PR #108, opened 2016-10-24
Bug 773436, use columnnames when adding data to: gnclock, numtest, versions

puggan:
If you add a field to a table, for example a timestamp to gnclock, the program can no longer load the data.

`ALTER TABLE gnclock ADD ts TIMESTAMP NOT NULL DEFALUT CURRENT_TIMESTAMP;`

If all insert-queries specify what columns they inserting the data to, the queries will still work with a newer version of the database.

So if a future version of gnucash need an extra column, it will not lock out the older version of gnucash, likewise if some external tool, want an extra column in a table, gnucash will still work.


jralls:
The bug is closed as invalid and this PR accordingly rejected. There isn't any reason to do this.

PR #91, opened 2016-06-12
Added forward declaration for struct GncGUID

limitedAtonement:
I also removed the guid.hpp include where I could.

PR #109, opened 2016-10-24
Guid header

limitedAtonement:
(As a refresher: https://github.com/Gnucash/gnucash/pull/91)

A new guid class has been written for use with C++. It doesn't extend a base class (such as boost::uuids::uuid), and it converts easily to and from the guid used in C code.

I wasn't sure how super secret we should be about the boost uuid thing; I was thinking of changing
`struct GUID : public boost::uuids::uuid`
to
`struct GUID : boost::uuids::uuid`
but instead I took the boost uuid out of the inheritance hierarchy. There is still a conversion constructor for it.


gjanssens:
Thanks for the new PR. It seems [travis has issues with it though](https://travis-ci.org/Gnucash/gnucash/builds/170229377).

limitedAtonement:
I see. I'll try a few force pushes to see if I can appease travis, and send another message back here when I'm done. It looks like travis is using gcc 4.8.4, but the one that comes with my distribution is 6.2.1.

limitedAtonement:
What is the docker image used by travis-ci? I think I see a few commands that are run when this docker container starts up https://travis-ci.org/Gnucash/gnucash/jobs/170240528/config.

jralls:
The problem Travis found is that [brace-initializer-list initialization](http://en.cppreference.com/w/cpp/language/list_initialization) is handled differently in C++11 and C++14. For  this cycle we need to stick to C++11, so you need to be careful about using initializer lists. Gcc should warn about that if you've got -std=c++11 set.

While you're at it, could you elaborate the last commit message to say that you've replaced inheritance from boost::guid::guid with composition?

Otherwise very nice. Composition is clearly a better choice here.

limitedAtonement:
Okay, made some changes to the initializer list handling. Using `export CXXFLAGS=-std=c++11` didn't help; I'm thinking there are some features gcc 4.8 just doesn't do? I understand initializer lists changed between 11 and 14 (and 17), but I'm not comfortable enough with the details to say whether this should be working in a c++11 compiler or not.

Anyway, Travis seems happy.

jralls:
Thanks!

PR #110, opened 2016-10-28
C++ backend

jralls:


jralls:
My turn for a code review. This branch isn't complete but it does pass tests and is ready for an interim merge so that other work can start to be based on it.

src/backend/sql/gnc-backend-sql.h
```diff
@@ -372,23 +372,23 @@ struct GncSqlColumnInfo
     GncSqlColumnInfo (std::string&& name, GncSqlBasicColumnType type,
                       unsigned int size = 0, bool unicode = false,
                       bool autoinc = false, bool primary = false,
-                      bool null_allowed = false) :
+                      bool not_null = false) :
```
gjanssens: You invert the logic of null_allowed to not_null throughout the code, except for the default value. This means a column is now by default allowed to store null values, where it wasn't before. Is this your intention ?

src/backend/sql/gnc-backend-sql.h
```diff
@@ -44,6 +45,13 @@ extern "C"
 #include "qofbackend-p.h"
 #include <gmodule.h>
 }
+
+#include <algorithm>
+#include <sstream>
+#include <string>
+#include <vector>
+
```
jralls: It's still more of a module header than a class header, so I haven't renamed it yet.

src/backend/sql/gnc-slots-sql.cpp
```diff
@@ -1100,6 +1100,10 @@ gnc_sql_init_slots_handler (void)
     static GncSqlObjectBackend be_data =
     {
         GNC_SQL_BACKEND_VERSION,
+// This was GNC_ID_ACCOUNT. If somethine blows up, change it back,
+// make the registry store a std::tuple<std::string,
+// GncSqlObjectBackendPtr>, and check the first string against types
+// in the functions that are called on each backend.
         GNC_ID_ACCOUNT,
```
jralls: I don't remember now what I tried, but since the registry stores a tuple whatever it was must have blown up.

src/backend/sql/gnc-address-sql.cpp
```diff
@@ -60,33 +60,32 @@ static GncSqlColumnTableEntry col_table[] =
...
 
+
     g_return_if_fail (be != NULL);
     g_return_if_fail (row != NULL);
     g_return_if_fail (pObject != NULL);
-    g_return_if_fail (table_row != NULL);
 
     addr = gncAddressCreate (be->book, QOF_INSTANCE(pObject));
-    for (subtable = col_table; subtable->col_name != NULL; subtable++)
+    for (auto const& subtable_row : col_table)
```
jralls: auto would get me a copy of the GncSqlColumnTableEntry and I need a const reference to it instead for polymorphism.

src/backend/sql/gnc-backend-sql.h
jralls: Yes. Making not_null the default blew up because there are far more columns that need to accept null values than there are ones that shouldn't. Note as well that this code is gone; the  GncSqlColumnInfo ctor now takes an or'd set of flags that correspond directly to the flags one passes to CREATE TABLE.

src/backend/dbi/gnc-backend-dbi.cpp
```diff
@@ -2156,143 +2156,95 @@ gnc_module_finalize_backend_dbi (void)
...
+    gnc_pop_locale (LC_NUMERIC);
+    return retval;
+}
+time64
+GncDbiSqlRow::get_time64_at_col (const char* col)
+{
+    auto type = dbi_result_get_field_type (m_result, col);
+    auto attrs = dbi_result_get_field_attribs (m_result, col);
+    if (type != DBI_TYPE_DATETIME)
+        throw (std::invalid_argument{"Requested double from non-double column."});
```
jralls: Yup, thanks. Fixed in its current location.

src/backend/sql/gnc-backend-sql.h
```diff
@@ -211,39 +213,82 @@ struct GncSqlConnection
...
 
 /**
- * @struct GncSqlResult
+ * Row of SQL Query results.
  *
- * Struct used to represent the result of an SQL SELECT statement.  SQL
- * backends must provide a structure which implements all of the functions.
+ * This is a "pointer" class of a pimpl pattern, the implementation being
+ * GncSqlResul::IteratorImpl. It's designed to present a std::forward_iterator
+ * like interface for use with range-for while allowing for wrapping a C API.
```
jralls: Thanks, fixed.

src/backend/sql/gnc-backend-sql.cpp
```diff
@@ -136,14 +136,29 @@ gnc_sql_register_backend(OBEEntry&& entry)
...
+GncSqlObjectBackendPtr
+gnc_sql_get_object_backend(const std::string& type)
+{
+    auto entry = std::find_if(backend_registry.begin(),
+                              backend_registry.end(),
+                              [type](const OBEEntry& entry){
+                                  return type == std::get<0>(entry);
+                              });
+    auto obe = std::get<1>(*entry);
+    if (entry != backend_registry.end())
```
jralls: Thanks, fixed.

gjanssens:
@jralls I was reading the current sql backend code this morning and was confused by the double use of 'be' as name for both a qofbackend and an sqlbackend. Can I suggest to deduplicate that name to 'qbe' and 'sbe' or similar ? That would definitely improve the readability of the code.

jralls:
OK, done with a slight difference: GncSqlBackend* are sql_be, GncDbiBackend* are dbi_be, and QofBackend* are qof_be.

Note that the distinction is temporary. The ultimate goal is to have GncSqlBackend and GncXmlBackend subclasses of QofBackend, and for GncDbiBackend to be a set of templated (on the underlying database) subclasses of GncSqlBackend. Once we've gotten to that point it will all of the backend references will disappear.

limitedAtonement:
I haven't said anything, but I've been watching. The changes look very good. I noticed the default null specification change, and I think it's good. @gjanssens has had some excellent comments; good eye.

gjanssens:
@jralls After I sent my comment it dawned on me the backends would probably still move. But thanks for renaming. It will help me dig into the backend code a bit easier. I'll note you've also introduced obe for GncSqlObjectBackend, which actually triggered my suggestion.

@limitedAtonement Thanks :)

PR #111, opened 2016-11-16
Webkit rep

Bob-IT:
This is the pull request for maint, I may of gone too far but have a look. This request incorporates all the webkit changes and the plot size option that will load the new pair from master, if it is a pixel pair, the value is used, if percentage, the default is used. If the reports are saved, then they updated with the old values and can see no way round this. There are also changes to the date-format widget with a new option 'No Fancy Date Format' which if set will return the format used in preference/date.

It all seems to work, if added to maint, or you pick the ones that can, I will probably start a new pull request with the rest of the changes made on master.

Bob


gjanssens:
Thanks. I believe all of this fits on the stable branch. I needed some time to reflect on the commit that adds the plot size option. My first reflex was - why not simple extend the existing number-range option instead of duplicating it. However then I noted the number-range option is used in many other places which don't need the extra check. So I believe your decision is best.
I tested this on maint and all seems to work fine, including make check.

PR #112, opened 2016-12-01
In response to bug 763640 followup

Bob-IT:
OK I failed at getting the rebase to work for me so I have reapplied the patches from pull 107 to form a new branch. I think this is all that has been discussed.
Bob


gjanssens:
Merged after a rebase. Thanks a lot!

PR #113, opened 2016-12-06
fix missing chmod +x in configure.ac

rtucker:
There is a missing "chmod +x" in configure.ac which, somehow, breaks compiles on my system.  This is fixed in this PR.


jralls:
Thanks, but we require changes that affect both branches to start in maint so they can merge up to master. This is trivial so I've made that change, attributed it to you, and pushed it to maint rather than using the PR directly.

rtucker:
Copy that, apologies for the inconvenience.  Thanks for the heads up!  -rt

PR #115, opened 2016-12-16
README typo

phil-davis:
I think this was intended to be "directly from Git".

### DIFF
--- a/README
+++ b/README
@@ -211,7 +211,7 @@ Building & Installing
 
 GnuCash uses GNU Automake to handle the build process, so for most of
 the details, see the generic instructions in INSTALL.  (If you are
-building directory from Git, read the README.git file for more instructions.)
+building directly from Git, read the README.git file for more instructions.)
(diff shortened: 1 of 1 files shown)


phil-davis:
Various docs talk about submitting patch files to other (older?) repositories. I guess these were previous to GitHub, and it seems that the code is now being managed via GitHub Pull Requests.
Is that so?
Should documentation be updated to say that?

jralls:
We use the ```master``` branch only for new development with all maintenance work like bug fixes committed to ```maint```. Since this change is trivial I opted to make it directly, crediting you as author, instead of rebasing it onto the correct branch.

As for the "various docs", if they refer explicitly to CVS or SVN repositories, they're obsolete and should either be deleted or revised. Our patch policy is at http://wiki.gnucash.org/wiki/Development#Submitting_Patches and any documents should refer to that page rather than repeating the instructions.

PR #117, opened 2017-01-07
configure: fix --disable-debug, --disable-profile and --disable-locale-specific-tax

stefantalpalaru:
All these options actually enabled the related features.


gjanssens:
Thanks for spotting this. Merged into maint. I added a follow-up commit to drop DEBUG_MEMORY which isn't used any more indeed.

stefantalpalaru:
You're welcome.

PR #120, opened 2017-01-28
Budget balance sheet, Liabilities section, negated sign

iulianu:
The Budget Balance Sheet report currently shows the budgeted liabilities with the wrong sign, for example it shows that liabilities _increase_ when you budget a liability repayment.
My understanding is that liabilities are budgeted with a negated sign. Positive amount means new liabilities, negative amount means liability repayment. 
This PR fixes the report for me.

### DIFF
--- a/src/report/standard-reports/budget-balance-sheet.scm
+++ b/src/report/standard-reports/budget-balance-sheet.scm
@@ -522,7 +522,7 @@
             (get-assoc-account-balances-budget
               budget
               liability-accounts
-              get-budget-account-budget-balance))
+              get-budget-account-budget-balance-negated))
 
           (set! liability-get-balance-fn
             (lambda (account start-date end-date)
@@ -566,7 +566,7 @@
 
           ;; Budgeted liabilities are liability repayments (negative liabilities).
           (set! liability-repayments
-            (get-assoc-account-balances-total-negated liability-account-budget-balances))
+            (gnc:get-assoc-account-balances-total liability-account-budget-balances))
 
           ;; New liabilities are then negated liability repayments.
           (set! new-liabilities
--- a/src/report/standard-reports/budget-balance-sheet.scm
+++ b/src/report/standard-reports/budget-balance-sheet.scm
@@ -566,7 +566,7 @@
 
           ;; Budgeted liabilities are liability repayments (negative liabilities).
           (set! liability-repayments
-            (gnc:get-assoc-account-balances-total liability-account-budget-balances))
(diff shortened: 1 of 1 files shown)


jralls:
What do you have selected for Preferences>Accounts, Reverse Balanced Accounts, and does changing it have any impact on the report?

iulianu:
My 'Reverse Balanced Accounts' setting is set to 'Credit accounts', changing it does not have an effect on the report.

Please find a test case attached for this scenario. It shows an existing liability of 1000 RON and a repayment of 900 RON. Also attached is the expected budget balance sheet report, and the actual report I am getting.
[PR120.zip](https://github.com/Gnucash/gnucash/files/752858/PR120.zip)

iulianu:
My 'Reverse Balanced Accounts' setting is set to 'Credit accounts', changing it does not have an effect on the report.

Please find a test case attached for this scenario. It shows an existing liability of 1000 RON and a repayment of 900 RON, budgeted, and executed in January. Also attached is the expected budget balance sheet report, and the actual report I am getting.
[PR120.zip](https://github.com/Gnucash/gnucash/files/752883/PR120.zip)

gjanssens:
Thanks for sending in the pull request, including an example file.

First a disclaimer: I don't use budgetting and know little of how it's supposed to work.

However having said that I have looked at your example file. (For reference if others wish to try this: I had to open the budget balance sheet report and reset the report currency to RON, otherwise there were no values shown).

It looks to me the issue is in how you entered values in the budget itself. GnuCash is double-entry so all values should balance. That means if you credit your bank account (withdraw money so a negative number on your budget), there should somewhere in your budget be a balancing amount. So from how I see it your credit card amount should be positive to effectively balance. Entering the numbers as such makes the Budget Balance Sheet behave correctly for me.

Considering my disclaimer I'd need more feedback from people really using the budget feature before I can take action on your pull request. I'll ask around on the mailing lists.

gjanssens:
[time passes]

The thread on gnucash-user was very insightful. Edward Doolittle did a great job analysing the current state of the budget code with respect to the Reverse Balanced Accounts preference and confirms what I suspected (though not explicitly mentioned on this PR). The budget is ignoring this preference and hence gives different results depending on the state of that preference.

Edward also explained how values should be entered in assets or liabilities depending the state of the preference. Using his signs, it appears the budget balance sheet report is showing the proper amounts without your patch unless I'm reading it incorrectly.

So unless you think I'm wrong (which is very possible as I'm not using the budget feature), I will not apply this patch. I do appreciate the effort you spent creating it though. So thanks!

PR #123, opened 2017-02-03
Numeric

jralls:
This fails on Ubuntu 14.04 because of a regex compilation problem which I believe is a bug in gcc-4.8's C++ standard library; I believe that because gcc-6.2.1 on F25 and clang on MacOS both pass. F25 also has a single rounding failure in test-gnc-rational.  I'll get those fixed soon, but I want to get this up for code review in parallel.

The plan for Ubuntu 14.04 is to replace std::regex with boost::regex. I haven't investigated the rounding error enough to know what exactly is the problem, though the test compares the results from converting a product to denominator 100 before and after calling round_to_numeric() on it.

### DIFF
--- a/src/libqof/qof/gnc-numeric.cpp
+++ b/src/libqof/qof/gnc-numeric.cpp
@@ -44,6 +44,8 @@ extern "C"
 #include "gnc-rational.hpp"
 
 static QofLogModule log_module = "qof";
+
+static const uint8_t max_leg_digits{17};
 static const int64_t pten[] = { 1, 10, 100, 1000, 10000, 100000, 1000000,
                                10000000, 100000000, 1000000000,
                                INT64_C(10000000000), INT64_C(100000000000),
@@ -57,8 +59,8 @@ static const int64_t pten[] = { 1, 10, 100, 1000, 10000, 100000, 1000000,
 int64_t
 powten (unsigned int exp)
 {
-    if (exp > 17)
-        exp = 17;
+    if (exp > max_leg_digits)
+        exp = max_leg_digits;
     return pten[exp];
 }
 
@@ -80,7 +82,8 @@ GncNumeric::GncNumeric(GncRational rr)
 
 GncNumeric::GncNumeric(double d) : m_num(0), m_den(1)
 {
-    if (isnan(d) || fabs(d) > 1e18)
+    static uint64_t max_leg_value{INT64_C(1000000000000000000)};
+    if (isnan(d) || fabs(d) > max_leg_value)
     {
         std::ostringstream msg;
         msg << "Unable to construct a GncNumeric from " << d << ".\n";
@@ -90,9 +93,9 @@ GncNumeric::GncNumeric(double d) : m_num(0), m_den(1)
     auto logval = log10(fabs(d));
     int64_t den;
     if (logval > 0.0)
-        den = powten(18 - static_cast<int>(floor(logval) + 1.0));
+        den = powten((max_leg_digits + 1) - static_cast<int>(floor(logval) + 1.0));
     else
-        den = powten(17);
+        den = powten(max_leg_digits);
     auto num = static_cast<int64_t>(floor(static_cast<double>(den) * d));
 
     if (num == 0)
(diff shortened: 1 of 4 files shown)


src/libqof/qof/gnc-rational.hpp
```diff
@@ -19,75 +19,261 @@
...
+}
+GncRational operator/(GncRational a, GncRational b);
+inline GncRational operator/(GncRational a, GncInt128 b)
+{
+    return a / GncRational(b, 1);
+}
+inline GncRational operator/(GncInt128 a, GncRational b)
+{
+    return b / GncRational(a, 1);
+}
```
gjanssens: This looks like a copy/paste error.
I assume you meant:
`return GncRational(a, 1) / b;`
?

src/libqof/qof/gnc-numeric.cpp
```diff
@@ -21,46 +21,501 @@
...
+    if (max_places > 17)
+        max_places = 17;
+    bool is_pwr_ten = true;
+    for (int pwr = 1; pwr <= 17 && m_den > powten(pwr); ++pwr)
+        if (m_den % powten(pwr))
+        {
+            is_pwr_ten = false;
+            break;
+        }
+
```
gjanssens: It looks like this will give invalid results if m_denom < 10.
In that case the for loop is never executed, yet the GncNumeric is not necessarily a power of ten. I discovered this by chance because my data holds an amount of 21/2. This will eventually result in an Arithmetic exception being thrown in xaccPrintAmountInternal because it makes reductions that assume the denominator is really a power of ten (in the while loop starting at src/app-utils/gnc-ui-utils.c:1509).
One way to mitigate this can be to properly initialize is_pwr_ten:
`bool is_pwr_ten = (m_den >= 10);`

src/libqof/qof/gnc-numeric.cpp
```diff
@@ -21,46 +21,501 @@
...
+    std::ostringstream out;
+    out << *this;
+    return out.str();
+}
+
+GncNumeric
+GncNumeric::to_decimal(unsigned int max_places) const
+{
+    if (max_places > 17)
+        max_places = 17;
```
gjanssens: While I'm in the area, perhaps this magic number is best defined somewhere near the top ?

src/libqof/qof/gnc-numeric.cpp
gjanssens: And this is still not enough as it still produces warnings for numbers with a denominator of 1 (which strictly speaking as also a power of 10). Testing with
`bool is_pwr_ten = (m_den == 1 || m_den >= 10);`
resolved all of the warnings and exceptions.

src/libqof/qof/gnc-numeric.hpp
```diff
@@ -0,0 +1,418 @@
...
+std::basic_ostream<charT, traits>& operator<<(std::basic_ostream<charT, traits>& s, GncNumeric n)
+{
+    std::basic_ostringstream<charT, traits> ss;
+    ss.copyfmt(s);
+    ss.width(0);
+    if (n.denom() == 1)
+         ss << n.num();
+    else if (n.denom() % 10 == 0)
+        ss << n.num() / n.denom() << "."
+           << (n.num() > 0 ? n.num() : -n.num()) % n.denom();
```
jralls: You're right, and not only that the condition is wrong: a denom of e.g. 130 would run that code and produce an incorrect representation.

src/libqof/qof/gnc-numeric.hpp
```diff
@@ -0,0 +1,418 @@
...
+     * @return 0 if this == 0 else 1/this.
+     */
+    GncNumeric inv() const noexcept;
+    /**
+     * @return -this if this < 0 else this.
+     */
+    GncNumeric abs() const noexcept;
+    /**
+     * Reduce this to an equivalent fraction with the least common multiple as
+     * the denominator.
```
jralls: How about "Return an equivalent fraction with all common factors between the numerator and denominator removed." ?

src/libqof/qof/gnc-rational.hpp
jralls: Maybe a "didn't think hard enough when copy-pasting error". operator-() was similarly wrong.

gjanssens:
Op maandag 20 februari 2017 09:30:49 CET schreef John Ralls:
> jralls commented on this pull request.
> 
> > +    int64_t denom() const noexcept { return m_den; }
> 
> +    /**
> +     * @return A GncNumeric with the opposite sign.
> +     */
> +    GncNumeric operator-() const noexcept;
> +    /**
> +     * @return 0 if this == 0 else 1/this.
> +     */
> +    GncNumeric inv() const noexcept;
> +    /**
> +     * @return -this if this < 0 else this.
> +     */
> +    GncNumeric abs() const noexcept;
> +    /**
> +     * Reduce this to an equivalent fraction with the least common multiple
> as +     * the denominator.
> 
> How about "Return an equivalent fraction with all common factors between the
> numerator and denominator removed." ?

That will do fine.

src/libqof/qof/gnc-numeric.cpp
```diff
@@ -21,46 +21,501 @@
...
+    m_num = r.num();
+    m_den = r.denom();
+}
+
+using boost::regex;
+using boost::smatch;
+using boost::regex_search;
+GncNumeric::GncNumeric(const std::string& str, bool autoround)
+{
+    static const std::string numer_frag("(-?[0-9]+)");
```
jralls: I tried to get operator>> to accept thousands separators, but it looks like the llvm libstdc++ doesn't support that, at least as of what Apple included in their latest Xcode.

For now I think we continue to use xaccPrintNumeric and xaccParseNumeric and look at merging that functionality into GncNumeric later when time permits.

PR #124, opened 2017-02-06
Add gncEntryGetPrice to enable net price in reports 

hcrohland:
Add gncEntryGetNetPrice
Create an option in taxinvoice to either use gncEntryGetNetPrice or
gncEntryGetPrice

So far taxinvoice would show net or gross prices dependent on the internal flag tax_included.
This is inconsistent for the reader of the invoice since there is no notion of that flag.
This patch adds the option to always show net prices.
It does not change the default behaviour even if I would consider it broken.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -20,7 +20,7 @@ dnl Process this file with autoconf to produce a configure script.
 
 # Autoconf initialization
 AC_PREREQ(2.60)
-AC_INIT([GnuCash], [2.6.15], [gnucash-devel@gnucash.org])
+AC_INIT([GnuCash], [2.6.15], [https://bugzilla.gnome.org/page.cgi?id=browse.html&product=GnuCash], , [http://www.gnucash.org/])
 AC_CONFIG_HEADERS(config.h)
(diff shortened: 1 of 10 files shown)


hcrohland:
Hi Geert,

I wasn't aware that this would be visible to you without further action.

My idea was to split it up into two requests: The backend part and the use 
in the report.

The backend changes would apply cleanly to both maint and master. And they 
would not have new strings.

The report changes I would only submit for master.

With that one could easily make a local change in the 2.6 report to show 
the net values

What do you think?

Gruß,
Christoph



Am 1. März 2017 6:42:35 nachm. schrieb Geert Janssens 
<notifications@github.com>:

gjanssens:
Ah, that sounds reasonable. I can agree with that and I'll wait until you're ready until I comment again :)

hcrohland:
Hi Geert,

this should now be ready for pulling. It only adds the backend functionality as discussed. 
I have the changes for master ready. I will create the corresponding pull request as soon as I have the local build of master running again. I am fighting with the Mac build...

Christoph

PR #127, opened 2017-02-18
Export gnc_get_current_session to guile bindings

sdementen:
this allows to retrieve the name of a book in a report by using (qof-session-get-url (gnc-get-current-session))


jralls:
OK, but don't you want it in ```maint``` so that you can use it after the next release instead of having to wait until 2.8.0 at the end of the year?

sdementen:
yes, i did a new one on the maint branch ... but forgot to start from main (instead of master ...)
i will correct both ...

PR #130, opened 2017-03-08
Let taxinvoice optionally show the net instead of the given price

hcrohland:
This is needed if there is a mix of entries with tax included and
excluded.


gjanssens:
Committed after a rebase. Thanks a lot!

PR #132, opened 2017-03-09
Charge back net value of bill entries in invoices - Bug 776380

hcrohland:
When charging entries from bills we need to exclude taxes since the tax
rate of the invoice might be different than that of the bill.

This fixes Bug 776380 - Gross value of bills charged back instead of
net value (https://bugzilla.gnome.org/show_bug.cgi?id=776380)


gjanssens:
Committed after a rebase. Thanks!

PR #133, opened 2017-03-10
Fix taxinvoice for bill entries with taxes

hcrohland:
For bills taxinvoice was throwing an exception when there were taxes on bill entries.
It also did scan the invoice members instead of the bill members to determine if there were taxes and discounts.

This should merge without problem with my previous change on taxinvoice.eguile.scm. I tested that too.

### DIFF
--- a/src/report/business-reports/taxinvoice.eguile.scm
+++ b/src/report/business-reports/taxinvoice.eguile.scm
@@ -82,18 +82,30 @@
                          (< (car (gnc-transaction-get-date-posted t1))
                             (car (gnc-transaction-get-date-posted t2))))))))
 
+
+      ;; Is this an invoice or something else
+      (if (not (null? opt-invoice))
+          (begin
+            (set! owner (gncInvoiceGetOwner  opt-invoice))
+            (let ((type (gncInvoiceGetType  opt-invoice)))
+              (cond
+               ((eqv? type GNC-INVOICE-CUST-INVOICE)
+                (set! cust-doc? #t))
+               
+               ))))
+      
       ; pre-scan invoice entries to look for discounts and taxes
       (for entry in entries do
           (let ((action    (gncEntryGetAction entry))
                 (qty       (gncEntryGetDocQuantity entry credit-note?))
                 (discount  (gncEntryGetInvDiscount entry))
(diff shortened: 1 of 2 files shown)


hcrohland:
I do not know why the tests fail. It seems to me that the failure is unrelated to my change.

src/report/business-reports/taxinvoice.eguile.scm
```diff
@@ -82,18 +82,30 @@
...
-                (taxtable  (gncEntryGetInvTaxTable entry)))
+                (taxable?  (if cust-doc? (gncEntryGetInvTaxable entry) (gncEntryGetBillTaxable entry)))
+                (taxtable  (if cust-doc? (gncEntryGetInvTaxTable entry) (gncEntryGetBillTaxTable entry))))
             (if (not (string=? action ""))
               (set! units? #t))
             (if (not (= (gnc-numeric-to-double qty) 1.0))
               (set! qty? #t))
-            (if (not (gnc-numeric-zero-p discount)) (set! discount? #t))
+            (if (not (or cust-doc? (gnc-numeric-zero-p discount))) (set! discount? #t))
             ;(if taxable - no, this flag is redundant
```
gjanssens: This looks like a logic error to me ? Assuming discounts are only possible for customer documents the condition is:

discount? is true if we're handling a customer document and and the discount value is non zero

Or translated into scheme

`(and cust-doc? (not (gnc-numeric-zero-p discount)))`

If you want to switch this to an or form it would become

`(not (or (not cust-doc?) (gnc-numeric-zero-p discount)))`

Feel free to pick your preferred format.

gjanssens:
I have noticed the intermittent failure of test-imap-map as well. I've even opened a bug report for it. Unfortunately the original author of the test (and the imap-map code) can't reproduce. And I've never been able to reproduce outside of travis as well. The fact it's intermittent makes this pretty difficult to debug.

hcrohland:
Oh, my bad. Wanted to check that later and forgot :-(

Made a third version which is better readable and in line with the style used a few lines below.

Gruß,
Christoph

gjanssens:
An even better readable option indeed. I've committed your actual invoice code after a rebase. I skipped the travis test code as we already have several builds now that illustrate the bug and print the test log.

PR #134, opened 2017-03-12

christopherlam:
Update MUR symbol.


fellen:
Thanks. But we decided, to choose the ligature '₨'.

PR #141, opened 2017-04-16
Fix crash while saving scheduled transaction template

nitinjavakid:
Currently Gnucash crashed when you try to save scheduled transactions. This fixes the issue.


jralls:
Good catch, thanks!

PR #142, opened 2017-04-18
Prices import

Bob-IT:
There have been several requests to import prices so here is one based on the new CSV importer.

I have kept this mostly separate from existing transaction importer as Geert was still going to make some changes so there is some duplication in the amount / date parsers which I believe might be moved.

I have added to the settings file and in the last commits changed the name as it potentially will not just have transaction settings. All seems to work as far as I can tell.

Bob


gjanssens:
Bob, I haven't had time yet to review your work. Your PR comment however prompted me to finish moving the date and amount parsers from the csv importer into the engine. FYI I'm almost done with the date part. I'll get back to this when these two are finished.

gjanssens:
Oops! Didn't mean to close :(

Bob-IT:
No worries, I did not expect it to be done quickly, just really a heads up on what I did.

jralls:
@Bob-IT Bob, this now needs a rebase and I think @gjanssens is done with the date work on the csv-importer.

Bob-IT:
I have rebased on unstable, made some changes for Gtk3 and changed the date parser. I think it does what I wanted to achieve but still might need some changes. Will await feedback.
Bob

gjanssens:
I realize my review is long overdue... Sorry about that.

I'll add comments as I go through the code.

My first stop is at your changes in the CsvTransSettings class: the way it was extended to support both transaction import settings and prices import settings is very much C-style. In C++ you would start with a base class that handles everything that's common to all settings types you want to support. Everything that's unique per import type will then be delegated to derived classes.

So in this particular case, you'd have a generic base class. CsvImportSettings would be a good name for such a base class. This class can deal with common settings already.
Then make two derived classes, like CsvTransImpSettings and CsvPriceImpSettings. Both classes delegate handling of the common settings to the base class and add whatever is needed for their type specific settings.

All code that needs a settings object can then instantiate one of the derived classes, depending on the need. Like that there is no need to globally define a settings_type to use in if statements all over the place.

It's probably possible to abstract this away even further, but for now this will suffice. I've also considered a template class, but I'm not sure that would be a good fit in this case (and tbh, I'm not that fluent in templates either).

gnucash/import-export/csv-imp/gnc-price-import.cpp
```diff
@@ -0,0 +1,758 @@
...
+
+const int num_currency_formats_price = 3;
+const gchar* currency_format_user_price[] = {N_("Locale"),
+                                       N_("Period: 123,456.78"),
+                                       N_("Comma: 123.456,78")
+                                      };
+
+
+/** Constructor for GncPriceImport.
+ * @return Pointer to a new GncCsvParseData
```
gjanssens: This is still a reference to the old c struct...

gjanssens:
@Bob-IT I remember you indicated in the past your C++ experience was fairly limited. If you want I can work with you on the changes I proposed to get this feature ready to get merged.

Bob-IT:
Geert, yes it is limited so will need to do some reading to understand your comments but it gives me an opportunity to learn. It will take a few days and will report back here on any or no progress. thanks for the offer of help.
Bob

Bob-IT:
Here is an attempt to change the way the import settings are handled along with the other easy updates.
Bob

gnucash/import-export/csv-imp/gnc-csv-price-import-settings.cpp
```diff
@@ -0,0 +1,258 @@
...
+#define CSV_COL_TYPES    "ColumnTypes"
+
+#define CSV_TO_CURR      "PriceToCurrency"
+#define CSV_FROM_COMM    "PriceFromCommodity"
+
+G_GNUC_UNUSED static QofLogModule log_module = GNC_MOD_IMPORT;
+
+preset_vec_price presets_price;
+
+static std::shared_ptr<CsvPriceImpSettings> create_int_no_preset(void)
```
gjanssens: Aside the template specialization this function is identical across settings types. So it would make sense to share the code as well. One way to do this would be to add the function as a templatized, static, protected class member to CsvImportSettings:
```
template<class SType>static std::shared_ptr<CsvImportSettings> create_no_preset()
```
I'm not 100% sure of the location of the static keyword. Note also the generic class type I'm returning rather than the specific one per import type (see other comments for the reasoning).

As your get_import_presets_* functions need to be friends of that class anyway, they can still call this common function as
```
CsvImportSettings::create_no_preset<CsvPriceImpSettings>()
```
for CsvTransImpSettings you only need to change the template specialization to <CsvTransImpSettings>

gnucash/import-export/csv-imp/gnc-csv-price-import-settings.cpp
```diff
@@ -0,0 +1,258 @@
...
+{
+    auto preset = std::make_shared<CsvPriceImpSettings>();
+    preset->m_name = get_no_settings();
+    preset->m_settings_type = settings_type;
+
+    return preset;
+}
+
+static std::shared_ptr<CsvPriceImpSettings> create_int_gnc_exp_preset(void)
```
gjanssens: This function can get a similar treatment as the no_preset variant. The primary difference here is that I don't know if all future importers we convert to this more generic import class structure will always have exactly one gnucash specific export format preset or that it's related to gnucash export features at all. 

(Sidenote, if we expand this in the future, the "Csv" part in many of the file and function names may have to be removed still, but that's not important now).

So for this one I would consider a more generic approach of adding a virtual function get_gnucash_presets (or some other name that conveys one or more presets that come with the system rather than user defined). This function should be overridden by each derived class. It's not possible to make this one common function for all because each derived class has a unique set of specialized settings. The function would return a vector of CsvImportSettings. The calling functions can then add these to the list of available presets they manage (after the "no_preset" and before user presets).

gnucash/import-export/csv-imp/gnc-csv-price-import-settings.cpp
```diff
@@ -0,0 +1,258 @@
...
+    return preset;
+}
+
+/**************************************************
+ * find
+ *
+ * find all settings entries in the state key file
+ * based on settings type.
+ **************************************************/
+const preset_vec_price& get_import_presets_price (void)
```
gjanssens: Other than a way to get the group prefix and whether or not there are export format presets, this code is identical across the two specializations we currently have. So this is another candidate to move to a common location.

The most logical place would be a public class method of CsvImportSettings. However as the function is used regardless of whether there are CsvImportSettings instances, it should again be declared static. To get the proper group prefix and the correct no_settings and other type specific internal presets, we do need to know the settings type though. So another template is in order:
```
template<class SType> const preset_vector& get_import_presets (void);
```
(Note the use of preset_vector rather than preset_vector_price, explained in another comment).
```

File: gnucash/import-export/csv-imp/gnc-csv-price-import-settings.cpp [hunk 5]
```diff
@@ -0,0 +1,258 @@
...
+    auto preset_names = std::vector<std::string>();
+    auto keyfile = gnc_state_get_current ();
+    gsize grouplength;
+    gchar **groups = g_key_file_get_groups (keyfile, &grouplength);
+
+    /* Start by building a sorted list of candidate presets as found in the state file */
+    for (gsize i=0; i < grouplength; i++)
+    {
+        auto group = std::string(groups[i]);
+        auto gp = get_prefix() + settings_type + " - ";
```
gjanssens: In the template function and with a static get_group_prefix() function defined per derived class, this would become
```
auto gp = SType::get_group_prefix()
```

And move it outside of the for loop for efficiency.

gnucash/import-export/csv-imp/gnc-csv-price-import-settings.cpp
```diff
@@ -0,0 +1,258 @@
...
+    g_strfreev (groups);
+
+    /* We want our settings to appear sorted alphabetically to the user */
+    std::sort(preset_names.begin(), preset_names.end());
+
+    /* Now add each preset to our global list */
+    presets_price.clear();
+
+    /* Start with the internally generated ones */
+    presets_price.push_back(create_int_no_preset());
```
gjanssens: Becomes
```
presets.price.push_back (create_no_preset<SType> ())
```

gnucash/import-export/csv-imp/gnc-csv-price-import-settings.cpp
```diff
@@ -0,0 +1,258 @@
...
+    //presets_price.push_back(create_int_gnc_exp_preset()); // Not Required
+
+    /* Then add all the ones we found in the state file */
+    for (auto preset_name : preset_names)
+    {
+        auto preset = std::make_shared<CsvPriceImpSettings>();
```
gjanssens:
```
auto preset = std::make_shared<SType>();
```

gnucash/import-export/csv-imp/gnc-csv-price-import-settings.cpp
```diff
@@ -0,0 +1,258 @@
...
+        preset->m_settings_type = settings_type;
```
gjanssens: Can be removed.

gnucash/import-export/csv-imp/gnc-csv-price-import-settings.cpp
```diff
@@ -0,0 +1,258 @@
...
+        preset->m_name = preset_name;
+        preset->load();
+        presets_price.push_back(preset);
```
gjanssens: This **may** need a cast to the parent class, but I suspect C++ will do this for you.

gnucash/import-export/csv-imp/gnc-csv-price-import-settings.cpp
```diff
@@ -0,0 +1,258 @@
...
+    if (preset_is_reserved_name (m_name))
+        return true;
+
+    GError *key_error = nullptr;
+    m_load_error = false;
+    auto keyfile = gnc_state_get_current ();
+    auto group = get_prefix() + m_settings_type + " - " + m_name;
+
+    // Start Loading the settings
+    m_load_error = load_common(); // load the common settings
```
gjanssens: Use ```CsvImportSettings::load()```. Similar for save_common() and so on below.

gjanssens:
Oh, and merry Christmass :)

jralls:
Sorry for not noticing sooner: C++ classes that have GObject member variables (that's anything derived from GtkWidget or QofInstance) must have custom destructors that g_object_unref the GObjects, otherwise they'll be leaked.

That said, I rather doubt that there's enough benefit to classes like CsvPriceImpAssist being C++ classes instead of GObject classes to outweigh the difficulties imposed by the different memory management and construction/destruction models.

If we really want to use C++ for GUI code we should use gtkmm where they've at least had a lot of time and practice reconciling the two models.

gjanssens:
Op zondag 24 december 2017 17:25:18 CET schreef John Ralls:
> Sorry for not noticing sooner: C++ classes that have GObject member
> variables (that's anything derived from GtkWidget or QofInstance) must have
> custom destructors that g_object_unref the GObjects, otherwise they'll be
> leaked.
> 
> That said, I rather doubt that there's enough benefit to classes like
> CsvPriceImpAssist being C++ classes instead of GObject classes to outweigh
> the difficulties imposed by the different memory management and
> construction/destruction models.
> 
> If we really want to use C++ for GUI code we should use gtkmm where they've
> at least had a lot of time and practice reconciling the two models.

As for the g_object_unref in a destructor, I **thought** I had done that, but 
you're right it's missing.

I didn't find that much difficulty otherwise in memory management or 
construction/destruction. Besides even if not written as a C++ class the 
memory management is tricky because the import functionality itself is written 
completely in C++. I'd even revert this statement: by not writing it as a C++ 
class I have to manage the lifecycle of **both** the C objects **and** the C++ 
ones. Now the class at least takes care of the C++ ones properly.

You make a fair point about gtkmm though. I'm hesitant there because
1. I know plain gtk pretty well, but can't say the same of gtkmm. So I have no 
idea of the potential pitfalls this would bring up.
2. I eventually want to switch to a true C++ gui framework in the future so 
don't feel like investing much in gtkmm.

Bob-IT:
Geert, thank you for your comments, more to think about probably in the new year and Merry Christmas to you all.
Bob

jralls:
>As for the g_object_unref in a destructor, I **thought** I had done that, but
you're right it's missing.

Darn, I didn't notice that in your original CSV importers. Also missing are explicit declarations of the copy and move constructors... they should probably be ```delete```, I doubt that we want to copy or move (or have operator=) for GUI elements. Not strictly required once you supply the destructor because that will suppress the compiler creating the default ones, but still good to explicitly suppress them to make it clear to humans.

>I didn't find that much difficulty otherwise in memory management or
construction/destruction. Besides even if not written as a C++ class the
memory management is tricky because the import functionality itself is written
completely in C++. I'd even revert this statement: by not writing it as a C++
class I have to manage the lifecycle of **both** the C objects **and** the C++
ones. Now the class at least takes care of the C++ ones properly.

After studying the code more, I see that the leakage problem isn't so bad, though there's a potential dangling pointer problem. Most of the widget pointers that are member variables are obtained with ```gtk_builder_get_object()```, which does not call ```g_object_ref()``` on them. They're owned by ```csv_imp_assist``` so when that's destroyed in the ```close``` handler all of the other widgets get destroyed too. The lone exception here is ```encselector```;  CsvImpTransAssist adds ```gnc_csv_importer_gui```. Those are owned directly by the C++ class and need to be cleaned up in the destructor. (But see below.)
 
Here's another, very serious, problem: In ```gnc_csv_foo_import``` you create a CsvImpFooAssist* with operator new and pass it to gnc_register_gui_component that adds it to a list of components maintained in libgnucash/app-utils/gnc-component-manager.c. The corresponding gnc_unregister_gui_component (called by the ```destroy``` signal handler) calls ```g_free()``` on the pointer. If that doesn't crash immediately (operator new tracks allocations separately from malloc) it frees the memory without calling the class destructor. 

Even if that were wired up correctly the ```destroy``` signal is separate from the ```close``` signal, and in between CsvImpFooAssist is populated with pointers to now deallocated memory. Any intervening operation on the class that tries to dereference any of them might crash or worse corrupt something.

That the import functionality  being in C++ isn't a reason to encapsulate the GtkAssistant in a class. Using the same assistant implementation as in assistant-csv-account-import in a .cpp file will allow you to access C++ features without the memory management consequences of the class wrapper.

>You make a fair point about gtkmm though. I'm hesitant there because
1. I know plain gtk pretty well, but can't say the same of gtkmm. So I have no
idea of the potential pitfalls this would bring up.
2. I eventually want to switch to a true C++ gui framework in the future so
don't feel like investing much in gtkmm.

I'm not enthusiastic about gtkmm either but mixing the GObject and C++ object models isn't easy and the gtkmm folks have been doing it for a long time. We have the choice of taking advantage of their experience, not trying to mix the two, or screwing things up as we recreate their experience. My preference is the middle one but the first is much better than the last.

gjanssens:
Op maandag 25 december 2017 19:38:01 CET schreef John Ralls:
> >As for the g_object_unref in a destructor, I **thought** I had done that,
> >but
> you're right it's missing.
> 
> Darn, I didn't notice that in your original CSV importers. Also missing are
> explicit declarations of the copy and move constructors... they should
> probably be ```delete```, I doubt that we want to copy or move (or have
> operator=) for GUI elements. Not strictly required once you supply the
> destructor because that will suppress the compiler creating the default
> ones, but still good to explicitly suppress them to make it clear to
> humans.

Yes, I should do that, thanks for reminding me.

> >I didn't find that much difficulty otherwise in memory management or
> > construction/destruction. Besides even if not written as a C++ class the
> > memory management is tricky because the import functionality itself is
> > written completely in C++. I'd even revert this statement: by not writing
> > it as a C++ class I have to manage the lifecycle of **both** the C objects
> > **and** the C++ ones. Now the class at least takes care of the C++ ones
> > properly.
> 
> After studying the code more, I see that the leakage problem isn't so bad,
> though there's a potential dangling pointer problem. Most of the widget
> pointers that are member variables are obtained with
> ```gtk_builder_get_object()```, which does not call ```g_object_ref()``` on
> them. They're owned by ```csv_imp_assist``` so when that's destroyed in the
> ```close``` handler all of the other widgets get destroyed too. The lone
> exception here is ```encselector```;

The memory management of encselector is not that clear to me. It's a 
GtkWidget. As far as I understand, by adding this widget to the dialog it 
becomes managed by the dialog from there on. Or should I ```g_object_unref``` 
it after adding it to the dialog widget ? This was not happening in the 
original code as far as I can tell so I figured I didn't need to either.

> CsvImpTransAssist adds
> ```gnc_csv_importer_gui```. Those are owned directly by the C++ class and
> need to be cleaned up in the destructor. (But see below.)

The ```GNCImportMainMatcher``` code is hopelessly convoluted and I didn't want 
to touch it yet to limit my scope. But there's no memory issue here as far as 
I understand. The ```gnc_csv_importer_gui``` is ```g_new0```'d in a call to 
```gnc_gen_trans_assist_new``` and it's ```g_freed()``` in the call to 
```gnc_gen_trans_list_delete```. That function is either called in 
```CsvImpTransAssist::assist_finish``` or internally by the main matcher that 
takes over for the final matching. This was no different in the original C 
code. Quite ugly and that should be cleaned up at some point.

> Here's another, very serious, problem: In ```gnc_csv_foo_import``` you
> create a CsvImpFooAssist* with operator new and pass it to
> gnc_register_gui_component that adds it to a list of components maintained
> in libgnucash/app-utils/gnc-component-manager.c. The corresponding
> gnc_unregister_gui_component (called by the ```destroy``` signal handler)
> calls ```g_free()``` on the pointer. If that doesn't crash immediately
> (operator new tracks allocations separately from malloc) it frees the
> memory without calling the class destructor.
> 
No, it does not. It calls ```g_free()``` and the internally managed 
```ComponentInfo* ci```. The ```CsvImpFooAssist``` is merely referenced as 
user_data pointer in that structure. The component manager doesn't do anything 
with that user_data except compare it to find a certain component. It's up to 
the calling code to free this, which I do in ```csv_tximp_assist_destroy_cb``` 
with a ```delete```.

> Even if that were wired up correctly the ```destroy``` signal is separate
> from the ```close``` signal, and in between CsvImpFooAssist is populated
> with pointers to now deallocated memory. Any intervening operation on the
> class that tries to dereference any of them might crash or worse corrupt
> something.
> 
It's wired up like this all over the gnucash code. So if you consider it wrong 
here, it's wrong everywhere. Not really a C++ / C mixture issue.

> That the import functionality  being in C++ isn't a reason to encapsulate
> the GtkAssistant in a class.

True.

> Using the same assistant implementation as in
> assistant-csv-account-import in a .cpp file will allow you to access C++
> features without the memory management consequences of the class wrapper.

I still don't see any real memory management consequences, except perhaps one 
must pay more attention to what to do when.

For me writing this as a class was almost a natural progression from the huge 
data structure that's already there. By making it a C++ class I could 
encapsulate all the functions operating on that data inside the same class. 
That saved me a lot of situations where I had to pass the data structure as an 
additional argument into the functions.

You could argue I still have to do this for all the widget callbacks. That is 
true. However I treated the callbacks as very simple wrappers that would call 
the appropriate function in the class. Perhaps this is overkill. I liked the 
clear separation and I still do.

> > You make a fair point about gtkmm though. I'm hesitant there because
> 
> > 1. I know plain gtk pretty well, but can't say the same of gtkmm. So I 
have
> > no idea of the potential pitfalls this would bring up.
> > 2. I eventually want to switch to a true C++ gui framework in the future 
so
> > don't feel like investing much in gtkmm.
> 
> I'm not enthusiastic about gtkmm either but mixing the GObject and C++
> object models isn't easy and the gtkmm folks have been doing it for a long
> time. We have the choice of taking advantage of their experience, not
> trying to mix the two, or screwing things up as we recreate their
> experience. My preference is the middle one but the first is much better
> than the last.

I don't think we can avoid mixing the two completely. There's always a point 
where the two models will intersect. But I'll check into gtkmm when I find 
some time. It won't be for 2.7.3 I'm afraid.

Out of curiosity, does gtkmm mix well with plain gtk ? I can't avoid using 
plain gtk altogether yet.

gjanssens:
Oh and by the way, thanks for digging in so thoroughly. I think it's good to reflect on these things carefully. It also reminds me I still have to unit test the whole thing (and probably try and improve the memory management and performance of the pure C++ parts of the csv importer).

jralls:
>No, it does not. It calls ```g_free()``` and the internally managed
```ComponentInfo* ci```. The ```CsvImpFooAssist``` is merely referenced as
user_data pointer in that structure. The component manager doesn't do anything
with that user_data except compare it to find a certain component. It's up to
the calling code to free this, which I do in ```csv_tximp_assist_destroy_cb```
with a ```delete```.

Ah, I see. I meant the line above, where it frees ci->component_class... but that's just a string. I somehow mixed it up with the ptr that goes into ci->user_data. Sorry.

jralls:
> It's wired up like this all over the gnucash code. So if you consider it wrong
here, it's wrong everywhere. Not really a C++ / C mixture issue.

No, it's right. The flow is the ```close``` handler calls ```gtk_widget_destroy()``` on csv_imp_asst and that emits the ```destroy``` signal the handler for which unregisters it from the component manager and deletes the CsvImpTransAssist. It took a few mental runs through all of the hops to see it, which suggests that some comments might be helpful.

jralls:
> The memory management of encselector is not that clear to me. It's a
GtkWidget. As far as I understand, by adding this widget to the dialog it
becomes managed by the dialog from there on. Or should I ```g_object_unref```
it after adding it to the dialog widget ? This was not happening in the
original code as far as I can tell so I figured I didn't need to either.

I think that this is correct. ```gtk_container_add``` takes ownership so the ```encselector``` should be destroyed with the rest of the GtkAssistant.

jralls:
>The ```GNCImportMainMatcher``` code is hopelessly convoluted and I didn't want
to touch it yet to limit my scope. But there's no memory issue here as far as
I understand. The ```gnc_csv_importer_gui``` is ```g_new0```'d in a call to
```gnc_gen_trans_assist_new``` and it's ```g_freed()``` in the call to
```gnc_gen_trans_list_delete```. That function is either called in
```CsvImpTransAssist::assist_finish``` or internally by the main matcher that
takes over for the final matching. This was no different in the original C
code. Quite ugly and that should be cleaned up at some point.

Hmm, a self-sinking object. No doubt the author thought it convenient. The only real issue with it is holding a pointer to it as a class member. ```assist_finish``` should null the member variable. It would be better if it was local to the page it lives on instead of part of the class.

jralls:
> Out of curiosity, does gtkmm mix well with plain gtk ? I can't avoid using
plain gtk altogether yet.

Well, gtkmm is just a wrapper around gtk, so according to the documentation it's possible to mix them. As long as one is just passing around pointers one is probably safe. References might produce some unexpected results.

jralls:
> I still don't see any real memory management consequences, except perhaps one
must pay more attention to what to do when.

You've shown that you have covered the memory management (well done!), but at considerable effort for both of us. I think that's an indication that the code is sufficiently unusual and spread around that it needs some explanatory comments.

jralls:
> For me writing this as a class was almost a natural progression from the huge
data structure that's already there. By making it a C++ class I could
encapsulate all the functions operating on that data inside the same class.
That saved me a lot of situations where I had to pass the data structure as an
additional argument into the functions.

> You could argue I still have to do this for all the widget callbacks. That is
true. However I treated the callbacks as very simple wrappers that would call
the appropriate function in the class. Perhaps this is overkill. I liked the
clear separation and I still do.

I'm less convinced of this. Aside from the Preview page (which looks to me like it deserves its own class) all of the class methods seem to be callbacks. Besides, you're still passing the argument, it's just hidden by the compiler.

gjanssens:
This thread has added a few interesting suggestions for improvement of the new importer code. I will evaluate whether gtkmm can help in this area. And  based on that I'll re-evaluate whether I'd keep the assistant as wrapped in a C++ class or not. In the end the code must be manageable by more devs than just me...

Back to @Bob-IT's work now. I'm having a timing question: next weekend we have planned to release 2.7.3. And if I understood correctly we'd like that version to be our feature freeze release. @Bob-IT you suggested a few comments back you would continue to work on this in the new year, which would be **after** the 2.7.3 release.

So what do we do with this prices import feature (and this branch) ? My feedback is mostly on coding style, not on whether it works or not. Is it working well enough to be included already ? If so, we probably should merge this branch before 2.7.3 and continue the coding style improvements afterwards. Is that then still for 2.8 or should that happen on master then ?

gjanssens:
Oh and in that context, I think I'll delay introducing gtkmm in any case until after 2.8. It's another dependency to manage on OS X and Windows as  well. I don't want to introduce new dependencies this late in the cycle. I prefer to continue stabilizing what's there to avoid any further delays.

Bob-IT:
Geert, I do not mind what is decided, there were people asking for a way to import prices so created one based on your work. I believe it does what it should as far as I understand it and if not included at least it has highlighted some improvements. I can continue to make the changes already mentioned in the new year.
Bob

gjanssens:
I have merged this into unstable for it to appear in 2.7.3. We'll continue to work on it for the 2.8. Thanks a lot for your work!

PR #143, opened 2017-04-18
Prices remove old prices

Bob-IT:
This request changes the way old prices can be removed. There is now an option to specify which commodities to remove prices from as well as options to leave a weekly Friday price, the last Friday of a monthly and a scaled option.

Bob

### DIFF
--- a/src/engine/gnc-pricedb.c
+++ b/src/engine/gnc-pricedb.c
@@ -1369,7 +1369,7 @@ free_data (gpointer data)
 }
 
 static void
-gnc_pricedb_remove_get_keep_dates (GHashTable* hash_dates, GDate period_begin, GDate period_end, PriceRemoveOptions leave)
+gnc_pricedb_remove_get_keep_dates (GHashTable* hash_dates, GDate period_begin, GDate period_end, PriceRemoveKeepOptions keep)
 {
     Recurrence *r;
     GDate recurrence_date_old, recurrence_date_next;
@@ -1386,11 +1386,12 @@ gnc_pricedb_remove_get_keep_dates (GHashTable* hash_dates, GDate period_begin, G
            g_date_get_month (&period_end), g_date_get_year (&period_end));
 
     g_date_set_day (&day_of_week_date, 1); // set date to 1st of month
+    g_date_subtract_days (&day_of_week_date, 1); // move to last day of previous month
 
     while (g_date_get_weekday (&day_of_week_date) != selected_day_of_week)
-        g_date_subtract_days (&day_of_week_date, 1); // go back till we find last Friday of month
+        g_date_subtract_days (&day_of_week_date, 1); // go back till we find last Friday of previous month
 
-    if (leave == PRICE_REMOVE_MONTHLY)
+    if (keep == PRICE_REMOVE_KEEP_MONTHLY)
         recurrenceSet (r, 1, PERIOD_LAST_WEEKDAY, &day_of_week_date, WEEKEND_ADJ_NONE); // Month set begin on last friday of month
     else
         recurrenceSet (r, 1, PERIOD_WEEK, &day_of_week_date, WEEKEND_ADJ_NONE); // Week set to begin on Friday of week
@@ -1416,8 +1417,8 @@ gnc_pricedb_remove_get_keep_dates (GHashTable* hash_dates, GDate period_begin, G
 gboolean
 gnc_pricedb_remove_old_prices (GNCPriceDB *db, GList *comm_list,
                               Timespec first, Timespec cutoff,
-                              PriceRemoveOptions source,
-                              PriceRemoveOptions leave)
+                              PriceRemoveSourceOptions source,
+                              PriceRemoveKeepOptions keep)
 {
     GHashTable *hash_dates = NULL;
     remove_info data;
@@ -1430,13 +1431,13 @@ gnc_pricedb_remove_old_prices (GNCPriceDB *db, GList *comm_list,
     data.delete_user = FALSE;
     data.delete_all = FALSE;
 
-    ENTER("Remove Prices for Source %d, leaving %d", source, leave);
+    ENTER("Remove Prices for Source %d, keeping %d", source, keep);
 
     // setup the options
-    if (source == PRICE_REMOVE_ALL)
+    if (source == PRICE_REMOVE_SOURCE_ALL)
         data.delete_all = TRUE;
 
-    if (source == PRICE_REMOVE_USER)
+    if (source == PRICE_REMOVE_SOURCE_USER)
         data.delete_user = TRUE;
 
     // Walk the list of commodities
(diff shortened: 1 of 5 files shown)


src/engine/gnc-pricedb.c
```diff
@@ -1345,68 +1356,147 @@ pricedb_remove_foreach_pricelist (gpointer key,
...
+        LEAVE("Empty price list");
         return FALSE;
+    }
+    DEBUG("Number of Prices in list is %d", g_slist_length (data.list));
+
+    if (leave != PRICE_REMOVE_DEFAULT)
+    {
+        GDate period_begin = timespec_to_gdate (first);
+        GDate period_end = timespec_to_gdate (cutoff);
+        hash_dates = g_hash_table_new_full (g_str_hash, g_str_equal, free_data, NULL);
```
gjanssens: In addition to the comment made earlier about the value of the hash table, I think it would be more efficient to make the key a guint32 and use the result of g_date_get_julian as keys.
That will
* avoid the need for a free function
* allow direct insertions instead of constructing a date string to insert

src/engine/gnc-pricedb.c
```diff
@@ -1345,68 +1356,147 @@ pricedb_remove_foreach_pricelist (gpointer key,
...
+    r = g_new (Recurrence, 1);
+
+    DEBUG("Period Begin date is %d/%d/%d and End date is %d/%d/%d", g_date_get_day (&period_begin),
+           g_date_get_month (&period_begin), g_date_get_year (&period_begin), g_date_get_day (&period_end),
+           g_date_get_month (&period_end), g_date_get_year (&period_end));
+
+    g_date_set_day (&day_of_week_date, 1); // set date to 1st of month
+
+    while (g_date_get_weekday (&day_of_week_date) != selected_day_of_week)
+        g_date_subtract_days (&day_of_week_date, 1); // go back till we find last Friday of month
```
gjanssens: This is confusing me: your comment says you will go back till the last Friday of month. However this is at best the last Friday of the **previous** month or it will be the first day of **this** month in case it happens to be a Friday.
Is this what you intend ?

src/engine/gnc-pricedb.c
```diff
@@ -1345,68 +1356,147 @@ pricedb_remove_foreach_pricelist (gpointer key,
...
 
 gboolean
-gnc_pricedb_remove_old_prices(GNCPriceDB *db,
-                              Timespec cutoff,
-                              gboolean delete_user,
-                              gboolean delete_last)
+gnc_pricedb_remove_old_prices (GNCPriceDB *db, GList *comm_list,
+                              Timespec first, Timespec cutoff,
+                              PriceRemoveOptions source,
+                              PriceRemoveOptions leave)
```
gjanssens: It's confusing that source and leave both use the same enum PriceRemoveOptions. This is also a potential source of nasty bugs if someone after you makes changes to this code and doesn't pay sufficient attention to which are the right combinations of options.

src/engine/gnc-pricedb.h
```diff
@@ -372,17 +372,29 @@ gboolean     gnc_pricedb_add_price(GNCPriceDB *db, GNCPrice *p);
...
 
+typedef enum
+{
+    PRICE_REMOVE_DEFAULT, // default option value
+    PRICE_REMOVE_USER,    // only user prices
+    PRICE_REMOVE_ALL,     // all prices
+    PRICE_REMOVE_WEEKLY,  // leave one every week on a friday
+    PRICE_REMOVE_MONTHLY, // leave one every last friday of month
+    PRICE_REMOVE_SCALED,  // leave one every week then one a month
+} PriceRemoveOptions;
```
gjanssens: Like a mentioned in my other comment, it's confusing to mix different option sets in one enum. Please separate this into two, one for the options to indicate which sources to modify and one for the options to specify which dates to keep.

An example of the confusion: when used in the context of sources to clean up PRICE_REMOVE_DEFAULT means not user source and not all sources. So that would be fq source. However in the context of the "leave" option PRICE_REMOVE_DEFAULT means remove all dates (within the selected sources that is). Those are very different meanings.

src/engine/gnc-pricedb.c
gjanssens: 'leave' means "dates to keep/retain" ? Perhaps that's a better name for the variable ? I'm associating 'leave" with something that goes away apparently while here it seems to mean something to leave in place.

Bob-IT:
Thank you for looking at this, hopefully I have made the required changes, one day I may propose some changes that do not require subsequent tweaks. I did try to rebase this but could not quiet get it to work for me.

gjanssens:
@Bob-IT Thanks for the follow up commits. They do address my suggestions nicely.

These suggestions came from source code review. I have now built and run gnucash with your changes and I have a few more issues with the GUI that I couldn't see in code review.

* "Remove Old Prices" should be the dialog title instead of a label in the dialog body.

* To me the interface for selecting sources to remove prices from is fairly confusing. How many sources are there ? In my code review I found two: FQ based prices and user prices. If these are the only two, why not change the two checkboxes to "Fetched online" and "Entered manually" or something similar ? Checking both would then become equivalent to "All". The advantage would be the user clearly sees which prices sources are considered. The current situation with "Include manually entered prices" and "All Prices" on the other hand leaves the user guessing at which prices sources are considered when neither option is enabled.
If there are more than 2 price sources that should be considered, you can add more check boxes or try to come up with a scheme to group them (like manually entered vs computer generated, whatever). Changing the meanings of the buttons will clearly also impact the non-gui code.

I realize that with my suggestion (naming the price sources directly) the user could potentially make an invalid selection (if no check box is ticked). This can be checked and the Ok button should be dimmed in that situation. Alternatively this can also be avoided by using radio buttons instead. You'd need 3 in this case: "Fetched online", "Manually entered" and "All". This will obviously only work if there are only two price sources.

* In the same line of thought, the Ok  button should be dimmed if non commodities are selected.

* Next I think it would benefit the user to mention "Friday" in the Keeping line, rearranging it to something like this:
```
Keep a price on Friday (if available):
(o) None (o) One per week (o) One per month (o) Scaled
```
That sets the expectations straight before they make a choice. You can keep the caution (though the capital K is not needed in the sentence) to make it extra clear.

* And the final nit: I think the UI will be slightly easier on the eye if the "Delete..." and "Before Date..." are also left-aligned.

jralls:
There are 9 sources, enumerated in [gnc-pricedb.h](https://github.com/Gnucash/gnucash/blob/master/src/engine/gnc-pricedb.h#L165). Most are variants of 'generated by the program during edit of a transaction'.

fellen:
> Keep a price on Friday (if available):
> (o) None (o) One per week (o) One per month (o) Scaled

I would prefer:

Keep last of
* Week
* Month
* Accounting period (usually year)

gjanssens:
So we have 9 price sources.
In the current PR, all 9 of them will be considered if "All Prices" is selected. If "All Prices" is not selected the dialog will only consider prices of either type PRICE_SOURCE_FQ or type PRICE_SOURCE_USER_PRICE depending on the setting of the other check box and the other 7 will be ignored.

I believe listing 9 price types may be exagerated. How can we group them to reduce the selection list to a set that still makes sense ?
A first stab at it:
* Fetched online (for PRICE_SOURCE_FQ)
* Entered manually (for PRICE_SOURCE_USER_PRICE and possibly PRICE_SOURCE_EDIT_DLG)
* Generated automatically (for all the others)

I'm hesitating on PRICE_SOURCE_EDIT_DLG. I'm not sure in which context this is used as opposed to PRICE_SOURCE_USER_PRICE.
The idea behind the grouping is that "Generated automatically" refers to all prices that were created as a side effect, but which is not strictly necessary for gnucash to function properly. For example when a multi-currency transaction is created, a price is added automatically. However that price can safely be deleted as it can still be recalculated based on the amounts/values of the relevant splits. Same when posting an invoice when multiple currencies are involved. FQ and manually entered prices on the other hand are usually recorded to track price changes and can't be recalculated based on split data.

jralls:
The difference between USER_PRICE and EDIT_DLG is, IIRC, that in the first case a user added the price with the "add" button on the Price Database window and in the second the user changed an existing price.

While you're right that the "automatic" prices could be regenerated, the event that causes them to be created is a transaction commit, so they won't be. The motivation for those is so that there's always a "nearest in time" price for every multi-commodity transaction so that the Accounts page and reports are able to correctly report values.

gjanssens:
Ok. So what should we do with those "automatic" prices in the context of this PR? Should we allow a user to delete the historical ones or not ?
If not, should we hide them in the interface as the user should not have control over them at all  (I hear Wm complaining already to such a suggestion...) ?

jralls:
Derek's "enough rope" principle says that we should provide the ability, perhaps with a tooltip explaining the ramifications. I think the 3 checkboxes you suggest is right, and I also like Frank's suggestion to which I'd add 'quarterly'.

I don't think we should hide those prices from the user as they can be important to understanding why a report or the CoA assigns a particular value to an asset.

gjanssens:
Good arguments. Thanks.

@Bob-IT So there you go:
* please make it 3 checkboxes. When ticked all together it should act as if all prices are involved and otherwise only on the subset defined by the ticked boxes.
* please add a "One per quarter" and "One per accounting period" options
* add a tooltip to the "Generated automatically" check box to explain that removing these quotes may make it harder to follow which prices a report is using when reporting on actual transactions. Perhaps add an warning icon to this option (in the end) to attract the attention to this.

Thanks for polishing this up!

Bob-IT:
Can I ask for clarification of what Frank's and John's  extra options mean.
Is it last Friday of Quarter, last Friday of account period or is it delete every thing older than a week, month, quarter, account period based on the date selected ?

jralls:
I meant keep only the last price of each quarter; I think Frank meant the same for weeks, months, and accounting periods.

ISTM removing all prices older than some date would cause problems unless the date is older than the original purchase of the longest-held priced asset. Figuring that out from the database is possible of course but I don't think it would be easy.

fellen:
Yes, keep the last available record in the period. 
Except for weekly it will not necessary be a friday.

For the balance sheet I am interested in the 31.12. If it has no price check the days before until you find one with a price. Keep that. 

Depending on your region and business branch the accounting period might use another day than 31.12. So it should respect Edit->Preferences->Accounting Period->End date.

Bad example for "last friday": 
If 31.12 is a Thursday, the last available Friday would be the 18. because the 25. is christmas.

Bob-IT:
I have change the options to last of and removed the Friday ones as they were too similar.

jralls:
Bob, do the GUI bits need any changes for Gtk3?

Bob-IT:
Closing this as it is superseded by pull 171, I failed to rebase this one so it was easier to start from master and add the changes.

PR #144, opened 2017-04-21
Add constructor to create GncDate from string and predefined date format

gjanssens:
Primary use case is for parsing dates from external sources (importers).
Current implementation causes the test to abort in a segfault. I don't see any reason for that unfortunately. It seems to happen in boost:regex' memory housekeeping at the end of the test run. Running under gdb results in the following bt:
```
Program received signal SIGSEGV, Segmentation fault.
malloc_consolidate (av=av@entry=0x7ffff5b4bae0 <main_arena>) at malloc.c:4204
4204              nextsize = chunksize(nextchunk);
Missing separate debuginfos, use: dnf debuginfo-install gtest-1.7.0-8.fc25.x86_64
(gdb) bt
#0  0x00007ffff580bfbb in malloc_consolidate (av=av@entry=0x7ffff5b4bae0 <main_arena>) at malloc.c:4204
#1  0x00007ffff580db20 in _int_free (av=0x7ffff5b4bae0 <main_arena>, p=0x6f25a0, have_lock=0) at malloc.c:4110
#2  0x00007ffff58112bc in __GI___libc_free (mem=<optimized out>) at malloc.c:2982
#3  0x00007ffff724f6db in boost::re_detail_106000::mem_block_cache::~mem_block_cache() (this=0x7ffff74a9160 <boost::re_detail_106000::block_cache>, __in_chrg=<optimized out>) at boost/regex/v4/mem_block_cache.hpp:53
#4  0x00007ffff57c578a in __cxa_finalize (d=0x7ffff74a2d60) at cxa_finalize.c:56
#5  0x00007ffff71f04f3 in __do_global_dtors_aux () at /lib64/libboost_regex.so.1.60.0
#6  0x00007fffffffd840 in  ()
#7  0x00007ffff7de818a in _dl_fini () at dl-fini.c:235
```

### DIFF
--- a/src/import-export/csv-imp/gnc-tx-import.hpp
+++ b/src/import-export/csv-imp/gnc-tx-import.hpp
@@ -68,16 +68,23 @@ extern const gchar* currency_format_user[];
 extern const int num_date_formats;
 extern const gchar* date_format_user[];
 
-/** Tuple to hold
+/** An enum describing the columns found in a parse_line_t. Currently these are:
  *  - a tokenized line of input
  *  - an optional error string
  *  - a struct to hold user selected properties for a transaction
- *  - a struct to hold user selected properties for one or two splits in the above transaction */
-#define PL_INPUT    0
-#define PL_ERROR    1
-#define PL_PRETRANS 2
-#define PL_PRESPLIT 3
-#define PL_SKIP     4
+ *  - a struct to hold user selected properties for one or two splits in the above transaction
+ *  - a boolean to mark the line as skipped by error and/or user or not */
+enum parse_line_cols {
+    PL_INPUT,
+    PL_ERROR,
+    PL_PRETRANS,
+    PL_PRESPLIT,
+    PL_SKIP
+};
+
+/** Tuple to hold all internal state for one parsed line. The contents of each
+ * colummn is described by the parse_line_cols enum. This enum should be used
+ * with std::get to access the columns. */
 using parse_line_t = std::tuple<StrVec,
                                 std::string,
                                 std::shared_ptr<GncPreTrans>,
--- a/src/import-export/csv-imp/gnc-tx-import.hpp
+++ b/src/import-export/csv-imp/gnc-tx-import.hpp
@@ -73,6 +73,11 @@ extern const gchar* date_format_user[];
  *  - an optional error string
  *  - a struct to hold user selected properties for a transaction
  *  - a struct to hold user selected properties for one or two splits in the above transaction */
+#define PL_INPUT    0
+#define PL_ERROR    1
+#define PL_PRETRANS 2
+#define PL_PRESPLIT 3
+#define PL_SKIP     4
 using parse_line_t = std::tuple<StrVec,
                                 std::string,
                                 std::shared_ptr<GncPreTrans>,
--- a/src/libqof/qof/gnc-datetime.cpp
+++ b/src/libqof/qof/gnc-datetime.cpp
@@ -29,12 +29,18 @@ extern "C"
 }
 #include <boost/date_time/gregorian/gregorian.hpp>
 #include <boost/date_time/posix_time/posix_time.hpp>
+#include <boost/regex.hpp>
+#include <libintl.h>
+#include <map>
 #include <memory>
 #include <iostream>
 #include <sstream>
 #include <string>
+#include <vector>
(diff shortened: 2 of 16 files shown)


src/libqof/qof/gnc-datetime.hpp
```diff
@@ -153,10 +153,40 @@ class GncDateTime
...
+private:
+    const std::string m_name;
+};
+
 class GncDate
 {
-    public:/** Construct a GncDate representing the current day.
-        */
+    public:
+        /** Returns a vector with all the formats supported by the string constructor
```
jralls: Doesn't *return* a vector, it *is* a vector.

Since it's a class member variable it should be c_formats.

src/libqof/qof/gnc-datetime.cpp
```diff
@@ -65,15 +65,15 @@ static constexpr auto ticks_per_second = INT64_C(1000000);
...
  * to parse each from an external source
  * Note: while the format names are using a "-" as separator, the
  * regexes will accept any of "-/.' " and will also work for dates
  * without separators.
  */
-static std::map<const std::string, const char *> gnc_date_fmt_map = {
-    {
-        std::string(N_("y-m-d")),
+const std::vector<GncDateFormat> GncDate::formats = {
+    GncDateFormat {
```
jralls: I think you need to wrap the initializer list in parentheses i.e. ```GncDateFormat({...})```.

src/libqof/qof/gnc-datetime.hpp
```diff
@@ -153,10 +153,40 @@ class GncDateTime
...
+private:
+    const std::string m_re;
+
+    friend class GncDateImpl;
+};
+
+/** Functor to check if a GncDateFormat refers to a specific date format.
+ * Intended to query GncDate::formats with std::find_if.
+ */
+struct GncDateFormatPred
```
jralls: Since only GncDateImpl can use the predicate it's better to use a lambda than a functor as that keeps the logic local to the function using it.

src/libqof/qof/gnc-datetime.cpp
gjanssens: It's a better optimization indeed. That will make it use the copy constructor instead of copy assignment.

Other than that it doesn't seem to make much of a difference. The abort remains.

src/libqof/qof/gnc-datetime.hpp
gjanssens: Good point.

gjanssens: Yay, I just wrote my first lambda function :)
If we continue at this pace I'll be a c++ guru by the end of the month :D 
Unfortunately it still aborts during cleanup.

src/import-export/csv-imp/gnc-tx-import.hpp
```diff
@@ -73,6 +73,11 @@ extern const gchar* date_format_user[];
  *  - an optional error string
  *  - a struct to hold user selected properties for a transaction
  *  - a struct to hold user selected properties for one or two splits in the above transaction */
+#define PL_INPUT    0
+#define PL_ERROR    1
+#define PL_PRETRANS 2
+#define PL_PRESPLIT 3
+#define PL_SKIP     4
```
jralls: Can't this be an enum instead of a cpp macro?

src/libqof/qof/gnc-datetime.hpp
```diff
@@ -152,11 +153,28 @@ class GncDateTime
     std::unique_ptr<GncDateTimeImpl> m_impl;
 };
 
+class GncDateFormat
```
jralls: Needs Doxygen docs.

src/import-export/csv-imp/gnc-tx-import.hpp
gjanssens: Yes it can. When writing this part I was considering a scoped enum as that's preferred by Meyers. Such a scoped enum doesn't cast into an int type unless explicitly using static_cast. Got a brain failure at that point and used cpp macros instead of an ordinary enum...

Fixed.

src/libqof/qof/gnc-datetime.hpp
gjanssens: Done. Included a general doxygen comment for the GncDate class as well.

PR #145, opened 2017-04-23
Save button active on every load

Bob-IT:
The gnc_numeric numval was being set with a invalid numeric and hence was always different to the parsed amount.

### DIFF
--- a/src/app-utils/gnc-sx-instance-model.c
+++ b/src/app-utils/gnc-sx-instance-model.c
@@ -78,7 +78,7 @@ scrub_sx_split_numeric (Split* split, const char *debcred)
     const char *numeric = is_credit ?
         "sx-credit-numeric" : "sx-debit-numeric";
     char *formval;
-    gnc_numeric numval;
+    gnc_numeric *numval;
     GHashTable *parser_vars = g_hash_table_new (g_str_hash, g_str_equal);
     char *error_loc;
     gnc_numeric amount = gnc_numeric_zero ();
@@ -93,7 +93,7 @@ scrub_sx_split_numeric (Split* split, const char *debcred)
     if (!parse_result || g_hash_table_size (parser_vars) != 0)
         amount = gnc_numeric_zero ();
     g_hash_table_unref (parser_vars);
-    if (gnc_numeric_eq (amount, numval))
+    if (gnc_numeric_eq (amount, *numval))
         return FALSE;
     qof_instance_set (QOF_INSTANCE (split),
 		  numeric, &amount,


jralls:
Bob, can you explain your reasoning here? ```&numeric``` is passed (eventually) as a parameter to [gnc_numeric_boxed_copy_func](https://github.com/Gnucash/gnucash/blob/master/src/libqof/qof/gnc-numeric.cpp#L1211), as a gpointer (aka void*) that's cast to a gnc_numeric* not a gnc_numeric**.

Bob-IT:
Just to confirm my findings, with the original lines and this line added after the qof_instance_get...
g_print("formval %s, numval %s\n", formval, gnc_num_dbg_to_string (numval));

Which resulted in the following output...
formval 100.00, numval 37057072/140024340781676
formval (null), numval 37259408/140024340781676
formval (null), numval 37581840/140024340781676
formval 100.00, numval 37646208/140024340781676

With the changes and a change to the print statement *numval, the output is...
formval 100.00, numval 100/1
formval (null), numval 0/1
formval (null), numval 0/1
formval 100.00, numval 100/1

So I thought that was the problem unless this is just down to my setup?
Bob

Bob-IT:
I am having another look at this, with my changes I get a segfault when the numval values are missing.

jralls:
Sorry, I should have responded sooner to your test results. Clearly you're right that it's turning into a gnc_numeric**, I think as a consequence of having to "box" the struct. The actual bug is at https://github.com/Gnucash/gnucash/blob/master/src/libqof/qof/gnc-numeric.cpp#L1216 that should be
```
if (!(in_gnc_numeric && *in_gnc_numeric))
    return nullptr;
/* The returned pointer will be g_free()d so we must allocate with malloc, not operator new. */
auto newvalue = static_cast<gnc_numeric*>(g_malloc (sizeof (gnc_numeric)));
*newvalue =  *(static_cast<gnc_numeric*>(*in_gnc_numeric));
```
Fixing that will also require checking that all of the other places where we retrieve gnc_numerics from KVP to make sure that the right level of dereferencing is used.

Yes, of course it will segfault if numval == nullptr. You'd need to check that before dereferencing it:
```
if (!numval || gnc_numeric_eq(amount, *numval))
```

Bob-IT:
I think I have misunderstood your comment above. I changed the boxed copy function to ...
```
static gpointer
gnc_numeric_boxed_copy_func( gpointer in_gnc_numeric )
{
    if (!in_gnc_numeric)
        return nullptr;
    /* The returned pointer will be g_free()d so we must allocate with malloc, not operator new. */
    auto newvalue = static_cast<gnc_numeric*>(g_malloc (sizeof (gnc_numeric)));
    *newvalue =  *(static_cast<gnc_numeric*>(in_gnc_numeric));

    return newvalue;
}
```
note the changes to get it to compile, and still required the changes I made. I thought I would have to revert my changes and then go hunting for other places like you mentioned.

jralls:
Here's a more understandable version that actually compiles:
```
    auto in_gnc_numeric = static_cast<gnc_numeric**>(in_ptr);
    if (!(in_gnc_numeric && *in_gnc_numeric))
	return nullptr;

    /* newvalue will be passed to g_free so we must allocate with g_malloc. */
    auto newvalue = static_cast<gnc_numeric*>(g_malloc (sizeof (gnc_numeric)));
    *newvalue = **in_gnc_numeric;

    return newvalue;
```
The point was that the incoming is a gnc_numeric**. It doesn't work because it's supposed to be a GBoxedCopyFunc and those need to be symmetrical and besides, the real issue is that the return value is a gpointer. GValue isn't able to use direct copy semantics on anything bigger than sizeof(int*).

So your solution is correct, except that it also needs to ```g_free (numval)```... not that you'd have known that.

And I need to go through and find every *other* instance where we're handling a GValue with a G_TYPE_NUMERIC in it and make sure that it's getting g_freed.

jralls:
Eh, make that ```g_boxed_free (GNC_TYPE_NUMERIC, numval);```.

gjanssens:
Another can of worms... Yay :(

gjanssens:
@jralls, if I read the [G_TYPE_STRV example](https://developer.gnome.org/gobject/unstable/gobject-Boxed-Types.html#G-TYPE-STRV:CAPS) correctly, the gnc_numeric* returned by qof_object_get should be freed with g_free, not g_boxed_free.

As far as I can tell g_object_get (and by extension qof_instance_get) will return the type directly, not a boxed version of it so there's no box to free. Or is there something I'm missing here ?

gjanssens:
I decided to have a look at this today. I replaced gnc_numeric_boxed_copy_func with @jralls' code. It compiles fine, but I right after that 8 tests segfault. To investigate I've picked test-job as that's small and only has one gnc_numeric type property, "job-rate". This property is stored in kvp, so it's slightly different from the above property in gnc_sx_instance_model. My finding is that in this case the value passed to gnc_numeric_boxed_copy_func is a (gnc_numeric*) and not a (gnc_numeric**). I've copied my debug session to [this gist](https://gist.github.com/gjanssens/be5b59d8ac644f23ada314df581d36c7) to illustrate.

So because I got confused I decided to revert all changes and have the debugger stop in gnc_numeric_boxed_copy_func when triggered from the qof_instance_get call in scrub_sx_split_numeric. And lo and behold, the function parameter in_gnc_numeric is a (gnc_numeric*) as well instead of a (gnc_numeric**). I've added this gdb run to the gist as well.
So much of this conversation has been based on a false assumption or I misunderstood what @jralls was trying to explain.

Next I've adjusted @jralls' version of gnc_numeric_boxed_copy_func to assume a (gnc_numeric*) as input (but keeping a reduced nullptr test to avoid dereferencing a null input), together with @Bob-IT's change to the scrub function and adding a g_free to avoid leaking the returned (gnc_numeric*).

As far as I'm concerned that solves the original issue of the file always being marked dirty. We still have to go through all of the uses of the boxed gnc_numeric's and check whether a g_free is needed to be added (after retrieving the boxed value) or a g_boxed_free after setting the boxed value. The latter is unlikely as the most common way we create boxed gnc_numerics is via &value with value a simple gnc_numeric (stack based). And by extension we should verify all other boxed types we use for the proper g_(boxed_)free calls upon setting or getting.

gjanssens:
That turned out to be a fairly small change for gnc_numeric based (boxed) GValues. In most cases the values were set based on the &value construct as I suggested, so no need to free the original values after creating the boxed ones. There were a few places where the retrieved values had to be freed, but in most cases g_value_get_boxed was called  which returns a pointer without transferring ownership. So again there was nothing to free for in most cases.
Lastly a slightly trickier construct:
```
    g_value_init (&v, GNC_TYPE_NUMERIC);
    g_value_set_boxed (&v, &rate);
    qof_instance_set_kvp (QOF_INSTANCE (job), GNC_JOB_RATE, &v);
```
As far as I understand this the g_value_set_boxed will store a copy of &rate. qof_instance_set_kvp will use g_value_get_boxed to access this copy (without transferring ownership once more) to set the kvp value internally. This leaves us with the GValue v which is no longer needed after the call to qof_instance_set_kvp. Here I assume a GValue will call g_boxed_free(GNC_NUMERIC_TYPE, <internal-copy-of-rate>) when it goes out of scope to free the internal copy of &rate. A simple run in gdb testing with job-rate seems to confirm this.

jralls:
Thanks for cleaning out the can!

PR #146, opened 2017-05-06
Modify KRW fraction of ISO4217 currency settings

hkjinlee:
Chon(=Jeon), the smaller part of KRW, is no longer used in Korea.
Leaving it as it is causes annoying .00 attached after every value.

(FYI, JPY is the same, but it is already marked as having 1 smallest fraction, not 100. 
Someone must have updated it before)

Reference:
- https://en.wikipedia.org/wiki/ISO_4217 :
Number of digits after decimal separator is 0 for KRW
- https://en.wikipedia.org/wiki/South_Korean_won :
“The jeon is no longer used for everyday transactions”

### DIFF
--- a/src/engine/iso-4217-currencies.xml
+++ b/src/engine/iso-4217-currencies.xml
@@ -1452,8 +1452,8 @@
   partname="chon"
   namespace="ISO4217"
   exchange-code="410"
-  parts-per-unit="1"
-  smallest-fraction="100"
+  parts-per-unit="100"
+  smallest-fraction="1"
   local-symbol="₩"
 />
 <!-- "KWD" - "Kuwaiti Dinar"


fellen:
Thanks to bring it up again!

There was some discussion in https://bugzilla.gnome.org/show_bug.cgi?id=666536  - smallest fraction of KRW is invalid:

https://en.wikipedia.org/wiki/ISO_4217 is often inaccurate. They are only interested in cash, not accounting.
If I understand it right, http://ko.wikipedia.org/wiki/%EB%8C%80%ED%95%9C%EB%AF%BC%EA%B5%AD_%EC%9B%90 shows 전: 1/100 and notes “Jeon is used only for accounting” but that is what we do here.

The current problem is Bug 408444 - "Smallest fraction" disregarded: You can set the smallest currency unit of some accounts like "Cash" to something greater than the default, but the other way has issues.

So I would suggest to continue the discussion in above bugzilla entry.

hkjinlee:
Didn't know this was already discussed before. Thank you for letting me know.
I will try to find official accounting rules, and post it on bugzilla thread.

PR #147, opened 2017-05-13
Bug 776396 - Foreign currency reports, amounts are not aligned correctly 

bclaesen:
https://bugzilla.gnome.org/show_bug.cgi?id=776396

Fixes the bug with the css change proposed in the bugzilla report.

### DIFF
--- a/src/report/report-system/html-acct-table.scm
+++ b/src/report/report-system/html-acct-table.scm
@@ -1163,7 +1163,7 @@
 	  )
 	 )))
     ;;below line as fix to Bug 776396 - make sure the commidity aligns the same as the rest of the currency (on the right)
-    (gnc:html-table-set-style! table "table" 'attribute(list "style" "width:100%; max-width:20em"))
+    (gnc:html-table-set-style! table "table" 'attribute(list "style" "width:100%; max-width:20em") 'attribute (list "cellpadding" "0"))
     ;;end fix
     table)
   )


bclaesen:
Hi Chris

That makes sense (although it looks a bit weird when testing), I can see how this fixes the alignment. I am new to this, so trying to find how I can add a commit to this pull request. Feel free to add it yourself (you seem to have more experience with github), what you wrote in the comment will work.

gjanssens:
Bert, thank you for your PR.

The fix itself is perfectly ok. Still a few remarks:
- You have added a lengthy comment referring to the bug being fixed in the first patch surrounding the one-line fix. Please don't do that. Instead it's much more useful to add this information in the commit message. A short comment in code is ok if it explains non-obvious considerations. I don't think your code really needs one in this case. It's fairly obvious to understand you are adding style information to the table element.

- Next your PR is against our master branch. However bugfixes like this should go into the maint branch.

- While doing so I also have a suggestion regarding the PR structure: it appears you started from a feature branch (bclaesen-acct-table) which you then merged in your own github gnucash repository's master branch via a PR and then created a new PR for gnucash master. This would make our git history needlessly confusing after the final merge. Better would be to make a PR directly from your feature branch to our master (well, in this case rather maint) branch, skipping the intermediate merge in your own repository. So I'd like to see this cleaned up first.

- Lastly for the convenience of our release manager it's useful if the commit message starts with the bug number and subject in the first line and then (after one empty line) an optional message body with more details about the commit, so for example:

````
Bug 776396 - Foreign currency reports, amounts are not aligned correctly

This commit will improve the styling of the table element used
for foreign currencies in the Balance sheet report so they
end up aligned with the other currencies.
````
This is just an example. Feel free to write as much or as little detail in the message body as you see fit.

How comfortable are you with git to tweak your commits to fit these requests into a PR against maint ? If you need some guidance for this, feel free to ask.

bclaesen:
Hi Geert

Real newbie at git and github at the moment but I'll try to get it right later this weekend. Always good to learn so the next one will be more right.

That's what I get for reading the gnucash wiki early in the morning, it's all there in retrospect :) 
Bert

bclaesen:
Ok, I hope I have all your feedback included. I really liked your example description, so used it. Think the only thing I should change in the future is add a feature branch on my end and then PR against that.
https://github.com/Gnucash/gnucash/pull/148

Is it OK to close this one?

gjanssens:
Lovely!
I was still in the process of writing some instructions to help you further, but you managed it all by yourself. This one can be closed now indeed. I'll follow up on the other PR.

PR #148, opened 2017-05-13
Bug 776396 - Foreign currency reports, amounts are not aligned correctly

bclaesen:
Bug 776396 - Foreign currency reports, amounts are not aligned correctly

This commit will improve the styling of the table element used
for foreign currencies in the Balance sheet report so they
end up aligned with the other currencies.

Follow up on feedback on:

- https://github.com/Gnucash/gnucash/pull/147
- https://bugzilla.gnome.org/show_bug.cgi?id=776396

### DIFF
--- a/src/report/report-system/html-acct-table.scm
+++ b/src/report/report-system/html-acct-table.scm
@@ -1162,6 +1162,7 @@
 	   )
 	  )
 	 )))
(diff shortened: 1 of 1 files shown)


gjanssens:
Thanks a lot and congratulations on your first inclusion in the gnucash code base! Well done :)

PR #149, opened 2017-05-16
Fix coredump in Python when using "get_namespaces"

TheBiggerGuy:
The Python API incorectly had GncCommodityTable.get_namespaces() defined
as a list of GncCommodityNamespace over the correct list of String. This
fixes the issue and adds a test for GncCommodityTable.get_namespaces()
and GncCommodityTable.get_namespaces_list().

Note: This is not a direct fix as I could not get SWIG to detect/convert
"GList *" to a "gchar *" to a Python str list.

### DIFF
--- a/src/optional/python-bindings/gnucash_core.py
+++ b/src/optional/python-bindings/gnucash_core.py
@@ -367,7 +367,8 @@ class GncCommodityTable(GnuCashCoreClass):
     which includes most of the world's currencies.
     """
 
-    pass
+    def _get_namespaces_py(self):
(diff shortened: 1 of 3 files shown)


TheBiggerGuy:
#150 should be merged before this so the Python tests are run on Travis

gjanssens:
Committed after a rebase on top or your PR150.

Just for future reference, you could also have added the single commit of PR 150 into this PR (and reordering the commits in this PR as needed). That would have avoided my rebase. While I'm fine with doing the manual rebase it sometimes confuses the PR submitters.

Regardless, thanks for your contributions!

PR #150, opened 2017-05-16
Enable python testing in travis

TheBiggerGuy:
This should be merged before #149

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,4 +18,4 @@ install:
   - sudo apt-get install -qq libboost-all-dev
   - sudo apt-get --reinstall install -qq language-pack-en language-pack-fr
   - git clone https://github.com/google/googletest -b release-1.8.0 ~/gtest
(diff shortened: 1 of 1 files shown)


jralls:
PRs 149 and 150 are identical, and in both the third commit reverts the first.

The second commit, which only passes --enable-python in the config arguments, seems unlikely to work without at least installing python-dev and is in any case pointless once the first commit is reverted by the third.

Perhaps you don't realize that pull requests apply to a branch rather than to a commit...

TheBiggerGuy:
Sorry for the confusion I didn't realise I branched #150 of #149 instead of master.

#150 (this request) has the correct one line diff to enable Python testing in Travis. It is passing in Travis and the logs correctly show "runTests.py" is been run.

#149 Does not have #150 and was never the same.

jralls:
Please rebase this branch so that the it has only 2a6ce28.

TheBiggerGuy:
Sorry I did not know you could force push after a pull request was opened. I have now fixed the pull request to be only the single commit.

jralls:
Perfect, thanks, I'll merge them both Thursday unless Geert beats me to it.

gjanssens:
Thanks!

PR #151, opened 2017-05-17
Fixes, tests and clean up for GncNumeric in Python

TheBiggerGuy:
This is mainly a fix for incorrect type conversion in GncNumeric (44cbb5afb53243552cc4a12606b42fac468b969b) but also includes extra tests and a small clean up of the string representation.

### DIFF
--- a/src/app-utils/Makefile.am
+++ b/src/app-utils/Makefile.am
@@ -118,7 +118,7 @@ endif
 endif
 swig-app-utils-python.c: app-utils.i ${top_srcdir}/src/base-typemaps.i
 	$(SWIG) -python -Wall -Werror $(SWIG_ARGS) \
-	-I${top_srcdir}/src -o $@ $<
+	-I${GLIB_CFLAGS} -I${top_srcdir}/src -o $@ $<
 endif
 
 if WITH_PYTHON
--- a/src/base-typemaps.i
+++ b/src/base-typemaps.i
@@ -169,38 +169,15 @@ typedef char gchar;
 #elif defined(SWIGPYTHON) /* Typemaps for Python */
 
 %import "glib.h"
+%include <stdint.i>
 
-%typemap(in) gint8, gint16, gint32, gint64, gshort, glong {
-    $1 = ($1_type)PyInt_AsLong($input);
-}
-
-%typemap(out) gint8, gint16, gint32, gint64, gshort, glong {
-    $result = PyInt_FromLong($1);
-}
-
-%typemap(in) guint8, guint16, guint32, guint64, gushort, gulong {
-    $1 = ($1_type)PyLong_AsUnsignedLong($input);
-}
-
-%typemap(out) guint8, guint16, guint32, guint64, gushort, gulong {
-    $result = PyLong_FromUnsignedLong($1);
-}
-
-%typemap(in) gdouble {
-    $1 = ($1_type)PyFloat_AsDouble($input);
-}
-
-%typemap(out) gdouble {
-    $result = PyFloat_FromDouble($1);
-}
-
-%typemap(in) gchar * {
-    $1 = ($1_ltype)PyString_AsString($input);
-}
-
-%typemap(out) gchar * {
-    $result = PyString_FromString($1);
-}
+%apply int { gint };
+%apply long { glong };
+%apply int64_t { gint64 };
+%apply unsigned long { gulong };
+%apply uint64_t { guint64 };
+%apply double { gdouble };
+%apply char* { gchar* };
 
 %typemap(in) gboolean {
     if ($input == Py_True)
--- a/src/base-typemaps.i
+++ b/src/base-typemaps.i
@@ -22,18 +22,16 @@
  *                                                                  *
 \********************************************************************/
 
+typedef void * gpointer; // Not sure why SWIG doesn't figure this out.
+%typemap(newfree) gchar * "g_free($1);"
 
-/* Not sure why SWIG doesn't figure this out. */
+#if defined(SWIGGUILE)
+
 typedef int gint;
 typedef gint64 time64;
 typedef unsigned int guint;
 typedef double gdouble;
 typedef float gfloat;
-typedef void * gpointer;
-
-%typemap(newfree) gchar * "g_free($1);"
-
-#if defined(SWIGGUILE)
 typedef char gchar;
 
 %typemap (out) char * {
@@ -169,6 +167,9 @@ typedef char gchar;
 }
 %enddef
 #elif defined(SWIGPYTHON) /* Typemaps for Python */
+
+%import "glib.h"
+
 %typemap(in) gint8, gint16, gint32, gint64, gshort, glong {
     $1 = ($1_type)PyInt_AsLong($input);
 }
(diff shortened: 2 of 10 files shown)


TheBiggerGuy:
The main question I have for this fix is should we use the Python code to convert str/float to (int, int) or use the C/C++ [double_to_gnc_numeric](https://github.com/Gnucash/gnucash/blob/master/src/libqof/qof/gnc-numeric.h#L295) and [string_to_gnc_numeric](https://github.com/Gnucash/gnucash/blob/master/src/libqof/qof/gnc-numeric.h#L301)?

This also goes for the __str__ using [gnc_numeric_to_string](https://github.com/Gnucash/gnucash/blob/master/src/libqof/qof/gnc-numeric.h#L334)? (Note I have tried this but l can not get SWIG to work. It looks like the Python-SWIG code is missing str/gchar* converters)

gjanssens:
We will be seeing the missing type conversions in swig more and more as we progress with converting the code from C to C++. I think the proper solution is to add these and have the python bindings use the gnucash api instead of doing the conversions directly in Python.
Failing to do so will open the gates to a flood of subtle behavioural differences between the gnucash C++ code and the python bindings. In this particular case I'm thinking of rounding or reducing of numbers which python and our C++ code may use different rules for.

TheBiggerGuy:
OK I have updated to code to use the native C/C++ str/double parsers and to_string.
Outstanding issues:
* What should the default __denom__ be for double (I picked GNC_DENOM_AUTO)
* What should the default __how__ be for double (I picked GNC_HOW_DENOM_FIXED | GNC_HOW_RND_NEVER)
* Correct tests to exercise the above two decision. I do not understand it enough to write them.

gjanssens:
Thanks for the update.

As to your question on what denom and how should be, I need to answer in 3 parts.
- The code you have chosen to use is the legacy C interface. It will go away at some point as it's being replaced by a cleaned up C++ interface. For GncNumeric this exists already. You can find it in gnc-numeric.hpp. There are no swig bindings yet for this interface so I'm fine with your choice of the C interface for now (perhaps you're interested in helping out getting our C++ interface wrapped in swig ?)
- for the C interface, the function leaves it up to the user to define "denom" and "how" values. So it would make sense if the python api also allows this to stay as close as possible to the C api. You could implement this by adding another branch in your if clause that tests for 3 parameters, which are then assumed to be double, "denom" and "how", which you can pass to double_to_gnc_numeric.
- The C++ interface on the other hand changes the assumptions. We've decided to defer rounding to where it's needed (in the financial objects like transactions and splits), and hence the (C++) code hence to create a GncNumeric from a double will try to keep a maximum of significant digits. So for example 1.2345678 will be created as 12345678/10000000. It will round though if there are too many significant digits. So the case where a python gncnumeric is to be created from a single double parameter should in theory round as well (where you chose GNC_HOW_RND_NEVER). In practise however what you chose as defaults will result in the same behaviour, because the rounding in this scenario (too many sigfigs) happens **before** the passed in rounding parameters are considered. By the time the code will evaluate GNC_HOW_RND_NEVER the double has already been rounded to fit the internal temporary GncNumeric. Hence that temporary one won't need further rounding so GNC_HOW_RND_NEVER is just fine.

A short summary:
- the defaults you have chosen for double are ok
- please add one more option to generate a python gncnumeric using a double, denom and how value
- alternatively if you feel up to it, investigate how to wrap the C++ interface in swig and use that instead

Thanks for your work so far!

TheBiggerGuy:
I've updated the code and now have two issues:
* _GNC_HOW_DENOM_SIGFIGS_ is broken
* Still more tests around the float/double code to ensure there is nothing broken there

Also some better docs would be nice but that definitely is not my strong point.

TheBiggerGuy:
I have cleaned up the c from/to Python type conversions so I am happy there will be more under/overflows correctly handled. I am still trying to work out the scope for Python floats to C double overflows. SWIG seems to do less error checking with float/double than with int/long.

jralls:
It would be *much* better to use Standard C types than GLib ones, and there's no good reason to use gint, gfloat, or gchar instead of the real names int, float, and char. For those cases where specifying the number of bits is necessary use the Standard C types int32_t, int64_t, etc.

There's no overflow checking because C doubles and Python reals are the same thing, see https://docs.python.org/3/reference/datamodel.html#objects-values-and-types. I think that GnuCash's Scheme code uses only the "inexact real" representation in Guile and that's also just a wrapper around a C double.

TheBiggerGuy:
The SWIG wrapper needs to use the same types as the GNC code uses (at least during calls to GNC). The patch just maps the GLib types onto the standard C types so the default SWIG code can work. I think this is better than custom type wrappers as the SWIG team are more likely to get the conversion correct. Or am I getting the wrong end of the stick on your comment @jralls ?? (I am now on IRC as TheBiggerGuy)

Also I am looking into the Travis error. It works on my PC so need to find out whats the difference.

TheBiggerGuy:
I have fixed the main issues by making all SWIG changes Python only. However I am unable to reproduce the Travis test failure on my device.
Is anyone able to test this branch and post the test failure?

TheBiggerGuy:
Updated to fix the CMake build. I'm still looking into the Travis failure.

jralls:
The errors in runTests.py are:
```
test_memo (test_split.TestSplit) ... ok
test_transaction (test_split.TestSplit) ... swig/python detected a memory leak of type 'guint *', no destructor found.
swig/python detected a memory leak of type 'guint *', no destructor found.
ok
test_clone (test_transaction.TestTransaction) ... swig/python detected a memory leak of type 'guint *', no destructor found.
ok
test_countsplits (test_transaction.TestTransaction) ... ok
```
```
test_iso_currency (test_commodity.TestCommodity) ... ok
test_namespaces (test_commodity.TestCommodityNamespace) ... <class 'gnucash.gnucash_core.GncCommodityTable'>
['AMEX', 'NYSE', 'NASDAQ', 'EUREX', 'FUND', 'template', 'CURRENCY']
ok
test_namespaces_list (test_commodity.TestCommodityNamespace) ... ok
```
```
test_to_str (test_numeric.TestGncNumeric) ... ok

======================================================================
ERROR: test_from_long (test_numeric.TestGncNumeric)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/john/gnucash-git/src/optional/python-bindings/tests/test_numeric.py", line 32, in test_from_long
    GncNumeric((2**64)+1)
  File "/home/john/gnucash-build/src/optional/python-bindings/gnucash/gnucash_core.py", line 288, in __init__
    kargs['instance'] = GncNumeric.__args_to_instance(args)
  File "/home/john/gnucash-build/src/optional/python-bindings/gnucash/gnucash_core.py", line 298, in __args_to_instance
    return gnc_numeric_create(arg ,1)
  File "/home/john/gnucash-build/src/optional/python-bindings/gnucash/gnucash_core_c.py", line 2321, in gnc_numeric_create
    return _gnucash_core_c.gnc_numeric_create(*args)
TypeError: in method 'gnc_numeric_create', argument 1 of type 'gint64'
```

jralls:
```-I${GLIB_CFLAGS}``` expands to ```-I-pthreads -I/usr/include/glib-2.0  -I/usr/lib/x86_64-linux-gnu/glib-2.0/include``` on my Ubuntu 14.04 VM. The nonsensical ```-I-pthreads``` for some reason stops swig's include processing from looking in the cwd unless one adds -I${srcdir} to the args as well. Leaving off the leading "-I" passes the naked ```-pthreads``` argument which swig doesn't understand.

One option would be to create a GLIB_INCLUDE_DIRS in configure with ```pkg-config --cflags-only-I glib-2.0``` and to use that, but frankly in light of the leaks and test errors perhaps the custom typedefs and conversions are better left in place.

jralls:
This now needs a rebase to the new file layout (replace src with libgnucash), and if you want it in 2.8 it needs to be a PR on the unstable branch.

Or have you abandoned it?

jralls:
Merged after a rebase and some fixups. Thanks!

PR #152, opened 2017-06-03
Implement 'dist', 'distcheck' and 'uninstall' targets for CMake

codesmythe:
This PR implements the 'dist' , 'distcheck' and 'uninstall' targets for CMake.

The tarballs created by 'dist' (src/cmake_modules/MakeDist.cmake) can be built with either CMake or Autotools.

'Distcheck' (src/cmake_modules/MakeDistCheck.cmake) will untar a tarball, run a CMake configure on the tarball, and then build. After that it will to an 'install' , 'check' and 'dist'. Assuming all that passes, the autotools distcheck is then run.

Tested in  Ubuntu 16.04, Ubuntu 14.04, Debian Jessie, Centos 7, Fedora 25 and MacOS. Tested that the build still works on Windows. 

Fedora 25 and Ubuntu 16.04 use automake 1.15, which breaks the autotools distcheck by changing the build directory.  For Ubuntu 16.04, automake 1.11 is an easy apt install away. For Fedora, you can use AUTOTOOLS_IN_DIST=OFF to skip the autotools part. The resulting tarball with then only build with CMake.

Once this is merged into maint, I can prepare a PR for master.


jralls:
Rob, Thanks!

But was there no way to get it to build and check the tarball's contents without resorting to Automake?

jralls:
Thanks for the update. Unfortunately I've been consumed with getting GnuCash to build under mingw64, which we need to move master to Gtk3, and now I need to work on the 2.6.17 release. I won't be able to do anything with this until July, sorry.

jralls:
Merged after a rebase a couple of weeks ago, thanks.

PR #153, opened 2017-06-04
Bug 776396 Further Fix

christopherlam:
This commit will further improve the styling of the table element used for foreign 
currencies in reports with stocks or foreign currencies. Adds non-breaking space 
between foreign currency/stock and report currency.


gjanssens:
Thanks :)

PR #154, opened 2017-06-25
Fix import-export compile on windows

kcwu:
libgncmod-test-engine.la uses some functions defined in libtest-core.la, so libtest-core.la should be put after libgncmod-test-engine.la


gjanssens:
Hmm, I merged the PR into master, but in retrospect I think it should have been merged into maint. As we're in a release weekend I'll defer the second merge (to maint) until the release is out.

gjanssens:
I have now cherry-picked this commit to maint as well. Thanks!

PR #155, opened 2017-07-03
Fixing and improving hello-world.scm 

christopherlam:
1) I have fixed the "contact gnucash-devel@gnucash.org" which was previously buggy, overwriting the previous line.
2) I have added an eval-string option which will be useful to learn scheme in gnucash. A mini-REPL of sort. With this I will be finally be productive in scheme (after years of hacking at it!)


jralls:
2 changes for different purposes should be 2 commits, even if they're only a few lines apart.

PR #156, opened 2017-07-03
Fixing hello-world.scm

christopherlam:
Hello-world improved by fixing nesting in contact details, and added new eval-string option to ease scheme programming for reports.

### DIFF
--- a/src/report/utility-reports/hello-world.scm
+++ b/src/report/utility-reports/hello-world.scm
@@ -354,9 +354,9 @@ or extending existing reports.")))
 new, totally cool report, consult the mailing list %s.")
           (gnc:html-markup-anchor 
            "mailto:gnucash-devel@gnucash.org"
-           (gnc:html-markup-tt "gnucash-devel@gnucash.org")))
-         (_ "For details on subscribing to that list, see &lt;http://www.gnucash.org/&gt;.")
-         (_ "You can learn more about writing scheme at &lt;http://www.scheme.com/tspl2d/&gt;."))
+           (gnc:html-markup-tt "gnucash-devel@gnucash.org"))
+          (_ "For details on subscribing to that list, see &lt;http://www.gnucash.org/&gt;. ")
+          (_ "You can learn more about writing scheme at &lt;http://www.scheme.com/tspl2d/&gt;.")))
 
         (gnc:html-markup-p
          (gnc:html-markup/format
--- a/src/report/utility-reports/hello-world.scm
+++ b/src/report/utility-reports/hello-world.scm
@@ -105,7 +105,7 @@
     (add-option
      (gnc:make-string-option
       (N_ "Hello, World!") (N_ "Eval Option")
-      "cc" (N_ "This is an eval option.") (N_ "(+ 2 3)")))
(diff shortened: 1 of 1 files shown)


src/report/utility-reports/hello-world.scm
```diff
@@ -96,7 +96,17 @@
...
+
+    ;; This is an eval option. Users can type a scheme expression.
+    ;; The default value is "(+ 2 3)". The scheme expression is 
+    ;; evaluated left to right, and the last value returned is
+    ;; to be printed into the report. This will be useful to 
+    ;; learn hacking scheme in gnucash.
+    (add-option
+     (gnc:make-string-option
+      (N_ "Hello, World!") (N_ "Eval Option")
+      "cc" (N_ "This is an eval option.") (N_ "(+ 2 3)")))
```
gjanssens: As the default value is not a string you expect to translate there is no need to wrap it in (N_  ...). In fact it's even better not to because as written "(+ 2 3)" will be added to the list of strings to translate...

src/report/utility-reports/hello-world.scm
```diff
@@ -354,9 +354,9 @@ or extending existing reports.")))
 new, totally cool report, consult the mailing list %s.")
           (gnc:html-markup-anchor 
            "mailto:gnucash-devel@gnucash.org"
-           (gnc:html-markup-tt "gnucash-devel@gnucash.org")))
-         (_ "For details on subscribing to that list, see &lt;http://www.gnucash.org/&gt;.")
-         (_ "You can learn more about writing scheme at &lt;http://www.scheme.com/tspl2d/&gt;."))
+           (gnc:html-markup-tt "gnucash-devel@gnucash.org"))
+          (_ "For details on subscribing to that list, see &lt;http://www.gnucash.org/&gt;. ")
+          (_ "You can learn more about writing scheme at &lt;http://www.scheme.com/tspl2d/&gt;.")))
```
gjanssens: This time I have applied your patch and ran some tests.

The change above appears to do different things for me than for you. Before your patch, I have this:

> For help on writing reports, or to contribute your brand new, totally cool report, consult the mailing list gnucash-devel@gnucash.org .For details on subscribing to that list, see <http://www.gnucash.org/>.You can learn more about writing scheme at <http://www.scheme.com/tspl2d/>.

After your patch I get this:

> For help on writing reports, or to contribute your brand new, totally cool report, consult the mailing list gnucash-devel@gnucash.org .

The rest of the text is gone completely. Can you check this ?

christopherlam: Whoops you're right.

In my Windows 7 machine the original had 4 elements within html-markup-p. It exports to HTML with correct markup, but within gnucash there is overwritten text after the <a href> element. I guess this is just a Webkit bug. 

My patched one is wrong because gnc:html-markup/format only needs 2 elements and will discard the other 2. Fixing bugs is hard!

image

gjanssens: Perhaps another approach to fixing this would be to add a line break after the a href ?

gjanssens:
You have closed this PR. What should I do with the small patch to add a mini REPL option ?

PR #157, opened 2017-07-09
Gtk3 update

Bob-IT:
This is the completion of the gtk3 changes after Geert copied my branch. Also included are the updates he proposed and fixes he found that I had missed and a couple I found.

Things to follow are getting rid of surplus GTK_CHECK_VERSION's, add application css file, and maybe combine the icon/pixmap folder.


gjanssens:
Thanks for the updates.

I have pushed the PR to master after a rebase and after dropping the pixmap related changed. As it happened I already worked on those myself today because the state of the pixmaps makefile was breaking make install. And I needed to get that fixed because it was blocking my fedora copr builds.

PR #159, opened 2017-07-15
Gtk3 update2

Bob-IT:
Replace deprecated functions in currency_accounting_widget and some missing accelerator keys along with a couple of other changes I found.


jralls:
Thanks, merged after a rebase.

PR #160, opened 2017-07-19
Minor changes to get build working for GCC7

bhardwajs:
- Edited guid.hpp and guid.cpp for exception specification deprecated in C++11
- Edited gnc-dense-cal.c for GCC7 warning for snprintf buffer overflow.


jralls:
I beat you to the changes in guid.[hc]pp, so you don't need that. We don't want the merge commits, you need to clean them out along with the add of gnc-vcs-info.h, which is generated at build time and shouldn't be in git.

That leaves increasing the size of dayNumBuf from 3 to 4. What's that about?

bhardwajs:
That is a gcc 7 warning for this change:

The -Wformat-truncation=*level* option detects certain and likely output
truncation in calls to the snprintf family of formatted output functions.
-Wformat-truncation=1 is included in -Wall and enabled without optimization
but works best with -O2 and higher.
https://gcc.gnu.org/gcc-7/changes.html

I can clean up rest and send a new pull request tonight if that's okay.
Please let me know.

Thanks,
Sumit

bhardwajs:
Abandoning the PR as it had multiple changes including guid.{c,h}pp changes that John already merged in master.

PR #161, opened 2017-07-21
membuf changes in gnc-dense-cal.c for gcc7 warnings

bhardwajs:
This PR fixes a gcc 7 warning for this [change](https://gcc.gnu.org/gcc-7/changes.html):

The -Wformat-truncation=*level* option detects certain and likely output truncation in calls to the snprintf family of formatted output functions. -Wformat-truncation=1 is included in -Wall and enabled without optimization but works best with -O2 and higher.


jralls:
Merged with a rebase, thanks.

PR #162, opened 2017-07-23
Gtk3 update3

Bob-IT:
Some fixes and the creation of an application css file for static css so it is not spread through out the source files. May need tweaking on location but it seemed a sensible location is in the ui folder.

Also some code clean up on unused variables, variables set but not used and some defines.
All seems OK to me and still builds and passed the travis tests.


jralls:
Brilliant, Thanks. Merged after rebase.

PR #163, opened 2017-07-27
Use portable test(1) operator.

0-wiz-0:
The standard test(1) comparison operator is "=".


0-wiz-0:
I've added a second patch to my fork which fixes the compilation in pkgsrc. pkg-config does not report the outside search path so that `<glib-2.0/glib.h>` will work, so I changed two files to use the standard `<glib.h>` include.

gjanssens:
Thanks :)

PR #164, opened 2017-08-01
Add vertical-align:top to default text fields.

christopherlam:
Add vertical-align top.

Previously, text fields would have vertical-align:default = centre. This means long transaction descriptions would cause all other fields to be vertically centred. This amendment will ensure the transactional report text are aligned from the top.

Cosmetic change only.


gjanssens:
Thanks :)

PR #165, opened 2017-08-01
Gtk3 update4

Bob-IT:
More updates without any of the colour changes that I had played with so hopefully this should be OK.

### DIFF
--- a/src/register/register-gnome/gnucash-header.c
+++ b/src/register/register-gnome/gnucash-header.c
@@ -464,7 +464,6 @@ gnc_header_event (GtkWidget *widget, GdkEvent *event)
             header->resize_col_width = cd->pixel_width;
             header->resize_x = x;
         }
-
         break;
     }
     case GDK_BUTTON_RELEASE:
@@ -485,7 +484,6 @@ gnc_header_event (GtkWidget *widget, GdkEvent *event)
             header->resize_col = -1;
             gnc_header_request_redraw (header);
         }
-
         break;
     }
 
@@ -516,14 +514,12 @@ gnc_header_event (GtkWidget *widget, GdkEvent *event)
             header->resize_col = -1;
             gnc_header_auto_resize_column (header, resize_col);
         }
-
     }
     break;
 
     default:
         break;
     }
-
     return FALSE;
 }
 
@@ -572,7 +568,6 @@ gnc_header_set_property (GObject *object,
     case PROP_SHEET:
         header->sheet = GNUCASH_SHEET (g_value_get_object (value));
         gtk_scrollable_set_hadjustment (GTK_SCROLLABLE(layout), header->sheet->hadj);
-
         needs_update = TRUE;
         break;
     case PROP_CURSOR_NAME:
--- a/src/register/register-gnome/gnucash-header.c
+++ b/src/register/register-gnome/gnucash-header.c
@@ -601,8 +601,8 @@ gnc_header_init (GncHeader *header)
     header->width = 400;
     header->style = NULL;
 
-    // Set the style context for this widget so it can be easily manipulated with css
-    gtk_style_context_add_class (gtk_widget_get_style_context (GTK_WIDGET(header)), "GncRegisterHeader");
+    // This sets a style class for when Gtk+ version is less than 3.20
+    gnc_widget_set_css_name (GTK_WIDGET(header), "header");
 
     gtk_widget_add_events(GTK_WIDGET(header), (GDK_EXPOSURE_MASK
                           | GDK_BUTTON_PRESS_MASK
@@ -621,6 +621,10 @@ gnc_header_class_init (GncHeaderClass *header_class)
 {
     GObjectClass  *object_class = G_OBJECT_CLASS (header_class);
     GtkWidgetClass *item_class = GTK_WIDGET_CLASS (header_class);
+
+#if GTK_CHECK_VERSION(3,20,0)
+    gtk_widget_class_set_css_name (GTK_WIDGET_CLASS(header_class), "header");
+#endif
 
     parent_class = g_type_class_peek_parent (header_class);
 
--- a/src/register/register-gnome/gnucash-header.c
+++ b/src/register/register-gnome/gnucash-header.c
@@ -51,10 +51,15 @@ enum
 static gboolean
 gnc_header_draw (GtkWidget *header, cairo_t *cr)
 {
-    cairo_save (cr);
-    cairo_set_source_surface (cr, GNC_HEADER(header)->surface, 0, 0);
+    GnucashSheet *sheet = GNC_HEADER(header)->sheet;
+    GdkWindow *sheet_layout_win = gtk_layout_get_bin_window (GTK_LAYOUT(sheet));
+    gint x, y;
+
+    // use this to get the scroll x value to align the header
+    gdk_window_get_position (sheet_layout_win, &x, &y);
+
+    cairo_set_source_surface (cr, GNC_HEADER(header)->surface, x, 0);
     cairo_paint (cr);
-    cairo_restore (cr);
 
     return TRUE;
 }
--- a/src/register/register-gnome/gnucash-header.c
+++ b/src/register/register-gnome/gnucash-header.c
@@ -106,13 +106,18 @@ gnc_header_draw_offscreen (GncHeader *header)
                                                 header->height);
 
     cr = cairo_create (header->surface);
+    // Fill background color of header
     cairo_rectangle (cr, 0.5, 0.5, header->width - 1.0, header->height - 1.0);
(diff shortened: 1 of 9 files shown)


src/register/register-gnome/gnucash-sheet-private.c
```diff
@@ -699,19 +699,28 @@ gnucash_sheet_draw_cursor (GnucashCursor *cursor, cairo_t *cr)
...
+
+    if (cursor->y == 0)
+        cairo_rectangle (cr, cursor->x - x + 0.5, cursor->y - y + 0.5, cursor->w - 1.0, cursor->h - 3.0);
+    else
+    {
+        if (cursor->x == 0)
+            cairo_rectangle (cr, cursor->x - x + 0.5, cursor->y - y - 0.5, cursor->w - 1.0, cursor->h - 2.0);
+        else
+            cairo_rectangle (cr, cursor->x - x - 0.5, cursor->y - y - 0.5, cursor->w, cursor->h - 2.0);
+    }
```
gjanssens: It seems your desire to display a frame border on first row/cell does require quite a bit of extra work, considering this commit, the next and parts of the commit in which you redo all the border drawing.
How about instead keeping the borders as they were and draw one black frame around the whole visible area of the gtklayout ? That would avoid all the exceptions you have to code now at the expense of some redundant lines at the bottom and lower parts of the left and right border if the row don't completely fill the visible area.
A variation would be to calculate the vertical position of the bottom most row and draw the left-most and right-most border in one go at the end. when all rows are drawn.

src/register/register-gnome/gnucash-header.c
```diff
@@ -606,6 +606,9 @@ gnc_header_init (GncHeader *header)
     header->width = 400;
     header->style = NULL;
 
+    // Set the style context for this widget so it can be easily manipulated with css
+    gtk_style_context_add_class (gtk_widget_get_style_context (GTK_WIDGET(header)), "GncRegisterHeader");
+
```
gjanssens: Ideally we should be using
`gtk_widget_class_set_css_name` rather than `gtk_style_context_add_class`. Unfortunately this is only available as of gtk 3.20, so that's of little use now. It looks like the css code was in heavy flux in the range of gtk versions we currently support.
I'm not sure what is the best alternative. `gtk_style_context_add_class` or `gtk_widget_set_name`. For now I don't think this needs changing, but it's something to keep in mind for the future.

src/register/register-gnome/gnucash-sheet-private.c
```diff
@@ -342,7 +342,7 @@ draw_hatching (cairo_t *cr,
     cairo_move_to (cr, h_x, h_y);
     cairo_rel_line_to (cr, h_size, h_size);
 
-    cairo_rel_move_to (cr, -h_x, 0);
+    cairo_move_to (cr, h_x , h_y + h_size);
```
gjanssens: Nevermind this one. In the bigger context your line works equally fine.

src/register/register-gnome/gnucash-sheet-private.c
```diff
@@ -423,42 +423,41 @@ draw_cell (GnucashSheet *sheet,
     /* top */
     draw_cell_line (cr, bg_color,
                     (borders.top >= borders.left ? x : x + 1.0),
-                    y + 0.5,
+                    (y == 0 ? y + 0.5 : y - 0.5),
```
gjanssens: The lost double line is due to another commit. When you reworked the cursor drawing, you apparently removed the extra line drawn there...

src/register/register-gnome/gnucash-header.c
gjanssens: More generally, I would use a naming convention similar to what gtk is doing. Don't add "Gnc" and use lowercase for the names.
Ideally you shouldn't even have to mention "register" because the header is a child widget of the register widget. Unfortunately that only works if you can have real css nodes, and may be more difficult to implement with css classes.

src/register/register-gnome/gnucash-item-edit.c
```diff
@@ -505,6 +505,9 @@ gnc_item_edit_new (GnucashSheet *sheet)
                            NULL);
     gtk_layout_put (GTK_LAYOUT(sheet), GTK_WIDGET(item_edit), 0, 0);
 
+    // Set the style context for this widget so it can be easily manipulated with css
+    gtk_style_context_add_class (gtk_widget_get_style_context (GTK_WIDGET(item_edit)), "GncRegisterItemEdit");
+
```
gjanssens: GncRegisterItemEdit is another name example of describing the implementation instead of the function. This widget is an implementation of a cell cursor (the "purpose"), so use that as css node/class description. The idea is the theme designer is not interested in how something is implemented, but what the function is of a given node so the defined style can support this function. And ideally it should not matter how we implement it internally and the chosen style should work equally well regardless of whether the register is implemented using gtklayout/gtkentry or gtktreeview. (That's unlikely to realize fully, but it helps to think in those terms to better separate function from implementation).

gjanssens: To elaborate a bit on this, ideally we should create nodes/classes (I'm always mentioning both due to the functions that only first appear in gtk 3.20) for

- the register itself
- the register header
- the register body
- a register row (optionally providing even/odd differentiation via flags)
- the (row) cursor
- the cell cursor

Each of these can take css style properties such as color, font, padding, border color, border width,...
We don't have to define all of those ourselves though. We should implement in a way these are taken into account when a theme designer sets them.

And I also think this may be more work than we have time for before 2.8. Yet we should lay at least the groundwork now that we are redoing the register. And these considerations go for all of our custom widgets by the way, but most of them are just a combination of existing widgets, so they style automatically. The register and the SX calendar overview are two exceptions because they are based on GtkLayout so that makes us responsible for properly handling all styling. (Which makes me wonder again whether going the GtkTreeView route may still be the better solution in the long run...)

src/register/register-gnome/gnucash-sheet-private.c
Bob-IT: I have changed this anyway.

src/register/register-gnome/gnucash-sheet-private.c
```diff
@@ -369,7 +369,7 @@ draw_divider_line (cairo_t *cr, VirtualLocation virt_loc,
 
     cairo_set_source_rgb (cr, fg_color->red, fg_color->green, fg_color->blue);
 
-    cairo_set_line_width (cr, 3.0);
+    cairo_set_line_width (cr, 2.0);
```
Bob-IT: I have changed this back to 3px and made it straddle the two rows centrally.

src/register/register-gnome/gnucash-sheet-private.c
Bob-IT: Changed this to always subtract 0.5 from the y axis so all rows should be the same. The first row uses the header border for the top border.

src/register/register-gnome/gnucash-sheet-private.c
```diff
@@ -423,42 +423,41 @@ draw_cell (GnucashSheet *sheet,
...
-                    x + 0.5,
-                    (borders.left > borders.top ? y : y + 1),
-                    x + 0.5,
+                    (x == 0 ? x + 0.5 : x - 0.5),
+                    (borders.left > borders.top ? y : y),
+                    (x == 0 ? x + 0.5 : x - 0.5),
                     (borders.left > borders.bottom ?
-                     y + height : y + height - 1),
+                     y + height : y + height),
                     borders.left);
```
Bob-IT: Sorry, I just thought it was a mistake as I could see all these one pixel dots every where.

src/register/register-gnome/gnucash-sheet-private.c
```diff
@@ -561,7 +560,7 @@ draw_cell (GnucashSheet *sheet,
 
 
 
-    cairo_move_to (cr, rect.x + x_offset, rect.y + 1);
+    cairo_move_to (cr, rect.x + x_offset, rect.y - 1); //FIXME Not sure about this change from + 1
```
Bob-IT: Changed back.

Bob-IT:
Hope this is better, I wanted the left border shown as I always thought it looked odd when the right one aligned with the balance cell border. As I have changed the first row alignment, this has reduced the complexity and hopefully is acceptable. The problem with just drawing a vertical line in the left most cell is when splits an in view you do not have a complete vertical line.

Bob

gjanssens:
Much better, thanks, though it looks like you missed the resizing if the GncItemEdit while on the first line :)
It's one pixel too small on that line.

Bob-IT:
I missed that, will change and push later, just run gnucash with 'Adwaita-dark', looks totally different but doable I think, will sort the calendar out first.

gjanssens:
Thanks!

PR #166, opened 2017-08-01
Fix income statement when trading accounts not used

christopherlam:
Fix cosmetic bug
http://gnucash.1415818.n4.nabble.com/unknown-column-in-Income-Statement-Profit-amp-Loss-td4680661.html#a4692940

Insert an empty-cell instead of #unspecified, which is the result of (if predicate? result-if-true) where predicate? is #false

### DIFF
--- a/src/report/standard-reports/income-statement.scm
+++ b/src/report/standard-reports/income-statement.scm
@@ -647,14 +647,16 @@
 		  (if standard-order?
 		      (list
 		       (gnc:make-html-table-cell inc-table)
-		       (if (not (null? trading-accounts))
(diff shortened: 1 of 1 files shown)


gjanssens:
Merged after a rebase. Thanks.

Just a note - if you intend to make additonal small changes, there's no need to create independent pull requests for each. You can simply update your existing pull request (as long as it's open) by pushing extra commits to the same branch in your github repo. So if you make one branch "small-updates" or similar, and create a first PR from this branch, you can continue to add commits to it until it gets merged. This makes it easer for me or another dev too as there's only one PR to handle.

PR #167, opened 2017-08-03
PR for partial fix for bug 660027

christopherlam:
This will partially fix bug 660027 when the report is HTML Exported, and the resultant file is opened and printed from a modern browser. I can confirm the table headers will then be repeated in every page... perhaps one day webkit will be upgraded!

### DIFF
--- a/src/report/report-system/html-document.scm
+++ b/src/report/report-system/html-document.scm
@@ -286,7 +286,7 @@
                       (attr (gnc:warn "non-string attribute" attr)))))
              (build-first-tag
               (lambda (tag)
-                (push "\n<") (push tag)
+                (push "<") (push tag)
                 (if attr (hash-fold add-attribute #f attr))
                 (if extra-attrib (for-each addextraatt extra-attrib))
                 (if (not end-tag?)
@@ -336,7 +336,7 @@
             (let ((addtag (lambda (t)
                             (push "</")
                             (push tag)
-                            (push ">"))))
+                            (push ">\n"))))
               (cond
                ((string? tag)
                 (addtag tag))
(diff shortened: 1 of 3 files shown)


christopherlam:
Apologies for spam; the first 3 commits will definitely help toward fixing this bug.

gjanssens:
@christopherlam can you reduce this PR to two commits only please?

- one that squashes the first 2 commits together (as "(Partial) Fix for bug 660027" or similar commit message)
- one for the cosmetic changes (whitespace adjustments)

This will make the git history slightly cleaner.

You can achieve this with `git rebase -i <parent commit>` (you need to replace <parent commit> with a reference to the commit right before the start of your own branch. This can be a commit hash or another branch that point to that (or an older) commit). In the interactive (text)window that will appear replace 'pick' with 'fixup' (or simple 'f') for each commit you want to melt together with a previous commit. So in this PR that would be for the second commit (to integrate it into the first commit) and for commit 4,5,6 to (to integrate them into the 3rd commit). In the same rebase action you can also tell git you want to change the commit message on the 3rd commit (to something like "Improve whitespace in source and rendered report") by changing the "pick" before that commit by 'edit' (or 'e').
Final step is to save and close the change overview. Only then git will start the actual changes and it will prompt you for the new commit message at some point.

This assumes you know how to use a command line editor such as vi(m), emacs, nano, pico,... git will by default use the one that's the default on your system. You could change this (even to use a graphical text editor) by setting environment variable EDITOR before calling git rebase ...
For example

  EDITOR=kwrite git rebase -i master

gjanssens:
@christopherlam did you receive my feedback on this ? I'd like to merge your work so I was wondering if you are capable of performing my suggestions. If not, I can do them for you in this case as well if that would help you.

christopherlam:
Apologies, please go ahead and merge as appropriate. I am not within
broadband connection and cannot do git console commands right now. Thanks.

christopherlam:
PS thank you for the tutorials on git

gjanssens:
You're welcome and I have now applied your changes to maint after a rebase and squashing the minor changes together. Thanks!

PR #168, opened 2017-08-09
Gtk3 update5

Bob-IT:
Highlights...
Add a fall back css file that has a lower priority than themes.
Change the way the dense calendar is drawn, now theme-able and can look more like the default calendar, below is some css that hopefully should give you an idea, the only thing I was not sure about was the default names...
Changed the up/down arrow to be drawn based on theme which brings it more inline with the rest of the look.
Bob

calendar {
  color:yellow;
  background-color: lightgrey;
}

calendar.header {
 background-color: lightgreen;
}

calendar.highlight { color: red;
  background-color: pink;
}

calendar:indeterminate { color: alpha(currentColor,0.2); }

calendar:selected { 
  color: white;
  background-color: grey;
  border-radius: 10px;
}

.calendar {
  border-color: white;
}

.calendar.header {
 background-color: lightgreen;
}

.calendar:selected { 
  color: white;
  background-color: grey;
  border-radius: 10px;
}

.calendar.primary {
  background-color: darksalmon;
}

.calendar.secondary {
  background-color: darkseagreen;
}

.calendar.markers {
  background-color: indianred;
}

#dense-cal-popup treeview {
  background-color: lightcoral;
}

### DIFF
--- a/gnucash/gnome-utils/gnc-dense-cal.c
+++ b/gnucash/gnome-utils/gnc-dense-cal.c
@@ -1030,7 +1030,15 @@ gnc_dense_cal_draw_to_buffer(GncDenseCal *dcal)
                 center_y = (y1 + y2 ) / 2;
                 radius = MIN((x2 - x1), (y2 - y1)) * .75;
 
-                gtk_render_background (stylectxt, cr, center_x - radius - 2, center_y - radius, (radius * 2) + 4, radius * 2);
+                // try to compensate for row height being odd or even
+                if (((y2 -y1) % 2) == 0)
+                    gtk_render_background (stylectxt, cr,
+                                           center_x - radius - 2, center_y - radius - 1,
+                                            (radius * 2) + 4, radius * 2);
+                else
+                    gtk_render_background (stylectxt, cr,
+                                           center_x - radius - 2, center_y - radius,
+                                            (radius * 2) + 4, (radius * 2) + 1);
             }
         }
         gtk_style_context_restore (stylectxt);
--- a/gnucash/gnome-utils/gnc-dense-cal.c
+++ b/gnucash/gnome-utils/gnc-dense-cal.c
@@ -923,6 +923,20 @@ gnc_style_context_get_border_color (GtkStyleContext *context,
     gdk_rgba_free (c);
 }
 
+static gboolean
+is_color_dark (GdkRGBA *color)
+{
+    gboolean is_dark = FALSE;
+
+    // Counting the perceptive luminance - human eye favors green color...
+    double a = (0.299 * color->red + 0.587 * color->green + 0.114 * color->blue);
+
+    if (a > 0.5)
+        is_dark = TRUE;
+
+    return is_dark;
+}
+
 static void
 gnc_dense_cal_draw_to_buffer(GncDenseCal *dcal)
 {
@@ -935,6 +949,7 @@ gnc_dense_cal_draw_to_buffer(GncDenseCal *dcal)
     PangoLayout *layout;
     GTimer *timer;
     cairo_t *cr;
+    const gchar *primary_color_class, *secondary_color_class, *marker_color_class;
 
     timer = g_timer_new();
     g_debug("drawing");
@@ -957,6 +972,22 @@ gnc_dense_cal_draw_to_buffer(GncDenseCal *dcal)
     gtk_render_background (stylectxt, cr, 0, 0,
                            cairo_image_surface_get_width (dcal->surface),
                            cairo_image_surface_get_height (dcal->surface));
+
+    /* get the colors */
+    {
+         GdkRGBA color;
+         gchar *color_extension = NULL;
+
+         gtk_style_context_get_color (stylectxt, GTK_STATE_FLAG_NORMAL, &color);
+
+          if (is_color_dark (&color))
+              color_extension = "-dark";
+
+          primary_color_class = g_strconcat ("primary", color_extension, NULL);
+          secondary_color_class = g_strconcat ("secondary", color_extension, NULL);
+          marker_color_class = g_strconcat ("markers", color_extension, NULL);
+    }
+
 
     /* Fill in alternating month colors. */
     {
@@ -984,20 +1015,15 @@ gnc_dense_cal_draw_to_buffer(GncDenseCal *dcal)
             for (mcListIter = mcList; mcListIter != NULL; mcListIter = mcListIter->next)
             {
                 rect = (GdkRectangle*)mcListIter->data;
+                gtk_style_context_save (stylectxt);
 
                 if (i % 2 == 0)
-                {
-                    gtk_style_context_add_class (stylectxt, "primary");
-                    if (gtk_style_context_has_class (stylectxt, "secondary"))
-                        gtk_style_context_remove_class (stylectxt, "secondary");
-                }
+                    gtk_style_context_add_class (stylectxt, primary_color_class);
                 else
-                {
-                    if (gtk_style_context_has_class (stylectxt, "primary"))
-                        gtk_style_context_remove_class (stylectxt, "primary");
-                    gtk_style_context_add_class (stylectxt, "secondary");
-                }
+                    gtk_style_context_add_class (stylectxt, secondary_color_class);
+
                 gtk_render_background (stylectxt, cr, rect->x, rect->y, rect->width, rect->height);
+                gtk_style_context_restore (stylectxt);
             }
             g_list_foreach(mcList, free_rect, NULL);
             g_list_free(mcList);
@@ -1013,7 +1039,7 @@ gnc_dense_cal_draw_to_buffer(GncDenseCal *dcal)
         int x1, x2, y1, y2;
 
         gtk_style_context_save (stylectxt);
-        gtk_style_context_add_class (stylectxt, "markers");
+        gtk_style_context_add_class (stylectxt, marker_color_class);
 #if GTK_CHECK_VERSION(3,22,0)
         gtk_style_context_add_class (stylectxt, GTK_STYLE_CLASS_VIEW);
         gtk_style_context_set_state (stylectxt, GTK_STATE_FLAG_SELECTED);
@@ -1064,12 +1090,16 @@ gnc_dense_cal_draw_to_buffer(GncDenseCal *dcal)
         w = col_width(dcal) - COL_BORDER_SIZE - dcal->label_width;
         h = col_height(dcal);
 
+        gtk_style_context_save (stylectxt);
+
         /* draw the outside border [inside the month labels] */
+        gtk_style_context_add_class (stylectxt, GTK_STYLE_CLASS_FRAME);
+
+        gtk_render_frame (stylectxt, cr, x, y, w + 1, h + 1);
+
         gnc_style_context_get_border_color (stylectxt, state_flags, &color);
         cairo_set_source_rgb (cr, color.red, color.green, color.blue);
         cairo_set_line_width (cr, 1);
-        cairo_rectangle (cr, x + 0.5, y + 0.5, w, h);
-        cairo_stroke (cr);
 
         /* draw the week separations */
         for (j = 0; j < num_weeks_per_col(dcal); j++)
@@ -1089,6 +1119,7 @@ gnc_dense_cal_draw_to_buffer(GncDenseCal *dcal)
             cairo_stroke (cr);
         }
         cairo_restore (cr);
+        gtk_style_context_restore (stylectxt);
 
 
         /* draw the day of the week labels */
--- a/gnucash/gnome-utils/gnc-dense-cal.c
+++ b/gnucash/gnome-utils/gnc-dense-cal.c
@@ -25,6 +25,7 @@
 #include "gnc-dense-cal.h"
 #include "gnc-dense-cal-model.h"
 #include "gnc-engine.h"
+#include "gnc-gtk-utils.h"
 #include <glib.h>
 #include <glib/gi18n.h>
 #include <gtk/gtk.h>
@@ -923,21 +924,6 @@ gnc_style_context_get_border_color (GtkStyleContext *context,
     gdk_rgba_free (c);
 }
 
-static gboolean
-is_color_light (GdkRGBA *color)
-{
-    gboolean is_light = FALSE;
-
-    // Counting the perceptive luminance - human eye favors green color...
-    double a = (0.299 * color->red + 0.587 * color->green + 0.114 * color->blue);
-
-    if (a > 0.5)
-        is_light = TRUE;
-
-    return is_light;
-}
-
-static void
 gnc_dense_cal_draw_to_buffer(GncDenseCal *dcal)
 {
     GtkWidget *widget;
@@ -980,7 +966,7 @@ gnc_dense_cal_draw_to_buffer(GncDenseCal *dcal)
 
          gtk_style_context_get_color (stylectxt, GTK_STATE_FLAG_NORMAL, &color);
 
-          if (is_color_light (&color))
+          if (gnc_is_dark_theme (&color))
               class_extension = "-dark";
 
           primary_color_class = g_strconcat ("primary", class_extension, NULL);
--- a/gnucash/gnome-utils/gnc-dense-cal.c
+++ b/gnucash/gnome-utils/gnc-dense-cal.c
@@ -69,11 +69,6 @@ static const int DENSE_CAL_DEFAULT_HEIGHT = 105;
 static const int MINOR_BORDER_SIZE = 1;
 static const int COL_BORDER_SIZE = 3;
 
-static const gchar* MONTH_THIS_COLOR = "lavender";
-static const gchar* MONTH_THAT_COLOR = "SlateGray1";
-
-static const gchar* MARK_COLOR = "Yellow";
-
 #undef G_LOG_DOMAIN
 #define G_LOG_DOMAIN "gnc.gui.dense-cal"
 
@@ -226,6 +221,10 @@ gnc_dense_cal_class_init(GncDenseCalClass *klass)
     object_class = G_OBJECT_CLASS (klass);
     widget_class = GTK_WIDGET_CLASS (klass);
 
+#if GTK_CHECK_VERSION(3,20,0)
+    gtk_widget_class_set_css_name (GTK_WIDGET_CLASS(klass), "calendar");
+#endif
+
     parent_class = g_type_class_peek_parent (klass);
 
     object_class->finalize = gnc_dense_cal_finalize;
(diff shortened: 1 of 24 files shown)


Bob-IT:
Forgot to add that with gtk version 3.22 as specified on line 1018 of gnc-dense-cal.c the default behaviour of the mark colour would be the theme selected colour, it may work on earlier versions but certainly not 3.10, I thought this was a good idea but if not the if block can be removed.

Bob

Bob-IT:
Reworked the css for the negative label colour, this does not include the register as that will be done when that is converted to use css. On that, if the default colours are in the fallback css file and can be over-ridden in the theme or by the user is there a need for the preference ?

Bob-IT:
This completes the move of the css to the fallback file with hopefully better choices of names.
Bob

gjanssens:
To answer your question on the need for a preference first: I think it should be kept for now and will explain here why I believe so.

There are 3 parties involved in styling gnucash: the gnucash developers, the gtk theme designers and the users. In general widgets should blend in with their environment, that is match in style with surrounding widgets, unless there is good reason not to.

This is primarily the job gtk theme designers do. Provide a coherent visual representation. However this usually only works for common widgets (that is the widgets in the Gtk library). Widgets that aren't in this common set, but instead are drawn directly by the application (such as our register in its current implementation) will usually not receive styling info from theme designers, so we'd have to set up a fallback that mimics this behaviour: blend in as well as possible with the other widgets in the gtk set.

On top of that, gnucash developers in the past were of the opinion the register concept could be much easier to grasp if it resembled a real ledger on paper and so decided to use the yellow/green theming. Clearly there has been opposition to this as well because there is now an option to enable or disable this custom theme. Back to our consideration this theme is clearly something gnucash specific, not general theme specific. So by Gtk's guidelines that belongs in the APPLICATION priority.

And finally a user not happy with any of the options (fallback, theme, application) can tweak the style even more because we will parse a user level css file as well.

Now suppose you would make the yellow/green theme the fallback theme instead of an application theme (which I think you were suggesting in your question). Then

1. it will be very unlikely there will be a global gtk theme that will override this style, because the register widget is gnucash specific and no general theme is likely to care about it.
2. the user can override it via custom css in her home directory. That would essentially mean copy the css rules from our fallback theme and tweak it's values to mimic the current general theme.

However... ideally the fallback theme is compiled in as a gresource to ensure it always exists, so it will be difficult for the user to copy these in the first place. In addition whenever the general theme changes (either by user choice or because of a system update) the user's local changes have to be updated as well. Not very user friendly and probably beyond the technical skills of many users of an accounting package.

Hence it's not a good idea to use the custom yellow/green theme for our fallback. Our fallback should query theme properties from the general theme and use those instead. I haven't investigated this in detail myself yet, but I would look at the gtk-specific extensions for colors like @bg_color and the like as described on the [gtk css overview](https://developer.gnome.org/gtk3/unstable/chap-css-overview.html).

src/gnome-utils/ui/gnucash-fallback.css
```diff
@@ -2,3 +2,16 @@
...
+  background-color: lavender;
+}
+
+*.secondary {
+  background-color: SlateGray1;
+}
+
+*.markers {
+  background-color: yellow;
+}
```
gjanssens: If we want to keep these as default colors, we'll have to set text colors as well to match them. In the dark theme the text color is white, which will be pretty hard to read.

I'd propose however  let go of these colors and instead expermiment with colors that blend in with the theme like background color +/-10%. Or if we must have a gnucash specific style, we can try to reuse the yellow/green theme also used in the register.
These gnucash specific colors can be activated/deactivated via the same single preference we already have for the register. In that case the preference should become more generic of course and probably be presented in a different preference tab.

src/register/register-gnome/gnucash-item-edit.c
```diff
@@ -167,8 +167,7 @@ gnc_item_edit_init (GncItemEdit *item_edit)
 
     item_edit->popup_toggle.ebox = NULL;
     item_edit->popup_toggle.tbutton = NULL;
-    item_edit->popup_toggle.arrow_up = NULL;
-    item_edit->popup_toggle.arrow_down = NULL;
+    item_edit->popup_toggle.direction = TRUE;
```
gjanssens: This is a variablename that doesn't suggest a boolean value. Better would be something like arrow_points_up or something like that.
Other than that I very much appreciate this change. Much cleaner.

src/register/register-gnome/gnucash-item-edit.c
```diff
@@ -333,6 +332,32 @@ unblock_toggle_signals(GncItemEdit *item_edit)
 }
 
 
+static gboolean
+draw_arrow_cb (GtkWidget *widget, cairo_t *cr, gpointer data)
```
gjanssens: Why not reuse the arrow drawing callback you wrote for the rest of gnucash ? That is if you add direction as data instead of item_edit, you have essentially the same interface in both cases. Particularly as you only use item_edit in the function below to extract the arrow direction.

src/gnome-utils/dialog-utils.c
```diff
@@ -69,7 +69,7 @@ gnc_set_label_color(GtkWidget *label, gnc_numeric value)
     deficit = gnc_numeric_negative_p (value);
 
     if (deficit)
-        gnc_widget_set_style_context (GTK_WIDGET(label), "negative-color");
+        gnc_widget_set_style_context (GTK_WIDGET(label), "negative-numbers");
```
gjanssens: Yes, numbers is much more to the point than color. Thanks!

src/import-export/import-main-matcher.c
```diff
@@ -82,9 +82,9 @@ enum downloaded_cols
     NUM_DOWNLOADED_COLS
 };
 
-#define COLOR_RED    "brown1"
-#define COLOR_YELLOW "gold"
-#define COLOR_GREEN  "DarkSeaGreen1"
+#define COLOR_RED_CLASS    "intervention-required"
+#define COLOR_YELLOW_CLASS "intervention-probably-required"
+#define COLOR_GREEN_CLASS  "intervention-not-required"
```
gjanssens: The values are good (intervention-xyz). Now the same should still happen with the parameter itself. COLOR_RED_CLASS etc is meaningless in the code as well.

src/import-export/import-main-matcher.c
```diff
@@ -215,13 +215,13 @@ on_matcher_help_clicked (GtkButton *button, gpointer user_data)
     gnc_builder_add_from_file (builder, "dialog-import.glade", "matcher_help_dialog");
 
     box = GTK_WIDGET(gtk_builder_get_object (builder, "red"));
```
gjanssens: And the names in glade should be updated equally.

src/import-export/import-main-matcher.c
```diff
@@ -481,10 +477,6 @@ gnc_gen_trans_init_view (GNCImportMainMatcher *info,
              "cell-background", DOWNLOADED_COL_COLOR,
              NULL);
     renderer = gtk_cell_renderer_text_new();
-    g_object_set(G_OBJECT(renderer),
-                 "foreground", "black",
-                 "foreground-set", TRUE,
-                 NULL);
```
gjanssens: Hmm, as with the dense cal, I think you should set a text color at the same place you set a background color. Since we define default background colors based on the intervention required, we should set accompanying foreground colors IMO. Otherwise it will depend on the theme whether the text rendered is readable on our chose backgrounds. Moreover I even think we should define separate foreground colors for each of our 3 classes.

src/gnome-utils/ui/gnucash-fallback.css
```diff
@@ -30,6 +30,13 @@
   border-color: black;
 }
 
+/* Make Label bold and larger */
+.emphasize-label {
```
gjanssens: Good choice of class name! The comment is redundant though or too specific. It can be "Make label more important". We happened to choose to do this by making it bold and larger, but someone else may want to do this differently by overriding our class.

gjanssens:
Aside from the code review, I have also observed this while merging this PR:

- the newly added file gnucash-fallback.css will be stored in the wrong location (src/gnome-utils/ui instead of gnucash/gnome-utils/ui). This is to be expected, but we need to keep this in mind in the final PR merge or you have to remember it if you decide to rebase your PR to past my directory reshuffling.
- when I merge this PR, I get guile errors while loading reports. I haven't had time to investigate this. The relevant error messages are here:
```
In ice-9/boot-9.scm:
 174: 19 [with-throw-handler #t #<catch-closure 2797760> #<catch-closure 2797740>]
In unknown file:
   ?: 18 [apply-smob/1 #<catch-closure 2797760>]
   ?: 17 [call-with-input-string "(gnc:report-run 3)" ...]
In ice-9/boot-9.scm:
2404: 16 [save-module-excursion #<procedure 2a01fc0 at ice-9/eval-string.scm:65:9 ()>]
In ice-9/eval-string.scm:
  44: 15 [read-and-eval #<input: string 315eea0> #:lang ...]
  37: 14 [lp (gnc:report-run 3)]
In report.scm:
 753: 13 [gnc:report-run 3]
In ice-9/boot-9.scm:
 160: 12 [catch ignore #<procedure 2a01d20 at main.scm:116:4 ()> ...]
In unknown file:
   ?: 11 [lazy-catch #t #<procedure 2a01cc0 at main.scm:118:18 ()> ...]
In ice-9/boot-9.scm:
 174: 10 [with-throw-handler #t #<catch-closure 27973a0> #<catch-closure 2797380>]
In unknown file:
   ?: 9 [apply-smob/1 #<catch-closure 27973a0>]
In report.scm:
 757: 8 [#<procedure 2a01de0 at report.scm:754:5 ()>]
 728: 7 [gnc:report-render-html # #t]
In view-column.scm:
  94: 6 [render-view #]
In ice-9/boot-9.scm:
 771: 5 [for-each #<procedure 2a01bd0 at view-column.scm:95:7 (_)> (# # # # ...)]
In view-column.scm:
 103: 4 [#<procedure 2a01bd0 at view-column.scm:95:7 (_)> #<variable 2897f00 value: #>]
  65: 3 [make-child-options-callback # #f]
In ice-9/boot-9.scm:
1384: 2 [#<procedure 2b594b0 at ice-9/boot-9.scm:1383:4 (obj)> #f]
 105: 1 [#<procedure 2ad15c0 at ice-9/boot-9.scm:100:6 (thrown-k . args)> wrong-type-arg ...]
In unknown file:
   ?: 0 [apply-smob/1 #<catch-closure 2797380> wrong-type-arg ...]
ERROR: In procedure apply-smob/1:
ERROR: In procedure struct_vtable: Wrong type argument in position 1 (expecting struct): #f
* 18:39:49  WARN <gnc.scm> In ice-9/boot-9.scm:
 174: 19 [with-throw-handler #t #<catch-closure 2797760> #<catch-closure 2797740>]
In unknown file:
   ?: 18 [apply-smob/1 #<catch-closure 2797760>]
   ?: 17 [call-with-input-string "(gnc:report-run 3)" ...]
In ice-9/boot-9.scm:
2404: 16 [save-module-excursion #<procedure 2a01fc0 at ice-9/eval-string.scm:65:9 ()>]
In ice-9/eval-string.scm:
  44: 15 [read-and-eval #<input: string 315eea0> #:lang ...]
  37: 14 [lp (gnc:report-run 3)]
In report.scm:
 753: 13 [gnc:report-run 3]
In ice-9/boot-9.scm:
 160: 12 [catch ignore #<procedure 2a01d20 at main.scm:116:4 ()> ...]
In unknown file:
   ?: 11 [lazy-catch #t #<procedure 2a01cc0 at main.scm:118:18 ()> ...]
In ice-9/boot-9.scm:
 174: 10 [with-throw-handler #t #<catch-closure 27973a0> #<catch-closure 2797380>]
In unknown file:
   ?: 9 [apply-smob/1 #<catch-closure 27973a0>]
In report.scm:
 757: 8 [#<procedure 2a01de0 at report.scm:754:5 ()>]
 728: 7 [gnc:report-render-html # #t]
In view-column.scm:
  94: 6 [render-view #]
In ice-9/boot-9.scm:
 771: 5 [for-each #<procedure 2a01bd0 at view-column.scm:95:7 (_)> (# # # # ...)]
In view-column.scm:
 103: 4 [#<procedure 2a01bd0 at view-column.scm:95:7 (_)> #<variable 2897f00 value: #>]
  65: 3 [make-child-options-callback # #f]
In ice-9/boot-9.scm:
1384: 2 [#<procedure 2b594b0 at ice-9/boot-9.scm:1383:4 (obj)> #f]
 105: 1 [#<procedure 2ad15c0 at ice-9/boot-9.scm:100:6 (thrown-k . args)> wrong-type-arg ...]
In unknown file:
   ?: 0 [apply-smob/1 #<catch-closure 2797380> wrong-type-arg ...]
ERROR: In procedure apply-smob/1:
ERROR: In procedure struct_vtable: Wrong type argument in position 1 (expecting struct): #f
```
- I also saw some Gtk critical warnings about an invalid cast from a GtkCombobox to a GtkEntry, which I think happened when playing with the SX overview, but that may be inaccurate.

src/gnome-utils/gnc-gnome-utils.c
```diff
@@ -182,13 +182,20 @@ gnc_add_css_file (void)
 
     provider_user = gtk_css_provider_new ();
     provider_app = gtk_css_provider_new ();
+    provider_fallback = gtk_css_provider_new ();
     display = gdk_display_get_default ();
     screen = gdk_display_get_default_screen (display);
+
+    gtk_style_context_add_provider_for_screen (screen, GTK_STYLE_PROVIDER (provider_fallback), GTK_STYLE_PROVIDER_PRIORITY_THEME - 50);
```
Bob-IT: When I first tried this I thought it did not work, tested again and it does, will change.

src/gnome-utils/ui/gnucash-fallback.css
```diff
@@ -0,0 +1,4 @@
+/* Note: Widgets obtained from Glade files will not be addressable
+   directly by name as these are not brought in when loaded, only
+   the widget type can be configured unless they are named in code */
```
Bob-IT: Looks like the widgets need to be named in the glade files, then you can use some thing like #imap-dialog or you can add classes to them

src/gnome-utils/gnc-dense-cal.c
```diff
@@ -1021,21 +1031,25 @@ gnc_dense_cal_draw_to_buffer(GncDenseCal *dcal)
...
     {
+        GdkRGBA color;
         gint x, y, w, h;
         gint j;
 
+        cairo_save (cr);
         pango_layout_set_text(layout, "S", -1);
         pango_layout_get_pixel_size(layout, NULL, &dcal->dayLabelHeight);
 
+        gdk_rgba_parse (&color, "black");
```
Bob-IT: The black is just a default, the colour should come from the border.

src/register/register-gnome/gnucash-item-edit.c
Bob-IT: Changed this.

src/register/register-gnome/gnucash-item-edit.c
Bob-IT: Interesting one this, in the other locations the arrow is only required to be set in one direction so it would not work in this location.

src/import-export/import-main-matcher.c
Bob-IT: Changed this.

src/import-export/import-main-matcher.c
Bob-IT: And also this.

src/import-export/import-main-matcher.c
Bob-IT: I do not think one should change the text colour but make the background colours darker if the text colour is light just like I have done in the dense calendar.

Bob-IT:
I have made some changes which hopefully are acceptable the main being making the background colours of the dense calendar a darker shade when the text colour is white. If this is OK, I can do the something similar for the import matcher ?

Bob

Bob-IT:
Forgot to mention that I have not seen the errors above but will keep an eye out...
Bob

Bob-IT:
OK, I took a gamble and made similar changes to the import-main-matcher background colours that I did in the dense calendar.
Bob

gnucash/gnome-utils/gnc-dense-cal.c
```diff
@@ -976,16 +976,16 @@ gnc_dense_cal_draw_to_buffer(GncDenseCal *dcal)
...
          GdkRGBA color;
-         gchar *color_extension = NULL;
+         gchar *class_extension = NULL;
 
          gtk_style_context_get_color (stylectxt, GTK_STATE_FLAG_NORMAL, &color);
 
-          if (is_color_dark (&color))
-              color_extension = "-dark";
+          if (is_color_light (&color))
+              class_extension = "-dark";
```
gjanssens: While absolutely correct, this got me confused ad first reading. Can you add a short comment explaining why a light color suggests a dark theme ? (I understand now this is because "color" is the text color, and if that's a light color that means the background is dark -> hence dark theme). We may have forgotten this logic in a few years time again...

And perhaps we should call the function "is_dark_theme" rather than "is_color_light" and then the comment can be inside that function. That would make it more clear to the caller what the purpose of the function is. Devs interested in how this is determined can the read the comment with the function.

gnucash/import-export/import-main-matcher.c
```diff
@@ -497,6 +511,20 @@ gnc_gen_trans_init_view (GNCImportMainMatcher *info,
...
+    gboolean is_light = FALSE;
+
+    // Counting the perceptive luminance - human eye favors green color...
+    double a = (0.299 * color->red + 0.587 * color->green + 0.114 * color->blue);
+
+    if (a > 0.5)
+        is_light = TRUE;
+
+    return is_light;
+}
```
gjanssens: Hmm, writing the same function twice is never a good idea. Can you move it to somewhere both locations can use it ? First suggestion would be gnucash/gnome-utils/gnc-gtk-utils.[ch]

Bob-IT: Yes, just did not want to change another file if this approach was wrong.

Bob-IT:
Will update and push later.

Bob-IT:
Changes made.

gjanssens:
Exellent work!

kubrickfr:
Hi,
I apologise as it's probably not the right place to ask this question, but are you taking bug reports for GTK3 already? If yes, where can I report them?
Regards.

gjanssens:
@kubrickfr The right place to ask would have been the [gnucash mailing lists](https://lists.gnucash.org/mailman/listinfo). I'll answer here anyway :)

Our Gtk3 port isn't fully stable yet. There are a few known issues still. However I don't think these are formally documented, so bug reports would be a good way to do just that. Bug reports should go in [Gnome's bugzilla](https://bugzilla.gnome.org) against the GnuCash product. Gtk3 related issues probably belong under the "User Interface General" component.

PR #173, opened 2017-08-16
Account fullname filter for transaction.scm

christopherlam:
This small change will introduce an account full-name filter. Only accounts whose fullname containing substring will be selected e.g. ":Travel" will match Expenses:Travel:Holiday and Expenses:Business:Travel. This can be left blank, which will mimic previous behavior. This will ease accounts selection tremendously, much easier than scrolling through the tiny account tree box and selecting individual accounts.

### DIFF
--- a/src/report/standard-reports/transaction.scm
+++ b/src/report/standard-reports/transaction.scm
@@ -33,7 +33,6 @@
 
 (use-modules (gnucash main)) ;; FIXME: delete after we finish modularizing.
 (use-modules (srfi srfi-1))
-(use-modules (srfi srfi-13))
 (use-modules (gnucash gnc-module))
 (use-modules (gnucash gettext))
 
@@ -666,7 +665,7 @@
     gnc:pagename-accounts (N_ "Account Substring")
     "a5" (N_ "Match only above accounts whose fullname contains substring e.g. ':Travel' will \
 match Expenses:Travel:Holiday and Expenses:Business:Travel. Can be left blank, which will \
-disable the substring filter. This filter is case-sensitive.")
+disable the substring filter. This filter is case-sensitive. Gnucash 2.8 or later.")
     ""))
 
   (gnc:register-trep-option
@@ -1431,7 +1430,6 @@ Credit Card, and Income accounts.")))))
   (gnc:report-starting reportname)
   (let ((document (gnc:make-html-document))
 	(c_account_1 (opt-val gnc:pagename-accounts "Accounts"))
-	(c_account_substring (opt-val gnc:pagename-accounts "Account Substring"))
 	(c_account_2 (opt-val gnc:pagename-accounts "Filter By..."))
 	(filter-mode (opt-val gnc:pagename-accounts "Filter Type"))
         (begindate (gnc:timepair-start-day-time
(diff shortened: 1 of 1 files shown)


gjanssens:
I just saw you had requested feedback on the gnucash-devel list for this feature. I'm sorry I didn't catch that earlier.

While thinking of this, an even better improvement IMO would be if the full text search you have implemented actually interacted with the Filter Accounts option. I mean if the user enters text in the full text search, the list of accounts available to filter on is reduced to all accounts that match the text entered. That would allow even more flexible filtering (perhaps the user wants to select only some accounts that match "Travel").

I realize this is a more difficult work. I'm mentioning this because I anticipate this will be what users will request next after they will learn your current improvement...

christopherlam:
Hi Geert

You're right that if a prior version loads a saved report generated by a future version, and saves, the filter substring will be lost. The risk is small though, and can be mitigated by documenting in the popup help text. In any case, a prior version loading an enhanced saved report will just show all selected accounts, unfiltered, without complaining.

I wouldn't know how to implement backward-compatibility with the various branches. If this substring filter can be applied to master instead (for 2.8 onwards, documented properly) then it will be absolutely fine. For the maint branch I guess the option will just need to exist without being used at all? I can't figure out this part.

--

I thought about the next suggestion, which is to apply a substring matcher to the "Filter By..." menu -- I guess this can be done, but the underlying API does not make for a good UI. I was thinking of using a selector with a callback function to modify the filterby list but this does not exist. The callback functions only exist for boolean checkbox and for radio buttons. Also this is materially a very different filter which applies to the _Other Account_ for each split originating from the (filtered) account list as above.

In other words, for my filter, the user would select account EXPENSES - Select Children - choose Substring: "Travel" - and end up choosing Expenses:Travel:Personal and Expenses:Business:Travel. Whereas the existing Filter-By will require the choose ASSETS - Select Children - Use existing Filter-By all Travel accounts.

Thoughts?

christopherlam:
Another possibility is to modify the account selector to allow "Select Substring" as an addition to selecting All/Clear/Children/Default however this resides in /gnucash/gnome-utils/dialog-options.c and I'm afraid C is inaccessible to me.

gjanssens:
To solve the backwards compatibility issue, you add the same option to the report in maint, except you make it an invisible option. That's all. No need to act on the option at all. It can't be set from maint because it's invisible so users running 2.6.x will never set the option. Backwards compatibility with all previous versions is guaranteed. If a user runs 2.8 and sets the option then switches to 2.6 (more recent than 2.6.16 obviously) for example on another machine, the 2.6 code will load the option, but won't act on it. When closing  gnucash the option will be properly saved again because it was set. Backwards compatibility between 2.8 and 2.6.16+ established.

As for the other suggestion, we can just leave it as is. I agree this would require C changes and not required to accept this PR.

christopherlam:
Reduced this code to compatibility shim.

gjanssens:
@christopherlam Unfortunately just commenting out the option is doing more than hiding it. As far as the report system is concerned it no longer exists. I meant hiding it from the *user* :(

While you were adding this last commit I was already busy merging your PR. I added one more commit to change the option from a string option to an internal option and pushed the result to my personal repo to run travis on it. You can remove your last commit from this PR so it remains in sync with what I'm about to push.

You can do this by checking out your patch-3 branch and then run
git reset --hard 19fe7d8
git push -f <your remote> patch-3

christopherlam:
Hi Geert, from my testing, commenting out the (gnc:register-trep-option)
but allowing the (gnc:make-string-option) did seem to hide from user but
still allow loading of report from saved-reports.

But yours is the official one, so, go ahead and thank you :)

gjanssens:
Yes, loading would indeed still work. But the option would get lost also when saving the report again. That's what I wanted to avoid.

PR #174, opened 2017-08-23
Link test_import_pending_matches as C++

esiegerman:
This fixes the problem that was discussed in http://gnucash.1415818.n4.nabble.com/building-master-tc4691183.html#a4691186.


gjanssens:
Thank you for your patch. As it happens I had a similar commit ready in my local repository waiting to be pushed to master... I didn't know about the dummy.cpp trick, but instead chose to make the test file itself a C++ file. That is how similar issues have been solved in our repo previously and since gnucash is slowly becoming a C++ application this fits that long term goal even better.

Your contribution is appreciated nonetheless and please continue to send patches :)

PR #175, opened 2017-08-25
Account fullname filter for transaction.scm

christopherlam:
This small change will introduce an account full-name filter. Only accounts whose fullname containing substring will be selected e.g. ":Travel" will match Expenses:Travel:Holiday and Expenses:Business:Travel. This can be left blank, which will mimic previous behavior. This will ease accounts selection tremendously. This filter is case sensitive. For Gnucash 2.8 or later.

### DIFF
--- a/src/report/standard-reports/transaction.scm
+++ b/src/report/standard-reports/transaction.scm
@@ -33,6 +33,7 @@
 
 (use-modules (gnucash main)) ;; FIXME: delete after we finish modularizing.
 (use-modules (srfi srfi-1))
+(use-modules (srfi srfi-13))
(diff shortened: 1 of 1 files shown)


gjanssens:
Thanks for the new PR. As with the previous one I was already handling the merge into master locally. I'm sorry I had you do double work :(

Instead of pulling this PR (which would result in a merge conflict) I have merged maint and reverted the two commits that hid and disabled the option on maint to restore the full functionality.

gjanssens:
But the good news is your work is now available on both branches and hence will appear in 2.6.18 and up. Well done!

PR #176, opened 2017-08-29
Kvpvaltostring

limitedAtonement:
If I understand correctly, this can't be accepted until that unstable branch is created, etc. at which time I'll be happy to rebase this pull request to make sure it's still right.


gjanssens:
I think it can still be added to master. From what I saw on IRC there is still some debugging to do on Windows before we can branch 2.7. So I have just pushed your PR. If I was wrong I'll backport the changes to 2.7. manually.
Thanks!

PR #178, opened 2017-09-01
Add all gncmod modules as dependency of scheme compiles

codesmythe:
This change causes all of the gncmod-*.so files to be built before any of the scheme code is compiled and thus removes the complaints from gnc-module.c about dlopened .so files being too short.

### DIFF
--- a/src/cmake_modules/GncAddSchemeTargets.cmake
+++ b/src/cmake_modules/GncAddSchemeTargets.cmake
@@ -87,6 +87,15 @@ MACRO(GNC_ADD_SCHEME_TARGETS _TARGET _SOURCE_FILES _OUTPUT_DIR_IN _GUILE_MODULES
     ENDIF(__DEBUG)
     SET(_TARGET_FILES "")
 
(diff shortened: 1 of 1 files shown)


codesmythe:
After IRC discussion, I'm going to withdraw this and think about it some more.

limitedAtonement:
We were talking about this recently on IRC; do you have any further thoughts? Why did you withdraw this pull request? I'll try to find a conversation in IRC logs, too.

gjanssens:
@limitedAtonement IIRC I made a reservation on the fact the gnc-mod* objects are all listed together in one spot, the macro that adds scheme files. I considered this to be non-ideal because it's hidden away from the actual rules to build the gnc-mod* files themselves. So adding, renaming or removing a gnc-mod* in the future would require the dev  to remember to update this list. For renaming and removing cmake will be able to warn because of a missing dependency. However forgetting to update the list while adding a new gnc-mod* file would go unnoticed.

codesmythe:
I agree with @gjanssens recollection of the issues, and haven't been able to give this much attention lately.

PR #179, opened 2017-09-01
Enable CMake build in Travis CI

codesmythe:
This PR has the Travis build system do CMake builds (Makefile and Ninja) in addition to the autotools build.

We did not need a PPA for cmake, as Trusty provides a cmake3 package. 

Also, I did not move the script part to a separate file, as I figured out that using 'set -e' will allow us to put one command per line in the 'script:' section and have a failing line fail the whole build.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,29 @@ compiler:
 before_install:
   - sudo apt-get update -qq
   - sudo apt-get build-dep -qq gnucash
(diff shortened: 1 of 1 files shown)


gjanssens:
Wonderful. Your solution is even more elegant that adding a separate script. I have committed a followup to split the build into a matrix of 3 jobs, for performance and ease of use.

PR #181, opened 2017-09-02
Added WITH_OFX fix to CMakeLists in gnucash/import-export/ofx

bhardwajs:
Added condition for WITH_OFX to CMakeList in gnucash/import-export


gjanssens:
As this issue was also present on maint I have rebased your fix there first. Then merged into master.

Thanks!

PR #183, opened 2017-09-03
L10N: DE, automatic "make pot" changes.

quazgar:
Do these changes need to be committed manually?
@Mechtilde

### DIFF
--- a/po/de.po
+++ b/po/de.po
@@ -15,7 +15,7 @@ msgstr ""
 "Project-Id-Version: GnuCash 2.6.15\n"
 "Report-Msgid-Bugs-To: \n"
 "POT-Creation-Date: 2017-03-22 02:17+0100\n"
-"PO-Revision-Date: 2017-03-22 14:22+0100\n"
+"PO-Revision-Date: 2017-09-04 17:10+0200\n"
 "Last-Translator: Mechtilde <ooo@mechtilde.de>\n"
 "Language-Team: GnuCash-de <gnucash-de@gnucash.org>\n"
 "Language: de\n"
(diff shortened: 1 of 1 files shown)


po/de.po
```diff
@@ -7147,10 +7142,10 @@ msgid ""
 "any valid strftime string; for more information about this format, read the "
 "manual page of strftime by \"man 3 strftime\"."
 msgstr ""
-"Wenn das Datumsformat auf Benutzerdefiniert gesetzt ist, kann in dieser Ausdruck"
-"als Argument an das strftime-Format übergegeben werden. Es kann eine beliebige "
-"gültige strftime Zeichenkette sein. Für mehr Informationen "
-"lesen Sie die Ausgabe von »man 3 strftime«."
+"Wenn das Datumsformat auf Benutzerdefiniert gesetzt ist, kann in dieser "
+"Ausdruckals Argument an das strftime-Format übergegeben werden. Es kann eine "
```
gjanssens: This looks like an error to me: I think it should stay "Ausdruck als Argument" and not become "Ausdrukals Argument". But that may very well have been an error in the original translation as well. I think that first line should have ended in a space.

gjanssens:
Thanks!

po/de.po
```diff
@@ -14824,26 +14796,6 @@ msgstr "Nicht vorgemerkt"
 msgid "Select occurrence date above."
 msgstr "Wählen Sie oben ein Datum des Auftretens."
 
-#: ../src/gnome-utils/gtkbuilder/gnc-frequency.glade.h:7
```
fellen: The following 3 removed sections are the result of https://bugzilla.gnome.org/show_bug.cgi?id=759844
Best results are currently obtained by using a separate out of tree build directory.

PR #188, opened 2017-09-06
Adding .gitignore entries for ctags and vim swaps

limitedAtonement:


bhardwajs:
@limitedAtonement I was about to send this PR; you beat me to it. For vim swap files, I will change that to include ".*.sw*". vim swap files go from swp to swn to swo ...

jralls:
I don't think that you want that leading '.' in the vim swap file rule: With it git will ignore only swap files on hidden files, e.g. .gitignore.swp. Also unless vim will use more than one character after the 'sw' you want '?' instead of '*' for the trailing wildcard. With those changes the rule would be `*.sw?'.

bhardwajs:
@jralls AFAIK, vim swap files are hidden files. For example, I just checked on Ubuntu, the swap file for "blah" is ".blah.swp". You are right about "?". So, the pattern to use should be ".*.sw?".

limitedAtonement:
Yes on the leading `.`. My current status says
<pre>lmat@ryle~/source/gnucash$ git st
On branch newtravis
Untracked files:
  (use "git add <file>..." to include in what will be committed)

        .travis.yml.swp
        libgnucash/engine/.Account.cpp.swp
        libgnucash/engine/.qof-string-cache.cpp.swp
        tags

nothing added to commit but untracked files present (use "git add" to track)
lmat@ryle~/source/gnucash$</pre>
I don't care between `.*.sw?` or `.*.sw*`.

gjanssens:
I have applied this to maint first and then merged into master.

Thanks!

PR #189, opened 2017-09-07
Allow to export the last commit's date

bmwiedemann:
for use in release tarballs
to give users a hint on how old their software is.

Partial implementation of an idea discussed in PR #180

### DIFF
--- a/util/gnc-vcs-info
+++ b/util/gnc-vcs-info
@@ -2,9 +2,11 @@
 #
 # Usage:
 #   gnc-vcs-info -r <srcdir>
+#   gnc-vcs-info -d <srcdir>
 #   gnc-vcs-info -t <srcdir>
 #
 # With -r prints the revision number to stdout and exits 0 on success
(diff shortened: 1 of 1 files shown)


gjanssens:
This looks like a good start. Do you intend to add this only for our future 2.8 series or do you also want to achieve reproducible builds for our current stable 2.6 series ? If the latter, please reapply your changes on the maint branch and redo the PR. If you're fine with only getting this in 2.8 the master branch is fine.

bmwiedemann:
I think, doing it on master is sufficient, as this is a long term effort... all other versions will 'soon' be obsolete and forgotten ;-)

gjanssens:
I have merged this PR after a rebase in the unstable branch (soon to become the 2.8 release). I have continued your work and made the necessary changes in the rest of the code to use the commit date in all messages and in the man page.
Hopefully this will get you one step further to your reproducible build goal.

Good luck and thanks !

PR #190, opened 2017-09-07
Configure travis to use docker for CI

limitedAtonement:
Currently, there is an ubuntu 14.04 and archlinux build and they both
use cmake with ninja, cmake with make, and autotools to build. It should
be straightforward to add another configuration.


limitedAtonement:
Oh yeah, I forgot to add some things like after-failure script, etc. I'll push those changes in a bit.

util/ci/arch-testscript
```diff
@@ -0,0 +1,38 @@
...
+
+# Python is python 3, but gnucash doesn't work with python 3. There
+# doesn't seem to be a way to tell automake to use /usr/bin/python2,
+# so we'll mock this up
+mkdir autotools_bin;
+ln -s /usr/bin/python2 autotools_bin/python;
+export PATH=/autotools_bin:"$PATH";
+echo path is "$PATH";
+echo python version is "$(python --version)";
+git clone --depth 1 https://github.com/Gnucash/gnucash.git;
```
codesmythe: With this clone, you won't be using the repository under test (you'll get whatever is currently on master).

util/ci/ubuntu-14.04-testscript
```diff
@@ -0,0 +1,28 @@
...
+
+git clone --depth 1 https://github.com/Gnucash/gnucash.git;
+mkdir build;
+cd build;
+export TZ="America/Los_Angeles";
+ls -al /;
+ls -al /gtest;
+mkdir -p /root/.local/share
+export GTEST_ROOT=/gtest/googletest GMOCK_ROOT=/gtest/googlemock;
+if [[ "$BUILDTYPE" == "cmake-make" ]]; then
```
codesmythe: What do you think about putting this 'if buildtype...' block in a separate file and calling it from the various distro test scripts?

gjanssens: That will of course only work if the commands are exactly the same for all distros, but it's worth investigating. For instance, on arch, googletest is installed on the system path while on Ubuntu 14.04 we install it ourselves locally and hence have to pass GTEST_ROOT/GMOCK_ROOT.

util/ci/arch-testscript
limitedAtonement: Yikes good catch.

limitedAtonement: For what it's worth, I don't think it would always be checking `master`, but it would always be checking whatever is `HEAD` in github. Not quite as bad, but still quite bad. I have a fix committed now, although I'll close this pull request and open a new one to clean things up.

PR #191, opened 2017-09-08
Run the python bindings test when using CMake

codesmythe:


gjanssens:
Pushed after a rebase. Thanks!

gjanssens:
Note the python bindings test fails after merging it into master. I have changed the python_path but it still fails due to missing unittest support. I don't have time right now to debug this further.

PR #192, opened 2017-09-08
Configure travis to use docker for CI

limitedAtonement:
Currently, there is an ubuntu 14.04 and archlinux build and they both
use cmake with ninja, cmake with make, and autotools to build. It should
be straightforward to add another configuration.

Travis checks out the correct branch of the source code for us, and we
expose that source code to the docker container using a docker run
"volume".


util/ci/ubuntu-14.04-testscript
```diff
@@ -0,0 +1,16 @@
...
+#  exit immediately.
+
+mkdir -p /root/.local/share
+
+mkdir build;
+cd build;
+export TZ="America/Los_Angeles";
+
+ls -al /;
+ls -al /gtest;
```
gjanssens: Considering you didn't add these lines to the arch testscript I assume the two lines were only necessary to debug your docker setup ?

.travis.yml
```diff
@@ -1,26 +1,22 @@
...
-  - git clone https://github.com/google/googletest -b release-1.8.0 ~/gtest
-script: ./util/travis-script.sh
-after_failure:  ./util/travis-after-failure.sh
+    - sudo docker --version
+    - echo BUILDENV="$BUILDENV"
+    - sudo docker build -f util/ci/${BUILDENV}-docker -t ${BUILDENV}-gnucashbuild util/ci
+script:
+    - echo BUILDTYPE="$BUILDTYPE"
+    - pwd
+    - ls
```
gjanssens: Is this useful output or was it added to debug while creating these docker based tests ?

limitedAtonement:
Eek! I'm appalled that I committed .swp files; my apologies. It looks like the gitignore change was applied to maint. I've been creating pull requests against master. Let me know if I need to change anything. For what it's worth, tracked files aren't affected at all by `.gitignore`, so they'll need to be removed explicitly either way.

.travis.yml
```diff
@@ -1,26 +1,22 @@
...
-  - sudo apt-get install -qq swig xsltproc libdbd-sqlite3 cmake3 texinfo ninja-build
-  - sudo apt-get install -qq libboost-all-dev libgtk-3-dev libwebkit2gtk-3.0-dev
-  - sudo apt-get --reinstall install -qq language-pack-en language-pack-fr
-  - git clone https://github.com/google/googletest -b release-1.8.0 ~/gtest
-script: ./util/travis-script.sh
-after_failure:  ./util/travis-after-failure.sh
+    - sudo docker --version
+    - echo BUILDENV="$BUILDENV"
+    - sudo docker build -f util/ci/${BUILDENV}-docker -t ${BUILDENV}-gnucashbuild util/ci
+script:
```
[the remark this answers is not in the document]
limitedAtonement: When I run docker, I have to run `sudo` on everything in order to communicate with the root-run daemon. I see that I don't need to use `sudo`; that means `sudo: required` means "run everything as root"? Thanks for pointing it out.

.travis.yml
limitedAtonement: I'll just remove that.

util/ci/ubuntu-14.04-testscript
limitedAtonement: It's related to the way we get google test differently in ubuntu than arch linux. I'll remove this, too.

limitedAtonement:
Oh, I see the merge to master; I'll rebase accordingly.

gjanssens:
Thanks!

PR #193, opened 2017-09-08
Fix problems that occur when running out of CMake build dir

codesmythe:
Back in July, @gjanssens reported some issues he ran into when running GnuCash out of the CMake build directory. This PR fixes the two issues mentioned: locale settings not honored and segfaults when finding vendors.


gjanssens:
I can confirm both issues are fixed after applying this PR.

Pushed after a rebase. Thanks!

PR #195, opened 2017-09-09
Fix swig wrapper for gncTaxTableGetTaxTables

christopherlam:
Bug: (gncTaxTableGetTables (gnc-get-current-book)) currently returns a #\<swig-pointer GncInvoiceList\>
Fix: It should return a list of #\<swig-pointer GncTaxTable\> entries.

This patches gncTaxTableGetTables to return a GncTaxTableList, according to formula in old commit https://github.com/Gnucash/gnucash/commit/8221aada49ec01ad91f6d3a69aefd473ef6608f1 from https://bugzilla.gnome.org/show_bug.cgi?id=573645

I don't have a dev environment and cannot really test above. But I'll still document this (partial?) fix.


gjanssens:
Aplied after squashing them together in one commit and rebasing on current maint.

Thanks!

PR #196, opened 2017-09-09
Toggle "Account Name" should effect "Use Full Account Name" in Options.

christopherlam:
From https://bugzilla.gnome.org/show_bug.cgi?id=776263

Currently, setting "Use Account Name" / "Other Account Name" to #f would disable the Account column, rendering the "Use Full Account Name" option useless. This setting ensures the user understands that toggling (Other) Account Name effects the next option.

No effect is anticipated upon backwards compatibility, saved-reports, or report content. This purely enhances UI of report options.


christopherlam:
Will it be useful that I patch all relevant .scm in this PR?

christopherlam:
I now realise there are changes in master. I'll rebase these changes onto master.

PR #197, opened 2017-09-09
Made ofx-gschema dependent on WITH_OFX flag

bhardwajs:
Change cmake config to move ofx-gschemas out of main list and check for WITH_OFX flag.

### DIFF
--- a/cmake/CMakeLists.txt
+++ b/cmake/CMakeLists.txt
@@ -16,11 +16,15 @@ IF (${CMAKE_VERSION} VERSION_GREATER 3.1)
 ENDIF()
 
 SET(schema-targets csv-exp-gschema csv-imp-gschema
-    generic-import-gschema gnome-gschema gnome-utils-gschema ofx-gschema qif-imp-gschema)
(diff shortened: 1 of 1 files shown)


bhardwajs:
Somehow, I didn't get a notification yesterday that Travis build failed. Taking a look now.

Surprised that Travis fails for autotools when the PR doesn't touch any of autotools part.

jralls:
It was test-import-map which fails intermittently for no particular reason. Don't worry about it.

gjanssens:
Thanks!
Note your changes got me thinking about this business of conditionally excluding some directories from the build path and I believe this has negative side effects on the generation of our distribution tarballs. I've opened [a new bug](https://bugzilla.gnome.org/show_bug.cgi?id=787497) for this so we can discuss this. It's not an issue you have introduced. We already did this wrong for aqbanking an the python-bindings.

bhardwajs:
Thanks @jralls .
@gjanssens I did check today to confirm that WITH_OFX and WITH_AQBANKING are handled consistently. Let me look at the bug that you created. Thanks!

PR #198, opened 2017-09-11
Adding python tests to cmake

limitedAtonement:
And removing some debugging comments from travis docker files.


gjanssens:

PR #199, opened 2017-09-11
Tweak "Display", overhaul "Sorting" tab to improve UI

christopherlam:
Usability fix. This should help newcomers. Two changes:

1. Toggling Display/Account Name will also enable/disable Display/Full Account Name
2. Overhaul of the Sorting tab only presents *valid* options. e.g. 

 - Changing Primary or Secondary Sort-key to 'none will disable its descendant options. 
 - Changing to 'date or 'register-order will enable Date-interval choice. 
 - Changing to 'account-name or 'account-code will enable subtotal checkbox
 - Changing subtotal checkbox to #t will allow relevant account-name display options
 - Changed order for better flow of options

This has no backward/forward compatibility issues

### DIFF
--- a/gnucash/report/standard-reports/transaction.scm
+++ b/gnucash/report/standard-reports/transaction.scm
@@ -51,9 +51,13 @@
 (define pagename-sorting (N_ "Sorting"))
 (define optname-prime-sortkey (N_ "Primary Key"))
 (define optname-prime-subtotal (N_ "Primary Subtotal"))
+(define optname-prime-sortorder (N_ "Primary Sort Order"))
 (define optname-prime-date-subtotal (N_ "Primary Subtotal for Date Key"))
+(define optname-full-account-name (N_ "Show Full Account Name"))
+(define optname-show-account-code (N_ "Show Account Code"))
 (define optname-sec-sortkey (N_ "Secondary Key"))
 (define optname-sec-subtotal (N_ "Secondary Subtotal"))
+(define optname-sec-sortorder  (N_ "Secondary Sort Order"))
 (define optname-sec-date-subtotal (N_ "Secondary Subtotal for Date Key"))
 (define optname-void-transactions (N_ "Void Transactions"))
 (define optname-table-export (N_ "Table for Exporting"))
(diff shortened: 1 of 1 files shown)


gnucash/report/standard-reports/transaction.scm
```diff
@@ -972,11 +1022,35 @@ disable the substring filter. This filter is case-sensitive.")
...
+                 gnc:*transaction-report-options*
+                 gnc:pagename-display
+                 (N_ "Use Full Account Name")
+                 x))))
+  
+  ;; Ditto for Other Account Name #t -> Use Full Other Account Name is selectable
+  (gnc:register-trep-option
+   (gnc:make-complex-boolean-option
+    gnc:pagename-display (N_ "Other Account Name")
+    "h"  (N_ "Display the other account name? (if this is a split transaction, this parameter is guessed).") #f
```
gjanssens: "h" is already in use as position indicator for the "Amount of detail to display per transaction." option. On my system this results in a wrong ordering of these two options. You can fix this by choosing "hx" for the "Other Account Name" option.

christopherlam:
> * When the primary sort key is not set to an account related value, the two checkboxes at the bottom are disabled. However they remain disabled if the second sort key is set to an account related value. Wouldn't you want the same flexibility for both options ?

The (primary or secondary) sortkey must be set to an account related value _and_ its sortkey subtotal must be enabled, and this will allow the bottom two checkboxes.

>  * When opening the Report Options dialog for newly opened transaction report (so with default options set) the Display tab will show "Account Name" and "Use Full Account Name" as ticked. That's good. However "Other Account Name" is not ticked and still "Use Other Full Account Name" is enabled (and ticked).

I think this is caused by prematurely running (apply-selectable-by-name-options) upon report-load within (trep-options-generator). I've removed it, and I think it works well now. In any case "Display/Use Other Full Account Name" is now #f by default.

>  * Closing and Reopening the Report Options dialog will spew lots of errors like this one:
>    - 19:07:32 CRIT <gnc.app-utils> gnc_option_set_selectable: assertion 'option->odb->set_selectable != NULL' failed

As above, I think this is caused by prematurely running (apply-selectable-by-name-options) upon report-load within (trep-options-generator).

christopherlam:
I'd think we could disable sec-sortkey if prime-sortkey is 'none but this would probably introduce incompatibility with existing reports.

gjanssens:
On dinsdag 12 september 2017 01:01:23 CEST christopherlam wrote:
> > * When the primary sort key is not set to an account related value, the
> > two checkboxes at the bottom are disabled. However they remain disabled
> > if the second sort key is set to an account related value. Wouldn't you
> > want the same flexibility for both options ?
> The (primary or secondary) sortkey must be set to an account related value
> _and_ its sortkey subtotal must be enabled, and this will allow the bottom
> two checkboxes.

Oh indeed... I missed that detail in my previous testing.

> >  * When opening the Report Options dialog for newly opened transaction
> >  report (so with default options set) the Display tab will show "Account
> >  Name" and "Use Full Account Name" as ticked. That's good. However "Other
> >  Account Name" is not ticked and still "Use Other Full Account Name" is
> >  enabled (and ticked).
> I think this is caused by prematurely running
> (apply-selectable-by-name-options) upon report-load within
> (trep-options-generator). I've removed it, and I think it works well now.
> In any case "Display/Use Other Full Account Name" is now #f by default.

It behaves better now.

> >  * Closing and Reopening the Report Options dialog will spew lots of 
errors like this one:
> >    - 19:07:32 CRIT <gnc.app-utils> gnc_option_set_selectable: assertion
> >    'option->odb->set_selectable != NULL' failed
> As above, I think this is caused by prematurely running
> (apply-selectable-by-name-options) upon report-load within
> (trep-options-generator).

It still happens on my system each time I open the Report Options window 
except for the first time it's opened on a report.

christopherlam:
>>>  * Closing and Reopening the Report Options dialog will spew lots of errors like this one:
>>>    - 19:07:32 CRIT <gnc.app-utils> gnc_option_set_selectable: assertion   'option->odb->set_selectable != NULL' failed

>> As above, I think this is caused by prematurely running (apply-selectable-by-name-options) upon report-load within (trep-options-generator).

>It still happens on my system each time I open the Report Options window
>except for the first time it's opened on a report.

This is the unusual one. I'm sure this peculiarity has existed for a long time: occasionally cannot change options; e.g. multichoice options such as the prime-sortkey suddenly shows no entries (only a single horizontal line), clicking 'Display/Memo' does not toggle 'Display/Notes', etc. GTK getting confused.

I've just reverted to current master/transaction.scm and I have still managed to trigger this error. Usually the best option is to close down the tab and start again. I cannot recall other .scm reports - I haven't experimented too much with them.

I'll look further but my suspicion is somewhere within C being too fast, and it cannot find the selectable before the window is open. But perhaps I'll try small changes. So far I can't find a functional difference between master and the latest .scm.

gjanssens:
@christopherlam With your last commits the options that should be enabled conditionally are now always enabled. They never dim on my system. Instead each time I alter an option that should trigger enabling/disabling another option I get the same error as I pointed out before.
It looks like this is a more severe case of [the bug](https://bugzilla.gnome.org/show_bug.cgi?id=647805), for which I unfortunately don't have a solution yet. I see the same behaviour that the pointer to the GncOptionsDatabase changes between the time the report options dialog is opened and the time the options should be enabled/disabled. Contrary to what the bug reports, it doesn't appear to resolve itself after a few tries.
Do you see this behaviour as well ?

gjanssens:
Oh and can you please pay attention to your commit messages ? "." is not a very useful one. Please improve it (using git rebase -i).

christopherlam:
> Do you see this behaviour as well ?

In my Win10 2.6.17 this bug can usually be sidestepped by closing and opening report options until it works. If not, I can close tab and start again. Perhaps this bug is more severe because I'm abstracting and introducing small delays. We can always undo abstraction, which may work but will be uglier. Time to go work!

christopherlam:
> Oh and can you please pay attention to your commit messages ? "." is not a very useful one. Please improve it (using git rebase -i).

No worries. Done. I was actually thinking that this type of PR would normally best be squashed and merged rather than recording works in progress. After all I'm not a professional programmer and wouldn't want to record my numerous learning mistakes in the official tree.

gjanssens:
Ok, I've got the "set_selectable != NULL" issue sorted. I tried to apply your PR again and now it works as advertised.
I propose you do a final rebase to clean up and then it can be pushed to our main repository.

christopherlam:
\m/ I've survived my first rebase \m/

christopherlam:
Squash-and-merge would be better?

christopherlam:
Well looks like my rebase wasn't quite right.
How to fix this?

jralls:
Back it out like I explained to you on IRC yesterday and try again. You might read the [Pro Git section on rebasing](https://git-scm.com/book/en/v2/Git-Branching-Rebasing) through a couple of times first. You can save yourself a bit of angst by checking the results of a rebase with ```git log --oneline``` before pushing anything.

gjanssens:
It looks like a git PR remembers its original base commit and hence will show all commits starting from that commit also after a rebase. If I pull this PR, it applies correctly so you can ignore the commits you didn't do yourself.

And yes, I meant for you to "squash and merge" as you call it yourself. In my mind this cleanup could be done using "git rebase -i", which is why I said "rebase to clean up". Sorry about the confusion.

gjanssens:
Oh, and while you're cleaning up, please remove the trailing whitespace (and whitespace on empty lines) that snuck in. The gitk tool unfortunately doesn't show those, but 'git diff <base-commit>' on the command line does. <base-commit> should be replaced with a reference to the commit on which your branch is based. After a rebase this can be "master" or "origin/master".

gjanssens:
Thanks and congrats 👍

PR #201, opened 2017-09-14
Add maint compatibility for transaction-type-filter

christopherlam:
After https://github.com/Gnucash/gnucash/pull/199/


christopherlam:
I'm aware there was lack of discussion but I think this would be useful for some.

christopherlam:
not very useful after all...

PR #205, opened 2017-09-20
Catalan accounts missing

jpftgn:
Catalan accounts missing

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -504,6 +504,8 @@ SET(gnucash_DOCS
     ChangeLog.2012
     ChangeLog.2013
     ChangeLog.2014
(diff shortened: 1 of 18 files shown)


gjanssens:
Thanks for your PR. You made it against master, which means it would only appear in gnucash 2.8 or later. Is that your intention or would you like these account charts to be included in the next 2.6 version already ?

@fellen Other than the choice of base branch this PR seems ok to me. As our internationalization epxert, do you have additional comments ?

jpftgn:
I would prefer them to appear in next 2.6 version, shall I do a new PR? Or will you be able to do it? Thanks for confirming

gjanssens:
A new PR would be best. Thanks!

PR #206, opened 2017-09-21
Add missing export (gnc:date-get-week)

christopherlam:
Hit this bug during development. meanwhile I'm using (gnc:timepair-get-week) instead


gjanssens:

PR #207, opened 2017-09-21
Catalan accounts missing

jpftgn:
This is the new PR. Further information: https://github.com/Gnucash/gnucash/pull/205

### DIFF
--- a/accounts/ca/CMakeLists.txt
+++ b/accounts/ca/CMakeLists.txt
@@ -1,0 +1,21 @@
+SET(account_DATA
+  acctchrt_brokerage.gnucash-xea
+  acctchrt_carloan.gnucash-xea
+  acctchrt_cdmoneymkt.gnucash-xea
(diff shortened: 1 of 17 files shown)


gjanssens:
Wonderful. I have applied your PR after a rebase, together with a followup commit to get the Catalan account charts installed and distributed. They will appear in the next stable release (due this weekend).

Thanks!

fellen:
At first thanks to your work, but I have a few notes:

1. I would have choosen a nicer commit message like "New: Catalan account templates".

2. xmllint fails, because you did not add the namespace section xmlns... as shown in https://wiki.gnucash.org/wiki/Account_Hierarchy_Template.
Yes, I know, they are missing in all templates which where not touched in the last ten years, but at some point we should require xmllint in the accounts subtree as part of make check.

3. Your users would like it, if the headers of account groups were already marked as placeholder. See, how it is done in the script in above wiki page.

PR #210, opened 2017-10-02
Move CSS files to Gresource file

Bob-IT:
I think this covers all that was proposed.
Bob

### DIFF
--- a/gnucash/CMakeLists.txt
+++ b/gnucash/CMakeLists.txt
@@ -27,6 +27,7 @@ ENDIF (WIN32)
 
 SET (gnucash_SOURCES
   gnucash-bin.c
+  gnucash-gresources.c
   ${GNUCASH_RESOURCE_FILE}
 )
 
@@ -42,6 +43,40 @@ TARGET_LINK_LIBRARIES (gnucash
    ${GUILE_LDFLAGS} ${GLIB2_LDFLAGS} ${GTK3_LDFLAGS} ${GTK_MAC_LDFLAGS}
 )
 
+# Get glib executable for generating the gresource file
+EXECUTE_PROCESS(
+   COMMAND
+       ${PKG_CONFIG_EXECUTABLE} gio-2.0 --variable glib_compile_resources
+   OUTPUT_VARIABLE
+       GLIB_COMPILE_RESOURCES_EXECUTABLE
+   OUTPUT_STRIP_TRAILING_WHITESPACE
+)
+
+# Get the dependencies of the gresource
+EXECUTE_PROCESS(
+  OUTPUT_VARIABLE
+      gr_files
+  COMMAND ${GLIB_COMPILE_RESOURCES_EXECUTABLE}
+     --sourcedir=${CMAKE_CURRENT_SOURCE_DIR}
+     --generate-dependencies
+     ${CMAKE_CURRENT_SOURCE_DIR}/gnucash-gresources.xml
+)
+
+STRING (REPLACE "\n" ";" gresource_files ${gr_files})
+
+ADD_CUSTOM_COMMAND(
+   OUTPUT gnucash-gresources.c
+   COMMAND
+    ${GLIB_COMPILE_RESOURCES_EXECUTABLE}
+    --target=gnucash-gresources.c
+    --sourcedir=${CMAKE_CURRENT_SOURCE_DIR}
+    --generate-source
+    ${CMAKE_CURRENT_SOURCE_DIR}/gnucash-gresources.xml
+   DEPENDS
+    ${CMAKE_CURRENT_SOURCE_DIR}/gnucash-gresources.xml ${gresource_files}
+   WORKING_DIRECTORY
+    ${CMAKE_CURRENT_BINARY_DIR}
+)
 
 IF (MAC_INTEGRATION)
   TARGET_COMPILE_OPTIONS(gnucash PRIVATE ${OSX_EXTRA_COMPILE_FLAGS})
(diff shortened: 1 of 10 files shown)


gnucash/CMakeLists.txt
```diff
@@ -42,6 +43,46 @@ TARGET_LINK_LIBRARIES (gnucash
...
+ENDFOREACH ()
+
+ADD_CUSTOM_COMMAND(
+   OUTPUT ${CMAKE_CURRENT_BINARY_DIR}/gnucash-gresources.c
+   COMMAND
+    ${GLIB_COMPILE_RESOURCES_EXECUTABLE}
+    --target=${CMAKE_CURRENT_BINARY_DIR}/gnucash-gresources.c
+    --sourcedir=${CMAKE_CURRENT_SOURCE_DIR}
+    --generate-source
+    gnucash-gresources.xml
```
gjanssens: It would make slightly more sense to me to run this command with WORKING DIRECTORY set to ${CMAKE_CURRENT_BINARY_DIR} (file generation should generally happen there as much as possible).
You can then probably omit most ${CMAKE_CURRENT_BINARY_DIR}/ prefixes (including in the OUTPUT line) and need to add a ${CMAKE_CURRENT_SOURCE_DIR}/ prefix to gnucash-gresources.xml.

gnucash/CMakeLists.txt
```diff
@@ -42,6 +43,46 @@ TARGET_LINK_LIBRARIES (gnucash
...
+   DEPENDS
+    gnucash-gresources.xml ${gnucash_GRESOURCE}
```
gjanssens: Same here: you may have to specify this as ${CMAKE_CURRENT_SOURCE_DIR}/gnucash-gresources.xml

gnucash/CMakeLists.txt
```diff
@@ -42,6 +43,46 @@ TARGET_LINK_LIBRARIES (gnucash
...
+   WORKING_DIRECTORY
+    ${CMAKE_CURRENT_SOURCE_DIR}
```
gjanssens: And this becomes ${CMAKE_CURRENT_BINARY_DIR} then or can probably even be omitted.

Bob-IT:
Made the changes requested.
Bob

gnucash/CMakeLists.txt
```diff
@@ -44,6 +45,46 @@ TARGET_LINK_LIBRARIES (gnucash
...
+  WORKING_DIRECTORY
+     ${CMAKE_CURRENT_SOURCE_DIR}
+  COMMAND ${GLIB_COMPILE_RESOURCES_EXECUTABLE}
+     --generate-dependencies
+     ${CMAKE_CURRENT_SOURCE_DIR}/gnucash-gresources.xml
+)
+
+STRING (REPLACE "\n" ";" gresource_files ${gr_files})
+
+SET (DEPENDS "")
```
gjanssens: What does this do ? It seems to work without clearing DEPENDS as well for me ?

gnucash/CMakeLists.txt
```diff
@@ -44,6 +45,46 @@ TARGET_LINK_LIBRARIES (gnucash
...
+FOREACH (cur_file ${gresource_files})
+    LIST (APPEND gnucash_GRESOURCE "${cur_file}")
+ENDFOREACH ()
```
gjanssens: I don't think this is necessary. Any string with semicolons in it is considered a list by cmake. So you can remove this part work with gresource_files directly everywhere below instead of gnucash_RESOURCE.

gnucash/CMakeLists.txt
```diff
@@ -44,6 +45,46 @@ TARGET_LINK_LIBRARIES (gnucash
...
+       GLIB_COMPILE_RESOURCES_EXECUTABLE
+   OUTPUT_STRIP_TRAILING_WHITESPACE
+)
+
+# Get the dependencies of the gresource
+EXECUTE_PROCESS(
+  OUTPUT_VARIABLE
+      gr_files
+  WORKING_DIRECTORY
+     ${CMAKE_CURRENT_SOURCE_DIR}
```
gjanssens: No need to set a working directory for this process. You work with full pathnames in the command below.

Bob-IT:
OK I have made the changes required but had to put the working directory change back as the cmake travis ubuntu builds failed with out it.
Bob

gnucash/CMakeLists.txt
Bob-IT: Needed to leave this as the travis cmake ubuntu builds would fail

gjanssens: Nope. The better solution would be to set --sourcedir to the command below rather than a WORKING_DIRECTORY. But you're right my suggestion did not work on ubuntu 14.04 either. Adding --sourcedir does work though.

gjanssens:
Thanks!

A sidenote: when you force-push changes, take the opportunity to rebase on the most recent branch head (unstable,master or maint). It helps reduce merge conflicts and superfluous merge commits in the final tree.

Bob-IT:
I thought my unstable was up to date but will check next time, I was going to ask why was using --source better that WORKING_DIRECTORY ?

Regards,
Bob

gjanssens:
For me it's a matter of principle and good coding practise. When building out of tree, the build scripts should stay out of the source directories as much as possible.
Setting WORKING_DIRECTORY does cause glib_compile_resources to do the equivalent of "cd SOURCEDIR", effectively switching to that directory. If the command creates intermediate files it would do so in that directory in this situation. I don't know of glib_compile_resources does. If so it properly cleans up when done. But regardless, it's better to not risk this. On the other hand specifying '--source' will not cause the command to change working directory, it only will use that path to search for files as mentioned in the xml file.

PR #212, opened 2017-10-10
Item edit cell padding and code reduction of cell colours

Bob-IT:
The first lot of commits get the cell margin, border and padding from the CSS values. Previously the border would take up 1px when not specified and the padding default was 2px. This is still the case but with margin so there should be 3px top and bottom which should be enough space for the double line so I have moved it back up and does not interfere with the cell. Also the cell alignment is now a common set of functions which fixes a header alignment issue. I have cast a new toggle button so that the button width can be controlled which leads in to commit 'Change the look...', with these changes the button is removed and replaced with the arrow and possible frame, this could be adjusted if it was a way to go. The latter commits were the code reduction for the cell colours.
Bob

To test here is some CSS...

/* Register sheet font setting */
sheet {
  font: 18px arial, sans-serif;
}

sheet calendar {
  font: 12px arial, sans-serif;
}

/* Register Cursor settings, top, right, bottom, left */
cursor entry {
  margin: 10px 5px 10px 5px;
  padding: 2px 2px 2px 2px;
  border-width: 2px 2px 2px 2px;
  border-color: blue red blue green;
  background-color: #4dd2ff;
}

cursor button {
  margin: 1px 1px 1px 1px;
}

/* Test */
cursor toggle-button {
  margin: 1px 1px 1px 1px;
  background-color: pink;
}

### DIFF
--- a/gnucash/gnucash-310.css
+++ b/gnucash/gnucash-310.css
@@ -27,7 +27,7 @@
   padding: 2px 2px 2px 2px; /* all work with different values, around the text blue area */
 }
 
-.cursor .button {
+.cursor .toggle-button {
   margin: 1px 1px 1px 1px; /* does not work, not used, here for completeness */
 }
 
--- a/gnucash/gnucash-310.css
+++ b/gnucash/gnucash-310.css
@@ -21,13 +21,14 @@
   color: mix (currentColor, grey, 0.2);
 }
 
-/* Register Cursor padding settings, make sure entry matches sheet.h */
-cursor entry {
-  padding: 2px 5px 2px 5px;
+/* Register Cursor settings, top, right, bottom, left */
+.cursor .entry {
+  margin: 2px 5px 2px 5px;  /* this only works by doing it in code, yellow area */
+  padding: 2px 2px 2px 2px; /* all work with different values, around the text blue area */
 }
 
-cursor button {
-  padding: 1px 1px 1px 1px;
+.cursor .button {
+  margin: 1px 1px 1px 1px; /* does not work, not used, here for completeness */
 }
 
 /* Register defaults */
@@ -37,7 +38,7 @@ cursor button {
 @define-color register_split_bg_color #EDE7D3;
 @define-color register_cursor_bg_color #FFEF98;
 
-.register-header {
+*.register-header {
   background-color: @register_header_bg_color;
 }
 
--- a/gnucash/register/ledger-core/split-register-model.c
+++ b/gnucash/register/ledger-core/split-register-model.c
@@ -514,110 +514,20 @@ get_trans_total_balance (SplitRegister *reg, Transaction *trans)
     return xaccTransGetAccountBalance(trans, account);
 }
 
-static guint32
-gnc_split_register_get_color_internal (VirtualLocation virt_loc,
-                                       SplitRegister *reg,
-                                       gboolean foreground)
+static gboolean
+gnc_split_register_use_negative_color (VirtualLocation virt_loc,
+                                       SplitRegister *reg)
 {
-    const char *cursor_name;
-    VirtualCell *vcell;
-    gboolean is_current;
-    gboolean double_alternate_virt;
-    guint32 colorbase = 0; /* By default return background colors */
-
-    if (foreground)
-        colorbase = COLOR_UNKNOWN_FG; /* a bit of enum arithmetic */
-
-    if (!reg)
-        return (colorbase + COLOR_UNKNOWN_BG);
-
-    if (gnc_table_virtual_location_in_header (reg->table, virt_loc))
-        return (colorbase + COLOR_HEADER_BG);
-
-    vcell = gnc_table_get_virtual_cell (reg->table, virt_loc.vcell_loc);
-    if (!vcell || !vcell->cellblock)
-        return (colorbase + COLOR_UNKNOWN_BG);
-
-    if ((virt_loc.phys_col_offset < vcell->cellblock->start_col) ||
-            (virt_loc.phys_col_offset > vcell->cellblock->stop_col))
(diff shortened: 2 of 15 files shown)


gnucash/register/register-gnome/gnucash-sheet-private.c
```diff
@@ -418,7 +418,10 @@ draw_cell (GnucashSheet *sheet,
     gtk_render_background (stylectxt, cr, x, y, width, height);
 
     gdk_rgba_parse (&color, "black");
-    gnc_style_context_get_background_color (stylectxt, GTK_STATE_FLAG_NORMAL, &color);
+    if (gtk_style_context_get_state (stylectxt) == GTK_STATE_FLAG_INSENSITIVE)
+        gnc_style_context_get_background_color (stylectxt, GTK_STATE_FLAG_INSENSITIVE, &color);
+    else
+        gnc_style_context_get_background_color (stylectxt, GTK_STATE_FLAG_NORMAL, &color);
```
gjanssens: Ah, another case of unnecessary command doubling which can be avoided like this:
```
gnc_style_context_get_background_color (stylectxt,
                      gtk_style_context_get_state (stylectxt),
                      &color);
```

gnucash/register/register-core/table-model.h
```diff
@@ -189,25 +185,14 @@ TableGetCellIOFlagsHandler gnc_table_model_get_io_flags_handler
...
- TableGetBGColorHandler io_flags_handler,
+ TableGetCellColorHandler io_flags_handler,
  const char * cell_name);
-void gnc_table_model_set_default_bg_color_handler
+ void gnc_table_model_set_default_cell_color_handler
 (TableModel *model,
- TableGetBGColorHandler io_flags_handler);
-TableGetBGColorHandler gnc_table_model_get_bg_color_handler
+ TableGetCellColorHandler io_flags_handler);
+TableGetCellColorHandler gnc_table_model_get_cell_color_handler
```
gjanssens: This commit will break the build until the next commit, because you replace the fg/bg function definitions with a single cell variant, but the rest of the code is still calling the fg/bg functions.

You should try to avoid this as it makes bisecting the git history more difficult. git bisect is a crucial tool to figure out in which commit a bug first appears, but if there are commits with build errors the bisect can't be used accurately.

If you really have a good reason to deliberately insert a commit that breaks the build, this should be explained in the commit message. In this case I don't see a good reason. This and the next commit are functionally part of the same change, or at least the parts in the next commit where you replace the fg/bg functions. I'm not sure if you can or want to keep the ENUM reduction separately.

Other than that, I'm very much in favor of this simplification :)

gnucash/register/ledger-core/split-register-model.c
```diff
@@ -570,101 +622,40 @@ gnc_split_register_get_color_internal (VirtualLocation virt_loc,
...
+            return (colorbase + COLOR_SPLIT_ACTIVE);
 
-        return (colorbase + COLOR_SPLIT_BG);
+        return (colorbase + COLOR_SPLIT);
     }
 
     PWARN ("Unexpected cursor: %s\n", cursor_name);
 
-    return (colorbase + COLOR_UNKNOWN_BG);
+    return (colorbase + COLOR_UNDEFINED);
```
gjanssens: Same here. This commit should be squashed together with the previous one, because the previous one won't compile without the changes you make here as I understand it.

gnucash/register/register-gnome/gnucash-header.c
```diff
@@ -88,8 +88,8 @@ gnc_header_draw_offscreen (GncHeader *header)
 
     gtk_style_context_save (stylectxt);
 
-    // Get the background color type and apply the css class
-    color_type = gnc_table_get_bg_color (table, virt_loc, NULL);
+    // Get the color type and apply the css class
+    color_type = gnc_table_get_color (table, virt_loc, NULL);
```
gjanssens: And for this commit as well... The previous one wouldn't compile without these changes.

gnucash/gnucash-310.css
```diff
@@ -27,7 +27,7 @@
   padding: 2px 2px 2px 2px; /* all work with different values, around the text blue area */
 }
 
-.cursor .button {
+.cursor .toggle-button {
```
gjanssens: I'd prefer not to do this. Granted the new version is lighter. On the flip side though it's also no longer similar to other arrow buttons used in for example a combobox. So it is slightly alienating the register in a gnome widget context.

I would propose to try and use more vertical space per line so the toggle button gets slightly more room. This will make the register less compact, but it's no longer a pixel perfect match with the old register anyway.

This may also solve a few other glitches which I mention here (but are due to other code sections):
- If text in the cell cursor as glyphs that drop below the base line (like 'g', 'j',...) these glyphs glue against the double line. This is probably because you moved this double line inside the cell again. And perhaps the theme has something to do with it as well.
- Similarly if I click in a cell having such glyphs, the lower line of the glyph is gone (overwritten with yellow somewhere).
- The bottom padding/margin is not correct when using  a huge font. There also the bottom part of similar glyphs is gone. Note the current situation is already a big improvement compared to before your changes. Before the lines wouldn't even scale at all and I saw as much as half of the text being cut off vertically.

Bob-IT:
I see your comments in the git email, did you mean to commit this PR as it will need some changes to get rid of the alternate toggle arrangement. I will look at the alignment again as I thought it looked right with lowercase letters 'gpjq'. I will create a new PR to change the toggle tomorrow if a new version is being released at the weekend.
Bob

gjanssens:
Oh, crap. No it was not my intention to merge this one :(

I have reverted my merge commit. Feel free to reuse your branch for the necessary refactorings.

gjanssens:
To avoid confusion, you probably should rebase and create a new PR when done.

PR #214, opened 2017-10-11
report/net-barchart: create table using gnc-monetary values

jmarino:
The code calculates the income/expense/net values as lists of doubles
that can be easily added to the chart. However, when these values are
added to the table they are displayed as plain numbers instead of being
formatted as monetary values. For example, a value of 12.34 dollars is
displayed as "12.34" instead of "$12.34".

This commit converts the doubles to gnc-monetary objects before they
are added to the table.


jralls:
A better fix would be to get rid of the doubles.

jmarino:
I can see why the original author of the report used doubles, it makes this operation easy:
net = assets - liabilities

The same operation with gnc-monetary is not trivial. I came up with this function to add two gnc-monetary:

```
(define (monetary+ a b)
      (if (and (gnc:gnc-monetary? a) (gnc:gnc-monetary? b))
	  (let ((same-currency? (eq? (gnc:gnc-monetary-commodity a) (gnc:gnc-monetary-commodity b)))
		(denom (gnc:gnc-numeric-denom (gnc:gnc-monetary-amount a)))
		(amount (+ (gnc-numeric-to-double (gnc:gnc-monetary-amount a))
			   (gnc-numeric-to-double (gnc:gnc-monetary-amount b))))
		 )
	    (if same-currency?
		(gnc:make-gnc-monetary
		 (gnc:gnc-monetary-commodity a)
		 (gnc:make-gnc-numeric (inexact->exact (floor (* amount denom))) denom))
		(warn "incompatible currencies in monetary+: " a b))
	    )
	  (warn "wrong arguments for monetary+: " a b)))
```
Basically, we end up building gnc-monetary objects in the same way as the original patch.

christopherlam:
I haven't had a chance to test this code, but you can call gnc-numeric-add instead.
NB: use gnc-commodity-equal instead of eq? because occasionally eq? will trip you up. It took me a while to debug this.
```
(define (monetary+ a b)
  (if (and (gnc:gnc-monetary? a) (gnc:gnc-monetary? b))
      (let ((same-currency? (gnc-commodity-equal (gnc:gnc-monetary-commodity a) (gnc:gnc-monetary-commodity b)))
            (amount (gnc-numeric-add (gnc:monetary-amount a) (gnc:monetary-amount b))))
        (if same-currency?
            (gnc:make-gnc-monetary (gnc:gnc-monetary-commodity a) amount)
            (warn "incompatible currencies in monetary+: " a b)))
      (warn "wrong arguments for monetary+: " a b)))

```

jralls:
I'm pretty sure gnc_numeric_subtract() is exposed to Scheme, but Scheme also has its own rational number class and one could pretty easily convert a GncNumeric to a Scheme rational. I don't think that it would have the rounding control that GncNumeric provides but it would still be much more accurate than using doubles.

jmarino:
I reverted my original commit and added a new one where all values are computed and stored as gnc-monetary. I used @jralls much nicer and cleaner implementation of monetary+.

christopherlam:
This looks good; I assume you will also fix the line charts?

jralls:
It might look good but it causes the standard net barchart test to crash. That's got to get fixed for it to be merged.

jmarino:
I have found a bug in my patch that leads to the test failure. I mistakenly default to empty lists when there are no assets/liabilities:
```
(set! assets-list (if assets (car assets) '()))
(set! liability-list (if liabilities (car liabilities) '()))
```
In the original code, empty assets/liabilities default to a list of 0s. In the test there are no liabilities and the default empty list causes trouble. So that's easy to fix, I can default to a list of `<gnc-monetary>` 0s as the original code does (I never should have changed that). However, the test still fails. The new failure is due to something else: the table rendering code in the test produces invalid output that cannot be successfully parsed by the test. The test generates the report using `gnc:html-document-render` and this is what is produced for one row of the table:
```
<tr><td><string> 10/01/17</td>
<td class="number-cell"><gnc-monetary> #[unknown]</td>
<td class="number-cell"><gnc-monetary> #[unknown]</td>
<td class="number-cell"><gnc-monetary> #[unknown]</td>
</tr>
```

In the actual report inside gnucash those `<gnc-monetary>` values are displayed as numbers with currency, but `gnc:html-document-render` outputs `"<gnc-monetary> #[unknown]"`.

Is this a bug in `gnc:html-document-render` or is this the intended result?

jmarino:
It took some digging but I finally figured out the issue with the unit test and the mysterious `"<gnc-monetary> #[unknown]"` entries. It was caused by the lack of stylesheets in the unit test. Without stylesheets, the html code defaults to a trivial renderer that cannot handle `gnc-monetary` objects. After initializing the stylesheets, the `gnc-monetary` elements in the table are properly formatted and the table can be parsed by the test code to extract their values.

I have just pushed two commits where I fix the bug I introduced by defaulting to empty lists and I fix the unit test by initializing the stylesheets and adapting the parsing code to the new format.

Would it be worth it to do some restructuring of the commits to get rid of the 1st commit followed by its revert? That would provide a cleaner history.

gjanssens:
Yes, please do that restructuring. You can take the opportunity to rebase to the current unstable head rather than master. I think this fix is worth including in gnucash 2.7/2.8.

jmarino:
Rebased branch on current unstable (commit 5e7c2106e) and cleaned up the history a bit (got rid of commit+revert).

jmarino:
I just had a thought. I've been changing things so that computations are done in gnc-monetary. But, does it make more sense to do it in gnc-numeric?

I understand the point is to get away from floating point operations but using gnc-monetary may be overkill. An operation between gnc-monetary's requires conversions to gnc-numeric and then back to gnc-monetary. It seems things would be simpler to calculate in gnc-numeric and then convert to double for the charts and to gnc-monetary for the display.

jralls:
The second block in "files changed" is ensuring that the accumulation is same-commodity, so ISTM you need to use gnc-monetary. The size difference is a single pointer and the compiler will optimize away the extra indirection on the addition, nothing to worry about. It's fine.

jmarino:
Added fix for net-linechart report and its unit test.

jmarino:
I was in the process of applying the same fixes to category-barchart.scm  but I need some advice.
The original code applies an averaging factor when converting from collector to double:
```
(*
 (gnc-numeric-to-double
  (gnc:gnc-monetary-amount
   (gnc:sum-collector-commodity
    c report-currency
    (lambda (a b) (exchange-fn a b date)))))
 averaging-multiplier))
```

I want to apply this averaging-multiplier factor to a gnc-monetary, i.e., calculate a product between a gnc-monetary and a double. What is the best way to do it?

christopherlam:
Will this work? It's contrived, but may be a start. There are probably better ways.

```
(gnc:make-gnc-monetary
 REPORT-CURRENCY
 (gnc-numeric-mul
  (gnc:gnc-monetary-amount
   (gnc:sum-collector-commodity
    c report-currency
    (lambda (a b) (exchange-fn a b date))))
  (gnc:make-gnc-numeric (inexact->exact averaging-multiplier) 1)
  GNC-DENOM-AUTO
  GNC-RND-ROUND))
```

jralls:
It would work, but a better solution would be to change the computation of averaging-multiplier to result in a gnc-numeric.

jmarino:
rebased on current unstable and converted category-barchart.scm to gnc-monetary computations.

jmarino:
For the record, I followed @jralls advice and compute `averaging-multiplier` as a `gnc-numeric`. The code that applies the factor to a gnc-monetary then looks like this:
```
(gnc:make-gnc-monetary
 report-currency
 (gnc-numeric-mul
  (gnc:gnc-monetary-amount
   (gnc:sum-collector-commodity
    c report-currency
    (lambda (a b) (exchange-fn a b date))))
  averaging-multiplier currency-frac GNC-RND-ROUND)
 )
```

Where `currency-frac` comes from `(gnc-commodity-get-fraction report-currency)`.

christopherlam:
I'm reviewing all these .scm files, and I realize net-linechart and net-barchart are actually identical (except for styling options) and could be combined. But this will create a headache for backward compatibility...

jralls:
Backward compatibility for developers or for users? If it's users that's a concern. Developers not so much.

jmarino:
Applied the same idea discussed in PR #226. That is: configure load-path in the build system instead of in the code.

jralls:
If you rebase onto the commit I just pushed to unstable the CMake tests should pass; they do when I tested locally.

jmarino:
Ok, I should at least get rid of the `(add-to-load-path "../../stylesheets/")` lines.
I'll do a rebase from unstable in any case.

jralls:
If you've already rebased, do it again. I fixed the arch linux gwenhywfar compile error so barring a guile segfault in the arch autotools build it should pass now.

jmarino:
Rebased from current unstable.

jralls:
Thanks!

PR #215, opened 2017-10-12
Fix report colors

jmarino:
Many of the standard reports specify colors to use in their charts. However, these colors are ignored in the chart rendering code.

These commits modify the rendering code of all 3 types of charts (bar, pie and line) to fix this issue. This is done by adding the jqplot option "seriesColors"  to pass the requested bar/pie/line colors to jqplot. If no colors are requested, we let jqplot use default colors by setting "seriesColors: false".


jralls:
Thanks!

Now wouldn't it be nice to have options in report options so that the user can set the colors... ;-)

jmarino:
That would indeed be very nice. I'll look into it.

PR #216, opened 2017-10-16
Fix spelling insensative->insensitive

christopherlam:
Not sure whether to patch unstable or master. Both have spelling mistake.


jralls:
Thanks!

fellen:
@christopherlam: we usually merge maint (2.6) -> unstable (2.7/8) -> master (later),
so branch from the lowest target, where it should be applied.

PR #217, opened 2017-10-16
Fixed null pointer dereferencing during first transaction retrieval

ethaden:
This fixes the issue that the first attempt to import transactions into a newly created account synchronized via HBCI crashes Gnucash because there is no date of last retrieval available for that account.

### DIFF
--- a/gnucash/import-export/aqb/gnc-ab-kvp.c
+++ b/gnucash/import-export/aqb/gnc-ab-kvp.c
@@ -99,9 +99,17 @@ Timespec
 gnc_ab_get_account_trans_retrieval(const Account *a)
 {
     Timespec *t = NULL;
+    Timespec never;
+    never.tv_nsec = 0;
+    never.tv_sec = 0;
+    
     qof_instance_get (QOF_INSTANCE (a),
 		      "ab-trans-retrieval", &t,
 		      NULL);
+    if (!t)
+    {
+      return never;
+    }
     return *t;
 }
 


jralls:
Timespec 0 isn't "never", it's 1970-01-01 00:00:00. That particular time causes trouble with older versions of Windows and with MySQL. It's safer to return timespec_now() and probably makes more sense as well.

Also this change should go on the maint branch, not onto master.

ethaden:
Of course, you are right, technically that timespec does not mean
"never". But I've chosen that variable name from a semantic point of
view: "Never" represents a special valuation for the timespec var which
can easily be checked for by calling methods.

Unfortunately, there is a design flaw in method
"gnc_ab_get_account_trans_retrieval" in file
"gnucash/import-export/aqb/gnc-ab-kvp.c". That method has the
precondition, that for every HBCI-enabled account the date of last
transaction retrieval is available. Based on the precondition,
dereferencing "t" can be done safely.

But for accounts for which HBCI just has been configured, that
precondition does not hold as there has never been a retrieval for them
and no initial last retrieval date is set.

My first idea was to make sure that the precondition always holds by
adding an initial date to the account during HBCI configuration. But
unfortunately this would not work for accounts which are already
HBCI-enabled. The user would have to reconfigure them actively,
otherwise gnucash would keep crashing. Not a good choice in my opinion.

A better approach would be to relax the precondition by modifying the
method such that it checks for the existence of a date before
dereferencing the pointer and providing some initial value otherwise.

Another approach would be to throw an exception (of course this would
work only if the code is transformed to C++ already). But not having a
date of last retrieval is not an unexpected situation in my opinion. One
should not use exception for that.

Providing the current time is - in my opinion - semantically wrong: The
method is meant to provide the date of last transaction retrieval and
should be able to tell the caller, that no retrieval has been performed
yet.

Btw, in file "gnc-ab-gettrans.c, the following code is used:

    last_timespec = gnc_ab_get_account_trans_retrieval(gnc_acc);
    if (last_timespec.tv_sec == 0)
    {
        use_last_date = FALSE;
        timespecFromTime64 (&last_timespec, now);
    }

The author of those lines expect a) to always get a valid timespec in
the first line and b) (tv_sec==0) to represent the case where no last
update is available. Obviously, by providing the current date as special
 value would break his second assumption.

My suggestion: Either use my patch or modify it slightly by changing the
name "never" to something which better explains the intention.

Regarding the MySQL and Windows: Currently that "never" date is used
only inside of the HBCI dialog for retrieving the latest transactions
via HBCI. It is not saved into MySQL. I don't think that there is a
issue with Windows either. Of course, you'll never know where a method
will be used in the future. But for now, I suggest not to introduce
workarounds for things with very low likeliness of becoming issues in
the future.

Regarding the branch "maint": I'm still trying to understand the branch
model used for Gnucash. When writing the patch, I didn't understand how
commits in "maint" would go to "master" (and I still don't know). But I
could see in the repository that from the git perspective, the aqbanking
code has been reorganized in master compared to the "maint" branch (I
guess the folder names have been changed). I suspect that merging back
changes to the aqbanking code in "maint" into "master" will cause some
trouble. Therefore I started directly on "master".

gjanssens:
The branch model in gnucash is as follows. At any one time there will be 2 or 3 primary branches:
- maint: shorthand for maintenance, for commits that should go into the current stable release series (currently 2.6)
- unstable: only used near the end of a big development cycle, tracks changes that lead up to the upcoming next major release and will be used for beta-releases.
- master: for work that's intended for the next major release. As long as there's no unstable branch, this is the immediatly next major release (like 2.8). When unstable is branched work anything for 2.8 should go on unstable, and master is open for longer term development for 2.10/3.0.

The branch to base your work on is always the oldest one that requires the fix. If the bug already exists in maint and can be fixed there, do it there. If the bug only affects the upcoming major release it should go on unstable. Otherwise on maint.

We regularly merge all changes from maint->unstable->master so the more recent branches eventually get all changes from the less recent ones.

And regarding the directory restructuring, you don't have to worry about that. In 99% of the cases git handles this very well. The only exception I have seen so far is if you add a new file in a new subdirectory. That case needs extra attention during merge.

As to the core of the patch. Using Timespec {0, 0} looks ok on maint. There are other locations in the code where we still have to assume the 1970-01-01 date is meant the "invalid" date. This will very likely change for 2.8 though. I think here the other option is to alter the function signature to return a Timespec* instead and test for NULL in the calling functions. That will work on all branches IMO.

jralls:
I've pushed my simpler fix. Thanks for the report.

PR #221, opened 2017-10-23
use correct pointer (gnc_numeric**) with qof_instance_get

jmarino:
Function `qof_instance_get` needs a `gnc_numeric**` instead of a `gnc_numeric*` to work correctly.
The proper usage of function `qof_instance_get` with a `gnc_numeric` is shown in function:
```
split-register-model.c:gnc_template_register_get_debcred_entry
```

This bug had an interesting side effect. It was causing the session to be marked as dirty on startup: the Save button is active and an asterisk (*) appears in the window title.
One of the last steps when reading the user data file is to replay any scheduled transaction. This bug produced wrong values for `"sx-debit-numeric"` and `"sx-credit-numeric"` which forced the transaction
to be modified marking the whole session as dirty.

### DIFF
--- a/libgnucash/app-utils/gnc-sx-instance-model.c
+++ b/libgnucash/app-utils/gnc-sx-instance-model.c
@@ -78,7 +78,7 @@ scrub_sx_split_numeric (Split* split, const char *debcred)
     const char *numeric = is_credit ?
         "sx-credit-numeric" : "sx-debit-numeric";
     char *formval;
-    gnc_numeric numval;
+    gnc_numeric *numval = NULL;
     GHashTable *parser_vars = g_hash_table_new (g_str_hash, g_str_equal);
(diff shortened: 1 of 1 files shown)


gjanssens:
This PR seems closely related to [PR#145](https://github.com/Gnucash/gnucash/pull/145). Perhaps they should be considered together.

jmarino:
Sorry I didn't see PR #145. I see you guys are on top of this already.
I must admit I did not fully understand how things fully work or why my patch fixed anything at all. I just replicated the way gnc-numerics are accessed somewhere else.

gjanssens:
Well, since two people have now offered (the same) solution to the problem, this tickled me enough to complete the issues brought up in PR #145 

It should now be solved, using your (and @Bob-IT's patch). Thanks again!

PR #222, opened 2017-10-23
Flat Bayes Import Map KVP

limitedAtonement:
This also required compiling `Account.c` as `Account.cpp`. There were a number of changes to `kvp-frame` (and exposed by `qofinstance`) to support flat KVP structures. This allows callers to search KVP by prefix. Each commit passes make check and has somewhat detailed comments.

### DIFF
--- a/common/test-core/test-stuff.h
+++ b/common/test-core/test-stuff.h
@@ -57,7 +57,9 @@ Otherwise, only failures are printed out.
 #include <glib.h>
 #include <stdlib.h>
 
-
+#ifdef __cplusplus
+extern "C" {
+#endif
 
 /* Privately used to indicate a test result. You may use these if you
  * wish, but it's easier to use the do_test macro above.
@@ -150,5 +152,8 @@ gint64 get_random_gint64(void);
 double get_random_double(void);
 const char* get_random_string_in_array(const char* str_list[]);
 
+#ifdef __cplusplus
+} /*extern "C"*/
+#endif
 
 #endif /* TEST_STUFF_H */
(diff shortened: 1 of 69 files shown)


limitedAtonement:
I think I pushed too quickly and broke the travis build (https://travis-ci.org/Gnucash/gnucash/builds/292637860). Can someone restart it? Or perhaps I should rewrite and force push the branch again?

gjanssens:
@limitedAtonement Do you not have authority to do so yourself ? I see a "Restart Build" button on the right hand side of the build's overview page. And there are also buttons to restart each individual job. If they are not there for you, I'll just click them for you...

limitedAtonement:
I don't have authority, but I know what you're talking about. On my travis-ci account, I see them, but not on Gnucash's:

image
image

gjanssens:
Interesting. I would expect one to have authorization to restart tests for PR's you created yourself, but that's a wrong assumption. I have restarted for you.

limitedAtonement:
The build failed because of a docker image problem; Travis says, "We’re investigating reports of timeouts in builds while pulling images from Docker Hub." Could you restart the two arch linux builds?

gjanssens:
Done. BTW, the Arch autotools build fails more frequently than the other builds. I don't know why though.

limitedAtonement:
Yeah, I don't know why, either. It seems to be a race condition in the build scripts, but I don't know why it would be exposed more on an arch docker image. It does happen on the ubuntu build sometimes, too, right?

limitedAtonement:
It looks like the travis CI integration isn't working...I don't see it at least. I'll close and reopen.

PR #223, opened 2017-10-24
Further income-gst-statement and transaction enhancements

christopherlam:
This PR will do the following

1. Enhance the 'new feature' account selector from a substring matcher to full regex capabilities. This will apply to both _transaction.scm_ and _income-gst-statement.scm_ - I think the average gnucash user is capable of handling basic regex matching techniques.
2. The (split-adder) function in income-gst-statement gets the full functional programming treatment, avoiding (set!) calls.
3. Change 'Total Sales/Purchases' headings to uppercase which will aid differentiation for the accountant. Improve Input/Output tax account reporting to display 'None' if no tax accounts were selected.


christopherlam:
I think while 2.7.0 is still cooking there's still time to upgrade the account substring into a regex matcher.

Plus, there is a demand for a description/notes/memo filter in transaction.scm - perhaps a regex matcher can also be offered? This may need another tab, I can add in another PR.

jralls:
I disagree that the average user is able to handle regexes, though since we offer regexes in the Find dialog that's not a good reason not to offer it here, too. That said, most users should be comfortable with globs (they'll know them as "wildcards"). Would that be sufficient for your use case?

christopherlam:
The main motivation towards regex matcher stems from Doug's popular transaction.scm enhancements; see screenshot to illustrate a "complex" description substring matcher but has very limited options. I have some reservations about his suggested options and the general code cleanliness, but I figure incremental upgrades to transaction.scm will be welcome. 

My own use case is actually achieved via coding Expenses:Chris^1:(numerous subaccounts) and I'll just scan for ^1, therefore I don't actually need regex. But description/notes/memo matcher could benefit from a substring/regex filter too.

What are your thoughts?

Most importantly I wouldn't know how to formulate a glob matcher in scheme... regex was easily callable and seemed to fit the bill.

image

jralls:
Slib provides [filename-match and filename-substitute](http://soc.if.usp.br/manual/slib/slib_4.html#Filenames) that you could probably bend to your will. Another angle would be to parse the submission into a regex by substituting '.*' for '*', '.?' for '?' and passing the result to regex. In that implementation you could even provide a checkbox that would enable full regex support, but it will need to say what flavor of regex (POSIX extended) either in documentation or a tooltip.

Separately neither globbing nor regex matching should be the default: Users who don't know what a regex is but who include regex or glob special characters in their account names are going to be unpleasantly surprised when their strings don't match or match more than they want because they didn't understand that they have to escape some of the characters.

christopherlam:
I've checked out glob which is nice, but doesn't provide OR matching.

How does the following look?

image

jralls:
Yes, it's true that globs are a lot more limited than regexes. That's kind of the point: No one has ever needed to write a 300-page book called "Mastering Globs".

The check box is OK, but spell out "regular expressions" and the default should be unchecked.

christopherlam:
Screenshot for Transaction Matcher. Tooltips displayed.

I'd add to General tab. If we add another tab, the tablist will be vertical.

image

christopherlam:
I think this is now complete. Upgrades income-gst-statement and transaction.scm. The main visible changes are

1. Account substring matcher upgraded to optional regex
2. Introduce a Transaction Desc/Memo/Notes matcher - is also a substring (or regex) matcher
3. Some strings improved for further visibility/clarity

## Under the hood
1. income-gst-statement core (split-adder) function upgraded to pure FP
2. some internal functions (splits-filter-unique-transactions) and (is-filter-member) upgraded to use srfi-1 functions
3. combined filtering (to/from accounts, substring/regex matcher) seems elegant without losing clarity

tested using 2.7 and 2.6.18

jralls:
Merged, thanks!

PR #224, opened 2017-10-26
Account to cpp

limitedAtonement:
I think this will fix https://bugzilla.gnome.org/show_bug.cgi?id=761682, but I would appreciate another pair of eyes. It involves a run-once conversion of the bayes import data.


limitedAtonement:
I thought of another case that I need to test...Sorry for on again and off again, it seems there are a lot of moving parts to this one.

PR #225, opened 2017-10-27
Updating home directory to fix CI error messages

limitedAtonement:
Also removed some redundant build configurations.


jralls:
Merged via unstable. Thanks.

PR #226, opened 2017-10-27
Cashflow barchart report

jmarino:
Three small changes to the just merged new report cashflow-barcharct.scm

The first 2 simplify the code a little bit and the 3rd adds a new feature: the ability to control the display of money in, out or net independently.

### DIFF
--- a/gnucash/report/standard-reports/cashflow-barchart.scm
+++ b/gnucash/report/standard-reports/cashflow-barchart.scm
@@ -217,12 +217,12 @@
                (in-list '())
                (out-list '())
                (net-list '())
+               (total-in #f)
+               (total-out #f)
+               (total-net #f)
                (in-value-list #f)
                (out-value-list #f)
                (net-value-list #f)
-               (in-total-collector (gnc:make-commodity-collector))
-               (out-total-collector (gnc:make-commodity-collector))
-               (net-total-collector (gnc:make-commodity-collector))
                )
 
           ;; Helper function to convert currencies
@@ -236,6 +236,22 @@
             (gnc:sum-collector-commodity
              collector report-currency exchange-fn)
             )
+
+          ;; Add two or more gnc-monetary objects
+          (define (monetary+ a . blist)
+            (if (null? blist)
+                a
+                (let ((b (apply monetary+ blist)))
+                  (if (and (gnc:gnc-monetary? a) (gnc:gnc-monetary? b))
+                      (let ((same-currency? (gnc-commodity-equal (gnc:gnc-monetary-commodity a) (gnc:gnc-monetary-commodity b)))
+                            (amount (gnc-numeric-add (gnc:gnc-monetary-amount a) (gnc:gnc-monetary-amount b) GNC-DENOM-AUTO GNC-RND-ROUND)))
+                        (if same-currency?
+                            (gnc:make-gnc-monetary (gnc:gnc-monetary-commodity a) amount)
+                            (warn "incompatible currencies in monetary+: " a b)))
+                      (warn "wrong arguments for monetary+: " a b)))
+                )
+            )
+
           ;; Convert gnc:monetary to number (used to generate data for the chart)
           (define (monetary->double monetary)
             (gnc-numeric-to-double (gnc:gnc-monetary-amount monetary))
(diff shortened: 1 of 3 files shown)


jralls:
These changes break test-cashflow-barchart:
irectory: /Users/john/Development/Gnucash-Build/Gnucash-master-git/build/gnucas
h-git/gnucash/report/standard-reports/test
"test-cashflow-barchart" start time: Oct 28 14:17 PDT
Output:
Backtrace:
In ice-9/boot-9.scm:
 160: 11 [catch #t #<catch-closure 10f78f580> ...]
In unknown file:
   ?: 10 [apply-smob/1 #<catch-closure 10f78f580>]
In ice-9/boot-9.scm:
  66: 9 [call-with-prompt prompt0 ...]
In ice-9/eval.scm:
 432: 8 [eval # #]
In ice-9/boot-9.scm:
2404: 7 [save-module-excursion #<procedure 10f814a00 at ice-9/boot-9.scm:4051:3
()>]
4056: 6 [#<procedure 10f814a00 at ice-9/boot-9.scm:4051:3 ()>]
In /Users/john/Development/gtk-sources/gnucash-git/gnucash/report/standard-repor
ts/test/test-cashflow-barchart.scm:
  34: 5 [#<procedure 10f930180 ()>]
In ice-9/boot-9.scm:
3589: 4 [process-use-modules (((gnucash report stylesheets)))]
 705: 3 [map #<procedure 10f79cc80 at ice-9/boot-9.scm:3589:25 (mif-args)> ((#))
]
3590: 2 [#<procedure 10f79cc80 at ice-9/boot-9.scm:3589:25 (mif-args)> (#)]
2870: 1 [resolve-interface (gnucash report stylesheets) #:select ...]
In unknown file:
   ?: 0 [scm-error misc-error #f ...]

ERROR: In procedure scm-error:
ERROR: no code for module (gnucash report stylesheets)
<end of output>
Test time =   0.54 sec
Test Failed.
"test-cashflow-barchart" end time: Oct 28 14:17 PDT
"test-cashflow-barchart" time elapsed: 00:00:00

jmarino:
That's weird, I don't see that here. The error happens in test-cashflow-barchart.scm when it tries to load the stylesheets module and these 3 changes did not modify that file. Sometimes I've seen weird errors like this one that got fixed after a make clean ; make.

Is the loading of the stylesheets module done correctly?
```
(add-to-load-path "../../stylesheets/")
(use-modules (gnucash report stylesheets))
```
I don't really know if this is the proper way of doing it.

The module needs to be explicitly loaded in the test files to ensure that `gnc-monetary` values are rendered correctly. I believe this is not needed during normal operation because the stylesheets module is loaded from c code (in gnucash-bin.c).

jralls:
The problem seems to be on the cmake side, and I discovered that it fails after the earlier PR as well. 

Yes, (add-to-load-path "../../stylesheets") is the likely problem. I'd expect the path to be relative to cwd rather than the location of the source file.

jralls:
Hmm, but it passes on Travis. The failures are on both mac  and my Fedora 26 VM.

jmarino:
What about this patch, moving the load-path addition to the build system:
```
diff --git a/gnucash/report/standard-reports/test/CMakeLists.txt b/gnucash/report/standard-reports/test/CMakeLists.txt
index 7be4c9c67..77c6badd6 100644
--- a/gnucash/report/standard-reports/test/CMakeLists.txt
+++ b/gnucash/report/standard-reports/test/CMakeLists.txt
@@ -10,11 +10,13 @@ SET(scm_test_standard_reports_SOURCES
   test-generic-net-linechart.scm
 )
 
+SET(GUILE_LOAD_DIRS    libgnucash/app-utils libgnucash/engine libgnucash/core-utils libgnucash/gnc-module gnucash/gnome-utils gnucash/report/report-system libgnucash/scm gnucash/report/stylesheets)
+
 GNC_ADD_SCHEME_TARGETS(scm-test-standard-reports
   "${scm_test_standard_reports_SOURCES}"
   gnucash/report/standard-reports/test
   ""
-  ""
+  "${GUILE_LOAD_DIRS}"
   ""
   "scm-gnc-module;scm-test-report-system"
   FALSE
diff --git a/gnucash/report/standard-reports/test/Makefile.am b/gnucash/report/standard-reports/test/Makefile.am
index 2a4e80803..78ebd09b7 100644
--- a/gnucash/report/standard-reports/test/Makefile.am
+++ b/gnucash/report/standard-reports/test/Makefile.am
@@ -30,6 +30,7 @@ GNC_TEST_DEPS = \
   --guile-load-dir ${top_builddir}/gnucash/report/report-system/test \
   --guile-load-dir ${top_builddir}/gnucash/report/standard-reports \
   --guile-load-dir ${top_builddir}/gnucash/report/standard-reports/test \
+  --guile-load-dir ${top_builddir}/gnucash/report/stylesheets \
 \
   --library-dir    ${top_builddir}/gnucash/report/report-system \
   --library-dir    ${top_builddir}/libgnucash/core-utils \
diff --git a/gnucash/report/standard-reports/test/test-cashflow-barchart.scm b/gnucash/report/standard-reports/test/test-cashflow-barchart.scm
index e4f7dde65..eacbfe61d 100644
--- a/gnucash/report/standard-reports/test/test-cashflow-barchart.scm
+++ b/gnucash/report/standard-reports/test/test-cashflow-barchart.scm
@@ -30,7 +30,6 @@
 (use-modules (gnucash report report-system test test-extras))
 (use-modules (gnucash report standard-reports cashflow-barchart))
 
-(add-to-load-path "../../stylesheets/")
 (use-modules (gnucash report stylesheets))
 
 (use-modules (ice-9 format))
```

jralls:
That would actually work, but I bet it's not for the reason you think, and removing add-to-load-path would break the autotools build without some Makefile.am changes.

It turns out that the problem is that I usually call ninja check, relying on the test dependencies to get everything built. It doesn't build stylesheets.go, so the test fails. If I run ninja && ninja check the test passes. CMake builds put the build products in one place so the add-to-load-path dance is superfluous.

jmarino:
That explains why I couldn't make the test fail when building with cmake. 

Then the extra load-path is only needed when building with autotools (which is what I was using) and this patch should do the trick, right?
```
diff --git a/gnucash/report/standard-reports/test/Makefile.am b/gnucash/report/standard-reports/test/Makefile.am
index 2a4e80803..78ebd09b7 100644
--- a/gnucash/report/standard-reports/test/Makefile.am
+++ b/gnucash/report/standard-reports/test/Makefile.am
@@ -30,6 +30,7 @@ GNC_TEST_DEPS = \
   --guile-load-dir ${top_builddir}/gnucash/report/report-system/test \
   --guile-load-dir ${top_builddir}/gnucash/report/standard-reports \
   --guile-load-dir ${top_builddir}/gnucash/report/standard-reports/test \
+  --guile-load-dir ${top_builddir}/gnucash/report/stylesheets \
 \
   --library-dir    ${top_builddir}/gnucash/report/report-system \
   --library-dir    ${top_builddir}/libgnucash/core-utils \
diff --git a/gnucash/report/standard-reports/test/test-cashflow-barchart.scm b/gnucash/report/standard-reports/test/test-cashflow-barchart.scm
index e4f7dde65..eacbfe61d 100644
--- a/gnucash/report/standard-reports/test/test-cashflow-barchart.scm
+++ b/gnucash/report/standard-reports/test/test-cashflow-barchart.scm
@@ -30,7 +30,6 @@
 (use-modules (gnucash report report-system test test-extras))
 (use-modules (gnucash report standard-reports cashflow-barchart))
 
-(add-to-load-path "../../stylesheets/")
 (use-modules (gnucash report stylesheets))
 
 (use-modules (ice-9 format))
```

jralls:
Sorry, I stared right past the Makefile.am part of the patch. Yes, adding it there is better than calling add-to-load-path in test-cashflow-barchart.scm. Please also delete the line above add-to-load-path so that the gnucash modules are grouped together.

The cmake change isn't quite right as I said and while it would probably work with ninja it probably won't with cmake and make. The GNC_ADD_SCHEME_TEST function needs a way to set dependencies for its targets. I'll work on that a bit; meanwhile go ahead and push the autotools patch and I'll merge this PR.

jmarino:
Committed patch as requested.

jralls:
Thanks!

PR #227, opened 2017-10-31
delete ignored files from repository

jmarino:
These 3 files are part of the repository and ignored at the same time. They are automatically generated, they belong in .gitignore and shouldn't be tracked by git.

These files were found by running:
```
for i in `git ls-tree --full-tree -r HEAD | awk '{print $4}'`; do grep "^$i$" .gitignore; done
```

I'm not sure about the ChangeLog file though. Do we want this file to be tracked by git? If so it should be deleted from .gitignore.


jralls:
Geert added unittest_support.py by accident when he moved the directories around, that one should go.

ChangeLog is special: We want it in git but we want it updated only during a release.

I think POTFILES.in is in git to make sure that it gets into the release tarballs where it's needed to generate gnucash.pot. @fellen, do you know?

gjanssens:
I can't find a hard statement on the internet that POTFILES.in has to be in the source tree. The [GNU gettext manual](https://www.gnu.org/software/gettext/manual/html_node/po_002fPOTFILES_002ein.html#po_002fPOTFILES_002ein) considers this a source file and seems to assume it's manually maintained. Gnucash on the other hand generates it automatically based on a few scripts so that slightly changes things. I don't see a strong reason to keep it in git so far. It think it should go in the dist tarball though. And if we ever want our git tree to be a full dist tarball we should commit this. We currently don't support source as direct dist tarball (due to swig bindings), so I think it can be removed from git for now.

Out of curiosity, why do we want the ChangeLog to be in git ?

jralls:
ChangeLog is special because automake breaks the build if it's not there (along with AUTHORS, NEWS, COPYING, and README). We can get rid of it when we drop autotools.

I think POTFILES.in is in VCS because it always has been. As long as we're careful to ensure that it gets in the tarballs it doesn't need to be in git.

BTW, I just removed unittest_support.py as part of fixing python tests from tarball builds.

gjanssens:
Thanks for the unittest_support.py fix.

I would go for dropping POTFILES.in from the git repo and only include it in the dist tarballs. And I  can imagine we can even drop it from there the day we drop autotools support. From what I understand cmake will regenerate the file on each cmake invocation (which is triggered automatically whenever a CMakeLists.txt file changes, which is turn is the only way we could possibly have changes in the files to include in POTFILES.in).

To the best of my knowledge this file is only used to generate a new gnucash.pot file, which we also ship in the distribution tarballs. I can't imagine a use case or workflow where someone would require the POTFILES.in file in a situation where it's not possible to regenerate it by running cmake/make/ninja-build or similar. Of course, I may be wrong. In that case the file is easily readded.

jralls:
Hmm, true, as long as gnucash.pot is present the rest is immaterial. If autotools can be made to not complain maybe it doesn't need to be in the tarball either.

jmarino:
There's no value in keeping my original commit. I propose we drop this PR and I let you guys fix it. I should have filed this as a bug report instead of a PR.

jralls:
Jmarino: OK. Thanks for bringing it up.

PR #228, opened 2017-11-06
report: fix negative colors in barcharts

jmarino:
Bar colors specified by a report only apply to positive bars in the plot. Negative bars still have the default colors, which are slightly darker shades of the default positive colors.

This commit forces negative bars to have the same colors as positive bars.


jralls:
Thanks!

PR #229, opened 2017-11-08
Initial commit refactored transaction.scm

christopherlam:
## Main changes are:
1. Refactored whole report
2. Centralised sortkeys and sortkeys data
3. Large-text comments to ease navigation

## User-visible changes
1. Removed 'exact-time sorting option
2. Introduce Transaction substring/regex matcher
3. Introduce Reconciled-status filter
4. Move Account, Transaction and Reconciled status filter to new Filter tab
5. Debit/Credit amounts (i.e. double-amount display) now no longer test for reversed? amounts signs
   They test for signs based on internal representation, + = debit, - = credit. More testing/confirmation required
6. Optional custom sorter in scheme which allows better date sorting options
7. Add 'Show Account Description' into subheadings for account-related sortkeys

## Pending:
1. Overhaul of amounts display to allow multicolumns and multicolumn subtotals. Further discussion welcome.


gnucash/report/standard-reports/transaction.scm
```diff
@@ -933,21 +572,28 @@ Use a period (.) to match a single character e.g. '20../.' will match 'Travel 20
...
         (set! prime-sortkey x)
         (apply-selectable-by-name-sorting-options))))
-    
+
     (gnc:register-trep-option
      (gnc:make-simple-boolean-option
       pagename-sorting optname-full-account-name
       "j1"
-      (N_ "Show the full account name for subtotals and subtitles?")
+      (N_ "Show the full account name for subtotals and subheadings")
```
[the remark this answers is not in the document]
christopherlam: Sure, will fix; I think originally there were ?s

gnucash/report/standard-reports/transaction.scm
```diff
@@ -46,627 +48,385 @@
...
+;   ;   ;  ;   ;   ;; ;   ;     ;  ;   ;   ;  ;   ;  ;   ; 
+;   ;;     ;   ;   ;      ;     ; ;    ;   ;   ;  ;  ;;    
+;     ;;   ;   ;   ;      ;     ;;;    ;;;;;   ; ;     ;;  
+;       ;  ;   ;   ;      ;     ; ;;   ;       ; ;       ; 
+;   ;   ;  ;   ;   ;      ;     ;  ;;  ;;      ; ;   ;   ; 
+;   ;;;;    ;;;   ;;;;     ;;;  ;   ;;  ;;;;    ;    ;;;;  
+;                                               ;          
+;                                              ;;          
+;                                             ;            
+;
```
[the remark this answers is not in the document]
christopherlam: These are useful in the code map in DrRacket, can remove

christopherlam:
> I guess this is a debug statement you forgot ?

No can't see debug code in there

christopherlam:
https://github.com/christopherlam/gnucash/pull/6 for more commentary on changes

gnucash/report/standard-reports/transaction.scm
```diff
@@ -1577,30 +1732,34 @@ Credit Card, and Income accounts."))))))
...
-               (gnc:html-document-add-object!
-                document                 
-                (infobox)))
-
+                (gnc:html-document-add-object!
+                 document
+                 (gnc:make-html-text
+                  (gnc:html-markup-h2 NO-MATCHING-TRANS-HEADER)
+                  (gnc:html-markup-p NO-MATCHING-TRANS-TEXT)))
+
```
gjanssens: Whitespace on empty line

gnucash/report/standard-reports/transaction.scm
```diff
@@ -1577,30 +1732,34 @@ Credit Card, and Income accounts."))))))
...
+                (gnc:html-document-add-object!
+                 document
+                 (gnc:make-html-text
+                  (gnc:html-markup-h2 NO-MATCHING-TRANS-HEADER)
+                  (gnc:html-markup-p NO-MATCHING-TRANS-TEXT)))
+                
+                (gnc:html-document-add-object!
+                 document                 
+                 (infobox)))
+
```
gjanssens: Whitespace on empty line

gnucash/report/standard-reports/transaction.scm
```diff
@@ -13,6 +13,9 @@
 ;; - introduced account/transaction substring/regex matcher
 ;; - add custom sorter in scheme
 ;; - common currency - optionally show original currency amount
+;;   and enable multiple data columns
+;; - add informational box, summarising options used, useful
+;;   to troubleshoot reports
```
gjanssens: In general try to avoid doing things in a commit that you didn't specify in the commit message.
In this case it's pretty harmless as it's just an extra comment. I got more confused with your later commits.

If you feel like it you could take this opportunity to practice rewriting your git history using a mixture of git reset/git add -i and so on. If you do, create a new branch at this commit, from which you can the revert (mixed mode) the last commit. That will put your changes back in your working directory ready to be added/committed. Then a git add -i will allow you to only select some changes to go into a commit. Commit that and repeat until all changes are split up in separate commits. From there on you can cherry-pick or rebase the later commits from your other branch on top of this new one and continue to clean up.

While not mandadory it's certainly useful to learn these techniques as it will allow you to write better history (both for yourself as the other developers looking into your changes).

gnucash/report/standard-reports/transaction.scm
```diff
@@ -907,8 +907,7 @@ Credit Card, and Income accounts."))))))
                               (begin
                                 (if column-amount
                                     (set! merging-subtotal
-                                          (merge-fn merging-subtotal column-amount
-                                                    GNC-DENOM-AUTO GNC-RND-ROUND)))
+                                          (merge-fn merging-subtotal column-amount)))
```
gjanssens: This is not an "old comment removed" and I'm still not sure why you chose to change the rounding strategy. So your commit message is a bit lacking here/your commit mixes two separate topics.

gnucash/report/standard-reports/transaction.scm
```diff
@@ -417,7 +417,12 @@ tags within description, notes or memo. ")
                   (N_ "Include transactions to/from filter accounts only."))
           (vector 'exclude
                   (N_ "Exclude Transactions to/from Filter Accounts")
-                  (N_ "Exclude transactions to/from all filter accounts.")))))
+                  (N_ "Exclude transactions to/from all filter accounts.")))
+   #f
+   (lambda (x)
+     (gnc-option-db-set-option-selectable-by-name
+      options gnc:pagename-accounts optname-filterby
+      (not (eq? x 'none))))))
```
gjanssens: Your commit message doesn't explain this very well. And while on that subject, it looks like this commit is a mixture of at least 2 changes that are not related - a comment cleanup and a change in the info box functionality. These beg for separate commits. Another opportunity for git history rewriting...

gnucash/report/standard-reports/transaction.scm
```diff
@@ -1513,7 +1521,22 @@ Credit Card, and Income accounts."))))))
...
+	     (map (lambda (x) (assq-ref (list (cons #\c "cleared")
+					      (cons #\f "frozen")
+					      (cons #\n "non-reconciled/cleared")
+					      (cons #\y "reconciled")
+					      (cons #\v "void"))
+					x))
+			       reconcile-status-filter)
+	     ", ")
+	    "<br>")
+	   "")
```
gjanssens: I mentioned before the infobox message have translation issues. I'm mentioning it again as this is in a separate commit which may or may not hide the old review comment.

gnucash/report/standard-reports/transaction.scm
```diff
@@ -1487,9 +1482,10 @@ Credit Card, and Income accounts."))))))
             "<b>Split Reconciled Status: </b>"
             (string-join
              (map (lambda (item)
-                    (cdr (assq item (list (cons #\n "unreconciled")
-                                          (cons #\c "cleared")
-                                          (cons #\y "reconciled")))))
+                    (case item
+                      ((#\n) "unreconciled")
+                      ((#\c) "cleared")
+                      ((#\y) "reconciled")))
```
gjanssens: Can we reuse the strings as used in the combobox (in this case that means capitalized) ?
That would make for less translatable strings (that is they can be reused more that once).

christopherlam:
Hi. Thank you for going through with a fine comb. My aim is to seek feedback on new "features" eg string/regex matchers/filters, multicolumn display, while learning new scheme idioms.

More importantly I'm really needing to confirm my functional changes e.g.:

- strategy on reversing signs + debit - credit is correct?

- my new subtotal strategy which adds internal values rather than (reversed) display amounts.

- my custom sorter which overrides qof-query sorting strategy but is IMHO better in cases described in http://gnucash.1415818.n4.nabble.com/transaction-scm-date-subtotal-options-calling-wizards-td4693895.html -- the custom sorter should work better but is still a hack; and if this is approved I'd refactor its strategy into the main sortkeys. Also "Use custom sorter" isn't a good option to present. Do I keep it? Rename it? Wait for qof-query sorter to upgrade its strategy (when sort=date is chosen and subtotal=week for example it'll always strictly sort by daily dates; we want it to keep weeks alone!)?

- I know the infobox is hacking English into scheme and hard-codes HTML ;) this is also experimental and unsure whether to keep. And whether to add toggle in General. Feedback?

- multicolumn data display is an actual change in code whereas most of the above are refactoring/embelishments - seeking feedback- Add new column ideas? Remove?

These are actual functional changes which can cause long-term users to wonder why suddenly the reports have changed (in a good way I think).

With regards to backward-compatible options - there's code somewhere within helper modules that will dynamically rename options https://github.com/Gnucash/gnucash/blob/037c93faea77f51a36d7391d0f3ef403b5036843/libgnucash/app-utils/options.scm#L1685 - I don't think we have to necessarily add it there, but it's food for thought at least. I'll need to think how back/forward compatibility can be maintained.

Overall while I think the functional changes are positive and useful, I'm not sure whether they fit in the long-term strategy on what the TR should be. I don't necessarily wish to make it my exclusive project!

gnucash/report/standard-reports/transaction.scm
christopherlam: Hmm because I discovered gnc-numeric-add-fixed and thought it was rather underused :) ... and if I strictly use additions/subtractions then the rounding strategy will always be irrelevant? IIRC the rounding errors tend to happen when calculating taxes which involve multiplication. I haven't looked deeply into the rounding options but my guess is they're all harmless when adding numbers?

christopherlam:
PS As far as rewriting history... well this will be a whole new learning curve! :)

gnucash/report/standard-reports/transaction.scm
jralls: >  I haven't looked deeply into the rounding options but my guess is they're all harmless when adding numbers?

Nope. GncNumeric is a rational number. As long as the denominators are the same then there's rounding only on very large numbers, but if the denominators differ and don't have smallish common factors then rounding can become significant.

christopherlam: We're adding the Debit/Credit subtotal numbers here. So best to leave it as GNC-DENOM-AUTO GNC-RND-ROUND?

gjanssens:
Op zaterdag 25 november 2017 04:49:06 CET schreef christopherlam:
> PS As far as rewriting history... well this will be a whole new learning
> curve! :)

True. Take it at your own pace and feel free to ask for guidance. I'm happy to 
help out.

gjanssens:
Op zaterdag 25 november 2017 00:55:57 CET schreef christopherlam:
> Hi. Thank you for going through with a fine comb. My aim is to seek feedback
> on new "features" eg string/regex matchers/filters, multicolumn display,
> while learning new scheme idioms.
> 
> More importantly I'm really needing to confirm my functional changes e.g.:
> 
Well, you're stretching my accounting experience as well here :) But I'll 
insert my thoughts below;

> - strategy on reversing signs + debit - credit is correct?
> 
From real testing I think it's mostly ok. The one glitch I see is in the 
subtotals while displaying debit/credit columns. While in that form all 
transactions have positive numbers, the subtotals can still be displayed as 
negative numbers (income for example). I would expect the debit/credit idiom 
to be extended to the subtotals in that case or if that's not an option I 
would to at least expect to be able to control sign reversals on the 
subtotals.

A few side considerations that came up while thinking about this:
1. There is a global "Revert balanced accounts" option in Edit->Preferences-
>Accounts. It may be worth to evaluate whether the transaction report should 
just follow that option rather than have its own.

2. There is also a global option "Use formal accounting labels" under the same 
preference tab. When enabled all account registers will use debit/credit 
labels. When disabled the registers use labels that are better understood by 
people with no accounting background (charge/income, deposit/withdrawal, tot 
buy/sell,  payment/bill, invoice/payment,...). It might be a nice extra 
improvement if the report would follow this general preference as well (in 
double column mode).

> - my new subtotal strategy which adds internal values rather than (reversed)
> display amounts.
> 
Definitely an improvement and less error prone. I agree it's better to work 
with the internal amounts. The same goes for rounding in fact. We should 
really only round right before display or while storing numbers. John and I 
came to that conclusion a few months back when thinking about the new 
GncNumeric code. Because it's a fresh idea, almost none of the current code 
works like that though...

> - my custom sorter which overrides qof-query sorting strategy but is IMHO
> better in cases described in
> http://gnucash.1415818.n4.nabble.com/transaction-scm-date-subtotal-options-> calling-wizards-td4693895.html -- the custom sorter should work better but
> is still a hack; and if this is approved I'd refactor its strategy into the
> main sortkeys. Also "Use custom sorter" isn't a good option to present. Do
> I keep it? Rename it? Wait for qof-query sorter to upgrade its strategy
> (when sort=date is chosen and subtotal=week for example it'll always
> strictly sort by daily dates; we want it to keep weeks alone!)?

I would prefer to see this solved in qof really for a couple of reasons:
- firstly you wrote somewhere it's slightly slower. Depending on the 
performance of the PC and the amount of data to work with this "slightly" can 
become significant.
- qof is the code responsible for queries. In the course of the big 
refactoring it will be thoroughly reworked (probably close to or completely 
being replaced with sql queries) which should make it even more efficient at 
requesting the exact data in the proper sorting/grouping.

Having said all that, qof is not capable yet of sorting the way you need, so 
until that time your code gives better results for certain sorting/grouping 
combinations. How hard would it be to use your custom sorter only for those 
cases where qof sorting fails to produce the proper result ? That would give 
the best performance where qof can handle it, and guarantees proper results 
where qof can't.

I would like to avoid making it an option if at all possible.
> 
> - I know the infobox is hacking English into scheme and hard-codes HTML ;)
> this is also experimental and unsure whether to keep. And whether to add
> toggle in General. Feedback?
> 
I don't have much use for this, but I see the other accounting program I use 
also prints the filtering options on the report. So I believe there is value 
in this option. I can't decide though whether to add it under Display rather 
than General or not.

> - multicolumn data display is an actual change in code whereas most of the
> above are refactoring/embelishments - seeking feedback- Add new column
> ideas? Remove?
> 
I presume this refers to the ability to show the amount in the original 
currency ? That's a nice addition imo. Further refinements could be:
1. make this value follow the single/double column display as well. Currently 
if amount is set to double, the original amount is still in single mode and 
the user has no way of controlling sign reversals on that column.
2. It would definitely be nicer if the column headings for the original 
amounts were using the same labels (debit/credit or amount) with a currency or 
commodity mentioned between brackets. Considering the labels are only added 
once at the top, it's probably easier to add the report currency to the column 
in the chosen report currency rather than the original currency (which can be 
different from transaction to transaction) in the original currency column.

> These are actual functional changes which can cause long-term users to
> wonder why suddenly the reports have changed (in a good way I think).
> 
I consider all of them as improvements.

> With regards to backward-compatible options - there's code somewhere within
> helper modules that will dynamically rename options
> https://github.com/Gnucash/gnucash/blob/037c93faea77f51a36d7391d0f3ef403b50
> 36843/libgnucash/app-utils/options.scm#L1685 - I don't think we have to
> necessarily add it there, but it's food for thought at least. I'll need to
> think how back/forward compatibility can be maintained.
> 
I had forgotten about that code. I do think it's a good start and surely 
useful to manage it in one central place. For the specific cases we have here 
it's lacking some flexibility though. From what I understand it can only 
handle a name change for options. It can't detect moving options to another 
section, nor can it deal with changes in multi-choice option values (consider 
removing the 'exact-time value, with 'date being the new equivalent)


> Overall while I think the functional changes are positive and useful, I'm
> not sure whether they fit in the long-term strategy on what the TR should
> be. I don't necessarily wish to make it my exclusive project!

There is no long term road map for the TR that I know of. I'm happy you are 
proposing improvements and actually implement those. I also think most of them 
are well considered and address real practical use cases. So I believe you are 
on the right track.

gjanssens:
Oh, and on the subject of moving options: what would you think about moving the Void transactions option to the filter tab as well  (assuming we find a user transparant way to migrate saved options) ?

christopherlam:
>
>
> > - strategy on reversing signs + debit - credit is correct?
> >
> From real testing I think it's mostly ok. The one glitch I see is in the
> subtotals while displaying debit/credit columns. While in that form all
> transactions have positive numbers, the subtotals can still be displayed
> as
> negative numbers (income for example). I would expect the debit/credit
> idiom
> to be extended to the subtotals in that case or if that's not an option I
> would to at least expect to be able to control sign reversals on the
> subtotals.
>
>
In my understanding this is currently impossible. Because the TR allows,
for example, selection of *all* accounts, and the subtotals will
obligatorily receive split-values from all account types. It can't detect
the account type to consider reversing. Or maybe this can be hacked but
will require an upgraded collector system. As you have noticed, selecting
all accounts, and enabling subtotals will lead to the peculiarity that
income subtotals will be negative. The upside is that the grand total will
reach $0.00 when all accounts are selected, or if I've chosen "income &
expense accounts" it will show an indication of "profit" (with negative
sign...)... or if "assets & liabilities" accounts it should show an
indication of "net worth". For this to happen, the subtotal collectors must
acquire and process the unreversed split values... and if these split
values are negative, they will be displayed as such.

If the accounts chosen are tightly controlled, therefore the account types
can be inferred, *then* I feel the subtotals can be reversed.
Perhaps this will lead to complaints but I can't think how to fix this
unless the account types are specifically limited in options.

A few side considerations that came up while thinking about this:
> 1. There is a global "Revert balanced accounts" option in
> Edit->Preferences-
> >Accounts. It may be worth to evaluate whether the transaction report
> should
> just follow that option rather than have its own.
>

Will investigate.


> 2. There is also a global option "Use formal accounting labels" under the
> same
> preference tab. When enabled all account registers will use debit/credit
> labels. When disabled the registers use labels that are better understood
> by
> people with no accounting background (charge/income, deposit/withdrawal,
> tot
> buy/sell, payment/bill, invoice/payment,...). It might be a nice extra
> improvement if the report would follow this general preference as well (in
> double column mode).
>
>
As above, this will be fine if the account types are tightly controlled.
The headers (debit/credit/original/running balance) are displayed once per
table (and also at the top of every page according to the HTML renderer
<g>) and cannot be changed mid-table.


> > - my new subtotal strategy which adds internal values rather than
> (reversed)
> > display amounts.
> >
> Definitely an improvement and less error prone. I agree it's better to
> work
> with the internal amounts. The same goes for rounding in fact. We should
> really only round right before display or while storing numbers. John and
> I
> came to that conclusion a few months back when thinking about the new
> GncNumeric code. Because it's a fresh idea, almost none of the current
> code
> works like that though...
>
>
Actually the rounding strategies elude me... my understanding of GncNumeric
was that it allows arbitrary precision rational numbers but with @jrall's
comment that large numbers can be rounded off I'm slightly confused. Let me
know which rounding strategy is appropriate for the various adders and I'll
amend.


> > - my custom sorter which overrides qof-query sorting strategy but is IMHO
> > better in cases described in
> > http://gnucash.1415818.n4.nabble.com/transaction-scm-
> date-subtotal-options-> calling-wizards-td4693895.html -- the custom
> sorter should work better but
> > is still a hack; and if this is approved I'd refactor its strategy into
> the
> > main sortkeys. Also "Use custom sorter" isn't a good option to present.
> Do
> > I keep it? Rename it? Wait for qof-query sorter to upgrade its strategy
> > (when sort=date is chosen and subtotal=week for example it'll always
> > strictly sort by daily dates; we want it to keep weeks alone!)?
>
> I would prefer to see this solved in qof really for a couple of reasons:
> - firstly you wrote somewhere it's slightly slower. Depending on the
> performance of the PC and the amount of data to work with this "slightly"
> can
> become significant.
>
- qof is the code responsible for queries. In the course of the big
> refactoring it will be thoroughly reworked (probably close to or
> completely
> being replaced with sql queries) which should make it even more efficient
> at
> requesting the exact data in the proper sorting/grouping.
>
> Having said all that, qof is not capable yet of sorting the way you need,
> so
> until that time your code gives better results for certain
> sorting/grouping
> combinations. How hard would it be to use your custom sorter only for
> those
> cases where qof sorting fails to produce the proper result ? That would
> give
> the best performance where qof can handle it, and guarantees proper
> results
> where qof can't.
>
>
Will also investigate. This could be worked on, by detecting if
prime-sortkey or sec-sortkey is a date-subtotal-type... This means that the
whole custom sorter code will need to remain as is (a block which can be
removed in the future) until qof-query is upgraded.


> I would like to avoid making it an option if at all possible.
>

+1

>
> > - I know the infobox is hacking English into scheme and hard-codes HTML
> ;)
> > this is also experimental and unsure whether to keep. And whether to add
> > toggle in General. Feedback?
> >
> I don't have much use for this, but I see the other accounting program I
> use
> also prints the filtering options on the report. So I believe there is
> value
> in this option. I can't decide though whether to add it under Display
> rather
> than General or not.
>
> > - multicolumn data display is an actual change in code whereas most of
> the
> > above are refactoring/embelishments - seeking feedback- Add new column
> > ideas? Remove?
> >
> I presume this refers to the ability to show the amount in the original
> currency ? That's a nice addition imo. Further refinements could be:
> 1. make this value follow the single/double column display as well.
> Currently
> if amount is set to double, the original amount is still in single mode
> and
> the user has no way of controlling sign reversals on that column.
> 2. It would definitely be nicer if the column headings for the original
> amounts were using the same labels (debit/credit or amount) with a
> currency or
> commodity mentioned between brackets. Considering the labels are only
> added
> once at the top, it's probably easier to add the report currency to the
> column
> in the chosen report currency rather than the original currency (which can
> be
> different from transaction to transaction) in the original currency column.
>
>
Will consider both. In my experimentation, the purpose of "Original" column
was to print the internal-representation of split-value, for debugging or
for understanding reverse-signs. But it can be tweaked as above.


> > These are actual functional changes which can cause long-term users to
> > wonder why suddenly the reports have changed (in a good way I think).
> >
> I consider all of them as improvements.
>

Thank you :)


> > With regards to backward-compatible options - there's code somewhere
> within
> > helper modules that will dynamically rename options
> > https://github.com/Gnucash/gnucash/blob/037c93faea77f51a36d7391d0f3ef4
> 03b50
> > 36843/libgnucash/app-utils/options.scm#L1685 - I don't think we have to
> > necessarily add it there, but it's food for thought at least. I'll need
> to
> > think how back/forward compatibility can be maintained.
> >
> I had forgotten about that code. I do think it's a good start and surely
> useful to manage it in one central place. For the specific cases we have
> here
> it's lacking some flexibility though. From what I understand it can only
> handle a name change for options. It can't detect moving options to
> another
> section, nor can it deal with changes in multi-choice option values
> (consider
> removing the 'exact-time value, with 'date being the new equivalent)
>
>
I'll see what I do to hack it :)


>
> > Overall while I think the functional changes are positive and useful, I'm
> > not sure whether they fit in the long-term strategy on what the TR should
> > be. I don't necessarily wish to make it my exclusive project!
>
> There is no long term road map for the TR that I know of. I'm happy you
> are
> proposing improvements and actually implement those. I also think most of
> them
> are well considered and address real practical use cases. So I believe you
> are
> on the right track.
>

> Oh, and on the subject of moving options: what would you think about
moving the Void transactions option to the filter tab as well (assuming we
find a user transparant way to migrate saved options) ?

Agreed, will investigate.

Thank you for thoughts as above -- they are definitely valuable and will
influence strategy.

I can imagine the final TR could look like a whole lot of definitions
consisting of strings, alists of alists, formulas, and a tiny 'interpreter'
for the data presented above. Woe to the next hacker :-O

jralls:
> Actually the rounding strategies elude me... my understanding of GncNumeric
was that it allows arbitrary precision rational numbers but with @jrall's
comment that large numbers can be rounded off I'm slightly confused. Let me
know which rounding strategy is appropriate for the various adders

Not arbitrary precision. We considered that but found that it's too slow because it uses lots of allocations, so we went with 128-bit integers in classes GncInt128 and GncRational. GncNumeric is a 64-bit class that passes all of its computations to GncRational which then rounds the results back to 64-bit if necessary. gnc_numeric is a C wrapper around GncNumeric and I think that's the only facility that's wrapped for Scheme at present.

Scheme on the other hand does provide an arbitrary-precision rational type, so you could experiment with importing a GncNumeric into it and doing your calculations there. Be sure to test with a large dataset to make sure that you don't cause a performance problem.

The various rounding strategies are listed with rather sparse explanation in [the Doxygen docs](https://code.gnucash.org/docs/MASTER/gnc-numeric_8h.html) at "Arguments Standard Arguments to Most Functions". GNC_HOW_RND_ROUND is probably best for your purposes: It rounds to the nearest even number and so is consistent for both positive and negative numbers. Don't use GNC_HOW_DENOM_SIGFIGS. It doesn't actually work as advertised.

gjanssens:
Op zaterdag 25 november 2017 15:33:52 CET schreef christopherlam:
> > > - strategy on reversing signs + debit - credit is correct?
> > 
> > I would expect the debit/credit idiom
> > to be extended to the subtotals in that case or if that's not an option I
> > would to at least expect to be able to control sign reversals on the
> > subtotals.
> 
> In my understanding this is currently impossible. Because the TR allows,
> for example, selection of *all* accounts, and the subtotals will
> obligatorily receive split-values from all account types. It can't detect
> the account type to consider reversing. Or maybe this can be hacked but
> will require an upgraded collector system. As you have noticed, selecting
> all accounts, and enabling subtotals will lead to the peculiarity that
> income subtotals will be negative. The upside is that the grand total will
> reach $0.00 when all accounts are selected, or if I've chosen "income &
> expense accounts" it will show an indication of "profit" (with negative
> sign...)... or if "assets & liabilities" accounts it should show an
> indication of "net worth". For this to happen, the subtotal collectors must
> acquire and process the unreversed split values... and if these split
> values are negative, they will be displayed as such.
> 
> If the accounts chosen are tightly controlled, therefore the account types
> can be inferred, *then* I feel the subtotals can be reversed.
> Perhaps this will lead to complaints but I can't think how to fix this
> unless the account types are specifically limited in options.
> 
Controlling sign reversals was actually my second choice :)
I would prefer to have the subtotal appear in the appropriate debit or credit 
column in case double amounts was chosen. You are right that adding sign 
reversal would be challenging in all cases except where grouping was done one 
accounts.

<snip>

> > 2. There is also a global option "Use formal accounting labels" under the
> > same
> > preference tab. When enabled all account registers will use debit/credit
> > labels. When disabled the registers use labels that are better understood
> > by
> > people with no accounting background (charge/income, deposit/withdrawal,
> > tot
> > buy/sell, payment/bill, invoice/payment,...). It might be a nice extra
> > improvement if the report would follow this general preference as well (in
> > double column mode).
> 
> As above, this will be fine if the account types are tightly controlled.
> The headers (debit/credit/original/running balance) are displayed once per
> table (and also at the top of every page according to the HTML renderer
> <g>) and cannot be changed mid-table.
> 
Right. The choice of a single html table already makes this impossible. I 
should have thought of that because we once received a patch splitting the 
table into several smaller ones. I found the result much less pleasing because 
the table widths became quite inconsistent. So never mind this suggestion.

<snip>

> > How hard would it be to use your custom sorter only for those
> > cases where qof sorting fails to produce the proper result ? That would
> > give
> > the best performance where qof can handle it, and guarantees proper
> > results
> > where qof can't.
> 
> Will also investigate. This could be worked on, by detecting if
> prime-sortkey or sec-sortkey is a date-subtotal-type... This means that the
> whole custom sorter code will need to remain as is (a block which can be
> removed in the future) until qof-query is upgraded.
> 

<snip>

> I can imagine the final TR could look like a whole lot of definitions
> consisting of strings, alists of alists, formulas, and a tiny 'interpreter'
> for the data presented above. Woe to the next hacker :-O

That sounds like it's conceptually heading in the direction of the eguile 
code: it is a system to separate logic from presentation. A templating system 
written in guile if you will. I'm just mentioning it should you be interested 
in transforming the TR into an eguile based report (not a requirement).

christopherlam:
Checklist:
- [x] use custom sorter if and only if sortkeys require it
- [x] instead of merging arbitrary columns, add dual-columns only and place in correct column
- [x] restore rounding strategy to GNC_HOW_RND_ROUND
- [x] moved void-status filter to filter tab
- [ ] add backward compatibility shim somewhere for account-substring-filter and void-filter and 'exact time removal
- [ ] add dual column for 'original currency amount' and enable sign-reversal
- [ ] explore using global prefs for sign reversals
- [ ] add commodity to data columns if common-currency is ticked. This will require some refactoring unfortunately to access the headings in calculated-cells.

christopherlam:
Now just need upgraded options.scm to handle option sections being moved!
I'm not sure how to handle sortkey previously being 'exact-time - this pref isn't kept in reports.

gnucash/report/standard-reports/transaction.scm
```diff
@@ -1168,7 +1141,7 @@ tags within description, notes or memo. ")
                                                currency
                                                (gnc-numeric-neg amount))))))
                    (vector (if (and reverse?
-                                    (member (xaccAccountGetType account) account-types-to-reverse))
```
christopherlam: Alternatively we can keep this UI option and add the default "Use Global Preference" which will then pull from global?

christopherlam:
I think this madness can take a break now. It seems to work well on my 2.6.18 and unstable.
Not sure how to handle 'exact-time.
I'd think I've handled N_ and _ correctly.

christopherlam:
You'll have seen the increased number of commits. These relate to the previous work-in-progress during the initial refactoring. Unfortunately it's practically impossible to deconstruct into individual working changes.

I've also noticed the following living standard for multilevel headings. I'll continue experimenting in a separate branch. This may be worth while considering for account-type subheadings... https://www.w3.org/WAI/tutorials/tables/multi-level/

christopherlam:
I did it :) Can be styled; can add toggle to switch off, of course... "Sorting/Show informal account names for double column mode".  Adds clutter. Vote yay/nay? Optional toggle?

But I think it really only works well for sortkeys 'account-name or 'account-code -- IMHO using sortkeys 'other-acc-name or 'other-acc-code has always given spurious results, displaying wrong signs etc. (Try display=single, switching acc-name/other-account-name sortkeys). I've disabled for them. I'd be keen to remove them altogether myself.

image

christopherlam:
New Feature! Adds 0-2 blank columns to aid alignment, depending on subtotal options.

image

christopherlam:
Reconciliation report... Could be formalised here as a supplementary definition, or let users learn to play with the options until they find their preferred settings?

Eventually I should be able to merge the GST report here too :-D and will take extra effort to maintain clear separation for maintainability. Alternatively if I can understand how to generate another scm which will subclass the tr then it'll be cleaner.

christopherlam:
This branch obsolete.

PR #230, opened 2017-11-12
Fix dense calendar on Windows

Bob-IT:
This PR fixes Gnucash crashing when in Windows you click on a date that does not have any scheduled transactions marked. It seems to stem from the treeview being empty and not being displayed so if there are no entries I add a blank. There are also a couple of changes to help improve when the popup is displayed.
Bob

### DIFF
--- a/gnucash/gnome-utils/gnc-dense-cal.h
+++ b/gnucash/gnome-utils/gnc-dense-cal.h
@@ -59,6 +59,7 @@ struct _GncDenseCal
     GtkWindow *transPopup;
     gint screen_width;
     gint screen_height;
+    gint doc;
 
     gint min_x_scale;
     gint min_y_scale;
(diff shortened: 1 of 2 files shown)


gnucash/gnome-utils/gnc-dense-cal.h
```diff
@@ -57,6 +57,9 @@ struct _GncDenseCal
 
     gboolean showPopup;
     GtkWindow *transPopup;
+    gint screen_width;
+    gint screen_height;
+    gint doc;
```
CarwynNelson: I'm not very familiar at all with the gnucash codebase, but whilst reading through this pull request I can't figure out what `doc` means or stands for.

I guess my first question is what does `doc` stand for or refer to? And my second question would be if it has a "long-form" name would it make sense to use that instead?

jralls:
Looks like "day on calendar" might be a possible expansion. It's pretty clear from the implementation code (e.g. line 1220)  that it's the number of days from the first day of the month.

PR #231, opened 2017-11-16
report: fix default colors incorrectly shown as transparent

jmarino:
In all the scheme files that create a color option with "gnc:make-color-option" the default colors are being passed as a list of rgba values. For example, white color is given as: (list #xff #xff #xff 0)
Where the last value 0 is meant as transparency 0, i.e., fully opaque.

However, gtk uses a different convention. From the docs for GdkRGBA: "alpha: The opacity of the color from 0.0 for completely translucent to 1.0 for opaque"
Because of this disagreement, gui color buttons display a transparent checkerboard pattern instead of the intended default color.

In this commit I convert from the gnucash convention to the gtk convention when setting or getting
colors. Another way to fix this would be to change the gnucash alpha convention to match the gtk convention, changing all the default color alpha values from 0 to #xff. I don't know which solution would be more preferable.

There aren't many users of the color-option, only the example report "hello-world.scm", the standard report "price-scatter.scm", and some "stylesheets-*.scm" files appear to use it.

### DIFF
--- a/gnucash/report/report-system/doc/report-html.txt
+++ b/gnucash/report/report-system/doc/report-html.txt
@@ -363,7 +363,7 @@ We have three different kinds of style control information available:
             (gnc:make-color-option
               (_ "Style Sheet Options")
 	      (_ "Background Color") "a" (_ "Background color for reports.")
-	      (list #xff #xff #xff 0)
+	      (list #xff #xff #xff #xff)
 	      255 #f))
           options))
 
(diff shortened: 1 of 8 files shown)


gjanssens:
Thanks for the PR. Considering we have changed from GdkColor to GdkRGBA in this development series it makes more sense to me to adjust the default alpha values to #ff than to insert a conversion function. The latter requires the developer to keep two color definitions in mind, depending on where to change a color. Let's just stick to the GdkRGBA convention everywhere instead. Can you provide the alternative patch ?

jmarino:
Changed the patch to use `alpha=#xff` in all instances of `gnc:make-color-option`.
Also, rebased from current unstable.

gjanssens:

PR #232, opened 2017-11-18
Bugzilla 790526 Correct weeknum calculator

christopherlam:
This change will fix 'num-of-weeks-since-1/jan/1970' which formerly used quotient to remove
the fractional part of the division. For negative values of num-of-weeks, the number is truncated
in the wrong direction (i.e. towards 0). This change uses floor instead to ensure the num-of-weeks
found is the nearest integer LESS than the fractional number.

Needs testing to ensure the weeks produced match the previous calculator for dates > 1/1/1970.


jralls:
Are you working on a test?

christopherlam:
Ok will fix this this evening :)

jralls:
That got it. Thanks.

PR #233, opened 2017-11-22
Fix bayes

limitedAtonement:


libgnucash/engine/kvp-frame.hpp
```diff
@@ -204,11 +204,15 @@ struct KvpFrameImpl
      * @return The value at the key or nullptr.
      */
     KvpValue* get_slot(Path keys) const noexcept;
-    /** Convenience wrapper for std::for_each, which should be preferred.
+
+    void for_each_slot(void (*)(char const *key, KvpValue*, void*data), void*data) const noexcept;
```
gjanssens: As far as I can understand you declare this function but removed its definition (in kvp-frame.cpp). Yet the function is still called in qofinstance-p.h:187
I'm surprised this even compiles, but perhaps you can explain ?

libgnucash/engine/kvp-value.cpp
```diff
@@ -161,53 +159,66 @@ struct to_string_visitor : boost::static_visitor<void>
...
         {
             output << tmp1;
             g_free(tmp1);
         }
-        output << ")";
+        else
+        {
+            output << "(null)";
+        }
+        output << " (timespec)";
```
gjanssens: This should probably be ```output << " (gnc_numeric)";```

limitedAtonement:
It turned out to be an error of mine in stack string allocation size: a buffer overflow. I'm glad fedora caught it so reproducibly! I added another commit that hopefully addresses the problem.

Also, you mentioned that you're not so familiar with the import map stuff, so I thought I would give an introduction to what I know:

There are always two accounts to which reference is being made: the account into which transactions are being imported, and the "other side" which we're trying to use the import map to predict. The import maps are always attached to the accounts into which import is happening. For instance, after importing transactions for `Assets:Chequeing`, that account will have an import map KVP attached to it to assist future imports.

You can see this being used at `import-backend.c`:558 (the "find" parts are at line 487).

**REGULAR MATCHING STRATEGY**

There is an Account GUID at this location in the KVP: `import-map` -> `[category]` -> `key`. This is an hierarchical KVP. The category either corresponds to "description" or "memo", and the key will be the description or memo itself. It looks like matches are always made in full text on the memo or description. Naturally, the Account GUID placed at that kvp location is the account requested by the user.

For instance, when importing against the chequeing account, when a transaction with description "Kroger" is matching with `Expenses:Groceries`, the following entry is created:

Kvp frame `import-map` -> `DESC` -> `Kroger` containing the guid for `Expenses:Groceries`.

**BAYES**

The KVP for Bayes looks like this:

`import-map-bayes/token/32-digit account GUID` and at this location, a `numeric` is stored. 

This KVP is flat -- not hierarchical. It's flat to fix bugzilla bug 761682. In the bayes model, each time you import transactions, several "tokens" are created relevant to the transaction. Tokens like the day of week, words in the description, etc. (Tokenization is done at `import-backend.c`:405.) For each of these tokens, an import map KVP record is created. Naturally, the account GUID corresponds to the account with which this transaction is matched. The numeric at this location corresponds to how many times this import has happened this way; thus this is a learning algorithm. The more often you import a transaction with tokens like "Bridge" and "Monday" (for your Monday night bridge night fees), the better the bayes import map will be able to predict that the split will be going into `Expenses:Bridge Fees`.


That's a quick-and-dirty. Let me know if I can be of any further help.

libgnucash/engine/kvp-frame.hpp
limitedAtonement: I lost track of this one. Let me see what I can see.

limitedAtonement: Good ol' C++ (probably C, too): only pay for what you use. It's called in `qofinstance-p.h`:187 as you say, but since *that* function (`qof_instance_foreach_slot_temp`) is never called, the definition is never needed. Thank you for finding this. I'll commit a proper removal presently.

libgnucash/engine/Account.cpp
```diff
@@ -5234,260 +5157,157 @@ struct account_probability
...
+    /*By convention, the key ends with the account GUID.*/
+    this_account.account_guid = account_guid.substr(account_guid.size() - GUID_ENCODING_LENGTH);
+    this_account.token_count = value->get<int64_t>();
+    tokenInfo.accounts.push_back(this_account);
 }
 
+/** We scale the probability values by PROBABILITY_FACTOR.
+  ie. with PROBABILITY_FACTOR of 100000, 10% would be
+  0.10 * 100000 = 10000 */
+#define PROBABILITY_FACTOR 100000
```
jralls: I'd prefer ```static constexpr int probability_factor 10000;``` .

libgnucash/engine/Account.cpp
```diff
@@ -5591,138 +5398,76 @@ build_bayes_layer_two (const char *key, const GValue *value, gpointer user_data)
...
 
-    guid = g_new (GncGUID, 1);
-
-    if (string_to_guid (key, guid))
-        map_account = xaccAccountLookup (guid, book);
-
-    g_free (guid);
-
-    imapInfo_node = g_malloc(sizeof(*imapInfo_node));
+    imapInfo_node = static_cast <imap_info*> (g_malloc(sizeof(*imapInfo_node)));
```
jralls: struct imap_info is typedeffed to GncImapInfo. Using that typename is preferable to `imap_info`.

libgnucash/engine/Account.cpp
```diff
@@ -5797,11 +5535,12 @@ gnc_account_delete_map_entry (Account *acc, char *full_category, gboolean empty)
 /*******************************************************************************/
 
 static gchar *
-look_for_old_separator_descendants (Account *root, gchar *full_name, const gchar *separator)
+look_for_old_separator_descendants (Account *root, gchar const *full_name, const gchar *separator)
 {
     GList *top_accounts, *ptr;
     gint   found_len = 0;
     gchar  found_sep;
+    gchar * new_name = nullptr;
```
jralls: This is a c++ file, no need to stick all the identifiers at the top.

common/test-core/unittest-support.h
```diff
@@ -178,6 +178,14 @@ GSList *test_log_set_fatal_handler (GSList *list, TestErrorStruct *error,
...
 void test_free_log_handler (gpointer item);
 
+/**
+ * Check that the user_data error message is a substring of the
+ * actual error otherwise assert.  Displays the error (and asserts
+ * if G_LOG_FLAG_FATAL is TRUE) if NULL is passed as user_data,
+ * but a NULL or 0 value member matches anything.
+ */
+gboolean test_checked_substring_handler (const char *log_domain, GLogLevelFlags log_level,
+                                         const gchar *msg, gpointer user_data);
```
jralls: Good idea!

lawsa:
That's right, other kvp uses are still hierarchical rather than flat. I wanted to get this change implemented quickly. There is no delimiter character: kvp keys are not parsed at all now for delimiters but are stored intact.

libgnucash/engine/kvp-frame.hpp
```diff
@@ -187,28 +173,51 @@ struct KvpFrameImpl
...
-    KvpValue* get_slot(Path keys) const noexcept;
-    /** Convenience wrapper for std::for_each, which should be preferred.
+    KvpValue* get_slot(Path keys) noexcept;
+
+    /** The function should be of the form:
+     * <anything> func (char const *, KvpValue *, data_type &);
+     * Do not pass nullptr as the function.
+     */
+    template <typename func_type, typename data_type>
+    void for_each_slot_temp(func_type const &, data_type &) const noexcept;
```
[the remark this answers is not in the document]
lawsa: I think the problem is that `for_each_slot` is `extern "C"`, so it can't be overloaded.

libgnucash/engine/kvp-frame.hpp
```diff
@@ -218,8 +227,69 @@ struct KvpFrameImpl
...
     private:
     map_type m_valuemap;
+
+    KvpFrame * get_child_frame_or_nullptr (Path const &) noexcept;
+    KvpFrame * get_child_frame_or_create (Path const &) noexcept;
+    void flatten_kvp_impl(std::vector <std::string>, std::vector <std::pair <std::vector <std::string>, KvpValue*>> &) const noexcept;
+    KvpValue * set_impl (std::string const &, KvpValue *) noexcept;
 };
 
+template<typename func_type>
```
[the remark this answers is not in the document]
lawsa: Thank you for making reference to that. I'll review the coding standards.

libgnucash/engine/qofinstance-p.h
```diff
@@ -113,23 +116,24 @@ void qof_instance_set_idata(gpointer inst, guint32 idata);
...
+ * frames will be created if necessary. Commits the change to the QofInstance.
  * @param inst: The QofInstance on which to set the value.
- * @param key: The key for the slot or '/'-delimited path
+ * @param key: The path to the slot.
  * @param value: A GValue containing an item of a type which KvpValue knows
- * how to store.
+ *           how to store.
  */
-void qof_instance_set_kvp (QofInstance *inst, const gchar *key, const GValue *value);
+void qof_instance_set_var_kvp (QofInstance *, GValue const * value, unsigned count, ...);
```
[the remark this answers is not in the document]
lawsa: I started by adding the `_var_` function, then worked through changing all uses to that, then deleted the original. After that, I renamed the function back (to get rid of `_var_`), so all should be okay now.

limitedAtonement:
That's right, other kvp uses are still hierarchical rather than flat. I wanted to get this change implemented quickly. There is no delimiter character: kvp keys are not parsed at all now for delimiters but are stored intact.

libgnucash/engine/kvp-frame.hpp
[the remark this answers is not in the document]
limitedAtonement: I think the problem is that for_each_slot is extern "C", so it can't be overloaded.

libgnucash/engine/kvp-frame.hpp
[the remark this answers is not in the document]
limitedAtonement: Thank you for making reference to that. I'll review the coding standards.

libgnucash/engine/Account.cpp
```diff
@@ -3067,7 +3053,7 @@ gnc_account_get_full_name(const Account *account)
 
     /* Get all the pointers in the right order. The root node "entry"
      * becomes the terminating NULL pointer for the array of strings. */
-    names = g_malloc(level * sizeof(gchar *));
+    gchar* names[level*sizeof(gchar*)];
```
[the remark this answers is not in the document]
limitedAtonement: Yeah, not sure what went wrong here. I'll take care of it.

libgnucash/engine/qofinstance-p.h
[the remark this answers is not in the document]
limitedAtonement: Oops, I thought I had put those back. I just pushed a new commit for this purpose.

libgnucash/engine/Account.cpp
```diff
@@ -5534,29 +5535,23 @@ gnc_account_delete_map_entry (Account *acc, char *full_category, gboolean empty)
...
-
+    PINFO("Incoming full_name is '%s', current separator is '%s'", full_name.c_str (), separator);
     /* Go through list of top level accounts */
     for (ptr = top_accounts; ptr; ptr = g_list_next (ptr))
     {
         const gchar *name = xaccAccountGetName (static_cast <Account const *> (ptr->data));
-
         // we are looking for the longest top level account that matches
-        if (g_str_has_prefix (full_name, name))
+        if (g_str_has_prefix (full_name.c_str (), name))
```
jralls: Come on. If you're going to introduce a C++ feature, **use it!** Possibilities:
```
if (full_name.find(name) == full_name.begin())
```
or
```
auto const name == std::string(xaccAccountGetName(static_cast<Account const*>(ptr-data));
if (full_name.substr(0, name.length()) == name)
```
Not a blocker for merging, just a plea to write C++ when you're changing code.

limitedAtonement: Nobody wants to get rid of `g_` from the code base more than me ;-) I'm in a bit of an hurry, and I'm not considering this "the Account C++ rewrite". Be comforted by the fact that I'm with you heart and soul on this point.

libgnucash/engine/gnc-features.c
```diff
@@ -46,6 +46,7 @@ static gncFeature known_features[] =
     { GNC_FEATURE_KVP_EXTRA_DATA, "Extra data for addresses, jobs or invoice entries (requires at least GnuCash 2.6.4)" },
     { GNC_FEATURE_BOOK_CURRENCY, "User specifies a 'book-currency'; costs of other currencies/commodities tracked in terms of book-currency (requires at least GnuCash 2.7.0)" },
     { GNC_FEATURE_GUID_BAYESIAN, "Use account GUID as key for Bayesian data (requires at least GnuCash 2.6.12)" },
+    { GNC_FEATURE_GUID_FLAT_BAYESIAN, "Use account GUID as key for bayesian data and store KVP flat" },
```
gjanssens: Like in 2.6.19, this should refer to the first version of gnucash that does support the feature.

limitedAtonement: Ah yes, thanks for pointing this out.

limitedAtonement:
Oof; pushed again.

gjanssens:
Thanks!

PR #234, opened 2017-11-22
GUID/Flat bayes handling in 2.6

limitedAtonement:
With 2.8 soon able to write guid and flat bayes kvps, 2.6 will also need
to be able to read them. Due to the nature of the change, though, 2.6
will also be able to write guid and flat bayes kvps. 2.8 is able to read
2.6-type bayes kvps, and will convert them to a guid/flat model. With this
commit, 2.6 can read and write in this model, but is not able to convert
a database between the two.

### DIFF
--- a/src/core-utils/gnc-features.c
+++ b/src/core-utils/gnc-features.c
@@ -45,6 +45,7 @@ static gncFeature known_features[] =
     { GNC_FEATURE_NUM_FIELD_SOURCE, "User specifies source of 'num' field'; either transaction number or split action (requires at least GnuCash 2.5.0)" },
     { GNC_FEATURE_KVP_EXTRA_DATA, "Extra data for addresses, jobs or invoice entries (requires at least GnuCash 2.6.4)" },
     { GNC_FEATURE_GUID_BAYESIAN, "Use account GUID as key for Bayesian data (requires at least GnuCash 2.6.12)" },
+    { GNC_FEATURE_GUID_FLAT_BAYESIAN, "Use account GUID as key for bayesian data and store KVP flat" },
     { NULL },
 };
 
@@ -159,4 +160,35 @@ void gnc_features_set_used (QofBook *book, const gchar *feature)
     qof_book_kvp_changed (book);
 
 
+}
+
+struct check_feature
+{
+    gchar const * checked_feature;
+    gboolean found;
+};
+
+static void
+gnc_feature_check_feature_cb (gchar const * key, KvpValue * value, gpointer data)
+{
+    struct check_feature * check_data = data;
+    if (!g_strcmp0 (key, check_data->checked_feature))
+        check_data->found = TRUE;
+}
+
+gboolean gnc_features_check_used (QofBook *book, const gchar * feature)
+{
+    KvpFrame *frame = qof_book_get_slots (book);
+    KvpValue *value;
+    struct check_feature check_data = {feature, FALSE};
+    /* Setup the known_features hash table */
+    gnc_features_init();
+    g_assert(frame);
+    value = kvp_frame_get_value(frame, "features");
+    if (!value) return FALSE;
+    frame = kvp_value_get_frame(value);
+    g_assert(frame);
+    /* Iterate over the members of this frame for unknown features */
+    kvp_frame_for_each_slot (frame, &gnc_feature_check_feature_cb, &check_data);
+    return check_data.found;
 }
--- a/src/import-export/import-match-map.c
+++ b/src/import-export/import-match-map.c
@@ -33,6 +33,7 @@
 #include "import-match-map.h"
 #include "gnc-ui-util.h"
 #include "gnc-engine.h"
+#include "gnc-features.h"
 
 /********************************************************************\
  *   Constants   *
@@ -173,8 +174,8 @@ void gnc_imap_add_account (GncImportMatchMap *imap, const char *category,
 
 struct account_token_count
 {
-    char* account_name;
-    gint64 token_count; /**< occurances of a given token for this account_name */
+    char * account_guid;
+    gint64 token_count; /**< occurances of a given token for this account_guid */
 };
 
 /** total_count and the token_count for a given account let us calculate the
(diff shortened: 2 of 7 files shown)


src/import-export/import-match-map.c
```diff
@@ -306,6 +373,7 @@ Account* gnc_imap_find_account_bayes(GncImportMatchMap *imap, GList *tokens)
     struct account_info account_i;
     kvp_value* value;
     kvp_frame* token_frame;
+    gboolean guid_bayes = gnc_features_check_used (imap->book, GNC_FEATURE_GUID_BAYESIAN);
```
jralls: I'm a little worried about keying off of GNC_FEATURE_GUID_BAYESIAN instead of using a new feature because GUID_BAYESIAN has been in place since 2.6.12, meaning that someone could open a 2.8.x book with a GnuCash version between that and 2.6.18, in which case GnuCash wouldn't be able to read the Bayesian import map and would write data to it in the old format.

gjanssens: Is gnucash actually taking initiative to start *using* guid's for bayesian in 2.6.12 or is it only able to handle them if it gets to open a file that is already using guid's for baysian ? The latter would be safe, the former would indeed be a problem.

jralls: IIUC Bob's approach was to be able to read guid accounts in 2.6.12. Any writes would be with names and would unset the feature flag so that the next time the book was opened in 2.7/8 it would go through and fix all of the "new" entries.

The new code in PR 233 changes the format of the KVP keys in a way that I think 2.6.12-2.6.18 won't understand but since those versions know about the feature flag will open the book anyway and the import matcher will generate "non-flat" tokens from the import and they might not match the flattened tokens, so that transactions that should match, won't. It would also unset the feature flag and the next time the book is opened with GnuCash 2.7/8 the PR233 conversion code will be presented with a mixture of flat and not-flat token keys representing what should be the same key.

gjanssens: You're right. So what we should do is  indeed use another feature key instead that supercedes GNC_FEATURE_GUID_BAYESIAN and will only be known starting from whenever this gets merged.
That will mean the GNC_FEATURE_GUID_BAYESIAN key will never be used in practice (except for a few people that have been experimenting with early 2.7 releases. So in theory we can drop GNC_FEATURE_GUID_BAYESIAN as no stable release code will ever set it.

lawsa: Bob's plan was to have gnucash 2.6 be able to read both styles, but write in the "old" fashion creating a mixed KVP. This pull request (in concern with the other pull request) makes it so that only 2.7 and up can convert an import map to GUID, and 2.6 can read to and write to it in the appropriate style.

@jralls, I see what you're saying. A file created in 2.7 then read in 2.6.18 will not be interpreted properly. As @gjanssens says, a new feature flag that supersedes the old flag makes sense. I'll do some more thinking about it.

src/import-export/test/test-flat-bayes.c
```diff
@@ -0,0 +1,186 @@
...
+
+static gchar const * suitename = "/import-export/test-flat-bayes";
+
+int
+main (int argc, char *argv[])
+{
+    int result;
+    qof_init();
+    g_test_init (&argc, &argv, NULL);
+#ifdef HAVE_GLIB_2_38
```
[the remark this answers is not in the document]
lawsa: That looks like a copy-paste artifact. I'll remove it.

src/import-export/import-match-map.c
limitedAtonement: I added a short design document here: https://wiki.gnucash.org/wiki/User:Lmat It has a question in it; would you please weigh in there?

gjanssens: I have added my input to your short design document.

src/core-utils/gnc-features.c
```diff
@@ -45,6 +45,7 @@ static gncFeature known_features[] =
     { GNC_FEATURE_NUM_FIELD_SOURCE, "User specifies source of 'num' field'; either transaction number or split action (requires at least GnuCash 2.5.0)" },
     { GNC_FEATURE_KVP_EXTRA_DATA, "Extra data for addresses, jobs or invoice entries (requires at least GnuCash 2.6.4)" },
     { GNC_FEATURE_GUID_BAYESIAN, "Use account GUID as key for Bayesian data (requires at least GnuCash 2.6.12)" },
+    { GNC_FEATURE_GUID_FLAT_BAYESIAN, "Use account GUID as key for bayesian data and store KVP flat" },
```
gjanssens: I think it would be useful to add the minimum required version that supports the feature in the help string. This string is displayed to the user when a file using the corresponding feature is opened with a gnucash version that doesn't support it. Setting the version string gives the user a hint as to which gnucash version to install.

src/import-export/test/test-flat-bayes.c
```diff
@@ -0,0 +1,289 @@
...
+main (int argc, char *argv[])
+{
+    int result;
+    qof_init();
+    g_test_init (&argc, &argv, NULL);
+    GNC_TEST_ADD (suitename, "normal_bayes_entry", Fixture, NULL, setup, &write_normal_bayes_entry, teardown);
+    GNC_TEST_ADD (suitename, "guid_bayes_entry", Fixture, NULL, setup, &write_guid_bayes_entry, teardown);
+    GNC_TEST_ADD (suitename, "normal_bayes_find", Fixture, NULL, setup, &find_normal_bayes_account, teardown);
+    GNC_TEST_ADD (suitename, "guid_bayes_find", Fixture, NULL, setup, &find_guid_bayes_account, teardown);
+    GNC_TEST_ADD (suitename, "guid_bayes_find", Fixture, NULL, setup, &find_normal_bayes_when_non_exists, teardown);
```
gjanssens: This results in the following test failure:
```
GLib-ERROR **: duplicate test case path: /import-export/test-flat-bayes/guid_bayes_find
```

Can simply be solved by renaming one of these.

gjanssens:
Except for these last small issues, I have no further comments. Unfortunately I don't have time to merge this PR either. @jralls unless you have additional remarks, feel free to fix the two minor issues and merge for the upcoming 2.6.19 release.

jralls:
Done. I also added the new test to automake.
Merged, thanks. On to the release!

PR #235, opened 2017-11-23
Item edit cell padding and code reduction of cell colours

Bob-IT:
OK I have squashed and dropped and made the changes recommended in PR 212. I have not changed the item edit entry alignment but will have another look.

Also included are a couple of changes to remove deprecated widgets and fixes to glade files that were corrupted when merged from maint.

Bob

### DIFF
--- a/gnucash/gnome-utils/gtkbuilder/dialog-preferences.glade
+++ b/gnucash/gnome-utils/gtkbuilder/dialog-preferences.glade
@@ -1,5 +1,3 @@
-<?xml version="1.0" encoding="utf-8"?>
-<?xml version="1.0" encoding="utf-8"?>
 <?xml version="1.0" encoding="UTF-8"?>
 <!-- Generated with glade 3.20.0 -->
 <interface>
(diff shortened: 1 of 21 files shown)


gjanssens:
Thanks!

Bob-IT:
Looks like I was 15mins too late for another change but it can wait for another PR. I found that on some small fonts, the pixel height is 1 pixel less than required so I was going to add 1px to all returned values. This seems to fix the cut off descenders. With this change every thing looked good from 10px to 72px but the exact vertical position is font dependant so can you advise what font you are using so I can try.

gjanssens:
My current font is Abyssinica SIL 10.

I had some time available this evening so I spent it reviewing PR's, considering this weekend we'll release 2.7.2. If you want you can put up your change, I can try and get it merged in tomorrow still.

Bob-IT:
Geert,

Interesting font, I managed to install it and set it system wide at size 10
with my original changes and yes the bottom of the descenders were cut off
also not sure if you noticed but on the splash screen all the g's were cut
off.
So I thought my fix would work, it did sort of work but there is some thing
strange with this font, for starters the underline character is below the
descenders and is not factored into the font height and so is not
highlighted in any entry. You can see this if you go to tools/import map
editor, enter some text with an underscore and descenders and highlight it.

Next, I restored my application font and added your font to my CSS file, at
10px it seemed OK with the extra 1px, 18px the bottom of the descenders
started to be missing, 32px you loose the underscore and the bottom of the
descenders.

Long story short this is going to require further thought...

Bob

PR #236, opened 2017-11-23
Bug 790620 - Failed to create file “/usr/share/glib-2.0/schemas/gsche…

codesmythe:
…mas.compiled.XY789Y”

Changed install rule for gschema.compiled to use a target.


jralls:
Thanks!

PR #237, opened 2017-11-26
Upgrade lookup-value to learn section changes

christopherlam:
I think this will take care of backward-compatibility section changes.
This is conceptually similar to previous but adds another #f state where there is a name match but the section and new-section are the same.

### DIFF
--- a/libgnucash/app-utils/options.scm
+++ b/libgnucash/app-utils/options.scm
@@ -1681,32 +1681,49 @@
           (let ((option-hash (hash-ref section-hash name)))
             (if option-hash
                 option-hash
-                ; Option name was not found. Perhaps it was renamed ?
-                ; Let's try to map it to a known new name
+                ;; Option name was not found. Perhaps it was renamed ?
+                ;; Let's try to map it to a known new name.
+                ;; This list will try match names - if one is found
+                ;; the next item will describe a pair.
+                ;; (cons newsection newname)
+                ;; If newsection is #f then reuse previous section name.
+                ;;
+                ;; Please note the rename list currently supports renaming
+                ;; individual option names, or individual option names moved
+                ;; to another section. It does not currently support renaming
+                ;; whole sections.
                 (let* ((new-names-list (list
-                        "Accounts to include" "Accounts"
-                        "Exclude transactions between selected accounts?" "Exclude transactions between selected accounts"
-                        "Filter Accounts" "Filter By..."
-                        "Flatten list to depth limit?" "Flatten list to depth limit"
-                        "From" "Start Date"
-                        "Report Accounts" "Accounts"
-                        "Report Currency" "Report's currency"
-                        "Show Account Code?" "Show Account Code"
-                        "Show Full Account Name?" "Show Full Account Name"
-                        "Show Multi-currency Totals?" "Show Multi-currency Totals"
-                        "Show zero balance items?" "Show zero balance items"
-                        "Sign Reverses?" "Sign Reverses"
-                        "To" "End Date"
-                        "Use Full Account Name?" "Use Full Account Name"
-                        "Use Full Other Account Name?" "Use Full Other Account Name"
-                        "Void Transactions?" "Void Transactions"
-                        ))
+                                        "Accounts to include" (cons #f "Accounts")
+                                        "Exclude transactions between selected accounts?" (cons #f "Exclude transactions between selected accounts")
+                                        "Filter Accounts" (cons #f "Filter By...")
+                                        "Flatten list to depth limit?" (cons #f "Flatten list to depth limit")
+                                        "From" (cons #f "Start Date")
+                                        "Report Accounts" (cons #f "Accounts")
+                                        "Report Currency" (cons #f "Report's currency")
+                                        "Show Account Code?" (cons #f "Show Account Code")
+                                        "Show Full Account Name?" (cons #f "Show Full Account Name")
+                                        "Show Multi-currency Totals?" (cons #f "Show Multi-currency Totals")
+                                        "Show zero balance items?" (cons #f "Show zero balance items")
+                                        "Sign Reverses?" (cons #f "Sign Reverses")
+                                        "To" (cons #f "End Date")
+                                        "Use Full Account Name?" (cons #f "Use Full Account Name")
+                                        "Use Full Other Account Name?" (cons #f "Use Full Other Account Name")
+                                        "Void Transactions?" (cons "Filter" "Void Transactions")
+                                        "Void Transactions" (cons "Filter" "Void Transactions")
+                                        ))
                        (name-match (member name new-names-list)))
 
-                      (if name-match 
-                        (let ((new-name (cadr name-match)))
-                             (lookup-option section new-name))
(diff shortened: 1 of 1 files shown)


libgnucash/app-utils/options.scm
```diff
@@ -1688,32 +1688,46 @@
...
             (if option-hash
                 option-hash
-                ; Option name was not found. Perhaps it was renamed ?
-                ; Let's try to map it to a known new name
+                                        ; Option name was not found. Perhaps it was renamed ?
+                                        ; Let's try to map it to a known new name.
+                                        ; This list will try match names - if one is found
+                                        ; the next item will describe a pair.
+                                        ; (cons newsection newname)
+                                        ; If newsection is #f then reuse previous section name.
```
jralls: Is there a good reason for the huge indent here?

libgnucash/app-utils/options.scm
```diff
@@ -1688,32 +1688,46 @@
...
+                                        "Show Full Account Name?" (cons #f "Show Full Account Name")
+                                        "Show Multi-currency Totals?" (cons #f "Show Multi-currency Totals")
+                                        "Show zero balance items?" (cons #f "Show zero balance items")
+                                        "Sign Reverses?" (cons #f "Sign Reverses")
+                                        "To" (cons #f "End Date")
+                                        "Use Full Account Name?" (cons #f "Use Full Account Name")
+                                        "Use Full Other Account Name?" (cons #f "Use Full Other Account Name")
+                                        "Void Transactions?" (cons #f "Void Transactions")
+                                        "Void Transactions" (cons "Filter" "Void Transactions")
+                                        "Transaction Matcher" (cons "Filter" "Transaction Matcher")
```
gjanssens: As far as I know this option is new in 2.7 and will be moved by your transaction report PR. There is no need to add compatibility code for option changes within a development cycle. Had the option existed in 2.6, the compatibility should have been added.

christopherlam: Correct; it's "Account Substring" that should be added here.

libgnucash/app-utils/options.scm
christopherlam: emacs made me do it...

libgnucash/app-utils/options.scm
```diff
@@ -1688,33 +1688,46 @@
...
+                                        "Report Currency" (cons #f "Report's currency")
+                                        "Show Account Code?" (cons #f "Show Account Code")
+                                        "Show Full Account Name?" (cons #f "Show Full Account Name")
+                                        "Show Multi-currency Totals?" (cons #f "Show Multi-currency Totals")
+                                        "Show zero balance items?" (cons #f "Show zero balance items")
+                                        "Sign Reverses?" (cons #f "Sign Reverses")
+                                        "To" (cons #f "End Date")
+                                        "Use Full Account Name?" (cons #f "Use Full Account Name")
+                                        "Use Full Other Account Name?" (cons #f "Use Full Other Account Name")
+                                        "Void Transactions?" (cons #f "Void Transactions")
```
gjanssens: You could update this one as well to use the new section name already. That saves one recursion when looking up the ancient "Void Transactions?" parameter (if a user opens an ancient report...)

It's just a minor detail though.

libgnucash/app-utils/options.scm
```diff
@@ -1688,33 +1688,46 @@
...
+                                        "Void Transactions" (cons "Filter" "Void Transactions")
+                                        "Account Substring" (cons "Filter" "Account Matcher")
```
gjanssens: Can you split out the "Account Substring" entry into a separate commit please ? That change can only be applied after your new transaction report gets merged. The other changes however can go in now already.

libgnucash/app-utils/options.scm
```diff
@@ -1688,33 +1688,46 @@
...
+                             (new-name (cdr (cadr name-match))))
+                         ;; compare if new-section name exists.
+                         (if new-section
+                             ;; if so, if it's different to current section name
+                             ;; then try new section name
+                             (and (not (string=? new-section section))
+                                  (lookup-option new-section new-name))
+                             ;; else reuse section-name with new-name
+                             (lookup-option section new-name))))))))
+      #f)))
```
gjanssens: There is a closing parenthesis mismatch in here somewhere. When I pull this PR I get a build error (using guile 2.2):
```
options.scm:1669:0: no expressions in body in form (lambda () (define option-hash (make-hash-table 23)) (define options-changed #f) (define changed-hash (make-hash-table 23)) (define callback-hash (make-hash-table 23)) (define last-callback-id 0) (define (lookup-option section name) (let ((section-hash (hash-ref option-hash section))) (if section-hash (let ((option-hash (hash-ref section-hash name))) (if option-hash option-hash (let* ((new-names-list (list "Accounts to include" (cons #f "Accounts") "Exclude transactions between selected accounts?" (cons #f "Exclude transactions between selected accounts") "Filter Accounts" (cons #f "Filter By...") "Flatten list to depth limit?" (cons #f "Flatten list to depth limit") "From" (cons #f "Start Date") "Report Accounts" (cons #f "Accounts") "Report Currency" (cons #f "Report's currency") "Show Account Code?" (cons #f "Show Account Code") "Show Full Account Name?" (cons #f "Show Full Account Name") "Show Multi-currency Totals?" (cons #f "Show Multi-currency Totals") "Show zero balance items?" (cons #f "Show zero balance items") "Sign Reverses?" (cons #f "Sign Reverses") "To" (cons #f "End Date") "Use Full Account Name?" (cons #f "Use Full Account Name") "Use Full Other Account Name?" (cons #f "Use Full Other Account Name") "Void Transactions?" (cons "Filter" "Void Transactions") "Void Transactions" (cons "Filter" "Void Transactions"))) (name-match (member name new-names-list))) (and name-match (let ((new-section (car (cadr name-match))) (new-name (cdr (cadr name-match)))) (if new-section (and (not (string=? new-section section)) (lookup-option new-section new-name)) (lookup-option section new-name)))))))) #f)))
```

Looking at options.scm, there's an unmatched closing parenthesis 1899, but the issue starts somewere in the code above.

Also while studying this, it occurred to me this code will not handle renaming a complete section. It will only handle moving an option from one (still existing) section to another and renaming the option.

We haven't ever renamed a complete section yet, but it may be worth adding a comment to show at least we considered it and are aware this is currently not supported.

libgnucash/app-utils/options.scm
gjanssens: Ok on splitting this option. Next step is to remove it from this PR and add it to your transaction.scm rework PR. I think you should squash it together with the exact commit in which the option gets renamed.
I realize this is a bit more complicated  because that makes your TR rework branch dependent on this branch (and hence should be branched off of it in your local repo as well...). The benefit of doing so is your commit history results in more bisect-friendly commits.

libgnucash/app-utils/options.scm
christopherlam: Pushed 1 change, will do further TR work tonight.

christopherlam:
This PR has been embedded onto #244

PR #239, opened 2017-11-30
Reconciled split warning Bug 771667

Bob-IT:
Added the patch I made for bug 771667 with changes so it would apply to unstable. I believe it achieves all that was requested on that bug.
Bob

### DIFF
--- a/gnucash/register/ledger-core/split-register-model.c
+++ b/gnucash/register/ledger-core/split-register-model.c
@@ -1993,7 +1993,9 @@ gnc_split_register_confirm (VirtualLocation virt_loc, gpointer user_data)
     Split *split;
     char recn;
     const char *cell_name;
-    gboolean change_ok;
+    gboolean protected_split_cell, protected_trans_cell;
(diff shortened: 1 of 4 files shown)


gnucash/register/ledger-core/split-register-model.c
```diff
@@ -2007,6 +2009,10 @@ gnc_split_register_confirm (VirtualLocation virt_loc, gpointer user_data)
     trans = xaccSplitGetParent (split);
     if (xaccTransWarnReadOnly(trans))
         return FALSE;
+
+    if (xaccTransHasReconciledSplits (trans) == FALSE)
```
gjanssens: Nitpick: while correct, the preferred style is `if (!xaccTransHasReconciledSplits (trans))`

Bob-IT:
I have made the changes requested, still forget the obvious, shame there isn't a coding style checker!!!
Bob

gjanssens:
Thanks!

PR #241, opened 2017-12-01
Handle cases where LIBDIR is not "lib"

codesmythe:
For example, Fedora installs to lib64, Debian to lib/<arch-tuple>


jralls:
Thanks!

PR #242, opened 2017-12-03
Partial fix for Bug 787497 - Disabling options cripples dist package

codesmythe:
WITH_OFX, WITH_SQL and WITH_AQBANKING handling is fixed. The
dist files for these features are always included in the dist
tarball.

I added code for WITH_GNUCASH=OFF to error out when trying to
create a dist tarball. Not fully tested because WITH_GNUCASH=OFF
itself does not build.

Still need to backport fixes from unstable for WITH_PYTHON=OFF.


codesmythe:
Ooops, hit the wrong button. Closing.

PR #244, opened 2017-12-12
transaction.scm refactoring and upgrade

christopherlam:
Offshoot from #229 refactoring and major upgrade of transaction.scm with clean history, removed all trailing whitespace.


christopherlam:
Very annoying. Not a formal bug report, just tearing hair out figuring out why it doesn't work between Windows installation types.

I copy this final .scm into Ubuntu /usr/share etc and it works well on v2.6.17
I copy into v2.6.18 on Windows 10 and it works well.
I copy into GNC Portable 2.6.18 at work and it does not work. 

Tracefile as follows. [gnucash.trace.02IBBZ.log](https://github.com/Gnucash/gnucash/files/1557986/gnucash.trace.02IBBZ.log)

gnucash/report/standard-reports/transaction.scm
```diff
@@ -1018,19 +1018,7 @@ Credit Card, and Income accounts."))))))
 
 
     (define (do-rows-with-subtotals splits
-                                    table
```
gjanssens: Can this commit work without the next one ? If not (and it doesn't look like that to me), it shouldn't be two commits. Perhaps you forgot to squash them together ?

gnucash/report/standard-reports/transaction.scm
```diff
@@ -214,14 +214,14 @@ options specified in the Options panels."))
 (define (sortkey-get-info sortkey info)
   (cdr (assq info (cdr (assq sortkey sortkey-list)))))
 
-(define (timepair-year tp)    (gnc:timepair-get-year tp))
```
gjanssens: Will this commit work without the changes you have been discussing with @limitedAtonement on IRC recently ? If not I'd prefer to keep the timepair work on a separate branch in order to get your transaction report work merged in. It's becoming quite large and hard to follow for me :(
If these changes stand on their own, they can stay of course.

christopherlam: Yes it does work without the time64 upgrade elsewhere. 

When this tr is merged in, the next work of applying time64 general upgrade will only require a couple of lines upgrading here.

gnucash/report/standard-reports/transaction.scm
christopherlam: I think you're right. I'll recheck.

gnucash/report/standard-reports/transaction.scm
```diff
@@ -1867,9 +1886,15 @@ tags within description, notes or memo. ")
 
 ;; Define the report.
 (gnc:define-report
+ 'version 2
```
gjanssens: Why did you change the version number ? I'm not sure what the side-effects are in general.

gjanssens:
As for the runtime failure you see with guile 1.8, you are probably right guile 1.8 can't handle this construct. I'm inclined to drop guile 1.8 support in gnucash 2.8 as it is no longer maintained and hasn't been for quite a while. All platforms we care about currently come with guile 2. But not all distros went through the effort of packaging gnucash for guile 2, so they may not like it if we drop guile 1.8 support.

gjanssens:
To get back on getting this branch merged, if you still have refactoring ideas queued or planned, please start a new branch for those (probably as a subbranch of this one, which makes it slightly harder for you to manage locally as you need to rebase that branch onto this one with each commit you add to this one). I'd like to ask that you only add commits to this one that directly related to review comments or bugs you find.

Otherwise we may get in an infinite loop of reviewing new enhancements/refactorings which may need fixes which get intertwined with new enhancements which need reviewing again...

I'm in general happy with the improvements you made so far and I'd like to get them into the hands of our testers as soon as possible now.

christopherlam:
Ok when you're happy with these commits let me know and I'll do a final rewriting history for amendments as above / any strings to rename etc *without* any further enhancements. I'd like to believe this is bugfree on guile 2.0 onwards ;-) any further refactoring will be put on hold and I'll go back to time64/income-GST-statement work.

I'm 90% sure on the reconciliation report defaults. The benefit here is a printable report. On the other hand, it's probably redundant to the formal reconciliation tool and may lead to further upgrade requests that I'm not particularly bothered about.

gnucash/report/standard-reports/transaction.scm
christopherlam: personal pride :) will revert.

gjanssens:
Oh, I did forget to mention, but I'm quite impressed with your history rewriting in general. It's much easier to follow what you did and why now. So thanks for that!

christopherlam:
> Oh, I did forget to mention, but I'm quite impressed with your history rewriting in general. It's much easier to follow what you did and why now. So thanks for that!

Not only you have an encyclopedic knowledge of this code, you also can coach... :-o

christopherlam:
Fixed according to comments as above. I had to do rebase from >30 commits ago because  a commit (custom sorter) was mixed up with another one (original currency & multicolumn).

gjanssens:
@jralls' recent changes to use guile's rational instead of our own homebrewn gnc:numeric class have broken your report. Can you rebase to the latest unstable and fix these issues ? I get this error when trying to run your report:
```
ERROR: In procedure module-lookup: Unbound variable: gnc:make-gnc-numeric
```

christopherlam:
Looks like gnc-numeric won't be available anymore (Woohoo) but (gnc-numeric-to-double) still available and the best one to convert numeric to number? Will retest tomorrow during the day. (Last commit is a WIP and may be amended).

gjanssens:
It looks like that. However verify you really need it. It's only in one place in your commits so far and that seems to set an amount that's not used (at least not in the diff of that commit).

jralls:
Please don't use gnc-numeric-to-double unless you actually want a double for display or a graphing function. gnc_numeric_to_scm and gnc_scm_to_numeric are exported by engine.i so gnc-numeric-to-scm and gnc-scm-to-numeric should be pretty widely available. The code is at https://github.com/Gnucash/gnucash/blob/unstable/libgnucash/engine/engine-helpers.c#L1999. Construct scheme rationals using (/ numerator denominator).

christopherlam:
Do we need:
- filter for TransactionHasAttachment status?
- additional left-column with attachment icon/text to link to above attached file/uri?
This will be a quick job

gjanssens:
Op donderdag 28 december 2017 03:12:00 CET schreef christopherlam:
> Do we need:
> - filter for TransactionHasAttachment status?
> - additional left-column with attachment icon/text to link to above attached
> file/uri? This will be a quick job

I'm not sure that would be very useful, but I'm sure someone will have a 
different opinion at some point :)

I prefer to leave it as is for now and wait until someone really asks for it.

christopherlam:
> Do we need:
>  - filter for TransactionHasAttachment status?
>  - additional left-column with attachment icon/text to link to above attached
>  file/uri? This will be a quick job
> I'm not sure that would be very useful, but I'm sure someone will have a
> different opinion at some point :)
> 
> I prefer to leave it as is for now and wait until someone really asks for it.

Agree, I tried and found it particularly useless. Tried unicode &#128206; for paperclip which works for &less;a href>, but the notebook tab does not allow target="_blank" which leaves UI undesirable. So we'll leave it as it is for now.

Please double check strings/ordering of options and let me know any last-minute changes.

gjanssens:
So a few minor last nit-picks :)

String improvement suggestions:
1. `Account Name Filter` sounds more natural and precise to me than `Account Matcher`.
1. Similarly I propose `Transaction Filter` instead of `Transaction Matcher`. Particularly given these options are on the `Filter` page after all. What do you think ?
1. In the accompanying regex options, I would use all lowercase letters except for the very first word. So there it would become `Account name filter uses regular expressions...`. Or perhaps you can come up with a more concise description ? It's pretty long for a typical check box label.

Another suggestion would be to implement unit tests for this report. It has been changed substantially and each change holds a potential bug. Unit tests will help discover those in the current code and in future changes. I would suggest to do this in another PR as I assume this would take some time...

christopherlam:
These changes sound good :)
Creating unit tests will be another learning curve, to add onto January's TODO...

gjanssens:
Op vrijdag 29 december 2017 21:49:10 CET schreef christopherlam:
> These changes sound good :)
> Creating unit tests will be another learning curve, to add onto January's
> TODO...

Thanks. That's better already although I think we shouldn't abbreviate regular 
expression into regex. (Yes, I know... I asked for a shorter description...)

I think the tooltip strings need amending as well...

"Matches only accounts whose full name is matched." sounds a bit awkward to 
me. How about "Show only accounts whose full name matches this filter". A 
native English speaker can probably formulate the more elegantly still.

Also the tooltips for the regex buttons still refer to Account/Transaction 
matcher rather than Account name/Transaction filter.

christopherlam:
Let me know any further changes!

gjanssens:
That's perfectly fine. Thanks a lot!

gjanssens:
Oh, almost right after pushing I remember another detail I saw while experimenting with your new report: one can configure the info box to only show when no transactions are matched. This is a useful option IMO. However when no account is matched (which also means no transactions are matched) the info box is not displayed. I think it should.

christopherlam:
Hmm I'll look into it, and will create small PR when ready.

PR #245, opened 2017-12-17
Change the relative path to a full one for rpath

Bob-IT:
I mentioned this on the list and the response was that it was appropriate.
Bob


jralls:
Merged, but apparently Github can't see it.

PR #246, opened 2017-12-21
Eliminate gnc configure in favor of cmake's own configure_file

gjanssens:
GNC_CONFIGURE had the disadvantage cmake was not aware at all a trackable build artifact was created. Using configure_file solves this.

I'm creating a PR because I'm actually stuck on the gschemas.compile target. I can't get it to rerun when one of the input schema files is changed. Unfortunately the reason why escapes me. The target seems to have the proper dependencies set. So I'm soliciting some input from others with hopefully more cmake experience than I have.


gjanssens:
Dammit. Wrong branch...

PR #247, opened 2017-12-21
Replace GNC_CONFIGURE(2) with configure_file

gjanssens:
GNC_CONFIGURE had the disadvantage cmake was not aware at all a trackable build artifact was created. Using configure_file solves this.

I'm creating a PR because I'm actually stuck on the gschemas.compile target. I can't get it to rerun when one of the input schema files is changed. Unfortunately the reason why escapes me. The target seems to have the proper dependencies set. So I'm soliciting some input from others with hopefully more cmake experience than I have.


gjanssens:
@codesmythe, @jralls do you see anything obvious I'm missing ?
Note what I have now is not worse than it was: any change in a gschema file used to require a re-run of cmake anyway. I just don't understand why the dependencies are not tracked properly :(

gjanssens:
Note travis build fails, suggesting there are still other issues to look into. The autotools one should not be too hard to fix. The cmake-make one suggests a target dependency resolution issue so it may be related to what I started this PR for...

jralls:
MAIN_DEPENDENCY doesn't do what @codesmythe thought it does. From the add_custom_command manual: " Note that MAIN_DEPENDENCY is completely optional and is used as a suggestion to visual studio about where to hang the custom command. In makefile terms this creates a new target in the following form:

OUTPUT: MAIN_DEPENDENCY DEPENDS
        COMMAND"

`configure_file` doesn't create a target, so there's no dependency chain from the file that's actually edited, foo.gschema.xml.in.in, to the actual dependency for the intltool custom command foo.gschema.xml.in. 

Try moving the configure_file into the first custom_command, deleting the dependency line and changing main_dependency to dependency.

gjanssens:
I don't think the configure_file use is the problem here. Whenever the <abc>.gschema.xml.in.in file is changed this will trigger a rerun of cmake, resulting in an updated <abc>.gschema.xml.in file. The <abc>.gschema.xml.in file is used as dependency to trigger a recreation of the <abc>.gschema.xml file. This works reliably.

However an update of the <abc>.gschema.xml file should trigger a rerun of the gschemas.compile target because all of these *.gschema.xml files are added as dependencies for a few top-level targets (like gnome-gschema) and the gschemas.compile target is set to depend on these. Yet that is not happening. This target is only created when it doesn't exist and never updated once it exists.

So to me the dependency chain seems to break somewhere between updating the xml file and gschemas.compile. But so far I haven't figured out why that happens.

gjanssens:
The real issue was indeed in our dependency handling and mostly my lack of understanding about it.
As @codesmythe mentioned on IRC, to catch source file changes one needs to set a file-level dependency.
So in my first attempt I replaced the target-level dependencies set on gschemas.compiled with file-level dependencies. This fixed the build with ninja, but not with make.

It took me some time to discover that "add_custom_command" rules are only defined in the directory in which it's invoked. For ninja based builds this doesn't really matter because all of ninja's build rules are gathered in one toplevel build.ninja file. However with make's recursive makefiles (as generated by cmake) the build rules for the *.gschema.xml files were not visible to the makefile for gschemas.compiled and so it couldn't find how to build its dependencies.

The solution was to set both type of dependencies: file level and target level. That's what I have pushed now.

PR #248, opened 2017-12-23
BUGFIX: Added parent window as argument

Abschiedsstein:
gnc_file_export expects the parent window as an argument. Without this
fix a compilation error is caused when --disable-dbi is used during
configuration.


gjanssens:
Although I thank you for your patch, this fix was already inserted in the unstable branch (and will be upmerged soonish to master):
https://github.com/Gnucash/gnucash/commit/76921b5e28

I saw your question on IRC regarding which branch to PR against only 5 minutes after you left.
Our branching strategy is explained [here](https://wiki.gnucash.org/wiki/Git#Branches).

While your patch came too late for this particular issue, I'll be looking forward to future fixes on your behalf :)

PR #249, opened 2017-12-24
Make the notebook pages the primary focus

Bob-IT:
The main focus of these changes make the notebook page contents the primary focus so that you can use the keyboard straight away. Also stepping through the pages with ctrl+alt+pg_up/pg_down works consistently.

All seems to work as expected.
Bob

### DIFF
--- a/gnucash/gnome-utils/gnc-main-window.c
+++ b/gnucash/gnome-utils/gnc-main-window.c
@@ -3539,6 +3539,18 @@ gnc_main_window_window_menu (GncMainWindow *window)
 #endif
 };
 
+/* This is used to prevent the tab having focus */
+static gboolean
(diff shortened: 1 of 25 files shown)


christopherlam:
This commit (notebook has default focus), and the tab doesn't get focus, is a lifesaver!

Just wondering if you know how to modify currency multichoice picker - some e.g. the price editor can accept keyboard jumps (i.e. U -> USD) (but PgUp/PgDn don't behave as expected) but most don't (e.g. within Account Editor). There's no scrollbar either, therefore currency selection usually involves mouse to up/down arrow and a loooong wait for currency to appear then click.

jralls:
Brilliant! That fixes something that's been bugging a bunch of folks for a very long time indeed!

PR #250, opened 2017-12-28
bugfix incorrect min-date handling

christopherlam:
This bugfix affects net-linechart and net-barchart - assumes the min-date for dates must be 0. Net worth bar & line charts with dates earlier than 01-01-1970 were incorrect. I'll fix in time64-ftw as well.

### DIFF
--- a/src/report/report-system/report-collectors.scm
+++ b/src/report/report-system/report-collectors.scm
@@ -155,7 +155,7 @@
     (list min-date max-date datepairs)))
 
 (define (category-report-dates-accumulate dates)
-  (let* ((min-date (gnc:secs->timepair 0))
+  (let* ((min-date (decdate (car (list-min-max dates gnc:timepair-lt)) DayDelta))
 	 (max-date (cdr (list-min-max dates gnc:timepair-lt)))
 	 (datepairs (reverse! (cdr (fold (lambda (next acc)
 					   (let ((prev (car acc))


jralls:
Unfortunately this exposes more than one bug in the tests. that need to be tracked down and fixed before it can be merged.

christopherlam:
Spent all evening to track down this bug!

1. I think the observation that category-barchart.scm was buggy because of make-date-list end-day-time end-day-time (twice) is incorrect - the report logic was sound. The report combines both inc-exp report and asset/liability report - the inc-exp report will require amounts from start-day-time of start-date to end-day-time of end-date, however the asset/liability report *typically* cares about end-of-day balances. Hence end-of-day-time twice was correct!

2. The original min-date (gnc:secs->timepair 0) was obviously wrong and produced incorrect charts for txns < 1970; the amended min-date definition of (decdate min-dates DayDelta) is now valid because it can now calculate accumulated assets/liabilities balances from any date. Well, except for edge case of a transaction dated negative MAXINT, and whose previous date is an overflow. Let's leave this one pass.

Phew!

christopherlam:
Further explanation - I gathered by manually running the test suite: say there's daily transactions from 1.1.18 to 31.1.8

The category-report-dates-accumulate function is only used in the net worth chart. Net worth are calculated from end-of-day balances in assets+income-expense. 

My original fix to define min-date of (car (list-min-max dates gnc:timepair-lt)) produced min date for 1.1.8

The net worth calculator for the category-barchart would not include transaction for 1.1.18, so I had to amend min-date to the previous day, i.e. (decdate (car (list-min-max dates gnc:timepair-lt)) DayDelta) will decrease the original min-date by DayDelta.

christopherlam:
TLDR: I haven't looked too deep into the report code, but my original fix producing min-date of 1.1.18, meant that the calculators would be taking values from end-of-day from 1.1.18 to 31.1.18 and would therefore miss the 1.1.18 daytime one.

jralls:
Thanks for the explanation and the fix.

PR #252, opened 2017-12-29
Bug 616709 - Stop delete key on Accounts page

Bob-IT:
This fix adds some callbacks which monitor the editing-started, edited and cancelled so that the delete key binding can be disabled/enabled as appropriate.

Bob


jralls:
Merged to unstable, which is what you really meant.
Thanks!

PR #253, opened 2017-12-30
ENH: display infobox when no accounts matched

christopherlam:
Bugfix infobox display when no accounts were matched. Good catch @gjanssens


gjanssens:
Thanks for the quick fix :)

PR #255, opened 2018-01-02
bugfix: fix report crash if acc-depth too low

christopherlam:
If accounts / show level < minimum-level of accounts
Previously the report would crash
This commit now displays an empty report.


jralls:
Thanks.

PR #257, opened 2018-01-05
Remove gnc-value-portfolio references from dist lists.

codesmythe:
File itself was removed in c0fd3b3.


jralls:
Thanks for cleaning up my mess!

PR #258, opened 2018-01-06
de.po: some text changes

DiMan:


po/de.po
```diff
@@ -14091,7 +14091,7 @@ msgstr "Beispiel:"
 
 #: ../src/gnome-utils/gtkbuilder/dialog-preferences.glade.h:48
 msgid "<b>Account Color</b>"
-msgstr "<b>_Kontofarbe</b>"
```
fellen: This is one of many dialog labels, where the source is missing the mnenomic ("_").
Currently I am planning to set the attribute `<property name="use_underline">True</property>`
in the glade files. Then we can use them already in the translation.

If the MsgIds should be changed now - breaking the string freeze - or later should be discussed the next days on gnucash-devel.

DiMan: Hi, I'm happy to be back :-)

It seems in this case like a mistake in german translation:
bildschirmfoto_2018-01-06_10-59-17_kontofarbe

fellen: Right, instead the next 2 entries should get some.

fellen:
commit 34200a7
Because it was on a higher branch, a merge resulted in conflicts. So I cherrypicked it.

PR #256, opened 2018-01-02
timepair->time64 mostly scheme & some API changes in C

christopherlam:
This commit incorporates some C API changes by lmat, and as much timepair eradication as I found in scheme.

PR #259, opened 2018-01-08
timepair->time64 scheme & some API changes in C

christopherlam:
Follow on from #256 time64 changes manually verified
Context: https://code.gnucash.org/logs/2018/01/08.html#T06:25:15


jralls:
This looks pretty good, except that there are a bunch of renames in scheme that should be put back, e.g. moddate64 can go back to moddate now that all instances of the old one are removed.

jralls:
This is mostly merged, except for the last commit, which I had to modify a  bit.

PR #262, opened 2018-01-13
Remove Autotools

codesmythe:


fellen:
Before autotools get removed, we need updates of several wiki pages, how to use cmake.

jralls:
Actually only one, [Building](https://wiki.gnucash.org/wiki/Building), needs instructions for cmake and it already has them. I've added notes to it saying which versions support which build system.  I've removed an old BuildSystem page that had only links to various autotools manuals and updated the pointers to Building, and updated a few other spots like [Release Process](https://wiki.gnucash.org/wiki/Release_Process) to use cmake procedures.

There are a couple of German pages that may need updating and the [Eclipse](https://wiki.gnucash.org/wiki/Eclipse) page surely does, but I don't know either German or Eclipse well enough to do it.

fellen:
Translation was written for autotools,
Customizing Reports, IIRC, ...

I hope, docs and htdocs will not change in the next time.

While I know eclipse and german, I never used CMake. I will have to set up a new machine first and later update the current before they will fulfill the changed dependencies. And I am still busy with I18N issues.

jralls:
Thanks!

PR #263, opened 2018-01-16
correct typo

Mechtilde:
correct typo in SKR 3


fellen:
Thanks, but a note for the remaining 2.6 lifecycle:

Because it has become almost impossible to merge 2.6 po's in 2.7/3.0 po's, create in the future separate patches/commits for
* 2.6 po files.
* 2.7 po files and
* other stuff like account templates, which still can easily be merged from maint in unstable.

jralls:
That won't really help, ```git merge``` will still try to merge the po files. Maybe a better idea is to not accept any translation changes for the last 2.6 release and say that all translation changes should go to unstable.

PR #264, opened 2018-01-20
ru.po: update translation on maint

DiMan:


po/ru.po
```diff
@@ -14738,7 +14737,7 @@ msgstr "дней."
 #, fuzzy
```
fellen: You should remove this line, if "Каждые" instead of "Every" should be displayed.
https://wiki.gnucash.org/wiki/Translation#Gettext_source_.28.po.29_file_format

Can you review the msgctxt "Daily"..msgctxt "Weekly" section, which fuzzy flags are obsolete?
If you like, you can also scan the whole file. I count 38 occurrences.

DiMan:
Hello Frank,

yes, you are right. The commit 2836e02 contains only the changes from "msgmerge" after make a new pot file. Thanks for the comment to the frequency words. I have just corrected it.

Dmitriy

fellen:
msgfmt: найдена 71 критическая ошибка
https://wiki.gnucash.org/wiki/Translation#Special_characters_and_other_tips
in case you want to fix that.

4677 переведённых сообщений, 3 неточных перевода, 4 непереведённых сообщения.

Thanks, commit cb2b039, you can now delete this branch.

PR #267, opened 2018-01-24
TP->T64: gnucash/report/standard-reports/transaction.scm

christopherlam:
This fix was needed due to crossed-wires whereby time64 merge didn't take a previous fix into account. This is needed for common-currency to work.


gjanssens:
Thanks!

PR #268, opened 2018-01-27
Translation27 de

Mechtilde:
some more translations for unstable.

as I demonstrate in the last commit I want to add ellipses to fit Common User Access (CUA) in the menus.

Should I also do it vor the English original? Or when is the best time to do so?

Mechtilde


fellen:
I had the same idea some time before, but after some discussion with @jralls, who prefers plain ASCII in source files i delayed it. 

Can you give some references for CUA?

Mechtilde:
here the link

https://en.wikipedia.org/wiki/IBM_Common_User_Access

It is part of the SAA standard https://de.wikipedia.org/wiki/Systems_Application_Architecture

Instead of an ellipsis we can also use three dots

fellen:
Oh, I almost forgot the "good old time" of OS/2. But I think it was also the base for most later standards.
At least of the time of GNOME 2 we tried to follow their Human Interface Guideline. I did not check their current version: https://developer.gnome.org/hig/stable/.

But yes, each menu entry without submenu or prompt execution - or the other way: which calls a dialog - should have a trailing hellip. So, can you open a bug report "missing hellip in menus" or similar against the GUI and add all occurrences you see. Then we can fix them after lifting the string freeze.

po/de.po
```diff
@@ -2118,11 +2118,10 @@ msgstr "Betrag"
 #. Translators: %d is the number of bills/credit notes due. This is a
 #. ngettext(3) message.
 #: ../gnucash/gnome/dialog-invoice.c:3435
-#, fuzzy, c-format
```
gjanssens: Shouldn't the c-format indicator stay in this case as the plural forms contain a %d ? I am not that well aquainted with the po rules, so I may be mistaken as well.

po/de.po
```diff
@@ -3127,15 +3127,15 @@ msgstr "Alte Daten abschließen und in vorigen Buchführungsperioden speichern."
 
 #: ../gnucash/gnome/gnc-plugin-basic-commands.c:191
 msgid "_Price Database"
-msgstr "Kurs-Datenbank"
+msgstr "Kurs-Datenbank..."
```
gjanssens: What should be do with these ? Should I already commit them ? If so, all the German message strings will get marked as fuzzy the day we add the ellipses in the English strings as well.

Mechtilde:
Hello Geert,

this is no problem for me, doing the German translation. I can filter
for the fuzzy translations.

But why is only the German translation affected?

Mechtilde

Am 31.01.2018 um 18:36 schrieb Geert Janssens:
> gjanssens commented on this pull request.
> 
> 
>> @@ -2118,11 +2118,10 @@ msgstr "Betrag"
>  #. Translators: %d is the number of bills/credit notes due. This is a
>  #. ngettext(3) message.
>  #: ../gnucash/gnome/dialog-invoice.c:3435
> -#, fuzzy, c-format
> 
> Shouldn't the c-format indicator stay in this case as the plural forms contain a %d ? I am not that well aquainted with the po rules, so I may be mistaken as well.
> 
>> @@ -3127,15 +3127,15 @@ msgstr "Alte Daten abschließen und in vorigen Buchführungsperioden speichern."
>  
>  #: ../gnucash/gnome/gnc-plugin-basic-commands.c:191
>  msgid "_Price Database"
> -msgstr "Kurs-Datenbank"
> +msgstr "Kurs-Datenbank..."
> 
> What should be do with these ? Should I already commit them ? If so, all the German message strings will get marked as fuzzy the day we add the ellipses in the English strings as well.
> 

-- 
Mechtilde Stehmann
## Apache OpenOffice
## Freie Office Suite für Linux, MacOSX, Windows
## Debian Developer
## Loook, calender-exchange-provider, libreoffice-canzeley-client
## PGP encryption welcome
## F0E3 7F3D C87A 4998 2899  39E7 F287 7BBA 141A AD7F

po/de.po
fellen: Good catch!

Yes, if we have at least one parameter '#, c-format' has to stay.

po/de.po
fellen: It depends on how we decide about the pending release and string freeze.

fellen:
@Mechtilde In all other translations, the respective strings wil be marked, too.

Another note:
msgfmt -c --check-accelerators="_" --statistics de.po returns a bunch of missing accelerators. Note that currently 2 are false alarms. (Place_holder and Author_credit)

fellen:
OK, it seems you rebased your branch on unstable. But only github was confused. In my local repo it is usable.

If you remove the `#, fuzzy` flag, remove the whole line if no other flags are there. The empty line is the record separator and might have unexpected results.

When you removed underlines, did you test, they were not working for mnemonics? I fear, we worked parallel in different directions: while I enabled them in the code, you removed them from de.po.

Merged in commit 5cf5875.
Oops, I forgot the "#, c-format", found a few others missing, too:
commit  2598dbe.

You should now  rebase your repo, delete this branch and start a new one.
Thank you!

PR #271, opened 2018-01-30
BUGFIX: invoice posted-date is no longer 0 for unposted invoices

christopherlam:
This commit fixes business reports caused by 91f4b190394df6
whereby the posted-date for unposted invoice was changed
from 0 to MAXINT. Now we need to use (gncInvoiceIsPosted)
or (gncInvoiceDateExists) instead.


gjanssens:
Thanks !

PR #272, opened 2018-01-30
Bug792809 Add a default register foreground css class

Bob-IT:
When using a dark theme with the existing register colours it can be hard to read the text so add a default foreground colour of black. Also mentioned in the bug was moving the user CSS settings file to $HOME/.config/gtk-3.0-gnucash.css so added in separate commit.


gnucash/gnome-utils/gnc-gnome-utils.c
```diff
@@ -223,7 +223,7 @@ gnc_add_css_file (void)
     if (var)
     {
         gchar *str;
-        str = g_build_filename (var, ".gtk-3.0-gnucash.css", (char *)NULL);
+        str = g_build_filename (var, ".config/.gtk-3.0-gnucash.css", (char *)NULL);
```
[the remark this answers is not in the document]
jralls: MacOS is BSD, so g_get_user_config_dir works on MacOS as it does on Linux and vanilla BSD: If XDG_CONFIG_HOME (it seldom is on MacOS) is set it uses that otherwise it uses ~/.config.
Apple thinks the right place is actually ~/Library/Application Support/Gnucash and that we should be forcing all of the various config stuff--aqbanking, gtk, whatever--to go there.

Bob-IT:
I am not confident I have made the correct change. I am assuming that it would be the user who would create the 'GnuCash' folder and the css file or maybe the app should with just an empty file. In that case other changes would be needed possibly in gnc_filepath_init to create the directory and empty file.
Bob

gnucash/gnome-utils/gnc-gnome-utils.c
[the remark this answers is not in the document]
gjanssens: @jralls would it make sense to propose a patch to glib to have it return ~/Library/Application Support for g_get_user_config_dir and g_get_user_data_dir for OS X ? Unless overridden by XDG_abc_HOME environment variables of course. For the record I also think they made the wrong choice on Windows. glib prefers the local APPDATA_DIR csid instead of the roaming one. They were clearly not thinking their software would be used in an Active Directory managed network...

gjanssens:
@Bob-IT This is technically doing what I asked. The only issue with it is that it is giving less than optimal results on anything but linux.
1. On Windows it will also default to the local APPDATA_DIR instead of the roaming one.
1. On OS X it will search in $HOME/.config, but that's a directory that normally doesn't exist on that platform.

As John indicates the proper location on OS X is ~/Library/Application Support. This happens to be the same as what we configure as user data dir. And for Windows the roaming APPDATA_DIR is also what we configure as user data dir. So for both platforms, you can improve the user experience by calling gnc_build_userdata_path and keep the g_get_config_dir call for linux.

doc/gtk-3.0-gnucash.css
```diff
@@ -0,0 +1,86 @@
+/* Note: Widgets obtained from Glade files will not be addressable
```
gjanssens: Perhaps we should rearrange this comment a bit to make it easier to follow for the users:
1. start by explaining this is a sample css file
2. then where it should be installed and how it should be named to actually use it (you can explain the paths on the 3 platforms here as well)
3. Then follow up with the note about the limitation on Glade file widgets

Bob-IT:
OK, I think I have made the changes required also I do normally add a comment after changes when Travis finishes but I must of got side tracked.
Bob

libgnucash/core-utils/gnc-filepath-utils.cpp
```diff
@@ -543,6 +543,33 @@ get_userdata_home(void)
...
+       $HOME/.config is used */
+    if (config_dir)
+    {
+        userconfig_home = config_dir;
+        g_free(config_dir);
+    }
+    else
+        userconfig_home = g_get_user_config_dir();
+
+    userconfig_home = userconfig_home / PACKAGE;
```
gjanssens: Unfortunately this is the wrong spot to add the PACKAGE part of the path.
userconfig_home is supposed to be the equivalent of $HOME/.config
**gnc**_userconfig_home is intended to be $HOME/.config/gnucash

I am picky about this because I want userconfig and userdata to follow the same logic.

libgnucash/core-utils/gnc-filepath-utils.cpp
```diff
@@ -684,6 +711,24 @@ gnc_userdata_dir (void)
...
+ *  - Linux: $XDG_CONFIG_HOME/Gnucash (or the default $HOME/.config/Gnucash)
+ *
+ *  @return An absolute path to the configuration directory. This string is
+ *  owned by the gnc_filepath_utils code and should not be freed by the user.
+ */
+const gchar *
+gnc_userconfig_dir (void)
+{
+    auto path_string = get_userconfig_home();
+    return g_strdup(path_string.string().c_str());
```
gjanssens: This is where you add PACKAGE.

gjanssens:
Well, I've realized the last requested change is so minimal I decided to do it myself. Thanks for making this happen @Bob-IT !

We are now free desktop compliant. Yay!

PR #273, opened 2018-02-02
Scheme progress

christopherlam:
This PR includes several (I believe are well-designed) commits:
1. xaccSplitGetReconcile to identify reconciled txns instead of (reconciled-date == 0)
2. (infobox)->(gnc:render-options-changed options) is more generalized, reduces strings.
3. TR: omit display $0 - I managed to eliminate these extra monetary subtotals.
4. Merge TR-no-account and TR-no-accounts-after-filtering, reduces strings
5. income-gst-statement.scm reworked into a transaction.scm derivative.
6. Add 2 helper funcs in options.scm to support 5 above.

### DIFF
--- a/gnucash/report/standard-reports/income-gst-statement.scm
+++ b/gnucash/report/standard-reports/income-gst-statement.scm
@@ -45,7 +45,17 @@ accounts must be of type ASSET for taxes paid on expenses, and type LIABILITY fo
 (define (income-gst-statement-renderer rpt)
   (trep-renderer rpt
                  #:custom-calculated-cells gst-calculated-cells
-                 #:empty-report-message TAX-SETUP-DESC))
+                 #:empty-report-message TAX-SETUP-DESC
+                 #:custom-split-filter gst-custom-split-filter))
+
+(define (gst-custom-split-filter split)
+  ;; split -> bool
+  ;;
+  ;; additional split filter - returns #t if split must be included
+  ;; we need to exclude Closing, Link and Payment transactions
+  (let ((trans (xaccSplitGetParent split)))
+    (and (member (xaccTransGetTxnType trans) (list TXN-TYPE-NONE TXN-TYPE-INVOICE))
+         (not (xaccTransGetIsClosingTxn trans)))))
 
 (define (gst-statement-options-generator)
   ;; Retrieve the list of options specified within the transaction report
--- a/gnucash/report/standard-reports/income-gst-statement.scm
+++ b/gnucash/report/standard-reports/income-gst-statement.scm
@@ -2,16 +2,9 @@
 ;; Income-GST-Statement.scm : Produce report suitable for
 ;; annual income tax returns and periodic VAT/GST reporting.
 ;;
-;; Original transaction.scm report by Robert Merkel <rgmerk@mira.net>
-;; Contributions by Bryan Larsen <blarsen@ada-works.com>
-;; More contributions for new report generation code by Robert Merkel
-;; More contributions by Christian Stimming <stimming@tuhh.de>
-;; Modified to support the intersection of two account lists by
-;; Michael T. Garrison Stuber
-;; Modified account names display by Tomas Pospisek
-;; <tpo_deb@sourcepole.ch> with a lot of help from "warlord"
-;; Heavily amended by Christopher Lam to add calculations
-;; appropriate for GST/VAT, building on efforts by Doug Doughty.
+;; Original Income and GST Report designed by Christopher Lam
+;; Will reuse the Transaction Report with customised options
+;; and calculated cells.
 ;;
 ;; This program is free software; you can redistribute it and/or
 ;; modify it under the terms of the GNU General Public License as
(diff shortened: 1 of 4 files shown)


christopherlam:
It includes several bug fixes for income-gst-statement, so, after review, should be merged asap.

gnucash/report/standard-reports/transaction.scm
```diff
@@ -870,17 +870,24 @@ tags within description, notes or memo. ")
...
 
-(define (make-split-table splits options)
+(define (make-split-table splits options custom-calculated-cells)
 
-  (define (opt-val section name) (gnc:option-value (gnc:lookup-option options section name)))
+  (define (opt-val section name)
+    (let ((option (gnc:lookup-option options section name)))
+      (if option
+          (gnc:option-value option)
+          (error (format #f "cannot find ~a ~a" section name)))))
```
gjanssens: I'll note the use of error again so I've looked it up in the guile manual to be sure.

First off, it looks like the format specifier is redundant ? The error function takes a message and arguments as parameters.
Secondly, it's not entirely clear to me what this would do and where the error message would get written. Do you want this message to be captured in the gnucash trace file ? Do you want it displayed in the html to the user ?

(I can't easily pull your PR from here, my laptop is not properly configured for that, so I can only ask...).

libgnucash/app-utils/options.scm
```diff
@@ -1721,7 +1721,13 @@
...
                                         "Void Transactions?" (cons "Filter" "Void Transactions")
                                         "Void Transactions" (cons "Filter" "Void Transactions")
-                                        "Account Substring" (cons "Filter" "Account Matcher")
+                                        "Account Substring" (cons "Filter" "Account Name Filter")
+                                        "Account Matcher" (cons "Filter" "Account Name Filter")
+                                        "Transaction Matcher" (cons "Filter" "Transaction Filter")
+                                        "Account Matcher uses regular expressions for extended matching"
+                                        (cons "Filter" "Use regular expressions for account name filter")
+                                        "Transaction Matcher uses regular expressions for extended matching"
+                                        (cons "Filter" "Use regular expressions for transaction filter")
```
gjanssens: Why have you added all these in the option migrations ? Aren't these newly introduced options in master/unstable only ? Or did they exist in 2.6 already ?

christopherlam: I'd thought they needed to be included because GST options using old section/name were in a previous 2.7.x release and should be migrated. However agree they're not in maint.

gnucash/report/standard-reports/transaction.scm
christopherlam: Hmm from memory they were displayed with --debug. I'll modify to gnc:error.

christopherlam:
Hi, @gjanssens, thank you, good points

The report asymmetry was borne out of necessity, wishing to keep the TR renderer intact as far as possible. I can certainly break down the renderer further to accept either default-calculated-cells or income-gst-statement's calculated cells in proper OO fashion.

As far as GST's options, my view was that the TR could be a moving target (with new features if required) and the GST's options would do a few 'adjustments' to TR's options e.g. to retrieve the appropriate tax accounts.

But if there must be separation between the TR options + base renderer, and GST options + reuse of TR's renderer, then I will trim down the GST's options to the minimum required to achieve a good report? e.g. numerous amount/num/etc sortkeys can be removed from GST's options.

In conclusion, if it's preferable for GST's options to be independent, then I'll rewrite them all over next couple days, and pass them all to TR's base renderer with numerous invisible ('internal') options e.g. friendly debit/credit names. I don't anticipate any new strings therefore can be merged afterwards. Let me know if anything else is desirable!

gjanssens:
I have been giving this some more thought. I think we can leave things as you have written them. Perhaps a few stategically-placed comments can clarify your intent. If it turns out later the TR and GSTR will start to live more separate lives, we can still do the refactoring then.

In general what I'm more cautious about is that we are now in feature freeze (in addition to string freeze). That means we should now be really focussed on **stabilizing**  the features that made the feature freeze. That means cleaning up bugs, regressions and some minor polishing of the code.

Although your work on bringing GST and TR closer to getter is very welcome, it is by definition also much more intrusive that stabilizing what's there. It's still a fairly big refactoring. And my experience over the years tells me that always comes with an increased risk on new bugs and with it decreasing the stability. So it makes me a bit uneasy to include this change still for 3.0, even though I agree the new code is better in principle. I'm particularly uneasy in this case since we don't have comprehensive unit tests yet that could at least help us guard against regressions.

christopherlam:
Hi Geert

Thank you, my initial feeling during development of TR was that it was borrowing some ideas from the GSTR hence they should eventually merge.

After all, the TR is now essentially a SQL-like query language, and is so versatile that it should be the one receiving more thorough attention. 

I've been going through the 52 options, and their complicated master/slave dynamics, and feel if there must be separation of TR and GSTR options generator, it will definitely be much more messy with much more duplicating code. Hence the strict asymmetry in this commit to my eyes is easier to work with an understand. I can also confirm that unstable's GSTR is currently buggy (with foreign currency conversions) and this commit fixes it.

w.r.t. tests- I haven't done TDD and I understand unit tests are essentially an insurance against future breakage? If so, I think TR can receive more thorough unit tests, and GSTR will receive a few strategic tests involving its personalized options and custom-calculated-cells, and all will be fine.

gnucash/report/standard-reports/income-gst-statement.scm
```diff
@@ -33,1792 +26,222 @@
...
-                 (car splits) table width def:primary-subtotal-style used-columns))
-            (if secondary-subheading-renderer
-                (secondary-subheading-renderer
-                 (car splits) table width def:secondary-subtotal-style used-columns))
-
-            (do-rows-with-subtotals splits table used-columns width
-                                    multi-rows? #t
-                                    export?
-                                    account-types-to-reverse
-
```
christopherlam: Note to self: must change to (gnc:warn) also next line

christopherlam: Note to self: need to annotate always #f for GST report

christopherlam: Am I right that:
- (_ xx) strings are translated in UI and code
- (N_ xx) strings are translated in UI but are unchanged in code, i.e. saved-settings are not translated?
If I'm right then I guess should these option names also be changed to: (N_ "xx")?

gjanssens:
(_ xx) will lookup a translation for whatever is returned by xx or the 
litteral xx string depending on whether xx is a function or not.
So

  (_ "A value")

will be replaced with "Une valeur" on a computer with French locale.

  (set! some_value "A value")
  (_ some_value)

would do the same provided "A value" has been marked as translatable 
somewhere. In this limited example it wasn't, so (_ ) will probably not find a 
translation and will return the untranslated English string.

This is where (N_ ) comes in. It is used to mark strings for translation 
without translating them on the spot. This means they will get added to the 
catalog of translatable strings for later translation use in a call to (_ )

So to fix the previous issue you can change it to

  (set! some_value (N_ "A value"))
  (_ some_value)

In this case some_value is still set to "A value" - it's untranslated. But the 
translation system get's a queue that at some point we would like to use a 
translated variant, which we do in the next line.

This could raise the question: why not immediately use

  (set! some_value (N_ "A value"))

Because in this case some_value is always a translated string (ignoring for a 
moment our translations may not be complete). And that's not always what we 
want.

In particular, we don't want the option names to be translated when stored in 
for example a "saved report setting". That would break all saved settings if 
the user switched locale. But we do want to translate the option name when 
displaying it to the user for his/her convenience
So, in code:

  (set! some-value (N_ "A value"))
  (set! option-save-name some-value)
  (set! option-display-name (_ some_value))

christopherlam:
> (_ xx) will lookup a translation for whatever is returned by xx or the litteral xx string depending on whether xx is a function or not. So

... this N_ vs _ finally makes sense!

christopherlam:
For a couple hours I added the breaking down of renderer into multistage - however it would confuse things. I've just restored the closing/link/payment filter that was forgotten in the income-gst-statement rewrite above.

fellen:
@gjanssens 

> In particular, we don't want the option names to be translated when stored in for example a "saved report setting". That would break all saved settings if the user switched locale. But we do want to translate the option name when displaying it to the user for his/her convenience So, in code: `(set! some-value (N_ "A value")) (set! option-save-name some-value) (set! option-display-name (_ some_value))`

This is not the intention of noop_gettext "N()". It should be used, where calling a function like gettext "_()" would result in a compiler error like assigning default values to an array in the declaration part.

Our SCM code should be cleaned up and distinuish between option_X_key, option_X_label and option_X_default. 
```
 ;;; key for value in gschema:
 (set! some-value_key  "A value")
 ;;; optional if used in composed strings, window titles, ...:
 (set! some-value_name (_ some-value_key)) 
 ;;; Label with mnemonic for the gui:
 (set! some-value_label (_ "A _value"))
 ;;; gschema description and dialog tooltip:
 (set! some-value_hint (_ "The purpose of a value is ..."))
 ;;; a default for value:
 (set! some-value_default  (_ "void"))
```
See also https://wiki.gnucash.org/wiki/GUI_Guidelines

christopherlam:
If anyone reviewing I'd think this PR should be ripe

gjanssens:
Op zaterdag 10 februari 2018 04:54:46 CET schreef Frank H. Ellenberger:
> @gjanssens
> 
> > In particular, we don't want the option names to be translated when stored
> > in for example a "saved report setting". That would break all saved
> > settings if the user switched locale. But we do want to translate the
> > option name when displaying it to the user for his/her convenience So, in
> > code: `(set! some-value (N_ "A value")) (set! option-save-name
> > some-value) (set! option-display-name (_ some_value))`
> This is not the intention of noop_gettext "N()". It should be used, where
> calling a function like gettext "_()" would result in a compiler error like
> assigning default values to an array in the declaration part.
> 
> Our SCM code should be cleaned up and distinuish between option_X_key,
> option_X_label and option_X_default. ```
>  ;;; key for value in gschema:
>  (set! some-value_key  "A value")
>  ;;; optional if used in composed strings, window titles, ...:
>  (set! some-value_name (_ some-value_key))
This will not work as some-value_key is not a translatable string, but a 
variable. So xgettext will not have a string to extract and to translate.
Nowhere in your example you give xgettext a litteral "A value" string to add 
to the potfile. To fix this, you should
(set! some-value_key  (N_ "A value"))

However, if you are thinking of cleaning up our scm option code, I'd suggest 
to do it properly and cut the link between key and name completely. The above 
example would then become:

;;; key for value in gschema, note no more spaces:
(set! some-value-key  "a-key")
;;; optional if used in composed strings, window titles, ...:
(set! some-value_name (_ "A name"))

PR #274, opened 2018-02-04
fix syntax of example qif file

jbclements:
Attempting to import the given example .qif file results in an error about an unrecognized credit limit format. It appears that this is caused by a lack of transaction breaks between the !Account header and the individual transactions. Adding the breaks as in this pull request appears to fix the example; I can successfully import this .qif file into GnuCash, and the transaction appears correct.

### DIFF
--- a/gnucash/import-export/qif-imp/file-format.txt
+++ b/gnucash/import-export/qif-imp/file-format.txt
@@ -392,6 +392,8 @@ Investment Transactions
 !Account
 NAssets:Investments:Mutual Fund
 TInvst
+^
+!Type:Invst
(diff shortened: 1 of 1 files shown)


gjanssens:
Thanks for your PR. I don't know much about qif, but I verified your change gets the otherwise broken sample file successfully imported.

I have cherry-picked the fix to our maint branch rather than merging into our master branch. The former is for 2.6, while the latter is for post 3.0 and the unstable branch is for preparing the 3.0 release. As your fix is equally valid for 2.6 I applied it there first.

jbclements:
Thanks!

PR #276, opened 2018-02-08
bug 793278 fix

christopherlam:
This is caused by commit 766e74096 - min-date was
erroneously thought to mean 'min date of date-list'
but actually meant 'negative infinity date'. This
commit changes date comparison logic to always
return #t when comparing (<= min-date date) for
the first date interval.

Test case also created.

### DIFF
--- a/gnucash/report/report-system/report-collectors.scm
+++ b/gnucash/report/report-system/report-collectors.scm
@@ -93,7 +93,8 @@
          (slotset (make-slotset (lambda (split)
                                   (let* ((date (split->date split))
                                          (interval-index (binary-search-lt (lambda (pair date)
-                                                                             (<= (car pair) date))
+                                                                             (or (not (car pair))
+                                                                                 (<= (car pair) date)))
                                                                            date
                                                                            date-vector))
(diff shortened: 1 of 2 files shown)


gjanssens:
Pushed after a rebase. Thanks! I'm rebasing to unclutter git history a bit. I should remember to ask you to do it right before pulling...

DiMan:
Hello gjannsens,

Thank you for checking and accepting the bugfix.

This bug also exists in maint. Would it be possible to apply this bug fix to maint as well?
So I can use the stable maint version without bug.

Thanks, DM

christopherlam:
How embarassing. Shall I add onto #280 ?

jralls:
No, if @DiMan is right it should get its own PR... but will e3c2595 really apply to maint? I thought that it was fixing a previous commit that had gone only on unstable.

christopherlam:
> No, if @DiMan is right it should get its own PR... but will e3c2595 really apply to maint? I thought that it was fixing a previous commit that had gone only on unstable.

766e740 was my original misunderstood commit to maint. Will PR in a few hours back home.

PR #277, opened 2018-02-12
Main code

Mechtilde:
I alos correct typo and tried to make a string translatable

### DIFF
--- a/data/accounts/de_DE/acctchrt_skr03.gnucash-xea
+++ b/data/accounts/de_DE/acctchrt_skr03.gnucash-xea
@@ -1145,7 +1145,7 @@
     <act:parent type="new">d0c00095e4b00f9f25253307be05e885</act:parent>
   </gnc:account>
   <gnc:account version="2.0.0">
-    <act:name>4957 Abschluß- u. Prüfungskosten</act:name>
+    <act:name>4957 Abschluss- u. Prüfungskosten</act:name>
     <act:id type="new">bf093094a55c036a193357ebcafcab67</act:id>
     <act:type>EXPENSE</act:type>
     <act:commodity>
--- a/libgnucash/app-utils/gnc-sx-instance-model.c
+++ b/libgnucash/app-utils/gnc-sx-instance-model.c
@@ -970,7 +970,7 @@ _get_template_split_account(const SchedXaction* sx,
         char guid_str[GUID_ENCODING_LENGTH+1];
         gchar* err;
         guid_to_string_buff((const GncGUID*)acct_guid, guid_str);
-        err = g_strdup_printf ("Unknown account for guid [%s], cancelling SX [%s] creation.",
+        err = g_strdup_printf (_("Unknown account for guid [%s], cancelling SX [%s] creation."),
                         guid_str, xaccSchedXactionGetName(sx));
         g_critical("%s", err);
         if (creation_errors != NULL)
--- a/data/accounts/de_DE/acctchrt_skr04.gnucash-xea
+++ b/data/accounts/de_DE/acctchrt_skr04.gnucash-xea
@@ -8884,7 +8884,7 @@
     <act:parent type="new">984e3a4bbeb1a32db7dd6df9437e6ee1</act:parent>
   </gnc:account>
   <gnc:account version="2.0.0">
-    <act:name>Rückstellungen für Abschluß- und Prüfungskosten</act:name>
+    <act:name>Rückstellungen für Abschluss- und Prüfungskosten</act:name>
     <act:id type="new">9dd399dbda29584d1bf898838ecf2eac</act:id>
     <act:type>LIABILITY</act:type>
     <act:commodity>
@@ -21392,7 +21392,7 @@
     <act:parent type="new">34be7b6c3246314364ef3ad2c6aead4c</act:parent>
   </gnc:account>
   <gnc:account version="2.0.0">
-    <act:name>Abschluß- und Prüfungskosten</act:name>
+    <act:name>Abschluss- und Prüfungskosten</act:name>
     <act:id type="new">0a45ddd44e05c5daa293526d57a82346</act:id>
     <act:type>EXPENSE</act:type>
     <act:commodity>


libgnucash/app-utils/gnc-sx-instance-model.c
```diff
@@ -970,7 +970,7 @@ _get_template_split_account(const SchedXaction* sx,
         char guid_str[GUID_ENCODING_LENGTH+1];
         gchar* err;
         guid_to_string_buff((const GncGUID*)acct_guid, guid_str);
-        err = g_strdup_printf ("Unknown account for guid [%s], cancelling SX [%s] creation.",
+        err = g_strdup_printf (_("Unknown account for guid [%s], cancelling SX [%s] creation."),
                         guid_str, xaccSchedXactionGetName(sx));
         g_critical("%s", err);
```
fellen: This is a log procedure [https://developer.gnome.org/glib/stable/glib-Message-Logging.html#g-critical]. It should return english text because the devs are not fluent in all languages for which we offer translations.

Did you see this string in the GUI and where?

data/accounts/de_DE/acctchrt_skr03.gnucash-xea
```diff
@@ -1145,7 +1145,7 @@
     <act:parent type="new">d0c00095e4b00f9f25253307be05e885</act:parent>
   </gnc:account>
   <gnc:account version="2.0.0">
-    <act:name>4957 Abschluß- u. Prüfungskosten</act:name>
+    <act:name>4957 Abschluss- u. Prüfungskosten</act:name>
     <act:id type="new">bf093094a55c036a193357ebcafcab67</act:id>
```
fellen: It was no typo but the only correct form in de:1906. The text was copied from DATEVs SKRxx.pdf.
I would suggest to make a complete update based on more recent editions of the SKRs by datev.de instead of some punching.

Mechtilde:
Hello Frank,
Am 14.02.2018 um 06:45 schrieb Frank H. Ellenberger:
> fellen commented on this pull request.
> 
>> @@ -970,7 +970,7 @@ _get_template_split_account(const SchedXaction* sx,
>          char guid_str[GUID_ENCODING_LENGTH+1];
>          gchar* err;
>          guid_to_string_buff((const GncGUID*)acct_guid, guid_str);
> -        err = g_strdup_printf ("Unknown account for guid [%s], cancelling SX [%s] creation.",
> +        err = g_strdup_printf (_("Unknown account for guid [%s], cancelling SX [%s] creation."),
>                          guid_str, xaccSchedXactionGetName(sx));
>          g_critical("%s", err);
> 
> This is a log procedure [https://developer.gnome.org/glib/stable/glib-Message-Logging.html#g-critical]. It should return english text because the devs are not fluent in all languages for which we offer translations.

This is ok
> 
> Did you see this string in the GUI and where?

It pops up after starting GnuCash and the schedule commes up. there I
press <ok>. Then the message box appears

The first string <Voided transaction> is translated, but not the second one.

It seems to come from </libgnucash/engine/Transaction>

I hope this helps

Kind regards

-- 
Mechtilde Stehmann
## Apache OpenOffice
## Freie Office Suite für Linux, MacOSX, Windows
## Debian Developer
## Loook, calender-exchange-provider, libreoffice-canzeley-client
## PGP encryption welcome
## F0E3 7F3D C87A 4998 2899  39E7 F287 7BBA 141A AD7F

fellen:
While investigating I found in maint and unstable in [libgnucash/]app-utils/gnc-sx-instance-model.c  several occurrences of the form:
```
    if (<condition>)
    {
        gchar *err = g_strdup_printf(<untranslatedformatString>, <param1, ...>);
/* logfile output: untranslated */
        g_critical("%s", err);
        if (creation_errors != NULL)
/* list for the user should be translated */
            *creation_errors = g_list_append(*creation_errors, err);
        else
            g_free(err);
        return FALSE;
    }

```
it should become something like:
```
  msg=N_(<untranslatedformatString>)
  g_critical(msg, <param1, ...>)
  :
  *creation_errors = g_list_append(*creation_errors, g_strdup_printf(_(msg), <param1, ...>))
  :
```
@jralls 
@gjanssens 
Because there are several occurrences should we have a separate procedure or patch all?

jralls:
You'd have to write a variadic macro or function to consolidate it. They can be kind of fiddly, so I guess it depends on how much editing you'll save compared to how much time you'll spend getting the macro or function to work the way you want.

fellen:
@Mechtilde I think I know how to fix it. But I do not know, how to test it. Every time when I try to delete or rename one of the related accounts, the missing account gets already catched by some other dialog.

> It pops up after starting GnuCash and the schedule commes up. 

That is "Action->Scheduled Transactions->Since last  run", which will autostart if set in Edit->Preferences->Scheduled Transactions.

So, how did you break the SX?

fellen:
I beleave, I solved the issue with the untranslated SX messages. Can you please test it and report back which can be seen by which action and conditions.

To avoid work on translation, I got the following idea, but did not test it:
in unstable:
- make pot
- msgmerge ...
then try to msgcat a recent copy of de.po from maint branch into that of the unstable branch.

gjanssens:
@fellen The merge procedure you suggest to get translations from maint into unstable is more or less what I have used so far when we got translations on the maint branch. It works more or less, but the git diffs are almose undecipherable. So you have to trust msgmerge/msgcat for doing a decent job.

msgcat should be used very carefully here. By default it will merge translations if there's one for the same msgid in several po files. I think this is almost never what we want. So I'm tempted to use the '--use-first' command line switch to use only the first translation found. Then one can have some control over which translation is picked by playing with the order in which the po files are passed to msgcat.
But what if there are two conflicting translations (which may or may not be fuzzy): which one would you want ? Was the specific string most recently updated in unstable or in maint ? This is very hard to track.

gjanssens:
Aside from the merge discussion, what should happen with this PR now ? The issue in commit 
b9ef8a6 has been fixed by @fellen, so this commit is obsolete (also causing a merge conflict which can be resolved fairly easily). As I understand the two other commits were more or less rejected by @fellen on the premise this should better be handled in one global new import of SKR.

So it looks like there is nothing left to do here but closing the PR.

fellen:
@gjanssens Agreed, can you improve https://wiki.gnucash.org/wiki/Translation#Merging_with_divergent_source_trees ?

PR #278, opened 2018-02-13
Updated info and migrated to the latest AppStream specs

GitHaarek:
+ Specified metadata license
+ Added required nametag
+ Added required project license
+ Added required summary
+ Changed urls to https
+ Fixed screenshot tags
+ Added category
+ Added url for bugtracker, faq, help, donation and translate

I have not removed the current underscore prefixes as you may want to migrate from the deprecated intltool to modern gettext before doing so. Relevant information:
- https://blogs.gnome.org/mclasen/2016/07/21/using-modern-gettext/
- https://wiki.gnome.org/MigratingFromIntltoolToGettext

### DIFF
--- a/gnucash/gnome/gnucash.appdata.xml.in
+++ b/gnucash/gnome/gnucash.appdata.xml.in
@@ -1,34 +1,48 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<application>
- <id type="desktop">org.gnucash.Gnucash</id>
- <licence>CC0</licence>
- <description>
-  <_p>
-   GnuCash is a program for personal and small-business financial-accounting.
-  </_p>
-  <_p>
-   Designed to be easy to use, yet powerful and flexible, GnuCash allows you
-   to track bank accounts, stocks, income and expenses. As quick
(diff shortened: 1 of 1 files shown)


gjanssens:
Thanks for your PR. We are indeed still supporting gnucash on platforms that don't have a recent enough gettext, however I'll keep your links for future reference when we decide to drop old gettext versions support.

Can you rebase you work on the unstable branch please ? That's currently the most active branch, leading up to 3.0. Master branch is for 4.0.

gjanssens:
Hmm, I'm not sure what you did exactly, but it didn't have the effect I was hoping.
The commit is still based on the master branch, but the PR now asks to pull it into unstable. What I had in mind was to update the PR that the commit itself would be based on the current unstable head. That would require a git cherry-pick or a git rebase command to achieve.

In this case as the merge is trivial, I can do this for you. The only side effect is that github will not notice I have now merged your change. So I'll close the PR manually. For any future PR's, please start from unstable until 3.0 has been formally released.

Your commit is now part of unstable. Thanks !

PR #279, opened 2018-02-14
BUGFIX: numeric->number in commodity-utilities.scm 

christopherlam:
I think this patches correctly for now.
Addendum to e0300d3 - was crashing reports.
And 45f61a3 had couple bugs.
value/share swapped
no need to define comm-list in create-commodity-list


jralls:
The bug fix needs to be done in maint and the numeric->number in unstable.

christopherlam:
Ok will do, currently at work!

gjanssens:
Thanks !

PR #280, opened 2018-02-14
BUGFIX: create-commodity-list in commodity-utilities.scm

christopherlam:
45f61a3 had couple bugs.
- value/share was inadvertently swapped
- no need to define comm-list in create-commodity-list to return it


jralls:
Thanks.

PR #281, opened 2018-02-15
bug 793278 fix for maint

christopherlam:
This is caused by commit 766e74096 - min-date was
erroneously thought to mean 'min date of date-list'
but actually meant 'negative infinity date'. This
commit changes date comparison logic to always
return #t when comparing
(gnc:timepair-le min-date date) for
the first date interval.

Test case also created. The test is derived from the
equivalent test in unstable.

I believe this fixes maint. I think merging onto unstable
will lead to conflicts, both to report-collectors.scm
and to the unit test. When resolving the conflict,
unstable must take priority.

This is the maint equivalent to #276


jralls:
Thanks.

I've also merged maint into unstable and discarded the changes from this commit.

PR #283, opened 2018-02-20
WITH_PYTHON uses backwards compatible print().

gitluwum:
# Issue
If  `python3` is aliased or installed as `python` you will receive the following error.
```
CMake Error at CMakeLists.txt:468 (MESSAGE):
  Could not determine Python site-package directory:
    File "<string>", line 1
      from distutils import sysconfig; print sysconfig.get_python_lib(prefix='/usr', plat_specific=True)
                                                     ^
  SyntaxError: invalid syntax

```
# Proposed Solution
Use the print function with parenthesis for compatibility between versions 2 and 3.


jralls:
Is that really the only blocker for Py3?

gitluwum:
It isn't yet :disappointed:
Going to be looking into it.

PR #284, opened 2018-02-21
Cache a couple of split register preferences

Bob-IT:
Two preferences were being used heavily in the drawing of the register especially when it is obscured so it makes sense for them to be saved in the split_register structure so they can be easily be referenced.

### DIFF
--- a/gnucash/register/ledger-core/split-register.c
+++ b/gnucash/register/ledger-core/split-register.c
@@ -1876,7 +1876,7 @@ gnc_split_register_get_account_by_name (SplitRegister *reg, BasicCell * bcell,
     }
 
     /* Now have the account. */
-    account_name = gnc_get_account_name_for_register (account);
+    account_name = gnc_get_account_name_for_split_register (account, reg->show_leaf_accounts);
     if (g_strcmp0(account_name, gnc_basic_cell_get_value(bcell)))
     {
         /* The name has changed. Update the cell. */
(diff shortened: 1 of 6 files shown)


jralls:
Thanks, Bob.

gnucash/register/ledger-core/split-register.c
```diff
@@ -2691,6 +2691,14 @@ split_register_pref_changed (gpointer prefs, gchar *pref, gpointer user_data)
     {
         info->separator_changed = TRUE;
     }
+    else if (g_str_has_suffix(pref, GNC_PREF_SHOW_LEAF_ACCT_NAMES))
+    {
+        reg->show_leaf_accounts = !reg->show_leaf_accounts;
```
gjanssens: This is a potentially dangerous construct. If the setting-changed signal fails to trigger at some point, or triggers twice while the option is only changed once show_leaf_accounts would go out of sync with the real option.
Considering the option change doesn't happen that often, I would prefer to just use gnc_prefs_get_bool here. That way you're sure to always get the current value of the preference.

Other than that, nice work. Thanks!

Bob-IT: I had it that way to start with but thought this way was simpler but I can see your point so expect a new PR to change it shortly.

gjanssens:
Thanks!

PR #285, opened 2018-02-22
tests: Skip tests for dates on 32-bit platforms which cannot be representated

Whissi:
Tests "test_IANA_Perth_tz" and "test_IANA_Minsk_tz" are failing on 32-bit
platforms because the earliest timestamp which can be representated on a
32-bit platform is "1901-12-13 20:45:52" but the test ranges start before
that date.

To be safe, this commit will add code to skip tests before year 1902 on
32-bit platforms.

Bug: https://bugs.gentoo.org/647596

### DIFF
--- a/libgnucash/engine/test/gtest-gnc-timezone.cpp
+++ b/libgnucash/engine/test/gtest-gnc-timezone.cpp
@@ -154,6 +154,14 @@ TEST(gnc_timezone_constructors, test_IANA_Perth_tz)
         }
         else if (year < 1916)
 #else
+        if (year < 1902)
+        {
+            // Earliest timestamp which can be represented on a 32-bit
+            // system is "1901-12-13 20:45:52" -- so skip tests until we
+            // reach a year >1901 to be safe
+            continue;
+        }
+
         if (year < 1916)
 #endif
         {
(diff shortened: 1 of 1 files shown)


jralls:
No, we're using boost::date_time and a 64-bit time representation specifically to avoid the time representation problems with 32-bit time_t (which even some 64-bit libc implementations retain). The reason for the ```#ifdef __LP64__``` is that 32-bit systems tend to provide only TZIF files and those don't include the local apparent noon timezone; the first one they have is the "standard" timezone for the locale.

Whissi:
So are you going to apply a different solution or are you going to keep a broken test suite on 32-bit?

jralls:
The test suite works on Debian-9 32-bit. If it's failing on BSD 32-bit, file a bug.

Whissi:
It is failing on Gentoo Linux x86, see our linked bug report. Here you can see the test failure: https://647596.bugs.gentoo.org/attachment.cgi?id=520072

jralls:
Now that's interesting: It's loading the 64-bit section of the zoneinfo file, so if you *remove* the #ifdef __LP64__ it should pass. I guess that means that __LP64__ is too general a test, though it probably means that something else is broken too.

Please file a GnuCash bug. You can just post the output of bin/test-gnc-timezone.

PR #286, opened 2018-02-23
Unstable prepare srfi64

christopherlam:
A few commits, mainly to prepare for srfi-64 and sxpath-based tests which will need well-formed XML renders.
1. Change html break tag from `<br>` to `<br />`
2. Opening `<html>` tag was mistakenly deleted two years ago.
3. Test for SRFI-64 and enable testing group. This group will be run on systems containing guile-2.0.10 onwards, such as travis-ci arch.

And a minor amendment to TR according to guile guide.

After this, TR-unittest #270 can progress.

### DIFF
--- a/gnucash/report/standard-reports/test/CMakeLists.txt
+++ b/gnucash/report/standard-reports/test/CMakeLists.txt
@@ -6,6 +6,9 @@ SET(scm_test_standard_reports_SOURCES
   test-standard-net-linechart.scm
 )
 
+set(scm_test_with_srfi64_SOURCES
+)
+
 SET(scm_test_report_SUPPORT
   test-generic-category-report.scm
   test-generic-net-barchart.scm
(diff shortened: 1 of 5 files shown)


gnucash/report/standard-reports/test/CMakeLists.txt
```diff
@@ -6,6 +6,9 @@ SET(scm_test_standard_reports_SOURCES
   test-standard-net-linechart.scm
 )
 
+set(scm_test_with_srfi64_SOURCES
```
gjanssens: I believe the scm_test_with_srfi64_SOURCES variable should be added to the dist list at the end of this file (in SET_DIST_LIST(test_standard_reports_DIST ...))
The set is currently empty but it won't remain empty...

gjanssens: Travis wouldn't detect this issue. It would only surface in a make/ninja-build dist-check, and even then only after you start adding unit test sources to the variable.
But since this is the commit in which you introduce the variable, I believe it would be most elegant to make it a complete introduction :)

gjanssens:
And with that it's pushed. Thanks!

PR #287, opened 2018-02-23
Use the gnc_prefs_get_bool values to update split-register cache values

Bob-IT:
This way is a safer way to update the cache values in case the settings
changed signal fails to trigger or there is a double trigger and the
cached value gets out of sync with the preference value.


gjanssens:
Thanks!

PR #288, opened 2018-02-27
bugzilla 793898 and ongoing scheme progress

christopherlam:
No user-visible changes at all, except that TR may have to be recreated/resaved.

- TR multichoice -> symbols instead of lists. This may break saved-reports/reports across sessions. Best merge this in unstable
- Fixes unintended copyright statements.
- Clean up `<br/>` tag to include `\n` in HTML.
- Homogenize TR blank-report messages

### DIFF
--- a/gnucash/report/standard-reports/daily-reports.scm
+++ b/gnucash/report/standard-reports/daily-reports.scm
@@ -2,7 +2,6 @@
 ;; daily-reports.scm: reports based on the day of the week
 ;;
 ;; Copyright (C) 2003, Andy Wingo <wingo at pobox dot com>
-;; Christopher Lam upgrade to time64 (2017)
 ;;
 ;; based on account-piecharts.scm by Robert Merkel (rgmerk@mira.net)
 ;; and Christian Stimming <stimming@tu-harburg.de> with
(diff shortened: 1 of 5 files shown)


christopherlam:
I think this part is all done now

jralls:
WRT bug 793898, it was inappropriate for Dmitry to even raise the issue. I've closed it as invalid. Whether a particular commit is worth a claim of copyright is an internal matter for the development team. If anyone thinks it needs discussion or guidance we can take it up on the dev list and work up some guidance for the wiki.

gjanssens:
I interpreted the bug claim differently. I thought the copyright statement itself was oddly formed (as if the result of a poor copy and paste). Usually a copyright statement only states "Copyright (c)" followed by a year, the copyright holder's name and e-mail address. The statement in sx-summary.scm read more like a file change history. So I understood Dmitry's bug report to point out this confusion.

@christopherlam For clarity, you can explicitly claim copyright for your work if you prefer this, but don't combine it with file change history records. However in a git environment I have always assumed that superfluous: it's very easy to track all your contributions in git history (which was not always the case in cvs or svn). IANAL though, so maybe we should indeed discuss this in more detail and verify our licence and copyright statements reflect this sufficiently clearly.

gjanssens:
Hmm, for some reason github doesn't detect this has been merged. Perhaps because the PR was still in state "Changes requested"...

Anyway: thanks!

PR #289, opened 2018-02-28
Add -Werror=implicit-fallthrough

nielsegberts:
In two instances the code seems to fall through unintentionally, though
I think it did not result in a reproducible bug.

### DIFF
--- a/libgnucash/engine/Account.cpp
+++ b/libgnucash/engine/Account.cpp
@@ -409,6 +409,7 @@ gnc_account_get_property (GObject         *object,
         break;
     case PROP_SORT_REVERSED:
         g_value_set_boolean(value, xaccAccountGetSortReversed(account));
+        break;
     case PROP_LOT_NEXT_ID:
         /* Pre-set the value in case the frame is empty */
         g_value_set_int64 (value, 0);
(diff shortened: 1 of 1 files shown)


nielsegberts:
I noticed CI did not pass, previous versions of GCC don't support this flag. I think I can work around that in the cmake config.

gjanssens:
Interesting.

Is -Werror=implicit-fallthrough and [[gnu::fallthrough]] supported by clang as well ?
And do we want this as error or as warning ?
And related - as we have -Wall, isn't implicit-fallthrough then automatically set as a warning when it exists ?

nielsegberts:
The implicit-fallthrough check is not part of -Wall, but is part of -Wextra.

It is supported by clang, but had to replace it with [[fallthrough]] (C++17) so it works with both. But then gcc 4.8 throws a warning which is treated as an error unless we disable unknown-attributes. I could not get it to work the way I wanted to.

I have changed the commit to just add the two break statements that I think were missing.

gjanssens:
Yes, for unstable we only accept C++11 features. Once 3.0 is released we'll bump this to at least C++14 and depending on the target platforms we decide to support, maybe even to C++17.
So for now the extra check can't be applied universally yet. But thanks for running the tests and proposing a fix. I have applied your PR to unstable, though github doesn't seem to be aware of this.

PR #295, opened 2018-03-04
Add spacing between border and text in "New Accounts Hierarchy Setup"

DiMan:
There is no spacing between the text and the frame in the dialog "New File" > "New Accounts Hierarchy Setup". In some cases - when changing the window size - the text appears next to the frame. It does not look very clean. As a solution, I have defined here a small spacing: 5px.

Example 1:
1_margin_step_1

Example 2:
2_margin_step_2

### DIFF
--- a/gnucash/gtkbuilder/assistant-hierarchy.glade
+++ b/gnucash/gtkbuilder/assistant-hierarchy.glade
@@ -14,6 +14,8 @@
       <object class="GtkLabel" id="intro_page_label">
         <property name="visible">True</property>
         <property name="can_focus">False</property>
+        <property name="margin_left">5</property>
+        <property name="margin_right">5</property>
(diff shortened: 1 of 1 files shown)


gjanssens:
Thanks!

fellen:
@gjanssens No backport?

gjanssens:
No. To start this is a gtk3 related fix. maint is still on Gtk2 where the assistants look completely different.

Secondly I consider maint in the "extended maintenance" phase of its life cycle, which means I only consider important bug fixes still. I make exceptions if someone submits a PR specifically against maint, but otherwise I won't bother any more. We need all our energy and time to get 3.0 ready, which in theory should be released in a few weeks. I have only a few days left because I'll be afk for two weeks starting Sunday so I'm starting to feel the pressure...

PR #296, opened 2018-03-05
scheme progress - bugfix ~s, modernise colours, use xhtml <br/>

christopherlam:
- Longstanding 'bug' in aging.scm - only matters if there are unpaid invoices dated earlier than 1970. It's not worth fixing in maint is it?
- `<br>` to `<br/>`
- `~s` to `~a` in (format)
- modernize colours to new red/blue/green from http://clrs.cc/ - from http://i.imgur.com/81U9huO.png to http://i.imgur.com/RocVPUz.png :)


christopherlam:
I've also added some changes to borrowed -- but I can't understand CMakeLists.txt.
- move jqplot to borrowed, and upgrade to latest 1.0.9 - most plugins are unchanged, some new ones
- add guile-json

- jqplot would still be copied to ${CMAKE_INSTALL_DATADIR}/gnucash/jqplot wouldn't it?
- guile-json - we must be able to (use-modules (json builder)) from any scheme file, how to add this external module into the build dir?

gnucash/report/locale-specific/us/taxtxf.scm
```diff
@@ -3452,29 +3449,29 @@
                         (gnc:html-markup-p
                          (gnc:html-markup/format
                           (string-append
-                             "Selected Report Options:<BR>"
+                             "Selected Report Options:<BR />"
```
gjanssens: While you're changing the html BR notes, can you lowercase them at the same time ?

christopherlam: Ok. There are lots of <br> elsewhere which I haven't touched because they're translatable.

christopherlam:
I think these are clean and I'd hope the colours are welcome!

gjanssens:
Thanks!
I agree it's not worth backporting the date issue to maint.

christopherlam:
Ok. Perhaps I'll rebase any further scheme work onto mas ter instead, unless I come across any glaring bugs.

gjanssens:
Yes, that's preferable.

PR #297, opened 2018-03-05
Upgrade scheme dependencies

christopherlam:
1. Move jqplot to borrowed
2. Upgrade it to jqplot 1.0.9
3. Add guile-json-0.6.0 to borrowed

I'm not sure how to fix CMakeLists.txt and would value expert guidance :)

Ideally jqplot would still be copied in the same location in builddir
And guile-json would be accessible to all scheme programs.


christopherlam:
jqplot can be removed. I've completed jqplot -> chartjs transplant. We won't need these commits anymore.

PR #298, opened 2018-03-05
Fix CPack package version number

TheBiggerGuy:
Currently running "ninja-build package" produces a file with and incorrect
version number. This patch passes the CMake version variables into CPack to get
the correct filename.

Old "gnucash-0.1.1-Linux.tar.gz" new "gnucash-2.7.5-Linux.tar.gz"


jralls:
The supported way to generate tarballs is ```ninja dist```. Unless a CPack-generated tarball runs SWIG and includes the resulting C files in the tarball, such a tarball will be unbuildable.

gjanssens:
I was wondering about that indeed. I did merge the PR because there were other CPACK related commands in our scripts already. If that was already broken, it won't be broken more by setting the version number straight.

Oh, wait, cpack makes a tarball from the installed directory, not from the sources! So it serves a different purpose which we don't use ourselves. Perhaps it's useful for distro packagers though.

jralls:
That strikes me as even more misguided because unlike an RPM or a .deb a tarball carries no dependency information. Looking over the docs it appears that there's more to getting CPack to work correctly than simply setting the version and generator. ISTM we should get @codesmythe involved.

codesmythe:
The CPack stuff is left over from cstim's original CMake files. We should remove it unless someone wants to make the effort to to get it to work.

TheBiggerGuy:
Totally agree with the remove it all or get it working sentiment. I am starting to get my head around CMake, and this projects use of it, so can't say how easy it is to get CPack to work. However the next issue I have found #299 is a generally useful cleanup I believe. So there may be merit in me progressing even if I give up and it is removed.

If I do get CPack working would the project be happy with it as a replacement for the custom source tarball rules? And would the project be interested in using it to generate binary distributions tar/deb/rpm/osx-installers/win-installers?

ps. Do people want a bugzilla issue to keep it github?

jralls:
ISTM we can leave this issue here.
Your #299 is utterly the wrong thing to do as I explained when I closed it just now.

What I propose is to remove completely the CPack stuff for now. If you're really interested in making CPack work, then you can present a new PR with a complete implementation. As a minimum it should build an RPM and a Deb that successfully install on recent stable releases of the respective distros. It should *not* make binary tarballs of any sort, that's just asking for dependency hell.

PR #293, opened 2018-03-03
Revision and extension of accounts for the Russian language.

DiMan:
Revision and extension of accounts for the Russian language.
In addition, replacing of commodity: USD => RUB.

PR #300, opened 2018-03-05
Adjustment of Account Hierarchy Templates.

DiMan:
Adjustment of Account Hierarchy Templates.
(mentioned in PR #293 on GitHub)

1) Added namespace section (header)
2) Added missing placeholder

After adjustment, I checked it with **xmllint** => no errors.

Two questions:
* I use two versions of GnuCash: de and ru. I have seen that in the German version the most files haven't definition of namespace section. Should I adjust the "de"-templates? How about "C"-templates? 
* There are three comments at the end of each hierarchy template file. Do they have any meaning? Can they be removed?
`< !-- Local variables: -->`
`< !-- mode: xml        -->`
`< !-- End:             -->`

### DIFF
--- a/accounts/ru/acctchrt_auto.gnucash-xea
+++ b/accounts/ru/acctchrt_auto.gnucash-xea
@@ -1,14 +1,46 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<gnc-account-example>
-    <gnc-act:title>
-      Автомобиль
-    </gnc-act:title>
-    <gnc-act:short-description>
-      Расходы на автомобиль
-    </gnc-act:short-description>
-    <gnc-act:long-description>
-      Счета для учёта расходов связанных с автомобилем.
-    </gnc-act:long-description>
+<gnc-account-example 
+  xmlns="http://www.gnucash.org/XML/"
+  xmlns:act="http://www.gnucash.org/XML/act"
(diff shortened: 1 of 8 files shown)


fellen:
Good questions!
* I just created to get some consent about your first question. Until then I would use namespaces in new files and leave old files untouched.
* I *believe* they are from one of the dozen xml-modes for emacs or some other multiversal editor. Perhaps a real emacs user can answer this?

Your template looks well.

DiMan:
Thanks!

Am 06.03.2018 6:04 vorm. schrieb "Frank H. Ellenberger" <
notifications@github.com>:

jralls:
Yes, the comments at the end are to help emacs recognize that the files are xml since they don't have 'xml' as an extension. IMO it makes more sense for an emacs user who frequently edits these files (a population that I believe numbers around 0 because *nobody* edits these files frequently) to just add 'gnucash-xea' to the auto-mode-alist for their preferred xml mode. Occasional editors can just m-x xml-mode when they open the file.

That's a long way of saying that removing the comments is OK.

jralls:
If an xml file uses namespaces it *must* declare them. Without the declaration it is not well-formed xml. The fact that our current parser is broken--the xml spec says that parsers should abort immediately if they encounter ill-formed xml--is immaterial.

PR #301, opened 2018-03-06
Increase default options dialog size

christopherlam:
The old 400x400 was woefully inadequate especially options involving account trees.


christopherlam:
It would seem the better approach would be to make the options window size responsive to the contents but I have no idea how to do this.

jralls:
I'm a bit concerned about hard-coding dialog sizes to bigger than 640x480. We've had complaints from users with small or low-res screens of dialog boxes being too big to fit.

PR #302, opened 2018-03-06
Fixes for a couple of bugs 

Bob-IT:
Bug 793699, I could not get it to fail but have refactored so the start_spath is defined in the while loop and freed also in the while loop so should fix the issue.

Bug 794031, this looks like it has not worked for a very long time but I can see no reason for it not to work so have enabled the cell render toggle for the placeholder column and provided a call back. Looks OK with my testing but maybe I have missed something, the other way to clear this bug is to delete the surplus text.

Bob


jralls:
The first one could have been fixed by simply NULLing the declaration, but moving it into the scope where it's used and declaring it with an initializer to the right value is better, plus it fixes a leak. Plenty more room for cleanup there, unfortunately as usual.

Thanks!

PR #303, opened 2018-03-06
Remove the emacs comments at the end of files

DiMan:
Remove the emacs comments at the end of files:
`<! -- Local variables: -->`
`<! -- mode: xml        -->`
`<! -- End:             -->`

This comments are to help emacs recognize that the files are xml
since they don't have 'xml' as an extension (see PR #300).


fellen:
Thanks!

I just run 

> file accounts/ru/acctchrt_common.gnucash-xea
accounts/ru/acctchrt_common.gnucash-xea: XML document text

So file's mime magic can still determinate the type, because it starts with "<xml [...]"

PR #304, opened 2018-03-07
Fix Segmentation fault when going to File->New

Bob-IT:
When selecting the File->New with existing register pages open a crash
can happen as the register pages get closed forcing a page-changed
signal to queue the page focus function when there is no page. Add a
call to remove the page focus function in the ..destroy_widget function.

Also with the addition of g_idle_remove before the add when page-changed
is emitted the page focus function only gets run once.


jralls:
Thanks!

PR #305, opened 2018-03-07
Clean Up for Account Hierarchy Templates: "C" (english)

DiMan:
Clean Up for Account Hierarchy Template: "C" (english):

* Add XML namespaces
  The parser in GC 2.6 makes no use of it, but future versions might change.
  Defining the namespace has the benefit, you can check the syntax of template files with `xmllint`:

 * Remove emacs-comments at the end of files:
  `<! -- Local variables: -->`
  `<! -- mode: xml -->`
  `<! -- End: -->`

 * Add `<gnc-act:exclude-from-select-all>1</gnc-act:exclude-from-select-all>` for `acctchrt_full.gnucash-xea`
   This template should not be selected if the user clicks the button "Select All",
   because it consists of all other non-business templates, which would be selected with "Select All".
   BTW: This template is currently not active.

If these changes are not ok, please just cancel this PR. Thanks in advance!
For more Info see PRs #293, #300 and #303.

### DIFF
--- a/accounts/C/acctchrt_brokerage.gnucash-xea
+++ b/accounts/C/acctchrt_brokerage.gnucash-xea
@@ -1,14 +1,47 @@
 <?xml version="1.0" encoding="utf-8"?>
-<gnc-account-example>
-    <gnc-act:title>
-     Investment Accounts
-    </gnc-act:title>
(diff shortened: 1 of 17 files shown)


fellen:
Thanks!

PR #306, opened 2018-03-07
 Clean Up for Account Hierarchy Templates: "de_DE" (german)

DiMan:
Clean Up for Account Hierarchy Templates: "de_DE" (german):

* Add XML namespaces
  The parser in GC 2.6 makes no use of it, but future versions might change.
  Defining the namespace has the benefit, you can check the syntax of template files with `xmllint`:

 * Remove emacs-comments at the end of files:
  `<! -- Local variables: -->`
  `<! -- mode: xml -->`
  `<! -- End: -->`

 * Add `<gnc-act:exclude-from-select-all>1</gnc-act:exclude-from-select-all>` for `acctchrt_full.gnucash-xea`
   This template should not be selected if the user clicks the button "Select All",
   because it consists of all other non-business templates, which would be selected with "Select All".
   BTW: This template is currently not active.

If these changes are not ok, please just cancel this PR. Thanks in advance!
For more Info see PRs #293, #300 and #303.

### DIFF
--- a/accounts/de_DE/acctchrt_auto.gnucash-xea
+++ b/accounts/de_DE/acctchrt_auto.gnucash-xea
@@ -1,14 +1,47 @@
 <?xml version="1.0" encoding="utf-8"?>
-<gnc-account-example>
-    <gnc-act:title>
-      Autobesitz
-    </gnc-act:title>
-    <gnc-act:short-description>
-      Ausgaben von Autobesitzern
-    </gnc-act:short-description>
-    <gnc-act:long-description>
-      Diese Auswahl erstellt Konten, die Ausgaben für ein Auto repräsentieren.
-  </gnc-act:long-description>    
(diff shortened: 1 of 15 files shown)


fellen:
Thanks!

Can you apply this changes also on the other locales? Then the templates would be consistent and we could call xmllint by 'make check'.

fellen:
Just a note:

> Add <gnc-act:exclude-from-select-all>1</gnc-act:exclude-from-select-all> for acctchrt_full.gnucash-xea

 is not required, because it is usually after modularization no longer in the user distributed files (account_DATA), but in EXTRA_DIST for translators and maintainers. See the respective `*Make*.* `files.

DiMan:
Hello fellen,

I added xml namespaces for all other languages. See PR #311.

Thanks, DM

PR #299, opened 2018-03-05
Export compiled GSchema when using CPack

TheBiggerGuy:
This uses the already generated compiled GSchema file, from the
"compiled-schemas" target, for installation. Previously the compiler was
rerun at install time.

PR #307, opened 2018-03-07
Add comment to CMake GSchema build

TheBiggerGuy:
After a pull request would have reinroduced a bug it may be wise to add
a comment to explain why the system is as is.

https://github.com/Gnucash/gnucash/pull/299
https://bugzilla.gnome.org/show_bug.cgi?id=792884


jralls:
No, that's what git blame is for. GitHub even provides a handy button so you can see the blame for any file.

PR #308, opened 2018-03-09
Fix some gtk transient parent warnings

Bob-IT:
Some fixes for GTK transient parent warnings that I have come across.
Bob


jralls:
Yay! Thanks!

PR #310, opened 2018-03-11
Some for missing transient parent warnings

Bob-IT:
These are mainly to do with the Qif, OFX, CSV importers.


Bob-IT:
Added some more fixes to transient parent warnings which has partly made me to change the new-user dialog to use a GtkWindow. No strings were changed just the container so I think it should be OK.

PR #312, opened 2018-03-12
update GC version in NEWS

DiMan:


jralls:
Sigh. Thanks. Since it's gotten delayed I've taken the liberty of changing the date as well, which means that github won't notice the merge.

DiMan:
thanks!

PR #314, opened 2018-03-16
Allow creation of prices in Python bindings

tomason:
Using the function gnc_price_create and book as a parameter, it is possible to create the new GncPrice object. This will remove the necessity of cloning the prices from existing ones in Python scripts.


jralls:
Thanks!

PR #315, opened 2018-03-16
Bug 791831 - Add python3 support

juw:
This switches swig to use python 3 specific features when generating the
bindings, switches the build over to python 3 and makes the neccesary
python 2 to 3 conversions in the bindings and tests.


jralls:
Merged, will be in 2.7.7. Thanks!

PR #316, opened 2018-03-16
[4.x] jqplot->chartjs and upgrade reports

christopherlam:
**This PR will be slowly deconstructed**

~I'm opening this PR for an alpha review - it won't be ready to merge yet. This is mainly the fruitful result of experimenting on html-barchart and linechart.~

~I'm not entirely sure how to add borrowed/guile-json onto GUILE_LOAD_PATH. In my local machine I've installed guile-json onto /opt/share/guile/site, hence the hard-coded path in html-chart.scm. I think this is the main reason it'll fail on travis (also the tests haven't been updated for html-chart!). Please let me know how to allow html-chart.scm to find guile-json libdir.~

~Barring this hurdle,~ this PR will:
1. ~install guile-json as a general purpose scm<->json converter.~
2. ~remove jqplot and replace it with the more flexible and reponsive chartjs~
3. ~merge html-barchart/linechart/scatter/piechart all into html-chart.scm~
4. html-chart will encompass common options and create the charts.
5. options have been chosen to mimic jqplot style as far as possible (colours, fonts etc)
6. ~restore the URL facility allowing inter-report jumps and jumps to the account register.~
7. ~appropriate refactoring of standard-reports' piechart/barchart/linecharts to access the html-chart api~.
8. copy some existing charts into the Utility menu, clearly labelled 'experimental'

### Notes
1. ~installation currently requires guile-json in /opt/share/guile/site~
2. url feature is currently hacked via a fixed-position `<a href>` anchor - I cannot seem to trigger the Gnucash calls to gnc:register and gnc:report to create new reports on the fly. For now the `<a href>` seems the least intrusive method to trigger these jumps.
3. html-piechart/barchart etc still exist - there are numerous reports using them.
4. ~I can't see *any* rationale for using doubles - I'll eventually refactor to pass scheme numbers directly into guile-json which will fill the data fields correctly.~
5. commits to test-graphing is not exactly reachable by UI but is included as an illustration how to change code. This is probably buggy because I cannot test it!


christopherlam:
Ideally I'd like to know how to get webkit to call the url handler described at - https://github.com/Gnucash/gnucash/blob/master/gnucash/gnome/top-level.c#L114

borrowed/guile-json/CMakeLists.txt
```diff
@@ -1,9 +1,11 @@
...
-GNC_ADD_SCHEME_TARGETS(guile-json
+SET(guile-json_DATA
   json.scm
   json/builder.scm
   json/parser.scm
   json/syntax.scm
 )
 
-# SET_DIST_LIST(guile-json_DIST CMakeLists.txt ${guile-json_DATA})
+INSTALL(FILES ${guile-json_DATA} DESTINATION ${CMAKE_INSTALL_DATADIR}/gnucash/guile-json)
```
christopherlam: I'm having real problems trying to add guile-json into the appropriate folder where it can be accessed via `(use-modules (json))` - I think it's because I'm not setting the appropriate GUILE_DEPENDS in standard-reports. Can anyone help me set the right folders?

gjanssens:
As the report and register url handlers are both properly registered, webkit should automatically handle them.

The way to generate proper links in your html files is by using the functions found in [html-utilities.scm](https://github.com/Gnucash/gnucash/blob/master/gnucash/report/report-system/html-utilities.scm#L59) .

gjanssens:
I can look at properly installing guile-json after gnucash 3.0 is out.

christopherlam:
Changed base ('target') to maint. I think I'll create another PR to attempt install guile-json; it'll be cleaner.

christopherlam:
As described on IRC https://code.gnucash.org/logs/2018/04/23.html#T18:35:00 it seems that stripping all escaped-strings is acceptable, however, I don't have a Windows build to test https://bugzilla.gnome.org/show_bug.cgi?id=794965 or https://bugzilla.gnome.org/show_bug.cgi?id=721768 -- I think if anyone wishes to provide a windows build to test, all it needs is adding 4 .scm files from guile-json somewhere into guile-load-path and it should compile and run reports correctly.

christopherlam:
I could do another PR which encompasses most refactoring *except* the introduction of chartjs? However I'm not sure if the URL facility can be restored.

christopherlam:
This series amended - it adds html-chart API which constructs a chartjs-based chart, and leaves the old JQPlot-based API alone. Any custom report which uses old API should still work. Perhaps in 1 year or so, jqplot can be removed, and old API will map into new one.

jralls:
OK, so it goes to maint, not master, and we remove the jqplot API in master after this gets merged up. Otherwise we're stuck with dragging the jqplot stuff around until 4.900.

There seem to be a lot of code-cleanup changes (e.g. net-charts.scm line 208) mixed in with the API changes, though. Those should have separate commits.

christopherlam:
https://github.com/christopherlam/gnucash/tree/maint-chartjs-upgrades is work so far... so html-bar|line|pie|scatter.scm are now mapping to html-chart, so that most existing reports are unchaged! Only price-scatter.scm needs minimal adjustment.

christopherlam:
So, I'd love to merge this in, and I'm 99% sure the existing charts API is not significantly changed.
html-chart API is introduced, but I'm not sure the exact html-chart API names/range of features to expose. Perhaps if it is merged as-is it can be refined in time?

christopherlam:
> So, I'd love to merge this in, and I'm 99% sure the existing charts API is not significantly changed.
> html-chart API is introduced, but I'm not sure the exact html-chart API names/range of features to expose. Perhaps if it is merged as-is it can be refined in time?

Non-destructive amendments. Adds html-chart api which is hopefully not frozen yet.

christopherlam:
Hi

Thank you for reviewing. I have designed html-chart aiming for maximum compatibility with jqplot-based API first and foremost, then secondarily for clarity of code and ease of maintenance.

The first aim is still a work in progress - html-bar/line/chart/scatter have broadly similar features (e.g. gnc:html-*-set-title!) whereas some differences exist (e.g. gnc:html-linechart-set-markers? vs gnc:html-scatter-set-marker?) with slightly different meanings. Therefore any combining of API will necessarily lead to compromises (e.g. x/y charts have col-colors whereas pie has pie-color; so we compromise on col-color), and occasionally breakage (e.g. jqplot scatter marks have 'x' whereas chartJS has 'cross' etc... and jqplot has minor/major grid toggles whereas chartjs has x/y grid toggles).

I would love some help in reviewing compatibility -- old reports will necessarily use old API calls (e.g. html-barchart-set-title!) rather than newer ones (eg gnc:html-chart-set-title!), therefore, the migration path will be: (1) create and finalise html-chart and freeze the API, (2) modify standard-reports charts to call html-chart instead, and (3) modify html-chart/line/etc to call html-chart instead, (4) finally we can remove jqplot.

As you can see, setting an API is a very difficult problem and I do not think that API must be changed arbitrarily once it's published unless there are glaring errors or upgrades (see 3e9cd1fc1170165299a1fe30c434825444eeab2a as an example of a safe upgrade)... occasionally the API designer makes mistakes which are then worked around by custom report writers. This is why tests exists -- to prove that API is consistent.

All above IMHO, IANAP, YMMV etc.

If anything I'd think the test creation process must be documented somewhere I think.

christopherlam:
>     * clone your repository and send PRs to you?
> 
>     * wait until this PR is merged into maint and create new PRs on top of gnucash/maint?

Oh I forgot to answer... if PR is created in my branch then we lose benefit of assistance by core devs... so, perhaps you can clone and I can apply specific commits... but perhaps best converse in here?

christopherlam:
I have had a thought about this line of work.

Instead of trying to achieve 100% old-reports compatibility, I think it will be equally thoughtful to attempt achieve 100% parity with chartjs functionality... this means all API names and parameters will reflect exactly what ChartJS has to offer, and this module named 'html-chartjs.scm'.

My reasoning is, any old-report compatibility layers will be coded in the various html-bar/line/scatter/piechart.scm, and any legacy report can call them and expect a decent chart.

So, the migration path will be:
1. code html-chartjs.scm (and incorporate table-modifier fixes discovered by @GnuFiBux )
2. convert standard-report charts to call html-chartjs api instead of html-bar/line/scatter/pie
3. convert html-bar/line/scatter/pie to use html-chartjs instead... any jqplot-specific functionality which is not present in chartjs will just lead to a deprecation warning and a NOP. Any legacy reports which expect buggy behavior in chart table modifier can be translated enroute to calling html-chartjs api. Therefore API stability is maintained.
4. remove jqplot.

The benefit here is, a future upgrade path to another chart engine (e.g. chartist.js) can be done by chart.js->chartist.js compatibility layer too.

GnuFiBux:
+1 from me

one comment though:
keep the naming html-chart.scm if possible (as currently done).
In case for a later move to something else html-chart.scm might be the place to do the transformation in code, and no messing with the build system (i.e. update of make files) is needed just to rename html-chartjs.scm to html-chart<whatsoever>.scm.

christopherlam:
> Regarding API:
> If I did not overlook anything, I see that the renderer is called with a second parameter "doc" which is not used. Should it be removed?

I don't think so. All renderer functions (eg gnc:html-table-render), (gnc:html-text-render) will get called somewhere with 2 parameters.

I will be keen to see your data modifier tests.

I will slowly modify the API to reflect ChartJS capabilities. Reports can choose to use them. This will take time.

gjanssens:
I agree with @GnuFiBux it would make more sense to name the interface html-chart.scm to make it independent of the underlaying charting engine.

Next with respect to the API to expose I would suggest to be pragmatic about this. That is start by exposing what is currently used and expose extra api on an as-needed basis. This will reduce the maintenance burden and may be less work if you decide to switch to yet another charting engine in the future.

And with regards to your migration plan: That looks ok. Just remember I expect the current reports to remain untouched on the maint branch until users had the chance to test the new charting engine. So plan to expose the migrated charts in an Experimental or Tech Preview or whatever menu for a couple of maint releases.

christopherlam:
Ok this html-chart will slowly be amended.

The html-chart object is currently IMHO a nice elaborate hack generate options which try to mimic jqplot. I'll see about unhacking it so that it has no/minimal conditionals in the options generator, which will improve ease of maintenance.

GnuFiBux:
I have not translated the tests for bar/line charts. Pie and Scatter would be next.
This includes amendments in html-chart.scm to make the charts use tables from html-tables.scm, so table handling can be re-used and does not require extra logic.
If you would like to see an intermediate result, let me know, then I can paste the files here or upload a branch into my repository so you can have a look there.

At the moment I let all tests fail where I think rework is required (no matter in functional code or in test code). Shall I do changes to make everything pass so it can go into maint and create bugs accordingly so the issues won't get lost?

gjanssens:
@GnuFiBux Please work in small steps. In particular adding lots of tests that fail is backwards. You can add plenty of tests that pass for code that is already correct. For incorrect code, please write your commits in such a way that each commit introduces one test together with the required code changes to ensure the test passes. We're not really interested in failing tests in the repo. They are only distracting when we have to bisect commit history.

GnuFiBux:
@gjanssens Don't worry, I did get the point from the previous discussons: as soon as there is any chance that code will be taken into PRs the tests will pass.

GnuFiBux:
@christopherlam During the rendering testing I discovered that the rendering only works for all datasets if the number of ticks is equal or greater as the number of datasets.

You can see the effect if you take the test (test-bar-chart-rendering) and remove the last row label "date3". When running the tests, from the test failure output you see that dataset3 is missing.

I think the reason is in the renderer in html-chart.scm in line "(map (lambda (rownum col collabel color)".
The mapping stops after the shortest list has been full processed, this case (gnc:html-chart-row-labels chart).
I assume we hit a similar problem with (apply zip data) if the data table is not "square" (i.e. the number of rows does not match the number of columns).
By that, also the definition of colors becomes mandatory, which should be avoided if you ask me.

I haven't thought of a solution yet, but I think it is worthwhile mentioning already now.

christopherlam:
Hi @GnuFiBux thank you - this is a nice set of tests for html-chart, and it's clever to reuse html-table data table modifiers. I still have concerns that old API may call these data modifiers using different data and not sure how it should act; strict backward compatibility would mean html-bar/line/chart/scatter keep the old data modifiers, or use a translating function.

The question of an incomplete data table or colors is an interesting one too. During design of html-chart I was mainly testing by running current reports. As we know the current standard reports will always send a complete data set - colors, 'rectangular' data tables (via appropriate lengths append-column! calls), row/col labels. How should html-chart react to such inconsistencies?

My current views: html-chart should be agnostic. It should pass on any incomplete data to chartjs and let chartjs handle it. I haven't coded it this way because I simply never thought about it. My ideal is that html-chart is just a transparent converter from scheme objects to javascript code, and every standard report will be reviewed to send correct chartjs data objects. The only issue is old reports which remain tied to html-bar/line/chart/scatter and jqplot idiosyncrasies.

In the html-chart header section, the datatable is described as a direct representation of a spreadsheet (i.e. cols=series, rows=dates), and I have a strong opinion that it's the wrong way round... instead of meaninglessly adding "rows"/"columns" to a chart data table, and setting row/col labels, a report should really be adding a dataset consisting of "series" and setting "series-labels". This merely requires renaming of functions and adds more meaning to the API. The dataset as follows will not require `(apply zip data)`.

````
(list (list 1.00 2.00 3.00) ;from acc-1
      (list 4.00 5.00 6.00) ;from acc-2
      (list 7.00 8.00 9.00));from acc-3
````

So, I'm quite tempted to suggest the following shortened migration path.
1. Finalise html-chart to be a simple wrapper for chartjs
2. Copy existing standard-reports charts to experimental-reports, review all API calls to send appropriate data to chartjs. Advise users strongly to *not* save experimental reports.
3. When ready, rename standard-reports charts to legacy-reports and rename experimental-reports charts to standard-reports. As long as the options are unchanged, the new charts will just take over the old ones.

Lastly the GNU coding style strongly suggests avoiding dangling parens (minor nitpick)! https://www.gnu.org/software/emacs/manual/html_node/elisp/Coding-Conventions.html#Coding-Conventions

**Comment by GnuFiBux** [comment]
hi there,
I have added now the tests for pie charts and scatter charts.
I also cleaned up a bit, so there is now a test for the common parts, and for the specific charts it is merely about checking the charts options.

My understanding is that I have gone as far as I should at this point of time, no further updates/extensions/improvements planned. Please let me know if there is anything else that I can help with.

Regarding table setup, exchanging the meaning of rows and colomns: I do not have a strong opinion here, but for me it is not difficult to work with the current design -> for me no need for a change.

Regarding the 'rectangular' tables:
My understanding is that with the current design it is html-chart.scm that is adding a limitation here. Having 'non-rectangular' tables does not mean the tables are incomplete. I agree to the idea that html-chart.scm should be as agnostic as can be. But at the moment that is not the case - nothing urgent, but to keep in mind for later design enhancement.

**Comment by christopherlam** [comment]
While html-chart is still being debated, would it be wrong to actually merge in chartjs first? Then they'll be plenty of time to decide how to implement html-chart separately.

My thoughts now are that html-chart must be a minimal wrapper with little/no error checking, and reports will need to change to  chartjs lingo.

**Comment by christopherlam** [comment]
Ok this PR will be deconstructed further. I now think the above code is an elaborate, but a monster hack to fake compatibility with old charts. And I think some will conside it stinky code.

1) Merge chartjs .js files somewhere accessible to reports.
2) I will consider the final html-chart.scm API -- a very skinny wrapper to chartjs -- with a very strict minimum number of conditionals. This means creating html-chart tests will be very easy. As far as possible, mimic previous jqplot look&feel for colours, line widths, font sizes etc.
3) I will slowly offer to modify each chart report to use html-chart instead of html-bar/line/scatter/pie and set the options carefully to create a decent chart. Prior to merging, I'll canvas gnucash-user for views on look&feel.
4) html-bar/line/scatter/pie will then be gracefully deprecated, and removed one day. Then jqplot can be removed. I'll miss it and its idiosyncrasies!

I could alternatively clean up all 4 charts and prettify all of them with assistance of guile-json, and maintain jqplot which is now obsolete... but I think time must move on, urls must be resurrected, and gnucash charts must modernize.

Good plan?

**Comment by gjanssens** [comment]
Other than that I'm fine with your plan of action.

**Comment by gjanssens** [comment]
@GnuFiBux I haven't studied them in detail yet, but your tests look quite clean now.

**Comment by christopherlam** [comment]
> About commit [2e4895d](https://github.com/Gnucash/gnucash/commit/2e4895d4c7cc0579582c59c8d579bb17835075e2):
> If I understand you correctly you're replacing tests for the old api with tests for the new api. I'd prefer to see tests _added_ instead of replaced.
> Only when the old api is dropped the accompanying tests should be replaced.

Hi @gjanssens, as it stands, test-graphing is actually dead code. I've never bothered to maintain it because when I started to look at charting, it has never complained.

This commit served merely as an example for calls to charting engine. I tried enabling it in CMakeLists.txt and can't seem to run it from Reports/Sample&Custom menu. Moreover from my memory of TDD the strict order of operations is (1) write tests (2) write code to pass tests :smile_cat: -- in any case this set of commits will be redone when I have more time in the next few months, and I will aim to (1) merge in Chart.js (2) create/build a skinny html-chart.scm incrementally, building api *as the reports require* instead of a 'grand intelligent design'. So, lots of work ahead. (3) I'll also create good tests for html-chart.scm and borrow some of @GnuFiBux ideas for table modification.

The idea for the above plan is that html-bar/chart/line/pie will remain 'frozen/maintainable' until html-chart is fully fleshed out and ready, and all existing reports have been converted safely. And I think is a good plan now.

**Comment by christopherlam** [comment]
Hi all

What I've been generating (in my head while travelling :small_airplane: ) is a method to generate the chartjs options, and expose it directly to reports. The report writer will then modify the options tree to apply customizations, and the chart html generator will then convert it to json string. The advantage here is a very thin wrapper for chartjs, and also *all* chartjs capabilities are then exposed. The disadvantage is that the chartjs options api is not immediately obvious, but I'd hope that the converted reports and the explanations will make sense.

Consider the raw chartjs options object:
```
(list (cons 'type 'bar)
      (cons 'options (list (cons 'maintainAspectRatio #f)
                           ...etc...))
      (cons 'data (list (cons 'labels (list "A" "B" "C")))
                        (cons 'datasets #f)))
```
will be modified via calls as follows:
```
(gnc:html-chart-set-options! chart '(type) 'pie)
(gnc:html-chart-set-options! chart '(options maintainAspectRatio) #t)
(gnc:html-chart-set-options! chart '(data datasets)
    (list (list (cons 'data (list 10 20 30)))))
```

to the result
```
(list (cons 'type 'pie)
      (cons 'options (list (cons 'maintainAspectRatio #t)
                           ...etc...))
      (cons 'data (list (cons 'labels (list "A" "B" "C")))
                        (cons 'datasets (list (list (cons 'data (list 10 20 30)))))))
```

Notice in the definition for `(gnc:html-chart-set-options! chart path val)`, the `path` will be a list of symbols to recursively determine the kvp pair to be modified with the new value `val`.

Sounds good?

**Comment by jralls** [comment]
The code snippet appears to begin with a default set of chart options that are modified by gnc:html-chart-set-options to the final options for producing the report. Is that what you mean?

I don't understand the last bit about KVP pairs. Do you mean to store chart options using slots on some book object (which one?) or are you just saying that options in saved configurations (which go in the user's configuration directory and are available to all books) get stored as key-value pairs?

**Comment by christopherlam** [comment]
> The code snippet appears to begin with a default set of chart options that are modified by gnc:html-chart-set-options to the final options for producing the report. Is that what you mean?

Correct. A brand new chart will have a skeleton options object which sets fonts/colours only. The report writer will modify this options object to add customized data/labels/charttype.

> I don't understand the last bit about KVP pairs. Do you mean to store chart options using slots on some book object (which one?) or are you just saying that options in saved configurations (which go in the user's configuration directory and are available to all books) get stored as key-value pairs?

kvp doesn't mean gnucash's kvp slots. It means a scheme cons pair, whereby the car is a symbol (eg 'data) and the cdr is either a value (eg a boolean, a number, a string, a list of numbers, or another list of pairs). Sorry for confusion. As above options demonstrate, chartjs expects data/options as a nested json list... this can easily be generated effectively by nesting scheme's list of pairs, and transforming this list to json using guile-json.

**Comment by jralls** [comment]
OK, sounds reasonable to me. @gjanssens?

**Comment by gjanssens** [comment]
Same here.

**Comment by christopherlam** [comment]
I've pushed this mid-way job to illustrate how reports will set options.
As an addendum to the above snippets, the `path` list will also support integers to navigate through simple lists. So, the following `path` is also supported.

`(gnc:html-chart-set-option! chart '(scales xAxes 0 gridLines display) #t)`

will set the following
```
(cons 'scales (list
               (cons 'xAxes (list
                             (list
                              (cons 'gridLines (list (cons 'display #t))))))))
```

This is necessary because chartJS supports multiple x- and y- axes.

**Comment by christopherlam** [comment]
Ok. The current dilemma is that jqplot was generating a chart using the call `plot = $.jqplot(id,data,options)` whereby id was the xx in the `<div id=xx>` attribute. `data` and `options` were separate javascript objects. I haven't quite analysed the syntax; but `data` seems to be a list of datasets, each element was a list (label, value); `options` was a complex javascript object whereby `options.legend.show = true;` or `options.series[0].label = "Assets"` were valid statements to set options.

In ChartJS, a chart is generated using the call `chart = new Chart(chartid, options)` whereby `options` contains both chart-options and data. `options.type` was a string pie/scatter/line/bar, and `options.data` was a complex javascript object whereby `options.maintainAspectRatio = true;`, `options.chartArea.backgroundColor = "red"`, `options.data.datasets[0].labels = ["01/01/80", "01/02/80"];`, `options.data.datasets[0].data = [1,2,3];` are used to set data. The full syntax is described in https://www.chartjs.org/docs/latest/

I've designed html-chart to be able to accept a default set of options which mimics jqplot fonts/colours. To allow options to be modified, I wished to create a mechanism whereby the specific option *leaf* is modifiable, hence the `(nested-list-set! obj path newval)` utility function.

The datasets are sent to chartjs via the options object, and are also recorded as nested javascript objects.

The current iteration of account-piecharts-upgraded.scm shows how to send data to chartjs --
```
(gnc:html-chart-set-data!
        chart
        (list (cons 'labels date-iso-string-list)
              (cons 'datasets
                         (list
                          (list
                           (cons 'label (_ "Assets")))
                           (cons 'backgroundColor "#0074D9")
                           (cons 'borderColor "#0074D9")
                           (cons 'fill #f)
                           (cons 'pointRadius (if markers 4 0))
                           (cons 'data (map gnc:gnc-monetary-amount assets-list)))
                          (list
                           (cons 'label (_ "Liabilities"))
                           (cons 'backgroundColor "#FF4136")
                           (cons 'borderColor "#FF4136")
                           (cons 'fill #f)
                           (cons 'pointRadius (if markers 4 0))
                           (cons 'data (map gnc:gnc-monetary-amount liability-list)))))))
```

The current iteration of html-chart will construct the chartjs options via
```
(list
 (cons 'type (gnc:html-chart-type chart)
 (cons 'options (gnc:html-chart-options-internal chart)
 (cons 'data (gnc:html-chart-data chart)))
```

Should the `data` modifier also use the nested-list path syntax?

If yes:
- allows reports to add data incrementally

If no
- much simpler
- but reports must be careful to set data carefully, copying from (future) existing reports

My views: no. Because most reports should be simple to construct, and the use case for incrementally adding data series is tiny.

**Comment by jralls** [comment]
It's a lot easier to see what you're doing if you post diffs instead of whole files.

I find the old version more readable.

**Comment by christopherlam** [comment]
Wouldn't it be more clever to try trim api to a strict minimum? The first style will produce a long list of various html-chart modifiers.

**Comment by christopherlam** [comment]
Ok this sounds reasonable; when you've read ChartJS api as much as I have, they seem very straightforward. Perhaps I'll add these intermediate modifiers as needed and will clean up when all's done.

**Comment by christopherlam** [comment]
Ok a quick review of last commit is warranted.
There will be (for now) an expanded API list. The basic ones are:
```
(gnc:html-chart-set-width! chart '(pixels . 500))
(gnc:html-chart-set-height! chart '(percent . 50))
(gnc:html-chart-set-urls! chart list-of-urls)
(gnc:html-chart-set-type! chart 'line)
(gnc:html-chart-set-title! chart "mega-chart")
(gnc:html-chart-set-data-labels! chart '("Jan" "Feb" "Mar"))
(gnc:html-chart-add-data-series! chart "Income" '(200 210 240) "green"
 'borderWidth 2
 'fill #f)
(gnc:html-chart-set-x-axis-label! chart "Months")
(gnc:html-chart-set-y-axis-label! chart "USD")
(gnc:html-chart-set-currency-symbol! chart "$")
```

and the generic `(gnc:html-chart-set! chart path newval)` will remain as an advanced modifier for people (i.e. @christopherlam for now) who wish to modify any property in the options object.

Note the `(gnc:html-chart-add-data-series!)` modifier can accept further arguments to modify the dataseries according to any ChartJS property eg https://www.chartjs.org/docs/latest/charts/bar.html#dataset-properties.

This is now a lesser hack, and still very flexible.

christopherlam:
There is a last commit to refine the formatting of monetary amounts in javascript.

As we know the (xaccPrintAmount amount print-info) is very good in formatting numbers into locale-specific currencies eg USD 1 = "$1.00", EUR -2.50 = "-€2.50", but is inaccessible to javascript.

html-barchart etc had never worried about currencies; amounts were rendered as numbers in yAxis with a suitable yAxis title "USD".

We can use javascript https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toLocaleString number formatter which can accept a locale + currency to properly format the number, e.g. `var value = 2.5; value.toLocaleString(undefined, {style:'currency', currency:"USD"})` will output "$2.50" -- it will tailor the currency display according to currency and locale. e.g. AUD will show "$2.50" in en_AU locale, but will render as "A$2.50" in C locale.

I think this is a nice addition to charts, but may generate a brand new set of behaviours for charts; users may complain that the numeric amounts are not consistently rendered across GnuCash builds, which is resolved by setting the right locale in the environment. This is a purely Javascript capability, and I'm not 100% sure is available in Webkit1. Perhaps a Win32 build is in order?

The last commit will implement this feature. I was faking it in the past via `var currency = "€"; str = currency + number` which did not produce perfect strings.

gjanssens:
Hmm, it looks like this PR now has conflicts with maint due to me merging PR#391...

jralls:
Only one very minor one, they both add a test to CMakeLists.txt but the changeset on this one doesn't know about the added line from #391.

gjanssens:
Other than that I have reverted locally back to before Carsten's changes and then merged your PR for testing.

In general the reports seem to work fine, so nice job on the backend replacement.
The ability to drill down is a nice new feature, though I hope we can eventually improve the actual implementation. It's not really obvious that one has to click on the "Load" link in order to see effectively jump. I have read your motivations as to why it currently is as it is in an earlier comment. I don't know what's needed to catch a javascript document.location event. On the other hand it may be possible to make the link appear underneath the mousepointer in a floating (or rather absolute-positioned) block element and style that whole block to look like something the user can click on. In the end what the user really clicks on is still a plain old <a href=> type link, but with some css and javascript hackery it can be made be better discoverable.

I also found a first feature regression compared to jquery ;-p
It used to be possible to drag a frame over some portion of the chart and then jqplot would zoom in on the y-axis for that part. This no longer happens. Note it would be better still it it zoomed in on the x- and y-axes, so if chartjs has such a feature that would be a great improvement as well.

gjanssens:
Op donderdag 31 januari 2019 20:59:51 CET schreef John Ralls:
> Only one very minor one, they both add a test to CMakeLists.txt but the
> changeset on this one doesn't know about the added line from #391.

I'll admit I didn't look too deeply at this point as I could easily test by 
temporarily reverting back.

christopherlam:
> Other than that I have reverted locally back to before Carsten's changes and then merged your PR for testing.

Already rebased, thanks magit :)

> In general the reports seem to work fine, so nice job on the backend replacement.
> The ability to drill down is a nice new feature, though I hope we can eventually improve the actual implementation. It's not really obvious that one has to click on the "Load" link in order to see effectively jump. I have read your motivations as to why it currently is as it is in an earlier comment. I don't know what's needed to catch a javascript document.location event. On the other hand it may be possible to make the link appear underneath the mousepointer in a floating (or rather absolute-positioned) block element and style that whole block to look like something the user can click on. In the end what the user really clicks on is still a plain old <a href=> type link, but with some css and javascript hackery it can be made be better discoverable.

I had already considered this :smile:  the `//commented-out` line in html-chart.scm:417 does exactly this. Click the data point and the anchor appears near the mouse pointer. I'll leave CSS hacking to someone else!

> I also found a first feature regression compared to jquery ;-p
> It used to be possible to drag a frame over some portion of the chart and then jqplot would zoom in on the y-axis for that part. This no longer happens. Note it would be better still it it zoomed in on the x- and y-axes, so if chartjs has such a feature that would be a great improvement as well.

Ok personally I never really liked it, hence didn't make effort. chartJS has https://github.com/chartjs/chartjs-plugin-zoom but is rather clunky. I haven't been able to consistently make it work.

I'd welcome further feedback! I'd like to clean up the commits after review so that 1 commit = 1 file, before merging, (and `git diff` = #null).

@jralls I'm fully aware this javascript Intl formatter does introduce another surface for locale-sensitive behaviours, hence I needed to put up a disclaimer here. I'd hope that webkit2 is good enough to behave correctly; webkit1 is simply clueless and uses a simplistic string formatter. If there are enough complaints I'll see how http://openexchangerates.github.io/accounting.js/ works -- it seems well-written. From my testing `LANG=C ./gnucash` `LANG=de_DE.utf8 ./gnucash` the javascript currency seems to work very well.

christopherlam:
Just to clarify ... I would not call this PR mergeable yet -- it needs some win32 beta builds, specifically https://github.com/christopherlam/gnucash/tree/maint-chartjs-budget-barchart which is the same as this branch with a git revert for the gnc-date issues... I've never tested chartjs on win32.

christopherlam:
@jralls or @gjanssens - what would be the status of chartjs changes? is it worth while keeping it updated? if yes I'll keep it up; it not I'll work on fixing the jqplot implementation from time to time.

gjanssens:
@christopherlam I think the first step forward is to get clarification on the licence uncertainty regarding guile-json being GPLv3 only while gnucash is GPLv2 or later.

I have searched around a bit more and according to [this bit on gnu.org](https://www.gnu.org/licenses/license-list.html#GPLv2) as we add the "or later" it looks like we are allowed to combine them. Frustratingly that same bit refers to [the all compatibility faq](https://www.gnu.org/licenses/gpl-faq.html#AllCompatibility) which is slightly ambiguous to me. In the matrix it states both can be combined, but in the footnote is suggests

> If you have the ability to release the project under GPLv2 or any later version, you can choose to release it under GPLv3 or any later version—and once you do that, you'll be able to incorporate the code released under GPLv3.

It is unclear to me whether that means we should explicitly relicense to GPLv3 or later, or if our current license implies the effective license terms that apply are GPLv3 in this combination. The latter would be the best, but I can't determine whether that's what's meant.

For the other option I don't know if we can relicense to GPLv3 only simply because our license supports that or whether we'd need consent of all contributors (active and past) to make that change.

To be complete this whole discussion may be moot anyway because Guile is Lesser GPLv3 licensed and according to the same FAQ matrix that already forces us to distribute (they use the term "convey") GnuCash according to the terms of the GPLv3 license:

> Because GPLv2 does not permit combinations with LGPLv3, you must convey the project under GPLv3's terms in this case, since it will allow that combination.

It's not exactly the same thing so a bit of ambiguity remains.

To get to the bottom of this I have just sent a mail to the fsf's license e-mail address for some clarification.

jralls:
This PR is for chartjs, not guile-json.  chartjs is licensed under one of the so-called MIT licenses. It's very permissive and doesn't interfere with the GPL at all.

I think we should merge this, the only question is whether it should be to maint or master. Either way the change to the root CMakeLists.txt needs to go: If it's going on master then it needs to be rebased onto that branch or it will screw up history again and in that case CMakeLists.txt will already have the right version; if to maint then the version shouldn't change to 3.900.

gjanssens:
Oops I'm sorry I missed that. I thought chartjs implied use of guile-json for option handling.

Are there changes to the save-report interface ? I mean does the switch to chartjs add/remove/change any report options ?

jralls:
I don't think so. I'd thought that guile-json was in #477 but it isn't and now I can't find the PR where it was. Getting #477 into 4.0 would be nice too, though one of us might have to finish it up as   @christopherlam has set out other priorities for himself.

gjanssens:
guile-json has been merged into maint already, but we're not using it yet. Depending on the outcome of the license thing we can push forward with #477 or have to remove it from our code again.
Other than that I agree it would be nice to have #477 in master. Let's see if we can pick it up.

christopherlam:
This PR *does* have a couple option changes in `budget.scm` to ensure it requests budget-period rather than start/end-dates. I think I'll need to move this change to another PR. This PR is purely about chartJS; it doesn't deprecate jqplot supporting files. I'll review and clarify. If need I'll modify this branch so that it performs a chartjs-upgrade suitable for maint, and another PR for master (i.e. budget.scm amendments, deprecating/removing jqplot, etc)

christopherlam:
> Getting #477 into 4.0 would be nice too

I think you're referring to #459 to save accounting-period into book kvp?

jralls:
Yeah, that one too.

christopherlam:
Rebased to master. Changes in this branch:
* all charts upgraded to use html-chart
* all old URL links restored
* cashflow-barchart.scm also gets URL upgrade to run cashflow.scm
* jqplot modules still present (I vote to remove dead code cf. deprecate them)
* budget-barchart also upgraded to select start/end-period instead of start/end-dates -- it is too difficult to extricate this change from chartjs work. This change does not lead to report crash; the start/end dates are simply ignored; it defaults to showing all budget periods. It is near impossible to consider a backward-compatibility path here.

gjanssens:
Looks good to me. Feel free to merge into master.

PR #317, opened 2018-03-17
Add COMPILE_SCHEMA to allow disabling gschema compilation

luc14n0:
Adding the COMPILE_SCHEMA option allows a choice to compile the gschema or not, defaulting to 'ON'. This is useful for distributions that choose to not compile the gschema when producing packages for some good reasons.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -71,6 +71,9 @@ OPTION (ALLOW_OLD_GETTEXT "allow to configure build with a gettext version older
 
 # These are also settable from the command line in a similar way.
 
+# Use gsettings-desktop-schema for a better integration with GNOME
+SET(COMPILE_SCHEMA ON CACHE BOOL "Compile the GSettings schema")
+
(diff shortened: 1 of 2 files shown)


luc14n0:
Let me know if something can be improved.

CMakeLists.txt
```diff
@@ -71,6 +71,9 @@ OPTION (ALLOW_OLD_GETTEXT "allow to configure build with a gettext version older
 
 # These are also settable from the command line in a similar way.
 
+# Use gsettings-desktop-schema for a better integration with GNOME
+SET(COMPILE_SCHEMA ON CACHE BOOL "Compile the GSettings schema")
```
fellen: If there is not already some convention, I would name the option COMPILE_GSCHEMA
to have 2 letters distance from SCHEME stuff

luc14n0: It's a good idea.

jralls:
I merged this in error and reverted it.

Do you really want to disable schema compilation at build? I understood from our IRC discussion the other day that you only wanted to disable installation.

PR #318, opened 2018-03-18
Add COMPILE_GSCHEMAS option to allow disabling gschemas compilation

luc14n0:
This option will allow to not compile the gschemas, defaulting to 'ON'. It's useful for distributions where the gschemas compilation is not handled by the app itself.


jralls:
Merged, after squashing the two commits.

PR #319, opened 2018-03-20
Bug794360 error when deleting accounts with subaccounts

Bob-IT:
Test for valid widgets before trying to get there sensitivity. Also if the register was open when the account is deleted an invalid component_id error is logged so check for it being valid before using.

### DIFF
--- a/gnucash/gnome/gnc-plugin-page-account-tree.c
+++ b/gnucash/gnome/gnc-plugin-page-account-tree.c
@@ -82,6 +82,8 @@ static QofLogModule log_module = GNC_MOD_GUI;
 #define DELETE_DIALOG_TRANS_MAS    "trans_mas"
 #define DELETE_DIALOG_SA_MAS       "sa_mas"
 #define DELETE_DIALOG_SA_TRANS_MAS "sa_trans_mas"
+#define DELETE_DIALOG_SA_TRANS     "sa_trans"
+#define DELETE_DIALOG_SA_SPLITS    "sa_has_split"
 #define DELETE_DIALOG_OK_BUTTON    "deletebutton"
 
 enum
@@ -1286,8 +1288,36 @@ gppat_populate_trans_mas_list(GtkToggleButton *sa_mrb,
 void
 gppat_set_insensitive_iff_rb_active(GtkWidget *widget, GtkToggleButton *b)
 {
+    GtkWidget *dialog = gtk_widget_get_toplevel(widget);
+    GtkWidget *subaccount_trans = g_object_get_data(G_OBJECT(dialog), DELETE_DIALOG_SA_TRANS);
+    GtkWidget *sa_mas = g_object_get_data(G_OBJECT(dialog), DELETE_DIALOG_SA_MAS);
+    gboolean have_splits = GPOINTER_TO_INT (g_object_get_data(G_OBJECT(dialog), DELETE_DIALOG_SA_SPLITS));
+
     gtk_widget_set_sensitive(widget, !gtk_toggle_button_get_active(b));
-    set_ok_sensitivity(gtk_widget_get_toplevel(widget));
+
+    // If we have subaccount splits & delete subaccounts, enable subaccount_trans
+    if ((have_splits) && !gtk_widget_is_sensitive(sa_mas))
+        gtk_widget_set_sensitive(subaccount_trans, TRUE);
+    else
+        gtk_widget_set_sensitive(subaccount_trans, FALSE);
+
+    set_ok_sensitivity(dialog);
+}
+
+static GtkWidget *
+gppat_setup_account_selector (GtkBuilder *builder, GtkWidget *dialog,
+                              const gchar *hbox, const gchar *sel_name)
+{
+    GtkWidget *selector = gnc_account_sel_new();
+    GtkWidget *box = GTK_WIDGET(gtk_builder_get_object (builder, hbox));
+
+    gtk_box_pack_start (GTK_BOX(box), selector, TRUE, TRUE, 0);
+    g_object_set_data(G_OBJECT(dialog), sel_name, selector);
+
+    gppat_populate_gas_list(dialog, GNC_ACCOUNT_SEL(selector), TRUE);
+    gtk_widget_show_all(box);
+
+    return selector;
 }
 
 static void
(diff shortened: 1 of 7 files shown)


gnucash/gnome/gnc-plugin-page-account-tree.c
```diff
@@ -1460,11 +1460,11 @@ gnc_plugin_page_account_tree_cmd_delete_account (GtkAction *action, GncPluginPag
             g_free(acct_name);
             return;
         }
-        if (trans_mas && gtk_widget_is_sensitive(trans_mas))
+        if (GTK_IS_WIDGET(trans_mas) && gtk_widget_is_sensitive(trans_mas))
```
jralls: The first line of gtk_widget_is_sensitive() is
```c
g_return_if_fail(GTK_IS_WIDGET(widget), FALSE);
```
If widget isn't NULL (G_IS_OBJECT checks for NULL and returns false) but is an invalid address then all you've done is move the crash from gtk_widget_is_sensitive() to here.

Where do you think that the widgets are getting destroyed?

gnucash/register/ledger-core/gnc-ledger-display.c
```diff
@@ -903,5 +903,6 @@ gnc_ledger_display_close (GNCLedgerDisplay *ld)
     if (!ld)
         return;
 
-    gnc_close_gui_component (ld->component_id);
+    if (ld->component_id)
+        gnc_close_gui_component (ld->component_id);
```
jralls: Won't this leave the register open with a zombie account?

Bob-IT: I checked that the close_handler in gnc_ledger_display was being called before the gnc_ledger_display_close which unregistered the ld->component_id before it frees the ld so thought this would be OK.

gnucash/gnome/gnc-plugin-page-account-tree.c
Bob-IT: I think I may of confused my self but there is a gtk_widget_destroy on "subaccount_trans" GtkBox which encompasses the hbox that the Gas widget is packed and so assumed that was destroyed also.

jralls: Well, the logic in that function isn't great. If the child accounts have no splits then indeed subaccount_trans gets destroyed, after which sa_trans_mas is created and packed into the destroyed box. trans_mas and sa_mas, on the other hand, are part of other components that aren't destroyed. The invalid pointer that causes the crash in the bug is indeed sa_trans_mas, so perhaps the problem is its unconditional creation; even if it doesn't crash it still leaks. So I think the fix should be to rearrange the subaccount_trans logic: If the subaccount has transactions, then create sa_trans_mas and do the read-only check, else delete subaccount_trans.

Incidentally, GTK_IS_WIDGET(sa_trans_mas) should return TRUE; it's freshly allocated, after all. gtk_widget_is_sensitive is probably crashing as it climbs the chain of parents; parent->parent is undefined after the parent's destruction.

gnucash/register/ledger-core/gnc-ledger-display.c
jralls: What?! The close_handler is supposed to be called by gnc_close_gui_component. It shouldn't get run first.

Regardless, since the close handler doesn't then call component_id = 0 (or better compnent_id = NO_COMPONENT, but you'd have to change this test accordingly) the test won't change anything.

gnucash/gnome/gnc-plugin-page-account-tree.c
Bob-IT: I wondered about changing the order so will have another stab at this.

gnucash/register/ledger-core/gnc-ledger-display.c
Bob-IT: Will need to have another look at this and check the order of things again.

gnucash/gnome/gnc-plugin-page-account-tree.c
Bob-IT: This dialog is strange, the vbox subaccount_trans which contains Gas widget sa_trans_mas is never sensitive / greyed out so there is no point in it being there as far as I can see. I thought it was a mistake on unstable but it is the same on maint.
Should this section "Sub-Account Transactions" get removed from the dialog and associated code or am I missing something ?

jralls: ISTM it's together with the messed-up logic in the if statements: The subaccount_trans box should be made sensitive if there are subaccount transactions.

But IIRC it's a HIG failure to delete GUI elements anyway, they should just be made insensitive. So maybe the gtk_wdget_destroy calls should be removed.

Bob-IT: What I was getting at, if the account you are deleting has sub accounts and you specify to move them why would you move any sub account transactions some where else.

Bob-IT: The dialog comprises of a vbox containing three vbox's , 'transactions', 'subaccounts' and 'subaccount-trans'. What I think should be displayed is if there are transactions in the deleted account 'transactions' is left, if there are sub accounts then you may have 'transactions' and 'subaccounts' or you may have 'transctions' and 'subaccount-trans'.

gnucash/register/ledger-core/gnc-ledger-display.c
Bob-IT: I think I know what is wrong with this, when the register is registered with the component manager you end up with this setup...
Components: id's
  GncPluginPageRegister:	6
  register-single:	                5
  GncPluginPageRegister:	4
  register-single:	                3
  summary-bar:	                2
  plugin-page-acct-tree:	        1

After the account is destroyed, there is a 'gnc_resume_gui_refresh' which then calls 'gnc_gui_refresh_internal' and on line 692 it calls 'find_component_ids_by_class' which is used in order resulting in 'register-single' being unregistered before 'GncPluginPageRegister' and hence 'ld' being freed but the register not knowing this. If I change g_list_prepend to g_list_append in 'find_component_ids_by_class' the list is reversed or use g_list_reverse and so no error.

What I do not know is whether that is a valid fix ?

jralls: g_list_append is slow. It would be better to call g_list_reverse and use the reversed list.
We also need ```ld = NO_COMPONENT``` after it has been freed.

gnucash/gnome/gnc-plugin-page-account-tree.c
jralls: OK. Will it be ready by tomorrow morning so that I can release 2.7.8 tomorrow?

Bob-IT:
Here are the changes I have worked on, the first two replace my original ones. The next fixes a couple of transient parent warnings and the last is a proposed fix for bug 794644. I do not have that many prices but I think it should do the trick but still waiting for confirmation in the bug so commit if you think it is appropriate. If there are any minor changes please update as I probably wont have time to fix tomorrow but if it is rubbish it will have to wait till next week.

jralls:
Thanks!

PR #320, opened 2018-03-25
Bug 794030 relative date functions compute wrong day of month

christopherlam:
Convert `>` to `<` and rename `month-length` to `month-days` for consistency.


jralls:
Thanks!

PR #321, opened 2018-03-26
Bugzilla 792800 fix and eradicate some gnc-numeric in scheme

christopherlam:
1. Fix Advanced Portfolio Report crash caused by negative-stock-balance situation
2. Some schemifying of APR
3. Eradicate gnc-numeric in some more reports.


jralls:
Chris, can you separate out the bug fix or are the other changes prerequisites to that? I'd rather not merge this big a change until after 3.0's release.

christopherlam:
Hmm I guess I could just add an exception handler in the current APR and leave the other parts intact.

PR #322, opened 2018-03-26
Bug 792800 - Advanced Portfolio Report tries to divide by zero balance and aborts

christopherlam:
This fixes a div/0 error identified in bug 792800. It is thought to be related to negative-stock-balance situation rather than high-precision security error. From #321.


christopherlam:
Better overflow error message coming soon.

jralls:
Thanks!

PR #326, opened 2018-03-30
Bug794806 - Autosize too small

Bob-IT:
When double clicking on header to autosize columns need to add the
padding_border values to get correct width.


jralls:
Thanks!

PR #327, opened 2018-03-30
Eradicate gncInvoiceDateExists

christopherlam:
This commit replaces gncInvoiceDateExists calls to gncInvoiceIsPosted. I think it's a reasonable assumption that all posted invoices will have posted-date.


jralls:
Thanks!

PR #328, opened 2018-03-31
Various gwenywfar fixes

Bob-IT:
The first and fourth commits were errors migrating from the gtk2 Gui to the gtk3 version.
The second and third errors I observed while trying to fix the first one which I think are correct.
The last commit fixes a couple of transient parent warnings when doing a MT940 file import.

I do not use the Aqbanking stuff but was able to create a user and account via the wizard, a strange thing when setting the currency for a account is that there are two 'Pound Sterling' options but was unable to set it to either, maybe because it was the default currency, others worked OK.

If the first four fixes are accepted then the gwenywdar library would need to patched on Windows along with the Aqbanking compile fix.

Bob

### DIFF
--- a/borrowed/gwengui-gtk3/w_combobox.c
+++ b/borrowed/gwengui-gtk3/w_combobox.c
@@ -258,10 +258,13 @@ int Gtk3Gui_WComboBox_Setup(GWEN_WIDGET *w) {
   /* create widget */
   store=gtk_list_store_new(1, G_TYPE_STRING);
   if (flags & GWEN_WIDGET_FLAGS_READONLY)
-    g=gtk_combo_box_new_with_model_and_entry(GTK_TREE_MODEL(store));
+    g=gtk_combo_box_new_with_model(GTK_TREE_MODEL(store));
   else
+  {
     /* TODO: why the heck does *this* combo box have two columns in the list?? */
     g=gtk_combo_box_new_with_model_and_entry(GTK_TREE_MODEL(store));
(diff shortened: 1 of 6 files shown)


jralls:
By "patched on windows" you mean that the first four patches need to be pushed upstream and gwenhywfar-4.20.0 needs to have them applied, right?

Bob-IT:
Ultimately yes but if version 3.0 is being released before upstream is patched then the gwenhywfar version 4.20.0 on the build server would need to be patched I assume.

jralls:
That will work for Windows and Mac but not for Linux distros that have already added Gwen4.20.0. Is there a way to test for the bug in CMake and use borrowed/gwenhywfar-gtk3?

Bob-IT:
Do not know, in the mean time would be nice if someone else can test that it fixes the issue, will ponder...

jralls:
Well, I had to apply the patch to gwehywfar because I already have 4.20 installed on my Mac and that's where I have aqbanking accounts to test. Yes, it fixes the problem. I'll go ahead and merge this; if you have time to create a cmake test today, by all means do... otherwise I'll look at it this afternoon after I get a couple of other issues sorted.

cstim:
I've committed those patches to gwenhywfar itself, so that it will be included in a gwenyhwfar 4.21 that is to come sometime in the future.

jralls:
@cstim Thanks. Be sure to let Martin know, I sent him an email with a git patch.

PR #329, opened 2018-04-04
Allow building against a shared libgmock (an option in gtest-1.8.0)

wenottingham:
Fedora builds it as a shared library.


jralls:
Thanks.

PR #331, opened 2018-04-04
Adjust how gnc-version.h is created.

wenottingham:
This avoids 3.0 builds showing up as 'development'.

Someone who lives more in CMake can probably redo this into a much shorter patch, feel free to do so.


jralls:
Unfortunately this approach won't work: gnc-version.h is created only when when building from VCS and the result gets included in the tarball.

I've fixed it a different way, but thanks for the effort!

PR #332, opened 2018-04-04
Remove GNUCASH_MICRO_VERSION from VERSION...

DiMan:
and removing of variable CPACK_PACKAGE_VERSION_PATCH, 
because the variable GNUCASH_MICRO_VERSION does not exist


jralls:
Thanks!

PR #333, opened 2018-04-07
FIX: Fixed compilation error for target gwengui-gtk3

Abschiedsstein:
Compilation led to the error
Scanning dependencies of target gwengui-gtk3
[  0%] Building C object borrowed/gwengui-gtk3/CMakeFiles/gwengui-gtk3.dir/gtk3_gui_dialog.c.o
In file included from /usr/include/gtk-3.0/gdk/gdk.h:30:0,
                 from /usr/include/gtk-3.0/gtk/gtk.h:30,
                 from /home/henrik/gnucash2/borrowed/gwengui-gtk3/gtk3_gui.h:14,
                 from /home/henrik/gnucash2/borrowed/gwengui-gtk3/gtk3_gui_dialog_l.h:14,
                 from /home/henrik/gnucash2/borrowed/gwengui-gtk3/gtk3_gui_dialog_p.h:14,
                 from /home/henrik/gnucash2/borrowed/gwengui-gtk3/gtk3_gui_dialog.c:10:
/usr/include/gtk-3.0/gdk/gdkconfig.h:13:10: fatal error: glib.h: No such file or directory
 #include <glib.h>

I fixed this by adding GLIB2_LDFLAGS to the linked libraries in the
CMakeLists.txt.

The problem was also reported in an email of 04/05/2018 07:34 AM with
the title "Gnucash 3.0 Build error".


jralls:
Your fix doesn't match the problem: GLIB2_LDFLAGS is for the link stage and the error is that it can't find a header. For that particular header you'd normally add ${GLIB2_INCLUDE_DIRS} to TARGET__INCLUDE_DIRECTORIES, but if your system is configured correctly that directory (/usr/include/glib-2.0) should be in GTK3_INCLUDE_DIRS; the GLIB2_LDFLAGS should similarly be part of GTK3_LDFLAGS.

Note as well that we run CI on Arch and it doesn't fail this way, more evidence that the problem is with your system rather than GnuCash.

PR #334, opened 2018-04-07
Typo fix and README update

Abschiedsstein:
Fixed typos in common/cmake_modules/GncAddSwigCommand.cmake
Updated README

### DIFF
--- a/README
+++ b/README
@@ -189,8 +189,7 @@ Building & Installing
 ---------------------
 
 GnuCash uses CMake to handle the build process. Details are available
-in cmake/README_CMAKE.txt (If you are building directly from Git, read
-the README.git file for more instructions.)
+in cmake/README_CMAKE.txt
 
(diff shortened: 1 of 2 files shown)


jralls:
The typo is OK, but why remove the reference to README.git?

Abschiedsstein:
The reference is in building and installing section. But there is no such information in README.git.
If the reference should be kept it should be in the context of how to behave when making changes to gnucash (which is the information addressed in README.git)

jralls:
OK, I see your point. I think that whole section needs overhauling: Most of the content either duplicates or belongs in README_CMAKE.txt.

gjanssens:
Thanks. I have merged this. It wasn't clear to me who is supposed to do the overhauling. @Abschiedsstein  if you would take a stab at this, feel free to create a new PR.

PR #336, opened 2018-04-09
Some Fixes after 3.0 released

Bob-IT:
I will add some fixes here as I try and fix them till this PR gets merged

### DIFF
--- a/gnucash/gnome-utils/dialog-transfer.c
+++ b/gnucash/gnome-utils/dialog-transfer.c
@@ -272,6 +272,16 @@ lookup_price(PriceReq *pr, PriceDate pd)
             prc = gnc_pricedb_lookup_latest (pr->pricedb, pr->from, pr->to);
             break;
     }
+
+    if (!prc) //no price found
+    {
+        PINFO("No price Found for %s, %s",
+              gnc_commodity_get_mnemonic(pr->from),
+              gnc_commodity_get_mnemonic(pr->to));
+        pr->price = NULL;
+        return FALSE;
+    }
+
     if (gnc_commodity_equiv(gnc_price_get_currency(prc), pr->from))
     {
         pr->reverse = TRUE;
@@ -279,7 +289,6 @@ lookup_price(PriceReq *pr, PriceDate pd)
               gnc_commodity_get_mnemonic(pr->to),
               gnc_numeric_to_double(gnc_price_get_value(prc)),
               gnc_commodity_get_mnemonic(pr->from));
-
     }
     else
     {
@@ -288,8 +297,6 @@ lookup_price(PriceReq *pr, PriceDate pd)
               gnc_numeric_to_double(gnc_price_get_value(prc)),
               gnc_commodity_get_mnemonic(pr->to));
     }
-    if (!prc)
-        return FALSE;
     pr->price = prc;
     return TRUE;
 }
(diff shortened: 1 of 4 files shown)


gnucash/gnome-utils/dialog-transfer.c
```diff
@@ -275,18 +275,35 @@ lookup_price(PriceReq *pr, PriceDate pd)
     if (gnc_commodity_equiv(gnc_price_get_currency(prc), pr->from))
```
gjanssens: This whole if clause is only to print a debug message. The added logic doesn't make sense. If prc doesn't exist no price is found. Period. The concept of reverse or normal doesn't apply in that case. Only when prc exists you can test for reverse or not. So the test sequence should be:
1. is prc NULL ? => info message no price found and return FALSE
2. is reverse ? => info message reporting the found reverse conversion rate
3. otherwise => info message reporting the normal conversion rate
In the end set pr->price and return TRUE

gnucash/gtkbuilder/dialog-preferences.glade
```diff
@@ -3280,6 +3280,7 @@ many months before the current month:</property>
                     <property name="visible">True</property>
                     <property name="can_focus">True</property>
                     <property name="tooltip_text" translatable="yes">To retrieve online quotes from Alphavantage, this key needs to be set. A key can be retrieved from the Alpha Vantage website.</property>
+                    <property name="width_chars">20</property>
```
gjanssens: On my system this is still not sufficient if the key happens to consist of only very wide glyphs (try 16x 'W'). Instead you could try to make the field expand (under common widget properties in glade).

gnucash/gnome-utils/dialog-transfer.c
Bob-IT: Brain dead, just needed to move prc not null test to before the info stuff.

gnucash/gtkbuilder/dialog-preferences.glade
Bob-IT: I assumed these were hexadecimal characters for the key.

Bob-IT:
Made the required changes hopefully and also added fix for bug794953

DiMan:
Hello Bob-IT and gjanssend,

I also looked at the problem with tabs a few weeks ago. I noticed that in the following part of the code two different functions are used to determine the length of the tab text: `g_utf8_strlen` and `strlen`. Can you please check if it is right.
```
        if (g_utf8_strlen (lab_text, -1) < tab_width)
        {
            gtk_label_set_width_chars (GTK_LABEL(label), strlen (lab_text));
            gtk_label_set_ellipsize (GTK_LABEL(label), PANGO_ELLIPSIZE_NONE);
        }
        ...
```

In addition, can you please check, if it makes sense to use the PANGO_ELLIPSIZE_END option instead of PANGO_ELLIPSIZE_MIDDLE. This will cut off the text at the end. 

For Example:
PANGO_ELLIPSIZE_MIDDLE: "Savings Interest" => "Savin...erest"
PANGO_ELLIPSIZE_END:       "Savings Interest" => "Savings In..."

Thanks,
DiMan

jralls:
Shouldn't it be getting the layout width from pango rather than counting codepoints (or worse, characters)?

DiMan:
@jralls: It would be also a good solution. Something like this?

```
if (tab_width != 0)
{
       gtk_label_set_width_chars (GTK_LABEL(label), tab_width);

       if (g_utf8_strlen (lab_text, -1) > tab_width)
       {
           gtk_label_set_ellipsize (GTK_LABEL(label), PANGO_ELLIPSIZE_MIDDLE);          
        }
       else
       {
          gtk_label_set_ellipsize (GTK_LABEL(label), PANGO_ELLIPSIZE_NONE);          
      }
}
else
{
      gtk_label_set_width_chars (GTK_LABEL(label), 15);
      gtk_label_set_ellipsize (GTK_LABEL(label), PANGO_ELLIPSIZE_NONE);
 }
```

PR #337, opened 2018-04-10
TR: catch invalid regex instead of report crashing.

christopherlam:
The main fix here is an exception catcher for invalid-regex for account/transaction matcher. There are also a few other IMHO harmless refinements. Please cherry-pick if this is preferred.


gjanssens:
Thanks!

PR #338, opened 2018-04-11
Bug 795064 - Enable subtotal-only if sortkeys are dates

christopherlam:
This will allow 'show subtotals only' if sortkeys are date-types. This allows, for example, prime-sortkey=date, yearly. sec-sortkey=date, monthly.

This can still produce invalid combinations eg both sortkeys = date and subtotals = none means this option is invalid. But I figure the user will need to make some effort to produce this combination.


jralls:
Thanks.

fellen:
Ahem, will it now appear in 3.1 or 3.900?
With a respective message we can close the bug, too.

PR #341, opened 2018-04-18
Fix syntax

fkoester:


fellen:
Right, but the orphaned trailing bracket should be removed

fkoester:
> Right, but the orphaned trailing bracket should be removed

Right, fixed.

gjanssens:
The review request was an accident. Please ignore...

gjanssens:
Merged, thanks!

PR #342, opened 2018-04-19
Add guile-json into borrowed

christopherlam:
This PR will copy guile-json into borrowed/guile-json and make preliminary modifications to borrowed/CMakeLists.txt

Unfortunately it's not installed into guile-load-path. How to fix this?

I've added test to /CMakeLists.txt to check if guile-json is callable from guile.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -343,16 +343,16 @@ if (GNC_SRFI64_RESULT EQUAL 0)
 endif ()
 
 # Test that guile has guile-json.
-execute_process (COMMAND ${GUILE_EXECUTABLE} -c "(use-modules (json builder) (json parser) (json syntax))"
+execute_process (COMMAND ${GUILE_EXECUTABLE} -c "(load-from-path \"json.scm\") (use-modules (json builder))"
   RESULT_VARIABLE GNC_GUILE_JSON
-  ERROR_QUIET
+#  ERROR_QUIET
 )
 
 if (GNC_GUILE_JSON EQUAL 0)
   message (STATUS "guile-json is present.")
   set (HAVE_GUILE_JSON TRUE)
 else ()
-  message (FATAL_ERROR "guile-json cannot be imported. Abort.")
+  message (STATUS "guile-json cannot be imported. Abort.")
 endif ()
 
 # ############################################################
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -342,6 +342,19 @@ if (GNC_SRFI64_RESULT EQUAL 0)
   set (HAVE_SRFI64 TRUE)
 endif ()
 
+# Test that guile has guile-json.
+execute_process (COMMAND ${GUILE_EXECUTABLE} -c "(use-modules (json builder) (json parser) (json syntax))"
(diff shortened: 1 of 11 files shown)


jralls:
This thing is GPLv3, so incorporating it into GnuCash will require GnuCash to also re-license as GPLv3.

Does it really make sense to add this code directly to the project instead of either as an external dependency or a git subproject?

Plus, I'm still concerned that using a Scheme instead of a C/C++ implementation puts this at the wrong level.

Meanwhile, why did you push two commits where the second cancels out the first? Please reset to 896ecf6 and force-push.

christopherlam:
Sorry mistake. README looked like .md but is not really. I will reset and force-push when at home.

This module is very handy for converting scheme data to javascript for jqplot/chartjs. While it is convenient, I agree relicensing everything to GPL3 is not realistic. I'll think of a solution. I'm guessing having another dependency is not ideal either.

jralls:
I just had a look at Guile and it's also GPLv3. We have to change anyway.

gjanssens:
Guile is LGPL according to its version string? If that's true we don't have to change IIUC

gjanssens:
[LICENSE file for guile 2.2.3](http://git.savannah.gnu.org/cgit/guile.git/tree/LICENSE?h=v2.2.3)

jralls:
COPYING from the same: http://git.savannah.gnu.org/cgit/guile.git/tree/COPYING?h=v2.2.3
But they have COPYING.LESSER there as well, and reference both in the LICENSE file. Ah, and so does guile-json. A check of the top of guile-json.scm shows that it, too, is LGPLv3. So no problem after all.

christopherlam:
Thank you, disaster averted! I'll let you decide whether it is copied into borrowed, or added as a dependency. I put into borrowed because that's all I knew!

gjanssens:
I've merged this after a rebase and fixing the issues.

These were the issues:
1. The cmake presence test for guile-json was misplaced. Such test should only be written for external dependencies. However we have included guile-json in our source tree. In addition you can't test for files the will eventually get installed at cmake time. So I have remove this test.
2. gnc_add_scheme_targets can't handle sources in subdirectories. To handle this, I have split the CMakeLists.txt file such that the json directory gets its own.
3. I have added the copying and authors files also in the dist tarball.

Note I have chosen to merge this into maint as nothing depends on it yet. I don't expect the chartjs to be included in maint as well. That would be too intrusive for a stable release. However by adding it in maint I hope to reduce the number of merge conflicts you may get due to me lowercasing cmake commands right after your PR.

christopherlam:
@gjanssens thanks!
The CMake hack was meant as an easy test to check if I've done guile-json/CMakeLists.txt correctly. No wonder I could never do!
I'm still struggling to do `(use-modules (json builder))` - any clue how to make its exported functions available? Does app-utils.scm need amending?

gjanssens:
No that was my bad. I installed json under the gnucash namespace while I shouldn't have. Under that configuration it would have required ```(use-module (gnucash json builder))``` and that would also require you to alter json.scm to make it work. I have now pushed a followup fix to make it work as you would expect.

PR #343, opened 2018-04-21
More Fixes for 3.0

Bob-IT:
Three bug fixes for version 3.0 tried on my Linux box and Windows 10
Bug 794990 fixes the double clicking on saved reports followed by a fix for the tooltips and a code clean in that file.
Bug 795132 fixes tabbing, I think it is correct but can not try on a mac
Bug794976 replaces the entry context menu with the register so it is the same as 2.6 version.

Bob


Bob-IT:
I will give some thought to this after things have settled down. I was in two minds to possibly add items to the entry context menu which would involve grabbing possible menu items from the register or use the existing which seemed simpler and quicker to achieved the same as 2.6.x

PR #344, opened 2018-04-25
Maint string html escape

christopherlam:
Creating a html string sanitization function TDD-style.

This is different from jqplot-escape-string which will escape double-quotes as well.

I think I'd do it using reliable string->chars->string loop constructs, or using ports, rather than C-style regex calls. In practice, the string-sanitize algorithm will only be run for report labels, descriptions, account names, and can be done in scheme; it does not need C-level speed.

OSS development 😀


gjanssens:
I was investigating this yesterday evening, but I didn't have time to analyse everything.
I mainly focussed on chart legend labels, like in the Expense bar chart. Here are my findings so far.

The current jqplot-escape-string function (or more precisely regex-substitute/global) mangles unicode strings so it should be replaced with something else.
jqplot-escape-string is used on strings that will be inserted in javascript template code. That is guile will concatenate a number of fixed strings with the outputs from jqplot-escape-string to create valid javascript code. Guile itself of course is completely unaware of the semantics of javascript, it just sees a bunch of strings. A simplified example:
```
(push "series.push({ label: \"")
                         (push (jqplot-escape-string label))
                         (push "\"});\n\n")
```
This will result in following output:
```
{ label: "<value-from-jqplot-escape-string>"};
```
Now let's make label being passed to jqplot-escape-string 'Books from "Maggy" bookstore', an imaginary but valid account name. That would make the above javascript snippet:
```
{ label: "Books from "Maggy" bookstore"};
```
This is no longer valid javascript code because there is now a string with unescaped double quotes in it.
Which gets me to my point: you skipped escaping double quotes in your sanitizing function while I believe it's the single most important escape of all. The second one which would really be there is the backslash.

As for all the other escapes (html oriented), I don't know whether that is even needed. I tried disabling those for test and the barchart still worked absolutely fine even when the account name has a < or & in it. jqplot-escape-string's function is to escape strings before they go into javascript code, not into html code. And as your own tests also suggested javascript seems to handle this escaping just fine by itself.

To be complete I wrote at the start I haven't had time to analyse all uses of jqplot-escape-string. If it's also used in contexts where the escaped string is inserted directly in html context rather than in javascript context, it would make sense to also provide html escapes. However I would suggest to make that a separate escape function. Html and javascript have different escape needs IMO.

Having said all that I like your TDD approach. Your tests however will need some refinement. The emoticon test is a good start, but the test string is supposed to be unchanged. I don't know whether regex-substitute/global would hav mangled it in that case. You could add a 'Smile:"🙂" something' test, in which the double quotes should get escaped correctly and the smiley unchanged.

christopherlam:
Ok to summarize, a basic string-sanitize will not need to escape " quotes, depending on its use.

1) the current mechanism is: `(push "<var chartid=\"") (push (jqplot-escape-string chartid)) (push "\">\n")` - notice the `\"` surrounding the jqplot-escape-string. The regex-free string-sanitize-js can be used here.
2) my preferred mechanism is `(push (format #f "<var chartid=~s>" chartid))` -- notice the absence of `\"`; the chartid formatted as ~s will automatically insert the double-quotes.

I'll need to review my chartJS PR to sanitize any strings that will potentially be user-selectable.

So far I believe any jqplot-escape-string can use the string-sanitize-html or -js as above, without using regex.

gjanssens:
Hmm, seeing this again I tend to agree your original method is more elegant. It looks like the last guile maintainers before you (including me) were not really aware of guile's format function.
So one sanitize function should be ok, combined with rewriting the various push commands to use
`(format #f "some string with ~s" (string-sanitize value))`

Will you take on the job of replacing all current jqplot-escape-string calls ?

And then a nitpick: I noticed you have added the function in report-utilities.scm. There is also html-utilities.scm and html-jqplot.scm. The latter holds the current jqplot-escape-string and two small functions.

As sanitizing is really html oriented I'd add the function in html-utilities, reusing the naming conventions of that file (like gnc:html-sanitize-string).

And while we're in the area, you can also move the two remaining functions from html-jqplot.scm to html-utilities.scm and completely drop html-jqplot.scm. Of course only after all calls to jqplot-escape-string are replaced :)

christopherlam:
Sure I'll be a bit busy with real work over next few days but will do the necessary, sanitizing where I feel it'll be necessary.

jqplot-escape-string (and hopefully *all* guile->C-style regex, and encoding woes) will be banished.

christopherlam:
(partial work done, will ping when complete)

gnucash/report/report-system/test/CMakeLists.txt
```diff
@@ -29,6 +33,10 @@ set(GUILE_DEPENDS
   )
 GNC_ADD_SCHEME_TESTS(${scm_test_report_system_SOURCES})
 
+if (HAVE_SRFI64)
+  gnc_add_scheme_tests ("${scm_test_report_system_with_srfi64_SOURCES}")
+endif (HAVE_SRFI64)
```
gjanssens: Are there environments we support that don't have SRFI64 ?

christopherlam: erm the travis-ci ubuntu one doesn't have SRFI64. i've copied from test-transaction.

gnucash/report/standard-reports/net-barchart.scm
```diff
@@ -39,7 +39,6 @@
 
 ;; included since Bug726449
 (use-modules (ice-9 regex)) ;; for regexp-substitute/global, used by jpqplot
```
gjanssens: Is regex still used in other parts of this file ?

christopherlam: I don't think so! I think I'll remove these as a last commit

gnucash/report/report-system/test/CMakeLists.txt
gjanssens: Ah indeed. Ubuntu 14.04 ships guile 2.0.9 and srfi-64 was introduced in 2.0.10. So there we go, we need the conditional...

PR #345, opened 2018-05-01
Fix the incorrect default sort for the reconcile window

Bob-IT:
This should fix the default sort for the reconcile window and have enabled the sort for the reconcile column. Also changed the setting of the right margin for the totals from the gtk box to the label.


gnucash/register/register-gnome/gnucash-item-edit.c
```diff
@@ -814,6 +814,14 @@ button_press_cb (GtkWidget *widget, GdkEventButton *event, gpointer *pointer)
...
         return TRUE;
     }
+
+    // Middle mouse button, paste
+    if (event->button == 2 && event->type == GDK_BUTTON_PRESS)
+    {
+        gnc_item_edit_paste_clipboard (GNC_ITEM_EDIT(sheet->item_editor));
+        return TRUE;
+    }
+
```
gjanssens: Middle mouse button paste is a default feature of a GtkEntry widget. You don't have to explicitly do this. It's one of the examples of where we were reimplementing standard Gtk behaviour.

gnucash/register/register-gnome/gnucash-sheet.c
```diff
@@ -1465,9 +1465,7 @@ gnucash_sheet_button_press_event (GtkWidget *widget, GdkEventButton *event)
         button_1 = TRUE;
         break;
     case 2:
-        if (event->type != GDK_BUTTON_PRESS)
-            return FALSE;
-        gnc_item_edit_paste_clipboard (GNC_ITEM_EDIT(sheet->item_editor));
+        // This middle button option has been moved to gnucash-item-edit.c
```
gjanssens: It may be a bit more complicated than this.
The intent of a middle mouse button click is to paste the current clipboard at the current location.
By no longer handling a middle mouse button click here this works when clicking in the item_edit (we get this for free from Gtk).

However a middle-click somewhere else on the sheet now does nothing anymore. I think it should
- move the cursor to where the user clicked (like a left-click)
- pass the middle-click on to the item-edit to handle the past action

Note that if you get to redo the itemedit popup as I suggested in a previous PR, this function really reduces to moving the cursor and passing any event on to the itemedit in its new location...

gnucash/gnome/window-reconcile.c
```diff
@@ -1200,6 +1194,12 @@ gnc_reconcile_window_create_view_box(Account *account,
     gtk_box_pack_start(GTK_BOX(hbox), label, FALSE, FALSE, 0);
     *total_save = label;
 
+#if GTK_CHECK_VERSION(3,12,0)
+    gtk_widget_set_margin_end (GTK_WIDGET(label), 10);
+#else
+    gtk_widget_set_margin_right (GTK_WIDGET(label), 10);
+#endif
+
```
gjanssens: I only see now this is hardcoded to 10px. How does this work out with themes that use different margins ? Or are all the relevant margins hardcoded by us in this area ?

Bob-IT: We are specifying them in gnc_reconcile_window_create_view_box of window-reconcile.c

gnucash/register/register-gnome/gnucash-item-edit.c
Bob-IT: In the current setup we do, I think this is as a result of setting the entry so it can not be focused.

gnucash/register/register-gnome/gnucash-sheet.c
Bob-IT: This would be nice but it does not happen in 2.6 version. What I was aiming at getting was what is missing working and then do the focus change on master as I do not think it will be that straightforward for me.

gjanssens: Hmm, not sure I can follow. 2.6 is not a reference as it was still using gnomecanvas. I already made plenty of changes for 3.0 in this area. Pasting already works in the itemedit. What's not working is moving the cursor on a middle click. IMO that can probably solved by handling left-click and middle-click exactly the same here in the sheet. You're now returning prematurely for a middle-click.

When I said it may be a bit more complicated I was thinking of how to process the event such that it gets passed to the itemedit after the cursor is updated. But we can wait with that part if you prefer.

gnucash/gnome/window-reconcile.c
gjanssens: Ok, then they remain in sync.

Bob-IT:
I have changed the fix for the middle mouse button and it now moves the cursor and pastes in one go.

gjanssens:
Note I have already pushed the other two commits. You can rebase this PR if you like to clean that part up and focus the remainder on the middle-click issue.

Bob-IT:
Did not know there was a difference when pasting but trying in a text editor I can see what you mean. Before I have another go at this, can you advise if middle clicking on a different transaction should replace all text in that field, as it currently does or should it insert selected/copied text where the mouse pointer is ?

gjanssens:
I would expect it to paste the copied text where the mouse pointer is.

Bob-IT:
Here is another go, I think it does what it is supposed to do but I am sure you will advise otherwise.
;-)

christopherlam:
something weird happening with test-income-gst.scm - please feel free to disable/ignore this test while i investigate

christopherlam:
erm I've just pushed the exact same branch which passes on travis. looks like something is temperamental in test-income-gst. https://travis-ci.org/christopherlam/gnucash/jobs/378369544 - a travis restart should fix it.

PR #346, opened 2018-05-03
Some scheme progress

christopherlam:
### A few harmless refinements
1. `(define-macro)` is apparently frowned upon nowadays; use `(define-syntax)` instead.
2. TR sortkey capabilities were listed manually; change to dynamically check according to their attributes.
3. TR: enable subtotal/grouping of Desc/Notes/Memo.
4. test-transaction.scm modified to change "out-NN.html" output in /tmp to "test-trep-TESTNAME.html"
5. TR/GSTR: begin string sanitizing project

### User-visible changes TR

1. Description/Notes/Memo can be grouped and subtotalled. The original TR did not allow grouping for these sortkeys, and there is no reason we should disallow grouping on these freetext fields which can allow arbitrary personal coding.
2. Common-currency amount/debit/credit changed from 2-line `Amount + commodity` to 1-line `Amount (commodity)` - this is due to the table-col-headers cannot receive gnc:make-html-text objects.
3. If subtotal-only was selected, the left-columns (date, desc etc) were still rendered. Cosmetic bugfix.
4. Add Filter/Closing transactions, and enable it by default? I think it's a good thing.

### DIFF
--- a/gnucash/report/business-reports/customer-summary.scm
+++ b/gnucash/report/business-reports/customer-summary.scm
@@ -97,8 +97,10 @@
 
 ;; ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 
-(define-macro (addto! alist element)
-  `(set! ,alist (cons ,element ,alist)))
+(define-syntax addto!
+  (syntax-rules ()
+    ((addto! alist element)
+     (set! alist (cons element alist)))))
 
 (define (set-last-row-style! table tag . rest)
   (let ((arg-list 
--- a/gnucash/report/business-reports/customer-summary.scm
+++ b/gnucash/report/business-reports/customer-summary.scm
@@ -97,13 +97,6 @@
 
 ;; ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 
-(define (set-last-row-style! table tag . rest)
-  (let ((arg-list 
-         (cons table 
-               (cons (- (gnc:html-table-num-rows table) 1)
-                     (cons tag rest)))))
-    (apply gnc:html-table-set-row-style! arg-list)))
-
 (define (date-col columns-used)
   (vector-ref columns-used 0))
 (define (num-col columns-used)
@@ -591,7 +584,7 @@
     (gnc:html-table-append-row!
      table
      (list "<br>"))
-    (set-last-row-style!
+    (gnc:html-table-set-last-row-style!
      table "td"
      'attribute (list "valign" "top"))
     table))
@@ -609,7 +602,7 @@
      table "table"
      'attribute (list "border" 0)
      'attribute (list "cellpadding" 0))
-    (set-last-row-style!
+    (gnc:html-table-set-last-row-style!
      table "td"
      'attribute (list "valign" "top"))
     table))
--- a/gnucash/report/business-reports/customer-summary.scm
+++ b/gnucash/report/business-reports/customer-summary.scm
@@ -97,11 +97,6 @@
 
 ;; ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 
-(define-syntax addto!
-  (syntax-rules ()
-    ((addto! alist element)
-     (set! alist (cons element alist)))))
-
 (define (set-last-row-style! table tag . rest)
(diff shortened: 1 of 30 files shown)


gnucash/report/standard-reports/transaction.scm
```diff
@@ -400,6 +398,16 @@ Credit Card, and Income accounts."))
   ;; it checks whether a renderer-fn is defined.
   (keylist-get-info sortkey-list sortkey 'renderer-fn))
 
+(define (CUSTOM-SORTING? sortkey)
+  ;; sortkey -> bool
+  ;;
+  ;; this returns which sortkeys which *must* use the custom sorter.
+  ;; it filters whereby a split-sortvalue is defined (i.e. the splits
+  ;; can be compared according to their sortvalue) but the QofQuery
+  ;; sortkey is not defined (i.e. 'split-sortvalue is #f).
```
christopherlam: oops typo should be "(i.e. 'sortkey is #f)

gnucash/report/business-reports/customer-summary.scm
```diff
@@ -97,8 +97,10 @@
 
 ;; ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 
-(define-macro (addto! alist element)
-  `(set! ,alist (cons ,element ,alist)))
+(define-syntax addto!
+  (syntax-rules ()
+    ((addto! alist element)
+     (set! alist (cons element alist)))))
```
gjanssens: Considering you're altering the same macro in 8 files, perhaps it should live somewhere in a common file ?

gnucash/report/report-system/html-utilities.scm
```diff
@@ -818,6 +818,65 @@
        "")
       (_ "Edit report options")))))
 
+(define* (gnc:render-options-changed options #:optional plaintext?)
```
gjanssens: Good choice for the new location.
I would suggest updating the name to start with "gnc:html" to be more consistent with the surrounding functions. That makes it slightly easier to trace back the code.

gjanssens:
On a more general note - while testing your changes I found two options related to account summaries:
1. Display->Subtotal table
1. Show subtotals only.

If I enable both I actually get two subtotals tables. One in normal font and one in boldface with "Total for" added to each account name. There seems to be no (obvious) way to only show the first and not the second ?

gnucash/report/business-reports/customer-summary.scm
christopherlam: Hmm, this seems a fundamental type function. Perhaps in utilities.scm?

christopherlam:
>     Display->Subtotal table
>     Sorting->Show subtotals only.
> 
> If I enable both I actually get two subtotals tables. One in normal font and one in boldface with "Total for" added to each account name. There seems to be no (obvious) way to only show the first and not the second ?

I think there are always more edge cases. There are 8 combinations from 3 booleans (above 2 and Display/Totals), and 4 types of data to print (i.e. transactions, subtotals, grandtotal, grid). I think that maybe I should have named it "Display->Hide transactions"  instead of "Sorting->Subtotals only (hide txns)" but I created this option mainly for 'daily subtotals.

Perhaps we need to define exactly what we want. Display/Subtotal-Table can be extended into a multichoice - i.e. #f, #t, and 'grid-only and this will satisfy DavidT's wish for a single "Grand-Total" type report. This will create 12 combinations then!

gnucash/report/business-reports/customer-summary.scm
christopherlam: I've just tried to insert into utilities.scm and I don't think it's easy - https://www.gnu.org/software/guile/manual/html_node/R6RS-Incompatibilities.html - I keep getting an error whichever way I try to export a macro from utilities.scm... best leave as it is

christopherlam:
> Display->Subtotal table
>     Sorting->Show subtotals only.
> 
> If I enable both I actually get two subtotals tables. One in normal font and one in boldface with "Total for" added to each account name. There seems to be no (obvious) way to only show the first and not the second ?

Or, the report can detect if both Display/Subtotal-Table and Sorting/Show-subtotals-only are enabled, it decides to hide main-table and display grid only, as a special case :)

christopherlam:
Added Closing Status filter... I think is a good thing to enable by default.

christopherlam:
The last few commits relate to cleaning up SRFI64 utility functions to make easier tests, and to begin creating tests for business-reports. I'd like to merge various invoices.scm and need to create tests first.

If preferred I can split TR work and Testing commits into different PRs.

PR #347, opened 2018-05-09
Scheme progress

christopherlam:
Ok putting this together again from #346 -- ~Travis failure is not of my doing!~ I believe most of these commits are self-contained. But some have dependencies (eg centralizing sortkeys -> allows easier addition of sortkeys)

## User visible changes
1. TR can now subtotal/group by Desc/Notes/Memo. This can be useful and allows users to use arbitrary strings to code their transactions.
2. TR if hiding transactions, hide headers as well (date, desc etc)
3. TR Add Closing Filter, and enable by default (comments?)

## Under the hood
1. TR - centralize custom-sorter values, dynamically check sortkey capabilities.
2. TR/GSTR - begin string sanitizing.
3. Centralize general functions/macros - addto!, gnc:html-table-set-last-row-style!
4. Centralize test utility functions - test-runner, gnc:options->sxml
5. Purge dead code in test-extras.scm, eliminate list-extras.scm
6. Initialize business-reports/test for creating tests
7. Restore report-system/test/CMakeLists.txt which erroneously disabled tests
8. Simplify/schemify numerous functions.
9. Introduce GST Report test.
10. After all this prep work, I'll work on business-report tests.


christopherlam:
I'd think this is ready. I'll be hacking on business-tests on https://github.com/christopherlam/gnucash/commits/scheme-progress-business-tests now; and will be basing on this branch.

gnucash/report/standard-reports/transaction.scm
```diff
@@ -555,6 +574,16 @@ tags within description, notes or memo. ")
...
+  (gnc:register-trep-option
+   (gnc:make-multichoice-option
+    pagename-filter optname-closing-transactions
+    "l" (_ "By default most users should not include closing \
+transactions in a transaction report. Closing transactions are \
+transfers from INCOME and EXPENSE accounts to equity, and must usually \
+be excluded from periodic reporting.")
+    'exclude-closing
+    (keylist->vectorlist show-closing-list)))
+
```
gjanssens: This particular commit adds plenty of new translatable strings. Our translators won't be happy if we do that on the maint branch. There was a previous commit that introduced one new string which I would let pass.  I'm counting 8 here.

gnucash/report/standard-reports/income-gst-statement.scm
```diff
@@ -35,24 +35,27 @@
...
+ which will require Tax Tables to be set up correctly. Please see the documentation.")
+   (gnc:html-markup-br)
+   (gnc:html-markup-br)
    (_ "From the Report Options, you will need to select the accounts which will \
 hold the GST/VAT taxes collected or paid. These accounts must contain splits which document the \
 monies which are wholly sent or claimed from tax authorities during periodic GST/VAT returns. These \
 accounts must be of type ASSET for taxes paid on expenses, and type LIABILITY for taxes collected on sales.")
-   "<br/><br/>"))
+   (gnc:html-markup-br)
+   (gnc:html-markup-br)))
```
gjanssens: Three more changed translatable strings...

gjanssens:
Other than the translatable string changes this looks very good.
@jralls, @fellen I don't know very well what to decide on the translatable strings. I thought we tried to keep changes in this area to an absolute minimum. What do you propose in this case ?

jralls:
While I imagine that translators are OK with changes that have an obvious purpose--e.g. something has changed in the code, or rewriting a message for better clarity--these seem mostly trivial and gratuitous: Pedantic changes to markup, replacing "&" with "and", and similar. The sole exception was adding "Filter" to the menu options and that's already an available string so there's no problem with it anyway.

On the other hand since the changes are generally trivial I expect that msgmerge will just mark them fuzzy and move on.

gjanssens:
> While I imagine that translators are OK with changes that have an obvious purpose--e.g. something has changed in the code, or rewriting a message for better clarity--these seem mostly trivial and gratuitous: Pedantic changes to markup, replacing "&" with "and", and similar.

These follow from the effort to sanitize all html strings. The "&" is a problem in pure html hence the replacement.

> The sole exception was adding "Filter" to the menu options and that's already an available string so there's no problem with it anyway.

Well not quite. That commit starts with adding "Filter" and continues with 7 more new strings in transaction.scm to add the new "Closing transactions" filter option. I was more concerned about the new strings.

So for you it would be fine if we merge this into maint as is ? I don't find any formal policy about this on the wiki and I don't remember where I got this idea that new translations on maint should be avoided.

jralls:
Sigh, bushwacked again by github's habit of hiding "large" diffs so that find-in-page didn't find them.

Those additional strings are just the sort that I think translators don't have a problem with. They're clearly a necessary part of the new feature.

Take a look at the msgid counts for the 2.6 series at the [Translation Project](http://translationproject.org/domain/gnucash.html). 2.6.0 had 4592 and it grew more or less steadily to 4685 by 2.6.19. If we're serious about "slipstreaming" more improvements into 3.x we'll need to accept a higher level still.

christopherlam:
1) Sure, new strings are unavoidable for Closing Txn filter, and can be reverted if they are a problem. But my feeling is we should disable them by default for a useful TR. This is perhaps the only substantial change in this series of commits and may be unexpected for long-term users who already accommodate for these closing txns.

2) While discussing Closing Txns, there's also the weird Closing Entries specifier in the "Income Statement" report, and the Adjusting/Closing Entries specifier in the "Trial Balance" --> but perhaps these would best be discussed in -user.

gjanssens:
So it's settled. New strings are added in maint. Thanks!

christopherlam:
Thankfully, the merge has added the commits in the tidier intended order, rather than github's confusing order!

PR #349, opened 2018-05-14
Bug 796081 Tax Schedule Report - An error occurred while running the report

christopherlam:
Missed out from timepair->time64 conversion.


jralls:
Excellent, thanks!

PR #351, opened 2018-05-17
Modify date formats to strip out certian modifiers.

wenottingham:
strptime/strftime supports various modifiers to their parameters.
- 'E' and 'O': alternate locale-specific formats
  (used in Persian, Oriya, Azerbaijani)
- '-': padding
  (used in Czech)

GnuCash passes dates as integer y/m/d without using locale-specific
formats, so we need to strip out 'E' and 'O' from the format. '-'
goes wrong in boost somewhere, as far as I can tell, so strip it out
as well.

Fixes https://bugzilla.gnome.org/show_bug.cgi?id=795247.

### DIFF
--- a/libgnucash/engine/gnc-date.cpp
+++ b/libgnucash/engine/gnc-date.cpp
@@ -728,6 +728,32 @@ floordiv(int a, int b)
     }
 }
 
+/* Normalize the localized date format to avoid date scanning issues.
+ *
+ *   The 'O' and 'E' format modifiers are for localized input/output
+ *   characters. Remove them as we are always using Arabic numbers.
+ */
+static char *
+normalize_format (const char *format)
+{
+    gint counter = 0, n_counter = 0;
+    gchar *normalized;
+
+    normalized = g_strdup(format);
+    while (format[counter] != '\0')
+    {
+        normalized[n_counter] = format[counter];
+        if ((format[counter] == '%') && \
+            (format[counter+1] == 'E' || format[counter+1] == 'O'))
+            counter++;  // skip format modifier
+
+        counter++;
+        n_counter++;
+    }
+    normalized[n_counter] = '\0';
+    return normalized;
+}
+
 /* Convert a string into  day, month and year integers
 
     Convert a string into  day / month / year integers according to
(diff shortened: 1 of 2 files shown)


libgnucash/engine/gnc-date.cpp
```diff
@@ -585,6 +607,8 @@ void qof_date_completion_set(QofDateCompletion dc, int backmonths)
 */
 const gchar *qof_date_format_get_string(QofDateFormat df)
```
jralls: This affects date string output as well as parsing. That breaks localized dates on checks, so one gets e.g. 18/05/17 instead of ۸۱/۵۰/۷۱ (which is actually backwards because this editor doesn't support RTL). It probably also breaks fancy dates on invoices, but I can't read Farsi well enough to get that set up. Rather than munging qof_date_format_get_string and qof_date_text_format_get_string you should apply normalize_dt_format only for passing to strptime and to the one strftime call in this file where the string is used to determine the date separator.

libgnucash/engine/gnc-date.cpp
```diff
@@ -126,6 +126,30 @@ gnc_localtime_r (const time64 *secs, struct tm* time)
     }
 }
 
+char *
```
jralls: static char*, see below.

libgnucash/engine/gnc-date.h
```diff
@@ -221,6 +221,13 @@ gchar* gnc_ctime (const time64 *secs);
  */
 time64 gnc_time (time64 *tbuf);
 
+/** \brief Normalize a strftime/strptime format string
+ *  \param d_fmt: A strftime/strptime format string
+ *  \return A format string without modifiers. This string is allocated
+ * and must be freed by the caller
+ */
+char *normalize_d_fmt (const char *d_fmt);
```
jralls: If we really wanted to add this to the published API in would have to be gnc_date_normalize_d_fmt(const char *d_fmt). Fortunately it can stay a static because you'll need to copy it in GncDateTimeImpl::format(), see below.

wenottingham:
OK.

FWIW, sent upstream to boost at https://github.com/boostorg/date_time/issues/83 (and downstream in Fedora at https://bugzilla.redhat.com/show_bug.cgi?id=1580216)

wenottingham:
v3:
Outside of review comments:
-  updated source code comments and some var names
- don't strip '-' for gnc-date, it's not needed due to strptime() fixes in glibc

jralls:
Thanks!

PR #353, opened 2018-05-24
Fix permanent storage of vendor details.

p-wassi:
Since the data types did not match for the billterms and taxtable,
those references/guids were not saved to the database.
Also, while at it, fix another copy'n'paste error in sql backend.


jralls:
Christian Stimming's fantasies not withstanding the 2.6 branch is closed.

PR #354, opened 2018-05-24
Fix permanent storage of vendor details.

p-wassi:
Since the data types did not match for the billterms and taxtable,
those references/guids were not saved to the database.

Similar to the PR before, but for the maint branch here.


jralls:
Amazing that that has gone unnoticed for 6 years. Thanks!

PR #355, opened 2018-05-25
TR: upgrade grid to display averages and multiple commodities

christopherlam:
## Visible Changes
1. Grid now supports multiple commodities. Each commodity gets a separate row, exactly like in the main table.
2. Grid is disabled if primary-subtotal is disabled. I'm think it could also be disabled if secondary-subtotal is disabled too?
3. Grid now calculates and displays row averages. If prime-sortkey = accounts, and sec-sortkey = dates with subtotals, it provides a useful average periodic income/expenditure. This is a small feature that's compatible with multiple commodities above.

## Invisible changes
4. Refactoring add-subtotal-row - the previous (addto!) use is not functional programming. Also deduplicate code - first-commodity and subsequent-commodities don't need to be handled separately.


gjanssens:
Thanks!

PR #356, opened 2018-05-26
Fix build on case-sensitive filesystems

ryandesign:
Fixes build on case-sensitive filesystems. See https://trac.macports.org/ticket/56536


jralls:
Thanks!

PR #357, opened 2018-05-27
invoice.scm upgrades; generates 3 invoice variants

christopherlam:
Work so far. ~A few housekeeping commits, and the test-invoice.scm which tests the 3 invoice reports variants. This test provides full coverage for invoice functionality, except gncOrder rendering and the 'no-invoice-specified' blank page.~

* invoice.scm cleanup
* initialize custom css in invoice.scm which will be supersized to generate other invoice variants.
* ~the stylesheets have been modified to pass the *report* style-text to the *style-sheet* style-text. The combined CSS will have the default old css, and add any custom css as well. See invoice.scm.~

## ~Housekeeping:~
- ~easy-invoice was handwriting html using attributes `cellpadding=0` which cannot be parsed, need to enclose value with quotes: `cellpadding='0'`~
- ~options->sxml can now parse html containing `&nbsp;`~
- ~options->sxml now ignores whitespace in html. Travis was complaining randomly.~
- ~test-cashflow-barchart was writing whole html to stdout. Disable.~

~Note commit 'last test inv-8 is paid up' adds a comment that calling (gncEntryGetDocValue) can trigger a subtle bug: it seems to recalculate the entry amount and introduce rounding errors. As a result, calling it results in the last test (which calls gncInvoiceIsPaid) failing.~

## Next steps
During refactoring and merging of invoices, I expect the tests to be modified, so they won't need to handle variants separately.


christopherlam:
This branch is based on #364 and has now merged all invoice variants.

jralls:
Thanks!

PR #359, opened 2018-05-31
Couple of bug fixes

Bob-IT:
Bug 795471 hopefully fixes the budget issue. There are still the existing gtk warnings about size allocation which I can get rid of if I move the gbv_treeview_resized_cb to the totals_tree_view but then I can not find a way to reliably keep the tree views in sync. The other budget changes seem appropriate but can be removed if not.
Bug 796083 could be fixed by dropping the model from the tree view and reconnecting but that raised other issues so opted to queue a redraw of the tree view instead.


jralls:
Thanks!

PR #360, opened 2018-06-03
Bug787401 test report definition

GnuFiBux:
Hi,
please note that this my first pull request, it may be not the most mature pull request you have ever come across.

This pull request has three aspects:
- it fixes a bug in the current code: if the report definition is triggered for a report without a GUID, then GnuCash terminates with an error message from Guile. With this fix, an error dialog message is popping up to notify the user, and after acknowledgment the start-up of GnuCash resumes as normal.
- it adds test cases for the report definition in report.scm
- it adds a flag to report.scm that can be set to avoid calls to GUI dialogs that are not working for the tests, so by setting the tests before test-begin (and resetting after test-end) these calls can be suppressed.
Looging forward to your comments.

### DIFF
--- a/gnucash/report/report-system/test/test-report-system.scm
+++ b/gnucash/report/report-system/test/test-report-system.scm
@@ -1,0 +1,68 @@
+(use-modules (gnucash gnc-module))
+
+(gnc:module-begin-syntax (gnc:module-load "gnucash/app-utils" 0))
+
+(use-modules (gnucash engine test test-extras))
+(use-modules (gnucash report report-system))
+(use-modules (srfi srfi-64))
+(use-modules (gnucash engine test srfi64-extras))
+
+(define (run-test)
+    (set! test-report-system-flag #t)
+    (test-runner-factory gnc:test-runner)
+    (test-begin "Testing/Temporary/test-report-system") ;; if (test-runner-factory gnc:test-runner) is commented out, this
+                                                            ;; will create Testing/Temporary/test-asset-performance.log
+    (test-assert "Minimum Report Definition" (test-check1))
+    (test-assert "Missing GUID detection" (test-check2))
+    (test-assert "Detect double GUID" (test-check3))
+    (test-assert "Report with Full Argument Set" (test-check4))
+    (test-end "Testing/Temporary/test-report-system")
+    (set! test-report-system-flag #f)
+)
+
+;; -----------------------------------------------------------------------
+
+(define (test-check1)
+  (gnc:define-report 'version "1" 'name "Test Report Template" 'report-guid "54c2fc051af64a08ba2334c2e9179e23")
+)
+
+;; -----------------------------------------------------------------------
+
+(define (test-check2)
+  (if (not (gnc:define-report 'version "1" 'name "Test Report Template"))
+    #t
+    #f
+  )
+)
+
+;; -----------------------------------------------------------------------
+
+(define (test-check3)
+  (if (not (gnc:define-report 'version "1" 'name "Test Report Template" 'report-guid "54c2fc051af64a08ba2334c2e9179e23" 'parent-type "Parent Type" 'options-generator "Options Generator" 'renderer "Renderer" 'options-cleanup-cb "Options Clean-Up" 'options-changed-cb "Options Changed" 'in-menu? #f 'menu-path "Menu Path" 'menu-name "Menu Name" 'menu-tip "Menu Tip" 'export-types "Export Types" 'export-thunk "Export Thunk"))
+    #t
+    #f
+  )
+)
+
+;; -----------------------------------------------------------------------
+
+(define (test-check4)
+  (and
+    (gnc:define-report 'version "1" 'name "Test Report Template" 'report-guid "54c2fc051af64a08ba2334c2e9179e24" 'parent-type "Parent Type" 'options-generator "Options Generator" 'renderer "Renderer" 'options-cleanup-cb "Options Clean-Up" 'options-changed-cb "Options Changed" 'in-menu? #f 'menu-path "Menu Path" 'menu-name "Menu Name" 'menu-tip "Menu Tip" 'export-types "Export Types" 'export-thunk "Export Thunk")
+    (string=? (gnc:report-template-version (gnc:find-report-template "54c2fc051af64a08ba2334c2e9179e24")) "1")
+    (string=? (gnc:report-template-name (gnc:find-report-template "54c2fc051af64a08ba2334c2e9179e24")) "Test Report Template")
+    (string=? (gnc:report-template-report-guid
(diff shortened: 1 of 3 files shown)


jralls:
This needs some fixing in order to merge cleanly. In the following instructions I've used ```origin```for https://github.com/gnucash/gnucash and ```github``` for git@github:GnuFiBux/gnucash, following the [GnuCash Git Wiki Page](https://wiki.gnucash.org/wiki/Git#User_and_.22Normal.22_Contributors).
* Your feature branch is based on master, so it has a bunch of commits that don't belong in maint. To fix this please do the following:
```shell
git branch -D Bug787401-TestReport-Definition
git checkout -b maint -t upstream/maint
git checkout -b Bug787401-TestReport-Definition
git cherry-pick master..7c76f30
```
* Please don't use the same commit description for multiple commits. You can use
```shell
git rebase -i
```
to select each commit in turn and give it a meaningful description, or to collapse commits together if that makes sense.
After that you can
```shell
git push -f github Bug787401-TestReport-Definition
```
```-f``` means 'force' and it will replace your existing branch on Github with the new commits and automatically update this pull request.

File: gnucash/report/report-system/test/test-report-system.scm [hunk 1]
```diff
@@ -0,0 +1,68 @@
...
+    (set! test-report-system-flag #t)
+    (test-runner-factory gnc:test-runner)
+    (test-begin "Testing/Temporary/test-report-system") ;; if (test-runner-factory gnc:test-runner) is commented out, this
+                                                            ;; will create Testing/Temporary/test-asset-performance.log
+    (test-assert "Minimum Report Definition" (test-check1))
+    (test-assert "Missing GUID detection" (test-check2))
+    (test-assert "Detect double GUID" (test-check3))
+    (test-assert "Report with Full Argument Set" (test-check4))
+    (test-end "Testing/Temporary/test-report-system")
+    (set! test-report-system-flag #f)
```
christopherlam: I think it'll be better to move the (set!) call 1 line up - the (test-end) statement *must* be the last one in (run-test) to successfully return #t or #f to the test-runner. If (set!) is the last statement, its return value is unspecified which renders the test useless.

gnucash/report/report-system/test/test-report-system.scm
```diff
@@ -0,0 +1,68 @@
...
+
+;; -----------------------------------------------------------------------
+
+(define (test-check1)
+  (gnc:define-report 'version "1" 'name "Test Report Template" 'report-guid "54c2fc051af64a08ba2334c2e9179e23")
+)
+
+;; -----------------------------------------------------------------------
+
+(define (test-check2)
```
christopherlam: stylistic choice.... `(define (test-check2) (if (not (gnc:define-report ...etc)) #t #f))` can be shortened to `(define (test-check2) (not (gnc:define-report ...etc))` directly

gnucash/report/report-system/test/test-report-system.scm
```diff
@@ -0,0 +1,68 @@
...
+    (string=? (gnc:report-template-renderer (gnc:find-report-template "54c2fc051af64a08ba2334c2e9179e24")) "Renderer")
+    (string=? (gnc:report-template-options-cleanup-cb (gnc:find-report-template "54c2fc051af64a08ba2334c2e9179e24"))"Options Clean-Up")
+    (string=? (gnc:report-template-options-changed-cb (gnc:find-report-template "54c2fc051af64a08ba2334c2e9179e24"))"Options Changed")
+    (not (gnc:report-template-in-menu? (gnc:find-report-template "54c2fc051af64a08ba2334c2e9179e24")))
+    (string=? (gnc:report-template-menu-path (gnc:find-report-template "54c2fc051af64a08ba2334c2e9179e24"))"Menu Path")
+    (string=? (gnc:report-template-menu-name (gnc:find-report-template "54c2fc051af64a08ba2334c2e9179e24"))"Menu Name")
+    (string=? (gnc:report-template-menu-tip (gnc:find-report-template "54c2fc051af64a08ba2334c2e9179e24"))"Menu Tip")
+    (string=? (gnc:report-template-export-types (gnc:find-report-template "54c2fc051af64a08ba2334c2e9179e24"))"Export Types")
+    (string=? (gnc:report-template-export-thunk (gnc:find-report-template "54c2fc051af64a08ba2334c2e9179e24"))"Export Thunk")
+  )
```
christopherlam: this function can be renamed as

    (define (test-check4)
      (test-assert "test-report template"
        (gnc:define-report 'version "1" 'name "Test Report Template" '...)))
      (test-equal "test-report-version is 1"
        "1"
        (gnc:report-template-version (gn....)))))
etc

GnuFiBux:
All  comments have been integrated into the latest commit, except for the last one.
If (test-check4) is updates as proposed above, then the logging in LastTest.log changes to

[pass] line:15, test: Minimum Report Definition
[pass] line:16, test: Missing GUID detection
[pass] line:17, test: Detect double GUID
[pass] line:65, test: test-report-version
[pass] line:65, test: test-report-version  (--> missing: [pass] line:18, test: Report with Full Argument Set)

which I think is not nice. So I kept (test-check4) as it is.

christopherlam:
Ok no probs with last function. It's a stylistic choice, I like to generate lots of [pass] statements myself.

My next concern is that this commiit changes main code to use a different code path during test or regular run. *I'm not sure this happens anywhere else through code*. You're effectively telling code 'run this path only when not testing' which somewhat defeats the purpose of testing. I'll defer to more experienced devs here - my thoughts would be to drop the test-report-system-flag -- if the test suite needs to call gnc_error_dialog then so be it; let it run and succeed/fail.

My other thought would be stylistic again- instead of calling variables 'old-style' better call them 'legacy-2.x' or something? In the future your changes will become old too :)

jralls:
About the test-system-flag thing: Turning off an error dialog during automated tests isn't a problem; in fact it's probably necessary because it would be unusual for the GUI to be loaded during unit testing.

The problem--with the original code, not with this PR--is that there's GUI code at all in report-system. GUI code should be confined to report-gnome, and the report-system code should be signaling the error somehow. How the error is handled is the responsibility of other code: It could be the GUI displaying a message box, the html including an error message, or logging the error to the tracefile... but what ever it is, it's not report.scm's responsibility.

jralls:
No, it's too squishy for a bug report. It's more something to keep in mind as we (and @christopherlam in particular) reengineer the report system.

christopherlam:
I can understand the need to remove gui calls but not entirely sure exactly how the `(gnc:define-report)` is being called. I know they're being triggered with every report code e.g. at the end of transaction.scm, but not sure what exactly is causing the transaction.scm to be run. So, where will the gui error message be stated? As a html report perhaps?

jralls:
Thanks!

PR #361, opened 2018-06-04
Eradicate gtkhtml in .scm

christopherlam:
This commit does away with anything gtkhtml related. Pulled out from another PR because it's different enough.
Also POTFILES.in is always regenerated during CMake - it means during development it was always annoyingly being scanned for changes. Deleting from repo means it is not tracked as directed by .gitignore in project root.


gjanssens:
Thanks!

PR #362, opened 2018-06-05
[test-transaction] modify test to be acceptable to MacOS

christopherlam:
This modification will convert string to number (using a customized function defined in same file) to remove need to encode £ sign in scheme.

@jralls test-html-utilities-srfi64.scm uses emoji in scheme code... will this also fail in MacOS? (or is it OSX?)

### DIFF
--- a/gnucash/report/standard-reports/test/test-transaction.scm
+++ b/gnucash/report/standard-reports/test/test-transaction.scm
@@ -172,7 +172,7 @@
 
     ;; Here we set foreign currencies
 
-    (gnc-commodity-set-user-symbol foreign2 "£")
+    (gnc-commodity-set-user-symbol foreign2 "#")
 
     (with-account
      gbp-bank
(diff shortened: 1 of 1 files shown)


jralls:
@christopherlam I think that the problem with £ is that you encoded it in UTF-8 and that guile expects some other encoding. The output of test-transaction on Debian 9, while it passes, includes ```[pass] line:588, test: GBP original currency totals = ?4```. That suggests that Guile doesn't understand the UTF-8 there either, it's just that the ones-and-zeros happen to be the same in both the file and the output so it matches.

jralls:
Oh, and it's currently "MacOS". From 10.0 to 10.12 it was "Mac OS X", and "X" was Roman numeral 10.

christopherlam:
Ok I'll avoid using extended chars in future; if it passes with '?' output on console I think it's still fine.

jralls:
It would probably be better to figure out how to get extended characters into and out of guile.

Another approach that you can use in the html code is to use html character entities, eg ```&#128077;``` for 👍.

christopherlam:
Hm the use of html code is the only one that's not going to work... there's no inbuilt html entity parser in guile; AFAIU it uses Unicode internally everywhere (except when linking with C, or when manipulating bytes directly). I'm trying every combination of hex/unicode to print pound signs in console and they don't work. So, guile->console isn't perfect, yet guile->html-report is absolutely fine.

jralls:
Of course. Webkit takes care of converting the html entities to glyphs. For unit tests guile sees only the characters, so it will match.

christopherlam:
This is better? no more £ just use # instead lol.
This is not a test for html so let's use an easy currency symbol.

jralls:
That does the trick, thanks, though I still think it would be useful to figure out how to portably tell guile "£".

PR #363, opened 2018-06-05
add tags to enable div/span and css for 'modern' layout

christopherlam:
Not quite 100% sure this is final. But it's a great start.
1. stylesheets and documents. this mess of a hack normally renders stylesheet's style-text only and report's contents. This commit will make stylesheet style-text *append* document's CSS style-text. As a result, a report can call `(gnc:html-document-make-style-text! doc "custom-css")` which will be tagged at the end of stylesheet's CSS.
2. `<div>` and `<span>` tags are here! Any html-object (e.g. plain string, or html-table, or html-markup-b etc) can be enclosed with `(gnc:make-html-div obj)` or `(gnc:make-html-span obj)` tag, or `(gnc:make-html-div/class "classname" obj)` or `(gnc:make-html-span/class "classname" obj)`.

Together these two commits should allow css-based layouts.


christopherlam:
hmm instead of /class, legacy code calls them /markup... I've renamed to `(gnc:make-html-div/markup "classname" div-contents ...)`

christopherlam:
So this iteration can be merged.

I've piggy-backed onto the existing gnc html "style" to allow creation of `<div class="bla">abc</div>` by using `(gnc:make-html-div/markup "bla" "abc")`.

Unfortunately the old-school "style" is really attempting to be a poor-man's CSS, so, the following construct `(gnc:make-html-div/markup "parent-class" (gnc:make-html-div "child-text"))` which is meant to produce `<div class="parent-class"><div>child-text</div></div>` ends up becoming `<div class="parent-class"><div class="parent-class">child-text</div></div>` because the old wizards were trying to create css using html attributes.

So I've had to define (gnc:make-html-div) to call (gnc:make-html-div/markup) with an empty-string classname whose resultant html becomes `<div class="parent-class"><div class="">child-text</div></div>`, which is better than the unintended inherited classname.

Old html-style mechanism -> :-1: 

Perhaps when all can be converted to CSS I'll get rid of html-style altogether.

jralls:
Nice work! Thanks.

PR #364, opened 2018-06-07
test-invoice

christopherlam:
this tests invoice, easy-invoice, fancy-invoice content.
eventually tests will be amended when layout+css is introduced.


jralls:
Thanks!

PR #365, opened 2018-06-09
charts upgrade part 1

christopherlam:
Part 1 of html-chart & chartjs upgrades
Mainly combining net-linechart & net-barchart


jralls:
This is failing to build for me. When merged directly into current maint compiling the reports segfaults; pulling it onto its own branch matching yours (i.e. starting at 6f0a3c43) and bisecting, it's broken at the first commit because " In procedure primitive-load-path: Unable to find file "html-jqplot" in load path". html-jqplot was removed in ba2c1f6a 27 April.

christopherlam:
I'll fix this and ping when ready.

jralls:
It still segfaulted, but this time I looked at the crash report and found that it was in gtkstyleproviderprivate.c. Starting from a clean build directory works so I imagine the problem is that a .go needs to be rebuilt and somehow ninja isn't picking that up.

So merged. Thanks.

christopherlam:
I saw these build issues (seem to bork with invoice.go?)... they do not reappear upon clean rebuild. Will investigate, but this is not my forte!

jralls:
Don't put too much effort into it. It's probably just a CMakeLists.txt dependency issue and it's not that big a deal to have to build from a clean build-dir.

PR #367, opened 2018-06-11
Use HINTS instead of PATHS to tell cmake where to look for GTEST + GMOCK

goodvibes2:
files so it selects the paths in GTEST_ROOT + GMOCK_ROOT rather than the
system libraries. See https://cmake.org/cmake/help/v3.5/command/find_path.html?highlight=find_path

### DIFF
--- a/common/cmake_modules/GncAddTest.cmake
+++ b/common/cmake_modules/GncAddTest.cmake
@@ -125,9 +125,9 @@ function(gnc_gtest_configure)
     set(GMOCK_ROOT $ENV{GMOCK_ROOT})
   endif()
   find_path(GTEST_INCLUDE_DIR gtest/gtest.h
-    PATHS ${GTEST_ROOT}/include ${GMOCK_ROOT}/gtest/include /usr/include)
+    HINTS ${GTEST_ROOT}/include ${GMOCK_ROOT}/gtest/include /usr/include)
   find_path(GTEST_SRC_DIR src/gtest-all.cc
-    PATHS ${GTEST_ROOT} ${GMOCK_ROOT}/gtest /usr/src/gtest)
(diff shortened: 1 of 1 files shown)


goodvibes2:
There were no other examples of this problem in maint.

goodvibes2:
Should I do another commit to update googletest dependency to 1.8.0 in README.dependencies?

jralls:
Thanks!

> Should I do another commit to update googletest dependency to 1.8.0 in README.dependencies?

No, 1.7.0 works fine. The versions in README.dependencies are supposed to be the minimum that's know to work, not necessarily what everyone uses.

gjanssens:
> No, 1.7.0 works fine. The versions in README.dependencies are supposed to be the minimum that's know to work, not necessarily what everyone uses.

Which begs the question then why @goodvibes2 couldn't build with Ubuntu's default googletest (1.7.x) ? From the mailing list thread I remember 1.8.0 wasn't picked up because 1.7 was present on his system...

jralls:
I suspect that he didn't try because I told him to follow the recipe in the 14.04 docker script. 14.04 didn't have googletest at all, 16.04 has 1.7.0, as does Debian Stretch that I use frequently without any issues.

goodvibes2:
It failed when I used 
cmake args -DGTEST_ROOT=$HOME/github/gtest/googletest -DGMOCK_ROOT=$HOME/github/gtest/googlemock
because GncAddTest.cmake used PATHS instead of HINTS, and it looked for .h in /usr/include/gtest.
Just did a test now without GTEST_ROOT + DGMOCK_ROOT and indeed it compiles using /usr/include gtest.
I'll update the Ubuntu wiki again.

jralls:
Right. And this PR's change is the Right Thing To Do because we want GTEST_ROOT to override the system if it's set.

PR #368, opened 2018-06-12
Bug 796537 - Transaction Report cannot sort by "num"

christopherlam:
This is technically a bug in the TR regarding book split-action property. The reworked TR did not retrieve book split-action properly which means it was always assuming split-action was #f for sorting.

This commit redefines sortkey-list to accept split-action? as a parameter to generate the correct sortkeys. Unfortunately it will make it more complicated; I'm not sure if or how to simplify.


jralls:
Thanks!

PR #370, opened 2018-06-13
test-net-charts

christopherlam:
This adds basic net-barchart and net-linechart testing.

It will be a prelude to #365, ensuring the reports survive. It does not yet test the data generated.


christopherlam:
sorry wrong approach

PR #371, opened 2018-06-13
Maint test net charts

christopherlam:
This adds basic net-barchart and net-linechart testing. It will be a prelude to #365, ensuring the reports survive. I think #365 is now ready as well.

It does not test the data generated... data is encoded in javascript `d0.push([x,y])` methods.


christopherlam:
With the final html-chart.scm, the data is sent to chartjs in a JSON object, and this can be tested back using guile-json.

PR #372, opened 2018-06-15
Further testing commits

christopherlam:
So... while I've only just written test-net-charts, it turns out expanding to create basic null-report for all charts is not too difficult. This is to ensure all standard charts survive ChartJS upgrade. The groundwork has ben laid and any further tests can be added here.


christopherlam:
I think these are harmless - upgrades tests (net-charts -> all-charts / transaction->foreign-currency / basic balance-sheet), eradicates `css?` in eguile, sanitizing.
The tests are for now rather basic; I'll upgrade specific testing prior to work.

jralls:
Thanks. Merged, though there was a hiccup between code and github so github doesn't know it yet.

PR #373, opened 2018-06-16
Add tooltips to the register

Bob-IT:
First up, not sure if this should be against maint or master but can be changed if required.
The first patch adds the ability to create register tooltips based on cell type.
The second adds a couple of tooltips, one is based on the recn cell, this shows the reconciled date and the void reason if the transaction is voided. The other is based on the associate cell and shows the uri associated to a transaction if present. The reason for tooltips was I was thinking about a wide register layout and this would allow me to remove the void reason cell.

This could be extended to tooltips for the header but not sure if that would be required.

The last patch is for Bug 796545, not sure whether this covers the problem but maybe a start to see what is possible. There was another bug which mentions a tooltip for the date cell but I am unsure about that.

### DIFF
--- a/gnucash/register/ledger-core/split-register-model.c
+++ b/gnucash/register/ledger-core/split-register-model.c
@@ -497,6 +497,17 @@ gnc_split_register_get_fcredit_label (VirtualLocation virt_loc,
     return _("Credit Formula");
 }
 
+
+static char *
+gnc_split_register_get_default_tooltip (VirtualLocation virt_loc,
+                                     gpointer user_data)
+{
+    SplitRegister *reg = user_data;
+    const char *tooltip = gnc_table_get_entry(reg->table, virt_loc);
+
+    return g_strdup (tooltip);
+}
+
 static gnc_numeric
 get_trans_total_amount (SplitRegister *reg, Transaction *trans)
 {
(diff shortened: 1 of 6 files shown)


jralls:
In a multi-currency transaction the register with the "other" (i.e. not the transaction) currency shows the amount in the "other" currency decorated with the transaction currency's symbol. E.g, a transaction for C$3.75 from a CAD asset account to US$3.38 in a USD expense account; when the expense account is showing hovering over the debit amount shows "C$3.38 [CAD]". That's not right: It should either be "$3.38 [CAD]" (which is kind of redundant except for showing the transaction currency) or "C$3.75 [CAD]".

Bob-IT:
OK, I know what I think I did wrong trying to fix the 'General Journal' view, I changed line 722 from ...
    print_info = gnc_account_print_info (account, TRUE);
to
    print_info = gnc_commodity_print_info (currency, TRUE);
but I should of added another case statement...
    case GENERAL_JOURNAL:

Seems to work locally, maybe you can make the changes and test before I push. I think I may remove the 'transaction currency' in brackets, will probably confuse people. Also should there be a similar tooltip for balance ?

jralls:
Well, that does change it to $3.38 [CAD]: Correct but as I said rather useless; it already says that in the field. If you remove the transaction currency in brackets it's *completely* useless. Why have a tooltip that just repeats the content of the field?

Bob-IT:
OK, it looks like I do not understand the requirements for that bug so have dropped the commit from the PR.

jralls:
OK, thanks.

PR #375, opened 2018-06-21
Bug795471 - Some more changes to the budget view

Bob-IT:
Just some small changes to the budget view based on feedback on Bug 795471


gjanssens:
Thanks!

PR #376, opened 2018-06-23
Reconciliation report amendments for 796614

christopherlam:
This commit mainly amends the reconcile report in a few ways:
1. date filter applies to reconciliation date, rather than posted date. If reconciliation date does not exist (i.e. unreconciled split), the transaction is included anyway.
2. the balances are split to debit & credit (i.e. 'funds in', 'funds out') yet the subtotals are counted separately.

Supporting amendments:
1. test for reconciliation report
2. splits filtering is now done before sorting, because we want to sort a short splitlist.
3. modify trep-renderer to accept keyword args to support reconciliation report.
4. (unrelated) place the render-options summary above subtotal-grid


jralls:
Thanks!

PR #377, opened 2018-06-27
Multicolumn balsheet & profit'n'loss report :)

christopherlam:
Features
1. as above
2. balsheet has anchors which lead to the exact split that documents the account balance
3. profit&loss has anchor to TR to 'expand' income/expense period
4. both have anchor to charts in net-charts.scm
5. testing included

Both original-currency and report common-currency supported
Balsheet common-currency will use prices appropriate to balsheet date, or latest.
Profit&Loss common-currency will use prices appropriate for begin/midpoint/end of period, or latest.
NO cost-price valuation is implemented.
Almost 100% functional approach.... there's no multicolumn-acct-table-add-row or multicolumn-acct-add-accounts or add-columns API... all data and options are sent and processed immediately in add-multicolumn-acct-table. 
I wonder if it should be labelled "beta, please verify numbers and totals are correct."?

It is planned that the add-multicolumn-acct-table eventually gets upgraded into report-utilities.scm... It can be a nice replacement in other reports - trial-balance etc. Perhaps in the future I'll add 'contracts' i.e. `(if (not (list? col-data)) (throw "col-data must be a list"))` to aid in development.

### DIFF
--- a/gnucash/report/standard-reports/CMakeLists.txt
+++ b/gnucash/report/standard-reports/CMakeLists.txt
@@ -9,6 +9,7 @@ set (standard_reports_SCHEME_2
     advanced-portfolio.scm
     average-balance.scm
     balance-sheet.scm
+    balsheet-pnl.scm
     budget-balance-sheet.scm
     budget-barchart.scm
     budget-flow.scm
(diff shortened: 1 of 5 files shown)


christopherlam:
Closing down. Need to rebuild with an appropriate subtotal strategy.

PR #378, opened 2018-07-01
Stress tests for options

christopherlam:
This complicated test suite is designed to prevent further issues such as bug 796696.

First, a general test-stress-optionslist.scm was generated programmatically - see source in test-stress-options.scm. This documents all boolean and multichoice options obtainable in reports in memory. I've skipped other option types - unsure how to test them.

The test-stress-options.scm will load the optionlist, then iterate over each report (ie. most reports, except those that produce unusual errors while loading eg eguile-based reports). Each report is rendered as is; after which the option list is cycled until all options are tested. Not all combinatorial permutations are performed - this would take far too long.

e.g. if Transaction Report is tested, the following sample set of options illus:
Display / Account Name - true/false
Display / Account Code - true/false
Display / Amount Display - none/single/double
Report / Price source = average/latest/nearest/weighted

The total number of combinations should be 2x2x3x4 = 48; however the above combination is only run 4 times, with options true/true/none/average, false/false/single/latest, true/true/double/nearest, false/false/none/weighted. (This skips many combinations e.g. false/true/any/any)... This provides some coverage for testing. NO report-contents are tested. It simply ensures the report does not crash. Ideally it would offer 12 combinations for pairwise combinatorics, but I do not know an easy scheme-based algorithm.

Incidentally I found commodity-utilities.scm would crash if avg-prices or weighted-avg were selected on a sparse datafile, and I've fixed it. And job-report.scm would crash on startup when the 'account' variable was empty (indicating no AR type account exists). I've fixed this too in this PR.

On my machine this test runs in 5s.


christopherlam:
PS not sure how to treat test-stress-optionslist.scm - should it be included in CMakeLists.txt?

christopherlam:
I'd anticipate a future report can record its options in test-stress-optionslist.scm to allow testing.
A benefit of this approach is illustrated by cherrypicking 9cd7d36 only, observing test failure, and cherrypicking 5a4481e to fix it.

The test-stress-optionslist.scm can always be easily regenerated by the snippet in test-stress-options.scm; the snippet can be added in a report, and called by a `(mydump)` in the renderer.

christopherlam:
I have an idea brewing to generate automated regression tests for reports, and wish to canvas opinion.

As we know there is very little coverage of test options in current code, and the above 'stress-test' will only test a fraction of report-option combinations -- it only tests a report as many times as the largest number of possibilities per option - eg if all options were booleans, it'd test twice - all off and all on. Ideally we'd implement pairwise testing. It is very difficult to generate a comprehensive test suite, and I have not found a scheme-based pairwise testing generation code. I thought of two possible ideas, both hackish. 

1) We can add public-domain http://burtleburtle.net/bob/math/jenny.html into borrowed so that it is run (as a guile subprocess) by test-stress-options.scm once for each report, and read its output back into scheme, and for each report, run every combination in the test to ensure the report is run. This method IMO is 'hackish but elegant'. My preference is this one, but it does create a call to an "external" test suite generator.

2) My other proposal is to modify *every* report slightly to add, somewhere within the report definition a 'report-test-parameters' object -- eg in transaction report, there would be 

```
(define-report options-test-parameters
        (cons 'parameters 
              (list '("Display" "Amount")
                    '("Display" "Memo")
                    '("Display" "Account")
                    ...etc...)
        (cons 'combinatorics
             (list '(2 0 1)
                   '(1 1 0)
                   '(0 1 1)
                   ...etc...))
```

The `parameters` specify options to be tested; the `combinatorics` will specify the output generated by jenny, tailored for each report. Then test-stress-options.scm can set options for each combination and test the report does not crash. This is more labour intensive and will require careful instruction on how to amend reports.

jralls:
Proposals like that belong on the developer list rather than on a pull request.
Did you not read in the first paragraph of Jenny.c "testing every combination is overkill"?

There's also a time concern: Testing works by being run routinely as part of the development cycle, and it already takes as long to run the tests as it does to build the program. Even your limited "stress test"  adds half-again to the test time; even minimal combinatorics testing will balloon that many times.

 Let's instead focus on consolidating the duplication in reports to make the code smaller and more manageable and on getting good unit-test coverage of the consolidated code.

christopherlam:
Will do, meanwhile please feel free to cherry-pick the other commits except the two related to test-stress-options. There are some crashers, and I've removed my homedir from the coverage functions.

christopherlam:
Ok the latest changes are: 
1. rebased onto #380 which fixes a few internal functions to allow more reports to run.
2. I've split (tests) to run tests twice - first on an empty book, second with a few transactions. This dislodged a few more report crashers, already fixed in 1 above.
3. the test-stress-options will by default run the simple tests - if all options are bools, run twice only - all off, all on. If the environment variable COMBINATORICS is set to any value, run the combinatorial stress tests.

christopherlam:
ready...

jralls:
Thanks. I added borrowed/jenny/jenny.c for completeness, haven't yet added a CMakeLists.txt for it.

COMBINATORICS needs a full-path, relative doesn't work, presumably because the tests are run cd'd to somewhere.

PR #379, opened 2018-07-03
add tests for report options

GnuFiBux:
Hi,

here is my next follow-up on Bug787401 "tests for the report system", this time for report options.

### DIFF
--- a/gnucash/report/report-system/test/CMakeLists.txt
+++ b/gnucash/report/report-system/test/CMakeLists.txt
@@ -18,6 +18,7 @@ set(scm_test_report_system_SOURCES
 
 set (scm_test_report_system_with_srfi64_SOURCES
   test-html-utilities-srfi64.scm
+  test-report-options.scm
   test-report-system.scm
   )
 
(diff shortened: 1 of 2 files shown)


jralls:
One doesn't normally waste test time on simple setters. Is there a point to these tests?

jralls:
> ideally a unit test would test the functionality of a module rather than its implementation...we want to test the function does exactly what we want..."

Not quite. A test obviously tests the implementation, it being a bit difficult to test anything else. The objective is to ensure that the implementation achieves the desired functionality (as @christopherlam says in the later bit). My point is that if a function simply takes a list of arguments and assigns them to a set of variables there's not much point in testing it: It can be easily determined by inspection that it does that. Expanding, if a function starts off setting a bunch of variables and then does other stuff there's no need to test the setting, get straight to the other stuff.

GnuFiBux:
Hi John, Christopher,
thanks for your feedback.
Many interesting aspects that I think are good to open endless philosophical discussions.
One thing I totally agree with: this set of tests is not complete.

Not to waste even more time on those kind of discussions:
Consider it an offer. Take it or leave it.
Let's see if I can make time to "bug" you with the findings that came of of this waste. I think I had the one or the other spot...

jralls:
It's a very practical issue: Unnecessary/not-useful tests add code clutter and take additional run time with no payback.
Since the tests all pass I will be interested to see your bug reports, and since I don't see any benefit to the tests I'll "leave it".

PR #380, opened 2018-07-10
report bugfixes

christopherlam:
A few further report crashers found on stress-tests.


christopherlam:
This is now included in #378

PR #382, opened 2018-07-14
Increase of MAX_TAB_COUNT to change appearance of dialog "Book Options"

DiMan:
Increase of MAX_TAB_COUNT to change appearance of dialog "Book Options" in german language.

The dialog "Book Options" looks in german different, because it has 5 tabs: four standard tabs and additionaly one tab for tax settings, which appears only for german language. The 1 to 4 tabs appearance on top (horizontal mode). The appearance switchs to vertical mode (with tabs on the left side) for 5 and more tabs. Here the setting MAX_TAB_COUNT is increased to 5. This change should make the appearance of dialog "Book Options" similar in german and other languages.

Dialog "Book Options" in english, only for comparison:
bildschirmfoto_2018-07-14_20-58-57

Dialog "Book Options" in german before change:
bildschirmfoto_2018-07-14_20-58-05

Dialog "Book Options" in german after change:
bildschirmfoto_2018-07-14_20-52-51

### DIFF
--- a/gnucash/gnome-utils/dialog-options.c
+++ b/gnucash/gnome-utils/dialog-options.c
@@ -74,7 +74,7 @@ static QofLogModule log_module = GNC_MOD_GUI;
  * Point where preferences switch control method from a set of
  * notebook tabs to a list.
  */
-#define MAX_TAB_COUNT 5
+#define MAX_TAB_COUNT 8
 
 /* A pointer to the last selected filename */
 #define LAST_SELECTION "last-selection"
--- a/gnucash/gnome-utils/dialog-options.c
+++ b/gnucash/gnome-utils/dialog-options.c
@@ -74,7 +74,7 @@ static QofLogModule log_module = GNC_MOD_GUI;
  * Point where preferences switch control method from a set of
  * notebook tabs to a list.
  */
-#define MAX_TAB_COUNT 4
+#define MAX_TAB_COUNT 5
 
 /* A pointer to the last selected filename */
 #define LAST_SELECTION "last-selection"


jralls:
This is a global change that will affect all GtkNotebooks in dialogs with 5 tags. That includes several reports' option dialogs. I think we should discuss it with the user community first, so please start a thread on gnucash-user.

DiMan:
I just sent an email to gnucash-user.

DiMan:
I have change the value MAX_TAB_COUNT to 8. See small discussion under https://lists.gnucash.org/pipermail/gnucash-user/2018-July/078224.html and the listing of all GnuCash dialogs: 
[GnuCash - Dialogs with GtkNotebooks.pdf](https://github.com/Gnucash/gnucash/files/2207239/GnuCash.-.Dialogs.with.GtkNotebooks.pdf)

DiMan:
As written in mailing list, the report with 7 tabs still looks fine (<800x600). 
https://lists.gnucash.org/pipermail/gnucash-user/2018-July/078229.html

Which setting would you prefer:
* `MAX_TAB_COUNT=6`: all current reports are covered with this setting
* `MAX_TAB_COUNT=7`: + 1 tab additionally as reserve for future flexibility. (window < 800x600)
* `MAX_TAB_COUNT=8`: + 2 tabs additionally as reserve for more future flexibility (windows may be> 800x600 for some design themes)

jralls:
Let's stop at 6. I merged and changed it in the merge commit. I actually think that the number of options pages on some reports is already a bit much.

DiMan:
Thanks!

PR #383, opened 2018-07-15
Some Bug Fixes

Bob-IT:
Bug 796739 has also been mentioned on the user list
Also another stab at the reconcile tree views, had some inspiration about getting the vertical scrollbar width and using that as padding for the last column.
Bug 796751 restores the column order and by using previous commit the tick boxes are not covered.
The last commit aligns the totals label with the account column at first display but does not track it just like before.

### DIFF
--- a/gnucash/gnome-utils/dialog-preferences.c
+++ b/gnucash/gnome-utils/dialog-preferences.c
@@ -1229,6 +1229,8 @@ gnc_preferences_dialog_create(GtkWindow *parent)
     gnc_builder_add_from_file (builder, "dialog-preferences.glade", "retain_days_adj");
     gnc_builder_add_from_file (builder, "dialog-preferences.glade", "tab_width_adj");
     gnc_builder_add_from_file (builder, "dialog-preferences.glade", "date_formats");
+    gnc_builder_add_from_file (builder, "dialog-preferences.glade", "toolbar-styles");
+    gnc_builder_add_from_file (builder, "dialog-preferences.glade", "toolbar-icon-sizes");
     gnc_builder_add_from_file (builder, "dialog-preferences.glade", "gnucash_preferences_dialog");
 
     dialog = GTK_WIDGET(gtk_builder_get_object (builder, "gnucash_preferences_dialog"));
--- a/gnucash/gnome-utils/gnc-embedded-window.c
+++ b/gnucash/gnome-utils/gnc-embedded-window.c
@@ -1,6 +1,6 @@
 /*
  * gnc-main-window.c -- GtkWindow which represents the
- *	GnuCash main window.
+ *  GnuCash main window.
  *
  * Copyright (C) 2003 Jan Arne Petersen <jpetersen@uni-bonn.de>
  * Copyright (C) 2003 David Hampton <hampton@employees.org>
@@ -38,6 +38,7 @@
 #include "gnc-plugin-manager.h"
 #include "gnc-ui.h"
 #include "gnc-window.h"
+#include "gnc-prefs.h"
 #include "dialog-utils.h"
 
 /* Static Globals *******************************************************/
@@ -308,6 +309,13 @@ gnc_embedded_window_add_widget (GtkUIManager *merge,
     if (GTK_IS_TOOLBAR (widget))
     {
         priv->toolbar = widget;
+
+        gtk_toolbar_set_style (GTK_TOOLBAR(priv->toolbar),
+            gnc_prefs_get_int (GNC_PREFS_GROUP_GENERAL, GNC_PREF_TOOLBAR_STYLE));
+
+        // prefs has only small and large icons so add 2 to get right enum
+        gtk_toolbar_set_icon_size (GTK_TOOLBAR(priv->toolbar),
+           (gnc_prefs_get_int (GNC_PREFS_GROUP_GENERAL, GNC_PREF_TOOLBAR_ICON_SIZE)) + 2);
     }
 
     gtk_box_pack_start (GTK_BOX (priv->menu_dock), widget, FALSE, FALSE, 0);
--- a/gnucash/gnome-utils/gnc-main-window.c
+++ b/gnucash/gnome-utils/gnc-main-window.c
@@ -3341,6 +3341,28 @@ gnc_main_window_update_tab_position (gpointer prefs, gchar *pref, gpointer user_
     LEAVE ("");
 }
 
+static void
+gnc_main_window_update_toolbar (gpointer prefs, gchar *pref, gpointer user_data)
+{
+    GncMainWindow *window;
+    GtkToolbar *tb;
+    GncMainWindowPrivate *priv;
+    gint selection;
+
+    window = GNC_MAIN_WINDOW(user_data);
+    priv = GNC_MAIN_WINDOW_GET_PRIVATE (window);
+    tb = GTK_TOOLBAR(priv->toolbar);
+
+    selection = gnc_prefs_get_int (GNC_PREFS_GROUP_GENERAL, GNC_PREF_TOOLBAR_STYLE);
+    if (gtk_toolbar_get_style (tb) != selection)
+        gtk_toolbar_set_style (tb, selection);
+
+    selection = gnc_prefs_get_int (GNC_PREFS_GROUP_GENERAL, GNC_PREF_TOOLBAR_ICON_SIZE);
+    // prefs has only small and large icons so add 2 to get right enum
+    if (gtk_toolbar_get_icon_size (tb) != selection + 2)
+        gtk_toolbar_set_icon_size (tb, selection + 2);
+}
+
 /*
  * Based on code from Epiphany (src/ephy-window.c)
  */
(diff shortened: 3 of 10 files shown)


jralls:
Great, thanks, but when fixing bugs please paste in the bug number and summary from bugzilla, either as the whole title line and explain what you did in the rest of the commit message *or* summarize the change in the  title line, add more explanation if needed, then paste in the bug number and summary at the end, maybe preceded by  something like "fixes", "partly fixes", or "fix part n". Otherwise at release time I have to notice that you did the commit and get the actual from the bug for the release notes.

I've changed the relevant commit messages and pushed.

Bob-IT:
Ok, will try and remember for future fixes.

gjanssens:
Unfortunately I was unable to review this before it got committed as I have several reservations on the changes proposed here.

To start with the most important one: the addition of preferences to tweak visual style. I explicitly removed the one for toolbar button style in gnucash 2.6 with reason so I prefer not to add it back. Here is why I think these should not be added:

1. The [Gnome Human Interface Guidelines design principles](https://developer.gnome.org/hig/stable/design-principles.html.en) advise to "Use configuration options sparingly". (I thought to have read other references to this but can't find them at the moment). So let's consider whether this option is proper to add.

1. I think gnucash should focus primarily on accounting and reserve code complexity for areas that improve our accounting capabilities. So if we add options we should first consider whether it improves the flexibility or power of gnucash in the context of accounting.

1. The two options added on the other hand are visual tweaks that don't affect accounting in any way. It's mostly a user preference of how to represent the tool bar.

1. Second question to ask is whether the preference is gnucash only or platform related. For these two, I think the answer is clearly the second. If a user wants the tool bar to have both icons and text, it's a reasonable assumption the user would want that for all the tool bars in other applications as well. The same goes for icon size. That's a good clue this should be handled at a higher level, not in gnucash directly.

1. Thirdly it's always valuable to do some field study. For example look at other (mature) applications for comparison. Digikam uses small icons with text, this style can't be changed. Libre Office uses small icons by default with no text. Icon size can be changed, but text can't be added. Gimp doesn't really have  toolbar, but a tool button pallet (considering the huge amount of buttons). They are fixed size and no text. We may look at typical Windows applications as well but I currently don't have access to any. You will note that in most cases the choice is made by the application designer, not the user.

1. Gtk developers have removed the facility for the user to decide on tool bar style by themselves. I don't always agree on such decisions but in this case it makes sense. It's really the application that should be designed in such a way there is no need for a preference (read the option reference again in the Gnome HIG).

1. The issue was raised on the user list that the icons are hard to interpret without labels. Only the tool tip gives more details but that slows down the user. That is a valid complaint and also holds the clue for a true solution: we should either improve our icons or choose for ourselves to add labels if we can't improve the icons.

So to summarize: adding these options are giving in to weak design. Our toolbar needs improvement in its own right. Adding an option to toggle visual styles doesn't solve that. It only papers over it.

What can we do ?

1. evaluate the icons we use. Are the expressing their meaning clearly enough ?
1. evaluate which actions to add on the tool bar in the first place. The hig suggests we should only add actions that act on the whole page or application wide (like switching to a different view). There are several buttons there though that only act on a small part of the page. For example the cancel button, the duplicate button, the view splits button all act on the currently selected transaction. Perhaps they can be displayed differently ? For example using an [action bar](https://developer.gnome.org/hig/stable/action-bars.html.en) or a [pop over](https://developer.gnome.org/hig/stable/popovers.html.en) or in some way directly inline. To be investigated further.

gjanssens:
Then I want to come back also on the reconcile changes.

I'm pleased you found a way to take the scroll bar width into account. On the other and I'm not convinced we should move the check box column back to the end although I'm less black and white on this one.

The reason is again integration in the wider platform. Most contexts I have seen in which you can select more than one item by means of a check box will put the check box in front. So doing the same in the reconcile view helps new users work with it more easily as they can reuse knowledge taken from other applications.

I know there was an argument somewhere having the check box separate from the amount makes it harder to reconcile. I can see two potential reasons why this issue was raised:

1. The user thinks s/he has to click in the check box in order to tick an item. I can't test now, but I believe the user can click anywhere on the line including the amount to tick or untick one particular line. If that's not the case we have to fix that.

1. When reviewing what was ticked it may be more difficult to see which amounts are ticked and which ones aren't. I have no clear answer to this one. I would like to find "prior art" but haven't had time to look how other applications handle such multiple select situations. I know Digikam allows multi-select on tags, but I consider that a bad implementation (cumbersome to use).

gjanssens:
Coming back to the tool bar story I would like to see the new options removed from the code again before the 3.3 release.

We may think of other ways to accommodate users if the need remains after a serious redesign of the tool bar. However removing options is always a worse user experience than not adding one in the first place.

jralls:
OK, then it has to be words-and-icons for everything. The current words-on-a-few-random-icons sucks, especially when many of the icons are unintelligible. e.g., how on earth does a speed limit sign with an x instead of a number mean "delete the current transaction"? 

For 4.0 I'd be happy to get rid of the toolbar completely.

jralls:
Dang. I meant to lead with "Welcome Back!"

gjanssens:
Heh, I'm not fully back but I had a sleepless night and wifi at my disposal :) I'll still need some more time to get through the backlog.

As for the tool bar, I agree icons and text is currently the best option. Given the limited space available I would propose to use small icons as well.

I also gave the check boxes some more thought. Indeed the first motivation to move them was the overlapping scroll bars. I added back then already I liked it because it made the check boxes more consistent with other selection lists. After some sleep I now think we should move away from check boxes completely and instead only use an image of a check mark. A check box urges the user to click there. On the other hand I perceive a check mark image more as a read-only indicator and having less forcefully attracting the mouse pointer.

jralls:
Huh? The check-boxes aren't read-only, they're how the user tells GnuCash that the split is reconciled. They'll be pre-checked if the split has been set to cleared either because the user did so or the splits were imported via AQB, OFX, or QIF. Sometimes in that case the user may need to un-check them if she downloaded past her statement date.

gjanssens:
From an earlier comment:
>  ... I believe the user can click anywhere on the line including the amount to tick or untick one particular line. If that's not the case we have to fix that.

I assumed this in my later comment.

(I don't have gnucash on the system I'm using to type these messages so I can't verify for myself)

So in my view we don't need a check box as a target to click on. The selection unit is a whole line. The user clicks on the lines s/he wants to add or remove. This is a multi-select list paradigm. What we do need is a visual indicator which lines are selected. Hence the suggestion of a check mark image/icon as such an indicator. The indicator could also be a  full line highlight (dark blue in the default theme), but that conflicts with navigating through the list by keyboard. So I think a check mark is better.

I also meant to add that for me it feels more natural to add a check mark icon at the end of the line than a full check box.

For completeness I do find some conflict in my own reasoning. I don't know whether this would be discoverable enough for new users. All modern selectable lists that come to mind right now always have true check boxes at the beginning of the line, never at the end (tags in digikam, almost all selection methods on Android,...). So that's an established pattern. I vaguely remember this may be different on iOS devices, which (in my memory) behave more like I'm proposing now.

I really want to have more examples of how other applications have dealt with similar multiple select lists before finally deciding. I won't have time to do this research before somewhere next week though.

jralls:
That doesn't work, at least on MacOS or Debian 9. If you think about it it would be kind of weird if it did. What does work and is perhaps what you're thinking of is selecting a row by clicking on it and then hitting the spacebar to toggle the checkbox. But that's not really discoverable and I don't think that it would work without having a checkbox there or writing a custom event handler for the whole row... and the absence of a checkbox would make even clicking rather undiscoverable.

I'm actually experiencing a similar "user experience" in Chrome on MacOS 10.14 beta: Most of the time the usual controls don't render. I can sometimes guess where to click if I'm familiar with the page but often I have to switch to Firefox in order to fill out forms. It's not something I'd want to inflict on users.

Making things more familiar for new users is all well and good but we already have a pretty substantial bunch of established users and they've made it clear that they're not terribly happy with the UI changes from moving to Gtk3. Bob moved the reconcile checkbox back to the right side because of users complaining, after all.

gjanssens:
After reading the code (on github) I see what you mean. The current implementation is based on a GNCQueryView and it has no facility to act directly on a click event anywhere in the line. So at least I agree this is not something we want to change for the 3.x series. It has enough behavioral changes as is already.

I still think it's worth to work out my idea though and see I how far I can go. But that can happen on a feature branch geared to master.

As for the user feedback I think we should always search for the core issue they point out and find the optimal solution for it, which can be either "as it was before" or something new that has additional benefits. In this case the complaint was that it was hard that the point required to click was far away from the most important information (the amount). This is because the current view requires the user to click on the check box. The "as it was before" solution would be to just move the check box back to the end. Which is what Bob did and that's sufficient for 3.x.
However that doesn't solve what I perceive as an additional issue: the click target is relatively small (I'm sensitive to RSI so I tend to care for such things). Making the view act on any click anywhere on the line would solve that. Whether or not the state representation of the line is a check box or a check mark is an additional design decision with still other motivations (as I explained earlier).

I should really do a mockup. It takes way too many words to explain all this. To be continued somewhere in the future and not in the comments of this PR.

PR #384, opened 2018-07-15
Color accounts

Bob-IT:
Just some thing I have been meaning to do, this dialog allows you to select an account and use that accounts colour for all sub accounts.


jralls:
Nice change, thanks!

PR #385, opened 2018-07-16
Do not execute test-real-data.sh with $SHELL

vinc17fr:
Avoid a failure when the SHELL environment variable is set to /bin/zsh, for instance.
Fixes https://bugs.gnucash.org/show_bug.cgi?id=796763

### DIFF
--- a/libgnucash/backend/xml/test/CMakeLists.txt
+++ b/libgnucash/backend/xml/test/CMakeLists.txt
@@ -103,6 +103,6 @@ set(test-real-data-env
 )
 add_test(NAME test-real-data
    COMMAND ${CMAKE_COMMAND_TMP}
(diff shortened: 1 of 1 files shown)


onlyjob:
That's right, shebang should be sufficient, there is no need to force shell.

gjanssens:
I believe the $SHELL notation is needed on Windows or at leat will be if we ever want to support the ninja build system.

gjanssens:
It has taken a while (which had mostly to do with me getting access again at a Windows development box), but I have landed a fix to make our build system independent of the user's preferred shell. It is still explicitly executing a shell, but we actively configure it to use bash everywhere (unless the user deliberately overrides it by setting the GNC_SHELL environment variable). That should make the system work on linux, Windows and MacOS.

While these fixes obsolete this patch, your contribution has been appreciated nonetheless.

PR #386, opened 2018-07-22
Change some dialogues to GtkWindows with some small changes

Bob-IT:
Not sure if this all can be on maint but happy to split if required.

### DIFF
--- a/gnucash/gnome/dialog-trans-assoc.c
+++ b/gnucash/gnome/dialog-trans-assoc.c
@@ -48,7 +48,7 @@ enum GncAssocColumn {DATE_TRANS, DESC_TRANS, URI_U, AVAILABLE, URI_SPLIT, URI, U
 
 typedef struct
 {
-    GtkWidget    *dialog;
+    GtkWidget    *window;
     GtkWidget    *view;
     const gchar  *path_head;
     gboolean      valid_path_head;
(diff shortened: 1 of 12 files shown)


jralls:
Thanks, but...
Changing the dialogs to windows means that they pop up at coordinates 0,0, which on my dual-screen setup is a long way from where I have the main window parked. Although it broke the stacking behavior (and that's more important, which is why I went ahead and merged this) the transient-for setting put the dialog box on top of the calling window. Can we get the location of the (former) parent and try to place the windows close by?

Bob-IT:
I assume that is on first load, but they do save there position right for subsequent loading correctly ?

jralls:
That is on first load. If I move it somewhere on the left-side monitor then its position is remembered and stored (though just for the currently-open book, of course). If I put it on the right-side monitor it's restored to the same position on the left-side one, so it would seem that the coordinates that you're getting and storing for dialogs are relative to the origin of the monitor that they're on but when the window is restored it goes to those coordinates relative to the whole screen.

The main window gets restored to the monitor on which is was last displayed, so there's some difference in logic somewhere.

Bob-IT:
OK, will have a look...

christopherlam:
About 5ad584ca3b67236c563f5da25f935ee9f188592e - Interesting approach...
It seems to make the Report Options always load on the bottom-left of my ubuntu xfce, instead of the centre of the desktop - haven't tried win32 nightly. What's more annoying is that Esc does not close the Report/Book options anymore. I know the 'C'ancel hotkey exists, but usually prefer Esc...

Bob-IT:
Chris, I assume you did not change any settings after window placement, may need to change that. Will look at Esc key.

christopherlam:
> Chris, I assume you did not change any settings after window placement, may need to change that. Will look at Esc key.

You're right -
initial load bottom left, move window, press Alt-C to cancel -> reloads at bottom-left
initial load bottom left, move window, change options, press OK -> reload at saved location

gnucash/gnome/dialog-trans-assoc.c
```diff
@@ -299,7 +305,10 @@ get_trans_info (AssocDialog *assoc_dialog)
     /* Get list of Accounts */
     accts = gnc_account_get_descendants_sorted (root);
 
+    /* disconnect the model from the treeview */
     model = gtk_tree_view_get_model (GTK_TREE_VIEW(assoc_dialog->view));
+    g_object_ref (G_OBJECT(model));
+    gtk_tree_view_set_model (GTK_TREE_VIEW(assoc_dialog->view), NULL);
```
gjanssens: Same remark as with the find account dialog. You don't have to get-clear-set the model (and it seems to me you missed the clear step in this case). Just create a new model and set it.

gnucash/gnome/dialog-trans-assoc.c
```diff
@@ -165,7 +165,10 @@ assoc_dialog_update (AssocDialog *assoc_dialog)
     GtkTreeIter       iter;
     gboolean          valid;
 
+    /* disconnect the model from the treeview */
     model = gtk_tree_view_get_model (GTK_TREE_VIEW(assoc_dialog->view));
+    g_object_ref (G_OBJECT(model));
+    gtk_tree_view_set_model (GTK_TREE_VIEW(assoc_dialog->view), NULL);
```
gjanssens: Do you expect long transaction association trees such that it requires disconnecting the model (which is fine, just curious) ?

libgnucash/app-utils/gnc-component-manager.c
```diff
@@ -827,6 +827,9 @@ gnc_close_gui_component_by_session (gpointer session)
 
     list = find_components_by_session (session);
 
+    // reverse the list so class like dialog-options close before window-report
+    list = g_list_reverse (list);
+
```
gjanssens: Clever!

PR #387, opened 2018-07-24
[3.x] [test-balance-sheet] upgrade to test sample book with 2 subtotal strategies

christopherlam:
This PR incorporates majority balance-sheet and income-statement test suite. It creates some crafted equity transactions dated in 1970 and some income transactions in 1980.

- creates equity->asset transactions for USD, incorporating both leaf accounts and parent account, and several buy/sell amounts USD/GBP and USD/FUND, in year 1970. This aims to test various commodity prices conversion strategies.
- creates several pricedb entries for both USD/GBP and USD/FUND from 1971 to 1975
- creates income->asset amounts for year 1980
- generates reports with weighted-average/average-cost/pricedb-nearest/pricedb-latest amounts, and reports with toggled incl-zb-accts and omit-zb-bals, and toggled show-fcur / show-exchange-rates.
- generates 2 subtotal strategies - recursive-bal and multilevel
- tests amounts from balance-sheet and income-statement from above report options. It tests only the expected changes e.g. omit-zb-bals being true/false will hide/show a $0.00 amount.

The above tests are preparation for upcoming balsheet-pnl upgrade, a single .scm which creates both balsheet and pnl reports. I expect all important amounts to survive the upgrade. (Note upcoming balsheet-pnl does not render a small original-currency->report-currency table in table cell - it will render either original currency amounts, or the report-currency amounts). I expect the combined balsheet-pnl.scm report to generate the whole report more quickly than previous separate files, because it has less dependencies and uses pure functional programming concepts.


christopherlam:
When the new balsheet-pnl is successfully merged after these commits, I'll start on account-summary, cash-flow etc and will probably upgrade balsheet-pnl's add-multicolumn-acct-table to a supporting module.

christopherlam:
Yay!

PR #389, opened 2018-07-28
Bugfix 796783 - handling invalid date range in "get transaction..."

DiMan:
The GnuCash attempts to get the transactions from the online bank even
though the date range entered is incorrect.

Examples for incorrect user inputs:
1) "to"-date is smaller than "from"-date
2) "to"-date is in the future

In both cases GnuCash tries to get a connection to the online bank.
It takes a while. After that GnuCash shows an error message:
"Error on executing job. Status: error - job contains errors".
The error message is very general and does not say what is wrong.

It is better to check the user inputs *before* the request
will be send to the bank, and to show a error message if the entered date
range is wrong, *without* creating the connection to the online bank.

### DIFF
--- a/gnucash/import-export/aqb/gnc-ab-gettrans.c
+++ b/gnucash/import-export/aqb/gnc-ab-gettrans.c
@@ -143,6 +143,23 @@ gnc_ab_gettrans(GtkWidget *parent, Account *gnc_acc)
         g_debug("gnc_ab_gettrans: gettrans_dates aborted");
         goto cleanup;
     }
+
+    /* validate date range: from_date should be less than or equal to to_date. */
+    if(GWEN_Time_toTime_t(from_date) > GWEN_Time_toTime_t(to_date))
+    {
+        g_debug("gnc_ab_gettrans: invalid date range - from_date is greater than to_date");
+        gnc_error_dialog (GTK_WINDOW (parent), _("Invalid date range: \nStart date cannot be greater than the End date!"));
+        goto cleanup;
+    }
+
+    /* validate to_date: to_date cann't be in the future */
+    if(GWEN_Time_toTime_t(to_date) > gnc_time(NULL))
+    {
+        g_debug("gnc_ab_gettrans: invalid date range - to_date is in the future");
+        gnc_error_dialog (GTK_WINDOW (parent), _("Invalid date range: \nEnd date cannot be in the future!"));
+        goto cleanup;
+    }
+
     /* Use this as a local storage for the until_time below. */
     until = GWEN_Time_toTime_t(to_date);
 


DiMan:
Please check if the texts of the error messages are ok, as I am not a native speaker... thanks!

gnucash/import-export/aqb/gnc-ab-gettrans.c
```diff
@@ -143,6 +143,23 @@ gnc_ab_gettrans(GtkWidget *parent, Account *gnc_acc)
...
+        g_debug("gnc_ab_gettrans: invalid date range - from_date is greater than to_date");
+        gnc_error_dialog (GTK_WINDOW (parent), _("Invalid date range: \nStart date cannot be greater than the End date!"));
+        goto cleanup;
+    }
+
+    /* validate to_date: to_date cann't be in the future */
+    if(GWEN_Time_toTime_t(to_date) > gnc_time(NULL))
+    {
+        g_debug("gnc_ab_gettrans: invalid date range - to_date is in the future");
+        gnc_error_dialog (GTK_WINDOW (parent), _("Invalid date range: \nEnd date cannot be in the future!"));
```
gjanssens: Is a date in the future effectively causing a bank error ?

jralls:
I hadn't heard of in-app notifications before and the design pattern doesn't offer any clues on implementation. This [stack overflow article](https://stackoverflow.com/questions/45431512/gtk-in-app-notifications-api-referece) does. It's non-trivial, and it's a significant style change that shouldn't be applied in just one place: We first need to mock this up and see how it looks on Windows and MacOS before we decide to use it and if we do we should consider every GtkMessageBox in GnuCash for conversion. For this PR let's stick with a GtkMessageBox.

gjanssens:
Hmm, agreed. Looks like I'm piling up design projects for 4.x these days...

gjanssens:
Coming back to this. While the in-app notifications are new, design that doesn't interrupt the user isn't. And I have personally implemented several such changes in the past. For example the preview page in the csv importer will dynamically update an error list and will prevent clicking the next button until all errors are fixed (or skip errors is used). Same for the business payment dialog. It will only enable the ok button if the input makes sense. This dialog will use a less intrusive icon to indicate something is missing or incorrect.

This design can also be applied to the get transaction... dialog. At a minimum if the dates are not correct disable the button to continue. Adding a status field or an icon with tooltip does make the solution complete without using new paradigms.

I don't see the added benefit of a message box. They should be avoided, I'm gradually working on this and point out all new situations where they could be avoided instead of introduced.

gnucash/import-export/aqb/gnc-ab-gettrans.c
DiMan: Yes, in cases with the invalid date range GnuCash tries to get the connection to the bank and the bank send an error messages. It looks in GnuCash like this:
bildschirmfoto_2018-07-24_20-23-54

If I repeate the same requests on the console with `aqbanking-cli` I get the following error codes:

**"to-date" > today** or **"to-date" > "from-date"**:
```
...
Dialog not aborted, assuming PIN is ok
AqHBCI finished.
5:2018/07/31 22-45-28:aqbanking(3915):banking_online.c:  119: Error executing backend's queue
4:2018/07/31 22-45-28:aqbanking(3915):banking_online.c:  137: Not a single job successfully executed
Postprocessing jobs
Job Get Transactions: error
Resetting provider queues
Executing Jobs: Finished.

```

DiMan:
I agree, my solution is not user-friendly because it interrupts the user action: The error message appears after the "get transaction" dialog has been closed. It would be better to avoid invalid user input directly in the dialog, for example by disabling the "OK" button for incorrect user data.

I'm coming from the Java and Data Analysis world, and I still have a lot to learn how to do things like that in C. I will be glad if you will implement it.

gjanssens:
It was and is not my intention to take this over from you (there's enough to do for me already). I was rather hoping that my pointers would allow you implement this yourself.

DiMan:
Can you please say me, where can I find a business payment dialog in GnuCash?
I will look at the solution with the validation of user input and make it similar in the "get transaction" dialog.

DiMan:
Thanks for info!

jralls:
@DiMan , are you still working on this?

DiMan:
Hello Jon,
I did not continue working at this point because I currently have no time.
Sorry. Can you please close this PR. I will commit it later when I have more time for GnuCash.
Thanks, DiMang

jralls:
OK, thanks for taking time to get back to me.

PR #390, opened 2018-07-29
Some more fixes hopefully

Bob-IT:
The first commit adds the use of the Escape key to close the GtkWindow on the three dialogues I had changed previously.

In the next commit I think I fixed the restoring of the dialogues geometry when multiple monitors are used as it was previously only looking at the primary one. I do not have multiple monitors but was able to test it by adding one to a VirtualBox VM which seemed to work.
There is also a similar fix to the adjust for screen size function that was also only looking at the primary monitor.

On first use of these dialogues, with the dialogues that do not have any dimensions specified, GTK seems to align the new dialogue on top of the parent but if a dimension is specified it would be located on the left side of the monitor. The next commit tries to locate these dialogues when first used by using the parent window as a reference point. The subsequent two commits could be squashed but wanted to distinguish between the changes to the other files as simple changes and the slightly more involved.

Also fixed a couple of transient parent warnings on some dialogues.

### DIFF
--- a/gnucash/gnome-search/dialog-search.c
+++ b/gnucash/gnome-search/dialog-search.c
@@ -1345,7 +1345,7 @@ gnc_search_dialog_create (GtkWindow *parent,
 
     gnc_search_dialog_init_widgets (sw, title);
     if (sw->prefs_group)
-        gnc_restore_window_size(sw->prefs_group, GTK_WINDOW(sw->dialog));
+        gnc_restore_window_size(sw->prefs_group, GTK_WINDOW(sw->dialog), parent);
     gtk_window_set_transient_for(GTK_WINDOW(sw->dialog), parent);
     gtk_widget_show(sw->dialog);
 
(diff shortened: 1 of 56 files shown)


jralls:
Bob,
I'm a bit puzzled about your saying that Gtk places dialogs without dimensions centered on the parent. Is that the dialogs that you've converted from being GtkDialogs to GtkWindows? If so, what parent? I thought that the point of that change was that they'd be new top-levels so that the user could change their stacking.

Bob-IT:
Bad choice of words, I was using parent as the main application window. If it is a top level GtkWindow / GtkAssistant with no dimensions defined it would be placed centrally on screen which may of been by chance on top of the main application window. If dimensions were defined it would be aligned to the left side. This is only on the first display as the geometry would be saved and then restored subsequently.

jralls:
Wow, a month escaped! Sorry for taking so long. It would be better if we could always put windows near the main window, it can be a real pain on a multi-monitor setup to have them open on the other monitor. At least dragging them back will stick. now.

jralls:
Thanks!

PR #391, opened 2018-07-31
Bug787401 - Test for the report system - HTML Tests

GnuFiBux:
Hi there, next step to transform the tests from Bug787401 into pull request.

Here are some test for the HMTL code generation.
Mainly simple positive tests, only a few simple negative tests.

You find some hints for potential bugs in there which I will take care of later, for now I ignore the tests that might fail due to these potential bugs (I want to finish the transformation of Bug787401 into pull requests first).


gnucash/report/report-system/test/test-report-html.scm
```diff
@@ -0,0 +1,217 @@
...
+      (gnc:html-document? test-doc)
+      (not (gnc:html-document-style-sheet test-doc))
+      (null? (gnc:html-document-style-stack test-doc))
+      (gnc:html-style-table? (gnc:html-document-style test-doc))
+      (not (gnc:html-document-style-text test-doc))
+      (string-null? (gnc:html-document-title test-doc))
+      (not (gnc:html-document-headline test-doc))
+      (null? (gnc:html-document-objects test-doc))
+      (string=? (gnc:html-document-render test-doc)
+        "<html>\n<head>\n<meta http-equiv=\"content-type\" content=\"text/html; charset=utf-8\" />\n<title>\n</title></head><body></body>\n</html>\n"
```
christopherlam: I'd think guile does allow multiline strings which are definitely easier to read, and also if the html is amended in code it'll be much easier to make small amendments to this test string.

christopherlam:
Feedback on string output given inline.

I also personally would still advocate multiple (test-equal "test" a b) statements... my thinking is the main purpose of creating tests is ensuring that the module still works while I refactor inner functions... so, if I were to modify (gnc:html-markup-i) and introduce a bug, I'd hope *all* other tests would pass while (gnc:html-markup-i) test would fail. This approach has helped me tremendously hunt down bugs and refactor correctly.

For example I'd test (gnc:html-markup-i) building up from something like the following -- this way I know 100% that html-markup-i works well.

```
(define (render t)
 (let ((test-doc (gnc:make-html-document)))
   (gnc:html-document-append-objects! test-doc (list (gnc:make-html-text t))
   (gnc:html-document-render test-doc))

 (test-equal "html-markup-i"
   "<i>italic</i>"
   (render (gnc:html-markup-i "italic"))
```

GnuFiBux:
The "long" lines have now been converted to multi-lines, definitely easier to read.
No additional test statement have been added. As mentioned in PR#360 it has impact on LastTest.log which I think should be avoided for this kind of "regression" test.
For development testing I agree that more test statements are useful, and the contents of LastTest.log in that case is of lower interest.

christopherlam:
P.S. grateful for tests - jqplot tests will be worthwhile too, and will also validate an onward upgrade to chartjs in #316

GnuFiBux:
the \n characters actually belong to the return value of the renderer (not added by me ;-) )

GnuFiBux:
hi John,

regarding the newline:
indeed, the renderer returns something like "text1\ntext2\ntext3\n"
and all I did is to put it to
"text1\n\
text2\n\
text3\n"
i.e. all I did was adding the slashes at the end of the lines for text1 and text2 and pressing enter to make it more readable.

Regarding your comments about how to write unit tests and your repo example:
cool stuff
I'll try to incorporate it as good as I can, thanks for the intro

gnucash/report/report-system/test/test-report-html.scm
```diff
@@ -0,0 +1,380 @@
...
+
+;; -----------------------------------------------------------------------
+
+(define (test-check1)
+  (let (
+         (test-doc (gnc:make-html-document))
+       )
+    (and
+      (gnc:html-document? test-doc)
+      (not (gnc:html-document-style-sheet test-doc))
```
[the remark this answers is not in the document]
GnuFiBux: see commit from today

gnucash/report/report-system/test/test-report-html.scm
```diff
@@ -0,0 +1,380 @@
...
+      ;; the row-markup hash table which is
+      ;;      - not updated on deletion of a row
+      ;;      - not updated anywhere else in the code
+      ;;      - not used anywhere else in GnuCash
+      ;;      --> should be removed
+      ;;      (same goes for gnc:html-table-row-markup,
+      ;; gnc:html-table-set-row-markup-table! gnc:html-table-set-row-markup!)
+
+      ;; Reset table data due to bug above:
+      (gnc:html-table-set-data! test-table '())
```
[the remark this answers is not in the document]
GnuFiBux: see changes in the comit from today

GnuFiBux:
I hope I could address all changes with the commit from today

gnucash/report/report-system/html-document.scm
```diff
@@ -128,7 +128,8 @@
 ;; returns the html document as a string, I think.
 (define (gnc:html-document-render doc . rest)
   (let ((stylesheet (gnc:html-document-style-sheet doc))
-        (headers? (if (null? rest) #f (if (car rest) #t #f)))
+        (headers? (if (null? rest) #t (if (not (car rest)) #f #t)))
```
jralls: Did you really mean to flip the first #f to #t?

gnucash/report/report-system/html-document.scm
```diff
@@ -391,7 +392,8 @@
                (lambda (obj doc)
                  (gnc:html-document-render-data doc obj))
                ;; if the object is #f, make it a placeholder
-               (if obj obj " ")))
+               ;;(if obj obj " ")))
+               obj))
```
jralls: This change causes:
The following tests FAILED:
	 93 - test-invoice (Failed)
	107 - test-transaction (Failed)
	108 - test-balsheet-pnl (Failed)
	109 - test-income-gst (Failed)

With an error 
```
Throw to key `parser-error' with args `((#<input: string 10b506f50> "[GIMatch] b
roken for " (END . td) " while expecting " END boolean))'.
```

gnucash/report/report-system/html-document.scm
GnuFiBux: yes. if no further options are given, then the default behaviour should be to create the header. I think that is how it was before. possibly I made it #f myself for testing and then forgot to make it #t again...

gnucash/report/report-system/html-document.scm
GnuFiBux: yepp, I can also reproduce it on my end.
After looking into this a bit, it seems that #f was used as value for "empty" cell. If this is changed to "" then the test is passed. I have not found all places yet that use #f for empty cells to pass all tests. I keep looking into this.

Apart from this:
The xml->sxml converter doesn't like the #f value as only value in cells if they are used with the <boolean> tag, because it expects a closing tag that is not implemented.
Actually, that also should go for the <string> tag.

Enabling the <boolean> tag for #f and correcting the append-colomn! to produce more empty cells intensifies the effect on the failing reports.

I try to come up with a proposal, if it is small enough. Otherwise it might be more clever to simply not apply the changes for 796831 and 796821.

GnuFiBux:
hm, still no success in travis even though all tests are passing now on my end.
I have downloaded the travis log.txt, will make a comparison, and post my result here.

christopherlam:
I'd have *major* concerns that for the tests to pass, the reports need to be amended.

Be aware that not all reports have adequate unit tests. 

test-transaction was failing because the amendments in this PR are causing #f cells to display something which is not valid XML. Therefore amending transaction.scm to pass this test is not good practice.

I'm 100% sure that many other reports are still using #f as an empty-cell to be rendered as " "; and are not being detected; and if this html-document amendment is merged, many reports will be rendering invalid XHTML.

I'd encourage you to find an appropriate fix that still allows #f to be renderable as empty-cell valid XHTML.

GnuFiBux:
Certainly, Christopher has a good point here.
But I am lacking a good idea how to make this displayable value a non-displayable value for such occurances - it feels more like a workaround for a workaround.

In my eyes, the remaining risk is that for those reports not convered in this PR there might be empty cells that are not empty anymore but show #f instead. The  reports as such should still work. Users might complain that the reports are not nice anymore, but the funtionality would still be there.

GnuFiBux:
voila, all tests passed

gnucash/report/report-system/html-document.scm
jralls: No, it was #f. The only place gnc:html-document-render-data is called with only one arg is in gnc:options->render and that isn't used in a way that the presence or absence of headers matters so it's really moot what the default is, but it's also not germane to your change, but changing the default of a function deserves a separate commit of its own with an explanation of why the old default is wrong, and it shouldn't be buried in a PR that's ostensibly about adding tests.

The second part of the change is non-functional: ```(if (car rest) #t #f)``` and ```(if (not (car rest)) #f #t)``` are the same thing.

GnuFiBux: yes, no, and yes

yes
it was #f before, so my statement above is incorrect, I flipped it by intention

no
and I do not vote for reverting back to 
reason:
because of  Bug796826 - Report HTML - not possible to suppress the document header
the header was always present, i.e. no matter if args are given or not, the default behaviour was to always include the header. If we leave it like that, chances are that it'll break user's custom reports.
Please confirm to revert back to #f.

yes
The second part of the change is non-functional and should be reverted back.

gnucash/report/report-system/html-document.scm
GnuFiBux: As said before, I do not see a good way to go around this.
According to my understanding we need an additional condition that makes it possible to decide between
#f -> <boolean>#f
and
#f -> " "

The change above is in
html-document.scm: (gnc:make-html-object obj)

which is used by
html-document.scm: (gnc:html-object-render obj doc)
html-document.scm: (gnc:html-document-add-object! doc obj)

both of them would require a change of interface (some additional arg, I guess).
but (gnc:html-document-add-object! doc obj) is spread, so here we would need to touch a couple of report implementations, too.

Only way I see:  simply revert back and stick to how it has been up to now.
As you might guess, that is not my favourite.
BTW: I also had to rework my custom reports for new time handling. Writing own reports bears a risk that  it might require maintenance ...

Suggestions/Advice welcome.

gnucash/report/report-system/html-document.scm
jralls: Ah, got it. The default here didn't matter because of the error on line 152. OK.

gnucash/report/report-system/html-document.scm
```diff
@@ -179,9 +179,10 @@
                (gnc:report-percent-done (* 100 (/ work-done work-to-do)))))
            objs)
 
-          (if (not (null? headers?))
+          (push "</body>\n")
+
+          (if headers?
               (begin
```
jralls: Don't need the begin any more.

gnucash/report/report-system/html-document.scm
jralls: Maybe we're thinking of this incorrectly: What is the actual use-case for putting a "#t" or "#f" in a report? Reports are for humans, so rather than outputting a raw boolean object one should output the appropriate (localized) word for the situation, e.g. "True"/"False" or "Yes"/"No".
In most other cases we're using ```#f``` to mean "nothing to display" and the conversion to the empty string is the right thing to do.

christopherlam: > Maybe we're thinking of this incorrectly: What is the actual use-case for putting a "#t" or "#f" in a report? Reports are for humans, so rather than outputting a raw boolean object one should output the appropriate (localized) word for the situation, e.g. "True"/"False" or "Yes"/"No".
> In most other cases we're using `#f` to mean "nothing to display" and the conversion to the empty string is the right thing to do.

I think this is the right answer. It is a very lispy convention that `#f` means default, or absence of an object. I do not think there's a need for appending boolean objects into a report at all.

Report objects are gnc:html-table, html-text, html-barchart, etc. From my understanding of old code, raw string or numbers are handled separately, stemming from legacy reports that have never been upgraded. Raw numbers are rendered with 2 decimal points, and strings are rendered raw. Unknown objects added to the document are rendered with invalid xml.

IMHO invalid objects should produce an error, which will signal the report writer to send proper document objects. We can do any report without resorting to coding boolean rendering separately, and without needing adjustments to current reports. I look forward to an amended PR incorporating tests... I will be very happy to refactor html-document which will explicitly clarify and document all acceptable document objects.

GnuFiBux:
to cut a long story short I would suggest:
- close this PR without merging
- open up a new PR only with tests that pass on the current design (only add a test file, no changes in existing code)
- open a seperate PR for title and header handling (bugs 796826 and 796827)
- open a separate PR for table row append/prepend handling (bugs 796829 and 796831)
- forget about the boolean false handling (leave it as it is, cancel bug 796828)

what do you think?

jralls:
I don't think it's necessary to make a new PR. I do think we should document that booleans should be handled directly by report code rather than generically and be done with it. You can just rebase -i dropping ac8f057 and force-push to continue with this PR.

GnuFiBux:
thanks for the hint - what a nice tool ... I hope I did not overdo it with the clean-up.

gnucash/report/report-system/test/test-report-html.scm
gjanssens: As far as I can see this is handled in the next commit.

gjanssens:
As far as I can tell all review requests have been addressed and travis passes.

The only issue left is that the PR branch has a number of unwanted merge commits from maint and master, preventing me from merging the PR directly. In order to no longer delay this PR, I have rebased the Bug787401-Test-Report-Html onto current maint in my local repository. That has eliminated all the unwanted merge commits and I have pushed the result directly from there.

So with that I'm closing this PR.

Thanks for your contributions!

PR #393, opened 2018-08-02
[test-stress-options] remove need for test-stress-optionlist.scm

christopherlam:
This commit modifies test-stress-options.scm to retrieve the report-list, and for each report, retrieve their options, and for each multichoice and boolean option, add to optionslist. Therefore there's no need to maintain a test-stress-optionslist.scm anymore. Yay! 

In the future the tests could be expanded to other option types eg string/date, however I'd only do on a private branch because there are far more code conditionals from boolean or multichoice options than any other.

During test-stress-options.scm development somehow I couldn't retrieve all the reports but now for some reason all reports now are accessible from scheme.

Still only 6.54s for non-combinatorics test-stress-options.scm.


gnucash/report/standard-reports/test/test-stress-optionslist.scm
```diff
@@ -1,972 +0,0 @@
-(define optionslist
```
christopherlam: This file is embarrassing. Hoping it's deleted soon!

gjanssens:
Thanks!

PR #396, opened 2018-08-07
Bug796792

Bob-IT:
Possible fix for Bug796792


gjanssens:
That was spot on. Thanks!

PR #397, opened 2018-08-14
Bug 796785 - dates cannot be keyboard entered on vendor bill

Bob-IT:
transactions

Make the gncEntryLedgerModel 'get_date_help' function the same format
as that used in the register.


Bob-IT:
Sorry about the lack of information, when I first looked at this I thought one was used only for the help message and the other one required the warning message as that was the one that saved the date.
Some times I wonder when a bug becomes something more, I had noticed the read-only flag before but did not think much of it as I assumed once set people will not normally change.

PR #398, opened 2018-08-15
scheme progress

christopherlam:
~This commit will modify gnome-utils.scm to maintain an internal list of gui-log messages. This list is modified by calling (addto-gui-log type str), type being either 'warning or 'error.~

~When the GUI is ready, a call to (process-gui-log) will call the gnc-warning-dialog or gnc-error-dialog calls, and empty gui-log.~

~I'm not 100% sure of function call names. Ideally the list would be maintained in C[++] but this can be a useful stop-gap to separate report definition and gui calls. The only functional change is that the report error dialogs will now be displayed *after* all reports are loaded, instead of *during* each report load.~

~The main benefit here is to separate GUI calls from scheme reports. There's now no need to maintain a test-report-system-flag boolean in report.scm.~

Strategy updated to use much simpler (gnucash-ui-is-running) call. I think it'd be nicer to handle errors outside report definition, but this would require some C work I think.

### DIFF
--- a/gnucash/report/report-system/report-utilities.scm
+++ b/gnucash/report/report-system/report-utilities.scm
@@ -362,10 +362,10 @@
     (lambda (action commodity amount)
       (case action
 	((add) (add-commodity-value commodity amount))
-	((merge) (add-commodity-clist 
-		 (gnc-commodity-collector-list commodity)))
+	((merge) (add-commodity-clist
+                  (commodity 'list #f #f)))
 	((minusmerge) (minus-commodity-clist
-		      (gnc-commodity-collector-list commodity)))
+                       (commodity 'list #f #f)))
 	((format) (process-commodity-list commodity commoditylist))
 	((reset) (set! commoditylist '()))
 	((getpair) (getpair commodity amount))
@@ -383,28 +383,6 @@
     (for-each (lambda (collector) (merged 'merge collector #f)) collectorlist)
     merged))
 
-;; Bah. Let's get back to normal data types -- this procedure thingy
-;; from above makes every code almost unreadable. First step: replace
-;; all 'action function calls by the normal functions below.
-
-;; CAS: ugh.  Having two usages is even *more* confusing, so let's
-;; please settle on one or the other.  What's Step 2?  How 'bout
-;; documenting the new functions?
-(define (gnc-commodity-collector-add collector commodity amount)
-  (collector 'add commodity amount))
-(define (gnc-commodity-collector-merge collector other-collector)
-  (collector 'merge other-collector #f))
-(define (gnc-commodity-collector-minusmerge collector other-collector)
-  (collector 'minusmerge other-collector #f))
-(define (gnc-commodity-collector-map collector function)
-  (collector 'format function #f))
-(define (gnc-commodity-collector-assoc collector commodity sign?)
-  (collector 'getmonetary commodity sign?))
-(define (gnc-commodity-collector-assoc-pair collector commodity sign?)
-  (collector 'getpair commodity sign?))
-(define (gnc-commodity-collector-list collector)
-  (collector 'list #f #f))
-
 ;; Returns zero if all entries in this collector are zero.
 (define (gnc-commodity-collector-allzero? collector)
   (every zero?
@@ -417,8 +395,7 @@
 (define (gnc:account-get-balance-at-date account date include-children?)
   (let ((collector (gnc:account-get-comm-balance-at-date
                     account date include-children?)))
-    (cadr (gnc-commodity-collector-assoc-pair
-	   collector (xaccAccountGetCommodity account) #f))))
+    (cadr (collector 'getpair (xaccAccountGetCommodity account) #f))))
 
 ;; This works similar as above but returns a commodity-collector, 
 ;; thus takes care of children accounts with different currencies.
@@ -434,7 +411,7 @@
       (if include-children?
 	  (for-each 
 	   (lambda (x) 
-	     (gnc-commodity-collector-merge balance-collector x))
+	     (balance-collector 'merge x #f))
 	   (gnc:account-map-descendants
 	    (lambda (child)
 	      (gnc:account-get-comm-balance-at-date child date #f))
@@ -454,9 +431,9 @@
       (qof-query-destroy query)
 
       (if (and splits (not (null? splits)))
-	  (gnc-commodity-collector-add balance-collector
-				       (xaccAccountGetCommodity account)
-				       (xaccSplitGetBalance (car splits))))
+	  (balance-collector 'add
+                             (xaccAccountGetCommodity account)
+                             (xaccSplitGetBalance (car splits))))
       balance-collector))
 
 ;; Calculate the increase in the balance of the account in terms of
@@ -473,7 +450,7 @@
     (if include-children?
         (for-each
          (lambda (x)
-           (gnc-commodity-collector-merge value-collector x))
+           (value-collector 'merge x #f))
          (gnc:account-map-descendants
           (lambda (d)
             (gnc:account-get-comm-value-interval d start-date end-date #f))
@@ -520,10 +497,10 @@
   (let ((collector (gnc:make-commodity-collector)))
     (for-each 
      (lambda (acct)
-       ((if (reverse-balance-fn acct)
-	    gnc-commodity-collector-minusmerge
-	    gnc-commodity-collector-merge)
-	collector (get-balance-fn acct)))
+       (collector
+        (if (reverse-balance-fn acct) 'minusmerge 'merge)
+        (get-balance-fn acct)
+        #f))
      accounts)
     collector))
 
@@ -576,8 +553,7 @@
 (define (gnc:account-get-balance-interval account from to include-children?)
   (let ((collector (gnc:account-get-comm-balance-interval
                     account from to include-children?)))
-    (cadr (gnc-commodity-collector-assoc-pair
-	   collector (xaccAccountGetCommodity account) #f))))
+    (cadr (collector 'getpair (xaccAccountGetCommodity account) #f))))
 
 ;; the version which returns a commodity-collector
 (define (gnc:account-get-comm-balance-interval account from to include-children?)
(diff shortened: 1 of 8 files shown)


christopherlam:
I'm refining transaction.scm to be very well-documented.
The subsequent commits will simplify common loops into srfi-1 and schemey forms.
I wonder if the API trimming effort in 140b1f4 and 104bb1b are welcome?

jralls:
I'm in favor of trimming but with the binding APIs we have to consider the possibility that a user has written a custom report that uses API that GnuCash itself doesn't. Geert and I discussed this on IRC a couple of weeks ago in the context of my removing Timespec from everywhere and we decided that we should mark as deprecated API that we want to remove in 4.0. Guile provides an undocumented function for this, ```issue-deprecation-warning``` that takes a string description argument. The string should be along the lines of "foo is deprecated use bar instead".

christopherlam:
Ok, I can think of 2 methods to maintain deprecated functions for these external reports. As you can see I'm a noob here!
1. reintroduce in existing *-utilities.scm files, with some (issue-deprecation-warning)-type warning - perhaps it can check gnucash-version, and fail if version >= 4?
2. add a module *-utilities-deprecated.scm similar to engine-deprecated.[ch], but no idea how to automatically add this module onto the environment for these external reports.

gjanssens:
Option 2 is the cleanest.
You can add a (use-modules (path-to-*-utilities-deprecated.scm)) in gnucash/report/report-system/report-system.scm to keep them available.
I would group the use-modules line together with the export lines for deprecated functions. That makes it easier to do the cleanup in the 4.0 branch.
I'm not sure the gnucash version check is useful other than a reminder to the devs we should still drop these functions in 3.900. But I'm not sure this code is seen when the function isn't actually executed. As we don't have code in gnucash using these deprecated functions we wouldn't run into the version check failure. Unless you find a way this version check happens at compile time already (which is perhaps possible by adding it in an eval-when clause, or our begin-syntax macro).

jralls:
Right. Note that you *didn't* remove the export lines from report-system.scm, so the removal is anyway incomplete.

There is indeed no point in the version check, we'll be removing the deprecated functions as part of the preparation for releasing 3.900. As an aside, any version sensitive code needs to test for 3.900, not 4.0, otherwise it won't work until the actual 4.0 release.

christopherlam:
I'll pause these refactoring... report-utilities.scm will get some freshly-cooked tests soon.

christopherlam:
This set is now rebased on #406.

I'm struggling with *-deprecated.scm - I'd think easiest is to tack these deprecated functions at the end of report-system.scm...

jralls:
We don't want any code in report-system.scm, just use-modules, load-path, and export. If you can't figure out how to add a file just put the functions back where you deleted them from and add the issue-deprecation-warning.

christopherlam:
> We don't want any code in report-system.scm, just use-modules, load-path, and export. If you can't figure out how to add a file just put the functions back where you deleted them from and add the issue-deprecation-warning.

It is done. This branch based on #406 except the last failing commit.

gnucash/report/standard-reports/transaction.scm
```diff
@@ -1157,16 +1157,13 @@ be excluded from periodic reporting.")
...
                                  (let* ((currency (xaccTransGetCurrency (xaccSplitGetParent split)))
                                         (scu (gnc-commodity-get-fraction currency))
                                         (price (xaccSplitGetSharePrice split))
-                                        (price-decimal (gnc-numeric-convert price
-                                                                            (if (< scu 10000)
-                                                                                (* scu 100)
-                                                                                scu)
-                                                                            GNC-HOW-RND-ROUND)))
+                                        (price-decimal (gnc-numeric-convert
+                                                        price (max scu 10000) GNC-HOW-RND-ROUND)))
```
jralls: I think you want ```(max (* scu 100) 10000)```.

jralls:
7fde021 and e751795 have "[deprecate]" in their summaries, but they're not deprecating anything, they're removing local functions.

christopherlam:
Lots of undocumented (i.e. IMNSHO completely useless) "utility" functions in collectors.scm - although they're exported, i vote to eradicate them...

jralls:
Undocumented doesn't mean "completely useless" or justify removal, it means that the functions need to be documented.

Even if our code doesn't use a function and you can't think of a reason to use it, it's still possible that it's crept into some user's custom report. We need to give users a warning before we deliberately break things. That's why we're using the deprecation mechanism.

gnucash/report/standard-reports/transaction.scm
```diff
@@ -1157,16 +1157,14 @@ be excluded from periodic reporting.")
...
                                         (scu (gnc-commodity-get-fraction currency))
                                         (price (xaccSplitGetSharePrice split))
-                                        (price-decimal (gnc-numeric-convert price
-                                                                            (if (< scu 10000)
-                                                                                (* scu 100)
-                                                                                scu)
-                                                                            GNC-HOW-RND-ROUND)))
+                                        (price-decimal (gnc-numeric-convert
+                                                        price (max (* 100 scu) 10000)
+                                                        GNC-HOW-RND-ROUND)))
```
jralls: The commit description doesn't describe what the code does, which is set the price's displayed precision to at least 4 decimal places and at most the smallest currency unit's decimal places plus two more. Bitcoin and possibly other e-currencies has an SCU of 10^-9, so prices in BTC will be displayed with 11 decimal places. This does that. Japanese Yen, on the other hand, have an SCU of 1: It's all pennies. I'm not sure that anything is gained by displaying prices in JPY to 4 rather than 2 decimal places, but I'm also not sure that much is gained by displaying BTC prices with 11 instead of 9.

jralls:
Well, surgically undo is more likely to succeed than revert, but I think you're better off working forward rather than trying to go back 5 years. Fortunately the contents of neither collectors.scm nor report-collectors.scm is exported, though it looks like some specified wrappers are, so you have a fairly free hand to rewrite/remove them.

christopherlam:
> Well, surgically undo is more likely to succeed than revert, but I think you're better off working forward rather than trying to go back 5 years. Fortunately the contents of neither collectors.scm nor report-collectors.scm is exported, though it looks like some specified wrappers are, so you have a fairly free hand to rewrite/remove them.

Ok, any work in this direction will be based upon #316. I'd think the new chartjs-based charts will eventually be merged? Perhaps an announcement on userlist that any user charts (using old API) will eventually be obsolete? I'm happy with the refactoring and merging of charts myself in this PR.

I'll try undo the optimization from #316 onwards and experiment based on that branch. But otherwise I think #316 is feature-complete.

gnucash/report/report-system/test/test-report-utilities.scm
```diff
@@ -41,4 +55,385 @@
...
+      lst)
+    (list-set-safe! lst 5 1)
+    (test-equal "list-set-safe out-of-bounds"
+      '(1 3 #f #f #f 1)
+      lst))
+  (test-end "list-set-safe"))
+
+(define (test-gnc:monetary->string)
+  (test-begin "gnc:monetary->string")
+  (let* ((book (gnc-get-current-book))
```
jralls: gnc-get-current-book calls gnc-get-current-session and that creates a session if one doesn't already exist; it does so in C where Guile's garbage collection can't see it, so if you don't call gnc-clear-current-session at the end of the test, it leaks the session memory. The fix is to use (test-group-with-cleanup) instead of (test-begin) and use either your teardown function or (gnc-clear-current-session) as the cleanup function.

gnucash/report/report-system/report-utilities.scm
```diff
@@ -20,27 +20,25 @@
...
+                 (i (length l)))
+        (if (< i elt)
+            (loop (cons #f filler) (1+ i))
+            (set! l (append! l filler)))))
   l)
 
 ;; pair is a list of one gnc:commodity and one gnc:numeric
 ;; value. Deprecated -- use <gnc-monetary> instead.
 (define (gnc-commodity-value->string pair)
+  (issue-deprecation-warning "gnc-commodity-value->string deprecated")
```
jralls: The deprecation warning should tell the caller what they should do instead as well as say that the function is deprecated.

christopherlam:
All fixed... this is now based on #406 with improved deprecation warning.

jralls:
Thanks!

christopherlam:
Hmm sorry about that. Would the following addition to test-report-utilities.scm fix it?

;; Explicitly set locale to make the report output predictable
(setlocale LC_ALL "C")

gjanssens:
No but running
```
LC_ALL=C ninja-build check
```
does.

So you're on the right track. It also means you can probably reproduce the issue by running
```
LC_ALL=nl_BE ninja-build check
```

christopherlam:
It's `LC_ALL=nl_BE.utf8 ninja check` for me and seems to reproduce.

I admit I don't understand locale settings very much. IMO it's simpler to change tests to expect list of monetary pairs rather than strings, and should be locale-independent.

````
(define (collector->list coll)
  ;; input:  collector
-  ;; output: list of strings e.g. '("$25.00" "-£15.00")
-  (map gnc:monetary->string (coll 'format gnc:make-gnc-monetary #f)))
+ ;; output: list of monetary-mnemonic and amount pairs
+  (define (monetary->pair comm amt)
+    (cons (gnc-commodity-get-mnemonic comm) amt))
+  (coll 'format monetary->pair #f))
````

The above patch will change collector->list output from `'("$25.00" "£20.00")` to `'(("USD" . 25) ("GBP" . 20))` -- to my mind this is reasonable?

gjanssens:
That looks reasonable to me indeed.

christopherlam:
Ok https://github.com/christopherlam/gnucash/tree/maint-fix-test-report-utilities has my attempt. It seems to pass various `LC_ALL=xxx ninja check`. I'm hypothesizing that setting LC_ALL=C in code will set default report currency to USD, would you be able to check? If it works please feel free to merge latest commit.

It may be even more elegant to define collector->list as
````
(define (collector->list coll)
  ;; input:  collector
 ;; output: list of monetary-mnemonic and amount pairs
  (define (monetary->pair comm amt)
    (cons (string->symbol (gnc-commodity-get-mnemonic comm) amt)))
  (coll 'format monetary->pair #f))
````
which would output pure symbols & numbers eg `'((USD 25) (GBP . 20))` but some scheme purist consider this an abuse of symbols...

gjanssens:
That does fix it. So I have applied your commit. Thanks!

PR #399, opened 2018-08-16
Bug796725 Date Posted options fail

Bob-IT:
gnc_date_edit_get_date returns a value set to the start of the day 00:00:00 so there is a need to adjust value to 23:59:59 for LessThanEqual and GreaterThan and also use QOF_DATE_MATCH_DAY when comparing dates on one day only.

### DIFF
--- a/gnucash/gnome-search/search-date.c
+++ b/gnucash/gnome-search/search-date.c
@@ -179,9 +179,22 @@ gncs_validate (GNCSearchCoreType *fe)
 }
 
 static void
+gnc_search_date_set_date_from_edit (GNCSearchDate *fe, GNCDateEdit *de)
+{
+   /* The gnc_date_edit_get_date function returns a value set to the
+    * start of the day 00:00:00, use gnc_date_edit_get_date_end to get
+    * value for day end 23:59:59 for LessThanEqual and GreaterThan */
+
+    if (fe->how == QOF_COMPARE_LTE || fe->how == QOF_COMPARE_GT)
+        fe->tt = gnc_date_edit_get_date_end (de);
+    else
+        fe->tt = gnc_date_edit_get_date (de);
+}
+
+static void
 date_changed (GNCDateEdit *date_edit, GNCSearchDate *fe)
 {
-    fe->tt = gnc_date_edit_get_date (date_edit);
+    gnc_search_date_set_date_from_edit (fe, date_edit);
 }
 
 static GtkWidget *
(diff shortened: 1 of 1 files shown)


gnucash/gnome-search/search-date.c
```diff
@@ -178,10 +178,23 @@ gncs_validate (GNCSearchCoreType *fe)
...
+    /* time64 time_in comes from the gnc_date_edit_get_date which
+     * returns a value set to the start of the day 00:00:00, adjust
+     * value to 23:59:59 for LessThanEqual and GreaterThan */
+
+    if (fi->how == QOF_COMPARE_LTE || fi->how == QOF_COMPARE_GT)
+        return gnc_time64_get_day_end (time_in);
+    else
+        return time_in;
+}
+
```
gjanssens: Good.

Should we anticipate that gnc_date_edit_get_date may one day be changed to return 10:59:59 as time (which is our new default neutral time) ? That is to say this function may be more robust if it also acted on QOF_COMPARE_LT and QOF_COMPARE_GTE to return gnc_time64_get_day_begin.

jralls:
> Should we anticipate that gnc_date_edit_get_date may one day be changed to return 10:59:59 as time (which is our new default neutral time) ? 

gnc_date_edit_get_date should eventually return a GncDate which would be passed to gnc_time64_get_day_(begin|neutral|end) to obtain the time64 that one wants for a particular use-case. When that happens all of the using code will have to be refactored anyway and any "future proofing" is as likely to make that job more difficult rather than less so.

gjanssens:
Uhm, I read that as, yes, we should already use gnc_time64_get_day_(begin|neutral|end) now to explicitly set our intent. Not doing so would increase the risk of missing the subtlety of this situation when doing large refactorings in the future.
Quiet contrary to what you implied I'm afraid.

gnucash/gnome-search/search-date.c
jralls: Better:
```C
static void
gnc_search_date_set_date_from_edit(GncSearchDate* fe, GncDateEdit* ed)
{
    if (fi->how == QOF_COMPARE_LTE || fi->how == QOF_COMPARE_GT)
        fe->tt = gnc_date_edit_get_date_end (ed);
    else
       fe->tt = gnc_date_edit_get_date (ed);
}

static void
date_changed (GncDateEdit *date_edit, GNCSearchDate *fe)
{
    gnc_search_date_set_date_from_edit (fe, date_edit);
}
```
and same for gncs_get_predicate. ```gnc_date_edit_get_date_end``` sets the time to midnight using the ```struct tm``` from ```gnc_date_edit_get_date_internal``` so it will be faster than gnc_time64_get_day_end's round-trip into a GncDateTime and it clearly expresses the intent for future refactoring of gnc_date_edit.

Bob-IT:
I have used John's function and updated the PR. I have not used gnc_time64_get_day_(begin|neutral|end) as I was unsure if required. I did ponder on suggesting adding gnc_date_edit_get_date_(begin|neutral|end) with appropriate gnc_tm_set_day_neutral as an interim and going through occurrences of GNCDateEdit and choosing appropriate function but not sure if that would achieve any thing. Also should GNCDateEdit be GncDateEdit ?

jralls:
gnc_time64_get_day_(begin|neutral|end) isn't required, gnc_date_edit_get_date already returns the time64 for the beginning of the day as you know and gnc_date_edit_get_date_end returns the end of the day.

A gnc_date_edit_get_date_neutral might be useful for cases where we want to set neutral time on some object's member variable from a date edit. It wouldn't be useful for the query case.

Yes, GncDateEdit would better comply with our style guide.

gjanssens:
Committed, thanks! Note due to the newly discovered bug https://bugs.gnucash.org/show_bug.cgi?id=796839 I'm not able to do a user test, but the code looks good and compiles fine.

PR #400, opened 2018-08-22
Register Date bugs

Bob-IT:
These bugs were reported by Geert and hopefully fix the issues raised with the date warning and the removal of the gdate function.


Bob-IT:
Added another two fixes, first one fixes the book options dialogue not being raised and the second was to cache the read only number of days value.

Bob-IT:
Looks like I forgot to change the tests for the read only cache change, will sort tomorrow.

gjanssens:
Thanks!

PR #401, opened 2018-08-24
Ignore deprecation warnings from glib

vorlonofportland:
In glib 2.58, g_type_class_add_private is deprecated, which causes gnucash
to fail to build due to -Werror.  To get gnucash building in Ubuntu, we have
turned this particular error back into a warning with
-Wno-error=deprecated-declarations.

This also drops the -Wno-deprecated-register flag, which doesn't appear
to be recognized at all by gcc-8.

An alternate solution of course is to update gnucash to use the current glib API.

https://launchpad.net/ubuntu/+source/gnucash/1:3.2-1build1/+build/15137868

### DIFF
--- a/libgnucash/engine/CMakeLists.txt
+++ b/libgnucash/engine/CMakeLists.txt
@@ -235,7 +235,7 @@ add_dependencies (gncmod-engine swig-runtime-h iso-4217-c)
 target_link_libraries(gncmod-engine gnc-core-utils gnc-module ${Boost_DATE_TIME_LIBRARIES}  ${Boost_REGEX_LIBRARIES} ${REGEX_LDFLAGS} ${GMODULE_LDFLAGS} ${GLIB2_LDFLAGS} ${GOBJECT_LDFLAGS} ${GUILE_LDFLAGS})
 
 target_compile_definitions (gncmod-engine PRIVATE -DG_LOG_DOMAIN=\"gnc.engine\")
-target_compile_options (gncmod-engine PRIVATE -Wno-deprecated-register)
+target_compile_options (gncmod-engine PRIVATE -Wno-error=deprecated-declarations)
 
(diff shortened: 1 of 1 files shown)


gjanssens:
@jralls Do you have a preference for -Wno-error=deprecated-declarations vs -Wno-deprecated-declarations ? I'm inclined to set the former in both our default CXX_FLAGS and our C_FLAGS by default as explained in my earlier comment.

jralls:
I prefer -Wno-deprecated-declarations. I don't think we gain anything from warnings we're going to routinely ignore. When someone gets around to cleaning up the deprecated decls he can turn off the flag locally to find them all.

The change also removes -Wno-deprecated-register. IIRC it's there because one of the Boost libraries uses the ```register``` storage class in at least one header. If we're going to turn the error back on we need to find which one and in what Boost version it's fixed and update the minimum in the root CMakeLists.txt.

gjanssens:
Ah yes, the removal of -Wno-deprecated-register. I did some experiments and apparently gcc doesn't know the "Wdeprecated-register" warning. I have checked with gcc 7 (Fedora 27) and gcc 4.8 (Ubuntu 14.04).
Eventually I ran into a github page [listing all warnings known by different versions of gcc and clang](https://github.com/Barro/compiler-warnings). There I found Wdeprecated-register is a clang alias for Wregister. So I have changed our build config to use that one instead.

Note I had to add it to the global CMAKE_CXX_FLAGS instead of adding it via target_compile_options. It's a c++ only flag (at least in gcc) and our gncmod-engine target is still mixed C/C++.

In addition I have also added -Wno-deprecated-declarations to the CMAKE_CXX_FLAGS, as indicated by @vorlonofportland 
Hopefully this fixes the glib 2.58 issues you experienced on Ubuntu. Thanks for pointing us in the right direction!

PR #402, opened 2018-08-24
S390x build fixes

vorlonofportland:
Debugging a build failure of gnucash 3.2 on Ubuntu s390x, I identified a number of issues.  Here is a series of patches that address these issues and allow gnucash to build on this architecture.

### DIFF
--- a/common/config.h.cmake.in
+++ b/common/config.h.cmake.in
@@ -1,3 +1,5 @@
+#include <endian.h>
+
 /* Template file for processing by the cmake command CONFIGURE_FILE */
 
 /* If we're using this file it's a CMAKE build, so say so */
@@ -356,7 +358,11 @@
 # endif
 #else
 # ifndef WORDS_BIGENDIAN
+#  if __BYTE_ORDER == __BIG_ENDIAN
+#   define WORDS_BIGENDIAN 1
+#  else
 #cmakedefine WORDS_BIGENDIAN
+#  endif
 # endif
 #endif
 
--- a/gnucash/gnome-utils/dialog-utils.c
+++ b/gnucash/gnome-utils/dialog-utils.c
@@ -23,6 +23,9 @@
  *                                                                  *
 \********************************************************************/
 
+/* RTLD_DEFAULT */
+#define _GNU_SOURCE
+
 #include <config.h>
 
 #include <gtk/gtk.h>
--- a/gnucash/import-export/bi-import/dialog-bi-import-helper.c
+++ b/gnucash/import-export/bi-import/dialog-bi-import-helper.c
@@ -19,6 +19,8 @@
  * Boston, MA  02110-1301,  USA       gnu@gnu.org
  */
 
+/* strptime */
+#define _XOPEN_SOURCE
 #ifdef HAVE_CONFIG_H
 #include <config.h>
 #endif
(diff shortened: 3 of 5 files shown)


gjanssens:
Thank you for submitting these patches for s390x! This will bring us a step closer to getting gnucash 3 into debian again.

WRT to the big-endianness, I think WORDS_BIGENDIAN is not working because we simply don't evaluate test for endianness during a cmake run. We could fix this by using cmake's TEST_BIG_ENDIAN function.

I'll note the WORDS_BIGENDIAN seems to have been introduced in support of MacOS on PPC. IIUC that platform has been obsoleted in gnucash 3 so I think the MacOS specialization code can probably go out as well.

As I don't have access to bigendian hardware (nor have experience in simulating it) could you test this patch:
```
diff --git a/CMakeLists.txt b/CMakeLists.txt
index b2fcffb17..675584717 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -658,6 +658,11 @@ install(FILES ${gnucash_DOCS} DESTINATION ${CMAKE_INSTALL_DOCDIR})
 
 set (GETTEXT_PACKAGE "gnucash")
 
+test_big_endian (IS_BIGENDIAN)
+if (IS_BIGENDIAN)
+    set (WORDS_BIGENDIAN)
+endif (IS_BIGENDIAN)
+
 if (WIN32)
 if (MINGW)
 set (HAVE_SCANF_LLD 1)
```
And let me know if that also works instead of your proposed commit ?

gjanssens:
As for the other two fixes, I'm happy they solve the issues on s390x. However I think it would be nicer if these defines were included in config.h and only when needed. That keeps our code cleaner and will avoid regressions if the problematic functions get used in other files and the developer forgets to add the defines there.

The way to do this would require two changes:
1. in toplevel CMakeLists.txt one should check for the target architecture. If the architecture is s390x, set _XOPEN_SOURCE, _DEFAULT_SOURCE and _GNU_SOURCE variables.
1. in common/config.h.in, add "cmakeconfig" lines for each of the variables above.

I'll admit I'm not sure how to detect architecture in cmake. I found a trick involving the C preprocessor [on Stack Overflow](https://stackoverflow.com/questions/3877344/detect-endianness-with-cmake#3877387), but there may be better ways by now.

jralls:
WORDS_BIGENDIAN was introduced for MacOS-PPC, but the code in question is reading tzinfo files and so applies to all Unix, not just MacOS.

_GNU_SOURCE used to be defined by configure.ac, it just didn't make it over to CMake. With that globally defined _XOPEN_SOURCE [isn't needed any more](https://stackoverflow.com/questions/29201515/what-does-d-default-source-do). tm.tm_gmtoff doesn't appear in our code so there's no need for _DEFAULT_SOURCE at all.

All of which said we can probably get rid of strptime. Do we still need the dlsym workaround for BSDs?

jralls:
@vorlonofportland please see if in addition to @gjanssens's patch 
```cmake
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -545,6 +545,7 @@ check_c_compiler_flag(-Wstringop-truncation have_stringop_truncation)
 if (have_stringop_truncation)
     set(HAVE_STRINGOP_TRUNCATION TRUE)
 endif()
+add_definitions(-D_GNU_SOURCE)

 if (APPLE)
   include (CheckCXXCompilerFlag)
```
allows you to get a clean build.

jralls:
It seems @vorlonofportland was a drive-by. Since the changes are inconsequential everywhere else I've pushed them in the hope that it mostly fixes the S390x build.

PR #403, opened 2018-08-24
Use alignment-safe buffer handling

vorlonofportland:
Casting a char* to a struct containing a uint32_t is not universally safe
due to alignment constraints on reads on some platforms.  Copy our possibly
unaligned source data into an aligned area of memory to avoid SIGBUS on
armhf.


gjanssens:
I have committed John's optimized version for this issue. If this doesn't work for some reason, please let us know.
Thanks for pointing out the issue and creating a initial fix!

PR #404, opened 2018-08-26
Bug787401 - Tests for jqplot

GnuFiBux:
Hi there, this is last step to transform the tests from Bug787401 into pull request.

Here are some test for jqplog related functions for report generation.
Mainly simple positive tests, only a few simple negative tests.

You find some hints for potential bugs in there which I will take care of next, for now I ignore the tests that might fail due to these potential bugs.


jralls:
You need to get this to pass on both Travis platforms...

GnuFiBux:
if my understanding is correct (see below) then .... LOL

In PR#379 you commented
[...] Since the tests all pass I will be interested to see your bug reports [...]

Now I let the tests which discover bugs fail, and it breaks the travis check for ninja because the tests fail.
The Ubuntu part only passes because the tests do not get executed.

I can comment out again all tests that fail, to make the ninja part work, and make them active again once I have found fixes (through seperate PRs) - not only here but also for PR#391.
Which opens the principle question what happens to deactivated tests for which the underlying faults are fixed by persons that do not know about these tests. But that is not me to answer.

Regarding the Ubuntu build, I have no clue how I can help to get the tests executed.

jralls:
Since you haven't followed my instructions about how to write a SRFI-64 test it's impossible to tell from the output what is the problem... including knowing whether there's actually a bug in the gnc-html code or if it's in your tests.  Yes, that's a defect in most of the Scheme test code, but fixing it was one of the main goals of adopting a proper unit test framework.

If your tests did find bugs then the PR needs to include fixes as separate commits. We're not merging failing tests and we're not merging non-tests. Consider that both of the failing tests in this PR seem not to matter: All of the existing report tests pass and the line charts work in use. If an important jqplot module isn't present or if the line chart plotting code was actually defective on Arch Linux that wouldn't be the case.

GnuFiBux:
The adaptions for SRFI-64 have been done now, and I have added to solution proposals so the test data setup can be kept to the essentials.

Still the travis ninja build brakes because of two changes in the return value when rendering the test line chart:

- other date format used then on my mashine
expected: options.axes.xaxis.tickOptions.formatString = '%m/%d/%y'
obtained: options.axes.xaxis.tickOptions.formatString = '%d.%m.%Y'

- compile data scrambled with the return value
expected: var hasVScroll = document.body.scrollHeight > document.body.clientHeight;
obtained: var hasVScroll = document.body.scrollHeight > document.body.clien;;; note: auto-compilation is enabled, set GUILE_AUTO_COMPILE=0
;;;       or pass the --no-auto-compile argument to disable.
;;; compiling /gnucash/gnucash/report/report-system/test/test-report-jqplot.scm
;;; compiled /root/.cache/guile/ccache/2.2-LE-8-3.A/gnucash/gnucash/report/report-system/test/test-report-jqplot.scm.go
tHeight;

So for me it looks like the failure is not a code error but caused by not alinged test setup.
Any help is appreciated how to get the test setups (travis <-> my local) aligned.

gjanssens:
Op woensdag 19 september 2018 18:17:39 CEST schreef GnuFiBux:
> The adaptions for SRFI-64 have been done now, and I have added to solution
> proposals so the test data setup can be kept to the essentials.
> 
> Still the travis ninja build brakes because of two changes in the return
> value when rendering the test line chart:
> 
> - other date format used then on my mashine
> expected: options.axes.xaxis.tickOptions.formatString = '%m/%d/%y'
> obtained: options.axes.xaxis.tickOptions.formatString = '%d.%m.%Y'
> 
This can probably be fixed by forcing the C locale early in your test script.

> - compile data scrambled with the return value
> expected: var hasVScroll = document.body.scrollHeight >
> document.body.clientHeight; obtained: var hasVScroll =
> document.body.scrollHeight > document.body.clien;;; note: auto-compilation
> is enabled, set GUILE_AUTO_COMPILE=0 ;;;       or pass the
> --no-auto-compile argument to disable.
> ;;; compiling
> /gnucash/gnucash/report/report-system/test/test-report-jqplot.scm ;;;
> compiled
> /root/.cache/guile/ccache/2.2-LE-8-3.A/gnucash/gnucash/report/report-system
> /test/test-report-jqplot.scm.go tHeight;

If you properly added the test in the build system, this file should have been 
precompiled before running it and hence the warning should not appear. You can 
compare with other scm test examples to learn how they are added in the 
CMakeLists.txt files to ensure they are precompiled.

GnuFiBux:
Thanks for the quick hint.
Forcing the C locale does the trick.

gnucash/report/report-system/test/test-report-jqplot.scm
```diff
@@ -0,0 +1,699 @@
...
+
+(define (test-bar-chart-data)
+
+  (test-begin "Jqplot - HTML Bar Chart Data")
+
+  (let (
+         (html-bar-chart (gnc:make-html-barchart))
+       )
+    ;; prepare test data
+    (gnc:html-barchart-set-data! html-bar-chart '((5.00)))
```
jralls: You were doing well until now. This tests 4 separate function at one go and on failure requires careful analysis to figure out what's gone wrong.

Instead, populate a chart with a canned set of elements as a fixture and then run each function in its own test to see that it does what it's supposed to,.

gnucash/report/report-system/test/test-report-jqplot.scm
```diff
@@ -0,0 +1,699 @@
...
+    (gnc:html-barchart-set-x-axis-label! html-bar-chart "Test x-Axis Label barchart")
+    (gnc:html-barchart-set-y-axis-label! html-bar-chart "Test y-Axis Label barchart")
+    (gnc:html-barchart-set-col-labels! html-bar-chart '("Col1" "Col2" "Col3"))
+    (gnc:html-barchart-set-row-labels! html-bar-chart '("Row1" "Row2" "Row3"))
+    (gnc:html-barchart-set-data! html-bar-chart '((5.00)))
+
+    ;; do the test
+    (test-equal "Jqplot - Render Bar Chart"
+      (string-concatenate (drop (gnc:html-document-tree-collapse (gnc:html-barchart-render html-bar-chart #f)) 11))
+"\" style=\"width:800px;height:400px;\"></div>\n\
```
jralls: This is terribly brittle and any errors will be difficult to diagnose. That's partly due to gnc:html-foo-render doing too much at once, which makes it impossible to test the different points in its execution. Since Chris Lam is working on replacing these functions (and indeed replacing jqplot) entirely it's not worth the effort to refactor them into something more testable. I suggest that you just remove the tests.

GnuFiBux: Agree.
I was putting it here as a regression test that is only supposed to act as indicator for "something has changed". The idea was: To have an indication is better than not having any indication. For diagnosis this will not help, but at least you know that there might be something to analyze.
I do not have a strong opinion here. But to keep the possibility for easy regression, I would suggest to only comment this test out, so there is a quick possibility to enable it again "offline" if deemed useful for whatever reason. I can also decorate it with corresponding comments so other contributors/developers are aware of that idea.
Is that ok?

gnucash/report/report-system/test/test-report-jqplot.scm
GnuFiBux: Agree.
The return values right now are giving good indications in which direction to look if you are familiar with how the table creation works. If not, than your proposal will make it clearer, so let's go for that and do it similar as for the table data tests in PR#391.

This brings me to a another point:
As the table data manipulation is handled similarily in html-barchart.scm, html-linechart.scm and html-table.scm, should that be centralized? E.g. html-barchat.scm and html-linechart.scm are using the functions from html-table.scm?

Is Chris Lam maybe working on this already?
What I am after: The corresponding tests for html-table.scm came along with a bugfix for row extension. I think it will be the same for bar and line chart. If Chris is working on a new concept than I will drop corresponding tests until this is done, and only keep the simple tests as implemented so far.

gnucash/report/report-system/test/test-report-jqplot.scm
christopherlam: It's best to remove them completely. It's just clutter and will really annoy anyone trying to develop charts.

gnucash/report/report-system/test/test-report-jqplot.scm
christopherlam: #316 is the current work-in-progress for new chart generation, which eventually aims to replace html-pie/line/bar/scatter. I would definitely NOT write any whole-document string-comparison tests. I'd write a test for each functionality, with good comments what's being tested.

Compare the following test output:
````
[pass] html-chart-add-row! on an empty table
[pass] html-chart-add-row! on a single row-table
[fail] html-chart-add-row! on a large table
 expected: xxx
 obtained: yyy
[pass] html-chart-remove row
````

is 1000% more informative than
````
[fail] html-chart row manipulation tests
````

GnuFiBux: thanks for the quick response and your clear words.
I understand
- having nothing is better than having anything until it is replaced by something proper
- you have not followed my recent changes by which 999% of "my" tests should return expected/obtained values, have a look at #391 - maybe you noticed that I agreed to John's comment
- this PR is a waste of time, it will anyway be replaced by Chris' work, thanks for letting me know

times appear not to be easy for voluntary contributors, actually, seems like you do not want them

gnucash/report/report-system/test/test-report-jqplot.scm
GnuFiBux: ah, come on,  let's close the issue

christopherlam:
Hi apologies for the impression, I did not mean to be rude.

As you can see in the upcoming merge requests, I have not made changes because frankly I do not 100% understand the *existing* meaning of the various html-chart modifiers. I can see many cases whereby the meaning != actual code, and I've refrained from making gratuitous changes which match my understanding, because buggy implementations usually are accompanied by workarounds by other coders who can't/won't fix the implementations.

So while I agree the various -add-row! or -add-column! are not completely bug-free, I've copied them "as is" into html-chart which calls chartJS without any personal changes I wouldn't claim to know the exact intention of the original coder; ultimately the idea is  there will be a migration path from html-bar/line/scatter/piechart (plugs to jqplot) into html-chart (plugs to chartjs), and I do not wish to annoy any report developer who expects the 4 chart types to behave in a particular way. However I do think your changes to -add-row! and -add-column! fix the implementation safely so I'd welcome your views on #316 html-chart.scm which is as yet not final nor mergeable... once merged the API should be considered "frozen for all future purposes" and I'm finding this is quite difficult.

Moreover the javascript/html output of html-chart is currently not frozen yet. Therefore testing the html-document output against a frozen implementation will fail every time a change is made in html-chart rendering code, and if e.g. a feature request is merged in, the developer would need to create suitable tests for the new feature, code it, verify its output, *and* amend the html-document render test in the string test. This is needless duplication of work!

PR #406, opened 2018-08-31
scheme-progress: some tests for report-utilities.scm

christopherlam:
Majority coverage of report-utilities.scm functions

Does not include:
- budget-related functions - too annoying!
- gnc:account-get-pos-trans-total-interval - don't understand
- gnc:account-get-trans-type-splits-interval - don't understand

@jralls - I can't understand (teardown) - if we call it prior to calling (test-get-account-balances) in the main test-runner, it resets all GBP symbol from "£" to "RON" causing the subsequent tests to fail unfairly. Is there another call to 'start new session'?

### DIFF
--- a/gnucash/report/report-system/test/test-report-utilities.scm
+++ b/gnucash/report/report-system/test/test-report-utilities.scm
@@ -188,7 +188,7 @@
    (gnc-commodity-get-namespace (gnc-default-report-currency))
    sym))
 
-(define structure
+(define (structure)
   (list "Root" (list (cons 'type ACCT-TYPE-ASSET))
         (list "Asset"
               (list "Bank")
@@ -205,7 +205,7 @@
 
 (define (create-test-data)
   (let* ((env (create-test-env))
-         (account-alist (env-create-account-structure-alist env structure))
+         (account-alist (env-create-account-structure-alist env (structure)))
          (asset (cdr (assoc "Asset" account-alist)))
          (bank (cdr (assoc "Bank" account-alist)))
          (gbp-bank (cdr (assoc "GBP Bank" account-alist)))
--- a/gnucash/report/report-system/test/test-report-utilities.scm
+++ b/gnucash/report/report-system/test/test-report-utilities.scm
@@ -31,6 +31,7 @@
   (map gnc:monetary->string (coll 'format gnc:make-gnc-monetary #f)))
 
 (define (test-account-get-trans-type-splits-interval)
+  (test-group-with-cleanup "test-account-get-trans-type-splits-interval"
   (let* ((env (create-test-env))
          (ts-now (gnc-localtime (current-time)))
          (test-day (tm:mday ts-now))
@@ -57,7 +58,8 @@
 	;; 10 is the right number (5 days, two splits per tx)
 	(test-equal "length splits = 10"
           10
-          (length splits))))))
+          (length splits)))))
+  (teardown)))
 
 (define (teardown)
   (gnc-clear-current-session))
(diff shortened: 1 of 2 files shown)


jralls:
@christopherlam, you don't call teardown from the main test runner. You call it as the last item in ```(test-group-with-cleanup...```, e.g. https://github.com/Gnucash/gnucash/blob/maint/gnucash/report/report-system/test/test-commodity-utils.scm#L385. ```setup``` is called as the first ```let*``` item:
```scheme
(define (test-get-exchange-totals)
  (test-group-with-cleanup "gnc:get-exchange-totals"
  (let* ((account-alist (setup #f))
```

That encapsulates each test function in its own session so there's no shared state between test functions.

christopherlam:
Hmm I did try.

I tried encapsulating the old (test-account-get-trans-type-splits-interval) with the `(test-group-with-cleanup "old-test" (let* (...) ...) (teardown))` however the very act of running (gnc-clear-current-session) seems to make GBP symbol change. It doesn't detract from this PR however; the old (test-account-get-trans-type-splits-interval) uses modern dates whereas my other tests use 1970-2000 dates, so, they'll never clash.

jralls:
Well, push your attempt and I'll take a look at it. Bear in mind that I'll be AFK for the weekend.

christopherlam:
My output after ec0e9a6 - [PR406.txt](https://github.com/Gnucash/gnucash/files/2340592/PR406.txt)

jralls:
The problem isn't teardown _per se_, it's that you haven't created anything to teardown. ```create-test-env``` just seeds the random number generator, it doesn't create a session or a book.

christopherlam:
I've traced this GBP symbol issue to something about the print-info data structure then got lost in C. The (gnc-commodity-get-nice-symbol GBP) output is "£" as it should. In any case I'd prefer to not get too lost in C, and skip the last commit here. i.e. #398 could be deemed mergeable because it incorporates this PR excluding the last commit.

jralls:
You've encountered that problem before. At bottom it's that Guile 2.x's charset handling is pretty badly broken. £ is codepoint U+00A3 encoded as 0xC2 0xA3 in UTF-8. How that's getting converted to "RON" is pretty mystifying. My solution is to avoid currencies and their non-ascii symbols.

christopherlam:
> You've encountered that problem before. At bottom it's that Guile 2.x's charset handling is pretty badly broken. £ is codepoint U+00A3 encoded as 0xC2 0xA3 in UTF-8. How that's getting converted to "RON" is pretty mystifying. My solution is to avoid currencies and their non-ascii symbols.

I still do not think that guile's charset is at blame. I tried changing my testing from GBP ("£" gets changed to "RON") to other ASCII and non-ASCII currencies.
"HK$" -> "CN¥"
"CLP" -> "SDP"
"CNY" -> "JP¥"
"DEM" -> "GHC"
"EUR" -> "$n"

lol... I'll try to isolate this bug...

gnucash/report/report-system/test/test-report-utilities.scm
```diff
@@ -41,4 +56,388 @@
...
+      lst)
+    (list-set-safe! lst 5 1)
+    (test-equal "list-set-safe out-of-bounds"
+      '(1 3 #f #f #f 1)
+      lst))
+  (test-end "list-set-safe"))
+
+(define (test-gnc:monetary->string)
+  (test-begin "gnc:monetary->string")
+  (let* ((book (gnc-get-current-book))
```
jralls: Needs a teardown to free the session created by gnc-get-current-book.

christopherlam:
done!

christopherlam:
well, if you're happy with both, we could merge #406 first then merge #398, or I can rebase #398 over #406 and we can merge #398 and close #406.

jralls:
I opted for the first choice. Thanks!

PR #407, opened 2018-09-02
Bug796839

Bob-IT:
The first commit I think fixes Bug796839, in my testing it does what I think it should but as always another take on it may find some thing I missed.

The second commit updates a previous commit to reduce the number of times the display ledger is refreshed at start, I had missed the setting of double line mode and register style. As these are set via callbacks I have changed to setting/resetting the register parameter 'enable_refresh' to control the enabling/disabling of the display ledger refresh, this is now potentially reducing it from 7 to 2.

### DIFF
--- a/gnucash/gnome/gnc-plugin-page-register.c
+++ b/gnucash/gnome/gnc-plugin-page-register.c
@@ -139,8 +139,8 @@ static gchar *gnc_plugin_page_register_get_filter (GncPluginPage *plugin_page);
 void gnc_plugin_page_register_set_filter (GncPluginPage *plugin_page, const gchar *filter);
 static void gnc_plugin_page_register_set_filter_tooltip (GncPluginPageRegister *page);
 
-static void gnc_ppr_update_status_query (GncPluginPageRegister *page, gboolean refresh);
-static void gnc_ppr_update_date_query (GncPluginPageRegister *page, gboolean refresh);
+static void gnc_ppr_update_status_query (GncPluginPageRegister *page);
+static void gnc_ppr_update_date_query (GncPluginPageRegister *page);
 
 /* Command callbacks */
 static void gnc_plugin_page_register_cmd_print_check (GtkAction *action, GncPluginPageRegister *plugin_page);
@@ -548,6 +548,7 @@ typedef struct GncPluginPageRegisterPrivate
 
     gint lines_default;
     gboolean read_only;
+    gboolean enable_refresh; // used to reduce ledger display refreshes
     Query *search_query;     // saved search query for comparison
     Query *filter_query;     // saved filter query for comparison
 
@@ -808,6 +809,7 @@ gnc_plugin_page_register_init (GncPluginPageRegister *plugin_page)
     priv->read_only         = FALSE;
     priv->fd.cleared_match  = CLEARED_ALL;
     priv->fd.days           = 0;
+    priv->enable_refresh    = TRUE;
     priv->search_query      = NULL;
     priv->filter_query      = NULL;
 }
@@ -1119,6 +1121,7 @@ gnc_plugin_page_register_create_widget (GncPluginPage *plugin_page)
     gchar **filter;
     gchar *order;
     int filter_changed = 0;
+    gboolean create_new_page = FALSE;
 
     ENTER("page %p", plugin_page);
     page = GNC_PLUGIN_PAGE_REGISTER (plugin_page);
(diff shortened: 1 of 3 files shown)


gnucash/gnome/gnc-plugin-page-register.c
```diff
@@ -2417,6 +2466,13 @@ gnc_ppr_update_status_query (GncPluginPageRegister *page, gboolean refresh)
     gnc_plugin_page_register_set_filter_tooltip (page);
 
     if (refresh)
+    // clear previous filter query
```
jralls: Oops. ```if (refresh)``` needs to be *after* this block. That confused me when I looked at the next commit.

gnucash/gnome/gnc-plugin-page-register.c
```diff
@@ -2489,6 +2551,13 @@ gnc_ppr_update_date_query (GncPluginPageRegister *page, gboolean refresh)
     gnc_plugin_page_register_set_filter_tooltip (page);
 
     if (refresh)
```
jralls: As above, the inserted block should go before ```if (refresh)```.

gnucash/gnome/gnc-plugin-page-register.c
```diff
@@ -1181,7 +1192,7 @@ gnc_plugin_page_register_create_widget (GncPluginPage *plugin_page)
         g_free(order);
 
         priv->sd.reverse_order = gnc_plugin_page_register_get_sort_reversed(plugin_page);
-        gnc_split_reg_set_sort_reversed(priv->gsr, priv->sd.reverse_order, FALSE);
+        gnc_split_reg_set_sort_reversed(priv->gsr, priv->sd.reverse_order, priv->enable_refresh);
```
jralls: priv->enable_refresh is always FALSE here, so it's confusing to change it.

gnucash/gnome/gnc-plugin-page-register.c
```diff
@@ -1264,11 +1275,15 @@ gnc_plugin_page_register_create_widget (GncPluginPage *plugin_page)
...
+    gnc_ppr_update_date_query (page);
 
-    /* Now do the refresh */
-    gnc_ledger_display_refresh(priv->ledger);
+    /* Now do the refresh if this is a new page instaed of restore */
+    if (create_new_page)
+    {
+        priv->enable_refresh = TRUE;
+        gnc_ledger_display_refresh(priv->ledger);
+    }
```
jralls: Instead, move the ```if (priv->enable_refresh)``` block from above down here:
```C
/* We always want to call gnc_ppr_update_status_query and gnc_ppr_update_date_query with refresh disabled here */
     if (priv->enable_refresh)
     {
           create_new_page = TRUE;
           priv->enable_refresh = FALSE;
      }
      gnc_ppr_update_status_query (page);
      gnc_ppr_update_date_query (page);
      if (create_new_page)
      {
           priv->enable_refresh = TRUE;
           gnc_ledger_display_refresh(priv->ledger);
      }
```

That keeps the machination all in one place so it's clearer what's going on.

jralls: However, ISTM it would be be a better design to not have gnc_ledger_display_refresh() as a side effect of gnc_ppr_update_status_query, gnc_ppr_update_date_query, gnc_split_reg_change_style, and gnc_split_reg_double_line and to instead call it explicitly when it's the right thing to do and not otherwise.

gnucash/gnome/gnc-plugin-page-register.c
```diff
@@ -1359,6 +1363,12 @@ gnc_plugin_page_register_destroy_widget (GncPluginPage *plugin_page)
         memset(&priv->fd, 0, sizeof(priv->fd));
     }
 
+    if (priv->search_query)
```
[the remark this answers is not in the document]
Bob-IT: OK, did not realise so have removed tests

gnucash/gnome/gnc-plugin-page-register.c
```diff
@@ -2373,6 +2383,34 @@ gnc_plugin_page_register_sort_order_reverse_cb (GtkToggleButton *button,
...
+
+        // if filter_query is NULL, then the dialogue find has been run
+        // before coming here. if query_tmp does not equal filter_query
+        // then the dialogue find has been run again before coming here
+        if ((priv->filter_query == NULL) ||
+           (!qof_query_equal (query_tmp, priv->filter_query)))
+        {
+            if (priv->search_query != NULL)
+                qof_query_destroy (priv->search_query);
+            priv->search_query = qof_query_copy (query_tmp);
```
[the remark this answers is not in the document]
Bob-IT: The query_tmp is only a pointer to the ledger->query so I needed to save a copy of it to priv->search_query. When the ledger is updated with gnc_ledger_display_set_query, the original query is destroyed and a copy of the provided one is made.

gnucash/gnome/gnc-plugin-page-register.c
Bob-IT: This was a mistake in trying to separate work done into the two commits.

gnucash/gnome/gnc-plugin-page-register.c
Bob-IT: I think I got carried away with the find and replace changes.

gnucash/gnome/gnc-plugin-page-register.c
Bob-IT: Moved as suggested.

Bob-IT:
I have made the changes and all still works as I expect.
Regarding the last comment, in master I will need to look in this area to change the saving/loading of the filter/sort settings to .gcm file only and was intending to separate out all the filter/sort bits to a new file along with the conversion routine so will have a look to see if possible then.

gnucash/gnome/gnc-plugin-page-register.c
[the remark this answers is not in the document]
jralls: Ah, OK.

PR #409, opened 2018-09-07
Block registered prefs when preference dialogue loaded

Bob-IT:
When the preference dialogue is loaded and options are set, the ones with registered callbacks fire causing parts of Gnucash to be updated. This was observed with gnc_split_register_load being executed 5 times for each open register when the preference dialogue was loaded.

To overcome this, a couple of functions have been created to block and unblock all registered prefs and used while the preference dialogue is loaded. I checked to see if in the source any had the capability of being blocked but did not find any so I think it would be safe to use in the situation indicated.

There is one thing I want to change but could not work out how, that is the two gsetting functions are defined with a text parameter but should be void but when I made what I thought were the appropriate changes, it compiled OK but would they not execute.

So hopefully the idea is sound but await feedback.

### DIFF
--- a/libgnucash/app-utils/gnc-gsettings.c
+++ b/libgnucash/app-utils/gnc-gsettings.c
@@ -53,6 +53,8 @@ static GHashTable *schema_hash = NULL;
 static const gchar *gsettings_prefix;
 static xmlExternalEntityLoader defaultEntityLoader = NULL;
 
+static GHashTable *registered_handlers_hash = NULL;
+
 /* This static indicates the debugging module that this .o belongs to.  */
 static QofLogModule log_module = "gnc.app-utils.gsettings";
 
@@ -131,6 +133,19 @@ static GSettings * gnc_gsettings_get_settings_ptr (const gchar *schema_str)
     return gset;
 }
 
+static void
+handlers_hash_block_helper (gpointer key, gpointer settings_ptr, gpointer pointer)
+{
+    g_signal_handler_block (settings_ptr, (gulong)key); // block signal_handler
+    PINFO("Block handler_id %ld for settings_ptr %p", (gulong)key, settings_ptr);
+}
+
+static void
+handlers_hash_unblock_helper (gpointer key, gpointer settings_ptr, gpointer pointer)
+{
+    g_signal_handler_unblock (settings_ptr, (gulong)key); // unblock signal_handler
+    PINFO("UnBlock handler_id %ld for settings_ptr %p", (gulong)key, settings_ptr);
+}
 
 /************************************************************/
 /*                      GSettings Utilities                 */
@@ -203,6 +218,17 @@ gnc_gsettings_register_cb (const gchar *schema,
 
     retval = g_signal_connect (settings_ptr, signal, G_CALLBACK (func), user_data);
 
+    if (!registered_handlers_hash)
+        registered_handlers_hash = g_hash_table_new (g_direct_hash, g_direct_equal);
+
+    if (retval != 0)
+    {
+        g_hash_table_insert (registered_handlers_hash,
+                             GINT_TO_POINTER(retval), settings_ptr); //key, value
+
+        PINFO("schema: %s, key: %s, settings_ptr: %p, handler_id: %ld",
+               schema, key, settings_ptr, retval);
+    }
     g_free (signal);
 
     LEAVE("");
@@ -218,6 +244,7 @@ gnc_gsettings_remove_cb_by_func (const gchar *schema,
 {
     gint matched = 0;
     GQuark quark = 0;
+    gulong handler_id = 0;
 
     GSettings *settings_ptr = gnc_gsettings_get_settings_ptr (schema);
     g_return_if_fail (G_IS_SETTINGS (settings_ptr));
@@ -228,7 +255,7 @@ gnc_gsettings_remove_cb_by_func (const gchar *schema,
     if ((key) && (gnc_gsettings_is_valid_key(settings_ptr, key)))
         quark = g_quark_from_string (key);
 
-    matched = g_signal_handlers_disconnect_matched (
+    handler_id = g_signal_handler_find (
                   settings_ptr,
                   G_SIGNAL_MATCH_DETAIL | G_SIGNAL_MATCH_FUNC | G_SIGNAL_MATCH_DATA,
                   g_signal_lookup ("changed", G_TYPE_SETTINGS), /* signal_id */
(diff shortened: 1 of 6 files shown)


libgnucash/app-utils/gnc-gsettings.c
```diff
@@ -850,3 +868,51 @@ void gnc_gsettings_version_upgrade (void)
...
+            remove_copy = g_slist_prepend (remove_copy, ri); // add for removal
+    }
+
+    if (g_slist_length (remove_copy) > 0)
+    {
+        PINFO("remove_copy length is %d", g_slist_length (remove_copy));
+        for (node = remove_copy; node; node = node->next)
+        {
+            RegisteredSignalInfo *ri = node->data;
+            GSList *found_item = g_slist_find (registered_signals, ri);
```
gjanssens: This is fairly inefficient and performance will decrease exponentially with every settings callback we register. I suspect it's not needed though, more on that in the other commits.

libgnucash/app-utils/gnc-gsettings.c
```diff
@@ -53,6 +53,14 @@ static GHashTable *schema_hash = NULL;
 static const gchar *gsettings_prefix;
 static xmlExternalEntityLoader defaultEntityLoader = NULL;
 
+typedef struct
+{
+    gulong signal_handler;
+    gpointer instance;
+} RegisteredSignalInfo;
+
+static GSList *registered_signals = NULL;
```
gjanssens: Slightly pedantic: you're not keeping a list of signals but a list of handlers.

libgnucash/app-utils/gnc-gsettings.c
```diff
@@ -850,3 +868,51 @@ void gnc_gsettings_version_upgrade (void)
...
+    if (g_slist_length (remove_copy) > 0)
+    {
+        PINFO("remove_copy length is %d", g_slist_length (remove_copy));
+        for (node = remove_copy; node; node = node->next)
+        {
+            RegisteredSignalInfo *ri = node->data;
+            GSList *found_item = g_slist_find (registered_signals, ri);
+
+            if (found_item)
+                registered_signals = g_slist_remove (registered_signals, ri);
```
gjanssens: Why are you removing callbacks here ?

My guess is because you didn't adjust the gnc_gsettings_remove_by_xyz functions to update the registered_signals list opposite to how you update it during gnc_gsettings_register_callback.

libgnucash/app-utils/gnc-gsettings.h
```diff
@@ -86,6 +86,16 @@ void gnc_gsettings_set_prefix (const gchar *prefix);
 const gchar *gnc_gsettings_get_prefix (void);
 
 
+/** Block all prefs callbacks, used while preference dialog is loaded.
+ */
+void gnc_gsettings_block_all (const gchar *text);
```
gjanssens: Can you give more details of how this was implemented with void ? You can also create a temporary commit with all the changes if you like (even if it doesn't work). That may be easier to figure out what is wrong.

Bob-IT:
I have renamed the list and removed the g_slist_find and used g_slist_nth_data as I already know at what position in the registered_handlers I need to remove. Also worked out why settings the functions with void did not work, when used in gnc_prefs, I missed off the empty brackets like..
(prefsbackend->block_all) ();

I am still removing the old entries from the registered_handlers list in the blocking function for the reason above and if I moved that to the '...remove_cb...' function I would then have to use slist find which as you say is inefficient.

gjanssens:
Thanks for the update.

It's a bit unfortunate I focused on performance when pointing out the odd construction of removing obsolete handlers while blocking them. My main point is really this is not very good design. A callback blocking function should do just that: block callbacks. I was puzzled the first time I saw your PR and it took me a couple of minutes digging through the rest of the code why you would need to remove callbacks at that point.

In short: the logical flow of the code would be much clearer if you add callbacks in the gnc_gsettings_register_cb function and remove them again in the gnc_gsettings_remove_by_* functions. When a coder encounters the first, s/he will automatically understand the second.

Add to that the current construct is a potential memory leak. What if handlers get added, the block function is not used afterwards, yet the handler gets removed again using gnc_gsettings_remove_* ? That means the handler will no longer be removed from the list. I'm not saying the current code does this, but it may do so at some point as the result of future changes. In the current construct this isn't too obvious to detect.

As for performance: I agree g_slist_nth_data is probably faster than g_slist_find. On the other hand the remove_by_* functions will usually not be called on all handlers at once, so their impact per call is relatively small. And perhaps a hash table may be even better in this case, using the handler ID as key and the instance as value. That would probably be faster for the random access needed in gnc_gsettings_remove_* and with a g_hash_for_each it would be a very concise and efficient handler blocking function. Note my use of "may" and "would" as only profiling would prove it :)

So please do move the handler removal to gnc_gsettings_remove_*, and check if you can implement this using a hash table.

libgnucash/app-utils/gnc-gsettings.c
```diff
@@ -850,3 +918,25 @@ void gnc_gsettings_version_upgrade (void)
...
         gnc_gsettings_set_int (GNC_PREFS_GROUP_GENERAL, GNC_PREF_VERSION, cur_maj_min);
 }
+
+
+void gnc_gsettings_block_all (void)
+{
+    PINFO("registered_handlers_hash block list size before is %d",
+           g_hash_table_size (registered_handlers_hash));
+    g_hash_table_foreach_remove (registered_handlers_hash,
+                                 handlers_hash_helper, GINT_TO_POINTER(BLOCK));
```
gjanssens: As I said in earlier comments it doesn't make sense to combine blocking signals with a hash table cleanup (that is removing stale handler ids). So g_hash_table_foreach_remove is out of place here.
Instead you can rewrite the helper function to block or unblock the passed in handlerid based on a boolean passed in (instead of the enum). In fact as the hash helper function is so minimal, I would write two: one for blocking a signal handler and one for unblocking it. And then don't pass anything as user data. But that's a free choice.

If you worry this helper function could ever get called on a handler that's no longer connected you can add a test in the helper first which emits a warning and returns if you run into this situation. I expect the code to never hit it though.

libgnucash/app-utils/gnc-gsettings.c
```diff
@@ -850,3 +918,25 @@ void gnc_gsettings_version_upgrade (void)
...
+           g_hash_table_size (registered_handlers_hash));
+}
+
+
+void gnc_gsettings_unblock_all (void)
+{
+    PINFO("registered_handlers_hash unblock list size before is %d",
+           g_hash_table_size (registered_handlers_hash));
+    g_hash_table_foreach_remove (registered_handlers_hash,
+                                 handlers_hash_helper, GINT_TO_POINTER(UNBLOCK));
```
gjanssens: Same remarks as for gnc_gsettings_block_all.

gjanssens:
Thanks. This is now very clean and readable.

I have chosen the merge this already and look forward to a followup PR to handle proper cleanup of the lingering signal handlers.

Coming back to the leak of the hash table structure itself (in addition to the lingering signal handlers), it occurred to me we can also  handle that elegantly, as the code is now quite symmetric: there is one single function that adds elements to the hash table (`gnc_gsettings_register_cb`) and one single function that removes them (`gnc_gsettings_remove_cb_by_id`). As you create the initial hash table in that first function if it doesn't exist yet, you can equally destroy and nullify it in the latter if it has no elements left.

If you like you can add this while fixing the lingering 35 signal handlers.

PR #408, opened 2018-09-06
Commit

c-holtermann:
Based on discussion on bugtracker: https://bugs.gnucash.org/show_bug.cgi?id=796137

Fix problems with qof_query_search_for in python bindings resulting from switch to Python 3 and different string handling.

PR #410, opened 2018-09-07
 Fix string handling in qof_query_search_for to work with python3 

c-holtermann:
Fix problems with qof_query_search_for in python bindings resulting from switch to Python 3 and different string handling.

1) Provides typemaps for QofIdType and QofIdTypeConst (patch by David Osguthorpe)
2) modification to python Query class so that search_for is a python method that creates a buffer string that stays alive as long as the query lives and calls qof_query_search_for. The latter is being imported as python Query object method _search_for.
3) Prevent Query.search_for to be overwritten by add_constructor_and_methods_with_prefix after declaration of method search_for as described above. To get that an exclude option is being added to add_constructor_and_methods_with_prefix.

Based on discussion on bugtracker: https://bugs.gnucash.org/show_bug.cgi?id=796137

This is a cleanup of pull request https://github.com/Gnucash/gnucash/pull/408.

### DIFF
--- a/bindings/python/function_class.py
+++ b/bindings/python/function_class.py
@@ -125,19 +125,23 @@ class ClassFromFunctions(object):
         return method_function
 
     @classmethod
-    def add_methods_with_prefix(cls, prefix):
-        """Add a group of functions with the same prefix
+    def add_methods_with_prefix(cls, prefix, exclude=[]):
+        """Add a group of functions with the same prefix, exclude methods
+        in array exclude.
         """
         for function_name, function_value, after_prefix in \
-            extract_attributes_with_prefix(cls._module, prefix):
+                extract_attributes_with_prefix(cls._module, prefix):
+
+            if not (function_name in exclude):
                 cls.add_method(function_name, after_prefix)
 
     @classmethod
-    def add_constructor_and_methods_with_prefix(cls, prefix, constructor):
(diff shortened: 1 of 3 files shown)


jralls:
That's better, but rather than a catch-all pull requests branch you really want a separate feature branch for each pull request. Always start that from an up-to-date tracking branch of the appropriate canonical repository branch, e.g. https://github.com/Gnucash/gnucash maint. You can maintain a separate branch or set of branches that you merge the canonical branches into if you want, but don't use those for generating pull requests.

jralls:
Looks good, just one more thing: Please change the last commit to have "
Bug 796137 - query.search_for outputs critical qof.object errors and prevents queries being run" as the commit title or with "fixes" in the message so that I remember to include that the bug is fixed in the release notes.

c-holtermann:
Thank you John for the explanation of pull request branches. I'd create one for developing every feature and another one to cherry-pick the relevant commits for pull-requests. I would have a lot of branches in the end which seems inconvenient. I see that archiving branches by making them a tag seems a possibility.

c-holtermann:
I see that some of your suggestions for using github are included in the wiki https://wiki.gnucash.org/wiki/Git#Using_the_Github_Repository. I'll have a closer look at that and maybe add if necessary what I just learned :)

jralls:
Thnaks!

PR #411, opened 2018-09-14
Optimize interval charts

christopherlam:
These commits should partially undo https://github.com/Gnucash/gnucash/commit/8aed5c3f660951bbde9d01fdba62c2915cd42651 and reoptimize the net-worth and income-chart calculations using xaccAccountGetBalance API instead of qof-query splitlist cycling.

1. a few more tests for test-charts
2. test-charts output table is tested for net-charts.
3. net-charts is cleaned up to use old (process-datelist) which uses the slow (gnc:account-get-comm-balance-at-date)...
4. ... which is immediately optimized to use C-based xaccAccountGetBalanceAsOfDate

Unfortunately I don't think there's a ready-made C-based code for (gnc:account-get-comm-value-interval), so, category-barchart.scm will need to remain ugly for now.


christopherlam:
> the code is a lot more readable

word

My Net worth linechart is definitely faster. let me know num(trans) num(accts) num(commodities) and I'll try replicate

jralls:
34206 transactions, 763 accounts, 409 commodities. Networth line chart from 1 Jan 2001 to 31 Dec 2018 took 26 minutes with your new code, 16 with GnuCash 3.2.

christopherlam:
I'm rather not happy this code has slowed down the balance calculation... it may be that the C code is not necessarily faster than scheme. I'll try another approach which will use far fewer queries, but will use more memory, and let you know.

christopherlam:
To speed this up I can think of 3 options:
1. revert 941acee to use p.broadbery's utility functions. They are still present. I don't really understand them at all.
2. speed up xaccAccountGetBalanceAsOfDate - it's performing a linear search on every account splitlist until it finds the whereby (split->date <= date) and (split->next->date > date) -- this could easily (to a C programmer :wink:) be converted into binary search. This is be generally useful.
3. I can customize (process-datelist) in net-charts to cycle through the accounts *once*, and each account cycles through the splitlist *once*, and will collect relevant data _for all dates_ along the way. This will stretch my scheme skills!

I may be able to achieve 3 but I think someone else would be better than me at 2...

jralls:
Unfortunately 2 isn't feasible: The split list is a linked list and so must be traversed to get any particular record. One of the major deficiencies of GLib is that it doesn't have a  vector class.

It looks to me that 941acee04e3598c18eda8cb68c3b024da6f73ab1 removes actually implements 3: Isn't it creating a nested alist of (account . (date . (commodity . balance)))?

christopherlam:
> It looks to me that [941acee](https://github.com/Gnucash/gnucash/commit/941acee04e3598c18eda8cb68c3b024da6f73ab1) removes actually implements 3: Isn't it creating a nested alist of (account . (date . (commodity . balance)))?

Perhaps you're right. I've been trying to understand his optimizations for months but could never get anywhere. Unfortunately they carry a lot of baggage. Perhaps I should throw the towel. My attempt as follows, so far, after 1.5 days of work!
Hint:
- for each account
-- take splitlist, distribute splits into separate bins, according to their posted-date
-- each bin is processed by splitlist->amount to obtain balance
-- and balances are summed for every account.

Feel free to revert my attempt. I'll ping back if I succeed!
````
modified   gnucash/report/standard-reports/net-charts.scm
@@ -268,27 +268,61 @@
     ;; reversed according to income-sign-reverse general option
     ;; settings. Uses the collector->monetary conversion function
     ;; above. Returns a list of gnc-monetary.
-    (define (process-datelist accounts dates income?)
-      (map
-       (lambda (date)
-         (collector->monetary
-          ((if inc-exp?
-               (if income?
-                   gnc:accounts-get-comm-total-income
-                   gnc:accounts-get-comm-total-expense)
-               gnc:accounts-get-comm-total-assets)
-           accounts
-           (lambda (account)
-             (if inc-exp?
-                 ;; for inc-exp, 'date' is a pair of time values, else
-                 ;; it is a time value.
-                 (gnc:account-get-comm-balance-interval
-                  account (first date) (second date) #f)
-                 (gnc:account-get-comm-balance-at-date
-                  account date #f))))
-          (if inc-exp? (second date) date)))
-       dates))
-
+    (define (process-datelist accountlist dates income?)
+      (define (split->date s)
+        (and s (xaccTransGetDate (xaccSplitGetParent s))))
+      (define (splitlist->amount lst)
+        (if inc-exp?
+            (- (xaccSplitGetBalance (last lst))
+               (xaccSplitGetBalance (car lst)))
+            (xaccSplitGetBalance (car lst))))
+      (define (split->bin split)
+        (let ((date (split->date split)))
+          (let loop ((pos 0) (dates dates))
+            (and (pair? dates)
+                 (if inc-exp?
+                     (if (<= (car (car dates)) date (cadr (car dates)))
+                         pos
+                         (loop (1+ pos) (cdr dates)))
+                     (if (<= date (car dates))
+                         pos
+                         (loop (1+ pos) (cdr dates))))))))
+      (define (list->bins lst fn)
+        (define result (make-hash-table))
+        (let loop ((lst lst)
+                   (maximum 0))
+          (if (null? lst)
+              result
+              (let* ((elt (car lst))
+                     (existing (hash-ref result (fn elt))))
+                (if (fn elt)
+                    (hash-set! result (fn elt) (if existing (cons elt existing) (list elt))))
+                (loop (cdr lst)
+                      (max (or (fn elt) 0) maximum))))))
+      (define (myadd a b)
+        (if a (if b (+ a b) a) b))
+      (let loop ((accounts accountlist)
+                 (result '()))
+        (if (null? accounts)
+            (map
+             (lambda (idx)
+               (let ((acc-result (hash-ref result idx))
+                     (acc (list-ref accountlist idx)))
+               (gnc:make-gnc-monetary
+                (xaccAccountGetCommodity acc)
+                (if acc-result
+                    (map splitlist->amount acc-result)
+                    0))))
+             (iota (length accountlist)))
+            (let* ((account (car accounts))
+                   (splits (stable-sort!
+                            (xaccAccountGetSplitList account)
+                            (lambda (a b)
+                              (< (split->date a)
+                                 (split->date b))))))
+              (loop (cdr accounts)
+                    (cons (list->bins splits split->bin)
+                          result))))))
     (gnc:report-percent-done 1)
     (set! commodity-list (gnc:accounts-get-commodities
                           (append
@@ -315,6 +349,10 @@
````

PR #412, opened 2018-09-17
Bug 796858 - Add gint32 to typemap for int

loftx:
Pull request to resolve issue reported in bug


jralls:
Solved a different way.

PR #413, opened 2018-09-20
add basic test for python query

c-holtermann:
add test for creating query object and setting search_for
related to Bug 796137 and fix in commit 1a7c5b9a32df1d0ccd7e739a70b3d0f38e4dda36


jralls:
Thanks!

PR #414, opened 2018-09-21
Schemey net-charts optimization

christopherlam:
### This commit will
1. Redo the net-charts deoptimization in 72c628d
2. Add basic test-optimize-net-charts. This is similar to test-standard-net-linechart with customized intervals.
3. Add (gnc:strify) and (gnc:pk) as an upgrade to the builtin (pk) debugging tool - this understands gnc-objects and pretty-prints for ease of debugging.
4. Replace calls to xaccAccountGetBalanceAsOfDate to (account->balancelist) which will scan the accounts splitlist only ~once~ ~twice~ *once* (using scheme list cycling of account->splits until we find splits at date boundary, while ignoring any closing transactions), instead of as many times as there are number-of-dates, multiplied by up to 5 times for income charts. Perhaps this can be coded in C in the future?
5. note it is purely functional therefore uses NO set! calls so guile should be able to optimize nicely internally...

Inc-exp/balance +/- signs and amount balances look fine to me.

### Not completely tested
1. foreign commodity conversions but they are calculating well for me
2. real-world performance on large datafiles.

### Pending
1. ~adjustments for closing entries~ done, supports kvp-based closing entries filter only

### DIFF
--- a/gnucash/report/standard-reports/net-charts.scm
+++ b/gnucash/report/standard-reports/net-charts.scm
@@ -258,9 +258,6 @@
                 (warn "incompatible currencies in monetary+: " a b)))
           (warn "wrong arguments for monetary+: " a b)))
 
-    (define (monetary->double monetary)
-      (gnc-numeric-to-double (gnc:gnc-monetary-amount monetary)))
-
     (define (split->date s)
       (xaccTransGetDate (xaccSplitGetParent s)))
 
@@ -469,11 +466,11 @@
 
        ;; Add the data
        (when show-sep?
-         (add-column! (map monetary->double minuend-balances))
-         (add-column! (map - (map monetary->double subtrahend-balances))))
+         (add-column! (map gnc:gnc-monetary-amount minuend-balances))
+         (add-column! (map - (map gnc:gnc-monetary-amount subtrahend-balances))))
 
        (if show-net?
-           (add-column! (map monetary->double difference-balances)))
+           (add-column! (map gnc:gnc-monetary-amount difference-balances)))
 
        ;; Legend labels, colors
        ((if linechart?


christopherlam:
My benchmarks on a conventional multiyear datafile, report 1yr:

resolution= week
p.broadbery's version - 0.455 0.438 0.54
deoptimized = DNF
my version - 0.375 0.350 0.358

resolution = day
p.broadbery's version - 0.728 0.690 0.704
deoptimized = DNF
my version - 0.62 0.69 0.69

so, marginally faster... I wonder if C will be even faster.

gnucash/report/standard-reports/net-charts.scm
```diff
@@ -263,6 +263,61 @@
...
+    ;; this function will scan through the account splitlist, building
+    ;; a list of balances along the way. it will use the dates
+    ;; specified in the variable dates-list.
+    ;; input: account
+    ;;  uses: dates-list (list of time64)
+    ;;   out: (list account bal0 bal1 ...)
+    (define (account->balancelist account)
+      (let loop ((splits (stable-sort! (xaccAccountGetSplitList account)
+                                       (lambda (a b)
+                                         (< (split->date a) (split->date b)))))
```
christopherlam: @jralls from experimentation, it would seem that xaccAccountGetSplitList will *always* return a splitlist, sorted by posted-date. If this is correct, I can remove sorting here which will be rather nice. Can you confirm if the engine will enforce this assumption? Reviewing xaccAccountGetBalanceAsOfDate will call sorting but states it would usually be a NOOP. https://github.com/Gnucash/gnucash/blob/maint/libgnucash/engine/Account.cpp#L3293

jralls: @christopherlam You actually wanted L3701, but yes, it calls xaccAccountSortSplits to make sure that they're in the correct order. That's a bit more complicated than posted date because it has more sorts when there are more than one split posted on a particular date, but date is the outermost order so sorting the returned list by date is a no-op.

christopherlam: Happy times!

christopherlam:
Ok I've run out of ideas to speed this up...

jralls:
Have you used [statprof](https://www.gnu.org/software/guile/manual/html_node/Statprof.html)?

christopherlam:
Hmm just tried; doesn't give me any more information.

The above commits will lead to:

1) single splitlist pass for each account selected, collecting amounts along the way, ignore closing-txns for inc-exp report, and at each date-list boundary will accumulate balance-list.
2) single balance-list pass for each account, creating type-specific (asset/liability/inc/exp) balance-list.

The only remaining bottlenecks are the commodity-collector system which is a pain to optimize, and possibly upgrade xaccAccountRecomputeBalance to track no-closing balances, which will lead to slower regular use (I think).

christopherlam:
Please let me know benchmarks.

I still think kvp searching is expensive, and the 2 ways to speed further is either (1) change engine to store ClosingTxn status as a split->closing? property instead of having to search kvp, tree or (2) modify xaccAccountRecalculate to track balances-ignoring-closing.

jralls:
Exactly 12 minutes for the usual networth line chart.
Here's the top of the profiling list:
```
%     cumulative   self
time   seconds    seconds   calls   procedure
52.40   612.81    369.22            commodity-utilities.scm:51:18
11.01    77.62     77.57            xaccSplitGetParent
 9.28    65.42     65.40            xaccAccountGetCommodity
 5.26    37.09     37.06            xaccTransGetCurrency
 4.80    33.82     33.81            xaccSplitGetAccount
 4.56    32.13     32.13            gnc-commodity-equiv
 4.00    28.33     28.17            display
 3.69    26.01     26.01            gnc-commodity-print-info
```
The first one is the filter-lambda in gnc:get-match-commodity-splits. Note that the cumulative seconds for the lambda (613) is pretty close to the self seconds for the lambda and the next five, which it calls. That suggests that there's a fair amount of overhead there.
The comment explaining the lambda isn't quite correct. It says "Filter such that we get only those splits which have two different commodities involved" (sic), but if a commodity is passed in it also filters for those splits where either the split commodity or the transaction commodity are the same as the passed-in commodity.

If looking up the split's closed status even appears in the profile I don't see it. In fact, I can't find any reference to it in the reports module. Where is it?

Since the top 4 C functions are just pointer dereferences, the overhead must be in Swig wrapping the returned pointers. A single C function taking a split and a commodity and doing all of the testing, maybe
```C
gboolean
gnc_split_are_commodities_equivalent(Split *split, gnc_commodity *comm)
{
    gnc_commodity *trans_comm = NULL, *split_comm = NULL;
    g_return_val_if_fail (split != NULL, FALSE);
    trans_comm = xaccTransGetCurrency (xaccSplitGetParent (split));
    split_comm = xaccAccountGetCommodity (xaccSplitGetAccount (split));
    return (trans_comm && split_comm && gnc_commodity_equiv (split_comm, trans_comm) &&
            (comm == NULL || gnc_commodity_equiv (comm, trans_comm) || 
             gnc_commodity_equiv (comm, split_comm)));
}
```
(Which is pretty much verbatim your Scheme code with different punctuation.)
After that the top-5 in the profile became:
```
%     cumulative   self
time   seconds    seconds   calls   procedure
 73.07    505.02    505.02          commodity-utilities.scm:51:18
 23.46    162.10    162.10          gnc-split-are-commodities-equivalent
  0.91      6.31      6.29          display
  0.68      4.73      4.73          gnc-window-show-progress
  0.68      4.72      4.72          qof-query-run
```
Interestingly the overall time didn't change at all.

christopherlam:
If I understand your profiling, you are using 'weighted-average which calls (gnc:get-commodity-totalavg-prices) and is spending a lot of time scanning the *whole* book for 'interesting' transactions involving >2 commodities, which means this is a particularly complicated book currency-wise? I've always found weighted-average to be slow and haven't optimized this part at all.

I have only optimized net-chart's process-datelist -- e.g. prior to p.broadbery, if 100 accounts were selected for weekly networth over 5 years, it'd scan each account 52*5 times to get balances but now it scans only once. p.broadbery does broadly the same thing but converts list->vector and does binary search to find splits, then grabs split->balance.

I think your 12 minutes is mostly being spent calculating the prices for weighted-average rather than accumulating balances, which means my optimization is working well.

> If looking up the split's closed status even appears in the profile I don't see it. In fact, I can't find any reference to it in the reports module. Where is it?

I was referring to account->balancelist in net-charts.scm which is optionally querying closing-txn status, which means a hike up kvp tree for every split, but I've enabled for inc-exp charts only.

christopherlam:
> (Which is pretty much verbatim your Scheme code with different punctuation.)

That's not actually my scheme code :)

jralls:
Well, if I'm testing the wrong thing it certainly explains why there's no time difference. Yes, the times were the default weighted average price. I'm waiting the 12 minutes for that to run now so that I can change the option, but I'll change the default so I don't have to wait...
But I'll have to bail for the day in a few minutes.

christopherlam:
Ok here's a low hanging-fruit idea for optimizing weighted-average:
replace (qof-query-run query) with (xaccQueryGetSplitsUniqueTrans query) in line 63... it seems to make no difference on my complicated test book, and should slice your weighted-average calculator?

jralls:
Sounds worthwhile, queries are *really* expensive. But for testing this PR would nearest-in-time be better?

christopherlam:
> Sounds worthwhile, queries are _really_ expensive. But for testing this PR would nearest-in-time be better?

Yes :)

jralls:
OK, I've changed the default for net-chart to 'nearest-in-time, at least temporarily.

jralls:
That's relatively intantaneous:
```
d13.6344 t729.545: (running report...)
%     cumulative   self
time   seconds     seconds  procedure
 47.92      0.44      0.44  srfi/srfi-1.scm:918:15
 18.01      0.17      0.17  time64CanonicalDayTime
  6.09      0.07      0.06  report-utilities.scm:134:17
  3.60      0.49      0.03  find-tail
  2.49      0.04      0.02  report-utilities.scm:356:4
  2.49      0.02      0.02  gnc-window-show-progress
...
Sample count: 361
Total time: 0.926078 seconds (0.0 seconds in GC)
d1.0046 t730.550: (completed report.)
```

jralls:
Now to switch branches, instrument maint, and try again.

christopherlam:
(Don't forget to test maint together with the gnc:pk and the benchmarking commit)

jralls:
Yup.
```
d34.7796 t34.780: (running report...)
%     cumulative   self
time   seconds     seconds  procedure
 46.88      2.15      2.15  list?
  7.15      0.33      0.33  gnc-window-show-progress
  4.72      0.22      0.22  time64CanonicalDayTime
  4.41      1.31      0.20  collectors.scm:301:22
  4.41      0.40      0.20  report-utilities.scm:356:4
  4.11      0.19      0.19  gnc-numeric-convert
  3.65      0.24      0.17  report-utilities.scm:134:17
  3.50      0.16      0.16  xaccTransGetIsClosingTxn
  1.67      0.08      0.08  gnc-commodity-get-mnemonic
  1.52      0.36      0.07  report-collectors.scm:174:31
  1.52      0.07      0.07  report-utilities.scm:74:10
  1.22      0.06      0.06  report-utilities.scm:249:4
...
Sample count: 657
Total time: 4.589019 seconds (0.247322 seconds in GC)
d4.5294 t39.309: (completed report.)
```
So you've made a 75% speedup. Not bad!

christopherlam:
\o/

jralls:
GTG.

christopherlam:
Looking at profiling, my code spends 0.0s in garbage collector which is a warm feeling...

My main motivation here is to deprecate the ugly collectors.scm and friends, and remove the incomprehensible magic. Speed is a nice side effect! I'll clean up and merge test-optimize into test-charts, and repush.

When I can spare some time I'll redo category-barchart as well which means I can probably reuse my functions.

jralls:
Did you use gcprof? IIU [statprof](https://www.gnu.org/software/guile/manual/html_node/Statprof.html) correctly that's the only way to see what the gc is doing.

BTW, I suggest turning off count-calls. The docs say it has significant overhead that distorts the profiling results.

christopherlam:
> Did you use gcprof? IIU [statprof](https://www.gnu.org/software/guile/manual/html_node/Statprof.html) correctly that's the only way to see what the gc is doing.

No, I believe my code is clean because avoids any (set!) calls, therefore doesn't lead to dangling objects.

> BTW, I suggest turning off count-calls. The docs say it has significant overhead that distorts the profiling results.

Ok removed from this PR anyway, which is now tidied up. Next project: weighted average price calculator.

christopherlam:
The only further comment I can offer here is that the functions defined here are very likely to be reusable for category-barchart and my intention is to "upgrade" account->balancelist to report-utilities.scm once I start to examine category barchart to know exactly what it needs.

gnucash/report/standard-reports/net-charts.scm
```diff
@@ -263,6 +263,72 @@
     (define (monetary->double monetary)
       (gnc-numeric-to-double (gnc:gnc-monetary-amount monetary)))
 
+    (define (split->date s)
+      (xaccTransGetDate (xaccSplitGetParent s)))
+
+    ;; this function will scan through the account splitlist, building
+    ;; a list of balances along the way. it will use the dates
+    ;; specified in the variable dates-list.
+    ;; input: account
```
gjanssens: Any reason not to specify dates-list as a parameter to the function rather than assuming it's available in the parent scope ?

christopherlam: > Any reason not to specify dates-list as a parameter to the function rather than assuming it's available in the parent scope ?
> Oh never mind. The function is meant to be called in a map context, which only takes one argument...

Yeah trying to be elegant for map... :)

Later on when exporting this function I'll likely be creating a wrapper for it... I'd like to call this xaccAccountGetBalancesAsOfDates but this wouldn't be consistent with other scm functions :-/

christopherlam:
11 months later, would be nice to know how much b9382d2c had an effect in speeding up weighted-average ;) from 12 minutes to ???

jralls:
3-4 seconds.

christopherlam:
wow...

jralls:
The trial balance report still takes ~150 seconds... just in case you're still looking for something to speed up. I can also get the networth linechart to take 12 seconds if I stretch out the time span to 10 years.

PR #416, opened 2018-09-24
Update documentation of python binding example

myaser:


gjanssens:
Hmm, while your changed help text is certainly better than the plain wrong old one, it's still not very nice: it has a typo and is still a bit fuzzy as to what will happen exactly. How about:

```
Recurse over all accounts in a book and marks the first one having target_account_code as tax related
```

myaser:
it is not Recursing over accounts in a book, but over sub-accounts of one root account and its children
```
mark an account having target_account_code as tax related, else Recurse over all of its sub-accounts  marks the first one having target_account_code as tax related
```

gjanssens:
Thanks!

PR #417, opened 2018-09-26
General progress .scm files

christopherlam:
~(gnc:pk) is a generally useful debugging tool. Surround any procedure call, or object, with (gnc:pk) and the program will run unchanged, yet the object will be dumped in the tracefile and 2 timestamps printed; both running timestamp and delta timestamp since the last (gnc:pk) call.~

transaction.scm upgraded to optionally toggle transaction/account anchors. seems to be an often-requested feature upgrade.


jralls:
Thanks!

PR #419, opened 2018-10-01
Bug796872 Import OFX fix for V3.3

DaveC49:
Incorporates fixes to import-main-matcher.c and dialog-import.glade from bug796778 and corrects import OFX problem of not being able to select transactions in the import matcher


jralls:
The problem turned out to have been operator error.

PR #420, opened 2018-10-09
Generate a filtered splitlist early for weighted-average reports

christopherlam:
Experimental PR. Instead of repeating a whole-book query to look for prices for every commodity, cache the splitlist and reuse it for each commodity. This should dramatically cut down weighted-average ~and average-cost~ for books with large numbers of commodities. If book has 100 commodities, it would previously be running 100 queries and perform very similar analyses of the resulting splitlist.

~I've added `(get-all-splits #f)`  to clear the cache between reports, but I am unsure if I've inserted at the correct location.~

Mainly for verifying numbers are still correct, and testing how much speedup is possible. (in theory, book with N commodities with report taking t will reduce to, at most, t/N)

~This is not my preferred approach -- I'd refactor and rework but this would be disruptive to the exported functions.~

### DIFF
--- a/gnucash/report/report-system/commodity-utilities.scm
+++ b/gnucash/report/report-system/commodity-utilities.scm
@@ -95,7 +95,7 @@ construct with gnc:make-gnc-monetary and gnc:monetary->string instead.")
 ;; Returns true if the given pricealist element is a non-zero price.
 (define (gnc:price-is-not-zero? elem)
   (and (second elem)
-       (not (gnc-numeric-zero-p (second elem)))))
+       (not (zero? (second elem)))))
 
 ;; Create a list of all prices of 'price-commodity' measured in the currency
 ;; 'report-currency'. The prices are taken from all splits in
@@ -123,8 +123,8 @@ construct with gnc:make-gnc-monetary and gnc:monetary->string instead.")
 (define* (gnc:get-commodity-totalavg-prices
           currency-accounts end-date price-commodity report-currency
           #:key commodity-splits)
-  (let ((total-foreign (gnc-numeric-zero))
-        (total-domestic (gnc-numeric-zero)))
+  (let ((total-foreign 0)
+        (total-domestic 0))
     (filter
      gnc:price-is-not-zero?
      (map-in-order
@@ -133,9 +133,9 @@ construct with gnc:make-gnc-monetary and gnc:monetary->string instead.")
                                   (xaccSplitGetParent a)))
                (account-comm (xaccAccountGetCommodity
                               (xaccSplitGetAccount a)))
-               (share-amount (gnc-numeric-abs
+               (share-amount (abs
                               (xaccSplitGetAmount a)))
-               (value-amount (gnc-numeric-abs
+               (value-amount (abs
                               (xaccSplitGetValue a)))
                (transaction-date (xaccTransGetDate
                                   (xaccSplitGetParent a)))
@@ -182,8 +182,8 @@ construct with gnc:make-gnc-monetary and gnc:monetary->string instead.")
                            ") =? "
                            (gnc:monetary->string
                             (gnc:make-gnc-monetary
-                            report-currency (gnc-numeric-zero))))
-                     (gnc-numeric-zero))
+                            report-currency 0)))
+                     0)
                    (begin
                      (set! total-foreign (gnc-numeric-add total-foreign
                                                           (third foreignlist)
@@ -222,8 +222,6 @@ construct with gnc:make-gnc-monetary and gnc:monetary->string instead.")
     (< (xaccTransGetDate (xaccSplitGetParent a))
        (xaccTransGetDate (xaccSplitGetParent b))))
   (let* ((currency-accounts
-          ;;(filter gnc:account-has-shares?
-          ;; -- use all accounts, not only share accounts, since gnucash-1.7
           (gnc-account-get-descendants-sorted (gnc-get-current-root-account)))
          (all-splits (get-all-splits currency-accounts end-date))
          (interesting-splits (sort (filter interesting-split? all-splits) date<?))
@@ -261,9 +259,9 @@ construct with gnc:make-gnc-monetary and gnc:monetary->string instead.")
                                 (xaccSplitGetParent a)))
              (account-comm (xaccAccountGetCommodity
                             (xaccSplitGetAccount a)))
-             (share-amount (gnc-numeric-abs
+             (share-amount (abs
                             (xaccSplitGetAmount a)))
-             (value-amount (gnc-numeric-abs
+             (value-amount (abs
                             (xaccSplitGetValue a)))
              (transaction-date (xaccTransGetDate
                                 (xaccSplitGetParent a)))
@@ -303,8 +301,8 @@ construct with gnc:make-gnc-monetary and gnc:monetary->string instead.")
                        ") =? "
                        (gnc:monetary->string
                         (gnc:make-gnc-monetary
-                         report-currency (gnc-numeric-zero))))
-               (gnc-numeric-zero))
+                         report-currency 0)))
+               0)
              (if (not (zero? (third foreignlist)))
                  (gnc-numeric-div
                   (second foreignlist)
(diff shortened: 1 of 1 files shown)


christopherlam:
Ok RFC.... and benchmarks please :) for weighted-average. I think average-cost speed is unchanged.

jralls:
I'm away tomorrow, I'll have a go at it Thursday (Friday for you ;-) ).

christopherlam:
I think I will amend - I'm not a fan of closures which needs keeping track of state. I'll make further cleaner amendments (and still fast) and won't require bubbling.

christopherlam:
@jralls this is independent of #414 so, they should work together nicely :)

jralls:
Statprof results:
weighted-average:
%     cumulative   self
time   seconds     seconds  procedure
 25.15     11.72     11.04  display
 22.79     10.01     10.01  gnc-commodity-print-info
 11.89      9.80      5.22  commodity-utilities.scm:232:14:split-has-commodity?
  3.97      1.74      1.74  xaccSplitGetParent
  3.25      1.48      1.42  commodity-utilities.scm:350:22
  3.17      1.39      1.39  gnc-commodity-equiv
  2.95      1.29      1.29  list?
  2.71      1.19      1.19  gnc-window-show-progress
  2.48      1.09      1.09  xaccAccountGetCommodity
  2.07      0.91      0.91  gnc-numeric-convert
  1.81      0.79      0.79  gnc-numeric-mul
  1.65      0.72      0.72  %after-gc-thunk
...
Sample count: 11122
Total time: 43.901055 seconds (6.303563 seconds in GC)

average-cost:
%     cumulative   self
time   seconds     seconds  procedure
 19.60      1.20      1.20  list?
 15.08      1.47      0.92  commodity-utilities.scm:47:3
  7.54      0.46      0.46  gnc-window-show-progress
  6.53      0.66      0.40  report-utilities.scm:355:4
  4.19      0.26      0.26  gnc-numeric-convert
  4.02      0.30      0.25  report-utilities.scm:133:17
  3.85      1.99      0.23  collectors.scm:301:22
  3.69      0.22      0.22  xaccAccountGetCommodity
  3.35      0.20      0.20  xaccTransGetIsClosingTxn
  2.68      0.16      0.16  xaccSplitGetParent
...
Sample count: 597
Total time: 6.097589 seconds (0.147358 seconds in GC)

christopherlam:
Ok cleaned up this PR. No profiling, and cleaned up. If this is acceptable, this PR and #414 should be imminently mergeable. Both are, I believe, 100% backward compatible, are faster, and have fewer dependencies. As @code-gnucash-org in https://github.com/Gnucash/gnucash/pull/423#issuecomment-428961416 says, once merged in, I can move on to the next.

christopherlam:
Perhaps one last idea is to swap the order of the two (or) arguments in the following:
```
(define (split-has-commodity? s)
         (or (gnc-commodity-equiv c (xaccTransGetCurrency (xaccSplitGetParent s)))
             (gnc-commodity-equiv c (xaccAccountGetCommodity (xaccSplitGetAccount s)))))
```

I can't figure out which one would be more common in a typical multicurrency and multicommodity large book, and would evaluate to #t first. Please experiment and swap to the faster version!

jralls:
I don't think there's going to be a "most common". I started to claim that trans would be, because it's more likely to be the book currency and the report is most likely to be run in the book currency. But a user could flip that by starting most transactions in the "other commodity" accounts. For someone trading in two currencies, say a British firm with business in Euro, there's no benefit to recording the transaction in one or the other. Someone trading securities it might find it easier to enter those trades in the securities registers where they can see the amount, price, and value. That's doubly true for sales as one can't create the capital gain/loss splits in a currency register.

jralls:
@gjanssens Any concerns about merging this and #414?

gjanssens:
I'll try to do my review somewhere this weekend.

gnucash/report/report-system/commodity-utilities.scm
```diff
@@ -129,8 +115,14 @@ construct with gnc:make-gnc-monetary and gnc:monetary->string instead.")
...
 ;; The intended second price is 6500/300, or 21.67.
 
-(define (gnc:get-commodity-totalavg-prices
-         currency-accounts end-date price-commodity report-currency)
+;; NOTE there is an optional #:commodity-splits argument which should
+;; contain a sorted splitlist with pricing information related to the
+;; price-commodity. See (gnc:get-commoditylist-totalavg-prices) how to
+;; generate this splitlist. This is NOT to be used by external
+;; reports.
+(define* (gnc:get-commodity-totalavg-prices
```
gjanssens: If only the comment states this is NOT to be used by external reports, chances are someone will eventually ignore the warning and try to use the function in external reports anyway. Is there a way to prevent this in code also ?
Like splitting this up in a (define (gnc:get-commodity-totalavg-prices ...)) without the optional parameter and a, say (gnc:det-commodity-totalavg-prices-internal ...)) with the optional parameter. The former is exported, the latter isn't. The former will simply call the latter, and internal code can call the latter as well with optional parameter.

jralls: More to the point, why shouldn't it be used by external reports if they can generate the splitlist?

gnucash/report/report-system/commodity-utilities.scm
```diff
@@ -23,49 +23,35 @@
...
-                                  (not commodity)
-                                  (gnc-commodity-equiv
-                                   commodity trans-comm)
-                                  (gnc-commodity-equiv
-                                   commodity acc-comm)))))
-                  (qof-query-run query)))
-    (qof-query-destroy query)
-    splits))
+  ;; Filter such that we get only those splits which have two
+  ;; *different* commodities involved.
```
gjanssens: As @jralls also mentioned in PR #414 this comment is incomplete. I had noticed this but eventually forgot to raise a comment here. As it is explained well in the whole function's comment, perhaps this one can be dropped ?

christopherlam:
> Like splitting this up in a (define (gnc:get-commodity-totalavg-prices ...)) without the optional parameter and a, say (gnc:det-commodity-totalavg-prices-internal ...)) with the optional parameter. The former is exported, the latter isn't. The former will simply call the latter, and internal code can call the latter as well with optional parameter.

Hmm no particular reason... except for me trying to economise on number of functions present in this utility file.



> More to the point, why shouldn't it be used by external reports if they can generate the splitlist?

Again no particular reason except for trying not to export too much functionality.

I think the 'internal' function could do the trick well.

I'll do all in 1 final commit and push -- ideally I'd fixup...

jralls:
Thanks.

> Too many little functions = too many head scratching by future hackers. 

Most people think the opposite: Lots of short functions composed together in a control function is easier to understand than having the same  code in a single hundred-line (or worse, hundreds-of-lines) function. I certainly find that to be the case.

christopherlam:
> Most people think the opposite: Lots of short functions composed together in a control function is easier to understand than having the same code in a single hundred-line (or worse, hundreds-of-lines) function. I certainly find that to be the case.

Ok in many ways I agree many functions are useful if these short functions are defined in the control-function scope rather than globally. My stalled baIsheet-pnl does have numerous internal functions *enclosed* within a control function. In any case this is now set in stone.

For the sake of completeness and documentation, a final benchmark wouldn't be unwelcome :) trying to isolate the price calculators only i.e. keep same size of book, disable table display... incorporating both optimizations merged in. I'll still try find ways to slice the 45s!

christopherlam:
I think:
````
(use-modules (statprof))

(statprof
 (lambda ()
    (set! html (gnc:report-render-html report #t)))
 #:hz 1000)
````

christopherlam:
@derekatkins Maybe... but not from 42.4s to 2.91s. Original jralls's large book of, I guess 15-20 commodities, weighted-average was 12mins, optimized to 45s in this PR, and to 42s in conjunction with #414. I don't think average-cost speed has changed.

I think this time could be trimmed further if (qof-query-run) was replaced with (xaccQueryGetSplitsUniqueTrans) -- after all we're retrieving interesting-splits from book, and I think there's no point retrieving them twice. But I still can't figure out the price-calculators completely yet. Perhaps I will one day, and can smash this target.

I suspect (but can't confirm) this speedup will propagate to all reports which use weighted-average price API. We've shown, in this experiment, that guile is as fast as C :+1: and needs no malloc nor free. guile-3 should be even faster...

jralls:
> I guess 15-20 commodities

409

> We've shown, in this experiment, that guile is as fast as C 

Not really. You've shown that traversing a linked list of ~70K splits ~400 times is slow no matter what language you do it in.

PR #424, opened 2018-10-13
Bug796896

Bob-IT:
Have moved the OK button to the assistant action area and hidden the forward button. If this is acceptable then I could do the same for the csv import assistants on this PR. Looks OK on Linux and have tried on Windows 10 and works there too.
Looking about the last commit, maybe that should of been a translation change.

### DIFF
--- a/gnucash/import-export/csv-imp/csv-account-import.c
+++ b/gnucash/import-export/csv-imp/csv-account-import.c
@@ -78,7 +78,8 @@ fill_model_with_match(GMatchInfo *match_info,
  * Parse the file for a correctly formatted file
  *******************************************************/
 csv_import_result
-csv_import_read_file (const gchar *filename, const gchar *parser_regexp,
+csv_import_read_file (GtkWindow *win, const gchar *filename,
+                      const gchar *parser_regexp,
                       GtkListStore *store, guint max_rows)
 {
     gchar      *locale_cont, *contents;
--- a/gnucash/gtkbuilder/assistant-csv-account-import.glade
+++ b/gnucash/gtkbuilder/assistant-csv-account-import.glade
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<!-- Generated with glade 3.20.0 -->
+<!-- Generated with glade 3.20.4 -->
 <interface>
   <requires lib="gtk+" version="3.10"/>
   <object class="GtkAdjustment" id="num_hrows_adj">
@@ -279,7 +279,6 @@ Enter file name and location for the Import...
         </child>
       </object>
       <packing>
-        <property name="page_type">progress</property>
         <property name="title" translatable="yes">Import Account Preview, first 10 rows only</property>
         <property name="complete">True</property>
         <property name="has_padding">False</property>
--- a/gnucash/gtkbuilder/assistant-csv-price-import.glade
+++ b/gnucash/gtkbuilder/assistant-csv-price-import.glade
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<!-- Generated with glade 3.20.2 -->
+<!-- Generated with glade 3.20.4 -->
 <interface>
   <requires lib="gtk+" version="3.10"/>
   <object class="GtkAdjustment" id="end_row_adj">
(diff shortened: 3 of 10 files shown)


jralls:
Yes, the last change needs to be echoed in the po files (ex, fa, no, sk, tr, and uk) that have the string already.

The "Next" button isn't hidden, it's just disabled. Granted having the "OK" button on the same row is a bit less confusing, especially on Linux where the assistant buttons are across the top (who thought *that* was a good idea?). It helps on Linux that the Back and Cancel buttons are on the left and the OK and Next ones are on the right. That doesn't work as well on MacOS where they're all in a row on the bottom, OK, Cancel, Back, Next. 

It's also not so good that the OK button now appears enabled on every page, putting up the "No file name is selected" message box when clicked. Not only that but clicking it on the wrong page disables the Next button so that the assistant has to be canceled and restarted.

Bob-IT:
Well that is disappointing, on my Linux box and on the Windows machine I tested on the buttons are along the bottom of the assistant, the OK button is hidden on all pages apart from the file page. The Next button is visible only on the required pages apart from the file page.

I will need to check this again but am sure of that, so the only other option is to have some thing similar to what's in the QIF importer, a GtkEntry or GtkLabel and a button that launches a FileChooserDialog.

gjanssens:
Op woensdag 17 oktober 2018 00:23:30 CEST schreef John Ralls:
> It's probably a styling issue, different distros take liberties with CSS to
> differentiate themselves. That's what the article Geert posted on
> Gnucash-devel yesterday was about. I tested on Debian Stretch. What Linux
> did you use?

On my Fedora machine the buttons are below as well.
> 
> Why not just lose the OK button and use the Next button?

That's what I was hoping as a solution as well.

However ISTR from past attempts we didn't find a way to intercept the file 
chooser's file selection handling to enable or disable page forwarding.

I looked again, and I believe we  can do so now via
https://developer.gnome.org/gtk3/stable/GtkFileChooser.html#GtkFileChooser-selection-changed

So I think there's no need for a separate ok button any more. Instead
1. listen for the file chooser's selection-changed signal
2. determine whether the selection is a single, valid file (so no directory)
3. If so mark the page as complete (so the next button is enabled) otherwise 
mark is as not complete.

Bob-IT:
I have removed the OK button and enabled the 'forward' button when a file is selected and made similar changes on three other CSV assistants.

The last commit is for Bug779565 which I reported a while ago and remembered while testing to fix the first use of the treeview header preview combo's. There is still something else wrong with the combo's, they slightly move when mouse is over them which can also be seen on windows I assume that's down to gtk. I wonder if a popup menu widget might be better but that would be a master fix.

gjanssens:
There is still an issue in at least the csv transaction importer, and maybe in the others as well.
You have replaced `file_confirm_cb` with `file_selection_changed_cb`. However the scope of these functions is different. `file_confirm_cb was` called when the user has make his/her final file selection. `file_selection_changed_cb` is called each time the selection changes, so during active navigation.

The way `file_selection_changed_cb` is currently implemented it will actively try to load and parse the currently selected file each time the selection changes. This can have several side effects. In my short tests I have experienced these already:
1. If the file the user clicks on is empty, a pop up will be displayed that parsing failed.
1. If the file is large, the parsing will take a while and the dialog will temporarily be unresponsive.

I think `file_selection_changed_cb`'s functionality should be restricted to only setting the page to complete or not based on some selection criteria. For now that means a valid file is selected and it's not a directory. We could refine this if we like, but for now that should be sufficient.

Only when the user clicks "Forward" the importer should try to parse the file. So that bit should probably move to `assist_preview_page_prepare`. If parsing fails, a message will be popped up (as is now already). And the "next" page should be reset to the file selection page (new).

gjanssens:
And I see similar issues when using selecting a file in the export assistant: when I click on a file it will immediately pop up a question whether I want to overwrite this file. That question should only happen after a click on forward. The user should not be hindered in any way while navigating the file system tree. Only the forward button should reflect whether the current selection *can* serve as a file to go forward.

Bob-IT:
I think I have made the required changes for the processing after the forward button is pressed. Will await feedback...

gnucash/import-export/csv-imp/csv-account-import.c
```diff
@@ -78,7 +78,8 @@ fill_model_with_match(GMatchInfo *match_info,
  * Parse the file for a correctly formatted file
  *******************************************************/
 csv_import_result
-csv_import_read_file (const gchar *filename, const gchar *parser_regexp,
+csv_import_read_file (GtkWindow *win, const gchar *filename,
```
gjanssens: An oversight probably: you've added a parent widget in the function call but as far as I can see you're not using it ?

gjanssens:
Thanks!

PR #425, opened 2018-10-17
Doing some german translations

hauptsignal:
I have done some translations. Please review my changes carefully because this is my first try editing this file.


hauptsignal:
There is an error.

PR #426, opened 2018-10-17
 Doing some german translations

hauptsignal:
I have done some german translations. Please check my changes carefully because this is my first try to edit this file.

### DIFF
--- a/po/de.po
+++ b/po/de.po
@@ -9157,7 +9157,7 @@ msgid ""
 "Enable debugging mode: provide deep detail in the logs.\n"
 "This is equivalent to: --log \"=info\" --log \"qof=info\" --log \"gnc=info\""
 msgstr ""
-"Debugging aktivieren: liefere ausführliche Details in den Logs.\"
+"Debugging aktivieren: liefere ausführliche Details in den Logs.\n"
 "Das ist äquivalent zu: --log \"=info\" --log \"qof=info\" --log \"gnc=info\""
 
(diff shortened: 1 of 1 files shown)


gjanssens:
Thanks for your contribution.

I can't comment on the translations itself as I only speak a little bit of German. So I'll leave that to someone with more authority.

I have these side remarks for you though:
* I can't deduce from your PR message whether you contacted the previous translators before starting your work. While your effort is surely appreciated, it would be frustrating to all involved if this would lead to two or more people doing the same thing. So if you didn't, be sure to contact the last active translators as found at the beginning the de.po file.
* If you did coordinate with the previous translators, you can add your name to the list of translators at the top of the file and in the translator credits and update the PO revision date.
* I saw you closed the previous PR and started a new one. However that is not necessary. If you make fixes you can simply push them to your existing branch and they will appear in the PR automatically. That will save you some work.
* I haven't done so already, I recommend you also read our (translation wiki page)[https://wiki.gnucash.org/wiki/Translation]. It contains a lot of documentation and tips to help you with translating.

fellen:
@gjanssens https://lists.gnucash.org/pipermail/gnucash-de/2018-October/010478.html

gjanssens:
Ah, so this was discussed on the German list. Perfect! I will leave it to @fellen now :)

hauptsignal:
@gjanssens Thank you for your remarks.

PR #430, opened 2018-10-26
Bug #408444 - Fix "Smallest fraction" disregarded

Stummi:
When importing a CSV-File into a account with a non standard SCU configured, the imported Transactions will still round to the currencies default SCU. This PR fixes that issue.

### DIFF
--- a/libgnucash/engine/Split.c
+++ b/libgnucash/engine/Split.c
@@ -1240,6 +1240,9 @@ qofSplitSetValue (Split *split, gnc_numeric amt)
 void
 xaccSplitSetValue (Split *s, gnc_numeric amt)
 {
+    const gnc_commodity *currency;
+    const gnc_commodity *commodity;
+
     gnc_numeric new_val;
     if (!s) return;
 
@@ -1249,8 +1252,19 @@ xaccSplitSetValue (Split *s, gnc_numeric amt)
            s->value.num, s->value.denom, amt.num, amt.denom);
 
     xaccTransBeginEdit (s->parent);
-    new_val = gnc_numeric_convert(amt, get_currency_denom(s),
+
+    currency = xaccTransGetCurrency (s->parent);
+    commodity = xaccAccountGetCommodity (s->acc);
+
+    if (gnc_commodity_equiv(currency, commodity))
+    {
+        // if transaction and account currency are the same, use the accounts SCU as denom for split entry
+        new_val = gnc_numeric_convert(amt, xaccAccountGetCommoditySCU(s->acc),
+                                      GNC_HOW_RND_ROUND_HALF_UP);;
+    } else {
+        new_val = gnc_numeric_convert(amt, get_currency_denom(s),
                                   GNC_HOW_RND_ROUND_HALF_UP);
+    }
     if (gnc_numeric_check(new_val) == GNC_ERROR_OK &&
         !(gnc_numeric_zero_p (new_val) && !gnc_numeric_zero_p (amt)))
         s->value = new_val;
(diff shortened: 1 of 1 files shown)


jralls:
Did this pass ```make check``` on your system?

Stummi:
`make check` failed for me for both, my HEAD and the upstream maint head, with the same error message:

```
/usr/bin/ld: /usr/lib/gcc/x86_64-linux-gnu/8/../../../x86_64-linux-gnu/libgtest_main.a(gtest_main.cc.o): in function `main':
(.text.startup+0x21): undefined reference to `testing::InitGoogleTest(int*, char**)'
/usr/bin/ld: (.text.startup+0x26): undefined reference to `testing::UnitTest::GetInstance()'
/usr/bin/ld: (.text.startup+0x2e): undefined reference to `testing::UnitTest::Run()'
make[3]: *** [gnucash/import-export/csv-imp/test/CMakeFiles/test-tx_import.dir/build.make:100: bin/test-tx_import] Fehler 1
make[2]: *** [CMakeFiles/Makefile2:9478: gnucash/import-export/csv-imp/test/CMakeFiles/test-tx_import.dir/all] Fehler 2
make[1]: *** [CMakeFiles/Makefile2:233: CMakeFiles/check.dir/rule] Fehler 2
make: *** [Makefile:223: check] Fehler 2
```

jralls:
That's a different error from the one Travis is reporting and indicates that you didn't get a good build. You'll need to fix that before you can proceed to fixing the failures that Travis reported. We don't merge PRs with failing tests, but sometimes the two environments Travis uses don't match the submitter's build environment and tests will pass for the submitter and fail on Travis. The problem Travis is complaining about is splits with unset accounts (i.e. split->acct == NULL). That's a situation that happens in testing but shouldn't when running GnuCash, so the solution may be to fix the tests.

That aside, your patch will work only in cases where all of the accounts in a particular currency use the same SCU. If they don't there will be an imbalance that must be allocated somewhere. For example suppose you have account A with an SCU of 1 and account B with SCU of 100, both in dollars. You transfer $2.75 from B to A, where the value is rounded to $3.00. B's value remains $2.75, so the transaction has a $.25 imbalance. That's a variant of the issue you posted to IRC yesterday, where the imbalance arose because the expense split's value was "banker's rounded" to the nearest even number (0) and the bank split's value was rounded the usual way to the nearest integer (1), both divided by the currency SCU, 100. Had they both rounded the same way your 9-mil transaction would have simply turned into a 1-cent one.

The transaction balancing code expects all of the split->values to be in the transaction currency and, since balancing is an add operation it needs a common denominator. Using the transaction currency's SCU provides that common denominator . When split->amount is in a different currency the conversion is accomplished by a price, but there's no mechanism to accommodate different SCUs.

In addition to transaction balancing GnuCash must keep its version of the accounting equation in balance: Assets + Income = Liabilities + Equity + Expenses. (In formal accounting Income and Expense are both part of Equity, but GnuCash's way is equivalent and makes some calculations easier to program.) That balance is performed by the Balance Sheet and Trial Balance reports. They use the amounts and convert them to a report currency (which isn't necessarily one of the currencies in any particular transaction) at a single price. That means that any scheme to handle different SCUs in a transaction must deal with amounts as well as values.

Stummi:
Hello, sorry for the late reply, somehow totally forgot about this.

I am not sure if its worth to put additional work into this, since as far as I understood the current described behavior is actually expected, and there is no much chance of any changes in this regard getting merged into upstream. The actual issue is providing the user with the possibility of setting custom lower SCUs for an Account, which is not intended, but that is something that would have to be fixed in the UI, and personally I can live with that as I know now I should not use that.

Apart from this I am a bit busy personally right now, so I wouldn't have time to work on this in the next days.

So I don't think I am going to invest some more time into the issue soon. If no one else is going to pick it up, it can be closed

jralls:
OK, Thanks for replying.

PR #431, opened 2018-11-01

rcalpha:
*) Fix VEB and VEF local-symbol
*) Add new currency for Venezuela (VES - Bolivar Soberano) since August 20th 2018
This is related to Bug 796927


fellen:
Commited, thanks!

PR #434, opened 2018-11-11
Decorate Python GetBalTaxValues to return tuples of GncCommodity and GncNumeric

c-holtermann:
Let GetBalTaxValues return a list of tuples (GncCommodity, GncNumeric) by using decorate_monetary_list_returning_function. This function needs an additional arg to work here. So add args and pass them through to original function. Add some doc as well.

I used this tuple solution because there was a function that did it this way (decorate_monetary_list_returning_function). I wonder if it makes more sense to create a GncMonetary class instead.

### DIFF
--- a/bindings/python/gnucash_business.py
+++ b/bindings/python/gnucash_business.py
@@ -374,3 +374,6 @@ methods_return_instance(Entry, entry_dict)
 Entry.decorate_functions(
     decorate_to_return_instance_instead_of_owner,
     'GetBillTo' )
+
+from gnucash.gnucash_core import decorate_monetary_list_returning_function
+Entry.decorate_functions(decorate_monetary_list_returning_function, 'GetBalTaxValues')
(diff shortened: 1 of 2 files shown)


jralls:
gnc_monetary is a class in [gnc-commodity.h](https://github.com/Gnucash/gnucash/blob/maint/libgnucash/engine/gnc-commodity.h#L1005) albeit with a very limited API and used only in imbalance calculations in Transaction.c. I think that's probably a poor design choice. There's a similar but unrelated gnc:monetary class in the reporting code--at least until Chris Lam realizes that it's not very useful and gets rid of it. ;-)
In both cases they're effectively equivalent to the tuple you've already got. But what's your use-case for exposing this to python ?

c-holtermann:
Exposing the class gnc_monetary would be more for completeness sake. I think it's useful to remain close to the c/c++ structure. If the use of this class is limited to few occurences then it's ok with me to stick with just passing a tuple in python and not explicitly creating that class. I just wanted to bring that question to discussion.

jralls:
Thanks!

c-holtermann:
You're welcome and thanks for accepting !

PR #435, opened 2018-11-17
Python example scripts: Update query for get_all_customers

federvieh:
Use same type of query in get_all_customers as in get_all_invoices


federvieh:
Just found out that the rest API uses the same query in getCustomers.

jralls:
Thanks. The master branch is where we develop the next major release. Since I doubt that you want to wait 2-3 years for your commit to appear in a release and since it's a simple change I took the liberty of merging it into maint instead.

PR #436, opened 2018-12-02
Some follow up changes after OK button PR

Bob-IT:
Geert suggested a couple of changes which I have implemented here namely..
CSV import alignment change,
A missing update that should of used a parent window parameter,
And the ability to double click on a file name for file selection.

There are a couple of other bug fixes that I think should be OK but as always will wait for feedback.

### DIFF
--- a/gnucash/gnome-utils/dialog-options.c
+++ b/gnucash/gnome-utils/dialog-options.c
@@ -2505,6 +2505,7 @@ gnc_option_set_ui_widget_text (GNCOption *option, GtkBox *page_box,
     value = gtk_text_view_new();
     gtk_text_view_set_wrap_mode(GTK_TEXT_VIEW(value), GTK_WRAP_WORD);
     gtk_text_view_set_editable(GTK_TEXT_VIEW(value), TRUE);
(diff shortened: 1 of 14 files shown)


Bob-IT:
Added a couple of new fixes that disallow tabs in text views. Bug 796944 is about the options text view that seems to be used for notes and CSS. The last one was for the account dialogue that I seemed to remember being reported but could not find the Bug till now, Bug795845, this is for the notes field, not sure, can be dropped if required.

fellen:
@Bob-IT I have reopened Bug 795845. Gnucash is no text processor and should use the tabulator for navigation.

Nice cleanup, BTW.

gjanssens:
A nice round of fixes, thanks a lot!
And no comments from me this time. You're really into it these days :)

fellen:
I took the freedom to close Bug 795845, 796886, 796944, 796948, 796960. ;-)

fellen:
@Bob-IT while testing I found the accept-tab issue also on the Notes field in the New/Edit Customer/Vendor dialogs. Can you look for them.

I am currently not sure what is the best way to handle issue: Remove the attribut in the glade files or in the C files, prbably defining adjusted classes of widgets.

PR #438, opened 2018-12-04
add CMAKE CXX compile flag -Wno-deprecated-declarations

c-holtermann:
As discussed on the mailing list compilation fails (on recent OpenSuse Tumbleweed) because of
https://gitlab.gnome.org/GNOME/glib/commit/7e5db31d36532274c2b2428ef9bace796928785e
as `g_type_class_add_private()` is deprecated. It's in use fairly often in the source.
`grep g_type_class_add_private * -R | grep ':'` leads to 68 files. But that is mainly c files. The only cpp ones are libgnucash/engine/Account.cpp and libgnucash/engine/qofinstance.cpp.


c-holtermann:
I think I was wrong here as the issue had already been adressed in
https://github.com/Gnucash/gnucash/commit/4c87dd05ec8f1da40292f3e2f0004418a573ab89#diff-af3b638bc2a3e6c650974192a53c7291
Maybe some temporary files had not been removed while I compiled.
I suppose I will redraw this PR.

PR #440, opened 2018-12-06
Fixes to the appdata file

kalev:
Rename the ID from org.gnucash.Gnucash to org.gnucash.GnuCash to match
flathub, and add a launchable ID to link it to the desktop file.


gjanssens:
Thanks. I have applied your patch.

I agree the ID should be GnuCash with capital C, although I disagree this should be because flathub does so :)
It should be because the GnuCash project recently decided independently (and actually rather by accident) that GnuCash should be the formal name of the application.

Your patch supports this, so I happily accepted it. So thanks again!

kalev:
Great, thanks! :)

PR #442, opened 2018-12-20
Some more fixes

Bob-IT:
I think they are all fairly straight forward which should be OK hopefully.
Not sure about the last two but it's the only way I could think off simply. Maybe the last one could get a list of split accounts, scan the open registers and then only try to refresh those registers or maybe the rollback should do it somehow.

For info, I can probably make changes up to Christmas.

### DIFF
--- a/gnucash/gnome-utils/dialog-account.c
+++ b/gnucash/gnome-utils/dialog-account.c
@@ -230,12 +230,13 @@ gnc_account_to_ui(AccountWindow *aw)
 
     string = xaccAccountGetColor (account);
 
-    if ((string == NULL) || (g_strcmp0 ("Not Set", string) == 0))
+    if (!string)
         string = DEFAULT_COLOR;
-    if (gdk_rgba_parse(&color, string))
-    {
-        gtk_color_chooser_set_rgba(GTK_COLOR_CHOOSER(aw->color_entry_button), &color);
-    }
+
+    if (!gdk_rgba_parse (&color, string))
+        gdk_rgba_parse (&color, DEFAULT_COLOR);
+
+    gtk_color_chooser_set_rgba(GTK_COLOR_CHOOSER(aw->color_entry_button), &color);
 
     commodity = xaccAccountGetCommodity (account);
     gnc_general_select_set_selected (GNC_GENERAL_SELECT (aw->commodity_edit),
@@ -380,12 +381,19 @@ gnc_ui_to_account(AccountWindow *aw)
 
     gtk_color_chooser_get_rgba(GTK_COLOR_CHOOSER(aw->color_entry_button), &color );
     string = gdk_rgba_to_string(&color);
+
     if (g_strcmp0 (string, DEFAULT_COLOR) == 0)
-        string = "Not Set";
+        string = NULL;
 
     old_string = xaccAccountGetColor (account);
-    if (g_strcmp0 (string, old_string) != 0)
-        xaccAccountSetColor (account, string);
+
+    if (!string && old_string)
+        xaccAccountSetColor (account, ""); // remove entry
+    else
+    {
+        if (g_strcmp0 (string, old_string) != 0)
+            xaccAccountSetColor (account, string); // update entry
+    }
 
     commodity = (gnc_commodity *)
                 gnc_general_select_get_selected (GNC_GENERAL_SELECT (aw->commodity_edit));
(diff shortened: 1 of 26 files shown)


gnucash/gnome/gnc-split-reg.c
```diff
@@ -1305,6 +1305,9 @@ gsr_default_delete_handler( GNCSplitReg *gsr, gpointer data )
     trans = xaccSplitGetParent(split);
     cursor_class = gnc_split_register_get_current_cursor_class (reg);
 
+    /* test for blank_split reference pointing to split */
+    gnc_split_register_change_blank_split_ref (reg, split);
+
```
jralls: It would be better to separate the test from the action so that this line becomes
```C
    if (gnc_split_register_is_blank_split (reg, split))
        gnc_split_register_change_blank_split_ref (reg, split);
```
Or change the name of the function to something that expresses the real outcome, maybe ```gnc_split_register_ensure_not_blank_split_ref (reg, split)```.

gnucash/gnome-utils/dialog-account.c
```diff
@@ -230,12 +230,13 @@ gnc_account_to_ui(AccountWindow *aw)
...
 
-    if ((string == NULL) || (g_strcmp0 ("Not Set", string) == 0))
+    if (string == NULL)
         string = DEFAULT_COLOR;
-    if (gdk_rgba_parse(&color, string))
-    {
-        gtk_color_chooser_set_rgba(GTK_COLOR_CHOOSER(aw->color_entry_button), &color);
-    }
+
+    if (gdk_rgba_parse (&color, string) == FALSE)
```
gjanssens: This is typically written as `if (!gdk_rgba_parse (&color, string))`

gnucash/gnome/gnc-plugin-page-budget.c
```diff
@@ -335,6 +335,7 @@ gnc_plugin_page_budget_init (GncPluginPageBudget *plugin_page)
     priv->fd.visible_types = -1; /* Start with all types */
     priv->fd.show_hidden = FALSE;
     priv->fd.show_zero_total = TRUE;
+    priv->fd.show_unused = TRUE;
```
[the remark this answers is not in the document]
Bob-IT: I have dropped this.

gnucash/import-export/csv-imp/assistant-csv-trans-import.cpp
```diff
@@ -1501,9 +1504,18 @@ make_column_header_model (bool multi_split)
         if (sanitize_trans_prop(col_type.first, multi_split) == col_type.first)
         {
             GtkTreeIter iter;
+            gchar *label;
             gtk_list_store_append (combostore, &iter);
+
+            if (formal && (col_type.first == GncTransPropType::DEPOSIT))
+                label = _("Debit");
+            else if (formal && (col_type.first == GncTransPropType::WITHDRAWAL))
+                label = _("Credit");
```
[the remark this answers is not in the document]
Bob-IT: I have dropped this.

gnucash/gnome-utils/dialog-account.c
Bob-IT: I have a tendency to write the long version as it is clearer and then try to reduce then I have something working but missed this.

gnucash/gnome/gnc-split-reg.c
Bob-IT: Changed this to the first option as it makes more sense.

gnucash/import-export/csv-imp/assistant-csv-trans-import.cpp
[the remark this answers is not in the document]
jralls: Let's not use "inverted". That usually means "divided into 1" e.g. 1/2 is the inverse of 2. I think accountants would use "reversed" for applying the amount with the sign changed.

That said, we absolutely *can* assume that the data follow Debit/Credit, Deposit/Withdrawal, Charge/Payment, or whatever convention because *that's what the user is telling us by selecting the column heading*. If there's a negative amount in the Credit column then the user should expect it to be the same as putting a negative amount in that column in the register. The tricky bit for infomal labels is getting the right set, which depends upon the account being imported into. Hardwiring Deposit/Withdrawal won't be right if the account isn't a bank account.
Now if the importer doesn't work like that but instead wants a signed amount that it puts in the debit or credit column depending on the sign then the heading should be "Amount" as you say.

Bob-IT:
I have made some changes that hopefully are OK. I have dropped the two commits Geert had concerns about. Will look at the budget one in new year. I have also added a routine to remove all account color "Not Set " slots which should only run once on a book.

gnucash/import-export/csv-imp/assistant-csv-trans-import.cpp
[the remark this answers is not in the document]
jralls: Resolving since the commit is dropped. @gjanssens should we change the Debit and Credit to Amount and Amount (Reversed) reflecting what the importer is really doing?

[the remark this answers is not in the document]
gjanssens: Back to your previous comment first: the tricky bit of selecting the right informal label set. We simply can't because the csv importer is generic enough it can import transactions from and to multiple accounts. Importing into a single account is only a subset of that (which is hence also supported). In addition, assume we're importing into one single account, that account is not always known in the preview page, so we won't always be able to determine the proper set of informal labels in time to be useful.

So in a way Debit and Credit are better as they are always correct. But they also frequently lead to confusion as many users don't use these terms (and frankly I also continue to struggle with them as I don't use them enough). So I do vote for changing Deposit and Withdrawal to Amount and Amount (Reversed).

libgnucash/engine/Account.h
```diff
@@ -1454,6 +1454,11 @@ void gnc_account_delete_map_entry (Account *acc, char *head, char *category,
  */
 void gnc_account_delete_all_bayes_maps (Account *acc);
 
+/** Remove color slots that have a "Not Set" value
+ *  This should only be run once on a book
+ */
+void gnc_account_fix_color_slots (QofBook *book);
```
jralls: This would fit best as part of xaccAccountScrubKvp except that that isn't presently called in the SQL backend. Generally the scrubs need some cleanup; calls to them are splattered around instead of consolidated and some of them are probably obsolete. To facilitate including this in an eventual cleanup let's call it xaccAccountScrubColorNotSet.

jralls: Yours needs to run every session too in case a user edits an account with an older version of GnuCash.

I'll merge it and make any more required changes before the release.

PR #444, opened 2018-12-26
 Appdata: remove old property 

City-busz:
This fixes the following notice detected by `appstreamcli validate cherrytree.appdata.xml`:
```
I - gnucash.appdata.xml.in:org.gnucash.GnuCash:3
    The id tag for "org.gnucash.GnuCash" still contains a 'type' property, probably 
    from an old conversion.
```
And the metainfo files should be installed into /usr/share/metainfo.

See: https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#spec-component-location

### DIFF
--- a/gnucash/gnome/CMakeLists.txt
+++ b/gnucash/gnome/CMakeLists.txt
@@ -180,7 +180,7 @@ endif()
 
 add_custom_target(gnucash-appdata ALL DEPENDS ${CMAKE_CURRENT_BINARY_DIR}/gnucash.appdata.xml)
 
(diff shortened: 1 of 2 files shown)


gjanssens:
Thanks for the patch.

The link you passed mentions the new metadata location is well supported by older distros such as Centos/RHEL 7 and Ubuntu 16.04LTS. It doesn't mention Ubuntu 14.04LTS which we also still support. However I doubt appdata was already in active use in that distro. So I have merged your fixes. They will appear in the next release (3.4).

PR #445, opened 2018-12-28
Maint scheme progress

christopherlam:
A few commits that *can* be merged in maint. Comments welcome!

1. remove any dependency on (ice-9 streams) by pattern-streamer which was using regex for parsing html. ugh. https://stackoverflow.com/questions/1732348/regex-match-open-tags-except-xhtml-self-contained-tags -- use xml parsing via guile sxml module
2. ~fix income-gst-report remittance label~
3. upgrade some tests to use srfi-64
4. deprecate some unused functions options-utilities.scm
5. deprecate collectors.scm and report-collectors.scm and cease testing them
6. remove duplicated test-test-extras.scm


fellen:
I cherry.picked commit a1f1e14. Thanks!

PR #447, opened 2018-12-31
Re-use user locale when formatting time - 10% improvement save speed

chriscarson60187:
This 2-liner saves 10-20% of the user CPU for XML save.
Note that the benefit in XML save time is likely temporary given
jralls' effort to switch to UTC for XML save data.

### DIFF
--- a/libgnucash/engine/gnc-datetime.cpp
+++ b/libgnucash/engine/gnc-datetime.cpp
@@ -425,11 +425,21 @@ std::string
 GncDateTimeImpl::format(const char* format) const
 {
     using Facet = boost::local_time::local_time_facet;
+    static std::locale cachedLocale;
+    static bool cachedLocaleAvailable = false;
     std::stringstream ss;
+
+    if(!cachedLocaleAvailable)
+    {
+        cachedLocale = std::locale("");
+	cachedLocaleAvailable = true;
+    }
+
     //The stream destructor frees the facet, so it must be heap-allocated.
     auto output_facet(new Facet(normalize_format(format).c_str()));
     // FIXME Rather than imbueing a locale below we probably should set std::locale::global appropriately somewhere.
-    ss.imbue(std::locale(std::locale(""), output_facet));
+    // At that point the use of cachedLocale mechanism should be removed.
+    ss.imbue(std::locale(cachedLocale, output_facet));
     ss << m_time;
     return ss.str();
 }


jralls:
Nope. static variables are initialized before calling main() so the locale will be frozen at what's in the user's environment... something that doesn't even exist on Windows. There's a pile of code in gnucash-bin.c that gets the locale from the OS on Macs and Windows as well as reading the environment file in case the user wants to run GnuCash in a different locale from the rest of the system.  The comment is correct: Use std::locale::global(). The catch is that you need to find a good place to initialize it after the locale-setting code in main(). That's why the comment is there.

Style note: "myfoo" is for poorly written introductory texts. Don't write it in real code.

chriscarson60187:
Thanks for the insights.  I have been keeping my changes as localized as possible precisely because I am early on in my understanding of gnucash internals.  The switch to a static was obviously overly minimal.

The change is worthwhile as long as the format routine is used to write xml. The 10% performance improvement estimate is quite conservative.

So your advice:  Should I:
1) Drop this, or
2) Re-do with caching locale on first use, or
3) Go after the underlying problem by setting global locale in main?  I can set up a test environment on windows but not mac os/x.

chriscarson60187:
For completeness I have pushed a commit with the cache-on-first-use approach.

Based on an average of 3 runs of an init/load and 3 runs of an init/load/save scenario
in my environment:
User mode CPU for the version without this change:  10.2 seconds
User mode CPU for the version with this change: 7.6 seconds

christopherlam:
The test-transaction failure is embarrassing. I'll fix it asap.

chriscarson60187:
Thank you!  I was baffled. Is there a button I can push to re-run the CI tests?

jralls:
No, only developers have access to the CI controls. Besides, @christopherlam doesn't have push either, so I have to merge and push his PR before the CI tests will pass again.

The first-use cache is acceptable, though obviously setting std::locale::global somewhere would be better still. The backends are C++ so they might be a good place.

Please rebase your branch into a single commit.

jralls:
I've merged @christopherlam's PR. If you rebase this PR onto current ```maint``` (squashing it into 1 commit in the process) and force-push that will get CI to re-test and it should come up clean.

chriscarson60187:
But wait...  we may have an oh-poop moment here.

@jralls  wrote: _"Nope. static variables are initialized before calling main() so the locale will be frozen at what's in the user's environment..."_  But my understanding is that's exactly what locale("") returns, the environment frozen at what's in the user's environment, regardless of where it's run  The reason it's slow is that it rebuilds the locale by scanning environment variables.

If the design intent was to re-use the locale created by the setlocale(...) calls in gnucash/gnucash-bin.c, it looks like the edited routine was wrong from the start.  Those setlocale() changes are, as I read the documentation,  saved in the C locale, so the call should have been to locale("C").  And that, I've already established, is quite fast.

If agreed, I could make that change, but I'm reluctant to do that because it theoretically changes the contents of the saved XML file.  I'm not sure I understand all of those impacts.

(I studied the Gnu standard library code that implements locale("") before writing this.)

jralls:
No, because the runtime environment can--and is, in gnucash-bin.c--be changed with calls to setenv().
No, the "C" locale is the libc default locale. Setting it turns off localization. That's probably why std::locale("C") is so fast.

Ideally the C++ standard library would set std::locale::global from C setlocale() calls so that a call to std::locale() would get the C-runtime (*not the "C"*) locale, but it works only the other way: std::locale::global(locale) will set the C-runtime locale. Until we convert gnucash-bin.c to C++ so we can set std::locale::global there what you've got is the best we can do.

But for writing XML files we don't want to use this function because we don't want to localize the dates in the XML files (IOW we *do* want to use the "C" locale, but we can do that by simply not using localization at all) and we want something faster than the generic boost::date_time formatter. Christian and I discussed this a bit in the gnucash-devel thread that lead you here. I hope I'll get back to that later today.

jralls:
Thanks!

PR #448, opened 2019-01-01
Scheme progress

christopherlam:
includes a fix for a embarrassing test-transaction test hard-coded to 2018. it's now 2019.


jralls:
Thanks for the quick remediation.

PR #449, opened 2019-01-07
Fixes for time64 setter/getters in Python examples

rgreid:
The Python examples were updated for Python3, but not the breaking changes in the bindings.  From the 3.0 release notes:

> Replaced Timespec with time64 in the Scheme and Python bindings, introducing many new C time64 functions to accomodate the replacement.

This PR fixes the examples to use the time64 bindings.

### DIFF
--- a/bindings/python/example_scripts/priceDB_test.py
+++ b/bindings/python/example_scripts/priceDB_test.py
@@ -38,7 +38,7 @@ value = latest.get_value()
 pl = pdb.get_prices(arm,gbp)
 for pr in pl:
    source = pr.get_source()
-   time = pr.get_time()
+   time = pr.get_time64()
    v=pr.get_value()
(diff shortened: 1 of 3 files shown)


gjanssens:
Thanks for your fixes. They look good to me.
Unfortunately I can't easily test them as I currently don't have a book with price information in it and it's been a while since I have looked at the python code.

@c-holtermann Any thoughts on this PR?

c-holtermann:
It's a necessary change as the get/set_time() methods are gone since fee589b28cb0c554109cb934c24fa573c1dd9757
Thanks for the fix !

c-holtermann:
The transition from python 2 to python 3 is not complete. I started a bug for that: https://bugs.gnucash.org/show_bug.cgi?id=796863.
But anyway the change from time to time64 is not related to that from python2 to python3.

PR #450, opened 2019-01-10
scheme progress - refactoring only

christopherlam:
Various REFACTORING commits. 
- ~fix for https://bugs.gnucash.org/show_bug.cgi?id=773169~
- ~transaction report was disabling sorting/indenting inappropriately for date-grouped sortkeys, fix~
- ~transaction report - if display-subtotal = true, hide-transactions = true, then also hide the main table.~
- ~reconciliation-report will now hopefully display a header similar to the reconciliation tool.~
- option-getter will warn if trying to retrieve old sectionname/optionname
- register.scm and report.scm cleanup
- gnc:account-get-balances-at-dates upgrade
- further deprecation of unused/silly functions/modules

~EDIT: I had to remove rewrite make-date-list tests to pass in all time zones~


christopherlam:
After IRC discussion, and some tinkering, I've designed `gnc:make-date-list` and `gnc:make-interval-date-list` to handle month/quarter/halfyear/year deltas separate from day/week deltas.

Premise: for the larger deltas, any D/M/Y will be expected to have M modified by *delta* months. if the resulting gnc_localtime(gnc_mktime(M+delta))'s month is not equal to M+delta, then we decrease D by 1 and try again, until the new month is M+delta. And the M+delta is always calculated from the original startdate; so, a startdate of 31/08/2019, quarterly deltas, will reach 31/11/2019 (fixed to 30/11/2019), 31/02/2020 (fixed to 29/02/2020), and 31/05/2020 and so on. I think this is the most intuitive outcome we'll get, and I think can be considered a bug fix.

libgnucash/app-utils/config-var.scm
```diff
@@ -33,6 +33,11 @@
 ;;; permanent, and if they leave the variable value different from the
 ;;; default, should be saved to the auto configuration file.
 
+;; this module should be deprecated.
+;; (issue-deprecation-warning "config-var.scm must be deprecated.")
+;; dummy line above to be searchable in master
+;; its defined functions are no longer used in code.
+
```
gjanssens: Hmm, I don't understand this part: why can't we add the (issue-deprecation-warning ...) message to each of the individual gnc:config-var-* functions ?
Is this because just compiling that code would already trigger the deprecation notices ?

libgnucash/engine/engine-utilities.scm
```diff
@@ -93,33 +93,31 @@
   (hashx-set! account-hash account-assoc table account value))
 
 ;; Splits
-(export split-same?)
-(export split-in-list?)
-
+(export split-same?)                    ;deprecated
+(export split-in-list?)                 ;deprecated
 (define (split-same? s1 s2)
+  (issue-deprecation-warning "split-same? is deprecated.")
```
gjanssens: On a general note: it would be *very* helpful to function consumers if your deprecation warnings hinted at the alternative.
So make the message something like

`"split-same? is deprecated. Please use xyz instead.`

While no doubt the replacement is blindingly obvious to you, it certainly isn't for me :)
And we want our code to be conductive for newcomers also...

libgnucash/engine/test/test-split.scm
```diff
@@ -1,28 +0,0 @@
...
-	 (wallet-account (cdr (assoc "Wallet" account-alist)))
-	 (tx1 (env-create-transaction env today bank-account wallet-account 20/1))
-	 (tx2 (env-create-transaction env today bank-account expense-account 10/1))
-	 (splits-tx1 (xaccTransGetSplitList tx1))
-	 (splits-tx2 (xaccTransGetSplitList tx2)))
-    (and (split-in-list? (first splits-tx1) splits-tx1)
-	 (split-in-list? (second splits-tx1) splits-tx1)
-	 (not (split-in-list? (first splits-tx1) splits-tx2))
-	 (not (split-in-list? (second splits-tx1) splits-tx2))
-	 (not (split-in-list? (first splits-tx1) '())))))
```
gjanssens: Again, deprecated functions still need to work reliably until they are finally removed. So tests should only be removed when the deprecated functions are.

libgnucash/engine/test/test-account.scm
```diff
@@ -1,50 +0,0 @@
...
-(define (test-account-list-predicate-generic predicate)
-  (let* ((env (create-test-env))
-	 (account-alist (env-create-test-accounts env))
-	 (bank-account (cdr (assoc "Bank" account-alist)))
-	 (wallet-account (cdr (assoc "Wallet" account-alist)))
-	 (other-account (cdr (assoc "Other" account-alist)))
-	 (bank-or-wallet? (predicate (list bank-account wallet-account))))
-    (and (bank-or-wallet? bank-account)
-	 (bank-or-wallet? wallet-account)
-	 (not (bank-or-wallet? other-account)))))
```
gjanssens: Same story about premature removal...

libgnucash/app-utils/config-var.scm
christopherlam: No, it wouldn't trigger... it's fine to do individually. Except I don't understand this module which is unused apart from a comment by linas in 1999 in prefs.scm:20 -- I wouldn't know alternative either since this module is completely unused.

christopherlam:
Shall I make amends and repush? Or better add reverting commits?

gjanssens:
In the past when I reviewed @jralls ' work on his request I saw he just added followup commits, mentioning these were to fix review remarks. That's an elegant way of keeping history and perhaps what we should stick to.
I used to be a rebase fan, but perhaps that's my svn origins showing. In the git world it seems it's generally avoided unless for work that was never published...

gjanssens:
And on second review, I believe the config-vars is sufficiently old and unused you can delete it just like that as far as I'm concerned. I believe it predates the use of gconf and now gsettings.

jralls:
> In the past when I reviewed @jralls ' work on his request I saw he just added followup commits, mentioning these were to fix review remarks. That's an elegant way of keeping history and perhaps what we should stick to.

But depending on the complexity of the PR I might rebase the whole thing into a logical sequence after all of the reviews are over.

Reverts are something of a special case, particularly if its removing then restoring (or adding then removing) whole functions. That makes for confusing history, so IMO they usually call for rebasing the change away instead of reverting it.

gjanssens:
Ok. It's probably better to have clear and concise history than to be able to track git PR comments. So you can do a cleanup rebase after review.

gnucash/report/standard-reports/register.scm
```diff
@@ -596,10 +586,10 @@
   (let* ((table (gnc:make-html-table))
          (used-columns (build-column-used options))
          (width (num-columns-required used-columns))
-         (multi-rows? (reg-report-journal?))
-         (ledger-type? (reg-report-ledger-type?))
-         (double? (reg-report-double?))
          (leader (splits-leader splits))
+         (multi-rows? (opt-val "__reg" "journal"))
+         (ledger-type? (opt-val "__reg" "ledger-type"))
+         (double? (opt-val "__reg" "double"))
```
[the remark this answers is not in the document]
christopherlam: Ok I'll remove this commit.

gjanssens:
So I got lost here. What's the status of this PR ? Did you push all of this ? I can't seem to find all of the commits on maint...

christopherlam:
This branch is being deconstructed projectwise and more tests being added in-between. I'll close it eventually.

If I rebase to current maint, the entire commit history will disappear from this PR.

christopherlam:
This has now been merged safely. Numerous tests have been added.

gjanssens:
Yay!

PR #451, opened 2019-01-14
Fixes for bugs 797029 7970331 797030 in Import Customers & Vendors

ekinonnakapito:
[797029 Import Customer & Vendors: blank name and company in import data row crashes GnuCash](https://bugs.gnucash.org/show_bug.cgi?id=797029)
[797031 Import customer & Vendors: import can create customer and vendors without address data](
https://bugs.gnucash.org/show_bug.cgi?id=797031)
[797030 Import Customer & Vendors: several issues with the matching of data rows](https://bugs.gnucash.org/show_bug.cgi?id=797030)

### DIFF
--- a/gnucash/import-export/customer-import/dialog-customer-import-gui.c
+++ b/gnucash/import-export/customer-import/dialog-customer-import-gui.c
@@ -109,7 +109,7 @@ gnc_plugin_customer_import_showGUI(GtkWindow *parent)
 
     gtk_window_set_transient_for (GTK_WINDOW (gui->dialog), parent);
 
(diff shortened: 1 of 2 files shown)


gjanssens:
Hey, thanks for the PR. I'm a bit busy right now, but I'll try to check it out over the weekend.

gjanssens:
Thanks for the PR.

The code changes look good so I have committed them.

However for future work, I want to ask you to try to stick with one bugfix per commit. That makes looking back in the future much easier and it helps should we require to debug other (new or old) bugs in the same code area.

As the PR is fairly small I have chosen exceptionally to merge it anyway as is in this case.

PR #453, opened 2019-01-20
Bug796965 - Change the default to show unused accounts

Bob-IT:
Another attempt at this one...
The first commit changes the default for budgets to show unused accounts so it matches the Account Treeview.

With the addition of two functions that will save and restore the account filter settings to the state file the budget will now remember these settings when the page is closed on exit. They are still being saved as part of the page saving routines as well for compatibility and in 4.x these settings could be removed from the those routines.

I have checked in version 2.6.21 that these settings are ignored and left as is but to achieve this I needed to change the key names to include an underscore so that version would not crash, this was fixed in 3.2.

The last commit adds the same capability to the 'Accounts TreeView' so if you do open a new accounts page it will have your preferred default. It doesn't take much imagination in expanding this to saving different account trees based on account type filtering if that would be of use.

### DIFF
--- a/gnucash/gnome-utils/gnc-tree-view-account.c
+++ b/gnucash/gnome-utils/gnc-tree-view-account.c
@@ -2286,6 +2286,7 @@ account_filter_dialog_create(AccountFilterDialog *fd, GncPluginPage *page)
     LEAVE(" ");
 }
 
+// page state section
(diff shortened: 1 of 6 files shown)


gjanssens:
That does it. Thanks, @Bob-IT !

PR #454, opened 2019-02-01
scheme bugfixes

christopherlam:
Some quick bugfixes with associated tests, separate from ~the difficult PRs~ refactoring commits.

### DIFF
--- a/gnucash/gnome-utils/gnome-utils.scm
+++ b/gnucash/gnome-utils/gnome-utils.scm
@@ -41,3 +41,27 @@
 
 (load-from-path "gnc-menu-extensions")
 
+;; this function will receive 1 boolean argument, and can be used for
+;; any UI init/shutdown routines. For now it will set the
+;; gnc:ui-warn/error/msg tracefile routines to display dialog messages
+;; in addition to tracefile logging.
+(define-public gnc:set-ui-status
+  (let ((save-warn gnc:gui-warn)
+        (save-error gnc:gui-error)
+        (save-msg gnc:gui-msg))
+    (lambda (status)
+      (cond
+       (status
+        (set! gnc:gui-warn (lambda (constr guistr)
+                             (save-warn constr guistr)
+                             (gnc-warning-dialog '() guistr)))
+        (set! gnc:gui-error (lambda (constr guistr)
+                              (save-error constr guistr)
+                              (gnc-error-dialog '() guistr)))
+        (set! gnc:gui-msg (lambda (constr guistr)
+                            (save-msg constr guistr)
+                            (gnc-info-dialog '() guistr))))
+       (else
+        (set! gnc:gui-warn save-warn)
+        (set! gnc:gui-error save-error)
+        (set! gnc:gui-msg save-msg))))))
--- a/libgnucash/app-utils/date-utilities.scm
+++ b/libgnucash/app-utils/date-utilities.scm
@@ -430,8 +430,14 @@
 (define (gnc:get-absolute-from-relative-date date-symbol)
   (let ((rel-date-data (hash-ref gnc:relative-date-hash date-symbol)))
     (if rel-date-data
-       ((gnc:reldate-get-fn rel-date-data))
-       (gnc:error "Tried to look up an undefined date symbol"))))
+        ((gnc:reldate-get-fn rel-date-data))
+        (let* ((msg (_ "Tried to look up an undefined date symbol \
+'~a'. This report was probably saved by a later version of GnuCash. \
+Defaulting to today."))
+               (conmsg (format #f msg date-symbol))
+               (uimsg (format #f (_ msg) date-symbol)))
+          (gnc:gui-warn conmsg uimsg)
+          (current-time)))))
 
 (define (gnc:get-relative-date-strings date-symbol)
   (let ((rel-date-info (hash-ref gnc:relative-date-hash date-symbol)))
--- a/libgnucash/app-utils/date-utilities.scm
+++ b/libgnucash/app-utils/date-utilities.scm
@@ -215,32 +215,78 @@
 (define (gnc:time64-ge-date t1 t2)
   (gnc:time64-le-date t2 t1))
 
-;; Build a list of time intervals. 
+(define (incdate-months date nmonths)
+  (let* ((new-date (gnc-localtime date))
+         (newmonth (+ (tm:mon new-date) nmonths))
+         (new-month-proper (floor-remainder newmonth 12))
+         (new-year-proper (+ (tm:year new-date) (floor-quotient newmonth 12))))
+    (set-tm:year new-date new-year-proper)
+    (set-tm:mon new-date new-month-proper)
+    (let loop ((new-mday (tm:mday new-date)))
+      (set-tm:mday new-date new-mday)
+      (let ((res (gnc-mktime new-date)))
+        (if (= new-month-proper (tm:mon (gnc-localtime res)))
+            res
+            (loop (1- new-mday)))))))
+
+;; Build a list of time intervals.
 ;;
 ;; Note that the last interval will be shorter than <incr> if
 ;; (<curd>-<endd>) is not an integer multiple of <incr>. If you don't
 ;; want that you'll have to write another function.
-(define (gnc:make-date-interval-list current-date end-date increment)
-  (if (< current-date end-date)
-      (let ((next-date (incdate current-date increment)))
-        (if (< next-date end-date)
-            (cons (list current-date (decdate next-date SecDelta) '())
-                  (gnc:make-date-interval-list next-date end-date increment))
-            (cons (list current-date end-date '())
-                  '())))
-      '()))
-  
+(define (gnc:make-date-interval-list startdate enddate incr)
+  (define month-delta
+    (assv-ref MonthDeltas incr))
+  (let loop ((result '())
+             (date startdate)
+             (idx 0))
+    (cond
+     ((>= date enddate)
+      (reverse result))
+     (month-delta
+      (let* ((curr (incdate-months startdate (* month-delta idx)))
+             (next (incdate-months startdate (* month-delta (1+ idx)))))
+        (loop (cons (list curr
+                          (if (< next enddate)
+                              (decdate next SecDelta)
+                              enddate))
+                    result)
+              next
+              (1+ idx))))
+     (else
+      (let ((next (incdate date incr)))
+        (loop (cons (list date
+                          (if (< next enddate)
+                              (decdate next SecDelta)
+                              enddate))
+                    result)
+              next
+              (1+ idx)))))))
+
 ;; Build a list of times.  The dates are evenly spaced with the
 ;; stepsize 'incr'. If the difference of 'startdate' and 'enddate' is
 ;; not an integer multiple of 'incr', 'enddate' will be added as the
 ;; last element of the list, thus making the last interval smaller
 ;; than 'incr'.
 (define (gnc:make-date-list startdate enddate incr)
-  (if (< startdate enddate)
-      (cons startdate
-            (gnc:make-date-list (incdate startdate incr)
-                                enddate incr))
-      (list enddate)))
+  (define month-delta
+    (assv-ref MonthDeltas incr))
+  (let loop ((result '())
+             (date startdate)
+             (idx 0))
+    (cond
+     ((>= date enddate)
+      (reverse (cons enddate result)))
+     (month-delta
+      (let* ((curr (incdate-months startdate (* month-delta idx)))
+             (next (incdate-months startdate (* month-delta (1+ idx)))))
+        (loop (cons curr result)
+              next
+              (1+ idx))))
+     (else
+      (loop (cons date result)
+            (incdate date incr)
+            (1+ idx))))))
 
 ; A reference zero date - the Beginning Of The Epoch
 ; Note: use of eval is evil... by making this a generator function, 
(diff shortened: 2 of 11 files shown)


christopherlam:
I've pulled the book-fy-end property commit from #450 to this PR which contains mainly bugfixes.

libgnucash/app-utils/business-prefs.scm
```diff
@@ -125,6 +125,48 @@
...
+
+  (reg-option
+   (gnc:make-date-option
+    gnc:*book-accounting-period-label* "Custom End of Financial Year"
+    "a2"
+    "Please choose the custom end of financial year. Please be aware \
+only the day and month are used. The year stated is unused."
+    (lambda () (cons 'absolute (current-time)))
+    #f 'absolute #f))
+
```
gjanssens: With this commit the user now has two ways to specify an accounting period:

- via book options in File->Properties
- user-wide via Edit->Preferences

IMO this is a good recipe for confusion.

I do prefer the book options method. It's how it should have been implemented in the first place. However I think that means we need to migrate away from the user-wide option. Which in turn means backwards incompatible changes, so this should go on master to be handled properly.

gjanssens:
Other than the per-book accounting period all other commits are good to go imo

christopherlam:
Ok; so that you don't lose place in reviewing commits; I'll *stop* rebasing on maint.
The 9168ad9 is meant to handle backward compatibility and unfortunately does it silently: only a gnc:warn is emitted. Is this good practice? If all good to go perhaps best to cherrypick the rest? Or you prefer I remove the book-accounting and repush?

gjanssens:
Silently falling back to a default in case of backwards compatibility issues is not ideal, but probably better than a crash.
Even better would be a mechanism to pop up a message box to the user. Of course the tricky part here is you can't define a gui anywhere in libgnucash.

What you can define in libgnucash is a generic "messenger" function object. By default it simply writes a message to the log or stdout or whatever. Make it so that the gui initialization code alters its functionality such that it pops up a gui message box instead.

Something like this in pseudo guile code:

```
; Somewhere in engine initialization code
; This function should probably be added to the global scope
(define (messenger)
  (print message to file))

; Somewhere in the gui initialization routines
(set! (messenger)
  (pop-up-dialog with message))

; Anywhere you'd want to inform the user
(messenger "message")
```

It can be refined to support several message types, like WARN, DEBUG, ERROR,...

gjanssens:
As for rebasing, better not indeed for now to save history. I'm hoping you can handle the merge yourself one of these days :)
For now you can push a reverting commit perhaps.

christopherlam:
OK, I'll think about your solution which sounds good. Meanwhile, perhaps cherry picking everything except the last two commits will be useful? I'll move book accounting to master.

christopherlam:
Ok I've managed to make the gui warning dialog work;

My view is that `(gnc:warn) (gnc:error) (gnc:msg)` can be redefined upon gnome initialization to output to GUI instead of console (or it can do both). It works well for me! Loading a saved-report "from the future" triggers the warning dialog appropriately. If the gui is not loaded then it reverts to console. I've succeeded by amending gnc-gnome-utils.c and augmenting `gnc_ui_start_event_loop`.

There's lots of gnc:warn gnc:error all over, left gratuitously by various coders for their own internal use, never removed. I'll go through and remove them all or make sure they are suitable for GUI.

Views?

christopherlam:
See last commit which will convert all gnc:* to trigger gui warnings :-) so gnc:warn/error/msg will be *much* more visible. I think this is the most appropriate method to inform users!

gjanssens:
I'm glad you got that working. However gnc:warn and friends are not meant to inform the user! Those are meant to generate a trace log, the "trace file" we often ask the users in case of issues. It would reduce the usefulness of the trace file if those entries would be removed.
So it would be better to make an additional set of functions specifically to display messages to the user.

christopherlam:
> I'm glad you got that working. However gnc:warn and friends are not meant to inform the user! Those are meant to generate a trace log, the "trace file" we often ask the users in case of issues. It would reduce the usefulness of the trace file if those entries would be removed.

Ok personally I felt the reports are far too wordy (e.g. I think the barchart will gnc:warn the whole script!) and needed trimming. Will repush soon.

christopherlam:
I think this is now good.
To test I'd suggest save a new saved-report, modify the saved-reports file to modify a multichoice to an unknown symbol, and try running it...

christopherlam:
I think this finally works. I don't have ability to test on ubuntu 14.04 unfortunately.
On ubuntu disco 19.04 it works well.

christopherlam:
Unless someone objects I'll try my first formal github merge tomorrow. I've removed book-accounting-period commit. Bugfixes only.

gnucash/gnome-utils/gnome-utils.scm
```diff
@@ -41,3 +41,23 @@
 
 (load-from-path "gnc-menu-extensions")
 
+(define-public gnc:set-ui-status
+  (let ((save-warn gnc:gui-warn)
+        (save-error gnc:gui-error)
+        (save-msg gnc:gui-msg))
+    (lambda (status)
```
gjanssens: I'm a bit puzzled how "status" gets set. Can you explain this a bit ?

I see you're calling scm_call_1 ... in in gnc_ui_start_event_loop with an SCM boolean, but I can't understand how this eventually ends up in the "status" variable here.

libgnucash/app-utils/options.scm
```diff
@@ -18,6 +18,12 @@
 ;; Boston, MA  02110-1301,  USA       gnu@gnu.org
 (use-modules (ice-9 regex))
 
+(define (rpterror-earlier type newoption fallback)
+  (let ((msg (format #f "This report was saved using a later version of GnuCash. \
+One of the newer ~a options '~a' is not available, fallback to the option \
+'~a'." type newoption fallback)))
+    (gnc:gui-warn msg)))
```
gjanssens: This is a message that will be shown to the end user so it should be translatable...

libgnucash/app-utils/date-utilities.scm
```diff
@@ -430,8 +430,12 @@
...
   (let ((rel-date-data (hash-ref gnc:relative-date-hash date-symbol)))
     (if rel-date-data
-       ((gnc:reldate-get-fn rel-date-data))
-       (gnc:error "Tried to look up an undefined date symbol"))))
+        ((gnc:reldate-get-fn rel-date-data))
+        (let ((msg (format #f "Tried to look up an undefined date symbol '~a'. \
+This report was probably saved by a later version of GnuCash. Defaulting \
+to today." date-symbol)))
+          (gnc:gui-warn msg)
+          (current-time)))))
```
gjanssens: This message should be translable as well.

gjanssens: The warning is no longer printed to the trace file as suggested by the commit message. Was that deliberate ?

gjanssens:
Only a few remarks left :-)

gnucash/gnome-utils/gnome-utils.scm
christopherlam: It's a lispy closure.

1. Two ways of defining a function in guile. `(define (f x) ...)` or `(define f (lambda (x) ...))` These two are *exactly* the same. `x` is hopefully useful in `f`.

2. Any scheme expression can create a local environment using `(let ((vars..)) ...)` - I use it extensively to create intermediate values instead of long complicated expressions; vars are only visible within the let environment, and *discarded* when the (let) is finished.

Combining 1 and 2 leads to
```
(define gui-set
  (let ((old-fn gnc:gui-warn))
    (lambda (status)
      (if status (set! gnc:gui-warn ...)
                 (set! gnc:gui-warn old-fn))
      ...)))
```
whereby gui-set is a *function* (see the lambda) being created within the local environment containing old-fn which is accessible only *within* the `(let)` environment. So gui-set is now a lambda, accepting 1 argument `status` which receives #t/#f, and uses it to modify *global* gnc:gui-warn and can also save it in *local* old-fn.

Very schemey and compact. Avoids using global variables. No tricks, no pythonic generators. Similar to https://www.gnu.org/software/guile/manual/html_node/Shared-Variable.html#Shared-Variable

libgnucash/app-utils/date-utilities.scm
christopherlam: It should... in the redefinition of `gnc:gui-warn` does call the old `gnc:warn` after clicking Close... perhaps I'll reverse the order...

gnucash/gnome-utils/gnome-utils.scm
gjanssens: Very schemey and compact... and very hard to grasp for a non schemer :(
I'll try to remember next time I see such a construct...

libgnucash/app-utils/date-utilities.scm
gjanssens: Ah, I missed that call. I see now.

However that turns the issue around: somewhere below you call  rpterror_earlier followed by a gnc:error. As the former calls gnc:warn you now send two messages to the trace log. One being a warning and the other being an error. That's confusing unfortunately.

Also in the light of the translations, be careful to send the English text to the log and present the translated version to the user...

christopherlam: > However that turns the issue around: somewhere below you call rpterror_earlier followed by a gnc:error. As the former calls gnc:warn you now send two messages to the trace log. One being a warning and the other being an error. That's confusing unfortunately.

Ok found the double log. No difficulty removing old one.

> Also in the light of the translations, be careful to send the English text to the log and present the translated version to the user...

Ack, this one *not* easy! The formatted string gets sent to both gnc:warn and gui-warning-dialog!! If this is important to resolve perhaps *every* gnc:gui-error/warn/msg must send two strings - one for UI and the other for console. 
```
(define (rpterror-earlier type newoption fallback)
  (let ((template (N_ "This report was saved using a later version of \
GnuCash. One of the newer ~a options '~a' is not available, fallback to the \
option '~a'."))
        (console-msg (format #f template type newoption fallback))
        (ui-msg (format #f (_ template) type newoption fallback)))
    (gnc:gui-warn console-msg ui-msg)))
```

Is this desirable?

christopherlam: I've created last amendment for translations but think they add unnecessary complexity. Every gnc:gui-error will need to generate two strings.

cf price-quotes.scm also generates either `gnc-error-dialog` or `gnc:warn` and uses translated strings only. I'm not sure if price-quotes.scm ever gets to run without UI that's beside the point.

jralls: Yes, price-quotes.scm is run from the command-line via the --get-price-quotes option. Many users put it in a cron job/scheduled task to get quotes at the end of each trading day.

Your approach declaring the string in (N_ ...) and calling (_ ...) for dispatch to the gui is correct.

christopherlam: Ok, I've fixed up the gui functions and the crasher. Await travis, then merge!

gnucash/gnome-utils/gnome-utils.scm
christopherlam: There's lots of previous examples in code using a more elaborate closure modelled from the manual
https://www.gnu.org/software/guile/manual/html_node/OO-Closure.html#OO-Closure

e.g. https://github.com/Gnucash/gnucash/blob/3a105f0728984df7f063110acc8390c93722d581/gnucash/report/report-system/report-utilities.scm#L293 whereby `commoditylist` is the local variable

https://github.com/Gnucash/gnucash/blob/3a105f0728984df7f063110acc8390c93722d581/libgnucash/app-utils/options.scm#L204 whereby `value` and `value->string` are local variables... and are miles uglier than the compact stuff above...

gjanssens:
Op maandag 18 februari 2019 07:42:05 CET schreef Christopher Lam:
> christopherlam commented on this pull request.
> 
> > @@ -41,3 +41,23 @@
> 
>  (load-from-path "gnc-menu-extensions")
> 
> +(define-public gnc:set-ui-status
> +  (let ((save-warn gnc:gui-warn)
> +        (save-error gnc:gui-error)
> +        (save-msg gnc:gui-msg))
> +    (lambda (status)
> 
> There's lots of previous examples in code using a more elaborate closure
> modelled from the manual
> https://www.gnu.org/software/guile/manual/html_node/OO-Closure.html#OO-Clos
> ure
> 
> e.g.
> https://github.com/Gnucash/gnucash/blob/3a105f0728984df7f063110acc8390c9372
> 2d581/gnucash/report/report-system/report-utilities.scm#L293 whereby
> `commoditylist` is the local variable
> 
> https://github.com/Gnucash/gnucash/blob/3a105f0728984df7f063110acc8390c93722
> d581/libgnucash/app-utils/options.scm#L204 whereby `value` and
> `value->string` are local variables... and are miles uglier than the
> compact stuff above...

Right. And I never claimed I fully understood all of that code. Your code is 
fine. I am just learning more guile here.

christopherlam:
I forgot to consider that 65bfeaf5de8c52c77ba0f4e8f4d5d6ceeb45b33e is potentially important to highlight in release 3.5 - it will change the dates produced in many reports eg asset-chart, networth-chart etc. It shouldn't raise any complaints -- dates can be selected e.g. monthly networth on 31th December would formerly produce 31/12, 31/01, 03/03, 03/04 etc but now produce 31/12, 31/01, 28/02, 31/03 etc.

The main side effect here would be any reports using the previous dates list would now produce different numbers.

A major benefit will be reports can now produce true monthly net-worth figures. Previously users would try start-date = 01-jan, interval = monthly, and would unfortunately get end-of-day networth. Now they can select start-date = 31-12, interval = monthly, and produce net worth figures "at the beginning of every month".

jralls:
That's good to know but unfortunately I'm unlikely to remember to include it in the release notes 3 weeks from now.

PR #456, opened 2019-02-12
gnc-tree-view-split-reg.c:  strncpy into datebuff risked non-null-terminated string.

brentjens:
gcc version 8.2.1 20181127 (GCC) gave a warning pointing at the problem.


gjanssens:
Thanks!

PR #457, opened 2019-02-12
Bugs and improvements Import Bills & Invoices

ekinonnakapito:
Several bug fixes and improvements in Import Bills & Invoices.

* Bug [797024 Import Bills & Invoices: import matches csv data rows with too few separators, messing up the import data](https://bugs.gnucash.org/show_bug.cgi?id=797024)
* Take all header data for an invoice from the first data row of that invoice. Previously, the import function took the header data for posting from the last row, and all other header data from the first row of an invoice.
* Bug [796987 Import Bills & Invoices: fixing import items only works in special case](https://bugs.gnucash.org/show_bug.cgi?id=796987)
* Validate header fields only for the first data row of an invoice.
* An error in a data row will cause the full invoice to be ignored; previously only that row would be ignored.
* Bug [796986 Import Bills & Invoices: in case of invalid posting date in import file, invoice is posted with unexpected date](https://bugs.gnucash.org/show_bug.cgi?id=796986)
* Bug [797022 Import Bills & Invoices: expense/income account is not validated - leads to one-sided posting if account is invalid](https://bugs.gnucash.org/show_bug.cgi?id=797022)
* Bug [797025 Import Bills & Invoices: the type of the post to account is not validated - enabling A/P and A/R postings on regular accounts](https://bugs.gnucash.org/show_bug.cgi?id=797025)
* Bug [796985 Import Bills & Invoices: option 'open not yet posted docs in tab' does not open tab if invoice could not be posted](https://bugs.gnucash.org/show_bug.cgi?id=796985)
* Bug [796984 Import Bills & Invoices: date of line item is not validated; can lead to unexpected results](https://bugs.gnucash.org/show_bug.cgi?id=796984)
* Bug [797023 Import Bills & Invoices: no user confirmation requested for update of invoices, if new invoices is created first](https://bugs.gnucash.org/show_bug.cgi?id=797023)
* Improved error reporting and feedback.
* Added and updated doxygen information.

Rewrote chapter Import Business Data of the guide, to document the functionality; submitted in pull request to gnucash-docs, pull request 117.

### DIFF
--- a/gnucash/import-export/bi-import/dialog-bi-import.c
+++ b/gnucash/import-export/bi-import/dialog-bi-import.c
@@ -25,9 +25,7 @@
  * @brief core import functions for invoice import plugin
  * @author Copyright (C) 2009 Sebastian Held <sebastian.held@gmx.de>
  * @author Mike Evans <mikee@saxicola.co.uk>
- * @todo Create an option to import a pre-formed regex when it is present
- * to enable the use of custom output csv formats.
- * @todo Open the newly created invoice(es).
+ * @author Rob Laan <rob.laan@chello.nl>
  */
 
 #ifdef HAVE_CONFIG_H
@@ -75,6 +73,27 @@
 
 static QofLogModule log_module = G_LOG_DOMAIN; //G_LOG_BUSINESS;
 static char * un_escape(char *str);
+
+/** \brief Imports a csv file with invoice data into a GtkListStore.
+ 
+ Opens the csv file and attempts to match each row with the regular
+ expression provided in parser_regexp. This is a regular expression
+ that matches each field of the import row and the user selected field
+ separators (, or ;), optionally with the fields enclosed in quotes.
+ 
+ If the match is succesful, the fields of the import row are transferred to
+ a row in the GtkListStore store. If the the match is not succesful, the
+ row is ignored. Maintains information about number of rows imported,
+ the number of rows ignored, and the actual ignored rows.
+ 
+ @param filename      The csv filename to read
+ @param parser_regexp The regular expression with which to match the import rows
+ @param store         To store the matched data
+ @param max_rows      The maximum number of rows to import; use 0 for no maximum.
+ @param stats         Return information about matched and non-matched rows. Use NULL if the information is not required.
+ 
+ */
+
 bi_import_result
 gnc_bi_import_read_file (const gchar * filename, const gchar * parser_regexp,
                          GtkListStore * store, guint max_rows,
@@ -217,12 +236,36 @@ gnc_bi_import_read_file (const gchar * filename, const gchar * parser_regexp,
 }
 
 
-//! \brief try to fix some common errors in the csv representation of invoices
-//! * corrects the date format
-//! * corrects ambiguous values in multi line invoices
-//! * ensures customer exists
-//! * if quantity is unset, set to 1
-//! * if price is unset, delete row.
+/** \brief Adjusts and validates invoice import data.
+ 
+ Replaces missing or invalid data with defaults:
+ - if quantity is not set, default to 1
+ - if date_opened is not set or invalid, default to today
+ - if date is not set or invalid, default to date_opened
+ - if due date is not set or invalid, default to date_posted
+ 
+ Validates the import data; any error causes all rows of the same invoice
+ to be deleted from the import data:
+ - id is not set
+ - owner_id is not set, or customer/vendor does not exist
+ - date_posted is not valid
+ - account_posted does not exist
+ - account posted is not the applicable type, A/P or A/R
+ - price is not set
+ - account does not exist
+ 
+ Adjustment and validation for header fields is only done for the first row of an invoice,
+ which is assumed to hold the header data for all items of the same invoice.
+ Currency related validation is done in subsqequent processing by gnc_bi_import_create_bis.
+ 
+ @param store Holds the rows of invoice import data
+ @param n_rows_fixed Increased for every data row that is adjusted in this function
+ @param n_rows_ignored Increased for every data row that is deleted in this function
+ @param info Updated with the error messages from this function
+ @param type The type of the import data, BILL or INVOICE
+ 
+ */
+
 void
 gnc_bi_import_fix_bis (GtkListStore * store, guint * n_rows_fixed, guint * n_rows_ignored,
                        GString * info, gchar *type)
@@ -236,9 +279,6 @@ gnc_bi_import_fix_bis (GtkListStore * store, guint * n_rows_fixed, guint * n_row
     guint dummy;
     gint row = 1, fixed_for_invoice = 0;
     const gchar* date_format_string = qof_date_format_get_string (qof_date_format_get()); // Get the user set date format string
-
-
-    //date_format_string = qof_date_format_get_string (qof_date_format_get());
 
     DEBUG("date_format_string: %s",date_format_string);
     // allow the call to this function with only GtkListeStore* specified
@@ -518,14 +558,27 @@ gnc_bi_import_fix_bis (GtkListStore * store, guint * n_rows_fixed, guint * n_row
 }
 
 
-/***********************************************************************
- * @todo Maybe invoice checking should be done in gnc_bi_import_fix_bis (...)
- * rather than in here?  But that is more concerned with ensuring the csv is consistent.
- * @param GtkListStore *store
- * @param guint *n_invoices_created
- * @param guint *n_invoices_updated
- * @return void
- ***********************************************************************/
+/** \brief Creates and updates invoices from validated import data.
+ 
+ Loops through the import data to create and update invoices.
+ The first data row for an invoice is assumed to hold the header data.
+ 
+ If an invoice already exists, the user is asked, once per import,
+ to confirm that invoices should be updated.
+ If not confirmed, any rows for existing invoices are ignored.
+ If confirmed, entries are added to existing invoices.
+ Posted invoices, however, are never updated.
+ 
+ If the field date_posted is set, the system will
+ attempt to also post the invoice. The system will not
+ post the invoice if the entries of the invoice hold different currencies,
+ or if the currency of the invoice differs from the currency of the account_posted.
+ 
+ As per user selection, the system displays tabs for either all affected invoices,
+ all affected invoices not yet posted, or no invoices at all.
+ 
+ */
+
 void
 gnc_bi_import_create_bis (GtkListStore * store, QofBook * book,
                           guint * n_invoices_created,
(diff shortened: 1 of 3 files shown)


gnucash/import-export/bi-import/dialog-bi-import.c
```diff
@@ -231,7 +231,6 @@ gnc_bi_import_fix_bis (GtkListStore * store, guint * fixed, guint * deleted,
     gboolean valid, row_deleted, row_fixed;
     gchar *id = NULL, *date_opened = NULL, *date_posted = NULL, *due_date = NULL,
         *owner_id = NULL, *date = NULL, *quantity = NULL, *price = NULL;
-    GString *prev_id, *prev_date_opened, *prev_date_posted, *prev_owner_id, *prev_date;	// needed to fix multi line invoices
```
gjanssens: This intermediary commit got me on the wrong track... The final new behaviour of the row validation function is actually defined by the next commit.
Please squash these two commits together so we only see the final and intended behaviour. This intermediate step is confusing to a reviewer.

gnucash/import-export/bi-import/dialog-bi-import.c
```diff
@@ -263,126 +268,100 @@ gnc_bi_import_fix_bis (GtkListStore * store, guint * fixed, guint * deleted,
...
-            if (strlen (owner_id) == 0)
+        }
+        
+        // Validate and fix item data.
+        if (!row_deleted)
+        {
+            // Validate the price.
+            if (strlen (price) == 0)
             {
+                // No valid price, delete the row
```
gjanssens: This is a tricky thing to do: you're validating the first line of a potentially multi-line invoice definition. All the invoice's header fields validated fine in this case. It's only the entry fields on this line that don't validate.
If you now remove the line based on that, you risk that all subsequent lines will be removed as well as they may not have invoice header data at all.
In contrast if a non-first line is missing a price, only that line will be deleted from the invoice. This results in an inconsistent user experience.

I actually think the whole concept of simply removing lines automatically give a poor user experience. It would be much better if the user is shown the bad lines with an explanation of why they were considered bad. That would greatly help the user go in and fix the csv file. But that may be a future improvement. It's not blocking this PR for me.

gnucash/import-export/bi-import/dialog-bi-import.c
```diff
@@ -392,6 +371,28 @@ gnc_bi_import_fix_bis (GtkListStore * store, guint * fixed, guint * deleted,
...
+
+        // If this row was the last row of the invoice...
+        if (!valid || (valid && g_strcmp0 (id, running_id->str) != 0))
+        {
+            on_first_row_of_invoice = TRUE;
+        }
+        else
+        {
+            // The next row has the same id, so it is not the first row of an invoice.
+            // But if the first row of an invoice was deleted, the next row is again the first row of the invoice.
```
gjanssens: > If the first row of an invoice was deleted, the next row is again the first row of the invoice

... except you may have lost the header data as not all lines should be required to carry it. You may have to cater for that case by setting the missing data on this new row based on what you had on the removed one.

gnucash/import-export/bi-import/dialog-bi-import.c
```diff
@@ -392,6 +371,28 @@ gnc_bi_import_fix_bis (GtkListStore * store, guint * fixed, guint * deleted,
 
         if (!row_deleted)
             valid = gtk_tree_model_iter_next (GTK_TREE_MODEL (store), &iter);
+        if (valid) gtk_tree_model_get (GTK_TREE_MODEL (store), &iter, ID, &id, -1);
+
+        // If this row was the last row of the invoice...
+        if (!valid || (valid && g_strcmp0 (id, running_id->str) != 0))
```
gjanssens: In addition to an equal id, an empty id should also have you treat the new line as a line of the current invoice.

gnucash/import-export/bi-import/dialog-bi-import.c
```diff
@@ -528,18 +532,10 @@ gnc_bi_import_create_bis (GtkListStore * store, QofBook * book,
                                       gnc_search_customer_on_id (book, owner_id));
             gncInvoiceSetOwner (invoice, owner);
             gncInvoiceSetCurrency (invoice, gncOwnerGetCurrency (owner));	// Set the invoice currency based on the owner
-            if (strlen (date_opened) != 0)	// If a date is specified in CSV
-            {
-                // FIXME: Must check for the return value of qof_scan_date!
+
                 qof_scan_date (date_opened, &day, &month, &year);
```
gjanssens: Can you fix the indentation here ?

gnucash/import-export/bi-import/dialog-bi-import.c
```diff
@@ -251,12 +251,12 @@ gnc_bi_import_fix_bis (GtkListStore * store, guint * fixed, guint * deleted,
     
     // Init control variables
     running_id = g_string_new("");
+    ignore_invoice = FALSE;
```
gjanssens: I may be missing something here.
I can follow you are validating all rows in the import file. And if any one field is invalid you will skip further validation of all rows related to the invoice this particular field is part of. And a single message for that invoice is stored in info.

The rows themselves however remain in the list store.

Doesn't that mean the importer will try to import those lines anyway ?

And also a multi-line invoice may have multiple invalid values in the entry related fields. I think it would be useful if those were all reported to the user so the user can fix all the issues the import file in one go instead of having to iterate over each one as the become reported by the importer.

gnucash/import-export/bi-import/dialog-bi-import.c
```diff
@@ -606,100 +605,94 @@ gnc_bi_import_create_bis (GtkListStore * store, QofBook * book,
...
-        {
-             DEBUG( "Creating a new : %s\n", type );
-            // new invoice
-            invoice = gncInvoiceCreate (book);
-            /* Protect against thrashing the DB and trying to write the invoice
-             * record prematurely */
-            gncInvoiceBeginEdit (invoice);
-            gncInvoiceSetID (invoice, id);
-            owner = gncOwnerNew ();
             if (g_ascii_strcasecmp (type, "BILL") == 0)
```
gjanssens: This unwanted extra indentation makes this commit diff pretty hard to read... Can you fix it please ?

jralls:
@ekinonnakapito Was this a fire-and-forget? You've missed getting this into GnuCash 3.5.

ekinonnakapito:
Not a fire-and-forget. I will be working on this in the coming days. Sorry to have missed the 3.5 release.

gjanssens:
That's fine. We just wanted to figure out how to proceed. It will be for 3.6 then :)

jralls:
We're just over 5 weeks from releasing 3.6 and no progress...

gjanssens:
@ekinonnakapito any progress on this ?

ekinonnakapito:
Not much progress yet. After an upgrade of my OSX, I ran into a lot of issues (not GnuCash related), and had to migrate to a new user account. That took up a lot of my time. But I'm up and running again. The coding for the changes is not that complex, but I need to improve my understanding of Git first, to understand how to process these changes. Splitting the single commit into one per bug was not a smooth exercise for me in Git; I need to better understand how to work efficiently in Git.

jralls:
I highly recommend [Scott Chacun's ProGit](https://git-scm.com/book). You can read the book online for free or buy a dead-trees copy and help support the effort. Also at that site are nice web versions of all of the [man pages](https://git-scm.com/docs). If you're GUI-tool oriented, [Atlassian's SourceTree](https://www.sourcetreeapp.com/) is a lot more complete and easier to use than Xcode's git support.

gnucash/import-export/bi-import/dialog-bi-import.c
ekinonnakapito: Yes, I agree that this results in an inconsistent user experience. And I could have added code that takes the validated header data of a first row, and uses it for subsequent rows of the same invoice, even if this first row is deleted because the item data does not validate.
However, in my view, if an invoice has one row that does not validate, you should reject the complete invoice; else you will import only the correct items of the invoice, which leads to an incorrect total amount of that invoice in the system. I implemented that functionality (all invoice items should validate, or the invoice will not be imported) in the next commit. So I did not put the effort into solving the somewhat inconsistent user experience here, because I implemented a different and consistent user experience in the next commit.
I also agree that it would be better to show all bad lines to the user, but, as you say, that may be a future improvement. For now, I at least improved the user feedback on the first erroneous row, in a subsequent commit of the pull request.

gnucash/import-export/bi-import/dialog-bi-import.c
ekinonnakapito: If a row of an invoice does not validate, ignore_invoice is set. At the end of the processing of that row (at line 365), all rows of the same invoice are removed from the store.

I will add the reporting on all errors in an invoice (as opposed to only the first error) in a separate commit.

gnucash/import-export/bi-import/dialog-bi-import.c
ekinonnakapito: I removed that logic, because I think it is bad practice to manage data records without the records' keys. But anyway, I have added it again (locally for now).

gnucash/import-export/bi-import/dialog-bi-import.c
ekinonnakapito: I think the compare is confused because the same line of code appears twice in the source code; the second occurrence has an extra indent. The change here is fairly straightforward: previously, the code would check for an existing invoice for every row of the import; in the new version, the code does this check only for the first row of the invoice. So the code for 'check of invoice exists, and if not, create it' is now included in the 'if on_first_row_of_invoice' statement, so this code gets an extra indent. But the compare now matches second ocurrence of the said line of code in the old version, with the first occurrence to that line of code in the new version. That makes the compare difficult to read. I think the indents are ok here. I will try add make a comment, to prevent this 'compare confusion'.

gnucash/import-export/bi-import/dialog-bi-import.c
```diff
@@ -526,9 +526,9 @@ gnc_bi_import_create_bis (GtkListStore * store, QofBook * book,
     GtkWidget *dialog;
     time64 today;
     InvoiceWindow *iw;
-    gchar *new_id = NULL;
     gint64 denom = 0;
     gnc_commodity *currency;
+    GString *running_id;
```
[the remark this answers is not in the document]
ekinonnakapito: Well, that is just my lack of understanding. I used `GString*` because the original author used the same for a similar purpose in the current version of this file, line 234:

`GString *prev_id, *prev_date_opened, *prev_date_posted, *prev_owner_id, *prev_date;	// needed to fix multi line invoices`

ekinonnakapito:
I've made the requested changes, or added comments. Update of the documentation for the substitution of blank invoice ID with invoice ID from previous row is in Gnucash-docs pull request #123.

gnucash/import-export/bi-import/dialog-bi-import.c
gjanssens: Commits have been squashed together, thanks.

gnucash/import-export/bi-import/dialog-bi-import.c
gjanssens: I missed that. Your choice to report all errors is better.

gnucash/import-export/bi-import/dialog-bi-import.c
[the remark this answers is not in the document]
gjanssens: That's fair enough.

gnucash/import-export/bi-import/dialog-bi-import.c
gjanssens: I can't remember what this was about exactly. In the final commit series I no longer see "unwanted indentation".

gnucash/import-export/bi-import/dialog-bi-import.c
gjanssens: Same here. I can't detect indentation problems here any more.

gjanssens:
Thanks again!

PR #459, opened 2019-02-14
[master] book-accounting-period

christopherlam:
Hopefully one of many positive changes
* book-accounting-period

### DIFF
--- a/libgnucash/app-utils/business-prefs.scm
+++ b/libgnucash/app-utils/business-prefs.scm
@@ -30,6 +30,9 @@
 (use-modules (gnucash app-utils options))
 (use-modules (gnucash app-utils business-options))
 
+(define gnc:*accounting-period-tab* (N_ "Accounting Period"))
+(define gnc:*accounting-period-eofy* (N_ "End of Fiscal Year Date"))
+
 (define gnc:*option-section-counters* (N_ "Counters"))
 
 ;; This defines all available counter types to show options for. This a
@@ -133,6 +136,15 @@
     gnc:*business-label* gnc:*fancy-date-label*
     "g" (N_ "The default date format used for fancy printed dates.")
     #f))
+
+  ;; Accounting Period
+  (reg-option
+   (gnc:make-date-option
+    gnc:*accounting-period-tab* gnc:*accounting-period-eofy*
+    "a2" (N_ "Please choose the custom end of financial year date. Please be \
+aware only the day and month are used. The year stated is unused.")
+    (lambda () (cons 'absolute (current-time)))
+    #f 'absolute #f))
 
   ;; Accounts tab
 
--- a/libgnucash/app-utils/date-utilities.scm
+++ b/libgnucash/app-utils/date-utilities.scm
@@ -574,51 +574,71 @@ Defaulting to today."))
     (gnc-mktime now)))
 
 (define (gnc:get-start-accounting-period)
+  (issue-deprecation-warning "gnc:get-start-accounting-period is deprecated")
   (gnc-accounting-period-fiscal-start))
 
 (define (gnc:get-end-accounting-period)
+  (issue-deprecation-warning "gnc:get-end-accounting-period is deprecated")
   (gnc-accounting-period-fiscal-end))
 
+;; returns the stored book eofy option, or #f if it is not set.
+(define (get-book-eofy)
+  (qof-book-get-option
+   (gnc-get-current-book)
+   (list gnc:*accounting-period-tab* gnc:*accounting-period-eofy* "value")))
+
 ;; returns the book-eofy, initialized to this-year and adjusted for
-;; yeardelta years. e.g. (book-eofy 2) means end-of-fiscal-year
-;; for the year (this year+2).
+;; yeardelta years. e.g. (book-eofy 2) means end-of-fiscal-year for
+;; the year (this year+2). this should only be called if book has eofy
+;; property.
 (define (book-eofy yeardelta)
-  (let* ((custom (qof-book-get-option (gnc-get-current-book)
-                                      (list gnc:*accounting-period-tab*
-                                            gnc:*accounting-period-eofy*
-                                            "value")))
-         (custom-date (or custom (gnc-accounting-period-fiscal-end)))
+  (let* ((eofy (get-book-eofy))
          (deltayears (+ (- (tm:year (gnc-localtime (current-time)))
-                           (tm:year (gnc-localtime custom-date)))
+                           (tm:year (gnc-localtime eofy)))
                         yeardelta)))
-    (gnc:time64-end-day-time
-     (incdate-months custom-date (* 12 deltayears)))))
+    (incdate-months eofy (* 12 deltayears))))
 
 ;; returns the start current book accounting period
 (define (gnc:get-start-book-accounting-period)
-  (let ((now (current-time))
-        (eofy (book-eofy 0)))
-    (gnc:time64-start-day-time
-     (incdate (if (< now eofy) (book-eofy -1) eofy) DayDelta))))
+  (if (get-book-eofy)
+      (let ((eofy (book-eofy 0)))
+        (gnc:time64-start-day-time
+         (incdate (if (< (current-time) eofy)
+                      (book-eofy -1)
+                      eofy)
+                  DayDelta)))
+      (gnc-accounting-period-fiscal-start)))
 
 ;; returns the end current book accounting period
 (define (gnc:get-end-book-accounting-period)
-  (let ((now (current-time))
-        (eofy (book-eofy 0)))
-    (if (< now eofy) eofy (book-eofy 1))))
+  (if (get-book-eofy)
+      (let ((eofy (book-eofy 0)))
+        (gnc:time64-end-day-time
+         (if (< (current-time) eofy)
+             eofy
+             (book-eofy 1))))
+      (gnc-accounting-period-fiscal-end)))
 
 ;; returns the start previous book accounting period
 (define (gnc:get-start-book-prev-accounting-period)
-  (let ((now (current-time))
-        (eofy (book-eofy 0)))
-    (gnc:time64-start-day-time
-     (incdate (if (< now eofy) (book-eofy -2) (book-eofy -1)) DayDelta))))
+  (if (get-book-eofy)
+      (let ((eofy (book-eofy 0)))
+        (gnc:time64-start-day-time
+         (incdate (if (< (current-time) eofy)
+                      (book-eofy -2)
+                      (book-eofy -1))
+                  DayDelta)))
+      (decdate (gnc-accounting-period-fiscal-start) YearDelta)))
 
 ;; returns the end previous book accounting period
 (define (gnc:get-end-book-prev-accounting-period)
-  (let ((now (current-time))
-        (eofy (book-eofy 0)))
-    (if (< now eofy) (book-eofy -1) (book-eofy 0))))
+  (if (get-book-eofy)
+      (let ((eofy (book-eofy 0)))
+        (gnc:time64-end-day-time
+         (if (< (current-time) eofy)
+             (book-eofy -1)
+             eofy)))
+      (decdate (gnc-accounting-period-fiscal-end) YearDelta)))
 
 (define (gnc:get-start-this-month)
   (let ((now (gnc-localtime (current-time))))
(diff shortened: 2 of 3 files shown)


gjanssens:
So this PR would add a new Accounting Period scheme which would be book specific. In itself I think that's a good thing.
On the other hand that now means we'd have two places to define an Accounting period:
1. In the general preferences
2. In the book specific preferences

That is IMO a confusing situation. I'm inclined to deprecate the global preference in the 4.x cycle and only keep the book specific one from 5.x going forward. That gives users a grace period to change all uses of the old option to the new one.

For this I would:
- Ensure everywhere the user currently can select the global accounting period as an option, the user can equally select the book specific accounting period as an option. IIUC the current PR does so for the reports. Are there other places in the gnucash code the use the global accounting period that aren't covered yet ?
- Rename all visual elements referring to the global accounting period to "Accounting Period (Deprecated)" and update any tool tips explaining the book specific accounting period should be preferred.

christopherlam:
I'll be adding further work soon
1. i18n strings
2. multichoice should be symbols rather than symbols+`#f`... I'll make the 'custom-eofy' a symbol instead of `#f` for consistency
3. deprecating global-accounting-period shouldn't be difficult.... `gnc_accounting_period_fiscal_end` can call transparently call `gnc:get-end-book-accounting-period` and the strings can be amended for transitioning.
4. the options will be changed so that it defaults to custom-eofy on the date `gnc_accounting_period_fiscal_end` although I'll need to be careful to avoid circular references...

christopherlam:
I still can't read C. Is https://github.com/Gnucash/gnucash/blob/maint/libgnucash/app-utils/gnc-accounting-period.c#L106 relating to an unused prior book-accounting-period work?

jralls:
That's more a history question than a C question. It seems to be a leftover from a previous incarnation of the summary bar.

christopherlam:
There's no kvp fy-end setter in code, so it's effectively dead. I'm guessing the following will be safe?

```patch
modified   libgnucash/app-utils/gnc-accounting-period.c
@@ -96,22 +96,11 @@ lookup_end_date_option (GDate *fy_end)
     return time;
 }
 
-static GDate *
-get_fy_end (void)
-{
-    QofBook *book;
-    GDate *date = NULL;
-
-    book = gnc_get_current_book();
-    qof_instance_get (QOF_INSTANCE (book), "fy-end", &date, NULL);
-    return date;
-}
-
 time64
 gnc_accounting_period_fiscal_start (void)
 {
     time64 t;
-    GDate *fy_end = get_fy_end();
+    GDate *fy_end = NULL;
     t = lookup_start_date_option (fy_end);
     if (fy_end)
         g_date_free (fy_end);
@@ -122,7 +111,7 @@ time64
 gnc_accounting_period_fiscal_end (void)
 {
     time64 t;
-    GDate *fy_end = get_fy_end();
+    GDate *fy_end = NULL;
 
     t = lookup_end_date_option (fy_end);
     if (fy_end)
```

I wouldn't do it myself; I don't know what to do with the other instances of `"fy-end"` and `fy_end` in code base.

jralls:
Safe, but not correct. All of the ```fy_end``` code should be removed, probably more than you want to undertake given that it's C.

christopherlam:
Another thought. Instead of adding another option to date selectors, we can potentially *amend* the existing https://github.com/Gnucash/gnucash/blob/master/libgnucash/app-utils/gnc-accounting-period.c#L80  to query from scheme rather than from global preferences. Would that be cleaner?

P.S.  I don't understand this code -- it seems to retrieve book kvp's fy-end then ignores it and retrieves global preferences.

jralls:
It's not necessarily throwing away fy_end, it's maybe passing it to `gnc_accounting_period_end_gdate` via `gnc_accounting_period_end_time64`.  If `which` is of the right value it's used to calculate the timestamp of either the current or previous fiscal year according to the book option. Whether it does that and what is the value of `which` is depends on the preference settings queried in `lookup_enddate_option`.

christopherlam:
Instead of trying to transition from global to book accounting period, there's enough space in town for both?

The book accounting period can be set per-book, and can also be set to 'use global preference'?

jralls:
I don't see the use-case for that, but regardless that discussion should happen on gnucash-user.

christopherlam:
>  don't see the use-case for that, but regardless that discussion should happen on gnucash-user.

The purpose of allowing both was to modify the scheme calls to fiscal-year-start and fiscal-year-end to try book accounting period first, and if it doesn't exist or set to 'global', choose global preference. The benefit would be a much easier upgrade path because we wouldn't need to remove the global preference. The difficulty is that the global preference asks for 2 distinct dates whereas this branch asks for EOFY date only and assumes a 1yr accounting period.

gjanssens:
I don't see a use case for a global accounting period either. This is really a book property.

On the other hand I do care about providing smooth upgrade paths.
If we just drop it, users would have to adjust their book properties the first time they use a newer version of gnucash.
If instead we deprecate it for 4.x and only remove it for 5.x, we can use the 4.x cycle to initialize the book property based on the value of the global property.

That is, if no fy book property is set, try to deduce one from the global accounting period data. We should also make sure that global property is really never used any more in our own code and change it's widget and/or tooltip to inform the user of this.

christopherlam:
Ok. I'll try see how to make it work.

Next item: how best to set the option: my suggestion here is to find an eofy date and deduce startdate 1 year prior. Old preference asks both start and send dates. How to proceed?

jralls:
I think we should keep the two-date setting. It's sometimes necessary to have a "short" fiscal year; one example is when changing it, say to start in October instead of July. Most tax authorities require that an FY be no longer than 12 months so in that situation one must have a one-off July-October FY.

gjanssens:
I'm aware of non-year boundary fiscal years. The first fiscal year of our own company was 18 months (so yes, more than 1 year actually).

If we fully want to support this, we're in a similar situation as with reconciling and statements: in that case we need to record each and every fiscal year with start and end date. Not something I'm looking forward to. Interestingly an older version of Odoo worked that way. It had shortcut buttons to quickly add new fiscal years (which were then subdivided into accounting periods; either months or quarters). Newer versions no longer do that. They now just work as @christopherlam proposes: you can set an ending date (day and month) and they will just calculate fiscal years from there.

Further thoughts:
* even if non-year fiscal years exist, I would assume them to be uncommon, like the first or last fiscal year of a company's full life time. Or an exceptional one in the middle, but not the norm. Particularly as gnucash is targeting management of personal finances of small businesses. 
* fiscal year is merely used to determine start and end dates for reporting. Having a single selector like "current fiscal year" or "previous fiscal year" are very handy shortcuts that would work for most fiscal years, namely those that are really one year long. However if there are exceptional fiscal years, one can easily generate the reports for that fiscal year by manually choosing the proper start and end dates on the report options. We're not losing functionality or introduce errors by having the "fiscal year" selector assume a year is 365 days. It merely may not be the appropriate selector for fiscal years that are not exactly one year long. I see it as a convenience selector that works year after year without having to change it all the time (which the current accounting period preference requires).

jralls:
OK.

christopherlam:
Next question: is it worth the shortcut and (minimal) increased complexity of having preset eofy dates? (I think having a preset will lead to requests for more dates!) Or simplify to use a single date option?

image
image

gjanssens:
Single end of year entry would suffice. And ideally it would just be two fields: one day and one month, not the year. I remember you found that too challenging way back. Do you think this is within reach now ?

jralls:
Making a day-and-month option will require a new option type in options.scm and a new widget in dialog-options.c. It would be a little uglier to use the existing absolute-date option type and widget and just ignore the year in the option implementation, but a lot less work.

gjanssens:
It could be two separate options, one for day (text field with spinner) and a month field (combo box). I think we have both widget types already. And we can even link two options such that the spinner's maximum number is adjusted depending on the month that's set. Shouldn't be that much work. However it may be more than is worth. I just worry about user confusion.

jralls:
That would make two options that would have to be stitched together.

christopherlam:
Actually the choice of an date option was a @derekatkins request.

fellen:
I am a little late ...
> I'm aware of non-year boundary fiscal years. The first fiscal year of our own company was 18 months (so yes, more than 1 year actually).
> 
> If we fully want to support this, we're in a similar situation as with reconciling and statements: in that case we need to record each and every fiscal year with start and end date. Not something I'm looking forward to. Interestingly an older version of Odoo worked that way. It had shortcut buttons to quickly add new fiscal years (which were then subdivided into accounting periods; either months or quarters). Newer versions no longer do that. They now just work as @christopherlam proposes: you can set an ending date (day and month) and they will just calculate fiscal years from there.

I like that idea.

> Further thoughts:
> 
>     * even if non-year fiscal years exist, I would assume them to be uncommon, like the first or last fiscal year of a company's full life time. Or an exceptional one in the middle, but not the norm. Particularly as gnucash is targeting management of personal finances of small businesses.

* I have once worked for a company, which was bought by another and had to adjust their year.
* The government could force a new fiscal year.
 
>     * fiscal year is merely used to determine start and end dates for reporting. Having a single selector like "current fiscal year" or "previous fiscal year" are very handy shortcuts that would work for most fiscal years, namely those that are really one year long. However if there are exceptional fiscal years, one can easily generate the reports for that fiscal year by manually choosing the proper start and end dates on the report options. We're not losing functionality or introduce errors by having the "fiscal year" selector assume a year is 365 days. 

~365.2425 d in the Gregorian calendar, which is a solar calendar, but  there are also
* lunar  (e.g. islamic with 354 or 355 days) and
* lunisolar (e.g. Hebrew 353-385 days, trad. Chinese and Hindu) calendars.

Solar Hijri is the official calendar in AF and IR and begins on the March equinox (Nowruz). I could imagine more countries leaving the Gregorian calendar in this strange times.

see https://en.wikipedia.org/wiki/Year and Calendar etc.

> Single end of year entry would suffice. 

End of february is hard to encode.

jralls:
Let's leave non-Gregorian calendars for another PR.

christopherlam:
If that's all that's required then I think this is ready to merge to master. Adds date options to retrieve book eofy. If book-eofy doesn't exist it defaults to today. Next step is to softly deprecate global option, and in 5.x:

```patch
(define (gnc:get-start-accounting-period)
-  (gnc-accounting-period-fiscal-start))
+  (gnc:get-start-book-accounting-period))

(define (gnc:get-end-accounting-period)
-  (gnc-accounting-period-fiscal-end))
+  (gnc:get-end-book-accounting-period))
```

gjanssens:
Another minor detail: as we introduce a new value to the date selectors in 4.x, we probably should change the saved-reports version number. Otherwise the user choice may be lost of the report is opened in version 3.x

gjanssens:
And lastly, let's just drop my idea of separate options. While technically possible it's still not very nice.

christopherlam:
Soft deprecation meant to allow both global and book periods, but all UI modified to nudge users away from global period.

E.g. on load, if book option isn't set, a dialog to nudge them to set it. A saved report which uses global option can still run but may show popup to nudge user to select book option.

Basically allow both, but don't specifically use any heuristics to guess the book property.

Would this be too intrusive?

jralls:
> Another minor detail: as we introduce a new value to the date selectors in 4.x, we probably should change the saved-reports version number. Otherwise the user choice may be lost of the report is opened in version 3.x

Why would we be introducing a new value into the date selectors for reports as part of this PR?

christopherlam:
> > Another minor detail: as we introduce a new value to the date selectors in 4.x, we probably should change the saved-reports version number. Otherwise the user choice may be lost of the report is opened in version 3.x
> 
> Why would we be introducing a new value into the date selectors for reports as part of this PR?

This PR adds 2 new date selectors `start/end of current book accounting period`, and `start/end of previous book accounting period`.

jralls:
We already have start/end of accounting period. I'd suggest coding that to get the book option if it's set and the preference if it isn't. Otherwise users will get it into their heads that there should be both and will complain when the preference goes away.

There isn't a previous period option now because it doesn't make sense with the preference, but it does with the  book option.

christopherlam:
> We already have start/end of accounting period. I'd suggest coding that to get the book option if it's set and the preference if it isn't. Otherwise users will get it into their heads that there should be both and will complain when the preference goes away.

Are you sure this is the best option? I think they'll be more upset when reports suddenly change when they haven't had a chance to set to book property yet. To allow old option yet emit an annoying dialog is another way.

jralls:
What part of " if it's set and the preference if it isn't" do you not understand?

christopherlam:
Hadn't quite caught that. With this, I think it does the trick.

gjanssens:
> We already have start/end of accounting period. I'd suggest coding that to get the book option if it's set and the preference if it isn't. Otherwise users will get it into their heads that there should be both and will complain when the preference goes away.
> 
They are not really equivalent. The preference can set a period that is not a year. A user may have chosen a quarter instead. Book-eofy can only be one year. So by combining those under the same selector you can change the behaviour of their reports without them really being aware of that. Namely if both the preference is set (to say quarter) and book-eofy, any report using accounting period as date selector would suddenly show different results (a full year instead of a quarter). Users will probably complain about that just as much.

But perhaps if we add sufficient feedback in various places it can pass. Here's a proposal to that effect:
1. I stick with my recommendation to alter the preferences tab (change tab name, add label with deprecation notice, move the two summary options elsewhere)
1. Introduce some kind of notification mechanism in reports. That is a block of text that can be displayed in the report with a warning or info message. For this particular situation, if a date selector is set to `Accounting Period` and book-eofy is not set (hence the preference is used) use this mechanism to display a notification at the top of the report. Something along the lines of

> Note this report has been generated based on the deprecated Accounting Period preference. For this message to go away, either configure the new Fiscal Year book property or choose another Start Date/End Date)

christopherlam:
@gjanssens try the latest iteration b8c4623 -- this version will retain the old 'start/end accounting period' and introduce the new option 'start/end previous accounting period'.

if book property isn't set:
* start/end accounting period will use global preference
* start/end previous accounting period will use global preference, less 1 year

if book property set:
* start/end accounting period which includes today
* start/end previous accounting period prior to today's accounting period

christopherlam:
See latest amendment, unsure if useful. Now the follow results are obtained:

if book property isn't set:
   * start/end accounting period return the global preference period.
   * start/end previous accounting period return the period with the same number of days immediately preceding from global preference period.

flywire:
The displayed dates are wrong. Can displayed dates be updated when the period is updated?

PR #460, opened 2019-02-15
Transaction Association Changes

Bob-IT:
I decided to create a new PR with the transaction association changes highlighted in PR432 and when this is accepted both can be closed as follows...

Changed the location association to require a scheme to be present, normally that would be http:// or https:// but if the user knows what they are doing it could be any valid uri with a scheme.

This allows the associate kvp to have the following...
* `file:///c:/somedir/some_file` is absolute file path on windows

* `file:///usr/somedir/some_file` is absolute as well on linux

* `somedir/some_file` is relative file path on both linux and windows

* `some_file` is relative file on both linux and windows

* `http://www.gnucash.org` is location url

So if the kvp entry does not have a valid scheme, it is treated as a relative file path.

The relative path head is only set by the GtkFileChooser so it will be an absolute file path uri of the form...
* `file:///n:/` is a root of a network share on windows

* `file:///n:/somedir` is a subdirectory on network share on windows

* `file:///mnt/somedir` is subdirectory on linux

When a file is selected in the file chooser, the prefix of the file path is compared to that of the path head setting, if they match then the path head is removed from the chosen file path and this remainder is stored in the kvp entry.

I have used the existing uri's functions to achieve this and tested both on Windows 10 and Linux and it does what I think it should do but as always will wait and see.

### DIFF
--- a/gnucash/gnome-utils/gnc-gnome-utils.c
+++ b/gnucash/gnome-utils/gnc-gnome-utils.c
@@ -45,6 +45,7 @@
 #include "dialog-commodity.h"
 #include "dialog-totd.h"
 #include "gnc-ui-util.h"
+#include "gnc-uri-utils.h"
 #include "gnc-session.h"
 #include "qofbookslots.h"
 #ifdef G_OS_WIN32
@@ -482,32 +483,33 @@ void
 gnc_launch_assoc (const char *uri)
 {
     wchar_t *winuri = NULL;
-    char* scheme = g_uri_parse_scheme(uri);
     /* ShellExecuteW open doesn't decode http escapes if it's passed a
      * file URI so we have to do it. */
-    if (scheme && strcmp (scheme, "file") == 0)
+    if (gnc_uri_is_file_uri (uri))
     {
-	gchar *filename = g_filename_from_uri (uri, NULL, NULL);
-	winuri = (wchar_t *)g_utf8_to_utf16(filename, -1, NULL, NULL, NULL);
+        gchar *uri_ue = g_uri_unescape_string (uri, NULL);
+        gchar *filename = gnc_uri_get_path (uri_ue);
+        winuri = (wchar_t *)g_utf8_to_utf16(filename, -1, NULL, NULL, NULL);
+        g_free (filename);
+        g_free (uri_ue);
     }
     else
-	winuri = (wchar_t *)g_utf8_to_utf16(uri, -1, NULL, NULL, NULL);
+        winuri = (wchar_t *)g_utf8_to_utf16(uri, -1, NULL, NULL, NULL);
 
     if (winuri)
     {
-	wchar_t *wincmd = (wchar_t *)g_utf8_to_utf16("open", -1,
-						     NULL, NULL, NULL);
-	if ((INT_PTR)ShellExecuteW(NULL, wincmd, winuri,
-				   NULL, NULL, SW_SHOWNORMAL) <= 32)
-	{
-	    const gchar *message =
-		_("GnuCash could not find the associated file");
-	    gnc_error_dialog(NULL, "%s: %s", message, uri);
-	}
-	g_free (wincmd);
-	g_free (winuri);
+        wchar_t *wincmd = (wchar_t *)g_utf8_to_utf16("open", -1,
+                                 NULL, NULL, NULL);
+        if ((INT_PTR)ShellExecuteW(NULL, wincmd, winuri,
+                       NULL, NULL, SW_SHOWNORMAL) <= 32)
+        {
+            const gchar *message =
+            _("GnuCash could not find the associated file");
+            gnc_error_dialog(NULL, "%s: %s", message, uri);
+        }
+        g_free (wincmd);
+        g_free (winuri);
     }
-    g_free (scheme);
 }
 
 #else
--- a/gnucash/gnome/dialog-trans-assoc.c
+++ b/gnucash/gnome/dialog-trans-assoc.c
@@ -525,7 +525,6 @@ gnc_assoc_dialog_create (GtkWindow *parent, AssocDialog *assoc_dialog)
     gtk_tree_view_column_set_title (tree_column, _("Relative"));
     gtk_tree_view_append_column (GTK_TREE_VIEW(assoc_dialog->view), tree_column);
     gtk_tree_view_column_set_alignment (tree_column, 0.5);
-    gtk_tree_view_column_set_expand (tree_column, TRUE);
     cr = gtk_cell_renderer_pixbuf_new();
     gtk_tree_view_column_pack_start (tree_column, cr, TRUE);
     // connect 'active' and set 'xalign' property of the cell renderer
@@ -534,6 +533,10 @@ gnc_assoc_dialog_create (GtkWindow *parent, AssocDialog *assoc_dialog)
 
     g_signal_connect (assoc_dialog->view, "row-activated",
                       G_CALLBACK(row_selected_cb), (gpointer)assoc_dialog);
+
+    // set the Associate column to be the one that expands
+    tree_column = GTK_TREE_VIEW_COLUMN(gtk_builder_get_object (builder, "uri-entry"));
+    gtk_tree_view_column_set_expand (tree_column, TRUE);
 
     // Set grid lines option to preference
     gtk_tree_view_set_grid_lines (GTK_TREE_VIEW(assoc_dialog->view), gnc_tree_view_get_grid_lines_pref ());
--- a/gnucash/gnome/dialog-trans-assoc.c
+++ b/gnucash/gnome/dialog-trans-assoc.c
@@ -38,6 +38,8 @@
 #include "gnc-ui.h"
 #include "gnc-ui-util.h"
 #include "gnc-gnome-utils.h"
+#include "gnc-uri-utils.h"
+#include "gnc-filepath-utils.h"
 #include "Account.h"
 
 #define DIALOG_ASSOC_CM_CLASS    "dialog-trans-assoc"
@@ -51,7 +53,7 @@ typedef struct
     GtkWidget    *window;
     GtkWidget    *view;
     const gchar  *path_head;
-    gboolean      valid_path_head;
+    gboolean      path_head_set;
 }AssocDialog;
 
 /* This static indicates the debugging module that this .o belongs to.  */
@@ -76,7 +78,7 @@ gnc_assoc_dialog_window_destroy_cb (GtkWidget *object, gpointer user_data)
 }
 
 static gboolean
-gnc_assoc_dialog_window_key_press_cb(GtkWidget *widget, GdkEventKey *event, gpointer user_data)
+gnc_assoc_dialog_window_key_press_cb (GtkWidget *widget, GdkEventKey *event, gpointer user_data)
 {
     AssocDialog *assoc_dialog = user_data;
 
@@ -139,37 +141,40 @@ assoc_dialog_sort (AssocDialog *assoc_dialog)
     gtk_tree_sortable_set_sort_column_id (sortable, URI, order);
 }
 
-static const gchar *
-convert_uri_relative_to_uri (AssocDialog *assoc_dialog, const gchar *uri)
+static gchar *
+convert_uri_to_filename (AssocDialog *assoc_dialog, const gchar *uri, gchar *scheme)
 {
-    const gchar *new_uri;
+    gchar *file_path = NULL;
 
-    if (assoc_dialog->valid_path_head && g_str_has_prefix (uri,"file:/") &&
-        !g_str_has_prefix (uri,"file://")) // path is relative
+    if (!scheme) // relative path
     {
-        const gchar *part = uri + strlen ("file:");
-        new_uri = g_strconcat (assoc_dialog->path_head, part, NULL);
+        if (assoc_dialog->path_head_set) // not default entry
+            file_path = gnc_file_path_absolute (gnc_uri_get_path (assoc_dialog->path_head), uri);
+        else
+            file_path = gnc_file_path_absolute (NULL, uri);
     }
+
+    if (gnc_uri_is_file_scheme (scheme)) // absolute path
+        file_path = gnc_uri_get_path (uri);
+
+    return file_path;
+}
+
+static gchar *
+convert_uri_to_unescaped (AssocDialog *assoc_dialog, const gchar *uri, gchar *scheme)
+{
+    gchar *uri_u = NULL;
+    gchar *file_path = NULL;
+
+    // if scheme is null or 'file' we should get a file path
+    file_path = convert_uri_to_filename (assoc_dialog, uri, scheme);
+
+    if (file_path)
+        uri_u = g_uri_unescape_string (file_path, NULL);
     else
-        new_uri = g_strdup (uri);
+        uri_u = g_uri_unescape_string (uri, NULL);
 
-    return new_uri;
-}
-
-static gchar *
-convert_uri_to_filename (AssocDialog *assoc_dialog, const gchar *uri)
-{
-    const gchar *new_uri = convert_uri_relative_to_uri (assoc_dialog, uri);
-    gchar *filename = g_filename_from_uri (new_uri, NULL, NULL);
-
-    return filename;
-}
-
-static gchar *
-convert_uri_to_unescaped (AssocDialog *assoc_dialog, const gchar *uri)
-{
-    const gchar *new_uri = convert_uri_relative_to_uri (assoc_dialog, uri);
-    gchar *uri_u = g_uri_unescape_string (new_uri, NULL);
+    g_free (file_path);
 
     return uri_u;
 }
@@ -195,12 +200,15 @@ assoc_dialog_update (AssocDialog *assoc_dialog)
         GSocketConnectable *conn;
         gchar              *uri;
         gchar              *filename;
+        gchar              *scheme;
 
         gtk_tree_model_get (model, &iter, URI, &uri, -1);
 
-        filename = convert_uri_to_filename (assoc_dialog, uri);
+        scheme = gnc_uri_get_scheme (uri);
 
-        if (filename != NULL)
+        filename = convert_uri_to_unescaped (assoc_dialog, uri, scheme);
+
+        if (!scheme || gnc_uri_is_file_scheme (scheme))
         {
             if (g_file_test (filename, G_FILE_TEST_EXISTS))
                 gtk_list_store_set (GTK_LIST_STORE(model), &iter, AVAILABLE, _("File Found"), -1);
@@ -209,18 +217,21 @@ assoc_dialog_update (AssocDialog *assoc_dialog)
         }
         else
         {
+            gchar *escaped = g_uri_escape_string (uri, ":/.", TRUE);
             nm = g_network_monitor_get_default ();
-            conn = g_network_address_parse_uri (uri, 80, NULL);
+            conn = g_network_address_parse_uri (escaped, 80, NULL);
 
-            if (conn != NULL)
+            if (conn)
             {
                 if (g_network_monitor_can_reach (nm, conn, NULL, NULL))
                     gtk_list_store_set (GTK_LIST_STORE(model), &iter, AVAILABLE, _("Address Found"), -1);
                 else
                     gtk_list_store_set (GTK_LIST_STORE(model), &iter, AVAILABLE, _("Address Not Found"), -1);
             }
+            g_free (escaped);
         }
         g_free (uri);
+        g_free (scheme);
         g_free (filename);
         valid = gtk_tree_model_iter_next (model, &iter);
     }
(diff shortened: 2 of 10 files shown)


gnucash/gnome/gnc-split-reg.c
```diff
@@ -1054,33 +1055,50 @@ gsr_default_associate_handler_file (GNCSplitReg *gsr, Transaction *trans, gboole
 
     gtk_file_chooser_set_local_only (GTK_FILE_CHOOSER(dialog), FALSE);
 
-    if ((path_head != NULL) && (g_strcmp0 (path_head, "") != 0)) // not default entry
-        valid_path_head = TRUE;
+    if (path_head && g_strcmp0 (path_head, "") != 0) // not default entry
```
gjanssens: While you're optimizing the if statement it's probably even more efficient to test for a non-empty string using
```
if (path_head && *path_head != '\0')
```
The dereferencing is safe here as you first verify path_head is a valid non-null pointer.

And going yet one step further you could even eliminate the if statement completely and assign the result of that test directly to path_head_set like so:
```
path_head_set =  (path_head && *path_head != '\0');
```
Of course this is mostly a matter of taste so feel free to take or ignore this suggestion.

gnucash/gnome/gnc-split-reg.c
```diff
@@ -1054,33 +1055,50 @@ gsr_default_associate_handler_file (GNCSplitReg *gsr, Transaction *trans, gboole
...
+            g_free (file_path);
         }
-        else
-            new_uri = g_strdup (uri);
 
-        filename = g_uri_unescape_string (new_uri, NULL);
-        uri_label = g_strconcat (_("Existing Association is "), filename, NULL);
-        gtk_file_chooser_set_extra_widget (GTK_FILE_CHOOSER(dialog), gtk_label_new (uri_label));
-        gtk_file_chooser_set_uri (GTK_FILE_CHOOSER(dialog), new_uri);
+        if (g_strcmp0 (scheme, "file") == 0) // absolute path
```
gjanssens: I know the implicit assumption is that file associations can only be set via a GtkFileChooserDialog, hence file is the only scheme to expect. However in the future other parts of gnucash may also set file associations (like python bindings) and they may be using another scheme as supported by gnc_uri_utils (sqlite or xml come to mind). So to be future safe it's probably better to use the gnc_uri_utils provided function to test whether the given scheme is file compatible.

gnucash/gnome/gnc-split-reg.c
```diff
@@ -1089,23 +1107,35 @@ gsr_default_associate_handler_file (GNCSplitReg *gsr, Transaction *trans, gboole
 
     if (response == GTK_RESPONSE_ACCEPT)
     {
+        gboolean add_trailing_slash = FALSE;
         gchar *dialog_uri = gtk_file_chooser_get_uri (GTK_FILE_CHOOSER (dialog));
 
-        PINFO("Dialog File URI: %s\n", dialog_uri);
+        // relative paths do not start with a / so need to add one depending on path_head
```
gjanssens: While this is a good catch, I believe it's more efficient to ensure the prefix always has a trailing slash when you save the prefix (wherever that code is stored). Setting a prefix happens less frequently than testing for relative paths against a prefix.

gnucash/gnome/gnc-split-reg.c
```diff
@@ -1206,21 +1263,24 @@ gsr_default_associate_handler (GNCSplitReg *gsr, gboolean uri_is_file)
...
+
     if (is_trans_readonly_and_warn (GTK_WINDOW(gsr->window), trans))
         return;
 
-    // get the existing uri
-    uri = xaccTransGetAssociation (trans);
-
     // Check for uri is empty or NULL
-    if (g_strcmp0 (uri, "") != 0 && g_strcmp0 (uri, NULL) != 0)
+    if (uri && g_strcmp0 (uri, "") != 0)
```
gjanssens: Same remark as earlier above.

gnucash/gnome/gnc-split-reg.c
```diff
@@ -1206,21 +1263,24 @@ gsr_default_associate_handler (GNCSplitReg *gsr, gboolean uri_is_file)
...
     {
+        gchar *scheme = gnc_uri_get_scheme (uri);
         have_uri = TRUE;
 
-        if (g_str_has_prefix (uri, "file:")) // use the correct dialog
+        if (!scheme || g_strcmp0 (scheme, "file") == 0) // use the correct dialog
```
gjanssens: Same as above.

gnucash/gnome/dialog-trans-assoc.c
```diff
@@ -139,37 +141,40 @@ assoc_dialog_sort (AssocDialog *assoc_dialog)
...
+        if (assoc_dialog->path_head_set) // not default entry
+            file_path = gnc_file_path_absolute (gnc_uri_get_path (assoc_dialog->path_head), uri);
+        else
+            file_path = gnc_file_path_absolute (NULL, uri);
     }
-    else
-        new_uri = g_strdup (uri);
 
-    return new_uri;
+    if (g_strcmp0 (scheme, "file") == 0) // absolute path
```
gjanssens: See remark on other file about more generic file compatible scheme testing.

gnucash/gnome/dialog-trans-assoc.c
```diff
@@ -195,12 +200,15 @@ assoc_dialog_update (AssocDialog *assoc_dialog)
...
 
         gtk_tree_model_get (model, &iter, URI, &uri, -1);
 
-        filename = convert_uri_to_filename (assoc_dialog, uri);
+        scheme = gnc_uri_get_scheme (uri);
+
+        filename = convert_uri_to_unescaped (assoc_dialog, uri, scheme);
 
-        if (filename != NULL)
+        if (!scheme || g_strcmp0 (scheme, "file") == 0)
```
gjanssens: And once again :)

gnucash/gnome/dialog-trans-assoc.c
```diff
@@ -346,12 +403,14 @@ get_trans_info (AssocDialog *assoc_dialog)
             if (g_list_find (trans_list, trans) != NULL)
                 continue;
 
-            uri = xaccTransGetAssociation (trans);
+            // fix an earlier error when storing relative paths in version 3.3
+            uri = gsr_convert_associate_uri (trans);
 
-            if (g_strcmp0 (uri, "") != 0 && g_strcmp0 (uri, NULL) != 0)
+            if (uri && g_strcmp0 (uri, "") != 0)
```
gjanssens: Mentioned earlier.

gnucash/register/ledger-core/split-register-model.c
```diff
@@ -824,12 +840,16 @@ gnc_split_register_get_associate_entry (VirtualLocation virt_loc,
     uri = xaccTransGetAssociation (trans);
 
     // Check for uri is empty or NULL
-    if (g_strcmp0 (uri, "") != 0 && g_strcmp0 (uri, NULL) != 0)
+    if (uri && g_strcmp0 (uri, "") != 0)
     {
-        if (g_str_has_prefix (uri, "file:"))
+        gchar *scheme = gnc_uri_get_scheme (uri);
+
+        if (!scheme || g_strcmp0 (scheme, "file") == 0)
```
gjanssens: Same remark on file compatible scheme test.

gnucash/register/ledger-core/split-register-model.c
```diff
@@ -552,23 +554,37 @@ gnc_split_register_get_associate_tooltip (VirtualLocation virt_loc,
...
+            if (path_head && g_strcmp0 (path_head, "") != 0) // not default entry
+                file_path = gnc_file_path_absolute (gnc_uri_get_path (path_head), uri);
+            else
+                file_path = gnc_file_path_absolute (NULL, uri);
+
             g_free (path_head);
-            return g_strdup (new_uri);
         }
+
+        if (g_strcmp0 (scheme, "file") == 0) // absolute path
```
gjanssens: Same remark on file compatible scheme test.

gnucash/gnome/gnc-split-reg.c
```diff
@@ -1160,8 +1163,19 @@ gsr_default_associate_handler_location (GNCSplitReg *gsr, Transaction *trans, gb
...
     {
         const gchar *dialog_uri = gtk_entry_get_text (GTK_ENTRY (entry));
+        gchar *scheme = gnc_uri_get_scheme (dialog_uri);
+
+        if (!scheme)
+        {
+            gnc_error_dialog (GTK_WINDOW(dialog), "%s",
+                         _("This is not a valid URL, does not start with a valid scheme.\n"
+                           "Maybe you missed starting with http:// or https://"));
+            goto response;
```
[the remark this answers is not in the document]
jralls: Concur. ```goto``` should be used only to escape to a common cleanup label at the end of a function, never to make a loop. The label shouldn't be the same as another identifier, either. 

That aside this is clunky and unfriendly. ```dialog_uri``` should be checked when the user exits the GtkEntry and if it's invalid then raise the message box and disable the "OK" button.

gjanssens:
One potential user improvement that comes to mind is this:
If the user sets/changes the prefix it may be very helpful if gnucash scans the current file associations for absolute paths that are now descendants of the new prefix and ask the user whether or not to convert those to relative as well.
Examples:
1. No prefix set, one file association to file:///home/user/gnucash/support/fileA
* user sets prefix to file:///home/user/gnucash/
* at this point ask whether file:///home/user/gnucash/support/fileA should be converted to support/fileA that is relative to file:///home/user/gnucash/

2. Existing prefix is file:///home/user/gnucash/, one relative file association exists as "fileB" (which is hence in said prefix), one absolute one in file:///home/user/gnc2/fileC and one more absolute one in file:///home/user/somplace/fileD
* user sets new prefix to file:///home/user/gnc2/
* at this point ask whether file:///home/user/gnc2/fileC should be converted to fileC, that is a relative file association instead of absolute.
* fileB is not changed as it's already relative.
* fileD is not changed as it's not a descendant of the new prefix

It's just an idea, and if you are interested in implementing this, make it a new PR. I hope to get this PR #460 merged in before 3.5.

gjanssens:
And another potential user convenience: when a new prefix is added, ask whether relative associations existing in the old prefix should be copied or moved to the new prefix. We should ask because maybe the files had been moved by the user behind our backs already or the files will be coming from somewhere else.

gnucash/gnome-utils/gnc-gnome-utils.c
```diff
@@ -482,30 +483,30 @@ void
...
     wchar_t *winuri = NULL;
-    char* scheme = g_uri_parse_scheme(uri);
+    char* scheme = gnc_uri_get_scheme (uri);
     /* ShellExecuteW open doesn't decode http escapes if it's passed a
      * file URI so we have to do it. */
     if (scheme && strcmp (scheme, "file") == 0)
     {
-	gchar *filename = g_filename_from_uri (uri, NULL, NULL);
-	winuri = (wchar_t *)g_utf8_to_utf16(filename, -1, NULL, NULL, NULL);
+        gchar *filename = g_filename_from_uri (uri, NULL, NULL);
```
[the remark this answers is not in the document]
Bob-IT: Changed...

gnucash/gnome/gnc-split-reg.c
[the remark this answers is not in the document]
Bob-IT: Changed this to disable the OK button unless a valid scheme is present...

libgnucash/core-utils/gnc-filepath-utils.cpp
```diff
@@ -214,6 +214,26 @@ gchar *gnc_file_path_relative_part (const gchar *prefix, const gchar *path)
...
 
+gchar *gnc_file_path_absolute (const gchar *prefix, const gchar *relative)
+{
+    bfs::path path_relative (relative);
+    bfs::path absolute_path;
+
+    if (prefix == nullptr)
+    {
+        bfs::path temp_path = absolute (path_relative);
+        absolute_path = absolute (path_relative, temp_path.root_directory());
```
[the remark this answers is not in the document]
Bob-IT: I have changed this to use the userdata directory as default and imbue, not sure if that is the correct way, just looked at the code below...

libgnucash/core-utils/gnc-filepath-utils.cpp
```diff
@@ -214,6 +214,26 @@ gchar *gnc_file_path_relative_part (const gchar *prefix, const gchar *path)
...
+    }
+    else
+    {
+        bfs::path path_prefix (prefix);
+        absolute_path = absolute (path_relative, path_prefix);
+    }
+    auto abs_path = absolute_path.string();
+
+    return g_strdup (abs_path.c_str());
```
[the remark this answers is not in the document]
Bob-IT: Changed...

gnucash/gnome/dialog-trans-assoc.c
Bob-IT: I have changed most of these to use gnc_is_file_scheme but on the one in gnc_split_reg I have left it as that code will end up at the GtkFileChooser which seems correct to me...

Bob-IT:
I think these changes are correct and have tested it but I am having trouble with my Windows build setup so have not fully tested on there...

For information, I will look at the other suggestions later as I have already a branch with associations for invoices sorted along with an improved dialog and another branch using glyphs on the register.

libgnucash/core-utils/gnc-filepath-utils.cpp
```diff
@@ -1017,6 +1018,24 @@ gnc_userdata_dir_as_path (void)
...
 }
 
+gchar *gnc_file_path_absolute (const gchar *prefix, const gchar *relative)
+{
+    bfs::path path_relative (relative);
+    path_relative.imbue (bfs_locale);
+    bfs::path absolute_path;
+
+    if (prefix == nullptr)
+        absolute_path = absolute (path_relative, gnc_userdata_dir_as_path());
```
gjanssens: While I agree with you gnc_user_data_dir is less random than working directory, it's not the current file's directory. In addition the GNC_USERDATA_DIR is a hidden directory (or at least well hidden) so it's not really a very user friendly solution.

However thinking about this I do understand my request comes with a few challenges. Allow me to think out loud a bit here:

* The mysql and postgres backends don't have a file's directory. So what to do for those ?
* We could state that for those backends relative paths and an empty prefix are not allowed.
* But then what to do if one converts a file to a db ?
* In that case we could run a conversion if the prefix is empty.
* File->Db: replace the empty prefix with the directory of the file being saved from.
* If we can guarantee a non-empty prefix like that while converting to db, the inverse wouldn't be a problem. If we convert from db to file, the db should have either a prefix or no relative file associations.

The corner case(s) here are still in db land:
* if the user creates a new db, it won't have a prefix set. Which is fine as adding file associations will then store absolute paths (though we may want to add a sanity check in the lower level path setting routines to catch invalid values being passed via bindings)
* suppose the user had set a prefix in the db and has added a number of relative paths. And then suppose the user decides to remove the prefix. That would lead to invalid file paths.

I believe we have/had a similar issue to solve for the transaction logs in the sql db's. The current solution is to store those in GNC_USERDATA_DIR/translog. Perhaps it's time we bring some order and logic into this all for gnucash 4.x (other work in that direction is ongoing with saved-reports).

Until we can do so, perhaps it's better to not allow relative file associations with an empty prefix. And if you can do that, the default path you have chosen here is less important.

gnucash/gnome/dialog-trans-assoc.c
gjanssens: That's reasonable.

libgnucash/engine/gnc-uri-utils.c
```diff
@@ -176,7 +176,15 @@ void gnc_uri_get_components (const gchar *uri,
 
     if ( gnc_uri_is_file_scheme ( *scheme ) )
     {
-        *path     = gnc_resolve_file_path ( splituri[1] );
+        /* a true file uri on windows can start file:///N:/
+           so we come here with /N:/                       */
+        if (g_str_has_prefix (splituri[1], "/") && g_strstr_len (splituri[1], -1,  ":") != NULL)
+        {
+            gchar *ptr = splituri[1];
+            *path = gnc_resolve_file_path ( ptr + 1 );
```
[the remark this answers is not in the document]
gjanssens: @Bob-IT is this memory leak gone in your other branches ?

gjanssens:
Other than preventing an empty prefix if there are relative file associations (and possibly the memory leak) this is good to go.

libgnucash/core-utils/gnc-filepath-utils.cpp
Bob-IT: Using a database was the reason why I chose the user data directory, I did not want to make it more complicated and it really is only used in the situation when the path_head is not set.

This can happen currently if it was set and then cleared, (Note: this would be fixed by your other suggestion of scanning for existing relative paths and making them absolute, possibly with a question to the user) or buy using the file on another computer with the preference not set and so to make valid paths the default is used. These paths are obviously incorrect but allow the dialogs to display this and hopefully the user will notice and correct.

I think that absolute paths should be stored unless the user has set the path head, at least they have made a conscious decision to do so. I suppose the interface could be improved by highlighting the path head is not set and so relative paths will not be used, I can look at that when I create the PR for invoice associations.

 Currently if it is not set, then you can not store relative paths so absolute paths will be stored.

Bob-IT:
I have looked at the possible memory leak issue but can not find one. I copied the relevant parts to a new source file and ran that with valgrind and no warnings for the source file were highlighted. I even missed of a g_free to make sure it would show and it did.

gjanssens:
Op vrijdag 29 maart 2019 13:11:19 CET schreef Bob-IT:
> Bob-IT commented on this pull request.
> 
> > @@ -1017,6 +1018,24 @@ gnc_userdata_dir_as_path (void)
> 
>      return gnc_userdata_home;
>  }
> 
> +gchar *gnc_file_path_absolute (const gchar *prefix, const gchar *relative)
> +{
> +    bfs::path path_relative (relative);
> +    path_relative.imbue (bfs_locale);
> +    bfs::path absolute_path;
> +
> +    if (prefix == nullptr)
> +        absolute_path = absolute (path_relative,
> gnc_userdata_dir_as_path());
> 
> Using a database was the reason why I chose the user data directory,

A valid consideration.

> I did
> not want to make it more complicated and it really is only used in the
> situation when the path_head is not set.
> 
"really only used" is the key point here. An unset path_head is the default 
and hence the most common situation. So whatever default we choose is used 
most commonly. Unless of course we guarantee there can't be relative paths in 
case of no path head being set (which we can't :(  ).

Having said that, I still think user data directory is a poor choice. As I 
said this directory is hidden away on most systems (%APPDATA%\gnucash, 
$HOME/.local/share/gnucash,...) and should really only be used to store 
application internal state files.

I believe a much better choice would be the output of
g_get_user_special_dir (G_USER_DIRECTORY_DOCUMENTS);
which (by default) resolves to
- /home/<user>/Documents (linux)
- c:\Users\<user>\Documents (windows)
- the same equivalent on MacOS

That's where I would expect user documents should go if no path is predefined. 
If you change this particular bit your PR is ready to be merged as far as I'm 
concerned.


What follows below are musings for a follow-up PR, which I don't expect will 
make it for 3.5.

> This can happen currently if it was set and then cleared, (Note: this would
> be fixed by your other suggestion of scanning for existing relative paths
> and making them absolute, possibly with a question to the user) or buy
> using the file on another computer with the preference not set and so to
> make valid paths the default is used. These paths are obviously incorrect
> but allow the dialogs to display this and hopefully the user will notice
> and correct.

My first reflex was to suggest path_head should be a book preference then. But 
that's obviously wrong. The location of the associated files can be different 
on different PCs, depending on where the files got copied. It should be a 
book/user type preference. I can very much imagine wanting a different 
location for associated files per book and that location could change when 
moving to a different PC. I have other considerations on what the 
configuration side of this could be, but that's beyond the scope of this PR.

Thinking this through further I believe we should completely avoid assuming a 
path_head default. There's no way we can know where the user did store these 
files on a different PC

Instead of focusing on whether we have a valid absolute path, the more general 
test could be "can we open the file in the association". This encompasses 
several sub-cases:
1. the file has a relative path, but no path_head is set
2. the file has a relative path, but the file is not found relative to the 
current path_head
3. the file has an absolute path, but it can't be found.

I think in all 3 cases we should simply ask the user to resolve this issue and 
offer options for this.
In case 1 and 2 we can ask to either locate the file for us (using a 
GtkFileChooser) or to set a (new) path_head (again using a file_chooser).
In the third case only locating the file should be offered.
In all cases if we got an absolute path from the GtkFileChooser, attempt to 
convert it to relative in case a path_head exists (and the absolute path is a 
descendant of that path_head).
That's probably the sanest thing to do and it would probably work for all 
similar situations - in case the gnucash data file is opened on a different 
PC, when the path head gets cleared or changed, when the user moves files 
around behind our backs,...
> 
> I think that absolute paths should be stored unless the user has set the
> path head, at least they have made a conscious decision to do so. I suppose
> the interface could be improved by highlighting the path head is not set
> and so relative paths will not be used, I can look at that when I create
> the PR for invoice associations.
> 
>  Currently if it is not set, then you can not store relative paths so
> absolute paths will be stored.

Agreed, though my assertion was meant for when there *are* relative paths and 
the user subsequently clears the path_head. I meant to say *that* should be 
prevented. But we can't in case the gnucash data file is moved to another 
computer or loaded by a different user. So the above is probably better.

Bob-IT:
Not sure what just happened with this branch but it looked odd, have made the changes and re-pushed a clean branch rebased on maint.

gjanssens:
No more comments. Feel free to push your work upstream to the maint branch.

Bob-IT:
I had a look for a boost equivalent but the only thing I could find is using getenv("HOME") on Unix and concatenating the results of getenv("HOMEDRIVE") and getenv("HOMEPATH") on Windows.

Should I close this ?

jralls:
You should merge it to maint and push the result to code. Once Github sees the merge it will close it.

Bob-IT:
Well I have pushed to code, is merging using the green 'Merge pull request' button

jralls:
Huh? No, don't use the "Merge pull request" button. That does the merge on Github and leaves code out of the loop. It looks like you rebased onto maint instead of merging your tassoc branch so Github doesn't recognize the merge. We have to close this manually in that case.

PR #461, opened 2019-02-15

sicelo:
Change the 'local-symbol' property for the SZL currency from "L" to "E."
Fixes https://bugs.gnucash.org/show_bug.cgi?id=797105.

### DIFF
--- a/libgnucash/engine/iso-4217-currencies.xml
+++ b/libgnucash/engine/iso-4217-currencies.xml
@@ -2414,6 +2414,7 @@
   local-symbol=""
 />
 <!-- "SZL" - "Lilangeni"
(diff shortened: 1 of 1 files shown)


libgnucash/engine/iso-4217-currencies.xml
```diff
@@ -2414,6 +2414,7 @@
   local-symbol=""
 />
 <!-- "SZL" - "Lilangeni"
+  2019-02-16 : The plural "E" is the common local-symbol; Bug 797105
```
fellen: No date here, please. The dates in the comments mean 'date of replacement' followed by new currency and frozen rate.

sicelo: Thanks for clarifying. Fixed.

PR #462, opened 2019-02-21
Fixes to traceback calls in pycons

j00bar:
There is no call `traceback.print` - these are obviously mistakes. And they can even cause compile-time errors as `print` is a reserved function name in Python.


j00bar:
@wenottingham points out this was probably a search-and-replace failure from trying to upgrade to Python 3, when `print` began to require parentheses.

gjanssens:
Good catch, thanks a lot!

PR #463, opened 2019-02-21
[test-advanced-portfolio] initial commit

christopherlam:
Wish to put it out there for comment. I don't use APR so I wouldn't know exact ins & outs of this report.
Are the transactions set up correctly? 
Do we need many more buy/sell txns to thoroughly test the APR?

### DIFF
--- a/gnucash/report/reports/standard/advanced-portfolio.scm
+++ b/gnucash/report/reports/standard/advanced-portfolio.scm
@@ -225,128 +225,110 @@ by preventing negative stock balances.<br/>")
        (cons (cons (* units-ratio units) (* value-ratio value))
              (apply-basis-ratio rest units-ratio value-ratio)))))
 
-  ;; this builds a list for basis calculation and handles average, fifo and lifo methods
-  ;; the list is cons cells of (units-of-stock . price-per-unit)... average method produces only one
-  ;; cell that mutates to the new average. Need to add a date checker so that we allow for prices
-  ;; coming in out of order, such as a transfer with a price adjusted to carryover the basis.
+  ;; this builds a list for basis calculation and handles average,
+  ;; fifo and lifo methods the list is cons cells of (units-of-stock
+  ;; . price-per-unit)... average method produces only one cell that
+  ;; mutates to the new average. Need to add a date checker so that we
+  ;; allow for prices coming in out of order, such as a transfer with
+  ;; a price adjusted to carryover the basis.
   (define (basis-builder b-list b-units b-value b-method currency-frac)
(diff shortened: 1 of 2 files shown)


jralls:
@mtalexander I think you're the best choice to review this.

christopherlam:
This is partially stalled because I don't think it's possible to create a "Stock Split" (i.e. number of shares increases without any increase in value, or something) transactions with scheme APIs. Or I haven't found yet exactly how these are created.

mtalexander:
[AdvPortfolioTest.gnucash.zip](https://github.com/Gnucash/gnucash/files/3011925/AdvPortfolioTest.gnucash.zip)

From the point of view of the APR, a stock split is just a transaction that has a split in the stock's account with a zero value field and a non-zero amount field.  The function that creates these for the stock split assistant is gnc_stock_split_assistant_finish in assistant-stock-split.c.  This function does a few other things, but most of them are not relevant to the report.

In the simple case it creates a transaction with one split.  After setting the currency, date, and comment in the transaction, it adds one split to the transaction.  It sets the amount in the split to the number of shares gained or lost and then calls xaccSplitMakeStockSplit on it.  This sets the value to zero and marks the split as a stock split.  The APR currently doesn't use that mark.

The last thing it does on the split is call gnc_set_num_action on it.  This is a little complicated because of the new option to overload the use of the number field.  Depending on the parameters and the setting of the "Set book option on new files to use split "action" field for "Num" field on registers/reports" it will call either xaccTransSetNum or xaccSplitSetAction.  In the case we care about it will either call xaccSplitSetAction or do nothing depending on the setting of that option.  The parameters are (null, split, null, "Action Column|Split").

The gnc_stock_split_assistant_finish function may also create a price DB entry and perhaps add two more splits to the transaction to record cash in lieu of a fractional share.  The price DB entry is not relevant to the APR.  The cash splits are potentially relevant since it tries to handle them too, but there is nothing special about them.

All of the functions used by gnc_stock_split_assistant_finish are available in Scheme, so this code could be replicated there without too much trouble.  It would also be possible to factor out the code that creates the transaction and price DB entry from the code that handles the button event.

Does this help at all?  I've also attached a file I've used to test the Advanced Portfolio Report in case that is useful.

christopherlam:
@mtalexander thanks I managed to replicate the stock-split procedure.

I think I have now successfully completed 100% coverage for portfolio.scm -- will ping @jralls to confirm that I've understood the stock-split process well, especially cash-in-lieu instead of fractional shares. See `test-portfolios.scm` especially `stock-split` calls. advanced-portfolio tests are in progress.

PR #468, opened 2019-02-28
Croatian accounts correction

milotype:
Just noticed, that I forgott to change the "guid" string in these two files, as specified in the wiki: "Just replace all occurrences of <act:id type="guid"> by <act:id type="new">".

Only the croatian files for "eduloan" and "fixedassets" needed the change. Sorry for the inconvienience!


fellen:
Commited, thanks!

PR #469, opened 2019-03-03
Fix gnc-recurrence type

christopherlam:
Just a quick query for C types.

Looks like gnc-recurrence-new was wrongly typed. The fix suggested here seems to work. Is this correct?

Also why is this in gnome-utils... shouldn't it be moved to engine?


christopherlam:
Purpose: unless I can create a recurrence, I cannot test budgets.
Even with the above change, (gnc-recurrence-new) is still not accessible from scheme. Likely because it's part of gnome-utils.

christopherlam:
ok don't need this. still messy though.

PR #470, opened 2019-03-06
Toolar UI for Pending Scheduled Transactions

darthpale:
I use scheduled transactions a lot.  While I want to be aware of "pending" items, choosing the option to show the "Since last run" dialog box on open can be disruptive to my workflow  This change implements a tool bar item which show the number of scheduled transactions that either in the reminder or ready to create mode. 

Clicking on that button will open the "Since last run" dialog.

I've been using this feature since 1.95, and rebasing for years.  Thought I would give submitting them again.


jralls:
I'm inclined to say that the toolbar is already too big. I also doubt that this would have wide appeal.

gjanssens:
I agree with @jralls it's not a good fit to have on the tool bar all of the time and for every type of page. The tool bar space is precious and many tool bars are already overcrowded.

I do see it as a useful addition to the "Scheduled Transactions" page though. That's in the right context and that tool bar is not too crowded yet.

While that may not be enough for your use it would allow you to carry a smaller patch for making it more generically available from there.

jralls:
Closing as this seems to have been a drive-by. @darthpale by all means submit a new PR conforming to @gjanssens's recommendation if you're so inclined.

darthpale:
Sorry, I didn't know I needed to close it myself after the initial 
rejection.  The rationale was that its quite common for apps that have 
calendar based action items have some way of notifying the user when the 
item due.  Meeting reminders 15 minutes before meeting in calendaring 
apps etc.   I get the idea you want to do away with the toolbar, I 
didn't really though "get" the idea that it was overcrowded as even on a 
reasonably small window my toolbar has lots of unused space on the 
right. That being said, would putting the "pending" SX count on the 
status bar (which is only optionally displayed anyway based on settings) 
if the pending count > 0?  That would still make it so there was some UI 
cue that there are pending SX that need action, but without having the 
app always open the "Since Last Run..." dialog box which would need to 
be dismissed, if (for what ever reason) the user didn't want to actually 
deal with the pending SX right way.

jralls:
@darthpale You didn't need to close it necessarily, you needed to respond. You didn't, so after 3 months I closed it.

If you open GnuCash full-screen on a large monitor then there seems to be plenty of space on the toolbar. Based on the number of complaints we get it's pretty common not to do that. 

But this toolbar button isn't a button: It doesn't do anything, not even run the SLR calculation to update itself. It's just a display icon, and it would be the only one that GnuCash has. Text on the SX Editor status bar would fit better with the rest of GnuCash's UI.

Note as well that the numbers update only when the SLR dialog is run. You probably want to change that so that they're updated every time the SX Editor tab gets focus since the SLR dialog runs automatically only at GnuCash startup.

darthpale:
On 6/21/19 10:57 AM, John Ralls wrote:
>
> @darthpale <https://github.com/darthpale> You didn't need to close it 
> necessarily, you needed to respond. You didn't, so after 3 months I 
> closed it.
>
OK Thanks - your response in March seemed definitive so I assumed that 
was just end of it.

> If you open GnuCash full-screen on a large monitor then there seems to 
> be plenty of space on the toolbar. Based on the number of complaints 
> we get it's pretty common not to do that.
>
I don't do that either, but I get it...
>
> But this toolbar button isn't a button: It doesn't do anything, not 
> even run the SLR calculation to update itself. It's just a display 
> icon, and it would be the only one that GnuCash has. Text on the SX 
> Editor status bar would fit better with the rest of GnuCash's UI.
>
The button is not technically "display only" as clicking it does open 
the SLR dialog, but I get your point.

Putting the pending text only the status on the SX Editor status still 
requires the user to take an action to see how many are pending.

What I was try to accomplish was the following: alert in a non intrusive 
way if the user has pending SX that need attention, without the user 
having to take any specific action to .  The way it works now is that 
one ether needs to have the SLR dialog auto open each file open (at 
least in my use case, I don't want to necessarily have to interact with 
the SLR dialog if there are pending SX -- such as  having other TX to 
enter first), or have to open the SLR dialog to see if there are any.

> Note as well that the numbers update only when the SLR dialog is run. 
> You probably want to change that so that they're updated every time 
> the SX Editor tab gets focus since the SLR dialog runs automatically 
> only at GnuCash startup.
>
That's not accurate if I understand you correctly.  Any new or edited 
item from the SX editor will cause the number to be updated.   Just 
re-tested.  Currently I had 2 pending.  Opened the SX editor to increase 
the reminder days on another SX to and as soon as I confirmed the edit 
the pending count updated to 3.  I call to update it in all the places 
that at least in my use case made sense at the time.

TBS, if it's not useful/desirable on a wider scale then no worries.  
Fortunately it's easy to roll my own with a periodic git pull 
official/maint followed by a git merge.

At least I didn't just get on the user-list and complain about the 
things I don't like.  You do a good job.

Thanks.

jralls:
> The button is not technically "display only" as clicking it does open the SLR dialog, but I get your point.

Ah, I see it now.
> Putting the pending text only the status on the SX Editor status still
requires the user to take an action to see how many are pending.

As would having a button only on the SX Editor toolbar. Either way one has to switch to the SX Editor tab to see it. So what you really want would be to split off a bit of the status bar for your indicator.

>>Note as well that the numbers update only when the SLR dialog is run.

> That's not accurate if I understand you correctly.  Any new or edited item from the SX editor will cause the number to be updated. 

I wouldn't expect that to be a daily activity either. I meant that it doesn't have a timer event registered that looks for new pending SXes every midnight. Not all users restart GnuCash every day.

> The way it works now is that one ether needs to have the SLR dialog auto open each file open (at least in my use case, I don't want to necessarily have to interact with the SLR dialog if there are pending SX -- such as  having other TX to enter first)

You know it's not modal, right? You can just click the main window and do whatever and then get back to it when you're ready.

> TBS, if it's not useful/desirable on a wider scale then no worries...At least I didn't just get on the user-list and complain about the things I don't like.  

But that's the best way to find out if other users would also like the feature. And since you've already done most of the work you could lead off with "I've written this cool toolbar button that shows how many pending SXes there are and you can click it to open the SLR! Who else would like something like that?" If the general response is "Who needs that, what's wrong with just opening the SLR?" then you know you don't have much company, but if it's "Oooh! I want one!" then you have a good case to reopen the PR.

PR #471, opened 2019-03-14
Fix  TZ dependency in average balance reports (Bug 748431)

christopherlam:
This PR aims to deprecate `gnc:make-dr-cr-collector` and `gnc:make-stats-collector`.

The `analyze-splits` ~function from `average-balance` was copied nearly verbatim into `daily-reports` by wingo.~

~First it was deduped. Then~ converted stepwise to use scheme numbers via named-let loops instead of these collectors.

I was using test-average-balance to confirm analyze-splits behaviour was unchanged... however when the test is run in a different TZ the amounts differ; this corresponds to a longstanding bug whereby account period averages are sensitive to TZ.

Finally `gnc:make-dr-cr-collector` and `gnc:make-stats-collector` can both be deprecated.

Comments welcome -- I think the scheme named let loops is a much cleaner approach than using the original accumulators...

### DIFF
--- a/gnucash/report/report-system/report-system.scm
+++ b/gnucash/report/report-system/report-system.scm
@@ -686,10 +686,10 @@
 (export gnc:accounts-get-commodities)
 (export gnc:get-current-account-tree-depth)
 (export gnc:acccounts-get-all-subaccounts)
-(export gnc:make-stats-collector)
-(export gnc:make-drcr-collector)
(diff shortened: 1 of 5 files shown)


christopherlam:
I've decided that `analyze-splits` is a lost cause. Redo with named let; this seems to work well :)

jralls:
Thanks!

PR #472, opened 2019-03-19
[balance-forecast] Allow negative values for reserve and target

ZDBioHazard:
Whoops! I could have sworn I had allowed the reserve and target lines to be negative, but when I went to chart some liability accounts just now, turns out I missed it. >.>


christopherlam:
This is low impact and can be merged. My tiny issue is the interpretation of reserve/target in these negative amounts will need to be creative. Is one equivalent to a credit card limit? ;)

It can be merged. I'm restoring my dev environment after a silly mishap.

jralls:
Thanks!

PR #473, opened 2019-03-21
Bug 639049 - Asset Barchart Report includes also the first day of next month transactions

christopherlam:
If the original date is an end-of-month date, we take it as an indicator they always want month-delta dates to be end-of-months. This works for monthly/quarterly/halfyearly/annual. Also I'd forgotten to activate a test in test-date-utilities. Enable it.

This works for any month-end date, even leapyears.

monthly from 28/02/16 = 28/02/16, 28/03/16, 28/04/16, 28/05/16...
monthly from 29/02/16 = 29/02/16, 31/03/16, 30/04/16, 31/05/16...

Apparently human psychology is preferable to cold logic...

### DIFF
--- a/libgnucash/app-utils/date-utilities.scm
+++ b/libgnucash/app-utils/date-utilities.scm
@@ -215,6 +215,13 @@
 (define (gnc:time64-ge-date t1 t2)
   (gnc:time64-le-date t2 t1))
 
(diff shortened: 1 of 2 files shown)


jralls:
Thanks!

PR #474, opened 2019-03-22
[FUTURE] save dates as iso-string instead of time64

christopherlam:
This commit will:
1. change the date serialization routines from ("absolute" . time64) to ("absolute" . "yyyy-mm-dd").
2. introduce compatibility layers to read existing ('absolute . time64) dates.

Relative dates saved are unchanged ("relative" . "relative-date-symbol")

The in-memory representation of dates is still time64, parsed from strings into time64.

Hopefully saved-reports will no longer be sensitive to TZ changes.

This means future versions will read any previous saved-reports however previous cannot read future saved options. If this is agreed as a good change I'll incorporate into #459 :-)

### DIFF
--- a/libgnucash/app-utils/options.scm
+++ b/libgnucash/app-utils/options.scm
@@ -571,6 +571,7 @@ the option '~a'."))
          show-time
          subtype
          relative-date-list)
+  (define regex-iso "[0-9][0-9][0-9][0-9]-[0-9][0-9]-[0-9][0-9]")
   (define (date-legal date)
     (and (pair? date)
          (or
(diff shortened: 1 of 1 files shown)


jralls:
I don't have a problem with writing ISO8601 format date strings instead of time64, but...
1. We have date parsing routines in C++, prefer them to rolling your own.
2. All of the options code that's in Scheme is going to be rewritten in C++, don't add to what needs to be converted.
3. Since transaction dates are encoded as 10:59:00 UTC setting the interval begin as 00:00:00 UTC and the interval end as 23:59:59 UTC ensures that the intervals are consistent regardless of timezone. In other words it's easier to just take the timezone out of the equation than to recompute midnight for the timezone every time you run the report.

christopherlam:
Thanks this is more of a proof-of-concept to see how it works. The main issue is switching TZ seems to make some report dates change when saved/shutdown and load/restart with another TZ. The only C api I found converts time64 to full y/m/d/h/m/s/tz iso8601 which is overkill.

PR #475, opened 2019-03-23
Improve TR display/other-account and Sortkey=other-account renders

christopherlam:
A few scheme housekeeping commits.

Transaction Report can be amended to render other-account better in single-split and compound-txn cases. This amends display for both Display/Other-account-name/code, and Sortkey=corresponding-account-name.

image

### DIFF
--- a/gnucash/report/standard-reports/transaction.scm
+++ b/gnucash/report/standard-reports/transaction.scm
@@ -1157,14 +1157,18 @@ be excluded from periodic reporting.")
                                                      (column-uses? 'account-name)
                                                      (column-uses? 'account-full-name)))))
 
-               (add-if (or (column-uses? 'other-account-name) (column-uses? 'other-account-code))
+               (add-if (or (column-uses? 'other-account-name)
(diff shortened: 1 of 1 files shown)


christopherlam:
I'll remove "--None--" for single-split transactions.
I have a few other cleanups pending.
For multisplit transactions do we keep the terminology as "Split Transaction" ?

jralls:
Yes for now. It's a substantial change and IMO should be a major release change. @gjanssens seemed to agree with that on IRC the other day.

gjanssens:
> Yes for now. It's a substantial change and IMO should be a major release change. @gjanssens seemed to agree with that on IRC the other day.

While my agreement on the IRC discussion was about the term "Split" in itself the other discussion is about our artificial distinction between two-split and more-than-two-split transactions. That debate is still ongoing but if we can find better terms to make this distinction (the current favourites are "simple" vs "compound") that may be a less intrusive change that could potentially be done for 3.x somewhere already.

But in any case I'd wait until we have come to a final conclusion.

PR #476, opened 2019-03-24
Updated croatian hr.po for app and glossary

milotype:
Some wording corrections


fellen:
Thanks!

fellen:
Just to keep you informed: In between I had to change a few strings (mnenoics of menu entries, tooltips) in the code.

milotype:
?? Which files are you talking about? Anything that needs to be translated?

fellen:
after a fresh 'make pot;msgmerge ...' I get
```
grep -nr "fuzzy" hr.po
5447:#, fuzzy
14704:#, fuzzy
14725:#, fuzzy
18476:#, fuzzy, c-format
18503:#, fuzzy, c-format
18700:#, fuzzy, c-format
18705:#, fuzzy, c-format
18805:#, fuzzy
18816:#, fuzzy
18825:#, fuzzy
18862:#, fuzzy
18871:#, fuzzy
19070:#, fuzzy
19312:#, fuzzy
22010:#, fuzzy
22022:#, fuzzy
22347:#, fuzzy
22478:#, fuzzy
22750:#, fuzzy
25055:#, fuzzy
25070:#, fuzzy
25086:#, fuzzy
25091:#, fuzzy
25101:#, fuzzy
25116:#, fuzzy
25127:#, fuzzy
29201:#, fuzzy

```
In this sections the MsgId was changed and you should review the MsgStr's, see [wiki:Translation#The fuzzy flag](https://wiki.gnucash.org/wiki/Translation#The_fuzzy_flag)

milotype:
... probably changes between pot 3.4 and 3.5. My file was based on 3.4.

BTW, any chance of having current/valid `.pot` files within their dedicated folders on GitHub? This would be good to have in the other "subprojects" as well (like gnucash-htdocs/po)

That would be practical/great for people like me, who are no programmes, don't work with local branches, etc. As it is now, I always have to search for it on on https://translationproject.org/domain/gnucash.html

jralls:
Not for GnuCash. It's a build product and having build products in the repo causes trouble with using the repo; it's the source of "refused to switch a dirty tree" errors when you pull. Maybe we can get the nightly API documentation build to build it and upload it to [code](https://code.gnucash.org/docs).

Until then the TP is probably the easiest way to get a release's gnucash.pot. It's also in the release tarballs, but then you have to download all of the sources and extract them somewhere to get the potfile.

 I'm ambivalent about gnucash-htdocs. Is it really that hard for you to type ```make pot``` after pulling? Are there even frequent enough translatable changes to worry about it much?

milotype:
**gnucash**
So, I'll stick with TP for gnucash.pot. I'm fine with that. I've actually already updated my croatian .po with the 3.5 pot file, so I might do a pull request today.

**gnucash-htdocs**
1. I wouldn't know where to type `make pot`.  As mentioned i only work through the internet. I fork the needed "subprojects" to my account/repositories, then I make a new branch for my changes. I upload the files from my computer to the branch, and then make a pull request. No typing of any commands after that.
2. No, I don't expect frequent changes - but it's not on me. it depends on how often authors of the website make changes.
3. But ... I've noticed that something isn't working quite as it should. On the website, the last paragraph on the book of Ashok Ramachandran doesn't show the translations in different langauges (checked only de, hr, it), though the strings have been translated (at least when looking at the current .po files on the "master"). Interestingly a french translation is being shown, but neither the `fr.po` nor the french `.mo` files doesn't have the translation.
> Our own developer Christian Stimming has ...

milotype:
No, no, sorry, french doesn't work either ...

gjanssens:
That's probably because that paragraph has recently been changed by John and so the translations need updating.

jralls:
By "only through the internet" you presumably mean that you exclusively use github.com's web interface. 

WRT the Small Business Accounting book blurb, French doesn't work because it isn't translated. The others are though, and I don't see any difference between the msgid in the po files and the text in the webpage. I changed only the URLs and I did so in the po files as well as menu.php, see https://github.com/Gnucash/gnucash-htdocs/commit/ccc9c7d0a77345c815fb622fedc9ed818689caf7.

The anchor-tag in the string has an embedded double-quoted URL with the quotes escaped. I suspect the problem is that the potfile extraction doesn't recognize that the backslash needs to be tripled up so that gettext sees both the backslash and the embedded quote. I'll try that with German.

jralls:
No joy.
The right thing to do is to use php to format the string with the URL as a parameter. I'll try that next.

milotype:
Yes, you are wright, i > exclusively use github.com's web interface.

milotype:
When you're done, please let me know. I do have some minor corrections to do in the croatian translation on the website.
Translating is a strange thing – once you see your own translation "live" on a website, suddenly you realize, that the text doesn't transport the message as you intended :(

BTW, somewhat off topic – is there any interest/project to change/modernize the website's design? Even with some minor css-changes it could be improved. Should that be discussed through bugzilla?

jralls:
I'm done for now. I tried changing the string to "Our own developer Christian Stimming has been contributing to this book as a reviewer, and we can wholeheartly recommend the outcome of this fruitful collaboration. Feel free to have a look at this book over at %s, and, best of all, the publisher has committed to allocate some percentage of the book sales back to the GnuCash project. Have fun with this book!" and wrapping it in a ```sprintf()``` call with the URL as a parameter.  Still not translated and I made sure that the string in the po file matches the one in the phtml file.

We just had a redesign a couple of years ago. If you want to propose a new one the right place to start is the developer list, gnucash-devel@lists.gnucash.org.

fellen:
@jralls Either you forgot to run 'make mos' or to add de.mo to the repo.

fellen:
@jralls I believe a steps are menioned in https://wiki.gnucash.org/wiki/Website_Maintenance#Managing_Translations

fellen:
@milotype Which ways are offered by other projects, where you contribute and which of them do you prefer?

milotype:
@fellen Don't quite understand the question "Which ways are offered ..."
Do you mean how they provide `.pot` files?

fellen:
@milotype Yes - or in which ways translation is done there in general (web based, ...)?

milotype:
Most projects I have been working on, have .pot files directly within the `.po` folders. Others allow to make a copy of an exiting translation and use that `.po` file as a base (obviously they don’t have frequent changes in the code).
Then again there are those who use external translation platforms (Transifex, Crowdin, POEditor, etc.) with translation managers, reviewers and "plain" translators. I don't know if project owners pay fees to such platforms, for translators the accounts are usually free of charge.

Inkscape uses GitLab with subprojects (similar to GitHub). The `.pot` files are in folders, where the `.po` files are saved. Most translations concearning the **application** and **general web interface** are run through `.po` files from GitHub. The **content** of the website gets translated by registered translators directly on the website. I think they need that, because their website isn't as "static" as GnuCash's. For instance, they have frequent news. And they use Wiki mostly for developer stuff, not for user relevant issues.

milotype:
For me, it's the simplest when I find current `.pot` files within the po-folders.

milotype:
One more general question:
When do updated translations become part of the downloadable files on the website? For instance, now there is GnuCash 3.5. Will my croatian update be included in 3.5.1 or 3.6? Or does the current 3.5 version get updated without notice? (Nightly build, if that's what it's called)

milotype:
@jralls Your fix of the german translation of the "Our own developer Christian Stimming has …" seems to work now! I can see it on the website.

jralls:
That's because I missed a step as @fellen pointed out yesterday. I imagine that he did them and now it works. I'll propagate it to the other translations in the next few days as I have time.

As for downloadable files, we do [quarterly releases](https://wiki.gnucash.org/wiki/Release_Schedule) and [nightly builds for Windows](https://code.gnucash.org/builds/win32/maint). In both cases translations are updated at the next build after a translation is merged in our repository.

Then there's Linux. @gjanssens has an experimental nightly build in flatpak, but for the most part Linux users are at the mercy of their distribution packagers for updates and they're not very good about making GnuCash releases available to their users.

fellen:
The other languages should be fixed in commit adcaaff.

jralls:
@fellen, thanks for following though.

PR #477, opened 2019-03-26
reports: reports are saved in book kvp

christopherlam:
This commit will add a hook after file-load to grab the book custom-report-templates and replace the in-memory custom report templates. Please note this does not hook into the initial datafile loading... I'm not sure exactly where the initial file-load can be hooked. From experimentation if Gnucash is quit with a book-with-report, and relaunched, the book-reports may not be read... until File/Load exact same datafile.

saved-reports is now loaded in memory, and is *no longer updated*. I do not think there is any other place in code apart from `gnc:save-all-reports` which writes saved-reports, but I'm not sure.

I thought about updating saved-reports but this means that loading/saving book-saved-reports will keep modifying the in-memory custom-reports list, and can overwrite the saved-report and obliterate previous saved-reports.

book-reports saved in qof-book-option system. This is IMHO rather safe and does sanitize strings. I don't know how to guard book load/save with a feature.

I expect lots of scrutiny and discussion prior to merge :)

Task list:
- [x] find safe serialization for gnc-query2scm objects
- [ ] determine appropriate kvp path for reports & embedded-reports.


christopherlam:
oh and all saved-reports are still live scheme code. we *could* amend it, but to what? json? or nested-kvp paths? all are much the same imho and live scheme code has been debugged...

christopherlam:
fao @gjanssens re: 'custom-report' types -- I think I'll skip `generate-restore-forms` mechanism. I'll see about scanning a custom-report options list and save directly into kvp tree. I am fairly confident to succeed here.

re: 'saved-reports' ie tabbed-reports part is so far untouched. As you know the C part is still opaque to me :) - but we can already discuss where the settings will live... in the book? in the .gcm files?

christopherlam:
My current think is that book-reports are saved in kvp tree, using existing kvp<->scm code.

tabbed-reports are saved as json in .gcm files. I'll need to write serialize/deserialize functions.

Good?

christopherlam:
@gjanssens so as discussed if the nested-kvp is undesirable, the 'report: saved reports can be loaded from book' commit will load/save scheme-code into toplevel kvp '("custom-templates") path.

I'll try json serialization routines but this will require augmenting each and every option-type with a serialize/deserialize function...... they already do it internally so it'll be a matter of exposing them.

eg. https://github.com/Gnucash/gnucash/blob/maint/libgnucash/app-utils/options.scm#L382

christopherlam:
@gjanssens I think this is now complete. This set of commits will update infrastructure to save reports into book instead of saved-reports-2.8. The latter is read&loaded, but not updated anymore. If book has no saved-reports, the global saved-reports is used.

One issue: make-query-option uses gnc-query2scm which converts query to scm form. This is a 'complex' list. I have two options to serialize/deserialize:
1. use the preexisting value->string which converts to something like `"'(query-v2 (terms ((((\"book\" \"guid\") #f guid 3 1 (\"2aaa93e65adb416db7952e11acb6761a\")) ((\"account\" \"guid\") #f guid 3 1 (\"4c2d70ed8833402ca53a61f695e0529a\"))))) (search-for Split) (primary-sort ((\"QofQueryDefaultSort\") 0 #t)) (secondary-sort #f) (tertiary-sort #f) (max-results -1))"` i.e. a live string representation of query, which is easily decoded via `eval-string`.
2. unpack the above structure into component terms/search-for/primary-sort etc...... code it properly, and repack for reload.

Option 1 uses `eval-string` which is known to be evil. But easy to implement.
Option 2 is cleaner but assumes that I understand the scm encoding very well (I don't) and assumes the scm representation is permanent. A good example of option 2 is https://github.com/Gnucash/gnucash/commit/92ef26bd148c5e9dea57704f1c726fc05b43a0c5#diff-5c2239f1b11cd8b6c3a0f10d9bafe025R1495 - see how I'm decoding the input x to extract dateformat values.

I'm using option 1 for now. Any thoughts?

christopherlam:
Addendum I think option 2 is ultimately the better choice because the datafile containing report definition or a live string, even within the confines of a query2scm string, now poses a security risk. I think I *have* to understand the query2scm format 100% to properly pack/unpack it.

christopherlam:
The last idea about this is to save/load the report-options as a scheme string instead of json string.

Advantages - reduce dependency upon guile-json, no need to obey guile-json's idiosyncrasies-- as a result, simpler code...
Disadvantages - json string is more readable by C

I'll push a commit to show...

jralls:
Huge disadvantage: Reading and executing a scheme string means that arbitrary code can be injected into a data file and GnuCash will execute it. Granted we have that now from a different location but we should take this opportunity to fix it.

christopherlam:
I enquired with #guile yesterday, and we can neutralize any scheme code being evaluated with `read`. Assuming str is a malicious string:

```scheme
(define str "(display 'hello)(newline)")

(define R1 (eval-string str)) ==> will output 'hello'
(unspecified? R1) ==> #t

(define R2 (with-input-from-string str read)) ==> will not run code
(equal? R2 (list 'display (list 'quote 'hello))) ==> #t
```
The `read` call will not run code, and R2 will become a properly scheme-quoted, list of the first s-exp found in `str`

jralls:
OK, but in that case you could as easily just use a string of key1:value1;key2:value2;.... No need for Scheme code at all and perhaps less overhead.

christopherlam:
Agree however some options cannot be serialized easily e.g. make-query-option is something like `(list 'query-v2 ((((terms ...)))) (account . "xx") (search-for . "yy"))` produced by gnc_query2scm. I don't wish to decode this because I don't understand the structure, and there's no easy way to serialize this apart from the pair `(format #f "~s" scm)` and `(with-input-from-string serialized read)`.

wrt the current system being live scheme code, the risk is contained, but with book-reports I can craft a payload into a .gnucash file :smiling_imp:

jralls:
It's pretty straightforward, it's just pairs of predicate type and value, serialized by ```gnc_query2scm``` and deserialized by ```gnc_scm2query```. They're in ```libgnucash/engine/engine-helpers.c. ```

FWIW it's used only by the Register (a.k.a. General Journal) report and it's not exposed in the report options, so it may not need to be written out to file at all.

I guess I should also point out that we want to rewrite the scheme code in libgnucash/app-utils into C++ and to replace QofQuery with SQL queries, both ideally for GnuCash 4.

christopherlam:
I saw the pairs. It's the terms that baffle me; I cant figure out whether they're a simple nested-nested-list, and whether any list can be arbitrarily long. After all a query in my mind may have multiple and/or search terms.

If we're moving all away from scheme (thereby my area of control/concern will diminish :smirk: ) then I think either json or yaml will be better?

The json implementation is IMHO complete and seems to be unicode-safe; we just need to fix bug 797163 and I'll make changes so that we don't have to encode a list-of-guids in a leaf frame.

jralls:
QofQueries are simple tuples of {field, condition, value}. There's no grouping and the and/or is global to the whole query.

Since you've already done json let's stick with that.

fellen:
From https://lists.gnucash.org/logs/2019/04/02.html#T10:43:20:
I have the strange feeling, moving saved-reports into the book is a bad idea.
There is this old rule: keep data and data representation forms separate.

I think making them book specific is valid, but that could also done by replacing the user specific file by a report file aside of the book i.e. as <bookname>-reports.scm.

It is a feeling and I have no very good example: Lets say to fulfil governmental rules after closing the year I sign the file with a key and save it onto a DVD. Then we get new external reporting rules or I want to improve my internal reporting. Then I have to keep an new copy of the file although I might only have changed one line in the reports.

gjanssens:
@fellen Imagine a spreadsheet in which you have recorded some data in columns. And then in that spreadsheet you have also created a graph based on this data. Where will these parameters be stored ? In the same file I would think.

And the same reservations apply there as well: if you have to keep an unchanged version of the file for compliance, and the spreadsheet application moves on to an incompatible data format, you have a problem.

The same would go for our own data format, regardless of where we store report parameters. If we make an incompatible change (think for example flattening our kvp structure as we did for 3.x), the old file will have to be converted before it can be opened again with a newer version of gnucash.

I do take your point in general though: gnucash does not have a formal and stable data file format, nor a universal archival format. So if you want to be sure your signed version on DVD can still run, you'll have to save it together with a complete environment that can run the version of gnucash that last touched this file. A discussion of formalizing our data format or what an archival format would look like is beyond the scope of this change though.

And I don't think that discussion should block this change either. Storing the report parameters (not the reports themselves) with the data does have important benefits as well:
- several of these options are book specific, so storing them globally (as we have done so far) makes no sense and this has been the source of subtle bugs in the past
- adding them in the book directly solves the long-standing bug that sometimes a book gets saved but the reports aren't
- the reports will go with your book, so if you share your book for example with your accountant s/he will also get your reports instead of having to export those separately. (Note about the sharing: our concurrent multi-user restriction still applies of course).

christopherlam:
Re: tabbed-reports backward-compatibility can be done.

`gnc-plugin-page-report.c` modified as follows means that `gnc:json->report` can try convert the `option_string` into json; if this fails then `eval-string` it to recreate the tabbed-report. I'll see about a suitable neutralizing function...

```
-        scm_id = scm_c_eval_string(option_string);
+        scm_json_to_report = scm_c_eval_string("gnc:json->report");
+        scm_id = scm_call_1(scm_json_to_report, option_string);
```

fellen:
@gjanssens I think, such fundamental changes should not be decided by one or two devs here. Instead we should present it on gnucash-devel. Probably somebody has a much better idea or better arguments (pro and con)..

jralls:
@fellen, there's nobody on gnucash-devel who isn't also here with sufficient understanding of GnuCash's internals to contribute constructively. There *is* Wm Tarr who *thinks* he understands GnuCash's internals better than all of us put together, but do you really want to engage him? I sure don't.

With that out of the way, consider the reverse situation to the one you proposed; one I think is more likely: You have a report that you've configured to match some government requirement. You archive your data to CDR every year. The government changes the requirement and you modify your saved report configuration accordingly. Some years later you need to show the old data *including the report in the old format* to an auditor. If the report configurations are saved in the data file, no problem. If not and you foresaw the possibility and also archived your saved report configurations, still no problem. If not and you only archived your data file, you have to figure out again how to generate that report, which might be a problem.

The whole thing is a bit of a strawman though: GnuCash isn't for entities whose accounting systems are subject to audit. If there's a possibility that the auditor will want to see a report from n years ago, run the report and save it as a PDF at the end of the accounting period. The government auditor isn't going to be interested in your GnuCash file, nor would they be interested in

jralls:
I should add that I'm agnostic about whether the report configurations are part of the data file or if we store them in a separate file. I prefer that we not put them in slots: It's a breaking change regardless of how its done so there's no excuse for a kluge. They should have a properly designed set of elements in XML and a table in SQL. In fact it's an opportunity to get rid of slots and have properly designed elements and tables for everything, though demanding that would raise the bar so high that it would never get done.

christopherlam:
@jralls assuming we *don't* want live scheme in book, I suspect that json-in-kvp is probably the least hackish solution for now. I know myself I do prefer them in book for ease of everything - datafile backup etc. If a proper table with report-options API (xaccReportSetOption xaccReportGetOption etc) is published in the future, it's not difficult to read json and write API as a compatibility layer... The API will still need to handle serialization issues.

The benefits of this work above are:
- eradicate `generate-restore-forms` the scheme code generator for saved-reports and style-sheets which are loaded at startup
- which also means fixing style-sheets save mechanism which uses `generate-restore-forms`; I think I can make it save as json in style-sheets instead. I will think about backward compatibility later.

json is very convenient to serialize/deserialize both report-options and stylesheet-options.

If we need to save stylesheet-options as CSS then it means: (1) I have to learn the required CSS (2) we need to write CSS when saving stylesheet (3) we need to read CSS when loading stylesheet.

I'm sure you are aware scheme is incredible for reading and parsing streams of chars. But this means an additional layer of complexity that guile-json has solved already. I don't mind learning how to parse chars and convert these into scheme data structures to set stylesheet options. Is this a good direction to go? (see qif-import .scm files for example how messy it is!) (it is however comparatively easy to use `ice-9 peg` parser which is a late addition to guile)

The other alternative is to think of CSS stylesheets as a write-only mechanism, to be user-modifiable, loaded & inserted upon report-generation, and overwritten *only* when user edits styles and generates new ones. But I'd still prefer to *not* write a reader and use guile-json instead, because I don't have experience writing a reader.

gjanssens:
I also prefer to see this implemented as regular xml nodes/sql tables rather than slots. It will require a bit more work on the C(++) side, but I don't think it is *that* much more. Most of the work in the PR can be reused for that IMO.

As for stylesheets in css, I agree this is a bigger challenge. It would require an overhaul of the complete style system in reporting: take all styling we do on all reports, try to make a common css file based on it, make all reports use it. Then decide whether we want to leave it a that (and only allow users to provide custom css for style changes) or figure out how we can provide a gui that will allow users to customize their style sheets (that gui should be then probably be able to parse existing css to provide the user with a representation of the current styling). We could also limit the gui to only a few styling options, or only offer a way to import/export css styles written by other users. It is in any case challenging as it essentially means ripping out the current "stylesheet" system and replacing it with a new one based on css. Something that should eventually happen IMO as that's the native styling used for html. However it's probably better considered as an independent project from loading/saving report presets/style options directly to/from a data file.

Having said that and as the end goal IMO is having a css based style system, I prefer not to make the intermediate step to json for style sheets as that would mean we have to introduce data file incompatibility twice for one data type.

christopherlam:
> I also prefer to see this implemented as regular xml nodes/sql tables rather than slots. It will require a bit more work on the C(++) side, but I don't think it is _that_ much more. Most of the work in the PR can be reused for that IMO.

The infrastructure is ready. The dilemma will be how to serialize each option, because they're all different. I've created `gnc:options-list->scm` and `gnc:scm->options-list` to package all scm string/bool/number/list *modified* options into a form suitable for guile-json, and vice-versa.

Re: style-sheets I'll need to think about that as well. Maybe I'll complete the json implementation as a reference, and try combine the stylesheets separately.

My personal issue is that I'm not sure I can allocate a lot of time for this in 12 months time or so :wink:

jralls:
I haven't had time to look at the code for this PR before today, and I'm sorry to say that the philosphical  discussion over JSON vs. CSS and storage on the book object vs. KVP has been a bit of a distraction.

I have a much bigger problem with this PR: It implements the storage functions in Scheme and calls those functions from C, making the options system's C<->Scheme bouncing problem worse. Storing the saved report configurations in the book is worthy, but not worth increasing technical debt that severely.

christopherlam:
@jralls fully aware - at this stage the point of this PR is to eliminate scheme writing scheme and determine the right data structures to use, a blueprint of sorts to know what needs to be saved and in what format. I'd think implementing this work in C is not that difficult. I'll park this one now, and move on.

christopherlam:
> I have a much bigger problem with this PR: It implements the storage functions in Scheme and calls those functions from C, making the options system's C<->Scheme bouncing problem worse. Storing the saved report configurations in the book is worthy, but not worth increasing technical debt that severely.

FWIW this set of commits will actually dramatically reduce the technical debt. I'll push a final commit soon whereby I can demonstrate numerous scheme-writing functions can be removed without affecting the saved-reports facility, nor removing backwards compatibility for loading saved-reports-2.8. This is in addition to the deprecated functions already identified. (in other words I am confident that new-lines < deleted-lines).

The scheme functions being called are not actually new code -- they are rewritten gnc:report-serialize into proper scheme, and the prior code was doing a simple `eval-string` on the generated scheme code hence was very concise, but now is deserialized safely.

The report.scm code will handle all report templates, whether standard or custom, and IMO the report-template serializing/loading should be done in the same language...

The main new requirement for this line of work is (1) guile-json for serializing and (2) access to a storage mechanism, currently chosen to be kvp, but can use any other mechanism. When options.scm is converted to C IMHO it must still have a serialization mechanism e.g. an open Register report can be saved, which will require the gnc-query being serialized... and this will be rather difficult.

So, not pushing for an early commit here, but I'd wager that C code will pose very similar challenges as this branch.

In summary I think:
- report.scm must handle loading/saving of report-templates (to be called from C) because it accesses the report-templates directly
- html-style-sheets.scm must also read/write saved style sheets.
- options.scm can be rewritten in C but must expose a serialization/deserialization method

jralls:
> IMO the report-template serializing/loading should be done in the same language...
You need to rethink your opinion. Guile isn't capable of doing that, particularly if it's going into the book or involves transcoding. Since some options are stored in the book it has to be C/C++. Since on Windows any file path operations involve transcoding (because Windows uses UTF16 for filepaths and GnuCash uses UTF8), Guile can't do that either. (This latter is entirely the fault of the Guile developers who route transcoding through a port encoded in the system code page. It might have been fixed in Guile-2.2, but since I haven't managed to get that to actually work on Windows that's for the moment moot.) 

Your Guile-json argument is circular: You insist on using json because Guile-json is a readily-available serialization library and then insist that that makes Guile-json mandatory.   [Guile-json is also not widely supported in distros](https://repology.org/project/guile-json/versions). That is as likely to get us dropped as to get guile-json added, not so good. Considering the minimal unit tests your claim that it's well-tested is at best debatable. It's also a single-author project with a rather spotty development history.

The book is read and written using the C++ backends. Options are stored in a C g_hashtable because Scheme historically didn't do data structures. It does now have KVP lists that can work something like hashtables, but either way for the option system to read and write the book it needs a C++ interface so nothing is gained by converting it. On the other end the GUI is also in C, so we have a ton of code that goes C->guile->C. You want to add more. Even if it's fewer LOC it's still more debt.

Same thing with stylesheets and html: There's no good reason for either of them to be in Scheme: The StyleSheet Editor is in C. It could save a lot of work by just reading and writing the CSS classes, keys and values; the Scheme report generator just needs to stuff them as-is into the ```<style>``` element. The stylesheets can be stored as CSS or YAML directly in C, there's no good reason for Scheme intermediation, it just adds extra steps to convert the strings in and out of Scheme and makes it harder to understand the code.

christopherlam:
Ok thank you for the roadmap... I admit I can't understand the C part of it so I'll have to defer to your higher strategic thinking.

Let me know what I can do to ease though... I mentioned before I may not be able to maintain productivity here for too long. I have it in my sights to merge stylesheets/*.scm similarly to invoice.scm.

christopherlam:
This is an example (prettified) book.gcm SchemeOptions prior to this commit; it's the register.scm report generated from a live register, applied some filters, and run Reports > Account Report. It illustrates how `gnc-query2scm` option is serialized. The next snippet is the equivalent JSON object. The account parameters are:
- selected account
- filter on reconciliation reconcile/unreconcile/void
- sort by date

```scheme
(let ((options (gnc:report-template-new-options/report-guid "22104e02654c4adba844ee75a3f8d173" "Register")))

  (let ((option (gnc:lookup-option options "General" "Title")))
    ((lambda (option) 
       (if option ((gnc:option-setter option) "Assets:EUR cash - Transaction Report")))
     option))

  (let ((option (gnc:lookup-option options "__reg" "credit-string")))
    ((lambda (option)
       (if option ((gnc:option-setter option) '"Decrease")))
     option))

  (let ((option (gnc:lookup-option options "__reg" "debit-string")))
    ((lambda (option)
       (if option ((gnc:option-setter option) '"Increase")))
     option))

  (let ((option (gnc:lookup-option options "__reg" "query")))
    ((lambda (option)
       (if option
           ((gnc:option-setter option)
            '(query-v2 (terms (((("book" "guid") #f guid 3 1 ("581d833e6c6a429c94e9a1ecb2957001"))
                                (("account" "guid") #f guid 3 1 ("33e1638860624ac99caff1d15f43d3e5"))
                                (("reconcile-flag") #f character 3 1 "ynv")
                                (("trans" "date-posted") #f date 5 1 1517414400))))
                       (search-for Split)
                       (primary-sort (("QofQueryDefaultSort") 0 #t))
                       (secondary-sort #f)
                       (tertiary-sort #f)
                       (max-results -1)))))
      option))

  (if (defined? 'gnc:restore-report-by-guid-with-custom-template)
      (gnc:restore-report-by-guid-with-custom-template 1 "22104e02654c4adba844ee75a3f8d173" "Register" "7a167cf9462148d2956c1b3fa93a4448" options)
      (gnc:restore-report-by-guid 1 "22104e02654c4adba844ee75a3f8d173" "Register" options)))
```

```json


{
  "guid": "22104e02654c4adba844ee75a3f8d173",
  "name": "Register",
  "template": {
    "name": "Register",
    "parenttype": false,
    "menupath": false,
    "sub-reports": [
      
    ],
    "options": [
      [
        "__reg",
        "query",
        "(query-v2 (terms ((((\"book\" \"guid\") #f guid 3 1 (\"581d833e6c6a429c94e9a1ecb2957001\")) ((\"account\" \"guid\") #f guid 3 1 (\"33e1638860624ac99caff1d15f43d3e5\"))))) (search-for Split) (primary-sort ((\"QofQueryDefaultSort\") 0 #t)) (secondary-sort #f) (tertiary-sort #f) (max-results -1))"
      ],
      [
        "__reg",
        "debit-string",
        "Increase"
      ],
      [
        "__reg",
        "credit-string",
        "Decrease"
      ],
      [
        "General",
        "Title",
        "Assets:EUR cash - Transaction Report"
      ]
    ]
  }
}
```

jralls:
Maybe we could finesse the licensing issue by doing the json serialization in C++. [Boost::Property Tree](https://www.boost.org/doc/libs/1_70_0/doc/html/property_tree.html) is a possible substitute though it apparently uses std::string and allocates a lot, which can cause performance problems. [RapidJSON](https://github.com/Tencent/rapidjson/) seems well regarded and uses a so-called MIT license for itself and the BSD license for one included dependency and the JSON license for another.

christopherlam:
To clarify this PR is about saving saved-report configurations in book's kvp tree rather than saved-reports-2.8 -- it is heavily scheme based because that's what I know. From my understanding it's fully functional.

christopherlam:
A thought regarding `saved-reports-2.8` which is now *read-only*.

As of this message, this set of commits will, upon file-load, run `gnc:load-book-custom-templates` -- the gist of this function is: 
- first it queries if book-reports exist:
- if yes, remove saved-reports-2.8 custom-reports, and load book-reports
- if no, revert to global saved-reports-2.8

A missing behaviour is -- if user has saved-reports-2.8, loads datafile, deletes all saved-reports, and saves datafile, the saved-reports-2.8 is now completely inaccessible. The book has already been 'tainted' with an empty book-reports list. It's impossible to restore saved-reports-2.8 reports. If this functionality is desired, more UI must be built.

christopherlam:
Some notes on backward compatibility so far:
* Tabbed-reports loading from .gcm is does not currently load existing live-scheme-code. To load previous tabbed-reports, we need to reactivate the `SchemeOptions=(live-code)` handling; I've chosen `JSONOptions=...` for json-tabbed-reports instead otherwise json/scheme clash badly when switching builds.
* stylesheets-2.0 rename to stylesheets-3.0; I think I've removed loading previous style-sheets.
* saved-reports-2.8 now read but never written.

christopherlam:
@derekatkins tried this.
If you open an old datafile you'll get old reports. No good.

derekatkins:
@christopherlam  --
> @derekatkins tried this. If you open an old datafile you'll get old reports. No good.

I honestly don't remember trying this...

christopherlam:
> @christopherlam --
> 
> > @derekatkins tried this. If you open an old datafile you'll get old reports. No good.
> 
> I honestly don't remember trying this...

Oh I meant if the custom reports are stored within the data file, then the above behaviour will happen. Hence this idea in this pr was dropped.

christopherlam:
> @derekatkins tried this. If you open an old datafile you'll get old reports. No good.

By _you_ I meant the user, I didn't mean you Warlord specifically 🤦‍♂️

gjanssens:
> @derekatkins tried this. If you open an old datafile you'll get old reports. No good.

Why is that a reason to close this PR ?

I can see this is undesired behaviour. You can perhaps avoid this by keeping the old mechanism of writing to saved-reports in place for at least one major release cycle. So when someone opens the book in an older version of gnucash, it will also see the updated reports.

What you still need then is a way to detect if a user made report configuration changes while using the older gnucash version. I would propose to write some no-op marker to the saved-reports file when using the new gnucash version. This could be a comment even. A report change in an older gnucash version will result in the saved-reports file being rewritten without this marker. So now when you open the book again in the new gnucash, you will know whether to re-read the saved-reports file.

PR #478, opened 2019-03-26
balsheet-pnl ongoing work.

christopherlam:
all commits from scratch.
lots of branching out, backtracking, refactoring.
added reporting of unrealized-gains and retained-earnings as part of equity in balsheet.
numbers seem to match formal balance-sheet.scm
a bit messy now :)

Task List
- [ ] refactor to break out lambdas


christopherlam:
formatting were set to 'my' preferences ;) if this is merged, I'd guess all commits will be fixup together. that's fine. I can't remember the earlier commits anymore. I *may* be able to upgrade `add-multicolumn-acct-table` to be used by other reports... I dare say it's more capable than gnc:make-html-acct-table that many reports use.

ggam:
The income statement with monthly period duration is absolutely great! The only thing I miss is the option to filter accounts by origin/destination like the Transaction reports does.

I enter my salary with taxes as a split transaction, but for this monthly income/expenses report I'd like to see only the net income. I can exclude the taxes expenses account, but then the salary account still shows the total gross income including taxes.

Would it be possible to add that option?

PR #405, opened 2018-08-27
Bug 796778

DaveC49:
Updated version of the implementation of Multiple selection in the import_main_matcher.c (https://bugs.gnucash.org/show_bug.cgi?id=796778).  In the matcher, select multiple rows using Ctrl-Click or Ctrl_Shift then either right click or ShiftF10 to bring up a popupmenu with Select transfer account.Click on that to assign a selected transfer account to all of the selected transactions. The click OK as usual to import into the accounts.

PR #479, opened 2019-03-26
Aqb6

jralls:
Martin Preuss's changes to support AQBanking version 6. Martin had based his patches on master, so this includes rebasing to maint. There were a few conflicts that I resolved related to removing Ktoblzcheck and converting the transaction datestamp to neutral time.

### DIFF
--- a/README.dependencies
+++ b/README.dependencies
@@ -85,8 +85,8 @@ Libraries/Deps
 
   optional
   --------
-  aqbanking		4.0.0			 online banking; also requires
-						 gwehywfar and ktoblzcheck
+  aqbanking		5.3.4			 online banking; also requires
+						 gwehywfar.
 
   libdbi		0.8.3			 SQL backend; also requires at
 						 least one of libdbd-sqlite3,
--- a/gnucash/import-export/aqb/gnc-ab-utils.c
+++ b/gnucash/import-export/aqb/gnc-ab-utils.c
@@ -29,15 +29,18 @@
 
 #include <config.h>
 
+#include "gnc-ab-utils.h"
+
 #include <glib/gi18n.h>
 #include <gwenhywfar/gwenhywfar.h>
 #include <aqbanking/banking.h>
-
+#ifdef AQBANKING6
+# include <aqbanking/types/balance.h>
+#endif
 #include "window-reconcile.h"
 #include "Transaction.h"
 #include "dialog-ab-trans.h"
 #include "gnc-ab-kvp.h"
-#include "gnc-ab-utils.h"
 #include "gnc-glib-utils.h"
 #include "gnc-gwen-gui.h"
 #include "gnc-prefs.h"
@@ -73,11 +76,11 @@ struct _GncABImExContextImport
     guint awaiting;
     gboolean txn_found;
     Account *gnc_acc;
-    AB_ACCOUNT *ab_acc;
+    GNC_AB_ACCOUNT_SPEC *ab_acc;
     gboolean execute_txns;
     AB_BANKING *api;
     GtkWidget *parent;
-    AB_JOB_LIST2 *job_list;
+    GNC_AB_JOB_LIST2 *job_list;
     GNCImportMainMatcher *generic_importer;
     GData *tmp_job_list;
 };
@@ -230,10 +233,10 @@ gnc_AB_BANKING_fini(AB_BANKING *api)
     return 0;
 }
 
-AB_ACCOUNT *
+GNC_AB_ACCOUNT_SPEC *
 gnc_ab_get_ab_account(const AB_BANKING *api, Account *gnc_acc)
 {
-    AB_ACCOUNT *ab_account = NULL;
+    GNC_AB_ACCOUNT_SPEC *ab_account = NULL;
     const gchar *bankcode = NULL;
     const gchar *accountid = NULL;
     guint32 account_uid = 0;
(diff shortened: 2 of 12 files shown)


gnucash/gtkbuilder/dialog-import.glade
```diff
@@ -14,6 +14,9 @@
     <property name="default_height">600</property>
     <property name="type_hint">dialog</property>
     <signal name="map" handler="gnc_ui_generic_account_picker_map_cb" swapped="no"/>
+    <child>
```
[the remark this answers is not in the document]
Bob-IT: I have in the passed removed them in a text editor if it was unrelated to a change but I see you are using a newer glade version, need to get that version and see.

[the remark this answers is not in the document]
jralls: This is actually an artifact of a difference between master and maint, totally unrelated to Martin's work. It's from https://github.com/Gnucash/gnucash/commit/cae8ecde8f9ddd615ff47c783ce8136395bd71a9, part of PR #405. It seems to have gotten applied to master instead of maint and I need to figure out why and probably get it into maint.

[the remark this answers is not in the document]
gjanssens: I believe PR #405 went on master as it implements a feature request and at the time of that PR I was still advocating to have new features on master instead of maint. There were too many moving targets in our effort to stabilize the 3.x branch to add random new features. I think I'm ready to relax this a bit as 3.x is now in a better shape than it was back then.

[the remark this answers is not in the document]
jralls: OK, but not this time.
After working through separating the two PRs I looked at Martin's changes and decided that while it's big it would be too much code-copying to make a separate directory so I'm hand-applying the changes with ifdefs after having cleaned out all of the old pre AQBanking 5.3 ifdefs. That's the version in ubuntu 14.03LTS so there's no reason to drag around the older code.

The whole thing needs a refactor, there's a lot of code duplication, but there isn't time for that this week. I don't think there's really time for a proper code review either so I'm just going to get it building both ways and jam it in. More technical debt, sigh.

gnucash/import-export/aqb/gnc-ab-utils.c
```diff
@@ -276,25 +228,30 @@ gnc_ab_get_ab_account(const AB_BANKING *api, Account *gnc_acc)
...
+        gint rv;
+
+        rv = AB_Banking_GetAccountSpecByUniqueId(api, account_uid, &ab_account);
 
-        if (!ab_account && bankcode && *bankcode && accountid && *accountid)
+        if ( (rv<0 || !ab_account) && bankcode && *bankcode && accountid && *accountid)
         {
             g_message("gnc_ab_get_ab_account: No AB_ACCOUNT found for UID %d, "
                       "trying bank code\n", account_uid);
+#if 0 /* TODO */
```
jralls: I find this and the following item a bit distressing. It implies that Martin intends to reimplement AB_Banking_GetAccountByCodeAndNumber(). This breaks the function for AQB6 and ISTM means that trying to support AQB6 for GC3.5 is perhaps misguided.

m53956: I don't intend to remove account lookup by bank code and account id, I just wanted to get it done quickly to see whether it could be done and improve later (in case there was any interest in my changes, which didn't seem to be the case until now, so I didn't continue to work on it).

I chose to start with lookup by unique id because the combination of account number and bank code is very much outdated, at least in Europe, since there is only IBAN now. Banks seem to support account numbers for some time, but that will eventually go away.
Also, some banks in Germany chose to use the same bank code and account number for different accounts, so this combination is not unique with those banks.

Anyhow, in AqBanking6 there already is a function which can use several account attributes to find matching accounts (currently IBAN, BIC, account number, bank code, account type, currency etc.) which should be used here.

m53956: Regards
Martin (AqBanking)

jralls: Martin, 
Great, thanks. What's the function?

jralls:
OK, completely redone to work with both AQB 5 and 6, builds cleanly on both. Support for versions earlier than AQB 5.3.4 is removed.

fellen:
update of CMakeLists.txt, README.dependencies and https://wiki.gnucash.org/wiki/Dependencies?

fellen:
Once we started here: https://bugs.gnucash.org/show_bug.cgi?id=796956- Aqbanking 6 will drop DTAUS

I assume 

> File/import/Dataus [& send]

 will fail then?
Sorry, I have no example file at hand.

jralls:
I had a local branch-naming that prevented my force-push from replacing Martin's. Fixed now, includes CMakeLists.txt and README.dependencies. I've also updated the wiki page.

README.dependencies
```diff
@@ -85,8 +85,8 @@ Libraries/Deps
 
   optional
   --------
-  aqbanking		4.0.0			 online banking; also requires
-						 gwehywfar and ktoblzcheck
+  aqbanking		5.3.4			 online banking; also requires
+						 gwehywfar.
```
gjanssens: Just to be clear here: does this mean ktoblzcheck is not needed anymore at all ? I understood it only to be dropped for aqbanking 6.

gjanssens: Or did you remove it as it's a dependency of aqbanking and not a direct dependency of gnucash ?

jralls: I suppose it's no longer needed, @cstim removed it entirely in https://github.com/Gnucash/gnucash/commit/dac94ce03806d5791ac7a8a0ffcba03896e8c237#diff-7d8b963127def26b06716af6e17b35d0.

gjanssens: Oh, right. I forgot about that bit. So yay, one dependency less!

PR #480, opened 2019-04-04
Pr python2to3 str methods gnc jinja

c-holtermann:
Work on python example scripts. Change str_methods.py and gnc_jinja.py from python 2 to python 3. Some improvements to gnc_jinja.py.


jralls:
Thanks!

PR #481, opened 2019-04-04
Pr python2to3

c-holtermann:
I ran the python 2to3 utility on all the example files. Shebang has been modified to reflect the step to python3.

### DIFF
--- a/bindings/python/example_scripts/account_analysis.py
+++ b/bindings/python/example_scripts/account_analysis.py
@@ -117,7 +117,7 @@ def next_period_start(start_year, start_month, period_type):
 def period_end(start_year, start_month, period_type):
     if period_type not in PERIODS:
         raise Exception("%s is not a valid period, should be %s" % (
-                period_type, str(PERIODS.keys()) ) )
+                period_type, str(list(PERIODS.keys())) ) )
 
     end_year, end_month = next_period_start(start_year, start_month,
                                             period_type)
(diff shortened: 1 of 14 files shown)


jralls:
Thanks!

fellen:
Shouldn't we update https://github.com/Gnucash/gnucash/blob/c38c4557a56370a51236903e326f4429fd5b2caf/CMakeLists.txt#L477-L478
this to 3.2, too?

jralls:
Yes.

fellen:
But which one:
https://github.com/Gnucash/gnucash/blob/7b1a262eea619dfea7126c3ef5dbd3e252a65f1a/README.dependencies#L97
or https://wiki.gnucash.org/wiki/Python_Bindings#Setting_things_up (3.2)
And how is https://ironpython.net/ (Latest 2.7.8) related?

BTW, I fear I messed up https://wiki.gnucash.org/wiki/Python_Bindings#Mac_OSX by replacing 2[.]7.

jralls:
Looks like the answer for Iron Python is [don't](https://github.com/IronLanguages/ironpython3). I guess we should remove iron python from the wiki until that's ready to use and somebody feels motivated to try it.

The only problem with the MacOS bindings is that the PYTHONPATH depends on the actual version of python. We require >=3.2 but most folks will now have 3.6 or 3.7.

c-holtermann:
Python 3.2 or 3.4 ? The version should be the same in the CMake checks and README.dependencies. Otherwise I don't know if 3.4 is necessary. Change the README.dependencies to 3.2 ?

fellen:
I think, the 3.4 was a typo in https://github.com/Gnucash/gnucash/commit/36c771b156f5147d142599bf524f12cd8e98c31d#diff-eefe61d1e670296756c590c5164f7130L94 and have updated both to 3.2.

PR #483, opened 2019-04-06
Rewrite price calculators

christopherlam:
This is upcoming... I'm trusting that `test-commodity-utils` has ~nearly~ 100% coverage of the relevant price calculators... ~if not I'll analyse in the near future and add more tests.~

Rewrite loops, cleaner IMHO and uses scheme rational arithmetic. `set!` is banished. Instead of building price-lists and removing invalid prices, build price-list using valid prices only. Fewer conditionals. Tail-calls. As a result these loops should allocate vars on stack rather than heap.

Pending:
- [x] confirm 100% test code coverage
- [ ] consider renaming vars to show clearer intent.
- [ ] more documentation how these calculators work
- [ ] in future - consider using scheme numbers instead of value-collectors - these `cdadr` are annoying. would need amending test suite.

### DIFF
--- a/gnucash/report/report-system/commodity-utilities.scm
+++ b/gnucash/report/report-system/commodity-utilities.scm
@@ -484,75 +484,80 @@ construct with gnc:make-gnc-monetary and gnc:monetary->string instead.")
 ;; amount total of 2000 GBP and a value of 2400 EUR. Returns a report-list.
 
 (define (gnc:get-exchange-totals report-commodity end-date)
-  (let ((curr-accounts
-         (gnc-account-get-descendants-sorted (gnc-get-current-root-account)))
-        (sumlist (list (list report-commodity '()))))
+  (let ((curr-accounts (gnc-account-get-descendants-sorted
+                        (gnc-get-current-root-account))))
 
-    (if (not (null? curr-accounts))
-        ;; Go through all splits and add up all value-amounts
-        ;; and share-amounts
-        (for-each
-         (lambda (a)
-           (let* ((transaction-comm (xaccTransGetCurrency
-                                     (xaccSplitGetParent a)))
-                  (account-comm (xaccAccountGetCommodity
-                                 (xaccSplitGetAccount a)))
-                  ;; Always use the absolute value here.
-                  (share-amount (abs
-                                 (xaccSplitGetAmount a)))
-                  (value-amount (abs
-                                 (xaccSplitGetValue a)))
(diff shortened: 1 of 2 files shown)


jralls:
@christopherlam Having now learned about named let loops, I'm concerned about C functions in tails. It *looks* to me that the only place you have them is in the warnings, which are terminal and so OK. Am I right?

In general, this new code is much easier to follow. Good work.

christopherlam:
I *think* the `(warn)` calls are terminal because they are triggered when there are no suitable prices, so, the next statement with `(cdr commodity-splits)` *should* be null. I'll recheck tonight.

(aside, I wouldn't think that needing to emit a warning *during* splitlist scanning is necessarily the best code... I think the cleanest would be to test commodity vs splitlist and *not* call the price-calculators if the commodity is missing... so, if I can make it work I'll see about that in the future).

jralls:
@christopherlam The question is are any *other* C functions in "tail position", see the bit about stack frames in [control flow](https://www.gnu.org/software/guile/manual/html_node/Control-Flow.html) and consider a loop with many thousands of iterations.

christopherlam:
From my understanding, all named-let loops are being called properly with tail calls, because they're all using the general template (also heavily used in guile source)
```scheme
(let loop args
 (cond
  (pred?
   ...
   (loop ...))))
```
compared with the non-tail-call format
```scheme
(let loop args
 (cond
  (pred?
   ...
   (some-function (loop ...)))))
```

jralls:
Very good then.

christopherlam:
Just wondering what's the general feeling regarding unfinished task list above? Is it worth the effort to complete them all?

jralls:
They're all worthwhile, but is there more worthwhile stuff to spend your time on? That's something of a personal decision for you: My own priorities bounce between fixing significant bugs (where significant means crash, loss of important function, or lots of users whining) and advancing on the long-term goals. I do code cleanup and technical debt paydown when I can combine it with the first two. Those don't have to be *your* priorities.

christopherlam:
Ok my primary motivation is that a project of this importance should not have embarrassing code :)
Perhaps you can guide me in which direction to apply effort within scm files to ease long-term goals e.g. merge style-sheets/*.scm, create options.scm tests, or elsewhere?

jralls:
I partly tested this by running the trial balance report on a large file. There was no significant speed difference, so you can go ahead and merge when you're ready. Just one thing, please squash 8c4b75d and  f2403d1 into 49e3cc8  first.

PR #485, opened 2019-04-06
PR python bindings test: extra test target to create test dependencies

c-holtermann:
As described in https://bugs.gnucash.org/show_bug.cgi?id=797181 running make test on a blank build leads to missing dependencies for the python bindings. On Stackoverflow I found a suggestion to include an additional test to build the actual tests dependencies: https://stackoverflow.com/a/10824578/3720561.


c-holtermann:
John Ralls wrote in the bug report that make test doesn't create dependencies while make check does. This PR would introduce a different behavior. So this should rather be a draft as basis for discussion or closed and discussion continued on the bug report.

jralls:
OK, let's close this and discuss on the bug.

PR #487, opened 2019-04-15
Scheme progress

christopherlam:
- Various cleanup efforts
- Avoids string-ports which munge Unicode in Windows

### DIFF
--- a/gnucash/report/stylesheets/stylesheet-easy.scm
+++ b/gnucash/report/stylesheets/stylesheet-easy.scm
@@ -4,23 +4,23 @@
 ;;
 ;; Copyright 2004 James Strandboge <jstrand1@rochester.rr.com>
 ;;
-;; This program is free software; you can redistribute it and/or    
-;; modify it under the terms of the GNU General Public License as   
-;; published by the Free Software Foundation; either version 2 of   
-;; the License, or (at your option) any later version.              
-;;                                                                  
-;; This program is distributed in the hope that it will be useful,  
-;; but WITHOUT ANY WARRANTY; without even the implied warranty of   
-;; MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the    
-;; GNU General Public License for more details.                     
-;;                                                                  
+;; This program is free software; you can redistribute it and/or
+;; modify it under the terms of the GNU General Public License as
+;; published by the Free Software Foundation; either version 2 of
+;; the License, or (at your option) any later version.
+;;
+;; This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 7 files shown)


christopherlam:
why schemers build lists in reverse order
https://stackoverflow.com/questions/51279697/how-efficient-is-guile-schemes-reverse-function

jralls:
Ah, and a string is just a list-of-wchar_t. Figures. Prepend-and-reverse is a common strategy for linked lists regardless of language... but in normal languages strings are contiguous memory, not splattered all over the place and plastered together with links.

christopherlam:
> Ah, and a string is just a list-of-wchar_t. Figures

Note quite... https://www.gnu.org/software/guile/manual/html_node/Strings.html and http://git.savannah.gnu.org/cgit/guile.git/tree/libguile/srfi-13.c#n283 -- strings are still pointer and memory. `string->list` written in C shows there is a boolean property `sci_i_is_narrow_string` which generates a list of char instead of a list of wchar

jralls:
Aha! The smoking gun on string ports.

gnucash/report/standard-reports/account-piecharts.scm
```diff
@@ -570,8 +570,8 @@ balance at a given time"))
                                     (qof-print-date to-date)))
                        (if show-total?
                            (let ((total (apply + (unzip1 combined))))
-                             (format
-                              #f ": ~a"
+                             (string-append
+                              ": "
```
fellen: In both cases I am not sure, how the result will look in LTR languages like ar or he. Can you test it with LANG=he_IL.utf8 or an ar variant?

christopherlam:
> Aha! The smoking gun on string ports.

Which part?

jralls:
"One warning though: in Guile 2.0, string ports only support characters representable in the %default-port-encoding, which defaults to the encoding of the current locale."

Works fine on Unix, blows up on Windows in the presence of code pages.

christopherlam:
> "One warning though: in Guile 2.0, string ports only support characters representable in the %default-port-encoding, which defaults to the encoding of the current locale."
> 
> Works fine on Unix, blows up on Windows in the presence of code pages.

Oh ok. I am sure you know Mark Weaver is now a core guile dev. I kept a link to this particular email handy because I was thinking of replacing the ugly gnc:substring-replace-from-to in utilities.scm with a better one; didn't realize it had a nugget of wisdom regarding strings in guile-2.0. Perhaps I'll summarize my experiments on the current state of string functions in bugzilla?

jralls:
OK, if it's on one of the existing  transcoding bugs.

jralls:
Looks like this is just whitespace now. Why don't you just merge it?

PR #488, opened 2019-04-15
Bug 797195 - computeInterestIncrement function in fin.scm only works for annual compounding

thetedmunds:
https://bugs.gnucash.org/show_bug.cgi?id=797195
Changed gnc:computeInterestIncrement to use a new futureValue calculation that takes time as a number of periods.

### DIFF
--- a/libgnucash/app-utils/fin.scm
+++ b/libgnucash/app-utils/fin.scm
@@ -63,12 +63,9 @@
     ;; formula from http://www.riskglossary.com/articles/compounding.htm
   (* a (expt (+ 1 (/ r n)) (* n t))))
 
-(define (gnc:computeInterestIncrement amount interest periods i)
-  (let ((thisVal (gnc:futureValue amount interest periods i))
-        (prevVal (gnc:futureValue amount interest periods (- i 1))))
-    (- thisVal prevVal)
-  )
-)
+(define (gnc:computeInterestIncrement pv ann-rate compounds period)
+  (let ((rate (/ ann-rate compounds)))  
+    (* rate (* pv (expt (+ 1 rate) (- period 1))))))
 
 ;;;;;
 ;; below: not-exposed/"private" functions, used by the "public" functions


jralls:
```computeInterestIncrement``` is the only place ```gnc:futureValue``` was used in GnuCash. The new ```gnc:futureValuePeriods``` is more correct, so ISTM we should deprecate ```gnc:futureValue``` and plan to remove it in GnuCash 4.0.

**Comment by thetedmunds** [comment]
Since it's exposed with the `gnc` prefix, I assume that there might be some people who are using it as part of various work-arounds in the STX editor.  That's the only reason I didn't just modify the existing function.  But the same could be said for people using `computeInterestIncrement`.  So deprecation seems reasonable.

**Comment by jralls** [comment]
After a little more thought I don't like either of them. An interest function should either take fv, pv, 
periodic rate, periods or it should take fv, pv , annual rate, periods-per-year, years, solving for one of the  values. IMO the first is more straightforward. That's even more true for ```gnccomputeInterestIncrement```: As you point out it doesn't make a lot of sense to compute a whole year's interest, so the rate should be the period rate, not the annual rate. You said in the bug that you can't pass it ```.0375/12```, but what about ``` (/ .0375 12)```? 

But since the function signature is already set, what about
```
(define (gnc:computeInterestIncrement pv ann-rate compounds period)
   (let ((exponent (+ 1 (/ ann-rate compounds))))
      (* pv (- (expt exponent period) (expt exponent (- period 1))))))
```
instead of generating a new function?

thetedmunds:
Well, I'm not super invested in the functionality (because I want a function that allows you to specify a compounding frequency that is different from the transaction frequency).  But I would say that it is more intuitive to specify the interest rate as an annual rate (because that's generally how people talk about interest rates, and the periodic rate is rarely a nice round number) and the period number (because the scheduled transaction editor provides the special variable `i` that corresponds to the period).  That way people can enter the values they have (annual rate and transaction number) instead of having to enter scheme code in the transaction editor to either convert the rate or convert the "time".

But, as you say, since the function signature is already set, we can eliminate the double call to `gnc:futureValue`, and do it all "in-function".  Only one `expt` is needed:

    (define (gnc:computeInterestIncrement pv ann-rate compounds period)
       (let ((rate (/ ann-rate compounds)))  
          (* pv (* (expt (+ 1 rate) (- period 1)) rate))))

In the end, the relative efficiency of * vs expt is probably not worth worrying about, this implementation still doesn't do any intermediate rounding (to deal with the fact that most payers of interest don't pay fractions of a cent), and - given that it seems to have been calculating an unexpected result since 2005 without complaint - it seems likely that it is not heavily used.  So any of the proposed solutions seem fine :)

orcainbox:
Hey guys I am a CPA.

I agree with the following comment below.  K|$$ principle



"But I would say that it is more intuitive to specify the interest rate as an annual rate (because that's generally how people talk about interest rates, and the periodic rate is rarely a nice round number) and the period number "

________________________________

jralls:
@orcainbox, KISS takes you to using the period rate. Consider that the standard [spreadsheet FV function](https://help.libreoffice.org/Calc/Financial_Functions_Part_Three#Syntax_14) takes the rate per-period, not per-year.  Users have somehow managed with that for 40 years.

jralls:
@thetedmunds As long as mantissas stay small enough to work in the CPU's floating point unit there's no difference in execution speed, but even without that it's not in a loop so maintainability in the sense of being able to reason about the code is more important than execution speed. On one hand <tt>pr(1+r)<sup>t-1</sup></tt> is the fully-solved formula that one will likely find in a text, on the other from a coding standpoint it's more clear what's going on with <tt>p((1+r)<sup>t</sup> - (1+r)<sup>t-1</sup>)</tt>. I guess it doesn't really matter, whichever way you want to code it is fine.

thetedmunds:
I've pushed an update.  I went with an ordering that seems intuitive (calculate the balance at the end of the last period, then multiply by r to find the amount of interest in one period).

And, not to knock Excel's TVM functions, but you know the old saying: show me a person using the FV function in one cell, and I'll show you a person with four other cells containing the nominal rate, the compounding frequency, the payment frequency, and a periodic rate calculation using EFFECT, NOMINAL, and division.  At least, I think it's an old saying...

jralls:
@thetedmunds multiplication being associative, the order of the multiplicands doesn't matter. This is fine.

Spreadsheets will do the .05/12 inline. LibreOffice will even do 5%/12, which is pretty friendly. ```gnc:futureValue``` doing periods and years is a somewhat misguided attempt to work around the fact that the SX parser won't even do .05/12. Using both NORMAL and EFFECT is weird, they're mutually reciprocal... and you probably don't want to pass the result of either of them to FV.

jralls:
Thanks!

PR #489, opened 2019-04-15
Pr pycons2to3 - fix python console in python3

c-holtermann:
Make pycons great again. Turning on the ipython shell does not yet work as the ipython api from 5 on has changed a lot.


jralls:
Fixes the crash on startup after building WITH_PYTHON=ON, so I'm happy even if it isn't quite finished.
Thanks!

PR #492, opened 2019-04-19
make python shell startup more understandable

c-holtermann:


gnucash/python/init.py
```diff
@@ -85,8 +85,13 @@ def refresh (self):
 
 # Change this to "if True:" to switch on a python console at gnucash
 # startup:
+# shelltype can either be "python" or "ipython" (the latter is not yet fully functional)
 if False:
-    console = Console(argv = [], shelltype = 'python', banner = [['woop', 'title']], size = 100)
+    shelltype = "ipython"
```
jralls: If ```ipython``` doesn't quite work shouldn't ```python``` be the default, particularly since CPython is far more widely used in the non-Windows world?

c-holtermann: Yes. My mistake. I took the patch from my development environment trying to fix ipython. I will change that. Thanks !

jralls:
Would you mind squashing that to a single commit and force-pushing?

c-holtermann:
force pushed, renamed branch to shorter name as requested by John Ralls. Started a new pull request. https://github.com/Gnucash/gnucash/pull/501

PR #494, opened 2019-04-22
test-utilities

christopherlam:
test suite for utilities.scm.

note the test-utilities.scm *should* be in libgnucash/scm/test/test-utilities.scm but I don't know how to create CMakeLists.txt -- if anyone can offer some guidance, I'll move this file.

also fixes comment in `gnc:substring-replace-from-to` which was incorrect.
also refactor substring substitution algorithms to use a much more efficient one by Mark Weaver.


christopherlam:
Do you guys mind if test-utilities is in libgnucash/engine/test/test-utilities.scm rather than libgnucash/scm/test/test-utilities.scm? It's too annoying to keep hammering at CMakeLists until it works... :)

jralls:
I'd really rather have the tests in the same place as the code its testing, otherwise the dependency graph gets messy. Put your test file in ```libgnucash/scm/tests``` and I'll write a CMakeLists.txt for you tomorrow.

christopherlam:
Cool, will do tonight.

From my understanding of guile internals (I may be wrong), if a large string was sent for substitutions, and there are numerous substitutions, the various `substring` calls would not copy the strings at all, but would simply select the pointer & length to the appropriate locations from the original string to build a list of substrings including the replacement substring, which is then collected in 1 go via `string-concatenate-reverse` into the final result.

christopherlam:
My attempt at CMakeLists.txt - will successfully run `test-utilities`, no idea about `set_dist_list` stuff.

gjanssens:
set_dist_list is required to add all source files to our distribution tarball. It should at least contain the CMakeLists.txt and your test file. If not, consumers of our distribution tar ball won't be able to run your newly added tests simply because the files to do so are missing.

jralls:
So why then did you change the set-dist from what I wrote for you yesterday?

christopherlam:
@jralls I think your commit was https://github.com/Gnucash/gnucash/commit/09eb58d945277f331ebde43905d53f6b04244160 - I saw `scm/CMakeLists.txt` but not `scm/test/CMakeLists.txt`, and this caused travis failure https://travis-ci.org/christopherlam/gnucash/jobs/523734983 -- anyway I made set_dist_list - is this the right format?

jralls:
```libgnucash/scm/test/test-utilities.scm``` seems to have disappeared... When it comes back, please rename the test ```test-scm-utilities```.  ```test-utilities``` is too generic, particularly in a list of > 120 tests. We have lots of ```utilities```.

In fact, since the number of tests is getting large maybe we need a naming convention to make it easier to identify tests from the summary report. Two obvious options come to mind: The directory path to the test (so this one would become ```libgnucash-scm-utilities```) or the logging domain (which this directory doesn't have, but for engine tests would be ```gnc.engine.whatever```).

christopherlam:
`test-utilities` has disappeared because it's now committed as requested https://github.com/Gnucash/gnucash/pull/494#issuecomment-485627138; I'll push a rename to `libgnucash-scm-utilites.scm` commit.

gjanssens:
As the log domain assignment is incomplete, I think the directory path based approach is more useful. However does this mean the test *files* have to be named like that or just the test references used in the file ?

christopherlam:
It's nice to have a consistent test- prefix, so, `test-libgnucash-scm-utilities.scm` would be better I think

christopherlam:
test then refactoring commits merged individually.

jralls:
> However does this mean the test files have to be named like that or just the test references used in the file ?

As far as I care just the names of the tests that show up on the result summary and in Testing/Temporary/LastTest.log. It would be *really* redundant to include the path in test file names, and IMO it's clearest if, at least for unit tests, are named ```test-the-file-being-tested```. Inside the file each unit test should be named ```test-function-name```. If the function has more than one execution path and one is following the generally accepted unit test best practice then there should be a separate, appropriately named, test function exercising each path. I'm guilty of not doing this and instead testing all the paths (or all the paths that I can figure out or if I'm being lazy the one path that I'm worried about that motivated me to write the test) in a single function.

gjanssens:
Right, that makes more sense.

PR #496, opened 2019-04-25
python console - fix deprecated positional arguments with the GObject constructor 

c-holtermann:
fix Bug 797205
https://bugs.gnucash.org/show_bug.cgi?id=797205


jralls:
Thanks.
Could you use shorter branch names so that "Merged Chrisoph Holtermann's 'mybranch' into maint." is less than 80 characters?

c-holtermann:
I'll have a look at the length of the branch names

PR #497, opened 2019-04-26
ImportMatcherMultiSelect

DaveC49:
Mods to import-main-matcher.c and dialog-import.glade to implement multi-selection of transactions to import and assignment of a single transfer account to the selection.

John these should now be the correct files David

### DIFF
--- a/gnucash/gtkbuilder/dialog-import.glade
+++ b/gnucash/gtkbuilder/dialog-import.glade
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<!-- Generated with glade 3.20.4 -->
+<!-- Generated with glade 3.22.1 -->
 <interface>
   <requires lib="gtk+" version="3.10"/>
   <object class="GtkImage" id="account_new_icon">
@@ -14,6 +14,9 @@
     <property name="default_height">600</property>
     <property name="type_hint">dialog</property>
     <signal name="map" handler="gnc_ui_generic_account_picker_map_cb" swapped="no"/>
+    <child>
+      <placeholder/>
+    </child>
     <child internal-child="vbox">
       <object class="GtkBox" id="account_picker_vbox">
         <property name="visible">True</property>
@@ -165,7 +168,6 @@
     <property name="page_increment">10</property>
   </object>
   <object class="GtkDialog" id="format_picker_dialog">
-    <property name="visible">False</property>
     <property name="can_focus">False</property>
     <property name="title" translatable="yes">Choose a format</property>
     <property name="resizable">False</property>
@@ -173,6 +175,9 @@
     <property name="default_width">600</property>
     <property name="default_height">400</property>
     <property name="type_hint">dialog</property>
+    <child>
+      <placeholder/>
+    </child>
     <child internal-child="vbox">
       <object class="GtkBox" id="dialog-vbox16">
         <property name="visible">True</property>
@@ -301,6 +306,9 @@
     <property name="visible">True</property>
     <property name="can_focus">False</property>
     <property name="title" translatable="yes">Preferences</property>
+    <child>
+      <placeholder/>
+    </child>
     <child>
       <object class="GtkGrid" id="matcher_prefs">
         <property name="visible">True</property>
@@ -554,6 +562,9 @@
     <property name="default_width">600</property>
     <property name="default_height">400</property>
     <property name="type_hint">dialog</property>
+    <child>
+      <placeholder/>
+    </child>
     <child internal-child="vbox">
       <object class="GtkBox" id="dialog-vbox14">
         <property name="visible">True</property>
@@ -726,6 +737,9 @@
     <property name="resizable">False</property>
     <property name="type_hint">dialog</property>
     <signal name="close" handler="on_matcher_help_close_clicked" swapped="no"/>
+    <child>
+      <placeholder/>
+    </child>
     <child internal-child="vbox">
       <object class="GtkBox" id="dialog-vbox18">
         <property name="visible">True</property>
@@ -1107,8 +1121,11 @@
             <property name="reorderable">True</property>
             <property name="rules_hint">True</property>
             <property name="enable_search">False</property>
+            <property name="rubber_banding">True</property>
             <child internal-child="selection">
-              <object class="GtkTreeSelection" id="treeview-selection3"/>
+              <object class="GtkTreeSelection" id="treeview-selection3">
+                <property name="mode">multiple</property>
+              </object>
             </child>
           </object>
         </child>
@@ -1145,6 +1162,9 @@
     <property name="default_height">400</property>
     <property name="type_hint">dialog</property>
     <signal name="close" handler="on_matcher_cancel_clicked" swapped="no"/>
+    <child>
+      <placeholder/>
+    </child>
     <child internal-child="vbox">
       <object class="GtkBox" id="transaction_matcher_vbox">
         <property name="visible">True</property>
--- a/gnucash/import-export/import-main-matcher.c
+++ b/gnucash/import-export/import-main-matcher.c
@@ -64,16 +64,13 @@ struct _main_matcher_info
     GNCTransactionProcessedCB transaction_processed_cb;
     gpointer user_data;
     GNCImportPendingMatches *pending_matches;
-    GtkTreeViewColumn *account_column;
 };
 
 enum downloaded_cols
 {
-    DOWNLOADED_COL_DATE_TXT = 0,
-    DOWNLOADED_COL_DATE_INT64,
+    DOWNLOADED_COL_DATE = 0,
     DOWNLOADED_COL_ACCOUNT,
     DOWNLOADED_COL_AMOUNT,
-    DOWNLOADED_COL_AMOUNT_DOUBLE,
     DOWNLOADED_COL_DESCRIPTION,
     DOWNLOADED_COL_MEMO,
     DOWNLOADED_COL_ACTION_ADD,
@@ -98,9 +95,28 @@ void on_matcher_help_clicked (GtkButton *button, gpointer user_data);
 void on_matcher_help_close_clicked (GtkButton *button, gpointer user_data);
 
 /* Local prototypes */
-static void
-refresh_model_row(GNCImportMainMatcher *gui, GtkTreeModel *model,
+static void gnc_gen_trans_assign_transfer_account(
+                                                GtkTreeView *treeview, 
+                                                gboolean *first,
+                                                gboolean *is_selection,
+                                                GtkTreePath *path,
+                                                Account **new_acc,
+                                                GNCImportMainMatcher *info);
+static void gnc_gen_trans_assign_transfer_account_to_selection_cb (
+                                          GtkMenuItem *menuitem,
+                                          GNCImportMainMatcher *info);
+static void gnc_gen_trans_view_popup_menu (GtkTreeView *treeview, 
+                                          GdkEvent *event,
+                                          GNCImportMainMatcher *info);
+static gboolean gnc_gen_trans_onButtonPressed_cb (GtkTreeView *treeview,
+                                                  GdkEvent *event,
+                                                  GNCImportMainMatcher *info);
+static gboolean gnc_gen_trans_onPopupMenu_cb (GtkTreeView *treeview,
+                                              GdkEvent *event, 
+                                              GNCImportMainMatcher *info);
+static void refresh_model_row(GNCImportMainMatcher *gui, GtkTreeModel *model,
                   GtkTreeIter *iter, GNCImportTransInfo *info);
+/* end local prototypes */
 
 void gnc_gen_trans_list_delete (GNCImportMainMatcher *info)
 {
@@ -361,53 +377,243 @@ gnc_gen_trans_update_toggled_cb (GtkCellRendererToggle *cell_renderer,
 }
 
 static void
-gnc_gen_trans_row_activated_cb (GtkTreeView           *view,
-                                GtkTreePath           *path,
-                                GtkTreeViewColumn     *column,
-                                GNCImportMainMatcher  *gui)
+gnc_gen_trans_assign_transfer_account(GtkTreeView *treeview, 
+                                      gboolean *first,
+                                      gboolean *is_selection,
+                                      GtkTreePath *path,
+                                      Account **new_acc,
+                                      GNCImportMainMatcher *info)
 {
     GtkTreeModel *model;
     GtkTreeIter iter;
     GNCImportTransInfo *trans_info;
+    Account *old_acc;
+    gboolean ok_pressed;
 
-    model = gtk_tree_view_get_model(gui->view);
-    if (!gtk_tree_model_get_iter(model, &iter, path))
-        return;
-    gtk_tree_model_get(model, &iter, DOWNLOADED_COL_DATA, &trans_info, -1);
-
-    switch (gnc_import_TransInfo_get_action (trans_info))
+    ENTER("");
+    DEBUG("first = %s",*first?"true":"false");
+    DEBUG("is_selection = %s",*is_selection?"true":"false");
+    DEBUG("path  = %s", gtk_tree_path_to_string(path));
+    DEBUG("account passed in = %s", gnc_get_account_name_for_register(*new_acc));
+    model = gtk_tree_view_get_model(treeview);
+    if (gtk_tree_model_get_iter(model, &iter, path))
     {
-    case GNCImport_ADD:
-        if (gnc_import_TransInfo_is_balanced(trans_info) == FALSE)
+        gtk_tree_model_get(model, &iter, DOWNLOADED_COL_DATA, &trans_info, -1);
+
+        switch (gnc_import_TransInfo_get_action (trans_info))
         {
-            run_account_picker_dialog (gui, model, &iter, trans_info);
+        case GNCImport_ADD:
+            if (gnc_import_TransInfo_is_balanced(trans_info) == FALSE)
+            {
+                ok_pressed = TRUE;
+                old_acc  = gnc_import_TransInfo_get_destacc (trans_info);
+                if (*first)
+                {
+                    ok_pressed = FALSE;
+                    *new_acc = gnc_import_select_account(info->main_widget,
+                        NULL,
+                        TRUE,
+                        _("Destination account for the auto-balance split."),
+                        xaccTransGetCurrency(
+                              gnc_import_TransInfo_get_trans(trans_info)),
+                        ACCT_TYPE_NONE,
+                        old_acc,
+                        &ok_pressed);
+                    *first = FALSE;
+                    DEBUG("account selected = %s", 
+                            gnc_account_get_full_name(*new_acc));                    
+                }
+                if (ok_pressed)
+                        gnc_import_TransInfo_set_destacc (trans_info, 
+                                                    *new_acc, TRUE);
+            }
+            break;
+        case GNCImport_CLEAR:
+        case GNCImport_UPDATE:
+            if (first && !is_selection) run_match_dialog (info, trans_info);
+            break;
+        case GNCImport_SKIP:        
+            break;
+        default:
+            PERR("InvalidGNCImportValue");
+            break;
         }
-        break;
-    case GNCImport_CLEAR:
-    case GNCImport_UPDATE:
-        run_match_dialog (gui, trans_info);
-        break;
-    case GNCImport_SKIP:
-        /*The information displayed is only informative, until you select an action*/
-        break;
-    default:
-        PERR("I don't know what to do! (Yet...)");
-        break;
+        refresh_model_row(info, model, &iter, trans_info);
     }
-    refresh_model_row(gui, model, &iter, trans_info);
+    LEAVE("");
+}
+
+static void
+gnc_gen_trans_assign_transfer_account_to_selection_cb (GtkMenuItem *menuitem,
+                                             GNCImportMainMatcher *info)
+{
+    GtkTreeView *treeview;
+    GtkTreeSelection *selection;
+    GtkTreeModel *model;
+    GtkTreeIter iter;
+    GNCImportTransInfo *trans_info;
+    Account *assigned_account;
+    GList *selected_rows, *l;
+    gboolean first, is_selection;
+
+    ENTER("");
+    treeview = GTK_TREE_VIEW(info->view);
+    model = gtk_tree_view_get_model(treeview);
+    selection = gtk_tree_view_get_selection(treeview);
+    selected_rows = gtk_tree_selection_get_selected_rows (selection, &model);
+    assigned_account = NULL;
+    first = TRUE;
+    is_selection = TRUE;    
+    DEBUG("Rows in selection = %i",
+          gtk_tree_selection_count_selected_rows(selection));
+    DEBUG("Entering loop over selection");
+
+    if (gtk_tree_selection_count_selected_rows(selection) > 0)
+    {
+        for (l = selected_rows; l != NULL; l = l->next)
+        {
+            DEBUG("passing first = %s",
+                                                first?"true":"false");
+            DEBUG("passing is_selection = %s",
+                                                is_selection?"true":"false");
+            DEBUG("passing path = %s",
+                           gtk_tree_path_to_string(l->data));
+            DEBUG("passing account value = %s",
+                        gnc_account_get_full_name(assigned_account));
+            gnc_gen_trans_assign_transfer_account(treeview,
+                           &first, &is_selection, l->data, &assigned_account, info);
+            DEBUG("returned value of account = %s", 
+                        gnc_account_get_full_name(assigned_account));
+            DEBUG("returned value of first = %s",first?"true":"false");
+            if (assigned_account == NULL) break;
+            gtk_tree_selection_unselect_path(selection, l->data);
+        }
+    }
+    g_list_free_full (selected_rows, (GDestroyNotify) gtk_tree_path_free);
+    LEAVE("");
+}
+
+static void
+gnc_gen_trans_row_activated_cb (GtkTreeView *treeview,
+                                GtkTreePath *path,
+                                GtkTreeViewColumn *column,
+                                GNCImportMainMatcher *info)
+{
+    Account *assigned_account;
+    gboolean first, is_selection;
+
+    ENTER("");
+    assigned_account = NULL;
+    first = TRUE;
+    is_selection = FALSE;
+    gnc_gen_trans_assign_transfer_account(treeview, 
+                            &first,  &is_selection, path,
+                            &assigned_account, info);
+    DEBUG("account returned = %s", gnc_account_get_full_name(assigned_account));
+    LEAVE("");
 }
 
 static void
 gnc_gen_trans_row_changed_cb (GtkTreeSelection *selection,
-                              GNCImportMainMatcher *gui)
+                              GNCImportMainMatcher *info)
 {
     GtkTreeModel *model;
     GtkTreeIter iter;
+    GtkSelectionMode mode;
 
-    if (!gtk_tree_selection_get_selected(selection, &model, &iter))
-        return;
-    gtk_tree_selection_unselect_iter(selection, &iter);
+    ENTER("");
+    mode = gtk_tree_selection_get_mode(selection);
+    switch (mode)
+    {
+        case GTK_SELECTION_MULTIPLE:
+            DEBUG("mode = GTK_SELECTION_MULTIPLE, no action");
+            break;
+        case GTK_SELECTION_NONE:
+            DEBUG("mode = GTK_SELECTION_NONE, no action");
+            break;
+        case GTK_SELECTION_BROWSE:
+            DEBUG("mode = GTK_SELECTION_BROWSE->default");
+        case GTK_SELECTION_SINGLE:
+            DEBUG("mode = GTK_SELECTION_SINGLE->default");
+        default:
+            DEBUG("mode = default unselect selected row");
+            if (gtk_tree_selection_get_selected(selection, &model, &iter))
+            {
+                gtk_tree_selection_unselect_iter(selection, &iter);
+            }
+    }
+    LEAVE("");
 }
+
+static void
+gnc_gen_trans_view_popup_menu (GtkTreeView *treeview, 
+                               GdkEvent *event,
+                               GNCImportMainMatcher *info)
+{    
+    GtkWidget *menu, *menuitem;
+    GdkEventButton *event_button;
+
+    ENTER ("");
+    menu = gtk_menu_new();
+    menuitem = gtk_menu_item_new_with_label(_("Assign a transfer account."));
+    g_signal_connect(menuitem, "activate",
+                     G_CALLBACK(
+                     gnc_gen_trans_assign_transfer_account_to_selection_cb),
+                     info);
+    DEBUG("Callback to assign destination account to selection connected");
+    gtk_menu_shell_append(GTK_MENU_SHELL(menu), menuitem);
+    gtk_widget_show_all(menu);
+    event_button = (GdkEventButton *) event;
+    /* Note: event can be NULL here when called from view_onPopupMenu;
+     *  gdk_event_get_time() accepts a NULL argument */
+    /* replace deprecated function for GTK versions >=3.22 */
+#if GTK_CHECK_VERSION(3,22,0)
+   gtk_menu_popup_at_pointer(GTK_MENU(menu),(GdkEvent*)event);
+#else
+    gtk_menu_popup(GTK_MENU(menu), NULL, NULL, NULL, NULL, 
+                    (event_button != NULL) ? event_button->button : 0, 
+                    gdk_event_get_time((GdkEvent*)event)); 
+#endif  
+    LEAVE ("");
+}
+
+static gboolean
+gnc_gen_trans_onButtonPressed_cb (GtkTreeView *treeview,
+                                  GdkEvent *event,
+                                  GNCImportMainMatcher *info)
+{
+    GdkEventButton *event_button;
+    ENTER("");
+    g_return_val_if_fail (treeview != NULL, FALSE);
+    g_return_val_if_fail (event != NULL, FALSE);
+    /* handle single click with the right mouse button? */
+    if (event->type == GDK_BUTTON_PRESS)
+    {
+        event_button = (GdkEventButton *) event;
+        if (event_button->button == GDK_BUTTON_SECONDARY)
+        {
+            DEBUG("Right mouseClick detected- popup the menu.");
+            gnc_gen_trans_view_popup_menu(treeview, event, info);
+            LEAVE("return TRUE");
+            return TRUE;
+        }
+    }
+    LEAVE("return FALSE");
+    return FALSE;
+}
+
+static gboolean
+gnc_gen_trans_onPopupMenu_cb (GtkTreeView *treeview,
+                              GdkEvent *event, 
+                              GNCImportMainMatcher *info)
+{
+    ENTER("");
+    /* respond to Shift-F10 popup menu hotkey */
+    gnc_gen_trans_view_popup_menu(treeview, NULL, info);
+    LEAVE ("");
+    return TRUE;
+}
+
 
 static GtkTreeViewColumn *
 add_text_column(GtkTreeView *view, const gchar *title, int col_num)
@@ -421,19 +627,7 @@ add_text_column(GtkTreeView *view, const gchar *title, int col_num)
               "text", col_num,
               "background", DOWNLOADED_COL_COLOR,
               NULL);
-
-    // If date column, use the time64 value for the sorting.
-    if (col_num == DOWNLOADED_COL_DATE_TXT)
-        gtk_tree_view_column_set_sort_column_id(column, DOWNLOADED_COL_DATE_INT64);
-    else if (col_num == DOWNLOADED_COL_AMOUNT) // If amount column, use double value
-    {
-        gtk_cell_renderer_set_alignment (renderer, 1.0, 0.5); // right align amount column
-        gtk_cell_renderer_set_padding (renderer, 5, 0); // add padding so its not close to description
-        gtk_tree_view_column_set_sort_column_id(column, DOWNLOADED_COL_AMOUNT_DOUBLE);
-    }
-    else
-        gtk_tree_view_column_set_sort_column_id(column, col_num);
-
+    gtk_tree_view_column_set_sort_column_id(column, col_num);
     g_object_set(G_OBJECT(column),
                  "reorderable", TRUE,
                  "resizable", TRUE,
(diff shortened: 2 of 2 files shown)


gnucash/gtkbuilder/dialog-import.glade
```diff
@@ -1122,21 +1142,6 @@
     <child>
       <placeholder/>
     </child>
-    <child>
```
Bob-IT: Why are you removing the Show source account button?

gnucash/import-export/import-main-matcher.c
```diff
@@ -64,16 +64,13 @@ struct _main_matcher_info
     GNCTransactionProcessedCB transaction_processed_cb;
     gpointer user_data;
     GNCImportPendingMatches *pending_matches;
-    GtkTreeViewColumn *account_column;
 };
 
 enum downloaded_cols
 {
-    DOWNLOADED_COL_DATE_TXT = 0,
```
Bob-IT: You are also removing the date sort and amount sort model columns?

gnucash/import-export/import-main-matcher.c
```diff
@@ -421,19 +623,7 @@ add_text_column(GtkTreeView *view, const gchar *title, int col_num)
               "text", col_num,
               "background", DOWNLOADED_COL_COLOR,
               NULL);
-
```
Bob-IT: And this lot which is to do with the sorting

gnucash/import-export/import-main-matcher.c
```diff
@@ -476,23 +666,20 @@ gnc_gen_trans_init_view (GNCImportMainMatcher *info,
     GtkTreeSelection *selection;
 
     view = info->view;
-    store = gtk_list_store_new(NUM_DOWNLOADED_COLS, G_TYPE_STRING, G_TYPE_INT64,
```
Bob-IT: And here...

gnucash/import-export/import-main-matcher.c
```diff
@@ -361,54 +377,240 @@ gnc_gen_trans_update_toggled_cb (GtkCellRendererToggle *cell_renderer,
...
+    gtk_menu_shell_append(GTK_MENU_SHELL(menu), menuitem);
+    gtk_widget_show_all(menu);
+    event_button = (GdkEventButton *) event;
+    /* Note: event can be NULL here when called from view_onPopupMenu;
+     *  gdk_event_get_time() accepts a NULL argument */
+    /* replace deprecated function
+    gtk_menu_popup(GTK_MENU(menu), NULL, NULL, NULL, NULL, 
+                    (event_button != NULL) ? event_button->button : 0, 
+                    gdk_event_get_time((GdkEvent*)event)); */
+    gtk_menu_popup_at_pointer(GTK_MENU(menu),(GdkEvent*)event);
```
jralls: CI is failing on Ubuntu 14.04 because this function didn't appear until gtk 3.22. You need to leave the call to gtk_menu_popup and guard it like this:
```
#if ! GTK_MAJOR_VERSION < 3 || GTK_MAJOR_VERSION == 3 && GTK_MINOR_VERSION < 22
   gtk_menu_popup(GTK_MENU(menu), NULL, NULL, NULL, NULL, 
                    (event_button != NULL) ? event_button->button : 0, 
                    gdk_event_get_time((GdkEvent*)event));
#else
    gtk_menu_popup_at_pointer(GTK_MENU(menu),(GdkEvent*)event);            
#endif
 ```   
This should be a separate commit called something like "Silence deprecation warning". As a general rule don't comment out old code, just replace it. Old code in comments is just noise.

File: gnucash/gtkbuilder/dialog-import.glade (hunk 1 above)
Comment (DaveC49) [reply gnucash/gtkbuilder/dialog-import.glade]: Bob,
I wasn't aware any of that had changed. This was a recommit of some changes from August last year that got lost, most likely as a result of my inexperience with git and github. I copied the changed file into the commit rather than reapplying the required changes so it is likely I overwrote some changes since my original commit.  I will reapply the required changes to a current copy of the file and resubmit it. I have to make a change John has suggested as well so I will incorporate that in the next commit as well.

David

File: gnucash/import-export/import-main-matcher.c (hunk 5 above)
Comment (DaveC49) [reply gnucash/import-export/import-main-matcher.c]: Thanks John 
I'll add that and fix up the problems Bob identified above and recommit it 

David

File: gnucash/gtkbuilder/dialog-import.glade (hunk 1 above)
Comment (DaveC49) [reply gnucash/gtkbuilder/dialog-import.glade]: Bob,

That was indeed what had happened. I only had to change the treeview and treeview selection settings to enable multiple selection so I have copied the current dialog-import.glade from the maint branch into my feature branch and then modified it.  Will test it, commit and then do another pull request.

David

**Comment by DaveC49** [comment]
John, Bob 
Reapplied changes to enable multiple selection in the treeview to the current version of dialog-import.glade from the maint branch and made replacement of deprecated gtk_popup_menu conditional on GTK version being >=3.22.

David Cousens

File: gnucash/gtkbuilder/dialog-import.glade (hunk 1 above)
Comment (Bob-IT) [reply gnucash/gtkbuilder/dialog-import.glade]: Dave,
I still think it does not look right, had a look at your branch and there are now two commits, the first one with all the removal and a second one with John's change request with the glade changes but no sign of all the required changes in the 'c' source file. You can probably fix this with some git commands but it may be simpler to update your git maint branch so it is current, create a new branch for your changes and then make your changes there.
Bob

**Comment by jralls** [comment]
Dave, Please squash the two commits so that we don't have the broken tests in history.

**Comment by DaveC49** [comment]
Again my apologies John. 
Unfortunately I reforked the current state of GnuCash into my local repository on my computer and then pushed it my
Github repository yesterday when I decided to start from scratch on the documentation changes I was making yesterday.  I
have overwritten both and no longer have the record of those commits - my bad.  I will have to take the time out to try
and fully understand git and github and how to keep everything in sync and reverse things properly so I don't get into
the same situation again. I'll setup a couple of dummy repositories on my github and play until I understand what is
actually happening. Is there any way for you to squash them from your end?

David Cousens

**Comment by jralls** [comment]
@DaveC49 Nah, it takes way more effort than that to lose work once you've put it on github. It's still here. Dunno what your local repo looks like, but suppose that [this](https://github.com/DaveC49/gnucash) is the remote named ```origin```. Then
```
git fetch origin
git checkout -b ImportMultiSelect -t origin/ImportMultiSelect
```
will restore it to your local repo.

**Comment by DaveC49** [comment]
OK John I think I have achieved that. I have pushed the now combined commit to my repo.  Can you get them from there
using the previous pull request.

If I have it right these two commits from the ImportMultiSelect branch

* f27e8f114 (HEAD -> ImportMultiSelect, origin/ImportMultiSelect) Silence Deprecation Warning
* 7d7bf3a72 ImportMatcherMultiSelect

should be combined in the following commit

* 74f61597a (HEAD -> ImportMultiSelect) ImportMatcherMultiSelect

I used the following command sequence: 
$git fetch origin       # origin was set to https://github.com/DaveC49/gnucash.git
$git checkout ImportMultiSelect
$git status
On branch ImportMultiSelect
nothing to commit, working tree clean
$ git log --graph --decorate --pretty=oneline --abbrev-commit
* f27e8f114 (HEAD -> ImportMultiSelect, origin/ImportMultiSelect) Silence Deprecation Warning
* 7d7bf3a72 ImportMatcherMultiSelect
.
.
.
$git rebase -i HEAD~2
  pick 7d7bf3a72 ImportMatcherMultiSelect
  squash f27e8f114 Silence Deprecation Warning
..
$ git log --graph --decorate --pretty=oneline --abbrev-commit
* 74f61597a (HEAD -> ImportMultiSelect) ImportMatcherMultiSelect

$ git push -f --set-upstream origin ImportMultiSelect

**Comment by jralls** [comment]
David,

Yup, perfect. @Bob-IT Are you happy with it? @gjanssens do you agree this is a small enough UI change for maint?

File: gnucash/import-export/import-main-matcher.c [hunk 6]
```diff
@@ -361,54 +377,243 @@ gnc_gen_trans_update_toggled_cb (GtkCellRendererToggle *cell_renderer,
...
+                     gnc_gen_trans_assign_transfer_account_to_selection_cb),
+                     info);
+    DEBUG("Callback to assign destination account to selection connected");
+    gtk_menu_shell_append(GTK_MENU_SHELL(menu), menuitem);
+    gtk_widget_show_all(menu);
+    event_button = (GdkEventButton *) event;
+    /* Note: event can be NULL here when called from view_onPopupMenu;
+     *  gdk_event_get_time() accepts a NULL argument */
+    /* replace deprecated function for GTK versions >=3.22 */
+#if ! GTK_MAJOR_VERSION < 3 || GTK_MAJOR_VERSION == 3 && GTK_MINOR_VERSION < 22
```
gjanssens: There is an easier to understand macro for this:

    #if GTK_CHECK_VERSION(3,22,0)
        // use the Gtk3.22 function
    #else
        // use the deprecated function
    #endif

gjanssens:
More importantly, when applying this PR to maint, I can consistently crash gnucash after applying this PR. Here's the backtrace:
```
Thread 1 "gnucash" received signal SIGSEGV, Segmentation fault.
0x00007ffff7b04552 in g_slice_alloc () from /lib64/libglib-2.0.so.0
(gdb) bt
#0  0x00007ffff7b04552 in g_slice_alloc () at /lib64/libglib-2.0.so.0
#1  0x00007ffff7ae19ea in g_list_prepend () at /lib64/libglib-2.0.so.0
#2  0x00007ffff72964dd in  () at /lib64/libgdk-3.so.0
#3  0x00007ffff7296ada in  () at /lib64/libgdk-3.so.0
#4  0x00007ffff7274bac in  () at /lib64/libgdk-3.so.0
#5  0x00007ffff7ae297b in  () at /lib64/libglib-2.0.so.0
#6  0x00007ffff7ae606d in g_main_context_dispatch () at /lib64/libglib-2.0.so.0
#7  0x00007ffff7ae6438 in  () at /lib64/libglib-2.0.so.0
#8  0x00007ffff7ae6762 in g_main_loop_run () at /lib64/libglib-2.0.so.0
#9  0x00007ffff758ec5d in gtk_main () at /lib64/libgtk-3.so.0
#10 0x00007ffff2c31840 in gnc_ui_start_event_loop () at /home/janssege/Development/gnucash/maint/gnucash/gnome-utils/gnc-gnome-utils.c:671
#11 0x0000000000405eac in inner_main (closure=0x0, argc=1, argv=0x7fffffffcf98) at /home/janssege/Development/gnucash/maint/gnucash/gnucash-bin.c:680
#12 0x00007ffff7d7a901 in invoke_main_func () at /lib64/libguile-2.2.so.1
#13 0x00007ffff7d5b85e in c_body () at /lib64/libguile-2.2.so.1
#14 0x00007ffff7ddf4d7 in vm_regular_engine () at /lib64/libguile-2.2.so.1
#15 0x00007ffff7de518f in scm_call_n () at /lib64/libguile-2.2.so.1
#16 0x00007ffff7dd2e0f in catch () at /lib64/libguile-2.2.so.1
#17 0x00007ffff7d5be74 in scm_i_with_continuation_barrier () at /lib64/libguile-2.2.so.1
#18 0x00007ffff7d5bf59 in scm_c_with_continuation_barrier () at /lib64/libguile-2.2.so.1
#19 0x00007ffff7dd1802 in with_guile () at /lib64/libguile-2.2.so.1
#20 0x00007ffff7bcc389 in GC_call_with_stack_base () at /lib64/libgc.so.1
#21 0x00007ffff7dd1c0c in scm_with_guile () at /lib64/libguile-2.2.so.1
#22 0x00007ffff7d7aae6 in scm_boot_guile () at /lib64/libguile-2.2.so.1
#23 0x0000000000406375 in main (argc=1, argv=0x7fffffffcf98) at /home/janssege/Development/gnucash/maint/gnucash/gnucash-bin.c:938
(gdb) 
```

That crash happens after I select two non-adjacent transactions using ctrl-click (both are already green) and then click on the Assign transactions... in the popup menu.

This is with a build stared in a clean build directory.

Unfortunately the backtrace is of little help. It doesn't hint at where in the gnucash code this issue starts :(

jralls:
You need to install the [debuginfo](https://fedoraproject.org/wiki/StackTraces#Installing_debuginfo_RPMs_using_DNF) for glib-2.0 and gtk to get an intelligible stack trace. 

I suspect from the trace that there's a problem with creating the closure for the [menu activation signal](https://github.com/Gnucash/gnucash/pull/497/files#diff-6c7aa6b28614a3c4055d11fcf4803b61R558) since all of the processing is in glib and gdk.

gjanssens:
That indeed does give a little more details:

```
Thread 1 "gnucash" received signal SIGSEGV, Segmentation fault.
magazine_chain_pop_head (magazine_chunks=<optimized out>) at gslice.c:538
538     gslice.c: No such file or directory.
(gdb) bt
#0  0x00007ffff7b04552 in magazine_chain_pop_head (magazine_chunks=<optimized out>) at gslice.c:538
#1  0x00007ffff7b04552 in thread_memory_magazine1_alloc (tmem=<optimized out>, ix=1) at gslice.c:841
#2  0x00007ffff7b04552 in g_slice_alloc (mem_size=mem_size@entry=24) at gslice.c:1015
#3  0x00007ffff7ae19ea in g_list_prepend (list=list@entry=0x0, data=0x4237660) at glist.c:314
#4  0x00007ffff72943cd in convert_toplevel_coords_to_window (window=<optimized out>, 
    window@entry=0x4237660 [GdkX11Window], toplevel_x=toplevel_x@entry=989, toplevel_y=toplevel_y@entry=227, window_x=window_x@entry=0x7fff50001b78, window_y=window_y@entry=0x7fff50001b80) at gdkwindow.c:7745
#5  0x00007ffff7295fe6 in send_crossing_event
    (display=display@entry=0x45b0e0 [GdkX11Display], toplevel=toplevel@entry=0x385bc90 [GdkX11Window], window=0x4237660 [GdkX11Window], type=type@entry=GDK_ENTER_NOTIFY, mode=mode@entry=GDK_CROSSING_UNGRAB, notify_type=notify_type@entry=GDK_NOTIFY_VIRTUAL, subwindow=0x42374d0 [GdkX11Window], device=0x489810 [GdkX11DeviceCore], source_device=0x489810 [GdkX11DeviceCore], toplevel_x=toplevel_x@entry=989, toplevel_y=toplevel_y@entry=227, mask=GDK_MOD2_MASK, time_=36651190, event_in_queue=0x0, serial=116011) at gdkwindow.c:8337
#6  0x00007ffff72963a9 in _gdk_synthesize_crossing_events
    (display=0x45b0e0 [GdkX11Display], src=<optimized out>, dest=<optimized out>, device=0x489810 [GdkX11DeviceCore], source_device=0x489810 [GdkX11DeviceCore], mode=GDK_CROSSING_UNGRAB, toplevel_x=989, toplevel_y=227, mask=GDK_MOD2_MASK, time_=36651190, event_in_queue=0x0, serial=116011, non_linear=<optimized out>) at gdkwindow.c:8481
#7  0x00007ffff727b80c in synthesize_crossing_events
    (display=display@entry=0x45b0e0 [GdkX11Display], device=device@entry=0x489810 [GdkX11DeviceCore], source_device=source_device@entry=0x489810 [GdkX11DeviceCore], src_window=0x42a1320 [GdkX11Window], dest_window=dest_window@entry=0x42377f0 [GdkX11Window], crossing_mode=crossing_mode@entry=GDK_CROSSING_UNGRAB, time=36651190, serial=116011) at gdkdisplay.c:975
#8  0x00007ffff727c6a6 in switch_to_pointer_grab
    (display=display@entry=0x45b0e0 [GdkX11Display], device=device@entry=0x489810 [GdkX11DeviceCore], source_device=source_device@entry=0x489810 [GdkX11DeviceCore], grab=grab@entry=0x0, last_grab=last_grab@entry=0x42d4020, time=time@entry=36651190, serial=116011) at gdkdisplay.c:1121
#9  0x00007ffff727c812 in _gdk_display_device_grab_update
    (display=display@entry=0x45b0e0 [GdkX11Display], device=device@entry=0x489810 [GdkX11DeviceCore], source_device=source_device@entry=0x489810 [GdkX11DeviceCore], current_serial=current_serial@entry=116011) at gdkdisplay.c:1200
#10 0x00007ffff72974c0 in _gdk_windowing_got_event (display=display@entry=0x45b0e0 [GdkX11Display], event_link=0x3edc520 = {...}, event=event@entry=0x3daa350, serial=116011) at gdkwindow.c:9921
#11 0x00007ffff72b27b6 in _gdk_x11_display_queue_events (display=0x45b0e0 [GdkX11Display]) at gdkeventsource.c:348
#12 0x00007ffff727bc34 in gdk_display_get_event (display=display@entry=0x45b0e0 [GdkX11Display]) at gdkdisplay.c:438
#13 0x00007ffff72b2506 in gdk_event_source_dispatch (source=<optimized out>, callback=<optimized out>, user_data=<optimized out>) at gdkeventsource.c:363
#14 0x00007ffff7ae606d in g_main_dispatch (context=0x48b6e0) at gmain.c:3182
#15 0x00007ffff7ae606d in g_main_context_dispatch (context=context@entry=0x48b6e0) at gmain.c:3847
#16 0x00007ffff7ae6438 in g_main_context_iterate (context=0x48b6e0, block=block@entry=1, dispatch=dispatch@entry=1, self=<optimized out>) at gmain.c:3920
#17 0x00007ffff7ae6762 in g_main_loop_run (loop=0x5f90f0) at gmain.c:4116
#18 0x00007ffff758ec5d in gtk_main () at gtkmain.c:1323
#19 0x00007ffff2c31840 in gnc_ui_start_event_loop () at /home/janssege/Development/gnucash/maint/gnucash/gnome-utils/gnc-gnome-utils.c:671
#20 0x0000000000405eac in inner_main (closure=0x0, argc=1, argv=0x7fffffffcf98) at /home/janssege/Development/gnucash/maint/gnucash/gnucash-bin.c:680
#21 0x00007ffff7d7a901 in invoke_main_func (body_data=0x7fffffffce20) at init.c:340
#22 0x00007ffff7d5b85e in c_body (d=0x7fffffffcd60) at continuations.c:422
#23 0x00007ffff7ddf4d7 in vm_regular_engine (thread=0x20, vp=0xaf7f30, registers=0x7fffee228df0, resume=1) at vm-engine.c:784
#24 0x00007ffff7de518f in scm_call_n (proc=proc@entry=0xbd12a0, argv=argv@entry=0x0, nargs=nargs@entry=0) at vm.c:1257
#25 0x00007ffff7d6190d in scm_call_0 (proc=proc@entry=0xbd12a0) at eval.c:481
#26 0x00007ffff7dd2e0f in catch (tag=tag@entry=0x404, thunk=0xbd12a0, handler=0xbd1280, pre_unwind_handler=0xbd1260) at throw.c:137
#27 0x00007ffff7dd3199 in scm_catch_with_pre_unwind_handler (key=key@entry=0x404, thunk=<optimized out>, handler=<optimized out>, pre_unwind_handler=<optimized out>) at throw.c:254
#28 0x00007ffff7dd3383 in scm_c_catch
    (tag=tag@entry=0x404, body=body@entry=0x7ffff7d5b850 <c_body>, body_data=body_data@entry=0x7fffffffcd60, handler=handler@entry=0x7ffff7d5baf0 <c_handler>, handler_data=handler_data@entry=0x7fffffffcd60, pre_unwind_handler=pre_unwind_handler@entry=0x7ffff7d5b940 <pre_unwind_handler>, pre_unwind_handler_data=0xaf0080) at throw.c:377
#29 0x00007ffff7d5be74 in scm_i_with_continuation_barrier
    (body=body@entry=0x7ffff7d5b850 <c_body>, body_data=body_data@entry=0x7fffffffcd60, handler=handler@entry=0x7ffff7d5baf0 <c_handler>, handler_data=handler_data@entry=0x7fffffffcd60, pre_unwind_handler=pre_unwind_handler@entry=0x7ffff7d5b940 <pre_unwind_handler>, pre_unwind_handler_data=0xaf0080) at continuations.c:360
#30 0x00007ffff7d5bf59 in scm_c_with_continuation_barrier (func=<optimized out>, data=<optimized out>) at continuations.c:456
#31 0x00007ffff7dd1802 in with_guile (base=base@entry=0x7fffffffcdc0, data=data@entry=0x7fffffffcdf0) at threads.c:660
#32 0x00007ffff7bcc389 in GC_call_with_stack_base (fn=fn@entry=0x7ffff7dd1730 <with_guile>, arg=arg@entry=0x7fffffffcdf0) at misc.c:1949
#33 0x00007ffff7dd1c0c in scm_i_with_guile (dynamic_state=<optimized out>, data=0x7fffffffcdf0, func=0x7ffff7d7a8e0 <invoke_main_func>) at threads.c:703
#34 0x00007ffff7dd1c0c in scm_with_guile (func=func@entry=0x7ffff7d7a8e0 <invoke_main_func>, data=data@entry=0x7fffffffce20) at threads.c:709
#35 0x00007ffff7d7aae6 in scm_boot_guile (argc=1, argv=0x7fffffffcf98, main_func=0x405cae <inner_main>, closure=0x0) at init.c:323
#36 0x0000000000406375 in main (argc=1, argv=0x7fffffffcf98) at /home/janssege/Development/gnucash/maint/gnucash/gnucash-bin.c:938
```

DaveC49:
Geert , John,
I have just rebuilt the ImportMultiSelect branch on my computer (Linux Mint 19.1) from scratch and the compiled version is running with with no problems at all and happily selecting multiple transactions, assigning a transfer account to the selection and importing the transactions. I am wondering whether I am using different library versions for libglib, libgdk and gtk-3.0 perhaps. These are what I currently have installed:
$ dpkg -l | grep libglib
ii  libglib-object-introspection-perl                                0.044-2 
ii  libglib-perl                                                                3:1.326-1build1
ii  libglib2.0-0:amd64                                                   2.56.4-0ubuntu0.18.04.2
ii  libglib2.0-bin                                                            2.56.4-0ubuntu0.18.04.2
ii  libglib2.0-cil                                                             2.12.40-2
ii  libglib2.0-cil-dev                                                      2.12.40-2
ii  libglib2.0-data                                                         2.56.4-0ubuntu0.18.04.2
ii  libglib2.0-dev:amd64                                              2.56.4-0ubuntu0.18.04.2 
ii  libglib2.0-dev-bin                                                    2.56.4-0ubuntu0.18.04.2
ii  libglib2.0-doc                                                          2.56.4-0ubuntu0.18.04.2 
ii  libglib2.0-tests                                                        2.56.4-0ubuntu0.18.04.2
ii  libglib3.0-cil                                                            2.99.3-2
ii  libglib3.0-cil-dev                                                     2.99.3-2
ii  libglibmm-2.4-1v5:amd64                                      2.56.0-1
ii  libglibmm-2.4-dev:amd64                                      2.56.0-1

$ dpkg -l | grep gdk
ii  gir1.2-gdkpixbuf-2.0:amd64                                    2.36.11-2
ii  libgdk-pixbuf2.0-0:amd64                                       2.36.11-2
ii  libgdk-pixbuf2.0-bin                                                2.36.11-2
ii  libgdk-pixbuf2.0-common                                       2.36.11-2
ii  libgdk-pixbuf2.0-dev                                               2.36.11-2
ii  libgdk3.0-cil                                                            2.99.3-2
ii  libgdk3.0-cil-dev                                                     2.99.3-2

ii  libgtk-3-0:amd64                                                    3.22.30-1ubuntu3
ii  libwebkitgtk-3.0-0:amd64                                       2.4.11-3ubuntu3
ii  libwebkitgtk-3.0-dev:amd64                                   2.4.11-3ubuntu3
ii  libgtk3.0-cil                                                             2.99.3-2
ii  libgtk3.0-cil-dev                                                      2.99.3-2
ii  libwxgtk3.0-0v5:amd64                                          3.0.4+dfsg-3
ii  libwxgtk3.0-dev                                                      3.0.4+dfsg-3
ii  libwxgtk3.0-gtk3-0v5:amd64                                  3.0.4+dfsg-3
ii  libwxgtk3.0-gtk3-dev                                              3.0.4+dfsg-3

David

Bob-IT:
OK, I am still confused about this but maybe you have not rebased your branch against the current upstream maint branch as clicking on the files tab above and then the '...' and view file for the source file it is still missing current maint changes. 
Any way I manually added your changes to my local maint branch and tried it after changing the pop-up gtk versioning to Geert's recommended changes.

All seemed to work until I hit cancel on the 'select Account' dialogue causing a crash, this was tracked down to the last line of 'gnc_gen_trans_assign_transfer_account_to_selection_cb', you do not need the g_list_free(l,....

There is one other thing I would change, if no rows are selected, the pop-up still works but does nothing so check for a selection before popping.

jralls:
Good catch, @Bob-IT! @DaveC49 do you understand why that second ```g_list_free_full``` causes a crash?

DaveC49:
Thanks Bob. I am getting the same behavior using cancel from the dialog. John, I don't at the moment but I will. My guess is I am trying to free something which hasn't been allocated when you use cancel. I will check out what happens on the OK and cancel paths.

 I am not sure what my problem is with git at the moment. My son who is a software engineer visited yesterday and gave me a few pointers. I am starting to understand what happens in rebasing. I have really only used github in the past in a situation where at most there was one other developer in the loop and not much turnover so it is a new experience for me. I will incorporate the change Geert mentioned to use GTK_CHECK_VERSION(3,22,0) and sort the problem with using Cancel on the account selection dialog.

jralls:
```C
       for (l = selected_rows; l != NULL; l = l->next)
       {
 ...
            if (assigned_account == NULL) break;
...
      }
      g_list_free_full (selected_rows, (GDestroyNotify) gtk_tree_path_free);
      g_list_free_full(l, (GDestroyNotify) gtk_tree_path_free);
```
David,
```l``` and ```selected_rows``` start off pointing to the same list and ```g_list_free_full (selected_rows)``` frees all of the nodes on the list. If the loop has completed then ```l``` is ```NULL``` and ```g_list_free_full``` does nothing, so no problem. But if the ```break``` line happened then ```l``` points at one of the already-freed nodes in the list and the second ```g_list_free_full``` call tries to free it again and that crashes.

File: gnucash/import-export/import-main-matcher.c [hunk 7]
```diff
@@ -361,54 +377,247 @@ gnc_gen_trans_update_toggled_cb (GtkCellRendererToggle *cell_renderer,
...
+                     gnc_gen_trans_assign_transfer_account_to_selection_cb),
+                     info);
+    DEBUG("Callback to assign destination account to selection connected");
+    gtk_menu_shell_append(GTK_MENU_SHELL(menu), menuitem);
+    gtk_widget_show_all(menu);
+    event_button = (GdkEventButton *) event;
+    /* Note: event can be NULL here when called from view_onPopupMenu;
+     *  gdk_event_get_time() accepts a NULL argument */
+    /* replace deprecated function for GTK versions >=3.22 */
+#if GTK_CHECK_VERSION(3,22,0)
```
jralls: You did this backwards. If ```GTK_CHECK_VERSION(3,22,0) ``` is true then ```gtk_menu_popup_at_pointer``` is available, otherwise fall back on ```gtk_menu_popup```.

DaveC49:
Oops! I'll fix thatOn Mon, 2019-05-27 at 13:49 -0700, John Ralls wrote:
> @jralls commented on this pull request.
> 
> 
> 
> 
> In gnucash/import-export/import-main-matcher.c:
> 
> > ++    ENTER ("");+    menu = gtk_menu_new();+    menuitem = gtk_menu_item_new_with_label(_("Assign a transfer
> account."));+    g_signal_connect(menuitem,
> "activate",+                     G_CALLBACK(+                     gnc_gen_trans_assign_transfer_account_to_selection_c
> b),+                     info);+    DEBUG("Callback to assign destination account to selection
> connected");+    gtk_menu_shell_append(GTK_MENU_SHELL(menu),
> menuitem);+    gtk_widget_show_all(menu);+    event_button = (GdkEventButton *) event;+    /* Note: event can be NULL
> here when called from view_onPopupMenu;+     *  gdk_event_get_time() accepts a NULL argument */+    /* replace
> deprecated function for GTK versions >=3.22 */+#if GTK_CHECK_VERSION(3,22,0)
> 
> You did this backwards. If GTK_CHECK_VERSION(3,22,0)  is true then gtk_menu_popup_at_pointer is available, otherwise
> fall back on gtk_menu_popup.
> 
> 
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub, or mute the thread.
> 
> [
> {
> "@context": "http://schema.org",
> "@type": "EmailMessage",
> "potentialAction": {
> "@type": "ViewAction",
> "target": "
> https://github.com/Gnucash/gnucash/pull/497?email_source=notifications\u0026email_token=ABW4CAZC6W3KV5DEOWJYBRDPXRCOBA5CNFSM4HITSV4KYY3PNVWWK3TUL52HS4DFWFIHK3DMKJSXC5LFON2FEZLWNFSXPKTDN5WW2ZLOORPWSZGOBZZISLY#pullrequestreview-242387247
> ",
> "url": "
> https://github.com/Gnucash/gnucash/pull/497?email_source=notifications\u0026email_token=ABW4CAZC6W3KV5DEOWJYBRDPXRCOBA5CNFSM4HITSV4KYY3PNVWWK3TUL52HS4DFWFIHK3DMKJSXC5LFON2FEZLWNFSXPKTDN5WW2ZLOORPWSZGOBZZISLY#pullrequestreview-242387247
> ",
> "name": "View Pull Request"
> },
> "description": "View this Pull Request on GitHub",
> "publisher": {
> "@type": "Organization",
> "name": "GitHub",
> "url": "https://github.com"
> }
> }
> ]
-- 
Dr David R Cousens
B.Sc, M.Prof. Acc., Ph.D., G.C.Ed

DaveC49:
@jralls John, I have put the changes Bo and Geert reommended in and Cancel on the account selection is no longer causing the segmentation error. I have squashed it onto the previous commit and pushed the fix to my repo. I take it you will be notified of that when the pull request is still active.

gnucash/import-export/import-main-matcher.c
```diff
@@ -361,54 +377,247 @@ gnc_gen_trans_update_toggled_cb (GtkCellRendererToggle *cell_renderer,
...
-        /*The information displayed is only informative, until you select an action*/
-        break;
-    default:
-        PERR("I don't know what to do! (Yet...)");
-        break;
     }
-    refresh_model_row(gui, model, &iter, trans_info);
+
+    g_list_free_full (selected_rows, (GDestroyNotify) gtk_tree_path_free);
+/*    g_list_free_full(l, (GDestroyNotify) gtk_tree_path_free);
```
jralls: Delete it, don't comment it out.

jralls:
@Bob-IT, @gjanssens are you satisfied with the latest version?

Bob-IT:
Unless I am using the file button on github incorrectly I still think this PR needs updating with current maint. Looking at your branch, it says its is 232 commits behind maint so I would advise you update your maint by doing something like the following in your git directory...
git fetch upstream
git checkout maint
git rebase upstream/maint
git checkout ImportMatchMultiSeleect
git rebase upstream/maint
git push --force

There is probably a easier / better way but that should get the job done.

jralls:
@Bob-IT it doesn't matter if the branch is behind as long as it will merge cleanly. Github tests that and reports at the bottom.

jralls:
@gjanssens I think the only question left is whether this should go on maint or master, considering that it's a minor UI change.

@DaveC49 I think this is pretty undiscoverable. What about a change to Help documenting it?

DaveC49:
@Jralls John the Help section has already been written and incorporated in the current release.  I had originally put this as a branch off master last year but when the mofified Help came out  before the branch was incorporated, I revisited it as the current  branch off maint.  @Bob-IT I will try and sort out the rebasing issue and get the workflow right even if it is mainly for future reference.

jralls:
@DaveC49 There's no need to rebase.

Bob-IT:
OK, that would be me being over cautious as I try to test stuff I cobble together on the current source just to verify I have not mucked it up again.

gjanssens:
As this is a small change, I'm fine with this going into maint.

I'm also happy to report it no longer crashes. There's one more glitch that I think should be fixed:
If the user doesn't select a transaction, but instead directly right-clicks one, one can select the Assign transaction from the pop-up menu that appears, but it does nothing. There's no visual feedback at all which is pretty confusing.
So either the pop-up should be disabled when no line is selected or a right-click on a previously unselected line should do two things:
1. select that line
2. show the pop-up menu

Note that selecting in that case should still adhere to any modifier keys like shift or ctrl.

jralls:
@Bob-IT the recipe is:
```
git checkout -b ImportMultiSelect -t maint
git pull https://github.com/DaveC49/gnucash ImportMultiSelect
```
That will create a merge commit. Change the summary line to "Merge David Cousens's ImportMultiSelect into maint."

Test away. When satisfied,
```
git checkout maint
git merge ImportMultiSelect
```
That will do a fast-forward including the already prepared merge commit.

Bob-IT:
Thanks John, another snippet to be noted for later.

DaveC49:
Geert , John,
I've added the extra changes so that if there is no selection made then the popup menu can't be pulled up with either a right click or Shift-F10.  I thought I had done it but it I hadn't gone far enough back in the chain. The pull request is closed so will I need to do another one?

gjanssens:
Normally I would simply reopen it but for some reason that button is greyed out (hovering the button states it was force-pushed as reason).

Regardless, I can also merge a closed PR so don't bother creating a new one. I'll try to look into this in the afternoon.

gjanssens:
@DaveC49 The extra changes take care of the final issue I had with this PR. The code works well now.

On the other hand the commit history was still a bit odd: there were two commits starting from most recent maint both with identical commit message and almost identical diff. These two commits were then merged together on your maint branch. I have chosen to rebase the most recent commit on the older one (which resulted in a difference of exactly your last fix). I have then further squashed these two together and only pushed this unified commit to maint.

So your changes are now officially queued for gnucash 3.6 (after a bit of massaging of the commit history). Thanks for your contriubtions!

christopherlam:
I'm on the sideline here but very very grateful about this work :)

Bob-IT:
Looks like all the column sorting I added for date and amount has been removed, I will add it all back as a separate commit at the weekend.

gjanssens:
I didn't realize that. Sorry.

Bob-IT:
No problem.

DaveC49:
@Bob-IT  Apologies if I wiped out the column sorting with my incompetence with git. I think I've got the hang of it now with the recipe you gave me above and i now have the hang of what it is doing. I'm now doing the fetch and rebase before I start doing any work. Thanks to you John and Geert for all the help and he work on GnuCash.

Bob-IT:
multi-select
I was adding the Date and amount sorting back and while trying your multi-select I noticed some odd behaviour, hopefully you can see the image OK.
In the Image I can select the lines 'Dividens', 'Insurance' and right mouse works but the following does not 'Subscriptions', 'Books' it right mouses OK but no account dialogue ?
I thought it was the order but that makes no difference...

jralls:
Should one be able to combine an add and a reconcile in a single interaction? When done one at a time they present different dialog boxes.

DaveC49:
@Bob-IT I think you will first need to uncheck the R box as this is flagging the transaction as matching an existing transaction already in GnuCash and not to be imported. (The matcher often matches  my pension  payments as being already present. They occur every two weeks so must be within the date range window for matching and the amount and description are usually identical.) To be able to assign a transfer account to them and import them, I uncheck the R box which then changes the background to orange and checks the A box flagging it to be imported. If you only try to select lines which are flagged for import (i.e. A checked) does it work as expected. In the case of transactions flagged for Updating or as exact matches ( U or R checked) I think it would be the correct behavior to be not able to assign a transfer account (as presumably there is already an existing one assigned in the existing GnuCash record) and for that transaction to be not even selectable for matching or inclusion in a selection. AFAIK there is nothing in the code I added which enforces the latter as the treeview's selection process does not know implicitly about the A U and R checkboxes. The action field in the trans_info does records the state of the checkboxes so it may be some logic around that. I could perhaps add something to the selection call back to check the import action field to prevent lines with U or R checked being added to a selection at all  if you feel that is necessary. We all know that in user land if it can be done, it will be, whether it should be or not.

christopherlam:
Apologies for begging; would you guys please augment the `"Reconcile (auto) match"` text to read `"Reconcile (auto) to 'existing-desc'/'existing-memo'"` whereby these refer to the matched existing split details? Ditto `Reconcile (manual) match`.

My workflow includes manually-entered transactions (with useful free-text details) and periodic OFX imports (to catch missing txns) and it really grates me that I *cannot* visually verify that the auto-match heuristics identified the correct transaction for matching.

Bob-IT:
@christopherlam Have a look at PR 518 and give it a spin if you can.

Bob-IT:
@DaveC49 I think I know why it is different, the selection is processed from top to bottom and the looping in gnc_gen_trans_assign_transfer_account_to_selection_cb bombs out when the assigned_account is NULL and as it decides the first row is an update, jumps to GNCImport_UPDATE but 'first' and 'is_selected' is TRUE so does not run the match dialog.
I think this should deal with rows that are the same which I assume is 'New' ones and so on right mouse the selection should be scanned and any that are not new unselected or block on selection if there are multiple row.

DaveC49:
@Bob-It The best way to handle it may be to check the selected rows in gnc_gen_trans_onButtonPressed_cb and again skip popping up the menu at all if the rows in the selection are not marked to be Imported. Any check laeter on and the popup to assign the transfer account is already activated,  I might add this as another new bug branch and a separate pull request to avoid complication the already complicated.

Bob-IT:
@DaveC49  You could probably do it all in gnc_gen_trans_view_popup_menu before doing the popup, so test for selection greater than 1, scan selection for just 'New' ones and unselect any others and then do pop. If in this function, you can remove the count test from ....Pressed_cb and ....Menu_cb as they both end up here. 
Yes do al this in a new branch and create a new PR, make sure you update your local maint before creating your new branch, any way have fun.

jralls:
> You could probably do it all in gnc_gen_trans_view_popup_menu 

But please don't. Please use the [MVC pattern](https://en.wikipedia.org/wiki/Model%E2%80%93view%E2%80%93controller) at least by separating the parts into different functions, ideally with separate files. Prior failure to do this is an enormous source of technical debt in GnuCash, we don't want to add to that debt with new code.

DaveC49:
John There is a select function in treeview which is called when a row is selected which can return a TRUE/FALSE which either allows the row to be selected or not. I will look at creating a function to do that based on the trans_info content for the row. This will probably be the best way to do it. I raised Bug 797275 to cover this and will do it in a new branch.  David

jralls:
OK.

christopherlam:
There's an important regression introduced by this commit.

Doubleclick on an import transaction should pop up a 'select matching existing transaction' dialog with but this is not possible anymore. This is important to verify the matcher found the correct register transaction, and select the correct one if it mismatched.

DaveC49:
Chris,
If the reconcile or the update and reconcile check boxes are checked this behavior should be the expected response as an existing transaction has been matched and similarly if all the checkboxes are unchecked - no action is defined. . It is only with the Add checkbox checked that a double click will bring up the select account dialog. The changes for the multiple selection didn't change the behavior in this regard as it is set by switch case code in the functions initiated by the callbacks.
David

christopherlam:
@DaveC49 I am extremely grateful for your work.

But there's definitely regression -- see my OFX import -- the two OFX transactions have been nicely matched to register transactions, but (1) I have zero clue whether they matched with the correct one -- this is very important to me to avoid accidentally entering transactions twice (2) I can double-click the automatically matched transaction to bring up the list of candidates and the matched one. I was running build cc12fcf32 just prior to merging this work.

After running build 5f51fc540 (merging this work), doubleclick on the matched transaction doesn't do anything; I have no ability to verify the matching was accurate, nor amend an inaccurate matching.

Bob-IT:
I had a look at this, and the problem comes down to this code...
            if (first && !is_selection)
                run_match_dialog (info, trans_info);

The problem with this is that you are passing pointers for gboolean first and is_selection so the above should be if (*first && !*is_selection) but I am look at why you were passing them in the first place and not just gboolean's.

Bob-IT:
OK I have pushed a fix for this, changed the two gboolean pointers to just gbooleans, worked locally so it should work for you but please test.

DaveC49:
Bob I don't think there was any good reason for passing the pointers rather than the values other than it was probably the only bit of C++ coding I've done since 1989. i'll duplicate the changes and check it out.

David

christopherlam:
Thank you :) :) :)

DaveC49:
Bob,

 Your changes wotk fine on my system as well and the matching dialog is coming up as Chris  said it should

David

DaveC49:
Bug https://bugs.gnucash.org/show_bug.cgi?id=797275 raised to cover selection of transactions which are not to be imported.  I have looked at the treeview docs and it is possible to set  a select function which is called before a row is selected which returns TRUE if the row  can be selected and FALSE if not. It should be possible to define a function which will test the row content to see what flags are set

DaveC49:
Bob,
Actually there was a reason for passing the first and is_selection flags as pointers, at least for the first flag. It is initially set but was being unset in the gnc_gen_trans_assign_transfer_account() call so that on subseqent calls the dialog for the transfer account selection would not be displayed and the value selected on the first one is used for the subsequent members of the selection. If it is passed by value rather than passed by pointer, the first flag in the gnc_gen_trans_assign_transfer_account_to_selection_cb is not unset so an incorrect value is then passed for the subsequent  members of the selection. If it is simply unset in gnc_gen_trans_assign_transfer_account_to_selection_cb then if the account selection is cancelled rather than ok_pressed in this function this is not passed back to gnc_gen_trans_assign_transfer_account_to_selection_cb. I have a view-Selection_function working for the treeview now so I'll fix up the logic on this one so it fixes this and still displays the dialog  that was blocked in the bug branch 797275 .
David

Bob-IT:
OK, I missed that, maybe you could add an inline comment where set. The other thought I had but not necessary to look at was that new_acc would be set to a non null value on the first use so may be you could use that as your test.

DaveC49:
Bob Chris  I've created a pull reqest for branch bug797275 which I think both fixes the problem of only allowing selection of rows flagged for adding and allows the match dialog to be run on double clicking. I think it now works OK but I hyaven't got a good test case for using the match dialog so i would appreciate it if you could check it out Chris.

jralls:
Reminder for all: This needs to be merged by 1900 UTC (Noon California time) Saturday if it's going to make 3.6. @Bob-IT I'm leaving it to you for thumbs up/down.

PR #498, opened 2019-04-26
Modularise the transaction report

christopherlam:
This commit will modify:
* transaction.scm
* income-gst-statement.scm
* general-ledger.scm

and creates 
* reconcile-report.scm
* a new module `trep-engine.scm` which houses the full transaction report engine.

The 4 reports all use functions from `trep-engine.scm` which means all reports are now independent of each other, and can be `load`ed cleanly. The test suite has minimal amendments, otherwise run unchanged, confirming the report behaviours are functionally unaltered.


christopherlam:
This line of work supports https://code.gnucash.org/logs/2019/04/04.html#T03:09:55 would mean all standard-reports should be standalone reports, @gjanssens can take up the mantle whenever to change `resolve-interface` to `(for-each load reportlist)` or something.

fellen:
Perhaps rename trep-engine.scm to txn-rep-engine.scm? So already the name would give a clue of its content.

christopherlam:
If the team don't mind I'll be aiming to merge this soon, and also merge the CSV output https://github.com/christopherlam/gnucash/tree/maint-export-csv - the latter is a simple csv generator which simply lists the headings their overall totals in .csv format.

I think this approach can be used to generate a UK-specific VAT report, based upon the pre-existing VAT account structure offered in the New Accounts wizard.

PR #500, opened 2019-05-03
[UK] VAT and [AU] BAS report

christopherlam:
DRAFT awaiting VAT-report finalisation.

This commit enables CSV export of transaction-report and gst-report monetary columns and grand-totals.

Sample CSV output
```csv
"from","2018-01-01"
"to","2018-03-31"
"Debit",4270.88
"Credit",8314.44
```
Note dates are iso-format strings.

If a grand-total has multiple currencies, all will be listed without currency; e.g. grand-total 100GBP + 50USD will show as: `"Amount",100,50`. The order is not specified.

### DIFF
--- a/gnucash/report/reports/standard/income-gst-statement.scm
+++ b/gnucash/report/reports/standard/income-gst-statement.scm
@@ -33,36 +33,60 @@
 (use-modules (gnucash app-utils))
 (use-modules (gnucash report))
 (use-modules (srfi srfi-1))
+(use-modules (ice-9 match))
 
 ;; Define the strings here to avoid typos and make changes easier.
 (define reportname (N_ "Income and GST Statement"))
 (define pagename-sorting (N_ "Sorting"))
 (define pagename-filter (N_ "Filter"))
+(define pagename-format (N_ "Format"))
+
 (define TAX-SETUP-DESC
   (gnc:make-html-text
-   (_ "This report is useful to calculate periodic business tax payable/receivable from \
-authorities. From 'Edit report options' above, choose your Business Income and Business Expense accounts. \
-Each transaction may contain, in addition to the accounts payable/receivable or bank accounts, \
-a split to a tax account, e.g. Income:Sales -$1000, Liability:GST on Sales -$100, Asset:Bank $1100.")
-   (gnc:html-markup-br)
-   (gnc:html-markup-br)
-   (_ "These tax accounts can either be populated using the standard register, or from Business Invoices and Bills \
-which will require Tax Tables to be set up correctly. Please see the documentation.")
-   (gnc:html-markup-br)
-   (gnc:html-markup-br)
-   (_ "From the Report Options, you will need to select the accounts which will \
+
+   (gnc:html-markup-p
+    (_ "This report is useful to calculate periodic business tax \
+payable/receivable from authorities. From 'Edit report options', \
+choose your business sales and purchase accounts. Each transaction \
+may contain, in addition to the asset, liability, A/Payable or \
+A/Receivable accounts, a split to a tax account, e.g. Income:Sales \
+-$1000, A/Receivable $1100, Liability:GST on Sales -$100."))
+
+   (gnc:html-markup-p
+    (_ "These tax accounts can either be populated using the standard register, or from Business Invoices and Bills \
+which will require Tax Tables to be set up correctly. Please see the documentation."))
+
+   (gnc:html-markup-p
+    (_ "From the Report Options, you will need to select the accounts which will \
 hold the GST/VAT taxes collected or paid. These accounts must contain splits which document the \
 monies which are wholly sent or claimed from tax authorities during periodic GST/VAT returns. These \
-accounts must be of type ASSET for taxes paid on expenses, and type LIABILITY for taxes collected on sales.")
-   (gnc:html-markup-br)
-   (gnc:html-markup-br)))
+accounts must be of type ASSET for taxes paid on expenses, and type LIABILITY for taxes collected on sales."))
 
-(define (gst-statement-renderer rpt)
-  (gnc:trep-renderer
-   rpt
-   #:custom-calculated-cells gst-calculated-cells
-   #:empty-report-message TAX-SETUP-DESC
-   #:custom-split-filter gst-custom-split-filter))
+   (gnc:html-markup-p
+    (_ "Note the UK variant may specify EU VAT accounts may be tagged \
+with *EUVAT* in the VAT account description. EU Goods sales and purchase \
+accounts may be tagged with *EUGOODS* in the account description."))
+
+   (gnc:html-markup-p
+    (_ "This message will be removed when tax accounts are specified."))))
+
+(define* (gst-statement-renderer rpt #:optional export-type file-name)
+  (define (opt-val section name)
+    (gnc:option-value
+     (gnc:lookup-option (gnc:report-options rpt) section name)))
+  (define sales-purch-accounts
+    (append (opt-val "Accounts" "Sales") (opt-val "Accounts" "Purchases")))
+  (define document
+    (gnc:trep-renderer
+     rpt
+     #:custom-calculated-cells gst-calculated-cells
+     #:custom-source-accounts sales-purch-accounts
+     #:custom-split-filter gst-custom-split-filter
+     #:export-type export-type
+     #:filename file-name))
+  (when (null? (opt-val "Accounts" "Tax Accounts"))
+    (gnc:html-document-add-object! document TAX-SETUP-DESC))
+  document)
 
 (define (gst-custom-split-filter split)
   ;; split -> bool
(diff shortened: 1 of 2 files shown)


jralls:
How does the user enable the CSV export and select the file path?

christopherlam:
File/ export / report will show options

jralls:
So draft pending VAT report and someone writing GUI for you...

christopherlam:
I don't think I'll need a GUI like the us tax system. All will be from report options, awaiting formulas and algorithms for vat/EC vat calculations from maf

jralls:
Well, there is no ```File>Export>Report``` menu item and the File menu isn't written in Scheme. In addition to the menu item you'll need some way to enable it if the current tab is a report able to export CSV and disable it otherwise. That will also have to be at least partly in C even if most of the functionality is in the options system. There's also popping a GtkFileChooser to set the output filename; more C.

christopherlam:
In this branch, Transaction Report:
image

File>Export>Export Report
image

Popup
image

FileChooser
image

Tada!
image

My current problem is - how to handle dual subtotals because it'll show debit/credit subtotals separately and not sure how these should be exported.

Example:
image
becomes:
```csv
"from","2018-01-01"
"to","2019-05-04"
"Debit (CNY)",150.0
"Credit (CNY)",750.0
"Debit",20.0
"Credit",100.0
```
and I *could* try put the difference subtotal into the appropriate column, but this means more code...

christopherlam:
I think I'll just enable csv output for reports where no monetary columns are dual-subtotals. It *would* be possible to calculate the dual-subtotal difference and place in the correct column, but would pretty much duplicate the `add-subtotal-row` function.

christopherlam:
So, from my **naive** understanding of VAT rules in EU, all it involves is charging VAT then immediately reversing it. eg. in UK receive 100 GBP from French customer, assuming 20% VAT rate in UK:

* Asset:Bank +100 GBP
* Income:ECSales -100 GBP
* VAT:VAT on Salers -20 GBP
* VAT:EC:Reverse EC Sales +20 GBP

I'm hoping @gjanssens can confirm my understanding. The above is easily reported, summed, and exported as CSV using the existing `income-gst-statement.scm` report.

I'll be refining `eu-vat-report.scm` whose *sole* upgrade cf income-gst-statement.scm is to allow more account-types in sales/purchase accounts. Whereas the previous report classified sales/purchases from the income/expense acct types, the new report would allow *any* account type for sales/purchases but would default to income/expense accounts. This will then allow capital asset purchases to be allowable as a purchase transaction with VAT claims.

jralls:
@gjanssens isn't in the UK, he's in Belgium. Maf King and @EvansMike are probably your best resources for UK Vat and MTD.

gjanssens:
I actually don't know whether all EU countries use the exact same VAT rules.
I do know the intracommunity agreement allows for a displacement of VAT. That is, I have to report it but can immediately offset it in the same report. And this is an agreement between all EU countries. Brexit may change this for the UK if/when it gets in effect.
How exactly this is recorded/reported may change per country or not. I have no idea.

In Belgium reporting intracommunity VAT involves at least 3 grids (one for the base amount, one for the charged VAT and one for the offset). However the grid for the base amount is not the same as the grid for the base amount for in-country sales. So the split you do above is right, but the income of that invoice should be reported separately (in a different grid) from income from sales in Belgium.

And then credit notes are treated differently still. Just to give you an idea, odoo (which I currently use to manage our businesses) has 75 VAT rules manipulating 30 VAT grids for Belgium alone !

For now I think you best focus on the UK situation as that is likely most urgent. We may evaluate later if we can implement a more generic rule system the fits multiple countries/jurisdictions.

gjanssens:
> I don't think I'll need a GUI like the us tax system. All will be from report options, awaiting formulas and algorithms for vat/EC vat calculations from maf

Unfortunately I suspect we will need a gui for the EU tax system(s). I highly doubt each EU country uses the exact same VAT rules.

frakturfreak:
> So, from my **naive** understanding of VAT rules in EU, all it involves is charging VAT then immediately reversing it. eg. in UK receive 100 GBP from French customer, assuming 20% VAT rate in UK:
> 
>     * Asset:Bank +100 GBP
> 
>     * Income:ECSales -100 GBP
> 
>     * VAT:VAT on Sales [liability] -20 GBP
> 
>     * VAT:EC:Reverse EC Sales [liability] +20 GBP
> 
> 
> I'm hoping @gjanssens can confirm my understanding. The above is easily reported, summed, and exported as CSV using the existing `income-gst-statement.scm` report.
> 
> I'll be refining `eu-vat-report.scm` whose _sole_ upgrade cf income-gst-statement.scm is to allow more account-types in sales/purchase accounts. Whereas the previous report classified sales/purchases from the income/expense acct types, the new report would allow _any_ account type for sales/purchases but would default to income/expense accounts. This will then allow capital asset purchases to be allowable as a purchase transaction with VAT claims.

You’ve got it totally wrong.

You only use the EC input and output VAT acccount for purchases.

For sales you’ve got to differentiate if you’re selling to another business or an end customer:

For another business it’s just the income account and nothing else, for sales to end customer it’s income + the standard output VAT rate of your country up to a certain threshold and afterwards the VAT of the customers country. (And this is just the basic case, there are different rules for e. g. billing online services).

christopherlam:
The VAT report now is a small upgrade from GST report.

Sales and Purchases accounts (they default to income+AR & expense+AP accounts) are now separately requested but can choose *any* account type including asset&liability accounts. VAT accounts chosen from Asset&Liability accounts.

I thought about how to detemine whether income/expense/VAT account relate to EC -- initially would create another accountlist option, but this becomes too many accounts to deal with. The current iteration now searches accounts via freetext tags: sales & purchase accounts can be tagged with "*ECGOODS*" somewhere in account description; VAT accounts can be tagged "*ECVAT*" and will be flagged and handled especially by the report.

For now the report is tailored to UK VAT report - it analyzes all splits from sales & purchase accounts within the date range (defaults to last quarter); for each split it analyzes the transaction, pulls out sales/purchases/EC-goods-sales and purchases/VAT/reverse-EC-VAT components and places the amounts in the correct columns labelled box 1-9.

Finally the columns 1-9 are summed and exportable as CSV. 

Opinion from interested parties welcome @EvansMike and possibly anyone who handles other EU VAT.

christopherlam:
what's left of VAT-report:

1. a decision on accounts selection - I've reduced accounts selection from 7 to 3 -- sales accts, purch accts, vat accounts. individual accounts can be tagged with \*ECVAT\* or \*ECGOODS\* to be handled differently. i.e. sales/purchase accounts can be tagged ECGOODS to be counted in box 8 and 9 UK VAT form; and VAT accounts can be tagged ECVAT to highlight EC handling. Is there a better way to tag these EC accounts?

2. the box 1-9 are tailored to UK VAT form therefore file named `uk-vat-report.scm`... it *could* be made more general (via eval-string text option?) but I suspect it'll be easier to just copy this report for other countries. Their definitions and labels will be different and will need different logic. I'd think this is now legible enough for any novice schemer to adapt at leisure...

3. more trep-options can be hidden...

christopherlam:
Further work suggestions:
1. modify income-gst-statement.scm to request sales and purchase accounts separately.
2. move tax-specific options to separate tab `Format` 
3. Add multichoice option to output as per UK-VAT rules.

Default
image

UK-VAT
image

fellen:
> The VAT report now is a small upgrade from GST report.
> 
> Sales and Purchases accounts (they default to income+AR & expense+AP accounts) are now separately requested but can choose _any_ account type including asset&liability accounts. VAT accounts chosen from Asset&Liability accounts.
> 
> I thought about how to detemine whether income/expense/VAT account relate to EC -- initially would create another accountlist option, but this becomes too many accounts to deal with. The current iteration now searches accounts via freetext tags: sales & purchase accounts can be tagged with "*ECGOODS*" somewhere in account description; VAT accounts can be tagged "*ECVAT*" and will be flagged and handled especially by the report. This strategy is obviously hackish and could be done via GUI, but is not nessary...
> 
> For now the report is tailored to UK VAT report - it analyzes all splits from sales & purchase accounts within the date range (defaults to last quarter); for each split it analyzes the transaction, pulls out sales/purchases/EC-goods-sales and purchases/VAT/reverse-EC-VAT components and places the amounts in the correct columns labelled box 1-9.
> 
> Finally the columns 1-9 are summed and exportable as CSV.
> 
> Opinion from interested parties welcome @EvansMike and possibly anyone who handles other EU VAT.

There is since over a decade no longer an EC. It became the EU. The ancestor of the EC has been the EEC, a pure economic community. They were also emitting other currencies  `"Ecu"+<number of members>` with ISO codes starting with "XE" instead of "EUR". So, please do not use historic names for recent organisations.

christopherlam:
> There is since over a decade no longer an EC. It became the EU. The ancestor of the EC has been the EEC, a pure economic community. They were also emitting other currencies `"Ecu"+<number of members>` with ISO codes starting with "XE" instead of "EUR". So, please do not use historic names for recent organisations.

Thank you @fellen - would changing the magic string to `*EUGOODS*` and `*EUVAT*` be better?

I learned from Reverse EC charges, literally from GnuCash GB account structure, and I guess it stuck through. https://github.com/Gnucash/gnucash/blob/93a24b62ce54dfc0d42f8409f2a7aaefac2830b1/data/accounts/en_GB/uk-vat.gnucash-xea#L140

I still do not think that using the Tax-Report-Options dialog is suitable for the purposes of this report because I do not wish to limit my report to a specific locale.

gjanssens:
> I still do not think that using the Tax-Report-Options dialog is suitable for the purposes of this report because I do not wish to limit my report to a specific locale.

Do you literally mean "locale" or also not limited to specific jurisdictions ?

If the first, I agree. The user should be able to select the tax rules that apply to their accounting data. And we should change our code such that this is possible independent of the locale that happens to be set for the user.

However every good tax report inevitably depends on tax details specific to the user's jurisdiction in my opinion. And then offering the user an interface to configure this tax data in detail rather than hard-coding it is highly valuable IMO. Our current Tax-Report-Options dialog is a first attempt in that direction. I agree though it needs much improvement to become more universally helpful.

christopherlam:
> Do you literally mean "locale" or also not limited to specific jurisdictions ?

locale :) because the tax-report-options gives non-deterministic outputs.

fellen:
> Thank you @fellen - would changing the magic string to `*EUGOODS*` and `*EUVAT*` be better?

How about `*EU-GOODS*` and `*EU-VAT*` for easier understanding.

> I learned from Reverse EC charges, literally from GnuCash GB account structure, and I guess it stuck through.
> 
> https://github.com/Gnucash/gnucash/blob/93a24b62ce54dfc0d42f8409f2a7aaefac2830b1/data/accounts/en_GB/uk-vat.gnucash-xea#L140

Fixed in commit f4379bb.

> I still do not think that using the Tax-Report-Options dialog is suitable for the purposes of this report because I do not wish to limit my report to a specific locale.

Trying some generalization: I assume most governments distinguish between domestic and foreign puchases/sales  in their tax handling. But many have between them also a category for their Regional Economic Community, REC. See https://en.wikipedia.org/wiki/Regional_Economic_Communities. So for a global use `REC-*` might be more appropriate. In this case, you should add translator comments and probably an entry in the glossary.

PR #501, opened 2019-05-03
make python shell startup more understandable

c-holtermann:


jralls:
Thanks.

It occurs to me that those strings should be localized at some point, which in addition to marking them for gettext means that they should switch from including the variable to parameterizing them, e.g.
```
banner = _("Welcome to {0}\n").format(title)
```
instead of
```
banner = "Welcome to " + title + "\n"
```

c-holtermann:
John, thank you very much for that suggestion. I had been thinking about the same thing. I'll try to implement that localization in the python bindings. I have files a bug (enhancement proposal): https://bugs.gnucash.org/show_bug.cgi?id=797228.

jralls:
You don't need a bug report if you're going to do the work, a PR is sufficient.

c-holtermann:
Some thoughts why I like a bugreport in addition to a pull request.
A bug report seems to be more stable to me. A pull request can be accepted and is then gone.
In the workflow I like to have a task description before actually doing the coding. Pull requests seems more focused on the actual coding than bug reports. When there is just an idea or suggestion I think it is useful to file it as an enhancement suggestion which can be discussed before writing code.

jralls:
No, pull requests live as long as Github does. This one is "closed" but we're still using it, and if you go to the listing page you'll see tabs for open and closed. Plus, once a PR is merged it's in the commit history.

If design discussion is needed then the developer's list is a better place for it so that it gets a wider audience and more viewpoints. It's hard to believe that there needs to be much design discussion for localizing the python code.

c-holtermann:
The commit history is a good point. How development talk takes place is a matter of agreements as there are different possibilities. I like the bugtracker. I like to have a todo list where I can strike out things, even for small things. And even if something is a little thing to you to others it might be a big thing. But anyway it its a more generic question and not so much about localizing python.

PR #502, opened 2019-05-04
Fixed division to return negative remainders.

DavPalma:
The division method was returning negative remainders only when the absolute value of the dividend was less than the value of the divisor for all other cases it didn't respect (a/b)*b + a%b = a. Also changed rounding to be based on the sign of the remainder so that numbers between -1 and 0(quotient 0 and remainder negative) are rounded correctly fixing Bug 796949.

### DIFF
--- a/libgnucash/engine/gnc-rational-rounding.hpp
+++ b/libgnucash/engine/gnc-rational-rounding.hpp
@@ -25,6 +25,12 @@
 #include "gnc-numeric.h"
 #include "gnc-int128.hpp"
 
+template <typename T> inline bool
+quotient_is_positive(T dividend, T divisor)
+{
+    return (dividend > 0 && divisor > 0) || (dividend < 0 && divisor < 0);
+}
+
 enum class RoundType
 {
     floor = GNC_HOW_RND_FLOOR,
@@ -72,8 +78,23 @@ round(T num, T den, T rem, RT2T<RoundType::floor>)
 //              << ", and rem " << rem << ".\n";
     if (rem == 0)
         return num;
-    if (num < 0)
-        return num + 1;
+    // floor num==0 that is the quotient of two numbers with opposite signs
+    if (num < 0 || (num == 0 && !quotient_is_positive(rem, den)))
+        return num - 1;
+    return num;
+}
+
+
+template <> inline GncInt128
+round<GncInt128>(GncInt128 num, GncInt128 den, GncInt128 rem,
+                 RT2T<RoundType::floor>)
+{
+//    std::cout << "Rounding to floor  with num " << num << " den " << den
+//              << ", and rem " << rem << ".\n";
+    if (rem == 0)
+        return num;
+    if (num.isNeg())
+        return num - 1;
     return num;
 }
 
@@ -82,7 +103,18 @@ round(T num, T den, T rem, RT2T<RoundType::ceiling>)
 {
     if (rem == 0)
         return num;
-    if (num > 0)
+    if (num > 0 || (num == 0 && quotient_is_positive(rem, den)))
+        return num + 1;
+    return num;
+}
+
+template <> inline GncInt128
+round<GncInt128>(GncInt128 num, GncInt128 den, GncInt128 rem,
+      RT2T<RoundType::ceiling>)
+{
+    if (rem == 0)
+        return num;
+    if (!num.isNeg())
         return num + 1;
     return num;
 }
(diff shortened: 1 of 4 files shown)


libgnucash/engine/gnc-rational-rounding.hpp
```diff
@@ -98,7 +98,7 @@ round(T num, T den, T rem, RT2T<RoundType::promote>)
 {
     if (rem == 0)
         return num;
-    return num + (num < 0 ? -1 : 1);
+    return num + (rem < 0 ? -1 : 1);
```
jralls: Need a comment explaining this. It will seem weird without context.

DavPalma: What I did there was wrong and ended up checking the signs of the numerator and remainder to ascertain whether zero should be rounded up or down.

libgnucash/engine/gnc-rational-rounding.hpp
```diff
@@ -72,8 +74,23 @@ round(T num, T den, T rem, RT2T<RoundType::floor>)
 //              << ", and rem " << rem << ".\n";
     if (rem == 0)
         return num;
-    if (num < 0)
-        return num + 1;
+    // floor num==0 that is the quotient of two numbers with opposite signs
+    if (num < 0 || (num == 0 && (SIGN_NUM(rem) * SIGN_NUM(den) < 0)))
+        return num - 1;
```
DavPalma: I also fixed a mistake in flooring and the corresponding test that was rounding up negatives.

libgnucash/engine/gnc-rational-rounding.hpp
```diff
@@ -25,6 +25,8 @@
 #include "gnc-numeric.h"
 #include "gnc-int128.hpp"
 
+#define SIGN_NUM(number) ((number) == 0 ? 0 : (number) / std::abs(number))
```
jralls: No macros please, this is C++. Use a template.
Did you profile ```number/std::abs(number)``` against ```number <  0 ? -1 :number > 0 ? 1 : 0```?

DavPalma: Sorry, I am not that much comfortable with C++, but I guess it's alright now.

bhardwajs:
```
template<typename T>
constexpr bool sign_number(T number)
{
    static_assert(std::is_arithmetic<T>::value);
    return number == std::abs(number);    
}
```

DavPalma: @bhardwajs I ended up writing a template function that checks if the quotient between the two numbers is positive.

jralls: @bhardwajs Not quite the same. ```SIGN_NUM``` returns -1, 0, or 1 and the usage is ```SIGN_NUM(numerator) * SIGN_NUM(denominator)```. That obviously won't work for a function that returns type bool. Also it's operating on runtime values so it can't be ```constexpr```.

jralls: @DavPalma Looks good.

jralls:
One last thing before I can merge this: You need to rebase away the two merge commits.

libgnucash/engine/gnc-rational-rounding.hpp
bhardwajs: Yes, @jralls , the usage `SIGN_NUM(numerator) * SIGN_NUM(denominator)` would require a numeric return type. Sorry about that!

For reference, `constexpr` can be used for functions operating on runtime values. For example, following code is valid and tested with `g++ -Wall -std=c++11`

```
template<typename T>
constexpr bool sign_number(T number)
{
    static_assert(std::is_arithmetic<T>::value);
    return number == std::abs(number);    
}

int main()
{
    int b;
    std::cin >> b;
    std::cout << sign_number(b) << '\n';
    
    return 0;
}
```

jralls: @bhardwajs Well, sort of. If you don't use the function in a ```constexper``` context then the keyword is ignored. If instead you do
```
#include <iostream>
#include <type_traits>
#include <cmath>

template<typename T>
constexpr bool sign_number(T number)
{
    static_assert(std::is_arithmetic<T>::value);
    return number == std::abs(number);    
}

template<bool v>
struct constN
{
    constN() { std::cout << v << '\n'; }
};

int main()
{
    constN<sign_number(-17)> out;
    return 0;
}
```
So that it gets used as an actual constexpr and with Apple clang 10.0.1 you get
```
/Users/john/testconstexper.cpp:23:12: error: non-type template argument is not a
      constant expression
    constN<sign_number(-17)> out;
           ^~~~~~~~~~~~~~~~
/Users/john/testconstexper.cpp:9:22: note: non-constexpr function 'abs' cannot
      be used in a constant expression
    return number == std::abs(number);
                     ^
/Users/john/testconstexper.cpp:23:12: note: in call to 'sign_number(-17)'
    constN<sign_number(-17)> out;
```
Bottom line: Don't stick ```constexpr``` on template functions unless you're actually going to run the function at compile time, you might be introducing latent bugs.

jralls:
Thanks!

PR #505, opened 2019-05-05
Bug 751290 - Rethinking the invoice ledger view

EvansMike:
Have information if this invoice has been paid
This uses an unused label in the Invoice Information box to
display PAID or UNPAID as appropriate.

### DIFF
--- a/gnucash/gnome/dialog-invoice.c
+++ b/gnucash/gnome/dialog-invoice.c
@@ -161,7 +161,8 @@ struct _invoice_window
     GtkWidget  * posted_date_hbox;
     GtkWidget  * posted_date;
     GtkWidget  * active_check;
-
+    GtkWidget  * paid_label;
+    
     GtkWidget  * owner_box;
     GtkWidget  * owner_label;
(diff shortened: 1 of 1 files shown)


gnucash/gnome/dialog-invoice.c
```diff
@@ -1949,6 +1948,11 @@ gnc_invoice_update_window (InvoiceWindow *iw, GtkWidget *widget)
         gtk_widget_set_sensitive (iw->notes_text, TRUE);
     }
 
+    if(gncInvoiceIsPaid (invoice))
+            gtk_label_set_text(GTK_LABEL(iw->paid_label),  _("PAID"));
```
jralls: Indentation.

EvansMike: Thanks John. Will do.

fellen: It would save the some translation if we would use some `uppercase (_("paid"|"unpaid))` function in this module.
Also a translator comment  about the context is always welcome for single words.

PR #507, opened 2019-05-07
Qif Import Assistant changes

Bob-IT:
Just a heads up before I try and push some bug fixes for the QIF import assistant the main being the first bug that deals with tradable commodities being added as assistant pages and if you have a lot you loose the assistant buttons of the screen so on the commodities assistant page I added a notebook and then each commodity is then added as a notebook page.

The other bugs were relatively straightforward so I think should be OK with some others that should be closeble as a result.

The last commit fixes one issue with the cancel/undo but there is still another one coming from scheme that I have not figured out so may look at it separately.

### DIFF
--- a/gnucash/import-export/qif-imp/assistant-qif-import.c
+++ b/gnucash/import-export/qif-imp/assistant-qif-import.c
@@ -62,10 +62,10 @@
 #include "guile-mappings.h"
 #include <gfec.h>
 
-#define ASSISTANT_QIF_IMPORT_CM_CLASS "assistant-qif-import"
-#define GNC_PREFS_GROUP   "dialogs.import.qif"
-#define GNC_PREF_SHOW_DOC "show-doc"
-#define GNC_PREF_DEFAULT_TRANS_STATUS_CLEARED "default-status-cleared"
+#define ASSISTANT_QIF_IMPORT_CM_CLASS            "assistant-qif-import"
+#define GNC_PREFS_GROUP                          "dialogs.import.qif"
+#define GNC_PREF_SHOW_DOC                        "show-doc"
+#define GNC_PREF_DEFAULT_TRANS_STATUS_CLEARED    "default-status-cleared"
 #define GNC_PREF_DEFAULT_TRANS_STATUS_NOTCLEARED "default-status-notcleared"
 #define GNC_PREF_DEFAULT_TRANS_STATUS_RECONCILED "default-status-reconciled"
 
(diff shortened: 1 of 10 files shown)


Bob-IT:
I will push this on Sunday if no one objects

gnucash/import-export/qif-imp/assistant-qif-import.c
```diff
@@ -94,8 +94,10 @@ enum qif_trans_cols
 {
     QIF_TRANS_COL_INDEX = 0,
     QIF_TRANS_COL_DATE,
+    QIF_TRANS_COL_DATE_INT64,
     QIF_TRANS_COL_DESCRIPTION,
     QIF_TRANS_COL_AMOUNT,
+    QIF_TRANS_COL_AMOUNT_DOUBLE,
```
gjanssens: Perhaps add a developer comment here to explain why you chose a double instead of a gnc_numeric. I suppose (after some thought) this is because you can store a double directly in a list store where you can't store a gnc_numeric directly.

gnucash/import-export/qif-imp/dialog-account-picker.c
```diff
@@ -318,6 +326,40 @@ gnc_ui_qif_account_picker_map_cb(GtkWidget * w, gpointer user_data)
 }
 
 
+/****************************************************************
+ * dialog_response_cb
+ *
+ ****************************************************************/
+static void
+dialog_response_cb (GtkDialog *dialog, gint response_id, gpointer user_data)
```
gjanssens: An error dialog is always disruptive. Can we avoid it ? For example can we disable the Ok button when a placeholder is selected ? Potentially with a status message somewhere indicating why the Ok button was disabled.

gnucash/import-export/qif-imp/assistant-qif-import.c
jralls: If it's just for sorting numerically it should be OK. For that purpose converting to a common denominator (100 or 1000 seem reasonable) is an alternative, not sure which would be faster nor if it would matter.

Bob-IT: I will add a comment to fact it is only used for the sorting, I don't think speed is an issue here.

gnucash/import-export/qif-imp/dialog-account-picker.c
Bob-IT: I just copied what was similar in the OFX importer I think but can have another look at it before pushing.

PR #508, opened 2019-05-13
[dialog-tax-table] implement editing of Tax Table name (WIP)

christopherlam:
Adventures into C and GLADE :-)
This commit aims to implement tax-table name editing.

So far - managed to copy current taxtable's name into textbox, handled OK, but can't get gtk_entry's contents. Clues requested -- why I can't retrieve textbox's contents via `gtk_entry_get_text GTK_ENTRY(textbox)` !

image
image
image

```
This is a development version. It may or may not work.
Report bugs and other problems to gnucash-devel@gnucash.org
You can also lookup and file bug reports at https://bugs.gnucash.org
To find the last stable version, please refer to https://www.gnucash.org/
Found Finance::Quote version 1.47.
* 16:32:51  WARN <GLib-GObject> invalid unclassed pointer in cast to 'GtkEntry'
* 16:32:51 ERROR <Gtk> gtk_entry_get_text: assertion 'GTK_IS_ENTRY (entry)' failed
```
PS I am fully aware `char_dialog` should have a better name and belongs somewhere else, but not sure where. It's a copypasta from `gnc_choose_radio_option_dialog`.

### DIFF
--- a/gnucash/gnome-utils/dialog-tax-table.c
+++ b/gnucash/gnome-utils/dialog-tax-table.c
@@ -60,6 +60,7 @@ enum tax_entry_cols
 };
 
 void tax_table_new_table_cb (GtkButton *button, TaxTableWindow *ttw);
+void tax_table_rename_table_cb (GtkButton *button, TaxTableWindow *ttw);
 void tax_table_delete_table_cb (GtkButton *button, TaxTableWindow *ttw);
 void tax_table_new_entry_cb (GtkButton *button, TaxTableWindow *ttw);
 void tax_table_edit_entry_cb (GtkButton *button, TaxTableWindow *ttw);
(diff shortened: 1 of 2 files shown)


Bob-IT:
Looks to me like you are destroying the dialog before retrieving the value, swap 618 and 620 and change that to get a const gchar that you have defined previously and then return that.

christopherlam:
Thanks, this works! Hopefully will finalise soon.

christopherlam:
Oh and I have *no* idea whether var/function-names, function location, are style-consistent, nor are memories freed appropriately.

Bob-IT:
Some thoughts...
Why not create the dialog in the glade file and you could then get rid of the char_dialog function and do it all in the edit_table_cb function as all you realy will be left with is the building and dialog run.

In your existing code there are some white space formatting issues, line 634 you are passing non translatable strings and the parent variable is NULL. Oh if you want to keep it as is, make the char_dialog fun static then you do not need line 62, I also do not think you need to g_strdup the returning value but for avoiding the set you will need to use g_strcmp0

christopherlam:
> Some thoughts...
> Why not create the dialog in the glade file and you could then get rid of the char_dialog function and do it all in the edit_table_cb function as all you realy will be left with is the building and dialog run.

Good question. For now I'm practically bashing words until the compiler doesn't barf. I'll look into GLADE editor; would need to wait a few days.

> In your existing code there are some white space formatting issues, line 634 you are passing non translatable strings and the parent variable is NULL. Oh if you want to keep it as is, make the char_dialog fun static then you do not need line 62, I also do not think you need to g_strdup the returning value but for avoiding the set you will need to use g_strcmp0

g_strdup was needed because (I think) text is pointing to the gtk_entry_get_text's result; removing it meant the contents was destroyed alongside the dialog.

Feedback very appreciated.

christopherlam:
converted to `static`. remaining issues:
1. after successfully renaming taxtable, the selected taxtable changes unnecessarily. not sure how to pin it down.
2. there's *no* checking for newname is already an existing taxtable name.

gnucash/gnome-utils/dialog-tax-table.c
```diff
@@ -550,6 +551,94 @@ tax_table_new_table_cb (GtkButton *button, TaxTableWindow *ttw)
     new_tax_table_dialog (ttw, TRUE, NULL, NULL);
 }
 
+
+static const char
+*char_dialog (GtkWidget *parent,
```
jralls: How about ```rename_tax_table_dialog```?

jralls:
@christopherlam I don't understand what you mean about the tax table changes unnecessarily.

gnucash/gnome-utils/dialog-tax-table.c
```diff
@@ -625,19 +620,22 @@ const char
...
-                           gncTaxTableGetName (ttw->current_table));
+    oldname = gncTaxTableGetName (ttw->current_table);
+    newname = char_dialog (ttw->dialog, (_("Rename")),
+                           (_("Rename Tax Table")),
+                           NULL, oldname);
 
-    if (!(newname == NULL || *newname == '\0'))
-      {
+    if (!(newname == NULL || *newname == '\0' || *newname == *oldname))
+    {
```
Bob-IT: I would change this to...
    if (newname && *newname != '\0')
    {
        if (g_strcmp0 (oldname, newname) != 0)
            gncTaxTableSetName (ttw->current_table, newname);
    }

Bob-IT:
I think Chris is talking about the underlying tree views, I would imagine the selection is changing after edit, not sure if they are sorted or maybe the model is being refreshed?

christopherlam:
> I think Chris is talking about the underlying tree views, I would imagine the selection is changing after edit, not sure if they are sorted or maybe the model is being refreshed?

Correct; initial state, standard input is selected *=selection

Tax Tables | Tax Table Entries
-----------|------------------
Standard Input(*) | 20% Input
Standard Output |


Click Edit, edit to "New Standard Input". Selection moves to another tax-table

Tax Tables | Tax Table Entries
-----------|------------------
New Standard Input | 20% Output
Standard Output(*) |

christopherlam:
Done. Will not rename if newname matches existing taxtable name. Ideally it would error out and offer chance to retry but I'm not too bothered. Ideally the taxtable selection won't move after rename but I don't know how to fix.

Bob-IT:
I know why the selection moves and will push a fix at some point along with setting up default sorting.

christopherlam:
Thank you @bob-it for assistance. If there are no further comments nor missed g_free I'll merge this soon.

Bob-IT:
Chris, the integer i is not used so remove, button_name seems redundant so might be an idea to remove that also and check your white spaces, some are missing or misplaced like...
function_name (GTK_WIDGET(xyz));
g_free (xyz);
I still think the dialogue could be created in glade and possibly a couple of callbacks to check for duplicate name and presenting a warning message in the dialogue as that may be less disruptive but may be for another time.

christopherlam:
fully agree @Bob-IT -- any remaining limitations above are from my own lack of skill. I think this is now ok?

christopherlam:
merged.

PR #509, opened 2019-05-14
[numeric] Fix constructor from strings in range (0 ,1)

ianchi:
Fix error that caused strings in the form 0.nnn to be
converted to negative numerics.


jralls:
Thanks!

PR #510, opened 2019-05-17
Report cleanup

gjanssens:
This PR does some housekeeping around our reports. Points of note:

- all but the locale-specific reports are now stored in one of two directories: reports/standard or reports/example. The reports in business-reports and utility-reports have been integrated there as well.
- all reports (again except for the locale-specific ones and the 'new' dashboard report) are now loaded using the same load function. This function loops a preset list of directories and loads each scm file it finds in there. This is how it was done for standard reports before, it has now been extended to most reports in reports/standard and reports/example.
- the "Welcome to gnucash" report has been reworked into a dashboard report as proposed in [bug 773199](https://bugs.gnucash.org/show_bug.cgi?id=773199). The "welcome to gnucash" report was loaded rather unusually via a snippet of code near the end of report-gnome.scm. I have chosen not to change this as it serves as an example of how scheme code can add items to the gnucash menus. Hence this report is not loaded via the common load mechanism described above.
- the new dashboard report is added to a new "Multicolumn" sub menu, together with the "Custom Multicolumn Report" as suggested in [bug 773200](https://bugs.gnucash.org/show_bug.cgi?id=773200)
- the reports as mentioned in [bug773198](https://bugs.gnucash.org/show_bug.cgi?id=773198) have been moved to an Example sub menu which replaces the Sample & Custom menu. As these reports are mostly useful as examples for someone wanting to write their own this menu is now hidden by default. It can be made visible by running gnucash with the "--extra" command line option.
- my intention is to eventually include the locale-specific reports into this scheme as well. However those require a larger redesign than I now have time for so that part is deferred.

Todo's:
- ~~I wasn't too sure what to do with "aging.scm". For the time being it's parked in the reports directory. It's really a report engine, much like @christopherlam 's trep-engine. Should I bring those together in their own subdirectory (reports/engines) ?~~ Will be deferred until a future cleanup of reports and report-system.
- ~~the reports/* directories and files are currently installed under gnucash/report as there's already a gnucash/reports directory curiously holding a couple of test scripts totally unrelated to reporting... I'll look into moving those and the move the reports directory next to report to flatten the hierarchy a bit.~~ The test scripts' compiled objects have all been moved to ccache/2.2/tests. Other than that I won't move the reports directory up a level. For now it makes more sense to keep everything report related in the report directory.
- I'm tempted to replace the explicit report lists in gnucash/report/reports/CMakeLists.txt with file globs, much like we do for our pixmaps in data/pixmaps. Are there compelling reasons not do to this when the contents of the reports sub directories is strictly defined ?

Are there other issues/remarks that require fixing before merging this into maint ?


gnucash/report/report-system/html-utilities.scm
```diff
@@ -51,7 +51,67 @@
 (define (gnc:price-anchor-text price)
   (gnc-build-url URL-TYPE-PRICE
 		      (string-append "price-guid=" (gncPriceGetGUID price))
-		      ""))
+		      ""))(define (guid-ref idstr type guid)
```
jralls: Needs a newline.

libgnucash/engine/Account.cpp
```diff
@@ -248,6 +249,78 @@ mark_account (Account *acc)
     qof_instance_set_dirty(&acc->inst);
 }
 
+Account*
+gnc_account_from_tuple (const AccountTuple& t, QofBook* book)
```
gjanssens: I suppose this slipped in  by accident ?

gnucash/report/report-system/html-utilities.scm
gjanssens: Thanks. Added a fixup commit.

gjanssens:
Note I have added a commit instead of fixing the original commit and force-pushing. I think that's easier during review. I'll do a final commit cleanup round before merging.

gjanssens:
@jralls  Your relocation of the test-*.go files works well. Thanks!

christopherlam:
I will refrain from tinkering reports until this I finalised.

I recall some were using average-balance as a periodic income chart. Perhaps we should inform the users before banishing it behind a switch? I think renaming welcome report to dashboard is nice.

I'm glad the test-extras mess will be fixed. I cannot wait to rm collectors.scm too.

Re:trep-engine & aging vs rest, up to you; I find the distinction between report engine and report helper files very gray indeed.....
There's HTML generators, charting engine, pricing calculators, etc, a bit difficult to delineate between engine and utilities. If I have time I'll try create an overall dependency map.

gjanssens:
> I will refrain from tinkering reports until this I finalised.
> 
> I recall some were using average-balance as a periodic balance chart. Perhaps we should inform the users before banishing it behind a switch?

Good point. I have just sent a message to the list.

> I think renaming welcome report to dashboard is nice.
> 
> I'm glad the test-extras mess will be fixed. I cannot wait to rm collectors.scm too.

I'm not sure what had to be fixed there ? The only thing that changed is the file name. None of the contents have.

> Re:trep-engine & aging vs rest, up to you; I find the distinction between report engine and report helper files very gray indeed.....
> There's HTML generators, charting engine, pricing calculators, etc, a bit difficult to delineate between engine and utilities. If I have time I'll try create an overall dependency map.

Agreed. I'll probably leave that for a future cleanup round. reports and report-system can do with a bit of a restructuring as well.

gjanssens:
I have pushed a few more fixes for the test scripts' object files.

Other than that, any thoughts on this:

> I'm tempted to replace the explicit report lists in gnucash/report/reports/CMakeLists.txt with file globs, much like we do for our pixmaps in data/pixmaps. Are there compelling reasons not do to this when the contents of the reports sub directories is strictly defined ?

gjanssens:
Hmm, travis fails the ubuntu 14.04 test:
https://travis-ci.org/gjanssens/gnucash/jobs/534157193
I'll look at it somewhere next week.

libgnucash/engine/Account.cpp
jralls: Yup, sorry, it belongs to another branch. I thought I had a clean tree.

jralls:
```report/reports/data``` seems to be just the eguile reports. Maybe it should be ```report/reports/equile``` instead?

**Comment by gjanssens** [comment]
> `report/reports/data` seems to be just the eguile reports. Maybe it should be `report/reports/equile` instead?

To be exact it's the template and css files for the eguile reports. Each eguile report consists of 3 files: the report file (stored with the other reports), a template file and a css style sheet.
As for now only the eguile reports have templates and css files we could use `report/reports/eguile` instead as well. I chose data to be a bit more generic thinking it the future there may be more of these. For example if we do switch to css based styling, or if we go for report templates based on mustache,...
Generally I thought this directory could be a the location for all report related support files that don't need compilation (so that aren't scheme files we have to run through guild). 

If you consider that too far in the future or too uncertain, I am fine with calling it eguile as well.

**Comment by jralls** [comment]
OK, though ```data``` is perhaps a bit too generic. Maybe call it ```support``` add a one-line ```README``` for explanation?

**Comment by gjanssens** [comment]
Will do.

**Comment by christopherlam** [comment]
My opinion regarding hiding reports -- I would think it'll be fairly safe to hide any underused reports via `--extras`. I'm not sure how exactly you were planning to code this -- there is, within `(gnc:define-report)` mechanism, the optional report property `'in-menu?` which controls the visibility of a report in menus. I believe any saved-reports or unit tests based on hidden reports will still work. If there's outcry it wouldn't be difficult to reverse a decision.

Example from `average-balance.scm` definition, assuming `SCM_BOOL gnc_enable_extras_argument` exists.

```
(gnc:define-report
 'version 1
 'name reportname
 'in-menu? (gnc-enable-extras-argument)
 'report-guid "d5adcc61c62e4b8684dd8907448d7900"
 'menu-path (list gnc:menuname-example)
 'options-generator options-generator
 'renderer renderer)
```

File: gnucash/report/report-gnome/report-gnome.scm [hunk 3]
```diff
@@ -102,7 +103,8 @@
   (gnc-add-scm-extension income-expense-menu)
   (gnc-add-scm-extension asset-liability-menu)
   (gnc-add-scm-extension budget-menu)
-  (gnc-add-scm-extension utility-menu)
+  (if (gnc-prefs-is-extra-enabled)
+    (gnc-add-scm-extension example-menu))
```
gjanssens: @christopherlam This line hides the complete "Examples" menu unless users set the --extra command line.

gjanssens:
> My opinion regarding hiding reports -- I would think it'll be fairly safe to hide any underused reports via `--extras`. I'm not sure how exactly you were planning to code this

I'm not planning it, it's done :)
See the comment I have added inline to show how I implemented it. The examples menu is fully hidden. Any report that should be visible in all circumstances has to be moved back to another menu.

 -- there is, within `(gnc:define-report)` mechanism, the optional report property `'in-menu?` which controls the visibility of a report in menus. I believe any saved-reports or unit tests based on hidden reports will still work. If there's outcry it wouldn't be difficult to reverse a decision.

Indeed.

fellen:
IMHO
1. --extras is the wrong flag. It is until now used for experimental stuff. While the examples are useful for beginners, the register2 stuff is definitely not. How about adding an `--examples` flag?
 BTW: The examples have been a strong point for me to choose gnucash - whithout them I had probably choosen a different program.
1. On the user list I see advanced english natives, who want to get rid of the examples and non-english users, who want to keep them exposed. This probably related to the lack of up to date translated documentation. How about 

> if LANG=C hide examples, else not

or make it a preferemce?

christopherlam:
>    On the user list I see advanced english natives, who want to get rid of the examples and non-english users, who want to keep them exposed. This probably related to the lack of up to date translated documentation. How about if LANG=C hide examples, else not; or make it a preferemce?

This looks like a bad hack! I think we need to rid the code base from unexpected behaviour like this.

I think if documentation is lacking then the documentation should be improved. My vision for the code base is that all reports will use a standardised API, with a common core using clean scheme calls. It would seem anything lesser will just drag down progress.

I can probably rewrite all existing reports much more cleanly than the current set, which will ease deprecating old exported functions, but I cannot promise 100% backward compatibility with old saved-reports.

gjanssens:
I agree with @christopherlam here that using locale as poor man's substitute for a preference is definitely not a good idea.

And I'm not in favour of adding an extra command line switch for the example reports. That's an additional code path to test for little gain. I don't feel that much difference between exposing example reports or register2 in the same switch. Both are to be exposed only for testing or development purposes.

Note that adding them to --extra was a convenience consideration. I also seriously considered simply dropping the reports completely from the menus. The reasoning behind this is these reports serve as examples of what is possible with the report system. They are no longer considered standard/supported reports (again the fact they're still built and tested is a convenience service so at least the examples we offer are proper examples).
Our canonical way to add unsupported reports is to add them in your GNC_CONFIG_HOME, write a load line in user-config.scm and off you go.
However I thought for our default examples it could save a few configuration steps and allow users to evaluate them by simply using the --extra command line switch.

To state this explicitly: no user should ever permanently add the --extra switch to their gnucash invocation! They are examples. If they offer the functionality a user really wants as is, that user should turn it into a custom report and use the custom report procedures to get permanent access to it.

Putting it differently the --extra switch enables a number of playground features in gnucash. And to my mind both the example reports and Register2 fit that category.

gjanssens:
Hmm force-pushing a PR after rebasing it onto another branch is apparently not a good idea. I'll close this PR and create a new one.

PR #511, opened 2019-05-24
Rptcleanup

gjanssens:
This PR started off as a set of changes to clean up the file locations of our reports and the build system around (see PR #510 ). While that PR was against maint, most of these changes have instead been merged into master.

One commit needs more discussion. This has been split of into this PR.

So this PR deals only with:
- the reports as mentioned in [bug773198](https://bugs.gnucash.org/show_bug.cgi?id=773198) have been moved to an Example sub menu which replaces the Sample & Custom menu. As these reports are mostly useful as examples for someone wanting to write their own this menu is now hidden by default. It can be made visible by running gnucash with the "--extra" command line option.


gjanssens:
Based on comments on PR #510 there is one controversial commit in this PR: the one hiding away a couple of rarely used reports.
Are there objections to merge the remainder of these commits while we continue the discussion on this controversial bit ? That would unblock further work on the reports.

Note that compared to PR #510 I have now also squashed the fixup commits to clean up the commit history. Each commit should build cleanly and pass all tests in itself.

gjanssens:
I went ahead an pushed everything except the commit wrt to moving/hiding the example reports.

jralls:
FWIW the controversy isn't over moving the reports because that doesn't really affect users. The controversy is over putting the four reports into a separate menu that's hidden unless the user runs GnuCash with the ```--extra``` command-line parameter.

gjanssens:
Right. So I have pushed that part of the PR to master. The reports have been moved to the example menu, which for now is permanently visible.
Although I've force pushed the bit to hide the menu unless the --extra option is set, I think I'll just close the PR instead. I currently don't care *that* much about this detail. I may think of a more elegant way to deal with this in a future redesign.

PR #512, opened 2019-05-25
Budget: more fixes

ianchi:
Adds some more fixes to "Budget Reports":
- add multicurrency support
- remove unused option "rollup budget" (it is always used)
- fix color formatting in budget column

Adds a fix to the budget register:
- Respect read-only state of a book and date threadshold for editing budget values

### DIFF
--- a/gnucash/report/standard-reports/budget.scm
+++ b/gnucash/report/standard-reports/budget.scm
@@ -60,8 +60,6 @@
 (define opthelp-show-difference (N_ "Display the difference as budget - actual."))
 (define optname-show-totalcol (N_ "Show Column with Totals"))
 (define opthelp-show-totalcol (N_ "Display a column with the row totals."))
-(define optname-rollup-budget (N_ "Roll up budget amounts to parent"))
-(define opthelp-rollup-budget (N_ "If parent account does not have its own budget value, use the sum of the child account budget values."))
 (define optname-show-zb-accounts (N_ "Include accounts with zero total balances and budget values"))
 (define opthelp-show-zb-accounts (N_ "Include accounts with zero total (recursive) balances and budget values in this report."))
 
@@ -264,10 +262,6 @@
       "s4" opthelp-show-totalcol #f))
     (add-option
      (gnc:make-simple-boolean-option
-      gnc:pagename-display optname-rollup-budget
-      "s4" opthelp-rollup-budget #f))
-    (add-option
-     (gnc:make-simple-boolean-option
       gnc:pagename-display optname-show-zb-accounts
       "s5" opthelp-show-zb-accounts #t))
 
@@ -292,7 +286,6 @@
          (show-budget? (get-val params 'show-budget))
          (show-diff? (get-val params 'show-difference))
          (show-totalcol? (get-val params 'show-totalcol))
-         (rollup-budget? (get-val params 'rollup-budget))
          (use-ranges? (get-val params 'use-ranges))
          (num-rows (gnc:html-acct-table-num-rows acct-table))
          (numcolumns (gnc:html-table-num-columns html-table))
@@ -347,11 +340,9 @@
     ;;   colnum - starting column number
     ;;   budget - budget to use
     ;;   acct - account being displayed
-    ;;   rollup-budget? - rollup budget values for account children
-    ;;                    if account budget not set
     ;;   exchange-fn - exchange function (not used)
     (define (gnc:html-table-add-budget-line!
-             html-table rownum colnum budget acct rollup-budget?
+             html-table rownum colnum budget acct
              column-list exchange-fn)
       (let* ((comm (xaccAccountGetCommodity acct))
              (reverse-balance? (gnc-reverse-balance acct))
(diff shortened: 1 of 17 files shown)


gnucash/report/standard-reports/budget.scm
```diff
@@ -262,10 +260,6 @@
      (gnc:make-simple-boolean-option
       gnc:pagename-display optname-show-totalcol
       "s4" opthelp-show-totalcol #f))
-    (add-option
-     (gnc:make-simple-boolean-option
```
christopherlam: Instead of deleting option I think it's safer to change this to `(gnc:make-internal-option)` so that saved-reports will still load. And add comment to explain this option was never used.

ianchi: I've made the change, but I'm not sure if it is the right way. I'm not familiar at all with Scheme and all the internals of reports

christopherlam: Your scheme skills are not that bad :) you're right "__" in section will hide it, but we want to use `(gnc:make-internal-option gnc:pagename-display "Roll up budget amounts to parent" #f)` instead because options are found via section/name strings, and saved-reports will try to restore options by searching them. (I'd suggest `#f` because in scheme it is conventionally used for all NULL/unused/missing data). So it should be something like:

```scheme
 (add-option
  (gnc:make-internal-option
   gnc:pagename-display "Roll up budget amounts to parent" #f))
```

ianchi:
Regarding the "nearest-price" conversion, I understand that it is restricted to just that strategy, but I wanted the buget's values shown in the report to coincide with the ones in the budget register page, which uses only "nearest-price".

ianchi:
Added a fix for inconsistent button behavior 
acting on the whole budget and not on selected rows.

ianchi:
Hi @christopherlam / @jralls ,
please tell me if there is anything else pending.

Thnks

gnucash/report/standard-reports/budget.scm
christopherlam: Well I'll need to put foot from mouth.
It looks like removing an option is actually harmless.
Apologies for misinformation.

ianchi: No problem.
So now I reverted back, and completely removed the option.
This leaves a cleaner code

ianchi:
Added new functionality to include notes to budgets values.
This fixes issue
Bug 693180 - Add notes to budgeting values

jralls:
This goes *way* beyond fixes. Anything involving backend schema changes is a major enhancement and goes to master, never maint. 

There's also a problem that you're persisting a new field in the SQL backend and not in the XML backend. That's also a non-starter.

ianchi:
OK, I'll split the last set of commits to a different branch, and leave the first fixes for maint, if it is OK for you
Regarding the persistence in XML, it doesn't need any change as it is  actually persisted with the slots.
I have already checked that.

jralls:
Separate the actual bugs from the enhancements, see the second "Importance" field on each bug for a first cut at telling them apart, but submitters don't always mark enhancements so also consider whether each bug describes something wrong with existing functionality (that's a bug) or some new behavior (that's an enhancement). Each enhancement should be a separate PR and we'll decide on a case basis whether it should go to maint or master.

If the notes are going to be in slots then it's slots for both XML and SQL, no need to add anything to either backend.

ianchi:
@jralls another question. The notes functionality will work exactly the same if we simple drop the commit with the schema change.
Changing the backend is not really necessary, I only did it to have a more consistent schema.
Would you agree to add the “notes” functionality to maint if I do that?

Additionally in the master branch I could send a fix to the budget’s sql backend, as right now it has a weird behavior, and everything is saved twice, once in budget_amounts and once in slots. With the later being the one actually used.
(This is why the commit with the schema change can be dropped).
Another thing that in my opinion also needs rethinking is the way editing works: now editing in the view is slow, as each cell change triggers a complete delete of all the records (for all accounts and periods) and insertion of all the values

christopherlam:
> Another thing that in my opinion also needs rethinking is the way editing works: now editing in the view is slow, as each cell change triggers a complete delete of all the records (for all accounts and periods) and insertion of all the values

This is probably the cause for major slowdowns on Windows in 2.6 series. Not so much on 3.x. I'd think this is a good bugfix if it can be done safely.

jralls:
@ianchi I, along with the other core devs, will *consider* adding such an enhancement to maint, but remember that we're planning to release 4.0 this winter so it's not much of a delay to put an enhancement in master instead.

Saving the same data in two places is indeed dumb, and if the SQL backend is doing it and the XML one not that's rather a poor design. 

The last issue sounds like a major design flaw, which means a lot of work to fix it. Depending on how much that might trip another (debatable) reason to put it in master, that the magnitude of the change might affect introduce enough new bugs to make the supposedly stable series not so stable.

ianchi:
> Separate the actual bugs from the enhancements
...
Each enhancement should be a separate PR and we'll decide on a case basis whether it should go to maint or master.

I've split the commits and send PR #513 with only "smaller" fixes
After they are merged I'll send a reduced version of "notes" without be changes, for your consideration. (it will most probably have merge conflicts if I send two parallel PR)

Afterwards in the master branch we discuss on the other more complex issues

PR #513, opened 2019-06-02
Budget Report fixes

ianchi:
Adds some more fixes to "Budget Reports":

add multicurrency support
remove unused option "rollup budget" (it is always used)
fix color formatting in budget column
Adds a fix to the budget register:

Respect read-only state of a book and date threadshold for editing budget values

This replaces #512

### DIFF
--- a/gnucash/report/report-system/report-utilities.scm
+++ b/gnucash/report/report-system/report-utilities.scm
@@ -981,14 +981,22 @@ flawed. see report-utilities.scm. please update reports.")
 ;;   budget - budget to use
 ;;   children - list of children
 ;;   period - budget period to use
+;;   currency - currency to use to accumulate the balances
 ;;
 ;; Return value:
 ;;   budget value to use for account for specified period.
-(define (budget-account-sum budget children period)
-  (apply + (map
-            (lambda (child)
-              (gnc:get-account-period-rolledup-budget-value budget child period))
-            children)))
+(define (budget-account-sum budget children period currency)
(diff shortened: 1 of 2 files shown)


jralls:
[577967](https://bugs.gnucash.org/show_bug.cgi?id=577967) is an enhancement request and a significant UI change, so that one needs to go to ```master```.

gnucash/gnome/gnc-budget-view.c
```diff
@@ -1230,7 +1230,12 @@ gbv_refresh_col_titles(GncBudgetView *view)
     gchar title[MAX_DATE_LENGTH];
     GList *col_list;
     gint i;
+    gboolean editable, read_only;
+    GtkCellRenderer *renderer;
+    GDate *readonly_threshold =
+        qof_book_get_autoreadonly_gdate(gnc_get_current_book());
```
jralls: Do you really want to apply the read-only threshold to budgets? It's intended to apply only to transactions so that you can't create transactions more than n days in the past nor modify transactions more than n days old. Budgets aren't part of the accounting record and it would seem that users might want retrospective budget changes even when they don't want to allow transaction changes.

ianchi:
@jralls  I only left the reports fixes in this PR.
For the other more complex and "polemic" ones I'll be sending individual PR so we discuss on each of them.

gnucash/gnome/gnc-budget-view.c
ianchi: Will leave this for another PR

jralls:
Thanks!

PR #514, opened 2019-06-03
Adds Notes to budgets

ianchi:
Functionality to add "notes" to budget values. This solves:
`Bug 693180 - Add notes to budgeting values`

@jralls I've dropped the commit with BE changes, so this version stores notes in slots.

I leave all the BE issues for a future PR to the master branch.

### DIFF
--- a/gnucash/gnome-utils/gnc-cell-renderer-text-flag.c
+++ b/gnucash/gnome-utils/gnc-cell-renderer-text-flag.c
@@ -1,0 +1,238 @@
+/**
+ * gnc-cell-renderer-text-flag.c -- text cell renderer with flag.
+ *
+ * Copyright (C) 2019 Adrian Panella <ianchi74@outlook.com>
+ * All rights reserved.
+ **/
+
+/* GnuCash is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU Library General Public License as
+ * published by the Free Software Foundation; either version 2 of
+ * the License, or (at your option) any later version.
+ *
+ * Gnucash is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Library General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, contact:
+ *
+ * Free Software Foundation           Voice:  +1-617-542-5942
+ * 51 Franklin Street, Fifth Floor    Fax:    +1-617-542-2652
+ * Boston, MA  02110-1301,  USA       gnu@gnu.org
+ */
+
+#include <config.h>
+#include <gtk/gtk.h>
+
+#include "gnc-cell-renderer-text-flag.h"
+
+/**
+ * @Title: GncCellRendererTextFlag
+ *
+ * A #GncCellRendererTextFlag extends the GtkCellRendererText
+ * adding the avility to show a color triangle on the top right corner
+ * to flag the cell.
+ */
+
+
+static void gnc_cell_renderer_text_flag_get_property(GObject *object,
+                                                     guint param_id,
+                                                     GValue *value,
+                                                     GParamSpec *pspec);
+static void gnc_cell_renderer_text_flag_set_property(GObject *object,
+                                                     guint param_id,
+                                                     const GValue *value,
+                                                     GParamSpec *pspec);
+
+static void gnc_cell_renderer_text_flag_render(
+    GtkCellRenderer *cell, cairo_t *cr, GtkWidget *widget,
+    const GdkRectangle *background_area, const GdkRectangle *cell_area,
+    GtkCellRendererState flags);
+
+enum
+{
+    PROP_0,
+
+    PROP_FLAG_SIZE,
+    PROP_FLAG_COLOR,
+    PROP_FLAG_COLOR_RGBA,
+    PROP_FLAGGED,
+
+    LAST_PROP
+};
+
+G_DEFINE_TYPE_WITH_PRIVATE(GncCellRendererTextFlag,
+                           gnc_cell_renderer_text_flag,
+                           GTK_TYPE_CELL_RENDERER_TEXT)
+
+static void
+gnc_cell_renderer_text_flag_init(GncCellRendererTextFlag *celltext)
+{
+    GncCellRendererTextFlagPrivate *priv;
+    GtkCellRendererText *cell = GTK_CELL_RENDERER_TEXT(celltext);
+
+    celltext->priv =
+        gnc_cell_renderer_text_flag_get_instance_private(celltext);
+    priv = celltext->priv;
+
+    priv->size = 8;
+    gdk_rgba_parse(&priv->color, "red");
+    priv->flagged = FALSE;
+}
+
+static void
+gnc_cell_renderer_text_flag_class_init(GncCellRendererTextFlagClass *class)
+{
+    GObjectClass *object_class       = G_OBJECT_CLASS(class);
+    GtkCellRendererClass *cell_class = GTK_CELL_RENDERER_CLASS(class);
+
+    object_class->get_property = gnc_cell_renderer_text_flag_get_property;
+    object_class->set_property = gnc_cell_renderer_text_flag_set_property;
+
+    cell_class->render = gnc_cell_renderer_text_flag_render;
+
+    g_object_class_install_property(
+        object_class, PROP_FLAG_SIZE,
+        g_param_spec_int("flag-size", "Flag size", "Flag size", 0, 50,
+                         8, G_PARAM_READWRITE));
+
+    g_object_class_install_property(
+        object_class, PROP_FLAG_COLOR,
+        g_param_spec_string("flag-color", "Flag color name",
+                            "Flag color as a string", "red",
+                            G_PARAM_WRITABLE));
+
+    g_object_class_install_property(
+        object_class, PROP_FLAG_COLOR_RGBA,
+        g_param_spec_boxed("flag-color-rgba", "Flag color as RGBA",
+                           "Flag color as a GdkRGBA", GDK_TYPE_RGBA,
+                           G_PARAM_READWRITE));
+
+    g_object_class_install_property(
+        object_class, PROP_FLAGGED,
+        g_param_spec_boolean("flagged", "Flag set",
+                             "Flag indicator is set", FALSE,
+                             G_PARAM_READWRITE));
+}
+
+static void
+gnc_cell_renderer_text_flag_get_property(GObject *object, guint param_id,
+                                         GValue *value, GParamSpec *pspec)
+{
+    GncCellRendererTextFlag *celltext    = GNC_CELL_RENDERER_TEXT_FLAG(object);
+    GncCellRendererTextFlagPrivate *priv = celltext->priv;
+
+    switch (param_id)
+    {
+    case PROP_FLAGGED:
+        g_value_set_boolean(value, priv->flagged);
+        break;
+
+    case PROP_FLAG_SIZE:
+        g_value_set_int(value, priv->size);
+        break;
+
+    case PROP_FLAG_COLOR_RGBA:
+      g_value_set_boxed (value, &priv->color);
+      break;
+
+    default:
+        G_OBJECT_WARN_INVALID_PROPERTY_ID(object, param_id, pspec);
+        break;
+    }
+}
+
+static void
+gnc_cell_renderer_text_flag_set_property(GObject *object, guint param_id,
+                                         const GValue *value,
+                                         GParamSpec *pspec)
+{
+    GncCellRendererTextFlag *celltext    = GNC_CELL_RENDERER_TEXT_FLAG(object);
+    GncCellRendererTextFlagPrivate *priv = celltext->priv;
+    switch (param_id)
+    {
+    case PROP_FLAG_COLOR:
+    {
+        GdkRGBA rgba;
+
+        if (!g_value_get_string(value))
+            break;
+        else if (gdk_rgba_parse(&rgba, g_value_get_string(value)))
+            priv->color = rgba;
+        else
+            g_warning("Don't know color '%s'", g_value_get_string(value));
+    }
+    break;
+
+    case PROP_FLAG_COLOR_RGBA:
+    {
+        GdkRGBA *rgba;
+
+        rgba = g_value_get_boxed(value);
+        if (rgba) 
+            priv->color = *rgba;
+    }
+    break;
+
+    case PROP_FLAGGED:
+        priv->flagged = g_value_get_boolean(value);
+        break;
+
+    case PROP_FLAG_SIZE:
+        priv->size = g_value_get_int(value);
+        break;
+
+    default:
+        G_OBJECT_WARN_INVALID_PROPERTY_ID(object, param_id, pspec);
+        break;
+    }
+}
+
+/**
+ * gnc_cell_renderer_text_flag_new:
+ *
+ * Creates a new #GtkCellRendererTextFlag. 
+ * It is a standard GtkCellRendererText extended to optionally show a
+ * coloured triangle as a flag in the top right corner
+ *
+ * Returns: the new cell renderer
+ **/
+GtkCellRenderer *
+gnc_cell_renderer_text_flag_new(void)
+{
+    return g_object_new(GNC_TYPE_CELL_RENDERER_TEXT_FLAG, NULL);
+}
+
+static void
+gnc_cell_renderer_text_flag_render(GtkCellRenderer *cell, cairo_t *cr,
+                                   GtkWidget *widget,
+                                   const GdkRectangle *background_area,
+                                   const GdkRectangle *cell_area,
+                                   GtkCellRendererState flags)
+
+{
+    GncCellRendererTextFlag *celltext    = GNC_CELL_RENDERER_TEXT_FLAG(cell);
+    GncCellRendererTextFlagPrivate *priv = celltext->priv;
+
+    // call the parent renderer to do the standard drawing
+    GTK_CELL_RENDERER_CLASS(gnc_cell_renderer_text_flag_parent_class)
+        ->render(cell, cr, widget, background_area, cell_area, flags);
+
+    // add the flag (triangle in the top right corner)
+    if (priv->flagged) 
+    {
+        guint size = MIN(MIN(background_area->height, priv->size),
+                         background_area->width);
+        double x   = background_area->x + background_area->width - size;
+        double y   = background_area->y;
+
+        cairo_move_to(cr, x, y);
+        cairo_rel_line_to(cr, size, 0);
+        cairo_rel_line_to(cr, 0, size);
+        cairo_close_path(cr);
+        gdk_cairo_set_source_rgba(cr, &priv->color);
+        cairo_fill(cr);
+    }
+}
--- a/gnucash/gnome-utils/gnc-cell-renderer-text-flag.c
+++ b/gnucash/gnome-utils/gnc-cell-renderer-text-flag.c
@@ -59,6 +59,8 @@ enum
     PROP_FLAG_COLOR,
     PROP_FLAG_COLOR_RGBA,
     PROP_FLAGGED,
+    PROP_FLAG_COLOR_SELECTED,
+    PROP_FLAG_COLOR_RGBA_SELECTED,
 
     LAST_PROP
 };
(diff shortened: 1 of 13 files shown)


jralls:
@ianchi It's still a candidate for master because it's a new feature. The migration from slots to the primary data structures is a long-term goal, and we'll want to do it all at once so that we don't have a string of storage compatibility changes. That's not on the table for GnuCash 4.

ianchi:
Pity that it cannot go to maint, as now it is not disruptive to any current behavior. (I'd like to have a win executable with this, but have only a linux dev environment)

Do you need me to rebase onto master and send a new PR?
I understood about the BE stuff, and left it out of this PR.

jralls:
No need to do anything just yet. master is nearly identical to maint and since you're the only one playing in the budget code it will be identical after the next maint->master merge.

gnucash/gnome/gnc-budget-view.c
```diff
@@ -984,6 +985,18 @@ budget_col_source(Account *account, GtkTreeViewColumn *col,
...
+
+    note = gnc_budget_get_account_period_note(budget, account, period_num);
+
+    if (note == NULL)
+    {
+        g_object_set(cell, "background", NULL, NULL);
+    }
+    else
+    {
+        g_object_set(cell, "background", "beige", NULL);
```
gjanssens: Don't hardcode colours. This doesn't work well with custom themes. Instead query the theme for a proper colour based on the function of the color.

If you can't find a suitable colour in the theme, you should provide it in our own css files so at least a user can override it. This will probably mean you'll have to set a custom css class for each cell which is different for cells with note or without and then have a css snippet defining a background color based on this css class.
Note our css files cater for both light or dark themes.

gnucash/gtkbuilder/gnc-plugin-page-budget.glade
```diff
@@ -805,4 +817,119 @@
       <action-widget response="-7">close_button</action-widget>
     </action-widgets>
   </object>
+  <object class="GtkDialog" id="budget_note_dialog">
+    <property name="can_focus">False</property>
+    <property name="border_width">5</property>
+    <property name="title" translatable="yes">Budget notes</property>
```
gjanssens: Small nitpick: I believe in English titles each word is typically capitalized, so this should become `Budget Notes`.

gnucash/gnome-utils/gnc-cell-renderer-text-flag.c
```diff
@@ -0,0 +1,222 @@
+/**
+ * gnc-cell-renderer-text-flag.h -- text cell renderer with flag.
+ *
+ * Note that this widget will track changes in the account tree, and update
+ * itself accordingly.  If an account with the same name exists in the
+ * freshly-retrieved account list, the widget will re-select that account.
```
gjanssens: As far as I understand from reading the code, there is nothing in here to actually track accounts. It merely has a property "flagged" which can be set by the caller.
Please move this comment elsewhere that actually implements this.

gnucash/gnome-utils/gnc-cell-renderer-text-flag.h
```diff
@@ -0,0 +1,68 @@
+/**
+ * gnc-cell-renderer-text-flag.h -- text cell renderer with flag.
+ *
+ * Note that this widget will track changes in the account tree, and update
+ * itself accordingly.  If an account with the same name exists in the
+ * freshly-retrieved account list, the widget will re-select that account.
```
gjanssens: Same for this comment.

gnucash/gnome-utils/gnc-cell-renderer-text-flag.c
```diff
@@ -0,0 +1,222 @@
...
+
+    g_object_class_install_property(
+        object_class, PROP_FLAG_SIZE,
+        g_param_spec_int("flag-size", "Flag size", "Flag size", 0, 50,
+                         8, G_PARAM_READWRITE));
+
+    g_object_class_install_property(
+        object_class, PROP_FLAG_COLOR,
+        g_param_spec_string("flag-color", "Flag color name",
+                            "Flag color as a string", "red",
```
gjanssens: While this will replace the hardcoded "beige" colour from a previous commit, the basic idea remains the same: don't hardcode the colour. Make the colour a css/styling property of this widget. That way users can override it via css.

gnucash/gnome/gnc-budget-view.c
```diff
@@ -987,15 +988,7 @@ budget_col_source(Account *account, GtkTreeViewColumn *col,
...
-
-    if (note == NULL)
-    {
-        g_object_set(cell, "background", NULL, NULL);
-    }
-    else
-    {
-        g_object_set(cell, "background", "beige", NULL);
-    }
+    g_object_set(cell, "flagged", note != NULL, NULL);
```
gjanssens: As the flagged text entry widget replaces the background color changes from an earlier commit, consider squashing them together.

gnucash/gnome-utils/gnc-cell-renderer-text-flag.c
ianchi: Sorry, it was a paste error. I copied the copyright notice from another file and left that part unchanged.

gnucash/gnome-utils/gnc-cell-renderer-text-flag.c
ianchi: I leave this part with a default "red", and configure in the view from css

gnucash/gnome/gnc-budget-view.c
ianchi: OK, added a css styling.
Changed to "blue" so it is visible also on rows highlighted with selection.

ianchi:
@gjanssens I pushed an updated version addressing your comments

gnucash/gnome/gnc-budget-view.c
```diff
@@ -913,6 +914,15 @@ get_negative_color (void)
...
 
+static void
+get_flag_color (GdkRGBA *color)
+{
+    GtkWidget *label = gtk_label_new ("Color");
+    GtkStyleContext *context = gtk_widget_get_style_context (GTK_WIDGET(label));
+    gtk_style_context_add_class (context, "flag-color");
+    gtk_style_context_get_color (context, GTK_STATE_FLAG_NORMAL, color);
+    gtk_widget_destroy(label);
+}
```
gjanssens: While this works for the most basic scenario determining the flag_color from an arbitrary and temporary widget the user never interacts with is fragile.
You should query the style context of the exact widget that will be using these style properties to render in order to allow the user more fine-grained control over styling.

The way it's implemented now only it's not possible for a user to style as follows:
```
even-row > .flag-color {
    color: pink;
}

odd-row > .flag-color {
    color: magenta;
}
```

Currently this is a rather hypothetical example, it's just to show the point. If any css selector is used other than just the .flag-color class, the css snippet will not be found in your test.

Secondly `flag-color` is not a good name for a css class selector. It already presumes only color will be set and prevents future expansion. Another user may want to add properties (and accompanying code) to add margins for example. flag-color would not really cover the class any more.

A better name could be 'notes-flag' or even 'bugdet-notes-indicator'.

gjanssens: Thinking some more about this, it may even be better to query this color insed the gnc_cell_renderer_text_flag code as part of the initialization function. In that case the name should probably also become more generic, something like 'txt-renderer-flag' or something like that.

Doing it in the initialization function of the gnc_cell_renderer_text_flag widget allows a consumer of that widget to still override the color before rendering by setting the `flag-color-rgba` property of the widget. I personally would not recommend anyone to do so programmatically, but it keeps the renderer generic and customizable, with a css based default style.

ianchi: @gjanssens the flag renderer uses an attribute to define the color, using the same approach as the standard widget.
The renderer itself is not a widget, and thus it doesn’t have a widget context of it own. I don’t see how this can be done with a generic style.

Regarding setting the color in the initialization of the renderer, as there is no widget, you have no context and so no possibility to use css.

I agree that the “flag-color” approach has it limitations, but they are the same as in other places of gnucash (like getting the negative color).
Additionally, the tree widget doesn’t allow to do much customizing via css (the odd-row example you used is not supported), most customization must be done via attributes.

I don’t find a more generic way to do it. The only widget present is the whole treeview. Using that widget as context for the css wouldn’t work as it would set the “color” css attribute to the whole view, not just the notes label.

gjanssens: It's a bit more complicated than I first thought indeed as the renderer doesn't have a widget.
And unfortunately Gtk doesn't expose its private GtkNode interface. That would have allowed you to add another css node to the budget GtkTreeView.

I still think we can come close to what I have in mind by adding the label widget directoy as a hidden child to the treeview. That way you'd have an extra css node users can style under the treeview instead of requiring users to style some pre-defined class. Have you investigated this possibility ?

That aside, I also would like you to change the class name to something that's more generic than color. There may be other useful css settings in the future and changing names after they have been introduced is a pain.

ianchi: I don’t think I follow your idea:

```
I still think we can come close to what I have in mind by adding the label widget directoy as a hidden child to the treeview.
```

There is no “label widget”, just the renderer, which does not create any child widget. It just draws on the treeview canvas.
Creating child widgets defeats the whole “renderer” concept of the treeview.
 I’ll change the class name and rebase, but I don’t find a different approach.

ianchi: Rebased and changed css class name from `flag-color` to `bugdet-notes-indicator`.

ianchi: Regarding your comment:

```
That way you'd have an extra css node users can style under the treeview instead of requiring users to style some pre-defined class. 
```
I'm not sure sure I understand. You mean that instead of styling a *predefined* class, the should style a non-standard hierarchy like:
```
treeview label { 
   color: ....
```

I think it *might* be possible, but I really don't see how that is better or more intuitive.
In fact I find it is much more difficult to understand what it targets.

In general I see that in GnuCash all styling is done with *predefined classes*:
```
*.register-foreground {
*.register-header {
.negative-numbers {
```
why use such a different approach in this case only?

gjanssens: Ok, I took a pause to give this some more thought.

First an answer to your question "why use a different approach in this case only?"
Because  the case is different. The examples you mention define these style classes inside the widgets to which they refer, where your style class is defined outside of that widget. That is the core of my issue with your approach.

However stepping back I think we may be over-engineering this whole thing. I actually think that creating a custom renderer only to be able to add a visual clue to a note may be a bit too much. A treeview column can have more than one renderer. So why not just add two renderers to each cell ? One to display the number and the other to display a small indicator (which can be a small icon) whenever a note is attached to the value. It seems to me this would be more straightforward and reusing already functionality in gtk.

If you like the icon to follow the current theme you can pick one of [the standard icons](https://standards.freedesktop.org/icon-naming-spec/icon-naming-spec-latest.htmlurl). You could also find the names by opening the glade ui editor and create an GtkImage widget. Then the Image selector will list all stock images you can use directly in your image renderer.


Aside from that, I have built your branch and played with it. It looks pretty nice. I would only suggest to add the "Note" feature to the "Edit" menu, not only the tool bar.

gjanssens:
Aside from that, I have built your branch and played with it. It looks pretty nice. I would only suggest to add the "Note" feature to the "Edit" menu, not only the tool bar.

gnucash/gnome/gnc-budget-view.c
ianchi: Hi @gjanssens,
I don’t think that adding an icon to the cell is a good idea to just signal a note.
It would be too intrusive, potentially cluttering the grid if there are many notes, and it would take too much space, making columns wider and so less columns would fit on the screen.
I tried not to reinvent the wheel here, and used the most “standard” and proven approach used by Excel (and so I think also more intuitive to users) of using just a small triangle to mark notes.
As the use case is very specific, I don’t think the icon should be customizable (apart from color).

Regarding the custom renderer approach, the truth is that I’m not a gtk expert and I just couldn’t find a way to use two std renderers (text renderer + pixbuf renderer) and make the pixbuf draw the triangle exactly at the corner and without taking any extra space in the cell. 
If you can give me some directions, I could give it a new try.

Anyhow the custom renderer is rather simple and doesn’t add too much complexity nor size.

gnucash/gnome/gnc-budget-view.c
gjanssens: Unfortunately on my system the highlight color is also blue, so the flag corner disappears in that context.

gnucash/gnome/gnc-budget-view.c
gjanssens: > I tried not to reinvent the wheel here, and used the most “standard” and proven approach used by Excel (and so I think also more intuitive to users) of using just a small triangle to mark notes.

That assumes users are using Excel. I don't and I think most linux user don't. I haven't seen this in LibreOffice yet.
I mostly mention this to point out we should be careful with our assumptions. Other than that I think a triangle is fine. I didn't suggest the "icon" had to be a full fledged icon. An image of a triangle may have worked as well. However I have no idea how far you can configure two renderers to share the same rendering area. I suspect it's possible, but I have never experimented with this. Drawing an image in the top-right corner by itself is easy. One can simply set xalign to 1 (right aligned) and yalign to 0 (top aligned).

> Anyhow the custom renderer is rather simple and doesn’t add too much complexity nor size.

That is a matter of perspective of course... Even simple code can add bugs, and your code actually adds a potential memory leak already (I'll mark it in a minute). In addition it adds another code branch in the gnucash tree view (initialization with or without a custom renderer). In testing that would mean a factor two more of parameter combinations to check.

But none of that is really the point of my issue with this PR. My main issue is that the CSS selector you have set up for the flag color is not logically tied to the widget using it and the suggestion of using a combination of two renderers was one thought experiment to get rid of this. To be honest I don't believe either it will be such a clean solution.

Your custom cell renderer in itself is conceptually ok. We need to figure out a cleaner way to define the color(s) it should use. I am a strong proponent of reusing what's already available if it comes to styling. So here's another avenue to explore:
* The treeview uses different background colors for a cell (or row) that's selected and one that's not
* These colors come from the current css theme and hence should be carefully chosen by the theme designer
* So as colors for the flag triangle I propose to use the opposite colors: when a cell is not selected use the treeview's selection background color as flag color. For a selected cell use the treeview's normal background color as flag color. That way
1.  we stay in the theme
2. we should have flag triangles that are visible in all situations
3. you don't need the whole clutch to create our own theme overrides.

You can probably fine-tune it a little more even by using selected+disabled's  background color if the cell is disabled. This is typically slightly dimmed.

gnucash/gnome-utils/gnc-tree-view-account.c
```diff
@@ -1881,12 +1881,25 @@ gnc_tree_view_account_add_custom_column(GncTreeViewAccount *account_view,
                                         GncTreeViewAccountColumnTextEdited
                                         col_edited_cb)
 {
-    GtkCellRenderer *renderer;
+    GtkCellRenderer *renderer = gtk_cell_renderer_text_new();
```
gjanssens: In the original function it was first verified that account_view is actually of type GNC_TREE_VIEW_ACCOUNT before a renderer is created. If not the function returned NULL.

With your changes the function will always create the renderer yet it may never be used if the account_view parameter is not of type GNC_TREE_VIEW_ACCOUNT. So you're potentially leaking a renderer for each invalid call to this function.

Be sure to first test the inputs and only then create the renderer.

gnucash/gnome/gnc-budget-view.c
```diff
@@ -1380,20 +1397,17 @@ gnc_budget_view_refresh(GncBudgetView *view)
     /* Create any needed columns */
     while (num_periods_visible < num_periods)
     {
-        GtkCellRenderer* renderer;
-
-        col = gnc_tree_view_account_add_custom_column(
+        GtkCellRenderer *renderer = gnc_cell_renderer_text_flag_new ();
+        g_object_set(renderer, "flag-color-rgba", &flag_color, NULL);
```
gjanssens: I suggest to keep all renderer configuration together in the call to budget_col_source function. This will make even more sense if you reuse the selection and disabled background colors.

gnucash/gnome-utils/gnc-tree-view-account.c
ianchi: fixed

gnucash/gnome/gnc-budget-view.c
ianchi: I could use some help here, as I'm not much of an expert on gtk.
I can't find a simple way to get the treview's current "selected" color and the current row's color.
Some hints?

gjanssens: You can pretty much reuse the code you have in your `get_flag_color` function. However in this case you get the style context from the budgetview widget.
The colors you're looking for are background colors which you can retrieve with `gtk_style_context_get_background_color`

The background color used for highlighting rows is the one for the `GTK_STATE_FLAG_SELECTED` state. The background color for other rows is the one you get for the `GTK_STATE_FLAG_NORMAL` state.

I think you can forget what I wrote about the dimmed state. That probably doesn't apply here as we never disable the treeview widget.

ianchi: Ok, understood.

But in gtk doc it states:
```
gtk_style_context_get_background_color has been deprecated since version 3.16 and should not be used in newly-written code.
Use gtk_render_background() instead.
```

That’s why I originally discarded this function, but the doc proposed `gtk_render_background` doesn’t apply to our use case.

Should we proceed with a function marked as deprecated?

gjanssens: Probably not. However there's also the more generic and non-deprecated gtk_style_context_get, which you can use to achieve the same thing.

gnucash/gnome/gnc-budget-view.c
ianchi: Switched to using "selected" and "normal" background colors for flag. No custom css

gnucash/gnome/gnc-budget-view.c
ianchi: Switched from css to using inverted "selected"/"normal" cell's background color for note's flag

gnucash/gnome/gnc-budget-view.c
ianchi: done

ianchi:
Hi @gjanssens, I think I've addressed all the requested changes.

Now there is no css color configuration, and the note's flag color is taken from the *inverted* "selected"/"normal" cell's background color.

My only concern is that doing all this in `budget_col_source` could degrade a bit more the already awful performance of the budget view.
This function already does a lot of things on the fly, and it gets called hundreds of time by the TreeView (even when only moving the mouse over a cell).

But fixing the performance problem is a completely different issue (which I think would need a complete redesing of the view's approach).

Please, tell me if you prefer that I squash the last three fix commits with the original ones (I left them separate for easier revision).

ianchi:
I've thought it a bit over and added 2 new commits changing the approach and moving all the notes color logic out of `budget_col_source` so as not to impact on performance.

Now the colors are determined only once (when creating the renderers) and then the renderer selects the correct color based on state.
This way no costly Account selection lookup is needed on every mouse move.

gjanssens:
Thanks for the updates. The code looks good to me now, including the performance enhancement.

As for squashing: it would be useful to squash small fixups with the commits that required the fixup. Like the potential memory leak fix can be squashed with the commit that introduced that potential memory leak.
The commits reimplementing the color selection from a hidden widget to using inverted selection colors can also be squashed into the commit implementing the original color selection.
Other than that, I like the commit history telling the story of how this feature has been constructed.

Lastly, I think it's now time to rebase on master as well before the final merge.

ianchi:
Rebased and squashed commits.

christopherlam:
@ianchi I've taken the liberty of changing this PR base from maint to master to remove superfluous commits.

gjanssens:
> Rebased and squashed commits.

When pulling this PR it was rebased, but still on the maint branch (branched off of 315bbb5d0539d7d911b867b2eac8406f8d3e4484 ). I don't know if this is due to @christopherlam 's change of the PR base or a rebase gone wrong.

No worries, I have properly rebased on master and pushed the result. I'm mainly mentioning this to explain why this PR will be closed by hand while your changes have properly been merged in our repo.

Thanks a lot for your work!

ianchi:
Hi @gjanssens / @christopherlam ,

The Notes functionality for budgets still sits on master, and after two releases it hasn't made it to maint.
Is there a reason for this? When are you planning on including this in production version?
Is there any changes that you would like me to make?

Thanks!

jralls:
@ianchi, It's going into 4.0, planned for release at the end of June with testing releases beginning 26 April. We do not generally add new features, especially ones that add a pile of new storage items, to the stable series. That's part of what makes it stable.

ianchi:
Clear.
Thanks

PR #515, opened 2019-06-03
Restructure shell detection code

gjanssens:
It will now first check for an environment variable GNC_SHELL and use that if it exists. Then hardcode a shell if cmake detects the old mingw environment (not mingw64), and finally fall back to cmake's
built-in method of finding bash.

This requires a fix in gnucash-on-osx for the build on that platform to continue to work: on that platform, we should set GNC_SHELL before starting the build. A patch has been submitted to do that automatically for the user.


jralls:
It works on Mac, so I pushed it for you.

PR #516, opened 2019-06-03
Some changes started from bug 797266

Bob-IT:
OK, bug 797266 was about a missing translation of a button so in fixing this I decided to look at all translations in the glade files and enable any missing in that commit. In doing so I noticed some other things which I have added in separate commits to try and make the dialogues be consistent.

I have changed the ab buttons to remove the icons Frank had added just to make the application buttons all the same, if icons are wanted then I think it should be applied consistently in one go maybe for 4.0 but I am willing to drop the change which is why it is a separate commit.

I no one objects I would like to push this at the weekend.

### DIFF
--- a/gnucash/gnome-utils/gnc-frequency.c
+++ b/gnucash/gnome-utils/gnc-frequency.c
@@ -229,7 +229,7 @@ gnc_frequency_init(GncFrequency *gf)
     /* Add the new widget to the table. */
     {
         GtkWidget *table = GTK_WIDGET(gtk_builder_get_object (builder, "gncfreq_table"));
-        gtk_grid_attach(GTK_GRID(table), GTK_WIDGET(gf->startDate), 4, 0, 1, 1);
+        gtk_grid_attach(GTK_GRID(table), GTK_WIDGET(gf->startDate), 3, 0, 1, 1);
         gtk_widget_set_vexpand (GTK_WIDGET(gf->startDate), FALSE);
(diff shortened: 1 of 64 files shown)


gjanssens:
This is a nice round of cleanups. My only reflection is whether this should go on maint or master (as we approach a new master release in December/January).

In particular, this PR introduces 88 translation changes. I feel that's a bit much for maint. Doing it on master will give translators a bit more time to translate for the future release. @fellen what do you think ?

Similarly I wonder how to treat the various tweaks to the dialogs. Things like replacing VBoxes with Grids may be refactorings that are better on master, where smaller tweaks could fit on maint. If it's difficult to separate I'd be inclined to do most gui tweaks on master now.

Bob-IT:
Yes I was in two minds which one of the reasons for the PR, I am happy to push it all to master or maybe I could separate out the actual bug fix reported and the last two to maint and push the rest to master.

gjanssens:
I'm fine with the bugfix and last two commits on maint.

Bob-IT:
After some more thought, I would like to push to maint the first two commits, the ab commits and the last two commits. As I understand it, there should already be translations for most of the buttons I am enabling,( _OK, _Cancel, _Close) so there should be nothing to do and the second commit does not make any visible changes, its just changes to the layout of the glade files.

gjanssens:
Ok, that's reasonable. I can imagine it would become harder to work on glade files if you can't do the glade file changes on maint.

Bob-IT:
I am closing this as I have added the commits to maint, the remaining ones will be added to master when ready.

PR #520, opened 2019-06-14
[html-pie/scatter/bar/linechart] defer to html-chart renderer

christopherlam:
- [x] change old jqplot renderer to call html-chart's renderer instead.
- [x] modify old-chart modifiers and add deprecation warnings

### DIFF
--- a/gnucash/report/html-piechart.scm
+++ b/gnucash/report/html-piechart.scm
@@ -41,6 +41,16 @@
 (define gnc:html-piechart? 
   (record-predicate <html-piechart>))
 
+(define-syntax-rule (gnc:guard-html-chart api)
+  ;; this macro applied to old html-bar/line/scatter/pie apis will
+  ;; guard a report writer from passing html-chart objects. this
+  ;; should be removed in 5.x series.
+  (let ((old-api api))
+    (set! api
+      (lambda args
+        (if (and (pair? args) (gnc:html-chart? (car args)))
+            (gnc:warn "using old-api " (procedure-name api) " on html-chart object. set options via gnc:html-chart-set! or its shortcuts gnc:html-chart-set-title! etc, and set data via gnc:html-chart-add-data-series! see sample-graphs.scm for examples.")
+            (apply old-api args))))))
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;  <html-piechart> class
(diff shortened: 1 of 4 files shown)


christopherlam:
fao @gjanssens second commit is the right approach to deprecate old api :)

gjanssens:
So if I understand this code correctly, you have
1. first rewritten the old chart generators to defer actual generation to the new chartjs based api such that no code will actually depend on jqplot any more.
2. written convertors for the various setters from the old api (set width and so on) to do the same in the new api
3. added code to issue a warning to the user in case they still use any of the old api calls

Is that correct ?

gnucash/report/report.scm
```diff
@@ -37,6 +37,21 @@
...
+  ;; this macro applied to old html-bar/line/scatter/pie apis will
+  ;; guard a report writer from passing html-chart objects. this
+  ;; should be removed in 5.x series.
+  (syntax-rules ()
+    ((_ api)
+     (let ((old-api api))
+       (set! api
+         (lambda args
+           (if (and (pair? args) (gnc:html-chart? (car args)))
+               (gnc:error "trying to use old-api on html-chart object.")
```
gjanssens: I think a gnc:warning is more appropriate: gnc:error is meant to inform the user the requested command can't be executed at all and hence the expected result won't be achieved. In this case you have a fallback that still does what the user is expecting, but that fallback will eventually go away. And we want to warn the user well in advance.

I'd also try to write the message in a way it is more conductive to a solution. This only tells something is wrong it doesn't help much in making it right. Remember the user/developer may not be very deep into our api as you and I would be.

For example, mention the specific function that was called (if possible), and indicate it's deprecated. And explain in the new api chart properties are set via xyz or point at some ready to use documentation or examples.

gnucash/report/html-piechart.scm
```diff
@@ -53,13 +53,17 @@
...
   (gnc:make-html-piechart-internal '(pixels . -1) '(pixels . -1) #f #f #f #f #f #f #f #f #f #f #f))
 
 (define gnc:html-piechart-data
   (record-accessor <html-piechart> 'data))
 
-(define gnc:html-piechart-set-data!
-  (record-modifier <html-piechart> 'data))
+(define (gnc:html-piechart-set-data! pie newdata)
+  (if (gnc:html-chart? pie)
+      (issue-deprecation-warning "uhhh wrong api.")
```
gjanssens: Uhhh, horrible deprecation warning :(
For all the reasons I have mentioned earlier: this only tells the user they did something wrong, but it won't give them a clue as to how to fix it.

christopherlam: forgot this one

christopherlam:
> * first rewritten the old chart generators to defer actual generation to the new chartjs based api such that no code will actually depend on jqplot any more.
> * written convertors for the various setters from the old api (set width and so on) to do the same in the new api
> * added code to issue a warning to the user in case they still use any of the old api calls

yes, no, yes

The various old-api getters/setters still set old html-bar/line/scatter/pie objects. The old-API renderer will create html-chart object, copy values from old-object into html-chart properties, and call html-chart renderer.

PR #524, opened 2019-06-17
[master] UI to change currency of root account

christopherlam:
## UI to set root currency
Effects: it changes the currency of the summary bar. I am not sure any other side effects.
This does not change format of datafile, so, it could be merged into maint if wished.

## Internal
It adds `gnc:make-currency-complex-option` with callbacks.

### DIFF
--- a/libgnucash/app-utils/app-utils.scm
+++ b/libgnucash/app-utils/app-utils.scm
@@ -77,6 +77,7 @@
 (export gnc:make-text-option)
 (export gnc:make-font-option)
 (export gnc:make-currency-option)
+(export gnc:make-currency-complex-option)
 (export gnc:make-commodity-option)
 (export gnc:make-simple-boolean-option)
 (export gnc:make-complex-boolean-option)
@@ -272,6 +273,7 @@
 (load-from-path "date-utilities")
 
 ;; Business options
+(define gnc:*book-base-currency* (N_ "Book Base Currency"))
 (define gnc:*business-label* (N_ "Business"))
 (define gnc:*company-name* (N_ "Company Name"))
 (define gnc:*company-addy* (N_ "Company Address"))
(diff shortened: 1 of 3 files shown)


jralls:
@christopherlam Can you explain the name of the option? It seems a bit odd to me: Complex in what way? 

More broadly, ISTM that the root currency's being invariant is a design decision from way back and we need to be absolutely certain of the all of the effects of changing it after the book's creation before we make it not-invariant. Part of that review should consider what the current application preference currency is used for and whether each of those uses should instead use the root account currency and vice-versa. For example I think a good case could be made that the summary bar should use the preference default currency rather than the root account's currency, and that the Transaction dialog should use the root account's currency instead of the preference default currency.

christopherlam:
It's not really complex however follows the naming convention of https://github.com/Gnucash/gnucash/blob/maint/libgnucash/app-utils/options.scm#L468 -- perhaps it's best named `gnc:make-currency-callback-option` following https://github.com/Gnucash/gnucash/blob/maint/libgnucash/app-utils/options.scm#L873

I recall @derekatkins mentioned it would be reasonable to have UI to change ROOT currency (to allow changing summay-bar currency) however not sure if it'll have nasty side effects elsewhere.  Let's leave this pending for now?

jralls:
If it's just about the summary bar, make it follow the default currency preference. The summary bar is a display thing and if anything follows a global preference it should display things. It would be a worthwhile enhancement to make a new option, root-account currency, and maybe it should be the default instead of locale.

The much larger question about which parts should be controlled by a Preference and which by a book option is a matter for discussion in gnucash-devel. I don't think that there's time to have the discussion and make the changes for 4.0, but the sooner we start the discussion the better. Alex's hapf done "book-currency" feature is a piece of that puzzle and I think it could do with a proper strategy discussion as well.

christopherlam:
Then https://github.com/Gnucash/gnucash/blob/maint/gnucash/gnome-utils/window-main-summarybar.c#L361 is a simple change to retrieve the default-report-currency, which I'd think is a good change. I'll amend this PR to that effect soon.

It would help me to know which .scm PRs seem reasonable to complete and thoroughly test in the next 10 days...

PR #526, opened 2019-06-21
Set the summarybar currency to default_currency

christopherlam:
Offshoot from #524 -- sets the summary-bar display currency to default_currency.

This seems uncontroversial -- modifies display currency only. It's also updated live in Edit / Preferences.

### DIFF
--- a/gnucash/gnome-utils/window-main-summarybar.c
+++ b/gnucash/gnome-utils/window-main-summarybar.c
@@ -359,10 +359,10 @@ gnc_main_window_summary_refresh (GNCMainSummary * summary)
 
 
     root = gnc_get_current_root_account ();
-    options.default_currency = xaccAccountGetCommodity(root);
+    options.default_currency = gnc_default_currency ();
     if (options.default_currency == NULL)
     {
-        options.default_currency = gnc_default_currency ();
+        options.default_currency = xaccAccountGetCommodity(root);
     }
 
     options.grand_total =


jralls:
You might consider leaving `xaccAccountGetCommodity(root)` as the fallback in case `get_default_currency()` returns `NULL`, as it will if the user has set it to Locale and the Locale is 'C'.

christopherlam:
So, a straightforward swap, as per latest commit?

jralls:
Yup.

fellen:
What the hell is happening here?
What are the reasons?
Was there a discussion somewhere?

jralls:
Calm down @fellen. All this does is use Preferences>Accounts>Default Currency instead of the root account currency to set the currency for the summary bar so that users can change it if they want to.

fellen:
And if the user has files in different currencies, he has to go to edit->preferences... each time, he opens a different file?

christopherlam:
@fellen  what would you suggest? There have been many requests for facility to change summary-bar currency. I think this is a valid request; I may have started a book with USD but decide to create most accounts in GBP and am now stuck with summarybar showing dollars.

1. apply UI as #524 to change root account's currency. Unknown number of side effects.

2. add new book option "book-summarybar-currency", in addition to aycinena's book-currency-accounting option, IMHO a useful idea, is still unfinished, 4 years on. More options burden.

3. use heuristics to determine book's preferred currency -- I've used a customised one `balsheet-pnl.scm` -- it has some harmless magic but isn't used anywhere else at https://github.com/Gnucash/gnucash/blob/03d9bf902a74d0b8e2a3dc5f1a178583a4993c89/gnucash/report/standard-reports/balsheet-pnl.scm#L833

With this commit, I agree there's a requirement of changing default_currency, is to me the *least* harmful option to allow different summarybar currencies.

fellen:
Most users have only one file, others have all files in the same locale and currency, This users are not heavily affected.
But some users are using Gnucash for their tax reports - tax authorithies usually accept only reports with the final statements in their official currency - and are bordercrossers, i.e. they have their business in an EUR country, but reside in an country which still has its own currency, e.g. PLN, (I still remember the times, when I had always BEF/LUF, DEM and NLG in my wallet.) This people are forced to have more than one book in different currencies.

From your list, I would prefer 2,, but how about using the Edit->Preferences->Reports->"Default Report Currency" also for the summary bar?

Probably it is worth to expand both dialogs to use 1. book currency, (which is set on file creation), 2. locale, 3. choose.

About you beginning error
1. could we improve the docs so that others can avoid it?
2. can you not replace the USD accounts accounts one by one with GBP accounts, finally export all transactions to import them in a file with the right root currency?

jralls:
> From your list, I would prefer 2,, but how about using the Edit->Preferences->Reports->"Default Report Currency" also for the summary bar?

How would that be any different from using the Preferences>Accounts default currency beyond being less intuitive because the summary bar isn't a report?

> Probably it is worth to expand both dialogs to use 1. book currency, (which is set on file creation), 2. locale, 3. choose.

+1. We could add a fourth, "account currency" for the summary bar on register pages.

It might be worthwhile to detect when accounts can't be priced in the selected currency and so the summary bar or report is incorrect and display that somehow.

christopherlam:
> About you beginning error
> 
> 1. could we improve the docs so that others can avoid it?
> 2. can you not replace the USD accounts accounts one by one with GBP accounts, finally export all transactions to import them in a file with the right root currency?

I'm sorry it's completely unreasonable to expect users to read docs top-to-bottom before experimenting with a datafile. Ditto, do you have any idea how difficult it is to export & reimport gnucash data? If it was that easy, I would never have spent hours and hours to learn scheme two years ago; I'd have everything I need via PivotTables. With the newer CSV import & export, I haven't experimented with it; there's lots of metadata within the transactions that I am sure are not exportable -- business data, file-association, online_id for ofx matching etc.

So, it seems we want some global and book options.

Global:
- preferred locale-specific currency -- (to be used for what exactly?)

Book:
- default book currency - for new accounts etc. (and the "Transfer Funds" window?) (and what else?)
- default report currency - obvious - for reports
- default book gains/loss currency - for aycinena
- default summarybar currency - for the summary bar

So do you really suggest add new book-kvp option? This commit changes from invariant book-root-currency to modifiable global-default-currency (or could be changed to your suggestion, the global-default-report-currency) -- agree we should set a book-currency but there's no such book property for now.

@jralls 
> It might be worthwhile to detect when accounts can't be priced in the selected currency and so the summary bar or report is incorrect and display that somehow.

Agree -- most exchange-functions convert foreign to $0 domestic when no conversion is possible; this is why my balsheet-pnl.scm will avoid this and render original-currency instead of target-currency when conversion is impossible. There are other niggles in the summary-bar e.g. clicking it with a multicurrency book shows an unpopulated long drop down.

christopherlam:
> Probably it is worth to expand both dialogs to use 1. book currency, (which is set on file creation), 2. locale, 3. choose.

+1 This is not a bad idea -- so, the function gnc_default_currency() will then query the current book's root's currency and return it? Effectively this function becomes suddenly much more useful, because it returns a book-specific property.

jralls:
I suppose `gnc_default_currency()` would check the option (that should probably default to `book currency`) and return the appropriate value. That would fix a lot of currently questionable code that uses `gnc_default_currency()` where `xaccAccountGetCommodity(root)` is more likely to be correct.

christopherlam:
Ok here's a patch -- the UI part is too difficult for me. Note `gnc-ui-util.c` has code for returning book-currency already but depends on aycinena's code being completed. This is visible in the 2nd section of the patch below

```patch
modified   libgnucash/app-utils/gnc-ui-util.c
@@ -61,6 +61,7 @@
 #include "gnc-guile-utils.h"
 
 #define GNC_PREF_CURRENCY_CHOICE_LOCALE "currency-choice-locale"
+#define GNC_PREF_CURRENCY_CHOICE_ROOT   "currency-choice-root"
 #define GNC_PREF_CURRENCY_CHOICE_OTHER  "currency-choice-other"
 #define GNC_PREF_CURRENCY_OTHER         "currency-other"
 #define GNC_PREF_REVERSED_ACCTS_NONE    "reversed-accounts-none"
@@ -1221,6 +1222,11 @@ gnc_default_currency_common (gchar *requested_currency,
     if (gnc_book_use_book_currency (gnc_get_current_book ()))
         return gnc_book_get_book_currency (gnc_get_current_book ());
 
+    if (gnc_prefs_get_bool (section, GNC_PREF_CURRENCY_CHOICE_ROOT))
+    {
+        currency = xaccAccountGetCommodity (gnc_get_current_root_account ());
+    }
+
     if (gnc_prefs_get_bool (section, GNC_PREF_CURRENCY_CHOICE_OTHER))
     {
         mnemonic = gnc_prefs_get_string(section, GNC_PREF_CURRENCY_OTHER);
@@ -2659,12 +2665,16 @@ gnc_ui_util_init (void)
                           gnc_configure_reverse_balance, NULL);
     gnc_prefs_register_cb(GNC_PREFS_GROUP_GENERAL, GNC_PREF_CURRENCY_CHOICE_LOCALE,
                           gnc_currency_changed_cb, NULL);
+    gnc_prefs_register_cb(GNC_PREFS_GROUP_GENERAL, GNC_PREF_CURRENCY_CHOICE_ROOT,
+                          gnc_currency_changed_cb, NULL);
     gnc_prefs_register_cb(GNC_PREFS_GROUP_GENERAL, GNC_PREF_CURRENCY_CHOICE_OTHER,
                           gnc_currency_changed_cb, NULL);
     gnc_prefs_register_cb(GNC_PREFS_GROUP_GENERAL, GNC_PREF_CURRENCY_OTHER,
                           gnc_currency_changed_cb, NULL);
modified   libgnucash/core-utils/gnc-prefs.h
@@ -91,6 +91,7 @@
 /* Currency preferences */
 #define GNC_PREF_CURRENCY_OTHER      "currency-other"
 #define GNC_PREF_CURRENCY_CHOICE_LOCALE "currency-choice-locale"
+#define GNC_PREF_CURRENCY_CHOICE_ROOT   "currency-choice-root"
 #define GNC_PREF_CURRENCY_CHOICE_OTHER  "currency-choice-other"
 
 /** @name Early bird functions, needed before any backend has been set up
```

fellen:
I would prefer to have choice-root as the first and default choice.

fellen:
N.B. I wonder, if really both existing option are required, if they by default are set to book-currency.

jralls:
Maybe not, regardless of what is the default. Based on the placement I suspect that the original intent of the Accounts one is to set the default commodity when creating a new account. In that light one can almost see a use-case for a separate preference for default report currency if one squints really hard... but if the default currency is used to drive the UI display one has to squint even harder. Why would a user want the UI to display in one currency and reports in a different one?

Then there's Alex's book-currency project. I'll re-read those emails where he lays out his plan, but I have to wonder if he might be going in the wrong direction.

jralls:
Heh, the tooltip text on the Accounts one even says "Use the specified currency for all newly created accounts".

jralls:
As for the UI, it should be a simple matter of firing up Glade and adding a button to each of the default currency radio buttons. @fellen I think that's where the order and default are set, not in the properties themselves.

fellen:
Right, but for future readers it would look more nice, if it would already follow the logic behind.

fellen:
@Bob-IT, can you improve the GUI?

christopherlam:
While I feel that gnc_default_currency() potentially returning book's root's currency is a powerful upgrade, I am thinking that it could also have side effects:

- calls to `gnc_default_print_info (TRUE)` code retrieves it and prints an amount with the currency obtained; this is present in https://github.com/Gnucash/gnucash/blob/2d06d703f2164eeeb7782f61a379fc1dc06735ba/gnucash/gnome-utils/dialog-tax-table.c#L381 although I don't see where `row_text[1]` is then actually used.

- called by https://github.com/Gnucash/gnucash/blob/2d06d703f2164eeeb7782f61a379fc1dc06735ba/gnucash/gnome/dialog-fincalc.c#L156 to variable `commodity` which is immediately only used to find its SCU -- I feel is an invalid usage

- used by dialog-vendor.c and dialog-employee.c when the vendor/employee is NULL

- etc.

It seems gnc_default_currency() is being abused!

I'll aim to eradicate its use in .scm files.

Bob-IT:
@fellen I have created a PR #528 that updates the preference gui along with the patch above. Please look at and suggest changes on that PR. I have not been following so I may of got the wording wrong. 

The selection does change summary bar currency but does not affect 'New Account' creation default currency, I assume it is pendent on parent account currency.

Also I have a test file that may of been created from a Qif import long ago and it does not have a currency for the root account and so in the preference that would show up as blank.

fellen:
> > From your list, I would prefer 2,, but how about using the Edit->Preferences->Reports->"Default Report Currency" also for the summary bar?
> 
> How would that be any different from using the Preferences>Accounts default currency beyond being less intuitive because the summary bar isn't a report?

Both are about output, while Accounts>Default Currency is about input (creation of accounts), which follows a different logic. BTW the default means "parent account", not locale.

jralls:
I think that in reality account creation pays no attention at all to this preference and that both the name and tooltip are obsolete. However this preference is (mis)used in a lot of other places that are neither explained, obvious, nor in most places even correct.

GUI != Reports. If there's going to be a single preference it should be on the General tab, not the Reports or the Accounts tab.

fellen:
> If there's going to be a single preference it should be on the General tab, not the Reports or the Accounts tab.

+1

christopherlam:
> GUI != Reports. If there's going to be a single preference it should be on the General tab, not the Reports or the Accounts tab.

I can eradicate gnc_default_currency in reports -- I don't think reports have any business knowing this. However gnc_default_report_currency is definitely of use. I'd prefer if the latter is also capable of looking up root currency.

PR #527, opened 2019-06-23
Check and UTF8-encode the names if needed

wrobelda:
Fixes the "Wide character in subroutine entry" error for some UTF8 entries


jralls:
Thanks!

PR #528, opened 2019-06-24
Summary bar update to previous commit

Bob-IT:
Add preference options to select root account currency.
This was requested in PR #526 for the preference gui to be updated.
Screen shot below, report option is the same as account option but tooltip different wording but along the same lines of the other options.

Screenshot_2019-06-24_11-39-10

### DIFF
--- a/gnucash/gnome-utils/dialog-preferences.c
+++ b/gnucash/gnome-utils/dialog-preferences.c
@@ -1348,6 +1348,7 @@ gnc_preferences_dialog_create (GtkWindow *parent)
     GtkTreePath *path;
     GtkTreeIter iter;
     gnc_commodity *locale_currency;
+    gnc_commodity *root_currency;
     const gchar *currency_name;
     GDate fy_end;
     gboolean date_is_valid = FALSE;
@@ -1501,6 +1502,13 @@ gnc_preferences_dialog_create (GtkWindow *parent)
     label = GTK_WIDGET(gtk_builder_get_object (builder, "locale_currency"));
     gtk_label_set_label (GTK_LABEL(label), currency_name);
     label = GTK_WIDGET(gtk_builder_get_object (builder, "locale_currency2"));
+    gtk_label_set_label (GTK_LABEL(label), currency_name);
+
+    root_currency = xaccAccountGetCommodity (gnc_get_current_root_account ());
+    currency_name = gnc_commodity_get_printname (root_currency);
+    label = GTK_WIDGET(gtk_builder_get_object (builder, "root_currency"));
+    gtk_label_set_label (GTK_LABEL(label), currency_name);
+    label = GTK_WIDGET(gtk_builder_get_object (builder, "root_currency2"));
     gtk_label_set_label (GTK_LABEL(label), currency_name);
 
     button = GTK_WIDGET(gtk_builder_get_object (builder, "pref/general/save-on-close-expires"));
--- a/gnucash/gnome-utils/dialog-preferences.c
+++ b/gnucash/gnome-utils/dialog-preferences.c
@@ -80,6 +80,7 @@
 #include "gnc-component-manager.h"
 #include "dialog-preferences.h"
 #include "dialog-doclink-utils.h"
+#include "qofinstance.h"
 
 #define DIALOG_PREFERENCES_CM_CLASS "dialog-newpreferences"
 #define GNC_PREFS_GROUP             "dialogs.preferences"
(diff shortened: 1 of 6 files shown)


fellen:
> @fellen I have created a PR #528 that updates the preference gui along with the patch above. Please look at and suggest changes on that PR. I have not been following so I may of got the wording wrong.

I personally would prefer to have the default (parent/root account) as first choice.

> The selection does change summary bar currency but does not affect 'New Account' creation default currency, I assume it is pendent on parent account currency.

You are right here and most users would prefer "Parent account" over "Root account" for new accounts. and that seems to be system settings.

> Also I have a test file that may of been created from a Qif import long ago and it does not have a currency for the root account and so in the preference that would show up as blank.

That is an interesting use case. I assume you should file a bug against the importers to verify the existence of a root account, if new files are created.

Bob-IT:
Changed the default to the Root Account.
Added a fix for when the Qif importer is used from the 'New User' dialogue so it checks for the root currency when the apply button is used at the end of the assistant.
For the preference, when used a check is made for the root account currency and if missing it is set to locale currency as that is what the patch for gnc-ui-util.c would return. I could not think of another way short of presenting a currency edit widget for a one time user set.

jralls:
For the missing root-account currency one could look at the currencies of top-level accounts. In the case of differences I'd give precedence to equity as being the most likely to hold the user's preferred currency for the book. That leaves the possibility of multiple top-level equity accounts in different currencies, in which case popping up a currency picker is perhaps the right thing to do. Until we sort that out I think the best thing to do if the root account has no currency is to simply disable that selection and explain it in the tool-tip. 

I think the vast majority of users have root currency == locale currency == preferred currency and that for the few who don't having the defaults follow the root currency will be the preferred behavior so let's go ahead and merge this.

christopherlam:
While I think that having gnc_default_currency and gnc_default_report_currency return a book-property is a good upgrade, it still doesn't cover the requirement/wish to use a modifiable book-preferred-currency.

Perhaps in the future the default_currency will save into book rather than global prefs. This means `gnc_default_currency_common` will request from `qof_book_get_option` rather than `gnc_prefs_get_bool`.

gjanssens:
The root account's currency could be promoted to be the book level default currency. Then there would be no need for an extra kvp option.
All we need is a bit of gui code that pops up a list of possible currencies from which the user can select a preferred currency. This list could be primed with the current root account currency and any change will be saved into the root account and be used as future default currency.
This would also solve the issue of missing root account currency: the user can simply set one. If we have a gui in place to set it, it may be useful to warn the user at startup if this default currency is missing and ask for it.

Gotchas in this approach:
* I vaguely remember that due to a bug in gnucash some books can have more than one root account. I don't think this happens still with newly created books, however older books are still around. And I don't think we ever wrote code to clean this up. If we want to use the root account currency as default currency, this will have to be cleaned up first.
* I think this would be the first book option that's not stored as kvp, so it may require additional plumbing to fit into the book properties dialog.

christopherlam:
@gjanssens so... https://github.com/Gnucash/gnucash/pull/524 ?

jralls:
There's already GUI for setting the root account currency in the New Account Hierarchy Assistant, but that's easily bypassed. If one does and just starts creating accounts then GnuCash will create a root account with no currency. As it happens I did that 15 years ago when I started using GnuCash and my primary book has no root account currency. 

Heh, and we already have a way to change it: Edit>New Account Hierarchy... runs the New Account Hierarchy Assistant that includes the root account currency page. When run on an existing book it will overwrite the root account currency. Perhaps that was where the bug was that would create a second root account?

jralls:
Whether that's the right place for changing the root-account currency is another matter. I think it isn't and that File>Properties would be a better menu selection, though the resulting "Book Options" dialog is perhaps not the right place. Perhaps File>Properties could open a submenu with items Set Root Currency... and Book Options...

fellen:
@jralls In theory all items from the New file assistant should be editable from book properties, but at the time of writing most of us considered the root currency immutable. Probably because nobody was willing to check the all implications of a changeable root currency.

christopherlam:
While this is waiting to be merged, git-master currently doesn't work very well. See snippet from today 2020-04-24's master.

image

fellen:
I fear, we have forgotten this.
About commit 0ce312e - Make sure Root account has currency for New User Qif import do other importers not deserve such a check?

jralls:
I've rebased this on master, checking to make sure each commit builds and works, and changed the target branch.

To @fellen's comment about importers, I think he's right but doesn't go far enough: It should apply when the first account is created in an empty book no matter how it's created.

Bob-IT:
Thanks, I had forgotten about this, will have to reread to understand what it was about.

jralls:
Rebased on current stable, fixing a conflict.

I'm tempted to merge this. There's clearly no strong motivation to deal with the other pieces and maybe having this in stable will provide it.

jralls:
7 years of dithering. Time to merge. It'll be in GnuCash 6.

PR #529, opened 2019-06-25
[customer-summary] work with multiple currencies

christopherlam:
This line of work starts from fixing customer-summary.scm -- has issues, too numerous to mention
* counts all monetary amounts irrespective of currency
* can't handle multiple customer currencies
* formatting
* now has URL to owner-report.
* hidden useless options removed -- this report hardwired for customers; not much benefit creating parallel reports for vendors/employees/jobs. (saved-report trying to set a removed option does not cause a crash).

owner-report.scm also slightly upgraded
* running Report > Account Report (single transaction) will now run owner-report and find the appropriate APAR account according to owner's currency.
* other small srfi-1 refactoring
* timepair->t64 fix too

I'll leave this here a small while, I do not think these changes are uncontroversial, nor unsafe.

### DIFF
--- a/gnucash/report/business-reports/business-reports.scm
+++ b/gnucash/report/business-reports/business-reports.scm
@@ -28,6 +28,7 @@
 (use-modules (gnucash app-utils))
 (use-modules (gnucash report report-system))
 (use-modules (gnucash report standard-reports))
+(use-modules (srfi srfi-8))
 
 ;; to define gnc-build-url
 (gnc:module-load "gnucash/html" 0)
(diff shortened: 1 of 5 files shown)


christopherlam:
I'll need to remove the sorting options other than owner-name. Because now each owner has sales/expense/profit in several currencies, and it's impossible to find a good sorting strategy.
Any objections?

gjanssens:
Group-by-currency with sort on any of the other columns seems like a sensible use case to me.

christopherlam:
Oh nice idea. I've done this two ways (no option needed). This will enforce groupby currency. See results:

1. The 'owner' entries are grouped by currency, and sort by sales. The 'no-owner' entries are rendered after 'owner' entries, before the total lines.
image

2. All entries 'owner' and 'no-owner' are mixed together, grouped by currency, then sorted by sales.
image

I'd vote for 2 myself...

gjanssens:
Number 2 looks better to me too.

As you force grouping on currency I'd even consider adding a group header with the currency name/symbol and dropping the currency column.
Bonus points: if there's only one currency in the report, drop the group header.

christopherlam:
Latest. Currency section headers only shown when >1 currency and group-by currency is in force i.e. sales/expense/profit sorting.
image

christopherlam:
I do not think this is a complicated report so I'll fixup appropriate commits and merge soon.

PR #530, opened 2019-06-27
Bug797275

DaveC49:
Fix to the multiple selection to only allow multiple selection of rows with the Add checkbox set and to correct the blocking of the running of the match_dialog by implementing a view_selection_function to enforce this action. This also corrects a problem with  toggling a row marked U+rR or R to A which left the row selected such that if it was then toggled back to U+R or R without deselecting the row which then left it in a state where it could not be unselected and blocked selecting other rows. This was occurring because the treeview selection was activated at the same time as the row was toggled adding the toggled row to the selection. This now unselects the row immediately after it is toggled to the A stateby creating a flag which is set in the add_toggled_cb and cleared in the view_selection_function after blocking selection of the row.

### DIFF
--- a/gnucash/import-export/import-main-matcher.c
+++ b/gnucash/import-export/import-main-matcher.c
@@ -54,6 +54,7 @@
 
 #define GNC_PREFS_GROUP "dialogs.import.generic.transaction-list"
 
+
 struct _main_matcher_info
 {
     GtkWidget *main_widget;
@@ -65,6 +66,7 @@ struct _main_matcher_info
     gpointer user_data;
     GNCImportPendingMatches *pending_matches;
     GtkTreeViewColumn *account_column;
+    gboolean add_toggled;   // flag to indicate that add has been toggled to stop selection
 };
 
 enum downloaded_cols
@@ -90,7 +92,10 @@ enum downloaded_cols
 #define CSS_INT_PROB_REQUIRED_CLASS "intervention-probably-required"
 #define CSS_INT_NOT_REQUIRED_CLASS  "intervention-not-required"
 
-static QofLogModule log_module = GNC_MOD_IMPORT;
+/* Define log domain for extended debugging of matcher */
+#define G_MOD_IMPORT_MATCHER "gnc.import.main-matcher"
+/*static QofLogModule log_module = GNC_MOD_IMPORT;*/
+static QofLogModule log_module = G_MOD_IMPORT_MATCHER;
 
 void on_matcher_ok_clicked (GtkButton *button, GNCImportMainMatcher *info);
 void on_matcher_cancel_clicked (GtkButton *button, gpointer user_data);
@@ -117,6 +122,11 @@ static gboolean gnc_gen_trans_onButtonPressed_cb (GtkTreeView *treeview,
 static gboolean gnc_gen_trans_onPopupMenu_cb (GtkTreeView *treeview,
                                               GdkEvent *event,
                                               GNCImportMainMatcher *info);
+static gboolean view_selection_function (GtkTreeSelection *selection,
+                         GtkTreeModel *model,
+                         GtkTreePath *path,
+                         gboolean path_currently_selected,
+                         gpointer data);
 static void refresh_model_row (GNCImportMainMatcher *gui, GtkTreeModel *model,
                                GtkTreeIter *iter, GNCImportTransInfo *info);
 /* end local prototypes */
@@ -306,6 +316,7 @@ gnc_gen_trans_add_toggled_cb (GtkCellRendererToggle *cell_renderer,
     GtkTreeIter iter;
     GNCImportTransInfo *trans_info;
 
+    ENTER("");
     model = gtk_tree_view_get_model (gui->view);
     if (!gtk_tree_model_get_iter_from_string (model, &iter, path))
         return;
@@ -319,8 +330,10 @@ gnc_gen_trans_add_toggled_cb (GtkCellRendererToggle *cell_renderer,
     else
     {
         gnc_import_TransInfo_set_action (trans_info, GNCImport_ADD);
+        gui->add_toggled =TRUE;  //flag A(dd) has just been toggled to view_selection_function
     }
     refresh_model_row (gui, model, &iter, trans_info);
+    LEAVE("");
 }
 
 static void
@@ -331,8 +344,10 @@ gnc_gen_trans_clear_toggled_cb (GtkCellRendererToggle *cell_renderer,
     GtkTreeModel *model;
     GtkTreeIter iter;
     GNCImportTransInfo *trans_info;
-
+    
+    ENTER("");
     model = gtk_tree_view_get_model (gui->view);
+    
     if (!gtk_tree_model_get_iter_from_string (model, &iter, path))
         return;
     gtk_tree_model_get (model, &iter, DOWNLOADED_COL_DATA, &trans_info, -1);
@@ -347,6 +362,7 @@ gnc_gen_trans_clear_toggled_cb (GtkCellRendererToggle *cell_renderer,
         gnc_import_TransInfo_set_action (trans_info, GNCImport_CLEAR);
     }
     refresh_model_row (gui, model, &iter, trans_info);
+    LEAVE("");
 }
 
 static void
@@ -357,8 +373,10 @@ gnc_gen_trans_update_toggled_cb (GtkCellRendererToggle *cell_renderer,
     GtkTreeModel *model;
     GtkTreeIter iter;
     GNCImportTransInfo *trans_info;
-
+    
+    ENTER("");
     model = gtk_tree_view_get_model (gui->view);
+    
     if (!gtk_tree_model_get_iter_from_string (model, &iter, path))
         return;
     gtk_tree_model_get (model, &iter, DOWNLOADED_COL_DATA, &trans_info, -1);
@@ -373,6 +391,7 @@ gnc_gen_trans_update_toggled_cb (GtkCellRendererToggle *cell_renderer,
         gnc_import_TransInfo_set_action (trans_info, GNCImport_UPDATE);
     }
     refresh_model_row (gui, model, &iter, trans_info);
+    LEAVE("");
 }
 
 static void
@@ -457,7 +476,7 @@ gnc_gen_trans_assign_transfer_account_to_selection_cb (GtkMenuItem *menuitem,
     GList *selected_rows, *l;
     gboolean first, is_selection;
 
-    ENTER("");
+    ENTER("assign_transfer_account_to_selection_cb");
     treeview = GTK_TREE_VIEW(info->view);
     model = gtk_tree_view_get_model (treeview);
     selection = gtk_tree_view_get_selection (treeview);
@@ -481,7 +500,7 @@ gnc_gen_trans_assign_transfer_account_to_selection_cb (GtkMenuItem *menuitem,
             DEBUG("passing account value = %s",
                         gnc_account_get_full_name (assigned_account));
             gnc_gen_trans_assign_transfer_account (treeview,
-                           &first, is_selection, l->data, &assigned_account, info);
+                &first, is_selection, l->data, &assigned_account, info);
             DEBUG("returned value of account = %s",
                         gnc_account_get_full_name (assigned_account));
             DEBUG("returned value of first = %s", first ? "true" : "false");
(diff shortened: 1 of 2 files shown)


gnucash/import-export/import-main-matcher.c
```diff
@@ -90,7 +91,10 @@ enum downloaded_cols
 #define CSS_INT_PROB_REQUIRED_CLASS "intervention-probably-required"
 #define CSS_INT_NOT_REQUIRED_CLASS  "intervention-not-required"
 
-static QofLogModule log_module = GNC_MOD_IMPORT;
+/* Define log domain for extended debugging of matcher */
+#define G_MOD_IMPORT_MATCHER "gnc.import-export.main-matcher"
```
Bob-IT: Why define it, it is only used here, I may be wrong but why not just use...
static QofLogModule log_module = "gnc.import.main-matcher";
Also I think it should be just 'import'.

gnucash/import-export/import-main-matcher.c
```diff
@@ -117,10 +121,22 @@ static gboolean gnc_gen_trans_onButtonPressed_cb (GtkTreeView *treeview,
 static gboolean gnc_gen_trans_onPopupMenu_cb (GtkTreeView *treeview,
                                               GdkEvent *event,
                                               GNCImportMainMatcher *info);
+gboolean view_selection_function (GtkTreeSelection *selection,
```
Bob-IT: This should be 'static' I think

gnucash/import-export/import-main-matcher.c
```diff
@@ -117,10 +121,22 @@ static gboolean gnc_gen_trans_onButtonPressed_cb (GtkTreeView *treeview,
...
 static void refresh_model_row (GNCImportMainMatcher *gui, GtkTreeModel *model,
                                GtkTreeIter *iter, GNCImportTransInfo *info);
 /* end local prototypes */
 
+/* define the add_toggled gboolean flag for use by the vie-selection_funcc-tion
+   to avoid a row being selected immediately after the row's Add checkbox has 
+   been toggled on. This casues a problem if the U+R or R checkboxes are toggled
+   without the row being unselected
+*/
+gboolean add_toggled = FALSE;
```
Bob-IT: I probably would not do this, instead I would add the gboolean to _main_matcher_info with a suitable comment and then you will need to add the info structure to the function setup and access it from there...

    gtk_tree_selection_set_select_function
                               (selection,
                                view_selection_function,
                                info,
                                NULL);

gnucash/import-export/import-main-matcher.c
```diff
@@ -613,14 +638,16 @@ gnc_gen_trans_onPopupMenu_cb (GtkTreeView *treeview,
...
 {
     GtkTreeSelection *selection;
-    ENTER("");
+    ENTER("onPopupMenu_cb");
     /* respond to Shift-F10 popup menu hotkey */
     selection = gtk_tree_view_get_selection (treeview);
     if (gtk_tree_selection_count_selected_rows (selection) > 0)
     {
       gnc_gen_trans_view_popup_menu (treeview, NULL, info);
+      LEAVE ("TRUE");
```
Bob-IT: Can you check you spacing, indents should be 4 spaces.

gnucash/import-export/import-main-matcher.h
```diff
@@ -36,6 +36,15 @@
...
 #include "import-backend.h"
 
+/* global variable to flag that the add checkbox has been toggled so that 
+  the view_select_function for the treeview can negate selection of a row
+  immediately after the checkbox has been toggled. Avoids a situation where
+  if one of the other checkboxes is (U+R or R) toggled, the row is left in
+  a permanently selected state and cannot be unselected unless the Add 
+  checkbox is toggled again
+*/
+extern gboolean add_toggled;
```
Bob-IT: I do not think you need this, not certain but if you add it to info as suggested then you definitely wont.

Bob-IT:
I am wondering if the commits that show up before the main commit will cause a problem.
I normally start with an up to date maint branch, created a new branch from that 'Bugxxxx', make my changes and then I would git fetch --all and git rebase upstream/maint to make sure my working branch is up to date with maint.

jralls:
> I am wondering if the commits that show up before the main commit will cause a problem.

Indeed. This branch isn't mergeable.
@DaveC49 you have commits named "merge remote tracking branch Gnucash/maint into maint". Don't do that, it creates weird extra commit paths in history. Always rebase your local branch on or reset it to the mainline repository.

The very first commit, "Put the version back to 3.900" is a master commit and shouldn't be in your maint branch.

To clean this mess up, do this in your local repo:
```
git fetch
git checkout maint
git reset --hard Gnucash/maint
git branch -m Bug797275 Bug797275-old
git checkout -b Bug797275
git cherry-pick Bug797275-old
git branch -D Bug797275-old
git push -f davec49 Bug797275 #Substitute your actual remote name for "davec49"
git checkout maint
git push -f davec49 maint
```

gnucash/import-export/import-main-matcher.c
DaveC49: OK hadn't thought of that. I'll give that a go. I was not happy about putting a global in.

gnucash/import-export/import-main-matcher.h
DaveC49: agreed I'll try that out

gnucash/import-export/import-main-matcher.c
DaveC49: Will do Bob

gnucash/import-export/import-main-matcher.c
DaveC49: I was just wanting debug information from the main-matcher and not from the whole import process. I have changed it back to gnc.import.main-matcher so that i can set just the matcher to provide debug information while running
One thing I am unsure of with the log domain definitions is whether if gnc.import and gnc.import.main-matcher were defined and the log-level for gnc.import was set as debug it would automatically include any subdomains at that log-level.  I will research thiswhen I have a bit of time

gnucash/import-export/import-main-matcher.c
DaveC49: Agreed

DaveC49:
Added all the fixes above with the add_toggled flag in the _main_matcher_info structure as Bob suggested, fixed indenting.

David

gnucash/import-export/import-main-matcher.c
gjanssens: > I was just wanting debug information from the main-matcher and not from the whole import process. I have changed it back to gnc.import.main-matcher so that i can set just the matcher to provide debug information while running
> One thing I am unsure of with the log domain definitions is whether if gnc.import and gnc.import.main-matcher were defined and the log-level for gnc.import was set as debug it would automatically include any subdomains at that log-level. I will research thiswhen I have a bit of time

I believe subdomains will inherit log levels from their parent domains.

DaveC49: Thanks Geert, I thought that might be the case and was working on that assumption.

gnucash/import-export/import-main-matcher.c
DaveC49: Bob Worked like a charm

Bob-IT:
David,
Thanks for the update, can you please squash the second commit so it is all part of the first one.
git rebase -i with the commit to start from.
On that subject, I have pushed the correction to my previous double-click correction which was part of this change just in case this PR is not finished in time for 3.6 so please rebase against the current maint branch first.

I have tried this change and I think it is almost there, there are still some 'spaces' that are not correct like blank lines with spaces on them and indentation like at 616, 637 and 818. Not sure what editor you use, I use Geany which has the option to strip spaces from lines or mark line endings.

When using the matcher, it starts of with no lines selected, which is OK and you can not right mouse any where. Double click works as before and you can only select 'Add' lines which I think is also OK. Now if you select one, you can right mouse on any line which I do not think is correct, I would of thought it would be only the selected lines that could be right moused. Also was wondering if the red text at the bottom and/or help dialogue might need a reference to the multiselection.

DaveC49:
Bob,
I've squashed the commits after rebasing on GnuCash/gnucash/maint and pushed them back to my repo. The right mouse click to bring up the popup menu can happen anywhere on the screen. It doesn't perform any actions on the rows selected or any other row which the cursor might be on at all. I can make that clear in a change to the documentationwhich is already in place since the 3.5 release. I have wondered about how it make it clear that multiselection is enabled. I'm not familiar with the code around the status bar so that might take me a while to sort out. If you are happy with it as it is now, I will look into how to indicate that multiple selection is enabled for the next release. There is really no place for a tooltip that I can see.
David

DaveC49:
Bob,

I'm using the Linux xed editor at the moment. It also gets used for modifying shell scripts etc. which all use the 2 space tabs.  I have used eclipse in the past but found it  did more than I wanted. I might give it or Geany a go so I don't have to keep swapping the editor preferences when editing C++ code.

DaveC49:
Bob , 
I've now addeed a section on multiple selection at the bottom of the help dialog poup. Also ammended the right click popup menu messaage to Assign a transfer account to selection. i did this in a separate branch which I then merged into Bug797275. rebased it again to Gnucash/gnucash/maint just in case there had been any changes and pushed it to my repo.

christopherlam:
> I'm using the Linux xed editor at the moment. It also gets used for modifying shell scripts etc. which all use the 2 space tabs. I have used eclipse in the past but found it did more than I wanted. I might give it or Geany a go so I don't have to keep swapping the editor preferences when editing C++ code.

It's worth while learning Emacs which has numerous useful packages; the most unique one is magit which makes short work of any git task.

Sashachua.com has some cartoon cheat sheets

DaveC49:
Thanks Chris, I have used emacs in the past in the days before mice. I've taken a look at it a couple of times over the last couple of years, but having been weaned off keyboard shortcuts, I find it difficult to go back to learning them all again. It has certainly come a long way with a pretty good menu interface now.   I'm using Gitkraken as a git interface at the moment.  Now that I've worked out how to do rebasing properly with it, I quite like it. The visual display of the commit history is quite good. It has some nice features for interactive merging which I found utterly confusing when using git at the command line. Like all programs you have to invest the time to get the best out of them.

jralls:
I'm also a long-time (as in 30+ years) emacs user, and while it's possible to do a lot of things from the menus it's still very much optimized for touch-typing and no mouse. I find that a feature, but it may well be because it's now hard-wired in my fingers.
Microsoft's [Visual Studio Code](https://code.visualstudio.com/) is available for Linux and has gotten a lot of traction.

DaveC49:
I am taking another look at emacs and Geany. Eclipse was good but I think qualified as bloat ware. Given emacs was 100M to install that may be a moot point. My preference is for mouse driven interfaces, particularly as I am now aging and remembering what I went into the next room for is sometimes a challenge, let alone remembering a large number of hot key combinations, so a menu is a great comfort. I only have to remember what the menu items do.

gjanssens:
Well as we're apparently on a DE promo tour, I have switched from Eclipse to KDevelop a few years back. Compared to a decent text editor you could also consider it "bloated". If you're only interested in a good text editor, Kate would be the more lightweight version.
What I like about kdevelop is it's checking as you type feature based on LLVM. So while I'm typing my code it will hint me at possible typos and syntax errors. Because it does inline compilation it can also assist me when I quickly need to know the type of a variable or a function signature. It shows me all that info while hovering the mouse over any word. From there it has direct links to related code parts (function declaration->definition, list of locations where a function is used,...). In short lots of handy navigation support.

When I started my professional IT path, I learned emacs as well. After that first job I lost track of it (it didn't fit in the new work environment), so by now I have lost my muscle memory as well. And I don't remember how well it was in quick navigation and other coding support.

Anyway, just another suggestion.

At the same time I'm also curious at what Gnome's Builder is capable of. It has received a lot of attention and has a few promising features (like tight integration with gperf). It's on my list for when-time-permits.

jralls:
What does KDevelop do with Scheme? I think @christopherlam's adoption of emacs was because emacs can format Scheme out of the box and there's an add-on that can interpret Scheme in a buffer. 
(Not that it would make any difference to @DaveC49!)

gjanssens:
I did indeed target this at @DaveC49 , ignoring scheme. KDevelop can do syntax highlighting for scheme code and does so quite nicely. It doesn't have a guile interpreter built in though. So for my guile work it's sufficient, it my not be for @christopherlam where guile is the main language.

DaveC49:
KDevelop does sound nice with the typo and syntax checking as you type which may be handy given my novice status with C++. Builder also looks interesting as well.  Too much choice

Bob-IT:
@DaveC49 Thank you for the update, have pushed this to maint, sorry it missed 3.6 but as always other stuff gets in the way.

DaveC49:
No drama Bob.  It works as expected without that commit anyway. What is obvious to you when you are writing the code is not necessarily obvious to general users so I am sure to have a few more minor improvements before the next release.

PR #531, opened 2019-06-30
fix master maint merge and reports

christopherlam:
a couple mishaps during recent work

@gjanssens is the intention that reports need to `(use-modules (gnucash reports))`? without this line, customer-summary and owner-report crash. i'm not sure why other reports don't need it.

### DIFF
--- a/gnucash/report/reports/reports.scm
+++ b/gnucash/report/reports/reports.scm
@@ -26,6 +26,7 @@
 
 (define-module (gnucash reports))
 (use-modules (srfi srfi-13))
+(use-modules (srfi srfi-8))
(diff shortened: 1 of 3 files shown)


gjanssens:
That's not intentional. I think it's an artefact from business-reports.scm so you probably need it for former business reports.
What's the guile backtrace you get when they crash ?

christopherlam:
It reports a regular "variable not defined" error or something. If the fix is acceptable then I'll merge; if you prefer another fix to promote `gnc:owner-report-create` then go ahead...

gjanssens:
At this point that fix is fine. I don't have a better idea ATM and not really time to search for one either...

PR #533, opened 2019-07-06
[budget] upgrade to show YTD values

christopherlam:
Merges ideas from Phil Longstaff's ytd-budget.scm report

Copied from #532 which had become messy.

- [ ] awaiting feedback on budget-period headers -- it's rather crowded; at most it will print bgt / act / diff / bgt(ytd) / act(ytd) values

Screenshot:
image

### DIFF
--- a/gnucash/report/standard-reports/budget.scm
+++ b/gnucash/report/standard-reports/budget.scm
@@ -55,6 +55,8 @@
 (define opthelp-show-actual (N_ "Display a column for the actual values."))
 (define optname-show-difference (N_ "Show Difference"))
 (define opthelp-show-difference (N_ "Display the difference as budget - actual."))
+(define optname-use-envelope (N_ "Use envelope budgeting"))
+(define opthelp-use-envelope (N_ "Values are accumulated across periods."))
 (define optname-show-totalcol (N_ "Show Column with Totals"))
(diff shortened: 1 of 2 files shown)


christopherlam:
As an alternative to adding more headers causing clutter, I think it would be better to switch Bgt / Act headers to Bgt (YTD) / Act (YTD) headers when option is toggled? I think the Diff header will still be fine.

christopherlam:
Ok I've changed this branch -- option "Display" "Use envelope budgeting" will, instead of period bgt/act/diff amounts, show the accumulated bgt/act/diff amounts. I think this is the simplest upgrade to budget chart. See screenshots.

ss1
ss2

christopherlam:
Unless there are objections I'll merge soon.

Removed options, never used:
* general / show-full-account-name?
* general / price-source

Add option:
* general / use-envelope-budgeting? -- changes budgeting strategy from period-only budgeting to accumulated amounts from the start of budget. Test added.

christopherlam:
@gjanssens this requires your attention as well, it's already merged -- ec28835d7 adds 'ytd' budget i.e. unused budgeted amounts carry over across periods. What option name to use? I'm not too sure 
'use envelope budgeting' is accurate.

PR #534, opened 2019-07-07
Wrapqoflog

christopherlam:
Figuring out how to wrap qoflog.h

Still unsure how to check appropriate log domain is debug-enabled.

Checking "gnc" will be too noisy; checking "gnc.reports" means gnc:debug isn't working in libgnucash/scm


jralls:
Works now, and cuts the time to run tests by almost half. Merge & push whenever you're ready.

PR #535, opened 2019-07-07
Unify job&owner-report, display more metadata...

christopherlam:
This PR aims to merge `job-report.scm` into `owner-report.scm`.

Some handling is necessary -- job always map into 1 customer/vendor, but can map into multiple invoices/bills, all using originating original invoice/bill currency.

The owner-report aims to limit search of invoices/bills into 1 AP/AR account (thereby limiting currency)... I think it could be a reasonable upgrade, later on, to remove the account option, thereby searching *all* APAR accounts and supporting multiple currencies.


christopherlam:
Candidate commit to supercharge owner-report

Various types of invoices / payments (full, multi-invoice "1 txn for 3 pmts", partial-payments "inv-13")

image

christopherlam:
Supercharged `owner-report.scm`, redux.

Invoice lines show associated payment details + any outstanding amount, 1 in each line.
Payment lines show associated invoices paid, 1 in each line.
Which format is nicer?

image

christopherlam:
A lot of inlining single-caller functions have been done to reduce complexity and reduce reliance on `set!` calls -- nothing wrong with them, but these can modify variables out-of-scope and makes code unreadable; I'll respin them into separate functions with much better documentation.

With regards to removing AP/AR account option -- I have some doubts as well... Doesn't seem to make a lot of sense to request customer/vendor/employee/job, and also limit to one APAR account. We could limit the search to AP or AR (or both, for jobs) only according to owner-type. The only reason to enable an account option is to satisfy the register hook (i.e. from register APAR acct, click `Report > Account Report - Single Transaction`) which sends the originating account.... according to my branch, the originating account is then ignored, and creates split's owner's owner-report.

jralls:
I'm on board with checking all e.g. AP accounts for a customer report, I just wonder if it makes sense to check AR accounts as well since in theory a customer wouldn't have any transactions in them. OTOH if the code is making a single pass through the accounts container and just using a broader criterion for selecting it doesn't matter at all; if it's making two passes, one for AP and one for AR, that's a significant performance hit. Traversing hash tables is expensive.

christopherlam:
I can probably refactor the query so that only 1 is done (with all APAR accounts), then handle each account separately. IMHO It would be useful to highlight if a customer invoice split lives in an AP account.

christopherlam:
Ok further work done, should be more understandable now. Feedback on option or column labels welcome.

This branch will remove `gnc:make-account-sel-limited-option` -- the single-account selector in options. `owner-report.scm` won't need it anymore. `receivables.scm` and `payables.scm` will be next.

`gnc:make-account-sel-limited-option` is buggy somewhere in C -- see https://bugs.gnucash.org/show_bug.cgi?id=797158

christopherlam:
With regards to showing all AP/AR for owner-report, I think showing different tables for each account is better than one table for all. See experiment below. I'll leave multiple tables.

image

gjanssens:
I haven't had a chance for a more detailed review yet. Just a few initial comments:
* I think this change is big enough to go into master rather than maint ?
* The earlier screenshot shows extra columns with details ("1 invoice, 3 payments",...), the later one doesn't. I seem to recall a discussion with Derek about this, but there are no comments on this PR reflecting that. Can you point to that discussion or add a summary for completeness?
* I would prefer something in between the one table per AR and one single table for all. In particular I like it when the columns line up across tables. It's a bit messy to look at the whole report if they don't. Perhaps you can stick to the one table for all approach but apply some smart styling to give it the look of the one table per AR layout ? In particular, remove borders on rows that were not in rows in the earlier example, make the AR description rows render like titles,...

christopherlam:
>     * I think this change is big enough to go into master rather than maint ?

I don't think there are many incompatibility changes here. An extra option maybe but other doesn't actually need merge into master. We can decide later where to merge.

>     * The earlier screenshot shows extra columns with details ("1 invoice, 3 payments",...), the later one doesn't. I seem to recall a discussion with Derek about this, but there are no comments on this PR reflecting that. Can you point to that discussion or add a summary for completeness?

Sure it's https://code.gnucash.org/logs/2019/07/07.html onwards

>     * I would prefer something in between the one table per AR and one single table for all. In particular I like it when the columns line up across tables. It's a bit messy to look at the whole report if they don't. Perhaps you can stick to the one table for all approach but apply some smart styling to give it the look of the one table per AR layout ? In particular, remove borders on rows that were not in rows in the earlier example, make the AR description rows render like titles,...

This was another cosmetic idea I had.

The main sticking point I have is the *aging table* -- how should it handle invoices/creditnotes especially cases whereby these fall into different buckets, and are yet unattached to each other? Or overpayment cases? Note aging.scm and owner-report.scm handle them very differently.

i.e. invoice due 60-90 days ago $100, customer-note offered 30-60 days ago $80, both unlinked
or invoice due $100, payment $200.

christopherlam:
> I would prefer something in between the one table per AR and one single table for all. In particular I like it when the columns line up across tables. It's a bit messy to look at the whole report if they don't. Perhaps you can stick to the one table for all approach but apply some smart styling to give it the look of the one table per AR layout ? In particular, remove borders on rows that were not in rows in the earlier example, make the AR description rows render like titles,...

I've been experimenting using only 1 html-table for all.

The difficulty is: (1) to hide left&right bordes for the APAR account rows, while maintaining border=1 for the whole html-table is quite difficult. (2) using h3 for APAR account row leads to alignment issues because there's padding underneath h3. But I also like having columns aligned, so, I'm happy to show 1 table for all.

I've anticipated @gjanssens feature request to hide APAR account row if only 1 is used. So, most books with only 1 APAR account will not have ugly H3 rows.

See screenshots: (1) customer with invoices in only 1 account (2) in multiple accounts

image
image

christopherlam:
Merged in as commit d893723479e9c63a9288fe56834d78a3b65ffe64

PR #536, opened 2019-07-07
[Master] Create a New properties dialogue

Bob-IT:
This was a bit of test and to see whether it was doable and potentially for a couple of bugs with a bit more work. It is based on the preference dialogue and so looks similar which I think is a good thing. If the requirement is for a static dialogue, then more code could be removed that dealt with the adding of pages and options.

I have created items for all existing properties and stubs in the source for other widgets when required.
I have added a new menu option, 'Properties New' so they can be compared for testing.

With the possible move of some preferences to properties hopefully this makes it simpler.
Will await feedback.

### DIFF
--- a/gnucash/gnome-utils/dialog-properties.c
+++ b/gnucash/gnome-utils/dialog-properties.c
@@ -1,0 +1,1740 @@
+/********************************************************************\
+ * dialog-properties.c -- properties dialog                         *
+ *                                                                  *
+ * Copyright (C) 2005 David Hampton                                 *
+ * Copyright (C) 2013 Geert Janssens                                *
+ * Copyright (C) 2019 Robert Fewell                                 *
+ *                                                                  *
+ * This program is free software; you can redistribute it and/or    *
+ * modify it under the terms of the GNU General Public License as   *
+ * published by the Free Software Foundation; either version 2 of   *
+ * the License, or (at your option) any later version.              *
+ *                                                                  *
+ * This program is distributed in the hope that it will be useful,  *
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of   *
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the    *
+ * GNU General Public License for more details.                     *
+ *                                                                  *
+ * You should have received a copy of the GNU General Public License*
+ * along with this program; if not, contact:                        *
+ *                                                                  *
+ * Free Software Foundation           Voice:  +1-617-542-5942       *
+ * 51 Franklin Street, Fifth Floor    Fax:    +1-617-542-2652       *
+ * Boston, MA  02110-1301,  USA       gnu@gnu.org                   *
+\********************************************************************/
+
+/** @addtogroup Dialogs
+    @{ */
+/** @addtogroup PropDialog Properties Dialog
+    @{ */
+/** @file dialog-properties.c
+    @brief Dialog for handling Gnucash file properties.
+    @author Copyright (c) 2019 Robert Fewell
+
+    These functions are the external API available for the
+    properties dialog. This dialog allows a user to modify
+    several Gnucash properties.
+    Any module may add a page (or partial page) of properties
+    to the dialog.  These additions are done by providing
+    the name of a glade file and the content to load from that
+    file along with a widget in that file.  If a partial
+    page is added, the widget name provided must be that of
+    a GtkGrid containing four columns. If a full page is added,
+    the widget name provided to this code can be any kind of
+    widget, but for consistency it should probably be the same.
+
+    The argument *is* a glade file, so if your code has special
+    requirements (e.g. make one widget insensitive until another is
+    selected) feel free to go ahead and add your own callbacks to the
+    glade file.  This code will connect any callbacks that exist in
+    the glade file.
+*/
+
+#include <config.h>
+#include <gtk/gtk.h>
+#include <glib/gi18n.h>
+#include "dialog-utils.h"
+#include "gnc-gobject-utils.h"
+#include "gnc-engine.h"
+
+#include "gnc-ui.h"
+#include "gnc-ui-util.h"
+#include "gnc-component-manager.h"
+#include "dialog-properties.h"
+
+#include "gncTaxTable.h"
+#include "gnc-date-format.h"
+#include "gnc-tree-model-budget.h"
+#include "gnc-currency-edit.h"
+#include "gnc-date-edit.h"
+#include "gnc-period-select.h"
+
+
+#define DIALOG_PROPERTIES_CM_CLASS  "dialog-properties"
+#define GNC_PREFS_GROUP             "dialogs.properties"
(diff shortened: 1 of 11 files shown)


gnucash/gnome-utils/dialog-properties.c
```diff
@@ -0,0 +1,1702 @@
...
+/** Connect a GtkFontButton widget to its properties value.
+ *
+ *  @internal
+ *
+ *  @param fb A pointer to the font button that should be connected.
+ */
+static void
+gnc_props_connect_font_button (GtkFontButton *fb)
+{
+    gchar *path;
```
jralls: Init to NULL. This (and similar in other functions, `free` at lines 688, 964, 1029, 1049) causes a "might be used uninitialized warning and fails to build. That's the current cause of the test failures.

jralls:
Where's the connection to the options API?

gnucash/gnome-utils/dialog-properties.c
Bob-IT: Sorry about that, missed this as travis is not running on my master, need to look into that.

Bob-IT:
The only connection to the options API is the KVP path you specify for the GTK control which is similar to that for the preference which is the gsetting path.

gnucash/gnome-utils/dialog-properties.c
```diff
@@ -0,0 +1,1702 @@
...
+
+    gnc_props_split_widget_name (gtk_buildable_get_name(GTK_BUILDABLE(spin)), &path);
+
+    val = qof_book_get_int64_option (gnc_get_current_book(), path);
+
+    // some spin values are doubles
+    if (val == 0)
+        vald = qof_book_get_double_option (gnc_get_current_book(), path);
+
+    if ((vald > 0) || g_strcmp0 (path, "options/Accounts/Day Threshold for Read-Only Transactions (red line)") == 0)
```
jralls: Make a separate `gnc_props_connect_spin_button_double`. Hard-coding the path here is asking for maintenance trouble later.

jralls:
That won't work. The option values are read from hashtables that are loaded from KVP at startup and saved during a file save. Changes made with your dialog won't have any effect and will be overwritten when the file is saved.

gjanssens:
While this is an interesting experiment, I have a few concerns with it.

1. As you write it's based on the preferences code. Without having made an in-depth comparison, it seems to me the only real differences are which pages to inject and the callback handling. Wouldn't it have been better to factor out those relatively limited areas of code instead of copying large chunks into separate files ?
2. My other concern is that conceptually book properties are exactly that: properties of the book object. The book object is defined in qofbook.c/h. In a decent OO design (granted we're not there yet) I would expect a book's properties to be defined with the object. In this experiment however the property definitions are now defined in a glade interface file. Whatever gets defined there will become a book property. It should be that qofbook defines the properties and if any code tries to get or set a property not defined there, it should give an error.

We could combine strongly defined book properties with a manually crafted gui as you have now created. The advantage is more tight control over the dialog layout. The disadvantage is more manual work in case properties are added/removed. For the book in particular this may still be feasible. It will be more difficult for reports that currently use the same options dialog code as the book options.

gnucash/gnome-utils/dialog-properties.c
Bob-IT: I realised that but could not think of an alternative till now by specifying double in the glade file widget name.

Bob-IT:
@gjanssens 
1. Some times it is just easier to copy stuff and see where it leads and then go back and see what actually was required to change and yes this could probably be refactored with the preference stuff sharing a common util file with all the glade specific stuff. Also this may not lead any where so I just like to keep it separate.

2. That sounds great but that is probably beyond my knowledge.

This was purely to do with the 'File->Properties' menu item that happens to use the 'options' system and not a use for reports which is why I was confused by Johns last statement as every thing I tried did work like having an invoice report open, change the Business details, refresh the report and the details were updated.

Any way as I said, just an experiment, will close later.

gjanssens:
> @gjanssens
> 
>     1. Some times it is just easier to copy stuff and see where it leads and then go back and see what actually was required to change and yes this could probably be refactored with the preference stuff sharing a common util file with all the glade specific stuff. Also this may not lead any where so I just like to keep it separate.
> 

Fair point.

>     2. That sounds great but that is probably beyond my knowledge.
> 
> 
> This was purely to do with the 'File->Properties' menu item that happens to use the 'options' system and not a use for reports which is why I was confused by Johns last statement as every thing I tried did work like having an invoice report open, change the Business details, refresh the report and the details were updated.
> 
> Any way as I said, just an experiment, will close later.

I appreciate you did this proof of concept. Transforming the options away from the odd guile/C mix they are in now is an important work. And there is at least one interesting benefit to your approach: the better control over the layout of the options dialog. So if you close the PR, don't throw out your branch just yet. We may come back to it in the future when the options conversion has progressed further.

Bob-IT:
@gjanssens With my very limited understanding, what you are asking for is say, two qofbook functions, a getter and setter so the getter would retrieve all information and the setter would set it. So you could have gc_book_get_property ("prop/options/read-only threshold") that would return a structure with a GValue for the value and gchar for tooltip, you may want limit values for numbers also. That function could get the required information from a static array. The glade file would still control what was displayed and where but if the prop is not defined it would be disabled.

Bob-IT:
Closing as not required.

gjanssens:
> @gjanssens With my very limited understanding, what you are asking for is say, two qofbook functions, a getter and setter so the getter would retrieve all information and the setter would set it. So you could have gc_book_get_property ("prop/options/read-only threshold") that would return a structure with a GValue for the value and gchar for tooltip, you may want limit values for numbers also. That function could get the required information from a static array. The glade file would still control what was displayed and where but if the prop is not defined it would be disabled. Not very c++ but it might be something I could do.

I actually have no idea yet how this will eventually be implemented. But the general idea would probably be as you write here. In C++ it will likely be implemented using boost::variant rather than GValue or we may even try to add some strong typing in there. It's too soon to tell (and I've been away for a while so my memory is a bit foggy still).

jralls:
And besides, I'm the one [working on it](https://github.com/jralls/gnucash/tree/c++options). ;-)

`gnc_book_get_property` will get you to the KVP, but that's only going to have the value of the option and in some cases (mostly having to do with reports) only if it's different from the default for the option.  The current design is driven by the report system's need to create an arbitrary options dialog from Scheme for each report. You'll see from [gnc-option.hpp](https://github.com/jralls/gnucash/blob/c%2B%2https://github.com/jralls/gnucash/blob/c%2B%2Boptions/libgnucash/app-utils/gnc-optiondb.hppBoptions/libgnucash/app-utils/gnc-option.hpp) and [gnc-optiondb.hpp](https://github.com/jralls/gnucash/blob/c%2B%2Boptions/libgnucash/app-utils/gnc-optiondb.hpp) that my current thinking is to compose that functionality of successive wrapper classes. It's still a bit tentative because the Scheme design does a very poor job of separating the pieces and an insanely large API (nearly 200 public functions).

PR #538, opened 2019-07-19
Refactor aging report

christopherlam:
Rebuild aging report from the ground up. Much easier using srfi-1. I can further break out larger functions for greater clarity. Some tricks:
* one query for the whole AP/AR query all splits, filtered for txn-type-invoice/payment types only
* all above analysed to obtain list of owners
* for-each owner, whole splitlist is `partition`ed into splits-belonging-to-owner and other-splits
  * owner-splits are analysed via lots to create aging-table
  * other-splits are then recycled to scan the next owner.
* all AP or AR accounts are scanned, which creates report with subtotals with all their currencies.

Removed features
* APAR-account selection
* sorting now works on account-name only; removed sort by aging-table or amount
* "show multicurrency subtotals" which had never worked properly.
* printing of owner address

Pending
* [ ] Clarification on overpayments & credit-notes -- how should overpayments be shown on aging-table? Should credit-notes automatically be applied to reduce invoices?
* Reinstate printing of owner address -- is this important?

Result
image

### DIFF
--- a/gnucash/report/business-reports/aging.scm
+++ b/gnucash/report/business-reports/aging.scm
@@ -26,6 +26,9 @@
 
 (define-module (gnucash report aging))
 
+(use-modules (srfi srfi-1))
+(use-modules (srfi srfi-9))             ;record
+(use-modules (ice-9 receive))
 (use-modules (gnucash utilities))
 (use-modules (gnucash gnc-module))
 (use-modules (gnucash gettext))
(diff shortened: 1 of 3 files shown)


christopherlam:
P.S. this report had a much more extensive change than #535 because it was so complex I basically rebuilt from the ground up using the options object. Still pending a verdict on the aging table and subsequent debugging.

jralls:
Nice that it cuts the code size in half. `aging-renderer` is still pretty ugly, but progress is best made incrementally.

christopherlam:
Merged into maint as a new pair of reports in the experimental submenu.

PR #539, opened 2019-07-19
Bug797301

Bob-IT:
This is a two part fix, the first commit adds a fix for the menu View->refresh that only worked on the report pages, the other pages did not connect to the menu item so for the register and budget they are reloaded and for other pages a GTK queue draw is requested.
The second fixes when in a sub-account register and you create another account which is also a sub-account but the register does not update. To fix this the number of sub-accounts is tracked and if it is different on the refresh the query is recreated as opposed to just rerun.

The error cancelling commit fixes when you are asked whether you want to create a new account in the register, possibly due to a typo and you answer no, the same dialogue will appear another three times before resetting the account cell to the original value. This change eliminate that by returning you back to the cell with the invalid entry so that you can amend / cancel or use the dialogue again to create a new account based on an amended entry. It also fixes if you create a placeholder account, which brings up a warning dialogue but currently will still allow the transaction to be added to the placeholder.

I have tried this on Linux and Windows and I think it works as expected but would appreciate if some else gave it a spin before I push it.


jralls:
Looks fine to me.

Bob-IT:
Thanks John, will push later.

PR #542, opened 2019-07-28
[Master] Remove preference callbacks on quit / window close

Bob-IT:
This is a follow up for the blocking of registered preference callbacks to remove the callbacks on GncWindow close and quit.
The first few commits are straightforward adding functions to remove the preference callbacks setup in various source files.
There are two 'register' preference callbacks that are setup with a 'g_once' that are used till quit so the easiest way to remove them was to save the preference id and use that on quit.
With these changes it took the remaining callbacks from about 50 to 3 on quit.
These last few are setup via plugins so there is a couple of commits dealing with that with the final commit destroying the hash table when there are no callbacks left.

I have tested this by quitting with multiple windows open and closing windows in different orders and all preference callbacks are removed. Will leave here for feedback and push end of week if all OK.

### DIFF
--- a/gnucash/gnome-utils/account-quickfill.c
+++ b/gnucash/gnome-utils/account-quickfill.c
@@ -65,7 +65,7 @@ static void
 shared_quickfill_destroy (QofBook *book, gpointer key, gpointer user_data)
 {
     QFB *qfb = user_data;
-    gnc_prefs_remove_cb_by_func (GNC_PREFS_GROUP_GENERAL_REGISTER,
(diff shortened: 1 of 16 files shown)


jralls:
This looks fine to me but I'm curious: What's the motivation?

Bob-IT:
Thanks John, all stemmed from #409 which was blocking/unblocking the callbacks for when the preference dialogue was loaded. I noticed at the time on my system I still had some hash entries left on exit and Geert suggested that they be found and fixed as appropriate. I started that at the time but did not complete and while reviewing the patches I had saved came across them so thought I would finish it and then move onto some others to complete.

jralls:
Very good.

PR #543, opened 2019-07-29
[master] sample-graphs.scm add tutorial for html-chart

christopherlam:
Add option in `sample-graphs.scm` to show html-chart tutorial.
Is it worth while completing?

image

### DIFF
--- a/gnucash/report/reports/example/sample-graphs.scm
+++ b/gnucash/report/reports/example/sample-graphs.scm
@@ -24,7 +24,8 @@
 ;; new reports for GnuCash.
 
 (define-module (gnucash reports example sample-graphs))
-(use-modules (gnucash utilities)) 
+(use-modules (gnucash utilities))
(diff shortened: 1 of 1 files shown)


jralls:
More documentation is good, but I'm not sure that a report is the right way to do it. It's not really where someone would go looking. What about putting the narrative part in [Custom Reports](https://wiki.gnucash.org/wiki/Custom_Reports) with a pointer to the sample code in Github?

christopherlam:
Fair comment. Some good scheme tricks using `local-eval` in this branch converting data to string and to code :). There's already section in wiki/Custom Reports for html-barchart. Herewith the final tutorial, to be added to wiki:

image

jralls:
That seems a lot clearer than what's there now.

christopherlam:
https://wiki.gnucash.org/wiki/Tutorial_for_html-chart.scm#Tutorial_to_html-chart.scm all done thanks pandoc

PR #545, opened 2019-08-01
Travis CI: Lint Python code for syntax errors and undefined names

cclauss:
3 syntax errors and 3 undefined names...
https://travis-ci.org/Gnucash/gnucash/jobs/566592702#L488

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,4 @@
 # Test
-sudo: required
 language: c++
 compiler: gcc
 env:
@@ -7,6 +6,9 @@ env:
     - BUILDENV=ubuntu-14.04 BUILDTYPE=cmake-make
 services:
     - docker
+before_install:
+    - python3 -m pip install --user flake8
+    - flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
 install:
     - docker --version
     - echo BUILDENV="$BUILDENV"
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,7 @@ env:
 services:
     - docker
 before_install:
+    - python3 -m pip --version || wget -qO- https://bootstrap.pypa.io/get-pip.py | sudo -H python3
     - python3 -m pip install --user flake8
     - flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
 install:
--- a/contrib/gnuc2ooo.py
+++ b/contrib/gnuc2ooo.py
@@ -1,6 +1,6 @@
 #**********************************************************************
 #   Gnu Cash to OpenOffice.org (gnuc2ooo.py)
-#   
+#
 #   Makros for the download of GnuCash Data to OpenOffice.org Base
 #
 #   Makro SetGnuCashFilePaths:
@@ -11,7 +11,7 @@
 #   Makro fillGnuCashDB:
 #       Download data from the GnuCash data source to OpenOffice.org
 #       Base
-#   
+#
 #
 #**********************************************************************
 #   Copyright (c) 2008 Knut Gerwens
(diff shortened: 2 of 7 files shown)


.travis.yml
```diff
@@ -1,12 +1,14 @@
 # Test
-sudo: required
```
jralls: Why?

cclauss:
Undefined name 'XSCRIPTCONTEXT' in contrib/gnuc2ooo.py

https://travis-ci.org/Gnucash/gnucash/jobs/566603018#L472

$ __flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics__
```
./contrib/gnuc2ooo.py:573:19: F821 undefined name 'XSCRIPTCONTEXT'
            doc = XSCRIPTCONTEXT.getDocument()
                  ^
1     F821 undefined name 'XSCRIPTCONTEXT'
1
```
@gjanssens

.travis.yml
```diff
@@ -1,12 +1,14 @@
...
 language: c++
 compiler: gcc
 env:
     - BUILDENV=arch BUILDTYPE=cmake-ninja
     - BUILDENV=ubuntu-14.04 BUILDTYPE=cmake-make
 services:
     - docker
+before_install:
+    - python3 -m pip install --user flake8
```
jralls: flake8 should be installed using the package manager in the test platform docker files.

.travis.yml
```diff
@@ -1,12 +1,14 @@
...
+    - flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
```
jralls: The actual test should be run in the platform testscript file.

.travis.yml
cclauss: [Travis are now recommending removing the __sudo__ tag](https://blog.travis-ci.com/2018-11-19-required-linux-infrastructure-migration).  Mention  of the __sudo:__ tag has been removed from all Travis docs because the sudo command is now _always_ available and there is no way to turn it off.

jralls: Ok, that should be a separate commit. It's unrelated to linting python.

contrib/gnuc2ooo.py
```diff
@@ -569,22 +569,22 @@ def MessageBox(MsgText, MsgTitle="GnuCash to Ooo", MsgType="messbox", MsgButtons
     global last_input_MessageBox
 ##    last_input_MessageBox = str(MsgText)
 
-    if not ParentWin:
-            doc = XSCRIPTCONTEXT.getDocument()
+    if not ParentWin:  # TODO: Where is XSCRIPTCONTEXT defined?
+            doc = XSCRIPTCONTEXT.getDocument()  # noqa: F821
```
gjanssens: FYI, this script is intended to be used from within LibreOffice as a macro. LibreOffice will set the XSCRIPTCONTEXT variable before the macro is run.

cclauss: Then it is proper to put the statement __global XSCRIPTCONTEXT__ before this line of code.

gjanssens: Considering this script doesn't make sense outside of LibreOffice, I'd rather make it so it exists gracefully if XSCRIPTCONTEXT is not defined. Something like this:

```
try:
    XSCRIPTCONTEXT
except NameError:
    print "This script can't be used standalone. Instead install it as a macro for LibreOffice and run it from there."
    exit
```

cclauss: __print()__ and __exit()__ are both functions in Python 3.

gjanssens: Right. That was just from the top of my head to explain the idea.

PR #546, opened 2019-08-01
Remove the deprecated sudo tag from .travis.yml

cclauss:
As discussed in #545, [Travis are now recommending removing the __sudo__ tag](https://blog.travis-ci.com/2018-11-19-required-linux-infrastructure-migration).  Mention  of the __sudo:__ tag has been removed from all Travis docs because the sudo command is now _always_ available and there is no way to turn it off.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,4 @@
 # Test
-sudo: required
 language: c++
 compiler: gcc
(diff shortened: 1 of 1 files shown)


jralls:
Thanks!

PR #548, opened 2019-08-10
Improve price quotes to be more responsive

christopherlam:
Calling C aficionados.

This is the outcome. See window-title. It shows current F::Q job, and restores the window title when completed.
image

The implementation is, I'm sure, not ideal.

In particular, GtkWindow vs GncWindow. The correct file to add my window-title getter/setters. Alternatively, a multiline GtkTextBox to add the progress notes will also be useful, but I don't know how to do this. Any C style mistakes.

Otherwise, price-quotes.scm is cleaned up a bit.

### DIFF
--- a/libgnucash/scm/price-quotes.scm
+++ b/libgnucash/scm/price-quotes.scm
@@ -78,7 +78,7 @@
 (define gnc:*finance-quote-helper*
   (string-append (gnc-path-get-bindir) "/gnc-fq-helper"))
 
-(define (gnc:fq-get-quotes requests)
+(define (gnc:fq-get-quotes requests window)
   ;; requests should be a list where each item is of the form
   ;;
   ;; (<fq-method> sym sym ...)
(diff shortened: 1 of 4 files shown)


fellen:
I would prefer to use the/a status line or at least `<normal title>: <status>`.

libgnucash/scm/price-quotes.scm
```diff
@@ -151,50 +117,50 @@
...
+      (when quoter
+        (map
+         (lambda (request)
+           (catch 'z
+             (lambda ()
+               (gnc:debug "handling-request: " request)
+               ;; we need to display the first element (the method,
+               ;; so it won't be quoted) and then write the rest
+               (gnc-window-set-title
+                window (format #f "Retrieving ~a: ~s..."
```
fellen: (_ "Retrieving ~a: ~s...")

jralls:
I agree with @fellen that a status line would be better, though I'm not convinced of the utility:  The messages will go by too fast except in the face of Alphavantage's throttling and just having the redraws will slow down the quote retrieval.

christopherlam:
Ok, back to drawing board. I was aiming for a terminal-type output with ability to cancel process, but this requires multithreading and is too difficult.

PR #552, opened 2019-08-14
Remove GTEST_SRC

gruberchr:
GTEST_SRC, which references the file src/gtest_main.cc within GoogleTest
source code, is used to add this file to the sources of several test
applications. But that's actually not necessary, since this source file
is already compiled into libgtest_main.a (referenced by GTEST_MAIN_LIB,
which is part of GTEST_LIB). As long as GTEST_LIB is linked to the test
application, GTEST_SRC can be omitted from source file list.


gruberchr:
This pull request is replaced by new pull request #555

PR #555, opened 2019-08-15
Rework source file handling of GoogleTest build

gruberchr:


jralls:
Thanks. Not quite what I thought you were going to do but still a simplification.

PR #557, opened 2019-08-17
Bug 797344 - Mixing of preinstalled GoogleTest libraries with headers from source directory

gruberchr:


gruberchr:
By the way, I asked myself, if it would be saver to completely skip search for preinstalled GoogleTest libraries, if variables GTEST_ROOT and GMOCK_ROOT are defined. I guess, if developers set GTEST_ROOT and GMOCK_ROOT, they expect to build GoogleTest from source code and want to see an error, if this was not successful instead of using possibly preinstalled libraries.

What do you think?

jralls:
Makes sense, with some permutations: IIRC you found that your distro provides gtest and not gmock, so one might want to use the built-in gtest and the built-from-source gmock. For that you'd want to define only GMOCK_ROOT and have the search mechanism find the gtest libs. OTOH if you want to build both and since they're in the same repo since 1.8.0 it would be nice to be able to define only GTEST_ROOT and have it find both... but for 1.7.0 it's the reverse because the 1.7.0 gtest tarball didn't have gmock but the gmock tarball did have gtest. Since the current code was written for 1.7.0 defining GMOCK_ROOT only will find gtest if you have the gmock tarball.

gruberchr:
Ok, I guess wording should be clarified first. Since v1.8.0 it becomes quite confusing. I used GoogleTest as a placeholder for both GoogleTest and GoogleMock here. I should be more precise.

> IIRC you found that your distro provides gtest and not gmock, so one might want to use the built-in gtest and the built-from-source gmock.

Do you refer to https://bugs.gnucash.org/show_bug.cgi?id=797342 ? There I used GoogleTest as a placeholder as well. My distro provides both gtest and gmock and both are installed.

But you are right, one could even build only gtest from source and use prebuilt gmock. Haven't thought about that.

> OTOH if you want to build both and since they're in the same repo since 1.8.0 it would be nice to be able to define only GTEST_ROOT and have it find both.

That's a further interesting aspect.

But if we ignore this at the moment my proposal is as follows:
1. If GTEST_ROOT is defined, search for GoogleTest sources and throw an error if not found
1. If GTEST_ROOT is not defined, search for gtest libraries and throw an error if not found
1. If GMOCK_ROOT is defined, search for GoogleMock sources and throw an error if not found
1. If GMOCK_ROOT is not defined, search for gmock libraries and throw an error if not found

gruberchr:
Any comments regarding my proposal?

If not then I have a remark. I saw, that there are further directories searched for GoogleTest and GoogleMock source code, namely /usr/src/gtest and /usr/src/gmock. In my proposal these are now searched only when GTEST_ROOT and GMOCK_ROOT are defined respectively. Does this really make sense? If not what could by a meaningful search strategy?

By the way after looking for preinstalled GoogleTest and GoogleMock sources on my system (Ubuntu 19.04) I found them in /usr/googletest/googletest and /usr/googletest/googlemock (googletest v1.8.1).

jralls:
No, if GTEST/GMOCK_ROOT are set then there should be no need to search at all. Check the path to make sure that the required files are present and error out if not.

The various paths were for finding the installation locations on various Linux distros. They haven't been updated in a while and I'd expect distros to tweak CMake's FindGtest as necessary for their layout.

jralls:
As for `/usr/googletest`, eww.

gruberchr:
> As for `/usr/googletest`, eww. Are you sure it's not `/usr/src/googletest`?

Right, that's a typo. Of course it's `/usr/src/googletest`.

gruberchr:
That means, you would even remove further search paths like `/usr/src/gtest` and `/usr/src/gmock`? Or would you leave it as it is?

jralls:
I'd remove them from the HINTS for finding the paths when GTEST_ROOT or GMOCK_ROOT are set as they'll prevent erroring out if the GTEST/MOCK_ROOT path is wrong but the sought-after files are installed by the system in those places. BTW according to the CMake find_path docs we should be using PATH instead of HINTS.

Since you're not using FindGTest and there is no FindGMock you're relying on find_path and find_library paths to be able to find them on the various CMAKE_SYSTEM_FOO_PATHs. That *should* be OK for the precompiled libraries but IIRC somebody provided sources in /usr/src and precompiled libraries in /usr/lib but not headers in /usr/include so we might need some /usr/src PATH specs for that/those distros... or we could leave it and let them file bugs if it doesn't build for them.

jralls:
As for the 4 cases of your proposal ISTM that if the distro provides sources then it would be better to compile from them unless the user specifies otherwise by setting GTEST/GMOCK_ROOT. So the flow would be for each of gtest and gmock:
* if GFOO_ROOT is defined, check if it's a googlefoo source directory. Use it if it is, error out if not. If only one of GTEST_ROOT and GMOCK_ROOT is defined check to see if the other one is also present in that directory.
* Otherwise look for a system-provided googlefoo source directory. Use it if found.
* Otherwise look for system-provided headers and libs, preferring static to shared if both are present. Use if found.
* Otherwise error.

gjanssens:
> I'd remove them from the HINTS for finding the paths when GTEST_ROOT or GMOCK_ROOT are set as they'll prevent erroring out if the GTEST/MOCK_ROOT path is wrong but the sought-after files are installed by the system in those places. BTW according to the CMake find_path docs we should be using PATH instead of HINTS.
> 
> Since you're not using FindGTest and there is no FindGMock you're relying on find_path and find_library paths to be able to find them on the various CMAKE_SYSTEM_FOO_PATHs. That _should_ be OK for the precompiled libraries but IIRC somebody provided sources in /usr/src and precompiled libraries in /usr/lib but not headers in /usr/include so we might need some /usr/src PATH specs for that/those distros... or we could leave it and let them file bugs if it doesn't build for them.

If we're going to remove these HINTS I would suggest rebasing this PR to master as we're not sure which downstreams we're breaking here. Such unexpected build failures should be avoided in only a minor release. For a major release most packagers are willing to review build issues. For minor releases less so. So let's avoid this kind of frustration now all packagers have more or less figured out how to build the 3.x series for their platforms.

gruberchr:
> If only one of GTEST_ROOT and GMOCK_ROOT is defined check to see if the other one is also present in that directory.

This is not clear to me. Is this meant for GoogleTest >= v1.8.0, where both GoogleTest and GoogleMock are combined in one repository? And should for instance `GTEST_ROOT` be interpreted as the root path to that repository?
In this case I would at least only allow to use `GTEST_ROOT` as root path, but not `GMOCK_ROOT`, because the name of the combined repository is `googletest`.

> BTW according to the CMake find_path docs we should be using PATH instead of HINTS.

Correct, but interestingly I found the following commit 5f53e29, where `PATHS` has been replaced by `HINTS`.

jralls:
Both before and after 1.8.0. Before 1.8.0 tarballs of gmock included gtest. That check is already in place. With 1.8.0 the two projects/repositories were combined into a single one with the structure we're now familiar with. In that case I think we should configure both if:
1. If `GTEST_ROOT` or `GMOCK_ROOT` exist and (representing either as just `ROOT`) `ROOT/googletest` and/or `ROOT/googlemock` exist and contain the needed files.
2. if `GTEST_ROOT` exists and `GTEST_ROOT/../googlemock` exists and contains the gmock files.
3. If `GMOCK_ROOT` exists and `GMOCK_ROOT/../googletest` exists and contains the gtest files.

That's to make it easier on the user and not get carried away with being unnecessarily strict, plus allow just setting `GMOCK_ROOT` to work regardless of the googletest version. 

As for `HINTS` vs. `PATHS`, apparently the distinction based on whether the path in question is computed or absolute is true only for older versions. The latest docs don't make that distinction, but `HINTS` is searched before the system and platform paths while `PATHS` is searched after. I think that's been true since before 3.0. In that light @goodvibes2 is probably right.

gruberchr:
I pushed new commits. I'm not sure, if this is what's in your mind. Have a look.

And I changed using PATHS instead of HINTS with a little amendment. Argument NO_CMAKE_SYSTEM_PATH should solve the problem, that system paths are searched before paths specified with PATHS.

common/cmake_modules/GncAddTest.cmake
```diff
@@ -117,11 +117,14 @@ function(gnc_gtest_configure)
...
-    find_path(GTEST_INCLUDE_DIR gtest/gtest.h
-      HINTS ${GTEST_ROOT}/include ${GMOCK_ROOT}/gtest/include)
     find_path(GTEST_SRC_DIR src/gtest-all.cc
       HINTS ${GTEST_ROOT} ${GMOCK_ROOT}/gtest /usr/src/gtest)
-    if (GTEST_SRC_DIR AND GTEST_INCLUDE_DIR)
+    if (GTEST_SRC_DIR)
+      if (EXISTS ${GTEST_SRC_DIR}/include/gtest/gtest.h)
+        set(GTEST_INCLUDE_DIR ${GTEST_SRC_DIR}/include CACHE PATH "" FORCE)
+      else()
+        message(FATAL_ERROR "GTEST sources found, but no include directory")
```
jralls: Say "doesn't have 'gtest/gtest.h'". It might well have an include directory and telling the user that it doesn't would be confusing and slow down troubleshooting.

common/cmake_modules/GncAddTest.cmake
```diff
@@ -117,27 +117,35 @@ function(gnc_gtest_configure)
   endif()
   unset(GTEST_SRC_DIR CACHE)
   if (GTEST_ROOT)
+    find_path(GTEST_SRC_DIR src/gtest-all.cc NO_CMAKE_SYSTEM_PATH
+      PATHS ${GTEST_ROOT} PATH_SUFFIXES googletest)
+  elseif (GMOCK_ROOT)
+    find_path(GTEST_SRC_DIR src/gtest-all.cc NO_CMAKE_SYSTEM_PATH
+      PATHS ${GMOCK_ROOT}/gtest)
```
jralls: This handles the googletest v1.7 case but not v 1.8+, where we might have `${GMOCK_ROOT}/googletest/src/gtest_all.cc` or `${GMOCK_ROOT}/../googletest/src/gtest_all.cc`.

common/cmake_modules/GncAddTest.cmake
```diff
@@ -147,36 +155,45 @@ function(gnc_gtest_configure)
   message(STATUS "Checking for GMOCK")
   unset(GMOCK_SRC_DIR CACHE)
   if (GMOCK_ROOT)
+    find_path(GMOCK_SRC_DIR src/gmock-all.cc NO_CMAKE_SYSTEM_PATH
+      PATHS ${GMOCK_ROOT})
+  elseif (GTEST_ROOT)
+    find_path(GMOCK_SRC_DIR src/gmock-all.cc NO_CMAKE_SYSTEM_PATH
+      PATHS ${GTEST_ROOT}/googlemock)
```
jralls: or ${GTEST_ROOT}/../googlemock

jralls:
It seems that Arch Linux installs the headers in `/usr/include/gtest` and `/usr/include/gmock`, the `cc` sources in `/usr/src/gtest/src` and `/usr/src/gmock`, and *shared*, not static libraries at `/usr/lib/libgtest.so`, `/usr/lib/libgtest_main.so`, `/usr/lib/gmock.so`, and `/usr/lib/libgmock_main.so`. CMake failed because the C++ sources are separate from the headers.

gjanssens:
> It seems that Arch Linux installs the headers in `/usr/include/gtest` and `/usr/include/gmock`, the `cc` sources in `/usr/src/gtest/src` and `/usr/src/gmock`, and _shared_, not static libraries at `/usr/lib/libgtest.so`, `/usr/lib/libgtest_main.so`, `/usr/lib/gmock.so`, and `/usr/lib/libgmock_main.so`. CMake failed because the C++ sources are separate from the headers.

Again, I think a major rework of the google test/mock detection fits better on master. There have been plenty of tweaks to get it all working on various distros and we risk missing a few of those subtle differences in the rewrite.

In addition, we may be able to simplify a bit if we set the minimum version on master to 1.8 ? Ubuntu 16.04 is the only distro -according to repology- we intended to support that doesn't ship gtest 1.8.x yet. Considering gnucash 4 will release about the same time as Ubuntu 20.04 that would still mean we support 2 Ubuntu LTS releases. Perhaps that's enough also.

jralls:
Requiring Googletest 1.8.0+ would remove one of three paths from line 124. A simplification to be sure but not a big one.

These changes cover a much wider variety of installations than the existing does, but it will indeed break patches that packagers might be using to work around their distro's installation foibles. I don't feel strongly about what branch into which to merge.

I don't understand the comment about supporting two Ubuntu LTS releases. Our CI tests one nearly-obsolete Ubuntu LTS and one constantly-updated bleeding edge distro, Arch Linux. That pretty much ensures that GnuCash is buildable on the nearly-obsolete and everything newer as long as the distro doesn't do something weird with paths that confuses CMake or pkg-config. That would be perverse.

common/cmake_modules/GncAddTest.cmake
gruberchr: Do you really want to allow both GTEST_ROOT and GMOCK_ROOT as root directory for v1.8+, i.e. should they really be equivalent?
Does this make sense? Since v1.8+ the name of the repository is googletest. From my point of view GTEST_ROOT is sufficient for future versions and GMOCK_ROOT is deprecated since v1.8+.

common/cmake_modules/GncAddTest.cmake
gruberchr: Ok, I understand. This should handle the case, that only GTEST_ROOT or GMOCK_ROOT are defined, but not both and GnuCash nevertheless finds both, if they are located next to each other like in googletest repo v1.8+.

gruberchr:
> It seems that Arch Linux installs the headers in `/usr/include/gtest` and `/usr/include/gmock`, the cc sources in `/usr/src/gtest/src` and `/usr/src/gmock`

Ok, that's a case, I didn't have in mind. I'll consider this, but only if sources are found in `/usr/src` to not mix up different headers again. If sources are found in `GTEST_ROOT` or `GMOCK_ROOT` than headers must be there as well.

common/cmake_modules/GncAddTest.cmake
gruberchr: What do you recommend?

gruberchr:
> You're not really reverting commit 5f53e29, you're using `NO_CMAKE_SYSTEM_PATH` instead of using HINTS to change the search order.

Right, I'll extend the commit message.

common/cmake_modules/GncAddTest.cmake
jralls: My reasoning is that the user may have a script written (or a habit formed) for 1.7 where they set GMOCK_ROOT only. Why break it when it's so simple not to? I suppose the counter argument is that a user might want to use the distribution-provided gtest but a separate gmock by setting GMOCK_ROOT and they'd be frustrated because having GMOCK_ROOT set would either find both sets of sources or wouldn't find the googletest sources and so fail.

ISTM the second scenario is less likely.

jralls:
@gruberchr do you have any objection to merging this to master vs. maint in light of @gjanssens comments?

common/cmake_modules/GncAddTest.cmake
gruberchr: At first to avoid misunderstandings, I already considered the case `${GMOCK_ROOT}/../googletest/src/gtest_all.cc` in my last proposal. My question above is only related to the case `${GMOCK_ROOT}/googletest/src/gtest_all.cc`, where `GMOCK_ROOT` is used as the GoogleTest v1.8+ root directory. Maybe this already clarifies your last remark.

>  My reasoning is that the user may have a script written (or a habit formed) for 1.7 where they set GMOCK_ROOT only. Why break it when it's so simple not to? I suppose the counter argument is that a user might want to use the distribution-provided gtest but a separate gmock by setting GMOCK_ROOT

I don't know exactly what case do have in mind. Why should something get broken?

gruberchr:
> @gruberchr do you have any objection to merging this to master vs. maint in light of @gjanssens comments?

I agree with @jralls:

> These changes cover a much wider variety of installations than the existing does, but it will indeed break patches that packagers might be using to work around their distro's installation foibles. I don't feel strongly about what branch into which to merge.

None of the previously used search paths has been removed. The main change is, that these paths are now searched in a well-defined order. I point out that this was the initial idea of this PR to avoid mixing headers and libraries (or even sources), which do not belong together. And the last changes even extend the search paths. 

I don't agree with @jralls
> Requiring Googletest 1.8.0+ would remove one of three paths from line 124. A simplification to be sure but not a big one.

If requiring GoogleTest 1.8.0+ I would simplify much more. I would remove variable `GMOCK_ROOT` completely, which removes many lines and special cases discussed above.

Now it's up to you. But let me at least squash the fixup commits before merging.

common/cmake_modules/GncAddTest.cmake
jralls: Suppose a packager's distro currently provides googlemock 1.7 from the googletest distribution tarball as their gmock source package and so they've written their gnucash build script to pass GMOCK_ROOT=somewhere/gmock and are relying on GnuCash finding the google*test* sources as a subdir. Now their next release is going to have googletest 1.8 with both googletest and googlemock subdirs. It would be kind to not require them to change their GnuCash buildscript, especially in 3.7/3.8. *But* if we decide to apply this to GnuCash 4 (i.e.master), require googletest 1.8 or later , and remove GMOCK_ROOT entirely then it's reasonable to expect packagers to change their scripts as necessary.

jralls:
@gjanssens has acquiesced on IRC to making the GC4 minimum distro Ubuntu 18.04LTS, so let's do that.

common/cmake_modules/GncAddTest.cmake
gruberchr: If only `GMOCK_ROOT` is defined, but not `GTEST_ROOT`, even on current `maint` branch it's not possible that GnuCash finds the GoogleTest sources, when switching to v1.8.
Therefore I don't exactly understand, where this requirement comes from.

gruberchr:
Ok, but this is a new PR and we keep this PR #557 as it is, right?

jralls:
Since @gjanssens objects to #557 going to maint I don't think so. Either close this one and replace it with a new PR or change this one to master and googletest >= 1.8.0.

gruberchr:
I rebased branch `rework_gtest_integration` on `master` and therefore close this PR and open a new one to be merged into `master`.

gjanssens:
Thanks. Feel free to simplify for a gtest 1.8 world only now :)

PR #558, opened 2019-08-18
[master] Remove "Report Title", use "Report name" instead

christopherlam:
Premise: *ALL* reports have, as a minimum, a reportname and a stylesheet. Most reports will use this reportname for the document `<title>` tag or for the chart heading. 10 reports however also define a "Report Title" (or "Title") option, which is then used to populate the `<title>` tag.

Proposal: remove Report Title, and reuse reportname to set document title. No compatibility issues are expected moving forward.

A backward-compatibility path is inserted in `options.scm` to ensure saved-reports pick up the new option-name. Will this file continue to exist in master?

### DIFF
--- a/gnucash/report/reports/standard/account-summary.scm
+++ b/gnucash/report/reports/standard/account-summary.scm
@@ -69,9 +69,6 @@
 (define accsum-reportname (N_ "Account Summary"))
 (define fsts-reportname (N_ "Future Scheduled Transactions Summary"))
 
-(define optname-report-title (N_ "Report Title"))
-(define opthelp-report-title (N_ "Title for this report."))
-
 (define optname-party-name (N_ "Company name"))
 (define opthelp-party-name (N_ "Name of company/individual."))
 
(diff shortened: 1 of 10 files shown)


jralls:
We're getting close to setting feature freeze and the start of 3.9xx releases, so the fate of `options.scm` in GnuCash 4 depends on how much progress I can make on the options rewrite in the next few weeks. I've had less time to work on it than I'd like and the existing code is more convoluted than I realized.

gjanssens:
There may be a compatibility issue: the report name is also used as text for the notebook tab page. Notebook pages (our "tabs") can be renamed via `Actions->Rename Page`. This is true for all page types, not only reports.
For reports in particular this will change the report name option, but not the report title option. So for reports there are two ways to change the name: via

It's not clear to me how this ever came about and got inconsistent over the years. However I can imagine one would want long report titles for certain reports (like for example "Balance sheet for fiscal year 2018-2019"), but a shorter description in the page's tab.

In addition there seems to be a use of report name in saved report configurations.

Superficially I'd think report name should have been/should become a hidden option, that can only indirectly be manipulated via `Actions->Rename Page` or in the saved report configurations code.

Report title on the other hand should hold a value to be actually printed on the report for those reports where it makes sense. Or even better if all reports that require one can have a hard-coded title that works equally well without the need for customization I'd go for that. But I realize this is sensitive territory.

christopherlam:
>     * `Report Name` should be reserved for use outside of the report itself

Don't understand, so, reportname field should be hidden from report options?

gjanssens:
That's indeed what I'm putting up and front to discuss.

christopherlam:
likely to cause numerous complaints

PR #560, opened 2019-08-19
 Complete Hebrew he.po file for v 3.6 

avma:


fellen:
About the templates: There is no region with the abbreviation _HE_.
As long as there is no region specific template, the folder should be simply named **he**.
Then one can use it with LANG=he_IL, he_PL, he_PS...
See https://wiki.gnucash.org/wiki/Locale_Settings for details.

Some users committed region specific templates, mostly based on different tax rules.
So i.e. the business account templates for AT, CH, DE are different for the language de.
Then we create region specific directories.

avma:
I'm not sure I know how to create a new fork without deleting the previous one. If I delete a fork before the PR was actually merged into the target repository will the merge still happen?

gjanssens:
No. In that case the work will be gone.

Are you working from a git repository on your local system or purely on github ?
You don't have to create new forks, only new branches within your fork. Each branch should start from the maint branch though, not stacked on top of each other. If they are stacked, they will effectively end up in the same PR at some point.

avma:
I'm working on Github (uploading from my pc to Github.com)

gjanssens:
Oh, I see my question was ambiguous.
A PR is always created on github.com.
However where do you actually perform your translation work - with an editor on your PC (like poedit, virtaal,...) or on the github webpage ?

Both should be fine as long as you make separate branches for each large body of work. You can have multiple branches in one fork both on github or on your local PC. The way to create branches may be different though depending exactly on which tools you use to manipulate your fork/clone.

PR #562, opened 2019-08-22
Bug797341, Gtk Negative content width warning

Bob-IT:
I was going to push this but decided to see if there were any concerns.

The cause of this warning is when the spacer column of the GNC_TREE_VIEW is reduced to a size that the header button will not fit, So I have removed this column and made the menu select column do the expanding to fill any space. This resulted in the menu being activated on the whole column header so added a button_press_event callback for the 'go-down' image along with a vertical separator which I think looks OK and works.

I have tested this in all the places I think the GNC_TREE_VIEW is used and also LTR. If there are no concerns I will push this at the weekend.


jralls:
Looks good to me.

PR #563, opened 2019-08-22
[Master] Glade changes

Bob-IT:
I created these commits originally for maint but it was decided that they were more appropriate for master. Mainly they are changes to alignment, spacing and with some changes from gtk boxes to use gtk grids. There are also some text changes that will probably cause translation changes.

If no one has any concerns I will push these at the weekend.

### DIFF
--- a/gnucash/gnome-utils/gnc-frequency.c
+++ b/gnucash/gnome-utils/gnc-frequency.c
@@ -229,7 +229,7 @@ gnc_frequency_init(GncFrequency *gf)
     /* Add the new widget to the table. */
     {
         GtkWidget *table = GTK_WIDGET(gtk_builder_get_object (builder, "gncfreq_table"));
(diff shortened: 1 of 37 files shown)


fellen:
Why do you think, it should go to master. I see mostly fixes of our 3.x GUI?
As a user I would prefer to have the fixes ASAP.
Or are you concerned of our translators? I prefer to review 100 messages/release over 1000/branch.

Something different would be a new sequence of the dialogs.

Bob-IT:
At the time it was thought that there were too many strings changed that may need a new translation and the change of GtkBox widgets to GtkGrid was too much of a change for maint.

fellen:
Please put the string changes in maint, else I could change them in another way.
While I review translations, I often mark "bad strings" and fix them.

Bob-IT:
Closing as superseded by PR614

PR #518, opened 2019-06-08
Add option to display the Description / Memo in the matcher [Master]

Bob-IT:
On the Import matcher window, with reconciled matches it is not obvious
which transaction / split was matched so add the option to display the
transaction description and the split memo.

I would of pushed this but not sure if I have affected the translations ?

PR #565, opened 2019-08-23
[Master] Import Matcher changes from PR518

Bob-IT:
These changes are as a result of ideas in #518 which I will close when I push this.

The main change is from a list store to a tree store and using a child row to display the matched information. Other changes involved are restricting the selected account currency to that of the import if provided and some changes to the tree view.

@christopherlam Could you give this a spin, hopefully I have included the good stuff and like my change to your idea.

### DIFF
--- a/common/debug/valgrind/valgrind-compiz.supp
+++ b/common/debug/valgrind/valgrind-compiz.supp
@@ -1,0 +1,919 @@
+# -*- tab-width: 3; indent-tabs-mode: nil -*-
+#
+# GNOME supressions base.supp:
+#  git://github.com/dtrebbien/GNOME.supp.git
+#  Copyright (C) 2012  Daniel Trebbien
+#
+#  This program is free software; you can redistribute it and/or
+#  modify it under the terms of the GNU General Public License
+#  as published by the Free Software Foundation; either version 2
+#  of the License, or (at your option) any later version.
+#
+#  This program is distributed in the hope that it will be useful,
+#  but WITHOUT ANY WARRANTY; without even the implied warranty of
+#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#  GNU General Public License for more details.
+#
+#  You should have received a copy of the GNU General Public License
+#  along with this program; if not, write to the Free Software
+#  Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+#
+# Python supressions valgrind-python.supp:
+#  http://hg.python.org/cpython
+#  Copyright (c) 2001, 2002, 2003, 2004, 2005, 2006, 2007, 2008, 2009, 2010,
+#  2011, 2012 Python Software Foundation; All Rights Reserved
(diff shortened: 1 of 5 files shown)


christopherlam:
This is a nice approach.

Also nice that clicking A/A+R/R adds/removes the angle-icon. Perhaps, if 'show matched txns' is true, changing from A to A+R/R should also auto-open the matched txn as well?

Bob-IT:
> This is a nice approach.
> 
> Also nice that clicking A/A+R/R adds/removes the angle-icon. Perhaps, if 'show matched txns' is true, changing from A to A+R/R should also auto-open the matched txn as well?

Interesting, I do not think I tried that, will have a look...

Bob-IT:
@christopherlam I made some changes for your suggestion so if you could try it and confirm it covers what you wanted I will push this next week baring any other concerns.

christopherlam:
Looks good now.

Should the OK button be disabled if matchlist is invalid state? eg user selects A+R/R but does not select a matched book split?

Hmm: Bug? From import matcher, don't click OK, close dialogbox with Alt-F4 (or cancel, or 'x'), examine book: it still has imported transactions. This is not always reproducible.

Bob-IT:
@christopherlam  I assume the line is red, unless you are saying something else. I think the OK button should always be enabled.

For the second part, that could be similar to if you have the source account open, start the import, deselect a couple of lines, hit OK and then in the register you see a number of blank transactions, view->refresh gets rid of them so maybe force a refresh of the source account register. OK may of just seen your scenario, did you have the little squares on the debit/credit columns?

christopherlam:
> For the second part, that could be similar to if you have the source account open, start the import, deselect a couple of lines, hit OK and then in the register you see a number of blank transactions, view->refresh gets rid of them so maybe force a refresh of the source account register. OK may of just seen your scenario, did you have the little squares on the debit/credit columns?

Yes, very unbalanced unusable book, with the import-matcher dialog now gone. Can't figure out exact steps to reproduce yet, so, may not be relevant to this PR.

christopherlam:
I think I managed to reproduce the bug: from the import-matcher window:
* click [Cancel] or press [Esc] - book is intact as it should.
* press [Alt-F4] - book is now completely unbalanced, filled with unmatched imported transactions .

Looks like the 'close' signal isn't being fired to launch `on_matcher_cancel_clicked` when Alt-F4 is used to close window. But this is present in maint, therefore nothing to do with this PR.

christopherlam:
With #571 gone, I think this is now fine. Thank you for doing this!

common/debug/valgrind/valgrind-compiz.supp
```diff
@@ -0,0 +1,919 @@
+# -*- tab-width: 3; indent-tabs-mode: nil -*-
+#
+# GNOME supressions base.supp:
+#  git://github.com/dtrebbien/GNOME.supp.git
+#  Copyright (C) 2012  Daniel Trebbien
+#
```
christopherlam: I'm glad this change is finalised. But this file is merged in error isn't it?

Bob-IT: Yes, not sure how that got in there... removed.

PR #566, opened 2019-08-24
Fix for bug in string_to_gnc_numeric where negative decimal fractions…

craigmj:
… did not parse as negative (e.g. -0.1 would parse as 0.1)

For some reason I cannot run `make check` - my test-gnc-numeric always fails, even before my patch. So please check that for me :)


jralls:
Merged, with changes to accommodate the case where there's no leading 0, e.g. `.12345` and `-.12345` .

Thanks.

PR #569, opened 2019-08-25
Hebrew COA sets

avma:


fellen:
Hi Avi,
the failed check is not your fault.

For me it looks OK, but I had only 2 lessons Hebrew at school.

Just for curiosity, why did you adjust one currency in acctchrt_brokerage.gnucash-xea:427?
While the currency finally gets set by your selection in the new file assistant, it is not important, but for consistency, I would prefer to have the same in all accounts.

I could adjust this one or you could adjust all other.

avma:
I guess I forgot to set it back to USD, I was playing with it testing if it makes any difference...did that caused the failer?

fellen:
No, the failure came from a change in the build system by somebody else.

Thank you for your work and patience!

PR #570, opened 2019-08-25
Hebrew Translation for v3.6

avma:


fellen:
Hi,

I amended the commit message of commit 7e5be75, so it became d8468b9

In commit 8d1c5f8 I inserted the contexts because for bug 797349 - "A"ssociate header badly translated - I will need more time.

Finally with commit abc0964 everything became part of our stable branch.

Thank you!

PR #572, opened 2019-08-29
TaxTable ref count on entries

Bob-IT:
I was getting this in my trace file with an xml Gnucash file...
 * 11:58:50  WARN <gnc.io> [taxtable_reset_refcount()] Fixing refcount on
taxtable 2dca815fad04436cbb5497517803cc79 (2 -> 1)

after some investigation it was due to the tax table on the blank entry being counted if it was still open when GnuCash is closed. I fixed this by adding a gboolean to the GncEntry indicating whether it was the blank entry and so could be counted or not which also fixed the total entries count.

I will push this at the weekend if no one sees a problem with it.

### DIFF
--- a/gnucash/register/ledger-core/gncEntryLedgerControl.c
+++ b/gnucash/register/ledger-core/gncEntryLedgerControl.c
@@ -93,6 +93,8 @@ gnc_entry_ledger_save (GncEntryLedger *ledger, gboolean do_commit)
         time64 time = gnc_time (NULL);
         gncEntrySetDateEntered (blank_entry, time);
 
+        gncEntrySetBlankEntry (blank_entry, FALSE);
+
         switch (ledger->type)
         {
         case GNCENTRY_ORDER_ENTRY:
@@ -100,6 +102,9 @@ gnc_entry_ledger_save (GncEntryLedger *ledger, gboolean do_commit)
             break;
         case GNCENTRY_INVOICE_ENTRY:
         case GNCENTRY_CUST_CREDIT_NOTE_ENTRY:
+            /* saving blank entry so increase the tax table ref if present */
+            if (gncEntryGetInvTaxable (blank_entry))
+                gncTaxTableIncRef (gncEntryGetInvTaxTable (blank_entry));
             /* Anything entered on an invoice entry must be part of the invoice! */
             gncInvoiceAddEntry (ledger->invoice, blank_entry);
             break;
@@ -107,6 +112,9 @@ gnc_entry_ledger_save (GncEntryLedger *ledger, gboolean do_commit)
         case GNCENTRY_EXPVOUCHER_ENTRY:
         case GNCENTRY_VEND_CREDIT_NOTE_ENTRY:
         case GNCENTRY_EMPL_CREDIT_NOTE_ENTRY:
+            /* saving blank entry so increase the tax table ref if present */
+            if (gncEntryGetBillTaxable (blank_entry))
+                gncTaxTableIncRef (gncEntryGetBillTaxTable (blank_entry));
             /* Anything entered on an invoice entry must be part of the invoice! */
             gncBillAddEntry (ledger->invoice, blank_entry);
             break;
--- a/gnucash/register/ledger-core/gncEntryLedgerLoad.c
+++ b/gnucash/register/ledger-core/gncEntryLedgerLoad.c
@@ -364,6 +364,9 @@ void gnc_entry_ledger_load (GncEntryLedger *ledger, GList *entry_list)
             gnc_suspend_gui_refresh ();
 
             blank_entry = gncEntryCreate (ledger->book);
+
+            gncEntrySetBlankEntry (blank_entry, TRUE);
+
             gncEntrySetDateGDate (blank_entry, &ledger->last_date_entered);
             ledger->blank_entry_guid = *gncEntryGetGUID (blank_entry);
 
--- a/gnucash/register/ledger-core/gncEntryLedgerLoad.c
+++ b/gnucash/register/ledger-core/gncEntryLedgerLoad.c
@@ -365,6 +365,8 @@ void gnc_entry_ledger_load (GncEntryLedger *ledger, GList *entry_list)
 
             blank_entry = gncEntryCreate (ledger->book);
 
+            gncEntryBeginEdit (blank_entry);
+
             gncEntrySetBlankEntry (blank_entry, TRUE);
 
             gncEntrySetDateGDate (blank_entry, &ledger->last_date_entered);
--- a/libgnucash/backend/xml/gnc-entry-xml-v2.cpp
+++ b/libgnucash/backend/xml/gnc-entry-xml-v2.cpp
@@ -796,8 +796,13 @@ entry_sixtp_parser_create (void)
 static void
 do_count (QofInstance* entry_p, gpointer count_p)
 {
-    int* count = static_cast<decltype (count)> (count_p);
-    (*count)++;
+    GncEntry* entry = (GncEntry*) entry_p;
+
+    if (!gncEntryIsBlankEntry(entry))
+    {
+        int* count = static_cast<decltype (count)> (count_p);
+        (*count)++;
+    }
 }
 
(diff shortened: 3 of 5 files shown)


jralls:
This makes no sense to me at all. Why would the blank entry be persisted? It's a GUI implementation detail to provide a place for the user to add a new entry. Like the blank split in a regular register it should be created when the parent Invoice/Bill/CreditNote is opened for editing and destroyed without persistence when the object is committed or rolled back.

I particularly dislike calling `gncEntryBeginEdit` with no corresponding `gncEntryCommitEdit`. That leaves the instance with a an edit count so future edits must commit twice to actually commit the edit and save the result.

Bob-IT:
Maybe I did not explain it well.

> This makes no sense to me at all. Why would the blank entry be persisted? It's a GUI implementation detail to provide a place for the user to add a new entry. Like the blank split in a regular register it should be created when the parent Invoice/Bill/CreditNote is opened for editing and destroyed without persistence when the object is committed or rolled back.

The blank entry is created when the ledger is loaded and destroyed when it is closed. The point being is that if the ledger is open when the file save is run which is before the gui is destroyed and if the blank entry has a taxtable set it will be counted for the ref count of the taxtable. When the file is loaded the ref count will be wrong and this is what triggers the refcount_reset scrub routine.

> I particularly dislike calling `gncEntryBeginEdit` with no corresponding `gncEntryCommitEdit`. That leaves the instance with a an edit count so future edits must commit twice to actually commit the edit and save the result.

There is a begin edit and commit edit in the entry save routine, all I have done for this is add a begin edit to the load before the set entry date which is similar to what is done for the split register so I do not see what your concern is.

jralls:
>The point being is that if the ledger is open when the file save is run which is before the gui is destroyed and if the blank entry has a taxtable set it will be counted for the ref count of the taxtable. 

The point being that if the ledger has an editlevel > 0 *file save should be using the rollback copy*. That's the whole point of edit and commit: The changes aren't applied and are certainly not saved until committed.

> There is a begin edit and commit edit in the entry save routine, all I have done for this is add a begin edit to the load before the set entry date which is similar to what is done for the split register so I do not see what your concern is.

`gnc_entry_ledger_load` is called only in `gnc_entry_ledger_refresh_internal`, which is the functional part of `gnc_entry_ledger_display_refresh`.  That function gets called a lot, so by the time gnc_entry_ledger_save calls gncEntryCommitEdit the editlevel will be much > 1 and the commit will be a no-op.

Bob-IT:
Thank you, did not think of that, will scrap this.

Bob-IT:
> `gnc_entry_ledger_load` is called only in `gnc_entry_ledger_refresh_internal`, which is the functional part of `gnc_entry_ledger_display_refresh`. That function gets called a lot, so by the time gnc_entry_ledger_save calls gncEntryCommitEdit the editlevel will be much > 1 and the commit will be a no-op.
OK, after sleeping on it and trying to work out why I may of got this so wrong and did not notice in any testing. Yes the gnc_entry_ledger_load can be called many times but there is a test for the blank_entry_guid being NULL, if so it creates a new blank_entry with gncEntryCreate and I added the gncEntryBeginEdit immediately after. Once created on the next ledger_load it bypasses this create/begin until one does the save on the blank_entry which has the commit, clears the blank_entry guid from the ledger and then on the next load creates a new blank_entry.

With testing by adding a print statement to gncEntryBeginEdit and Commit to get the qof edit level it only ever gets to 2.
It still might be the wrong thing to do but I just wanted to try and understand why.

jralls:
Ah, I see, so the extra begin_edit() is called only when creating the blank entry. GncEntry also has an evil hack, `GncEntryIsOpen()` that returns true if edit_level > 0 and is used to conditionally block the other calls to `gncEntryBeginEdit()`. Those are in `gnc_entry_ledger_save`, `gnc_entry_ledger_clear_blank_entry`, and `gnc_entry_ledger_delete_current_entry`; the latter two destroy the entry so it's not clear to me how the edit_level gets to 2, but even that will prevent running `qof_commit_edit_part2` and mess up the save. Maybe protecting your added `gncEntryBeginEdit()` with `if (gncEntryIsOpen())` as is done elsewhere will work.

jralls:
As for the taxtable reference on the blank entry, wouldn't it be simpler to not increment the taxtable reference count until the blank entry gets turned into a real entry in `gnc_entry_ledger_save()`?

Bob-IT:
OK, did some more testing and all I am doing in the following is with the ledger loaded, add some text to the description cell of the blank entry and press Enter...
Original order,

Ledger load
create blank entry
gncEntryBeginEdit leave edit level 1,  set date,
gncEntryCommitEdit, gncEntryOnDone, leaves edit level 0
gncEntryBeginEdit leave edit level 1,  set invtaxtable,
gncEntryCommitEdit, gncEntryOnDone, leaves edit level 0
update description and press Enter
gnc_entry_ledger_save, entry not open do gncEntryBeginEdit leave edit level 1
gncEntryBeginEdit leave edit level 2,  set description,
gncEntryCommitEdit, leave edit level 1
gncEntryBeginEdit leave edit level 2,  set date entered,
gncEntryCommitEdit, leave edit level 1
gncEntryBeginEdit leave edit level 2,  set invoice,
gncEntryCommitEdit, leave edit level 1
gnc_entry_ledger_save, do gncEntryCommitEdit, gncEntryOnDone,  leave edit level 0
Ledger load
create blank entry....

With my change

Ledger load
create blank entry, gncEntryBeginEdit leave edit level 1
gncEntryBeginEdit leave edit level 2,  set date,
gncEntryCommitEdit, gncEntryOnDone, leaves edit level 1
gncEntryBeginEdit leave edit level 2,  set invtaxtable,
gncEntryCommitEdit, gncEntryOnDone, leaves edit level 1
update description and press Enter
gnc_entry_ledger_save, entry already open
gncEntryBeginEdit leave edit level 2,  set description,
gncEntryCommitEdit, leave edit level 1
gncEntryBeginEdit leave edit level 2,  set date entered,
gncEntryCommitEdit, leave edit level 1
gncEntryBeginEdit leave edit level 2,  set invoice,
gncEntryCommitEdit, leave edit level 1
gnc_entry_ledger_save, do gncEntryCommitEdit, gncEntryOnDone,  leave edit level 0
Ledger load
create blank entry....

As can be seen, they both end up with an edit level of 0, also I do not think there is any point protecting my addition of gncEntryBeginEdit as the blank entry has just been created.

Bob-IT:
> As for the taxtable reference on the blank entry, wouldn't it be simpler to not increment the taxtable reference count until the blank entry gets turned into a real entry in `gnc_entry_ledger_save()`?

I started with that idea also and came up with the first commit but maybe there is another way of doing it. The problem is on creation of the blank_entry, it could have a default tax table or not.

So, create blank_entry with, added tax table, decrease refcount of tax table and then save blank_entry with increase tax table refcount. But if the tax table is cleared, the save would not change the tax table refcount so one would have to store whether the blank_entry was created with or without a taxtable and which one to do the right thing.

The commit I proposed did the following...
mark the entry as a blank_entry
if it is the blank entry, do not change any tax table recounts at all
on save, if it is the blank_entry and there is a tax table, increase the refcount of the tax table and clear the blank_entry mark from the entry.

jralls:
Oh, you hadn't said that edit_level was getting raised-and-lowered in the setters so I thought you meant that you had an edit_level of 2 in gnc_entry_ledger_load. 

The existence of a blank entry really shouldn't leak out of the register code where it's an implementation detail. It looks to me like part of the problem is that the taxtable ref count is changed gncEntrySetBillTaxTable and gncEntrySetInvoiceTaxTable instead of in gncEntryCommitEdit between the calls to qof_commit_edit and qof_commit_edit_part2, but for that to work GncEntryLedger should call gncEntryBeginEdit when the user begins editing an entry instead of at the beginning of gnc_entry_ledger_save. In fact, instead of calling gncEntryBeginEdit if edit_level is 0 gnc_entry_level_save should return without doing anything; then if the user has never edited the blank entry it's simply destroyed.

Is that feasible?

PR #574, opened 2019-09-03
Bug 797344 - Mixing of preinstalled GoogleTest libraries with headers from source directory

gruberchr:
This PR replaces PR #557


gruberchr:
I simplified GoogleTest integration for v1.8.0 now.
Unfortunately I can not build locally at the moment on `master`. Target `scm-engine-0` fails when compiling *libgnucash/engine/engine-utilities.scm*:

```
Backtrace:
           1 (primitive-load-path "gnucash/engine/gnc-numeric")
In ice-9/boot-9.scm:
   752:25  0 (dispatch-exception _ _ _)

ice-9/boot-9.scm:752:25: In procedure dispatch-exception:
In procedure primitive-load-path: Unable to find file "gnucash/engine/gnc-numeric" in load path
```

gjanssens:
This is usually the result of reusing a maint build directory for a master build or alternatively the linker picks up a 3.x version of gnucash installed system-wide.
You are building in a dedicated build directory I hope ? If so, remove it completely and restart from running cmake. Or if you have installed a version of gnucash via your package manager, temporarily remove it.

jralls:
The latest commit needs to change line 6 of `util/ci/ubuntu-16.04-testscript`:
```diff
- export GTEST_ROOT=/gtest/googletest GMOCK_ROOT=/gtest/googlemock
+ export GTEST_ROOT=/gtest
```

gruberchr:
> This is usually the result of reusing a maint build directory for a master build or alternatively the linker picks up a 3.x version of gnucash installed system-wide. You are building in a dedicated build directory I hope ?

Yes, I built in a separate build directory. But I have GnuCash 3.5 installed from package manager. How does it interfere with my local build?

> Or if you have installed a version of gnucash via your package manager, temporarily remove it.

What means "temporarily"? Do I always have to remove it, when running CMake for GnuCash? Is there another solution than removing?

jralls:
The problem is that Guile loads C libraries with dlopen and that looks on the system path (`/usr/lib, /usr/lib64, /usr/lib/x86_64_gnu_linux` etc depending on your distro) before it looks in the build directory. Rather than uninstalling you could try renaming the libraries temporarily. Supposing that your system library dir is `/usr/lib` you might try
```shell
sudo mkdir /usr/lib/hidden
sudo mv /usr/lib/libgnc*.so /usr/lib/hidden
sudo mv /usr/lib/gnucash /usr/lib/hidden
```
and when you want to use the production GnuCash again,
```shell
sudo mv /usr/lib/hidden/libgnc*.so /usr/lib
sudo mv /usr/lib/hidden/gnucash /usr/lib
```

jralls:
All tests pass now, so merged to master. Thanks!

PR #579, opened 2019-09-09
Update he.po to v3.7

avma:
Plus fix some bidi conversion when Latin and Hebrew chars appear both in the same sentence.


fellen:
Thanks!

PR #583, opened 2019-09-22
show backtrace in report error

christopherlam:
Couple commits to show backtrace in report crash.

* exposes a SCM string `gnc:last-captured-error` containing last backtrace
* when rendering report-crash window, include it in `<pre>` block
* we *do* need to upgrade `gnc:html-string-sanitize` to `libgnucash` to neutralize the captured-error -- otherwise if backtrace contains `</pre>` it will prematurely end the block and mess up the error. Or even worse, backtrace could contain a `<script>` block :)
* note this snippet is not thread-safe -- `gnc:last-captured-error` could be modified in another thread and the errorpage gets the wrong string, but we'll leave it for another decade...

image

### DIFF
--- a/gnucash/report/report-gnome/window-report.c
+++ b/gnucash/report/report-gnome/window-report.c
@@ -274,10 +274,16 @@ gnc_html_report_stream_cb (const char *location, char ** data, int *len)
 
     if (!ok)
     {
+        SCM captured = scm_c_eval_string ("gnc:last-captured-error");
+        gchar *captured_str = gnc_scm_to_utf8_string(captured);
+
         *data = g_strdup_printf ("<html><body><h3>%s</h3>"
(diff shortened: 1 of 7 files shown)


christopherlam:
Would be good for a C reviewer. I don't think I need to explicitly free anything in the C amendments.

jralls:
But you do. Look at the documentation for `gnc_scm_to_utf8_string` in `libgnucash/core-utils/gnc-guile-utils.h`.

christopherlam:
this is better?
/me still clueless -- I've seen non-freed code like https://github.com/Gnucash/gnucash/blob/maint/gnucash/gnome-utils/dialog-options.c#L744

jralls:
Yes, it's right now. And the example you found is a memory leak. I've no doubt at all that there are others that are similar.

PR #584, opened 2019-09-24
Update hr.po file with 3.7 pot-file

milotype:
update and some corrections


fellen:
Thanks, committed.
For your convenience I merged also a recent gnucash,pot, but until now there are only 3 changed messages.

PR #585, opened 2019-09-28
[master] one-time fix for budget amount signs

christopherlam:
I've reviewed all calls to `gnc_budget_set_account_period_value` to ensure that internal budget amounts are *not* reversed, yet *displayed* or *added* according to account-reversal policy.

Fixing reports should not be too difficult. Note this does not fix budget amounts already stored reversed. I do not know how to create backward-compatibility path for existing budgets.

- [x] also fix budget-reports sign-reversal handling
- [x] fix budget amounts already stored.
- [x] all code converted to C


christopherlam:
rebased to *master* with only 1 commit...

christopherlam:
Budget signs fixing would look like:
```scheme
;; transform each budget's periods' signs according to
;; account-reversal property, as a one-off fix

(define (fixbudgets)
  (unless (gnc-features-check-used book BUDGET-AMOUNTS-FIXED)
    (for-each
     (lambda (budget)
       (for-each
        (lambda (account)
          (for-each
           (lambda (period)
             (when (gnc-reverse-balance account)
               (let ((amt (gnc-budget-get-account-period-value
                           budget account period)))
                 (unless (null? amt)
                   (gnc-budget-set-account-period-value
                    budget account period (- amt))))))
           (iota (gnc-budget-get-num-periods budget))))
        all-accounts))
     all-budgets)
    (gnc-features-set-used book BUDGET-AMOUNTS-FIXED)))
```

christopherlam:
P.S. fixing the budget amount signs will work correctly if and only if the user has not changed the account-reversal book option between budget generation and running above fix.

christopherlam:
I've added the `gnc-features` guard and the one-time fix, but my C is probably ugly.

gnucash/gnome-utils/gnc-file.c
```diff
@@ -640,6 +643,35 @@ gnc_file_query_save (GtkWindow *parent, gboolean can_cancel)
...
+    SCM scm_budget_guid = scm_from_utf8_string (guid_to_string(budget_guid));
+    scm_call_1 (reverse_proc, scm_budget_guid);
+}
+
+/* For GnuCash 4.0 onwards - fix budget signs */
+/* guard is set if we have completed reversal, or there are no
+   budgets in book */
+static void consider_fix_budget_signs (QofBook *book)
+{
+    if (!(gnc_features_check_used (book, GNC_FEATURE_BUDGET_UNREVERSED)))
```
gjanssens: To avoid nesting loops in C you may consider returning early from checks. So instead you'd write

```
if (gnc_features_check_used (book, GNC_FEATURE_BUDGET_UNREVERSED))
    return;
if (!gnc_budget_get_default (book))
    return;
if (...
```

christopherlam:
Agree on all counts. This PR is Mk.1, I was dragged down by failing to create ScrubBudget.c from scratch, and I'll refrain from rebasing on master until it's complete.

> So 3.8 needs a way to determine which data format it's looking at and act accordingly. We can use your newly added feature check `GNC_FEATURE_BUDGET_UNREVERSED` to make this distinction. But contrary to what you did that means 3.8 should be made aware of this flag so it should already be defined in 3.8. The clever trick lies in the fact the 3.x series will _never_ actively set that flag. Only 4.x will. That will ensure that once 4.x converts the data, gnucash versions older than 3.7 will refuse to work on that data. 3.8 and up on the other hand will, and can use the presence of the flag to determine how to interpret the budget data.

I'm sure you're aware that currently data&reports are in an awkward situation that the fix will only work well if & only if the reversal-pref set *during* budget-generation is the same one set *during* the one-time fix. And, as has already been documented in numerous bugs, *during* budget-report generation. With the 4.0 upgrade it's not too bad; I inserted the gnc_verify_dialog with placeholder text, considering warning "if the budget signs are incorrect, please don't save this datafile; set the reversal-pref to the same one used during budget-generation, and reload to retry the fix."

With 3.8 forward-compatibility read, I am not sure how this will work. For 3.8 budget reports I can retain the two possibilities (ugly but doable); for budget .c files I don't know what it should do.

christopherlam:
> With 3.8 forward-compatibility read, I am not sure how this will work. For 3.8 budget reports I can retain the two possibilities (ugly but doable); for budget .c files I don't know what it should do.

Uhh re-reading your comment, now I understood you'd already considered the possibility :-/

libgnucash/engine/ScrubBudget.c
```diff
@@ -0,0 +1,95 @@
...
+           (accounts (gnc-account-get-descendants (gnc-book-get-root-account book))) \
+           (periods (iota (gnc-budget-get-num-periods budget)))) \
+      (for-each \
+       (lambda (account) \
+         (for-each \
+          (lambda (period) \
+            (let ((amt (gnc-budget-get-account-period-value budget account period))) \
+              (gnc-budget-set-account-period-value budget account period (- amt)))) \
+          periods)) \
+       (filter gnc-reverse-balance accounts))))";
```
jralls: C-calling-scheme, please. Based on the names all of those functions are wrapped from C anyway.

gjanssens:
>     * I'll aim to create another PR for maint to read/write pre/post feature books
> 
>     * I do not know how gnc-features work -- will a 4.0 featured book refuse to load in 3.7, or will it refuse to save?

It will refuse to load. And by the way to the user you have to report it as "requiring at least GnuCash 3.8", not a 4.0 feature. You define the feature already for 3.8, so that will be the first version able to read a book with this feature set.

libgnucash/engine/ScrubBudget.c
christopherlam: Baby steps! The previous commit would load normal scheme in gnome-utils.scm. We'll want to move scrub to libgnucash which incidentally mean there'll be no more dialog to offer conversion.

jralls: Rewrite-to-C *then* move it to `engine`.

christopherlam:
Remaining issues:
* an info message will be displayed if the budgets have been sign-reversed. If there are no budgets, the feature flag is still set. Should the message be amended to omit the budget reports, yet retain the backward-incompatibility with older versions?
* gnc_reverse_balance still needs to be accessible.

christopherlam:
We've decided to use heuristics instead of querying pref to determine accounts to reverse.
This mean no need for `gnc_account_reverse` anymore.

https://code.gnucash.org/logs/2019/10/19.html#T19:43:39

christopherlam:
Without shame, this iteration is *functional* :scream: 

Scrubbing is all done via C:
* The book is analysed, and if already featured, return 0; if no budget return 1 for silent feature; if  desirable return 2 which asks gnc_verify_dialog (default No); if agree then proceed with reversal.
* Each budget analysed via heuristics done in scheme -- returns 5=inc-exp, 6=none, 7=credit-accounts. e.g. inc-exp means reverse inc or exp only; none means skip reversal; credit means reverse liability/AP/equity/income/credit. Perform actual reversal in C.

Example log output:
```
;;; ("Reversal=Inc/Exp")
type asset number 4
type liability number -1
type income number 1
type expense number -1
type equity number -2
inc/exp
* 23:59:50  WARN <gnc.engine.scrub> [fix_budget_acc_sign()] budget account [Income] sign-reversed as 5!
* 23:59:50  WARN <gnc.engine.scrub> [fix_budget_acc_sign()] budget account [Expenses] sign-reversed as 5!

;;; ("Reversal=None")
type asset number 4
type liability number -1
type income number 1
type expense number 1
type equity number -1
none

;;; ("Reversal=Credit-Accounts")
type asset number 4
type liability number 1
type income number 1
type expense number 1
type equity number 2
credit-accounts
* 23:59:50  WARN <gnc.engine.scrub> [fix_budget_acc_sign()] budget account [Income] sign-reversed as 7!
* 23:59:50  WARN <gnc.engine.scrub> [fix_budget_acc_sign()] budget account [Equity] sign-reversed as 7!
* 23:59:50  WARN <gnc.engine.scrub> [fix_budget_acc_sign()] budget account [Opening Balances] sign-reversed as 7!
* 23:59:50  WARN <gnc.engine.scrub> [fix_budget_acc_sign()] budget account [Liabilities] sign-reversed as 7!
```

christopherlam:
Example reversal tracefile output:

```
;;; ("Reversal=Inc/Exp")
type asset number 4
type liability number -1
type income number 1
type expense number -1
type equity number -2
inc/exp
* 14:09:58  WARN <gnc.engine.scrub> [fix_budget_sign()] processing budget [Reversal=Inc/Exp] for reversal
* 14:09:58  WARN <gnc.engine.scrub> [fix_budget_acc_sign()] budget account [Income] is inc/exp. reverse!
* 14:09:58  WARN <gnc.engine.scrub> [fix_budget_acc_sign()] budget account [Income] period 1 sign-reversed!
* 14:09:58  WARN <gnc.engine.scrub> [fix_budget_acc_sign()] budget account [Income] period 2 sign-reversed!
* 14:09:58  WARN <gnc.engine.scrub> [fix_budget_acc_sign()] budget account [Income] period 3 sign-reversed!
* 14:09:58  WARN <gnc.engine.scrub> [fix_budget_acc_sign()] budget account [Income] period 4 sign-reversed!
* 14:09:58  WARN <gnc.engine.scrub> [fix_budget_acc_sign()] budget account [Income] period 5 sign-reversed!
* 14:09:58  WARN <gnc.engine.scrub> [fix_budget_acc_sign()] budget account [Income] period 6 sign-reversed!
* 14:09:58  WARN <gnc.engine.scrub> [fix_budget_acc_sign()] budget account reversal [Income] completed!
* 14:09:58  WARN <gnc.engine.scrub> [fix_budget_acc_sign()] budget account [Expenses] is inc/exp. reverse!
* 14:09:58  WARN <gnc.engine.scrub> [fix_budget_acc_sign()] budget account [Expenses] period 1 sign-reversed!
* 14:09:58  WARN <gnc.engine.scrub> [fix_budget_acc_sign()] budget account [Expenses] period 2 sign-reversed!
* 14:09:58  WARN <gnc.engine.scrub> [fix_budget_acc_sign()] budget account [Expenses] period 3 sign-reversed!
* 14:09:58  WARN <gnc.engine.scrub> [fix_budget_acc_sign()] budget account [Expenses] period 4 sign-reversed!
* 14:09:58  WARN <gnc.engine.scrub> [fix_budget_acc_sign()] budget account [Expenses] period 5 sign-reversed!
* 14:09:58  WARN <gnc.engine.scrub> [fix_budget_acc_sign()] budget account [Expenses] period 6 sign-reversed!
* 14:09:58  WARN <gnc.engine.scrub> [fix_budget_acc_sign()] budget account reversal [Expenses] completed!

;;; ("Reversal=None")
type asset number 4
type liability number -1
type income number 1
type expense number 1
type equity number -1
none
* 14:09:58  WARN <gnc.engine.scrub> [fix_budget_sign()] processing budget [Reversal=None] for reversal

;;; ("Reversal=Credit-Accounts")
type asset number 4
type liability number 1
type income number 1
type expense number 1
type equity number 2
credit-accounts
* 14:09:58  WARN <gnc.engine.scrub> [fix_budget_sign()] processing budget [Reversal=Credit-Accounts] for reversal
* 14:09:58  WARN <gnc.engine.scrub> [fix_budget_acc_sign()] budget account [Income] is credit-account. reverse!
* 14:09:58  WARN <gnc.engine.scrub> [fix_budget_acc_sign()] budget account [Income] period 1 sign-reversed!
* 14:09:58  WARN <gnc.engine.scrub> [fix_budget_acc_sign()] budget account [Income] period 2 sign-reversed!
* 14:09:58  WARN <gnc.engine.scrub> [fix_budget_acc_sign()] budget account [Income] period 3 sign-reversed!
* 14:09:58  WARN <gnc.engine.scrub> [fix_budget_acc_sign()] budget account [Income] period 4 sign-reversed!
* 14:09:58  WARN <gnc.engine.scrub> [fix_budget_acc_sign()] budget account [Income] period 5 sign-reversed!
* 14:09:58  WARN <gnc.engine.scrub> [fix_budget_acc_sign()] budget account [Income] period 6 sign-reversed!
* 14:09:58  WARN <gnc.engine.scrub> [fix_budget_acc_sign()] budget account reversal [Income] completed!
* 14:09:58  WARN <gnc.engine.scrub> [fix_budget_acc_sign()] budget account [Equity] is credit-account. reverse!
* 14:09:58  WARN <gnc.engine.scrub> [fix_budget_acc_sign()] budget account [Equity] period 1 sign-reversed!
* 14:09:58  WARN <gnc.engine.scrub> [fix_budget_acc_sign()] budget account [Equity] period 2 sign-reversed!
* 14:09:58  WARN <gnc.engine.scrub> [fix_budget_acc_sign()] budget account [Equity] period 3 sign-reversed!
* 14:09:58  WARN <gnc.engine.scrub> [fix_budget_acc_sign()] budget account [Equity] period 4 sign-reversed!
* 14:09:58  WARN <gnc.engine.scrub> [fix_budget_acc_sign()] budget account [Equity] period 5 sign-reversed!
* 14:09:58  WARN <gnc.engine.scrub> [fix_budget_acc_sign()] budget account [Equity] period 6 sign-reversed!
* 14:09:58  WARN <gnc.engine.scrub> [fix_budget_acc_sign()] budget account reversal [Equity] completed!
* 14:09:58  WARN <gnc.engine.scrub> [fix_budget_acc_sign()] budget account [Opening Balances] is credit-account. reverse!
* 14:09:58  WARN <gnc.engine.scrub> [fix_budget_acc_sign()] budget account [Opening Balances] period 1 sign-reversed!
* 14:09:58  WARN <gnc.engine.scrub> [fix_budget_acc_sign()] budget account [Opening Balances] period 2 sign-reversed!
* 14:09:58  WARN <gnc.engine.scrub> [fix_budget_acc_sign()] budget account [Opening Balances] period 3 sign-reversed!
* 14:09:58  WARN <gnc.engine.scrub> [fix_budget_acc_sign()] budget account [Opening Balances] period 4 sign-reversed!
* 14:09:58  WARN <gnc.engine.scrub> [fix_budget_acc_sign()] budget account [Opening Balances] period 5 sign-reversed!
* 14:09:58  WARN <gnc.engine.scrub> [fix_budget_acc_sign()] budget account [Opening Balances] period 6 sign-reversed!
* 14:09:58  WARN <gnc.engine.scrub> [fix_budget_acc_sign()] budget account reversal [Opening Balances] completed!
* 14:09:58  WARN <gnc.engine.scrub> [fix_budget_acc_sign()] budget account [Liabilities] is credit-account. reverse!
* 14:09:58  WARN <gnc.engine.scrub> [fix_budget_acc_sign()] budget account [Liabilities] period 1 sign-reversed!
* 14:09:58  WARN <gnc.engine.scrub> [fix_budget_acc_sign()] budget account [Liabilities] period 2 sign-reversed!
* 14:09:58  WARN <gnc.engine.scrub> [fix_budget_acc_sign()] budget account [Liabilities] period 3 sign-reversed!
* 14:09:58  WARN <gnc.engine.scrub> [fix_budget_acc_sign()] budget account [Liabilities] period 4 sign-reversed!
* 14:09:58  WARN <gnc.engine.scrub> [fix_budget_acc_sign()] budget account [Liabilities] period 5 sign-reversed!
* 14:09:58  WARN <gnc.engine.scrub> [fix_budget_acc_sign()] budget account [Liabilities] period 6 sign-reversed!
* 14:09:58  WARN <gnc.engine.scrub> [fix_budget_acc_sign()] budget account reversal [Liabilities] completed!
* 14:09:58  WARN <gnc.engine.scrub> [scrub_budget_signs_add_feature()] Budget sign reversals complete
```

christopherlam:
Removed all traces of scheme in this PR!
See very latest commit for details. Heuristics slightly different in conversion scm->C.

gnucash/gnome-utils/gnc-file.c
```diff
@@ -1028,9 +1029,24 @@ gnc_post_file_open (GtkWindow *parent, const char * filename, gboolean is_readon
...
+    {
+    case 0:
+        break;
+    case 1:
+        scrub_budget_signs_add_feature (new_book);
+        break;
+    case 2:
+        if (gnc_verify_dialog (parent, FALSE, (_("A one-time fix should be applied to budget amounts so that they are stored internally with natural rather than reversed numbers. After this fix, please review budgets and ensure amounts are as expected. If this fix is not applied, the budget reports will be incorrect. Should the fix be applied? Please note after this fix, the data will be unreadable by GnuCash version 3.7 or earlier." ))))
+            scrub_budget_signs_doit (new_book);
+    }
```
gjanssens: I'm still unsure we covered all possible scenarios:
* User applies fix, and accepts changes (the ideal case)
* User applies fix, rejects changes (back to square 1, will be asked again next time)
* User refuses fix, anything goes as we simply don't know what the user will do next. Will the user manually fix the budgets? Change sign reversal strategy? What? Assuming the user chose to fix the budgets manually, it would be annoying to get the same fix request again on next startup. And again and again until s/he finally allows the scrub to run once because that's currently the only way the feature flag gets set if there are already budgets. And perhaps that's exactly what the user didn't want to do! Perhaps our heuristics were wrong after all in a some obscure corner case situation.

So I think we're missing a solid path for the case where the user wants to fix the budgets manually.

Based on that I would propose a slightly different approach
First determine the state the current book is in:
* feature set or no budgets -> do nothing at file loading
* budgets but no fix needed -> inform the user we think budgets are in good shape and want to prevent older gnucash versions from messing this up. Explain that doing so will prevent 3.7 or older to open the book, and not doing so will mean the budgets may have sign issues when switching between 3.x and 4.x. Question will be repeated at book loading until user agrees to prevent loading in older versions or user makes a budget change.
* budgets that need fixing -> inform the user we think budgets need fixing and that doing so will prevent 3.7 or older from opening the book. Not doing so will continue to pop up this dialog when opening a book until the user agrees to fix or does the fixes him/herself. In both cases older versions will no longer be able to open the book.

Then we should also catch any budget value changes made directly by the user. On any such change check whether the feature flag is already set. If not pop up a warning explaining that making budget value changes will 
1. prevent older gnucash versions from opening the book and
2. stop gnucash from any future attempt to do natural sign budget  fixes in the future. So potentially the user may want to check any other budgets as well. If continue anyway, set feature flag and leave the rest to the user. If not revert attempted change.

libgnucash/engine/ScrubBudget.c
```diff
@@ -0,0 +1,256 @@
...
+ * 2: offer scrubbing */
+guint scrub_budget_signs_check (QofBook *book)
+{
+    /* feature set. skip scrubbing. */
+    if (gnc_features_check_used (book, GNC_FEATURE_BUDGET_UNREVERSED))
+        return 0;
+
+    /* there are no budgets. set feature. */
+    if (!(gnc_budget_get_default (book)))
+        return 1;
```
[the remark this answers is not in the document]
christopherlam: > Instead defer this to when the user makes changes to some budget value. In fact I think that's the only place where you should set it. Even scrub doesn't have to set it explicitly as it will also make changes to some budget values which then in turn will set the feature code for you.

This is not a bad idea. So leave book unchanged until budget-editor processes a budget-editor-cell input. Will need to think about that some more.

libgnucash/engine/ScrubBudget.c
```diff
@@ -0,0 +1,256 @@
...
+
+/* analyse book, and return following:
+ * 0: don't offer scrubbing
+ * 1: set the flag without scrubbing
+ * 2: offer scrubbing */
+guint scrub_budget_signs_check (QofBook *book)
+{
+    /* feature set. skip scrubbing. */
+    if (gnc_features_check_used (book, GNC_FEATURE_BUDGET_UNREVERSED))
+        return 0;
```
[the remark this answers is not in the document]
christopherlam: ok easy step done in my branch

gnucash/gnome-utils/gnc-file.c
christopherlam: So, your view is:
* feature set, or no budgets -> do nothing
* budgets but no fix needed -> only applies if *all* budgets seem to follow 'reversed balance=none' which is extremely unlikely -- I think the Estimate tool only works when pref is credit-accounts
* budgets need fixing -> offer Yes/No dialog to "allow automatic fixing? if yes it'll be done silently. if no, you may need to fix budgets manually" which applies my fixes and sets budget.

And when budget cells are modified:
* if book is not featured, then show a pop up once: "please be advised editing budget manually will set a flag that will cause the datafile to be unreadable by 3.7 or earlier, please confirm you wish to continue editing?"  yes=set feature, no=abort.

This is not a bad idea. My original idea was that unfeatured datafiles' budgets were always fixed which mean we could remove some old code.

gjanssens: > So, your view is:

It's almost my view...

> 
>     * feature set, or no budgets -> do nothing
> 
>     * budgets but no fix needed -> only applies if _all_ budgets seem to follow 'reversed balance=none' which is extremely unlikely -- I think the Estimate tool only works when pref is credit-accounts
> 
This is really `budgets but we *think* no fix needed`. Someone may be budgetting only reimbursed expenses for example, which would cause the heuristics to come to the wrong conclusion.

>     * budgets need fixing -> offer Yes/No dialog to "allow automatic fixing? if yes it'll be done silently. if no, you may need to fix budgets manually" which applies my fixes and sets budget.
> 
> 
> And when budget cells are modified:
> 
>     * if book is not featured, then show a pop up once: "please be advised editing budget manually will set a flag that will cause the datafile to be unreadable by 3.7 or earlier, please confirm you wish to continue editing?"  yes=set feature, no=abort.
> 
That's the idea though for the best user experience it should also test (again) whether the budget may require fixing and if we think so add an additional message that the signs may be wrong but manually editing the budget would prevent gnucash from offering to fix this on future runs.

> 
> This is not a bad idea. My original idea was that unfeatured datafiles' budgets were always fixed which mean we could remove some old code.

I think that would be a false assumption. Unfeatured datafile's budgets are the ones the may need fixing, because these have been created with older gnucash versions. Or did you mean something else ?

christopherlam: > > * budgets but no fix needed -> only applies if _all_ budgets seem to follow 'reversed balance=none' which is extremely unlikely -- I think the Estimate tool only works when pref is credit-accounts
> 
> This is really `budgets but we *think* no fix needed`. Someone may be budgetting only reimbursed expenses for example, which would cause the heuristics to come to the wrong conclusion.

Ok -- in this case I am not sure I can reliably detect that budgets don't need fixing, i.e.  I think whereby sign-reversal will do nothing. Any ideas here?

> > And when budget cells are modified:
> > * if book is not featured, then show a pop up once: "please be advised editing budget manually will set a flag that will cause the datafile to be unreadable by 3.7 or earlier, please confirm you wish to continue editing?"  yes=set feature, no=abort.
> 
> That's the idea though for the best user experience it should also test (again) whether the budget may require fixing and if we think so add an additional message that the signs may be wrong but manually editing the budget would prevent gnucash from offering to fix this on future runs.

Ok I'll see what I can do.

> > This is not a bad idea. My original idea was that unfeatured datafiles' budgets were always fixed which mean we could remove some old code.
> 
> I think that would be a false assumption. Unfeatured datafile's budgets are the ones the may need fixing, because these have been created with older gnucash versions. Or did you mean something else ?

I meant: gnucash 4.x would aim to set the feature in every book (possibly after applying fixes)  which means that the old code to handle reversal could be removed. But it's fine to use your approach above.

gjanssens: > > > * budgets but no fix needed -> only applies if _all_ budgets seem to follow 'reversed balance=none' which is extremely unlikely -- I think the Estimate tool only works when pref is credit-accounts
> > 
> > 
> > This is really `budgets but we *think* no fix needed`. Someone may be budgetting only reimbursed expenses for example, which would cause the heuristics to come to the wrong conclusion.
> 
> Ok -- in this case I am not sure I can reliably detect that budgets don't need fixing, i.e. I think whereby sign-reversal will do nothing. Any ideas here?

I agree there isn't a 100% reliable way of detecting this. Which is why I proposed to tell the user what we *think* is the right way forward and propose to act according to it:
- either we think a fix is needed, so propose to apply the fix, but offer the user to decide not to do so
- or we think budgets are correct, but we want to carve that in stone by by setting a flag. Again if the user is still using gnucash pre-3.8 on any other machines allow him to not set that flag until he's ready.

We might mention in these initial message already in case of decline the nagging will continue on future reopening of the book until either the user accepts the proposed solution or makes a manual change to any existing budget. End result will always be the book can only be opened in 3.8 or later from then on.

Having said all that, there's another way to look at this: pre-3.8 budget format is considered read-only in 4.x onward. So you could also consider to formulate the user messaging in that sense:

> Hi, we notice your budgets are still in 3.x format. 4.x has a new format to store budgets. We can convert automatically to this new format if you like. Beware that once converted you'll need at least gnucash 3.8 be able to open your book.
> On the other hand if you want to make changes to your budgets in 4.x you have to convert it to the new format. GnuCash 4.x can't write the old format any more.
> Note the conversion may involve a on-time sign correction on your budgeted values. Until you run the conversion some budgeted values may show with wrong signs. However for completeness there is a small chance the budgeted values show up with the wrong sign *after* the conversion. In that unfortunate case you'll have to manually correct these yourself.
> Would you like gnucash to convert your budgets now ?

This message focuses on the impact it has on the user:
- convert (with or without sign conversion which is now an implementation detail the user doesn't need to worry about), which has the following impact
  - budgets can be manipulated
  - books older than 3.8 can't open this book any more
  - have a small chance some budgeted values need manual sign fixes
- keep as is, in which case
  - budgets are read-only
  - several budgeted values are likely to be presented with wrong signs
  - but can still be manipulated in 3.7 and before

This also assumes my previous counter example is the most uncommon case and that your heuristics have a high chance of being right. I actually think they have.
In that case the decision to be able to open in pre-3.8 is for most users the more important factor.

Combined with this opening message, I would effectively make budgets read-only until the conversion has run. I believe we already test on the book's read-onlyness in budget views. It should not be too hard to test for a second read-only factor, being the natural sign feature being unset.

And as a convenience to users that skipped the conversion at book opening we can also add a menu item in the budget menu to convert budgets. That item would only be shown when the feature flag is not set.

What we lose with this scenario is the case where we think conversion is needed and the user believes it is not. In the previous scenario the user could just decline conversion, go to the budget and make one single change. Gnucash would then set the feature flag without any extra conversions. In the new scenario, the user eventually has to go through the conversion at some point. And if unlucky, that means some budgets would get the wrong signs. We have evaluated this is the uncommon case (a theoretical case I just maded up).

In the end I think the new scenario would be easier to cope with for the user. Asking the user to decide how to do the conversion may be stressful for all but the most experienced users.

My apologies for turning everything upside down once more...

christopherlam: I think the current branch already functions pretty much as described. Only the string will need amend. So, please feel free to test this PR with a book with or without existing budgets, and try budget editor and reports.

> Combined with this opening message, I would effectively make budgets read-only until the conversion has run. I believe we already test on the book's read-onlyness in budget views. It should not be too hard to test for a second read-only factor, being the natural sign feature being unset.

fwiw the budget editor doesn't obey the readonly property. Thanks to earlier work in maint, both budget editor and reports can handle both unreversed and reversed.

christopherlam: > And as a convenience to users that skipped the conversion at book opening we can also add a menu item in the budget menu to convert budgets. That item would only be shown when the feature flag is not set.

I think this is the only feature missing. I'd think this is overengineering...

jralls:
You have this pointed at `maint` so it's dragging in all of the master merge commits. The PR title says [master] but that's left over from last year. Don't you really want to cherry-pick your two commits over to maint and force-push?

christopherlam:
> You have this pointed at `maint` so it's dragging in all of the master merge commits. The PR title says [master] but that's left over from last year. Don't you really want to cherry-pick your two commits over to maint and force-push?

A late night mistake indeed.

libgnucash/engine/ScrubBudget.c
```diff
@@ -0,0 +1,238 @@
...
+    for (i = 0; i < heuristics->num_periods; ++i)
+    {
+        if (!gnc_budget_is_account_period_value_set (heuristics->budget, account, i))
+            continue;
+        val = gnc_budget_get_account_period_value (heuristics->budget, account, i);
+        total = gnc_numeric_add_fixed (total, val);
+    }
+
+    i = gnc_numeric_compare (total, gnc_numeric_zero ());
+    PWARN ("acct=%s, total=%ld/%ld, i=%d", xaccAccountGetName (account),
```
jralls, line 76: You can't count on `int64_t` mapping to `long`, so instead use either the standard C `PRId64` or the GLib equivalent `G_GINT64_FORMAT`  :
```
"acct=%s, total=%" PRId64 "/%" PRId64 ", i=%d",
```

libgnucash/engine/ScrubBudget.c
```diff
@@ -0,0 +1,238 @@
...
+    GncBudget* budget;
+    HeuristicsOutput policy;
+} Reversal_t;
+
+typedef struct
+{
+    gint asset,liability,equity,income,expense;
+    gint num_periods;
+    GncBudget* budget;
+} Heuristics;
```
jralls, line 55: These aren't heuristics, they're accumulators (most of which are used only to print a progress message) and a couple of arguments for `heuristics_on_budget` to pass to `process_heuristics_acct`.  Something like `ProcessData` would be more descriptive.

jralls (changes requested):

libgnucash/engine/ScrubBudget.c
christopherlam: `PRId64` errors out. I'll change to `gnc_numeric_to_string`

gnucash/gnome/gnc-budget-view.c
```diff
@@ -1689,3 +1692,30 @@ gnc_budget_view_refresh (GncBudgetView *budget_view)
...
+a budget cell manually. To proceed, we need to set a feature to prevent \
+budget signs errors. This is important to fix inconsistencies in budget \
+reports. This change will cause the data to be unreadable in 3.7 or earlier. \
+If you proceed, It is important to recheck all budgets to ensure the amounts \
+signs are correct.\n\nDo you wish to proceed to set the feature, and allow \
+editing?"))))
+        /* user declined feature. abort editing. */
+        return FALSE;
+
+    /* accept feature. allow editing. */
```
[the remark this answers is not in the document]
christopherlam: That's the confusing part: https://github.com/Gnucash/gnucash/pull/585#pullrequestreview-303462160 - then the user is expected to fix signs manually.

libgnucash/engine/ScrubBudget.c
jralls: You might need to `#include inttypes.h>` to get `PRId64`. `gnc_numeric_to_string` uses `G_GINT64_FORMAT`, but it also allocates so you'll need to assign the string to a variable that you can free after the message.

libgnucash/engine/ScrubBudget.c
```diff
@@ -39,68 +41,70 @@ typedef enum
...
         val = gnc_budget_get_account_period_value (heuristics->budget, account, i);
         total = gnc_numeric_add_fixed (total, val);
     }
 
-    i = gnc_numeric_compare (total, gnc_numeric_zero ());
-    PWARN ("acct=%s, total=%ld/%ld, i=%d", xaccAccountGetName (account),
-           gnc_numeric_num (total), gnc_numeric_denom (total), i);
+    sign = gnc_numeric_compare (total, gnc_numeric_zero ());
+    PWARN ("acct=%s, total=%s, sign=%d",
+           xaccAccountGetName (account), gnc_numeric_to_string (total), sign);
```
jralls: Leaks the string returned by `gnc_numeric_to_string`.

libgnucash/engine/ScrubBudget.c
```diff
@@ -112,29 +116,32 @@ static HeuristicsOutput heuristics_on_budget (GncBudget * budget, Account *root)
 
     gnc_account_foreach_descendant (root, process_heuristics_acct, &heuristics);
 
-    if (heuristics.expense < 0) /* total budget expense is negative */
+    if (heuristics.expense < 0) /* budget expenses are mostly negative */
         result = HEURISTICS_INC_EXP;            /* conclude inc-exp */
-    else if (heuristics.liability > 0) /* total budget liability is positive */
+    else if (heuristics.liability > 0) /* budget liabilitieies are mostly positive */
```
jralls, line 125: lia-what? ;-)

libgnucash/engine/ScrubBudget.h
```diff
@@ -20,21 +20,21 @@
...
-#include "gnc-prefs.h"
+#include <qofbook.h>
 
 /* ================================================================ */
 
-gboolean scrub_budget_signs_check (QofBook *book);
-void scrub_budget_signs_doit (QofBook *book);
+/** analyse book, and return following:
+ * FALSE: don't offer scrubbing
+ * TRUE:  offer scrubbing
```
jralls, line 30:
```
* @param book The book to check
* @return Whether to offer scrubbing
```

jralls (changes requested):
Squash the last two commits so that you don't leave a build failure to screw up bisects.

gnucash/gnome/gnc-budget-view.c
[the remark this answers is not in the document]
christopherlam: FWIW the approach implemented in this branch is as suggested by @gjanssens is to allow the intermediate state - budgets are read-only until auto-fix is allowed, or budget is unlocked for editing. This means both branches of code need to be maintained forever.

My own preference was to remove the intermediate state and enforce auto-fix; this would enable old code handling reversed budgets to be removed.

I think this branch works well.

gnucash/gnome-utils/gnc-file.c
```diff
@@ -1103,9 +1104,29 @@ gnc_post_file_open (GtkWindow *parent, const char * filename, gboolean is_readon
...
+On the other hand if you want to make your own changes to your budgets \
+in 4.x you have to convert it to the new format manually. GnuCash 4.x \
+can't write the old format any more.\n\n Note the conversion may involve \
+a one-time sign correction on your budgeted values. Until you run \
+the conversion some budgeted values may show with wrong signs. \
+However for completeness there is a small chance the budgeted \
+values show up with the wrong sign after the conversion. In \
+that unfortunate case you'll have to manually correct these \
+yourself. \n\nWould you like gnucash to convert \
+your budgets now?" ))))
```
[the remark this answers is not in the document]
christopherlam: Suggestions? For ease of editing, text copied below

"Hi, we notice your budgets are still in 3.x format. 4.x has a new format to store budgets. We can convert automatically to this new format if you like. Beware that once converted you'll need at least gnucash 3.8 be able to open your book. On the other hand if you want to make your own changes to your budgets in 4.x you have to convert it to the new format manually. GnuCash 4.x can't write the old format any more.\n\n Note the conversion may involve a one-time sign correction on your budgeted values. Until you run the conversion some budgeted values may show with wrong signs. However for completeness there is a small chance the budgeted values show up with the wrong sign after the conversion. In that unfortunate case you'll have to manually correct these yourself. 

Would you like gnucash to convert your budgets now?"

gnucash/gnome/gnc-budget-view.c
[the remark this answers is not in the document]
christopherlam: https://code.gnucash.org/logs/2021/01/24.html#T19:17:42

Do we:
1. consider merging this branch, i.e. offer scrubbing; if declined then budgets are read-only until manually fixed?
2. enforce scrubbing and advise manual review of all budgets?
3. offer scrubbing + readonly budgets for 4.x; and enforce scrubbing and remove read-only budgets for 5.x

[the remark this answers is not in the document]
jralls: I'm for #2.

But I think there's a documentation component to this as well: Will anything described in the T&CG [budget chapter](https://www.gnucash.org/docs/v4/C/gnucash-guide/chapter_budgets.html) change? Does it need to explain the impact of different values of the sign reversal option to keep users from being confused?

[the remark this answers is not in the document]
christopherlam: Option 2 is my preference too -- but Geert prefers 1.

I'll review and add another commit to illustrate simplicity of option 2.

The budget chapter could be amended slightly; it'd mean the budget amount signs _will_ be similar to the account signs: e.g. if they expect income accounts to be negative, then budget income amounts will be negative too.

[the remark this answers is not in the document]
gjanssens: I think my opinion weighs less today given my limited availability...

Having said that, I  do prefer some form of safety net for unsuspecting users. It doesn't have to be as suggested in option 1 though.

Here's the use case I want to cover:
- User installs gnucash 5.x without having read the release notes about sign fixes in budgets. User as a sql based book, and was careless enough to not make a backup.
- User opens  book for the first time and is confronted with budget scrubbing.
- Imagine it goes horribly wrong for whatever reason (a corner case we didn't consider for example). Sql backend means changes are applied immediately so there's no easy rollback.
- Lots of user frustration and support time spent to fix this up again.

Of course the user *should* have taken that backup and have read the release notes.

We can make user's lives somewhat more pleasant by informing the user at startup a budget fix is about to happen if user continues with loading this book (even if that will be enforced), we could offer the option to not open the book instead (and to make that backup before trying again). That will give the user a choice:
- continue and take the "risk"
- bail out, make that forgotten backup first and then try again
- bail out and stay on 4.x for some time longer to hear the experience of others first
The dialog itself doesn't have to offer 3 options though. It can shortly explain what's about to happen, make mention of the importance of a good backup and finally ask if the user wants to proceed or quit.

IMHO that could  become a more generic strategy on major version upgrades in the future for any data format conversion we do.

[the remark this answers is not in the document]
jralls: I like that. Gramps does something like "This new version of Gramps uses a new database format and we need to upgrade it. If you don't have a good backup, please make one now before continuing." with Cancel and OK buttons. Cancel doesn't do the upgrade and closes the database. 

The difference for GnuCash in doing something like that is that we tend to do scrubs when using some feature and it might be well into a session. We might need a backend-dependent action in that case where if it's XML we can just do an autosave but if it's SQL the user has to stop and either copy the SQLite3 file or do whatever backup procedure is correct for the server in question. Cancel in our case might just cancel whatever action triggered the scrub rather than closing the file because the rest of the program might work OK.

Where it gets interesting is when the user accepts the first scrub and then encounters a second one and doesn't want to accept it. Depending on what are the first versions that support each feature there's a potential that they'll end up in a situation where they can't go back to the pre-scrub of the second feature because it's blocked by the minimum version of the first.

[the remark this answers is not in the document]
christopherlam: [groan] More accidental complexity.

So: we'd need to overhaul the `gnc_features` feature, ensure all scrubs are centralised and triggered upon load (cf upon first use) and wrap around them to offer the bail-out mechanism?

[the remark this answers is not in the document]
gjanssens: Yes, I think our current strategy to only do conversions the moment a new feature is used for the first time makes our compatibility story very complicated. I have considered it more than once in the past to rework this into the scenario you suggest: run any needed conversion at file load time. We can still keep the "features" concept and only set features that are actually used. For example if a user doesn't have budgets there's no need to set the sign reversal fix flag so older versions of gnucash will continue to work. The only difference is the timing of the conversions.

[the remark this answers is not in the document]
christopherlam: I had a thought how to add feature only if required: add `set_budget_feature (book)` - will count budgets; if >=1 then will set feature, otherwise remove feature:
1. any budget modifier (i.e. create/delete) will call it
2. the budget scrubber, when trigged on file load, will also call it.

jralls:
I think this [bug report](https://bugs.gnucash.org/show_bug.cgi?id=689754) might be related to this PR. @christopherlam do you agree?

christopherlam:
Agree. I've mainly dealt with the regular budget report; the "budget balance sheet" (WTH does it mean anyway) has been off my radar in this PR.

christopherlam:
By Agree I mean it's likely to be fixed by this one-time fix.

jralls:
You've got some test failures to fix:
```
budget.scm
[pass] line:81, test: all display OFF, table has 15 cells
[pass] line:91, test: all display ON, table has 228 cells
[pass] line:94, test: bank
[fail] line:99, test: income
budget/budget.scm
 -> expected: ("Income" "-$55.00 " "1" "-$55.00" "$0.00" "." "$0.00" "." "-$65.00
" "-$67.00" "-$2.00" "-$75.00" "-$77.00" "-$2.00" "." "$0.00" "." "." "$0.00" "."
 "-$195.00" "-$199.00" "-$4.00")
 -> obtained: ("Income" "-$55.00 " "1" "-$55.00" "$0.00" "." "$0.00" "." "-$65.00
" "-$67.00" "$2.00" "-$75.00" "-$77.00" "$2.00" "." "$0.00" "." "." "$0.00" "." "
-$195.00" "-$199.00" "$4.00")
[pass] line:104, test: expense
[pass] line:109, test: budget notes
[pass] line:118, test: only next period - 135 cells
[pass] line:122, test: only next period- bank
[pass] line:130, test: only last period - 102 cells
[pass] line:134, test: only last period - bank
[pass] line:146, test: exact periods - 134 cells
[pass] line:150, test: exact periods - bank
[pass] line:157, test: use accumulated amounts

budget-income-statement.scm
[pass] line:172, test: basic test
[fail] line:176, test: budgeted income amounts
budget/budget-income-statement.scm
 -> expected: ("$195.00" "Income")
 -> obtained: ("-$195.00" "Income")
[fail] line:181, test: net loss for budget
budget/budget-income-statement.scm
 -> expected: ("Net loss for Budget test budget" "$285.00")
 -> obtained: ("Net income for Budget test budget" "$285.00")

budget-balance-sheet.scm
[pass] line:195, test: basic test
[pass] line:199, test: existing assets
[fail] line:203, test: allocated assets
budget/budget-balance-sheet.scm
 -> expected: ("Allocated Assets" "$120.00")
 -> obtained: ("Allocated Assets" "-$120.00")
[fail] line:207, test: unallocated assets
budget/budget-balance-sheet.scm
 -> expected: ("Unallocated Assets" "-$405.00")
 -> obtained: ("Unallocated Assets" "-$165.00")
[pass] line:211, test: total assets
[pass] line:215, test: existing liab
[pass] line:219, test: new liab
[pass] line:223, test: total liab
[pass] line:227, test: retained earnings
[pass] line:231, test: retained losses
[pass] line:235, test: unrealized losses
[pass] line:239, test: existing equity
[pass] line:243, test: new equity
[pass] line:247, test: total equity
[pass] line:251, test: total liab and equity
Source:test-budget.scm
pass = 26, fail = 5
<end of output>
Test time =   0.52 sec
Test Failed.
"test-budget" end time: Mar 05 05:11 PST
"test-budget" time elapsed: 00:00:00
```

libgnucash/engine/ScrubBudget.h
```diff
@@ -0,0 +1,39 @@
...
+ *                                                                  *
+\********************************************************************/
+
+#include <glib.h>
+#include <qofbook.h>
+
+/* ================================================================ */
+
+/** Fix budget signs
+ * A guard is set if we have completed reversal, or there are no
```
jralls, line 29: Say "feature".

But speaking of guards, this header needs them. You do that by adding immediately under the license block
```
#ifndef _GNC_SCRUBBUDGET_H_
#define _GNC_SCRUBBUDGET_H_
```
and then at the end
```
#endif // _GNC_SCRUBBUDGET_H_
```

The purpose is to avoid re-declaration errors if the header is included more than once in a compilation unit.

libgnucash/engine/ScrubBudget.h
```diff
@@ -0,0 +1,39 @@
...
+ * budgets in book.
+ *
+ * @param book The book to scrub
+ *
+ * returns NULL if feature is already set
+ * returns a newly allocated string describing outcome of scrub.
+ */
+gchar * gnc_scrub_budget_signs (QofBook *book);
+
+/* ==================== END OF FILE ==================== */
```
jralls, line 39: Useless noise.

libgnucash/engine/ScrubBudget.c
```diff
@@ -0,0 +1,229 @@
...
+} ReversalType;
+
+typedef struct
+{
+    gint asset,liability,equity,income,expense;
+    gint num_periods;
+    GncBudget* budget;
+} ProcessData;
+
+static gboolean this_book_needs_fixing;
```
jralls, line 59: Not used, remove.

libgnucash/engine/ScrubBudget.c
```diff
@@ -0,0 +1,229 @@
...
+
+    ENTER ("processing budget [%s] for reversal", gnc_budget_get_name (budget));
+    gnc_account_foreach_descendant (root, fix_budget_acc_sign, &reversal);
+    LEAVE ("completed budget [%s] for reversal", gnc_budget_get_name (budget));
+}
+
+/* For GnuCash 4.4 onwards - fix budget signs */
+/* A guard is set if we have completed reversal, or there are no
+   budgets in book. */
+gchar *
```
jralls, line 200: Return a `gboolean`, `TRUE` if we did the scrub.

gnucash/gnome-utils/gnc-file.c
```diff
@@ -1102,9 +1103,20 @@ gnc_post_file_open (GtkWindow *parent, const char * filename, gboolean is_readon
...
     }
 
-    // Fix account color slots being set to 'Not Set', should run once on a book
     qof_event_suspend();
+
+    // Fix account color slots being set to 'Not Set', should run once on a book
     xaccAccountScrubColorNotSet (gnc_get_current_book());
+
+    /* Fix budget signs */
+    budget_scrubbing_outcome = gnc_scrub_budget_signs (new_book);
```
jralls, line 1112: `gnc_post_file_open` is already 10x longer than is reasonable so put the new stuff  into a static function `maybe_scrub_budget_signs` that:
1. Checks the feature and whether there's a budget. It should call `gnc_scub_budget_signs` only if necessary.
2. `gnc_scrub_budget_signs` can return a boolean that's TRUE if it actually found something to scrub.
3. Display a dialog box only if `gnc_scrub_budget_signs` that tells the user that the scrub occurred and that they should check their budgets to make sure it did it right. Don't forget to mark the message for translation. 

Call `maybe_scrub_budget` here.

gnucash/gnome-utils/gnc-file.c
jralls: This whole line of reasoning is overthinking. We do this a lot: Fix a problem, set a feature. We usually don't even tell the user about it, but in this case the fix can misfire so we tell the user to check that it went OK.

jralls (changes requested):
The cleanup yesterday fixed a lot of my code issues so that there's only one serious one left and even that's a simple refactor.

I didn't flag them but there are a few mentions of 4.4 in here and the next release is 4.5, but do we really need to mention versions in comments?

gnucash/gnome-utils/gnc-file.c
christopherlam: Leaving the feature_check and budget_check to `ScrubBudget.c` seems to make more sense to me? See latest.

gnucash/gnome-utils/gnc-file.c
```diff
@@ -700,6 +701,18 @@ gnc_file_query_save (GtkWindow *parent, gboolean can_cancel)
...
 
+static void
+maybe_scrub_budget_signs (QofBook *book, GtkWindow *parent)
+{
+    /* Fix budget signs */
+    if (gnc_scrub_budget_signs (book))
+    {
+        gnc_info_dialog (parent, "%s", _("This book has budgets. \
+The internal representation of amounts is now fixed. Please review \
+budgets and amend signs if necessary.."));
```
jralls, line 712: Typo, two dots at the end.

libgnucash/engine/ScrubBudget.c
```diff
@@ -0,0 +1,205 @@
...
+        return FALSE;
+
+    if (!gnc_budget_get_default (book))
+        return FALSE;
+
+    qof_collection_foreach (qof_book_get_collection (book, GNC_ID_BUDGET),
+                            fix_budget_sign, root);
+
+    gnc_features_set_used (book, GNC_FEATURE_BUDGET_UNREVERSED);
+    return TRUE;
```
jralls, line 203: Is it necessarily true that every current budget needs changing? If not this should be conditional on fix_budget_sign actually doing something.

libgnucash/engine/ScrubBudget.c
```diff
@@ -0,0 +1,205 @@
...
+{
+    Account* root = gnc_book_get_root_account (book);
+    gchar *retval = NULL;
+
+    if (gnc_features_check_used (book, GNC_FEATURE_BUDGET_UNREVERSED))
+        return FALSE;
+
+    if (!gnc_budget_get_default (book))
+        return FALSE;
+
```
jralls, line 198: If these checks are here then this function should be `gnc_scrub_budget_signs_maybe`. IMO that is reason enough to have the checks in `gnc-file.c` instead.

libgnucash/engine/ScrubBudget.c
```diff
@@ -0,0 +1,205 @@
...
+           heuristics.asset, heuristics.liability, heuristics.income,
+           heuristics.expense, heuristics.equity, result);
+
+    return result;
+}
+
+static void
+fix_budget_acc_sign (Account *acc, gpointer user_data)
+{
+    gnc_numeric amt;
```
jralls, line 131: Belongs in the loop block where it's used.

libgnucash/engine/ScrubBudget.c
christopherlam: I'd add the features check in ScrubBudget.c. Because `gnc_maybe_scrub_budget_signs` is a *dangerous* function which can modify budgets, it must bail if the book has already been scrubbed.

libgnucash/engine/ScrubBudget.c
christopherlam: I'll rename the function to `check_and_fix_budget_sign`.

jralls: That doesn't answer the question.

christopherlam: Ok not all budgets need changing. If algorithm concludes `HEURISTICS_NONE` then the signs are not reversed, but the feature must still be set to prevent repeating scrubbing.

jralls:
Right, but then it unnecessarily blocks opening the book on GnuCash < 3.8. I don't know if that's a significant problem or not. @gjanssens what do you think?

gjanssens:
The trade-off we face here is either needlessly blocking users on versions of gnucash older than 3.8 versus repeatedly wasting time on scrubbing in gnucash newer than 4,5.
A few factors that should help our decision:
- how long does scrubbing take ? Is it long enough the user notices a delay ? Is there visual feedback in that case ?
- 3.8 is the version of gnucash that ships with Ubuntu 20.04LTS. So that's an audience that may be switching between gnucash versions because they can also install a more recent flatpak. For 20.04 that's fine, 3.8 is recent enough. People on older LTS's would have 2.6 as system package, so their version is too old to  be compatible with 4.x anyway.

With that in mind I'm tempted to just block anyone on versions older than 3.8 to keep things simple.

christopherlam:
Scrubbing should be very quick. AFAIU all budget-amount getters/setters touch a kvp tree only. This branch doesn't have to go to 4.5 -- 5.x is fine, and more cleanup on master.

christopherlam:
I think this branch is ready to be merged to master. Any pre-5.x budgets will be fixed and set feature; books without budgets will not have feature. 5.x onwards will handle unreversed budgets only.

Pre-5.x only "Budget Report" has ability to handle both reversed & unreversed amounts; after this merge and we can now start fixing other budget reports.

libgnucash/engine/ScrubBudget.c
```diff
@@ -0,0 +1,211 @@
...
+{
+    ProcessData heuristics = {0, 0, 0, 0, 0, gnc_budget_get_num_periods (budget),
+                              budget};
+    SignReversals result;
+
+    gnc_account_foreach_descendant (root, process_heuristics_acct, &heuristics);
+
+    result =
+        heuristics.expense < 0 ? HEURISTICS_INC_EXP :
+        heuristics.liability > 0 ? HEURISTICS_NONE :
```
miklcct, line 116: I suggest we use income rather than liability in the second step of heuristics. This is because a liability check fails when budgeting for a new loan such as mortgage.

In summary:
First we check for expense. If it is negative then we are certain that the setting is reverse income / expense.
Now we just need to distinguish "reverse credit account" and "none", where income is one of them. A positive income means "reverse credit account" is set so we should reverse it.

christopherlam: I think you're right fwiw

christopherlam: On second thoughts, I do not think I will be able to continue this line of work; I do not use budgets at all, and will not want to handle users' bug reports after the budgets fix. I'll be keen to revert this budget scrubbing altogether, including the 4.x handling of unreversed budgets. I'll be happy to reinstate a PR to document the work done, and perhaps someone else may take up the work.

christopherlam: On third thoughts, I think I found the bug in this branch, see f7ee6446 and 85a62c03 just merged in. Please test master again.

PR #589, opened 2019-10-14
 Bug 797453 - Chart of Accounts is slow to update / redraw

Bob-IT:
I think this covers what is required for this Bug. A hash table is used to cache the account model values and so eliminate the constant recalculating of balance/totals and so should be quicker to redraw.
Changes to an account will clear the cache for that account and its parents so values can be updated. Changes to the preference negative colour clears the whole cache.

It does what I think it should and seems quicker and updates to the accounts happened as before in my testing but would appreciate some one else testing / review.

### DIFF
--- a/gnucash/gnome-utils/gnc-tree-model-account.c
+++ b/gnucash/gnome-utils/gnc-tree-model-account.c
@@ -97,6 +97,9 @@ typedef struct GncTreeModelAccountPrivate
     Account *root;
     gint event_handler_id;
     const gchar *negative_color;
+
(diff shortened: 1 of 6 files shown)


Bob-IT:
OK, a couple of things I had not thought off, to get the refresh button on the Chart of Accounts page to clear the cache and what to do when a price is changed / added which would influence the balance / total columns.

Bob-IT:
I have fixed the refresh button and that partly fixed the totals changing with changes in prices, just needed to add a couple more gnc_gui_refresh_all's.

PR #591, opened 2019-10-18
Update hr.po

milotype:
minor changes and adapted to .pot file changes


fellen:
Merged as commit e6875c1, Thanks!

For your convenience I merged in the next commit a recent gnucash.pot, which contained 1 new message.

BTW: You can remove the https://wiki.gnucash.org/wiki/Translation#Orphaned_Records at the end of the file.

PR #592, opened 2019-10-19
[maint] budget code handles both natural and reversed amounts

christopherlam:
I guess this branch will be merged into maint, then merge maint to master, then merge #585 scrubbing function.

Modify maint branch to handle both reversed & unreversed budgets.

Scrubbing existing budget data is not done until master.

Draft. Not sure I've handled all possibilities. But I think this is safe.

### DIFF
--- a/gnucash/gnome/gnc-plugin-page-budget.c
+++ b/gnucash/gnome/gnc-plugin-page-budget.c
@@ -902,7 +902,7 @@ estimate_budget_helper(GtkTreeModel *model, GtkTreePath *path,
                               GNC_HOW_DENOM_SIGFIGS(priv->sigFigs) | 
                               GNC_HOW_RND_ROUND_HALF_UP);
 
-        if (gnc_reverse_balance(acct))
+        if (gnc_reverse_budget_balance(acct, FALSE))
             num = gnc_numeric_neg(num);
 
         for (i = 0; i < num_periods; i++)
@@ -920,7 +920,7 @@ estimate_budget_helper(GtkTreeModel *model, GtkTreePath *path,
 
             if (!gnc_numeric_check(num))
             {
-                if (gnc_reverse_balance(acct))
+                if (gnc_reverse_budget_balance(acct, FALSE))
                     num = gnc_numeric_neg(num);
 
                 num = gnc_numeric_convert(num, GNC_DENOM_AUTO,
@@ -1020,7 +1020,7 @@ allperiods_budget_helper(GtkTreeModel *model, GtkTreePath *path,
 {
     Account *acct;
     guint num_periods, i;
-    gnc_numeric num;
+    gnc_numeric num, allvalue;
     GncPluginPageBudgetPrivate *priv;
     GncPluginPageBudget *page = data;
 
@@ -1028,7 +1028,9 @@ allperiods_budget_helper(GtkTreeModel *model, GtkTreePath *path,
     priv = GNC_PLUGIN_PAGE_BUDGET_GET_PRIVATE(page);
     acct = gnc_budget_view_get_account_from_path(priv->budget_view, path);
     num_periods = gnc_budget_get_num_periods(priv->budget);
-    num = priv->allValue;
+    allvalue = priv->allValue;
+    if (gnc_reverse_budget_balance(acct, TRUE))
+        allvalue = gnc_numeric_neg(allvalue);
 
     for (i = 0; i < num_periods; i++)
     {
@@ -1036,7 +1038,7 @@ allperiods_budget_helper(GtkTreeModel *model, GtkTreePath *path,
         {
         case ADD:
             num = gnc_budget_get_account_period_value(priv->budget, acct, i);
-            num = gnc_numeric_add(num, priv->allValue, GNC_DENOM_AUTO,
+            num = gnc_numeric_add(num, allvalue, GNC_DENOM_AUTO,
                                   GNC_HOW_DENOM_SIGFIGS(priv->sigFigs) |
                                   GNC_HOW_RND_ROUND_HALF_UP);
             gnc_budget_set_account_period_value(priv->budget, acct, i, num);
@@ -1053,7 +1055,7 @@ allperiods_budget_helper(GtkTreeModel *model, GtkTreePath *path,
             break;
         default:
             gnc_budget_set_account_period_value(priv->budget, acct, i,
-                                                priv->allValue);
+                                                allvalue);
             break;
         }
     }
--- a/gnucash/gnome/gnc-budget-view.c
+++ b/gnucash/gnome/gnc-budget-view.c
@@ -94,8 +94,9 @@ enum
 {
     TOTALS_TYPE_INCOME, /**< This total is Income type*/
     TOTALS_TYPE_EXPENSES, /**< This total is Expenses type*/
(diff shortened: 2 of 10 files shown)


christopherlam:
Ok I'll leave this one for now. It would be good to have another tester. I attach a test datafile with gnc-feature enabled (budget amounts have natural signs)

Test the following, using all 3 sign-reversal preferences.
* verify income/equity/expense signs are correct
* verify income/expense/transfer/total amounts are correct
* verify all right-hand-side (i.e. allperiods) totals have correct signs.
* verify entering data is as expected
* verify estimate & all-period budget tasks perform as expected.
* verify budget.scm report is stable as well -- the signs may be +ve/-ve but the difference/actuals  
should be stable
* other budget reports are less tested

Please report any issues here!

christopherlam:
I think this branch is fine except the "Transfers" and "Total Budget" summary rows are now inconsistent. I have no idea how to interpret these numbers. Summary "Income" and "Expense" are ok, "Transfer" and "Total" make no sense.

christopherlam:
I believe this is now done.

gnucash/gnome/gnc-plugin-page-budget.c
```diff
@@ -902,7 +902,7 @@ estimate_budget_helper(GtkTreeModel *model, GtkTreePath *path,
                               GNC_HOW_DENOM_SIGFIGS(priv->sigFigs) | 
                               GNC_HOW_RND_ROUND_HALF_UP);
 
-        if (gnc_reverse_balance(acct))
+        if (gnc_reverse_budget_balance(acct, FALSE))
```
gjanssens: Can you help me understand why only two calls to gnc_reverse_budget_balance use FALSE as second parameter (that is, why they emulate pre-4.x behaviour) ? It's difficult to grasp when not having studied the code fully.

christopherlam: Or another way to consider this is that FALSE calls can be removed altogether in master.

gjanssens: Ok, that makes sense indeed. Conceptually your code looks fine.

I'll try to test-run it one of the coming days.

gjanssens:
I have built gnucash with this PR applied. I'm not a budget user so my testing is only superficial and I don't know exactly what to expect.

Here are my findings:
* Gnucash with a book that doesn't have the new feature flag set seems to behave exactly the same before and after applying your patch: that is there is no visual difference in the budget or budget report visual output regardless of how the sign reversal preference is set.
* With a book that does have the feature flag set, several values will reverse sign based on the sign reversal preference setting. The only ones that don't seem to change are the totals in the budget view (Income, Expense, Asset, Liability, Equity). They appear to remain positive regardless of the sign reversal setting. I don't know if that's intended or not, but to me this seems inconsistent and confusing.
* Sums are stable (other than expected sign reversals) which is good.

christopherlam:
Thanks for short review

> * Gnucash with a book that doesn't have the new feature flag set seems to behave exactly the same before and after applying your patch: that is there is no visual difference in the budget or budget report visual output regardless of how the sign reversal preference is set.

Correct, old books are unchanged in maint.
 
> * With a book that does have the feature flag set, several values will reverse sign based on the sign reversal preference setting. The only ones that don't seem to change are the totals in the budget view (Income, Expense, Asset, Liability, Equity). They appear to remain positive regardless of the sign reversal setting. I don't know if that's intended or not, but to me this seems inconsistent and confusing.

Had your book with feature flag set been processed via #585 to fix the internal values? If yes then the budget editor/viewer *should* be stable. If no, then all bets are off.

A good test should be: create *new* book, manually *add* feature flag, and create a budget using new book data. IMHO the budget editor/viewer (and reports) should be stable.

Alternatively https://github.com/Gnucash/gnucash/pull/592#issuecomment-544150327 has a test *feature-enabled* book with budget data.

christopherlam:
>     * With a book that does have the feature flag set, several values will reverse sign based on the sign reversal preference setting. The only ones that don't seem to change are the totals in the budget view (Income, Expense, Asset, Liability, Equity). They appear to remain positive regardless of the sign reversal setting. I don't know if that's intended or not, but to me this seems inconsistent and confusing.

Ahh I remember I'd *enabled* sign-reversal in the budget-viewer to display signs according to global sign-reversal policy. I figured if user selected 'None' then they'd expect every Equity and Income type amounts to be *negative* including budget-value i.e. change of income during budget period is negative with the 'none' policy. IOW this is by design. The rule would be that budget signs are the same as the main Account tab signs.

christopherlam:
I guess it would be an idea to ensure that the budget editor/viewer always shows amounts assuming 'credit-amounts' reversal policy, yet store amounts unreversed?

christopherlam:
> I guess it would be an idea to ensure that the budget editor/viewer always shows amounts assuming 'credit-amounts' reversal policy, yet store amounts unreversed?

https://bugs.gnucash.org/show_bug.cgi?id=780717#c4 would suggest that this branch behaviour is the correct one.

gjanssens:
> Alternatively [#592 (comment)](https://github.com/Gnucash/gnucash/pull/592#issuecomment-544150327) has a test _feature-enabled_ book with budget data.

JFTR my observations were done with this test book.

gjanssens:
> > * With a book that does have the feature flag set, several values will reverse sign based on the sign reversal preference setting. The only ones that don't seem to change are the totals in the budget view (Income, Expense, Asset, Liability, Equity). They appear to remain positive regardless of the sign reversal setting. I don't know if that's intended or not, but to me this seems inconsistent and confusing.
> 
> Ahh I remember I'd _enabled_ sign-reversal in the budget-viewer to display signs according to global sign-reversal policy. I figured if user selected 'None' then they'd expect every Equity and Income type amounts to be _negative_ including budget-value i.e. change of income during budget period is negative with the 'none' policy.  IOW this is by design.

I follow that design choice, but in my tests the **totals** on the budget page don't follow it. For example if I select 'None' for reversal policy, the income values in the account list will be displayed as negative (good), but the income total remains positive (bad).

christopherlam:
> I follow that design choice, but in my tests the **totals** on the budget page don't follow it. For example if I select 'None' for reversal policy, the income values in the account list will be displayed as negative (good), but the income total remains positive (bad).

Ok I see what you mean now. The *totals* don't get redrawn as long as the Preferences window is open. When Preferences are finalised via OK, the *total* sign will reflect sign-reversal policy. This is beyond my ability to fix.

gjanssens:
Oh, interesting. It didn't occur to me to close the Preferences window. It should be a small change, adding the proper event processing to the totals window. If too difficult for you, feel free to write a followup bug for it. Me or @Bob-IT should be able to get the ironed out.

christopherlam:
Ok if there are no further concerns I'll merge this one soon.

I'll need to clarify there *is* a change to the Totals section - from Income/Expense/Transfers/Totals to 5 account types. I had to change to 5 account types because the previous transfers & totals summation would be completely wrong with unreversed numbers.

If there is outcry at this change then someone will need to figure out how to calculate the 'transfers' and 'totals' section, using an appropriate algorithm...

gjanssens:
Agreed

christopherlam:
Latest push was a rebase only... Still haven't quite got the hang of `--ff-only` or `--no-ff` :neutral_face:

gjanssens:
Don't worry. That doesn't matter too much :)

PR #593, opened 2019-10-19
Fixed German IBAN error message

stovocor:


cstim:
Thanks for spotting this!

PR #595, opened 2019-10-24
Fix bug 743943

mauritslamers:
An attempt at fixing Gnucash#743943. The approach taken is to define parametrized strings and perform the actual translation depending on the invoice type and update of the actions on every call to `gnc_plugin_page_invoice_update_menus` accordingly.
The PR also include one tiny edit in the documentation to add a word I noticed was missing.

### DIFF
--- a/gnucash/gnome/gnc-plugin-page-invoice.c
+++ b/gnucash/gnome/gnc-plugin-page-invoice.c
@@ -139,12 +139,12 @@ static GtkActionEntry gnc_plugin_page_invoice_actions [] =
     },
     {
         "EditPostInvoiceAction", GNC_ICON_INVOICE_POST, N_("_Post Invoice"), NULL,
-        N_("Post this Invoice to your Chart of Accounts"),
+        N_("Post this invoice to your Chart of Accounts"),
         G_CALLBACK (gnc_plugin_page_invoice_cmd_post)
     },
     {
         "EditUnpostInvoiceAction", GNC_ICON_INVOICE_UNPOST, N_("_Unpost Invoice"), NULL,
-        N_("Unpost this Invoice and make it editable"),
+        N_("Unpost this invoice and make it editable"),
         G_CALLBACK (gnc_plugin_page_invoice_cmd_unpost)
     },
 
@@ -195,19 +195,19 @@ static GtkActionEntry gnc_plugin_page_invoice_actions [] =
     /* Business menu */
     {
         "BusinessNewInvoiceAction", GNC_ICON_INVOICE_NEW, N_("New _Invoice"), "",
-        N_("Create a new Invoice for the same owner as the current one"),
+        N_("Create a new invoice for the same owner as the current one"),
         G_CALLBACK (gnc_plugin_page_invoice_cmd_new_invoice)
     },
     {
         "ToolsProcessPaymentAction", GNC_ICON_INVOICE_PAY, N_("_Pay Invoice"), NULL,
-        N_("Enter a payment for the owner of this Invoice"),
+        N_("Enter a payment for the owner of this invoice"),
         G_CALLBACK (gnc_plugin_page_invoice_cmd_pay_invoice)
     },
 
     /* Reports menu */
     {
         "ReportsCompanyReportAction", NULL, N_("_Company Report"), NULL,
-        N_("Open a company report window for the owner of this Invoice"),
+        N_("Open a company report window for the owner of this invoice"),
         G_CALLBACK (gnc_plugin_page_invoice_cmd_company_report)
     },
 };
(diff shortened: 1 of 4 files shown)


gnucash/gnome/gnc-plugin-page-invoice.c
```diff
@@ -421,6 +467,30 @@ gnc_plugin_page_invoice_update_menus (GncPluginPage *page, gboolean is_posted, g
...
+
+
+    for (i = 0; i < parametrized_n_actions; i++)
+    {
+        const gchar *action_name = parametrized_actions[i];
+        // need action by name, check whether in plugin page actions
+        for (j = 0; j < gnc_plugin_page_invoice_n_actions; j++)
+        {
+            action = gnc_plugin_page_invoice_actions[j];
+            if (action_name == action.name)
```
gjanssens: Minor performance consideration: you are iterating more than is needed. If the correct action is found, break out of the inner loop instead of continuing to scan the remaining actions.

More optimization could be achieved if you only loop over the plugin page actions once. That would require another approach though. As we are deeply in gui land here (using gtk and gtkactions) one way to achieve this would set a custom property on your parameterized actions (they are gobjects, so you can [add arbitrary data](https://developer.gnome.org/gobject/stable/gobject-The-Base-Object-Type.html#g-object-set-dataurl)). That custom property could be a gint with value 0 for a normal action or 1 for a parameterized option. I'm deliberately suggesting an int as that is one of the easiest to set/get as object data using `GINT_TO_POINTER` and `GPOINTER_TO_INT`
You can then loop over the plugin page's action, test the custom property and if it's 1, do the parameter substitution dance. For readability though it's probably best to use an enum and give the numbers 0 and 1 a clear enum name.

gnucash/gnome/gnc-plugin-page-invoice.c
```diff
@@ -421,6 +467,30 @@ gnc_plugin_page_invoice_update_menus (GncPluginPage *page, gboolean is_posted, g
...
+            {
+                // found the action, update label and tooltip
+                const gchar *parametrized_label = action.label;
+                const gchar *parametrized_tooltip = action.tooltip;
+                GtkAction *gtk_action = gtk_action_group_get_action (action_group, action_name);
+                gtk_action_set_label(gtk_action,
+                                    g_strdup_printf ( _(parametrized_label),
+                                                    invoice_type_name));
```
gjanssens: I'm pretty sure this leaks memory. A GtkAction takes a `const gchar*`, which typically means it internally makes a copy of the string and won't further touch the pointer passed in. However g_strdup_printf allocates a new string and this string is never freed.

gnucash/gnome/gnc-plugin-page-invoice.c
```diff
@@ -194,20 +193,20 @@ static GtkActionEntry gnc_plugin_page_invoice_actions [] =
...
+        "ToolsProcessPaymentAction", GNC_ICON_INVOICE_PAY, N_("_Pay %s"), NULL,
         N_("Enter a payment for the owner of this Invoice"),
         G_CALLBACK (gnc_plugin_page_invoice_cmd_pay_invoice)
     },
 
     /* Reports menu */
     {
         "ReportsCompanyReportAction", NULL, N_("_Company Report"), NULL,
-        N_("Open a company report window for the owner of this Invoice"),
+        N_("Open a company report window for the owner of this %s"),
```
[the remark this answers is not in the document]
fellen: No, both are female: 
https://www.duden.de/rechtschreibung/Rechnung, 
https://www.duden.de/rechtschreibung/Gutschrift.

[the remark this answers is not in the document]
gjanssens: > No, both are female:
> https://www.duden.de/rechtschreibung/Rechnung,
> https://www.duden.de/rechtschreibung/Gutschrift.

Ok. So google translate tricked me. The issue remains the same though only the example is invalid.

gnucash/gnome/gnc-plugin-page-invoice.c
mauritslamers: I was aware it could be leaking, but I saw things similar to this elsewhere in the gnucash code, and I hoped (rather than assumed) that somehow the freeing would be done on update automatically. The rewrite to static strings made the use of `g_strdup_printf` no longer necessary.

gnucash/gnome/gnc-plugin-page-invoice.c
mauritslamers: True, I was planning to do that, but forgot to put it in.

gnucash/gnome/gnc-plugin-page-invoice.c
```diff
@@ -195,7 +195,7 @@ static GtkActionEntry gnc_plugin_page_invoice_actions [] =
     /* Business menu */
     {
         "BusinessNewInvoiceAction", GNC_ICON_INVOICE_NEW, N_("New _Invoice"), "",
-        N_("Create a new invoice for the same owner as the current one"),
+        N_("Create a new Invoice for the same owner as the current one"),
```
fellen: lowercase in English

gnucash/gnome/gnc-plugin-page-invoice.c
```diff
@@ -106,7 +106,7 @@ static GtkActionEntry gnc_plugin_page_invoice_actions [] =
         G_CALLBACK (gnc_plugin_page_invoice_cmd_new_account)
     },
     {
-        "FilePrintAction", "document-print", N_("Print Invoice"), "<primary>p",
+        "FilePrintAction", "document-print", N_("_Print Invoice"), "<primary>p",
```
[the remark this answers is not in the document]
mauritslamers: Was intending to do exactly that, but obviously didn't do it. Fixed in latest commit.

gnucash/gnome/gnc-plugin-page-invoice.c
```diff
@@ -261,22 +261,113 @@ static const gchar *can_unpost_actions[] =
...
+static action_toolbar_labels invoice_action_labels[] =
+{
+    {"FilePrintAction", N_("_Print Invoice")},
+    {"EditEditInvoiceAction", N_("_Edit Invoice")},
+    {"EditDuplicateInvoiceAction", N_("_Duplicate Invoice")},
+    {"EditPostInvoiceAction", N_("_Post Invoice")},
+    {"EditUnpostInvoiceAction", N_("_Unpost Invoice")},
+    {"BusinessNewInvoiceAction", N_("New _Invoice")},
+    {"ToolsProcessPaymentAction", N_("_Pay Invoice")}
+};
```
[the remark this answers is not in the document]
mauritslamers: Definitely safer. Fixed in latest commit.

Bob-IT:
@mauritslamers Hi, just wondering if you will have time to make the final changes @gjanssens  asked for, would be nice to get this in 3.8 which is due for release on 29/12/2019 and I may need to add to it for version 4.0

mauritslamers:
@Bob-IT I will try to get it done before 24/12

mauritslamers:
@Bob-IT I tried to implement @gjanssens suggestions using `git rebase -i`. I didn't see a way to properly split up commit 7, so I followed his suggestion to squash it with the others, and add a note in the commit message. Could you please confirm whether this PR can now be approved as is?

gjanssens:
Looks good except for the commit message on the commit fixing the bug.
This is partially my fault: I now see github didn't like my markdown text where I explained what I expected the commit message to look like.
It should have said:
`Bug <number> - <bug title>`
Sorry about that.
Can you still correct that bit ?

mauritslamers:
@gjanssens Updated. Thanks for teaching me about `git rebase -i`.

gjanssens:
That was it; thanks a lot and congratulations on your first accepted contribution :)

PR #598, opened 2019-10-30
Implemented quick&dirty fix for a regression with aqbanking6.

ingoha:
Until now, there is now bug filed due to time limitations on my side :-(

### DIFF
--- a/gnucash/import-export/import-account-matcher.c
+++ b/gnucash/import-export/import-account-matcher.c
@@ -86,7 +86,9 @@ static gpointer test_acct_online_id_match(Account *acct, gpointer param_online_i
     const gchar * current_online_id = gnc_import_get_acc_online_id(acct);
     if ( (current_online_id != NULL
             && param_online_id != NULL )
-            && strcmp( current_online_id, param_online_id ) == 0 )
+            // discard any extra characters of the passed online_id
+            // as some (German) banks tack stuff to it when using HBCI/FinTS
+            && strncmp( current_online_id, param_online_id, strlen(current_online_id ) == 0 )
     {
         return (gpointer *) acct;
     }


jralls:
Does that EUR AQB6 is tacking on to the account id by any chance  indicate a Euro account? If so the hack won't work if the account is instead denominated in a different currency, say USD .

cstim:
There was some discussion on gnucash-de, resulting in this statement by someone who was running into this bug: https://lists.gnucash.org/pipermail/gnucash-de/2019-October/011033.html  The last paragraph says "I'm using a local workaround: removing all non-digits from the end". @ingoha Do you think you could implement this as well?
Also, I think https://bugs.gnucash.org/show_bug.cgi?id=797432 is caused by the same issue, so you could reference that bug as well.

jralls:
@cstim removing all non-digits seems risky: It assumes that the bank in question uses only numerics in its online account id. Remember that this code operates on *all importers except QIF* so any code here can't be AQB-specific. Do you know the answer to my question about why AQB is tacking "EUR" onto the end of @ingoha's online-account-id?

Perhaps this problem is actually caused by the two calls to gnc_import_select_account in gnc-aqb-utils.c calling different AQB account number functions (`AB_ImExporterAccountInfo_GetAccountNumber` in `gnc_ab_accinfo_to_gnc_acc` and `AB_Transaction_GetLocalAccountNumber` in `gnc_ab_txn_to_gnc_acc`)? Regardless, since it's an AQB problem maybe a better place to hack a fix (assuming that an actual fix is too hard) would be in one of those two functions instead of in the general-purpose `import-account-matcher.c`.

ingoha:
I think @jralls is perfectly right; I did not think very much about the consequences of my patch for other settings.
As a matter of fact, I did search for the root cause of the EUR at the end quite a time. So the question would be: Why are the two different AQB account number functions called?
btw: Reading through the AQB documentation, the functions used here are not stable yet :-/

ingoha:
My purpose: Add an abstraction layer in gnc-aqb-utils.c; e.g. a new function called gnc_ab_generate_online_id and do the lifting there. What do you think?

jralls:
I think that generating the account is AQB's job and that it should do so consistently.

After looking at the code a bit more I see why different functions are called: One is used in a callback on an account list iteration, the other on a transaction list iteration. Unless there's AQB API that works in both contexts you're not going to be able to make a single function that will work in both.

Repeated traversals of the account tree are absolutely *not* the right answer. That's expensive and should be done only once.

Maybe the best course would be to look at Martin's AqFinance code to see how he reconciles the two.

cstim:
Martin said (in the same thread, see https://lists.gnucash.org/pipermail/gnucash-de/2019-September/011028.html ) he recommends using the settings from the AccountInfo. But I haven't checked the code in details myself so far.

jralls:
I don't see any way to get the AccountInfo from the transaction. AB_Transaction exposes set and get functions for uniqueAccountId, localAccountNumber, and remoteAccountNumber, nothing about AccountInfo. It looks like `SetLocalAccountNumber` is called by several of the imexporters, SetRemoteAccountNumber is set only by eri2 and openhbci1, and `SetUniqueAccountId` isn't called by any of them. Interestingly neither swift nor sepa call `SetLocalAccountNumber`.

Reading the whole thread that @cstim referenced it seems that different banks stick different bits of junk onto the end of the account number.

We could change `test_acct_online_id_match` to use `strstr(param_online_id, current_online_id) == current_online_id` instead of `strcmp(current_online_id, param_online_id) == 0`, the idea being that if it's true then the `param_online_id` is `current_online_id` plus some more characters and the accounts match. That does depend on the book not having two online account ids one of which is a substring of the other.  I guess we could do that, then double-check it with `strcmp` and if that fails stow the match in a list and keep looking. If we later find a perfect match then we quit and use that. If we have only one imperfect match then we're good and use it, and if there's more than one imperfect match we put up a dialog box and ask the user.

ingoha:
OMG, looking deeper makes the problem way bigger ;-) 
btw: I have found that another bank puts junk also after the bank code...
The solution proposed by @jralls seems reasonable. 
However, this idea could be a bit simpler:
* Change the way the online_id is created in the first place (from the account) and stick a dividing character between bank code and account number (e.g. |)
* When matching a transaction to an online_id (created as described above), test if a) the account's bank code is a substring of the transaction's and b) the account's account number is a substring of the transaction's. This should solve all problems in an elegant way, or am I wrong?

btw: I volunteer to implement it :-)

jralls:
@ingoha that seems reasonable. The only problem will be compatibility with existing files that already have the online-id with no separator. 

I guess that the account-info stage is done to configure AQBanking to use the right user information, so perhaps we could cache the bank id and account id from that step and use them to check the numbers from the transaction instead of the Gnucash Account's online-id? Is it possible for a FinTS "get transactions" to retrieve transactions for more than one account?

ingoha:
> @ingoha that seems reasonable. The only problem will be compatibility with existing files that already have the online-id with no separator.
Yes, I will think about that - in the worst case we would only handle online_ids in the "new" format using the new algorithm. Users would then have to renew the link from gnucash account to AQB account thus updating the online_id. Perhaps we should use a new slot named e.g. online_id_new?
> 
> I guess that the account-info stage is done to configure AQBanking to use the right user information, so perhaps we could cache the bank id and account id from that step and use them to check the numbers from the transaction instead of the Gnucash Account's online-id? 
Yeah, but where should they be stored?
> Is it possible for a FinTS "get transactions" to retrieve transactions for more than one account?
I don't think so (looking at the AQB API).

jralls:
> Yeah, but where should they be stored?
In the `GncABImExContextImport`, which is available in `txn_transaction_cb` that calls `gnc_ab_txn_to_gnc_acc` and to `txn_accountinfo_cb` that calls `gnc_ab_accinfo_to_gnc_acc`.

A more radical suggestion:  If AQBanking can import transactions for only a single account, don't bother with `gnc_ab_txn_to_gnc_acc` at all and just use the account already stored in the `GncABImExContextImport`; it looks like that should be the [fallback anyway](https://github.com/Gnucash/gnucash/blob/master/gnucash/import-export/aqb/gnc-ab-utils.c#L718).

cstim:
> A more radical suggestion: If AQBanking can import transactions for only a single account, don't bother with `gnc_ab_txn_to_gnc_acc` at all and just use the account already stored in the `GncABImExContextImport`

No, this is not the case. Aqbanking can very well import transactions for a set of multiple accounts. Gnucash only doesn't use this feature so far. 

@ingoha If a new solution fits better but will require people to newly select the account for one single time, I think this is totally fine and doesn't need extra effort. Only if that new choice isn't stored for whatever reason we would have a problem. So just take the existing kvp slot.

ingoha:
Hmm, perhaps we don't even need the divider inside the online_id. As it is composed of the bank code (BLZ) and the account number, we know where the divider is: The BLZ is always 10 digits long :-)

jralls:
Setting aside that US routing ids are 9 digits, the problem isn't just deciding where the bankid ends, it's also deciding where the account id starts.

However, if we add the bankid and account id to `struct _GncABImExContextImport` then we can use them to test the same info coming from the transaction.  `gnc_ab_txn_to_gnc_txn` has the numbers separately. If they both start with the saved numbers then we use the saved `Account*`--no need to search again--and if not we log an error and skip the transaction. Future direction would be to signal a need to get the second account info using `gnc_ab_accinfo_to_gnc_acc` and maintain an array (or better, convert to C++ and use a std::vector) of `struct _GncABImExContextImport` to pick the `Account*` to set as the primary account for the transaction.

ingoha:
Ok, @jralls thanks for the clarification. I will try to use the method described by Martin on the mailing list. Working on the patch...

ingoha:
While working on the fix, I got another idea: what about storing the IBAN instead of the account number in the `accountid` slot?
This should always match with `AB_Transaction_GetLocalIban`...

jralls:
@ingoha This code needs to work with OFX too. OFX doesn't know about IBAN.

You're also assuming that no bank tacks on noise or extra info onto the IBAN field in the transfer or that `AB_Transaction_getLocalIban` filters it back out if it does. Do you have reason to believe that assumption is correct?

cstim:
> what about storing the IBAN instead of the account number in the `accountid` slot?
> This should always match with `AB_Transaction_GetLocalIban`
In the old times there hasn't been any IBAN, so that's why the old code for creating the online_id didn't use it. If it is there, then yes, indeed it could be used. Then again, I don't know what the aqbanking ofxconnect code inserts into the iban field.

ingoha:
My current code (cf. files of this PR) does only change stuff in the AQB-specific part of Gnucash. It fixes my problems with both banks (comdirect which tags EUR at the end of the account id and DKB which tags two zeros at the end of the bank code).
Is this enough for a bug fix? Should I add a condition to exclude OFX?

jralls:
@ingoha, AQB supports OFX DirectConnect, EBICS, HBCI, and Paypal so GnuCash's interface needs to either be generic enough to work for all 4.

gnucash/import-export/aqb/gnc-ab-utils.c
```diff
@@ -690,6 +692,19 @@ gnc_ab_txn_to_gnc_acc(GtkWidget *parent, const AB_TRANSACTION *transaction)
...
+    if(gnc_acc)
+    {
+      bankcode = gnc_ab_get_account_bankcode(gnc_acc);
+      accountnumber = gnc_ab_get_account_accountid(gnc_acc);
+      online_id = g_strconcat(bankcode ? bankcode : "",
+                                accountnumber ? accountnumber : "",
+                                (gchar*)NULL);
+    }
+    #endif
+
```
jralls: @ingoha, You get the `bankcode` and `accountid` from `gnc_acc`'s KVP, construct an `online_id` from them, then search *the entire account tree* for an account with the matching KVP. Expecting to find what, if not `gnc_acc`?

ingoha: Sorry, I took some days off 8-)
@jralls Thanks for the comment - I certainly suffered from tunnel vision while coding this...

So could the algorithm (for importing a transaction) be: If gnc_acc contains bankcode and accountnumber (it does for a SWIFT transaction), use it. Else try to find the right gnc_acc using the online_id as before?

jralls: @ingoha, You slacker! ;-)
Using the `online_id` *is* effectively using the bank code and account number. The `online_id` is just a concatenation of the two. All import methods do that; the difference is where the numbers come from.  Either way, that's not what the problem is. The problem is that the bank code and account number that AQBanking retrieves from the transactions has extra characters on it so it doesn't match the bank code and account number that AQBanking retrieves when setting up a userid so GnuCash fails to match them.

Each AQB session retrieves a bank code and account number with `AB_ImExporterAccountInfo` so we have the undecorated versions available, and we use that to find an `Account*`. The simplest fix would be to stop there and just use that as the account for the import. However that would break if the import contains transactions for a different account, so perhaps we should cache the bank id and account numbers from that and check that each transaction's bank id and account number begin with exactly those values. I proposed something link that almost 3 weeks ago and I don't think that either you or @cstim have suggested anything else that would work.

We also need to decide what to do when there's a mismatch. One option is to log an error in the tracefile and maybe raise a dialog at the end of the import telling the user that some of the transactions were dropped for having the wrong account. Another is to cycle through all of the AQB user ids, retrieving their bank codes and account numbers and doing the same begins-with match on each and if/when we find a match combine them and search for a matching account. It being possible that we wouldn't find one we'd still need to log an error and raise a warning dialog at the end.

SimJoSt:
Is there a workaround users can use to get this working without new code? Like editing a config file by hand or something similar?

Additionally after setting up the accounts in the aqbanking wizard and the association of the online account with an account in GnuCash doesn't work. It does save the necessary information in the `.gnucash` file, but after opening the association screen again, it shows up like it's not configured. Also all the online actions options are grayed out/disabled. Is this a known side effect of this issue or something else entirely?

Using version 2019-11-23-git-3.7-239-g17d393886+

cstim:
After observing this ongoing discussion for a while without a really good general solution and recognizing continuous bugreports on gnucash-de for German bank accounts, I'm inclined to agree to the very first workaround: We're talking about a problem which currently only occurs for aqbanking-6 and HBCI users (which, by definition, is available only in Germany). Hence for now I would agree to simply check for this trailing currency symbol, which for 99% of the cases is "EUR". Hence we would improve the situation for 99% of the users by simply checking for ends_with("EUR") and removing this suffix. I'd stop looking for a more general solution until other users start bugreports for different combination, but for now we'd have at least a solution for those who currently complain. Thanks!

ingoha:
I would agree, just to be sure: HBCI is only available in Germany? Then I could add a check to activate the workaround only in an HBCI context...

jralls:
Yes as far as anyone knows FinTS (formerly HBCI) is only Germany. ISTM if the messed up bankid is really rare you could just change https://github.com/Gnucash/gnucash/blob/maint/gnucash/import-export/import-account-matcher.c#L89
from
```
&& strcmp( current_online_id, param_online_id ) == 0 )
```
to
```
&& strncmp( current_online_id, param_online_id, strlen(current_online_id ) == 0 )
```
That will discard the tail of param_online_id and be more general and faster than looking for a special string, and safer against the potential (though admittedly unlikely) case of an account id that actually ends in 'EUR'. It is also safe for non-FinTS backends.

Andre-J:
Has anyone taken an eye on SubAccountIds? As far as I know Commerzbank uses SubAccounts under the same account number. So just using the account number (with different SubAccountId) you may run into problems.
See https://lists.gnucash.org/pipermail/gnucash-de/2019-November/011243.html for an example.

jralls:
Great. Those are exposed in AB_AccountInfo but not on AB_Transaction.

ingoha:
Okay, I propose to just close the loop on this PR and come back to the original title: q&d fix. @Andre-J please file a bug for the subAccount problem (is it a problem?).

With the change in https://github.com/Gnucash/gnucash/blob/maint/gnucash/import-export/import-account-matcher.c (@jralls), the problem with suffixes on the account_id would be fixed for the 99% of German users mentioned by @cstim.

However, another problem persists: E.g. the DKB bank adds junk that ends up in the _middle_ of the online_id. To solve this, the comparison between the transaction's and the account's accountnumber should be: If txn_accountnumber is longer than acc_accountnumber, do a substring search inside the txn_accountnumber for the acc_accountnumber. This would also solve the "suffix problem".

Any thoughts on this?

jralls:
@ingoha it's what I told you a month ago and reiterated last week. 

In the absence of anything else useful happening I've pushed my `import-account-matcher.c` work around. The more complete solution requires quite a bit more work that I don't have time for right now.

sauerburger:
@ingoha I'm one of the 1% with a DKB account. Is there a dirty solution where I change the config files manually to match the  transaction's and the account's accountnumber?

ingoha:
I am afraid no :-( That has to be changed in the code. As soon as I have time I will update the package in my PPA and incorporate the fix by @jralls which solved the DKB problem for me.

PR #599, opened 2019-11-03
UI enhancements

christopherlam:
I think the ideas are sound. Need code quality review please.

1. UI jump from any split which has an associated invoice, instead of the awkward Business / Customer / Find Invoice menu. This commit *could* be modified to call function defined in `gnc-split-reg.c` similar to `gnc_plugin_page_register_cmd_execassociated_transaction` if wished.

2. When applying payment, warn if there are no documents selected.

screenshot1
screenshot2


gnucash/gnome/gnc-plugin-page-register.c
```diff
@@ -184,6 +185,7 @@ static void gnc_plugin_page_register_cmd_transaction_report (GtkAction *action,
 static void gnc_plugin_page_register_cmd_associate_file_transaction (GtkAction *action, GncPluginPageRegister *plugin_page);
 static void gnc_plugin_page_register_cmd_associate_location_transaction (GtkAction *action, GncPluginPageRegister *plugin_page);
 static void gnc_plugin_page_register_cmd_execassociated_transaction (GtkAction *action, GncPluginPageRegister *plugin_page);
+static void gnc_plugin_page_register_cmd_execassociated_invoice (GtkAction *action, GncPluginPageRegister *plugin_page);
```
Bob-IT: Is this a good name, maybe ...cmd_open_associated_invoice or jump

gnucash/gnome/gnc-plugin-page-register.c
```diff
@@ -4307,6 +4325,25 @@ gnc_plugin_page_register_cmd_execassociated_transaction (GtkAction *action,
...
+{
+    GncPluginPageRegisterPrivate *priv = GNC_PLUGIN_PAGE_REGISTER_GET_PRIVATE(plugin_page);
+    SplitRegister *reg = gnc_ledger_display_get_split_register (priv->gsr->ledger);
+    Split *split = gnc_split_register_get_current_split (reg);
+    Transaction *trans = xaccSplitGetParent(split);
+    GncInvoice *invoice = gncInvoiceGetInvoiceFromTxn(trans);
+
+    ENTER("(action %p, plugin_page %p)", action, plugin_page);
+
+    g_return_if_fail(GNC_IS_PLUGIN_PAGE_REGISTER(plugin_page));
```
Bob-IT: You are already using plugin_page for GncPluginPageRegisterPrivate so the g_return_if_fail is not doing its job. See the blank transaction function for example.

christopherlam: so, remove this line?

gnucash/gnome/gnc-plugin-page-register.c
```diff
@@ -1009,6 +1022,11 @@ gnc_plugin_page_register_ui_update (gpointer various, GncPluginPageRegister *pag
                                          "ExecAssociatedTransactionAction");
     gtk_action_set_sensitive (GTK_ACTION(action), (uri && *uri));
 
+    /* Set 'ExecAssociatedInvoice' */
+    inv = gncInvoiceGetInvoiceFromTxn(trans);
+    action = gnc_plugin_page_get_action (GNC_PLUGIN_PAGE(page), "ExecAssociatedInvoiceAction");
+    gtk_action_set_sensitive (GTK_ACTION(action), (!(!inv)));
```
Bob-IT: You have two ! here, I am not sure about this, not come across this in existing files.

gnucash/gnome/gnc-plugin-page-register.c
Bob-IT: No, do this...
    GncPluginPageRegisterPrivate *priv;
    SplitRegister *reg;
    Split *split;
    Transaction *trans;
    GncInvoice *invoice;

    ENTER("(action %p, plugin_page %p)", action, plugin_page);

    g_return_if_fail(GNC_IS_PLUGIN_PAGE_REGISTER(plugin_page));

   priv = GNC_PLUGIN_PAGE_REGISTER_GET_PRIVATE(plugin_page);
   reg = gnc_ledger_display_get_split_register (priv->gsr->ledger);

etc...

and you could just do this if you wanted...
    invoice = gncInvoiceGetInvoiceFromTxn (xaccSplitGetParent (gnc_split_register_get_current_split (reg)));

and get rid of 'split' and 'trans' unless you want to test for a valid 'split' and 'trans'

gnucash/gnome/gnc-plugin-page-register.c
christopherlam: I remember this trick from somewhere... converts a valid invoice to TRUE and null-invoice to NULL

gjanssens:
This unfortunately went in too fast.
I'm fine with the new jump functionality, I'm not so happy with the warning dialog being added.
I went to great lengths to make the payment experience interrupt free. That is, it will not pop up any extra dialog unless absolutely needed.
So rather than having a pop up, I would prefer to see a warning icon at the bottom of the screen. Pointing your mouse at this icon will show a tooltip with the warning. That's where I left off a while ago. You should find an example of its use somewhere in the dialog-payment code. A further improvement would be to provide a status field to display any warnings. But please avoid popping up dialog boxes as it is highly interruptive.

christopherlam:
> You should find an example of its use somewhere in the dialog-payment code. A further improvement would be to provide a status field to display any warnings. But please avoid popping up dialog boxes as it is highly interruptive.

That's very fair. My initial thought was to explicitly have a checkbox 'Do not assign to any invoices' which would disable the documents list, and the OK button would be disabled until either checkbox was ticked, or documents were selected, but agree with your plan. I'll see what I can find.

christopherlam:
Found the warning. I'm not sure I'm best placed to modify as above. The following can retain icon but allow payment; not sure how to show proper status field.

```patch
diff --git a/gnucash/gnome/dialog-payment.c b/gnucash/gnome/dialog-payment.c
index b1f96523a..95b48b58f 100644
--- a/gnucash/gnome/dialog-payment.c
+++ b/gnucash/gnome/dialog-payment.c
@@ -218,6 +218,8 @@ gnc_payment_window_check_payment (PaymentWindow *pw)
     const char *conflict_msg = NULL;
     gnc_numeric amount_deb, amount_cred;
     gboolean enable_xfer_acct = TRUE;
+    gboolean OK_enabled = TRUE;
+    GtkTreeSelection *selection;
 
     if (!pw)
         return FALSE;
@@ -226,6 +228,7 @@ gnc_payment_window_check_payment (PaymentWindow *pw)
     if (!pw->post_acct)
     {
         conflict_msg = _("You must enter a valid account name for posting.");
+        OK_enabled = FALSE;
         goto update_cleanup;
     }
 
@@ -234,6 +237,7 @@ gnc_payment_window_check_payment (PaymentWindow *pw)
     if (!gncOwnerIsValid(&pw->owner))
     {
         conflict_msg = _("You must select a company for payment processing.");
+        OK_enabled = FALSE;
         goto update_cleanup;
     }
 
@@ -255,9 +259,21 @@ gnc_payment_window_check_payment (PaymentWindow *pw)
         if (!pw->xfer_acct)
         {
             conflict_msg = _("You must select a transfer account from the account tree.");
+            OK_enabled = FALSE;
+            goto update_cleanup;
         }
     }
 
+    /* this last test checks whether documents were selected. if none,
+       emit warning but still allow as an unattached payment. */
+    selection = gtk_tree_view_get_selection (GTK_TREE_VIEW(pw->docs_list_tree_view));
+    if (gtk_tree_selection_count_selected_rows (selection) == 0)
+    {
+        conflict_msg = _("No documents were selected to assign this payment to. This may create an unattached payment.");
+        OK_enabled = TRUE;
+    }
+
+
 update_cleanup:
     gtk_widget_set_sensitive (pw->acct_tree, enable_xfer_acct);
 
@@ -276,15 +292,14 @@ update_cleanup:
     if (conflict_msg)
     {
         gtk_widget_show (pw->payment_warning);
-        gtk_widget_set_sensitive (pw->ok_button, FALSE);
         return FALSE;
     }
     else
     {
         gtk_widget_hide (pw->payment_warning);
-        gtk_widget_set_sensitive (pw->ok_button, TRUE);
     }
 
+    gtk_widget_set_sensitive (pw->ok_button, OK_enabled);
     return TRUE;
 }
```

gjanssens:
Thanks. That would stay within the spirit of the current behaviour. So ok with me. The status area can be added later as a different feature.

PR #600, opened 2019-11-03
Better UI Refinements

christopherlam:
A better #599 - Show warning if no documents were assigned to payment, and show warnings in gtklabel instead of a tooltip. Warning icon is still shown/hidden appropriately.
status-msg

### DIFF
--- a/gnucash/gnome/dialog-payment.c
+++ b/gnucash/gnome/dialog-payment.c
@@ -218,6 +218,8 @@ gnc_payment_window_check_payment (PaymentWindow *pw)
     const char *conflict_msg = NULL;
     gnc_numeric amount_deb, amount_cred;
     gboolean enable_xfer_acct = TRUE;
+    gboolean allow_payment = TRUE;
+    GtkTreeSelection *selection;
 
     if (!pw)
         return FALSE;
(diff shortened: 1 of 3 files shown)


christopherlam:
This needs a review. When payment is unattached, it currently returns FALSE which is only used in https://github.com/Gnucash/gnucash/blob/maint/gnucash/gnome/dialog-payment.c#L1042 seemingly to decide on auto-click OK. Is this acceptable?

christopherlam:
> When payment is unattached, it currently returns FALSE which is only used in https://github.com/Gnucash/gnucash/blob/maint/gnucash/gnome/dialog-payment.c#L1042 seemingly to decide on auto-click OK. Is this acceptable?

Looks like double-click on Transfer Account (i.e. asset/liability) is considered equivalent to OK. So I think it is we should return TRUE for unattached payments, and user doubleclicks a transfer account.

christopherlam:
This PR also modifies the "Jump to Invoice" functionality -- it was flawed when a payment spanned multiple invoices. Previously it would jump to the first invoice found. Now it jumps from a payment split to the corresponding invoice.

The disadvantage is that jump from bank register is now disabled; this would be difficult because multiple invoices would either require UI to choose invoices to open, or we'd need to open all invoices.

PR #602, opened 2019-11-12
Maint

christopherlam:
Some subjectively better changes.

Consider main table = owner-report, and invoice->payments and payment->invoices table = link-table. Changes in this series are:

1. main table 'payment' links to transfer account split **instead** of APAR split.
2. link table invoice->payments show **full** payment amount for multi-invoice payments instead of *partial* payment.
3. link table invoice->payments show **transfer account currency** instead of APAR currency whereby payments were made in another currency.

image


fellen:
More descriptive titles than "Maint" or "Scheme progress" for PR's would be nice. ;-)

PR #605, opened 2019-11-16
Add budget-simple.scm to standard reports

push-eax:
I'd like to share a budget report I've written, budget-simple.scm. I like the budgeting functionality in GnuCash, but the default budget reports have some shortcomings that make it difficult to quickly get the information I need. I wanted a report that could tell me, at a glance:
 - whether the accounts I had budgeted were within budget
 - whether (and how much) I was spending on accounts without a budget. Ideally, one wouldn't be spending very much (or anything) outside their budget, but sometimes this happens.

The default reports I used were budget.scm and budget-flow.scm. Here are my criticisms of both:

- Both reports display many accounts that have no budget, which makes them much less succinct and less efficient for humans to parse. In budget.scm, there is an option to "include accounts with zero total balances and budget values", but this includes accounts with balances that don't have budgets, which isn't really what I wanted.

Problems with budget.scm:
- Displays too many periods. This is very useful for determining budget trends over time, but it can get confusing if you want to look at budget performance for the current period only. It's pretty easy to display only one period, though.
- Displays too many accounts. There are too many irrelevant accounts onscreen, and that costs the user time in finding what they want. Even with the "include accounts with zero total balances and budget values" option deselected, too many accounts show up. A solution to this is to manually deselect accounts in the account selector, but that introduces another problem:
- Incorrect totals when accounts are removed from the account selector. For example, if I deselect Expenses:Taxes, the reported value of the Expenses top-level account doesn't change, even though the Expenses:Taxes subtree is no longer displayed in the report. Invisible accounts affecting the totals of visible ones makes me queasy, and more importantly, forces the user to manually subtract the values of the invisible accounts, which is what software is supposed to eliminate.

Problems with budget-flow.scm:
- Does not display parent accounts. Parent account values include the values of their children, and writing a report that deals with this properly is challenging. Whoever wrote this report chose not to deal with this at all by ignoring parent accounts entirely. This makes the account tree very ugly and confusing.

I've implemented budget-simple.scm to address the problems in both budget.scm and budget-flow.scm. It's made of some code repurposed from both, plus a new account tree traversal function written from scratch (accounts are recursively traversed, with the budgets and values of parent accounts trusted, i.e. recursion stops at budgeted parent accounts).

In writing this report, I made some assumptions that might become bugs in certain situations. Here are as many as I can think of:

- Budgets are not set on top-level accounts. I don't know why someone would do this, since budget values for (non-budgeted) parent accounts are shown in the budget editor as the sum of the budgets of their children. budget-simple.scm does not look at top-level accounts to avoid double-counting.
- Budgets are not set on child accounts of other budgeted accounts. I suppose that there MAY be a case for doing this if, for example, the user wants to set a budget for a group of accounts and then set a limit on a specific account within that group. From my tests, doing this does not change the budget of the parent. I've never set a budget for a child of a budgeted parent, and I don't know how I'd treat that from a reporting standpoint (e.g. if a child was overbudget, but the parent wasn't, are we overbudget since the parent budget also contains child spending?) so I chose not to implement it.

I should note that while I've been programming for a long time, I'm new to Scheme and had never touched it or any Lisp prior to August of this year when I wrote this. The code might not be in line with Scheme best practices or conventions as a result, but it works and I've documented it as best I can. I'm also not an accountant (and maybe accountants see no distinction between spending on overbudget accounts and spending on non-budgeted accounts, but this information is useful to me).

Ironically, I stopped using GnuCash almost immediately after I wrote budget-simple.scm for other reasons, mainly a lack of a web interface and poor support for importing brokerage transactions via OFX. However, I want to make my work available to the GnuCash project since GnuCash is really great software despite the fact that it no longer suits my use case.

Happy budgeting.

### DIFF
--- a/gnucash/report/standard-reports/budget-simple.scm
+++ b/gnucash/report/standard-reports/budget-simple.scm
@@ -1,0 +1,433 @@
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;; budget-simple.scm: simplified budget report
+;;
+;; Kiernan Roche <self@kiernanro.ch> 8/14/2019
+;;
+;; based on budget.scm and budget-flow.scm by:
+;; Chris Shoemaker <c.shoemaker@cox.net>
+;;
+;; This program is free software; you can redistribute it and/or
+;; modify it under the terms of the GNU General Public License as
+;; published by the Free Software Foundation; either version 2 of
+;; the License, or (at your option) any later version.
+;;
+;; This program is distributed in the hope that it will be useful,
+;; but WITHOUT ANY WARRANTY; without even the implied warranty of
+;; MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+;; GNU General Public License for more details.
+;;
+;; You should have received a copy of the GNU General Public License
+;; along with this program; if not, contact:
+;;
+;; Free Software Foundation           Voice:  +1-617-542-5942
+;; 51 Franklin Street, Fifth Floor    Fax:    +1-617-542-2652
+;; Boston, MA  02110-1301,  USA       gnu@gnu.org
+;;
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+
+(define-module (gnucash report standard-reports budget-simple))
+
+(use-modules (gnucash utilities)) 
+(use-modules (gnucash gnc-module))
+(use-modules (gnucash gettext))
+
+(use-modules (gnucash engine))
+
+(use-modules (srfi srfi-1))
+
+(gnc:module-load "gnucash/report/report-system" 0)
+(gnc:module-load "gnucash/gnome-utils" 0) ;for gnc-build-url
+
+(define reportname (N_ "Budget Simple"))
+
+;; Options definitions
+(define optname-accounts (N_ "Account"))
+(define optname-budget (N_ "Budget"))
+(define optname-report-currency (N_ "Report currency"))
+(define optname-price-source (N_ "Price source"))
+(define optname-use-current-period (N_ "Use current period"))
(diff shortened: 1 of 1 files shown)


christopherlam:
Hi @push-eax thank you for your contribution.

While your work is interesting, I would be inclined to *not* add another budget report to the current list. The reasons are multiple:

1. the budget module and reports were designed with some principles in mind; as incorrect/inadequate as they are, the existing transactional and budget data leads to people adopting practices to suit them. I'm afraid that an alternative calculation of budget vs actual amounts (especially with accounts traversal & with recursive subtotals) will lead to much confusion to new users, and if there's no one to document and explain (and defend) the alternate calculations, will only lead to developer burden.

2. many of the shortcomings could be fixed by simply augmenting/improving current reports. `budget.scm` will always be the canonical budget report; it has further improvement planned in maint and master branches, and will be fixed as soon as bugs are found; `budget-flow.scm` could be improved with suitable bug reports.

3. I can see bugs in this report and do not wish to fix them without understanding the assumptions in this code. e.g. the 'non-budget-account' simply tests for budget-value=0 but forgets the possibility of the account having budget in other periods. No explanation, no defense = poor fix.

4. if you're happy to leave GnuCash forever then no one will be using this report as intended.

So, but it will need much more attention, documentation and someone to explain to new and existing users how it is 'better' than existing code than I suspect you're willing to provide.

gjanssens:
I agree with @christopherlam here. If you were still a gnucash user I would have encouraged you to investigate how your work could be integrated in the existing budget reports. But as you had to swtich to another system, there's probably not much incentive to invest that time of course.

push-eax:
It's definitely buggy - I hacked it together quickly to see if I could get something working. If I were still a GnuCash user, I'd polish it up. Maybe the account traversal code will be of some use in the future, maybe not.

> as incorrect/inadequate as they are, the existing transactional and budget data leads to people adopting practices to suit them.

I disagree strongly with this attitude. Adopting practices to cope with incorrect or inadequate tooling instead of improving the tooling only creates long-term frustration, which is why I wrote this report in the first place (and ultimately why I stopped using GnuCash). However, I realize that my use case and requirements are not universal, so even with bugfixes, including the report in the project might not be appropriate.

Thanks for taking a look at it.

PR #606, opened 2019-11-18
Remove colon and trailing spaces

Bob-IT:
This PR removes the trailing colon from dialogue labels to make all labels of the same format, it also removes some trailing spaces that were added as a spacer which have been replaced with the use of padding and margins. The status bars have been left as is. 

I think I have identified all of them but if I come across them on master when I change some widgets to grids I will change them then.

This obviously will require updated translations and may be ultimately some reduction if the labels are consistent and it was expressed that it should be done on maint.

If no one has any concerns, I will push this on Friday and hopefully merge this to master once I worked out how to do that and then I can change some dialogues to use grids instead of boxes and also change the label alignment to the right.

### DIFF
--- a/gnucash/gnome-utils/dialog-account.c
+++ b/gnucash/gnome-utils/dialog-account.c
@@ -2153,7 +2153,7 @@ gnc_account_cascade_color_dialog (GtkWidget *window, Account *account)
                       G_CALLBACK(default_color_button_cb), (gpointer)color_button);
 
     string = g_strdup_printf(_( "Set the account color for account '%s' "
-                                "including all sub-accounts to the selected color:"),
+                                "including all sub-accounts to the selected color"),
                              gnc_account_get_full_name(account));
     gtk_label_set_text (GTK_LABEL(color_label), string);
     g_free (string);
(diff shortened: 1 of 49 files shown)


fellen:
Can you start the commit messages with "I18N: " or append " [I18N]?
@jralls Do you have a special preference for writing the release info?

jralls:
Maybe a paragraph in the merge commit explaining what they need to do after merging in the new `gnucash.pot`. What's going to happen to the msgids with whitespace or colons on the end and their corresponding msgstrs? Do they get marked "fuzzy"?

fellen:
> Maybe a paragraph in the merge commit explaining what they need to do after merging in the new `gnucash.pot`.

I know nothing which is not explained in
https://wiki.gnucash.org/wiki/Translation#Translating_the_Program

> What's going to happen to the msgids with whitespace or colons on the end and their corresponding msgstrs? Do they get marked "fuzzy"?

Yes.

fellen:
OK, it might be easier for them to search in a first run for ':"' and ': "', And for non-TP projects, we could do it.

PR #607, opened 2019-11-24
Bug 797506 - New Aging errors out with guile backtrace in case of a few uncommon transactions

christopherlam:
1. Don't error out when payment doesn't have owner. Log and skip it.
2. Render the skipped list in report. Links to offending split.

image

### DIFF
--- a/gnucash/report/business-reports/new-aging.scm
+++ b/gnucash/report/business-reports/new-aging.scm
@@ -213,6 +213,7 @@ exist but have no suitable transactions."))
          (show-zeros (op-value gnc:pagename-general optname-show-zeros))
          (date-type (op-value gnc:pagename-general optname-date-driver))
          (query (qof-query-create-for-splits))
+         (invalid-splits '())
          (document (gnc:make-html-document)))
 
     ;; for sorting and delete-duplicates. compare GUIDs
@@ -223,6 +224,10 @@ exist but have no suitable transactions."))
     (define (owner<? a b)
       ((if (eq? sort-order 'increasing) string<? string>?)
        (gncOwnerGetName a) (gncOwnerGetName b)))
+
+    (define (html-markup-ol lst)
+      (apply gnc:html-markup "ol"
+             (map (lambda (elt) (gnc:html-markup "li" elt)) lst)))
 
     ;; set default title
     (gnc:html-document-set-title! document report-title)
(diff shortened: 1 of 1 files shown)


gjanssens:
This fixes all crashes on my test book and nicely reports the odd splits. Thanks!

On the other hand seems quite noisy in the logs: it now fills with plenty of messages like:

> d0.0003 t12.664: ('next=invoice Split<d:01/07/2009,acc:Rekening Courrant Zaakvoerder,amt:€ 457,60-,val:€ 457,60-> Bill<Post:01/07/2009,Owner:Restaurantkosten,Notes:,Total:[€ 457,60]> 0)
> d0.0005 t12.664: ('payment Split<d:25/08/2009,acc:Rekening Courrant Zaakvoerder,amt:€ 457,60,val:€ 457,60> 2288/5 -> 0)
> * 17:41:08  WARN <gnc.scm     > split Split<d:22/10/2011,acc:Vorderingen klanten,amt:€ 20,00-,val:€ 20,00-> has no owner
> d117.8078 t130.472: ('processing: 24/11/2019 'duedate 'receivable? #t)
> d0.0007 t130.473: ('next=invoice Split<d:27/05/2005,acc:Vorderingen klanten,amt:€ 198,44,val:€ 198,44> Invoice<Post:27/05/2005,Owner:AbitMORE,Notes:,Total:[€ 198,44]> 0)
> d0.0005 t130.473: ('payment Split<d:30/06/2005,acc:Vorderingen klanten,amt:€ 198,44-,val:€ 198,44-> -4961/25 -> 0)

The warning makes sense, the rest would only make sense in info or debug mode. Should these be written with gnc:debug or gnc:message instead ?

christopherlam:
I *don't* quite understand the purpose/difference between info vs debug... which one is more appropriate? [is this it?](https://stackoverflow.com/questions/6860186/log-info-vs-log-debug)

jralls:
Sort of. For our purposes `info` messages are added to the trace file when the user launches GnuCash with the (perhaps misnamed) `--debug` command line option and `debug` messages for a particular log module when the user launches with `--log gnc.log.module=debug` ( a real example of a log module would be `gnc.backend.dbi`; I'll tell a user to do that when I need to see the SQL queries for a SQL backend problem; since the queries will leak information that most would consider confidential I'll also tell the user to send the trace file directly to me instead of to the list or attached to a bug report.

Another example: The engine has `ENTER` and `LEAVE` logging macros at the entry and exit points of many functions that log at the `info` level and so go into the tracefile when the `--debug` flag is used. That can be helpful when isolating a crash on Windows where generating a stack trace is hard.

PR #608, opened 2019-11-25
drop emacs_trailer from xml files.

fellen:
Modern versions of emacs know, how to behave on xml files.


fellen:
Some time ago, we removed the emacs trailers from our templates. Today I found the part, which created them.

Are there any objections?

Did I miss the read counterpart?

Other Annotations?

jralls:
Go right ahead. Editor customizations are the user's business, they don't belong in our repo nor should we have code that embeds them in data files.

And FWIW emacs is my primary editor.

PR #610, opened 2019-11-27
Drop gnc-module use from libgnucash

gjanssens:
Primary goal of this branch is to eliminate use of gnc-module from libgnucash. gnc-module is still exposed to consumers of libgnucash though.

This branch also starts the work of making libgnucash no longer depend on guile. This is complete for all but app-utils and tax. Work on this will continue at a later date. Guile wrappers have been moved to bindings/guile for consistency with the bindings/python code.

As part of this cleanup the guile wrapper for gnc-module has been deprecated. Guile access to the libgnucash functionality should be handled via guile's own (use-modules ...) interface. There are guile modules for each gnc-module that has been removed:
- gnucash/engine -> (gnucash engine)
- gnucash/app-utils -> (gnucash app-utils)
- gnucash/tax/us -> (gnucash locale tax us)
- gnucash/tax/de_DE -> (gnucash locale tax de_DE)

In addition, following gnc-modules were commonly loaded directly in guile code. While these gnc-modules can't be loaded from guile directly any more their functionality is exposed via equivalent guile modules:
- gnucash/html -> (gnucash html)
- gnucash/report/report-system -> (gnucash report)
- gnucash/gnome-utils -> (gnucash gnome-utils)


gjanssens:
Grmbl. Wrong branch :(

PR #611, opened 2019-11-27
Drop gnc-module use from libgnucash

gjanssens:
Primary goal of this branch is to eliminate use of gnc-module from libgnucash. gnc-module is still exposed to consumers of libgnucash though.

This branch also starts the work of making libgnucash no longer depend on guile. This is complete for all but app-utils and tax. Work on this will continue at a later date. Guile wrappers have been moved to bindings/guile for consistency with the bindings/python code.

As part of this cleanup the guile wrapper for gnc-module has been deprecated. Guile access to the libgnucash functionality should be handled via guile's own (use-modules ...) interface. There are guile modules for each gnc-module that has been removed:
- gnucash/engine -> (gnucash engine)
- gnucash/app-utils -> (gnucash app-utils)
- gnucash/tax/us -> (gnucash locale tax us)
- gnucash/tax/de_DE -> (gnucash locale tax de_DE)

In addition, following gnc-modules were commonly loaded directly in guile code. While these gnc-modules can't be loaded from guile directly any more their functionality is exposed via equivalent guile modules:
- gnucash/html -> (gnucash html)
- gnucash/report/report-system -> (gnucash report)
- gnucash/gnome-utils -> (gnucash gnome-utils)


gnucash/register/ledger-core/split-register-copy-ops-internals.c
```diff
@@ -0,0 +1,455 @@
...
+}
+
+/* type predicate */
+gboolean is_gnc_float_split (const floating_split *fs)
+{
+    if (fs)
+        return TRUE;
+    else
+        return FALSE;
+}
```
jralls: Eww. `return fs != NULL`.

gnucash/register/ledger-core/split-register-copy-ops-internals.h
```diff
@@ -0,0 +1,137 @@
...
+    Split       *m_split;
+    Account     *m_account;
+    Transaction *m_transaction;
+    const char  *m_memo;
+    const char  *m_action;
+    time64       m_reconcile_date;
+    char         m_reconcile_state;
+    gnc_numeric  m_value;
+    gnc_numeric  m_amount;
+} floating_split;
```
jralls: Typenames in camel-case, please.

bindings/guile/glib-guile.c
```diff
@@ -173,3 +204,139 @@ gnc_glist_string_p(SCM list)
...
+gnc_detach_process (Process *proc, const gboolean kill_it)
+{
+    g_return_if_fail (proc && proc->pid);
+
+    errno = 0;
+    close (proc->fd_stdin);
+    if (errno)
+    {
+        g_message ("Close of childs stdin (%d) failed: %s", proc->fd_stdin,
+                   g_strerror (errno));
```
jralls: s/childs/child's/.

gnucash/register/ledger-core/split-register-copy-ops-internals.c
gjanssens: Heh, indeed. I was in minimal transliteration mode apparently. Fixed.

gjanssens:
Op zaterdag 30 november 2019 01:54:13 CET schreef John Ralls:
> jralls commented on this pull request.
> 
> > +    if (*retptr == -1)
> 
> +        g_warning ("Pipe to childs file descriptor %d is -1", std_fd);
> +    return *retptr;
> +}
> +
> +void
> +gnc_detach_process (Process *proc, const gboolean kill_it)
> +{
> +    g_return_if_fail (proc && proc->pid);
> +
> +    errno = 0;
> +    close (proc->fd_stdin);
> +    if (errno)
> +    {
> +        g_message ("Close of childs stdin (%d) failed: %s", proc->fd_stdin,
> +                   g_strerror (errno));
> 
> s/childs/child's/.


Thanks for the feedback. Interestingly I can't find this review comment on 
github, so I don't know how to mark it as resolved...

It is now.

gjanssens:
> Sorry, I wasn't finished writing up the review last night. I guess I pressed the wrong button.
> 
> price-quotes.scm doesn't really belong in bindings, it's not wrapping anything. How about putting it in `gnucash` alongside `gnucash-bin.c`?
> 

True. This is a side-effect of my initial double goal of getting libgnucash guile-free as well as no longer using gnc-module. The guile cleanup turned out to be bigger than I could do in one sweep so I decided towards the end to focus on the gnc-module aspect first to get the branch into a mergable state.

What I really want for price-quotes.scm is to rewrite it in C(++) and move it back into the engine. I would like to keep it available for the bindings as well, not only the gui application. I had started this and then saw it depends on some gui code to display warning messages :(
Our guile code tends to pull tricks like that. That got me thinking of possibly implementing a more generic user messaging system with configurable output backends. The default one would print out messages on the console and would be sufficient for the bindings. The gui however could provide a backend that pops up a dialog with the user instead.
And then the project got bigger than I have currently time for so I left it at that for the time being. So until I could revisit all this, it can as well live alongside gnucash-bin.c indeed.

> Shouldn't the transaction copy/paste rewrite be in a separate PR? Regardless, the C implementation needs tests.

Fair enough. I'll add tests and see if it's reasonable to split it off into a separate PR.

bindings/guile/price-quotes.scm
```diff
@@ -28,6 +28,7 @@
 (use-modules (gnucash engine))
```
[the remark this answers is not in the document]
gjanssens: Right. Typo. Will be fixed in my next force-push.

gjanssens:
> price-quotes.scm doesn't really belong in bindings, it's not wrapping anything. How about putting it in `gnucash` alongside `gnucash-bin.c`?
> 
This is lives in `gnucash` until I rewrite it in C++.

> Shouldn't the transaction copy/paste rewrite be in a separate PR? Regardless, the C implementation needs tests.
I have moved this bit to the beginning of the commit history in this PR and added tests. Do you really want to split it off into a separate PR for separate review and merge or is this sufficient ?

gjanssens:
I did push the copy operations port to C as a separate branch in order to set it apart a bit more.

PR #612, opened 2019-11-29
Remove not needed semicolons in Python bindings

andrey-legayev:
Python syntax fix - remove not needed semicolons.


fellen:
Just a note:
https://www.python.org/dev/peps/pep-0008/ does not cover the case and we have still no python section in https://wiki.gnucash.org/wiki/CodingStandard.

About the commit: Yes, it looks more consistent.

christopherlam:
Agree semicolons are punctuation which is defitely unpythonic. Thanks!

andrey-legayev:
I reviewed what's doable with help of Python bindings and PyCharm shown me these places. 
I couldn't skip fixing it :)

PR #615, opened 2019-11-30
[Master] Change dialog-options to use a two column GtkGrid

Bob-IT:
Change the dialogue options to use a two column grid and setup to be the same as other dialogues. In the first column there is the 'name' of the option and the right column the option widget. There are two options that do not conform to this and they are the 'account-select' and 'list' options that have frames with the 'name' contained and so these two options use both columns.

This is a follow on from PR614 but was not sure if this would affect John's work so created separate PR, @jralls this can wait if this would interfere with your changes.

I have tested all the options and all looks good apart from radiobutton-option and currency-accounting-option-... which I can find being used.
@christopherlam Have I missed them, would it be possible to add a radiobutton example to Hello, World' and maybe fix the list option?

### DIFF
--- a/gnucash/gnome-utils/dialog-options.c
+++ b/gnucash/gnome-utils/dialog-options.c
@@ -1160,7 +1160,7 @@ gnc_option_create_radiobutton_widget(char *name, GNCOption *option)
     frame = gtk_frame_new (name);
 
     /* Create the button box */
-    box = gtk_box_new (GTK_ORIENTATION_HORIZONTAL, 5);
+    box = gtk_box_new (GTK_ORIENTATION_VERTICAL, 5);
     gtk_box_set_homogeneous (GTK_BOX (box), FALSE);
     gtk_container_add (GTK_CONTAINER (frame), box);
 
@@ -1528,6 +1528,9 @@ gnc_option_create_account_widget(GNCOption *option, char *name)
     g_signal_connect(G_OBJECT(button), "clicked",
                      G_CALLBACK(gnc_option_default_cb), option);
 
+    gtk_widget_set_margin_start (GTK_WIDGET(bbox), 6);
+    gtk_widget_set_margin_end (GTK_WIDGET(bbox), 6);
+
     if (multiple_selection)
     {
         /* Put the "Show hidden" checkbox on a separate line since the 4 buttons make
@@ -1640,7 +1643,7 @@ gnc_option_create_list_widget(GNCOption *option, char *name)
 
     bbox = gtk_button_box_new (GTK_ORIENTATION_VERTICAL);
     gtk_button_box_set_layout(GTK_BUTTON_BOX(bbox), GTK_BUTTONBOX_SPREAD);
-    gtk_box_pack_start(GTK_BOX(hbox), bbox, FALSE, FALSE, 10);
+    gtk_box_pack_end(GTK_BOX(hbox), bbox, FALSE, FALSE, 0);
 
     button = gtk_button_new_with_label(_("Select All"));
     gtk_box_pack_start(GTK_BOX(bbox), button, FALSE, FALSE, 0);
@@ -1663,6 +1666,8 @@ gnc_option_create_list_widget(GNCOption *option, char *name)
     g_signal_connect(G_OBJECT(button), "clicked",
                      G_CALLBACK(gnc_option_default_cb), option);
 
+    g_object_set (G_OBJECT(hbox), "margin", 3, NULL);
+
     gnc_option_set_widget (option, GTK_WIDGET(view));
 
     return frame;
@@ -1681,8 +1686,7 @@ gnc_option_font_changed_cb(GtkFontButton *font_button, GNCOption *option)
 }
 
 static void
-gnc_option_set_ui_widget(GNCOption *option,
-                         GtkBox *page_box)
+gnc_option_set_ui_widget(GNCOption *option, GtkGrid *page_box, gint grid_row)
 {
     GtkWidget *enclosing = NULL;
     GtkWidget *value = NULL;
@@ -1691,6 +1695,7 @@ gnc_option_set_ui_widget(GNCOption *option,
     char *name, *documentation;
     char *type;
     GNCOptionDef_t *option_def;
+    GtkLabel *name_label;
 
     ENTER("option %p(%s), box %p",
           option, gnc_option_name(option), page_box);
@@ -1718,11 +1723,13 @@ gnc_option_set_ui_widget(GNCOption *option,
     else
         documentation = NULL;
 
+    name_label = GTK_LABEL(gtk_label_new (name));
+
     option_def = gnc_options_ui_get_option (type);
     if (option_def && option_def->set_widget)
     {
         value = option_def->set_widget (option, page_box,
-                                        name, documentation,
+                                        name_label, documentation,
                                         /* Return values */
                                         &enclosing, &packed);
     }
@@ -1730,6 +1737,13 @@ gnc_option_set_ui_widget(GNCOption *option,
     {
         PERR("Unknown option type. Ignoring option \"%s\".\n", name);
     }
+
+    /* attach the name label to the first column of the grid and
+       align to the end */
+    gtk_grid_attach (GTK_GRID(page_box), GTK_WIDGET(name_label),
+                     0, grid_row, // left, top
+                     1, 1);  // width, height
+    gtk_widget_set_halign (GTK_WIDGET(name_label), GTK_ALIGN_END);
 
     if (!packed && (enclosing != NULL))
     {
@@ -1739,11 +1753,10 @@ gnc_option_set_ui_widget(GNCOption *option,
 
         gtk_container_add (GTK_CONTAINER (eventbox), enclosing);
 
-        /* Allow the text widget to expand and fill remaining space */
-        if (g_strcmp0 (type, "text") == 0)
-            gtk_box_pack_start (page_box, eventbox, TRUE, TRUE, 0);
-        else
-            gtk_box_pack_start (page_box, eventbox, FALSE, FALSE, 0);
+        /* attach the option widget to the second column of the grid */
+        gtk_grid_attach (GTK_GRID(page_box), eventbox,
+                         1, grid_row, // left, top
+                         1, 1);  // width, height
 
         gtk_widget_set_tooltip_text (eventbox, documentation);
     }
@@ -1761,9 +1774,10 @@ gnc_option_set_ui_widget(GNCOption *option,
 
 static void
 gnc_options_dialog_add_option(GtkWidget *page,
-                              GNCOption *option)
+                              GNCOption *option, gint row)
 {
-    gnc_option_set_ui_widget(option, GTK_BOX(page));
+    g_object_set_data (G_OBJECT(page), "options-grid-row", GINT_TO_POINTER(row));
+    gnc_option_set_ui_widget(option, GTK_GRID(page), row);
 }
 
 static gint
@@ -1801,6 +1815,7 @@ gnc_options_dialog_append_page(GNCOptionWin * propertybox,
 
     /* Build this options page */
     page_content_box = gtk_box_new (GTK_ORIENTATION_VERTICAL, 2);
+    gtk_widget_set_name (page_content_box, "page-content-box");
     gtk_box_set_homogeneous (GTK_BOX (page_content_box), FALSE);
 
     gtk_container_set_border_width(GTK_CONTAINER(page_content_box), 12);
@@ -1809,8 +1824,12 @@ gnc_options_dialog_append_page(GNCOptionWin * propertybox,
     gtk_box_pack_start(GTK_BOX(page_content_box), options_scrolled_win, TRUE, TRUE, 0);
 
     /* Build space for the content - the options box */
-    options_box = gtk_box_new (GTK_ORIENTATION_VERTICAL, 5);
-    gtk_box_set_homogeneous (GTK_BOX (options_box), FALSE);
+    options_box = gtk_grid_new(); // this will have two columns
+    gtk_widget_set_name (options_box, "options-box");
+    gtk_grid_set_row_homogeneous (GTK_GRID(options_box), FALSE);
+    gtk_grid_set_column_homogeneous (GTK_GRID(options_box), FALSE);
+    gtk_grid_set_row_spacing (GTK_GRID(options_box), 6);
+    gtk_grid_set_column_spacing (GTK_GRID(options_box), 6);
 
     gtk_container_set_border_width(GTK_CONTAINER(options_box), 0);
     gtk_container_add (GTK_CONTAINER(options_scrolled_win), GTK_WIDGET(options_box));
@@ -1821,7 +1840,7 @@ gnc_options_dialog_append_page(GNCOptionWin * propertybox,
     for (i = 0; i < num_options; i++)
     {
         option = gnc_get_option_section_option(section, i);
-        gnc_options_dialog_add_option(options_box, option);
+        gnc_options_dialog_add_option(options_box, option, i);
     }
 
     /* Add a button box at the bottom of the page */
@@ -2403,9 +2422,19 @@ gnc_options_dialog_destroy(GNCOptionWin * win)
  * The widget you return from this function should be the widget in
  * which you're storing the option value.
  */
+ 
+static void
+gnc_option_set_ui_label_alignment (GtkLabel *name_label)
+{
+    /* some option widgets have a large vertical foot print so align
+       the label name to the top and add a small top margin */
+    gtk_widget_set_valign (GTK_WIDGET(name_label), GTK_ALIGN_START);
+    gtk_widget_set_margin_top (GTK_WIDGET(name_label), 6); 
+}
+
 static GtkWidget *
-gnc_option_set_ui_widget_boolean (GNCOption *option, GtkBox *page_box,
-                                  char *name, char *documentation,
+gnc_option_set_ui_widget_boolean (GNCOption *option, GtkGrid *page_box,
+                                  GtkLabel *name_label, char *documentation,
                                   /* Return values */
                                   GtkWidget **enclosing, gboolean *packed)
 {
@@ -2413,7 +2442,7 @@ gnc_option_set_ui_widget_boolean (GNCOption *option, GtkBox *page_box,
 
     *enclosing = gtk_box_new (GTK_ORIENTATION_HORIZONTAL, 5);
     gtk_box_set_homogeneous (GTK_BOX (*enclosing), FALSE);
-    value = gtk_check_button_new_with_label(name);
+    value = gtk_check_button_new ();
 
     gnc_option_set_widget (option, value);
     gnc_option_set_ui_value(option, FALSE);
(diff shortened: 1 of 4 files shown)


jralls:
@Bob-IT looks fine to me. I'm about to start integrating the C++ options with dialog-options.c so better that you go ahead and merge this when you're ready and I'll use it for integration; that will make for a cleaner merge of C++ options.

christopherlam:
Not sure what you mean by fixing the list option.

```patch
modified   gnucash/report/utility-reports/hello-world.scm
@@ -207,6 +207,20 @@ Your reports probably shouldn't have an \
 option like this.") 
       #f))
     
+    (add-option
+     (gnc:make-radiobutton-option
+      (N_ "Hello Again") "A radiobutton option" "z" "helptext" 'good
+      (list (vector 'good
+                    (N_ "The Good")
+                    (N_ "Good option."))
+            (vector 'bad
+                    (N_ "The Bad")
+                    (N_ "Bad option."))
+            (vector 'ugly
+                    (N_ "The Ugly")
+                    (N_ "Ugly option.")))))
+
     (gnc:options-set-default-section options "Hello, World!")      
     options))
```

PS options look *good*. I'm not settled whether right-aligning option names is better than left-align, you make that call :) also check File>Properties>Business>Fancy date format isn't aligned with others.

jralls:
Maybe it would be better to remove the radio button option since it's not used anywhere.

christopherlam:
> Maybe it would be better to remove the radio button option since it's not used anywhere.

Agree; the functionality is identical to make-list-option and radio buttons take up more screen. I'll work on deprecation tonight and removal in master?

Bob-IT:
That's fine with me also, do you want me to remove it from dialog-options ?
Can you see where currency-account-options-... are used, maybe they are also a candidate for removal?

@christopherlam Thanks for the code snippet, it did allow me to make a few tweaks and consider changing the orientation. Regarding the fancy date alignment, took me a minute and then realised I had already fixed it in my other PR as it showed up on the print cheque dialogue.

gjanssens:
Please don't rush the removal of the radio button options. I am afk until Tuesday but want to discuss this a bit more first.

christopherlam:
> That's fine with me also, do you want me to remove it from dialog-options ?
> Can you see where currency-account-options-... are used, maybe they are also a candidate for removal?

Currency accounting options are aycinena's project, unfinished for at least 4 years. Perhaps a vote in devel would be opportune. They are untested, and incomprehensible.

> Please don't rush the removal of the radio button options. I am afk until Tuesday but want to discuss this a bit more first.

Ok.

Bob-IT:
@christopherlam  The fix I mentioned is shown below, I do not know why 'List items selected' is stacked like that when above you have 'Items you selected:', just did not look write to me.

hello

Just a thought on the radio buttons, if we keep them there is possibly some merit in being able to select orientation, I could probably do that by duplicating some of the existing radio button option but there is probably a way in scheme to add an 'orientation' value and then be able to retrieve it in C to set the GtkBox orientation.

jralls:
@Bob-IT remember that options are going to C++. A radiobutton option is just a display variant of a multichoice. The simplest approach would be to have [`GncOptionUIType`](https://github.com/jralls/gnucash/blob/c%2B%2Boptions/libgnucash/app-utils/gnc-option.hpp#L92) of `RADIO_HORIZ` and `RADIO_VERT`. instead of `RADIOBUTTON`.

Bob-IT:
@jralls Thank you for that, I think I just missed off the ++, will have a think about that and wait for what Geert has to say before changing any thing.

christopherlam:
> List items selected' is stacked like that when above you have 'Items you selected

Because the `hello-world.scm` report is not of major interest. Here's a small fix to make it less ugly. Don't think anything wrong in C code.

```patch
modified   gnucash/report/utility-reports/hello-world.scm
@@ -407,6 +407,9 @@ new, totally cool report, consult the mailing list ~a.")
           (let ((table (gnc:make-html-table)))
             (gnc:html-table-append-column! 
              table (map symbol->string list-val))
+            (gnc:html-table-set-style! table "table"
+             'attribute (list "style" "width:200px"))
             (gnc:html-table-set-caption! table 
                                          (_ "List items selected"))
             (gnc:html-document-add-object! document table))
```

Bob-IT:
I have made some tweaks to the original push which are all options have two columns, the left being the label and the right the option widget. Changed some spacing and alignment also. The radio button has been changed to vertical packing as it would work better with long text and I will look at adding an orientation variant after John has done his changes if required.

Have tried this on both Windows and Fedora and looks OK to me and in RTL format.

Forgot to change the commit message so will change that later and if there are no concerns I will push this later along with my other PR.

christopherlam:
Hi @Bob-IT -- a couple feedback;

1. in my Adwaita theme there's a bounding box as follows in some options -- is this fixable?
2. see screenshot; I feel we can homogenize the height by setting a `minimum_height` property equal to GtkGrid height? It would look slightly more tidy?
3. I've tested your work with 1 small change: `GTK_ALIGN_START` instead of `GTK_ALIGN_END`... personally I prefer left-align option name -- don't think what others think?

Left-align:
image

Right-align:
image

christopherlam:
>     3\. 've tested your work with 1 small change: `GTK_ALIGN_START` instead of `GTK_ALIGN_END`... personally I prefer left-align option name

Or leave it right aligned... which will probably better with long option names, and will make GnuCash's report options display rather unique, which is not a bad thing :) -- note other options eg Preferences are left-aligned.

Bob-IT:
@christopherlam  I added the frame around some options when there were more than one widget which sort tide in with the more complex options like 'account-selection' and 'lists' but if it is not a good idea I can remove, I just thought it looked better.

2, Not sure what you mean

3, Part of this is to align the text in all dialogues, in my other PR I am setting those to pack end also so the application is consistent, there were a mixture of start and end and according to [this](https://developer.gnome.org/hig/stable/visual-layout.html.en) seemed the better option. I left the preferences alone for now as a special case but may test changing it later.

christopherlam:
1. np
2. I meant the "Report Name" -> "Stylesheet" vertical spacing is larger than "Show original currency amount" -> "Table for Exporting" because the latter are small checkboxes... slightly irregular, but I think to make them all equally vertically spaced would be uglier.
3. np

Bob-IT:
OK, yes making all the option rows the same height is tricky, it could work on the General tab but it would not work in the accounts tab. The other option would be to add some top and bottom margin for tick boxes but then the display tab has loads of them so that would increase the page height and maybe scrolling.

gjanssens:
So... first of all, I'm glad you're taking up this work @Bob-IT 

On the subject of alignment I'm with you: I think right aligning the labels is better for usability as explained in the link you refer to.

Coming back to the radio button options, I think it was me that removed most of them a long time ago when converting from Gconf to GSettings. IIRC the storage format for a radio button option couldn't be easily converted to a GSettings format without needing migration code. That was a practical trade-off at the time.

However radio buttons have their value as a UI element regardless and maybe we should restore their use in some cases. There is plenty of literature on the web on the topic of radio buttons vs dropdown lists. One example (not terribly well written but it does bring up several good points) is [this one](https://blog.prototypr.io/7-rules-of-using-radio-buttons-vs-drop-down-menus-fddf50d312d1)

I'm just pointing out a few considerations:
* when the list of choices is long, a dropdown list is better
* for choices where none of the choices is a clear, recommended default, radio buttons are preferred as that allows the user to more easily compare the choices
* on the other hand for options where there's a clear, recommended default that's not expected to be changed in many cases, a dropdown list is fine
* things we easily recognize, like days of the week or months, dropdown lists work well as well

I don't think reviewing our existing options for radio button use cases should be done as part of this PR. But I think it would be a valuable UX improvement if we do that at some point. Until then I prefer to keep the radio button code around and up to date.

Bob-IT:
OK, To sum up it looks like it is OK to push this and I will have a look at changing the radio button option to have an orientation property as just for example on the Transaction report 'Display->Detail Level' and 'Display->Amount' would work as horizontal radio buttons while 'Accounts->Filter Type' would work as vertical radio buttons.

jralls:
OK. It actually makes no difference for the options backend except that I need a `gnc_register_radio_button_option` function to go with `gnc_register_multichoice_option`.

@Bob-IT the C++ implementation uses an enum to set the UI type, and I'd provide a `HORIZONTAL_RADIO` and `VERTICAL_RADIO` replacing the current `RADIO` to drive dispatch in `dialog-options.cpp`. That's not exposed to Scheme so the choices are to have either a `HORIZONTAL|VERTICAL`parameter on that `gnc_register_radio_button_option` function or to have two functions, `gnc_register_horizontal_radio_option` and `gnc_register_vertical_radio_option`. The latter is a bit simpler to implement.

fellen:
I found a missing page:
In Business->Billing Terms Editor 
PR6115_proximo
choose Type proximo
Also the common header part is not properly aligned.

Bob-IT:
@fellen took me a while to see the problem, the fix is pushed to master. Not sure why your image shows the common header part wrong as on my machine all looks OK.

fellen:
Oh, I had forgotten switching to grid waas only applied on Master. My screenshot was from maint.
Can we backport the changed labels as we did with the others?

christopherlam:
@Bob-IT may I ask one change: to toggle checkbox is now more difficult because the target box is small. How about enlarging the target to include the option name as well? 3.x was better in this respect.

image

Bob-IT:
@christopherlam that may be difficult as they are two separate widgets in a grid but try this in your CSS file...

#gnc-id-options check {
min-height: 20px;
min-width: 20px;
}

as you guessed, it changes the size of the square. What values look good for you, it could be added to the application CSS and still be overridden by users.

gjanssens:
@christopherlam brings up a valid usability issue I didn't consider earlier.
@Bob-IT we may have to move the checkbox labels to the right of the check box and as part of the check box themselves. The checkbox itself can remain on the same alignment as all the other actionable widgets like the combobox or gtkentry.

It may be slightly less clean but would greatly improve the usability. Unless you find a way to make the separate label in the first column clickable as well and make it toggle the check box too.

Bob-IT:
@christopherlam 
@gjanssens 
Your wish is my command, after thinking about it, it was not that difficult, added the label to an event box and callback to do the toggle.

gjanssens:
:D 

Unfortunately it's not a complete replacement of the way the check box' own label behaves. I see two potential issues:

1. Rapidly clicking on a check box will toggle on each mouse-click. Doing the same on the label in the event box will consume each second click for a double-click event. This is a minor yet annoying inconsistency
2. The check box' own label can have mnemonics so you can toggle it using <ALT>-<key>. I'm not sure the same can be done with a label in an event box.

There may be others we haven't even come across yet.

I think the core problem here is we want to use the GtkWidgets in ways they were not designed. While it certainly shows your skill in Gtk programming (which I enjoy), I'm not sure it is a good idea. What you're building is a GtkCheckBox of which you can freely place the label in another spot in the parent container, yet we want it to continue to behave as the GtkCheckBox. That takes extra code to achieve now and will take extra maintenance in the future each time Gtk changes their widget behaviour. I'm not sure that's a good use of our limited time.

Bob-IT:
@christopherlam 
@gjanssens 
As always you think of something that never occurred to me,
1) the double click may be fixable, would needed to ponder some more.
2) I am not sure, again would need to ponder, the only thing I would say is that they are not used currently.

It is a simple matter of using gtk_checkbox_new_with_label but not sure of the look..

Screenshot_2020-05-14_12-51-37
Screenshot_2020-05-14_12-52-28

So I am not sure what to do!!

christopherlam:
Instead of checkbox for boolean, is https://developer.gnome.org/hig/stable/figures/visual-alignment-correct.svg available for use?

gjanssens:
@Bob-IT thanks for adding the screenshots. That gives us something to talk about.

Another tweak you'd have had to do would have been disabling the label when the check box gets disabled (like the `Use Full Other Account Name` option in the screenshot above). In the changes as currently pushed to master the checkbox is disabled but you can still toggle it by clicking on the label. See what I mean we're quickly adding plenty of extra effort by not using widgets as they were designed ?

On the other hand I'll give you the result is not ideal. For the `General` tab the layout is not that bad. On the `Display` tab the few non-check boxes stand out this way.

However comparing this with how it was in 3.x, I think the result is still cleaner and easier on the eye so I would stick with it.

And taking a step back the issue may not be the location of the check box label itself but the general organisation of the options on the page. Looking at the `Display` tab with all the labels on the left is still not very nice. Reordering the options would help already, for example just from a first look:
* Group all options that show or hide columns together. 
* The `Detail Level` (Single line vs Multiline line mode) option is of a higher order in that it changes the whole format to display transactions in. It would make sense to put that option at the beginning of the tab, as it will affect the column selection options. (It's now right above the options it will affect, but in hindsight that's asking users to switch mindsets when scanning through the options from which columns to show to how to structure transactions and back).
* `Running balance` controls visibility of a column, so it should go with the other column options
* I have my doubts whether `Sign Reverses` should really be a report option, but if so it's again an option affecting the whole of the report so it could be grouped with the `Detail Level` option. By the way shouldn't this be "Sign Reversals" rather than "Sign Reverses" ?
* The same goes for `Subtotal Table`
* And finally I'm not sure where `Totals` should live, but as it's at the end of each account or at the end of the report, putting it at the end of the Display tab seems reasonable.

This rearrangement would help in several ways: check box options will be grouped more logically and there will be less intermediate indentation switches.

Another bit that would help tremendously would be the ability to visually group options together. I understand the option tabs are assembled from code and we currently provide only limited control over the layout other than display order and which tab to go in. And to make things worse we have tightly coupled option name with visual representation (the tab name is part of the option name). A future improvement (though highly unlikely for 4.x) could be to separate option name from layout. And then extend the layout parameters we support to include an option group within an option tab. When we have that we could for example have a group of check boxes in a GtkFrame with a descriptive title (like "Columns"). That's far away and hypothetical though. So for now we have to work with the tools we have.

gjanssens:
> Instead of checkbox for boolean, is https://developer.gnome.org/hig/stable/figures/visual-alignment-correct.svg available for use?

That is the GtkSwitch widget. For starters it doesn't have an integrated label, so that brings the burden of label synchronisation back. Next, reading the [HIG recommendations](https://developer.gnome.org/hig/stable/switches.html.en) for this widget it would depend on the context. For just selecting which columns to show the HIG seems to prefer [check boxes](https://developer.gnome.org/hig/stable/check-boxes.html.en).

gjanssens:
The HIG does suggest grouping check boxes using a descriptive group title as I mentioned in the other comment. We don't have code that supports this currently so we can't yet.

Bob-IT:
@christopherlam 
@gjanssens 

Just for visual info...
Screenshot_2020-05-14_14-29-43
Screenshot_2020-05-14_14-30-04

Bob-IT:
@gjanssens 
So for now, do we leave it or change it?
I think I have a fix for the double click and also for the prelight for when the mouse enters and I think another call back on the checkbox would sort the sensitivity of the label.
I do see what you mean about adding the extras but I think that would probably be it.

gjanssens:
> @christopherlam
> @gjanssens
> 
> Just for visual info, it does get rid of the size issue...
> Screenshot_2020-05-14_14-29-43
> Screenshot_2020-05-14_14-30-04

I prefer not to go this route. The on/off switch is not meant for simple things like selecting columns to show or hide.

gjanssens:
> @gjanssens
> So for now, do we leave it or change it?
> I think I have a fix for the double click and also for the prelight for when the mouse enters and I think another call back on the checkbox would sort the sensitivity of the label.
> I do see what you mean about adding the extras but I think that would probably be it.

If you ask my opinion I'd prefer to use a checkbox as it is intended; with its own embedded label. The added glue code to make the external label behave similarly to the embedded one is a maintenance burden for little gain.

christopherlam:
https://bugs.documentfoundation.org/show_bug.cgi?id=99104 they have similar issues. It would seem that the consensus is that checkboxes take up the whole horizontal space cf other widgets.

https://developer.gnome.org/hig-book/unstable/controls-check-boxes.html.en

gjanssens:
> https://bugs.documentfoundation.org/show_bug.cgi?id=99104 they have similar issues. It would seem that the consensus is that checkboxes take up the whole horizontal space cf other widgets.
> 
It depends. They would if there's no caption for the checkbox group. In case of a group caption they do indent. Depending on the HIG, they will be below the caption and slightly indented or they will start to the right of the caption.

> https://developer.gnome.org/hig-book/unstable/controls-check-boxes.html.en

That looks like an older version of the GNOME HIG. I think that was for GNOME 2. However the newer version is nearly identical.

As I mentioned before the main issue with our dialogs is we don't support group captions in our options code. Not something we can solve quickly. With that lack in mind we may test the tdf idea of having check boxes start in the same column as the labels to see how that would look.

Bob-IT:
@christopherlam 
@gjanssens 
If you mean this, the check button has been placed in the first grid column with a width of two columns and the other labels have been left aligned as well.

The rest of the controls are placed in the second grid column, they would all line up at the point of the longest label in the first grid column.

I am not totally convinced on this layout but if am willing to go with it.

Screenshot_2020-05-14_18-12-43
Screenshot_2020-05-14_18-12-58

gjanssens:
I agree it doesn't look great either.

In summary:
* I advocate using widgets as they were designed and what they were designed for. That means I think we should continue to use check boxes for the show/hide choices (rather than gtk switches) and that we should use the check box' built-in label.
* None of the alignments is ideal; IMO this is primarily due to the limited subset of design patterns we currently support in our option code (remember option definition defines layout ATM). Extra design patterns such as an option group, a group caption or a separator would allow for a much nicer view.
* If I stick with my first summary item these are the remaining possibilities so far:
  1. All options align on the second column, labels (except those for check boxes) are in the first column and left aligned
  2. All options align on the second column, labels (except those for check boxes) are in the first column and right aligned
  3. All options except the check boxes align on the second column, checkboxes are left aligned in the first column, labels (except those for check boxes) are in the first column and left aligned
  4. All options except the check boxes align on the second column, checkboxes are left aligned in the first column, labels (except those for check boxes) are in the first column and left aligned
* Personally of these 4 choices I prefer number 2: all option widgets in the second column with labels right-aligned in the first (and check boxes use their internal label).
* If we want further improvements we can
  1. Go through the option tabs and evaluate where it would help to reorder the options.
  2. Work on code support for extra design patterns. This is a case of fixing problems at the right level.

Bob-IT:
That's fine, ii) it is, I will push the change later and as you say think about extra layout options at some other time maybe after John has pushed his changes.

christopherlam:
And the layout is easy to change in `trep-engine.scm`

```patch
index c317941b4..8e5d25f77 100644
@@ -931,8 +931,8 @@ be excluded from periodic reporting.")
      ;; One list per option here with: option-name, sort-tag,
      ;; help-string, default-value
      (list
-      (list (N_ "Date")                         "a"  (_ "Display the date?") #t)
-      (list (N_ "Reconciled Date")              "a2" (_ "Display the reconciled date?") #f)
+      (list (N_ "Date")                         "a5" (_ "Display the date?") #t)
+      (list (N_ "Reconciled Date")              "a6" (_ "Display the reconciled date?") #f)
       (if BOOK-SPLIT-ACTION
           (list (N_ "Num/Action")               "b"  (_ "Display the check number?") #t)
           (list (N_ "Num")                      "b"  (_ "Display the check number?") #t))
@@ -991,7 +991,7 @@ be excluded from periodic reporting.")
     (gnc:register-trep-option
      (gnc:make-multichoice-callback-option
       gnc:pagename-display optname-detail-level
-      "h" (_ "Amount of detail to display per transaction.")
+      "a1" (_ "Amount of detail to display per transaction.")
       'single
       (list (vector 'multi-line
                     (_ "Multi-Line")
@@ -1007,7 +1007,7 @@ be excluded from periodic reporting.")
     (gnc:register-trep-option
      (gnc:make-multichoice-callback-option
       gnc:pagename-display (N_ "Amount")
-      "m" (_ "Display the amount?")
+      "a2" (_ "Display the amount?")
       amount-value
       (list
        (vector 'none   (_ "None") (_ "No amount display."))
@@ -1026,7 +1026,7 @@ be excluded from periodic reporting.")
     (gnc:register-trep-option
      (gnc:make-multichoice-option
       gnc:pagename-display (N_ "Sign Reverses")
-      "m1" (_ "Reverse amount display for certain account types.")
+      "a3" (_ "Reverse amount display for certain account types.")
       'global
       (keylist->vectorlist sign-reverse-list))))
```

image

gjanssens:
> That's fine, ii) it is, I will push the change later and as you say think about extra layout options at some other time maybe after John has pushed his changes.

Yes, I agree it makes more sense to add these after @jralls option work is merged.

gjanssens:
> And the layout is easy to change in `trep-engine.scm`

I know, which is partly why I brought it up :)
It's obviously limited to reordering within a tab, but it's a start. Similar small improvements can be evaluated for most other reports as well.

Note I'm not sure about the amount combo box. It's a three way variation of show/hide: show none, show in a single column or show in two columns. From that perspective it may have to stay with the column check boxes.
As Enable links only affects the presentation of amounts it should probably stick close to the amount combo box.
And subtotal table can go to the top of the check box list. Enabling it adds another section to the top of the report, not an extra column. @christopherlam I have noticed this option is disabled if Account view is set to two columns. Why is that again ?

christopherlam:
> And subtotal table can go to the top of the check box list. Enabling it adds another section to the top of the report, not an extra column. @christopherlam I have noticed this option is disabled if Account view is set to two columns. Why is that again ?

As an implementation detail, the subtotal table accumulates the subtotals encountered into a list called `grid` (as in 'sparse grid'). Showing dual Dr/Cr columns means there are two subtotals being accumulated per row. I did not know 100% whether Dr/Cr amount would be safely stored into the grid, so, I enabled it only for single-amount columns.

christopherlam:
> @christopherlam I have noticed this option is disabled if Account view is set to two columns. Why is that again

> As an implementation detail, the subtotal table accumulates the subtotals encountered into a list called `grid` (as in 'sparse grid'). Showing dual Dr/Cr columns means there are two subtotals being accumulated per row. I did not know 100% whether Dr/Cr amount would be safely stored into the grid, so, I enabled it only for single-amount columns.


For posterity... fixed by https://github.com/Gnucash/gnucash/commit/641985746226d16f63d56169af96effdd231715f

PR #617, opened 2019-12-06
[Master] Bug 797472 - Add option to choose account templates

Bob-IT:
In the account hierarchy assistant the choice of account template was
decided by the your locale and so could be limited. Add two combos so
you can select account templates by language and region.
@fellen Frank does this do what you want?

### DIFF
--- a/gnucash/gnome/assistant-hierarchy.c
+++ b/gnucash/gnome/assistant-hierarchy.c
@@ -94,6 +94,8 @@ typedef struct
 
     GtkWidget   *language_combo;
     GtkWidget   *region_combo;
+    GtkWidget   *region_label;
+
     const gchar *gnc_accounts_dir;
 
     GtkTreeView *categories_tree;
@@ -315,6 +317,7 @@ region_combo_changed_cb (GtkComboBox *widget, hierarchy_data  *data)
     if (gtk_combo_box_get_active_iter (widget, &filter_iter))
     {
         GtkListStore *cat_list = GTK_LIST_STORE(gtk_tree_view_get_model (data->categories_tree));
+        GtkTreeModel *cat_model = gtk_tree_view_get_model (data->categories_tree);
         GtkTreeSelection *selection = gtk_tree_view_get_selection (GTK_TREE_VIEW(data->categories_tree));
         GSList *list;
         GtkTreePath *path;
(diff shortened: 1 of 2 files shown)


jannick0:
+1 idea!

fellen:
...one giant leap for mankind.

Can we have (here or in the intro page) an invitation like "If not satisfied with the available templates read https://wiki.gnucash.org/wiki/Account_Hierarchy_Template and share your new or improved template.", - ideally with the URL as a parameter.

Minor issue:
For non-programmers "C" might be confusing. Can we replace it by "en_US" in the display mechanic?

Bob-IT:
@fellen I have made the changes so 'C' comes up as en_US, also added some text and a link label. Please give it a spin and report back.

jralls:
@fellen we want new features like that in 4.0 rather than 3.x.

gjanssens:
I'm quite happy with this step forward!

I have a couple of remarks.

First off, my testing is with an existing book, on the Accounts tab and then selecting `Actions`->`New Account Hierarchy...`

* On my system switching region and/or language can takes a long time. For example, switching from en_US to de_DE took 71 seconds. Going from there to de_CH took another 12 seconds. That strikes me as an abnormally long time to create the lists ?

* To reduce clutter I propose a few layout changes as can be seen in [this image](https://i.imgur.com/MuaKj0V.pngA):
1.  the message concerning creating your own hierarchies together with the note the selected hierarchies are only supplementary information, not key to using this page. So it would IMO be better to add this info at the bottom of the page as a bullet list of notes rather than at the beginning.
2. If you move the two dropdown items next to `Categories` they are probably self-explanatory and don't need separate `Language` and `Region` labels. You can mention language and region in the first sentence still if you like by changing it to "Select **language and region specific** categories that correspond to the ways..."

* When there's only one region for a selected language you preselect it by default, which I like very much. I'd go a teeny bit further and change the region dropdown to a label in that case.

* Same if there is no region for the given language. I like your choice of "--" in that case, but change it to a label.

* In case there are multiple regions for one language, select one already. While we can't know which one is the best, it may save clicks if the first selection happens to be the correct one. For this to be usable though it's important selecting a different language/region is **fast**.

* Do we want to support selecting categories from multiple languages/regions in one go ? Imagine this use case: a Swiss user wants to set up a hierarchy that's a mix of the de_DE and de_CH categories because not all de_DE categories are available for de_CH but the de_DE ones are close enough to be used anyway. Currently switching from one region to the other undoes all selections.

Bob-IT:
@gjanssens Interesting observations, I am not sure I have ever noticed the action->'New account hierarchy' way, my testing was always from File->'New file'
I have made the gui changes which I thought were an improvement.
Note sure about the last bullet point, that would probably require some list of selected categories to be visible and all that goes behind that, they can easily run the hierarchy assistant again, maybe for later.

What I am confused about is the delay which seems to stem from gnc_load_example_account_list.
For me, using Actions.... that function takes about 0.5 of a second to load en_GB at the start but if I then choose en_US and then back to en_GB it takes about 3 times longer to load. In fact they all take approximately 3 times longer to load, starting with LANG=de_DE gnucash, defaults to en_US which takes 0.5 at start and then takes 2 seconds.

Another strange thing is if I use 'File->New File' the initial load time is quicker and the same for subsequent changes.

More thought required...

Bob-IT:
I think I need to use gnc_suspend_gui_refresh and resume around gnc_load_example_account_list, more testing...

gjanssens:
Yes, that's a likely candidate.

Bob-IT:
OK my git foo needs work..
Looks like the glade changes were committed to master by mistake, not sure how.
Any way I have rebased on master and I think made the changes to my updated files so hopefully when I push this will be OK.
@gjanssens To fix the delay, using gnc_suspend_gui_refresh and resume around gnc_load_example_account_list did improve the speed but the real improvement was to wrap it with qof_event_suspend and resume. It looks like the existing account model was getting events for all the new accounts, comparing with its book and deciding it was not for it and returning so a lot of events.
Do not think I have used qof_event_suspend/resume before but all looks to work OK unless I have missed some thing...

PR #618, opened 2019-12-08
Fixes Bug797421

gsegovia2018:


gsegovia2018:
Tried to clean pull request up but won't allow me to push changes while "git rebase -i HEAD~n"

jralls:
There's a lot of garbage here that doesn't belong because you cloned somebody else's clone instead of our register directly.

That aside, no, changing the SCU on XXX is not a fix for Bug 797421.

PR #620, opened 2019-12-10
POC SplitListNode for gnc:account-accumulate-at-dates

christopherlam:
also upgrades `gnc:account-get-balances-at-dates`. These are used by `category-barchart`, `net-charts`, experimental multicolumn reports

note sure if this speeds it up significantly. after all the GList traversal takes place only *once* per account per report. times in seconds. the best test would involve an account north of 10,000 splits. from guile, `,time (define x (iota 10000000))` to create 10E7 list runs in 0.22s.

old GList: 0.612, 0.553, 0.50 0.50 0.523
new SplitListNode: 0.58 0.55 0.50 0.51 0.52

### DIFF
--- a/gnucash/report/report-system/report-utilities.scm
+++ b/gnucash/report/report-system/report-utilities.scm
@@ -498,7 +498,7 @@ flawed. see report-utilities.scm. please update reports.")
           (nosplit->elt #f)
           (split->date (compose xaccTransGetDate xaccSplitGetParent))
           (split->elt xaccSplitGetBalance))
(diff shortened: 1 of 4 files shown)


jralls:
That's why you should profile before spending time optimizing. ;-)

christopherlam:
After all the noise and interruption, I'll return to optimizing for extreme readability and test coverage then :man_shrugging: ... we did learn a *lot* in the process though.

I suspect Guile 2.2's VM optimiser is also much more clever when processing proper lists.

PR #621, opened 2019-12-27
[stylesheet-css] a user-customisable CSS based stylesheet

christopherlam:
New stylesheet. Exposes a single text box for full CSS customisability.

image


gjanssens:
Good enough for for an initial implementation as far as I'm concerned. If we get feedback we can start to fine-tune.

christopherlam:
merged in 1 commit. the search string is now `</style` because `</style etc>` also closes the tag.

PR #623, opened 2019-12-30
create a more consistant online_id

d-dittmann:
The online_id will generated based on bankcode and accountnumber. The
accountnumber may have leading zeros which shall be ignored to make the
creation reliable. In my case the accountnumber of the accountinfo has no
leading zeros, but the accountnumber of the imported transactions has leading
zeros. In the result the generated online_id was different and the widget
to assign the transactions to a account appears always.

### DIFF
--- a/gnucash/import-export/aqb/gnc-ab-utils.c
+++ b/gnucash/import-export/aqb/gnc-ab-utils.c
@@ -63,6 +63,7 @@ G_GNUC_UNUSED static QofLogModule log_module = G_LOG_DOMAIN;
 static AB_BANKING *gnc_AB_BANKING = NULL;
 static gint gnc_AB_BANKING_refcount = 0;
 
+static gchar* create_online_id(const gchar *bankcode, const gchar *accountnumber);
 static gpointer join_ab_strings_cb(const gchar *str, gpointer user_data);
 static Account *gnc_ab_accinfo_to_gnc_acc(GtkWidget *parent,
     AB_IMEXPORTER_ACCOUNTINFO *account_info);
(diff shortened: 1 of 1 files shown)


gnucash/import-export/aqb/gnc-ab-utils.c
```diff
@@ -316,6 +317,28 @@ gnc_AB_VALUE_to_readable_string(const AB_VALUE *value)
...
+static gchar*
+create_online_id(const gchar *bankcode, const gchar *accountnumber)
+{
+    gchar *online_id;
+
+    /* The accountnumber may have leading zeros, depending on where them
+     * accountnumver is came from, e.g. the accountnumber of accountinfo
+     * has no leading zeros while the (local)accountnumber of a transaction
+     * has leading zeros.
+     * So remove all leading '0', to get a consistant online_id.
```
jsoref:
```suggestion
     * So remove all leading '0', to get a consistent online_id.
```

gnucash/import-export/aqb/gnc-ab-utils.c
```diff
@@ -316,6 +317,28 @@ gnc_AB_VALUE_to_readable_string(const AB_VALUE *value)
...
 }
 
+
+static gchar*
+create_online_id(const gchar *bankcode, const gchar *accountnumber)
+{
+    gchar *online_id;
+
+    /* The accountnumber may have leading zeros, depending on where them
+     * accountnumver is came from, e.g. the accountnumber of accountinfo
```
fellen: * accountnumber ...

fellen:
> The change probably solves a bunch of unexpected behaviours for various user, and is unobtrusive enough to just go in. Thanks!

Yes, it is described in https://bugs.gnucash.org/show_bug.cgi?id=797432#c15

fellen:
I fixed the mentioned typos and applied it then.
Thank you!

PR #627, opened 2020-01-01
po/de.po: fix translation bug

jannick0:


fellen:
Never use githubs webinterface for merging for any gnucash project!
Because code.gnucash.org is canonical, the merge got overwritten by the next push.

PR #629, opened 2020-01-02
reports: bug fixes in translation strings

jannick0:
Please check if - and amend if needed - the syntax of menu paths (I used `>`) is the correct one for the project and/or the German locale.

Thanks.


jralls:
Thanks.

PR #631, opened 2020-01-09
Fix online account matching code to support subaccounts and adding accounts.

jralls:
@mtalexander complained in [gnucash-devel](https://lists.gnucash.org/pipermail/gnucash-devel/2020-January/044473.html) that 7853f5a24a4ed6e5cf32fdf6e5eba217a1e312cd had broken two features of OFX file imports, subaccounts and the ability to detect new accounts and ask if they should be created. This PR addresses and I think corrects both issues.


gnucash/import-export/import-account-matcher.c
```diff
@@ -101,29 +101,49 @@ static AccountPickerDialog* gnc_import_new_account_picker(void)
...
+        }
+        else
+        {
+            const char *partial_online_id =
+                gnc_import_get_acc_online_id(match->partial_match);
+            int partial_len = strlen(partial_online_id);
+            if (partial_online_id[partial_len - 1] == ' ')
+                --partial_len;
+            /* Both partial_online_id and acct_online_id are substrings of
+             * match->partial_id, but whichever is longer is the better match.
```
mtalexander: I think this comment should refer to match->online_id

gnucash/import-export/import-account-matcher.c
```diff
@@ -101,29 +101,49 @@ static AccountPickerDialog* gnc_import_new_account_picker(void)
...
+             */
+            if (partial_len < acct_len)
+                match->partial_match = acct;
```
mtalexander: You should reset match->count to 1 here since this is a new partial match.

mtalexander:
I tried this with my TIAA OFX file and it seemed to do the right thing.  My two comments are both trivial changes.

gnucash/import-export/import-account-matcher.c
jralls: Right.

gnucash/import-export/import-account-matcher.c
jralls: It is, but we're discarding the old one because we found a more complete partial match and so there's no ambiguity. See the test at gtest-import-account-matcher.cpp line 154: "Broker", "BrokerStocks", and "BrokerStocksHPE" are partial matches of "BrokerStocksHPEUSD but "BrokerStocksHPE" is unambiguously the best match so we want `count == 1` so that it will be the selected account.

I'll add an explanatory comment.

mtalexander:
.What happens if you get two matches for the same short substring and then
a match for a longer substring? Won't that leave the count set too 2
causing the longer substring too be rejected? I'm doing this from memory so
I may be wrong.

jralls:
I thought of that later this morning. I've added reseting `match->count = 1` in that case and logging an error if there's ambiguity because that would mean two accounts with the same online-id.

fellen:
@cstim can you test that it does not break the FinTS side?

gnucash/import-export/import-account-matcher.c
pkzw: Just a cosmetic remark:
The pattern 
```c
n = strlen(s);
if (s[n-1] == ' ')
    --n;
```
occurs 3 times within this code. It might be replaced by a function:
```c
size_t effstrlen(const char *s)
{
    size_t len = strlen(s);
    while(len > 0 && s[len - 1] == ' ')
        --len;
    return len;
}
...
n = effstrlen(s);
```
This example would be prepared also for more than one trailing blank.

PR #632, opened 2020-01-10
Add missing dependency on the math library

AdamMajer:
while building gnucash,

```
/usr/lib64/gcc/x86_64-suse-linux/9/../../../../x86_64-suse-linux/bin/ld: libgnucash/app-utils/CMakeFiles/gncmod-app-utils.dir/calculation/fin.c.o: in function `eff_int':
/home/abuild/rpmbuild/BUILD/gnucash-3.7/build/../libgnucash/app-utils/calculation/fin.c:1505: undefined reference to `pow'
/usr/lib64/gcc/x86_64-suse-linux/9/../../../../x86_64-suse-linux/bin/ld: /home/abuild/rpmbuild/BUILD/gnucash-3.7/build/../libgnucash/app-utils/calculation/fin.c:1509: undefined reference to `exp'
/usr/lib64/gcc/x86_64-suse-linux/9/../../../../x86_64-suse-linux/bin/ld: libgnucash/app-utils/CMakeFiles/gncmod-app-utils.dir/calculation/fin.c.o: in function `_fi_calc_num_payments':
/home/abuild/rpmbuild/BUILD/gnucash-3.7/build/../libgnucash/app-utils/calculation/fin.c:1289: undefined reference to `log'
```

### DIFF
--- a/libgnucash/app-utils/CMakeLists.txt
+++ b/libgnucash/app-utils/CMakeLists.txt
@@ -90,6 +90,29 @@ endif()
 
 
 add_library (gncmod-app-utils ${app_utils_ALL_SOURCES} ${SWIG_APP_UTILS_GUILE_C})
+
+include(CheckFunctionExists)
+
+if(HAVE_LIBM AND NOT POW_FUNCTION_EXISTS AND NOT NEED_LINKING_AGAINST_LIBM)
+  SET(LIBM_CFLAGS_STORAGE ${CMAKE_C_FLAGS})
+  SET(CMAKE_C_FLAGS "")
+  CHECK_FUNCTION_EXISTS(pow POW_FUNCTION_EXISTS)
+  if(NOT POW_FUNCTION_EXISTS)
+    list(APPEND CMAKE_REQUIRED_LIBRARIES m)
+    CHECK_FUNCTION_EXISTS(pow POW_FUNCTION_EXISTS_IN_M)
+    if(POW_FUNCTION_EXISTS_IN_M)
+      set(NEED_LINKING_AGAINST_LIBM True CACHE BOOL "" FORCE)
+    else()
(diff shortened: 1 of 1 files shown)


AdamMajer:
This could also be done in `${app_utils_ALL_LIBRARIES}` or some other place.

jralls:
OSes that don't have libm (which is everything that GnuCash overtly supports) need to be protected from this change. See [this stackoverflow](https://stackoverflow.com/questions/32816646/can-cmake-detect-if-i-need-to-link-to-libm-when-using-pow-in-c) for details.

AdamMajer:
Unfortunately using this fix hits another problem with cmake https://cmake.org/Bug/view.php?id=8246 . It almost seems like the best solution is to have a local fix instead of trying to do this in a general fashion.

jralls:
That's easy to work around: stash CMAKE_C_FLAGS, clear it, run the test, and then restore it.

AdamMajer:
@jralls Indeed. :+1:

This seems to now work and compile. But it looks a little bit hackey to me. What do you think about moving the logic about -lm detection to the main CMakeLists.txt?

jralls:
Although it doesn't really matter where it is, I think it's more idiomatic  to put the check in the top level `CMakeLists.txt`, or better yet make it a function, maybe `gnc_needs_libm` and put it in a module (they go in `common/cmake_modules`) and call it from `libgnucash/app-utils/CMakeLists.txt`.

jralls:
@AdamMajer Do you intend to move the check?

jralls:
I've pushed an equivalent test to master, but I can't be sure that it actually works because GnuCash builds just fine on OpenSuSE Tumbleweed without it (as long as you don't let it upgrade Guile to 3.0, but that's a separate issue).

PR #636, opened 2020-01-18
aging speedup

christopherlam:
Speeds up new-aging.scm

There were several splits loops that are reduced: (1) to find valid/invalid owners (2) to find unique owners (3) process each owner sequentially. Changed to (1) find next split's owner; if invalid then skip; process owner splits.

Will request @gjanssens to test with unusual APAR transactions.

### DIFF
--- a/gnucash/report/business-reports/new-aging.scm
+++ b/gnucash/report/business-reports/new-aging.scm
@@ -193,7 +193,7 @@ exist but have no suitable transactions."))
 ;; simpler version of gnc:owner-from-split. must be gncOwnerFree after
 ;; use! see split-has-owner? above...
 (define (split->owner split)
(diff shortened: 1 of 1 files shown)


christopherlam:
Ok a complicated book reduced report from 5s to 1.5s

gjanssens:
My PC is too fast/my dataset too slow to notice any difference with my typical test data set. I have also run this with a huge book I received from another user a couple of years back. The difference there is 6 minutes 30 seconds vs 4 minutes 40 seconds. Not as an impressive improvement ratio as you get but still in absolute terms almost 2 minutes gained. So yay!

Aslo the resulting report yields the same results before and after the patch, which is good. Unfortunately both results have an error which I'll report in a separate bug report.

christopherlam:
Thanks! I may try further speedups later on.

PR #637, opened 2020-01-20
Add Invoice.Unpost() to Python Bindings

loftx:
gncInvoiceUnpost isn't exposed as part of the Invoice object in the Python bindings so this minor commit added it for easier unposting from Python.


gjanssens:
Looks good. Thanks!

PR #638, opened 2020-01-22
[gnc-lot.c] speed up gncInvoiceGetInvoiceFromLot by caching invoice

christopherlam:
I'd hope this does not add debt. `cached_invoice` only used/set in gncInvoice API

### DIFF
--- a/libgnucash/engine/gnc-lot.c
+++ b/libgnucash/engine/gnc-lot.c
@@ -52,6 +52,7 @@
 #include "cap-gains.h"
 #include "Transaction.h"
 #include "TransactionP.h"
+#include "gncInvoice.h"
 
 /* This static indicates the debugging module that this .o belongs to.  */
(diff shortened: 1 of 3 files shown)


christopherlam:
Ideally I'd add `cached_owner` to use in `gncOwnerGetOwnerFromLot` too but is more annoying to do.

jralls:
Seems mostly harmless, but does it really provide a significant speedup? I wouldn't think that the lot->invoice lookup would be exercised that often.

christopherlam:
> Seems mostly harmless, but does it really provide a significant speedup? I wouldn't think that the lot->invoice lookup would be exercised that often.

It does get called many times on new-aging and new-owner-report. Running these reports on mta's book with dozens of customers and vendors is very slow otherwise.

christopherlam:
There's `owner_from_lot` but is not exportable.

christopherlam:
Merged in `master`

PR #639, opened 2020-01-29
Add a warning for gettext < 0.20

fellen:
They do not recognize <developer_name>GnuCash Project</developer_name>
in gnucash.appdata

https://savannah.gnu.org/bugs/?50408

I put it at first here, because formally it is changing a dependency.
OTOH, we lost already translations by this issue and it can be overwritten with cmake -DALLOW_OLD_GETTEXT ...


fellen:
OK, but then we to review the build scripts off our bundles and travis for ubuntu, if they need to set `cmake -DALLOW_OLD_GETTEXT` or its equivalent.

gjanssens:
> OK, but then we to review the build scripts off our bundles and travis for ubuntu, if they need to set `cmake -DALLOW_OLD_GETTEXT` or its equivalent.

Done.

- Our travis-ci ubuntu docker image has a very old version of gettext and we already added the -DALLOW_GETTEXT_OLD_GETTEXT  for that
- Our travis-ci arch docker image has gettext 0.20.1
- Our Windows build uses 0.19.8.1, I have added the parameter above.
- I don't know about the MacOS build. I trust @jralls will add the flag if required.

gjanssens:
Forgot to mention: flatpak build uses 0.20.1.

fellen:
Merged after rewordening: mention bug reference

PR #640, opened 2020-01-30
Bug 797570 Import Customers/Vendors, Bills/Invoices: list view for not imported rows does not expand with window size

ekinonnakapito:
…imported rows does not expand with window size

In the import of Customer/Vendors en Bills/Invoice, the dialog that lists rows that were not imported shows a scrollable view of only four rows.
The view does not use the available space of the window, and does not expand with resizing the window. This change fixes this issue.
In the same functionalities, the dialog for a user defined regular expression has a similar issue. This changes also fixes that issue.

### DIFF
--- a/gnucash/import-export/bi-import/dialog-bi-import-gui.c
+++ b/gnucash/import-export/bi-import/dialog-bi-import-gui.c
@@ -425,14 +425,14 @@ gnc_input_dialog (GtkWidget *parent, const gchar *title, const gchar *msg, const
 
     // add a label
     label = gtk_label_new (msg);
-    gtk_container_add (GTK_CONTAINER (content_area), label);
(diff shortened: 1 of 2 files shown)


ekinonnakapito:
I was not aware of that setting. I have just a basic understanding of the GtkWidget stuff. I created a small GTK application to reproduce the problem, applied an example from the web to fix it, and applied the same to the GnuCash code.
I think your proposal would still require 4 additions (not 2), as the issue is fixed in 4 instances. And it's a few extra lines of code, if I understand correctly, as it is an addition, not a replacement. Whether it is simpler, I cannot judge.

gjanssens:
Actually both methods would probably work. I'll stick with the one of this PR as the inner child of a dialog is effectively a GtkBox so using the GtkBox features makes for clearer code. Changing the call to add the label is perhaps not really necessary, but it's more consistent.

Thanks!

PR #641, opened 2020-02-02
Bug 797587  - Token handling for bayesian import matching is incorrect 

gruberchr:
This PR fixes bug [797587](https://bugs.gnucash.org/show_bug.cgi?id=797587)

### DIFF
--- a/gnucash/import-export/import-backend.c
+++ b/gnucash/import-export/import-backend.c
@@ -387,8 +387,24 @@ tokenize_string(GList* existing_tokens, const char *string)
     /* add each token to the token GList */
     while (stringpos && *stringpos)
     {
-        /* prepend the char* to the token GList */
-        existing_tokens = g_list_prepend(existing_tokens, g_strdup(*stringpos));
+        if (strlen(*stringpos) > 0)
+        {
+            /* check for duplicated tokens */
+            gboolean duplicated = FALSE;
+            for (GList* token = existing_tokens; token != NULL; token = token->next)
+            {
+                if (g_strcmp0(token->data, *stringpos) == 0)
+                {
+                    duplicated = TRUE;
+                    break;
+                }
+            }
+            if (duplicated == FALSE)
+            {
+                /* prepend the char* to the token GList */
+                existing_tokens = g_list_prepend(existing_tokens, g_strdup(*stringpos));
+            }
+        }
 
         /* then move to the next string */
         stringpos++;
--- a/libgnucash/engine/Account.cpp
+++ b/libgnucash/engine/Account.cpp
@@ -5281,13 +5281,15 @@ struct AccountInfo
 static void
 build_token_info(char const * suffix, KvpValue * value, TokenAccountsInfo & tokenInfo)
 {
-    tokenInfo.total_count += value->get<int64_t>();
-    AccountTokenCount this_account;
-    std::string account_guid {suffix};
-    /*By convention, the key ends with the account GUID.*/
-    this_account.account_guid = account_guid.substr(account_guid.size() - GUID_ENCODING_LENGTH);
-    this_account.token_count = value->get<int64_t>();
-    tokenInfo.accounts.push_back(this_account);
+    if (strlen(suffix) == GUID_ENCODING_LENGTH)
+    {
+        tokenInfo.total_count += value->get<int64_t>();
+        AccountTokenCount this_account;
+        /*By convention, the key ends with the account GUID.*/
+        this_account.account_guid = std::string{suffix};
+        this_account.token_count = value->get<int64_t>();
+        tokenInfo.accounts.push_back(this_account);
+    }
 }
 
 /** We scale the probability values by probability_factor.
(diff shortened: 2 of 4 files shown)


libgnucash/engine/kvp-frame.hpp
```diff
@@ -247,7 +247,7 @@ void KvpFrame::for_each_slot_prefix(std::string const & prefix,
                 return;
             /* Testing for prefix matching */
             if (std::mismatch(prefix.begin(), prefix.end(), temp_key.begin()).first == prefix.end())
-                func (a.first, a.second);
+                func (&a.first[prefix.size()], a.second);
```
cstim: This kind of expression in C++ makes me suspicious always (the `first[prefix.size()]` part). How can you guarantee that the container `a.first` has enough elements so that `prefix.size()` is not out of bounds? If there is a possibility to write this differently (such as using `back()`), please change that. Or add a comment for why this is ok here.

jralls: @cstim It's guaranteed that by line 245, which returns if `strlen(a.first) < prefix.size()`, but what you're complaining about isn't part of this PR, it's existing code.

@gruberchr If you want a performance win don't create a std::string just to use operator==(), use `if (strncmp(a.first, prefix.c_str(), prefix.size()))`. Even if the std::string ctor doesn't allocate it's still unnecessary overhead and I doubt that std::mismatch is any less efficient than either strncmp or std::string::operator==(); they all have to iterate over the char arrays and compare each until there's a mismatch. Any performance hit is still much more likely to be from the std::string ctor and if you saw an improvement it would be because with your test data the substring was able to avoid an alloc because it fit in your std library's SSTO window and the whole a.first didn't.

gruberchr: @jralls I didn't measure performance. I simply tried to exchange `std::mismatch` with string comparison as you suggested in the corresponding bug ticket. And AFAIKS string comparison uses `std::memcmp` internally, which seems to be faster than `std::mismatch`. Since `strncmp` has comparable performance than `memcmp`, the only performance difference between using C++ and C for string comparison should be the ctor overhead as you already mentioned.
I'll change this using C string comparison instead of C++ string comparison.

gruberchr: Any further remarks?

jralls: Yeah, this template should be removed, it's not used anywhere--even in tests. Only the three-argument version is used.

gnucash/import-export/import-backend.c
```diff
@@ -387,8 +387,25 @@ tokenize_string(GList* existing_tokens, const char *string)
     /* add each token to the token GList */
     while (stringpos && *stringpos)
     {
-        /* prepend the char* to the token GList */
-        existing_tokens = g_list_prepend(existing_tokens, g_strdup(*stringpos));
+        /* check for empty strings */
```
jralls: Leave out the comments that state the obvious.

libgnucash/engine/Account.cpp
```diff
@@ -5665,38 +5667,36 @@ build_non_bayes (const char *key, const GValue *value, gpointer user_data)
     g_free (guid_string);
 }
 
-static std::tuple<std::string, std::string, std::string>
+static std::tuple<std::string, std::string>
```
jralls: This function is expensively silly. Do it inline in `build_bayes` and use pointer math on suffix instead of constructing a string just to use `substr()`. The only string construction should be `account_guid` because `gnc::GUID::from_string()` takes a `const std::string&`.

libgnucash/engine/Account.cpp
```diff
@@ -5665,38 +5667,36 @@ build_non_bayes (const char *key, const GValue *value, gpointer user_data)
...
-build_bayes (const char *key, KvpValue * value, GncImapInfo & imapInfo)
+build_bayes (const char *suffix, KvpValue * value, GncImapInfo & imapInfo)
 {
-    auto parsed_key = parse_bayes_imap_info (key);
+    auto parsed_key = parse_bayes_imap_info (suffix);
     GncGUID guid;
     try
     {
-        auto temp_guid = gnc::GUID::from_string (std::get <2> (parsed_key));
+        auto temp_guid = gnc::GUID::from_string (std::get <1> (parsed_key));
```
jralls: No need for `temp_guid`, assign directly to `guid`.

libgnucash/engine/Account.cpp
```diff
@@ -5279,15 +5279,17 @@ struct AccountInfo
...
-    this_account.token_count = value->get<int64_t>();
-    tokenInfo.accounts.push_back(this_account);
+    if (strlen(suffix) == GUID_ENCODING_LENGTH)
+    {
+        tokenInfo.total_count += value->get<int64_t>();
+        AccountTokenCount this_account;
+        /*By convention, the key ends with the account GUID.*/
+        this_account.account_guid = std::string{suffix};
+        this_account.token_count = value->get<int64_t>();
+        tokenInfo.accounts.push_back(this_account);
```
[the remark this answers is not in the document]
gruberchr: Done

libgnucash/engine/kvp-frame.hpp
gruberchr: Done

gnucash/import-export/import-backend.c
gruberchr: Done

libgnucash/engine/Account.cpp
gruberchr: Done

libgnucash/engine/Account.cpp
gruberchr: Done

PR #642, opened 2020-02-03
Fix white text on white background

pau4o:
When dark theme is active in Gnome Shell it sets text colour to white, combined with background colour set in default style to white lead to unreadable reports.


gjanssens:
Thanks, that's a good fix for the problem.

PR #644, opened 2020-02-05
Follow up to previous commit 94cb965 / Bug 797522

Bob-IT:
This commit moves the setting up of the page changed signal callback to when the plugin page is inserted and also records the id used. The call back functions have been changed to reflect the signal spec and cast to the correct types. The id is used to disconnect this callback when the page is moved to a different window with the gnc_main_window_cmd_window_move_page function so it gets disconnected as early as possible before pages start changing. It is also used when the page is destroyed.

### DIFF
--- a/gnucash/gnome-utils/gnc-main-window.c
+++ b/gnucash/gnome-utils/gnc-main-window.c
@@ -2795,6 +2795,9 @@ gnc_main_window_disconnect (GncMainWindow *window,
     g_signal_handlers_disconnect_by_func(G_OBJECT(page->notebook_page),
                                          G_CALLBACK(gnc_main_window_button_press_cb), page);
 
+    // Remove the page_changed signal callback
+    gnc_plugin_page_disconnect_page_changed (GNC_PLUGIN_PAGE(page));
+
     /* Disconnect the page and summarybar from the window */
     priv = GNC_MAIN_WINDOW_GET_PRIVATE(window);
     if (priv->current_page == page)
@@ -3712,7 +3715,7 @@ gnc_quartz_shutdown (GtkosxApplication *theApp, gpointer data)
     /* Do Nothing. It's too late. */
 }
 /* Should quit responds to NSApplicationBlockTermination; returning
- * TRUE means "don't terminate", FALSE means "do terminate". 
+ * TRUE means "don't terminate", FALSE means "do terminate".
  */
 static gboolean
 gnc_quartz_should_quit (GtkosxApplication *theApp, GncMainWindow *window)
@@ -4324,6 +4327,9 @@ gnc_main_window_cmd_window_move_page (GtkAction *action, GncMainWindow *window)
     notebook = GTK_NOTEBOOK (priv->notebook);
     tab_widget = gtk_notebook_get_tab_label (notebook, page->notebook_page);
     menu_widget = gtk_notebook_get_menu_label (notebook, page->notebook_page);
+
+    // Remove the page_changed signal callback
+    gnc_plugin_page_disconnect_page_changed (GNC_PLUGIN_PAGE(page));
 
     /* Ref the page components, then remove it from its old window */
     g_object_ref(page);
--- a/gnucash/gnome-utils/gnc-plugin-page.c
+++ b/gnucash/gnome-utils/gnc-plugin-page.c
@@ -58,6 +58,9 @@ static void gnc_plugin_page_get_property (GObject         *object,
         GValue          *value,
         GParamSpec      *pspec);
 
+static void gnc_plugin_page_default_focus (GncPluginPage *plugin_page,
+                                           gboolean on_current_page);
+
 enum
 {
     INSERTED,
@@ -102,6 +105,9 @@ typedef struct _GncPluginPagePrivate
     gchar *page_color;
     gchar *uri;
     gchar *statusbar_text;
+
+    gulong page_changed_id;
+
 } GncPluginPagePrivate;
 
 GNC_DEFINE_TYPE_WITH_CODE(GncPluginPage, gnc_plugin_page, G_TYPE_OBJECT,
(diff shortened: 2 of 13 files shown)


jralls:
There's an awful lot of code duplication. Can't `gnc_plugin_page_foo_main_window_changed` and gnc_plugin_page_foo_inserted` be `GncPluginPage` functions with the latter the callback for the `inserted` signal connected in each `gnc_plugin_page_foo_create_widget`? I think that would also keep the `GncPluginPage.priv->page_connected_id` completely internal to `GncPluginPage`, removing the need for the getter and setter.

Bob-IT:
Yes, I could not think of a simpler way or probably could not see it. I have tried this, created this gnc_plugin_page function...

void
gnc_plugin_page_inserted_cb (GncPluginPage *page, gpointer user_data)
{
    GncPluginPagePrivate *priv;
    GFunc page_changed_cb = user_data;

    g_return_if_fail (GNC_IS_PLUGIN_PAGE(page));
    priv = GNC_PLUGIN_PAGE_GET_PRIVATE(page);
    priv->page_changed_id = g_signal_connect (G_OBJECT(page->window), "page_changed",
                                              G_CALLBACK(page_changed_cb),
                                              page);
//    gnc_plugin_page_register_main_window_changed (GNC_MAIN_WINDOW(window),
//                                                  G_OBJECT(plugin_page),
//                                                  plugin_page);
}

and used it from the register page create function like this...
    g_signal_connect (G_OBJECT(plugin_page), "inserted",
                      G_CALLBACK(gnc_plugin_page_inserted_cb),
                      gnc_plugin_page_register_main_window_changed);

which mostly works but can not figure out if I can still use the commented out part, it is required when the page is moved to set the initial focus other wise I could emit the 'page_changed' signal in gnc_main_window_connect after the 'page_added' one.

As for the gnc_plugin_page_foo_main_window_changed functions, they are not all the same and if it was moved to the plugin, I can not see how one would access the specific idle_add function that is used for each page focus.

jralls:
Why wouldn't the commented out call work? The target ptr is cast all the way down to a GObject. 

`gnc_plugin_page_foo_main_window_changed ` seem to differ only in the name of the function that they set the idle on, which is `gnc_plugin_page_foo_focus`.  There seems to be a bit of variation there about what widget each passes focus to, including at least one where it's conditional so it's probably most straightforward to store the GObject* of the item that is passed to `g_idle_remove_by_data` in GncPluginPage and to [virtualize](https://developer.gnome.org/gobject/stable/howto-gobject-methods.html) `gnc_plugin_page_foo_focus` so that it can easily be called from `gnc_plugin_page_main_window_changed`.

Bob-IT:
I have made these changes on what I understand...
Changed all the page focus functions to contain the page specific logic so the idle function just takes a plugin_page. In doing so the 'gnc_plugin' function gnc_plugin_page_inserted_cb now sets up the callback gnc_plugin_page_main_window_changed for the 'page_changed' signal and uses the gpointer to record a pointer to the page focus function so that it can be added/removed.

Also setup a virtual function 'focus_page_widget' that is used for the register pages as that page focus function is more involved.

jralls:
Much better, but why did you make both a `priv->page_focus_func` and a `class->focus_page_widget`? You might also hide the virtual function dispatch (including the test of which function is implemented) with `gnc_plugin_page_focus_page_widget()` as demonstrated in the [GObject doc](https://developer.gnome.org/gobject/stable/howto-gobject-methods.html#virtual-private-methods).

Bob-IT:
I did not think I could achieve what I wanted with just one way, this is how I see it...
All plugins by default follow this sequence...
gnc_plugin_page_foo_create_widget sets up the page 'inserted' call back to gnc_plugin_page_inserted_cb with the user_data pointing to the specific plugin page focus function.

gnc_plugin_page_inserted_cb sets up the 'page_changed' call back to gnc_plugin_page_main_window_changed and also saves the focus function pointer.

gnc_plugin_page_main_window_changed tests for the plugin_page equalling the current_page and then call the class->focus_page_widget, if one is not set it now uses the default one gnc_plugin_page_default_focus_widget which setup the g_idle function with the saved focus function pointer if it is the current_page.

With the register plugin I needed to do more specific actions if on the current_page or not so it has an over ride function for class->focus_page_widget.

jralls:
I think I got the question wrong. It should be why is there `priv->page_focus_func` used by every class *except* `GncPluginPageRegister ` and `klass->focus_page_widget` that's overridden only by `GncPluginPageRegister`?  And why `priv->page_focus_func` instead of the more usual `klass->page_focus_func`? 

I can answer the first question on my own: Only GncPluginPageRegister needs to conditionally set the idle, so all of the other subclasses can use the parent class focus_page_widget code, and since GncPluginPageRegister has a custom `focus_page_widget` it can directly set the idle with `gnc_plugin_page_register_focus_func` instead of using the `priv->page_focus_func` indirection. But if `gnc_plugin_page_register_focus_widget` did
```
if (on_current_page)
    parent->focus_page_widget(plugin_page);  // set the idle
else
   priv->page_focus_func(plugin_page);
```
It would reduce the copied code still more *and* make the intent clearer. A comment on the function explaining why registers alone need that special handling would be helpful down the road.

Next, if `page_focus_func` was a class func instead of an instance funcptr you wouldn't need to pass it as user_data in the signal closure, and `gnc_plugin_page_default_focus_widget` could simply clobber the old idle and create a new one with `klass->page_focus_func`, protected by `if (G_LIKELY(klass->page_focus_func))` for safety.

Bob-IT:
I started down the passing the function pointer route which seemed a good fit for most of the plugins and came up with the different approach for the register. I did not see a way of simply combining but from your thoughts I think I should off. Calling a parent object function from a child is some thing that never occurred.

I think this latest push encompasses your feed back, thank you.

jralls:
Very good. Push when ready.

PR #646, opened 2020-02-09
quality-of-life improvements for python bindings

andygoblins:
Two changes:
1. Make equality tests work for Split and Transaction. When calling things like transaction.GetSplitList(), you get new class instances of the Splits which will not compare correctly with existing instances. See unit tests for details. I went with a GUID compare--if you think it should be more than that, let me know.
2. Make Session into a context manager. This makes the API feel a little more pythonic, and helps people like me who forget to call end() or save(). Since this isn't a substantial change I didn't write any unit tests; I updated one of the example scripts as a demo.

Let me know what suggestions you have.

### DIFF
--- a/bindings/python/gnucash_core.py
+++ b/bindings/python/gnucash_core.py
@@ -434,6 +434,9 @@ class Transaction(GnuCashCoreClass):
         return self.do_lookup_create_oo_instance(
             gncInvoiceGetInvoiceFromTxn, Transaction )
 
+    def __eq__(self, other):
+        return self.Equal(other, True, False, False, False)
+
 def decorate_monetary_list_returning_function(orig_function):
     def new_function(self, *args):
         """decorate function that returns list of gnc_monetary to return tuples of GncCommodity and GncNumeric
@@ -460,6 +463,9 @@ class Split(GnuCashCoreClass):
     another.
     """
     _new_instance = 'xaccMallocSplit'
+
+    def __eq__(self, other):
+        return self.Equal(other, True, False, False)
 
 class Account(GnuCashCoreClass):
     """A GnuCash Account.
--- a/bindings/python/gnucash_core.py
+++ b/bindings/python/gnucash_core.py
@@ -117,6 +117,15 @@ class Session(GnuCashCoreClass):
                 self.destroy()
                 raise
 
+    def __enter__(self):
+        return self
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        # Roll back changes on exception by not calling save. Only works for XMl backend.
+        if not exc_type:
+            self.save()
+        self.end()
+
     def raise_backend_errors(self, called_function="qof_session function"):
         """Raises a GnuCashBackendException if there are outstanding
         QOF_BACKEND errors.
--- a/bindings/python/example_scripts/simple_book.py
+++ b/bindings/python/example_scripts/simple_book.py
@@ -11,15 +11,15 @@ from gnucash import Session
 uri = "xml:///tmp/simple_book.gnucash"
 
 print("uri:", uri)
-ses = Session(uri, is_new=True)
-book = ses.get_book()
+with Session(uri, is_new=True) as ses:
+    book = ses.get_book()
 
-#Call some methods that produce output to show that Book works
-book.get_root_account().SetDescription("hello, book")
+    #Call some methods that produce output to show that Book works
+    book.get_root_account().SetDescription("hello, book")
+    print("Book is saved:", not book.session_not_saved())
+
+    #As long as there's no exceptions, book is automatically saved
+    #when session ends.
+    print("saving...")
+
 print("Book is saved:", not book.session_not_saved())
-
-print("saving...")
-ses.save()
-
-print("Book is saved:", not book.session_not_saved())
-ses.end()
--- a/bindings/python/example_scripts/simple_session.py
+++ b/bindings/python/example_scripts/simple_session.py
@@ -19,18 +19,13 @@ except GnuCashBackendException as backend_exception:
 
 
 # create a new file, this requires a file type specification
-session = Session("xml://%s" % FILE_2, is_new=True)
-session.save()
-session.end()
-session.destroy()
+with Session("xml://%s" % FILE_2, is_new=True) as session:
+    book = session.book
+    root = book.get_root_account()
 
 # open the new file, try to open it a second time, detect the lock
-session = Session(FILE_2)
-try:
-    session_2 = Session(FILE_2)
-except GnuCashBackendException as backend_exception:
-    assert( ERR_BACKEND_LOCKED in backend_exception.errors )
-session.end()
-session.destroy()
-
-    
+with Session(FILE_2) as session:
+    try:
+        session_2 = Session(FILE_2)
+    except GnuCashBackendException as backend_exception:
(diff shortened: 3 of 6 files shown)


bindings/python/gnucash_core.py
```diff
@@ -117,6 +117,15 @@ def __init__(self, book_uri=None, ignore_lock=False, is_new=False,
                 self.destroy()
                 raise
 
+    def __enter__(self):
+        return self
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        # Roll back changes on exception by not calling save. Only works for XMl backend.
```
gjanssens: To avoid unneeded saves you can first test if the book is dirty before calling save. As sql based books are never in a dirty state that also avoids calling save on sql based books.

andygoblins: My understanding is that the underlying code is already checking for a dirty book:
 Session.save() is qof_session_save, which calls QofSessionImpl::save, which does nothing if !qof_book_session_not_saved.

andygoblins:
Updated two more scripts as context manager examples. A lot of the other scripts need more significant makeovers, which is probably the topic of a future PR.

bindings/python/gnucash_core.py
gjanssens: That's true indeed. So why the comment this would only work for XML backend ?

gjanssens:
> Updated two more scripts as context manager examples. A lot of the other scripts need more significant makeovers, which is probably the topic of a future PR.

I don't see the two extra modified scripts in the PR yet. Do you intend to keep those for your future PR as well ?

bindings/python/gnucash_core.py
andygoblins: My understanding is that qof_session_save is a noop for sql backends, so the rollback would not occur for sql backends. Is that correct? I'm not an expert on the codebase.

Is there a better way I can word that comment to be clear?

andygoblins:
Oops. I forgot to push those changes (^_^)

gjanssens:
Hmm, while preparing to merge I ran ninja-build check and got this test failure:

```
UpdateCTestConfiguration  from :/home/janssege/Development/builds/gnucash/master/DartConfiguration.tcl
UpdateCTestConfiguration  from :/home/janssege/Development/builds/gnucash/master/DartConfiguration.tcl
Test project /home/janssege/Development/builds/gnucash/master
Constructing a list of tests
Done constructing a list of tests
Updating test list for fixtures
Added 0 tests to meet fixture requirements
Checking test dependency graph...
Checking test dependency graph end
test 120
    Start 120: python-bindings

120: Test command: /usr/bin/python3 "/home/janssege/Development/gnucash/master/bindings/python/tests/runTests.py.in"
120: Environment variables: 
120:  GNC_BUILDDIR=/home/janssege/Development/builds/gnucash/master
120:  PYTHONPATH=/home/janssege/Development/builds/gnucash/master/lib64/python3.7/site-packages:/home/janssege/Development/builds/gnucash/master/lib64/gnucash:/home/janssege/Development/builds/gnucash/master/common/test-core
120: Test timeout computed to be: 10000000
120: * 20:24:44  WARN <gnc.core-utils> no backend loaded, or the backend doesn't define register_cb, returning 0
120: * 20:24:44  WARN <gnc.core-utils> no backend loaded, or the backend doesn't define register_cb, returning 0
120: * 20:24:44  WARN <gnc.core-utils> no backend loaded, or the backend doesn't define register_cb, returning 0
120: * 20:24:44  WARN <gnc.core-utils> no backend loaded, or the backend doesn't define register_cb, returning 0
120: * 20:24:44  WARN <gnc.core-utils> no backend loaded, or the backend doesn't define register_cb, returning 0
120: ..........................* 20:24:44 ERROR <gnc.engine> [xaccTransScrubSplits()] Transaction doesn't have a currency!
120: * 20:24:44 ERROR <gnc.engine.scrub> [xaccScrubUtilityGetOrMakeAccount()] No currency specified!
120: F......F.........
120: ======================================================================
120: FAIL: test_equal (test_split.TestSplit)
120: ----------------------------------------------------------------------
120: Traceback (most recent call last):
120:   File "/home/janssege/Development/gnucash/master/bindings/python/tests/test_split.py", line 63, in test_equal
120:     self.assertTrue( self.split == TRANS.GetSplitList()[0] )
120: AssertionError: False is not true
120: 
120: ======================================================================
120: FAIL: test_equal (test_transaction.TestTransaction)
120: ----------------------------------------------------------------------
120: Traceback (most recent call last):
120:   File "/home/janssege/Development/gnucash/master/bindings/python/tests/test_transaction.py", line 50, in test_equal
120:     self.assertTrue( self.trans == SPLIT.GetParent() )
120: AssertionError: False is not true
```

Can you look into this ?

gjanssens:
Note you can ignore the missing backend warning. This is was already there before.
Also note our CI doesn't test the python bindings, so you need to run `make check` or `ninja(-build) check` yourself to run our unit test suite.

gjanssens:
Oh, and while you're at it, please squash the commit that introduces the equal tests with the commit that implements the equal functions.
Commits should never deliberately break the test suite. That would needlessly complicate future git bisects.

andygoblins:
Commits squashed.

That failure you see is the test I wrote for the __eq__ functions. It seems to be working fine for me. Did you remember to build before testing? I made that mistake a few times and it caused my test to fail.
```
119/119 Testing: python-bindings
119/119 Test: python-bindings
Command: "/usr/bin/python3" "/home/andy/hack/gnucash/gnucash-git/bindings/python/tests/runTests.py.in"
Directory: /home/andy/hack/gnucash/build/bindings/python/tests
"python-bindings" start time: Feb 29 15:59 CST
Output:
* 15:59:23  WARN <gnc.core-utils> no backend loaded, or the backend doesn't define register_cb, returning 0
* 15:59:23  WARN <gnc.core-utils> no backend loaded, or the backend doesn't define register_cb, returning 0
* 15:59:23  WARN <gnc.core-utils> no backend loaded, or the backend doesn't define register_cb, returning 0
* 15:59:23  WARN <gnc.core-utils> no backend loaded, or the backend doesn't define register_cb, returning 0
* 15:59:23  WARN <gnc.core-utils> no backend loaded, or the backend doesn't define register_cb, returning 0
..........................* 15:59:23 ERROR <gnc.engine> [xaccTransScrubSplits()] Transaction doesn't have a currency!
* 15:59:23 ERROR <gnc.engine.scrub> [xaccScrubUtilityGetOrMakeAccount()] No currency specified!
.................
----------------------------------------------------------------------
Ran 43 tests in 0.235s

OK
<end of output>
Test time =   0.50 sec
Test Passed.
"python-bindings" end time: Feb 29 15:59 CST
"python-bindings" time elapsed: 00:00:00

End testing: Feb 29 15:59 CST
```

gjanssens:
It turns out rebuilding was not sufficient, for the test to succeed I need to start from a clean build directory. This points in the direction of a build dependency issue. The most likely source is [the file (COPY) command](https://github.com/Gnucash/gnucash/blob/master/bindings/python/CMakeLists.txt#L105).

However this is not caused by your PR, so I'm merging this in and will create a followup bug report.

Thanks for your improvements!

gjanssens:
[Followup bug report](https://bugs.gnucash.org/show_bug.cgi?id=797634)

PR #647, opened 2020-02-10
Scrub Import Map Editor

christopherlam:
Creates a `Scrub` button in Import Map Editor. It scans all gtk_list rows, finds mapping entries whereby map_account is NULL, and deletes them.

* probably missed g_free ()
* gui doesn't immediately refresh, need to close and reopen dialog to see the nomapping entries removed.

### DIFF
--- a/gnucash/gnome/dialog-imap-editor.c
+++ b/gnucash/gnome/dialog-imap-editor.c
@@ -181,27 +181,27 @@ delete_selected_row (GtkTreeModel *model, GtkTreeIter *iter, ImapDialog *imap_di
         g_free (full_source_account);
 }
 
-static void
+static gboolean
 gnc_delete_nomapping (GtkTreeModel *model, GtkTreePath *path,
                       GtkTreeIter *iter, gpointer data)
 {
     Account     *source_account = NULL;
     Account     *map_account = NULL;
     gchar       *head;
-    GtkTreePath *tree_path;
     gint         depth;
 
     gtk_tree_model_get (model, iter, SOURCE_ACCOUNT, &source_account,
                         MAP_ACCOUNT, &map_account, HEAD, &head, -1);
-    tree_path = gtk_tree_model_get_path (model, iter);
-    depth = gtk_tree_path_get_depth (tree_path);
+    depth = gtk_tree_path_get_depth (path);
     gtk_tree_path_free (tree_path);
 
-    if ((source_account == NULL) || (map_account != NULL) || (depth == 1))
-        return;
-
-    PWARN ("Delete map: Acc=%s, head=%s", xaccAccountGetName(source_account), head);
-    gnc_account_delete_map_entry (source_account, head, NULL, NULL, FALSE);
+    if (!((source_account == NULL) || (map_account != NULL) || (depth == 1)))
+    {
+        PWARN ("Delete map: Acc=%s, head=%s", xaccAccountGetName(source_account), head);
+        gnc_account_delete_map_entry (source_account, head, NULL, NULL, FALSE);
+    }
+    g_free(head);
+    return FALSE;
 }
 
 static void
--- a/gnucash/gnome/dialog-imap-editor.c
+++ b/gnucash/gnome/dialog-imap-editor.c
@@ -182,6 +182,40 @@ delete_selected_row (GtkTreeModel *model, GtkTreeIter *iter, ImapDialog *imap_di
 }
 
 static void
+gnc_delete_nomapping (GtkTreeModel *model, GtkTreePath *path,
+                      GtkTreeIter *iter, gpointer data)
+{
+    Account     *source_account = NULL;
+    Account     *map_account = NULL;
+    gchar       *head;
+    GtkTreePath *tree_path;
+    gint         depth;
+
+    gtk_tree_model_get (model, iter, SOURCE_ACCOUNT, &source_account,
+                        MAP_ACCOUNT, &map_account, HEAD, &head, -1);
+    tree_path = gtk_tree_model_get_path (model, iter);
+    depth = gtk_tree_path_get_depth (tree_path);
+    gtk_tree_path_free (tree_path);
+
+    if ((source_account == NULL) || (map_account != NULL) || (depth == 1))
+        return;
+
+    PWARN ("Delete map: Acc=%s, head=%s", xaccAccountGetName(source_account), head);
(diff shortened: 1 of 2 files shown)


gnucash/gnome/dialog-imap-editor.c
```diff
@@ -181,6 +181,40 @@ delete_selected_row (GtkTreeModel *model, GtkTreeIter *iter, ImapDialog *imap_di
         g_free (full_source_account);
 }
 
+static void
+gnc_delete_nomapping (GtkTreeModel *model, GtkTreePath *path,
```
Bob-IT: this function is defined wrong, look at the definition of GtkTreeMOdelForeachFunc and you see it should be a static gboolean and you will have to return TRUE/FALSE, in your case probably FALSE.

gnucash/gnome/dialog-imap-editor.c
```diff
@@ -181,6 +181,40 @@ delete_selected_row (GtkTreeModel *model, GtkTreeIter *iter, ImapDialog *imap_di
...
+{
+    Account     *source_account = NULL;
+    Account     *map_account = NULL;
+    gchar       *head;
+    GtkTreePath *tree_path;
+    gint         depth;
+
+    gtk_tree_model_get (model, iter, SOURCE_ACCOUNT, &source_account,
+                        MAP_ACCOUNT, &map_account, HEAD, &head, -1);
+    tree_path = gtk_tree_model_get_path (model, iter);
```
Bob-IT: Why not use the tree path being passed in?

gnucash/gnome/dialog-imap-editor.c
```diff
@@ -181,6 +181,40 @@ delete_selected_row (GtkTreeModel *model, GtkTreeIter *iter, ImapDialog *imap_di
         g_free (full_source_account);
 }
 
+static void
+gnc_delete_nomapping (GtkTreeModel *model, GtkTreePath *path,
+                      GtkTreeIter *iter, gpointer data)
+{
+    Account     *source_account = NULL;
+    Account     *map_account = NULL;
+    gchar       *head;
```
Bob-IT: You will need to free the head

christopherlam:
Ideally a dialog stating how many entries were deleted too.

Bob-IT:
Chris,
Does 'Scrub' mean any thing to the user, should there be a tooltip for the button.
You should be aware that this tree view connects to a model filter that connects to a model so if you are using iters and the such you may need to convert from filter to model to filter.
In your existing setup you can just add get_account_info (imap_dialog) before you resume the gui in gnc_imap_scrub.

What I can not decide is if you should disconnect the model from the tree view, generally depends on how many entries you are deleting.

Bob-IT:
What I would do for the number of deleted entries is create another foreach function that did nothing but count the number of possible deletions and return it. Then in gnc_imap_scrub use one of the gnc dialogs to present this number with a question to proceed, yes/no and depending on this continue with the foreach that does delete them.

christopherlam:
> Does 'Scrub' mean any thing to the user, should there be a tooltip for the button.
> You should be aware that this tree view connects to a model filter that connects to a model so if you are using iters and the such you may need to convert from filter to model to filter.

Yeah, about that: if there is Filter applied, I guess it'll only delete nomap entries in filter. I wonder if this is useful. IMHO any entry whereby account can't be found is of no use, and should be scrubbed upon load. I noticed this when importing my .OFX and found that several tokens never matched because the account-guid was obsolete.

> In your existing setup you can just add get_account_info (imap_dialog) before you resume the gui in gnc_imap_scrub.

Thanks, I guess `get_account_info()` will refresh thedialog.

christopherlam:
> Yeah, about that: if there is Filter applied, I guess it'll only delete nomap entries in filter. I wonder if this is useful. IMHO any entry whereby account can't be found is of no use, and should be scrubbed upon load. I noticed this when importing my .OFX and found that several tokens never matched because the account-guid was obsolete.

So, my view is that this extra button *shouldn't* be needed at all, rather the mapping entries scrubbed upon load?

gjanssens:
Op donderdag 13 februari 2020 03:46:33 CET schreef Christopher Lam:
> > Yeah, about that: if there is Filter applied, I guess it'll only delete
> > nomap entries in filter. I wonder if this is useful. IMHO any entry
> > whereby account can't be found is of no use, and should be scrubbed upon
> > load. I noticed this when importing my .OFX and found that several tokens
> > never matched because the account-guid was obsolete.
> So, my view is that this extra button *shouldn't* be needed at all, rather
> the mapping entries scrubbed upon load?

That decision depends a bit on the performance of the scrub. If the time it takes is 
negligible, I don't think a button is needed.

For comparison the conversion to flat bayes that happens while converting a 2.6.x 
book to a 3.x book can take such a long time users currently think gnucash is locked 
up. It was very unfortunate we don't have any user feedback during that conversion. 
Let's not make that mistake again.

Bob-IT:
I did think of another approach, on load it could count the number of 'No maps' and maybe bring up a dialog stating this with an option to delete them now, if they say no add the button. If a large number I would disconnect the tree view before deleting.

christopherlam:
> That decision depends a bit on the performance of the scrub. If the time it takes is negligible, I don't think a button is needed. For comparison the conversion to flat bayes that happens while converting a 2.6.x book to a 3.x book can take such a long time users currently think gnucash is locked up. It was very unfortunate we don't have any user feedback during that conversion. Let's not make that mistake again.

Sounds like a job for the status bar and the progress bar :) either upon file-load, or upon loading import-map-editor.

> I did think of another approach, on load it could count the number of 'No maps' and maybe bring up a dialog stating this with an option to delete them now, if they say no add the button. If a large number I would disconnect the tree view before deleting.
> If it is a small number, I would get the model iter and emit a row delete signal to update the treeview which would save time as you would not be doing a full reload.

Sounds more difficult that I'd know how to code 😕

Bob-IT:
> Sounds more difficult that I'd know how to code confused

Most off the required information is already in the file, if you look at add_to_store you see where the 'No map' text gets added so you could add a gint to imapInfo to record the number, you could also get the total here also. get_account_info has some model stuff and you could plumb your dialog into the end of the create function.
Have a think about it and if you are not going to pursue, I may, just need to complete some other bits first.

christopherlam:
@Bob-IT  Feel free to take over as you like. I'd aimed to insert a hack avoiding the other dialog functions, and you'd be much better at merging it with the existing workflow.

christopherlam:
Superceded by a higher authority

PR #648, opened 2020-02-17
[Master] State file changes

Bob-IT:
Currently the comments, which I think are useful as the group entries are based on guid's, used in the state file are only saved for the open registers as on load all comments are removed. Tried to use G_KEY_FILE_KEEP_COMMENTS but that still did not work in the way wanted even if you deleted the group.
So the first commit changes saving a comment to using a dummy key and add the comment there along with also adding it to the budget groups which are also based on guid's.

Invoice/Bill/Vouchers never saved there sheet layouts so have added that as a separate commit. This works OK but a bit tedious having to change each layout so have added an option to save a layout for each as a default layout. As there maybe many, once an invoice/bill/voucher is paid the saved layout is dropped and the default one used if present.

As I was in the area, I think there was mention of a register reset layout option so have added that also.

If there are no issues with this I aim to push this at the weekend.

### DIFF
--- a/gnucash/gnome/dialog-invoice.c
+++ b/gnucash/gnome/dialog-invoice.c
@@ -518,19 +518,19 @@ gnc_invoice_window_help_cb (GtkWidget *widget, gpointer data)
     gnc_gnome_help(HF_HELP, HL_USAGE_INVOICE);
 }
 
-static gchar *
+static const gchar *
 gnc_invoice_window_get_state_group (InvoiceWindow *iw)
 {
     switch (gncOwnerGetType (gncOwnerGetEndOwner (&iw->owner)))
     {
         case GNC_OWNER_VENDOR:
-            return g_strdup ("Vendor documents");
+            return "Vendor documents";
             break;
         case GNC_OWNER_EMPLOYEE:
-            return g_strdup ("Employee documents");
+            return "Employee documents";
             break;
         default:
-            return g_strdup ("Customer documents");
+            return "Customer documents";
             break;
     }
 }
@@ -542,10 +542,9 @@ void
 gnc_invoice_window_save_document_layout_to_user_state (InvoiceWindow *iw)
 {
     Table *table = gnc_entry_ledger_get_table (iw->ledger);
-    gchar *group = gnc_invoice_window_get_state_group (iw);
+    const gchar *group = gnc_invoice_window_get_state_group (iw);
 
     gnc_table_save_state (table, group, NULL);
-    g_free (group);
 }
 
 /* Removes the user state layout information for Invoice/Bill/Voucher
@@ -555,11 +554,10 @@ void
 gnc_invoice_window_reset_document_layout_and_clear_user_state (InvoiceWindow *iw)
 {
     GnucashRegister *reg = iw->reg;
-    gchar *group = gnc_invoice_window_get_state_group (iw);
+    const gchar *group = gnc_invoice_window_get_state_group (iw);
 
     gnucash_register_reset_sheet_layout (reg);
     gnc_state_drop_sections_for (group);
-    g_free (group);
 }
 
 /* Checks to see if there is user state layout information for
@@ -570,10 +568,8 @@ gboolean
 gnc_invoice_window_document_has_user_state (InvoiceWindow *iw)
 {
     GKeyFile *state_file = gnc_state_get_current ();
-    gchar *group = gnc_invoice_window_get_state_group (iw);
-    gboolean has_group = g_key_file_has_group (state_file, group);
-    g_free (group);
-    return has_group;
+    const gchar *group = gnc_invoice_window_get_state_group (iw);
+    return g_key_file_has_group (state_file, group);
 }
 
 void
@@ -2351,7 +2347,6 @@ gnc_invoice_save_page (InvoiceWindow *iw,
                        const gchar *group_name)
 {
     Table *table = gnc_entry_ledger_get_table (iw->ledger);
-    gchar *group = g_strdup (group_name);
     gchar guidstr[GUID_ENCODING_LENGTH+1];
     guid_to_string_buff(&iw->invoice_guid, guidstr);
     g_key_file_set_string(key_file, group_name, KEY_INVOICE_TYPE,
@@ -2373,8 +2368,7 @@ gnc_invoice_save_page (InvoiceWindow *iw,
         g_key_file_set_string(key_file, group_name, KEY_OWNER_GUID, guidstr);
     }
     // save the open table layout
-    gnc_table_save_state (table, group, NULL);
-    g_free (group);
+    gnc_table_save_state (table, group_name, NULL);
 }
 
 GtkWidget *
@@ -2568,21 +2562,20 @@ gnc_invoice_create_page (InvoiceWindow *iw, gpointer page)
     /* Create the register */
     {
         GtkWidget *regWidget, *frame, *window;
-        gchar *default_group = gnc_invoice_window_get_state_group (iw);
-        gchar *group;
+        const gchar *default_group = gnc_invoice_window_get_state_group (iw);
+        const gchar *group;
 
         // if this is from a page recreate, use those settings
         if (iw->page_state_name)
-            group = g_strdup (iw->page_state_name);
+            group = iw->page_state_name;
         else
-            group = g_strdup (default_group);
+            group = default_group;
 
         /* Watch the order of operations, here... */
         regWidget = gnucash_register_new (gnc_entry_ledger_get_table
                                           (entry_ledger), group);
         gtk_widget_show(regWidget);
-        g_free (default_group);
-        g_free (group);
+
         frame = GTK_WIDGET (gtk_builder_get_object (builder, "ledger_frame"));
         gtk_container_add (GTK_CONTAINER (frame), regWidget);
 
--- a/gnucash/gnome/dialog-invoice.c
+++ b/gnucash/gnome/dialog-invoice.c
@@ -71,6 +71,7 @@
 #include "gnc-plugin-business.h"
 #include "gnc-plugin-page-invoice.h"
 #include "gnc-main-window.h"
+#include "gnc-state.h"
 
 #include "dialog-transfer.h"
 
@@ -141,6 +142,7 @@ struct _invoice_window
 
     GtkWidget  * dialog;         /* Used by 'New Invoice Window' */
     GncPluginPage *page;        /* Used by 'Edit Invoice' Page */
+    const gchar * page_state_name;    /* Used for loading open state information */
 
     /* Summary Bar Widgets */
     GtkWidget  * total_label;
@@ -514,6 +516,64 @@ void
 gnc_invoice_window_help_cb (GtkWidget *widget, gpointer data)
 {
     gnc_gnome_help(HF_HELP, HL_USAGE_INVOICE);
+}
+
+static gchar *
+gnc_invoice_window_get_state_group (InvoiceWindow *iw)
+{
+    switch (gncOwnerGetType (gncOwnerGetEndOwner (&iw->owner)))
+    {
+        case GNC_OWNER_VENDOR:
+            return g_strdup ("Vendor documents");
+            break;
+        case GNC_OWNER_EMPLOYEE:
+            return g_strdup ("Employee documents");
+            break;
+        default:
+            return g_strdup ("Customer documents");
+            break;
+    }
+}
+
+/* Save user state layout information for Invoice/Bill/Voucher
+ * documents so it can be used for the default user set layout
+ */
+void
+gnc_invoice_window_save_document_layout_to_user_state (InvoiceWindow *iw)
+{
+    Table *table = gnc_entry_ledger_get_table (iw->ledger);
+    gchar *group = gnc_invoice_window_get_state_group (iw);
+
+    gnc_table_save_state (table, group, NULL);
+    g_free (group);
+}
+
+/* Removes the user state layout information for Invoice/Bill/Voucher
+ * documents and also resets the current layout to the built-in defaults
+ */
+void
+gnc_invoice_window_reset_document_layout_and_clear_user_state (InvoiceWindow *iw)
+{
+    GnucashRegister *reg = iw->reg;
+    gchar *group = gnc_invoice_window_get_state_group (iw);
+
+    gnucash_register_reset_sheet_layout (reg);
+    gnc_state_drop_sections_for (group);
+    g_free (group);
+}
+
+/* Checks to see if there is user state layout information for
+ * Invoice/Bill/Voucher documents so it can be used for the
+ * default user layout
+ */
+gboolean
+gnc_invoice_window_document_has_user_state (InvoiceWindow *iw)
+{
+    GKeyFile *state_file = gnc_state_get_current ();
+    gchar *group = gnc_invoice_window_get_state_group (iw);
+    gboolean has_group = g_key_file_has_group (state_file, group);
+    g_free (group);
+    return has_group;
 }
 
 void
@@ -2122,7 +2182,7 @@ find_handler (gpointer find_data, gpointer user_data)
 static InvoiceWindow *
 gnc_invoice_new_page (QofBook *bookp, InvoiceDialogType type,
                       GncInvoice *invoice, const GncOwner *owner,
-                      GncMainWindow *window)
+                      GncMainWindow *window, const gchar *group_name)
 {
     InvoiceWindow *iw;
     GncOwner *billto;
@@ -2158,6 +2218,7 @@ gnc_invoice_new_page (QofBook *bookp, InvoiceDialogType type,
     iw->invoice_guid = *gncInvoiceGetGUID (invoice);
     iw->is_credit_note = gncInvoiceGetIsCreditNote (invoice);
     iw->width = -1;
+    iw->page_state_name = group_name;
 
     /* Save this for later */
     gncOwnerCopy (gncOwnerGetEndOwner (owner), &(iw->owner));
@@ -2270,7 +2331,7 @@ gnc_invoice_recreate_page (GncMainWindow *window,
     g_free(tmp_string);
     g_free(owner_type);
 
-    iw = gnc_invoice_new_page (book, type, invoice, &owner, window);
+    iw = gnc_invoice_new_page (book, type, invoice, &owner, window, group_name);
     return iw->page;
 
 give_up:
@@ -2289,6 +2350,8 @@ gnc_invoice_save_page (InvoiceWindow *iw,
                        GKeyFile *key_file,
                        const gchar *group_name)
 {
+    Table *table = gnc_entry_ledger_get_table (iw->ledger);
+    gchar *group = g_strdup (group_name);
     gchar guidstr[GUID_ENCODING_LENGTH+1];
     guid_to_string_buff(&iw->invoice_guid, guidstr);
     g_key_file_set_string(key_file, group_name, KEY_INVOICE_TYPE,
@@ -2309,6 +2372,9 @@ gnc_invoice_save_page (InvoiceWindow *iw,
         guid_to_string_buff(gncOwnerGetGUID(&iw->owner), guidstr);
         g_key_file_set_string(key_file, group_name, KEY_OWNER_GUID, guidstr);
     }
+    // save the open table layout
+    gnc_table_save_state (table, group, NULL);
+    g_free (group);
 }
 
 GtkWidget *
@@ -2502,12 +2568,21 @@ gnc_invoice_create_page (InvoiceWindow *iw, gpointer page)
     /* Create the register */
     {
         GtkWidget *regWidget, *frame, *window;
+        gchar *default_group = gnc_invoice_window_get_state_group (iw);
+        gchar *group;
+
+        // if this is from a page recreate, use those settings
+        if (iw->page_state_name)
+            group = g_strdup (iw->page_state_name);
+        else
+            group = g_strdup (default_group);
 
         /* Watch the order of operations, here... */
         regWidget = gnucash_register_new (gnc_entry_ledger_get_table
-                                          (entry_ledger), NULL);
+                                          (entry_ledger), group);
         gtk_widget_show(regWidget);
-
+        g_free (default_group);
+        g_free (group);
         frame = GTK_WIDGET (gtk_builder_get_object (builder, "ledger_frame"));
         gtk_container_add (GTK_CONTAINER (frame), regWidget);
 
@@ -2779,7 +2854,7 @@ gnc_ui_invoice_edit (GtkWindow *parent, GncInvoice *invoice)
 
     iw = gnc_invoice_new_page (gncInvoiceGetBook(invoice), type,
                                invoice, gncInvoiceGetOwner (invoice),
-                               GNC_MAIN_WINDOW(gnc_ui_get_main_window (GTK_WIDGET (parent))));
+                               GNC_MAIN_WINDOW(gnc_ui_get_main_window (GTK_WIDGET (parent))), NULL);
 
     return iw;
 }
--- a/gnucash/gnome/dialog-invoice.c
+++ b/gnucash/gnome/dialog-invoice.c
@@ -544,7 +544,7 @@ gnc_invoice_window_save_document_layout_to_user_state (InvoiceWindow *iw)
     Table *table = gnc_entry_ledger_get_table (iw->ledger);
     const gchar *group = gnc_invoice_window_get_state_group (iw);
 
-    gnc_table_save_state (table, group, NULL);
+    gnc_table_save_state (table, group);
 }
 
 /* Removes the user state layout information for Invoice/Bill/Voucher
@@ -2368,7 +2368,7 @@ gnc_invoice_save_page (InvoiceWindow *iw,
         g_key_file_set_string(key_file, group_name, KEY_OWNER_GUID, guidstr);
     }
     // save the open table layout
-    gnc_table_save_state (table, group_name, NULL);
+    gnc_table_save_state (table, group_name);
 }
 
 GtkWidget *
--- a/gnucash/gnome/dialog-invoice.h
+++ b/gnucash/gnome/dialog-invoice.h
@@ -110,6 +110,10 @@ void gnc_invoice_window_duplicateCB (GtkWidget *widget, gpointer data);
 void gnc_invoice_window_payment_cb (GtkWindow *parent, gpointer data);
 void gnc_invoice_window_report_owner_cb (GtkWindow *parent, gpointer data);
 
+void gnc_invoice_window_save_document_layout_to_user_state (InvoiceWindow *iw);
+void gnc_invoice_window_reset_document_layout_and_clear_user_state (InvoiceWindow *iw);
+gboolean gnc_invoice_window_document_has_user_state (InvoiceWindow *iw);
+
 void gnc_invoice_window_entryUpCB (GtkWidget *widget, gpointer data);
 void gnc_invoice_window_entryDownCB (GtkWidget *widget, gpointer data);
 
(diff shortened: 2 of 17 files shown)


gjanssens:
@Bob-IT Did you ever see my e-mail reply on the list wrt to the state file comments ? I never saw a reply from you at least. This is what I asked then:

> Have you tried completely removing a group from the key file before updating 
> it ? Perhaps deleting the group will also delete its related comment.
> 
> Regards,
> 
> Geert

Bob-IT:
@gjanssens I think I missed it, but I tried that, from what I can remember the following happens
You start with a saved entry in the gcm of say this...

#Register state for "Assets:Current Assets:Savings Account"
[Register 5ca76dbe6668a75daeffd33b1229fa0e]
date_width=111...

you then start up and the gcm file loads so you get this...

#Register state for "Assets:Current Assets:Savings Account"

#Register state for "Assets:Current Assets:Savings Account"
[Register 5ca76dbe6668a75daeffd33b1229fa0e]
date_width=111...

and if you remove the group, you are left with this in the file...
#Register state for "Assets:Current Assets:Savings Account"

gnucash/ui/gnc-plugin-page-invoice-ui.xml
```diff
@@ -40,6 +40,7 @@
       <menu name="Business" action="BusinessAction">
         <placeholder name="BusinessPlaceholderMiddle">
           <menuitem name="ToolsProcessPayment"     action="ToolsProcessPaymentAction"/>
+          <menuitem name="ToolsSaveLayout"         action="ToolsSaveLayoutAction"/>
```
gjanssens: I would put this under the "Window" menu rather than the tools menu as i is window state you are manipulating here.

gjanssens: As with the register, provide a menu item as well to restore the state back to the original built-in state for consistency.

gjanssens:
> @gjanssens I think I missed it, but I tried that, from what I can remember the following happens
> You start with a saved entry in the gcm of say this...
> 
> #Register state for "Assets:Current Assets:Savings Account"
> [Register 5ca76dbe6668a75daeffd33b1229fa0e]
> date_width=111...
> 
> you then start up and the gcm file loads so you get this...
> 
> #Register state for "Assets:Current Assets:Savings Account"
> 
> #Register state for "Assets:Current Assets:Savings Account"
> [Register 5ca76dbe6668a75daeffd33b1229fa0e]
> date_width=111...
> 
That is odd. How is loading the gcm file already adding a comment ?
Are you sure this happens while loading the gcm file and not when writing ?


> and if you remove the group, you are left with this in the file...
> #Register state for "Assets:Current Assets:Savings Account"

So removing does also remove the comment that's directly related. The mystery to solve then is where that orphan comment is created.

Bob-IT:
> That is odd. How is loading the gcm file already adding a comment ?
> Are you sure this happens while loading the gcm file and not when writing ?

I will have to check this again and report back, I defiantly remember deleting the group before saving the group with a comment.
And there was me thinking not much wrong here!!

gjanssens:
Hey, not really much wrong. Your work just got me thinking a bit furhter...

gnucash/gnome/gnc-budget-view.c
```diff
@@ -716,11 +716,25 @@ gnc_budget_view_save_account_filter (GncBudgetView *budget_view)
...
-    gnc_tree_view_account_save_filter (GNC_TREE_VIEW_ACCOUNT(priv->tree_view),
-                                       priv->fd, gnc_state_get_current (),
-                                       gnc_tree_view_get_state_section (
-                                       GNC_TREE_VIEW(priv->tree_view)));
+    if (GNC_IS_BUDGET(priv->budget))
+    {
+        const gchar *name = gnc_budget_get_name (priv->budget);
+        gchar *date = gnc_print_time64 (gnc_time (NULL), qof_date_format_get_string (QOF_DATE_FORMAT_ISO));
+        GKeyFile *key_file = gnc_state_get_current();
+        gchar *group_comment = g_strdup_printf ("Budget state for \"%s\" - %s", name, date);
```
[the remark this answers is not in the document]
Bob-IT: I was thinking it would be nice to know when it was last opened but in light of other comments it would not make sense for doing so just for budgets

gnucash/gnome/dialog-invoice.c
```diff
@@ -501,29 +502,74 @@ gnc_invoice_window_help_cb (GtkWidget *widget, gpointer data)
...
+
+    if (g_key_file_has_group (key_file, group) && !use_default)
+        return group;
+    else
+    {
+        switch (gncOwnerGetType (&iw->owner))
+        {
+            case GNC_OWNER_VENDOR:
+                return g_strdup ("Bill 00000000000000000000000000000000");
+                break;
```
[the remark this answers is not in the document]
Bob-IT: I was just going on the General Register format, makes no difference so can be changed

gnucash/gnome/gnc-plugin-page-invoice.c
```diff
@@ -283,6 +290,7 @@ static action_toolbar_labels bill_action_labels[] =
     {"EditUnpostInvoiceAction", N_("_Unpost Bill")},
     {"BusinessNewInvoiceAction", N_("New _Bill")},
     {"ToolsProcessPaymentAction", N_("_Pay Bill")},
+    {"ToolsSaveLayoutAction", N_("_Save Layout for Bill Entries")},
```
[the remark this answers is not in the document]
Bob-IT: Need to check this as I did not notice.

gnucash/register/register-gnome/gnucash-register.c
```diff
@@ -161,6 +161,37 @@ gnucash_register_refresh_from_prefs (GnucashRegister *reg)
...
+
+    sheet = GNUCASH_SHEET(reg->sheet);
+
+    g_return_if_fail (sheet != NULL);
+    g_return_if_fail (GNUCASH_IS_SHEET (sheet));
+
+    current_width = sheet->window_width - 1;
+
+    widths = gnc_header_widths_new ();
+    gnucash_sheet_set_header_widths (sheet, widths);
```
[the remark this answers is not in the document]
Bob-IT: It does not drop the state as the register will be saved on close unless I have not noticed that a default register layout does not save.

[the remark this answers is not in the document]
gjanssens: Hmm, good point. Though I think that if we are currently also saving default state for each register we didn't change we definitely should stop doing that if we want to introduce a customizable default state.
Consider the following scenario assuming state is always saved per register (default or not) and we have the option to redefine a default state as well:
- open a register
- close it again. Assuming we always save state on close this will save the current default state, though we didn't really ask for it.
- now open another register, tweak the columns and save the new layout as default state
- reopen the first register. This register will still be use the original default state as that's what we saved.

However as the user never explicitly manipulated the state of that first register, s/he would probably expect it to use the new defaults.

Had we not saved the default state as register specific state when closing that first register, it would have picked up the new default and use that on reopening.

gnucash/gnome/gnc-budget-view.c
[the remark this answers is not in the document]
gjanssens: Well, to your credit you consistently added it for invoices and bills as well :-)

I'm just trying to see the use case for a date stored in a state file. This is something users normally never open. If they have to our design has failed. But I may be missing something of course, so I'm interested to hear what use you see for it?

Bob-IT:
> That is odd. How is loading the gcm file already adding a comment ?
> Are you sure this happens while loading the gcm file and not when writing ?

OK slightly remembered it wrong, the problem is with closed registers with saved data, here is what I tested. With the keep comment setting and adding a g_key_file_remove_group (state_file, state_section, NULL) after gnucash_sheet_get_header_widths and before the for loop in table-gnome.c I get this in the state file...
Open Gnucash with no registers open...
.
.
[Window1]

[Page1]

Start Gnucash, Open one register and leave open when shutting down...
.
.
[Window1]

[Page1]

#Register state for Checking Account
[Register .....71]

Reopen and close Gnucash, file same...
Reopen and close register and then Gnucash....
.
.
#Register start for Checking Account
[Register .....71]

[Window1]

[Page1]

Reopen and close Gnucash...
.
.
#Register state for Checking Account

#Register state for Checking Account
[Register ....71]

[Window1]

[Page1]

Open Gnucash, open register, close register and close Gnucash
.
.
#Register state for Checking Account

#Register state for Checking Account
[Register ....71]

[Window1]

[Page1]

Reopen Gnucash and Close it
.
.
#Register state for Checking Account

#Register state for Checking Account

[Register ....71]

[Window1]

[Page1]

As you can see, once the comment is detached from the group it can not be deleted and it just keeps growing if that register is opened and closed. The only way that seems to work is if the comment is for a key, for registers the first key is 'date_width' so maybe that is the way to go, choose the first key of the group.
If the changes to use register types is done, then we probably do not require comments for these or business types so that would leave just using for budgets.

gnucash/gnome/gnc-budget-view.c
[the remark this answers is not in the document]
Bob-IT: I was thinking that there might be a lot of state information saved for registers / budgets that have not been accessed in a while and hence one might add a routine to remove these groups based on date but with a move to using register types this would not be the case.

Bob-IT:
@gjanssens  I think I have encompassed all your ideas with this update as follows...
Open Business items save to page section, renamed the three default groups, menus actions moved to windows and updated.
Open Registers save to page section, using 5 default groups, menu actions moved to windows and registers now DO NOT save there state, it is up to the user to use the menu option to update the  default state.

On first use, if the register has state information and there is no default state for that register type then the register state is used and copied to the default state for subsequent loads and also the register state is dropped.

With these changes I did not see a reason for adding any comments so this was also removed.

gnucash/gnome/gnc-plugin-page-invoice.c
```diff
@@ -274,6 +288,13 @@ static action_toolbar_labels invoice_action_labels[] =
     {NULL, NULL},
 };
 
+static action_toolbar_labels invoice_action_layout_labels[] =
+{
+    {"WindowsSaveLayoutAction", N_("_Save Layout for Customer Documents")},
+    {"WindowsResetLayoutAction", N_("_Reset Layout for Customer Documents")},
```
gjanssens: Same here. To get `default` in there:
"Reset  Default Layout for Customer Documents"
or simply
"Reset  Default Layout"

The tooltip can do the rest.

gnucash/gnome/gnc-plugin-page-invoice.c
```diff
@@ -556,21 +651,21 @@ gnc_plugin_page_invoice_update_menus (GncPluginPage *page, gboolean is_posted, g
...
-
-    for (i = 0; tooltip_list[i].action_name; i++)
-    {
-        /* update the action tooltips */
-        action = gtk_action_group_get_action(action_group,
-                                             tooltip_list[i].action_name);
-        gtk_action_set_tooltip(action, _(tooltip_list[i].label));
-    }
+    /* update the action labels */
+    gnc_plugin_page_invoice_action_update (action_group, label_list, (void*)gtk_action_set_label);
```
gjanssens: Nice touch putting this in its own function.

gnucash/gnome/gnc-plugin-page-invoice.c
```diff
@@ -556,21 +651,21 @@ gnc_plugin_page_invoice_update_menus (GncPluginPage *page, gboolean is_posted, g
...
+    /* update the action tooltips */
+    gnc_plugin_page_invoice_action_update (action_group, tooltip_list, (void*)gtk_action_set_tooltip);
+
+    // if there is no default layout do not enable reset action
+    layout_action = gnc_plugin_page_get_action (page, "WindowsResetLayoutAction");
+    if (gnc_invoice_window_has_default_state (priv->iw))
+        has_default = TRUE;
+    gtk_action_set_sensitive (layout_action, has_default);
```
gjanssens: You're calling this code twice exactly like this. Perhaps spin it out into a helper function instead ?

gnucash/gnome/gnc-plugin-page-invoice.c
```diff
@@ -324,6 +359,12 @@ static action_toolbar_labels invoice_action_tooltips[] = {
     {NULL, NULL},
 };
 
+static action_toolbar_labels invoice_action_layout_tooltips[] = {
+    {"WindowsSaveLayoutAction", N_("Save this layout for the default Customer Document")},
+    {"WindowsResetLayoutAction", N_("Reset column widths and remove the default layout for Customer Documents")},
```
gjanssens: "Reset default layout for all customer invoices and credit notes back to built-in default and update current page accordingly"

gnucash/gnome/gnc-plugin-page-register.c
```diff
@@ -439,6 +441,18 @@ static GtkActionEntry gnc_plugin_page_register_actions [] =
...
+
+    /* Windows menu */
+    {
+        "WindowsSaveLayoutAction", NULL, "_Save Layout for this Register type", NULL,
+        "Save this layout for this register type to the default layout",
+        G_CALLBACK (gnc_plugin_page_register_cmd_save_layout)
+    },
+    {
+        "WindowsResetLayoutAction", NULL, "_Reset Layout for this Register type", NULL,
+        "Reset column widths and remove the default layout for this register type",
```
gjanssens: These strings should be marked for translation.
And like with the business options, I propose to use
"Use as Default Layout for this Register Type"
"Reset Default Layout for this Register Type"
Or more concisely
"Use as Default Layout"
"Reset Default Layout"

And tooltips:
"Use the current layout as default for all registers of this type"
"Reset default layout for all registers of this type back to built-in default and update page accordingly"

Is there a way to insert the register type in here ?

gnucash/gnome/gnc-split-reg.c
```diff
@@ -402,6 +407,57 @@ gsr_setup_table( GNCSplitReg *gsr )
...
+        case CURRENCY_REGISTER:
+        {
+            return g_strdup ("Register Default State Group4");
+            break;
+        }
+        case PORTFOLIO_LEDGER:
+        {
+            return g_strdup ("Register Default State Group5");
+            break;
+        }
```
gjanssens: What is this grouping of register types based on ?

gnucash/gnome/gnc-split-reg.c
```diff
@@ -412,22 +468,40 @@ gsr_create_table( GNCSplitReg *gsr )
...
 
     ENTER("gsr=%p", gsr);
 
+    sr = gnc_ledger_display_get_split_register (gsr->ledger);
+    default_state_section = gnc_split_reg_get_state_group (sr);
+
+    // if this is from a page recreate and no register state use those settings,
+    // register state is dropped at the end of function.
+    if (gsr->page_state_name && !g_key_file_has_group (gnc_state_get_current (), register_state_section))
+        group = g_strdup (gsr->page_state_name);
```
gjanssens: I didn't do the exercise here but I suspect you can do similar simplifications by ensuring group is always treated as a `const gchar*`

gnucash/gnome/gnc-split-reg.c
Bob-IT: I looked at gnc_split_register_set_cells and used the same switch statement.

gnucash/gnome/dialog-invoice.c
```diff
@@ -2293,6 +2346,9 @@ gnc_invoice_save_page (InvoiceWindow *iw,
         guid_to_string_buff(gncOwnerGetGUID(&iw->owner), guidstr);
         g_key_file_set_string(key_file, group_name, KEY_OWNER_GUID, guidstr);
     }
+    // save the open table layout
+    gnc_table_save_state (table, group, NULL);
+    g_free (group);
```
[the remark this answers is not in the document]
Bob-IT: OK, that makes sense

gnucash/gnome/dialog-invoice.c
```diff
@@ -2254,7 +2305,7 @@ gnc_invoice_recreate_page (GncMainWindow *window,
     g_free(tmp_string);
     g_free(owner_type);
 
-    iw = gnc_invoice_new_page (book, type, invoice, &owner, window);
+    iw = gnc_invoice_new_page (book, type, invoice, &owner, window, group_name);
```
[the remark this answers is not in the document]
Bob-IT: I think this was down to the order the functions were being called, will need to check this.

gnucash/gnome/dialog-invoice.c
```diff
@@ -501,6 +503,54 @@ gnc_invoice_window_help_cb (GtkWidget *widget, gpointer data)
     gnc_gnome_help(HF_HELP, HL_USAGE_INVOICE);
 }
 
+static gchar *
+gnc_invoice_window_get_state_group (InvoiceWindow *iw)
+{
+    switch (gncOwnerGetType (&iw->owner))
```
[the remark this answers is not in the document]
Bob-IT: I am not sure, what are jobs part off ?
Would need to check this.

gnucash/gnome/dialog-invoice.h
```diff
@@ -109,6 +109,9 @@ void gnc_invoice_window_blankCB (GtkWidget *widget, gpointer data);
 void gnc_invoice_window_duplicateCB (GtkWidget *widget, gpointer data);
 void gnc_invoice_window_payment_cb (GtkWindow *parent, gpointer data);
 void gnc_invoice_window_report_owner_cb (GtkWindow *parent, gpointer data);
+void gnc_invoice_window_save_layout_to_default_state (InvoiceWindow *iw);
+void gnc_invoice_window_reset_layout_and_clear_state (InvoiceWindow *iw);
+gboolean gnc_invoice_window_has_default_state (InvoiceWindow *iw);
```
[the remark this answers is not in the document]
Bob-IT: May be I should add 'user' in the function names along with some text.

gnucash/gnome/gnc-plugin-page-invoice.c
Bob-IT: OK, your text is much better.

gnucash/gnome/gnc-plugin-page-register.c
Bob-IT: I think that was a copy and paste error from the invoice stuff, I wondered about register type, will have a think after making other changes.

gnucash/gnome/gnc-split-reg.c
Bob-IT: Possibly, will have a look.

gnucash/gnome/dialog-invoice.c
```diff
@@ -501,6 +503,54 @@ gnc_invoice_window_help_cb (GtkWidget *widget, gpointer data)
...
+}
+
+void
+gnc_invoice_window_save_layout_to_default_state (InvoiceWindow *iw)
+{
+    Table *table = gnc_entry_ledger_get_table (iw->ledger);
+    gchar *group = gnc_invoice_window_get_state_group (iw);
+
+    gnc_table_save_state (table, group, NULL);
+    g_free (group);
```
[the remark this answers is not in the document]
jralls: That's not true.  `gnc_invoice_window_get_state_group` calls `g_strdup`, which always allocates and transfers ownership, on its returned string.  Making the variable const isn't going to get the compiler to do the free for you.

gnucash/gnome/dialog-invoice.c
```diff
@@ -2482,11 +2538,21 @@ gnc_invoice_create_page (InvoiceWindow *iw, gpointer page)
...
     {
         GtkWidget *regWidget, *frame, *window;
+        gchar *default_group = gnc_invoice_window_get_state_group (iw);
+        gchar *group;
+
+        // if this is from a page recreate, use those settings
+        if (iw->page_state_name)
+            group = g_strdup (iw->page_state_name);
+        else
+            group = g_strdup (default_group);
```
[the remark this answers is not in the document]
jralls: And leaks memory because if `group` is `const` the compiler will raise an error when you try to `free` it.

gnucash/gnome/gnc-split-reg.c
gjanssens: Ok, that seems like a good start. However just copying this over is fairly fragile. Clearly both code locations want to group account types for the same purpose. It may be better to factor this grouping out into a separate function and have it return a unique group identifier. I'm thinking of a separate enum named "SplitRegisterTypeGroup" with 5 values; one for each of the groups you define above.

I think these enum names may be fitting:
REG_TYPE_GROUP_SIMPLE (or REG_TYPE_GROUP_CURRENCY)
REG_TYPE_GROUP_APAR
REG_TYPE_GROUP_STOCK
REG_TYPE_GROUP_JOURNAL
REG_TYPE_GROUP_PORTFOLIO

With that you can also define a list of (translatable) names for these.
As a first suggestion:
- Currency account registers
- Business account registers
- Stock account registers
- Journal registers
- Portfolio registers

gnucash/gnome/dialog-invoice.c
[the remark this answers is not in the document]
gjanssens: It was one of my other remarks to have that function return a const gchar* instead as well.

gnucash/gnome/dialog-invoice.c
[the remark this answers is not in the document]
gjanssens: Removing g_free was implied. I have  not exhaustively pointed out all cases.

gnucash/gnome/dialog-invoice.c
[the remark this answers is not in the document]
jralls: That seems to have gotten lost and anyway the implication of your comments that making the values `const char*` is misleading. Not calling `g_strdup` in `gnc_invoice_window_get_state_group` is what relieves the need to free its return value and so simplifies the code.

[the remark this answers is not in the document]
gjanssens: Actually I was assuming `gnc_invoice_window_get_state_group` was written that way to accommodate for `gnc_table_save_state` that currently takes a non-const gchar* as input. That's something I pointed out in another comment as well (for line 2349 in dialog-invoice.c).
That particular comment is what started me on the const train. Unfortunately you seem to have gotten the github messages in a random order which makes that less obvious.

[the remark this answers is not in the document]
gjanssens: And curiously the comment about removing g_strdup in gnc_invoice_window_get_state_group got lost indeed. Maybe I forgot to save it. So apologies for the confusion.

gnucash/gnome/dialog-invoice.c
```diff
@@ -501,6 +503,54 @@ gnc_invoice_window_help_cb (GtkWidget *widget, gpointer data)
     gnc_gnome_help(HF_HELP, HL_USAGE_INVOICE);
 }
 
+static gchar *
+gnc_invoice_window_get_state_group (InvoiceWindow *iw)
+{
+    switch (gncOwnerGetType (&iw->owner))
+    {
+        case GNC_OWNER_VENDOR:
+            return g_strdup ("Bill Default State Group");
```
gjanssens: So to complete the whole story: these g_strdups should go if we want to make the state_group string const throughout the code.
What I *am* not sure about is whether these strings should be made static top-level strings or can be returned directly from the function here.

jralls: They have to be declared `static` somewhere to be passable, otherwise the compiler will just create a temporary and return a ptr to it, but the ptr won't be valid because the temporary will have gone out of scope. It's OK to do the static decl in the function so that the strings don't pollute the compilation-unit's name space:
```C
#include <stdio.h>
#include <stddef.h>

const char*
string_func(int which)
{
     static const char* string1 = "I'm the first string!";
     static const char* string2 = "I'm the second string!";
     static const char* string3 = "I'm the last string!";

     switch (which)
     {
     case 0:
          return string1;
     case 1:
          return string2;
     case 2:
          return string3;
     default:
          return NULL;
     }
}

int
main(int argc, char** argv)
{
     for (int i = 0; i < 3; ++i)
          printf("%s\n", string_func(i));
     return 0;
}
```

jralls: I just tried without the static decl and it still works:
```C
#include <stdio.h>
#include <stddef.h>

const char*
string_func(int which)
{
     switch (which)
     {
     case 0:
          return "I'm the first string!";
     case 1:
          return "I'm the second string!";
     case 2:
          return  "I'm the last string!";
     default:
          return NULL;
     }
}

void
print_string(const char* string)
{
     printf("%s\n", string);
}

int
main(int argc, char** argv)
{
     for (int i = 0; i < 3; ++i)
     {
          const char *string = string_func(i);
          print_string(string);
     }
     return 0;
}
```
So it seems that the compiler (clang) makes a static const char* from the literal without being told to.

gnucash/gnome/dialog-invoice.c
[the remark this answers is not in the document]
Bob-IT: I think I had it there to start with but it is too late for correct use, this is how I see it for page recreate, starts here.
gnc_invoice_recreate_page, on the line 2308, iw = ....
gnc_invoice_new_page, on 2195 I set iw->page_state_name = group_name
                                       on 2212 it calls gnc_main_window_open_page which ends up at
gnc_plugin_page_invoice_create_widget, line 742
gnc_invoice_create_page and this is where it is used, line 2545
It is not till all this is done we get back to the next line, just could not see an easier way of doing it.

gnucash/gnome/gnc-plugin-page-invoice.c
Bob-IT: Will do.

gnucash/gnome/dialog-invoice.c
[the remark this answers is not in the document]
gjanssens: You are correct. I missed the part that gnc_invoice_new_page already creates the window and hence needs the group name.

gnucash/gnome/gnc-split-reg.c
Bob-IT: OK, Have almost got this sorted, just wondering if I need both translated for the menu and an untranslated for the state group section. Not sure on the policy of having translated group sections in the state file?

gjanssens:
I would not translate in the state file, only in the gui.

Op zondag 1 maart 2020 20:55:24 CET schreef Bob-IT:
> Bob-IT commented on this pull request.
> 
> > +        case BANK_REGISTER:
> +        case CASH_REGISTER:
> +        case ASSET_REGISTER:
> +        case CREDIT_REGISTER:
> +        case LIABILITY_REGISTER:
> +        case INCOME_REGISTER:
> +        case EXPENSE_REGISTER:
> +        case EQUITY_REGISTER:
> +        case TRADING_REGISTER:
> +        {
> +            return g_strdup ("Register Default State Group1");
> +            break;
> +        }
> +        case PAYABLE_REGISTER:
> +        case RECEIVABLE_REGISTER:
> +        {
> +            return g_strdup ("Register Default State Group2");
> +            break;
> +        }
> +        case INCOME_LEDGER:
> +        case GENERAL_JOURNAL:
> +        case SEARCH_LEDGER:
> +        {
> +            return g_strdup ("Register Default State Group3");
> +            break;
> +        }
> +        case STOCK_REGISTER:
> +        case CURRENCY_REGISTER:
> +        {
> +            return g_strdup ("Register Default State Group4");
> +            break;
> +        }
> +        case PORTFOLIO_LEDGER:
> +        {
> +            return g_strdup ("Register Default State Group5");
> +            break;
> +        }
> 
> OK, Have almost got this sorted, just wondering if I need both translated
> for the menu and an untranslated for the state group section. Not sure on
> the policy of having translated group sections in the state file?

Bob-IT:
Made some more changes and hopefully they are correct.

Bob-IT:
@gjanssens I have had another look at this and as far as I can see it does what I think it should. Made all the changes for the const gchar state_section and the testing I have done results in what I am  expecting. I would like to push this on Wednesday or before if you have no objection/changes.

gnucash/gnome/gnc-split-reg.c
gjanssens: I'm sorry for my delayed response. I think the state file should use untranslated section names. Otherwise just changing language would cause gnucash to no longer recognize the sections.

gjanssens: (Oh I did reply but via mail... apparently that fails to add the reply to the proper thread)

gjanssens:
Good to go.

Bob-IT:
This has been applied to master but PR did not close.

PR #652, opened 2020-02-27
[Master] Transaction / Invoice Associations

Bob-IT:
I have been tinkering with this on and off and adding here for comment.
The main changes are...
The first few commits involve moving the existing transaction association dialog to a neutral named dialog-assoc file and moving the gnc-split-reg functions to this new file.
A couple of enhancement bugs for sorting and reloading the transaction list.
A new dialog has been added for when the path head is changed which will update the existing relative paths based on the user request to make them absolute ones.
Have changed from using a file chooser widget to a button so that they should get a native file dialog.
Added the use of glyphs for the association cell if there current font supports it.
The next few commits deal with adding the ability to add an association to the business items Invoice/Bill and Voucher and reusing the existing association dialogs. I have used a 'link button' on the plugin page below the notes field which is displayed only when there is an association. 
@gjanssens  Not sure if this is the best location but it is a start point.

@christopherlam I need help with the commit 'Bug 796531' which you commented on and added some scheme code. To fix the issue I have created a new anchor for the association so when clicked the association will open out side of the webkit window just like the existing association code. With the change to trep-engine.scm, tests 90 and 93 fail but other changes are needed, column title, display option, maybe follow 'enable links', what ever you think.


gnucash/report/trep-engine.scm
```diff
@@ -1226,6 +1226,16 @@ be excluded from periodic reporting.")
...
+               (add-if 'association-column?
+                       (vector ""
+                               (lambda (split transaction-row?)
+                                 (let ((url (xaccTransGetAssociation (xaccSplitGetParent split))))
+                                   (if (not (string-null? url))
+                                     (gnc:make-html-table-cell/markup
+                                      "text-cell"
+                                      (gnc:html-transaction-association-anchor (xaccSplitGetParent split) (_ "Association")))
+                                       " ")))))
+
```
christopherlam: That's rather impressive. Here's a follow-up commit that you can fixup to this one. Minor nitpick: Wouldn't `trans-association-guid` be more consistent?

```
@@ -937,6 +937,7 @@ be excluded from periodic reporting.")
       (list (N_ "Other Account Code")           "j"  (_ "Display the other account code?") #f)
       (list (N_ "Shares")                       "k"  (_ "Display the number of shares?") #f)
       (list (N_ "Price")                        "l"  (_ "Display the shares price?") #f)
+      (list (N_ "Association")                  "l5" (_ "Display the transaction association") #f)
       ;; note the "Amount" multichoice option in between here
       (list optname-grid                        "m5" (_ "Display a subtotal summary table. This requires Display/Amount being 'single") #f)
       (list (N_ "Running Balance")              "n"  (_ "Display a running balance?") #f)
@@ -1061,6 +1062,7 @@ be excluded from periodic reporting.")
                      (opt-val gnc:pagename-display (N_ "Other Account Name"))))
           (cons 'shares (opt-val gnc:pagename-display (N_ "Shares")))
           (cons 'price (opt-val gnc:pagename-display (N_ "Price")))
+          (cons 'association (opt-val gnc:pagename-display "Association"))
           (cons 'amount-single (eq? amount-setting 'single))
           (cons 'amount-double (eq? amount-setting 'double))
           (cons 'common-currency (opt-val gnc:pagename-general optname-common-currency))
@@ -1226,15 +1228,17 @@ be excluded from periodic reporting.")
                                   "number-cell"
                                   (xaccSplitGetAmount split)))))
 
-               (add-if 'association-column?
+               (add-if (column-uses? 'association)
                        (vector ""
                                (lambda (split transaction-row?)
-                                 (let ((url (xaccTransGetAssociation (xaccSplitGetParent split))))
-                                   (if (not (string-null? url))
-                                     (gnc:make-html-table-cell/markup
-                                      "text-cell"
-                                      (gnc:html-transaction-association-anchor (xaccSplitGetParent split) (_ "Association")))
-                                       " ")))))
+                                 (let ((url (xaccTransGetAssociation
+                                             (xaccSplitGetParent split))))
+                                   (and (not (string-null? url))
+                                        (gnc:make-html-table-cell/markup
+                                         "text-cell"
+                                         (gnc:html-transaction-association-anchor
+                                          (xaccSplitGetParent split)
+                                          (_ "📎"))))))))
 
                (add-if (column-uses? 'price)
                        (vector (_ "Price")
```

Bob-IT: @christopherlam Thanks for the update and the suggested change which I have done.

gnucash/report/trep-engine.scm
```diff
@@ -1226,6 +1228,20 @@ be excluded from periodic reporting.")
...
+                                 (let ((url (xaccTransGetAssociation
+                                             (xaccSplitGetParent split))))
+                                   (and (not (string-null? url))
+                                        (gnc:make-html-table-cell/markup
+                                         "text-cell"
+                                         (if opt-use-links?
+                                          (gnc:html-transaction-association-anchor
+                                           (xaccSplitGetParent split)
+                                           (_ "📎"))
+                                           (_ "📎"))))))))
```
christopherlam: ahem.

Bob-IT: Not sure what you mean by this?

christopherlam: There should be only one `(_ "📎"))` -- delete line 1242

christopherlam: Ah sorry didn't see your `opt-use-links?`... my bad. The indenting made it confusing.

christopherlam: This is why emacs indent *strictly* according to parentheses level as follows:
```scheme
               (add-if (column-uses? 'association)
                       (vector ""
                               (lambda (split transaction-row?)
                                 (let ((url (xaccTransGetAssociation
                                             (xaccSplitGetParent split))))
                                   (and (not (string-null? url))
                                        (gnc:make-html-table-cell/markup
                                         "text-cell"
                                         (if opt-use-links?
                                             (gnc:html-transaction-association-anchor
                                              (xaccSplitGetParent split)
                                              (_ "📎"))
                                             (_ "📎"))))))))
```

Bob-IT: There was me just about to point that out, what should the indenting be and I will change it.

christopherlam: Thinking back... if `opt-use-links?` is #FALSE then we shouldn't even display the paperclip :laughing:

Bob-IT: OK, I will change that and push later.

Bob-IT: It does show that there is an association even if they can not click on it. If the user does not want it to show like that, they just do not display the column.

gnucash/register/register-gnome/gnucash-sheet-private.c
```diff
@@ -479,6 +479,20 @@ draw_cell (GnucashSheet *sheet,
...
+    {
+        const gchar *name = gnc_table_get_cell_name (table, virt_loc);
+
+        if (g_strcmp0 (name, "associate") == 0 && g_strcmp0 (text, " ") != 0)
+        {
+            if (g_strcmp0 (text, "f") == 0) // only have two options
+                text = GLYPH_PAPERCLIP;
+            else
+                text = GLYPH_LINK;
+        }
```
gjanssens: Can `text` ever be NULL in this function or is it always one of " ", "f" or "l" ?

If it's never null it's likely to be more efficient to test `*text != ' '` and `*text == 'f'` rather than going through g_strcmp0. If it can be NULL you may have to guard the whole test with a `if (text)`

gjanssens: A bigger issue here is though that I don't think this is the spot to do this. You have added code in the generic sheet that now needs to distinguish based on the presence of a very specific column. The sheet is supposed to be column type agnostic.
Instead you should define a new cell type for the association that does this glyph replacement for you and returns either a single letter or a glyph as text when requested for it.

Bob-IT:
@gjanssens I have created an assoc cell as suggested and made appropriate changes to be used, I did wonder if the changes to AsscoCell should be added to RecnCell but then it is probably good to keep them separate.
Do not know why travis is failing on an unrelated test, 109 test-report-utilities, maybe it is my github clone as it fails on my master but not Gnucash.
Will await feedback.

christopherlam:
This is bug 797139.

jralls:
@Bob-IT Bug 797139 is fixed so if you'll rebase and force-push Travis should pass.

gnucash/register/register-core/assoccell.h
```diff
@@ -0,0 +1,99 @@
...
+
+#ifndef ASSOC_CELL_H
+#define ASSOC_CELL_H
+
+#include <glib.h>
+
+#include "basiccell.h"
+
+#define GLYPH_PAPERCLIP "\360\237\223\216"
+#define GLYPH_LINK      "\360\237\224\227"
```
jralls: Please comment these with their unicode code points:
```
#define GLYPH_PAPERCLIP "\360\237\223\216"  // Codepoint  U+1F4CE
#define GLYPH_LINK      "\360\237\224\227"         // Codepoint U+1F517
```
These should be put somewhere that the sheet and report code can get to them too so there's only one definition.

gnucash/register/register-gnome/gnucash-sheet.c
```diff
@@ -789,6 +789,29 @@ gnucash_sheet_finalize (GObject *object)
     g_object_unref (sheet->im_context);
 }
 
+#define GLYPH_PAPERCLIP "\360\237\223\216"
+#define GLYPH_LINK      "\360\237\224\227"
+
+gboolean
+gnucash_sheet_use_glyphs (GnucashSheet *sheet)
```
jralls: Can't this code be made a function of assoccell instead of sheet so that the determination of whether to use them is colocated with the code that does use them. Speaking of which, I don't see any code that actually enables using them.

gnucash/report/trep-engine.scm
```diff
@@ -1226,6 +1228,20 @@ be excluded from periodic reporting.")
...
+                                             (xaccSplitGetParent split))))
+                                   (and (not (string-null? url))
+                                        (gnc:make-html-table-cell/markup
+                                         "text-cell"
+                                         (if opt-use-links?
+                                             (gnc:html-transaction-association-anchor
+                                              (xaccSplitGetParent split)
+                                              (_ "📎"))
+                                             (_ "📎"))))))))
+
```
jralls: Did you mean to display the same character for both legs of the if? That aside, does it make sense to put the emoji-set characters in the Scheme code? Aside from it being better to have only one place where the codepoints are set so that it's easy to change them if we decide to, you can't send them like that to the html code, they need to be the corresponding html enitities. If they're going to be translated here they should be translated in the register too, and they need a translator comment explaining what they're for.

libgnucash/app-utils/gnc-ui-util.c
```diff
@@ -903,6 +903,56 @@ gnc_get_reconcile_flag_order (void)
...
+        PERR("Bad association flag\n");
+        return NULL;
+    }
+}
+
+/********************************************************************\
+ * gnc_get_association_valid_flags                                  *
+ *   return a string containing the list of associated flags        *
+ *                                                                  *
+ * Returns: the i18n'd reconciled flags string                      *
```
jralls: They're not internationalized.

libgnucash/app-utils/gnc-ui-util.h
```diff
@@ -189,6 +189,13 @@ const char * gnc_get_reconcile_str (char reconciled_flag);
 const char * gnc_get_reconcile_valid_flags (void);
 const char * gnc_get_reconcile_flag_order (void);
 
+#define WASSOC 'w'
```
bhardwajs: Can this be `constexpr char WASSOC = 'w'`? I see the use in `gnc_get_association_str` and `gnc_get_association_flag_order` where `char` would fit in directly.

gnucash/gnome-utils/dialog-assoc-utils.c
```diff
@@ -0,0 +1,454 @@
...
+    if (g_str_has_prefix (uri, "file:") && !g_str_has_prefix (uri,"file://"))
+    {
+        // fix an earlier error when storing relative paths before version 3.5
+        // relative paths are stored without a leading "/" and in native form
+        if (g_str_has_prefix (uri,"file:/") && !g_str_has_prefix (uri,"file://"))
+            part = uri + strlen ("file:/");
+        else if (g_str_has_prefix (uri,"file:") && !g_str_has_prefix (uri,"file://"))
+            part = uri + strlen ("file:");
+
+        if (part)
```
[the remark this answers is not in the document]
gjanssens: If I read this properly `part` will NULL if prefix is "file://"

gnucash/gnome-utils/dialog-assoc-utils.c
```diff
@@ -0,0 +1,454 @@
...
+        {
+            if (!xaccTransGetReadOnly (trans) && !book_ro)
+                xaccTransSetAssociation (trans, part);
+            return g_strdup (part);
+        }
+    }
+    return g_strdup (uri);
```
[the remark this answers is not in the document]
gjanssens: No. This processing is for internal storage. Absolute paths are stored as file uris, relative parts as just simple relative paths. The idea is to store in uri format as much as possible so remote uris and system local uris can be treated as much as possible in the same way. However there's not valid format for relative paths, so those are stored as just that. This makes it easier to combine the relative path with a base uri later on as well.
In previous implementations relative paths were stored as file:<relative path> or file:/<relative path>. But those are non-standard and add extra processing overhead later on so in earlier discussions we decided to move away from those.

libgnucash/app-utils/gnc-ui-util.h
gjanssens: `constexpr` is a C++11 feature. This is C code. So no, I don't think this can be used here. It's a valid improvement in a future C++ conversion though.

gnucash/gnome-utils/dialog-assoc-utils.c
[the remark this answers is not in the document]
Bob-IT: I will push my changes at the weekend. I have slightly changed this and removed the 'if (part)' and also the second parameter to the interior if statements as not required.
Basically the test is for 'file:' and 'file:/' and if found remove that part so we can store a relative path which does not start with a '/'. Paths starting 'file://' are treated as absolute paths which obviously includes 'file:///' format.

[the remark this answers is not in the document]
gjanssens: > If I read this properly `part` will NULL if prefix is "file://"

Nevermind. I'm clearly still jetlagged :(

gnucash/gnome-utils/dialog-assoc-utils.c
```diff
@@ -0,0 +1,454 @@
...
+        return NULL;
+
+    if (g_str_has_prefix (uri, "file:") && !g_str_has_prefix (uri,"file://"))
+    {
+        // fix an earlier error when storing relative paths before version 3.5
+        // relative paths are stored without a leading "/" and in native form
+        if (g_str_has_prefix (uri,"file:/") && !g_str_has_prefix (uri,"file://"))
+            part = uri + strlen ("file:/");
+        else if (g_str_has_prefix (uri,"file:") && !g_str_has_prefix (uri,"file://"))
+            part = uri + strlen ("file:");
```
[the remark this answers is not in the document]
gjanssens: IIRC I wrote the original gnc_uri_utils under the false assumption uri's would start with file:// instead of file:///. So current code other than the file associations also work under this assumption. I'm fine with cleaning that up, though that's probably best done in a separate PR.

libgnucash/app-utils/gnc-ui-util.c
Bob-IT: Copy and paste error, changed

gnucash/report/trep-engine.scm
Bob-IT: Yes, I have now changed them to be the same as the register column title.

gnucash/register/register-gnome/gnucash-sheet.c
Bob-IT: I did not think it could but having tried, yes it can so moved.

gnucash/register/register-core/assoccell.h
Bob-IT: Done.

gnucash/gnome-utils/dialog-assoc-utils.c
```diff
@@ -0,0 +1,454 @@
...
+/***********************************************************************/
+
+static gchar *
+convert_uri_to_abs_path (const gchar *path_head, const gchar *uri, gchar *uri_scheme, gboolean return_uri)
+{
+    gboolean path_head_set = FALSE;
+    gchar *file_path = NULL;
+    gchar *ret_value = NULL;
+
+    if (path_head && g_strcmp0 (path_head, "") != 0) // not default entry
```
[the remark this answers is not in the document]
Bob-IT: Changed, I tend to just use what I know, will try to remember.

gnucash/gnome-utils/dialog-assoc-utils.c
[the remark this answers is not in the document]
Bob-IT: I have changed this, the test is for 'file:' and 'file:/' with that part being removed from the string so that relative paths are stored without a scheme start and without a '/' if relative. Absolute uri paths do store with 'file:///'

gnucash/gnome-utils/dialog-assoc-utils.h
```diff
@@ -0,0 +1,38 @@
...
+ *                                                                  *
+ * Free Software Foundation           Voice:  +1-617-542-5942       *
+ * 51 Franklin Street, Fifth Floor    Fax:    +1-617-542-2652       *
+ * Boston, MA  02110-1301,  USA       gnu@gnu.org                   *
+\********************************************************************/
+
+#ifndef DIALOG_ASSOC_UTILS_H
+#define DIALOG_ASSOC_UTILS_H
+
+gchar * gnc_assoc_get_unescape_uri (const gchar *path_head, const gchar *uri, gchar *uri_scheme);
```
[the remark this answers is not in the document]
Bob-IT: Done.

gnucash/gnome-utils/dialog-assoc-utils.c
```diff
@@ -0,0 +1,454 @@
...
+    /* if book is read only, nothing to do */
+    if (book_ro)
+        return;
+
+    /*
+     *  Transactions
+     */
+    trans_hash = g_hash_table_new (g_direct_hash, g_direct_equal);
+
+    /* Get list of Accounts */
```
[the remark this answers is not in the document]
Bob-IT: Extracted transactions and invoice to separate functions.

gnucash/gnome/dialog-assoc.h
```diff
@@ -20,9 +20,14 @@
...
+#ifndef DIALOG_ASSOC_H
+#define DIALOG_ASSOC_H
 
-void gnc_trans_assoc_dialog (GtkWindow *parent);
+gchar * gnc_assoc_get_uri_dialog (GtkWindow *parent, const gchar *title, const gchar *uri);
+
+void gnc_assoc_open_uri (GtkWindow *parent, const gchar *uri);
+
+void gnc_assoc_business_dialog (GtkWindow *parent);
+void gnc_assoc_trans_dialog (GtkWindow *parent);
```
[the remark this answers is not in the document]
Bob-IT: Done.

gnucash/gnome/dialog-assoc.c
```diff
@@ -0,0 +1,1102 @@
...
+        PINFO("Open uri scheme is '%s', uri is '%s'", run_scheme, run_uri);
+
+        if (run_scheme) // make sure we have a scheme entry
+        {
+            gnc_launch_assoc (GTK_WINDOW (parent), run_uri);
+            g_free (run_uri);
+            g_free (run_scheme);
+        }
+        else
+            gnc_error_dialog (GTK_WINDOW (parent), "%s", _("This associated entry has no valid scheme."));
```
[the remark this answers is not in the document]
Bob-IT: After some more testing it is not required so removed.

gnucash/gnome/dialog-assoc.c
```diff
@@ -0,0 +1,1102 @@
...
+                     GtkTreeViewColumn  *col, gpointer user_data)
+{
+    AssocDialog   *assoc_dialog = user_data;
+    GtkTreeModel  *model = gtk_tree_view_get_model (GTK_TREE_VIEW(assoc_dialog->view));
+    GtkTreeIter    iter;
+    GncInvoice    *invoice;
+    const gchar   *uri;
+
+    if (!gtk_tree_model_get_iter (model, &iter, path))
+        return; /* path describes a non-existing row - should not happen */
```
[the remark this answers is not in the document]
Bob-IT: Changed.

gnucash/gnome-utils/dialog-assoc-utils.c
```diff
@@ -0,0 +1,454 @@
...
+static QofLogModule log_module = GNC_MOD_GUI;
+
+/***********************************************************************/
+
+static gchar *
+convert_uri_to_abs_path (const gchar *path_head, const gchar *uri, gchar *uri_scheme, gboolean return_uri)
+{
+    gboolean path_head_set = FALSE;
+    gchar *file_path = NULL;
+    gchar *ret_value = NULL;
```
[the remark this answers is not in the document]
Bob-IT: I do try to do that, this one got missed.

gnucash/gnome/dialog-assoc.c
```diff
@@ -0,0 +1,1102 @@
...
+            case GNC_INVOICE_VEND_CREDIT_NOTE:
+                inv_type = _("Bill");
+                break;
+            case GNC_INVOICE_EMPL_INVOICE:
+            case GNC_INVOICE_EMPL_CREDIT_NOTE:
+                inv_type = _("Voucher");
+                break;
+            default:
+//            case GNC_INVOICE_CUST_INVOICE:
+//            case GNC_INVOICE_CUST_CREDIT_NOTE:
```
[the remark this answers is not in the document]
Bob-IT: Changed.

gnucash/gnome/dialog-assoc.c
```diff
@@ -0,0 +1,1102 @@
...
+    for (ptr = accts; ptr; ptr = g_list_next (ptr))
+    {
+        Query  *query = qof_query_create_for (GNC_ID_SPLIT);
+        Account *acc = ptr->data;
+
+        qof_query_set_book (query, book);
+        xaccQueryAddSingleAccountMatch (query, acc, QOF_QUERY_AND);
+
+        /* Run the query */
+        for (splits = qof_query_run (query); splits; splits = splits->next)
```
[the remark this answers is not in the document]
Bob-IT: Had not used qof collections in the past so never occurred to me to use, changed.

gnucash/gnome/dialog-assoc.c
```diff
@@ -0,0 +1,1102 @@
...
+    return(TRUE);
+}
+
+/********************************************************************\
+ * gnc_assoc_business_dialog                                        *
+ * opens a window showing the Associations of all Business Items    *
+ *                                                                  *
+ * Args:   parent  - the parent of the window to be created         *
+ * Return: nothing                                                  *
+\********************************************************************/
```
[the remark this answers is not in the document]
Bob-IT: Done.

Bob-IT:
I have made the recommended changes and some others I found, bit disappointed that you found so many g_free's after checking several times, need to be more careful/observant.

gnucash/gnome/dialog-assoc.c
```diff
@@ -0,0 +1,1102 @@
...
+    if (gtk_tree_view_get_column (GTK_TREE_VIEW(assoc_dialog->view), AVAILABLE - 1) == col)
+    {
+
+        QofBook      *book = gnc_get_current_book();
+        gboolean      book_ro = qof_book_is_readonly (book);
+        gchar        *ret_uri;
+
+        /* This should never be true, but be paranoid */
+        if (!invoice)
+            return;
```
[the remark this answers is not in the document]
Bob-IT: Changed.

gnucash/gnome/dialog-assoc.c
```diff
@@ -0,0 +1,1102 @@
...
+                ret_uri = g_strdup (part);
+            }
+            else
+                ret_uri = g_strdup (dialog_uri);
+
+            PINFO("Dialog File URI: '%s'", ret_uri);
+            g_free (dialog_uri);
+        }
+    }
+    if (result == GTK_RESPONSE_REJECT)
```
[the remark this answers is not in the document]
Bob-IT: Changed.

gnucash/gnome/dialog-assoc.c
```diff
@@ -0,0 +1,1102 @@
...
+    gboolean      is_list_trans;
+}AssocDialog;
+
+/* This static indicates the debugging module that this .o belongs to. */
+static QofLogModule log_module = GNC_MOD_GUI;
+
+/***********************************************************************/
+
+static gboolean
+assoc_set_path_head_label (GtkWidget *path_head_label, const gchar *path_head)
```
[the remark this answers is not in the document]
Bob-IT: Changed.

Bob-IT:
OK I have gone through this again and I think I have sorted all the g_free errors, created a new cell type assoc which handles all the glyph parts for the assoc cell. Split some functions and changed the way the list of associations is collected. It would be nice to get this in 4.0.

gnucash/gnome/dialog-assoc.c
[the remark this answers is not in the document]
jralls: Umm, it's still here, and since it starts with `/**` Doxygen will pick it up.

gnucash/register/register-core/assoccell.h
```diff
@@ -0,0 +1,101 @@
...
+void gnc_assoc_cell_set_string_getter (AssocCell *cell,
+                                       AssocCellStringGetter getter);
+
+/** note that chars is copied into the AssocCell directly, but remains
+ * the "property" of the caller.  The caller must maintain the chars
+ * pointer, and the caller must setup a mechanism to 'free' the chars
+ * pointer.  The rationale is that you may have many AssocCell objects
+ * that use the same set of flags -- this saves you an alloc/free for
+ * each cell.  - warlord  2001-11-28
+ */
```
jralls, line 87: I think `chars` should be `flags`. The note should be preceded by `@param flags` to make it completely clear what it refers to, and the signature line is unnecessary.

jralls:
I couldn't find any missed leaks and the code generally looks OK. There are still a ton of comments in the implementation files that are going to get fed to Doxygen because the ornate frames begin with `/**`, but are otherwise not in Doxygen format. That doesn't really affect the code, just our sanity later, so I've marked it approved.

Bob-IT:
@gjanssens Do you have any concerns, would like to push this.

PR #653, opened 2020-02-29
Bug 797613  Due Invoices Reminder shows Job Name instead of Company Name

ekinonnakapito:
In the invoices due dialog, the column 'Company' should show the name of the customer/vendor of an invoice/bill.
Currently the content is set to the owner name, but if a job is assigned to the invoice, that will show the job name.
This change changes the content to the name of the owner parent, which is always  the customer/vendor name.


gjanssens:
Sweet, thanks !

PR #658, opened 2020-03-01
[Master] Invalid Imap mapping dialog

Bob-IT:
This PR takes over from #647 in presenting a dialog when the 'Import Map Editor' starts if it finds any invalid account mappings. The dialog asks the user if they would like to remove them now, if not a button is added to the editor so they can remove them later.

@christopherlam I think this does what you wanted but would be interested in feedback and any text changes.


christopherlam:
:trophy: works beautifully... that's all. Thank you!

I didn't realize we could filter via 'Map Account NOT Found' to show only these invalid mappings...

Bob-IT:
Merged.

PR #659, opened 2020-03-02
Bug 797636 - Incorrect Online account matches cannot be deleted

pkzw:
A push button with text "Delete selected matches" is added to the
"Match Online accounts with GnuCash accounts" page of the "Online
Banking Setup" tool.
Additionally, the underlying tree view is set to multiple selection
mode from single selection mode.
Pushing the button will remove the matched GnuCash account from
all selected AqBanking (Online) accounts.

### DIFF
--- a/gnucash/import-export/aqb/assistant-ab-initial.c
+++ b/gnucash/import-export/aqb/assistant-ab-initial.c
@@ -89,6 +89,7 @@ gboolean aai_key_press_event_cb(GtkWidget *widget, GdkEventKey *event, gpointer 
 
 void aai_wizard_page_prepare (GtkAssistant *assistant, gpointer user_data);
 void aai_wizard_button_clicked_cb(GtkButton *button, gpointer user_data);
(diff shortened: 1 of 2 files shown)


pkzw:
I still have the following questions with respect to the suggested changes:

1. Should this PR be included as bug fix into maint or should it go as a new feature to master? Within the related bug report [797636](https://bugs.gnucash.org/show_bug.cgi?id=797636) I have given a workaround.
2. The "Delete selected matches" button will delete the matches without a further confirmation. Is this acceptable or shall there an additiona confirmation popup.
3. What about the translation of the added button text? Do I need to trigger something?

Bob-IT:
Thank you for the fix, I think this is safe for maint as it does not have a big impact.
I think the premise for confirmation button is whether it just gets in the way and how easy is it to undo. In this case undoing is just remapping the account which is fairly easy so I do not think it needs one.
Any new translations will be picked up so there is nothing to trigger.
If no one else has any issues, I will push this tomorrow.

PR #660, opened 2020-03-05
Bug 797640 - The Reconciliation Window starting balance calculator needs to ignore splits after statement date

christopherlam:
adds a couple helper functions in account.cpp and gnc-ui-balances.c

### DIFF
--- a/libgnucash/engine/Account.cpp
+++ b/libgnucash/engine/Account.cpp
@@ -3428,6 +3428,24 @@ xaccAccountGetNoclosingBalanceAsOfDate (Account *acc, time64 date)
     return GetBalanceAsOfDate (acc, date, TRUE);
 }
 
+gnc_numeric
+xaccAccountGetReconciledBalanceAsOfDate (Account *acc, time64 date)
+{
+    gnc_numeric balance = gnc_numeric_zero();
+
+    g_return_val_if_fail(GNC_IS_ACCOUNT(acc), gnc_numeric_zero());
+
+    for (GList *node = GET_PRIVATE(acc)->splits; node; node = node->next)
+    {
+        Split *split = (Split*) node->data;
+        if ((xaccSplitGetReconcile (split) == YREC) &&
+            (xaccSplitGetDateReconciled (split) <= date))
+            balance = gnc_numeric_add_fixed (balance, xaccSplitGetAmount (split));
+    };
+
+    return balance;
+}
+
 /*
  * Originally gsr_account_present_balance in gnc-split-reg.c
  *
--- a/gnucash/gnome/window-reconcile.c
+++ b/gnucash/gnome/window-reconcile.c
@@ -281,7 +281,8 @@ recnRecalculateBalance (RecnWindow *recnData)
 
     /* update the starting balance */
     include_children = xaccAccountGetReconcileChildrenStatus(account);
-    starting = gnc_ui_account_get_reconciled_balance(account, include_children);
+    starting = gnc_ui_account_get_reconciled_balance_as_of_date
+        (account, recnData->statement_date, include_children);
     print_info = gnc_account_print_info (account, TRUE);
 
     /*
--- a/libgnucash/app-utils/gnc-ui-balances.c
+++ b/libgnucash/app-utils/gnc-ui-balances.c
@@ -179,11 +179,11 @@ gnc_ui_account_get_print_report_balance (xaccGetBalanceInCurrencyFn fn,
     return g_strdup(xaccPrintAmount(balance, print_info));
 }
 
-
-gnc_numeric
-gnc_ui_account_get_balance_as_of_date (Account *account,
-                                       time64 date,
-                                       gboolean include_children)
+static gnc_numeric
+account_get_balance_as_of_date (Account *account,
+                                time64 date,
+                                gboolean include_children,
+                                xaccGetBalanceAsOfDateFn fn)
 {
     QofBook *book = gnc_account_get_book (account);
     GNCPriceDB *pdb = gnc_pricedb_get_db (book);
@@ -194,7 +194,7 @@ gnc_ui_account_get_balance_as_of_date (Account *account,
         return gnc_numeric_zero ();
 
     currency = xaccAccountGetCommodity (account);
-    balance = xaccAccountGetBalanceAsOfDate (account, date);
+    balance = fn (account, date);
 
     if (include_children)
     {
@@ -210,7 +210,7 @@ gnc_ui_account_get_balance_as_of_date (Account *account,
 
             child = node->data;
             child_currency = xaccAccountGetCommodity (child);
-            child_balance = xaccAccountGetBalanceAsOfDate (child, date);
+            child_balance = fn (child, date);
             child_balance =
                 gnc_pricedb_convert_balance_latest_price (pdb, child_balance,
                                                           child_currency,
@@ -226,6 +226,24 @@ gnc_ui_account_get_balance_as_of_date (Account *account,
         balance = gnc_numeric_neg (balance);
 
     return balance;
+}
+
+gnc_numeric
+gnc_ui_account_get_balance_as_of_date (Account *account,
+                                       time64 date,
+                                       gboolean include_children)
+{
+    return account_get_balance_as_of_date (account, date, include_children,
+                                           xaccAccountGetBalanceAsOfDate);
+}
+
+gnc_numeric
+gnc_ui_account_get_reconciled_balance_as_of_date (Account *account,
+                                                  time64 date,
+                                                  gboolean include_children)
+{
+    return account_get_balance_as_of_date (account, date, include_children,
+                                           xaccAccountGetReconciledBalanceAsOfDate);
 }
 
 
--- a/libgnucash/app-utils/gnc-ui-balances.h
+++ b/libgnucash/app-utils/gnc-ui-balances.h
@@ -110,6 +110,10 @@ gnc_ui_account_get_print_report_balance (xaccGetBalanceInCurrencyFn fn,
 gnc_numeric gnc_ui_account_get_balance_as_of_date (Account *account,
 						   time64 date,
 						   gboolean include_children);
+gnc_numeric
+gnc_ui_account_get_reconciled_balance_as_of_date (Account *account,
+                                                  time64 date,
+                                                  gboolean include_children);
 
 /********************************************************************
  * Balance calculations related to owners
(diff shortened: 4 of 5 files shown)


libgnucash/engine/Account.cpp
```diff
@@ -3397,10 +3403,16 @@ GetBalanceAsOfDate (Account *acc, time64 date, gboolean ignclosing)
...
+            switch (baltype) {
+            case 0:
                 balance = xaccSplitGetBalance( (Split *)lp->prev->data );
+                break;
+            case 1:
+                balance = xaccSplitGetNoclosingBalance( (Split *)lp->prev->data );
+                break;
+            case 2:
+                balance = xaccSplitGetReconciledBalance( (Split *)lp->prev->data );
+            }
```
jralls: This doesn't do what you want because it operates on `date_posted` instead of `reconcile_date`.  You can't assume that a split posted before a statement date was necessarily reconciled in that statement. If it was reconciled in the next statement, as is likely, then xaccAccountRecomputeBalances will treat it as balanced and the resulting reconciled balance won't match the statement's opening balance.

christopherlam: Explained in https://bugs.gnucash.org/show_bug.cgi?id=797640#c11 why I don't think the reconcile_date matters here.

jralls: You explained why you don't understand: You're laboring under the misconception that all transactions posted in a particular month clear the bank that month.

christopherlam: Now I know `reconcile_date` means `statement_date` I've amended `xaccAccoutGetReconciledBalanceAsOfDate` to find the _last_ split whose reconciled date fits, and return its reconciled balance. If this last attempt is still not right then I am not sure whether I can pursue this further.

libgnucash/engine/Account.cpp
```diff
@@ -3397,10 +3440,16 @@ GetBalanceAsOfDate (Account *acc, time64 date, gboolean ignclosing)
...
-            else
+            switch (baltype) {
+            case BalanceType::FULL:
                 balance = xaccSplitGetBalance( (Split *)lp->prev->data );
+                break;
+            case BalanceType::NO_CLOSING:
+                balance = xaccSplitGetNoclosingBalance( (Split *)lp->prev->data );
+                break;
+            case BalanceType::RECONCILED:
+                balance = xaccSplitGetReconciledBalance( (Split *)lp->prev->data );
```
jralls: Close but not quite. The balances attached to the splits were computed with `xaccAccountRecomputeBalances` in `posted_date` order. If you're looking a couple of months back you'll have splits that were reconciled after the reconcile date of interest and so will be included in the reconciled balance computation.

jralls: This whole clause also has an off-by-one error: If trans_time <= date then the balance of that split, not of the previous one, is the balance that you eventually want to return.

libgnucash/engine/Account.cpp
```diff
@@ -3375,15 +3396,37 @@ GetBalanceAsOfDate (Account *acc, time64 date, gboolean ignclosing)
...
+    case BalanceType::RECONCILED:
+        GList *YREC_splits = NULL;
+        balance = priv->reconciled_balance;
+        for (lp = priv->splits; lp; lp = lp->next)
+        {
+            if (xaccSplitGetReconcile ((Split*)lp->data) == YREC)
+                YREC_splits = g_list_prepend (YREC_splits, lp->data);
+        }
+        YREC_splits = g_list_reverse (YREC_splits);
+        lp = g_list_sort (YREC_splits, (GCompareFunc)compare_reconciled_date);
```
jralls: I just thought of a serious speedup for this bit. No need to sort the splits or even create a new list:
```C++
             balance = gnc_numeric_zero();
             for (GList *node = priv->splits; node ; node = g_list_next(node))
                  if (xaccSplitGetReconcile((Split*)node->data) == YREC && 
                       xaccSplitGetDateReconciledSec((Split*)node->data) <= date))
                       balance = gnc_numeric_add(balance, xaccSplitGetAmount((Split*)node-data));
              return balance;
```
I think it would make more sense to simply have a separate `get_reconciled_balance_as_of_date(Account*, time64 date)` and leave `GetBalanceAsOfDate(Account*, gboolean)` alone since the logic is substantially different.

christopherlam: Done. Reverted Account.cpp commits and recreated in last commit.

libgnucash/engine/Account.cpp
```diff
@@ -3448,32 +3448,14 @@ xaccAccountGetReconciledBalanceAsOfDate (Account *acc, time64 date)
...
-    {
-        Split *split = static_cast<Split*>(node->data);
-
-        if (xaccTransGetDate (xaccSplitGetParent (split)) <= today)
-            return xaccSplitGetBalance (split);
-    }
-
-    return gnc_numeric_zero ();
+    return xaccAccountGetBalanceAsOfDate (GNC_ACCOUNT (acc),
+                                          gnc_time64_get_today_end ());
```
christopherlam: I've slipstreamed an easy refactoring optimisation. Instead of traversing whole account splitlist then backtracking, reuse function as described in comment.

libgnucash/engine/Account.cpp
```diff
@@ -3428,6 +3428,24 @@ xaccAccountGetNoclosingBalanceAsOfDate (Account *acc, time64 date)
...
+{
+    gnc_numeric balance = gnc_numeric_zero();
+
+    g_return_val_if_fail(GNC_IS_ACCOUNT(acc), gnc_numeric_zero());
+
+    for (GList *node = GET_PRIVATE(acc)->splits; node; node = node->next)
+    {
+        Split *split = (Split*) node->data;
+        if (xaccSplitGetReconcile (split) != YREC) continue;
+        if (xaccSplitGetDateReconciled (split) > date) continue;
```
jralls: Any reason not to combine these into a single `if` with `||` between the conditions?

christopherlam: No. I'll do before merge. It was just a stylistic choice, to reduce indentation. Similar to:
```scheme
(cond
 ((fail-condition-1) #f)
 ((fail-condition-2) #f)
 (else (do-actual work)))
```

christopherlam: The amended code will test for pass-conditions instead of fail-conditions instead, which is even clearer.

christopherlam:
> In addition to the inline comments, you need to somehow compute the closing reconciled balance for the statement you're examining and to provide for displaying the splits that were reconciled in the statement being checked so that the reconcile window makes sense. The reconcile window currently displays only the splits that are unreconciled.

I haven't addressed the above request at all. I *think* this reconciliation work is still incomplete:
- we can now reconcile past statements
- however the reconcile information dialog pasted as follows *will* attempt to calculate a reasonable ending balance whenever the statement date is modified, as long as the ending balance hasn't been modified by the user. I *think* this calculation is still flawed. If I set statement date to 31/01/20, the ending-balance resets to $72.00
- ditto the starting balance calculator ($126.00) in the reconcile information dialog.
image

jralls:
That's not the ending balance I was thinking of. It defaults to the last regular balance on the day given and that's probably as close to right as it can be because there's no way for GnuCash to know which transactions are reflected in the bank's statement. The user is expected to replace that default with the amount on the statement.

The ending balance I meant is the ending reconciled balance used to match the statement balance, and it will be the reconciled balance as of the end of the statement date using the new calculation from this PR.

That leaves the display of the already-reconciled transactions. ISTM the display for an old statement should show both the unreconciled transactions and the transactions previously reconciled on the statement date so that the user can review the whole statement. One might go further and show also the transactions posted before the statement date that were reconciled later with some indication of that.

fellen:
I am not shure if it is useful in this context, but actions->online actions->get balance and some file formats like MT942 contain a balance.

christopherlam:
> That's not the ending balance I was thinking of. It defaults to the last regular balance on the day given and that's probably as close to right as it can be because there's no way for GnuCash to know which transactions are reflected in the bank's statement. The user is expected to replace that default with the amount on the statement.

I just had a thought that every reconciliation involves a `statement_date` and `ending_balance`, and it *would* be possible to store these amounts somewhere which then act as a soft balance assertion and could be used to sanity-check an account via a custom report. Re-reconciliation can also retrieve these old statement_date+ending_balance pairs.

> The ending balance I meant is the ending reconciled balance used to match the statement balance, and it will be the reconciled balance as of the end of the statement date using the new calculation from this PR.

You mean the following `Ending Balance` which has not been changed at all?
image

> That leaves the display of the already-reconciled transactions. ISTM the display for an old statement should show both the unreconciled transactions and the transactions previously reconciled on the statement date so that the user can review the whole statement. One might go further and show also the transactions posted before the statement date that were reconciled later with some indication of that.

This will be rather difficult in the Reconcile Window I think, and may be best left to an augmented Reconciliation Report.

jralls:
I guess you could store the balances in the Account's KVP.

No, I mean Reconciled Balance. Ending Balance is what the user entered in the Reconcile Info dialog.

christopherlam:
Here's some further changes which fixes the suggested `ending_balance` in the reconcile info dialog. I *think* this is useful but would prefer confirmation.

```patch
modified   gnucash/gnome/window-reconcile.c
@@ -388,8 +388,8 @@ gnc_start_recn_date_changed (GtkWidget *widget, startRecnWindowData *data)
         return;
     new_date = gnc_date_edit_get_date_end (gde);
     /* get the balance for the account as of the new date */
-    new_balance = gnc_ui_account_get_balance_as_of_date (data->account, new_date,
-                  data->include_children);
+    new_balance = gnc_ui_account_get_reconciled_balance_as_of_date
+        (data->account, new_date, data->include_children);
     /* update the amount edit with the amount */
     gnc_amount_edit_set_amount (GNC_AMOUNT_EDIT (data->end_value),
                                 new_balance);
```

image

jralls:
No, that would be wrong: Consider the normal situation where you're reconciling a just-arrived statement. This would make the proposed ending balance equal to the starting balance.

The best default GnuCash can use is the cleared balance on the statement date if it's different from the starting balance, otherwise it should use the plain balance. That's the current behavior.

christopherlam:
I tried as follows but doesn't seem to produce good ending_balances. May be best wait for #667 to be completed then. https://github.com/Gnucash/gnucash/compare/maint...christopherlam:maint-reconcile-ending-balance?expand=1

jralls:
Not surprising since you didn't test for inequality to the reconciled balance and fall back to balance-as-of-date. Did you create a test account with cleared but not reconciled splits to test with?

PR #661, opened 2020-03-08
If no transactions are found in the match dialog, close the dialog up…

jeanlaroche:
…on OK

### DIFF
--- a/gnucash/import-export/import-main-matcher.c
+++ b/gnucash/import-export/import-main-matcher.c
@@ -187,6 +187,11 @@ gboolean gnc_gen_trans_list_empty(GNCImportMainMatcher *info)
     return !gtk_tree_model_get_iter_first (model, &iter);
 }
 
+void gnc_gen_trans_list_show_all(GNCImportMainMatcher *info)
+{
+    gtk_widget_show_all (GTK_WIDGET (info->main_widget));
+}
+
 void
 on_matcher_ok_clicked (GtkButton *button, GNCImportMainMatcher *info)
 {
@@ -907,7 +912,8 @@ show_account_column_toggled_cb (GtkToggleButton *togglebutton,
 GNCImportMainMatcher *gnc_gen_trans_list_new (GtkWidget *parent,
         const gchar* heading,
         gboolean all_from_same_account,
-        gint match_date_hardlimit)
+        gint match_date_hardlimit,
+        gboolean show_all)
 {
     GNCImportMainMatcher *info;
     GtkBuilder *builder;
@@ -962,7 +968,8 @@ GNCImportMainMatcher *gnc_gen_trans_list_new (GtkWidget *parent,
         gtk_label_set_text (GTK_LABEL (heading_label), heading);
 
     gnc_restore_window_size (GNC_PREFS_GROUP, GTK_WINDOW(info->main_widget), GTK_WINDOW (parent));
-    gtk_widget_show_all (GTK_WIDGET (info->main_widget));
+    if(show_all)
+        gtk_widget_show_all (GTK_WIDGET (info->main_widget));
 
     info->transaction_processed_cb = NULL;
 
--- a/gnucash/import-export/import-main-matcher.c
+++ b/gnucash/import-export/import-main-matcher.c
@@ -177,6 +177,17 @@ void gnc_gen_trans_list_delete (GNCImportMainMatcher *info)
     g_free (info);
 }
 
+gboolean gnc_gen_trans_list_empty(GNCImportMainMatcher *info) {
+    GtkTreeModel *model;
+    GtkTreeIter iter;
+    GNCImportTransInfo *trans_info;
+    
+    g_assert (info);
+    
+    model = gtk_tree_view_get_model (info->view);
+    return !gtk_tree_model_get_iter_first (model, &iter);
+}
+
 void
 on_matcher_ok_clicked (GtkButton *button, GNCImportMainMatcher *info)
 {
@@ -189,8 +200,11 @@ on_matcher_ok_clicked (GtkButton *button, GNCImportMainMatcher *info)
     /*   DEBUG ("Begin") */
 
     model = gtk_tree_view_get_model (info->view);
-    if (!gtk_tree_model_get_iter_first (model, &iter))
+    if (!gtk_tree_model_get_iter_first (model, &iter)) {
+        // No transaction, we can just close the dialog. JEAN OK CLICKED NO TRANS
+        gnc_gen_trans_list_delete (info);
         return;
+      }
 
     /* Don't run any queries and/or split sorts while processing the matcher
     results. */
--- a/gnucash/import-export/import-main-matcher.c
+++ b/gnucash/import-export/import-main-matcher.c
@@ -182,9 +182,7 @@ gboolean gnc_gen_trans_list_empty(GNCImportMainMatcher *info)
     GtkTreeModel *model;
     GtkTreeIter iter;
     GNCImportTransInfo *trans_info;
-    
     g_assert (info);
-    
     model = gtk_tree_view_get_model (info->view);
     return !gtk_tree_model_get_iter_first (model, &iter);
 }
--- a/gnucash/import-export/import-main-matcher.c
+++ b/gnucash/import-export/import-main-matcher.c
@@ -177,7 +177,8 @@ void gnc_gen_trans_list_delete (GNCImportMainMatcher *info)
     g_free (info);
 }
 
-gboolean gnc_gen_trans_list_empty(GNCImportMainMatcher *info) {
+gboolean gnc_gen_trans_list_empty(GNCImportMainMatcher *info)
+{
     GtkTreeModel *model;
     GtkTreeIter iter;
     GNCImportTransInfo *trans_info;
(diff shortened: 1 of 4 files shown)


jeanlaroche:
This fixes bug report 797042.

christopherlam:
Hi @jeanlaroche it is more useful, for posterity, to make the commit message title the same as the bug being fixed. So this one would be `Bug 797042 - "OK" button in OFX importer doesn't work if nothing is imported`.

Have you found a way to avoid showing the dialog, instead showing a `gnc_info_dialog`?

jeanlaroche:
Ah OK thanks! I'll go ahead and push a new message.
I'm not sure what you mean by "showing the dialog, instead showing a 
|gnc_info_dialog". The only thing I changed is this: when you import an 
ofx and you get no transaction in the import window, you can click OK 
instead of Cancel. This wasn't the case previously. It's a one-line fix!|

|Jean
|

Bob-IT:
I think what Chris was eluding to, if there is nothing to import, why not display a gnc_info_dialog instead of the 'Generic import transaction matcher' stating that and maybe giving the number of transactions in the import processed so the user is confident that the file was processed.

jeanlaroche:
ah yes, that makes sense, I get it now. I'll take a look at it, but can 
we at least accept the proposed change? That's already an improvement 
over what we currently have!
Jean

jeanlaroche:
If there are no transactions to import, the code now shows a small dialog stating that the OFX file was imported correctly but no new transactions were found. I don't have an easy way to get the number of transactions processed (but not imported).

gnucash/import-export/import-main-matcher.h
```diff
@@ -178,5 +178,11 @@ gboolean gnc_gen_trans_list_run (GNCImportMainMatcher *info);
  */
 GtkWidget *gnc_gen_trans_list_widget (GNCImportMainMatcher *info);
 
+/** Checks whether there are no transactions to match.
+ */
+gboolean gnc_gen_trans_list_empty(GNCImportMainMatcher *info);
+
+
+
```
jralls: Please remove the extra whitespace.

gnucash/import-export/import-main-matcher.c
```diff
@@ -189,8 +200,11 @@ on_matcher_ok_clicked (GtkButton *button, GNCImportMainMatcher *info)
     /*   DEBUG ("Begin") */
 
     model = gtk_tree_view_get_model (info->view);
-    if (!gtk_tree_model_get_iter_first (model, &iter))
+    if (!gtk_tree_model_get_iter_first (model, &iter)) {
```
jralls: Please put the `{` on its own line.  See the [GnuCash Coding Style](https://wiki.gnucash.org/wiki/CodingStandard).

gnucash/import-export/import-main-matcher.c
```diff
@@ -189,8 +200,11 @@ on_matcher_ok_clicked (GtkButton *button, GNCImportMainMatcher *info)
...
+        // No transaction, we can just close the dialog. JEAN OK CLICKED NO TRANS
```
jralls: What's that all-caps bit on the end about?

gnucash/import-export/import-main-matcher.h
```diff
@@ -178,5 +178,11 @@ gboolean gnc_gen_trans_list_run (GNCImportMainMatcher *info);
  */
 GtkWidget *gnc_gen_trans_list_widget (GNCImportMainMatcher *info);
 
+/** Checks whether there are no transactions to match.
```
jralls: Document the parameter and return as well.

gnucash/import-export/ofx/gnc-ofx-import.c
```diff
@@ -1085,7 +1086,15 @@ void gnc_file_ofx_import (GtkWindow *parent)
 #endif
 
         DEBUG("Opening selected file");
+        // JEAN: Where the ofx file is loaded.
         libofx_proc_file(libofx_context, selected_filename, AUTODETECT);
+        // Now would be a good time to see whether the view has anything in it!
+        gboolean is_empty = gnc_gen_trans_list_empty(gnc_ofx_importer_gui);
```
jralls: You can't declare variables in the middle of a block in C, that's a C++ thing.  You could move the declaration to the top of the function, but since `is_empty` is used only in the following if statement, just call `gnc_gen_trans_list_empty` as the condition.

gnucash/import-export/ofx/gnc-ofx-import.c
```diff
@@ -1085,7 +1086,15 @@ void gnc_file_ofx_import (GtkWindow *parent)
...
+        if(is_empty) {
```
jralls: Formatting, see the coding standards.

jeanlaroche:
Sorry about the messy PR. I fixed all the issues I believe. Won't happen again (hopefully!). Thanks for the feedback.

gnucash/import-export/import-main-matcher.c
```diff
@@ -177,6 +177,18 @@ void gnc_gen_trans_list_delete (GNCImportMainMatcher *info)
...
 }
 
+gboolean gnc_gen_trans_list_empty(GNCImportMainMatcher *info)
+{
+    GtkTreeModel *model;
+    GtkTreeIter iter;
+    GNCImportTransInfo *trans_info;
+    
+    g_assert (info);
+
```
christopherlam: We want to remove whitespace wherever possible

gnucash/import-export/ofx/gnc-ofx-import.c
```diff
@@ -1086,6 +1086,12 @@ void gnc_file_ofx_import (GtkWindow *parent)
 
         DEBUG("Opening selected file");
         libofx_proc_file(libofx_context, selected_filename, AUTODETECT);
+        // Now would be a good time to see whether the view has anything in it!
+        if(gnc_gen_trans_list_empty(gnc_ofx_importer_gui))
+        {
+            gnc_gen_trans_list_delete (gnc_ofx_importer_gui);
+            gnc_info_dialog(parent,_("OFX file imported, no new transactions"));
+        }
```
christopherlam: I really wonder if this is the right location to test for empty view.

1. The UI will flash the importer before showing the no-transaction info dialog
2. (I may be wrong here) The dialog is only shown on OFX import; other importers could use it. Do you think you can replicate functionality in `import-export`?

gnucash/import-export/import-main-matcher.c
jeanlaroche: Sure. I'm only replicating code that I see around, in the hope that I replicate coding standards. See on_matcher_ok_clicked() just below, which I did not touch.

gnucash/import-export/ofx/gnc-ofx-import.c
jeanlaroche: Well you can't test until all the transactions have been processed, which occurs one line above:
```
libofx_proc_file(libofx_context, selected_filename, AUTODETECT);
```
I'm not seeing the UI flash. Does this command (a few lines above)
```
gnc_ofx_importer_gui = gnc_gen_trans_list_new (GTK_WIDGET(parent), NULL, FALSE, 42);
```
Causes the window to appear? Or simply creates it without showing it.
About 2) yes, I can try to replicate it in import-export. Let me take a look.

jeanlaroche: mmm, I looked, and there isn't a whole lot of code reuse in import-export, I think I'd rather not touch anything.

Bob-IT: It does create it and show it, if you look at that function, near the bottom is a gtk_widget_show_all. I would think you could remove that line, create a new function in import_main_matcher to retrieve the main_widget and then add an else clause to your if statement to do the gtk_widget_show_all but you will need to find any other occurrence of gnc_gen_trans_list_new and make appropriate changes and this would be a separate commit on this PR.

Bob-IT:
If you did want to show the number of transaction processed, I would imagine that you could track the number of times ofx_proc_transaction_cb is called and present the number minus the number of transactions in the list, whether it is worth it I am not sure.

jeanlaroche:
> 
> 
> If you did want to show the number of transaction processed, I would imagine that you could track the number of times ofx_proc_transaction_cb is called and present the number minus the number of transactions in the list, whether it is worth it I am not sure.

True! that might be more informative. I'll code that up.

gnucash/import-export/ofx/gnc-ofx-import.c
jeanlaroche: Ah thanks. Normally I would add an optional input parameter with a default to False, and use that, which would ensure back-compatibility but this is C code so that's not an option. Let me see how much I would have to change if I added that parameter.

jeanlaroche:
Implemented display of the number of transactions processed. This required using a static gint. DOn't know how you guys feel about that.
Also implemented the idea of not showing the match dialog until we know for sure that there are transactions to match.

jralls:
It's single-threaded code so the `static` is harmless.

jeanlaroche:
> 
> 
> It's single-threaded code so the `static` is harmless.

Some people object to it on a coding-style basis! :)

jralls:
> Some people object to it on a coding-style basis! :)

Justifiably, they're a pretty bad code smell. But `gnc-ofx-import.c` already has 5 and two of them are even more egregious as they're not marked `static`. At least they're not exposed in the header so no other files can see them, but still...

It's a bit beyond scope for this PR but replacing the globals with a struct that includes the parent widget ptr and passing the struct to ofx_set_transaction_cb would be a big improvement, as would refactoring `ofx_proc_transaction_cb`. 500 lines is a lot of function to hold in one's head.

jralls:
Still outside the scope of the PR but it would be still better to collect a list of transactions in that struct and pass the list to the `gnc_ofx_importer_gui` instead of calling `gnc_gen_trans_list_add_trans`.

jralls:
Thanks.

PR #662, opened 2020-03-09
Remove unwanted space from xmlns sodipodi URI.

shastry:
SVG icon doesn't open with latest librsvg (2.48) due to bad URI.

XML parse error: error code=99 (3) in (null):10:72: xmlns:sodipodi: 'http://inkscape.sourceforge.net/DTD/s odipodi-0.dtd' is not a valid URI


fellen:
Merged, thanks!

I believe, we should add a test 'xmllin ...'

PR #663, opened 2020-03-14
clean up some account.cpp funcs

christopherlam:
These look like yak-shaving, but IMHO clarifies and simplifies them greatly. I believe these are well tested.

### DIFF
--- a/libgnucash/engine/Account.cpp
+++ b/libgnucash/engine/Account.cpp
@@ -3364,56 +3364,27 @@ GetBalanceAsOfDate (Account *acc, time64 date, gboolean ignclosing)
      * xaccAccountForEachTransaction by using gpointer return
      * values rather than gints.
      */
-    AccountPrivate *priv;
-    GList   *lp;
-    gboolean found = FALSE;
-    gnc_numeric balance;
+    Split *latest = nullptr;
 
     g_return_val_if_fail(GNC_IS_ACCOUNT(acc), gnc_numeric_zero());
 
     xaccAccountSortSplits (acc, TRUE); /* just in case, normally a noop */
     xaccAccountRecomputeBalance (acc); /* just in case, normally a noop */
 
-    priv = GET_PRIVATE(acc);
+    for (GList *lp = GET_PRIVATE(acc)->splits; lp; lp = lp->next)
+    {
+        if (xaccTransGetDate (xaccSplitGetParent ((Split *)lp->data)) >= date)
+            break;
+        latest = (Split *)lp->data;
+    }
+
+    if (!latest)
+        return gnc_numeric_zero();
+
     if (ignclosing)
-        balance = priv->noclosing_balance;
+        return xaccSplitGetNoclosingBalance (latest);
     else
-        balance = priv->balance;
-
-    lp = priv->splits;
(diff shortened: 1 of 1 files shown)


libgnucash/engine/Account.cpp
```diff
@@ -2170,16 +2170,16 @@ xaccAccountRecomputeBalance (Account * acc)
 
         balance = gnc_numeric_add_fixed(balance, amt);
 
-        if (NREC != split->reconciled)
+        switch (split->reconciled)
         {
+        case FREC:
```
bhardwajs: Consider using `[[fallthrough]]`. It should be no-op in C++ 14 and a valuable compiler check in C++ 17.

christopherlam: Thanks!

christopherlam: Unfortunately it fails on old travis.

christopherlam: https://travis-ci.org/github/christopherlam/gnucash/jobs/662362784

libgnucash/engine/Account.cpp
```diff
@@ -3430,32 +3401,14 @@ xaccAccountGetNoclosingBalanceAsOfDate (Account *acc, time64 date)
...
-    {
-        Split *split = static_cast<Split*>(node->data);
-
-        if (xaccTransGetDate (xaccSplitGetParent (split)) <= today)
-            return xaccSplitGetBalance (split);
-    }
-
-    return gnc_numeric_zero ();
+    return xaccAccountGetBalanceAsOfDate (GNC_ACCOUNT (acc),
+                                          gnc_time64_get_today_end ());
```
jralls: This is used exactly once each in `gnucash/gnome/gnc-split-reg.c` and its near-copy `gnucash/gnome/gnc-split-reg2.c` and in its unit test in utest-Account.cpp. Why not just remove it?

libgnucash/engine/Account.cpp
```diff
@@ -2170,16 +2170,16 @@ xaccAccountRecomputeBalance (Account * acc)
...
             cleared_balance = gnc_numeric_add_fixed(cleared_balance, amt);
-        }
-
-        if (YREC == split->reconciled ||
-                FREC == split->reconciled)
-        {
-            reconciled_balance =
-                gnc_numeric_add_fixed(reconciled_balance, amt);
+        case NREC:
+            break;
```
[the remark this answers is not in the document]
christopherlam: Will remove. I'd recently about https://en.wikipedia.org/wiki/Duff%27s_device and was misled.

libgnucash/engine/Account.cpp
christopherlam: Don't understand. xaccAccountGetPresentBalance is used by `gnc-split-reg[-2].c` as a function being passed as a parameter. Can't remove it.

jralls: Crud. Those callbacks should be static functions inside gnc-split-reg.c instead of increasing the public API. Oh, well, it's outside of the scope of this PR.

jralls:
Looks good.

PR #664, opened 2020-03-14
Implement type-ahead account name completion

jeanlaroche:
Implement type-ahead account name completion


christopherlam:
Please see https://wiki.gnucash.org/wiki/CodingStandard -- in no particular order:

- no personal name stamps
- { } indenting style and space after function name
- smaller commits, e.g. one to introduce pref, another for GUI, another for the actual code.
- implement sanity checks for function arguments `g_return_if_fail`
- no code samples in comments
- don't use g_print in code, use DEBUG ("..etc..")
- no TODO

From my POV I think the code looks good, but doesn't follow house style. A more experienced dev will be able to offer more pertinent feedback.

jeanlaroche:
> Please see https://wiki.gnucash.org/wiki/CodingStandard -- in no particular order:
> 
>     * no personal name stamps
> 
>     * { } indenting style and space after function name
> 
>     * smaller commits, e.g. one to introduce pref, another for GUI, another for the actual code.
> 
>     * implement sanity checks for function arguments `g_return_if_fail`
> 
>     * no code samples in comments
> 
>     * don't use g_print in code, use DEBUG ("..etc..")
> 
>     * no TODO
> 
> 
> From my POV I think the code looks good, but doesn't follow house style. A more experienced dev will be able to offer more pertinent feedback.

Sorry about that. I'm fully aware that there are quite a few things that need to be cleaned up. I wanted to get some feedback on the current state of the code.
On other projects I worked on, it was crucial to squash all commits into 1 before the pull request. i should have checked with the coding standards.

gnucash/register/register-gnome/combocell-gnome.c
```diff
@@ -87,28 +87,43 @@ static void gnc_combo_cell_destroy (BasicCell *bcell);
...
+}
+
 static gpointer
 gnc_combo_cell_autopop_init (gpointer unused)
 {
     gulong id;
     auto_pop_combos = gnc_prefs_get_bool (GNC_PREFS_GROUP_GENERAL_REGISTER,
                                           GNC_PREF_AUTO_RAISE_LISTS);
-
-    id = gnc_prefs_register_cb (GNC_PREFS_GROUP_GENERAL_REGISTER,
```
Bob-IT: Why have you changed this, with your changes there is now no way to remove the auto raise callback, you will probably have to do some thing similar to remove your callback unless I have missed something.

gnucash/register/register-gnome/combocell-gnome.c
```diff
@@ -546,25 +566,90 @@ gnc_combo_cell_modify_verify (BasicCell *_cell,
...
     match_str = gnc_quickfill_string (match);
+    the_store = cell->shared_store_full;
 
-    if ((match == NULL) || (match_str == NULL))
+    g_debug ("type ahead search flag: %d val %s\n", type_ahead_search,newval);
+    if (type_ahead_search)
+    {
+        // Type ahead account name match code.
+        gchar *str_data;
+        gchar *case_normalized_newval;
```
Bob-IT: You need to set this to NULL.

Bob-IT:
I am not sure the g_debug statements in combocell-gnome are required, normally we would use the macro DEBUG() but sometimes it is not warranted.
I found it hard to work out what is going on, probably down to me but I would expect the following...
After making changes, rebase and squash to reduce the clutter.
If are going to change the store name, make all the changes in one commit or possibly change your new store name instead.
Pay attention to where you are / not adding spaces, it makes it more readable for me any way.

I did give it a try and seems to work but one thing I did notice is that the list does not update when you backspace which I think it should.

gnucash/gtkbuilder/dialog-preferences.glade
```diff
@@ -2021,6 +2021,25 @@ many months before the current month:</property>
...
                 </child>
+                <child>
+                  <object class="GtkCheckButton" id="pref/general.register/type-ahead-search">
+                    <property name="label" translatable="yes">_Use type ahead</property>
+                    <property name="visible">True</property>
+                    <property name="can_focus">True</property>
+                    <property name="receives_default">False</property>
+                    <property name="has_tooltip">True</property>
+                    <property name="tooltip_markup">Use the type-ahead search for accounts</property>
+                    <property name="tooltip_text" translatable="yes">Use the type-ahead search for accounts</property>
```
fellen: I know, most tooltips are currently of this kind, but better would be a short explanation, which a new user would understand:
"The normal behaviour is ... \n
Type ahead search allows to ..."

gnucash/register/register-gnome/combocell-gnome.c
jeanlaroche: Crap. i think that's a mistake. I don't know how it happened. Sorry about that. I'll revert that change.

jeanlaroche:
> 
> 
> I am not sure the g_debug statements in combocell-gnome are required, normally we would use the macro DEBUG() but sometimes it is not warranted.
> I found it hard to work out what is going on, probably down to me but I would expect the following...
> After making changes, rebase and squash to reduce the clutter.
> If are going to change the store name, make all the changes in one commit or possibly change your new store name instead.
> Pay attention to where you are / not adding spaces, it makes it more readable for me any way.
> 
> I did give it a try and seems to work but one thing I did notice is that the list does not update when you backspace which I think it should.

- I can remove the debug statement, I don't think it's needed.
- The point of the new store was precisely to *not* rename the existing one and ensure that all the code that relies on it works the same as ever.
- Understood about spaces. I need to be be more careful about that.
- Yes, the list does not update when you backspace, but that's the same behavior as the current search, so I didn't try to change that (I'm trying to minimize the impact of my changes, to minimize the chances of messing something up).

gnucash/gtkbuilder/dialog-preferences.glade
jeanlaroche: 100% agreed. I'll improve it, thanks for the suggestion.

gnucash/gnome-utils/account-quickfill.c
```diff
@@ -408,6 +442,22 @@ listen_for_account_events (QofInstance *entity, QofEventId event_type,
         DEBUG("other %s", name);
         break;
     }
+    // Now that qfb->list_store_full has been updated, qfb->list_store also needs to be updated in
+    // case we're using the regular search.
```
jralls: We prefer to use `/* ... */` for multiline comments. And lest you object that it makes commenting out code hard, we prefer `#if 0 ... #endif` for that.

gnucash/gnome-utils/account-quickfill.c
```diff
@@ -249,6 +279,8 @@ listen_for_account_events (QofInstance *entity, QofEventId event_type,
     find_data data = { 0 };
     GtkTreePath *path;
     GList *tmp;
+    gboolean valid;
+
```
jralls: Please remove the extra blank line.

gnucash/gnome-utils/account-quickfill.c
```diff
@@ -176,6 +187,8 @@ build_shared_quickfill (QofBook *book, Account *root, const char * key,
     qfb->load_list_store = TRUE;
     qfb->list_store = gtk_list_store_new (NUM_ACCOUNT_COLUMNS,
                                           G_TYPE_STRING, G_TYPE_POINTER);
+    qfb->list_store_full = gtk_list_store_new (NUM_ACCOUNT_COLUMNS,
+                                          G_TYPE_STRING, G_TYPE_POINTER);
```
jralls: Please fix the indentation so that the parameters line up throughout your changes.

gnucash/register/register-gnome/combocell-gnome.c
```diff
@@ -534,37 +555,99 @@ gnc_combo_cell_modify_verify (BasicCell *_cell,
...
 
     /* If we are inserting in the middle, just accept */
-    if (*cursor_position < _cell->value_chars)
+    if (*cursor_position < _cell->value_chars && !type_ahead_search)
     {
         gnc_basic_cell_set_value_internal (_cell, newval);
         return;
     }
-
+    // Legacy account name match.
```
jralls: This is confusing because it seems to introduce a large block of mostly new code. What part is "legacy"?

gnucash/register/register-gnome/combocell-gnome.c
```diff
@@ -534,37 +555,99 @@ gnc_combo_cell_modify_verify (BasicCell *_cell,
     }
```
jralls: Nearly everything from here on is wrapped in a `type_ahead_search` conditional. I think it would be clearer to instead have two helper functions, one for type-ahead and one not.

jeanlaroche: agreed. I'll address your other comments in the updated code. Thanks.

jeanlaroche:
If I didn't forget anything all your suggestions have been adopted. I moved the search in a separate function and also simplified the rest of the code (something I was previously reluctant to do, for fear of breaking stuff). LMK what you think.

gnucash/register/register-gnome/combocell-gnome.c
```diff
@@ -501,6 +519,62 @@ gnc_combo_cell_set_value (ComboCell *cell, const char *str)
     gnc_basic_cell_set_value (&cell->cell, str);
 }
 
+// This function looks through full_store for a partial match with newval. Note: match_str needs to be freed upon return.
+static void gnc_combo_cell_type_ahead_search(const gchar* newval, GtkListStore* full_store, PopBox *box, char **match_str)
```
jeanlaroche: I think I need to make that new function return a char* rather than use a char** as an input. It will be a lot more in-line with the coding style.

gnucash/register/register-gnome/combocell-gnome.c
```diff
@@ -501,6 +519,57 @@ gnc_combo_cell_set_value (ComboCell *cell, const char *str)
     gnc_basic_cell_set_value (&cell->cell, str);
 }
 
+/* This function looks through full_store for a partial match with newval and returns the first
+ match (which must be subsequently freed). It fills out box->item_list with found matches */
+static gchar* gnc_combo_cell_type_ahead_search(const gchar* newval, GtkListStore* full_store, PopBox *box)
```
jralls: Decls get everything on one line, but impls should have the storage class and return type on one line with the function name on a second line. Aside from style consistency it makes it easy to grep for the implementation, e.g.
`grep -r ^gnc_combo_cell_type_ahead_search *gnucash`.

gnucash/register/register-gnome/combocell-gnome.c
jralls: Looks like the indentation got messed up when you put it back.

gnucash/register/register-gnome/combocell-gnome.c
jeanlaroche: ok I see, never thought of that.

gnucash/register/register-gnome/combocell-gnome.c
```diff
@@ -501,6 +519,57 @@ gnc_combo_cell_set_value (ComboCell *cell, const char *str)
...
+    gchar *match_str = NULL;
+
+    if (normalized_newval)
+        case_normalized_newval = g_utf8_casefold (normalized_newval, -1);
+    valid = gtk_tree_model_get_iter_first (GTK_TREE_MODEL(full_store), &iter);
+    // Clear result list.
+    gnc_item_list_clear(box->item_list);
+    while (valid)
+    {
+        static gint MAX_NUM_MATCHES = 30;
```
jralls: `static const gint...`

jralls:
Looks a lot better, mostly formatting issues left.

gnucash/register/register-gnome/combocell-gnome.c
```diff
@@ -606,56 +601,42 @@ gnc_combo_cell_modify_verify (BasicCell *_cell,
...
-        ss2 = newval_chars;
-        es2 = -1;
-        cp2 += change_chars;
-        // Update the string in the type-in box with the matched string.
-        box_string2 = match_str2;
-    }
-    
-    if(match_str2 != NULL)
+    // Try the start-of-name match using the quickfill
+    QuickFill *match = NULL;
```
jralls: C++ must be deeply ingrained in your fingers, you keep getting bitten by declaring variables in the middle. Did it compile without error on Xcode?

jeanlaroche: Ah crap, sorry about that. Yes, it compiles fine on Xcode. The check in is temporary of course, since i will remove the preference when I'm done, but I can't believe I let this one slip through. Do you know if there's a way for travis to alert you if your build fails?

jralls: I think you need to set up travis on your own repo to get notifications. I wonder if Xcode is compiling it as c++ because of something internal or because cmake is telling it to.

jeanlaroche:
One more iteration. The type-ahead search now supports ":" If typing ":" does not autocomplete in the old-style search (for example, because you're already doing a type-ahead search) then ":" is replaced by ".*:.*" in the typed string and the match is done using a regular expression.
For example, if you type "aut:pense" you'd match "Auto:Expense" and "aut:ten:ange" would match "Auto:Maintenance:Oilchange". 
Let me know what you think. I find it pretty intuitive because it mirrors what was happening previously, but in a more flexible way.

gnucash/register/register-gnome/combocell-gnome.c
```diff
@@ -501,6 +519,59 @@ gnc_combo_cell_set_value (ComboCell *cell, const char *str)
...
+    gchar *match_str = NULL;
+    GError *gerror = NULL;
+    GMatchInfo *match_info = NULL;
+    GRegex *regex = NULL;
+    gchar *rep_str = NULL;
+    GRegex *regex0 = g_regex_new (":", 0, 0, &gerror);
+
+    // Replace ":" in newval with ".*:.*" so we can use regexp to match.
+    rep_str = g_regex_replace (regex0,newval,-1,0,".*:.*",0,&gerror);
+    // Then compile the regular expression based on rep_str.
```
jralls: Replace the hard-coded `:` with the return value of `gnc_get_acount_separator_string()`. Users may set a different separator character (which thanks to UTF8 might be a string as far as C is concerned).

jralls:
I think conceptually it's just right.

Don't forget to remove the type-ahead preference code and please do clean up the white space. In particular, commas should be followed with a space and second-line parameters should line up with ones above. Note on the latter that using tabs instead of spaces messes up github, so the alignment might look right in Xcode but won't here.

gnucash/register/register-gnome/combocell-gnome.c
jeanlaroche: Holy crap, thanks for reminding me, I completely forgot about that.

jeanlaroche:
> 
> 
> I think conceptually it's just right.
> 
> Don't forget to remove the type-ahead preference code and please do clean up the white space. In particular, commas should be followed with a space and second-line parameters should line up with ones above. Note on the latter that using tabs instead of spaces messes up github, so the alignment might look right in Xcode but won't here.

On some projets, people use clang-format to ensure that the right style is respected. We don't have anything similar here, right? (it's not very easy to not miss *any* formatting mishaps, that's a perfect task for an automated tool).

jralls:
We've used [astyle](http://astyle.sourceforge.net/) for that in the past:
```
astyle -xd -k1 -m0 -M60 -xL -xC79 -OHpUcZns4 --brackets=break
```
In order to make sure that your new work isn't buried in a pile of whitespace adjustments it's a good idea to run it on the file without your changes and commit the whitespace adjustments, then merge your changes onto the fixed up files and run `astyle` again.

We used to have a developer who liked to run `astyle` on the whole codebase every 6 months or so. That makes a mess of history and history in git is a lot more useful than it was in subversion so we prefer  not to do that anymore.

 We don't have any clang tooling set up because clang is installed and used by default only on MacOS.

jeanlaroche:
I squashed a few changes, removed the prefs. I then added one last commit after running astyle, was easier than doing it the other way around.

jralls:
Great, thanks for sticking with it.

christopherlam:
https://www.reddit.com/r/GnuCash/comments/hi7p91/gnucash_40/ they like it...

jeanlaroche:
> 
> 
> https://www.reddit.com/r/GnuCash/comments/hi7p91/gnucash_40/ they like it...

Thanks for linking this here, always fun to hear positive feedback! :) "I really thought development had stopped. I'm glad to see it's continuing." <- this made me laugh, with all the work you guys continuously put in! ah ah ah

gjanssens:
Op dinsdag 30 juni 2020 17:44:04 CEST schreef jean laroche:
> > https://www.reddit.com/r/GnuCash/comments/hi7p91/gnucash_40/ 
they like
> > it...
> Thanks for linking this here, always fun to hear positive feedback! :) "I
> really thought development had stopped. I'm glad to see it's 
continuing."
> <- this made me laugh, with all the work you guys continuously put in! 
ah
> ah ah

It does show our marketing department is a bit underpowered...

Geert

christopherlam:
I have to say I like both the supercharged account selector in the register; also have recently used the import matcher account selector for the first time.

Is it hard(TM) to make the import matcher account selector show only the matched accounts as well? Then Up/Down to pick the right one?

jeanlaroche:
I thought I would try to replicate it, but it's going to have to be a 
deep dive. The current selector (in the import matcher) uses the default 
TreeView search thing.
I like the new type-ahead much better... but that wasn't originally 
written in a modular way (before I added the type ahead) so not so easy 
to re-use somewhere else.

Jean

christopherlam:
> 
> I thought I would try to replicate it, but it's going to have to be a deep dive. The current selector (in the import matcher) uses the default TreeView search thing. I like the new type-ahead much better... but that wasn't originally written in a modular way (before I added the type ahead) so not so easy to re-use somewhere else. Jean

Hmm may be best leave as it is... there are advantages to _not_ filter the list early: we can _esc_ and highlight a non-typeahead account.

PR #665, opened 2020-03-14
17136391-ofx-file-import When selecting a file for OFX import, restri…

jeanlaroche:
…ct the files listed to .OFX (or .QFX)

### DIFF
--- a/gnucash/import-export/ofx/gnc-ofx-import.c
+++ b/gnucash/import-export/ofx/gnc-ofx-import.c
@@ -1020,6 +1020,7 @@ double ofx_get_investment_amount(const struct OfxTransactionData* data)
     }
 }
 
+// Let the user selecct a ofx qfx file and start the import
 void gnc_file_ofx_import (GtkWindow *parent)
 {
     extern int ofx_PARSER_msg;
@@ -1031,6 +1032,8 @@ void gnc_file_ofx_import (GtkWindow *parent)
     char *selected_filename;
     char *default_dir;
     LibofxContextPtr libofx_context = libofx_get_new_context();
+    GList *filters = NULL;
+    GtkFileFilter* filter = gtk_file_filter_new ();
 
     ofx_PARSER_msg = false;
     ofx_DEBUG_msg = false;
@@ -1042,9 +1045,13 @@ void gnc_file_ofx_import (GtkWindow *parent)
     DEBUG("gnc_file_ofx_import(): Begin...\n");
 
     default_dir = gnc_get_default_directory(GNC_PREFS_GROUP);
+    gtk_file_filter_set_name (filter, "ofx/qfx files (*.ofx,*.qfx)");
+    gtk_file_filter_add_pattern (filter, "*.[oqOQ][fF][xX]");
+    filters = g_list_prepend( filters, filter );
+
     selected_filename = gnc_file_dialog(parent,
                                         _("Select an OFX/QFX file to process"),
-                                        NULL,
+                                        filters,
                                         default_dir,
                                         GNC_FILE_DIALOG_IMPORT);
     g_free(default_dir);


gnucash/import-export/ofx/gnc-ofx-import.c
```diff
@@ -1042,9 +1043,15 @@ void gnc_file_ofx_import (GtkWindow *parent)
     DEBUG("gnc_file_ofx_import(): Begin...\n");
 
     default_dir = gnc_get_default_directory(GNC_PREFS_GROUP);
+    GList *filters = NULL;
+    GtkFileFilter* filter = gtk_file_filter_new ();
+    gtk_file_filter_set_name (filter, "ofx/qfx files (*.ofx,*.qfx)");
+    gtk_file_filter_add_pattern (filter, "*.[oqOQ][fF][xX]");
+    filters = g_list_append( filters, filter );
```
christopherlam: It doesn't matter here; but as a matter of habit we'd want to use `g_list_prepend`. See https://developer.gnome.org/glib/stable/glib-Doubly-Linked-Lists.html#g-list-append

christopherlam:
I think this is a good usability fix, but can be merged in `maint` instead whose next release is in 2 weeks? `master` will not be ready for the next few months.

jralls:
@jeanlaroche please attach a link to the bug, I can't find anything like it in the bug tracker and the number has too many digits.

The benefits are debatable as well. On the one hand if one isn't in the habit of cleaning up one's download folder highlighting the files with [oq]fx extensions might be a help. On the other banks can't be trusted to put one of those extensions on the downloaded file, so with this filter the user would have to rename the file before they could import it.

gnucash/import-export/ofx/gnc-ofx-import.c
jeanlaroche: OK I'll change it for good practice

jeanlaroche:
> 
> 
> I think this is a good usability fix, but can be merged in `maint` instead whose next release is in 2 weeks? `master` will not be ready for the next few months.
Sure. I'll do it in maint. I was under the impression that "new features" had to be in master. I would much rather have it in maint if it's going to be available faster.

jeanlaroche:
> 
> 
> @jeanlaroche please attach a link to the bug, I can't find anything like it in the bug tracker and the number has too many digits.
> 
> The benefits are debatable as well. On the one hand if one isn't in the habit of cleaning up one's download folder highlighting the files with [oq]fx extensions might be a help. On the other banks can't be trusted to put one of those extensions on the downloaded file, so with this filter the user would have to rename the file before they could import it.

There's no bug report for that. It was a suggestion, whose number I attached to the commit message. Yes, it's debatable, we use the filter for QIF files so it seemed logical to do it for these files. Of course, the filter can be disabled: you don't have to rename your file, there's an option in the open dialog for "all files"... But that's one extra requirement for user.
Ideally, the open dialog would remember your choice (filter or no filter). Not sure if that's possible.

jralls:
We use the number-summary format only for bug fixes,  not for uservoice suggestions. Put a link to it in the body of the commit message.

Also very few of the user voice suggestions are good ideas so unless @gjanssens, @fellen, or I have commented positively on one (and I don't remember ever doing so) I suggest that you discuss them on gnucash-devel before starting work.

For disabling the filter, OK.

gnucash/import-export/ofx/gnc-ofx-import.c
```diff
@@ -1042,9 +1043,15 @@ void gnc_file_ofx_import (GtkWindow *parent)
     DEBUG("gnc_file_ofx_import(): Begin...\n");
 
     default_dir = gnc_get_default_directory(GNC_PREFS_GROUP);
+    GList *filters = NULL;
+    GtkFileFilter* filter = gtk_file_filter_new ();
```
jralls: This is a C file and declaring variables after code isn't legal in C. Move these lines to the declarations block above.

jeanlaroche:
Closing this and moving to maint https://github.com/Gnucash/gnucash/pull/666

jeanlaroche:
> We use the number-summary format only for bug fixes, not for uservoice suggestions. Put a link to it in the body of the commit message.
> 
> Also very few of the user voice suggestions are good ideas so unless @gjanssens, @fellen, or I have commented positively on one (and I don't remember ever doing so) I suggest that you discuss them on gnucash-devel before starting work.

OK understood. Feel free to reject the PR then.

jralls:
That was meant more as a caution for other work that about this particular case. As you say, we do it for QIF so it makes some amount of sense to do it for the other types as well.

But I do want you to change the commit summary so that it doesn't look like a bug-fix commit. I'll bring that up on #666.

PR #666, opened 2020-03-14
17136391-ofx-file-import When selecting a file for OFX import, restri…

jeanlaroche:
…ct the files listed to .OFX (or .QFX)

### DIFF
--- a/gnucash/import-export/ofx/gnc-ofx-import.c
+++ b/gnucash/import-export/ofx/gnc-ofx-import.c
@@ -1032,6 +1032,8 @@ void gnc_file_ofx_import (GtkWindow *parent)
     char *selected_filename;
     char *default_dir;
     LibofxContextPtr libofx_context = libofx_get_new_context();
+    GList *filters = NULL;
+    GtkFileFilter* filter = gtk_file_filter_new ();
 
     ofx_PARSER_msg = false;
     ofx_DEBUG_msg = false;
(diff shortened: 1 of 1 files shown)


jralls:
Please change the commit summary so that it doesn't look like a bug-fix.

gnucash/import-export/ofx/gnc-ofx-import.c
```diff
@@ -1020,6 +1020,7 @@ double ofx_get_investment_amount(const struct OfxTransactionData* data)
     }
 }
 
+// Let the user selecct a ofx qfx file and start the import
```
jralls: This comment is superfluous, the function is documented in gnc-ofx-import.h. Besides, you mis-spelled 'select'.

jralls:
Please squash these to a single commit with the second commit's summary and force-push.

jralls:
Thanks!

gjanssens:
When I saw the commits I was worried for a moment this would prevent users from selecting files not matching the proper extension. However on testing I see the gtk open dialog automatically adds the option to show all files. So good work!

gnucash/import-export/ofx/gnc-ofx-import.c
```diff
@@ -1043,9 +1045,13 @@ void gnc_file_ofx_import (GtkWindow *parent)
     DEBUG("gnc_file_ofx_import(): Begin...\n");
 
     default_dir = gnc_get_default_directory(GNC_PREFS_GROUP);
+    gtk_file_filter_set_name (filter, "ofx/qfx files (*.ofx,*.qfx)");
```
gjanssens: Shouldn't this string be marked translatable ?

And related to this, I wonder if it wouldn't be better to use a more elaborate file type description rather than using the shorthand twice. I'd propose

```
_("Open/Quicken Financial Exchange file (*.ofx, *.qfx)")
```

jeanlaroche: Did you implement these two suggestions before merging? ^^
I think you're right about them.

gjanssens: I didn't implement anything. I'm just back from a two-week absence. I'm still catching up. And while doing so these two suggestions occurred to me.
Feel free to create a followup PR with these.

gjanssens: In retrospect it would be overkill to make a PR for this one-liner, so don't bother. I have it queued to be pushed after 3.9 release is out.

PR #667, opened 2020-03-16
Add & verify balance assertions

christopherlam:
Add account API to store reconciliation pairs of `statement_date` and `ending_balance`
Upon reconciliation complete, add into account's kvp. Overwrite any previous reconciliation pair.
There's off-by-one error somewhere. Buggy. No `g_free` at all yet. 
Assistance would be nice.
I'll follow on with tools to actually use it and possibly augment the reconciliation report to find errant transactions.

PR #669, opened 2020-03-19
Bug 685102 - Scheduled Transactions don't always respect weekends for…

jeanlaroche:
… first occurrence

### DIFF
--- a/gnucash/gnome/dialog-sx-editor.c
+++ b/gnucash/gnome/dialog-sx-editor.c
@@ -1591,10 +1591,26 @@ gnc_sxed_update_cal(GncSxEditorDialog *sxed)
     GDate start_date, first_date;
 
     g_date_clear(&start_date, 1);
-
+    
+    /* There is an issue is with gnc_frequency_save_to_recurrence: it sets the first date to
+     * the selected one without checking that it's not a weekend, which is problematic if the
+     * user does not want recurrences to fall on a weekend. */
     gnc_frequency_save_to_recurrence(sxed->gncfreq, &recurrences, &start_date);
+    /* One simple solution: if monthly, subtract 1 month from the first and second (in case of by-montly)
+     * instances of recurrence. The mechanism for selecting the next recurrences then fixes the problem. */
+    Recurrence* first_rec = (Recurrence*) recurrences->data;
+    if(first_rec != NULL && first_rec->ptype == PERIOD_MONTH)
+    {
+        g_date_subtract_months(&first_rec->start,1);
+        GList* second = recurrences->next;
+        if(second != NULL)
+            g_date_subtract_months(&((Recurrence*) second->data)->start,1);
+    }
+    
     g_date_subtract_days(&start_date, 1);
     recurrenceListNextInstance(recurrences, &start_date, &first_date);
+    /* Re-add 1 to the start date, otherwise you'll allow the first recurrence to fall before the start date */
+    g_date_add_days(&start_date, 1);
 
     /* Deal with the fact that this SX may have been run before [the
      * calendar should only show upcoming instances]... */


Bob-IT:
Did you look at the travis error, self explanatory ?
Also I do not think this is the place to fix this if as you say there is an underlying issue.
Why not fix that instead?

jralls:
> Why not fix that instead?

Indeed. GnuCash quite enough hacks already. 

@jeanlaroche I've pointed out to you before that C requires declarations at the beginning of a block before any non-declaration code. You seem not to be learning and you seem not to be building and running tests before pushing. Every commit should build and pass tests; commits that don't break `git bisect`, an important technique for tracking down bugs.

I've also cautioned you before on excessive comments. If your code is such a hack that it needs more comment than code you have a design problem and need to think about the problem some more.

jeanlaroche:
> Did you look at the travis error, self explanatory ?
Missed, that, sorry.
> Also I do not think this is the place to fix this if as you say there is an underlying issue.
> Why not fix that instead?
Because this would mean duplicating code, or moving the duplicated code out of its original place and into a separate function which is prone to side-effects. Both solutions don't seem good to me. The solution I suggested isn't bad: it reuses the existing code to fix the issue.
But I really don't care if you guys reject the PR.

jeanlaroche:
> @jeanlaroche I've pointed out to you before that C requires declarations at the beginning of a block before any non-declaration code. You seem not to be learning and you seem not to be building and running tests before pushing. Every commit should build and pass tests; commits that don't break `git bisect`, an important technique for tracking down bugs.
True, I need to run the tests, which I haven't done, my bad.
 
> I've also cautioned you before on excessive comments. If your code is such a hack that it needs more comment than code you have a design problem and need to think about the problem some more.

I don't know if you noticed but the existing gnucash code is almost 100% free of any comment, which makes it impossible to figure out why someone 20 years ago decided to do something a certain way. This ensures nobody can possibly contribute to the code, unless they were part of the original team that made the decision.
Closing this PR.

christopherlam:
Don't be discouraged. The issue is not simply the code quality. It is consistency and understanding the true source of a bug, and fixing that instead. 20yrs of hacks to fix misunderstood bugs have led to many difficult bugs, and it's better fix the source rather than apply a superficial bandaid. Maybe try explore why Recurrence gives unsuitable dates, and try fix that instead. Maybe fixing recurrence will also fix other subtle bugs.

jeanlaroche:
> Don't be discouraged. The issue is not simply the code quality. It is consistency and understanding the true source of a bug, and fixing that instead. 20yrs of hacks to fix misunderstood bugs have led to many difficult bugs, and it's better fix the source rather than apply a superficial bandaid. Maybe try explore why Recurrence gives unsuitable dates, and try fix that instead. Maybe fixing recurrence will also fix other subtle bugs.

Thanks. I do understand where the bug comes from. The first occurrence isn't tested for possible weekend restrictions because it's created using a different function than the function used for subsequent occurrences and this function does no have the code to check for weekend exceptions. 
So there are four solutions: 1) extract the exception-checking code into a separate function that gets called by both functions, 2) reproduce the exception-checking code (that's a no-go of course) 3) ensure that all occurrences that matter are added using the same function (which is what I'm doing), so any change to that function will apply to all or 4) ensure the first and subsequent occurrences are created using a single function.
I chose 3) because that's the solution that seems the safest to me, with the fewest side effects, given that I'm not extremely familiar with the code. 
1) is going to be ugly (one more function, in an already crowded code) and the ideal solution 4) is quite dangerous to me: lots of changes and possibilities for breaking things.

jralls:
Breaking up a long function into a bunch of short functions with a composing function is a standard refactor. It's not ugly at all, especially in the context of a 70-line function if the alternative is making it a 90-line function. To that add that this is ostensibly a View module and calculating the recurrences is a Model function the big-picture right thing to do is to move and perhaps consolidate the functionality with the rest of the Model code.  I guess that's a piece of what you mean by choice 4.

The way to keep the right design from being dangerous is to write tests, and BTW long functions doing more than one thing are a lot harder to test that short functions doing a single thing each that can be composed to do the work of the former long function. 

Oh, and the right place for explaining why you did something is in the commit message, not in comments in the code.

christopherlam:
@jeanlaroche you sound experienced and the code base would definitely benefit from extra eyes. If you are able to break down old code into nicer ones (such as #663) it would be extremely welcome. Adding documentation to functions (such as #670), and more tests (such as 8b616449) equally welcome. For myself I'm much more familiar with guile than C(++) and glib in #667, and desperate for C++ hand holding. Note none of the merged code feature any hackish commentary. Larger useful comments are shown in comments such as 1f83cfaf or in bugzilla. Conclusion: please familiarise with the code base, and help fix bugs and hacks! To implement new features is always difficult because of the 20 year history of hacks, and this has to stop.

With regards to your suggested solutions my view would be that Recurrence suggested date should meet the user-set criteria at all times, so, (4) sounds more correct, however we'd need to verify all uses of Recurrence and decide that every change would be reasonable.

jeanlaroche:
@christopherlam, thanks for the suggestions, they're very welcome. I have to add that JR's testy feedback "you seem not to be learning", "I've cautioned you" achieves absolutely nothing except drive potential contributors away. But I thank you guys for the feedback/suggestions, it always takes a while to adjust to the "culture" of an open-source project.

PR #670, opened 2020-03-20
Bug 797648 - Fix problem reconciling multiple splits in a transaction

goodvibes2:
Force all splits in a transaction for the account being reconciled
(and its children), to have the same reconciliation status. This fixes
a problem when the splits have different statuses.

### DIFF
--- a/gnucash/gnome/reconcile-view.c
+++ b/gnucash/gnome/reconcile-view.c
@@ -524,18 +524,17 @@ gnc_reconcile_view_toggle_split (GNCReconcileView *view, Split *split)
 }
 
 
-/*****************************************************************************\
- * gnc_reconcile_view_rec_or_unrec_split                                     *
- *   insert or remove a child split from the list of splits to be reconciled *
- *   (view->reconciled) so that all other splits in the same transaction     *
- *   for the account being reconciled (including children), are the same     *
- *   reconciliation state as the split that has  been toggled                *
- *                                                                           *
- * Args: view           - the view to use                                    *
- *       split          - the split to be inserted or removed                *
- *       reconcile      - TRUE=insert, FALSE=remove                          *
- * Returns: void                                                             *
-\*****************************************************************************/
+/** Insert or remove a split from the list of splits to be reconciled
+ *   (view->reconciled) so that all other splits in the same transaction
+ *   for the account being reconciled (including children), are the same
+ *   reconciliation state as the split that has been toggled.
+ *
+ *  @param view The view to use.
+ *
+ *  @param split The split to be inserted or removed
+ *
+ *  @param reconcile TRUE=insert, FALSE=remove
+ */
 static void
 gnc_reconcile_view_rec_or_unrec_split (GNCReconcileView *view, Split *split, gboolean reconcile)
 {
--- a/gnucash/gnome/reconcile-view.c
+++ b/gnucash/gnome/reconcile-view.c
@@ -501,8 +501,43 @@ gnc_reconcile_view_class_init (GNCReconcileViewClass *klass)
 }
 
 
-static void
+static gboolean
 gnc_reconcile_view_toggle_split (GNCReconcileView *view, Split *split)
+{
+    Split *current;
+
+    g_return_val_if_fail (GNC_IS_RECONCILE_VIEW (view), FALSE);
+    g_return_val_if_fail (view->reconciled != NULL, FALSE);
+
+    current = g_hash_table_lookup (view->reconciled, split);
+
+    if (current == NULL)
+    {
+        g_hash_table_insert (view->reconciled, split, split);
+        return TRUE;
+    }
+    else
+    {
+        g_hash_table_remove (view->reconciled, split);
+        return FALSE;
+    }
+}
+
+
+/*****************************************************************************\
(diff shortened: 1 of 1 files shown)


gnucash/gnome/reconcile-view.c
```diff
@@ -501,25 +501,60 @@ gnc_reconcile_view_class_init (GNCReconcileViewClass *klass)
...
+ *   insert or remove a child split from the list of splits to be reconciled *
+ *   (view->reconciled) so that all other splits in the same transaction     *
+ *   for the account being reconciled (including children), are the same     *
+ *   reconciliation state as the split that has  been toggled                *
+ *                                                                           *
+ * Args: view           - the view to use                                    *
+ *       split          - the split to be inserted or removed                *
+ *       reconcile      - TRUE=insert, FALSE=remove                          *
+ * Returns: void                                                             *
+\*****************************************************************************/
```
jralls: Please format the comment with [Doxygen](http://www.doxygen.nl/) markup, particularly when you lead the comment with `/**` that flags it for Doxygen to process it:
* No need to repeat the function name
* No need to specify that it returns void (there's no need for that in a not-Doxygen comment either, it's obvious from the function signature).
* Begin each parameter line with `@param`.

jralls:
@goodvibes2 Looks good aside from the comment formatting.

gnucash/gnome/reconcile-view.c
goodvibes2: Done. Thanks John.

derekatkins:
WAIT...  I'm just noticing this now..  WHAT was the problem that this was trying to solve??
Forcing all splits of a transaction to have the same reconcile status is WRONG.  Here's an example:
I make a purchase at Amazon on my credit card.  Total is $100.
They then split the order into $75 that ships immediately and $25 that ships 2 weeks later.
In my mind, the best way to handle this is to split that $100 CC charge into $75 and $25 (and leave the other end of the transaction the same).  HOWEVER, what happens if that $75 charge hits one month and the $25 charge hits the next?
This change will not allow this to happen and will force the $25 to get reconciled when it can't be..

christopherlam:
I am not following this closely but agree with @derekatkins -- e.g. I pay my credit-card bill from my current account. I will reconcile both legs of the same splits at different times.

goodvibes2:
@derekatkins, I had not thought of your use case. The problem I had was that the old code would toggle the checking of the split checkbox for all splits in a transaction for the reconciliation account and its Subaccounts, which meant that a checked split became unchecked when you checked another unchecked split. I think the fix for this is to change it so only the specific split being checked or unchecked is changed. Agree?

@christopherlam, I'm not sure I understand what you are saying. Are you saying your credit card and current account are all part of the account being reconciled? You get one statement that includes transactions for both your credit card and current accounts? In this case my suggestion to only change the specific split being checked or unchecked should still be ok.

christopherlam:
I didn't actually check the new behaviour. So ignore me please :flushed:

goodvibes2:
@derekatkins,
It doesn't show the same split twice. It shows both of the splits for Person1 & Person2 but if you tick the checkbox for Person2, it unticks the checkbox for Person1, so it is not possible in the Reconcile Window to reconcile both of them if to start 1 is not reconciled and 1 is cleared. This is different to having multiple splits to the reconcile account in the same transaction.
I suggest checking or unchecking Person1 should have no effect on Person2 and visa versa.

derekatkins:
> @derekatkins,
> It doesn't show the same split twice. It shows both of the splits for Person1 & Person2 but if you tick the checkbox for Person2, it unticks the checkbox for Person1, so it is not possible in the Reconcile Window to reconcile both of them if to start 1 is not reconciled and 1 is cleared. This is different to having multiple splits to the reconcile account in the same transaction.
> I suggest checking or unchecking Person1 should have no effect on Person2 and visa versa.

So just to be clear, in your case you have a single transaction that goes into 4 accounts (2 expense, 2 Liability), and the 2 Liability splits toggle together?  Honestly, I'm not exactly sure why you'd have a transaction that looks like this, but GnuCash I guess should support it.

IMHO, yes, they should toggle independently, regardless of whether they are in the same account or different sub-accounts.  I don't understand why they would ever toggle together!  What was THAT use-case?

goodvibes2:
@derekatkins,
Yes, that's right.
I have a transaction like that because I want to keep track of who is responsible for the credit card balance when I pay off the credit card from the savings account, so then I can see how much of the savings account balance is mine (because it also has person subaccounts).
I've never seen an explanation of why GC toggled all the subaccount splits but I assume they just thought they would all be reconciled together and hadn't come across your use case.
Perhaps there is a possibility of some-one having a multitude of subaccount splits and just wanted to tick 1. After my mod, they can still easily select all the splits and click the Reconcile button.

goodvibes2:
As this PR was merged into maint, should my change (so that other splits for the reconciliation subaccounts are not automatically updated) go into maint or master?

derekatkins:
@goodvibes2 
> As this PR was merged into maint, should my change (so that other splits for the reconciliation subaccounts are not automatically updated) go into maint or master?

IMHO -- it should go into maint as well.

goodvibes2:
PR https://github.com/Gnucash/gnucash/pull/713 fixes @derekatkins concern above.

PR #672, opened 2020-03-22
Fix path errors as reported by eclipse as problem warnings

goodvibes2:


goodvibes2:
The warnings in the first commit are just typos.
The warnings in the 2nd commit (39c9aa4) I'm *not* sure need to be fixed, but eclipse reports that 
.../src/lib does not exist (from src/gnucash/import-export/csv-imp/test/CMakeLists.txt  ${CMAKE_SOURCE_DIR}/lib
and
/include does not exist (from src/libgnucash/engine/test-core/CMakeLists.txt  ${CMAKE_PREFIX_PATH}/include

Making these changes, regenerating the source project, and rebuilding the project removes these warnings. Hopefully I am helping...

jralls:
Those have been around for a while. I wonder why they didn't raise errors before.

Anyway, thanks!

PR #673, opened 2020-03-22
[window-reconcile] refactor common actions into WidgetSetAmount

christopherlam:
Previous code was ugly - reused char *amount for all strings including non-amount statement-date, gratuitous balance reversals and immediate re-reversals. Refactor common code to render and colorise amount into widget. Verified matches current behaviour in both BANK and CREDITCARD accounts. Instead of reversing sign, render, and re-reversing sign, the rendering function will handle sign reversal using a local variable. I have a suspicion WidgetSetAmount can be upgraded to a common utility library.

### DIFF
--- a/gnucash/gnome/window-reconcile.c
+++ b/gnucash/gnome/window-reconcile.c
@@ -252,6 +252,16 @@ recn_get_account (RecnWindow *recnData)
 }
 
 
+static void
+gnc_add_colorized_amount (gpointer obj, gnc_numeric amt,
+                          GNCPrintAmountInfo print_info, gboolean reverse)
+{
+    if (!obj) return;
+    if (reverse) amt = gnc_numeric_neg (amt);
+    gnc_set_label_color (GTK_WIDGET (obj), amt);
+    gtk_label_set_text (GTK_LABEL (obj), xaccPrintAmount (amt, print_info));
+}
+
 /********************************************************************\
  * recnRecalculateBalance                                           *
  *   refreshes the balances in the reconcile window                 *
@@ -264,7 +274,6 @@ static gnc_numeric
 recnRecalculateBalance (RecnWindow *recnData)
 {
     Account *account;
-    const char *amount;
     gnc_numeric debit;
     gnc_numeric credit;
     gnc_numeric starting;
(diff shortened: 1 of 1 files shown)


gnucash/gnome/window-reconcile.c
```diff
@@ -250,6 +250,16 @@ recn_get_account (RecnWindow *recnData)
 }
 
 
+static void
+WidgetSetAmount (GtkWidget *widget, gnc_numeric amt, GNCPrintAmountInfo print_info,
+                 gboolean reverse)
```
jralls: Except for the old xaccFooDoBar api we should use lower-case-with-underscores for function names, reserving camel case for types. Also, better to be more specific than `widget`. How about `label_set_amount` or `label_add_amount_string_and_color`?

gnucash/gnome/window-reconcile.c
```diff
@@ -274,78 +283,34 @@ recnRecalculateBalance (RecnWindow *recnData)
...
-    ending = recnData->new_ending;
-    if (reverse_balance)
-        ending = gnc_numeric_neg (ending);
-    amount = xaccPrintAmount(ending, print_info);
-    gnc_set_label_color(recnData->ending, ending);
-    gtk_label_set_text(GTK_LABEL(recnData->ending), amount);
-    if (reverse_balance)
-        ending = gnc_numeric_neg (ending);
+    starting = gnc_ui_account_get_reconciled_balance_as_of_date
+        (account, recnData->statement_date, include_children);
```
jralls: I think it would be clearer to call 
```
label_set_amount (recnData->starting, starting, print_info, FALSE);
```
rather than potentially negating it twice.

gnucash/gnome/window-reconcile.c
christopherlam: `gnc_add_colorized_amount`, and can take any gobject.

gnucash/gnome/window-reconcile.c
```diff
@@ -274,78 +283,38 @@ recnRecalculateBalance (RecnWindow *recnData)
...
-    /* update the ending balance */
-    ending = recnData->new_ending;
-    if (reverse_balance)
-        ending = gnc_numeric_neg (ending);
-    amount = xaccPrintAmount(ending, print_info);
-    gnc_set_label_color(recnData->ending, ending);
-    gtk_label_set_text(GTK_LABEL(recnData->ending), amount);
-    if (reverse_balance)
-        ending = gnc_numeric_neg (ending);
+    starting = gnc_ui_account_get_reconciled_balance_as_of_date
```
jralls, line 292: @christopherlam This needs to be `gnc_ui_account_get_reconciled_balance` now, right? 

Aside from that I think you can push this.

PR #674, opened 2020-03-29
Bug 797006 - Balance is misleading in open subaccounts when different…

jeanlaroche:
… currencies are involved
I implemented the solution suggested by John to prevent opening subaccounts if there is a currency / commodity mismatch.
I think a possibly better solution would be to open as read-only, and fix the display issue (i.e., mark each subaccount with its currency and remove the balance column).

### DIFF
--- a/gnucash/gnome-utils/gnc-tree-model-split-reg.c
+++ b/gnucash/gnome-utils/gnc-tree-model-split-reg.c
@@ -451,7 +451,7 @@ gnc_tree_model_split_reg_dispose (GObject *object)
 /* Create a new tree model */
 GncTreeModelSplitReg *
 gnc_tree_model_split_reg_new (SplitRegisterType2 reg_type, SplitRegisterStyle2 style,
-                        gboolean use_double_line, gboolean is_template)
+                        gboolean use_double_line, gboolean is_template, gboolean mismatched_commodities)
 {
     GncTreeModelSplitReg *model;
     GncTreeModelSplitRegPrivate *priv;
@@ -469,6 +469,7 @@ gnc_tree_model_split_reg_new (SplitRegisterType2 reg_type, SplitRegisterStyle2 s
     model->style = style;
     model->use_double_line = use_double_line;
     model->is_template = is_template;
+    model->mismatched_commodities = mismatched_commodities;
 
     model->sort_col = 1;
     model->sort_depth = 1;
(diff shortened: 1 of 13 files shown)


jralls:
> I think a possibly better solution would be to open as read-only, and fix the display issue (i.e., mark each subaccount with its currency and remove the balance column).

I think that would be a better solution too.  Note that there's one other display problem shown the second screenshot on the bug: the splits in the different currency (Indian Rupees, FWIW) don't display the debit/credit amounts.

jeanlaroche:
It took me a while to find the place where columns are chosen for different types of ledgers/registers. (split_register_layout.c if you're curious). So the fix is simple and clean: a new flag is added to the split register indicating mixed commodities, and if so, the columns chosen are slightly different (there's already code similar to that in that c file).

> BTW, there's a gotcha hiding in the register code: There's two copies of lots
> of it, one for the original implementation that we're still using, the other
> for a reimplementation of the display as a GtkTreeView called register2. When
> the latter was created the author thought it would be easy and just whole files
> and modified only the bits that needed changing, figuring that he'd just delete
> the originals when everything worked. Unfortunately he got tantalizingly close
> but never quite got all of the bugs worked out. We've been trying to keep the
> two sets at least in sync so that when someone takes up register2 again the
> bitrot won't be too overwhelming.

So I should duplicate my code there (somewhat blindly since that's not tested)? There seems to also be a duplicate INCOME_LEDGER named INCOME_LEDGER2 in gnc-tree-view-split-register.c . Can you clarify why that is there? Wondering if some changes are needed there as well.

jralls:
Yes, duplicate the code. You can test it by starting gnucash with `--extra` then use Extensions>Register2>OpenSubaccounts.

There is `INCOME_LEDGER` and `INCOME_LEDGER2`, but not in `gnc-tree-view-split-reg.`; that's a Reg2 file.  The Reg1 version is `gnucash/register/ledger-core/split-register.c` where you'll find INCOME_LEDGER filling a similar role. The only file with both of them is `gnucash/gnome/gnc-plugin-page-register2.c` and it's either a mistake or for some reason the Reg1 enum works in that case.

jeanlaroche:
Updated a few 2* files, things are working, but I don't think these changes achieve anything because there does not seem to be a problem with displaying subaccount when using --extra.

jralls:
Thanks!

PR #675, opened 2020-03-30
Fix minor typos

yurchor:
Many thanks in advance for reviewing.

### DIFF
--- a/gnucash/gnome-utils/gnc-tree-view-split-reg.c
+++ b/gnucash/gnome-utils/gnc-tree-view-split-reg.c
@@ -592,7 +592,7 @@ gnc_tree_view_split_reg_pref_changed (gpointer prefs, gchar *pref, gpointer user
 
 /* Define which columns are in which views */
 static ViewCol *
-gnc_tree_view_split_reg_get_colummn_list (GncTreeModelSplitReg *model)
+gnc_tree_view_split_reg_get_column_list (GncTreeModelSplitReg *model)
 {
     DEBUG("Model-type is %d", model->type);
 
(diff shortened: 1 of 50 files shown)


gnucash/import-export/log-replay/gnc-log-replay.h
```diff
@@ -29,7 +29,7 @@
 
 /** The gnc_file_log_replay() routine will pop up a standard file
  *     selection dialogue asking the user to pick a log file to replay. If one
- *     is selected the the .log file is opened and read.  It's contents
+ *     is selected the .log file is opened and read.  Its contents
```
fellen: No, AFAIK in English genitive is `~'s`,
plural often `~s` and
genitive plural then `~s'`.
@jralls right?

gnucash/gnome-utils/gnc-tree-view-split-reg.c
```diff
@@ -592,7 +592,7 @@ gnc_tree_view_split_reg_pref_changed (gpointer prefs, gchar *pref, gpointer user
 
 /* Define which columns are in which views */
 static ViewCol *
-gnc_tree_view_split_reg_get_colummn_list (GncTreeModelSplitReg *model)
+gnc_tree_view_split_reg_get_column_list (GncTreeModelSplitReg *model)
```
fellen: At least I seen no external use of this. So it should be OK.

gnucash/import-export/log-replay/gnc-log-replay.h
jralls: Nope, because of the collision with the contraction of "it is" that takes precedence over the possessive. "It's" is the contraction, "its" is the genetive/possessive. It's a common error, even native speakers screw it up about as often as they use "loose" (release from confinement or make not tight) when they want "lose" (give up possession or misplace).

yurchor:
The requested change is squashed with the commit.

Many thanks for reviewing.

fellen:
Thanks for contributing.

PR #676, opened 2020-03-31
A functional gncOwnerFromLot

christopherlam:
Convenience wrapper around `owner_from_lot` and the resulting `*owner` is freed automatically in SWIG.


jralls:
It would be immensely better as well as less work to use gncOwnerGetOwnerFromLot which is already available to Scheme.

christopherlam:
Dead end.

PR #677, opened 2020-04-02
high level python wrappers for app-utils

c-holtermann:
I'd like to have wrapper methods that return instances

### DIFF
--- a/bindings/python/CMakeLists.txt
+++ b/bindings/python/CMakeLists.txt
@@ -1,7 +1,7 @@
 add_subdirectory(example_scripts)
 add_subdirectory(tests)
 
-set(PYEXEC_FILES  __init__.py function_class.py gnucash_business.py gnucash_core.py)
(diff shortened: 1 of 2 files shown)


jralls:
You only need the `Session` one. The other two are already wrapped as `Session.get_book()`  and `Book.get_root_account()`.

c-holtermann:
@jralls ok. I'll remove them. I'll look throught the rest of the functions visible for python in _sw_app_utils and see if any other makes sense.

jralls:
#678 replaces this, right?

c-holtermann:
This is different to and based on #678.

jralls:
@c_holterman, it is now. It wasn't when I wrote that. This is OK now, I'll merge it when #678 is ready.

c-holtermann:
Thanks!

jralls:
Thanks!

PR #678, opened 2020-04-02
add instance argument to Session constructor

c-holtermann:
if Session is used only as a wrapper for an existing session instance
its necessary to be able to pass this instance as an argument

Session(instance=gnucash._sw_app_utils.gnc_get_current_session()) is impossible otherwise as can be used in the python shell

### DIFF
--- a/bindings/python/gnucash_core.py
+++ b/bindings/python/gnucash_core.py
@@ -75,7 +75,7 @@ class Session(GnuCashCoreClass):
     """
 
     def __init__(self, book_uri=None, ignore_lock=False, is_new=False,
-                 force_new= False):
+                 force_new=False, instance=None):
         """A convenient constructor that allows you to specify a book URI,
         begin the session, and load the book.
 
@@ -95,13 +95,15 @@ class Session(GnuCashCoreClass):
         ignore_lock is passed to qof_session_begin's argument of the
         same name and is used to break an existing lock on a dataset.
 
+        instance argument can be passed if new Session is used as a
+        wrapper for an existing session instance
 
 
         This function can raise a GnuCashBackendException. If it does,
         you don't need to cleanup and call end() and destroy(), that is handled
         for you, and the exception is raised.
         """
-        GnuCashCoreClass.__init__(self)
+        GnuCashCoreClass.__init__(self, instance=instance)
         if book_uri is not None:
             try:
                 self.begin(book_uri, ignore_lock, is_new, force_new)
--- a/bindings/python/tests/test_session.py
+++ b/bindings/python/tests/test_session.py
@@ -1,0 +1,34 @@
+# test cases for Session wrapper object
+#
+# test for get_book may belong in test_book but it makes sense here
+# to see if get_current_session works
+# test for app_utils on the other hand could go to a subfolder of
+# /libgnucash/app-utils
+#
+# @date 2020-04-03
+# @author Christoph Holtermann <mail@c-holtermann.net>
+
+from unittest import TestCase, main
+
+from gnucash import Session
+
+class TestSession(TestCase):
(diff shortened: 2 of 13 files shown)


c-holtermann:
Fails for ninja, I'll have to check why.

c-holtermann:
They only work together. Could have made them one commit.

c-holtermann:
Problem found. ClassFromFunctions.__init__(instance=None) will not create a new instance but an object with its __instance=None. I don't really think that makes sense but I'll keep it that way and use a workaround for creating new Session objects.

c-holtermann:
Passes the tests but doesn't work yet.

c-holtermann:
Test provokes this error. I found that SWIG translates QofSession differently for app-utils and python bindings. Importing python-bindings into swigging app-utils solves this.

c-holtermann:
This looks finished to me. If it's ok for you reviewers then I'll squash the commits.

bindings/python/gnucash_core.py
```diff
@@ -95,13 +95,21 @@ def __init__(self, book_uri=None, ignore_lock=False, is_new=False,
...
         you don't need to cleanup and call end() and destroy(), that is handled
         for you, and the exception is raised.
         """
-        GnuCashCoreClass.__init__(self)
+        if instance:
+            GnuCashCoreClass.__init__(self, instance=instance)
+        else:
+            # even if instance is None GnuCashCoreClass.__init__ will still
+            # not create a new instance, only omitting the instance parameter
+            # works. Maybe change that?
```
jralls, line 111: Yes, change that. Otherwise you risk passing `None` to the C API and it will crash.

common/cmake_modules/GncAddSwigCommand.cmake
```diff
@@ -84,7 +84,8 @@ macro (gnc_add_swig_python_command _target _out_var _py_out_var _output _py_outp
             ${CMAKE_SOURCE_DIR}/common
             ${CMAKE_SOURCE_DIR}/libgnucash/engine
             ${CMAKE_SOURCE_DIR}/libgnucash/app-utils
-        )
+       	    ${CMAKE_SOURCE_DIR}/bindings/python
```
jralls, line 87: Please fix the indentation.

bindings/python/tests/test_session.py
```diff
@@ -0,0 +1,34 @@
...
+# /libgnucash/app-utils
+#
+# @date 2020-04-03
+# @author Christoph Holtermann <mail@c-holtermann.net>
+
+from unittest import TestCase, main
+
+from gnucash import Session
+
+class TestSession( TestCase ):
```
jralls, line 15: Whitespace around parentheses is [unpythonic](https://www.python.org/dev/peps/pep-0008/#whitespace-in-expressions-and-statements).

jralls:
Thanks. I made one small change in `bindings/python/tests/CMakeLists.txt`:
```
     # Because it hasn't been built yet
     set(test_core_dir ${CMAKE_BINARY_DIR}/common/test-core)
   endif()
-  add_custom_target(test-python-bindings ALL DEPENDS unittest_support gnucash-core-c-build gnucash-core-c-py)
+  add_custom_target(test-python-bindings ALL DEPENDS unittest_support gnucash-core-c-build gnucash-core-c-py swig-app-utils-python)
   add_dependencies(check test-python-bindings)
   add_test(python-bindings ${PYTHON_EXECUTABLE} ${CMAKE_CURRENT_SOURCE_DIR}/runTests.py.in)
   set_property(TEST python-bindings PROPERTY ENVIRONMENT
```
Because otherwise one has to run `ninja` before `ninja check` for the test to pass.

c-holtermann:
Thanks!

PR #679, opened 2020-04-03
Make pycons/ishell.py compatible with Python 3 and current IPython

cjmayo:
I've only tested this with python -Wd and not actually used it.

ipapi.get() has been deprecated since 0.11 (2011) apparently:
https://stackoverflow.com/questions/13864126/no-module-named-ipapi#comment19164908_13864322

### DIFF
--- a/gnucash/python/pycons/ishell.py
+++ b/gnucash/python/pycons/ishell.py
@@ -16,12 +16,11 @@
 import os
 import sys
 import re
-from StringIO import StringIO
+from io import StringIO
 try:
     import IPython
-    from IPython import ipapi
-except Exception,e:
-    raise "Error importing IPython (%s)" % str(e)
+except Exception as e:
+    raise Exception("Error importing IPython (%s)" % str(e))
 
 
 # ------------------------------------------------------------------ class Shell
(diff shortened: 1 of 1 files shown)


c-holtermann:
There have been a lot of changes to ipython that need to be taken care of as well. Pycons has been used in other projects. Some time ago I had a look if someone elsewhere had modified it to work with recent versions but didn't find a working solution. Maybe that has changed.

c-holtermann:
I'm finde with these changes. I tried to make ipython shell work some while ago but didn't finish the process. I made about the same changes but did not commit them. I went some steps further: https://github.com/Gnucash/gnucash/compare/maint...c-holtermann:pycons2to3#diff-fd858d01561e31d51b110e2d2bb8c7f5 
That diff link is a mix of changes to ipython shell and other stuff but you can have a look at the ipython shell file.

cjmayo:
I was only trying to silence some errors when byte-compiling pycons, rather than get ishell working - plus I can see from your work there is a lot more to that.
Do take this if you want to, else probably the best solution for me is just not to install ishell.py.

c-holtermann:
I'll put together the changes I have made to ipython shell so far.

c-holtermann:
My changes are here: https://github.com/c-holtermann/gnucash/commit/617cf417659c05b6c3cb4469d575bb235dc22b48
I'm fine with this PR and can put my changes on top later on.

jralls:
OK, merged. Thanks.

PR #681, opened 2020-04-04
797318 - Amounts ending in zero displayed as fractions

jeanlaroche:
The function gnc_numeric_to_decimal() should return TRUE if num is 0 but currently does not.

### DIFF
--- a/libgnucash/app-utils/gnc-ui-util.c
+++ b/libgnucash/app-utils/gnc-ui-util.c
@@ -1605,7 +1605,9 @@ PrintAmountInternal(char *buf, gnc_numeric val, const GNCPrintAmountInfo *info)
         *buf = '\0';
         return 0;
     }
-
+    
+    // Value may now be decimal, for example if the factional part is zero
+    value_is_decimal = gnc_numeric_to_decimal(&val, NULL);
     /* print the integer part without separators */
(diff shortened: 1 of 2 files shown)


libgnucash/engine/iso-4217-currencies.xml
```diff
@@ -1744,7 +1744,7 @@
   namespace="ISO4217"
   exchange-code="478"
   parts-per-unit="5"
-  smallest-fraction="100"
+  smallest-fraction="5"
```
jralls, line 1747: We follow [ISO-4217](https://www.currency-iso.org/dam/downloads/lists/list_one.xml) and it says
```
<CcyNtry>
<CtryNm>MAURITANIA</CtryNm>
<CcyNm>Ouguiya</CcyNm>
<Ccy>MRU</Ccy>
<CcyNbr>929</CcyNbr>
<CcyMnrUnts>2</CcyMnrUnts>
</CcyNtry>
```
So smallest fraction stays at 100. Users may set 5 in accounts if they like.

jeanlaroche: OK I"ll revert that change, but the rest can stay in case the user changes it, right?

jralls: Yes, the bit about returning true on 0 is a good catch indeed.

jralls:
Thanks!

PR #682, opened 2020-04-06
Add account balance variable for scheduled transactions

jeanlaroche:
I'd like to suggest adding a variable "balance" to the already existing "i" in the schedule transaction editor. This variable would contain the value of the account balance at the time the scheduled transaction is executed. It can be used, for example (and mostly I imagine) to compute the interests on a loan or a mortgage.
In the US at least, mortgage interests are typically computed based on the account balance at the current month, so that makes it very easy to create a scheduled transaction for mortgage payment that will be robust to overpayment and will assign the proper ratio to principal reduction and the rest for interests. 
This was mentioned as a request, and I personally wished it was available to allow flexible mortgage scheduled transactions (a-la quicken) so I went ahead an implemented it. Let me know if you think that's a really bad idea.

### DIFF
--- a/libgnucash/app-utils/gnc-exp-parser.c
+++ b/libgnucash/app-utils/gnc-exp-parser.c
@@ -526,7 +526,7 @@ gnc_exp_parser_parse_separate_vars (const char * expression,
                       trans_numeric, numeric_ops, negate_numeric, g_free,
                       func_op);
 
-    error_loc = parse_string (&result, expression, pe); // JEAN VALUE COMPUTED HERE?
+    error_loc = parse_string (&result, expression, pe);
 
     pnum = result.value;
 
--- a/libgnucash/app-utils/gnc-exp-parser.c
+++ b/libgnucash/app-utils/gnc-exp-parser.c
@@ -526,7 +526,7 @@ gnc_exp_parser_parse_separate_vars (const char * expression,
                       trans_numeric, numeric_ops, negate_numeric, g_free,
                       func_op);
 
-    error_loc = parse_string (&result, expression, pe);
+    error_loc = parse_string (&result, expression, pe); // JEAN VALUE COMPUTED HERE?
 
     pnum = result.value;
 
--- a/libgnucash/app-utils/gnc-sx-instance-model.c
+++ b/libgnucash/app-utils/gnc-sx-instance-model.c
@@ -353,9 +353,6 @@ _clone_sx_var_hash_entry(gpointer key, gpointer value, gpointer user_data)
     g_hash_table_insert(to, g_strdup(key), var);
 }
 
-#define GET_PRIVATE(o)  \
-((AccountPrivate*)g_type_instance_get_private((GTypeInstance*)o, GNC_TYPE_ACCOUNT))
-
 static GncSxInstance*
 gnc_sx_instance_new(GncSxInstances *parent, GncSxInstanceState state, GDate *date, void *temporal_state, gint sequence_num)
 {
@@ -380,18 +377,20 @@ gnc_sx_instance_new(GncSxInstances *parent, GncSxInstanceState state, GDate *dat
 
     {
         int instance_i_value;
-        gnc_numeric i_num;
+        gnc_numeric num_value;
         GncSxVariable *as_var;
+        gchar* var_name = NULL;
 
-        // JEAN: Xch TRANS is re-created.
+        var_name = "i";
         instance_i_value = gnc_sx_get_instance_count(rtn->parent->sx, rtn->temporal_state);
-        i_num = gnc_numeric_create(instance_i_value*1000, 1);
-        as_var = gnc_sx_variable_new_full("i", i_num, FALSE);
-        g_hash_table_insert(rtn->variable_bindings, g_strdup("i"), as_var);
+        num_value = gnc_numeric_create(instance_i_value, 1);
+        as_var = gnc_sx_variable_new_full(var_name, num_value, FALSE);
+        g_hash_table_insert(rtn->variable_bindings, g_strdup(var_name), as_var);
         
-        i_num = gnc_numeric_create(1111, 1);
-        as_var = gnc_sx_variable_new_full("j", i_num, FALSE);
-        g_hash_table_insert(rtn->variable_bindings, g_strdup("j"), as_var);
+        var_name = "balance";
+        num_value = gnc_numeric_create(0, 1);
+        as_var = gnc_sx_variable_new_full(var_name, num_value, FALSE);
+        g_hash_table_insert(rtn->variable_bindings, g_strdup(var_name), as_var);
     }
 
     return rtn;
@@ -1041,7 +1040,7 @@ _get_sx_formula_value(const SchedXaction* sx,
         {
             parser_vars = gnc_sx_instance_get_variables_for_parser(variable_bindings);
         }
-        if (!gnc_exp_parser_parse_separate_vars(formula_str, // JEAN: THE FORMULA IS PARSED HERE.
+        if (!gnc_exp_parser_parse_separate_vars(formula_str,
                                                 numeric,
                                                 &parseErrorLoc,
                                                 parser_vars))
@@ -1081,15 +1080,6 @@ _get_debit_formula_value(GncSxInstance *instance, const Split *template_split,
                           "sx-debit-numeric", instance->variable_bindings);
 }
 
-static void
-xaccJean(Account* acc)
-{
-    gnc_numeric balance = xaccAccountGetBalance(acc);
-    const gchar* name = gnc_account_get_full_name(acc);
-    const gchar* desc = xaccAccountGetDescription(acc);
-    g_print("ACCOUNT NAME %s Desc %s Bal %d %d\n",name,desc,(gint)balance.num,(gint) balance.denom);
-
-}
 static gnc_numeric
 split_apply_formulas (const Split *split, SxTxnCreationData* creation_data)
 {
@@ -1098,16 +1088,11 @@ split_apply_formulas (const Split *split, SxTxnCreationData* creation_data)
     gnc_numeric final;
     gint gncn_error;
     SchedXaction *sx = creation_data->instance->parent->sx;
-    // JEAN: The value of the formula is computed here?
+    
     _get_credit_formula_value(creation_data->instance, split, &credit_num,
                               creation_data->creation_errors);
     _get_debit_formula_value(creation_data->instance, split, &debit_num,
                              creation_data->creation_errors);
-
-    Account* foo;
-    _get_template_split_account(NULL, split, &foo,NULL);
-    xaccJean(foo);
-
 
     final = gnc_numeric_sub_fixed(debit_num, credit_num);
 
@@ -1237,6 +1222,7 @@ create_each_transaction_helper(Transaction *template_txn, void *user_data)
     SchedXaction *sx = creation_data->instance->parent->sx;
     gnc_commodity *txn_cmdty = get_transaction_currency (creation_data,
                                                          sx, template_txn);
+    GncSxVariable *as_var = NULL;
 
     /* No txn_cmdty means there was a defective split. Bail. */
     if (txn_cmdty == NULL)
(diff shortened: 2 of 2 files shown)


jralls:
No doubt it would be useful and would make the normal sinking fund loan handling a lot better. The catch is that once it's out there users will want to use it for other things and will complain when it doesn't work.

jeanlaroche:
> No doubt it would be useful and would make the normal sinking fund loan handling a lot better. The catch is that once it's out there users will want to use it for other things and will complain when it doesn't work.

Yes! And it will be a bit tricky to make it work the right way (I set it up on my account, and I had to redo the scheduled transaction a few times, before I got the debit/credit trickiness right between the asset/liabilities income/expense parts of it!). A good guide will help, but only to a point, as we know.
This being said, formulas in scheduled transactions are for "advanced users" I would imagine, right? Like even the "i" variable thing (that counts occurrences) is kind of hidden in the doc... 
But still, that's not a very good approach: better do it right and make it easy to use and robust.

jeanlaroche:
Would the right thing be to create a new type of account "Loan" with the right fields (accural, rate, balloon penalty etc), and all the mechanics to handle interest accrual?
If so, is that an effort that's worth it in your opinion?

christopherlam:
The SX formula is approaching a DSL. I can see lots of featuritis and desired further hacks.

One way to proceed is to ensure that users can extend SX formulas. Currently only `i` is available. Could the current `account` also be accessible? Therefore further extension goes to `fin.scm`?

OTOH I'm not sure `fin.scm` belonging to `libgnucash/app-utils` is likely undesirable. Maybe it should be moved to `gnucash/somewhere else`.

jeanlaroche:
> The SX formula is approaching a DSL. I can see lots of featuritis and desired further hacks.
> 
Don't know what DSL stands for in that context!
> One way to proceed is to ensure that users can extend SX formulas. Currently only `i` is available. Could the current `account` also be accessible? Therefore further extension goes to `fin.scm`?
> 
My PR makes account balance available. We could add more, of course, but the objections above stand (the values we have have to be robustly defined when the sx formula is calculated).

> OTOH I'm not sure `fin.scm` belonging to `libgnucash/app-utils` is likely undesirable. Maybe it should be moved to `gnucash/somewhere else`.

christopherlam:
> DSL

https://en.wikipedia.org/wiki/Domain-specific_language

derekatkins:
> Would the right thing be to create a new type of account "Loan" with the right fields (accural, rate, balloon penalty etc), and all the mechanics to handle interest accrual?
> If so, is that an effort that's worth it in your opinion?

No, I don't think we need that.  Most of the data necessary is already captured in the Loan & Mortgage Repayment assistant.

> > One way to proceed is to ensure that users can extend SX formulas. Currently only `i` is available. Could the current `account` also be accessible? Therefore further extension goes to `fin.scm`?
> 
> My PR makes account balance available. We could add more, of course, but the objections above stand (the values we have have to be robustly defined when the sx formula is calculated).

Personally I think that just having it be the balance of the "post" date of the transaction should be sufficient.  I.e., whatever the Date will be on the posted transaction, it should be the balance as of that date.  Of course the outstanding questions are:
1) Is that sufficient, and
2) Can you assume the account from which you get the balance?  
In a mortgage repayment transaction there are a minimum of 3 accounts, usually at least 4!  There is the Asset (bank where payment is coming from), Liability (Loan Principal), and Expense (Interest).  But there could be another Asset, Escrow, and other expenses, like PMI.

Then there's the order-of-operations when you have potentially multiple SXes firing off.   It's one thing to ensure posting order when you have multiple instances of one SX, but what happens if you have multiple SXes that need to interact?

I am not trying to discourage you -- this is a GREAT feature idea.  But as always, the devil is in the details and users will want to use it for their things..  So it would be good to at least think about it all up front, even if we decide that we're only going to support Loan repayments.

jeanlaroche:
> Personally I think that just having it be the balance of the "post" date of the transaction should be sufficient. I.e., whatever the Date will be on the posted transaction, it should be the balance as of that date. Of course the outstanding questions are:
> 
>     1. Is that sufficient, and

I think it is. I can only think of very twisted cases where the end of time balance wouldn't be the right one (like, inserting a transaction with a date posterior to the scheduled x but *before* the scheduled one has run, not very easy to do!). 


>     2. Can you assume the account from which you get the balance?
>        In a mortgage repayment transaction there are a minimum of 3 accounts, usually at least 4!  There is the Asset (bank where payment is coming from), Liability (Loan Principal), and Expense (Interest).  But there could be another Asset, Escrow, and other expenses, like PMI.

You're right. I thought about this aspect, but I think it's simpler than it looks. I think there's only 1 account whose balance you're interested in, it's the liability one. My assumption so far is that that's the account in the first part of the split (arbitrary). But of course, we could add some smarts to detect the liability account if it's not the first one. Or to provide variables for the balance for each account of the splits (overkill I would say).

As far as the order: I believe the solution is somewhat simple: sort the transactions in the order they are supposed to execute, and update the balances in between.

jralls:
> I think it is. I can only think of very twisted cases where the end of time balance wouldn't be the right one (like, inserting a transaction with a date posterior to the scheduled x but before the scheduled one has run, not very easy to do!).

No, that's really easy to do. Just manually create some transactions posted in the future after the SX post_date (e.g. scheduled bill payments or even an ACH on an extra sinking fund payment), then quit GnuCash and don't start it back up until after the SX post_date.

> You're right. I thought about this aspect, but I think it's simpler than it looks. I think there's only 1 account whose balance you're interested in, it's the liability one. My assumption so far is that that's the account in the first part of the split (arbitrary). But of course, we could add some smarts to detect the liability account if it's not the first one. Or to provide variables for the balance for each account of the splits (overkill I would say). You can get most of the way to addressing that by using `xaccAccountGetBalanceAsOfDate()` as @derekatkins suggests, though if there's more than one transaction on the liability account on the day of the SX you probably need to apply the interest rate to the balance from the day before for loans.

There's a related problem, though: If there have been changes in the balance of the account over the period that you're computing interest then the only way to get it right is to iterate over the dates and calculate the interest amount based on the average balance for the compounding period. That would call for more variables, e.g. `avg_balance` and `daily_balance`; the former could be computed using the same approach as the present `balance`; the latter would have to invoke a callback taking the date range and the interest rate.

You can't depend on the order of the splits but it's reasonable to stipulate that the balance variable works only when there's a single liability account in the SX and that's the balance that will be retrieved.

> As far as the order: I believe the solution is somewhat simple: sort the transactions in the order they are supposed to execute, and update the balances in between.

Not so fast. First of all, how do you decide the correct order for transactions posted on a particular day?  Second, the SLR logic creates the transactions, optionally presents them for review, and then posts them.

BTW, first testing release is in 19 days and feature freeze is 31 May.

jeanlaroche:
> There's a related problem, though: If there have been changes in the balance of the account over the period that you're computing interest then the only way to get it right is to iterate over the dates and calculate the interest amount based on the average balance for the compounding period. That would call for more variables, e.g. `avg_balance` and `daily_balance`; the former could be computed using the same approach as the present `balance`; the latter would have to invoke a callback taking the date range and the interest rate.

Wait. My assumption is that this is only good for "simple" mortgage type loans where interests are compounded monthly based on the balance at the end of the month. If interests are compounded daily, things are significantly more complicated, it would be a bit of a nightmare to make it work! In the simple case, if there's a payment during the month, the date does not matter, as long as it's before the cutoff at which time the interest is computed. 
Or am I missing something?

jralls:
> Or am I missing something?

You're missing that other users will have other use cases and will consider it a bug if it doesn't work for their case.

jeanlaroche:
> > Or am I missing something?
> 
> You're missing that other users will have other use cases and will consider it a bug if it doesn't work for their case.

I don't really know how we can make this work for a loan that compounds daily, it would be a tremendous hack. Or you'd have to add a scheduled transaction that fires every day!

For this type of scenario, I think it would make a lot more sense to have a dedicated account type (loan) as I suggested above, with the right settings (compound type etc) so the right computations can be done, not using scheduled transactions. 

In the meantime, is it useful to provide this solution, even if it does not do the right thing for all loans?

jralls:
> For this type of scenario, I think it would make a lot more sense to have a dedicated account type (loan) as I suggested above

Not an account, but perhaps an object that has an account reference member  along with member variables for duration, interest, and compounding and an interest-calculation member function.

> In the meantime, is it useful to provide this solution, even if it does not do the right thing for all loans?

Maybe, but you should run it past the users on the mailing list.
Be clear that it's limited to loans that allow only one payment per period, credited on the due date, and that interest compounds per period. The additional balance variable is to account for the previous payment being different from the planned payment.

jeanlaroche:
> Maybe, but you should run it past the users on the mailing list.
> Be clear that it's limited to loans that allow only one payment per period, credited on the due date, and that interest compounds per period. The additional balance variable is to account for the previous payment being different from the planned payment.

You can make as many payments to the loan as you want during the month though and it does not matter when you make them, they're assumed to decrease the principal (that's an assumption, but I think many banks make that assumptions, over-payments go to principal reduction). The computation of the interest/principal reduction will be accurate at the end of the month (or period if you prefer) as long as interested are compounded per period.

jralls:
I would be very surprised to learn of a loan that allows those terms. Consider an extreme example where you owe $1000 and make a payment for $999 that clears the day before the due date. You'd pay interest on only $1.

jeanlaroche:
That's the way my mortgage loan (and many out there in the US) works. I make an extra payment toward the end of the month, and that's deducted from the principal. The next month's interest is computed based on the reduced principal. That's actually pretty standard I think. I looked online to verify that and read that banks have to agree to this, but it's generally the case (for *mortgages* that is!)

jralls:
I couldn't find a website that said anything at all about how interest is calculated, just a lot of calculators showing the total interest saved by paying down the loan early. What the URI of the one you found?

Calculating through the examples on those calculators shows that the savings come from having payed down the loan early (a common example is making bi-weekly payments so you effectively make an extra payment every year, which pays off a 30 year loan in 25 years). The difference in the monthly accrued interest is in the noise.

Regardless, I suppose you can carefully explain that assumption about inter-period payments and warn that users should carefully examine their loan documents and consult their loan officers before deciding that it's valid for them. Not a licensed accountant, not accounting advice, yada yada yada.

jeanlaroche:
Here's one:
https://www.mtgprofessor.com/A%20-%20Early%20Payoff/making_your_extra_mortgage_payments_count.html

Of interest here:

> Simple interest mortgages, on which interest accrues daily, are an exception

>  Is There a Best Time Within the Month to Make an Extra Payment to Principal? Yes, the best time within the month to make an extra payment is the last day on which the lender will credit you for the current month, rather than deferring credit until the following month. If it is the 15th, for example, an extra payment made within the first 15 days of January will reduce your balance that month and the interest due in February. Payments made the 16th or later will not be credited until February, and the interest deduction will be deferred until March.  
There is no universal lender practice in crediting extra payments. Some lenders will credit payments received anytime during the month while others are much more restrictive. In most cases, extra payments sent in with the scheduled payment will be credited the same month, but it is a good idea to ask your lender what their rule is.  

So you're right in the sense that the user should know what they're doing otherwise the calculations will not reflect what's going on with their loan. 
The saving grace is that using variables in scheduled transactions is a bit of an advanced topic, so one could assume the user knows what they're doing? Or perhaps not! :)

jralls:
Surely you know about what assume makes out of you and me? ;-)

jeanlaroche:
:)

christopherlam:
> That's the way my mortgage loan (and many out there in the US) works. I make an extra payment toward the end of the month, and that's deducted from the principal. The next month's interest is computed based on the reduced principal. That's actually pretty standard I think

From my experience in UK and Australia, banks calculate interest on _daily_ mortgage balances. Nonetheless my view is that you can try make it work, but it'll be limited to your use case only, and _very_ difficult to amend the formulas when circumstances change (e.g. change of interest rate method, mortgage interest holiday such as with today's looming recession/depression, paying down principal with lumpsum, home loan offset account / HELOC). This feature _won't_ affect me because I trust the bank interest calculation more than I trust my own, so I don't mind this feature at all. But interested to know how you'd do it.

Personally I'd simply augment the interface to `fin.scm` to pass on some functions e.g. the SX formula could accept `xaccAccountGetBalanceAsOfDate` to be `local-eval`uated by `fin.scm` but IMHO this is _very_ dangerous because it'd allow arbitrary scheme code execution by a well-crafted .gnucash file (which I can make!).

derekatkins:
> Personally I'd simply augment the interface to `fin.scm` to pass on some functions e.g. the SX formula could accept `xaccAccountGetBalanceAsOfDate` to be `local-eval`uated by `fin.scm` but IMHO this is _very_ dangerous because it'd allow arbitrary scheme code execution by a well-crafted .gnucash file (which I can make!).

Well, users can already run arbitrary scheme code.  However, the problem is that there isn't enough information going into fin.scm to run an xaccAccountGetBalanceAsOfDate() from the fin.scm/SX code.  Therefore, IMHO, adding a "balance" variable really solves that particular problem.  I can tell you for certain that this feature would definitely work for *my* use case of pre-paying mortgage amounts.

jeanlaroche:
> I can tell you for certain that this feature would definitely work for _my_ use case of pre-paying mortgage amounts.

Same here, it would 100% solve my problem. But I realize that's quite specific to certain US users. One issue with fin.scm is that it's *really* cryptic for people who are not familiar with scheme! Definitely not for the casual user.

jeanlaroche:
I added code to sort scheduled transactions between posting them. This was a bit involved as they were presented in two nested loops (outer one for each scheduled transaction, inner one for each date each transaction needed to be posted), which prevented pre-sorting them. I flattened the nested loop into a single loop over all combined transactions, which allowed the list to be date-sorted. With that I was able to verify that transactions post in the right order, and the balanced used in each formula is accurate.
A unit test would be great for that but I have no idea how to write one for this type of operation. Any hints would be greatly welcome.

Remaining to do: code to identify the liability account whose balance will be used.

jeanlaroche:
I added the code to detect the first liability account. I believe this should complete this PR, unless I missed something obvious.

jralls:
You mean aside from getting it to compile?
```
/home/john/gnucash-git/libgnucash/app-utils/gnc-sx-instance-model.c: In function ‘create_each_transaction_helper’:
/home/john/gnucash-git/libgnucash/app-utils/gnc-sx-instance-model.c:1285:9: error: ISO C90 forbids mixed declarations and code [-Werror=declaration-after-statement]
 1285 |         GNCAccountType acct_type = xaccAccountGetType (split_acct);
      |         ^~~~~~~~~~~~~~
cc1: all warnings being treated as errors
```

jeanlaroche:
> You mean aside from getting it to compile?
:)  I'll fix that. Maybe we can figure out how to force xcode to compile this as c, not c++. I try to catch these by inspecting my code, but obviously I'm not as good as a c compiler!

PR #683, opened 2020-04-07
[Master] Fix issue when editing the account tree view notes field

Bob-IT:
I think this is correct and works OK but just a question.
This fix adds a new cell renderer based on a text view to fix editing the notes field on the account tree view. As part of this, when editing the notes field I am using the return key to add new lines and then to complete the edit I am using 'shift and return/keypad enter', Should this be the other way round or some other key combinations?

I have also added gnc_tree_view_add_text_view_column but there is really only one line different to that of gnc_tree_view_add_text so extracted common code to be called from these functions.

### DIFF
--- a/gnucash/gnome-utils/CMakeLists.txt
+++ b/gnucash/gnome-utils/CMakeLists.txt
@@ -51,6 +51,8 @@ set (gnome_utils_SOURCES
   gnc-cell-renderer-popup.c
   gnc-cell-renderer-popup-entry.c
   gnc-cell-renderer-text-flag.c
+  gnc-cell-renderer-text-view.c
+  gnc-cell-view.c
   gnc-combott.c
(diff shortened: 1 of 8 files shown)


christopherlam:
I think Excel uses shift-Enter to insert newline, and Enter to finalise cell?

Bob-IT:
Just tried on my old version of excel, 'alt+enter' adds a line break, 'shift+enter' moves up a cell and enter moves down a cell. On my LibreOffice calc it is 'ctrl+enter'

Bob-IT:
I changed it locally so enter would finalise and 'shift+enter' would add new line. This way matches changing 'Account code' and 'Description' but doesn't feel as natural the other way round but as it probably will not be used much may not matter.

jralls:
Thanks.

PR #684, opened 2020-04-07
[Master] Css changes

Bob-IT:
This PR is mostly about correcting some style classes that I added at the start of version 3.x
Instead of adding widget names I added style contexts based on appropriate structure naming which was wrong. An email awhile ago decided on widget names starting 'gnc-id-' and classes would start 'gnc-class-' followed by some friendly text. 
This PR makes those corrections and also adds some classes where I thought may be useful like 'gnc-class-highlight' which would allow appropriate text to be changed.

As a result of these changes, there will be some disruption when released as the setting in users gtk-3.0.css files will be wrong, mostly it will be just a matter of adding the prefixes above but not always. I have updated the example file with the new entries which I think is liked from the wiki.

It is possible to extract these id's and classes with the following bash file run from the source directory I cobbled together...

#/bin/bash
grep -IrF \"gnc-id- | sed 's/\"name\"/blank/g' > /tmp/id.txt
sed -i 's/[^"]*"\([^"]*\).*/\1/' /tmp/id.txt
grep -IrF \"gnc-class- | sed 's/[^"]*"\([^"]*\).*/\1/' > /tmp/class.txt
sort -u /tmp/class.txt > /tmp/out.txt
sort -u /tmp/id.txt >> /tmp/out.txt
cp /tmp/out.txt class-and-id-list.txt

### DIFF
--- a/doc/gtk-3.0.css
+++ b/doc/gtk-3.0.css
@@ -19,78 +19,254 @@
   the widget type can be configured unless they are named in code.
 */
 
-/* Register sheet font setting */
-.sheet {
-  font: 18px arial, sans-serif;
-}
 
-.sheet calendar {
+/* Application wide font setting */
+* {
   font: 12px arial, sans-serif;
 }
 
 
-/* Register User Colors */
-.header-color {
+/* Scollbar size */
+scrollbar slider {
+    /* Size of the slider */
+    min-width: 20px;
+    min-height: 20px;
+    border-radius: 22px;
+
+    /* Padding around the slider */
+    border: 5px solid transparent;
+}
+
+
+/* Register sheet font setting, use .gnc-class-sheet < gtk3.20 */
+gnc-id-sheet {
+  font: 10px arial, sans-serif;
+}
+
+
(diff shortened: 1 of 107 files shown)


Bob-IT:
For ease, here is the output list...
[get-class-and-ids.txt](https://github.com/Gnucash/gnucash/files/4443768/get-class-and-ids.txt)

jralls:
Thanks!

doc/gtk-3.0.css
```diff
@@ -19,78 +19,254 @@
...
+    min-width: 20px;
+    min-height: 20px;
+    border-radius: 22px;
+
+    /* Padding around the slider */
+    border: 5px solid transparent;
+}
+
+
+/* Register sheet font setting, use .gnc-class-sheet < gtk3.20 */
```
gjanssens, line 41: As we now require Gtk 3.22, I think all references to use classes for older gtk versions like the above can be removed.

Bob-IT: OK will have a look, that also means getting rid of gnucash-310.css, will sort this out during week

PR #685, opened 2020-04-07
Handle future reconcile statement dates

christopherlam:
Here are 3 approaches to handling future reconcile dates. Not sure which one will be most effective.

1. Double-check before accepting a future reconcile date. Uses `goto` because I can't make `do..while` work.
2. ~`xaccAccountGetReconciledBalanceAsOfDate` also includes future reconciled splits likely in error. An ugly hack.~
3. ~Disallow reconcile date > 1week from today.~

### DIFF
--- a/gnucash/gnome/window-reconcile.c
+++ b/gnucash/gnome/window-reconcile.c
@@ -114,6 +114,8 @@ typedef struct _startRecnWindowData
     GtkWidget     *startRecnWindow; /* the startRecnWindow dialog              */
     GtkWidget     *xfer_button;     /* the dialog's interest transfer button   */
     GtkWidget     *date_value;      /* the dialog's ending date field          */
+    GtkWidget     *future_icon;
+    GtkWidget     *future_text;
     GNCAmountEdit *end_value;       /* the dialog's ending balance amount edit */
     gnc_numeric    original_value;  /* the dialog's original ending balance    */
     gboolean       user_set_value;  /* the user changed the ending value       */
@@ -383,9 +385,57 @@ gnc_start_recn_date_changed (GtkWidget *widget, startRecnWindowData *data)
     gnc_numeric new_balance;
     time64 new_date;
 
+    gboolean show_warning = FALSE;
+    gint days_after_today;
+    static const time64 secs_per_day = 86400;
+    static const time64 secs_per_hour = 3600;
+
+    new_date = gnc_date_edit_get_date_end (gde);
+
+    /* Add secs_per_hour to the difference to compensate for the short
+     * day when transitioning from standard to daylight time.
+     */
+    days_after_today = (gnc_time64_get_day_end (new_date) -
+                        gnc_time64_get_today_end () +
+                        secs_per_hour) / secs_per_day;
+
+    if (days_after_today > 0)
+    {
+        /* Translators: This is a ngettext(3) message, %d is the
+           number of days in the future */
+        gchar *str = g_strdup_printf
+            (ngettext ("Statement Date is %d day after today.",
+                       "Statement Date is %d days after today.",
+                       days_after_today),
+             days_after_today);
+
+        /* Translators: This is a ngettext(3) message, %d is the
+           number of days in the future */
+        gchar *tip_start = g_strdup_printf
+            (ngettext ("The statement date you have chosen is %d day in the future.",
+                       "The statement date you have chosen is %d days in the future.",
+                       days_after_today),
+             days_after_today);
+
+        gchar *tip_end = g_strdup (_("This may cause issues for future reconciliation \
+actions on this account. Please double-check this is the date you intended."));
+        gchar *tip = g_strdup_printf ("%s %s", tip_start, tip_end);
+
+        show_warning = TRUE;
+
+        gtk_label_set_text (GTK_LABEL(data->future_text), str);
+        gtk_widget_set_tooltip_text (GTK_WIDGET(data->future_text), tip);
+        g_free (str);
+        g_free (tip_end);
+        g_free (tip_start);
+        g_free (tip);
+    }
+    gtk_widget_set_visible (GTK_WIDGET(data->future_icon), show_warning);
+    gtk_widget_set_visible (GTK_WIDGET(data->future_text), show_warning);
+
     if (data->user_set_value)
         return;
-    new_date = gnc_date_edit_get_date_end (gde);
+
     /* get the balance for the account as of the new date */
     new_balance = gnc_ui_account_get_balance_as_of_date (data->account, new_date,
                   data->include_children);
@@ -771,6 +821,10 @@ startRecnWindow(GtkWidget *parent, Account *account,
         data.end_value = GNC_AMOUNT_EDIT(end_value);
         data.original_value = *new_ending;
         data.user_set_value = FALSE;
+
+        data.future_icon = GTK_WIDGET(gtk_builder_get_object (builder, "future_icon"));
+        data.future_text = GTK_WIDGET(gtk_builder_get_object (builder, "future_text"));
+
         box = GTK_WIDGET(gtk_builder_get_object (builder, "ending_value_box"));
         gtk_box_pack_start(GTK_BOX(box), end_value, TRUE, TRUE, 0);
         label = GTK_WIDGET(gtk_builder_get_object (builder, "end_label"));
@@ -820,6 +874,9 @@ startRecnWindow(GtkWidget *parent, Account *account,
         }
 
         gtk_widget_show_all(dialog);
+
+        gtk_widget_hide (data.future_text);
+        gtk_widget_hide (data.future_icon);
 
         gtk_widget_grab_focus(gnc_amount_edit_gtk_entry
                               (GNC_AMOUNT_EDIT (end_value)));
--- a/gnucash/gnome/window-reconcile.c
+++ b/gnucash/gnome/window-reconcile.c
@@ -1867,6 +1867,43 @@ recnWindowWithBalance (GtkWidget *parent, Account *account, gnc_numeric new_endi
     g_signal_connect (recnData->window, "key_press_event",
                       G_CALLBACK(recn_key_press_cb), recnData);
 
+
+    /* if account has a reconciled split where reconciled_date is
+       later than statement_date, emit a warning into statusbar */
+    {
+        GtkStatusbar *bar = GTK_STATUSBAR (statusbar);
+        guint context = gtk_statusbar_get_context_id (bar, "future_dates");
+        GtkWidget *box = gtk_statusbar_get_message_area (bar);
+        GtkWidget *image = gtk_image_new_from_icon_name
+            ("dialog-warning", GTK_ICON_SIZE_SMALL_TOOLBAR);
+
+        for (GList *n = xaccAccountGetSplitList (account); n; n = n->next)
+        {
+            Split* split = n->data;
+            time64 recn_date = xaccSplitGetDateReconciled (split);
+            if ((xaccSplitGetReconcile (split) != YREC) ||
+                (recn_date <= statement_date))
+                continue;
+
+            PWARN ("split posting_date=%s, recn_date=%s",
+                   qof_print_date (xaccTransGetDate (xaccSplitGetParent (split))),
+                   qof_print_date (recn_date));
+
+            gtk_statusbar_push (bar, context, _("WARNING! Account contains \
+splits whose reconcile date is after statement date. Reconciliation may be \
+difficult."));
+
+            gtk_widget_set_tooltip_text (GTK_WIDGET (bar), _("This account \
+has splits whose Reconciled Date is after this reconciliation statement date. \
+These splits may make reconciliation difficult. If this is the case, you may \
+use Find Transactions to find them, unreconcile, and re-reconcile."));
+
+            gtk_box_pack_start (GTK_BOX(box), image, FALSE, FALSE, 0);
+            gtk_box_reorder_child (GTK_BOX(box), image, 0);
+            break;
+        }
+    }
+
     /* The main area */
     {
         GtkWidget *frame = gtk_frame_new(NULL);
--- a/gnucash/gnome/dialog-find-transactions.c
+++ b/gnucash/gnome/dialog-find-transactions.c
@@ -126,6 +126,8 @@ gnc_ui_find_transactions_dialog_create(GtkWindow *parent, GNCLedgerDisplay * ori
         params = gnc_search_param_prepend (params, N_("Date Posted"), NULL,
                                            type, SPLIT_TRANS, TRANS_DATE_POSTED,
                                            NULL);
+        params = gnc_search_param_prepend (params, N_("Reconciled Date"), NULL,
+                                           type, SPLIT_DATE_RECONCILED, NULL);
         params = gnc_search_param_prepend (params, (num_action
                                                     ? N_("Number/Action")
                                                     : N_("Action")), NULL,
--- a/gnucash/gtkbuilder/window-reconcile.glade
+++ b/gnucash/gtkbuilder/window-reconcile.glade
@@ -219,6 +219,43 @@
                 <property name="position">2</property>
               </packing>
             </child>
+            <child>
+              <object class="GtkBox" id="future_warning">
+                <property name="can_focus">False</property>
+                <property name="halign">center</property>
+                <child>
+                  <object class="GtkImage" id="future_icon">
+                    <property name="visible">True</property>
+                    <property name="can_focus">False</property>
+                    <property name="stock">gtk-dialog-warning</property>
+                    <property name="icon_size">3</property>
+                  </object>
+                  <packing>
+                    <property name="expand">False</property>
+                    <property name="fill">True</property>
+                    <property name="position">0</property>
+                  </packing>
+                </child>
+                <child>
+                  <object class="GtkLabel" id="future_text">
+                    <property name="visible">True</property>
+                    <property name="can_focus">False</property>
+                    <property name="label" translatable="yes">Statement Date is after today</property>
+                    <property name="wrap">True</property>
+                  </object>
+                  <packing>
+                    <property name="expand">False</property>
+                    <property name="fill">True</property>
+                    <property name="position">1</property>
+                  </packing>
+                </child>
+              </object>
+              <packing>
+                <property name="expand">False</property>
+                <property name="fill">True</property>
+                <property name="position">3</property>
+              </packing>
+            </child>
           </object>
           <packing>
             <property name="expand">False</property>


gnucash/gnome/window-reconcile.c
```diff
@@ -382,11 +382,19 @@ gnc_start_recn_date_changed (GtkWidget *widget, startRecnWindowData *data)
 {
     GNCDateEdit *gde = GNC_DATE_EDIT (widget);
     gnc_numeric new_balance;
-    time64 new_date;
+    time64 new_date, cutoff_date;
 
     if (data->user_set_value)
         return;
     new_date = gnc_date_edit_get_date_end (gde);
+    cutoff_date = gnc_time64_get_today_end () + (7*24*60*60);
```
jralls, line 390: I agree with Dale Phurrough that an arbitrary cut-off date is hackish. How about adding another date picker to the reconcile info dialog so that the user can decide?

libgnucash/engine/Account.cpp
```diff
@@ -3403,14 +3403,16 @@ gnc_numeric
...
 
     g_return_val_if_fail(GNC_IS_ACCOUNT(acc), gnc_numeric_zero());
 
     for (GList *node = GET_PRIVATE(acc)->splits; node; node = node->next)
     {
         Split *split = (Split*) node->data;
+        time64 reconciled_date = xaccSplitGetDateReconciled (split);
         if ((xaccSplitGetReconcile (split) == YREC) &&
-            (xaccSplitGetDateReconciled (split) <= date))
+            ((reconciled_date <= date) || (reconciled_date > today)))
```
jralls, line 3415: I don't like this one at all: It means that every past-reconcile will have the future reconcile errors attached to its balance. I think that will make it even harder to find and fix the bad reconcile dates.

gnucash/gnome/window-reconcile.c
```diff
@@ -851,6 +852,11 @@ startRecnWindow(GtkWidget *parent, Account *account,
         if (gnc_reverse_balance(account))
             *new_ending = gnc_numeric_neg (*new_ending);
 
+        if ((*statement_date > gnc_time64_get_today_end ()) &&
+            (!gnc_verify_dialog (GTK_WINDOW (parent), FALSE,
+                                 _("Statement date is in the future. This is very likely to be an error. Using future statement dates is very likely to cause reconciliation difficulties in the future. Please confirm you wish to use a future statement date."))))
+            goto restart_dialog;
+
```
jralls, line 859: This is the wrong place to check. Do it inside the dialog code and disable OK.

This is a fine preventative measure going forward but doesn' t do anything for users whose books already have bad reconcile dates.

jralls:
Another idea, that will directly help users find problems: Add Reconcile Date to the Edit>Find dialog so the user can easily get a register with all of them shown and simply unreconcile them.

gnucash/gnome/window-reconcile.c
derekatkins: > This is the wrong place to check. Do it inside the dialog code and disable OK.
No, disabling OK means that you break a useful feature of being able to reconcile a day or two ahead which I consider important.  I still think allowing 1 week into the future, albeit arbitrary, is perfectly reasonable.  Or perhaps split the difference?  Allow 1 week into the future without a popup, and if it is > 1 week then pop up the warning (with or without "OK").
I can tell you now that I **will** change the code myself to allow future reconciles if they get turned off.

derekatkins:
> Another idea, that will directly help users find problems: Add Reconcile Date to the Edit>Find dialog so the user can easily get a register with all of them shown and simply unreconcile them.

This is a good idea and should be relatively straightforward to add.  I'm actually surprised it's not there already -- probably because the reconcile date isn't actually visible in the UI.  IMHO it should be in a different PR and not this one.

gnucash/gnome/window-reconcile.c
derekatkins: > I agree with Dale Phurrough that an arbitrary cut-off date is hackish. How about adding another date picker to the reconcile info dialog so that the user can decide?

I don't understand what a second date-picker would do here?   The point is to prevent the user from accidentally choosing a day WAY in the future.

gnucash/gnome/window-reconcile.c
christopherlam: I'd still vote for *any* future reconcile date pops a loud warning on OK: "Are you sure?". Only @derekatkins will face needing an extra click... I tried to do it inside the date_changed signal handler but it is either too obnoxious (click after each date change) or not visible enough (a GtkLabel appearing/disappearing).

jralls: The solution to that is to include the "remember this" and "remember this for this session" buttons. I don't know offhand where they live, but @Bob-IT  probably does.

Bob-IT: What you want is to create a dialog and use 'gnc_dialog_run' which use gnc-warnings.h/c and also schema org.gnucash.warnings.gschema.xml.in

Do a search for 'gnc_dialog_run' for examples.

christopherlam:
Better approach
1. `Find Transactions` can now search reconciled_date. It should really be combined with search reconciled=y as well.
2. Add GtkStatusBar warning if account being reconciled has splits whose recdate>today. It should be a separate vbox with GtkLabel and :warning: icon. But I don't know how to add a vbox.
3. Selecting a future statement date is allowable but pops a gnc_dialog_run warning which can be bypassed in the future. This doesn't quite work yet -- gnc_dialog_run (dialog, PREF) - I don't know how to define PREF globally.

```
* 20:42:54 ERROR <gnc.app-utils.gsettings> [gnc_gsettings_get_int()] Invalid key allow-future-reconcile for schema warnings.permanent
* 20:42:54 ERROR <gnc.app-utils.gsettings> [gnc_gsettings_get_int()] Invalid key allow-future-reconcile for schema warnings.temporary
```

Screenshot 1
image

Screenshot 2: see statusbar
image

Screenshot 3
image

Bob-IT:
@christopherlam  Try this...
change "allow-future-reconcile"  to GNC_PREF_WARN_ALLOW_FUTURE_RECONCILE

add this line to gnc-warnings.h
#define GNC_PREF_WARN_ALLOW_FUTURE_RECONCILE "allow-future-reconcile"

add this to gnc-warnings.c

  {GNC_PREF_WARN_ALLOW_FUTURE_RECONCILE,
    "Some text, Short form",
   "Detailed text long form",
  }

and finally add this to org.gnucash.warnings.gschema.xml.in
   <key name="allow-future-reconcile" type="i">
   <default>0</default>
   <summary>Some text, Short form</summary>
   <description>"Detailed text long form"</description>
 </key>
You will need to do the above section twice, once for permanent and the other for temporary, once you look at the file you will get the format.

If you make the above changes, I will pull this and have a look at the warning icon, just updating my VM at the moment.

Bob-IT:
Sorry about that, I forgot those files are auto-generated, I think the preference is generated form what is used in the gschema file, I will have a look soon.

christopherlam:
> because the reconcile date isn't actually visible in the UI

@derekatkins reconcile date is visible when hovering over the 'y' reconcile column in the register.

derekatkins:
> @derekatkins reconcile date is visible when hovering over the 'y' reconcile column in the register.

Oh, I guess they are -- although for some reason popups like that don't actually work for me (but it's a me-thing -- they don't work on any app -- they just flicker).  I'm sure it's an artefact of how my system upgraded from F25 to F29 but I've never really tried to debug it.

Bob-IT:
@christopherlam  The preference is based on the name used in the gschema file.
For other bits in window-reconcile.c
add at top
#include "gnc-warnings.h" 

line 1850 below guint context....

GtkWidget *box = gtk_statusbar_get_message_area (bar);
GtkWidget *image = gtk_image_new_from_icon_name ("dialog-warning", GTK_ICON_SIZE_SMALL_TOOLBAR);

 line below gtk_status_push...
gtk_box_pack_start (GTK_BOX(box), image, FALSE, FALSE, 0);
gtk_box_reorder (GTK_BOX(box), image, 0);


Just a comment, the status bar text seems a bit long, maybe you can shorten it and add the text to a tooltip.

christopherlam:
Thanks @Bob-IT !
Feedback on strings is very welcome.
image

christopherlam:
The commits are now 'merge-ready' assuming strings are acceptable; however still undecided whether to merge, or revert f182d9f91. In any case I think we should add the warning for future-dated reconciliations, and the search by reconciled_date.

christopherlam:
Further food for thought: Instead of reverting the change completely, it could be made optional subject to the book property or book feature "Use strict reconciliation dates for reconciled balances". This will effect Reconciliation only, and be very well documented. Other warnings will remain. Later a scrub may be devised to repair future reconciled_dates, set this feature, and also enable the reconcile-audit-report in https://github.com/Gnucash/gnucash/pull/667#issuecomment-606648840.

gjanssens:
> Further food for thought: Instead of reverting the change completely, it could be made optional subject to the book property or book feature "Use strict reconciliation dates for reconciled balances". This will effect Reconciliation only, and be very well documented.

Personally I think that would be over-engineering. And besides book features should only be used in case incompatible data format changes are introduced (new data objects or different interpretations of existing data objects). I don't think that is the case here. You only want to restrict or regulate data entry more tightly in this case.

> Other warnings will remain. Later a scrub may be devised to repair future reconciled_dates, set this feature, and also enable the reconcile-audit-report in [#667 (comment)](https://github.com/Gnucash/gnucash/pull/667#issuecomment-606648840).

The warnings are a good addition. A scrub would be helpful as well if we find a way to fix dates that's not arbitrary or a way for users to enter one.

gjanssens:
> Better approach
>     3. Selecting a future statement date is allowable but pops a gnc_dialog_run warning which can be bypassed in the future.

I am a bit late in the discussion. However I'm not sure a gnc_dialog_run warning is the best approach here. It has a few issues IMO.

1. What we want to achieve is inform the user the date chosen is in the future, yet allow the user to continue anyway if s/he chooses to do so anyway. The gnc_dialog_run warning can only guarantee that the first time. If the user then ticks the box to no longer see this warning in the future, we effectively no longer inform the user in the future even if the entered date really is a bad type (like way in the future). Take @derekatkins ' case (which I'm not convinced yet is a valid one but still a good illustration for this). He deliberately wants to be able to reconcile one or at most two days in the future. The pop-up warning will quickly annoy him so I'm pretty sure he would turn it off. If he then on a future reconcile really fat-fingers and mistypes a date way into the future, that would still break his following reconciles yet he wouldn't be informed about them anymore. So my preference would go to a method to inform the user that can't be turned off.

2. Popping up warnings is considered disruptive. And a dialog popping up a dialog is even worse. On the other hand information that can't be turned off (as I suggested earlier) should should not be disruptive as that would lead to a lot of user irritation. So instead of a popup warning  I would propose to add a text label to the reconcile info window itself that appears the moment the date entered is in the future. It could state something like 

> Note the statement date you have chosen is <x days> in the future. This may cause issues for future reconciliation actions on this account. Please double-check this is the date you intended.

3. If @derekatkins ' case can be dealt with otherwise (as my suggestion on the mailing list) we would not have a valid use case for future dates at all. I have yet to receive a bank or credit card statement with a future date on it. In that case the message in point 2 could be slightly changed to note this is not allowed and disable the Ok button as long as the date entered is in the future.

christopherlam:
GtkLabel, warning, and tooltip
image

christopherlam:
Remaining Issues:
1. `days_after_today` is calculated as `(end_date - today) / 86400` then cast to `gint`. there's probably a better way eg. `time64tojulian(end) - time64tojulian(today)`. The simple division may break during DST changes.
2. I don't know how to internationalise `day(s)` in the strings.

gnucash/gnome/window-reconcile.c
```diff
@@ -385,14 +386,35 @@ gnc_start_recn_date_changed (GtkWidget *widget, startRecnWindowData *data)
     gnc_numeric new_balance;
     time64 new_date;
 
-    GtkWidget *future_box;
-    gboolean date_is_future;
+    GtkWidget *future_icon, *future_text;
```
Bob-IT, line 389: Some suggestions...
change this to gboolean hide_warning = TRUE;

gnucash/gnome/window-reconcile.c
```diff
@@ -385,14 +386,35 @@ gnc_start_recn_date_changed (GtkWidget *widget, startRecnWindowData *data)
...
+        gtk_label_set_text (GTK_LABEL (future_text), str);
+        gtk_widget_set_tooltip_text (future_text, tip);
+        gtk_widget_show (future_text);
+        gtk_widget_show (future_icon);
+    }
+    else
+    {
+        gtk_widget_hide (future_text);
+        gtk_widget_hide (future_icon);
+    }
```
Bob-IT, line 417: change from 400 to this...

        /* Translators: This is a ngettext(3) message, %d is the number of days in the future */
        gchar *str = g_strdup_printf (ngettext ("Statement Date is %d day after today.",
                                                "Statement Date is %d days after today.",
                                                days_after_today),
                                                days_after_today);

         gchar *tip_end = g_strdup (_("This may cause issues for future reconciliation actions on this account. \
 Please double-check this is the date you intended."));

        /* Translators: This is a ngettext(3) message, %d is the number of days in the future */
        gchar *tip_start = g_strdup_printf (ngettext ("The statement date you have chosen is %d day in the future.",
                                                      "The statement date you have chosen is %d days in the future.",
                                                      days_after_today),
                                                      days_after_today);

        gchar *tip = g_strdup_printf ("%s\n%s", tip_start, tip_end);

        hide_warning = FALSE;

        gtk_label_set_text (GTK_LABEL(data->future_text), str);
        gtk_widget_set_tooltip_text (GTK_WIDGET(data->future_text), tip);
        g_free (str);
        g_free (tip_end);
        g_free (tip_start);
        g_free (tip); 
    }
    gtk_widget_set_visible (GTK_WIDGET(data->future_icon), !hide_warning);
    gtk_widget_set_visible (GTK_WIDGET(data->future_text), !hide_warning);

Bob-IT:
>     2\. I don't know how to internationalise `day(s)` in the strings.

See above for 2, for 1 I would not think it mattered for the warning label

christopherlam:
Based on latest findings that `reconciled_date` is unreliable, I think now we could amend the reconcile window warning:

if the account has splits whose `reconciled_date` is later than `statement_date` then show warning "This account has splits reconciled later than statement date. This may make reconciliation difficult. Either you are reconciling a past statement, or a past reconciliation was set with an incorrect statement date."

jralls:
With f182d9f9 reverted is it really true that reconciliation will be made difficult?

christopherlam:
> With [f182d9f](https://github.com/Gnucash/gnucash/commit/f182d9f9121130b49320e82810a924b73d69a6b9) reverted is it really true that reconciliation will be made difficult?

For the average user, not anymore. For people trying to reconcile an old statement, yes.

gnucash/gnome/window-reconcile.c
```diff
@@ -384,9 +387,51 @@ gnc_start_recn_date_changed (GtkWidget *widget, startRecnWindowData *data)
...
     time64 new_date;
 
+    gboolean show_warning = FALSE;
+    GtkWidget *future_icon, *future_text;
+    gint days_after_today;
+
+    new_date = gnc_date_edit_get_date_end (gde);
+
+    days_after_today = round ((gnc_time64_get_day_end (new_date) -
+                               gnc_time64_get_today_end ()) / 86400);
```
jralls, line 397: `round()` is for floating-point. This compiles only because the `int64_t` is implicitly cast to a `double`, rounded, and then the returned double is implicitly cast back to an `int64_t`.

In C integer division always returns an integer of the same type as the larger operand and the remainder is always discarded. The `%` operator returns the remainder, but in general there won't be any. The exception would be crossing a DST boundary, where going from standard time to daylight time will lose a whole day because the interval is less than 86400 seconds. To deal with that case I suggest adding 3600 to the difference before dividing.

Also, it's bad form to stick a raw constant into calculation. Instead add a constant decl to the function's declaration block (or put at the top of the file if you need the constant in more than one place):
```
static const time64 secs_per_day = 86400;
static const time64 secs_per_hour = 3600;
...
/* Add secs_per_hour to the difference to compensate for the short day when
 * transitioning from standard to daylight time.
 */
days_after_today = (gnc_time64_get_day_end (new_date) - 
                                    gnc_time64_get_today_end + secs_per_hour) / secs_per_day;
```

gnucash/import-export/aqb/gnc-ab-utils.c
```diff
@@ -1062,13 +1062,13 @@ bal_accountinfo_cb(AB_IMEXPORTER_ACCOUNTINFO *element, gpointer user_data)
 #else
             time64 secs = GWEN_Time_toTime_t(ti);
 #endif
-            booked_tt = gnc_time64_get_day_neutral(secs);
+            booked_tt = gnc_time64_get_day_end(secs);
```
jralls, line 1065: This is probably not what caused the problem with [Bug 797673](https://bugs.gnucash.org/show_bug.cgi?id=797673) because as I noted in [comment 17](https://bugs.gnucash.org/show_bug.cgi?id=797673#c17) the reconciliation date on the two errant splits was before neutral time.

One could argue that reconciling to a downloaded balance should reflect the time of the download: The balance might change later in the day and the user might do another reconcile. The alternative I suppose would be to use end-of-day, resulting in a new reconciled balance for what becomes as far as GnuCash is concerned the same reconcile.

christopherlam: I'll forgo this bugfix because I don't understand it enough.

gnucash/gnome/window-reconcile.c
christopherlam: the secs_adjustment may be up to 3hours sometimes... see Brazil in https://en.wikipedia.org/wiki/Daylight_saving_time, so, to be very general I think best adjust by `secs_per_day/2`

jralls: Read that more carefully: It says "In early 2008, central Brazil was one, two, or three hours ahead of the eastern United States, depending on the date." That's because Brazil's DST wasn't (and probably isn't) aligned with the US's. Fortunately for Brazilians it doesn't mean that a DST shift in Brazil can be 3 hours in a month's time.

We're not looking for TZ shifts here either: The two times are generated within a few microseconds of each other and aside from Dr. Who nobody's changing timezones in a fraction of a second.

christopherlam: Ok :) this is very much small print.

christopherlam:
Ok rebased to be merge-ready. No breaking change this time; adds search by `reconciled_date`, adds warning if `statement_date>today`, and warn if `account->splits->reconciled_date` exists after `statement_date`.

christopherlam:
adds a small PWARN if reconcile_window detects a split whereby recn_date>statement_date. can be removed or changed to PINFO if wished.

christopherlam:
No breaking change. Adds 2 warnings only in UI, and find by reconciled date.

christopherlam:
PS I realized after push that I didn't specifically test the 32-bit windows build especially time64->gint conversion... please check reconcile still works prior to release?

jralls:
Don't worry about that: You're not converting a time64 to an int, you're putting the difference between two time64s divided by 86400 into an int. I'll leave it as an exercise for you to figure out how long that interval would have to be to overflow an int.

christopherlam:
Ok it works 👍
image

christopherlam:
> I'll leave it as an exercise for you to figure out how long that interval would have to be to overflow an int

The exact answer is of course MAXINT days, where 32-bit = 2147483647 days = 5883516 years from now.

My query came from the observation that the formula starts with 64-bit `time64` numbers, performs arithmetic and division, results in a number that may or may not be in range for `int` nor a whole number, and C will agree to convert to an `int` including truncating fractions if it has to.

jralls:
> results in a number that may or may not be in range for int nor a whole number

No, it always results in a whole number because both divisor and dividend are integer types. That part is implemented in hardware. If one of the operands was a floating-point type then the compiler would generate instructions to use a floating-point divide instead and different circuits in the CPU would perform the arithmetic. (In the 8086 processor used in 1980's-era PCs the floating point circuitry was on a separate chip called the 8087 floating-point unit. It was optional, so compilers would generate code to detect it and branch to either floating-point arithmetic instructions if it was present or use integer arithmetic with code like our Rational class if it wasn't.)

The only way it can go out of range for an int is if one of the dates is invalid because we use INT64_MAX for that. Otherwise the valid date range is 1400-9999, a mere 8599 years whose number of days will easily fit into an int32_t.

PR #687, opened 2020-04-12
Bug 797114 - Fixing an SX due to deleted account stuck in an error loop

jeanlaroche:
Added call to qof_instance_get_destroying() so we can ignore a non-existing account if we're in the process of destroying the split.


jeanlaroche:
Yes, confirmed fixed! Thanks!

jralls:
Thanks!

PR #688, opened 2020-04-13
Bug 797605 - RFE: Customer/Vendor/Employee Overview should also have …

christopherlam:
…"Process Payment" toolbar

image

### DIFF
--- a/gnucash/gnome/gnc-plugin-page-owner-tree.c
+++ b/gnucash/gnome/gnc-plugin-page-owner-tree.c
@@ -46,6 +46,7 @@
 #include "dialog-employee.h"
 #include "dialog-invoice.h"
 #include "dialog-job.h"
+#include "dialog-payment.h"
 
 #include "gncOwner.h"
 #include "dialog-utils.h"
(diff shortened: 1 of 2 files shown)


Bob-IT:
I have not pull this yet but noticed that the three objects have the same text so do you really need three duplicates or maybe you forgot to change some text?

gnucash/gnome/gnc-plugin-page-owner-tree.c
```diff
@@ -314,6 +346,9 @@ static action_owners_struct action_owners[] =
     { "OTVendorReportAction",          GNC_OWNER_VENDOR },
     { "OTCustomerReportAction",        GNC_OWNER_CUSTOMER },
     { "OTEmployeeReportAction",        GNC_OWNER_EMPLOYEE },
+    { "OTProcessVendorPaymentAction",  GNC_OWNER_VENDOR },
+    { "OTProcessCustomerPaymentAction",GNC_OWNER_CUSTOMER },
+    { "OTProcessEmployeePaymentAction",GNC_OWNER_EMPLOYEE },
```
christopherlam, line 351: @Bob-IT I think this is why... action for each type of owners

Bob-IT:
OK I pulled this and if you are not requiring different text for each owner then it can just be one entry, just do not add any thing to the section above and use a general string for the action like 'OTProcessPaymentAction' unless I am missing something.

christopherlam:
@Bob-IT Correct as usual :+1:

jralls:
@christopherlam please revert this and apply it to master instead.

christopherlam:
I'll wait till the next maint->master merge then apply to master.

jralls:
Merged & pushed. Go for it.

PR #689, opened 2020-04-14
Fix building with link-time-optimization.

wenottingham:
Patch from Jeff Law (law@redhat.com):

Rename one instance of struct _iterate to struct _qofid_iterate to avoid ODR problems with LTO.

(downstream Fedora patch)

### DIFF
--- a/libgnucash/engine/qofid.cpp
+++ b/libgnucash/engine/qofid.cpp
@@ -303,7 +303,7 @@ qof_collection_set_data (QofCollection *col, gpointer user_data)
 
 /* =============================================================== */
 
-struct _iterate
+struct _qofid_iterate
 {
     QofInstanceForeachCB      fcn;
     gpointer                data;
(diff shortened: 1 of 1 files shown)


libgnucash/engine/qofid.cpp
```diff
@@ -303,7 +303,7 @@ qof_collection_set_data (QofCollection *col, gpointer user_data)
 
 /* =============================================================== */
 
-struct _iterate
+struct __iterate
```
bhardwajs, line 306: I think we shouldn't use this name here: https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#nl8-use-a-consistent-naming-style

jralls: WTF?1 @bhardwajs is right. That's not just core guidelines, identifiers with double underscores are reserved *[in the standard](https://en.cppreference.com/w/cpp/language/identifiers) and using them is UB. So how the heck is UB going to improve LTO?

jralls: Also [reserved by gcc](https://www.gnu.org/software/libc/manual/html_node/Reserved-Names.html).

wenottingham: Not sure why he chose that as the rename. I assume the rationale for the patch is that there's a `struct _iterate` defined in both `qofid.cpp` and `qofbook.cpp`. I'll ask.

wenottingham: Rebased with a different name.

jralls:
Thanks!

PR #690, opened 2020-04-14
Bug 797115 - Can't 're-activate' an expired SX Fix 3 problems with th…

jeanlaroche:
…e scheduled transactions calculations
While reproducing the bug I found quite a few problems with the scheduled transactions: inconsistencies between what's displayed in the main dialog and in the edit dialog, mistakes by 1 day in the calculation of the first instance, lack of removal of marks in the calendar in cases where no instances can possibly occur.
These were mostly corner cases (playing with the start date, the previous occurrence date, the end date, etc), but things seem to be working well now.
I haven't run the tests, I'll check later to ensure they all pass.

### DIFF
--- a/gnucash/gnome-utils/gnc-dense-cal.c
+++ b/gnucash/gnome-utils/gnc-dense-cal.c
@@ -1939,10 +1939,13 @@ gdc_model_added_cb(GncDenseCalModel *model, guint added_tag, gpointer user_data)
 static void
 gdc_model_update_cb(GncDenseCalModel *model, guint update_tag, gpointer user_data)
 {
-    GncDenseCal *cal = GNC_DENSE_CAL(user_data);
-    g_debug("gdc_model_update_cb update for tag [%d]\n", update_tag);
-    gdc_mark_remove(cal, update_tag, FALSE);
-    gdc_add_tag_markings(cal, update_tag);
+    GncDenseCal *cal = GNC_DENSE_CAL (user_data);
+    gint num_marks = 0;
+    g_debug ("gdc_model_update_cb update for tag [%d]\n", update_tag);
+    num_marks = gnc_dense_cal_model_get_instance_count (cal->model, update_tag);
+    // We need to redraw if there are no mark, to ensure they're all erased.
+    gdc_mark_remove (cal, update_tag, num_marks==0);
+    gdc_add_tag_markings (cal, update_tag);
 
 }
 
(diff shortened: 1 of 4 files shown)


christopherlam:
It may be easier to write tests that would fail on previous code, and fix code to make the tests pass? 😉  IMHO both test and fix may be in the same commit. Alternatively, each corner case gets its own test and commit.

e.g. 447de7f64dc644be9b0684c5beaebe2db90cc3c9 and d711cc35f80ff0d483fc174cca17fd3ea4b56b64

jeanlaroche:
> It may be easier to write tests that would fail on previous code, and fix code to make the tests pass? 😉 IMHO both test and fix may be in the same commit. Alternatively, each corner case gets its own test and commit.
> 
> e.g. [447de7f](https://github.com/Gnucash/gnucash/commit/447de7f64dc644be9b0684c5beaebe2db90cc3c9) and [d711cc3](https://github.com/Gnucash/gnucash/commit/d711cc35f80ff0d483fc174cca17fd3ea4b56b64)

I agree, I'll look into writing such tests.... I don't feel super comfortable fixing bugs that nobody's reported on! (but me)

christopherlam:
> I agree, I'll look into writing such tests.... I don't feel super comfortable fixing bugs that nobody's reported on! (but me)

On the contrary... clear and simple tests will be useful to confirm that your fixes are correct and necessary.

jeanlaroche:
I think you misunderstood me, I was agreeing with you! :)

gjanssens:
Ok, not code related but code management.

When you write commit messages, please leave an empty line between the first line ('message subject') and any additional description you add. Failing to to so would result in these odd commit messages that consists of your bug title and the beginning of the extra description.

This then ripples through into the PR title and message body.

Both are distracting.

jeanlaroche:
I took a look at extending the unit test, but this is going to be problematic: The unit test only tests functionality in recurrence.c. There was no problem in that file, however there were several problems in dialog-sx-editor.c, gnc-dense-cal.c and SchedXaction.c
I don't see an easy way to create a simple test to check for example SchedXaction.c: you'd have to create a SchedXaction structure, which has lots of fields that would have to be properly set... 
It seems to me like a lot of work, and I'm not sure how much bang for the buck we'd get out of that. I have to say I'm balking at investing much time other than the time it too to fix the 3 issues that I could see were wrong... Hope that's OK with you guys.
Jean

christopherlam:
I think UI bugfixes will always be difficult to formally test.  So I think the fairest approach is to illustrate the exact fixes in the commit message exactly -- a sequence of actions to create the circumstances of the fix. Alternatively the bugzilla report may be enough to recreate the bug (& the fix!)

christopherlam:
Ok I cannot understand the C code but I can confirm your 3 cases described in comment are fixed :+1:

jeanlaroche:
Awesome thanks!

jralls:
Merged after editing the commit message to remove the bug reference. This doesn't fix that bug, it's just that @jeanlaroche found these issues while he was investigating it.
Thanks!

PR #691, opened 2020-04-14
Adding lines missing from fix for Bug 797196

thetedmunds:
Adding definition of gnc:amort_balance() that got lost in the original pull-request for #490.
(Using guile syntax proposed by @christopherlam)

### DIFF
--- a/libgnucash/app-utils/fin.scm
+++ b/libgnucash/app-utils/fin.scm
@@ -298,6 +298,22 @@
 
 ;; Calculate the balance remaining after the nth payment
 ;; (n must be greater than or equal to zero)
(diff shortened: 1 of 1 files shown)


christopherlam:
With #692 merged in, now you can return `#f` or any non-number to signify invalid input. Care to rewrite?

thetedmunds:
Should that be done in this patch, or would it be better to merge this, then do another patch for all the `-1` -> `#f` changes in the whole fin.scm file?

christopherlam:
Or you can add to this branch and it'll all be merged in one go.

christopherlam:
Thanks I've amended to return `#f` instead of `-1`. Any further changes please submit a new PR.

PR #692, opened 2020-04-14
[gnc-exp-parser] don't crash when gnc:fn returns non-number

christopherlam:
If gnc:fn returns anything other than a finite real number, abort rather than crash. Precursor to #691

test by modifying `fin.scm`:

```scheme
(define (gnc:foobar val) (and (positive? val) val))
```

and in SX formula try both `foobar(20)` and `foobar(-20)` -- now gnc:function can return anything, and will only be processed if it returns a number.


christopherlam:
Safe enough. Don't know how to upgrade test-exp-parser though.

PR #693, opened 2020-04-14
L10N:fi: Update translations and add glossary

kohtala:
I was looking if GnuCash would fit my accounting needs, but found the Finnish translations in bad shape. This is not very complete, but thought I better make this pull request to make the effort visible to others.

I have not yet done any accounting using GnuCash and if I actually get to it I'm sure there will be more translations and corrections.

While incomplete, I think this is an improvement and good enough for merge.

### DIFF
--- a/po/fi.po
+++ b/po/fi.po
@@ -1,18 +1,19 @@
-# gnucash/fi.
+# Finnish translation of GnuCash.
 # This file is distributed under the same license as the gnucash package.
 # Copyright (C) 2008 gnucash project
+# Marko Kohtala <marko.kohtala@gmail.com>, 2020.
 # Tuomo Hartikainen <tth@harski.org>, 2017.
 # Tuomo Kohvakka <tuomo.kohvakka@iki.fi>, 2008.
 #
 msgid ""
 msgstr ""
-"Project-Id-Version: GnuCash 2.6.15\n"
+"Project-Id-Version: GnuCash 3.10\n"
 "Report-Msgid-Bugs-To: https://bugs.gnucash.org/enter_bug.cgi?"
 "product=GnuCash&component=Translations\n"
 "POT-Creation-Date: 2020-03-28 12:48-0700\n"
-"PO-Revision-Date: 2017-03-13 20:59+0200\n"
-"Last-Translator: Tuomo Hartikainen <tth@harski.org>\n"
-"Language-Team: UNKNOWN\n"
+"PO-Revision-Date: 2020-04-20 10:44+0300\n"
+"Last-Translator: Marko Kohtala <marko.kohtala@gmail.com>\n"
+"Language-Team: \n"
 "Language: fi\n"
 "MIME-Version: 1.0\n"
 "Content-Type: text/plain; charset=UTF-8\n"
@@ -439,7 +440,7 @@ msgid ""
 "the manual under the Help menu."
 msgstr ""
 "GnuCashin verkko-oppaassa on paljon hyödyllistä tietoa. Pääset siihen "
-"käsiksi \"Apua\"-valikon alta."
+"käsiksi \"Ohje\"-valikon alta."
 
 #: doc/tip_of_the_day.list.c:4
 msgid ""
@@ -448,8 +449,8 @@ msgid ""
 "org"
 msgstr ""
 "GnuCashin tekijät on helppo tavoittaa. Useiden sähköpostilistojen lisäksi "
-"voit jutella heidän kanssaan IRC:ssä. Liity #gnucash -kanavalle irc.gnome."
-"org-verkossa."
+"voit jutella heidän kanssaan IRC:ssä! Liity #gnucash -kanavalle irc.gnome."
+"org:ssa."
 
 #: doc/tip_of_the_day.list.c:8
 msgid ""
@@ -460,7 +461,7 @@ msgid ""
 msgstr ""
 "Voit helposti tuoda vanhan kirjanpidon Quickenistä, MS Moneystä tai muista "
 "ohjelmista jotka osaavat viedä tiedot QIF- tai OFX-tiedostoiksi. Avaa "
-"\"Tiedosto\"-valikko, valitse \"Tuo\"-alavalikko ja klikkaa QIF- tai OFX-"
+"Tiedosto-valikko, valitse Tuo-alavalikko ja klikkaa QIF- tai OFX-"
 "tiedoston tuonti. Seuraa tuonnin ohjeita jatkaaksesi."
 
 #: doc/tip_of_the_day.list.c:13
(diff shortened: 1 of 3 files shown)


jralls:
Thanks for trying, but there are some problems:
```
po/fi.po:18315: number of format specifications in 'msgid' and 'msgstr' does not match
po/fi.po:22786: format specifications in 'msgid' and 'msgstr' are not equivalent
po/fi.po:23444: format specifications in 'msgid' and 'msgstr' are not equivalent
po/fi.po:23449: format specifications in 'msgid' and 'msgstr' are not equivalent
msgfmt: found 4 fatal errors
2753 translated messages, 4 fuzzy translations, 2666 untranslated messages.
```
The first one is because you added an extra `%u` parameter. The function call supplies only one parameter.

In the last three you correctly figured out that `~a` in Scheme is the same as `%s` in C, but since those strings will be passed to Scheme `format` the substitution is incorrect.

kohtala:
Thanks. I did read quite a lot of documentation, but missed some verifications. I now updated the commit with fixes.

kohtala:
Thanks for review. I did not have time yesterday. I think I'll have time to update the commit if not tonight during the weekend.

po/glossary/fi.po
```diff
@@ -0,0 +1,789 @@
+# SOME DESCRIPTIVE TITLE.
+# Copyright (C) YEAR Free Software Foundation, Inc.
+# FIRST AUTHOR <EMAIL@ADDRESS>, YEAR.
```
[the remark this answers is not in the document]
kohtala: Other glossary files had the Free Software Foundation, so I left it in there. I updated slightly the main po/fi.po file header as well for the latest commit.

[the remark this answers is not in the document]
fellen: See options [--copyright-holder](https://www.gnu.org/software/gettext/manual/gettext.html#index-_002d_002dcopyright_002dholder_002c-xgettext-option) and [--foreign-user](https://www.gnu.org/software/gettext/manual/gettext.html#index-_002d_002dforeign_002duser_002c-xgettext-option).

The question is not, what other files for unclear historical reasons have. Old gettext versions had no option to disable the FSF line.

The questions is: do you have or want a contract with them?

po/fi.po
```diff
@@ -1,18 +1,19 @@
...
 # Tuomo Kohvakka <tuomo.kohvakka@iki.fi>, 2008.
 #
 msgid ""
 msgstr ""
-"Project-Id-Version: GnuCash 2.6.15\n"
+"Project-Id-Version: GnuCash 3.10\n"
 "Report-Msgid-Bugs-To: https://bugs.gnucash.org/enter_bug.cgi?"
 "product=GnuCash&component=Translations\n"
 "POT-Creation-Date: 2020-03-28 12:48-0700\n"
 "PO-Revision-Date: 2017-03-13 20:59+0200\n"
```
fellen, line 14: Update the date, please.

po/fi.po
```diff
@@ -1,18 +1,19 @@
...
-"Last-Translator: Tuomo Hartikainen <tth@harski.org>\n"
```
fellen, line 14: Had you talked with Tuomo?

po/fi.po
```diff
@@ -1,18 +1,19 @@
...
-"Language-Team: UNKNOWN\n"
+"Last-Translator: Marko Kohtala <marko.kohtala@gmail.com>\n"
+"Language-Team: Finnish <fi@li.org>\n"
```
fellen, line 16: Is this the email address of a real language team like e.g. for the translation project translation-team-fi@lists.sourceforge.net?
If you are not associated to a team, insert "none" or leave it empty.

po/fi.po
```diff
@@ -1430,7 +1429,7 @@ msgstr "Yritys"
 #: gnucash/gnome/dialog-job.c:594 gnucash/gnome/dialog-vendor.c:742
 #: gnucash/gnome-utils/gnc-tree-view-owner.c:349
 msgid "ID #"
-msgstr "ID nro"
+msgstr "ID #"
```
fellen, line 1432: The english use "#" as abbreviation for "Number" in forms, but in other laguages we use "Nr.", "No.", ...

po/fi.po
kohtala: I had contacted @harski in 2018. He told he got some other work and was happy someone would be continuing. I understood he would not mind. I just tonight remembered to drop him a note I've created this pull request in case he has time to take a look.

po/fi.po
kohtala: This is just what gettext proposed. I'll change. I had done a quick search and found https://www.gnu.org/software/gettext/manual/gettext.html#Trans-Intro-0 mention it, but now looking it again there is a note the section is outdated.

kohtala:
I noticed the glossary/fi.po was not up to date with gnc-glossary.txt. Updated it and the Project-Id-Version. Also other comments applied there as well.

po/fi.po
```diff
@@ -720,9 +721,8 @@ msgstr "Aikaväli"
 
 #: gnucash/gnome/assistant-acct-period.c:590
 #: gnucash/gtkbuilder/dialog-book-close.glade:86
-#, fuzzy
 msgid "Closing Date"
-msgstr "Vientipäivä"
```
harski, line 725: Why simply delete this, or was it a mistake? If you think it's wrong, could you suggest a better translation?

po/fi.po
harski: I'm cool with this, I haven't done any further translations since we last spoke.

harski:
Great job @kohtala!

The diff is huge and I didn't go through all of it, but I have one note: `Expense account` shouldn't be translated as `Matkalasku` as it could be other type of expense too (you buy danishes to a housing cooperative meeting). Maybe `Kulutosite` would be more appropriate?

po/fi.po
kohtala: In some cases I thought there was misleading translation. Having untranslated string is better than having misleading translation. In this case "Vientipäivä" would mean export date.

You review is appreciated. I am not sure I understood this right. What I understood was that this would be the last day of the books. It could be "Tilinpäätöspäivä". I just was not sure at the time I looked at it.

kohtala:
> Great job @kohtala!

Thanks.

> The diff is huge and I didn't go through all of it, but I have one note: `Expense account` shouldn't be translated as `Matkalasku` as it could be other type of expense too (you buy danishes to a housing cooperative meeting). Maybe `Kulutosite` would be more appropriate?

Hmm. I see `Expense account` translated as `Kulutili`. `Expense Vouchers` is translated as `Matkalaskut`. `Voucher` is also translated as `Matkalasku`, as glossary tells "The terms 'Voucher' and 'Expense Voucher' are used interchangeably". But I have a bit of suspicion if there are places where `Voucher` could be used to mean `Kulutosite` or maybe `Kuitti`?

I saw `Expense Vouchers` to be used to track business expense debt from employer to employee (to have proof the payment is not a taxable salary payment). Under different employers I've always created `Matkalasku` for this regardless if any travel was involved. `Kulutosite` is more generic and does not require an employee to be involved, which I saw to go out of context here. A slightly different case is eg. purchases using business credit card where business never pays any money to employee, but to the credit company, and `Kuitti` would work here. In either case, if it is necessary to track the employee, I do not see any confusion using `Matkalasku`. I believe that is what some businesses do anyway with other software.

po/fi.po
fellen: > In some cases I thought there was misleading translation. Having untranslated string is better than having misleading translation. 

In such a case you can insert the `, fuzzy`flag. Then the translation is not used.

kohtala: Thanks. I'll fuzzy them if similar comes up. I changed this to `Tilinpäätöspäivä`. I see it being used meaning the date you have to value your assets (stock prices etc.) when closing books, which I understood this field to mean.

I did remove quite a few other translations that were already fuzzy as I thought they were created by msgmerge and too far from correct to be helpful. I rather view them as untranslated when I get time to think for a correct translation.

fellen:
I have committed this part.
Thank you!
Feel free to send updates in a new PR.

kohtala:
Thank you. I only now noticed you had commented on the FSF Copyright. You are right. If FSF was not involved, they need not be mentioned as copyright holder. I'll fix it in a future PR.

fellen:
OK

PR #694, opened 2020-04-16
General UI - improve consistency of menu actions

christopherlam:
Adds UI to get reversed transaction date. Current code acts immediately and doesn't jump to new split. This commit offers a chance to cancel.

This code also tries to jump to the reversing transaction, but fails. Not sure why this jump doesn't actually work.

### DIFF
--- a/gnucash/gnome-utils/dialog-dup-trans.c
+++ b/gnucash/gnome-utils/dialog-dup-trans.c
@@ -183,7 +183,8 @@ gnc_dup_trans_dialog_create (GtkWidget * parent, DupTransDialog *dt_dialog,
 }
 
 static gboolean
-gnc_dup_trans_dialog_internal (GtkWidget * parent, const char* title,
+gnc_dup_trans_dialog_internal (GtkWidget * parent,
+                               const char* window_title, const char* title,
                                gboolean show_date, time64 *date_p,
                                GDate *gdate_p, const char *num, char **out_num,
                                const char *tnum, char **out_tnum,
@@ -217,6 +218,9 @@ gnc_dup_trans_dialog_internal (GtkWidget * parent, const char* title,
         entry = gde->date_entry;
         gtk_widget_grab_focus (entry);
     }
+
+    if (window_title)
+        gtk_window_set_title (GTK_WINDOW (dt_dialog->dialog), window_title);
 
     if (title)
     {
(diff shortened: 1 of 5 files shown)


gnucash/gnome/gnc-plugin-budget.c
```diff
@@ -54,6 +54,8 @@ static void gnc_plugin_budget_cmd_open_budget (GtkAction *action,
         GncMainWindowActionData *data);
 static void gnc_plugin_budget_cmd_copy_budget (GtkAction *action,
         GncMainWindowActionData *data);
+static void gnc_plugin_budget_cmd_delete_budget (GtkAction *action,
+                                                 GncMainWindowActionData *data);
```
Bob-IT, line 58: I think I would be inclined to align GncMain... like the others above.

gnucash/gnome/gnc-plugin-budget.c
```diff
@@ -233,6 +241,24 @@ gnc_plugin_budget_cmd_copy_budget (GtkAction *action,
...
+gnc_plugin_budget_cmd_delete_budget (GtkAction *action,
+                                     GncMainWindowActionData *user_data)
+{
+    QofBook *book = gnc_get_current_book ();
+    QofCollection *col = qof_book_get_collection (book, GNC_ID_BUDGET);
+    GncBudget *bgt;
+
+    if (qof_collection_count (col) == 0) return;
+
+    g_return_if_fail (user_data != NULL);
```
Bob-IT, line 255: The return test I think should be below GncBudget *bgt;
and move the get collection to be below the g_return_if_fail, if you are testing a parameter, there is no point in doing any thing unless it passes and you could even do this ...

    if (qof_collection_count (qof_book_get_collection (book, GNC_ID_BUDGET)) == 0)
        return;

gnucash/gnome-utils/dialog-dup-trans.c
```diff
@@ -184,6 +184,7 @@ gnc_dup_trans_dialog_create (GtkWidget * parent, DupTransDialog *dt_dialog,
 
 static gboolean
 gnc_dup_trans_dialog_internal (GtkWidget * parent, const char* title,
+                               char* window_title,
```
Bob-IT, line 187: maybe this should be a const char*

gnucash/gnome-utils/dialog-dup-trans.h
```diff
@@ -73,4 +73,8 @@ gnc_dup_trans_dialog_gdate (GtkWidget * parent, GDate *gdate_p,
 gboolean
 gnc_dup_date_dialog (GtkWidget * parent, const char* title, GDate *date);
 
+gboolean
```
Bob-IT, line 76: Needs some text explaining this, see gnc_dup_date_dialog

Bob-IT:
As for the jump not working, it was not working before your changes. Not sure why it was using the qof_event_suspend/resume but if you change to gnc_suspend/resume_gui_refresh fixes the jumping.

christopherlam:
Thanks @Bob-IT now things work as expected :) (irrelevant issue: the Delete Budget menu item is different from other budget menu items: it has a not totally unwelcome but unexpected icon!).

Now that there is a Delete Budget menu, I wonder if we should delete the Budget Editor's `Delete` toolbar button?

PR #695, opened 2020-04-17
U+R and R changed to U+C and C

DaveC49:
The labels in the import matcher dialog and message s in the code imply that a transaction split to an account is reconciled when its status is actually only set to cleared 'c'. The labels and reference in the code are changed to U+C and C to better reflect the actual actions on import.

These are changes to the codse so it now matches the documentation in the Help manual


jralls:
No. This is a UI change that belongs in GnuCash 4.0 and it's already merged there.

PR #697, opened 2020-04-18
Add ability to import multiple OFX files at one time

jeanlaroche:
This adds support for importing multiple OFX files in one shot. The one thing that might be objectionable to some is the change to gnc_file_dialog returning void* instead of char*. This is something that's done quite frequently in C and in gnucash in particular (in many callback functions etc), so I don't find it particularly objectionable.
If you guys object, I can add a GSList** parameter to gnc_file_dialog, and change all existing calls to send NULL.

### DIFF
--- a/gnucash/gnome-utils/gnc-file.c
+++ b/gnucash/gnome-utils/gnc-file.c
@@ -60,29 +60,15 @@ static QofLogModule log_module = GNC_MOD_GUI;
 static GNCShutdownCB shutdown_cb = NULL;
 static gint save_in_progress = 0;
 
-
-/********************************************************************\
- * gnc_file_dialog                                                  *
- *   Pops up a file selection dialog (either a "Save As" or an      *
- *   "Open"), and returns the name of the file the user selected.   *
- *   (This function does not return until the user selects a file   *
- *   or presses "Cancel" or the window manager destroy button)      *
- *                                                                  *
- * Args:   title        - the title of the window                   *
- *         filters      - list of GtkFileFilters to use, will be    *
-                          freed automatically                       *
- *         default_dir  - start the chooser in this directory       *
- *         type         - what type of dialog (open, save, etc.)    *
- * Return: containing the name of the file the user selected        *
-\********************************************************************/
-
-char *
-gnc_file_dialog (GtkWindow *parent,
-                 const char * title,
-                 GList * filters,
-                 const char * starting_dir,
-                 GNCFileDialogType type
-                )
+// gnc_file_dialog_int is used both by gnc_file_dialog and gnc_file_dialog_multi
+static GSList *
+gnc_file_dialog_int (GtkWindow *parent,
+                     const char * title,
+                     GList * filters,
+                     const char * starting_dir,
+                     GNCFileDialogType type,
+                     gboolean multi
+                     )
 {
     GtkWidget *file_box;
     const char *internal_name;
@@ -91,6 +77,7 @@ gnc_file_dialog (GtkWindow *parent,
     const gchar *ok_icon = NULL;
     GtkFileChooserAction action = GTK_FILE_CHOOSER_ACTION_OPEN;
     gint response;
+    GSList* file_name_list = NULL;
 
     ENTER(" ");
 
@@ -130,6 +117,9 @@ gnc_file_dialog (GtkWindow *parent,
                    action,
                    _("_Cancel"), GTK_RESPONSE_CANCEL,
                    NULL);
+    if (multi)
+        gtk_file_chooser_set_select_multiple (GTK_FILE_CHOOSER (file_box), TRUE);
+    
     if (ok_icon)
         gnc_gtk_dialog_add_button(file_box, okbutton, ok_icon, GTK_RESPONSE_ACCEPT);
     else
@@ -174,23 +164,85 @@ gnc_file_dialog (GtkWindow *parent,
 
     if (response == GTK_RESPONSE_ACCEPT)
     {
-        /* look for constructs like postgres://foo */
-        internal_name = gtk_file_chooser_get_uri(GTK_FILE_CHOOSER (file_box));
-        if (internal_name != NULL)
+        if (multi)
         {
-            if (strstr (internal_name, "file://") == internal_name)
+            file_name_list = gtk_file_chooser_get_filenames (GTK_FILE_CHOOSER (file_box));
+        }
+        else
+        {
+            /* look for constructs like postgres://foo */
+            internal_name = gtk_file_chooser_get_uri(GTK_FILE_CHOOSER (file_box));
+            if (internal_name != NULL)
             {
-                /* nope, a local file name */
-                internal_name = gtk_file_chooser_get_filename(GTK_FILE_CHOOSER (file_box));
+                if (strstr (internal_name, "file://") == internal_name)
+                {
+                    /* nope, a local file name */
+                    internal_name = gtk_file_chooser_get_filename(GTK_FILE_CHOOSER (file_box));
+                }
+                file_name = g_strdup(internal_name);
             }
-            file_name = g_strdup(internal_name);
+            file_name_list = g_slist_append (file_name_list, file_name);
         }
     }
     gtk_widget_destroy(GTK_WIDGET(file_box));
     LEAVE("%s", file_name ? file_name : "(null)");
+    return file_name_list;
+}
+
+/********************************************************************\
+ * gnc_file_dialog                                                  *
+ *   Pops up a file selection dialog (either a "Save As" or an      *
+ *   "Open"), and returns the name of the file the user selected.   *
+ *   (This function does not return until the user selects a file   *
+ *   or presses "Cancel" or the window manager destroy button)      *
+ *                                                                  *
+ * Args:   title        - the title of the window                   *
+ *         filters      - list of GtkFileFilters to use, will be    *
+ *                        freed automatically                       *
+ *         default_dir  - start the chooser in this directory       *
+ *         type         - what type of dialog (open, save, etc.)    *
+ * Return: containing the name of the file the user selected        *
+ \********************************************************************/
+char *
+gnc_file_dialog (GtkWindow *parent,
+                 const char * title,
+                 GList * filters,
+                 const char * starting_dir,
+                 GNCFileDialogType type
+                 )
+{
+    gchar* file_name = NULL;
+    GSList* ret = gnc_file_dialog_int (parent, title, filters, starting_dir, type, FALSE);
+    if (ret)
+        file_name = g_strdup (ret->data);
+    g_slist_free_full (ret, g_free);
     return file_name;
 }
 
+/********************************************************************\
+ * gnc_file_dialog_multi                                            *
+ *   Pops up a file selection dialog (either a "Save As" or an      *
+ *   "Open"), and returns the name of the files the user selected.  *
+ *   Similar to gnc_file_dialog with allowing multi-file selection  *
+ *                                                                  *
+ * Args:   title        - the title of the window                   *
+ *         filters      - list of GtkFileFilters to use, will be    *
+ *                        freed automatically                       *
+ *         default_dir  - start the chooser in this directory       *
+ *         type         - what type of dialog (open, save, etc.)    *
+ * Return: GList containing the names of the selected files         *
+ \********************************************************************/
+
+GSList *
+gnc_file_dialog_multi (GtkWindow *parent,
+                       const char * title,
+                       GList * filters,
+                       const char * starting_dir,
+                       GNCFileDialogType type
+                       )
+{
+    return gnc_file_dialog_int (parent, title, filters, starting_dir, type, TRUE);
+}
 
 gboolean
 show_session_error (GtkWindow *parent,
--- a/gnucash/gnome-utils/gnc-file.h
+++ b/gnucash/gnome-utils/gnc-file.h
@@ -149,6 +149,12 @@ char * gnc_file_dialog (GtkWindow *parent,
                         const char * starting_dir,
                         GNCFileDialogType type);
 
+GSList * gnc_file_dialog_multi (GtkWindow *parent,
+                                const char * title,
+                                GList * filters,
+                                const char * starting_dir,
+                                GNCFileDialogType type);
+
 gboolean gnc_file_open_file (GtkWindow *parent,
                              const char *filename,
                              gboolean open_readonly);
--- a/gnucash/gtkbuilder/dialog-import.glade
+++ b/gnucash/gtkbuilder/dialog-import.glade
@@ -945,6 +945,21 @@
             <property name="position">1</property>
           </packing>
         </child>
+        <child>
+            <object class="GtkCheckButton" id="reconcile_after_close_button">
+                <property name="label" translatable="yes">Reconcile after match</property>
+                <property name="visible">True</property>
+                <property name="can_focus">True</property>
+                <property name="receives_default">False</property>
+                <property name="use_underline">True</property>
+                <property name="draw_indicator">True</property>
+            </object>
+            <packing>
+                <property name="expand">False</property>
+                <property name="fill">True</property>
+                <property name="position">2</property>
+            </packing>
+        </child>
       </object>
       <packing>
         <property name="expand">False</property>
--- a/gnucash/import-export/import-main-matcher.c
+++ b/gnucash/import-export/import-main-matcher.c
@@ -69,6 +69,7 @@ struct _main_matcher_info
     GtkTreeViewColumn *account_column;
     GtkWidget         *show_account_column;
     GtkWidget         *show_matched_info;
+    GtkWidget         *reconcile_after_close;
     gboolean add_toggled;   // flag to indicate that add has been toggled to stop selection
     gint id;
 };
@@ -196,6 +197,8 @@ gboolean gnc_gen_trans_list_empty(GNCImportMainMatcher *info)
 void gnc_gen_trans_list_show_all(GNCImportMainMatcher *info)
 {
     gtk_widget_show_all (GTK_WIDGET (info->main_widget));
+    // By default, do not show this check box.
+    gnc_gen_trans_list_show_reconcile_after_close (info, FALSE, FALSE);
 }
 
 void
@@ -985,6 +988,9 @@ GNCImportMainMatcher *gnc_gen_trans_list_new (GtkWidget *parent,
     g_signal_connect (G_OBJECT(info->show_matched_info), "toggled",
                       G_CALLBACK(show_matched_info_toggled_cb), info);
 
+    // Create the checkbox, but do not show it by default.
+    info->reconcile_after_close = GTK_WIDGET(gtk_builder_get_object (builder, "reconcile_after_close_button"));
+
     show_update = gnc_import_Settings_get_action_update_enabled (info->user_settings);
     gnc_gen_trans_init_view (info, all_from_same_account, show_update);
     heading_label = GTK_WIDGET(gtk_builder_get_object (builder, "heading_label"));
@@ -1006,7 +1012,7 @@ GNCImportMainMatcher *gnc_gen_trans_list_new (GtkWidget *parent,
     gtk_builder_connect_signals_full (builder, gnc_builder_connect_full_func, info);
 
     g_object_unref (G_OBJECT(builder));
-    
+
     // Register this UI, it needs to be closed when the session is closed.
     info->id = gnc_register_gui_component (IMPORT_MAIN_MATCHER_CM_CLASS,
                                            NULL, /* no refresh handler */
@@ -1068,6 +1074,8 @@ GNCImportMainMatcher * gnc_gen_trans_assist_new (GtkWidget *parent,
     info->show_matched_info = GTK_WIDGET(gtk_builder_get_object (builder, "show_matched_info_button"));
     g_signal_connect (G_OBJECT(info->show_matched_info), "toggled",
                       G_CALLBACK(show_matched_info_toggled_cb), info);
+
+    info->reconcile_after_close = GTK_WIDGET(gtk_builder_get_object (builder, "reconcile_when_close_button"));
 
     show_update = gnc_import_Settings_get_action_update_enabled (info->user_settings);
     gnc_gen_trans_init_view (info, all_from_same_account, show_update);
@@ -1447,6 +1455,18 @@ void gnc_gen_trans_list_add_trans (GNCImportMainMatcher *gui, Transaction *trans
     gnc_gen_trans_list_add_trans_with_ref_id (gui, trans, 0);
     return;
 }/* end gnc_import_add_trans() */
+
+void gnc_gen_trans_list_show_reconcile_after_close(GNCImportMainMatcher *info, gboolean reconcile_after_close, gboolean active)
+{
+    gtk_widget_set_visible (info->reconcile_after_close,reconcile_after_close);
+    gtk_toggle_button_set_active (GTK_TOGGLE_BUTTON (info->reconcile_after_close), active);
+}
+
+GtkWidget*
+gnc_gen_trans_list_get_reconcile_widget(GNCImportMainMatcher *info)
+{
+    return info->reconcile_after_close;
+}
 
 void gnc_gen_trans_list_add_trans_with_ref_id (GNCImportMainMatcher *gui, Transaction *trans, guint32 ref_id)
 {
--- a/gnucash/import-export/ofx/gnc-ofx-import.c
+++ b/gnucash/import-export/ofx/gnc-ofx-import.c
@@ -67,15 +67,19 @@ static Account *ofx_parent_account = NULL;
 // Structure we use to gather information about statement balance/account etc.
 typedef struct _ofx_info
 {
-    gint num_trans_processed;
-    GSList* statement;
     GtkWindow* parent;
     GNCImportMainMatcher *gnc_ofx_importer_gui;
     Account *last_import_account;
     Account *last_investment_account;
     Account *last_income_account;
-    GList *created_commodites   ;
+    gint num_trans_processed;               // Number of transactions processed
+    struct OfxStatementData* statement;     // Statement, if any
+    gboolean run_reconcile;                 // If TRUE the reconcile window is opened after matching.
+    GList *created_commodites;
+    GSList* file_list;                      // List of OFX files to import
 } ofx_info ;
+
+GList *ofx_created_commodites = NULL;
 
 /*
 int ofx_proc_status_cb(struct OfxStatusData data)
@@ -118,13 +122,11 @@ set_associated_income_account(Account* investment_account,
     xaccAccountCommitEdit(investment_account);
 }
 
-int ofx_proc_statement_cb(struct OfxStatementData data,
-                          void *statement_user_data);
-int ofx_proc_security_cb(const struct OfxSecurityData data,
-                         void *security_user_data);
+int ofx_proc_statement_cb (struct OfxStatementData data, void * statement_user_data);
+int ofx_proc_security_cb (const struct OfxSecurityData data, void * security_user_data);
 int ofx_proc_transaction_cb (struct OfxTransactionData data, void *user_data);
-int ofx_proc_account_cb(struct OfxAccountData data, void * account_user_data);
-static double ofx_get_investment_amount(const struct OfxTransactionData* data);
+int ofx_proc_account_cb (struct OfxAccountData data, void * account_user_data);
+static double ofx_get_investment_amount (const struct OfxTransactionData* data);
 
 static const gchar *gnc_ofx_ttype_to_string(TransactionType t)
 {
@@ -917,11 +919,10 @@ int ofx_proc_transaction_cb(struct OfxTransactionData data, void *user_data)
 int ofx_proc_statement_cb (struct OfxStatementData data, void * statement_user_data)
 {
     ofx_info* info = (ofx_info*) statement_user_data;
-    struct OfxStatementData* statement = g_new (struct OfxStatementData, 1);
-    *statement = data;
-    info->statement = g_slist_append (info->statement, statement);
+    info->statement = g_new (struct OfxStatementData, 1);
+    *info->statement = data;
     return 0;
-}//end ofx_proc_statement()
+}
 
 
 int ofx_proc_account_cb(struct OfxAccountData data, void * account_user_data)
@@ -1003,22 +1004,23 @@ int ofx_proc_account_cb(struct OfxAccountData data, void * account_user_data)
 
         gnc_utf8_strip_invalid(data.account_name);
         gnc_utf8_strip_invalid(data.account_id);
-        account_description = g_strdup_printf( /* This string is a default account
+        account_description = g_strdup_printf (/* This string is a default account
                                                   name. It MUST NOT contain the
                                                   character ':' anywhere in it or
                                                   in any translation.  */
-                                  "%s \"%s\"",
-                                  account_type_name,
-                                  data.account_name);
-        /* use the info->parent as import-matcher is not displayed yet */
-        account = gnc_import_select_account (GTK_WIDGET(info->parent),
+                                               "%s \"%s\"",
+                                               account_type_name,
+                                               data.account_name);
+        account = gnc_import_select_account (gnc_gen_trans_list_widget(info->gnc_ofx_importer_gui),
                                              data.account_id, 1,
                                              account_description, default_commodity,
-                                             default_type, info->last_import_account, NULL);
+                                             default_type, NULL, NULL);
+
         if (account)
         {
             info->last_import_account = account;
         }
+
         g_free(account_description);
     }
     else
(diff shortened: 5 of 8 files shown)


jralls:
Well, it is true that `void*` is widely used in C for type erasure, but it's usually for parameters rather than for return types. This approach is dangerously hackish, but your proposed alternative of adding a GSList** parameter isn't much better.

Why not return a GSList* and change the 10 other uses to get only the first node's data?

jeanlaroche:
> 
> Why not return a GSList* and change the 10 other uses to get only the first node's data?

You prefer that to passing a GList** set to NULL? In any case, I"m open to either, I don't really mind. The 10 other uses also have to free the GList, but they already had to free the char I imagine...

gjanssens:
The other approach is to write two file dialog functions

* `gnc_file_dialog` as is returning a `char*`
* `gnc_file_dialog_multi` returning a `GSList*`

`gnc_file_dialog_multi` would be the function as you have rewritten except it will return a `GSList*`. gnc_file_dialog would simply be a wrapper around it that takes the returned `GSList*`, copies the first `char*` from it takes care of freeing the GSList and returns the copied `char*`.

jeanlaroche:
I like that approach best ^^ . I'll implement that. Thanks for the suggestion!

gnucash/gnome-utils/gnc-file.c
```diff
@@ -130,6 +117,8 @@ gnc_file_dialog (GtkWindow *parent,
                    action,
                    _("_Cancel"), GTK_RESPONSE_CANCEL,
                    NULL);
+    if (multi)
+        gtk_file_chooser_set_select_multiple (GTK_FILE_CHOOSER (file_box), TRUE);
```
Bob-IT, line 121: I would add a blank line after this.

gnucash/gnome-utils/gnc-file.c
```diff
@@ -175,22 +164,78 @@ gnc_file_dialog (GtkWindow *parent,
     if (response == GTK_RESPONSE_ACCEPT)
     {
         /* look for constructs like postgres://foo */
-        internal_name = gtk_file_chooser_get_uri(GTK_FILE_CHOOSER (file_box));
-        if (internal_name != NULL)
+        if (!multi)
```
Bob-IT, line 167: I think it would be better to have if (multi), so swap the order

gnucash/gnome-utils/gnc-file.h
```diff
@@ -149,6 +149,12 @@ char * gnc_file_dialog (GtkWindow *parent,
                         const char * starting_dir,
                         GNCFileDialogType type);
 
+GSList * gnc_file_dialog_multi (GtkWindow *parent,
+                        const char * title,
+                        GList * filters,
+                        const char * starting_dir,
+                        GNCFileDialogType type);
+
```
Bob-IT, line 157: Can you align the parameters to that of GtkWindow

gnucash/gnome-utils/gnc-file.c
```diff
@@ -175,22 +164,78 @@ gnc_file_dialog (GtkWindow *parent,
...
+ *   or presses "Cancel" or the window manager destroy button)      *
+ *                                                                  *
+ * Args:   title        - the title of the window                   *
+ *         filters      - list of GtkFileFilters to use, will be    *
+ *                        freed automatically                       *
+ *         default_dir  - start the chooser in this directory       *
+ *         type         - what type of dialog (open, save, etc.)    *
+ * Return: containing the name of the file the user selected        *
+ \********************************************************************/
+
```
Bob-IT, line 205: Remove blank line

gnucash/gnome-utils/gnc-file.c
```diff
@@ -175,22 +164,78 @@ gnc_file_dialog (GtkWindow *parent,
...
     return file_name;
 }
 
+/********************************************************************\
+ * gnc_file_dialog_multi                                            *
+ *   Identical to gnc_file_dialog but allows selecting multiple     *
+ *   files                                                          *
+ * Return: GList containing the names of the selected files         *
+ \********************************************************************/
+
```
Bob-IT, line 228: I would be inclined to add all the parameters so you do nit have to go looking for them and the remove blank line.

gnucash/import-export/ofx/gnc-ofx-import.c
```diff
@@ -1091,22 +1094,22 @@ void gnc_file_ofx_import (GtkWindow *parent)
         selected_filename = conv_name;
 #endif
 
-        DEBUG("Opening selected file");
+        DEBUG("Opening selected file(s)");
         num_trans_processed = 0;
-        libofx_proc_file(libofx_context, selected_filename, AUTODETECT);
-        // Now would be a good time to see whether the view has anything in it!
+        for (iter = selected_filenames; iter; iter = iter->next)
+            libofx_proc_file (libofx_context, iter->data, AUTODETECT);
```
Bob-IT, line 1100: Should we not be doing the conv_name procedure on all the file names in the list?

gnucash/import-export/ofx/gnc-ofx-import.c
```diff
@@ -1091,22 +1094,22 @@ void gnc_file_ofx_import (GtkWindow *parent)
...
             gnc_gen_trans_list_delete (gnc_ofx_importer_gui);
             if(num_trans_processed)
-                gnc_info_dialog(parent,_("OFX file imported, %d transactions processed, no transactions to match"),num_trans_processed);
+                gnc_info_dialog (parent, _("OFX file(s) imported, %d transactions processed, no transactions to match"), num_trans_processed);
         }
         else
         {
             gnc_gen_trans_list_show_all(gnc_ofx_importer_gui);
         }
-        g_free(selected_filename);
```
Bob-IT, line 1108: Why remove the g_free here, maybe I am misunderstanding this.

gnucash/import-export/ofx/gnc-ofx-import.c
jeanlaroche: You're right.

gnucash/import-export/ofx/gnc-ofx-import.c
jeanlaroche: My bad, g_slist_free() by default does not free the strings themselves.

gjanssens: This construct is slightly complex (no blame, it was already before you started working on it). I would like this better if the GSList functions would be used to fully manage the list. I mean if list memory is freed with `g_slist_free_full`. Also do this as soon as you no longer need the memory, which is right after the for loop.

The non-Windows case is straightforward. It can work directly with `iter->data` (via `selected_filename` as is set right before the Windows block. With a `g_slist_free_full` right after the for loop, memory is handled cleanly.

The Windows case needs special treatment because we need to localize the name. Which means we create a variable that needs to be freed. In the current Windows block, just reset `selected_filename` to the result of `g_win32_locale_filename_from_utf8`. No need for a temporary variable and don't free anything there. Instead at the very end of the for loop, add another Windows only block and `g_free` the `selected_filename` variable in there.

That makes the flow on each platform more consistent:
* non-windows: the only memory management is done by `g_slist_free_full`.
* on windows: the list is memory-managed by the same `g_slist_free_full` and the localized variable by a `g_free`.

gnucash/gnome-utils/gnc-file.c
```diff
@@ -60,29 +60,15 @@ static QofLogModule log_module = GNC_MOD_GUI;
...
-char *
-gnc_file_dialog (GtkWindow *parent,
-                 const char * title,
-                 GList * filters,
-                 const char * starting_dir,
-                 GNCFileDialogType type
-                )
+// gnc_file_dialog_aux is used both by gnc_file_dialog and gnc_file_dialog_multi
+static GSList *
+gnc_file_dialog_aux (GtkWindow *parent,
```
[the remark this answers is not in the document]
jeanlaroche: Done

gnucash/gnome-utils/gnc-file.c
```diff
@@ -174,23 +164,85 @@ gnc_file_dialog (GtkWindow *parent,
...
+                 GList * filters,
+                 const char * starting_dir,
+                 GNCFileDialogType type
+                 )
+{
+    gchar* file_name = NULL;
+    GSList* ret = gnc_file_dialog_aux (parent, title, filters, starting_dir, type, FALSE);
+    if (ret)
+        file_name = g_strdup (ret->data);
+    g_slist_free (ret);
```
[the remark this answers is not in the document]
jeanlaroche: Done

christopherlam:
The commit title is now not very useful :)

Bob-IT:
Firstly, the import-matcher changes should really be in a separate commit.
Secondly if you cancel at the file dialogue, the app crashes. This is because the file list will be empty so have added suggestions in line...

gnucash/import-export/ofx/gnc-ofx-import.c
```diff
@@ -1047,11 +1045,12 @@ void gnc_file_ofx_import (GtkWindow *parent)
     extern int ofx_INFO_msg;
     extern int ofx_STATUS_msg;
     char *selected_filename;
```
Bob-IT, line 1047: Move this line to you for loop.

gnucash/import-export/ofx/gnc-ofx-import.c
```diff
@@ -1069,12 +1068,13 @@ void gnc_file_ofx_import (GtkWindow *parent)
...
-                                        filters,
-                                        default_dir,
-                                        GNC_FILE_DIALOG_IMPORT);
+    selected_filenames = gnc_file_dialog_multi (parent,
+                                               _("Select one or multiple OFX/QFX file(s) to process"),
+                                               filters,
+                                               default_dir,
+                                               GNC_FILE_DIALOG_IMPORT);
     g_free(default_dir);
+    selected_filename = selected_filenames->data;
```
Bob-IT, line 1077: Remove this line.

gnucash/import-export/ofx/gnc-ofx-import.c
```diff
@@ -1069,12 +1068,13 @@ void gnc_file_ofx_import (GtkWindow *parent)
...
 
     if (selected_filename != NULL)
```
Bob-IT, line 1079: Change this to..
    if (g_slist_length (selected_filenames) > 0)

gnucash/import-export/ofx/gnc-ofx-import.c
```diff
@@ -1091,7 +1091,7 @@ void gnc_file_ofx_import (GtkWindow *parent)
         DEBUG("Filename found: %s", selected_filename);
```
Bob-IT, line 1091: Move this to for loop.

gnucash/import-export/ofx/gnc-ofx-import.c
```diff
@@ -1104,20 +1104,25 @@ void gnc_file_ofx_import (GtkWindow *parent)
...
+        DEBUG("Opening selected file(s)");
+        for (iter = selected_filenames; iter; iter = iter->next)
+        {
+            selected_filename = iter->data;
 #ifdef G_OS_WIN32
-        conv_name = g_win32_locale_filename_from_utf8(selected_filename);
-        g_free(selected_filename);
-        selected_filename = conv_name;
+            selected_filename = g_win32_locale_filename_from_utf8 (selected_filename);
 #endif
```
Bob-IT, line 1113: Change this to...
#ifdef G_OS_WIN32
            char *selected_filename = g_win32_locale_filename_from_utf8 (iter->data);
#else
            char *selected_filename = g_strdup (iter->data);
#endif
            DEBUG("Filename found: %s", selected_filename);

gnucash/import-export/ofx/gnc-ofx-import.c
```diff
@@ -1104,20 +1104,25 @@ void gnc_file_ofx_import (GtkWindow *parent)
...
-
-        DEBUG("Opening selected file");
-        libofx_proc_file(libofx_context, selected_filename, AUTODETECT);
+            libofx_proc_file (libofx_context, selected_filename, AUTODETECT);
+#ifdef G_OS_WIN32
+            g_free (selected_filename);
+#endif
```
Bob-IT, line 1117: Change this to...
g_free (selected_filename);

gnucash/import-export/import-main-matcher.c
```diff
@@ -1014,12 +1016,12 @@ GNCImportMainMatcher *gnc_gen_trans_list_new (GtkWidget *parent,
     g_object_unref (G_OBJECT(builder));
     
     // Register this UI, it needs to be closed when the session is closed.
-    id = gnc_register_gui_component (IMPORT_MAIN_MATCHER_CM_CLASS,
+    info->id = gnc_register_gui_component (IMPORT_MAIN_MATCHER_CM_CLASS,
```
Bob-IT, line 1019: Realign the parameters below this line and also remove the 'gint id' on line 956

jeanlaroche:
Thanks for the review and the suggested code changes, and sorry for missing the crash on cancel ... I separated the commits into two.

gnucash/import-export/ofx/gnc-ofx-import.c
jralls: An empty GSlist is a `nullptr`, so no point in counting the list. `if (selected_filenames)` does everything needed with the least ceremony.

Bob-IT: @jeanlaroche Did you see John's suggestion, it's simpler and better than my change.

jeanlaroche: Yep, saw that. Will change it. :)

Bob-IT:
Also you do not need conv_name and look closely at line 1106

jeanlaroche:
> Also you do not need conv_name and look closely at line 1106

Crap! Sorry about that. Will fix as well...

Bob-IT:
I made a comment in #PR710 to separate out the import-main-matcher fix.

jeanlaroche:
I found an issue with the current solution: If you import 2 OFX from banks A and B, with a transfer from bank A to bank B, the transfer appears in both OFX, and as a result appears *twice* in the matcher.
If instead the OFX are imported one after the other, the transaction is added when the first OFX is imported, and then flagged as previously matched when the second OFX is imported.
So that's a weakness of the current solution (which consists of combining *all* the OFX data into a single list of transactions that are matched in one shot).
One solution would be for GC to import the selected OFX one by one, each time opening a new matching window. This would essentially do what you do by hand when importing multiple OFX files in a row, saving you a few clicks. 
One issue with that solution is it wouldn't be easy to allow the user to cancel *all* imports if they change their minds: they would have to go through all imports and click cancel each time.

On the other hand, detecting the double transaction would require re-using the code form the matcher, and I'm not sure how easy that would be.

What do you guys think?
Jean

jralls:
> One issue with that solution is it wouldn't be easy to allow the user to cancel all imports if they change their minds: they would have to go through all matching windows and click cancel each time.

That shouldn't be too hard to address, connect to the match window's `response` signal and if it's "cancel" pop a dialog box offering to cancel all of the remaining imports.

> Another solution would be for GC to open as many transaction matching windows are there are OFX files. I believe if you complete the match in one window, the other window where the corresponding transaction occurs would then be updated to show a previously reconciled transaction (need to check that).

Not only do I think that's not the case, I think implementing it would be difficult.

christopherlam:
> I found an issue with the current solution: If you import 2 OFX from banks A and B, with a transfer from bank A to bank B, the transfer appears in both OFX, and as a result appears _twice_ in the matcher.
> If instead the OFX are imported one after the other, the transaction is added when the first OFX is imported, and then flagged as previously matched when the second OFX is imported.
> So that's a weakness of the current solution (which consists of combining _all_ the OFX data into a single list of transactions that are matched in one shot).

Remember that OFX obligatorily have a `FITID` field which is unique for each transaction. You could engineer a delete-duplicate check somewhere.

jeanlaroche:
> 
> 
> > I found an issue with the current solution: If you import 2 OFX from banks A and B, with a transfer from bank A to bank B, the transfer appears in both OFX, and as a result appears _twice_ in the matcher.
> > If instead the OFX are imported one after the other, the transaction is added when the first OFX is imported, and then flagged as previously matched when the second OFX is imported.
> > So that's a weakness of the current solution (which consists of combining _all_ the OFX data into a single list of transactions that are matched in one shot).
> 
> Remember that OFX obligatorily have a `FITID` field which is unique for each transaction. You could engineer a delete-duplicate check somewhere.

I this case, the OFX files come from 2 different banks, so they will not have matching FITID. It's just that they represent the two ends of a unique transaction, and so only one should appear in the match window (as a transfer between the two accounts in question)

jeanlaroche:
Next try! I changed the process so multiple files are imported one after the other (instead of simultaneously). I think this is a lot better. It takes care of the case where corresponding transactions are present in two OFX files.

To make this happen, I connected to the closing of the matching dialog, so the next file can be processed. The list of files to process is stored in the info structure, and the list is cleared after the last OFX file is imported.

gnucash/import-export/ofx/gnc-ofx-import.c
```diff
@@ -70,9 +70,10 @@ static Account *ofx_parent_account = NULL;
 // Structure we use to gather information about statement balance/account etc.
 typedef struct _ofx_info
 {
-    gint num_trans_processed;
-    GSList* statement;
     GtkWindow* parent;
+    gint num_trans_processed;               // Number of transactions processed
+    GSList* file_list;                      // List of OFX files to import
+    GSList* file_iter;                      // Current OFX file in list above.
```
Bob-IT, line 76: Do you need this, why not just remove the processed file from the list when done.

gnucash/import-export/ofx/gnc-ofx-import.c
```diff
@@ -1036,6 +1026,101 @@ double ofx_get_investment_amount(const struct OfxTransactionData* data)
     }
 }
 
+// Forward declaration, required because several static functions depend on one-another.
+static void
+gnc_file_ofx_import_int (ofx_info* info);
```
Bob-IT, line 1031: If you need this I think it should be at the start and I not sire on function name, why not something like gnc_file_ofx_import_process_file

gnucash/import-export/ofx/gnc-ofx-import.c
```diff
@@ -1036,6 +1026,101 @@ double ofx_get_investment_amount(const struct OfxTransactionData* data)
...
+            /* FIXME: Present some result window about the newly created
+             * commodities */
+            g_warning ("Created %d new commodities during import", g_list_length (ofx_created_commodites));
+            g_list_free (ofx_created_commodites);
+            ofx_created_commodites = NULL;
+        }
+        else
+        {
+            //g_warning("No new commodities created");
+        }
```
Bob-IT, line 1060: Can not see the point of leaving this.

gnucash/import-export/ofx/gnc-ofx-import.c
```diff
@@ -1036,6 +1026,101 @@ double ofx_get_investment_amount(const struct OfxTransactionData* data)
...
+{
+    ofx_info* info = (ofx_info*) user_data;
+    gnc_ofx_process_next_file (dialog, info);
+}
+
+// Aux function to process the OFX file in info->file_iter
+static void
+gnc_file_ofx_import_int (ofx_info* info)
+{
+    LibofxContextPtr libofx_context = libofx_get_new_context();
```
Bob-IT, line 1077: Do you need to create a new context for each file?
If yes, I see no libofx_free_context
Maybe you can use the unused one in gnc_file_ofx_import and move the callback setup to there also ?

gnucash/import-export/ofx/gnc-ofx-import.c
```diff
@@ -1036,6 +1026,101 @@ double ofx_get_investment_amount(const struct OfxTransactionData* data)
...
+    {
+        // Done with the previous OFX file, process the next one if any.
+        info->file_iter = info->file_iter->next;
+        gnc_file_ofx_import_int (info);
+    }
+    else
+    {
+        // No more OFX file to process, final cleanup.
+        g_slist_free_full (info->file_list, g_free);
+        g_free (info);
```
Bob-IT, line 1048: This seems more logical to be added at the end of gnc_file_ofx_import

Bob-IT:
Just an observation, if I select two files, select OK on the first matcher and the second file has nothing to do you get the message 'OFX file(s) imported, 2 transactions processed, no transactions to match' but behind it is the matcher dialog for the first file, this looks confusing, the first matcher should of been closed.

gnucash/import-export/ofx/gnc-ofx-import.c
jeanlaroche: Don't you need the list to delete the GSList at the very end? Or can you save the next pointer, and delete the current one? Does that work exactly the same?

gnucash/import-export/ofx/gnc-ofx-import.c
jeanlaroche: I'll move to the head. I was told that aux functions typically have an _int suffix in GC (which seems strange to me). I'm all for changing it to gnc_file_ofx_import_process_file

gnucash/import-export/ofx/gnc-ofx-import.c
jeanlaroche: Me neither. If you don't I'll remove it.

gnucash/import-export/ofx/gnc-ofx-import.c
jeanlaroche: There was no libofx_free_context() in the original code, I don't think (unless somehow I removed it)? I'll try to reuse the same one, that's cleaner.

gnucash/import-export/ofx/gnc-ofx-import.c
jeanlaroche: mmm, all the calls in gnc_file_ofx_import() are non-blocking so the end of the function is reached instantly before info is done being used. So you can't free it there. You have to free it after the last time it's used, correct?

gnucash/import-export/ofx/gnc-ofx-import.c
Bob-IT: I think you may be referring to the file changes, probably not the right words, in that case there were to two functions declared in the h file that really just provided the right parameters to the common _int function.

jeanlaroche:
> Just an observation, if I select two files, select OK on the first matcher and the second file has nothing to do you get the message 'OFX file(s) imported, 2 transactions processed, no transactions to match' but behind it is the matcher dialog for the first file, this looks confusing, the first matcher should of been closed.
> 

That's because the code that processes the next file is in the callback for the "response" signal of the first matcher, which happens after you press OK but before the window is closed. 
I could try to connect to the "close" signal instead, I think that would fix it, but then I wouldn't get the user's answer in the match dialog (OK/Cancel) which for now does not matter, but if eventually I add "reconcile after match" it will matter. 
I could connect to both, get the answer with the "response" signal, and continue with the imports in the "close" signal. Seems like over-engineering to me, but I'm OK doing it if you think it's worth it. 
Also if you see a simpler solution let me know. BTW GC does not use multithreading, which I think is a good call! (but it would help in situations like this).

gnucash/import-export/ofx/gnc-ofx-import.c
jeanlaroche: Yes, that's it. I'll change the name then.

gnucash/import-export/ofx/gnc-ofx-import.c
Bob-IT: I was thinking of just removing the filename from the list with g_slist_remove, I have not tested this but was just an idea.

jeanlaroche: I'll give that a shot. BTW, it does not look like I see memory leaks if I forget to free something (in Xcode). Is there an easy way to turn that on?

gnucash/import-export/ofx/gnc-ofx-import.c
Bob-IT: Possibly it relied on there being one matcher and it all got freed when the dialog closed or it was just missed and it should be there so may be worth trying to add it just for clarity.

Bob-IT:
Maybe you just hide it?

jeanlaroche:
> Maybe you just hide it?
I'll try that, but I suspect the hiding won't happen as it's called in the same thread as the closing. But I'll try it, thanks for the suggestion!

gnucash/import-export/ofx/gnc-ofx-import.c
Bob-IT: I do not know.

gnucash/import-export/ofx/gnc-ofx-import.c
Bob-IT: That's OK then, I did not see that but yes it is best to free them when no longer required.

Bob-IT:
The other thing is maybe using g_signal_connect_after

gnucash/import-export/ofx/gnc-ofx-import.c
jralls: It would be a miraculous compiler that could detect memory leaks or their cousins use-after-free and double-free. Even static analyzers generally can't detect memory errors. Instruments includes a leak checker tool, Apple's [malloc library](https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man3/malloc.3.html) has some nice built-in facilities that you can enable by setting environment variables, and if you find something really hairy there's [address sanitizer](https://developer.apple.com/documentation/code_diagnostics/address_sanitizer).

jeanlaroche: Oh, I didn't mean at compile time! I meant at run time. When you quit the program, it's pretty easy for the debugger to flag arrays that haven't been properly freed. MSVC++ did that (used it for many years in the past).

gnucash/import-export/ofx/gnc-ofx-import.c
```diff
@@ -1038,6 +1025,95 @@ double ofx_get_investment_amount(const struct OfxTransactionData* data)
...
+gnc_ofx_process_next_file (GtkDialog *dialog, gpointer user_data)
+{
+    ofx_info* info = (ofx_info*) user_data;
+    if (info->file_iter->next)
+    {
+        // Done with the previous OFX file, process the next one if any.
+        GSList* temp = info->file_iter;
+        info->file_iter = info->file_iter->next;
+        temp = g_slist_delete_link(temp, temp);
+        gnc_file_ofx_import_process_file (info);
```
gjanssens, line 1039: Likely you can shorten this whole list juggling to just this:

```
    ofx_info* info = (ofx_info*) user_data;
    info->file_iter = g_slist_delete_link(info->file_iter, info->file_iter);
    if (info->file_iter)
        gnc_file_ofx_import_process_file (info);
    else
        ...
```

g_slist_delete_link returns the new head of the GSList after the requested element has been deleted.

However I also think both options still have a string memory leak. The filenames stored in the GSList are not freed by g_slist_delete_link. It will only free the GSList data structure. So you probably want to add a g_free before calling g_slist_delete_link.

Lastly your function assumes it will never be called with an info->file_iter that's set to NULL. While the code you have now written probably adheres to that constraint it's fragile to just assume this to be the case forever. So I propose to add a guard upfront checking that both info and info->file_iter are not NULL before any dereferencing.

gnucash/import-export/ofx/gnc-ofx-import.c
```diff
@@ -1038,6 +1025,95 @@ double ofx_get_investment_amount(const struct OfxTransactionData* data)
...
+    }
+    else
+    {
+        // No more OFX file to process, final cleanup.
+        g_free (info);
+        if (ofx_created_commodites)
+        {
+            /* FIXME: Present some result window about the newly created
+             * commodities */
```
gjanssens, line 1048: Is this planned as part of the current PR ?

gnucash/import-export/ofx/gnc-ofx-import.c
```diff
@@ -1038,6 +1025,95 @@ double ofx_get_investment_amount(const struct OfxTransactionData* data)
...
+{
+    LibofxContextPtr libofx_context = libofx_get_new_context();
+    char* filename = info->file_iter->data;
+    GtkWindow *parent = info->parent;
+
+#ifdef G_OS_WIN32
+    char * selected_filename = g_win32_locale_filename_from_utf8 (filename);
+#else
+    char * selected_filename = g_strdup (filename);
+#endif
```
gjanssens, line 1077: Nitpick: the g_strdup on non-Windows platforms is not really needed. The selected_filename is used as is (no changes made to it). So you can avoid the memory copy penalty here if you also make the g_free at the end Windows only.

gnucash/import-export/ofx/gnc-ofx-import.c
gjanssens: This would also be easy if the ofx importer was transformed into an assistant.

gnucash/import-export/ofx/gnc-ofx-import.c
jeanlaroche: Yes, I don't know how consistently that's done in GC, sometime I see it done, sometimes not. But I can certainly add that! I'll check the freeing stuff, thanks for the review.

gnucash/import-export/ofx/gnc-ofx-import.c
jeanlaroche: It was part of the old code, so I didn't remove it....

gnucash/import-export/ofx/gnc-ofx-import.c
jeanlaroche: ah, OK I didn't realize that.

gnucash/import-export/ofx/gnc-ofx-import.c
gjanssens: No problem - it just caught my eye.

gnucash/import-export/ofx/gnc-ofx-import.c
gjanssens: It's not done very consistently everywhere indeed. I think in this case it would be relatively safe to assume you are never getting a NULL info parameter. I'm more worried about dereferencing info->file_iter in the future as it's less obvious.

gjanssens: Oh, one more nitpick; you can call it `file_list` rather than `file_iter`. You're not really using it to iterate in a loop.

jeanlaroche:
Thanks again for the review. I used your simplification of the list juggling, much cleaner. 
I fixed the freeing of the filename, as well as not duplicating the filename in non-windows platforms. Also added a guard against a NULL file-list.

About the assistant: I'm not personally a huge fan of that, I tend to think it makes things a bit more heavy, less streamlined (comparing for example the OFX import and the QIF import). But I also see your point. I have another PR to add the reconcile at the end of the ofx import(s) and that would fit nicely into the assistant idea.

jeanlaroche:
To avoid the conflict that needed to be resolved, I had to add handling the automatic reconcile after import. I kept all the suggestions from the above comments however.

gjanssens:
Youir last change needs more work. You add a checkbox in a dialog that is used by three different importers (CSV, OFX and AQB) but only do something with it in the OFX importer. Can you find a way to add this only for OFX ?
Also I think it should be presented to the user only once. Not again for each file.

jeanlaroche:
> 
> 
> Youir last change needs more work. You add a checkbox in a dialog that is used by three different importers (CSV, OFX and AQB) but only do something with it in the OFX importer. Can you find a way to add this only for OFX ?
> Also I think it should be presented to the user only once. Not again for each file.

Ah yes, I'll change it so it's only for OFX. You're assuming that people will want to uniformly reconcile/not-reconcile for all OFX? Is it OK if I make it remember its value, but leave it available? If the user wants to change that?

jeanlaroche:
> > Ah yes, I'll change it so it's only for OFX. You're assuming that people will want to uniformly reconcile/not-reconcile for all OFX? Is it OK if I make it remember its value, but leave it available? If the user wants to change that?
> 
> No, I mean that users would will want to reconcile or not all files loaded at once in one import run. It can be different on a future ofx import run. However the way I understand your code the option will be there for each file in a single run. But only the state of the option on the last file to import will take effect. Did I misinterpret this ?

It does not work quite that way: the reconcile happens after *each* import, not at the end of all imports. So the check box controls whether you'll reconcile for this ofx import. I changed it so it remember from ofx file to ofx file. I also changed it so the check box does not show by default (for other parts of the code that use the matcher).

gnucash/import-export/ofx/gnc-ofx-import.c
gjanssens: Do you plan to change this still ?

gnucash/gtkbuilder/dialog-import.glade
```diff
@@ -945,6 +945,21 @@
             <property name="position">1</property>
           </packing>
         </child>
+        <child>
+            <object class="GtkCheckButton" id="reconcile_after_close_button">
+                <property name="label" translatable="yes">Reconcile after match</property>
+                <property name="visible">True</property>
```
gjanssens, line 951: Have you tried setting the default visibility to False ? If I'm not mistaken gtk_widget_show_all won't automatically display such widgets. That may make it no longer necessary to call `gnc_gen_trans_list_show_reconcile_after_close` in `gnc_gen_trans_list_show_all`.

jeanlaroche: > Have you tried setting the default visibility to False

That was the first thing I tried, and it didn't work. But I'll try again. It's quite possible that I goofed something the first time around (perhaps didn't build ALL).

> Do you plan to change this still ?

I had fixed that but it looks like I lost it when I rebased against master. Let me put it back, sorry about that.

gnucash/import-export/ofx/gnc-ofx-import.c
Bob-IT: This has been done.

gnucash/gtkbuilder/dialog-import.glade
Bob-IT: For information, what you needed to do was set the visibility of the button to false and also enable the setting to ignore gtk_widget_show_all

PR #698, opened 2020-04-18
Bug797236 Reconciliation - Retain visibility of current split

goodvibes2:
Ensure current split (or last selected), is visible after all
operations.
After a deletion, select next split so can delete multiple trans using
just the Delete button. If no next split, select previous.

### DIFF
--- a/gnucash/gnome/reconcile-view.c
+++ b/gnucash/gnome/reconcile-view.c
@@ -901,12 +901,26 @@ void
 gnc_reconcile_view_refresh (GNCReconcileView *view)
 {
     GNCQueryView *qview;
+    GtkTreeSelection *selection;
+    GList *path_list, *node;
 
     g_return_if_fail (view != NULL);
     g_return_if_fail (GNC_IS_RECONCILE_VIEW (view));
 
     qview = GNC_QUERY_VIEW (view);
     gnc_query_view_refresh (qview);
+
+    /* Ensure last selected split, if any, can be seen */
+    selection = gtk_tree_view_get_selection (GTK_TREE_VIEW (qview));
+    path_list = gtk_tree_selection_get_selected_rows (selection, NULL);
+    node = g_list_last(path_list);
+    if (node)
+    {
+        GtkTreePath *tree_path = node->data;
+        gtk_tree_view_scroll_to_cell (GTK_TREE_VIEW (qview),
+                                      tree_path, NULL, FALSE, 0.0, 0.0);
+    }
+    g_list_free_full (path_list, (GDestroyNotify) gtk_tree_path_free);
 
     /* Now verify that everything in the reconcile hash is still in qview */
     if (view->reconciled)
--- a/gnucash/gnome/window-reconcile.c
+++ b/gnucash/gnome/window-reconcile.c
@@ -659,7 +659,7 @@ gnc_save_reconcile_interval(Account *account, time64 statement_date)
 
     /*
      * See if we need to remember days(weeks) or months.  The only trick
-     * value is 28 days which could be wither 4 weeks or 1 month.
+     * value is 28 days which could be either 4 weeks or 1 month.
      */
     if (days == 28)
     {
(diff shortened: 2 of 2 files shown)


gnucash/gnome/window-reconcile.c
```diff
@@ -1365,11 +1421,50 @@ gnc_ui_reconcile_window_delete_cb(GtkButton *button, gpointer data)
...
     }
 
+    /* Save the view and split to be selected after split is deleted */
+    save_view = GNC_RECONCILE_VIEW (recnData->debit);
+    selection = gtk_tree_view_get_selection (GTK_TREE_VIEW (save_view));
+    if (gtk_tree_selection_count_selected_rows(selection) == 0)
+    {
+        save_view = GNC_RECONCILE_VIEW (recnData->credit);
+        selection = gtk_tree_view_get_selection (GTK_TREE_VIEW (save_view));
+    }
```
Bob-IT, line 1431: This took me a while to understand the above, as the delete button is used for credit and debits you are deciding what view to use. I would suggest adding a comment here or add a function to return the view, you could probably base it on  gnc_reconcile_window_get_current_split and return a view instead of a split.

Bob-IT:
I was wondering for getting the next split you could use the selection that you have, there being only one selected for delete actions, get the path, use path_next, do your checks and step on if required. You may be able to move and do the selection before the delete of the transaction or if needed use a row reference to track the path and use that do the selection.

gnucash/gnome/reconcile-view.c
```diff
@@ -901,13 +901,27 @@ void
...
     g_return_if_fail (view != NULL);
     g_return_if_fail (GNC_IS_RECONCILE_VIEW (view));
 
     qview = GNC_QUERY_VIEW (view);
     gnc_query_view_refresh (qview);
 
+    /* Ensure last selected split, if any, can be seen */
+    selection = gtk_tree_view_get_selection (GTK_TREE_VIEW (qview));
+    path_list = gtk_tree_selection_get_selected_rows (selection, NULL);
+    node = g_list_last(path_list);
```
Bob-IT, line 916: can you add a space after g_list_last

jralls:
One more thing: Master, not maint. We don't want to introduce any behavior changes in the final 3.x release.

gnucash/gnome/window-reconcile.c
goodvibes2: Fixed in PR 707 (against master). There is new function (gnc_reconcile_window_get_selection_view) for getting the view containing the selection with Doxygen comment.

gnucash/gnome/reconcile-view.c
goodvibes2: Fixed in PR 707 (against master). I've implemented your method in gnc_reconcile_window_delete_set_next_selection() which pre-selects the next split before the deletion.

gnucash/gnome/window-reconcile.c
```diff
@@ -1239,6 +1239,59 @@ gnc_reconcile_window_create_view_box(Account *account,
...
+            }
+        }
+        if (valid)
+        {
+            if (trans != xaccSplitGetParent (pointer))
+                next_split = pointer;
+            valid = gtk_tree_model_iter_next (model, &iter);
+        }
+    }
+    return next_split;
```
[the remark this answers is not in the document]
goodvibes2: Fixed in PR 707 (against master). The deletion changes in this PR are more to do with maintaining visibility of where you're working than enabling faster multiple transaction deletes.
Without my changes, everytime the list refreshes, it would scroll to the top, even though a selection would still exist if the selected split had not been deleted.

Re the nested loops: I have fixed and must remember not to copy old code.

gnucash/gnome/window-reconcile.c
```diff
@@ -1365,11 +1421,50 @@ gnc_ui_reconcile_window_delete_cb(GtkButton *button, gpointer data)
...
     xaccTransDestroy(trans);
 
+    /* set the selected split to be the one after the deleted one,
+     *  or if there were no more, the last one.
+     *  gnc_reconcile_view_refresh() triggered by
+     *  the deletion will ensure the selected split can be seen.
+     */
+    if (next_split)
+    {
+        gpointer pointer;
```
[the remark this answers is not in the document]
goodvibes2: Thanks for the explanation. Fixed in PR 707.

goodvibes2:
Close this PR has been replaced by PR 707 (against master).

PR #700, opened 2020-04-19
Add reconcile immediately after ofx import, based on statement

jeanlaroche:
To make this work, I had to add the right header and make a change to the CMakeLists file.

The ofx code used to have a static int to count imported transactions, which was a bit hacky.
I replaced it with a structure that's passed to all the callbacks. The structure has
the transaction count, as well as 2 lists if accounts + statement balances.
These lists are used after the import of the ofx file to do the reconciliation.
NOTE: I wouldn't need lists since the current code only process one ofx at a time. However
I have another PR for importing several ofx files in one shot, so this PR is ready for that.
The reconciliation code is modeled after what's done in aqbanking.
- Question: should I add a preference to enable/disable automatic reconciliation after ofx import?
- Also: I just noticed the "Reverse balanced accounts" preference (I hadn't noticed it). Would this option affect the sign of the balance we should report in the reconciliation, depending on the type of account?

### DIFF
--- a/gnucash/import-export/ofx/gnc-ofx-import.c
+++ b/gnucash/import-export/ofx/gnc-ofx-import.c
@@ -50,6 +50,7 @@
 #include "gnc-ui.h"
 #include "dialog-account.h"
 #include "dialog-utils.h"
+#include "window-reconcile.h"
 
 #define GNC_PREFS_GROUP "dialogs.import.ofx"
 #define GNC_PREF_AUTO_COMMODITY "auto-create-commodity"
@@ -66,7 +67,13 @@ static QofLogModule log_module = GNC_MOD_IMPORT;
 GNCImportMainMatcher *gnc_ofx_importer_gui = NULL;
 static gboolean auto_create_commodity = FALSE;
 static Account *ofx_parent_account = NULL;
-static gint num_trans_processed = 0;
+// Structure we use to gather information about statement balance/account etc.
+typedef struct _ofx_info
+{
+    gint num_trans_processed;
+    GSList* statement;
+    GtkWindow* parent;
+} ofx_info ;
 
 GList *ofx_created_commodites = NULL;
 
@@ -111,6 +118,7 @@ set_associated_income_account(Account* investment_account,
     xaccAccountCommitEdit(investment_account);
 }
 
+int ofx_proc_statement_cb(struct OfxStatementData data, void * statement_user_data);
 int ofx_proc_security_cb(const struct OfxSecurityData data, void * security_user_data);
 int ofx_proc_transaction_cb (struct OfxTransactionData data, void *user_data);
 int ofx_proc_account_cb(struct OfxAccountData data, void * account_user_data);
@@ -409,7 +417,8 @@ int ofx_proc_transaction_cb(struct OfxTransactionData data, void *user_data)
     Transaction *transaction;
     Split *split;
     gchar *notes, *tmp;
-    GtkWindow *parent = GTK_WINDOW (user_data);
+    ofx_info* info = (ofx_info*) user_data;
+    GtkWindow *parent = GTK_WINDOW (info->parent);
 
     g_assert(gnc_ofx_importer_gui);
 
@@ -890,16 +899,20 @@ int ofx_proc_transaction_cb(struct OfxTransactionData data, void *user_data)
         xaccTransDestroy(transaction);
         xaccTransCommitEdit(transaction);
     }
-    num_trans_processed += 1;
+    info->num_trans_processed += 1;
     return 0;
 }//end ofx_proc_transaction()
 
-/*
-int ofx_proc_statement_cb(struct OfxStatementData data, void * statement_user_data)
+
+int ofx_proc_statement_cb (struct OfxStatementData data, void * statement_user_data)
 {
-  return 0;
+    ofx_info* info = (ofx_info*) statement_user_data;
+    struct OfxStatementData* statement = g_new (struct OfxStatementData, 1);
+    *statement = data;
+    info->statement = g_slist_append (info->statement, statement);
+    return 0;
 }//end ofx_proc_statement()
-*/
+
 
 int ofx_proc_account_cb(struct OfxAccountData data, void * account_user_data)
 {
@@ -910,6 +923,8 @@ int ofx_proc_account_cb(struct OfxAccountData data, void * account_user_data)
     /* In order to trigger a book options display on the creation of a new book,
      * we need to detect when we are dealing with a new book. */
     gboolean new_book = gnc_is_new_book();
+    ofx_info* info = (ofx_info*) account_user_data;
+    Account* account = NULL;
 
     const gchar * account_type_name = _("Unknown OFX account");
 
@@ -985,10 +1000,10 @@ int ofx_proc_account_cb(struct OfxAccountData data, void * account_user_data)
                                   "%s \"%s\"",
                                   account_type_name,
                                   data.account_name);
-        gnc_import_select_account(gnc_gen_trans_list_widget(gnc_ofx_importer_gui),
-                                  data.account_id, 1,
-                                  account_description, default_commodity,
-                                  default_type, NULL, NULL);
+        account = gnc_import_select_account( gnc_gen_trans_list_widget(gnc_ofx_importer_gui),
+                                             data.account_id, 1,
+                                             account_description, default_commodity,
+                                             default_type, NULL, NULL);
         g_free(account_description);
     }
     else
@@ -1034,6 +1049,9 @@ void gnc_file_ofx_import (GtkWindow *parent)
     LibofxContextPtr libofx_context = libofx_get_new_context();
     GList *filters = NULL;
     GtkFileFilter* filter = gtk_file_filter_new ();
+    GSList *iter = NULL;
+    // Create the structure we're using to gather reconciliation information.
+    ofx_info info = {0, NULL, parent};
 
     ofx_PARSER_msg = false;
     ofx_DEBUG_msg = false;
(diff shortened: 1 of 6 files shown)


christopherlam:
If you download and import an old OFX, the `ledger_balance_date` and `ledger_balance` will be old data, and this wouldn't reconcile...

christopherlam:
Having said that, it's not a big deal if the old OFX doesn't reconcile; the reconcile window status bar will explain there are reconciled splits after `statement_date`. I think to harmonise OFX to mimic AQB is a fine idea.

christopherlam:
>     * Also: I just noticed the "Reverse balanced accounts" preference (I hadn't noticed it). Would this option affect the sign of the balance we should report in the reconciliation, depending on the type of account?

From experience this preference does not affect the ability to reconcile at all.

jeanlaroche:
> 
> 
> Having said that, it's not a big deal if the old OFX doesn't reconcile; the reconcile window status bar will explain there are reconciled splits after `statement_date`. I think to harmonise OFX to mimic AQB is a fine idea.

Also, if you've already imported the old ofx, then there won't be any reconciliation window because all transactions will already have been reviewed and imported. And as you said, the reconciliation date is the date of the statement sent by your bank, so indeed you'll be warned if some of your transactions fall past that date...

jeanlaroche:
> 
> 
> > Also: I just noticed the "Reverse balanced accounts" preference (I hadn't noticed it). Would this option affect the sign of the balance we should report in the reconciliation, depending on the type of account?
> 
> From experience this preference does not affect the ability to reconcile at all.

I was wondering if I needed to present the negative of the statement balance based on that preference, but I did a test and apparently GC takes care of it (i.e. the right balance is presented to the user in all cases).

christopherlam:
> > Having said that, it's not a big deal if the old OFX doesn't reconcile; the reconcile window status bar will explain there are reconciled splits after `statement_date`. I think to harmonise OFX to mimic AQB is a fine idea.
> 
> Also, if you've already imported the old ofx, then there won't be any reconciliation window because all transactions will already have been reviewed and imported. And as you said, the reconciliation date is the date of the statement sent by your bank, so indeed you'll be warned if some of your transactions fall past that date...

Not necessarily. It is (to me) common to enter most transactions manually, and also import an OFX and match all lines to their OFX. This ensures my manual splits get their `online_id` field populated.

gnucash/import-export/ofx/gnc-ofx-import.c
```diff
@@ -1131,14 +1131,27 @@ void gnc_file_ofx_import (GtkWindow *parent)
...
+                    switch(xaccAccountGetType (account))
+                    {
+                        case ACCT_TYPE_CREDIT:
+                        case ACCT_TYPE_LIABILITY:
+                            sign = -1;
+                        default:
+                            sign = 1;
+                    }
+                    // Grab the balance value and date from the statement and open a reconcile window for this account.
+                    gnc_numeric value = double_to_gnc_numeric (sign*statement->ledger_balance, xaccAccountGetCommoditySCU (account), GNC_HOW_RND_ROUND_HALF_UP);
```
Bob-IT, line 1147: Why do you not use gnc_numeric_neg ?

gnucash/import-export/ofx/gnc-ofx-import.c
```diff
@@ -1131,14 +1131,27 @@ void gnc_file_ofx_import (GtkWindow *parent)
...
+                {
+                    // The balance needs to be corrected for credit-card and loan accounts.
+                    float sign = 1;
+                    switch(xaccAccountGetType (account))
+                    {
+                        case ACCT_TYPE_CREDIT:
+                        case ACCT_TYPE_LIABILITY:
+                            sign = -1;
+                        default:
+                            sign = 1;
```
christopherlam, line 1144: This is not comprehensive enough. There is a distinct lack of `xaccAccountTypeGetFundamental` function which converts gnucash's account-types to one of the 5 fundamental types -- acount/liability/income/expense/equity. This also affects, for example, budget totals which sum only `ACCT_TYPE_ASSET` and ignores `ACCT_TYPE_BANK` budget amounts. See the following snippet onwards: https://github.com/Gnucash/gnucash/blob/802306503057a8f4476acc830eaa0e1f777e4c86/gnucash/gnome/gnc-budget-view.c#L1197

I suspect adding the following to `Account.cpp` is easy enough?
```cpp
GNCAccountType
xaccAccountTypeGetFundamental (GNCAccountType t)
{
    switch (t)
    {
    case ACCT_TYPE_BANK:
    case ACCT_TYPE_STOCK:
    case ACCT_TYPE_MUTUAL:
    case ACCT_TYPE_CURRENCY:
    case ACCT_TYPE_CASH:
    case ACCT_TYPE_ASSET:
    case ACCT_TYPE_RECEIVABLE:
        return ACCT_TYPE_ASSET;
    case ACCT_TYPE_CREDIT:
    case ACCT_TYPE_LIABILITY:
    case ACCT_TYPE_PAYABLE:
        return ACCT_TYPE_LIABILITY;
    case ACCT_TYPE_INCOME:
        return ACCT_TYPE_INCOME;
    case ACCT_TYPE_EXPENSE:
        return ACCT_TYPE_EXPENSE;
    case ACCT_TYPE_EQUITY:
        return ACCT_TYPE_EQUITY;
    case ACCT_TYPE_TRADING:
    default:
        return ACCT_TYPE_NONE;
    }
}
```

jeanlaroche:
Yes, I was looking for something like xaccAccountTypeGetFundamental(). I will add that, and use that during the reconcile.

christopherlam:
> Yes, I was looking for something like xaccAccountTypeGetFundamental(). I will add that, and use that during the reconcile.

... and you'll have your big dilemma whether `ACCT_TYPE_TRADING`'s mapping should be `ACCT_TYPE_NONE` or `ACCT_TYPE_EQUITY`...

jeanlaroche:
After more testing, I realize that there is no need to reverse the sign of the balance based on the type of account we're dealing with. I'm keeping the function that was added to Account.cpp.
I rebased against upstream/master, fixed the CMakeList conflict. Also switched to using g_slist_free_full() to free the Glist of statements.

Bob-IT:
@christopherlam Did you try this?, I do not do OFX imports but I think you do, I just built 3.901 to check Jean's other PR and tried a test import that results in 3 transactions indicating 'New, Unbalanced' but with a reconcile window open. Does not seem logical to me?

christopherlam:
> @christopherlam Did you try this?, I do not do OFX imports but I think you do, I just built 3.901 to check Jean's other PR and tried a test import that results in 3 transactions indicating 'New, Unbalanced' but with a reconcile window open. Does not seem logical to me?

Agree I think the `recnWindowWithBalance` call is premature; it should probably wait until import matcher is complete before offering to reconcile.

Bob-IT:
@jeanlaroche Could you look at changing this ?

jralls:
It is what AQBanking does. Neither the reconcile window nor the matcher dialog are modal so there's nothing forcing the user to do the reconcile first. OTOH in both cases it would be more intuitive for the reconcile check to happen after the matcher runs. That might involve rearranging the flow a bit if the matcher is called outside of the actual importer. That rearranging would have another benefit: If the import provides transfer accounts then the importer could pop a dialog asking if the user wants to run the matcher or just go with what's in the import.

jeanlaroche:
> That rearranging would have another benefit: If the import provides transfer accounts then the importer could pop a dialog asking if the user wants to run the matcher or just go with what's in the import.

I'm not understanding what you're saying, what are the transfer accounts that the import might provide? Can you elaborate a bit?

I'll start looking at how to defer the reconcile until after the match is done. But I'm reiterating my question above:
> I suspect a difficulty: The match dialog call returns immediately (it's non-modal) so how do you suggest I pop the reconcile dialog after the matching is done?
The only solution I can see is to add the reconcile info (account, balance, date) to the matching dialog call (as a parameter, probably a structure, or a list of structures since we can have several accounts to reconcile) indicating "open the reconcile when you're done". But that's a rather pervasive change.
Can you think of anything better?

jralls:
"Transfer account" is the register basic-view name for the "other" account. OFX imports have only two splits, one for the OFX file's account and one for the "other" or "transfer" account. CSV imports can have several splits but when they do they need to include the full account name of each split's account. Even when importing a one-line CSV transaction record it can still provide the "other" account. When it does the match dialog might be thought superfluous.

For knowing when the Matcher is done connect to its [`response` signal](https://developer.gnome.org/gtk3/stable/GtkDialog.html#GtkDialog-response). That has the added benefit of letting you decide whether to launch the reconcile based on the supplied `response_id`. You wouldn't want to if it's `GTK_CANCEL`.

jeanlaroche:
> For knowing when the Matcher is done connect to its [`response` signal](https://developer.gnome.org/gtk3/stable/GtkDialog.html#GtkDialog-response). That has the added benefit of letting you decide whether to launch the reconcile based on the supplied `response_id`. You wouldn't want to if it's `GTK_CANCEL`.

Ah ok great, I would add the reconcile data as user_data I imagine, so it's available to my callback when it finally gets called, is that right?

jralls:
At least that. If you need more stuff you can make a struct to pass a ptr to. It's async, so be careful about making sure that whatever you pass in is still around when the signal fires and that you don't leak.

christopherlam:
An idea: how would it be to show the reconcile window if, after importing (and matching), the reconcile window appears to mark splits as reconciled if the ofx balance matches the book balance at statement date?

jralls:
Do you mean that it wouldn't offer to reconcile if the balance didn't match? That would ignore the ordinary case where there was a check that hadn't yet cleared for which reconciling is appropriate and easily done. Even if one used the cleared balance and there was a mismatch, I'd think that the user would like to know about it and the reconciliation window is an excellent tool for helping to find the discrepancy.

jeanlaroche:
How do you guys feel about having one more check box on the transaction 
match dialog, that say "open reconcile dialog when done" or something 
like that? Unchecked by default, but if you check it while matching 
transaction, the reconcile dialog opens when you're done?

My guess is many people will *not* like the reconcile window opening by 
default after every ofx import. So we have to offer the choice to 
open/not open the reconcile dialog. This is less invasive than a new option.
J.

jralls:
I guess that would be OK if it was only displayed if there was balance info in the OFX or FinTS download.

I agree that popping the Reconcile window without asking isn't going to make people happy, it should at least put up a dialog box first like the AQB importer does already.

jeanlaroche:
It's not going to make people happy if they have to respond "no" to a 
dialog every time they import an ofx so the reconcile does not open.
Ideally, it would be a response that sticks, would that be OK?

So:
- you import your ofx
- the importer tells you, there's a balance info do you want to 
reconcile after matching. You can say yes, no, and it remembers for this 
session (and the next?)

What if the user changes their mind, and want to open the reconcile the 
next time around? Is there a way to reset theses sticky responses (like 
when you say "don't warn me when I delete transactions")?
Jean

jralls:
Actions>Reset Warnings...

But you have to use the right functions to set the "don't ask again" preference.

jeanlaroche:
OK Great. I'll go in that direction then. I'll find some examples in the 
code.
Jean

christopherlam:
> Do you mean that it wouldn't offer to reconcile if the balance didn't match? That would ignore the ordinary case where there was a check that hadn't yet cleared for which reconciling is appropriate and easily done. Even if one used the cleared balance and there was a mismatch, I'd think that the user would like to know about it and the reconciliation window is an excellent tool for helping to find the discrepancy.

My use case was that I am not always very consistent in ensuring all my OFX imports will cover missing dates. i.e. I may have reconciled transactions until Dec 2019, and my ofx import is from Feb to Apr 2020, and obviously the reconcile will fail.

My thinking was that if OFX contains valuable balance information, it could populate a `statement_data` account metadata as we discussed in mailing list, and offer future reconciliation. But this is currently unworkable.

As far as transitioning from import matcher to reconciliation, I think two options:
1. when import is complete, gnc-dialog offer "The OFX confirms balance is $X on Y/Y/YYYY. Do you wish to reconcile?" [Reconcile] [Postpone] whereby `postpone` will populate the `statement_data`.
2. the import matcher may be augmented to offer reconciliation as you have already discussed.

PR #701, opened 2020-04-20
Python app utils 2

c-holtermann:
An import statement was missing to actually utilize the app_utils wrapper. Modify python shell to use this wrapper.

### DIFF
--- a/bindings/python/__init__.py
+++ b/bindings/python/__init__.py
@@ -4,6 +4,7 @@
 # instead of
 # >>> from gnucash.gnucash_core import thingy
 from gnucash.gnucash_core import *
+from . import app_utils
 ##  @file
 #   @brief helper file for the importing of gnucash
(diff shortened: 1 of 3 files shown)


c-holtermann:
@gjanssens  - in a way this is a bugfix / completion for the commit I did for app-utils e2c37359ef07132e5bb8568fa9c06ac68202e5f8 which is not really accessible right now. To me it seems that it makes sense in maint. But you decide.

gjanssens:
> @gjanssens - in a way this is a bugfix / completion for the commit I did for app-utils [e2c3735](https://github.com/Gnucash/gnucash/commit/e2c37359ef07132e5bb8568fa9c06ac68202e5f8) which is not really accessible right now. To me it seems that it makes sense in maint. But you decide.

Ok, I didn't realize that. maint is fine then.

PR #703, opened 2020-04-20
Python gettext localize master

c-holtermann:
Patches of PR #655 applied on master.


c-holtermann:
This is still the same state as #655 - needs work.

c-holtermann:
@jralls would you please try this version?

bindings/python/tests/CMakeLists.txt
```diff
@@ -16,6 +16,7 @@ endif()
 
 set(test_python_bindings_DATA
         runTests.py.in
+	test_gettext.py
```
gjanssens, line 19: Cosmetic: can you use spaces instead of a tab here ? That will align the file with the others.

jralls:
It works. Please clean out the tabs (some programming editors make that easy, e.g. emacs has `M-X untabbify`, vim has `:retab`) and force push and I'll merge it. (Note that feature freeze for 4.0 is this weekend so it would be *really good* to get it in by then.)

c-holtermann:
> 
> 
> Looks good. Do you intend to work on the unit test still ?
> Also, you may have to rerun make or ninja and commit an updated POTFILES.in afterwards. If I run ninja right after merging your branch a few lines change in that file.

I don't have gettext 0.2. I'd be glad if someone could supply me with POTFILES.in after this PR has been merged.

bindings/python/tests/CMakeLists.txt
c-holtermann: Sure. I'm using vim and on python files that gets done automatically. But this is a txt.

jralls:
Thanks!

fellen:
> I don't have gettext 0.2. I'd be glad if someone could supply me with POTFILES.in after this PR has been merged.

Sent by private email.

PR #705, opened 2020-04-25
Bug 797351 - General ledger register transaction becomes "zombie" aft…

jeanlaroche:
…er deletion

This issue arises every time a pending transaction is deleted (it does not arise if blank splits are deleted). This PR fixes the issue by calling gnc_split_register_redraw() once the pending transaction has be deleted.

### DIFF
--- a/gnucash/register/ledger-core/split-register.c
+++ b/gnucash/register/ledger-core/split-register.c
@@ -1223,6 +1223,7 @@ gnc_split_register_delete_current_trans (SplitRegister* reg)
         xaccTransCommitEdit (trans);
     }
     gnc_resume_gui_refresh();
+    gnc_split_register_redraw (reg);
     LEAVE (" ");
 }
 
--- a/gnucash/register/ledger-core/gnc-ledger-display2.c
+++ b/gnucash/register/ledger-core/gnc-ledger-display2.c
@@ -809,7 +809,6 @@ gnc_ledger_display2_internal (Account *lead_account, Query *q,
 
     ld->use_double_line_default = use_double_line;
 
-    // JEAN: add mismatched_commodities
     ld->model = gnc_tree_model_split_reg_new (reg_type, style, use_double_line, is_template, mismatched_commodities);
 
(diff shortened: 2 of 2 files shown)


gnucash/register/ledger-core/split-register.c
```diff
@@ -1223,6 +1223,7 @@ gnc_split_register_delete_current_trans (SplitRegister* reg)
         xaccTransCommitEdit (trans);
     }
     gnc_resume_gui_refresh();
+    gnc_split_register_redraw (reg);
```
jralls, line 1226: @Bob-IT Does it make sense to force a redraw or would it be better to move the `gnc_resume_gui_refresh() to before `if (was_open)`?

jeanlaroche: just FYI I tried this
```
    was_open = xaccTransIsOpen (trans);
    xaccTransDestroy (trans);
    gnc_resume_gui_refresh();
    if (was_open)
    {
        DEBUG ("committing");
        xaccTransCommitEdit (trans);
    }
```
But this does not fix the issue, the split is still un-editable.

jralls: OK, and looking at the rest of the file,
```
gnc_resume_gui_refresh();
gnc_split_register_redraw (reg);
```
is used in 3 other places. We'll go with it.

jralls:
Thanks!

jralls:
Thanks!

gnucash/register/ledger-core/split-register.c
jeanlaroche: That was also my reasoning, but I have to admit, I don't really know why/how it fixes the issue. I do know that the GUID for the blank split is invalid after the delete, and that's what cause the split to become un-editable, but I don't really understand which part of gnc_split_register_redraw fixes that (but it does)...

jralls: The register holds a dangling guid reference to the old blank split. Redrawing the register causes it to destroy those cells and make a new empty transaction with a new blank split

PR #706, opened 2020-04-25
[master] guid dance for new-aging/new-owner and balsheet-pnl

christopherlam:
* old reports' guids have "-old" appended, and hidden into `--extras`
* new reports guids supplant old ones, and also replace menu items
* incompatible options are ignored
* tests and hooks amended
* experimental messages removed
* [ ] any saved-report with experimental report will now fail; if we were to create a guid redirect, it'd mean supporting it forever?


christopherlam:
is this on the cards for 4.0 feature freeze?

jralls:
I don't have any problem with it. I doubt that many users will have saved configurations on the experimental reports, plus being *experimental* implies a certain amount of instability. OTOH a scrub of saved-reports-2.8 -> saved-reports-4.0 could replace the guids and could be removed in 4.900.

@gjanssens, any thoughts?

christopherlam:
If this is merged soon, the 4.0 release notes _must_ include notification that these reports are now using different options and renderer, the interpretation and layout will be different... Old reports still accessible via `--extras` but will be removed, and any missing feature to be reported in bugzilla to attempt porting to new reports. From my view, I think the only unported feature is the 'flatten list to depth limit' from accounts tab.

PR #707, opened 2020-04-26
Bug 797236 - Reconciliation - Retain visibility of selected split

goodvibes2:
Ensure selected split (or last selected if multiples selected), is
visible after all operations when the list is refreshed.
After a deletion using the Reconcile window Delete button, the
selected split has been deleted, so preselect the next split after
the split to be deleted, so the working position in the list is
still visible. If no next split in the list, select previous.
The split preselected must have a different parent transaction to
the split to be deleted, as all splits in the transaction will be
deleted.

This PR (master) replaces maint PR 698.

### DIFF
--- a/gnucash/gnome/window-reconcile.c
+++ b/gnucash/gnome/window-reconcile.c
@@ -1320,12 +1320,93 @@ gnc_ui_reconcile_window_unrec_cb(GtkButton *button, gpointer data)
 }
 
 
+/** Get the debit or credit view that has at least 1 split selected.
+ *   gnc_reconcile_window_focus_cb() ensures only 1 view
+ *   has a selection.
+ * @param window The reconcile window.
+ */
+static GNCReconcileView *
+gnc_reconcile_window_get_selection_view (RecnWindow *recnData)
+{
+    if (gnc_reconcile_view_num_selected (GNC_RECONCILE_VIEW (recnData->debit)) > 0)
+        return GNC_RECONCILE_VIEW (recnData->debit);
+
+    if (gnc_reconcile_view_num_selected (GNC_RECONCILE_VIEW (recnData->credit)) > 0)
+        return GNC_RECONCILE_VIEW (recnData->credit);
+
+    return NULL;
+}
+
+
+/** Select the next split in the debit or credit view so that after the Delete
+ *   button is actioned, the working position in the list is still in view.
+ *  Unless this is done, the list will be scrolled to the top.
+ *  The new split selected must have a different parent transaction as all splits
+ *   for the transaction will be deleted.
+ */
+static void
+gnc_reconcile_window_delete_set_next_selection (RecnWindow *recnData, Split *split)
+{
+    GNCReconcileView *view = gnc_reconcile_window_get_selection_view (recnData);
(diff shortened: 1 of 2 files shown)


gnucash/gnome/window-reconcile.c
```diff
@@ -1320,12 +1320,95 @@ gnc_ui_reconcile_window_unrec_cb(GtkButton *button, gpointer data)
...
 static void
 gnc_ui_reconcile_window_delete_cb(GtkButton *button, gpointer data)
 {
     RecnWindow *recnData = data;
     Transaction *trans;
-    Split *split;
+    Split *split, *next_split;
+    GtkTreePath *save_path = NULL;
+    GNCReconcileView *view;    /* debit or credit view containing selection */
+    GtkTreeSelection *selection;
```
Bob-IT, line 1411: I think these may of been left behind in you chnages

gnucash/gnome/window-reconcile.c
```diff
@@ -1320,12 +1320,95 @@ gnc_ui_reconcile_window_unrec_cb(GtkButton *button, gpointer data)
...
+    if (!view)
+        return; // no selected split
+
+    path_list = gtk_tree_selection_get_selected_rows (selection, &model);
+    // get path of the first split selected - there should be only 1 selected
+    node = g_list_first (path_list);
+    if (!node)
+        return;
+    path = node->data;
+    save_del_path = gtk_tree_path_copy(path);
```
Bob-IT, line 1368: I think you may of missed the tree free for this and can you add a space after _copy.

goodvibes2: Hi Bob, Thanks for reviewing. Can I just add a commit to fix these problems or should I back out both commits, then redo with the fix to the comment typos first, in case John would like other changes?

Bob-IT: What I would do is add a commit to make the changes and then do a rebase and squash the new commit to be part of the first commit and then do a force push

gjanssens: For reviewing purposes it's often easier to just add a commit. When all review comments are handled satisfactorily, a final cleanup round before merging the PR is often useful.

This final cleanup round effectively consists of squashing fixup commits together with the commits they fix, possibly improve commit messages if needed, or whatever. This is done via an interactive rebase (check the man page for details on `rebase -i`) which is what @Bob-IT is referring to.

gnucash/gnome/window-reconcile.c
goodvibes2: Fixed in commit added today.

gnucash/gnome/window-reconcile.c
goodvibes2: Thanks people. Fixed in commit added today.

Bob-IT:
@goodvibes2 , Chris this looks OK to me, now what I suggest before I can merge to master is you create a commit for the reconcile-view changes for the bug, a commit for the delete split with suitable text and a commit for the typo's and then rebase on current master. How you do that is up to you and what feel comfortable in doing and as always there are many ways.

goodvibes2:
@Bob-IT, commits has been restructured as requested.

Bob-IT:
Thank you.

PR #708, opened 2020-04-26
Add check to alert user if target account for moved currencies does n…

jeanlaroche:
…ot have the right currency

This fixes the issue described in the bug. An alert panel is displayed warning the user, but allowing to continue (or pick another account).
One question though: I didn't add anything to check subaccount transactions. Should we do something about that? But what should we do? Check that the subaccounts have the same currency as the account to be deleted? Check that all the transactions are in the same currency, which matches account where the transactions are to be moved?
It seems a bit like a can of worms to me, so I'd be happy to leave it at what's proposed in this PR.

### DIFF
--- a/gnucash/gnome/gnc-plugin-page-account-tree.c
+++ b/gnucash/gnome/gnc-plugin-page-account-tree.c
@@ -74,6 +74,25 @@
 /* This static indicates the debugging module that this .o belongs to.  */
 static QofLogModule log_module = GNC_MOD_GUI;
 
+
+/********************************************************************
+ * delete_account_helper
+ * See if this account has any splits present.  Set the user data
+ * and return the same value to stop walking the account tree if
+ * appropriate.
+ ********************************************************************/
+typedef struct _delete_helper
+{
+    gboolean has_splits;
+    gboolean has_ro_splits;
+} delete_helper_t;
+
+
+static void delete_account_next (GtkAction *action, GncPluginPageAccountTree *page, Account* ta,
+                                 Account* sta, Account* saa, delete_helper_t delete_res);
+static void  delete_account_final (Account* account, Account* saa, Account* sta, Account* ta);
+
+
 #define PLUGIN_PAGE_ACCT_TREE_CM_CLASS "plugin-page-acct-tree"
 #define STATE_SECTION "Account Hierarchy"
 
@@ -1207,18 +1226,6 @@ gnc_plugin_page_account_tree_cmd_cascade_color_account (GtkAction *action, GncPl
     LEAVE(" ");
 }
 
-/********************************************************************
- * delete_account_helper
- * See if this account has any splits present.  Set the user data
- * and return the same value to stop walking the account tree if
- * appropriate.
- ********************************************************************/
-typedef struct _delete_helper
-{
-    gboolean has_splits;
-    gboolean has_ro_splits;
-} delete_helper_t;
-
 static gpointer
 delete_account_helper (Account * account, gpointer data)
 {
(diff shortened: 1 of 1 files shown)


gnucash/gnome/gnc-plugin-page-account-tree.c
```diff
@@ -1499,21 +1500,51 @@ gnc_plugin_page_account_tree_cmd_delete_account (GtkAction *action, GncPluginPag
...
+        while (1)
         {
-            /* Account deletion is cancelled, so clean up and return. */
-            gtk_widget_destroy(dialog);
-            g_list_free(filter);
-            g_free(acct_name);
-            return;
+            gboolean ta_match = TRUE;
+            response = gtk_dialog_run(GTK_DIALOG(dialog));
+            if (GTK_RESPONSE_ACCEPT != response)
```
jralls, line 1510: Don't use constant-first comparisons. The claim that it prevents assignment mistakes is BS, compilers know about that and error out.

jralls, line 1510: Don't use constant-first comparisons. The claim that it prevents assignment mistakes is BS, compilers know about that and error out.

gnucash/gnome/gnc-plugin-page-account-tree.c
```diff
@@ -1499,21 +1500,51 @@ gnc_plugin_page_account_tree_cmd_delete_account (GtkAction *action, GncPluginPag
          * Note that one effect of the modal dialog is preventing
          * the account selectors from being repopulated.
          */
-        response = gtk_dialog_run(GTK_DIALOG(dialog));
-        if (GTK_RESPONSE_ACCEPT != response)
+        account_currency = gnc_account_get_currency_or_parent (account);
+        // Ideally we should check all subaccount currencies, and verify they're all identical. If not, it's not advisable
+        // to move subaccount currencies to a new account.
+        while (1)
```
jralls, line 1506: use `TRUE` instead of `1`,  but better to use `!ta_match`. 

Plus this function is an appalling 350 lines with multiple unconditional blocks, so extracting this to a function would be even better.

gnucash/gnome/gnc-plugin-page-account-tree.c
```diff
@@ -1499,21 +1500,51 @@ gnc_plugin_page_account_tree_cmd_delete_account (GtkAction *action, GncPluginPag
...
+            }
+            if (trans_mas && gtk_widget_is_sensitive(trans_mas))
+            {
+                ta = gnc_account_sel_get_account(GNC_ACCOUNT_SEL(trans_mas));
+                ta_match = gnc_account_get_currency_or_parent (ta) == account_currency;
+            }
+            if (sa_mas && gtk_widget_is_sensitive(sa_mas))
+                saa = gnc_account_sel_get_account(GNC_ACCOUNT_SEL(sa_mas));
+            if (sa_trans_mas && gtk_widget_is_sensitive(sa_trans_mas))
+                sta = gnc_account_sel_get_account(GNC_ACCOUNT_SEL(sa_trans_mas));
```
jralls, line 1526: There's absolutely no point fo doing this inside the loop. Even given that the loop will seldom run more than once it only makes the code harder to read to include them here.

jeanlaroche:
I completely agree!!! I never use this construct which I've always 
thought was complete BS as well.
I'm only doing it here because it's everywhere in the code already. But 
I"ll happily change it back.
These days, compilers are so freaking good, almost every try to steer 
them in the right direction is wasted effort.

gnucash/gnome/gnc-plugin-page-account-tree.c
jeanlaroche: If I put this in a separate function, I'll have to pass 3 Account** parameters to the function for ta, saa and sta, or a structure, or a list. Worth it? I have no problem passing Account** params myself, but I've heard you object to it.

gnucash/gnome/gnc-plugin-page-account-tree.c
jeanlaroche: mmm, if that's not in the while(), the user has no chance to change the account or cancel.

gnucash/gnome/gnc-plugin-page-account-tree.c
jeanlaroche: Yes, I don't use that construct myself, only replicating the coding style of the function. Modern compilers issue a warning if you do ```if (response = GTK_RESPONSE_ACCEPT)```

jralls:
> One question though: I didn't add anything to check subaccount transactions. Should we do something about that? But what should we do? Check that the subaccounts have the same currency as the account to be deleted? Check that all the transactions are in the same currency, which matches account where the transactions are to be moved?

I'd think that if an account has sub-accounts it's too big an assumption that the user wants to delete them and move all of their transactions to the same account as the parent's. The simplest way to deal with that would be to refuse to delete an account with subaccounts: Pop a dialog saying "This account has subaccounts. Delete them one-by-one or move them to a different parent account before deleting this one."

gnucash/gnome/gnc-plugin-page-account-tree.c
jralls: True about the subaccount choices. I think that `sa_trans_mas` is anyway misguided. It's unlikely that it would make sense to put all of the transactions from all subaccounts into a single account. It would be OK if it were restricted to a single account but that would also need the commodity check.

Good that you brought up cancellation: You don't handle it so cancelling will just re-run the dialog. That would be frustrating.

jeanlaroche:
> I'd think that if an account has sub-accounts it's too big an assumption that the user wants to delete them and move all of their transactions to the same account as the parent's. The simplest way to deal with that would be to refuse to delete an account with subaccounts: Pop a dialog saying "This account has subaccounts. Delete them one-by-one or move them to a different parent account before deleting this one."

Wait, deleting sub-accounts and moving transactions to another user-selected account is something that's currently available to the user in the account delete dialog. Are you saying we should remove this option?

gnucash/gnome/gnc-plugin-page-account-tree.c
```diff
@@ -1499,21 +1501,49 @@ gnc_plugin_page_account_tree_cmd_delete_account (GtkAction *action, GncPluginPag
...
+                // The currencies in the account that we're deleting and the account we're moving transactions to don't match.
+                char* message = g_strdup_printf (_("Account %s does not have the same currency as the account you're deleting.\nAre you sure you want to do this?"), gnc_account_get_full_name (ta));
+                GtkWidget *error_dialog = gtk_message_dialog_new (GTK_WINDOW(window),
+                                                                  GTK_DIALOG_DESTROY_WITH_PARENT,
+                                                                  GTK_MESSAGE_ERROR,
+                                                                  GTK_BUTTONS_NONE,
+                                                                  "%s", message);
+                gtk_dialog_add_buttons (GTK_DIALOG(error_dialog),
+                                        _("_Pick another account"), GTK_RESPONSE_CANCEL,
+                                        _("_Do it anyway"), GTK_RESPONSE_ACCEPT,
```
jralls, line 1539: As Dale Phurrough notes in his bug report, allowing a change in the split's commodity can make a mess even if the user thinks it's the right thing to do: Each split must be revalued in the transaction currency and if  trading accounts are in use those splits will need to be deleted and re-created with different trading accounts.

gnucash/gnome/gnc-plugin-page-account-tree.c
jralls: Actually, no, it's not true, I was right in the first place. I guess you were thinking of reading `sa_mas` and `sa_trans_mas` before entering the loop and I was thinking of after. Only the values on the last run of the dialog matter.
*But* `sa_trans_mas` should either be removed or should only be enabled if there's only one subaccount, and if it's enabled it should also get the commodity-equal check.

jeanlaroche: But how can I check the match of the currency if I don't read the selected account in the loop? I"m not understanding... 
So, should I change the behavior so we only allow moving sub-account transactions is there's only 1 sub-account? You don't expect people getting annoyed at this "regression"?

gnucash/gnome/gnc-plugin-page-account-tree.c
jralls: @christopherlam doesn't like out-params because it's not "functional". I don't mind them, though I think that returning a struct is better. I wonder if gcc does RVO on C or only on C++...

Even better, you could call the functions to move the accounts and transactions from the function that runs the dialog and return a gboolean to indicate whether the account deletion should proceed.

gnucash/gnome/gnc-plugin-page-account-tree.c
jralls: > But how can I check the match of the currency if I don't read the selected account in the loop? I"m not understanding...

This is only about `sa_mas` and `sa_trans_mas`. Only `trans_mas` is significant for the loop as it's written, though `sa_trans_mas` either needs the same treatment or it needs to be removed.

> So, should I change the behavior so we only allow moving sub-account transactions is there's only 1 sub-account? You don't expect people getting annoyed at this "regression"?

I don't expect that anyone would notice because it's such a spectacularly stupid thing to do that nobody would try it. I think there would be whining about the  more general reduction in features I suggested in another comment, disallowing deleting an account that has any subaccounts, but I do think that's the best approach. It doesn't prevent the user from shuffling their CoA at will, but it does greatly reduce the opportunity for them to blow up their book while doing so.

jeanlaroche:
I split the long function into two function (right after the user selects what to do with transactions/accounts). I also added code to reject deleting accounts with more than 1 sub-account. Added code to check matching currencies in all 3 cases (account is moved, sub-account is moved, sub-account transactions are moved)

gnucash/gnome/gnc-plugin-page-account-tree.c
christopherlam: I object to mutating arguments only for functions that are expected to be used in guile reports :wink:  I have an ongoing fight with `gncOwner.c` as a result; not only many functions mutate the `gncOwner *` argument, they also force me to `gncOwnerFree` it. :vomiting_face:

gnucash/gnome/gnc-plugin-page-account-tree.c
```diff
@@ -1348,6 +1348,158 @@ gppat_setup_account_selector (GtkBuilder *builder, GtkWidget *dialog,
...
+            xaccAccountCommitEdit (saa);
+        }
+        else if (NULL != sta)
+        {
+            /* Move the splits of its subaccounts, if any. */
+            gnc_account_foreach_descendant(account,
+                                           (AccountCb)xaccAccountMoveAllSplits,
+                                           sta);
+        }
+        if (NULL != ta)
```
christopherlam, line 1466:
```c
if (saa) ...
if (sta) ...
if (ta) ...
```

gnucash/gnome/gnc-plugin-page-account-tree.c
```diff
@@ -1521,14 +1692,22 @@ gnc_plugin_page_account_tree_cmd_delete_account (GtkAction *action, GncPluginPag
...
             {
-                // The currencies in the account that we're deleting and the account we're moving transactions to don't match.
-                char* message = g_strdup_printf (_("Account %s does not have the same currency as the account you're deleting.\nAre you sure you want to do this?"), gnc_account_get_full_name (ta));
+                char* message = NULL;
+                char* a_name = (!ta_match) ?  gnc_account_get_full_name (ta) : (!sta_match) ?  gnc_account_get_full_name (sta) :
+                                              gnc_account_get_full_name (saa);
+                message = g_strdup_printf (_("Account %s does not have the same currency as the account you're deleting.\nAre you sure you want to do this?"), a_name);
                 GtkWidget *error_dialog = gtk_message_dialog_new (GTK_WINDOW(window),
                                                                   GTK_DIALOG_DESTROY_WITH_PARENT,
                                                                   GTK_MESSAGE_ERROR,
```
Bob-IT: This is the travis failure, similar to the one in the other PR, move GtkWidget *error_dialog; to the beginning of the if branch and then change this line to error_dialog = ... and do not forget to realign the following lines.

gnucash/gnome/gnc-plugin-page-account-tree.c
```diff
@@ -1459,6 +1613,21 @@ gnc_plugin_page_account_tree_cmd_delete_account (GtkAction *action, GncPluginPag
                           GTK_WIDGET(gtk_builder_get_object (builder, "subaccount_trans")));
 
         // Does the selected account have sub accounts
+        if (gnc_account_n_children(account) > 1) {
+            // We don't delete accounts with more than 1 subaccount.
+            gchar* message = g_strdup_printf("Account %s has more than 1 subaccount, move subaccounts or delete them before attempting to delete this account.", acct_name);
```
Bob-IT, line 1618: Why not just use 'one' instead of '1'

Bob-IT:
One further comment, I might be inclined to change %s to '%s' where it involves account names, it may look better especially if they have spaces in them oh and other thing, you may like to add window titles of "Delete Account" to the dialogues you have added.

gnucash/gnome/gnc-plugin-page-account-tree.c
christopherlam: Agree, I'd vote to disallow this. The QIF importer handles this when importing accounts, if the currencies are incompatible, will simply create a new `AccountName 2` to receive the splits.

goodvibes2:
@jeanlaroche, I believe the usual practice is to assign (take) a bug to yourself when you start working on it so others know. I also add a bug comment when a PR is created so others can follow progress without having to manually check for new PRs. AFAIK, there is no way of automatically being informed of new PRs...

bhardwajs:
> 
> 
> @jeanlaroche, I believe the usual practice is to assign (take) a bug to yourself when you start working on it so others know. I also add a bug comment when a PR is created so others can follow progress without having to manually check for new PRs. AFAIK, there is no way of automatically being informed of new PRs...

If one sets the repo for `Watching`, one should get notifications for new PRs. I have done that for `gnucash` and `MSVC/STL` and get emails for each and every PR.

jeanlaroche:
My apologies, I must have forgotten to mention I had started on that 
bug... Hope I didn't waste too much of your time.

goodvibes2:
@jeanlaroche, No problem. I didn't get far and still learnt something. Silly me, forgetting about watching a repo!

gnucash/gnome/gnc-plugin-page-account-tree.c
```diff
@@ -1499,167 +1670,64 @@ gnc_plugin_page_account_tree_cmd_delete_account (GtkAction *action, GncPluginPag
...
-            xaccAccountDestroy (account);
-            gnc_resume_gui_refresh ();
-            gnc_unset_busy_cursor(NULL);
         }
-    }
-    g_free(acct_name);
+        gtk_widget_destroy(dialog);
+        g_list_free(filter);
+    } /* (NULL != splits) || (NULL != children) */
+    gnc_plugin_page_account_tree_cmd_delete_account_int (action, page, ta, sta, saa, acct_name, delete_res);
```
jralls, line 1730: `delete_res` is never changed from its initialization value `{FALSE, FALSE}`.
It's better to allocate and free things (`acct_name` in this case) in the same function and to pass them as `const` ptrs to helper functions so you don't have to chase around to make sure that all of the functions that are supposed to sink them do so. The obvious exception is returning an allocated object: In that case the function that ultimately uses the object considers itself the allocator as well.

You're still trying to do too much in one function.
```
if (confirm_account_delete_dialog (page, acct_name, ta, sta, saa, subaccount_has_splits))
{
    if (ta || sta || saa)
        do_move_splits (account, ta, sta, saa);
   do_delete_account (account);
}
``` 
would be a much clearer decomposition.

gnucash/gnome/gnc-plugin-page-account-tree.c
```diff
@@ -1499,167 +1670,64 @@ gnc_plugin_page_account_tree_cmd_delete_account (GtkAction *action, GncPluginPag
...
-            xaccAccountBeginEdit (account);
-            xaccAccountDestroy (account);
-            gnc_resume_gui_refresh ();
-            gnc_unset_busy_cursor(NULL);
         }
-    }
-    g_free(acct_name);
+        gtk_widget_destroy(dialog);
+        g_list_free(filter);
+    } /* (NULL != splits) || (NULL != children) */
```
jralls, line 1729: Needing a comment at the end of a block is a sure sign that it's too long.

gnucash/gnome/gnc-plugin-page-account-tree.c
```diff
@@ -1499,167 +1670,64 @@ gnc_plugin_page_account_tree_cmd_delete_account (GtkAction *action, GncPluginPag
...
+    gnc_plugin_page_account_tree_cmd_delete_account_int (action, page, ta, sta, saa, acct_name, delete_res);
 }
```
jralls, line 1731: Requiring functions to sink their parameters is a good way to have leaks or double-frees because it's too hard to keep track of whether they do or not. Pass `acct_name` as `const char*` and free it when the function returns... but see my note about allocating it where it's used since you're retrieving the account from the page there anyway. That way you could free it when it's last used in this function.

Aside from having an unwieldy name, `gnc_plugin_page_account_tree_cmd_delete_account_int` does too many things. A better decomposition would have
```
if (confirm_account_delete_dialog (page, ta, sta, saa, delete_res.has_splits))
{
    if (ta || sta || saa)
       move_subaccounts (account, ta, sta, saa);
    delete_account (account);
}
```

gnucash/gnome/gnc-plugin-page-account-tree.c
```diff
@@ -1348,6 +1348,158 @@ gppat_setup_account_selector (GtkBuilder *builder, GtkWidget *dialog,
     return selector;
 }
 
+// This is a helper function for gnc_plugin_page_account_tree_cmd_delete_account
+static void
+gnc_plugin_page_account_tree_cmd_delete_account_int (GtkAction *action, GncPluginPageAccountTree *page, Account* ta,
+                                                    Account* sta, Account* saa, gchar* acct_name, delete_helper_t delete_res)
+{
+    Account *account = gnc_plugin_page_account_tree_get_current_account (page);
```
jralls, line 1356: Since you're getting `account` here why not use it to generate `acct_name` locally instead of passing it in?

gnucash/gnome/gnc-plugin-page-account-tree.c
```diff
@@ -1348,6 +1348,158 @@ gppat_setup_account_selector (GtkBuilder *builder, GtkWidget *dialog,
     return selector;
 }
 
+// This is a helper function for gnc_plugin_page_account_tree_cmd_delete_account
+static void
+gnc_plugin_page_account_tree_cmd_delete_account_int (GtkAction *action, GncPluginPageAccountTree *page, Account* ta,
+                                                    Account* sta, Account* saa, gchar* acct_name, delete_helper_t delete_res)
```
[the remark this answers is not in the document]
jeanlaroche: I was instructed that helper functions tend to end with _int, but I guess I misunderstood that. I'll fix.

gnucash/gnome/gnc-plugin-page-account-tree.c
```diff
@@ -1348,6 +1348,158 @@ gppat_setup_account_selector (GtkBuilder *builder, GtkWidget *dialog,
...
+{
+    Account *account = gnc_plugin_page_account_tree_get_current_account (page);
+    GList* splits = xaccAccountGetSplitList(account);
+    GtkWidget* window = gnc_plugin_page_get_window(GNC_PLUGIN_PAGE(page));
+    gint response;
+    
+    /*
+     * Present a message to the user which specifies what will be
+     * deleted and what will be reparented, then ask for verification.
+     */
```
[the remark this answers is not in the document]
jeanlaroche: ah ah ah, this wasn't from me. Copied/moved from the existing code. But thanks for the scolding, much appreciated.

gnucash/gnome/gnc-plugin-page-account-tree.c
```diff
@@ -1348,6 +1348,158 @@ gppat_setup_account_selector (GtkBuilder *builder, GtkWidget *dialog,
     return selector;
 }
 
+// This is a helper function for gnc_plugin_page_account_tree_cmd_delete_account
```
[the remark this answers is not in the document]
jeanlaroche: That one's from me.

gnucash/gnome/gnc-plugin-page-account-tree.c
```diff
@@ -1348,6 +1348,158 @@ gppat_setup_account_selector (GtkBuilder *builder, GtkWidget *dialog,
...
+        {
+            name = gnc_account_get_full_name(ta);
+            format = _("All transactions in this account will be moved to "
+                       "the account %s.");
+            lines[++i] = g_strdup_printf(format, name);
+        }
+        else if (splits)
+        {
+            format = _("All transactions in this account will be deleted.");
+            lines[++i] = g_strdup_printf("%s", format);
```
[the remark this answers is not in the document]
jeanlaroche: Not from me either, but happy to fix.

gnucash/gnome/gnc-plugin-page-account-tree.c
jeanlaroche: Aggreed, comment is not from me either, but will fix by separating.

gnucash/gnome/gnc-plugin-page-account-tree.c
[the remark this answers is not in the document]
jralls: I haven't run across any '-int' names elswhere. There are some cases where we have a master static function called `gnc-foo-internal` that takes a bunch of options and public convenience functions `gnc-foo-bar`, `gnc-foo-baz`, etc. that call `gnc-foo-internal` with different option values.  A more common convention is to prefix public function names with the object name, e.g `gnc_numeric_add()` and private implementation functions without. In that convention GObject virtual functions do have the prefix but are still static. You won't find much of that in GnuCash code because there are few if any virtual functions.

gnucash/gnome/gnc-plugin-page-account-tree.c
jeanlaroche: `delete_res` can be changed during             `gnc_account_foreach_descendant_until(account,delete_account_helper, &delete_res);` if I'm not mistaken.

jralls: I did find that and thought I'd removed all of the comments about `delete_res`. Guess I missed one. Sorry.

jeanlaroche:
Addressed most of your comments, but I didn't break up the remaining long function. I attempted, but realized that I would have to pass so many parameters it would become quite ugly... Recall that this is a fix to a very silly bug. No point in sweating it too much.

jralls:
I don't see it as a silly bug at all. Transferring splits to an account having a different commodity wrecks the book: It becomes nearly impossible to get a passing trial balance afterwards.

As for "no point in sweating it too much", that is just never true. GnuCash is about user's money and they're understandably really twitchy about it. Plus, as I've told you before, GnuCash has more than enough technical debt. We can't afford to add any more and need to take every opportunity to reduce it.

jeanlaroche:
> I don't see it as a silly bug at all. Transferring splits to an account having a different commodity wrecks the book: It becomes nearly impossible to get a passing trial balance afterwards.
> 

OK fair enough. It's just not a bug that many people will run into. 

> As for "no point in sweating it too much", that is just never true. GnuCash is about user's money and they're understandably really twitchy about it. Plus, as I've told you before, GnuCash has more than enough technical debt. We can't afford to add any more and need to take every opportunity to reduce it.

The bug needs to be fixed, that's for sure. Now, the question is whether anybody should spent hours to cleanup and refactor code until you're happy with the result. To that question I answer no, there are more pressing things to fix than to refactor existing code. Not add to the mess = yes. Cleanup and beautify until you JR are happy with the result = not so much.

jralls:
It shouldn't take hours. What takes hours is analyzing the messy 400-line functions that try to do 5 different things, and turning them into 600-line functions that do 6 things makes it worse. It's faster to extract helper functions and once that's done it's generally easy to understand, to isolate the bug you were after or to add the functionality you want, and it has the added payoff that you don't have to spend all those analysis hours the next time you work on the function.

Too many parameters to pass? Make a struct and pass pointers to it.

Think the gatekeepers on a FOSS project are too strict? Either take it as a learning opportunity to make yourself a better software engineer or go find a project with less experienced leadership.

jeanlaroche:
> Think the gatekeepers on a FOSS project are too strict? Either take it as a learning opportunity to make yourself a better software engineer or go find a project with less experienced leadership.

Do you actually *read* what you write? "less experienced leadership", really?
You're one of a kind man!
In any case, I'm done with this PR. Feel free to reject.

christopherlam:
@jeanlaroche the issue is still code clarity and maintainability for future hackers.  You will have noticed *lots* of hackish code in the git (ex-svn) history from numerous contributors over the past 20+ years. All those past ones are now *gone*. I think you have skill and ability, and would be sorry to see you go. But we must also ensure any new code is as correct as we can be. If this means delaying new code until old code is refactored, then refactoring must take place (safely) before adding new code. See https://github.com/Gnucash/gnucash/pull/716#issuecomment-625326668 - I will fix the `wind->load_stop` behaviour before fixing this code. This is unfortunate but necessary. The function you're working on in this is 400 lines long, although it's acceptable to you, is unintelligible to the rest of us. It's definitely worthwhile breaking it down first, then add new code. Otherwise future hackers (or @jralls  who is still hacking after many years) will have a 600-line headache rather than a 400-line headache.

jeanlaroche:
Well here's my take on it.
- The function is messy, not doubt. Not the way I would have written it.
- I fixed the bug, *and* left the function in a cleaner state than it was before.
- Further breaking down the function for the sake of making it shorter is entirely cosmetic, and does not help readability or understandability. I just makes some people happier.
- Refactoring the entire thing (not just one function) so it's more logical *would* be better, but it's fraught with danger and the bang for the buck is zero.
- The question is, is it more valuable to have me spend my time making a "very experienced leadership" happier, or is it more valuable for me to spend my time fixing other bugs? The answer to that is very obvious to me.
- Any future hacker will do the same as I did. Scratch their head for a little while, figure it out, and hopefully leave the function in a nicer state yet. Leaving helpful comments in the code would actually be a good thing, but that's apparently frowned upon by the VEL.
- Most likely there will be a full rewrite of the entire code that will make this entire discussion completely moot.

So in short, I'm done rewriting code to make the VEL happy and nothing else. If that's no going to work with you devs, then that's just fine with me. I'll working on my fork and be done with it. It's too bad for the rest of the community but so be it. I understand the need for gatekeepers to prevent people from writing atrocious code, but it's not a "my way or the highway" thing, sorry.

jralls:
Merged into master and cleaned up.

PR #709, opened 2020-04-27
Bug 797388 - Add watch for each subacct of the acct to be reconciled

goodvibes2:
Enable the Reconcile window debit & credit unreconciled split
lists to be automatically updated when transactions for
subaccounts of the account to be reconciled are added, modified or
deleted. This already occurs for the reconciliation account.


Bob-IT:
Thank you.

PR #711, opened 2020-04-30
Fix the crash issue by unregistering the ui component when the window…

jeanlaroche:
… is destroyed.

This fix was happening in two separate PRs, creating a mess. It's now extracted into its own, still on master, even though it's really a bug fix.


Bob-IT:
You are correct that this should go to maint, so please rebase this PR on current maint.

PR #713, opened 2020-05-01
Bug 797648 Redo - Reconciliation - Treat each split independently

goodvibes2:
The previous change under this bug which propagated the status
change (reconcile or unreconcile) of any split for the account to
be reconciled (and its subaccounts) in a transaction, to all
splits for the account to be reconciled (and its subaccounts) in
the transaction, is incorrect. Each split needs to be able to be
checked or unchecked independently of any other split in the
transaction.
For instance, if an order delivery was split into 2 parts delivered and billed in different statement periods, the 2 splits need to be reconciled independently.

If anyone needs to reconcile or unreconcile multiple splits in a transaction, they can still do a multi-split selection, then use either the 'Reconcile Selection' or 'Unreconcile Selection' buttons. This is slightly more work, but we need the ability to reconcile splits independently.

See discussion under previous PR https://github.com/Gnucash/gnucash/pull/670

### DIFF
--- a/gnucash/gnome/reconcile-view.c
+++ b/gnucash/gnome/reconcile-view.c
@@ -519,92 +519,12 @@ gnc_reconcile_view_toggle_split (GNCReconcileView *view, Split *split)
 
 
 static void
-gnc_reconcile_view_toggle_children (Account *account, GNCReconcileView *view, Split *split)
-{
-    GList       *child_accounts, *node;
-    Transaction *transaction;
-
-    /*
-     * Need to get all splits in this transaction and identify any that are
-     * in the same hierarchy as the account being reconciled (not necessarily
-     * the account this split is from.)
-     *
-     * For each of these splits toggle them all to the same state.
-     */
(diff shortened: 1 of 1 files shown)


Bob-IT:
@goodvibes2 Chris is this just a PR to revert what you did in #670 ?
If so can you change the commit message to start "Revert Bug...., @christopherlam  did one last week, not sure how he did it but there is 'git revert'

goodvibes2:
Hi Bob-IT,
No, it is not just a revert. It removes the code that was there previously that propagates toggling ( fn  gnc_reconcile_view_toggle_children) and also my mods to change the toggling to propagation). I thought about using revert but decided against it, partly because I haven't used it before. I can do a revert, then another commit if you like?

Bob-IT:
Chris, I think that is a good idea, it makes it clear what has happened.

goodvibes2:
Hi @Bob-IT, I have reverted my previous changes and now the last commit is only the changes from before I started.

Bob-IT:
Thank you.

PR #716, opened 2020-05-05
Bug 797218 - Support QIF import with any encoding

christopherlam:
Inputs the encoding as a string. Better to use a `GtkComboBox`, but choose an appropriate encoding and QIF can import.

image
image

### DIFF
--- a/gnucash/import-export/qif-imp/CMakeLists.txt
+++ b/gnucash/import-export/qif-imp/CMakeLists.txt
@@ -8,6 +8,18 @@ set (qif_import_SOURCES
     gnc-plugin-qif-import.c
 )
 
+set (qif_import_remote_HEADERS
+  ${CMAKE_SOURCE_DIR}/borrowed/goffice/go-charmap-sel.h
+  ${CMAKE_SOURCE_DIR}/borrowed/goffice/go-optionmenu.h
+  ${CMAKE_SOURCE_DIR}/borrowed/goffice/go-glib-extras.h
+  )
+
+set (qif_import_remote_SOURCES
+  ${CMAKE_SOURCE_DIR}/borrowed/goffice/go-charmap-sel.c
+  ${CMAKE_SOURCE_DIR}/borrowed/goffice/go-optionmenu.c
+  ${CMAKE_SOURCE_DIR}/borrowed/goffice/go-glib-extras.c
+  )
+
 # Add dependency on config.h
 set_source_files_properties (${qif_import_SOURCES} PROPERTIES OBJECT_DEPENDS ${CONFIG_H})
 
@@ -17,7 +29,13 @@ set (qif_import_noinst_HEADERS
     gnc-plugin-qif-import.h
 )
 
-add_library	(gnc-qif-import ${qif_import_SOURCES} ${qif_import_noinst_HEADERS})
+add_library (gnc-qif-import
+  ${qif_import_noinst_HEADERS}
+  ${qif_import_remote_HEADERS}
+  ${qif_import_HEADERS}
+  ${qif_import_remote_SOURCES}
+  ${qif_import_SOURCES}
+)
 
 target_link_libraries(gnc-qif-import
     gnc-app-utils
@@ -31,6 +49,10 @@ target_include_directories(gnc-qif-import
 )
 
 target_compile_definitions(gnc-qif-import PRIVATE -DG_LOG_DOMAIN=\"gnc.import.qif.import\")
+
+target_include_directories (gnc-qif-import PRIVATE
+  ${CMAKE_SOURCE_DIR}/borrowed/goffice
+  )
 
 if (APPLE)
   set_target_properties (gnc-qif-import PROPERTIES INSTALL_NAME_DIR "${CMAKE_INSTALL_FULL_LIBDIR}/gnucash")
--- a/gnucash/import-export/qif-imp/CMakeLists.txt
+++ b/gnucash/import-export/qif-imp/CMakeLists.txt
@@ -6,19 +6,10 @@ set (qif_import_SOURCES
     dialog-account-picker.c
     assistant-qif-import.c
     gnc-plugin-qif-import.c
+    ${CMAKE_SOURCE_DIR}/borrowed/goffice/go-charmap-sel.c
+    ${CMAKE_SOURCE_DIR}/borrowed/goffice/go-optionmenu.c
+    ${CMAKE_SOURCE_DIR}/borrowed/goffice/go-glib-extras.c
 )
-
-set (qif_import_remote_HEADERS
-  ${CMAKE_SOURCE_DIR}/borrowed/goffice/go-charmap-sel.h
-  ${CMAKE_SOURCE_DIR}/borrowed/goffice/go-optionmenu.h
-  ${CMAKE_SOURCE_DIR}/borrowed/goffice/go-glib-extras.h
-  )
-
-set (qif_import_remote_SOURCES
-  ${CMAKE_SOURCE_DIR}/borrowed/goffice/go-charmap-sel.c
-  ${CMAKE_SOURCE_DIR}/borrowed/goffice/go-optionmenu.c
-  ${CMAKE_SOURCE_DIR}/borrowed/goffice/go-glib-extras.c
-  )
 
 # Add dependency on config.h
 set_source_files_properties (${qif_import_SOURCES} PROPERTIES OBJECT_DEPENDS ${CONFIG_H})
@@ -27,15 +18,12 @@ set (qif_import_noinst_HEADERS
     dialog-account-picker.h
     assistant-qif-import.h
     gnc-plugin-qif-import.h
+    ${CMAKE_SOURCE_DIR}/borrowed/goffice/go-charmap-sel.h
+    ${CMAKE_SOURCE_DIR}/borrowed/goffice/go-optionmenu.h
+    ${CMAKE_SOURCE_DIR}/borrowed/goffice/go-glib-extras.h
 )
 
-add_library (gnc-qif-import
-  ${qif_import_noinst_HEADERS}
-  ${qif_import_remote_HEADERS}
-  ${qif_import_HEADERS}
-  ${qif_import_remote_SOURCES}
-  ${qif_import_SOURCES}
-)
+add_library	(gnc-qif-import ${qif_import_SOURCES} ${qif_import_noinst_HEADERS})
 
 target_link_libraries(gnc-qif-import
     gnc-app-utils
--- a/gnucash/import-export/qif-imp/assistant-qif-import.c
+++ b/gnucash/import-export/qif-imp/assistant-qif-import.c
@@ -1877,7 +1877,7 @@ gnc_ui_qif_import_load_progress_start_cb (GtkButton * button,
     gnc_progress_dialog_pop (wind->load_progress);
     wind->ask_date_format = FALSE;
     wind->date_format = NULL;
-    }
+
     if (parse_return == SCM_BOOL_T)
     {
         /* Canceled by the user. */
@@ -1992,6 +1992,7 @@ gnc_ui_qif_import_load_progress_start_cb (GtkButton * button,
 
     gtk_widget_set_sensitive (wind->load_pause, FALSE);
     wind->busy = FALSE;
+    }
 
     if (wind->load_stop == FALSE)
     {
--- a/gnucash/import-export/qif-imp/assistant-qif-import.c
+++ b/gnucash/import-export/qif-imp/assistant-qif-import.c
@@ -1590,7 +1590,7 @@ gnc_ui_qif_import_load_file_complete (GtkAssistant  *assistant,
 
         /* See if the file is already loaded. */
         if (scm_call_2 (qif_file_loaded,
-                        scm_from_locale_string (path_to_load ? path_to_load : ""),
+                        scm_from_utf8_string (path_to_load ? path_to_load : ""),
                         wind->imported_files) == SCM_BOOL_T)
             gnc_error_dialog (GTK_WINDOW(assistant), "%s",
                               _("That QIF file is already loaded. "
@@ -1802,7 +1802,7 @@ gnc_ui_qif_import_load_progress_start_cb (GtkButton * button,
     gnc_progress_dialog_push (wind->load_progress, 0.7);
     load_return = scm_call_5 (qif_file_load,
                               SCM_CAR(imported_files),
-                              scm_from_locale_string (path_to_load ? path_to_load : ""),
+                              scm_from_utf8_string (path_to_load ? path_to_load : ""),
                               wind->ticker_map,
                               progress,
                               scm_from_utf8_string (qif_encoding));
@@ -3572,6 +3572,14 @@ void gnc_ui_qif_import_prepare_cb (GtkAssistant  *assistant, GtkWidget *page,
     }
 }
 
+inline static GtkWidget*
+make_encoding_widget (GObject* container)
+{
+    GtkWidget* encoding_widget = go_charmap_sel_new (GO_CHARMAP_SEL_TO_UTF8);
+    gtk_container_add (GTK_CONTAINER (container), encoding_widget);
+    return encoding_widget;
+}
+
 
 /********************************************************************
  * get_assistant_widgets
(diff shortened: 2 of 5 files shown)


Bob-IT:
There is an encoding widget used in the Transaction CSV import that you may be able to use. GOCharmapSel and I think that returns a string.

christopherlam:
Thanks @Bob-IT 

image
image

christopherlam:
Old code has an invalid-character stripping function which worked only on UTF-8; now if invalid encoding is selected, the resulting data is now as follows. Also rebased to master.

image

jralls:
Code looks generally OK. 

I'm not so sure about blindly accepting the user's encoding and carrying on to try to process the import. In particular not all encodings will render ASCII in a way that's usable to guile and so the tags might not be parsable. You might consider changing the port conversion strategy to `error` so that you'll get an exception if guile can't convert a character, at which point you could pop a dialog saying that there's been a conversion error from encoding xxx and offering the user the opportunity to go back and change the encoding or to bail out.

christopherlam:
Don't proceed unless the charset is acceptable.

image

christopherlam:
I think this branch must be stalled because of an assistant-qif-import.c bug -- `wind->load_stop` doesn't currently work well; my 406db87 changes the logic in the qif importer; without it the charset-failure stage moves directly to the qif-parsing stage. But I'm not sure if 406db87 is correct. I think assistant-qif-import.c must be broken down into several smaller functions, one for each step.

diablodale:
I continue to believe the workflow should include additional UI and steps. As I wrote in https://bugs.gnucash.org/show_bug.cgi?id=797218#c9 so I don't duplicate the conversation.

christopherlam:
@jralls would you have any objection to merging this? The QIF C importer part is still difficult for me to decode, but I think it maintains a good workflow and handles invalid-charset correctly. I'm not sure if multiple-QIF-import is handled correctly.

gnucash/import-export/qif-imp/CMakeLists.txt
```diff
@@ -8,6 +8,18 @@ set (qif_import_SOURCES
     gnc-plugin-qif-import.c
 )
 
+set (qif_import_remote_HEADERS
```
jralls, line 11: No need for a separate set of cmake variables, just add the files to `qif_import_FOO`.

gnucash/import-export/qif-imp/assistant-qif-import.c
```diff
@@ -1796,11 +1800,12 @@ gnc_ui_qif_import_load_progress_start_cb (GtkButton * button,
 
     /* This step will fill 70% of the bar. */
     gnc_progress_dialog_push (wind->load_progress, 0.7);
-    load_return = scm_call_4 (qif_file_load,
+    load_return = scm_call_5 (qif_file_load,
                               SCM_CAR(imported_files),
                               scm_from_locale_string (path_to_load ? path_to_load : ""),
```
jralls, line 1805: Not yours but probably wrong: path_to_load is coming from a Gtk widget and will be UTF8, not the locale encoding.

gnucash/import-export/qif-imp/assistant-qif-import.c
```diff
@@ -1987,6 +1992,7 @@ gnc_ui_qif_import_load_progress_start_cb (GtkButton * button,
 
     gtk_widget_set_sensitive (wind->load_pause, FALSE);
     wind->busy = FALSE;
+    }
```
jralls, line 1995: The `if` condition this brace closes is `!wind->load_stop` and the following line is a new if with the same condition, so if this is really right then you should remove it too. But it's a major change: You're making an additional 120 lines execute only if `!wind->load_stop` that previously executed regardless of `load_stop`'s value. Are you sure that's right?

Aside from that the whole block needs to be indented.

gnucash/import-export/qif-imp/assistant-qif-import.c
```diff
@@ -3579,6 +3585,11 @@ get_assistant_widgets (QIFImportWindow *wind, GtkBuilder *builder)
 
     wind->window             = GTK_WIDGET(gtk_builder_get_object (builder, "qif_import_assistant"));
     wind->filename_entry     = GTK_WIDGET(gtk_builder_get_object (builder, "qif_filename_entry"));
+    {
+        GtkWidget *enc_con = GTK_WIDGET (gtk_builder_get_object (builder, "encoding_container"));
+        wind->qif_encoding = go_charmap_sel_new (GO_CHARMAP_SEL_TO_UTF8);
+        gtk_container_add (GTK_CONTAINER (enc_con), wind->qif_encoding);
+    }
```
jralls, line 3592: Don't make blocks like this, it's very poor style. If you can't do this in the builder file make a helper function:
```C
inline static GtkWidget*
make_encoding_widget (GtkWidget* container)
{
    GtkWidget* encoding_widget = go_charmap_sel_new (GO_CHARMAP_SEL_TO_UTF8);
    gtk_container_add (GTK_CONTAINER (container), encoding_widget);
    return encoding_widget;
}
...
    wind->qif_encoding = make_encoding_widget (gtk_builder_get_object (builder, "encoding_container"));
```

gnucash/import-export/qif-imp/qif-file.scm
```diff
@@ -525,18 +507,32 @@
...
           (gnc-progress-dialog-set-value progress-dialog 1))
 
-        retval))))
+        retval))
+
+    (lambda (key . rest)
+      (let ((msg (string-append
+                  (case key
+                    ((decoding-error)
+                     (format #f (_ "Invalid charset ~a specified.") encoding))
```
jralls, line 530: "Invalid charset" implies that `encoding` contains a value that's not a legal charset, and that's not likely since it's selected with a chooser that should contain only valid charset names. It also doesn't really explain to the user what is the problem. Something like "The file contains values that aren't part of charset ~a. Please choose a different one." would be better.

gnucash/import-export/qif-imp/assistant-qif-import.c
christopherlam: > The `if` condition this brace closes is `!wind->load_stop` and the following line is a new if with the same condition, so if this is really right then you should remove it too. But it's a major change: You're making an additional 120 lines execute only if `!wind->load_stop` that previously executed regardless of `load_stop`'s value. Are you sure that's right?

This is from testing with and without the `if` condition changes. I think 755c42a has created a flaw in the workflow and this commit fixes it.

> Aside from that the whole block needs to be indented.

Agreed, will be reindented properly iff the `if` change is confirmed correct.

jralls: That's hand waving and you're not changing the condition, you're extending the block that it prevents from operating. What failed?

That aside, rather than extending the block you might as well just flip the test, set `wind->load_stop = FALSE;` and return.

gnucash/import-export/qif-imp/qif-file.scm
```diff
@@ -525,18 +507,33 @@
...
           (gnc-progress-dialog-set-value progress-dialog 1))
 
-        retval))))
+        retval))
+
+    (lambda (key . rest)
+      (let ((msg (string-append
+                  (case key
+                    ((decoding-error)
+                     (format #f (_ "The file contains values that are not \
```
christopherlam, line 530: s/values/characters :)

gnucash/import-export/qif-imp/assistant-qif-import.c
christopherlam: I find the C part confusing still. Maybe best I remove the C changes commit, expose the bug, and leave it for you?

jralls: No surprise, it's a really ugly function with far too many moving parts. But I can move a brace around, just tell me what is the bug!

christopherlam: Ok not at home atm to test; from memory if you apply this branch but revert 026ff3a0, and try load QIF with weird charsets available at https://bugs.gnucash.org/show_bug.cgi?id=797218, the incorrect-charset error isn't handled very well.

christopherlam: i.e. without 026ff3a:
select shift-JIS.qif for import, chose iso-8859-1
incorrect-charset appears twice (from memory)
try press back - can't remember if it works.

the qif-import mechanism also allows loading multiple qif files (eg. from same quicken book, load each qif-account, to achieve mass import) and I'm not 100% sure how it should handle multiple-qif-files, each with its own charset

gnucash/import-export/qif-imp/qif-file.scm
jralls: A bit odd leaving review comments on your own PR! ;-)

christopherlam: I reviewed myself at work and forgot to update!

gnucash/import-export/qif-imp/assistant-qif-import.c
jralls: With the braces as-is when I try to import shiftjis.qif using Unicode(UTF-8) the load completes. The message "The file contains characters that are not part of charset UTF-8. Please choose a different one." With the brace in its original position it also pops a message box saying "A bug was detected while parsing the QIF file." and repeats the same thing in the log window along with the first message. The back button works in either case.

shiftjis.qif imports with no error if you tell it Western(iso-8859-1), I suppose because there aren't any illegal bytes in iso-8859 charsets.

Incidentally, several of Phurrough's qif files fail on MacOS with a guile error "cannot convert narrow string to output locale" or "cannot convert wide string to output locale" from scm_to_stringn depending on what encoding one selects. Haven't tested on Windows yet.

christopherlam: Haven't seen this narrow/wide char particular issue. Would this error come up only when an invalid charset is chosen? If so then guile's exception handler can catch it as well:

```patch
modified   gnucash/import-export/qif-imp/qif-file.scm
@@ -526,7 +526,7 @@
     (lambda (key . rest)
       (let ((msg (string-append
                   (case key
-                    ((decoding-error)
+                    ((decoding-error encoding-error)
                      (format #f (G_ "The file contains characters that are not \
 part of charset ~a. Please choose a different one.") encoding))
                     ((system-error) (G_ "I/O system error. Try again."))
```

christopherlam:
P.S. this branch rebased to maint, and modified to handle guile-3 with `G_`. No other changes otherwise.

gnucash/import-export/qif-imp/assistant-qif-import.c
jralls: Well, this is disappointing: I set a breakpoint on the wide string to output locale error. Here's the bt:
```
(lldb) bt 16
* thread #1, queue = 'com.apple.main-thread', stop reason = breakpoint 2.1
  * frame #0: 0x00000001008d32ae libguile-2.2.1.dylib`scm_to_stringn(str=0x0000000151230b40, lenp=0x0000000000000000, encoding="ASCII", handler=SCM_FAILED_CONVERSION_ERROR) at strings.c:2270:39
    frame #1: 0x00000001008d2f91 libguile-2.2.1.dylib`scm_to_locale_stringn(str=0x0000000151230b40, lenp=0x0000000000000000) at strings.c:1909:10
    frame #2: 0x00000001008d2f49 libguile-2.2.1.dylib`scm_to_locale_string(str=0x0000000151230b40) at strings.c:1903:10
    frame #3: 0x0000000100917a17 libguile-2.2.1.dylib`scm_regexp_exec(rx=0x000000015128d4a0, str=0x0000000151230b40, start=0x0000000000000904, flags=0x0000000000000002) at regex-posix.c:274:11
    frame #4: 0x0000000100847901 libguile-2.2.1.dylib`scm_apply_subr(sp=0x000000011f4fbbf0, nslots=5) at gsubr.c:311:14
```

It's the same old UTF-8 conversion error passing through `scm_to_locale_string` problem that makes Guile unsuitable for string handling. You'll need to rewrite qif_file_load to C/C++.

christopherlam: You know very well that I can't write C well... if @derekatkins could not write a qif importer in C then there's no chance I can. Any idea which guile function is calling `scm_to_locale_string`?

jralls: Sorry, I thought it obvious. It's your attempt to pass `encoding` to `with-input-from-file` in `qif-file:read-file`. That only works on Linux because on MacOS the "locale" encoding is hard-coded to ASCII and on Windows to whatever is the active code page for CMD. You need to use C to read the contents of the file and transcode it to UTF8 and pass that to `scm-from-utf8-string`. *You ripped out the code that did that.*

christopherlam: Ok. Ripping out the code was meant to allow guile to throw a `decoding-error` exception to report via UI. But "cannot convert narrow/wide string to output locale" seems to be a result of `encoding-error` in guile, hence, the idea to catch it as well *may* work?

jralls: Sure, you can catch the exception. The problem is that it will get thrown *every time* you try to transcode on macOS or Windows because Guile's transcoding design using ports is irretrievably broken.

christopherlam: Does the exception get thrown if the correct charset is selected?

christopherlam: If there's a need for complete scm->C rewrite then I definitely won't be able to do it. If it's only 1-2 functions then maybe I can.

jralls: > Does the exception get thrown if the correct charset is selected?

Yes.

>  If it's only 1-2 functions then maybe I can.

There are a couple of possible approaches. One would be to extend the existing C transcoding functions that `qif-file:read-file` already calls to take a proposed encoding argument and raise errors that Scheme can pass back through `qif-file:read-file`. The other would be to rewrite `qif-file:read-file` in C, returning an SCM string to pass on to the importer. I think the second would be cleaner because it gets guile out of the middle for transcoding, error handling, and UI.

christopherlam: Ok -- this means handle the charset test and failure, and BOM stripping, all in C rather than SCM. It means the majority of `qif-file:read-file` to decode each line will remain in scheme. This much more reasonable.

jralls:
Pretty much. Instead of `qif-file:read-file` it would be something like `parse-qif-string-list`: The assistant would have a function to read in the whole file, transcode it into whatever Unicode representation Guile can inhale most efficiently, and create a list of SCM strings to pass to `parse-qif-string-list` to do the actual parsing and transaction creation. 

There are  three choices for transcoding: iconv, icu, and boost::locale. C++ stdlib also has a transcoding facility but it doesn't work very well. ICU is the most powerful and flexible, and boost::locale is really just a wrapper around ICU so I'd be inclined to use ICU. Oh, GLib has a wrapper around iconv as well.

christopherlam:
There's also `scm_from_stringn` in https://www.gnu.org/software/guile/manual/html_node/Conversion-to_002ffrom-C.html but I guess we'll skip guile's implementation 😆.

High level overview:

1. load file into buffer
```c
gsize length;
gchar *contents, *length;
g_file_get_contents(filename, &contents, &length, NULL);
```
2. pass `contents` through http://userguide.icu-project.org/conversion/converters using user-specified charset
3. then `scm_from_utf8_string`

christopherlam:
See latest commit. Managed to convert using glib. Loading ICU was too difficult. If invalid charset, g_assert fail :)

gnucash/import-export/qif-imp/assistant-qif-import.c
```diff
@@ -1800,9 +1800,30 @@ gnc_ui_qif_import_load_progress_start_cb (GtkButton * button,
...
+        gchar *contents;
+        gchar *utf8_string;
+        GError *gerror;
+
+        g_file_get_contents(path_to_load, &contents, &length, NULL);
+
+        utf8_string = g_convert (contents, length, "UTF-8",
+                                 qif_encoding, NULL, NULL, &gerror);
+
+        /* bwahaha. crash. */
```
jralls, line 1815: *Rude*

gnucash/import-export/qif-imp/assistant-qif-import.c
```diff
@@ -1800,9 +1800,30 @@ gnc_ui_qif_import_load_progress_start_cb (GtkButton * button,
...
+        g_assert (gerror == G_CONVERT_ERROR_NO_CONVERSION);
+
```
jralls, line 1817: I thought that the whole point of this exercise was to handle `G_CONVERT_ERROR_ILLEGAL_SEQUENCE`, but regardless if you don't handle the other errors you'll crash anyway from passing `NULL` to `scm_from_utf8_string`.

jralls:
This now crashes on your G_CONVERT_ERROR_NO_CONVERSION assert on every try on Mac. I'll debug later, maybe not until after the release this weekend.

gnucash/import-export/qif-imp/assistant-qif-import.c
christopherlam: Yes the point is to expose (and eventually handle) a conversion error in pure C, and I'll leave the handling & workflow rearrangement to someone more capable than me :)

gnucash/import-export/qif-imp/assistant-qif-import.c
christopherlam: This whole `{section}` will probably be moved into a static function.

gnucash/import-export/qif-imp/assistant-qif-import.c
jralls: OK, so close this and add a note on the bug that you've realized you're out of your depth on fixing it.

PR #717, opened 2020-05-08
Bug 797737 - Import matching can match an imported transaction to an …

jeanlaroche:
…existing, previously matched transaction

A simple fix that skips transactions that have an online_id in the matching process.
This fixes the issue, and also speeds up the import.

### DIFF
--- a/gnucash/import-export/import-backend.c
+++ b/gnucash/import-export/import-backend.c
@@ -627,6 +627,10 @@ static void split_find_match (GNCImportTransInfo * trans_info,
         Transaction *new_trans = gnc_import_TransInfo_get_trans (trans_info);
         Split *new_trans_fsplit = gnc_import_TransInfo_get_fsplit (trans_info);
 
+        // Do not consider transactions that have been previously matched.
+        if (gnc_import_split_has_online_id (split))
+            return;
+
         /* Matching heuristics */
 
         /* Amount heuristics */


jralls:
Thanks!

christopherlam:
Jean I haven't been able to thoroughly test your code; I trust you have. If I import ofx which contains previously imported splits, I trust that it will happily match to account splits with `online_id`?

jeanlaroche:
> 
> 
> Jean I haven't been able to thoroughly test your code; I trust you have. If I import ofx which contains previously imported splits, I trust that it will happily match to account splits with `online_id`?

This behavior isn't changed. If the OFX transaction was previously imported and matched, its online_id appears in one of the register transactions and the OFX transaction is simply ignored. It does not appear in the list of transactions to match. That's the same as it's always been.

christopherlam:
I think this causes https://bugs.gnucash.org/show_bug.cgi?id=798128

Can you verify?

christopherlam:
This is a fix.

```patch
@@ -1785,8 +1785,8 @@ create_hash_of_potential_matches (GList *candidate_txns,
     {
         Account* split_account;
         GSList* split_list;
-        if (gnc_import_split_has_online_id (candidate->data))
-            continue;
         split_account = xaccSplitGetAccount (candidate->data);
         /* g_hash_table_steal_extended would do the two calls in one shot but is
          * not available until GLib 2.58.
```

jeanlaroche:
But will this make it so an imported transaction can be matched to a register transaction that was previously matched?
Perhaps a more robust solution would be to create a fake online ID for imported transactions that don't have one, based on amounts, date, recipient etc, such that if you re-import the same transaction the generated online id will be identical, and will refuse to match a register transaction that was previously matched?

christopherlam:
> But will this make it so an imported transaction can be matched to a register transaction that was previously matched?

Yes that's the point. If I import the same CSV many times I expect the matcher to tell me it's a duplicate.

> Perhaps a more robust solution would be to create a fake online ID for imported transactions that don't have one, based on amounts, date, recipient etc, such that if you re-import the same transaction the generated online id will be identical, and will refuse to match a register transaction that was previously matched?

Not sure we're a fan of faking internal data. I could import CSV then later import ofx which has online_id.

jeanlaroche:
Oh I see, I read the comments in the bug more carefully. Faking the online ID would definitely not help in that case.
If we revert the change as you suggested, then this will make it so an imported transaction with ID=1 can be matched to a register transaction that have already been matched with ID=2 as if the register transaction had never been matched before.
This was the issue the change was put in to address: if you had similar transactions (i.e., coffee every morning), it was possible for coffee from 2 separate days to match a single coffee transaction in the register.

christopherlam:
Thank you for checking. So this change should be reverted and all will be well? Will the daily coffee situation be handled properly by your improved matcher?

jeanlaroche:
No, precisely, if we revert the change, the coffee situation will go back to being broken...
It's an unusual situation when you re-import a transaction and its ID has changed (or disappeared in your case). It has happened in the past (for example, because a bank didn't make it so the ID would remain the same for a particular transaction) and it's a pain.
So I don't know what's worse, breaking the coffee situation or breaking the re-import of CSV files after importing an ofx.

christopherlam:
We can justify either approach via methods: how often do users have regular daily identical txns _vs_ how often would CSV txns overlap?
- for ofx this point is moot because it sets online_id which ensures a perfect match
- for csv we have this problem

jeanlaroche:
One option is to make this user adjustable, but that's not great either because it's not that easy to understand/explain.
We could do a compromise:
- If the incoming transaction has an ID, then we keep things as before (refuse to match any transaction in the register that has an online ID because it's been matched already).
- If the incoming transaction has no ID, then we let anything match it the way it used to be done (based on amounts etc). Note that in that case, the coffee problem would exist, but I don't see a way to avoid the problem if nothing differentiates two consecutive identical transactions.

christopherlam:
The above seems very reasonable, unless another dev has further views.

christopherlam:
P.S. your approach would exceed my knowledge of this code, so, a PR would be very welcome...

PR #718, opened 2020-05-10
 Bug 797730 - Transaction matching can match multiple imported transa…

jeanlaroche:
…ctions to the same existing one

This is a simple solution to the conflict problem. It does not guarantee an optimal match (i.e. a match with the best combined match score) but it removes conflicts and finds an acceptable solution. The idea is simple:
- Start with the first imported transaction,
- Find all other imported transations that conflict with it (i.e., want to match the same register transaction).
- Of all these conflicting transactions, find the one with the best match to the register transaction
and set it as the winner: it keeps its match, and all other transactions go to their next favorite match.
- If any conflict happened, start back from the top as new conflicts can arise when imported transactions lose their top match and go to their second best one (if any).

This loop is guaranteed to terminate because every time we go to the top, we remove at least one match from the sum of all matches.

### DIFF
--- a/gnucash/import-export/import-backend.c
+++ b/gnucash/import-export/import-backend.c
@@ -113,6 +113,22 @@ gnc_import_TransInfo_get_match_list (const GNCImportTransInfo *info)
     return info->match_list;
 }
 
+void
+gnc_import_TransInfo_set_match_list (GNCImportTransInfo *info, GList* match_list)
+{
+    g_assert (info);
+    info->match_list = match_list;
+    if (match_list)
+    {
+        info->selected_match_info.selected_match = match_list->data;
+    }
+    else
+    {
+        info->selected_match_info.selected_match = NULL;
+        gnc_import_TransInfo_set_action (info, GNCImport_ADD);
+    }
+}
+
 Transaction *
 gnc_import_TransInfo_get_trans (const GNCImportTransInfo *info)
 {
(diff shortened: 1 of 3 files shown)


jeanlaroche:
> Looks generally pretty good. I'd like to merge it in tandem with #721

Does that mean I should merge them together as a single PR?

jralls:
No need to unless you think that merging them serially would create conflicts.

jeanlaroche:
There's a conflict I need to fix here...

jralls:
Thanks!

jeanlaroche:
John, this isn't going to work. The function `resolve_conflicts` is declared and implemented, but never called... Gonna have to reopen this.

jralls:
Hmm, looks like it got eaten in the merge.

jralls:
It got stepped on by the #721 change to `gnc_gen_trans_list_show_all`. Is this right:
```
void gnc_gen_trans_list_show_all(GNCImportMainMatcher *info)
{
    gnc_gen_trans_list_create_matches (info);
    resolve_conflicts (info);
    gtk_widget_show_all (GTK_WIDGET (info->main_widget));
}
```
?

jeanlaroche:
I just checked, put the line back, compiled and tested and yes, it's right. This works.
But I wish I had unit tests for all these changes to make 100% sure that I'm not breaking anything. This is working without a net, and I don't like it. But from what I see, writing a quick unit test isn't easy at all, unless I'm mistaken. The mock stuff seems pretty cumbersome...

jralls:
OK, I commited and pushed that.

It's not that it's hard to write a quick unit test, it's that the functionality here is really hard to test in a meaningful way without a ton of set up.

christopherlam:
Hi, @jeanlaroche sorry to tag you, there's https://bugs.gnucash.org/show_bug.cgi?id=798887 which involves conflicting transactions.

jeanlaroche:
Thanks for bringing that to my attention. Not sure when I'll have some 
time to take a look but I'll try

Jean

PR #719, opened 2020-05-11
[trep-engine] add option Filter/Transaction Filter is case insensitive

christopherlam:
add Filter option to make Transaction Filter case insensitive. Into maint or master?


jralls:
Master.

christopherlam:
Any comments on the option names? @fellen 

```
name "Transaction Filter is case insensitive"
popup "If this option is selected, transactions matching filter is not case sensitive."
```

PR #720, opened 2020-05-11
Bug 797743 - Monetary amounts are occasionally rendered in fractions

christopherlam:
If monetary is already in decimal, don't convert. if monetary is exact e.g. 3/7 then round to its SCU.

As a result, monetary amounts in an arbitrary precision will be displayed unchanged, e.g. `$0.1442`, whereas exact monetary amounts will be displayed using the currency's SCU e.g. $1/3 -> `$0.33`


christopherlam:
This fix will presume that there's _no_ use case for rendering monetary amounts in fractions. Agreed?

christopherlam:
This branch will eventually convert price renderers.

Prices are generally identical to gnc-monetary objects, so, reports will need to call `(gnc:default-price-renderer price)` where `price` is a `gnc-monetary`, to be rendered either exact `$1 + 2/3` or inexact `$1.6667`

christopherlam:
The idea for this PR has changed.

Any `(gnc:make-gnc-monetary comm amount)` will be rendered in decimal using the commodity SCU.

Reports which need to render prices are modified; previously they were rendering monetary objects as above and are changed to use a customised price renderer `(gnc:default-price-renderer currency amount)` whereby `curr` refers to the target currency.

christopherlam:
Merged into `master`

PR #721, opened 2020-05-12
Speeds up the import of ofx files by only doing one query at the end.

jeanlaroche:
The previous ofx import code performed one query for each imported transaction, which
was quite slow. The change consists of gathering all ofx transactions before doing the
query. The query must be wider to search for all matching accounts (in case the imported
transactions come from different accounts) and an enlarged date range (according to the
earliest and latest imported transaction). The rest of the code is identical to what was
done before. The final query is performed just before the matching dialog is displayed.

### DIFF
--- a/gnucash/import-export/import-backend.c
+++ b/gnucash/import-export/import-backend.c
@@ -607,7 +607,7 @@ matchmap_store_destination (GncImportMatchMap *matchmap,
 
 /** @brief The transaction matching heuristics are here.
  */
-static void split_find_match (GNCImportTransInfo * trans_info,
+void split_find_match (GNCImportTransInfo * trans_info,
                               Split * split,
                               gint display_threshold,
                               double fuzzy_amount_difference)
@@ -820,66 +820,6 @@ static void split_find_match (GNCImportTransInfo * trans_info,
                            match_info);
     }
 }/* end split_find_match */
-
-
-/** /brief Iterate through all splits of the originating account of the given
-   transaction, and find all matching splits there. */
-void gnc_import_find_split_matches(GNCImportTransInfo *trans_info,
-                                   gint process_threshold,
-                                   double fuzzy_amount_difference,
-                                   gint match_date_hardlimit)
-{
-    GList * list_element;
-    Query *query = qof_query_create_for(GNC_ID_SPLIT);
-    g_assert (trans_info);
-
-    /* Get list of splits of the originating account. */
-    {
-        /* We used to traverse *all* splits of the account by using
-           xaccAccountGetSplitList, which is a bad idea because 90% of these
-           splits are outside the date range that is interesting. We should
-           rather use a query according to the date region, which is
-           implemented here.
-        */
-        Account *importaccount =
-            xaccSplitGetAccount (gnc_import_TransInfo_get_fsplit (trans_info));
-        time64 download_time = xaccTransGetDate (gnc_import_TransInfo_get_trans (trans_info));
-
-        qof_query_set_book (query, gnc_get_current_book());
-        xaccQueryAddSingleAccountMatch (query, importaccount,
-                                        QOF_QUERY_AND);
-        xaccQueryAddDateMatchTT (query,
-                                 TRUE, download_time - match_date_hardlimit * 86400,
-                                 TRUE, download_time + match_date_hardlimit * 86400,
-                                 QOF_QUERY_AND);
-        list_element = qof_query_run (query);
-        /* Sigh. Doesn't help too much. We still create and run one query
-           for each imported transaction. Maybe it would improve
-           performance further if there is one single (master-)query at
-           the beginning, matching the full date range and all accounts in
-           question. However, this doesn't quite work because this function
-           here is called from each gnc_gen_trans_list_add_trans(), which
-           is called one at a time. Therefore the whole importer would
-           have to change its behaviour: Accept the imported txns via
-           gnc_gen_trans_list_add_trans(), and only when
-           gnc_gen_trans_list_run() is called, then calculate all the
-           different match candidates. That's too much work for now.
-        */
-    }
-
-    /* Traverse that list, calling split_find_match on each one. Note
-       that xaccAccountForEachSplit is declared in Account.h but
-       implemented nowhere :-( */
-    while (list_element != NULL)
-    {
-        split_find_match (trans_info, list_element->data,
-                          process_threshold, fuzzy_amount_difference);
-        list_element = g_list_next (list_element);
-    }
-
-    qof_query_destroy (query);
-}
-
 
 /***********************************************************************
  */
@@ -1221,13 +1161,6 @@ gnc_import_TransInfo_init_matches (GNCImportTransInfo *trans_info,
 {
     GNCImportMatchInfo * best_match = NULL;
     g_assert (trans_info);
-
-
-    /* Find all split matches in originating account. */
-    gnc_import_find_split_matches(trans_info,
-                                  gnc_import_Settings_get_display_threshold (settings),
-                                  gnc_import_Settings_get_fuzzy_amount (settings),
-                                  gnc_import_Settings_get_match_date_hardlimit (settings));
 
     if (trans_info->match_list != NULL)
     {
--- a/gnucash/import-export/import-backend.c
+++ b/gnucash/import-export/import-backend.c
@@ -623,10 +623,6 @@ void split_find_match (GNCImportTransInfo * trans_info,
         Transaction *new_trans = gnc_import_TransInfo_get_trans (trans_info);
         Split *new_trans_fsplit = gnc_import_TransInfo_get_fsplit (trans_info);
 
-        // Do not consider transactions that have been previously matched.
-        if (gnc_import_split_has_online_id (split)) // JEAN THIS CAN NOW BE REMOVED.
-            return;
-
         /* Matching heuristics */
 
         /* Amount heuristics */
@@ -1081,7 +1077,7 @@ static gint check_trans_online_id(Transaction *trans1, void *user_data)
     }
 }
 
-static gint collect_trans_online_id(Transaction *trans, void *user_data) //JEAN COLLECT
+static gint collect_trans_online_id(Transaction *trans, void *user_data)
 {
     Split *split;
     GHashTable *id_hash = user_data;
@@ -1101,7 +1097,7 @@ static gint collect_trans_online_id(Transaction *trans, void *user_data) //JEAN 
 
 /** Checks whether the given transaction's online_id already exists in
   its parent account. */
-gboolean gnc_import_exists_online_id (Transaction *trans, GHashTable* id_hash)
+gboolean gnc_import_exists_online_id (Transaction *trans, GHashTable* acct_id_hash)
 {
     gboolean online_id_exists = FALSE;
     Account *dest_acct;
@@ -1113,15 +1109,13 @@ gboolean gnc_import_exists_online_id (Transaction *trans, GHashTable* id_hash)
     // Create a hash per account of a hash of all transactions IDs. Then the test below will be fast if
     // we have many transactions to import.
     dest_acct = xaccSplitGetAccount (source_split);
-    if (!g_hash_table_contains (id_hash, dest_acct))
+    if (!g_hash_table_contains (acct_id_hash, dest_acct))
     {
         GHashTable* new_hash = g_hash_table_new (g_str_hash, g_str_equal);
-        g_hash_table_insert (id_hash, dest_acct, new_hash);
-        printf ("CREATE HASH\n");
+        g_hash_table_insert (acct_id_hash, dest_acct, new_hash);
         xaccAccountForEachTransaction (dest_acct, collect_trans_online_id, new_hash);
-        printf ("CREATE DONE\n");
     }
-    online_id_exists = g_hash_table_contains (g_hash_table_lookup (id_hash, dest_acct),
+    online_id_exists = g_hash_table_contains (g_hash_table_lookup (acct_id_hash, dest_acct),
                                               gnc_import_get_split_online_id (source_split));
     
     /* If it does, abort the process for this transaction, since it is
--- a/gnucash/import-export/import-backend.c
+++ b/gnucash/import-export/import-backend.c
@@ -426,8 +426,6 @@ TransactionGetTokens(GNCImportTransInfo *info)
     time64 transtime;
     struct tm *tm_struct;
     char local_day_of_week[16];
-    Split* split;
-    int split_index;
 
     g_return_val_if_fail (info, NULL);
     if (info->match_tokens) return info->match_tokens;
@@ -458,12 +456,10 @@ TransactionGetTokens(GNCImportTransInfo *info)
     tokens = g_list_prepend(tokens, g_strdup(local_day_of_week));
 
     /* make tokens from the memo of each split of this transaction */
-    split_index = 0;
-    while ((split = xaccTransGetSplit(transaction, split_index)))
+    for (GList *split=xaccTransGetSplitList (transaction); split; split=split->next)
     {
-        text = xaccSplitGetMemo(split);
+        text = xaccSplitGetMemo(split->data);
         tokens = tokenize_string(tokens, text);
-        split_index++; /* next split */
     }
 
     /* remember the list of tokens for later.. */
@@ -628,7 +624,7 @@ void split_find_match (GNCImportTransInfo * trans_info,
         Split *new_trans_fsplit = gnc_import_TransInfo_get_fsplit (trans_info);
 
         // Do not consider transactions that have been previously matched.
-        if (gnc_import_split_has_online_id (split))
+        if (gnc_import_split_has_online_id (split)) // JEAN THIS CAN NOW BE REMOVED.
             return;
 
         /* Matching heuristics */
@@ -1085,9 +1081,27 @@ static gint check_trans_online_id(Transaction *trans1, void *user_data)
     }
 }
 
+static gint collect_trans_online_id(Transaction *trans, void *user_data) //JEAN COLLECT
+{
+    Split *split;
+    GHashTable *id_hash = user_data;
+    int i=0;
+    
+    const gchar* online_id = gnc_import_get_trans_online_id (trans);
+    if (online_id)
+        g_hash_table_add (id_hash, (void*) online_id);
+
+    for (GList *splits = xaccTransGetSplitList (trans); splits; splits = splits->next)
+    {
+        if (gnc_import_split_has_online_id (splits->data))
+            g_hash_table_add(id_hash, (void*) gnc_import_get_split_online_id (splits->data));
+    }
+    return 0;
+}
+
 /** Checks whether the given transaction's online_id already exists in
   its parent account. */
-gboolean gnc_import_exists_online_id (Transaction *trans)
+gboolean gnc_import_exists_online_id (Transaction *trans, GHashTable* id_hash)
 {
     gboolean online_id_exists = FALSE;
     Account *dest_acct;
@@ -1096,13 +1110,20 @@ gboolean gnc_import_exists_online_id (Transaction *trans)
     /* Look for an online_id in the first split */
     source_split = xaccTransGetSplit(trans, 0);
     g_assert(source_split);
-
-    /* DEBUG("%s%d%s","Checking split ",i," for duplicates"); */
-    dest_acct = xaccSplitGetAccount(source_split);
-    online_id_exists = xaccAccountForEachTransaction(dest_acct,
-                       check_trans_online_id,
-                       source_split);
-
+    // Create a hash per account of a hash of all transactions IDs. Then the test below will be fast if
+    // we have many transactions to import.
+    dest_acct = xaccSplitGetAccount (source_split);
+    if (!g_hash_table_contains (id_hash, dest_acct))
+    {
+        GHashTable* new_hash = g_hash_table_new (g_str_hash, g_str_equal);
+        g_hash_table_insert (id_hash, dest_acct, new_hash);
+        printf ("CREATE HASH\n");
+        xaccAccountForEachTransaction (dest_acct, collect_trans_online_id, new_hash);
+        printf ("CREATE DONE\n");
+    }
+    online_id_exists = g_hash_table_contains (g_hash_table_lookup (id_hash, dest_acct),
+                                              gnc_import_get_split_online_id (source_split));
+    
     /* If it does, abort the process for this transaction, since it is
        already in the system. */
     if (online_id_exists == TRUE)
--- a/gnucash/import-export/import-backend.h
+++ b/gnucash/import-export/import-backend.h
@@ -66,34 +66,20 @@ typedef enum _action
  * online_id. */
 gboolean gnc_import_exists_online_id (Transaction *trans);
 
-/** Iterate through all splits of the originating account of the given
- * transaction, find all matching splits there, and store them in the
- * GNCImportTransInfo structure.
+/** Evaluates the match between trans_info and split using the provided parameters.
  *
- * @param trans_info The TransInfo for which the corresponding
- * matching existing transactions should be found.
+ * @param trans_info The TransInfo for the imported transaction
  *
- * @param process_threshold Each match whose heuristics are smaller
- * than this value is totally ignored.
+ * @param split The register split that should be evaluated for a match.
  *
- * @param fuzzy_amount_difference For fuzzy amount matching, a certain
- * fuzzyness in the matching amount is allowed up to this value. May
- * be e.g. 3.00 dollars for ATM fees, or 0.0 if you only want to allow
- * exact matches.
+ * @param display_threshold Minimum match score to include split in the list of matches.
  *
- * @param match_date_hardlimit The number of days that a matching
- * split may differ from the given transaction before it is discarded
- * immediately. In other words, any split that is more distant from
- * the given transaction than this match_date_hardlimit days will be
- * ignored altogether. For use cases without paper checks (e.g. HBCI),
- * values like 14 (days) might be appropriate, whereas for use cases
- * with paper checks (e.g. OFX, QIF), values like 42 (days) seem more
- * appropriate.
+ * @param fuzzy_amount_difference Maximum amount difference to consider the match good.
  */
-void gnc_import_find_split_matches(GNCImportTransInfo *trans_info,
-                                   gint process_threshold,
-                                   double fuzzy_amount_difference,
-                                   gint match_date_hardlimit);
+void split_find_match (GNCImportTransInfo * trans_info,
+                       Split * split,
+                       gint display_threshold,
+                       double fuzzy_amount_difference);
 
 /** Iterates through all splits of the originating account of
  * trans_info. Sorts the resulting list and sets the selected_match
--- a/gnucash/import-export/import-backend.h
+++ b/gnucash/import-export/import-backend.h
@@ -64,7 +64,7 @@ typedef enum _action
  *
  * @param trans The transaction for which to check for an existing
  * online_id. */
-gboolean gnc_import_exists_online_id (Transaction *trans);
+gboolean gnc_import_exists_online_id (Transaction *trans, GHashTable* Foo);
 
 /** Evaluates the match between trans_info and split using the provided parameters.
  *
--- a/gnucash/import-export/import-backend.h
+++ b/gnucash/import-export/import-backend.h
@@ -64,7 +64,7 @@ typedef enum _action
  *
  * @param trans The transaction for which to check for an existing
  * online_id. */
-gboolean gnc_import_exists_online_id (Transaction *trans, GHashTable* id_hash);
+gboolean gnc_import_exists_online_id (Transaction *trans, GHashTable* acct_id_hash);
 
 /** Evaluates the match between trans_info and split using the provided parameters.
  *
--- a/gnucash/import-export/import-backend.h
+++ b/gnucash/import-export/import-backend.h
@@ -64,7 +64,7 @@ typedef enum _action
  *
  * @param trans The transaction for which to check for an existing
  * online_id. */
-gboolean gnc_import_exists_online_id (Transaction *trans, GHashTable* Foo);
+gboolean gnc_import_exists_online_id (Transaction *trans, GHashTable* id_hash);
 
 /** Evaluates the match between trans_info and split using the provided parameters.
  *
--- a/gnucash/import-export/import-main-matcher.c
+++ b/gnucash/import-export/import-main-matcher.c
@@ -75,7 +75,7 @@ struct _main_matcher_info
     gboolean add_toggled;   // flag to indicate that add has been toggled to stop selection
     gint id;
     GSList* temp_trans_list; // Temporary list of imported transactions
-    GHashTable* id_hash;     // Hash table, per account, of list of transaction IDs.
+    GHashTable* acct_id_hash;     // Hash table, per account, of list of transaction IDs.
     GSList* edited_accounts; // List of accounts currently edited.
 };
 
@@ -147,36 +147,29 @@ static void refresh_model_row (
 
 static gboolean delete_hash (gpointer key, gpointer value, gpointer user_data)
 {
-    // Value is a hash table that needs to be destroyed. JEAN: I should free the memory used for the keys.
+    // Value is a hash table that needs to be destroyed. 
     g_hash_table_destroy (value);
     return TRUE;
 }
 
 static void
-commit_all (GNCImportMainMatcher *info)
+update_all_balances (GNCImportMainMatcher *info)
 {
-    // Commit the edits for all accounts for which we've called inc_account_edit.
     for (GSList* iter = info->edited_accounts; iter; iter=iter->next)
     {
-        int level = qof_instance_get_editlevel (iter->data);
-        // This should not happen.
-        if (level != 1)
-            PERR ("import-main-macher.c: Edit level not 1");
-        while (qof_instance_get_editlevel (iter->data) > 0)
-            xaccAccountCommitEdit (iter->data);
+        gnc_account_set_defer_bal_computation(iter->data,FALSE);
+        xaccAccountRecomputeBalance(iter->data);
     }
     g_slist_free (info->edited_accounts);
     info->edited_accounts = NULL;
 }
 
 static void
-inc_account_edit (GNCImportMainMatcher *info, Account* acc)
+defer_bal_computation (GNCImportMainMatcher *info, Account* acc)
 {
-    // Call BeginEdit on accounts that haven't been edited yet. This will speeds things up
-    // a lot by preventing commits until we're all done.
-    if (qof_instance_get_editlevel (acc) == 0)
+    if (!gnc_account_get_defer_bal_computation(acc))
     {
-        xaccAccountBeginEdit (acc);
+        gnc_account_set_defer_bal_computation (acc, TRUE);
         info->edited_accounts = g_slist_prepend (info->edited_accounts, acc);
     }
 }
@@ -186,7 +179,6 @@ void gnc_gen_trans_list_delete (GNCImportMainMatcher *info)
     GtkTreeModel *model;
     GtkTreeIter iter;
     GNCImportTransInfo *trans_info;
-    Split* first_split = NULL;
 
     if (info == NULL)
         return;
@@ -219,14 +211,14 @@ void gnc_gen_trans_list_delete (GNCImportMainMatcher *info)
     else
         gnc_import_Settings_delete (info->user_settings);
 
-    // We've called xaccAccountBeginEdit on many accounts to defer the commits until now.
-    // Let's do it now that we're done.
-    commit_all (info);
-
     g_slist_free_full (info->temp_trans_list,(GDestroyNotify) gnc_import_TransInfo_delete);
     info->temp_trans_list = NULL;
-    g_hash_table_foreach_remove (info->id_hash, delete_hash, NULL);
-    info->id_hash = NULL;
+    
+    // We've deferred balance computations on many accounts. Let's do it now that we're done.
+    update_all_balances (info);
+    
+    g_hash_table_foreach_remove (info->acct_id_hash, delete_hash, NULL);
+    info->acct_id_hash = NULL;
     g_free (info);
 }
 
@@ -269,15 +261,14 @@ on_matcher_ok_clicked (GtkButton *button, GNCImportMainMatcher *info)
     /* Don't run any queries and/or split sorts while processing the matcher
     results. */
     gnc_suspend_gui_refresh();
-    printf("START OK LOOP\n");
     do
     {
         gtk_tree_model_get (model, &iter,
                             DOWNLOADED_COL_DATA, &trans_info,
                             -1);
 
-        // JEAN: the account of all splits should be set as edited. IF THERE'S ONLY 1 SPLIT, A SPLIT
-        // WILL BE CREATED with unbalanced, and the account will not have its edit level increased...
+        // Note: if there's only 1 split (unbalanced) one will be created with the unbalanced account,
+        // and for that account the defer balance will not be set. So things will be slow.
 
         if (gnc_import_process_trans_item (NULL, trans_info))
         {
@@ -289,7 +280,6 @@ on_matcher_ok_clicked (GtkButton *button, GNCImportMainMatcher *info)
         }
     }
     while (gtk_tree_model_iter_next (model, &iter));
-    printf("DONE OK LOOP\n");
     
     gnc_gen_trans_list_delete (info);
 
@@ -392,8 +382,8 @@ run_account_picker_dialog (GNCImportMainMatcher *info,
              &ok_pressed);
     if (ok_pressed)
     {
-        gnc_import_TransInfo_set_destacc (trans_info, new_acc, TRUE); // JEAN: BeginEdit on this account
-        inc_account_edit (info, new_acc);
+        gnc_import_TransInfo_set_destacc (trans_info, new_acc, TRUE);
+        defer_bal_computation (info, new_acc);
     }
 }
 
@@ -547,8 +537,8 @@ gnc_gen_trans_assign_transfer_account (GtkTreeView *treeview,
                 }
                 if (ok_pressed)
                 {
-                    gnc_import_TransInfo_set_destacc (trans_info, *new_acc, TRUE); // JEAN Set Begin_edit on this account.
-                    inc_account_edit (info, *new_acc);
+                    gnc_import_TransInfo_set_destacc (trans_info, *new_acc, TRUE);
+                    defer_bal_computation (info, *new_acc);
                 }
             }
             break;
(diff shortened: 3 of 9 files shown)


gnucash/import-export/import-main-matcher.c
```diff
@@ -1352,26 +1354,116 @@ void gnc_gen_trans_list_add_trans_with_ref_id (GNCImportMainMatcher *gui, Transa
...
+
+    query = qof_query_create_for (GNC_ID_SPLIT);
+    // Gather the list of splits that could match one of the imported transactions
+    list_element = create_list_of_potential_matches (model, query, match_date_hardlimit);
+
+    // For each imported transaction, go through list_element, select the ones that have the
+    // right account, and evaluate how good a match they are.
+    valid = gtk_tree_model_get_iter_first (model, &iter);
+    while(valid)
+    {
```
jralls, line 1430: A `for` loop (see below) is better, but `gtk_tree_model_foreach()` would be better still.

gnucash/import-export/import-main-matcher.c
```diff
@@ -1352,26 +1354,116 @@ void gnc_gen_trans_list_add_trans_with_ref_id (GNCImportMainMatcher *gui, Transa
...
+                            DOWNLOADED_COL_DATA, &transaction_info,
+                            -1);
+        importaccount = xaccSplitGetAccount (gnc_import_TransInfo_get_fsplit (transaction_info));
+
+        iter_2 = list_element;
+        while (iter_2 != NULL)
+        {
+            // One issue here is that the accounts may not match since the query was
+            // called for all accounts so we need to check here.
+            Account* split_account = xaccSplitGetAccount(iter_2->data);
```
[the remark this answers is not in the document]
jeanlaroche: Do you have a hunch as to whether that's a significant enough gain to warrant the added complexity? Actually it's not that much more complex, I'll give it a shot.

gnucash/import-export/import-main-matcher.c
jeanlaroche: gtk_tree_model_foreach() requires defining a function, IMHO the added function obscures the code a bit, for a gain that's pretty minimal. But if you insist, I'll do it, no problem.

jralls: We use the `foo_for_each` idiom quite a bit in GnuCash, and the STL equivalent is what's taught as The Right Thing To Do™ in the C++ universe nowadays. In general decomposing to functions and algorithms has been considered a fundamental [Good Thing™](https://martinfowler.com/books/refactoring.html) for at least 20 years.

Unlike the STL case there's probably no performance advantage over using a naked loop so you can take it as a suggestion instead of requirement. If you decide not to then do replace the `while`s with `for`s.

christopherlam: I also prefer `for-each` much more than `do..while` or `while`.
This branch will also speed up all imports except QIF, right?

bhardwajs: > Unlike the STL case there's probably no performance advantage over using a naked loop
Not sure I understand you @jralls - are you saying in C++ `std::for_each` might introduce a performance penalty in comparison to naked loop? It will be a bit of surprise to see that in optimized builds.

jeanlaroche: Here's my beef with gtk_tree_model_foreach():  I need to pass a bunch of auxiliary data to the function that will be called, but GTK only provides a single pointer. So I'd have to pack everything into a structure, and pass the pointer to that structure. On the function side, I'd need to unpack the structure into the data I need.
That, to me, is ugly as hell, and bloats the code. Not sure about the stl version but some foreach constructs allow passing a variable number of arguments so you don't have that issue. Not so here.

jralls: >  @jralls - are you saying in C++ std::for_each might introduce a performance penalty in comparison to naked loop? It will be a bit of surprise to see that in optimized builds.

Read what you quoted more carefully: ***Unlike** the STL case*...

The STL algorithms can access private data and functions in STL containers that naked loops cannot, STL implementations are closely tied to their associated compilers and so can make use of compiler features both public and private that portable code cannot, and the STL implementors have time and motivation to focus on wringing the best possible performance out of the standard algorithms.

GLib's foreach functions have none of those advantages and some of them suffer from what Dr. Knuth calls "premature pessimization". Unfortunately profiling has shown that just about everything associated with GtkTreeModel iteration performs pretty poorly but there's no way around that if one wants to use a GtkTreeView.

jralls: @bhardwajs 
  >  @jralls - are you saying in C++ std::for_each might introduce a performance penalty in comparison to naked loop? It will be a bit of surprise to see that in optimized builds.

Read what you quoted more carefully: Unlike the STL case...

The STL algorithms can access private data and functions in STL containers that naked loops cannot, STL implementations are closely tied to their associated compilers and so can make use of compiler features both public and private that portable code cannot, and the STL implementors have time and motivation to focus on wringing the best possible performance out of the standard algorithms.

GLib's foreach functions have none of those advantages and some of them suffer from what Dr. Knuth calls "premature pessimization". Unfortunately profiling has shown that just about everything associated with GtkTreeModel iteration performs pretty poorly but there's no way around that if one wants to use a GtkTreeView.

jralls: @jeanlaroche Your antipathy to structs is misplaced; varargs are a much heavier-weight way of passing multiple parameters.

You don't "unpack" structs. The only extra effort to using a parameter in a struct is typing `foo->bar` instead of `bar`.

No struct needed in this case, though: the function would need only `list_item` or the hashtable I proposed.

jeanlaroche: > No struct needed in this case, though: the function would need only `list_item` or the hashtable I proposed.

I need model, list_element, display_threshold, fuzzy_amount, and gui

gnucash/import-export/import-main-matcher.c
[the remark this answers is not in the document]
jralls: The speed up would depend on how many accounts and how many transactions. With a single account there'd obviously be no speed up at all, in fact a slight slowdown because of the extra indirection; similarly if there were only a few transactions the speedup wouldn't be noticeable. On the other hand 5 accounts with 100 transactions each would get a 5N speedup where N is the number of items in the listmodel.

gnucash/import-export/import-main-matcher.c
jralls: Not `model`, it's passed to the `GtkTreeModelForeachFunc` by `gtk_tree_model_for_each` along with `path`, which you don't need and `iter`, which you do. I'll concede the other 3, I lost track of where the loop stopped.

jeanlaroche:
I simplified and improved the code so things run faster. BUT: there are still some major issues when importing large files into large account. For example, importing 6000+ transactions into an empty account can be very lengthy, then importing the same 6000 transactions into the same account can take absolutely forever (even though no transaction should be imported).
More details:
- Big ofx -> small accounts: things are reasonably fast, but the final import (i.e., adding the matched transactions into the account after the user clicks OK) takes a LONG time. Every transactions has to be committed, and this triggers a balance computation! (for each transaction) so this looks straight like O(N^2) cost! A simple solution would be to add an option to not compute the balance until the end. But I haven't tested that.
- Big ofx -> big account is insanely fast: the step of finding whether an imported transaction has been seen before becomes unmanageable (it's straight O(N^2)! So if N is 6000 we're talking about millions of operations/calls. A hash of transaction IDs would make that far faster, but I haven't tried that either.

The question is: is all this worth the trouble? Or is it over engineering? 6000 transactions is something like 20 years of data? (or something like that?). Who would import 20 years of data into an account that already has 20 years of data?

I'm not entirely sure what a reasonable test would be here. If you guys could offer some guidance that would be helpful: how big should I assume the account to be (in total number of transactions) and how big the import?
Then we can decide whether to do something about the slow import or not...

christopherlam:
I tend to think there's no such thing as enough speedups 😁
To prevent balance computations there's such things as xaccFooBeginEdit xaccFooCommitEdit

jeanlaroche:
> 
> 
> I tend to think there's no such thing as enough speedups 😁
> To prevent balance computations there's such things as xaccFooBeginEdit xaccFooCommitEdit

I didn't know about these! Really??? xaccFooBeginEdit  "Foo" ???? That's a riot!

christopherlam:
>     * Big ofx -> big account is insanely fast: the step of finding whether an imported transaction has been seen before becomes unmanageable (it's straight O(N^2)! So if N is 6000 we're talking about millions of operations/calls. A hash of transaction IDs would make that far faster, but I haven't tried that either.

You mean insanely slow? Without reading C how is it done? Matching FITID strings? Can't we make the safe assumption that in an OFX FITIDs are unique, and ditto for datafile's FITIDs? Then it becomes a simple application of hungarian (sorry, this again!) or any similar matching algorithm.

christopherlam:
> > I tend to think there's no such thing as enough speedups grin
> > To prevent balance computations there's such things as xaccFooBeginEdit xaccFooCommitEdit
> 
> I didn't know about these! Really??? xaccFooBeginEdit "Foo" ???? That's a riot! Oh, sorry, now I get it. xacc*BeginEdit... I see, now, you should start with a beginEdit, do all the stuff, then do a commitEdit. This is clearly NOT what's being done here. You get a begin/commit per transaction.

I guess, if 1 OFX = 1 account, you'd wrap the whole import with xaccAccountBeginEdit / xaccAccountCommitEdit. But beware the balance computations seem to handle such things as capital-gains computations.

jeanlaroche:
> You mean insanely slow? Without reading C how is it done? Matching FITID strings? Can't we make the safe assumption that in an OFX FITIDs are unique, and ditto for datafile's FITIDs? Then it becomes a simple application of hungarian (sorry, this again!) or any similar matching algorithm.

It's done brute force: you get an incoming ID, you look through all transactions you have for the target account, and see whether this ID exists. So if the account has N transactions, and you receive M ofx transactions you get N*M "comparisons" which themselves involve quite a few function calls... This type of stuff is much better done with a hash table which is completely optimal for that (instead of N^2 you'd get NLogN which is a LOT better).

jeanlaroche:
> 
> 

> I guess, if 1 OFX = 1 account, you'd wrap the whole import with xaccAccountBeginEdit / xaccAccountCommitEdit. But beware the balance computations seem to handle such things as capital-gains computations.

I'm not getting what you're hinting at...

christopherlam:
> It's done brute force: you get an incoming ID, you look through all transactions you have for the target account, and see whether this ID exists. So if the account has N transactions, and you receive M ofx transactions you get N*M "comparisons" which themselves involve quite a few function calls... This type of stuff is much better done with a hash table which is completely optimal for that (instead of N^2 you'd get NLogN which is a LOT better).

Agree that a hash-table approach would be perfect here. string-hash all the account's split's online_ids, and compare ofx's split's fitid, skip if exist.

> > I guess, if 1 OFX = 1 account, you'd wrap the whole import with xaccAccountBeginEdit / xaccAccountCommitEdit. But beware the balance computations seem to handle such things as capital-gains computations.
> 
> I'm not getting what you're hinting at...

I may have misread code.

christopherlam:
> > > I guess, if 1 OFX = 1 account, you'd wrap the whole import with xaccAccountBeginEdit / xaccAccountCommitEdit. But beware the balance computations seem to handle such things as capital-gains computations.
> > 
> > 
> > I'm not getting what you're hinting at...
> 
> I may have misread code.

Edit: this is probably irrelevant to `xaccAccountCommitEdit` --> https://github.com/Gnucash/gnucash/blob/b07d09328fd87a9cd1db9e100ace29328717d774/libgnucash/engine/Transaction.c#L1662

jralls:
Each QofInstance has an edit level that xaccFooBeginEdit increments and xaccFooCommit decrements; xaccFooCommit only does work when the edit level drops to zero. @christopherlam's idea was that  the account rebalancing could be deferred by calling xaccAccountBeginEdit before entering the import loop that creates the transactions. Even if that would work it one would have to do it on all of the matched accounts as well as the main target accounts or it would only defer the rebalances on the main target accounts. Unfortunately it won't even do that: xaccAccountRecomputeBalance is called from xaccSplitCommit and doesn't look at the account's edit level.

jralls:
I think 6000-item imports are pretty rare, but my primary book has a dozen accounts with more than 2000 splits each. Reimporting the same 6000 transactions into a book containing only those transactions is a bit of a degenerate case because there's obviously no way to filter the splits in the book to limit the range. A more realistic import will be 100 splits into an account having 3000 splits where the query can be optimized to return only a couple hundred splits. 

But that's a separate optimization from the one in this PR, so maybe we should defer it to another PR and get this one polished up and merged.

jeanlaroche:
Ah, thanks for the clarification. If there aren't side effects, I can 
make xaccAccountRecomputeBalance look at the edit level. I'll take a look.
I also found that transactions are created during the loading of the ofx 
file, then removed if they're already been imported. The removing 
process, once more, involves commits for each transactions! It can be 
super slow in my tests. So it's a good thing if I can avoid all this by 
incrementing the edit level...

J.

jralls:
A cursory check shows that adding an edit level check to `xaccAccountRecomputeBalance` should be OK, it does get indirectly called in `xaccAccountCommitEdit`. But that's a *cursory* check so you should test thoroughly.

jeanlaroche:
> `xaccAccountRecomputeBalance`

Wait, I see this in xaccAccountRecomputeBalance()
```
    if (qof_instance_get_editlevel(acc) > 0) return;
```
So it looks like it's already there.

jeanlaroche:
I add a good number of improvements...

To further increase the import speed, it's necessary to prevent any account commit to happen until the very end of the import (OK or Cancel), because account commits trigger very length balance computations. For this, I call xaccAccountBeginEdit on all the accounts involved in the import, keeping a list of them so BeginEdit is called only once. At the end of the import, commit is called on all the accounts in the list. Note that when the user selects a target account for an imported transaction, xaccAccountBeginEdit is called on the account, and it is added to the list.
27859cd66 Instead of saving the imported transaction into the treeview, which takes more time I now save them into a temporary list. A single query is done for all imported transactions and the resulting register splits are put into a hash table of lists with accounts as key, which will speed things up when multiple accounts are found. My tests of large imports on large accounts seems to ate that most of the time is spent verifying whether the imported transactions has already been imported, then computing the balance repeatedly for each imported transaction (!)  when the user clicks OK to add all the transactions to the account. So there's still room for improvement here!

christopherlam:
PS No further comments other than this is a pain point for my regular OFX imports, so, very thankful for your work.

gnucash/import-export/import-backend.c
```diff
@@ -1145,9 +1085,26 @@ static gint check_trans_online_id(Transaction *trans1, void *user_data)
...
+    
+    const gchar* online_id = gnc_import_get_trans_online_id (trans);
+    if (online_id)
+        g_hash_table_add (id_hash, (void*) online_id);
+    for (i=0,split = xaccTransGetSplit (trans, 0); split; split = xaccTransGetSplit (trans, ++i))
+    {
+        if (gnc_import_split_has_online_id (split))
+            g_hash_table_add(id_hash, (void*) gnc_import_get_split_online_id (split));
+    }
+    return 0;
```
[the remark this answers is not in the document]
jeanlaroche: Yes, I forgot to remove all my comments, sorry about that. They're crucial to me to allow navigating from file to file, in Xcode it's a royal pain the butt, there are no bookmarks so I use these comments as bookmarks. Lame.
About the for() construct, I was following advice given in the code itself, but I'll switch, the GList construct is a lot nicer.
The online_id are collated *per account* and it's possible that imported transactions correspond to different accounts (it's unusual, but GC does support ofx files that combine several accounts, I checked and it works well).

gnucash/import-export/import-backend.c
```diff
@@ -1156,13 +1113,20 @@ gboolean gnc_import_exists_online_id (Transaction *trans)
...
+    // Create a hash per account of a hash of all transactions IDs. Then the test below will be fast if
+    // we have many transactions to import.
+    dest_acct = xaccSplitGetAccount (source_split);
+    if (!g_hash_table_contains (id_hash, dest_acct))
+    {
+        GHashTable* new_hash = g_hash_table_new (g_str_hash, g_str_equal);
+        g_hash_table_insert (id_hash, dest_acct, new_hash);
+        printf ("CREATE HASH\n");
+        xaccAccountForEachTransaction (dest_acct, collect_trans_online_id, new_hash);
+        printf ("CREATE DONE\n");
```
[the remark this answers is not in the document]
jeanlaroche: I will definitely remove these, sorry about that.

gnucash/import-export/import-main-matcher.c
```diff
@@ -138,11 +145,46 @@ static void refresh_model_row (
...
+
+static void
+commit_all (GNCImportMainMatcher *info)
+{
+    // Commit the edits for all accounts for which we've called inc_account_edit.
+    for (;info->edited_accounts; info->edited_accounts=info->edited_accounts->next)
+    {
+        // JEAN Check that the counter is 1 so the edit does	 something.
+        int level = qof_instance_get_editlevel (info->edited_accounts->data);
+        g_assert(level == 1);
```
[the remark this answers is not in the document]
jeanlaroche: Yes, I was wondering what the best way to report this type of problem is, without causing a hard crash.

jeanlaroche:
> PS No further comments other than this is a pain point for my regular OFX imports, so, very thankful for your work.

Your ofx imports are that slow? I import transactions all the time, and I've never noticed anything slow. But to be honest I only import a few transactions at a time (cause I do it regularly). I'm curious how slow it is for you, and how many transactions you typically import.

gnucash/import-export/import-backend.c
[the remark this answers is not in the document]
jralls: Some Xcode users apparently [use breakpoints](https://www.jessesquires.com/blog/2020/01/21/xcode-tip-breakpoints-as-bookmarks/) for bookmarks.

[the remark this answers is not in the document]
jralls: Where was the advice about `xaccTransGetSplit`? It should make clear that it's only for getting a single split and not for iteration.

One of the interesting issues that comes up when importing for multiple accounts occurs when there's a transfer between two of the accounts in the import: There will be two entries in the import for that transfer transaction and of course they'll have different FITIDs.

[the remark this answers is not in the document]
jeanlaroche: > 
> 
> Some Xcode users apparently [use breakpoints](https://www.jessesquires.com/blog/2020/01/21/xcode-tip-breakpoints-as-bookmarks/) for bookmarks.

Yes! I've read about that, and tried to do it, but it's lame, you can't put any tag on the breakpoint, so it's not very useful... Xcode is really not top notch as far as I'm concerned but I can't possibly go command line, that's too painful.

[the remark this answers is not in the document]
jralls: Oh, and traversing with `xaccTransGetSplit` is O(N!), much worse than O(N^2).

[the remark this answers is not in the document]
jeanlaroche: > Where was the advice about `xaccTransGetSplit`? It should make clear that it's only for getting a single split and not for iteration.
> 

From Account.h
```
/** The xaccTransGetSplit() method returns a pointer to each of the
    splits in this transaction.
    @param trans The transaction
    @param i The split number.  Valid values for i are zero to
    (number_of__splits-1).  An invalid value of i will cause NULL to
    be returned.  A convenient way of cycling through all splits is
    to start at zero, and keep incrementing until a null value is returned. */
```
This advice should be removed...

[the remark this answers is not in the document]
jralls: Done.

gnucash/import-export/import-backend.h
```diff
@@ -64,36 +64,22 @@ typedef enum _action
  *
  * @param trans The transaction for which to check for an existing
  * online_id. */
-gboolean gnc_import_exists_online_id (Transaction *trans);
+gboolean gnc_import_exists_online_id (Transaction *trans, GHashTable* Foo);
```
jralls, line 67: Let's not use `Foo` and besides the [coding standards](https://wiki.gnucash.org/wiki/CodingStandard) say that capitalized and camel case identifiers should be type names. Variables should be lower-case and joined with underscores.

jeanlaroche: Yep, apologies, definitely slipped through the cracks.

gnucash/import-export/import-main-matcher.c
```diff
@@ -138,11 +145,45 @@ static void refresh_model_row (
...
+commit_all (GNCImportMainMatcher *info)
+{
+    // Commit the edits for all accounts for which we've called inc_account_edit.
+    for (;info->edited_accounts; info->edited_accounts=info->edited_accounts->next)
+    {
+        int level = qof_instance_get_editlevel (info->edited_accounts->data);
+        g_assert(level == 1);
+        xaccAccountCommitEdit (info->edited_accounts->data);
+    }
+    g_slist_free (info->edited_accounts);
```
jralls, line 165: That won't work, `info->edited_accounts` is `NULL` now. You need an iterator in the `for` loop:
```C
for(GSList* iter = info->edited_accounts; iter; iter=iter->next)
```
and use `iter` instead of `info->edited_accounts` inside the loop.

gnucash/import-export/import-main-matcher.c
```diff
@@ -138,11 +145,45 @@ static void refresh_model_row (
...
+}
+
+static void
+commit_all (GNCImportMainMatcher *info)
+{
+    // Commit the edits for all accounts for which we've called inc_account_edit.
+    for (;info->edited_accounts; info->edited_accounts=info->edited_accounts->next)
+    {
+        int level = qof_instance_get_editlevel (info->edited_accounts->data);
+        g_assert(level == 1);
```
jralls, line 162: Don't crash!  A SQL backend will already have written out the already-committed accounts so crashing will make a corrupted book. Instead log an error and then commit repeatedly until `editlevel == 0`.

gnucash/import-export/import-main-matcher.c
```diff
@@ -1435,38 +1499,151 @@ void gnc_gen_trans_list_add_trans_with_ref_id (GNCImportMainMatcher *gui, Transa
...
-
-        model = gtk_tree_view_get_model (gui->view);
-        gtk_tree_store_append (GTK_TREE_STORE(model), &iter, NULL);
+        
+        gtk_tree_store_append (GTK_TREE_STORE (model), &iter, NULL);
         refresh_model_row (gui, model, &iter, transaction_info);
     }
+    
+    qof_query_destroy (query);
+    g_slist_free (gui->temp_trans_list);
```
jralls, line 1641: This leaks the `trans_info`s. You need a `trans_info_delete` function and use `g_slist_free_full`.

gnucash/import-export/import-backend.c
```diff
@@ -1156,13 +1110,18 @@ gboolean gnc_import_exists_online_id (Transaction *trans)
     /* Look for an online_id in the first split */
     source_split = xaccTransGetSplit(trans, 0);
     g_assert(source_split);
```
jralls, line 1112: Because you've removed the atomicity of inserting splits you need to remove *all* of the asserts, even the ones you didn't write, and make sure that once `inc_account_edit` has been called `commit_all` will be too. Note that `g_return_if_fail`, `g_return_val_if_fail` call `g_log` with `G_LOG_LEVEL_ERROR` and that under some conditions that will call `g_assert` so you need to log the error with `PWARN` and return.

gnucash/import-export/import-main-matcher.c
jralls, line 162: See the note on asserting. Here the only safe thing to do if edit_level > 1 is to keep calling `xaccAccountCommitEdit` until it's 0.

gnucash/import-export/import-main-matcher.c
jeanlaroche: The freeing of trans_info is done in gnc_gen_trans_list_delete(), but your comment made me realize that the destruction of this list is better placed in gnc_gen_trans_list_delete().

gnucash/import-export/import-backend.c
jeanlaroche: I'm not getting this. What's a situation where g_assert(source_split) would trigger, that's caused by what I changed? I don't mind removing all the assert, but I'd like to understand why that's necessary!
Thanks for the thorough review BTW!

jralls: A transaction with a `NULL` split llist. Shouldn't happen, but that doesn't make it impossible... and even without having raised the account edit level it would hardly be a reason to crash the program, just return NULL and make sure that the calling code knows to not add it to the list and continue. Same for the `g_assert(trans)` at the top of the function.

Why it's necessary is that the SQL backend writes the db as part of processing the commit. You're blocking committing any changes to the accounts table but allowing commits to the transactions and splits tables. It's not protected with an encompassing SQL transaction that would cause the SQL server to roll everything back if GnuCash crashes, so a crash would cause a database with reference errors that either wouldn't load or wouldn't balance right and would be difficult to repair.

Even removing all of the asserts can't completely prevent crashes so I'm coming to think that the performance boost isn't worth the risk.

jeanlaroche: Ah yes, I understand now. Transaction changes are allowed but account changes are not, so in case of a crash this would be bad. Here's another option: keep all the commits as they are (i.e. remove all beginEdits) but add code to disable the computation of the balance until the end. In case of a crash, that wouldn't be a big deal at all, but it would still save a bunch of time. 
I might even be simpler to add that code than what I had to add to implement the beginEdit... 
What do you think?

jralls: That would surely be safer, though I'm skeptical that it would be simpler. What do you have in mind?

jralls:
The CI failure is:
```
/usr/sbin/ld: gnucash/import-export/test/CMakeFiles/test-import-backend.dir/__/import-backend.c.o: in function `TransactionGetTokens':

/gnucash/gnucash/import-export/import-backend.c:459: undefined reference to `xaccTransGetSplitList'

/usr/sbin/ld: gnucash/import-export/test/CMakeFiles/test-import-backend.dir/__/import-backend.c.o: in function `collect_trans_online_id':

/gnucash/gnucash/import-export/import-backend.c:1094: undefined reference to `xaccTransGetSplitList'

```

gnucash/import-export/import-backend.c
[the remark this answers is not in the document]
jralls: > Xcode is really not top notch as far as I'm concerned but I can't possibly go command line, that's too painful.

What's so painful about typing `ninja && ninja check`?
IIUC you can set up VSCode to do pretty much everything else. That might get you an environment close enough to VS for comfort.

gnucash/import-export/import-backend.c
jeanlaroche: > That would surely be safer, though I'm skeptical that it would be simpler. What do you have in mind?

How about adding a flag to the account_s structure
`gint defer_bal_computation ;`
that's 0 by default. In the code that I added which calls xaccAccBeingEdit, raise this flag instead (and add the account to the list as is currently done. Add code to not compute the balance when the flag is up. Then the final function that currently calls xaccAccCommitEdit, replace the call by set_defer_bal_computation(false) and in that function trigger the balance computation.
It's annoying to add a flag to the account structure just for that, but it's not a huge deal... It's possible that this flag/function could be reused in other parts of the code when batch modifications have to be made (i,e., move a bunch of transactions, etc).

jeanlaroche:
OK I fixed the issues that John flagged. In the second commit, I implemented the idea of adding a flag to the account's structure to defer the computation of the balance. Instead of using BeginEdit and CommitEdit, I now set the flag, and reset it at the end of the import operation. This way, all commits are still atomic, but we're still avoiding the constant balance recomputation.

jeanlaroche:
I'm not understanding the compilation error. For one thing, compilation works fine on my machine. For another, it's a linking error! Transaction.h is included so things compile fine, but I'm not understanding why linking is failing.
Before my latest change, the code was calling xaccTransGetSplit() which is defined in the same file (Transaction.c), which I changed to xaccTransGetSplitList() (the recommended way to cycle through splits). 
Can someone shed light on this for me?

jralls:
The failure is in a test program that uses a mock Transaction and xaccTransGetSplit is mocked but xaccTransGetSplitList isn't. See libgnucash/engine/mocks/gmock-Transaction.cpp.

jeanlaroche:
> The failure is in a test program that uses a mock Transaction and xaccTransGetSplit is mocked but xaccTransGetSplitList isn't. See libgnucash/engine/mocks/gmock-Transaction.cpp.

Well, that will teach me to build and run tests as well.

jeanlaroche:
All right so the test is fixed, and I think the issue of non atomic 
database writes is no longer valid.
Check whether the speed improvement is significant enough.
J.

jralls:
There's actually already a flag controlling balance recalculation, `balance_dirty`,  set to `TRUE` in `xaccSplitCommitEdit` just before it calls `xaccAccountRecomputeBalances`.  Changing that would have to do so through `xaccTransCommitEdit` so it would be a bit messier than the `defer` flag on accounts, so this approach is OK.

jralls:
> Check whether the speed improvement is significant enough.

@christopherlam Do you have a recurring slow import that you can test with?

jeanlaroche:
Or share with us, which would allow me to see where things are still 
slow, if they are. @john is there a way to get a build from a PR? I 
don't think so, right? Unless I set my repo to build...

gnucash/import-export/import-main-matcher.c
```diff
@@ -138,11 +145,42 @@ static void refresh_model_row (
...
+}
+
+static void
+update_all_balances (GNCImportMainMatcher *info)
+{
+    for (GSList* iter = info->edited_accounts; iter; iter=iter->next)
+    {
+        xaccAccountBeginEdit(iter->data);
+        gnc_account_set_defer_bal_computation(iter->data,0);
+        xaccAccountCommitEdit(iter->data);
```
jralls, line 162: `xaccAccountRecomputeBalance` just returns if the edit level > 0, so calling `xaccAccountBeginEdit` stops it from working.

gnucash/import-export/import-main-matcher.c
```diff
@@ -138,11 +145,42 @@ static void refresh_model_row (
...
+        info->edited_accounts = g_slist_prepend (info->edited_accounts, acc);
+    }
+}
+
 void gnc_gen_trans_list_delete (GNCImportMainMatcher *info)
 {
     GtkTreeModel *model;
     GtkTreeIter iter;
     GNCImportTransInfo *trans_info;
+    Split* first_split = NULL;
```
jralls, line 183: Unused.

jralls:
> @john is there a way to get a build from a PR? I don't think so, right? Unless I set my repo to build...

The only way to get a build from a PR is to pull it and build it yourself. TravisCI builds every PR but it discards everything after running the tests.

gnucash/import-export/import-main-matcher.c
```diff
@@ -72,6 +74,9 @@ struct _main_matcher_info
     GtkWidget         *reconcile_after_close;
     gboolean add_toggled;   // flag to indicate that add has been toggled to stop selection
     gint id;
+    GSList* temp_trans_list; // Temporary list of imported transactions
+    GHashTable* id_hash;     // Hash table, per account, of list of transaction IDs.
```
jralls, line 78: Could this be renamed `acct_hash` both here and as the arg to `gnc_import_exists_online_id` to make it clearer that `id_hash` used in `collect_trans_online` is `new_hash` from `gnc_import_exists_online_id`?

libgnucash/engine/Account.cpp
```diff
@@ -1859,6 +1859,31 @@ gnc_account_set_balance_dirty (Account *acc)
...
+    
+    g_return_if_fail (GNC_IS_ACCOUNT (acc));
+    
+    if (qof_instance_get_destroying (acc))
+        return;
+    
+    priv = GET_PRIVATE (acc);
+    priv->defer_bal_computation = defer;
+    if (!defer)
+        xaccAccountRecomputeBalance (acc);
```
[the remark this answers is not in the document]
jeanlaroche: Agreed

gnucash/import-export/import-main-matcher.c
jeanlaroche: How about acct_id_hash, since it's a hash of id, per account?

gnucash/import-export/import-main-matcher.c
```diff
@@ -1435,38 +1495,149 @@ void gnc_gen_trans_list_add_trans_with_ref_id (GNCImportMainMatcher *gui, Transa
...
+create_list_of_potential_matches (GNCImportMainMatcher *gui, GtkTreeModel* model, Query *query, gint match_date_hardlimit)
+{
+    GList* query_return = NULL;
+    Account *import_trans_account;
+    time64 import_trans_time, min_time=G_MAXINT64, max_time=0;
+    GList* all_accounts = NULL;
+    static const int secs_per_day = 86400;
+    GSList* imported_trans;
+    GList* potential_match;
+    GHashTable* lists_per_accounts = g_hash_table_new (g_direct_hash, g_direct_equal);
```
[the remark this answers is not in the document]
jeanlaroche: This does not work because when I do `g_hash_table_replace (lists_per_accounts, import_trans_account, per_account_list);` at line 1563 the delete function is called on the list. We don't want the list to be destroyed, to replace it with a new list, we only want to add an element to the existing list. I can't use append instead of prepend as that's also likely to change the pointer.
I did try, and got a crash.

jeanlaroche:
How do we make it possible for christopherlam to test the improvement though? It's very significant in my test accounts/ofx files, but I'd like to know whether it makes any different in his case...

christopherlam:
> How do we make it possible for christopherlam to test the improvement though? It's very significant in my test accounts/ofx files, but I'd like to know whether it makes any different in his case...

I usually import OFX on my Windows box. Still, testing on Linux importing 6 months transactions results in before=5s and after=1s so I'm rather happy about this progress (and cleanup).

gnucash/import-export/import-main-matcher.c
jralls: Sure. The main thing is that it's a different name.

gnucash/import-export/import-main-matcher.c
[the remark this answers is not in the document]
jralls: Oh, right. You could change line 1563 to
```C
   g_hash_table_steal (lists_per_accounts, import_trans_account);
   g_hash_table_add (lists_per_accounts, import_trans_account, per_account_list);
```
But that's pretty opaque.  You could add a level of indirection with a struct containing only the GSList* and provide a ctor and dtor so that the hash table's value would be stable, but I think it not worth the effort.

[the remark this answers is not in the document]
jeanlaroche: Right. It's annoying that you can't append to a list without having the original pointer possibly invalidated, I don't understand what that's the case (I can see why for prepend, but not for append). If that were the case, it would make the process a lot simpler, you wouldn't have to change anything to the hash table, just keep adding to the list.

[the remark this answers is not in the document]
jralls: Maybe this, though, replacing 1561-1563:
```C
// Steal the GSList from the lists_per_accounts hash so that the value destructor isn't called on it.
g_hash_table_steal_extended (lists_per_accounts, import_trans_account, NULL, &per_account_list);
per_account_list = g_slist_prepend (per_account_list, potential_match->data);
g_hash_table_add (lists_per_accounts, import_trans_account, per_account_list);
```
I checked the code and `g_hash_table_steal_extended` does `NULL` the output ptrs if the key isn't found.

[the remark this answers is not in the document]
jeanlaroche: Ah, nice, yes, that way you get both the initialization and the update in one shot. I'll do that.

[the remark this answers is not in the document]
jralls: You wouldn't want to append anyway. It has to traverse the list to find the end so it gets really slow as the list grows. That's why the docs say that in general you should reverse, prepend, and reverse if you need to append in a loop and care about the order. That said the [returned pointer *is* stable unless the incoming is NULL](https://gitlab.gnome.org/GNOME/glib/-/blob/master/glib/gslist.c#L227).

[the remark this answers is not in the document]
jeanlaroche: Ah! The doc says that the pointer might change, I guess they mean if you pass a NULL! But yes, append is quite slow for a long list.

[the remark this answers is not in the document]
jralls: I think it's more likely that whoever wrote the doc comment copied the one from `g_list_prepend` and was too lazy to edit it to be more precise. Either that or they envisioned eventually creating a pre-allocated memory block to keep the list elements contiguous, which would vastly speed up traversal by improving cache usage but would also require reallocating when the list got too big for the existing block. That would require a stronger type guarantee that's difficult to pull off when you're in the habit of passing `void*` to everything so no doubt the vision died young.

gnucash/import-export/import-main-matcher.c
```diff
@@ -1435,38 +1496,141 @@ void gnc_gen_trans_list_add_trans_with_ref_id (GNCImportMainMatcher *gui, Transa
...
+    // Now put all potential matches into a hash table based on their account.
+    for (potential_match=query_return; potential_match!=NULL; potential_match=potential_match->next)
+    {
+        Account* split_account;
+        GSList* per_account_list;
+        if (gnc_import_split_has_online_id (potential_match->data))
+            continue;
+        split_account = xaccSplitGetAccount (potential_match->data);
+        // This will be NULL the first time around, which is fine for a GSList
+        g_hash_table_steal_extended (lists_per_accounts, import_trans_account, NULL, (gpointer)&per_account_list);
```
christopherlam, line 1561: This fails on ubuntu 18.04 -- it has glib 2.56.4 < 2.58 doesn't have `g_hash_table_steal_extended` -- how annoying!

gjanssens:
@jeanlaroche A sidenote relating to all your PR's, not just this one: I noted from some recent merges your commits are signed off by "you@example.com". I'd be very impressed if you were the owner of that domain, however I rather suspect you didn't configure an e-mail for git yet on your local PC. As it is important for attribution under the GPL, can you please replace this for future work by a real e-mail address ?

gnucash/import-export/import-main-matcher.c
jralls: Well, fooey. So it has to be either the original code or a `g_hash_table_lookup();g_hash_table_steal()` two-step. If the original code then use `g_hash_table_foreach_remove` to traverse the hash table only once for destruction.

jeanlaroche:
> 
> 
> @jeanlaroche A sidenote relating to all your PR's, not just this one: I noted from some recent merges your commits are signed off by "[you@example.com](mailto:you@example.com)". I'd be very impressed if you were the owner of that domain, however I rather suspect you didn't configure an e-mail for git yet on your local PC. As it is important for attribution under the GPL, can you please replace this for future work by a real e-mail address ?

My bad, I'll take a look at my gitconfig.

gnucash/import-export/import-main-matcher.c
```diff
@@ -147,36 +147,27 @@ static void refresh_model_row (
...
+defer_bal_computation (GNCImportMainMatcher *info, Account* acc)
 {
-    // Call BeginEdit on accounts that haven't been edited yet. This will speeds things up
-    // a lot by preventing commits until we're all done.
-    if (qof_instance_get_editlevel (acc) == 0)
+    if (!gnc_account_get_defer_bal_computation(acc))
     {
-        xaccAccountBeginEdit (acc);
+        gnc_account_set_defer_bal_computation (acc, TRUE);
+        xaccAccountRecomputeBalance (acc);
```
jralls, line 170: Doesn't this belong in `update all balances`?

jeanlaroche: Well, first of `xaccAccountRecomputeBalance (acc);` should not be there, for one thing it does nothing, and second it's irrelevant here, since we're trying to avoid any balance recomputation!
But `defer_bal_computation` is called from multiple places to set the flag on the accounts, and add them to our list so we can reset the flags at the end.
Or maybe I don't understand your comment.

jralls: I think you just put the `xaccAccountRecomputeBalance (acc)` call in the wrong place. Doesn't it belong in `update_all_balances` after the call to `gnc_account_set_defer_bal_computation (acc, FALSE)` call?

christopherlam:
> My bad, I'll take a look at my gitconfig. EDIT: it had the wrong email, but I can't really change it for this PR. I'll do it next time I branch out. Sorry about that.

https://confluence.atlassian.com/bitbucketserverkb/how-do-you-make-changes-on-a-specific-commit-779171729.html will probably modify your commit details correctly.

gnucash/import-export/import-backend.c
```diff
@@ -1145,9 +1077,27 @@ static gint check_trans_online_id(Transaction *trans1, void *user_data)
     }
 }
 
+static gint collect_trans_online_id(Transaction *trans, void *user_data)
+{
+    Split *split;
+    GHashTable *id_hash = user_data;
+    int i=0;
```
christopherlam, line 1084: i=0 is now unused.

gnucash/import-export/import-main-matcher.c
jeanlaroche: You're absolutely right. Dang it, I need to be more careful. Thanks.

jralls:
Finally got it!. Thanks.

gnucash/import-export/import-backend.c
```diff
@@ -1145,9 +1077,27 @@ static gint check_trans_online_id(Transaction *trans1, void *user_data)
...
+        g_hash_table_add (id_hash, (void*) online_id);
+
+    for (GList *splits = xaccTransGetSplitList (trans); splits; splits = splits->next)
+    {
+        if (gnc_import_split_has_online_id (splits->data))
+            g_hash_table_add(id_hash, (void*) gnc_import_get_split_online_id (splits->data));
+    }
+    return 0;
+}
+
```
christopherlam, line 1097: I am aware this function `collect_trans_online_id` was moved away, but this is the first occurrence of the `g_hash_table_add` usage. @jeanlaroche would you be able to review the very recent commit 579ba443172fde921fd6526e37a0a30be7d2da5e which has changed `g_hash_table_add (hash, id)` to `g_hash_table_insert (hash, id, 1)` -- the reason is that `id` is a newly allocated string and and needs to be freed after use, and if we use `id` as both key and value in the hash table, then I'm not sure whether to free in the key or value.

jeanlaroche: First, I don't think there's any issue with using g_hash_table_insert (hash, id, 1), it might be slightly less efficient memorywise but I can't imagine that being a problem.
Second, if you use g_hash_table_add , I imagine you'd just free the key when the hash table is destroyed.

christopherlam: Thanks for clarification. Agree it probably doesn't matter that much, and now there will be fewer leaks.

jralls: From a leak perspective `g_hash_table_add` has a wrinkle: It's a wrapper of `g_hash_table_replace` that, like its name suggests, discards the existing key and value and inserts the new one. To prevent leaks it's therefore important to set the `key_destroy_func` or the `value_destroy_func` when creating the hash table so that one or the other gets cleaned up; there's no other way to get the old key/value to free it.

PR #723, opened 2020-05-18
POC: run report from CLI

christopherlam:
gnucash --run-report="Transaction Report" --output-file=~/rpt.html
--run-report specify either reportname or reportguid
--output-file default is gnucash.html

Session doesn't always load file.
I've probably loaded too many modules.

Next:
- [x] also allow --export-type=csv to trigger export code :)
- [x] check validity of report arguments *before* loading datafile.

### DIFF
--- a/gnucash/gnucash-bin.c
+++ b/gnucash/gnucash-bin.c
@@ -107,6 +107,9 @@ static char        *log_to_filename  = NULL;
 static int          nofile           = 0;
 static const gchar *gsettings_prefix = NULL;
 static const char  *add_quotes_file  = NULL;
+static const char  *run_report       = NULL;
+static const char  *export_type      = NULL;
+static const char  *output_file      = NULL;
 static char        *namespace_regexp = NULL;
 static const char  *file_to_load     = NULL;
 static gchar      **args_remaining   = NULL;
@@ -158,6 +161,18 @@ static GOptionEntry options[] =
         /* Translators: Argument description for autohelp; see
            http://developer.gnome.org/doc/API/2.0/glib/glib-Commandline-option-parser.html */
         N_("FILE")
+    },
+    {
+        "run-report", '\0', 0, G_OPTION_ARG_STRING, &run_report,
+        N_("Runs a report"), N_("FILE")
+    },
+    {
+        "export-type", '\0', 0, G_OPTION_ARG_STRING, &export_type,
+        N_("Specify export type"), NULL
+    },
+    {
+        "output-file", '\0', 0, G_OPTION_ARG_STRING, &output_file,
+        N_("Output file for report default gnucash.html"), N_("FILE")
     },
     {
         "namespace", '\0', 0, G_OPTION_ARG_STRING, &namespace_regexp,
@@ -595,6 +610,82 @@ get_file_to_load()
         return gnc_history_get_last();
 }
 
+static void
+report_session_percentage (const char *message, double percent)
+{
+    static double previous = 0.0;
+    if ((percent - previous) < 5.0)
+        return;
+    fprintf (stderr, "\r%3.0f%% complete...", percent);
+    previous = percent;
+    return;
+}
+
+static void
+inner_main_run_report (void *closure, int argc, char **argv)
+{
+    QofSession *session = NULL;
+    SCM cmdline, report, type, file;
+    const gchar *datafile;
+
+    scm_c_eval_string("(debug-set! stack 200000)");
+    scm_c_use_module ("gnucash utilities");
+    scm_c_use_module ("gnucash app-utils");
+    scm_c_use_module ("gnucash reports");
+
+    gnc_report_init ();
+    load_system_config();
+    load_user_config();
+    gnc_prefs_init ();
+    qof_event_suspend ();
+    datafile = get_file_to_load();
+
+    cmdline = scm_c_eval_string ("gnc:cmdline-run-report");
+    report = scm_from_utf8_string (run_report);
+    type = export_type ? scm_from_utf8_string (export_type) : SCM_BOOL_F;
+    file = output_file ? scm_from_utf8_string (output_file) : SCM_BOOL_F;
+
+    /* dry-run? is #t: check options to check validity of options */
+    if (scm_is_false (scm_call_4 (cmdline, report, type, file, SCM_BOOL_T)))
+        goto fail;
+
+    fprintf (stderr, "Loading datafile %s...\n", datafile);
+
+    session = gnc_get_current_session ();
+    if (!session) goto fail;
+
+    qof_session_begin (session, datafile, TRUE, FALSE, FALSE);
+    if (qof_session_get_error (session) != ERR_BACKEND_NO_ERR) goto fail;
+
+    qof_session_load (session, report_session_percentage);
+    if (qof_session_get_error (session) != ERR_BACKEND_NO_ERR) goto fail;
+
+    fprintf (stderr, "\n");
+
+    /* dry-run? is #f: run the report */
+    scm_call_4 (cmdline, report, type, file, SCM_BOOL_F);
+
+    qof_session_end (session);
+    if (qof_session_get_error (session) != ERR_BACKEND_NO_ERR) goto fail;
+
+    qof_session_destroy (session);
+
+    qof_event_resume ();
+    gnc_shutdown (0);
+    return;
+fail:
+    if (session)
+    {
+        if (qof_session_get_error (session) != ERR_BACKEND_NO_ERR)
+            g_warning ("Session Error: %d %s",
+                       qof_session_get_error (session),
+                       qof_session_get_error_message (session));
+        qof_session_destroy (session);
+    }
+    qof_event_resume ();
+    gnc_shutdown (1);
+}
+
 extern SCM scm_init_sw_gnome_module(void);
 
 static void
@@ -914,6 +1005,14 @@ main(int argc, char ** argv)
     {
         scm_boot_guile(argc, argv, inner_main_add_price_quotes, 0);
         exit(0);  /* never reached */
+    }
+
+
+    /* If asked via a command line parameter, run report only */
+    if (run_report)
+    {
+        scm_boot_guile (argc, argv, inner_main_run_report, 0);
+        exit(0);
     }
 
     /* We need to initialize gtk before looking up all modules */
--- a/gnucash/report/report-core.scm
+++ b/gnucash/report/report-core.scm
@@ -801,13 +801,12 @@ not found.")))
 
 (define-public (gnc:cmdline-run-report report export-type output-file dry-run?)
   (let ((template (or (gnc:find-report-template report)
-                      (let ((retval #f))
-                        (hash-for-each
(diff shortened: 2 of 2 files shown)


gnucash/gnucash-bin.c
```diff
@@ -159,6 +161,14 @@ static GOptionEntry options[] =
...
         N_("FILE")
     },
+    {
+        "run-report", '\0', 0, G_OPTION_ARG_STRING, &run_report,
+        N_("Runs a report"), N_("FILE")
+    },
+    {
+        "output-file", '\0', 0, G_OPTION_ARG_STRING, &report_output,
+        N_("Output file for report default gnucash.html"), N_("FILE")
+    },
```
gjanssens, line 171: Why not printing to stdout by default ? That allows for more scripting flexibility.

gnucash/gnucash-bin.c
```diff
@@ -600,6 +610,72 @@ get_file_to_load()
...
+    scm_c_use_module("gnucash report-menus");
+
+    /* scm_set_current_module (scm_c_resolve_module ("gnucash report")); */
+    /* scm_set_current_module (scm_c_resolve_module ("gnucash reports")); */
+    gnucash_register_add_cell_types ();
+    gnc_report_init ();
+    load_gnucash_plugins();
+    load_gnucash_modules();
+    load_system_config();
+    load_user_config();
```
gjanssens, line 633: I think this does way too much. Only load what you need.
I doubt you need to load_gnucash_modules. The only modules left are the optional ones (ofx, python and aqbanking). You don't need those to generate a report. In the same way you don't need any of the plugins loaded with load_gnucash_plugins. They too are related to importers and log-replay. Generating a report doesn't import and won't change transactions.

gnucash/gnucash-bin.c
```diff
@@ -600,6 +610,72 @@ get_file_to_load()
...
+    qof_session_begin (session, get_file_to_load(), FALSE, FALSE, FALSE);
+    if (qof_session_get_error (session) != ERR_BACKEND_NO_ERR) goto fail;
+
+    qof_session_load (session, NULL);
+    if (qof_session_get_error (session) != ERR_BACKEND_NO_ERR) goto fail;
+
+    if (report_output)
+        scm_report_output = scm_from_utf8_string (report_output);
+    else
+        scm_report_output = scm_from_utf8_string ("gnucash.html");
```
gjanssens, line 653: If you would consider outputting to stdout it may be easier to keep output file handling in C(++) and reduce guile's responsibility to generate the html to a string.

christopherlam:
1. https://code.gnucash.org/logs/2020/05/17.html#T12:45:14
2. some qofsession bug: qof_session_get_error_message doesn't quite work well. If file is open via `qof_session_begin (session, datafile, FALSE, FALSE, FALSE);` instead of `qof_session_begin (session, datafile, TRUE, FALSE, FALSE);` and the lock file exists, the qof_session_get_error returns 7 yet message is null. Hence the following double error dump. Looks like `push_error` isn't being used correctly in qofsession.cpp.
```c
            g_warning ("Session Error: %d %s",
                       qof_session_get_error (session),
                       qof_session_get_error_message (session));
```
3. if I remove `qof_session_save (session, NULL);` then the QofBackend complains:
 08:39:09  WARN <gnc.backend> [GncXmlBackend::session_end()] Error on g_unlink(/datafile.gnucash.LCK): 2: No such file or directory

christopherlam:
Currently this branch allows running of standard reports and saved reports with preset options. However we need to discuss on _whether_ and _how_ to allow setting report options from CLI. The difficulty will be complex objects such as dates or documents: `--set-option="General/Invoice Number/000030" or `--set-option="General/Customer/John Doe"`  -- none of the following is ideal...

1. allow arbitrary guile scripts with access to full API `(local-eval str (the-environment))`:- too unsafe! 
2. use json-like external file `--options=options.json`

jralls:
Users could always write an options paragraph into their `saved-reports-2.8`. Since we still haven't resolved how we're going to store saved report settings in 5.x I think it's premature to discuss how or whether to create a report options file for command-line usage.

That said, IMO ini/yaml would be easier for users to edit than a json one.

christopherlam:
Although this branch *could* be ready, I'm not yet 100% sure of its behaviour when run in parallel with an open datafile. I *think* it will delete the existing .LCK file. I do not know how to open a read-only qofsession. There is a `book->read_only` flag but I am not convinced it will leave .LCK file alone.

jralls:
I think @gjanssens wants to wrap this up with his more general CLI change. He'll be back tomorrow.

crossan007:
Would also be nice to be able to "save" a MySQL database as a `.gnucash` file from the CLI

jralls:
@crossan007 Please keep enhancement requests on the [bug tracker](https://bugs.gnucash.org/enter_bug.cgi).

christopherlam:
The idea to use [.ini files](https://devblogs.microsoft.com/oldnewthing/20071126-00/?p=24383) or [.yaml files](https://github.com/esilkensen/yaml/blob/master/yaml/parser.rkt), are both basically both a type of [DSL](https://en.wikipedia.org/wiki/Domain-specific_language). An approach is still to use guile, augmented with a very limited number of report-API functions, and neutralized to deny access to gnucash API. I may not have time for this.

```scheme
(let ((report (get-report "Balance Sheet")))
  (set-option! report "Accounts" "Accounts" (get-all-accounts))
  (set-option! report "General" "Date" (get-relative-date 'today))
  (run-report report #:file-name "bs.html"))

(let ((report (get-report "Transaction Report")))
  (set-option! report "Accounts" "Accounts" (get-account "Asset:Bank"))
  (set-option! report "General" "End Date" (get-absolute-date 31 12 (get-year)))
  (run-report report #:file-name "tr.html"))
```

derekatkins:
> The idea to use [.ini files](https://devblogs.microsoft.com/oldnewthing/20071126-00/?p=24383) or [.yaml files](https://github.com/esilkensen/yaml/blob/master/yaml/parser.rkt), are both basically both a type of [DSL](https://en.wikipedia.org/wiki/Domain-specific_language). An approach is still to use guile, augmented with a very limited number of report-API functions, and neutralized to deny access to gnucash API. I may not have time for this.

GnuCash already has support for INI files; the .chk files are INI format.  So we could (should?) leverage that code.

jralls:
> The idea to use .ini files or .yaml files, are both basically both a type of DSL.

B.S. `ini`/`yaml` (the latter being a generalized flavor of the former) is a categorized key-value store similar to but simpler and more human-friendly than `json` or `xml`.

christopherlam:
Again I may not have time nor energy for this line of work. INI/YAML/anything else will still be awkward. Not easy to figure out how to decode the following. Unicode, delimiters, multichoice, complex objects etc. I won't be bothered myself... quite happy to do one report at a time as this PR currently stands.

```
[Accounts]
Accounts=Asset:Bank, Asset:Savings, Asset:Foreign:Niños

[General]
Report Title=你好
Customer=My Client
Invoice=0000001
Start Date=today
End Date=2021-12-31
Period=monthly
```

derekatkins:
> Again I may not have time nor energy for this line of work. INI/YAML/anything else will still be awkward. Not easy to figure out how to decode the following. Unicode, delimiters, multichoice, complex objects etc. I won't be bothered myself... quite happy to do one report at a time as this PR currently stands.
> 
> ```
> [Accounts]
> Accounts=Asset:Bank, Asset:Savings, Asset:Foreign:Niños
> 
> [General]
> Report Title=你好
> Customer=My Client
> Invoice=0000001
> Start Date=today
> End Date=2021-12-31
> Period=monthly
> ```

Chris, you don't even need to think about most of that.  C.f. libgnucash/core-utils/gnc-gkeyfile-utils.[ch] which loads (and saves, if you wish) a GKeyFile data structure.  So all the *file parsing* is done for you.  You just have to map the appropriate section/key/value pairs to the proper report options.

jralls:
Chris, remember we're talking about after `c++options` lands and we decide how we want to store saved reports going forward. The implementation will be in C++ and nobody expects you to do it. For now and for 4.x saved report configs are scheme fragments in `saved-reports-2.8`.

christopherlam:
Ok I'll leave it as it is. What I'd personally miss is the ability to select the _latest_ or any arbitrary invoice for printing and rendering to PDF but can live without it for now.

jralls:
What has that to do with saved report configurations? Or rather if the user has saved their latest invoice as a saved report configuration why would the current arrangement not be able to print it from the command line?

christopherlam:
As [above](https://github.com/Gnucash/gnucash/pull/723#issuecomment-631887183) I'd ideally like to use `--set-option="General/Invoice Number/000030"` or `--set-option="General/Customer/John Doe"` but (1) to translate 000030 into a document is not easy and requires custom string->object for each option type, (2) setting many options from commandline becomes unwieldly and suggests using an options datafile instead. This is completely separate from `saved-reports-2.8` and would be mainly for ad-hoc reports generated from CLI.

christopherlam:
A couple last minutes issues/decisions
1. do we allow *all* reports to be run from CLI or saved-reports only? It's currenlty the former, I think a change to the latter would be better to reinforce the view that options cannot be set from cli.
2. as above the behaviour of this code's session_load when the book is already open (and .LCK is present) is currently untested -- I'm not familiar enough with C++ to decode qofsessions.cpp behaviour for `ignore_lock`. Maybe this needs to be clarified in the release notes (i.e. don't open sessions concurrently) or someone else will need to verify the .LCK file is not prematurely deleted by this code.

derekatkins:
> A couple last minutes issues/decisions
> 
>     1. do we allow _all_ reports to be run from CLI or saved-reports only? It's currently the former, I think a change to the latter would be better to reinforce the view that options cannot be set from cli.
> 
>     2. as above the behaviour of this code's session_load when the book is already open (and .LCK is present) is currently untested -- I'm not familiar enough with C++ to decode qofsessions.cpp behaviour for `ignore_lock`. Maybe this needs to be clarified in the release notes (i.e. don't open sessions concurrently) or someone else will need to verify the .LCK file is not prematurely deleted by this code.

I don't have an answer for question 2.
As for question 1, I think it would be most useful if it can open any report.

christopherlam:
modified to dump in-menu?=#t, but accepts any reportname or report-guid.

jralls:
No. Why would there be?

jguddas:
>  [x] also allow --export-type=csv to trigger export code :)

I was wondering what this entails and if it means I can trigger `Export Transactions to CSV` from the CLI.

I now understand that this not the case.

christopherlam:
> > [x] also allow --export-type=csv to trigger export code :)
> 
> I was wondering what this entails and if it means I can trigger `Export Transactions to CSV` from the CLI.
> 
> I now understand that this not the case.

Why? Not all reports have CSV export.

christopherlam:
This is not a report export option. The --export-type CSV refers to reports exporing CSV:
`./gnucash-cli -R run --name "Income and GST Statement" --export-type CSV ~/datafile.gnucash`

Transactions cannot be exported to CSV from CLI.

PR #724, opened 2020-05-19
Bug 797759 - Some transactions are not highlighted in the matching wi…

jeanlaroche:
…ndow

This is a change from the previous code which used gtk_tree_selection_set_select_function() to prevent some transactions from being selectable.
In this solution, I removed the select_function, instead, during the cursor moved callback, I look at the list of selected transactions and de-select those that should be de-selected, but only if we have more than one transaction selected.
This allows the user to arrow up/down and see the transaction under the cursor be highlighted, but when rubber-banding selections, only the right ones are added to the selection.

I removed the gboolean add_toggled flag, because there no longer seems to be any need for it, but perhaps I'm wrong on this one.


jralls:
Thanks.

PR #725, opened 2020-05-22
Gnc item edit changes

Bob-IT:
@jralls I have added your changes to this PR along with a couple of other changes I had noticed.
Note, I also moved the gnc_item_list_select and the unblock_list_signals to outside of the if block.


jralls:
> Note, I also moved the gnc_item_list_select and the unblock_list_signals to outside of the if block.

I don't see that change...

Bob-IT:
@jralls  I am confused by your comment, all lines from your patch are present it is just I moved  gnc_item_list_select and the unblock_list_signals out of the if block to following, that way any items in the action list are selected.
I have also added getting the row height from the tree view cell-renderer height.

jralls:
Ah, now I see it. OK.

PR #726, opened 2020-05-22
Speed up lot->owner by caching it

christopherlam:
Offshoot from #638 caching lot's owner. First attempt caches the guid only; second attempt caches the whole gncOwner. In the second approach, a gncOwner object copying is done for every cache hit.


christopherlam:
Closing, found another way to speed up new-aging.scm

PR #657, opened 2020-03-01
Python gettext localize pot

c-holtermann:
This is the pull request with translations for the python strings. It needs to be applied on top of PR #655. Draft in #506.

PR #727, opened 2020-05-29
Python gettext localize pot

c-holtermann:
Unintentional closing of same PR #657.


po/de.po
```diff
@@ -22251,9 +22251,8 @@ msgid "Have a nice day!"
 msgstr "Einen schönen Tag noch."
 
 #: gnucash/python/init.py:118
-#, fuzzy, python-format
```
fellen: Remove `, fuzzy`, but not `, python-format`. Otherwise it would really display "Willkommen zur GnuCash %s Shell" instead of the parameter.

Fixed in my following commit.

Something else: use in the future more descriptive commit messages, e.g. in this case by using the prefix `L12N:de: `

PR #730, opened 2020-06-02
Fix typo: Exsiting -> Existing

yurchor:
Thanks in advance for reviewing.


fellen:
Commit 1467f33
I took the freedom to prefix the commit message by "I18N: ". That shoukd help us translators to see for us relevant commits.

Thanks!

PR #731, opened 2020-06-02
make (parts of) gnucash logging accessible from python modules

c-holtermann:
Small change, big effect. This makes (parts of) logging accessible to python bindings.

With this it's possible to go

```
gnucash.gnucash_core_c.qof_log_set_level("", gnucash.gnucash_core_c.QOF_LOG_DEBUG)
gnucash.gnucash_core_c.qof_log_set_level("gnc", gnucash.gnucash_core_c.QOF_LOG_DEBUG)
gnucash.gnucash_core_c.qof_log_set_level("qof", gnucash.gnucash_core_c.QOF_LOG_DEBUG)
```

and see all the debug information when calling the c/c++ code. Very helpful for coding/debugging python.

It could be worth a try
* enable code using python modules to read the logging config file as done in gnucash bin gnc_log_init()
  * it would be nice to be able to just call gnc_log_init() but there's no header file for that as that is in gnucash-bin.c - could that be put into a separate file to make it accessible from python? Otherwise mimic it
* use the gnucash logging functions to write out python info

### DIFF
--- a/bindings/python/gnucash_core.i
+++ b/bindings/python/gnucash_core.i
@@ -215,6 +215,8 @@ static const GncGUID * gncEntryGetGUID(GncEntry *x);
 %include <cap-gains.h>
 %include <Scrub3.h>
 
+%include <qoflog.h>
+
 %init %{
 gnc_environment_setup();
 qof_log_init();


jralls:
This shouldn't be needed, qoflog.h is already `%include`d in bindings/engine.i.

c-holtermann:
Thanks for your reply. Interesting, I'll have a look at that.

jralls:
@c-holtermann Did you ever test logging without this PR?

c-holtermann:
@jralls - I tried without smime time ago. I was quite busy the last time so I could not work on this. I will try it out as soon as possibe to finish this pull request. Sorry for the massive delay!

jralls:
The right answer is that yes, engine.i has `%include <qof-log.h>`, but the python bindings don't actually use engine.i. gnucash_core.i should at some point get melded into engine.i to remove the  redundancy, but for now I've merged this.

Thanks!

PR #732, opened 2020-06-02
Python c++ QofSessionImpl (draft)

c-holtermann:
This is an attempt to wrap the c++ parts of gnucash, starting with QofSessionImpl as suggested by @jralls ([back in 2018](https://lists.gnucash.org/pipermail/gnucash-devel/2018-August/042393.html)).
Please have a look at it. It's development in progress. But the example python file and first unit tests work. I'll add some more tests before finishing this PR, clean up the commits and change it according to the discussion which I'm looking forward to.

**Goal**: First steps of making c++ parts of gnucash accessible for python bindings. From my understanding the code base will be moved from c to c++ so the python bindings will need to follow this transition.
**Implementation**: Use QofSessionImpl as a a first c++ structure to wrap. Implement a second python module gnucash_core_cc  to house the c++ methods besides gnucash_core_c which holds the python wrappers of gnucash c files as they are now. Gnucash_core_cc has the results of a swig call with command line argument '-c++'. Cmake files control those two separate swig calls, one for each language that create the respective python modules. They are submodules of the python gnucash module that contains the high level wrappers. High level wrapping for c++ will be a bit easier as the objects don't need to be patched together from the c function calls by the procedures provided by function_class.py but get provided by swig directly.
**History**: I started on this in 2018 but due to involvement in other issues it rested until spring 2020 when I found the time to pick it up again. Due to time passing some misfits might have come up.
**Future**: 
* provide the small amount of high level wrapping to bring the QofSessionImpl object from gnucash_core_cc to gnucash
   * adjust example and unittest to that
* have a look into which other c++ parts make sense to wrap
   *  I did some work on the c++ GncNumeric class
* Gnucash_core_cc may be joined into gnucash_core_c if that is possible and makes more sense.
* Finally move everything from gnucash_core_c to gnucash_core_cc as gnucash will be completely written in c++.
* it might make sense to talk about the content of python modules gnucash, gnucash_core_c and gnucash_core_cc. Methods in gnucash could be called stable, those in gnucash_core_c and gnucash_core_cc private and subject to change for example.

### DIFF
--- a/bindings/python/CMakeLists.txt
+++ b/bindings/python/CMakeLists.txt
@@ -3,7 +3,9 @@ add_subdirectory(tests)
 
 set(PYEXEC_FILES  __init__.py function_class.py gnucash_business.py gnucash_core.py app_utils.py)
 
-set(SWIG_FILES ${CMAKE_CURRENT_SOURCE_DIR}/gnucash_core.i ${CMAKE_CURRENT_SOURCE_DIR}/time64.i)
+set(SWIG_FILES ${CMAKE_CURRENT_SOURCE_DIR}/gnucash_core.i ${CMAKE_CURRENT_SOURCE_DIR}/time64.i )
+set(SWIG_FILES_CC ${CMAKE_CURRENT_SOURCE_DIR}/gnucash_core_cc.i ${CMAKE_CURRENT_SOURCE_DIR}/time64.i )
+
 set(GNUCASH_CORE_C_INCLUDES
     ${CONFIG_H}
     ${CMAKE_SOURCE_DIR}/libgnucash/engine/qofsession.h
@@ -35,6 +37,11 @@ set(GNUCASH_CORE_C_INCLUDES
     ${CMAKE_SOURCE_DIR}/libgnucash/app-utils/gnc-prefs-utils.h
 )
 
+set(GNUCASH_CORE_CC_INCLUDES
+    ${GNUCASH_CORE_C_INCLUDES}
+    ${CMAKE_SOURCE_DIR}/libgnucash/engine/gnc-numeric.h
+)
+
 gnc_add_swig_python_command (swig-gnucash-core
     SWIG_GNUCASH_CORE_C SWIG_GNUCASH_CORE_PY
     gnucash_core.c gnucash_core_c.py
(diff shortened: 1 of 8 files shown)


jralls:
Did I really tell you to wrap QofSessionImpl?
I've read through the code. It's not at all clear to me what you're trying to do.
Suggestion: First, write up your problem and your plan to solve it in English in a post to `gnucash-devel`. It's important to work out the general design before you start coding. Next code to your plan and clean up everything into a single commit for every step in the plan. That doesn't mean that you can't have more commits like "fix typo" and "cleanup white space" *while you're working*, it means that you do those commits in a way that you can squash them into a commit per step so that when you present the work for review the reviewer can step through the commits see a logical progression to the work. The 17 commits in this PR don't tell a story, they just bounce around.
The PR description should either restate the problem and sketch out the design and the steps to accomplish it or link to the email discussion (and links should point at https://lists.gnucash.org/pipermail, *not* at a parasite service like nabble and gmane).

c-holtermann:
Hey John, thank you for the comments. The talk had been in 2018:
https://lists.gnucash.org/pipermail/gnucash-devel/2018-August/042393.html

Sorry for not presenting my stuff in an adequate manner. I'll work on that.

jralls (changes requested):
There's no need to have separate C and C++ wrappers, you can simply make the whole thing gnucash-core.cpp. SWIG is smart enough to deal with both. You can look at the SWIG parts of [my c++options](https://github.com/jralls/gnucash/tree/c%2B%2Boptions/libgnucash/app-utils) branch, in particular [this commit](https://github.com/jralls/gnucash/commit/9fc3e848d0dbc7de64343fee17245e4c6fc52375), for an example.

I'm not convinced that automatically determining whether to set `-c++` based on the output file extension is a good idea; I am sure that basing it on the last characters of a variable name is not. But a simple wrapper function something like
```
macro (gnc_add_swig_c++_python_command _target _out_var _output _input _include_dirs)
    set(SWIG_ARGS "-c++" "-procdoc" "sw-gnc-option-doc" "-procdocformat" "plain")
    gnc_add_swig_python_command(_target _out_var _output _input _include_dirs)
    unset(SWIG_ARGS)
endmacro()
```
might be easier to understand.

c-holtermann:
> There's no need to have separate C and C++ wrappers, you can simply make the whole thing gnucash-core.cpp. SWIG is smart enough to deal with both. You can look at the SWIG parts of [my c++options](https://github.com/jralls/gnucash/tree/c%2B%2Boptions/libgnucash/app-utils) branch, in particular [this commit](https://github.com/jralls/gnucash/commit/9fc3e848d0dbc7de64343fee17245e4c6fc52375), for an example.

Thank you for you valuable feedback. I'm looking into putting the c++ part into the c stuff (or for testing the other way around). I'll need some time for that.

jralls:
Superseded by #2186

PR #733, opened 2020-06-06
cli report upgrades

christopherlam:
Lists available reports, show options modified in a specified report.

I think this is simple enough.

This mechanism would need to change if future reports are saved within datafile.

### DIFF
--- a/gnucash/gnucash-cli.cpp
+++ b/gnucash/gnucash-cli.cpp
@@ -102,6 +102,7 @@ Gnucash::GnucashCli::configure_program_options (void)
     ("report,R", bpo::value (&m_report_cmd),
      _("Execute report related commands. The following commands are supported.\n\n"
      "  list: \tLists available reports.\n"
+     "  show: \tDescribe the options modified in the named report.\n"
      "  run: \tRun the named report in the given GnuCash datafile.\n"))
     ("name", bpo::value (&m_report_name),
      _("Name of the report to run\n"))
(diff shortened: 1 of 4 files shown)


gnucash/gnucash-cli.cpp
```diff
@@ -100,7 +100,9 @@ Gnucash::GnucashCli::configure_program_options (void)
     bpo::options_description report_options(_("Report Generation Options"));
     report_options.add_options()
     ("report,R", bpo::value (&m_report_cmd),
-     _("Execute report related commands. Currently only one command is supported.\n\n"
+     _("Execute report related commands. Currently three commands are supported.\n\n"
```
gjanssens, line 103: No need to count the commands if you have more than one. I added this because it felt a bit odd to make a list of only one command without hinting more would follow in the future. Now you have three, this can simply become
`The following commands are supported.`
And we can add as many as we like in the future without having to change this line ever again.

gnucash/gnucash-cli.cpp
```diff
@@ -150,6 +152,10 @@ Gnucash::GnucashCli::start ([[maybe_unused]] int argc, [[maybe_unused]] char **a
                 return Gnucash::run_report(m_file_to_load, m_report_name,
                                            m_export_type, m_output_file);
         }
+        else if (*m_report_cmd == "list")
+                return Gnucash::report_list ();
```
gjanssens, line 156: I initially was confused for not passing a file name to this command, but eventually it dawned on my saved reports are not stored in our data files yet so we don't need it. That also explains your comment in the PR comment. Perhaps add a comment in the code as well pointing this out. It's not *that* obvious from a distance.

PR #736, opened 2020-06-09
Python bindings - adapt to use of sessionOpenMode in qof_session_begin

c-holtermann:
https://github.com/Gnucash/gnucash/commit/6a1cb5eecd794d65ab0d15161b1628750b29acb6#diff-7692ab9d58b634e3cd67f41873697da5L34 introduced SessionOpenMode as parameters to qof_session_begin. This needs to be reflected in the python bindings. This is a preliminary patch. I'd like to keep the original usage but deprecate it.

### DIFF
--- a/bindings/python/example_scripts/gncinvoice_jinja.py
+++ b/bindings/python/example_scripts/gncinvoice_jinja.py
@@ -39,6 +39,7 @@ try:
     import str_methods
     import jinja2
     from gncinvoicefkt import *
+    from gnucash import SessionOpenMode
 except ImportError as import_error:
     print("Problem importing modules.")
     print(import_error)
@@ -137,7 +138,7 @@ def main(argv=None):
         print("or file://filename")
         print("or mysql://user:password@host/databasename")
         print()
-        print("-f             force open = ignore lock")
+        print("-f             force open = ignore lock (read only)")
         print("-l             list all invoices")
         print("-h or --help   for this help")
         print("-I ID          use invoice ID")
@@ -150,8 +151,15 @@ def main(argv=None):
 
     # Try to open the given input
     try:
-        print("Opening", input_url, ".")
-        session = gnucash.Session(input_url, ignore_lock=ignore_lock)
+        print(
+            "Opening", input_url, " (ignore-lock = read-only)." if ignore_lock else "."
+        )
+        session = gnucash.Session(
+            input_url,
+            SessionOpenMode.SESSION_READ_ONLY
+            if ignore_lock
+            else SessionOpenMode.SESSION_NORMAL_OPEN,
+        )
     except Exception as exception:
         print("Problem opening input.")
         print(exception)
--- a/bindings/python/example_scripts/gncinvoice_jinja.py
+++ b/bindings/python/example_scripts/gncinvoice_jinja.py
@@ -45,9 +45,11 @@ except ImportError as import_error:
     print(import_error)
     sys.exit(2)
 
+
 class Usage(Exception):
     def __init__(self, msg):
         self.msg = msg
+
 
 def main(argv=None):
     if argv is None:
@@ -69,27 +71,27 @@ def main(argv=None):
         try:
             opts, args = getopt.getopt(argv[1:], "fhliI:t:o:OP:", ["help"])
         except getopt.error as msg:
-             raise Usage(msg)
+            raise Usage(msg)
 
         for opt in opts:
             if opt[0] in ["-f"]:
                 print("ignoring lock")
                 ignore_lock = True
-            if opt[0] in ["-h","--help"]:
+            if opt[0] in ["-h", "--help"]:
                 raise Usage("Help:")
             if opt[0] in ["-I"]:
                 invoice_id = opt[1]
-                print ("using invoice ID '" + str(invoice_id) + "'.")
+                print("using invoice ID '" + str(invoice_id) + "'.")
             if opt[0] in ["-i"]:
-                print ("Using ipshell")
+                print("Using ipshell")
                 with_ipshell = True
             if opt[0] in ["-o"]:
                 filename_output = opt[1]
                 print("using output file", filename_output)
             if opt[0] in ["-O"]:
                 if filename_output:
-                    print ("given output filename will be overwritten,")
-                print ("creating output filename from Invoice data.")
+                    print("given output filename will be overwritten,")
+                print("creating output filename from Invoice data.")
                 filename_from_invoice = True
             if opt[0] in ["-t"]:
                 filename_template = opt[1]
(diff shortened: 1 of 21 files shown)


jralls:
If it breaks the bindings how is it that the tests pass?

It does, however, put the bindings out of sync with the C API. Yes, maintaining the old and deprecating it in favor of the new is the correct approach.

Looks like the `SessionOpenMode` enum needs to be imported to Python.

c-holtermann:
The tests don't use params on calling Session(), that's why they don't fail. That should be improved.

c-holtermann:
Now a deprecation warning will be raised if any of the deprecated arguments get used. I need to adapt the docstring and include this to the unittests. I don't see any deprecation warnings in the python bindings. It may be useful to have some deprecation helper functions and decorators.

jralls:
We don't guarantee the stability of the C API yet, so there's no need for deprecation checks on `qof_session_begin`.

You still need to create the `sessionOpenMode` values in python. Yes, it compiles now, but users will have to pass an int instead of the name that explains what the mode does.

c-holtermann:
> We don't guarantee the stability of the C API yet, so there's no need for deprecation checks on `qof_session_begin`.

I was unclear. I thought of checking the python 'Session.begin()' methods arguments as well as the checks I implemented are for 'Session.__init__()'. But I don't know if anyone uses that path.

jralls:
Unfortunately there's no way to tell what parts of API are used or not. That has two implications: First, yes, both Session.begin() and Session.__init__() need to have those arguments deprecated. The second is broader: Minimize the exposed API and keep private as much implementation detail as possible, preferring the natural idioms of the target language. QofSession will eventually morph into a proper C++ class with RAII. Python also encourages that idiom so write the Python Session class to do that, keeping the C++ implementation hidden.

c-holtermann:
In `Session.__init__()` there is a check for is_new. I wonder (https://github.com/Gnucash/gnucash/commit/6a1cb5eecd794d65ab0d15161b1628750b29acb6#r39852123) if it would be useful to provide the deprecated arguments values as consts or enum as they are the basis for calculation of SessionOpenMode.

bindings/python/gnucash_core.py
```diff
@@ -243,7 +243,8 @@ def __init__(self, book_uri=None, sessionOpenMode=None, instance=None):
                 # has to be loaded before it can be altered
                 # Any existing store obviously has to be loaded
                 # More background: https://bugs.gnucash.org/show_bug.cgi?id=726891
-                if book_uri[:3] != "xml" or not is_new:
+                is_new = 2
```
c-holtermann, line 198: It would be useful to have this as const or enum.

bindings/python/function_class.py
```diff
@@ -88,10 +88,11 @@ def get_instance(self):
     def add_method(cls, function_name, method_name):
         """Add the function, method_name to this class as a method named name
         """
-        def method_function(self, *meth_func_args):
+        def method_function(self, *meth_func_args, **meth_func_kwargs):
```
c-holtermann, line 91: This needs some more testing

c-holtermann:
Works for me. Commits need to be cleaned up -> squash, explain, remove irrelevant parts.

bindings/python/example_scripts/gncinvoice_jinja.py
```diff
@@ -150,8 +151,8 @@ def main(argv=None):
 
     # Try to open the given input
     try:
-        print("Opening", input_url, ".")
-        session = gnucash.Session(input_url, ignore_lock=ignore_lock)
+        print("Opening", input_url, " (ignore-lock = read-only)." if ignore_lock else ".")
+        session = gnucash.Session(input_url, SessionOpenMode.SESSION_READ_ONLY if ignore_lock else SessionOpenMode.SESSION_NORMAL_OPEN)
```
jralls, line 155: Can you do something about the line length here, maybe make a `mode` variable on a separate line?

bindings/python/gnucash_core.py
```diff
@@ -28,6 +28,8 @@
 #  @author Jeff Green,   ParIT Worker Co-operative <jeff@parit.ca>
 #  @ingroup python_bindings
 
+from enum import IntEnum
+
```
jralls, line 32: Why the blank line?

bindings/python/gnucash_core.py
jralls: It would be clearest to copy what I did in C++: 
```Python
is_new = (mode in (SessionOpenMode.SESSION_NEW_STORE, SessionOpenMode.SESSION_NEW_OVERWRITE)
if book_uri[:3] != "xml" or not is_new:
```
And as long as you're squashing magic numbers, slicing `book_uri` counts as one.

bindings/python/gnucash_core.py
c-holtermann: I was following [PEP 8](https://www.python.org/dev/peps/pep-0008/#imports):

> Imports should be grouped in the following order:
>
>   1. Standard library imports.
>   2. Related third party imports.
>   3. Local application/library specific imports.
>
>_You should put a blank line between each group of imports._

(italics inserted by me)

Enum is a standard library (group 1) while the following gnucash imports are local/application specific (group 3).

bindings/python/gnucash_core.py
c-holtermann: @jralls I don't understand
> And as long as you're squashing magic numbers, slicing book_uri counts as one.

could you please explain?

jralls: The `3` in `book_uri[:3]` is also a [magic number](https://en.wikipedia.org/wiki/Magic_number_(programming)#Unnamed_numerical_constants). To code it clearly you'd want to find the position of `:` in the URI and get the substring of that (or better call `gnc_uri_get_components()` or python's [`urllib.parse.parse`](https://docs.python.org/3/library/urllib.parse.html#url-parsing)) to set a variable `scheme` and test `scheme == 'xml'`.

OTOH it's not your code so you don't have to clean it up if you don't want to. If you do, there's more: The logic is a bit tortured: `if not (is_new and scheme == 'xml'):` would be clearer.

c-holtermann: I had thought of that too but as you said as it is not my code didn't want to take care of it yet. But the direction you point looks good and I'll implement that.

libgnucash/engine/qofsession.h
```diff
@@ -154,24 +154,28 @@ void qof_session_swap_data (QofSession *session_1, QofSession *session_2);
...
  * `SESSION_NEW_STORE`: Check for an existing file or database at the provided
  * uri and if none is found, create it. If the file or database exists post a
  * QOF_BACKED_STORE_EXISTS and return.
+ * @par
  * `SESSION_READ_ONLY`: Find an existing file or database and open it without
  * disturbing the lock if it exists or setting one if not. This will also set a
  * flag on the book that will prevent many elements from being edited and will
  * prevent the backend from saving any edits.
+ * @par
  * `SESSION_OVERWRITE`: Create a new file or database at the provided uri,
```
pkzw, line 172: Shouldn't it be `SESSION_NEW_OVERWRITE` instead of `SESSION_OVERWRITE`?

c-holtermann: good point. I've changed that too.

jralls:
Thanks!

PR #737, opened 2020-06-14
Bug 797787 - Create new plugin page next to current page

Bob-IT:
Currently new pages are created in the last position which
could result in a lot of scrolling if a large number of
tabs are open. This commit changes that so new pages are
created next to the current page.

@jralls 
@gjanssens 
Not sure if this is OK for 4.0


Bob-IT:
Removed the extra blank lines, forgot to remove as I like to space them out when I am modifying them. Still left based on master but can rebase and push to maint if required.

jralls:
OK, go ahead and push. I'm ambivalent about whether this should go to maint for 4.1 or if it's enough of a behavior change to warrant going into master and so waiting for 5.0, but I'm inclined towards 4.1.

Bob-IT:
Looks like I messed things up here, will close and push new branch to maint.

PR #738, opened 2020-06-15
Fix implementation of mockups

gruberchr:
This PR fixes some open issues on new mockups implemented in PR #650.

### DIFF
--- a/libgnucash/app-utils/mocks/gmock-gnc-prefs.cpp
+++ b/libgnucash/app-utils/mocks/gmock-gnc-prefs.cpp
@@ -10,6 +10,8 @@ gmock_gnc_prefs_set_backend(MockPrefsBackend *backend)
     prefsbackend = backend;
 }
 
+extern "C"
+{
 gboolean
 gnc_prefs_get_bool (const gchar *group, const gchar *pref_name)
 {
@@ -63,3 +65,5 @@ gnc_prefs_get_coords (const gchar *group, const gchar *pref_name, gdouble *x, gd
     if (prefsbackend != nullptr)
         prefsbackend->get_coords(group, pref_name, x, y);
 }
+
+} // extern "C"
--- a/libgnucash/app-utils/mocks/gmock-gnc-prefs.cpp
+++ b/libgnucash/app-utils/mocks/gmock-gnc-prefs.cpp
@@ -13,41 +13,53 @@ gmock_gnc_prefs_set_backend(MockPrefsBackend *backend)
 gboolean
 gnc_prefs_get_bool (const gchar *group, const gchar *pref_name)
 {
-    return prefsbackend->getBool(group, pref_name);
+    EXPECT_NE(prefsbackend, nullptr);
+    return prefsbackend ? prefsbackend->getBool(group, pref_name) : FALSE;
 }
 
 gint
 gnc_prefs_get_int (const gchar *group, const gchar *pref_name)
 {
-    return prefsbackend->getInt(group, pref_name);
+    EXPECT_NE(prefsbackend, nullptr);
+    return prefsbackend ? prefsbackend->getInt(group, pref_name) : 0;
 }
 
 gint64
 gnc_prefs_get_int64 (const gchar *group, const gchar *pref_name)
 {
-    return prefsbackend->getInt64(group, pref_name);
+    EXPECT_NE(prefsbackend, nullptr);
+    return prefsbackend ? prefsbackend->getInt64(group, pref_name) : 0;
 }
 
 gdouble
 gnc_prefs_get_float (const gchar *group, const gchar *pref_name)
 {
-    return prefsbackend->getFloat(group, pref_name);
+    EXPECT_NE(prefsbackend, nullptr);
+    return prefsbackend ? prefsbackend->getFloat(group, pref_name) : 0.0;
 }
 
 gchar *
 gnc_prefs_get_string (const gchar *group, const gchar *pref_name)
 {
-    return prefsbackend->getString(group, pref_name);
+    EXPECT_NE(prefsbackend, nullptr);
+    return prefsbackend ? prefsbackend->getString(group, pref_name) : NULL;
 }
 
 gint
 gnc_prefs_get_enum (const gchar *group, const gchar *pref_name)
 {
-    return prefsbackend->getEnum(group, pref_name);
+    EXPECT_NE(prefsbackend, nullptr);
+    return prefsbackend ? prefsbackend->getEnum(group, pref_name) : 0;
 }
 
 void
 gnc_prefs_get_coords (const gchar *group, const gchar *pref_name, gdouble *x, gdouble *y)
 {
-    prefsbackend->getCoords(group, pref_name, x, y);
+    EXPECT_NE(prefsbackend, nullptr);
+
+    *x = 0.0;
+    *y = 0.0;
+
+    if (prefsbackend != nullptr)
+        prefsbackend->getCoords(group, pref_name, x, y);
 }
--- a/libgnucash/app-utils/mocks/gmock-gnc-prefs.cpp
+++ b/libgnucash/app-utils/mocks/gmock-gnc-prefs.cpp
@@ -2,46 +2,52 @@
 
 #include "gmock-gnc-prefs.h"
 
-PrefsBackend* prefsbackend = NULL;
+static MockPrefsBackend* prefsbackend = nullptr;
+
+void
+gmock_gnc_prefs_set_backend(MockPrefsBackend *backend)
+{
+    prefsbackend = backend;
+}
 
 gboolean
 gnc_prefs_get_bool (const gchar *group, const gchar *pref_name)
 {
-    return ((MockPrefsBackend*)prefsbackend)->getBool(group, pref_name);
+    return prefsbackend->getBool(group, pref_name);
 }
 
 gint
 gnc_prefs_get_int (const gchar *group, const gchar *pref_name)
 {
-    return ((MockPrefsBackend*)prefsbackend)->getInt(group, pref_name);
+    return prefsbackend->getInt(group, pref_name);
 }
 
 gint64
 gnc_prefs_get_int64 (const gchar *group, const gchar *pref_name)
 {
-    return ((MockPrefsBackend*)prefsbackend)->getInt64(group, pref_name);
+    return prefsbackend->getInt64(group, pref_name);
 }
 
 gdouble
 gnc_prefs_get_float (const gchar *group, const gchar *pref_name)
 {
-    return ((MockPrefsBackend*)prefsbackend)->getFloat(group, pref_name);
+    return prefsbackend->getFloat(group, pref_name);
 }
 
 gchar *
 gnc_prefs_get_string (const gchar *group, const gchar *pref_name)
 {
-    return ((MockPrefsBackend*)prefsbackend)->getString(group, pref_name);
+    return prefsbackend->getString(group, pref_name);
 }
 
 gint
 gnc_prefs_get_enum (const gchar *group, const gchar *pref_name)
 {
-    return ((MockPrefsBackend*)prefsbackend)->getEnum(group, pref_name);
+    return prefsbackend->getEnum(group, pref_name);
 }
 
 void
 gnc_prefs_get_coords (const gchar *group, const gchar *pref_name, gdouble *x, gdouble *y)
 {
-    ((MockPrefsBackend*)prefsbackend)->getCoords(group, pref_name, x, y);
+    prefsbackend->getCoords(group, pref_name, x, y);
 }
--- a/libgnucash/app-utils/mocks/gmock-gnc-prefs.cpp
+++ b/libgnucash/app-utils/mocks/gmock-gnc-prefs.cpp
@@ -14,42 +14,42 @@ gboolean
 gnc_prefs_get_bool (const gchar *group, const gchar *pref_name)
 {
     EXPECT_NE(prefsbackend, nullptr);
-    return prefsbackend ? prefsbackend->getBool(group, pref_name) : FALSE;
+    return prefsbackend ? prefsbackend->get_bool(group, pref_name) : FALSE;
 }
 
 gint
 gnc_prefs_get_int (const gchar *group, const gchar *pref_name)
 {
     EXPECT_NE(prefsbackend, nullptr);
-    return prefsbackend ? prefsbackend->getInt(group, pref_name) : 0;
+    return prefsbackend ? prefsbackend->get_int(group, pref_name) : 0;
 }
 
 gint64
 gnc_prefs_get_int64 (const gchar *group, const gchar *pref_name)
 {
     EXPECT_NE(prefsbackend, nullptr);
-    return prefsbackend ? prefsbackend->getInt64(group, pref_name) : 0;
+    return prefsbackend ? prefsbackend->get_int64(group, pref_name) : 0;
 }
 
 gdouble
 gnc_prefs_get_float (const gchar *group, const gchar *pref_name)
 {
     EXPECT_NE(prefsbackend, nullptr);
-    return prefsbackend ? prefsbackend->getFloat(group, pref_name) : 0.0;
+    return prefsbackend ? prefsbackend->get_float(group, pref_name) : 0.0;
 }
 
 gchar *
 gnc_prefs_get_string (const gchar *group, const gchar *pref_name)
 {
     EXPECT_NE(prefsbackend, nullptr);
-    return prefsbackend ? prefsbackend->getString(group, pref_name) : NULL;
+    return prefsbackend ? prefsbackend->get_string(group, pref_name) : NULL;
 }
 
 gint
 gnc_prefs_get_enum (const gchar *group, const gchar *pref_name)
 {
     EXPECT_NE(prefsbackend, nullptr);
-    return prefsbackend ? prefsbackend->getEnum(group, pref_name) : 0;
+    return prefsbackend ? prefsbackend->get_enum(group, pref_name) : 0;
 }
 
 void
@@ -61,5 +61,5 @@ gnc_prefs_get_coords (const gchar *group, const gchar *pref_name, gdouble *x, gd
     *y = 0.0;
 
     if (prefsbackend != nullptr)
-        prefsbackend->getCoords(group, pref_name, x, y);
+        prefsbackend->get_coords(group, pref_name, x, y);
 }
--- a/libgnucash/engine/mocks/fake-qofquery.cpp
+++ b/libgnucash/engine/mocks/fake-qofquery.cpp
@@ -1,6 +1,6 @@
 #include <config.h>
 
-#include "gmock-qofquery.h"
+#include "fake-qofquery.h"
 #include "gmock-qofbook.h"
 
 QofQuery *
--- a/libgnucash/engine/mocks/fake-qofquery.cpp
+++ b/libgnucash/engine/mocks/fake-qofquery.cpp
@@ -1,46 +1,109 @@
 #include <config.h>
 
+#include <list>
+
 #include "fake-qofquery.h"
 #include "gmock-qofbook.h"
+
+
+
+/* class QofFakeQueryPool */
+
+static class QofFakeQueryPool
+{
+public:
+    void addQuery(QofFakeQuery *query)
+    {
+        m_queriesNew.push_back(query);
+    }
+
+    QofFakeQuery* requestQuery()
+    {
+        QofFakeQuery* query = nullptr;
+
+        if (!m_queriesNew.empty())
+        {
+            query = m_queriesNew.front();
+            m_queriesNew.pop_front();
+            m_queriesUsed.push_back(query);
+        }
+
+        EXPECT_NE(query, nullptr);
+        return query;
+    }
+
+    bool queryUsed(QofQuery *query)
+    {
+        auto it = std::find(m_queriesUsed.begin(), m_queriesUsed.end(), (QofFakeQuery*)query);
+
+        return (it != m_queriesUsed.end());
+    }
+
+    void releaseQuery(QofFakeQuery *query)
+    {
+        ASSERT_TRUE(query_used((QofQuery*)query));
+        auto it = std::find(m_queriesUsed.begin(), m_queriesUsed.end(), query);
+        m_queriesUsed.erase(it);
+        m_queriesConsumed.push_back(*it);
+    }
+
+    void removeQuery(QofFakeQuery *query)
+    {
+        ASSERT_FALSE(queryUsed((QofQuery*)query));
+        auto it = std::find(m_queriesConsumed.begin(), m_queriesConsumed.end(), (QofFakeQuery*)query);
+        if (it != m_queriesConsumed.end())
+            m_queriesConsumed.erase(it);
+        else
+        {
+            it = std::find(m_queriesNew.begin(), m_queriesNew.end(), (QofFakeQuery*)query);
+            bool query_found = (it != m_queriesNew.end());
+            ASSERT_TRUE(query_found);
+            m_queriesNew.erase(it);
+        }
+    }
+
+private:
+    std::list<QofFakeQuery*> m_queriesNew {};
+    std::list<QofFakeQuery*> m_queriesUsed {};
+    std::list<QofFakeQuery*> m_queriesConsumed {};
+} queryPool;
+
+
+
+/* class QofFakeQuery */
+
+QofFakeQuery::QofFakeQuery()
+{
+    queryPool.addQuery(this);
+}
+
+QofFakeQuery::~QofFakeQuery()
+{
+    queryPool.removeQuery(this);
+}
+
+
+
+/* mock functions */
 
 QofQuery *
 qof_query_create_for (QofIdTypeConst obj_type)
 {
-    return (QofQuery*)qof_query_factory.create();
-/*
-    // \todo create typed query objects
-    QofQuery *ret = NULL;
+    return (QofQuery*)queryPool.requestQuery();
+}
 
-    if (g_strcmp0(obj_type, GNC_ID_SPLIT) == 0)
-        ret = (QofQuery*)qof_query_factory.createForSplit();
-//    else
-//        FAIL();
-
-    return ret;
-*/
+void
+qof_query_destroy (QofQuery *query)
+{
+    queryPool.releaseQuery((QofFakeQuery*)query);
 }
 
 void
 qof_query_set_book (QofQuery *query, QofBook *book)
 {
+    ASSERT_TRUE(queryPool.queryUsed(query));
     ASSERT_TRUE(QOF_IS_MOCK_BOOK(book));
     ((QofFakeQuery*)query)->setBook(book);
-}
-
-GList *
-qof_query_run (QofQuery *query)
-{
-    GList *matching_objects = NULL;
-
-    // \todo use typed mock objects
-    auto matchingObjects = ((QofFakeQuery*)query)->run();
-
-    for (auto object : matchingObjects)
-    {
-        matching_objects = g_list_append(matching_objects, static_cast<gpointer>(object));
-    }
-
-    return matching_objects;
 }
 
 void
@@ -52,17 +115,33 @@ xaccQueryAddDateMatchTT (
         time64 ett,
         QofQueryOp op)
 {
+    ASSERT_TRUE(queryPool.queryUsed(query));
     ((QofFakeQuery*)query)->addDateMatchTT(use_start, stt, use_end, ett, op);
 }
 
 void
 xaccQueryAddSingleAccountMatch(QofQuery *query, Account *acc, QofQueryOp op)
 {
+    ASSERT_TRUE(queryPool.queryUsed(query));
     ((QofFakeQuery*)query)->addSingleAccountMatch(acc, op);
 }
 
-void
-qof_query_destroy (QofQuery *query)
+GList *
+qof_query_run (QofQuery *query)
 {
-    ((QofFakeQuery*)query)->destroy();
+    GList *matching_objects = NULL;
+    bool  query_used        = queryPool.queryUsed(query);
+
+    EXPECT_TRUE(query_used);
+    if (query_used)
+    {
+        auto matchingObjects = ((QofFakeQuery*)query)->run();
+
+        for (auto object : matchingObjects)
+        {
+            matching_objects = g_list_append(matching_objects, static_cast<gpointer>(object));
+        }
+    }
+
+    return matching_objects;
 }
--- a/libgnucash/engine/mocks/fake-qofquery.cpp
+++ b/libgnucash/engine/mocks/fake-qofquery.cpp
@@ -124,6 +124,8 @@ qof_query_set_book (QofQuery *query, QofBook *book)
     ((QofFakeQuery*)query)->set_book(book);
 }
 
+extern "C"
+{
 void
 xaccQueryAddDateMatchTT (
         QofQuery *query,
@@ -143,6 +145,8 @@ xaccQueryAddSingleAccountMatch(QofQuery *query, Account *acc, QofQueryOp op)
     ASSERT_TRUE(queryPool.query_used(query));
     ((QofFakeQuery*)query)->add_single_account_match(acc, op);
 }
+
+} // extern "C"
 
 GList *
 qof_query_run (QofQuery *query)
--- a/libgnucash/engine/mocks/fake-qofquery.cpp
+++ b/libgnucash/engine/mocks/fake-qofquery.cpp
@@ -1,3 +1,7 @@
+/**
+ * @file fake-qofquery.cpp
+*/
+
 #include <config.h>
 
 #include <list>
@@ -12,11 +16,14 @@
 static class QofFakeQueryPool
 {
 public:
+
+    /* Add QofFakeQuery object to the pool */
     void add_query(QofFakeQuery *query)
     {
         m_queriesNew.push_back(query);
     }
 
+    /* Request to use a QofFakeQuery object */
     QofFakeQuery* request_query(QofIdTypeConst obj_type)
     {
         QofFakeQuery* query = nullptr;
@@ -36,6 +43,8 @@ public:
         return query;
     }
 
+    /* Check if a QofFakeQuery object is currently used, i.e. it has been
+     * requested before */
     bool query_used(QofQuery *query)
     {
         auto it = std::find(m_queriesUsed.begin(), m_queriesUsed.end(), (QofFakeQuery*)query);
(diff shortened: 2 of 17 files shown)


libgnucash/engine/mocks/fake-qofquery.cpp
```diff
@@ -0,0 +1,155 @@
...
+        auto it = std::find(m_queriesUsed.begin(), m_queriesUsed.end(), (QofFakeQuery*)query);
+
+        return (it != m_queriesUsed.end());
+    }
+
+    void releaseQuery(QofFakeQuery *query)
+    {
+        auto it           = std::find(m_queriesUsed.begin(), m_queriesUsed.end(), query);
+        bool query_in_use = (it != m_queriesUsed.end());
+        ASSERT_TRUE(query_in_use);
```
jralls, line 52: Use `ASSERT_TRUE(queryUsed(query))`.

libgnucash/engine/mocks/fake-qofquery.cpp
```diff
@@ -0,0 +1,155 @@
...
+
+
+/* class QofFakeQueryPool */
+
+static class QofFakeQueryPool
+{
+public:
+    QofFakeQueryPool() {};
+
+    void addQuery(QofFakeQuery *query)
```
jralls, line 17: Camel case is for type names only, use snake case for function names. See our [C++ Coding Guide](https://wiki.gnucash.org/wiki/CodingStandard#C.2B.2B_Format)

libgnucash/engine/mocks/fake-qofquery.cpp
```diff
@@ -0,0 +1,155 @@
...
+#include <list>
+
+#include "fake-qofquery.h"
+#include "gmock-qofbook.h"
+
+
+
+/* class QofFakeQueryPool */
+
+static class QofFakeQueryPool
```
jralls, line 12: This needs Doxygen documentation explaining how to use the query pool, highlighting the life-cycle of a QofFakeQuery and the importance of ensuring that `query_pool->request_query` and `query_pool->release_query` are called on every `QofFakeQuery` or the test program will `ASSERT`.

gruberchr:
Let me explain, why I decided for a pool to manage the life cycle of `QofFakeQuery` objects.

The `QofQuery` Interface, which should be mocked, comprises the functions
- `qof_query_create_for()`
- several functions to describe the search criteria on a created `QofQuery` object
- `qof_query_run()`
- `qof_query_destroy()`

During the life cycle of a query object, these functions are invoked in this order. Function `qof_query_create_for()` is expected to return a new query object, which should be a mock object here, when mocking the `QofQuery` interface. The expected behaviour of a mock object has to be described by the test application before the mock object is used. Since there is no interaction with the test application code during the call to function `qof_query_create_for()` the query object has to be created and its behaviour has to be described before the test application calls `qof_query_create_for()`. To provide these prepared query objects to the function `qof_query_create_for()`, I decided to use a pool. This pool also controls the further life cycle of the prepared query objects until they are destroyed with `qof_query_destroy()`.

libgnucash/engine/mocks/fake-qofquery.cpp
```diff
@@ -1,46 +1,112 @@
...
 #include "gmock-qofbook.h"
 
+
+
+/* class QofFakeQueryPool */
+
+static class QofFakeQueryPool
+{
+public:
+    QofFakeQueryPool() {};
```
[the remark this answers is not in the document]
gruberchr: Ok, then I would omit the definition. This has to be changed for the constructor of class `MockAccount` as well, right?

libgnucash/engine/mocks/fake-qofquery.cpp
gruberchr: You're right, this has the same behaviour. My intention was to do an explicit assertion on the iterator `it` used here instead of an assertion on a function, which searches for the same iterator implicitely.

libgnucash/engine/mocks/fake-qofquery.cpp
gruberchr: Oh, this means I have to change the function names of all mock class methods now, right?

libgnucash/engine/mocks/fake-qofquery.cpp
gruberchr: No, it's not necessary, that a `QofFakeQuery` object is requested (and released) by the test application. It can also be removed from the pool, if it has not been consumed. See the implementation of `QofFakeQueryPool::removeQuery()`. If a `QofFakeQuery` object is not found in the list of consumed query objects, it is searched in the list of new query objects. There are only two assertions in this function:
- when the `QofFakeQuery` object is currently in use (requested, but not released yet)
- when the `QofFakeQuery` object has never been added to the pool (which is actually not possible)

libgnucash/app-utils/mocks/gmock-gnc-prefs.cpp
```diff
@@ -13,41 +13,71 @@ gmock_gnc_prefs_set_backend(MockPrefsBackend *backend)
 gboolean
 gnc_prefs_get_bool (const gchar *group, const gchar *pref_name)
 {
-    return prefsbackend->getBool(group, pref_name);
+    EXPECT_NE(prefsbackend, nullptr);
+    if (prefsbackend != nullptr)
+        return prefsbackend->getBool(group, pref_name);
+    else
+        return FALSE;
```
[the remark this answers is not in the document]
gruberchr: Looks good. Changed.

libgnucash/engine/mocks/fake-qofquery.cpp
jralls: That's just useless code duplication. It's no less explicit in the function call.

libgnucash/engine/mocks/fake-qofquery.cpp
jralls: Yes.

libgnucash/engine/mocks/fake-qofquery.cpp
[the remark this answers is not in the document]
jralls: Yes.

libgnucash/engine/mocks/fake-qofquery.cpp
jralls: There are three asserts: One in `releaseQuery` if the query isn't in use as well as the two in `removeQuery` that you describe. `removeQuery` is run by the query's dtor and if the query is never requested it will indeed happily delete it. If OTOH it has been requested and not released then the `removeQuery` will assert; since there's no point in creating the query if you're not going to request it, it's important to request and release during the query's lifetime and the documentation needs to explain that.

libgnucash/engine/mocks/fake-qofquery.cpp
gruberchr: Ok, I changed it

libgnucash/engine/mocks/fake-qofquery.cpp
gruberchr: Done

libgnucash/engine/mocks/fake-qofquery.cpp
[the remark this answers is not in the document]
gruberchr: That aspect was very interesting. I removed the empty constructor for classes `QofFakeQueryPool` and `MockPrefsBackend` and everything was fine.

But after removing the empty constructor for class `MockAccount`, the test application crashed with a `SIGSEV` in the `delete` operator. What happened?

I debugged this and found out, that the default constructor does zero initialization and the empty constructor does not. And since the constructor runs *after* the `new` operator, which does `gobject` initialization, the default constructor overwrites all `gobject` data, but the empty constructor does not.

cppreference says about [value initialization](https://en.cppreference.com/w/cpp/language/value_initialization) since C++11: *"if T is a class type with no default constructor or with a user-provided or deleted default constructor, the object is default-initialized"*, but *"if T is a class type with a default constructor that is neither user-provided nor deleted (that is, it may be a class with an implicitly-defined or defaulted default constructor), the object is zero-initialized and then it is default-initialized if it has a non-trivial default constructor"*

So we can not remove the empty constructor here.

libgnucash/engine/mocks/fake-qofquery.cpp
gruberchr: What doxygen style do you prefer? I've seen Javadoc style quite often in the code. Is there any style guide for code documentation as well?

libgnucash/engine/mocks/fake-qofquery.cpp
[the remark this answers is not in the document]
jralls: Interesting. That either needs a comment or a second think about doing GObject initialization in operator new (and I suppose dispose/finalize in operator delete). As I'm sure you know, operators new and delete are for allocating and freeing memory and initialization/finalization are supposed to happen in the constructor and destructor.

libgnucash/engine/mocks/fake-qofquery.cpp
jralls: I think most of it's `/** .. */` with `@` for keywords. Best to stay consistent with that.

libgnucash/engine/mocks/fake-qofquery.cpp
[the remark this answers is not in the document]
gruberchr: Yes, I know that. But since `g_object_new()` does allocation and initialization in one, you can not simply separate both, except of reimplementing parts of this function, which is challenging after looking into glib code.

On the other hand at least the correct order of allocation and initialization is fulfilled. At first memory is allocated. And the first step of object construction would be the initialization of the base class anyway. Since `GObject` can be seen as the "base class", everything is fine. The next step is the construction of the derived class, which is done by the C++ ctor. The only problem is, that this base class is hidden to the C++ compiler. Therefore some coding rules have to be followed as for instance using empty constructors instead of default constructors.

I don't see another opportunity at the moment except of reimplementing `g_object_new()` (and `g_object_unref()`).

I'll add a comment.

libgnucash/engine/mocks/fake-qofquery.cpp
gruberchr: Class `QofFakeQueryPool` is a `static` class and at the moment doxygen does not generate documentation for static classes according to the current doxygen configuration (see `EXTRACT_STATIC = NO`). That means doxygen comments on this class are not part of the doxygen documentation.

Is it intended, that `EXTRACT_STATIC` is set to `NO`? And should class `QofFakeQueryPool` be documented anyway?

jralls: That's reasonable, `QofFakeQueryPool` is an implementation detail so it probably shouldn't be documented in Doxygen. The documentation would be better on `QofFakeQuery`.

libgnucash/engine/mocks/fake-qofquery.h
```diff
@@ -0,0 +1,51 @@
...
+
+extern "C"
+{
+#include <Query.h>
+}
+
+/** Fake object providing functionality similar to \c QofQuery
+ *
+ * @note \c QofQuery is a \c typedef for \c struct _QofQuery, which is not
+ * public. Therefore class QofFakeQuery is not derived from \c QofQuery.
```
jralls, line 22: `@c` instead of `\c`, and note that `@c struct _QofQuery` puts only `struct` in tt-face; to get both in use either `<tt>struct _QofQuery</tt>` or `@code struct _QofQuery @endcode`. But be consistent: This is the only place you marked words for tt-face.

libgnucash/engine/mocks/fake-qofquery.h
```diff
@@ -0,0 +1,51 @@
...
+ * To use a QofFakeQuery object simply create it before the GnuCash code
+ * performs a query. Check that the QofFakeQuery object is created with the
+ * correct object type. Also define all expectations and return values on the
+ * created QofFakeQuery object before the GnuCash code performs the query.
+ *
+ * After the query is finished, the QofFakeQuery object can be destroyed.
+ *
+ * @note Internally each created QofFakeQuery object is registered at a
+ * QofFakeQueryPool, which provides it to the GnuCash code on request.
+ * This pool observes the life-cycle of each QofFakeQuery object.
```
jralls, line 33: This would a ggod place to explain that the test will fail with a GTest `ASSERT` in QofFakeQueryPool if `qof_query_create_for()` and `qof_query_destroy()` aren't paired.

libgnucash/engine/mocks/gmock-Account.h
```diff
@@ -20,6 +20,9 @@ GType gnc_mock_account_get_type(void);
 class MockAccount : public Account
 {
 public:
+    // note: don't use default constructor instead of empty constructor, since
+    // it does zero initialization, which would overwrite GObject
+    // initialization, which is already done in the new operator.
```
jralls, line 25: Please use `/*...*/` for multi-line comments.

libgnucash/engine/mocks/gmock-Split.cpp
```diff
@@ -30,113 +30,143 @@ gnc_mock_split_class_init (MockSplitClass *klass)
...
 QofBook *
 xaccSplitGetBook (const Split *split)
 {
-    g_return_val_if_fail(GNC_IS_MOCK_SPLIT(split), NULL);
-    return ((MockSplit*)split)->getBook();
+    EXPECT_TRUE(GNC_IS_MOCK_SPLIT(split));
+    if (GNC_IS_MOCK_SPLIT(split))
+        return ((MockSplit*)split)->get_book();
+    else
+        return nullptr;
```
jralls, line 47: It's [conventional for `GObjects`](https://developer.gnome.org/gobject/stable/gtype-conventions.html) to have a macro to do this, in this case named `GNC_MOCK_SPLIT`. Conventionally the macro would be implemented as
```C
#define GNC_MOCK_SPLIT (split) \
   g_return_val_if_fail (GNC_IS_MOCK_SPLIT (split), NULL)
```
so that it logs an error as well, but I think that would be undesirable in test code.

jralls: Never mind, that's not what you're doing.

libgnucash/engine/mocks/fake-qofquery.h
gruberchr: > `@c` instead of `\c`

Done

> and note that `@c struct _QofQuery` puts only `struct` in tt-face [...] But be consistent: This is the only place you marked words for tt-face

The point is, I tried to mark up only names like `QofQuery`, which are not linked to any doxygen generated documentation. Therefore I used tt-face for `QofQuery` but not for `_QofQuery` and `QofFakeQuery` for instance. But this is actually problematic, since missing doxygen documentation may be only a temporary state or a question of doxygen configuration. Therefore I removed most of the `@c` and kept it only for `typedef` and `struct`.

Or should I completely remove all `@c`?

libgnucash/engine/mocks/gmock-Split.cpp
gruberchr: I already replaced `g_return_if_fail()` and `g_return_val_if_fail()` by `ASSERT_TRUE` and `EXPECT_TRUE` in commit bb0088c, since you mentioned this in PR #650. That means `g_return_val_if_fail()` has to be replaced by `if GNC_IS_MOCK_SPLIT (split) { return NULL }` in this macro?

What about cases with return values other than `NULL` or `nullptr` in case of incorrect type?

Shouldn't this macro also return a type-casted pointer?

libgnucash/engine/mocks/gmock-Account.h
gruberchr: Done

libgnucash/engine/mocks/gmock-Split.cpp
jralls: I didn't show the whole macro. The whole thing is
```C
#define GNC_MOCK_SPLIT (split)                                        \
{                                                                                              \
    g_return_val_if_fail (GNC_IS_MOCK_SPLIT (split), NULL); \
    return (MockSplit*)split;                                                              \
}
```
The point being to create a checked typecast. That's necessary because C is a bit loose on type enforcement for pointer types and doesn't really understand types beyond its POD ones. So GObject has a run-time type system that requires runtime checks and carefully nested structs to provide a semblance of polymoprhism. C++ has all of that built into the language, understands types very deeply, and enforces good type safety at compile time. That's been tightened up in the last 10 years. "NULL" is a C hack: In the C headers somewhere you'll find `#define NULL (void*)0`, while in C++11 and later `nullptr` is a language keyword that is typeless, the intent being to (and soon, we just got a bug that gcc11 enforces this) make it illegal to return `void*` when the function signature is `foo*`.

gruberchr: I'm a bit confused. In PR #650 you commented code line
```C
g_return_if_fail(GNC_IS_MOCK_ACCOUNT(account));
```
with

> ISTM you should use EXPECT_TRUE() or maybe ASSERT_TRUE instead of g_return_if_fail. The latter will just write a message to std::cerr instead of causing the test to fail.

Therefore I replaced all `g_return_if_fail()` with `ASSERT_TRUE` in commit bb0088c, which is straightforward, if the current function has no return value ([`ASSERT_TRUE`](https://github.com/google/googletest/blob/release-1.8.1/googletest/docs/primer.md#basic-assertions) is a fatal assertion, i.e. it immediately aborts the current function, which is similar to the behaviour of `g_return_if_fail()`.
But did this comment also mean, that all `g_return_val_if_fail()` should be replaced by `EXPECT_TRUE`? This is, what I did as well in commit bb0088c. This contradicts to your proposal to use `g_return_val_if_fail()` in macros like `GNC_MOCK_SPLIT`. Or should `g_return_val_if_fail()` only be complemented with `EXPECT_TRUE`?

Would it make sense to define macro `GNC_MOCK_SPLIT` as
```C
#define GNC_MOCK_SPLIT (split)                              \
{                                                           \
    EXPECT_TRUE(GNC_IS_MOCK_SPLIT(split));                  \
    g_return_val_if_fail (GNC_IS_MOCK_SPLIT(split), NULL);  \
    return (MockSplit*)split;                               \
}
```
or
```C
#define GNC_MOCK_SPLIT (split)                              \
{                                                           \
    EXPECT_TRUE(GNC_IS_MOCK_SPLIT(split));                  \
    if (GNC_IS_MOCK_SPLIT(split))                           \
        return (MockSplit*)split;                           \
    else                                                    \
        return NULL;                                        \
}
```
?

jralls: Yes, `EXPECT_TRUE` is better that `g_return_if_fail`. `EXPECT_TRUE` will fail the test, `g_return_if_fail` won't. You might do instead
```
static inline MockSplit*
mock_split(Split* split)
{
    if (GNC_IS_MOCK_SPLIT(split))
           return static_cast<MockSplit*>(split);
    ADD_FAILURE() << "Not a MockSplit!";
    return nullptr;
}
```
So you need do the test only once. This is C++, better to not use the preprocessor. BTW you'll need a second one for `const Split*` -> `const MockSplit*`.

I decided to build a simple one-function test of Account using MockSplit and QofMockBook so I could play around with this a bit. You can see the result at https://github.com/jralls/gnucash/tree/gmock-tryout. I had to spend a fair amount of time building out the mocked functions and I opted for not doing checks in them. But even though I know GnuCash is spaghetti like I was surprised how long it took to collect all of the source files needed to compile Account.cpp (56, not including the mocks).

Some important lessons:
* The C replacement function definitions need to be wrapped in `extern "C" {...}` unless the real implementation file is C++; otherwise they get C++ linkage and the linker won't recognize them.
* Functions that take a `const` first argument need to be mocked with `MOCK_CONST_METHOD`.
* `foo_get_type` needs to be faked to `return mock_foo_get_type()` so that the runtime GType checking in the code under test succeeds.
* Runtime checks of other classes in mock code should check for the base-class type, not for its mock-class if it has one. Mocks shouldn't assume that they'll be used only with the other mocks.

gruberchr: Using `ADD_FAILURE()` to avoid testing twice is a good idea. I'll implement it this way.

But do you also mean to implement the type conversion as  a function, `mock_split()` in this case? I see this in your code on branch `gmock-tryout`. Or would it still be sufficient to write a macro `GNC_MOCK_SPLIT` with the content of your example `mock_split()` according to the GObject conventions?

If a function is intended, it should better be named `gnc_mock_split()`, I guess.

>  BTW you'll need a second one for const Split* -> const MockSplit*.

How is this done in other GObject code, if type conversion is implemented as a macro?

>  The C replacement function definitions need to be wrapped in `extern "C" {...}` unless the real implementation file is C++; otherwise they get C++ linkage and the linker won't recognize them.

The *definitions*, not the *declarations* only? So it isn't sufficient to wrap `#include <Split.h>` and `#include <SplitP.h>` in *gmock-Split.h*? Why is the test application `test-import-backend` already linked successfully then? I don't have precise knowledge about C++ linkage.

jralls: Preprocessor macros are *very* strongly discouraged in C++, so yes, an `inline static` function is better. GObject is in C and abuses macros and other C features rather heavily in its lame efforts to replicate C++. It's something to use when one must but not something to emulate.

Why would you name the class `MockSplit` and want to call the casting function `gnc_mock_split`? 

> I don't have precise knowledge about C++ linkage.

You need to learn about that, it's important when integrating C and C++. 

The short explanation is that C++ "mangles" function names with information about their namespaces and parameters, which is necessary to make namespaces and overloading work. C doesn't, so in order to be able to use a C++ function from C you have to tell the compiler not to mangle it by declaring it `extern "C"`. Marking the decls only works if the marked decls are the first ones for the function that the compiler sees because the compilation guards prevent the compiler from looking inside subsequent ones. The compiler will also raise an error if it encounters multiple decls with differing signatures and `extern "C"` is part of the signature. 

> Why is the test application test-import-backend already linked successfully then?

Because it's all C++.

gruberchr: >  Why would you name the class `MockSplit` and want to call the casting function `gnc_mock_split`?

Because of the GObject type name `gnc_mock_split`, which is used for the GObject related functions `gnc_mock_split_init()` and `gnc_mock_split_class_init()` already.

I have another question related to your code on branch `gmock-tryout`. In contrast to your definition of function `mock_split()` here (see above) you don't do any GObject runtime type check on branch `gmock-tryout` before type conversion. This is only left out for your private use, right?

>  Marking the decls only works if the marked decls are the first ones for the function that the compiler sees because the compilation guards prevent the compiler from looking inside subsequent ones.

But doesn't this mean, that it's sufficient to mark the declarations only, since the header file containing the declarations is included in the cpp file before the definitions? And do I understand https://en.cppreference.com/w/cpp/language/language_linkage correctly, that using notation `extern "C" { ... }` marks the declaration and the definition, even if only the declaration is contained inside the braces?

jralls: > Because of the GObject type name gnc_mock_split

Then the class should be GncMockSplit.

If we were using gobject-introspection (be glad that we don't!) the class would have to be GncMocksplit and the functions gnc_mocksplit_ for the scanner to be able to correctly separate the namespace, class, and function names.

> I have another question related to your code on branch gmock-tryout...

Not entirely. I was also experimenting with whether the compiler would protect the mock code from getting a non-Split*. It does as long as everything is C++ so it works in this case and would as well in the tests you wrote. It's not safe generally so the checks should go in.

> But doesn't this mean, that it's sufficient to mark the declarations only, since the header file containing the declarations is included in the cpp file before the definitions? And do I understand https://en.cppreference.com/w/cpp/language/language_linkage correctly, that using notation extern "C" { ... } marks the declaration and the definition, even if only the declaration is contained inside the braces?

You do understand the cppreference.com item correctly. What you're missing is that just because `#include <Split.h>` is in this file before the definitions doesn't mean that that particular include is the one that actually includes the declarations. If `Split.h` is already indirectly included then the compile guard has already been invoked and the later include doesn't do anything.

I think that's actually not a problem for MockSplit. It *was* a problem for QofBook because Qof has a rather convoluted inclusion arrangement thanks to its time as a separate library.

gruberchr: >  Then the class should be GncMockSplit.

And why is the GObject class `Split` defined as follows then?
```
G_DEFINE_TYPE(Split, gnc_split, QOF_TYPE_INSTANCE)
```
It's the same naming scheme. The camel case type name is `Split` and the lowercase type name is `gnc_split`.

>  If Split.h is already indirectly included then the compile guard has already been invoked and the later include doesn't do anything.

That's it, I got it.

jralls: Most of the split functions are named xaccSplitFoo because Split is one of the original X-Accounting classes from before Gnome. The port to GObject came almost 10 years later and when @derekatkins [converted it to GObject](https://github.com/Gnucash/gnucash/commit/cb8f58790c3a59f3236d3f66adfc99ef97241695) he didn't change the name. BTW he often says he regrets having done that instead of making them C++ classes.

gruberchr: Interesting. This means that `Split` should have been actually named `GncSplit`? GObject recommends to use prefixes as namespaces, which is `Gnc` or `gnc_` here.

And if I understand you correctly, the prefix should be omitted for the mock classes, since it's C++. But this means, the lower case type name of `MockSplit` should be changed to `mock_split` instead of `gnc_mock_split`, right?

jralls: > Sorry, I ask again regarding placement of extern "C". Wouldn't it be the best to wrap the declarations directly in the C headers together with the #ifdef __cplusplus guard? This would avoid the described problem, that someone could include Split.h for instance without extern "C" wrapper by accident.

Yes, we'll need to do that sooner or later, though accidents of that sort aren't to worrying as they're caught at link time.

jralls: > Interesting. This means that Split should have been actually named GncSplit?

Well, renamed that way.

> And if I understand you correctly, the prefix should be omitted for the mock classes, since it's C++.

It would be more about the mock classes being for testing and therefore not exported out to where namespaces matter. Given that C++ has real namespaces and that we ultimately want to remove GObject from everywhere that doesn't really need it, we'd end up with something like `Gnc::Engine::Split` and `Gnc::Engine::Mock::Split`. Once Split is a C++ class we'll also be able to dispense with all of this GObject runtime crap. The compiler will make sure that we get a Split(-subclass) in a function that takes one.

>  But this means, the lower case type name of MockSplit should be changed to mock_split instead of gnc_mock_split, right?

Depends. If you're following GObject rules the type name would be `GncMocksplit` and functions would be named `gnc_mocksplit_foo`. There are two reasons to follow those rules: One is that you're using gobject-introspection, which needs things formatted that way in order to create the object->function names for languages that work that way. We don't, so that's out. The other would be that you're using GObject's built in lifetime management: The macros that manage that stuff also need names that follow those rules, but not quite as strictly. The original X-accounting classes don't use those macros and so don't follow the rules, the Qof and business classes do use the macros and so they have names like QofInstance and GncInvoice.

There's no point in any `gnc_mock_foo_` function of any kind, it's never called. The mocks have to provide all of the `gnc_foo_` functions that are actually used in the code being tested so that it will link. Those can be implemented in C++, just as you already did with the mocked functions. The only reason to provide for the lifecycle functions (`gnc_foo_init`, `gnc_foo_class_init`, `gnc_foo_dispose`, and `gnc_foo_finalize`) is if the function under test creates one. If you're going to create the object in `SetUp` and destroy it in `TearDown` then it can be a pure C++ object.
Things get a bit more interesting when one wants to test QofInstance behavior (e.g. `qof_instance_get(QOF_INSTANCE(split))` )  through one of the classes, but let's leave that for later work.

gruberchr: I did part of the suggested changes now.

Next I wanted to implement the following suggestion:
> Runtime checks of other classes in mock code should check for the base-class type, not for its mock-class if it has one. Mocks shouldn't assume that they'll be used only with the other mocks.

This led me to the problem, that checking for the base-class type, for instance `GNC_TYPE_ACCOUNT`, requires the base-class to be always linked to the test application. In this example the corresponding function `gnc_account_get_type()` is defined in *Account.cpp*, which has be always linked to the application in this case. But don't we run again into the same problem, we discussed in PR #650 already? We wanted to completely **replace** `libgncmod-engine.so` by mockups to not get in conflict with doubled defined symbols. Linking *Account.cpp* contradicts to that decision. By the way, that decision is the reason why `GNC_TYPE_MOCKACCOUNT` is not derived from `GNC_TYPE_ACCOUNT` but replaces it instead.

On the other hand you furthermore suggested

> `foo_get_type` needs to be faked to return `mock_foo_get_type()` so that the runtime GType checking in the code under test succeeds.

Isn't that the same as using `GNC_IS_MOCKACCOUNT()` instead of `GNC_IS_ACCOUNT()`?

jralls: > This led me to the problem, that checking for the base-class type, for instance GNC_TYPE_ACCOUNT, requires the base-class to be always linked to the test application.

You missed from [gmopck-tryout](https://github.com/jralls/gnucash/commit/088a53ea6d37ed884408822b930384dbc38b392b)
```C++


GType gnc_split_get_type(void)
{
    return gnc_mock_split_get_type();
}
```

jralls: > Isn't that the same as using GNC_IS_MOCKACCOUNT() instead of GNC_IS_ACCOUNT()?

Nope. With `gnc_split_get_type` returning `gnc_mocksplit_get_type` all of the type macros become interchangeable. That's important if you're going to pass a MockSplit to a C function taking a Split.

gruberchr: Fakes for `_get_type()` functions added and runtime type checks changed.

The last issue, adding `extern "C"` for C linkage, still confuses me. I'll update that later.

gruberchr: Ok, I now also added missing C linkage statements to the best of my knowledge and belief.

But there are differences to your `gmock-tryout` branch:
- I didn't wrap *gmock-Split.cpp* and only wrapped `#include <SplitP.h>` in *gmock-Split.h*, since *Split.h* already contains `extern "C"`
- I didn't wrap *gmock-qofbook.cpp*, since *qofbook.h* and *qofbook-p.h* already contain `extern "C"`

jralls (commented):
Commit 5408854c is harmless but unnecessary. We're not going to use gobject-introspection or gtk-doc so there's no need to comply with Gnome's naming patterns.

jralls:
The commit message on 0e634c5 is misleading. It isn't possible to combine real and mock objects in the same program because both the mocks and the real object declare the same C functions and C functions can't be overloaded. What it should say is that code being tested is going to call the "real" run-time type check and when it has been fed a mock that type check needs to check against the mock's GType instead of the real one.

gruberchr:
> The commit message on [0e634c5](https://github.com/Gnucash/gnucash/commit/0e634c54460463de2c8bc1204835041d03a014e0) is misleading. It isn't possible to combine real and mock objects in the same program because both the mocks and the real object declare the same C functions and C functions can't be overloaded. What it should say is that code being tested is going to call the "real" run-time type check and when it has been fed a mock that type check needs to check against the mock's GType instead of the real one.

Sorry, I don't know how to write down this precisely. "It should be possible to use mockup objects and non-mockup objects within the same application." is not meant for objects of the same type. I wanted to say, that you can use mockup accounts, but real transactions for instance.
Please feel free to rewrite my commit message.

libgnucash/engine/mocks/gmock-Transaction.cpp
```diff
@@ -29,121 +29,105 @@ void
...
-        return nullptr;
+    SCOPED_TRACE("");
+    const MockTransaction* mocktrans = gnc_mocktransaction(trans);
+    return mocktrans ? mocktrans->get_notes() : "";
 }
 
 void
 xaccTransSetNotes (Transaction *trans, const char *notes)
 {
     ASSERT_TRUE(GNC_IS_MOCKTRANSACTION(trans));
```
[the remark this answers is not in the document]
gruberchr: No, this was intended. I left `ASSERT_TRUE` in all void functions, since it seems to simplify things. The alternative would be four code lines instead of two:
```cpp
SCOPED_TRACE("");
MockTransaction* mocktrans = gnc_mocktransaction(trans);
if (mocktrans)
    mocktrans->set_description(notes);
```

What do you think? If you like I can change it.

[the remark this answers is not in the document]
gruberchr: Oops, there seems to be a copy&paste error in the code. Of course function `xaccTransSetNotes()` has to call mock function `mocktrans->set_notes()`. I'll correct this, too.

libgnucash/engine/mocks/fake-qofquery.cpp
```diff
@@ -12,11 +16,14 @@
 static class QofFakeQueryPool
 {
 public:
+
+    /* Add QofFakeQuery object to the pool */
```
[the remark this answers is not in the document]
gruberchr: This is a static class, which is not part of the doxygen documentation. This is therefore no doxygen comment.

libgnucash/engine/mocks/gmock-Split.h
```diff
@@ -82,4 +82,24 @@ class MockSplit : public Split
...
+gnc_mocksplit (Split *split)
+{
+    if (GNC_IS_MOCKSPLIT(split))
+        return static_cast<MockSplit*>(split);
+    ADD_FAILURE() << "Expected 'split' to be of type 'MockSplit'";
+    return nullptr;
+}
+
+static inline const MockSplit*
+gnc_mocksplit (const Split *split)
```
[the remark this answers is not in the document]
gruberchr: Makes sense. Is this the correct const implementation?

```cpp
static inline const MockSplit*
gnc_mocksplit (const Split *split)
{
    const MockSplit* mocksplit = gnc_mocksplit(split);
    return mocksplit;
}

File: libgnucash/engine/mocks/gmock-Account.cpp [hunk 13]
```diff
@@ -27,45 +27,39 @@ void
...
 QofBook *
 gnc_account_get_book(const Account *account)
 {
-    EXPECT_TRUE(GNC_IS_MOCKACCOUNT(account));
-    if (GNC_IS_MOCKACCOUNT(account))
-        return ((MockAccount*)account)->get_book();
-    else
-        return nullptr;
+    SCOPED_TRACE("");
+    const MockAccount* mockaccount = gnc_mockaccount(account);
```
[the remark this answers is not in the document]
gruberchr: Done

libgnucash/engine/mocks/gmock-Transaction.cpp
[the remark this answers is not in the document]
jralls: > since it seems to simplify things.

OK.

jralls:
I've resolved the conflict, fixed one omission (you hadn't mocked MockTransaction::get_list_split), and merged the result. Looks like my changes were enough to confuse github so that it didn't recognize the merge.

Thanks!

gruberchr:
Great!

PR #739, opened 2020-06-17
Add missing #include <algorithm> into gnc-imp-settings-csv-price.cpp

fellen:
Found in failed win maint nightlies.

Suggesstion by shaggy.

I am not sure about [un]conditional include.

[Edit] [IRC log](https://lists.gnucash.org/logs/2020/06/16.html#T20:52:51)


jralls:
Oh, it should be unconditional. You can also add
```C++
#include <string>
#include <vector>
#include <memory>
```
Because `gnc-imp-settings-csv-price.cpp` is depending on other includes to bring those in and that could change in the future and break the build.

PR #740, opened 2020-06-20
commodity-collector must not round

christopherlam:
ensure the commodity-collector adds exact amounts instead of rounding during accumulation.


christopherlam:
Not crucial during regular additions but may make price calculations more accurate in commodity-utilities.

PR #741, opened 2020-06-21
I18n: Fix grossly wrong wording

cstim:
Editing the associations must be called editing, not "update" or "change".

The wording in the code is grossly misleading. I understand we are currently in string freeze, but this change needs to go in as soon as string freeze is lifted again. In creating the de.po translation, I had to translate as if the English texts already say "edit", otherwise the texts would be non-comprehensible.

### DIFF
--- a/gnucash/gnome/dialog-assoc.c
+++ b/gnucash/gnome/dialog-assoc.c
@@ -537,7 +537,7 @@ row_selected_bus_cb (GtkTreeView *view, GtkTreePath *path,
             return;
         }
 
-        ret_uri = gnc_assoc_get_uri_dialog (GTK_WINDOW(assoc_dialog->window), _("Change a Business Association"), uri);
+        ret_uri = gnc_assoc_get_uri_dialog (GTK_WINDOW(assoc_dialog->window), _("Edit a Business Association"), uri);
 
         if (ret_uri && g_strcmp0 (uri, ret_uri) != 0)
         {
@@ -628,7 +628,7 @@ row_selected_trans_cb (GtkTreeView *view, GtkTreePath *path,
             g_free (uri);
             return;
         }
-        ret_uri = gnc_assoc_get_uri_dialog (GTK_WINDOW(assoc_dialog->window), _("Change a Transaction Association"), uri);
+        ret_uri = gnc_assoc_get_uri_dialog (GTK_WINDOW(assoc_dialog->window), _("Edit a Transaction Association"), uri);
 
         if (ret_uri && g_strcmp0 (uri, ret_uri) != 0)
         {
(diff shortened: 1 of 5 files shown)


jralls:
> I had to translate as if the English texts already say "edit", otherwise the texts would be non-comprehensible.

Why? A native English speaker has no difficulty recognizing the equivalence of "change", "update", and "edit". If German (or Chinese,Tamil, or whatever) don't share the equivalence then it's up to the translator to write an idiomatic sentence in that language.

What's more, "edit" implies changing the words in a string and that would be poor usage if, as in the case of the file association, the selection is made using the file chooser. For that case "change" or "update" is much better English usage.

cstim:
In the UI the point where the menu items appear is clearly where the user expects to "edit" stuff. It is common convention to say we "edit" there. Not change, update, modify, work on, handle, treat, process, or whatever other wording authors can come up with. This particular strings are the only exception from the convention which is otherwise fortunately strictly followed within gnucash.

cstim:
Even worse, the UI item for *adding* the very first file (or URL) association is currently the one labeled "update association". But the user's expectation for "update" is to change something that's already existing, so this is completely non-suitable. "editing" isn't fully matching, too, but at least it points better to the right direction, because I would expect to edit something so that it will be there subsequently.

I'm just saying this particular area of gnucash is a stark exception from the otherwise quite coherent UI wording, and it needs to be reworked from a UI wording point of view as soon as the i18n string freeze is over. The attached patch is the first step of this.

jralls:
> This particular strings are the only exception from the convention which is otherwise fortunately strictly followed within gnucash.

Not quite true:
```
grep -r 'name="label".*>[Cc]hange' gnucash/gtkbuilder/ | wc -l
       5
grep -r 'name="label".*>[Ee]dit' gnucash/gtkbuilder/ | wc -l
       0
grep -r 'name="label".*>[Uu]pdate' gnucash/gtkbuilder/ | wc -l
       0
grep -r '_("_\?[Cc]hange' gnucash | wc -l
      12
grep -r '_("_\?[Ee]dit' gnucash | wc -l
      62
grep -r '_("_\?[Uu]pdate' gnucash | wc -l
      13
```

> the UI item for adding the very first file (or URL) association is currently the one labeled "update association". But the user's expectation for "update" is to change something that's already existing, so this is completely non-suitable. "editing" isn't fully matching, too, but at least it points better to the right direction, because I would expect to edit something so that it will be there subsequently.

That I partly agree with, though I'd prefer "set" to either "update" or "edit".

fellen:
Before it gets forgotten, there was a discussion today:
https://lists.gnucash.org/logs/2020/07/07.html#T15:42:30

PR #743, opened 2020-06-23
Bug 797819 - When upgrading open register widths lost

Bob-IT:
I was going to push the fix but then decided to created this PR so it can be decided if you want me to revert this change for account registers and keep for Business or something else.

What would happen with this change, on first load of open registers, the page state section will not have width information so the register state section will be used. In the new format the width information is also stored with the page state information but that may be reverted depending on what is required to do.

On first load of a closed register,  a check is made to see if there is user default state for that register group type, if so that will be used for widths. If there is no default state, then the old register state information is used.

Once the register is loaded, if there was no user default state for that register group type, the current register state is saved to the register group state and hence used for any other register of that group type. Once this is done the old register state is removed as there is no longer a need to save it all the time.

As I see it, if you want to keep individual register state also so...
For already open tabs, use the page state as that also has the width information but as stated above that may be reverted.
For opening register pages, use register state if present, then register group state if present and then finally calculated widths, the last two would only seed the register state widths.

This could probably be achieved with a slight change of code and not deleting the register state, so setting the register group state widths would only be used for new registers layouts to seed there own register state information.

Part of this was an ability to reset the register width state, this could be added at a later date with another menu option that reset the widths to what ever the register group widths are.

@jralls 
@gjanssens 
I hope this makes sense, I probably have enough time to make the changes so please advise.

### DIFF
--- a/gnucash/gnome/gnc-split-reg.c
+++ b/gnucash/gnome/gnc-split-reg.c
@@ -555,10 +555,15 @@ gsr_create_table( GNCSplitReg *gsr )
     // see if register group has the date_width key, old format pre 4.0
     has_date_width = g_key_file_has_key (state_file, register_state_section, "date_width", NULL);
 
-    // if this is from a page recreate and no register state use those settings,
+    // if this is from a page recreate and no date_width use register state for settings,
     // register state width information is dropped at the end of function.
-    if (gsr->page_state_name && !has_date_width)
-        group = gsr->page_state_name;
+    if (gsr->page_state_name)
+    {
+        if (g_key_file_has_key (state_file, gsr->page_state_name, "date_width", NULL))
+            group = gsr->page_state_name;  // new format with widths
+        else
+            group = register_state_section;
+    }
     else
     {
         // if no default state, use register state if available


jralls:
Confirmed that the commits so far fix the [open-registers bug](https://bugs.gnucash.org/show_bug.cgi?id=797819).

>For opening register pages, use register state if present, then register group state if present and then finally calculated widths, the last two would only seed the register state widths.

>This could probably be achieved with a slight change of code and not deleting the register state, so setting the register group state widths would only be used for new registers layouts to seed there own register state information.

That would at least prevent the loss of users' carefully created per-register layouts. That will greatly reduce the howling, so please do that. With that "slight change" would subsequent changes to the layout be saved or would that register be frozen at the 3.x state unless the user leaves the tab open?  If the latter, how hard would it be to restore the saving of changes to individual registers not left open?

>Part of this was an ability to reset the register width state, this could be added at a later date with another menu option that reset the widths to what ever the register group widths are.

I take it that you saw [my IRC comment](https://code.gnucash.org/logs/2020/06/22.html#T12:46:11) and the [ensuing discussion](https://code.gnucash.org/logs/2020/06/22.html#T14:02:53) with @gjanssens. Might as well get his [email](https://lists.gnucash.org/pipermail/gnucash-devel/2020-June/045175.html) [links](https://lists.gnucash.org/pipermail/gnucash-devel/2020-June/045182.html) into the record here too. Yes, I think a menu item to reset the current register to the type group default (effectively to discard the per-account customizations) would be useful, and perhaps one to do so for all registers of a type group as well.

And while we're at it, one more thing: Is the Windows menu really the right place for this or would the View menu be more appropriate?  [Gnome's HIG](https://developer.gnome.org/hig/stable/menu-bars.html.en) suggest that the Format menu would be the right place, but GnuCash hasn't used one so far and [Apple's](https://developer.apple.com/design/human-interface-guidelines/macos/menus/menu-bar-menus/#window-menu) say "The Window menu lets people navigate, organize, and manage an app’s windows. This menu doesn’t include menu items for customizing the appearance of windows. Those functions are provided by the View menu." [Microsoft](https://docs.microsoft.com/en-us/windows/win32/uxguide/cmd-menus) doesn't provide guidance about the uses of particular main menu bar items.

Bob-IT:
> That would at least prevent the loss of users' carefully created per-register layouts. That will greatly reduce the howling, so please do that. With that "slight change" would subsequent changes to the layout be saved or would that register be frozen at the 3.x state unless the user leaves the tab open? If the latter, how hard would it be to restore the saving of changes to individual registers not left open?

I think I would just restore the existing setup, each register would save on close, the register group would only be used to seed a new register layout which would then in it's self be saved on close just like it is now. Changes to the register group can be made at any time but would only effect opening any new registers with no state. If I remember correctly I started off with them in the view menu so I can move them there, starting of with the existing two and move them to a sub menu when adding the following two after 4.0

With the addition of possibly two further menu options, one to reset an individual register to that of the register group and one possibly to reset all registers of a particular group I think would cover all possibilities, the registers affected will still have there own state but would be copied from the register group and as before could be further changed. I think these changes could be added after 4.0.

jralls:
That seems reasonable. I agree that it's too late to add the menu items for 4.0, we can put them in 4.1.

@gjanssens, any comments?

jralls:
Another thought, from the "least surprise" department: Would it be more surprising if the customized layout for an account stay attached to that account or be removed when one uses **Use as Default Layout for this Register Group**? What about **Restore Default Layout for this Register Group**: Should that retain the layout on the currently focussed register (and save it in the state file) or should it immediately reset that layout to the calculated value? I recognize that those are questions with which @gjanssens wrestled and that led him to decide that maintaining both group and individual settings was too hard. Certainly letting the user choose is too hard, and if coding one way or the other adds a lot of complexity let's go with the simpler solution and document it--and I'm working on documenting column widths now.

gjanssens:
> Another thought, from the "least surprise" department: Would it be more surprising if the customized layout for an account stay attached to that account or be removed when one uses **Use as Default Layout for this Register Group**? What about **Restore Default Layout for this Register Group**: Should that retain the layout on the currently focussed register (and save it in the state file) or should it immediately reset that layout to the calculated value?

The honest answer is it depends on the user's intention which we can't guess. And probably several users wouldn't grasp the full effect of either choice if asked to make it. So I do agree with your later remark letting the user choose is too hard.

When considering design I always try to keep this basic rule in mind: gnucash should help the user focus on effective accounting. Any time spent not on accounting is time lost. In this particular case there were several complaints about how cumbersome it is to have to change layouts on each and every register. The design goal was to reduce that overhead.

Will we really achieve this adding a middle level in addition to the calculated and individually customizable layouts ? I agree the user will have complete freedom that way, but will it reduce the overhead ? I doubt it. On the contrary the whole layout system becomes fairly complex. Making layout changes now suddenly requires careful consideration by the user. Is this change meant for this register only or do you want to apply it to the full register class ? That requires even more mental energy than the 3.x behaviour where it was just a matter of changing what you see (with the big disadvantage there you had to repeat this for every register).

I'll admit this extra consideration already started with my two level save state design. It compensated the partly by eliminating the per register save state. One level less to worry about.

To be honest at this point I would prefer to revert the per register class save state and just go back to what it was in 3.x (with the addition of save states for invoices and bills, which is what started the whole process). The reason is I believe doing this right requires another redesign which may no longer need that middle layer. However if we add it now and pull it once it has been in the wild for a while we will again receive plenty of complaints.

If that's no longer an option then yes, the added menu options will help at least restore the freedom lost.

I also agree the View menu is probably a better location to keep those. I don't remember why I suggested the `Windows` menu.

As to what to do with individual register layouts when setting or restoring a default layout. That really is a hard one. Coding wise it's easiest just to set/restore the default register class state and leave individual register states alone. But to my mind that would result in a very annoying scenario for users that have been tweaking registers for say a data column that's consistently too narrow. At some point the figure out there's a way to set the default and will do so. But the ones that were manually set won't change with it. Or imagine (as also mentioned in my mails) on some registers other columns were changed, like the description or the transfer column, but not yet the date column. Now open another register in the same class that was never changed. Use that one to fix the date column width and save the default. The registers that had their description or transfer width already set will still open with a too narrow date column because they already had an individual layout.

On the other hand updating all individual layouts when setting/resetting the default would frustrate the other camp. Take the last example. If we reset individual layouts when the user sets the default with a changed date column, suddenly the carefully set description and transfer columns are lost.

Which use case is most frequent ? I don't know.

gjanssens:
> I was going to push the fix but then decided to created this PR so it can be decided if you want me to revert this change for account registers and keep for Business or something else.
> 
> What would happen with this change, on first load of open registers, the page state section will not have width information so the register state section will be used. In the new format the width information is also stored with the page state information but that may be reverted depending on what is required to do.
> 
> On first load of a closed register, a check is made to see if there is user default state for that register group type, if so that will be used for widths. If there is no default state, then the old register state information is used.
> 
> Once the register is loaded, if there was no user default state for that register group type, the current register state is saved to the register group state and hence used for any other register of that group type. Once this is done the old register state is removed as there is no longer a need to save it all the time.
> 
> As I see it, if you want to keep individual register state also so...
> For already open tabs, use the page state as that also has the width information but as stated above that may be reverted.
> For opening register pages, use register state if present, then register group state if present and then finally calculated widths, the last two would only seed the register state widths.

What would happen if the user changes the register state and closes the register again ? Will you save that widths with the register state ?

That then would result in 4 levels of state:

* calculated
* register class default
* register individual (in register state)
* open register individual (in page state)

In that case I agree the open register individual state is redundant even though page state intuitively is the more logical place to store layout related info.

Bob-IT:
> What would happen if the user changes the register state and closes the register again ? Will you save that widths with the register state ?

Basically yes, register page width state would be dropped, each register would save its own width state so it would be exactly the same as in 3.x. The only difference would be for opening new registers, instead of going straight to calculated it would first look to see if there was a register class default and use that if present. That could be achieved with the current two menu options but may need some wording changes.

It is surprising to me that what looked like a simple change for the better causes so much trouble.

@jralls 
@gjanssens 
Maybe we just drop all this for the register setup and keep the Business changes but move the menu options to the 'View' menu.

gjanssens:
The recent discussion on gnucash-devel with Adrien only pushes me more in the direction of reverting this change for 4.0. It's insufficiently polished and as I said before pushing what we have now will make it harder to redesign.

@Bob-IT let's wait for @jralls to voice in as well, but if we revert would it be hard to rewrite the invoice/bill saving methods to behave like account saves ? That is, that you store individual invoice/bill column widths like we do for account registers ?

Bob-IT:
@gjanssens  I think it may of started out like that and thought it would all be in the PR #648 but I suppose a force push clears that.
To much to try and replicate from scratch now so happy to dump the lot and start again once a definite requirement is ironed out. I will just use the PR and reverse most of the changes.

gjanssens:
Yes, you did start out like that. I seem to remember one of us  was worried the list of entries in the key file would become too long - there are plenty more invoices and bills over time than there will be accounts.
So adding just that may also be the precursor to a painful fix in the future (I don't think key files are optimized for large data quantities).
Perhaps just fully reverting the PR is the best option. IIRC there were a few bugs closed as part of merging the PR. We'll have to reopen those as well.

By the way, I think @jralls will only reply late today or tomorrow.

jralls:
Is it really the case that the user complaints are about registers embedded in business objects, e.g. "Why does the stupid date field go back to the wrong size every time I create a new invoice?" and not account registers? In that case I agree that it's better to just address that use case unless and until we can create a more general facility that works for both the "I want a custom default" and "I want to customize individual accounts differently" camps.

jralls:
The discussion with Adrian is about hiding columns, right? That's rather orthogonal to what we've been talking about here: He actually was trying to use it for business use. His column-width issue isn't Mac-only, is it? Is it so common a use case that it means that the whole feature should be tossed?

Bob-IT:
> The discussion with Adrian is about hiding columns, right? That's rather orthogonal to what we've been talking about here: He actually was trying to use it for business use. His column-width issue isn't Mac-only, is it? Is it so common a use case that it means that the whole feature should be tossed?

Yes, Adrian was trying to hide columns which I never thought of, the saving code in table-gnome.c does not save 0 width columns so on the next load the column width is back to calculated default. This I think could be fixed by adding these lines to gnucash-header.c, line 477

            if (header->resize_col_width == 0)
                header->resize_col_width = 1;

So when you drag the right column border to the left border and release you end up with a thicker border, effectively the column width is 1px wide and this does save. It also indicates that you have hidden columns so that may be a good thing.

So like I said, maybe we just ditch the account register changes and have a rethink but keep the business changes with a move of the menu options to View.

gjanssens:
> Is it really the case that the user complaints are about registers embedded in business objects, e.g. "Why does the stupid date field go back to the wrong size every time I create a new invoice?" and not account registers?

I don't know that for sure. Looking back at @Bob-IT 's initial PR, he didn't mention this. It had state file fixes and added saving layouts for business objects with no reference to any bug or feature request. The extension to all account registers came on my request. I was searching for a solution consistent across all registers. And while doing so I remembered the complaints I had seen in the 3.x series where users were forced to fix each and every register due to poorly calculated defaults.

Searching bugzilla there are open bugs in all directions:
* [one particularly pointing at the inability to save invoice and bill column widths](https://bugs.gnucash.org/show_bug.cgi?id=456461)
* [general complaint of non-intuitive behaviour](https://bugs.gnucash.org/show_bug.cgi?id=563588)
* [wish to remove invoice columns](https://bugs.gnucash.org/show_bug.cgi?id=455801)
* [business column resizing complaint](https://bugs.gnucash.org/show_bug.cgi?id=796954). Note this one also complains about a missing horizontal scroll bar, but that's tangential to this PR.
* [balance column sizing failure on macos](https://bugs.gnucash.org/show_bug.cgi?id=751873)

* [(Closed) blast from the past](https://bugs.gnucash.org/show_bug.cgi?id=666576). I'm adding this here because it shows my past self actively pushing for a per account storage of column widths....

The list is not exhaustive in any way and doesn't take into account all of the time spent on the mailing lists explaining to people how it currently works. Plenty of first-time users get lost in the current mechanism, only a few go as far as to file it as a bug.

At this point I think the quickest win we can achieve is to drop the Description-Sizes-To-Remaining-Width behaviour and just let the user freely decide column widths for all columns even if that means making the total page width wider or narrower than the available window width.

> In that case I agree that it's better to just address that use case unless and until we can create a more general facility that works for both the "I want a custom default" and "I want to customize individual accounts differently" camps.

The 3.x design favours the right hand side camp, the current 3.90x design adds a new tool for the left hand side camp but neglects the right hand side camp. The proposal to add a reset option enables all requests at the cost of added complexity. I don't know how to express this properly, but my fundamental question is whether the added complexity and effort is worth the added benefit. To my mind it is not as I expect this to
* distract users from doing actual accounting. There has always been some layout management overhead due to poorly chosen default widths and the uncommon Description field behaviour. The overhead so far was only linked to what you saw on screen directly. This layout management overhead now takes another level because you have to decide whether your changes only apply to what you see on screen or also to plenty of other objects you currently don't see.
* distract developers from working on actual accounting improvements. I expect plenty of support requests on the lists. Many of those will be because it's just new. Those are normal for any change. However those aside there will also be many with corner cases we didn't think of and need tweaking, refining, adding exceptions,...

So yes, I'm all for withdrawing it for now and reconsider the fundamental design. I don't want further design iterations to be made more difficult by newly formed user habits based on the 3.90x half-way design.

gjanssens:
> > The discussion with Adrian is about hiding columns, right? That's rather orthogonal to what we've been talking about here: He actually was trying to use it for business use. His column-width issue isn't Mac-only, is it? Is it so common a use case that it means that the whole feature should be tossed?
> 
> Yes, Adrian was trying to hide columns which I never thought of, 

A nice example of what I mean with corner cases we didn't consider (not in any way meant as a criticism to you @Bob-IT ). And this is an easy one.

the saving code in table-gnome.c does not save 0 width columns so on the next load the column width is back to calculated default. This I think could be fixed by adding these lines to gnucash-header.c, line 477
> 
> ```
>         if (header->resize_col_width == 0)
>             header->resize_col_width = 1;
> ```
> 
It's a fair short term fix for this. And you can add this if we continue on this design.

More generally though it's a code design issue that we depend on a 0 width metric to 'hide' columns to differentiate layouts. I get that this stems from our custom-built register code. And that in itself is a huge effort drain that we should eventually fix by replacing it with standard gtk widgets. The countless hours @jralls has spent so far fixing proper input method support alone is disheartening. Input method support is a solved problem in gtk widgets. We should not have had to waste that much time on it. But I digress.

> So when you drag the right column border to the left border and release you end up with a thicker border, effectively the column width is 1px wide and this does save. It also indicates that you have hidden columns so that may be a good thing.

In the current state of affairs that's the best we can do. A bigger effort would be do model the registers as we do the account hierarchy and have a drop-down menu with the available columns and let users select the ones to show or hide. Depending on register class we need to define a few mandatory columns of course.

> So like I said, maybe we just ditch the account register changes and have a rethink but keep the business changes with a move of the menu options to View.

Ok with me. We just need to consider in which form the business changes should go in. I acknowledge invoice/bills are not the same as accounts. We start with an account hierarchy with many accounts, but over time this list changes only very little. The list of invoices/bills on the other hand will only grow over time, at some point far outnumbering the number of accounts. I'll repeat my concern about hogging the state file with layout information (as @Bob-IT also expressed in the original PR by the way). So perhaps for invoices and bills we should go with the default layouts method as implemented so far and not set individual invoice layouts. At first sight that may break consistency. On the other hand accounts and business documents are different objects which may be a reasonable motivation to treat their layout constraints differently.

jralls:
> distract users from doing actual accounting. 

A complaint about using computers for office work that goes back to the original word processors in the 1980s when people started "wasting time" on the appearance of documents instead of the content. That battle is long since lost. Users expect to be able to adjust the presentation to what they perceive as their needs.

It's wrong for another way, too: A well designed UI should help users to focus on the accounting by presenting them with the information they need in an easily understood way. The register has a bad record of requiring user intervention to do that, the Date, Transfer Account, and Balance columns being the worst offenders. 

> distract developers from working on actual accounting improvements

We developers are responsible for the whole package, not just the accounting parts. User experience is hugely important to a successful application. The fact that we're having this discussion at all reflects that we understand that. It's pretty clear from the frequency resizing columns comes up on the user list and IRC that it's a serious wart and that we need to fix it. It's too late now to do that properly for 4.0 so we need to decide what to do in the next 48 hours or so to not have too much of a mess.

So:

@Bob-IT's 2-line patch to unbreak saving hidden columns is a no-brainer. Let's do it.

At this point I think that reverting the general register changes and leaving the customized default for business objects is also the right way forward. IIUC we have not in the past recorded custom column widths on those objects, so we're not taking away any features, we're just fixing something.

Beyond that I agree that we need to rethink the register, but this isn't the right place for that. Let's have that discussion on gnucash-devel.

Bob-IT:
OK, I will make the changes Friday afternoon, busy in morning.

Bob-IT:
I have pushed the changes to master that should cover what has been discussed here.

jralls:
@Bob-IT The commits are made and this can be closed, right?

Bob-IT:
Yes, closing as required commits have been made, further work will be done on a new PR.

PR #744, opened 2020-06-25
i18n: Fix missing context in single-character translated strings

cstim:
The single-character column heading "C" is used for "Color" in one place, and for "Cleared" in another place. Obviously this must be fixed by adding context, otherwise grossly wrong translations will show up (as is currently the case for German and the "Cleared" column in the import matcher, see screenshot)


cstim:
grafik

fellen:
Good catch!
I ammended it to use the same context format as we used in other contexts for easier grepping.
Commit c69f0eb 

Thanks!

PR #745, opened 2020-06-28
Bug 787295 - Allow UI jump from Business accounts to their relevant invoice/bill/voucher

christopherlam:
Pros: shows all invoices associated with txn.

Cons: if txns has both invoice-payment splits (have lots) and overpayment (has no lot), and the APAR overpayment split is selected, it'll prompt user to select from the invoices.

Consider a payment which pays 2 invoices for $225 and $120, and also an overpayment of $10. It'll have 4 splits:
- AP $225
- AP $120
- AP $10
- Bank -$355

We'd expect AP splits for $225 and $120 to immediately open the associated invoice, and the bank -$355 to offer selection. This works. However AP $10 split->invoice will offer selection as well, which is not necessarily intuitive. It may be considered "overengineering" to prevent this.

### DIFF
--- a/gnucash/gnome/gnc-plugin-page-register.c
+++ b/gnucash/gnome/gnc-plugin-page-register.c
@@ -76,6 +76,7 @@
 #include "gnc-window.h"
 #include "gnc-main-window.h"
 #include "gnc-session.h"
+#include "gnc-ui.h"
 #include "gnc-warnings.h"
 #include "gnucash-sheet.h"
 #include "dialog-lot-viewer.h"
@@ -278,6 +279,7 @@ static void gnc_plugin_page_register_event_handler (QofInstance* entity,
                                                     GncEventData* ed);
 
 static GncInvoice* invoice_from_split (Split* split);
+static GList* invoices_from_transaction (Transaction* trans);
 
 /************************************************************/
 /*                          Actions                         */
@@ -1064,7 +1066,7 @@ gnc_plugin_page_register_ui_update (gpointer various,
     GtkAction* action;
     gboolean expanded, voided, read_only = FALSE;
     Transaction* trans;
-    GncInvoice* inv;
+    GList* invoices;
     CursorClass cursor_class;
     const char* uri;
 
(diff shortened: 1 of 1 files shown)


christopherlam:
image

christopherlam:
@gjanssens any opinion on this?

Simply upgrades to allow jump from any payment split, including the Asset/Liability account. Small issue - this also allows jump from APAR overpayment split, and will pop up invoice selection.

christopherlam:
This probably does it better.
Widget sensitivity depends on whether the txn has splits with invoices.
Jump action will first try split's invoice; if this fails will try split->parent->invoices; if only one then open it, otherwise ask user to select.

gnucash/gnome/gnc-plugin-page-register.c
```diff
@@ -4603,10 +4609,42 @@ gnc_plugin_page_register_cmd_jump_associated_invoice (GtkAction* action,
...
+                GncInvoice* inv = (GncInvoice*) node->data;
+                amt = xaccPrintAmount
+                    (gncInvoiceGetTotal (inv),
+                     gnc_account_print_info (gncInvoiceGetPostedAcc(inv), TRUE));
+                details = g_list_prepend
+                    (details,
+                     g_strdup_printf ("Date=%s, ID=%s, Amount=%s",
+                                      qof_print_date (gncInvoiceGetDatePosted (inv)),
+                                      gncInvoiceGetID (inv), amt));
+            }
```
christopherlam, line 4636: This string must be translatable. Not sure the best string to describe an invoice.

jralls: The description on the corresponding `GncEntry`?

gjanssens: There is not necessarily a direct mapping between a GncEntry and a transaction split. Invoices can have multiple entries, which could be merged together or not when creating the invoice transaction.

IMO an invoice is well defined by a combination of date, owner, id and possibly total amount.

jralls: So the dialog box should present Owner name, Invoice Number, Date, and Amount? How about whether it's a bill, invoice, or voucher as well? The user will need as much context as possible to work out which one they want to jump to.

christopherlam: Candidate invoice description string in latest commit:

image

gjanssens: > So the dialog box should present Owner name, Invoice Number, Date, and Amount? How about whether it's a bill, invoice, or voucher as well? The user will need as much context as possible to work out which one they want to jump to.

Good point. Add to that it can also be credit notes.

So any transaction can have a combination of these documented associated with it:
- bills and vendor credit notes
- invoices and customer credit notes
- employee vouchers and credit notes

gjanssens: > Candidate invoice description string in latest commit:
> 
> image

Looks good, except "Several invoices..." is incomplete as per my previous comment. It may be explicitly expanded to
"Several invoices and/or credit notes..." but I find that cumbersome.

Perhaps we should introduce a more generic term for these, like "customer documents", "vendor documents", "employee documents". That would still hold if someone in the future for example also implements pro-forma invoices or sales orders,...

christopherlam: > So any transaction can have a combination of these documented associated with it:
> 
>     * bills and vendor credit notes
>     * invoices and customer credit notes
>     * employee vouchers and credit notes

`gncInvoiceGetTypeString` can return "Invoice" "Bill" "Expense" or "Credit Note". I _think_ it's adequate.

> Perhaps we should introduce a more generic term for these, like "customer documents", "vendor documents", "employee documents".

"business object"? "document"? Then the menu item in use "Open Associated Invoice" must also be modified.

christopherlam: Resolved by renaming `invoice` to `document` in new strings. Existing menu-item `"Open Associated Invoice"` is unchanged.

PR #746, opened 2020-06-29
Bug 797825 - OFX import crashes on import of investment transaction

jeanlaroche:
The crash is caused by an uninitialized pointer in ofx_info.
This fixed it.

### DIFF
--- a/gnucash/import-export/ofx/gnc-ofx-import.c
+++ b/gnucash/import-export/ofx/gnc-ofx-import.c
@@ -1252,6 +1252,10 @@ void gnc_file_ofx_import (GtkWindow *parent)
         info = g_new(ofx_info,1);
         info->num_trans_processed = 0;
         info->statement = NULL;
+        info->last_investment_account = NULL;
+        info->last_import_account = NULL;
+        info->last_income_account = NULL;
+        info->created_commodites = NULL;
         info->parent = parent;
         info->run_reconcile = FALSE;
         info->file_list = selected_filenames;
--- a/gnucash/import-export/ofx/gnc-ofx-import.c
+++ b/gnucash/import-export/ofx/gnc-ofx-import.c
@@ -75,11 +75,8 @@ typedef struct _ofx_info
     gint num_trans_processed;               // Number of transactions processed
     struct OfxStatementData* statement;     // Statement, if any
     gboolean run_reconcile;                 // If TRUE the reconcile window is opened after matching.
-    GList *created_commodites;
     GSList* file_list;                      // List of OFX files to import
 } ofx_info ;
-
-GList *ofx_created_commodites = NULL;
 
 /*
 int ofx_proc_status_cb(struct OfxStatusData data)
@@ -290,8 +287,6 @@ int ofx_proc_security_cb(const struct OfxSecurityData data, void * security_user
             /* Remember the commodity */
             gnc_commodity_table_insert(gnc_get_current_commodities(), commodity);
 
-            /* Remember this new commodity for us as well */
-            info->created_commodites = g_list_prepend(info->created_commodites, commodity);
 	    g_free (commodity_namespace);
 
         }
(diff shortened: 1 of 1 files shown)


jeanlaroche:
> After thinking about this, I think that this change is not quite correct. When the change was made to process multiple files per call, the initialization code for the info struct was lost. Also the code that does something with created_commodites was lost so it is now leaked. I think it would be better to remove created_commodites entirely (it never did anything useful) and move the initialization of the other three fields to the top level gnc_file_ofx_import function where the info struct is allocated. 

I'm confused. The info struct is allocated in `gnc_file_ofx_import()` which is called only once. Then `gnc_file_ofx_import()` is called which triggers a series of file imports. So the fields are not re-initialized at each file. Only for the first file of the series. 

I do agree that `ofx_created_commodites` does nothing useful! I didn't want to take it upon myself to remove it but I don't mind removing it if others think it's OK if it disappears entirely!

mtalexander:
> I'm confused. The info struct is allocated in `gnc_file_ofx_import()` which is called only once. Then `gnc_file_ofx_import()` is called which triggers a series of file imports. So the fields are not re-initialized at each file. Only for the first file of the series.

This doesn't seem to make sense.   You mention gnc_file_ofx_import twice in seemingly contradictory ways.  I think you're saying the same thing I did: the initialization should be moved to gnc_file_ofx_import where the info struct is allocated, right?

jeanlaroche:
> > I'm confused. The info struct is allocated in `gnc_file_ofx_import()` which is called only once. Then `gnc_file_ofx_import()` is called which triggers a series of file imports. So the fields are not re-initialized at each file. Only for the first file of the series.
> 
> This doesn't seem to make sense. You mention gnc_file_ofx_import twice in seemingly contradictory ways. I think you're saying the same thing I did: the initialization should be moved to gnc_file_ofx_import where the info struct is allocated, right?

Argh my bad, copy paste mistake! `gnc_file_ofx_import()` is called at the start of the import, and only once. Then `gnc_file_ofx_import()` is called repeatedly.
The initialization *is already* in `gnc_file_ofx_import()` and is called only once. So I'm not sure what you want changed...

mtalexander:
I think we're both getting confused here between gnc_file_ofx_import and gnc_file_ofx_import_process_file.  You mention gnc_file_ofx_import three times in the last comment and at least one of them probably should be gnc_file_ofx_import_process_file.  On the other hand I was confused about where your change is located.  It is already in gnc_file_ofx_import which is where I think it should be.  I think we can end this somewhat confused discussion.  Sorry for the confusion.

jeanlaroche:
Holy moly I'm losing my marbles!!! You're right, I repeated the same stupid mistake in my last comment! Maaaan, sorry about that! :D 
Still, should I remove the stuff about ofx_created_commodites ? Keep the code cleaner?

mtalexander:
Yes, I would say remove ofx_created_commodites.  It never did much.  Before the changes to handle multiple files there were some comments about some day adding code to tell the user about commodities that had been created, but it was never written.  The comments and the code that released the list of commodities was deleted as part of that change.  All that's left now is the code that creates the list which is then leaked when the import is done.

If you want to be completely clear about it you could do two commits, one to initialize things and the other to delete ofx_created_commodites.  Then someone coming back in the future would more easily be able to see what happened.  Removing  ofx_created_commodites in the first commit avoids having to decide whether to initialize it and makes the two commits more independent of each other.

jeanlaroche:
> If you want to be completely clear about it you could do two commits, one to initialize things and the other to delete ofx_created_commodites. Then someone coming back in the future would more easily be able to see what happened. Removing ofx_created_commodites in the first commit avoids having to decide whether to initialize it and makes the two commits more independent of each other.

I think I'm confused again! :D I thought we agreed the initializations were done in the right place currently. So all that needs to be done is to remove ofx_created_commodites everywhere in the code. Why would that be best done in 2 commits? Sorry if I'm a bit slow here, but I'm really confused! :)

mtalexander:
It's confusing.  When I said two commits, I meant two with respect to current master not with respect to your current pull request.  You're right that the pull request does the initialization correctly.  The easiest thing to do would be to add another commit to the pull request that deletes ofx_created_commodites.  This would be fine with me.

The reason I suggested two commits is that it keeps each commit to a single,, well-defined purpose.  In this case both are so simple that it isn't a big deal, but it's generally better not to mix different changes in the same commit.  It makes things less confusing and makes it easier to undo changes if desired.  Again, this is so simple that it doesn't really matter, but it's a good habit nevertheless.

So the bottom line is that I would suggest that you keep what you have done as it is and add a second commit that deletes ofx_created_commodites.

jeanlaroche:
OK Got it. I went ahead and removed the code related to ofx_created_commodites

jralls:
Thanks!

PR #748, opened 2020-06-30
 Adjusted python bindings for GnuCash 4.0 to support new session open mode

pkzw:
A GnuCash session may be either created from Python
with new keyword parameter open_mode and a valid session open mode
or for compatibility reasons still with the Boolean triple
(ignore_lock, is_new, force_new).
If both are specified they must be consistent.


jralls:
Thanks, but we already have an open #736 for this.

pkzw:
Sorry, I did not recognize it. I will have a deeper look at https://github.com/Gnucash/gnucash/pull/736.

PR #750, opened 2020-07-07
Bug 797847 - Best match probability calculation on import is too pess…

jeanlaroche:
…imistic

Lowering the minimum value for the auto-clear preference. This is to allow user to still auto
clear even when date don't match quite exactly.


jralls:
Merged, but I had to recreate the commit because the preference is now in `dialog-preferences.glade`

PR #754, opened 2020-07-12
[guile-3] change _ to G_

christopherlam:
In guile-3 _ is a reserved symbol. Change to G_ for gettext by guile convention.


christopherlam:
From my understanding the `G_` must be added somewhere to be recognised as translatable? Not sure where.

jralls:
You did that at line 44 of core-utils.scm.

christopherlam:
> You did that at line 44 of core-utils.scm.

I'm referring to `gnucash-pot.cmake:24` need to add `--keyword=G_`?

jralls:
Ah, different meaning of "recognized". That's correct.

PR #755, opened 2020-07-12
[gnucash-cli] -R show should accept & try load datafile

christopherlam:
* file_to_load argument, if present, would be a candidate for loading.
* if loading fails, show report details anyway.

How to fix?
```
* 23:44:26 ERROR <gnc.engine> gnc_uri_get_components: assertion 'uri != NULL && strlen (uri) > 0' failed
* 23:44:26 ERROR <gnc.engine> gnc_uri_get_components: assertion 'uri != NULL && strlen (uri) > 0' failed
terminate called after throwing an instance of 'std::logic_error'
  what():  basic_string::_M_construct null not valid
Aborted (core dumped)
```

### DIFF
--- a/gnucash/gnucash-cli.cpp
+++ b/gnucash/gnucash-cli.cpp
@@ -103,7 +103,8 @@ Gnucash::GnucashCli::configure_program_options (void)
     ("report,R", bpo::value (&m_report_cmd),
      _("Execute report related commands. The following commands are supported.\n\n"
      "  list: \tLists available reports.\n"
-     "  show: \tDescribe the options modified in the named report.\n"
+     "  show: \tDescribe the options modified in the named report. A datafile \
+may be specified to describe some saved options.\n"
      "  run: \tRun the named report in the given GnuCash datafile.\n"))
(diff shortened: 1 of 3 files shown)


jralls:
> How to fix?

First write up the problem: What exact command caused the crash and the stack trace from the crash.

christopherlam:
1. gnucash-cli -R show --name "reportname"
2. gnucash-cli -R show --name "reportname" datafile.gnucash

jralls:
Well, it doesn't crash for me, I guess you figured that out and fixed it.

But I'm not convinced that there's any point to opening a session only to destroy it so this is more of a hook than anything else. If the user passes in a big book it could take quite a while to load and for absolutely no benefit. In any case we're not going to change the location of saved report configurations in a stable branch so this should go to master.

christopherlam:
The only use case I can think of is to describe a report which has custom accounts. Eg transaction report's source accounts option. Without loading session, these accounts are described as "null, null, null". With session they show "Bank, Savings, House" appropriately.

jralls:
Ah, OK.

PR #756, opened 2020-07-13
changes for guile-3

christopherlam:
Last couple changes for guile-3 compatibility.


christopherlam:
I think this is ready for testing

PR #758, opened 2020-07-16
Modify association dialog titles and menu items.

jralls:
A counter proposal to #741.

With additional changes as discussed on [IRC](https://code.gnucash.org/logs/2020/07/07.html#T15:42:30) and [again](https://code.gnucash.org/logs/2020/07/16.html#T16:58:55).


gnucash/gnome/gnc-plugin-page-invoice.c
```diff
@@ -218,20 +218,15 @@ static GtkActionEntry gnc_plugin_page_invoice_actions [] =
         G_CALLBACK (gnc_plugin_page_invoice_cmd_new_invoice)
     },
     {
-        "BusinessAssociationAction", NULL, "_Update Association for Invoice", NULL,
-        "Update Association for current Invoice",
+        "BusinessAssociationAction", NULL, "_Manage Associated Document...", NULL,
```
fellen, line 221: We want to switch to "…" ([altgr]+[.] on my keyboard)

gnucash/gnome/gnc-plugin-page-register.c
```diff
@@ -302,10 +309,9 @@ static GncInvoice* invoice_from_split (Split* split);
...
-#define ASSOCIATE_TRANSACTION_TIP        N_("Update Association for the current transaction")
-#define ASSOCIATE_TRANSACTION_OPEN_TIP   N_("Open Association for the current transaction")
-#define ASSOCIATE_TRANSACTION_REMOVE_TIP N_("Remove the association from the current transaction")
-#define JUMP_ASSOCIATED_INVOICE_TIP      N_("Open the associated invoice")
+#define ASSOCIATE_TRANSACTION_TIP        N_("Manage the document associated with the current transaction")
+#define ASSOCIATE_TRANSACTION_OPEN_TIP   N_("Open the associated document for the current transaction")
+#define JUMP_ASSOCIATED_INVOICE_TIP      N_("Jump to the associated invoice")
 #define CUT_SPLIT_TIP                    N_("Cut the selected split into clipboard")
 #define COPY_SPLIT_TIP                   N_("Copy the selected split into clipboard")
 #define PASTE_SPLIT_TIP                  N_("Paste the split from the clipboard")
```
fellen, line 317: Tips should be full sentences with full stop.

gnucash/gnome/gnc-plugin-page-register.c
```diff
@@ -302,10 +309,9 @@ static GncInvoice* invoice_from_split (Split* split);
 #define PASTE_TRANSACTION_TIP            N_("Paste the transaction from the clipboard")
 #define DUPLICATE_TRANSACTION_TIP        N_("Make a copy of the current transaction")
 #define DELETE_TRANSACTION_TIP           N_("Delete the current transaction")
-#define ASSOCIATE_TRANSACTION_TIP        N_("Update Association for the current transaction")
-#define ASSOCIATE_TRANSACTION_OPEN_TIP   N_("Open Association for the current transaction")
-#define ASSOCIATE_TRANSACTION_REMOVE_TIP N_("Remove the association from the current transaction")
-#define JUMP_ASSOCIATED_INVOICE_TIP      N_("Open the associated invoice")
+#define ASSOCIATE_TRANSACTION_TIP        N_("Manage the document associated with the current transaction")
```
fellen, line 312: Perhaps we should replace "Manage" in the tooltips by "Add, change or remove" to be more explicit?

gnucash/gnome/gnc-plugin-page-register2.c
```diff
@@ -370,7 +370,7 @@ static GtkActionEntry gnc_plugin_page_register2_actions [] =
         G_CALLBACK (gnc_plugin_page_register2_cmd_exchange_rate)
     },
     {
-        "JumpTransactionAction", GNC_ICON_JUMP_TO, N_("_Jump"), NULL,
+        "JumpTransactionAction", GNC_ICON_JUMP_TO, N_("_Jump to other account"), NULL,
```
fellen, line 373: Shouldn't it be "…to the other…"?

gnucash/gnome/dialog-assoc.c
```diff
@@ -537,7 +537,9 @@ row_selected_bus_cb (GtkTreeView *view, GtkTreePath *path,
             return;
         }
 
-        ret_uri = gnc_assoc_get_uri_dialog (GTK_WINDOW(assoc_dialog->window), _("Change a Business Association"), uri);
+/* Translators: This is the title of a dialog box for associating an external
+ * file or URI with the current bill, invoice, transaction, or voucher. */
```
fellen, line 541: Results in
```
#. Translators: This is the label of a dialog box that lists all of the
#. * transaction that have files or URIs associated with them.

```
Better no beginning comment continuation star in translator comments. Same issue below, too

gnucash/gnome/gnc-plugin-page-register.c
```diff
@@ -483,7 +484,10 @@ static GtkActionEntry gnc_plugin_page_register_actions [] =
         G_CALLBACK (gnc_plugin_page_register_cmd_exchange_rate)
     },
     {
-        "JumpTransactionAction", GNC_ICON_JUMP_TO, N_ ("_Jump"), NULL,
+/* Translators: This is a menu item that will open a register tab for the
+ * account of the first other account in the current transaction's split list
+ * with focus on the current transaction's entry in that register. */
+        "JumpTransactionAction", GNC_ICON_JUMP_TO, N_ ("_Jump to Other Account"), NULL,
```
fellen, line 490: "_Jump to other account" is already used. Capitalization required?

gnucash/gnome/gnc-plugin-page-register.c
```diff
@@ -302,10 +309,9 @@ static GncInvoice* invoice_from_split (Split* split);
 #define PASTE_TRANSACTION_TIP            N_("Paste the transaction from the clipboard")
 #define DUPLICATE_TRANSACTION_TIP        N_("Make a copy of the current transaction")
 #define DELETE_TRANSACTION_TIP           N_("Delete the current transaction")
-#define ASSOCIATE_TRANSACTION_TIP        N_("Update Association for the current transaction")
-#define ASSOCIATE_TRANSACTION_OPEN_TIP   N_("Open Association for the current transaction")
-#define ASSOCIATE_TRANSACTION_REMOVE_TIP N_("Remove the association from the current transaction")
-#define JUMP_ASSOCIATED_INVOICE_TIP      N_("Open the associated invoice")
+#define ASSOCIATE_TRANSACTION_TIP        N_("Manage the document associated with the current transaction")
+#define ASSOCIATE_TRANSACTION_OPEN_TIP   N_("Open the associated document for the current transaction")
+#define JUMP_ASSOCIATED_INVOICE_TIP      N_("Jump to the associated invoice")
```
fellen, line 314: In the tip we can be more explicit:
"Jump to the associated bill, invoice or voucher"

gnucash/gnome/gnc-plugin-page-register.c
```diff
@@ -483,7 +484,10 @@ static GtkActionEntry gnc_plugin_page_register_actions [] =
         G_CALLBACK (gnc_plugin_page_register_cmd_exchange_rate)
     },
     {
-        "JumpTransactionAction", GNC_ICON_JUMP_TO, N_ ("_Jump"), NULL,
+/* Translators: This is a menu item that will open a register tab for the
+ * account of the first other account in the current transaction's split list
+ * with focus on the current transaction's entry in that register. */
+        "JumpTransactionAction", GNC_ICON_JUMP_TO, N_ ("_Jump to Other Account"), NULL,
         N_ ("Jump to the corresponding transaction in the other account"),
```
fellen, line 491: Unrelated, but IMHO wrong. Either
"Jump to the corresponding split [of this transaction] in the other account." or
"Show this transaction in the other account."

fellen:
Another thought of unification:
We could use "Show …" for data file internal items and "Open …" for external items.

@cstim , @jralls your opinion?

cstim:
I'm on holiday and AFK, I can comment on this in detail only after July 28th.

Aside, I think the wording for the associated URL or document should rather be similar to when adding a URL or hyperlink in Microsoft Word documents or Google doc documents. The association itself is called a "link" I think, and editing it is called editing I think, but in general I vote for following the examples from there instead of inventing our own wording.

jralls:
> Aside, I think the wording for the associated URL or document should rather be similar to when adding a URL or hyperlink in Microsoft Word documents or Google doc documents. The association itself is called a "link" I think, and editing it is called editing I think, but in general I vote for following the examples from there instead of inventing our own wording.

Nope, at least not on Google Docs and LibreOffice where it's Insert>Link. Setting aside that it's a somewhat different use case I don't think that it makes sense to create an Insert menu just for one item.

gnucash/gnome/gnc-plugin-page-register.c
jralls: How about "Open a new register tab for the other account with focus on this transaction."?

gnucash/gnome/gnc-plugin-page-register.c
jralls: OK, but few if any are, so they should all be done in a single commit and that should be done at the same time as the ellipsis change.

jralls:
> We could use "Show …" for data file internal items and "Open …" for external items.

I prefer "Jump" to "Show" because we've used jump forever and because "Show" by itself under the bent arrow on the toolbar is less meaningful than "Jump" is.

gnucash/gnome/gnc-plugin-page-register.c
fellen: You can also jump between already open register tabs, e.g. forward and back.

gnucash/gnome/gnc-plugin-page-register.c
fellen: For this section yes. But my fear is, there could be other sections containing e.g. only an infinitive instead of a full sentence. Then it is not trivial.

gnucash/gnome/gnc-plugin-page-register.c
jralls: I don't like "corresponding" at all, but I'm not wild about "associated" either.  

For jumping to the other register it's the same transaction, neither corresponding nor associated. IIUC the operation that connects a transaction to a bill/invoice/voucher is called posting so maybe this tooltip could be "Jump to the posting bill/invoice/voucher".

If it weren't for the possible confusion about URLs vs. local files I'd go for changing "associate" to "link" for transactions and external documents. Attach presents the same confusion potential in the other direction. Possible synonyms connect, couple, join, and tie aren't terribly satisfying either.

gnucash/gnome/gnc-plugin-page-register.c
jralls: Well, you have ~2 years to find them. ;-)

gnucash/gnome/gnc-plugin-page-register.c
jralls: True. "Jump to register tab of the other account with focus on this transaction."? "Jump to register tab of the other account, opening one if needed, with focus on this transaction."? How wordy do we want to get? The explanation in the [Help Manual section](https://www.gnucash.org/docs/v4/C/gnucash-help/acct-jump.html) isn't very good either.

fellen: "Jump to register tab of the other account with focus on this transaction."  is sufficient here and "Jump to register tab of the other account, opening one if needed, with focus on this transaction." in help, but I am open for improvements.

There are additional reference sections in Help:
https://www.gnucash.org/docs/v4/C/gnucash-help/ch04s03.html#Trans-menus
https://www.gnucash.org/docs/v4/C/gnucash-help/ch04s03.html#Trans-toolbar
(should get some rework so that we have to define an entry only once)

gnucash/gnome/gnc-plugin-page-invoice.c
jralls: OK, but I think that should be a separate commit that fixes all of them. And since it will blow up every translation it should go on master.

jralls:
@cstim, a reminder... I hope you enjoyed your vacation.

cstim:
@jralls  thanks for the reminder. I've reviewed this PR now as a whole.

- The PR is surely an improvement over what we currently have. Unless we agree on something else, it should simply go in.
- Some nitpick on the wording fine-tune: Strictly speaking, the "manage" (edit) menu item does not manage the associated document but rather "manage document association"... the document itself is managed somewhere else but not from within gnucash. Ok, just regard this as a humorous nitpick.
- But apart from the nitpick, I propose to drop the word "association" ("associated document") and instead switch to "**link**" ("linked document").  "manage document link" or "manage linked document", "open linked document" and so on. The original author introduced the whole feature under the headline of "association", but I would consider "link" even more appropriate and it is not yet used (except in very few html report places).

jralls:
@cstim, I actually think your nitpick is correct: The dialog does indeed manage the association rather than the associated document.

'Link' has the advantage of being shorter that 'association', I'm just concerned that users will think that it has to be an http URI, plus they've gotten used to association by now.

gjanssens:
FWIW, I actually like "Link" better as well than "Association" even though "Association" was initially used. For me it's not a strong reference to an http URI alone. It just indicates this is something outside of gnucash we are referring to.
Of course I'm a developer, so my brain may be trained differently from an ordinary user. However if LibreOffice also uses the term "link" to refer to an external document (as opposed to "copy") that's a good extra motivation.

jralls:
OK, I've revised the first commit to use Link in place of Association.

The second commit is the beginning of the work needed to get all of the symbol names to match. On the one hand that might seem a bit pedantic; on the other it will make life easier later when we've forgotten why we changed the UI names and we're trying to find the "link" code.

fellen:
@jralls Did you miss gnucash/gnome/gnc-plugin-page-register.c?

jralls:
@fellen, I won't say that I didn't miss something, but there's 55 lines of changes in `gnc-plugin-page-register.c`. What in particular are you refering to?

fellen:
> @fellen, I won't say that I didn't miss something, but there's 55 lines of changes in `gnc-plugin-page-register.c`. What in particular are you refering to?

```
/gnucash/gnucash/gnome/gnc-plugin-page-register.c:420:40: error: ‘LINK_TRANSACTION_LABEL’ undeclared here (not in a function); did you mean ‘COPY_TRANSACTION_LABEL’?

  420 |         "LinkTransactionAction", NULL, LINK_TRANSACTION_LABEL, NULL,

      |                                        ^~~~~~~~~~~~~~~~~~~~~~

      |                                        COPY_TRANSACTION_LABEL

/gnucash/gnucash/gnome/gnc-plugin-page-register.c:1019:5: error: initialization of ‘const char *’ from incompatible pointer type ‘GtkActionEntry *’ [-Werror=incompatible-pointer-types]

 1019 |     LINK_TRANSACTION_LABEL,

      |     ^~~~~~~~~~~~~~~~~~~~~~

```

jralls:
Thanks, fixed.

gnucash/gnome/gnc-plugin-page-register.c
```diff
@@ -4518,7 +4518,7 @@ gnc_plugin_page_register_cmd_delete_transaction (GtkAction* action,
 }
 
 static void
-gnc_plugin_page_register_cmd_associate_transaction (GtkAction *action,
+gnc_plugin_page_register_cmd_linked_transaction (GtkAction *action,
                                                     GncPluginPageRegister* plugin_page)
```
gjanssens, line 4522: Complete nitpick: the function name is shorter now, but the function parameters on the second line are still aligned on the old function name length. I only saw this because you fixed it for a single instance a bit further down this file.

fellen:
Still a few occurences:
Association

fellen:
> Edit: Retracted, wrong version.
No, wrong language. So it still exists:
Association

jralls:
I think I've got them all this time.

Bob-IT:
@jralls This has got much bigger than when I last looked at it, have not been doing much as too busy with non Gnucash stuff.

Did you mean to change the KVP name and head preference setting as my test file is not showing the DocLinks ?
This could be fixed with a scrub function or possible just by changing xaccTransGetDocLink to look at old 'assoc-uri' first, if set immediately save to 'doclink-uri' and delete old entry. Similar changes would need to be done for the invoice KVP and path head setting.

jralls:
@Bob-IT, yes, I did a pretty thorough scrub through the code replacing `assoc` with `doclink` to protect against confusion in a year or two when we'd forgotten about the old name. I didn't consider that the symbol names might be persisted in KVP slot names. While I like the idea of the backward-compatible transition it would break forward compatibility--opening a file in 4.1 after it had been touched by 4.2 wouldn't find the slot--so I think it would be safer to just put the slot and preference names back to assoc with a comment explaining why.

Bob-IT:
@jralls , I have made the changes to use the old name and added some notes. Also changed some other references that used a form of associate.

jralls:
Thanks @Bob-IT.

PR #759, opened 2020-07-17
sankey cash-flow chart

christopherlam:
1. upgrade chartjs to 3.0.0beta, fix html-chart to handle chartjs breaking changes
2. add chartjs-chart-sankey.js
3. add sankey.scm

with permission from @kurkle: https://github.com/kurkle/chartjs-chart-sankey/issues/1

image


jralls:
No beta dependencies, and this is a new feature so it goes to master.

christopherlam:
Simple algorithm - eg from book with the following:

* Income:Sales -> Asset:Bank $100
* Asset:Bank -> Asset:Savings $20
* Asset:Bank -> Expense:Fuel $40
* Asset:Bank -> Expense:Food $20

Originating account = Asset:Bank

1. collate all non-originating account, non-trading splits:
* Income:Sales -$100
* Asset:Savings +$20
* Expense:Fuel +$40
* Expense:Food +$40

2. Tally all accounts, *including* accounts' parents:
* Income:Sales -$100
* Income -$100
* Expense:Fuel +$40
* Expense:Food +$40
* Expense +$80

3. Insert directed flows into chart: account->parent if negative, parent->account if positive

* Income:Sales -> Income $100
* Income -> ROOT $100
* Expense -> Expense:Fuel $40
* Expense -> Expense:Food $40
* ROOT -> Expense $80

ROOT refers to original account "Asset:Bank"

christopherlam:
@gjanssens is there any merit to the idea of including both chartjs-2 and chartjs-3 in the distribution? Problem: sankey requires chartjs-3, and I'm not 100% sure that chartjs-3 won't leave webkit on Windows behind. So, to upgrade chartjs to 3 may cause windows charts to suffer. If we include both 2 and 3, then we can use advanced chartjs features only when necessary.

jralls:
And have a feature that doesn't work for most of our users. -1.

gjanssens:
I agree this is not a good option.

christopherlam:
> And have a feature that doesn't work for most of our users. -1.

I presume you mean most users use Win32 and may be limited.

I've never been able to generate win32 builds therefore have never been able to test whether chartjs3-on-webkit1 is broken. Hence the idea: keep chartjs2 (which will eventually become stale) for most charts, and add chartjs3 for sankey. Sankey will be initially hidden via --extras and if I'm confident that chartjs3-on-win32works well then we could transition the plumbing to use chartjs3 throughout.

gjanssens:
> > And have a feature that doesn't work for most of our users. -1.
> 
> I presume you mean most users use Win32 and may be limited.
> 
> I've never been able to generate win32 builds therefore have never been able to test whether chartjs3-on-webkit1 is broken. 

As far as I can see chart.js is just a bunch of javascript files and its consumers are written in guile. That means you should be able to test without rebuilding. I t hink you can install a nightly of  gnucash on a win32 system, replace chartjs files with v3 and then make the necessary tweaks in the report code to work with chartjs3. You'll probably have to make those changes with elevated privileges though.

jralls:
Please don't combine PRs. You'll make a commit cycle in history. 

But @gjanssens and my comments from 4 years ago still stand: A feature that isn't available on all platforms is a bad idea.

jralls:
Superseded by #2110 and #2112

PR #760, opened 2020-07-25
[business-urls] slightly better url query parsing

christopherlam:
Preparation for [bug 797852](https://bugs.gnucash.org/show_bug.cgi?id=797852) is to make url query parsing less ugly. There's https://uriparser.github.io/doc/api/latest/ but we don't need its full capabilities. I *think* the hash-table is freed appropriately. Any additional `g_free` leads to double-free.

After this, the real bug fix is to encode aging-report end-date into string or number for the url, and decode it in business-urls.c to create owner report with custom end-date.

### DIFF
--- a/gnucash/gnome/business-urls.c
+++ b/gnucash/gnome/business-urls.c
@@ -163,120 +163,116 @@ jobCB (const char *location, const char *label,
 
 /* ================================================================= */
 
-#define RETURN_IF_NULL(inst)                                            \
+#define DISABLE_REPORT_IF_NULL(inst)                                    \
   if (NULL == inst)                                                     \
   {                                                                     \
     result->error_message =                                             \
       g_strdup_printf (_("No such owner entity: %s"), location);        \
-    return FALSE;                                                       \
+    show_report = FALSE;                                                \
   }
+
+/* parses a string "user=john&pass=smith&age=41" into a string-keyed
+   GHashTable. String must not contain non-ASCII chars. Duplicate keys
+   will be ignored. */
+static GHashTable *parse_parameters (const gchar *parms)
+{
+    GHashTable *rethash;
+    gchar *query, *p, *orig_query;
+
+    rethash = g_hash_table_new_full (g_str_hash, g_str_equal, g_free, g_free);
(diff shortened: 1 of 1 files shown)


gnucash/gnome/business-urls.c
```diff
@@ -168,78 +168,77 @@ jobCB (const char *location, const char *label,
...
+{
+    GHashTable *rethash;
+    gchar *query, *p;
+
+    rethash = g_hash_table_new_full (g_str_hash, g_str_equal, g_free, NULL);
+    query = g_strdup (parms);
+
+    while ((p = strsep (&query, "&\n")))
+    {
+        gchar *var = strtok (p, "=");
```
jralls, line 187: Since there should be exactly one instance of '=' in the string this would be more efficient:
```C
gchar * val = strchr (p, '=');
*(val++) = '\0';
if (val && !g_has_table_contains (rethash, p))
   g_hash_table_insert (rethash, g_strdup (p), g_strdup (val));
```

gnucash/gnome/business-urls.c
```diff
@@ -168,78 +168,77 @@ jobCB (const char *location, const char *label,
...
 
+/* parses a string "user=john&pass=smith&age=41" into a string-keyed
+   GHashTable. String must not contain non-ASCII chars. Duplicate keys
+   will be ignored. */
+static GHashTable *parse_parameters (const gchar *parms)
+{
+    GHashTable *rethash;
+    gchar *query, *p;
+
+    rethash = g_hash_table_new_full (g_str_hash, g_str_equal, g_free, NULL);
```
jralls, line 182: Both the key and value are `g_strdup`ed so both need to be `g_free`d: 
```C
rethash = g_hash_table_new_full (g_str_hash, g_str_equal, g_free, g_free);
```

Your alternative of freeing `ownerptr` and `acctptr` assumes that those are the only values in the hashtable. Leave the ownership with the hashtable and you don't have to worry about that, nor about freeing them at the end of the function.

jralls:
Looks good to me.

PR #761, opened 2020-07-28
Make python bindings GncLot.make_default() into a classmethod.

QWork66:
GncLot.make_default creates a new GncLot for an account.
It can now do so without an existing instance.  Python bindings
classmethod capability is fixed and expanded to make this possible.
Tests are added.


jralls:
The generic changes should be in a separate commit from the GncLot-specific ones. `ya_add_classmethod` is used in example_scripts/str_methods.py so you need to rename it there too.

QWork66:
Thanks for your comments.  I will make these changes.

QWork66:
Hmm. I see that I have made this PR very messy by rebasing it on the updated master. I am very sorry for that.  It is doubtless due my poor understanding of git.  I will close this and start again.

PR #762, opened 2020-07-29
Add support for specifying an opening balance account

rhabacker:
This pull request extends the gnucash sources to support dedicated opening balance accounts.  A script is added to add opening balance sheet accounts to the account templates. The account template extensions adapted by the script are included, as well as individual manual adjustments not covered by the script.

For background information, see https://bugs.gnucash.org/show_bug.cgi?id=797836.


fellen:
I have applied your first commit as it is not related to this PR. The rest needs some rethinking, which I will do on the bug report.

If you like you can rebase and force push this PR.

gnucash/gnome-utils/dialog-account.c
```diff
@@ -1273,6 +1302,7 @@ commodity_changed_cb (GNCGeneralSelect *gsl, gpointer data)
 
     selection = gtk_tree_view_get_selection (GTK_TREE_VIEW (aw->transfer_tree));
     gtk_tree_selection_unselect_all (selection);
+    gnc_account_opening_balance_button_update (aw, currency);
```
jralls, line 1305: This could be a problem: What if the user changes the commodity on an account with the opening_balance flag set and there's already another account of that same commodity with the flag set?

libgnucash/app-utils/gnc-ui-util.c
```diff
@@ -966,6 +966,13 @@ gnc_find_or_create_equity_account (Account *root,
     g_return_val_if_fail (currency != NULL, NULL);
     g_return_val_if_fail (root != NULL, NULL);
 
+    if (equity_type == EQUITY_OPENING_BALANCE)
+        account = gnc_account_lookup_by_opening_balance (root, currency);
+
+    if (account &&
+            gnc_commodity_equiv (currency, xaccAccountGetCommodity (account)))
+        return account;
```
jralls, line 974: You need to initialize account to NULL at declaration to prevent a crash in the EQUITY_RETAINED_EARNINGS case. But it would be better to `if (account) return account` in the previous if-block. There's no need to test the commodity match, `gnc_account_lookup_by_opening_balance` already did that.

libgnucash/engine/Account.cpp
```diff
@@ -113,6 +113,7 @@ enum
 
     PROP_LOT_NEXT_ID,                   /* KVP */
     PROP_ONLINE_ACCOUNT,                /* KVP */
+    PROP_OPENING_BALANCE,               /* KVP */
```
jralls, line 116: Name this PROP_IS_OPENING_BALANCE to avoid confusion with the value of the account's opening balance.

libgnucash/engine/Account.cpp
```diff
@@ -4088,6 +4138,18 @@ xaccAccountSetPlaceholder (Account *acc, gboolean val)
     set_boolean_key(acc, {"placeholder"}, val);
 }
 
+gboolean
+xaccAccountGetOpeningBalance (const Account *acc)
+{
+    return boolean_from_key(acc, {"opening-balance"});
```
jralls, line 4144: KVP lookups are expensive and this is going to get called a lot when searching for the opening balance account. You can speed up the check a bit by
```
       return GET_PRIVATE(acc)->type == ACCT_TYPE_EQUITY && boolean_from_key(acc, {"opening-balance"});
```

libgnucash/engine/Account.h
```diff
@@ -1170,6 +1176,22 @@ gboolean xaccAccountGetPlaceholder (const Account *account);
  *  @param val The new state for the account's "placeholder" flag. */
 void xaccAccountSetPlaceholder (Account *account, gboolean val);
 
+/** Get the "opening-balance" flag for an account.  If this flag is set
+ *  then the account is used for opening balance transactions.
+ *
+ *  @param account The account whose flag should be retrieved.
+ *
+ *  @return The current state of the account's "opening-balance" flag. */
+gboolean xaccAccountGetOpeningBalance (const Account *account);
```
jralls, line 1185: Name these `xaccAccount[GS]etIsOpeningBalance`. `xaccAccount[GS]etOpeningBalance` suggests retrieving the balance itself.

libgnucash/engine/Account.h
```diff
@@ -1542,6 +1564,7 @@ const char * dxaccAccountGetQuoteTZ (const Account *account);
 #define ACCOUNT_NOTES_		"notes"
 #define ACCOUNT_BALANCE_	"balance"
 #define ACCOUNT_NOCLOSING_	"noclosing"
+#define ACCOUNT_OPENING_BALANCE "opening-balance"
```
jralls, line 1567: `ACCOUNT_IS_OPENING_BALANCE_`

Dunno why the trailing underscore, but be consistent.

libgnucash/app-utils/gnc-ui-util.c
```diff
@@ -985,7 +992,11 @@ gnc_find_or_create_equity_account (Account *root,
 
     if (account &&
             gnc_commodity_equiv (currency, xaccAccountGetCommodity (account)))
+    {
+        if (equity_type == EQUITY_OPENING_BALANCE && !xaccAccountGetOpeningBalance (account))
```
jralls, line 996: The second test should be unnecessary: If the opening_balance property was already set the account should have been found and returned up at line 970. If it somehow got by there's no harm in setting it again.

libgnucash/app-utils/gnc-ui-util.c
```diff
@@ -997,7 +1008,11 @@ gnc_find_or_create_equity_account (Account *root,
...
+            xaccAccountSetOpeningBalance (account, TRUE);
```
jralls, line 1013: The second test shouldn't be necessary.

jralls:
@fellen's comment on [the bug](https://bugs.gnucash.org/show_bug.cgi?id=797836) and my reply gives me the idea that we should expand this. Instead of a slot specific to opening balances we should name the slot something like "special-purpose" and give it a value. The two obvious ones to begin with are opening balance and retained earnings, mirroring the two types in gnc-ui-utils.c. Two others come immediately to mind: imbalance and orphaned capital gains, the latter used by the lot facility. You need not implement more than opening balances for this PR.

gjanssens:
> @fellen's comment on [the bug](https://bugs.gnucash.org/show_bug.cgi?id=797836) and my reply gives me the idea that we should expand this. Instead of a slot specific to opening balances we should name the slot something like "special-purpose" and give it a value. The two obvious ones to begin with are opening balance and retained earnings, mirroring the two types in gnc-ui-utils.c. Two others come immediately to mind: imbalance and orphaned capital gains, the latter used by the lot facility.

Probably trading accounts would benefit from this as well.

rhabacker:
> Instead of a slot specific to opening balances we should name the slot something like "special-purpose" and give it a value. 

You are refering to something like this ?
```
static const std::string  KEY_SPECIAL_PURPOSE("special-purpose");
...
return boolean_from_key (acc, {KEY_SPECIAL_PURPOSE, "opening-balance"});
...
set_boolean_key (acc, {KEY_SPECIAL_PURPOSE, "opening-balance"}, val);
...

**Comment by christopherlam** [comment]
> > Instead of a slot specific to opening balances we should name the slot something like "special-purpose" and give it a value.
> 
> You are refering to something like this ?

I think some of the dilemma is how to encode these `special-purpose`

- account slot `special-purpose/opening-balance`=`true` or `false`
- account slot `special-purpose/retained-earnings`=`true` or `false`
- account slot `special-purpose/imbalance`=`true` or `false`
- account slot `special-purpose/orphan`=`true` or `false`

or

- account slot `special-purpose` = `opening-balance` or `retained-earnings` or `imbalance` or `orphan` , etc

I'd lean towards the latter...

**Comment by jralls** [comment]
> Probably trading accounts would benefit from this as well.

Trading accounts are `ACCT_TYPE_TRADING` so they don't need this.

**Comment by jralls** [comment]
> I'd lean towards the latter...

Me too.

File: gnucash/gnome-utils/dialog-account.c [hunk 9]
```diff
@@ -204,6 +205,23 @@ gnc_account_commodity_from_type (AccountWindow * aw, gboolean update)
...
 }
 
+static void
+gnc_account_opening_balance_button_update (AccountWindow *aw, gnc_commodity *commodity)
+{
+    Account *account, *obAccount;
+    int hasSplits;
+    account = aw_get_account (aw);
+    obAccount = gnc_account_lookup_by_opening_balance (gnc_book_get_root_account (aw->book), commodity);
+    hasSplits = xaccAccountCountSplits (account, FALSE) > 0;
```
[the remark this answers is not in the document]
rhabacker: > It's better style to declare and initialize in a single statement where possible, e.g.

At opensuse 15.1 with g++ 7.5 I get

~/src/gnucash/gnucash/gnome-utils/dialog-account.c: In function ‘gnc_account_opening_balance_button_update’:
~/src/gnucash/gnucash/gnome-utils/dialog-account.c:214:5: error: ISO C90 forbids mixed declarations and code [-Werror=declaration-after-statement]
     Account *ob_account = gnc_account_lookup_by_opening_balance (gnc_book_get_root_account (aw->book), commodity);
     ^~~~~~~

gnucash/gnome-utils/gnc-tree-view-account.c
```diff
@@ -256,6 +256,29 @@ gnc_tree_view_account_placeholder_toggled (GtkCellRendererToggle *cell,
...
+    account = gnc_tree_view_account_get_account_from_path (tree_view, s_path);
+    if (account)
+    {
+        opening_balance = !gtk_cell_renderer_toggle_get_active (cell); // hasn't changed yet.
+        xaccAccountSetOpeningBalance (account, opening_balance);
+    }
+
+    /* Clean up */
+    gtk_tree_path_free (s_path);
+}
```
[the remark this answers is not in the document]
rhabacker: > This lacks the controls of dialog-account that ensure that the account is of the right type, there are no splits in the account and that there isn't already an opening balance account for the account's commodity.

Because this check is already be performed in gnc_account_opening_balance_button_update(): would it not be enough to check if the toggle button is enabled and if not, return without any change instead of duplicating the checks ?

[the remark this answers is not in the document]
rhabacker: Is that event handler really required ? It looks that it can be removed complety because setting a break point at  xaccAccountSetIsOpeningBalance() shows that it is only called on closing the dialog using the following call tree

1   xaccAccountSetIsOpeningBalance    Account.cpp      4173 0x7fffec140e4b 
2   gnc_ui_to_account                 dialog-account.c 470  0x7fffef8a123e 
3   gnc_finish_ok                     dialog-account.c 599  0x7fffef8a18d6 
4   gnc_edit_account_ok               dialog-account.c 913  0x7fffef8a2b00 
5   gnc_account_window_response_cb    dialog-account.c 989  0x7fffef8a2fe2 

while gnc_tree_view_account_opening_balance_toggled (), into which is also set a breakpoint, is never called while editing the account dialog.

gnucash/gnome-utils/dialog-account.c
rhabacker: In that case the commodity change could not be processed and the user should be informed about this issue.

jralls: And that happens where?

gnucash/gnome-utils/dialog-account.c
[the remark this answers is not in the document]
jralls: Do the decl and init at the beginning of the function:
```
static void
gnc_account_opening_balance_button_update (AccountWindow *aw, gnc_commodity *commodity)
{
    Account *account account = aw_get_account (aw);
    Account *ob_account = gnc_account_lookup_by_opening_balance (gnc_book_get_root_account (aw->book), 
                                                                                                                     commodity);
    gboolean has_splits  xaccAccountCountSplits (account, FALSE) > 0;

    if (xaccAccountGetType (account) != ACCT_TYPE_EQUITY)
        return;
...
```

gnucash/gnome-utils/gnc-tree-view-account.c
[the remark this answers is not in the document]
jralls: > Because this check is already be performed in gnc_account_opening_balance_button_update(): 

OK, mostly. That bails out if the account isn't an equity account, so if the user changes the account from equity to not-equity there's no way to make the check box not sensitive or `FALSE`.

> s that event handler really required ? 

Sorry, I've lost you. Which event handler?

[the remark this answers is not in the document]
rhabacker: > Sorry, I've lost you. Which event handler?

Sorry for be'ing somehow unspecific - I refered to gnc_tree_view_account_opening_balance_toggled() which handles the button "toggle" event.

> s no way to make the check box not sensitiv

This is simple to solve by using something like this
```
static void
gnc_account_opening_balance_button_update (AccountWindow *aw, gnc_commodity *commodity)
{
    Account *account = aw_get_account (aw);
    Account *ob_account = gnc_account_lookup_by_opening_balance (gnc_book_get_root_account (aw->book), commodity);
    gboolean has_splits = xaccAccountCountSplits (account, FALSE) > 0;

    if (xaccAccountGetType (account) != ACCT_TYPE_EQUITY)
    {
        gtk_widget_set_sensitive (aw->opening_balance_button, FALSE);
        return;
    }
```

gnucash/gnome-utils/dialog-account.c
rhabacker: in commodity_changed_cb(), see updated patch

rhabacker:
>   account slot special-purpose/opening-balance=true or false
>   account slot special-purpose/retained-earnings=true or false
>   account slot special-purpose/imbalance=true or false
>   account slot special-purpose/orphan=true or false

If there is a possibility that several of these flags can be used together, this form is required.

gjanssens:
> > Probably trading accounts would benefit from this as well.
> 
> Trading accounts are `ACCT_TYPE_TRADING` so they don't need this.

Right. I'm conflating things.

christopherlam:
> > account slot special-purpose/opening-balance=true or false
> > account slot special-purpose/retained-earnings=true or false
> > account slot special-purpose/imbalance=true or false
> > account slot special-purpose/orphan=true or false
> 
> If i understand this notation correctly as: there should be 4 independent slots this is required if there is a possibility that several of these flags can be used together.
> 
> > account slot special-purpose = opening-balance or retained-earnings or imbalance or orphan , etc
> 
> Does this notation mean that there is only one slot named `special-purpose` with type 'string', which could have the mentioned values ?

That's the idea yet. I can think of *no* use case whereby an account will be both imbalance and retained earnings. 

I think there's a 5th equity-type account: `Owner's Drawings`  ???

jralls:
> Does this notation mean that there is only one slot named special-purpose with type 'string', which could have the mentioned values ?

Exactly. As @christopherlam says, they're mutually exclusive so there's no real need for a separate boolean for each of them.

> I think there's a 5th equity-type account: Owner's Drawings ???

That's one of several equity accounts that users might have, but there's no GnuCash automatic process that uses them so there's no need to include them in this mechanism.

rhabacker:
> I think there's a 5th equity-type account: Owner's Drawings ?

Does that mean that the new slot specifies a kind of account type ?  

If so, wouldn't a more meaningful key name be better such as 
```
account slot equity-type = opening-balance or retained-earnings or imbalance or orphan , etc
```

jralls:
@rhabacker I like the name `equity-type` better than `special-purpose`, but as I responded to @christopherlam we need not try to work out an exhaustive enumeration. We're only interested in tagging those accounts that GnuCash needs to find for automatic operations and I think that's just the 4 we've identified.

fellen:
This comment is primary "loud thinking" about further potential flags:

I see a bunch more potential slots in the future. From e.g. https://www.datev.de/web/de/datev-shop/material/kontenrahmen-datev-skr-04/ (use "kostenfreier Download")
Currently we use the column Bilanzposten/GuV-Posten as parent account.
There are bunch of end notes:

EÜR/SB/HB: Some accounts should only be used for a **specific accounting method** do determinate the profit. Small organizations will use Einnahme-Überschuß-Rechnung, bigger Steuer-Bilanz (balance according tax law), but can additional create a Handels-Bilanz (according trading law).

U/G/K account  required to calculate/declare a **specific tax**. Currently only U is implemented as tax-related. the others should go there, too.

Some e.g. statistical data accounts have a specific **counter-account**.

jralls:
@fellen But aren't those relationships managed by the user rather than by automatic processes in GnuCash?

fellen:
> @fellen But aren't those relationships managed by the user rather than by automatic processes in GnuCash?

The user has to manage them according the notes, because we did not implement them. But there is potential to improve UX.

jralls:
> The user has to manage them according the notes, because we did not implement them. But there is potential to improve UX.

With KVP slots that aren't visible to the user? Seems unlikely, but please elaborate.

fellen:
> > The user has to manage them according the notes, because we did not implement them. But there is potential to improve UX.
> 
> With KVP slots that aren't visible to the user? Seems unlikely, but please elaborate.

e.g. I can think of a slot <close-into>guid</close-into>, which then gets used by close-book to get a more detailed result.

jralls:
That actually sounds worthwhile. The close-book function is currently pretty basic and could stand improvement, beginning with a less hacky way of keeping the closing transactions from messing up reports.

Completely orthogonal to this PR though.

libgnucash/engine/Account.h
```diff
@@ -924,6 +924,12 @@ Account *gnc_account_lookup_by_full_name (const Account *any_account,
 Account *gnc_account_lookup_by_code (const Account *parent,
                                      const char *code);
 
+/** The gnc_account_lookup_by_by_opening_balance () subroutine
+ *  works like gnc_account_lookup_by_name, but checks for the
+ *  opening balance flag set.
+ */
```
jralls, line 930: Doxygen will fill in the function name so lead off with an action verb. Don't refer to other functions in the description, they might change making your description inaccurate. Describe each parameter beginning with `@param` and what the function returns beginning with @return:
```C
/** Find the opening balance account for the currency.
 *
 *  @param parent The account of which the sought-for account is a descendant.
 *  @param commodity The commodity in which the account should be denominated
 *  @return The descendant account of EQUITY_TYPE_OPENING_BALANCE or NULL if one doesn't exist.
 */
```

jralls (changes requested):
@rhabacker Are you going to implement the generalization from `opening_balance` to `equity_type`?

util/add-opening-balances
```diff
@@ -0,0 +1,98 @@
...
+    fi
+done
+
+echo "------------------------------------------------------------"
+echo -e "The following files do not have an account with slot key 'opening-balance'"
+echo -e $nooba
+echo
+echo "------------------------------------------------------------"
+echo -e "The following files have an account with slot key 'opening-balance'"
+echo -e $oba
```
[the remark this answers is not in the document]
rhabacker: will be fixed in next push

gnucash/gtkbuilder/dialog-account.glade
```diff
@@ -1531,6 +1531,21 @@
                             <property name="top_attach">7</property>
                           </packing>
                         </child>
+                        <child>
+                          <object class="GtkCheckButton" id="opening_balance_button">
+                            <property name="visible">True</property>
+                            <property name="sensitive">False</property>
+                            <property name="can_focus">True</property>
+                            <property name="receives_default">False</property>
+                            <property name="tooltip_text" translatable="yes" comments="">This account holds opening balance transactions. Only one account can hold opening balance transactions.</property>
```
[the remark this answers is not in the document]
rhabacker: will be fixed in next push

rhabacker:
> @rhabacker Are you going to implement the generalization from `opening_balance` to `equity_type`?
yes, should be ready with current patch set.

jralls (changes requested):
There's two more issues:
1. A book with using the equity-type KVP and with opening balance accounts that aren't named Opening Balance in the user's language opened in an older version of GnuCash won't be able to find the right opening balance account so GnuCash will create a new one. We usually handle that with a feature flag and a two-phase introduction: The first phase recognizes the feature flag and can find opening balance accounts using `gnc_account_lookup_by_opening_balance`; the second phase in a later release can set the event-type KVP and the feature flag.
2. A user could rename an opening balance account to something else and not know that they need to set the opening balance flag. The first half of that is an existing problem, but it can be dealt with going forward by noticing the name and setting the flag automatically.

rhabacker:
>2. A user could rename an opening balance account to something else and not know that they need to set the opening balance flag. The first half of that is an existing problem, but it can be dealt with going forward by noticing the name and setting the flag automatically.

This use case lacks some basic information about the initial state of the opening balance account, so I can explain only the recent implementation yet.

In case a user runs gnucash with "equity-type" support and the opening balance account has a untranslated or translated name matching the implemented rule for opening balance accounts, on first access (gnc_account_create_opening_balance) to that account it will by default be tagged as opening balance account, see 
https://github.com/rhabacker/gnucash/blob/add-opening-balance-accounts/libgnucash/app-utils/gnc-ui-util.c#L996
https://github.com/rhabacker/gnucash/blob/add-opening-balance-accounts/libgnucash/app-utils/gnc-ui-util.c#L1012
https://github.com/rhabacker/gnucash/blob/add-opening-balance-accounts/libgnucash/app-utils/gnc-ui-util.c#L1045

rhabacker:
> We usually handle that with a feature flag and a two-phase introduction: The first phase recognizes the feature flag 

Can you give a real world example as implementation hint ?

rhabacker:
3. If I create a new book with gnucash compiled with the patches from this pr and a template having equity-type slot set (e.g. de/skr03) an additional account named "Anfangsbestand" is created and also tagged as opening-balance account., which should not be the case. Any hint, where this happens ?

christopherlam:
> > We usually handle that with a feature flag and a two-phase introduction: The first phase recognizes the feature flag
> 
> Can you give a real world example as implementation hint ?

See #585 -- `gnc_features_check_used (book, GNC_FEATURE_BUDGET_UNREVERSED)` -- current 4.x is already able to handle books with budget_unreversed feature but there are no current books in use, because the feature is not set yet. When this PR is merged, `gnc_features_set_used (book, GNC_FEATURE_BUDGET_UNREVERSED)` will set the feature and allow GnuCash to use a different code path for budgeting module and reports.

As a result, all 4.x can read books with feature, but only a future version can read and write books with feature.

jralls:
> In case a user runs gnucash with "equity-type" support and the opening balance account has a untranslated or translated name matching the implemented rule for opening balance accounts, on first access (gnc_account_create_opening_balance) to that account it will by default be tagged as opening balance account, see

Right. That check also needs to be run in the edit account dialog.

jralls:
> If I create a new book with gnucash compiled with the patches from this pr and a template having equity-type slot set (e.g. de/skr03) an additional account named "Anfangsbestand" is created and also tagged as opening-balance account., which should not be the case. Any hint, where this happens ?

The account creation would be from [assistant-hierarchy](https://github.com/Gnucash/gnucash/blob/maint/gnucash/gnome/assistant-hierarchy.c#L1412). My first guess as to why is that that runs before the template account is registered in KVP. Failing that it must be getting around the check for an existing account some other way.

rhabacker:
> Right. That check also needs to be run in the edit account dialog.
and on deleting account as implemented with commit 6ccdeab

rhabacker:
> See #585 -- `gnc_features_check_used (book, GNC_FEATURE_BUDGET_UNREVERSED)` -- current 4.x is already able to handle books with budget_unreversed feature but there are no current books in use, because the feature is not set yet. When this PR is merged, `gnc_features_set_used (book, GNC_FEATURE_BUDGET_UNREVERSED)` will set the feature and allow GnuCash to use a different code path for budgeting module and reports.
 
How do you think that such a thing would be necessary with this pr ?

rhabacker:
I found a case, when on deleting the opening balance account the transactions are moved to another account. That account will not have the equity-type flag set, which will create a new opening balance account.

jralls:
> How do you think that such a thing would be necessary with this pr ?

I explained that in my review comment on Sept. 17. This is a show-stopper: This PR cannot be merged without GNC_FEATURE protection.

rhabacker:
> I explained that in my review comment on Sept. 17. This is a show-stopper: This PR cannot be merged without GNC_FEATURE protection.

I tried to map the design pattern mentioned for #585 to this PR, and here is what I understand

> current 4.x can already process books with the function budget_unreversed 

You want me to use and set the value 'opening balance' from the slot 'equity type' and show the new checkbox in the dialog account only if gnc_features_check_used (book, GNC_FEATURE_EQUITY_TYPE_OPENING_BALANCE) returns true ?

> but there are no current books in use because the feature is not yet set.

This would require splitting the current commits
```
Add script to create slots with the key 'equity-type' and the value 'opening-bal...
add slot 'equity-type' with the value 'opening-balance' to the account templates
Add support for the flag for opening balance sheet accounts
+ Follow-up commits, which are merged into the latter commits
```
in 
```
Add script to create slots with the key 'equity-type' and the value 'opening-bal...
Add support for the flag for opening balance accounts
+ Correction for gnc_features_check_used (Book, GNC_FEATURE_EQUITY_TYPE_OPENING_BALANCE) 
```
which are merged with this pr and a second pr with the following commits 
```
Add slot "Share type" with the value "Opening balance" to the account templates
Activate feature GNC_FEATURE_EQUITY_TYPE_OPENING_BALANCE
```
to activate this support sometime in the future ?

If yes, when should this happen ?

Just to understand the code: After merging the second pr, gnucash files saved with the new gnucash version are no longer readable for older gnucash versions ?  I have found a corresponding hint in the code related to credit_note.

christopherlam:
> to activate this support sometime in the future ?

Yes

> If yes, when should this happen ?

When you are 100% absolutely-bonkers sure that no one will *ever* complain about the changes being done. When 100% coverage of tests is complete.  When you are able to watch the mailing list and answer questions to defend the changes. When you can predict that absolutely no one will oppose the change and request a revert. And probably some more ;)

> Just to understand the code: After merging the second pr, gnucash files saved with the new gnucash version are no longer readable for older gnucash versions ? I have found a corresponding hint in the code related to credit_note.

Correct; older Gnucash will refuse to load book with feature.

christopherlam:
... and documentation is amended too!

rhabacker:
> This would require splitting the current commits
> 
After rereading the thread in detail I recognized that @pali already confirmed this proposal in parts:

> The first phase recognizes the feature flag and can find opening balance accounts using gnc_account_lookup_by_opening_balance; 

> the second phase in a later release can set the event-type KVP and the feature flag.

jralls:
>>If yes, when should this happen ?

>When you are 100% absolutely-bonkers sure that no one will ever complain about the changes being done. When 100% coverage of tests is complete. When you are able to watch the mailing list and answer questions to defend the changes. When you can predict that absolutely no one will oppose the change and request a revert. And probably some more ;)

@christopherlam was going a little over the top there . The scenario we're trying to protect against is a user who upgrades to the GnuCash with the feature, finds a (probably unrelated) problem, and decides to go back to the previous version only to find that their file won't open any more. That means that the "uses" phase must be at least one release before the "generates" phase. In the case of a substantial change, especially in UI, we'd want the "generates" phase to go to the master branch so that it would wait until the next major release, i.e. 5.0 in this case. I don't think that this change needs to wait that long.

Your "feature" commit needs one more piece: In addition to defining the feature you need to add it to the `known_features` array in `libgnucash/engine/gnc-features.c` so that `gnc_features_test_unknown()` will pass at startup.

rhabacker:
> Your "feature" commit needs one more piece: In addition to defining the feature you need to add it to the `known_features` array in `libgnucash/engine/gnc-features.c` so that `gnc_features_test_unknown()` will pass at startup.

Added to the updated patches

> I don't think that this change needs to wait that long.
> 
The version need to be specified as it is used in the `known_features` array

> That means that the "uses" phase must be at least one release before the "generates" phase

Because in the maint branch the next release will be 4.2 and therefore I set the minimum version to 4.3

To activate the feature I would add an additional pull request with the following change. If there is a better place, please let me know.
```
gnc_plugin_page_register_new_common (GNCLedgerDisplay* ledger)
.... 
+    // added for version 4.3 onwards
+    if (!gnc_using_equity_type_opening_balance_account (gnc_get_current_book()))
+        gnc_set_use_equity_type_opening_balance_account (gnc_get_current_book());
+
```

jralls:
We released [4.2](https://github.com/Gnucash/gnucash/releases/tag/4.2) on 27 September, 4.3 is next, see the [Release Schedule](https://wiki.gnucash.org/wiki/Release_Schedule).

jralls:
OK, this looks pretty good, but we need that last commit broken out into a separate PR to be merged in January after 4.3 is released.

jralls (approved):

rhabacker:
> OK, this looks pretty good, but we need that last commit broken out into a separate PR to be merged in January after 4.3 is released.
Nice to hear :)
The patches have been rebased and the mentioned commit has been moved to https://github.com/Gnucash/gnucash/pull/820, which needs to be updated after the merge of this pr.

jralls:
Thanks!

gnucash/gnome/gnc-plugin-page-account-tree.c
```diff
@@ -1669,6 +1669,11 @@ gnc_plugin_page_account_tree_cmd_delete_account (GtkAction *action, GncPluginPag
           gnc_account_n_children (account)))
     {
         do_delete_account (account, NULL, NULL, NULL);
+
+        /* update opening balance account */
+        gnc_find_or_create_equity_account (gnc_account_get_root(account),
+                                           EQUITY_OPENING_BALANCE,
+                                           xaccAccountGetCommodity (account));
```
jralls, line 1676: @rhabacker Why would you create an opening balance account when deleting the account that might or might not use it?

rhabacker: If the account to be deleted is the only opening balance account, I think a default account should be created.

jralls:
a, that's not what this snippet does, and b, why should we force an opening balance account on a user who has no opening balance transactions?

rhabacker:
> a, that's not what this snippet does,

My reading is as follows: After an account is deleted, make sure there is an equity account. If the deleted account was not the equity account, nothing happens. In the other case, an equity account is created.

> b, why should we force an opening balance account on a user who has no opening balance transactions?

Are you suggesting that a "default" equity account be created when the first transaction is created?  
But doesn't this require an equity account to have it in the list of selectable accounts?

christopherlam:
> > a, that's not what this snippet does,
> 
> My reading is as follows: After an account is deleted, make sure there is an equity account. If the deleted account was not the equity account, nothing happens. In the other case, an equity account is created.

But why??? If I'm deleting an account, it is empty. If it's not empty I expect the UI to ask the destination account for all existing account splits. No opening balance machinations are expected.

jralls:
@rhabacker Your reading is incorrect. There's no test of the deleted account to see if it's an O.B. account. @christopherlam is correct that the user will have had to designate a different account to receive the splits.  If the deleted account *is* an O.B. account then the account receiving the splits could a) be tested to ensure that it is also an Equity account and b) be set to be the O.B. account for the currency. Creating a new account with no splits in it is wrong.

PR #763, opened 2020-07-30
ACB tool

christopherlam:
https://bugs.gnucash.org/show_bug.cgi?id=797796
ACB calculator, with more explicit calculations than the advanced-portfolio. Very early draft.
image


christopherlam:
Merged into #818

PR #764, opened 2020-08-01
Add Python example export_account_totals.py

xuhdev:
This example exports acount totals of all accounts into a CSV file.

This has been asked before: https://money.stackexchange.com/questions/111786/exporting-account-totals-in-gnucash

### DIFF
--- a/bindings/python/example_scripts/export_account_totals.py
+++ b/bindings/python/example_scripts/export_account_totals.py
@@ -1,0 +1,69 @@
+#!/usr/bin/env python3
+
+# export_account_totals -- export the account totals of all accounts.
+
+# Copyright (C) 2020 Hong Xu <hong@topbug.net>
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License as
+# published by the Free Software Foundation; either version 2 of
+# the License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# @author Hong Xu <hong@topbug.net>
+
+# Usage:
+#
+#    python3 export_account_totals.py /path/to/gnucash_book_file > output.csv
+
+import copy
+import sys
+
+from gnucash import GncNumeric, Session, SessionOpenMode
+
+
+def get_all_sub_accounts(account, names=[]):
+    "Iterate over all sub accounts of a given account."
+
+    for child in account.get_children_sorted():
+        child_names = names.copy()
+        child_names.append(child.GetName())
+        yield child, '::'.join(child_names)
+        yield from get_all_sub_accounts(child, child_names)
+
+
+def to_string_with_decimal_point_placed(number: GncNumeric):
+    """Convert a GncNumeric to a string with decimal point placed if permissible.
+    Otherwise returns its fractional representation.
+    """
+
+    number = copy.copy(number)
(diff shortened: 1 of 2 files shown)


bindings/python/example_scripts/export_account_totals.py
```diff
@@ -0,0 +1,50 @@
...
+    print('Name,Commodity,Totals,Totals (USD)')
+    with gnucash.Session(sys.argv[1], SessionOpenMode.SESSION_READ_ONLY) as session:
+        commodity_table = session.book.get_table()
+        USD = commodity_table.lookup("ISO4217", "USD")
+        root_account = session.book.get_root_account()
+        for account, account_name in get_all_sub_accounts(root_account):
+            print(','.join([str(account_name),
+                            account.GetCommodity().get_mnemonic(),
+                            str(round(float(account.GetBalance().to_fraction()), 2)),
+                            str(round(float(account.GetBalanceInCurrency(USD, True).to_fraction()), 2))
```
jralls, line 49: Don't ever use floating point in financial software. Use `gnc_numeric_to_string`.

xuhdev: If I use `gnc_numeric_to_string`, it outputs things like `12345/100`, which I don't think is what most people want. Do you know whether there's a helper that converts a gnc_numeric to a decimal-looking string?

xuhdev: I added my own helper to the script. Feel free to let me know if there is already a more robust helper within gnucash!

bindings/python/example_scripts/export_account_totals.py
```diff
@@ -0,0 +1,69 @@
...
+
+    nominator = str(number.num())
+    point_place = str(number.denom()).count('0')  # How many zeros in the denominator?
+    if point_place == 0:
+        return nominator
+
+    if len(nominator) <= point_place:  # prepending zeros if the nominator is too short
+        nominator = '0' * (point_place - len(nominator)) + nominator
+
+    return '.'.join([nominator[:-point_place - 1], nominator[-point_place:]])
```
jralls, line 55: That's a rather naive algorithm. What if the result is 46398756401/179207364?

xuhdev: It output its fractional form (line 45)

jralls: No, it doesn't. There's a 0 in the denominator.

xuhdev: What do you mean? I tried the following and it should return its fractional form on line 45:

```
In [1]: import gnucash

In [2]: a=gnucash.GncNumeric(46398756401, 179207364)

In [3]: a
Out[3]: <gnucash.gnucash_core.GncNumeric at 0x7fce3826e470>

In [4]: str(a)
Out[4]: '46398756401/179207364'

In [5]: a.to_decimal(None)
Out[5]: False

In [6]: str(a)
Out[6]: '46398756401/179207364'
```

jralls: Sorry, you're right. I was looking at the wrong internal function.

To answer your earlier question about a better helper, there is: `xaccPrintAmount`, but it seems not to be exposed in python.

xuhdev: Got it, thanks. Looks like we would also have to expose `GNCPrintAmountInfo` before exposing `xaccPrintAmount`. I'm not very familiar with the Python binding exposure development though. Does this script look good for now, before `xaccPrintAmount` is exposed?

jralls: Yes.

jralls:
Thanks.

PR #766, opened 2020-08-08
[dialog-preferences.c] Warn on account separator only once

kohtala:
The validate callback may be called by Gtk repeatedly and if the account separator is not valid the warning dialog may be shown many times with no way to get it closed.

To reproduce, go to Edit / Preferences and select Accounts. Change the Separator Character to something that is in an account name. Select other tab, eg. Business, and come back to Accounts. I can not get away from the warning dialog as there is always another dialog when I close one. Sometimes when I open without the warning, change the tab, it does not occur, but after some retries it reproduces reliably.

Prevent reopening the dialog again until the preference is changed.

I checked on https://bugs.gnucash.org/ and the other bugs on this preference did not seem to apply.

Quickly looking I did not find other fields that would have same problem.

### DIFF
--- a/gnucash/gnome-utils/dialog-preferences.c
+++ b/gnucash/gnome-utils/dialog-preferences.c
@@ -117,6 +117,8 @@ typedef struct addition_t
  *  structures. */
 GSList *add_ins = NULL;
 
+static gboolean invalid_account_separator_warned = FALSE;
+
 static gchar *gnc_account_separator_is_valid (const gchar *separator,
                                               gchar **normalized_separator)
 {
@@ -181,6 +183,8 @@ gnc_account_separator_pref_changed_cb (GtkEntry *entry, GtkWidget *dialog)
     else
         gtk_widget_hide (GTK_WIDGET(image));
 
+    invalid_account_separator_warned = FALSE;
+
     g_free (separator);
 }
 
@@ -189,16 +193,22 @@ gboolean
 gnc_account_separator_validate_cb (GtkEntry *entry, GdkEvent *event, GtkWidget *dialog)
 {
     gchar *separator;
-    gchar *conflict_msg = gnc_account_separator_is_valid (gtk_entry_get_text (entry), &separator);
+    gchar *conflict_msg;
 
-    /* Check if the new separator clashes with existing account names */
+    if (!invalid_account_separator_warned) {
+        conflict_msg = gnc_account_separator_is_valid (gtk_entry_get_text (entry), &separator);
 
-    if (conflict_msg)
-    {
-        gnc_warning_dialog (GTK_WINDOW (dialog), "%s", conflict_msg);
-        g_free ( conflict_msg );
+        /* Check if the new separator clashes with existing account names */
+
+        if (conflict_msg)
+        {
+            gnc_warning_dialog (GTK_WINDOW (dialog), "%s", conflict_msg);
+            g_free ( conflict_msg );
+            invalid_account_separator_warned = TRUE;
+        }
+        g_free (separator);
     }
-    g_free (separator);
+
     return FALSE;
 }
 
@@ -770,7 +780,7 @@ file_chooser_selected_cb (GtkFileChooser *fc, gpointer user_data)
     else
         gnc_assoc_pref_path_head_changed (GTK_WINDOW(gtk_widget_get_toplevel (
                                           GTK_WIDGET(fc))), old_path_head_uri);
-    
+
     g_free (old_path_head_uri);
     g_free (folder_uri);
 }


gjanssens:
Thank you for your pach. Unfortunately it defeats the purpose of the warning.

The intention of the warning dialog is to prevent the user from closing the preferences dialog as long as an invalid account separator is entered. With your patch, when the user clicks close, the warning will be shown only the first time. The second time the user clicks close, gnucash kindly closes the dialog window and will start using the invalid account separator.

A way to properly handle this would be to run the validation test that emits the warning dialog only when the user hits the close button, but run it each time the close button is hit to prevent the dialog from being closed as long as there's invalid input somewhere.

The warning pop-up may not be strictly needed at the time the bad value is entered. There is already a warning icon appearing next to the field in that case, although with my gtk theme it's a bit weak.

Lastly, while testing I found that the validation test is not run at all when the user hits the window manager's close button (the "cross" icon at the top left or right of the window). While you're at this, it may be a big improvement if that case would be caught as well.

Thanks!

gnucash/gnome-utils/dialog-preferences.c
```diff
@@ -189,16 +193,22 @@ gboolean
 gnc_account_separator_validate_cb (GtkEntry *entry, GdkEvent *event, GtkWidget *dialog)
 {
     gchar *separator;
-    gchar *conflict_msg = gnc_account_separator_is_valid (gtk_entry_get_text (entry), &separator);
+    gchar *conflict_msg;
```
gjanssens, line 196: When you later test for the existence of this message, it's safer to explicitly initialize conflict_msg to NULL. Not all compilers to this by default so you may run into an uninitialized variable warning here (at best) or a NULL dereference at runtime (at worst).

kohtala:
> Thank you for your pach. Unfortunately it defeats the purpose of the warning.

Thanks for fast review.

I studied and considered this a bit further. I first thought I write here a long reply, but then realized it would belong to Bugzilla and you practically already agree with what I was getting into at https://bugs.gnucash.org/show_bug.cgi?id=797320#c7.

Even while a solution to a problem is nice, I much prefer not to have a problem at all.

Currently this patch does fix a problem where the user will be stuck and has to kill the program to get out. A good solution would be to get rid of the link between preference and account names so we could accept any value. So motivation to put time into finding a great solution to failed validation is not high if we could have no validation fail.

> The intention of the warning dialog is to prevent the user from closing the preferences dialog as long as an invalid account separator is entered. With your patch, when the user clicks close, the warning will be shown only the first time. The second time the user clicks close, gnucash kindly closes the dialog window and will start using the invalid account separator.

I see a small security issue here. Finding an acceptable setting before letting the user leave can be very difficult for the user if the file has all kinds of characters in account names. Someone could just send a file with all characters in account names and user is stuck without being able to close the preferences dialog. I think the user should always have the option to leave without making a change. And as it is about changing his preferences just because he has a file open, I find it rather invasive even when the file is not malicious. He might prefer to close the file instead.

A side note, that I had an idea a partial help would be if the separator could be a string, for example " / ", with the spaces. I've been trying to import accounting from another program and there are accounts that say for example "tuotteet/tavarat". It should be no problem to have that with " / " separator. There are also accounts with "Alv:n" in it (the colon is part of Finnish language) and it would not be a problem if the separator was " : ". When I tested to see what would happen, I noticed it also allows you type for example " & " as the value of separator character. It gets saved, but does cause a bunch of problems.

> Thanks!

I appreciate the work you've put into gnucash. I hope to be helpful.

I don't have much time for this and would hope to be able to continue importing my data and work on the Finnish language translations instead. But if no one else steps in and declares they want to work on these, I can try to find some time at some point to make minimal changes to get this one problem solved.

gnucash/gnome-utils/dialog-preferences.c
kohtala: I agree a `=NULL` is safe and clean. All paths however have the value of `conflict_msg` set before use, so there is no possibility of NULL dereference and no compiler will warn here. However the variable `separator` is suspect as generally there is no promise compiler should know if the `gnc_account_separator_is_valid` will initialize it. It does, so there is no g_free of uninitialized pointer.

gjanssens: Right. I got confused by reading the diff instead of the final code.

However in that case, my recommendation changes to: only define `conflict_msg` in the scope you use it, which s inside the `if (!invalid_separator_warned)` block.

Additionally, please use the same coding style as already in the file. That is put the `{` on a new line by itself.

gjanssens:
> > Thank you for your pach. Unfortunately it defeats the purpose of the warning.
> 
> Thanks for fast review.
> 
> I studied and considered this a bit further. I first thought I write here a long reply, but then realized it would belong to Bugzilla and you practically already agree with what I was getting into at https://bugs.gnucash.org/show_bug.cgi?id=797320#c7.
> 
> Even while a solution to a problem is nice, I much prefer not to have a problem at all.
> 
I fully agree, though I'm currently not in a position to make such a big change myself.

> Currently this patch does fix a problem where the user will be stuck and has to kill the program to get out.

For the record I tried to reproduce, but I can't. When I switch tabs and go back, I always get one warning dialog per tab change.

> I see a small security issue here. Finding an acceptable setting before letting the user leave can be very difficult for the user if the file has all kinds of characters in account names. Someone could just send a file with all characters in account names and user is stuck without being able to close the preferences dialog. I think the user should always have the option to leave without making a change. And as it is about changing his preferences just because he has a file open, I find it rather invasive even when the file is not malicious. He might prefer to close the file instead.

You're about to open a can of worms... In fact you have just described the reason I believe the account separator should be a per book configurable option, not user-general. Unfortunately fixing *that* fits in a much bigger project of re-evaluating all preferences and options.

> A side note, that I had an idea a partial help would be if the separator could be a string, for example " / ", with the spaces. I've been trying to import accounting from another program and there are accounts that say for example "tuotteet/tavarat". It should be no problem to have that with " / " separator. There are also accounts with "Alv:n" in it (the colon is part of Finnish language) and it would not be a problem if the separator was " : ". When I tested to see what would happen, I noticed it also allows you type for example " & " as the value of separator character. It gets saved, but does cause a bunch of problems.

> I don't have much time for this and would hope to be able to continue importing my data and work on the Finnish language translations instead. But if no one else steps in and declares they want to work on these, I can try to find some time at some point to make minimal changes to get this one problem solved.

Please do.

I think for a minimal change you can do as I suggested - only pop up the warning when the user tries to close the dialog (either via the close button or via the window manager 'X' button). It would be slightly more user-friendly if the dialog would propose two options
1. reset the account separator to its original state and close the preferences anyway
2. keep the preferences open and let the user choose another account separator

This extra bit does require to store the current separator and make sure not to block on that one (even if it's not usable) to allow the user a no-change way out.

If you do find it important enough you're obviously also welcome to go for the bigger change and make gnucash smarter at full-account-name-to-id detection as suggested in the bug you referenced.

gnucash/gnome-utils/dialog-preferences.c
kohtala: Ok. A few other functions like gnc_prefs_build_widget_table and gnc_prefs_find_page seemed to use style of declaring all variables at start of function, so I followed that. The brace is just a mistake of too much use like that.

gjanssens: Heh, the old code is not always well-formed unfortunately. We're trying to do better these days, but there are still plenty of ugly bits around...

Bob-IT:
I have implemented the minimum changes required so am closing this PR.

kohtala:
Thanks. I'll see the new solution once I get back to gnucash.

PR #768, opened 2020-08-10
handle report export in C rather than guile

christopherlam:
Previously, exporting CSV/TXF would call renderer with args: `scm_call_3 (export_thunk, priv->cur_report, choice, file_scm);` -- reports are changed to accept `SCM_BOOL_F` instead of `file_scm`, thereby triggering a different path to return the file contents as an SCM string instead.

C code must now be changed to accept the SCM string and write to file. gnucash-cli code is so far unchanged. Tested using GST:gb: and TXF:us: reports; XML:de: export untested-- someone else must test using appropriate datafile. @jralls may take over at leisure...


christopherlam:
Later commits in this branch remove the filename handling in guile completely; not sure if it's worth while keeping the guile file io code. C and CPP must be modified to handle the file io.

christopherlam:
This also means that `gnucash-cli -R run --export-type TXF datafile.gnucash` does not need to specify an outputfile... can dump to console.

christopherlam:
@jralls not sure which C file IO api to use -- `g_fopen`?

jralls:
Can't you just copy the code from my html output?

christopherlam:
> Can't you just copy the code from my html output?

But this one uses C++ and gnc-plugin-page-report.c (IIRC) must also write to file

jralls:
Ah, OK. Yes, a bit too soon to start converting the gnome code. I think [g_file_set_contents](https://developer.gnome.org/glib/stable/glib-File-Utilities.html#g-file-set-contents) will save a lot of boilerplate work. File ops aren't guaranteed to work, especially on Windows, so be sure to check for an error and pop a gtk_message_dialog if it fails.

On that point, memo to self: the cli code should grow some error handling too.

christopherlam:
Note I don't use taxtxf and can't fully test it.

christopherlam:
Small dilemma: there are many moving parts that expect the renderer to return an html-document object.

My proposal to return the CSV string instead is a nice goal, but must also handle error output too. This means the renderer will return either (1) html-document, (2) the CSV string, or (3) a `(list errormessage)`. This becomes rather awkward; see trep-engine.scm changes -- all cond branches which return document must be modified to potentially return a string.

Another idea: the renderer *always* returns an html-document, and a successful CSV/TXF export will _store_ the resulting string into the html-document record in a dedicated field, and the C/C++ caller will retrieve CSV/TXF from the renderer output. This is much cleaner. But also sounds like a clever hack.

I'm leaning towards latter solution...?

christopherlam:
I think this is ready for review -- removes file I/O from guile report code, handles file IO in C or C++, also allows CSV/TXF output to console with CLI.

christopherlam:
I have no data suitable for taxtxf nor taxtxf_de testing

christopherlam:
Ok I can remove the #f argument. This branch has several approaches to solving this problem, so, may be best to look at the overall patch instead of individual commits. I'll fixup each file into its own commit eventually. If we don't need to handle old custom reports then life is easier but we'll need to document it thoroughly.

christopherlam:
Last few commits have removed the backward compatibility path for `export-thunk`.

gnucash/gnome/gnc-plugin-page-report.c
```diff
@@ -1689,17 +1689,6 @@ gnc_plugin_page_report_export_cb( GtkAction *action, GncPluginPageReport *report
...
-        {
-            /* compatibility path -- old report which does not effect
-               export-string and export-error during export code- call
-               with filepath */
-            SCM file = scm_from_locale_string (filepath);
-            scm_c_eval_string ("(issue-deprecation-warning \"Old report \
-with export-thunk encountered. Please upgrade report to ignore filename \
-and sets html-document export-string and/or export-error instead.\")");
-            scm_call_3 (export_thunk, priv->cur_report, type, file);
-        }
```
jralls: Shouldn't there still be an `else` clause that raises an error instead of silently failing, just to protect developer sanity when one messes up an `export-thunk` so that it returns neither a string nor an error?

gnucash/gnucash-commands.cpp
```diff
@@ -192,11 +192,9 @@ scm_run_report (void *data,
     auto report = scm_from_locale_string (args->run_report.c_str());
     auto type = !args->export_type.empty() ?
                 scm_from_locale_string (args->export_type.c_str()) : SCM_BOOL_F;
-    auto file = !args->output_file.empty() ?
-                scm_from_locale_string (args->output_file.c_str()) : SCM_BOOL_F;
 
-/* dry-run? is #t: try report, check validity of options */
-    if (scm_is_false (scm_call_3 (check_report_cmd, report, type, file)))
+    /* dry-run? is #t: try report, check validity of options */
```
jralls, line 196: This comment no longer makes sense. My fault for leaving it in.

gnucash/gnucash-commands.cpp
```diff
@@ -213,12 +211,34 @@ scm_run_report (void *data,
...
+            }
+            else
+            {
+                std::cout << output << std::endl;
+            }
+        }
+        else if (scm_is_string (export_error))
+        {
+            auto err = scm_to_utf8_string (export_error);
+            std::cout << err << std::endl;
```
jralls, line 239: The error report should go to `std::cerr` in case the user is piping stdout.

jralls:
A few minor things in-line, after which I think it's good to merge.

gnucash/gnome/gnc-plugin-page-report.c
christopherlam: Correct. I'll amend. Also I'll handle another condition where the renderer doesn't return an html-document object.

PR #769, opened 2020-08-11
Maint cli webserver

christopherlam:
Gnucash Report Webserver :+1: 

image


fellen:
I fear, many users dislike a webserver with their financial data for security reasons.

christopherlam:
Then there's https://fava.pythonanywhere.com/example-beancount-file/income_statement/

jralls:
I don't think running an http server should be part of GnuCash.  It's is an open invitation for remote exploitation since it is trivially easy to create a malicious web page that sends requests to localhost from the browser.  I doubt very much that the Guile team puts much thought or effort into the web server's security, but that shouldn't be seen as an opening to go look for a different implementation.

christopherlam:
This is POC for now... May be a private project which uses libgnucash in the future, but can also harness existing report code base.

PR #772, opened 2020-08-16
Etype might have not been initialized when used

xuhdev:
It is used here:

    g_strdup_printf (_("Entity type does not match %s: %s"), etype, location);

But previous it might not have been initialized.


christopherlam:
Thanks! Applied your commit in f662935

PR #773, opened 2020-08-17
L10N: nb: partial updated translation

jeblad:
Only a partial update, way too much to do in one batch! The po-file now has 2793 translations of 5573 strings, or 50.12 % of the strings are now translated! ;)

Perhaps a slight warning: A lot of strings are translated with the fuzzy solution, and many of them are really long way off. Some are so far off I wonder if users may run into problems due to this. I have no solutions though, it is only an observation from the (mis)translated strings.

### DIFF
--- a/po/nb.po
+++ b/po/nb.po
@@ -11,7 +11,7 @@ msgstr ""
 "Report-Msgid-Bugs-To: https://bugs.gnucash.org/enter_bug.cgi?"
 "product=GnuCash&component=Translations\n"
 "POT-Creation-Date: 2020-06-13 20:45+0200\n"
-"PO-Revision-Date: 2018-02-16 13:42+0100\n"
+"PO-Revision-Date: 2020-08-17 18:49+0200\n"
 "Last-Translator: John Erling Blad <jeblad@gmail.com>\n"
 "Language-Team: Norwegian <>\n"
 "Language: no\n"
@@ -19,7 +19,7 @@ msgstr ""
 "Content-Type: text/plain; charset=UTF-8\n"
 "Content-Transfer-Encoding: 8bit\n"
 "Plural-Forms: nplurals=2; plural=n != 1;\n"
-"X-Generator: Poedit 2.0.4\n"
+"X-Generator: Poedit 2.3\n"
 
 #: bindings/guile/commodity-table.scm:36
 msgid "ALL NON-CURRENCY"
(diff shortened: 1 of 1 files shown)


po/nb.po
```diff
@@ -31681,9 +30911,8 @@ msgid "Transaction Voided"
 msgstr "Transaksjonen er annullert"
```
fellen, line 30912: Below are the [Orphaned Records](https://wiki.gnucash.org/wiki/Translation#Orphaned_Records), which you can remove.

po/nb.po
```diff
@@ -9104,6 +8833,7 @@ msgstr ""
 "Kjartan Maraas <kmaraas@gnome.org>, 2000. \n"
 "Tor Harald Thorland <tortho@strigen.com>, 2005-2006.\n"
 "Sigve Indregard <sigve.indregard@gmail.com>, 2006.\n"
+"John Erling Blad <jeblad@gmail.com>, 2020.\n"
```
fellen, line 8836: `"John Erling Blad <jeblad@gmail.com>, 2018-2020.\n"` as far as I can see.
BTW I would reorder the list: current users are more interested in the current translator than the history. See "Special strings" in [Adjust the header and special strings](https://wiki.gnucash.org/wiki/Translation#Adjust_the_header_and_special_strings).

jeblad: Some projects are very picky with the ordering! ;)

po/nb.po
jeblad: Always some… ;)

jeblad:
Accelerator marks does not translate well into Norwegian. Without a list of which strings goes together it isn't easy to create meaningful accelerators without collisions.

The fuzzy-matching does not work very well, and I don't think it should be used. Usually fuzzy-matching fails badly as soon as the strings are more than a few words. There are a lot of strings now that needs checking, some of them are quite meaningless. Some are opposite of what they should be, and some describe other features altogether. I'm not even sure part of the application makes sense now, transactions, lots, and other features are messed up.

I did not update the pot file, there were so much outstanding translations already. (Nearly 2000 additional fuzzy-strings, that would be way too much for me right now. I doubt I do more than around two-hundred strings a day.)

fellen:
> I did not update the pot file, there were so much outstanding translations already. (Nearly 2000 fuzzy-strings, that would be way too much for me right now. I doubt I do more than around two-hundred strings a day.)
> 
> (I must probably make a PR with additional translations before I start using a new pot-file…)

I did it for you in commit 2dd576f. just rebase your work on it.

fellen:
> Accelerator marks does not translate well into Norwegian. Without a list of which strings goes together it isn't easy to create meaningful accelerators without collisions.

Build the program with your translation. When running the program hold down the [Alt] key to see them. Then compare the entries inside a sub menu.
While recent Qt versions claim to generate them automaticaly, for GTK3 we have to enter them manually.

fellen:
> The fuzzy-matching does not work very well, and I don't think it should be used. Usually fuzzy-matching fails badly as soon as the strings are more than a few words. There are a lot of strings now that needs checking, some of them are quite meaningless. Some are opposite of what they should be, and some describe other features altogether. I'm not even sure part of the application makes sense now, transactions, lots, and other features are messed up.

There are different reasons, why a string is marked fuzzy:
1. We fixed a typo (or more) in the msgID. In case of a typo, your translation should often not be affected. If we changed more, apply the change.
2. For a new string, msgmerge made a (often not very smart) guess, often only containing the first keyword. You will have to review and add the rest. Keep in mind, the gettext tools are very simple.

PR #774, opened 2020-08-18
Multi-account selection and Delete

jeanlaroche:
This is an ease-of-use improvement, especially for new users. Most of us don't typically change our accounts once our books are setup I imagine, but new users do.

Trying to keep this message brief.

Somebody in the past must have looked at multi-account actions because there already was functioning (but never called) code to get the list of selected accounts.

Only the Delete action is currently available when multiple accounts are selected in the account page. But I will follow up with another PR that implements the multi-account edit action.

- GC used to delete accounts on the spot if there was no transactions or subaccounts in them. I removed that: you always have to confirm.
- Once you've chosen what to do with the transactions and subbacounts, I check that the selected actions are compatible with all accounts selected for delete. I ask for confirmation, then I delete all accounts.
- Moving transactions used to be allowed only to an account of the same type as the deleted account. I'm not sure why this is strictly enforced, but I removed it as it does not seem warranted. I could add a warning to tell the user.

### DIFF
--- a/gnucash/gnome/gnc-plugin-page-account-tree.c
+++ b/gnucash/gnome/gnc-plugin-page-account-tree.c
@@ -92,7 +92,7 @@ typedef struct _delete_helper
 #define STATE_SECTION "Account Hierarchy"
 
 #define DELETE_DIALOG_FILTER       "filter"
-#define DELETE_DIALOG_ACCOUNT      "account"
+#define DELETE_DIALOG_ACCOUNT_LIST "account_list"
 #define DELETE_DIALOG_TRANS_MAS    "trans_mas"
 #define DELETE_DIALOG_SA_MAS       "sa_mas"
 #define DELETE_DIALOG_SA_TRANS_MAS "sa_trans_mas"
@@ -146,7 +146,7 @@ static void gnc_plugin_page_account_tree_double_click_cb (GtkTreeView        *tr
 
 static void gnc_plugin_page_account_tree_selection_changed_cb (GtkTreeSelection *selection,
         GncPluginPageAccountTree *page);
-void gppat_populate_trans_mas_list(GtkToggleButton *sa_mrb, GtkWidget *dialog);
+void gppat_populate_trans_mas_list(GtkToggleButton *sa_mrb, GtkWidget *dialog, GList* acct_list);
 void gppat_set_insensitive_iff_rb_active(GtkWidget *widget, GtkToggleButton *b);
 
 /* Command callbacks */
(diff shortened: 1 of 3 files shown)


gnucash/gnome/gnc-plugin-page-account-tree.c
```diff
@@ -1077,8 +1098,24 @@ gnc_plugin_page_account_tree_selection_changed_cb (GtkTreeSelection *selection,
...
         g_return_if_fail(GTK_IS_TREE_SELECTION(selection));
         view = gtk_tree_selection_get_tree_view (selection);
-        account = gnc_tree_view_account_get_selected_account (GNC_TREE_VIEW_ACCOUNT(view));
+        GList* acct_list = gnc_tree_view_account_get_selected_accounts (GNC_TREE_VIEW_ACCOUNT (view));
+        if (acct_list)
+        {
+            Account* parent = NULL;
+            account = acct_list->data;
+            parent = gnc_account_get_parent (account);
+            multi_account = g_list_length (acct_list) > 1;
```
christopherlam, line 1107: or `multi_account = (!acct_list->next);` avoids O(N) g_list_length...

gnucash/gnome/gnc-plugin-page-account-tree.c
```diff
@@ -1287,28 +1329,29 @@ gppat_populate_gas_list(GtkWidget *dialog,
                         GNCAccountSel *gas,
                         gboolean exclude_subaccounts)
 {
-    Account *account;
     GList *filter;
+    GList* acct_list = g_object_get_data(G_OBJECT(dialog), DELETE_DIALOG_ACCOUNT_LIST);
 
     g_return_if_fail(GTK_IS_DIALOG(dialog));
     if (gas == NULL)
```
christopherlam, line 1336: You're accessing `dialog` before the argument check `g_return_if_fail`...

```c
GList *acct_list;
g_return_if_fail(GTK_IS_DIALOG(dialog));
acct_list = g_object_get_data(G_OBJECT(dialog), DELETE_DIALOG_ACCOUNT_LIST);
```

gnucash/gnome/gnc-plugin-page-account-tree.c
```diff
@@ -1447,18 +1496,17 @@ account_subaccount (Account* account)
 }
 
 static GtkWidget*
-account_delete_dialog (Account *account, GtkWindow *parent, Adopters* adopt)
+account_delete_dialog (GList* acct_list, GtkWindow *parent, Adopters* adopt)
```
christopherlam, line 1499: function should be renamed to `accountlist_delete_dialog`?

gnucash/gnome/gnc-plugin-page-account-tree.c
```diff
@@ -1556,104 +1601,150 @@ account_delete_dialog (Account *account, GtkWindow *parent, Adopters* adopt)
 
     /* default to cancel */
     gtk_dialog_set_default_response (GTK_DIALOG(dialog), GTK_RESPONSE_CANCEL);
+    if (g_list_length(acct_list) == 1)
```
christopherlam, line 1604: `if (acct_list && !acct_list->next)` to avoid `g_list_length`

jeanlaroche:
Thanks for the comments christopherlam, I've adopted all your suggestions. Good trick about avoiding length().

jralls:
> Somebody in the past must have looked at multi-account actions because there already was functioning (but never called) code to get the list of selected accounts.

If you mean `gnc_tree_view_account_get_selected_accounts`, it's used for the account selectors in reports, budgets, tax info, and Find.

PR #777, opened 2020-08-19
Multi account edit

jeanlaroche:
The PR is meant as a continuation of multi_account_delete.
To make multi-account simpler I only allow multi-account editing of accounts under the same parent (for now).
The difficulty is to present the data to the user when the accounts have mismatched attributes. Generally, the UI indicates a mismatch (see below) and only when the user makes a selection is the mismatch removed and replaced by the user's selection. If the user does not change the mismatched attributes, they're not modified in the accounts.
Details:

- Mismatched strings are easy: show "mismatched".
- Mismatched flags are easy: the GTK button has a "inconsistent" state.
- Mismatched currency: I add a temporary fake currency "MULTIPLE" which is removed as soon as the user edits the currency.
- Mismatched colors: I don't have a good solution for that: the GTK color button does no offer a way to show "inconsistency". At the moment, I show gray, but that's not very good. Ideas welcome.
- Mismatched parents: Impossible given the current restriction (see above). Changing the parent is supported of course.
- Mismatched Account Types:
   . I switch to a multi-selection type view, and show the different types of the accounts as selected (so you'll see more than 1 type in blue)
   . As soon as the user selects a common type, I switch back to single-selection.

The entire thing is complicated by the fact that GTK does not differentiate (something_changed callbacks triggered when the code changes something internally), and (something_changed callback triggered because the user clicked something). Only when the user makes a selection should the mismatch be collapsed, not when callbacks are triggered internally. In addition, GTK does not have a way to temporary disable callbacks.
An example of this is if the user selects a different parent: this typically changes data in the account type tree/view but we do not want these to understood as if the user had made them.

The solution I'm using is a flag in the aw structure to indicate whether callbacks should be on or not.

### DIFF
--- a/gnucash/gnome-utils/dialog-account.c
+++ b/gnucash/gnome-utils/dialog-account.c
@@ -71,6 +71,22 @@ typedef enum
     EDIT_ACCOUNT
 } AccountDialogType;
 
+// Structure for multi-account data.
+typedef struct _MultiMatchInfo
+{
+    gboolean         freeze_cb;
+    gboolean         description_entry_match;
+    gboolean         color_entry_match;
+    gboolean         notes_text_match;
+    gboolean         type_match;
+    gboolean         parent_match;
+    gboolean         commodity_match;
+    gboolean         tax_related_match;
+    gboolean         placeholder_match;
+    gboolean         hidden_match;
+    gnc_commodity*   invalid_commodity;
+} MultiMatchInfo;
+
 typedef struct _AccountWindow
 {
     QofBook *book;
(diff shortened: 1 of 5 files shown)


jralls:
This doesn't seem like a good idea, even without having to deal with the mismatches. Editing accounts at all should be a pretty rare occurrence, what's the use case for batch edits?

jeanlaroche:
> This doesn't seem like a good idea, even without having to deal with the mismatches. Editing accounts at all should be a pretty rare occurrence, what's the use case for batch edits?

When you setup your accounts, it's a lot easier to change things in one shot. For example moving a bunch of expense accounts to a different parent. I remember finding painful to setup my account tree, even using the template... It's basically a ease-of-use improvement.

shaform:
I think it is useful to do batch edit for the account colors.

PR #778, opened 2020-08-21
Fix warnings: missing sentinel in function calls

g4jc:
When compiling with `-Werror`,  GNUCash fails to build due to various `error: missing sentinel in function call`.
This patch resolves the issue by passing the appropriate sentinels.


jralls (changes requested):
Reiterating: re-casting `NULL` from `void*` to `char*` is not the correct solution. Use `nullptr`.

g4jc:
Thanks for the reminder. Passed `nullptr` as requested and confirmed working here.

jralls (approved):

jralls:
Thanks!

PR #779, opened 2020-08-22
Bug 797900  - Crash caused by Quitting while Check and Repair All is …

jeanlaroche:
…running

The account tree page didn't have a "finish" function normally used to verify a page can close.
I added one, along with two flags that indicate whether a scrubbing operation is currently ongoing
and whether we should quit when the scrubbing is done.
The result is: If a user attempts to quit while scrubbing isn't done, an alert pops up asking whether the
user wants to abort the scrub. If so, the scrub is aborted (safely) and GC quits.
If not the app does not quit.

I have to say, I'm not sure this is the right way to do this. In my view, the right way would be:
- Disable the "quit" menu when scrubbing is happening (for some reason gnc_suspend_gui_refresh() does
not cause the quit menu to be grayed) so there's no chance of quitting while scrubbing is ongoing
- If needed, add an abort scrubbing button to the main window. Not sure whether that's desirable or not.

Let me know what you think: is what I have what we need, or would the above be better.

### DIFF
--- a/gnucash/gnome-utils/gnc-main-window.c
+++ b/gnucash/gnome-utils/gnc-main-window.c
@@ -1146,6 +1146,11 @@ gnc_main_window_all_finish_pending (void)
             return FALSE;
         }
     }
+    if (gnc_gui_refresh_suspended ())
+    {
+        gnc_warning_dialog (NULL, "%s", "An operation is still pending, wait for it to complete before quitting ");
+        return FALSE;
+    }
     return TRUE;
 }
 
--- a/gnucash/gnome-utils/gnc-main-window.c
+++ b/gnucash/gnome-utils/gnc-main-window.c
@@ -3712,7 +3712,7 @@ static gboolean
 gnc_quartz_should_quit (GtkosxApplication *theApp, GncMainWindow *window)
 {
     if (gnc_main_window_all_finish_pending())
-        return gnc_main_window_quit (window);
+        return !gnc_main_window_quit (window);
     return TRUE;
 }
 
--- a/gnucash/gnome/gnc-plugin-page-account-tree.c
+++ b/gnucash/gnome/gnc-plugin-page-account-tree.c
@@ -414,6 +414,30 @@ gnc_plugin_page_account_tree_new (void)
 G_DEFINE_TYPE_WITH_PRIVATE(GncPluginPageAccountTree, gnc_plugin_page_account_tree, GNC_TYPE_PLUGIN_PAGE)
 
 static void
+prepare_scrubbing ()
+{
+    gnc_suspend_gui_refresh ();
+    gnc_set_abort_scrub (FALSE);
+}
+
+static gboolean
+finish (GncPluginPage* page)
+{
+    if (gnc_get_ongoing_scrub ())
+    {
+        gboolean ret = gnc_verify_dialog (NULL, FALSE, "%s", "A scrubbing operation is currently pending, do you want to abort it?");
+        if (ret)
+        {
+            gnc_set_abort_scrub (TRUE);
+            gnc_resume_gui_refresh (); // This is so quit does not complain about an ongoing operation.
+            return TRUE;
+        }
+        return FALSE;
+    }
+    return TRUE;
+}
+
+static void
 gnc_plugin_page_account_tree_class_init (GncPluginPageAccountTreeClass *klass)
 {
     GObjectClass *object_class = G_OBJECT_CLASS (klass);
@@ -430,6 +454,8 @@ gnc_plugin_page_account_tree_class_init (GncPluginPageAccountTreeClass *klass)
     gnc_plugin_class->save_page       = gnc_plugin_page_account_tree_save_page;
     gnc_plugin_class->recreate_page   = gnc_plugin_page_account_tree_recreate_page;
     gnc_plugin_class->focus_page_function = gnc_plugin_page_account_tree_focus_widget;
+    gnc_plugin_class->finish_pending = finish;
+
 
     plugin_page_signals[ACCOUNT_SELECTED] =
         g_signal_new ("account_selected",
@@ -1913,7 +1939,7 @@ gnc_plugin_page_account_tree_cmd_scrub (GtkAction *action, GncPluginPageAccountT
 
     g_return_if_fail (account != NULL);
 
-    gnc_suspend_gui_refresh ();
+    prepare_scrubbing ();
 
     window = GNC_WINDOW(GNC_PLUGIN_PAGE (page)->window);
     gnc_window_set_progressbar_window (window);
@@ -1939,7 +1965,7 @@ gnc_plugin_page_account_tree_cmd_scrub_sub (GtkAction *action, GncPluginPageAcco
 
     g_return_if_fail (account != NULL);
 
-    gnc_suspend_gui_refresh ();
+    prepare_scrubbing ();
 
     window = GNC_WINDOW(GNC_PLUGIN_PAGE (page)->window);
     gnc_window_set_progressbar_window (window);
@@ -1962,7 +1988,7 @@ gnc_plugin_page_account_tree_cmd_scrub_all (GtkAction *action, GncPluginPageAcco
     Account *root = gnc_get_current_root_account ();
     GncWindow *window;
 
-    gnc_suspend_gui_refresh ();
+    prepare_scrubbing ();
 
     window = GNC_WINDOW(GNC_PLUGIN_PAGE (page)->window);
     gnc_window_set_progressbar_window (window);
--- a/gnucash/gnome/gnc-plugin-page-register.c
+++ b/gnucash/gnome/gnc-plugin-page-register.c
@@ -1905,6 +1905,25 @@ gnc_plugin_page_register_update_edit_menu (GncPluginPage* page, gboolean hide)
     gtk_action_set_visible (action,  !hide || can_paste);
 }
 
+static gboolean abort_scrub = FALSE;
+static gboolean is_scrubbing = FALSE;
+
+static gboolean
+finish_scrub (GncPluginPage* page)
+{
+    if (is_scrubbing)
+    {
+        gboolean ret = gnc_verify_dialog (NULL, FALSE, "%s", "A scrubbing operation is currently pending, do you want to abort it?");
+        if (ret)
+        {
+            abort_scrub = TRUE;
+            gnc_resume_gui_refresh (); // This is so quit does not complain about an ongoing operation.
+            return TRUE;
+        }
+        return FALSE;
+    }
+    return TRUE;
+}
 
 static gboolean
 gnc_plugin_page_register_finish_pending (GncPluginPage* page)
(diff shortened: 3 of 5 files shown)


jralls:
@jeanlaroche,
It works a bit differently running from the command line or by double-clicking the icon in Finder, but you can see the Apple Notification by control-clicking the Dock icon and selecting Quit.
That said, the notification does appear to work by putting an event on the CFRunLoop so it's fired from the GMainLoop in the _nextEventMatchingEventMask call in gdkeventloop-quartz's `poll_func` so it should still interleave OK with the occasional stops to update the progress bar (and does, when I test with this PR).

I don't think that you can disable the Quit menu in macOS, it's controlled by AppKit. There's also the Dock Menu Quit and Quit from Activity Monitor. Returning `FALSE` from finish_pending is the right way to prevent 

And yes, this probably does have the potential to crash in any operation that doesn't update the plugin page's `finish_pending`.

Including, BTW, `gnc_plugin_page_register_finish_pending` which can also do check and repair. I got it to crash by quitting after starting a C&R from a register tab.

Another problem: The close order is messed up. The main window closes with the save or discard changes dialog still unanswered. Aside from being a little scary, it means that the book's state file is empty so when you reopen the book there are no tabs, not even an Account Tree.

christopherlam:
To @jeanlaroche the qif-importer includes guile code, during lengthy imports, to call `(update-progress)` https://github.com/Gnucash/gnucash/blob/c210ceb3cbfa422effe2a33f8b87d8f6887943ca/gnucash/import-export/qif-imp/qif-to-gnc.scm#L270 which calls `(qif-import:check-pause ...)` to detect UI cancel/pause events. Your suggestion is very similar.

jeanlaroche:
> And yes, this probably does have the potential to crash in any operation that doesn't update the plugin page's `finish_pending`.
> 
> Including, BTW, `gnc_plugin_page_register_finish_pending` which can also do check and repair. I got it to crash by quitting after starting a C&R from a register tab.

Ah, I have to take a look at that, I missed it. Note however, that I added a further protection against quitting in case a page does not implement finish() the right way. The main loop check whether refresh_ui is on or off before quitting. If it's off, it assumes an operation is ongoing, issues an alert and does not quit.
 
> Another problem: The close order is messed up. The main window closes with the save or discard changes dialog still unanswered. Aside from being a little scary, it means that the book's state file is empty so when you reopen the book there are no tabs, not even an Account Tree.

Yes, I was wondering what that came from. It's because I was attempting to quit the app by destroying the main window (hoping to trigger the chain of checks before it actually happened). That didn't work apparently. So now I'm closing the app by using the callback function which I've exposed in gnc-main-window.h

jralls:
@jeanlaroche The premise here is that the shutdown logic is already engaged, you need only ask the user if they'd rather quit the scrub and the program, in which case `finish_pending` should stop the scrub and return `TRUE` or finish the scrub and quit later in which case it should return `FALSE`.

jeanlaroche:
> @jeanlaroche The premise here is that the shutdown logic is already engaged, you need only ask the user if they'd rather quit the scrub and the program, in which case `finish_pending` should stop the scrub and return `TRUE` or finish the scrub and quit later in which case it should return `FALSE`.

Yes, that's what I was hoping to do! Alas this does not work because everything is in the same thread: finish_pending should "stop the scrub" but finish_pending is called from the scrub thread (remember, it's called by the display_progress() which pumps pending events), so it has no way to "wait" for the scrub to be done. It sets the "abort the scrub" flag, but that does nothing until finish_scrub returns.
Personally I don't like it one bit. For one thing, pumping UI messages as part of the "show progress" thing is asking for trouble. The right way to do it would be to bite the bullet and run the scrubbing in another thread. But that could be a can of works, as GC isn't designed for multi-threading. Yet, this would allow finish_pending to set the abort flag, wait for the abort to finish, then return TRUE.

jralls:
Yes, everything in GnuCash itself is in one thread; the other threads are created by WebKit and Guile for their own nefarious purposes. You've no doubt figured out that as the scrub proceeds it counts the splits processed and every 100th split it calls `gnc_window_show_progress` that updates the progress bar and then flushes the event queue so that a draw can happen. That also allows other events to process including our quit event (or sometimes a should-quit? event on Macs) and that calls `finish_pending` on every tab in every window including the tab running the scrub.

I suspect that the cause of the crash is that the quit wasn't stopped, there being no check on the scrub status so `gnc_main_window_quit` put up the "save before quitting" dialog and Fred naturally told it to save. `gnc_main_window_quit` sets a repeating 250ms timer, meaning that it places an event on the queue that when it's processed will call `gnc_main_window_timed_quit` 250ms later. `gnc_window_timed_quit` checks to see if the save is done and if it isn't returns `FALSE`, which causes a new timer event to be queued. Eventually all of the events are used up and the scrub resumes, the timer fires and since the save is completed shuts down with the scrub in the middle of doing something that causes a crash.

There's no need to block the quit, you're already at a safe place to stop. Just set `abort_now` so that when control eventually returns to the scrub function it breaks the loop. The `abort_now` check should be immediately *after* `g_free (progressmsg)`, not at the beginning of the loop. `abort_now` means **now** not "after one more scrub".

jeanlaroche:
> There's no need to block the quit, you're already at a safe place to 
> stop. Just set |abort_now| so that when control eventually returns to 
> the scrub function it breaks the loop. The |abort_now| check should be 
> immediately /after/ |g_free (progressmsg)|, not at the beginning of the 
> loop. |abort_now| means *now* not "after one more scrub".

I believe you're right. If I check the abort *immediately* after 
show_progress(), in a way that performs no further checks/operations, 
things should not crash. But I do anything after show_progress() 
returns, there's potential for a crash because the quit process will 
already have been carried out.
I'll try that.
J.

jeanlaroche:
John your suggestion worked, so I checked in all these changes. The code is now much simpler. I also fixed the crash that occurs when a scrub is triggered from the account pane.
I left in the code that checks for gnc_gui_refresh_suspended() before quitting to alert the user if there's an ongoing operation, this is just in case somebody runs into another problem like the one encountered in scrubbing.
Also there was a logic reversal in the dialog that asks whether to quit without saving or cancel in gnc_quartz_should_quit() . I'm not sure how that was never caught before.
The fix is a lot cleaner now.

jralls:
The `gnc_quartz_should_quit` fix should go in its own commit and it's no longer necessary to make `gnc_main_window_cmd_file_quit` public.

jeanlaroche:
> The `gnc_quartz_should_quit` fix should go in its own commit and it's no longer necessary to make `gnc_main_window_cmd_file_quit` public.

True. Fixed now.

jralls:
Thanks!

christopherlam:
In addition to https://github.com/Gnucash/gnucash/pull/779#commitcomment-42066454 I think it makes the UI partially responsive during the Check & Repair process: I can click various accounts and tabs, but Menus, Right-click and Toolbar items are unresponsive. For that matter I think this is a cosmetic regression - previously the action would clearly indicate the book was out of action for a while, and now it looks broken. Is there a way of setting the progress bar in the main window properly? "Check & Repair 4 of 47 accounts, 110 of 1234 splits"?

christopherlam:
Looks like there is already mechanism to update the progressbar text but it's currently broken. We can _hack_ it by adding a message into the statusbar; the proper fix is to fix the GtkProgressBar to display the message as directed:

```c
modified   gnucash/gnome-utils/gnc-window.c
@@ -191,7 +191,12 @@ gnc_window_show_progress (const char *message, double percentage)
     else
     {
         if (message)
+        {
+            GtkStatusbar *status = GTK_STATUSBAR (gnc_window_get_statusbar (window));
+            gtk_statusbar_pop (status, 0);
+            gtk_statusbar_push (status, 0, message);
             gtk_progress_bar_set_text(GTK_PROGRESS_BAR(progressbar), message);
+        }
         if ((percentage == 0) &&
                 (GNC_WINDOW_GET_IFACE(window)->ui_set_sensitive != NULL))
             GNC_WINDOW_GET_IFACE(window)->ui_set_sensitive(window, FALSE);
```

christopherlam:
Both issues fixed with 99101c191 and 979a050ae.

christopherlam:
Minor usability issue remaining: launch a lengthy check&repair, hit Alt-F4 or macOS equivalent: this will show the "A scrubbing operation is currently pending, do you want to abort it?" dialog twice.

jeanlaroche:
> 
> 
> Both issues fixed with [99101c1](https://github.com/Gnucash/gnucash/commit/99101c191c59a41d426382c33a8884169bc4e75e) and [979a050](https://github.com/Gnucash/gnucash/commit/979a050ae9bd53555414bd5b5548a59c3bae057a).

Thanks for fixing these! I'll incorporate your changes. I'll try to repro the Alt-F4 issue.

jralls:
@jeanlaroche this is already merged and @christopherlam pushed his changes onto maint.

christopherlam:
@jeanlaroche sorry to kick this again -- I've modified `abort_scrub` slightly to test every split rather than every 100 splits in 67a18df6a. IMHO scrubbing should be cancellable via UI - <ESC> or <STOP> button. Do you think you can work it?

jeanlaroche:
Sure, I'll see if I can add that. I'm thinking an "abort" button next to 
the progress bar would make sense?

christopherlam:
> Sure, I'll see if I can add that. I'm thinking an "abort" button next to the progress bar would make sense?

Maybe. My thinking is handling an abort button needs full UI therefore something like the following from https://developer.gnome.org/gtk3/stable/gtk3-General.html#gtk-events-pending would need to be inserted into the scrubbing splitlist loop? As I mentioned before https://github.com/Gnucash/gnucash/pull/779#issuecomment-678723176, gets hairy. Or, I may be completely wrong.

```c
while (gtk_events_pending ())
  gtk_main_iteration ();
```

jralls:
This is a merged, and therefore *closed* pull request. It shouldn't be used for ongoing discussion of a topic only marginally related to the original purpose of the PR. New feature discussions should anyway happen on the mailing lists to bring in the widest audience, and in particular UI design discussions should go on the user list to get feedback.

@christopherlam as for the specific question, the progress bar update code runs the event loop, that's how the bug that originated this PR happened: The quit event got processed during a progress bar update and pulled the rug out mid-C&R. An abort button event would be handled the same way.

gjanssens:
> This is a merged, and therefore _closed_ pull request. It shouldn't be used for ongoing discussion of a topic only marginally related to the original purpose of the PR. New feature discussions should anyway happen on the mailing lists to bring in the widest audience, and in particular UI design discussions should go on the user list to get feedback.
> 

I have my doubts about the latter. I would prefer to turn it around: users interested in discussing UI design should subscribe to gnucash-devel.

jralls:
The problem with that is that users think that they're not interested in UI design, but complain mightily when a UI change they don't like appears in a release.

gjanssens:
I understand and I think this is something that is no longer in proper proportions in the gnucash community. My current view on this is if a user cares enough about the design they should invest their time and energy to follow design discussions on gnucash-devel and they are most welcome to do so of course. Otherwise they have to accept others have spent a lot of volunteer effort to make the program as good as possible but perhaps not exactly as they had it in mind.
Just my opinion though.

PR #780, opened 2020-08-29
I18N: Add po copyright-holder

fellen:
I watched, the copyright holder of new po[t] files is currently set to:
`# Copyright (C) 2020 THE GnuCash'S COPYRIGHT HOLDER`

 The commit would set it to "by the translators:"
because the individual (C) comments follow.

Should it stay 

- in this form or 
- be set in a variable,
-  - here or
-  -  more up in the hierarchy?

Older files have:
`Free Software Foundation, Inc.`: old default, OK for files from TP
`THE PACKAGE'S COPYRIGHT HOLDER`: Default, if no package is set
`Gnumatic, Inc.`: dissolved
`gnucash project`
`C-DAC, GIST, Pune, India`.: OK
`by the translators`
`by the respective translator`
Individuals

An interesting aspect from the gettext manual: The msgids were created by the programmers.

### DIFF
--- a/po/gnucash-pot.cmake
+++ b/po/gnucash-pot.cmake
@@ -32,6 +32,7 @@ execute_process(
                         --keyword=translate:1c,2,3,4t
                         --package-name=${PACKAGE_NAME}
                         --package-version=${PROJECT_VERSION}
+                        --copyright-holder=by\ the\ GnuCash\ developers\ and\ the\ translators:
                         --msgid-bugs-address=https://bugs.gnucash.org/enter_bug.cgi?product=GnuCash&component=Translations
   WORKING_DIRECTORY ${PO_BIN_DIR}
   RESULT_VARIABLE GNUCASH_POT_RESULT


jralls:
This [Debian & gettext bug](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=682580) has an interesting discussion about copyrights and the `copyright-holders` xgettext option. Based on the current [manpage](https://linux.die.net/man/1/xgettext) it looks like none of those changes are in the current code but it's probably worth digging into the sources to be sure.

As the bug discussion indicates the pot file copyright is in fact held by all of authors of translatable strings. Figuring who those are would be challenging: I don't think there's an automatic way to do it. For example a naive use of `git blame` would assign all of the strings in Scheme files to @christopherlam because of the recent change from `(_ "foo")` to `(G_ "foo")`. To do it perfectly a human would have to review the history of every single string and in some cases decide what is a big enough change to warrant gaining copyright. I don't think that would be a productive use of anyone's time. It would also produce a ridiculously long list of copyright assertions in the file. How about `--copyright-holder GnuCash\ Developers\ and\ the\ translators:`?

ISTM we should also specify `--foreign-user` to disable the FSF copyright notice.

fellen:
> ISTM we should also specify `--foreign-user` to disable the FSF copyright notice.

From man xgettext: `This option is equivalent to ‘--copyright-holder=''’.` 
The existing FSF entries date back from before we used `--foreign-user`or `--copyright-holder`.

I did not change them because a TP member usually will transfer his copyright to them.
In the case, where a FSF (C) exists in a file, non-TP translators files can insert a second (C) line. I am not sure: should we add a section about it in https://wiki.gnucash.org/wiki/Translation#Adjust_the_header_and_special_strings and a link from https://wiki.gnucash.org/wiki/Translation#Committers

fellen:
> This [Debian & gettext bug](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=682580) has an interesting discussion about copyrights and the `copyright-holders` xgettext option. Based on the current [manpage](https://linux.die.net/man/1/xgettext) it looks like none of those changes are in the current code but it's probably worth digging into the sources to be sure.

There was no patch of their man page.  ;-)
I see the last patch on my system in `/usr/share/gettext/po/Makefile.in.in`, so they are applied.

Because we don't use autotools, gettextize does not copy that file. If we want to use `$(srcdir)/$(DOMAIN).pot-header`, we have to look for corresponding changes in https://cmake.org/cmake/help/latest/module/FindGettext.html or apply it in our handmade po/gnucash-pot.cmake. 
I currently see no need for a customized copyright header.

fellen:
Current header of pot:
```
# SOME DESCRIPTIVE TITLE.
# Copyright (C) YEAR by the GnuCash developers and the translators:
# This file is distributed under the same license as the GnuCash package.
# FIRST AUTHOR <EMAIL@ADDRESS>, YEAR.
#

```
and the result of msginit -l la
```
# Latin translations for GnuCash package.
# Copyright (C) 2020 by the GnuCash developers and the translators:
# This file is distributed under the same license as the GnuCash package.
# Frank H. Ellenberger <frank.h.ellenberger@gmail.com>, 2020.
#

```

jralls:
@fellen if that's what this PR produces it's fine, though it would be better if the second and third line changed places so that the list of translators started immediately after the copyright line. I imagine that's xgettext's fault rather than ours.

fellen:
> though it would be better if the second and third line changed places so that the list of translators started immediately after the copyright line.

I assume that could be done with a customized `$(srcdir)/$(DOMAIN).pot-header` after porting the makefile.in.in-patch on `po/gnucash-pot.cmake`.

> I imagine that's xgettext's fault rather than ours.

Their order has some logic, too:
1. copyright
2. license
3. details

jralls:
It does, but only if the copyright line stands on its own and ours doesn't. Maybe instead of 
```
--copyright-holder=by\ the\ GnuCash\ developers\ and\ the\ translators:
```
we should say 
```
--copyright-holder=by\ the\ GnuCash\ developers\ and\ the\ translators\ listed\ below.
```

fellen:
Right, commit 3f8ab26.

PR #784, opened 2020-09-09
Scrubbing and Gtk improvements

christopherlam:
Following https://github.com/Gnucash/gnucash/pull/779#issuecomment-688584221

Scrubbing (either from account-tree or register) can be aborted by pressing Escape key. Simpler than an abort button.

### DIFF
--- a/gnucash/gnome-utils/dialog-utils.c
+++ b/gnucash/gnome-utils/dialog-utils.c
@@ -535,6 +535,11 @@ gnc_handle_date_accelerator (GdkEventKey *event,
             g_date_to_struct_tm (&gdate, tm);
         return TRUE;
 
+    case GDK_KEY_question:
+        /* Translators: the []-+MHYRT keyboard shortcuts are hard coded
+           and are not localizable */
+        gnc_info_dialog (NULL, "%s", _("This is a date field. Keyboard shortcuts are available: \n- previous day\n+ next day\nshift- previous week\nshift+ next week\n[,alt- previous month \n],alt+ next month\nctrl- previous year\nctrl+next year\nM beginning of month\nH end of month\nY beginning of year\nR end of year\nT today"));
+        return TRUE;
     default:
         break;
     }
--- a/gnucash/gnome/gnc-plugin-page-register.c
+++ b/gnucash/gnome/gnc-plugin-page-register.c
@@ -5004,6 +5004,23 @@ gnc_plugin_page_register_cmd_scrub_current (GtkAction* action,
     LEAVE (" ");
 }
 
+static gboolean scrub_kp_handler (GtkWidget *widget, GdkEventKey *event, gpointer data)
+{
+    if (event->length == 0) return FALSE;
+
+    switch (event->keyval)
+    {
+    case GDK_KEY_Escape:
+        abort_scrub = gnc_verify_dialog
+            (GTK_WINDOW (widget), FALSE,
+             _("'Check & Repair' is currently running, do you want to abort it?"));
+        return TRUE;
+    default:
+        break;
+    }
+    return FALSE;
+}
+
 static void
 gnc_plugin_page_register_cmd_scrub_all (GtkAction* action,
                                         GncPluginPageRegister* plugin_page)
@@ -5013,6 +5030,7 @@ gnc_plugin_page_register_cmd_scrub_all (GtkAction* action,
     GncWindow* window;
     GList* node, *splits;
     gint split_count = 0, curr_split_no = 0;
+    guint scrub_kp_handler_ID;
     const char* message = _ ("Checking splits in current register: %u of %u");
 
     g_return_if_fail (GNC_IS_PLUGIN_PAGE_REGISTER (plugin_page));
@@ -5029,7 +5047,10 @@ gnc_plugin_page_register_cmd_scrub_all (GtkAction* action,
 
     gnc_suspend_gui_refresh();
     is_scrubbing = TRUE;
+    abort_scrub = FALSE;
     window = GNC_WINDOW (GNC_PLUGIN_PAGE (plugin_page)->window);
+    scrub_kp_handler_ID = g_signal_connect (G_OBJECT (window), "key-press-event",
+                                            G_CALLBACK (scrub_kp_handler), NULL);
     gnc_window_set_progressbar_window (window);
 
     splits = qof_query_run (query);
@@ -5043,19 +5064,20 @@ gnc_plugin_page_register_cmd_scrub_all (GtkAction* action,
         PINFO ("Start processing split %d of %d",
                curr_split_no + 1, split_count);
 
+        scrub_split (split);
+
+        PINFO ("Finished processing split %d of %d",
+               curr_split_no + 1, split_count);
+
         if (curr_split_no % 10 == 0)
         {
             char* progress_msg = g_strdup_printf (message, curr_split_no, split_count);
             gnc_window_show_progress (progress_msg, (100 * curr_split_no) / split_count);
             g_free (progress_msg);
         }
-
-        scrub_split (split);
-
-        PINFO ("Finished processing split %d of %d",
-               curr_split_no + 1, split_count);
     }
 
+    g_signal_handler_disconnect (G_OBJECT (window), scrub_kp_handler_ID);
     gnc_window_show_progress (NULL, -1.0);
     gnc_resume_gui_refresh();
     is_scrubbing = FALSE;
--- a/gnucash/gnome-utils/gnc-date-edit.c
+++ b/gnucash/gnome-utils/gnc-date-edit.c
@@ -884,6 +884,9 @@ create_children (GNCDateEdit *gde)
     gtk_container_add (GTK_CONTAINER (gde->date_button), hbox);
     gtk_widget_show (GTK_WIDGET(hbox));
 
+    gtk_widget_set_tooltip_text (gde->date_entry,
+                                 _("This is a date field. Keyboard shortcuts are available: \n- previous day\n+ next day\nshift- previous week\nshift+ next week\n[,alt- previous month \n],alt+ next month\nctrl- previous year\nctrl+next year\nM beginning of month\nH end of month\nY beginning of year\nR end of year\nT today"));
+
     /* Calendar label, only shown if the date editor has a time field */
     gde->cal_label = gtk_label_new (_("Calendar"));
     gnc_label_set_alignment (gde->cal_label, 0.0, 0.5);
--- a/gnucash/register/register-gnome/datecell-gnome.c
+++ b/gnucash/register/register-gnome/datecell-gnome.c
@@ -623,6 +623,8 @@ gnc_date_cell_realize (BasicCell *bcell, gpointer data)
     box->item_edit = item_edit;
     box->date_picker = GNC_DATE_PICKER (gnc_date_picker_new ());
     gtk_widget_show_all (GTK_WIDGET(box->date_picker));
+    gtk_widget_set_tooltip_text (GTK_WIDGET(item_edit),
+                                 _("This is a date field. Keyboard shortcuts are available: \n- previous day\n+ next day\nshift- previous week\nshift+ next week\n[,alt- previous month \n],alt+ next month\nctrl- previous year\nctrl+next year\nM beginning of month\nH end of month\nY beginning of year\nR end of year\nT today"));
     g_object_ref_sink(box->date_picker);
 
     /* to mark cell as realized, remove the realize method */


christopherlam:
Alternatively: hitting `ESC` will also call `finish_scrub` to offer dialog prior to aborting scrub.

christopherlam:
Date field can accept `?` key which will show the available shortcuts.

christopherlam:
Note the date keyboard handler `gnc_info_dialog` uses parent=`NULL` because I can't figure out how to find GtkWindow from the GdkWindow.

Screenshot: 
image

gjanssens:
> Note the date keyboard handler `gnc_info_dialog` uses parent=`NULL` because I can't figure out how to find GtkWindow from the GdkWindow.

According to the (GdkWindow documentation)[https://developer.gnome.org/gdk3/stable/gdk3-Windows.html#gdk-window-set-user-data] you can expect the event's user data to be the GtkWidget if non-null. Passing that widget to gnc_ui_get_gtk_window will probably get you a top-level widget to use as parent.

christopherlam:
> > Note the date keyboard handler `gnc_info_dialog` uses parent=`NULL` because I can't figure out how to find GtkWindow from the GdkWindow.
> 
> According to the (GdkWindow documentation)[https://developer.gnome.org/gdk3/stable/gdk3-Windows.html#gdk-window-set-user-data] you can expect the event's user data to be the GtkWidget if non-null. Passing that widget to gnc_ui_get_gtk_window will probably get you a top-level widget to use as parent.

Thanks. This PR is more or less suggestions for improving UI and responsiveness and raise discussions.

1. re: date documentation - not sure whether to trap "?" or `gtk_widget_set_tooltip_text` into the date field. There are pros and cons for both. Also macOS doesn't use Ctrl/Alt/Shift so the text won't be valid as is.

2. re: catching <ESC> to abort lengthy scrub/reports etc. I've had a thought how it could work:
(a) copy this abort-scrub process into other lengthy processes -- eg reports can also test gtk signals while scanning splitlist and abort when requested.
(b) `gnc:report-run` can catch an exception 'abort-report which is thrown by `gnc:report-percent-done` which throws it
(c) a global state `cease_all_lengthy_processes` default `FALSE`, and is tested during scrub or report splitlist scans, and is reset when we return to the main loop.

jralls:
> Also macOS doesn't use Ctrl/Alt/Shift

Not quite. macOS certainly does use those, plus a fourth key "Command", but it uses them differently from Linux and Windows: For most common shortcuts (e.g. cut, copy, paste, quit) command is used instead of control, and alt is used to extend the keyboard: For example alt-3 produces £ and alt-@ produces €. However keyboard handling in Gtk is done at a lower level so that if you write a key handler for <alt>-3 to do something that will be processed before it's passed to the GtkEntry whose handler will interpret it the Apple way. The datecell handler doesn't work correctly for alt-- and alt-+ (it looks like it's interpreting them as the extended keyboard keys – (n-dash) and ≠ and filtering them away because they're not legal date characters resulting in a blanked field) but does work correctly for ctrl and shift.

An aside: While the instructions are intuitively correct they're technically wrong, at least on US keyboards: + is the shifted state of = and the datecell handler gives shift-= priority over ctrl-=, so ctrl-+ advances the date by a week rather than a year.

jralls:
> copy this abort-scrub process into other lengthy processes -- eg reports can also test gtk signals while scanning splitlist and abort when requested.

It would be better to find a lower-level place for it rather than to copy the code around. The latter approach makes for a maintenance headache.

> a global state cease_all_lengthy_processes default FALSE, and is tested during scrub or report splitlist scans, and is reset when we return to the main loop.

Only modal dialog boxes have separate runloops, there is no "return to the main loop". Since we're currently single-threaded (not counting the threads guile and webkit create for their own purposes) there's only one lengthy process at a time so the <escape> handler can call `gnc_set_quit_activity()` and the scrub and report code can use `gnc_check_quit_activity()` to decide whether to bail and then `gnc_clear_quit_activity()` in the bail-out code. You might also provide a `gnc_confirim_quit_activity()` that puts up a yes-no dialog box for the user to confirm that they really meant to quit and it wasn't the cat stepping on the escape key.

christopherlam:
Last commit attempts to augment gnc_progressbar functions. Because the progressbar is always associated with a window, I suggest that the window will detect ESC keypress and set the `quitting_state` boolean, which can be queried by reports or scrubs.

Pressing ESC will trigger ask dialog "Do you wish to cancel?"; if yes will set `quitting_state` which is retrievable by `gnc_progressbar_get_quitting_state`. This means a datafile load/save *can* be cancellable too, so, may need to be disabled for these.

Good idea?

gjanssens:
> Last commit attempts to augment gnc_progressbar functions. Because the progressbar is always associated with a window, I suggest that the window will detect ESC keypress and set the `quitting_state` boolean, which can be queried by reports or scrubs.
> 
> Pressing ESC will trigger ask dialog "Do you wish to cancel?"; if yes will set `quitting_state` which is retrievable by `gnc_progressbar_get_quitting_state`. This means a datafile load/save _can_ be cancellable too, so, may need to be disabled for these.
> 
> Good idea?

Sounds reasonable if configurable (so it can be disabled for load/save indeed).

Minor nitpick: `quitting_state` is confusing to me as "quit" refers to ending the whole program which is not what this is about. The dialog that pops up uses the word "Cancel". That sounds more accurate. Can you use `cancelled_state` as alternative variable name ?

christopherlam:
> Sounds reasonable if configurable (so it can be disabled for load/save indeed).
> 
> Minor nitpick: `quitting_state` is confusing to me as "quit" refers to ending the whole program which is not what this is about. The dialog that pops up uses the word "Cancel". That sounds more accurate. Can you use `cancelled_state` as alternative variable name ?

I think then instead of boolean we need to have state: `gnc_progressbar_set_mode (state)` where `state` will be either `INACTIVE`, `CANCELLABLE`, `CANCELLING`, or `NONCANCELLABLE`.

Pressing Escape will ask `gnc_verify_dialog` iff it's `CANCELLABLE` and set it to `CANCELLING`, and the latter will signal reports and scrub to bail out.

christopherlam:
A salad of suggestions. Comments welcome.

christopherlam:
Latest iteration: register report is now cancellable during splitlist scan.

This means [ESC] will interrupt the table generation and send partially complete table to renderer. This means the report is completed with incomplete information.

Alternatively we can hook into the report-utilities or report-core, trap abort, and render a generic "Report generation cancelled". Opinions?

christopherlam:
In other words, one approach is to modify all reports individually to bail and render partial report, the other approach is to enable bail in all reports automatically and render generic message. Both approaches require that the report updates the progress bar.

christopherlam:
My approach above is to supercharge gnc_progressbar.

By default the behaviour is very similar to previous -- all callers can use it unchanged, but are encouraged via PWARN to set `gnc_progress_set_cancellable (boolean)` which sets state to `ACTIVE_CANCELLABLE` or `ACTIVE_NONCANCELLABLE`; if cancellable is enabled then we can trap ESC to set state to `ACTIVE_CANCELLING` which can then be tested via `gnc_progress_get_cancelling` by callers. Upon completion state becomes `INACTIVE`.

christopherlam:
Latest commit does add STOP button which works. But code is terrible. Ideally it'd hook to the main toolbar stop button, and set its sensitivity according to cancellable status.

image

christopherlam:
The approach described here is "low level" insofar it augments the progress bar code, but the caller still needs to detect cancelling status to bail appropriately and perform appropriate housekeeping eg freeing memory etc. Reports will also be able to render "cancelled report".

A perfect low-level code would detect stop/abort and find a way to "go-to main loop" which imho is much uglier.

jralls:
>  find a way to "go-to main loop" which imho is much uglier.

You seem not to understand how the event loop works. You can't go to somewhere when you're already there.

christopherlam:
Agree, I don't! Hence the last commit is pretty bad but seems to work...

gjanssens:
A more generic approach could be to allow callers to supply a callback function. If the progress gets a cancel trigger it can then invoke this callback function. That way users of the progress bar can define what should happen in case their progress is cancelled by the user.

Christopher Lam <notifications@github.com> schreef op 13 september 2020 07:19:39 CEST:
>The approach described here is "low level" insofar it augments the
>progress bar code, but the caller still needs to detect cancelling
>status to bail appropriately and perform appropriate housekeeping eg
>freeing memory etc. Reports will also be able to render "cancelled
>report".
>
>A perfect low-level code would detect stop/abort and find a way to
>"go-to main loop" which imho is much uglier.
>
>-- 
>You are receiving this because you commented.
>Reply to this email directly or view it on GitHub:
>https://github.com/Gnucash/gnucash/pull/784#issuecomment-691616219

Sent from my smartphone. Please excuse my brevity.

christopherlam:
> A more generic approach could be to allow callers to supply a callback function. If the progress gets a cancel trigger it can then invoke this callback function. That way users of the progress bar can define what should happen in case their progress is cancelled by the user.

I like that... similar to gtk.

But the current approach could be slightly modified so that the report code can test `cancelling`, and if so, display a partially generated report. I'm not sure guile can supply callback. The callback approach would need to be done from C therefore report cancellation shows a generic cancellation message.

gjanssens:
> > A more generic approach could be to allow callers to supply a callback function. If the progress gets a cancel trigger it can then invoke this callback function. That way users of the progress bar can define what should happen in case their progress is cancelled by the user.
> 
> I like that... similar to gtk.
> 
> But the current approach could be slightly modified so that the report code can test `cancelling`, and if so, display a partially generated report.

> I'm not sure guile can supply callback. The callback approach would need to be done from C therefore report cancellation shows a generic cancellation message.

It would certainly be more challenging, but I believe it could be done with a couple of helper functions. Instead of registering the callback function straight from guile, we could write a wrapper function in C that serves as callback function. That wrapper function can take some scheme procedure and data object as user data and when the wrapper is invoked by the callback mechanism, it fires the scheme procedure and on the data object. That data may hold the partially generated report and the procedure may write it out. The challenge is likely mostly in lifetime management of the scheme objects.

christopherlam:
> It would certainly be more challenging, but I believe it could be done with a couple of helper functions. Instead of registering the callback function straight from guile, we could write a wrapper function in C that serves as callback function. That wrapper function can take some scheme procedure and data object as user data and when the wrapper is invoked by the callback mechanism, it fires the scheme procedure and on the data object. That data may hold the partially generated report and the procedure may write it out. The challenge is likely mostly in lifetime management of the scheme objects.

Ok I am not sure which way is "proper" but if the preferred approach is to supply a callback then I'll modify this branch.

It would seem that `gnc_window_show_progress (NULL, -1)` is used to reset the progressbar, so, I think it's the perfect place to reset the callback.

christopherlam:
FWIW The scheme object containing it is usually `document` being heavily modified, and is not accessible from C at all until it's complete. So, if this is still desired, as a first attempt I'd not attempt render a partially written report.

jralls:
> FWIW The scheme object containing it is usually document being heavily modified, and is not accessible from C at all until it's complete. So, if this is still desired, as a first attempt I'd not attempt render a partially written report.

ISTM a partial report is unlikely to be useful so I suggest that you avoid the complication entirely.

gjanssens:
> > FWIW The scheme object containing it is usually document being heavily modified, and is not accessible from C at all until it's complete. So, if this is still desired, as a first attempt I'd not attempt render a partially written report.
> 
> ISTM a partial report is unlikely to be useful so I suggest that you avoid the complication entirely.

I completely agree on that.

christopherlam:
> > > FWIW The scheme object containing it is usually document being heavily modified, and is not accessible from C at all until it's complete. So, if this is still desired, as a first attempt I'd not attempt render a partially written report.
> > 
> > 
> > ISTM a partial report is unlikely to be useful so I suggest that you avoid the complication entirely.
> 
> I completely agree on that.

I was trying to mimic behaviour of web browsers :) I've transformed this PR completely to use callback.

gnucash/gnome-utils/gnc-window.c
```diff
@@ -139,6 +140,33 @@ gnc_window_set_status (GncWindow *window, GncPluginPage *page,
...
+progressbar_kp (GtkWidget *widget, GdkEventKey *event, gpointer data)
+{
+    if (event->length == 0) return FALSE;
+
+    switch (event->keyval)
+    {
+    case GDK_KEY_Escape:
+        if (escape_handler_cb)
+        {
+            (*escape_handler_cb)(widget);
```
christopherlam, line 156: instead of the `escape_handler_cb` being customised by the caller, we can also customise the whole g_signal handler. Would that be better?

christopherlam:
Another approach whereby `gnc-plugin-page-register.c` sets the keypress handler directly instead of augmenting progressbar.

gnucash/gnome-utils/gnc-window.c
jralls: > instead of the escape_handler_cb being customised by the caller, we can also customise the whole g_signal handler. Would that be better?

Which signal and customize how? Would the caller pass in a `GCallback` that some gnc-window function would connect to the signal or would the caller connect directly?

christopherlam: > Would the caller pass in a GCallback that some gnc-window function would connect to the signal

yes that'd be the idea. I have since then added 2 more commits which avoids augmenting progressbar and the scrubber itself creates the signal. This sounds cleaner...

gnucash/gnome/gnc-plugin-page-register.c
```diff
@@ -4973,6 +4973,13 @@ gnc_plugin_page_register_cmd_scrub_current (GtkAction* action,
     LEAVE (" ");
 }
 
+static void cancel_scrub (GtkWidget *window)
+{
+    abort_scrub = gnc_verify_dialog
+        (GTK_WINDOW (window), FALSE,
+         _("A scrubbing operation is currently pending, do you want to abort it?"));
```
jralls: [s/scrubbing/Check and Repair/, s/pending/running/](https://code.gnucash.org/logs/2020/09/16.html#T15:37:03)

gnucash/gnome/gnc-plugin-page-register.c
```diff
@@ -5019,6 +5028,8 @@ gnc_plugin_page_register_cmd_scrub_all (GtkAction* action,
             g_free (progress_msg);
         }
 
+        if (abort_scrub) break;
+
```
jralls, line 5044: Why?

gnucash/gnome/gnc-plugin-page-register.c
```diff
@@ -4973,11 +4973,21 @@ gnc_plugin_page_register_cmd_scrub_current (GtkAction* action,
     LEAVE (" ");
 }
 
-static void cancel_scrub (GtkWidget *window)
+static gboolean scrub_handler (GtkWidget *widget, GdkEventKey *event, gpointer data)
```
jralls, line 4976: How about `scrub_abort_handler`?

gnucash/gnome/gnc-plugin-page-register.c
christopherlam: Why move the `abort_scrub` test from the for loop to here? ISTM that `gnc_window_show_progress` which calls `gtk_main_iteration` seemed to modify something which results in the `split` object being invalid, therefore aborting the scrub led to a `g_warning`. I'll check again.

christopherlam: Confirmed. While checking `node && !abort_scrub` is more elegant, it means: if a lengthy scrub is running, and user closes via Alt-F4 (on Linux), and agrees to abort, the `split` becomes `NULL` which means `scrub_split(split)` fails.

gnucash/gnome-utils/dialog-utils.c
```diff
@@ -535,6 +535,11 @@ gnc_handle_date_accelerator (GdkEventKey *event,
             g_date_to_struct_tm (&gdate, tm);
         return TRUE;
 
+    case GDK_KEY_question:
+        /* Translators: the []-+MHYRT keyboard shortcuts are hard coded
+           and are not localizable */
+        gnc_info_dialog (NULL, "%s", _("This is a date field. Keyboard shortcuts are available: \n- previous day\n+ next day\nshift- previous week\nshift+ next week\n[,alt- previous month \n],alt+ next month\nctrl- previous year\nctrl+next year\nM beginning of month\nH end of month\nY beginning of year\nR end of year\nT today"));
+        return TRUE;
```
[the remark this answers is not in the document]
christopherlam: Here's tooltip- is this better?

image
image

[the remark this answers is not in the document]
fellen:
```
        gnc_info_dialog (NULL, "%s", _("This is a date field. Keyboard shortcuts are available:\n\
                                        [-]: previous day\n\
                                        [+]: next day\n\
                                        [shift]+[-]: previous week\n\
                                        [shift]+[+]: next week\n\
                                        [alt]+[-] or "[": previous month \n\
                                        [alt]+[+] or "]": next month\n\
                                        [ctrl]+[-]: previous year\n\
                                        [ctrl]+[+]: next year\n\
                                        [M]: beginning of month\n\
                                        [H]: end of month\n\
                                        [Y]: beginning of year\n\
                                        [R]: end of year\n\
                                        [T]: today"));

```
would be easier to read for
1. coders by the line breaks
2. users by marking keys
Note: on a german keyboard "[" is [AltGr]+[8] and "]" [AltGr]+[9]

Ideally the Characters would be translatable e.g. for de to
M, T, (Monat)
J, R (Jahr)
H (heute)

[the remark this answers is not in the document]
christopherlam: the MHYRT hotkeys are a direct descendant of Quicken -- you can test via https://archive.org/details/msdos_Quicken_7 and also I'd argue very difficult to translate.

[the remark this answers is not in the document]
fellen: … in the english version.

They are the first and the last character of the corresponding word like "M" and "H" in "MontH".
And it is in the responsibility of the translator to choose another in case of duplicates.

BTW, Fix:
```
                                        [alt]+[-] or \"[\": previous month\n\
                                        [alt]+[+] or \"]\": next month\n\

```

gnucash/gnome/gnc-plugin-page-register.c
```diff
@@ -5050,12 +5071,15 @@ gnc_plugin_page_register_cmd_scrub_all (GtkAction* action,
...
 
+        if (abort_scrub) break;
+
         scrub_split (split);
 
         PINFO ("Finished processing split %d of %d",
                curr_split_no + 1, split_count);
     }
 
+    g_signal_handler_disconnect (G_OBJECT (window), scrub_kp_handler_ID);
```
jralls, line 5082: I don't think you want to disconnect the handler, that will would mean it works only the first time check & repair is run. You *do* need to `abort_scrub = FALSE;` here other wise subsequent runs of check & repair will do only one split.

gnucash/gnome/gnc-plugin-page-register.c
jralls: So move `scrub_split` and its `PINFO` to before the progress bar update.

christopherlam:
Changed. I'm also dubious about the centralizing `gnc_resume_gui_refresh ()` - I *think* it's placed incorrectly in gnc-plugin-page-register.c `finish_scrub` and should be removed.

jralls:
The problem that extra `gnc_resume_gui_refresh` is addressing is that `gnc_plugin_page_register_finish_pending` has to complete before control returns to `gnc_plugin_page_register_cmd_scrub_all` and it needs to have `suspend_counter == 0` to do so cleanly, so I think `gnc_plugin_page_register_cmd_scrub_all` needs to
```
if (gnc_gui_refresh_suspended())
   gnc_resume_gui_refresh();
```
after the loop completes to prevent an underflow error getting written to the tracefile.

christopherlam:
> The problem that extra `gnc_resume_gui_refresh` is addressing is that `gnc_plugin_page_register_finish_pending` has to complete before control returns to `gnc_plugin_page_register_cmd_scrub_all` and it needs to have `suspend_counter == 0` to do so cleanly, so I think `gnc_plugin_page_register_cmd_scrub_all` needs to
> 
> ```
> if (gnc_gui_refresh_suspended())
>    gnc_resume_gui_refresh();
> ```
> 
> after the loop completes to prevent an underflow error getting written to the tracefile.

I strongly suspect this `gnc_resume_gui_refresh` was copied from `gnucash/gnome/gnc-plugin-page-account-tree.c` in #779 and doesn't really apply to `gnucash/gnome/gnc-plugin-page-register.c`. I'd think suspend and refresh should be balanced, and if code needs to test for suspended status then it's trying to wave the imbalance away. @jeanlaroche could review both cases and let us know thoughts.

jralls:
You're absolutely right that they need to be balanced. @Bob-IT would be a better choice for determining whether we need to resume refresh for `gnc_plugin_page_register_finish` or if it can wait for all of the pending events to complete. An alternative would be
```
     if (curr_split_no % 10 == 0)
        {
            char* progress_msg = g_strdup_printf (message, curr_split_no, split_count);
            gnc_resume_gui_refresh();
            gnc_window_show_progress (progress_msg, (100 * curr_split_no) / split_count);
            g_free (progress_msg);
            gnc_suspend_gui_refresh();
        }
...
gnc_resume_gui_refresh();
gnc_window_show_progress(NULL, -1.0);
is_scrubbing = FALSE;
LEAVE("");
}
```
That way the suspends and resumes stay balanced and the gui is refreshing as it should while the event loop is processed.

Bob-IT:
I will look at this this week.

Bob-IT:
I have looked at this and as a result of making a few changes decided to create a new PR #803 with them to make it easier. Hope this all makes sense, Bob.

Regarding the tool tip for dates, unfortunately the changes for the register are wrong as you will get a date tool tip for every edited cell. There is a way to do this by following the approach for the reconcile cell but to only display when the cell is being edited or being a bit more specific in which widget is connecting to the tool tip signal. Maybe this should be a new PR.

christopherlam:
The quality of your work will undoubtedly surpass mine, so, I'll defer 🙇.

With regards to date-cell tooltip, agree it is not easy to target specific date-editor cell types. A more general tooltip mechanism would be nice -- eg. Date, Action/Num, Account, Amount. Alternatively a common UI idiom is a short explanation in statusbar and longer one with hotkey eg. Ctrl-Shift-? or tooltip.

Bob-IT:
> With regards to date-cell tooltip, agree it is not easy to target specific date-editor cell types. A more general tooltip mechanism would be nice -- eg. Date, Action/Num, Account, Amount. Alternatively a common UI idiom is a short explanation in statusbar and longer one with hotkey eg. Ctrl-Shift-? or tooltip.

@christopherlam There already is a tooltip mechanism for the register as used by the doclink and recn cells, just need to tap into that so here is a patch that does that. I would assume a similar change would be needed for the business register if this was to be used. With this setup the tooltip does not timeout as it was not required for the mentioned cells, not sure if that is possible with the current implementation.

[0001-Register-date-cell-tooltip-for-shortcuts.txt](https://github.com/Gnucash/gnucash/files/5415333/0001-Register-date-cell-tooltip-for-shortcuts.txt)

I am on the fence about this tooltip, the status bar may be a better solution that mentions a shortcut to a help page.

christopherlam:
The statusbar currently shows the date in expanded form "Monday 12 October 2020". I think your #803 is of much better quality so I'll close this own down. Date shortcut keys just need better documentation somewhere. @fellen can organise localising hotkeys if he wishes!

PR #785, opened 2020-09-15
Bug 102787: Currency selection should do autocompletion

brooksidesw:
https://bugs.gnucash.org/show_bug.cgi?id=102787
This change allows GTK entry completion to take effect when a user types into the combo box.


brooksidesw:
In case there's any confusion, brooksidesw is the GitHub identity for Dean Jagels

jralls:
Excellent, thanks.

PR #786, opened 2020-09-16
Call to flatten bayesian import_map gets percentfunc

christopherlam:
Attempt to augment account.cpp's `check_import_map_data` to receive a `QofPercentageFunc` argument which is usually `gnc_window_show_progress`.

Ultimately the old code to call `check_import_map_data` will silently upgrade the import map, and new code runnable from UI will upgrade the import map *and* update progressbar. Also not sure which UI call to trigger the conversion -- I've added the trigger into account-tree scrub as a temporary hack.

But I can't figure out how to pass a pointer to callback to `gnc_account_check_import_map_data`, which is skipped if the `QofPercentageFunc` is `nullptr`.

### DIFF
--- a/gnucash/gnome/gnc-plugin-page-account-tree.c
+++ b/gnucash/gnome/gnc-plugin-page-account-tree.c
@@ -1939,6 +1939,9 @@ gnc_plugin_page_account_tree_cmd_scrub (GtkAction *action, GncPluginPageAccountT
 
     g_return_if_fail (account != NULL);
 
+    gnc_account_check_import_map_data (gnc_get_current_book (),
+                                       (QofPercentageFunc*)gnc_window_show_progress);
+
     prepare_scrubbing ();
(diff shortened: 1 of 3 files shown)


jralls:
@christopherlam The name of the function is its ptr:
```
static void imap_percentage_cb(const char* message, double progress);
```
...
```
gnc_account_check_import_map_data (gnc_account_get_book (acc), imap_percentage_cb);
```

jralls:
@christopherlam You've got the insertion right in your second commit. But this isn't a problem only for the map editor dialog, it's also a problem when running an import. I think we need to work on speeding up the check rather than making a progress bar for it.

jralls:
@christopherlam I think we don't need this anymore because of 9c8fa5cd2.

christopherlam:
Obsolete thanks!

PR #788, opened 2020-09-18
Bug797945 - Imap Changes

Bob-IT:
I was going to push this but decided to create a PR instead for comment.

This is a follow up to John's recent push which stops the imap convert to flat from running the same number of times equal to the number accounts in the list when there are no bayes KVP. With no bayes KVP entries the feature GNC_FEATURE_GUID_FLAT_BAYESIAN will never be set.

I think it is OK.


Bob-IT:
OK, forgot to run the tests and I think I know what I missed, just waiting for a rebuild.

jralls:
@Bob-IT looks ok to me too.

PR #789, opened 2020-09-26
Bug 797893 - Unable to change font size in charts 

christopherlam:
html-chart copies from stylesheet title & text font styles into chartjs JSON property.


jralls:
@christopherlam Not quite what I had in mind but definitely better than hard-coded values. Go ahead and merge if you have time in the next couple of hours, otherwise I will before preparing the release.

christopherlam:
I forgot to check if `document.querySelector` is available for use by libwebkit1. Could we delay release by 24h while I check today's build?

christopherlam:
> 
> 
> I forgot to check if `document.querySelector` is available for use by libwebkit1. Could we delay release by 24h while I check today's build?

Ok panic over, it works well. Let's carry on ✌️

PR #790, opened 2020-09-27
Update hr.po

milotype:
Update with current .pot file at https://translationproject.org/POT-files/gnucash-4.1.pot

### DIFF
--- a/po/hr.po
+++ b/po/hr.po
@@ -5,11 +5,11 @@
 #
 msgid ""
 msgstr ""
-"Project-Id-Version: GnuCash 4.0\n"
+"Project-Id-Version: GnuCash 4.1\n"
 "Report-Msgid-Bugs-To: https://bugs.gnucash.org/enter_bug.cgi?"
 "product=GnuCash&component=Translations\n"
-"POT-Creation-Date: 2020-06-27 15:34-0700\n"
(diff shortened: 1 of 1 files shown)


jralls:
Darn, a day late. I *just* released 4.2. 

Thanks!

fellen:
Milo, I have merged a fresh pot for you in commit 0666327. Why didn't you ask me before you started?

@jralls I see mainly 2 possibilities to improve this situation:

1. Make magic: If source files change, generate a new pot. There is also the possibility to add additional checks, see `xgettext --check --sentence-end`
If pot file changes, run msgmerge for ALL_LINGUAS
It would consume less time than mage check, but might be noissy in commits.

1. @milotype could help us to setup weblate.

jralls:
@fellen Running make is a bit of a strange thing to do when working on a website, but if it was set up as a git hook on code (like the one that creates gnucash-patches and gnucash-changes) it might work out OK.
 
As for weblate we need @derekatkins involvement to set up the instance on code.

milotype:
I don't know how I could help with Weblate, as I'm not a programmer ... :-(

milotype:
When setting up Wablate, I'd suggest the following:
- set the project "GnuCash"
- set components: "GnuCash App", GnuCash Website"
- ideally, the current GnuCash glossary files could be imported into language dependant glossaries for the GnuCash project

PR #792, opened 2020-09-29
Merge GtkCheckButton with GtkLabel

christopherlam:
... because using the gtkcheckbutton's label makes the checkbutton much easier to click. I think there _are_ other disjointed GtkLabel&GtkCheckButtons around which could be merged.

image


Bob-IT:
@christopherlam  You are probably right, I did change some when I did the Options dialog and then that was changed after a discussion. I still do not like the tick box before the text, it just does not seem to flow but understand the change so push it.
I do not think there were many instances like this.

christopherlam:
> While you are updating the account glade file you will no doubt want to update the 4 items for the account_dialog

Oops didn't see; not quite sure what you mean exactly.

PR #793, opened 2020-09-30
Maint deprecate xacc account count splits

christopherlam:
I'll put forward the claim that xaccAccountCountSplits is poorly used; the exact number of splits is hardly useful and the information desired can be gleaned much more effectively using `xaccAccountGetSplitList` or `gnc_account_and_descendants_empty`, a new method. I'd hope these commits are self explanatory and will be welcome.

### DIFF
--- a/gnucash/gnome-utils/dialog-account.c
+++ b/gnucash/gnome-utils/dialog-account.c
@@ -1389,7 +1389,7 @@ gnc_account_window_create(GtkWindow *parent, AccountWindow *aw)
                          &aw->commodity_mode);
 
     // If the account has transactions, prevent changes by displaying a label and tooltip
-    if (xaccAccountCountSplits (aw_get_account (aw), FALSE) > 0)
+    if (xaccAccountGetSplitList (aw_get_account (aw)) != NULL)
     {
         const gchar *sec_name = gnc_commodity_get_printname (xaccAccountGetCommodity(aw_get_account (aw)));
         GtkWidget *label = gtk_label_new (sec_name);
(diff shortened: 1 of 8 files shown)


jralls:
No need to deprecate C API, we still consider that private. Just go through the code and replace all instances of the old function. *However* if a C function is swigged you *do* need to deprecate the Python and Scheme API.

Keep in mind the two long term goals (stuff lives in a SQL database until needed and replacing GList with std::vector) when doing this sort of change. Replacing finding the list-length when you really want to know if it's empty as in this PR is an obvious win and can safely ignore those goals, but if any significant work is involved the effort would be better advancing those goals instead of trying to wring more efficiency out of a poorly-chosen container type.

christopherlam:
> No need to deprecate C API, we still consider that private. Just go through the code and replace all instances of the old function. _However_ if a C function is swigged you _do_ need to deprecate the Python and Scheme API.

All of Account.h is swigged, no? IIUC deprecation in it would mean adding @deprecated in the comments only?

jralls:
Putting `@deprecated` in the comments wouldn't be very useful. [`__attribute__((deprecated))`](https://gcc.gnu.org/onlinedocs/gcc-4.7.0/gcc/Function-Attributes.html) would be better. *But better still* to just replace all of the uses and add the Scheme and Python functions in the bindings code and mark that deprecated.

christopherlam:
I'll leave C[++] deprecations and .i bindings to you. There are quite a few dxacc* API which could be marked as well.

PR #794, opened 2020-10-01
Disable initial report animation

fiveop:
Previously, every reload caused the chart to move through an animation, which
prevented visual comparison of state before and after reload.

### DIFF
--- a/gnucash/report/html-chart.scm
+++ b/gnucash/report/html-chart.scm
@@ -155,6 +155,8 @@
                  (cons 'datasets #())))
     (cons 'options (list
                     (cons 'maintainAspectRatio #f)
+                    (cons 'animation (list
+                                      (cons 'duration 0)))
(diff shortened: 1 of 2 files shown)


jralls:
Please also remove the now redundant animation in budget-barchart.scm.

You might consider similarly disabling the hover animation as in budget-barchart.scm.

fiveop:
I now also removed the now redundant line in budget-barchart.scm, but added a comment, so that future readers of the code do not confused why one of the three disable options is missing.

I do not mind the hover and responsive animations (well, I have no Idea how to trigger the second. Window resize does not do it.), so I did not disable them globally.

Since I think the added comment in budget-barchart.scm necessary, the code would certainly be cleaner, if all animations were disabled globally.

christopherlam:
Thanks, nice clean commit!

PR #795, opened 2020-10-04
Owner-Report can display business document links

christopherlam:
Partial PR. Handle `invoice-doclink-guid=` links. This PR currently sends doclink URI in URL; it should send invoice GUID, and `top-level.c` would retrieve gncInvoice from GUID. Not sure how to do this. @Bob-IT ?

Not sure whether to add doclink as a separate column, or linkify the "Invoice" in the Type column.

Also "Display/Document Links" will now clash with "Display/Links" option.

Progress so far:
image

### DIFF
--- a/gnucash/report/reports/standard/new-owner-report.scm
+++ b/gnucash/report/reports/standard/new-owner-report.scm
@@ -57,7 +57,7 @@
 (define debit-header (N_ "Debits"))
 (define balance-header (N_ "Balance"))
 (define doclink-header (N_ "Document Links"))
-(define linked-txns-header (N_ "Links"))
+(define linked-txns-header (N_ "Transaction Links"))
 
 (define javascript "
 <script>
--- a/gnucash/report/reports/standard/new-owner-report.scm
+++ b/gnucash/report/reports/standard/new-owner-report.scm
@@ -56,6 +56,7 @@
 (define credit-header (N_ "Credits"))
 (define debit-header (N_ "Debits"))
 (define balance-header (N_ "Balance"))
+(define doclink-header (N_ "Document Links"))
 (define linked-txns-header (N_ "Links"))
 
 (define javascript "
@@ -162,6 +163,8 @@
   (vector-ref columns-used 8))
 (define (bal-col columns-used)
   (vector-ref columns-used 9))
+(define (doclink-col columns-used)
+  (vector-ref columns-used 10))
 
 (define (num-cols columns-used section)
   (let* ((date? (date-col columns-used))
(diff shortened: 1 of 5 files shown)


Bob-IT:
@christopherlam  change top-level.c as follows...
add #include "gncInvoice.h" at line 50 
add GncInvoice  * invoice; at line 124
and make the if statement at 185 look like....

    else if (strncmp ("invoice-doclink-guid=", location,
                      strlen ("invoice-doclink-guid=")) == 0)
    {
        if (!validate_type("invoice-doclink-guid=", location, GNC_ID_INVOICE,
                           result, &guid, &entity))
            return FALSE;

        invoice = (GncInvoice *) entity;
        gnc_doclink_open_uri (gnc_ui_get_gtk_window (GTK_WIDGET (result->parent)),
                              gncInvoiceGetDocLink (invoice));
        return TRUE;
    }

Now do your magic scheme stuff in new-owner-report.scm line 808
I fudged it by doing...
;;         (gncInvoiceGetDocLink invoice)
         (gncInvoiceReturnGUID invoice)
but I imagine you will need to do someing like the trep engine.

I would keep it as a separate column that can be added or not and maybe add 'Enable Links' like the 'Transaction Report'

christopherlam:
Nice, works well. Now what remains is i10n and strings. We should standardise them:

In Owner Report:
Document Links = doclinks
Links = invoice->payments and payment->invoices

In Transaction report:
Enable Links = allow HTML anchors. (This could be removed... is redundant because stylesheet has option to disable links)

image
image

gnucash/report/reports/standard/new-owner-report.scm
```diff
@@ -56,6 +56,7 @@
 (define credit-header (N_ "Credits"))
 (define debit-header (N_ "Debits"))
 (define balance-header (N_ "Balance"))
+(define doclink-header (N_ "Document Links"))
 (define linked-txns-header (N_ "Links"))
```
fellen, line 60: What kind of link are this? If we have "Document Links", it should become "Transaction Links" or whatever it is.

christopherlam: These are links from data file to external Uri describing files or web address.

fellen: doclink-header (N_ "Document Links") is obvisious.
I meant linked-txns-header (N_ "Links")

jralls: It's the transactions linked to the bill/invoice/voucher/credit note via the "post" operation.

christopherlam: So:

Owner Report:
`Document Links` = doclinks
`Transaction Links` vs `Business Links` = invoice->payments and payment->invoices

In Transaction report:
`Enable Links` - remove?

PR #796, opened 2020-10-07
Make lots more usable in python bindings.

QWork66:
The changes add usable versions of python methods related to lots.  The new methods return expected instances of python classes that are bound to gnucash objects.  This allows lots to be examined and manipulated from python.  For example python could be used to experiment with capital gains calculation on investments.


luzpaz:
Any traction on this ?

jralls:
Superseded by #2187.

PR #797, opened 2020-10-11
Speed up some autoclear

christopherlam:
Some harmless optimisations for account Autoclear

- return early if account cleared balance is same as input
- prepend&reverse instead of append

Also
- obey account's sign reversal
- rename 2nd `data` within same function to `sackdata`


christopherlam:
Some more optimisations.

Given a list of unreconciled splits, if the sum of split->amount of any subset is zero, a unique solution cannot be found. Abort early.

christopherlam:
One last refinement would be nice: if auto-clear is successful, it should (offer to) open the target account register, filtered to unreconciled & cleared only, to show the newly-cleared splits. @Bob-IT :)

Bob-IT:
> One last refinement would be nice: if auto-clear is successful, it should (offer to) open the target account register, filtered to unreconciled & cleared only, to show the newly-cleared splits. @Bob-IT :)

@christopherlam, I assume this is a per register action, never used it my self, I see no register menu option.
not sure what it does? 
what feedback do you get currently?

christopherlam:
It's not a register action; it's an account-tree action. It aims to set reconciled status from N to C according to some interesting heuristics. Uses https://en.wikipedia.org/wiki/Knapsack_problem. Writing to devel soon.

christopherlam:
So I think all I need is to open a register for an _account_, filtered to _cleared OR unreconciled_ only.

Bob-IT:
> It's not a register action; it's an account-tree action. It aims to set reconciled status from N to C according to some interesting heuristics. Uses https://en.wikipedia.org/wiki/Knapsack_problem. Writing to devel soon.

@christopherlam  if you do a search for autoClearWindow you will find it works on accounts and there is a callback in the plugin page register but there is no menu option defined.

I assume as part of the autoclear you will have a list of splits and with that you should be able to load a 'general ledger' just like the SLR does if you request to review created transactions.

jralls:
There's an action for it in gnc-plugin-page-register.c but it's not connected to anything. On an Accounts page it's connected to Actions>Auto Clear. ISTM it would make more sense to reverse that so that it's exposed on register pages and not on Accounts ones.

It's a rather bizarre thing to do though. I guess the use case is for when one gets a balance without details about what transactions are included, as one might at an ATM, and one wants to mark some transactions as cleared to get the cleared balance to match. Harmless entertainment, I suppose, but it will fail if one has forgotten to enter a transaction or if one has too many transactions with the same amount.

christopherlam:
Harmless, yes, except if there are 20+ unreconciled splits which means this action causes a UI freeze with no feedback. It should intuitively be runnable from the split register however the latter may have more than 1 source account.

jralls:
See my comment in gnucash-devel about the algorithm.

But frankly I think it would make more sense to just remove it than to put significant effort into making it usable.  It was a drive-by contribution that should have been declined in the first place and I think it unlikely that it's used much.

christopherlam:
@cristiklein a few optimisations -- please confirm the reasoning is acceptable:

1. If `end_value` == `cleared_balance` then there's no need to autoclear. Abort.
2. If too many uncleared splits (20?) then it'll lag. Abort.
3. If the sum a subset of uncleared splits amounts to zero, then the knapsack will always fail to find unique solution. Abort.
4. Select GtkEntry initially and when aborted to make modification easier.

PR #799, opened 2020-10-14
Add mailing lists as footer of translator credits

fellen:
In several languages, the language team is already set to the corresponding gnucash mailing list, but the "Language Team:" entry in the po headers is intended for bots. This PR appends it to the msgstr of msgid "translator-credits" to get it user visible in the translator section of Help->About->Credits.

Language, state, optional review:
de: set since years,
es: this commit, <yoann@laboussole.coop>
it:  set since years, <foti.giuseppe@gmail.com>
nl: this commit, @gjanssens 

Still not applied: [pt_] BR, which has 2 possible lists:
pt [_pt], Pedro is in https://translationproject.org/team/pt.html, "Language Team:" contains the TP address, but gnucash is there marked as external. We should perhaps talk with Pedro.
pt_BR has no conflicts. I will apply it later

### DIFF
--- a/po/es.po
+++ b/po/es.po
@@ -9121,7 +9121,9 @@ msgstr "Contabilidad financiara personal y pequeños negocios."
 msgid "translator-credits"
 msgstr ""
 "Francisco Javier Serrador <serrador@tecknolabs.com>  7-ene-2019\n"
-"Eneko Lacunza <enlar@enlar.net>"
+"Eneko Lacunza <enlar@enlar.net>\n"
+"\n"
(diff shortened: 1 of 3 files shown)


fellen:
> pt_BR has no conflicts. I will apply it later

Done, with an improved term. Perhaps we should change them to something like:
`Send your suggestions or criticism and questions to <ML>`

fellen:
In commit efb9226 the section was heavily improved (using links) for it.po and in commit 5671028 adapted to de.po. See the latter commit has only that change.

PR #800, opened 2020-10-14
Maint 608436

christopherlam:
Offshoot from #797  

@cristiklein a few optimisations -- please confirm the reasoning is acceptable:

1. If `end_value` == `cleared_balance` then there's no need to autoclear. Abort.
2. If too many ~uncleared splits (20?)~ reachable amounts (>1,000,000) then it'll lag. Abort.
3. ~If the sum a subset of uncleared splits amounts to zero, then the knapsack will always fail to find unique solution. Abort.~ - removed -- if split amounts are $1.00, $2.00 and -$3.00, a unique solution exists whereby target amount is $2.00
4. Select GtkEntry initially and when aborted to make modification easier.

### DIFF
--- a/gnucash/gnome/window-autoclear.c
+++ b/gnucash/gnome/window-autoclear.c
@@ -25,18 +25,14 @@
 #include <gtk/gtk.h>
 #include <glib/gi18n.h>
 
-#include "Scrub.h"
-#include "dialog-account.h"
-#include "dialog-transfer.h"
 #include "dialog-utils.h"
 #include "gnc-amount-edit.h"
-#include "gnc-component-manager.h"
-#include "gnc-date-edit.h"
 #include "gnc-event.h"
 #include "gnc-gnome-utils.h"
 #include "gnc-main-window.h"
 #include "gnc-plugin-page-register.h"
 #include "gnc-ui.h"
+#include "gnc-ui-balances.h"
 #include "window-autoclear.h"
 
 #define WINDOW_AUTOCLEAR_CM_CLASS "window-autoclear"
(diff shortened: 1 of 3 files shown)


cristiklein:
@christopherlam I generally like the ideas you suggest, however, one of the criticism that I have towards my past work is that it is not tested. Each such optimization should ideally come with some kind of unit test that shows that, indeed, time has significantly gone down. I was kind of hoping that, as people would start using the feature, they would start complaining about it and giving examples of where it falls short. 10 years later, it does not feel to me like this ended up being a killer feature. :smile: 

Instead of limiting the number of uncleared splits, we could add a progress bar and a cancel button, as well as make the clearing process not block the main thread. If all works well, the user should be able to decide for themselves if waiting $n$ minutes for auto-clearing is worth or not.

cristiklein:
To clarify on the complexity of the algorithm, the best knapsack solvers out there have a complexity of O(nW), where W is the "capacity" and n the number of items. For the auto-clear problem, W is equal to the uncleared amount and n the uncleared number of transactions. So if you want to clear from 100 to 3500, 20 uncleared transactions, you would need to perform 350000*20 operations (GnuCash accounts generally use 2 digit decimals, hence multiply by 100). I doubt this is optimal in too many cases, which is why I chose a "shittier" algorithm, whose complexity is in the number of uncleared splits. However, more testing and user feedback could reveal that the O(nW) algorithm might perform better.

christopherlam:
> I was kind of hoping that, as people would start using the feature, they would start complaining about it and giving examples of where it falls short. 10 years later, it does not feel to me like this ended up being a killer feature. :smile:

...because it's not been documented anywhere!

I did additional logging and the split vs reachable amount seems 2^n hence a progressbar would look very silly, quickly reaching 98% then stuck at 99% for 20s whereby 25th split is being worked.

> For the auto-clear problem, W is equal to the uncleared amount and n the uncleared number of transactions. So if you want to clear from 100 to 3500, 20 uncleared transactions, you would need to perform 350000*20 operations (GnuCash accounts generally use 2 digit decimals, hence multiply by 100). I doubt this is optimal in too many cases, which is why I chose a "shittier" algorithm, whose complexity is in the number of uncleared splits. However, more testing and user feedback could reveal that the O(nW) algorithm might perform better.

Feel free to dive back in... I'm not sure how to apply knapsack because classical knapsack assumes positive weights, and doesn't assume we have to reach target amount. This one uses +/- weights and we *must* reach target amount. This PR will shorten code and maybe a better algorithm can be plugged in.

christopherlam:
Please see patch and resulting log

```patch
Unstaged changes (1)
modified   gnucash/gnome/window-autoclear.c
@@ -135,6 +135,7 @@ gnc_autoclear_window_ok_cb (GtkWidget *widget,
     gnc_numeric toclear_value;
     GHashTable *sack;
     gchar *errmsg = NULL;
+    gint num, idx;
 
     g_return_if_fail (widget && data);
 
@@ -165,7 +166,8 @@ gnc_autoclear_window_ok_cb (GtkWidget *widget,
         goto skip_knapsack;
     }
 
-    if (g_list_length (nc_list) > 20)
+    num = g_list_length (nc_list);
+    if (num > 20)
     {
         errmsg = _("Too many uncleared splits");
         goto skip_knapsack;
@@ -187,16 +189,17 @@ gnc_autoclear_window_ok_cb (GtkWidget *widget,
      *  - value = last split we used to clear this amount (not managed by GHashTable)
      */
     PINFO("Knapsacking ...\n");
-    for (node = nc_list; node; node = node->next)
+    for (idx=0, node = nc_list; node; ++idx, node = node->next)
     {
         Split *split = (Split *)node->data;
         gnc_numeric split_value = xaccSplitGetAmount(split);
+        gint j=0;
 
         struct _sack_foreach_data_t s_data[1];
         s_data->split_value = split_value;
         s_data->reachable_list = 0;
 
        PINFO("  Split value: %s\n", gnc_numeric_to_string(split_value));
 
         /* For each value in the sack, compute a new reachable value */
         g_hash_table_foreach (sack, sack_foreach_func, s_data);
@@ -206,7 +209,7 @@ gnc_autoclear_window_ok_cb (GtkWidget *widget,
             (s_data->reachable_list, g_memdup (&split_value, sizeof (gnc_numeric)));
 
         /* Add everything to the sack, looking out for duplicates */
-        for (GList *s_node = s_data->reachable_list; s_node; s_node = s_node->next)
+        for (GList *s_node = s_data->reachable_list; s_node; ++j, s_node = s_node->next)
         {
             gnc_numeric *reachable_value = s_node->data;
 
@@ -230,6 +233,8 @@ gnc_autoclear_window_ok_cb (GtkWidget *widget,
                 g_hash_table_insert (sack, reachable_value, split);
             PINFO("\n");
         }
+
+        printf ("split_idx=%d, num_reachable_values=%d\n", idx, j);
         g_list_free(s_data->reachable_list);
     }
```

Log:
```
split_idx=0, num_reachable_values=1
split_idx=1, num_reachable_values=2
split_idx=2, num_reachable_values=4
split_idx=3, num_reachable_values=8
split_idx=4, num_reachable_values=16
split_idx=5, num_reachable_values=32
split_idx=6, num_reachable_values=64
split_idx=7, num_reachable_values=128
split_idx=8, num_reachable_values=256
split_idx=9, num_reachable_values=512
split_idx=10, num_reachable_values=1024
split_idx=11, num_reachable_values=2048
split_idx=12, num_reachable_values=4096
split_idx=13, num_reachable_values=8192
```

This is an account whose splits have amounts $1, $2, $4, $8, $16... thereby generating unique reachable_amounts for each additional split. idx>20 means looping >1Million amounts...

cristiklein:
@christopherlam The log you sent is really good to illustrate the problem and the algorithmic complexity of auto-clear. However, I am unsure what the solution would be. If we take a "smarter" Knapsack algorithm, that would have the complexity O(nW), so in the example you sent, it would mean n=13, W=91. Hence, Knapsack could be solved within 1183 operations. However, if you take into account that balances usually have double-digit resolution, then W=9100, so 118300 operations. This compares favorably with 8192 reachable values that the "less smart" algorithm had to devise.

However, let's assume that we want to support up to 20 splits and that we expect clearing never to exceed 10000$. This would mean that we could reduce the complexity of the algorithm from 2^20 to 20*10000*100, so 20,000,000 operations.

How much do you need this feature and how eager are you to further develop it? If we want to further develop it, I would move the main auto-clear logic from the UI code into "libgnucash" folder. Then I would develop some tests: For start, I would write a GnuCash file with 20 transactions, 7 already cleared, and 13 more to auto-clear. We could test for the following scenarios:

- All transactions need to be cleared;
- No transactions need to be cleared;
- All but the last need to be cleared;
- All but the first need to be cleared;
- All but the previous to the last need to be cleared;

Once we have these tests in place, I feel we could do a better job at actually measuring runtimes and the impact that various optimizations have.

What do you think?

christopherlam:
> However, let's assume that we want to support up to 20 splits and that we expect clearing never to exceed 10000$. This would mean that we could reduce the complexity of the algorithm from 2^20 to 20_10000_100, so 20,000,000 operations.

The 20 splits limit seems too arbitrary; I've changed to abort when num(reachable-values) > 1,000,000.

My interest in this feature is to perform general cleanup and possibly create some tests. It could move to libgnucash utility library which would ease creation of these tests.

cristiklein:
Nice! I like the API of `gnc_account_get_autoclear_splits`.

christopherlam:
Thanks; this is currently incorrect because I can't figure out mutable arguments in C and mutating an argument to a gchar error message.

jralls:
> Thanks; this is currently incorrect because I can't figure out mutable arguments in C and mutating an argument to a gchar error message.

Pass the argument as a pointer:
```
static void
foo(int* x)
{
    if (x)
      *x = 42;
}

int
main()
{
   int arg = 0;
   foo(&arg);
   printf ("Arg is %d\n", arg);
   return 0;
}
```
Will print "Arg is 42".

jralls:
@cristiklein You're looking at the wrong complexity value in Wikipedia's [Knapsack Problem](https://en.wikipedia.org/wiki/Knapsack_problem#Dynamic_programming_in-advance_algorithm). O(nW) is for the *unbounded* knapsack problem where one might have any number of items of any weight in the w vector. That's not the problem you're solving. The one you're solving is the 0-1 knapsack problem, but even the solutions proposed in that Wikipedia article are flawed, at least for the highly constrained version of the problem here. Those algorithms iterate over every possible value 0..W. Doing so when your objective is to find a combination of w<sub>i</sub> that sums to W would be beyond stupid.
A much better and more correct algo:
Sort w<sub>0</sub>...w<sub>n</sub>. Chop it off at w<sub>i</sub> < W. Note that if there's a w<sub>i</sub> == W then you've found one solution. Let that i be n'.
Construct an n' x n' binary array X and take the cross product of it and w<sub>0</sub>...w<sub>n'</sub>, then sum the results into W'<sub>0..n'</sub>. Iterate over that result array testing W'<sub>i</sub> == W. Any successes are possible solutions. 

In that algorithm the dominant calculation is computing the cross-product, which will be O(2<sup>n</sup>) only if you must compute every possible sum, but you probably don't.

Next, *don't* use linked lists and hashed linked lists past the sort-by-amount stage. Those are incredibly inefficient containers for this kind of algorithm. Once you've sorted the list of uncleared splits and found n' allocate a `Split*[n' + 1] s` and load that with the splits in amount order. 
The brute-force next step would iterate over 0..2<sup>n'</sup> and use the iterator as a bitfield to select which members of  s to include in that iteration's sum, compute the sum, test it against W, and save the iterator if they're equal. If you find a second equality you can abort the iteration. Considering that the split*[] is sorted by amount the sums will generally increase though not necessarily monotonically (as a simple experiment with w=(3,5,7,11) will show) but that could be used to optimize the search further: E.g. if the result R<sub>2<sup>n'-1</sup>-1</sub> < W there's no need to test lower indexes; similarly if R<sub>2<sup>n'-1</sup>+1</sub > W there's no need to test higher ones. I think a strategy along those lines could get the complexity pretty close to O(logn).

jralls:
> How much do you need this feature and how eager are you to further develop it? 

I don't think it's used much, nor am I convinced that it's particularly useful. The only use-case I can imagine is one where one gets a balance without the transaction information as happens when one uses an ATM. If one has online access to one's bank and can check the actual cleared transactions one is much better off doing that.

Don't let that dissuade you from writing unit tests using Googletest if you feel so inclined.

christopherlam:
> Pass the argument as a pointer: (snip)
> Will print "Arg is 42".

In this case I want to mutate argument would be a pointer to a newly-allocated string?

jralls:
> In this case I want to mutate argument would be a pointer to a newly-allocated string?

You mean you want to pass in an already-initialized string and change its contents? That's generally a bad idea unless you can guarantee that the new string won't be any larger than the old one. But you can do something like
```
static void
foo(char** string)
{
    *string = g_strdup_printf("The answer to life, the universe, and everything is %d", 42);
}

int main()
{
    char* the_answer = NULL;
    foo (&the_answer);
    printf ("%s\n", the_answer);
    return 0;
}
```
Alternatively a common idiom is to pick a size you're sure will be big enough:
```
static void
foo(char* str, size_t len)
{
     const char* message = "The answer is 42!";
     strncpy(str, message, len - 1);
}

int
main()
{
     char buffer[100];
     memset(buffer, 0, sizeof(buffer));
     foo(buffer, sizeof(buffer);
     printf("%s\n", buffer);
     return 0;
}
```

cristiklein:
> > How much do you need this feature and how eager are you to further develop it?
> 
> I don't think it's used much, nor am I convinced that it's particularly useful. The only use-case I can imagine is one where one gets a balance without the transaction information as happens when one uses an ATM. If one has online access to one's bank and can check the actual cleared transactions one is much better off doing that.

I was using this features as follows: I was rather disciplined and entered each transaction in GnuCash at the end of the day. Then, once a week, maybe 13 uncleared transactions accumulated and 10 were recorded by my bank, in a rather arbitrary order. Of course, I could have gone through each bank transaction and clear each split individually. However, given how "unique" the amounts were, I rather saved some time and just told GnuCash to auto-clear the transactions based on the balance on my bank. It simply felt a lot more efficient than having to find transactions one-by-one.

> In that algorithm the dominant calculation is computing the cross-product, which will be O(2n) only if you must compute every possible sum, but you probably don't.

My experience -- but I'm happy to get counter-examples -- is that for my use-case above these optimizations wouldn't bring a lot of value and almost all combinations need to be checked. Let me explain with an example. Given the amounts:

    -92 -29 -612 -79 -19.9 -109 -444 -636.1 -27 -154 -39 -220.42 1600

auto-clear to -4.9. These are actual values taken today from my checking account. (Currency is SEK, I'm not that rich. :smile:)

I am unsure to understand how any of your suggestions could reduce complexity in the general case. I totally agree that there are special cases -- like only one transaction needs to be cleared or all transactions need to be cleared -- that could be optimized. However, those are not the ones I personally encountered.

christopherlam:
Thanks @jralls @cristiklein it now works. `gnc-ui-balances.c` seems an appropriate location to park this algorithm. Last wishlist would be: after successful clear, open a register with all `toclear_list` because we do not want invisible magic. It now just needs documentation ideally in the Tutorial & Concepts guide.

For now the abort criteria are: `target_balance` == `current_cleared_balance`; or `num(reachable_amounts) > 1 million`. These seem reasonable.

christopherlam:
In case it's not obvious the reason for changing abort-critera to num(reachable)>1mil, the reasoning is: if uncleared splits have amounts `100.00` and `20 * 1.00`, the num(reachable) will be small (=41?), and if the target_amount is `100.00` then there is a unique solution, easily found.

christopherlam:
Output whereby Auto-clear amount needs to clear $350.00. Unfortunately the tab name is not set to 'Cleared Transactions' as desired.
image

fellen:
> It now just needs documentation ideally in the Tutorial & Concepts guide.

I think for you as developers it would be easier to add a section to help. If one exists use the order of the menu entries to find the right place. This will also allow you to add a button with context sensitive help (the section id in help).

Some often missed parts: In which context does it get used and why - hint: why did you develope it? Under which circumstances should it not be used, if any exist? Do alternative methodes exist? Finally the meaning of each GUI element should be explained.

See also https://wiki.gnucash.org/wiki/Documentation_Update_Instructions

jralls:
>  It now just needs documentation ideally in the Tutorial & Concepts guide.

Oh? I don't see any unit tests although @cristiklein said he'd write some.

cristiklein:
> > It now just needs documentation ideally in the Tutorial & Concepts guide.
> 
> Oh? I don't see any unit tests although @cristiklein said he'd write some.

I can do that next week, if @christopherlam isn't faster than me. :smile:

Bob-IT:
> Output whereby Auto-clear amount needs to clear $350.00. Unfortunately the tab name is not set to 'Cleared Transactions' as desired.

@christopherlam If you want to change the tab name, change show_cleared_splits to return the 'page' instead of void and then add this after that function call main_window_update_page_name (page, _("Cleared Transactions"));

I did wonder if the title should include the account name but it will make the tab text longer and there is no tool tip for search pages. Also I noticed if the autoclear fails, it displays an error message but it is not cleared if you change the amount and press 'OK'.

christopherlam:
> @christopherlam If you want to change the tab name, change show_cleared_splits to return the 'page' instead of void and then add this after that function call main_window_update_page_name (page, _("Cleared Transactions"));

Thanks @Bob-IT . Interestingly, calling `main_window_update_page_name (page, _("Cleared Transactions"));` from within `show_cleared_splits` didn't work; I needed to run `gdb` with breakpoint (+1 XP) to find that `xaccAccountCommitEdit` would somehow set the tab name back to 'Search Results'. Now works well. Opens _journal_ view with cleared splits. I'd prefer an account _ledger_ filtered to newly cleared splits, but this will do.

> I did wonder if the title should include the account name but it will make the tab text longer and there is no tool tip for search pages. Also I noticed if the autoclear fails, it displays an error message but it is not cleared if you change the amount and press 'OK'.

Can't reproduce. Works very well for me -- both clearing and aborting.

image

Bob-IT:
> Can't reproduce. Works very well for me -- both clearing and aborting.
> 
@christopherlam Maybe I did not explain it very well but I still see it, if you enter an amount that can not be cleared the status_label  displays error message 'The selected amount cannot be cleared', if you decide to correct the amount and try again the status_label will still be showing the error message until the procedure returns.

Here is what i did to change that...
Add this cb function,

static void
clear_status_label_cb (GtkEditable *editable, AutoClearWindow *data)
{
    gtk_label_set_text  (data->status_label, "");
}

connected to this signal...
g_signal_connect (data->end_value, "changed", G_CALLBACK(clear_status_label_cb), data);

I also contemplated adding these lines before your g_free (errmsg) but maybe not such a good idea...
        if (g_strcmp0 (errmsg, _("The selected amount cannot be cleared.")) == 0)
            gtk_widget_grab_focus (GTK_WIDGET(data->end_value));

cristiklein:
In the meanwhile, I worked on adding some tests to the auto-clear core: https://github.com/cristiklein/gnucash/tree/maint-608436-ck

Unfortunately, I don't managed to open the book:

```
$ mkdir -p build && cd build && cmake .. && make VERBOSE=1 -j4
[...]
$ make VERBOSE=1 test-app-utils && env CTEST_OUTPUT_ON_FAILURE=1 ctest -R app-utils
[...]
2/2 Test #9: test-app-utils ...................***Failed    0.26 sec
# random seed: R02Se2cd2b061767e86bfb3e442286762543
# gnc.backend.dbi-INFO: [gnc_module_init_backend_dbi] GNC_DBD_DIR not set: using libdbi built-in default

# gnc.backend.dbi-INFO: [gnc_module_init_backend_dbi] 1 DBD drivers found

# gnc.backend.dbi-INFO: [gnc_module_init_backend_dbi] Driver: sqlite3

Blah
1..4
# Start of app-utils tests
# Start of autoclear tests
/home/cklein/personal/gnucash/build/libgnucash/app-utils/test
total 52
drwxrwxr-x 10 cklein cklein  4096 okt 19 10:19 CMakeFiles
-rw-rw-r--  1 cklein cklein  1145 okt 19 10:17 cmake_install.cmake
-rw-rw-r--  1 cklein cklein 12925 okt 19 10:17 CTestTestfile.cmake
-rw-rw-r--  1 cklein cklein 23075 okt 19 10:17 Makefile
drwxrwxr-x  2 cklein cklein  4096 okt 19 10:21 test-files
total 12
-rw-rw-r-- 1 cklein cklein 11820 okt 19 10:21 autoclear.gnucash
# qof.session-INFO: [QofSessionImpl::load_backend]  Selected provider GnuCash Libdbi (SQLITE3) Backend
# gnc.backend.dbi-INFO: [QofDbiBackendProvider<Type>::type_check] exists, does not have SQLite format string -> not DBI
# qof.session-INFO: [QofSessionImpl::load_backend] Provider, GnuCash Libdbi (SQLITE3) Backend, reported not being usable for book, test-files/autoclear.gnucash.
# qof.session-INFO: [QofSessionImpl::load_backend] The provider providers access_method, sqlite3, but we're loading for access_method, file. Skipping.
# qof.session-INFO: [QofSessionImpl::load_backend] The provider providers access_method, xml, but we're loading for access_method, file. Skipping.
# qof.session-INFO: [QofSessionImpl::load_backend]  Selected provider GnuCash File Backend Version 2
# gnc.backend-INFO: [GncXmlBackend::session_begin] logpath=test-files/autoclear.gnucash
# qof.session-INFO: [QofSessionImpl::begin] Done running session_begin on backend
Bail out! gnc.engine-FATAL-CRITICAL: void QofSessionImpl::load(QofPercentageFunc): assertion 'm_book && qof_book_empty(m_book)' failed

(/home/cklein/personal/gnucash/build/bin/test-app-utils:495524): gnc.engine-CRITICAL **: 10:21:18.369: void QofSessionImpl::load(QofPercentageFunc): assertion 'm_book && qof_book_empty(m_book)' failed
SIGTRAP
```

Is this the correct way to open a file?

```
    qof_session_begin (fixture->session, TEST_FILENAME, SESSION_NORMAL_OPEN);
    qof_session_load (fixture->session, NULL);
```

christopherlam:
Interesting...I never managed to load a datafile for tests. All my tests needs to create data using API. This would ease creation of tests in the future especially for testing backward compatibility of data file formats. @jralls is probably best placed to assist.

jralls:
@cristiklein You need to create a `QofBook` and then do `qof_session_new(book);`.

Please use [GoogleTest](https://github.com/google/googletest/blob/master/googletest/docs/primer.md) instead of GLib tests for new test programs. It's vastly easier to use and besides GLib has changed the testing framework and deprecated the way that we use it.

jralls:
@cristiklein However @christopherlam is correct: You shouldn't need to create a whole book for unit testing the `gnc_account_get_autoclear_splits`, you just need to create a list of splits and a fake `xaccAccountGetSplitList()` to retrieve it.  You'll need at least three lists, one that's solvable, one that isn't, and one that has multiple solutions.

cristiklein:
@jralls Indeed, I felt the test file was making the tests someone intransparent. I'll go for programatic generation of test data.

christopherlam:
Fixup and merged. @cristiklein very welcome to rebase from maint for appropriate tests.

christopherlam:
Oops. (Always after merging) Function should have been written as follows to follow gnome rules:

```c
gboolean gnc_account_get_autoclear_splits (Account *account, gnc_numeric toclear_value,
                                           GList **toclear_splits, gchar **errmsg);
```

PR #801, opened 2020-10-15
Fix warning about misleading indentation from recent gcc.

wenottingham:
Patch by <law@redhat.com>


christopherlam:
Cherry-picked into maint 06110e2

PR #802, opened 2020-10-18
Maint 797982 - increase precision for exchange rates table

christopherlam:
Initially my approach was to modify `gnc:html-make-exchangerates` but it receives a generic `exchange-fn` which takes original-amount and target-currency.

This approach will upgrade the pricedb exchange-fn to consider more precision. This requires an underlying addition to `gnc-pricedb.c` functions to accept a `fraction` argument.

### DIFF
--- a/gnucash/report/html-utilities.scm
+++ b/gnucash/report/html-utilities.scm
@@ -207,6 +207,8 @@
 ;; function 'exchange-fn' and the 'accounts' determine which
 ;; commodities to show. Returns a html-object, a <html-table>.
 (define (gnc:html-make-exchangerates common-commodity exchange-fn accounts)
+  (issue-deprecation-warning
+   "gnc:html-make-exchangerates is deprecated. use gnc:html-make-rates-table instead.")
   (let ((comm-list (gnc:accounts-get-commodities accounts common-commodity))
         (markup (lambda (c) (gnc:make-html-table-cell/markup "number-cell" c)))
         (table (gnc:make-html-table)))


christopherlam:
Addendum- need to upgrade `gnc-pricedb.c` instead of navigating the pricedb in guile because the C converter has ability to convert currencies using an intermediate currency. To replicate this in guile means duplication of work.

jralls:
You've gone off in the wrong direction, maybe because the original code to generate the price table in reports did. Amounts and values must always be rounded to SCU because fractions of SCU are not legal in modern times. Chopping a gold Spanish dollar into 8 pieces isn't a thing any more.

The price table in reports needs to get the price, not the amount of 1 X in currency Y, and it needs to display it according to the force-price-to-decimal preference: SCU/100 if the preference is set and exact, with a fraction if necessary, if not.

christopherlam:
Sure, but this will mean more logic needs to be in scheme; the exchange table in reports is capable of showing the artificial price using an intermediate currency if direct conversion is impossible.

jralls:
It doesn't *have* to be in Scheme. It's not there already, it's just piggy-backing on the exchange logic in `gnc-pricedb.c`. Just extract the price logic from [indirect_balance_conversion](https://github.com/Gnucash/gnucash/blob/maint/libgnucash/engine/gnc-pricedb.c#L2547) into a separate function and add calls to it in the price lookup functions.

christopherlam:
Commits look complicated because price retrieval was intricately linked with the amount multiplier. Pulls out price retriever `gnc_pricedb_get_price_at_date` -- will try direct conversion; if this fails tries an intermediate currency. Exchange-rate table can use this price retriever for reporting.

Perhaps in the future an arbitrary number of intermediate currencies can be used; https://gist.github.com/christopherlam/0f14f47e1d222c080675e147f7605f6d

libgnucash/engine/gnc-pricedb.c
```diff
@@ -2513,6 +2513,93 @@ extract_common_prices (PriceList *from_prices, PriceList *to_prices,
...
+    PriceTuple tuple;
+    gnc_numeric zero = gnc_numeric_zero();
+    if (!from || !to)
+        return zero;
+    if (t == INT64_MAX)
+    {
+        from_prices = gnc_pricedb_lookup_latest_any_currency(db, from);
+        /* "to" is often the book currency which may have lots of prices,
+            so avoid getting them if they aren't needed. */
+        if (from_prices)
```
jralls: Shouldn't this be `!from_prices`, there being no point in getting `to_prices` if one has found `from_prices`?

jralls:
The one comment aside (which would be my mistake from some years ago), looks pretty good.

libgnucash/engine/gnc-pricedb.c
christopherlam: I haven't looked in detail; I think the underlying intermediate-currency pricing does require both `to_prices` and `from_prices`; if one of them is missing then there's no point retrieving the other.

christopherlam:
While this is being worked on @jralls assistance kindly requested. The `html-style-info.scm` change ensures price is possibly rendered exact, however if global preference to force decimal is enabled, a g_warning about "rounding required when never round is selected" is emitted.

gnucash/report/commodity-utilities.scm
```diff
@@ -764,6 +765,18 @@
...
 
+(define (gnc:case-price-fn source target-curr date)
+  (define pdb (gnc-pricedb-get-db (gnc-get-current-book)))
+  (case source
+    ((pricedb-nearest) (cut gnc-pricedb-get-price-at-date pdb <> target-curr date))
+    ((pricedb-latest)  (cut gnc-pricedb-get-price-latest pdb <> target-curr))
+    (else
+     (lambda (commodity)
+       (let* ((exchange-fn (gnc:case-exchange-fn source target-curr date))
+              (foreign-mon (gnc:make-gnc-monetary commodity 1))
```
jralls, line 776: @christopherlam You might put back @cstim's old hack where you convert 1000 instead of 1. Then you can div that by 1000 and have 6 decimals for your price.

jralls:
@christopherlam Meaning that some operation is called with `GNC_HOW_RND_NEVER`. Without more info I'd have to guess that it's coming from `convert_price` where it's used in a division; I think that the only other place `GNC_HOW_RND_NEVER` is used that would matter is in `PrintAmountInternal` but it's used with `GNC_DENOM_AUTO` in a subtraction of the integer part of a number from the number and that shouldn't cause rounding, but the warning message should tell you the function that issued the warning. You can set a breakpoint on the warning line in that function and back up a frame to see what function called to gnc_numeric operation and what values were passed.

gnucash/report/commodity-utilities.scm
christopherlam: I don't think this is needed. I've confirmed that: for all 4 price sources, and force-decimal true/false, the above code works well. Report currency USD:
£1.00	$1.8570
JP¥1	$0.0096

£1.00	$1.3484
JP¥1	$0.0096

£1.00	$1 + 24967/71671
JP¥1	$1/104

£1.00	$1 + 1157/1350
JP¥1	$1/104

For report currency JPY:

nearest in time, force-decimal off
£1.00	JP¥138.00
$1.00	JP¥104.00

average-cost, force-decimal off
£1.00	JP¥148 + 76956/128329
$1.00	JP¥104.00

average-cost, force-decimal on
£1.00	JP¥148.60
$1.00	JP¥104.00

nearest-in-time, force-decimal on
£1.00	JP¥138.00
$1.00	JP¥104.00

christopherlam:
> @christopherlam Meaning that some operation is called with `GNC_HOW_RND_NEVER`. Without more info I'd have to guess that it's coming from `convert_price` where it's used in a division; I think that the only other place `GNC_HOW_RND_NEVER` is used that would matter is in `PrintAmountInternal` but it's used with `GNC_DENOM_AUTO` in a subtraction of the integer part of a number from the number and that shouldn't cause rounding, but the warning message should tell you the function that issued the warning. You can set a breakpoint on the warning line in that function and back up a frame to see what function called to gnc_numeric operation and what values were passed.

I think this arises when price is a gnc_numeric e.g. `price = 19069648/128329`, and we're calling `(xaccPrintAmount price (gnc-price-print-info currency #t))` with print-info returning `(list 'print-info JPY 2 2 #t #t #t #t #f #f)` with the result being string `JP¥148 + 76956/128329`

christopherlam:
> @christopherlam Meaning that some operation is called with `GNC_HOW_RND_NEVER`. Without more info I'd have to guess that it's coming from `convert_price` where it's used in a division; I think that the only other place `GNC_HOW_RND_NEVER` is used that would matter is in `PrintAmountInternal` but it's used with `GNC_DENOM_AUTO` in a subtraction of the integer part of a number from the number and that shouldn't cause rounding, but the warning message should tell you the function that issued the warning. You can set a breakpoint on the warning line in that function and back up a frame to see what function called to gnc_numeric operation and what values were passed.

Confirmed. Try open a regular book with numerous stock transactions. Set pref force decimal to False. Open stock register. It'll show prices with exact fractions but numerous numeric warnings.

gnucash/report/commodity-utilities.scm
jralls: OK, the one of interest is average-cost. I'm a bit surprised that it doesn't round to JPY's SCU (1) given your earlier complaint and that AFAICT you haven't changed anything related to average-cost or weighted-average.

christopherlam: I think this JPY issuing 2 decimal points is a completely separate issue; I'd merge this branch first then investigate JPY separately.

christopherlam:
@jralls see latest commit for utest-pricedb. The numbers show exact prices e.g. 113289/161643 for USD->EUR. I think this is exactly what we want.

christopherlam:
Last dilemma:
`gnc_pricedb_get_price_at_date` and `gnc_pricedb_get_price_latest`, or
`gnc_pricedb_get_nearest_price` and `gnc_pricedb_get_latest_price`?

PR #805, opened 2020-10-23
Auto clear tests

cristiklein:
Hi @christopherlam,

Please find here a first draft of the autoclear tests. Unfortunately, right now the tests fail. The reason is that the test cases include two splits of the same amount. This leads to "Cannot uniquely clear splits. Found multiple possibilities.". This is a limitation of the current algorithm. It could be smarter and suggest a list of splits to clear when none or both of the two splits are included.

How should we proceed?
1. Accept the limitations of the algorithm and simplify the tests accordingly.
2. Improve the algorithm so it can deal with trickier cases.

Let me know what you think.

### DIFF
--- a/libgnucash/app-utils/test/test-autoclear.cpp
+++ b/libgnucash/app-utils/test/test-autoclear.cpp
@@ -31,7 +31,7 @@ extern "C" {
 
 #include "gtest/gtest.h"
 
-const gint64 DENOM = 100; //< Denomerator is always 100 for simplicity.
+static const int64_t DENOM = 100; //< Denomerator is always 100 for simplicity.
 
 struct SplitDatum {
     const char *memo;
@@ -39,100 +39,108 @@ struct SplitDatum {
     bool cleared;
 };
 
-struct TestCase {
+struct Tests {
     gint64 amount;
     const char *expectedErr;
 };
 
-std::vector<SplitDatum> easySplitData = {
-    { "Memo 01", -  8234, true },
-    { "Memo 02", -156326, true },
-    { "Memo 03", -  4500, true },
-    { "Memo 04", -694056, true },
-    { "Memo 05", -  7358, true },
-    { "Memo 06", - 11700, true },
-    { "Memo 07", - 20497, true },
-    { "Memo 08", - 11900, true },
-    { "Memo 09", -  8275, true },
-    { "Memo 10", - 58700, true },
-    { "Memo 11", +100000, true },
-    { "Memo 12", - 13881, true },
-    { "Memo 13", -  5000, true },
-    { "Memo 14", +200000, true },
-    { "Memo 15", - 16800, true },
-    { "Memo 16", -152000, true },
-    { "Memo 17", +160000, false },
-    { "Memo 18", - 63610, false },
-    { "Memo 19", -  2702, false },
-    { "Memo 20", - 15400, false },
-    { "Memo 21", -  3900, false },
-    { "Memo 22", - 22042, false },
-    { "Memo 23", -  2900, false },
-    { "Memo 24", - 10900, false },
-    { "Memo 25", - 44400, false },
-    { "Memo 26", -  9200, false },
-    { "Memo 27", -  7900, false },
-    { "Memo 28", -  1990, false },
-    { "Memo 29", -  7901, false },
-    { "Memo 30", - 61200, false },
+struct TestCase {
+    std::vector<SplitDatum> splits;
+    std::vector<Tests> tests;
 };
 
-std::vector<TestCase> easyTestCases = {
-    { 0, "The selected amount cannot be cleared.", },
-    { -869227, "Account is already at Auto-Clear Balance." }, // No splits need to be cleared.
-    { -869300, "The selected amount cannot be cleared." },
-    { -869230, NULL },
-    { -963272, NULL }, // All splits need to be cleared.
+TestCase easyTestCase = {
+    .splits = {
+        { "Memo 01", -  8234, true },
+        { "Memo 02", -156326, true },
+        { "Memo 03", -  4500, true },
+        { "Memo 04", -694056, true },
+        { "Memo 05", -  7358, true },
+        { "Memo 06", - 11700, true },
+        { "Memo 07", - 20497, true },
+        { "Memo 08", - 11900, true },
+        { "Memo 09", -  8275, true },
+        { "Memo 10", - 58700, true },
+        { "Memo 11", +100000, true },
+        { "Memo 12", - 13881, true },
+        { "Memo 13", -  5000, true },
+        { "Memo 14", +200000, true },
+        { "Memo 15", - 16800, true },
+        { "Memo 16", -152000, true },
+        { "Memo 17", +160000, false },
+        { "Memo 18", - 63610, false },
+        { "Memo 19", -  2702, false },
+        { "Memo 20", - 15400, false },
+        { "Memo 21", -  3900, false },
+        { "Memo 22", - 22042, false },
+        { "Memo 23", -  2900, false },
+        { "Memo 24", - 10900, false },
+        { "Memo 25", - 44400, false },
+        { "Memo 26", -  9200, false },
+        { "Memo 27", -  7900, false },
+        { "Memo 28", -  1990, false },
+        { "Memo 29", -  7901, false },
+        { "Memo 30", - 61200, false },
+    },
+    .tests = {
+        { 0, "The selected amount cannot be cleared.", },
+        { -869227, "Account is already at Auto-Clear Balance." }, // No splits need to be cleared.
+        { -869300, "The selected amount cannot be cleared." },
+        { -869230, NULL },
+        { -963272, NULL }, // All splits need to be cleared.
+    },
 };
 
-std::vector<SplitDatum> ambiguousSplitData = {
-    { "Memo 01", -10, false },
-    { "Memo 02", -10, false },
-    { "Memo 03", -10, false },
+TestCase ambiguousTestCase = {
+    .splits = {
+        { "Memo 01", -10, false },
+        { "Memo 02", -10, false },
+        { "Memo 03", -10, false },
+    },
+    .tests = {
+        { -10, "Cannot uniquely clear splits. Found multiple possibilities." },
+        { -20, "Cannot uniquely clear splits. Found multiple possibilities." },
+
+        // Forbid auto-clear to be too smart. We expect the user to manually deal
+        // with such situations.
+        { -30, "Cannot uniquely clear splits. Found multiple possibilities." },
+    },
 };
 
-std::vector<TestCase> ambiguousTestCases = {
-    { -10, "Cannot uniquely clear splits. Found multiple possibilities." },
-    { -20, "Cannot uniquely clear splits. Found multiple possibilities." },
+class AutoClearTest : public testing::TestWithParam<TestCase> {
+protected:
+    std::shared_ptr<QofBook> book_;
+    Account *account_; // owned by book_
+    TestCase testCase_;
 
-    // Commented out, auto-clear algorithm is not smart enough yet.
-    //{ -30, NULL },
-};
+public:
+    AutoClearTest() :
+        book_(qof_book_new(), qof_book_destroy),
+        account_(xaccMallocAccount(book_.get()))
+    {
+        testCase_ = GetParam();
 
-class AutoClearTest
-    : public testing::TestWithParam<
-        std::pair<
-            std::vector<SplitDatum>,
-            std::vector<TestCase>
-        >
-    > {
+        xaccAccountSetName(account_, "Test Account");
+        xaccAccountBeginEdit(account_);
+        for (auto &d : testCase_.splits) {
+            Split *split = xaccMallocSplit(book_.get());
+            xaccSplitSetMemo(split, d.memo);
+            xaccSplitSetAmount(split, gnc_numeric_create(d.amount, DENOM));
+            xaccSplitSetReconcile(split, d.cleared ? CREC : NREC);
+            xaccSplitSetAccount(split, account_);
+
+            gnc_account_insert_split(account_, split);
+        }
+        xaccAccountCommitEdit(account_);
+    }
 };
 
 TEST_P(AutoClearTest, DoesAutoClear) {
-    auto splitData = GetParam().first;
-    auto testCase = GetParam().second;
-
-    QofBook *book = qof_book_new ();
-    Account *account = xaccMallocAccount(book);
-    xaccAccountSetName(account, "Test Account");
-
-    xaccAccountBeginEdit(account);
-    for (auto &d : splitData) {
-        Split *split = xaccMallocSplit(book);
-        xaccSplitSetMemo(split, d.memo);
-        xaccSplitSetAmount(split, gnc_numeric_create(d.amount, DENOM));
-        xaccSplitSetReconcile(split, d.cleared ? CREC : NREC);
-        xaccSplitSetAccount(split, account);
-
-        gnc_account_insert_split(account, split);
-    }
-    xaccAccountCommitEdit(account);
-
-    for (auto &t : testCase) {
+    for (auto &t : testCase_.tests) {
         gnc_numeric amount_to_clear = gnc_numeric_create(t.amount, DENOM);
         char *err;
 
-        GList *splits_to_clear = gnc_account_get_autoclear_splits(account, amount_to_clear, &err);
+        GList *splits_to_clear = gnc_account_get_autoclear_splits(account_, amount_to_clear, &err);
 
         // Actually clear splits
         for (GList *node = splits_to_clear; node; node = node->next) {
@@ -142,20 +150,18 @@ TEST_P(AutoClearTest, DoesAutoClear) {
 
         ASSERT_STREQ(err, t.expectedErr);
         if (t.expectedErr == NULL) {
-            gnc_numeric c = xaccAccountGetClearedBalance(account);
+            gnc_numeric c = xaccAccountGetClearedBalance(account_);
             ASSERT_EQ(c.num, t.amount);
             ASSERT_EQ(c.denom, DENOM);
         }
     }
-
-    qof_book_destroy(book);
 }
 
 INSTANTIATE_TEST_SUITE_P(
     InstantiationAutoClearTest,
     AutoClearTest,
     testing::Values(
-        std::pair{ easySplitData, easyTestCases },
-        std::pair{ ambiguousSplitData, ambiguousTestCases }
+        easyTestCase,
+        ambiguousTestCase
     )
 );
--- a/libgnucash/app-utils/test/test-autoclear.cpp
+++ b/libgnucash/app-utils/test/test-autoclear.cpp
@@ -27,9 +27,8 @@
 extern "C" {
 #include "../gnc-ui-balances.h"
 }
-#include "Split.h"
-
-#include "gtest/gtest.h"
+#include <Split.h>
+#include <gtest/gtest.h>
 
 static const int64_t DENOM = 100; //< Denomerator is always 100 for simplicity.
 
--- a/libgnucash/app-utils/test/test-autoclear.cpp
+++ b/libgnucash/app-utils/test/test-autoclear.cpp
@@ -31,11 +31,11 @@ extern "C" {
 
 #include "gtest/gtest.h"
 
-const int DENOM = 100; //< Denomerator is always 100 for simplicity.
+const gint64 DENOM = 100; //< Denomerator is always 100 for simplicity.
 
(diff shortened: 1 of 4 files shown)


christopherlam:
Very nice... For now it would be sensible to merge the passing tests first, then consider how to fix the knapsack algorithm to consider non-unique amounts. IMHO having UI to pick the splits is overkill, but if it's an all-or-nothing scenario for duplicate splits then why not :)

@fellen is best placed to opine on wording of translatable strings or having html markup in ui.

libgnucash/app-utils/test/test-autoclear.cpp
```diff
@@ -0,0 +1,124 @@
...
+
+    for (auto &d : splitData) {
+        Split *split = xaccMallocSplit(book);
+        xaccSplitSetMemo(split, d.memo);
+        xaccSplitSetAmount(split, gnc_numeric_create(d.amount, DENOM));
+        xaccSplitSetReconcile(split, d.cleared ? CREC : NREC);
+
+        // This way of inserting a split, seems to actualy work best. :D
+        gnc_account_insert_split(account, split);
+    }
```
christopherlam, line 100: You'd want to wrap the whole account insertion with `xaccAccountBeginEdit` & `xaccAccountCommitEdit` to raise the edit-level during insertion. Should make it faster.

libgnucash/app-utils/test/test-autoclear.cpp
```diff
@@ -0,0 +1,124 @@
...
+            xaccSplitSetReconcile(split, CREC);
+        }
+
+        ASSERT_STREQ(err, t.expectedErr);
+        if (t.expectedErr == NULL) {
+            gnc_numeric c = xaccAccountGetClearedBalance(account);
+            ASSERT_EQ(c.num, t.amount);
+            ASSERT_EQ(c.denom, DENOM);
+        }
+    }
```
christopherlam, line 121: I'd think for each test case, there should be `qof_session_new ()`, insert_splits, test, `gnc_clear_current_session ()` to reset the splits to their desired initial reconcile status?

cristiklein: I'm a bit unsure here. I see the value in testing "incremental auto-clear" too. What do you think?

christopherlam: Ok agreed. Didn't realise this was intentional.

christopherlam:
> but if it's an all-or-nothing scenario for duplicate splits then why not :)

In case it's not obvious; if uncleared splits have amounts $100 $10 $10 $10, then:
- clearing $100 is easy - will succeed
- clearing $10 or $20 or $110 or $120 will fail - ambiguous splits
- clearing $30 or $130 currently fails but could succeed - all duplicate splits are cleared

Not sure how the algorithm could be amended to consider above.

christopherlam:
Once @fellen offers views on markup in strings we can fixup and merge. Would also be nice to insert the (commented-out) failing test which handles the all-duplicated-splits scenario as above for a future algorithm amendment.

cristiklein:
@christopherlam I added a test case for ambiguous cases. Please check. I'll run a background thread in my brain on how to deal with such cases.

christopherlam:
:+1:

jralls:
> Unfortunately, right now the tests fail. The reason is that the test cases include two splits of the same amount. This leads to "Cannot uniquely clear splits. Found multiple possibilities.". This is a limitation of the current algorithm. 

Change the test so that it passes, this is the expected behavior for an ambiguous set of splits.

> Not sure how the algorithm could be amended to consider above.

There's no general solution because there isn't enough information unless some of the ambiguous splits are on transactions dated after the bank balance in which case they can be excluded from consideration and there should be a test of that to make sure that it works (an opportunity to try out test-driven development!). Otherwise if possible the user could download and import the cleared transactions from the bank and the import matcher will mark them cleared... but if that's possible it's IMO what the user should be doing instead of using this dialog.

gnucash/gtkbuilder/window-autoclear.glade
```diff
@@ -72,7 +72,9 @@
                 <property name="visible">True</property>
                 <property name="can_focus">False</property>
                 <property name="label" translatable="yes">&lt;b&gt;Auto-Clear Information&lt;/b&gt;
-Use this dialog if you want GnuCash to automatically find which transactions are cleared, given an ending balance. For example, said ending balance can be the current balance given by your online bank.</property>
+Use this dialog if you want GnuCash to automatically find which transactions are cleared, given an ending balance. For example, said ending balance can be the current balance given by your online bank.
+
+&lt;b&gt;Note&lt;/b&gt;: The complexity of this algorithm is quite high (&lt;a href="https://en.wikipedia.org/wiki/Knapsack_problem#Computational_complexity"&gt;O(nW)&lt;/a&gt;). Use this only on a dozen of uncleared splits.</property>
```
jralls, line 77: Don't cite the Wikipedia article on the knapsack problem, that's not what you implemented.

jralls (changes requested):
Overall pretty good. The big issue is that setting up the account and splits should be done in the fixture, not in the test. Using the ctor and dtor for that is better RAII practice, but SetUp and TearDown are OK too.

Not sure what is Ubuntu 18.04's problem with `INSTANTIATE_TEST_CASE_P`, I'll have to fire up that VM and try it.

christopherlam:
Here's my thought on brute forcing a solution: consider 4 splits A=100, B=10, C=10, D=10. This brute forces the solution.

* Process A
 (empty)
to:
 100=[A]

* Process B
 100=[A]
to:
 100+10=take existing [A] add B=>[B,A]
 10=[B]

* Process C
 100=[A]
 10=[B]
 110=[B,A]
to:
 100+10=110=existing=>null
 10+10=20=existing=>null
 110+10=120=new=take existing [B,A] add C => [C,B,A]
 10=existing=>null

* Process D
 100=[A]
 10=null
 20=null
 110=null
 120=[C,B,A]
to:
 100+10=110=existing null
 10+10=20=existing null
 20+10=30=>[C]
 110+10=120=existing => null
 120+10=130=new=take existing [C,B,A] add D => [D,C,B,A]
 10=existing=>null

Easiest to understand is adding the 4th split D=10. Existing hash-table has keys for 100,10,20,110,120. To each key, add D.amount resulting in 110,20,30,120,130. Create a temporary hash-table with keys 110,20,30,120,130 from existing hash-table's values, adding D. Don't forget to process the solo split D. If existing-hash's value is null, ignore. After creating the temporary hash-table, merge onto main hash-table, overwriting existing keys. Dump temporary hash-table. Thus the hash-table now has `reachable_amount => toclear_list`.

jralls:
You're missing the point completely. If you have 3 splits with the same amount and a balance that indicates 2 of them are cleared you have a 1 in 3 chance of guessing correctly which are the 2, meaning that you're more likely to be wrong than right. The current behavior, to tell the user "sorry, couldn't figure out a solution, you'll have to go check the bank to find out what really cleared", is correct.

christopherlam:
Done. See scheme brute-force algorithm. The result is a hash-table where keys are reachable_amounts, values are toclear_list.

```scheme
(use-modules (ice-9 match))

(define items '(100 10 10 10))

(define (make-reachable-hash items)
  (define hash (make-hash-table))
  (for-each
   (lambda (item)
     (let ((temp-hash (make-hash-table)))
       (define (process key val old-val)
         (match (hash-ref hash key)
           (#f (hash-set! temp-hash key (cons val old-val)))
           ('dup #f)
           (_ (hash-set! temp-hash key 'dup))))
       (process item item '())
       (hash-for-each (lambda (k v) (process (+ k item) item v)) hash)
       (hash-for-each (lambda (k v) (hash-set! hash k v)) temp-hash)))
   items)
  hash)

(define hash (make-reachable-hash items))

(for-each
 (lambda (elt)
   (format #t "reachable=~a, to_clear=~a\n" (car elt) (cdr elt)))
 (hash-map->list cons hash))
```

Results in:
```
reachable=130, to_clear=(10 10 10 100)
reachable=20, to_clear=dup
reachable=100, to_clear=(100)
reachable=110, to_clear=dup
reachable=10, to_clear=dup
reachable=30, to_clear=(10 10 10)
reachable=120, to_clear=dup
```

christopherlam:
algorithm still buggy... fails with items '(100 10 10 10 20 20 20)...

christopherlam:
Old code removed, this one seems to be fine:

```scheme
(use-modules (ice-9 match))

(define items '(100 10 10 10))

(define (make-reachable-hash items)
  (define hash (make-hash-table))
  (for-each
   (lambda (item)
     (let ((temp-hash (make-hash-table)))
       (define (process key val old-val)
         (hash-set! temp-hash key
                    (if (or (hash-ref hash key) (eq? old-val 'dup))
                        'dup
                        (cons val old-val))))
       (process item item '())
       (hash-for-each (lambda (k v) (process (+ k item) item v)) hash)
       (hash-for-each (lambda (k v) (hash-set! hash k v)) temp-hash)))
   items)
  hash)

(define hash (make-reachable-hash items))

(for-each
 (lambda (elt)
   (format #t "reachable=~a, to_clear=~a\n" (car elt) (cdr elt)))
 (hash-map->list cons hash))
```
Results
```
reachable=130, to_clear=(10 10 10 100)
reachable=20, to_clear=dup
reachable=100, to_clear=(100)
reachable=110, to_clear=dup
reachable=10, to_clear=dup
reachable=30, to_clear=(10 10 10)
reachable=120, to_clear=dup
```

jralls:
I've replicated the compile failure on my Ubuntu 18.04 VM. It's because `INSTANTIATE_TEST_SUITE_P` isn't in googletest until 1.10. You'll have to replace it with a separate test for each case.

libgnucash/app-utils/test/test-autoclear.cpp
```diff
@@ -31,11 +31,11 @@ extern "C" {
 
 #include "gtest/gtest.h"
 
-const int DENOM = 100; //< Denomerator is always 100 for simplicity.
+const gint64 DENOM = 100; //< Denomerator is always 100 for simplicity.
```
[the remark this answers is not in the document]
cristiklein: N00b mistake. Thanks!

libgnucash/app-utils/test/test-autoclear.cpp
```diff
@@ -39,7 +39,12 @@ struct SplitDatum {
     bool cleared;
 };
 
-SplitDatum splitData[] = {
+struct TestCase {
+    gint64 amount;
+    const char *expectedErr;
+};
+
+std::vector<SplitDatum> easySplitData = {
```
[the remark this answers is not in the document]
cristiklein: I'm not sure to understand. If I pass `std::array`s around as test parameters, wouldn't that require the sizes of the split data and tests to be the same?

cristiklein:
@jralls Thanks for the review. I fixed all issues you raised, including compatibility with Ubuntu 18.04 / GoogleTest 1.8.

I didn't understand how to replace `std::vector` with `std::array`, since the number of splits and tests is different in each test case. Am I missing something?

christopherlam:
The following is my current attempt at converting https://github.com/Gnucash/gnucash/pull/805#issuecomment-715901309 to C:

```C
typedef struct _WorkData
{
    gnc_numeric amount;
    GList *toclear_list;
} *WorkData;

static void process_item (GHashTable *hash, GList **worklist,
                          gnc_numeric amount, Split *split, GList *old_val)
{
    struct _WorkData work[1];

    work->amount = amount;

    if (g_hash_table_lookup_extended (hash, &amount, NULL, NULL) ||
        // we can't use NULL as a marker for duplicate amounts.
        old_val == (gpointer)-1)
        work->toclear_list = (gpointer)-1;
    else
        work->toclear_list = g_list_prepend (old_val, split);

    *worklist = g_list_prepend (*worklist, work);
}

static void hash_foreach_func(gpointer key, gpointer value, gpointer user_data)
{
    WorkData work = (WorkData) user_data;
    gnc_numeric thisvalue = *(gnc_numeric *) key;
    gnc_numeric reachable_value = gnc_numeric_add_fixed (thisvalue, work->amount);

    work->toclear_list = g_list_prepend
        (work->toclear_list, g_memdup (&reachable_value, sizeof (gnc_numeric)));
}

GHashTable * get_reachable_amounts (GList *splits)
{
    GHashTable *hash = g_hash_table_new_full
        (ght_gnc_numeric_hash, ght_gnc_numeric_equal, g_free, NULL);

    for (GList *n = splits; n; n = n->next)
    {
        Split *split = n->data;
        GList *worklist = NULL;
        gnc_numeric amount = xaccSplitGetAmount (split);
        struct _WorkData hash_data[1];
        process_item (hash, &worklist, amount, split, NULL);
        g_hash_table_foreach (hash, hash_foreach_func, hash_data);

        for (GList *m = worklist; m; m = m->next)
        {
            WorkData *work = (WorkData*)m->data;
            g_hash_table_insert (hash, work.amount, work.toclear_list);
        }
    }
    return hash;
}
```

christopherlam:
@cristiklein You should subscribe to devel. There's more interest in refining this. https://lists.gnucash.org/pipermail/gnucash-devel/2020-October/045556.html

gnucash/gtkbuilder/window-autoclear.glade
```diff
@@ -72,7 +72,9 @@
                 <property name="visible">True</property>
                 <property name="can_focus">False</property>
                 <property name="label" translatable="yes">&lt;b&gt;Auto-Clear Information&lt;/b&gt;
-Use this dialog if you want GnuCash to automatically find which transactions are cleared, given an ending balance. For example, said ending balance can be the current balance given by your online bank.</property>
+Use this dialog if you want GnuCash to automatically find which transactions are cleared, given an ending balance. For example, said ending balance can be the current balance given by your online bank.
+
+&lt;b&gt;Note&lt;/b&gt;: The complexity of this algorithm is quite high. Use this only on a dozen of uncleared splits.</property>
```
fellen, line 77: I had already commented on the old PR, but am happy to write an updated version, as I found the dialog now in the menu.
This would lead to a po entry
```
#: gnucash/gtkbuilder/window-autoclear.glade:74
msgid ""
"<b>Auto-Clear Information</b>\n"
"Use this dialog if you want GnuCash to automatically find which transactions "
"are cleared, given an ending balance. For example, said ending balance can "
"be the current balance given by your online bank.\n"
"\n"
"<b>Note</b>: The complexity of this algorithm is quite high. Use this only "
"on a dozen of uncleared splits."
msgstr ""

```
Invalidating the old entry
```
msgid "<b>Auto-Clear Information</b>\n"
msgstr ""

```
which was already translated into several languages.

I think I would split it in 4 labels:
```
            <child>
              <object class="GtkLabel" id="label1">
                <property name="label" translatable="yes">&lt;b&gt;About Auto-Clear&lt;/b&gt;</property>
                <property name="use_markup">True</property>
              </object>
              <packing>
                <property name="expand">False</property>
                <property name="fill">False</property>
                <property name="position">0</property>
              </packing>
            </child>
            <child>
              <object class="GtkLabel" id="label2">
                <property name="label" translatable="yes">Use this dialog if you want GnuCash to automatically find which transactions are cleared, given an ending balance. For example, said ending balance can be the current balance given by your bank online.</property>
                <property name="wrap">True</property>
                <property name="max_width_chars">80</property>
              </object>
              <packing>
                <property name="expand">False</property>
                <property name="fill">False</property>
                <property name="position">1</property>
              </packing>
            </child>
            <child>
              <object class="GtkLabel" id="label3">
                <property name="label" translatable="yes">&lt;b&gt;Caution!&lt;/b&gt;</property>
                <property name="use_markup">True</property>
              </object>
              <packing>
                <property name="expand">False</property>
                <property name="fill">False</property>
                <property name="position">2</property>
              </packing>
            </child>
            <child>
              <object class="GtkLabel" id="label4">
                <property name="label" translatable="yes">The complexity of this algorithm is quite high. Use this only on a dozen of uncleared splits.</property>
                <property name="wrap">True</property>
                <property name="max_width_chars">80</property>
              </object>
              <packing>
                <property name="expand">False</property>
                <property name="fill">False</property>
                <property name="position">3</property>
              </packing>
            </child>
            <child>
              <object class="GtkGrid" id="table1">
                <property name="border_width">6</property>
                <property name="row_spacing">6</property>
                <property name="column_spacing">6</property>
                <child>
                  <object class="GtkLabel" id="end_label">
                    <property name="label" translatable="yes">_Ending Balance</property>
                    <property name="use_underline">True</property>
                  </object>
                  <packing>
                    <property name="left_attach">0</property>
                    <property name="top_attach">0</property>
                  </packing>
                </child>
                <child>
                  <object class="GtkBox" id="end_value_box">
                    <property name="hexpand">True</property>
                    <child>
                      <placeholder/>
                    </child>
                  </object>
                  <packing>
                    <property name="left_attach">1</property>
                    <property name="top_attach">0</property>
                  </packing>
                </child>
                <child>
                  <object class="GtkLabel" id="status_label">
                  </object>
                  <packing>
                    <property name="left_attach">0</property>
                    <property name="top_attach">1</property>
                    <property name="width">2</property>
                  </packing>
                </child>
              </object>
              <packing>
                <property name="expand">False</property>
                <property name="fill">True</property>
                <property name="position">4</property>
              </packing>
            </child>
          </object>
          <packing>
            <property name="expand">False</property>
            <property name="fill">False</property>
            <property name="padding">5</property>
            <property name="position">5</property>
          </packing>
        </child>

```
This way we get 2 sections with their own headers.
Textual changes :
1. "About …" seems more appropriate to me than "… Information. (Yes, now invalidate the existing translations, too)
1. "Note" seems too weak, if the user can lock the program for a long time

@Bob-IT any tricks to improve it like avoiding the markup?

Translators would then get:
```
#: gnucash/gtkbuilder/window-autoclear.glade:74
msgid "<b>About Auto-Clear</b>"
msgstr ""

#: gnucash/gtkbuilder/window-autoclear.glade:87
msgid ""
"Use this dialog if you want GnuCash to automatically find which transactions "
"are cleared, given an ending balance. For example, said ending balance can "
"be the current balance given by your bank online."
msgstr ""

#: gnucash/gtkbuilder/window-autoclear.glade:101
msgid "<b>Caution!</b>"
msgstr ""

#: gnucash/gtkbuilder/window-autoclear.glade:114
msgid ""
"The complexity of this algorithm is quite high. Use this only on a dozen of "
"uncleared splits."
msgstr ""

```
Because the chunks are smaller, later changes of the text are easier fixed by the translators.

Bob-IT: @fellen Yes there is, remove the bold markup and under the 'glade application label common tab' add the style class...
 gnc-class-emphasize-label

This is defined in gnucash-fallback.css as...
.gnc-class-emphasize-label {
  font-size: large;
  font-weight: bold;
}

You could even add gnc-class-highlight but that may be too much.

fellen: image
with
```
                <property name="label" translatable="yes">About Auto-Clear</property>
                <style>
                  <class name="gnc-class-highlight"/>
                </style>
```
and
```
                <property name="label" translatable="yes">Caution!</property>
                <style>
                  <class name="gnc-class-emphasize-label"/>
                </style>
```

fellen: > @fellen Yes there is

When the wiki is back, we should add it to [GUI Guidelines](https://wiki.gnucash.org/wiki/GUI_Guidelines) or an additional Glade page.

fellen: I have split off the GUI changes  as PR #811 .
@cristiklein you can remove commit c8bbf83 here.

cristiklein: Done, thanks!

christopherlam: FWIW the maint branch already has a safeguard (see `#define MAXIMUM_SACK_SIZE 1000000`) whereby the algorithm is aborted when the number of reachable_amounts reaches 1million. This can happen with 20 splits if the amounts are powers of 2, e.g. $1.00 $2.00 $4.00 $8.00 etc, but can easily handle many, many more splits if the amounts can create duplicates e.g. $100.00 $2.00 $2.00 $2.00 $2.00 $2.00 etc. So, IMHO the caution does not need to be too specific; it could state "Caution! The complexity of this algorithm is quite high; it may abort if the search size becomes excessive."

christopherlam:
I do think this is now complete. @jralls would confirm. We'd fixup all and merge in.

jralls (changes requested):
You're right about `std::array`, I'd forgotten that it's typed with the number of elements. So you could avoid the allocation by using a local C array with a sentinel or a size, but then you'd lose either the test_case architecture or range-based for. I think having to allocate the vector is a worthwhile expense for that gain.

Nice workaround for making the test case architecture work in Googletest 1.8.

PR #807, opened 2020-10-27
Maint 797989

christopherlam:
Bug 797989 - Sorting columns by alphabet with non-ASCII characters

Trial fix - for Transaction Report only. @fellen to verify sorting is localized...


fellen:
> @fellen to verify sorting is localized...

collate_report
should be: Ero…, Erö…, Eru…

libgnucash/core-utils/gnc-locale-utils.c
```diff
@@ -164,10 +164,8 @@ gchar *gnc_locale_name (void)
...
 {
-    char* akey = a ? g_utf8_collate_key (a, -1) : NULL;
-    char* bkey = b ? g_utf8_collate_key (b, -1) : NULL;
-    int result = g_strcmp0 (akey, bkey);
-    if (akey) g_free (akey);
-    if (bkey) g_free (bkey);
-    return result;
+    if (!a && !b) return 0;
+    if (!a) return -1;
+    if (!b) return 1;
```
jralls, line 169: This is a good place for the ternary operator:
```C
if (!(a && b))
   return a ? 1 : b ? -1 : 0;
```
or even
```C
return !a ? (!b ? 0 : -1) : !b ? 1 : g_utf8_collate (a, b);
```

gnucash/report/trep-engine.scm
```diff
@@ -2069,6 +2069,12 @@ warning will be removed in GnuCash 5.0"))
        (else
         (string-contains str transaction-matcher))))
 
+    (define (locale-string<? a b)
+      (< (gnc-collation-compare a b) 0))
+
+    (define (locale-string>? a b)
+      (> (gnc-collation-compare a b) 0))
+
```
jralls, line 2077: Do this in core-utils.scm so it's available to all scheme code instead of just trep-engine.scm users.

jralls (changes requested):

christopherlam:
So fell you mean the sorting isn't fixed properly by this branch? And the sorting was definitely incorrect prior to this change too?

libgnucash/core-utils/gnc-locale-utils.c
christopherlam: or, avoid negation:
```c
    return a ? (b ? g_utf8_collate (a, b) : 1) : b ? -1 : 0;
```

christopherlam:
Rewritten using existing `safe_utf8_collate`. @fellen the sorting isn't fixed yet because the trep sorting doesn't call the localized string comparison functions in the custom sorter; to use the custom sorter choose the following settings:

Primary Key: account
Secondary Key: date
Secondary Subtotal: monthly

Note this currently does *not* work because `safe_utf8_collate` is not importable into guile. Tried `(use-modules (sw_core_utils))` which doesn't seem to work at all.

fellen:
I had still no success, but I also do not know custom sorter.

If using C functions is too hard, how about using [scheme functions](https://www.gnu.org/software/guile/manual/html_node/Text-Collation.html)?

christopherlam:
> I had still no success, but I also do not know custom sorter.
> 
> If using C functions is too hard, how about using [scheme functions](https://www.gnu.org/software/guile/manual/html_node/Text-Collation.html)?

Ooh. Then it's easy; use sorting options as above (prime-sortkey account, sec-sortkey=date, sec-subtotal=month) with

```patch
@@ -2089,7 +2089,7 @@ warning will be removed in GnuCash 5.0"))
              (value-of-X (comparator-function split-X))
              (value-of-Y (comparator-function split-Y))
              (op (if (string? value-of-X)
-                     (if ascend? string<? string>?)
+                     (if ascend? string-locale<? string-locale>?)
                      (if ascend? < >))))
         (and value-of-X (op value-of-X value-of-Y))))
```
which results in
image

This can work well, but relies on guile correctly identifying the current locale in all OS.

fellen:
Somehow I succeeded to crash the report:
Fehler im Bericht

Beim Erstellen des Berichts ist ein Fehler aufgetreten.

          19 (gncp-option-invoke-callback #<swig-pointer GNCOptionChangeCallback 7f772451502f> #)
In c-interface.scm:
     24:4 18 (gnc:call-with-error-handling _ _)
In ice-9/boot-9.scm:
  1731:15 17 (with-exception-handler #<procedure 7f77159cfc90 at ice-9/boot-9.scm:1815:7 (exn)> _ …)
  1736:10 16 (with-exception-handler _ _ #:unwind? _ #:unwind-for-type _)
In c-interface.scm:
    29:37 15 (_)
In unknown file:
          14 (eval-string "(gnc:report-run 3)" #<undefined>)
In ice-9/boot-9.scm:
   2806:4 13 (save-module-excursion #<procedure 7f77159e9840 at ice-9/eval-string.scm:66:9 ()>)
In ice-9/eval-string.scm:
     38:6 12 (read-and-eval #<input: string 7f77159c6460> #:lang _)
In report-core.scm:
    693:4 11 (gnc:report-run _)
In c-interface.scm:
     63:2 10 (gnc:backtrace-if-exception _ . _)
     24:4  9 (gnc:call-with-error-handling _ _)
In ice-9/boot-9.scm:
  1731:15  8 (with-exception-handler #<procedure 7f77159cfb70 at ice-9/boot-9.scm:1815:7 (exn)> _ …)
  1736:10  7 (with-exception-handler _ _ #:unwind? _ #:unwind-for-type _)
In c-interface.scm:
    28:40  6 (_)
In report-core.scm:
   695:29  5 (_)
   677:25  4 (gnc:report-render-html #<<report> type: "2fe3b9833af044abb929a88d5a59620f" id: 3 …> …)
In trep-engine.scm:
  2205:21  3 (gnc:trep-renderer #<<report> type: "2fe3b9833af044abb929a88d5a59620f" id: 3 optio…> …)
In unknown file:
           2 (stable-sort! (#<swig-pointer Split * d138c0>) #<procedure primary-comparator? (X Y)>)
In core-utils.scm:
     62:5  1 (gnc:locale-string<? "Aufwendungen:Lebensmittel" "Aktiva:Barvermögen:Bargeld")
In ice-9/boot-9.scm:
  1669:16  0 (raise-exception _ #:continuable? _)

Unbound variable: safe-utf8-collate

christopherlam:
Yup. That's why it can't work yet. I'll introduce `gnc:string-locale<?` and this will work. In a couple days.

fellen:
The build failed with guile3 because there are still a bunch of "(_ " calls.

christopherlam:
@fellen you can use this branch to test transaction report. You can even set the following to *always* use the custom sorter (which uses locale-sensitive sort) in transaction report; if we want to fix sorting elsewhere (in register or account-tree) will need some deep string-comparison sort fix somewhere (in qofquery.cpp maybe?).

```patch
@@ -2044,12 +2044,7 @@ warning will be removed in GnuCash 5.0"))
                          (opt-val pagename-filter optname-closing-transactions)
                          'closing-match))
          (splits '())
-         (custom-sort? (or (and (memq primary-key DATE-SORTING-TYPES)
-                                (not (eq? primary-date-subtotal 'none)))
-                           (and (memq secondary-key DATE-SORTING-TYPES)
-                                (not (eq? secondary-date-subtotal 'none)))
-                           (or (CUSTOM-SORTING? primary-key BOOK-SPLIT-ACTION)
-                               (CUSTOM-SORTING? secondary-key BOOK-SPLIT-ACTION))))
+         (custom-sort? #t)
          (subtotal-table? (and (opt-val gnc:pagename-display optname-grid)
                                (if (memq primary-key DATE-SORTING-TYPES)
                                    (keylist-get-info date-subtotal-list
```

fellen:
Now I got a clean build and the right sort order.
At some point I might have tested on the wrong branch.

fellen:
> Now I got a clean build and the right sort order.
> At some point I might have tested on the wrong branch.

But not always. On my first test fike I still Umlauts are still sorted at the end.
I have currently no idea of the cause.

christopherlam:
I like screenshots ☺️

fellen:
I have no problems with new files and their reports.

The only problem is still that from https://github.com/Gnucash/gnucash/pull/807#issuecomment-717301919.
I tried already

-  close the report and open a new instance,
-  remove its .gcm; 
- edit the "ö" in the transaction, …
If nobody has an idea, where the old order could be hidden, I think we should apply the PR and close the bug with a corresponding note.

PR #808, opened 2020-10-27
Reports page changes

Bob-IT:
This has stemmed from Bug 797982 which causes Gnucash to sometimes crash when changing books when using 'File->Open' from the first book and there are open report pages. @jralls traced this down to the use of the g_idle_add to load the reports.

To fix this I have done away with the g_idle_add and the reports now load when the page has focus. To achieve this a flag is set at the start of the restoring window pages and reset at the end which is tested on report page focus. This has the benefit of reducing the load time if the last open page is not a report page BUT when you switch to a report page you will need to wait for the report to be rendered.

While looking at this I noticed that the page focus was focusing on the scroll window and not the Webkit webview so changed that which led to the commit for intercepting the Ctrl+Alt+Page_Up/Down on the report pages which is used for notebook navigation.

The last commit adds two key combinations Ctrl+Alt+Menu to raise the notebook page select menu and I think a translatable key combo of Ctrl+Alt+a to jump to the Accounts page for those heavy keyboard users. @christopherlam 

I have tested this on my Linux VM and Windows 10 and seems to be OK but please give it a try.

### DIFF
--- a/gnucash/gnome-utils/gnc-main-window.c
+++ b/gnucash/gnome-utils/gnc-main-window.c
@@ -67,6 +67,7 @@
 #include "gnc-ui-util.h"
 #include "gnc-uri-utils.h"
 #include "gnc-version.h"
+#include "gnc-warnings.h"
 #include "gnc-window.h"
 #include "gnc-prefs.h"
 #include "option-util.h"
@@ -1422,6 +1423,32 @@ gnc_main_window_delete_event (GtkWidget *window,
 
     if (already_dead)
         return TRUE;
+
+    if (g_list_length (active_windows) > 1)
+    {
+        gint response;
+        GtkWidget *dialog;
+        gchar *message = _("This window is closing and will not be restored.");
+
+        dialog = gtk_message_dialog_new (GTK_WINDOW (window),
+                                         GTK_DIALOG_DESTROY_WITH_PARENT,
+                                         GTK_MESSAGE_QUESTION,
+                                         GTK_BUTTONS_NONE,
+                                         "%s", _("Close Window?"));
+        gtk_message_dialog_format_secondary_text (GTK_MESSAGE_DIALOG(dialog),
+                                                  "%s", message);
+
+        gtk_dialog_add_buttons (GTK_DIALOG(dialog),
+                              _("_Cancel"), GTK_RESPONSE_CANCEL,
+                              _("_OK"), GTK_RESPONSE_YES,
+                               (gchar *)NULL);
+        gtk_dialog_set_default_response (GTK_DIALOG(dialog), GTK_RESPONSE_YES);
+        response = gnc_dialog_run (GTK_DIALOG(dialog), GNC_PREF_WARN_CLOSING_WINDOW_QUESTION);
+        gtk_widget_destroy (dialog);
+
+        if (response == GTK_RESPONSE_CANCEL)
+            return TRUE;
+    }
 
     if (!gnc_main_window_finish_pending(GNC_MAIN_WINDOW(window)))
     {
--- a/gnucash/gnome-utils/gnc-main-window.c
+++ b/gnucash/gnome-utils/gnc-main-window.c
@@ -2743,6 +2743,58 @@ gnc_main_window_destroy (GtkWidget *widget)
 }
 
 
+static gboolean
+gnc_main_window_key_press_event (GtkWidget *widget, GdkEventKey *event, gpointer user_data)
+{
+    GncMainWindowPrivate *priv;
+    GdkModifierType modifiers;
+
+    g_return_val_if_fail (GNC_IS_MAIN_WINDOW(widget), FALSE);
+
+    priv = GNC_MAIN_WINDOW_GET_PRIVATE(widget);
+
+    modifiers = gtk_accelerator_get_default_mod_mask ();
+
+    if ((event->state & modifiers) == (GDK_CONTROL_MASK | GDK_MOD1_MASK)) // Ctrl+Alt+
+    {
+        const gchar *account_key = C_ ("lower case key for short cut to 'Accounts'", "a");
+        guint account_keyval = gdk_keyval_from_name (account_key);
+
+        if ((account_keyval == event->keyval) || (account_keyval == gdk_keyval_to_lower (event->keyval)))
+        {
+            gint page = 0;
+
+            for (GList *item = priv->installed_pages; item; item = g_list_next (item))
+            {
+                 const gchar *pname = gnc_plugin_page_get_plugin_name (GNC_PLUGIN_PAGE(item->data));
+
+                 if (g_strcmp0 (pname, "GncPluginPageAccountTree") == 0)
+                 {
+                     gtk_notebook_set_current_page (GTK_NOTEBOOK(priv->notebook), page);
+                     return TRUE;
+                 }
+                 page++;
+            }
+        }
+        else if ((GDK_KEY_Menu == event->keyval) || (GDK_KEY_space == event->keyval))
+        {
+            GList *menu = gtk_menu_get_for_attach_widget (GTK_WIDGET(priv->notebook));
+
+            if (menu)
+            {
+                gtk_menu_popup_at_widget (GTK_MENU(menu->data),
+                                          GTK_WIDGET(priv->notebook),
+                                          GDK_GRAVITY_SOUTH,
+                                          GDK_GRAVITY_SOUTH,
+                                          NULL);
+                return TRUE;
+            }
+        }
+    }
+    return FALSE;
+}
+
+
 /*  Create a new gnc main window plugin.
  */
 GncMainWindow *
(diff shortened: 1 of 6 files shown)


Bob-IT:
As a follow up, I have noticed several bug like...
https://bugs.gnucash.org/show_bug.cgi?id=343228
https://bugs.gnucash.org/show_bug.cgi?id=411810
https://bugs.gnucash.org/show_bug.cgi?id=450717

That ask for the notebook navigation to be Ctrl+Page_Up/Down which I think conflicts with the tree view use which pages without selecting on the Accounts page but I think that would not be missed and jumping to the blank transaction. There may be other page conflicts.

Is this something that we want fixing or can we just say Ctrl+Alt+Page_UP/Down is the way and these bugs can be closed?

jralls:
@Bob-IT I think we could go ahead and override ctrl-pgup/pgdn on the accounts page. It works the same as plain pgup/pgdn on macOS. I'm a bit less sanguine about ctrl-pgdn on register pages. I imagine that there are users who have that hardwired into their fingers by now and will complain loudly if it suddenly starts meaning tab-right instead of blank transaction.

BTW it's 411910 not 411810. I've close all the others as duplicates of 343228. 

Another bug related to this also suggested alt-number for switching to a particular page but I suspect that would be painful to make work for those users with more than 10 pages. Raising the notebook select menu strikes be as a better solution but that leads to another BTW: [Macs don't have Menu keys.](https://developer.apple.com/documentation/appkit/1535851-function-key_unicodes/nsmenufunctionkey?language=objc).

christopherlam:
:+1: instead of Ctrl-Alt-A, how about Ctrl-Alt-Home?

Bob-IT:
> I think we could go ahead and override ctrl-pgup/pgdn on the accounts page.

@jralls I have added to the key-press-event callback to do this for all pages with the exception of using Ctrl+Page_Down on Register pages.

Bob-IT:
>instead of Ctrl-Alt-A, how about Ctrl-Alt-Home -- the counterpart Ctrl-Alt-End could work well to jump to the last tab.

@christopherlam Those key combinations will move the application to a different work spaces on my Linux VM. Also the Account page may not necessarily be the first one but when the menu pops up you can easily do PageUp/Down  to get to first and last one.

Bob-IT:
> BTW: [Macs don't have Menu keys.](https://developer.apple.com/documentation/appkit/1535851-function-key_unicodes/nsmenufunctionkey?language=objc).

@jralls I did not know that, maybe I should buy a second hand mac so I can build and run Gnucash, what would be the minimum spec. For the menu I could add Ctr+Alt+space as an alternative.

Bob-IT:
The last commit adds a confirmation dialog for when you have multiple windows open and you click the 'x' on the wrong one so the remaining window over writes the saved window / 'open pages' configuration which I have done several times.

christopherlam:
> > instead of Ctrl-Alt-A, how about Ctrl-Alt-Home -- the counterpart Ctrl-Alt-End could work well to jump to the last tab.
> 
> @christopherlam Those key combinations will move the application to a different work spaces on my Linux VM. Also the Account page may not necessarily be the first one but when the menu pops up you can easily do PageUp/Down to get to first and last one.

You'll be the best judge then. Design by committee always difficult. Personally I'd populate the Window menu item with the list of current window tabs. As long as this blooming Ctrl-PgUp/PgDn finally works reliably.

Bob-IT:
> You'll be the best judge then. Design by committee always difficult. Personally I'd populate the Window menu item with the list of current window tabs. As long as this blooming Ctrl-PgUp/PgDn finally works reliably.

@christopherlam The Window menu item is popped up with the list of open tabs as if you had right moused on a tab. Navigation is most reliable with Ctrl+Alt+Page_Up/Down and jumps from one tab to the next as expected. With a following commit Ctrl+Page_Up/Down will also work with one exception on the Register pages as Ctrl+Page_Down is used to jump to blank.

christopherlam:
> @christopherlam The Window menu item is popped up with the list of open tabs as if you had right moused on a tab. 

It doesn't actually... it lists open Windows. Create new window with "Window > New Window" and now you have two windows. I'd wager most users have many tabs rather than many windows, hence, the Window listing open Tabs would be nicer. Or it can list open windows *followed by* open tabs in current window. 

> Navigation is most reliable with Ctrl+Alt+Page_Up/Down and jumps from one tab to the next as expected. With a following commit Ctrl+Page_Up/Down will also work with one exception on the Register pages as Ctrl+Page_Down is used to jump to blank.

Nice, looking forward to this change.

Bob-IT:
> It doesn't actually... it lists open Windows. Create new window with "Window > New Window" and now you have two windows. I'd wager most users have many tabs rather than many windows, hence, the Window listing open Tabs would be nicer. Or it can list open windows _followed by_ open tabs in current window.

@christopherlam I miss understood what you were taking about. You don't like the builtin popup menu?

christopherlam:
> > It doesn't actually... it lists open Windows. Create new window with "Window > New Window" and now you have two windows. I'd wager most users have many tabs rather than many windows, hence, the Window listing open Tabs would be nicer. Or it can list open windows _followed by_ open tabs in current window.
> 
> @christopherlam I miss understood what you were taking about. You don't like the builtin popup menu?

I don't mind the built-in popup menu which appears when you right-click the tab. Unfortunately it requires the use of a fancy modern mouse :) My para above refers to the Window menu item which could be augmented with the tab list.

christopherlam:
In other words there's no reason this (which requires a mouse):
image
can't be tagged at the end of this (which can be accessed via Alt-W):
image

Bob-IT:
> I don't mind the built-in popup menu which appears when you right-click the tab. Unfortunately it requires the use of a fancy modern mouse :)

@christopherlam Not with the commit above, just Ctrl+Alt+Menu key and up pops the menu. Will ponder...

jralls:
> maybe I should buy a second hand mac so I can build and run Gnucash, what would be the minimum spec. 

@Bob-IT for 4.x it's anything that runs 10.13 or later, but in practice anything built in the last 5 or 6 years will run 11.0, the release now in beta.

jralls:
@christopherlam Putting tabs in the Window menu would be surprising to many users, most programs don't. 

Why are you using an Apple mouse?

gjanssens:
> > I think we could go ahead and override ctrl-pgup/pgdn on the accounts page.
> 
> @jralls I have added to the key-press-event callback to do this for all pages with the exception of using Ctrl+Page_Down on Register pages.

Personally I'm not a fan of sequences that are meant to work universally but don't on certain page types. That causes confusion  (as behaviour is not consistent) and frustration (as usually muscle memory kicks in faster than the brain can remind you it doesn't work on the current page). The frustration is aggrevated as the sequence has side-effects you didn't intend (jump to blank line).

So I would argue to either not implement it at all or implement it everywhere. The latter means we need to find a new shortcut for jumping to the blank line and require users to retrain their muscle memory. I do think however it's better to retrain muscle memory for a new, consistent behaviour than to continually have to fight muscle memory due to inconsistent behaviour.

Alternatively we could come up with a UI in which users can define their own shortcuts. We already write an accelerators file that can be tweaked manually so to some extent we already support customizing keyboard shortcuts. However edtiting a text file is not a very user friendly interface for a wider audience.

Bob-IT:
So it looks like...

- Do nothing for the Ctrl+Page_UP/Down, some pages would still work as it is part of the GtkNotebook but probably the combo would be used on the page

- Block Ctrl+Page_UP/Down on all pages apart from the Down on the Register page. At least it would be consistent and say that the only combo that works is Ctr+Alt+Page_Up/Down.

- Make Ctrl+Page_UP/Down work for all pages which means changing the go to Blank combo to maybe Ctrl+B

If reverse sorting was ever completed, the blank transaction being at the top then Ctrl+Page_Down would not make sense for the go to blank transaction.

Not sure what is the best option.

christopherlam:
Not sure how other users do, but I never understood Ctrl-PgDn to intuitively jump to blank transaction. Ctrl-Enter would seem more appropriate: hit return to complete current transaction changes; use Ctrl as modifier and also jumps to add another transaction.

Firefox does not expose key combinations in ui -- I'm not convinced the extra effort is really necessary.

gjanssens:
I agree ctrl-enter would be a useful keybinding to jump to the blank line. Considering the potential invasiveness of the change I also suggest to only do it in master and make a lot of noise around it.

jralls:
> I agree ctrl-enter would be a useful keybinding to jump to the blank line. Considering the potential invasiveness of the change I also suggest to only do it in master and make a lot of noise around it.

Concur.

jralls:
> Alternatively we could come up with a UI in which users can define their own shortcuts. We already write an accelerators file that can be tweaked manually so to some extent we already support customizing keyboard shortcuts. However edtiting a text file is not a very user friendly interface for a wider audience.

The [accelerator map](https://developer.gnome.org/gtk3/unstable/gtk3-Accelerator-Maps.html) is a Gtk feature, not our doing and it operates only on menu entries. We would implement a [GtkShortcutsWindow](https://developer.gnome.org/gtk3/unstable/GtkShortcutsWindow.html) to enable user access. I'd consider that a new feature that should be targeted at master. I appears to me that this requires migrating from GtkAccelerators to GtkShortcuts (introduced in Gtk+-3.20) and probably from the deprecated GtkActions and GtkMenu* to GAction and GMenuModel. That's heading into GUI overhaul territory and we've made almost no progress in extracting the M and C from the V that's something of a prerequisite for changing to another toolkit.

jralls:
@Bob-IT in light of the above please extract b20029423c from this PR and merge the rest.

PR #809, opened 2020-10-27
Add tooltip in import matcher window for description and memo column.

cstim:
Adding tooltips in a GtkTreeView is actually surprisingly straightforward, if one could use this word in the context of gtk.

The "description" field is often quite long for imported transactions. So users (and myself) may need to see the full content, but the screen has only finite width. Showing the full text as tooltip helps a lot.


masla84:
Nice work, you tamed gtk, and at this time (MEZ) ;-)

fellen:
Merged, thanks!

@Bob-IT Any annotations?

PR #810, opened 2020-10-27
Import matcher: add extra checkbox to make memo column invisible

cstim:
Does what the title says: In addition to the "account column" checkbox, another one for the memo column is added.

During implementation I was puzzled because of the super-obvious code duplication in the two functions `gnc_gen_trans_assist_new` and `gnc_gen_trans_list_new` - but I didn't feel called to clean up all of this duplication, only in the parts that I added. Anyone else may feel free to step up here.

### DIFF
--- a/gnucash/gtkbuilder/dialog-import.glade
+++ b/gnucash/gtkbuilder/dialog-import.glade
@@ -926,7 +926,7 @@
         <property name="halign">center</property>
         <child>
           <object class="GtkCheckButton" id="show_source_account_button">
-            <property name="label" translatable="yes">Show the _Account column</property>
+            <property name="label" translatable="yes">Show _Account column</property>
             <property name="visible">True</property>
             <property name="can_focus">True</property>
             <property name="receives_default">False</property>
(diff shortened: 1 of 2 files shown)


jralls:
@cstim Would you please split this into two commits, one for the extract-function refactor and the other for adding the memo column? The description of the latter could be improved a little as you're not just adding a checkbox, you're adding the column and providing a checkbox to control whether or not it's displayed.

cstim:
> @cstim Would you please split this into two commits

Is this a joke? Honestly: To request such an extra splitting of commits of the concerned approx. 15 lines sounds like quite ridiculous nit-picking to me. The commit adds a feature, and in doing so, has to add an extra button setup, and instead of continuing the existing stupid code pattern of code duplication, the relevant section (the button setup) is moved into one separate function so that the feature addition happens only once.

Given the average commit atomicity of the gnucash codebase, even when only considering recent years, this commit is already way more self-contained and not needing of any further split up compared to the majority of the other commits out there. Plus, that code section hasn't been touched by anyone except the original author (Bob-IT in this case), so obviously nobody else cares about it anyway. I guess if I hadn't mentioned it in the description, no question would have been asked anyway. I think this commit is small-sized enough, again comparing against the vast majority of gnucash commits so far.

jralls:
It isn't a joke and it would take you less time to just do it than to write that diatribe. Modern best practice is small commits that do only one thing.

Bob-IT (approved):
I can see nothing amiss, if you want I can factor out the common code as a commit as it looks like I started the code duplication and then add your changes on top.

Bob-IT:
I have refactored the two new functions so that they share some common code and also added this change to a following commit so am closing this PR.

Bob-IT:
I have refactored the two new functions so that they share some common code and also added this change to a following commit so am closing this PR.

PR #811, opened 2020-10-29
Amend autoclear dialog - spin-off from PR #805

fellen:
As the warning does not belong in the tests, I created a separate PR for the dialog improvement - mostly because code is still down.

### DIFF
--- a/gnucash/gtkbuilder/window-autoclear.glade
+++ b/gnucash/gtkbuilder/window-autoclear.glade
@@ -1,14 +1,11 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<!-- Generated with glade 3.22.2 -->
+<!-- Generated with glade 3.36.0 -->
 <interface>
   <requires lib="gtk+" version="3.22"/>
   <object class="GtkDialog" id="auto_clear_start_dialog">
     <property name="can_focus">False</property>
     <property name="resizable">False</property>
     <property name="type_hint">dialog</property>
-    <child type="titlebar">
-      <placeholder/>
-    </child>
     <child internal-child="vbox">
       <object class="GtkBox" id="dialog-vbox6">
         <property name="visible">True</property>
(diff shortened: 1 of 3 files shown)


fellen:
Oops, that is not, what I wanted.

gnucash/gtkbuilder/window-autoclear.glade
```diff
@@ -71,18 +68,59 @@
...
                 <property name="visible">True</property>
                 <property name="can_focus">False</property>
-                <property name="label" translatable="yes">&lt;b&gt;Auto-Clear Information&lt;/b&gt;
-Use this dialog if you want GnuCash to automatically find which transactions are cleared, given an ending balance. For example, said ending balance can be the current balance given by your online bank.
-
-&lt;b&gt;Note&lt;/b&gt;: The complexity of this algorithm is quite high (&lt;a href="https://en.wikipedia.org/wiki/Knapsack_problem#Computational_complexity"&gt;O(nW)&lt;/a&gt;). Use this only on a dozen of uncleared splits.</property>
-                <property name="use_markup">True</property>
+                <property name="label" translatable="yes">About Auto-Clear</property>
+                <style>
+                  <class name="gnc-class-highlight"/>
```
Bob-IT, line 73: Looking at this again, I think I would swap these, make the first label style 'gnc-class-emphasize-label' and add both to the second 'gnc-class-emphasize-label' and 'gnc-class-highlight'

I suppose one could go further and create 'gnc-primary-title', 'gnc-secondary-title' which could start of the same as 'gnc-class-emphasize-label' and also 'gnc-class-bold' just for bold text.

Bob-IT:
Forgot to mention you can also do the following in the glade application, highlight the label and on the General tab click on attributes and then edit attributes and there you will see several settings, the one for bold would be 'weight'.

gnucash/gtkbuilder/window-autoclear.glade
fellen: As I try to have a semantic view, obvisious the first is a title, but the second markup is more an admonition (note, info, tip, important, warning, caution, danger) than a subtitle. They can also be complemeted or replaced by icons.

Where is this stylesheet else used?

Usually (HTML, Docbook) "emphasis" means italic and "strong" is used for bold. Should we not lean to that standard?

gnucash/gtkbuilder/window-autoclear.glade
```diff
@@ -71,16 +68,59 @@
...
+                <property name="expand">False</property>
+                <property name="fill">False</property>
+                <property name="position">2</property>
+              </packing>
+            </child>
+            <child>
+              <object class="GtkLabel" id="label4">
+                <property name="visible">True</property>
+                <property name="can_focus">False</property>
+                <property name="label" translatable="yes">The complexity of this algorithm is quite high. Use this only on a dozen of uncleared splits.</property>
```
christopherlam, line 116: From https://github.com/Gnucash/gnucash/pull/805#discussion_r515151222

The caution does not need to be too specific; it could state "Caution! The complexity of this algorithm is quite high; it may abort if the search size becomes excessive."

fellen: > it may abort if the search size becomes excessive.

Can you include a magnitude? Worst case it will abort for …

christopherlam: As described worst case scenario is 1mil amounts reached after 20 splits.

gnucash/gtkbuilder/window-autoclear.glade
Bob-IT: > Where is this stylesheet else used?

The class 'gnc-class-emphasize-label', the name was what I came up with at the time, is used in a dialog that is shown when you change the account type. This class name can easily be changed to what ever, maybe ...
.gnc-class-underline-title {
  font-size: large;
  font-weight: bold;
  text-decoration-line: underline;
}
.gnc-class-title {
  font-size: large;
  font-weight: bold;
}
.gnc-class-strong {
  font-weight: bold;
}

gnucash/gtkbuilder/window-autoclear.glade
fellen: How about "Caution! The complexity of this algorithm is quite high; it may abort if the search size becomes excessive. Worst case the limit can be reached with 20 splits."?

christopherlam: Sure, that's fine

jralls: ISTM "The complexity of this algorithm is high" will be meaningless to the users who aren't coders; it might even be meaningless to some of the coders.

Something along the lines of "This feature can take a very long time if there are 20 or more uncleared splits." would be easier for the general user to understand.

fellen: > "This feature can take a very long time if there are 20 or more uncleared splits."

I think this is mixing 2 different things

1. It can take a very long time,

1. It can abort worst case for 20 or more uncleared splits.

fellen:
Current state:
grafik

gnucash/gtkbuilder/window-autoclear.glade
fellen: > "This feature can take a very long time if there are 20 or more uncleared splits."

Aren't you now mixing 2 different things

1. It can take a very long time,

1. It can worst case abort if there are 20 or more uncleared splits.

fellen: Strange, my answer had somehow disappeared and now it is there again.

jralls: Not really, though it perhaps mis-states that a largish number of uncleared splits somewhere around 20 will cause it to abort after taking a really long time. The actual abort occurs when the number of combinations reaches 1 million.

In any case, you're missing the point: Regardless of the actual message, the message should be understandable by users. The current wording makes sense only to programmers.

christopherlam: Actually I had a failsafe; it won't ever take too long. The most accurate statement is: "this feature is not enabled if the number of uncleared splits is more than approximately 20. If this is the case please reconcile manually."

fellen: > this feature is not enabled

Really, or "not recommend"?
Is there a better word for "feature"?

christopherlam: Try an account with 20+ uncleared splits, amounts 1, 2, 4, 8 etc up to around 1048576. It should fail to generate a split list to clear (without a long delay) unless @cristiklein or anyone can find a better algorithm.

christopherlam: How embarrassing. On testing with real data, with splits to 2^18 = $131072 onwards there is definitely slowdown. So, my #813 is not that clever. Either I'll rework it to cache `sack` instead of `toclear_list`, or revert it altogether.

fellen: > "this feature is not enabled if the number of uncleared splits is more than approximately 20. If this is the case please reconcile manually."

How about `This tool is not recommend if the number of uncleared splits is more than approximately 20. In that case please reconcile at least some of them manually.`

jralls: "clear" rather than "reconcile". "not recommend[ed]" doesn't really explain anything: "might be slow" and unless @christopherlam reverts his hack "won't work" are clearer.

PR #812, opened 2020-10-29
Fix CZK currency partname to nominative singular

McLaynV:
"Haléřů" is plural genetive, "Haléř" is singular nominative. 
https://cs.wikipedia.org/wiki/Hal%C3%A9%C5%99


fellen:
Merged, thanks!

PR #813, opened 2020-11-01
[auto-clear] provide live feedback when a valid end_amount is input

christopherlam:
@cristiklein @mtalexander 

Further refinement: if a valid `end_amount` is being written, the dialog will actually demonstrate the splits to be cleared. Every change to gnc_amount_edit calls `gnc_account_get_autoclear_splits` which must terminate in a reasonable time and I'm now confident it will terminate <1s due to the failsafe if num(reachable_amounts) exceeds 10^6. This feature renders the display of newly-cleared splits in a new register obsolete, but I'd leave it because the user may wish to review them afterwards.

image
image

### DIFF
--- a/gnucash/gnome/window-autoclear.c
+++ b/gnucash/gnome/window-autoclear.c
@@ -52,6 +52,7 @@ struct _AutoClearWindow
     GtkWidget *cancel_button;
     GtkWidget *show_cleared_splits_button;
     GtkLabel *status_label;
+    GList *toclear_list;
 };
 
 /** Callback prototypes************************************************/
@@ -122,47 +123,44 @@ void
 gnc_autoclear_window_ok_cb (GtkWidget *widget,
                             AutoClearWindow *data)
 {
-    GList *toclear_list;
     gnc_numeric toclear_value;
     gchar *errmsg = NULL;
 
     g_return_if_fail (widget && data);
 
-    toclear_value = gnc_amount_edit_get_amount(data->end_value);
+    /* sanity check: if toclear_list is null, bail out. but this
+       should not happen because the OK button is disabled if there is
+       autoclear error, and toclear_list is null */
+    g_return_if_fail (data->toclear_list);
 
-    if (gnc_reverse_balance(data->account))
-        toclear_value = gnc_numeric_neg (toclear_value);
+    xaccAccountBeginEdit (data->account);
+    for (GList *node = data->toclear_list; node; node = node->next)
+        xaccSplitSetReconcile (node->data, CREC);
+    xaccAccountCommitEdit (data->account);
 
-    toclear_value = gnc_numeric_convert
-        (toclear_value, xaccAccountGetCommoditySCU(data->account), GNC_HOW_RND_ROUND);
+    if (gtk_toggle_button_get_active
+        (GTK_TOGGLE_BUTTON (data->show_cleared_splits_button)))
+        show_cleared_splits (data->toclear_list);
 
-    toclear_list = gnc_account_get_autoclear_splits
-        (data->account, toclear_value, &errmsg);
+    /* Close window */
+    gnc_autoclear_window_cancel_cb (widget, data);
+}
 
-    if (errmsg)
-    {
-        gtk_label_set_text (data->status_label, errmsg);
-        gnc_amount_edit_set_amount (data->end_value, toclear_value);
-        gtk_editable_select_region (GTK_EDITABLE (data->end_value), 0, -1);
-        g_free (errmsg);
-    }
-    else
-    {
-        xaccAccountBeginEdit (data->account);
-        for (GList *node = toclear_list; node; node = node->next)
-            xaccSplitSetReconcile (node->data, CREC);
-        xaccAccountCommitEdit (data->account);
+static gboolean
+gnc_autoclear_window_delete_cb (GtkWidget *widget, GdkEvent  *event,
+                                AutoClearWindow *data)
+{
+    if (data->toclear_list)
+        g_list_free (data->toclear_list);
+    g_free(data);
+    return FALSE;
+}
 
-        if (gtk_toggle_button_get_active
-            (GTK_TOGGLE_BUTTON (data->show_cleared_splits_button)))
-            show_cleared_splits (toclear_list);
-
-        g_list_free (toclear_list);
-
-        /* Close window */
-        gtk_widget_destroy (data->window);
-        g_free (data);
-    }
+static void
+gnc_autoclear_end_value_activate_cb (GtkWidget *widget, AutoClearWindow *data)
+{
+    if (data->toclear_list)
+        gnc_autoclear_window_ok_cb (widget, data);
 }
 
 void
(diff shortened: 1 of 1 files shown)


christopherlam:
Note the buildup of status string with many splits is currently very leaky. Not sure how to code this correctly.

gnucash/gnome/window-autoclear.c
```diff
@@ -207,6 +240,7 @@ autoClearWindow (GtkWidget *parent, Account *account)
     // Set the name for this dialog so it can be easily manipulated with css
     gtk_widget_set_name (GTK_WIDGET(data->window), "gnc-id-auto-clear");
 
+    data->ok_button = GTK_WIDGET (gtk_builder_get_object (builder, "ok_button"));
     /* Add amount edit box */
     data->end_value = GNC_AMOUNT_EDIT(gnc_amount_edit_new());
     g_signal_connect(GTK_WIDGET(data->end_value), "activate",
```
christopherlam, line 246: @Bob-IT quick question -- why does the `end_value` "activate" signal trigger the `gnc_autoclear_window_ok_cb`? Shouldn't it connect to the `ok_button` instead? Is this a bug?

Bob-IT: @christopherlam It is there to respond to hitting the enter key after inputting amount, otherwise one would enter amount and then need to press OK

Bob-IT:
> This feature renders the display of newly-cleared splits in a new register obsolete, but I'd leave it because the user may wish to review them afterwards.

@christopherlam  Why not add a check button that when ticked will show the register view after just like the SLR, default not ticked.

Bob-IT:
> Note the buildup of status string with many splits is currently very leaky. Not sure how to code this correctly.

@christopherlam This does not look right, I think you are leaking status. In the loop I think you need a temporary gchar. On entering the loop make a copy of status with g_strdup, then g_free status then use the copy in your g_strdup_printf and then g_free the temporary one all in the loop.

gnucash/gnome/window-autoclear.c
```diff
@@ -171,7 +171,48 @@ gnc_autoclear_window_cancel_cb (GtkWidget *widget,
...
+    if (errmsg)
+    {
+        gchar *status = g_strdup_printf ("Cannot clear: %s", errmsg);
+        gtk_label_set_text (data->status_label, status);
+        g_free (status);
+        g_free (errmsg);
+    }
+    else
+    {
+        gchar *status = "The following splits will be cleared:\n";
```
Bob-IT: This needs to be translatable.

gnucash/gnome/window-autoclear.c
```diff
@@ -171,7 +171,48 @@ gnc_autoclear_window_cancel_cb (GtkWidget *widget,
...
+
+    toclear_value = gnc_numeric_convert
+        (toclear_value, xaccAccountGetCommoditySCU(data->account), GNC_HOW_RND_ROUND);
+
+    toclear_list = gnc_account_get_autoclear_splits
+        (data->account, toclear_value, &errmsg);
+
+    if (errmsg)
+    {
+        gchar *status = g_strdup_printf ("Cannot clear: %s", errmsg);
```
Bob-IT: This needs to be translatable.

christopherlam:
New autoclear algorithm to be more C-like and less lispy. Creates a amount => toclear_list, or amount => "dupe" hash table. I don't know how encode either a GList or a 'dupe' marker into the GHashTable value :

```scheme

(define (new-workitem hash reachable-amount elt old-val)
  (vector
   reachable-amount
   (if (or (hash-ref hash reachable-amount) (eq? old-val 'dup))
       'dup
       (cons elt old-val))))

(define (process-work hash workitem)
  (hash-set! hash
             (vector-ref workitem 0)
             (vector-ref workitem 1)))

(define (make-reachable-hash items)
  (define hash (make-hash-table))

  (for-each
   (lambda (item)
     (define worklist
       (list (new-workitem hash item item '())))

     (hash-for-each
      (lambda (k v)
        (set! worklist
          (cons (new-workitem hash (+ k item) item v) worklist)))
      hash)

     (for-each
      (lambda (workitem)
        (process-work hash workitem))
      worklist))
   items)

  hash)

(define hash (make-reachable-hash '(100 10 10 10)))

(hash-for-each
 (lambda (k v)
   (format #t "reachable=~a, to_clear=~a\n" k v))
 hash)
```

christopherlam:
Last commit - WIP - autoclear also accessible from Reconcile window. Idea is sound: picks up `toclear_value` from `recn->new_ending`. If error, should display in a GtkLabel. If no error, should *enable* checkbox for splits from debit & credit columns and *disable* other splits. But the latter requires traversing Debit and Credit GtkTreeView items to find the splits. Would need assistance how to navigate these TreeViews.

christopherlam:
> 
> 
> > This feature renders the display of newly-cleared splits in a new register obsolete, but I'd leave it because the user may wish to review them afterwards.
> 
> @christopherlam Why not add a check button that when ticked will show the register view after just like the SLR, default not ticked. I would also test to see what it looks like with a long description.

Agree best ellipsize description manually e.g. at 100 chars and sanitize string

christopherlam:
Ellipsizing long description. Arbitrarily maximum is 50 utf8 codepoints.

image

christopherlam:
Merged live-feedback into maint. Removed hook into Reconcile window: too difficult to scan debit&credit lists to toggle checkboxes. For reference by a future hacker; this is a starting point -- but it'd need to prescan all uncleared splits, reset their reconcile to `NREC`, set `toclear_list` splits to `CREC`, then regenerate the reconcile dialog.

```patch
modified   gnucash/gnome/window-reconcile.c
@@ -1207,6 +1207,31 @@ gnc_ui_reconcile_window_balance_cb(GtkButton *button, gpointer data)
 }
 
 
+static void
+gnc_ui_autoclear_cb(GtkButton *button, gpointer data)
+{
+    RecnWindow *recnData = data;
+    Account *account;
+    GList *splits;
+    gchar *errmsg = NULL;
+
+    account = recn_get_account (data);
+    splits = gnc_account_get_autoclear_splits (account, recnData->new_ending, &errmsg);
+
+    if (errmsg)
+    {
+        gnc_info_dialog (NULL, "%s", errmsg);
+        g_free (errmsg);
+    }
+    else
+    {
+        for (GList *n = splits; n; n = n->next)
+            xaccSplitSetReconcile (n->data, CREC);
+        g_list_free (splits);
+        recnRefresh (recnData);
+    }
+}
+
 static void
 gnc_ui_reconcile_window_rec_cb(GtkButton *button, gpointer data)
 {
@@ -2331,6 +2356,11 @@ static GtkActionEntry recnWindow_actions [] =
         "including statement date and ending balance."),
         G_CALLBACK (gnc_ui_reconcile_window_change_cb)
     },
+    {
+        "RecnAutoclearAction", "emblem-default", N_("_Auto-clear"), "<primary>a",
+        N_("Auto-clear splits"),
+        G_CALLBACK(gnc_ui_autoclear_cb)
+    },
     {
         "RecnFinishAction", "system-run", N_("_Finish"), "<primary>w",
         N_("Finish the reconciliation of this account"),
modified   gnucash/ui/gnc-reconcile-window-ui.xml
@@ -47,6 +47,7 @@
     <separator name="ToolbarSep2"/>
     <toolitem name="AccountOpenAccount" action="AccountOpenAccountAction"/>
     <separator name="ToolbarSep3"/>
+    <toolitem name="RecnAutoclear" action="RecnAutoclearAction"/>
     <toolitem name="RecnFinish" action="RecnFinishAction"/>
     <toolitem name="RecnPostpone" action="RecnPostponeAction"/>
     <toolitem name="RecnCancel" action="RecnCancelAction"/>
```

christopherlam:
Note this live status feedback introduces unacceptable delay, reverted with 6a6396edc
Will try again once algorithm is modified to be https://github.com/Gnucash/gnucash/pull/813#issuecomment-720077364 thus `reachable_hash` is cached and can be quickly looked up. This C is beyond my ability for now.

PR #814, opened 2020-11-06
[gnc-ui-util.c] reduce gnc_numeric val for display

christopherlam:
I'm 100% sure this is safe, but is it the correct fix? Example datafile at https://bugs.gnucash.org/show_bug.cgi?id=797796#c154 will spew a lot of warnings without this fix. These fractions reduce cleanly to 140/1, 170/1, 100/1 etc.

Removes warnings with large numerator & denominator e.g.

```
   GncNumeric 11900000000/85000000 could not be represented as a decimal without rounding.
   GncNumeric 11050000000/65000000 could not be represented as a decimal without rounding.
   GncNumeric 5000000000/50000000 could not be represented as a decimal without rounding.
   GncNumeric 11900000000/85000000 could not be represented as a decimal without rounding.
   GncNumeric 11050000000/65000000 could not be represented as a decimal without rounding.
   GncNumeric 5000000000/50000000 could not be represented as a decimal without rounding.
   GncNumeric 11900000000/85000000 could not be represented as a decimal without rounding.
   GncNumeric 11050000000/65000000 could not be represented as a decimal without rounding.
   GncNumeric 5000000000/50000000 could not be represented as a decimal without rounding.
```

### DIFF
--- a/libgnucash/app-utils/gnc-ui-util.c
+++ b/libgnucash/app-utils/gnc-ui-util.c
@@ -1531,7 +1531,7 @@ PrintAmountInternal(char *buf, gnc_numeric val, const GNCPrintAmountInfo *info)
 
     /* Print the absolute value, but remember sign */
     value_is_negative = gnc_numeric_negative_p (val);
-    val = gnc_numeric_abs (val);
+    val = gnc_numeric_abs (gnc_numeric_reduce (val));
 
     /* Try to print as decimal. */
     value_is_decimal = gnc_numeric_to_decimal(&val, NULL);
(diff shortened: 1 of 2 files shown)


jralls:
Nope, that's a bandaid fix. The problem is in `GncNumeric::to_decimal`. Start with adding a test case that fails, either instrument `to_decimal` with logging statements or run your test in the debugger, and figure out why it fails.

christopherlam:
This is the part which throws exception:

https://github.com/Gnucash/gnucash/blob/b6c0a62bbd33f63462e10bd8fcbaacd85ca8e7ec/libgnucash/engine/gnc-numeric.cpp#L381

jralls:
The actual fix is
```diff
--- a/libgnucash/engine/gnc-numeric.cpp
+++ b/libgnucash/engine/gnc-numeric.cpp
@@ -368,7 +368,7 @@ GncNumeric::to_decimal(unsigned int max_places) const
         rr_num *= factor;
         rr_den *= factor;
     }
-    while (!rr_num.isZero() && rr_num % 10 == 0)
+    while (!rr_num.isZero() && rr_num > 9 && rr_den > 9 && rr_num % 10 == 0)
     {
         rr_num /= 10;
         rr_den /= 10;
```
Without the protection if the numerator has more 0s than the denominator then the denominator goes to 0 and trying to create a `GncRational` with a 0 denominator quite reasonably throws.

I've pushed that fix along with the tests from your example up top.

PR #817, opened 2020-11-11
Bug 379126 - Combine multiple splits for the same transaction

Bob-IT:
This started from @christopherlam  bug 797998 which linked to bug 379126 by @gjanssens 

Currently when you have multiple splits for the same account in the same transaction they will appear in the register multiple times, one for each split when viewed in basic view. This has been confusing for users so this commit changes that so with the basic view, the splits values for the account are added together.

The action field is replaced with '-- Multi Line --' as suggested in the bug when the split actions are all blank or different as seen in the images.

I was surprised that you could change the debit/credit value in basic view when '-- Split Transaction --' is shown, you do get the little squares but for this change I think it would be safer to disallow changes to debit/credit and action fields when it is a multi line transaction and may be the '-- Split Transaction --' should be changed to some thing else, maybe '-- Multi Home Split Transaction --', not really sure what .

You will see for prices I have used the average value, not sure if that makes sense so maybe it should be blank.

Have enabled trading and the use transaction num/action swap and that looks OK. 
On the registers I have tested, the sums seem to be correct but further testing will be required.
Thoughts.

Some Screen shots...
Before
checking-before
stock-before
After
checking-after
stock-after

### DIFF
--- a/gnucash/register/ledger-core/split-register-load.c
+++ b/gnucash/register/ledger-core/split-register-load.c
@@ -556,8 +556,8 @@ gnc_split_register_load (SplitRegister* reg, GList* slist,
         }
     }
 
-    if (multi_line)
-        trans_table = g_hash_table_new (g_direct_hash, g_direct_equal);
+    /* create the has table for duplicat transactions */
+    trans_table = g_hash_table_new (g_direct_hash, g_direct_equal);
 
     /* populate the table */
     for (node = slist; node; node = node->next)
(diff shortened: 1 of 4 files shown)


christopherlam:
Interesting. Also needs testing:
- opening trading acct register
- open APAR register - Txn Type Invoice should be fine, Payment should be tested and a suitable 'multiple-invoice-payment' comment; Link would probably show blank amount.
- try complex stock book such as latest in bug 797796. I suspect price/amount would be meaningless.

Bob-IT:
Closing for now and will recreate/reopen after further development.

PR #818, opened 2020-11-12
Stock editor and ACB Calculator

christopherlam:
In progress: UI for creating well-crafted stock transactions.

- [x] stock transactions must have asset, asset (broker cash) income (capgains), income (dividends), expenses (commissions/fees)
- [x] these accounts will have split values with sign rules eg expenses can't be negative
- [x] these accounts *must* have currency same as stock's parent currency
- [x] these accounts should be saved/retrieved in stock-acct metadata
- [x] pricedb updated for stock vs parent currency
- [ ] the posted txns *may* be locked for editing and have special type TXN-TYPE-STOCK if we want ui to *edit* posted txns as well. For now it'll be to *create* these txns only.

Further thoughts
- `Account.cpp` to be upgraded to have `xaccAccountSetLinkedAccount (account, id, target_account)` and `xaccAccountGetLinkedAccount (account, id)` to store/retrieve in account's kvp. Thus other reports may retrieve such metadata.
- xaccAccountSetLinkedAccount (stock_acct, "proceeds-acct", asset_cash_proceeds)
- xaccAccountSetLinkedAccount (stock_acct, "capgains-acct", income_capgains)
- xaccAccountSetLinkedAccount (stock_acct, "dividend-acct", income_dividend)
- xaccAccountSetLinkedAccount (stock_acct, "expense-acct", expense_fees)

image

### DIFF
--- a/gnucash/gnome/assistant-stock-transaction.cpp
+++ b/gnucash/gnome/assistant-stock-transaction.cpp
@@ -1,0 +1,1225 @@
+/********************************************************************\
+ * assistant-stock-transaction.cpp -- stock assistant for GnuCash   *
+ * Copyright (C) 2022 Christopher Lam                               *
+ *                                                                  *
+ * This program is free software; you can redistribute it and/or    *
+ * modify it under the terms of the GNU General Public License as   *
+ * published by the Free Software Foundation; either version 2 of   *
+ * the License, or (at your option) any later version.              *
+ *                                                                  *
+ * This program is distributed in the hope that it will be useful,  *
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of   *
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the    *
+ * GNU General Public License for more details.                     *
+ *                                                                  *
+ * You should have received a copy of the GNU General Public License*
+ * along with this program; if not, contact:                        *
+ *                                                                  *
+ * Free Software Foundation           Voice:  +1-617-542-5942       *
+ * 51 Franklin Street, Fifth Floor    Fax:    +1-617-542-2652       *
+ * Boston, MA  02110-1301,  USA       gnu@gnu.org                   *
+\********************************************************************/
+
+#include <config.h>
+
+#include <gtk/gtk.h>
+#include <glib/gi18n.h>
+#include <vector>
+#include <string>
+#include <numeric>
+#include <algorithm>
+
+extern "C" {
+#include "Transaction.h"
+#include "engine-helpers.h"
+#include "dialog-utils.h"
+#include "assistant-stock-transaction.h"
+#include "gnc-account-sel.h"
+#include "gnc-amount-edit.h"
+#include "gnc-component-manager.h"
+#include "gnc-date-edit.h"
+#include "gnc-tree-view-account.h"
+
+static QofLogModule log_module = GNC_MOD_ASSISTANT;
+
+void stock_assistant_prepare (GtkAssistant  *assistant, GtkWidget *page,
+                              gpointer user_data);
+void stock_assistant_finish  (GtkAssistant *assistant, gpointer user_data);
+void stock_assistant_cancel  (GtkAssistant *gtkassistant, gpointer user_data);
+}
+
+enum class FieldMask : unsigned;
+bool operator &(FieldMask lhs, FieldMask rhs);
+FieldMask operator |(FieldMask lhs, FieldMask rhs);
+FieldMask operator ^(FieldMask lhs, FieldMask rhs);
+
+static const char* ASSISTANT_STOCK_TRANSACTION_CM_CLASS = "assistant-stock-transaction";
+
+enum assistant_pages
+{
+    PAGE_INTRO = 0,
+    PAGE_TRANSACTION_TYPE,
+    PAGE_TRANSATION_DETAILS,
+    PAGE_STOCK_AMOUNT,
+    PAGE_STOCK_VALUE,
+    PAGE_CASH,
+    PAGE_FEES,
+    PAGE_DIVIDEND,
+    PAGE_CAPGAINS,
+    PAGE_FINISH
+};
+
+enum split_cols
+{
+    SPLIT_COL_ACCOUNT = 0,
+    SPLIT_COL_MEMO,
+    SPLIT_COL_DEBIT,
+    SPLIT_COL_CREDIT,
+    NUM_SPLIT_COLS
+};
+
+/** structures *********************************************************/
+
+enum class FieldMask : unsigned
+{
+    DISABLED = 0,
+    ENABLED_DEBIT,
+    ENABLED_CREDIT,
+    ALLOW_ZERO = 4,
+    ALLOW_NEGATIVE = 8
+};
+
+FieldMask operator |(FieldMask lhs, FieldMask rhs)
+{
+    return static_cast<FieldMask> (static_cast<unsigned>(lhs) |
+                                   static_cast<unsigned>(rhs));
+};
+
+bool operator &(FieldMask lhs, FieldMask rhs)
+{
+    return (static_cast<unsigned>(lhs) & static_cast<unsigned>(rhs));
+};
+
+FieldMask operator ^(FieldMask lhs, FieldMask rhs)
+{
+    return static_cast<FieldMask> (static_cast<unsigned>(lhs) ^
+                                   static_cast<unsigned>(rhs));
+};
+
+struct TxnTypeInfo
+{
+    FieldMask stock_amount;
+    FieldMask stock_value;
+    FieldMask cash_value;
+    FieldMask fees_value;
+    bool fees_capitalize;
+    FieldMask dividend_value;
+    FieldMask capgains_value;
+    const char* friendly_name;
+    const char* explanation;
+};
+
+using StringVec = std::vector<std::string>;
+using TxnTypeVec = std::vector<TxnTypeInfo>;
+using AccountVec = std::vector<Account*>;
+
+static const TxnTypeVec starting_types
+{
+    {
+        FieldMask::ENABLED_DEBIT,          // stock_amt
+        FieldMask::ENABLED_DEBIT,          // stock_val
+        FieldMask::ENABLED_CREDIT,         // cash_amt
+        FieldMask::ENABLED_DEBIT | FieldMask::ALLOW_ZERO,          // fees_amt
+        true,                   // fees_capitalize
+        FieldMask::DISABLED,               // dividend_amt
+        FieldMask::DISABLED,               // capg_amt
+        // Translators: this is a stock transaction describing an
+        // Initial stock purchase
+        NC_("Stock Assistant: Transaction Type", "Open buy"),
+        N_("Initial stock purchase")
+    },
+    {
+        FieldMask::ENABLED_CREDIT,         // stock_amt
+        FieldMask::ENABLED_CREDIT,         // stock_val
+        FieldMask::ENABLED_DEBIT,          // cash_amt
+        FieldMask::ENABLED_DEBIT | FieldMask::ALLOW_ZERO,          // fees_amt
+        true,                   // fees_capitalize
+        FieldMask::DISABLED,               // dividend_amt
+        FieldMask::DISABLED,               // capg_amt
+        // Translators: this is a stock transaction describing an
+        // initial stock short-sale
+        NC_("Stock Assistant: Transaction Type", "Open short"),
+        N_("Initial stock short-sale")
+    }
+};
+
+static const TxnTypeVec long_types
+{
+    {
+        FieldMask::ENABLED_DEBIT,          // stock_amt
+        FieldMask::ENABLED_DEBIT,          // stock_val
+        FieldMask::ENABLED_CREDIT,         // cash_amt
+        FieldMask::ENABLED_DEBIT | FieldMask::ALLOW_ZERO,          // fees_amt
+        true,                   // fees_capitalize
+        FieldMask::DISABLED,               // dividend_amt
+        FieldMask::DISABLED,               // capg_amt
+        // Translators: this is a stock transaction describing
+        // purchase of stock.
+        NC_("Stock Assistant: Transaction Type", "Buy"),
+        N_("Buying stock.")
+    },
+    {
+        FieldMask::ENABLED_CREDIT,         // stock_amt
+        FieldMask::ENABLED_CREDIT,         // stock_val
+        FieldMask::ENABLED_DEBIT,          // cash_amt
+        FieldMask::ENABLED_DEBIT | FieldMask::ALLOW_ZERO,          // fees_amt
+        false,                  // fees_capitalize
+        FieldMask::DISABLED,               // dividend_amt
+        FieldMask::ENABLED_CREDIT | FieldMask::ALLOW_ZERO | FieldMask::ALLOW_NEGATIVE, // capgains_amt
+        // Translators: this is a stock transaction describing sale of
+        // stock, and recording capital gains/loss
+        NC_("Stock Assistant: Transaction Type", "Sell"),
+        N_("Selling stock, and record capital gains/loss")
+    },
+    {
+        FieldMask::DISABLED,               // stock_amt
+        FieldMask::DISABLED,               // stock_val
+        FieldMask::ENABLED_DEBIT,          // cash_amt
+        FieldMask::ENABLED_DEBIT | FieldMask::ALLOW_ZERO,          // fees_amt
+        false,                  // fees_capitalize
+        FieldMask::ENABLED_CREDIT,         // dividend_amt
+        FieldMask::DISABLED,               // capg_amt
+        // Translators: this is a stock transaction describing
+        // dividends issued to holder
+        NC_("Stock Assistant: Transaction Type", "Dividend"),
+        N_("Company issues dividends to holder")
+    },
+    {
+        FieldMask::ENABLED_DEBIT,          // stock_amt
+        FieldMask::ENABLED_DEBIT,          // stock_val
+        FieldMask::ENABLED_DEBIT,          // cash_amt
+        FieldMask::ENABLED_DEBIT | FieldMask::ALLOW_ZERO,          // fees_amt
+        true,                   // fees_capitalize
+        FieldMask::ENABLED_CREDIT,         // dividend_amt
+        FieldMask::DISABLED,               // capg_amt
+        // Translators: this is a stock transaction describing
+        // dividend issued to holder, and is reinvested. Some
+        // dividends are distributed as cash.
+        NC_("Stock Assistant: Transaction Type", "Dividend reinvestment (w/ remainder)"),
+        N_("Company issues dividend which is reinvested. Some dividends are paid to holder")
+    },
+    {
+        FieldMask::ENABLED_DEBIT,          // stock_amt
+        FieldMask::ENABLED_DEBIT,          // stock_val
+        FieldMask::DISABLED,               // cash_amt
+        FieldMask::ENABLED_DEBIT | FieldMask::ALLOW_ZERO,          // fees_amt
+        true,                   // fees_capitalize
+        FieldMask::ENABLED_CREDIT,         // dividend_amt
+        FieldMask::DISABLED,               // capg_amt
+        // Translators: this is a stock transaction describing
+        // dividend which is wholly reinvested.
+        NC_("Stock Assistant: Transaction Type", "Dividend reinvestment (w/o remainder)"),
+        N_("Company issues dividend which is wholly reinvested.")
+    },
+    {
+        FieldMask::DISABLED,               // stock_amt
+        FieldMask::ENABLED_CREDIT,         // stock_val
+        FieldMask::ENABLED_DEBIT,          // cash_amt
+        FieldMask::ENABLED_DEBIT | FieldMask::ALLOW_ZERO,          // fees_amt
+        true,                   // fees_capitalize
+        FieldMask::DISABLED,               // dividend_amt
+        FieldMask::DISABLED,               // capg_amt
+        // Translators: this is a stock transaction describing return
+        // of capital
+        NC_("Stock Assistant: Transaction Type", "Return of Capital"),
+        N_("Stock returns capital to holder")
+    },
+    {
+        FieldMask::DISABLED,               // stock_amt
+        FieldMask::ENABLED_DEBIT,          // stock_val
+        FieldMask::DISABLED,               // cash_amt
+        FieldMask::ENABLED_DEBIT | FieldMask::ALLOW_ZERO,          // fees_amt
+        false,                  // fees_capitalize
+        FieldMask::ENABLED_CREDIT,         // dividend_amt
+        FieldMask::DISABLED,               // capg_amt
+        // Translators: this is a stock transaction describing a
+        // notional distribution
+        NC_("Stock Assistant: Transaction Type", "Notional distribution"),
+        N_("Stock returns a notional distribution")
+    },
+    {
+        FieldMask::ENABLED_DEBIT,          // stock_amt
+        FieldMask::DISABLED,               // stock_val
+        FieldMask::DISABLED,               // cash_amt
+        FieldMask::ENABLED_DEBIT | FieldMask::ALLOW_ZERO,          // fees_amt
+        true,                   // fees_capitalize
+        FieldMask::DISABLED,               // dividend_amt
+        FieldMask::DISABLED,               // capg_amt
+        // Translators: this is a stock transaction describing a stock
+        // split
+        NC_("Stock Assistant: Transaction Type", "Stock split"),
+        N_("Stock price is fractionally reduced")
+    },
+    {
+        FieldMask::ENABLED_CREDIT,         // stock_amt
+        FieldMask::DISABLED,               // stock_val
+        FieldMask::DISABLED,               // cash_amt
+        FieldMask::ENABLED_DEBIT | FieldMask::ALLOW_ZERO,          // fees_amt
+        true,                   // fees_capitalize
+        FieldMask::DISABLED,               // dividend_amt
+        FieldMask::DISABLED,               // capg_amt
+        // Translators: this is a stock transaction describing a reverse split
+        NC_("Stock Assistant: Transaction Type", "Reverse split"),
+        N_("Stock price is fractionally increased.")
+    },
+    {
+        FieldMask::ENABLED_CREDIT,         // stock_amt
+        FieldMask::ENABLED_CREDIT,         // stock_val
+        FieldMask::ENABLED_DEBIT,          // cash_amt
+        FieldMask::ENABLED_DEBIT | FieldMask::ALLOW_ZERO,          // fees_amt
+        false,                  // fees_capitalize
+        FieldMask::DISABLED,               // dividend_amt
+        FieldMask::ENABLED_CREDIT | FieldMask::ALLOW_ZERO | FieldMask::ALLOW_NEGATIVE, // capgains_amt
+        // Translators: this is a stock transaction describing a
+        // reverse split. Some fractional stock is returned as cash.
+        NC_("Stock Assistant: Transaction Type", "Reverse split w/ cash in lieu for fractionals"),
+        N_("Stock price is fractionally increased. Fractional remaining stock is returned as cash.")
+    },
+};
+
+static const TxnTypeVec short_types
+{
+    {
+        FieldMask::ENABLED_DEBIT,          // stock_amt
+        FieldMask::ENABLED_DEBIT,          // stock_val
+        FieldMask::ENABLED_CREDIT,         // cash_amt
+        FieldMask::ENABLED_DEBIT | FieldMask::ALLOW_ZERO,          // fees_amt
+        true,                   // fees_capitalize
+        FieldMask::DISABLED,               // dividend_amt
+        FieldMask::DISABLED,               // capg_amt
+        // Translators: this is a stock transaction describing
+        // shorting of stock.
+        NC_("Stock Assistant: Transaction Type", "Short sell"),
+        N_("Shorting stock.")
+    },
+    {
+        FieldMask::ENABLED_DEBIT,          // stock_amt
+        FieldMask::ENABLED_DEBIT,          // stock_val
+        FieldMask::ENABLED_CREDIT,         // cash_amt
+        FieldMask::ENABLED_DEBIT | FieldMask::ALLOW_ZERO,          // fees_amt
+        false,                  // fees_capitalize
+        FieldMask::DISABLED,               // dividend_amt
+        FieldMask::ENABLED_DEBIT | FieldMask::ALLOW_ZERO | FieldMask::ALLOW_NEGATIVE,          // capg_amt
+        // Translators: this is a stock transaction describing cover
+        // buying stock, and recording capital gains/loss
+        NC_("Stock Assistant: Transaction Type", "Cover buy"),
+        N_("Cover buying stock, and record capital gains/loss")
+    },
+    {
+        FieldMask::DISABLED,               // stock_amt
+        FieldMask::DISABLED,               // stock_val
+        FieldMask::ENABLED_CREDIT,         // cash_amt
+        FieldMask::ENABLED_DEBIT | FieldMask::ALLOW_ZERO,          // fees_amt
+        false,                  // fees_capitalize
+        FieldMask::ENABLED_DEBIT,          // dividend_amt
+        FieldMask::DISABLED,               // capg_amt
+        // Translators: this is a stock transaction describing
+        // dividends retrieved from holder when shorting stock
+        NC_("Stock Assistant: Transaction Type", "Compensatory Dividend"),
+        N_("Company issues dividends to holder when shorting stock")
+    },
+    {
+        FieldMask::ENABLED_CREDIT,         // stock_amt
+        FieldMask::ENABLED_CREDIT,         // stock_val
+        FieldMask::ENABLED_DEBIT,          // cash_amt
+        FieldMask::ENABLED_DEBIT | FieldMask::ALLOW_ZERO,          // fees_amt
+        true,                   // fees_capitalize
+        FieldMask::ENABLED_DEBIT,          // dividend_amt
+        FieldMask::DISABLED,               // capg_amt
+        // Translators: this is a stock transaction describing
+        // dividends retrieved from holder when shorting stock. Some
+        // dividends are recovered from holder
+        NC_("Stock Assistant: Transaction Type", "Dividend reinvestment (w/ remainder)"),
+        N_("Company issues dividends to holder when shorting stock. Some dividends are recovered from holder")
+    },
+    {
+        FieldMask::ENABLED_DEBIT,          // stock_amt
+        FieldMask::ENABLED_DEBIT,          // stock_val
+        FieldMask::DISABLED,               // cash_amt
+        FieldMask::ENABLED_DEBIT | FieldMask::ALLOW_ZERO,          // fees_amt
+        true,                   // fees_capitalize
+        FieldMask::ENABLED_DEBIT,          // dividend_amt
+        FieldMask::DISABLED,               // capg_amt
+        // Translators: this is a stock transaction describing
+        // dividends retrieved from holder when shorting stock,
+        NC_("Stock Assistant: Transaction Type", "Dividend reinvestment (w/o remainder)"),
+        N_("Company issues dividend when shorting stock, which are wholly recovered from holder.")
+    },
+    {
+        FieldMask::DISABLED,               // stock_amt
+        FieldMask::ENABLED_DEBIT,          // stock_val
+        FieldMask::ENABLED_CREDIT,         // cash_amt
+        FieldMask::ENABLED_DEBIT | FieldMask::ALLOW_ZERO,          // fees_amt
+        true,                   // fees_capitalize
+        FieldMask::DISABLED,               // dividend_amt
+        FieldMask::DISABLED,               // capg_amt
+        // Translators: this is a stock transaction describing return
+        // of capital retrieved from holder when shorting stock
+        NC_("Stock Assistant: Transaction Type", "Compensatory Return of Capital"),
+        N_("Return retrieves capital from holder when shorting stock")
+    },
+    {
+        FieldMask::DISABLED,               // stock_amt
+        FieldMask::ENABLED_CREDIT,         // stock_val
+        FieldMask::DISABLED,               // cash_amt
+        FieldMask::ENABLED_DEBIT | FieldMask::ALLOW_ZERO,          // fees_amt
+        false,                  // fees_capitalize
+        FieldMask::ENABLED_DEBIT,          // dividend_amt
+        FieldMask::DISABLED,               // capg_amt
+        // Translators: this is a stock transaction describing a
+        // notional distribution when shorting stock
+        NC_("Stock Assistant: Transaction Type", "Compensatory Notional distribution"),
+        N_("Stock retrieves a notional distribution")
+    },
+    {
+        FieldMask::ENABLED_CREDIT,         // stock_amt
+        FieldMask::DISABLED,               // stock_val
+        FieldMask::DISABLED,               // cash_amt
+        FieldMask::ENABLED_DEBIT | FieldMask::ALLOW_ZERO,          // fees_amt
+        true,                   // fees_capitalize
+        FieldMask::DISABLED,               // dividend_amt
+        FieldMask::DISABLED,               // capg_amt
+        // Translators: this is a stock transaction describing a stock
+        // split when shorting stock
+        NC_("Stock Assistant: Transaction Type", "Stock split"),
+        N_("Stock price is fractionally reduced when shorting stock")
+    },
+    {
+        FieldMask::ENABLED_DEBIT,          // stock_amt
+        FieldMask::DISABLED,               // stock_val
+        FieldMask::DISABLED,               // cash_amt
+        FieldMask::ENABLED_DEBIT | FieldMask::ALLOW_ZERO,          // fees_amt
+        true,                   // fees_capitalize
+        FieldMask::DISABLED,               // dividend_amt
+        FieldMask::DISABLED,               // capg_amt
+        // Translators: this is a stock transaction describing a
+        // reverse split when shorting stock.
+        NC_("Stock Assistant: Transaction Type", "Reverse split"),
+        N_("Stock price is fractionally increased when shorting stock.")
+    },
+    {
+        FieldMask::ENABLED_DEBIT,          // stock_amt
+        FieldMask::ENABLED_DEBIT,          // stock_val
+        FieldMask::ENABLED_CREDIT,         // cash_amt
+        FieldMask::ENABLED_DEBIT | FieldMask::ALLOW_ZERO,          // fees_amt
+        false,                  // fees_capitalize
+        FieldMask::DISABLED,               // dividend_amt
+        FieldMask::ENABLED_DEBIT | FieldMask::ALLOW_ZERO | FieldMask::ALLOW_NEGATIVE,          // capg_amt
+        // Translators: this is a stock transaction describing a
+        // reverse split when shorting stock. Fractional remaining
+        // stock is retrieved as cash.
+        NC_("Stock Assistant: Transaction Type", "Reverse split w/ cash in lieu for fractionals"),
+        N_("Stock price is fractionally increased when shorting stock. Fractional remaining stock is retrieved as cash.")
+    },
+};
+
+typedef struct
+{
+    GtkWidget * window;
+    GtkWidget * assistant;
+
+    const TxnTypeVec * txn_types;
+    Account   * acct;
+    gnc_commodity * currency;
+
+    // transaction type page
+    GtkWidget * transaction_type_page;
+    GtkWidget * transaction_type_combo;
+    GtkWidget * transaction_type_explanation;
+    const TxnTypeInfo * txn_type;
+
+    // transaction details page
+    GtkWidget * transaction_details_page;
+    GtkWidget * date_edit;
+    GtkWidget * transaction_description_entry;
+
+    // stock amount page
+    gnc_numeric balance_at_date;
+    GtkWidget * stock_amount_page;
+    GtkWidget * prev_amount;
+    GtkWidget * next_amount;
+    GtkWidget * stock_amount_edit;
+
+    // stock value page
+    GtkWidget * stock_value_page;
+    GtkWidget * stock_value_edit;
+    GtkWidget * price_value;
+    GtkWidget * stock_memo_edit;
+
+    // cash page
+    GtkWidget * cash_page;
+    GtkWidget * cash_account;
+    GtkWidget * cash_memo_edit;
+    GtkWidget * cash_value;
+
+    // fees page
+    GtkWidget * fees_page;
+    GtkWidget * capitalize_fees_checkbox;
+    GtkWidget * fees_account;
+    GtkWidget * fees_memo_edit;
+    GtkWidget * fees_value;
+
+    // dividend page
+    GtkWidget * dividend_page;
+    GtkWidget * dividend_account;
+    GtkWidget * dividend_memo_edit;
+    GtkWidget * dividend_value;
+
+    // capgains page
+    GtkWidget * capgains_page;
+    GtkWidget * capgains_account;
+    GtkWidget * capgains_memo_edit;
+    GtkWidget * capgains_value;
+
+    // finish page
+    GtkWidget * finish_page;
+    GtkWidget * finish_split_view;
+    GtkWidget * finish_summary;
+} StockTransactionInfo;
+
+
+/******* implementations ***********************************************/
+static void
+stock_assistant_window_destroy_cb (GtkWidget *object, gpointer user_data)
+{
+    auto info = static_cast<StockTransactionInfo*>(user_data);
+    gnc_unregister_gui_component_by_data (ASSISTANT_STOCK_TRANSACTION_CM_CLASS, info);
+    g_free (info);
+}
+
+static void
+refresh_page_transaction_type (GtkWidget *widget, gpointer user_data)
+{
+    auto info = static_cast<StockTransactionInfo*>(user_data);
+
+    auto type_idx = gtk_combo_box_get_active (GTK_COMBO_BOX (widget));
+    if (type_idx < 0)           // combo isn't initialized yet.
+        return;
+
+    info->txn_type = &(info->txn_types->at (type_idx));
+
+    gtk_label_set_text (GTK_LABEL (info->transaction_type_explanation),
+                        _(info->txn_type->explanation));
+
+    // set default capitalize fees setting
+    gtk_toggle_button_set_active (GTK_TOGGLE_BUTTON (info->capitalize_fees_checkbox),
+                                  info->txn_type->fees_capitalize);
+}
+
+
+static void
+refresh_page_stock_amount (GtkWidget *widget, gpointer user_data)
+{
+    auto info = static_cast<StockTransactionInfo*>(user_data);
+
+    auto pinfo = gnc_commodity_print_info (xaccAccountGetCommodity (info->acct), true);
+    auto bal = info->balance_at_date;
+    gtk_label_set_text (GTK_LABEL(info->prev_amount), xaccPrintAmount (bal, pinfo));
+
+    gnc_numeric stock_amount;
+
+    if (gnc_amount_edit_expr_is_valid (GNC_AMOUNT_EDIT (info->stock_amount_edit),
+                                       &stock_amount, true, nullptr))
+        gtk_label_set_text (GTK_LABEL(info->next_amount), nullptr);
+    else
+    {
+        if (info->txn_type->stock_amount == FieldMask::ENABLED_CREDIT)
+            stock_amount = gnc_numeric_neg (stock_amount);
+        bal = gnc_numeric_add_fixed (bal, stock_amount);
+
+        gtk_label_set_text (GTK_LABEL(info->next_amount),
+                            xaccPrintAmount (bal, pinfo));
+    }
+}
+
+
+static void
+refresh_page_stock_value (GtkWidget *widget, gpointer user_data)
+{
+    auto info = static_cast<StockTransactionInfo*>(user_data);
+    gnc_numeric amount, value;
+
+    if (info->txn_type->stock_amount == FieldMask::DISABLED ||
+        info->txn_type->stock_value == FieldMask::DISABLED ||
+        gnc_amount_edit_expr_is_valid (GNC_AMOUNT_EDIT (info->stock_amount_edit), &amount, true, nullptr) ||
+        gnc_amount_edit_expr_is_valid (GNC_AMOUNT_EDIT (info->stock_value_edit),  &value,  true, nullptr) || 
+        gnc_numeric_zero_p (value))
+    {
+        // Translators: StockAssistant: N/A denotes stock price is not computable
+        const char* na_label =  N_("N/A");
+        gtk_label_set_text (GTK_LABEL (info->price_value), _(na_label));
+        return;
+    }
+
+    auto price = gnc_numeric_div (value, amount, GNC_DENOM_AUTO, GNC_HOW_RND_ROUND);
+    auto pinfo = gnc_commodity_print_info (info->currency, true);
+    gtk_label_set_text (GTK_LABEL (info->price_value), xaccPrintAmount (price, pinfo));
+}
+
+static void
+refresh_page_cash (GtkWidget *widget, gpointer user_data)
+{
+    return;
+}
+
+static void
+refresh_page_fees (GtkWidget *widget, gpointer user_data)
+{
+    auto info = static_cast<StockTransactionInfo*>(user_data);
+    auto capitalize_fees = gtk_toggle_button_get_active
+        (GTK_TOGGLE_BUTTON (info->capitalize_fees_checkbox));
+    gtk_widget_set_sensitive (info->fees_account, !capitalize_fees);
+}
+
+static void
+refresh_page_dividend (GtkWidget *widget, gpointer user_data)
+{
+    return;
+}
+
+static void
+refresh_page_capgains (GtkWidget *widget, gpointer user_data)
+{
+    return;
+}
+
+static void
+add_error (StringVec& errors, const char* format_str, const char* arg)
+{
+    gchar *buf = g_strdup_printf (_(format_str),
+                                  g_dpgettext2 (nullptr, "Stock Assistant: Page name", arg));
+    errors.emplace_back (buf);
+    g_free (buf);
+}
+
+static void
+add_error_str (StringVec& errors, const char* str)
+{
+    errors.emplace_back (_(str));
+}
+
+static void
+check_page (GtkListStore *list, gnc_numeric& debit, gnc_numeric& credit,
+            FieldMask splitfield, Account *acct, GtkWidget *memo, GtkWidget *gae,
+            gnc_commodity *comm, bool ignore_account,
+            const char* page, StringVec& errors)
+{
+    if (splitfield == FieldMask::DISABLED)
+        return;
+    const char* missing_str = N_("(missing)");
+    const gchar* amtstr;
+    gnc_numeric amount;
+    bool debit_side = (splitfield & FieldMask::ENABLED_DEBIT);
+
+    if (gnc_amount_edit_expr_is_valid (GNC_AMOUNT_EDIT (gae), &amount, true, nullptr))
+    {
+        if (splitfield & FieldMask::ALLOW_ZERO)
+            amtstr = "";
+        else
+        {
+            add_error (errors, N_("Amount for %s is missing"), page);
+            amtstr = _(missing_str);
+        }
+    }
+    else
+    {
+        if (!(splitfield & FieldMask::ALLOW_NEGATIVE))
+        {
+            if ((splitfield & FieldMask::ALLOW_ZERO) && gnc_numeric_negative_p (amount))
+                add_error (errors, N_("Amount for %s must not be negative."), page);
+            else if (!(splitfield & FieldMask::ALLOW_ZERO) && !gnc_numeric_positive_p (amount))
+                add_error (errors, N_("Amount for %s must be positive."), page);
+        }
+        if (gnc_numeric_negative_p (amount))
+        {
+            amount = gnc_numeric_neg (amount);
+            debit_side = !debit_side;
+        }
+        if (debit_side)
+            debit = gnc_numeric_add_fixed (debit, amount);
+        else
+            credit = gnc_numeric_add_fixed (credit, amount);
+        amtstr = xaccPrintAmount (amount, gnc_commodity_print_info (comm, true));
+    }
+
+    auto memostr = gtk_entry_get_text (GTK_ENTRY (memo));
+    const gchar *acctstr;
+
+    if (ignore_account)
+        acctstr = "";
+    else if (acct)
+        acctstr = xaccAccountGetName (acct);
+    else
+    {
+        add_error (errors, N_("Account for %s is missing"), page);
+        acctstr = _(missing_str);
+    }
+
+    GtkTreeIter iter;
+    gtk_list_store_append (list, &iter);
+    gtk_list_store_set (list, &iter,
+                        SPLIT_COL_ACCOUNT, acctstr,
+                        SPLIT_COL_MEMO, memostr,
+                        SPLIT_COL_DEBIT, debit_side ? amtstr : "",
+                        SPLIT_COL_CREDIT, !debit_side ? amtstr : "",
+                        -1);
+}
+
+static inline Account*
+gas_account (GtkWidget *gas)
+{
+    return gnc_account_sel_get_account (GNC_ACCOUNT_SEL (gas));
+}
+
+static void
+refresh_page_finish (StockTransactionInfo *info)
+{
+    auto view = GTK_TREE_VIEW (info->finish_split_view);
+    auto list = GTK_LIST_STORE (gtk_tree_view_get_model(view));
+    gtk_list_store_clear (list);
+
+    gnc_numeric debit = gnc_numeric_zero ();
+    gnc_numeric credit = gnc_numeric_zero ();
+    StringVec errors;
+
+    if (info->txn_type->stock_amount != FieldMask::DISABLED)
+    {
+        auto stock_amount = gnc_amount_edit_get_amount
+            (GNC_AMOUNT_EDIT(info->stock_amount_edit));
+        if (info->txn_type->stock_amount & FieldMask::ENABLED_CREDIT)
+            stock_amount = gnc_numeric_neg (stock_amount);
+        auto new_bal = gnc_numeric_add_fixed (info->balance_at_date, stock_amount);
+        if (gnc_numeric_positive_p (info->balance_at_date) &&
+            gnc_numeric_negative_p (new_bal))
+            add_error_str (errors, N_("Cannot sell more units than owned"));
+        else if (gnc_numeric_negative_p (info->balance_at_date) &&
+                 gnc_numeric_positive_p (new_bal))
+            add_error_str (errors, N_("Cannot cover buy more units than owed"));
+    }
+
+    check_page (list, debit, credit, info->txn_type->stock_value, info->acct,
+                info->stock_memo_edit, info->stock_value_edit, info->currency,
+                false, NC_ ("Stock Assistant: Page name", "stock value"), errors);
+
+    check_page (list, debit, credit, info->txn_type->cash_value,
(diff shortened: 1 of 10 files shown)


fellen:
Sorry, I was not aware the check boxes are clickable.

What is the purpose of this PR?
Do you want to replace another dialog or will it become an additional dialog?
Where is the border to other financial instruments like bonds, funds, options, …?
How can I trade shares in foreign currency while finally reporting in domestic currency?
Perhaps more questions will appear.

christopherlam:
> What is the purpose of this PR?
> Do you want to replace another dialog or will it become an additional dialog?
> Where is the border to other financial instruments like bonds, funds, options, …?
> How can I trade shares in foreign currency while finally reporting in domestic currency?
> Perhaps more questions will appear.

A dialog to assist creating well-crafted stock transactions. Templated from bug 797796. It'll be a new dialog. These questions will be unanswerable for now, unless CDB-Man in bug 797796 can comment further. Last commit will hide menu item behind `--extra` for beta feedback.

gnucash/gnome/dialog-stock-editor.h
```diff
@@ -0,0 +1,31 @@
+/********************************************************************\
+ * window-autoclear.h -- the autoclear window                       *
+ * Copyright (C) 2010 Cristian KLEIN                                *
```
Bob-IT, line 3: Needs changing

gnucash/gnome/dialog-stock-editor.h
```diff
@@ -0,0 +1,31 @@
...
+ * You should have received a copy of the GNU General Public License*
+ * along with this program; if not, contact:                        *
+ *                                                                  *
+ * Free Software Foundation           Voice:  +1-617-542-5942       *
+ * 51 Franklin Street, Fifth Floor    Fax:    +1-617-542-2652       *
+ * Boston, MA  02110-1301,  USA       gnu@gnu.org                   *
+\********************************************************************/
+
+#ifndef WINDOW_STOCKEDITOR_H
+#define WINDOW_STOCKEDITOR_H
```
Bob-IT, line 24: I would think these would best be DIALOG_

gnucash/gnome/dialog-stock-editor.h
```diff
@@ -0,0 +1,31 @@
...
+
+typedef struct _StockEditorWindow StockEditorWindow;
+
+StockEditorWindow *stockEditorWindow (GtkWidget *parent, Account *account);
+void gnc_ui_stockeditor_window_raise (StockEditorWindow * autoClearData);
```
Bob-IT, line 29: Not realy sure why you have done this as it seems to be not required unless you intend using the StockEditorWindow.
Have a look at the dialog-imap-editor
void gnc_stock_editor_dialog (GtkWidget *parent, Account *account);

gnucash/gnome/dialog-stock-editor.c
```diff
@@ -0,0 +1,689 @@
...
+#include "gnc-gnome-utils.h"
+#include "gnc-helpers.h"
+#include "gnc-main-window.h"
+#include "gnc-plugin-page-register.h"
+#include "gnc-session.h"
+#include "gnc-ui.h"
+#include "gnc-ui-balances.h"
+#include "dialog-stock-editor.h"
+
+#define DIALOG_BOOK_CLOSE_CM_CLASS "dialog-book-close"
```
Bob-IT, line 44: This may need changing or removing depending on if you change other stuff

gnucash/gnome/dialog-stock-editor.c
```diff
@@ -0,0 +1,689 @@
...
+    newstr = g_strconcat (prev_str, "\n", str, NULL);
+    g_free (str);
+    g_free (prev_str);
+    return newstr;
+}
+
+static void check_signs (GNCAmountEdit *gae, gchar mask, gchar **status,
+                         gboolean *passes, gchar *type)
+{
+    gchar *sign;
```
Bob-IT, line 173: this needs to be gchar *sign = NULL; as it may be used uninitialized.

gnucash/gnome/dialog-stock-editor.c
```diff
@@ -0,0 +1,689 @@
...
+        return;
+
+    if (mask == '1') sign = "positive";
+    else if (mask == '2') sign = "zero";
+    else if (mask == '3') sign = "zero or positive";
+    else if (mask == '#') sign = "zero or positive";
+    else if (mask == '4') sign = "negative";
+    else if (mask == '5') sign = "negative or positive";
+    else if (mask == '6') sign = "zero or negative";
+    *status = append_status (*status, g_strdup_printf ("%s must be %s", type, sign));
```
Bob-IT, line 198: Does this need translating?

gnucash/gnome/dialog-stock-editor.c
```diff
@@ -0,0 +1,689 @@
...
+    new_bal = gnc_numeric_add_fixed (old_bal, asset_amount);
+    gtk_label_set_text (data->new_balance_label,
+                        xaccPrintAmount (new_bal, printinfo));
+
+    if (gnc_numeric_negative_p
+        (gnc_numeric_mul
+         (old_bal, new_bal, GNC_DENOM_AUTO, GNC_HOW_RND_ROUND)))
+    {
+        status = append_status
+            (status, g_strdup ("Cannot sell more units than available."));
```
Bob-IT, line 254: Does this need translating?

gnucash/gnome/dialog-stock-editor.c
```diff
@@ -0,0 +1,689 @@
...
+    bal = asset_basis;
+    bal = gnc_numeric_add_fixed (bal, proceeds_val);
+    bal = gnc_numeric_add_fixed (bal, dividend_val);
+    bal = gnc_numeric_add_fixed (bal, fees_val);
+
+    if (gnc_date_edit_get_date (GNC_DATE_EDIT (data->date_entry)) <
+        data->latest_split_date)
+    {
+        status = append_status
+            (status, g_strdup ("Date is before latest split. Balances may not \
```
Bob-IT, line 267: Does this need translating?

gnucash/gnome/dialog-stock-editor.c
```diff
@@ -0,0 +1,689 @@
...
+be valid."));
+    }
+
+    if (req_acc_missing (data->proceeds_acc) ||
+        req_acc_missing (data->fees_acc) ||
+        req_acc_missing (data->capgains_acc) ||
+        req_acc_missing (data->dividend_acc))
+    {
+        status = append_status (status, g_strdup ("Account missing"));
```
Bob-IT, line 276: Does this need translating?

gnucash/gnome/dialog-stock-editor.c
```diff
@@ -0,0 +1,689 @@
...
+        passes = FALSE;
+    }
+
+    printinfo = gnc_commodity_print_info (data->parent_currency, TRUE);
+    if (!gnc_numeric_zero_p (bal))
+    {
+        status = append_status
+            (status, g_strdup_printf ("Imbalance of %s",
```
Bob-IT, line 284: Does this need translating?

gnucash/gnome/dialog-stock-editor.c
```diff
@@ -0,0 +1,689 @@
...
+                                      xaccPrintAmount (bal, printinfo)));
+        passes = FALSE;
+    }
+
+    check_signs (data->asset_amount, mask[0], &status, &passes, "Units");
+    check_signs (data->asset_basis,  mask[1], &status, &passes, "Basis");
+    check_signs (data->proceeds_val, mask[2], &status, &passes, "Proceeds");
+    check_signs (data->dividend_val, mask[3], &status, &passes, "Dividend");
+    check_signs (data->capgains_val, mask[4], &status, &passes, "CapGains");
+    check_signs (data->fees_val, mask[5], &status, &passes, "Fees");
```
Bob-IT, line 294: Does this need translating?

gnucash/gnome/dialog-stock-editor.c
```diff
@@ -0,0 +1,689 @@
...
+                                              G_TYPE_STRING, /* capgains_memo */
+                                              G_TYPE_STRING); /* fees_memo */
+    if (gnc_numeric_positive_p (balance))
+    {
+        add_action (store, "11400#", "Buy", "Source", "", "", "Fees");
+        add_action (store, "441073", "Sell", "Proceeds", "", "Capgains", "Fees");
+        add_action (store, "001403", "Dividend", "Proceeds", "Dividend", "", "Fees");
+        add_action (store, "010400", "Notional Distribution", "", "Notional Distribution", "", "");
+        add_action (store, "041000", "Return of Capital", "Proceeds", "", "", "");
+        add_action (store, "500000", "Stock Split", "", "", "", "");
```
Bob-IT, line 540: Does this need translating?

gnucash/gnome/dialog-stock-editor.c
```diff
@@ -0,0 +1,689 @@
...
+    }
+    else if (gnc_numeric_negative_p (balance))
+    {
+        add_action (store, "44100#", "Short Sell", "Source", "", "", "Fees");
+        add_action (store, "114073", "Short Buy", "Proceeds", "", "Capgains", "Fees");
+        add_action (store, "004103", "Compensatory Dividend", "Proceeds", "Dividend", "", "Fees");
+        add_action (store, "040100", "Compensatory Notional Distribution", "", "Notional Distribution", "", "");
+        add_action (store, "014000", "Compensatory Return of Capital", "Proceeds", "", "", "");
+        add_action (store, "500000", "Stock Split", "", "", "", "");
```
Bob-IT, line 549: Does this need translating?

gnucash/gnome/dialog-stock-editor.c
```diff
@@ -0,0 +1,689 @@
...
+    }
+    else
+    {
+        add_action (store, "11400#", "Open Long", "Source", "", "", "Fees");
+        add_action (store, "44100#", "Open Short", "Source", "", "", "Fees");
```
Bob-IT, line 554: Does this need translating?

gnucash/gnome/dialog-stock-editor.c
```diff
@@ -0,0 +1,689 @@
...
+
+/********************************************************************   \
+ * stockeditorWindow                                                *
+ *   opens up the window to stock-editor                            *
+ *                                                                  *
+ * Args:   parent  - the parent of this window                      *
+ *         account - the account to stock-edit                      *
+ * Return: stockeditorData - the instance of this StockEditorWindow *
+\********************************************************************/
+StockEditorWindow * stockEditorWindow (GtkWidget *parent, Account *account)
```
Bob-IT, line 577: See other comment in .h file.

gnucash/gnome/dialog-stock-editor.c
```diff
@@ -0,0 +1,689 @@
...
+{
+    GtkBox *box;
+    GtkBuilder *builder;
+    GList *types;
+    StockEditorWindow *data;
+
+    if (!xaccAccountIsPriced (account))
+    {
+        gnc_info_dialog (GTK_WINDOW (parent), "Stock Editor for Stock accounts only");
```
Bob-IT, line 586: Does this need translating?

Bob-IT:
I had a quick look at this with the following...
Any reason why you did not use the dialog response callback / setup?
It took me a while to find an account register that worked, currently you get no indication that it needs to be stock account, maybe only enable the menu option for the right account type.
Obviously I have no idea how to use it but glade wise you could add 'gnc-class-title' to the title label, I do not think the image does any thing so remove it. The other thing you could do is maybe add the memo as a separate line so it is like...
test

That way you see more of the account name and memo text, this needs some refining as it was just a quick move but you get the idea.
Also are you going to save and restore the window size?

Just some quick thoughts.

gnucash/gnome/dialog-stock-editor.c
jralls: No, it needs redesigning. It's about as wrong as it's possible to be.

christopherlam:
> Any reason why you did not use the dialog response callback / setup?

No idea what it is :)

> It took me a while to find an account register that worked, currently you get no indication that it needs to be stock account, maybe only enable the menu option for the right account type. Obviously I have no idea how to use it but glade wise you could add 'gnc-class-title' to the title label, I do not think the image does any thing so remove it. The other thing you could do is maybe add the memo as a separate line so it is like...
> test
> 
> That way you see more of the account name and memo text, this needs some refining as it was just a quick move but you get the idea.

Done; IIUC for complex stock splits, memos would generally be short eg "Cap gains" whereas desc would have more useful info.

> Also are you going to save and restore the window size?

Maybe when it's all done. This dialog isn't very nicely responsive for now.

> 
> Just some quick thoughts.

gnucash/gnome/dialog-stock-editor.h
christopherlam: Amended. A future possible extension will be to launch the dialog prepopulated with current stock txn details. Not sure how safe it'll be to recreate/amend existing txn to have new data. The *safest* approach would be to destroy the previous txn and recreate.

fellen:
> it is like...
> test

I would prefer Account on its own line as the longest path in german account templates is "Aufwendungen 2/4:Reparatur/Instandhaltung:4805 Reparatur u. Instandh. von Anlagen/Maschinen u. Betriebs- u. Geschäftsausst."

christopherlam:
Merged #763 into this branch -- the stock-editor and ACB-tool are complementary.

christopherlam:
@jralls if the stock transaction tool is merged in, what do you think is the appropriate/forward compatibility view on choosing a transaction currency? Should it be stock's parent's currency (most backward compatible) or should it be the root account currency (most likely to be the home currency)

I plan to augment the capital gains split so that it may be derived from the stock unit price on sale, or the unit price is displayed if the capital gain value is supplied.

fellen:
CDB-MAN mentioned recently on [IRC](https://lists.gnucash.org/logs/2020/11/18.html#T08:51:42), "ACB" is canadian terminology and suggested the more neutral "average cost tool".

gnucash/report/reports/standard/acb-tool.scm
```diff
@@ -0,0 +1,402 @@
...
+paid on purchase.")
+   (gnc:html-markup-p "This report is not appropriate for FIFO, LIFO, or
+specific-identification methods for cost basis accumulation and
+capital gain/loss reporting.  This report may not be appropriate for
+tax purposes, if the taxing jurisdiction requires a method other than
+the weighted average cost basis method.")
+   (gnc:html-markup-p "This report is not designed with options
+reporting in mind.  If your activity involves options and/or futures
+that are purchased, written, and/or exercised, there is no guarantee
+that this report will accurately portray this options activity.")))
```
fellen, line 49: No translation?

gnucash/report/reports/standard/acb-tool.scm
```diff
@@ -0,0 +1,402 @@
...
+(use-modules (gnucash core-utils))
+(use-modules (gnucash app-utils))
+(use-modules (gnucash engine))
+
+(define disclaimer
+  (gnc:make-html-text
+   (gnc:html-markup-p "This report is designed for cost basis
+accumulation and capital gain/loss reporting using the weighted
+average cost basis method, which is most consistent with typical
+accounting frameworks (US GAAP, IFRS, etc.).  This report allows for
```
fellen, line 37: @jralls Do we still want runoff style sentence ends (dot + 2 spaces)? I would prefer to unify sentence ends - xgettext offers a check. I would prefer  (dot + 1 space), but am open for other opinions.

jralls: [No.](https://www.theatlantic.com/technology/archive/2011/01/why-you-should-never-ever-use-two-spaces-between-sentences/69579/)

jralls:
@christopherlam Only testing will tell, but I think it's likely that making the transaction currency anything other than the stock-account's inherited currency will break assumptions in the rest of GnuCash, including most crucially online price retrieval.

> I plan to augment the capital gains split so that it may be derived from the stock unit price on sale, or the unit price is displayed if the capital gain value is supplied.

Why would that require augmenting the split? If you know the basis and the number of shares then the net unit price is easily derived from the net proceeds or gain and vice-versa.

christopherlam:
> @christopherlam Only testing will tell, but I think it's likely that making the transaction currency anything other than the stock-account's inherited currency will break assumptions in the rest of GnuCash, including most crucially online price retrieval.

I agree best use parent currency.

> Why would that require augmenting the split? If you know the basis and the number of shares then the net unit price is easily derived from the net proceeds or gain and vice-versa.

I meant augment the gtk controls to display sale price.

jralls:
How, beyond what they already do?
Screen Shot 2020-11-19 at 8 59 30 PM

gnucash/report/reports/standard/acb-tool.scm
christopherlam: This was a direct copy/paste from https://bugs.gnucash.org/show_bug.cgi?id=797796#c126 😏

gnucash/gnome/dialog-stock-editor.c
christopherlam: @jralls would you mind writing a template bitwise mask in `initialize_actions` and I'll convert other actions? e.g. a regular BUY.

christopherlam:
> How, beyond what they already do?
> Screen Shot 2020-11-19 at 8 59 30 PM

Hmm exactly the same as the register, with the added advantage that price (basis/amount) will updated underneath the basis field during `refresh_all` to get immediate feedback -- see `update_price`.

gnucash/gnome/dialog-stock-editor.c
jralls: @christopherlam no. The problem with this PR isn't that you don't know how to do bit masks, the problem is that you've decomposed the problem the wrong way. Giving you a plug-and-chug bit mask solution will neither fix the wrong decomposition nor teach you basic CS.

christopherlam:
@jralls is this any better?

gnucash/gnome/dialog-stock-editor.c
```diff
@@ -0,0 +1,927 @@
...
+    gnc_account_set_linked_account (data->asset_account, "fees_account", fees_acc);
+
+    xaccTransBeginEdit (txn);
+    xaccTransSetCurrency (txn, data->trans_currency);
+    xaccTransSetDatePostedSecsNormalized (txn, date);
+    xaccTransSetDescription (txn, desc);
+
+    create_split (txn, data->asset_account, data->asset_memo, data->asset_amount, data->asset_basis, FALSE);
+    create_split (txn, proceeds_acc, data->proceeds_memo, data->proceeds_val, data->proceeds_val, FALSE);
+    create_split (txn, dividend_acc, data->dividend_memo, data->dividend_val, data->dividend_val, FALSE);
```
jralls, line 545: Creating a dividend split should only happen if there's a dividend.

gnucash/gnome/dialog-stock-editor.c
```diff
@@ -0,0 +1,927 @@
...
+}
+
+static void
+cancel_button_cb (GtkWidget *widget, StockEditorWindow *data)
+{
+    stock_editor_destroy (data);
+}
+
+static gboolean
+gae_unfocus (GtkWidget *widget, const StockEditorWindow *data)
```
jralls, line 144: 'gae' and 'gas' are a bit terse, how about e.g. `account_edit_unfocus` and `account_sel_get_account`?

gnucash/gnome/dialog-stock-editor.c
```diff
@@ -0,0 +1,927 @@
...
+          (gnc_numeric_zero_p (num)     && (mask & MASK_ZERO)) ||
+          (gnc_numeric_negative_p (num) && (mask & MASK_NEGATIVE))))
+    {
+        *passes = FALSE;
+        *status = g_list_prepend (*status, g_strdup_printf (_("%s must be %s"),
+                                                            type, sign_str));
+    }
+
+    gtk_entry_set_placeholder_text (GTK_ENTRY (gae),
+                                    (mask == MASK_DISABLED) ? NULL : sign_str);
```
jralls, line 247: If `!mask` then you don't want to do all of the work, so check for that at the beginning of the function, set the placeholder text to NULL, and return.

gnucash/gnome/dialog-stock-editor.c
```diff
@@ -0,0 +1,927 @@
...
+    num = gae_get_amount (gae);
+
+    if (mask & MASK_POSITIVE)
+        signs = g_list_prepend (signs, g_strdup ("positive"));
+    if (mask & MASK_NEGATIVE)
+        signs = g_list_prepend (signs, g_strdup ("negative"));
+    if (mask & MASK_ZERO)
+        signs = g_list_prepend (signs, g_strdup ("zero"));
+
+    sign_str = string_join (signs, "/");
```
jralls, line 234: No need to g_strdup the strings, strcat can copy `static char*` into the buffer, but you need to translate them, they're displayed to users. Capitalized ("Positive") looks nicer in English. You'll just `g_list_free(signs)` instead of `g_list_free_full`.

You might also create a big static buffer here, 
```
static char* sign_str[256];
memset (sign_str, 0, sizeof(sign_str));
```
and pass it to `string_join` instead of allocating an just-right buffer on the heap.

jralls (changes requested):
Quite a few of the comments will become N/A when you rewrite this to use separate GtkAssistant pages for each transaction type instead of trying to cram every possibility into a single dialog. Not only will that make for a better UI it will also make for more understandable and maintainable code.

gnucash/gnome/dialog-stock-editor.c
christopherlam: This is handled by `gas_get_account` which checks the dividend account selector sensitivity, returning NULL if there's no dividend.

gnucash/gnome/dialog-stock-editor.c
```diff
@@ -0,0 +1,927 @@
...
+        if (n->next)
+            strcat (outstr, sep);
+    }
+
+    return outstr;
+}
+
+static void
+check_acct (GtkWidget *gas, GNCAmountEdit *gae,
+            gboolean *passes, GList **status, gchar *type)
```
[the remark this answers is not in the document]
christopherlam: Can do, however, you logic here[1] would suggest abort if the returned `passes` is FALSE. Instead I wished to collate all error status messages before presenting in the dialog.

[1]
```c
if (!check_acct (dividend_acc, dividend_val, &status, dividend_type) ||
    !check_acct (proceeds_acc, proceeds_val, &status, proceeds_type) ||
    ...
```

gnucash/gnome/dialog-stock-editor.c
```diff
@@ -0,0 +1,927 @@
...
+static void
+update_price (gnc_numeric amount, gnc_numeric basis, gnc_commodity *comm,
+              const StockEditorWindow *data, GNCPrintAmountInfo printinfo)
+{
+    const gchar *label = "";
+    if (!gnc_numeric_zero_p (amount))
+    {
+        gnc_numeric price = gnc_numeric_div
+            (basis, amount, GNC_DENOM_AUTO, GNC_HOW_DENOM_EXACT | GNC_HOW_RND_ROUND);
+        label = xaccPrintAmount (price, (gnc_price_print_info (comm, TRUE)));
```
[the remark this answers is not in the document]
christopherlam: I think it'll be ok -- see https://github.com/GNOME/gtk/blob/2e031e47dde324d0fe3002639e2d53dc15157587/gtk/gtklabel.c#L2922 where the label str is `g_strdup`ed.

gnucash/gnome/dialog-stock-editor.c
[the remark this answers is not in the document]
christopherlam: Addendum I've seen https://github.com/GNOME/gtk/blob/2e031e47dde324d0fe3002639e2d53dc15157587/gtk/gtklabel.c#L3013 which shows how to do it...

```c
fails = check_acct (dividend_acc, dividend_val, &status, dividend_type);
fails = check_acct (proceeds_acc, proceeds_val, &status, proceeds_type) || fails;
fails = check_acct (expenses_acc, expenses_val, &status, expenses_type) || fails;
```

libgnucash/app-utils/gnc-helpers.c
```diff
@@ -133,3 +134,35 @@ gnc_quoteinfo2scm(gnc_commodity *comm)
...
 }
+
+Account * gnc_account_get_linked_account (const Account *acc, const gchar *id)
+{
+    GValue v = G_VALUE_INIT;
+    GncGUID *guid = NULL;
+
+    g_return_val_if_fail (GNC_IS_ACCOUNT (acc), FALSE);
+
+    qof_instance_get_kvp (QOF_INSTANCE(acc), &v, 1, id);
```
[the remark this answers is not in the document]
christopherlam: `qof_instance_get` calls `g_object_get_valist` which calls `g_object_get_is_valid_property` and throws a warning for an account without these slots. I cannot find a suitable API to test that g_object has property. Any ideas?

```
sys:1: Warning: g_object_get_is_valid_property: object class 'Account' has no property named 'proceeds_account'
sys:1: Warning: g_object_get_is_valid_property: object class 'Account' has no property named 'dividend_account'
sys:1: Warning: g_object_get_is_valid_property: object class 'Account' has no property named 'capgains_account'
sys:1: Warning: g_object_get_is_valid_property: object class 'Account' has no property named 'fees_account'
```

gnucash/gnome/dialog-stock-editor.c
christopherlam: > Capitalized ("Positive") looks nicer in English.

But these joined strings e.g. "positive/zero" will be used both as placeholder text, and as the status error message eg "Fees must be positive/zero".

libgnucash/app-utils/gnc-helpers.c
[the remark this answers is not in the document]
christopherlam: fwiw the `qof_object_get/set` (as written in current branch) doesn't seem to work.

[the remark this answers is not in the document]
jralls: > Any ideas?
Yes. Declare the properties by adding them to the [properties enum](https://github.com/Gnucash/gnucash/blob/master/libgnucash/engine/Account.cpp#L85), call `g_object_class_install_property` on each of them in [`gnc_account_class_init`](https://github.com/Gnucash/gnucash/blob/master/libgnucash/engine/Account.cpp#L645), and add getters and setters to [`gnc_account_get_property`](https://github.com/Gnucash/gnucash/blob/master/libgnucash/engine/Account.cpp#L368) and [`gnc_account_set_property`](https://github.com/Gnucash/gnucash/blob/master/libgnucash/engine/Account.cpp#L516) respectively. You don't need to write explicit functions, see the examples of using `qof_instance_get_path` and `qof_instance_set_path` near the bottom of `gnc_account_get/set_property`.

gnucash/gnome/dialog-stock-editor.c
jralls: `positive/zero` is [not-a-number](https://en.wikipedia.org/wiki/NaN). The error should say "positive or zero", or better still "not negative". Not only that but for translators it's better to have "Fees must not be less than zero" than to compose a single `"%s must be %s.", field_name, value_type` to apply to every field. Just because that works in English doesn't mean that it does in other, especially non-European, languages. What's more even if you were to change the separator from '/' to "or" you can't assume that the placeholder text and the sentence text are interchangeable in different languages.

gnucash/gnome/dialog-stock-editor.c
jralls: How's that? Neither `ok_button_cb` nor `create_split` calls `gas_get_account`.

jralls:
1. If you're going to force-push, force push correct commits not a pair where one undoes the first.
2. Don't waste time fixing the comments in dialog-stockeditor.c, keep them in mind as you rewrite it into the GtkAssistant we discussed on IRC.

libgnucash/app-utils/gnc-helpers.c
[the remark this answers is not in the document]
jralls: > fwiw the qof_object_get/set (as written in current branch) doesn't seem to work.

No surprise, there's no such functions. It's `qof_instance_get/set`, and considering how much they're used it would be a major problem if they don't work:
```
grep -rl \\bqof_instance_get\\b *gnucash
gnucash/gnome/dialog-sx-editor.c
gnucash/gnome/gnc-split-reg.c
gnucash/gnome/dialog-sx-editor2.c
gnucash/gnome/gnc-plugin-page-register2.c
gnucash/gnome/dialog-invoice.c
gnucash/gnome/dialog-payment.c
gnucash/gnome/gnc-plugin-page-report.c
gnucash/register/ledger-core/split-register-model.c
gnucash/import-export/test/gtest-import-backend.cpp
gnucash/import-export/ofx/gnc-ofx-import.c
gnucash/import-export/import-utilities.c
gnucash/import-export/aqb/gnc-ab-kvp.c
gnucash/gnome-utils/dialog-preferences.c
gnucash/gnome-utils/gnc-tree-util-split-reg.c
gnucash/gnome-utils/gnc-tree-view-account.c
libgnucash/engine/mocks/gmock-qofinstance.cpp
libgnucash/engine/gncInvoice.c
libgnucash/engine/gncOwner.c
libgnucash/engine/test/test-engine-kvp-properties.c
libgnucash/engine/gnc-budget.c
libgnucash/engine/gnc-lot.c
libgnucash/engine/cap-gains.c
libgnucash/engine/qofinstance.cpp
libgnucash/engine/SX-book.c
libgnucash/engine/qofinstance.h
libgnucash/engine/qofbook.cpp
libgnucash/app-utils/gnc-accounting-period.c
libgnucash/app-utils/gnc-sx-instance-model.c
```

gnucash/gnome/dialog-stock-editor.c
christopherlam: > How's that? Neither `ok_button_cb` nor `create_split` calls `gas_get_account`.

Ok because of prior complaint this function was renamed `account_sel_get_account` 😬.

christopherlam:
Added a couple commits to model the account->linked_account setter/getter, modelled from gncInvoice.c's gncInvoiceAttachToLot / gncInvoiceGetInvoiceFromLot. But this doesn't seem to work.

gnucash/gnome/dialog-stock-editor.c
jralls: That's still not OK because the single-dialog, sensitivity determined by a list selector and behavior determined by various control sensitivity is too clever by 7/8 (that's a lot worse than too clever by half!). It's convoluted, brittle, and will confuse users and future maintainers.

christopherlam:
This is a heavily WIP partial work.

Instead of a crammed grid, this is a GtkAssistant approach. The 1st page is intro, 2nd page choose stock activity, 3rd->penultimate page selects account data (account/amount/value/memo), and last page will confirm.

image

I think this is a much better approach?

christopherlam:
Updated screenshot. Finish page summarises data entry, and lists errors.

image

christopherlam:
Starting:
image

Selecting SELL:
image

Details 1
image

Details 2
image

Details 3
image

details 4
image

details 5
image

details 6
image

summary
image

result
image

christopherlam:
This is currently hidden behind `--extra` and _could_ be merged in for beta testing.

gnucash/gnome/assistant-stock-transaction.cpp
```diff
@@ -0,0 +1,1148 @@
...
+
+static void refresh_page_transaction_type (GtkWidget *widget, gpointer user_data)
+{
+    StockTransactionInfo *info = (StockTransactionInfo *)user_data;
+
+    auto type_idx = gtk_combo_box_get_active (GTK_COMBO_BOX (widget));
+    if (type_idx < 0)           // combo isn't initialized yet.
+        return;
+
+    info->txn_type = info->txn_types.at (type_idx);
```
christopherlam, line 436: This currently leaks:
```
==38937== 14 bytes in 1 blocks are definitely lost in loss record 1,173 of 19,428
==38937==    at 0x4843FB3: operator new(unsigned long) (in /usr/libexec/valgrind/vgpreload_memcheck-amd64-linux.so)
==38937==    by 0x62171CE: std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >::_M_assign(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) (in /usr/lib/x86_64-linux-gnu/libstdc++.so.6.0.29)
==38937==    by 0x621752D: std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >::operator=(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) (in /usr/lib/x86_64-linux-gnu/libstdc++.so.6.0.29)
==38937==    by 0x4D96B62: TxnTypeInfo::operator=(TxnTypeInfo const&) (assistant-stock-transaction.cpp:95)
==38937==    by 0x4D91B36: refresh_page_transaction_type(_GtkWidget*, void*) (assistant-stock-transaction.cpp:436)
==38937==    by 0x5AAA96F: g_signal_emit_valist (in /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0.6800.4)
==38937==    by 0x5AAAAD2: g_signal_emit (in /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0.6800.4)
==38937==    by 0x51F0C06: ??? (in /usr/lib/x86_64-linux-gnu/libgtk-3.so.0.2404.26)
==38937==    by 0x51F0DD2: gtk_combo_box_set_active (in /usr/lib/x86_64-linux-gnu/libgtk-3.so.0.2404.26)
==38937==    by 0x4D9332C: stock_assistant_prepare (assistant-stock-transaction.cpp:701)
==38937==    by 0x5A8B624: g_cclosure_marshal_VOID__OBJECTv (in /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0.6800.4)
==38937==    by 0x5AAA96F: g_signal_emit_valist (in /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0.6800.4)
```

gnucash/gnome/assistant-stock-transaction.cpp
```diff
@@ -0,0 +1,1148 @@
...
+    auto curr_list = g_list_prepend (nullptr, currency);
+    gnc_account_sel_set_new_account_ability (GNC_ACCOUNT_SEL (gas), true);
+    gnc_account_sel_set_acct_filters (GNC_ACCOUNT_SEL (gas), acct_list, curr_list);
+    gtk_widget_show (gas);
+    gtk_grid_attach (GTK_GRID(table), gas, 1, row, 1, 1);
+    gtk_label_set_mnemonic_widget (GTK_LABEL(label), gas);
+    g_list_free (acct_list);
+    g_list_free (curr_list);
+    return gas;
+}
```
christopherlam, line 957: This one leaks `curr_list` however I can't figure out why. It's freed in #L955.
```
==38937== 24 bytes in 1 blocks are definitely lost in loss record 6,103 of 19,428
==38937==    at 0x4843839: malloc (in /usr/libexec/valgrind/vgpreload_memcheck-amd64-linux.so)
==38937==    by 0x4C3C088: g_malloc (in /usr/lib/x86_64-linux-gnu/libglib-2.0.so.0.6800.4)
==38937==    by 0x4C53374: g_slice_alloc (in /usr/lib/x86_64-linux-gnu/libglib-2.0.so.0.6800.4)
==38937==    by 0x4C2DF5A: g_list_copy_deep (in /usr/lib/x86_64-linux-gnu/libglib-2.0.so.0.6800.4)
==38937==    by 0x4F9A334: gnc_account_sel_set_acct_filters (gnc-account-sel.c:388)
==38937==    by 0x4D93E9E: create_gas(_GtkBuilder*, int, std::vector<GNCAccountType, std::allocator<GNCAccountType> >, gnc_commodity_s*, char const*, char const*) (assistant-stock-transaction.cpp:950)
==38937==    by 0x4D94C4E: stock_assistant_create(StockTransactionInfo*) (assistant-stock-transaction.cpp:1087)
==38937==    by 0x4D95030: gnc_stock_transaction_assistant (assistant-stock-transaction.cpp:1139)
==38937==    by 0x4E2C5C3: gnc_plugin_page_register_cmd_stock_assistant (gnc-plugin-page-register.c:4581)
==38937==    by 0x5A8CC0E: g_closure_invoke (in /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0.6800.4)
==38937==    by 0x5AA8EA5: ??? (in /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0.6800.4)
==38937==    by 0x5AAA883: g_signal_emit_valist (in /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0.6800.4)
==38937== 
```

gnucash/gnome/assistant-stock-transaction.cpp
christopherlam: I *tried* redefine `txn_type` and `txn_types` as

```cpp
TxnTypeInfo * txn_type;
TxnTypeVec * txn_types;
```

but doesn't compile. This _may_ require some fancy `shared_ptr`?

jralls:
@christopherlam I can't review a moving target. Let me know when you're done.

That said, your latest commit looks like a very bad idea indeed: non-constexpr statics make code thread-unsafe and can lead to all manner of other problems. The only reason to use ptrs instead of refs is for passing to a C function or you need a cheap way to indicate a "None" object, std::optional being the safer but somewhat painful alternative.

christopherlam:
Ok let's discard the latest commit which aims to plug memory leak. I'll leave this alone from now on.

gnucash/gnome/assistant-stock-transaction.cpp
```diff
@@ -0,0 +1,1151 @@
...
+    xaccSplitSetAccount (split, account);
+    xaccSplitSetMemo (split, gtk_entry_get_text (GTK_ENTRY (memo_entry)));
+    xaccSplitSetValue (split, value_numeric);
+    xaccSplitSetAmount (split, amount_numeric);
+    // g_warning ("creating %s split in account %s with Val(%s), Amt(%s). result Val(%s), Amt(%s)",
+    //            action, xaccAccountGetName (account),
+    //            gnc_numeric_to_string (value_numeric),
+    //            gnc_numeric_to_string (amount_numeric),
+    //            gnc_numeric_to_string (xaccSplitGetValue (split)),
+    //            gnc_numeric_to_string (xaccSplitGetAmount (split)));
```
jralls, line 796: Either delete this or uncomment it and change `g_warning` to `DEBUG`. Always prefer the libgnucash/core-utils/qoflog.h macros to calling GLib logging functions directly.

gnucash/gnome/assistant-stock-transaction.cpp
```diff
@@ -0,0 +1,1151 @@
...
+
+    //     split = xaccMallocSplit (gnc_get_current_book ());
+    //     xaccAccountBeginEdit (account);
+    //     xaccSplitSetAccount (split, account);
+    //     xaccSplitSetParent (split, trans);
+    //     xaccSplitSetAmount (split, gnc_numeric_neg (amount));
+    //     xaccSplitSetValue (split, gnc_numeric_neg (amount));
+    //     xaccSplitSetMemo (split, memo);
+    // }
+
```
jralls, line 888: Remove commented-out code.

gnucash/gnome/assistant-stock-transaction.cpp
jralls: It's not leaking `curr_list`, it's leaking the *copy* of `curr_list` in `gas` because [`gnc_account_sel_finalize`](https://github.com/Gnucash/gnucash/blob/7e299cb2b0f2d586462b8115440b3d17ff2b90dd/gnucash/gnome-utils/gnc-account-sel.c#L394) only frees `acctTypeFilters`.

gnucash/gnome/assistant-stock-transaction.cpp
```diff
@@ -0,0 +1,1148 @@
...
+    gnc_builder_add_from_file  (builder , "assistant-stock-transaction.glade", "stock_transaction_assistant");
+    info->window = get_widget (builder, "stock_transaction_assistant");
+
+    // Set the name for this assistant so it can be easily manipulated with css
+    gtk_widget_set_name (GTK_WIDGET(info->window), "gnc-id-assistant-stock-transaction");
+
+    auto balance = xaccAccountGetBalance (info->acct);
+    info->txn_types = gnc_numeric_zero_p (balance) ? starting_types
+        : gnc_numeric_positive_p (balance) ? long_types
+        : short_types;
```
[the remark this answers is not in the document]
jralls: Well, if you don't want a copy, don't tell it to copy.

Making `txn_types` a `TxnTypeInfoVec&` should prevent the copying and making the `std::string`->`const char*` and `std::vector`->`constexpr std::array` changes I suggested up top should fix a lot of the allocating. You can safely use `std::string_view`s if you need to use `std::string` functionality on those `const char*`s, since they'll be part of the `constexpr std::array<TxnTypeInfo>`s you don't have to worry about them going anywhere.

gnucash/gnome/assistant-stock-transaction.cpp
```diff
@@ -0,0 +1,1151 @@
...
+#include "engine-helpers.h"
+#include "dialog-utils.h"
+#include "assistant-stock-transaction.h"
+#include "gnc-account-sel.h"
+#include "gnc-amount-edit.h"
+#include "gnc-component-manager.h"
+#include "gnc-date-edit.h"
+#include "gnc-tree-view-account.h"
+
+void gnc_stock_transaction_dialog (GtkWidget *parent, Account * account);
```
jralls, line 41: This seems not to be used.

gnucash/gnome/assistant-stock-transaction.cpp
```diff
@@ -0,0 +1,1151 @@
...
+#include "Transaction.h"
+#include "engine-helpers.h"
+#include "dialog-utils.h"
+#include "assistant-stock-transaction.h"
+#include "gnc-account-sel.h"
+#include "gnc-amount-edit.h"
+#include "gnc-component-manager.h"
+#include "gnc-date-edit.h"
+#include "gnc-tree-view-account.h"
+
```
jralls, line 40: Needs a QofLoModule decl e.g.:
```
static QofLogModule log_module = GNC_MOD_ASSISTANT;
```
There are a bunch of GNC_MOD_FOO declared in libgnucash/engine/gnc-engine.h or you can inherit `gnc.gui` from CMakeLists.txt by using `G_LOG_DOMAIN`.

jralls (changes requested):
This still needs quite a bit of work.

gnucash/gnome/assistant-stock-transaction.cpp
```diff
@@ -0,0 +1,1151 @@
...
+    {
+        ENABLED_DEBIT,          // stock_amt
+        ENABLED_DEBIT,          // stock_val
+        ENABLED_CREDIT,         // cash_amt
+        ENABLED_DEBIT | ALLOW_ZERO,          // fees_amt
+        true,                   // fees_capitalize
+        DISABLED,               // dividend_amt
+        DISABLED,               // capg_amt
+        "Open buy",
+        "Initial stock purchase"
```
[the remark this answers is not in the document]
christopherlam: Of course. But I figure *if* this was merged soon for beta-testing hidden behind `--extra` it wouldn't be nice to annoy translators until it's ready to be deployed.

gnucash/gnome/assistant-stock-transaction.cpp
christopherlam: this will be fixed when the whole branch is condensed into 1 commit.

gnucash/gnome/assistant-stock-transaction.cpp
```diff
@@ -0,0 +1,1151 @@
...
+                        SPLIT_COL_ACCOUNT, acctstr,
+                        SPLIT_COL_MEMO, memostr,
+                        SPLIT_COL_DEBIT, debit_side ? amtstr : "",
+                        SPLIT_COL_CREDIT, !debit_side ? amtstr : "",
+                        -1);
+}
+
+static std::string join(const StringVec& v, char c)
+{
+    std::string s;
```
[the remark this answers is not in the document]
christopherlam: Ok, this means scanning the StringVec twice, added change as requested, not convinced it's worth the trouble.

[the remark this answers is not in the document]
christopherlam: btw see the fancy std::accumulate which is the srfi-1 `fold` equivalent...

christopherlam:
done most except
1. can't figure out how to use `std::array&` which requires 2nd parameter size, and starting_types=2, long_types=10, short_types=10 being all different, cannot be assigned to `txn_types`
2. not sure what to replace `uint` with.

gnucash/gnome/assistant-stock-transaction.cpp
```diff
@@ -0,0 +1,1151 @@
...
+    for (StringVec::const_iterator p = v.begin(); p != v.end(); ++p)
+    {
+        s += *p;
+        if (p != v.end() - 1)
+            s += c;
+    }
```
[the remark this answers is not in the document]
christopherlam: This (std::for_each) won't work because its 3rd arg where you've offered the lambda does not get the iterator as an argument. It gets the vector element itself. https://en.cppreference.com/w/cpp/algorithm/for_each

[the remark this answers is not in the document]
christopherlam: > This (std::for_each) won't work because its 3rd arg where you've offered the lambda does not get the iterator as an argument. It gets the vector element itself. https://en.cppreference.com/w/cpp/algorithm/for_each

Having said that^, this function `join` amended to use std::for_each which doesn't need the iterator.

gnucash/gnome/assistant-stock-transaction.cpp
```diff
@@ -0,0 +1,1150 @@
...
+static std::string
+join (std::string s, const StringVec& v, char c)
+{
+    auto add_len = [](int a, std::string b) { return a + 1 + b.size (); };
+    auto add_separator_and_string = [&](auto str) { s += c; s += str; };
+    auto length = std::accumulate (v.begin (), v.end (), 0, add_len);
+    s.reserve (length + s.size ());
+    std::for_each (v.begin(), v.end(), add_separator_and_string);
+    return s;
+}
```
gjanssens, line 610: As we're juggling algorithms, you could try [std::accumulate](https://en.cppreference.com/w/cpp/algorithm/accumulate) with a custom binary operator instead. That conveys the intent of the action even more. There's an example on that page. Freely translated to your use case:

```
auto newline_fold = [](std::string a, std::string b) {
                         return std::move(a) + '\n' + std::move(b);
                     };
 
     s = std::accumulate(std::next(v.begin()), v.end(),
                                    v[0], // start with first element
                                    newline_fold);
```

This ignores the pre-allocation of the result string `s` though. But that may well be a premature optimization. The strings to accumulate typically aren't that long and there may be optimisations already integrated in the accumulate code for strings in the stl itself. Only profiling will tell.

So for initial code I'd aim for simplicity here and just replace the whole `join`function (which appears to be used only once around line 696) by a single call to std::accumulate and embed the lambda function directly into the call, like so:

```
gtk_label_set_text (GTK_LABEL (info->finish_summary),
                    std::accumulate (errors.begin(), errors.end(),
                                     _("The following errors must be fixed:"),
                                     [](std::string a, std::string b) {
                                         return std::move(a) + '\n' + std::move(b);).c_str());
```

If that's too long, you can still declare that lambda just above the call to gtk_label_set_text and use it's variable name instead. If you need this in more places you can define the operator at the top-level as a function and use that function name instead.

jralls:
> can't figure out how to use std::array& which requires 2nd parameter size, and starting_types=2, long_types=10, short_types=10 being all different, cannot be assigned to txn_types

Actually long_types and short_types are the same and serve the same purpose, starting_types is the odd one out. The Modern C++ Way™ would be to use a std::variant, but that can get a bit complicated. You could use the same size for all three, but then you wouldn't be able to use `end()` for the terminating iterator. How about a little class:
```
using HoldingTypes = std::array<TxnType, 10>;
using StartingTypes = std::array<TxnType, 2>;
class TxnTypes
{
    const bool m_starting;
    const HoldingTypes* m_holding;
    const StartingTypes* m_starting;
public:
    TxnTypes() = deleted;
    TxnTypes(const HoldingTypes* holding) : m_starting{false}, m_holding{holding}, m_starting{nullptr};
    TxnTypes(const StartingTypes* starting) : m_starting{true}, m_holding{nullptr}, m_starting{starting};
    TxnTypes{const TxnTypes&) = default;
    TxnTypes(TxnTypes&&) = delete;
    operator=(const TxnTypes&) = default;
    operator=(TxnTypes&&) = delete;
    ~TxnTypes() {};

   const_iterator begin() { return m_starting ? m_starting->begin() : m_holding->begin() }
   const_iterator end() { return m_starting ? m_starting->end() : m_hoding->end() }
   // etc for whatever other std::array functions you need
}
```
That's a run-time version of what std::variant does with a much lower learning curve.

> not sure what to replace uint with.

The type that you're actually using, of course: `FieldMask`.

gnucash/gnome/assistant-stock-transaction.cpp
jralls: 👍

gnucash/gnome/assistant-stock-transaction.cpp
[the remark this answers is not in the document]
jralls: Uh, no. If it's going to the public it needs to have its public-facing strings marked for translation and we hope at least partially translated by the release. Hiding it behind --extra doesn't change that. Plus it's way too easy to forget that the strings all need to be marked when it comes time to remove `needs_extra_enabled`.

gnucash/gnome/assistant-stock-transaction.cpp
```diff
@@ -0,0 +1,1150 @@
...
+    {
+        ENABLED_DEBIT,          // stock_amt
+        ENABLED_DEBIT,          // stock_val
+        ENABLED_CREDIT,         // cash_amt
+        ENABLED_DEBIT | ALLOW_ZERO,          // fees_amt
+        true,                   // fees_capitalize
+        DISABLED,               // dividend_amt
+        DISABLED,               // capg_amt
+        _("Open buy"),
+        _("Initial stock purchase")
```
jralls, line 114: Static variables are created at program initialization before `main` so there's no locale or gettext text_domain. That means that these strings won't get translated. The fix is to use the `N_(x)` macro to mark them for xgettext and then call `gettext` (via the `_(x)` macro) when you pass them to the GtkWidget for display.

Short strings tend to be a bit cryptic in the isolation of the po file and the Weblate interface so it's helpful to translators to provide context. For that you use `NC_(x)` and `C_(x)`, and the context string goes first, e.g.:
```
NC_("Stock assistant open a position with a buy", "Open buy"),
```
and
```
          gtk_combo_box_text_append_text (GTK_COMBO_BOX_TEXT (info->transaction_type_combo),
                                             C_(it.friendly_name));
```

gnucash/gnome/assistant-stock-transaction.cpp
christopherlam: > So for initial code I'd aim for simplicity here and just replace the whole `join`function (which appears to be used only once around line 696) by a single call to std::accumulate and embed the lambda function directly into the call, like so:
> 
> ```
> gtk_label_set_text (GTK_LABEL (info->finish_summary),
>                     std::accumulate (errors.begin(), errors.end(),
>                                      _("The following errors must be fixed:"),
>                                      [](std::string a, std::string b) {
>                                          return std::move(a) + '\n' + std::move(b);).c_str());
> ```
> 
> If that's too long, you can still declare that lambda just above the call to gtk_label_set_text and use it's variable name instead. If you need this in more places you can define the operator at the top-level as a function and use that function name instead.

Ok. However std::accumulate 'inherits' the seed accumulator type so it can't be a char*:
`std::accumulatr (v.begin(), v.end(), std::string { seed }, newline_fold);`

christopherlam:
> > can't figure out how to use std::array& which requires 2nd parameter size, and starting_types=2, long_types=10, short_types=10 being all different, cannot be assigned to txn_types

(snip) I think best stick with pointer because `txn_types` and `txn_type` must be allowed to be uninitialized at the start of GtkAssistant; the `txn_types` will be initialized when the account balance is determined, and `txn_type` will be set when the user makes a choice.

> > not sure what to replace uint with.
> 
> The type that you're actually using, of course: `FieldMask`.

This doesn't work for `fees_amt` which can be initialized to `ENABLED_DEBIT | ALLOW_ZERO` which is not one of the FieldMask members.

gnucash/gnome/assistant-stock-transaction.cpp
christopherlam: > It's not leaking `curr_list`, it's leaking the _copy_ of `curr_list` in `gas` because [`gnc_account_sel_finalize`](https://github.com/Gnucash/gnucash/blob/7e299cb2b0f2d586462b8115440b3d17ff2b90dd/gnucash/gnome-utils/gnc-account-sel.c#L394) only frees `acctTypeFilters`.

This was a _sneaky_ leak...

jralls:
> This doesn't work for fees_amt which can be initialized to ENABLED_DEBIT | ALLOW_ZERO which is not one of the FieldMask members.

You need to provide `FieldMask operator|(FieldMask, FieldMask)`. See https://wiggling-bits.net/using-enum-classes-as-type-safe-bitmasks/ and https://accu.org/journals/overload/24/132/williams_2228/. Since you've got only the one enum type you needn't worry about the template stuff.

jralls:
> I think best stick with pointer because txn_types and txn_type must be allowed to be uninitialized at the start of GtkAssistant; the txn_types will be initialized when the account balance is determined, and txn_type will be set when the user makes a choice.

That's fine, but it's orthogonal to using the two std::array types and the combining class or std::variant. OTOH Geert's right about over-optimizing something that doesn't really need it; with the std::string memory issue fixed the existing std::vector approach is harmless if not necessarily best practice.

christopherlam:
> > I think best stick with pointer because txn_types and txn_type must be allowed to be uninitialized at the start of GtkAssistant; the txn_types will be initialized when the account balance is determined, and txn_type will be set when the user makes a choice.
> 
> That's fine, but it's orthogonal to using the two std::array types and the combining class or std::variant. OTOH Geert's right about over-optimizing something that doesn't really need it; with the std::string memory issue fixed the existing std::vector approach is harmless if not necessarily best practice.

👍 Hence this branch could be merged soon (in maint?) to start beta testing the mechanics.

The N_ vs _ issue is new to me. It makes translating the error messages more difficult. Does the _ apply inside or outside the std::move within the std:: accumulate ?

jralls:
> Hence this branch could be merged soon (in maint?) to start beta testing the mechanics.

If you can get the translations fixed before the string freeze this weekend you can merge it.

gnucash/gnome/assistant-stock-transaction.cpp
christopherlam: > Short strings tend to be a bit cryptic in the isolation of the po file and the Weblate interface so it's helpful to translators to provide context. For that you use `NC_(x)` and `C_(x)`, and the context string goes first, e.g.:

Doesn't quite work like this. Both `NC_` and `C_` require both context & string -- https://people.gnome.org/~ryanl/glib-docs/glib-I18N.html -- Perhaps it's best use `Translators:` comment to provide explanation.

jralls: Translator comments are a poor substitute because two instance of the same short English phrase sometimes has different translations in some language. While xgettext will copy `Translators` comments into gnucash.pot next to the `msgid`s without using context it will combine the same phrase into a single `msgid` so that the translator can provide only a single translation.

christopherlam: I'm still confused about NC_ and C_ -- both require context and translatable string. I guess for this file the context will be "StockAssistant" throughout this file? i.e. `NC("StockAssistant", "Cover Buy")`, and `C_("StockAssistant", it.friendly_name)`?

christopherlam: From https://people.gnome.org/~ryanl/glib-docs/glib-I18N.html#C-:CAPS using `C_` requires 2 args to be string _literals_. Maybe https://docs.gtk.org/glib/func.dpgettext2.html is the right one to use?

jralls: We're mostly using gettext directly so you should consult it's documentaion; for this purpose that's [Using Contexts to Solve Ambiguities](https://www.gnu.org/software/gettext/manual/gettext.html#Contexts) and [Special Cases](https://www.gnu.org/software/gettext/manual/gettext.html#Special-cases). Note at the bottom of the Contexts section there's `pgettext_expr` that takes non-literal arguments.

Yes, you must repeat the context. `NC_` marks the string pair for xgettext to extract and returns only the second string, the msgid, so you have to repeat the context in calling `C_`.

Of course we can simplify that with our own set of macros (untested, you might need to play with it a bit):
```
struct context_msgid
{
    const char* context;
    const char* msgid;
};

#undef NC_
#undef CN_

#define NC_(c, m) \
  context_msgid(c, m)

#define CN_(m) \
   pgettext_expr (m.context, m.msgid)
```

We might want that in libgnucash/core-utils/gnc-locale-utils.h. The only minor issue is that the header declaring `pgettext_expr` is installed at $PREFIX/share/gettext/gettext.h, a rather strange place that will require adding a `-I` compiler argument to the relevant CMakeLists.txt.

jralls:
Almost there!

Add `#include <glib/gi18n.h>` to gnc-locale-utils.h and remove it from gnc-locale-utils.c to fix the build failure and squash 3327384 into 56e9ef5.

jralls:
Don't you test build before pushing? You misspelt `gi18n.h`.

christopherlam:
I've added two sample commits to try debug C_.

```
[po/fr.po] sample fr.po]
[sample pgettext function]
```

The trace output is:

```
[Stock Assistant Action|Buy]=>[Buy]
[Stock Assistant Action|Sell]=>[Sell]
[Stock Assistant Action|Dividend]=>[Dividend]
[Stock Assistant Action|Dividend reinvestment (w/ remainder)]=>[Reinvestir les dividendes]
[Stock Assistant Action|Dividend reinvestment (w/o remainder)]=>[Dividend reinvestment (w/o remainder)]
[Stock Assistant Action|Return of Capital]=>[Return of Capital]
[Stock Assistant Action|Notional distribution]=>[Notional distribution]
[Stock Assistant Action|Stock split]=>[Stock split]
[Stock Assistant Action|Reverse split]=>[Reverse split]
[Stock Assistant Action|Reverse split w/ cash in lieu for fractionals]=>[Reverse split w/ cash in lieu for fractionals]
```

In this fr.po there's translation for `[Stock Assistant Action|Buy]`, `[Stock Assistant Action|Sell]` and `[Stock Assistant Action|Dividend reinvestment (w/ remainder)]` but only the last one is translated in UI.

jralls:
It's working perfectly, you just need to remove the fuzzy tags from the translations that have them:
```
[Stock Assistant Action|Buy]=>[Acheter des fonds]
[Stock Assistant Action|Sell]=>[Vendre des stocks]
[Stock Assistant Action|Dividend]=>[Dividendes]
[Stock Assistant Action|Dividend reinvestment (w/ remainder)]=>[Reinvestir les dividendes]
[Stock Assistant Action|Dividend reinvestment (w/o remainder)]=>[Dividend reinvestment (w/o remainder)]
[Stock Assistant Action|Return of Capital]=>[Return of Capital]
[Stock Assistant Action|Notional distribution]=>[Notional distribution]
[Stock Assistant Action|Stock split]=>[Division d'actions]
[Stock Assistant Action|Reverse split]=>[_Vérifier les répartitions pointées]
[Stock Assistant Action|Reverse split w/ cash in lieu for fractionals]=>[Reverse split w/ cash in lieu for fractionals]
```

jralls:
Don't forget to remove that sample fr.po before you merge this!

fellen:
> In this fr.po there's translation for `[Stock Assistant Action|Buy]`, `[Stock Assistant Action|Sell]` and `[Stock Assistant Action|Dividend reinvestment (w/ remainder)]` but only the last one is translated in UI.

That is not,what we want. You are reintroducing [Bug 797349 - "A"ssociate header badly translated](https://bugs.gnucash.org/show_bug.cgi?id=797349).

christopherlam:
> > In this fr.po there's translation for `[Stock Assistant Action|Buy]`, `[Stock Assistant Action|Sell]` and `[Stock Assistant Action|Dividend reinvestment (w/ remainder)]` but only the last one is translated in UI.
> 
> That is not,what we want. You are reintroducing [Bug 797349 - "A"ssociate header badly translated](https://bugs.gnucash.org/show_bug.cgi?id=797349).

No @fellen check the code and fr.po - the contexts are separated properly.

fellen:
> > > In this fr.po there's translation for `[Stock Assistant Action|Buy]`, `[Stock Assistant Action|Sell]` and `[Stock Assistant Action|Dividend reinvestment (w/ remainder)]` but only the last one is translated in UI.
> > 
> > 
> > That is not,what we want. You are reintroducing [Bug 797349 - "A"ssociate header badly translated](https://bugs.gnucash.org/show_bug.cgi?id=797349).
> 
> No @fellen check the code and fr.po - the contexts are separated properly.
I still do not have much time, My alarm ringed on the obsolete form in your comment, but your fr.po looks OK.

jralls:
fr.po is here just for demonstration purposes. It must be removed before the PR is merged and then the new strings will be available on Weblate for translation.  That's not to suggest that @christopherlam's French is in any way suspect, just that we want all translation work to flow through Weblate.

gnucash/gtkbuilder/assistant-stock-transaction.glade
```diff
@@ -0,0 +1,941 @@
...
+    <signal name="close" handler="stock_assistant_finish" swapped="no"/>
+    <signal name="prepare" handler="stock_assistant_prepare" swapped="no"/>
+    <child>
+      <object class="GtkLabel" id="intro_page_label">
+        <property name="visible">True</property>
+        <property name="can-focus">False</property>
+        <property name="valign">start</property>
+        <property name="margin-start">12</property>
+        <property name="margin-end">12</property>
+        <property name="label" translatable="yes">This assistant will help you record a stock transaction. The transaction type (purchase, sale, dividend, distribution, return of capital, stock split) will determine the transaction splits involved in the transaction.</property>
```
fellen, line 22: Is the meaning of `distribution`somewhere explained, perhaps in the glossary?

fellen: Another, more fundamental question: Is it only applicable for shares? Some other commodities like funds have similar tramsactions.

gnucash/gtkbuilder/assistant-stock-transaction.glade
```diff
@@ -0,0 +1,941 @@
...
+        <property name="can-focus">False</property>
+        <property name="margin-start">12</property>
+        <property name="border-width">6</property>
+        <property name="orientation">vertical</property>
+        <child>
+          <object class="GtkLabel" id="transaction_type_page_label">
+            <property name="visible">True</property>
+            <property name="can-focus">False</property>
+            <property name="margin-bottom">6</property>
+            <property name="label" translatable="yes">Select the type of stock activity that you wish to record. The available types depend on whether the current stock balance is positive, nil, or negative (i.e. when shorting stock). The type will determine the component splits.</property>
```
fellen, line 43: `(i.e. when shorting stock)`: while I know `short selling`  neither google translate nor me knows this term.

gnucash/gtkbuilder/assistant-stock-transaction.glade
```diff
@@ -0,0 +1,941 @@
...
+          </object>
+          <packing>
+            <property name="expand">False</property>
+            <property name="fill">False</property>
+            <property name="position">1</property>
+          </packing>
+        </child>
+      </object>
+      <packing>
+        <property name="title" translatable="yes" comments="Dialog title for the remains of a stock split">Cash</property>
```
fellen, line 521: At other places we use the term `cash in lieu`

gnucash/gtkbuilder/assistant-stock-transaction.glade
```diff
@@ -0,0 +1,941 @@
...
+        <property name="can-focus">False</property>
+        <property name="border-width">6</property>
+        <property name="orientation">vertical</property>
+        <property name="spacing">12</property>
+        <child>
+          <object class="GtkLabel" id="stock_fees_title">
+            <property name="visible">True</property>
+            <property name="can-focus">False</property>
+            <property name="margin-bottom">6</property>
+            <property name="label" translatable="yes">In this page, input any broker fees incurred in this transaction. The fee may be capitalized into the asset account, or expensed into a broker fees account. Typically fees on purchase are capitalized, and fees on sales are expensed. If there are no fees involved in this transaction, it can be left blank.</property>
```
fellen, line 537: `Typically fees on purchase are capitalized, and fees on sales are expensed.`
Doesn't this depend on the local tax law?

gnucash/gtkbuilder/assistant-stock-transaction.glade
jralls: "Shorting a stock" is a slang-ish way of saying "Selling short a stock", that is borrowing the stock from someone and selling it with the intent to buy it back at a lower price and to redeem the loan.

gnucash/gtkbuilder/assistant-stock-transaction.glade
jralls: A "distribution" in this meaning is a transfer of a company's retained wealth to the shareholders/partners. Dividends are one form of distribution used by regular companies and regulated investment companies (often known colloquially as open-end mutual funds). Other corporate structures, e.g. master limited partnerships, may call their distributions something else and they may have different tax implications.

gnucash/gtkbuilder/assistant-stock-transaction.glade
jralls: It does, though mostly as a matter of timing, particularly in jurisdictions where the capital gains and ordinary income tax rates are the same. By capitalizing the commissions and fees at purchase the government defers the deduction of them until the time of sale; for sale commissions and fees it doesn't matter whether the proceeds are net of commissions and fees or deducted afterwards.

FWIW in the US one reports both the basis and proceeds as net of commissions and fees, so not the "typical" arrangement. In the US it's become largely academic: There's no commission (brokers are making their money on order flow) and there's a flat $5.00 fee from the Securities and Exchange Commission on sales.

gnucash/gtkbuilder/assistant-stock-transaction.glade
```diff
@@ -0,0 +1,941 @@
...
+                <property name="left-attach">0</property>
+                <property name="top-attach">0</property>
+              </packing>
+            </child>
+            <child>
+              <object class="GtkLabel" id="prev_balance_amount">
+                <property name="visible">True</property>
+                <property name="can-focus">False</property>
+                <property name="halign">end</property>
+                <property name="label" translatable="yes">(Previous Balance)</property>
```
fellen, line 250: Is there no better solution here and later than parenthesis, perhaps some markup?

christopherlam: This label is overwritten during code. So translation must be disabled. Thanks.

jralls: Then you should change the `translatable` attribute value to "no" and add a comment `<!-- label will be overwritten in code -->`

christopherlam: > Then you should change the `translatable` attribute value to "no" and add a comment `<!-- label will be overwritten in code -->`

In my experiment with `ninja pot` removing the `translatable` attribute was sufficient to avoid an entry in gnucash.pot.

jralls: Of course. But it doesn't communicate to our future selves that the string is untranslated on purpose.

christopherlam:
saw this: https://ossia.io/posts/minimum-viable/ instead of gtk :)

fellen:
> saw this: https://ossia.io/posts/minimum-viable/ instead of gtk :)

If you search the history, cstim experimented in [CuteCash](https://wiki.gnucash.org/wiki/Cutecash) with QT. 

The problem whti switching the UI: There is still too much knowledge in the UI instead of the data model.

gnucash/gtkbuilder/assistant-stock-transaction.glade
```diff
@@ -0,0 +1,941 @@
...
+        <property name="can-focus">False</property>
+        <property name="valign">start</property>
+        <property name="margin-start">12</property>
+        <property name="margin-end">12</property>
+        <property name="label" translatable="yes">This assistant will help you record a stock transaction. The transaction type (purchase, sale, dividend, distribution, return of capital, stock split) will determine the transaction splits involved in the transaction.</property>
+        <property name="wrap">True</property>
+      </object>
+      <packing>
+        <property name="page-type">intro</property>
+        <property name="title" translatable="yes">Stock Transaction Assistant</property>
```
fellen, line 27: Is there any reason, why this  is not `Stock Split Assistant` like all other occurrences?

christopherlam: Because this tool is much more capable: stock split, stock purchase, stock sale, dividend, notional distribution.

PR #820, opened 2020-11-16
Activate feature GNC_FEATURE_EQUITY_TYPE_OPENING_BALANCE

rhabacker:
This commit should be applied after the release of gnucash version 4.3.

### DIFF
--- a/gnucash/gnome/gnc-plugin-page-register.c
+++ b/gnucash/gnome/gnc-plugin-page-register.c
@@ -742,6 +742,10 @@ gnc_plugin_page_register_new_common (GNCLedgerDisplay* ledger)
     if (!gnc_features_check_used (gnc_get_current_book(), GNC_FEATURE_REG_SORT_FILTER))
         gnc_features_set_used (gnc_get_current_book(), GNC_FEATURE_REG_SORT_FILTER);
 
+    // added for version 4.5 onwards
+    if (!gnc_using_equity_type_opening_balance_account (gnc_get_current_book()))
+        gnc_set_use_equity_type_opening_balance_account (gnc_get_current_book());
+
     /* Is there an existing page? */
     gsr = gnc_ledger_display_get_user_data (ledger);
     if (gsr)
--- a/gnucash/gnome/assistant-hierarchy.c
+++ b/gnucash/gnome/assistant-hierarchy.c
@@ -1424,6 +1424,9 @@ on_finish (GtkAssistant  *gtkassistant,
     ENTER (" ");
     com = gnc_currency_edit_get_currency (GNC_CURRENCY_EDIT(data->currency_selector));
 
+    if (!gnc_using_equity_type_opening_balance_account (gnc_get_current_book()))
+        gnc_set_use_equity_type_opening_balance_account (gnc_get_current_book());
+
     if (data->our_account_tree)
     {
         gnc_account_foreach_descendant (data->our_account_tree,


fellen:
I assume the build fails as this PR is based on PR #762.

rhabacker:
yes

If you wish I can duplicate the associated commits into this pr to fix that issue. Another option is to merge #762, which was recently approved, into `maint`, so that I can update this pr.

jralls:
I'll merge #762 shortly and rerun the CI tests.

rhabacker:
> mac-tests / macOS CI Tests (pull_request) Failing after 15m — macOS CI Tests
```
         Start  34: test-engine
 34/113 Test  #34: test-engine ......................Subprocess aborted***Exception:   0.18 sec
```
ctest should be started with the `--output-on-failure` command line option to get an idea if this test failure is caused by changes from #762

rhabacker:
> > mac-tests / macOS CI Tests (pull_request) Failing after 15m — macOS CI Tests

That issue seemed to be a temporary failure on the associated builder.

jralls:
> ctest should be started with the --output-on-failure command line option to get an idea if this test failure is caused by changes from #762

@rhabacker That's not necessary. When the test run fails that action saves the test log as an artifact.

gnucash/gnome/gnc-plugin-page-register.c
```diff
@@ -742,6 +742,10 @@ gnc_plugin_page_register_new_common (GNCLedgerDisplay* ledger)
     if (!gnc_features_check_used (gnc_get_current_book(), GNC_FEATURE_REG_SORT_FILTER))
         gnc_features_set_used (gnc_get_current_book(), GNC_FEATURE_REG_SORT_FILTER);
 
+    // added for version 4.3 onwards
```
jralls, line 745: Shouldn't that be 4.5 (because of the snap release of 4.4)?

fellen: or is it the result of https://bugs.gnucash.org/show_bug.cgi?id=798070#c4?

rhabacker: > Shouldn't that be 4.5 (because of the snap release of 4.4)?

yes, this need to be updated after fixing the issue mentioned at https://bugs.gnucash.org/show_bug.cgi?id=798070#c4? See https://github.com/Gnucash/gnucash/pull/856 for an associated mr.

rhabacker:
With recent patch set no additional opening balance account is added (checked with german skr03)

jralls:
This needs two fixes:
* In `gnc_account_opening_balance_button_update` the account-type test should be on `aw->type` rather than `xaccAccountGetType` so that it reflects the state of the dialog rather than the account type before the dialog started.
* One can lock up GnuCash solidly by creating a second opening balance account in the same currency. This is in part due to doing the check for an existing one only in `change_currency_cb` and partly because that check calls `gnc_general_select_set_selected` when the error dialog closes, recursively running `change_currency_cb`. The trick is to not edit the existing opening balance account first. If you do then the opening_balance checkbutton stays disabled on the new account.

rhabacker:
> This needs two fixes:
> * In `gnc_account_opening_balance_button_update` the account-type test should be on `aw->type` rather than `xaccAccountGetType` so that it reflects the state of the dialog rather than the account type before the dialog started.

See #1009

> * One can lock up GnuCash solidly by creating a second opening balance account in the same currency. This is in part due to doing the check for an existing one only in `change_currency_cb` and partly because that check calls `gnc_general_select_set_selected` when the error dialog closes, recursively running `change_currency_cb`. 

I'm sorry, I cannot find a function or method named `change_currency_cb()`, I guess you are refering to `commodity_changed_cb()`.

When trying to reproduce this point, I was unable to build gnucash from the maint branch with the patches from #820 and #1009 due to the bug described on https://bugs.gnucash.org/show_bug.cgi?id=798193.

jralls:
> I'm sorry, I cannot find a function or method named change_currency_cb(), I guess you are refering to commodity_changed_cb().

Sorry, yes, `commodity_changed_cb`, in particular your [new block](https://github.com/Gnucash/gnucash/blame/maint/gnucash/gnome-utils/dialog-account.c#L1323).

rhabacker:
> One can lock up GnuCash solidly by creating a second opening balance account in the same currency

This can be triggered e.g. by creating a new file with the German account template skr03 and skr04. Are there any other ways to reproduce the problem ?

rhabacker:
An issue with multiple opening balance accounts in the hierachy assistant has been fixed with #1011.

rhabacker:
> Sorry, yes, `commodity_changed_cb`, in particular your [new block](https://github.com/Gnucash/gnucash/blame/maint/gnucash/gnome-utils/dialog-account.c#L1323).

See # 1012 for a fix. I hope I get this right as I'm not familary with gtk signal/slot system.

PR #821, opened 2020-11-20
[date-utilities] Convert reldate list into srfi-9 records

christopherlam:
Cleanup, rearranges strings. Need to confirm i18n still works well. @fellen

### DIFF
--- a/libgnucash/app-utils/c-interface.scm
+++ b/libgnucash/app-utils/c-interface.scm
@@ -78,6 +78,8 @@
 ;; translated with gettext.
 (define (gnc:make-string-database)
   (define string-hash (make-hash-table))
+  (issue-deprecation-warning "gnc:make-string-database is deprecate. record \
+translatable strings with N_ and G_.")
   (lambda args
     (match args
       (('lookup key) (G_ (hash-ref string-hash key)))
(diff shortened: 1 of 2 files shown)


jralls:
This file gets removed in [c++options](https://github.com/jralls/gnucash/tree/c++options), no point in tidying it up.

fellen:
I compared gnucash.pot before and after the commit: "One Week A{go|head}[.]" disappeared. Resaon?

N.B. This file has always been a terrible example of I18n:
 The tooltips are no full sentences.
Are full number words really required instead of a ngettext expression?  …

christopherlam:
@fellen "One week ago" was translatable but not actually inserted into the reldate list. There are no complaints about it; I'd remove it. @jralls any ETA for c++options to be ready?

jralls:
@christopherlam for what value of ready? I guess there's still 2-3 months of solid work left and I've managed about 3 days worth since July.

christopherlam:
It'll be a while till c++options is ready then. I'll push this change which means we can deprecate `gnc:make-string-database`

christopherlam:
To my shame, this branch breaks reports spectacularly. I'll await c++options to be complete then.

PR #822, opened 2020-11-21
Update Indonesian translation for GnuCash.

triyanwn:
Update Indonesian translation.

### DIFF
--- a/po/id.po
+++ b/po/id.po
@@ -1,5 +1,5 @@
-# Indonesian translations for GnuCash package
-# Terjemahan bahasa inggris untuk paket GnuCash.
+# Indonesian translations for GnuCash package.
+# Terjemahan bahasa Indonesia untuk paket GnuCash.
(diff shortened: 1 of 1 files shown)


fellen:
Shouldn't 'split' become 'perpisahan', 'pemisahan' or similar? I wonder, why you replaced it.

Commited as commit 562b4be after adding the statistics.

Thank you

triyanwn:
At first, I use pisahan/pemisahan, but then I think these words are not common to use in finance, so I leave it as is.

fellen:
Perhapsyou can find a term starting with https://id.wikipedia.org/wiki/Pembukuan_berpasangan ?

triyanwn:
Thanks, I will check it. My other consideration for not translating it is that some other accounting software doesn't translate it either, including GnuCash for Android. Apart from that, in my opinion, this term is safer not to translate, because there is no correct word equivalent in Bahasa Indonesia. If this term is translated, it can be misinterpreted by the user.

PR #823, opened 2020-11-22
Bug309943 - When unable to obtain a lock, no option to open another file

Bob-IT:
I thought I would add this here for any comment, this PR changes the 'Create New File' option on the said dialog to 'Open with No File'.
I think this is a better choice as it bring up the main GnuCash window and allows you to do 'File->New' which is the equivalent to current option, 'File->Open' so you can choose a different file or use one from the history.

If there are no concerns I will push this next week.

### DIFF
--- a/gnucash/gnome-utils/gnc-file.c
+++ b/gnucash/gnome-utils/gnc-file.c
@@ -695,7 +695,7 @@ gnc_file_query_save (GtkWindow *parent, gboolean can_cancel)
 
 /* private utilities for file open; done in two stages */
 
-#define RESPONSE_NEW  1
+#define RESPONSE_NO_FILE 1
 #define RESPONSE_OPEN 2
 #define RESPONSE_QUIT 3
 #define RESPONSE_READONLY 4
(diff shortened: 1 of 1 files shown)


gnucash/gnome-utils/gnc-file.c
```diff
@@ -881,10 +881,9 @@ gnc_post_file_open (GtkWindow *parent, const char * filename, gboolean is_readon
             break;
         default:
             /* Can't use the given file, so just create a new
-             * database so that the user will get a window that
-             * they can click "Exit" on.
+             * Gnucash window so they can choose File->New, File->Open
+             * or just "Exit" on.
```
jralls, line 885: The "on" at the end doesn't make sense.

Bob-IT: Thanks, missed that on the change of note.

jralls (approved):
Looks good.

gjanssens:
@Bob-IT I saw this just too late.
I commented on https://github.com/Gnucash/gnucash/commit/cc64f197ae2b27a0d21ede05636fd75685ebbb27#commitcomment-44791303 with the suggestion to add two buttons (create new file and open other file) rather than the single open with no file. A gnucash window without a file is not very useful, and can only serve to create, open or quit. The latter is already in the dialog, the former two aren't. By adding those directly that saves the user a few clicks and some confusion. It's not *that* obvious that choosing no file is the path to open a new or other file. That's coder's logic, not end user's.

PR #824, opened 2020-11-22
Bug 128772 - Account Help Button does not work

Bob-IT:
I am unsure about this one, I was looking at the bug and the context help was not working. I do not know how long it has not been working but hopefully not from the report date.

From testing, when pressing the help button, the following uri is created for the new account help button...
'ghelp:gnucash-help?acct-create' and used to open the gnome help app.

This segfaults in yelp at the '?' as 'ghelp:gnucash-help' does work.
To prove try at a command prompt...
gnome-help ghelp:gnucash-help?acct-create which segfaults but this...
gnome-help ghelp:/usr/share/gnome/help/gnucash-help/C/gnucash-help.xml?acct-create works.

To fix this, I have changed the context help to use the full path to the help files.
Looking at this I think I should add a test for getting a valid path and maybe an error dialog but will wait for any feedback first.

### DIFF
--- a/gnucash/gnome-utils/gnc-gnome-utils.c
+++ b/gnucash/gnome-utils/gnc-gnome-utils.c
@@ -236,6 +236,61 @@ gnc_add_css_file (void)
     g_object_unref (provider_fallback);
 }
 
+/* This function fixes an issue with yelp that it does not work with the
+ * ?anchor varient, see https://gitlab.gnome.org/GNOME/yelp/issues/116
+ */
+static gchar *
+gnc_gnome_help_yelp_anchor_fix (const char *file_name, const char *anchor)
+{
+    const gchar * const *sdatadirs = g_get_system_data_dirs ();
+    const gchar * const *langs = g_get_language_names ();
+    gchar *lookfor = g_strconcat ("gnome/help/", file_name, NULL);
+    gchar *help_path = NULL;
+    gchar *help_file = NULL;
+    gchar *full_path = NULL;
+    gchar *uri = NULL;
+
+    for (; *sdatadirs; sdatadirs++)
+    {
+        gchar *filepath = g_build_filename (*sdatadirs, lookfor, NULL);
+        if (g_file_test (filepath, G_FILE_TEST_EXISTS))
+            help_path = g_strdup (filepath);
+        g_free (filepath);
+    }
+    g_free (lookfor);
+
+    if (!help_path)
+    {
+        gnc_error_dialog (NULL, "%s\n%s", _(msg_no_help_found), _(msg_no_help_reason));
+        PERR("Unable to find 'gnome/help' directory");
+        return NULL;
+    }
+
+    // add the file extension, currently .xml
+    help_file = g_strconcat (file_name, ".xml", NULL);
+
+    for (; *langs; langs++)
+    {
+        gchar *filename = g_build_filename (help_path, *langs, help_file, NULL);
+        if (g_file_test (filename, G_FILE_TEST_EXISTS))
+            full_path = g_strdup (filename);
+        g_free (filename);
+    }
+    g_free (help_path);
+    g_free (help_file);
+
+    if (full_path)
+        uri = g_strconcat ("ghelp:", full_path, "?", anchor, NULL);
+    else
+    {
+        gnc_error_dialog (NULL, "%s\n%s", _(msg_no_help_found), _(msg_no_help_reason));
+        PERR("Unable to find valid help language directory");
+        return NULL;
+    }
+    g_free (full_path);
+    return uri;
+}
+
 #ifdef MAC_INTEGRATION
 
 /* Don't be alarmed if this function looks strange to you: It's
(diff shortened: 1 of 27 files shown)


jralls:
Please split that into a separate static function.

I'm a little surprised that there have been so few complaints about this. Makes one suspect that context help just isn't used very much and that maintaining it is wasted effort.

gnucash/gnome-utils/gnc-gnome-utils.c
```diff
@@ -421,7 +421,39 @@ gnc_gnome_help (const char *file_name, const char *anchor)
...
+
+        for (; *sdatadirs; sdatadirs++)
+        {
+            gchar *filename = g_build_filename (*sdatadirs, lookfor, NULL);
+            if (g_file_test (filename, G_FILE_TEST_EXISTS))
+                help_path = g_strdup (filename);
+            g_free (filename);
+        }
+
+        help_file = g_strconcat (file_name, ".xml", NULL);
```
fellen, line 440: ".xml" should also get a more prominent place. The documenters could decide to use ".dokbook" or switch to another format.

gnucash/gnome-utils/gnc-gnome-utils.c
```diff
@@ -421,7 +421,39 @@ gnc_gnome_help (const char *file_name, const char *anchor)
...
+        for (; *langs; langs++)
+        {
+            gchar *filename = g_build_filename (help_path, *langs, help_file, NULL);
+            if (g_file_test (filename, G_FILE_TEST_EXISTS))
+                out = g_strdup (filename);
+            g_free (filename);
+        }
+
+        if (out)
+            uri = g_strconcat ("ghelp:", out, "?", anchor, NULL);
```
fellen, line 451: How about making it smarter. Imagine anchor was added to C, but de got no update. Can we then fall back to C?

fellen:
> I'm a little surprised that there have been so few complaints about this. Makes one suspect that context help just isn't used very much and that maintaining it is wasted effort.

I still remember, when it occured to me first, I didn't know about bugzilla, mailinglists - probably not in my language -, …
I simply thought "Another low quality program".

IMHO it is a bootstrapping problem: If  I can not start the docs, I do not know, how to complain.

gnucash/gnome-utils/gnc-gnome-utils.c
jralls: Yelp can handle .html, maybe we should switch to that (not in this PR of course). Even better, we could ditch yelp and Windows help viewer entirely and display the docs with gnc-html.

jralls:
> IMHO it is a bootstrapping problem: If I can not start the docs, I do not know, how to complain.

This isn't about not being able to start the docs, it's about context help and even when that works it won't tell you anything about filing bugs. That information lives in the T&CG.

Speaking of bugs, [one of yours](https://bugs.gnucash.org/show_bug.cgi?id=796852) indicates that the low-quality program would be yelp. IMO dropping support for that and the seriously obsolete Windows help viewer is a better way forward. We already have the infrastructure to display html in GnuCash.

gjanssens:
The bug is in yelp really, not in gnucash. According to their own user manual our incantation should be supported:
https://people.gnome.org/~bmsmith/build/yelp.html#yelp-advanced-cmdline

So you could work around it in GnuCash as a service to our users, but better would be to fix it in yelp (ie file a  bug at yelp or provide a patch to their code to get it fixed). This is how the open source ecosystem as a whole advances after all.

Having said all that though, I do agree with @jralls it's probably better to just switch to using our built-in webkit to display help.

fellen:
> The bug is in yelp really, not in gnucash. According to their own user manual our incantation should be supported:
> https://people.gnome.org/~bmsmith/build/yelp.html#yelp-advanced-cmdline

That is the 14years old Gnome2 variant - with scolkeeper, …
It's a shame that documentation tools are so badly documented. ;-) 
 > So you could work around it in GnuCash as a service to our users, but better would be to fix it in yelp (ie file a bug at yelp

https://gitlab.gnome.org/GNOME/yelp/issues/116
> or provide a patch to their code to get it fixed). This is how the open source ecosystem as a whole advances after all.
> 
> Having said all that though, I do agree with @jralls it's probably better to just switch to using our built-in webkit to display help.

There have been more other options been discussed on the bugs, IIRC.

gjanssens:
Thanks for the link to the yelp bug @fellen 
Looks like we shouldn't wait for that to get fixed :(

gjanssens:
Also, while I'm fine with this workaround, I would recommend adding a comment on top explaining why it is necessary. If for some reason switching to our built-in webkit takes a long time, we may forget and try to simplify it again at some point.

gnucash/gnome-utils/gnc-gnome-utils.c
Bob-IT: Can not see an easy way of doing this.

Bob-IT:
I have created a separate function and added testing for a valid path and reporting error with the use of gnc_error_dialog. 
Also added comment for why it is there with reference to yelp bug report.
As a result of testing, I have added an other commit to add a GtkWindow parameter to gnc_gnome_help so the parent window can be specified for the error dialog which resulted in the last commit for the commodity help. Could not see why it was setup that way so changed it to be like others.

Bob-IT:
I have pushed this as a temporary fix and will have a look at creating a help dialog using existing 'gnc-html'

fellen:
> I have pushed this as a temporary fix and will have a look at creating a help dialog using existing 'gnc-html'

Another approach would be to talk with libyelp directly. But I am not sure, it is it worth.

PR #825, opened 2020-11-29
Bug355496 - Read Only Registers

Bob-IT:
Hopefully straight forward bug.
The first commit adds the account name to the read only warning dialog so when you start Gnucash with a read only register open you can tell which register the dialog refers to.
The second commit adds a test for the register being read only so that the same actions disabled for a read only book can be applied.

As part of the bug report, there was a request for some indication that a register is read only. The second commit my well of fulfilled that by disabling the menu actions but was wondering if more is required. One simple idea would be to augment the title with some text as this changes for each register tab, another idea was maybe to change the account icon to a padlock, slightly more involved but doable.

I will push this next weekend if no further indication required.

### DIFF
--- a/gnucash/gnome-utils/gnc-main-window.c
+++ b/gnucash/gnome-utils/gnc-main-window.c
@@ -2442,6 +2442,76 @@ main_window_update_page_color (GncPluginPage *page,
 }
 
 
+void
+main_window_update_page_set_read_only_icon (GncPluginPage *page,
+                                            gboolean read_only)
+{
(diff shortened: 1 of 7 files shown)


christopherlam:
I think a padlock icon would be nice

Bob-IT:
Improved original commit, now the complete sheet is marked insensitive when register is read only and now the actions change when you clear the account placeholder setting. Also added changing the icon to a padlock as seen here...
Untitled

PR #826, opened 2020-11-29
Translation Update pt

pmraps:
Update glossary/pt.po
Update Spouse income accounts
Update home loan accounts


jralls:
Thanks!

po/glossary/pt.po
```diff
@@ -210,14 +208,12 @@ msgid "balance (noun)"
...
-#, fuzzy
 msgid "balance b/f"
-msgstr "reconciliar"
+msgstr "saldo"
 
 #. "Balance carried forward - usually the last entry of an account statement to be used as 'balance b/f' on the next billing period or page"
-#, fuzzy
 msgid "balance c/f"
-msgstr "reconciliar"
+msgstr "saldo"
```
fellen, line 216: Isn't there a difference between the saldo at the beginning and the end of the account statement?

pmraps: According to the comments, I believe it is the value at the end of, e.g. a month that passes to the next month. Beginning balance and ending balance. So no, the word should be the same. I could add "inicial" and "final"?

fellen:
> Update glossary/pt.po
> Update Spouse income accounts
> Update home loan accounts

In theory I had preferred 2 commits
- Update pt templates
- Update pt glossary
(small logical units)

po/glossary/pt.po
fellen: > I could add "inicial" and "final"?

Sounds good.

jralls:
@fellen Translators have enough trouble working with git and translations don't break things. Take what you get and be grateful!

Committing small logical units is important for code commits because it makes bisecting work better to isolate the cause of bugs. That's not the case for translations.

PR #828, opened 2020-11-30
Update Croatian translation for Glossary and Application

milotype:
Translation updates and some corrections.
Also, please review comment.


fellen:
I removed 
```
QUESTION:
I found at https://wiki.gnucash.org/wiki/Release_Schedule, that the next release will be "4.3 on 2020-12-27". 
Is there a date for a string freeze before the 4.3 release, so translators can update the translation?
```
from the 2. commit message. It is a good question but does not belong in the git history, release announcement, …

We are currently discussing it. I am since years for 2 weeks before release date, @jralls ?

Finally I merged a fresh pot in commit cc64f19 so you can continue, if you like.

jralls:
2 weeks seems reasonable to me, but we need to say so on the devel list.

fellen:
https://lists.gnucash.org/pipermail/gnucash-devel/2020-December/045588.html

PR #829, opened 2020-12-01
Bug 798036 - Transaction Report should offer price source

christopherlam:
Getting crowded isn't it?
image


jralls:
Many of the other reports put currency selection and price source on a separate commodities page.

christopherlam:
Better.

image

PR #830, opened 2020-12-02
WIP: fix modules in app-utils and engine

christopherlam:
A start to fix bug 798034

Use more modern guile module definitions

Use nested module names e.g. `(app-utils business-prefs)` which is imported by `(app-utils)` and re-exported.

I *think* it's safe to apply to maint because top-level modules retain the same name.


jralls:
Is this building on your local machine?

christopherlam:
No it is not. It was probably building previously due to previous build
artifacts. I'll prod on ...

christopherlam:
Needs much more work.

PR #832, opened 2020-12-07
sort with date then number

holymonson:
When sorting `BY_DATE`, transactions are sorted by `Date` then `Transaction ID`. But `Transaction ID` is merely a random unique byte array, its order is meaningless. On the order hand, `Number` is a user-assigned field, which likely contains sequential info.

Therefore sorting with `Date` then `Number` is more reasonable.


jralls:
It's [xaccSplitOrder](https://github.com/Gnucash/gnucash/blob/maint/libgnucash/engine/Split.c#L1482) which first calls [xaccTransOrder](https://github.com/Gnucash/gnucash/blob/maint/libgnucash/engine/Transaction.c#L1889). As one can easily see the standard order is `date_posted`, `is_closing`, `num`, `date_entered`, `description`, and finally `guid`.
In the case where the splits are in a single transaction they're further sorted by the split fields `memo`, `action`, `reconciled`, `amount`, `value`, `date_reconciled`, and finally `guid`.

Your proposed change does nothing.

PR #831, opened 2020-12-04
Modularisation guile code

christopherlam:
I think this is now _ready_ for more scrutiny. Most warnings removed. Some `format` warnings remain.

Top-level modules:

`(gnucash core-utils)` `(gnucash engine)` `(gnucash app-utils)` `(gnucash qif-import)` `(gnucash report)` `(gnucash reports)`

`app-utils`, `qif-import`, `report` have submodules. `app-utils` and `report` will re-export submodules public functions.

`qif-import` modularisation isn't yet debugged especially C code calling scheme.

PR #833, opened 2020-12-08
Improve test-report-utilities.scm

christopherlam:
Use unique split amounts to when testing `gnc:account-accumulate-at-dates`. The 2nd commit uncomments the 4th transaction in bank1. 

https://github.com/Gnucash/gnucash/pull/831#issuecomment-739575308 - I don't understand what you meant by "only 3 transactions can be added" - with this 4th transaction, the accumulated balance must be changed, as expected, to add 40.

Still passes on guile-2.2 and guile-3.0. This branch could be merged prior to #831

### DIFF
--- a/gnucash/report/test/test-report-utilities.scm
+++ b/gnucash/report/test/test-report-utilities.scm
@@ -634,7 +634,7 @@
 
       (env-transfer env 15 01 1970 income bank1 10)
       (env-transfer env 15 02 1970 income bank1 20)
-;      (env-transfer env 15 03 1970 income bank1 40)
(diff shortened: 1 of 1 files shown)


jralls:
>  I don't understand what you meant by "only 3 transactions can be added" - with this 4th transaction, the accumulated balance must be changed, as expected, to add 40.

You've taken it out of context: The 4th transaction gets dropped after the modularization changes when using guile 2.x but not guile 3.0. 

There's no harm to merging this separately from modularization, but the point of it is to help figure out what modularization is changing that breaks the test.

PR #835, opened 2020-12-13
Fix (format) strings, importing ice-9 format where appropriate

christopherlam:
* because simple-format uses only ~a and ~s
* ice-9's format handles ~f

If we merge this it'll modify strings, unfortunately.

### DIFF
--- a/gnucash/report/reports/standard/budget-income-statement.scm
+++ b/gnucash/report/reports/standard/budget-income-statement.scm
@@ -46,6 +46,7 @@
 (use-modules (gnucash core-utils))
 (use-modules (gnucash app-utils))
 (use-modules (gnucash report))
+(use-modules (ice-9 format))
 
 ;; define all option's names and help text so that they are properly
 ;; defined in *one* place.


fellen:
> If we merge this it'll modify strings, unfortunately.

If I didn't miss something, none of your improved strings is marked as translatable. So they are not affected by the string freeze.

jralls:
I don't understand. All of the format strings were already ice-9 ones, betrayed by the presence of `#f` as the first argument: [SRFI-28](https://www.gnu.org/software/guile/manual/html_node/SRFI_002d28.html) formats don't have that and anyway require `(use-modules (srfi srfi-28))`, something that's never used in GnuCash. If there are some modules now raising warnings about format being undefined then just commit a fix to add the `use-modules`. You don't need a PR for that.

jralls:
I agree that getting the rational representation makes more sense than a general float in `test-cash-flow.scm`, and that letting `format` handle the rational directly in TXF export (in any other report it should be `~h` instead of `~f`). Those changes have nothing to do with having left out explicitly using ice-9 format and belong in a separate commit, also one that you don't need a PR for. The one that puzzles me is the only translatable string: If budget periods can be alphanumeric then I guess using `~a` makes sense, but that would be a separate commit called something like "Allow display of alphanumeric budget periods in the Budget Income Statement Report."

christopherlam:
The problem was that modules hadn't imported `(ice-9 format)`; guile-core has a `format` which doesn't handle anything other than ~a and ~s. I've fixed by importing `ice-9 format` which means there's no need to modify strings.

PR #836, opened 2020-12-13
[4.4] Use custom unbound-variable exception printer

christopherlam:
When a guile coder uses a variable but omits `use-modules`, this code will scan *all* available modules and offer the appropriate module name. As test, modify a .scm file (eg options.scm or report-core.scm) and add a reference to `gnc-build-url`.

Before: Unbound variable: gnc-build-url
After: Unbound variable: gnc-build-url. Did you forget (use-module (sw_gnc_html))?

Mergeable after 4.3 is released.

### DIFF
--- a/bindings/guile/core-utils.scm
+++ b/bindings/guile/core-utils.scm
@@ -29,6 +29,8 @@
 (eval-when (compile load eval expand)
   (load-extension "libgnucash-guile" "gnc_guile_bindings_init"))
 
+(use-modules (srfi srfi-26))
+(use-modules (ice-9 match))
 (use-modules (ice-9 i18n))
 
 (export N_)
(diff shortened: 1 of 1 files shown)


jralls:
Cool! Why does it need to wait for the 4.3 release?

christopherlam:
> Cool! Why does it need to wait for the 4.3 release?

Because it's not yet battle tested... Does it introduce new breakage or slowdowns?

I originally saw it in guix and paraphrased it...

christopherlam:
BTW guile can't do magic. Introducing `gnc-build-url` in libgnucash/app-utils/options.scm will throw unbound-variable error without suggesting module. It will show suggestion in gnucash/report/report-core.scm

PR #839, opened 2020-12-15
New bayes account matching algorithm

gruberchr:
See announcement in the [mailing list](https://lists.gnucash.org/pipermail/gnucash-devel/2020-December/045612.html) as well.

### DIFF
--- a/libgnucash/engine/Account.cpp
+++ b/libgnucash/engine/Account.cpp
@@ -74,7 +74,7 @@ static const std::string AB_TRANS_RETRIEVAL("trans-retrieval");
 
 static gnc_numeric GetBalanceAsOfDate (Account *acc, time64 date, gboolean ignclosing);
 
-using FinalProbabilityVec=std::vector<std::pair<std::string, int32_t>>;
+using FinalProbabilityVec=std::vector<std::pair<int32_t, std::string>>;
 using TokenSelection=std::set<std::string>;
 using TokenInfoVec=std::vector<std::pair<std::string, struct TokenAccountsInfo>>;
 using FlatKvpEntry=std::pair<std::string, KvpValue*>;
@@ -5469,15 +5469,6 @@ struct TokenAccountsInfo
     int64_t total_count;
 };
 
-/** holds an account guid and its corresponding integer probability
-  the integer probability is some factor of 10
- */
-struct AccountInfo
-{
-    std::string account_guid;
-    int32_t probability;
-};
-
 static void
 build_token_info(char const * suffix, KvpValue * value, TokenAccountsInfo & tokenInfo)
 {
@@ -5582,21 +5573,11 @@ build_probabilities(TokenInfoVec const & token_info, TokenSelection token_select
     {
         for (auto const & account : sum_of_probabilities)
         {
-            ret.push_back({account.first,
-                (int32_t)(account.second / (double)num_tokens * probability_factor)});
+            ret.push_back({(int32_t)(account.second / (double)num_tokens * probability_factor),
+                account.first});
         }
     }
 
-    return ret;
-}
-
-static AccountInfo
-highest_probability(FinalProbabilityVec const & probabilities)
-{
-    AccountInfo ret {"", std::numeric_limits<int32_t>::min()};
-    for (auto const & prob : probabilities)
-        if (prob.second > ret.probability)
-            ret = AccountInfo {prob.first, prob.second};
     return ret;
 }
 
@@ -5777,14 +5758,12 @@ gnc_account_imap_find_account_bayes (GncImportMatchMap *imap, GList *tokens)
     auto final_probabilities = build_probabilities(token_info, selected_tokens);
     if (final_probabilities.empty())
         return nullptr;
-    auto best = highest_probability(final_probabilities);
-    if (best.account_guid == "")
-        return nullptr;
-    if (best.probability < account_probability_threshold)
+    auto best = std::max_element(final_probabilities.begin(), final_probabilities.end());
+    if (best->first < account_probability_threshold)
         return nullptr;
     gnc::GUID guid;
     try {
-        guid = gnc::GUID::from_string(best.account_guid);
+        guid = gnc::GUID::from_string(best->second);
     } catch (gnc::guid_syntax_exception&) {
         return nullptr;
     }
--- a/libgnucash/engine/Account.cpp
+++ b/libgnucash/engine/Account.cpp
@@ -75,6 +75,7 @@ static gnc_numeric GetBalanceAsOfDate (Account *acc, time64 date, gboolean igncl
 
 using FinalProbabilityVec=std::vector<std::pair<std::string, int32_t>>;
 using ProbabilityVec=std::vector<std::pair<std::string, struct AccountProbability>>;
+using TokenInfoVec=std::vector<struct TokenAccountsInfo>;
 using FlatKvpEntry=std::pair<std::string, KvpValue*>;
 
 enum
@@ -5463,6 +5464,22 @@ build_token_info(char const * suffix, KvpValue * value, TokenAccountsInfo & toke
     }
 }
 
+static TokenInfoVec
+get_token_info(GncImportMatchMap * imap, GList * tokens)
+{
+    TokenInfoVec ret;
+    /* query token frequencies from import match map */
+    for (auto current_token = tokens; current_token; current_token = current_token->next)
+    {
+        TokenAccountsInfo token_info{};
+        auto token_name = static_cast <char const*> (current_token->data);
+        auto path = std::string(IMAP_FRAME_BAYES "/") + token_name + "/";
+        qof_instance_foreach_slot_prefix (QOF_INSTANCE (imap->acc), path, &build_token_info, token_info);
+        ret.push_back(token_info);
+    }
+    return ret;
+}
+
 /** We scale the probability values by probability_factor.
   ie. with probability_factor of 100000, 10% would be
   0.10 * 100000 = 10000 */
@@ -5497,17 +5514,14 @@ highest_probability(FinalProbabilityVec const & probabilities)
 }
 
 static ProbabilityVec
-get_first_pass_probabilities(GncImportMatchMap * imap, GList * tokens)
+get_first_pass_probabilities(TokenInfoVec token_info)
 {
     ProbabilityVec ret;
     /* find the probability for each account that contains any of the tokens
      * in the input tokens list. */
-    for (auto current_token = tokens; current_token; current_token = current_token->next)
+    for (auto const & token : token_info)
     {
-        TokenAccountsInfo tokenInfo{};
-        auto path = std::string{IMAP_FRAME_BAYES "/"} + static_cast <char const *> (current_token->data) + "/";
-        qof_instance_foreach_slot_prefix (QOF_INSTANCE (imap->acc), path, &build_token_info, tokenInfo);
-        for (auto const & current_account_token : tokenInfo.accounts)
+        for (auto const & current_account_token : token.accounts)
         {
             auto item = std::find_if(ret.begin(), ret.end(), [&current_account_token]
                 (std::pair<std::string, AccountProbability> const & a) {
@@ -5516,16 +5530,16 @@ get_first_pass_probabilities(GncImportMatchMap * imap, GList * tokens)
             if (item != ret.end())
             {/* This account is already in the map */
                 item->second.product = ((double)current_account_token.token_count /
-                                      (double)tokenInfo.total_count) * item->second.product;
+                                      (double)token.total_count) * item->second.product;
                 item->second.product_difference = ((double)1 - ((double)current_account_token.token_count /
-                                              (double)tokenInfo.total_count)) * item->second.product_difference;
+                                              (double)token.total_count)) * item->second.product_difference;
             }
             else
             {
                 /* add a new entry */
                 AccountProbability new_probability;
                 new_probability.product = ((double)current_account_token.token_count /
-                                      (double)tokenInfo.total_count);
+                                      (double)token.total_count);
                 new_probability.product_difference = 1 - (new_probability.product);
                 ret.push_back({current_account_token.account_guid, std::move(new_probability)});
             }
@@ -5704,7 +5718,8 @@ gnc_account_imap_find_account_bayes (GncImportMatchMap *imap, GList *tokens)
     if (!imap)
         return nullptr;
     check_import_map_data (imap->book);
-    auto first_pass = get_first_pass_probabilities(imap, tokens);
+    auto token_info = get_token_info(imap, tokens);
+    auto first_pass = get_first_pass_probabilities(token_info);
     if (!first_pass.size())
         return nullptr;
     auto final_probabilities = build_probabilities(first_pass);
--- a/libgnucash/engine/Account.cpp
+++ b/libgnucash/engine/Account.cpp
@@ -74,7 +74,7 @@ static const std::string AB_TRANS_RETRIEVAL("trans-retrieval");
 
 static gnc_numeric GetBalanceAsOfDate (Account *acc, time64 date, gboolean ignclosing);
 
-using FinalProbabilityVec=std::vector<std::pair<double, std::string>>;
+using FinalProbabilityVec=std::vector<std::pair<double, gnc::GUID>>;
 using TokenSelection=std::set<std::string>;
 using TokenInfoVec=std::vector<std::pair<std::string, struct TokenAccountsInfo>>;
 using FlatKvpEntry=std::pair<std::string, KvpValue*>;
@@ -5455,7 +5455,7 @@ gnc_account_imap_delete_account (GncImportMatchMap *imap,
 
 struct AccountTokenProbability
 {
-    std::string account_guid;
+    gnc::GUID account_guid;
     int64_t token_count; /** occurrences of a given token for this account_guid */
     double probability; /* conditional probability of this account_guid for a given token */
 };
@@ -5474,9 +5474,14 @@ build_token_info(char const * suffix, const KvpValue* const value, TokenAccounts
 {
     if (strlen(suffix) == GUID_ENCODING_LENGTH)
     {
-        tokenInfo.total_count += value->get<int64_t>();
-        /*By convention, the key ends with the account GUID.*/
-        tokenInfo.accounts.emplace_back(AccountTokenProbability{std::string{suffix}, value->get<int64_t>(), 0.0});
+        try
+        {
+            /*By convention, the key ends with the account GUID.*/
+            auto guid = gnc::GUID::from_string(std::string(suffix));
+            tokenInfo.total_count += value->get<int64_t>();
+            tokenInfo.accounts.emplace_back(AccountTokenProbability{guid, value->get<int64_t>(), 0.0});
+        }
+        catch (gnc::guid_syntax_exception&) {}  /* ignore this entry, if GUID is not valid */
     }
 }
 
@@ -5538,7 +5543,7 @@ build_account_probabilities(TokenInfoVec const & token_info, TokenSelection toke
 {
     /* Summarize conditional probabilities P(A|T) from token_info over all
      * tokens, which are contained in token_selection */
-    std::map<std::string, double> sum_of_probabilities;
+    std::map<gnc::GUID, double> sum_of_probabilities;
     auto num_tokens = 0;
     for (auto const & token : token_info)
     {
@@ -5755,13 +5760,7 @@ gnc_account_imap_find_account_bayes (GncImportMatchMap *imap, GList *tokens)
     auto best = std::max_element(final_probabilities.begin(), final_probabilities.end());
     if (best->first < account_probability_threshold)
         return nullptr;
-    gnc::GUID guid;
-    try {
-        guid = gnc::GUID::from_string(best->second);
-    } catch (gnc::guid_syntax_exception&) {
-        return nullptr;
-    }
-    auto account = xaccAccountLookup (reinterpret_cast<GncGUID*>(&guid), imap->book);
+    auto account = xaccAccountLookup (reinterpret_cast<GncGUID*>(&best->second), imap->book);
     return account;
 }
 
--- a/libgnucash/engine/Account.cpp
+++ b/libgnucash/engine/Account.cpp
@@ -74,7 +74,7 @@ static const std::string AB_TRANS_RETRIEVAL("trans-retrieval");
 
 static gnc_numeric GetBalanceAsOfDate (Account *acc, time64 date, gboolean ignclosing);
 
-using FinalProbabilityVec=std::vector<std::pair<int32_t, std::string>>;
+using FinalProbabilityVec=std::vector<std::pair<double, std::string>>;
 using TokenSelection=std::set<std::string>;
 using TokenInfoVec=std::vector<std::pair<std::string, struct TokenAccountsInfo>>;
 using FlatKvpEntry=std::pair<std::string, KvpValue*>;
@@ -5525,11 +5525,6 @@ select_tokens(TokenInfoVec const & token_info)
     return ret;
 }
 
-/** We scale the probability values by probability_factor.
-  ie. with probability_factor of 100000, 10% would be
-  0.10 * 100000 = 10000 */
-static constexpr int probability_factor = 100000;
-
 /*
  * This function calculates the average posterior probabilities
  *    mean(P(A|T1), P(A|T2), ...)
@@ -5573,8 +5568,7 @@ build_account_probabilities(TokenInfoVec const & token_info, TokenSelection toke
     {
         for (auto const & account : sum_of_probabilities)
         {
-            ret.push_back({(int32_t)(account.second / (double)num_tokens * probability_factor),
-                account.first});
+            ret.push_back({account.second / (double)num_tokens, account.first});
         }
     }
 
@@ -5742,7 +5736,7 @@ check_import_map_data (QofBook *book)
     imap_convert_bayes_to_flat_run = true;
 }
 
-static constexpr double account_probability_threshold = .50 * probability_factor; /* 50% */
+static constexpr double account_probability_threshold = .50; /* 50% */
 
 /** Look up an Account in the map */
 Account*
--- a/libgnucash/engine/Account.cpp
+++ b/libgnucash/engine/Account.cpp
@@ -74,7 +74,7 @@ static const std::string AB_TRANS_RETRIEVAL("trans-retrieval");
 static gnc_numeric GetBalanceAsOfDate (Account *acc, time64 date, gboolean ignclosing);
 
 using FinalProbabilityVec=std::vector<std::pair<std::string, int32_t>>;
-using TokenInfoVec=std::vector<struct TokenAccountsInfo>;
+using TokenInfoVec=std::vector<std::pair<std::string, struct TokenAccountsInfo>>;
 using FlatKvpEntry=std::pair<std::string, KvpValue*>;
 
 enum
@@ -5500,7 +5500,7 @@ get_token_info(GncImportMatchMap * imap, GList * tokens)
         {
             for (auto & account : token_info.accounts)
                 account.probability = (double)account.token_count / (double)token_info.total_count;
-            ret.push_back(token_info);
+            ret.emplace_back(token_name, token_info);
         }
     }
     return ret;
@@ -5527,7 +5527,7 @@ build_probabilities(TokenInfoVec const & token_info)
     std::map<std::string, double> sum_of_probabilities;
     for (auto const & token : token_info)
     {
-        for (auto const & account : token.accounts)
+        for (auto const & account : token.second.accounts)
         {
             /* try to insert new entry into sum_of_probabilities */
             auto result = sum_of_probabilities.emplace(account.account_guid, account.probability);
--- a/libgnucash/engine/Account.cpp
+++ b/libgnucash/engine/Account.cpp
@@ -74,7 +74,6 @@ static const std::string AB_TRANS_RETRIEVAL("trans-retrieval");
 static gnc_numeric GetBalanceAsOfDate (Account *acc, time64 date, gboolean ignclosing);
 
 using FinalProbabilityVec=std::vector<std::pair<std::string, int32_t>>;
-using ProbabilityVec=std::vector<std::pair<std::string, struct AccountProbability>>;
 using TokenInfoVec=std::vector<struct TokenAccountsInfo>;
 using FlatKvpEntry=std::pair<std::string, KvpValue*>;
 
@@ -5418,16 +5417,35 @@ gnc_account_imap_delete_account (GncImportMatchMap *imap,
  Below here is the bayes transaction to account matching system
 --------------------------------------------------------------------------*/
 
-
-/** intermediate values used to calculate the bayes probability of a given account
-  where p(AB) = (a*b)/[a*b + (1-a)(1-b)], product is (a*b),
-  product_difference is (1-a) * (1-b)
+/* The transaction to account matching algorithm uses bayesian inference to
+ * classify a sequence of tokens (observations), i.e. to find the account which
+ * most probably fits to the observed token sequence. See for instance
+ * https://en.wikipedia.org/wiki/Bayesian_inference.
+ *
+ * The underlying probabilistic model is a token generator, which generates
+ * a sequence of identically distributed tokens of arbitrary length with
+ * probability P(T|A), which is the conditional probability of token T given
+ * account A. But the tokens are not independent random variables and therefore
+ * not i.i.d. Instead there is a strong relation between tokens from the same
+ * transaction. That means the joint probability of a token sequence
+ * P(T1,T2,...|A), which belong to the same transaction is not simply the
+ * product of the probabilities of the individual tokens
+ * P(T1|A)*P(T2|A)*... .
+ *
+ * The currently implemented algorithm handles each token individually and
+ * does not consider the joint probabilities between tokens. These statistics
+ * are not captured at the moment, only the token frequencies per account.
+ *
+ * The account classification of a token sequence is realized by an estimator,
+ * which maximizes the posterior probability P(A|T) of each token regarding
+ * the unknown account. This is known as maximum a posteriori (MAP) estimation.
+ *
+ * Since each token is handled individually, the estimation results for each
+ * token of the sequence have to be combined. This is done by calculating an
+ * average posterior probability:
+ *    mean(P(A|T1), P(A|T2), ...)
+ * This final averaged probability is maximized for classification.
  */
-struct AccountProbability
-{
-    double product; /* product of probabilities */
-    double product_difference; /* product of (1-probabilities) */
-};
 
 struct AccountTokenProbability
 {
@@ -5493,21 +5511,45 @@ get_token_info(GncImportMatchMap * imap, GList * tokens)
   0.10 * 100000 = 10000 */
 static constexpr int probability_factor = 100000;
 
+/*
+ * This function calculates the average posterior probabilities
+ *    mean(P(A|T1), P(A|T2), ...)
+ * for each account required for MAP estimation.
+ *
+ * Maximizing these probabilities over all accounts (MAP estimation) returns
+ * the most probably matching account.
+ */
 static FinalProbabilityVec
-build_probabilities(ProbabilityVec const & first_pass)
+build_probabilities(TokenInfoVec const & token_info)
 {
-    FinalProbabilityVec ret;
-    for (auto const & first_pass_prob : first_pass)
+    /* Summarize conditional probabilities P(A|T) from token_info over all
+     * tokens, which are contained in token_selection */
+    std::map<std::string, double> sum_of_probabilities;
+    for (auto const & token : token_info)
     {
-        auto const & account_probability = first_pass_prob.second;
-        /* P(AB) = A*B / [A*B + (1-A)*(1-B)]
-         * NOTE: so we only keep track of a running product(A*B*C...)
-         * and product difference ((1-A)(1-B)...)
-         */
-        int32_t probability = (account_probability.product /
-                (account_probability.product + account_probability.product_difference)) * probability_factor;
-        ret.push_back({first_pass_prob.first, probability});
+        for (auto const & account : token.accounts)
+        {
+            /* try to insert new entry into sum_of_probabilities */
+            auto result = sum_of_probabilities.emplace(account.account_guid, account.probability);
+            if (!result.second)
+            {
+                /* entry already exists > summarize P(A|T) */
+                result.first->second += account.probability;
+            }
+        }
     }
+
+    /* Calculate the final average posterior account probabilities for the set
+     * of selected tokens by dividing summarized conditional probabilities by
+     * the number of tokens (arithmetic mean value of P(A|T1), P(A|T2), etc.
+     * for each account). */
+    FinalProbabilityVec ret;
+    for (auto const & account : sum_of_probabilities)
+    {
+        ret.push_back({account.first,
+            (int32_t)(account.second / (double)token_info.size() * probability_factor)});
+    }
+
     return ret;
 }
 
@@ -5518,38 +5560,6 @@ highest_probability(FinalProbabilityVec const & probabilities)
     for (auto const & prob : probabilities)
         if (prob.second > ret.probability)
             ret = AccountInfo {prob.first, prob.second};
-    return ret;
-}
-
-static ProbabilityVec
-get_first_pass_probabilities(TokenInfoVec token_info)
-{
-    ProbabilityVec ret;
-    /* find the probability for each account that contains any of the tokens
-     * in the input tokens list. */
-    for (auto const & token : token_info)
-    {
-        for (auto const & current_account_token : token.accounts)
-        {
-            auto item = std::find_if(ret.begin(), ret.end(), [&current_account_token]
-                (std::pair<std::string, AccountProbability> const & a) {
-                    return current_account_token.account_guid == a.first;
-                });
-            if (item != ret.end())
-            {/* This account is already in the map */
-                item->second.product = current_account_token.probability * item->second.product;
-                item->second.product_difference = (1.0 - current_account_token.probability) * item->second.product_difference;
-            }
-            else
-            {
-                /* add a new entry */
-                AccountProbability new_probability;
-                new_probability.product = current_account_token.probability;
-                new_probability.product_difference = 1.0 - new_probability.product;
-                ret.push_back({current_account_token.account_guid, std::move(new_probability)});
-            }
-        } /* for all accounts in tokenInfo */
-    }
     return ret;
 }
 
@@ -5724,11 +5734,10 @@ gnc_account_imap_find_account_bayes (GncImportMatchMap *imap, GList *tokens)
         return nullptr;
     check_import_map_data (imap->book);
     auto token_info = get_token_info(imap, tokens);
-    auto first_pass = get_first_pass_probabilities(token_info);
-    if (!first_pass.size())
+    if (token_info.empty())
         return nullptr;
-    auto final_probabilities = build_probabilities(first_pass);
-    if (!final_probabilities.size())
+    auto final_probabilities = build_probabilities(token_info);
+    if (final_probabilities.empty())
         return nullptr;
     auto best = highest_probability(final_probabilities);
     if (best.account_guid == "")
--- a/libgnucash/engine/Account.cpp
+++ b/libgnucash/engine/Account.cpp
@@ -5539,7 +5539,7 @@ static constexpr int probability_factor = 100000;
  * the most probably matching account.
  */
 static FinalProbabilityVec
-build_probabilities(TokenInfoVec const & token_info, TokenSelection token_selection)
+build_account_probabilities(TokenInfoVec const & token_info, TokenSelection token_selection)
 {
     /* Summarize conditional probabilities P(A|T) from token_info over all
      * tokens, which are contained in token_selection */
@@ -5755,7 +5755,7 @@ gnc_account_imap_find_account_bayes (GncImportMatchMap *imap, GList *tokens)
     if (token_info.empty())
         return nullptr;
     auto selected_tokens = select_tokens(token_info);
-    auto final_probabilities = build_probabilities(token_info, selected_tokens);
+    auto final_probabilities = build_account_probabilities(token_info, selected_tokens);
     if (final_probabilities.empty())
         return nullptr;
     auto best = std::max_element(final_probabilities.begin(), final_probabilities.end());
--- a/libgnucash/engine/Account.cpp
+++ b/libgnucash/engine/Account.cpp
@@ -5429,10 +5429,11 @@ struct AccountProbability
     double product_difference; /* product of (1-probabilities) */
 };
 
-struct AccountTokenCount
+struct AccountTokenProbability
 {
     std::string account_guid;
     int64_t token_count; /** occurrences of a given token for this account_guid */
+    double probability; /* conditional probability of this account_guid for a given token */
 };
 
 /** total_count and the token_count for a given account let us calculate the
@@ -5440,7 +5441,7 @@ struct AccountTokenCount
  */
 struct TokenAccountsInfo
 {
-    std::vector<AccountTokenCount> accounts;
+    std::vector<AccountTokenProbability> accounts;
     int64_t total_count;
 };
 
@@ -5460,7 +5461,7 @@ build_token_info(char const * suffix, KvpValue * value, TokenAccountsInfo & toke
     {
         tokenInfo.total_count += value->get<int64_t>();
         /*By convention, the key ends with the account GUID.*/
-        tokenInfo.accounts.emplace_back(AccountTokenCount{std::string{suffix}, value->get<int64_t>()});
+        tokenInfo.accounts.emplace_back(AccountTokenProbability{std::string{suffix}, value->get<int64_t>(), 0.0});
     }
 }
 
@@ -5468,14 +5469,21 @@ static TokenInfoVec
 get_token_info(GncImportMatchMap * imap, GList * tokens)
 {
     TokenInfoVec ret;
-    /* query token frequencies from import match map */
     for (auto current_token = tokens; current_token; current_token = current_token->next)
     {
+        /* query token frequencies from import match map */
         TokenAccountsInfo token_info{};
         auto token_name = static_cast <char const*> (current_token->data);
         auto path = std::string(IMAP_FRAME_BAYES "/") + token_name + "/";
         qof_instance_foreach_slot_prefix (QOF_INSTANCE (imap->acc), path, &build_token_info, token_info);
-        ret.push_back(token_info);
+
+        /* calculate conditional account probabilities P(A|T) of account A given token T */
+        if (token_info.total_count > 0)
+        {
+            for (auto & account : token_info.accounts)
+                account.probability = (double)account.token_count / (double)token_info.total_count;
+            ret.push_back(token_info);
+        }
     }
     return ret;
 }
@@ -5529,18 +5537,15 @@ get_first_pass_probabilities(TokenInfoVec token_info)
                 });
             if (item != ret.end())
             {/* This account is already in the map */
-                item->second.product = ((double)current_account_token.token_count /
-                                      (double)token.total_count) * item->second.product;
-                item->second.product_difference = ((double)1 - ((double)current_account_token.token_count /
-                                              (double)token.total_count)) * item->second.product_difference;
+                item->second.product = current_account_token.probability * item->second.product;
+                item->second.product_difference = (1.0 - current_account_token.probability) * item->second.product_difference;
             }
             else
             {
                 /* add a new entry */
                 AccountProbability new_probability;
-                new_probability.product = ((double)current_account_token.token_count /
-                                      (double)token.total_count);
-                new_probability.product_difference = 1 - (new_probability.product);
+                new_probability.product = current_account_token.probability;
+                new_probability.product_difference = 1.0 - new_probability.product;
                 ret.push_back({current_account_token.account_guid, std::move(new_probability)});
             }
         } /* for all accounts in tokenInfo */
--- a/libgnucash/engine/Account.cpp
+++ b/libgnucash/engine/Account.cpp
@@ -49,6 +49,7 @@ extern "C" {
 
 #include <numeric>
 #include <map>
+#include <set>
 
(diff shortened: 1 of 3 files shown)


libgnucash/engine/Account.cpp
```diff
@@ -5417,120 +5419,155 @@ gnc_account_imap_delete_account (GncImportMatchMap *imap,
...
+            ret.emplace_back(token_name, token_info);
+        }
     }
     return ret;
 }
 
-static AccountInfo
-highest_probability(FinalProbabilityVec const & probabilities)
+/** Threshold for conditional account probabilities given individual tokens */
+static constexpr double token_account_probability_threshold = .90; /* 90% */
```
bhardwajs, line 5506: Does this need to be `static` if it's already `constexpr`?

gruberchr:
See announcement in the [mailing list](https://lists.gnucash.org/pipermail/gnucash-devel/2020-December/045612.html) as well.

libgnucash/engine/Account.cpp
jralls: Either `static` or in the anonymous namespace to keep the symbol from being global.

gruberchr:
The failed test on Mac could be bug [798073](https://bugs.gnucash.org/show_bug.cgi?id=798073) and PR #864 could probably solve it.

jralls:
> The failed test on Mac could be bug 798073 and PR #864 could probably solve it.

It is bug 798073 but #864 is a bandaid, not a fix. I pushed a couple of changes yesterday that might or might not fix it. Since it's something of a transient error that's hard to say. But it's a known issue and has nothing to do with this PR.

libgnucash/engine/Account.cpp
```diff
@@ -5417,120 +5419,155 @@ gnc_account_imap_delete_account (GncImportMatchMap *imap,
...
-  the integer probability is some factor of 10
- */
-struct AccountInfo
-{
-    std::string account_guid;
-    int32_t probability;
-};
-
 static void
 build_token_info(char const * suffix, KvpValue * value, TokenAccountsInfo & tokenInfo)
```
bhardwajs, line 5472: Should `KvpValue*` be `KvpValue const*`?

bhardwajs: I checked out your branch and using `KvpValue const *` compiles.

libgnucash/engine/Account.cpp
```diff
@@ -5417,120 +5419,155 @@ gnc_account_imap_delete_account (GncImportMatchMap *imap,
...
-         * NOTE: so we only keep track of a running product(A*B*C...)
-         * and product difference ((1-A)(1-B)...)
-         */
-        int32_t probability = (account_probability.product /
-                (account_probability.product + account_probability.product_difference)) * probability_factor;
-        ret.push_back({first_pass_prob.first, probability});
+        /* query token frequencies from import match map */
+        TokenAccountsInfo token_info{};
+        std::string token_name{static_cast <char const*> (current_token->data)};
+        auto path = std::string(IMAP_FRAME_BAYES "/") + token_name + "/";
```
bhardwajs, line 5491: Reasoning about this code a bit more, `token_name` is a string created in line 5490 and then 5491, there is a new string created, and the only time, it's used is in 5499 which is done only if `token_info.total_count > 0`.

Can the string object creation be postponed till needed?

How about?
```
auto token_name = static_cast<char const *>(current_token->data);
auto path = std::string(IMAP_FRAME_BAYES "/").append(token_name).append("/");```
```

Line 5499 should be left as it is, since string is implicitly convertible from `const char *`

libgnucash/engine/Account.cpp
```diff
@@ -5417,120 +5419,155 @@ gnc_account_imap_delete_account (GncImportMatchMap *imap,
...
+static constexpr int probability_factor = 100000;
+
+static FinalProbabilityVec
+build_probabilities(TokenInfoVec const & token_info, TokenSelection token_selection)
 {
-    ProbabilityVec ret;
-    /* find the probability for each account that contains any of the tokens
-     * in the input tokens list. */
-    for (auto current_token = tokens; current_token; current_token = current_token->next)
+    std::map<std::string, double> sum_of_probabilities;
```
bhardwajs, line 5535: Can this be `std::unordered_map`?

libgnucash/engine/Account.cpp
```diff
@@ -5417,120 +5419,155 @@ gnc_account_imap_delete_account (GncImportMatchMap *imap,
...
+    auto num_tokens = 0;
+    for (auto const & token : token_info)
     {
-        TokenAccountsInfo tokenInfo{};
-        auto path = std::string{IMAP_FRAME_BAYES "/"} + static_cast <char const *> (current_token->data) + "/";
-        qof_instance_foreach_slot_prefix (QOF_INSTANCE (imap->acc), path, &build_token_info, tokenInfo);
-        for (auto const & current_account_token : tokenInfo.accounts)
+        /* check if token is selected for account probability calculation */
+        if (token_selection.find(token.first) != token_selection.end())
```
bhardwajs, line 5540: Consider using `token_selection.cend()`.

bhardwajs:
@jralls @gruberchr - Now that I have my build setup fully functional, I look at the code at length. This is a good change. I left a few comments if you want to look at.

Looking forward to this getting merged. :)

jralls:
@bhardwajs Please expand on why (modulo your proposed changes) you think this an improvement?

libgnucash/engine/Account.cpp
jralls: Unless there's a really good reason not to, it should be `std::vector`.

libgnucash/engine/Account.cpp
jralls: West const, please, and no space between the type and `*`.
@bhardwajs You shouldn't need to compile to test that `const KvpValue*` (actually `const KvpValue* const` is appropriate since the function changes neither what `value` points to nor what it contains): `KvpValue::get()` is declared `const noexcept`.

libgnucash/engine/Account.cpp
jralls: [Makes no difference](https://en.cppreference.com/w/cpp/container/set/end).

libgnucash/engine/Account.cpp
gruberchr: The reason for using a map here is, that this avoids an explicit search for existing entries. This is implicitly done by function `std::map::emplace()` (see line 5545). That's a good reason for me to prefer a map to `std::vector`.

@bhardwajs What improvement do you have in mind, when suggesting `std::unordered_map`?

gruberchr:
> Please expand on why (modulo your proposed changes) you think this an improvement?

Do you think of copying (parts of) the announcement from the [mailing list](https://lists.gnucash.org/pipermail/gnucash-devel/2020-December/045612.html) to commit messages 21afc42 and 04fef3c?

At least this announcement is linked to this PR (see my first comment above).

libgnucash/engine/Account.cpp
gruberchr: What's the difference between [`std::operator+(std::string)`](https://en.cppreference.com/w/cpp/string/basic_string/operator%2B) and [`std::string::append()`](https://en.cppreference.com/w/cpp/string/basic_string/append)?

jralls:
@gruberchr, First of all, that was directed specifically at @bhardwajs in response to him saying that he looks forward to this being merged.

That post tells a story but it doesn't do much to explain what's wrong with the existing implementation (in fact you admit that you don't understand the existing implementation) nor even very much about how this one works. The comment you've added to Account.cpp sort of describes what you're trying to accomplish but lacks the detail to help a reader understand what code corresponds to the steps in the process. Worse, neither does your naming. For example the comment discusses an "estimator" and calculating an "average posterior probablity mean" (which seems redundant, as a mean is what the general public calls an average) but there's neither an `estimator()` nor a variable called anything resembling  `posterior_probability`.

gruberchr:
@jralls Ok, I do understand. You request more explanation in the code or at least some more references between the algorithm description at the beginning of the bayes transaction to account matching section in *Account.cpp* and the functions therein. Additionally you suggest to use more expressive naming to make the link obvious between the code and the description.

Additional note on your comment:
> calculating an "average posterior probablity mean" (which seems redundant, as a mean is what the general public calls an average)

This is a misunderstanding. The description is not about an "average posterior probability mean", but about an "average posterior probability `mean(P(A|T1),P(A,T2),...)`.

jralls:
> This is a misunderstanding. The description is not about an "average posterior probability mean", but about an "average posterior probability `mean(P(A|T1),P(A,T2),...)`".

Ah. The comment can't have the format to make it clear that you're making a pseudocode function call there. A colon (':') after "probability" would help a little, maybe putting the pseudocode on its own line and indenting it a bit would help more.

jralls:
>  Additionally you suggest to use more expressive naming to make the link obvious between the code and the description.

Especially that. Expressive naming combined with well-composed functions is the best way to write understandable, and therefore maintainable, code.

gruberchr:
I pushed a fixup commit, which extends code documentation to hopefully improve comprehensibility. Additionally I renamed function `build_probabilities()`.

libgnucash/engine/Account.cpp
gruberchr: Changed as suggested

libgnucash/engine/Account.cpp
gruberchr: String object creation postponed as suggested

libgnucash/engine/Account.cpp
```diff
@@ -73,8 +74,9 @@ static const std::string AB_TRANS_RETRIEVAL("trans-retrieval");
 
 static gnc_numeric GetBalanceAsOfDate (Account *acc, time64 date, gboolean ignclosing);
 
-using FinalProbabilityVec=std::vector<std::pair<std::string, int32_t>>;
-using ProbabilityVec=std::vector<std::pair<std::string, struct AccountProbability>>;
+using FinalProbabilityVec=std::vector<std::pair<int32_t, std::string>>;
```
gruberchr, line 77: I was thinking about using account GUIDs (type `gnc::GUID`) here instead of account names (type `std::string`) to reduce unnecessary string allocations. Do you think, this is a good idea?

libgnucash/engine/Account.cpp
```diff
@@ -5417,120 +5419,165 @@ gnc_account_imap_delete_account (GncImportMatchMap *imap,
...
+    /* Calculate the final average posterior account probabilities for the set
+     * of selected tokens by dividing summarized conditional probabilities by
+     * the number of tokens (arithmetic mean value of P(A|T1), P(A|T2), etc.
+     * for each account). */
+    FinalProbabilityVec ret;
+    if (num_tokens > 0)
+    {
+        for (auto const & account : sum_of_probabilities)
+        {
+            ret.push_back({(int32_t)(account.second / (double)num_tokens * probability_factor),
```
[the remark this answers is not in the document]
jralls: I'm at a loss about that too.

libgnucash/engine/Account.cpp
jralls: Yes. There are a couple of other benefits: A 'gnc::GUID' is only 16 bytes while a `std::string` representation of one is 57, and `gnc::GUID::operator<` is only two comparisons while `std::string::operator<` could be up to 32 (admittedly unlikely but odds are good that it will be more than 2).

libgnucash/engine/Account.cpp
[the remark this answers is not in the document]
gruberchr: I'll remove this.

libgnucash/engine/Account.cpp
bhardwajs: Complexity of `unordered_map::emplace` is amortized constant on average while that of `map::emplace` is logarithmic in the size of the container. If the container needs to have unique collection of key-value pairs, my preference is to use `unordered_map` unless the container needs to be sorted
https://en.cppreference.com/w/cpp/container/unordered_map/emplace
https://en.cppreference.com/w/cpp/container/map/emplace

jralls: More usefully, `std::map` builds a b-tree of nodes (hence the log<sub>2</sub> complexity) while `std::unordered_map` is more like `GHashTable`: It hashes the keys and makes a linked list of entries on each hash value. It has been repeatedly demonstrated that linear search on a std::vector is several times faster than indexed lookup on any other container, [an example](https://blog.quasardb.net/using-c-containers-efficiently) and several orders of magnitude faster than std::map and std::unordered_map. I've heard claims that [Abseil's `flat_hash_map`](https://abseil.io/docs/cpp/guides/container) is nearly as fast as std::vector but I've not seen any benchmarks and I'd rather not add the dependency.

libgnucash/engine/Account.cpp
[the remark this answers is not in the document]
gruberchr: Done

libgnucash/engine/Account.cpp
gruberchr: I get a strange error, when replacing `std::string` by `gnc::GUID`:

`error: no match for ‘operator<’ (operand types are ‘const gnc::GUID’ and ‘const gnc::GUID’)`

This error occurs at lines
* 5551: `auto result = sum_of_probabilities.emplace(account.account_guid, account.probability);`
* 5755: `auto best = std::max_element(final_probabilities.begin(), final_probabilities.end());`

The first code line is in function `build_account_probabilities()` and `sum_of_probabilities` is a map, which requires to compare its keys. And the second code line is in function `gnc_account_imap_find_account_bayes()`, where the account with the maximum probability is searched, and although mainly the probabilities (first element of each pair) are compared, it can also happen, that there exist two entries with equal probability and the GUIDs (second element of each pair) have to be compared as well.

Why does the error message say, that I try to compare two `const gnc::GUID` values, although I don't use a `const` type specifier? And why isn't it possible to compare these values?

jralls: > Why does the error message say, that I try to compare two const gnc::GUID values, although I don't use a const type specifier? And why isn't it possible to compare these values?

Because the definition of `operator<` that c++ expects takes two const refs as if it was `template <typename T> bool operator<(const T&, const T&)`. (It's not really a template, it's built into the language). C++ doesn't know how to call a member operator function in a non-class context, so we need to implement the two-operand member function.

Change the decl and impl of operator< in guid.[hc]pp to
```
bool operator<(GUID const &) const noexcept;
```
and add
```
inline bool operator<(const gnc::GUID& l, const gnc::GUID& r) { return l.operator<(r); }
```
to the bottom of guid.hpp.

bhardwajs: @jralls beat me to it :)

gruberchr: > Change the decl and impl of operator< in guid.[hc]pp to `bool operator<(GUID const &) const noexcept;`

Done

> Add `inline bool operator<(const gnc::GUID& l, const gnc::GUID& r) { return l.operator<(r); }`

This  seems to be unnecessary. It looks like the compiler uses the `gnc::GUID` member operator function automatically. Conversely if I add this line to *guid.hpp*, I get an ambiguous overload error for `operator<`.

jralls:
I've worked out the two algorithms.
The existing  one calculates an account probability per token **P**<sub>t</sub> as `account_count / token_total` and an account_difference_probability **D**<sub>t</sub> 1 - **P**<sub>t</sub>. It accumulates the *products* of each, **P** and **D**. The final probability is **P**/(**P** + **D**).

@gruberchr's new algorithm is simpler: There is no **D** and **P** is the simple arithmetic mean of **P**<sub>t</sub> over the number of tokens remaining after testing each **P**<sub>t</sub> and discarding any token that doesn't have at least one **P**<sub>t</sub> > .9.

Both algorithms require a final probability > .9 to decide to decide a match.

Thanks to the threshold is likely to produce any better results than the plain matcher.

Let's work a simple example: Consider four tokens, `foo`, `bar`, `baz` and `qux`. Ignoring order there are 6 permutations, so let there be 6 accounts, one corresponding to each permutation. Each account gets one transaction per period and we've trained the matcher for 6 periods so each account has a count of 6 for each of its tokens. For any pair of tokens then there will be 1 with a count of 6 for both tokens, 4 with that count for one of the tokens, and two with no count. The total count for each token is 18 so **P**<sub>t</sub> is 1/3 and **D**<sub>t</sub> is 2/3.

Ignoring that @gruberchr's algorithm would return `nullptr` at this point because no token meets the threshold, the final probabilities for the account that matches both tokens will be 1/5 for the current algorithm (1/3)<sup>2</sup>/((1/3)<sup>2</sup> + (2/3)<sup>2</sup>) and 1/3 for @gruberchr's while the other 4 accounts will have final probabilities of 1/3 for the original and 1/6 for @gruberchr's. Neither meets the .9 threshold.

The original algorithm's method of combining the probabilities can be [mathematically supported](https://en.wikipedia.org/wiki/Naive_Bayes_spam_filtering#Combining_individual_probabilities) though that's in a different context from the one at hand. The arithmetic mean providing a more-correct result (a higher final probability for the account with two matching tokens) is due to the original algorithm's ignoring the 0 probability of the second token matches.

Most of which is a distraction, though because the right answer for P(A| T<sub>x</sub> && T<sub>y</sub>) is 1 and neither algorithm gets remotely close. We're using the wrong model: Selecting accounts is fundamentally different from rejecting spam email.

jralls:
The [solution](https://en.wikipedia.org/wiki/Bayes%27_theorem#Bayes'_rule_with_3_events) to P(A|T<sub>x<//sub>,T<sub>y</sub>) is P(T<sub>x</sub>|A,T<sub>y</sub>)*P(A|T<sub>y</sub>)/P(T<sub>x</sub>,T<sub>y</sub>).

From our uniform distribution of 4 tokens example, for each token T there are three accounts A containing it, so P(A|T<sub>y</sub>) is 1/3 (see below). Of the three instances of T<sub>y</sub> one is associated with T<sub>x</sub> so P(T<sub>x</sub>, T<sub>y</sub>) is also 1/3. P(T<sub>x</sub>|A, T<sub>y</sub>) is either 0 for the two accounts that use T<sub>y</sub> but not T<sub>x</sub> and 1 for the account that does; for that last account P(A|T<sub>x</sub>, T<sub>y</sub>) = 1 * 1/3 / 1/3 = 1; for the others it's 0. That's the right answer, demonstrating that this is the correct way to find the combined probability.

P(A|T<sub>x</sub>) = P(T<sub>x</sub>|A) / ( P(T<sub>x</sub>|A)*P(A) + P(T<sub>x</sub>| !A) * P(!A); for our example P(T<sub>x</sub>|A) is either 0 or one depending on A; P(A) is 1/6, there being 6 accounts; P(T<sub>x</sub>| !A) is either 2/5 or 1 and P(!A) is 5/6. if P(T<sub>x</sub>|A) is 0 then P(A|T<sub>x</sub>) is too; otherwise P(A|T<sub>x</sub>) = 1 * 1/6/(1 * 1/6 + 2/5 * 5/6) = 1/3.

gruberchr:
Thanks @jralls for your detailed comparison of both approaches.

I'm very surprised, that you refer to the Bayes spam filtering algorithm here, because nobody else mentioned that up to now in my [thread](http://gnucash.1415818.n4.nabble.com/GNC-dev-Understanding-the-bayesian-import-matching-algorithm-td4719812.html) on the mailing list.

Now I finally see, where the formula for the currently implemented algorithm comes from, it's the formula in this section https://en.wikipedia.org/wiki/Naive_Bayes_spam_filtering#Combining_individual_probabilities, which is linked by @jralls above.
The main problem with this formula is, it's **wrong**! See the corresponding [talk](https://en.wikipedia.org/wiki/Talk:Naive_Bayes_spam_filtering#Combining_probabilities_section_is_wrong) and my [last mail](https://lists.gnucash.org/pipermail/gnucash-devel/2020-July/045333.html) from the mailing list thread mentioned above.

But even applying the correct formulas of the naive bayes classifier didn't help. I tried this as well. The problem with this classifier is, that it neglects the dependencies between the tokens of one transaction. You can not simply assume, that the tokens are independent. I mentioned these strong dependencies in the [code documentation](https://github.com/Gnucash/gnucash/pull/839/commits/a3a94d32eab3d3ad171221ba3f89b93efef3f101#diff-a2fe08453182373caa62b6e813e7050bd5944a14592a1e1f0b9b9607be8e7605R5429). And the solution proposed by @jralls for the example of two tokens uses joint token probability P(T<sub>x</sub>, T<sub>y</sub>) as well. (By the way, I just recognized that there is a little mistake in your formula. According to the provided link P(T<sub>x</sub>, T<sub>y</sub>) has to be replaced by P(T<sub>x</sub> | T<sub>y</sub>))

Unfortunately the general solution, which could possibly be deduced by extending @jralls's solution to an arbitrary number of tokens, seems to require a very complex model. Moreover the currently collected statistical information in the import match map (account token frequencies) do not consider any token dependencies at all. I guess, without some simplifications the problem won't be tractable. The challenge is to find good simplifications.

Therefore at the moment I searched for an algorithm, which can be built on the currently available statistical information. This is the algorithm proposed in this PR. Any other algorithm, which requires more statistical information, will also require larger changes to the code.

Maybe there might be powerful improvements to my algorithm in the future. Then my algorithm is one further intermediate step towards these improvements. Nevertheless I would give it a try and we will see, how far one can go with it. As I mentioned in my [announcement](https://lists.gnucash.org/pipermail/gnucash-devel/2020-December/045612.html), I still see space for improvement even with this simple approach neglecting any dependencies between the tokens.

jralls:
@gruberchr I disagree with you on just about every point. The difference between a spam classifier and an account classifier is that the former is binary. That allows one to assume that P(A) == P(!A) so they cancel out. *Your algorithm makes the same error.* Spam tokens are no more independent than tokens made of descriptions. That's no reason to avoid the algorithm, it's pretty common in real-world statistical analysis to assume independence in order to simplify the math. No, the denominator on the joint probability function is *not* P(T<sub>x</sub>|T<sub>y</sub>), it's [P(T<sub>x</sub>, T<sub>y</sub>)](https://machinelearningmastery.com/classification-as-conditional-probability-and-the-naive-bayes-algorithm/)

Yes, the theoretically ideal machine learning algorithm is the maximum a posteriori but it's mathematically complex--though there are libraries available to handle the heavy lifting. Google's [tensorflow](https://github.com/tensorflow/tensorflow) is readily available, though I think that for this purpose it would be swatting flies with a sledgehammer. Your arm-waving about an arithmetic mean of probabilities somehow being a way to compute a MAP model is remarkably unconvincing. I don't recall ever encountering a claim that the arithmetic mean of probabilities is meaningful for *anything*. Perhaps you can provide a reference?

I found a [collection of web tutorials](https://machinelearningmastery.com/bayes-theorem-for-machine-learning/) that I think will prove useful here.

gruberchr:
> The difference between a spam classifier and an account classifier is that the former is binary

Correct. The [Bayesian spam classifier](https://en.wikipedia.org/wiki/Naive_Bayes_spam_filtering) is a [Naive Bayes classifier](https://en.wikipedia.org/wiki/Naive_Bayes_classifier) with only two classes. An account classifier has more than two classes, namely one for each account. Of course you can apply the formulas of the Naive Bayes classifier for an arbitrary number of classes to the account matching problem. Just take the formulas at the end of [this section](https://en.wikipedia.org/wiki/Naive_Bayes_classifier#Probabilistic_model) and replace C<sub>k</sub> by A<sub>k</sub>, which is the account with index k, and x<sub>n</sub> by T<sub>n</sub>, which is the token with index n. The result is the same as in section II within [this mail](https://lists.gnucash.org/pipermail/gnucash-devel/2020-July/045333.html).

Using this approach has at least two difficulties:
1. If any of the probabilities P(T<sub>n</sub> | A<sub>k</sub>) is zero, the whole account probability P(A<sub>k</sub> | T<sub>1</sub>, ..., T<sub>N</sub>) will be zero. At the end all account probabilities might be zero due to a few irrelevant tokens. That means, we do not have enough statistical information to consider all tokens in the probability calculation. This does not happen, if there are no products of probabilities in the formulas but only sums as with an arithmetic mean value. At least we have to do token selection as in my proposed algorithm. (The token selection in my approach has other reasons)
1. If there are tokens, which always occur in conjunction within the same transaction, their probabilities P(T<sub>n</sub> | A) are multiplied, which is P(T<sub>n</sub> | A)<sup>m</sup> with the number of tokens m, if the probabilities are identical. The correct joint probability for these tokens is the same as for a single token, which is P(T<sub>n</sub> | A). This situation occurs quite often in real world, where transactions related to the same payee have almost identical text with many recurring words/tokens. After some time with increasing number of tokens emitted by account A, the probabilities P(T<sub>n</sub> | A) get quite low. And raising low probabilities to powers of two, three or more leads to much lower joint probabilities than the correct value P(T<sub>n</sub> | A). That means accounts with many relevant but jointly occurring tokens gets a much lower account probability than other accounts with only one relevant token.


> Your algorithm makes the same error

What error do you mean precisely and how does my proposed algorithm make this error?

Third if you want to apply [Bayes' rule with 3 events](https://en.wikipedia.org/wiki/Bayes%27_theorem#Bayes'_rule_with_3_events) to solve P(A|T<sub>x</sub>, T<sub>y</sub>), you have to replace B by T<sub>x</sub> and C by T<sub>y</sub> in this formula, if I am not mistaken. This means P(A|T<sub>x</sub>, T<sub>y</sub>) = P(T<sub>x</sub> | A, T<sub>y</sub>) * P(A | T<sub>y</sub>) / P(T<sub>x</sub> | T<sub>y</sub>).

> I don't recall ever encountering a claim that the arithmetic mean of probabilities is meaningful for anything. Perhaps you can provide a reference?

Yes. From my point of view this can be seen as [Bayesian model averaging](https://en.wikipedia.org/wiki/Ensemble_learning#Bayesian_model_averaging). For each token there is an individual model used for calculating the account probability. And all these probabilities are averaged to calculate a final account probability averaged over all models/tokens.

jralls:
>>Your algorithm makes the same error

>What error do you mean precisely and how does my proposed algorithm make this error?

Like the original algorithm it fails to calculate the priors. That's justified in a binary classifier like Spam  Assassin because the unbiased priors are all .5 and they cancel out. As you acknowledge the account classifier has more choices, so you must acknowledge as well that the priors don't cancel out and must be computed and applied.

jralls:
> At least we have to do token selection as in my proposed algorithm. (The token selection in my approach has other reasons)

Token selection, maybe. As in your algorithm, definitely not. The way you did it is crude and unjustifiable.

> If any of the probabilities P(Tn | Ak) is zero, the whole account probability P(Ak | T1, ..., TN) will be zero.

That's correct, meaning that any algorithm for which it's true is correct and any for which it's not true is wrong. Yours is the latter.

That's what training is about: At the beginning all of the tokens have P(T<sub>x</sub>|A<sub>i</sub>) == 0 for all (x, y). The user selects the account and the tokens are added to the model, and over time the model's predictive ability improves. Eventually the model will be able to generate a useful signal and at that point presentation of a P == 0 token creates a need for a policy decision. Following the process strictly would lead to not proposing a candidate. That's OK. It might mean that some other account gets proposed and the user has to correct it; it might also mean that no account is proposed and the user has to select one. That adds data to the model so the next time or the time after that the classifier will get the right answer. It might at some point be reasonable to refine that behavior somehow, particularly in the presence of a token that has no counts for any account, but it's not necessary.

jralls:
> If there are tokens, which always occur in conjunction within the same transaction, their probabilities P(Tn | A) are multiplied, which is P(Tn | A)m with the number of tokens m, if the probabilities are identical. The correct joint probability for these tokens is the same as for a single token, which is P(Tn | A).

Yes, the simplifying assumption of Naive Bayes is imperfect.

>  This situation occurs quite often in real world, where transactions related to the same payee have almost identical text with many recurring words/tokens. ... After some time with increasing number of tokens emitted by account A, the probabilities P(Tn | A) get quite low.

Not necessarily. If the tokens in question are used exclusively by that one account then P(T<sub>x</sub>|A<sub>i</sub>)<sup>n</sup> is 1 for all values of n. If the same set of tokens is used by more than one account then the Naive Bayes model *correctly discounts their predictive power* and the reduction in probability is perhaps overstated. Wishing the problem away with bogus math doesn't solve that problem, collecting token correlation information does.

jralls:
> From my point of view this can be seen as Bayesian model averaging. 

That's unfortunate. Bayes Model Averaging involves assigning probabilities to the results of Bayes analysis using different statistical models over the same data and applying Bayes theorem to obtain a consolidated probability. [This paper](https://www.researchgate.net/publication/282403435_Bayesian_Model_Averaging_A_Systematic_Review_and_Conceptual_Classification) presents a nice introduction.

jralls:
> Third if you want to apply Bayes' rule with 3 events to solve P(A|Tx, Ty), you have to replace B by Tx and C by Ty in this formula, if I am not mistaken. This means P(A|Tx, Ty) = P(Tx | A, Ty) * P(A | Ty) / P(Tx | Ty).

Ah, I see my mistake: The P(T<sub>x</sub>, T<sub>y</sub>) version is P(A|T<sub>x</sub>,T<sub>y</sub>) is    
  P(A|T<sub>x</sub>, T<sub>y</sub>) = P(T<sub>x</sub>|A,T<sub>y</sub>)P(A|T<sub>y</sub>)**P(T<sub>y</sub>)**/P(T<sub>x</sub>,T<sub>y</sub>).
Divide top and bottom by P(T<sub>y</sub>) and you get
   P(A|T<sub>x</sub>, T<sub>y</sub>) =  P(T<sub>x</sub>|A,T<sub>y</sub>)P(A|T<sub>y</sub>)**P(T<sub>y</sub>)**/P(T<sub>x</sub>|T<sub>y</sub>).

[This stack overflow](https://math.stackexchange.com/questions/1281454/bayes-rule-with-3-variables/1281558#1281558) shows the derivation and my error. Sorry.

gruberchr:
> Like the original algorithm it fails to calculate the priors.

Sorry, I need more explanation. As I wrote in the code documentation the applied bayesian model in my approach is a token generator, which generates a token T<sub>i</sub>, with a probability P(T<sub>i</sub> | A<sub>k</sub>) given account A<sub>k</sub>. According to Bayes rule the posterior probability equals

P(A<sub>k</sub> | T<sub>i</sub>) = P(T<sub>i</sub> | A<sub>k</sub>) * P(A<sub>k</sub>) / P(T<sub>i</sub>).

If n<sub>i,k</sub> is the (absolute) frequency, a transaction containing token T<sub>i</sub> has been assigned to account A<sub>k</sub>, then

P(T<sub>i</sub> | A<sub>k</sub>) = n<sub>i,k</sub> / (sum(n<sub>i,k</sub>) over i)

P(A<sub>k</sub>) = (sum(n<sub>i,k</sub>) over i) / (sum(n<sub>i,k</sub>) over i and k)

P(T<sub>i</sub>) = (sum(n<sub>i,k</sub>) over k) / (sum(n<sub>i,k</sub>) over i and k)

And therefore

P(A<sub>k</sub> | T<sub>i</sub>) = P(T<sub>i</sub> | A<sub>k</sub>) * P(A<sub>k</sub>) / P(T<sub>i</sub>) = n<sub>i,k</sub> / (sum(n<sub>i,k</sub>) over k)

This is what function `get_token_info()` calculates. And as you can see, even if it's not obvious, the prior probability P(A<sub>k</sub>) is considered in the calculation.

Honestly I'm even not fully convinced to consider the priors at all. Doesn't this give the accounts with a high number of transactions precedence over other accounts with lower number of transactions?

gruberchr:
> Token selection, maybe. As in your algorithm, definitely not. The way you did it is crude and unjustifiable.

I don't object. This approach is mainly designed to somehow emulate the existing algorithm.

gruberchr:
> > If any of the probabilities P(Tn | Ak) is zero, the whole account probability P(Ak | T1, ..., TN) will be zero.
>
> That's correct, meaning that any algorithm for which it's true is correct and any for which it's not true is wrong. Yours is the latter.

From Bayesian theory this and the following explanation is correct. I agree. From a practical point of view this is quite annoying, since it requires the user to always train the model again and again as soon as new unknown tokens occur.
Sure, there are other solutions to solve this problem than my solution. But if a Bayesian classifier should be used, I wouldn't consider any token with zero probability. If you do this, you must be pretty sure, that this token reliably excludes the corresponding account and is not only random.
Following this explanation I can for instance imagine to extend the model to a [noisy token generator](https://en.wikipedia.org/wiki/Supervised_learning#Noise_in_the_output_values), which does not only generate tokens, which reliably can be assigned to any account, but also random tokens, which aren't meaningful for account classification. This would be some kind of token selection as well. Another idea could be to exclude all tokens, which did not produce enough statistical information yet.

gruberchr:
> > This situation occurs quite often in real world, where transactions related to the same payee have almost identical text with many recurring words/tokens. ... After some time with increasing number of tokens emitted by account A, the probabilities P(Tn | A) get quite low.
> 
> Not necessarily. If the tokens in question are used exclusively by that one account then P(Tx|Ai)n is 1 for all values of n.

You have to be careful. Only P(A<sub>i</sub> | T<sub>x</sub>) equals 1, if token T<sub>x</sub> is always assigned to the same account. But P(T<sub>x</sub> | A<sub>i</sub>) is surely much lower than 1, since an account generates many different tokens, whose probabilities must sum to 1.

gruberchr:
> > From my point of view this can be seen as Bayesian model averaging.
> 
> That's unfortunate. Bayes Model Averaging involves assigning probabilities to the results of Bayes analysis using different statistical models over the same data and applying Bayes theorem to obtain a consolidated probability.

You're right. I admit that I am not an expert in bayesian model averaging. But in the simplest case the probability for each model is equal. Doesn't this result in an arithmetic mean value?

But as I already stated as well, my approach is only a first proposal and is open for improvement. Maybe there is way to assign probabilities to each each model/token.

jralls:
The upshot of all this is that we need to infer some more probabilities and the training data doesn't support it very well. We can pretty easily get P(A<sub>i</sub>) directly from the account's split list. Currently parsing the descriptions into tokens loses the connections between the tokens so P(T<sub>x</sub>|A<sub>i</sub>) and P(T<sub>x</sub>|T<sub>y</sub>) can't be calculated from the training data as-is. We can get those from the account split list too but that involves re-parsing the descriptions into tokens. That may be expensive on an account with a lot of history; on the other hand it dispenses with the rather kludgey KVP storage of the training data.

The alternative is to modify the data collection to also count the other tokens in a string for each token.

The second part is to analyze the counts and compute the probabilities into an (i + x)<sup>2</sup> matrix that the classifier can use to compute the posterior probabilities for each set of tokens in an import. It would be nice to be able to store just that matrix, updated from the results of each import, so that we could amortize the previous operations and not have a long busy cursor before the match window can come up.

jralls:
> From a practical point of view this is quite annoying, since it requires the user to always train the model again and again as soon as new unknown tokens occur.

Not necessarily: The classifier can simply ignore a token that it hasn't seen before, and both the original and your algos do that. The problem is that they also ignore that a token has been seen before but isn't correlated to an account. *That* case needs to generate a 0 P(A<sub>i</sub>|T<sub>x</sub>) and disqualify the account until trained otherwise.

jralls:
> You're right. I admit that I am not an expert in bayesian model averaging. But in the simplest case the probability for each model is equal. Doesn't this result in an arithmetic mean value?

Only if all of the models have the same probability, and with the precondition that the models are equivalent, e.g. that they differ by fitting different distributions to the same training data (e.g. gaussian vs. binary, see the machine learning tutorial I referenced earlier). That's not even close to what you're doing.

jralls:
> You have to be careful. Only P(Ai | Tx) equals 1, if token Tx is always assigned to the same account. But P(Tx | Ai) is surely much lower than 1, since an account generates many different tokens, whose probabilities must sum to 1.

Uh, no. If I have an account Expenses:Utilities:Electricity (which of course I do) and the description is Pacific Gas and Electric (which happens to be the name of the local electricity company) on every single transaction with that account as the "other" account, then P(T|A) = 1 for T = Pacific, Gas, and Electric give A = Expenses:Utilities:Electricity. On the other hand for Expenses:Groceries where I might have three local supermarket names, Safeway, Raleys, and Whole Foods the P(T|A) will be 1/3 for each token: Each has a 1 in 3 likelihood of being in a transaction going to Expenses:Groceries. But Whole and Foods don't have a P(T|A) of 1/6: They're just as likely as Safeway and Raleys to turn up in a Groceries transaction.

It's also likely that Safeway and Raleys will have P(A|T) = 1 for Expenses:Groceries and that Gas and Electric will do for Expenses:Utilities:Electricity. Pacific, Whole, and Foods are more generic and might show up in other accounts and have a lower probability.

jralls:
> Honestly I'm even not fully convinced to consider the priors at all. Doesn't this give the accounts with a high number of transactions precedence over other accounts with lower number of transactions?

It does if you decide to weight the accounts based on their prevalence, but you don't have to. You could say that the accounts that have already appeared in splits with this account are equally likely; you could as well say that all other accounts are equally likely. What you *can't* say unless there are only two other accounts is that P(A) == P(!A) and remove them from the equation.

That said I'm inclined to weight the accounts that way as a prior and let Bayes's theorem work as it's supposed to. If P(A|T) = 0 for a T in the training set then A is dropped regardless of its prior probability.

gruberchr:
> The upshot of all this is that we need to infer some more probabilities and the training data doesn't support it very well.

That's exactly how it is. The alternative are simpler approaches like mine, which deal with the available statistics.

> We can get those from the account split list too but that involves re-parsing the descriptions into tokens.

Just a remark. This is, what I already proposed for other reasons on [Bugzilla](https://bugs.gnucash.org/show_bug.cgi?id=797778).

But what's actually your final conclusion from all this discussion? Would you keep the current algorithm, although it turned out to be wrong, until there is an algorithm available, which solves all the discussed difficulties with a proper probabilistic model? Or would you be ok with my proposal, which is available right now and could replace the currently wrong algorithm immediately?

I'm using and testing my approach for two months now and I'm quite satisfied with it. I haven't noticed any strange mapping failures since then (see my motivation for all this in the [announcement](https://lists.gnucash.org/pipermail/gnucash-devel/2020-December/045612.html)). And the remaining failures are really more difficult transactions, which do not contain any token, which is unique for one acount. But these are below 10 % of the transactions. The rest of the failing transactions are new transactions, which haven't been mapped before.

Have you actually tried my algorithm?

gruberchr:
> > From a practical point of view this is quite annoying, since it requires the user to always train the model again and again as soon as new unknown tokens occur.
> 
> Not necessarily: The classifier can simply ignore a token that it hasn't seen before, and both the original and your algos do that. The problem is that they also ignore that a token has been seen before but isn't correlated to an account. _That_ case needs to generate a 0 P(Ai|Tx) and disqualify the account until trained otherwise.

Exactly

gruberchr:
> > You have to be careful. Only P(Ai | Tx) equals 1, if token Tx is always assigned to the same account. But P(Tx | Ai) is surely much lower than 1, since an account generates many different tokens, whose probabilities must sum to 1.
> 
> Uh, no. If I have an account Expenses:Utilities:Electricity (which of course I do) and the description is Pacific Gas and Electric (which happens to be the name of the local electricity company) on every single transaction with that account as the "other" account, then P(T|A) = 1 for T = Pacific, Gas, and Electric give A = Expenses:Utilities:Electricity. 

Really? The statistical information, which is available in the current frequency table, is P(T|A) = 1/4 for all four tokens, since the account generates these four tokens with equal probability and these probabilities must sum to 1.

> On the other hand for Expenses:Groceries where I might have three local supermarket names, Safeway, Raleys, and Whole Foods the P(T|A) will be 1/3 for each token.

Maybe we're not talking about the same probability. I guess you mean the probability, that these tokens occur *within a single transaction, which is assigned to a specific account*. But that's not the information, which is currently collected. At the moment the transactions aren't counted at all, only the tokens. At the moment you only see, that there are some tokens assigned to the same account, but you don't know, if they belong to the same transaction or to different transactions. The first example reflects the situation, I explained above, that several tokens always occur *in conjunction*. To capture this information you need conditional probabilities for tokens within the same transaction. In the first example P("Gas" | "Pacific") = P("and" | "Pacific") = P("Electric" | "Pacific") = 1. In the second example P("Raleys" | "Safeway") = P("Whole" | "Safeway") = P("Foods" | "Safeway") = 0, but P("Foods" | "Whole") = 1.

That's the main reason, why the independence assumption for the tokens doesn't hold for transaction to account matching.

gruberchr:
> > Honestly I'm even not fully convinced to consider the priors at all. Doesn't this give the accounts with a high number of transactions precedence over other accounts with lower number of transactions?
> 
> It does if you decide to weight the accounts based on their prevalence, but you don't have to. You could say that the accounts that have already appeared in splits with this account are equally likely; you could as well say that all other accounts are equally likely. What you _can't_ say unless there are only two other accounts is that P(A) == P(!A) and remove them from the equation.

Of course, I agree. But if you weight all accounts equally, the likelihoods are removed from the equation as well, aren't they?

> That said I'm inclined to weight the accounts that way as a prior and let Bayes's theorem work as it's supposed to. If P(A|T) = 0 for a T in the training set then A is dropped regardless of its prior probability.

I guess you mean P(T|A) = 0.

jralls:
> But what's actually your final conclusion from all this discussion? Would you keep the current algorithm, although it turned out to be wrong, until there is an algorithm available, which solves all the discussed difficulties with a proper probabilistic model? Or would you be ok with my proposal, which is available right now and could replace the currently wrong algorithm immediately?

No change until someone writes a correct implementation. Replacing an incorrect implementation with an even less correct one makes no sense.

> I'm using and testing my approach for two months now and I'm quite satisfied with it.

So it's wrong tuned for your book. That's as likely to mess up with my book as not.

jralls:
> Of course, I agree. But if you weight all accounts equally, the likelihoods are removed from the equation as well, aren't they?

No, though the result does wind up looking a lot like an arithmetic mean.

But that misses the point: The models in Bayes Model Averaging are *complete models*, differing by e.g. the distribution that's fitted to each variable. Your claim to using BMA rests on incomplete models each examining only one variable.

The rule upon which BMA is based is the OR combination rule: The combined probability P(A|v<sub>i</sub> OR v<sub>j</sub>) is P(A|v<sub>i</sub>)P(v<sub>i</sub>) + P(A|v<sub>j</sub>)P(v<sub>j</sub>). A classifier needs to find P(A}|v<sub>i</sub> AND v<sub>j</sub>). IOW the premise for BMA is "we believe that one of these models reflects the actual distribution of P(A|V) for a group of V but we're not sure which one, so we compute P(A|V) for all of them and build a weighted OR likelihood."  That's not what a classifier needs. It needs the joint P(A|V<sub>i</sub>) for all i, and *that* computation is Bayes's theorem.

Bottom line: The existing computation is closer to being right than yours is.

jralls:
> I guess you mean P(T|A) = 0.

I didn't, but that will work too.

jralls:
> Really? The statistical information, which is available in the current frequency table, is P(T|A) = 1/4 for all four tokens, since the account generates these four tokens with equal probability and these probabilities must sum to 1.

Yes, really. The current frequency table is immaterial, especially if it makes no sense.

> Maybe we're not talking about the same probability. I guess you mean the probability, that these tokens occur within a single transaction, which is assigned to a specific account. But that's not the information, which is currently collected. At the moment the transactions aren't counted at all, only the tokens. At the moment you only see, that there are some tokens assigned to the same account, but you don't know, if they belong to the same transaction or to different transactions. The first example reflects the situation, I explained above, that several tokens always occur in conjunction. To capture this information you need conditional probabilities for tokens within the same transaction. In the first example P("Gas" | "Pacific") = P("and" | "Pacific") = P("Electric" | "Pacific") = 1. In the second example P("Raleys" | "Safeway") = P("Whole" | "Safeway") = P("Foods" | "Safeway") = 0, but P("Foods" | "Whole") = 1.

Bingo!

There are two pieces of this puzzle: One is the counts that we collect (not enough) and the other is the way we compute the model (the probabilities from those counts). In order to apply Bayes's Theorem correctly we must get both right.

> That's the main reason, why the independence assumption for the tokens doesn't hold for transaction to account matching.

It doesn't have anything at all to do with independence. Naive Bayes assumes independence, but comes close even in the case of some dependency between the tokens. Full-on Bayes doesn't assume that and P(T<sub>i</sub>|T<sub>j</sub>) is what takes that dependence into account.

So let's back up a bit. The original classifier is based on Spam Assassin's classifier which isn't strictly Bayesian. 

Spam Assassin's classifier compares the ratio that a particular work appears in spam and ham and uses that ratio as a probability P(S|W), the likelihood that a message is Spam if word W appears in it. It computes that probability for every word in the message and aggregates them with product(S|W)/(product(S|W) + product(H|W)) and multiplies the result time some constant to make a score. It also looks and hundreds of other things about the message and makes a score for those as well, then adds up all of the scores and applies a threshold. If the aggregate score exceeds the threshold then the message is spam.

The original algo tries to do something similar: The two states are is and is not account A. If token T has appeared 50 times and 7 of them are account A then P(A|T) = 7/50 P(!A|T) = 43/50. Apply the same rule as Spam Assassin does to get a score for each account that has the tokens in the presented description, pick the one with the highest score, and then see if that score exceeds an arbitrary threshold. If so, propose the match, otherwise leave it up to the user.

Definitely not pure Bayes. Definitely mathematically suspect at best. But it has worked sort-of-mostly-OK for many years for a lot of users.

jralls:
> Have you actually tried my algorithm?

No, because you've provided no tests or test data. You haven't even presented an example of cases where the existing algo finds the wrong account and your algo finds the right one.

The current algorithm works quite well for my book, so it's not possible for yours to be an improvement, but that's no more meaningful than that it doesn't work well for you. Two real-life test cases isn't a very useful test.

jralls:
> Definitely not pure Bayes. Definitely mathematically suspect at best. But it has worked sort-of-mostly-OK for many years for a lot of users.

I lost track of my intended point on that comment: It's that the data collection design is correct for that implementation, but as noted elsewhere won't work for a more correct Bayes's Theorem algo. I rather doubt that it's possible to devise a more-correct classifier using it.

gruberchr:
> > But what's actually your final conclusion from all this discussion?
> 
> No change until someone writes a correct implementation. Replacing an incorrect implementation with an even less correct one makes no sense.

Very sad, so someone else can try his hand at this problem. I already spent a lot of time on it, because it's a tough nut to crack. I'll create at least a bug ticket about the current implementation, since it's clear now, that it's definitely wrong.

> > I'm using and testing my approach for two months now and I'm quite satisfied with it.
> 
> So it's wrong tuned for your book.

What do you think is tuned for my book? I use the same implementation for my book as provided in this PR. 

> That's as likely to mess up with my book as not.

Sorry, I don't understand that.

gruberchr:
> > Of course, I agree. But if you weight all accounts equally, the likelihoods are removed from the equation as well, aren't they?
> 
> No, though the result does wind up looking a lot like an arithmetic mean.
> 
> But that misses the point: The models in Bayes Model Averaging are _complete models_, differing by e.g. the distribution that's fitted to each variable. Your claim to using BMA rests on incomplete models each examining only one variable.
> 
> The rule upon which BMA is based is the OR combination rule: The combined probability P(A|vi OR vj) is P(A|vi)P(vi) + P(A|vj)P(vj). A classifier needs to find P(A}|vi AND vj). IOW the premise for BMA is "we believe that one of these models reflects the actual distribution of P(A|V) for a group of V but we're not sure which one, so we compute P(A|V) for all of them and build a weighted OR likelihood." That's not what a classifier needs. It needs the joint P(A|Vi) for all i, and _that_ computation is Bayes's theorem.
> 
> Bottom line: The existing computation is closer to being right than yours is.

Something got mixed up here. The initial thread was about the likelihoods of the accounts not about Bayes Model Averaging. If it does, then I must have misunderstood something quite a bit.

gruberchr:
> So let's back up a bit. The original classifier is based on Spam Assassin's classifier which isn't strictly Bayesian.

What is Spam Assassin's classifier? Can you provide a reference?

jralls:
Spam Assassin's classifier is what [this wikipedia article](https://en.wikipedia.org/wiki/Naive_Bayes_spam_filtering#Combining_individual_probabilities) is based on. The sources are somewhere in [here](https://github.com/apache/spamassassin).

jralls:
>  Something got mixed up here. The initial thread was about the likelihoods of the accounts not about Bayes Model Averaging. If it does, then I must have misunderstood something quite a bit.

Sorry, I got the two threads about averaging and combining, err, combined.

> Of course, I agree. But if you weight all accounts equally, the likelihoods are removed from the equation as well, aren't they?

No. Recall that the formula is P(A|Tx) = P(Tx|A)*P(A) / ( P(Tx|A)*P(A) + P(Tx| !A) * P(!A)). Obviously you can only divide out P(A) if P(A) == P(!A) and that can be true only if there's only one A and it has an even chance of being true or not true. If you're trying to choose among n possible A then the unbiased prior P(A) = 1/n and P(!A) = (n - 1)/n.

gruberchr:
> > Have you actually tried my algorithm?
> 
> No, because you've provided no tests or test data. You haven't even presented an example of cases where the existing algo finds the wrong account and your algo finds the right one.

You're right. But writing tests would be further work for me, for which I don't want to spent time at the moment. But I provided a list of all critical situations I could detect, where the current transaction to account mapping fails in [this](https://lists.gnucash.org/pipermail/gnucash-devel/2020-October/045549.html) mail from my mail thread. In brief the critical situations occur with quite irrelevant tokens, which showed up once before and are recurring again. In this case the other account, where that token showed up first, immediately gets a probability of 1, but it's mostly the wrong account. So there is more than one account with a probability of 1, the correct one and the wrong one. Or even worse there is only the wrong account with a probability of 1. That means that these tokens distract the algorithm heavily. Examples for such irrelevant tokens are time and date information, but I could also find other short text snippets.

> The current algorithm works quite well for my book, so it's not possible for yours to be an improvement, but that's no more meaningful than that it doesn't work well for you. Two real-life test cases isn't a very useful test.

Do you want to say, that it's "not possible to improve" an algorithm, which is not perfect but only "quite well"? Sounds strange.

gruberchr:
> > Of course, I agree. But if you weight all accounts equally, the likelihoods are removed from the equation as well, aren't they?
> 
> No. Recall that the formula is P(A|Tx) = P(Tx|A)*P(A) / ( P(Tx|A)*P(A) + P(Tx| !A) * P(!A)). Obviously you can only divide out P(A) if P(A) == P(!A) and that can be true only if there's only one A and it has an even chance of being true or not true. If you're trying to choose among n possible A then the unbiased prior P(A) = 1/n and P(!A) = (n - 1)/n.

Ok, if you apply this formula, you're right. But if you apply the full Bayes formula at the end of [this section](https://en.wikipedia.org/wiki/Naive_Bayes_classifier#Probabilistic_model) considering all accounts individually, the likelihoods can be cancelled out.

jralls:
> Sorry, I don't understand that.

Sorry, not surprised.

The basics are that you claim, without providing any supporting data, that your algorithm works on your personal book. That's very nice for you but provides no support whatever for the hypothesis that it will work at all, never mind better than the current implementation, on any other book.

> Ok, if you apply this formula, you're right. But if you apply the full Bayes formula at the end of this section considering all accounts individually, the likelihoods can be cancelled out.

Uh-huh.  Do you not see P(C<sub>k</sub> in that equation?

I'm sorry you wasted so much time trying to solve a problem you don't understand.

gruberchr:
> > Sorry, I don't understand that.
> 
> Sorry, not surprised.

I only meant, that I didn't know how to translate this.

> > Ok, if you apply this formula, you're right. But if you apply the full Bayes formula at the end of this section considering all accounts individually, the likelihoods can be cancelled out.
> 
> Uh-huh. Do you not see P(Ck in that equation?

Of course. Look at the equations: p(C<sub>k</sub> | x<sub>1</sub>, ..., x<sub>1</sub>) has the factor p(C<sub>k</sub>) in the nominator as well as in the denominator Z. In the nominator index k is for a specific account. In the denominator k is a running index for the sum over all accounts. And if all p(C<sub>k</sub>) are equal, they can be factored out of the sum in the denominator and finally cancelled out. 

> I'm sorry you wasted so much time trying to solve a problem you don't understand.

I think it is not a good idea to denigrate supporters of the GnuCash project like this. I see no reason to become rude and make unreasonable claims.

One problem is that we seem to talk past each other sometimes. It is not easy to explain a complex mathematical problem with simple textual means without being misunderstood.

And looking back at the discussion, I had to correct you several times, too.

I think I understood the problem very well, after all, I have dealt with it extensively. And I am well aware of the simplifications that I have made. I merely tried to approach the larger goal in small steps and not to rework everything right away. After all, my time is limited.

If you don't like my simplifications, that's your decision, but no reason to accuse me lack of knowledge.

By the way, nobody on the mailing list was able to give any hints about the current algo. So I had no knowledge at all about its mathematical background until now. Are you the only one, who is able to explain it?

I realized that you seem to be quite convinced  of the current implementation. I am not. The difference between you and me is, that I've dealt with the problem practically during the last months. You seem to look at it only theoretically at the moment. For me the current implementation is quite crude. It works only in the simple case, that there is at least one token, which is unique to one account and that there are no tokens, which are unique to any other account. Luckily this is the case many times. But as soon as these conditions are violated, the algorithm behaves more or less random.

I wish you good luck!

jralls:
> I only meant, that I didn't know how to translate this.

Sorry about misinterpreting you, then. Your English is good enough that I assumed you were perfectly fluent. If you can, please say what in particular you're not able to translate/understand so that I can try to reword it.

> Of course. Look at the equations: p(Ck | x1, ..., x1) has the factor p(Ck) in the nominator as well as in the denominator Z. In the nominator index k is for a specific account. In the denominator k is a running index for the sum over all accounts. And if all p(Ck) are equal, they can be factored out of the sum in the denominator and finally cancelled out.

OK, I see how you get there; the problem with it is that it's dependent on computing the overall P(X); that article proceeds to prescribe three distribution models for doing so, of which [Bernoulli](https://en.wikipedia.org/wiki/Naive_Bayes_classifier#Bernoulli_naïve_Bayes) is the one that applies in our case. It multiplies the terms p<sub>ki</sub>, explaining that that's what we've been denoting P(X<sub>i</sub>|C<sub>k</sub>). As we've already agreed, we don't collect enough data to compute that correctly.

> By the way, nobody on the mailing list was able to give any hints about the current algo. So I had no knowledge at all about its mathematical background until now. Are you the only one, who is able to explain it?

On the mailing list @derekatkins offered to help but the thread petered out without discussing anything substantial. That's doesn't mean that nobody was able. You never asked any concrete questions and Derek didn't follow up with an explanation unprompted. That doesn't mean that nobody was able to, it means nobody cared to take the time to write a detailed answer to a vague request. No, I'm not the only one who could have explained that it's an adaptation of the common spam-filtering algorithm, but while it took me a little time to sort out the crappy code, once I did it was pretty obvious.

> I realized that you seem to be quite convinced of the current implementation. I am not. 

Neither am I, and I've said so several times. What I am convinced of is that your implementation is worse. Sorry to be harsh, but I've spent several hours on this that would have been better used on more important work.

PR #840, opened 2020-12-16
move (gnucash string) from qif-imp into bindings

christopherlam:
The string.scm file contains general string utility functions, not limited to qif processing.

I think @gjanssens intends to move it to bindings?


jralls:
No, it's not a binding, it's part of the qif importer. It should stay put unless the 7 functions it provides can be replaced with standard Scheme ones in which case they should be so replaced and string.scm should be removed.

christopherlam:
Ok.

PR #841, opened 2020-12-16
Free the char 'code' after g_malloc0 in gnc-flicker-gui.c

Bob-IT:
I think this is correct but do not like changing stuff I can not test.
@jralls Do these changes look correct ?

Change flicker_data to return a guint so that the char code can be
freed in the function which also removes the need to use g_malloc0
twice.

### DIFF
--- a/gnucash/import-export/aqb/gnc-flicker-gui.c
+++ b/gnucash/import-export/aqb/gnc-flicker-gui.c
@@ -120,13 +120,13 @@ flicker_data (const char *challenge)
     /* bitfield is a clock bit and a 4-bit code with the bits reversed
        (bit 1 is the least significant and bit 4 the most
        so 0x1 is 1000 and 0x8 is 0001) */
-     static const _Bool bits[16][5] =
-     {
-           {0, 0, 0, 0, 0}, {0, 1, 0, 0, 0}, {0, 0, 1, 0, 0}, {0, 1, 1, 0, 0},
-           {0, 0, 0, 1, 0}, {0, 1, 0, 1, 0}, {0, 0, 1, 1, 0}, {0, 1, 1, 1, 0},
-           {0, 0, 0, 0, 1}, {0, 1, 0, 0, 1}, {0, 0, 1, 0, 1}, {0, 1, 1, 0, 1},
-           {0, 0, 0, 1, 1}, {0, 1, 0, 1, 1}, {0, 0, 1, 1, 1}, {0, 1, 1, 1, 1}
-     };
+    static const _Bool bits[16][5] =
+    {
+        {0, 0, 0, 0, 0}, {0, 1, 0, 0, 0}, {0, 0, 1, 0, 0}, {0, 1, 1, 0, 0},
+        {0, 0, 0, 1, 0}, {0, 1, 0, 1, 0}, {0, 0, 1, 1, 0}, {0, 1, 1, 1, 0},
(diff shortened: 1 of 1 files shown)


CWehli:
@Bob-IT, @jralls: I checked the correction and the TAN generation continues to work.

jralls:
@Bob-IT Good catch. Let's go one further and replace
```

     /* prepend synchronization identifier */
     const char pre[] = {'0', 'F', 'F', 'F'};
     guint challenge_length;
     size_t len = sizeof (pre) + strlen (challenge) + 1;
     char* code = (char*)g_malloc0 (len);
     memcpy (code, pre, sizeof (pre));
     memcpy (code + sizeof (pre), challenge, strlen (challenge));
```
with
```
     /* prepend synchronization identifier */
    char* code = g_strdup_printf ("0FFF%s", challenge);
    guint challenge_length = strlen (code);
```

CWehli:
@Bob-IT Still works if you also replace 'len' with 'challenge_length' in the for loop.

Bob-IT:
@CWehli I have made the suggested changes can you make sure they are still OK before I push tomorrow.
I forgot to change the commit text so will do that before I push.

gnucash/import-export/aqb/gnc-flicker-gui.c
```diff
@@ -114,38 +114,37 @@ get_num (const char ch)
...
-     for (guint i = 0; i < len - 1; i += 2)
-     {
-         guint val1 = get_num (code[i]);
-         guint val2 = get_num (code[i+1]);
-
-         memcpy (&bitarray[i], bits[val2], sizeof(bits[val2]));
-         memcpy (&bitarray[i+1], bits[val1], sizeof(bits[val1]));
-     }
-     return code;
+    for (guint i = 0; i < challenge_length - 1; i += 2)
```
jralls, line 137: I think this will leave off the last character as `strlen` doesn't count the terminal `'\0'`.

CWehli: @Bob-IT, @jralls: I have made bank transfers and the TAN is accepted.

I also changed the counter from "challenge_length-1" to "challenge_length-0" and the generation of the TAN still works.

Bob-IT: @CWehli So just need to change it to challenge_length, if that is correct I will do that tomorrow before I push.

jralls: > I also changed the counter from "challenge_length-1" to "challenge_length-0" and the generation of the TAN still works.

It's interesting that it worked without the last character. It makes one wonder if there's some padding included in the challenge.

CWehli:
@jralls , @Bob-IT : So, I have tested it again.
The last two numbers are a check digit. If I shorten "challenge_length" by 0 or 1 in the FOR loop of the counter, the TAN generation works correctly. If I subtract 2 then it no longer works. 

I suggest running the FOR loop up to the entire length of strlen(), i.e. using "for (i = 0, i < challenge_length, i += 2)".

jralls:
@CWehli Makes sense, thanks.

Bob-IT:
@CWehli Thank you for testing. In an unrelated matter, do you suffer from bug https://bugs.gnucash.org/show_bug.cgi?id=768547
I added a possible fix but can not test, no worries if not affected.

CWehli:
@Bob-IT : mhh, I don't remember any such behaviour. At least your improvement does not seem to have any negative effects.

PR #842, opened 2020-12-16
L10N:it:5545 messaggi tradotti.

MocioF:
Nuova traduzione italiana, per Gnucash-4.3


MocioF:
Please wait for new commit. This version is not the last one

fellen:
> Please wait for new commit. This version is not the last one

I saw this too late. 
Not sure, if we could have avoided this, if you had delete the branch.
Just in case, rebase. 
I will now look in your next PR.

PR #843, opened 2020-12-16
Nuova traduzione italiana e glossario aggiornato

MocioF:
Nuova traduzione italiana, per Gnucash-4.3

### DIFF
--- a/po/glossary/it.po
+++ b/po/glossary/it.po
@@ -15,13 +15,12 @@
 #
 msgid ""
 msgstr ""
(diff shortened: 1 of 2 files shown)


jralls:
@MocioF can you rebase out the first two commits or is that beyond your git knowledge?

fellen:
Commited as commit 49a47cc and commit 4a63dc7, thanks.

Do not remove keys (Language Team) from the first record, but empty lines are allowed.

PR #845, opened 2020-12-23
QR code in invoice

christopherlam:
after `$sudo apt install qrencode`:

`#include <qrencode.h>` does not seem to work:

```
/usr/bin/ld: lib/gnucash/libgnc-gnome-utils.so: undefined reference to `QRcode_encodeString'
/usr/bin/ld: lib/gnucash/libgnc-gnome-utils.so: undefined reference to `QRcode_free'
```


jralls:
That would be because you didn't tell the linker where to find the symbol. You do that by adding `libqrencode` to gnc-gnome-utils's `target_link_libraries` command in CMakeLists.txt.

You also need to check for qrencode being available in the root CMakeLists.txt, update Readme.dependencies in the root directory and add module installation in gnucash-on-windows, gnucash-on-osx, and gnucash-on-flatpak.

fellen:
> You also need to check for qrencode being available in the root CMakeLists.txt, update Readme.dependencies in the root directory and add module installation in gnucash-on-windows, gnucash-on-osx, and gnucash-on-flatpak.

It seems to be in most important distributions: https://repology.org/project/qrencode/versions.
As changing dependencies in maint is not recommend, you could also consider to make it optional.

christopherlam:
This requires `gcc -lqrencode` - how to add linker flag?

christopherlam:
Will it be:

```cmake
if (FOUND_QRENCODE)
  set (CMAKE_C_FLAGS "-lqrencode ${CMAKE_C_FLAGS}")
endif ()
```

jralls:
No, you add it to the list in the  target's `target_link_libraries` call. Since qrencode is also built with CMake you shouldn't need the  `-l`, you should be able to use `qrencode`.

christopherlam:
This branch has the C glue code to call `libqrencode`
It also has guile code to interpret binary to SVG.

Missing:
- [ ] appropriate CMake flags to link `qrencode` when compiling `gnc-menu-extensions.c` -- we'd eventually use a separate file
- [ ] template for string to incorporate information (eg invoice amounts, bank/routing numbers etc) to send to libqrencode and receive data

christopherlam:
@gjanssens this is now functional.
See the string-template is hard-coded.
The EPC QR code https://en.wikipedia.org/wiki/EPC_QR_code standard requires IBAN which is not currently a gnucash option.
My thought: introduce IBAN book company option, and introduce company QR template with templated language eg `{IBAN} {CURRENCY_ISO} {TOTAL}` fields as another company option.
Any other ideas?

gjanssens:
This will become a nice addition to our invoice printing functionality.

> @gjanssens this is now functional.
> See the string-template is hard-coded.
> The EPC QR code https://en.wikipedia.org/wiki/EPC_QR_code standard requires IBAN which is not currently a gnucash option.
> My thought: introduce IBAN book company option, and introduce company QR template with templated language eg `{IBAN} {CURRENCY_ISO} {TOTAL}` fields as another company option.
> Any other ideas?

There are two bits of required information that are currently not stored: IBAN and BIC. I agree they can be part of the book company info. The currency should be invoice currency. We don't need to store that as a company parameter.

From the specification it seems to me the only free form fields in the QR template are the reference and information fields. All other fields have strictly defined content. Unless you are considering other formats which follow different specifications ? In that case we will probably want a templating system similar to the check printing one. GnuCash can provide a number of standard formats and allow the user to create custom ones. But I'd wait with that until the need arises.
For now it's a matter of deciding what information we want in the two "freeform" fields. For reference I'd propose to use invoice number. For information we could use invoice notes ? Feel free to consider options of course.

As for the technical side of this PR:
I agree with @fellen we should not add required dependencies on maint. The program should continue build and work without the QR functionality if the library is missing. That means you will need to conditionally enable or disable this feature based on the presence of the library. In theory you have build time or run time dependencies and one could make either optional. In practice however I think we can claim that if the program is built with qrencode it requires qrencode to run. So you should tweak our build system such that it tests for the presence of the qrencode package. If present, build with the qr code functionality enabled. If missing disable the qr code functionality and print a message about this. That give builders interested in the functionality a hint to how they can enable it.

A few hints:
* the QREncode library exposes a package config file, so you can use `pkg_check_modules (QRENCODE IMPORTED_TARGET qrencode)` to test for the presence of the library (see [pkg_check_modules](https://cmake.org/cmake/help/latest/module/FindPkgConfig.html?highlight=pkg_check_modules) This also sets a number of environment variables you can use later in the build system.
* it will probably make your life easier if you move the C function that uses the QREncode library into a separate source file. If you do you can conditionally include that file as a source file in CMakeLists.txt based on the value of the generated variable QRENCODE_FOUND.
* Similarly you can use the generated import target PkgConfig::QRENCODE to set the proper linker details in target_link_libraries.
* It gets a bit nastier from there: you can only expose this function to scheme if it's defined of course, so in gnome-utils.i you'll have to check somehow if the function is available before wrapping it. I don't think we have code in place already to cater for such a check. A similar trick could be used here: use a separate interface file to expose the qrencode function, only generate that wrapper based on the value of QRENCODE_FOUND and only add the generated swig file as a source file based on the value of QRENCODE_FOUND.
* Similarly, the option you have added to invoice.scm to present a qr code should only be enabled (or visible) if gnucash is built with qrcode support. If we manage to fix the gnome-utils.i file test above you could use that file to expose a variable to scheme code you can test on for the presence of the qr encode functionality. I suppose scheme can even test for the presence of a function directly. So you could test for the existence of guile function `gnc-qrcode-encodestring` directly and enable/disable or show/hide the option based on the result.

As you can see, quite a few bits to change in order to make this an optional feature. If that's too much, the alternative is to only introduce this in master. I am totally fine with adding a new mandatory dependency there.

Note also that regardless of your target branch you can only make this mandatory when our build systems for MacOS, Windows and Flatpak are updated to also install or build qrencode.

gjanssens:
And in addition our ci scripts need updates to install qrencode before building gnucash. If we go the optional route, I'd only install qrencode in a few of our build environments to check if test succeed both with or without qrencode.

christopherlam:
> This will become a nice addition to our invoice printing functionality.
> 
> There are two bits of required information that are currently not stored: IBAN and BIC. I agree they can be part of the book company info. The currency should be invoice currency. We don't need to store that as a company parameter.

Ok yes two string options must be created. Note IBAN maximum 34 chars, BIC max 11 chars.

> From the specification it seems to me the only free form fields in the QR template are the reference and information fields. All other fields have strictly defined content. Unless you are considering other formats which follow different specifications ? In that case we will probably want a templating system similar to the check printing one. GnuCash can provide a number of standard formats and allow the user to create custom ones. But I'd wait with that until the need arises.
> For now it's a matter of deciding what information we want in the two "freeform" fields. For reference I'd propose to use invoice number. For information we could use invoice notes ? Feel free to consider options of course.

From my understanding it's reference _or_ information. For reference invoice ID seems natural.

> As for the technical side of this PR:
> I agree with @fellen we should not add required dependencies on maint. The program should continue build and work without the QR functionality if the library is missing. That means you will need to conditionally enable or disable this feature based on the presence of the library. In theory you have build time or run time dependencies and one could make either optional. In practice however I think we can claim that if the program is built with qrencode it requires qrencode to run. So you should tweak our build system such that it tests for the presence of the qrencode package. If present, build with the qr code functionality enabled. If missing disable the qr code functionality and print a message about this. That give builders interested in the functionality a hint to how they can enable it.
> 
> A few hints:
> (snip)

I'd suggest `gnc-qrencode.c` and `gnc-qrencode-stub.c` -- both expose `gnc_qrcode_encodestring` but the latter will always return `SCM_BOOL_F`

```cmake
if (HAVE_QRENCODE)
  target_link_libraries (qrencode)
  add_library (qr-encode.c)
else ()
  add_library (qr-encode-stub.c)
endif ()
```
>  Similarly, the option you have added to invoice.scm to present a qr code should only be enabled (or visible) if gnucash is built with qrcode support. If we manage to fix the gnome-utils.i file test above you could use that file to expose a variable to scheme code you can test on for the presence of the qr encode functionality. I suppose scheme can even test for the presence of a function directly. So you could test for the existence of guile function `gnc-qrcode-encodestring` directly and enable/disable or show/hide the option based on the result.

For simplicity I'd vote to _allow_ QR option in invoice.scm (otherwise saved-reports etc will complain), and if `gnc-qrcode-encodestring` does not return invalid data then the header will have textual 'QRencode library not available' with output to tracefile.

gjanssens:
> > From the specification it seems to me the only free form fields in the QR template are the reference and information fields. All other fields have strictly defined content. Unless you are considering other formats which follow different specifications ? In that case we will probably want a templating system similar to the check printing one. GnuCash can provide a number of standard formats and allow the user to create custom ones. But I'd wait with that until the need arises.
> > For now it's a matter of deciding what information we want in the two "freeform" fields. For reference I'd propose to use invoice number. For information we could use invoice notes ? Feel free to consider options of course.
> 
> From my understanding it's reference _or_ information. For reference invoice ID seems natural.
> 
No, there are actually two supported reference types, structured or unstructured. And in addition you  can have a free form information field. The structured reference follows the [ISO 11649](https://en.wikipedia.org/wiki/Creditor_Reference) format.

> I'd suggest `gnc-qrencode.c` and `gnc-qrencode-stub.c` -- both expose `gnc_qrcode_encodestring` but the latter will always return `SCM_BOOL_F`
> 
> ```cmake
> if (HAVE_QRENCODE)
>   target_link_libraries (qrencode)
>   add_library (qr-encode.c)
> else ()
>   add_library (qr-encode-stub.c)
> endif ()
> ```
> 

That's ok in principle. I assume you were using pseudocode here ?

> > Similarly, the option you have added to invoice.scm to present a qr code should only be enabled (or visible) if gnucash is built with qrcode support. If we manage to fix the gnome-utils.i file test above you could use that file to expose a variable to scheme code you can test on for the presence of the qr encode functionality. I suppose scheme can even test for the presence of a function directly. So you could test for the existence of guile function `gnc-qrcode-encodestring` directly and enable/disable or show/hide the option based on the result.
> 
> For simplicity I'd vote to _allow_ QR option in invoice.scm (otherwise saved-reports etc will complain), and if `gnc-qrcode-encodestring` does not return valid QR data then the header will have textual 'QRencode library not available' with output to tracefile.

Is it technically difficult to hide the option when qrcode is not available ? I don't mean make it non-existent, just hidden. The other option is to keep it around but disable it when qrencode is not found. That way it still exists but the user can't set it. A tooltip would then be helpful to explain why the option is disabled. The main difference between your and my approach is that in your case the error message only appears *after* the user has configured the report, which leads to frustration. If you can't set the option in the first place that saves the user a roundtrip to the options if it was set incorrectly.

christopherlam:
Have moved into `gnc-qrencode.c` and `gnc-encode-stub.c`. Both of them import `gnc-qrencode.h` which defines the exported function. For simplicity I've changed the 'no libqrencode' return value to `SCM_EOL`.

The invoice report currently places QR code in invoice header, as a new multichoice option for date/company-details/customer-details/picture. Alternatively we could add a _new_ option "QR code" which will obligatorily show in top-right corner, and whose multichoice options are 'None' 'EPC QR Code' for further expansion. Having a _new_ option will be much easier to selectively hide when libqrencode is not available.

I can't figure out CMakeLists changes to load either `qrencode package` + `gnc-qrencode.c` _or_ `gnc-qrencode-stub.c`. Assistance requested...

gnucash/gnome-utils/gnc-qrencode.c
```diff
@@ -0,0 +1,53 @@
...
+    QRcode *qr = NULL;
+    SCM output = SCM_EOL;
+
+    qr = QRcode_encodeString (str, 0, QR_ECLEVEL_M, QR_MODE_8, FALSE);
+
+    if (!qr)
+    {
+        perror ("QRcode_encodeString() failed");
+        goto ret;
+    }
```
christopherlam, line 40: @jralls is `perror` an acceptable error-dump mechanism here?

jralls: No, it's [for printing strings based on system `ERRNO`](https://man7.org/linux/man-pages/man3/perror.3.html), plus it writes to `stderr` and that will just go to `/dev/null` unless the user happens to have started GnuCash from the command line.  Use `PERR`.

gnucash/report/CMakeLists.txt
```diff
@@ -17,9 +17,21 @@ set (report_SOURCES
...
+set (qrencode_HEADERS gnc-qrencode.h)
+
+if (HAVE_QRENCODE)
+  set (qrencode_SOURCES gnc-qrencode.c)
+  set (qrencode_LIBRARY PkgConfig::QRENCODE)
+else ()
+  set (qrencode_SOURCES gnc-qrencode-stub.c)
+  set (qrencode_LIBRARY)
+endif ()
+
```
jralls, line 29: You need to add the new files to the `local_dist` list at the bottom so that they're included in the tarball. Since qrencode_SOURCES is variable you have to name the two files so that they're both added.

libgnucash/app-utils/business-options.scm
```diff
@@ -73,6 +75,8 @@
 (define gnc:*company-url* (N_ "Company Website URL"))
 (define gnc:*company-email* (N_ "Company Email Address"))
 (define gnc:*company-contact* (N_ "Company Contact Person"))
+(define gnc:*company-iban* (N_ "Company IBAN"))
+(define gnc:*company-bic* (N_ "Company BIC"))
```
fellen, line 79: If no IBAN is provided, you need BIC + traditional Bank Account Number for a bank transaction, so I would add it here, too. Optionally also the name (including branch) of the bank.
It is common to add this infos to the footer of a business letter.

[IBAN](https://en.wikipedia.org/wiki/International_Bank_Account_Number#Basic_Bank_Account_Number)
SWIFT-BIC: [ISO 9362](https://en.wikipedia.org/wiki/ISO_9362) and links.

gnucash/report/reports/standard/invoice.scm
```diff
@@ -346,7 +346,7 @@ for styling the invoice. Please see the exported report for the CSS class names.
   (gnc:register-inv-option
    (gnc:make-multichoice-option
     (N_ "Display") (N_ "QR Code")
-    "te" "QR Code"
+    "te" (if (gnc-qrcode-available) "QR Code" "QR Code (not available on this system")
```
jralls, line 349: >"QR Code (not available on this system"

S.B. "QR Code (not available on this system)"
but how about "QR Code (install libqrcode to enable this option)"

And of course both need to be marked for translation.

christopherlam: I may be wrong; it's not sufficient to install libqrencode on the local system isn't it? The build must be initiated with libqrencode present?

jralls: You're right, the message should say something like "rebuild with libqrcode to enable this option". That's not particularly useful for users. I'd suggest that it could be a mandatory dependency if you could accept qrencode 3.4.4; the requirement for >=4.0 locks out Ubuntu18.04. All of the other distros worth caring about [have at least 4.0](https://repology.org/project/qrencode/versions). OTOH it's kind of rude to stuff a new dependency in a stable release series.

christopherlam: https://fukuchi.org/works/qrencode/#news 3.4.4 should be fine -- main benefit for 4.0.0 is memory efficiency. The win32/mac/flatpak build scripts could be amended in maint, and only suggested in Linux builds (and mandatory for 5.x).

libgnucash/app-utils/business-options.scm
christopherlam: You mean we offer the new options IBAN, BIC, accnumber, and:
1. QR encoder picks up IBAN + BIC
2. invoice report prints IBAN & BIC+accnumber?

Otherwise what's the purpose of having acc-number?

fellen: I made a sketch with many of the relations in [Payment Routing](https://wiki.gnucash.org/wiki/Payment_Routing).

fellen: In the example in https://wiki.gnucash.org/wiki/File:Invoice_Option_Display.png I offer 3 bank connections. So you would better create a QR-code for each of them and disply it  together with the textual form at the bottom of the page:
```
Pay via
Bank: DeuBa VoBa SpK
IBAN: …
QR: …
```

christopherlam: I seriously don't think it's reasonable to require or display multiple QR codes.

fellen: > I seriously don't think it's reasonable to require or display multiple QR codes.

Scheme is list processing. Take the invoice amount and the list of IBANS (in the future) from the book properties to generate the payment section of the invoice.

The other problem: If there is a payment term like 2-10-30 (2% discount on payments in ten days), you have to generate 2 lists: for 98% and 100% of the invoice amount.

gjanssens: > In the example in https://wiki.gnucash.org/wiki/File:Invoice_Option_Display.png I offer 3 bank connections. So you would better create a QR-code for each of them and display it together with the textual form at the bottom of the page:
> 
> ```
> Pay via
> Bank: DeuBa VoBa SpK
> IBAN: …
> QR: …
> ```

A few thoughts to avoid dragging this proposal forever.

I would propose to to focus on implementing one specific scenario first and consider expanding it in future iterations based on user feedback. From reading the comments on this PR it looks like the current scenario is to implement EPC QR. I think that's a good start and it requires an IBAN account option and can have BIC as optional parameter. I think that's a good goal to reach first.

Before we go and implement variants with multiple bank account identifiers I would ask around for real world examples of invoices printing a QR code. IMO even if a company has multiple bank accounts, you can still pick one as the preferred one and use that for a QR code. Adding multiple QR codes to an invoice sounds fairly clumsy to me and a source of confusion to the the customer. I'm aware I say that while I have never ever seen a real invoice with QR codes yet so I'd like to have real world examples (even if obfuscated).

I also don't know if the comments about national bank account numbers and bank identifiers apply to this PR. The EPC QR specifically states it uses IBAN numbers,. So while it may be nice to print those traditional numbers and identifiers on your invoice they shouldn't be used to print QR codes. Unless there's another standard supporting that use case. Again if so, that could be a future improvement. Adding the ability to print the traditional bank numbers on your invoices should also be done in a separate PR as it will probably take a few layout iterations to work for many cases.

gjanssens: > The other problem: If there is a payment term like 2-10-30 (2% discount on payments in ten days), you have to generate 2 lists: for 98% and 100% of the invoice amount.

Same thing here. I don't know how this is handled in the real world today. I would probably not put QR codes at all if the payment amount depends on other factors or I would add one single QR code for the most common case. Real world samples would help a lot in this area.

gnucash/report/reports/standard/invoice.scm
gjanssens: FWIW I would not make libqrcode a new dependency on maint.

As for the case of missing libqrcode,  I still think the option should not be present at all if gnucash is not built with libqrcode. It's missing in previous 4.x release as well. Considering we want backwards compatibility in stable releases, the code  should still work if the option is not defined.

Imagine someone creates an invoice with QR code enabled in gnucash 4.5 and loads that report in 4.4 (on another system for example). We want that report to still show though ignoring the QR code option. I think the whole 4.x series should behave that way. Though you could add a helpful message to the user starting from the 4.5 release. Such message should probably only be  shown if they run gnucash 4.5 without qr code support and the saved invoice had the option enabled. In all other situations either the option works as advertised (gnucash 4.5+ built with libqrcode) or it is as if the option never existed (gnucash 4.5+ built without libqrcode  and loaded report didn't have option set in the first place).
So that message depends on the build configuration and could suggest something like "Report was configured with QR code enabled but this edition of gnucash lacks support for it. Please contact your package maintainer to request QR code support."
The tricky part is to display it to the user without it being printed on the actual invoice. Perhaps make it @media: screen only ?

christopherlam: The way it's currently written is imho "safe". Upgrading 4.4 to 4.5-with-qrencode will show new option working well. Now user saves such a report, and reverts to 4.4 will now have neither QR nor error. Upgrading to 4.5-without-qrencode and loading saved report will dump a small warning on tracefile but otherwise functional report. i.e. at worst a small warning if 4.5-without-qrencode encounters a saved-report requesting QR.

There's a printed QR error only if an invalid parameter exists eg IBAN > 34 characters. Or amount too large.

Whether the report options disable or hide QR option is another small UI issue. I have no particular preference.

Displaying QR warnings (missing QR library) on screen vs printed is an interesting trick.

gjanssens: Oh, I agree it's  safe.  My comments were only a suggestion to slightly improve the user experience. Not mandatory...

libgnucash/app-utils/business-options.scm
fellen: I would also now only implement something like
```
<bankconnections>
  <bankdetails>
    <bankname></bankname>
    <IBAN></IBAN>
    <BIC></BIC>
  </bankdetails>
</bankconnections>

```
with one instance in book company info.

About the layout.
1. AFAIK the size of a stamp is sufficient . In CH 46mm, 5mm border
2. I would put it below the sum as it is part of the content.
3. Alternative it can be printed as part of a regular transfer form 
[zahlschein.pdf](https://github.com/Gnucash/gnucash/files/5827916/zahlschein.pdf)
at the bottom or on the next page.

christopherlam: @fellen you know full well that the set of company options is currently a global pref and is out of scope for this PR. Please liaise with @jralls or @gjanssens how to move company options from global to book. This is not my game.

fellen: > the set of company options is currently a global pref

No,
image

christopherlam: This is still a major change in the datafile format, therefore out of scope for this bug.

fellen: 1. it is no bug, but an enhancement request.
2. There is a new dependency.
3. There is a change in the file structure.
So IMHO you should add the ability to read a file with IBAN … to maint and the rest in master.

jralls: @fellen That's absurd. The business options exist solely to create invoices. This use of IBAN/BIC is solely to put a QR on invoices. Business options are stored in KVP so no format changes and no compatibility worries.

The only issue with this PR is about the dependency.

fellen:
> And in addition our ci scripts need updates to install qrencode before building gnucash. If we go the optional route, I'd only install qrencode in a few of our build environments to check if test succeed both with or without qrencode.

@gjanssens how about converting your tips from this PR into a wiki page Adding_Dependencies?

jralls:
How about a one-line wiki page:
<h1>Adding Dependencies</h1>
<p>Don't.</p>

We already have too many dependencies and every added one makes GnuCash harder to build and more dependent on the whims of distribution packagers and in the case of single-contributor libraries like [libqrencode](https://github.com/fukuchi/libqrencode) the whims and survival of that one contributor. Note that it's been almost 10 months since the last commit to that repo.

fellen:
>> Adding Dependencies
> 
> Don't.

I hoped it would scare off in most cases. ;-)

fellen:
https://lists.gnucash.org/pipermail/gnucash-de/2022-February/012321.html:
From 1. Oktober 2022 QR codes are also required in CH.

christopherlam:
Here's one of the QR contents from the PDF. They also mandate precise layout containing the QR code.

```
SPC
0200
1
CH4431999123000889012
S
Max Muster & Söhne
Musterstrasse
123
8000
Seldwyla
CH







2500.25
CHF
S
Simon Muster
Musterstrasse 
1
8000
Seldwyla
CH
QRR
210000000003139471430009017

EPD

Name AV1: UV;UltraPay005;12345
Name AV2: XY;XYService;54321
```

fellen:
I assume you are talking about
https://www.paymentstandards.ch/dam/downloads/style-guide-en.pdf

They put the payment part below the invoice or on the next page. That is also common here in DE. The german form descriptions are in
https://die-dk.de/media/files/Richtlinie-ZV-Vordrucke-2016-DK_finale_Fassung_DK_Homepage.pdf with an update 
https://die-dk.de/media/files/Merkblatt_2017-Anpassung_der_Richtlinien_V11.01.2018_QOulHUn.pdf
QR code is almost at the end of the main document described.

As feared the formats are very different.

christopherlam:
> As feared the formats are very different.

Not a problem in this branch. The format it's extensible by modifying code, or even maybe in the future via options.

fellen:
> Not a problem in this branch. The format it's extensible by modifying code, or even maybe in the future via options.

Perhaps the printing of the payment part is easier integrated in the cheque printing facility. Most german and also other european bank forms are based on an A6 landscape form.

luzpaz:
What's the status on this PR. This is a fantastic feature

luzpaz:
bump

jralls:
The options parts need to be rewritten in C++.
[libqrencode](https://github.com/fukuchi/libqrencode) appears to have been abandoned with the [last commit 2020-09-27](https://github.com/fukuchi/libqrencode/commit/715e29fd4cd71b6e452ae0f4e36d917b43122ce8). We should consider taking the source into borrowed and including its test suite in our CI.

christopherlam:
QRcode_encodeString 4.11 has changed something and this isn't producing valid QR data anymore. Anyway rebased onto master, and using c++options. If libqrencode is added into borrowed then I'll try debug again.

jralls:
Putting this back to draft. libqrencode is surely abandoned, no new commits, 50 open bugs, and 16 open PRs. It occurs to me that a javascript one might be a better fit since we're drawing the QR on a report. [This one](https://www.npmjs.com/package/qrcode) came out at the top of a Google search.

PR #846, opened 2020-12-24
[app-utils] fix dependency chain

christopherlam:
I think this works?

business-prefs needs options and business-options
business-options needs options
options needs date-utilities

### DIFF
--- a/libgnucash/app-utils/CMakeLists.txt
+++ b/libgnucash/app-utils/CMakeLists.txt
@@ -169,11 +169,21 @@ install(FILES ${app_utils_HEADERS} DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/gnuca
 # And now handle scheme files
 
 set (app_utils_SCHEME_1
-    business-options.scm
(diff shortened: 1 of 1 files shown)


gjanssens:
Note ci is broken, but that's not due to this PR, that's because of the commit you pushed yesterday to deprecate gnc:make-string-database.

Other than that your PR should be ok.

Although as a last step you have to fix the generation of our  dist tarball: you have altered the source sets (and chose not to use variables for that). All source files in a given directory that need to end up in the distribution tarball have to appear in the set_local_dist function near the end of the CMakeLists.txt file.

As you have removed a few source files from variable app_utils_SCHEME_1 they will no longer be distributed, meaning our dist tarball is broken.

Note you already  broke it in your modularization effort earlier: there you split off scm-business-prefs.scm from the variable, dropping it from the dist tarball.

So please add the files no longer part of source environment variables to set_local_dist, Then you should be fine.

gjanssens (approved):
Looks good to me. :+1:

PR #848, opened 2020-12-27
Fix QIF-IMPORT dependency chain

christopherlam:
* fix dependency chain in qif-import
* some functions needed to be moved to resolve circular dependencies
* [ ] to test successful qif import

For initial review.


jralls:
You need to add the new `qif_import_SCHEME_X` lists to the `set_local_dist` call at the bottom of `CMakeLists.txt` so that they get included in the tarball.

christopherlam:
qif-merge-groups cannot find qif-utils. Can't figure out why; CMakeLists seems ok to me.

christopherlam:
too difficult to unravel circular dependencies. MAKE_LINKS in cmakelists is an acceptable solution for now.

PR #855, opened 2021-01-01
Remove deprecated functions from master.

christopherlam:
Mostly scheme code, and from Account.cpp and gnc-report.c.


jralls:
@christopherlam Looks fine.

christopherlam:
Ok there are more `gnc_add_scheme_deprecated_module` around but not 100% sure I'm capable of removing the deprecated modules completely.

jralls:
Because they're still in use?

christopherlam:
Because removing the `gnc_add_scheme_deprecated_module` statement doesn't necessarily mean the module source file is really visible?

jralls:
Well, if you've removed the `module` declaration then it can't be `use`d, though it can still be `load`ed. If it's neither `use`d nor `load`ed then it's dead. All that remains is the housekeeping of deleting the file (`git rm filename`) and the references to it in CMakeLists.txt.

gjanssens:
> Because removing the `gnc_add_scheme_deprecated_module` statement doesn't necessarily mean the module source file is really removed?

What  do you mean with this ? As far as I know there is no module source file being distributed for a  deprecated module created by `gnc_add_scheme_deprecated_module`. If you remove the  function call, no source file for the deprecated module will be generated, so  nothing will have to be removed. Or are you thinking of something else ?

christopherlam:
Not thinking of anything else. Sounds like it's safe to remove all such functions.

gjanssens:
> Not thinking of anything else. Sounds like it's safe to remove all such functions.

Yes, with one exception. We create two dummy deprecated modules to test whether `gnc_add_scheme_deprecated_module` itself works as expected. These should be kept to ensure that function continues to work properly in the  future.

PR #857, opened 2021-01-02
Don't use gnc:last-captured-error

christopherlam:
1. To avoid using the global var `gnc:last-captured-error`, this branch will augment some C functions and handle error by passing a parameter to receive the captured error.
2. ~Also `gnc_set_busy_cursor` is currently disabled -- they seem to belong to `gnc-ui.h` but importing the latter doesn't work.~ (fixed)

### DIFF
--- a/gnucash/gnome/gnc-plugin-report-system.c
+++ b/gnucash/gnome/gnc-plugin-report-system.c
@@ -137,14 +137,12 @@ static gboolean
 gnc_report_system_report_stream_cb (const char *location, char ** data, int *len)
 {
     gboolean ok;
+    gchar *captured_str;
 
-    ok = gnc_run_report_id_string (location, data);
+    ok = gnc_run_report_id_string_with_error_handling (location, data, &captured_str);
 
     if (!ok)
     {
-        SCM captured = scm_c_eval_string ("gnc:last-captured-error");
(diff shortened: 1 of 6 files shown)


gnucash/gnome/gnc-plugin-report-system.c
```diff
@@ -137,14 +138,14 @@ static gboolean
 gnc_report_system_report_stream_cb (const char *location, char ** data, int *len)
 {
     gboolean ok;
+    gchar *captured_str;
 
-    ok = gnc_run_report_id_string (location, data);
+    gnc_set_busy_cursor (NULL, TRUE);
+    ok = gnc_run_report_id_string_with_error_handling (location, data, &captured_str);
+    gnc_unset_busy_cursor (NULL);
```
jralls, line 146: Doesn't running a report run a progress bar?

jralls (approved):
Looks good, though you shouldn't set the busy cursor if there's a progress bar running.

gnucash/gnome/gnc-plugin-report-system.c
christopherlam: Correct, most report update a progress bar via `gnc:report-percent-done`. Will remove busy cursor.

PR #859, opened 2021-01-03
Some spelling errors fixed

comradekingu:


jralls:
Please separate all of the ellipsis changes into a separate PR.

Setup is a word in English. Those changes are rejected.

That leaves the "non-Bayesian" change. I've committed that separately.

comradekingu:
@jrails They are both words, and they mean different things. Specifically it "setup" is not a verb, whereas "set up" is.

For example "Set up relevant accounts" means "to set up", and "Setup relevant accounts" means accounts that "setup" is relevant to.

jralls:
@comradekingu I've had about enough of your pidgin-English sea lawyering. That distinction does not exist in modern American English usage. BTW, your second example requires the hyphen that you're so fond of scattering about: "setup-relevant" for relative to setup.

It is not possible to make changes to source files in Weblate: Github is not the canonical repository and we're not about to grant a public service access to our servers.

gnucash/gnome/gnc-plugin-basic-commands.c
```diff
@@ -113,7 +113,7 @@ static GtkActionEntry gnc_plugin_actions [] =
         G_CALLBACK (gnc_main_window_cmd_file_new)
     },
     {
-        "FileOpenAction", "document-open", N_("_Open..."), "<primary>o",
+        "FileOpenAction", "document-open", N_("_Open…"), "<primary>o",
```
fellen, line 116: About the usage of ellipsis see the discussion in https://bugs.gnucash.org/show_bug.cgi?id=792975

fellen:
@comradekingu FYI for the program most of the current development is done on maint 4.x series, while master will lead to 5.0. See https://wiki.gnucash.org/wiki/Release_Schedule

comradekingu:
@jralls In considering the distinction arbitrary, what is there to lose from dropping only the combinations that produce the error?
Is "modern" and "American" and non-perscriptive langauge a goal in and of itself? "Set up" vs "Setup", where the implied counterpart is "Don't set up" vs "Non-setup" warrants exclusivity IMO.

It isn't a personal goal of mine to possibly enlighten you, me, or both; at your or my expense, so I'll leave it at that.
From my poor understanding of English, it seems I could improve GnuCash a lot, yet objectively the push-back would suggest any level of change is likely to make matters worse.

jralls:
Yes, the C locale is American English and the strings provided for translation are the ones presented to American users so it is  as important that those strings be idiomatic American English as it is that your translations should be idiomatic Norwegian. So "modern" and "American" is a goal.

That said, many GnuCash contributors are like you not native English speakers and English is anyway a very broad language with lots of regional and generational variation. As long as strings effectively communicate what they're supposed to and don't mangle the language too badly we generally don't comment on them or correct them, simply to avoid the sorts of silly arguments that seem to delight you and annoy just about everyone else. Annoying contributors about their English is not a good way to get them to contribute C, especially because we must be pedantic about the C. That makes one meaning of "non-prescriptive"--that we don't prescribe strict adherence to textbook English that's anyway seldom used--also a goal.

Your proposed changes have all been pedantic; many of them have also been wrong. Even when correct there's little gained from pedantic changes: They don't make the English text more understandable and they make work for all translators by making them clear fuzzies to no useful end.

fellen:
> That leaves the "non-Bayesian" change. I've committed that separately.

… and as result of changing one letter, msgmerge produced 16GB diff noise.

jralls:
Wow. Are you sure you msgmerged the right branch? Should we revert the commit since it's of questionable value?

fellen:
> Are you sure you msgmerged the right branch?

I had cherry picked it into maint.
I got the message, that the patch notification awaits moderation as it had the tenfold of the size limit.

It was our first msgmerge since the string freeze, so many refs changed their line number in over 3 weeks.
> Should we revert the commit since it's of questionable value?

Not another big blob of noise.

jralls:
> It was our first msgmerge since the string freeze, so many refs changed their line number in over 3 weeks.

Right, nothing to do with this.

PR #863, opened 2021-01-03
I18N: glossary: New Term "Dashboard"

CWehli:


po/glossary/gnc-glossary.txt
```diff
@@ -76,6 +76,7 @@
 "credit note"	"A document that you give to a client that says you owe money to the client, i.e. the opposite of an invoice"
 "currency"	"The system of money used in a country"
 "Custom"	"Custom print format (i.e. according to the user's wishes) as opposed to a template choice."
+"Dashboard"	"-"
```
fellen, line 79: Without an  explanation it does not make much sense.

fellen (changes requested):

po/glossary/gnc-glossary.txt
CWehli: @fellen: Done.

fellen: I should have searched it before: https://en.wikipedia.org/wiki/Dashboard_(disambiguation)
Is everything important included?

CWehli: @fellen: Is my suggestion so bad?

fellen: I don't know it. I didn't know the word and have not used the report. So I posted just an idea and will trust you.

fellen (commented):
Ich habe noch den passenden wp link angehangen.

PR #864, opened 2021-01-04
Bug 798073 - Fix sporadic failure on MacOSX

rhabacker:
Fix for bug https://bugs.gnucash.org/show_bug.cgi?id=798073

### DIFF
--- a/libgnucash/engine/test/utest-Transaction.cpp
+++ b/libgnucash/engine/test/utest-Transaction.cpp
@@ -908,6 +908,14 @@ test_xaccTransEqual (Fixture *fixture, gconstpointer pData)
     qof_instance_set_guid (split1, qof_instance_get_guid (split0));
     g_assert (!xaccTransEqual (clone, txn0, TRUE, TRUE, TRUE, TRUE));
     g_assert (xaccTransEqual (clone, txn0, TRUE, FALSE, FALSE, TRUE));
+#ifdef __APPLE__
+    if (check->hits == 12)
+        g_assert_cmpint (check->hits, ==, 12);
+    else
+        g_assert_cmpint (check->hits, ==, 10);
+#else
+    g_assert_cmpint (check->hits, ==, 10);
+#endif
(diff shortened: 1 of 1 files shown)


rhabacker:
On Mac the output of the test has a different format 
```
Linux:   ok 1 /engine/Account/gnc set account separator
Mac:  /engine/Account/gnc set account separator: OK
```
probably caused by a different version of the used test framework. 

Since the test in question shows the same results
Linux:
```
/engine/Transaction/xaccTransEqual: <INFO> (gnc.engine) [xaccSplitEqualCheckBal] balances differ: 300000/1000 vs 100000/1000
OK
```
Mac:
```
/engine/transaction/xaccTransEqual: <INFO> (gnc.engine) [xaccSplitEqualCheckBal] balances differ: 300000/1000 vs 100000/1000
OK
```
the problem looks like a different verbosity level or similar and can be ignored.

jralls:
Wrong output. It's counting the occurrences of the message "[xaccSplitEqual] GUIDs differ" and if they differ 12 times instead of 10 that's a real test failure.

rhabacker:
Thanks for the pointer. 

I have a question about understanding the output of the tests that I did not find answered at https://wiki.gnucash.org/wiki/Testing#.

> <INFO> (gnc.engine) [xaccSplitEqualCheckBal] balances differ: 300000/1000 vs 100000/1000

This message indicates a test error ?

jralls:
Depends on the test. It indicates that the two splits being checked for equality aren't equal because the balances are different. It's output from xaccSplitEqualCheckBal, not from the test function. GLib-based tests either write out "OK" like your examples above or "ERROR" with the details of the failed assert like what you pasted into the bug report. Everything else comes from the code under test and is logged to help understand the cause of the test failure.

jralls:
As noted above this PR takes the wrong approach. Besides, the macOS test failures seem to have been fixed by https://github.com/Gnucash/gnucash/commit/ffa2f606e38c96a64cbebd4204f3795d5dd7c22d.

PR #873, opened 2021-01-13
Mark most parent accounts as placeholders in Danish templates

Munksgaard:
Most of these, like "Udgifter", correspond directly to directly to en_GB
equivalents, like "Expenses". However, unlike in the en_GB version, the parent
accounts in the Danish templates are not marked as placeholders.

I have used my own judgement in a few cases where there was no direct equivalent
in the en_GB templates.

This was also discussed on the gnucash mailing list: https://lists.gnucash.org/pipermail/gnucash-user/2021-January/094762.html (and replies).

### DIFF
--- a/data/accounts/da/acctchrt_car.gnucash-xea
+++ b/data/accounts/da/acctchrt_car.gnucash-xea
@@ -56,6 +56,12 @@
   </act:commodity>
   <act:description>Udgifter</act:description>
   <act:parent type="new">1972cce2e2364f95b2b0bc014502661d</act:parent>
+  <act:slots>
+    <slot>
+      <slot:key>placeholder</slot:key>
+      <slot:value type="string">true</slot:value>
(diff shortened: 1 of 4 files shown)


fellen (commented):
- You missed "Afstemning" [Edit] and "Underholdning" [/Edit]. Was that intended?

- I just watched, da is only a subset of C (en_US) if you like you can translate/adjust the others, too.

Munksgaard:
>     * You missed "Afstemning" [Edit] and "Underholdning" [/Edit]. Was that intended?

No, I will create a new PR with those fixed as well.

> * I just watched, da is only a subset of C (en_US) if you like you can translate/adjust the others, too.

I know next to nothing about how to to set up accounting systems, but I think many of the US accounts don't have meaningful Danish equivalents. I'll let someone else more knowledgeable decide.

PR #874, opened 2021-01-13
Rename "Amt" to "Region" in Danish template

Munksgaard:
There have not been any "Amt" in Denmark since 2007. The new name is "Region".


fellen:
Committed, thanks!

PR #878, opened 2021-01-15
Bug 798085 - Incorrect transactions import of deposit with large number amount

christopherlam:
Quicken 2005 introduces U amount which sometimes differs from T amount. U amount has larger range, and must override T amount whenever they're not equal.

Example log output:
```
* 21:57:17  WARN <gnc.scm> qif-split has ambiguous amounts. T-21449672.96 U21,500,000.00. using U amount.
* 21:57:17  WARN <gnc.scm> qif-split has ambiguous amounts. T-20949672.96 U22,000,000.00. using U amount.
* 21:57:17  WARN <gnc.scm> qif-split has ambiguous amounts. T-21469672.96 U21,480,000.00. using U amount.
* 21:57:17  WARN <gnc.scm> qif-split has ambiguous amounts. T-21474672.96 U21,475,000.00. using U amount.
* 21:57:17  WARN <gnc.scm> qif-split has ambiguous amounts. T-21474772.96 U21,474,900.00. using U amount.
* 21:57:17  WARN <gnc.scm> qif-split has ambiguous amounts. T-21474822.96 U21,474,850.00. using U amount.
* 21:57:17  WARN <gnc.scm> qif-split has ambiguous amounts. T-21474835.96 U21,474,837.00. using U amount.
* 21:57:17  WARN <gnc.scm> qif-split has ambiguous amounts. T-21474836.48 U21,474,836.48. using U amount.
* 21:57:17  WARN <gnc.scm> qif-split has ambiguous amounts. T-21449672.96 U21,500,000.00. using U amount.
* 21:57:17  WARN <gnc.scm> qif-split has ambiguous amounts. T-20949672.96 U22,000,000.00. using U amount.
* 21:57:17  WARN <gnc.scm> qif-split has ambiguous amounts. T-21469672.96 U21,480,000.00. using U amount.
* 21:57:17  WARN <gnc.scm> qif-split has ambiguous amounts. T-21474672.96 U21,475,000.00. using U amount.
* 21:57:17  WARN <gnc.scm> qif-split has ambiguous amounts. T-21474772.96 U21,474,900.00. using U amount.
* 21:57:17  WARN <gnc.scm> qif-split has ambiguous amounts. T-21474822.96 U21,474,850.00. using U amount.
* 21:57:17  WARN <gnc.scm> qif-split has ambiguous amounts. T-21474835.96 U21,474,837.00. using U amount.
* 21:57:17  WARN <gnc.scm> qif-split has ambiguous amounts. T-21474836.48 U21,474,836.48. using U amount.
```

### DIFF
--- a/gnucash/import-export/qif-imp/qif-file.scm
+++ b/gnucash/import-export/qif-imp/qif-file.scm
@@ -135,6 +135,12 @@
 	      (unread-char c1)
 	      #f))))
 
+      (define (qif-split-set-amount split value override?)
+        (when (and split
+                   (not-bad-numeric-string? value)
+                   (or override? (not (qif-split:amount split))))
+          (qif-split:set-amount! split value)))
+
       (qif-file:set-path! self path)
       (if (not (access? path R_OK))
           ;; A UTF-8 encoded path won't succeed on some systems, such as
@@ -261,9 +267,13 @@
 
                            ;; T : total amount
                            ((#\T)
-                            (if (and default-split
-                                    (not-bad-numeric-string? value))
-                                (qif-split:set-amount! default-split value)))
+                            (qif-split-set-amount default-split value #f))
+
+                           ;; U : total amount (handle larger amount
+                           ;; than T; present in Quicken 2005
+                           ;; exports). See bug 798085
+                           ((#\U)
+                            (qif-split-set-amount default-split value #t))
 
                            ;; P : payee
                            ((#\P)


christopherlam:
...not ready yet, testing the ideas

christopherlam:
Buggy.

derekatkins:
The issue is that the string interpretation is switched (later) to gnc-numeric via this code in qif-fie.scm:
```
      ;; this one's a little tricky... it checks and sets all the
      ;; split amounts for the transaction together.
      (check-and-parse-field
       qif-xtn:split-amounts qif-xtn:set-split-amounts! gnc-numeric-equal
       qif-parse:check-number-formats '(decimal comma)
       qif-parse:parse-numbers/format (qif-file:xtns self)
       qif-parse:print-numbers
       'guess-on-ambiguity add-error 'split-amounts
       update-progress)
```
But if you look at `qif-xtn:set-split-amounts!` in qif-objects.scm, you'll see if uses:
```
       (qif-split:set-amount! split (if neg? (gnc-numeric-neg amount) amount)))
```
Which will overwrite the 'T' amount but leaves the 'U' as a string.  So the *next* time when it tries to read the amount it will see that T != U and will return the U (string) which is (now) wrong.  So the (easy) thing to try for now would be to erase `uamount` in `qif-xtn:set-split-amounts!`

christopherlam:
Thanks Derek you still have skill. This works.

christopherlam:
@jralls the C snippet was not used because `amount` and `uamount` are not both parsed in parallel. This fix seems hackish but IMHO captures the spirit of the T vs U amounts used.

gnucash/import-export/qif-imp/qif-objects.scm
```diff
@@ -261,12 +263,29 @@
...
   (record-accessor <qif-split> 'amount))
 
+(define qif-split:uamount-internal
+  (record-accessor <qif-split> 'uamount))
+
+(define (qif-split:amount split)
+  (let ((uamount (qif-split:uamount-internal split))
+        (amount (qif-split:amount-internal split)))
+    (cond
+     ((number? amount) amount)
```
jralls, line 276: This should be the *second* condition: It's always true if there's a `T` field even when the `T` field is wrong.

jralls:
>  C snippet was not used because amount and uamount are not both parsed in parallel.

Huh? 
```
    ((and amount uamount (not (equal? amount uamount)))
       (if (not  (check-32bit-wrap amount uamount))
            (gnc:warn (format #f (G_ "qif-split has ambiguous amounts: T ~1,2f U ~1,2f. Using U amount.") amount uamount)))
       uamount)
```
There's no point in filling up Ruben Chen's trace file with bogus warnings. It could be argued that it should be an error printed in the QIF import dialog (IIRC there's a place for that) instead of sent to the trace file so that the user can immediately inspect and if necessary correct those transactions.

One might also argue that we should just use `uamount` if it's present:
```
 (cond 
     ((and uamount (number? uamount))
       (if (and amount  (not  (check-32bit-wrap amount uamount)))
            (gnc:warn (format #f (G_ "qif-split has ambiguous amounts: T ~1,2f U ~1,2f. Using U amount.") amount uamount)))
      ((and amount (number? amount)) amount))
      (else 
         (gnc:warn (G_ "Neither T nor U had numeric entries"))
         0)))
```

christopherlam:
@jralls the exact order seems to be
1. `qif-split` object will store into `amount` the _string_ representation during initial .qif parsing eg `12,345,678.90`
2. all split string amounts are parsed to determine the exact format `n,nnn.nn` or `n.nnn.nnn,nn`. If ambiguous UI will confirm numeric format.
3. the splits are modified to set the `amount` field to the gnc-numeric eg `1234567890/100`.

My approach will do the T vs U comparison if they're both strings; they can't be compared numerically until after 3 above. It's the simplest method I can find.

jralls:
Interesting. @derekatkins's extract from the sample made me believe that the T string would always differ because it lacked thousands separators, but on looking at the whole thing I see that it's only the wrapped values. Unfortunately the sample only shows debits so I've asked on the bug for another example with credits.

That said, I don't think there's any point to doing a simple string compare. If the strings are the same then it doesn't matter which you use. A far simpler, less noisy, and just as complete fix is to always use U if it's present and fall back on T if it's not. It's inappropriate to warn when T wrap; wrapping is expected behavior. The only reason to raise an error is if T and cast-to-int U don't match.

> EDIT] the customized warning box in the qif-importer is not accessible from the qif-objects.scm getters/setters

So? Make it accessible or store the errors and display them when it is accessible.

christopherlam:
IMHO This branch (when fixup'd) is a good fix. The `gnc:warn` commit exists because `mywarn` warnings, although they're added into the dialog via `gnc_progress_dialog_append_log`, are not visible because of a usability bug in `gnc_ui_qif_import_load_progress_start_cb` -- the loader returns `(#t error-message)` as described but the latter does not pause the wizard to show the log.

jralls:
There's no reason to issue a warning. T being the wrong value is expected in cases where U > 21,474,835.47 -- never mind that as written it declares *every single entry* as ambiguous and says overridden or ignored depending only on whether T or U comes first in the file. It's noisy, slow, and of no value whatsoever. 

Is the `gnc_ui_qif_import_load_progress_start_cb` bug documented? Searching BZ for that in a comment turns up Zarro.

christopherlam:
> There's no reason to issue a warning. T being the wrong value is expected in cases where U > 21,474,835.47 -- never mind that as written it declares _every single entry_ as ambiguous and says overridden or ignored depending only on whether T or U comes first in the file. It's noisy, slow, and of no value whatsoever.

Ok removed `mywarn` warnings. If this branch is ok I'll fixup and push.

> Is the `gnc_ui_qif_import_load_progress_start_cb` bug documented? Searching BZ for that in a comment turns up Zarro.

It's not quite documented but is closely related to my troubles in #716 and bug 797736 -- some logic error when handling qif loading. May be related to 755c42a0.

PR #880, opened 2021-01-16
Also make underholdning and afstemning placeholders

Munksgaard:
As per the discussion in https://github.com/Gnucash/gnucash/pull/873#pullrequestreview-567912602


fellen:
Thank you!

Just in case you are intersted to improve the translation:
https://hosted.weblate.org/projects/gnucash/glossary/da/

For the program you would have to talk with Joe Hansen from
https://translationproject.org/team/da.html

PR #881, opened 2021-01-16
Make prototypes static for autoclear unit test

bombadil:
With the explicit prototype in place unit test builds on Debian Buster (using
buster-backports) fail with "error: testing::internal::ParamGenerator<TestCase*>
gtest_InstantiationAutoClearTestAutoClearTest_EvalGenerator_() was declared
extern and later static [-fpermissive]".

According to the comment preceding the declaration the only intent for the
explicit prototype seems to be to silence a warning which would cause a build
failure when using -Werror. As builds on Debian unstable seem to build just
fine without this explicit declaration I consider it safe to just drop it.

This is the full build log showing the build failure on Debian Buster without this change.
https://buildd.debian.org/status/fetch.php?pkg=gnucash&arch=amd64&ver=1%3A4.4-1%7Ebpo10%2B1&stamp=1610592036

### DIFF
--- a/libgnucash/app-utils/test/test-autoclear.cpp
+++ b/libgnucash/app-utils/test/test-autoclear.cpp
@@ -157,9 +157,9 @@ TEST_P(AutoClearTest, DoesAutoClear) {
 }
 
 #ifndef INSTANTIATE_TEST_SUITE_P
-// Silence "no previous declaration for" which is treated as error, due to -Werror
-testing::internal::ParamGenerator<TestCase*> gtest_InstantiationAutoClearTestAutoClearTest_EvalGenerator_();
-std::string gtest_InstantiationAutoClearTestAutoClearTest_EvalGenerateName_(const testing::TestParamInfo<TestCase*>&);
+// Silence "no previous declaration for" (treated as error due to -Werror) when building with GoogleTest < 1.8.1
+static testing::internal::ParamGenerator<TestCase*> gtest_InstantiationAutoClearTestAutoClearTest_EvalGenerator_();
+static std::string gtest_InstantiationAutoClearTestAutoClearTest_EvalGenerateName_(const testing::TestParamInfo<TestCase*>&);
 
 INSTANTIATE_TEST_CASE_P(
 #else // INSTANTIATE_TEST_SUITE_P


bombadil:
Hmm, I see it failing in Ubuntu 18.04 CI Tests now.
```
In file included from /usr/src/googletest/googletest/include/gtest/gtest.h:62:0,
                 from ../libgnucash/app-utils/test/test-autoclear.cpp:32:
../libgnucash/app-utils/test/test-autoclear.cpp: In function ‘testing::internal::ParamGenerator<TestCase*> gtest_InstantiationAutoClearTestAutoClearTest_EvalGenerator_()’:
/usr/src/googletest/googletest/include/gtest/gtest-param-test.h:1423:7: error: no previous declaration for ‘testing::internal::ParamGenerator<TestCase*> gtest_InstantiationAutoClearTestAutoClearTest_EvalGenerator_()’ [-Werror=missing-declarations]
       gtest_##prefix##test_case_name##_EvalGenerator_() { return generator; } \
       ^
../libgnucash/app-utils/test/test-autoclear.cpp:160:1: note: in expansion of macro ‘INSTANTIATE_TEST_CASE_P’
 INSTANTIATE_TEST_CASE_P(
 ^~~~~~~~~~~~~~~~~~~~~~~
```
Clearly some more differentiated approach is needed. Anyone an idea?

jralls:
@bombadil does changing the decl to `static` silence the error or just make another one? What version of GoogleTest are you using?

bombadil:
@jralls asked:
> does changing the decl to `static` silence the error or just make another one?

Yes, thanks for the suggestion. I've updated this pull request accordingly.

> What version of GoogleTest are you using?

Debian Buster (where the explicit prototypes are conflicting when not `static`) has GoogleTest 1.8.1.
Ubuntu 18.04 (where the CI build fails when the explicit prototypes are missing entirely) has GoogleTest 1.8.0.

I suspect once support building Gnucash on Ubuntu 18.04 isn't needed anymore the best approach would be to drop these explicit prototypes and require GoogleTest >= 1.8.1 for building instead.

fellen:
> I suspect once support for building Gnucash on Ubuntu 18.04 isn't needed anymore the best approach would be to drop these explicit prototypes and require GoogleTest >= 1.8.1 for building instead.

How about putting it as Todo in the comment?

bombadil:
> > I suspect once support for building Gnucash on Ubuntu 18.04 isn't needed anymore the best approach would be to drop these explicit prototypes and require GoogleTest >= 1.8.1 for building instead.
> 
> How about putting it as Todo in the comment?

That depends on whether that's consensus. Even if it is, maybe that's a TODO for the `master` branch only, isn't it? And then I would prefer to just do the change instead (and try to see how CMake can be made to bail-out on outdated GoogleTest versions).

jralls:
> That depends on whether that's consensus. Even if it is, maybe that's a TODO for the master branch only, isn't it? And then I would prefer to just do the change instead (and try to see how CMake can be made to bail-out on outdated GoogleTest versions).

It would indeed be for master branch only. Best to merge this first and merge maint->master then change the Googletest version requirement in the root CMakeLists.txt and README.dependencies and remove the backdoor declarations. (Don't call them "explicit", that has special meaning in C++).

bombadil:
@jralls agreed. Should I reword my commits then?

jralls:
Yes please. I think you can leave out the particulars about the distro and Googletest version. The prototype warning is more likely to be down to the compiler anyway and not marking the prototypes `static` is an error even if the compiler doesn't say so.

For master it would be cleanest to just require Googletest 1.10.0 which has the conversion from TEST_CASE to TEST_SUITE, then you can get rid of the whole `#ifdef`. You'll need to disable the Ubuntu-18.04 test build and I'll have to upgrade gnucash-on-osx, but neither is a big deal.

jralls:
I merged this after a rebase.

PR #899, opened 2021-02-05
Fix typos in libgnucash/

luzpaz:
Found via `codespell -q 3 -L ans,ba,parm,parms,numer`


libgnucash/doc/finutil.html
```diff
@@ -696,7 +696,7 @@ <h1>Financial Calculator Usage</h1></a>
 [Hh](elp) to Display This Help
 [Qq](uit)? to Quit
 [Ss](tatus)? to Display Status of Computations
-[Uu](ser) Display User Defined Variables
```
fellen, line 699: wrong

libgnucash/doc/finutil.html
```diff
@@ -756,7 +756,7 @@ <h2>Calculator Commands</h2></a>
 Initial Payment Date: Thu Aug 01 00:00:00 1996(2450297)
 <>
 </pre>
-<li>[Uu](ser) Display User Defined Variables
+<li>[Uu](set) Display User Defined Variables
```
fellen, line 759: too

fellen (changes requested):
I found 2 wrong interpreted strings, but perhaps a native speaker should review the end 's'es.

libgnucash/doc/finutil.html
luzpaz: Sorry about that. Will revert.

fellen:
Applied after reverting the two undesired changes.

Thank you!

PR #900, opened 2021-02-06
Bug 797997 - Gnucash File can not be found on Windows from command line.

Bob-IT:
This is NOT complete, more of am I doing the right thing, probably not...

If the path to the Gnucash file is like C:\Temp\Comité\myfile.gnucash on a Windows PC, the file will not be found. This also applies to the log file when --logto C:\Temp\Comité\mylog.txt

I started looking at the m_file_to_load and I noticed that the m_log_to_filename was used differently, it used a conversion function g_utf16_to_utf8 so started to and try and get the --logto to work.

First off, the windows macros "\_\_MINGW64__" was used but I think this is wrong as we still build 32bit Windows, changed them "\_\_MINGW32__" which then highlighted a problem with it using guichar2. So after some thought I came up with this which does work for the --logto but not sure if it is any good and whether it can be applied to m_file_to_load which is used in many more places. The other thought was whether these member variables should be wstring's.

### DIFF
--- a/gnucash/gnome-utils/gnc-main-window.c
+++ b/gnucash/gnome-utils/gnc-main-window.c
@@ -1737,40 +1737,41 @@ static gchar *generate_statusbar_lastmodified_message()
     {
         if (gnc_uri_targets_local_fs (uri))
         {
-#ifdef HAVE_SYS_STAT_H
             /* The filename is a true file. */
             gchar *filepath = gnc_uri_get_path ( uri );
             gchar *filename = g_path_get_basename ( filepath );
+            GFile *file = g_file_new_for_uri (uri);
+            GFileInfo *info = g_file_query_info (file,
+                                                 G_FILE_ATTRIBUTE_TIME_MODIFIED,
+                                                 G_FILE_QUERY_INFO_NONE,
+                                                 NULL, NULL);
+
+            if (info && g_file_info_has_attribute (info, G_FILE_ATTRIBUTE_TIME_MODIFIED))
             {
-                // Access the mtime information through stat(2)
-                struct stat statbuf;
-                int r = stat(filepath, &statbuf);
-                if (r == 0)
-                {
-                    /* Translators: This is the date and time that is shown in
-                    the status bar after opening a file: The date and time of
-                    last modification. The string is a format string using
-                    boost::date_time's format flags, see the boost docs for an
-                    explanation of the modifiers. */
-                    char *time_string = gnc_print_time64(statbuf.st_mtime,
-                     _("Last modified on %a, %b %d, %Y at %I:%M %p"));
-                    //g_warning("got time %ld, str=%s\n", mtime, time_string);
-                    /* Translators: This message appears in the status bar after opening the file. */
-                    message = g_strdup_printf(_("File %s opened. %s"),
-                                              filename, time_string);
-                    free(time_string);
-                }
-                else
-                {
-                    g_warning("Unable to read mtime for file %s\n", filepath);
-                    // message is still NULL
-                }
+                guint64 modtime = g_file_info_get_attribute_uint64 (info, G_FILE_ATTRIBUTE_TIME_MODIFIED);
+
+                /* Translators: This is the date and time that is shown in
+                the status bar after opening a file: The date and time of
+                last modification. The string is a format string using
+                boost::date_time's format flags, see the boost docs for an
+                explanation of the modifiers. */
+                char *time_string = gnc_print_time64 (modtime,
+                 _("Last modified on %a, %b %d, %Y at %I:%M %p"));
+                //g_warning("got time %ld, str=%s\n", mtime, time_string);
+                /* Translators: This message appears in the status bar after opening the file. */
+                message = g_strdup_printf(_("File %s opened. %s"),
+                                          filename, time_string);
+                free(time_string);
+            }
+            else
+            {
+                g_warning("Unable to read mtime for file %s\n", filepath);
+                // message is still NULL
             }
             g_free(filename);
             g_free(filepath);
-#else
-            return NULL;
-#endif
+            g_object_unref (info);
+            g_object_unref (file);
         }
         // If the URI is not a file but a database, we can maybe also show
         // something useful, but I have no idea how to obtain this information.
--- a/gnucash/gnucash-cli.cpp
+++ b/gnucash/gnucash-cli.cpp
@@ -38,6 +38,9 @@ extern "C" {
 
 #include <boost/locale.hpp>
 #include <boost/optional.hpp>
+#ifdef __MINGW32__
+#include <boost/nowide/args.hpp>
+#endif
 #include <iostream>
 
 namespace bl = boost::locale;
@@ -195,7 +198,9 @@ int
 main(int argc, char **argv)
 {
     Gnucash::GnucashCli application (argv[0]);
-
+#ifdef __MINGW32__
+    boost::nowide::args a(argc, argv); // Fix arguments - make them UTF-8
+#endif
     application.parse_command_line (argc, argv);
     application.start (argc, argv);
 
--- a/gnucash/gnucash-core-app.cpp
+++ b/gnucash/gnucash-core-app.cpp
@@ -345,24 +345,14 @@ load_user_config(void)
     try_load_config_array(stylesheet_files);
 }
 
-
 static void
 gnc_log_init (const std::vector <std::string> log_flags,
               const boost::optional <std::string> &log_to_filename)
 {
     if (log_to_filename && !log_to_filename->empty())
     {
-#ifdef __MINGW64__
-        auto *utf8_filename = g_utf16_to_utf8 (log_to_filename->c_str(), -1, NULL, NULL, NULL);
-#else
         auto utf8_filename = log_to_filename->c_str();
-#endif
-
         qof_log_init_filename_special (utf8_filename);
-
-#ifdef __MINGW64__
-        g_free (utf8_filename);
-#endif
     }
     else
     {
--- a/gnucash/gnucash.cpp
+++ b/gnucash/gnucash.cpp
@@ -61,6 +61,9 @@ extern "C" {
 
 #include <boost/locale.hpp>
 #include <boost/optional.hpp>
+#ifdef __MINGW32__
+#include <boost/nowide/args.hpp>
+#endif
 #include <iostream>
 #include <gnc-locale-utils.hpp>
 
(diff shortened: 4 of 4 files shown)


gjanssens:
@Bob-IT I'm a bit puzzled here.

First off, the `__MINGW64__` define was also used before the introduction of gnucash-cli and the conversion to c++ of  gnucash. If it worked then, it would be odd that  it doesn't work now any more. It is possible of course the macro check didn't work before either but as we have changed option parsing library it was not a problem before and it now is. Have you run gnucash in a debugger to check which code path is taken under Windows ?

Also have you tried without any conversion ? It's quite possible the new option parsing library already converts strings to utf-8 and we no longer need any conversion at all.

What was the exact error message you got with __MINGW32__ ?

Again, have you run gnucash on windows in a  debugger to check if all the values are as we assume (like what's the encoding format of the log_to_filename string variable at the entry of gnc_log_init - utf-8 or wide character?) ?

In any case I'd be surprised we would have to write our own conversion function as you did. I see  there are similar examples in  the gnucash codebase, but that's code from 2007 and 2011. I would hope we can use C++'  standard interfaces for it rather than GLib's.

jralls:
`__MINGW64__` isn't about 64-bit, it's about [MinGW-w64](http://mingw-w64.org). [MinGW classic](http://www.mingw.org) defines `__MINGW32__`, MinGW-w64 defines both. For our purposes it shouldn't matter which define you use; you can also use `__WIN32__` if you want.

It's not safe to put `wchar_t` into a `std::string`, that's what `std::wstring` is for.

There's well-tested character conversion code already in core-utils/gnc-filepath-utils.cpp and that API should be the only way any GnuCash code interacts with the file system.

Bob-IT:
Thanks for looking...
First the "\_\_MINGW64__" issue, I did think that it may of been down to me having to re-setup my build system recently but I can not see that. For that I just ran the setup.ps1 file accepting defaults and ended up with the same directory structure as before. Most of the files under mingw32 and mingw64 has most of the directories empty and the package cache for mingw is based on 'mingw-w64-i686'. I ran this command...
echo | gcc -E -dM - | grep MING
and the only entry it showed was "\_\_MINGW32__" so that is why I thought it was wrong.

Bob-IT:
@gjanssens
So am I now, I do not know what is going on now, I downloaded the two builds mentioned in the bug report...
gnucash-3.902-2020-05-29-git-3.902-217-g5091d7a67+.setup	OK
gnucash-3.902-2020-05-31-git-3.902-289-g2a42078e9+.setup	doesn't run 
and they both work for me, I need to double check this, maybe it was a later build that changed things.
Will report back...

Bob-IT:
I tried again with downloaded gnucash-3.902-2020-05-29-git-3.902-217-g5091d7a67+.setup and both paths worked.
In that version gnucash-bin.c used both versions of the macro.
I built that version and added a print statement inside the #ifdef \_\_MINGW64__ to prove whether g_utf16_to_utf8 was being used, it was NOT and both paths worked.
I changed the macro to "\_\_MINGW32__" just to see if --logto still worked, it did NOT.

Tried again with gnucash-3.902-2020-05-31-git-3.902-289-g2a42078e9+.setup and both paths DO NOT work.
Will check this version out and try and build it and start from there.

jralls:
> echo | gcc -E -dM - | grep MING

Interesting. I do indeed find `__MINGW64__` only in the 64-bit gcc.

jralls:
@Bob-IT Keep in mind that Microsoft has trouble getting encoding straight on the command line. CMD only uses DOS codepages. [Powershell's documentation](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_character_encoding?view=powershell-7.1) says that 5.1 and earlier defaults to UTF16LE, 6.0 and later to utf8nobom. But $OutputEncoding on a Powershell 6.1 window I just opened says it's Windows CP1252.

jralls:
@Bob-IT as a further complication note that at the point you're looking at it the command line has already passed through boost::program_options and that  might have [munged the encoding](https://www.boost.org/doc/libs/1_75_0/doc/html/program_options/howto.html#id-1.3.32.6.8).

Bob-IT:
@jralls 
> > echo | gcc -E -dM - | grep MING
> Interesting. I do indeed find `__MINGW64__` only in the 64-bit gcc.

Was this your build machine or the build server?
I did a quick search and there are 33 occurrences of "\_\_MINGW32__" and 2 for "\_\_MINGW64__" which are the ones referred to above.

gjanssens:
> @jralls
> 
> > > echo | gcc -E -dM - | grep MING
> > > Interesting. I do indeed find `__MINGW64__` only in the 64-bit gcc.
> 
> Was this your build machine or the build server?
> I did a quick search and there are 33 occurrences of "__MINGW32__" and 2 for "__MINGW64__" which are the ones referred to above.

@Bob-IT I think your are misreading @jralls ' message. He is not referring to *our* sourcefiles, but to the defines gcc sets when being run in a mingw environment. According to @jralls ' test the 32-bit version of gcc doesn't set the `__MINGW64__` define. Only the 64-bit one does. That is actually consistent with what you suspected in the first place. So you'd be correct to want to use `__MINGW32__` in our sources.

gjanssens:
> @Bob-IT as a further complication note that at the point you're looking at it the command line has already passed through boost::program_options and that might have [munged the encoding](https://www.boost.org/doc/libs/1_75_0/doc/html/program_options/howto.html#id-1.3.32.6.8).

It's what I handwavingly suggested in my first reply. And I believe that's what we should focus on to fix this issue.
According to the link you refer to if we do nothing and use standard strings, the program option value would be in ascii encoding rather than unicode. That's probably in the system's local code page then, so we'd still need a conversion from that code page to utf-8 as all our internal string handling is in utf-8.
But I don't have the time to run these tests myself right now, so this is only theory.

gjanssens:
The alternative is to capture filesystem paths (like the logto parameter) in wstring format and just keep them that way. But that may complicate things further down the stack. I don't think our backends are ready to accept wstring formatted strings.
So the third option is to capture them in wstring format and convert them to utf-8 ourselves.

jralls:
> Was this your build machine or the build server?

Mine. The build server is set up only for mingw32.

jralls:
> According to the link you refer to if we do nothing and use standard strings, the program option value would be in ascii encoding rather than unicode. That's probably in the system's local code page then, so we'd still need a conversion from that code page to utf-8 as all our internal string handling is in utf-8.
But I don't have the time to run these tests myself right now, so this is only theory.

> The alternative is to capture filesystem paths (like the logto parameter) in wstring format and just keep them that way. But that may complicate things further down the stack. I don't think our backends are ready to accept wstring formatted strings.
So the third option is to capture them in wstring format and convert them to utf-8 ourselves.

Unfortunately it's harder than that: The problem begins with `main (int argc, char** argv)`: If you pass "gnucash-cli" encoded in UTF16LE no-bom and try to handle it as a char vector you have 1 empty and 11 single-character strings. Fortunately that's easy to detect, it just needs to be done in `main` before passing on `argv`.

Much harder to sort out is if the command line is encoded in a codepage. A locale-to-utf8 function would work but remember that gcc doesn't implement C++ locale facets on Windows so we must use C for that part.

This used to work, so the code either is or was around somewhere, maybe in more than one place. As I mentioned there's some of it in `gnc-filepath-utils.cpp` but unfortunately it's less than I remembered and handles only unicode conversion.

Bob-IT:
I tried a different way by using 'boost:nowide:args' as can be seen in the first commit. This change allowed me to open said Gnucash file and create a log file using the paths in opening statement. This strange thing is it worked without adding the extra boost library to the bin directory which seems strange. I also observed that the 'Last modified date' was not being displayed so fixed that in the second commit. Not sure if that is a good fix or whether any of the additional 'HAVE_SYS_STAT_H' entries could be removed.

I noticed the email about using mt.exe, so downloaded the SDK and found the mt.exe file.
Removed the 'nowide' changes to gnucash.cpp and gnucash-cli.cpp leaving other changes and created a new setup file.
Installed this and confirmed the said file/path could not be opened, then ran the following command from the bin directory after copying the manifest file, mt.exe and a dll file to the bin directory as I think I read somewhere it needs to be in the same directory as the exe but that maybe just the manifest file.

mt.exe -manifest win32-manifest.txt -outputresource:gnucash.exe;#1

Tried again to open said file/path and it worked.

jralls:
There is no library, boost::nowide is headers-only. Did you look at the code? It's pretty clever.

It's a good fix and has the advantages over the manifest change of working with older versions of windows and not requiring a build-system change to create and incorporate the manifest.

Bob-IT:
> There is no library, boost::nowide is headers-only. Did you look at the code? It's pretty clever.

Had some time to get back to this, I saw there was a libboost_nowide-mt.dll so thought it was needed. Just to prove it to my self I looked at dll file with 'Dependencies' to see what functions it supports and as you stated there is nothing to do with 'args' present.
Had a look at the header file and see it uses the 'Wide' versions of the Windows command line functions but the cleverness is still beyond me. If I recall this was how I found this, I was thinking about using the wide version of the command line parser.
If OK I can push this.

jralls:
>If OK I can push this.

Please do.

gjanssens:
I like this simple solution. Thanks  boost and thanks  @Bob-IT for finding it.

Bob-IT:
Hmm, this PR did not automatically get merged and closed as normal when I pushed my changes from branch.
Closing...

gjanssens:
> Hmm, this PR did not automatically get merged and closed as normal when I pushed my changes from branch.
> Closing...

No problem.

Though just for your information, this happens when you 
1. rebase your local branch before pushing to the gnucash github repo and
2. did not push the rebased branch to your own github repo first.

If you want to avoid this on future PR's, you have to first force-push your locally rebased branch to your own github repo and only then push the same branch to the common gnucash repo. That way github can see the branch you push to gnucash is the same one as the PR branch (it compares commit hashes for this,  which change when you rebase).

Bob-IT:
Thanks for that, yes I think I may of rebased and forgot to push to my github branch yesterday.

fellen:
> > Hmm, this PR did not automatically get merged and closed as normal when I pushed my changes from branch.
:
> This happens when you
>     1. rebase your local branch before pushing to the gnucash github repo and
>     2. did not push the rebased branch to your own github repo first.
> If you want to avoid this on future PR's, you have to first force-push your locally rebased branch to your own github repo and only then push the same branch to the common gnucash repo. That way github can see the branch you push to gnucash is the same one as the PR branch (it compares commit hashes for this, which change when you rebase).

Is it somewhere in our [Git pages](https://wiki.gnucash.org/wiki/Category:Development)? I lost the track.

PR #903, opened 2021-02-08
Html Help dialog - Experiment

Bob-IT:
Before Christmas it was suggested that maybe the help should be based on html pages and thus use the existing webkit.
So to see what it could look like I cobbled together this. I started of with extracting the html help files from the Mac dmg file to use as a basis and can be seen below.

There is only one dialog so successive use of any help button will use existing window.
The Guide button toggles between Help and Guide so pages open in dialog.
The three buttons bottom left open in default external browser.
The only links supported in the dialog are secure ones which again open in the default external browser.
The location of the help files are 'share\gnucash\help\x\gnucash-help' but that could be changed
The Windows help files are named based on chapters, e.g. ch08s09.html as opposed to full section names.
HTML page appearance looks set at build time and not easily changed unless maybe there is reference to an external CSS file in appropriate directory that the user could change, thinking of those users with dark themes might not like a white page background.

Not sure if this is going to progress but thought it worthwhile as an experiment.
Bob

.
mac-version

I then decided to use the Windows built html files, see below...
win-version
 Which also included a toc.hhc file and I parsed this to add a search entry...
win-version-ac

Below are the 'C' language files I used zipped up separately to get below the 10MB limit...
[help-guide.zip](https://github.com/Gnucash/gnucash/files/5943537/help-guide.zip)
[help-help.zip](https://github.com/Gnucash/gnucash/files/5943554/help-help.zip)

### DIFF
--- a/gnucash/gnome-utils/gnc-gnome-utils.c
+++ b/gnucash/gnome-utils/gnc-gnome-utils.c
@@ -230,269 +230,30 @@ gnc_add_css_file (void)
     g_object_unref (provider_fallback);
 }
 
-/* This function fixes an issue with yelp that it does not work with the
- * ?anchor varient, see https://gitlab.gnome.org/GNOME/yelp/issues/116
- */
-static gchar *
-gnc_gnome_help_yelp_anchor_fix (GtkWindow *parent, const char *file_name, const char *anchor)
-{
-    const gchar * const *sdatadirs = g_get_system_data_dirs ();
-    const gchar * const *langs = g_get_language_names ();
-    gchar *lookfor = g_strconcat ("gnome/help/", file_name, NULL);
-    gchar *help_path = NULL;
-    gchar *help_file = NULL;
-    gchar *full_path = NULL;
-    gchar *uri = NULL;
+static GFunc help_dialog = NULL;
 
-    for (; *sdatadirs; sdatadirs++)
-    {
-        gchar *filepath = g_build_filename (*sdatadirs, lookfor, NULL);
-        if (g_file_test (filepath, G_FILE_TEST_EXISTS))
-            help_path = g_strdup (filepath);
-        g_free (filepath);
-    }
-    g_free (lookfor);
-
-    if (!help_path)
-    {
-        gnc_error_dialog (parent, "%s\n%s", _(msg_no_help_found), _(msg_no_help_reason));
-        PERR("Unable to find 'gnome/help' directory");
-        return NULL;
-    }
-
-    // add the file extension, currently .xml
-    help_file = g_strconcat (file_name, ".xml", NULL);
-
-    for (; *langs; langs++)
-    {
-        gchar *filename = g_build_filename (help_path, *langs, help_file, NULL);
-        if (g_file_test (filename, G_FILE_TEST_EXISTS))
-            full_path = g_strdup (filename);
-        g_free (filename);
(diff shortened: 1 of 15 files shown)


fellen:
> The Windows help files are named based on chapters, e.g. ch08s09.html as opposed to full section names.

That is bad. Section and chapter numbers can change from version to version. If a user reports an error in the released or an older version, it can be totally different in the nightly version.

fellen:
> Before Christmas it was suggested that maybe the help should be based on html pages and thus use the existing webkit.

I am no big fan of our HTML format. The 2. image displays the terrible spacing of lists as an example. I experimented with spacing='compact', which is correctly rendered by yelp, but ignored by our html generation.

fellen:
FYI: @codesmythe is experimenting with the conversion of the sources from docbook to reST. That opens perhaps more alternatives.

codesmythe:
I've attached a ZIP of HTML files of guide/C as rendered by Sphinx. I'd be curious if they render okay in the Help widgets and if the javascript-based search field works. The first zip is the HTML, while the second contains the web fonts used by Sphinx. Extract them in the same directory. The base file is then index.html. 

[gnucash_docs_html_fonts.zip](https://github.com/Gnucash/gnucash/files/5945790/gnucash_docs_html_fonts.zip)
[gnucash_docs_html.zip](https://github.com/Gnucash/gnucash/files/5945791/gnucash_docs_html.zip)

Bob-IT:
@codesmythe  I have added your files and all looks good as seen below, links to pages work along with external HTTPS links...
I kept getting an error about missing custom.css file, just created an empty file to overcome that.
Context help also worked with this...
#define HL_IMPORT_BC         "guide/C/ch_import_business_data.html#importing-bills-and-invoices"
This generates this full path...
'file:///usr/share/gnucash/help/C/gnucash_docs_html/guide/C/ch_import_business_data.html#importing-bills-and-invoices'
Not sure what your final directory structure will be with help and guide in multiple languages but this can be changed as appropriate as I only made the minimum amount of changes to accommodate your layout.

'View page source' does not work but that is down to the minimum changes I made, probably not required if deployed.

Screenshot_2021-02-09_10-22-35
And with the search...
Screenshot_2021-02-09_10-22-56

codesmythe:
Hi Bob,

Thanks for trying out my HTML files. We're a long way off from restructuring (pun intended) our docs to be able to generate something like that, but it's useful to know that Sphinx generated output seems to work.

-- Rob

gjanssens:
Search is definitely a nice feature to have in our documentation!

You also make mention of external links. I think it's important we don't load those inside the embedded webkit, but in an external browser. We don't want to have to deal with all the security risks that come with full internet access. Just something to keep in mind while working on this.

christopherlam:
> in an external browser

Ooh open in external browser. Definitely something reports should also offer -- will allow Windows reports to use modern browsers!

Bob-IT:
> You also make mention of external links. I think it's important we don't load those inside the embedded webkit, but in an external browser. We don't want to have to deal with all the security risks that come with full internet access. Just something to keep in mind while working on this.

The only valid links in the dialog are file:// urls. If there is an https:// url that will open in the default external browser, http:// urls are ignored.

Bob-IT:
I am closing this PR as it was an experiment and can be revived if and when required.

PR #908, opened 2021-02-12
[options.scm API] yank canonically-tabbed parent-subtotal-mode

christopherlam:
'canonically-tabbed option is buggy and has never graduated out of experimental for 17 years. Time to put it to rest.

All previous saved reports with this option will be switched to 'f i.e. parent-subtotal-mode hidden.


libgnucash/app-utils/options.scm
```diff
@@ -1057,12 +1057,20 @@ the option '~a'."))
...
-                 (setter-function-called-cb x)))
-           (rpterror-earlier "multichoice" x default-value)))
+       (cond
+        ((and (equal? section "Display")
+              (equal? name "Parent account subtotals")
+              (equal? x 'canonically-tabbed))
+         (gnc:warn "canonically-tabbed obsolete. switching to 'f")
+         (set! value 'f)
+         (if (procedure? setter-function-called-cb)
+             (setter-function-called-cb 'f)))
```
jralls, line 1067: I don't much like sticking an very limited exception into otherwise very-generic code, but I suppose there's no better way to intercept the call from a saved report. Shouldn't it set the option to `'t` though so that the user still gets their subtotals?

I don't see why you need to test `setter-function-called-cb`, it's always `nil` because `gnc:options-add-subtotal-view` calls `gnc:make-multichoice-option` not `gnc:make-multichoice-callback-option`. If you really think that you must it would be better to work out how to do it without duplicating the code.

jralls:
This is something users might notice so it should be at least mentioned on gnucash-user.

libgnucash/app-utils/options.scm
christopherlam: Ok `'t` may be a better fallback. Will notify user list for comments.

christopherlam: FWIW the `gnc:make-date-option` also has backward-compatibility code to accept old-style timepairs, converting to time64 on the fly.

jralls: That's a bit different, it applies to all date options. You don't need to specify a particular option by section and name  to handle the exceptional case. Plus, it's probably time to remove it, wasn't it introduced in 3.0?

christopherlam: Yes between 2.7.3 and 2.7.4 -- easy to remove.

christopherlam:
```patch
modified   libgnucash/app-utils/options.scm
@@ -707,20 +707,8 @@ the option '~a'."))
          (or
           (and (eq? 'relative (car date)) (symbol? (cdr date)))
           (and (eq? 'absolute (car date))
-               (or (and (pair? (cdr date))   ; we can still accept
-                        (exact? (cadr date)) ; old-style timepairs
-                        (exact? (cddr date)))
-                   (and (number? (cdr date))
-                        (exact? (cdr date))))))))
-  (define (maybe-convert-to-time64 date)
-    ;; compatibility shim. this is triggered only when date is type
-    ;; (cons 'absolute (cons sec nsec)) - we'll convert to
-    ;; (cons 'absolute sec). this shim must always be kept for gnucash
-    ;; to reload saved reports, or reload reports launched at startup,
-    ;; which had been saved as timepairs.
-    (if (pair? (cdr date))
-        (cons (car date) (cadr date))
-        date))
+               (and (number? (cdr date))
+                    (exact? (cdr date)))))))
   (define (list-lookup full-list item)
     (or (list-index (lambda (i) (eq? i item)) full-list)
         (begin
@@ -734,7 +722,7 @@ the option '~a'."))
      (lambda () value)
      (lambda (date)
        (if (date-legal date)
-           (set! value (maybe-convert-to-time64 date))
+           (set! value date)
            (gnc:error "Illegal date value set:" date)))
      default-getter
      (gnc:restore-form-generator value->string)
```

PR #909, opened 2021-02-13
Maint scrub closing entries

christopherlam:
Prototype. Finds and sets feature "closing txns have KVP"

Let's call "Closing Entries" (localised) somewhere in description without KVP flag, "untagged closing entries". They will be tagged to set KVP flag. But we want to do it only once per book, therefore set a feature.

If a book has NO untagged closing entries, we could set the feature. Setting feature means we can skip the qofquery search. Not setting feature means we need to run qof-query after load. I think this is overcomplicating things.

### DIFF
--- a/gnucash/gnome-utils/gnc-file.c
+++ b/gnucash/gnome-utils/gnc-file.c
@@ -46,6 +46,7 @@
 #include "gnc-plugin-file-history.h"
 #include "qof.h"
 #include "Scrub.h"
+#include "Scrub4.h"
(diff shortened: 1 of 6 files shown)


christopherlam:
Alternatively instead of feature we could set a book KVP flag "scrubbed-closing-entries" to TRUE? Much easier?

christopherlam:
p.s. the need for this PR stems from aiming to simplify many reports to not require "Closing Entries" tab.

PR #911, opened 2021-02-13
Follow up to Bug 798098 fix

Bob-IT:
@jralls  I think this covers the destroying of main windows with no plugin pages. I have tested this on my Linux VM and also Windows10 by trying various combinations of windows with pages, closing pages and quitting, having all empty windows and quitting and nothing untoward showed up.
Does this look OK to you?

### DIFF
--- a/gnucash/gnome-utils/gnc-main-window.c
+++ b/gnucash/gnome-utils/gnc-main-window.c
@@ -924,7 +924,8 @@ cleanup:
     if (error)
         g_error_free(error);
     g_free(window_group);
-    gtk_widget_show(GTK_WIDGET(window));
+    if (window)
+        gtk_widget_show (GTK_WIDGET(window));
 }
 
 void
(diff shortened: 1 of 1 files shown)


gnucash/gnome-utils/gnc-main-window.c
```diff
@@ -1405,12 +1406,22 @@ gnc_main_window_quit(GncMainWindow *window)
...
 
-        for (w = active_windows; w; w = g_list_next (w))
+        for (w = active_windows; w; w = next)
         {
-            window = w->data;
-            window->window_quitting = TRUE; // set window_quitting on all windows
+            GncMainWindowPrivate *priv;
+            GncMainWindow *wind = w->data;
+
+            next = g_list_next (w);
```
jralls, line 1416: Why pull this out of the loop def?

Bob-IT: I thought there may be a chance that the 'next' pointer maybe removed as part of destroying the window as it removes the window from the active_windows list so I thought it would be safer. I can add a note to that effect if required.

jralls: Right. Yeah, a comment would be good.

PR #912, opened 2021-02-13
Fix build with glib2 2.67.3+

wenottingham:
glib headers should not be included with 'extern "C"'.

Change in glib that caused this: https://gitlab.gnome.org/GNOME/glib/-/merge_requests/1715

Discussion of whether to work around it in glib: https://gitlab.gnome.org/GNOME/glib/-/merge_requests/1935. This was rejected, the suggestion there was to fix it in the including programs.

It may be possible for an expert in the header inclusion hierarchy to make a more targeted fix - I just kept editing files until it built.

Build tested on Fedora rawhide, smoke tested by starting the program with a new account from scratch.

### DIFF
--- a/gnucash/gnome-utils/gnc-amount-edit.h
+++ b/gnucash/gnome-utils/gnc-amount-edit.h
@@ -32,6 +32,8 @@
 #include "qof.h"
 #include "gnc-ui-util.h"
 
+#include <gtk/gtk.h>
+
 #define GNC_TYPE_AMOUNT_EDIT          (gnc_amount_edit_get_type())
 #define GNC_AMOUNT_EDIT(obj)          G_TYPE_CHECK_INSTANCE_CAST (obj, GNC_TYPE_AMOUNT_EDIT, GNCAmountEdit)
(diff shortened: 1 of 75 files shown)


jralls:
Thanks! I didn't even see this one coming.

BTW it's not possible to do a more targeted fix because glib won't let you include most internal headers from outside. You have to import `glib.h`. There are a few exceptions that add special features like `glib/gprintf.h` and `glib/gi18n.h` but neither of those has anything to do with this.

One other thing: It's better to put `#include <glib.h>` before the `extern "C"` block to override includes inside other headers declared `extern "C"`.

christopherlam:
Shouldnt this go to maint?

jralls:
Yeah, the new glib will leak into something--likely MinGW64--before 5.0 gets released.

wenottingham:
It's already in Fedora rawhide (which is how I ran across it.)

jralls:
Does that mean that the next Fedora release will include it along with a recent GnuCash release (4.4 or 4.5 I suppose)? If so, why did you branch on master instead of maint?

wenottingham:
Because I goofed up the branch, no other reason. Would you like another PR, or is that merge something you can do?

jralls:
Easier for me to do it.

PR #917, opened 2021-02-18
Removal of the Combott Widget and translation changes

Bob-IT:
This PR removes the GncCombott Widget and replaces its use with a GtkCombobox Widget.

The first commit does that, that's the easy part.
The second commit removes the tool tip from the scheme files which I think is correct. I have not changed any aspect of the 'multichoice-option' which could be done at a later time if required all I have done is removed the tool tip from the item vector.

There is still a tool tip for the widget like all the others and this displays the set default text for the option.

As there is plenty of space on the combo selections, I have changed some entries to use the removed tool tip so hopefully no useful information is lost but may of gone too far so please check and I will make changes as required.


gnucash/report/options-utilities.scm
```diff
@@ -227,29 +217,24 @@
     ;; usually the option name is: (N_ "Parent account balances")
     optname-parent-balance-mode
     (string-append sort-tag "a")
-    (N_ "How to show the balances of parent accounts.")
+    (N_ "How to show the balances of parent accounts.\n\
+Account Balance in the parent account, excluding any subaccounts.\n\
```
gjanssens, line 221: If you want to list descriptions of the possible options in the widget level tooltip, it would help if they are presented in a bullet list. Goes for other options further down as well.

gnucash/report/reports/aging.scm
```diff
@@ -385,8 +385,8 @@ totals to report currency.")
        (N_ "Leading date.")
        'duedate
        (list
-         (vector 'duedate (N_ "Due Date") (N_ "Due date is leading.")) ;; Should be using standard label for due date?
-         (vector 'postdate (N_ "Post Date") (N_ "Post date is leading."))))) ;; Should be using standard label for post date?
+         (vector 'duedate (N_ "Due date is leading"))
```
gjanssens, line 388: In this case I think it's superfluous to add the full description. The option name is already "Leading date". The "is leading" in the option values adds little IMO.

gnucash/report/reports/locale-specific/de_DE/taxtxf.scm
```diff
@@ -151,24 +151,16 @@
...
-                  (N_ "Last Yr 2nd Est Tax Qtr")
-                  (N_ "Apr 1 - May 31, Last year."))
-          (vector '3rd-last
-                  (N_ "Last Yr 3rd Est Tax Qtr")
-                  (N_ "Jun 1 - Aug 31, Last year."))
-          (vector '4th-last
-                  (N_ "Last Yr 4th Est Tax Qtr")
-                  (N_ "Sep 1 - Dec 31, Last year.")))))
+    (list (vector 'from-to (N_ "Use From - To"))
+          (vector '1st-est (N_ "1st Est Tax Quarter, Jan 1 - Mar 31"))
```
gjanssens, line 155: Perhaps put the illustrative dates in parens ?
`(N_ "1st Est Tax Quarter (Jan 1 - Mar 31)"))`

It's probably a matter of taste though.

gjanssens (commented):
Looks good overall. I have made a few suggestions on some of the option value changes you made. I haven't reviewed them all (yet) though.

gnucash/report/reports/aging.scm
Bob-IT: When I first read this option, I was unsure of what it did, still don't, I assume it is some sort order but maybe I do not have enough dummy data to see. The option is 'Due or Post Date', the default tool tip was 'Leading Date'.

gnucash/report/reports/locale-specific/de_DE/taxtxf.scm
Bob-IT: I was thinking about putting it back to how it was, then read that quarters could be different so will change to use brackets.

gnucash/report/options-utilities.scm
Bob-IT: I am not sure I like this, if I could think of shorter appropriate text for the longest item of the option I would of used the tool tips, in this case...
'Calculate the subtotal for this parent account and all of its subaccounts, and show this as the parent account balance.'
and the one below...
'Show subtotals for selected parent accounts which have subaccounts'

christopherlam:
From my usability pov the changes are welcome. Tooltips in combobox were neither easily found, nor added much more information than the option names. New option names will take some adjustment period, but no concerns.

gnucash/report/reports/locale-specific/de_DE/taxtxf.scm
fellen: Two notes:

- In the german report only the regular quarters are used. The US Est Quarters exist only for backward compatibility with the US report.

- For the US report we should display one of the irregular quarters like "Apr 1 - May 31.". the first quarter is the same in both lists.

jralls:
You need to change `gnc:make-multichoice-callback-option` in libgnucash/app-utils/options.scm to remove vector-ref (car p-vals) 2) from `multichoice-strings` and the lambda immediate following `ok-values` in the call to `gnc:make-option`.

Bob-IT:
> You need to change `gnc:make-multichoice-callback-option` in libgnucash/app-utils/options.scm to remove vector-ref (car p-vals) 2) from `multichoice-strings` and the lambda immediate following `ok-values` in the call to `gnc:make-option`.

@jralls  Did this show up as an error somewhere? I do not understand scheme, but I think I have made the first change to..
   (define (multichoice-strings p-vals)
    (if (null? p-vals)
        '()
        (cons (vector-ref (car p-vals) 1)
              (multichoice-strings (cdr p-vals))))) 

Regarding the second change, is that remove the line...
     ` (lambda () (multichoice-strings ok-values)) `

jralls:
@Bob-IT I didn't see any errors, it's just code review: You changed all of the report code calling `gnc:make-multichoice.*option` to pass a vector with 2 elements but the code at the other end still tries to read a third element so it needs adjustment.

I think your first change is correct, best if @gjanssens or @christopherlam confirms. IIUC the string of `cons` is just Scheme's way of building a list one piece at a time and dropping one won't affect iterating over the list except that the result will have half the elements. This function is called very indirectly only by `gnc_owner_type_from_option` which queries an internal option that just sticks the enum in it, so no actual impact.

The second change is to remove [`(lambda (x) (vector-ref (list-ref ok-values x) 2))` ](https://github.com/Gnucash/gnucash/blob/maint/libgnucash/app-utils/options.scm#L1087). That lambda is invoked by [`gnc:option-index-get-description`](https://github.com/Gnucash/gnucash/blob/maint/libgnucash/app-utils/options.scm#L301) so that needs to be removed as well, and it's called in turn by [`getters.index_to_description`](https://github.com/Gnucash/gnucash/blob/maint/libgnucash/app-utils/option-util.c#L555) called by [`gnc_option_permissible_value_description`](https://github.com/Gnucash/gnucash/blob/maint/libgnucash/app-utils/option-util.c#L929). That was used by `dialog-options.c` and you've already removed all of those calls.

Oh, the [comments](https://github.com/Gnucash/gnucash/blob/maint/libgnucash/app-utils/options.scm#L194) explaining `gnc:option-data` and `gnc:option-data-fns` need to be fixed to match. 

But there's one more wrinkle: Some user might have written a custom report with a multichoice option and be passing it 3 strings. On the face of it a non-problem because the third element will be ignored, but it would be nice for `gnc:make-multichoice-callback-option` issued a deprecation warning suggesting that they consider making the descriptions into the names if it makes sense to.

christopherlam:
I tried to review removal of obsolete guile string handling, and found it more difficult than expected. Perhaps this PR can be considered complete as it is and I'll review in when I find time?

jralls:
> I tried to review removal of obsolete guile string handling, and found it more difficult than expected. Perhaps this PR can be considered complete as it is and I'll review in when I find time?

Uh, no. String handling isn't the primary issue, reading past the end of a vector is. That will crash or worse in C, what does it do in Scheme?

christopherlam:
> 
> 
> > I tried to review removal of obsolete guile string handling, and found it more difficult than expected. Perhaps this PR can be considered complete as it is and I'll review in when I find time?
> 
> Uh, no. String handling isn't the primary issue, reading past the end of a vector is. That will crash or worse in C, what does it do in Scheme?

Crash too... but this branch doesn't crash which tells me there's nothing that requests combo's items' tt anymore.

jralls:
Does it really crash or does it just raise a Scheme exception because the Scheme interpreter knows that the vector has only two elements? I hope it's the latter, that's an annoyance. A hard crash is a security problem because of user code.

I don't see this as blocked by a review of string handling, it's just a bit more cleanup that I've already explained.

christopherlam:
> Does it really crash or does it just raise a Scheme exception because the Scheme interpreter knows that the vector has only two elements? I hope it's the latter, that's an annoyance. A hard crash is a security problem because of user code.

Usually reports will throw a scheme exception which is caught by https://github.com/Gnucash/gnucash/blob/04d4038e3fa7022497d6525e9b332d80058d1cbc/gnucash/report/report-core.scm#L762 but options->scheme IIRC do not have the exception handler, therefore would crash.

gnucash/report/options-utilities.scm
```diff
@@ -83,12 +82,12 @@
    (gnc:make-multichoice-option
     pagename name-display-depth sort-tag help-string default-depth
     (list (vector 'all (N_ "All") (N_ "All accounts"))
```
gjanssens, line 84: I think you missed this multi-choice option. It still has both a name and a description.

gnucash/report/options-utilities.scm
gjanssens: I'm not sure I can follow your train of thought here so I can't tell if you understood what I suggested.

I'll try again - I think you have done the right thing to keep the short multi-choice names here (Account Balance, Calculate Subtotal, Do not show). However you have also chosen to add the longer descriptions in the common widget tooltip. In there the 3 descriptions are added just one after the other. It's not that clear these descriptions refer to 3 possible choices. So I would suggest to at least do this:

```
(N_ "How to show the balances of parent accounts.\n\
* Account Balance in the parent account, excluding any subaccounts.\n\
* Calculate the subtotal for this parent account and all of its subaccounts, and show this as the parent account balance.\n\
* Do not show any balances of parent accounts.")
```

gnucash/report/reports/aging.scm
gjanssens: Ok, I misinterpreted the option definition. The name of the option is "Due Date or Post Date" and the  tool tip is "Leading Date". That's an poor option name unfortunately, but I'm not to suggest option name changes in this PR.

Still your extended value names are not adding much compared to how the option  worked before. The "is leading" part would only appear in a tool tip anyway. So  before you would have a tool tip specifying "Due date is leading" or "Post date is leading". Now you have a common tip specifying "Leading date" and "Due date" or "Post date" as values. Adding the "is leading" part in the value names should not be necessary. If anything we could add a better explanation of what "Leading  date" means in the tool tip.

gjanssens:
There's another multi-choice option  that I don't see corrected here: the [stylesheet selector](https://github.com/Gnucash/gnucash/blob/maint/gnucash/report/report-core.scm#L290). It also generates per value tool tips, though these auto-generated ones are quite useless; So there's no loss in removing them.

More review will follow as I find time.

gjanssens:
> @Bob-IT I didn't see any errors, it's just code review: You changed all of the report code calling `gnc:make-multichoice.*option` to pass a vector with 2 elements but the code at the other end still tries to read a third element so it needs adjustment.
> 
> I think your first change is correct, best if @gjanssens or @christopherlam confirms.

Looks correct to me.

> The second change is to remove [`(lambda (x) (vector-ref (list-ref ok-values x) 2))` ](https://github.com/Gnucash/gnucash/blob/maint/libgnucash/app-utils/options.scm#L1087). That lambda is invoked by [`gnc:option-index-get-description`](https://github.com/Gnucash/gnucash/blob/maint/libgnucash/app-utils/options.scm#L301) so that needs to be removed as well, and it's called in turn by [`getters.index_to_description`](https://github.com/Gnucash/gnucash/blob/maint/libgnucash/app-utils/option-util.c#L555) called by [`gnc_option_permissible_value_description`](https://github.com/Gnucash/gnucash/blob/maint/libgnucash/app-utils/option-util.c#L929). That was used by `dialog-options.c` and you've already removed all of those calls.
> 

Agreed.

> Oh, the [comments](https://github.com/Gnucash/gnucash/blob/maint/libgnucash/app-utils/options.scm#L194) explaining `gnc:option-data` and `gnc:option-data-fns` need to be fixed to match.
> 

Ditto.

> But there's one more wrinkle: Some user might have written a custom report with a multichoice option and be passing it 3 strings. On the face of it a non-problem because the third element will be ignored, but it would be nice for `gnc:make-multichoice-callback-option` issued a deprecation warning suggesting that they consider making the descriptions into the names if it makes sense to.

To clarify, we can add a  test to see if the ok-data list still has 3-value vectors at the start of ]gnc:make-multichoice-callback-option`. Other than that I fully agree.

gnucash/report/reports/standard/advanced-portfolio.scm
```diff
@@ -79,28 +79,16 @@ by preventing negative stock balances.<br/>")
...
       "d" (N_ "The source of price information.") 'pricedb-nearest
-      (list (vector 'pricedb-latest
-		    (N_ "Most recent")
-		    (N_ "The most recent recorded price."))
-	    (vector 'pricedb-nearest
-		    (N_ "Nearest in time")
-		    (N_ "The price recorded nearest in time to the report date."))
-	    )))
+      (list (vector 'pricedb-latest (N_ "The most recent recorded price"))
+            (vector 'pricedb-nearest (N_ "The price recorded nearest in time to the report date")))))
```
gjanssens, line 83: Suggested names:
"Most recent"
"Closest to  report date" (or "Nearest to report date")

gnucash/report/reports/standard/advanced-portfolio.scm
```diff
@@ -79,28 +79,16 @@ by preventing negative stock balances.<br/>")
...
-	    (vector 'fifo-basis
-		    (N_ "FIFO")
-		    (N_ "Use first-in first-out method for basis."))
-	    (vector 'filo-basis
-		    (N_ "LIFO")
-		    (N_ "Use last-in first-out method for basis."))
-	    )))
+      (list (vector 'average-basis (N_ "Use average cost of all shares for basis"))
+            (vector 'fifo-basis (N_ "Use first-in first-out method for basis"))
+            (vector 'filo-basis (N_ "Use last-in first-out method for basis")))))
```
gjanssens, line 91: Value names should not be full sentences. Perhaps:
"Average cost of all shares"
"First in, first out"
"Last in, first out"

In any case the "Use " and "for basis" parts should go.

gnucash/report/reports/standard/advanced-portfolio.scm
```diff
@@ -112,16 +100,9 @@ by preventing negative stock balances.<br/>")
...
-            (vector 'include-in-gain
-                    (N_ "Include in gain")
-                    (N_  "Include brokerage fees in the gain and loss but not in the basis."))
-            (vector 'ignore-brokerage
-                    (N_ "Ignore")
-                    (N_ "Ignore brokerage fees entirely."))
-            )))
+      (list (vector 'include-in-basis (N_ "Include brokerage fees in the basis for the asset"))
+            (vector 'include-in-gain (N_ "Include brokerage fees in the gain and loss but not in the basis"))
+            (vector 'ignore-brokerage (N_ "Ignore brokerage fees entirely")))))
```
gjanssens, line 105: Also here I think the original value names were sufficient, though possible suggestions would be
(second value) "Include in gain/loss"
(last value) "Omit from report"

gnucash/report/reports/standard/balsheet-eg.scm
```diff
@@ -225,24 +225,15 @@
...
                   "b" opthelp-neg-format 'negsign
-                  (list (vector 'negsign
-                                (N_ "Sign")
-                                (N_ "Prefix negative amounts with a minus sign, e.g. -$10.00."))
-                        (vector 'negbrackets
-                                (N_ "Brackets")
-                                (N_ "Surround negative amounts with brackets, e.g. ($100.00).")))))
+                  (list (vector 'negsign (N_ "Prefix negative amounts with a minus sign, e.g. -$10.00"))
+                        (vector 'negbrackets (N_ "Surround negative amounts with brackets, e.g. ($100.00)")))))
+
```
gjanssens, line 236: Perhaps shorten to
"Sign: -$10.00"
"Brackets: ($10.00)"

gnucash/report/reports/standard/income-gst-statement.scm
```diff
@@ -159,20 +159,17 @@ for taxes paid on expenses, and type LIABILITY for taxes collected on sales.")
...
+purchase and tax accounts\nFor UK VAT Return. Specify sales, purchase and tax \
 accounts. EU rules may be used. Denote EU VAT accounts *EUVAT* in \
 account description, and denote EU goods sales and purchases accounts \
-with *EUGOODS* in the account description."))) #f
+with *EUGOODS* in the account description.")
+    'default
+    (list (vector 'default (G_ "Default Format"))
+          (vector 'au-bas (G_ "Australia BAS"))
+          (vector 'uk-vat (G_ "UK VAT Return")))
+     #f
```
gjanssens, line 172: As mentioned in an earlier comment, perhaps add some formatting to the tool tip. Without list indicators it's not so easy to read.

Also  out of curiosity, I see you are escaping newlines and adding "\n" in several places. Can this be simplified to not escape newlines and just put them where you consider them appropriate ?

For example I **think** these are equivalent:
```
"One\n\
two\n\
three"

"One
two
three"

```

The second one is definitely more readable.

gnucash/report/reports/standard/invoice.scm
```diff
@@ -110,25 +110,19 @@
 
 (define layout-key-list
   ;; Translators: "Their details" refer to the invoice 'other party' details i.e. client/vendor name/address/ID
-  (list (cons 'client (list (cons 'text (G_ "Their details"))
```
gjanssens, line 113: Translator comment has become redundant.

gnucash/report/reports/standard/invoice.scm
```diff
@@ -110,25 +110,19 @@
...
-                            (cons 'tip (G_ "Client or vendor name, address and ID"))))
+  (list (cons 'client (list (cons 'text (G_ "Client or vendor name, address and ID"))))
 
         ;; Translators: "Our details" refer to the book owner's details i.e. name/address/tax-ID
```
gjanssens, line 115: Translator comment has become redundant.

gnucash/report/reports/standard/new-aging.scm
```diff
@@ -121,14 +115,8 @@ exist but have no suitable transactions."))
...
       (list
-       ;; Should be using standard label for due date?
-       (vector 'duedate
-               (N_ "Due Date")
-               (N_ "Due date is leading."))
-       ;; Should be using standard label for post date?
-       (vector 'postdate
-               (N_ "Post Date")
-               (N_ "Post date is leading.")))))
+       (vector 'duedate (N_ "Due date is leading"))
```
gjanssens, line 118: Refer to earlier comment.

gnucash/report/reports/standard/new-aging.scm
```diff
@@ -415,12 +403,8 @@ exist but have no suitable transactions."))
...
       gnc:pagename-display optname-addr-source "a" (N_ "Address source.") 'billing
       (list
-       (vector 'billing
-               (N_ "Billing")
-               (N_ "Address fields from billing address."))
-       (vector 'shipping
-               (N_ "Shipping")
-               (N_ "Address fields from shipping address.")))))
+       (vector 'billing (N_ "Address fields from billing address"))
+       (vector 'shipping (N_ "Address fields from shipping address")))))
```
gjanssens, line 407: Option name itself is "Address Source".  The values can be reduce to "Billing Address" and "Shipping Address" IMO.

gnucash/report/reports/standard/new-owner-report.scm
```diff
@@ -944,17 +944,15 @@
...
-invoices and amounts.")))))
+    "hc" (N_ "Show linked transactions.
+Disabled: Linked transactions are hidden.\n\
+Simple: Invoices show if paid, payments show invoice numbers.\n\
+Detailed: Invoices show list of payments, payments show list of \
+invoices and amounts.")
+    'none
+    (list (vector 'none (N_ "Disabled"))
+          (vector 'simple (N_ "Simple"))
+          (vector 'detailed (N_ "Detailed")))))
```
gjanssens, line 955: Formatting

gnucash/report/reports/standard/new-owner-report.scm
```diff
@@ -966,14 +964,8 @@ invoices and amounts.")))))
...
-     ;; Should be using standard label for due date?
-     (vector 'duedate
-             (N_ "Due Date")
-             (N_ "Due date is leading."))
-     ;; Should be using standard label for post date?
-     (vector 'postdate
-             (N_ "Post Date")
-             (N_ "Post date is leading.")))))
+     (vector 'duedate (N_ "Due date is leading"))
+     (vector 'postdate (N_ "Post date is leading")))))
```
gjanssens, line 968: Refer to earlier comments.

gnucash/report/reports/standard/owner-report.scm
```diff
@@ -609,8 +609,8 @@
     (N_ "Leading date.") 
     'duedate 
     (list 
-     (vector 'duedate (N_ "Due Date") (N_ "Due date is leading.")) ;; Should be using standard label for due date? 
-     (vector 'postdate (N_ "Post Date") (N_ "Post date is leading."))))) ;; Should be using standard label for post date? 
+     (vector 'duedate (N_ "Due date is leading"))
+     (vector 'postdate (N_ "Post date is leading")))))
```
gjanssens, line 613: Refer to earlier comment.

gnucash/report/reports/standard/receivables.scm
```diff
@@ -65,8 +65,8 @@
         (N_ "Address source.")
         'billing
         (list
-         (vector 'billing (N_ "Billing") (N_ "Address fields from billing address."))
-         (vector 'shipping (N_ "Shipping") (N_ "Address fields from shipping address.")))))
+         (vector 'billing (N_ "Address fields from billing address"))
+         (vector 'shipping (N_ "Address fields from shipping address")))))
```
gjanssens, line 69: Refer to earlier comment.

gnucash/report/reports/standard/register.scm
```diff
@@ -399,8 +399,8 @@
     "ia" (N_ "Display the amount?")
     'double
     (list
-     (vector 'single (N_ "Single") (N_ "Single Column Display."))
-     (vector 'double (N_ "Double") (N_ "Two Column Display.")))))
+     (vector 'single (N_ "Single Column Display"))
+     (vector 'double (N_ "Two Column Display")))))
```
gjanssens, line 403: "Two-Column Display" ?  Not sure, a native speaker should verify.

gnucash/report/reports/standard/trial-balance.scm
```diff
@@ -200,15 +200,9 @@
...
-      (list (vector 'current
-                    (N_ "Current Trial Balance")
-                    (N_ "Uses the exact balances in the general journal"))
-            (vector 'pre-adj
-                    (N_ "Pre-adjustment Trial Balance")
-                    (N_ "Ignores Adjusting/Closing entries"))
-            (vector 'work-sheet
-                    (N_ "Work Sheet")
-                    (N_ "Creates a complete end-of-period work sheet")))))
+      (list (vector 'current (N_ "Uses the exact balances in the general journal"))
```
gjanssens, line 203: Suggestions:
"General journal exact balances"
"No adjusting/closing entries"
"Full end-of-period work sheet"

gnucash/report/stylesheets/footer.scm
```diff
@@ -102,15 +102,9 @@
...
-                    (N_ "Align the banner to the left."))
-            (vector 'center
-                    (N_ "Center")
-                    (N_ "Align the banner in the center."))
-            (vector 'right
-                    (N_ "Right")
-                    (N_ "Align the banner to the right.")))))
+      (list (vector 'left (N_ "Align the banner to the left"))
+            (vector 'center (N_ "Align the banner in the center"))
+            (vector 'right (N_ "Align the banner to the right")))))
```
gjanssens, line 107: I suggest to keep the short names. Together with the option name and tool tip that's fully self-explanatory.

gnucash/report/stylesheets/head-or-tail.scm
```diff
@@ -149,16 +149,9 @@
...
-            (vector 'center
-                    (N_ "Center")
-                    (N_ "Align the banner in the center."))
-            (vector 'right
-                    (N_ "Right")
-                    (N_ "Align the banner to the right."))
-            )))
+      (list (vector 'left (N_ "Align the banner to the left"))
+            (vector 'center (N_ "Align the banner in the center"))
+            (vector 'right (N_ "Align the banner to the right")))))
```
gjanssens, line 154: Refer to earlier comment.

gnucash/report/trep-engine.scm
```diff
@@ -1026,12 +975,8 @@ be excluded from periodic reporting.")
...
       "h" (G_ "Amount of detail to display per transaction.")
       disp-detail-level?
-      (list (vector 'multi-line
-                    (G_ "Multi-Line")
-                    (G_ "Display all splits in a transaction on a separate line."))
-            (vector 'single
-                    (G_ "Single")
-                    (G_ "Display one line per transaction, merging multiple splits where required.")))
+      (list (vector 'multi-line (G_ "Display all splits in a transaction on separate lines"))
+            (vector 'single (G_ "Display one line per transaction, merging multiple splits where required")))
```
gjanssens, line 979: How about
"One split per line"
"One transaction per line"

gnucash/report/trep-engine.scm
```diff
@@ -1043,9 +988,9 @@ be excluded from periodic reporting.")
       "m" (G_ "Display the amount?")
       amount-value
       (list
-       (vector 'none   (G_ "None") (G_ "No amount display."))
-       (vector 'single (G_ "Single") (G_ "Single Column Display."))
-       (vector 'double (G_ "Double") (G_ "Two Column Display.")))
+       (vector 'none   (G_ "No amount display"))
+       (vector 'single (G_ "Single Column Display"))
+       (vector 'double (G_ "Two Column Display")))
```
gjanssens, line 993: "Hide"
"Single column"
"Two columns"

Bob-IT:
OK I think I have done all the textual changes required.
@christopherlam Can you have a look at the remaining scheme changes requested by John and Geert from the 1st of March and 12th. I have not added the change I proposed as when I tried to do the second change it caused an error so you will need to add that.
gnc_option_permissible_value_description is still being used for radio button tool tips so am not sure how much needs changing.
Depending on the size of changes you could add a patch here and I will add to PR and push or I can push this so you can make the changes.

christopherlam (commented):
Here's a full patch which incorporates two changes:
1. reuse strings for my income-gst-statement.scm report
2. remove tooltip capability from `gnc:make-list-option` as well thereby allowing removal of `gnc:option-index-get-description`. The `gnc:make-list-option` is only used in `hello-world.scm` and `average-balance.scm` therefore is not a great loss.

christopherlam:
```patch
diff --git a/gnucash/report/reports/example/average-balance.scm b/gnucash/report/reports/example/average-balance.scm
index 49044e3c9..66a285a49 100644
--- a/gnucash/report/reports/example/average-balance.scm
+++ b/gnucash/report/reports/example/average-balance.scm
@@ -126,9 +126,9 @@
       gnc:pagename-display (N_ "Plot Type")
       "c" (N_ "The type of graph to generate.") (list 'AvgBalPlot)
       (list 
-       (vector 'AvgBalPlot (N_ "Average") (N_ "Average Balance."))
-       (vector 'GainPlot (N_ "Profit") (N_ "Profit (Gain minus Loss)."))
-       (vector 'GLPlot (N_ "Gain/Loss") (N_ "Gain And Loss.")))))
+       (vector 'AvgBalPlot (N_ "Average"))
+       (vector 'GainPlot (N_ "Profit"))
+       (vector 'GLPlot (N_ "Gain/Loss")))))
 
     (gnc:options-add-plot-size! 
      options gnc:pagename-display 
diff --git a/gnucash/report/reports/example/hello-world.scm b/gnucash/report/reports/example/hello-world.scm
index e077fa92e..77f8d5552 100644
--- a/gnucash/report/reports/example/hello-world.scm
+++ b/gnucash/report/reports/example/hello-world.scm
@@ -179,15 +179,9 @@
       (N_ "Hello Again") (N_ "A list option")
       "h" (N_ "This is a list option.")
       '(good)
-      (list (vector 'good
-                    (N_ "The Good")
-                    (N_ "Good option."))
-            (vector 'bad
-                    (N_ "The Bad")
-                    (N_ "Bad option."))
-            (vector 'ugly
-                    (N_ "The Ugly")
-                    (N_ "Ugly option.")))))
+      (list (vector 'good (N_ "The Good"))
+            (vector 'bad (N_ "The Bad"))
+            (vector 'ugly (N_ "The Ugly")))))
     
     ;; This option is for testing. When true, the report generates
     ;; an exception.
diff --git a/gnucash/report/reports/standard/income-gst-statement.scm b/gnucash/report/reports/standard/income-gst-statement.scm
index 9b1fea76b..179ee9072 100644
--- a/gnucash/report/reports/standard/income-gst-statement.scm
+++ b/gnucash/report/reports/standard/income-gst-statement.scm
@@ -155,13 +155,18 @@ for taxes paid on expenses, and type LIABILITY for taxes collected on sales.")
    options
    (gnc:make-multichoice-callback-option
     pagename-format (N_ "Report Format")
-    "a" 
-    (G_ "Report format.
-* Default Format.
-* Australia Business Activity Statement. Specify sales, purchase and tax accounts.
-* UK VAT Return. Specify sales, purchase and tax accounts. EU rules may be used.\
-Denote EU VAT accounts *EUVAT* in account description, and denote EU goods sales \
-and purchases accounts with *EUGOODS* in the account description.")
+    "a"
+    (string-join
+     (list
+      (G_ "Report format")
+      (G_ "Default Format")
+      (G_ "Australia Business Activity Statement. Specify sales, \
+purchase and tax accounts.")
+      (G_ "UK VAT Return. Specify sales, purchase and tax \
+accounts. EU rules may be used. Denote EU VAT accounts *EUVAT* in \
+account description, and denote EU goods sales and purchases accounts \
+with *EUGOODS* in the account description."))
+     "\n* ")
     'default
     (list (vector 'default (G_ "Default Format"))
           (vector 'au-bas (G_ "Australia BAS"))
diff --git a/libgnucash/app-utils/option-util.c b/libgnucash/app-utils/option-util.c
index c36656ac0..ceadeff52 100644
--- a/libgnucash/app-utils/option-util.c
+++ b/libgnucash/app-utils/option-util.c
@@ -95,7 +95,6 @@ struct _Getters
     SCM value_validator;
     SCM option_data;
     SCM index_to_name;
-    SCM index_to_description;
     SCM index_to_value;
     SCM value_to_index;
     SCM number_of_indices;
@@ -552,8 +551,6 @@ initialize_getters(void)
         scm_c_eval_string ("gnc:option-value-validator");
     getters.option_data = scm_c_eval_string ("gnc:option-data");
     getters.index_to_name = scm_c_eval_string ("gnc:option-index-get-name");
-    getters.index_to_description =
-        scm_c_eval_string ("gnc:option-index-get-description");
     getters.number_of_indices = scm_c_eval_string ("gnc:option-number-of-indices");
     getters.index_to_value = scm_c_eval_string ("gnc:option-index-get-value");
     getters.value_to_index = scm_c_eval_string ("gnc:option-value-get-index");
@@ -928,21 +925,9 @@ gnc_option_permissible_value_name (GNCOption *option, int index)
 char *
 gnc_option_permissible_value_description (GNCOption *option, int index)
 {
-    SCM help;
-
-    if (index < 0)
-        return NULL;
-
     initialize_getters ();
 
-    help = scm_call_2 (getters.index_to_description, option->guile_option,
-                       scm_from_int (index));
-    if (help == SCM_UNDEFINED)
-        return NULL;
-    if (!scm_is_string (help))
-        return NULL;
-
-    return gnc_scm_to_utf8_string (help);
+    return NULL;
 }
 
diff --git a/libgnucash/app-utils/options.scm b/libgnucash/app-utils/options.scm
index 2dba75358..b8699c719 100644
--- a/libgnucash/app-utils/options.scm
+++ b/libgnucash/app-utils/options.scm
@@ -89,7 +89,6 @@
 (export gnc:option-generate-restore-form)
 (export gnc:option-get-value)
 (export gnc:option-getter)
-(export gnc:option-index-get-description)
 (export gnc:option-index-get-name)
 (export gnc:option-index-get-value)
 (export gnc:option-kvp->scm)
@@ -298,11 +297,6 @@ the option '~a'."))
          (name-fn (vector-ref option-data-fns 2)))
     (name-fn index)))
 
-(define (gnc:option-index-get-description option index)
-  (let* ((option-data-fns (gnc:option-data-fns option))
-         (name-fn (vector-ref option-data-fns 3)))
-    (name-fn index)))
-
 (define (gnc:option-index-get-value option index)
   (let* ((option-data-fns (gnc:option-data-fns option))
          (name-fn (vector-ref option-data-fns 1)))
@@ -1047,8 +1041,7 @@ the option '~a'."))
     (if (null? p-vals)
         '()
         (cons (vector-ref (car p-vals) 1)
-              (cons (vector-ref (car p-vals) 2)
-                    (multichoice-strings (cdr p-vals))))))
+              (multichoice-strings (cdr p-vals)))))
 
   (let* ((value default-value)
          (value->string (lambda ()
@@ -1084,7 +1077,7 @@ the option '~a'."))
      (vector (lambda () (length ok-values))
              (lambda (x) (vector-ref (list-ref ok-values x) 0))
              (lambda (x) (vector-ref (list-ref ok-values x) 1))
-             (lambda (x) (vector-ref (list-ref ok-values x) 2))
+             #f                         ;old tooltip
              (lambda (x)
                (gnc:multichoice-list-lookup ok-values x)))
      (lambda () (multichoice-strings ok-values)) 
@@ -1139,8 +1132,7 @@ the option '~a'."))
     (if (null? p-vals)
         '()
         (cons (vector-ref (car p-vals) 1)
-              (cons (vector-ref (car p-vals) 2)
-                    (radiobutton-strings (cdr p-vals))))))
+              (radiobutton-strings (cdr p-vals)))))
 
   (let* ((value default-value)
          (value->string (lambda ()
@@ -1170,7 +1162,7 @@ the option '~a'."))
      (vector (lambda () (length ok-values))
              (lambda (x) (vector-ref (list-ref ok-values x) 0))
              (lambda (x) (vector-ref (list-ref ok-values x) 1))
-             (lambda (x) (vector-ref (list-ref ok-values x) 2))
+             #f                         ;old tooltip
              (lambda (x)
                (gnc:multichoice-list-lookup ok-values x)))
      (lambda () (radiobutton-strings ok-values)) 
@@ -1204,8 +1196,7 @@ the option '~a'."))
     (if (null? p-vals)
         '()
         (cons (vector-ref (car p-vals) 1)
-              (cons (vector-ref (car p-vals) 2)
-                    (list-strings (cdr p-vals))))))
+              (list-strings (cdr p-vals)))))
 
   (let* ((value default-value)
          (value->string (lambda ()
@@ -1242,7 +1233,7 @@ the option '~a'."))
      (vector (lambda () (length ok-values))
              (lambda (x) (vector-ref (list-ref ok-values x) 0))
              (lambda (x) (vector-ref (list-ref ok-values x) 1))
-             (lambda (x) (vector-ref (list-ref ok-values x) 2))
+             #f                         ;old tooltip
              (lambda (x) (gnc:multichoice-list-lookup ok-values x)))
      (lambda () (list-strings ok-values)) #f)))
 
```

christopherlam:
And another patch to add deprecation warnings for custom reports.

```patch
diff --git a/libgnucash/app-utils/options.scm b/libgnucash/app-utils/options.scm
index b8699c719..83dfa376c 100644
--- a/libgnucash/app-utils/options.scm
+++ b/libgnucash/app-utils/options.scm
@@ -994,6 +994,14 @@ the option '~a'."))
         (rpterror-earlier "multichoice" item (car full-lst))
         0)))
 
+(define (check-ok-values ok-values fn)
+  (for-each
+   (lambda (ok-value)
+     (when (> (vector-length ok-value) 2)
+       (issue-deprecation-warning
+        (format #f "~a: the tooltip in ~a is not supported anymore. Please remove." fn ok-value))))
+   ok-values))
+
 ;; multichoice options use the option-data as a list of vectors.
 ;; Each vector contains a permissible value (scheme symbol), a
 ;; name, and a description string.
@@ -1043,6 +1051,8 @@ the option '~a'."))
         (cons (vector-ref (car p-vals) 1)
               (multichoice-strings (cdr p-vals)))))
 
+  (check-ok-values ok-values "gnc:make-multichoice-[callback-]option")
+
   (let* ((value default-value)
          (value->string (lambda ()
                           (string-append "'" (gnc:value->string value)))))
@@ -1198,6 +1208,8 @@ the option '~a'."))
         (cons (vector-ref (car p-vals) 1)
               (list-strings (cdr p-vals)))))
 
+  (check-ok-values ok-values "gnc:make-list-option")
+
   (let* ((value default-value)
          (value->string (lambda ()
                           (string-append "'" (gnc:value->string value)))))
```

Bob-IT:
@christopherlam  Thank you for the scheme changes, just applied locally and noticed that you have removed the tool tips from the radio button option. As far as I can see it is only used in hello-world.scm so no big change but I have not removed them from the C code.
@jralls , @gjanssens
I am just confirming this is what we want and I will make the changes.

fellen:
> removed the tool tips from the radio button option

I would  prefer to keep them.

christopherlam:
> > removed the tool tips from the radio button option
> 
> I would prefer to keep them.

Why? They are not used in reports at all, cause only clutter.

fellen:
> Why? They are not used in reports at all, cause only clutter.

From my perspective the question is more: Could the UX get improved by using them?

christopherlam:
> > Why? They are not used in reports at all, cause only clutter.
> 
> From my perspective the question is more: Could the UX get improved by using them?

Nobody has found any good use of tooltips on radio buttons in options for 20 years. Is now a good time to think about saving them just when we're about to do some long overdue cleanup?

Bob-IT:
I think I have completed the changes required with the help of Chris, will push this at the weekend if there are no concerns.

gnucash/report/reports/standard/trial-balance.scm
```diff
@@ -200,15 +200,9 @@
...
-                    (N_ "Uses the exact balances in the general journal"))
-            (vector 'pre-adj
-                    (N_ "Pre-adjustment Trial Balance")
-                    (N_ "Ignores Adjusting/Closing entries"))
-            (vector 'work-sheet
-                    (N_ "Work Sheet")
-                    (N_ "Creates a complete end-of-period work sheet")))))
+      (list (vector 'current (N_ "General journal exact balances"))
+            (vector 'pre-adj (N_ "No adjusting/closing entries"))
+            (vector 'work-sheet (N_ "Full end-of-period work sheet")))))
```
fellen, line 205: I wonder if translators need some explanation.

fellen (commented):
It seems to be primary a nice simplification, but I added a few notes inline.

gnucash/report/reports/locale-specific/de_DE/taxtxf.scm
```diff
@@ -151,24 +151,16 @@
...
-          (vector '2nd-last
-                  (N_ "Last Yr 2nd Est Tax Qtr")
-                  (N_ "Apr 1 - May 31, Last year."))
-          (vector '3rd-last
-                  (N_ "Last Yr 3rd Est Tax Qtr")
-                  (N_ "Jun 1 - Aug 31, Last year."))
-          (vector '4th-last
-                  (N_ "Last Yr 4th Est Tax Qtr")
-                  (N_ "Sep 1 - Dec 31, Last year.")))))
+    (list (vector 'from-to (N_ "Use From - To"))
```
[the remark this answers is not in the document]
christopherlam: I never use txf reports. I'll happily leave it to bitrot and let someone else make that call.

bzbarsky:
@Bob-IT It looks like this change made existing (pre-4.6) saved reports with non-default account level settings selected not work properly, because the saved reports have things like:
```
(let ((option (gnc:lookup-option options
                                 "Accounts"
                                 "Show Accounts until level")))
  ((lambda (o) (if o (gnc:option-set-value o '3))) option))
```
but the option got renamed to `"Levels of Subaccounts"` and hence the setting is not getting applied anymore.  Same for reports which just happen to be open before upgrading to 4.6: they effectively reset their account level setting to the default.

Was that behavior change intended?

christopherlam:
> Was that behavior change intended?

It means that https://github.com/Gnucash/gnucash/blob/23bd7164511bf26af231097d2ee10da78fdd1fb7/libgnucash/app-utils/options.scm#L1797 should have received an option rename entry thus:

```scheme
   ("Show Accounts until level" #f "Levels of Subaccounts")
```

christopherlam:
Fixed with a5edacf

bzbarsky:
@christopherlam Thank you!

fellen:
> @fellen, @Bob-IT forgot to update `budget.scm`
> 
> This is a preliminary patch; I'll leave it to you to push the required changes. I'd happily reuse the existing names, but I feel you're much more opiniated on replacement names.
> 
> ```diff
> modified   gnucash/report/reports/standard/budget.scm
> @@ -108,24 +108,12 @@
>            (lambda (new-option)
>              (gnc:register-option options new-option)))
>           (period-options
> -          (list (vector 'first
> -                        (N_ "First")
> -                        (N_ "The first period of the budget"))
> -                (vector 'previous
> -                        (N_ "Previous")
> -                        (N_ "Budget period was before current period, according to report evaluation date"))
> -                (vector 'current
> -                        (N_ "Current")
> -                        (N_ "Current period, according to report evaluation date"))
> -                (vector 'next
> -                        (N_ "Next")
> -                        (N_ "Next period, according to report evaluation date"))
> -                (vector 'last
> -                        (N_ "Last")
> -                        (N_ "Last budget period"))
> -                (vector 'manual
> -                        (N_ "Manual period selection")
> -                        (N_ "Explicitly select period value with spinner below"))))
> +          (list (vector 'first (N_ "First period"))
> +                (vector 'previous (N_ "Previous period"))
> +                (vector 'current (N_ "Current period"))
> +                (vector 'next (N_ "Next period"))
> +                (vector 'last (N_ "Last period"))
> +                (vector 'manual (N_ "Manual period selection"))))
>           (ui-use-periods #f)
>           (ui-start-period-type 'current)
>           (ui-end-period-type 'next))
> ```

No objections.

gnucash/report/options-utilities.scm
```diff
@@ -63,14 +63,13 @@
...
-	  (vector 'HalfYearDelta (N_ "Half Year") (N_ "Half Year."))
-	  (vector 'YearDelta (N_ "Year") (N_ "One Year."))
-	  ))))
+    (list (vector 'DayDelta (N_ "One Day"))
+          (vector 'WeekDelta (N_ "One Week"))
+          (vector 'TwoWeekDelta (N_ "Two Weeks"))
+          (vector 'MonthDelta (N_ "One Month"))
+          (vector 'QuarterDelta (N_ "One Quarter"))
+          (vector 'HalfYearDelta (N_ "Half Year"))
+          (vector 'YearDelta (N_ "One Year"))))))
```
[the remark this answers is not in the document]
flywire: Seems Fiscal Year is missing from this list

PR #919, opened 2021-02-21
Use init.py from user data directory if it exists

tunedal:
To make it easier for the user to customize `init.py`, allow it to be loaded from the user data directory (e.g. `~/.local/share/gnucash`) rather than the system directory, where it's normally owned by root and managed by the system package manager.

If the user doesn't have an `init.py` file, the old system-wide `init.py` (now renamed to `default_init.py`) is run instead, preserving the existing behavior by default. Both the system-wide and the user-specific python directories are added to sys.path, in order to allow importing pycons, etc.

I've done this in two separate commits because otherwise git fails to detect the rename and "git blame" loses track of history for `default_init.py`.


jralls:
The file selection should be done in gnucash/python/gncmod-python.c; that would also remove the need to rename init.py. ISTM it would be much safer if the path was set via an environment variable or a command line option instead of being hard-coded to $GNC_DATA_DIR/python/init.py; in spite of the Guile precedent $GNC_DATA_DIR isn't really an appropriate place for code.

fellen:
Perhaps @c-holtermann can help?

tunedal:
I noticed (after submitting this PR) that a `config-user.scm` file is loaded from `gnc_userconfig_home`, so perhaps a `config-user.py` could be loaded from the same directory?

jralls:
I'd prefer not. $GNC_CONFIG_HOME isn't any more appropriate than $GNC_DATA_HOME. Searching $PYTHONPATH for a file named `gnucash-init.py` would be less awful.

gjanssens:
> I'd prefer not. $GNC_CONFIG_HOME isn't any more appropriate than $GNC_DATA_HOME. Searching $PYTHONPATH for a file named `gnucash-init.py` would be less awful.

+1

And perhaps we should do the same for our custom guile code loading. That is, deprecate the use of config-user.scm and instead have gnucash look for a gnucash-init.scm somewhere on the guile-load-path and/or guile-load-compiled-path.

fellen:
> > I'd prefer not. $GNC_CONFIG_HOME isn't any more appropriate than $GNC_DATA_HOME. Searching $PYTHONPATH for a file named `gnucash-init.py` would be less awful.
> 
> +1
> 
> And perhaps we should do the same for our custom guile code loading. That is, deprecate the use of config-user.scm and instead have gnucash look for a gnucash-init.scm somewhere on the guile-load-path and/or guile-load-compiled-path.

I am not against it, but after decades of obsolete documentation we have the current state in wiki's Configuration_Locations and Flatpak.

I would ask not to forget to update at least the wiki.
@CWehli is working on an update of gnucash-help partialy based on wiki sections.

jralls:
@c-holtermann Do you have an opinion on this?

c-holtermann:
I like the idea of making  `init.py` more accessible. I only now find a bit of time looking into it. I'll give feedback in the next days.

c-holtermann:
As I have a little time now to look into gnucash I can come back here and we can hopefully finish this soon.

I have a suggestion similar to this PR that I have put in my tree: https://github.com/c-holtermann/gnucash/commit/c6a2347cd771e762a8a8bb25d902705df0940114

I would not replace the init.py but suggest loading a user file in addition to it. Any thoughts?

The suggestion could be extended by a return value of that script. Either skip the rest of init.py, abort gnucash completely or just continue. [Not implemented yet]

jralls:
> I would not replace the init.py but suggest loading a user file in addition to it. Any thoughts?

My first thought is that I'm not even sure there's a good use case for having a Python REPL running in a GnuCash session, but if you've got that REPL going then it's pretty simple to add whatever directory you want to `sys.path` and import your module. If you don't want a Python REPL window then it makes sense for GnuCash to load your init.py instead of the built-in one. If you just want to do some prep before starting the REPL then you can start it from your init.py.

My next thought is that modifying gnucash/python/init.py for this purpose is a bit of a hack: It should be done in `libgncmod_python_gnc_module_init`.

And one more thing: It would be much better to launch the Python console on demand from a menu item than automatically on startup.

PR #920, opened 2021-02-21
Update price database for imported transactions

dl-g:
When a transaction is added from the ledger, price database is updated properly.
But if the transaction is imported, there is no price db update.

This change adds the proper pricedb update in the import path (qfx/ofx/qif).
Tested with make check

### DIFF
--- a/gnucash/import-export/qif-imp/qif-to-gnc.scm
+++ b/gnucash/import-export/qif-imp/qif-to-gnc.scm
@@ -445,7 +445,7 @@
 
                   ;; rebalance and commit everything
                   (xaccTransCommitEdit gnc-xtn)
-		  (xaccTransRecordPrice gnc-xtn))))
+                  (xaccTransRecordPrice gnc-xtn))))
           (qif-file:xtns qif-file)))
        sorted-qif-files-list)
 
--- a/gnucash/import-export/qif-imp/qif-to-gnc.scm
+++ b/gnucash/import-export/qif-imp/qif-to-gnc.scm
@@ -445,7 +445,7 @@
 
                   ;; rebalance and commit everything
                   (xaccTransCommitEdit gnc-xtn)
-                  (xaccTransRecordPrice gnc-xtn))))
+                  (xaccTransRecordPrice gnc-xtn PRICE-SOURCE-SPLIT-IMPORT))))
           (qif-file:xtns qif-file)))
        sorted-qif-files-list)
 
--- a/gnucash/import-export/qif-imp/qif-to-gnc.scm
+++ b/gnucash/import-export/qif-imp/qif-to-gnc.scm
@@ -444,7 +444,8 @@
                                                  progress-dialog)
 
                   ;; rebalance and commit everything
-                  (xaccTransCommitEdit gnc-xtn))))
+                  (xaccTransCommitEdit gnc-xtn)
+		  (xaccTransRecordPrice gnc-xtn))))
           (qif-file:xtns qif-file)))
        sorted-qif-files-list)
 
--- a/libgnucash/engine/Transaction.c
+++ b/libgnucash/engine/Transaction.c
@@ -3116,25 +3116,27 @@ record_price(Split *split,
     time64 time;
     gboolean swap;
 
-    account = xaccSplitGetAccount(split);
-    if (!xaccAccountIsPriced(account)) {
+    account = xaccSplitGetAccount (split);
+    if (!xaccAccountIsPriced (account))
+    {
        return;
     }
-    amount = xaccSplitGetAmount(split);
-    if (gnc_numeric_zero_p(amount)) {
+    amount = xaccSplitGetAmount (split);
+    if (gnc_numeric_zero_p (amount))
+    {
        return;
     }
-    trans = xaccSplitGetParent(split);
-    value = gnc_numeric_div(xaccSplitGetValue(split), amount,
-                            GNC_DENOM_AUTO,
-                            GNC_HOW_DENOM_EXACT);
+    trans = xaccSplitGetParent (split);
+    value = gnc_numeric_div (xaccSplitGetValue (split), amount,
+                             GNC_DENOM_AUTO,
+                             GNC_HOW_DENOM_EXACT);
     book = qof_instance_get_book (QOF_INSTANCE (account));
     pricedb = gnc_pricedb_get_db (book);
     comm = xaccAccountGetCommodity (account);
     curr = xaccTransGetCurrency (trans);
     scu = gnc_commodity_get_fraction (curr);
     swap = FALSE;
-    time = xaccTransGetDate(trans);
+    time = xaccTransGetDate (trans);
     price = gnc_pricedb_lookup_day_t64 (pricedb, comm, curr, time);
     if (gnc_commodity_equiv (comm, gnc_price_get_currency (price)))
         swap = TRUE;
@@ -3148,7 +3150,7 @@ record_price(Split *split,
             gnc_price_unref (price);
             return;
         }
-        if (gnc_price_get_source (price) < PRICE_SOURCE_XFER_DLG_VAL)
+        if (gnc_price_get_source (price) < PRICE_SOURCE_SPLIT_IMPORT)
         {
             /* Existing price is preferred over this one. */
             gnc_price_unref (price);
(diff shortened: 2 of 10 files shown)


gnucash/import-export/qif-imp/qif-to-gnc.scm
```diff
@@ -444,7 +444,8 @@
                                                  progress-dialog)
 
                   ;; rebalance and commit everything
-                  (xaccTransCommitEdit gnc-xtn))))
+                  (xaccTransCommitEdit gnc-xtn)
+		  (xaccTransRecordPrice gnc-xtn))))
```
christopherlam, line 448: Use spaces, not tabs.

libgnucash/engine/Transaction.c
```diff
@@ -3100,5 +3100,106 @@ _utest_trans_fill_functions (void)
...
+    GNCPrice* price;
+    gnc_numeric price_value, value, amount;
+    int scu;
+    time64 time;
+    gboolean swap;
+
+    account = xaccSplitGetAccount(split);
+    if (!xaccAccountIsPriced(account)) {
+       return;
+    }
```
christopherlam, line 3122: See https://wiki.gnucash.org/wiki/CodingStandard - no dangling `{` `}` and a space before `(`

```c
    if (!xaccAccountIsPriced (account))
    {
       return;
    }
```

libgnucash/engine/Transaction.c
```diff
@@ -3100,5 +3100,106 @@ _utest_trans_fill_functions (void)
...
+    if (price)
+    {
+        price_value = gnc_price_get_value (price);
+        if (gnc_numeric_equal (swap ? gnc_numeric_invert (value) : value,
+                               price_value))
+        {
+            gnc_price_unref (price);
+            return;
+        }
+        if (gnc_price_get_source (price) < PRICE_SOURCE_XFER_DLG_VAL)
```
christopherlam, line 3151: I'd think better compare with individual enums instead of relying that the existing price source is less than a threshold?

libgnucash/engine/Transaction.h
```diff
@@ -767,6 +767,12 @@ time64 xaccTransGetVoidTime(const Transaction *tr);
 void xaccTransDump (const Transaction *trans, const char *tag);
 #endif
 
+/** The xaccTransRecord() method iterates through the splits and
+ *  and record the non-currency equivalent prices in the price database.
+ */
+void xaccTransRecordPrice (Transaction *trans);
```
christopherlam, line 773: Wouldn't it be better named `xaccTransUpdatePrices` ?

christopherlam (changes requested):
A few stylistic comments. This new behaviour may be a surprise to some. Would it benefit from an import option?

libgnucash/engine/Transaction.c
```diff
@@ -3100,5 +3100,106 @@ _utest_trans_fill_functions (void)
...
+   int i = 0;
+
+    /* XXX: move this into xaccSplitCommitEdit and other callers of
+    * gnc_pricedb_add_price from split-register.c, gnc-imp-props-price.cpp
+    * etc.
+    */
+   while ((s = xaccTransGetSplit(trans, i)) != NULL) {
+      record_price(s, PRICE_SOURCE_SPLIT_IMPORT);
+      i++;
+   }
```
christopherlam, line 3201: This is O(N^2) -- better:
```c
for (GList *n = xaccTransGetSplitList (trans); n; n = n->next)
  record_price (n->data, PRICE_SOURCE_SPLIT_IMPORT);
```

gnucash/import-export/qif-imp/qif-to-gnc.scm
dl-g: fixed.

libgnucash/engine/Transaction.c
dl-g: fixed.

libgnucash/engine/Transaction.c
dl-g: If I understand the code correctly, these enums are in order of preference.  So if the existing price source is preferred over this imported source,  we skip the update.

libgnucash/engine/Transaction.h
dl-g: I stole this code from John's original piece in split-register.c:record_price.  These are most likely new entries to be added to the database. So record is slightly more accurate than update? Let me know if I should rename it.

libgnucash/engine/Transaction.c
dl-g: Great catch. Thanks!

dl-g:
> This new behaviour may be a surprise to some. Would it benefit from an import option?

Thanks for the review.  If the user expects a db update from a manual ledger entry,  would the same expectation hold for imported transactions?  I was originally surprised and puzzled by those missing entries. It's quite tedious that I had to manually update the database, especially for those proprietary 401k mutual funds that have no FQ source.

libgnucash/engine/Transaction.h
jralls: So instead of copying you should move the code and replace all of the calls to `record_price` in `split-register.c` with `xaccTransRecordPrice`.

libgnucash/engine/Transaction.c
jralls: @dl-g is correct, it's a precedence check.

jralls:
> Thanks for the review. If the user expects a db update from a manual ledger entry, would the same expectation hold for imported transactions? I was originally surprised and puzzled by those missing entries. It's quite tedious that I had to manually update the database, especially for those proprietary 401k mutual funds that have no FQ source.

That's right. The original change was to prevent having no price for a commodity, which causes balances in that commodity to not get valued. The same logic would apply when the commodity is introduced via an import. An option is no more necessary for an imported transaction than for one created in the GUI.

libgnucash/engine/Transaction.h
dl-g: Sure. Will do. Thanks!

libgnucash/engine/Transaction.h
```diff
@@ -767,10 +768,10 @@ time64 xaccTransGetVoidTime(const Transaction *tr);
 void xaccTransDump (const Transaction *trans, const char *tag);
 #endif
 
-/** The xaccTransRecord() method iterates through the splits and
+/** The xaccTransRecordPrice() method iterates through the splits and
  *  and record the non-currency equivalent prices in the price database.
  */
```
jralls, line 773: You need to document the parameters, e.g.
```
* @param trans The transaction whose price is recorded
* @param source The price priority level
```

libgnucash/engine/Transaction.c
```diff
@@ -3188,14 +3188,14 @@ record_price(Split *split,
...
+xaccTransRecordPrice (Transaction *trans, PriceSource source)
 {
     /* XXX: move this into xaccSplitCommitEdit and other callers of
     * gnc_pricedb_add_price from split-register.c, gnc-imp-props-price.cpp
     * etc.
     */
    for (GList *n = xaccTransGetSplitList (trans); n; n = n->next)
-      record_price (n->data, PRICE_SOURCE_SPLIT_IMPORT);
+      record_price (n->data, source);
 }
```
jralls, line 3199: Please move both function definitions up to line 2940, above the QofObject stuff.

libgnucash/engine/Transaction.h
dl-g: fixed

libgnucash/engine/Transaction.c
dl-g: sure. thanks!

jralls:
@dl-g One more issue: You made this PR for master, meaning that users won't see it until 5.0 in about 2 years. I don't see any reason not to merge it into maint instead so that it will be in 4.5 release at the end of the month. Do you?

dl-g:
Oh, I didn't know how the different branches are tied to releases. Thought master would be on a faster track than maint.  There should not be an issue for this change to go into 4.5. I have been running it in 3.8 for two months now. Thanks!

christopherlam:
75209c024feb209b100b0bff3b6dacee12c7c8ca closes this PR. Thanks @dl-g !

PR #930, opened 2021-03-06
From Budget editor, add toolbar and Edit menu to run budget report

christopherlam:
This parallels the invoice editor "Print Invoice" functionality. Runs the budget report using current budget.

image

### DIFF
--- a/gnucash/gnome/gnc-plugin-page-budget.c
+++ b/gnucash/gnome/gnc-plugin-page-budget.c
@@ -263,7 +263,9 @@ typedef struct GncPluginPageBudgetPrivate
     gnc_numeric allValue;
     allperiods_action action;
 
-    /* the cached reportPage for this budget */
+    /* the cached reportPage for this budget. note this is not saved
+       into .gcm file therefore the budget editor->report link is lost
+       upon restart. */
     GncPluginPage *reportPage;
 } GncPluginPageBudgetPrivate;
 
@@ -1284,6 +1286,12 @@ gnc_plugin_page_budget_cmd_budget_note(GtkAction *action,
     g_object_unref(G_OBJECT(builder));
 }
 
+/* From the budget editor, open the budget report. This will reuse the
+   budget report if generated from the current budget editor. Note the
+   reuse is lost when GnuCash is restarted. This link may be restored
+   by: scan the current session tabs, identify reports, checking
+   whereby report's report-type matches a budget report, and the
+   report's budget option value matches the current budget. */
 static void
 gnc_plugin_page_budget_cmd_budget_report (GtkAction *action,
                                           GncPluginPageBudget *page)
--- a/gnucash/gnome/gnc-plugin-page-budget.c
+++ b/gnucash/gnome/gnc-plugin-page-budget.c
@@ -42,7 +42,11 @@
 #include <glib/gi18n.h>
 #include "gnc-date-edit.h"
 
+#include "swig-runtime.h"
+#include "libguile.h"
+
 #include "gnc-plugin-page-register.h"
+#include "gnc-plugin-page-report.h"
 #include "gnc-budget.h"
 #include "gnc-features.h"
 
@@ -127,6 +131,8 @@ static void gnc_plugin_page_budget_cmd_refresh (GtkAction *action,
                                                 GncPluginPageBudget *page);
 static void gnc_plugin_page_budget_cmd_budget_note (GtkAction *action,
                                                     GncPluginPageBudget *page);
+static void gnc_plugin_page_budget_cmd_budget_report (GtkAction *action,
+                                                      GncPluginPageBudget *page);
 static void allperiods_budget_helper (GtkTreeModel *model, GtkTreePath *path,
                                       GtkTreeIter *iter, gpointer data);
 
@@ -177,6 +183,12 @@ static GtkActionEntry gnc_plugin_page_budget_actions [] =
         N_("Edit note for the selected account and period"),
         G_CALLBACK (gnc_plugin_page_budget_cmd_budget_note)
     },
+    {
+        "BudgetReportAction", "system-run", N_("Budget Report"),
+        NULL,
+        N_("Run budget report"),
+        G_CALLBACK (gnc_plugin_page_budget_cmd_budget_report)
+    },
     /* View menu */
     {
         "ViewFilterByAction", NULL, N_("_Filter By..."), NULL, NULL,
@@ -211,6 +223,7 @@ static action_toolbar_labels toolbar_labels[] =
     { "EstimateBudgetAction",       N_("Estimate") },
     { "AllPeriodsBudgetAction",     N_("All Periods") },
     { "BudgetNoteAction",           N_("Note") },
+    { "BudgetReportAction",         N_("Run Report") },
     { NULL, NULL },
 };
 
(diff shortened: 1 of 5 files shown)


Bob-IT:
Looks OK to me but not sure on toolbar icon, maybe it should be 'system-run' and as in area edit note should be 'text-editor-symbolic'
Another thing, what would you expect if pressed multiple times?
Should it just keep opening new report pages or should it rerun the report in the page if available. I am not sure and I suppose as this parallels the invoice setup should the same reasoning apply.

christopherlam:
> Looks OK to me but not sure on toolbar icon, maybe it should be 'system-run' and as in area edit note should be 'text-editor-symbolic'

Ugh. This does not match the style!
image

> Another thing, what would you expect if pressed multiple times?
> Should it just keep opening new report pages or should it rerun the report in the page if available. I am not sure and I suppose as this parallels the invoice setup should the same reasoning apply.

The invoice editor does open multiple reports. Don't particularly see it as a wart myself...

Bob-IT:
@christopherlam That icon did not look so bold on my system so try 'text-x-generic' instead or just forget about the change.

As I said, I have no idea if pressing 'Run Report' button multiple times and creating many pages is an issue or not, just some thing I noticed. I suppose one could run the report, notice something needs changing, go back to budget or invoice, run the report again, get distracted and print the old one but that maybe just too convoluted to care about.

gjanssens:
> As I said, I have no idea if pressing 'Run Report' button multiple times and creating many pages is an issue or not, just some thing I noticed. I suppose one could run the report, notice something needs changing, go back to budget or invoice, run the report again, get distracted and print the old one but that maybe just too convoluted to care about.

IMO the button should switch to an already open report.

christopherlam:
I think this is now satisfactory.

christopherlam:
image

Alternative icons for 'Estimate' and 'All Periods' would be welcome to IMHO

christopherlam:
Maybe clicking 'Run Report' and switching to the existing Budget Report tab should also trigger a report refresh.

Bob-IT:
I do not think switching should reload report as this will increase the time stepping through the tabs but you could add the following to gnc-plugin-page-report.c at line 1430 with appropriate declaration in the .h file...

void
gnc_plugin_page_report_reload (GncPluginPageReport *report)
{
    gnc_plugin_page_report_reload_cb (NULL, report);
}
and then add after your open window in gnc-plugin-page-budget.c...
    gnc_plugin_page_report_reload (GNC_PLUGIN_PAGE_REPORT(priv->reportPage));

christopherlam:
I think this is now complete.

Interesting observation: keep the generated Budget Report tab, switch to Budget Editor, modify a budget cell, and click Run Report again. This refreshes the report, *then* switches the tab. However webkit's multitasking takes over and the report update appears very shortly *after* the tab switching.

For obvious reasons, the budget editor -> budget report linkage is not saved when the tabs are reloaded after restarting GnuCash.

Bob-IT:
> For obvious reasons, the budget editor -> budget report linkage is not saved when the tabs are reloaded after restarting GnuCash.

It maybe possible using report_id which seems to be unique. So one could possibly save the report_id associated to the budget page as part of the normal budget page save and then on load record the report_id to use. Then use this report_id in the run report command function that cycles through the report pages looking for the report_id that matches and then do the update on that page, if not found create a new on. I would have to try it to be sure...

christopherlam:
@Bob-IT I'm not convinced we can save the `report_id` because they seem to be unique per session. So this can be merged and you can experiment anyway off the main branch?

gjanssens:
> I think this is now complete.
> 
> Interesting observation: keep the generated Budget Report tab, switch to Budget Editor, modify a budget cell, and click Run Report again. This refreshes the report, _then_ switches the tab. However webkit's multitasking takes over and the report update appears very shortly _after_ the tab switching.
> 
> For obvious reasons, the budget editor -> budget report linkage is not saved when the tabs are reloaded after restarting GnuCash.

What are the "obvious"  reasons ?

Bob-IT (approved):
@christopherlam  Looks OK to me.

I had a play with report_id and it sort of worked. The id is saved as part of the open report and can be seen in the .gcm file. But if a report is closed and a new report is opened it would possibly use the same id.

The only way I can think of to make this work is if there were possibly a way to retrieve the budget guid from the report. So one would get a list of open report pages, execute a function against the page that tries to retrieve the budget guid, if successful it then gets compared to the guid in the budget page. In theory the same could happen on invoice pages but with a different function retrieving the invoice guid.

I am not sure it is worth doing but was worth exploring.

gjanssens:
> > For obvious reasons, the budget editor -> budget report linkage is not saved when the tabs are reloaded after restarting GnuCash.
> 
> ~What are the "obvious" reasons ?~
> Never mind. Reading the code does make it obvious...

Anyway having  said  that, I don't agree with your premise. You are looking at it from the wrong angle.

In your code you save a reference to the report you have opened for this budget. That reference is obviously gone after restarting gnucash. But there is a piece of information that is unique and persisting: the budget guid. It's used to find the budget when  creating the budget page, it's used by the budget report's options to figure out which budget to report on.

So what you can do - if the user clicks the open report button, loop through all open reports and check each report whether it's a report for the current budget based on the budget guid. If so, switch to that report.

That's also how for example the account hierachy page code works if you double-click on an account. It will loop through all account register pages that are currently open and jumps to  one that matches or creates a new one if none matches.

christopherlam:
> Anyway having said that, I don't agree with your premise. You are looking at it from the wrong angle.
> 
> In your code you save a reference to the report you have opened for this budget. That reference is obviously gone after restarting gnucash. But there is a piece of information that is unique and persisting: the budget guid. It's used to find the budget when creating the budget page, it's used by the budget report's options to figure out which budget to report on.

But then we'd scan all tabs whereby `(report-type == budget) AND (report-option->budget == current budget)`. This will give a false positive if the user has created their own budget report using same budget. Plus lots of scheme<->C code to identify the correct tab. I think it's overthinking it.

gjanssens:
> > Anyway having said that, I don't agree with your premise. You are looking at it from the wrong angle.
> > In your code you save a reference to the report you have opened for this budget. That reference is obviously gone after restarting gnucash. But there is a piece of information that is unique and persisting: the budget guid. It's used to find the budget when creating the budget page, it's used by the budget report's options to figure out which budget to report on.
> 
> But then we'd scan all tabs whereby `(report-type == budget) AND (report-option->budget == current budget)`.

Indeed. That won't take much time though.

>  This will give a false positive if the user has created their own budget report using same budget.

The question is whether that's really a false positive or rather an added benefit ? Unfortunately there could be use cases for both behaviours. So I agree to just stick with what you have come up with so far. But perhaps do add a comment somewhere that explains the limitation of this  implementation (namely that you lose the link between open budget and open budget report on restart) and that that was a conscious design choice.

gnucash/gnome/gnc-plugin-page-budget.c
```diff
@@ -172,11 +178,17 @@ static GtkActionEntry gnc_plugin_page_budget_actions [] =
...
-        "BudgetNoteAction", "system-run", N_("Edit Note"),
+        "BudgetNoteAction", "text-x-generic", N_("Edit Note"),
         NULL,
         N_("Edit note for the selected account and period"),
         G_CALLBACK (gnc_plugin_page_budget_cmd_budget_note)
     },
+    {
+        "BudgetReportAction", "system-run", N_("Budget Report"),
+        NULL,
+        N_("Run budget report"),
```
fellen, line 189: Translation toolkit will complain a missing full stop and a user might want an answer to "what will happen, if I press the button a second time?"

PR #931, opened 2021-03-08
Replace g_memdup with g_memdup2

bhardwajs:
`g_memdup` has been deprecated and the recommendation from Gnome team is to port to `g_memdup2`. This PR does that.

I am merging to master as `g_memdup2` is in GLlib >= 2.67.3.
https://discourse.gnome.org/t/port-your-module-from-g-memdup-to-g-memdup2-now/5538

### DIFF
--- a/gnucash/register/register-gnome/gnucash-style.c
+++ b/gnucash/register/register-gnome/gnucash-style.c
@@ -60,7 +60,7 @@ style_create_key (SheetBlockStyle *style)
 
     key = style->cursor->num_rows;
 
-    return g_memdup(&key, sizeof(key));
+    return g_memdup2(&key, sizeof(key));
 }
 
 static void
--- a/libgnucash/app-utils/gnc-ui-balances.c
+++ b/libgnucash/app-utils/gnc-ui-balances.c
@@ -374,7 +374,7 @@ static void sack_foreach_func(gpointer key, gpointer value, gpointer user_data)
     gnc_numeric reachable_value = gnc_numeric_add_fixed (thisvalue, data->split_value);
 
     data->reachable_list = g_list_prepend
-        (data->reachable_list, g_memdup (&reachable_value, sizeof (gnc_numeric)));
+        (data->reachable_list, g_memdup2 (&reachable_value, sizeof (gnc_numeric)));
 }
 
 GList *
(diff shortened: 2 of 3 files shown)


bhardwajs:
CI build is failing because `g_memdup2` is not present on glib version on CI machines. Recommendation is to copy the static inline version from this commit:
https://gitlab.gnome.org/GNOME/glib/-/merge_requests/1927/diffs?commit_id=5e5f75a77e399c638be66d74e5daa8caeb433e00#cea947955c282af60c41ec9c23c858cb3945d26d

Question on Gnucash end is whether we expect an overflow or not? Given our input, this might be a moot issue in which case this change can wait till GLib version used is upgraded.

jralls:
@bhardwajs 
With all due respect to you and Phil Withnall, this is nuts. 2.67 is an unstable series and it will be at the very best months before 2.68 makes it into cutting-edge distributions like Arch Linux and years for LTS ones like Ubuntu and Debian stable. Deprecating in one dot release of an unstable series and removing in the following one is going to break a ton of code and piss off just about everyone. It's stupid. However, since it's presented as a security vulnerability (the amount of memory being copied has to be enough to overflow a 32-bit uint, not exactly common and certainly not happening in GnuCash) it's possible that distros will try to make security patches. They'll probably screw it up.

Patching GLib in the test containers isn't practical and besides ignores the problem that we can't do that in the real world.

All of that aside, there are two ways to make this an acceptable PR. The easier one is to add a configuration check for `g_memdup2` being available and #ifdef accordingly. The other, arguably better, one is to replace the 4 `g_memdup` instances with `memcpy` and whatever boilerplate is needed to make that work.

The best, of course, is to rewrite the modules in C++ so that they don't use either `g_memdup` or `memcpy`, but that's a lot to ask for in a PR.

bhardwajs:
> @bhardwajs
> With all due respect to you and Phil Withnall, this is nuts. 2.67 is an unstable series and it will be at the very best months before 2.68 makes it into cutting-edge distributions like Arch Linux and years for LTS ones like Ubuntu and Debian stable. Deprecating in one dot release of an unstable series and removing in the following one is going to break a ton of code and piss off just about everyone. It's stupid. However, since it's presented as a security vulnerability (the amount of memory being copied has to be enough to overflow a 32-bit uint, not exactly common and certainly not happening in GnuCash) it's possible that distros will try to make security patches. They'll probably screw it up.
> 
> Patching GLib in the test containers isn't practical and besides ignores the problem that we can't do that in the real world.
> 
> All of that aside, there are two ways to make this an acceptable PR. The easier one is to add a configuration check for `g_memdup2` being available and #ifdef accordingly. The other, arguably better, one is to replace the 4 `g_memdup` instances with `memcpy` and whatever boilerplate is needed to make that work.
> 
> The best, of course, is to rewrite the modules in C++ so that they don't use either `g_memdup` or `memcpy`, but that's a lot to ask for in a PR.

Agree. I wouldn't have send this PR if the change wasn't billed as security vulnerability.

Let me go with replacing the `g_memdup` with `memcpy` and send a new iteration targeting later in the week.

While I agree that the best course would be to rewrite the module in C++, unfortunately, I wouldn't be able to commit the time. Sorry!

bhardwajs:
@jralls - I sent #933 which can be cherry-picked against master as well. Thanks!

PR #934, opened 2021-03-10
I18N: Improve "Search from "

fellen:


fellen:
Currently one of the most annoying strings in [weblate checks](https://hosted.weblate.org/translate/gnucash/gnucash/en/?q=label:%22Source%20needs%20review%22)
@Bob-IT I do not fully understand this dialog. A few thoughts:
1. Is the const…; call; free… frame really required?
2. Should we add mnemonics?

gjanssens:
> Currently one of the most annoying strings in [weblate checks](https://hosted.weblate.org/translate/gnucash/gnucash/en/?q=label:%22Source%20needs%20review%22)
> @Bob-IT I do not fully understand this dialog. A few thoughts:
> 
>     1. Is the const…; call; free… frame really required?
You can omit the const declaration, but you can't omit g_free after running g_strdup_printf. Without it you will leak the memory allocated by the g_strdup_printf. Note the g_free call doesn't act on sub_label_start,  but on sub_label.

So if the const declaration really bothers you, you could add the translatable string directly into the g_strdup_printf  call and drop the const sub_label_start line:
`sub_label = g_strdup_printf (_("Search sub-accounts of %s"), sub_full_name);`

And for some bikeshedding :wink: : the new string is much longer and for me takes more time to interpret when put in its full context.

The radios have two options:
1. Search from root
2. Search from <account name>

These two strings clearly are two alternative options - both  allow to "search from"

That would now change to 
1. Search from  root
2. Search sub-accounts of <account name>

This formulation takes more user interpretation to estimate what alternatives they offer.

I would propose to rework the  radios to have a group box with title. Something like "Search scope". Then the radios itself could be condensed to
1. All accounts
2. Sub-accounts of <account-name>

Note I did change "Root" to "All accounts".  "Root" is a technical term referring to our hierarchical account tree concept. But that's not something clearly visible to users, so for them that's not a good concept.

Having said all that, I'd like to step back a bit and question the usefulness of this dialog altogether. The account tree itself can be searched directly just by typing parts of the account name you're looking for. What's the real benefit of this dialog and is that benefit worth enough to keep it around ?

Bob-IT:
@gjanssens
I was going to point out the need for a g_free for gnc_account_get_full_name but you were quicker than me :smile: 
Also I was going to suggest some other changes and I like your group box suggestion.

The main reason I added this was that you can only use the built in search for accounts that are visible. Do a search for an account in a collapsed branch or if you hide an account and use the filter to not show hidden accounts you will not find that either.

So this dialog allows you to find any account and temporarily over ride the hidden option to make it visible.

Bob-IT:
@fellen 
Try these changes in this patch file, it makes some changes for item 1, item 2, was mostly there but have added some for the radio buttons and checkbox, the strange thing is that the search button was already configured with '_Search' but does not work, it does not like the '_S' for some reason and can not see why so changed it to 'Sea_rch' but you may have a better one.
item 3, have added this and item 4, fix the window placement.

[0001-mychanges.txt](https://github.com/Gnucash/gnucash/files/6115527/0001-mychanges.txt)

gjanssens:
@Bob-IT Right, I see what you mean. The built-in search functionality of the GtkTreeView will only scan visible lines.

I find the GtkTreeView api has functions to tweak search behaviour. You can set a custom [search function](https://developer.gnome.org/gtk3/stable/GtkTreeView.html#gtk-tree-view-set-search-equal-func) for example. Did you ever experiment with those ? There is for example a 

And can I do another more radical design proposal (which you can follow or ignore as you see fit) ?
The benefit of the  tree view's built in search is that it's non-blocking - it's interactive and on the same screen as the data you're searching. UX wise that's tremendously comfortable. The current state of your dialog is missing both qualities: first it is in a separate window and in that window the user has to first type some text and then click search (well, you could select directly from the list as well, but that's not really searching).

So in the design I imagine there there's a search bar that integrates directly on the Account hierarchy page. It would have the following properties/qualities:
1.  It's not visible by default, but made visible by selecting the "Find accounts" menu
2. It will have a close button to close it again
3. It has a text field in which users can type part of the account name to search. This can possibly be made broader to search for all visible text columns, like name, code, description,... Not user configurable though. We decide whether we want it to search in all visible text columns or not. I'm tempted to enable this. It matches the way the built-in GtkTreeView search works.
4. Typing should be interactive. That means each character typed should update some result list. As we already have a full account list on the page, I'd suggest to use that as result list. That means as soon as some characters are typed, reduce the list of displayed accounts to those that match the text typed.
5. Collapsed matches should be made visible automatically. That is if a collapsed subaccount matches the typed text, the subtree it's part of should be expanded. I think in practise this means the whole tree should be expanded as soon as we start filtering based on the entered text.
6. All parents of a matched subaccount should be in the search list together with the matched subaccount. This is because there may be more  than one subaccount with the same name in different parts of the tree. Adding the parents helps the user select the proper one. Possibly the parent accounts of which the name doesn't match the search text could be presented differently from accounts that do have names matching the search text. Possible ways to do so include greying out text of non-matching accounts, or making text for matching accounts bold, or only make the matched text bold. 
7. There can be a checkbox to enable searching for hidden accounts. By default it should be unticked so that only accounts that are not hidden are searched. Note that with "hidden" I mean accounts the user has explicitly flagged as hidden, not accounts that happen to be collapsed.
8. After the search dialog is closed the account hierarchy returns to the state it was in before search started. So the filter state and the expansion states should be kept. I guess that can be  done by saving the filter model before starting a search.

Note this design doesn't have the radio buttons that started this PR at all. I don't think they would be needed.

I believe that design solves the problem you originally intended to solve while also avoiding  a blocking interface. It's a big design and totally outside of the scope of this PR. If you are interested to implement this at some point in the future (possibly after more discussion), we'd best move this proposal to an enhancement request. If you are not interested, we'll just forget about it.

Bob-IT:
@gjanssens I can not remember regarding tweaking function but have been thinking about 'GtkSearchEntry' and friends. I came across it when I used in it in help dialog experiment #903 and subsequently was thinking about using it in Customer report and maybe other places like...
search
the advanced button takes you to the normal find dialog and also how it could be used here. Your suggestion is an interesting idea and worth preserving so file an enhancement bug and add me so I do not forget about it.

gjanssens:
I have created  the enhancement request: https://bugs.gnucash.org/show_bug.cgi?id=798141

PR #935, opened 2021-03-12
Increase the maximum threshold for date mismatch

Ekleog:
Some french banks (that issue credit cards) have OFX files that count
the transaction as dated of the 5 of the month that follows the actual
expense. This means that the date offset between the invoice (that is at
the date of the expense) and the OFX file can be up to around 30-35
days.

So without this change, “usual” scores for such transactions are:
  +3 for cent-accurate amount
  -5 for date that mismatches by more than the max threshold
  +1 when lucky for matching the description or memo
thus leading to a score of -1, which can literally never match, as the
setting for reducing the minimal score for matches can only go down to
zero. And thus OFX matching would be impossible with the window, without
duplicating the transactions and then manually doing the matching
manually.

The only alternative to making this change that I could see would be to
parse the actual date, which the bank OFX includes in the description
field, and to compare with this date. It would probably be a more
precise solution, but would I think be pretty awfully bank-specific,
which in turn makes it a probably bad idea.

-----

Note: I'm not sure whether to do this PR against maint or master… I'll probably be applying the patch to my local gnucash instance anyway, so I'm PRing against maint, but I'd be happy to retarget to master if you think that's the place such a change should happen!

Anyway, thank you for this awesome program, I started using it only recently but it's already being great :)

### DIFF
--- a/gnucash/import-export/import-backend.c
+++ b/gnucash/import-export/import-backend.c
@@ -58,8 +58,14 @@ static QofLogModule log_module = GNC_MOD_IMPORT;
  *   Constants, should ideally be defined a user preference dialog  *
 \********************************************************************/
 
-static const int MATCH_DATE_THRESHOLD = 4; /*within 4 days*/
-static const int MATCH_DATE_NOT_THRESHOLD = 14;
+/* If the date matches within 4 days, it's probably the same transaction */
+static const int MATCH_DATE_THRESHOLD = 4;
+
+/* If the date is apart of more than a month, it's very likely not the same
+ * transaction. Some credit card companies delay things by a month, so having
+ * less than a month here would cause issues with matching transactions for
+ * these users */
+static const int MATCH_DATE_NOT_THRESHOLD = 40;
 
 /********************************************************************\
  *   Forward declared prototypes                                    *


Ekleog:
(the CI failure looks… very unrelated to this PR to me)

jralls:
The CI failure is unrelated.

Unfortunately opening the not-match threshold to 40 days will cause lots of false matches to everyone not using one of those French banks. You could make it an option, but keep in mind that even users of one of those banks that post-dates all of the transactions to the next month might also have a normal card that doesn't and so would want to configure the imports differently.
Parsing dates in memo fields is rather perilous without the user being able to specify the date format their bank uses.

Ekleog:
I'm with you that this increases the risk of false positive.

Maybe some combination of the below could be a reasonable compromise:
1. Make perfect-amount-match give more, eg. +5 instead of +3 (in practice on like 90% of my lines there's only one perfect amount match and it's the right one, though I guess other people may have other  constraints, eg. professionals will probably have more identical-in-amount transactions)
2. Make a date mismatch of 14-40 days give a lower malus, eg. -2 instead of the current -5 (and keep -5 for 40+ days mismatches)
3. Allow the user to configure a negative minimal-value-to-display-match (overall, AFAIU, a transaction that gets no displayed match means that it'll necessarily be duplicated and then have to be manually matched with the correct one, thus leading to risky manual edition and deletion)

What do you think?

fellen:
Have OFX files no date? IIRC all other formats have at least one date columns.
Perhaps we can we restrict this extended treshold to file type OFX and account type credit card?

christopherlam:
TBH this use case is obscure enough that I would disagree having a country-specific or account-type-specific increased days_threshold. A false-positive match is worse than no-match because the latter is much more visible than a false-positive match. I'd vote for a global preference option, or gsettings option.

Ekleog:
@fellen
> Have OFX files no date? IIRC all other formats have at least one date columns.

They do, the issue is that for credit cards the date logged there is the date at which the amount was debited from the account, not the date at which the transaction was made (which 100% makes sense but makes matches fail) — the date at which the transaction was made ends up only in the description, probably because OFX files have no way to have two dates (one “transaction date” and one “cleared date”)

@christopherlam 
> A false-positive match is worse than no-match because the latter is much more visible than a false-positive match.

Uhh I'm not sure I understand. To clarify: without this change all the transactions I see are orange and have zero matches in the “potential matches” list, thus forcing the user to duplicate the transaction and then manually fiddle with history to fix everything up (basically making using the import strictly worse in number of operations made than not using it at all as soon as at least 1/3 transactions are credit card). With this change, at least in my use case I see all the transactions as red transactions, for which I can then manually select “R” and it gives me a list of possible transactions from which I can select the proper one.

My issue with the “no false-positive match” stance (for which I'd 100% agree if it said “no false-positive green match”) is that if the correct transaction is not even in the potential match list, it's a huge pain to fixup.

> I'd vote for a global preference option, or gsettings option.

100% agree with you (actually there's a comment just above the global variables I change in this PR that states these variables should be configurable), but changing the GUI is beyond my skill with C GTK :/

christopherlam:
FWIW the OFX spec isn't always followed by the banks themselves (see https://github.com/libofx/libofx/issues/32) and the preference for GnuCash is not necessarily to accept all kinds of broken files. Instead of parsing dates in memo fields, it may be simply much easier to use CSV downloads instead.

jralls:
@ekleog, I ran across [bug 646340](https://bugs.gnucash.org/show_bug.cgi?id=636340) the other day while cleaning up date-related bugs. Does your bank include `<DTUSER>` in transactions?

Ekleog:
@jralls Looks like no… here is an exerpt from an OFX file, censored, that corresponds to a credit card payment made on 2021-02-23 (so the date present in gnucash) that was withdrawn from my account on 2021-03-04 (so the date auto-detected by gnucash):
```xml
<STMTTRN>
<TRNTYPE>DEBIT
<DTPOSTED>20210304
<TRNAMT>-[censored]
<FITID>[looks probably private? censored, let me know if you need it]
<NAME>CB [censored]         FACT 230221
<MEMO>CB [censored]         FACT 230221
</STMTTRN>
```

jralls:
Pity. Thanks for checking.

Bob-IT:
@jralls 
@gjanssens 
Is there any mileage in adding these two values to the preferences as noted in the source file, the default value can be made the same as existing, I think it would not take much to do?

jralls:
@Bob-IT If you mean the two-level matching approach @Ekleog proposed 12 March, I think it's worth at least testing if you can implement it easily.

fellen:
> I would disagree having a country-specific or account-type-specific increased days_threshold.

If required, it should be set in the online banking slots of the respective account like
```
    <slot>
      <slot:key>ofx</slot:key>
      <slot:value type="frame">
         <slot>
           <slot:key>account-id</slot:key>
           <slot:value type="string">1234567890</slot:value>
         </slot>
          :
        </slot:value>
    </slot>

```

Bob-IT:
I have created PR #963 which adds the two constant defines to the preferences dialog so they can be changed by the user.
All works through testing as far as I can tell but maybe the text could be better.
@Ekleog If you have time you may want to test and give feedback.

jralls:
@Bob-IT Would it be at all difficult to make it a per-account preference as @fellen suggests instead of an app-wide one?

Bob-IT:
@jralls
Hmm, the only place I can see that it might work that gives access to change these settings would be in the 'Generic Import Matcher'. If one had a button that raised a dialog next to the help button that was enabled by selecting only the same account rows, as potentially there could be many accounts in the import file. This dialog would then have the five threshold options like the preference import ones allowing you to change a setting. On pressing the dialog OK button, the changed values would be saved as account KVP data and do a reload of the import file to use the new settings.

What I am not sure of is with a multi account file, does the import matcher allow individual account settings OR is it one set of settings for the import file, if one set how difficult it would be to change.

The existing KVP layout would need to be tested and migrated to new layout along with changes to the 'Import Map Editor' that would still just show the online_id mapping but would delete all setting if removed.

I assume this would also apply to CSV imports?

Probably not a simple change and is it worth doing.

christopherlam:
> @jralls
> Hmm, the only place I can see that it might work that gives access to change these settings would be in the 'Generic Import Matcher'. If one had a button that raised a dialog next to the help button that was enabled by selecting only the same account rows, as potentially there could be many accounts in the import file.

Or the Account properties dialog? It would get rather cluttered though.

gjanssens:
> > @jralls
> > Hmm, the only place I can see that it might work that gives access to change these settings would be in the 'Generic Import Matcher'. If one had a button that raised a dialog next to the help button that was enabled by selecting only the same account rows, as potentially there could be many accounts in the import file.
> 
> Or the Account properties dialog? It would get rather cluttered though.

You can always add an extra tab  in there specifically for import settings to keep the main tab clean.

gjanssens:
> The Edit Account dialog is what I was thinking of. I don't much like adding a tab to it though. Maybe if the current stack of checkboxes were split into two columns there'd be room for two spinners underneath for the thresholds without making the dialog any longer.
> 
The dialog is already tabbed, though it mostly shows only one tab, the "General" tab. When called from the Account Hierarchy Assistant a second tab for opening balances is visible. I also think the the amount of information to process on a single page on this dialog is getting much. Ofx specific import parameters don't concern everybody so they would just clutter the general account properties. The same goes for the Tax-related flag really, but making a separate tab for just a single, read-only preference is a bit overkill.

> As a slightly related UI point the Tax-Related indicator shouldn't look like it's an insensitive editable item. A label whose entry is determined by the tax status (e.g. `Tax Related: Yes` | `Tax Related: No`) would be better, but maybe it isn't even necessary. The best way to see which accounts are tax related is to enable the `Tax Info` column on the accounts page.

I agree.

fellen:
> > As a slightly related UI point the Tax-Related indicator shouldn't look like it's an insensitive editable item. A label whose entry is determined by the tax status (e.g. `Tax Related: Yes` | `Tax Related: No`) would be better, but maybe it isn't even necessary. The best way to see which accounts are tax related is to enable the `Tax Info` column on the accounts page.
> 
> I agree.

From the workflow perspective it would be better to have a button, which opens Tax Report Options with the current account preselected.

jralls:
> From the workflow perspective it would be better to have a button, which opens Tax Report Options with the current account preselected.

That shouldn't be too hard, the logic is already in place for selecting Edit>Tax Report Options from a register tab.

Ekleog:
@Bob-IT I'll be more than happy to test things out, thank you, it looks awesome! Just to know, as there were discussions of maybe making the preference account-local since your message, do you plan to go back on it and so I should wait until that for testing?

Bob-IT:
@Ekleog I would hang fire on testing, I am pondering on what has been said, thank you.

jralls:
@Bob-IT How's the pondering doing?

Bob-IT:
Sorry, another one that slipped my mind, will need to reread this and action, this may be can be combined with a bug request that I came across about changing the order of widgets. Will try and find that so all the suggested changes can be looked at.

Ekleog:
Seems like in latest gnucash there's in parameters>import a "likely match day threshold" and "unlikely match day threshold", so I think this has been implemented in a much better way than this PR.

So I'm closing this, thank you for actually doing this but better! :)

PR #939, opened 2021-03-14
[dialog-invoice] invoice editor reuses invoice report tab

christopherlam:
Adds linkage from editor to report similar to #930.

Similar to 9f4ba17d8 -- from invoice editor, click Print Invoice to create invoice report. This saves link to invoice, and reuses the report tab if Print Invoice is clicked again.

### DIFF
--- a/gnucash/gnome/dialog-invoice.c
+++ b/gnucash/gnome/dialog-invoice.c
@@ -70,6 +70,7 @@
 
 #include "gnc-plugin-business.h"
 #include "gnc-plugin-page-invoice.h"
+#include "gnc-plugin-page-report.h"
 #include "gnc-main-window.h"
 #include "gnc-state.h"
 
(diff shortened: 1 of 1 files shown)


christopherlam:
From my testing this works well...

From invoice editor -> click "Print Invoice" -> creates invoice report tab, which is reused.

However (from the Invoice report, enable Invoice Notes display option first):
- From invoice editor, modify an invoice entry -> click Print Invoice -> renders amended report with new entry details.
- From invoice editor, modify the invoice notes -> click Print Invoices -> switches to report tab but doesn't pick up new notes.

I suspect the notes GtkTextBuffer is not immediately saved into Invoice notes when Print Invoice is activated.

Bob-IT:
@christopherlam Of the top of my head, I would think that we could use a key sequence like used for the notes field on the accounts page. Will have a look later.

christopherlam:
Merged in. @Bob-IT if you have idea how to update invoice when the `Print Invoice` is clicked, please amend as desired.

PR #944, opened 2021-03-18
Restructure early locale initialization

gjanssens:
- Move Macos specific bits to its own source file. As this is objective-c use .mm extension
- Use common function signature for Macos and Windows init function and include via common header file

The only functional change is that the platform  specific functions now return a `char  *` where before the Windows function was passed a writable parameter and the MacOS function didn't have any means to return a value. It should otherwise behave exactly as before the change. More changes may follow based on discussion on the mailing list.

@jralls can you provide feedback on whether or not this will build on MacOS ? I have found the typical extension for objective-c is .m or .mm, but I wasn't sure which one would fit here. I have also made a few minor tweaks (adding a common header/set a return value for the main function) which I would like to know if they are correct.

When I get my Windows build working again, I can verify the Windows part myself.

### DIFF
--- a/gnucash/CMakeLists.txt
+++ b/gnucash/CMakeLists.txt
@@ -34,6 +34,7 @@ endif()
 set(gnucash_noinst_HEADERS
     gnucash-commands.hpp
     gnucash-core-app.hpp
+    gnucash-locale-platform.h
 )
 
 set (gnucash_SOURCES
@@ -45,7 +46,9 @@ set (gnucash_SOURCES
   )
 
 if (MINGW)
-  list(APPEND gnucash_SOURCES "gnucash-windows-locale.c")
+  list(APPEND gnucash_SOURCES "gnucash-locale-windows.c")
+elseif(MAC_INTEGRATION)
+    list(APPEND gnucash_SOURCES "gnucash-locale-macos.mm")
 endif()
 
 add_executable (gnucash
@@ -71,7 +74,10 @@ set(gnucash_cli_SOURCES
     )
 
 if (MINGW)
-  list(APPEND gnucash_cli_SOURCES "gnucash-windows-locale.c")
+  list(APPEND gnucash_cli_SOURCES "gnucash-locale-locale.c")
+endif()
+if (MAC_INTEGRATION)
+    list(APPEND gnucash_cli_SOURCES "gnucash-locale-macos.mm")
 endif()
 
 add_executable (gnucash-cli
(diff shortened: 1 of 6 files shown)


jralls:
@gjanssens I see you discovered that CI answers your macOS question. ;-)

I've no preference about .m vs. .mm.

gjanssens:
> @gjanssens I see you discovered that CI answers your macOS question. ;-)

:smiley: It does partly. When the build succeeds, I'll have to ask you to proof-run it to verify it actually still works. I'll ping you in due time...

gjanssens:
Congrats on the macos ci implementation btw. It is very helpful for making changes to code on that platform while not having an Apple system to work on.

gjanssens:
Ok, macos build & test succeeded. I have squashed all fixes. @jralls it's now ready to be test-run on your  system...

jralls:
Seems to work fine with System Preferences>Language & Region changed to German in Germany.

gnucash/CMakeLists.txt
```diff
@@ -45,7 +46,10 @@ set (gnucash_SOURCES
   )
 
 if (MINGW)
-  list(APPEND gnucash_SOURCES "gnucash-windows-locale.c")
+  list(APPEND gnucash_SOURCES "gnucash-locale-windows.c")
+endif()
```
bhardwajs, line 50: Should this be
```
if()
elseif()
endif()
```
Is there a scenario where both `MINGW` and `MAC_INTEGRATION` evaluate to true?

gnucash/CMakeLists.txt
```diff
@@ -71,7 +75,10 @@ set(gnucash_cli_SOURCES
     )
 
 if (MINGW)
-  list(APPEND gnucash_cli_SOURCES "gnucash-windows-locale.c")
+  list(APPEND gnucash_cli_SOURCES "gnucash-locale-locale.c")
+endif()
+if (MAC_INTEGRATION)
```
bhardwajs, line 80: Same as above.

gnucash/gnucash-locale-windows.c
```diff
@@ -24,17 +25,15 @@
...
+#include "gnucash-locale-platform.h"
 
 /* If one of the Unix locale variables LC_ALL, LC_MESSAGES, or LANG is
  * set in the environment check to see if it's a valid locale and if
  * it is set both the Windows and POSIX locales to that. If not
  * retrieve the Windows locale and set POSIX to match.
  */
-void
-set_win32_thread_locale(char **sys_locale)
+char *void
```
bhardwajs, line 35: Is that a typo? Function returning `char* void`?

gnucash/CMakeLists.txt
gjanssens: Agreed using elseif() expresses the intent more clearly. I have pushed with your suggestion.

gnucash/gnucash-locale-windows.c
gjanssens: Done in a followup-commit.

gnucash/CMakeLists.txt
gjanssens: Yep.

PR #951, opened 2021-03-27
remove debug information for time conversion

c-holtermann:
After a recent commit (a11065b823f14fbc32bb1fc132304e36f486c8b0) the python bindings have become exceedingly chatty. I rather remove this debug info than trying to make it optional as it seems a bit too complicated to me to achieve this. I suppose this has been used only for the work on @jralls commit and removing it is ok, John?


christopherlam:
Oops. I'm not John but I'm sure this is very reasonable.

jralls:
Yes, I left that in by mistake. Thanks.

PR #958, opened 2021-04-01
Fix to_string_with_decimal_point_placed

koldavi:
>>> nominator='123456'
A digit is missing in the string returned from to_string_with_decimal_point_placed. The -1 error.
>>> point_place=2
>>> nominator[:-point_place - 1], nominator[-point_place:]
('123', '56')


koldavi:
Just a small bug fix in Python binding example.

jralls:
Thanks!

PR #963, opened 2021-04-05
Add two new preferences for import for PR #935

Bob-IT:
In the import-backend.c source file two constants were defined with a
comment about they should ideally be added to the user preference
dialog. This commit does that, the two preferences are
match-date-threshold which is the number of days a transaction match is
likely and match-date-not-threshold which is the number of days a
transaction match is considered not likely.

### DIFF
--- a/gnucash/gnome-utils/dialog-preferences.c
+++ b/gnucash/gnome-utils/dialog-preferences.c
@@ -1360,6 +1360,8 @@ gnc_preferences_dialog_create (GtkWindow *parent)
     gnc_builder_add_from_file (builder, "dialog-preferences.glade", "auto_add_adj");
     gnc_builder_add_from_file (builder, "dialog-preferences.glade", "auto_clear_adj");
     gnc_builder_add_from_file (builder, "dialog-preferences.glade", "match_adj");
+    gnc_builder_add_from_file (builder, "dialog-preferences.glade", "likely_day_threshold");
+    gnc_builder_add_from_file (builder, "dialog-preferences.glade", "unlikely_day_threshold");
     gnc_builder_add_from_file (builder, "dialog-preferences.glade", "gnucash_preferences_dialog");
 
     dialog = GTK_WIDGET(gtk_builder_get_object (builder, "gnucash_preferences_dialog"));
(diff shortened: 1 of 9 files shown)


christopherlam:
This can be reused in `qif-merge-groups.scm` which currently matches with existing splits _one week_ before&after the imported splits date range.

christopherlam:
@jralls any objection? The following changes the QIF importer to accept up to user-supplied days difference when matching with previous. The default 14 is higher than the previous 7, which means it will match more existing transactions, thus unlikely to annoy anyone.

```patch
modified   gnucash/import-export/qif-imp/qif-merge-groups.scm
@@ -57,6 +57,9 @@
 ;;  and change #f to #t where duplication is found.
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 
+(define (max-days-difference)
+  (gnc-prefs-get-float "dialogs.import.generic" "match-date-not-threshold"))
+
 (define (gnc:account-tree-find-duplicates old-root new-root progress-dialog)
   (define old-accounts (gnc-account-get-descendants-sorted old-root))
   (define (progress v)
@@ -69,7 +72,7 @@
       ;; Get all the splits in the new tree, then iterate over them
       ;; trying to find matches in the old tree.  If there are
       ;; matches, push the splits' parent onto a list.
-      (let ((WeekSecs (* 60 60 24 7)))
+      (let ((WeekSecs (* 60 60 24 (max-days-difference))))
 
         (define new-splits
           (let ((q (qof-query-create-for-splits))
@@ -131,7 +134,7 @@
                         (string=?
                          (gnc-account-get-full-name (xaccSplitGetAccount old-split))
                          (gnc-account-get-full-name (xaccSplitGetAccount new-split)))
-                        ;; maximum 1 week date difference
+                        ;; maximum date difference according to user pref
                         (<= (abs (- (xaccTransGetDate (xaccSplitGetParent old-split))
                                     (xaccTransGetDate (xaccSplitGetParent new-split))))
                             WeekSecs)))
```

jralls:
No real objection but wouldn't it be better to replicate the two-level algo that the generic matcher uses? Or  maybe even better SWIG the generic matching functions and use them from the QIF importer?

christopherlam:
> No real objection but wouldn't it be better to replicate the two-level algo that the generic matcher uses? Or maybe even better SWIG the generic matching functions and use them from the QIF importer?

Gah. The qif importer is off limits. Too difficult!

PR #964, opened 2021-04-05
Bug 796761 - Control characters can be pasted in register fields

Bob-IT:
I was looking at Bug 798157 which maybe should be marked as a duplicate of Bug 796761

They are both partly dealing with control characters being pasted in to register fields so thought I would try and filter them out and came up with this. This basically gets the text from the clipboard when requested, filters it and sends it back to the clipboard before the gnc_item_edit does the paste thus leaving the GtkEditable to deal with the insertion. Control characters are stripped from the beginning and end of text with a space replacing them in the middle of text.

As far as I can tell it seems to do the job but as usual I may of overlooked some aspect.

### DIFF
--- a/gnucash/gnome-utils/gnc-amount-edit.c
+++ b/gnucash/gnome-utils/gnc-amount-edit.c
@@ -35,6 +35,7 @@
 #include "gnc-ui-util.h"
 #include "qof.h"
 #include "dialog-utils.h"
+#include "gnc-ui.h"
 
 #ifdef G_OS_WIN32
 # include <gdk/gdkwin32.h>
@@ -53,6 +54,8 @@ static void gnc_amount_edit_init (GNCAmountEdit *gae);
 static void gnc_amount_edit_class_init (GNCAmountEditClass *klass);
 static void gnc_amount_edit_changed (GtkEditable *gae,
                                      gpointer user_data);
+static void gnc_amount_edit_paste_clipboard (GNCAmountEdit *gae,
+                                             gpointer user_data);
 static gint gnc_amount_edit_key_press (GtkWidget   *widget,
                                        GdkEventKey *event);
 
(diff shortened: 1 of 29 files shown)


limitedAtonement:
Thanks for working on this bug!

> This basically gets the text from the clipboard when requested, filters it and sends it back to the clipboard

I would be irritated to find that my clipboard was modified by pasting to an application.

Bob-IT:
@limitedAtonement
I see you point, it was the simplest way, otherwise I would need to care about where the cursor was and what was already selected instead of letting the GtkEditable/GtkEntry just deal with it.
Will have another look.

Bob-IT:
> I would be irritated to find that my clipboard was modified by pasting to an application.

I have changed the way it is done, now done by getting the clipboard, filtering the text and then update the GtkEditable directly so clipboard is still intact.

jralls:
@Bob-IT I think this could also address [Bug 798144](https://bugs.gnucash.org/show_bug.cgi?id=798144).

jralls:
@Bob-IT 
It looks to me that this only fixes the control-characters problem and the paste would still fail because of the currency symbol and possibly the thousands sep and decimal mark if the pasted number doesn't match the user's localization.

[Bug 798144](https://bugs.gnucash.org/show_bug.cgi?id=798144) is another instance of a similar problem.

Bob-IT:
@jralls Yes, the filter_text function, should rename that, deals with stripping the control characters out only. As it is used for pasting all entry fields on the register, that's all it realy can do as you do not want monetary symbols stripped from fields like notes. I would of thought that would of maybe down to the expression parser, not sure I do not think I have looked at it before so need to have a look how that is used in the register.

As for number formats, I would of thought it should be down to the user to provide a number in the format currently being used. Again not something I have looked so will need to figure out.

As for Bug 798144, interesting, on a quick look the widget being used is a GNCAmountEdit which in this case uses the expression parser, gnc_exp_parser_parse, on 'OK'. A quick experiment with pasting £55.68 returns a numeric of '94846720793088/139875915098029' but also an indication of 'NO ERROR' while pasting 55.68 gives '1392/25'.

Will have a think.

jralls:
@Bob-IT I hadn't even thought of fiddling with the expression parser. You're absolutely right that that's the place to go for dealing with a lot of these number parsing issues.

Bob-IT:
I have think this completes the changes for the register for removing control characters and also pasting of an amount with a currency symbol. 
So for a GBP account, £1,000 will paste OK but $1,000 will not as before. For a share account, price and buy/sell will work with currency symbol of the parent account and number of shares will not. This was made possible with the last two commits.
If the expression parser should fail, a dialog will be shown with the position of the error. From testing it seems to work OK.

gnucash/register/register-core/pricecell.c
```diff
@@ -103,12 +117,18 @@ gnc_price_cell_modify_verify (BasicCell *_cell,
             !g_unichar_isalpha (uc) &&
             (decimal_point != uc) &&
             (thousands_sep != uc) &&
-            (g_utf8_strchr (toks, -1, uc) == NULL))
+            (g_utf8_strchr (tokens, -1, uc) == NULL))
+        {
+            g_free (tokens);
             return;
+        }
```
jralls, line 124: The above looks a lot like the code in formulacell.c. Can you extract a function to basiccell.c and call it in both places?

jralls (changes requested):

Bob-IT:
@jralls Thank you for looking, will have another look, especially the Unicode stuff.

gnucash/register/register-gnome/gnucash-item-edit.c
```diff
@@ -398,7 +399,55 @@ gnc_item_edit_copy_clipboard (GncItemEdit *item_edit)
 void
 gnc_item_edit_paste_clipboard (GncItemEdit *item_edit)
 {
-    gtk_editable_paste_clipboard(GTK_EDITABLE(item_edit->editor));
+    GtkClipboard *clipboard = gtk_widget_get_clipboard (GTK_WIDGET(item_edit->editor),
+                                                        GDK_SELECTION_CLIPBOARD);
+    gchar *text = gtk_clipboard_wait_for_text (clipboard);
+
+    if (text)
```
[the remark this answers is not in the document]
Bob-IT: Done.

gnucash/register/register-gnome/gnucash-item-edit.c
```diff
@@ -398,7 +399,55 @@ gnc_item_edit_copy_clipboard (GncItemEdit *item_edit)
...
+            gtk_editable_delete_text (GTK_EDITABLE(item_edit->editor), 0, -1);
+
+            gtk_editable_insert_text (GTK_EDITABLE(item_edit->editor),
+                                      changed_text, -1, &position);
+
+            position = start_pos + g_utf8_strlen (filtered_text, -1);
+
+            g_free (begin);
+            g_free (end);
+            g_free (changed_text);
```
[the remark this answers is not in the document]
Bob-IT: I am still not sure why I did all this, I know I had a problem to start with but any way changed.

libgnucash/app-utils/gnc-ui-util.c
```diff
@@ -2663,3 +2663,52 @@ gnc_ui_util_remove_registered_prefs (void)
...
+        return g_strdup (text);
+
+    filtered = g_string_new ("");
+
+    while (*text)
+    {
+        gunichar uc = g_utf8_get_char (text);
+
+        // check for starting with control characters
+        if (g_unichar_iscntrl (uc) && !text_found)
```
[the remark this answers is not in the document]
Bob-IT: Made changes.

gnucash/register/register-core/pricecell.c
Bob-IT: Factored out common code.

Bob-IT:
I have closed the other PR and added all changes from there to here. Now using GError for parsing error with additional code changes. I looked at factoring out the Gtk paste code to app-utils/gnc-ui-util.* but there are some differences and those files do not use Gtk functions so have not done that.

gnucash/gnome-utils/gnc-amount-edit.c
```diff
@@ -51,116 +45,216 @@
...
+
+    gae->entry = GTK_ENTRY(gtk_entry_new());
     gae->need_to_parse = FALSE;
     gae->amount = gnc_numeric_zero ();
     gae->print_info = gnc_default_print_info (FALSE);
     gae->fraction = 0;
     gae->evaluate_on_enter = FALSE;
+    gae->validate_on_change = FALSE;
+    gae->block_changed = FALSE;
+    gae->show_warning_symbol = TRUE;
```
jralls, line 160: `   gae->disposed = FALSE;`

jralls:
Looks good aside from forgetting to init `gae->disposed`.

gnucash/gnome-utils/gnc-amount-edit.c
jralls: Never mind, did it in the merge commit..

PR #965, opened 2021-04-06
[dialog-sx-editor]: Enable SX toggles if enabled.

christopherlam:
This branch aims to enable GtkWidget sensitivities if and only if the SX is enabled.

Upcoming:

- [ ] enable SX right-click to pop up shortcuts to Edit / Delete
- [ ] ~ESC to close SX dialog~

### DIFF
--- a/gnucash/gnome/dialog-sx-editor.c
+++ b/gnucash/gnome/dialog-sx-editor.c
@@ -128,6 +128,7 @@ struct _GncSxEditorDialog
     GtkToggleButton *optEndDate;
     GtkToggleButton *optEndNone;
     GtkToggleButton *optEndCount;
+    EndType end_type;
     GtkEntry *endCountSpin;
     GtkEntry *endRemainSpin;
     GNCDateEdit *endDateEntry;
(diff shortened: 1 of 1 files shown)


gnucash/gnome/dialog-sx-editor.c
```diff
@@ -1012,43 +1013,55 @@ gnc_sxed_save_sx( GncSxEditorDialog *sxed )
     gnc_sx_commit_edit( sxed->sx );
 }
 
+static void gnc_sxed_update_sensitivity (GncSxEditorDialog *sxed)
```
Bob-IT, line 1016: I would put gnc_sxed_update_sensitivity (GncSxEditorDialog *sxed) on a new line

Bob-IT (commented):
Only one change I could see. How many changes do you think you will be doing, the only reason for asking is I think the source file could do with a lot of white space realignment.

gnucash/gnome/dialog-sx-editor.c
christopherlam: done

christopherlam:
> Only one change I could see. How many changes do you think you will be doing, the only reason for asking is I think the source file could do with a lot of white space realignment.

Will update the top comment in this PR with todo list. I consider whitespace changes as a content-free modification that can wait till all behavioural changes are done :)

jralls:
@christopherlam you do realize that that commit message makes no sense at all, right? A more correct description of this might be `Extract function gnc_sxed_update_sensitivity`.

*But* since it's a static function it shouldn't really have the `gnc_sxed_` prefix. Reserving prefixes for public functions makes it easier to know which functions are public and which are not when reading the code.

> I consider whitespace changes as a content-free modification that can wait till all behavioural changes are done :)

Content-free indeed. If you're planning a series of changes consider doing the whitespace first to ensure that diffs on the "real" changes aren't made less clear by whitespace fixups made as part of editing.

christopherlam:
Unfortunately pressing ESC doesn't quite bring up the "Are you sure" dialog. Herewith the console output:

```
* 21:43:09 ERROR <gnc.gui> main_window_update_page_set_read_only_icon: assertion 'page && page->window && GNC_IS_MAIN_WINDOW(page->window)' failed
* 21:43:09 ERROR <gnc.gui> main_window_update_page_set_read_only_icon: assertion 'page && page->window && GNC_IS_MAIN_WINDOW(page->window)' failed
* 21:43:09 ERROR <gnc.gui> main_window_update_page_set_read_only_icon: assertion 'page && page->window && GNC_IS_MAIN_WINDOW(page->window)' failed
* 21:43:11 ERROR <Gtk> gtk_widget_event: assertion 'WIDGET_REALIZED_FOR_EVENT (widget, event)' failed
* ```

Bob-IT:
@christopherlam 
The first three lines you can ignore or I was thinking of changing gnc_main_window.c line 2479 to 
if (page && page->window && !GNC_IS_MAIN_WINDOW(page->window))
    return;
I think that is correct but have not tested.

The second point is the SX editor is a GtkDialog so should already handle the escape key, only GtkWindows need the extra bits.
I think this is another grab focus issue, with it open the escape key does nothing but change tabs with mouse escape key works.
Have not looked at code so do not know where the first tab should be being focused.

christopherlam:
Thanks Bob have removed the ESC handling you're right that if focus is held by a checkbox or entry field then ESC is handled well. Next one will be right click.

christopherlam:
May be better keep PRs small. This branch is complete?

christopherlam:
Any one knows if the "create in advance" and "remind in advance" sensitivity should depend on "create automatically" ?
image

Bob-IT:
> Any one knows if the "create in advance" and "remind in advance" sensitivity should depend on "create automatically" ?

I do not think so, for things like that I have a version of 2.6.21 that I compare with just in case it was converted wrong in move to gtk3. What is different is the 'Notify me when created' is not indented.

Bob-IT:
> Thanks Bob have removed the ESC handling you're right that if focus is held by a checkbox or entry field then ESC is handled well. Next one will be right click.

Forgot to mention I think I know why the ESC does not work, need to do some more testing.

christopherlam:
> > Any one knows if the "create in advance" and "remind in advance" sensitivity should depend on "create automatically" ?
>
> I do not think so, for things like that I have a version of 2.6.21 that I compare with just in case it was converted wrong in move to gtk3. What is different is the 'Notify me when created' is not indented.

Thanks. I was mainly asking to see whether logically it should be dependent. But I agree it's probably fine as it stands (in this branch).

PR #978, opened 2021-04-18
average-cost-base report

christopherlam:
offshoot from #818 -- average-cost-base calculator report commits only.

### DIFF
--- a/bindings/guile/test/test-engine-extras.scm
+++ b/bindings/guile/test/test-engine-extras.scm
@@ -182,6 +182,7 @@
           num              ; string: num field   (def = null)
           notes            ; string: notes       (def = null)
           memo             ; string: memo        (def = null)
+          currency         ; commodity           (def = commodity of 1st split)
           )
   (env-create-multisplit-transaction
    env
@@ -193,6 +194,7 @@
    #:reconcile reconcile
    #:num num
    #:memo memo
+   #:currency currency
    #:notes notes))
 
 (define* (env-transfer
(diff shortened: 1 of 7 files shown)


bindings/guile/test/test-engine-extras.scm
```diff
@@ -239,42 +241,46 @@
...
+                               (vector-ref split 3)))
                   (newsplit (xaccMallocSplit book)))
               (xaccSplitSetParent newsplit txn)
               (xaccSplitSetAccount newsplit acc)
               (xaccSplitSetValue newsplit val)
               (xaccSplitSetAmount newsplit amt)
-              (if num (gnc-set-num-action txn newsplit num num))
-              (if memo (xaccSplitSetMemo newsplit memo))
+              (if memo   (xaccSplitSetMemo newsplit memo))
+              (if action (xaccSplitSetAction newsplit action))
```
jralls, line 270: That's not erroneous. There's a book option to use the split Action field for numbers so that you can have different numbers in different accounts on the same transaction. `gnc-set-num-action` reads the preference and puts the number in the appropriate field.

gnucash/report/reports/standard/test/test-register.scm
```diff
@@ -81,7 +81,7 @@
     (set-option options "__reg" "journal" #t)
     (let ((sxml (options->sxml options "journal")))
       (test-equal "table has 339 cells"
-        339
+        329
```
jralls, line 84: You should change the test descriptions to match the changed numbers.

bindings/guile/test/test-engine-extras.scm
christopherlam: I know and not sure how to engineer this test utility function. I find the split-action book property and API extremely inelegant and wish to enable it completely as a book upgrade*. The acb report uses the split action property to denote fees.

* TBD in devel I guess.

jralls: I suppose you mean "remove it" rather than "enable it", it's already enabled. Expect wailing and gnashing of teeth, there are users who think it's a Good Thing.

As for using it to designate a fee split I suppose it matters in what account. If it's Expenses:Investment:Commisisons and Fees then maybe it's unlikely that a user would want to put a number there.

christopherlam: I meant "enable and enforce it". ISTM if enabled then num becomes t-num and action becomes num, therefore more fields become available. Therefore they'll be no data loss. But whether to display num vs action in the register column would depend on the preference.

jralls: Sorry, you've lost me completely, but I haven't looked hard at the code in years. I thought that it hid the transaction object's num data in basic mode and pointed the  register column at the current account's split's action field; split view shows everything in its usual place. Do I mis-remember?

christopherlam:
An idea:

instead of trying to guess the stock-account's associated expense/proceeds/dividend/capgains accounts from Memo/Action heuristics, I'd retrieve them directly from the stock-account's kvp fields being suggested in #818 -- this means I'll create a PR for setting up the KVP fields soon, to be used by both the report and the dialog-stock-editor.

christopherlam:
This would mean the stock-account properties page should have another tab 'associated accounts' where the defaults can be set...

PR #979, opened 2021-04-20
Bug798148 - Accounts page Present (XXX) column uses future prices

Bob-IT:
I am not sure about this so would like a second opinion...
This is to do with when a security on the Chart of Accounts uses column 'Present(xxx)' and uses a price entered into the pricedb which is in the future to calculate the balance. I am not sure whether having prices in the future is OK, certainly one could do it by error.

It all starts from gnc-tree-model-account.c and the value returned from xaccAccountGetPresentBalanceInCurrency which used a recursive function that had no date parameter. Changing that to use a version with a date parameter lead to a function that did not use the date so that needed changing also.

If the future price was way off, then this may of been OK but the bug report was for a day or two and hence the report.

To fix this I added a function, gnc_pricedb_convert_balance_latest_before_price_t64 to gnc_pricedb that would get the last price before a date and changed associated pricedb functions so I could use existing code. What has confused me is that there are a couple of comments in gnc-pricedb.h about a couple of lookup functions that I have used  stating 'NOT USED, but see bug 743753'

If this progresses I need to add a couple of tests for this new function, probably squash the commits and there is also some inconsistency with using 'const Account' and 'const gnc_commodity' which I would try to align if worth doing.

### DIFF
--- a/libgnucash/engine/gnc-pricedb.c
+++ b/libgnucash/engine/gnc-pricedb.c
@@ -2024,8 +2024,8 @@ PriceList *
 gnc_pricedb_lookup_latest_any_currency(GNCPriceDB *db,
                                        const gnc_commodity *commodity)
 {
-    return gnc_pricedb_lookup_latest_before_any_currency_t64(db, commodity,
-                                                         gnc_time(NULL));
+    return gnc_pricedb_lookup_nearest_before_any_currency_t64(db, commodity,
+                                                              gnc_time(NULL));
 }
 
 PriceList *
@@ -2049,9 +2049,9 @@ gnc_pricedb_lookup_nearest_in_time_any_currency_t64(GNCPriceDB *db,
 }
 
 PriceList *
-gnc_pricedb_lookup_latest_before_any_currency_t64(GNCPriceDB *db,
-                                                  const gnc_commodity *commodity,
-                                                  time64 t)
+gnc_pricedb_lookup_nearest_before_any_currency_t64(GNCPriceDB *db,
+                                                   const gnc_commodity *commodity,
+                                                   time64 t)
 {
     GList *prices = NULL, *result;
     UsesCommodity helper = {&prices, commodity, t};
@@ -2405,10 +2405,10 @@ gnc_pricedb_lookup_nearest_in_time64(GNCPriceDB *db,
 
 
 GNCPrice *
-gnc_pricedb_lookup_latest_before_t64 (GNCPriceDB *db,
-                                      gnc_commodity *c,
-                                      gnc_commodity *currency,
-                                      time64 t)
+gnc_pricedb_lookup_nearest_before_t64 (GNCPriceDB *db,
+                                       gnc_commodity *c,
+                                       gnc_commodity *currency,
+                                       time64 t)
 {
     GList *price_list;
     GNCPrice *current_price = NULL;
@@ -2533,9 +2533,9 @@ indirect_price_conversion (GNCPriceDB *db, const gnc_commodity *from,
     }
     else if (before_date)
     {
-        from_prices = gnc_pricedb_lookup_latest_before_any_currency_t64 (db, from, t);
+        from_prices = gnc_pricedb_lookup_nearest_before_any_currency_t64 (db, from, t);
         if (from_prices)
-            to_prices = gnc_pricedb_lookup_latest_before_any_currency_t64 (db, to, t);
+            to_prices = gnc_pricedb_lookup_nearest_before_any_currency_t64 (db, to, t);
     }
     else
     {
@@ -2566,7 +2566,7 @@ direct_price_conversion (GNCPriceDB *db, const gnc_commodity *from,
     if (t == INT64_MAX)
         price = gnc_pricedb_lookup_latest(db, from, to);
     else if (before_date)
-        price = gnc_pricedb_lookup_latest_before_t64(db, (gnc_commodity*)from, (gnc_commodity*)to, t);
+        price = gnc_pricedb_lookup_nearest_before_t64(db, (gnc_commodity*)from, (gnc_commodity*)to, t);
     else
         price = gnc_pricedb_lookup_nearest_in_time64(db, from, to, t);
 
--- a/libgnucash/engine/gnc-pricedb.c
+++ b/libgnucash/engine/gnc-pricedb.c
@@ -2406,8 +2406,8 @@ gnc_pricedb_lookup_nearest_in_time64(GNCPriceDB *db,
 
 GNCPrice *
 gnc_pricedb_lookup_nearest_before_t64 (GNCPriceDB *db,
-                                       gnc_commodity *c,
-                                       gnc_commodity *currency,
+                                       const gnc_commodity *c,
+                                       const gnc_commodity *currency,
                                        time64 t)
 {
     GList *price_list;
@@ -2566,7 +2566,7 @@ direct_price_conversion (GNCPriceDB *db, const gnc_commodity *from,
     if (t == INT64_MAX)
         price = gnc_pricedb_lookup_latest(db, from, to);
     else if (before_date)
-        price = gnc_pricedb_lookup_nearest_before_t64(db, (gnc_commodity*)from, (gnc_commodity*)to, t);
+        price = gnc_pricedb_lookup_nearest_before_t64(db, from, to, t);
     else
         price = gnc_pricedb_lookup_nearest_in_time64(db, from, to, t);
 
--- a/libgnucash/engine/gnc-pricedb.c
+++ b/libgnucash/engine/gnc-pricedb.c
@@ -2516,7 +2516,7 @@ convert_price (const gnc_commodity *from, const gnc_commodity *to, PriceTuple tu
 
 static gnc_numeric
 indirect_price_conversion (GNCPriceDB *db, const gnc_commodity *from,
-                           const gnc_commodity *to, time64 t)
+                           const gnc_commodity *to, time64 t, gboolean before_date)
 {
     GList *from_prices = NULL, *to_prices = NULL;
     PriceTuple tuple;
@@ -2530,6 +2530,12 @@ indirect_price_conversion (GNCPriceDB *db, const gnc_commodity *from,
             so avoid getting them if they aren't needed. */
         if (from_prices)
             to_prices = gnc_pricedb_lookup_latest_any_currency(db, to);
+    }
+    else if (before_date)
+    {
+        from_prices = gnc_pricedb_lookup_latest_before_any_currency_t64 (db, from, t);
+        if (from_prices)
+            to_prices = gnc_pricedb_lookup_latest_before_any_currency_t64 (db, to, t);
     }
     else
     {
@@ -2550,7 +2556,7 @@ indirect_price_conversion (GNCPriceDB *db, const gnc_commodity *from,
 
 static gnc_numeric
 direct_price_conversion (GNCPriceDB *db, const gnc_commodity *from,
-                         const gnc_commodity *to, time64 t)
+                         const gnc_commodity *to, time64 t, gboolean before_date)
 {
     GNCPrice *price;
     gnc_numeric retval = gnc_numeric_zero();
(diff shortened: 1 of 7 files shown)


christopherlam:
IIUC this also fixes https://bugs.gnucash.org/show_bug.cgi?id=743753 -- I'd been trying to fix this but my C skills are obviously not good enough. If this is correct then we can introduce a new price source: `pricedb-before` (cf `pricedb-latest` and `pricedb-nearest`).

libgnucash/engine/gnc-pricedb.c
```diff
@@ -2577,21 +2585,22 @@ gnc_numeric
 gnc_pricedb_get_nearest_price (GNCPriceDB *pdb,
                                const gnc_commodity *orig_currency,
                                const gnc_commodity *new_currency,
-                               const time64 t)
+                               const time64 t,
+                               gboolean before_date)
```
christopherlam, line 2589: I'm not 100% sure it'll be nice to amend the signature of this function. As you notice, it affects custom reports. Wouldn't adding `gnc_pricedb_get_nearest_before_price` be a safer approach?

jralls:
> IIUC this also fixes https://bugs.gnucash.org/show_bug.cgi?id=743753

No, to fix that `latest-before` and `nearest-before` need to be added as price-source options. Both of the functions @Bob-IT refers to are already wrapped so I wouldn't think you'd need to code anything in C to add those options.

jralls:
> Wouldn't adding gnc_pricedb_get_nearest_before_price be a safer approach?

+1. That sort of boolean parameter is OK in internal implementation functions but makes for an ugly public API.

christopherlam:
The other issue is whether we should amend the behaviour of reports, or add another option. The following should add the new price option, however I'd think 'nearest' should really be amended to mean 'nearest before' but some users may complain.

```patch
diff --git a/gnucash/report/commodity-utilities.scm b/gnucash/report/commodity-utilities.scm
index 21f723832..6419bc243 100644
--- a/gnucash/report/commodity-utilities.scm
+++ b/gnucash/report/commodity-utilities.scm
@@ -741,6 +741,21 @@
              (gnc:gnc-monetary-commodity foreign)
              domestic (time64CanonicalDayTime date))))))
 
+(define (gnc:exchange-by-pricedb-nearest-before
+         foreign domestic date)
+  (and (record? foreign)
+       (gnc:gnc-monetary? foreign)
+       date
+       (or (gnc:exchange-by-euro foreign domestic date)
+           (gnc:exchange-if-same foreign domestic)
+           (gnc:make-gnc-monetary
+            domestic
+            (gnc-pricedb-convert-balance-latest-before-price-t64
+             (gnc-pricedb-get-db (gnc-get-current-book))
+             (gnc:gnc-monetary-amount foreign)
+             (gnc:gnc-monetary-commodity foreign)
+             domestic (time64CanonicalDayTime date))))))
+
 ;; Exchange by the nearest price from pricelist. This function takes
 ;; the <gnc-monetary> 'foreign' amount, the <gnc:commodity*>
 ;; 'domestic' commodity, a <gnc:time64> 'date' and the
@@ -783,6 +798,9 @@
                          (gnc:make-exchange-alist
                           report-currency to-date-tp)))
     ((pricedb-latest) gnc:exchange-by-pricedb-latest)
+    ((pricedb-before) (lambda (foreign domestic)
+                        (gnc:exchange-by-pricedb-nearest-before
+                         foreign domestic to-date-tp)))
     ((pricedb-nearest) (lambda (foreign domestic)
                          (gnc:exchange-by-pricedb-nearest
                           foreign domestic to-date-tp)))
@@ -803,6 +821,8 @@
 (define (gnc:case-price-fn source target-curr date)
   (define pdb (gnc-pricedb-get-db (gnc-get-current-book)))
   (case source
+    ((pricedb-before) (cut gnc-pricedb-get-nearest-before-price pdb <> target-curr
+                           (time64CanonicalDayTime date)))
     ((pricedb-nearest) (cut gnc-pricedb-get-nearest-price pdb <> target-curr
                             (time64CanonicalDayTime date)))
     ((pricedb-latest)  (cut gnc-pricedb-get-latest-price pdb <> target-curr))
@@ -850,6 +870,7 @@
                              pricealist foreign domestic date))))
     ((pricedb-latest) (lambda (foreign domestic date)
                         (gnc:exchange-by-pricedb-latest foreign domestic)))
+    ((pricedb-before) gnc:exchange-by-pricedb-nearest-before))
     ((pricedb-nearest) gnc:exchange-by-pricedb-nearest)
     (else
      (begin
diff --git a/gnucash/report/options-utilities.scm b/gnucash/report/options-utilities.scm
index 2d011b76e..ce982363b 100644
--- a/gnucash/report/options-utilities.scm
+++ b/gnucash/report/options-utilities.scm
@@ -144,6 +144,7 @@
     (list (vector 'average-cost (N_ "Average cost of purchases by volume-weighted"))
           (vector 'weighted-average (N_ "Weighted average of all past currency transactions"))
           (vector 'pricedb-latest (N_ "Most recent"))
+          (vector 'pricedb-before (N_ "Nearest before report date"))
           (vector 'pricedb-nearest (N_ "Nearest to report date"))))))
 
 ;; The width- and height- options for charts
```

jralls:
>however I'd think 'nearest' should really be amended to mean 'nearest before' but some users may complain.

It's use-case dependent. In one of the bugs I used the example of a net-worth chart and a user who retrieves prices only weekly. In that case it makes more sense to use the price closest to the chart point's date rather than the one before. A balance sheet OTOH shouldn't use a future date in most cases. I guess we could have a single option and try to guess which version is more appropriate for the report, but ISTM it's both safer and less work to have two (or maybe 3) options.

christopherlam:
>  I guess we could have a single option and try to guess which version is more appropriate for the report, but ISTM it's both safer and less work to have two (or maybe 3) options.

Ok we can simply amend default price source for each report.

libgnucash/engine/gnc-pricedb.c
Bob-IT: Agree, I have created functions as suggested.

Bob-IT:
OK, I have added new function as Chris suggested and added new tests for the new pricedb functions added.
Will have a look at the inconsistency with using 'const Account' and 'const gnc_commodity' tomorrow hopefully.

christopherlam:
Thanks. My last feedback is a request to standardise - use either `nearest_before` (better I think) or `latest_before`, but not both!

christopherlam:
> use either `nearest_before` (better I think) or `latest_before`, but not both!

followup - it would seem `gnc_pricedb_lookup_latest_before_t64` and `gnc_pricedb_lookup_latest_before_any_currency_t64` were defined and tested, but never used... perhaps they should be renamed to `nearest_before`???

jralls:
Since they're SWIGged it's possible, if unlikely, someone is using them in a custom report. So rename is OK if that's the name you prefer but add a deprecated scheme wrapper with the old name so as to not break anything.

Bob-IT:
I think that is it for me, @christopherlam I will leave you to add the scheme changes once I have pushed this. As you can see I needed to change gnucash_core.py for the rename of function which showed up as a test failure, needed to enable python bindings on my build system not that I know any thing about python as I did not see that test. I looked at the 'const Account' inconsistency but that was going to make allot of changes so will leave for a different PR.

jralls:
Mrmph. Didn't think about Python. I guess it needs a wrapper too. I'll do that in a bit.

@christopherlam  and I can push directly to this PR if it won't mess up your workflow too much.

Bob-IT:
Which ever is easiest, I can push this tomorrow.

christopherlam:
Go ahead, the `latest_before` renamer and test suite is ready. "After you".

jralls:
@christopherlam Do you want to push your wrappers here or wait?

christopherlam:
Not sure how to push to this branch.

```patch
modified   bindings/guile/engine.scm
@@ -31,11 +31,21 @@
                    (gnucash engine commodity-table)
                    (gnucash engine gnc-numeric))
 
+(export gnc-pricedb-lookup-latest-before-t64)
+(export gnc-pricedb-lookup-latest-before-any-currency-t64)
 (export gnc:account-map-descendants)
 (export gnc:account-map-children)
 (export account-full-name<?)
 (export accounts-get-children-depth)
 
+(define (gnc-pricedb-lookup-latest-before-t64 . args)
+  (issue-deprecation-warning "gnc-pricedb-lookup-latest-before-t64 has been renamed to gnc-pricedb-lookup-nearest-before-t64")
+  (apply gnc-pricedb-lookup-nearest-before-t64 args))
+
+(define (gnc-pricedb-lookup-latest-before-any-currency-t64 . args)
+  (issue-deprecation-warning "gnc-pricedb-lookup-latest-before-any-currency-t64 has been renamed to gnc-pricedb-lookup-nearest-before-any-currency-t64")
+  (apply gnc-pricedb-lookup-nearest-before-any-currency-t64 args))
+
 ;; A few account related utility functions which used to be in engine-utilities.scm
 (define (gnc:account-map-descendants thunk account)
   (issue-deprecation-warning "gnc:account-map-descendants is deprecated.")
modified   bindings/guile/test/CMakeLists.txt
@@ -71,6 +71,7 @@ if (HAVE_SRFI64)
 
     set (scm_tests_with_srfi64_SOURCES
         test-business-core.scm
+        test-scm-engine.scm
         )
 
     gnc_add_scheme_test_targets (scm-test-with-srfi64
new file   bindings/guile/test/test-scm-engine.scm
@@ -0,0 +1,20 @@
+(use-modules (srfi srfi-64))
+(use-modules (tests srfi64-extras))
+(use-modules (gnucash engine))
+
+(define (run-test)
+  (test-runner-factory gnc:test-runner)
+  (test-begin "test-engine")
+  (test-engine)
+  (test-end "test-engine"))
+
+(define (test-engine)
+  (test-begin "testing function availability")
+
+  (test-assert "testing gnc-pricedb-lookup-latest-before-t64"
+    (gnc-pricedb-lookup-latest-before-t64 '() '() '() 0))
+
+  (test-assert "testing gnc-pricedb-lookup-latest-before-any-currency-t64"
+    (gnc-pricedb-lookup-latest-before-any-currency-t64 '() '() 0))
+
+  (test-end "testing deprecated functions"))
```

Bob-IT:
@christopherlam I have just pushed to code so you can push your changes as a followup.
I also do not know how to push to a PR, @jralls would you like to enlighten us, thank you both.

jralls:
Create a branch that mirrors the PR branch.  Having rebased, @Bob-IT's bug798148 branch's root commit is a080a0c (visit the branch on GitHub and use History to find the latest mainline commit) so
```
git fetch
git checkout -b bug798148 a080a0c
git pull https://github.com/Bob-IT/gnucash bug798148
```
Pretty much what you'd always do when you want to test-build a PR.
Edit and commit, then push:
```
git push git@github.com:Bob-IT/gnucash bug798148
```

libgnucash/engine/gnc-pricedb.h
```diff
@@ -573,19 +571,27 @@ gnc_numeric gnc_pricedb_get_nearest_before_price (GNCPriceDB *pdb,
                                                   const gnc_commodity *new_currency,
                                                   const time64 t);
 
-/** @brief Retrieve the price one currency to another at specified date
+/** @brief Retrieve the price one currency to another using the price
```
fellen, line 574: That is doxygen and generates our api docs..
Docbook is used as source language in gnucash-docs.
;-)

christopherlam:
> Create a branch that mirrors the PR branch. Having rebased, @Bob-IT's bug798148 branch's root commit is a080a0c (visit the branch on GitHub and use History to find the latest mainline commit) so

I've immortalized how the classic programmers Alice can push onto Bob's branches in the wiki -- https://wiki.gnucash.org/wiki/Git#Pushing_onto_another_core_developer_branch -- please verify.

jralls:
Unfortunately you got it wrong*, plus I think that it's not about whether Bob is a core developer but whether the branch in his repo is the source for a pull request.

\* You left off making sure that Alice's new branch needs to branch from the same commit as Bob's PR branch, otherwise it will create a merge commit and pushing back will make a mess. If the branches in Alice's and Bob's repos aren't named the same then Alice needs to push with `bob-dev:maint`.

PR #983, opened 2021-04-24
xaccTransOrder sort on num in string compare

holymonson:
When `num` exceeds INT64_MAX (e.g. "2018112610193018150000000000140090141420") or contains non-digit chars (e.g. "SH210061023488"), sorting in integer compare will fail. We could trying using string compare as fallback.

### DIFF
--- a/libgnucash/engine/Transaction.c
+++ b/libgnucash/engine/Transaction.c
@@ -1890,7 +1890,8 @@ xaccTransOrder_num_action (const Transaction *ta, const char *actna,
                             const Transaction *tb, const char *actnb)
 {
     char *da, *db;
-    int na, nb, retval;
+    long long int na, nb;
+    int retval;
 
     if ( ta && !tb ) return -1;
     if ( !ta && tb ) return +1;
@@ -1907,19 +1908,27 @@ xaccTransOrder_num_action (const Transaction *ta, const char *actna,
             return (ta_is_closing - tb_is_closing);
     }
 
-    /* otherwise, sort on number string */
-    if (actna && actnb) /* split action string, if not NULL */
+    /* otherwise, sort on split action string, if not NULL */
+    if (actna && actnb)
     {
-        na = atoi(actna);
-        nb = atoi(actnb);
+        retval = g_strcmp0(actna, actna);
+        if (retval)
+            return retval;
     }
-    else                /* else transaction num string */
+
+    /* otherwise, sort on number string in integer compare, if it's "short" */
+    if (strlen(ta->num) <= 18 && strlen(tb->num) <= 18)
     {
-        na = atoi(ta->num);
-        nb = atoi(tb->num);
+        na = atoll(ta->num);
+        nb = atoll(tb->num);
+        if (na < nb) return -1;
+        if (na > nb) return +1;
     }
-    if (na < nb) return -1;
-    if (na > nb) return +1;
+
+    /* otherwise, sort on number string in string compare */
+    retval = g_strcmp0(ta->num, tb->num);
+    if (retval)
+        return retval;
 
     if (ta->date_entered != tb->date_entered)
         return (ta->date_entered > tb->date_entered) - (ta->date_entered < tb->date_entered);


jralls:
It would be `INT32_MAX`, but even then that use case is dubious at best. It's very unlikely that users would create transaction numbers in the billions. That's a good thing because passing a string that would overflow an `int` to `atoi` is undefined behavior.

You're on firmer ground with non-numerics in the string, but your approach is overly simplistic: 
* It relies on both `atoi` calls returning the same value, presumably 0. 
*  `strcmp` does lexical ordering so that "sh3" will sort after "sh21"; that may or may not be what the user expects.
* This is a two-value ordering function so it can produce inconsistent results if some num strings posted the same day resolve correctly with  `atoi` and others don't.
* `strcmp` produces really weird results when handed a string containing code points > U+007f. `g_utf8_collate` would at least deal with that, though not with the other problems.

One more problem: You ignore the possibility that the book uses the Split `action` field for `num`.

christopherlam:
@jralls, the transaction number shown here is a timestamp.

jralls:
@christopherlam Maybe "20181126101930" is, but no existing clock can discern 10<sup>-24</sup> seconds that would make it meaningful if you put a decimal separator after that first part. I think it more likely that it's a contrived number that happens to start off with something that might resemble a time stamp. Highly contrived: It would require 131 bits to represent it in binary, so it's too big even for a `GncInt128`.

holymonson:
In practice, `num` is usually not created by user but imported from external records, so it would be any arbitrary string (less than 2048 in length), not necessarily a number. For those `num` with long number or containing non-digit, they are very likely has the same length, so "sh3" vs "sh21" cases may be rare. We can only handle pure number and treat all other cases as literal string.

`atoll` will return 0 if the first char isn't a digit, otherwise it will parse integer as more as it can. `strcmp` may produce weird result for non-ascii, but at least the result is consistent. If people do use >U+007f chars in `num`, I think they maybe not really care about "right" order, as long as it's a consistent one.

At present `action` will be apply to `atoi`, it seems not we want. I have pushed a new commit, it doesn't fix all possible problems, but it may be the best I could offer.

jralls:
> In practice, num is usually not created by user but imported from external records,

In *your* practice, maybe. GnuCash's basic design principle is that the user is maintaining their own books and enters most transactions by hand as they are recognized. Imports are for converting from another program and as a convenience for reconciling one's books with other parties like banks and credit card companies.

> atoll will return 0 if the first char isn't a digit, otherwise it will parse integer as more as it can.

That's true as long as it doesn't overflow an `int`. If it does overflow then it's undefined behavior, meaning that it's up to the compiler and C standard library what it does, but the result can not be expected to be a good one. Discarding the extra digits and crashing would both be reasonable outcomes; overwriting memory outside the `int` would be careless on the part of the runtime library author but would still comply with the standard and it wouldn't surprise me a bit if some poorly-maintained runtimes (NetBSD comes to mind) would do that.

Next, if the first character is between U+0030 and U+0039 and a later character isn't then the numerical part gets returned and the rest of the string is ignored: `2SH1123456` and `2SH654321` will compare as equal. Not likely what the user wants.

> strcmp may produce weird result for non-ascii, but at least the result is consistent.

That's not correct because Unicode has 4 ways ("normalizations") of representing code points with diacritics.

>  I think they maybe not really care about "right" order, as long as it's a consistent one.

Users are picky about things like that. If it doesn't sort correctly according to their locale they'll complain.

> At present action will be apply to atoi, it seems not we want. I have pushed a new commit, it doesn't fix all possible problems

No, you misunderstand how the preference works: It changes the semantics of the split-action field so that it's used as a transaction number (and we use that word on purpose); it's intended to permit different splits in the same transaction to have different numbers: For example a user might put a check number in the checking account split and the vendor's invoice number in the expense account split. The same logic must be applied to both options. That's best done by extracting a function to do the comparison and apply it to the appropriate field depending on the value of the option.

> but it may be the best I could offer

To work correctly the actual function will need to check the lengths of the strings, test if both contain only numerics and is short enough to fit in a reasonable numeric value; the largest available in GnuCash is `GncInt128` so that would be 38 characters. If both meet that then do an numeric conversion for a numeric sort, otherwise do a lexical sort. It should be fully unicode-aware. Glib provides all the [unicode and utf8 functions you need](https://developer.gnome.org/glib/stable/glib-Unicode-Manipulation.html) to do this. That will address all of the problems except the somewhat odd results that will arise from trying to sort a list containing some all-numerics and some not where some string pairs are ordered numerically and others lexically.

There's nothing particularly difficult about implementing that, it's just a little tedious. There's no string constructor for `GncInt128` and no portable way in C to get `int64_t` (C++ has `strtoll`) so if you want to allow  numbers larger than 8 digits  you'll need to build it in pieces with `atoi`.

jralls:
@holymonson Have you abandoned this?

jralls:
@holymonson I think this is what you're trying to do:
```
--- a/libgnucash/engine/Transaction.c
+++ b/libgnucash/engine/Transaction.c
@@ -1885,12 +1885,37 @@ xaccTransOrder (const Transaction *ta, const Transaction *tb)
     return xaccTransOrder_num_action (ta, NULL, tb, NULL);
 }

+inline static gboolean
+is_all_digits (const char* str)
+{
+     int len = strlen(str);
+     for (int i = 0; i < len; ++i)
+          if (str[i] < 0x30 || str[i] > 0x39)
+               return FALSE;
+     return TRUE;
+}
+
+static int
+order_by_string_or_int64 (const char* a, const char* b)
+{
+     static const int max_int64_len = 18;
+     if (strlen (a) < max_int64_len && strlen (b) < max_int64_len &&
+         is_all_digits (a) && is_all_digits (b))
+     {
+          int64_t na = atoll(a);
+          int64_t nb = atoll(b);
+          return na < nb ? -1 : na > nb ? 1 : 0;
+     }
+     return strcmp(a, b);
+}
+
 int
 xaccTransOrder_num_action (const Transaction *ta, const char *actna,
                             const Transaction *tb, const char *actnb)
 {
     char *da, *db;
-    int na, nb, retval;
+    int retval;
+    int64_t na, nb;

     if ( ta && !tb ) return -1;
     if ( !ta && tb ) return +1;
@@ -1910,16 +1935,14 @@ xaccTransOrder_num_action (const Transaction *ta, const
char *actna,
     /* otherwise, sort on number string */
     if (actna && actnb) /* split action string, if not NULL */
     {
-        na = atoi(actna);
-        nb = atoi(actnb);
+         retval = order_by_string_or_int64 (actna, actnb);
     }
     else                /* else transaction num string */
     {
-        na = atoi(ta->num);
-        nb = atoi(tb->num);
+         retval = order_by_string_or_int64 (ta, tb);
     }
-    if (na < nb) return -1;
-    if (na > nb) return +1;
+    if (retval)
+         return retval;

     if (ta->date_entered != tb->date_entered)
         return (ta->date_entered > tb->date_entered) - (ta->date_entered < tb->date_entered);
```

holymonson:
@jralls I think your patch is great, would you make it a new PR yourself and closing this one?

jralls:
@holymonson I've done the commit and asked for comments in the gnucash-user mailing list about it. There's been a little push back and I'm considering how to refine it in response.

You're right, might as well close this PR.

PR #985, opened 2021-04-25
gnc-pricedb: plug memory leak

christopherlam:
plug memory leak. use `g_list_find`.

### DIFF
--- a/libgnucash/engine/gnc-pricedb.c
+++ b/libgnucash/engine/gnc-pricedb.c
@@ -2571,27 +2571,37 @@ direct_price_conversion (GNCPriceDB *db, const gnc_commodity *from,
     return retval;
 }
 
+static gnc_numeric
+get_nearest_price (GNCPriceDB *pdb,
+                   const gnc_commodity *orig_curr,
+                   const gnc_commodity *new_curr,
+                   const time64 t,
+                   gboolean before)
+{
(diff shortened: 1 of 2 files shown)


libgnucash/engine/gnc-pricedb.c
```diff
@@ -1901,17 +1891,17 @@ latest_before (PriceList *prices, const gnc_commodity* target, time64 t)
...
-        else
-        {
-            gnc_price_ref(price);
-            retval = g_list_prepend(retval, price);
-            found_coms = g_list_prepend(found_coms, com == target ? cur : com);
-        }
+
+        gnc_price_ref (price);
+        retval = g_list_prepend (retval, price);
+        found_coms = g_list_prepend (found_coms, com == target ? cur : com);
```
jralls, line 1902: `found_coms` seems redundant.

jralls (approved):

libgnucash/engine/gnc-pricedb.c
christopherlam: I've left it because I don't know the assumptions about `prices` enough to modify the logic...

jralls: Never mind, not redundant, the `retval` has prices and `found_coms` has commodities.

PR #989, opened 2021-04-28
Maint stock metadata

christopherlam:
Draft backend and UI for stock-account metadata. This will set up 4 "linked" accounts for a stock-account.

- Dividend & Capgains must be INCOME
- Cash/proceeds must be BANK/ASSET/CASH
- Fees must be EXPENSE

Only 1 account is allowable for each. All of these accounts *must* have the currency `gnc_account_get_currency_or_parent (stock_account)` and cannot be placeholder. Account Properties UI will get/set these accounts, and will be used by both #818 and #978.

### DIFF
--- a/gnucash/gtkbuilder/dialog-account.glade
+++ b/gnucash/gtkbuilder/dialog-account.glade
@@ -1866,6 +1866,139 @@
                 <property name="tab_fill">False</property>
               </packing>
             </child>
+            <child>
+              <object class="GtkBox" id="vbox303">
+                <property name="visible">True</property>
+                <property name="can_focus">False</property>
+                <property name="border_width">6</property>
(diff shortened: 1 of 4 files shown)


jralls:
>Dividend & Capgains must be INCOME
Cash/proceeds must be BANK/ASSET/CASH
Fees must be EXPENSE
Only 1 account is allowable for each. All of these accounts must have the currency gnc_account_get_currency_or_parent (stock_account) and cannot be placeholder.

Not a single word of that is true. If this is about planning for future work then it's premature to make a PR of it.

But has it occurred to you that this is going a bit far if it's useful only to Canadians?

christopherlam:
Outvoted. https://code.gnucash.org/logs/2021/04/29.html#T18:22:04

PR #996, opened 2021-05-04
Revist Bug798093

jralls:
Reverts ed486a5 as that was based on seriously misunderstanding what was going on.

0f8d9f6 changes the way trading accounts are found from looking at their names and ignoring account types to looking for account-types and commodities and ignoring names in order to fix [Bug 798093](https://bugs.gnucash.org/show_bug.cgi?id=798093) with the added (supposed) advantage of allowing the top level trading account to have any name.

Sadly I overlooked a key issue in that commit: The top-level trading account and any placeholders were created with the transaction currency of the transaction that caused them to be created. The new logic wouldn't find the existing ones if the current transaction has a different currency, so it would make a new hierarchy.

This PR changes in two ways: When looking for the top-level trading account and namespace placeholder account it will accept any account currency, and if it needs to create one or both it will get the book's root account currency (if the book is old and doesn't have one it will use the currency of the first top-level income account). If, thanks to 0f8d96 there's more than one top-level trading account or namespace placeholder it will prefer one in the root account-currency but if that doesn't exist it will use the first one found.

### DIFF
--- a/libgnucash/engine/Scrub.c
+++ b/libgnucash/engine/Scrub.c
@@ -530,50 +530,6 @@ get_trading_split (Transaction *trans, Account *base,
     return balance_split;
 }
 
-/* Find the trading split for a commodity, but don't create any splits
-   or accounts if they don't already exist. */
-static Split *
-find_trading_split (Transaction *trans, Account *root,
-                    gnc_commodity *commodity)
-{
-    Account *trading_account;
-    Account *ns_account;
-    Account *account;
-
-    if (!root)
-    {
-        root = gnc_book_get_root_account (xaccTransGetBook (trans));
-        if (NULL == root)
-        {
-            /* This can't occur, things should be in books */
-            PERR ("Bad data corruption, no root account in book");
-            return NULL;
-        }
-    }
-
-    trading_account = gnc_account_lookup_by_name (root, _("Trading"));
-    if (!trading_account)
-    {
-        return NULL;
-    }
-
-    ns_account = gnc_account_lookup_by_name (trading_account,
-                                             gnc_commodity_get_namespace(commodity));
-    if (!ns_account)
-    {
-        return NULL;
-    }
-
-    account = gnc_account_lookup_by_name (ns_account,
-                                          gnc_commodity_get_mnemonic(commodity));
-    if (!account)
-    {
-        return NULL;
-    }
-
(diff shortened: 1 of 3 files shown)


libgnucash/engine/Scrub.c
```diff
@@ -1450,43 +1454,75 @@ xaccAccountScrubColorNotSet (QofBook *book)
...
-
-        /* Hang the account off the root. */
-        gnc_account_append_child (root, acc);
-        xaccAccountCommitEdit (acc);
+        return construct_account (root, currency, accname,
+                                  acctype, placeholder);
     }
 
-    return acc;
+    if (g_list_length(acc_list) > 1)
```
christopherlam, line 1514:
```suggestion
    if (acc_list->next)
```

jralls: Not a huge improvement, the length is likely to be either 1 or 2, but I switched it to `if (g_list_next(acc_list))` which does the same thing as your version but uses the `GList` API.

mtalexander:
The changes to the TransScrubOrphansFast function and reports/standard/test/test-bdget.scm seem unrelated and I didn't include them in my testing.

Both find_root_currency and xaccScrubUtilityGetOrMakeAccount seem to leak a GList sometimes.

It seems like it would be better in xaccScrubUtilityGetOrMakeAccount if checkName is false and there are two or more accounts that match type and commodity, then prefer one that matches the name too, if any.  That way if the file has two trading accounts for the same commodity it will continue to use the one with the new name rather than (perhaps) switching back to the old one.

The transaction that can't be balanced still can't be balanced.   I'll add something to the bug report about this.

mtalexander:
find_root_currency calls gnc_account_get_children which returns a copy of the account list.  It is this copy that I think is being leaked.  Or it could just access the child list directly and not make a copy of it.

jralls:
Ah, you're right, `gnc_account_get_children` does indeed make a copy of the list.

mtalexander:
I tested your changes up to e7b94d7 (which I think is current) and everything seems to work much better.  It seems to use the existing trading account, if any, in all cases.  It also cleans up a transaction with multiple trading accounts when it is rebalanced.  Thanks for cleaning this up.  I clearly didn't think about renaming commodities when I wrote that code.

jralls:
@mtalexander Thanks for the reviews and catching the leaks.

PR #1006, opened 2021-05-18
Launch external browser to load current report

christopherlam:
Mainly of benefit for Windows builds; will use modern browser instead of libwebkit1. To reuse the previously generated `gnc-report-XXXXXX.html` is unreliable -- it may have been deleted between internal browser rendering and launching external browser.

### DIFF
--- a/gnucash/gnome/gnc-plugin-page-report.c
+++ b/gnucash/gnome/gnc-plugin-page-report.c
@@ -178,6 +178,7 @@ static void gnc_plugin_page_report_stop_cb(GtkAction *action, GncPluginPageRepor
 static void gnc_plugin_page_report_save_cb(GtkAction *action, GncPluginPageReport *rep);
 static void gnc_plugin_page_report_save_as_cb(GtkAction *action, GncPluginPageReport *rep);
 static void gnc_plugin_page_report_export_cb(GtkAction *action, GncPluginPageReport *rep);
+static void gnc_plugin_page_report_open_external_browser_cb (GtkAction *action, GncPluginPageReport *report);
 static void gnc_plugin_page_report_options_cb(GtkAction *action, GncPluginPageReport *rep);
 static void gnc_plugin_page_report_print_cb(GtkAction *action, GncPluginPageReport *rep);
 static void gnc_plugin_page_report_exportpdf_cb(GtkAction *action, GncPluginPageReport *rep);
@@ -1142,6 +1143,7 @@ static action_toolbar_labels toolbar_labels[] =
        to be used as toolbar button label. */
     { "ReportSaveAsAction", N_("Save Config As...") },
     { "FilePrintPDFAction", N_("Make Pdf") },
+    { "OpenExternalBrowserAction", N_("External") },
     { NULL, NULL },
 };
 
@@ -1205,6 +1207,11 @@ gnc_plugin_page_report_constr_init(GncPluginPageReport *plugin_page, gint report
             "FilePrintAction", "document-print", N_("_Print Report..."), "<primary>p",
             N_("Print the current report"),
             G_CALLBACK(gnc_plugin_page_report_print_cb)
+        },
+        {
+            "OpenExternalBrowserAction", "emblem-web",
+            N_("Open in _external browser..."), NULL, N_("Open in external browser"),
+            G_CALLBACK(gnc_plugin_page_report_open_external_browser_cb)
         },
         {
             "FilePrintPDFAction", GNC_ICON_PDF_EXPORT, N_("Export as P_DF..."), NULL,
@@ -1941,6 +1948,19 @@ gnc_plugin_page_report_print_cb( GtkAction *action, GncPluginPageReport *report 
 #endif
 
     g_free (job_name);
+}
+
+static void
+gnc_plugin_page_report_open_external_browser_cb (GtkAction *action,
+                                                 GncPluginPageReport *report)
+{
+    GncPluginPageReportPrivate *priv = GNC_PLUGIN_PAGE_REPORT_GET_PRIVATE(report);
+    GError *error = NULL;
+    if (!gnc_html_open_external_browser (priv->html, &error))
+    {
+        PWARN ("Failed to open uri: %s", error->message);
+        g_error_free (error);
+    }
 }
 
 static void
--- a/gnucash/html/gnc-html.c
+++ b/gnucash/html/gnc-html.c
@@ -509,6 +509,28 @@ gnc_html_copy_to_clipboard( GncHtml* self )
     }
 }
 
+gboolean
+gnc_html_open_external_browser (GncHtml *self, GError **error)
+{
+    static gchar *template = "gnc-report-XXXXXX.html";
+    gchar *filename = g_build_filename (g_get_tmp_dir(), template, NULL);
+
+#ifdef G_OS_WIN32
+    gchar *uri = g_strdup_printf ("file:///%s", filename);
+#else
+    gchar *uri = g_strdup_printf ("file://%s", filename);
+#endif
+
+    gboolean retval = gnc_html_export_to_file (self, filename);
+
+    if (retval)
+        retval = g_app_info_launch_default_for_uri (uri, NULL, error);
+
+    g_free (uri);
+    g_free (filename);
+    return retval;
+}
+
 /**************************************************************
  * gnc_html_export_to_file : wrapper around the builtin function in gtkhtml
  **************************************************************/
--- a/gnucash/html/gnc-html.h
+++ b/gnucash/html/gnc-html.h
@@ -195,6 +195,15 @@ void gnc_html_reload( GncHtml* html, gboolean view );
 void gnc_html_copy_to_clipboard( GncHtml* html );
 
 /**
+ * Launches external browser and load html
+ *
+ * @param html GncHtml object
+ * @param GError object
+ * @return TRUE if successful, FALSE if unsuccessful
+ */
+gboolean gnc_html_open_external_browser (GncHtml *html, GError **error);
+
+/**
  * Exports the html to an external file.
  *
  * @param html GncHtml object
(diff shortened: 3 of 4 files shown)


jralls (changes requested):
I'm surprised, I didn't think GAppInfo worked on macOS, but it does.
Does this really deserve a toolbar button? Should the account links be suppressed since they don't work in an external browser because there's no handler?

christopherlam:
> Does this really deserve a toolbar button?

Maybe not. But how to trigger? Menu item?

> Should the account links be suppressed since they don't work in an external browser because there's no handler?

This requires communicating with the style sheet, and is difficult. Or requires post-processing the HTML to remove `<A HREF>` tags, also difficult.

Bob-IT:
Just some thoughts...
Maybe this is a candidate for a preference to open all reports in external browser as it could be new behavior.
Is this so the new chart-js would work on Windows?
Then the preference could be open reports with charts in external browser but maybe that would be confusing.
I do think the links should be suppressed as they will not work.
Would each report be in a new browser or added to an existing browser?
I have no idea if this is possible maybe send them all to the same browser instance.

christopherlam:
> Just some thoughts...
> Maybe this is a candidate for a preference to open all reports in external browser as it could be new behavior.

It would be very awkward to set options then.

> Is this so the new chart-js would work on Windows?

No; chartjs-2 already exhibits https://bugs.gnucash.org/show_bug.cgi?id=797876 in old webkit and I suspect chartjs-3 will leave old webkit further behind. I haven't tested though. The purpose is to allow basic functionality in most reports (text or chart) but show all bells&whistles in a modern browser when the user wants final output.

> Then the preference could be open reports with charts in external browser but maybe that would be confusing.
> I do think the links should be suppressed as they will not work. If say a parameter was passed on normal report creation to disable links then I would think it all could be done as part of the report creation.

Ok but more complexity. The links being disabled will also change the output (I'm rather fond of the numbers being underlined and blue :open_mouth: ) so this will require a popularity vote to link this code to the report stylesheet.

> Would each report be in a new browser or added to an existing browser?

It currently launches a new tab in the default browser.

Bob-IT:
> It would be very awkward to set options then.

Had not thought about that.

Looks to me that this is bypassing the export to HTML file so maybe it would be possible to augment that option to export to file and then load it in the external browser with another button in that file dialog, this would remove the toolbar button.

gjanssens:
IMHO opening in an external browser should be considered the exceptional case as we lose integration (internal links no longer work, setting options will open a new browser tab,...)

Ideally even if webkit is becoming obsolete on Windows we should write a wrapper around another browser engine on that platform and use that in the gnc-html factory. I never looked at the details but WxWidgets does just that for their [wxWebView widget](https://docs.wxwidgets.org/3.0/classwx_web_view.html)

gnucash/gnome/gnc-plugin-page-report.c
```diff
@@ -1140,6 +1141,7 @@ static action_toolbar_labels toolbar_labels[] =
        to be used as toolbar button label. */
     { "ReportSaveAsAction", N_("Save Config As...") },
     { "FilePrintPDFAction", N_("Make Pdf") },
+    { "OpenExternalBrowserAction", N_("External") },
```
jralls, line 1144: "External" is a bit too terse even in English; it at least needs a context and maybe a translator comment to explain it.

jralls:
> To reuse the previously generated gnc-report-XXXXXX.html is unreliable -- it may have been deleted between internal browser rendering and launching external browser.

Why do you think that?

christopherlam:
> "External" is a bit too terse even in English; it at least needs a context and maybe a translator comment to explain it.

`Open in system browser`? Too long.

> > To reuse the previously generated gnc-report-XXXXXX.html is unreliable -- it may have been deleted between internal browser rendering and launching external browser.
> 
> Why do you think that?

The user may have purged /tmp between internal browser rendering and launching external browser. I note pressing Reload toolbar also creates multiple `gnc-report-XXXXXX.html` files.

jralls:
> The user may have purged /tmp between internal browser rendering and launching external browser. I note pressing Reload toolbar also creates multiple gnc-report-XXXXXX.html files.

The latter just makes more temp report files, it deletes nothing.

The former is extremely unlikely and might count as the user being malicious.  Such a malicious user could also suspend GnuCash and delete the file between writing it and telling the default browser to load it. There's no protection, but there's also no harm: If the temp file is deleted the browser just says that it couldn't find it.

christopherlam:
> The former is extremely unlikely and might count as the user being malicious. Such a malicious user could also suspend GnuCash and delete the file between writing it and telling the default browser to load it. There's no protection, but there's also no harm: If the temp file is deleted the browser just says that it couldn't find it.

Ok not saying it can't be done, just saying the existing `gnc-report-XXXXXX.html` mechanism lives in `gnc-html-webkit[12].c` and is not saved anywhere; to reuse it would mean a bunch of infrastructure works which seems slightly too much work: I figure if a simple "Reload" will create a new /tmp/ file, to launch external browser with a new /tmp/ file as well does not seem unreasonable.

Of more interest (to me) is whether to add a toolbar or a menu item.

gjanssens:
> > The former is extremely unlikely and might count as the user being malicious. Such a malicious user could also suspend GnuCash and delete the file between writing it and telling the default browser to load it. There's no protection, but there's also no harm: If the temp file is deleted the browser just says that it couldn't find it.
> 
> Ok not saying it can't be done, just saying the existing `gnc-report-XXXXXX.html` mechanism lives in `gnc-html-webkit[12].c` and is not saved anywhere; to reuse it would mean a bunch of infrastructure works which seems slightly too much work: I figure if a simple "Reload" will create a new /tmp/ file, to launch external browser with a new /tmp/ file as well does not seem unreasonable.

It would be unreasonable for reports that take a long time to generate. When users reload a report, that's usually the purpose (regenerating) so users expect it there. However just showing the same report in an external browser doesn't need the generation  step. It's already done.

> 
> Of more interest (to me) is whether to add a toolbar or a menu item, and what's the best text for it.

I would  make it a menu item first and wait for user reactions before making it a toolbar. I also like @Bob-IT 's idea to make it an export to html option.

christopherlam:
> It would be unreasonable for reports that take a long time to generate. When users reload a report, that's usually the purpose (regenerating) so users expect it there. However just showing the same report in an external browser doesn't need the generation step. It's already done.

Thankfully most reports are now fast.

Ok I think there's another use case -- generate report, go to register to modify data, return to report tab, and export report to HTML or PDF (or open with external browser). Reasoning would indicate the report shouldn't be regenerated. This means current PDF/HTML  exports are probably regenerating the reports incorrectly. So, I'd modify `gnc-html-webkit[12].c` to *save* the report string (or filename) in memory, and copy `gnc-report-XXXXXX.html` (or in-memory string) to the exported report filename when exporting HTML. Or is this taking it too far?

jralls:
I like @Bob-IT's idea better than making it a separate action at all, and I think it's acceptable as a minimum-effort addition. 

Links in exported html files is bugs [571318](https://bugs.gnucash.org/show_bug.cgi?id=571318) and [642223](https://bugs.gnucash.org/show_bug.cgi?id=642223). They're marked as enhancement requests but one could easily argue that exporting a private scheme into an html file for a normal browser's consumption is really a bug.

jralls:
> So, I'd modify gnc-html-webkit[12].c to save the report string (or filename) in memory, and copy gnc-report-XXXXXX.html (or in-memory string) to the exported report filename when exporting HTML/PDF. Or is this taking it too far?

No, I think it's right. `GtkWebKitWebView` has a string constructor and it's dumb to write the string out to a file only to read it back in again. Export *could* simply write the string out to a file, maybe making a trip through XSLT to strip the `gncFoo:` anchors, or it could regenerate the string through a modified-on-the-fly stylesheet that doesn't generate them in the first place, whichever is more performant.

> Thankfully most reports are now fast.

The report-generation part is. Crunching the data can take over a minute with a large book, nothing much to be done about that as long as we have to repeatedly traverse big GLists. That means there's two values of regenerate: The user clicking the reload tab usually means they've changed something in the book and it needs to be re-crunched, but exporting the html just means regenerating it without the links. Changing options is a depends: Things like changing the dates or the accounts will require a full regeneration; changing the stylesheet or the displayed columns just requires redoing the html (assuming that all of the possible column values are calculated on the first pass and cached).

christopherlam:
> The report-generation part is. Crunching the data can take over a minute with a large book, nothing much to be done about that as long as we have to repeatedly traverse big GLists. That means there's two values of regenerate: The user clicking the reload tab usually means they've changed something in the book and it needs to be re-crunched, but exporting the html just means regenerating it without the links. Changing options is a depends: Things like changing the dates or the accounts will require a full regeneration; changing the stylesheet or the displayed columns just requires redoing the html (assuming that all of the possible column values are calculated on the first pass and cached).

As mentioned in IRC, I'd be keen to sight such a datafile that takes a long time to create a report.

W.R.T. reusing report HTML there's mechanism in https://github.com/Gnucash/gnucash/blob/d174b9ad37688be663e03b017957f2fdef6e591a/gnucash/report/report-core.scm#L743 to cache the html text, but I do not understand from the C part why it's not used when calling `gnc:render-report` repeatedly.

jralls:
Because it's a lot easier to call C from SCM than the other way around?

christopherlam:
I can confirm in this branch, clicking 'External' to launch external browser does *not* call scheme to regenerate the report. Clicking 'Reload' does. So, the concerns raised in this PR don't apply. The last issue remaining is the UI to launch external browser. Either menu+toolbar (which could be limited to win32) or menu only.

gnucash/gnome/gnc-plugin-page-report.c
luzpaz: What would be more ideal ?

jralls: "Open in External Browser".

PR #1007, opened 2021-05-19
Bug 798188 - The Invoice Editor -> Printable Invoice toolbar button crashes on Windows

christopherlam:


jralls:
If you never put anything on the list, there will never be anything to find. If you add code to put ptrs on the list you also need code to take ptrs off the list when the pointed-to object is destroyed.

Why can't you use [GncComponentManager](https://github.com/Gnucash/gnucash/blob/maint/gnucash/gnome-utils/gnc-component-manager.h)?

christopherlam:
It's populated via `gnc_gobject_tracking_remember` in the `g_list_append` call. I'd copypasted this snippet from other code that use `gnc_gobject_tracking_get_list` to find existing pages (e.g. finding account or budget page).

jralls:
Ah, I didn't remember that lame effort to track leaks. I don't think repurposing it is a good idea. As an aside, there are better tools (e.g. valgrind) for finding leaks; `gnc_gobject_tracking` is not very useful overhead and we'd be better off without it. 

On the one hand this is equivalent and much less clear than just using `g_object_weak_ref` directly, that's how `gnc_gobject_tracking` works. On the other this won't work anyway because [`gnc_plugin_page_report_init`](https://github.com/Gnucash/gnucash/blob/maint/gnucash/gnome/gnc-plugin-page-report.c#L1154) doesn't chain up the parent's init function so the `GncPluginPageReport*` doesn't get added to the `gnc_gobject_tracking` hash table.

The good news is that the `GncPluginReportPage` ownership is passed to `GncMainWindow` and is [destroyed via `g_object_unref`](https://github.com/Gnucash/gnucash/blob/maint/gnucash/gnome-utils/gnc-main-window.c#L3330) in `gnc_main_window_close_page` so `g_object_weak_ref` will work.

christopherlam:
This works

christopherlam:
There's no way to avoid having to define `equal_fn` is there?

Bob-IT:
I do not think so, but I would change "window-report" to WINDOW_REPORT_CM_CLASS

christopherlam:
> I do not think so, but I would change "window-report" to `WINDOW_REPORT_CM_CLASS`

That's an idea except `WINDOW_REPORT_CM_CLASS` isn't actually exported by any `.h` file.

Bob-IT:
It looks like all files that require "window-reprt" depend on gnc-plugin-page-report.h so I would move the define there and remove the one in gnc-plugin-page-report.c and window-report.c unless I have missed something.

PR #1009, opened 2021-05-20
Fix for opening balance account support

rhabacker:
@jralls wrote at #820
>In gnc_account_opening_balance_button_update the account-type test should be on aw->type rather than xaccAccountGetType so that it reflects the state of the dialog rather than the account type before the dialog started.

Fix for #762


rhabacker:
I just updated the patch with the correct author name.

jralls:
Thanks!

PR #1010, opened 2021-05-22
Bug 798195 - No option in the user interface to set the opening balancece of an existing account

rhabacker:
Previously, opening balances could only be entered through the hierarchy
assistant or when creating a new account. This excludes the entry of opening
balances for accounts that were already included in the chart of accounts
when the file was created, but were not used. In these cases, it was
necessary to manually create an entry in the corresponding account.

By enabling the opening balance page when editing accounts when there are
no transactions, the same procedure as for new creation is enabled.


jralls:
Rebased to correct the commit message. It's always been possible to create an opening balance transaction by hand in the register or transfer dialog, so the bug was invalid and the original commit message incorrect.

Thanks.

PR #1011, opened 2021-05-22
Provide a selection option for the opening balance account in the hierarchy assistant

rhabacker:
When merging multiple account templates, it may happen that multiple
opening balance accounts are possible. With the newly added Opening
balance accounts display, users can now select the desired account.

If multiple opening balance accounts are selected when completing
the assistant, it cannot be completed, only canceled.

See #820


jralls:
This is inconsistent with the code in dialog-account.c's `commodity_changed_cb`, which doesn't permit creating a second opening balance account in a currency. Is there a legitimate use-case for having more than one opening balance account in a currency?

rhabacker:
> This is inconsistent with the code in dialog-account.c's `commodity_changed_cb`, which doesn't permit creating a second opening balance account in a currency. 

From inspecting the Hierarchy Wizard, I would say that it basically supports the creation of a Gnucash file with only one currency. Additional opening balance sheet accounts are possible with multiple selection of account trees in the hierarchy wizard, which is caught by this pr. When opening such a generated file in gnucash for the first time, only one opening balance account is available.

When editing the generated file with gnucash, additional opening balance accounts (each with a different currency) are generated in the following use cases:
1. adding a new account with a currency other than the file currency and entering an opening balance.
2. editing an account that does not contain splits: Selecting another currency and entering an opening balance.

Once an opening balance account is created for one currency, I cannot create a second opening balance account with the same currency.  Can you show me how to reproduce the problem you mentioned ?

> Is there a legitimate use case for having more than one opening balance account in one currency?
no.

rhabacker:
> Can you show me how to reproduce the problem you mentioned ?

I just noticed that the account dialog is not modal as I expected, but can be accessed multiple times for an account. So you can edit an account twice. In the first dialog define it as Opening Balance Account and save, in the second dialog change the account name, also define it as Opening Balance Account and save. Then you have two opening balance accounts. Is this what you mean, or something else ?

jralls:
Umm, you seem to have confused yourself. From your description:

>When merging multiple account templates, it may happen that multiple
opening balance accounts are possible. With the newly added Opening
balance accounts display, users can now select the desired account.

And this PR adds a dialog box to select an opening balance account when there is more than one in the same currency.

That means *you* think that having more than one opening balance accounts in a currency is acceptable. I don't, but I'm willing to reconsider if you can propose a use case.

rhabacker:
> When merging multiple account templates, it may happen that multiple
> opening balance accounts are possible. With the newly added Opening
> balance accounts display, users can now select the desired account.

Thanks for that comment - I'll rewrite the message to more clearly emphasize the fact that there can only be one account.

jralls:
OK, I lost the thread on this, went back and read everything from last August on. I guess this is to implement @fellen's [comment on the original bug](https://bugs.gnucash.org/show_bug.cgi?id=797836#c12) "In a later phase one could think about smarter ways like an additional GUI element in the new file assistant to select the right opening balance account from a set.". That was premised on some arcana of German accounting that I don't think fits in GnuCash's target audience. Aside from that it was premised on the idea that one could have more than one opening balance account in some currency and so need to select among them when creating an account hierarchy.

Since GnuCash *doesn't* allow more than one opening balance per currency, what's the point of a selector column in the account hierarchy assistant?

rhabacker:
The gnucash hierarchy wizard supports the selection of accounts from several predefined template files. Account template files can use a different name for the opening balance account. Selecting more than one template can result in multiple opening balance accounts, which requires letting the user choose which opening balance account to use.  

With the set of available account templates that I can select by executing 
```
find -name '*.gnucash-xea' -exec gawk '$0 ~ /act:name/ { n=$0; } $0 ~ /opening balance/ { b=n; } END { if (b != "") print FILENAME b; }' {} | sort | uniq > various-opening-balances.log
```
I have obtained [different-opening-balances.log](https://github.com/Gnucash/gnucash/files/6544209/different-opening-balances.log)

the following differences are visible:

./nl/acctchrt_checkbook.gnucash-xea <act:name>Beginsaldi</act:name>
./nl/acctchrt_full.gnucash-xea <act:name>Openingsbalans</act:name>

-> spelling issue 

./de_DE/acctchrt_full.gnucash-xea <act:name>Anfangsbestand</act:name>

-> template for private use

./de_DE/acctchrt_skr03.gnucash-xea <act:name>9000 Saldenvortrag Sachkonten</act:name>
./de_DE/acctchrt_skr04.gnucash-xea <act:name>Saldenvorträge Sachkonten</act:name>
./de_DE/acctchrt_skr49.gnucash-xea <act:name>9000 Saldenvorträge Sachkonten</act:name>

-> templates for business purposes

There may be more chances to get multiple opening balances when using account templates from different countries.

jralls:
That problem is created by #820.  It breaks GnuCash to accommodate KMyMoney, and that is completely unacceptable.

rhabacker:
> It breaks GnuCash to accommodate KMyMoney, and that is completely unacceptabl

I do not understand what kmymoney has to do with this. 

Please correct me if I'm wrong, but surely there was a decision by the gnucash developers to use slots to mark opening balance accounts to fix a long outstanding bug. It follows that when combining multiple gnucash account templates, each having a (different) opening balance account predefined in the related xml file, there may subsequently be multiple opening balance accounts in the combined account template, which must be resolved somehow. The approach with this pr is to leave the decision to the user.

If that is not desired, I see another option is to not even allow the user to select multiple account templates with different Opening Balance Account in the Hierarchy Wizard.  This ensures that there is always only one opening balance account.

Or do you see another possibility ?

jralls:
> I do not understand what kmymoney has to do with this.

This got started because you wrote in [Bug 797836](https://bugs.gnucash.org/show_bug.cgi?id=797836)
> KMyMoney, which uses the templates from gnucash, got support for specifing a dedicated "opening balance" account, which is used on creating new accounts...It would be nice if the gnucash xea files also had support for specifying such a flag

> Please correct me if I'm wrong, but surely there was a decision by the gnucash developers to use slots to mark opening balance accounts to fix a long outstanding bug. It follows that when combining multiple gnucash account templates, each having a (different) opening balance account predefined in the related xml file, there may subsequently be multiple opening balance accounts in the combined account template, which must be resolved somehow. The approach with this pr is to leave the decision to the user.

It follows that it's a possibility, it does not follow that it's permitted. I think I've been pretty consistent in saying that it's not permitted. I admit that I didn't review the changes to the template accounts in #762 to ensure that all of the templates in a particular currency were consistent.

Of course the new opening-balance KVPs are created by the templates but are ignored until #820 is merged.

>  I see another option is to not even allow the user to select multiple account templates with different Opening Balance Account in the Hierarchy Wizard. This ensures that there is always only one opening balance account.

> Or do you see another possibility ?

Generally speaking there are several ways to get there. Preventing combining templates (or adding them later via Actions>New Account Hierarchy) with conflicting opening balance accounts would require scanning the XML from the template files to see if there are conflicts. One could also catch it during account creation using the existing code you wrote for the New and Edit Account dialogs. At that point one could either replace the old one, ignore the new one, or let the user choose which one to designate as the opening balance account. In all three cases there are more choices about whether to delete the account that isn't the opening balance account, whether it matters if the old one already has splits, and if not what to do with those splits if the account is deleted.

FWIW the current behavior with #820 merged is that the code picks the first account with the `opening-balance` flag that it encounters when traversing the hierarchy.

Another wrinkle is that if two templates have an account with the same name and parent the second one encountered apparently overwrites the first. Thus using both the Allgemein Kontonen and the SKR04 templates in German results in an Anfangbestand placeholder account with its 3 children; only the Saldenvorträge Sachkonten child has the `opening-balance` slot set.

Aside from the translation error in data/accounts/nl/accrtchrt_full.gnucash-xea (that should just be fixed) this appears from your analysis to be limited to the German templates, though a hypothetical user could inflict it upon themselves by combining say a French and Spanish template because that would try to create two opening balance accounts with different names. In *that* case any of the account creation options is satisfactory to me. So the driver here is the German case. What makes sense there? Is it reasonable for a user to combine any of those 4 with another or are the SKR.* templates supposed to be used apart from any other template? If the latter is it worthwhile to do something besides telling a user who tries to combine them to not be stupid when he complains that GnuCash picked the wrong opening balance account from the choices? Consider that without #820 GnuCash won't use the right account even when an SKR.* template is used by itself and nobody has complained in the 14 years since those templates were created.

jralls:
In the end I think this isn't necessary. Selecting all three NL templates still creates only one opening balance account, leaving the German case, where selecting all of the templates yields only two, the ones for SKR03 and SKR04. The SKR49 one is broken because its Saldervorträge Sachkonten is of type Asset. IIUC nobody would ever more than one of them anyway, [it's an either-or choice](https://www.datev.de/web/de/m/ueber-datev/datev-im-web/datev-von-a-z/skr-standard-kontenrahmen/). GnuCash does the right thing in all but that case and ends up with the SKR03 one if someone messes up and does try to use more than one in the hierarchy assistant.

rhabacker:
> In the end I think this isn't necessary

Thanks for reviewing this pull request. 

> The SKR49 one is broken because its Saldervorträge Sachkonten is of type Asset.

I filed an associated [bug report](https://bugs.gnucash.org/show_bug.cgi?id=798732).

PR #1014, opened 2021-05-24
GNCAmountEdit changes, followup to PR964

Bob-IT:
These commits allow filtering of control characters and currency symbols when pasting in to a GNCAmountEdit widget and also the display of a warning triangle when the verification of the entry fails along with tool tip for the error.

To display the warning triangle, the GNCAmountEdit parent class is changed from GtkEntry to a GtkBox so that I can pack that with an entry and image widget.

Following that, the commits are mainly to do with accommodating the change of widget and use in various source files but also added some verification of the entries used as some dialogs were not testing for a valid entry before proceeding.

### DIFF
--- a/gnucash/gnome-utils/gnc-amount-edit.c
+++ b/gnucash/gnome-utils/gnc-amount-edit.c
@@ -35,6 +35,7 @@
 #include "gnc-ui-util.h"
 #include "qof.h"
 #include "dialog-utils.h"
+#include "gnc-ui.h"
 
 #ifdef G_OS_WIN32
 # include <gdk/gdkwin32.h>
@@ -53,6 +54,8 @@ static void gnc_amount_edit_init (GNCAmountEdit *gae);
 static void gnc_amount_edit_class_init (GNCAmountEditClass *klass);
 static void gnc_amount_edit_changed (GtkEditable *gae,
                                      gpointer user_data);
+static void gnc_amount_edit_paste_clipboard (GNCAmountEdit *gae,
+                                             gpointer user_data);
 static gint gnc_amount_edit_key_press (GtkWidget   *widget,
                                        GdkEventKey *event);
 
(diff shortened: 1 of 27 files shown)


christopherlam:
A screenshot always useful; shouldn't the warning be more visible? Font or background color of `gnc_amount_edit` cell changed to red, or tooltip on warning triangle, etc?

image

Bob-IT:
I thought the warning triangle would be enough and visible with a tool tip. I did not want to play with colours as we have no control of what the user has set for each dialog.
Interesting on all other dialogs there is a tool tip but the transfer one does not display it.

Bob-IT:
OK I found out why it was not working on amount, there was a spurious label widget in the glade file that I think was over lapping, removed.

jralls:
@Bob-IT You used the same branch for both PRs so this is identical to #964. You may as well close this one.

gnucash/gnome-utils/gnc-amount-edit.c
```diff
@@ -257,19 +257,67 @@ gnc_amount_edit_new (void)
...
+         {
+             found = TRUE;
+             break;
+         }
+    }
+
+    if (found)
+       ret = error_pos + (x * g_utf8_strlen (symbol, -1));
+    else
+       ret = g_utf8_strlen (string, -1) - 1;
```
jralls, line 369: How can `error_len` be larger than the sum of the lengths of `split_text`'s pieces?

gnucash/gnome-utils/gnc-amount-edit.c
```diff
@@ -120,22 +123,87 @@ gnc_amount_edit_init (GNCAmountEdit *gae)
...
+
+        gtk_editable_set_position (GTK_EDITABLE(gae), position);
+
+        g_signal_stop_emission_by_name (G_OBJECT(gae), "paste-clipboard");
+
+        g_free (text);
+        g_free (existing);
+        g_free (filtered_text);
+    }
+}
```
jralls, line 258: Argh, another copy of this code. Since this isn't in register the single function should be in app-utils/gnc-ui-utils.c with `gnc_filter_text*`.

jralls (changes requested):

gnucash/gnome-utils/gnc-amount-edit.h
```diff
@@ -143,7 +145,8 @@ double gnc_amount_edit_get_damount (GNCAmountEdit *gae);
  * Return *  0 if the parsing was successful (note that if !empty_ok,
  *             an empty field will parse to 0)
  *        * -1 if the field is empty and that's ok (empty_ok)
- *        * error position if there was a parsing error
+ *        * error position if there was a parsing error which is offset
+ *          by ERROR_POSITION_BASE
```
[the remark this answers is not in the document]
Bob-IT: Will look into that.

gnucash/gnome-utils/gnc-amount-edit.c
Bob-IT: I need to do some testing, I added this for a reason when I started but maybe I have changed a part and not checked if is needed now.

Bob-IT:
> @Bob-IT You used the same branch for both PRs so this is identical to #964. You may as well close this one.

I know what I did, I created a branch from the branch for PR #964 as it relied on parts of that branch, did not notice that this one had all of the other branch commits.

Bob-IT:
Closing with all modified code now added to PR #964

PR #1019, opened 2021-05-30
various memory leaks plugged

christopherlam:
~instead of `qof_print_date` string which must be freed.~

### DIFF
--- a/libgnucash/engine/Transaction.c
+++ b/libgnucash/engine/Transaction.c
@@ -2001,6 +2001,7 @@ xaccTransSetDatePostedGDate (Transaction *trans, GDate date)
     g_value_init (&v, G_TYPE_DATE);
     g_value_set_boxed (&v, &date);
     qof_instance_set_kvp (QOF_INSTANCE(trans), &v, 1, TRANS_DATE_POSTED);
+    g_value_unset (&v);
     /* mark dirty and commit handled by SetDateInternal */
     xaccTransSetDateInternal(trans, &trans->date_posted,
                              gdate_to_time64(date));
@@ -2060,6 +2061,7 @@ xaccTransSetDateDue (Transaction * trans, time64 time)
     xaccTransBeginEdit(trans);
     qof_instance_set_kvp (QOF_INSTANCE (trans), &v, 1, TRANS_DATE_DUE_KVP);
     qof_instance_set_dirty(QOF_INSTANCE(trans));
+    g_value_unset (&v);
     xaccTransCommitEdit(trans);
 }
 
@@ -2074,6 +2076,7 @@ xaccTransSetTxnType (Transaction *trans, char type)
     xaccTransBeginEdit(trans);
     qof_instance_set_kvp (QOF_INSTANCE (trans), &v, 1, TRANS_TXN_TYPE_KVP);
     qof_instance_set_dirty(QOF_INSTANCE(trans));
+    g_value_unset (&v);
     xaccTransCommitEdit(trans);
 }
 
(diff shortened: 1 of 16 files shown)


christopherlam:
This would leak numerous datebuff strings, 1 per register date-cell.

jralls:
@christopherlam Looks good.

christopherlam:
https://stackoverflow.com/questions/52523831/gvalue-initialization-finalization-necessity

christopherlam:
(see answer by philip withnall, the gnome dev IIRC)

libgnucash/engine/Transaction.c
```diff
@@ -2472,15 +2481,18 @@ xaccTransGetTxnType (const Transaction *trans)
...
     if (!trans) return TXN_TYPE_NONE;
     qof_instance_get_kvp (QOF_INSTANCE (trans), &v, 1, TRANS_TXN_TYPE_KVP);
     if (G_VALUE_HOLDS_STRING (&v))
+    {
          s = g_value_get_string (&v);
-    if (s && strlen (s) == 1)
-        return *s;
-
-    return TXN_TYPE_NONE;
+         if (s && (s[0] != '\0') && (s[1] == '\0'))
```
gjanssens, line 2491: Why this change ?

christopherlam: Small micro-optimisation. `strlen` is a function call and is O(N), whereas this translates to 6 assembly instructions...

gjanssens: Fair enough.

On the other hand the original code expresses the intent much more clearly where your version requires mental decoding to figure out we are trying the first character of a one-byte string, only if the string is exactly one byte. So unless you have measurable performance gains from this I would favour the more expressive form.

Just a nitpick though.

gjanssens: Indeed.

We're not in C++ land yet, but would there be a C++ alternative not using macros ? I hear a lot of backpush against macros in C++ conversations.

libgnucash/engine/Transaction.c
```diff
@@ -2472,15 +2481,18 @@ xaccTransGetTxnType (const Transaction *trans)
...
+             ret = s[0];
+    }
+    g_value_unset (&v);
```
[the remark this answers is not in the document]
gjanssens: I *think*  'g_value_take_string' works the other way around. It would be used to set a GValue with the given string such that it takes ownership of it. It's not a method to extract the string (a glaring omission in the GValue specification in retrospect).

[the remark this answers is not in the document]
gjanssens: https://developer.gnome.org/gobject/stable/gobject-Standard-Parameter-and-Value-Types.html#g-value-take-string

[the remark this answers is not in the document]
jralls: Darn, you're right.

libgnucash/engine/Transaction.c
jralls: Sort of, and it works in C too:
```
static inline char 
str_one_char (const char* s) 
{
   return (s && s[0] && !s[1]) ? *s : '\0'; 
}
```
like a macro the *definition* has to be part of the compilation unit (so it's either in the source file being compiled or in a header) or it can't be inlined. Also `inline` is a suggestion, the compiler doesn't have to inline it and probably won't if told `-O0` and if it doesn't inline it then the string has to be pretty long before it improves on `strlen`.

christopherlam: let's resolve this. `strlen` it is then.

libgnucash/engine/qof-backend.cpp
```diff
@@ -100,8 +100,9 @@ QofBackend::register_backend(const char* directory, const char* module_name)
     }
 
     auto absdir = directory;
+    auto pkgdir = gnc_path_get_pkglibdir ();
     if (!absdir || !g_path_is_absolute(absdir))
-        absdir = gnc_path_get_pkglibdir ();
+        absdir = pkgdir;
```
jralls, line 105: Why?

PR #1021, opened 2021-06-01
Add for context for translatable text at dialog-price.

brianhsu:
The context of Ask / Last in the price dialog is pretty different from plan English, and it might cause trouble for translators. 

For example, in zh-tw，the `Last` should be translate to `成交價` in the price context. But in the context of report options,it should be translated to `最後`, apparently it's very different.

This pull request added a context to the price dialog drop down menu.


brianhsu:
Oops, I didn't noticed that the latest code base already use 'Last budget period` instead of just `Last` in the budget report options. This PR is not needed.

fellen:
I think, it is a good idea, to add more msgcontext. Give me some time to think about it.

brianhsu:
Thanks, another word that I wish could be distinguished from each context is the word `Rate`, as following:

```
  #: gnucash/gnome/dialog-job.c:577 gnucash/gnome/dialog-job.c:590
  #: gnucash/gnome-utils/gnc-tree-view-split-reg.c:2929
  #: gnucash/gtkbuilder/dialog-job.glade:204
  #: gnucash/register/ledger-core/split-register-model.c:369
  msgid "Rate"

  #: gnucash/gnome-utils/gnc-tree-view-split-reg.c:3304
  #, fuzzy
  msgid "Enter the rate"                                                                                 
  msgstr "利率"
```

The issue is that rate could mean tax rate, exchange rate and something shows in the "Add job" dialog (which I have no idea what's the rate for). 

We never use the single word "rate" in our language, because it's to ambiguous. Without any context, we only know that it's some number divided by some number, and it's meaningless in our language without specify what's the rate for. (ex. exchange rate).

gnucash/gtkbuilder/dialog-price.glade
```diff
@@ -9,16 +9,16 @@
     </columns>
     <data>
       <row>
-        <col id="0" translatable="yes">Bid</col>
+        <col id="0" translatable="yes" context="Price">Bid</col>
```
fellen, line 12: We have started to distinguish a "Quote" –offered at some market– and a real "Price" –paid for an Item. So better use context="Quote" .

I will have to look, if the whole dialog should be renamed.

fellen:
> Thanks, another word that I wish could be distinguished from each context is the word `Rate`, as following:
:

That would be much harder to improve.
Rate is one english word derived from latin:ratio. Example "I still have a monitor with 3:4 ratio."
There a many places where you can either enter a quantified relation "rate" or an absolute "amount"/"value".

How was it solved for zh_CN?

jralls:
> That would be much harder to improve.

No, it's easy:  [Add context](https://www.gnu.org/software/gettext/manual/html_node/Contexts.html) to each one. I'd think it good practice to add context to *every* one word msgid and consider it for every two and three word one.

brianhsu:
> > Thanks, another word that I wish could be distinguished from each context is the word `Rate`, as following:
> > :
> 
> That would be much harder to improve.
> Rate is one english word derived from latin:ratio. Example "I still have a monitor with 3:4 ratio."
> There a many places where you can either enter a quantified relation "rate" or an absolute "amount"/"value".
> 
> How was it solved for zh_CN?

I don't think it was solved in zh_CN.

For zh_CN, they just translate it to '比率' (the equivalent of "rate")  which has the same problem I mentioned before. Although it correct word-to-word translation, it's very unnatural and ambiguous in our language. 

We almost never use that word without any prefix to specify which kind of rate we refer to in accounting. In accounting, we only use exact term like "tax rate" / "interest rate". 

Also I agree with @jralls , add context to each of them will solve this problem. The side effect will be that there would be duplicate entry for language that does not need to separate them.

brianhsu:
I've tried to add context to the word that would cause confuse in zh_TW, but also minimize the number of context that is needed to prevent too much duplicate entries for translators of other languages.

gnucash/gnome-utils/gnc-tree-view-split-reg.c
```diff
@@ -2926,7 +2926,7 @@ gtv_sr_titles (GncTreeViewSplitReg *view, RowDepth depth)
             {
             default: //FIXME These if statements may not be required
                 if (depth == TRANS1 || depth == TRANS2 || depth == SPLIT3)
-                    gtk_tree_view_column_set_title (tvc, _("Rate"));
+                    gtk_tree_view_column_set_title (tvc, C_("In register", "Rate"));
```
jralls, line 2929: This column is hidden so its heading shouldn't be translated.

brianhsu: Then it shouldn't have be `_("Rate")` either? 

If I understand how gettext works correctly, in the original code, this heading is also got translated to the string inside .po files.

jralls: > Then it shouldn't be _("Rate") neither?

Right, just `"Rate"`, maybe with a comment noting that it's a hidden column.

gnucash/gtkbuilder/dialog-price.glade
```diff
@@ -9,19 +9,19 @@
...
-        <col id="0" translatable="yes">Last</col>
+        <col id="0" translatable="yes" context="In commoditiy quote dialog">Last</col>
       </row>
       <row>
-        <col id="0" translatable="yes">Net Asset Value</col>
+        <col id="0" translatable="yes" context="In commoditiy quote dialog">Net Asset Value</col>
       </row>
       <row>
-        <col id="0" translatable="yes">Unknown</col>
+        <col id="0" translatable="yes" context="In commoditiy quote dialog">Unknown</col>
```
jralls, line 24: "In commodity quote dialog" isn't terribly helpful, the translator still has to go figure out what that dialog is (In spite of @fellen's protestation it's actually called the **Price Editor** dialog. Also in spite of his protestation `Bid` and `Ask` are types of quotes--bid if you want to sell, ask if you want to buy--but `Last` and `Net Asset Value` are prices--the former the price of the last trade on an exchange prior to the time of the price, the latter the price of an open-end mutual fund calculated after the market close each day and for which both buys and sells of that fund from the fund company are priced the following trading day.

That explanation is a bit wordy for translation contexts, so maybe `Commodity quote type` for `Bid` and `Ask`, `Commodity price type` for `Last` and `Unknown`, and `Open ended mutual fund price` for `NetAsset Value`.

gnucash/gtkbuilder/gnc-plugin-page-budget.glade
```diff
@@ -75,7 +75,7 @@
                 <property name="visible">True</property>
                 <property name="can_focus">False</property>
                 <property name="halign">end</property>
-                <property name="label" translatable="yes">Value</property>
+                <property name="label" translatable="yes" context="In edit all budget peroids dialog">Value</property>
```
jralls, line 78: s/peroids/periods/, but again saying where it goes isn't really helpful. Say what it means: "Budgeted value for account in all periods".

gnucash/register/ledger-core/split-register-model.c
```diff
@@ -366,7 +366,7 @@ static const char*
 gnc_split_register_get_rate_label (VirtualLocation virt_loc,
                                    gpointer user_data)
 {
-    return _ ("Rate");
+    return C_("In register", "Rate");
```
jralls, line 369: Also hidden like the one in gnc-tree-view-split-reg.c.

gnucash/gtkbuilder/dialog-price.glade
fellen: Yes, this context is totally redundant. The location[s] is/are already in the file before the msgid. This info is also displayed in weblate's interface in the lower right corner.

@jralls My definition

- Price: amount they user self (got) paid per piece.
- Quote: Offers the user got aware of. Often functions (Open/Close, Hi/Lo, (current) Ask/Bid …) of the sets of prizes paid by others. Also the NAV is the offer of the emitter to buy the funds back on the next working day.

But I agree with @brianhsu that the english language lacks precission in many use cases. ;-)

Can we delay this thread behind the next release?

jralls: @fellen, in English "quote" is actually short for "quote a price", where "quote" is a somewhat antique usage for "speak". It's not used for prices of transactions in the past, regardless of who transacted them.

> My definition

"When I use a word," [Humpty Dumpty](https://en.wikipedia.org/wiki/Humpty_Dumpty) said, in rather a scornful tone, "it means just what I choose it to mean—neither more nor less."

> Can we delay this thread behind the next release?

Yes.

brianhsu: > Yes, this context is totally redundant. The location[s] is/are already in the file before the msgid. This info is also displayed in weblate's interface in the lower right corner.
> 
> @jralls My definition
> 
>     * Price: amount they user self (got) paid per piece.
> 
>     * Quote: Offers the user got aware of. Often functions (Open/Close, Hi/Lo, (current) Ask/Bid …) of the sets of prizes paid by others. Also the NAV is the offer of the emitter to buy the funds back on the next working day.
> 
> 
> But I agree with @brianhsu that the english language lacks precission in many use cases. ;-)
> 
> Can we delay this thread behind the next release?

I totally agree it's redundant in current code base.  This PR is only useful in the older version, because some report use "Last" in their option drop-down menu to indicate time instead of type of price. And they should translate to quite different word in Chinese. (`成交價` vs `最後的`)

Fortunately, the latest code base does not have this issue, because "Ask" / "Last" does not been used in other place. 

Also if the "Rate" column inside gnc-tree-view-split-reg.c  / split-register-model.c is a hidden column and should never displayed to users, then I have no problem that it remain current status and has no context. 

Actually after all of these discussion and clarification,  I don't think this PR is actually needed and totally fine if this PR get closed without further action . Because all of these word that actually displayed to users does not have ambiguity in Chinese anymore.

The only two thing I'm not sure is that:

1. What the "Rate" in the job dialog actually means. Does it mean tax rate or exchange rate or interest rate or something else? This really impact the quality of Chinese translation as I mentioned before. And I couldn't find explanation of this field in the GnuCash manual on the website or the help documentation in the program.
2. Does the "Rate" in the gnc-tree-view-split-reg.c  / split-register-model.c refer to the account register column? I think it is but I'm not sure because I actually have a account register that display this column as the following screenshot, but according to above discussion it should be a hidden column. If this is just a bug and user should never see this column, I'm totally agree that "Rate" don't need any context.

Screenshot

jralls: > Fortunately, the latest code base does not have this issue, because "Ask" / "Last" does not been used in other place.

It doesn't matter, it still requires the translator to go trolling around in the code and hope to be able to figure out the meaning. Like I said earlier, *all* single-word msgids should have context so that translators don't need to do that.

> What the "Rate" in the job dialog actually means?

That sadly took a bit of digging to find out. It's from a [drive-by contribution](https://bugs.gnucash.org/show_bug.cgi?id=728136) 7 years ago and isn't documented anywhere. The answer is that it's the unit price of whatever the job is associated with: It could be an hourly or a piece rate for labor or a material unit price.

> Does the "Rate" in the gnc-tree-view-split-reg.c / split-register-model.c refer to the account register column? 

No. It's the same as the price column in a non-currency register, that is `value / amount`. It's used for convenience by the balancing code and is set by the Transfer Dialog's exchange rate section.

jralls:
@brianhsu wrote on 21 June 2021
> Actually after all of these discussion and clarification, I don't think this PR is actually needed and totally fine if this PR get closed without further action . Because all of these word that actually displayed to users does not have ambiguity in Chinese anymore.

PR #1027, opened 2021-06-07
Cleanup and update zh_TW.po to match current codebase.

brianhsu:
This translation file hasn't been maintained for years. It contains lots of useless / misleading comments or fuzzy entry left by msgmerge, and lots of file reference does not match the latest code base.

This makes it hard to translate both on Weblate, or use desktop program like PoEdit after download the po file from Weblate. 

I've updated and merge it with latest .pot template generated from latest source code in `maint` branch.

I've also tried upload it to Weblate directly, but apparently it won't delete those garbage comment and won't add msgid that didn't exist in the current po file.

### DIFF
--- a/po/zh_TW.po
+++ b/po/zh_TW.po
@@ -7,10 +7,10 @@
 msgid ""
 msgstr ""
 "Project-Id-Version: GnuCash-2.4.3\n"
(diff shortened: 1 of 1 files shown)


fellen:
Thanks!

https://hosted.weblate.org/projects/gnucash/gnucash/zh_Hant/ gots updated, too.

PR #1033, opened 2021-06-13
remove "..." from strings.

YtxCash:
I am working on translation from English to Chinese, Weblate.

There are lots of "..." in menu and somewhere, I do not know why contributors add them.

In my feeling, they are not so necessary. 

After remove them, the software looks more clear, so I remove almost all "..." from strings. Only in strings.

### DIFF
--- a/gnucash/gnome-utils/assistant-xml-encoding.c
+++ b/gnucash/gnome-utils/assistant-xml-encoding.c
@@ -1091,7 +1091,7 @@ gxi_parse_file (GncXmlImportData *data)
     }
 
     xaccLogDisable ();
-    gxi_update_progress_bar (_("Reading file..."), 0.0);
+    gxi_update_progress_bar (_("Reading file"), 0.0);
     qof_session_load (session, gxi_update_progress_bar);
     gxi_update_progress_bar (NULL, -1.0);
     xaccLogEnable ();
@@ -1114,7 +1114,7 @@ gxi_parse_file (GncXmlImportData *data)
     book = qof_session_get_book (session);
     backend = qof_book_get_backend (book);
 
-    gxi_update_progress_bar (_("Parsing file..."), 0.0);
+    gxi_update_progress_bar (_("Parsing file"), 0.0);
     success = gnc_xml2_parse_with_subst (backend, book, data->subst);
     gxi_update_progress_bar (NULL, -1.0);
 
@@ -1146,7 +1146,7 @@ gxi_save_file (GncXmlImportData *data)
     QofBackendError io_err;
     g_return_val_if_fail (data && data->session, FALSE);
 
-    gxi_update_progress_bar (_("Writing file..."), 0.0);
+    gxi_update_progress_bar (_("Writing file"), 0.0);
     qof_session_save (data->session, gxi_update_progress_bar);
     gxi_update_progress_bar (NULL, -1.0);
 
--- a/gnucash/gnome-utils/dialog-file-access.c
+++ b/gnucash/gnome-utils/dialog-file-access.c
@@ -300,14 +300,14 @@ gnc_ui_file_access (GtkWindow *parent, int type)
     switch ( type )
     {
     case FILE_ACCESS_OPEN:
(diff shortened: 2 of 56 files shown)


christopherlam:
It would seem that HIGs tend to support your view: ellipses only to be used if "further input is necessary":

https://developer.gnome.org/hig/stable/writing-style.html.en#ellipses
https://developer.apple.com/design/human-interface-guidelines/macos/menus/menu-anatomy/
https://docs.microsoft.com/en-us/windows/win32/uxguide/ctrl-command-buttons (Design Concepts)

However, (my opinion only) this UI change is jarring, and will bring out a lot of opiniated views... (such as mine!)

fellen:
https://bugs.gnucash.org/show_bug.cgi?id=792975
Imagine a menu entry
1. `Delete files` will immediately remove them, but
2. `Delete files…` or
3. `Open Cleanup Tool` would open a dialog to specify the selection and/or ask for confirmation before execution.

Tasks:
1. Unification: we should replace 3 dots by one typographical horizontal ellipse. 
2. Verification: Bug Comment 12 lists a few places, where we should remove them..

Each should ideally be done in the sources and the po files in one PR.

gnucash/gnome-utils/assistant-xml-encoding.c
```diff
@@ -1091,7 +1091,7 @@ gxi_parse_file (GncXmlImportData *data)
     }
 
     xaccLogDisable ();
-    gxi_update_progress_bar (_("Reading file..."), 0.0);
```
fellen, line 1094: No, this are status messages. Here the Ellipses means:
Work in progress, be patient!

TianXing-Yi:
> https://developer.gnome.org/hig/stable/writing-style.html.en#ellipses
> https://developer.apple.com/design/human-interface-guidelines/macos/menus/menu-anatomy/
> https://docs.microsoft.com/en-us/windows/win32/uxguide/ctrl-command-buttons

> Proper use of ellipses is important to indicate that users can make further choices before performing the action, or even cancel the action entirely. The visual cue offered by an ellipsis allows users to explore your software without fear.

Documents are helpful.

But ellipses are less useful for user. Like me, use Gnucash since 2014, for company, for home. Until today I know "..." means opening a dialog.  In other words, it doesn't matter if I don't know.

In my option, ellipses are more usefull for developers. For user, ellipses mean warning will be better, like Print... Delete...

There operation will not directly change user data should not use "...", like Import **, Export **, Find, Filter.
Otherwise, software will make user feel a bit verbose. It's my feeling.

gnucash/gnome-utils/assistant-xml-encoding.c
TianXing-Yi: > No, this are status messages. Here the Ellipses means:
> Work in progress, be patient!
> [Edit:]
> Often further details or a progress bar follow.

Yeah , it's suitable here. I made a mistake.

TianXing-Yi:
@fellen Hi sir, Is there a way that remove "..." only from Chinese? 
Weblate checks match.  The changes cannot through it.

fellen:
> @fellen Hi sir, Is there a way that remove "..." only from Chinese?
> Weblate checks match. The changes cannot through it.

The docs about checks start with https://docs.weblate.org/en/latest/user/checks.html. I did not find a way.

I am a big fan of international standards. They make many things easier. Sometimes rulesets get improved by others like from the File Hierarchy Standard (FHS) to the XDG standards. But as a inrtnational comunity we should try to stay close to the standards

I think, we have a common target:
Remove the dots from the menu entries listed in https://bugs.gnucash.org/show_bug.cgi?id=792975#c12 and perhaps some which Ii did not see on the first glimpse. That is the point where you should start.

jralls:
@YtxCash We'll want to merge this in December as preparation for the 4.900 release in January. Can you rebase it on a current master and do whatever you need to in order to make it ready for review?

Bob-IT:
@jralls Can this be closed, looks like your changes replace these.

jralls:
> looks like your changes replace these.

Indeed they do, I just forgot to close this out.

PR #1034, opened 2021-06-14
Fix minor typo: choosen -> chosen

yurchor:
Thanks in advance for reviewing.


fellen (approved):
You are right, but as @jralls declared string freeze last night, we have to wait 2 weeks.

jralls:
Thanks!

PR #1035, opened 2021-06-14
Developer log

christopherlam:
Developer Log -- see `hello-world.scm` to add an entry into deprecation log.
As discussed in 4e38b685e27589d0d4346cd2d5caf5a762f97877

Pending:
1. nice icon - warning, or info 'i' icon
2. ideally this toolbar item should be hidden if the devlog is empty
3. textarea instead of `gnc_info_dialog`?

image

### DIFF
--- a/gnucash/gnome-utils/gnc-main-window.c
+++ b/gnucash/gnome-utils/gnc-main-window.c
@@ -157,6 +157,7 @@ static void gnc_main_window_cmd_page_setup (GtkAction *action, GncMainWindow *wi
 static void gnc_main_window_cmd_file_properties (GtkAction *action, GncMainWindow *window);
 static void gnc_main_window_cmd_file_close (GtkAction *action, GncMainWindow *window);
 static void gnc_main_window_cmd_file_quit (GtkAction *action, GncMainWindow *window);
+static void gnc_main_window_cmd_dev_log (GtkAction *action, GncMainWindow *window);
 static void gnc_main_window_cmd_edit_cut (GtkAction *action, GncMainWindow *window);
 static void gnc_main_window_cmd_edit_copy (GtkAction *action, GncMainWindow *window);
 static void gnc_main_window_cmd_edit_paste (GtkAction *action, GncMainWindow *window);
@@ -311,6 +312,11 @@ static GtkActionEntry gnc_menu_actions [] =
         "FileQuitAction", "application-exit", N_("_Quit"), "<primary>Q",
         N_("Quit this application"),
         G_CALLBACK (gnc_main_window_cmd_file_quit)
+    },
+    {
+        "DevLogAction", NULL, N_("_Developer Log"), NULL,
+        N_("Show developer log"),
+        G_CALLBACK (gnc_main_window_cmd_dev_log)
     },
 
     /* Edit menu */
@@ -4354,6 +4360,23 @@ gnc_main_window_cmd_file_close (GtkAction *action, GncMainWindow *window)
     page = priv->current_page;
     gnc_main_window_close_page(page);
 }
+
+static void
+gnc_main_window_cmd_dev_log (GtkAction *action, GncMainWindow *window)
+{
+    gchar *logstr;
+    if (gnc_developer_log_empty ())
+    {
+        PWARN ("devlog empty. this button should be disabled.");
+        return;
+    }
+
+    logstr = gnc_get_developer_log ();
+    if (gnc_verify_dialog (NULL, TRUE, "Dev Log\n\n%s\n\nYes to clear", logstr))
+        gnc_clear_developer_log ();
+    g_free (logstr);
+}
+
 
 static void
 gnc_main_window_cmd_file_quit (GtkAction *action, GncMainWindow *window)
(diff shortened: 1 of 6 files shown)


gnucash/gnome-utils/gnc-main-window.c
```diff
@@ -312,6 +313,11 @@ static GtkActionEntry gnc_menu_actions [] =
         N_("Quit this application"),
         G_CALLBACK (gnc_main_window_cmd_file_quit)
     },
+    {
+        "DevLogAction", NULL, N_("_Developer Log"), NULL,
+        N_("Show developer log"),
```
fellen, line 318: String Freeze!

fellen (commented):
Can you add to the PR intro a few more detaiis?

gnucash/gnome-utils/gnc-main-window.c
jralls: Relax, @fellen, this is still preliminary. It won't be done in time for 4.6.

Bob-IT:
Some thoughts, have not tried this yet
Is this just for reports?
If so, the toolbar additions could be done in gnc-plugin-page-report.c
I assume the use of a depreciated function adds an entry to the trace file and so do these changes fire a dialog also so they will be displayed?
If so, do we need the toolbar entry at all?
Using a GList seems rather complicated, why not just use a static char* instead and just concat the messages, you could do a search to see if the message has already been added first.
Possibly also you could have a static boolean to indicate that the warning has been displayed in that session and only display again if a different warning is to be added. 
Maybe also the first line could be to mention the warning has been logged to the trace file.

jralls:
Maybe this is a bit over-done? I'd think it sufficient to just tell the user somehow that a deprecated function was used and they should inspect the trace file for the warnings. We need a wiki article explaining our deprecation policy anyway, the message could include a link to there.

christopherlam:
```scheme

(define gnc:issue-deprecation-warning
  (let ((strings-shown (make-hash-table)))
    (lambda (str)
      (issue-deprecation-warning str)
      (unless (hash-ref strings-shown str)
        (hash-set! strings-shown str #t)
        (gnc-info-dialog '() (format #f "Warning: ~a. Please check tracefile" str))))))
```

jralls:
That will pop a dialog for every use of every deprecated function. Not so good. Instead, set a flag when a deprecated function is encountered, finish processing the report, and pop the dialog if the flag is set.

christopherlam:
```scheme

;; render report. will return a 2-element list: either (list html #f)
;; where html is the report html string, or (list #f captured-error)
;; where captured-error is the error string.
(define (gnc:render-report report)
  (define old-deprecation-warning issue-deprecation-warning)
  (define (get-report) (gnc:report-render-html report #t))
  (define warnings-list '())
  (define (deprecation-warning-override str)
    (set! warnings-list (cons str warnings-list))
    (issue-deprecation-warning str))
  (define (collect-warnings-list)
    (format #f "Note this report has deprecations:\n\n~a\n
Please review the Tracefile") (string-join warnings-list "\n"))
  (set! issue-deprecation-warning deprecation-warning-override)
  (gnc:apply-with-error-handling get-report '())
  (unless (null? warnings-list)
    (gnc-info-dialog '() (collect-warnings-list)))
  (set! issue-deprecation-warning old-deprecation-warning))
```

or (but still buggy):

```scheme

Unstaged changes (2)
modified   bindings/guile/core-utils.scm
@@ -77,9 +77,13 @@
 (define gnc:string-locale<? string-locale<?)
 (define gnc:string-locale>? string-locale>?)
 
+(define gnc:issue-deprecation-warning
+  (let ((warnings '()))
+    (match-lambda
+      ('reset (set! warnings '()))
+      ('get (string-join warnings "\n"))
+      ('empty? (null? warnings))
+      (str (issue-deprecation-warning str) (set! warnings (cons str warnings))))))
 
 ;; Custom unbound-variable exception printer: instead of generic "In
 ;; procedure module-lookup: Unbound variable: varname", it will first
modified   gnucash/report/report-core.scm
@@ -763,7 +763,11 @@ not found.")))
 ;; where html is the report html string, or (list #f captured-error)
 ;; where captured-error is the error string.
 (define (gnc:render-report report)
-  (define (get-report) (gnc:report-render-html report #t))
+  (define (get-report)
+    (let ((html (gnc:report-render-html report #t)))
+      (unless (gnc:issue-deprecation-warning 'empty?)
+        (gnc-info-dialog '() (gnc:issue-deprecation-warning 'get)))
+      (gnc:issue-deprecation-warning 'reset)
+      html))
   (gnc:apply-with-error-handling get-report '()))
 
 ;; looks up the report by id and renders it with gnc:report-render-html
```

christopherlam:
I still think it should be done in C/C++ and available generally. Thus we could add an entry with deprecated API functions, and would be shown when custom .scm reports call them.

```patch
modified   libgnucash/engine/Account.cpp
@@ -3946,7 +3946,7 @@ xaccAccountCountSplits (const Account *acc, gboolean include_children)
 {
     gint64 nr, i;
 
-    PWARN ("xaccAccountCountSplits is deprecated and will be removed \
+    gnc_add_developer_log_entry ("xaccAccountCountSplits is deprecated and will be removed \
 in GnuCash 5.0. If testing for an empty account, use \
 xaccAccountGetSplitList(account) == NULL instead. To test descendants \
 as well, use gnc_account_and_descendants_empty.");
```

PR #1041, opened 2021-06-16
Explain out what AsOfDate means in the doc

xuhdev:
As discussed at
https://lists.gnucash.org/pipermail/gnucash-user/2021-June/096893.html


jralls:
Thanks!

PR #1046, opened 2021-06-19
Bug798162 -  Type Ahead Initialization Problem

Bob-IT:
@jralls 
These are my changes for bug798162, the last commit is a small refactor to the gnc_combo_cell_modify_verify. I think it does make it flow better, not sure what else can be achieved.

Still think it fixes the problem with further testing but as always can you give it a spin to confirm.

### DIFF
--- a/gnucash/register/register-gnome/combocell-gnome.c
+++ b/gnucash/register/register-gnome/combocell-gnome.c
@@ -633,16 +633,22 @@ gnc_combo_cell_modify_verify (BasicCell* _cell,
             return;
         }
         match_str = quickfill_match (box->qf, newval);
+
+        if (match_str != NULL) // Do we have a quickfill match
+        {
+            *start_selection = newval_chars;
+            *end_selection = -1;
+            *cursor_position += change_chars;
+            box_str = match_str;
+
+            block_list_signals (cell); // Prevent recursion
+            gnc_item_list_select (box->item_list, match_str);
+            unblock_list_signals (cell);
+        }
     }
 
-    if (match_str != NULL)
-    {
-        *start_selection = newval_chars;
-        *end_selection = -1;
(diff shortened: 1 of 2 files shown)


gnucash/register/register-gnome/combocell-gnome.c
```diff
@@ -664,9 +671,11 @@ gnc_combo_cell_modify_verify (BasicCell* _cell,
             unblock_list_signals (cell);
         }
         gnc_basic_cell_set_value_internal (_cell, newval);
-        block_list_signals (cell);
+
+        block_list_signals (cell); //Prevent recursion
         gnc_item_list_select (box->item_list, NULL);
         unblock_list_signals (cell);
+
```
jralls, line 678: How about
```
    block_list_signals (cell); //Prevent recursion       
    if (cell-&gt;shared_store &amp;&amp; gnc_item_list_using_temp (box-&gt;item_list))
    {
          gnc_item_list_set_temp_store (box-&gt;item_list, NULL);
          gtk_list_store_clear (box-&gt;tmp_store);
     }
     gnc_item_list_select (box-&gt;item_list, NULL);
     unblock_list_signals (cell);
    gnc_basic_cell_set_value_internal (_cell, newval);
```
Or does `gnc_basic_cell_set_value_internal (_cell, newval);` need to happen in between clearing the list store and clearing the selection?

jralls:
Aside from the one question on the refactor, looks good.

gnucash/register/register-gnome/combocell-gnome.c
Bob-IT: Did not see that, that part is dealing with when no type-ahead/quickfill found.
Made the change locally and after testing could not see any problems so will add it to PR and push this tomorrow.

PR #1047, opened 2021-06-19
Fix German translation for transaction filter case sensitivity

mkrieger1:
It had the opposite meaning of what was intended.


fellen:
Gut beobachtet, danke!

PR #1056, opened 2021-06-26
Bug 753307 Available invoice reports must be dynamic

christopherlam:
The invoice reports and GUIDs must be retrieved dynamically rather than hard coded. Expose list of invoice reports, for use in populating the global preference/book property. WIP. @Bob-IT

The invoice list is exposed out-of-order, to be reordered in C (or guile?)


jralls:
See my comments on [the bug](https://bugs.gnucash.org/show_bug.cgi?id=753307).

PR #1059, opened 2021-06-28
Support displaying 9 share digits in the portfolio reports

nomis:
I've got some fractional shares to 8 decimal places.


jralls:
Thanks!

PR #1060, opened 2021-06-29
Fix misc. typos

luzpaz:


fellen:
Thanks, merged as commit 95f8884 because of rewording.

PR #1064, opened 2021-07-02
Remove "New _Invoice..." from toolbar

TianXing-Yi:
I do not know why put this button on toolbar. 

It works almost same as "New invoice", very useless.

So I found source code location through weblate, and removed them.

### DIFF
--- a/gnucash/gnome/gnc-plugin-business.c
+++ b/gnucash/gnome/gnc-plugin-business.c
@@ -314,13 +314,6 @@ static GtkActionEntry gnc_plugin_actions [] =
         G_CALLBACK (gnc_plugin_business_cmd_test_init_data)
     },
 
-    /* Toolbar */
-    {
-        "ToolbarNewInvoiceAction", GNC_ICON_INVOICE_NEW, N_("New _Invoice..."), NULL,
-        N_("Open the New Invoice dialog"),
-        G_CALLBACK (gnc_plugin_business_cmd_customer_new_invoice)
-    },
-
     /* Register popup menu */
     {
         "RegisterAssignPayment", NULL, N_("Assign as payment..."), NULL,
--- a/gnucash/gtkbuilder/business-prefs.glade
+++ b/gnucash/gtkbuilder/business-prefs.glade
@@ -63,24 +63,6 @@
           </packing>
         </child>
         <child>
-          <object class="GtkCheckButton" id="pref/dialogs.business.invoice/enable-toolbuttons">
-            <property name="label" translatable="yes">Enable extra _buttons</property>
-            <property name="visible">True</property>
-            <property name="can_focus">True</property>
-            <property name="receives_default">False</property>
-            <property name="has_tooltip">True</property>
-            <property name="tooltip_markup">If active, extra toolbar buttons for common business functions are shown as well. Otherwise they are not shown.</property>
-            <property name="tooltip_text" translatable="yes">If active, extra toolbar buttons for common business functions are shown as well. Otherwise they are not shown.</property>
-            <property name="halign">start</property>
-            <property name="use_underline">True</property>
-            <property name="draw_indicator">True</property>
-          </object>
-          <packing>
-            <property name="left_attach">0</property>
-            <property name="top_attach">1</property>
-          </packing>
-        </child>
-        <child>
           <object class="GtkCheckButton" id="pref/dialogs.business.invoice/use-new-window">
             <property name="label" translatable="yes">_Open in new window</property>
             <property name="visible">True</property>


jralls:
Not useless, this one is for an optional toolbar button, `CustomerNewInvoiceOpenAction` is for the menu. I think that they're separate because the toolbar button is enabled only by an extra toolbar buttons preference in Preferences>Business. @Bob-IT can you confirm that the two can't be consolidated without the menu item being hidden too?

OTOH this is the only button controlled by the option so it probably makes sense to just display it always and get rid of the option and controlling code. That change would be for GnuCash 5.

fellen:
I agree, the "New Invoice" button is at least annoying, if no business objects exist in the file.
@gjanssens was there a reason to make it not more context sensitive?

gjanssens:
I believe this was the start of an effort to make an alternative interface geared more towards business users. The effort was never continued after that though. But that's also precisely why it's made optional. The intention really was to make a number of business only features much more prominently accessible to streamline that workflow.

There is another toolbar button to create new invoices in gnc-plugin-page-invoice.c (action name "BusinessNewInvoiceAction"). That one however is only visible when  an invoice is shown on screen.

@fellen if you don't like the button then don't set the option.

To all, while we can't combine the two actions I  think we can have them use  the same string to reduce the translation burden. If we change the name to "New _Invoice" (without the dots) it's the same name as used for the button on invoice pages.

fellen:
> @fellen if you don't like the button then don't set the option.

Where is it set?

gjanssens:
Op vrijdag 2 juli 2021 19:19:04 CEST schreef Frank H. Ellenberger:
> > @fellen if you don't like the button then don't set the option.
> 
> Where is it set?

Edit->Preferences->Business->Enable Extra Buttons

TianXing-Yi:
How about removing "Enable extra buttons" ?

The best way adding an invoice is opening customer overview, choosing a customer, then "New Invoice".
It will be more efficient than "New Invoice...".

christopherlam:
This is where digging into the archives will help: https://www.techrepublic.com/blog/five-apps/-five-tips-for-making-gnucash-work-for-your-business/ (from 10 years ago!) seems to like(d) it!

TianXing-Yi:
> This is where digging into the archives will help: https://www.techrepublic.com/blog/five-apps/-five-tips-for-making-gnucash-work-for-your-business/ (from 10 years ago!) seems to like(d) it!

In my opinion, this extra button should be abandoned.

It needs to search customer, while with many invoices from different customers, this action will be slower and not a good idea.

gjanssens:
> > This is where digging into the archives will help: https://www.techrepublic.com/blog/five-apps/-five-tips-for-making-gnucash-work-for-your-business/ (from 10 years ago!) seems to like(d) it!
> 
> In my opinion, this extra button should be abandoned.
> 
> It needs to search customer, while with many invoices from different customers, this action will be slower, using it is not a good idea.

I understand this button doesn't fit your way of using gnucash.  However I know of at least one business user that relies heavily on this  button (not me by the way).

As for the "extra" search action when using this button, you can also start typing the customer's name in the new invoice dialog. To me that's about as efficient as selecting a customer in a long list in the customer overview and then select the new invoice button.

Additionally, IIRC the original use case was to be able to create a new invoice from *anywhere* in gnucash, not only from pages that are business related.

Again, if the button doesn't work for you, simply disable it. But don't dismiss it just because it doesn't work for you.

TianXing-Yi:
> > > This is where digging into the archives will help: https://www.techrepublic.com/blog/five-apps/-five-tips-for-making-gnucash-work-for-your-business/ (from 10 years ago!) seems to like(d) it!
> 
> > 
> 
> > In my opinion, this extra button should be abandoned.
> 
> > 
> 
> > It needs to search customer, while with many invoices from different customers, this action will be slower, using it is not a good idea.
> 
> 
> 
> I understand this button doesn't fit your way of using gnucash.  However I know of at least one business user that relies heavily on this  button (not me by the way).
> 
> 
> 
> As for the "extra" search action when using this button, you can also start typing the customer's name in the new invoice dialog. To me that's about as efficient as selecting a customer in a long list in the customer overview and then select the new invoice button.
> 
> 
> 
> Additionally, IIRC the original use case was to be able to create a new invoice from *anywhere* in gnucash, not only from pages that are business related.
> 
> 
> 
> Again, if the button doesn't work for you, simply disable it. But don't dismiss it just because it doesn't work for you.

All right, I understand your point of view，also I know how to disable it.

Creating a new invoice from anywhere Looks very well.

But We already have an efficient way.
Why leave this not-so-good feature？

These are my thoughts， thanks.

Bob-IT:
I have had a look at this, with the preference 'extra' enabled, this button appears in two places with nothing populated in subsequent dialog apart from date...
On the customer overview page as 'New Invoice...', along with a 'New Invoice' button populated with customer
On the vendor overview page as 'New Invoice...', along with a 'New Bill' button populated with vendor
On the employee overview page as 'New Invoice...', along with a 'New Voucher' button populated with employee

The other place is on the Invoice/Bill/Voucher pages just like above.

I am inclined to just leave it, I see its use in allowing creating of an Invoice from all business pages and people will enable it if it suites them.

gjanssens:
> I have had a look at this, with the preference 'extra' enabled, this button appears in two places with nothing populated in subsequent dialog apart from date...

As far as I can see it will appear everywhere. On the account hierarchy page, any register page, reports pages,...

On the customer overview page and any customer invoice page the button is redundant, but I think the extra code to hide in on those pages may not be worth it.

fellen:
> Edit->Preferences->Business->Enable Extra Buttons

Then I think the other direction: Add a "New Bill" button, too. And the labels should be unified according section Ellipses in https://developer.gnome.org/hig/stable/writing-style.html to reduce translator burdens.

Bob-IT:
> As far as I can see it will appear everywhere. On the account hierarchy page, any register page, reports pages,...

You are right, not sure how I missed that, maybe I was just focused on the business pages.

jralls:
The button stays. Other ideas for improving this should be done with a new PR.

PR #1066, opened 2021-07-04
Bug 798229 - GncDate::c_formats is being created and destroyed twice...

jralls:
resulting in a double free crash on every exit.

Exclude by name the obsolete pre-GnuCash 4 modules from being inspected
so that they don't try to construct and destroy their static variables
twice.

### DIFF
--- a/libgnucash/gnc-module/gnc-module.c
+++ b/libgnucash/gnc-module/gnc-module.c
@@ -140,7 +140,6 @@ gnc_module_system_search_dirs(void)
  * gnc_module_system_init
  * initialize the module system
  *************************************************************/
-
 void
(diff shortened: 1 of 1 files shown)


gjanssens (approved):
This is a good solution. Perhaps update the comment in gnc_module_system_refresh to hint at why we want to exclude these libraries. We know now, but perhaps forget in a few years.

I did consider suggesting to use GNC_MODULE_PREFIX in the hard coded module names, but we better not: all versions of gnucash that could generate these modules used the 'libgncmod'  as prefix.

jralls:
Merged after adding the comment @gjanssens requested.

PR #1067, opened 2021-07-05
[gnc-pricedb] Don't lookup old prices when doing bulk updates

nomis:
There's no point setting old_price when doing bulk updates because it's not used.


jralls:
Good catch. Just move `old_price`'s decl into the `if` block where it's used.

nomis:
Done

jralls:
Thanks!

PR #1069, opened 2021-07-06
Maint Generic Import Transaction Matcher - fix crashes

goodvibes2:
Fix 2 core dumps. Both dumped core in gnc_import_TransInfo_get_action() because info
is null
1. After double clicking on a potential match line, then right clicking
on the the parent transaction of the potential match
2. After rubber banding a group of transactions which also includes an
expanded potential match. Sometimes also need to right click the
selection to trigger the core dump.

Also remove an unneeded duplicate call to get_action_for_path().

### DIFF
--- a/gnucash/import-export/import-main-matcher.c
+++ b/gnucash/import-export/import-main-matcher.c
@@ -870,6 +870,10 @@ get_action_for_path (GtkTreePath* path, GtkTreeModel *model)
     GtkTreeIter iter;
     gtk_tree_model_get_iter (model, &iter, path);
     gtk_tree_model_get (model, &iter, DOWNLOADED_COL_DATA, &trans_info, -1);
+    if (!trans_info)
+        // selected row is a potential match  (depth 2)
+        // instead of an imported transaction (depth 1)
+        return GNCImport_INVALID_ACTION;
     return gnc_import_TransInfo_get_action (trans_info);
 }
 
--- a/gnucash/import-export/import-main-matcher.c
+++ b/gnucash/import-export/import-main-matcher.c
@@ -976,8 +976,8 @@ gnc_gen_trans_onButtonPressed_cb (GtkTreeView *treeview,
                 GList* selected;
                 GtkTreeModel *model;
                 selected = gtk_tree_selection_get_selected_rows (selection, &model);
-                get_action_for_path (selected->data, model);
-                if (get_action_for_path (selected->data, model) == GNCImport_ADD)
+                if (gtk_tree_path_get_depth ((GtkTreePath *)selected->data) == 1 &&
+                    get_action_for_path (selected->data, model) == GNCImport_ADD)
                     gnc_gen_trans_view_popup_menu (treeview, event, info);
                 g_list_free_full (selected, (GDestroyNotify)gtk_tree_path_free);
             }


gnucash/import-export/import-main-matcher.c
```diff
@@ -977,7 +977,6 @@ gnc_gen_trans_onButtonPressed_cb (GtkTreeView *treeview,
                 selected = gtk_tree_selection_get_selected_rows (selection, &model);
                 if (gtk_tree_path_get_depth( (GtkTreePath *) selected->data) == 1)
                 {
-                    get_action_for_path (selected->data, model);
                     if (get_action_for_path (selected->data, model) == GNCImport_ADD)
```
jralls: With the intervening `get_action_for_path` removed you can combine the if statements with `&&`.

jralls (changes requested):
If the problem is passing `NULL` to `gnc_import_TransInfo_get_action` then it would be clearer to check for NULL and return `GNCImport_INVALID_ACTION` in `get_action_for_path`.

goodvibes2:
> If the problem is passing `NULL` to `gnc_import_TransInfo_get_action` then it would be clearer to check for NULL and return `GNCImport_INVALID_ACTION` in `get_action_for_path`.

So should I remove all my code and comments in this commit and just replace the 'g_assert (info);' in gnc_import_TransInfo_get_action with
```
if (!info)
     // selected row is a potential match (depth 2) instead of an imported transaction (depth 1)
     return GNCImport_INVALID_ACTION;

```
Or should I put some comments in generic-import-design.txt about the depth instead?
(I intend to update generic-import-design.txt somewhat as part of when I allow an imported transaction, for which there is already a transaction with a matching FITID, to be added to Generic Import Matcher for Bug 798205 (Some transactions dropped with no apparent cause or warnings for credit card ofx import).

gnucash/import-export/import-main-matcher.c
goodvibes2: > With the intervening `get_action_for_path` removed you can combine the if statements with `&&`.

That's going to be a long line... Is there a suggested max line length? If it needs to be split, how would you like the line to be split?
I could not see anything about that in https://wiki.gnucash.org/wiki/CodingStandard. I do see some very long lines in Account.cpp as well as lots of different styles of line splitting.

jralls: One should try to keep lines to less than 80 characters, but if there's no good place to break a line and you don't want to create a local variable then it's OK to go over.
But in this case there is a convenient break, right after the `&&`:
```
if (gtk_tree_path_get_depth( (GtkTreePath *) selected->data) == 1 &&
    get_action_for_path (selected->data, model) == GNCImport_ADD))
```

jralls:
I think that the `assert` in `gnc_import_TransInfo_get_action` should stay: It's an obvious precondition that getting a member of a TransInfo requires that the TransInfo* is valid. The check should go in `get_action_for_path` because not all paths have a TransInfo and it's the responsibility of the caller of `gnc_import_TransInfo_get_action` to honor its preconditions. You could claim that `get_action_for_path` has a precondition that the path has a TransInfo attached to it but that's more difficult to express clearly in code, so commenting *in the code* is necessary to explain what's going on. Documenting that sort of thing in a separate file isn't enough because someone working on another bug a year from now might not know to look there. In fact as a separate project it would be a Good Thing™ to convert `generic-import-design.txt` into Doxygen comments in `import-backend.h`.

gnucash/import-export/import-main-matcher.c
```diff
@@ -963,6 +964,8 @@ gnc_gen_trans_onButtonPressed_cb (GtkTreeView *treeview,
             DEBUG("Right mouseClick detected- popup the menu.");
             // Only pop up the menu if there's more than 1 selected transaction,
             // or the selected transaction is an ADD.
+            // Do NOT pop up the menu if the only row selected is a potential match row (depth 2)
+            //  instead of a transaction to be added (depth 1).
```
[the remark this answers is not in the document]
goodvibes2: Fixed and force pushed.

gnucash/import-export/import-main-matcher.c
goodvibes2: Fixed and force pushed

goodvibes2:
Is there something wrong with this PR that is stopping it from being reviewed again?
There's something wrong with my (or the GnuCash?) maint repo that I'd like to fix. When I pulled it last night, there were conflicts with 4 translation files.

gnucash/import-export/import-main-matcher.c
```diff
@@ -976,8 +976,8 @@ gnc_gen_trans_onButtonPressed_cb (GtkTreeView *treeview,
                 GList* selected;
                 GtkTreeModel *model;
                 selected = gtk_tree_selection_get_selected_rows (selection, &model);
-                get_action_for_path (selected->data, model);
-                if (get_action_for_path (selected->data, model) == GNCImport_ADD)
+                if (gtk_tree_path_get_depth ((GtkTreePath *)selected->data) == 1 &&
+                    get_action_for_path (selected->data, model) == GNCImport_ADD)
```
jralls, line 980: You've added another condition here as well as removed the extra call to `get_action_for_path`. The commit message should say why.

jralls (changes requested):
Sorry, it got pushed down the list and I didn't notice it needed another review. Just fix up the second commit message and it can be merged.

goodvibes2:
Replacement is https://github.com/Gnucash/gnucash/pull/1125

PR #1071, opened 2021-07-07
Bug 798234 - Use string cache in FloatingTxn and FloatingSplit

nomis:
Copied strings may be removed from the string cache while they're in the FloatingTxn/FloatingSplit so take a new reference to them.

Fix memory leaks of FloatingTxn and FloatingSplit when they're used temporarily.

Fix memory leak in FloatingTxn by freeing the m_splits list too.

### DIFF
--- a/gnucash/register/ledger-core/split-register-copy-ops.c
+++ b/gnucash/register/ledger-core/split-register-copy-ops.c
@@ -103,13 +103,13 @@ void gnc_float_split_set_transaction (FloatingSplit *fs, Transaction *transactio
 void gnc_float_split_set_memo (FloatingSplit *fs, const char *memo)
 {
     g_return_if_fail (fs);
-    fs->m_memo = memo;
+    CACHE_REPLACE (fs->m_memo, memo);
 };
 
 void gnc_float_split_set_action (FloatingSplit *fs, const char *action)
 {
     g_return_if_fail (fs);
-    fs->m_action = action;
+    CACHE_REPLACE (fs->m_action, action);
 };
 
 void gnc_float_split_set_reconcile_state (FloatingSplit *fs, char reconcile_state)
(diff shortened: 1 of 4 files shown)


jralls:
You've just kicked the crash down the road because you're passing the cached keys to the new transaction and its splits then deleting them from the cache and freeing them, leaving the ptrs you handed to the new transaction and splits to dangle.

The string cache doesn't do anything at all for you in this case. It's only helpful when a particular string is used a lot and these are one-offs. Just `strdup` the source object's strings into the floating object and *either* pass the copy's ptr to the new object or free it if the floating object is discarded instead of realized.

gnucash/register/ledger-core/split-register-copy-ops.c
```diff
@@ -427,3 +436,24 @@ void gnc_float_txn_to_txn_swap_accounts (const FloatingTxn *ft, Transaction *txn
...
+    CACHE_REMOVE (ft->m_description);
+    CACHE_REMOVE (ft->m_notes);
+    CACHE_REMOVE (ft->m_doclink);
+
+    for (iter = ft->m_splits; iter; iter = iter->next)
+    {
+        FloatingSplit *fs = iter->data;
+        gnc_float_split_free (fs);
+    }
+    g_list_free (ft->m_splits);
```
jralls, line 456: You can use `g_list_free_full (ft->m_splits, (GDestroyNotify)gnc_float_split_free);` instead of writing the loop yourself.

jralls:
Please copy and paste the bug summary line--"Bug 798234 - Cut Transaction discards the reference to the description/memo strings so that Paste Transaction will paste uninitialised data (or other strings)"--as the commit summary and explain what you did in the commit message. It makes it easier to generate the release notes; people complain if the bug link isn't verbatim.

nomis:
I've updated the freeing of m_splits and commit message.

jralls:
Thanks!

PR #1073, opened 2021-07-08
Bug 684507 - commodity namespace should be localized

fellen:
First draft for localized Namespaces.
ToDo:
- [x] Edit Security:Type should get a tooltip: "Select an existing type or enter your own. Usually this namespaces should be written uppercase."

A separate issue found while testing: If the type of an existing security changes and trading accounts are enabled, the path of the corresponding trading account should be adjusted.


fellen:
ninja check --rerun-failed --output-on-failure
ninja: unrecognized option '--rerun-failed'

fellen:
Failing Python test fixed.
Should I adjust bindings/guile/commodity-table.scm or can we create it by swig?

qif-import will no longer distinguish between between NASDAQ and NYSE while guessing.

gnucash/import-export/qif-imp/qif-dialog-utils.scm
```diff
@@ -747,21 +747,9 @@
...
-                (< 0 d 4)
-                (<= 2 (- l d) 3))
-           GNC_COMMODITY_NS_NYSE)
-
-          ;; Guess NASDAQ for symbols of 4 characters.
-          ((= l 4)
-           GNC_COMMODITY_NS_NASDAQ)
+          ;; Guess SHARE for symbols of 1-4 characters.
+          ((<= l 4)
+           GNC_COMMODITY_NS_SHARE)
```
gjanssens, line 752: I  suspect this won't work as GNC_COMMODITY_NS_SHARE is not defined in scheme code. GNC-COMMODITY-NS-SHARE probably is (thanks to the wrapper for gnc-commodity.h), so you can try to use that instead.

fellen: Changing all GNC_COMMODITY_NS_ to GNC-COMMODITY-NS- results in 
qif-dialog-utils.scm:739:51: warning: possibly unbound variable `GNC-COMMODITY-NS-MUTUAL'

libgnucash/engine/gnc-commodity.h
```diff
@@ -109,12 +109,10 @@ GType gnc_commodity_namespace_get_type(void);
...
-#define GNC_COMMODITY_NS_NASDAQ "NASDAQ"
-#define GNC_COMMODITY_NS_NYSE   "NYSE"
-#define GNC_COMMODITY_NS_EUREX  "EUREX"
-#define GNC_COMMODITY_NS_MUTUAL "FUND"
-#define GNC_COMMODITY_NS_AMEX   "AMEX"
-#define GNC_COMMODITY_NS_ASX    "ASX"
+#define GNC_COMMODITY_NS_BOND N_("BOND")
+#define GNC_COMMODITY_NS_SHARE N_("SHARE")
+#define GNC_COMMODITY_NS_MUTUAL _("FUND")
+#define GNC_COMMODITY_NS_OPTION _("OPTION")
```
fellen, line 115: The missng Ns was the culprit. No idea, why that change was incomplete.

jralls:
+1 on removing the US stock exchanges as defaults, but let's just lose all of the default categories except currency, they mostly confuse users. The guessed namespace feature in the QIF importer can go too. It's at best US centric and even in the US isn't reliable.

gjanssens:
> +1 on removing the US stock exchanges as defaults, but let's just lose all of the default categories except currency, they mostly confuse users. The guessed namespace feature in the QIF importer can go too. It's at best US centric and even in the US isn't reliable.

I  was tempted to suggest leaving at least a sample (with a name indicating that) to give users a hint namespaces exist to group their commodities. However while looking at namespace treatment  in the UI, it looks like that's not helpful. There seems to be no obvious way to manage these namespaces, in particular, there's no way to delete them. Or at least not one obvious to me. So the user would be stuck with this sample namespace.
At the same time namespaces are very easily created by accident by mistyping in the namespace field. Perhaps (though for another feature than this PR) it's better to add a more formal interface to manage these namespaces.

And with that I do agree it would be better to just drop the default namespaces. It would be a nice code cleanup.

fellen:
@christopherlam can you countercheck that my last commit didn't break anything? At least I saw no  warnings.

fellen:
After fixing distcheck, several normal checks fail by 
In procedure SWIG_Guile_scm2newstr: Wrong type argument in position 1: #<procedure GNC-COMMODITY-NS-CURRENCY ()

I consider to revert the last commit and return to the original task.

fellen:
@jralls , @gjanssens I see the example structure can also be put in the tooltip. i will look for it in a separate commit.

Then the question is, what will remain: template, 3 different consts for currency and  the (all non-currencies). That sounds like it could be optimized, but perhaps it is not worth the costs. 

Then there is the user acceptance. I can imagine some users will demand to revert this change immediately. ;-) So I think it's better done in smaller steps.

The whole guessing - if it contains something useful - should be general and not qif-only. I didn't further look in it.

jralls:
I'm not sure that existing users will notice, but some testing will be needed to make sure that removing the default namespaces doesn't affect a book that already uses those namespaces. I'll test that when I have some time tomorrow afternoon on Sunday; I conveniently have several books with securities in the NYSE and NASDAQ namespaces so it will be easy.

jralls:
Confirmed that removing the default namespaces has no effect on an existing book that has commodities already in the namespace. Users will notice only when they create a new book.

jralls:
@fellen Can I merge this or, considering that removing the default namespaces might be thought a major change, should we cherry-pick it to master instead?

fellen:
Seen from outside the US it is no major change, but a bug fix: Remove annoying US markets from default namespace.
And you found no problems loading you files containing NASDAQ etc.

Geert and you were against introducing basic namespaces 'BOND', 'SHARE', 'FUND', 'OPTION'. Then they should be moved to the instructions in the tooltip.

Thanks to climate change I am currently busy with RL tasks, so feel free to ammend that.

libgnucash/engine/gnc-commodity.h
jralls: Shouldn't `"CURRENCY"` be marked for translation too?

jralls:
> Seen from outside the US it is no major change, but a bug fix: Remove annoying US markets from default namespace.

Fair enough.

I've removed those namespaces and set the default for non-currency commodities to `GNC_COMMODITY_NS_NONCURRENCY' with corresponding changes to the QIF importer. I also added the requested tooltip.

gnucash/gtkbuilder/dialog-commodity.glade
```diff
@@ -305,7 +305,8 @@
                 <property name="model">liststore2</property>
                 <property name="has-entry">True</property>
                 <property name="entry-text-column">0</property>
-                <signal name="changed" handler="gnc_ui_commodity_changed_cb" swapped="no"/>
+                <property name="has-tooltip">True</property>
+                <property name="tooltip-markup" translatable="yes">Select a category for the commodity or enter a new one. One might use investment categories like STOCKS and BONDS or exchange names like NASDAQ and LSE.</property>                <signal name="changed" handler="gnc_ui_commodity_changed_cb" swapped="no"/>
```
gjanssens, line 309: You lost a newline here. Did that happen while using the glade editor or did you just change the file via a text editor ?

jralls: Fixed. I use emacs and avoid Glade.

gjanssens:
👍🏻

John Ralls ***@***.***> schreef op 3 maart 2022 22:36:58 CET:
***@***.*** commented on this pull request.
>
>
>
>> @@ -305,7 +305,8 @@
>                 <property name="model">liststore2</property>
>                 <property name="has-entry">True</property>
>                 <property name="entry-text-column">0</property>
>-                <signal name="changed" handler="gnc_ui_commodity_changed_cb" swapped="no"/>
>+                <property name="has-tooltip">True</property>
>+                <property name="tooltip-markup" translatable="yes">Select a category for the commodity or enter a new one. One might use investment categories like STOCKS and BONDS or exchange names like NASDAQ and LSE.</property>                <signal name="changed" handler="gnc_ui_commodity_changed_cb" swapped="no"/>
>
>Fixed. I use emacs and avoid Glade.
>
>-- 
>Reply to this email directly or view it on GitHub:
>https://github.com/Gnucash/gnucash/pull/1073#discussion_r819081734
>You are receiving this because you were mentioned.
>
>Message ID: ***@***.***>
-- 
Verzonden van mijn Android telefoon met K-9 Mail.

jralls:
@gjanssens Do you think this is ready to merge? I guess from @fellen's reply that he'll be too busy to vote before the 4.10 release.

gjanssens:
I have little experience with non-currency code, but it looks ok to me.

PR #1074, opened 2021-07-08
Fix Typo: between "new customer/vendor/employee" and "find customer/vendor/employee"

TianXing-Yi:
The strings doesn't match, so I do this, making it more clear while using these functions.


fellen:
Redone in PR #1075.

BTW there is no need to close a PR and create a new one. You can also force-push the commit:
`git push -f <your-remote>`

TianXing-Yi:
Get it, Thanks sir.

PR #1075, opened 2021-07-08
Fix Typo: rename some labels

TianXing-Yi:
Just making it more clear while using "new customer dialog" and "find customer dialog"

### DIFF
--- a/gnucash/gtkbuilder/dialog-preferences.glade
+++ b/gnucash/gtkbuilder/dialog-preferences.glade
@@ -1748,7 +1748,7 @@ many months before the current month</property>
                 </child>
                 <child>
                   <object class="GtkRadioButton" id="pref/general/retain-type-days">
-                    <property name="label" translatable="yes">For</property>
+                    <property name="label" translatable="yes">For Day(s)</property>
                     <property name="visible">True</property>
                     <property name="can_focus">True</property>
                     <property name="receives_default">False</property>


gnucash/gnome/dialog-vendor.c
```diff
@@ -713,9 +713,9 @@ gnc_vendor_search (GtkWindow *parent, GncVendor *start, QofBook *book)
     /* Build parameter list in reverse order */
     if (params == NULL)
     {
-        params = gnc_search_param_prepend (params, _("Billing Contact"), NULL, type,
+        params = gnc_search_param_prepend (params, _("Payment Contact"), NULL, type,
```
gjanssens, line 716: In accounting 'Billing' is not the same as 'Payment'. I think this is not a good substitute. I see however this is also used in the 'New Vendor' dialog window, so this may need some more discussion.

TianXing-Yi: In new vendor dialog, there's no field named "Billing Contact", only "Payment Address Name"

In find vendor dialog, one option is "Billing contact"

They describe one field in different strings.

So I do these: 
Rename "Name" to "Contact", "Billing contact" to "Payment Contact"

TianXing-Yi: @gjanssens 
Hi, sir, below are the screenshots, before and after changes. 
Would these changes be ok, or you give me some suggestions? 

customer
[Before](https://github.com/TianXing-Yi/GImage/blob/master/2.0-Before.png)
[After](https://github.com/TianXing-Yi/GImage/blob/master/2.0-After.png)

vendor
[Before](https://github.com/TianXing-Yi/GImage/blob/master/2.1-Before.png)
[After](https://github.com/TianXing-Yi/GImage/blob/master/2.1-After.png)

employee
[Before](https://github.com/TianXing-Yi/GImage/blob/master/2.2-Before.png)
[After](https://github.com/TianXing-Yi/GImage/blob/master/2.2-After.png)

gjanssens: As I said `Payment information` is not the proper term IMO. There is no such thing as a `Payment contact` or a `Payment address`. The only payment related element in all of the vendor options are the `Payment terms`.

So instead of changing `Billing` to `Payment` everywhere, you should go the other way around.

TianXing-Yi: Get it.

gjanssens (changes requested):
I think you have made a good start here. However your initial work reveals much more inconsistencies than just the ones you have uncovered so far. So rather than go through each change you have made, I'll offer a overarching design suggestion to work towards:

### Vendor
Vendor

#### In the vendor dialog
* Change `Identification` to `Vendor Identification`. That gives enough context to the two enclosed fields we can drop `Vendor` and `Company` from `Number` and `Name` respectively.
* Change to `Vendor Address` instead of `Billing Address`. Vendors only have one address.
* `Payment Information` -> `Billing Information` for the second tab
* `Contact` -> `Contact Name` to clarify this is an optional name for a contact at t his address.

#### In the find  dialog
* Use `Number` and `Name` without their prefixes. From the context it should be clear these refer to the vendor number and name. Benefit of the shorter name will be smaller columns.
* Use `Contact Name` rather than just `Contact` to be consistent with the vendor dialog.

### Customer
Customer

The changes are much like for Vendor. As with vendors, the customer's primary address is called `Customer Address`. Customers however can optionally have a second address, the shipping address. So to disambiguate between the contact names in these two addresses in  the search  form there are `Contact Name` (which refers to the `Customer Address` contact name and `Shipping Contact Name` referring to the contact name on the shipping address.
I have deliberately not chosen `Customer Contact Name` to refer to the `Customer Address` because `Customer` is implied by the way the dialogs are designed now.

### Employee
Employee

* Again changes similar to the other two, though with a few important distinctions. User name has simply become `Employee Identification` -> `Name`. According to @derekatkins `Username` was chosen with the idea in mind to add fine grained per user access controls to gnucash but that was never implemented. I also think if we ever implement that it should not be at the employee level. So opt here to bring this field in line with the same in vendor and customer. From that follows that the `Employee Address` -> `Name` also becomes the optional `Contact Name` again already in vendor and customer.
* Considering we have no implementation for access control, I also propose to drop the `Access Control` tab completely.
* To be consistent with vendors and customers the order of search parameters and result columns is changed to first show the `Employee Identification`->`Name` and end with `Contact Name`.


All these changes together should give us a more consistent user experience, reusing more terms than before (which means fewer translations), while retaining a space efficient result table.

gnucash/gnome/dialog-vendor.c
jralls: @gjanssens, payment for vendors and employees, billing for customers.

gjanssens: @jralls Why ? It sound very weird to me  to have an address related to a payment. That's why I suggested Vendor/Customer/Employee address in the followup design.

jralls: Why does it seem weird to have an address for payment? If you were to still pay bills with paper checks sent in the post you have to put an address on the envelope. 

Doesn't it seem weird to you to (send a) bill (to) a vendor?

gjanssens: I probably don't fully understand the term "billing" then.

Regardless however, my later point is that I don't think we should specialize the address type on anything if we only have one address. It's just the vendor's address, right ?
For customer there's a provision for an optional shipping address. So  some specialization is  needed. But for me it still is comfortable to call the main address simply the customer address and  keep shipping address as an optional extra specification. All customers have *an* address, but not all of them have a separate shipping address.

What would you propose as alternative for "Billing/Payment information" for vendors? For me "Payment Information" doesn't cover what's on that tab. A payment term, yes, that's payment information, but that's so both for customers and vendors. Currency is more than just payment information. It defines a property of all  amounts on the bill.  Of course that also impacts the payment,  but I'm not happy just calling it payment information. Similarly Tax details are an accounting property of the bill, that has among others an effect on the total amount. But I hope  you see my point: I think payment information is too narrow. Thinking about this "Bill/Invoice/Voucher defaults" probably comes closer - all but currency can be overridden on a per bill basis.

gnucash/gnome/dialog-customer.c
```diff
@@ -930,11 +930,11 @@ gnc_customer_search (GtkWindow *parent, GncCustomer *start, QofBook *book)
...
     if (columns == NULL)
     {
-        columns = gnc_search_param_prepend (columns, _("Contact"), NULL, type,
+        columns = gnc_search_param_prepend (columns, _("Billing Contact"), NULL, type,
                                             CUSTOMER_ADDR, ADDRESS_NAME, NULL);
-        columns = gnc_search_param_prepend (columns, _("Company"), NULL, type,
+        columns = gnc_search_param_prepend (columns, _("Company Name"), NULL, type,
                                             CUSTOMER_NAME, NULL);
-        columns = gnc_search_param_prepend (columns, _("ID #"), NULL, type,
+        columns = gnc_search_param_prepend (columns, _("Customer Number"), NULL, type,
```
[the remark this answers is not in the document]
fellen: As # in English often stands for number, it is redundant. I would usually use ID - short for some Identifying key - in contrast to a pure numerical number.

gnucash/gnome/dialog-vendor.c
fellen: The tab holds the customized Billing Terms.

fellen: > Why does it seem weird to have an address for payment? If you were to still pay bills with paper checks sent in the post you have to put an address on the envelope.
> 
> Doesn't it seem weird to you to (send a) bill (to) a vendor?

In Europe paper cheques are now rarely used. The IBAN is the primary address of payments followed by BIC+account-number.

Retranslated I would call it the (msgcontext=general) invoice address.

jralls: Maybe it would help to know that "Bill" and "Invoice" are essentially synonymous in this usage. A customer has a billing address because you send them a bill demanding payment and a vendor has a payment address because you'll be sending them a payment, not a bill.

The Vendor and Customer Edit dialogs are correct as they are now: The Vendor one says Payment Address and Payment Information, the Customer one says Billing Address and Billing Information; Customer has an additional Shipping Address tab.

I don't see any problem with changing Billing/Payment Address to just Address *in English* but I think that they'll need translation context to make them separate msgids because it's possible that they'd need different translations in some languages. Changing the Payment/Billing Information tabs to just Information would be confusing even in English.

TianXing-Yi:
Get it.

Thanks for your detailed explanation.
This night I will fix them.

gnucash/gnome/dialog-vendor.c
TianXing-Yi: I agree with gjanssens's point of view. 

For "Billing/Payment Address", the suitable labels may be "Bank Identifier Code, IBAN, Address of bank, Account".

According to existing labels, "Vendor Address" will be better than "Payment Address".

TianXing-Yi:
Hi, I have done some changes. Now, GUI looks like these:
[For-Preferences](https://github.com/TianXing-Yi/GImage/blob/master/2021-07-10_23-19.png)
[For-sx](https://github.com/TianXing-Yi/GImage/blob/master/2021-07-10_23-20.png)
[Sort](https://github.com/TianXing-Yi/GImage/blob/master/2021-07-10_23-21.png)
[Filter](https://github.com/TianXing-Yi/GImage/blob/master/2021-07-10_23-22.png)
[Customer](https://github.com/TianXing-Yi/GImage/blob/master/2021-07-10_23-23.png)
[Vendor](https://github.com/TianXing-Yi/GImage/blob/master/2021-07-10_23-24.png)
[Employee](https://github.com/TianXing-Yi/GImage/blob/master/2021-07-10_23-25.png)

Are there any suggestions?

gnucash/gtkbuilder/dialog-customer.glade
```diff
@@ -210,7 +210,7 @@
                       <object class="GtkLabel" id="label2">
                         <property name="visible">True</property>
                         <property name="can_focus">False</property>
-                        <property name="label" translatable="yes">Identification</property>
+                        <property name="label" translatable="yes">Information</property>
```
fellen, line 213: No, Name and Number are the Item which idenify the company. Information is too general, so it becomes meaningless

jralls:
@TianXing-Yi I don't see any differences between the first 4 screenshots and what GnuCash 4.6 looks like now and your Employee link 404s.

I don't think much of the changes in the Customer or Vendor dialogs either. They're better the way they are now.

I don't generally like to pull rank, but let's keep in mind that I'm the only native en_US speaker who's commented so far.. The dialogs *as they are now* are correct in en_US usage. If they're not clear for translators then they need translator comments or translation context to make them more clear, not rewriting. They especially shouldn't be rewritten to make them less clear to US users.

gnucash/gtkbuilder/dialog-preferences.glade
```diff
@@ -1748,7 +1748,7 @@ many months before the current month</property>
                 </child>
                 <child>
                   <object class="GtkRadioButton" id="pref/general/retain-type-days">
-                    <property name="label" translatable="yes">For</property>
+                    <property name="label" translatable="yes">For Day(s)</property>
```
jralls, line 1751: There's a 'days' label on the spin button so this will be redundant. @fellen is there a way to have context translations in GtkBuilder files?

gnucash/gnome/dialog-customer.c
```diff
@@ -917,24 +917,24 @@ gnc_customer_search (GtkWindow *parent, GncCustomer *start, QofBook *book)
...
     if (columns == NULL)
     {
-        columns = gnc_search_param_prepend (columns, _("Contact"), NULL, type,
+        columns = gnc_search_param_prepend (columns, _("Person"), NULL, type,
                                             CUSTOMER_ADDR, ADDRESS_NAME, NULL);
-        columns = gnc_search_param_prepend (columns, _("Company"), NULL, type,
+        columns = gnc_search_param_prepend (columns, _("Name"), NULL, type,
                                             CUSTOMER_NAME, NULL);
-        columns = gnc_search_param_prepend (columns, _("ID #"), NULL, type,
+        columns = gnc_search_param_prepend (columns, _("Number"), NULL, type,
```
fellen, line 937: It can be alphanumeric. E.g. I use their country code as prefix. So ID is more appropriate.

gnucash/gtkbuilder/dialog-sx.glade
```diff
@@ -1231,7 +1231,7 @@
                             </child>
                             <child>
                               <object class="GtkRadioButton" id="rb_num_occur">
-                                <property name="label" translatable="yes">For</property>
+                                <property name="label" translatable="yes">For Time(s)</property>
```
jralls, line 1234: Same problem as with the preferences dialog above, plus 'Time(s)' isn't as clear as 'occurrences'.

That aside in both cases these are effectively composed strings, the equivalent of writing `_("For") << secs << _("occurrences")` instead of `_("For %d occurrences")`. It may be impossible to translate correctly into some languages.

gnucash/gnome/gnc-plugin-page-register.c
```diff
@@ -4265,7 +4265,7 @@ gnc_plugin_page_register_cmd_view_filter_by (GtkAction* action,
                                   gnc_window_get_gtk_window (GNC_WINDOW (GNC_PLUGIN_PAGE (page)->window)));
 
     /* Translators: The %s is the name of the plugin page */
-    title = g_strdup_printf (_ ("Filter %s by..."),
+    title = g_strdup_printf (_ ("Filter %s by"),
```
fellen, line 4268: No, Filter is a command, but it is not immediately executed.

gnucash/gtkbuilder/dialog-customer.glade
```diff
@@ -240,7 +240,7 @@
                             <property name="visible">True</property>
                             <property name="can_focus">False</property>
                             <property name="halign">end</property>
-                            <property name="label" translatable="yes">Name</property>
+                            <property name="label" translatable="yes">Person</property>
```
fellen, line 243: No bots allowed any longer? ;-)

jralls: Or departments or mail stops...

Like I said, it's correct as-is for US English. Translate it how you want but leave the English alone; add context to distinguish it from other possibly different uses of the word. Yes, I understand that increases the load on some translators but without the context it might be impossible to correctly translate into some languages.

gnucash/gtkbuilder/dialog-preferences.glade
fellen: >  @fellen is there a way to have context translations in GtkBuilder files?

https://wiki.gnucash.org/wiki/I18N#Strings_in_Glade_files

gnucash/gnome/gnc-plugin-page-register.c
jralls: This is the dialog's title, so it's not about immediacy of execution. But it would be more correct to use a colon (`:`) than an ellipsis (`…`).

gnucash/gtkbuilder/dialog-sx.glade
fellen: > instead of `_("For %d occurrences")`

Because %d is the input, the rest output.
But
```
comments="Begin of 'For %d occurrences'"
…
comments="End of 'For %d occurrences'"
```
could help.

jralls: Better but still won't help for languages where the string would be "%d for occurrences" or "for occurrences %d".

TianXing-Yi:
Well, the best option is to leave it as it is. 

In Chinese translate, I cannot deal well with "For" at these two position. Rename one will be OK.

jralls:
> https://github.com/Gnucash/gnucash/blob/e1c153a516f0689b04efff7191639ad604a6b6b6/libgnucash/app-utils/gnc-ui-util.c#L129

Why won't setting context on them, which will create separate msgids, work?

fellen:
> Well, the best option is to leave it as it is.
> 
> In Chinese translate, I cannot deal well with "For" at these two position. Rename one will be OK.

The user friendly solution would be to add accellerators in Edit->Preferences.
Alternatively or in combination in General:Log/Backup we could also replace "For" by "Last"or even better "For last".
A comment should be added.

This might be easier than using a context.

I didn't look in the other occurrence.

PR #1076, opened 2021-07-08
add a shipping contact name column for FindCustomerDialog

TianXing-Yi:

### DIFF
--- a/gnucash/gnome/dialog-customer.c
+++ b/gnucash/gnome/dialog-customer.c
@@ -930,6 +930,8 @@ gnc_customer_search (GtkWindow *parent, GncCustomer *start, QofBook *book)
     /* Build the column list in reverse order */
     if (columns == NULL)
     {
+        columns = gnc_search_param_prepend (columns, _("Shipping Contact"), NULL, type,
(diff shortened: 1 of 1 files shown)


TianXing-Yi:
four parameter options, only three-column results.
Screenshot below:
[Before](https://github.com/TianXing-Yi/GImage/blob/master/1-Before.png)
[After](https://github.com/TianXing-Yi/GImage/blob/master/1-After.png)

gjanssens:
I'd have gone the other way around and add a column for shipping contact.

fellen:
No, some companies have 2 different addresses like
a real street for goods and a PO box for invoices or 
a factory for the goods, but the head office for invoices.

That is implemented for customers. 
@gjanssens The issue is, the shipping address is missing in vendors. How can I return goods?

gjanssens:
> That is implemented for customers.
> @gjanssens The issue is, the shipping address is missing in vendors. How can I return goods?

It was never implemented for vendors. As it stands now you'll have to dig up your paper invoice and find the proper return address there.

jralls:
Thanks!

PR #1080, opened 2021-07-11
Commit root accounts after loading from XML

nomis:
The root accounts start with a non-zero editlevel because BeginEdit is called for them during loading but not committed after loading.

If the book is then closed without performing any further edits that would require a commit, the Account book_end process does nothing because the root account is still being edited and so none of the accounts are freed.

### DIFF
--- a/libgnucash/backend/xml/io-gncxml-v2.cpp
+++ b/libgnucash/backend/xml/io-gncxml-v2.cpp
@@ -699,6 +699,7 @@ qof_session_load_from_xml_file_v2_full (
     QofBookFileType type)
 {
     Account* root;
+    Account* template_root;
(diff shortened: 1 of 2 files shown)


nomis:
One of the AQB test cases uses an XML file with no accounts. In this case the root account is auto-created and it will have already been committed. The other test case does have a root account and this change will commit it after loading the file.

However, even with https://github.com/Gnucash/gnucash/pull/1079 applied, test-load-xml2 already has a committed root account at this stage of the loading process.

Similar changes probably also need to be made to the other backends because they have the same "commit descendants of the root" behaviour.

jralls:
A collective thanks for the several PRs just merged!

PR #1081, opened 2021-07-11
qof_instance_dispose should remove its type string from the string cache

nomis:
If QofInstancePrivate has no collection then qof_instance_dispose() returns without completing the rest of its dispose processes.

Change the check for a collection so that it only affects the call to qof_collection_remove_entity().


jralls:
Your version is a little less tortured but it doesn't seem to have anything at all to do with the commit summary.

nomis:
I've fixed the commit message.

jralls:
Thanks.

PR #1082, opened 2021-07-11
Free template root accounts on book end

nomis:
The template root should be unset so that it's destroyed when the book is ending. This is required to free all the template accounts and referenced strings.


jralls:
Nope. The crash in macOS is because this frees the template root account before the sx itself is freed; in the course of the sx getting freed it wants to free its splits, but freeing template root frees all of its children so there's no way to get the split list.

jralls:
I found and fixed the problem and pushed the result.

PR #1029, opened 2021-06-08
More memory leaks

christopherlam:
Continuation of #1019

PR #1086, opened 2021-07-12
More memory leaks

christopherlam:
Continuation of #1029 from valgrind.

`owner_list` is populated with `gncBusinessGetOwnerList` which calls `get_ownerlist_cb` which calls `gncOwnerNew`, and must be freed with `gncOwnerFree`.

### DIFF
--- a/gnucash/gnome-utils/gnc-tree-model-owner.c
+++ b/gnucash/gnome-utils/gnc-tree-model-owner.c
@@ -192,6 +192,9 @@ gnc_tree_model_owner_finalize (GObject *object)
     model = GNC_TREE_MODEL_OWNER (object);
     priv = GNC_TREE_MODEL_OWNER_GET_PRIVATE(model);
 
+    if (priv->owner_list)
+        g_list_free_full (priv->owner_list, (GDestroyNotify) gncOwnerFree);
+
     priv->book       = NULL;
     priv->owner_list = NULL;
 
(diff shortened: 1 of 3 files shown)


Bob-IT:
@christopherlam Just for information, the problem with the freeing TaxTables you mentioned in a bug I have found, there is a similar one I think for BillTerms, also I think that Jobs are not being freed from Vendors and Customers. I will create a PR for you to try later.

gnucash/gnome-utils/gnc-tree-model-owner.c
```diff
@@ -1004,6 +1010,12 @@ gnc_tree_model_owner_event_handler (QofInstance *entity,
             DEBUG("can't generate path");
             break;
         }
+
+        g_assert_not_reached ();
+        /* above disables. untested; removing an owner is not allowed
+           in UI */
+        priv->owner_list = g_list_remove (priv->owner_list, &owner);
+        gncOwnerFree (&owner);
```
christopherlam, line 1018: This part is not tested live, hence has `g_assert_not_reached`. Otherwise I think this branch is safe?

jralls: No, deliberately crashing isn't safe. Besides, `assert`s are usually disabled in release builds so they're only useful as part of comprehensive testing.

Next, it's nonsensical to assert that a `switch case` isn't ever used and also provide implementation code for it.

Is the GtkTreeModel really the owner of the gncOwner? gncOwner is a subclass of QofInstance and most of those are owned by the QofCollections in qofid.c and are supposed to exist for the duration of the session unless the user explicitly deletes them. A QOF_REMOVE event suggests that that has happened, in which case the GtkTreeModel should remove its pointer to the gncOwner from the list as that object will soon be freed but shouldn't free it; that will be taken care of once the event processing is finished.

christopherlam: Agree pointless to do `g_list_remove` since the whole `owner_list` will be freed properly when the session is closed.

PR #1088, opened 2021-07-13
Object that are not freed / destroyed

Bob-IT:
Chris mentioned a problem with Tax Tables not being freed on closure so had a look and I think the first commit fixes that. In the same commit there is a fix for Bill Terms as they are very similar.

I thought I would have a look at the other objects and found Jobs were not being destroyed when Customer and Vendors are so that's the second commit followed by Accounts and Scheduled Transactions Accounts.
The last commit, which I have commented out, would delete Scheduled transaction accounts that have no splits when the root template account is destroyed, This is an extension of #1082, I had 30 which does work but it fails test 22, so not sure if I need to change the test or maybe the commit is wrong.

There are two other objects that I have not yet got to the bottom of...
I have 52 g_object_new prices but only 48 destroyed and Splits, g_object_new is used in three functions, xaccMallocSplit, xaccDupeSplit and xaccSplitCloneNoKvp which does not show up in my tests.
12034 malloced splits
12083 duped splits
24068 freed splits
12034 destroyed splits.
@christopherlam see if this helps with your memory leaks.

### DIFF
--- a/libgnucash/engine/Account.cpp
+++ b/libgnucash/engine/Account.cpp
@@ -1316,7 +1316,7 @@ xaccFreeAccount (Account *acc)
         PERR (" instead of calling xaccFreeAccount(), please call\n"
               " xaccAccountBeginEdit(); xaccAccountDestroy();\n");
 
-        /* First, recursively free children */
+        /* First, recursively free children, also frees list */
         xaccFreeAccountChildren(acc);
     }
 
@@ -5992,13 +5992,31 @@ gnc_account_delete_all_bayes_maps (Account *acc)
 /* QofObject function implementation and registration */
 
 static void
+destroy_all_child_accounts (Account *acc, gpointer data)
+{
+    xaccAccountBeginEdit (acc);
+    xaccAccountDestroy (acc);
+}
+
+static void
 gnc_account_book_end(QofBook* book)
 {
-    Account *root_account = gnc_book_get_root_account(book);
+    Account *root_account = gnc_book_get_root_account (book);
+    GList *accounts;
+
     if (!root_account)
         return;
-    xaccAccountBeginEdit(root_account);
-    xaccAccountDestroy(root_account);
+
+    accounts = gnc_account_get_descendants (root_account);
+
+    if (accounts)
+    {
+        accounts = g_list_reverse (accounts);
+        g_list_foreach (accounts, (GFunc)destroy_all_child_accounts, nullptr);
+        g_list_free (accounts);
+    }
+    xaccAccountBeginEdit (root_account);
+    xaccAccountDestroy (root_account);
 }
 
 #ifdef _MSC_VER
--- a/libgnucash/engine/Transaction.c
+++ b/libgnucash/engine/Transaction.c
@@ -1769,7 +1769,7 @@ xaccTransRollbackEdit (Transaction *trans)
             xaccSplitRollbackEdit(s);
             SWAP_STR(s->action, so->action);
             SWAP_STR(s->memo, so->memo);
-	    qof_instance_copy_kvp (QOF_INSTANCE (s), QOF_INSTANCE (so));
+            qof_instance_copy_kvp (QOF_INSTANCE (s), QOF_INSTANCE (so));
             s->reconciled = so->reconciled;
             s->amount = so->amount;
             s->value = so->value;
@@ -1778,7 +1778,6 @@ xaccTransRollbackEdit (Transaction *trans)
             //SET_GAINS_A_VDIRTY(s);
             s->date_reconciled = so->date_reconciled;
             qof_instance_mark_clean(QOF_INSTANCE(s));
-            xaccFreeSplit(so);
         }
         else
         {
@@ -1805,6 +1804,10 @@ xaccTransRollbackEdit (Transaction *trans)
         }
     }
     g_list_free(slist);
+
+    // orig->splits may still have duped splits so free them
+    for (node = orig->splits; node; node = node->next)
+        xaccFreeSplit(node->data);
     g_list_free(orig->splits);
     orig->splits = NULL;
 
--- a/libgnucash/engine/gncCustomer.c
+++ b/libgnucash/engine/gncCustomer.c
@@ -353,6 +353,8 @@ static void gncCustomerFree (GncCustomer *cust)
     gncAddressDestroy (cust->addr);
     gncAddressBeginEdit (cust->shipaddr);
     gncAddressDestroy (cust->shipaddr);
+
+    gncJobFreeList (cust->jobs);
     g_list_free (cust->jobs);
     g_free (cust->balance);
 
--- a/libgnucash/engine/gncCustomer.c
+++ b/libgnucash/engine/gncCustomer.c
@@ -358,11 +358,12 @@ static void gncCustomerFree (GncCustomer *cust)
     g_list_free (cust->jobs);
     g_free (cust->balance);
 
-    if (cust->terms)
-        gncBillTermDecRef (cust->terms);
-    if (cust->taxtable)
+    if (!qof_book_shutting_down (qof_instance_get_book (QOF_INSTANCE(cust))))
     {
-        gncTaxTableDecRef (cust->taxtable);
+        if (cust->terms)
+            gncBillTermDecRef (cust->terms);
+        if (cust->taxtable)
+            gncTaxTableDecRef (cust->taxtable);
     }
 
     /* qof_instance_release (&cust->inst); */
--- a/libgnucash/engine/gncJob.c
+++ b/libgnucash/engine/gncJob.c
@@ -231,6 +231,18 @@ GncJob *gncJobCreate (QofBook *book)
     return job;
 }
 
+static void free_job_list (GncJob *job)
+{
+    gncJobBeginEdit (job);
+    gncJobDestroy (job);
+}
+
+void gncJobFreeList (GList *jobs)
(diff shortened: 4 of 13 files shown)


Bob-IT:
I have found the discrepancy in prices and splits and added two new commits and now the figures add up. I think there is some overlap with Simon Arlot's PR's so will try and test his changes and may be add the required ones of mine on top.

Bob-IT:
I need to recheck all these commits as there appears to be an issue with the TaxTables commit.

jralls:
@Bob-IT #1082 tries to free the template accounts too soon, so if you're so inclined please add freeing them in the right place.

Bob-IT:
I think this covers what I set out to do, on testing the number of g_object_new's equaled the free's for the main objects.
@christopherlam Hopefully I think I have fixed the bug report 798252 you raised by wrapping the use of TaxTable and BillTerm DecRef functions in a test for shutdown, there was a possibility that TaxTables and BillTerms were already freed when the call to DecRef was made. If you could give it a spin with your test file which I unfortunately could not get to fail.
@jralls  For this PR I have left the use of 'xxxFree' instead of moving functions to a 'dispose' function which I may consider for a different PR.

jralls:
@Bob-IT 
> For this PR I have left the use of 'xxxFree' instead of moving functions to a 'dispose' function which I may consider for a different PR.

That's OK. @gjanssens and I agreed a couple of years ago that it makes more sense to resolve engine's screwed up memory management by replacing the GObject mis-implementation with C++ classes. We want to do that long-term anyway so there's no point in doing the job twice.

libgnucash/engine/gncCustomer.c
```diff
@@ -358,11 +358,12 @@ static void gncCustomerFree (GncCustomer *cust)
...
-    if (cust->terms)
-        gncBillTermDecRef (cust->terms);
-    if (cust->taxtable)
+    if (!qof_book_shutting_down (qof_instance_get_book (QOF_INSTANCE(cust))))
     {
-        gncTaxTableDecRef (cust->taxtable);
+        if (cust->terms)
+            gncBillTermDecRef (cust->terms);
+        if (cust->taxtable)
+            gncTaxTableDecRef (cust->taxtable);
```
jralls, line 366: This seems wrong at a deeper level: The refcount isn't about memory management, it's about whether a Bill Term or Tax Table is in use by an Owner. The refcount should be decremented only if the Term/Table is disconnected from the Owner or if the Owner is itself deleted by the user.

libgnucash/engine/Account.cpp
```diff
@@ -5991,14 +5991,32 @@ gnc_account_delete_all_bayes_maps (Account *acc)
 /* ================================================================ */
 /* QofObject function implementation and registration */
 
+static void
+destroy_all_child_accounts (Account *acc, gpointer data)
```
jralls, line 5995: Just `destroy_account` or `destroy_account_cb`. It only destroys one.

libgnucash/engine/Account.cpp
```diff
@@ -5991,14 +5991,32 @@ gnc_account_delete_all_bayes_maps (Account *acc)
...
+    accounts = gnc_account_get_descendants (root_account);
+
+    if (accounts)
+    {
+        accounts = g_list_reverse (accounts);
+        g_list_foreach (accounts, (GFunc)destroy_all_child_accounts, nullptr);
+        g_list_free (accounts);
+    }
+    xaccAccountBeginEdit (root_account);
+    xaccAccountDestroy (root_account);
```
jralls, line 6019: `xaccAccountDestroy` [should do this already](https://github.com/Gnucash/gnucash/blob/6bf5a618debee47d90317ff287767ba6d1fd32fa/libgnucash/engine/Account.cpp#L1455). If it's broken let's fix that instead of duplicating it.

libgnucash/engine/Transaction.c
```diff
@@ -1805,6 +1804,10 @@ xaccTransRollbackEdit (Transaction *trans)
         }
     }
     g_list_free(slist);
+
+    // orig->splits may still have duped splits so free them
+    for (node = orig->splits; node; node = node->next)
+        xaccFreeSplit(node->data);
     g_list_free(orig->splits);
```
jralls, line 1811: Why not `g_list_free_full(orig->splits, xaccFreeSplit)`?

That said I don't see any condition in `xaccTransRollbackEdit` that skips splits that aren't changed. The problem with this is that the splits that are freed aren't removed from the list so this will cause double-frees. Fortunately `xaccFreeSplit` tries to detect this and escape; unfortunately it does it in a way that's incompatible with GObject's way of marking freed objects so it can still crash.

Splits are leaked only if the edited transaction has fewer splits than the clone.

jralls (changes requested):

libgnucash/engine/gncJob.c
```diff
@@ -231,6 +231,18 @@ GncJob *gncJobCreate (QofBook *book)
     return job;
 }
 
+static void free_job_list (GncJob *job)
```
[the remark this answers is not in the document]
Bob-IT: OK changed.

libgnucash/engine/Account.cpp
Bob-IT: I forgot to change this after I made a similar change for the template root account. Changed locally and it does remove all accounts.

libgnucash/engine/Account.cpp
Bob-IT: Not needed.

libgnucash/engine/gncJob.c
```diff
@@ -231,6 +231,18 @@ GncJob *gncJobCreate (QofBook *book)
...
+{
+    gncJobBeginEdit (job);
+    gncJobDestroy (job);
+}
+
+void gncJobFreeList (GList *jobs)
+{
+    GList *job_list = g_list_copy (jobs);
```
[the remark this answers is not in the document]
Bob-IT: The reason I did it this way was that as part of gncJobFree there is a switch case that calls gncVendorRemoveJob or Customer which looks at the existing job list and removes a link to the job being freed.

libgnucash/engine/Transaction.c
Bob-IT: > Why not `g_list_free_full(orig->splits, xaccFreeSplit)`?

Did not think of it...
I am a bit confused about the removal from list, what list?
Mind you it does say at the start of this function it is 'terribly complex'.
I have added some print statements starting '=' to the rollback function and on startup I am seeing this...

dupe_trans called, from 0x558a576b8610, to 0x558a56ca4ba0
xaccDupeSplit called, 0x558a576d0660
xaccDupeSplit called, 0x558a576d08c0
xaccDupeSplit called, 0x558a576d09f0
xaccDupeSplit called, 0x558a576d0530
= xaccTransRollbackEdit called, 0x558a576b8610
= xaccTransRollbackEdit called1, split s is 0x558a576baa10
= xaccTransRollbackEdit called1, split s is 0x558a576bab40
= xaccTransRollbackEdit called1, split s is 0x558a576bac70
= xaccTransRollbackEdit called1, split s is 0x558a576ba8e0
= my free orig->splits list 0x558a576d0660
= my free orig->splits list 0x558a576d08c0
= my free orig->splits list 0x558a576d09f0
= my free orig->splits list 0x558a576d0530
  xaccFreeTransaction called, 0x558a56ca4ba0


As you can see for this duped transaction the 4 orig splits are not freed and in total equates to 49 splits which is what I was trying to free. From what I understand is happening in the for loop, it is looking at the transaction slist and iterating over it and if the split is not dirty it continues. So the corresponding split in the orig->slist is not freed.

libgnucash/engine/gncCustomer.c
Bob-IT: Correct, they are used to to stop the tables that are in use being deleted.
I do not think it is possible to delete a 'Vendor/Customer/Employee'.
I am not at all clear at the moment on the relationship of 'Owner' to the other objects.

jralls: gncOwner is a struct consisting of a type indicator and a union of gncCustomer, gncEmployee, and gncVendor. That's called a 'discriminated union' and it's similar in concept to the boost::variant we use in KVP and the std::variant in c++options.

So when I say that the Tax Table is in use by an Owner I mean that it's in use by an instance of one of the classes accessible by the Owner union.

The important point is that the TaxTable's and Bill Term's reference count should be incremented when it's added to one of those things and decremented when it's either removed or the thing that references it is deleted from the book, *not* when the owning instance is created/destroyed at session start or end.

libgnucash/engine/Transaction.c
jralls: OIC, I was looking at the flow the wrong way. In fact it doesn't ever free any of the cloned splits. I also see where it skips if the original (i.e. not cloned) split isn't marked dirty.

I think from the memory leak standpoint your change is good enough, but the function's assumptions leaves a lot of room for it to do the wrong thing. I guess we're lucky that those cases are rare in real-life usage. Oh well. In the future we'll just discard the edited transaction and get a new instance from the database.

libgnucash/engine/gncCustomer.c
jralls: >  not when the owning instance is created/destroyed at session start or end.

Which, on reinspection, I see that this change does: It prevents the ref counts being decremented at session end.

jralls:
Thanks!

PR #1096, opened 2021-07-26
[report-core] sandbox running report to prevent freeze

christopherlam:
Guile has mechanism to run some code with strict _time_ and/or _memory_ limits. See https://www.gnu.org/software/guile/manual/html_node/Sandboxed-Evaluation.html

This change will ensure any report (built-in or custom) will be aborted if it exceeds 10 seconds. eg. adding `(sleep 10)` to a report renderer will display error message instead of an indeterminate freeze. This change could also be applied to `gnc:apply-with-error-handling` which will also trap qif-importer. `call-with-time-and-allocation-limits` may also be used to limit memory used too.

Thoughts?

image

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -275,12 +275,11 @@ include (${SWIG_USE_FILE})
 string(REGEX MATCH "^[0-9]+[.]" SWIG_MAJOR ${SWIG_VERSION})
 
 # Find Guile and determine which version we are using.
-# Look for guile versions in this order: 3.0 > 2.2 > 2.0
+# Look for guile versions in this order: 3.0 > 2.2.1
 
 # guile library and include dir
 pkg_check_modules (GUILE3 guile-3.0 QUIET)
-pkg_check_modules (GUILE22 guile-2.2 QUIET)
-pkg_check_modules (GUILE2 guile-2.0>=2.0.9 QUIET)
+pkg_check_modules (GUILE22 guile-2.2>=2.2.1 QUIET)
 
 if (GUILE3_FOUND) # found guile-3.0
   add_definitions (-DHAVE_GUILE30)
@@ -295,7 +294,7 @@ if (GUILE3_FOUND) # found guile-3.0
   message(STATUS "Using guile-3.0.x")
   find_program (GUILE_EXECUTABLE NAMES guile3.0 guile)
 
-elseif (GUILE22_FOUND) # found guile-2.2
+elseif (GUILE22_FOUND) # found guile-2.2.1
   add_definitions (-DHAVE_GUILE22)
   set(HAVE_GUILE2 TRUE)
   set(GUILE_EFFECTIVE_VERSION 2.2)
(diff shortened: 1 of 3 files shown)


jralls:
Sandboxed evaluation was introduced in Guile 2.2 so this has to go to master and requires bumping the minimum guile version in CMakeLists.txt and Readme.dependencies.

christopherlam:
Rebased to master, updated guile minimum to 2.2.1

jralls:
Looks like we need to update some of the CI systems too.

I don't much care for "deprecate" in the commit message, that implies that it's still possible to use 2.0 and that support for it will be removed later.

fellen:
> I don't much care for "deprecate" in the commit message, that implies that it's still possible to use 2.0 and that support for it will be removed later.

Shouldnt the depreciation go to maint and the dropping to master?

CMakeLists.txt
```diff
@@ -308,21 +307,8 @@ elseif (GUILE22_FOUND) # found guile-2.2
...
-  find_program (GUILD_EXECUTABLE NAMES guild2.0 guild)
-  if (NOT GUILD_EXECUTABLE)
-    message (SEND_ERROR "The guild executable was not found, but is required. Please set GUILD_EXECUTABLE.")
-  endif()
-  message(STATUS "Using guile-2.0.x")
-  find_program (GUILE_EXECUTABLE NAMES guile2.0 guile)
-
 else()
-  message (FATAL_ERROR "Neither guile 3.0, guile 2.2, nor guile 2.0 were found GnuCash can't run without one of them. Ensure that one is installed and can be found with pkg-config.")
+  message (FATAL_ERROR "Neither guile 3.0, nor guile 2.2.1 were found GnuCash can't run without one of them. Ensure that one is installed and can be found with pkg-config.")
```
fellen, line 311: … were found. GnuCash can't …

gnucash/report/report-core.scm
```diff
@@ -748,7 +749,10 @@ not found.")))
 ;; where captured-error is the error string.
 (define (gnc:render-report report)
   (define (get-report) (gnc:report-render-html report #t))
-  (gnc:apply-with-error-handling get-report '()))
+  (call-with-time-limit
+   10
+   (lambda () (gnc:apply-with-error-handling get-report '()))
+   (lambda () (list #f "Report exceeds time limit. Aborted."))))
```
fellen, line 755: Perhaps inform the user about the value of the time limit?

jralls:
> Shouldnt the depreciation go to maint and the dropping to master?

Another reason not to use that word. No, it's routine for us to upgrade minimum dependency versions for major releases.

CMakeLists.txt
```diff
@@ -275,12 +275,11 @@ include (${SWIG_USE_FILE})
...
 
 # Find Guile and determine which version we are using.
-# Look for guile versions in this order: 3.0 > 2.2 > 2.0
+# Look for guile versions in this order: 3.0 > 2.2.1
 
 # guile library and include dir
 pkg_check_modules (GUILE3 guile-3.0 QUIET)
-pkg_check_modules (GUILE22 guile-2.2 QUIET)
-pkg_check_modules (GUILE2 guile-2.0>=2.0.9 QUIET)
+pkg_check_modules (GUILE22 guile-2.2.1 QUIET)
```
jralls, line 282: This is failing CI because it requires *exactly* guile-2.2.1 and all of the VMs except Ubuntu 20.04. are on 2.2.9. You want to say `(GUILE22 guile-2.2>=2.2.1 QUIET)`.

gnucash/report/report-core.scm
jralls: Does this get displayed in the html tab? If not, something should. If so it should be a bit less robotic and marked for translation.

CMakeLists.txt
jralls: And `Neither guile 3.0 nor guile 2.2.1+ were found.` A comma is used only when there are more than two elements in the list.

gnucash/report/report-core.scm
```diff
@@ -748,7 +749,10 @@ not found.")))
 ;; where captured-error is the error string.
 (define (gnc:render-report report)
   (define (get-report) (gnc:report-render-html report #t))
-  (gnc:apply-with-error-handling get-report '()))
+  (call-with-time-limit
+   10
```
gjanssens, line 753: I'm concerned with the hard-coded time limit. Generating a complex report on less capable hardware may easily exceed this.
Some way for the user to control what happens should be provided.
My ideal case would be what Firefox currently does when a javascript takes too long to respond: it will inform the user via a popup and offers the option to either continue to wait or to abort. This is ideal because the long-running process is not reset by default. Only on user request.

christopherlam: Sure. In this case I do not think there's anything guile has to offer unless we allow multi threading in guile. This branch was mainly to illustrate a new guile capability. Additionally looking into guile sources, the sandbox uses sigalarm currently but they want to use thread, so, we shouldn't use it.

PR #1098, opened 2021-07-27
Bug798205 import OFX: mods re FITID (online_id)

goodvibes2:
Do NOT exclude transaction from 'Generic import transaction matcher' window just because a transaction exists with the same online_id (FITID).

Also do not drop a potential match just because it has an online_id - it may be a reimport.

I'm not sure I should have deleted all newly unused functions - please advise.


christopherlam:
IMHO removing dead code should be done more often.
The reason for excluding the trans with online_id would seem to handle OFX with internal transfers between >1 accounts in the same OFX. Any idea how this possibility would manifest?

jralls:
Removing `gnc_import_get_trans_online_id` means there's no longer any point to having `gnc_import_set_trans_online_id` so that should be removed as well.

jralls:
> The reason for excluding the trans with online_id would seem to handle OFX with internal transfers between >1 accounts in the same OFX.

No. I [explained this](https://bugs.gnucash.org/show_bug.cgi?id=798205#c16)(see the last para) on the bug report.

goodvibes2:
> Removing `gnc_import_get_trans_online_id` means there's no longer any point to having `gnc_import_set_trans_online_id` so that should be removed as well.

John, OK I'll remove `gnc_import_set_trans_online_id` also.

goodvibes2:
> The reason for excluding the trans with online_id would seem to handle OFX with internal transfers between >1 accounts in the same OFX. Any idea how this possibility would manifest?

Christopher Lam,
Has John's response answered your question?

Your question got me thinking that I have not tested having transactions for 2 accounts in the 1 .ofx file, with both accounts having a common transfer transaction between them.

I did some tests, and the 2nd transaction always shows in the 'Generic import transaction matcher' window as defaulting to being added. I also tested this in std GnuCash 4.5 and it did the same.
It seems the code does not find (or look for?) a match in the previous transactions imported for a different account in the same import session (Note: There is no facility to import multiple .ofx files in 1 import session.)

I suspect that this is something that is never done in practice, at least for ofx imports. Maybe some other type of import, which uses the 'Generic import transaction matcher' window, could have this problem but I don't recall anyone ever raising it. In any case, it is a different problem.
I'll ask in GnuCash-user if anyone does include transactions for multiple accounts in the 1 .ofx file to see if this needs to be followed up.

christopherlam:
> Has John's response answered your question?

It doesn't bother me at all. I am repeating the argument brought in to support the feature as written in https://github.com/Gnucash/gnucash/pull/721#discussion_r457118918 and fwiw I agree that we shouldn't scan all transaction splits for online_id, we should scan only imported account splits. My bank Westpac doesn't let me import >1 account for ofx so I'm not bothered.

goodvibes2:
> > Removing `gnc_import_get_trans_online_id` means there's no longer any point to having `gnc_import_set_trans_online_id` so that should be removed as well.
> 
> John, OK I'll remove `gnc_import_set_trans_online_id` also.

John, Done, tested and force pushed.

goodvibes2:
> It seems the code does not find (or look for?) a match in the previous transactions imported for a different account in the same import session (Note: There is no facility to import multiple .ofx files in 1 import session.)
> 
> I suspect that this is something that is never done in practice, at least for ofx imports. Maybe some other type of import, which uses the 'Generic import transaction matcher' window, could have this problem but I don't recall anyone ever raising it. In any case, it is a different problem.
> I'll ask in GnuCash-user if anyone does include transactions for multiple accounts in the 1 .ofx file to see if this needs to be followed up.

I asked in GnuCash-User, and there are people who have this problem. When I tried to create a bug for this, I found there is already one:
https://bugs.gnucash.org/show_bug.cgi?id=105334
      import multiple account OFX files with transfers between the accounts, duplicate transactions are generated

jralls:
>I asked in GnuCash-User, and there are people who have this problem. When I tried to create a bug for this, I found there is already one:
https://bugs.gnucash.org/show_bug.cgi?id=105334
import multiple account OFX files with transfers between the accounts, duplicate transactions are generated

This came up in #721 (@christopherlam thanks for reminding me about that PR, I'd forgotten about it). @jeanlaroche said that he'd tested it and it seemed to work; the testimony on the mailing list suggests likewise. IIUC that change (I didn't re-review the details) collected all of the imports in a file and posted them at once. Maybe it would be reasonable to that for every account instead so that the imported splits and their FITIDs from the first account are processed and available for matching in subsequent accounts.

goodvibes2:
> This came up in #721 (@christopherlam thanks for reminding me about that PR, I'd forgotten about it). @jeanlaroche said that he'd tested it and it seemed to work; the testimony on the mailing list suggests likewise. IIUC that change (I didn't re-review the details) collected all of the imports in a file and posted them at once. Maybe it would be reasonable to that for every account instead so that the imported splits and their FITIDs from the first account are processed and available for matching in subsequent accounts.

That sounds reasonable. IIUC, I'm thinking, we would do a 'Generic transaction matcher' window for the transactions in the first account in the ofx file, process and commit those transactions, then do another 'Generic transaction matcher' window (& commit) for each for the extra accounts in the ofx file. Does that sound reasonable?

I'd like to have a go a doing this, unless @jeanlaroche wants to?

Are you waiting until the https://bugs.gnucash.org/show_bug.cgi?id=105334 issue is resolved until you merge this PR and https://github.com/Gnucash/gnucash/pull/1069 ?
I'm thinking Bug 105334 should be a separate PR.

christopherlam:
The use case I'd specifically wish to check is: import bank account ofx, _then_ import credit card account ofx. If the bank has a transfer to credit card, then the second import would _ignore_ it as "already imported" which is harmless but annoys me mildly. I have no idea how multi account ofx imports work.

goodvibes2:
> The use case I'd specifically wish to check is: import bank account ofx, _then_ import credit card account ofx. If the bank has a transfer to credit card, then the second import would _ignore_ it as "already imported" which is harmless but annoys me mildly. I have no idea how multi account ofx imports work.

@christopherlam, I tested this, the 2nd imported transaction is shown in the 'Generic transaction matcher' window automatically flagged as either 'Update+Clear' or 'Clear' depending on the import preferences.

jralls:
>Are you waiting until the https://bugs.gnucash.org/show_bug.cgi?id=105334 issue is resolved until you merge this PR and #1069 ?
I'm thinking Bug 105334 should be a separate PR.

Oh, OK. I had the impression from your comments on the bug that you were going to pick that up here.

PR #1101, opened 2021-07-30
Account.cpp optimisations

christopherlam:
Continuation of #1086, thanks to valgrind

### DIFF
--- a/libgnucash/engine/Account.cpp
+++ b/libgnucash/engine/Account.cpp
@@ -2695,7 +2695,6 @@ static void
 account_foreach_descendant (const Account *acc, AccountCb thunk,
                             void* user_data, bool sort)
 {
-    gpointer result = nullptr;
     GList *children;
 
     g_return_if_fail (GNC_IS_ACCOUNT(acc));
@@ -3146,20 +3145,7 @@ gnc_account_foreach_descendant (const Account *acc,
                                 AccountCb thunk,
                                 gpointer user_data)
 {
-    const AccountPrivate *priv;
-    GList *node;
-    Account *child;
-
-    g_return_if_fail(GNC_IS_ACCOUNT(acc));
-    g_return_if_fail(thunk);
-
-    priv = GET_PRIVATE(acc);
-    for (node = priv->children; node; node = node->next)
-    {
-        child = static_cast<Account*>(node->data);
-        thunk(child, user_data);
-        gnc_account_foreach_descendant(child, thunk, user_data);
-    }
+    account_foreach_descendant (acc, thunk, user_data, FALSE);
 }
 
 gpointer
--- a/libgnucash/engine/Account.cpp
+++ b/libgnucash/engine/Account.cpp
@@ -2894,20 +2894,18 @@ gnc_account_nth_child (const Account *parent, gint num)
     return static_cast<Account*>(g_list_nth_data(GET_PRIVATE(parent)->children, num));
 }
 
+static void
+count_acct (Account *account, gpointer user_data)
+{
+    auto count {static_cast<int*>(user_data)};
+    ++*count;
+}
+
 gint
 gnc_account_n_descendants (const Account *account)
 {
-    AccountPrivate *priv;
-    GList *node;
-    gint count = 0;
-
-    g_return_val_if_fail(GNC_IS_ACCOUNT(account), 0);
-
-    priv = GET_PRIVATE(account);
-    for (node = priv->children; node; node = g_list_next(node))
-    {
-        count += gnc_account_n_descendants(static_cast<Account*>(node->data)) + 1;
-    }
+    int count {0};
+    account_foreach_descendant (account, count_acct, &count, FALSE);
     return count;
 }
 
--- a/libgnucash/engine/Account.cpp
+++ b/libgnucash/engine/Account.cpp
@@ -2951,52 +2951,27 @@ gnc_account_get_tree_depth (const Account *account)
     return depth + 1;
 }
 
+static void
+collect_acct (Account *account, gpointer user_data)
+{
+    auto listptr{static_cast<GList**>(user_data)};
+    *listptr = g_list_prepend (*listptr, account);
+}
+
 GList *
 gnc_account_get_descendants (const Account *account)
 {
-    AccountPrivate *priv;
-    GList *child, *descendants;
-
-    g_return_val_if_fail(GNC_IS_ACCOUNT(account), NULL);
-
-    priv = GET_PRIVATE(account);
-    if (!priv->children)
-        return NULL;
-
-    descendants = NULL;
-    for (child = priv->children; child; child = g_list_next(child))
-    {
-        descendants = g_list_append(descendants, child->data);
-        descendants = g_list_concat(descendants,
-                gnc_account_get_descendants(static_cast<Account const *>(child->data)));
-    }
-    return descendants;
+    GList* list = nullptr;
+    account_foreach_descendant (account, collect_acct, &list, FALSE);
+    return g_list_reverse (list);
 }
 
 GList *
 gnc_account_get_descendants_sorted (const Account *account)
 {
-    AccountPrivate *priv;
-    GList *child, *children, *descendants;
-
-    /* errors */
-    g_return_val_if_fail(GNC_IS_ACCOUNT(account), NULL);
-
-    /* optimizations */
-    priv = GET_PRIVATE(account);
-    if (!priv->children)
-        return NULL;
-
-    descendants = NULL;
-    children = g_list_sort(g_list_copy(priv->children), (GCompareFunc)xaccAccountOrder);
-    for (child = children; child; child = g_list_next(child))
-    {
-        descendants = g_list_append(descendants, child->data);
-        descendants = g_list_concat(descendants,
-                gnc_account_get_descendants_sorted(static_cast<Account const *>(child->data)));
-    }
-    g_list_free(children);
-    return descendants;
+    GList* list = nullptr;
+    account_foreach_descendant (account, collect_acct, &list, TRUE);
+    return g_list_reverse (list);
 }
 
 Account *
--- a/libgnucash/engine/Account.cpp
+++ b/libgnucash/engine/Account.cpp
@@ -262,34 +262,29 @@ gchar *gnc_account_name_violations_errmsg (const gchar *separator, GList* invali
     return message;
 }
 
+struct ViolationData
+{
+    GList *list;
+    const gchar *separator;
+};
+
+static void
+check_acct_name (Account *acct, gpointer user_data)
+{
+    auto cb {static_cast<ViolationData*>(user_data)};
+    auto name {xaccAccountGetName (acct)};
+    if (g_strstr_len (name, -1, cb->separator))
+        cb->list = g_list_prepend (cb->list, g_strdup (name));
+}
+
 GList *gnc_account_list_name_violations (QofBook *book, const gchar *separator)
 {
-    Account *root_account = gnc_book_get_root_account(book);
-    GList   *accounts, *node;
-    GList   *invalid_list = NULL;
-
-    g_return_val_if_fail (separator != NULL, NULL);
-
-    if (root_account == NULL)
-        return NULL;
-
-    accounts = gnc_account_get_descendants (root_account);
-    for (node = accounts; node; node = g_list_next(node))
-    {
-        Account *acct      = (Account*)node->data;
-        gchar   *acct_name = g_strdup ( xaccAccountGetName ( acct ) );
-
-        if ( g_strstr_len ( acct_name, -1, separator ) )
-            invalid_list = g_list_prepend ( invalid_list, (gpointer) acct_name );
-        else
-            g_free ( acct_name );
-    }
-    if (accounts != NULL)
-    {
-        g_list_free(accounts);
-    }
-
-    return invalid_list;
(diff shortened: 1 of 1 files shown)


christopherlam:
@jralls @derekatkins I'm afraid the account.cpp refactor would stay as is. I can't optimize it further.

libgnucash/engine/Account.cpp
```diff
@@ -2921,52 +2921,46 @@ gnc_account_get_tree_depth (const Account *account)
...
     g_return_val_if_fail(GNC_IS_ACCOUNT(account), NULL);
 
+    /* optimizations */
     priv = GET_PRIVATE(account);
     if (!priv->children)
         return NULL;
 
-    descendants = NULL;
-    for (child = priv->children; child; child = g_list_next(child))
+    children = g_list_copy (priv->children);
```
jralls, line 2938: move this into `if (sort)`, no need to copy the unstorted case.

libgnucash/engine/Account.cpp
```diff
@@ -2921,52 +2921,46 @@ gnc_account_get_tree_depth (const Account *account)
...
-                gnc_account_get_descendants(static_cast<Account const *>(child->data)));
+        auto acc = static_cast<Account *>(child->data);
+        auto acc_descendants = account_get_descendants_internal (acc, sort);
+        descendants = g_list_prepend (descendants, acc);
+        for (GList *n = acc_descendants; n; n = n->next)
+            descendants = g_list_prepend (descendants, n->data);
+        g_list_free (acc_descendants);
     }
-    return descendants;
+    g_list_free(children);
```
jralls, line 2950: This should be `if (sort)` too.

libgnucash/engine/Account.cpp
```diff
@@ -2921,52 +2921,46 @@ gnc_account_get_tree_depth (const Account *account)
...
+    return g_list_reverse (descendants);
```
jralls, line 2951: This goes to the calling functions so that it's not done on every recursion.

libgnucash/engine/Account.cpp
```diff
@@ -2921,52 +2921,46 @@ gnc_account_get_tree_depth (const Account *account)
...
-    GList *child, *children, *descendants;
-
-    /* errors */
-    g_return_val_if_fail(GNC_IS_ACCOUNT(account), NULL);
-
-    /* optimizations */
-    priv = GET_PRIVATE(account);
-    if (!priv->children)
-        return NULL;
+    return account_get_descendants_internal (account, FALSE);
```
jralls, line 2957:
```
return g_list_reverse (account_get_descendants_internal (account, FALSE));
```

jralls:
> I can't optimize it further.

But you're *so close*!

christopherlam:
I was planning to demonstrate, with screenshots, how the final order of accounts would differ if reversing at the end. But experimentally I was proven wrong. So, I've switched as described. :-/

The most optimised is still
```scheme
(define-public (acc-descendants node)
  (define (lp a b) (fold lp (cons a b) (get-children a)))
  (cdr (reverse (lp node '()))))
```
but this branch is a fair optimization.

christopherlam:
Hmm tests show I was right, the order does change significantly.

jralls:
>I was planning to demonstrate, with screenshots, how the final order of accounts would differ if reversing at the end. But experimentally I was proven wrong. So, I've switched as described. :-/

That's because reversing in every recursion would screw up the order. Consider the tree
```
1
   A
   B
   C
2
   D
   E
   F
3
   G
   H
   I
```
In the first iteration before the reverse you have
```C, B, A, 1``` and after the reverse ```1, A, B, C```.
The second call *prepends* ```2, D, E, F``` and the third ```3, G, H, I```  so before the final reverse you have
```3, G, H, I, 2, D, E, F, 1, A, B, C```. That's not what you want and neither is the reverse. Prepending unreversed lists results in ```I, H, G, 3, F, E, D, 2, C, B, A, 1```, and a single reverse produces the desired order.


> The most optimised is still
```
(define-public (acc-descendants node)
  (define (lp a b) (fold lp (cons a b) (get-children a)))
  (cdr (reverse (lp node '()))))
```

You could write it that way if you want. The GList form of fold is [`g_list_foreach(GList*, GFunc, gpointer)`](https://developer-old.gnome.org/glib/stable/glib-Doubly-Linked-Lists.html#g-list-foreach). The resulting machine code will be pretty much the same as what you have now. So, BTW, would be the Scheme code. Underneath any abstraction the computer has to traverse the source loop copying the nodes to the destination loop. The fewer traversals the better.

**Comment by christopherlam** [comment]
Ok done in a different method. Much more scrutiny required :-o

File: libgnucash/engine/Account.cpp [hunk 5]
```diff
@@ -3178,54 +3130,85 @@ gnc_account_foreach_child (const Account *acc,
...
+        g_list_free (children);
+
+    return result;
 }
 
-gpointer
-gnc_account_foreach_descendant_until (const Account *acc,
-                                      AccountCb2 thunk,
-                                      gpointer user_data)
+typedef struct
```
jralls, line 3170: This is C++, no need for `typedef` and the name goes up top:
```
struct CollectAccounts
```

jralls:
Looks pretty good and it's more expressive (that's a Good Thing). Plus it passes the unit tests which the previous version didn't.

libgnucash/engine/Account.cpp
christopherlam: > **Edit**: On the other hand there's no point in a one-element struct. Just pass the `GList*` directly.

That's the one thing I've been fighting and losing. How? There's `**` magic somewhere.

jralls: Something like
```
GList* list = nullptr;
account_foreach_descendant (account, (AccountCb2)collect_acct, static_cast<gpointer>(&list), FALSE, FALSE);
```
 and 
```
auto listptr{static_cast<GList**>(user_data)};
*llistptr = g_list_prepend (*listptr, account);
```
Untested so it might take a little fiddling.

christopherlam: Works😊 now `CountAccounts` could be eliminated too...

jralls: \o/

christopherlam:
This branch repurposed to discuss optimisations and code reuse in account.cpp.

jralls:
Same way:
```
static gpointer count_acct (Account *account, gpointer user_data)
 {
     int *count = static_cast<int*>(user_data);
     ++*count;
     return count; //Not used, just here to satisfy the AccountCb2 spec
 }

gint
 gnc_account_n_descendants (const Account *account)
 {
     int count{0};
     account_foreach_descendant (account, (AccountCb2)count_acct, &count, FALSE, FALSE);
     return count;
}
```

christopherlam:
Final state of branch. Will push in 24h.

libgnucash/engine/Account.cpp
```diff
@@ -2691,6 +2691,45 @@ DxaccAccountSetCurrency (Account * acc, gnc_commodity * currency)
 /********************************************************************\
 \********************************************************************/
 
+static gpointer
+account_foreach_descendant (const Account *acc, AccountCb2 thunk,
+                            gpointer user_data, gboolean sort, gboolean can_abort)
```
jralls, line 2696: Use C++ types: `void*` instead of `gpointer` and `bool` instead of `gboolean`.

libgnucash/engine/Account.cpp
```diff
@@ -2691,6 +2691,45 @@ DxaccAccountSetCurrency (Account * acc, gnc_commodity * currency)
...
+{
+    const AccountPrivate *priv;
+    gpointer result = nullptr;
+    GList *children;
+
+    g_return_val_if_fail (GNC_IS_ACCOUNT(acc), nullptr);
+    g_return_val_if_fail (thunk, nullptr);
+
+    priv = GET_PRIVATE(acc);
```
jralls, line 2705: In C++ we declare at first use except when first use is inside a scope and we'll need it outside of the scope later and only specify the type when required.  Prefer brace initialization and reserving `=` for assignment.

`auto priv{GET_PRIVATE(acc)};`

libgnucash/engine/Account.cpp
```diff
@@ -2691,6 +2691,45 @@ DxaccAccountSetCurrency (Account * acc, gnc_commodity * currency)
...
+        children = priv->children;
+
+    for (auto node = children; node; node = node->next)
+    {
+        auto child = static_cast<Account*>(node->data);
+        result = thunk (child, user_data);
+        if (can_abort && result) break;
+
+        result = account_foreach_descendant (child, thunk, user_data, sort, can_abort);
+        if (can_abort && result) break;
```
jralls, line 2721: This is going one step too far for code reuse. `for_each` means that a thunk's side effects are applied to every node in the collection. Make a separate `gnc_account_find_first` that returns the desired value and call the function argument `predicate`.

libgnucash/engine/Account.cpp
```diff
@@ -2691,6 +2691,45 @@ DxaccAccountSetCurrency (Account * acc, gnc_commodity * currency)
...
+    }
+
+    if (sort)
+        g_list_free (children);
+
+    return result;
+}
+
+/********************************************************************\
+\********************************************************************/
```
jralls, line 2731: These decorations don't add anything when the functions are short... and long functions are a code smell all on their own.

jralls (changes requested):
Nice work.

For more fun have a look at [std::function](https://en.cppreference.com/w/cpp/utility/functional/function) and [lambda expressions](https://en.cppreference.com/w/cpp/language/lambda) which takes a bit of drudgery out of writing thunks.

libgnucash/engine/Account.cpp
christopherlam: Will continue later. I presume `gnc_account_find_first` will call `account_foreach_descendant` with `can_abort` being `TRUE`.

jralls: It would be better to just separate them. `foreach` whatever should always touch every `each`, no aborting. There's also no need for the sort logic in `find_first` so the duped code is down to the precondition check, 1thunk,(child)` and the recursion loop. One flavor returns, the other maybe sorts. Not that much duplication.

christopherlam: Ok in this case `gnc_account_foreach_descendant_until` doesn't need to change that much, just needs using C++ syntax. Pushed modifying changes and will merge commits to appropriate counterparts if satisfactory.

christopherlam:
> Nice work.
> 
> For more fun have a look at [std::function](https://en.cppreference.com/w/cpp/utility/functional/function) and [lambda expressions](https://en.cppreference.com/w/cpp/language/lambda) which takes a bit of drudgery out of writing thunks.

In other words, C++ is learning from lisp 😂😂😂

jralls:
> In other words, C++ is learning from lisp 😂😂😂

Of course.

jralls (approved):

jralls:
Looks good. I'm amused that there was already a function for `find_if` even though it's not named well.

jralls:
I trust you'll squash `(fixup)+` into more meaningful commits with proper messages...

gjanssens:
@christopherlam I'm impressed with the pace at which you are picking up C/C++ ! :+1:

christopherlam:
> Hi @christopherlam & @jralls, didn't think this was worth filing a bug for, but at least worth mentioning.
> 
> I have a few python scripts that were previously running against 4.4, and I recently upgraded them to run against 4.8 and found they broke in a surprising way.
> 
> I _believe_ that I have tracked it down to a change in the behavior of `gnc_account_lookup_by_name` in this PR. Specifically the function comment says
> 
> ```
> /** The gnc_account_lookup_by_name() subroutine fetches the account by
>  *  name from the descendants of the specified account.  The immediate
>  *  children are searched first.  If there is no match,, then a
>  *  recursive search of all descendants is performed looking for a
>  *  match.
> ```
> 
> To me this implies a breadth first search, I think this PR has changed it to a depth first search. The practical result is that I had some accounts like
> 
> ```
> A
> +-D
>   +-B
> B
> +-E
> C
> +-F
> ```
> 
> Searching for `B` used to return the `B` at the same level as `A` and `C` but now returns `A.D.B`.
> 
> Might just be worth updating the comment if the behavior isn't depended on anywhere more significant.

This is true, and hope the new behaviour can be worked around. We'll amend the function comment soon.

christopherlam:
Alternatively I think the breadth-first lookup may be restored as follows

```cpp
gpointer
gnc_account_foreach_descendant_until (const Account *acc,
                                      AccountCb2 thunk,
                                      gpointer user_data)
{
    gpointer result {nullptr};

    g_return_val_if_fail (GNC_IS_ACCOUNT(acc), nullptr);
    g_return_val_if_fail (thunk, nullptr);

    auto priv{GET_PRIVATE(acc)};

    for (auto node = priv->children; node && !result; node = node->next)
        result = thunk (static_cast<Account*>(node->data), user_data);

    for (auto node = priv->children; node && !result; node = node->next)
        result = gnc_account_foreach_descendant_until (static_cast<Account*>(node->data), thunk, user_data);

    return result;
}
```

christopherlam:
@austindoupnik thanks for the report. It has been fixed in maint for 4.11.

austindoupnik:
@christopherlam thank you! i'll keep my scripts as are and wait for that release! :)

PR #1109, opened 2021-08-13
Maint speedup and leaks

christopherlam:
`qof_book_use_trading_accounts` is called a lot especially from register code which calls `xaccSplitGetOtherSplit`. Cache its value from kvp. Also, more leaks.

### DIFF
--- a/libgnucash/engine/Split.c
+++ b/libgnucash/engine/Split.c
@@ -2077,45 +2077,27 @@ xaccSplitMergePeerSplits (Split *split, const Split *other_split)
 Split *
 xaccSplitGetOtherSplit (const Split *split)
 {
-    int i;
     Transaction *trans;
-    int count, num_splits;
     Split *other = NULL;
-    gboolean lot_split;
-    gboolean trading_accts;
 
     if (!split) return NULL;
     trans = split->parent;
     if (!trans) return NULL;
 
-    trading_accts = xaccTransUseTradingAccounts (trans);
-    num_splits = xaccTransCountSplits(trans);
-    count = num_splits;
-    lot_split = qof_instance_has_slot(QOF_INSTANCE (split), "lot-split");
-    if (!lot_split && !trading_accts && (2 != count)) return NULL;
-
-    for (i = 0; i < num_splits; i++)
+    for (GList *n = xaccTransGetSplitList (trans); n; n = n->next)
     {
-        Split *s = xaccTransGetSplit(trans, i);
-        if (s == split)
-        {
-            --count;
+        Split *s = n->data;
+        if ((s == split) ||
+            (xaccAccountGetType (xaccSplitGetAccount (s)) == ACCT_TYPE_TRADING) ||
+            (qof_instance_has_slot (QOF_INSTANCE (s), "lot-split")))
             continue;
-        }
-        if (qof_instance_has_slot (QOF_INSTANCE (s), "lot-split"))
-        {
-            --count;
-            continue;
-        }
-        if (trading_accts &&
-            xaccAccountGetType(xaccSplitGetAccount(s)) == ACCT_TYPE_TRADING)
-        {
-            --count;
-            continue;
-        }
+
+        if (other)
+            return NULL;
+
         other = s;
     }
-    return (1 == count) ? other : NULL;
+    return other;
 }
 
 /********************************************************************\


christopherlam:
This patch will show how often `qof_book_use_trading_accounts` is called -- mine tallies 200+ calls for every keyboard event in the register.
```patch
modified   libgnucash/engine/qofbook.cpp
@@ -1047,6 +1047,7 @@ qof_book_get_default_gain_loss_acct_guid (QofBook *book)
 gboolean
 qof_book_use_trading_accounts (const QofBook *book)
 {
+    static int count {0};
     if (book->cached_use_trading_accounts == -1)
     {
         char *opt = NULL;
@@ -1058,6 +1059,7 @@ qof_book_use_trading_accounts (const QofBook *book)
         g_free (opt);
     }
 
+    printf ("\rcount = %d", count++);
     return (book->cached_use_trading_accounts == 1);
 }
```

jralls:
It would be more interesting and useful to know what's calling it 200 times. That suggests that it's in a loop and the right thing to do is to move the call outside of the loop.

christopherlam:
> It would be more interesting and useful to know what's calling it 200 times. That suggests that it's in a loop and the right thing to do is to move the call outside of the loop.

Afaiu it's because `xaccSplitGetOtherSplit` is called many times per register refresh to populate the Transfer column.

jralls:
Testing if the book uses trading accounts isn't really necessary and in some cases might result in the function failing: Turning off trading accounts doesn't remove existing trading account splits.

There are other gains to be made in that function, starting with getting the split list and iterating on it directly instead of looping on the index.

christopherlam:
Simplified.

jralls:
You took out the persistence for `user_symbol`. Just making a space for it in the struct doesn't save it to file. Easy to fix, just put the qof_instance_get/set in the init and commit functions. More work, slightly faster: Modify the backends to save/load to the KVP slot. Don't try to save it directly on the commodity, that has all sorts of compatibility headaches.

I don't think anything is gained by using the string cache for the user symbol. GObject properties are slow and painful to use. Let's not create any more of them.

christopherlam:
Ok. This is a suggestion commit -- either expand the struct or continue kvp (and GValue) getters which need to handle leaking strings. What would be the best approach?

jralls:
> Ok. This is a suggestion commit -- either expand the struct or continue kvp (and GValue) getters which need to handle leaking strings. What would be the best approach?

It's not *that* either/or. It's either access KVP when you need it or access KVP when you create the object. If the item is accessed frequently then the latter will be faster.

There's at present no way around the GValue getters to talk to KVP, it's the GObject way of mimicking generics. If you move the symbol to a member of the struct then as long as you don't make a property for it there's no GObject.

The string cache is a separate issue. If there's a good chance that there's more than one place that the string value of the `user_symbol` would otherwise be created then the string cache makes sense. Otherwise the string cache is unnecessary overhead, just `strdup` the symbol from the KVP's GValue and hand out `const char*` references to it.

christopherlam:
> There's at present no way around the GValue getters to talk to KVP, it's the GObject way of mimicking generics. If you move the symbol to a member of the struct then as long as you don't make a property for it there's no GObject.

So, thinking of all these char* properties eg trans->notes, trans->doclink etc... to plug the leaks properly means copying the char* into an expanded trans struct. There's no better way?

jralls:
> So, thinking of all these char* properties eg trans->notes, trans->doclink etc... to plug the leaks properly means copying the char* into an expanded trans struct. There's no better way?

You're confusing KVP and Properties, I guess because both use GValues. They're otherwise quite distinct but neither is inherently leaky. Let's back up a couple of steps: Where is the leak on trans->notes occurring?

jralls:
> I guess because both use GValues.

Clarification: KVP doesn't itself use GValues, `qof_instance_set/get` do because they need to abstract away type without C++'s template mechanism. GValues are a convenient way to do that because they're already widely used in GnuCash.

christopherlam:
I've just confirmed with `cd install/bin` `valgrind --tool=memcheck --leak-check=full  ./gnucash --logto stdout --extra 2> valgrind.txt`. Expand a txn to doubleline, enter a 5 letter transaction notes "notes", save, exit.

```
==8097== 6 bytes in 1 blocks are definitely lost in loss record 249 of 19,075
==8097==    at 0x4842839: malloc (in /usr/libexec/valgrind/vgpreload_memcheck-amd64-linux.so)
==8097==    by 0x4C4BDE8: g_malloc (in /usr/lib/x86_64-linux-gnu/libglib-2.0.so.0.6800.1)
==8097==    by 0x4C609E3: g_strdup (in /usr/lib/x86_64-linux-gnu/libglib-2.0.so.0.6800.1)
==8097==    by 0x5A14157: ??? (in /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0.6800.1)
==8097==    by 0x5DF4430: qof_instance_get_kvp (qofinstance.cpp:1099)
==8097==    by 0x5D232D0: xaccTransGetNotes (Transaction.c:2398)
==8097==    by 0x6FA3632: gnc_split_register_get_notes_entry (split-register-model.c:1161)
==8097==    by 0x6FCE9C8: gnc_table_get_entry_internal (table-allgui.c:257)
==8097==    by 0x6FCFBFB: gnc_table_move_cursor_internal (table-allgui.c:863)
==8097==    by 0x6FCFD51: gnc_table_move_cursor_gui (table-allgui.c:891)
==8097==    by 0x6FA11D8: gnc_split_register_load (split-register-load.c:785)
==8097==    by 0x6F8D4F7: gnc_ledger_display_refresh_internal (gnc-ledger-display.c:874)
```

Because `return g_value_get_string (&v)` means the GValue which was created in the `xaccTransGetNotes` scope is never unset properly, which means its char* is not freed. One approach is

```patch
@@ -2394,11 +2394,14 @@ const char *
 xaccTransGetNotes (const Transaction *trans)
 {
     GValue v = G_VALUE_INIT;
+    static char* retval = NULL;
     if (!trans) return NULL;
     qof_instance_get_kvp (QOF_INSTANCE (trans), &v, 1, trans_notes_str);
-    if (G_VALUE_HOLDS_STRING (&v))
-         return g_value_get_string (&v);
-    return NULL;
+    if (retval)
+        g_free (retval);
+    retval = (G_VALUE_HOLDS_STRING (&v) ? g_value_dup_string (&v) : NULL);
+    g_value_unset (&v);
+    return retval;
 }
```
but is not threadsafe. Another approach is:
```patch
@@ -2390,15 +2390,19 @@ xaccTransGetDocLink (const Transaction *trans)
     return NULL;
 }
 
-const char *
+/* must be freed by the caller */
+char *
 xaccTransGetNotes (const Transaction *trans)
 {
     GValue v = G_VALUE_INIT;
+    char* retval = NULL;
     if (!trans) return NULL;
     qof_instance_get_kvp (QOF_INSTANCE (trans), &v, 1, trans_notes_str);
-    if (G_VALUE_HOLDS_STRING (&v))
-         return g_value_get_string (&v);
-    return NULL;
+    if (retval)
+        g_free (retval);
+    retval = (G_VALUE_HOLDS_STRING (&v) ? g_value_dup_string (&v) : NULL);
+    g_value_unset (&v);
+    return retval;
 }
```

christopherlam:
Anyway what's left here is a speedup to xaccSplitGetOtherSplit avoiding all kvp access and with early return.

jralls:
> Anyway what's left here is a speedup to xaccSplitGetOtherSplit avoiding all kvp access and with early return.

Well, a speedup to xaccSplitGetOtherSplit thanks to traversing the list the right way and early return. You're still calling `qof_instance_get_slot`.

christopherlam:
> You're still calling `qof_instance_get_slot`.

This is unavoidable isn't it?

jralls:
It is, it's just that you claimed to be avoiding KVP access and you can't. It's not a big deal in terms of performance: You're just checking the existence of a slot which is a straight memory lookup. It's reading or writing to slots that's slow.

PR #1113, opened 2021-08-19
Maint leaks

christopherlam:
More valgrind leaks... continuation of #1107 

because `gtk_container_get_children` always returns a new GList* and they were not consistently freed after use.

PR #1114, opened 2021-08-20
More leaks

christopherlam:
More valgrind. Continuation #1113 

1. `GPtrArray` was freed but not its contents. Was causing numerous leaks. This change seems safe but cannot confirm.
2. Returning static string for `gnc_commodity_get_user_symbol` in a thread-unsafe strategy. Is the `const` in the function signature reliable in signalling the function callers that the returned pointer must not be freed?


jralls:
1. The pointers aren't assigned in cellblock.c so that's not the right place to free them. 

2 is OK.

PR #1119, opened 2021-08-24
Locking fixes for the XML backend

nomis:
The lock file is closed even if it's not open, because m_lockfd is not initialised. Initialise it and be consistent in how it is checked for validity. (Tested)

**The lock file is deleted on session_end even if it's not acquired.** Fix this so that we don't unlock without holding the lock. (Tested)

The "break lock" process ignores the lock but does not try to acquire it. Unlink and try to acquire it. (Untested)

The NFS link checking process unlinks the lock file even if the check failed. We may not hold the lock at this point so don't do this. (Untested)

### DIFF
--- a/libgnucash/backend/xml/gnc-xml-backend.cpp
+++ b/libgnucash/backend/xml/gnc-xml-backend.cpp
@@ -723,7 +723,7 @@ GncXmlBackend::get_file_lock (bool break_lock)
         {
             close (m_lockfd);
             m_lockfd = -1;
-            g_unlink (m_lockfile.c_str());
+            /* Don't unlink m_lockfile, it does not belong to us. */
             m_lockfile.clear();
         }
         return false;
@@ -745,7 +745,7 @@ GncXmlBackend::get_file_lock (bool break_lock)
             g_unlink (linkfile.str().c_str());
             close (m_lockfd);
             m_lockfd = -1;
-            g_unlink (m_lockfile.c_str());
+            /* Don't unlink m_lockfile, it does not belong to us. */
             m_lockfile.clear();
         }
         return false;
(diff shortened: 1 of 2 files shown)


libgnucash/backend/xml/gnc-xml-backend.cpp
```diff
@@ -149,7 +149,7 @@ GncXmlBackend::session_begin(QofSession* session, const char* new_uri,
 
     /* Set the lock file */
     m_lockfile = m_fullpath + ".LCK";
-    auto locked = get_file_lock();
+    auto locked = get_file_lock(mode == SESSION_BREAK_LOCK);
```
bhardwajs, line 152: Could this be `const auto`?

libgnucash/backend/xml/gnc-xml-backend.hpp
```diff
@@ -60,7 +60,7 @@ class GncXmlBackend : public QofBackend
     std::string m_dirname;
     std::string m_lockfile;
     std::string m_linkfile;
-    int m_lockfd;
+    int m_lockfd = -1;
```
bhardwajs, line 63: How about use a `constexpr int INVALID_FILE_HANDLE = -1` and use that to compare `m_lockfd`?

nomis: That's an unnecessarily level of complexity for an invalid file descriptor...

libgnucash/backend/xml/gnc-xml-backend.cpp
nomis: Yes but I wouldn't normally litter the code with extra `const` specifiers.

gjanssens:
> As for NFS, I think it pretty unlikely that more than a very few GnuCash users use it. They're far more likely to use SMB for local shares or proprietary cloud services like Dropbox and Google Drive. The link trick is unlikely to work for any of those, so I think it would make sense to just remove that code. @gjanssens do you agree?

Yes. I'd even look for a decent multi-platform  file locking library and use that instead of our own implementation, but that's out of scope for this PR. Dropping the link stuff to simplify is fine with me.

jralls:
I've pushed the first two commits from this PR and a new commit that removes the lock-file-hardlink-count check.

PR #1123, opened 2021-08-29
Fix to allow gnucash-cli to run custom reports defined in config-user.scm

jarasco:
Custom reports normally loaded from the user account for the GUI are not available when running using the command line interface.  Attempting to run a custom report from gnucash-cli results in errors:
WARN <gnc.scm> BUG DETECTED: Scheme exception raised in report options generator procedure named options-gen
OTHER <> bad value

This PR incorporates logic to have gnucash-cli load the user's config-user.scm when running reports.

### DIFF
--- a/gnucash/report/gnc-report.c
+++ b/gnucash/report/gnc-report.c
@@ -92,6 +92,15 @@ load_custom_reports_stylesheets(void)
         return;
     else is_user_config_loaded = TRUE;
 
+    update_message("loading user configuration");
+    {
+        gchar *config_filename;
+        config_filename = g_build_filename (gnc_userconfig_dir (),
+                                                "config-user.scm", (char *)NULL);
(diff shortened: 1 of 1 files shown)


gjanssens:
While it solves your immediate problem, this change is unfortunately not in line with the intent behind `load_custom_reports_stylesheets`. As it's part of the report system code, it should only do stuff strictly related to reporting. `config-user.scm` is not restricted to code related to reporting so it shouldn't be loaded from there.

On the other hand currently only report commands in gnucash-cli can be influenced  by the contents of config-user.scm so I understand your choice to put these changes here.

I have chosen to fix this slightly differently while at the same time cleaning up a few artefacts of the last redesign in this area. The net effect is the same: with current maint you can now run the home-made reports defined in config-user.scm from gnucash-cli.

PR #1128, opened 2021-09-02
Bug 798235 Reconciliation Ending Balance not recalculated

goodvibes2:
The 2nd and subsequent times Statement Date is updated, the Ending Balance
was not recalculated. This was because at end of gnc_start_recn_date_changed(),
gnc_start_recn_update_cb() was called when it should not have been.

This caused the user_set_value flag to be set True (indicating statement
end amount was manually entered - which it wasn't) after the date is
changed. This causes the ending balance to NOT be recalculated when the
end date is subsequently changed.

### DIFF
--- a/gnucash/gnome/window-reconcile.c
+++ b/gnucash/gnome/window-reconcile.c
@@ -313,8 +313,21 @@ recnRecalculateBalance (RecnWindow *recnData)
 }
 
 
+/* amount_edit_focus_out_cb
+ *   Callback on focus-out event for statement Ending Balance.
+ *   Sets the user_set_value flag true if the amount entered is
+ *   different to the calculated Ending Balance as at the default
+ *   Statement Date. This prevents the entered Ending Balance
+ *   being recalculated if the Statement Date is changed.
+ *
+ * Args:   widget         - Ending Balance widget
+ *         event          - event triggering this callback
+ *         data           - structure containing info about this
+ *                          reconciliation process.
+ * Returns:  False - propagate the event to the widget's parent.
+ */
 static gboolean
-gnc_start_recn_update_cb(GtkWidget *widget, GdkEventFocus *event,
+amount_edit_focus_out_cb(GtkWidget *widget, GdkEventFocus *event,
                          startRecnWindowData *data)
 {
     gnc_numeric value;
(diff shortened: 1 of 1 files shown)


goodvibes2:
Perhaps I should also rename gnc_start_recn_update_cb() to gnc_start_recn_end_bal_focus_out_cb()?

jralls:
> Perhaps I should also rename gnc_start_recn_update_cb() to gnc_start_recn_end_bal_focus_out_cb()?

Since it's not a public function it really shouldn't start with `gnc`. The best name either combines the widget name and event, `amount_edit_focus_out_cb`, or describes the function's purpose, `update_ending_balance`.

jralls:
This reverts part of the [fix](https://github.com/Gnucash/gnucash/commit/faaed371ed3f5bb1083284bd762e3d22d4b1216f#diff-dc0f7b84dc083ae34c8f4f48366721469b7d77b8c8b46200c7a7f155b39646d3) for [Bug 798144](https://bugs.gnucash.org/show_bug.cgi?id=798144).

Comparing the before-and-after values when the widget has keyboard focus is a reasonable way to determine that the user entered a number with the keyboard, but that reasonableness breaks if it's not in response to a focus-out. @Bob-IT which piece of `gnc_start_recn_update_cb` did you need for your bug 798144 fix?

Bob-IT:
It maybe I added that line automatically after the gnc_amount_edit_set_amount but as the amount used is from a function it should always be a valid amount and as such can be removed. 
I should have some time at the weekend to have a look at this and test.

Bob-IT:
OK, after some testing it looks like that line is not required and was probably added as stated above. From testing with the change, if an ending balance is entered and is valid it will block the update from the date change call back which is what is required. If the ending balance has not been changed or is invalid and the date is then changed, the ending balance will be updated.

goodvibes2:
Hi John,
I have added a commit to rename a couple of static functions as per your suggestion Sep 3rd and documented them better. Bob's comments have confirmed my change is OK. Ready for review please.

gnucash/gnome/window-reconcile.c
```diff
@@ -313,8 +313,22 @@ recnRecalculateBalance (RecnWindow *recnData)
...
+ *   different to the calculated Ending Balance as at the default   *
+ *   Statement Date. This prevents the entered Ending Balance       *
+ *   being recalculated if the Statement Date is changed.           *
+ *                                                                  *
+ * Args:   widget         - Ending Balance widget                   *
+ *         event          - event triggering this callback          *
+ *         data           - structure containing info about this    *
+ *                          reconciliation process.                 *
+ * Returns:  False - propagate the event to the widget's parent.    *
+\********************************************************************/
```
jralls, line 329: static function descriptions shouldn't go into the API documentation so please change `/**` to `/* "`to hide it from Doxygen. They're not required, though they're certainly harmless as long as they're not done as Doxygen comments.

jralls (changes requested):

gnucash/gnome/window-reconcile.c
goodvibes2: @jralls, Done. Sorry, live and learn.

goodvibes2:
Nudge, nudge... GnuCash 4.7 is coming up :-)

jralls (approved):

jralls:
Thanks!

PR #1129, opened 2021-09-05
798298 reimport ofx

jeanlaroche:
Reverted the change be6fb1abe that entirely removed the check for online_id in existing registry transactions. Added code to ensure that during the import of a transaction, splits in register transactions that have the same FITID are only ignored if the split in question corresponds to the account into which the transaction is imported.
This re-instate the behavior prior to be6fb1abe which allowed us to re-import the same file twice in a row with the second time showing no transactions to import. It also fixes bug 798298 in cases where the online_id of an imported transaction exists in the split of a register transaction, but corresponds to a different account (there no guarantee that FITID are unique from bank to bank).

What this does not fix is non-compliant OFX files in which FITIDs are reused in the same bank account.

### DIFF
--- a/gnucash/import-export/import-backend.c
+++ b/gnucash/import-export/import-backend.c
@@ -1091,24 +1091,6 @@ static gint check_trans_online_id(Transaction *trans1, void *user_data)
     }
 }
 
-static gint collect_trans_online_id(Transaction *trans, void *user_data)
-{
-    Split *split;
-    GHashTable *id_hash = user_data;
-    int i=0;
-    
-    const gchar* online_id = gnc_import_get_trans_online_id (trans);
-    if (online_id)
-        g_hash_table_add (id_hash, (void*) online_id);
-
-    for (GList *splits = xaccTransGetSplitList (trans); splits; splits = splits->next)
-    {
-        if (gnc_import_split_has_online_id (splits->data))
-            g_hash_table_add(id_hash, (void*) gnc_import_get_split_online_id (splits->data));
-    }
-    return 0;
-}
-
 /** Checks whether the given transaction's online_id already exists in
   its parent account. */
 gboolean gnc_import_exists_online_id (Transaction *trans, GHashTable* acct_id_hash)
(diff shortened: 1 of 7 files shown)


christopherlam:
It should be possible to create tests for OFX import. It may require deconstructing the matching logic into testable functions. See https://github.com/Gnucash/gnucash/blob/maint/libgnucash/app-utils/test/test-autoclear.cpp

gnucash/import-export/import-backend.c
```diff
@@ -1091,10 +1091,18 @@ static gint check_trans_online_id(Transaction *trans1, void *user_data)
     }
 }
 
+typedef struct _collect_trans_online_id_data
```
jralls, line 1094: No need to name the struct when you typedef the definition, you can just say
```
typedef struct
{
/* ... */
} collect_trans_online_id_data;
```

jralls (changes requested):
This adds a layer of complexity when what's actually required is removing a layer of complexity:

An Account has a split list. `xaccAccountForEachTransaction` traverses the split list and on every split `s` calls the callback with `split->parent`. The original version of `collect_trans_online_id` examined every split in the transaction for an online id and added those splits to the hash. Your change ignores every split in the transaction's split list except the one that belongs to the account. That split is `s`, the one it started with.

A much better fix would just traverse the Account's split list and hash the online ids.

jralls:
> It should be possible to create tests for OFX import. It may require deconstructing the matching logic into testable functions. See https://github.com/Gnucash/gnucash/blob/maint/libgnucash/app-utils/test/test-autoclear.cpp

See also [Working Effectively With Legacy Code](https://www.worldcat.org/title/working-effectively-with-legacy-code/oclc/1156991884&referer=brief_results).

jeanlaroche:
> 
> 
> This adds a layer of complexity when what's actually required is removing a layer of complexity:
> 
> An Account has a split list. `xaccAccountForEachTransaction` traverses the split list and on every split `s` calls the callback with `split->parent`. The original version of `collect_trans_online_id` examined every split in the transaction for an online id and added those splits to the hash. Your change ignores every split in the transaction's split list except the one that belongs to the account. That split is `s`, the one it started with.
> 
> A much better fix would just traverse the Account's split list and hash the online ids.

I didn't know there was a way to traverse the account's split list so indeed that's a better way.

gnucash/import-export/import-backend.c
jeanlaroche: yep I'm aware of that. Only replicating what I see in the existing code so I'm changing the code style.

jeanlaroche:
Followed J. Ralls' suggestion which simplified the code quite a bit. Thanks.

jeanlaroche:
I don't get why compilation fails on jenkins but runs on my mac. There seems to be a missing library path.

jralls (approved):

jralls:
Thanks!

PR #1132, opened 2021-09-09
Add code to handle ofx files that include transfers between accounts

jeanlaroche:
To solve the issue in this bug, I'm processing the transactions in the ofx file per account, instead of all at once. This means that the matcher dialog is re-run for each account that's present in the ofx file (instead of once with all transactions for all accounts).
With this, the second transfer transaction in the OFX file is recognized as the other side of the transfer, and becomes a match.
One thing I had to change was to connect to "destroy" instead of "response" in the match dialog for the callback that continues processing (either with reconciling, processing transactions for the second account, or processing the next ofx file). This was needed because the imported transactions had to be recorded into the register before processing the ones corresponding to the second account, for them to be then recognized as matches.
Note that transfers between more than 2 accounts should also be handled by the new code.
Finally: I could add code to only separate the ofx transactions per account when we recognize that there's a likely transfer. This means that in most cases all transactions will be imported in one shot (that was the previous behavior), except if GC detects a likely transfer (for example 2 transactions that have matching amounts, but correspond to two different accounts). 

This would be a refinement, not sure whether it's worth it.

### DIFF
--- a/gnucash/import-export/ofx/gnc-ofx-import.c
+++ b/gnucash/import-export/ofx/gnc-ofx-import.c
@@ -76,7 +76,10 @@ typedef struct _ofx_info
     struct OfxStatementData* statement;     // Statement, if any
     gboolean run_reconcile;                 // If TRUE the reconcile window is opened after matching.
     GSList* file_list;                      // List of OFX files to import
+    GList* trans_list;
 } ofx_info ;
+
+static void runMatcher(ofx_info* info, char * selected_filename, gboolean go_to_next_file);
 
 /*
 int ofx_proc_status_cb(struct OfxStatusData data)
@@ -900,7 +903,7 @@ int ofx_proc_transaction_cb(struct OfxTransactionData data, void *user_data)
         if (xaccTransCountSplits(transaction) > 0)
         {
             DEBUG("%d splits sent to the importer gui", xaccTransCountSplits(transaction));
-            gnc_gen_trans_list_add_trans (info->gnc_ofx_importer_gui, transaction);
+            info->trans_list = g_list_prepend(info->trans_list,transaction);
         }
         else
         {
@@ -1088,7 +1091,7 @@ gnc_ofx_process_next_file (GtkDialog *dialog, gpointer user_data)
     info->statement = NULL;
 
     // Done with the previous OFX file, process the next one if any.
-    info->file_list = g_slist_delete_link(info->file_list, info->file_list);
+    info->file_list = g_slist_delete_link (info->file_list, info->file_list);
     if (info->file_list)
         gnc_file_ofx_import_process_file (info);
     else
@@ -1098,14 +1101,21 @@ gnc_ofx_process_next_file (GtkDialog *dialog, gpointer user_data)
     }
 }
 
-
-// This callback is called when the user is done matching transactions.
 static void
-gnc_ofx_match_done (GtkDialog *dialog, gint response_id, gpointer user_data)
+gnc_ofx_match_done (GtkDialog *dialog, gpointer user_data)
 {
     ofx_info* info = (ofx_info*) user_data;
-
-    if (response_id == GTK_RESPONSE_OK && info->run_reconcile && info->statement)
+    
+    if (info->trans_list && g_list_length (info->trans_list))
+    {
+        // Re-run the match dialog if there are transactions remaining in our list (happens if several accounts
+        // exist in the same ofx).
+        info->gnc_ofx_importer_gui = gnc_gen_trans_list_new (GTK_WIDGET (info->parent), NULL, FALSE, 42, FALSE);
+        runMatcher (info, NULL, true);
+        return;
+    }
+    
+    if (info->run_reconcile && info->statement)
     {
         // Open a reconcile window.
         Account* account = gnc_import_select_account (gnc_gen_trans_list_widget(info->gnc_ofx_importer_gui),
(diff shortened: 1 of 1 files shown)


jralls (commented):
Cleaned up the for loop, the redundant glist test, and some white space in the merge commit.

jralls:
Thanks!

gnucash/import-export/ofx/gnc-ofx-import.c
```diff
@@ -1098,14 +1101,21 @@ gnc_ofx_process_next_file (GtkDialog *dialog, gpointer user_data)
...
-// This callback is called when the user is done matching transactions.
 static void
-gnc_ofx_match_done (GtkDialog *dialog, gint response_id, gpointer user_data)
+gnc_ofx_match_done (GtkDialog *dialog, gpointer user_data)
 {
     ofx_info* info = (ofx_info*) user_data;
-
-    if (response_id == GTK_RESPONSE_OK && info->run_reconcile && info->statement)
+    
+    if (info->trans_list && g_list_length (info->trans_list))
```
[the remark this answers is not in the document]
jeanlaroche: Can you not have a non-null GList* with glist->data = NULL?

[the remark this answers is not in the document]
jralls: Of course, NULL is a perfectly valid value for `glist->data`, but in that case `g_list_length` will return something other than 0.

gnucash/import-export/ofx/gnc-ofx-import.c
```diff
@@ -1142,6 +1159,70 @@ reconcile_when_close_toggled_cb (GtkToggleButton *togglebutton, ofx_info* info)
...
+    info->trans_list = trans_list_remain;
+    
+    // See whether the view has anything in it and warn the user if not.
+    if (gnc_gen_trans_list_empty (info->gnc_ofx_importer_gui))
+    {
+        gnc_gen_trans_list_delete (info->gnc_ofx_importer_gui);
+        if (info->num_trans_processed)
+        {
+            gchar* acct_name = gnc_get_account_name_for_register (first_account);
+            gnc_info_dialog (parent, _("OFX file '%s', imported transactions for account '%s'\n%d transactions processed, no transactions to match"),
```
fellen, line 1196: As [some of our weblate translators found](https://github.com/Gnucash/gnucash/pull/1144/files), this has broken thee string freeze.

fellen: @jeanlaroche
While translating myself, I found it hard to understand. Have a look at the right column of
https://hosted.weblate.org/translate/gnucash/gnucash/en/?&offset=3677#history
for some of the problems.

Perhaps it looks nicer and is easier to understand with a colon after the first and a full stop after the second line.

Can you submit a fix after the string freeze is lifted?

jeanlaroche: Sorry about that, I'm not 100% sure what I should have done to avoid breaking things. 
In any case, can you clarify a bit what you think would be clearer? Can you give me an example of the format you think would be easier to understand?

fellen: > Sorry about that, I'm not 100% sure what I should have done to avoid breaking things.

You wrote it before the string freeze and on merging @jralls didn't see the changed string.
Usually we make no msgmerge in in te frozen time, but a translator found an unmarkes string.
So it was a concatenation of rare events.

jralls:
> You wrote it before the string freeze 

*Way* before the string freeze: He wrote it in April 2020.

> I'm not 100% sure what I should have done to avoid breaking things.

Every user-facing string needs to be marked for translation with the _() macro. [Wiki I18N](https://wiki.gnucash.org/wiki/I18N) has the gory details.

> Also I don't know how to make it possible for translators to reorder variables shown. If someone can show me an example, I'll make the appropriate changes.

Unfortunately that's not possible in C or Scheme. It is possible in C++ and is briefly mentioned in the wiki article. The capability could be made available to C by writing an `extern "C"` wrapper function for `boost::format`, but I wonder if it's too obscure for translators to readily use it.

gnucash/import-export/ofx/gnc-ofx-import.c
fellen: Ah, I start to understand. 
If it is a message box, use natural language. Only for table cells use short terms.

How about:
OFX import from file %s into account %s: \n …

How it the relation between  processing and matching?
%d transactions were processed, but no matchs were found.
or what do you try to express?

jeanlaroche: The meaning of this message is
%d transactions were processed, all of which were previously imported

(so you don't need to match any, hence we won't show you the matching dialog) would be the very-long version!
Note that I changed the string, but did not remove the _() macro, so I'm not sure how the string change was not detected.

fellen: The history:
2021-09-10: your commit as PR
2021-09-13 my msgmerge after string freeze (without your change)
2021-09-16: John's  merge of your PR into maint
2021-09-20: my extraordinary msgmerge into the .po files (with your change)
2021-09-21: translators started to update their translation

Just try to formulate as message normal sentences, which the user will understand. In some cases you might want to add some context for us translators—we do not see it in the context of the process and related GUI elements. Then add a [translator comment](https://wiki.gnucash.org/wiki/I18N#Translator_Comments_for_Single_Messages) to tell us the context.

goodvibes2:
Jean,
I'm probably missing something, but where does ofx_info get freed if there are no transactions in matcher or you cancel out of matcher?

jralls:
> I need to fix that. Do I need to open a bug just for that tiny fix?

You don't need to open a bug for anything that you're going to write a PR for. Just write the code and explain in the commit message what it's for.

PR #1148, opened 2021-09-23
Change the wording of the dialog when no new transactions need to be …

jeanlaroche:
…reviewed


goodvibes2:
Hi Jean,
I cannot review but just a suggestion...
This is much better but won't this message potentially appear for each account in an OFX file? If so, it should be:
"While importing transactions from OFX file '%s' into account '%s', found %d previously imported transactions, no new transactions." I.e files -> file, transaction -> transactions
If this should be better english, I would suggest:
"While importing transactions from OFX file '%s' into account '%s', there were %d previously imported transactions, and no new transactions."

jeanlaroche:
Thanks for the correction about files. I think "no transaction" should be "no transaction" and not "no transactions" but since I"m not a native English speaker, I'll go with the flow! 
I don't know why this isn't building. It's building on my machine, the error has nothing to do with anything I did, hopefully it's a fluke, but I have my doubts!

jralls:
"no transaction" is pedantically more correct but "no transactions" more idiomatic in US and I guess Australian English.

The build failures were my fault, I fixed a bug caused by an API change in libofx without if-deffing it for the distros still shipping the older version. Your rebase on my push yesterday fixed it.

fellen:
> , there were %d previously imported transactions, and no new transactions."

has more semantic content andso is easier to understand.

PR #1150, opened 2021-09-24
General progress

christopherlam:
post string freeze

* When reversing a transaction which is already reversed, offer a jump to the reversal transaction.

### DIFF
--- a/gnucash/gnome/gnc-plugin-page-register.c
+++ b/gnucash/gnome/gnc-plugin-page-register.c
@@ -4733,6 +4733,7 @@ gnc_plugin_page_register_cmd_jump_linked_invoice (GtkAction* action,
     SplitRegister* reg;
     GncInvoice* invoice;
     Transaction *txn;
+    GtkWidget *window;
 
     ENTER ("(action %p, plugin_page %p)", action, plugin_page);
 
@@ -4741,6 +4742,7 @@ gnc_plugin_page_register_cmd_jump_linked_invoice (GtkAction* action,
     reg = gnc_ledger_display_get_split_register (priv->gsr->ledger);
     txn = gnc_split_register_get_current_trans (reg);
     invoice = invoice_from_split (gnc_split_register_get_current_split (reg));
+    window = GNC_PLUGIN_PAGE(plugin_page)->window;
 
     if (!invoice)
     {
@@ -4775,7 +4777,7 @@ gnc_plugin_page_register_cmd_jump_linked_invoice (GtkAction* action,
             }
             details = g_list_reverse (details);
             choice = gnc_choose_radio_option_dialog
-                (GNC_PLUGIN_PAGE (plugin_page)->window, _("Select document"),
+                (window, _("Select document"),
                  _("Several documents are linked with this transaction. \
 Please choose one:"), _("Select"), 0, details);
             if (choice >= 0)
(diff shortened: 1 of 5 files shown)


gnucash/gnome/gnc-plugin-page-register.c
```diff
@@ -4088,6 +4081,15 @@ gnc_plugin_page_register_cmd_reverse_transaction (GtkAction* action,
         return;
     }
 
+    new_trans = xaccTransGetReversedBy (trans);
+    if (new_trans)
+    {
+        if (!gnc_verify_dialog (GTK_WINDOW (window), TRUE, "%s",
+                               _("A reversing entry has already been created for this transaction. Jump to transaction?")))
```
fellen, line 4088: I wonder if we should split the declarative sentence and the question in different lines, perhaps separated by an empty line.
From my expirience that could be faster understood.

christopherlam: Splitting done

gnucash/gnome/gnc-plugin-page-register.c
```diff
@@ -4088,6 +4081,17 @@ gnc_plugin_page_register_cmd_reverse_transaction (GtkAction* action,
         return;
     }
 
+    new_trans = xaccTransGetReversedBy (trans);
+    if (new_trans)
+    {
+        const char *rev = _("A reversing entry has already been created for this transaction.");
+        const char *jump = _("Jump to transaction?");
```
fellen, line 4088: Article missing. ;-)

gnucash/gnome/gnc-plugin-page-register.c
```diff
@@ -4563,13 +4563,15 @@ gnc_plugin_page_register_cmd_stock_split (GtkAction* action,
...
+    GtkWindow* window;
 
     ENTER ("(action %p, plugin_page %p)", action, page);
 
     g_return_if_fail (GNC_IS_PLUGIN_PAGE_REGISTER (page));
 
     account = gnc_plugin_page_register_get_account (page);
-    gnc_stock_split_dialog (NULL, account);
+    window = gnc_window_get_gtk_window (GNC_WINDOW (GNC_PLUGIN_PAGE (page)->window));
+    gnc_stock_split_dialog (GTK_WIDGET (window), account);
```
christopherlam, line 4574: @Bob-IT would be good to know if there are possibly any negative consequences of setting parent type properly.

jralls: There are negative consequences of *not* setting the parent: Instead of the dialog appearing centered in the parent it appears at the root coordinates, usually the top-left corner of the display. On my two-monitor setup that can be quite far away from where I'm looking. If the parent happens to be occupying that space then switching away from GnuCash and back can paint the parent over the dialog, and if the dialog is modal that will make the user think that GnuCash has hung.

christopherlam: Ok. I was thinking more in the lines of having a window without a parent, activating the register, modifying data etc. I don't understand the `g_signal` mechanism and wondered if these codes would run inappropriately. I'll push this branch through after release.

jralls:
g_signal is like our qof_event: A signal hold a list of closures that have been registered to it. When the signal is run it goes down the list calling each closure-function in turn. The closure returns a boolean that decides whether the signal will keep traversing the list.

PR #1153, opened 2021-09-26
Bug 797928 - Add test case for basic scheduled transactions

nomis:
Test these scheduled transactions:
* 2 splits with fixed amounts "123" and "-123"
* 2 splits with fixed amounts "0"
* 2 splits with empty amounts ""

Verify that automatically created scheduled transactions exist.

It should be possible to extend this to reproduce https://bugs.gnucash.org/show_bug.cgi?id=798262 too.


nomis:
Wrong bug number

PR #1154, opened 2021-09-26
Bug 798262 - Add test case for basic scheduled transactions

nomis:
Test these scheduled transactions:
* 2 splits with fixed amounts "123"
* 2 splits with fixed amounts "0"
* 2 splits with empty amounts ""

Verify that automatically created scheduled transactions exist.

This is for [798262](https://bugs.gnucash.org/show_bug.cgi?id=798262) but it should be possible to extend this to reproduce [797928](https://bugs.gnucash.org/show_bug.cgi?id=797928) too.


nomis:
This is expected because the bug hasn't been fixed yet:
```
  4/116 Test   #4: test-sx ..........................***Failed    0.04 sec
FAILURE Automatically created transactions /home/runner/work/gnucash/gnucash/libgnucash/app-utils/test/test-sx.cpp:320 for make_one_empty_transaction: auto_created_txns = 0, expected_txns = 1
Executed 89 tests. There was 1 failure.
```

jralls:
Merged with edits.

PR #1157, opened 2021-10-02
Bug797678 OFX import should append not replace, existing Notes & Desc

goodvibes2:
Add an "Append" checkbox to the bottom of the "Generic import transaction
matcher" window to the left of the "Reconcile after match" checkbox.
When ticked, this causes the imported Description/Notes to be appended
to the matched transaction Description/Notes respectively.
The selected ticked/unticked state of the "Append" checkbox is saved in
a key value pair for the import account, so the next import for that
account will automatically default it to the saved state.
As these mods are limited to the code for the matcher window, this should
work for all the imports that use it - ie ofx & csv file imports (both
tested) & aqbanking (I cannot test).

Can some-one please test aqbanking?

If this gets merged, I'll see if I can also do this for QIF import, and update the documentation.

### DIFF
--- a/gnucash/import-export/import-backend.c
+++ b/gnucash/import-export/import-backend.c
@@ -95,6 +95,9 @@ struct _transactioninfo
 
     /* Reference id to link gnc transaction to external object. E.g. aqbanking job id. */
     guint32 ref_id;
+
+    /* When updating a matched transaction, append Description and Notes instead of replacing */
+    gboolean append_text;
 };
 
 /* Some simple getters and setters for the above data types. */
@@ -237,6 +240,15 @@ gnc_import_TransInfo_set_ref_id (GNCImportTransInfo *info,
 {
     g_assert (info);
     info->ref_id = ref_id;
+}
+
+
+void
+gnc_import_TransInfo_set_append_text (GNCImportTransInfo *info,
+                                      gboolean append_text)
+{
+    g_assert (info);
+    info->append_text = append_text;
 }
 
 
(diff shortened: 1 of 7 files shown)


goodvibes2:
Oops, I accidentally changed a comment and forgot to rebuild and test. I have fixed, rebuilt, tested and force pushed. Ready for review again please.

benoitg:
Looks good to me.  And since you implemented the check to see if the string is already there, I think that may be a good default for most users.

gnucash/import-export/import-backend.c
```diff
@@ -827,6 +839,90 @@ void split_find_match (GNCImportTransInfo * trans_info,
...
+static void
+update_desc_and_notes (const GNCImportTransInfo* trans_info)
+{
+    GNCImportMatchInfo* selected_match =
+            gnc_import_TransInfo_get_selected_match (trans_info);
+
+    if (trans_info->append_text)
+    {
+        // Append if imported desc not already in matched desc
+        if ( !g_strrstr (xaccTransGetDescription ( selected_match->trans),
```
jralls, line 908: Why `g_strrstr` instead of `strstr`? Also you should call `g_utf8_normalize()` on both strings before searching to ensure that the representations of complex characters are the same.
It might also be worthwhile to check that `g_utf8_strlen(selected_match->trans) >= g_utf8_strlen(trans_info)` before doing a potentially expensive substring search.

goodvibes2: @jralls,
Re g_strrstr: Totally agree, strstr would be better. I cannot remember how I decided on g_strrstr. I shall fix.

goodvibes2: Oops, last comment was not finished because I keep forgetting not to do Ctrl-Enter for a newline, as in another app I often use, as that is equivalent to clicking 'comment' button in github.

Re normalize: Good idea, I shall do.

Re checking g_utf_strlen:
Please correct me, but it seems to me parsing 2 strings into utf-8 characters in order to count them would be much more processing than doing the strstr() byte by byte comparison?

jralls: > Please correct me, but it seems to me parsing 2 strings into utf-8 characters in order to count them would be much more processing than doing the strstr() byte by byte comparison?

Well, if you normalize first then you can compare the `strlen`s instead, that would save having to decode the utf8 sequences into gunichars twice (once for the nomalization and again for the length calculation). Even faster would be to convert both strings to normalized UTF32 (i.e. `gunichar`), normalize, then count and search the UTF32 because it would reduce the number of comparisons in languages other than English. Unfortunately GLib doesn't provide that functionality so you'd have to use ICU. That extra complication isn't worth the effort.

`strlen` is faster than char-by-char comparisons of two strings because it needs only one register load per character and has a very simple branching structure. `strstr` OTOH can be quite expensive because it may have to backtrack if there are partial matches of 'needle` in `haystack`.

gnucash/import-export/import-backend.c
```diff
@@ -827,6 +839,90 @@ void split_find_match (GNCImportTransInfo * trans_info,
...
+
+/* Append or replace transaction description and notes
+ * depending on the Append checkbox
+ */
+static void
+update_desc_and_notes (const GNCImportTransInfo* trans_info)
+{
+    GNCImportMatchInfo* selected_match =
+            gnc_import_TransInfo_get_selected_match (trans_info);
+
```
Bob-IT, line 904: Why not add...
`Transaction* trans = gnc_import_TransInfo_get_trans (trans_info);`
here and then substitute where used below. Also you could add 'selected_match' and 'trans' as parameters to the four functions like `notes_replace (trans_info, selected_match, trans)` that way you only be retrieving those parameters once.

jralls:
Thanks!

PR #1160, opened 2021-10-05
[import-main-matcher] show destination acct for auto-match

christopherlam:
More import matcher hackery. For matched splits, the "Additional Comments" is improved to show the existing splits' peer accounts. The same information can be gleaned by expanding the child row, but I think having the account(s) full name being visible is a definite improvement.

image


gnucash/import-export/import-main-matcher.c
```diff
@@ -1421,6 +1422,29 @@ update_child_row (GNCImportMatchInfo *sel_match, GtkTreeModel *model, GtkTreeIte
...
+    gchar *retval, *name;
+    for (GList *n = xaccTransGetSplitList (xaccSplitGetParent (split)); n; n = n->next)
+    {
+        Account *account = xaccSplitGetAccount (n->data);
+        if ((n->data == split) ||
+            (xaccAccountGetType (account) == ACCT_TYPE_TRADING) ||
+            (g_list_find (accounts_seen, account)))
+            continue;
+        name = gnc_account_get_full_name (account);
+        names = g_list_prepend (names, g_strdup_printf ("\"%s\"", name));
```
fellen, line 1438: Needs translation as some languages use different quotation symbols.

PR #1161, opened 2021-10-06
gnc:format

christopherlam:
better than `format`

```scheme
(gnc:format "${document} from ${owner}"
 'owner "customer"
 'document "invoice")
```

will return

`"invoice from customer"`

### DIFF
--- a/bindings/guile/core-utils.scm
+++ b/bindings/guile/core-utils.scm
@@ -32,6 +32,7 @@
 (use-modules (srfi srfi-26))
 (use-modules (ice-9 match))
 (use-modules (ice-9 i18n))
+(use-modules (ice-9 regex))
 
 (export N_)
 (export G_)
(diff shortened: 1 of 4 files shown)


jralls:
This handles only string arguments, i.e. `~a` or `~s`, right?

christopherlam:
well ~a and ~s can handle any scm object -- this code will handle string values only (thanks to the `(? string? value)` match pattern.

jralls:
True but the result might not be what you want or expect particularly with numbers. That might be a corner-case for translatable strings and this makes it a lot easier on translators.

jralls:
Converting all of the msgids is going to be a massive re-translation effort. We can automate the weblate languages but the TP translators maintain their own po files and ignore any changes we make. @fellen any thought on how/when to roll this out?

christopherlam:
> True but the result might not be what you want or expect particularly with numbers. That might be a corner-case for translatable strings and this makes it a lot easier on translators.

Ok. The motivation for limiting to strings now was to narrow the input. Anyway, I've modified to convert all to string. The latest change is to improve error reporting especially if translator doesn't insert a valid ${symbol}.

christopherlam:
As @fellen has requested this now has tests. I don't think all ~a or ~s need to be overhauled. Only the difficult strings could be modified.

PR #1165, opened 2021-10-11
hide prices==1 where amount==0

christopherlam:
and a few cleanups/speedups.

### DIFF
--- a/gnucash/gnome/reconcile-view.c
+++ b/gnucash/gnome/reconcile-view.c
@@ -313,6 +313,7 @@ gnc_reconcile_view_new (Account *account, GNCReconcileViewType type,
     GList            *accounts = NULL;
     GList            *splits;
     Query            *query;
+    QofNumericMatch   sign;
 
     g_return_val_if_fail (account, NULL);
     g_return_val_if_fail ((type == RECLIST_DEBIT) ||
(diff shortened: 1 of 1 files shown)


gnucash/register/ledger-core/split-register-model.c
```diff
@@ -1389,11 +1389,15 @@ gnc_split_register_get_price_entry (VirtualLocation virt_loc,
...
     split = gnc_split_register_get_split (reg, virt_loc.vcell_loc);
 
+    amount = xaccSplitGetAmount (split);
+    if (gnc_numeric_zero_p (amount))
+        return NULL;
+
     price = xaccSplitGetSharePrice (split);
-    curr = xaccTransGetCurrency (xaccSplitGetParent (split));
     if (gnc_numeric_zero_p (price))
         return NULL;
```
jralls, line 1398: Might it be clearer if it was
```
   if (gnc_numeric_zero_p(xaccSplitGetAmount(split)) || gnc_numeric_zero_p(xaccSplitGetValue(split)))
```
?

gnucash/register/ledger-core/split-register-model.c
```diff
@@ -1389,8 +1389,8 @@ gnc_split_register_get_price_entry (VirtualLocation virt_loc,
 
     split = gnc_split_register_get_split (reg, virt_loc.vcell_loc);
 
-    amount = xaccSplitGetAmount (split);
-    if (gnc_numeric_zero_p (amount))
+    if (gnc_numeric_zero_p (xaccSplitGetAmount(split)) ||
+        gnc_numeric_zero_p (xaccSplitGetValue(split)))
         return NULL;
```
jralls: With this change the test of `gnc_numeric_zero_p(price)` below is redundant and should be removed.

gnucash/gnome/reconcile-view.c
```diff
@@ -346,14 +347,11 @@ gnc_reconcile_view_new (Account *account, GNCReconcileViewType type,
     g_list_free (accounts);
 
     /* limit the matches to CREDITs and DEBITs only, depending on the type */
```
jralls, line 349: Drop the comment, it's no longer true.

gnucash/report/reports/standard/taxinvoice.scm
```diff
@@ -87,9 +87,9 @@
 (define optname-jobnumber-show		(N_ "Show Job number"))
 (define optname-netprice		(N_ "Show net price"))
 (define optname-invnum-next-to-title	(N_ "Invoice number next to title"))
-(define optname-border-collapse		(N_ "table-border-collapse"))
-(define optname-border-color-th		(N_ "table-header-border-color"))
-(define optname-border-color-td		(N_ "table-cell-border-color"))
+(define optname-border-collapse		"table-border-collapse")
+(define optname-border-color-th		"table-header-border-color")
+(define optname-border-color-td		"table-cell-border-color")
```
jralls, line 92: Nope, these are option names in the Display tab. They need to be translated. They'd look nicer capitalized, as would the row and column ones above.

PR #1166, opened 2021-10-12
More leaks

christopherlam:
For the Saved Reports dialog.
Free data when closing dialog, freeing still does not trigger when pressing Alt-F4.

### DIFF
--- a/gnucash/gnome/dialog-custom-report.c
+++ b/gnucash/gnome/dialog-custom-report.c
@@ -80,6 +80,7 @@ typedef struct _CustomReportDialog
 
 } CustomReportDialog;
 
+void custom_report_dialog_destroy_cb (GtkWidget* widget, gpointer data);
 void custom_report_dialog_close_cb(GtkWidget* widget, gpointer data);
 void custom_report_help_cb(GtkWidget* widget, gpointer data);
 void close_custom_report_clicked_cb(GtkWidget* widget, gpointer data);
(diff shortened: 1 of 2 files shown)


jralls (changes requested):
The second commit should be two commits, one for emptying the list store and one for handling the escape key.

gnucash/gnome/dialog-custom-report.c
```diff
@@ -508,6 +517,19 @@ custom_report_query_tooltip_cb (GtkTreeView  *view,
     return FALSE;
 }
 
+static gboolean
+custom_report_event_cb (GtkWidget *widget, GdkEventKey *event,
+                        gpointer user_data)
+{
+    if (event->keyval == GDK_KEY_Escape)
```
jralls, line 524: Why does the commit message say Alt-F4 and the code is for &lt;Esc&gt;?

gnucash/gnome/dialog-custom-report.c
```diff
@@ -94,14 +95,40 @@ gboolean custom_report_query_tooltip_cb (GtkTreeView  *view,
...
+                             NULL);
+     gtk_list_store_clear (GTK_LIST_STORE (model));
+}
+
+void
+custom_report_dialog_destroy_cb (GtkWidget* widget, gpointer data)
+{
+    CustomReportDialog *crd = data;
+    empty_tree_model (gtk_tree_view_get_model (GTK_TREE_VIEW(crd->reportview)));
+    gtk_widget_destroy (crd->dialog);
```
Bob-IT, line 121: I do not think you need this gtk_widget_destroy here, when the close is pressed or 'X' the close callback should call the destroy function which then calls the destroy callback. In that you are then trying to destroy again but crd is freed and so should I think crash. When you 'Alt+F4' which destroys the dialog and as such calls the destroy callback and so again tries to destroy the dialog again and may crash. This may be down to timing and would have to pull your changes and give it a try.

PR #1178, opened 2021-10-22
WIP autoclear upgrade

christopherlam:
upgrade autoclear.... can handle autoclear amounts eg splits $10 $10 $10 can clear $30 easily.

~@cristiklein would need amending test-autoclear.~ Offshoot from #811.

todo:
- [x] very leaky
- [x] reinstate limits on knapsack hash size
- [x] ~can separate knapsack generation and query and reinstate live feedback~
- [x] cache `nc_list` the account-specific list of uncleared splits. Will be reused when modifying date & autoclear toggle. If account has 1000s of cleared, and <100 uncleared splits, should speed up startRecnWindow as the user changes/tweaks the statement date.

### DIFF
--- a/gnucash/gnome-utils/gnc-autoclear.cpp
+++ b/gnucash/gnome-utils/gnc-autoclear.cpp
@@ -25,6 +25,7 @@
 
 #include <gtk/gtk.h>
 #include <glib/gi18n.h>
+#include <vector>
 
 #include "inttypes.h"
 #include "Account.h"
@@ -58,22 +59,22 @@ autoclear_quark (void)
     return g_quark_from_static_string ("autoclear");
 }
 
-typedef struct
+struct WorkItem
 {
-    GArray *workarray;
+    gint64 reachable_amount;
+    GList *list_of_splits;
+};
+
+struct sack_data
+{
+    std::vector<WorkItem*> workvector;
     GHashTable *sack;
     Split *split;
     gboolean debugging_enabled;
     guint nc_progress;
     guint nc_list_length;
     GtkLabel *label;
-} sack_data;
-
-typedef struct
-{
-    gint64 reachable_amount;
-    GList *list_of_splits;
-} WorkItem;
+};
 
 static GList *DUP_LIST;
 
@@ -125,7 +126,7 @@ sack_foreach_func (gint64 thisvalue, GList *splits, sack_data *data)
     gint64 new_value = thisvalue + itemval;
     WorkItem *item = make_workitem (data->sack, new_value, data->split, splits);
 
-    g_array_append_val (data->workarray, item);
+    data->workvector.emplace_back (item);
     looping_update_status (data);
 }
 
@@ -223,20 +224,18 @@ gnc_autoclear_get_splits (Account *account, gnc_numeric toclear_value,
         gnc_numeric amount = xaccSplitGetAmount (split);
         WorkItem *item = make_workitem (sack, amount.num, split, NULL);
         size_t work_size = g_hash_table_size (sack) + 1;
-        GArray *workarray = g_array_sized_new
-            (FALSE, FALSE, sizeof (WorkItem *), work_size);
-        sack_data s_data = { workarray, sack, split,
+        sack_data s_data = { {}, sack, split,
             debugging_enabled, nc_progress, nc_list_length, label };
 
-        g_array_append_val (workarray, item);
+        s_data.workvector.reserve (work_size);
+        s_data.workvector.emplace_back (item);
         g_hash_table_foreach (sack, (GHFunc) sack_foreach_func, &s_data);
-        for (int i = 0; i < (int)work_size; i++)
+
+        for (auto& i : s_data.workvector)
         {
-            WorkItem *a_item = g_array_index (workarray, WorkItem*, i);
-            process_work (a_item, &s_data);
-            g_free (a_item);
+            process_work (i, &s_data);
+            g_free (i);
         }
-        g_array_free (workarray, TRUE);
         nc_progress++;
         if (G_UNLIKELY ((clock () - start_ticks) > MAX_AUTOCLEAR_TICKS))
         {
--- a/gnucash/gnome-utils/gnc-autoclear.cpp
+++ b/gnucash/gnome-utils/gnc-autoclear.cpp
@@ -139,13 +139,13 @@ gnc_autoclear_get_splits (Account *account, gnc_numeric toclear_value,
 
     if (gnc_numeric_zero_p (toclear_value))
     {
-        g_set_error (error, autoclear_quark, AUTOCLEAR_NOP,
+        g_set_error (error, autoclear_quark, AUTOCLEAR_NOP, "%s",
                      _("Account is already at Auto-Clear Balance."));
         goto skip_knapsack;
     }
     else if (nc_vector.empty ())
     {
-        g_set_error (error, autoclear_quark, AUTOCLEAR_NOP,
+        g_set_error (error, autoclear_quark, AUTOCLEAR_NOP, "%s",
                      _("No uncleared splits found."));
         goto skip_knapsack;
     }
@@ -174,7 +174,7 @@ gnc_autoclear_get_splits (Account *account, gnc_numeric toclear_value,
             gint64 new_value;
             if (__builtin_add_overflow (map_value, amount.num, &new_value))
             {
-                g_set_error (error, autoclear_quark, AUTOCLEAR_OVERLOAD,
+                g_set_error (error, autoclear_quark, AUTOCLEAR_OVERLOAD, "%s",
                              "Overflow error: Amount numbers are too large!");
                 goto skip_knapsack;
             }
@@ -196,7 +196,7 @@ gnc_autoclear_get_splits (Account *account, gnc_numeric toclear_value,
 
         if (G_UNLIKELY ((clock () - start_ticks) > MAX_AUTOCLEAR_TICKS))
         {
-            g_set_error (error, autoclear_quark, AUTOCLEAR_OVERLOAD,
+            g_set_error (error, autoclear_quark, AUTOCLEAR_OVERLOAD, "%s",
                          _("Too many uncleared splits"));
             goto skip_knapsack;
         }
@@ -205,7 +205,7 @@ gnc_autoclear_get_splits (Account *account, gnc_numeric toclear_value,
         if (G_UNLIKELY (try_toclear != sack.end() &&
                         try_toclear->second.empty()))
         {
-            g_set_error (error, autoclear_quark, AUTOCLEAR_MULTIPLE,
+            g_set_error (error, autoclear_quark, AUTOCLEAR_MULTIPLE, "%s",
                          _("Cannot uniquely clear splits. Found multiple possibilities."));
             goto skip_knapsack;
         }
@@ -227,7 +227,7 @@ gnc_autoclear_get_splits (Account *account, gnc_numeric toclear_value,
     /* Check solution */
     if (sack.find (toclear_value.num) == sack.end())
     {
-        g_set_error (error, autoclear_quark, AUTOCLEAR_UNABLE,
+        g_set_error (error, autoclear_quark, AUTOCLEAR_UNABLE, "%s",
                      _("The selected amount cannot be cleared."));
         goto skip_knapsack;
     }
--- a/gnucash/gnome-utils/gnc-autoclear.cpp
+++ b/gnucash/gnome-utils/gnc-autoclear.cpp
@@ -63,12 +63,12 @@ autoclear_quark (void)
 struct WorkItem
 {
     gint64 reachable_amount;
-    GList *list_of_splits;
-    WorkItem (const gint64 amount, GList *splits)
-        : reachable_amount (amount), list_of_splits(splits){}
+    std::vector<Split*> splits_vector;
+    WorkItem (const gint64 amount, std::vector<Split*> splits)
+        : reachable_amount (amount), splits_vector(splits){}
 };
 
-static GList *DUP_LIST;
+static std::vector<Split*> DUP_VEC;
 
 static void status_update (GtkLabel *label, gchar *status)
 {
@@ -104,7 +104,7 @@ gnc_autoclear_get_splits (Account *account, gnc_numeric toclear_value,
 {
     GList *toclear_list = NULL;
     std::vector<Split*> nc_vector;
-    std::unordered_map<gint64, GList*> sack;
+    std::unordered_map<gint64, std::vector<Split*>> sack;
     guint nc_progress = 0;
     clock_t start_ticks;
     gboolean debugging_enabled = qof_log_check (G_LOG_DOMAIN, QOF_LOG_DEBUG);
@@ -112,7 +112,7 @@ gnc_autoclear_get_splits (Account *account, gnc_numeric toclear_value,
     g_return_val_if_fail (GNC_IS_ACCOUNT (account), FALSE);
     g_return_val_if_fail (splits != NULL, FALSE);
 
-    DUP_LIST = g_list_prepend (NULL, NULL);
+    DUP_VEC = { nullptr };
     *splits = nullptr;
 
     /* Extract which splits are not cleared and compute the amount we have to clear */
@@ -155,35 +155,34 @@ gnc_autoclear_get_splits (Account *account, gnc_numeric toclear_value,
         std::vector<WorkItem> workvector = {};
 
         workvector.reserve (sack.size () + 1);
-        workvector.emplace_back (amountnum,
-                                 sack.find (amountnum) != sack.end () ?
-                                 DUP_LIST : g_list_prepend (nullptr, split));
+        if (sack.find (amountnum) != sack.end ())
+            workvector.emplace_back (amountnum, DUP_VEC);
+        else
+        {
+            std::vector<Split*> newvec = { split };
+            workvector.emplace_back (amountnum, newvec);
+        }
+
         // printf ("new split. sack size = %ld\n", sack.size());
 
-        for (auto& [thisvalue, splits] : sack)
+        for (auto& [thisvalue, thissplits] : sack)
         {
             gint64 new_value = thisvalue + amountnum;
-            workvector.emplace_back
-                (new_value,
-                 (splits == DUP_LIST || sack.find (new_value) != sack.end ())
-                 ? DUP_LIST
-                 : g_list_prepend (g_list_copy (splits), split));
+            if (thissplits == DUP_VEC || sack.find (new_value) != sack.end ())
+                workvector.emplace_back (new_value, DUP_VEC);
+            else
+            {
+                auto newvec = thissplits;
+                newvec.emplace_back (split);
+                workvector.emplace_back (new_value, newvec);
+            }
             looping_update_status (label, nc_progress, nc_vector.size (),
                                    sack.size ());
         }
 
         for (auto& item : workvector)
-        {
-            auto existing = sack.find (item.reachable_amount);
-            if (existing != sack.end())
-            {
-                if (debugging_enabled)
-                    DEBUG ("removing existing for %" PRId64 "\n", item.reachable_amount);
-                if (existing->second != DUP_LIST)
-                    g_list_free (existing->second);
-            }
-            sack[item.reachable_amount] = item.list_of_splits;
-        }
+            sack[item.reachable_amount] = item.splits_vector;
+
         nc_progress++;
         if (G_UNLIKELY ((clock () - start_ticks) > MAX_AUTOCLEAR_TICKS))
         {
@@ -193,7 +192,7 @@ gnc_autoclear_get_splits (Account *account, gnc_numeric toclear_value,
         }
 
         auto try_toclear = sack.find (toclear_value.num);
-        if (try_toclear != sack.end() && try_toclear->second == DUP_LIST)
+        if (try_toclear != sack.end() && try_toclear->second == DUP_VEC)
         {
             g_set_error (error, autoclear_quark (), AUTOCLEAR_MULTIPLE,
                          _("Cannot uniquely clear splits. Found multiple possibilities."));
@@ -225,18 +224,17 @@ gnc_autoclear_get_splits (Account *account, gnc_numeric toclear_value,
     }
     else
         /* copy GList because std::unordered_map value will be freed */
-        *splits = g_list_copy (sack[toclear_value.num]);
+    {
+        *splits = nullptr;
+        for (auto& s : sack[toclear_value.num])
+            *splits = g_list_prepend (*splits, s);
+    }
 
  skip_knapsack:
 
-    for (const auto& [thisvalue, thissplits] : sack)
-        if (thissplits != DUP_LIST)
-            g_list_free (thissplits);
-    g_list_free (DUP_LIST);
-
     status_update (label, NULL);
 
-    return (*splits != NULL);
+    return (*splits != nullptr);
 }
 
 
--- a/gnucash/gnome-utils/gnc-autoclear.cpp
+++ b/gnucash/gnome-utils/gnc-autoclear.cpp
@@ -25,7 +25,6 @@
 
 #include <gtk/gtk.h>
 #include <glib/gi18n.h>
-#include <memory>
 #include <vector>
 #include <unordered_map>
 
@@ -58,8 +57,8 @@ typedef enum
 struct WorkItem
 {
     gint64 reachable_amount;
-    std::vector<std::shared_ptr<Split>> splits_vector;
-    WorkItem (const gint64 amount, std::vector<std::shared_ptr<Split>> splits)
+    std::vector<Split*> splits_vector;
+    WorkItem (const gint64 amount, std::vector<Split*> splits)
         : reachable_amount (amount), splits_vector(splits){}
 };
 
@@ -111,9 +110,8 @@ gnc_autoclear_get_splits (Account *account, gnc_numeric toclear_value,
                           time64 end_date,
                           GList **splits, GError **error, GtkLabel *label)
 {
-    std::vector<std::shared_ptr<Split>> nc_vector, DUP_VEC;
-    std::unordered_map<gint64, std::vector<std::shared_ptr<Split>>,
-                       Hasher, EqualFn> sack;
+    std::vector<Split*> nc_vector, DUP_VEC;
+    std::unordered_map<gint64, std::vector<Split*>, Hasher, EqualFn> sack;
     guint nc_progress = 0;
     clock_t start_ticks, next_update_tick;
     gboolean debugging_enabled = qof_log_check (G_LOG_DOMAIN, QOF_LOG_DEBUG);
@@ -141,7 +139,7 @@ gnc_autoclear_get_splits (Account *account, gnc_numeric toclear_value,
                  xaccTransGetDate (xaccSplitGetParent (split)) > end_date)
             DEBUG ("skipping split after statement_date %p", split);
         else
-            nc_vector.emplace_back (std::make_shared<Split>(split));
+            nc_vector.emplace_back (split);
     }
 
     if (gnc_numeric_zero_p (toclear_value))
@@ -161,7 +159,7 @@ gnc_autoclear_get_splits (Account *account, gnc_numeric toclear_value,
 
     for (auto& split : nc_vector)
     {
-        auto amount = xaccSplitGetAmount (split.get ());
+        auto amount = xaccSplitGetAmount (split);
         std::vector<WorkItem> workvector = {};
 
         workvector.reserve (sack.size () + 1);
@@ -169,7 +167,7 @@ gnc_autoclear_get_splits (Account *account, gnc_numeric toclear_value,
             workvector.emplace_back (amount.num, DUP_VEC);
         else
         {
-            std::vector<std::shared_ptr<Split>> new_splits = { split };
+            std::vector<Split*> new_splits = { split };
             workvector.emplace_back (amount.num, new_splits);
         }
 
@@ -228,7 +226,7 @@ gnc_autoclear_get_splits (Account *account, gnc_numeric toclear_value,
             else
                 for (auto& s : this_splits)
                     printf (" [%5.2f]",
-                            gnc_numeric_to_double (xaccSplitGetAmount (s.get ())));
+                            gnc_numeric_to_double (xaccSplitGetAmount (s)));
             printf ("\n");
         }
 
@@ -242,7 +240,7 @@ gnc_autoclear_get_splits (Account *account, gnc_numeric toclear_value,
 
     /* copy GList because std::unordered_map value will be freed */
     for (auto& s : sack[toclear_value.num])
-        *splits = g_list_prepend (*splits, s.get ());
+        *splits = g_list_prepend (*splits, s);
 
  skip_knapsack:
 
--- a/gnucash/gnome-utils/gnc-autoclear.cpp
+++ b/gnucash/gnome-utils/gnc-autoclear.cpp
@@ -22,11 +22,12 @@
  * Boston, MA  02110-1301,  USA       gnu@gnu.org                   *
  *******************************************************************/
 
+#include <numeric>
 #include <cmath>
 #include <vector>
-#include <unordered_map>
 #include <unordered_set>
 #include <chrono>
+#include <boost/unordered_map.hpp>
 
 #include <config.h>
 #include <gtk/gtk.h>
@@ -98,23 +99,6 @@ static void looping_update_status (GtkLabel *label, guint nc_progress,
     g_free (text);
 }
 
-class Hasher
-{
-public:
-    size_t operator() (gint64 const& key) const
-    {
-        return key;
-    }
-};
-class EqualFn
-{
-public:
-    bool operator() (gint64 const& t1, gint64 const& t2) const
-    {
-        return (t1 == t2);
-    }
-};
-
 static inline gint64
 normalize_num (gnc_numeric amount, int denom)
 {
@@ -123,11 +107,12 @@ normalize_num (gnc_numeric amount, int denom)
 };
 
 static gint
-count_unique (SplitVec nc_vector, int commodity_scu)
+count_unique (SplitVec nc_vector, int scu)
 {
-    std::unordered_set<gint64, Hasher, EqualFn> buckets {};
-    for (auto& n : nc_vector)
-        buckets.insert (normalize_num (xaccSplitGetAmount (n), commodity_scu));
+    std::unordered_set<gint64> buckets;
+    std::for_each (nc_vector.begin(), nc_vector.end(),
+                   [&buckets, &scu](const auto& e)
+                   { buckets.insert (normalize_num (xaccSplitGetAmount (e), scu)); });
     return buckets.size ();
 }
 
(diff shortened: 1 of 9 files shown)


christopherlam:
Moved into the Reconcile Info dialog instead. Will run autoclear prior to opening reconcile window.
image
@jralls I'd moved algorithm into `gnome-utils/gnc-autoclear.c` and `gnome-utils/test/test-autoclear.cpp` but the latter can't seem to find config.h or glib.h
Moreover if autoclear fails, there seem to be memory corruption somewhere...

Bob-IT:
@christopherlam  Just wondering instead of the commit 'comment out call to dump_sack' you just add this and leave dump_sack as is...

`if (qof_log_check (G_LOG_DOMAIN, QOF_LOG_DEBUG))`
`   dump_sack (thisvalue, splits, data);`

christopherlam:
> @christopherlam Just wondering instead of the commit 'comment out call to dump_sack' you just add this and leave dump_sack as is...
> 
> `if (qof_log_check (G_LOG_DOMAIN, QOF_LOG_DEBUG))` ` dump_sack (thisvalue, splits, data);`

`dump_sack` call had to be removed because somehow `sack` isn't accessible anymore in `sack_free`. It's probably false economy, but the `(qof_log_check (G_LOG_DOMAIN, QOF_LOG_DEBUG))` being cached into a gboolean seems an easy win...

christopherlam:
latest commit attempts conversion to cpp. gnucash-cpp doesn't compile. any assistance?

gnucash/gnome-utils/gnc-autoclear.hpp
```diff
@@ -0,0 +1,47 @@
...
+ * Boston, MA  02111-1301,  USA       gnu@gnu.org
+ *******************************************************************/
+
+#ifndef GNC_AUTOCLEAR_HPP
+#define GNC_AUTOCLEAR_HPP
+
+#include <glib.h>
+#include <stdint.h>
+#include <gtk/gtk.h>
+#include <Account.h>
```
jralls, line 30: Add here
```
#ifdef __cplusplus
extern "C"
{
```

gnucash/gnome-utils/gnc-autoclear.hpp
```diff
@@ -0,0 +1,47 @@
...
+ */
+GList * gnc_account_get_autoclear_splits (Account *account, gnc_numeric toclear_value,
+                                          gchar **errmsg);
+
+/* same as above, but returns TRUE if successful, and FALSE if
+   unsuccessful and sets GError appropriately */
+gboolean gnc_autoclear_get_splits (Account *account, gnc_numeric toclear_value,
+                                   time64 end_date,
+                                   GList **splits, GError **error, GtkLabel *label);
+
```
jralls, line 46: and here
```
#ifdef __cplusplus
}
#endif
```

jralls:
> latest commit attempts conversion to cpp. gnucash-cpp doesn't compile. any assistance?

It compiles fine, it just doesn't link. The reason it doesn't link is because of C++ mangling, which I'll leave as an exercise for you to go learn about. The cure is to name the header gnc-autoclear.h and to add the two chunks above.  That tells the compiler that `gnc_autoclear_get_splits` should get C linkage, meaning that the name doesn't get mangled.

christopherlam:
Summary. 
From original work (knapsack is a GHashTable with key = `gnc_numeric*` and val = `GList*`) test-autoclear takes 500ms.
Using key = gnc_numeric.num reduces benchmark to 125ms.
Using GArray* instead of GList* for worklist reduces to 115ms.
Using std::vector instead of GArray* for worklist increases to 120ms.
Using std::unordered_map instead of GHashTable for knapsack increases to 210ms.

christopherlam:
> Summary. 
> * Using key = gnc_numeric.num reduces benchmark to 125ms.
> * Using GArray* instead of GList* for worklist reduces to 115ms.
> * Using std::vector instead of GArray* for worklist increases to 120ms.
> * Using std::unordered_map instead of GHashTable for knapsack increases to 210ms.

addendum:
* Using GArray* instead of GList* for workitem->splits reduces from 115ms to 81ms...

Patch to apply onto 24715ff96
```patch
modified   gnucash/gnome-utils/gnc-autoclear.c
@@ -72,21 +72,27 @@ typedef struct
 typedef struct
 {
     gint64 reachable_amount;
-    GList *list_of_splits;
+    GArray *array_splits;
 } WorkItem;
 
-static GList *DUP_LIST;
+static GArray *DUP_ARRAY;
 
 static WorkItem *
 make_workitem (GHashTable *hash, gint64 amount,
-               Split *split, GList *splits)
+               Split *split, GArray *splits)
 {
     WorkItem *item = g_new (WorkItem, 1);
     item->reachable_amount = amount;
-    if (splits == DUP_LIST || g_hash_table_lookup (hash, (gpointer)amount))
-        item->list_of_splits = DUP_LIST;
+    if (splits == DUP_ARRAY || g_hash_table_lookup (hash, (gpointer)amount))
+        item->array_splits = DUP_ARRAY;
     else
-        item->list_of_splits = g_list_prepend (g_list_copy (splits), split);
+    {
+        if (splits)
+            item->array_splits = g_array_copy (splits);
+        else
+            item->array_splits = g_array_sized_new (FALSE, FALSE, sizeof (Split*), 1);
+        g_array_append_val (item->array_splits, split);
+    }
     return item;
 }
 
@@ -118,7 +124,7 @@ static void looping_update_status (sack_data *data)
 }
 
 static void
-sack_foreach_func (gint64 thisvalue, GList *splits, sack_data *data)
+sack_foreach_func (gint64 thisvalue, GArray *splits, sack_data *data)
 {
     gnc_numeric numeric = xaccSplitGetAmount (data->split);
     gint64 itemval = numeric.num;
@@ -129,37 +135,39 @@ sack_foreach_func (gint64 thisvalue, GList *splits, sack_data *data)
     looping_update_status (data);
 }
 
+/*
 static void dump_sack (gint64 thisvalue, GList *splits, sack_data *data)
 {
     DEBUG ("%" PRId64, thisvalue);
-    if (splits == DUP_LIST)
+    if (splits == DUP_ARRAY)
         DEBUG (" DUPE");
     else
         for (GList *n = splits; n; n = n->next)
             DEBUG (" [%5.2f]", gnc_numeric_to_double (xaccSplitGetAmount (n->data)));
     DEBUG ("\n");
 }
+*/
 
 static void
-sack_free (gint64 thisvalue, GList *splits, gpointer user_data)
+sack_free (gint64 thisvalue, GArray *splits, gpointer user_data)
 {
-    if (splits != DUP_LIST)
-        g_list_free (splits);
+    if (splits != DUP_ARRAY)
+        g_array_free (splits, FALSE);
 }
 
 static void
 process_work (WorkItem *item, sack_data *data)
 {
-    GList *existing = g_hash_table_lookup (data->sack,
-                                           (gpointer)item->reachable_amount);
+    GArray *existing = g_hash_table_lookup (data->sack,
+                                            (gpointer)item->reachable_amount);
     g_hash_table_insert (data->sack, (gpointer)item->reachable_amount,
-                         item->list_of_splits);
+                         item->array_splits);
     if (existing)
     {
         if (G_UNLIKELY (data->debugging_enabled))
             DEBUG ("removing existing for %" PRId64 "\n", item->reachable_amount);
-        if (existing != DUP_LIST)
-            g_list_free (existing);
+        if (existing != DUP_ARRAY)
+            g_array_free (existing, FALSE);
     }
     looping_update_status (data);
 }
@@ -169,7 +177,8 @@ gnc_autoclear_get_splits (Account *account, gnc_numeric toclear_value,
                           time64 end_date,
                           GList **splits, GError **error, GtkLabel *label)
 {
-    GList *nc_list = NULL, *toclear_list = NULL;
+    GList *nc_list = NULL;
+    GArray *toclear_array = NULL;
     GHashTable *sack;
     guint nc_progress = 0, nc_list_length = 0;
     clock_t start_ticks;
@@ -179,7 +188,7 @@ gnc_autoclear_get_splits (Account *account, gnc_numeric toclear_value,
     g_return_val_if_fail (splits != NULL, FALSE);
 
     sack = g_hash_table_new ((GHashFunc) g_direct_hash, (GEqualFunc) g_direct_equal);
-    DUP_LIST = g_list_prepend (NULL, NULL);
+    DUP_ARRAY = g_array_new (FALSE, FALSE, 1);
 
     /* Extract which splits are not cleared and compute the amount we have to clear */
     for (GList *node = xaccAccountGetSplitList (account); node; node = node->next)
@@ -245,7 +254,7 @@ gnc_autoclear_get_splits (Account *account, gnc_numeric toclear_value,
             goto skip_knapsack;
         }
         else if (G_UNLIKELY (g_hash_table_lookup (sack, (gpointer)toclear_value.num)
-                             == DUP_LIST))
+                             == DUP_ARRAY))
         {
             g_set_error (error, autoclear_quark (), AUTOCLEAR_MULTIPLE,
                          _("Cannot uniquely clear splits. Found multiple possibilities."));
@@ -255,28 +264,34 @@ gnc_autoclear_get_splits (Account *account, gnc_numeric toclear_value,
 
     status_update (label, "Cleaning up...");
 
-    toclear_list = g_hash_table_lookup (sack, (gpointer)toclear_value.num);
+    toclear_array = g_hash_table_lookup (sack, (gpointer)toclear_value.num);
 
     /* Check solution */
-    if (!toclear_list)
+    if (!toclear_array)
     {
         g_set_error (error, autoclear_quark (), AUTOCLEAR_UNABLE,
                      _("The selected amount cannot be cleared."));
         goto skip_knapsack;
     }
     else
+    {
         /* copy GList because GHashTable value will be freed */
-        *splits = g_list_copy (toclear_list);
+        size_t size = toclear_array->len;
+        *splits = NULL;
+        for (int i = 0; i < size; i++)
+            *splits = g_list_prepend (*splits, g_array_index
+                                      (toclear_array, Split*, i));
+    }
 
  skip_knapsack:
     g_hash_table_foreach (sack, (GHFunc) sack_free, NULL);
     g_hash_table_destroy (sack);
     g_list_free (nc_list);
-    g_list_free (DUP_LIST);
+    g_array_free (DUP_ARRAY, TRUE);
 
     status_update (label, NULL);
 
-    return (toclear_list != NULL);
+    return (toclear_array != NULL);
 }
 
```

gnucash/gnome-utils/gnc-autoclear.cpp
```diff
@@ -58,22 +59,22 @@ autoclear_quark (void)
...
+struct WorkItem
 {
-    GArray *workarray;
+    gint64 reachable_amount;
+    GList *list_of_splits;
+};
+
+struct sack_data
+{
+    std::vector<WorkItem*> workvector;
```
gjanssens: Your use of std::vector is probably suboptimal because you store  pointers of `WorkItem`. That means the cpu can't benefit from cache locality. Additionally the list of splits is also a list of  pointers reducing the efficiency even more. It's actually impressive std::vector is already that efficient compared to a plain GHash in the current implementation.

Secondly I don't know if struct WorkItem is a movable type by default as no constructors, destructors and assignment operators are defined. In that case each potential move operation is performed as a copy operation instead.

All of that just to champion for std::vector rather than GArray*. As we move to C++ it's worth using the standard containers proposed and in almost all cases the std::vector is the fastest option as long as the cpu can benefit from cache locality.

christopherlam: You're looking at an earlier commit whereby GArray was replaced with std::vector<Split*> so that it compiles. Would the latest commits improve things?

gnucash/gnome-utils/gnc-autoclear.cpp
```diff
@@ -250,27 +213,40 @@ gnc_autoclear_get_splits (Account *account, gnc_numeric toclear_value,
...
     {
         g_set_error (error, autoclear_quark (), AUTOCLEAR_UNABLE,
                      _("The selected amount cannot be cleared."));
         goto skip_knapsack;
     }
     else
-        /* copy GList because GHashTable value will be freed */
-        *splits = g_list_copy (toclear_list);
+        /* copy GList because std::unordered_map value will be freed */
+        *splits = g_list_copy (sack[toclear_value.num]);
```
gjanssens: In theory if you don't need the value of the std::unordered_map further down you should be able to std::move it into the splits variable. That will avoid an expensive copy. Only theory, not tested yet.

gnucash/gnome-utils/gnc-autoclear.cpp
```diff
@@ -225,18 +224,17 @@ gnc_autoclear_get_splits (Account *account, gnc_numeric toclear_value,
     }
     else
         /* copy GList because std::unordered_map value will be freed */
-        *splits = g_list_copy (sack[toclear_value.num]);
+    {
+        *splits = nullptr;
+        for (auto& s : sack[toclear_value.num])
+            *splits = g_list_prepend (*splits, s);
+    }
```
gjanssens, line 231: I don't know which works faster. Sticking with GLists (assuming you need those to interact with the rest of gnucash) and avoid copying using std::move or working with a std::vector internally in autoclear and copying between GLists and std::vectors at the interface boundaries.
Long-term I suspect all split lists will eventually become std:vectors so at that point it will definitely make sense. Depending on your performance tests it may be too early to do so here already.

gjanssens:
Op dinsdag 2 november 2021 15:30:15 CET schreef Christopher Lam:
> @christopherlam commented on this pull request.
> 
> >  {
> 
> -    GArray *workarray;
> +    gint64 reachable_amount;
> +    GList *list_of_splits;
> +};
> +
> +struct sack_data
> +{
> +    std::vector<WorkItem*> workvector;
> 
> You're looking at an earlier commit whereby GArray was replaced with
> std::vector<Split*> so that it compiles. Would the latest commits improve
> things?

I'll admit I haven't looked into the whole of the code and how it works exactly. Mine were just 
general observations.

Another one of those: C++ containers benefit greatly from using std::algorithm functions. 
There are lots of for loops in the code. You may be able to replace some of those with more 
specific algorithms (things like find_if, partition,...). I'm still fairly novice in working with 
algorithms myself so there's not too much I can help there, but I'm told the choice of a good 
container with the right std algorithm can result in very efficient code.

christopherlam:
> I'll admit I haven't looked into the whole of the code and how it works exactly. Mine were just general observations. Another one of those: C++ containers benefit greatly from using std::algorithm functions. There are lots of for loops in the code. You may be able to replace some of those with more specific algorithms (things like find_if, partition,...). I'm still fairly novice in working with algorithms myself so there's not too much I can help there, but I'm told the choice of a good container with the right std algorithm can result in very efficient code.

I'm not sure I'd need the std::algorithm (similar to srfi-1) -- this is brute force knapsack.

- scan splitlist (all NREC) one at a time, knapsack (a hash table initially empty)
- generate a worklist from knapsack contents (initially empty) + one split
- the whole worklist (initially 1 only) populates the knapsack (which then has 1 pair)
- then restart loop with next split

thus:
- 1st iteration: one split, zero in knapsack, one in worklist ==> one in knapsack
- 2nd iteration: one split, one in knapsack, two in worklist ==> at most three in knapsack (or two if splits are identical)
- 3rd iteration: one split, three in knapsack, four in worklist ==> at most seven in knapsack (or three if splits are identical)
- 4th iteration: one split, seven in knapsack, eight in worklist ==> at most fifteen in knapsack (or four if splits are identical)

Thus for N splits, the knapsack size will be between N and 2^N-1.

gnucash/gnome-utils/gnc-autoclear.cpp
```diff
@@ -0,0 +1,280 @@
...
+        else if (gnc_numeric_zero_p (xaccSplitGetAmount (split)))
+            DEBUG ("skipping zero-amount split %p", split);
+        else if (gnc_numeric_check (xaccSplitGetAmount (split)))
+            DEBUG ("skipping invalid-amount split %p", split);
+        else if (end_date != INT64_MAX &&
+                 xaccTransGetDate (xaccSplitGetParent (split)) > end_date)
+            DEBUG ("skipping split after statement_date %p", split);
+        else
+        {
+            auto s = static_cast<std::shared_ptr<Split>>(split);
```
jralls, line 146: I'm surprised this compiles. It doesn't create a shared_ptr from split, it reinterprets the address stored in split as a shared_ptr<Split>. shared_ptrs are bigger than raw ptrs so stack mayhem is a likely result. To create a shared_ptr from a raw ptr you'd
```
    std::shared_ptr<Split> s{split};
```
Doing that transfers ownership of split to s so that when s's ref_count goes to 0 s will call operator delete on split. That will be a problem for two reasons: First, split was allocated with malloc and freeing it with operator delete won't work. Second, there's exactly one split in the system and freeing it means all of the other references to it are invalid and a use-after-free crash is inevitable.

gnucash/gnome-utils/gnc-autoclear.cpp
```diff
@@ -0,0 +1,280 @@
...
+            auto new_s = new Split;
+            memcpy (new_s, &s, sizeof(Split));
+            nc_vector.emplace_back (s);
```
jralls, line 149: [emplace_back](https://en.cppreference.com/w/cpp/container/vector/emplace_back) will copy-construct a new shared_ptr<Split> from s; Since s's [control block ptr](https://www.nextptr.com/tutorial/ta1358374985/shared_ptr-basics-and-internals-with-examples) is undefined so will the one in the vector. If you wanted and had made a valid shared_ptr you'd normally do `nc_vector.push_back(std::move(s))` to quickly transfer s into the vector without the expense of copying and deleting it with the consequent two changes to the control block.

gnucash/gnome-utils/gnc-autoclear.cpp
```diff
@@ -0,0 +1,280 @@
...
+        else if (gnc_numeric_check (xaccSplitGetAmount (split)))
+            DEBUG ("skipping invalid-amount split %p", split);
+        else if (end_date != INT64_MAX &&
+                 xaccTransGetDate (xaccSplitGetParent (split)) > end_date)
+            DEBUG ("skipping split after statement_date %p", split);
+        else
+        {
+            auto s = static_cast<std::shared_ptr<Split>>(split);
+            auto new_s = new Split;
+            memcpy (new_s, &s, sizeof(Split));
```
jralls, line 148: What's this for?

gnucash/gnome-utils/gnc-autoclear.cpp
```diff
@@ -0,0 +1,280 @@
...
+                     _("No uncleared splits found."));
+        goto skip_knapsack;
+    }
+
+    start_ticks = next_update_tick = clock ();
+
+    for (auto& split : nc_vector)
+    {
+        auto amount = xaccSplitGetAmount (split.get ());
+        std::vector<WorkItem> workvector = {};
```
jralls, line 171: Leave off the `=`.

gnucash/gnome-utils/gnc-autoclear.cpp
jralls: The good news is that you can just put the `Split*`s in your vectors, you don't need to worry about them getting freed out from under you.

gnucash/gnome-utils/gnc-autoclear.cpp
```diff
@@ -0,0 +1,280 @@
...
+        return (t1 == t2);
+    }
+};
+
+gboolean
+gnc_autoclear_get_splits (Account *account, gnc_numeric toclear_value,
+                          time64 end_date,
+                          GList **splits, GError **error, GtkLabel *label)
+{
+    std::vector<std::shared_ptr<Split>> nc_vector, DUP_VEC;
```
jralls, line 115: All caps identifiers should only be used for preprocessor macros, not variables.

gnucash/gnome-utils/gnc-autoclear.cpp
```diff
@@ -0,0 +1,280 @@
...
+
+    start_ticks = next_update_tick = clock ();
+
+    for (auto& split : nc_vector)
+    {
+        auto amount = xaccSplitGetAmount (split.get ());
+        std::vector<WorkItem> workvector = {};
+
+        workvector.reserve (sack.size () + 1);
+        if (sack.find (amount.num) != sack.end ())
```
jralls, line 174: I don't see where you're ever adding anything to `sack`.

gnucash/gnome-utils/gnc-autoclear.cpp
```diff
@@ -0,0 +1,280 @@
...
+            workvector.emplace_back (amount.num, DUP_VEC);
```
jralls, line 175: Ditto DUP_VEC.

gnucash/gnome-utils/gnc-autoclear.cpp
```diff
@@ -0,0 +1,280 @@
...
+                auto new_splits = map_splits;
+                new_splits.emplace_back (split);
+                workvector.emplace_back (new_value, new_splits);
+            }
+            looping_update_status (label, nc_progress, nc_vector.size (),
+                                   sack.size (), &next_update_tick);
+        }
+
+        for (auto& item : workvector)
+            sack[item.reachable_amount] = item.splits_vector;
```
jralls, line 206: Ah, there it is!

gnucash/gnome-utils/gnc-autoclear.cpp
```diff
@@ -0,0 +1,280 @@
...
+
+        nc_progress++;
```
jralls, line 208: `++nc_progress` so that it doesn't create an unused temporary.

gnucash/gnome-utils/gnc-autoclear.cpp
jralls: Since there's never anything in it shouldn't it be empty_vec instead of dup_vec?

christopherlam: Yes dup_vec is meant to be a unique pointer, a marker to use as sack value to indicate the sack key (an int64) would have ambiguous split list. In C it works well, in C++ it's probably not a unique pointer anymore.

gnucash/gnome-utils/gnc-autoclear.cpp
christopherlam: It was my attempt to create a shared_ptr. I'll revert the two commits.

gnucash/gnome-utils/gnc-autoclear.cpp
jralls: It's not C++'s fault that you're passing around the object instead of the pointer.

gnucash/gnome-utils/gnc-autoclear.cpp
```diff
@@ -0,0 +1,272 @@
...
+        return (t1 == t2);
+    }
+};
+
+gboolean
+gnc_autoclear_get_splits (Account *account, gnc_numeric toclear_value,
+                          time64 end_date,
+                          GList **splits, GError **error, GtkLabel *label)
+{
+    std::vector<Split*> nc_vector, empty_vec;
```
jralls, line 113: Since you use `std::vector<Split*>` a lot you might want to make an alias for it:
```
using SplitVec = std::vector<Split*>;
```
then you can say
```
SplitVec nc_vector, empty_vec{}; // Always declare and initialize together.
```

gnucash/gnome-utils/gnc-autoclear.cpp
christopherlam: ok tried `static_cast<SplitVec>(1)` which does work...

gnucash/gnome-utils/gnc-autoclear.cpp
```diff
@@ -0,0 +1,274 @@
...
+                             "Overflow error: Amount numbers are too large!");
+                goto skip_knapsack;
+            }
+            if (map_splits == static_cast<SplitVec>(1) || sack.find (new_value) != sack.end ())
+                workvector.emplace_back (new_value, static_cast<SplitVec>(1));
+            else
+            {
+                auto new_splits = map_splits;
+                new_splits.emplace_back (split);
+                workvector.emplace_back (new_value, new_splits);
```
jralls, line 193:
```
new_splits.push_back(split);
```
`emplace_back` allocates and constructs a copy of  and stores the pointer to that in the vector. You don't need or want the extra layer of indirection. `push_back` copies the pointer into the slot in the split.

jralls:
Still much worse than the 210ms you had two days ago, never mind the low-water mark of 115ms with GHashtable and GArray worklists.

You're doing a lot of copying of vectors. Were you doing that with GArrays too?

christopherlam:
> Still much worse than the 210ms you had two days ago, never mind the low-water mark of 115ms with GHashtable and GArray worklists.
> 
> You're doing a lot of copying of vectors. Were you doing that with GArrays too?

Yes, see pure C https://github.com/Gnucash/gnucash/pull/1178#issuecomment-957409244

christopherlam:
`g_array_copy` uses memcpy internally which makes it speedy.

christopherlam:
the SplitVec during knapsack is always created with known length, and would be the perfect user for the ill-fated `dynarray`... https://stackoverflow.com/a/19659449

jralls:
std::dynarray would have been std::vector without the ability to expand after the initial allocation. You already get exactly the same behavior with std::vector with [`reserve`](https://github.com/Gnucash/gnucash/pull/1178/files#diff-94019ff9e3fd6b92d2a1f217438725b76d0958f314f16767d2d52edbc0121f55R158). Contrary to the claims in that SO it was free-store allocated; there's no way to get a runtime-sized container on the stack because stack allocation requires the compiler to know how big the object is.

jralls:
@christopherlam I've done some optimization on this, it's in my [new-autoclear](https://github.com/jralls/gnucash/tree/new-autoclear) branch. Please take it for a spin.

gnucash/gnome-utils/gnc-autoclear.cpp
```diff
@@ -0,0 +1,271 @@
...
+                SplitVec new_splits{};
+                new_splits.reserve(map_splits.size() + 1);
+                new_splits = map_splits;
+                new_splits.push_back (split);
+                workvector.emplace_back (new_value, std::move(new_splits));
+            }
+        }
+
+        for (auto& item : workvector)
+            sack.insert_or_assign(item.reachable_amount, std::move(item.splits_vector));
```
christopherlam, line 196: this part *may* be `std::thread`able

gnucash/gnome-utils/gnc-autoclear.cpp
```diff
@@ -0,0 +1,271 @@
...
+    g_free (text);
+}
+
+class Hasher
+{
+public:
+    size_t operator() (gnc_numeric const& key) const
+    {
+        double d = gnc_numeric_to_double (key);
+        return g_double_hash (&d);
```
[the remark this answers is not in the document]
jralls: Be *very* careful with `gnc_numeric_to_double`: It has a lot less precision than one would expect from the 64 bits it consumes. It will likely work just fine with OECD member currencies but may not for highly inflated developing ones at one end or highly deflated ones like Bitcoin.

gnucash/gnome-utils/gnc-autoclear.cpp
```diff
@@ -0,0 +1,281 @@
...
+        return (t1 == t2);
+    }
+};
+
+static inline gint64
+normalize_num (gnc_numeric amount, int first_denom)
+{
+    return ((amount.num * first_denom) / amount.denom);
+};
+
```
christopherlam, line 107: The concern here is that we need to account for large denoms which could overflow. This function is called infrequently (nc_vector_size() + 1) therefore doesn't need to be too fast.

christopherlam:
```cpp
static inline gint64
normalize_num (gnc_numeric amount, int first_denom)
{
    auto cmp = first_denom - amount.denom;
    if (G_LIKELY (cmp == 0))
        return amount.num;
    else if (cmp > 0)       // e.g. gnc_numeric is 1230/10 and first_denom is 100
    {
        auto ratio = first_denom / amount.denom;
        return amount.num * ratio;
    }
    else                  // e.g. gnc_numeric is 123000/1000 and first_denom is 100
    {
        auto ratio = amount.denom / first_denom;
        return amount.num / ratio;
    }
};
```

jralls: This is very unsafe against overflow if the denominators are large, especially if they don't share common factors. You also risk getting wedged where a total of several splits is no longer convertible to the denominator of toclear_value

Instead use  [gnc_numeric_convert](https://github.com/Gnucash/gnucash/blob/aa450cb00eb9c10d07113fe36ec992b684f981c0/libgnucash/engine/gnc-numeric.h#L457); it should always be possible and safe to convert to the commodity's SCU, after which you can safely use the numerator for all computation.

christopherlam:
Can't think of any further speedups. I'd suggest merge the first 3 commits independently, then fixup everything from c6255a1be22243ea660d6be51984cb5438c21fa5 onwards into 1 without force-push because this PR has lots of (IMO) interesting commentary. Any further comments?

gnucash/gnome-utils/gnc-autoclear.cpp
christopherlam: fwiw this part is technically parallelisable because all workitems (except possibly one -- the new split amount) have different reachable amounts, but all parallel workers would try to modify the same unordered_map, thus defeating the multithreading.

jralls:
In that last commit message
> splits with amounts $10 $20 $30 $40 $50 will lead to a sack size of 15 instead of the theoretical 5-split limit of 31.

Didn't you want two of the amounts to be duplicated?

gnucash/gnome-utils/gnc-autoclear.cpp
```diff
@@ -103,6 +105,15 @@ normalize_num (gnc_numeric amount, int denom)
     return normalized.num;
 };
 
+static gint
+count_unique (SplitVec nc_vector, int commodity_scu)
+{
+    std::unordered_set<gint64, Hasher, EqualFn> buckets {};
```
jralls, line 111: This consumes any benefit gained from `sack.reserve` because now the unordered set is allocating chunks for insertion.  See [this SO](https://stackoverflow.com/a/28100858) for a more performant algo to count unique members.

christopherlam: Isn't that similar to https://github.com/Gnucash/gnucash/blob/22f83c8697b2945345306fb633e63d0f4c8eb9dc/bindings/guile/utilities.scm#L182 - it requires we define a split sorter which sorts splits in order of inceasing amounts. Sounds like a lot of work for sorting `nc_vector` which is expected to be no more than a couple dozen elements.

christopherlam:
> In that last commit message
> > splits with amounts $10 $20 $30 $40 $50 will lead to a sack size of 15 instead of the theoretical 5-split limit of 31.

> Didn't you want two of the amounts to be duplicated?

This was meant to illustrate the theoretical maximum sacksize of 31 would be perfectly adequate with 5 non-identical splits which lead to multiple duplicated reachable amounts.

A better example would have been $1 $2 $3 $5 $6 $12 $15 $20 $25 $30 $40 $45 $60 $100 $120 $180 i.e. multiple numbers with numerous common factors; the algorithm would potentially handle a *large* number of splits because the sack size would be much less than the theoretical maximum.

jralls:
Read it again:
"splits with amounts $10 $20 $30 $40 $50 will lead to a ***sack size of 15 instead*** of the theoretical 5-split limit of 31."
With no duplicates that's not true.

gnucash/gnome-utils/gnc-autoclear.cpp
jralls: Not a sorter, just a less-than lambda: 
```
sort(nc_vector.begin(), nc_vector.end(),
       [](Split* s1, Split *s2)->bool { return gnc_numeric_compare(s1->amount, s2->amount) < 0; });
```

But I'll concede that you're right that cost of the ~24 allocations for stuffing the set isn't comparable to the 2<sup>24</sup> insertions into `sack`.

christopherlam:
Here's the debug output with 10,20,30,40,50 split amounts.

```patch

Unstaged changes (1)
modified   gnucash/gnome-utils/gnc-autoclear.cpp
@@ -236,7 +236,7 @@ gnc_autoclear_get_splits (Account *account, gnc_numeric toclear_value,
         }
     }
 
-    if (debugging_enabled)
+    if (true)
         for (auto& [this_value, this_splits] : sack)
         {
             printf ("dump: %" PRId64 " = ", this_value);
@@ -249,6 +249,8 @@ gnc_autoclear_get_splits (Account *account, gnc_numeric toclear_value,
             printf ("\n");
         }
 
+    printf ("sack_size = %ld\n", sack.size());
+
     /* Check solution */
     if (sack.find (toclear_normalized) == sack.end())
     {
```

dump: 11000 =  DUPE
dump: 13000 =  [10.00] [30.00] [40.00] [50.00]
dump: 14000 =  [20.00] [30.00] [40.00] [50.00]
dump: 15000 =  [10.00] [20.00] [30.00] [40.00] [50.00]
dump: 12000 =  DUPE
dump: 7000 =  DUPE
dump: 10000 =  DUPE
dump: 9000 =  DUPE
dump: 8000 =  DUPE
dump: 4000 =  DUPE
dump: 5000 =  DUPE
dump: 6000 =  DUPE
dump: 3000 =  DUPE
dump: 2000 =  [20.00]
dump: 1000 =  [10.00]
sack_size = 15

jralls:
That doesn't show what you think: It demonstrates that there are only 15 possible sack values with 5 splits, so the worst-case sack size is 2<sup>n-1</sup>-1 where n is the number of  values in `nc_vector`. 

Common factors aren't at play here, nor can they be since the operation at hand is addition not multiplication. What's at play is that there are common sums: 10 + 20 = 30, which is one of the values so 3000 is a DUPE. It's still a sack slot, it's just not a useful one because there's no way to decide whether to clear the 10 and 20 splits or the 30 one.

if there's one duplicate then there are 14 reachable values. e.g {10, 20, 20, 40, 50} yields 1000, 2000, 3000, 4000, 5000, 6000, 7000, 8000, 9000, 10000, 11000, 12000, 13000, and 14000. With 2 duplicates e.g. {10, 30, 30, 30, 50}  -> 1000, 3000, 4000, 6000, 7000, 8000, 9000, 10000, 11000, 12000, 14000, and 15000) the sack size is 12. {20, 20, 20, 20, 50}->2000, 4000, 6000, 7000, 8000, 9000, 11000, and 13000, 8 values, and 5 duplicates requires only 5 sack values. Sumarizing, with the columns n, u, n-u, sack size, and formula:
n  u  n-u   size equation
5 5 0 -> 15 2<sup>n-1</sup>-1
5 4 1 -> 14  2<sup>n-1</sup>-2
5 3 2 -> 12  2<sup>n-1</sup>-4
5 2 3 ->  8   2<sup>n-1</sup>-8
5 1 4 ->  5   2<sup>n-1</sup>-11

The first four can be found with 2<sup>n-1</sup>-2<sup>n-u</sup>; for the last one that yields 0 so for that special case the sack size is n.

The next interesting observation is that in each case there are only u usable values, the rest are dupes. If the user has 29 splits with 5 duplicates the odds of being able to auto-balance is 24/(2<sup>28</sup>-2<sup>5</sup> or 3 in 33,554,428 for which we will have allocated a  sack using over 4GB and performed 268,435,456 additions.

How exactly does this make sense?

christopherlam:
The use case isn't to try guess the autoclear balance; it would be to reduce tedium during reconciliation- if the bank offers a monthly bank statement with (e.g. 100) transactions, and has running balance alongside, we now the autoclear would take forever. So, I could pick a date e.g. (10th of the month) with its running balance, autoclear, visually check that the cleared splits seem legitimate, and move on to 20 then 30th of the month. If autoclear was _impossible_ then it's obvious there is a data entry error somewhere within the date interval of interest. If autoclear fails due to duplicate paths, then we need to revert to manual reconcile.

jralls:
In that scenario there's exactly one useful solution and there's no need to go through all of this rigamarole: Either all of the splits plus the beginning balance add up to the ending balance or they don't. That's one simple addition, no need for ginormous hash tables.

jralls:
If you're going to argue that some of the splits might not have cleared so you have to do this, then you have to accept that the odds that you'll be able to find a unique solution are on the order of 1 in 10 million. No user gets enough bank statements in a lifetime to have a reasonable chance of having a single autoclear success ever.

luzpaz:
Is there an ETA on this project ?

christopherlam:
> Is there an ETA on this project ?

It's not really that popular. What's your use case?

luzpaz:
Just looks like a useful feature to incorporate in to a workflow

christopherlam:
Merged in #1769 without cpp conversion... glib is much more lightweight than stl.

PR #1181, opened 2021-10-28
Separate the warnings for "cut split/transaction" and "delete split/transaction"

nomis:


fellen:
Question from the translator view: Cut like Copy&Paste?

Perhaps we should explain it somewhere?

jralls:
Yes, cut as in ctrl-x. What other meaning of cut do we use in GnuCash?

jralls:
Thanks!

christopherlam:
This code has invalidated the existing gsettings `reg-split-cut` and `reg-split-cut-recd` which means gnc-gsettings.cpp#L367 will log error e.g. when performing Check & Repair:

```cpp

gint
gnc_gsettings_get_int (const gchar *schema,
                       const gchar *key)
{
    GSettings *settings_ptr = gnc_gsettings_get_settings_ptr (schema);
    g_return_val_if_fail (G_IS_SETTINGS (settings_ptr), 0);

    if (gnc_gsettings_is_valid_key (settings_ptr, key))
        return g_settings_get_int (settings_ptr, key);
    else
    {
        PERR ("Invalid key %s for schema %s", key, schema);
        return 0;
    }
}
```

any clue how best to silence this PERR?

jralls:
No, I think your explanation does make sense. Somewhere some code is asking for the values of those two keys and since they've been removed from the schema it's throwing that error. The quickest way to track it down is to set a breakpoint on that line in the debugger and look at the backtrace to see where the request is coming from.

christopherlam:
Backtrace:

```
#0  0x00007ffff7b61d97 in g_logv () at /lib/x86_64-linux-gnu/libglib-2.0.so.0
#1  0x00007ffff7b62053 in g_log () at /lib/x86_64-linux-gnu/libglib-2.0.so.0
#2  0x00007ffff6cf9f63 in gnc_gsettings_get_int(gchar const*, gchar const*) (schema=0x7ffff78c01f4 "warnings.temporary", key=0x7ffff78de357 "reg-split-cut")
    at /home/chris/sources/gnucash/maint/libgnucash/app-utils/gnc-gsettings.cpp:377
#3  0x00007ffff6843487 in gnc_prefs_get_int (group=0x7ffff78c01f4 "warnings.temporary", pref_name=0x7ffff78de357 "reg-split-cut")
    at /home/chris/sources/gnucash/maint/libgnucash/core-utils/gnc-prefs.c:205
#4  0x00007ffff780a09e in gnc_reset_warnings_add_section (rw_dialog=0x5555561df030, prefs_group=0x7ffff78c01f4 "warnings.temporary", box=0x5555560c26f0)
    at /home/chris/sources/gnucash/maint/gnucash/gnome-utils/dialog-reset-warnings.c:300
#5  0x00007ffff780a80f in gnc_reset_warnings_dialog (parent=0x555555a32340) at /home/chris/sources/gnucash/maint/gnucash/gnome-utils/dialog-reset-warnings.c:402
#6  0x00007ffff784c3b3 in gnc_main_window_cmd_actions_reset_warnings (action=0x555555751680, window=0x555555a32340)
    at /home/chris/sources/gnucash/maint/gnucash/gnome-utils/gnc-main-window.c:4444
```

nomis:
Do you have outdated schema files installed?

jralls:
> Nothing has been removed from the schema by this change.

Quite right, those two are added with this change.

gjanssens:
I think the error happens because the keys are only added in the permanent warnings section and not in the temporary one as well. Copy the same keys in the schema `org.gnucash.GnuCash.warnings.temporary` and I suspect the errors will go away.

nomis:
I've created #1293 to fix this.

PR #1183, opened 2021-11-01
gnc_list_length comparator functions

christopherlam:
for review candidate g_list_length comparator function names... `gnc_list_length_gt` `gnc_list_length_lt` `gnc_list_length_eq`?

### DIFF
--- a/libgnucash/core-utils/gnc-glib-utils.c
+++ b/libgnucash/core-utils/gnc-glib-utils.c
@@ -351,3 +351,13 @@ gnc_g_list_stringjoin (GList *list_of_strings, const gchar *sep)
 
     return retval;
 }
+
+gint
+gnc_list_length_cmp (const GList *list, size_t len)
+{
+    for (GList *lst = (GList*) list;; lst = g_list_next (lst), len--)
+    {
+        if (!lst) return (len ? -1 : 0);
+        if (!len) return 1;
+    }
+}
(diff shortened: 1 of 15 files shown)


libgnucash/core-utils/gnc-glib-utils.c
```diff
@@ -351,3 +351,33 @@ gnc_g_list_stringjoin (GList *list_of_strings, const gchar *sep)
 
     return retval;
 }
+
+gboolean
+gnc_list_length_cmp (GList *lst, size_t len,
+                     gboolean less, gboolean same, gboolean more)
```
jralls, line 357: That's ugly API that shouldn't be public, so `static inline boolean list_length_cmp_internal`.

jralls:
That will work, but what's the use case?

christopherlam:
These would be used when undoing b5f5129f7.

jralls:
You only need `gnc_list_length_gt` for that, and the intent is less clear than `gnc_list_length_at_least` (which could be renamed `gnc_list_length_ge`. The indirection of the (very ugly) `list_length_cmp_internal` gains nothing: There's already `g_list_length` and `!gnc_list_length_at_least` is equivalent to `gnc_list_length_lt`.

jralls:
It occurs to me that you could change the cmp function to be like traditional c-library compares such as strcmp and return -1 for less, 0 for equal, and 1 for greater. That would get rid of the ugly boolean inputs. OTOH the only benefit to testing equality is if length is much less than g_list_length as in that case it would cut short the list traversal; the same is also true of less-than. In both cases one would get the benefit only if one expected the answer to be false in the majority of instances.

christopherlam:
In this case `cmp` will be the only function required. Callers can compare the `gint` result as desired, from the whole <= < == > => range.

jralls:
Good.

PR #1184, opened 2021-11-03
A couple of Right to Left Bugs.

Bob-IT:
The first bug commit, Bug798341 changes the alignment of monetary columns in tree views built with gnc_tree_view_add_numeric_column. By default they are right aligned but when an RTL language is used, GTK transposes the alignment so they become left aligned. This commit changes them to be right aligned for all languages. I do not know if there is a reason for them not to be but another option is to have a preference under the accounts tab like 'Swap monetary column alignment'.

The second commit, Bug797501 is illustrated in this image of part of the CoA on top of a terminal window...
rtl1
The terminal window shows the tree model string generated from xaccPrintAmount which is being changed when displayed in the tree view.

I think this is to do with GTK's implementation of 'Unicode Bidirectional Text Algorithm' which I have fixed by inserting a BiDi override character at the start of the string returned by xaccPrintAmount and a BiDi closing character at the end so you end up with this...
rtl5

The way I have done it affects all number strings created by xaccSPrintAmount which display a symbol as well as the number so corrects the reports also.

I think the test failures are probably down to comparing of strings which I have not looked at due to uncertainty of whether this is the correct fix. If it is I will look at and fix.


jralls:
Adding the BiDi overrides blows up every unit test that uses the output of `xaccPrintAmount`.

Bob-IT:
> Adding the BiDi overrides blows up every unit test that uses the output of `xaccPrintAmount`.

Yep, that is what I thought and why I said I was uncertain of whether it was the right fix. I changed the fix to be more local by adding a function to wrap some text with the BiDi override and then changing three functions only used for the account model.

jralls:
Looks good.

fellen:
@avma Is this the expected behaviour?

avma:
I couldn't tell, I couldn't bring it to work yet and test it...

fellen:
@avma check out this PR as separate branch, build and test it. Perhaps you can describe your problem?

jralls:
@Bob-IT Looks good, merge when you're ready.

Bob-IT:
I am considering a different approach based on my first change so have extracted other commits and pushed them separately.
Will update with new changes later that would hopefully deal with reports also.

jralls:
@Bob-IT ping...

Bob-IT:
@jralls Thanks for the ping, the following patch against maint was what I was thinking about proposing. It has the advantage of fixing **ALL** RTL amounts with symbols and I thought it would work for reports also. On testing it did not affect the webkit presented reports but opening an exported report with Firefox did work.

[0001-my-rtl-test.txt](https://github.com/Gnucash/gnucash/files/8190653/0001-my-rtl-test.txt)
 
As a result, I think it maybe better as a first step to use the wrapping function where required for the Gui, I have noticed the register status bar needs this and treat reports separately so will progress in that direction.

jralls:
@Bob-IT Have you tested the reports on Linux? The WebKit instance on Windows is so old that it's quite possible that it can't handle RTL correctly.

jralls:
As for the patch, please don't attach patches like that. Click the + at the top-right corner of the page and pick "New gist". That will open a page where you can paste in your patch. You can then link the page back here.

gnucash/gnome-utils/gnc-tree-model-account.c
```diff
@@ -583,7 +584,12 @@ gnc_tree_model_account_compute_period_balance (GncTreeModelAccount *model,
     if (negative)
         *negative = gnc_numeric_negative_p (b3);
 
-    return g_strdup(xaccPrintAmount (b3, gnc_account_print_info (acct, TRUE)));
+    print_info = gnc_account_print_info (acct, TRUE);
+
+    if (print_info.use_symbol)
+        return gnc_wrap_text_with_bidi_ltr_isolate (xaccPrintAmount (b3, print_info));
+    else
+        return g_strdup (xaccPrintAmount (b3, print_info));
```
jralls, line 592: Rather than repeating this test a gazillion times, move it into `XaccPrintAmount`.

Bob-IT:
> Have you tested the reports on Linux? The WebKit instance on Windows is so old that it's quite possible that it can't handle RTL correctly.

This was on a Linux Mint 20.2 VM that I use to switch between an English user and Hebrew user.

Bob-IT:
> "New gist"

Did not know about that option, always learning...

gnucash/gnome-utils/gnc-tree-model-account.c
Bob-IT: Can not do that, I did something similar for my first commit but that would affect all the report tests. What I have done is create another function similar to xaccAccountPrint that does do that and used that function where needed. Thinking about it I probably do not need to test for 'use_symbol' as all the places used do display the symbol but no harm in leaving it.

jralls: Much better, thanks.

libgnucash/app-utils/gnc-ui-util.c
```diff
@@ -1863,6 +1863,39 @@ xaccPrintAmount (gnc_numeric val, GNCPrintAmountInfo info)
...
+{
+    /* hack alert -- this is not thread safe ... */
+    static char buf[1024];
+
+    if (!xaccSPrintAmount (buf, val, info))
+        buf[0] = '\0';
+
+    if (buf[0] && info.use_symbol)
+    {
+        char temp_buf[1024];
```
jralls, line 1877: This is a bit of a problem because it needs 6 bytes more than `buf` to accommodate the two bidi code points. Maybe a `strlen` check to make sure that there's enough left over or use `strncat (temp_buf, buf, buflen - 7)` to ensure that `temp_buf` doesn't get overrun?

Since this gets called a lot maybe a bit of hackery would be worthwhile:
```
#define BUFLEN 1024
static char buf[BUFLEN];
static const char[] ltr_isolate = { 0xe2, 0x81, 0xa6 };
static const char[] ltr_pop_isolate = { 0xe2, 0x81, 0xa9 };
size_t offset = info.use_symbol ? 3 : 0;

memset (buf, 0, BUFLEN);
if (!xaccSPrintAmount (buf + offset, val, info))
{
    buf[0] = '\0';
    return buf;
};

if (!info.use_symbol)
    return buf;

memcpy (buf, ltr_isolate, 3);
if (buf[BUFLEN] != '\0')
{
    buf[BUFLEN] = '\0';
    memcpy (buflen + BUFLEN - 4, ltr_pop_isolate, 3);
}
else
{
    size_t length = strlen (buf);
    memcpy (buflen + length, ltr_pop_isolate, 3);
}

return buf;
```

libgnucash/app-utils/gnc-ui-util.c
```diff
@@ -1863,6 +1863,39 @@ xaccPrintAmount (gnc_numeric val, GNCPrintAmountInfo info)
...
+
+        strcpy (temp_buf, "\u2066"); // ltr isolate
+        strcat (temp_buf, buf);
+        strcat (temp_buf, "\u2069"); // pop directional formatting
```
jralls, line 1881: If you decide I've gone a bit over with a single buffer and for future reference it's faster to use `stpcpy`:
```
    char *pos = stpcpy (temp_buf, "\u2066");
    pos = stpncpy (pos, buf, buflen - 7);
    pos = stpcpy (pos, "\u2069");
```
Because it saves finding the end of the string every time.

jralls (changes requested):

libgnucash/app-utils/gnc-ui-util.c
```diff
@@ -1863,6 +1863,39 @@ xaccPrintAmount (gnc_numeric val, GNCPrintAmountInfo info)
     return buf;
 }
 
+const char *
+gnc_print_amount_with_bidi_ltr_isolate (gnc_numeric val, GNCPrintAmountInfo info)
+{
+    /* hack alert -- this is not thread safe ... */
+    static char buf[1024];
```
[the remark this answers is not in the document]
Bob-IT: Copy and paste job from xaccPrintAmount, have changed that to use the define BUFLEN

libgnucash/app-utils/gnc-ui-util.c
Bob-IT: Don't think I have used those mem functions, needed to look them up and seems a better approach, needed to tweak it slightly to get to work.

libgnucash/app-utils/gnc-ui-util.c
Bob-IT: Used your suggestion.

libgnucash/app-utils/gnc-ui-util.c
```diff
@@ -1863,6 +1865,53 @@ xaccPrintAmount (gnc_numeric val, GNCPrintAmountInfo info)
...
+    /* hack alert -- this is not thread safe ... */
+    static char buf[BUFLEN];
+    static const char ltr_isolate[] = { 0xe2, 0x81, 0xa6 };
+    static const char ltr_pop_isolate[] = { 0xe2, 0x81, 0xa9 };
+    size_t offset = info.use_symbol ? 3 : 0;
+
+    memset (buf, 0, BUFLEN);
+    if (!xaccSPrintAmount (buf + offset, val, info))
+    {
+        buf[0] = '\0';
```
jralls, line 1880: Don't need this anymore, the `memset` did it.

libgnucash/app-utils/gnc-ui-util.c
```diff
@@ -1863,6 +1865,53 @@ xaccPrintAmount (gnc_numeric val, GNCPrintAmountInfo info)
...
+
+    if (buf[BUFLEN - 1] != '\0')
+    {
+        buf[BUFLEN - 1] = '\0';
+        memcpy (buf + BUFLEN - 4, ltr_pop_isolate, 3);
+    }
+    else
+    {
+        size_t length = strlen (buf);
+        memcpy (buf + length, ltr_pop_isolate, 3);
```
jralls, line 1897: Oops, this could overrun the buffer and fail to be null-terminated if `strlen(buf) > BUFLEN - 4`.

To fix the first test could be `if ((uint32_t)buf[BUFLEN - 4])`. You'll need to add `#include <stdint.h>` up top. The reason for `uint32_t` is that Debian builds for a bunch of weird architectures like IBM360 and I'm not sure that `int` is 32 bits on all of them; and even if it is this year that can change as `sizeof(int)` is implementation-defined.

Bob-IT: I have changed it differently, so the if test is `if (buf[BUFLEN - 4] == '\0')` and now there should be room for the pop_isolate,
If the test fails, I set the last element of the buffer to `\0` and then `memcpy (buf + BUFLEN - 4, ltr_pop_isolate, 3)`. This obviously truncates the output but should never happen with the buffer size of 1024!
I did wonder about adding a PWARN for this but not sure it is worth it.

jralls: I think skip the test and always assign. 

Considering how unlikely it is that we'd see a 1017-byte representation of a number, hitting this branch means that we've somehow gotten a non-numeric string out of xaccSPrintAmount. That probably does justify a PWARN.

Bob-IT: Suggested changes made.

jralls:
Thanks!

PR #1186, opened 2021-11-06
Transaction Report and Reconciled Report, and filtering by Posted/Reconciled/Entered/other dates.

christopherlam:
* Transaction Report upgrade - can select the date to be used for date filter. Defaults to posted date.
* Transaction Report can still filter with a different date e.g. payment date, by a derived report via the `#:split->date` argument.
* Reconcile Report gains ability to filter by Posted Date, which IMHO shouldn't have been changed in the first place.


christopherlam:
Works well, and tests already exist in test-transaction.scm. The only (small) issue is that filtering by Data Entry date is potentially useful (eg what did I input last week) but there's no UI or report to show it.

PR #1201, opened 2021-11-25
xaccTransGetTxnType generates the type (and caches it) dynamically

christopherlam:
Generates TXN_TYPE dynamically (and caches it) by analysing transaction splits, instead of relying on transaction property. ~Also eliminates TXN_TYPE_LINK merging it as TXN_TYPE_PAYMENT.~ Needs much scrutiny.

### DIFF
--- a/libgnucash/engine/Transaction.c
+++ b/libgnucash/engine/Transaction.c
@@ -279,6 +279,7 @@ gnc_transaction_init(Transaction* trans)
     trans->notes = (char*) is_unset;
     trans->doclink = (char*) is_unset;
     trans->void_reason = (char*) is_unset;
+    trans->txn_type = TXN_TYPE_UNCACHED;
     LEAVE (" ");
 }
 
@@ -1705,6 +1706,7 @@ xaccTransCommitEdit (Transaction *trans)
         qof_instance_set_dirty(QOF_INSTANCE(trans));
     }
 
+    trans->txn_type = TXN_TYPE_UNCACHED;
     qof_commit_edit_part2(QOF_INSTANCE(trans),
                           (void (*) (QofInstance *, QofBackendError))
                           trans_on_error,
@@ -2551,22 +2553,43 @@ xaccTransRetDateDue(const Transaction *trans)
 }
 
 char
-xaccTransGetTxnType (const Transaction *trans)
+xaccTransGetTxnType (Transaction *trans)
 {
-    const char *s = NULL;
-    GValue v = G_VALUE_INIT;
-    char ret = TXN_TYPE_NONE;
+    gboolean has_nonAPAR_amount = FALSE;
 
     if (!trans) return TXN_TYPE_NONE;
-    qof_instance_get_kvp (QOF_INSTANCE (trans), &v, 1, TRANS_TXN_TYPE_KVP);
-    if (G_VALUE_HOLDS_STRING (&v))
+
+    if (trans->txn_type != TXN_TYPE_UNCACHED)
+        return trans->txn_type;
+
+    trans->txn_type = TXN_TYPE_NONE;
+    for (GList *n = xaccTransGetSplitList (trans); n; n = g_list_next (n))
     {
-         s = g_value_get_string (&v);
-         if (s && strlen (s) == 1)
-             ret = s[0];
+        Account *acc = xaccSplitGetAccount (n->data);
+
+        if (!acc)
+            continue;
+
+        if (!xaccAccountIsAPARType (xaccAccountGetType (acc)) &&
+            !gnc_numeric_zero_p (xaccSplitGetValue (n->data)))
+            has_nonAPAR_amount = TRUE;
+        else if (trans->txn_type == TXN_TYPE_NONE)
+        {
+            GNCLot *lot = xaccSplitGetLot (n->data);
+            GncInvoice *invoice = gncInvoiceGetInvoiceFromLot (lot);
+            GncOwner owner;
+
+            if (invoice && trans == gncInvoiceGetPostedTxn (invoice))
+                trans->txn_type = TXN_TYPE_INVOICE;
+            else if (invoice || gncOwnerGetOwnerFromLot (lot, &owner))
+                trans->txn_type = TXN_TYPE_PAYMENT;
+        }
     }
-    g_value_unset (&v);
-    return ret;
+
+    if (!has_nonAPAR_amount && (trans->txn_type == TXN_TYPE_PAYMENT))
+        trans->txn_type = TXN_TYPE_LINK;
+
+    return trans->txn_type;
 }
 
 const char *
--- a/bindings/python/tests/test_transaction.py
+++ b/bindings/python/tests/test_transaction.py
@@ -112,15 +112,6 @@ class TestTransaction(TransactionSession):
         self.trans.ClearReadOnly()
         self.assertEqual( None, self.trans.GetReadOnly() )
 
-    def test_txntype(self):
-        self.assertEqual( '\x00', self.trans.GetTxnType() )
-        TYPE = 'I'
-        self.trans.SetTxnType(TYPE)
-        self.assertEqual( TYPE, self.trans.GetTxnType() )
-        TYPE = 'P'
-        self.trans.SetTxnType(TYPE)
-        self.assertEqual( TYPE, self.trans.GetTxnType() )
-
     def test_num(self):
         NUM = '5'
         self.assertEqual( '', self.trans.GetNum() )
--- a/libgnucash/engine/Transaction.h
+++ b/libgnucash/engine/Transaction.h
@@ -121,6 +121,7 @@ GType gnc_transaction_get_type(void);
 /** @name Transaction Type field values
 @{
 */
+#define TXN_TYPE_UNCACHED '?' /** Transaction type not yet cached */
 #define TXN_TYPE_NONE	 '\0' /**< No transaction type       */
 #define TXN_TYPE_INVOICE 'I'  /**< Transaction is an invoice */
 #define TXN_TYPE_PAYMENT 'P'  /**< Transaction is a payment  */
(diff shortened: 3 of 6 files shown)


jralls:
@christopherlam I know this is motivated by a gnucash-user discussion, but I've lost which one. Can you please link it?

christopherlam:
> @christopherlam I know this is motivated by a gnucash-user discussion, but I've lost which one. Can you please link it?

https://lists.gnucash.org/pipermail/gnucash-user/2021-November/098516.html

Will take time to figure out why test-new-owner-report fails, and if this is a worth while change...

jralls:
Failures:
```
ERROR:/Users/john/Development/Gnucash-Build/Gnucash-master-git-10.13-x86_64/src/gnucash-git/libgnucash/engine/test/utest-Transaction.cpp:1845:void test_xaccTransGetTxnType(Fixture *, gconstpointer): assertion failed (i == xaccTransGetTxnType(txn)): (73 == 0)
```
and
```
Directory: /Users/john/Development/Gnucash-Build/Gnucash-master-git-10.13-x86_64
/build/gnucash-git/gnucash/report/reports/standard/test
"test-new-owner-report" start time: Feb 28 13:19 PST
Output:
new-owner-report tests:
[pass] line:230, test: line 1
[pass] line:234, test: line 2
[pass] line:237, test: line 3
[pass] line:240, test: line 4
[pass] line:244, test: line 5
[pass] line:248, test: line 6
[pass] line:252, test: line 7
[pass] line:256, test: line 8
[pass] line:260, test: line 9
[pass] line:263, test: line 10
[fail] line:267, test: line 11
test-new-owner-report/new-customer-report
 -> expected: ("1980-06-26" "Payment" "$11.75" "1980-06-24" "Invoice" "$1.00" "$
1.00" "$28.00")
 -> obtained: ("1980-06-26" "Refund" "$12.75" "1980-06-24" "Invoice" "$0.00" "$2
7.00" "$28.00")
[fail] line:273, test: line 12 refund $120 to partially repay
test-new-owner-report/new-customer-report
 -> expected: ("1980-06-28" "Payment" "-$148.25" "1980-06-30" "Refund" "$160.00"
 "-$40.00" "$50.00")
 -> obtained: ("1980-06-25" "Credit Note" "-$27.00" "-$27.00")
[fail] line:277, test: line 13 refund $120 to partially repay
test-new-owner-report/new-customer-report
 -> expected: ("1980-06-29" "Refund" "-$120.00" "$120.00")
 -> obtained: ("1980-06-26" "Payment" "$11.75" "1980-06-24" "Invoice" "$1.00" "$
1.00" "$28.00")
[fail] line:280, test: line 14 refund $120 to partially repay
test-new-owner-report/new-customer-report
 -> expected: ("1980-06-29" "Refund" "-$28.25" "1980-06-28" "Payment" "$120.00"
"$120.00" "-$120.00")
 -> obtained: ("1980-06-28" "Payment" "-$148.25" "1980-06-30" "Refund" "$160.00"
 "-$40.00" "$50.00")
[fail] line:284, test: line 15 refund $120 to partially repay
test-new-owner-report/new-customer-report
 -> expected: ("1980-06-30" "Refund" "$21.75" "1980-06-28" "Payment" "* " "$50.0
0" "$40.00" "-$40.00")
 -> obtained: ("1980-06-29" "Refund" "-$120.00" "$120.00")
[fail] line:288, test: line 16 refund $120 to partially repay
test-new-owner-report/new-customer-report
 -> expected: ("Pre-Payment" "$10.00")
 -> obtained: ("1980-06-29" "Refund" "-$28.25" "1980-06-28" "Payment" "$120.00"
"$120.00" "-$120.00")
[fail] line:291, test: line 17 refund $120 to partially repay
test-new-owner-report/new-customer-report
 -> expected: ("Period Totals" "$216.25" "$194.50" "$21.75" "* Amounts denoted t
hus are derived from, and do not match the transaction.")
 -> obtained: ("1980-06-30" "Refund" "$21.75" "1980-06-28" "Payment" "* " "$50.0
0" "$40.00" "-$40.00")
Source:test-new-owner-report.scm
pass = 10, fail = 7
<end of output>
Test time =   0.58 sec
```

Dunno about the report test, but the first one fails because the test just calls `xaccTransSetTxnType` and then tries to read it back. You removed the read it back part from `xaccTransGetTxnType` so it fails.

You may as well remove `xaccTransSetTxnType` and add a scrub to remove the KVP. You'll need to modify the test to create an invoice attached to the transaction and set its value to P or I and test `xaccTransGetTxnType` returns the correct type.

christopherlam:
The report failure happens because test-new-owner-report constructs the invoice/creditnote/payment test incorrectly. I'll see about fixing this test first and try again.

christopherlam:
> The report failure happens because test-new-owner-report constructs the invoice/creditnote/payment test incorrectly. I'll see about fixing this test first and try again.

No, the test-new-owner-report was constructing Invoice-post/CN-post/Overpayment and link Inv to CN correctly. However the old LINK txns would show as a zero-amount line. Small (hopefully harmless) modification to new-owner-report will omit a zero-amount line.

christopherlam:
Note if TxnType kvp is removed then the datafile shouldnt be read by previous builds. So, either we keep writing the kvp, or add a feature.

christopherlam:
Small change - can now return TXN_TYPE_LINK, thereby eliminating need to amend new-owner-report. IIUC xaccTransGetTxnType now returns None/Invoice/Payment/Link correctly, even if the datafile was damaged by clicking P on type column in APAR register. The old (possibly incorrect) kvp data is not modified. TXN_TYPE_LINK is concluded if all splits have lot, and lot has invoice, the transaction is not the invoice posting transaction.

It means after this merge
* Old txn_type kvp will be ignored, and heuristics used to return txn_type.
* New txn_type kvp will be written correctly.
* Thus there should be no further warnings in new-owner-report about invalid invoice txns.
* And Invoice/creditnotes/payments generated by future builds will be readable by previous gnucash correctly.

jralls:
> use new behaviour only if --extra enabled

I don't think changing invisible internal behavior based on a command-line option used for other purposes is a good idea.

christopherlam:
> > use new behaviour only if --extra enabled 
> I don't think changing invisible internal behavior based on a command-line option used for other purposes is a good idea.

Ok then I think this is "ready"

jralls:
> Ok then I think this is "ready"

It has commented-out code so it isn't ready.

jralls:
This strikes me as a good opportunity to change to a less generic name than "transaction type", perhaps "transaction business type", with `TXN_BIZ_TYPE_NONE` being commented as "This transaction is not associated with an invoice or bill."

libgnucash/engine/Transaction.c
```diff
@@ -2547,20 +2556,39 @@ xaccTransRetDateDue(const Transaction *trans)
...
+                    DEBUG ("%p txn type is %c", trans, txn->txn_type);
+                }
+                else if (gncOwnerGetOwnerFromLot (lot, &owner))
+                {
+                    txn->txn_type = TXN_TYPE_PAYMENT;
+                    DEBUG ("%p txn type is unattached payment", trans);
+                }
+            }
+        }
+        if (all_splits_have_lot && (txn->txn_type == TXN_TYPE_PAYMENT))
```
jralls, line 2588: This will break if there are any extra splits in the transaction. For example some users have mentioned that they insert random zero-value splits for tagging or for adding comments.  To handle this you could do
```
unsigned int count_lots = 0;
/* ... */
 for (GList *n = xaccTransGetSplitList (txn); n; n = g_list_next (n))
  {

      if (!lot)
         continue;
      if (txn->txn_biz_type !=  TXN_BIZ_TYPE_NONE)
      {
           ++count_lots;
           continue;
      }
     /* Do the type heurisitic */
}
if (count_lots > 1 && (txn->txn_type == TXN_TYPE_PAYMENT))
   txn->txn_type = TXN_BIZ_TYPE_LINK
```

But the comment for `TXN_TYPE_LINK` says "Transaction is a link between (invoice and payment) lots". This test  assigns it if the first split with a lot resolves to a payment and ignores the other splits with lots. What if the first split encountered is an invoice split? What if there are two payment splits, or more than two splits with lots?

christopherlam: IIUC `TXN_TYPE_LINK` and `TXN_TYPE_INVOICE` have always been strictly Read-only. @gjanssens could confirm. Wouldn't your code fail when a payment is applied to clear 2 invoices?

jralls: Wouldn't your code fail when a payment is applied to clear 2 invoices?

Yes, and so will yours.

> IIUC TXN_TYPE_LINK and TXN_TYPE_INVOICE have always been strictly Read-only.

Do you mean that splits with those flags set are read only? How is that significant here, where you're assigning the flags?

christopherlam: > > IIUC TXN_TYPE_LINK and TXN_TYPE_INVOICE have always been strictly Read-only.
> 
> Do you mean that splits with those flags set are read only? How is that significant here, where you're assigning the flags?

It would mean Invoice-posting transactions, and Lot-link Payments should never be modifiable by the user to add zero-amount splits to an arbitrary account. However @gjanssens had fixed some lot-link proliferation bugs so I'm not sure my assumption stands.

gjanssens: TXN_TYPE_LINK transactions are writable. Only TXN_TYPE_INVOICE transactions are read-only.

christopherlam: Ok last commit(s) amended: if we conclude TXN_TYPE_PAYMENT, and there are no non-APAR zero-amount splits, then it must be a TXN_TYPE_LINK.

libgnucash/engine/Transaction.c
```diff
@@ -2547,20 +2556,40 @@ xaccTransRetDateDue(const Transaction *trans)
...
     if (!trans) return TXN_TYPE_NONE;
-    qof_instance_get_kvp (QOF_INSTANCE (trans), &v, 1, TRANS_TXN_TYPE_KVP);
-    if (G_VALUE_HOLDS_STRING (&v))
+
+    if (trans->txn_type == TXN_TYPE_UNCACHED)
     {
-         s = g_value_get_string (&v);
-         if (s && strlen (s) == 1)
-             ret = s[0];
+        Transaction *txn = (Transaction*) trans;
```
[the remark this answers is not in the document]
christopherlam: `trans` is a const so its `trans->txn_type` won't be modifiable.

christopherlam:
Not sure why mac fails.

jralls:
The crash is because of a use-after-free; macOS is built so that bugs like that crash immediately. It's not completely your fault, the order of destruction was wrong in `teardown()`, you just happened to trip it by setting a lot on your splits.When the first of those splits were destroyed as part of `xaccAccountDestroy` it sent a QofEvent to the customer, but the customer had already been destroyed so when `cust_handle_qof_events` tried to free `cust->balance` at gncCustomer.c line 1872 cust was already freed and (here's the macOS part) its memory written over with `0xaa`.  Trying to `g_free(0xaaaaaaaaaaaaaaaa)` crashed.

The fix is simple enough, don't destroy the customer or vendor until after the accounts are destroyed.

jralls:
Sorry about dragging in the extra commit from maint. You'll have to rebase it away before merging.

christopherlam:
thanks to the fix, I've been able to move the payment and cn+link transactions into the appropriate `setup` functions.

christopherlam:
The last consideration is anticipating errors in heuristics -- I _think_ the invoice posting transactions will be protected via Readonly irrespective of txn_type.

christopherlam:
Ok final thoughts on this line of work; the positives of this branch https://github.com/Gnucash/gnucash/pull/1201#issuecomment-1126852946 remain. I will not particularly be inclined to write an internal messaging subsystem due to lack of time. Any last concerns?

libgnucash/engine/Transaction.c
```diff
@@ -2551,22 +2553,45 @@ xaccTransRetDateDue(const Transaction *trans)
...
-             ret = s[0];
+        Account *acc = xaccSplitGetAccount (n->data);
+
+        if (!acc)
+            continue;
+
+        if (!xaccAccountIsAPARType (xaccAccountGetType (acc)) &&
+            !gnc_numeric_zero_p (xaccSplitGetValue (n->data)))
+            has_nonAPAR_amount = TRUE;
+
```
jralls, line 2576: No blank lines between `if `clauses and their `else`s.

jralls (approved):

jralls:
LGTM other than the one whitespace niggle. Merge when ready.

PR #1191, opened 2021-11-12
C++options

jralls:
Reimplement the options system used for File Properties, Stylesheets properties, and individual report options in C++, replacing the existing mixed scheme and C mess.

This PR is too big and complicated for me to expect any kind of a thorough review. Unfortunately I couldn't see a good way to make the changes more incrementally without breaking along the way.

There are a few tasks to complete after the initial merge. I've left them off for now to make this PR a little more cohesive, but after it's merged I plan to:
* Move the C++ files to engine
* Move gnc-optiondb.i, options.scm, and the scheme tests to bindings and integrate gnc-optiondb.i into swig-engine.
This because the book options (file properties) are written to the book so it should be formally part of the gnucash library.
With that done we (I) can also move app-utils.i and any supporting scheme files to bindings as well. @gjanssens planned to do this early last year but demurred because I'd already started on this project.

Note that options.scm is now largely a backwards-compatibility shim: Nearly everything in it can be done, often more easily, directly using swigged functions. I have it in mind that we'll keep it through GnuCash 5 to give users time to adjust their custom reports and remove it in GnuCash 6.

PR #1207, opened 2021-12-03
[options.scm] gnc:make-budget-option stores guid instead of object

christopherlam:
Offshoot from #1191 

To reproduce bug:
- create budget
- create budget report, select it, run successfully
- delete budget
- reload budget report.

Result: segfault.

This fix stores guid instead of budget object, so, lookup a deleted guid returns null as it should. I haven't tested all combinations of saved-reports, gcm reports etc. But I'm sure it's safe.

### DIFF
--- a/libgnucash/app-utils/options.scm
+++ b/libgnucash/app-utils/options.scm
@@ -470,6 +470,13 @@ the option '~a'."))
 	 (selection-budget initial-budget)
          )
 
+
+    (define (convert-to-guid item)
+      (if (string? item) item (gncBudgetGetGUID item)))
+
+    (define (convert-to-budget item)
+      (if (string? item) (gnc-budget-lookup item (gnc-get-current-book)) item))
+
     (gnc:make-option
      section 
      name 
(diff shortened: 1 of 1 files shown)


jralls (approved):
Merge when ready.

christopherlam:
Iiuc this branch is incomplete. It does not modify the option scm->kvp and kvp->scm lambdas. I have no idea whether these are used anywhere.

libgnucash/app-utils/options.scm
```diff
@@ -470,6 +470,13 @@ the option '~a'."))
 	 (selection-budget initial-budget)
```
jralls, line 470: `gnc_budget_get_default` returns a `GncBudget*` so ISTM you need to change this to
```
(selection-budget (convert-to-guid initial-budget))
```

jralls:
They're used to set and retrieve the default budget on the budgeting page in File>Properties. The GUID conversion needs to be removed now that `selection-budget` is already a GUID.

PR #1209, opened 2021-12-05
Fetch all commodity namespaces always.

Takym:
**Motivation** & **Purpose**: Now, I cannot create a bank account for virtual currencies. This change helps to book transactions with virtual currencies.

P.S. I merely would like to propose to support virtual currencies.


jralls:
This change would break the price entry dialog. The function you're gutting is used only for that and for the QIF importer. How does it affect your ability to account for your crypto assets?

Takym:
I want to select custom virtual currencies, added with the security editor, at the currency picker for a bank account. However, there is only a type "Currencies" in the combo box.
image

PR #1213, opened 2021-12-06
Support the crypto-currencies

Takym:
Hello

I added a new namespace for crypto-currencies (virtual currencies).
This PR would be another resolution of #1209.

Thanks


fellen:
Hi, nice approach, but please see the discussion in #1073. We want reduce the number of hard coded name spaces, which for most users only clutter the selection box. So write a note there reminding me to mention crypto currencies in the examples and docs.

jralls:
What's more there's no need to add hard-coded namespaces. Security namespace is a free-text field. You can have as many as you want named anything you want *except* "Currencies", which is a special controlled namespace for ISO-4217 currencies.

PR #1214, opened 2021-12-12
Explain that most parts of online banking setup are from aqbanking

fellen:
Bug 798309 showed users don't know it and report problems to GnuCash.
Let's improve page 2 of the assistant:
image

@Bob-IT do we 
1. want the first commit
2. if yes is everything fine there?

General: Can we separate the URL from the text?

Perhaps Martin can create an English instance of that page?

### DIFF
--- a/gnucash/import-export/aqb/assistant-ab-initial.glade
+++ b/gnucash/import-export/aqb/assistant-ab-initial.glade
@@ -1,13 +1,13 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<!-- Generated with glade 3.36.0 -->
+<!-- Generated with glade 3.38.2 -->
 <interface>
   <requires lib="gtk+" version="3.22"/>
   <object class="GtkAssistant" id="aqbanking_init_assistant">
-    <property name="can_focus">False</property>
-    <property name="border_width">12</property>
+    <property name="can-focus">False</property>
+    <property name="border-width">12</property>
     <property name="title" translatable="yes">AqBanking Initial Assistant</property>
-    <property name="default_width">400</property>
-    <property name="default_height">500</property>
+    <property name="default-width">400</property>
+    <property name="default-height">500</property>
     <signal name="cancel" handler="aai_on_cancel" swapped="no"/>
     <signal name="close" handler="aai_on_finish" swapped="no"/>
     <signal name="prepare" handler="aai_on_prepare" swapped="no"/>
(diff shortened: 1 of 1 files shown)


Bob-IT:
@fellen 
1) Yes to first commit, it was my intention to go through all the glade files and open and save them with glade version 3.38.2 Will do that in January.
2) I would think so, probably add a GtkLinkButton, so increase number of items for wizard_page, add button with appropriate text and then reorder.

Bob-IT:
@fellen 
I had a quick look and the link button did not look right so I split the text up into four labels packed into the existing vbox.
With the third label set with markup and using a hyperlink anchor, all packing at start I came up with this.

Screenshot_2021-12-14_15-07-34

There is also a hyperlink on first page so you could make a similar change.

jralls:
@fellen, @Bob-IT string freeze is done, but this is also marked draft. Are you guys happy with it as-is?

fellen:
@Bob-IT The screenshot looks nice, perhaps we should move "You have to set up…" between the `Start Aqbanking…` and the `Continue` button.
Do you have the branch still around?
 
> There is also a hyperlink on first page so you could make a similar change.

of course

What me puzzled now: I can not go back from the 2. to the first page, but I imagine some users wish to reread the intro after they have seen page 2.

Bob-IT:
@fellen I am just in the middle of saving glade files with new version of glade so they are all consistent, should only take a couple of days and will redo my changes after that.

fellen:
> @fellen I am just in the middle of saving glade files with new version of glade so they are all consistent, should only take a couple of days and will redo my changes after that.

I wonder, if we should in that process also rename the first page of the assistants to `Introduction` instead of the assistant's title.

Bob-IT:
@fellen 
> I wonder, if we should in that process also rename the first page of the assistants to `Introduction` instead of the assistant's title.

Not sure what you mean, do you mean that all first pages should be changed to...
'Introduction to Online Banking Setup'
'Introduction to Importing Transactions'
'Introduction to Importing QIF files'
...
These would be the the initial assistant titles and left pane entry?

Bob-IT:
@fellen
These are the changes I have made as a patch file against current maint. There are also some cosmetic changes like removing references to wizard.
And the answer to your question is not that I can find.

[0001-My-Changes.txt](https://github.com/Gnucash/gnucash/files/7835094/0001-My-Changes.txt)

jralls:
Another string freeze is closing in on us. Can we get this merged soon?

jralls:
I've merged @Bob-IT's patch, which includes the string changes, into maint so that we can get the translators going on the clarification and getting rid of the word wizard.

I wonder, though, if we might improve the layout of both the first and second pages with [pango markup](https://docs.gtk.org/Pango/pango_markup.html)? Is that too technical for most translators?

@Bob-IT you might note that you can also [embed links](https://docs.gtk.org/gtk3/class.Label.html#links) in labels using markup.

PR #1215, opened 2021-12-12
Create crypto-currencies.md

Takym:
Hello @fellen,

I wrote a note for crypto-currencies.

Thanks


jralls:
@Takym That doesn't belong in the code base. It's a possible candidate for the wiki without the personal introduction.

PR #1222, opened 2021-12-21
Bug 798392 - Reports do not display when Reports JavaScript dependencies are located at a filepath that includes special characters like hash ("#")

christopherlam:
using a guile's uri.scm sanitizing function with expanded charset.

More test cases are available from https://github.com/Sporkmonger/Addressable/blob/master/spec/addressable/uri_spec.rb

### DIFF
--- a/gnucash/report/html-utilities.scm
+++ b/gnucash/report/html-utilities.scm
@@ -425,12 +425,12 @@
 (define (gnc:html-js-include file)
   (format #f
           "<script language=\"javascript\" type=\"text/javascript\" src=~s></script>\n"
-          (make-uri (gnc-path-find-localized-html-file file))))
+          (make-uri (gnc-resolve-file-path file))))
 
 (define (gnc:html-css-include file)
   (format #f
-          "<link rel=\"stylesheet\" type=\"text/css\" href=\"file:///~a\" />\n"
-          (gnc-path-find-localized-html-file file)))
+          "<link rel=\"stylesheet\" type=\"text/css\" href=~s />\n"
+          (make-uri (gnc-resolve-file-path file))))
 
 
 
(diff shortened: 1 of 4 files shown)


christopherlam:
Ok just checked Windows build, the HTML produced renders well on Firefox but not in gnucash. There's no access to JS console in libwebkit1 unfortunately.

```xml
<script language="javascript" type="text/javascript" src="file://C:/Program%20Files%20%28x86%29/gnucash/share/gnucash/chartjs/Chart.bundle.min.js"></script>
<h3 style='display:none'></h3><div style='width:100.0%;height:100.0%;'>
```

[x.zip](https://github.com/Gnucash/gnucash/files/7776517/x.zip)

jralls:
With the first commit it works for me on Windows in both a GnuCash reports tab and a Chrome window.

The second commit replaces `gnc-path-find-localized-html-file` with `gnc-resolve-file-path`. The latter is more direct than the former and can't be distracted with `$GNC_DOCS_PATH`.

I'm inclined to remove `gnc:html-css-file`: It's not used and the few css files we have are installed in reports directory where `gnc:html-css-file` isn't likely to find them.

gnucash/report/html-utilities.scm
```diff
@@ -413,11 +413,14 @@
...
+;;path must be absolute. On Windows an absolute path begins with a
+;;drive letter followed by a colon.
 (define (make-uri path)
-  (string-append
-   "file://"
-   (uri-encode (gnc:substring-replace path "\\" "/")
+  (let ((uri-path (uri-encode (gnc:substring-replace path "\\" "/")
                #:unescaped-chars unreserved-chars-rfc3986)))
+  (string-append
+   (if (eq? (car (string->list uri-path)) "/") "file://" "file:///")
```
christopherlam, line 422: `(eq? (car (string->list uri-path)) "/")` to `(char=? (string-ref uri-path 0) #\/)` will avoid creating intermediate list, and we'd want to be careful about equality functions.

https://www.gnu.org/software/guile/manual/html_node/Equality.html

jralls: OK, done.

PR #1229, opened 2021-12-31
Bug 798391 OFX import routine keeps prompting for accepted

jeanlaroche:
With some OFX_INCOME or OFX_REINVEST transactions, the FITID field was not always being set.


jralls:
The two existing calls to `gnc_set_split_online_id` are guarded by `if (data.fi_id_valid)`; shouldn't the new ones be too?

jeanlaroche:
Added if (data.fi_id_valid)

jralls:
Thanks!

PR #1227, opened 2021-12-30
[window-reconcile] speed up by postponing account balance recalc

christopherlam:
Reconciliation can be sped up tremendously by postponing all account
balance recalc when modifying splits' reconcile flags.

test: 
* open a large account.
* reconcile, finalize
* notice this is slow

by postponing account balance recalc until all splits' reconcile flags are set, we can speed this up tremendously.

PR #1230, opened 2022-01-01
Addendum to #1227 fd3ec14f6 optionally raises descendants' editlevel

christopherlam:
not sure if it's too clever.

### DIFF
--- a/gnucash/gnome/window-reconcile.c
+++ b/gnucash/gnome/window-reconcile.c
@@ -2257,6 +2257,19 @@ find_payment_account(Account *account)
     return NULL;
 }
 
+typedef void (*AccountProc) (Account *a);
+static void traverse_fn (Account *acct, AccountProc fn)
+{
+    fn (acct);
+}
+
+static void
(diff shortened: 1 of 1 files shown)


gnucash/gnome/window-reconcile.c
```diff
@@ -2288,10 +2301,10 @@ recnFinishCB (GtkAction *action, RecnWindow *recnData)
     recnData->delete_refresh = TRUE;
     account = recn_get_account (recnData);
 
-    xaccAccountBeginEdit (account);
+    acct_traverse_descendants (account, xaccAccountBeginEdit);
```
fellen, line 2304: It's indeed too smart for me. BeginEdit and CommitEdit are common database terminology. At least I would miss them.

christopherlam: There's no plan to remove BeginEdit and CommitEdit.

jralls (approved):
Looks fine to me.

PR #1232, opened 2022-01-04
Bug 798395 ofxdirectconnect downloads crash gnucash 4.9

jeanlaroche:
Bug 798395 ofxdirectconnect downloads crash gnucash 4.9
Add a check of temp_trans_list to avoid dereferencing a NULL pointer when there
are no transactions to match. Add a dialog to alert the user.

### DIFF
--- a/gnucash/import-export/import-main-matcher.c
+++ b/gnucash/import-export/import-main-matcher.c
@@ -456,6 +456,11 @@ gnc_gen_trans_list_show_all (GNCImportMainMatcher *info)
     // Set initial state of Append checkbox to same as last import for this account.
     // Get the import account from the first split in first transaction.
     temp_trans_list = info->temp_trans_list;
+    if (!temp_trans_list)
+    {
+        gnc_info_dialog (GTK_WINDOW (info->main_widget), _("While importing transactions found no new transactions."));
+        return;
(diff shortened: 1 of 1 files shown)


jralls:
Thanks!

gnucash/import-export/import-main-matcher.c
```diff
@@ -456,6 +456,11 @@ gnc_gen_trans_list_show_all (GNCImportMainMatcher *info)
     // Set initial state of Append checkbox to same as last import for this account.
     // Get the import account from the first split in first transaction.
     temp_trans_list = info->temp_trans_list;
+    if (!temp_trans_list)
+    {
+        gnc_info_dialog (GTK_WINDOW (info->main_widget), _("While importing transactions found no new transactions."));
```
fellen, line 461: @jralls Grammar?

fellen (commented):
Can we get a full sentence? English instead of Hackish, please.

gnucash/import-export/import-main-matcher.c
jralls: Is OK, though I suppose "Found no new transactions in this import." would be a bit less robotic-sounding.

PR #1236, opened 2022-01-05
Bug 798408 - The calculation in budgets UI and reports are wrong.

christopherlam:
gbv_get_accumulated_budget_amount returns a numeric, signed for DISPLAY and must be converted to unreversed before adding to info->total.

@miklcct should test this issue is fixed. Please test budgets existing in several account levels too.


christopherlam:
Thanks. I'm sure it will also fix issue with old broken budgets.

christopherlam:
@jralls merging this one into maint is possible, and will need a modification similar to 85a62c03ea when maint is merged into master. Unless you disagree I'll do the needful soon.

jralls:
@christopherlam go right ahead.

PR #1237, opened 2022-01-06
I18N: Add context to glossary files

fellen:
This starts a normalization of gnc-glossary.txt by using msgctxt.

To see the result run  `po/glossary/txt-to-pot.sh po/glossary/gnc-glossary.txt > po/glossary/gnc-glossary.pot`

Any suggestions for
- better contexts,
- additional terms
- …?

Todo:
- [ ] preparing patch for msgmerge


po/glossary/gnc-glossary.txt
```diff
@@ -1,212 +1,212 @@
...
+"account type"	"Expense"	"-"
+"account type"	"Income"	"-"
+"account type"	"Liability"	"A debt, a financial obligation, but see also 'account type: Passive'"
+"account type"	"money-market"	"-"
+"account type"	"Mutual fund"	"-"
+"account type"	"Passive"	"The right side of the balance sheet in T account form shows the source of funds and contains equity & liability. While not common in english, most languages would translate 'equity & liability' with 'passive'. Complement: Active. See also: Report Form Implementation: https://bugs.gnucash.org/show_bug.cgi?id=421766"
+"account type"	"Profit & Loss"	"Group of accounts tracking your success, complement of 'Assets & Liabilities'"
+"account type"	"saving"	"1. (US) any type of account that earns interest 2. (Brit) any type of bank account that earns a higher level of interest than a current account or deposit account"
+"account type"	"Stock"	"-"
+"account type"	"trading"	"This account type (new in gnucash-2.4.0) is used when exchanging or trading amounts from one currency into another"
```
jralls, line 22: We can lose the aside about GnuCash 2.4, that was long enough ago that no translator will care.

po/glossary/gnc-glossary.txt
```diff
@@ -1,212 +1,212 @@
...
-"value (in a split)"	"The worth of sth in terms of money or other commodities for which it can be exchanged"
-"vendor"	"In small business accounting: A person or company that sells items and is supplying goods"
-"voucher"	"The terms 'Voucher' and 'Expense Voucher' are used interchangeably in gnucash. The 'Expense Voucher' is also a bit of a misnomer -- it's more like an 'Expense Report' in gnucash.  The phrase is meant to be a list of expenses incurred by an employee for which the company will reminburse them."
-"withdraw (in the reconcile dialog)"	"see debit"
+"Context [optional]"	"Term"	"English Definition or explanation. Purpose of this record: document the file format for maintainers—no need to translate it."
+"Quoting convention"	"\"\""	"Opening and closing quote symbols and optionally their key combos like [altgr]+[Y]/[X]. Define the preferred style of quotation, see https://en.wikipedia.org/wiki/Quotation_mark#Summary_table"
+	"account"	"A detailed record of money spent and received"
+	"account code"	"An alphanumerical key of an account in GnuCash, not at the bank, can be used to sort. Some templates provide them or the user can enter them."
```
jralls, line 4: "An alphanumerical code applied to accounts for sorting. Some jurisdictions mandate codes and some GnuCash book templates include codes as part of the tree structure. Account codes can be set and changed in the Edit Account dialog."

po/glossary/gnc-glossary.txt
```diff
@@ -1,212 +1,212 @@
...
+	"account hierarchy"	"the tree view of all accounts"
+	"account name"	"-"
+"account type"	"Active"	"The left side of the balance sheet in T account form shows the application of funds in form of assets. Because it contains only assets use assets directly. Complement: Passive. See also: Report Form"
```
jralls, line 7: This is US English and that description won't make sense in to a US-based accountant. How about "The grouping of accounts whose balance increases with a debit. Complement: Passive."

thereeroyz (approved):

po/glossary/gnc-glossary-new.pot
```diff
@@ -0,0 +1,965 @@
...
+msgstr ""
+
+#. "The thing that the scatter plot uses to mark each data point"
+msgid "marker"
+msgstr ""
+
+#. "1. Some text annotation, but this meaning isn't used inside gnucash. 2. In the Customer summary report: The ratio of profit vs. sales, i.e. the profit amount divided by the sales amount, shown in percent."
+msgid "markup"
+msgstr ""
+
```
brianhsu, line 570: These two meaning are for quite different context. Maybe we should separate them to different context?

At least, in Traditional Chinese, they are translated to very different term. First one is `標籤` / second one is `純益率`。

Also, I believe that "this meaning isn't used inside gnucash" is not correct. As we have something like the following in the report configuration dialog:

> Notes added at end of invoice -- may contain HTML markup

fellen: Right, should get a context. `text` and `financial` OK?

jralls:
Closing in favor of #2287

PR #1240, opened 2022-01-07
Bug 798406 - 4.9 inappropriately converts dollars to shares at $1 a piece in pure cash transactions

christopherlam:
`xaccSplitGetSharePrice` returns price==0 instead of price==1 and don't save price==0 into pricedb.


jralls:
Looks good, works the way I expect.

PR #1242, opened 2022-01-09
[gnc-plugin-page-budget] estimate/allperiods wrap begin/commit pair

christopherlam:
@miklcct for https://bugs.gnucash.org/show_bug.cgi?id=798407 would you be able to verify this speeds up estimate/allperiods? I do not have access to a mysql server.


miklcct:
Tried this build - it's better now with setting all periods takes 26 seconds on my database on a remote VPS in the same country.

christopherlam:
> Tried this build - it's better now with setting all periods takes 26 seconds on my database on a remote VPS in the same country.

Nice, what was the previous benchmark?

PR #1243, opened 2022-01-09
pt_PT translation update

pmraps:
Added check book and business accounts trees


fellen:
Thanks, Pedro!
I have squashed this PR into commit de935c6, but had written there 1234 instead of 1243.

I wonder, why you have not set all accounts, which have subaccounts
```
    <slot>
      <slot:key>placeholder</slot:key>
      <slot:value type="string">true</slot:value>
    </slot>

``` 
Do you really want to have them in the list of target accounts?

PR #1094, opened 2021-07-21
Bug753307 - Custom Report be selectable as default Report for Printing

Bob-IT:
This commit changes the preference in Business->'Report for Printing' to be saved as a book property and allow the selection of any Invoice Report to be used as the default. This still needs work...

At the moment it works by returning all reports and I have successfully selected a new report scheme file, saved configuration and the existing 4 invoice reports. If the saved report can not be found, I have added an entry to the combo list with the missing guid and present a warning image with tool tip.

I am not convinced I have added to the correct file, 'dialog-utils.*', I think it probably should be in 'business-gnome-utils.*' but that is in the wrong directory for 'dialog-preference' to access it. I have not tried moving 'business-gnome-utils.*' to the 'gnome-utils' folder but that may be an option.

@christopherlam  Can you have a look at the required scheme changes to filter the list, at the moment I was able to work out how to return all reports but this should be limited to just invoice ones. Also for the function 'gnc_invoice_window_print_invoice', a test needs to be made for valid returned guid, maybe you can add some scheme code to return true/false to be combined with the test for 'reportname'.

Any further changes I make will be after the required scheme changes to make it useful.

As an after thought, I did wonder if this sort of change could be added for other similar use reports but the only one that may make sense is the 'Account Report'.

PR #1247, opened 2022-01-11
Allow to select saved report on print invoice button.

mildred:
Use case : I want to be able to select a saved report I made to print
my invoices without having to open the saved report, change the option
and choose the invoice number. I also want to be able to print some of
my invoices using a specific template I made for quotes.

This change opens up the saved invoice dialog when the print invoice
button is clicked (only if the settings are set for this) and when I
select the saved report, the invoice report is created with the given
defaults and the correct invoice number.

### DIFF
--- a/gnucash/gnome/dialog-custom-report.c
+++ b/gnucash/gnome/dialog-custom-report.c
@@ -77,6 +77,8 @@ typedef struct _CustomReportDialog
 
     /* data */
     SCM reportlist;
+    GncCustomReportSelected selected_cb;
+    gpointer selected_userdata;
 
 } CustomReportDialog;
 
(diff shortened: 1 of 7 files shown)


gjanssens:
Thanks for the PR. I had a cursory look at it it's a decent start.

I do have a few design and user experience questions/remarks though first:

1. Your code at some point opens the saved reports window for the user to select a specific saved report to use. What happens if the user selects a saved report that's not an invoice template ?
2. If I understand your implementation correctly, you have added an extra option to the list of reports in which the user selects the preferred layout to use when printing for his/her invoices. So if you "want to print some of [your] invoices using a specific template [you] made for quotes", that means you have to regularly go to preferences and adjust the report selection (to switch between a saved report or a standard one). That seems a bit cumbersome to me (though of course less cumbersome as to manually configure a report window each time just to be able to print an invoice).

Your use case has some similarity to PR #1094 . That one is about extending the list of possible candidates for invoice printing with compatible saved reports (setting one as default). Yours is about occasionally overriding that (extended) default. Perhaps both could be designed together ?

That aside I would suggest to rather implement printing with a non-standard template as a separate menu item. That gives users the convenience of a good default and a fast access to the occasional non-standard printout. Or put differently that would avoid  both a separate option and the need to go change it all the time.

Bob-IT:
Thank you for the PR, I think I have incorporated your idea in to my PR which will be in version 5.00.
If you want to try it, have a look at the pre-release version 4.901 that should be out this Sunday.

PR #1248, opened 2022-01-12
Budget backend efficiency improvements.

christopherlam:
The first time a budget value or note is accessed, the whole kvp tree is loaded onto ~memory~struct. 

All subsequent value or note access are speedy.

Writes to value or note are written to backend *and* ~memory~struct.

Draft. ~Leaky.~

### DIFF
--- a/libgnucash/engine/gnc-budget.cpp
+++ b/libgnucash/engine/gnc-budget.cpp
@@ -30,6 +30,9 @@
 #include <qof.h>
 #include <qofbookslots.h>
 #include <qofinstance-p.h>
+#include <unordered_map>
+#include <vector>
+#include <memory>
 
 #include "Account.h"
 
@@ -59,6 +62,16 @@ typedef struct
     QofInstanceClass parent_class;
 } BudgetClass;
 
+struct PeriodData
+{
+    std::string note;
+    bool value_is_set;
+    gnc_numeric value;
+};
+
+using PeriodDataVec = std::vector<PeriodData>;
+using AcctMap = std::unordered_map<const Account*, PeriodDataVec>;
+
 typedef struct GncBudgetPrivate
 {
     /* The name is an arbitrary string assigned by the user. */
@@ -70,7 +83,7 @@ typedef struct GncBudgetPrivate
     /* Recurrence (period info) for the budget */
     Recurrence recurrence;
 
-    GHashTable *acct_hash;
+    std::unique_ptr<AcctMap> acct_map;
 
     /* Number of periods */
     guint  num_periods;
@@ -87,24 +100,6 @@ struct _GncBudgetClass
 /* GObject Initialization */
 G_DEFINE_TYPE_WITH_PRIVATE(GncBudget, gnc_budget, QOF_TYPE_INSTANCE)
 
-typedef struct PeriodData
-{
-    gchar *note;
-    gboolean value_is_set;
-    gnc_numeric value;
-} PeriodData;
-
-static void acct_array_free (GArray *acct_array, gpointer user_data)
-{
-    for (guint i = 0; i < acct_array->len; i++)
-    {
-        PeriodData *perioddata = &g_array_index (acct_array, PeriodData, i);
-        if (perioddata->note)
-            g_free (perioddata->note);
-    }
-    g_array_free (acct_array, FALSE);
-}
-
 static void
 gnc_budget_init(GncBudget* budget)
 {
@@ -114,9 +109,7 @@ gnc_budget_init(GncBudget* budget)
     priv = GET_PRIVATE(budget);
     priv->name = CACHE_INSERT(_("Unnamed Budget"));
     priv->description = CACHE_INSERT("");
-
-    priv->acct_hash = g_hash_table_new_full (g_direct_hash, g_direct_equal, NULL,
-                                             (GDestroyNotify) acct_array_free);
+    priv->acct_map = std::make_unique<AcctMap>();
 
     priv->num_periods = 12;
     date = gnc_g_date_new_today ();
@@ -286,8 +279,7 @@ gnc_budget_free(QofInstance *inst)
 
     CACHE_REMOVE(priv->name);
     CACHE_REMOVE(priv->description);
-
-    g_hash_table_destroy (priv->acct_hash);
+    priv->acct_map = nullptr;   // nullify to ensure unique_ptr is freed.
 
     /* qof_instance_release (&budget->inst); */
     g_object_unref(budget);
@@ -471,13 +463,6 @@ gnc_budget_get_guid(const GncBudget* budget)
     return qof_instance_get_guid(QOF_INSTANCE(budget));
 }
 
-static void
-set_num_periods (GArray *perioddata, gpointer user_data)
-{
-    gint num_periods = GPOINTER_TO_INT (user_data);
-    g_array_set_size (perioddata, num_periods);
-}
-
 void
 gnc_budget_set_num_periods(GncBudget* budget, guint num_periods)
 {
@@ -493,8 +478,12 @@ gnc_budget_set_num_periods(GncBudget* budget, guint num_periods)
     qof_instance_set_dirty(&budget->inst);
     gnc_budget_commit_edit(budget);
 
-    g_hash_table_foreach (priv->acct_hash, (GHFunc) set_num_periods,
-                          GUINT_TO_POINTER (num_periods));
+    std::for_each (priv->acct_map->begin(),
+                   priv->acct_map->end(),
+                   [num_periods](auto& it)
+                   {
+                       it.second.resize(num_periods);
+                   });
 
     qof_event_gen( &budget->inst, QOF_EVENT_MODIFY, NULL);
 }
@@ -516,8 +505,9 @@ make_period_path (const Account *account, guint period_num, char *path1, char *p
     g_sprintf (path2, "%d", period_num);
 }
 
-static GArray* get_acct_array (const GncBudget *budget, const Account *account);
-
+static PeriodData& get_perioddata (const GncBudget *budget,
+                                   const Account *account,
+                                   guint period_num);
 
 /* period_num is zero-based */
 /* What happens when account is deleted, after we have an entry for it? */
@@ -525,17 +515,15 @@ void
 gnc_budget_unset_account_period_value(GncBudget *budget, const Account *account,
                                       guint period_num)
 {
-    GArray *acct_array;
     gchar path_part_one [GUID_ENCODING_LENGTH + 1];
     gchar path_part_two [GNC_BUDGET_MAX_NUM_PERIODS_DIGITS];
-    PeriodData *perioddata;
 
     g_return_if_fail (budget != NULL);
     g_return_if_fail (account != NULL);
+    g_return_if_fail (period_num < GET_PRIVATE(budget)->num_periods);
 
-    acct_array = get_acct_array (budget, account);
-    perioddata = &g_array_index (acct_array, PeriodData, period_num);
-    perioddata->value_is_set = FALSE;
+    auto& data = get_perioddata (budget, account, period_num);
+    data.value_is_set = false;
 
     make_period_path (account, period_num, path_part_one, path_part_two);
 
@@ -556,8 +544,6 @@ gnc_budget_set_account_period_value(GncBudget *budget, const Account *account,
 {
     gchar path_part_one [GUID_ENCODING_LENGTH + 1];
     gchar path_part_two [GNC_BUDGET_MAX_NUM_PERIODS_DIGITS];
-    GArray *acct_array;
-    PeriodData *perioddata;
 
     /* Watch out for an off-by-one error here:
      * period_num starts from 0 while num_periods starts from 1 */
@@ -570,8 +556,7 @@ gnc_budget_set_account_period_value(GncBudget *budget, const Account *account,
     g_return_if_fail (budget != NULL);
     g_return_if_fail (account != NULL);
 
-    acct_array = get_acct_array (budget, account);
-    perioddata = &g_array_index (acct_array, PeriodData, period_num);
+    auto& perioddata = get_perioddata (budget, account, period_num);
 
     make_period_path (account, period_num, path_part_one, path_part_two);
 
@@ -579,7 +564,7 @@ gnc_budget_set_account_period_value(GncBudget *budget, const Account *account,
     if (gnc_numeric_check(val))
     {
         qof_instance_set_kvp (QOF_INSTANCE (budget), NULL, 2, path_part_one, path_part_two);
-        perioddata->value_is_set = FALSE;
+        perioddata.value_is_set = false;
     }
     else
     {
@@ -588,8 +573,8 @@ gnc_budget_set_account_period_value(GncBudget *budget, const Account *account,
         g_value_set_boxed (&v, &val);
         qof_instance_set_kvp (QOF_INSTANCE (budget), &v, 2, path_part_one, path_part_two);
         g_value_unset (&v);
-        perioddata->value_is_set = TRUE;
-        perioddata->value = val;
+        perioddata.value_is_set = true;
+        perioddata.value = val;
     }
     qof_instance_set_dirty(&budget->inst);
     gnc_budget_commit_edit(budget);
@@ -627,9 +612,8 @@ gnc_budget_is_account_period_value_set (const GncBudget *budget,
                                         const Account *account,
                                         guint period_num)
 {
-    GArray *array = get_acct_array (budget, account);
-    PeriodData *data = (PeriodData*) &g_array_index (array, PeriodData, period_num);
-    return data->value_is_set;
+    g_return_val_if_fail (period_num < GET_PRIVATE(budget)->num_periods, false);
+    return get_perioddata (budget, account, period_num).value_is_set;
 }
 
 static gnc_numeric
@@ -661,9 +645,13 @@ gnc_budget_get_account_period_value (const GncBudget *budget,
                                      const Account *account,
                                      guint period_num)
 {
-    GArray *array = get_acct_array (budget, account);
-    PeriodData *data = (PeriodData*) &g_array_index (array, PeriodData, period_num);
-    return data->value_is_set ? data->value : gnc_numeric_zero ();
+    g_return_val_if_fail (period_num < GET_PRIVATE(budget)->num_periods,
+                          gnc_numeric_zero());
+    auto& data = get_perioddata (budget, account, period_num);
+    if (!data.value_is_set)
+        return gnc_numeric_zero();
+
+    return data.value;
 }
 
 void
@@ -672,8 +660,6 @@ gnc_budget_set_account_period_note(GncBudget *budget, const Account *account,
 {
     gchar path_part_one [GUID_ENCODING_LENGTH + 1];
     gchar path_part_two [GNC_BUDGET_MAX_NUM_PERIODS_DIGITS];
-    GArray *acct_array;
-    PeriodData *perioddata;
 
     /* Watch out for an off-by-one error here:
      * period_num starts from 0 while num_periods starts from 1 */
@@ -686,17 +672,16 @@ gnc_budget_set_account_period_note(GncBudget *budget, const Account *account,
     g_return_if_fail (budget != NULL);
     g_return_if_fail (account != NULL);
 
-    acct_array = get_acct_array (budget, account);
-    perioddata = &g_array_index (acct_array, PeriodData, period_num);
+    auto& perioddata = get_perioddata (budget, account, period_num);
 
     make_period_path (account, period_num, path_part_one, path_part_two);
 
     gnc_budget_begin_edit(budget);
-    if (perioddata->note)
-        g_free (perioddata->note);
-    perioddata->note = g_strdup (note);
     if (note == NULL)
+    {
         qof_instance_set_kvp (QOF_INSTANCE (budget), NULL, 3, GNC_BUDGET_NOTES_PATH, path_part_one, path_part_two);
+        perioddata.note.clear ();
+    }
     else
     {
         GValue v = G_VALUE_INIT;
@@ -705,6 +690,7 @@ gnc_budget_set_account_period_note(GncBudget *budget, const Account *account,
 
         qof_instance_set_kvp (QOF_INSTANCE (budget), &v, 3, GNC_BUDGET_NOTES_PATH, path_part_one, path_part_two);
         g_value_unset (&v);
+        perioddata.note = note;
     }
     qof_instance_set_dirty(&budget->inst);
     gnc_budget_commit_edit(budget);
@@ -713,21 +699,22 @@ gnc_budget_set_account_period_note(GncBudget *budget, const Account *account,
 
 }
 
-static gchar *
+static std::string
 get_account_period_note (const GncBudget *budget,
                          const Account *account, guint period_num)
 {
     gchar path_part_one [GUID_ENCODING_LENGTH + 1];
     gchar path_part_two [GNC_BUDGET_MAX_NUM_PERIODS_DIGITS];
     GValue v = G_VALUE_INIT;
-    gchar *retval;
+    std::string retval;
 
-    g_return_val_if_fail(GNC_IS_BUDGET(budget), NULL);
-    g_return_val_if_fail(account, NULL);
+    g_return_val_if_fail (GNC_IS_BUDGET(budget), "");
+    g_return_val_if_fail (account, "");
 
     make_period_path (account, period_num, path_part_one, path_part_two);
     qof_instance_get_kvp (QOF_INSTANCE (budget), &v, 3, GNC_BUDGET_NOTES_PATH, path_part_one, path_part_two);
-    retval = G_VALUE_HOLDS_STRING (&v) ? g_value_dup_string(&v) : NULL;
+    if (G_VALUE_HOLDS_STRING (&v))
+        retval = g_value_get_string(&v);
     g_value_unset (&v);
     return retval;
 }
@@ -737,9 +724,9 @@ gchar *
 gnc_budget_get_account_period_note (const GncBudget *budget,
                                     const Account *account, guint period_num)
 {
-    GArray *array = get_acct_array (budget, account);
-    PeriodData *data = (PeriodData*) &g_array_index (array, PeriodData, period_num);
-    return g_strdup (data->note);
+    g_return_val_if_fail (period_num < GET_PRIVATE(budget)->num_periods, nullptr);
+    auto& data = get_perioddata (budget, account, period_num);
+    return data.note.empty () ? nullptr : g_strdup (data.note.c_str());
 }
 
 time64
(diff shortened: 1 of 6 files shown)


christopherlam:
@jralls some unusual bugs in here... this branch seems to pass on mine and mac, but not on CI linux. Any off-by-one error you can see?

Also still buggy - if array is shortened, the obsolete array element structs are not yet freed properly. C++ would be nicer...

jralls:
> The first time a budget value or note is accessed, the whole kvp tree is loaded onto memory.

Uh, no, everything including KVP is loaded into memory at session start. You're just copying it. With the C iinterface to KVP that involves GValues, which are kinda slow.

>  C++ would be nicer...

Always. So make it C++.

> this branch seems to pass on mine and mac

As you no doubt saw the failure is 
```
In unknown file:
           0 (+ 0 0 0 0 0 0 #f)

In procedure +: Wrong type: #f
```

Seems strange that that would differ between Linux and macOS, but no matter. Where's that list coming from?

christopherlam:
> > this branch seems to pass on mine and mac
> 
> As you no doubt saw the failure is
> 
> ```
> In unknown file:
>            0 (+ 0 0 0 0 0 0 #f)
> 
> In procedure +: Wrong type: #f
> ```
> 
> Seems strange that that would differ between Linux and macOS, but no matter. Where's that list coming from?

from `gnc:get-account-period-rolledup-budget-value` which calls `gnc:budget-account-get-rolledup-net` which I'll need to review.

jralls:
@christopherlam Can you rename this PR to something a little less cryptic than "maint lightning budget"? What the heck is "maint lightning budget" supposed to mean?

christopherlam:
> @christopherlam Can you rename this PR to something a little less cryptic than "maint lightning budget"? What the heck is "maint lightning budget" supposed to mean?

It was meant to make the budget editor much faster on local or remote mysql. However this is not possible because editing cells need to write to kvp.

IIUC renaming a branch means closing down PR.

I will test budget editor with an overarching `gnc_budget_begin_edit` throughout the editing, and `gnc_budget_commit_edit` when running the report or closing the tab.

jralls:
> IIUC renaming a branch means closing down PR.

Nope, just click the edit button next to the title and write a new one, as I did just now.  By all means change it again if I didn't capture your intent correctly. The PR title and branch name are completely independent, the former should emphasize explaining the overall purpose of the PR while the latter is a mnemonic that you  can type quickly when interacting with git on the command line.

> I will test budget editor with an overarching gnc_budget_begin_edit throughout the editing, and gnc_budget_commit_edit when running the report or closing the tab.

That's likely to be a more successful approach, but don't throw away your engine changes. Even though they're not significant for a remote SQL backend they're still worthwhile.

libgnucash/engine/gnc-budget.cpp
```diff
@@ -461,6 +475,9 @@ gnc_budget_set_num_periods(GncBudget* budget, guint num_periods)
     qof_instance_set_dirty(&budget->inst);
     gnc_budget_commit_edit(budget);
 
+    for (auto& it : priv->acct_map)
+        it.second.resize (num_periods);
```
christopherlam, line 479: Resizing a std::vector may discard some `PeriodData` structs. The latter contains an std::string. Can this resize dive into the struct to free the std::string properly?

miklcct: This should be done in the destructor of the struct.

christopherlam:
> > It was meant to make the budget editor much faster on local or remote mysql. However this is not possible because editing cells need to write to kvp.
> 
> Are you talking about as single cell edit or using the 'Estimate/All Periods' buttons? Have not looked at this for a long time but you may have to disconnect the tree view from the model, do the updates and then reconnect.

I'm referring to the whole budget editor tab. It _could_ be initialized with `gnc_budget_begin_edit`, and all updates (cell editor / estimate / all periods) will be fast. Tab closure or Tab switch or "Run Report" will then need to `gnc_budget_commit_edit`

libgnucash/engine/gnc-budget.cpp
jralls: >  Can this resize dive into the struct to free the std::string properly?

Yes. The vector contains actual structs as it should, and PeriodData holds the actual strings so removing the elements from the vector destroys all of the struct's contents including the std::string. The std::string manages its character memory so you don't have to do anything.

> This should be done in the destructor of the struct.

It should not. There's nothing to do, the struct holds the string itself so when the PeriodData is destroyed the string goes out of scope and is automatically destroyed. Had the struct held a naked std::string* *then* `~PeriodData` would need to free it, but since C++11 naked pointers are considered [bad practice](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rr-newdelete), one [should use smart pointers](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rr-owner), preferably [std::unique_ptr](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rr-make_unique) instead.

libgnucash/engine/gnc-budget.cpp
```diff
@@ -65,6 +65,12 @@ struct PeriodData
     std::string note;
     bool value_is_set;
     gnc_numeric value;
+
+    ~PeriodData()
+    {
+        note.clear ();
+        note.shrink_to_fit ();
+    };
```
jralls: This is harmless but unnecessary. When the `PeriodData` is destroyed `note` will be too. std::string is responsible for its character memory, you don't need to do anything.

christopherlam:
@jralls see last commit -- has bug whereby setters are not working. try this branch, open budget editor, and modify any cell. logger will show the new value isn't being saved. any clues?

* 19:02:54  WARN <gnc.engine> [gnc_budget_set_account_period_value()] A. vec=0x7ffceea96440, perioddata=0x7ffceea96480, note='', value=33/1
* 19:02:54  WARN <gnc.engine> [gnc_budget_set_account_period_value()] B. vec=0x7ffceea96460, perioddata=0x7ffceea964c0, note='', value=0/1

christopherlam:
Done. There's potential for dataloss though; open a populated budget, edit options to reduce num_periods, then increase num_periods back to original. Previously the old data would reappear; but thanks to std::vector.resize() the old data is reset.

jralls (changes requested):
Pretty good for a first foray into C++.

The main source of your early problems is not realizing the difference between passing by pointer, reference (a non-nullable pointer), and value. Returning values makes a copy of the return variable so assigning to it or one of its members doesn't change the value in the struct. The beauty of C++'s `&` and `auto` is that you need change only the return type of the accessor function and everything works as references with no copying.

Note that `auto&` works *only* when you're assigning from another variable, not from a function return.

Move semantics can be really complicated but the basic idea is that you're transferring the memory block owned by one variable to another variable instead of copying the contents from one to the other. After a move the donor variable's contents are undefined.

Skipping the GValue dance will hugely speed up KVP access. The example I provided could be made cleaner with some API work on QofInstance and KVP, but that's beyond the scope of this PR.

libgnucash/engine/gnc-budget.cpp
```diff
@@ -623,33 +700,48 @@ gnc_budget_set_account_period_note(GncBudget *budget, const Account *account,
...
-    g_return_val_if_fail(GNC_IS_BUDGET(budget), NULL);
-    g_return_val_if_fail(account, NULL);
+    g_return_val_if_fail (GNC_IS_BUDGET(budget), "");
+    g_return_val_if_fail (account, "");
 
     make_period_path (account, period_num, path_part_one, path_part_two);
     qof_instance_get_kvp (QOF_INSTANCE (budget), &v, 3, GNC_BUDGET_NOTES_PATH, path_part_one, path_part_two);
-    retval = G_VALUE_HOLDS_STRING (&v) ? g_value_dup_string(&v) : NULL;
+    if (G_VALUE_HOLDS_STRING (&v))
+        retval = g_value_get_string(&v);
```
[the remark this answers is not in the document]
bhardwajs: @jralls has it right. Another option would be to use `std::move` which will pass the ownership of the string to the caller.

libgnucash/engine/gnc-budget.cpp
```diff
@@ -674,6 +766,42 @@ gnc_budget_get_account_period_actual_value(
...
+
+    return map_iter->second;
+}
+
+static void set_perioddata_vec (const GncBudget *budget,
+                                const Account *account,
+                                PeriodDataVec vec)
+{
+    auto map = GET_PRIVATE (budget)->acct_map;
+    map->insert_or_assign (account, vec);
```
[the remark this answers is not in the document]
bhardwajs: @jralls - `insert_or_assign` is C++ 17. Is gnucash already on C++ 17?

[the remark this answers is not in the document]
jralls: @bhardwajs Yes, since 4.0.

libgnucash/engine/gnc-budget.cpp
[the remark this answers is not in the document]
jralls: >  Another option would be to use std::move which will pass the ownership of the string to the caller.

No, it guts the string in the KVP slot (which would be bad, it wouldn't be usable afterwards) and puts the memory block containing the characters in the receiving string. Ownership applies to pointers, not to objects.

libgnucash/engine/gnc-budget.cpp
```diff
@@ -481,6 +500,12 @@ make_period_path (const Account *account, guint period_num, char *path1, char *p
     g_sprintf (path2, "%d", period_num);
 }
 
+static PeriodDataVec get_perioddata_vec (const GncBudget *budget,
```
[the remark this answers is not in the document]
christopherlam: Ok if we're able to get the original _reference_ then this function can become
`static PeriodData& get_perioddata (GncBudget *budget, Account *account, guint period_num)`.

libgnucash/engine/gnc-budget.cpp
```diff
@@ -461,6 +477,9 @@ gnc_budget_set_num_periods(GncBudget* budget, guint num_periods)
     qof_instance_set_dirty(&budget->inst);
     gnc_budget_commit_edit(budget);
 
+    for (auto& it : *(priv->acct_map))
+        it.second.resize (num_periods);
```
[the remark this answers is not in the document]
christopherlam: What's the big advantage of `for_each` vs `for`?
https://stackoverflow.com/questions/2047414/advantages-of-stdfor-each-over-for-loop
https://www.reddit.com/r/cpp/comments/fadeu2/for_each_vs_for/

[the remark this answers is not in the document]
jralls: You asked about templates on IRC. One of the interesting features of templates is that one can write explicit specializations that can work differently from the generic definition. The standard library implementors do that alot.

So the big difference between `std::for_each()` and range-for is that the former is a template function in the standard library that can be specialized depending on the type of the iterator that you pass to it and the latter is a non-templated language feature. `for (auto foo : bar)` is just syntactic sugar for `for (auto foo = bar.begin<forward_iterator>(); foo != bar.end(); ++foo)`. A `std::for_each<unordered_map<foo>>` implementation could use a different kind of iterator and a different mechanism for traversing them if the structure of the collection warrants it.

Of course from your pov the big benny is that `for_each` is more functional. ;-)

libgnucash/engine/gnc-budget.cpp
```diff
@@ -481,6 +500,12 @@ make_period_path (const Account *account, guint period_num, char *path1, char *p
...
 }
 
+static PeriodDataVec get_perioddata_vec (const GncBudget *budget,
+                                         const Account *account);
+
+static void set_perioddata_vec (const GncBudget *budget,
+                                const Account *account, PeriodDataVec vec);
+
 /* period_num is zero-based */
 /* What happens when account is deleted, after we have an entry for it? */
```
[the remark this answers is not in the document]
christopherlam: It doesn't store a copy of acct pointer. It uses Account* as key. If the code is:

```cpp
auto data = get_perioddata_vec (budget, acct1, periodnum)
xaccFreeAccount (acct1)
data.note = "s"
```
then this data still exists as an inaccessible record which will eventually be purged when AcctMap destructors launch?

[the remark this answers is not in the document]
jralls: Ah, right. What's persistent is the account's gncGUID in KVP, which also becomes unreachable. Is there a scrub_budget that removes slots whose gncGUID isn't in the account collection?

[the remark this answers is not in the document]
christopherlam: ^ of course not :) if an account is deleted, the budget data is left forever in datafile IIUC. This applies before and after this PR.

libgnucash/engine/gnc-budget.cpp
```diff
@@ -758,6 +758,8 @@ get_perioddata (const GncBudget *budget, const Account *account, guint period_nu
 {
     GncBudgetPrivate *priv = GET_PRIVATE (budget);
 
+    g_assert (period_num < priv->num_periods);
```
jralls, line 765: Unfortunately g_assert is a no-op in a release build. Since this is a static function the best fix would be to use `return vec.at(period_num);` instead of `return vec[period_num);` and handle the `std::out_of_range` exception at each call of `get_perioddata()`.

libgnucash/engine/gnc-budget.cpp
```diff
@@ -781,8 +781,7 @@ get_perioddata (const GncBudget *budget, const Account *account, guint period_nu
                 data.value = get_account_period_value (budget, account, i);
             vec.emplace_back (data);
         }
-        map->insert_or_assign (account, vec);
-        return vec[period_num];
+        map_iter = map->emplace(account, vec).first;
```
jralls, line 784: Nothing weird here, clang just does a better job of recognizing that you're trying to return a reference to something that won't be valid once the function exits. It's always nicer to get a compile error than a crash!

`insert` instead of `emplace`: The latter is for constructing a new element from the element type's constructor arguments, something you can't do in this case. Same for creating the vector element: If you construct the vector yourself use `push_back`; in this case you *could* do
```
   auto value_set = s_account_period_value_set(budget, account, i)
   vec.emplace_back(get_account_period_note(budget, account, i),
                                    is_set, is_set ? get_account_period_value(budget, account, i) : gnc_numeric_zero());
```
When you construct the value separately you should pass it wrapped with `std::move`, e.g. `map_iter = map->insert(account, std::move(vec)).first;` to transfer `vec`'s contents instead of copying them.

libgnucash/engine/gnc-budget.cpp
jralls: Or you could `thow std::out_of_range{}` yourself instead of the `assert`.

libgnucash/engine/test/utest-Budget.c
```diff
@@ -96,10 +96,10 @@ test_gnc_set_budget_num_periods_data_retention ()
     gnc_budget_set_num_periods(budget, 10);
     gnc_budget_set_num_periods(budget, 20);
 
-    /* value and note are retained */
-    g_assert (gnc_budget_is_account_period_value_set(budget, acc, 15));
+    /* value and note are lost */
+    g_assert (!gnc_budget_is_account_period_value_set(budget, acc, 15));
     note = gnc_budget_get_account_period_note (budget, acc, 11);
-    g_assert_cmpstr (note, ==, "undefined");
+    g_assert_cmpstr (note, ==, NULL);
```
christopherlam, line 102: I'd think this change of behaviour isn't considered unacceptable...

libgnucash/engine/gnc-budget.cpp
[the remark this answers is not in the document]
christopherlam: I guess a budget scrub would do `qof_collection_foreach (budget, purge_obsolete_accounts)`, and `purge_obsolete_accounts` would also purge values and notes whose period_nums >= num_periods...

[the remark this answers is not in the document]
jralls: That sounds reasonable.

libgnucash/engine/test/utest-Budget.c
jralls: I don't really understand the point. Asserts are for detecting programming errors when running unit tests and aren't much use for errors in the wild.

christopherlam:
sadly valgrind reports a leak somewhere...

```
==471166== 5,175 (104 direct, 5,071 indirect) bytes in 1 blocks are definitely lost in loss record 14,923 of 15,092
==471166==    at 0x4843FB3: operator new(unsigned long) (in /usr/libexec/valgrind/vgpreload_memcheck-amd64-linux.so)
==471166==    by 0x5E27E29: __gnu_cxx::new_allocator<std::__detail::_Hash_node_base*>::allocate(unsigned long, void const*) (new_allocator.h:127)
==471166==    by 0x5E27CCC: std::allocator_traits<std::allocator<std::__detail::_Hash_node_base*> >::allocate(std::allocator<std::__detail::_Hash_node_base*>&, unsigned long) (alloc_traits.h:460)
==471166==    by 0x5E27AC6: std::__detail::_Hashtable_alloc<std::allocator<std::__detail::_Hash_node<std::pair<account_s const* const, std::vector<PeriodData, std::allocator<PeriodData> > >, false> > >::_M_allocate_buckets(unsigned long) (hashtable_policy.h:1924)
==471166==    by 0x5E277DC: std::_Hashtable<account_s const*, std::pair<account_s const* const, std::vector<PeriodData, std::allocator<PeriodData> > >, std::allocator<std::pair<account_s const* const, std::vector<PeriodData, std::allocator<PeriodData> > > >, std::__detail::_Select1st, std::equal_to<account_s const*>, std::hash<account_s const*>, std::__detail::_Mod_range_hashing, std::__detail::_Default_ranged_hash, std::__detail::_Prime_rehash_policy, std::__detail::_Hashtable_traits<false, false, true> >::_M_allocate_buckets(unsigned long) (hashtable.h:425)
==471166==    by 0x5E26FF0: std::_Hashtable<account_s const*, std::pair<account_s const* const, std::vector<PeriodData, std::allocator<PeriodData> > >, std::allocator<std::pair<account_s const* const, std::vector<PeriodData, std::allocator<PeriodData> > > >, std::__detail::_Select1st, std::equal_to<account_s const*>, std::hash<account_s const*>, std::__detail::_Mod_range_hashing, std::__detail::_Default_ranged_hash, std::__detail::_Prime_rehash_policy, std::__detail::_Hashtable_traits<false, false, true> >::_M_rehash_aux(unsigned long, std::integral_constant<bool, true>) (hashtable.h:2361)
==471166==    by 0x5E267FB: std::_Hashtable<account_s const*, std::pair<account_s const* const, std::vector<PeriodData, std::allocator<PeriodData> > >, std::allocator<std::pair<account_s const* const, std::vector<PeriodData, std::allocator<PeriodData> > > >, std::__detail::_Select1st, std::equal_to<account_s const*>, std::hash<account_s const*>, std::__detail::_Mod_range_hashing, std::__detail::_Default_ranged_hash, std::__detail::_Prime_rehash_policy, std::__detail::_Hashtable_traits<false, false, true> >::_M_rehash(unsigned long, unsigned long const&) (hashtable.h:2340)
==471166==    by 0x5E25C2B: std::_Hashtable<account_s const*, std::pair<account_s const* const, std::vector<PeriodData, std::allocator<PeriodData> > >, std::allocator<std::pair<account_s const* const, std::vector<PeriodData, std::allocator<PeriodData> > > >, std::__detail::_Select1st, std::equal_to<account_s const*>, std::hash<account_s const*>, std::__detail::_Mod_range_hashing, std::__detail::_Default_ranged_hash, std::__detail::_Prime_rehash_policy, std::__detail::_Hashtable_traits<false, false, true> >::_M_insert_unique_node(unsigned long, unsigned long, std::__detail::_Hash_node<std::pair<account_s const* const, std::vector<PeriodData, std::allocator<PeriodData> > >, false>*, unsigned long) (hashtable.h:2000)
==471166==    by 0x5E24EB4: std::pair<std::__detail::_Node_iterator<std::pair<account_s const* const, std::vector<PeriodData, std::allocator<PeriodData> > >, false, false>, bool> std::__detail::_Insert_base<account_s const*, std::pair<account_s const* const, std::vector<PeriodData, std::allocator<PeriodData> > >, std::allocator<std::pair<account_s const* const, std::vector<PeriodData, std::allocator<PeriodData> > > >, std::__detail::_Select1st, std::equal_to<account_s const*>, std::hash<account_s const*>, std::__detail::_Mod_range_hashing, std::__detail::_Default_ranged_hash, std::__detail::_Prime_rehash_policy, std::__detail::_Hashtable_traits<false, false, true> >::try_emplace<account_s const* const&, std::vector<PeriodData, std::allocator<PeriodData> > >(std::__detail::_Node_const_iterator<std::pair<account_s const* const, std::vector<PeriodData, std::allocator<PeriodData> > >, false, false>, account_s const* const&, std::vector<PeriodData, std::allocator<PeriodData> >&&) (hashtable_policy.h:871)
==471166==    by 0x5E24515: std::pair<std::__detail::_Node_iterator<std::pair<account_s const* const, std::vector<PeriodData, std::allocator<PeriodData> > >, false, false>, bool> std::unordered_map<account_s const*, std::vector<PeriodData, std::allocator<PeriodData> >, std::hash<account_s const*>, std::equal_to<account_s const*>, std::allocator<std::pair<account_s const* const, std::vector<PeriodData, std::allocator<PeriodData> > > > >::insert_or_assign<std::vector<PeriodData, std::allocator<PeriodData> > >(account_s const* const&, std::vector<PeriodData, std::allocator<PeriodData> >&&) (unordered_map.h:656)
==471166==    by 0x5E23822: get_perioddata(budget_s const*, account_s const*, unsigned int) (gnc-budget.cpp:781)
==471166==    by 0x5E22923: gnc_budget_is_account_period_value_set (gnc-budget.cpp:616)
```

jralls:
Is that running test-budget?

christopherlam:
> Is that running test-budget?

No it's a regular session editing one cell then quit. Is the unique ptr being destroyed when g_object_unref is called?

jralls:
It *should* be, but since the object's private data is destroyed by GType, in C code, it's entirely possible that the unique_ptr's destructor doesn't get called. Try adding
```
priv->acct_map = nullptr;
```
to `gnc_budget_free` to see if it fixes the leak.

christopherlam:
> Try adding `priv->acct_map = nullptr;` to `gnc_budget_free` to see if it fixes the leak.

Yes it does fix it.

libgnucash/engine/gnc-budget.cpp
```diff
@@ -58,6 +61,16 @@ typedef struct
...
 
+struct PeriodData
+{
+    std::string note;
+    bool value_is_set;
+    gnc_numeric value;
+};
+
+using PeriodDataVec = std::vector<PeriodData>;
+using AcctMap = std::unordered_map<const Account*,PeriodDataVec>;
```
jralls, line 72: Needs a space after the comma.

libgnucash/engine/gnc-budget.cpp
```diff
@@ -631,25 +700,36 @@ gnc_budget_set_account_period_note(GncBudget *budget, const Account *account,
...
 }
 
+
+gchar *
+gnc_budget_get_account_period_note (const GncBudget *budget,
+                                    const Account *account, guint period_num)
+{
+    g_return_val_if_fail (period_num < GET_PRIVATE(budget)->num_periods, nullptr);
+    auto& data = get_perioddata (budget, account, period_num);
+    return data.note.empty () ? nullptr : g_strdup (data.note.c_str());
```
jralls, line 730: Should this function return `nullptr` if `!data.value_is_set`? IOW does it make sense for a period to have a valid note but not a valid value?

libgnucash/engine/test/utest-Budget.c
jralls: Oh, this is a unit test. Never mind.

jralls (changes requested):

libgnucash/engine/gnc-budget.cpp
christopherlam: Dunno... I've replicated existing behaviour...

jralls: OK, I guess, since none of us actually uses budgets. :-/

christopherlam:
Last 3 may be too much.

libgnucash/engine/gnc-budget.cpp
```diff
@@ -471,31 +496,46 @@ gnc_budget_get_num_periods(const GncBudget* budget)
...
+    std::string period_str { std::to_string (period_num) };
+    return { acct_guid, period_str };
 }
 
+static inline StringVec
+make_period_note_path (const Account *account, guint period_num)
+{
+    StringVec path { GNC_BUDGET_NOTES_PATH };
+    StringVec data_path { make_period_data_path (account, period_num) };
+    std::move (data_path.begin(), data_path.end(), std::back_inserter (path));
```
jralls, line 514: std::move doesn't do anything here.

libgnucash/engine/gnc-budget.cpp
```diff
@@ -523,18 +560,22 @@ gnc_budget_set_account_period_value(GncBudget *budget, const Account *account,
...
+        perioddata.value_is_set = false;
+    }
     else
     {
-        GValue v = G_VALUE_INIT;
-        g_value_init (&v, GNC_TYPE_NUMERIC);
-        g_value_set_boxed (&v, &val);
-        qof_instance_set_kvp (QOF_INSTANCE (budget), &v, 2, path_part_one, path_part_two);
-        g_value_unset (&v);
+        KvpValue* v = new KvpValue (*(gnc_numeric*)(&val));
```
jralls, line 575: What the heck is `(*(gnc_numeric*)(&val))` supposed to do?

jralls (changes requested):
Generally pretty good, just a couple of minor cleanups.

libgnucash/engine/gnc-budget.cpp
christopherlam: It's a different move: https://en.cppreference.com/w/cpp/algorithm/move because period_data_path is `{guid, period_num}`, and period_note_path is `{guid, period_num, "notes"}`.

libgnucash/engine/gnc-budget.cpp
christopherlam: It's copied directly from https://github.com/Gnucash/gnucash/blob/831b25f8f8a218e5e9b75bcc6c0161e8ba1bd6b1/libgnucash/engine/kvp-frame.cpp#L347

libgnucash/engine/gnc-budget.cpp
```diff
@@ -471,31 +496,46 @@ gnc_budget_get_num_periods(const GncBudget* budget)
...
-    const GncGUID *guid;
-    gchar *bufend;
-    guid = xaccAccountGetGUID (account);
-    guid_to_string_buff (guid, path1);
-    g_sprintf (path2, "%d", period_num);
+    gchar guid_buf [GUID_ENCODING_LENGTH + 1];
+    guid_to_string_buff (xaccAccountGetGUID (account), guid_buf);
+    std::string acct_guid { guid_buf };
+    std::string period_str { std::to_string (period_num) };
+    return { acct_guid, period_str };
```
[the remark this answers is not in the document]
christopherlam: done... needed small adjustments.

libgnucash/engine/gnc-budget.cpp
christopherlam: this is now simplified.

libgnucash/engine/gnc-budget.cpp
jralls: Oh, cool. Nice find.

libgnucash/engine/gnc-budget.cpp
jralls: > It's copied directly from 

OK, but there's a big difference between the `void*` return value of `g_value_get_boxed` and the `gnc_numeric` passed in by value in `val`.

jralls (approved):

jralls:
Looks good, go ahead and merge when ready.

libgnucash/engine/gnc-budget.cpp
```diff
@@ -674,6 +676,46 @@ gnc_budget_get_account_period_actual_value(
...
+            auto kval1 { budget_kvp->get_slot (make_period_data_path (account, i)) };
+            auto kval2 { budget_kvp->get_slot (make_period_note_path (account, i)) };
+
+            auto is_set = kval1 && kval1->get_type() == KvpValue::Type::NUMERIC;
+            auto num = is_set ? kval1->get<gnc_numeric>() : gnc_numeric_zero ();
+
+            if (kval2 && kval2->get_type() == KvpValue::Type::STRING)
+                note = kval2->get<const char*>();
+
+            PeriodData data { note, is_set, num };
```
christopherlam, line 709: I guess this should be `PeriodData data { std::move (note), is_set, num };`

PR #1252, opened 2022-01-15
L10N:nl: Better Invoice Number translation

frenkel:
In Dutch it is weird to say "Sales Invoice Number", instead you use
"Invoice Number". Also, the old value messed up the PDF layout because
it was so long.

5345 translated messages, 29 fuzzy translations, 6 untranslated messages.

### DIFF
--- a/po/nl.po
+++ b/po/nl.po
@@ -23694,7 +23694,7 @@ msgstr "Opmaaksjabloon"
 
 #: gnucash/report/report-core.scm:165
 msgid "Invoice Number"
-msgstr "Verkoopfactuurnummer"
+msgstr "Factuurnummer"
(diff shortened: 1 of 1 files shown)


fellen:
@gjanssens From my POV the text is OK, but are there side effects?

gjanssens (approved):
I have no issues with this change. As far as I can see it has no side effects elsewhere in the code.

fellen:
@frenkel Updating translations is easier done by https://hosted.weblate.org/projects/gnucash/

frenkel:
@fellen Thanks for the hint, but I like to test my translations in the actual application to see how it lines up etc.

PR #1255, opened 2022-01-18
L10n:nl: Better tax name translations for invoices

frenkel:
Most Dutch invoices do not talk about taxes and tax rates, but about VAT and VAT%. This change has the added benefit that the invoice PDF layout is much better:

Tax invoice pdf before:
Screenshot from 2022-01-18 14-29-11

Tax invoice pdf after:
Screenshot from 2022-01-18 14-28-41

### DIFF
--- a/po/nl.po
+++ b/po/nl.po
@@ -27114,7 +27114,7 @@ msgstr "Winst- en verliesrekening"
 #: gnucash/report/reports/standard/taxinvoice.scm:108
 #: gnucash/report/reports/standard/taxinvoice.scm:181
 msgid "Tax Amount"
-msgstr "Belastingbedrag"
+msgstr "BTW"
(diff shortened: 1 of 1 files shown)


jralls:
We handle all translations through our [weblate site](https://hosted.weblate.org/projects/gnucash/gnucash/nl/). Please offer your changes as suggestions there.

fellen:
@gjanssens Just to inform you as it's your language.

frenkel:
@jralls I think this case would better be handled as a PR. That weblate system still has my suggestions unhandled and it misses the ability to show something like a screenshot as the one above.

fellen:
@frenkel How about
grafik

frenkel:
Even better 😄

PR #1259, opened 2022-01-25
Bug741674 - Template accounts appearing in search register and gl

Bob-IT:
This is a proposed fix for the bug mentioned. The first commit I think is OK but am wondering if the second commit needs to be changed.

The second commit uses a hash table to remember the template accounts that have been excluded already. I am comparing the template account pointer to do this and am wondering if it should be based on the guid string. The template account pointer should be around while program is running but when the schedule is deleted, the template accounts will also be deleted.
As I am just comparing the pointer, will this matter?

### DIFF
--- a/gnucash/gnome/dialog-find-transactions.c
+++ b/gnucash/gnome/dialog-find-transactions.c
@@ -193,35 +193,7 @@ gnc_ui_find_transactions_dialog_create(GtkWindow *parent, GNCLedgerDisplay * ori
     {
         start_q = qof_query_create ();
         qof_query_set_book (start_q, gnc_get_current_book ());
-
-        /* In lieu of not "mis-using" some portion of the infrastructure by writing
-         * a bunch of new code, we just filter out the accounts of the template
-         * transactions.  While these are in a separate Account trees just for this
-         * reason, the query engine makes no distinction between Account trees.
-         * See Gnome Bug 86302.
-         * 	-- jsled
-         *
-         * copied from gnc-ledger-display.c:gnc_ledger_display_gl()  -- warlord
-         *
-         * <jsled> Alternatively, you could look for a GNC_SX_ACCOUNT [SchedAction.h]
-         * key in the KVP frame of the split.
-         */
-        {
-            Account *tRoot;
-            GList *al;
-
-            tRoot = gnc_book_get_template_root( gnc_get_current_book() );
-            al = gnc_account_get_descendants( tRoot );
-
-            if (g_list_length(al) != 0)
-                xaccQueryAddAccountMatch( start_q, al, QOF_GUID_MATCH_NONE, QOF_QUERY_AND );
-
-            g_list_free (al);
-            al = NULL;
-            tRoot = NULL;
-        }
-
-        ftd->q = start_q;		/* save this to destroy it later */
+        ftd->q = start_q; // save this to destroy it later
     }
 
     ftd->parent = parent;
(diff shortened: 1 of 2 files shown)


jralls:
You're checking ptrs and not dereferencing them so there's no harm if some are dangling.

Don't use `g_list_remove_link` to remove the current node in a list traversal: That re-traverses the list and also leaks the removed node. Operate directly on node and its prev and next nodes like so:
```
GList *node_to_free = node;
node->prev->next = node->next;
node->next->prev = node->prev;
node = node->prev; // So that the loop's call to g_list_next (node) will work
g_free (node_to_free);
```

Bob-IT:
@jralls 
I had a look at the code you gave and got a segfault. Added some if statements to get it to work with one element and got close but I still got a segfault. I downloaded the source of glib to understand why and looked at the function I used g_list_remove_link and that function does what you proposed and I can not see any traversing.

Yes I did not free the element and to do that I should of added a g_list_free_1 or just used the g_list_delete_link instead.

On an unrelated issue, while trying to debug this, I used gdb and got the following unrelated segfault, have you come across this?

(gdb) bt full
#0  0x00007ffff5d0adf6 in  () at //usr/lib64/libgc.so.1
#1  0x00007ffff5d15f68 in GC_init () at //usr/lib64/libgc.so.1
#2  0x00007ffff7da77e3 in scm_storage_prehistory () at /usr/lib64/libguile-2.2.so.1
#3  0x00007ffff7db17f5 in scm_i_init_guile () at /usr/lib64/libguile-2.2.so.1
#4  0x00007ffff7e0c0dd in scm_i_init_thread_for_guile.part () at /usr/lib64/libguile-2.2.so.1
#5  0x00007ffff7e0c236 in with_guile () at /usr/lib64/libguile-2.2.so.1
#6  0x00007ffff5d0a77e in GC_call_with_stack_base () at //usr/lib64/libgc.so.1
#7  0x00007ffff7e0c388 in scm_with_guile () at /usr/lib64/libguile-2.2.so.1
#8  0x00007ffff7db1780 in scm_boot_guile () at /usr/lib64/libguile-2.2.so.1
#9  0x00005555555650a2 in Gnucash::Gnucash::start(int, char**) (this=0x7fffffffdef0, argc=1, argv=0x7fffffffe2a8)
    at /var/tmp/portage/app-office/gnucash-4.9/work/gnucash-4.9/gnucash/gnucash.cpp:330
        user_file_spec = {nofile = 0, file_to_load = 0x55555557460b ""}
#10 0x000055555556396d in main(int, char**) (argc=<optimized out>, argv=<optimized out>)
    at /var/tmp/portage/app-office/gnucash-4.9/work/gnucash-4.9/gnucash/gnucash.cpp:354
        application =

jralls:
@Bob-IT, You're right, `g_list_remove_link` does work efficiently, and yes, `g_link_delete_list` would be the easiest way to stop the leak.

As for the segfault in bdwgc's GC_init, I do recall a bug some time in the last couple of years that featured a bdwgc crash but I can't find it. I *think* it had something to do with threading on Windows, though, and so it wouldn't be this one. This seems to be right at Guile startup, is that right? Are you able to reproduce it at all?

Bob-IT:
OK, I will change to use the `g_list_delete_link` function and push this tomorrow.

As for the segfault, yes it is at startup with gdb and every time, generally do...
gdb gnucash
run --g-fatal-warnings
and then stops with a segfault, but pressing 'c' will continue and gnucash will load successfully.

Fired up a test Xubunto VM, and all works as expected so looks like it's down to my Gentoo development VM's. Will have a look tomorrow and see if I need to do an update.

jralls:
Check the bdwgc version in the emerge script and have a look through [bdwgc issues](https://github.com/ivmai/bdwgc/issues). Maybe you found a new bug!

PR #1261, opened 2022-01-28
Unify labels and tooltips of dates in transaction report

fellen:
As said some time before on IRC I am not happy with the inconsistency of labels and tooltips in trep-engine. So here is my suggestion. But currently it causes 2 tests to fail.

Still open for improvements…

### DIFF
--- a/gnucash/report/trep-engine.scm
+++ b/gnucash/report/trep-engine.scm
@@ -922,9 +922,9 @@ be excluded from periodic reporting.")
      ;; One list per option here with: option-name, sort-tag,
      ;; help-string, default-value
      (list
-      (list (N_ "Date")                         "a"  (G_ "Display the date?") #t)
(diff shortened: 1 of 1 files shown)


christopherlam:
The tests fail because the options being tested are hard coded and need to be changed as well: https://github.com/Gnucash/gnucash/blob/5bcc3d6b14c561cb24b210115c9abef401bb3230/gnucash/report/reports/standard/test/test-transaction.scm#L465 -- There must also be backward-compatibility lines added in new-names-alist 
https://github.com/Gnucash/gnucash/blob/5bcc3d6b14c561cb24b210115c9abef401bb3230/libgnucash/app-utils/options.scm#L1805

From an Anglophone, "Posting Date", "Reconciliation Date", "Entry Date" would sound better.

jralls (commented):
I agree with @christopherlam about the labels, but I don't see it as a big deal, they're all understandable.

PR #1267, opened 2022-02-04
Bug 798438 - on reconcile window, 'R' column sorting broken 

Bob-IT:
This is a strange one, I by mistake added sorting of the 'R' column in version 3.0. As far as I can tell this never did the correct thing but the reporter was adamant it did. From what I can tell, the other columns are sorted by changing the query but the 'R' column is populated from a search param function and as such the model would need to do the sorting.

I have added this in the fourth commit along with a couple of commits that I will probably squash.

The subsequent commit adds the ability to 'scroll to selection' to gnc-query-view. This can be used for keeping a selected row in view when the sort order is changed which I thought I was going to set but had second thoughts. Currently only used from the reconcile-view to keep any selected row in view if a change is made to the register.

The last commit was an error I noticed when run from a terminal.

I think this is all works OK but thought I would park here for week for any comments.


jralls:
> I by mistake added sorting of the 'R' column in version 3.0.

This is a lot of work to reintroduce a supposed feature created by mistake, but otherwise it looks OK.

PR #1272, opened 2022-02-09
Category charts upgrade to chart allocation ratios instead of absolute values.

exxus:
I created a patch to show ratios in the barchart.
This is useful if you want to maintain a certain ratio between different asset classes, e.g. 60:40.
I contribute this, since I think it can help other users as well.
Before this can be merged, maybe some documentation needs to be added.
Thank you all for this really helpful piece of software.

Best regards
exxus

### DIFF
--- a/gnucash/report/reports/standard/category-barchart.scm
+++ b/gnucash/report/reports/standard/category-barchart.scm
@@ -328,23 +328,21 @@ developing over time"))
              ;; created.
              (other-anchor ""))
 
-        ;; Converts a commodity-collector into gnc-monetary in the report's
-        ;; currency using the exchange-fn calculated above. Returns a gnc-monetary
+        ;; Converts a commodity-collector into amount in the report's
+        ;; currency using the exchange-fn calculated above. Returns an amount
         ;; multiplied by the averaging-multiplier (smaller than one; multiplication
         ;; instead of division to avoid division-by-zero issues) in case
         ;; the user wants to see the amounts averaged over some value.
-        (define (collector->monetary c date)
-          (gnc:make-gnc-monetary
-           report-currency
-           (* averaging-multiplier
-              (gnc:gnc-monetary-amount
-               (gnc:sum-collector-commodity
-                c report-currency
-                (lambda (a b) (exchange-fn a b date)))))))
+        (define (collector->report-currency-amount c date)
+          (* averaging-multiplier
+             (gnc:gnc-monetary-amount
+              (gnc:sum-collector-commodity
+               c report-currency
+               (lambda (a b) (exchange-fn a b date))))))
 
         (define (all-zeros data)
           (cond
-           ((gnc:gnc-monetary? data) (zero? (gnc:gnc-monetary-amount data)))
+           ((number? data) (zero? data))
            ((pair? data) (every all-zeros data))
            (else (error 'huh))))
 
@@ -396,11 +394,11 @@ developing over time"))
                   (loop (cdr list-of-mon-collectors)
                         (cdr dates-list)
                         (cons (if do-intervals?
-                                  (collector->monetary
+                                  (collector->report-currency-amount
                                    (gnc:collector- (cadr list-of-mon-collectors)
                                                    (car list-of-mon-collectors))
                                    (cadr dates-list))
-                                  (collector->monetary
+                                  (collector->report-currency-amount
                                    (car list-of-mon-collectors)
                                    (car dates-list)))
                               result))))))
@@ -462,9 +460,7 @@ developing over time"))
         ;; Sort the account list according to the account code field.
         (set! all-data
           (sort
-           (filter (lambda (l)
-                     (not (zero? (gnc:gnc-monetary-amount
-                                  (apply gnc:monetary+ (cadr l))))))
+           (filter (lambda (l) (not (zero? (apply + (cadr l)))))
                    (traverse-accounts 1 topl-accounts))
            (case sort-method
              ((alphabetical)
@@ -480,8 +476,8 @@ developing over time"))
                                      (xaccAccountGetCode (car b)))))
              ((amount)
               (lambda (a b)
-                (> (gnc:gnc-monetary-amount (apply gnc:monetary+ (cadr a)))
-                   (gnc:gnc-monetary-amount (apply gnc:monetary+ (cadr b)))))))))
+                (> (apply + (cadr a))
+                   (apply + (cadr b))))))))
 
         (cond
          ((or (null? all-data) (all-zeros (map cadr all-data)))
@@ -494,7 +490,8 @@ developing over time"))
           (let* ((dates-list (if do-intervals?
                                  (list-head dates-list (1- (length dates-list)))
                                  dates-list))
-                 (date-string-list (map qof-print-date dates-list)))
+                 (date-string-list (map qof-print-date dates-list))
+                 (list-of-rows (apply zip (map cadr all-data))))
 
             ;; Set chart title, subtitle etc.
             (gnc:html-chart-set-type!
@@ -523,7 +520,7 @@ developing over time"))
                 (let* ((start (take all-data (1- max-slices)))
                        (finish (drop all-data (1- max-slices)))
                        (other-sum (map
-                                   (lambda (l) (apply gnc:monetary+ l))
+                                   (lambda (l) (apply + l))
                                    (apply zip (map cadr finish)))))
                   (set! all-data
                     (append start
@@ -551,7 +548,7 @@ developing over time"))
                               ((string? acct) (car series))
                               (show-fullname? (gnc-account-get-full-name acct))
                               (else (xaccAccountGetName acct))))
-                      (amounts (map gnc:gnc-monetary-amount (cadr series)))
+                      (amounts (cadr series))
                       (stack (if stacked? "default" (number->string stack)))
                       (fill (eq? chart-type 'barchart))
                       (urls (cond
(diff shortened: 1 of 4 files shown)


gnucash/report/reports/standard/category-barchart.scm
```diff
@@ -600,17 +621,23 @@ developing over time"))
...
-                    (append (list (make-cell date))
-                            (map make-cell row)
-                            (if cols>1?
-                                (list
-                                 (make-cell (apply gnc:monetary+ row)))
-                                '()))))
+                    ( let ((grandt (apply l1norm (map gnc:gnc-monetary-amount row))))
+                     (append (list (make-cell date))
+                             (if ratio-chart? 
+                                 (map make-cell-percent (map (lambda (x) (* (divide-number  (gnc:gnc-monetary-amount x) grandt) 100)) row))
```
christopherlam, line 635: To be more efficient scheme, don't nest `map` calls; nesting creates intermediate lists which are then immediately consumed by the outer `map` therefore increasing gc workload.

```scheme
(map (lambda (x) (make-cell-percent (* (/ (gnc:gnc-monetary-amount x) grandt) 100) row))
```

gnucash/report/reports/standard/category-barchart.scm
```diff
@@ -175,14 +182,21 @@ developing over time"))
       (N_ "Show table")
       "e" (N_ "Display a table of the selected data.")
       #f))
+      
+    (add-option
+     (gnc:make-simple-boolean-option
+      gnc:pagename-display
+      (N_ "Show ratios")
+      "f" (N_ "Display ratios instead of amounts.")
+      #f))
```
christopherlam, line 191: Instead of using `"f"` and modifying other options's sort strings, it's simpler to use `"e1"` here. Also `Show ratios` is incomplete -- it will no longer draw absolute values in a chart, but will instead report the percentage contribution of each account into the overall asset/liability/income/expense. Is there a better formal name to be used?

christopherlam (changes requested):
Some comments. This is an interesting chart modification, and needs further testing. I've also seen div/0 errors but cannot debug this at this time.

gnucash/report/reports/standard/category-barchart.scm
```diff
@@ -34,6 +34,13 @@
 (use-modules (gnucash app-utils))
 (use-modules (gnucash report))
 
+;; useful functions
+(define divide-number (lambda (x y) (gnc-numeric-div  x  y GNC-DENOM-AUTO (logior (GNC-DENOM-SIGFIGS 6) GNC-RND-ROUND) )))
```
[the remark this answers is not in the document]
exxus: If I use `/` directly, it will lead to overflow  errors due to division by 0.
I have some time spans where all assets are 0. In this case, ratios make no sense.
But using the gnc-numeric-divide function, the report does not crash but shows 0% for the ratio of all assets.

christopherlam:
Finally this modification would benefit from tests in `test-standard-category-report.scm`. Would you be able to watch the userlist and buzgilla for bug reports and respond? Alternatively a description for the release notes, which include hints how to handle edge cases will be needed. Merging this change will undoubtly generate queries and the core developers do not have all answers.

christopherlam:
> > Finally this modification would benefit from tests in `test-standard-category-report.scm`. Would you be able to watch the userlist and buzgilla for bug reports and respond? Alternatively a description for the release notes, which include hints how to handle edge cases will be needed. Merging this change will undoubtly generate queries and the core developers do not have all answers.

For copyright and attribution we may wish to add your github handle onto the report.

> I am not able to read all the mailing list, but I will try to reply and maintain the option when I get notifactions e.g. on github. I agree that is testcase is useful, I looked at the mentioned file, but couldn't figure out how to add a test. Is there testing database with which I can test edge cases?

`test-standard-category-report.scm` calls `env-create-daily-transactions` in test-engine-extras to create the book, dumps the html somewhere in /tmp/ and tests the resulting table contents using sxml. You may use the same test data, or (better) create custom ones.

> I guess the most significant edge cases are negative values, because there is no clear understanding to me how to divide e.g. a cake with negative pieces. I can write an explanation with the intended use case and what happens in case of negative values.

Yes it would be useful to write an explanation for the release notes. I think the most severe edge case will be how to handle mixed-account-type hierarchies such as Income[INCOME]:Salary[INCOME]:PAYETaxes[EXPENSE] for example, whereby both Salary and PAYETaxes are populated with splits (and the account balances may become negative!). But the existing reports are likely to fail too.

> I mainly use it to maintain a certain ratio between some of my asset classes (e.g. investment funds and fixed-interest securities). In the hope other will benefit from it as well, I wanted to share it. But if it is too complicated to maintain, understand or support or if it is not in the scope of the reports provided with gnucash, don't hesitate to close the pull request.

I think it is an interesting chart, and as bonus it is a single-option upgrade to an existing report.

gnucash/report/reports/standard/category-barchart.scm
```diff
@@ -34,6 +34,12 @@
 (use-modules (gnucash app-utils))
 (use-modules (gnucash report))
 
+;; useful functions
+(define (safe-/ x y)
+  (if (and (zero? x) (zero? y)) 0 (/ x y)))
```
christopherlam, line 39: Are you sure both x and y being 0 is the only div/0? What if the assets have only two accounts: cash $400 and loan -$400?

christopherlam:
While reviewing this PR I think the `collector->monetary` part to create a gnc-monetary and should be changed to `collector->amount`, an amount of the _known_ report currency. This will simplify this report a lot. I think this change can be done after merging this PR.

gnucash/report/reports/standard/category-barchart.scm
exxus: This is the part where the abs value will make y 800. In this case it makes no difference. Since if x=0, y will also always be 0. We can also stick with `(if (zero? y) 0 (/ x y)))`. I just added the check for x, because for x not 0, a divide by 0 error in this case would show that something is wrong.

gjanssens: > Since if x=0, y will also always be 0.

There is nothing in this function definition that guarantees this so your statement must be based on how this function is used. That's slightly risky because that opens doors to mistakes in future changes of the code.

>  We can also stick with `(if (zero? y) 0 (/ x y)))`. I just added the check for x, because for x not 0, a divide by 0 error in this case would show that something is wrong.

The added check adds nothing. The function will return 0 in both implementations so there's no way your calling code can tell the difference.

gnucash/report/reports/standard/category-barchart.scm
```diff
@@ -600,17 +620,23 @@ developing over time"))
...
+                     (append (list (make-cell date))
+                             (if ratio-chart? 
+                                 (map (lambda (x) (make-cell-percent (* (safe-/ (gnc:gnc-monetary-amount x) grandt) 100))) row)
+                                 (map make-cell row))
+                             (if cols>1?
+                                 (list
+                                  (make-cell (apply gnc:monetary+ row)))
+                                 '())))))
                  date-string-list
                  (apply zip (map cadr all-data)))
```
christopherlam, line 641: Instead of recreating grandt for each row, it'll be much more clever to reuse the `grandts` already generated. For that matter, you'll want to modify the `for-each` as follows:

```scheme
(for-each
 (lambda (date row grandt)
  (gnc:html-table-append-row! ...etc...and use grandt now available for use))
 date-string-list
 (apply zip (map cadr all-data))
 grandts)
```

gnucash/report/reports/standard/category-barchart.scm
```diff
@@ -552,6 +569,7 @@ developing over time"))
                               (show-fullname? (gnc-account-get-full-name acct))
                               (else (xaccAccountGetName acct))))
                       (amounts (map gnc:gnc-monetary-amount (cadr series)))
+                      (percentage (map (lambda (x grandt)  (safe-/ x grandt)) amounts grandts))
```
christopherlam, line 572: This line is ok for readability, but because `percentage` is only used iff `ratios-chart?` is `#t`, it would be marginally more efficient to generate the calls to `map` and `safe-/` only when required; thus:

```guile
  (gnc:html-chart-add-data-series!
   chart label (if ratio-chart? (map safe-/ amounts grandts) amounts) 
   color 'stack stack 'fill fill 'urls urls)
```

christopherlam (approved):
Final comments: I agree it is clever to show negative assets as a negative bar, it may not be strictly legitimate according to formal reporting. Many users are *extremely* sharp as far as accounting standards are concerned, and will ask questions. So, some documentation, including caveats on edge cases, and description how it handles negative amounts, will be crucial for users to understand its behaviour.

gnucash/report/reports/standard/category-barchart.scm
exxus: > There is nothing in this function definition that guarantees this so your statement must be based on how this function is used. That's slightly risky because that opens doors to mistakes in future changes of the code.

That is true, only in my use case if y=0, also x will be 0.
> Since if x=0, y will also always be 0.

Sorry, that was is mistake. In my usecase: y=abs(x_1)+...+abs(x_N). So if y=0, all x must have been 0.
> The added check adds nothing. The function will return 0 in both implementations so there's no way your calling code can tell the difference.

Consider the case y=0, x=1.
`(if (zero? y) 0 (/ x y)))` will return 0
`(if (and (zero? x) (zero? y)) 0 (/ x y)))` will fail due to division by 0: (/ 1 0).
In my implementation the case y=0, x=1 can't happen. So if it still occurs, something is wrong and the report will fail.

christopherlam:
With 5f06fc9 in maint, this branch is no longer current. Please direct all future discussions in #1274.

PR #1274, opened 2022-02-12
Category charts upgrade to chart allocation ratios instead of absolute values.

christopherlam:
Includes simplication of category-barchart.scm and code from @exxus, and test code.

It would be useful if @exxus would include a description for release notes, especially handling edge cases.

The last issue relates the table output format. It contains both percentages and currency. Wouldn't this be confusing? If this is intentional, it should also be included in the release notes.

image

### DIFF
--- a/gnucash/report/html-chart.scm
+++ b/gnucash/report/html-chart.scm
@@ -51,6 +51,8 @@
 (export gnc:html-chart-set-currency-iso!)
 (export gnc:html-chart-currency-symbol)
 (export gnc:html-chart-set-currency-symbol!)
+(export gnc:html-chart-format-style)
+(export gnc:html-chart-set-format-style!)
 (export gnc:html-chart-render)
 (export gnc:html-chart-set-custom-x-axis-ticks?!)
 (export gnc:html-chart-set-title!)
@@ -144,13 +146,14 @@
 
 (define-record-type <html-chart>
   (make-html-chart width height chart-options currency-iso
-                   currency-symbol custom-x-axis-ticks? custom-y-axis-ticks?)
+                   currency-symbol format-style custom-x-axis-ticks? custom-y-axis-ticks?)
   html-chart?
   (width html-chart-width html-chart-set-width)
   (height html-chart-height html-chart-set-height)
   (chart-options html-chart-chart-options html-chart-set-chart-options)
   (currency-iso html-chart-currency-iso html-chart-set-currency-iso)
   (currency-symbol html-chart-currency-symbol html-chart-set-currency-symbol)
+  (format-style html-chart-format-style html-chart-set-format-style)
   (custom-x-axis-ticks? html-chart-custom-x-axis-ticks?
                         html-chart-set-custom-x-axis-ticks?)
   (custom-y-axis-ticks? html-chart-custom-y-axis-ticks?
@@ -166,6 +169,8 @@
 (define gnc:html-chart-set-currency-iso! html-chart-set-currency-iso)
 (define gnc:html-chart-currency-symbol html-chart-currency-symbol)
 (define gnc:html-chart-set-currency-symbol! html-chart-set-currency-symbol)
+(define gnc:html-chart-format-style html-chart-format-style)
+(define gnc:html-chart-set-format-style! html-chart-set-format-style)
 (define gnc:html-chart-custom-x-axis-ticks? html-chart-custom-x-axis-ticks?)
 (define gnc:html-chart-set-custom-x-axis-ticks?! html-chart-set-custom-x-axis-ticks?)
 (define gnc:html-chart-custom-y-axis-ticks? html-chart-custom-y-axis-ticks?)
@@ -254,6 +259,7 @@
                                   (cons 'text ""))))))
    "XXX"     ;currency-iso
    "\u00A4"  ;currency-symbol
+   "currency";format-style
    #t        ;custom x-axis ticks?
    #t        ;custom y-axis ticks?
    ))
@@ -341,10 +347,10 @@
 // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toLocaleString
 var toLocaleStringSupportsOptions = (typeof Intl == 'object' && Intl && typeof Intl.NumberFormat == 'function');
 
-// format a number e.g. 2.5 into monetary e.g. \"$2.50\"
+// format a number e.g. 2.5 into monetary e.g. \"$2.50\" or other style formsty
 function numformat(amount) {
   if (toLocaleStringSupportsOptions) {
-      return amount.toLocaleString(undefined, {style:'currency', currency:curriso});
+      return amount.toLocaleString(undefined, {style:formsty, currency:curriso});
   } else {
       return currsym + amount.toLocaleString();
   }
@@ -457,6 +463,7 @@ document.getElementById(chartid).onclick = function(evt) {
     (push (format #f "<script id='script-~a'>\n" id))
     (push (format #f "var curriso = ~s;\n" (gnc:html-chart-currency-iso chart)))
     (push (format #f "var currsym = ~s;\n" (gnc:html-chart-currency-symbol chart)))
+    (push (format #f "var formsty = ~s;\n" (gnc:html-chart-format-style chart)))
     (push (format #f "var chartid = ~s;\n" id))
     (push (format #f "var jumpid = 'jump-~a';\n" id))
     (push (format #f "var loadstring = ~s;\n" (G_ "Load")))
--- a/gnucash/report/reports/standard/category-barchart.scm
+++ b/gnucash/report/reports/standard/category-barchart.scm
@@ -340,6 +340,9 @@ developing over time"))
                c report-currency
                (lambda (a b) (exchange-fn a b date))))))
 
+        (define acct->name
+          (if show-fullname? gnc-account-get-full-name xaccAccountGetName))
+
         (define (all-zeros data)
           (cond
            ((number? data) (zero? data))
@@ -465,11 +468,8 @@ developing over time"))
            (case sort-method
              ((alphabetical)
               (lambda (a b)
-                (if show-fullname?
-                    (gnc:string-locale<? (gnc-account-get-full-name (car a))
-                                         (gnc-account-get-full-name (car b)))
-                    (gnc:string-locale<? (xaccAccountGetName (car a))
-                                         (xaccAccountGetName (car b))))))
+                (gnc:string-locale<? (acct->name (car a))
+                                     (acct->name (car b)))))
              ((acct-code)
               (lambda (a b)
                 (gnc:string-locale<? (xaccAccountGetCode (car a))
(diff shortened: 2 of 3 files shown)


exxus:
Thank you for the further work on this.

Now we have two pull requests, #1272 and #1274. This is closed but more up to date, so I will comment on this one.
I have been away the last days, but I read more on the difficulties dealing with negative values.

It is no easy task and here are two articles that explain the problem:

https://polymathprogrammer.com/2008/11/28/percentage-calculation-with-negative-numbers/
https://polymathprogrammer.com/2010/10/06/negative-numbers-in-business-reports/

So I think an explanation is very important.

Description for release notes:

An option ("Replace amounts with percentage ratios.") was added to category-barchart reports.
This option will replace the total amount of each account with the ratio with which it contributed to the Grand Total.
The intended use case is to maintain a certain ratio between the account classes (e.g. investment funds and fixed-interest securities with a ratio of 60% : 40%).

The following screenshots explain this:
assets2
This picture shows the amount for the "fixed-interest securities" and the "investment funds" account. This report was already available.

assets2-ratios
This picture show the bar chart of the assets with the assigned ratios. One can see that at the beginning, most of the money was in "fixed-interest securities". Due  the changes of the amounts, at the end, most of the money is in "investment funds" with a ratio of 60% "investment funds" and 40% "fixed-interest securities.

The caveats are negative values, as there is no clear understanding what the share of the negative value in the Grand Total is.
Consider two accounts. Account one with $50 and account two with $-50. The Grand Total would be 0$.

In this implementation, the percentage ratio of account one would be 50$/(abs(50$)+abs(-50$)*100%=50%.
The percentage ratio of account two would be $-50/(abs($50)+abs($-50)*100%=-50%.
The sign is to show that one account contributed positively and the other account contributed negatively.

In general, for N accounts, the percentage ratio for each account i with amount a_i is calculated as:
a_i / ( abs(a_1) + abs(a_2) + ... + abs(a_N) ) * 100%.
The Grand Total is calculated as (a_1 + a_2 + ... a_N).

Example:
The "Savings Account" has an amount of $100. The "Checking Account" is overdrawn with $-50.
The Grand Total would be $50.

Assets1
The amounts are shown with bars. Positive amounts in the upward direction and negative amounts in the downward direction.

Assets_ratios
The percentage ratios are also are shown with bars. Positive ratios in the upward direction and negative ratios in the downward direction. But the interpretation is more difficult.The  "Savings Account" shows a ratio of 66.7% and the "Checking Account" a ratio of -33.3%. But this is not the ratio each account has on the Grand Total (that would be 200% for the "Savings Account" and -100% for the "Checking Account").

The implementation is chosen, so that the length of the bars adds up to 100% (abs(66.7%) + abs(-33.3%))=100%.

More interested users can read the following articles:
https://polymathprogrammer.com/2008/11/28/percentage-calculation-with-negative-numbers/
https://polymathprogrammer.com/2010/10/06/negative-numbers-in-business-reports/

I admit that showing the Grand Total in the table can be confusing.
In the case of only positive numbers it is meaningful, since on can directly see that e.g. 50% of the Grand Total 
of$ 5000 is in savings. Otherwise not so much.

Still the interpretation in case of negative numbers is not perfectly meaningful.
1. If the users can live with the above explanation, it is fine.
If not, one could
2. use another calculation, e.g a_i/(abs(Grand Total)). In case of positive values this would lead to the same result. In case of negative values, the results and the interpretation are different and very large percentage ratios >> 100% are possible. If Grand Total is 0, this will not work.
3. Only allow the option when all amounts are positive and show an error message otherwise. This is the intended use case.
4. Close the pull request without merging.

The more I think about this, I would go with option 3. or 4.
What are your thoughts?

christopherlam:
I think option 1 is absolutely fine, and a nice (and unconventional) workaround to handling negative amounts, as long as caveats are explicitly described so that users do not think unusual outputs are buggy.

As an example, my #759 is a nice cashflow-type report (currently stalled for various reasons) works well as long as the amounts are positive. However notice the odd line connecting Income to Rebates; this is due to negative amounts somewhere. This is not a blocker to merging this chart. The report description _will_ describe it as a known unknown behaviour.

Any one else objects?

jralls:
>Any one else objects?

Release notes aren't good enough, the explanation needs to be added to the manual. It's complicated enough that the PR should probably go to master instead of maint.

The wording is a bit clumsy as well. The phrase "percentage ratio" is redundant: Percentages are by definition ratios: Per cent is Latin for "divided by 100", i.e. a ratio expressed in hundredths. I suggest "Percentage chart", "Display account contributions as a percentage of the total assets for the period."

Accounting doesn't really do negative values: That negative checking account balance is formally a liability, and we use separate columns for amounts that increase or decrease the account's balance. You've borrowed money from the bank, so the savings account represents 100% of assets and, absent other liabilities, the checking account is 100% of liabilities. While it's true that the savings account is 200% and the checking account is -100% of net worth expressing it that way isn't terribly useful: It's more common when evaluating a business to report liabilities as a percentage of assets, but we don't have charts or reports for those common business accounting ratios. I propose an option 3.5: Negative balances are excluded from the Asset and Liability bar charts and the Net Worth bar chart doesn't include the percentage option. You could extend the logic to look for negative balances on the opposite accounts, e.g. a debit balance on a liability account is an asset. That would require changing the default account selections when the option is selected.

exxus:
Thank you for your extensive comment.

> It's complicated enough that the PR should probably go to master instead of maint.

I think the pull request is for maint.

> The wording is a bit clumsy as well. The phrase "percentage ratio" is redundant: Percentages are by definition ratios: Per cent is Latin for "divided by 100", i.e. a ratio expressed in hundredths. I suggest "Percentage chart", "Display account contributions as a percentage of the total assets for the period."

I agree the wording can be improved. What I wanted to express is that the ratios are shown in parts per hundred, that is e.g. not 0.7 but 70%. I am totally fine with your suggested wording.

> Accounting doesn't really do negative values ... 
 I propose an option 3.5: Negative balances are excluded from the Asset and Liability bar charts and the Net Worth bar chart doesn't include the percentage option.

I think your proposition is reasonable.
Ideally, negative values should not occur, and users should instead add it to liabilities. But unfortunately, it is common practice to have negative values in the assets accounts. So if I understand you correctly your option ignores negative balances for the asset bar chart, because a negative asset is not really an asset but a liability. So we ignore them also in the denominator:
 (+ (if (> a_1 0)  a_1 0) .... (if (> a_N 0)  a_N 0)  ) .

Should we also start ignoring negative values in the current behavior when the account values are shown?

What about sub-accounts? If you have an account Bank_A in assets with a positive value of $50, that consists of two sub-accounts Savings and Checking with $100 and $-50.
Should we already ignore the $-50 at the top level? So if we compare the assets of Bank_A and Bank_B ($300) do we show $50 for Bank_A or $100? Is the percentage  of Bank_A  in the total assets $50/($350)=14.29% or $100/($400)=25%?

>You could extend the logic to look for negative balances on the opposite accounts, e.g. a debit balance on a liability account is an asset. That would require changing the default account selections when the option is selected.

After reading your very helpful comment I am leaning more towards option 3. If negative values occur, we should show a warning that negative assets are liabilities and negative liabilities are assets and leave it in the responsibility of the user to get rid of the negative values by putting them as positive values in the opposite accounts.
So accounts (e.g. Checking, Credit Card) that can be overdrawn but also can contain assets should be added twice, to the Asset and to the Liability accounts.  Instead of putting negative values to the Asset account the user should add it to the corresponding Liability account. I don't know how well the practice is described in the documentation / manual.

> It's more common when evaluating a business to report liabilities as a percentage of assets, but we don't have charts or reports for those common business accounting ratios.

Something to work on ;)

Thank you again for your valuable inputs. I am looking forward to further comments.

jralls:
This PR is for maint, but nothing stops us from merging it into master instead.

I'll turn your subaccounts question around on you: What if instead of an overdrawn checking account at Bank A you have a credit card there with a credit balance? Would you deduct that from your Assets and reduce Bank A's percentage of total assets? After all, the only reason you have credit balance in your checking account is because you didn't take the trouble to set up a Liabilities:Current:BankA:Overdrafts to hold the liability balance. Considering the complexities, how much effort is appropriate to support an improperly set up book?

We're also drifting away from your original use-case, which was about asset allocation in an investment portfolio. In *that* case things get a bit more interesting, because investment management and accounting diverge in how one regards things like call options and short sales. If one accepts that those are speculative instruments that aren't really suitable for a portfolio that's seeking a certain asset allocation. What do you really want to model, and is it really an accounting question or would it be more suited to a portfolio management program?

exxus:
>This PR is for maint, but nothing stops us from merging it into master instead.

True, after I read about the difference, I also think it should go into master.
The explanation was not in
https://wiki.gnucash.org/wiki/Development
but I found it in
https://wiki.gnucash.org/wiki/Git#Pull_Requests

> because you didn't take the trouble to set up a Liabilities:Current:BankA:Overdrafts to hold the liability balance. Considering the complexities, how much effort is appropriate to support an improperly set up book?

I agree with you, so option 3 would be ok. Negative values only occur in improperly set up books and they don't need to be supported.

>  What do you really want to model, and is it really an accounting question or would it be more suited to a portfolio management program?

Since Gnucash offers the option to add shares, I still think the basic addition to the report would be useful.
If you want to do more complicated portfolio management, of course a separated, more suited program would be needed.

jralls:
> I agree with you, so option 3 would be ok. Negative values only occur in improperly set up books and they don't need to be supported.

OK. @christopherlam, your thoughts?

jralls:
>  "The ratio chart cannot report on accounts with negative balances. Please unselect accounts with negative balances."

That'll do, but let's think of a more specific name than "ratio chart".

exxus:
> That'll do, but let's think of a more specific name than "ratio chart".

What about "percentage chart", or "bar chart illustrating value distribution"?

jralls:
"Percentage chart" isn't any different than "ratio chart". "Bar chart illustrating value distribution" might work as a caption for a chart in a blog post, it's not suitable as a title.  "Allocation percentage chart" might work. @fellen, any ideas?

exxus:
@christopherlam thank you for the updates.
I noticed the following: The text still contains the wording "percentage ratios",
jralls sugested:
> "Percentage chart", "Display account contributions as a percentage of the total assets for the period."

About the description for release notes:
Should I update the description? It is much simpler now, neglecting negative amounts.

jralls:
@exxus Yes, you have to make the requested changes. Nobody is going to do it for you.

exxus:
Since it is not my pull request anymore I can't add changes/commits. I can either do a new pull request for 
exxus-category-barchart-patch or github.com/Gnucash/gnucash:master or  github.com/Gnucash/gnucash:maint
What is the preferred way to go?

Where to put the updated description?:

An option ("Replace amounts with percentage charts.") was added to category-barchart reports.
This option will replace the total amount of each account with the ratio with which it contributed to the total assets for the period.
The intended use case is to maintain a certain ratio between the account classes (e.g. investment funds and fixed-interest securities with a ratio of 60% : 40%).

The following screenshots explain this:
assets2
This picture shows the amount for the "fixed-interest securities" and the "investment funds" account. This report was already available.

assets2-ratios
This picture show the bar chart of the assets with the assigned ratios. One can see that at the beginning, most of the total money was in "fixed-interest securities". Due the changes of the amounts, at the end, most of the total money is in "investment funds" with a ratio of 60% "investment funds" and 40% "fixed-interest securities.

The caveats are negative values, as there is no clear understanding what the share of the negative value in the Grand Total is.
Since negative values only occur in improperly set up books, they are neglected.

christopherlam:
Thanks @exxus the strings relate to the error messages shown instead of a chart. I'll need to amend and push. Awaiting views from @fellen who is heavily involved in internationalization.

jralls:
@christopherlam, @fellen is busy with real-life stuff, so if he hasn't found time to comment by tomorrow go ahead and merge this. If there are i18n issues we can fix them later.

christopherlam:
Thanks @exxus !

PR #1276, opened 2022-02-13
taxinvoice.scm: removed trailing spaces

CWehli:


fellen:
Some questions:
Does it have an impact on  gnucash.pot? More, less different msgS?
Can you make pot before your change, rename it to gnucash-old.pot
Apply your change, make pot and compare it with gnucash-old.pot.

BTW I would prefer to move also the colon out of the translatable string. Perhaps @christopherlam has a better option generator in other modules?

jralls:
It's nothing to do with the option generator, those are the default values that are passed to the string option generator. I agree that since they're headings over columns that the colons aren't appropriate.

CWehli:
In invoice.scm there are also example entries with colons, but they have no trailing spaces. Two of the five entries are used in both reports and are placed in the header of the invoice. These are not column headers.

I am of the opinion that the colon already makes sense at this point.

jralls:
Ok, I looked more closely and these are block leaders, so the colons are correct. The trailing spaces are ignored because html. Thanks.

PR #1280, opened 2022-02-19
qof-string-cache.cpp to use C++ STL

christopherlam:
... instead of GHashTable.


libgnucash/engine/qof-string-cache.cpp
```diff
@@ -43,98 +41,60 @@ static QofLogModule log_module = QOF_MOD_UTIL;
...
-        }
+    if (!key)
+        return nullptr;
+
+    if (!key[0])
+        return "";
+
+    auto map_iter = cache.find (key);
+    if (map_iter == cache.end())
+        map_iter = cache.insert_or_assign (key, 1).first;
```
jralls, line 93: You've already established that `key` isn't present so there's no point in `insert_or_assign`, just use `emplace`.

libgnucash/engine/qof-string-cache.cpp
```diff
@@ -43,98 +41,60 @@ static QofLogModule log_module = QOF_MOD_UTIL;
...
 
 void
 qof_string_cache_destroy (void)
 {
-    if (qof_string_cache)
-    {
-        g_hash_table_destroy(qof_string_cache);
-    }
-    qof_string_cache = NULL;
+    cache.clear ();
```
jralls, line 59: Unnecessary, and since `qof_string_cache_init` and `qof_string_cache_destroy` are each used only twice--once in the unit test and once in qofutil.cpp--it's better to just remove them.

libgnucash/engine/qof-string-cache.cpp
```diff
@@ -43,98 +41,60 @@ static QofLogModule log_module = QOF_MOD_UTIL;
 /* =================================================================== */
 /* The QOF string cache                                                */
 /*                                                                     */
-/* The cache is a GHashTable where a copy of the string is the key,    */
+/* The cache is an std:unordered_map where an std::string is the key,  */
```
jralls, line 44: "a std::unordered_map": "std" is pronounced "stud" or "standard".

libgnucash/engine/qof-string-cache.cpp
```diff
@@ -43,98 +41,60 @@ static QofLogModule log_module = QOF_MOD_UTIL;
...
-    {
-        if (key[0] == 0)
-        {
-            return "";
-        }
+    if (!key)
+        return nullptr;
+
+    if (!key[0])
+        return "";
```
jralls, line 89:
```
if (!key || !*key)
    return key;
```

jralls (changes requested):
Pending benchmark results.

libgnucash/engine/qof-string-cache.cpp
christopherlam: I'm not 100% sure about this one. This function is meant to return the *cached* null terminated string. Is it safe to return the *original* empty string?

christopherlam:
Unfortunately benchmarks are slower with STL

STL
```
33: # Start of qof-string-cache tests
33: 1 insertions, 0 deletions: elapsed = 0.313046
33: 3 insertions, 0 deletions: elapsed = 0.452090
33: 5 insertions, 5 deletions: elapsed = 1.473400
33: 8 insertions, 12 deletions: elapsed = 3.060178
33: ok 80 /qof/qof-string-cache/string-cache
33: # End of qof-string-cache tests
```

Glib
```
33: # Start of qof-string-cache tests
33: 1 insertions, 0 deletions: elapsed = 0.113515
33: 3 insertions, 0 deletions: elapsed = 0.212478
33: 5 insertions, 5 deletions: elapsed = 0.702014
33: 8 insertions, 12 deletions: elapsed = 1.438659
33: ok 80 /qof/qof-string-cache/string-cache
```

libgnucash/engine/qof-string-cache.cpp
jralls: True, it could cause a use-after-free if a variable is set to CACHE_INSERT(some_allocated_emtpy_string). Better then to let it fall through and cache an empty std::string.

jralls:
Thought it might. That 8 insertions and 12 deletions takes a second and a half even on the C version suggests that the string cache itself might be misguided.

christopherlam:
> Thought it might. That 8 insertions and 12 deletions takes a second and a half even on the C version suggests that the string cache itself might be misguided.

"8 insertions" doesn't quite mean only 8 string insertions...

https://github.com/Gnucash/gnucash/blob/bf2f2880e460fbddcfd7a1e2bcec220bf9db96cd/libgnucash/engine/test/test-qof-string-cache.c#L53

jralls:
Ah, more like 2m operations in 1.4 seconds. That's more reasonable. But it's not a very good test: The last 400,000 deletes will be no-ops on an empty hash table/unordered_map and the strings are so similar that there will be frequent re-hashing and long bucket traversals.

christopherlam:
> Ah, more like 2m operations in 1.4 seconds. That's more reasonable. But it's not a very good test: The last 400,000 deletes will be no-ops on an empty hash table/unordered_map and the strings are so similar that there will be frequent re-hashing and long bucket traversals.

No, https://github.com/Gnucash/gnucash/blob/bf2f2880e460fbddcfd7a1e2bcec220bf9db96cd/libgnucash/engine/test/test-qof-string-cache.c#L104 ensures the map is populated  incrementally; each unit being 100,000 strings:- +1 +3 +5 -5 +8 -12 = 0. Anyway I suspect the overhead in std::string is to blame rather than unordered_map performance. Using https://stackoverflow.com/questions/20649864/c-unordered-map-with-char-as-key to use char* as key instead may work but makes it all 🤮.

jralls:
OK about the sum increments = decreements. The std::string allocations and any rehashing happens in the first `time_it (1, 0)` call, after that it's only incrementing and decrrementing the counts, so it isn't std::string's fault. It must be something about std::unordered_map vs. g_hash_table; maybe std::stringhash is slower than  the very simple hash used by g_hash_table? Just speculating, haven't profiled it.

christopherlam:
> OK about the sum increments = decreements. The std::string allocations and any rehashing happens in the first `time_it (1, 0)` call, after that it's only incrementing and decrrementing the counts, so it isn't std::string's fault. It must be something about std::unordered_map vs. g_hash_table; maybe std::stringhash is slower than the very simple hash used by g_hash_table? Just speculating, haven't profiled it.

Hmm. If the cache is `std::unordered_map<std::string,int>`, and we access cache via `auto map_iter = cache.find (key);`, where `key` is `char*` would this create a temporary `std::string` from the `char*` then immediately hash it for the lookup?

bhardwajs:
> > OK about the sum increments = decreements. The std::string allocations and any rehashing happens in the first `time_it (1, 0)` call, after that it's only incrementing and decrrementing the counts, so it isn't std::string's fault. It must be something about std::unordered_map vs. g_hash_table; maybe std::stringhash is slower than the very simple hash used by g_hash_table? Just speculating, haven't profiled it.
> 
> Hmm. If the cache is `std::unordered_map<std::string,int>`, and we access cache via `auto map_iter = cache.find (key);`, where `key` is `char*` would this create a temporary `std::string` from the `char*` then immediately hash it for the lookup?

@christopherlam - Could this be a situation where using `is_transparent` would provide faster lookups?
https://www.fluentcpp.com/2017/06/09/search-set-another-type-key/

christopherlam:
> > Hmm. If the cache is `std::unordered_map<std::string,int>`, and we access cache via `auto map_iter = cache.find (key);`, where `key` is `char*` would this create a temporary `std::string` from the `char*` then immediately hash it for the lookup?
> 
> @christopherlam - Could this be a situation where using `is_transparent` would provide faster lookups? https://www.fluentcpp.com/2017/06/09/search-set-another-type-key/

Bah. if @bhardwajs has the magic formula I'm all ears.

jralls:
No, is_transparent won't help because you're not using a set of structs, you're using a map of pairs. It's quite possible that passing a char* to std::hash<std::string> will cause construction of a std::string temporary and if the key is bigger than the small-string optimization limit for the compiler that will cause an allocation. You could try constructing a std::string_view from the char* and passing that to `find`, but one would hope that the libstdc++ authors would have provided a direct hashing function for char* to avoid the temporary.

Best way to know what's causing the slowdown is to profile it, but you could also have a look at it in [compiler explorer](https://www.godbolt.org) to see if anything glaring pops out of the assembly code.

christopherlam:
fd87eea in github actions reduces from 
```
# Start of qof-string-cache tests
1 insertions, 0 deletions: elapsed = 0.081506
3 insertions, 0 deletions: elapsed = 0.141935
5 insertions, 5 deletions: elapsed = 0.483330
8 insertions, 12 deletions: elapsed = 1.049015
ok 80 /qof/qof-string-cache/string-cache
# End of qof-string-cache tests
```
to
```
# Start of qof-string-cache tests
1 insertions, 0 deletions: elapsed = 0.074098
3 insertions, 0 deletions: elapsed = 0.121549
5 insertions, 5 deletions: elapsed = 0.396022
8 insertions, 12 deletions: elapsed = 0.827708
ok 80 /qof/qof-string-cache/string-cache
# End of qof-string-cache tests
```

christopherlam:
May need https://gist.github.com/facontidavide/95f20c28df8ec91729f9d8ab01e7d2df

jralls:
Maybe something like that, but that particular code won't work because it's the wrong string_view.

christopherlam:
Slighlty better with `unordered_map<std::string_view,int>` in 1e995f6 but insertion needs `g_strdup` and deletion needs `g_free`.

```
33: # Start of qof-string-cache tests
33: 1 insertions, 0 deletions: elapsed = 0.301827
33: 3 insertions, 0 deletions: elapsed = 0.396754
33: 5 insertions, 5 deletions: elapsed = 1.316918
33: 8 insertions, 12 deletions: elapsed = 2.691642
33: ok 80 /qof/qof-string-cache/string-cache
33: # End of qof-string-cache tests
```

reverting to C++17 standard in [e1ddaf8](https://github.com/Gnucash/gnucash/pull/1280/commits/e1ddaf859267dd9791b973e4146b1faaf79f8fd5) leads to tiny deterioration:
```
33: # Start of qof-string-cache tests
33: 1 insertions, 0 deletions: elapsed = 0.310707
33: 3 insertions, 0 deletions: elapsed = 0.424167
33: 5 insertions, 5 deletions: elapsed = 1.411321
33: 8 insertions, 12 deletions: elapsed = 2.858163
33: ok 80 /qof/qof-string-cache/string-cache
33: # End of qof-string-cache tests
```

jralls:
A small improvement over the original std::string impl but still much worse than the existing g_hash_table one.

You might find this Scott Meyers (the Effective C++ guy) [blog post](http://scottmeyers.blogspot.com/2015/09/should-you-be-using-something-instead.html) interesting, especially the discussion about hashing algorithms. The benchmark you're using produces unrealistic strings that, as I pointed out earlier, will be very hard to hash effectively so you might be ending up with overly large buckets. You can investigate that with the member functions `load_factor` and `bucket_size` and tune it with `max_size`, `max_load_factor`, and `max_bucket_count`. Remember that these are template functions and may not exist in the debugger unless you use them in the program.

christopherlam:
I'm not convinced the load factor during the tests is a major issue; in real use, the string-cache stores: currency ISO codes, currency names, strings "Book" "PriceDB" etc; the load factor hovers between 0.5 to 1.0. In the tests the same load factor is seen. It's simply clear that glib is simpler and faster than STL. So, I think this branch may either be withdrawn completely (or kept as a template), or allowed in as a code simplification.

```patch
modified   libgnucash/engine/qof-string-cache.cpp
@@ -88,6 +88,8 @@ qof_string_cache_insert(const char * key)
     else
         map_iter->second++;
 
+    printf ("inserting [%s], size=%ld, bucket-count=%ld, load-factor=%f\n",
+            key, cache.size(), cache.bucket_count(), cache.load_factor ());
     return map_iter->first.data();
 }
```

jralls:
Interesting. I agree about the real-world use, that was actually my point about the benchmark code.

> So, I think this branch may either be withdrawn completely (or kept as a template), or allowed in as a code simplification.

Well, it's not much of a simplification, it mostly just changes the containers. If you're reasonably sure that the benchmark performance translates to how it would perform in the application then I think the only option is to drop it.

gjanssens:
A question out of curiosity. At some point in the future, the idea is to drop glib. Will we have more effective mechanisms once the code base is fully C++ ? I mean is this reduction in performance now due to the mixed C/C++ world we currently live in ? Or will we have to accept a drop in performance at that point in the future ?

gjanssens:
> The converse--that having a string cache at all helps performance--remains to be proved.

That reminds me I did some changes to string caching years ago (on request of a customer no less) that significantly improved the performance on his huge book. IIRC those changes effectively disabled string caching in certain parts of gnucash. I only did limited profiling back then. But I remember opening certain account registers improved from taking minutes to seconds. The downside was more memory consumption (I believe I saw an increase of about 50Mb) on that enormous book. 

I didn't know much about profiling back then (and probably still don't tbh), but I think that was a reasonable trade-of in modern-day computing environments.

christopherlam:
The latest iteration of this branch doesn't use `std::string` at all, it uses a g_strdup'ed `char*` which is then managed via `std::string_view` and `g_free`d when erasing from unordered_map. Still slower than GHashTable. Nevertheless this branch aimed to modernize code base.

PR #1285, opened 2022-03-02
Preferences, Import: UPDATE AND RECONCILE should be UPDATE AND CLEAR

goodvibes2:
on preference "Enable update match action" tooltip text.


jralls:
Thanks!

PR #1286, opened 2022-03-02
various code progress

christopherlam:
~1. remove dead code~
2. fix a crasher in budget view

### DIFF
--- a/gnucash/gnome/gnc-budget-view.c
+++ b/gnucash/gnome/gnc-budget-view.c
@@ -884,7 +884,7 @@ gbv_treeview_resized_cb (GtkWidget *widget, GtkAllocation *allocation,
         if ((g_strcmp0 (name, "description") == 0) && (!priv->show_account_desc))
             total_node = g_list_next (total_node);
 
-        if (gtk_tree_view_column_get_visible (tree_view_col))
+        if (gtk_tree_view_column_get_visible (tree_view_col) && total_node != NULL)
         {
             gint col_width = gtk_tree_view_column_get_width (tree_view_col);
             GtkTreeViewColumn *totals_view_col = total_node->data;


christopherlam:
The code tries to access a null `total_node`. Hint: the gtk_tree_view_column_get_visible conditional above the crasher is usually FALSE when `total_node` is exhausted, but switching the account default currency from locale to user-supplied seems to cause the gtk_tree_view_column_get_visible to return TRUE thereby attempting to access `total_node->data` when `total_node` is NULL. I don't know why it happens; I think this branch should be merged imminently to band-aid the crasher.

gnucash/gnome-utils/gnc-tree-util-split-reg.c
```diff
@@ -550,42 +550,9 @@ gnc_tree_util_split_reg_rotate (GncTreeViewSplitReg *view, GtkTreeViewColumn *co
...
-
-        gnc_tree_view_split_reg_set_dirty_trans (view, trans);
-        if (!xaccTransIsOpen (trans))
-            xaccTransBeginEdit (trans);
-
-        xaccTransSetTxnType (trans, type);
-        return TRUE;
+        // thanks to commit 995f5d3ff, the APAR transaction type is no
+        // longer user modifiable.
+        PERR ("rotating a COL_TYPE viewcol. this shouldn't happen.");
```
jralls, line 555: Let's not have self-referring comments. Neither the comment nor the commit message explain why the transaction type shouldn't be user-modifiable. 

The error message isn't very meaningful either.
```
/* Changing from invoice to payment in the register can create a error condition
  * that can't be resolved from the user interface.
  * /
PERR ("Changing from invoice to payment in the register isn't permitted.");
```

Note that this file is for register2 only and doesn't affect the one that everyone actually uses.

But you do realize that this file is part of Register2, right? And that this function doesn't necessarily apply only to AP/AR accounts?

gnucash/gnome/gnc-budget-view.c
```diff
@@ -884,7 +884,7 @@ gbv_treeview_resized_cb (GtkWidget *widget, GtkAllocation *allocation,
         if ((g_strcmp0 (name, "description") == 0) && (!priv->show_account_desc))
             total_node = g_list_next (total_node);
 
-        if (gtk_tree_view_column_get_visible (tree_view_col))
+        if (gtk_tree_view_column_get_visible (tree_view_col) && total_node != NULL)
```
jralls, line 887: Wouldn't it make more sense to just break out of the loop when you run out of totals columns?

OTOH the premise here seems to be that if show_account_code or show_account_desc is false then `gtk_tree_view_column_get_visible()` will return false for those columns anyway so you could just track through both in the `for` statement:
```
for (GList *node = columns, *total_node = total_columns;
       node && total_node;
       node = g_list_next (node), total_node = g_list_next (total_node))
{
     GtkTreeViewColumn *tree_view_col = node->data;
     if (gtk_tree_view_column_get_visible (tree_view_col))
     {
         gint col_width = gtk_tree_view_column_get_width (tree_view_col);
         GtkTreeViewColumn *totals_view_col = total_node->data;
         if (GTK_IS_TREE_VIEW_COLUMN(totals_view_col))
             gtk_tree_view_column_set_fixed_width (totals_view_col, col_width);
     }
}
```

But if the goal is to hide the `totals_view_col` for the not-displayed `tree_view_col` shouldn't it be
```
     GtkTreeViewColumn *tree_view_col = node->data;
     GtkTreeViewColumn *totals_view_col = total_node->data;
     gint col_width = 0;
     if (gtk_tree_view_column_get_visible (tree_view_col))
         col_width = gtk_tree_view_column_get_width (tree_view_col);
     if (GTK_IS_TREE_VIEW_COLUMN(totals_view_col))
         gtk_tree_view_column_set_fixed_width (totals_view_col, col_width);
```

instead?

jralls:
The third commit message says
"because the GList modification works on the tail end, it's faster to reverse initially, modify, and re-reverse."

Shouldn't that be that modification works *faster* on the *head* end?

gnucash/gnome-utils/gnc-tree-util-split-reg.c
christopherlam: Ok will remove; @mtalexander has removed Register2 in master 3161df6 anyway.

christopherlam:
> The third commit message says "because the GList modification works on the tail end, it's faster to reverse initially, modify, and re-reverse."
> 
> Shouldn't that be that modification works _faster_ on the _head_ end?

Of course. I meant GList modification _in this code_ works on the tail end (see the g_list_last being chopped off sequentially). Modified message.

gnucash/gnome/gnc-budget-view.c
christopherlam: I'm not too familiar with the mechanics of gtk_tree_view_column; in this file columns and total_columns are of different lengths. In my budget with 2 periods, columns have 31 nodes (being derived from `gnc_tree_view_account_new` and has all acct properties) whereas total-columns have 6 nodes (presumably: acctname, desc, note, period1, period2, total) (?).

jralls: Well, that explains why it over-runs. Maybe this function is a left over from a previous design? What it's trying to do is set the column widths of the totals_view to be the same as the corresponding tree_view ones, presumably so that they'll line up. Do the first n tree_view columns correspond to the n totals_view ones so that that would make sense?

jralls: @christopherlam OK, I think. The last two columns entries correspond to totals and account currency?

christopherlam: Uh no idea... Hence the patch to block crasher. Orig discussion https://lists.gnucash.org/pipermail/gnucash-devel/2013-February/034938.html

gnucash/gnome-utils/gnc-tree-util-split-reg.c
christopherlam: > Let's not have self-referring comments. Neither the comment nor the commit message explain why the transaction type shouldn't be user-modifiable.
> 
> The error message isn't very meaningful either.

fwiw the txn_type became unmodifiable due to @Bob-IT nice work in 995f5d3ff because changing TXN_TYPE_NONE to TXN_TYPE_PAYMENT was previously possible (although illegal according to @derekatkins) and would irretrievably damage the book.

gnucash/gnome/gnc-budget-view.c
jralls: Sigh, another half-baked feature creating long-term technical debt.

There's not much point in optimizing the loop, it's short and executing in user time. Go ahead and push.

PR #1293, opened 2022-03-13
Fix suppression of "cut split/transaction" warnings for the current session

nomis:
These were not working because they were missing from the temporary warnings settings schema. The warnings could only be dismissed permanently.

Add them to the temporary warnings section of the settings schema.


jralls:
Thanks!

PR #1297, opened 2022-03-17
Account: replacing GValue use with Kvp direct access

christopherlam:
~eliminate GValue gradually~ consolidate GValue access where possible

### DIFF
--- a/libgnucash/engine/Account.cpp
+++ b/libgnucash/engine/Account.cpp
@@ -2561,6 +2561,42 @@ get_kvp_string_tag (const Account *acc, const char *tag)
     return get_kvp_string_path (acc, {tag});
 }
 
+static void
+set_kvp_int64 (Account *acc, std::vector<std::string> const & path, int value)
+{
+    GValue v = G_VALUE_INIT;
+    g_return_if_fail(GNC_IS_ACCOUNT(acc));
+
+    xaccAccountBeginEdit(acc);
+    g_value_init (&v, G_TYPE_INT64);
+    g_value_set_int64 (&v, value);
+    qof_instance_set_path_kvp (QOF_INSTANCE (acc), &v, path);
+    g_value_unset (&v);
+
+    mark_account (acc);
+    xaccAccountCommitEdit(acc);
+}
+
+static bool
+get_kvp_int64 (const Account *acc, std::vector<std::string> const & path,
+               gint64& value)
+{
+    GValue v = G_VALUE_INIT;
+    bool success = false;
+    g_return_val_if_fail(GNC_IS_ACCOUNT(acc), false);
+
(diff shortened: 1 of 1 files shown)


libgnucash/engine/Account.cpp
```diff
@@ -4084,33 +4075,46 @@ xaccAccountForEachLot(const Account *acc,
...
+
+    if (!kval)
+        return false;
+
+    if (kval->get_type () == KvpValue::Type::INT64)
+        return (kval->get<int64_t>() != 0);
+
+    if (kval->get_type () == KvpValue::Type::STRING)
+        return (!g_strcmp0 (kval->get<const char*>(), "true"));
+
```
christopherlam, line 4101: the old code has G_VALUE_HOLDS_BOOLEAN which doesn't seem to be a valid Kvp type so there's no equivalent code for g_value_get_boolean.

libgnucash/engine/Account.cpp
```diff
@@ -5094,14 +5061,8 @@ xaccAccountGetReconcileChildrenStatus(const Account *acc)
...
      */
-    GValue v = G_VALUE_INIT;
-    gboolean retval;
     if (!acc) return FALSE;
-    qof_instance_get_path_kvp (QOF_INSTANCE (acc), &v,
-            {KEY_RECONCILE_INFO, KEY_INCLUDE_CHILDREN});
-    retval = G_VALUE_HOLDS_INT64 (&v) ? g_value_get_int64 (&v) : FALSE;
-    g_value_unset (&v);
-    return retval;
+    return boolean_from_key (acc, {KEY_RECONCILE_INFO, KEY_INCLUDE_CHILDREN});
```
christopherlam, line 5065: This refactor is not completely indentical to previous. It can return true if the kvp value has string "true"; previous code would only return true if the kvp value is a non-zero int. This side effect *should* be unimportant in practice.

libgnucash/engine/Account.cpp
jralls: Yes, `g_value_from_kvp_value` can't return a boolean GValue so that branch wouldn't be exercised. This is fine.

jralls:
I'm not really happy about de-abstracting the KVP code. Yes, GValues are inefficient, but it's a long term goal to make KVP a backend detail instead of as an end-around the type system. I created the abstraction to make that easier when someone has the time to work on it and reversing it "for efficiency" isn't forward progress.

jralls:
Um, you missed the point. GCValue is an implementation detail of qof_instance_get/set that's necessitated by not being able to use std::variant in C code. An overloaded implementation of qof_instance_get for c++ that uses std::variant would be OK... but here you've done the worst possible by dropping the abstraction (qof_instance_get) and keeping its inefficient GValue implementation.

PR #1299, opened 2022-03-21
Python gnc numeric example

c-holtermann:
This is intended to be as an example file for GncNumeric to provide a better understanding (for me and others) of how gnucash handles numerics. It still has issues discussed in https://github.com/Gnucash/gnucash/pull/1298. I'd like to pick up that conversation. Just now the file has not changed in respect to the aforementioned PR. I will work on it this week and add changes to this PR.

One main topic is the conversion of float to GncNumeric. The basic practical question for me is how to achieve to have a specific decimal value like "2.2" as a GncNumeric. The way I'd expect it to work is GncNumeric(2.2) but this uses a float which does not have the exact value of 2.2 as has been showed by @jralls in https://github.com/Gnucash/gnucash/pull/1298#discussion_r830545454.

So I'm looking for practical ways to have exact decimal values represented in GncNumeric coming from python script or shell and to show possible problems. In this end this PR should show a practical solution but as now has flaws as John Ralls has pointed out.

The significance of the significant figures and their practical value and handling needs to be worked out.

### DIFF
--- a/bindings/python/example_scripts/GncNumeric.py
+++ b/bindings/python/example_scripts/GncNumeric.py
@@ -40,6 +40,15 @@ print(n4, "==", g4, ":", n4 == g4)
 print('Same float with higher precision: %.32f' % n4)
 print()
 
+# Create from float
+n5 = 2245.67
+g5 = GncNumeric(n5)
+print("5 - create from float")
+print(n5, "(", type(n5), ") =>", g5)
+print(n5, "==", g5, ":", n5 == g5)
+print('Same float with higher precision: %.32f' % n5)
+print()
+
 # Compare float, string and int
 print("Compare GncNumeric from float (4) to GncNumeric from string (2)")
 print(g2, "==", g4, ":", g2 == g4)
@@ -67,6 +76,18 @@ print()
 
 # Convert GncNumeric from float to remove errors
 print("Convert float to different denominator to remove conversion errors")
-g5 = g4.convert(1000, GNC_HOW_RND_ROUND)
-print(g4, "(", type(g4), ") =>", g5)
+g4b = g4.convert(1000, GNC_HOW_RND_ROUND)
+print(g4, "(", type(g4), ") =>", g4b)
+print()
+
+# Create from float with specified significant figures
+print(f"Create from float (Example 5, ={n5}) with specified significant figures")
+for sigfigs in range(0,20):
+    print(sigfigs, GncNumeric(n5, GNC_DENOM_AUTO, GNC_HOW_DENOM_SIGFIGS(sigfigs)))
+print()
+
+# Convert GncNumeric from float to remove errors
+print("Convert float to different denominator to remove conversion errors")
+g5b = g5.convert(1000, GNC_HOW_RND_ROUND)
+print(g5, "(", type(g5), ") =>", g5b)
 
--- a/bindings/python/example_scripts/GncNumeric.py
+++ b/bindings/python/example_scripts/GncNumeric.py
@@ -53,10 +53,16 @@ print("Compare GncNumeric from string (2) to GncNumeric from int (3)")
 print(g2, "==", g3, ":", g2 == g3)
 print()
 
-# Create from float with specified precision
-print("Create from float with specified precision")
-for n in range(0,20):
-    print(GncNumeric(n4, GNC_DENOM_AUTO, GNC_HOW_DENOM_SIGFIG | n * 0x100))
+def GNC_HOW_DENOM_SIGFIGS(n):
+    """Leaned on the C Macro"""
+    return GNC_HOW_DENOM_SIGFIG | (n << 8 & GNC_NUMERIC_SIGFIGS_MASK)
+
+GNC_NUMERIC_SIGFIGS_MASK = 0x0000ff00
+
+# Create from float with specified significant figures
+print(f"Create from float (Example 4, ={n4}) with specified significant figures")
+for sigfigs in range(0,20):
+    print(sigfigs, GncNumeric(n4, GNC_DENOM_AUTO, GNC_HOW_DENOM_SIGFIGS(sigfigs)))
 print()
 
 # Convert GncNumeric from float to remove errors
--- a/bindings/python/example_scripts/GncNumeric.py
+++ b/bindings/python/example_scripts/GncNumeric.py
@@ -37,6 +37,7 @@ g4 = GncNumeric(n4)
 print("4 - create from float")
 print(n4, "(", type(n4), ") =>", g4)
 print(n4, "==", g4, ":", n4 == g4)
+print('Same float with higher precision: %.32f' % n4)
 print()
 
 # Compare float, string and int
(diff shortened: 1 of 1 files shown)


bindings/python/example_scripts/GncNumeric.py
```diff
@@ -0,0 +1,66 @@
...
+print(g2, "==", g3, ":", g2 == g3)
+print()
+
+# Create from float with specified precision
+print("Create from float with specified precision")
+for n in range(0,20):
+    print(GncNumeric(n4, GNC_DENOM_AUTO, GNC_HOW_DENOM_SIGFIG | n * 0x100))
+print()
+
+# Convert GncNumeric from float to remove errors
```
c-holtermann, line 62: see https://github.com/Gnucash/gnucash/pull/1298#pullrequestreview-915135815

bindings/python/example_scripts/GncNumeric.py
```diff
@@ -0,0 +1,66 @@
...
+
+print("Compare GncNumeric from float (4) to GncNumeric from int (3)")
+print(g3, "==", g4, ":", g3 == g4)
+print()
+
+print("Compare GncNumeric from string (2) to GncNumeric from int (3)")
+print(g2, "==", g3, ":", g2 == g3)
+print()
+
+# Create from float with specified precision
```
c-holtermann, line 56: see https://github.com/Gnucash/gnucash/pull/1298#pullrequestreview-915135433

bindings/python/example_scripts/GncNumeric.py
c-holtermann: I've moved the sigfig calculation to a function like the C Makro in https://github.com/Gnucash/gnucash/blob/bbb4113a5a996dcd7bb3494e0be900b275b49a4f/libgnucash/engine/gnc-numeric.h#L218. I've removed the magic number which is identical to a bitshift by 8.

jralls:
> One main topic is the conversion of float to GncNumeric.

Just don't. Use [Decimal](https://docs.python.org/3/library/decimal.html).

>  The basic practical question for me is how to achieve to have a specific decimal value like "2.2" as a GncNumeric

The problem with 2.2 is that it's a trivial example.

bindings/python/example_scripts/GncNumeric.py
jralls: What SWIG does with macros depends to some extent on the backend, but in general I think that it wraps them in a function that it aliases in the target language.

While it's true that GNC_HOW_DENOM_SIGFIGS is part of the gnc_numeric API I think it's a remarkably bad idea to use it for accounting. Significant Figures *does not* mean decimal places, it means the number of digits in the numerator: 2.25 has 3 significant digits, 42.25 has 4. So does 4225, but 4225.00 has 6. It should be immediately apparent how a significant digits round can create an imbalance in an accounting computation.

All of which aside you're just replaced one magic number with another. In programming a "magic number" is an undocumented numeric literal; the solution is to declare a descriptively named constant, e.g. 
```C
GNC_NUMERIC_SIGFIGS_OFFSET = 8;
/* ... */
return GNC_HOW_DENOM_SIGFIG | (n << GNC_NUMERIC_SIGFIGS_OFFSETe & GNC_NUMERIC_SIGFIGS_MASK);
```

jralls:
There's a fundamental problem with adding this file to GnuCash; Most people will regard code in an example file as an illustration of how to do something. Much of this file illustrates using float and GNC_HOW_DENOM_SIGFIGS, neither of which is a good idea.

c-holtermann:
@jralls without knowing about the issues we are discussing I guess every naive person would expect that converting something like `n=2.2` into a GncNumeric would lead to an exact conversion without giving a thought that this includes a conversion to float that turns n into `nf = float(n)` which is similar to n but not exactly as it includes a small error e that is due to the implementation of floats. So `nf = n + e` so that `nf ≠ n`. At least I was surprised seeing that `nf ≠ n` or  expressed as GncNumeric:
```
>>> GncNumeric(2.2) != GncNumeric("2.2")
True
```

2.2 is a common value for accounting, at least in my case. I think it's relevant to look at trivial cases as well as they are the common ones while taking also into account border cases. Of course different people might use GnuCash for different things and different numbers.

I agree with you that an example file should demonstrate a way that is safe, exact, robust and suggested while also showing possible problems in other approaches. I will work on this until I understand it enough learning from your remarks. Until then lets take the example file as basis for our discussion. When we agree on that it also makes sense for me to add some information to the docs when what I can add is actually insightful ;)

From a practical point when entering a transaction through the python bindings I'd like to be able to type 2.2 in a short way and actually have 2.2 in gnucash.

I.e. when I create a transaction I'd like it to be as simple as something like Transaction(date, Split(AccA, "Text", n), Split(AccB, "Text", -n)) or even T(date, AccA, AccB, n, "Text1", "Text2"). In that case I would be happy to be able to use a float as n as that what python turns numbers into automatically, being aware of the problems.

The conversion from string to float is where the problem occurs. And that is in the python language, not in GnuCash.

GnuCash could warn each time a conversion from float occurs. Or the python bindings could be enabled to remove the errors automatically (maybe like I suggest below).

In any case the example file should contain a warning about the rounding issues and make the issue clear

I wonder how python shows 2.2 as 2.2 and not 2.20000000000000017763568394002505.

I wonder if there could be a default for significant figures in the conversion from float to GncNumeric in the python bindings that round nf back to n with a number of significant figures. The usage of significant figures could be set manually. This could be made optional.

Something like

```
>>> GncNumeric(2.2) == GncNumeric("2.2")
False
>>> GncNumeric.sigfigs = 2
>>> GncNumeric(2.2) == GncNumeric("2.2")
True
```

Maybe I'm completely off the track here. But in that case I'm sure you'll show the right direction :)

I have read the C++ GncNumeric remarks about sigfigs: https://github.com/Gnucash/gnucash/blob/bbb4113a5a996dcd7bb3494e0be900b275b49a4f/libgnucash/engine/gnc-numeric.hpp#L203.

I've included another number 2245.67 in the example. With this included I see that I don't really understand the GnuCash concept of significant figures.

I expected 2245.67 with two significant figures to be `224567/100`. But that is the result for `n=6` in the example file.

--work in progress--

c-holtermann:
I realized that I mistook significant numbers for decimal places.

jralls:
> 2.2 is a common value for accounting, at least in my case. I think it's relevant to look at trivial cases as well as they are the common ones while taking also into account border cases. Of course different people might use GnuCash for different things and different numbers.

Sorry, I wasn't clear enough in that comment. 2.2 is trivial because it has so few digits that it's easy to separate the decimal "signal" from the binary-representation "noise". Consider this more interesting case:
```
>>> from decimal import *
>>> Decimal(321901) / (1709)
Decimal('188.3563487419543592744294909')
>>> print("%.32f" % (321901 / 1709))
188.35634874195434917965030763298273
```
The number changes at the 17th digit just like your 2.2 case, but since it lacks the long string of zeros it's a lot less obvious. Note that it's digits, not decimals; for example,
```
>>> print("%.32f" % (4093082899/7))
584726128.42857146263122558593750000000000
>>> Decimal(4093082899) / Decimal (7)
Decimal('584726128.4285714285714285714')
```
And before you claim that nobody deals in that many digits in real life, consider the exchange rate between the Indonesian Rupiah and Bitcoin: 630,401,319.40 IDR/BTC. As a reminder, BTC's SCU is 10<sup>-9</sup>.

PR #1300, opened 2022-03-21
Example: plot account balance w pandas and matplotlib

c-holtermann:
Python example file.

Create a temporary file with two Accounts and
100 random transactions in the 20th century
Feed dates and balances to pandas and show plot
using matplotlib


luzpaz:
Can you show a screenshot of proposed results of script ?

luzpaz:
bump

PR #1303, opened 2022-03-23
[income-statement] single-column doesn't need separate inc/exp tables

christopherlam:
Because it's true that a single-column income-statement should not require nested html-tables. Thanks to AdrienM for debugging. https://lists.gnucash.org/pipermail/gnucash-user/2022-March/100354.html

### DIFF
--- a/gnucash/report/reports/standard/income-statement.scm
+++ b/gnucash/report/reports/standard/income-statement.scm
@@ -441,10 +441,7 @@
                                 trading-total
                                 (gnc:collector- expense-total)))
 
-               (inc-table (gnc:make-html-table))
-               (exp-table (gnc:make-html-table))
-               (tra-table (gnc:make-html-table))
-
+               (build-table (gnc:make-html-table))
                (table-env
                 (list
                  (list 'start-date start-date)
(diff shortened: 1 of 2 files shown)


gnucash/report/reports/standard/income-statement.scm
```diff
@@ -494,67 +491,86 @@
...
+                   (tra-cell (if (null? trading-accounts)
+                                 (gnc:html-make-empty-cell)
+                                 (gnc:make-html-table-cell tra-table)))
+                   (exp-cell (gnc:make-html-table-cell exp-table)))
+              (define (add-cells . lst) (gnc:html-table-append-row! build-table lst))
+              (add-rule (if standard-order? exp-table inc-table))
+              (add-report-line
+               (if standard-order? exp-table inc-table)
+               (string-append (G_ "Net income") period-for)
+               (string-append (G_ "Net loss") period-for)
```
fellen, line 533: Here and later: If they are headers, shouldn't the final strings be capitalized?

jralls: Yeah, and then translated to the uncapitalized form in en_GB.po. But since @christopherlam and @Bob-IT think in English English it might be asking too much for them to remember that. "Two peoples separated by a common tongue" and all that.

PR #1305, opened 2022-03-26
GNC_HOW_DENOM_SIGFIGS(n) to calculate denominator type for sigfigs

c-holtermann:
Python implementation of C macro GNC_HOW_DENOM_SIGFIGS(n) from gnc-numeric.h

### DIFF
--- a/bindings/python/gnucash_core.py
+++ b/bindings/python/gnucash_core.py
@@ -800,11 +800,20 @@ from gnucash.gnucash_core_c import \
     GNC_HOW_RND_PROMOTE, GNC_HOW_RND_ROUND_HALF_DOWN, \
     GNC_HOW_RND_ROUND_HALF_UP, GNC_HOW_RND_ROUND, GNC_HOW_RND_NEVER
 
+# GncNumeric bitmasks for HOW flags
+from gnucash.gnucash_core_c import \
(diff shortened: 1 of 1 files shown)


c-holtermann:
I have described my intention in https://github.com/Gnucash/gnucash/pull/1299#issuecomment-1077395035

I want to be able to enter floats in python and have them rounded to remove the implementation specific inadequacy so that `GncNumeric(2.2) == GncNumeric("2.2")`. I would like to be able to define a default behavior for conversion. Something like `GncNumeric.default_float_conversion = foo`.  It seemed to me that using the sigfigs is a valid way as this is already implemented. But thinking about it after your objections it comes to me that it might be adequate to set the denominator to something like 100 (if dealing with 1/100th as the smallest size) with adequate rounding than using sigfigs. Sigfigs would not actually meet what I intend. I guess sticking to them comes from my mistranslating sigfigs to figures after decimal point.

So my intended usecase just seems to faint away. I'll think about a bit more but it might be better then to retreat this PR and mention in the python bindings or the wiki that significant figure handling will not be ported to the bindings to have a clear message there.

jralls:
@c-holtermann Have you had enough of a think? Can we close this?

jralls:
@c-holtermann Three weeks later it's time to close it.

c-holtermann:
Thanks for cleaning up.

PR #1306, opened 2022-03-27
Move function from scheme to cpp

christopherlam:
Much more efficient. Avoids generating GLists of accounts' descendants repeatedly prior to deleting duplicates.

### DIFF
--- a/libgnucash/engine/Account.cpp
+++ b/libgnucash/engine/Account.cpp
@@ -49,6 +49,7 @@ extern "C" {
 
 #include <numeric>
 #include <map>
+#include <unordered_set>
 
 static QofLogModule log_module = GNC_MOD_ACCOUNT;
 
@@ -6226,6 +6227,21 @@ gboolean xaccAccountRegister (void)
     qof_class_register (GNC_ID_ACCOUNT, (QofSortFunc) qof_xaccAccountOrder, params);
 
     return qof_object_register (&account_object_def);
+}
+
+using AccountSet = std::unordered_set<Account*>;
+static void maybe_add_descendants (Account* acc, gpointer arg)
+{
+    if (static_cast <AccountSet*> (arg)->insert (acc).second)
+        g_list_foreach (GET_PRIVATE(acc)->children, (GFunc) maybe_add_descendants, arg);
+};
+
+GList *
+gnc_accounts_and_all_descendants (GList *accounts)
+{
+    AccountSet accset;
+    g_list_foreach (accounts, (GFunc) maybe_add_descendants, &accset);
+    return std::accumulate (accset.begin(), accset.end(), (GList*) nullptr, g_list_prepend);
 }
 
 /* ======================= UNIT TESTING ACCESS =======================
--- a/bindings/engine-common.i
+++ b/bindings/engine-common.i
@@ -48,10 +48,14 @@ AccountList * gnc_account_get_descendants (const Account *account);
 %newobject gnc_account_get_descendants_sorted;
 AccountList * gnc_account_get_descendants_sorted (const Account *account);
 
+%newobject gnc_accounts_and_all_descendants;
+AccountList * gnc_accounts_and_all_descendants (AccountList *accounts);
+
 %ignore gnc_account_get_children;
 %ignore gnc_account_get_children_sorted;
 %ignore gnc_account_get_descendants;
 %ignore gnc_account_get_descendants_sorted;
+%ignore gnc_accounts_and_all_descendants;
 %include <Account.h>
 
 %include <Transaction.h>
--- a/gnucash/report/html-acct-table.scm
+++ b/gnucash/report/html-acct-table.scm
@@ -697,7 +697,7 @@
           (for-each
            (lambda (acct)
              (this-collector 'merge (get-balance acct-balances acct) #f))
-           (gnc:accounts-and-all-descendants (list account)))
+           (cons account (gnc-account-get-descendants-sorted account)))
           this-collector))
 
       (let lp ((accounts (if less-p (sort accts less-p) accts))
--- a/gnucash/report/html-acct-table.scm
+++ b/gnucash/report/html-acct-table.scm
@@ -555,6 +555,7 @@
   (gnc:string-locale<? (xaccAccountGetName a)
                        (xaccAccountGetName b)))
 (define (gnc:account-path-less-p a b)
+  (issue-deprecation-warning "gnc:account-path-less-p is deprecated. use gnc:account-full-name<? instead")
   (gnc:string-locale<? (gnc-account-get-full-name a)
                        (gnc-account-get-full-name b)))
 
--- a/gnucash/report/report-utilities.scm
+++ b/gnucash/report/report-utilities.scm
@@ -234,9 +234,10 @@
 
 ;; Return accountslist *and* their descendant accounts
 (define (gnc:accounts-and-all-descendants accountslist)
-  (sort-and-delete-duplicates
-   (apply append accountslist (map gnc-account-get-descendants accountslist))
-   gnc:account-path-less-p equal?))
+  (issue-deprecation-warning "gnc:accounts-and-all-descendants is \
+now deprecated, use gnc-accounts-and-all-descendants instead. sort \
+with gnc:account-full-name<? if necessary.")
+  (sort (gnc-accounts-and-all-descendants accountslist) gnc:account-full-name<?))
 
 ;;; Here's a statistics collector...  Collects max, min, total, and makes
 ;;; it easy to get at the mean.
--- a/gnucash/report/reports/example/average-balance.scm
+++ b/gnucash/report/reports/example/average-balance.scm
@@ -296,7 +296,7 @@
          (accounts   (get-option gnc:pagename-accounts (N_ "Accounts")))
          (dosubs?    (get-option gnc:pagename-accounts optname-subacct))
          (accounts (if dosubs?
-                       (gnc:accounts-and-all-descendants accounts)
+                       (gnc-accounts-and-all-descendants accounts)
                        accounts))
          (plot-type  (get-option gnc:pagename-display (N_ "Plot Type")))
          (show-plot? (get-option gnc:pagename-display (N_ "Show plot")))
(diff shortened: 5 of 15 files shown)


christopherlam:
A previous iteration used xaccAccountOrder which prioritises acct->code and caused layout changes in reports. This fixed version uses locale-sensitive cpp collator on the acct fullnames to sort the accounts.

libgnucash/engine/Account.cpp
```diff
@@ -6226,6 +6227,39 @@ gboolean xaccAccountRegister (void)
     return qof_object_register (&account_object_def);
 }
 
+
+using AccountSet = std::unordered_set<Account*>;
+static void add_descendant (Account *acc, gpointer data)
+{
+    static_cast<AccountSet*> (data)->emplace (acc);
```
jralls, line 6234: `insert`, not `emplace`: `emplace` only works for C++ objects with constructors and expects a parameter list; `insert` expects an object reference and that's what you're passing.

libgnucash/engine/Account.cpp
```diff
@@ -6226,6 +6227,39 @@ gboolean xaccAccountRegister (void)
     return qof_object_register (&account_object_def);
 }
 
+
+using AccountSet = std::unordered_set<Account*>;
```
jralls, line 6231: I suppose you're using a set instead of a vector to enforce uniqueness, but what's the point of hashing the ptr? That is, why use an unordered_set instead of just a set?

christopherlam: https://www.geeksforgeeks.org/unordered_set-in-cpp-stl/

Unordered set uses a hash table and set uses a red black tree, which needs rebalancing after each insertion?

jralls: OK, but we've already demonstrated elsewhere that the C++ std library hash functions are expensive. You could fix that by providing hash and compare functions (compiles but not tested):
```
#include <unordered_set>
#include <cstdint>
/*...*/
constexpr auto hasher = [](const Account* const acct) { auto rv =std::reinterpret_cast<uintptr_t>(acct); };
constexpr auto compare = [](const Account* const lhs, const Account* const rhs) { return lhs == rhs; };
using AccountSet = std::unordered_set<Account*, hasher, compare>;
```

christopherlam: not sure it's worth while the added complexity: https://stackoverflow.com/questions/20953390/what-is-the-fastest-hash-function-for-pointers

jralls: Proves my point. What complexity are you complaining about? (BTW, I think his benchmarking is skewed by the assumption that all of the pointers are allocated from a single-type pool like GSlice so the pointers are aligned on `sizeof(T)`.  That's not the case in GnuCash so use `sizeof(unintptr_t)` to detect 32-bitness:
```
constexpr size_t shift_size = sizeof(uintptr_t) == 4 ? 1 : 3;
constexpr auto hasher = [](const Account* const acct) { auto rv = std::reinterpret_cast<uintptr_t>(acct) >> shift_size; };
```

OTOH, you're accumulating the Accounts by traversing the tree. There should be no duplicates and a std::vector is sufficient and much simpler.

christopherlam: > Proves my point. What complexity are you complaining about?

I mean the complexity of offering a custom hasher.

> OTOH, you're accumulating the Accounts by traversing the tree. There should be no duplicates and a std::vector is sufficient and much simpler.

There's plenty of duplicates. Consider the acct tree

```
A
A:B
A:B:C
A:B:D
A:E
A:E:F
A:G
```

calling `gnc_accounts_and_all_descendants` with GList containing A,B,C,E,F,G will traverse all descendants of each therefore B,C,D,E,F will be processed multiple times.

The current guile code is much worse -- it calls descendants for all accounts given, each time 
generating a GList of the descendants = O(N), converting to SCM list = O(N), appends them = O(N), sorts them = O(NLogN), then traverses again eliminating duplicates = O(N). Then the gc kicks in and mops up the mess. The CPP code as above is a major speedup. To micro-optimize further seems unnecessary.

jralls: >  it calls descendants for all accounts given

So does this

> converting to SCM list = O(N)

Saved by not using Scheme for the rest, but then spent by converting the std::set into a GList for sorting., albeit for a possibly smaller N.

>  sorts them = O(NLogN), then traverses again eliminating duplicates = O(N)

The sort set is still there, though also for a smaller N. The algorithm isn't that different, really. The only supposed savings is the remove duplicates step, but absent any profiling there's no reason to assume that it's more expensive than constructing the std::unordered_set.

> Then the gc kicks in and mops up the mess.

GC, like `free`, is cheap. It's the allocation that's expensive and they're both allocating. Plus GC runs asynchronously in its own thread, possibly on a different core, while the std::unordered_set element destructors and frees run synchronously in this thread.

> calling gnc_accounts_and_all_descendants with GList containing A,B,C,E,F,G will traverse all descendants of each therefore B,C,D,E,F will be processed multiple times. The output will be A,B,C,D,E,F,G.

That's insane, but I guess it's the result of selecting A and all of its children in the account selector and then mindlessly running `gnc_account_foreach_descendant` on the result.

It's a poorly designed brute-force algorithm. A little analysis up front would reduce N and is conceptually more likely to save time: Walk the tree backwards, eliminating any child accounts: In your example, C's and D's parent is B, so no need to traverse those two. F's parent is E, ditto. The remaining B, E, and G all have parent A, so they're gone too. We just need to call gnc_account_for_each_descendant on A and we know there will be no duplicates so we can put the results directly in a GList and then sort it.

But absent actual profiling data maybe this whole exercise is bike-shedding. Is gnc:accounts_and_all_descendants on the hot trace for any of the reports that use it? Does this conversion to C++ produce a significant savings in running those reports?

christopherlam: > But absent actual profiling data maybe this whole exercise is bike-shedding. Is gnc:accounts_and_all_descendants on the hot trace for any of the reports that use it? Does this conversion to C++ produce a significant savings in running those reports?

It's not a hot function, however it's an embarrassing function in scheme. Many speedups have taken place in the past 3 years (aka death by thousand cuts) so this one is a candidate too. The predecessor https://github.com/Gnucash/gnucash/blob/868489b1c4f9b1f64fe61130e6ae28b4bad16f22/src/report/report-system/report-utilities.scm#L148 was even worse, by keeping duplicates. The resulting expanded accountlist was used for QofQuery.

libgnucash/engine/Account.cpp
```diff
@@ -6251,6 +6251,8 @@ gnc_accounts_and_all_descendants (GList *accounts)
     for (GList *n = accounts; n; n = g_list_next (n))
     {
         auto acc = static_cast<Account*>(n->data);
+        if (accset.find (acc) != accset.end())
+            continue;
         add_descendant (acc, &accset);
```
jralls, line 6256: That skips head accounts already in the set, but doesn't to anything for descendants already in the set unless the top most account happens to be first in the list.
`gnc_account_foreach_descendant` is a simple recursor so you could do a recursion function that does the same check, or you could use the rv of `insert`, e.g.

```
static void 
maybe_add_descendants (AccountSet& set, Account* acc)
{
     auto [iter, succeed] = set.insert(acc);
     if (succeed)
         for (auto child = gnc_account_get_children(acc); child; child = c_list_next(child))
             maybe_add_descendants(set, static_cast<Account*>(child->data));
}

GList*
gnc_accounts_and_all_descendants(GList *accounts)
{
    AccountSet accset;
    for (GList *n = accounts; n; n = g_list_next(n))
         maybe_add_descendants(static_cast<Account*>(n->data));
   GList *ret = nullptr;
    std::for_each (accset.begin(), accset.end(),
                   [&ret](const auto& it) { ret = g_list_prepend (ret, it); });
    return g_list_sort (ret, (GCompareFunc) acct_fullname_cmp);
}
```

libgnucash/engine/Account.cpp
christopherlam: I think there's no need to utf8 sort this account list which generally is fed into QofQiery. This will effect a small change in tests.

jralls: Definitely. In half the calls the list is passed to `gnc:accounts-get-commodities` so it definitely doesn't need to be sorted. Then there's budget.scm and cash-flow.scm where the result is sorted again with `(sort accounts gnc:account-full-name<?))`.  It seems to get fed to a query only in examples/average-balance.scm and examples/daily-reports. scm. In category-barcharts.scm there's a call where its rv is passed to a lambda that creates an alist with the account as the key, you might want to sort that.

christopherlam: Ok I'll remove the expensive utf8 sorting. This is the only functional difference with the scheme version, and causes a small change in tests. If any reports eventually show accounts in a wrong order it'll be easy to reinstate sorting within the report itself.

christopherlam:
Changed to have both `gnc-accounts-and-all-descendants` and `gnc:accounts-and-all-descendants-sorted`, the latter in scheme only and offers the old utf8 sorting. This is to accommodate some reports which seem to break when offered an unsorted descendants list. Also cpp func is now terse.

libgnucash/engine/Account.cpp
```diff
@@ -6229,25 +6229,19 @@ gboolean xaccAccountRegister (void)
...
 {
     AccountSet accset;
-    for (GList *n = accounts; n; n = g_list_next (n))
-        maybe_add_descendants (accset, static_cast<Account*>(n->data));
-    GList *ret = nullptr;
-    std::for_each (accset.begin(), accset.end(),
-                   [&ret](const auto& it) { ret = g_list_prepend (ret, it); });
-    return ret;
+    g_list_foreach (accounts, (GFunc) maybe_add_descendants, &accset);
+    return std::accumulate (accset.begin(), accset.end(), (GList*) nullptr, g_list_prepend);
```
jralls, line 6244: Include what you use: Add `#include <algorithm>` at the top for `std::accumulate`.

jralls: Apparently you need a little background on Include What You Use. [Read this](https://www.fluentcpp.com/2021/01/01/include-what-you-use/).

PR #1312, opened 2022-04-08
gncEntry and invoice plugin page

christopherlam:
* removes rounding for Discount, Quantity allowing input of unrounded numbers
* invoice report shows numbers as fraction or decimal appropriately
* silence gnc-numeric warning


gnucash/register/ledger-core/gncEntryLedger.c
```diff
@@ -271,16 +271,6 @@ gnc_entry_ledger_config_cells (GncEntryLedger *ledger)
      gnc_table_layout_get_cell (ledger->table->layout, ENTRY_PRIC_CELL),
      GNC_COMMODITY_MAX_FRACTION);
```
jralls, line 273: Price should also be unrounded.

jralls:
The commit summary on 56a6f3c should be something like "Reduce logging level of gnc_numeric exception message" with an additional message "exceptions are expected in some use cases."

PR #1314, opened 2022-04-11
Bug 798504 - In the Generic import transaction matcher, prettyfy display of split transaction

christopherlam:
One approach is to show all splits from the matched transaction in the dropdown. The matched split is highlighted via '==>'.

image

### DIFF
--- a/gnucash/import-export/import-backend.c
+++ b/gnucash/import-export/import-backend.c
@@ -299,7 +299,6 @@ void gnc_import_TransInfo_delete (GNCImportTransInfo *info)
 GdkPixbuf* gen_probability_pixbuf(gint score_original, GNCImportSettings *settings, GtkWidget * widget)
 {
     GdkPixbuf* retval = NULL;
-    gint i, j;
     gint score;
     const gint height = 15;
     const gint width_each_bar = 7;
@@ -312,24 +311,20 @@ GdkPixbuf* gen_probability_pixbuf(gint score_original, GNCImportSettings *settin
     const gint num_colors = 5;
     gchar * size_str;
     gchar * none_color_str = g_strdup_printf("  c None");
-    gchar * green_color_str = g_strdup_printf("g c green");
-    gchar * yellow_color_str = g_strdup_printf("y c yellow");
-    gchar * red_color_str = g_strdup_printf("r c red");
-    gchar * black_color_str = g_strdup_printf("b c black");
+    gchar * green_color_str = g_strdup_printf("g c #2ECC40");
+    gchar * yellow_color_str = g_strdup_printf("y c #FFDC00");
+    gchar * red_color_str = g_strdup_printf("r c #FF4136");
+    gchar * black_color_str = g_strdup_printf("b c #111111");
     gchar * xpm[2+num_colors+height];
     gint add_threshold, clear_threshold;
 
     g_assert(settings);
     g_assert(widget);
-    if (score_original < 0)
-    {
-        score = 0;
-    }
-    else
-    {
-        score = score_original;
-    }
-    size_str = g_strdup_printf("%d%s%d%s%d%s", (width_each_bar * score) + width_first_bar/*width*/, " ", height, " ", num_colors, " 1"/*characters per pixel*/);
+
+    score = (score_original < 0) ? 0 : score_original;
+    size_str = g_strdup_printf("%d %d %d 1",
+                               (width_each_bar * score) + width_first_bar, /*width*/
+                               height, num_colors);
 
     /*DEBUG("Begin");*/
     xpm[0] = size_str;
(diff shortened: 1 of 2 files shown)


gnucash/import-export/import-main-matcher.c
```diff
@@ -1559,49 +1559,47 @@ get_required_color (const gchar *class_name)
...
+    GtkTreeStore *store = GTK_TREE_STORE (model);
+    GtkTreeIter child;
+    gtk_tree_store_append (store, &child, iter);
+    gtk_tree_store_set (store, &child, DOWNLOADED_COL_DATE_TXT, date, -1);
+    gtk_tree_store_set (store, &child, DOWNLOADED_COL_ACCOUNT, account, -1);
+    gtk_tree_store_set (store, &child, DOWNLOADED_COL_AMOUNT, amount, -1);
+    gtk_tree_store_set (store, &child, DOWNLOADED_COL_DESCRIPTION, narrative, -1);
+    gtk_tree_store_set (store, &child, DOWNLOADED_COL_ENABLE, FALSE, -1);
+}
+
```
Bob-IT, line 1582: You should combine the gtk_tree_store_set's to one but with my suggested changes this function is not required.

gnucash/import-export/import-main-matcher.c
```diff
@@ -1559,49 +1559,47 @@ get_required_color (const gchar *class_name)
...
+    for (GList *n = xaccTransGetSplitList (trans); n; n = n->next)
+    {
+        Split *s = n->data;
+        const gchar *acct_name = xaccAccountGetName (xaccSplitGetAccount (s));
+        const gchar *memo = xaccSplitGetMemo (s);
+        const gchar *amount = xaccPrintAmount (xaccSplitGetAmount (s),
+                                               gnc_split_amount_print_info (s, TRUE));
+        add_row (model, iter, (sel_match->split == s) ? "==>" : "",
+                 acct_name, amount, (memo && *memo) ? memo : "(no memo)");
+    }
```
Bob-IT, line 1602: I was not sure how to show my changes for this so added here...

    const Transaction *trans = sel_match->trans;
    const gchar *desc = xaccTransGetDescription (trans);
    gchar *datestr = qof_print_date (xaccTransGetDate (trans));

    remove_child_row (model, iter);

    for (GList *n = xaccTransGetSplitList (trans); n; n = n->next)
    {
        GtkTreeStore *store = GTK_TREE_STORE (model);
        GtkTreeIter child;
        Split *s = n->data;
        const gchar *acct_name = xaccAccountGetName (xaccSplitGetAccount (s));
        const gchar *memo = xaccSplitGetMemo (s);
        const gchar *amount = xaccPrintAmount (xaccSplitGetAmount (s),
                                               gnc_split_amount_print_info (s, TRUE));

        gtk_tree_store_append (store, &child, iter);
        gtk_tree_store_set (store, &child,
                            DOWNLOADED_COL_DATE_TXT, (sel_match->split == s) ? datestr : "",
                            DOWNLOADED_COL_ACCOUNT, acct_name,
                            DOWNLOADED_COL_AMOUNT, amount,
                            DOWNLOADED_COL_DESCRIPTION, (sel_match->split == s) ? desc : "",
                            DOWNLOADED_COL_MEMO, (memo && *memo) ? memo : "(no memo)",
                            DOWNLOADED_COL_ENABLE, FALSE, -1);
    }
    g_free (datestr);

You will see that I have added info to the memo column and also populated the existing transaction description and removed the extra date row as the date is now added to the split row.
Try it and see if it works for you.

Bob-IT:
As an after thought, with these changes is there a need for the split account names to be added to the 'Additional comments', if its just one split possibly but for multiple splits may be we can just say 'Use Show matched information' some how.
I do not use this so have no idea.

gnucash/import-export/import-main-matcher.c
christopherlam: The main reason for the separate function is that I've moved the existing splits' memo into the Description header, similar to the regular register. Otherwise the memos are squashed into a separate column while the description column is empty.

Bob-IT: I see what you mean, then maybe there should only be one column, combine Description and Memo and change column header appropriately on row selection. May work?
I still would get rid of the extra line though.

christopherlam: :+1: for combining Desc and Memo. They could even be combined via `\n` in the imported splits. But it needs to be tested for UX.

The extraline would be needed for the matched transaction - txn->description in 1 line, and 1 line per split->memo.

jralls:
> As an after thought, with these changes is there a need for the split account names to be added to the 'Additional comments', if its just one split possibly but for multiple splits may be we can just say 'Use Show matched information' some how.

No need, there's an account column and @christopherlam has already populated it with the splits' accounts.

gnucash/import-export/import-backend.c
```diff
@@ -312,24 +311,20 @@ GdkPixbuf* gen_probability_pixbuf(gint score_original, GNCImportSettings *settin
...
     gchar * size_str;
     gchar * none_color_str = g_strdup_printf("  c None");
-    gchar * green_color_str = g_strdup_printf("g c green");
-    gchar * yellow_color_str = g_strdup_printf("y c yellow");
-    gchar * red_color_str = g_strdup_printf("r c red");
-    gchar * black_color_str = g_strdup_printf("b c black");
+    gchar * green_color_str = g_strdup_printf("g c #2ECC40");
+    gchar * yellow_color_str = g_strdup_printf("y c #FFDC00");
+    gchar * red_color_str = g_strdup_printf("r c #FF4136");
+    gchar * black_color_str = g_strdup_printf("b c #111111");
```
fellen, line 317: I would prefer to use named constants—see https://www.w3schools.com/colors/colors_hex.asp— or something defined in our css files.

PR #1317, opened 2022-04-13
[Master] C++options posible fixes

Bob-IT:
@jralls  I have been looking at the new C++ options in master and have tried to fix a couple of additional errors I found so created this PR. Apart from the 'FancyDate' one, there is one I have come across that can not see where it is going wrong.

That is 'Plot Size', first off I did not realize it had changed so looking in wrong area. In the dialog, the plot size widget is 309 characters wide. I see that gnc_make_plot_size_option is getting the correct 'int' values but when it tries to create a GncOptionUIType::NUMBER_RANGE calling create_range_spinner,  the values returned do not match with the upper_bound being the very large number.

### DIFF
--- a/libgnucash/app-utils/gnc-option-impl.cpp
+++ b/libgnucash/app-utils/gnc-option-impl.cpp
@@ -587,7 +587,7 @@ qof_instance_from_guid(GncGUID* guid, GncOptionUIType type)
             qof_type = "gncInvoice";
             break;
         case GncOptionUIType::TAX_TABLE:
-            qof_type = "gncTaxtable";
+            qof_type = "gncTaxTable";
             break;
         case GncOptionUIType::ACCOUNT_LIST:
         case GncOptionUIType::ACCOUNT_SEL:
--- a/libgnucash/app-utils/gnc-option-impl.cpp
+++ b/libgnucash/app-utils/gnc-option-impl.cpp
@@ -224,7 +224,7 @@ GncOptionCommodityValue::reset_default_value()
 bool
 GncOptionCommodityValue::is_changed() const noexcept
 {
-    return m_namespace == m_default_namespace && m_mnemonic == m_default_mnemonic;
+    return m_namespace != m_default_namespace || m_mnemonic != m_default_mnemonic;
 }
 
 bool
--- a/gnucash/gnome-utils/dialog-options.cpp
+++ b/gnucash/gnome-utils/dialog-options.cpp
@@ -1319,6 +1319,11 @@ void
 BothDateEntry::toggle_relative(bool use_absolute)
 {
     m_use_absolute = use_absolute;
+
+    gtk_widget_set_sensitive(GTK_WIDGET(m_abs_entry->get_widget()),
+                             m_use_absolute);
+    gtk_widget_set_sensitive(GTK_WIDGET(m_rel_entry->get_widget()),
+                             !m_use_absolute);
 }
 
 void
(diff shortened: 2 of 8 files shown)


jralls:
@Bob-IT, Sorry, I should have acknowledged your comments on the commits sooner. I'm busy with cairo ATM and after that on the queue is working up a best and final MR for [Gtk Issue 3714](https://gitlab.gnome.org/GNOME/gtk/-/issues/3714). It may be 3-4 weeks before I can get back to this.

fellen:
We should merge maint for the docker fix, but there are conflicts which I do not fully understand.

jralls:
@fellen, I just merged and pushed. We'll get conflicts on options.scm every time it has changes; those changes should always be discarded. In this case they were just spelling changes in comments so no problem.

Bob-IT:
@jralls Hopefully these changes will be appropriate or at least give you a clue to a location for the correct fix. 
From what I can tell that only leaves the 'Fancy Date' to fix but I probably have not tested every thing. There is this error that comes up on options exit,` [gnc_option_db_destroy()] Direct Destroy called on GncOptionDB 94131654261200` and not sure if all the old plot parts are required now.
Will await your comments when you have time.

Bob-IT:
@jralls I will leave this for now, the last commit is probably totally wrong to do with the currency and commodity combo's.
There is one other thing I have noticed but have no idea on, that is to do with using the 'Head or Tail' style sheet. You can set it and the report looks OK but if you try and save the config GnuCash crashes or if the report is left open it crashes while loading.

jralls:
>  if you try and save the config GnuCash crashes or if the report is left open it crashes while loading.

That's a bit weird. It's surely a serialization problem, but it's strange that saving a config crashes while serializing but leaving the tab open serializes without crashing only to crash when deserializing.

jralls:
I figured out the Head or Tail stylesheet bug. c++options is saving the stylesheet option as
```
(let ((option (gnc:lookup-option options                                  "General"                                  "Stylesheet")))   ((lambda (o) (if o (gnc:option-set-value o 'Head or Tail))) option))
```
and maint saves it as
```
(let ((option (gnc:lookup-option options                                  "General"                                  "Stylesheet")))   ((lambda (o) (if o (gnc:option-set-value o '#{Head or Tail}#))) option))
```
The other stylesheets don't have that escape mechanism because they don't have spaces.

jralls:
More or less the same thing for the commodity crash. c++options is saving
```
(gnc:option-set-value o "NASDAQ" "AAPL")
```
and maint saves
```
 (gnc:option-set-value o '(commodity-scm "NASDAQ" "AAPL"))
```

libgnucash/app-utils/gnc-option-impl.cpp
```diff
@@ -224,7 +224,7 @@ GncOptionCommodityValue::reset_default_value()
 bool
 GncOptionCommodityValue::is_changed() const noexcept
 {
-    return m_namespace == m_default_namespace && m_mnemonic == m_default_mnemonic;
+    return m_namespace == m_default_namespace && m_mnemonic != m_default_mnemonic;
```
jralls, line 227: This should be either
```
return !(m_namespace == m_default_namespace && m_mnemonic == m_default_mnemonic);
```
or
```
return m_namespace != m_default_namespace || m_mnemonic != m_default_mnemonic;
```
To cover the possibility that two namespaces have the same mnemonic.

jralls:
I've pushed a fix for the stylesheet and commodity option crashes onto master.

For the rest, aside from the `is_changed` logic I think your changes all look reasonable. Go ahead and merge when you like.

jralls:
And thanks for the thorough testing!

PR #1319, opened 2022-04-15
[account.cpp] restore breadthwise search for 2 functions

christopherlam:
gnc_account_lookup_by_name and gnc_account_lookup_by_code were searching breadthwise.

as reported in https://github.com/Gnucash/gnucash/pull/1101#issuecomment-1098146573

### DIFF
--- a/libgnucash/engine/Account.cpp
+++ b/libgnucash/engine/Account.cpp
@@ -3048,6 +3048,30 @@ gnc_account_get_descendants_sorted (const Account *account)
     return g_list_reverse (list);
 }
 
+// because gnc_account_lookup_by_name and gnc_account_lookup_by_code
+// are described in Account.h searching breadth-first until 4.6, and
+// accidentally modified to search depth-first from 4.7
+// onwards. Restore breath-first searching in 4.11 onwards to match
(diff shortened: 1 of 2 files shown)


jralls:
Since the original behavior was for `account_foreach_descendant` to search by breadth perhaps it would be better to restore that behavior and rename the by-depth function `account_foreach_descendant_by_depth`, and call it only in those (new-ish) functions where that makes more sense, perhaps with comments explaining why.

jralls (commented):
The code looks fine.

christopherlam:
> Since the original behavior was for `account_foreach_descendant` to search by breadth perhaps it would be better to restore that behavior and rename the by-depth function `account_foreach_descendant_by_depth`, and call it only in those (new-ish) functions where that makes more sense, perhaps with comments explaining why.

I don't think this is accurate -- `foreach_descendant` and `foreach_descendant_until` were always depth-first.

jralls:
> I don't think this is accurate -- foreach_descendant and foreach_descendant_until were always depth-first.

Ah, got it: The problem was from too-aggressive code consolidation.

OK, merge when ready.

PR #1321, opened 2022-04-17
offer filtering for GNCAccountSel - hide placeholder and hidden accounts.

christopherlam:
for use in dialogs whereby account-sel is meant for posting new transactions.... feedback on var and api names ?

also useful for https://bugs.gnucash.org/show_bug.cgi?id=796481

### DIFF
--- a/gnucash/gnome/gnc-plugin-page-account-tree.c
+++ b/gnucash/gnome/gnc-plugin-page-account-tree.c
@@ -1359,6 +1359,7 @@ gppat_populate_gas_list(GtkWidget *dialog,
 
     /* Setting the account type filter triggers GNCAccountSel population. */
     gnc_account_sel_set_acct_filters (gas, filter, NULL);
+    gnc_account_sel_filter_placeholder (gas, TRUE);
 
     /* Accounts to be deleted must be removed. */
     gnc_account_sel_purge_account( gas, account, exclude_subaccounts);
--- a/gnucash/import-export/csv-imp/assistant-csv-trans-import.cpp
+++ b/gnucash/import-export/csv-imp/assistant-csv-trans-import.cpp
@@ -572,6 +572,7 @@ CsvImpTransAssist::CsvImpTransAssist ()
         acct_selector = gnc_account_sel_new();
         auto account_hbox = GTK_WIDGET(gtk_builder_get_object (builder, "account_hbox"));
         gtk_box_pack_start (GTK_BOX(account_hbox), acct_selector, TRUE, TRUE, 6);
+        gnc_account_sel_filter_placeholder (GNC_ACCOUNT_SEL(acct_selector), true);
         gnc_account_sel_set_hexpand (GNC_ACCOUNT_SEL(acct_selector), true);
         gtk_widget_show (acct_selector);
 
--- a/gnucash/gnome-utils/dialog-book-close.c
+++ b/gnucash/gnome-utils/dialog-book-close.c
@@ -352,6 +352,7 @@ void gnc_ui_close_book (QofBook* book, GtkWindow *parent)
     box = GTK_WIDGET(gtk_builder_get_object (builder, "income_acct_box"));
     cbw->income_acct_widget = gnc_account_sel_new();
     gnc_account_sel_set_hexpand (GNC_ACCOUNT_SEL(cbw->income_acct_widget), TRUE);
+    gnc_account_sel_filter_placeholder (GNC_ACCOUNT_SEL(cbw->income_acct_widget), TRUE);
     gnc_account_sel_set_acct_filters(GNC_ACCOUNT_SEL(cbw->income_acct_widget),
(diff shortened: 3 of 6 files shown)


Bob-IT (commented):
@christopherlam It looks OK to me but am not sure about 'writable' but can not think of another word, the only other thing is split it to two functions.
gnc_account_sel_filter_placeholder_accts (GNCAccountSel *gas, gboolean filter);
gnc_account_sel_filter_hidden_accts (GNCAccountSel *gas, gboolean filter);

christopherlam:
> @christopherlam It looks OK to me but am not sure about 'writable' but can not think of another word, the only other thing is split it to two functions. `gnc_account_sel_filter_placeholder_accts (GNCAccountSel *gas, gboolean filter); gnc_account_sel_filter_hidden_accts (GNCAccountSel *gas, gboolean filter);` but maybe it could be `gnc_account_sel_filter_read_only_accts (GNCAccountSel *gas, gboolean filter);`

The (small) advantage of defining api names according to purpose is it cane be amended in case the schema changes.

jralls:
> It looks OK to me but am not sure about 'writable' but can not think of another word, the only other thing is split it to two functions

`writable` isn't even true: Hidden accounts are writable.

Placeholder exclusion is implemented in the register via a `skip_cb` parameter to `gnc_get_shared_account_name_list_store` and `gnc_get_shared_account_name_quickfill`. ISTM we should use one of those functions (in account-quickfill.c) for comboxes and reserve gnc-account-sel for tree views where the placeholders need to be visible. Where it isn't already "Show Hidden" should be a checkbox selection on account selection treeviews. I think a lot of users would be happy to have an *optional* removal of hidden accounts from the account comboboxes too, though @gjanssens may object to another preference and I suppose it's also possible that users would want hidden accounts to appear in some places like the Close Book dialog and not others like a split register account field or the Stock Transaction Assistant.

gnucash/gnome/gnc-plugin-page-account-tree.c
```diff
@@ -1359,6 +1359,7 @@ gppat_populate_gas_list(GtkWidget *dialog,
 
     /* Setting the account type filter triggers GNCAccountSel population. */
     gnc_account_sel_set_acct_filters (gas, filter, NULL);
+    gnc_account_sel_filter_placeholder (gas, TRUE);
```
jralls, line 1362: Not here, that will break tree views.

gnucash/import-export/csv-imp/assistant-csv-trans-import.cpp
```diff
@@ -572,6 +572,7 @@ CsvImpTransAssist::CsvImpTransAssist ()
         acct_selector = gnc_account_sel_new();
         auto account_hbox = GTK_WIDGET(gtk_builder_get_object (builder, "account_hbox"));
         gtk_box_pack_start (GTK_BOX(account_hbox), acct_selector, TRUE, TRUE, 6);
+        gnc_account_sel_filter_placeholder (GNC_ACCOUNT_SEL(acct_selector), true);
```
jralls, line 575: Isn't this a tree view?

christopherlam:
These changes to CSV and plugin-page-account-tree do not affect treeviews; they affect simple GtkComboBox whereby all accounts are shown as full account names.

Bob-IT:
I have looked into this and been trying out the use of the account name quick fill.
`The existing gnc-account-sel is a combobox with a list store of account full names`, the list store being in the same format as the account name one. With such a change, from my tests all gnc-account-sel widgets would share the same QFKEY and as such the same list store and skip_cb so the first one loaded would govern the rest (there is also a note about it not listening for account destroy events which I did not check). For the existing filtering of account types, I suppose a model filter could be added on top so individual gnc-account-sel's could be tailored. I am not sure this is worth doing.

There was another bug, [[Bug 798475](https://bugs.gnucash.org/show_bug.cgi?id=798475)] that was about adding the same keyboard shortcuts like Ex: for 'Expenses:' that I have almost got working and while doing that was wondering if adding a secondary icon to the entry that would popup a two item toggle menu to include/exclude hidden/placeholder accounts, the default being both are hidden would work.

gnucash/import-export/csv-imp/assistant-csv-trans-import.cpp
jralls: Nope, it's a combo on the CSV characteristic page.

jralls:
> These changes to CSV and plugin-page-account-tree do not affect treeviews; they affect simple GtkComboBox whereby all accounts are shown as full account names.

Yes, sorry. I got lost in the spaghetti.

FWIW, the "simple GtkComboBox" died with Gtk2 (of course that's what all of this code was written for). In Gtk3 and Gtk4 a GtkComboBox is a special case of GtkTreeView.

christopherlam:
much better in #1332

PR #1324, opened 2022-04-22
Bug 797669 - Change Schedule C Returns and Allowances TXF Category

fellen:
to Income Account

@jralls I can't find Alex' nick here. But perhaps you can check this, too?


jralls:
@fellen I don't think Alex is on Github. This looks right on its face, but I have no more (and possibly less) knowledge of the TXF code than you do so no confidence that it's right.

jralls:
[Turns out](https://bugs.gnucash.org/show_bug.cgi?id=797669) that Alex has [already fixed it](https://github.com/Gnucash/gnucash/search?q=d48937c&type=commits).

PR #1326, opened 2022-04-29
Bug 798501 - Balance wrong date end of account period

Bob-IT:
With the accounting period preference settings set to 'Start/End of previous quarter' and balance sheet report using 'End of accounting period' the date is wrong, currently is 30/03/2022 when it should be 31/03/2022.

I fixed this by changing `gnc_gdate_set_prev_quarter_end` to make sure date is month end.
I think it looks correct/OK to me but after a second opinion as I am not really up on financial matters.

For the way the date was calculated... 
`gnc_gdate_set_prev_quarter_end` calls `gnc_gdate_set_quarter_end` and this function changes the date in the following way...
starts of as 01/04/2022
changes to 01/07/2022
changes to 30/06/2022
and then `gnc_gdate_set_prev_quarter_end` would subtract 3 months to get
30/03/2022

### DIFF
--- a/libgnucash/engine/gnc-date.h
+++ b/libgnucash/engine/gnc-date.h
@@ -702,7 +702,7 @@ void gnc_gdate_set_prev_month_end (GDate *date);
 
 /** This function modifies a GDate to set it to the first day of the
  *  quarter in which it falls.  For example, if this function is called
- *  with a date of 2003-09-24 the date will be modified to 2003-09-01.
+ *  with a date of 2003-09-24 the date will be modified to 2003-07-01.
  *
  *  @param date The GDate to modify. */
 void gnc_gdate_set_quarter_start (GDate *date);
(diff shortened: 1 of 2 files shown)


Bob-IT:
@jralls  Changed the function to use your code but needed to subtract 1 from the month otherwise December would result in quarter being 4. Also needed to subtract the 3 months before calling this modified function else we would end up in the same situation as before with returned date being 30/03/2022 and corrected some dates mentioned in the text describing functions.

libgnucash/engine/gnc-date.h
```diff
@@ -702,15 +702,15 @@ void gnc_gdate_set_prev_month_end (GDate *date);
 
 /** This function modifies a GDate to set it to the first day of the
  *  quarter in which it falls.  For example, if this function is called
- *  with a date of 2003-09-24 the date will be modified to 2003-09-01.
+ *  with a date of 2003-09-24 the date will be modified to 2003-07-01.
```
jralls, line 705: Huh? No quarter starts on 1 July. The original comment was correct.

jralls:
> needed to subtract 1 from the month

Right, G_DATE_JANUARY is 1, not 0 like in struct tm.

>  Also needed to subtract the 3 months before calling this modified function else we would end up in the same situation as before with returned date being 30/03/2022

Of course.

PR #1328, opened 2022-05-02
Add warning if adding stock transaction dated before the latest existing stock account transaction

christopherlam:
Further progress. Can show non-critical warnings as well as errors.

image

### DIFF
--- a/gnucash/gnome/assistant-stock-transaction.cpp
+++ b/gnucash/gnome/assistant-stock-transaction.cpp
@@ -58,8 +58,8 @@ static const char* ASSISTANT_STOCK_TRANSACTION_CM_CLASS = "assistant-stock-trans
 enum assistant_pages
 {
     PAGE_INTRO = 0,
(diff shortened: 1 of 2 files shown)


christopherlam:
@Bob-IT the middle GtkTreeView inside GtkScrolledWindow is too short. How to expand the height vertically?

Bob-IT:
@christopherlam  Would setting the scroll window vertical scroll bar to never work for you?
For this warning, you do not mention the actual date so how do I know if it is wrong?
I have not really looked at this, but did notice some layout things I would change but that will have to wait.

christopherlam:
> @christopherlam Would setting the scroll window vertical scroll bar to never work for you?

As usual, it does work!

> For this warning, you do not mention the actual date so how do I know if it is wrong?

Modified: 
image

> I have not really looked at this, but did notice some layout things I would change but that will have to wait.

Ok. Note also some fixmes in the code, mainly setting default focus in various pages don't seem to work.

PR #1329, opened 2022-05-06
stock assistant

christopherlam:
- guards against unassigned std::optional
- ~skip check txn_date < last txn date if the stock acct is empty~
- ~UX improvement, don't reset txn_types unnecessarily~

### DIFF
--- a/gnucash/gnome/assistant-stock-transaction.cpp
+++ b/gnucash/gnome/assistant-stock-transaction.cpp
@@ -29,6 +29,7 @@
 #include <numeric>
 #include <algorithm>
 #include <optional>
+#include <stdexcept>
 
 extern "C" {
 #include "Transaction.h"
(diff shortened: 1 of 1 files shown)


gnucash/gnome/assistant-stock-transaction.cpp
```diff
@@ -514,7 +515,17 @@ refresh_page_transaction_type (GtkWidget *widget, gpointer user_data)
     if (!info->txn_types)
         return;
 
-    info->txn_type = info->txn_types->at (type_idx);
+    try
+    {
+        info->txn_type = info->txn_types->at (type_idx);
+    }
+    catch (const std::out_of_range& e)
```
bhardwajs, line 522: The exception is not used in this block and would lead to a warning about unused variable in `-Wall`. How about `catch(const std::out_of_range&)`?

PR #1330, opened 2022-05-06
Force Configured Company Name option in Balance Sheet Report

debian-testing:
The current "Balance Sheet" report seems to read the 'Company Name" at first use, but if this report option has text in it (many or automatic), it's next use no longer takes the configured company name. This means that if a user configures a Balance Sheet report saves it, it's use on a separate company will not get updated. Also, I think that if GnuCash is not closed in between companies, the report continues to use the previous company name, even if the report option is force to blank.

Either way, the company name in the report has caused me alot of wasted time and confusion. However, this patch leaves the existing option as-is to avoid breaking gnucash for other current users. Long term, especially for Balance Sheet, I don't see why the company name in the report would be different than the configured company name, so many eventually remove this option? Or have a better way to deal with the specific needs (multi-company, multi-division)?


jralls:
The current design is that the option is populated with the book's Company Name value at report creation. If you override it then it's gone: A blank value in the Company Name option means no company name in the report title. That persists: GnuCash considers the option changed if the option was populated from the book properties at creation and removed using report options. Retyping the same value as the book property does remove the persisted option.

I think that a force-book-property checkbox should disable the text option field in the UI.
We need to make sure that this won't crash if a saved report is opened in an older GnuCash release.

@christopherlam Any thoughts?

gjanssens:
IMHO I agree with @debian-testing in that the `Company Name` option in the report options is superfluous and I would suggest to remove it. The only two scenarios I would support are:

* show the company name and use the one defined in the book properties in that case
* don't show the company name

The extra option to add (Show/hide company name) would then fit under the Display tab.

I am aware the risk is some users did (ab)use the company name option in unintended ways and could complain. I think however it's more straight forward design to have a book level company name and and option to show or hide it on the report than it is to have two company name options and the need for an extra option to force use of one of the two.

As for potential crashes - IIUC unknown options only generate a warning in the logs and are otherwise ignored.

fellen:
I agree with Geert. I suspect when the book option was introduced some reports were not adjusted.

debian-testing:
Thank you for the input.  The larger context here is that a few related issues with standard reports was forcing me to have copies of the reports for each book/company report and year.  These issues can be summarized as:  
-Company Name
-Select All Accounts
-Previous-Previous-Year
-Report Name (minor issue)

Standard reports like Balance Sheet, Income Statement and Full General Ledger should just work simple.  Anyone using an accounting system expects that.  I am willing to help as much as possible, this is why I submitted these PR, but feel free to rework them or provide other solutions.  Depending on the progress and direction of the "Company Name" solution, I will continue the other issues.

The "Select All" issue is that once a report is saved, the list of accounts is saved with it.  If the accounts change or the report is used on a different company with different account, the report either crashes or has incorrect results.  There should be a "Select All" checkbox that dynamically forces all relevant account (like pressing the "Select All" button) each time the report is run.  I don't really see why there is an option to select the accounts, since any missing accounts will cause an incorrect Balance Sheet and Income Statement and completely defeats the purpose of having a double-entry accounting system.  But I don't want to impose my perspective of changes for GnuCash, so at least there should be an option to force all account every time the report is run.  I can open another PR, Bug, or discussion elsewhere, as the team prefers.

christopherlam:
I agree there's major inconsistency in how the report title is set. Business reports tend to use it to populate the 'myname' section, and some reports use `(string-append companyname " " report-title " " ...)` format. My preference would be to ensure *all* reports prepend the company name if non-null rather than add an option.

jralls:
@debian-testing It looks like @gjanssens and @fellen want the company-name option removed from all reports and that rather than a `force-book-option` checkbox a `don't-use-company-name-in-title` one. Are you willing/able to code that?

debian-testing:
As requested, here is a version that removed the company name report field and always uses the book company name. I will send a separate PR per report. I will provide the most common reports first, since I use this option myself, thus it is unit tested.

I noted in the code, that the "get company-name should probably be modularized and standardized across reports". For now, I used the most common code pattern currently in use across the existing reports.

jralls:
> [Merge branch 'Gnucash:maint' into balance-sheet-force-party-name-book](https://github.com/Gnucash/gnucash/pull/1330/commits/91a99cb744681833e3c4adc440ebcab1a4c311ce)

Do not ever do that on a pull request branch, it makes commit loops in history. Always rebase your pull request on the target branch. Please do that now (rebasing will get rid of the merge commit) and force-push the result.

gnucash/report/reports/standard/balance-sheet.scm
```diff
@@ -290,8 +279,10 @@
 
   ;; get all option's values
   (let* (
+	 (book (gnc-get-current-book))
+	 ;; Get company-name should probably be modularized and standardized across reports
+	 (company-name (gnc:company-info book gnc:*company-name*))
```
jralls, line 284: I think the test failures can be fixed by replacing this line with 
```
(or (company-name (gnc:company-info book gnc:*company-name*)) "")
```

jralls:
> I will send a separate PR per report. 

How about batching several reports per PR instead? You can still do a commit per report if you want.

Please remember run `ninja check` on your local build before committing changes.

jralls:
> How about batching several reports per PR instead? You can still do a commit per report if you want.

On second thought, stop now. The option system is rewritten in master and this change will make merge conflicts with parked #1402. It's simpler to do this as part of that PR so it's introduced in GnuCash 5.

jralls:
> @debian-testing
[Merge branch 'Gnucash:maint' into balance-sheet-force-party-name-book](https://github.com/Gnucash/gnucash/pull/1330/commits/95c8b733af66dc6c23fc84fcc4481a3644e86d1e)

Not very good at following instructions, I see.

debian-testing:
Sorry, having trouble with the github webui.  The code adjustments didn't take long, but trying to do the GitHub PR cleanly is taking alot of time.   The reccomendations are good, both to merge the PRs and for the code adjustment to deal with empty company name.  I have been running these corrected version for a while, but was avoiding submitting a PR because it takes so much time and energy.  The purpose of submitting and doing in 2 separate PRs is for transparency and reusability.  

I also want to change/correct the General Ledger which is based on the transaction reporting engine, but that will be for another time.  So the purpose here was to contribute iteratively.  Feel free to do what you want with these two PRs.  I will try to fix within Github, but as you can see, it's making more of a mess....

jralls:
@debian-testing I'll take care of it for all reports on master and the changes will be in GnuCash 5 that will [start testing releases in January for initial release in March](https://wiki.gnucash.org/wiki/Release_Schedule). You don't need to do anything more.

PR #1333, opened 2022-05-12
[Master] Bug753307 - Custom Report be selectable as default Report for Printing

Bob-IT:
This PR is a rehash of the maint version PR #1094 and now uses the new options as part of saving in book properties.

This commit changes the preference in Business->'Report for Printing' to be saved as a book property in 'File>Properties->Business' and allow the selection of any Invoice Report to be used as the default. Every thing in the previous PR is the same apart from the saving to book properties and the inclusion of code to do that. The property being saved is a joining of the report guid and the report name which is used if said report is deleted.

Also included is code which I think achieves what PR #1247 does. This was done by presenting a dialog with the default for the report combo set to the properties setting. This dialog has a timeout which is adjustable under properties, (setting of 0 disables this) and will stop if a key is pressed or combo popped so that a different report template can be selected and used. There is also an 'OK' button that stops the time out and prints and a 'Cancel' button which cancels the print.

All seems to work as before with the last commit removing the preference setting and setting up the removal of the gschema setting in future versions.

### DIFF
--- a/gnucash/gnome-utils/gnc-report-combo.c
+++ b/gnucash/gnome-utils/gnc-report-combo.c
@@ -1,0 +1,551 @@
+/********************************************************************\
+ * gnc-report-combo.c -- report select widget for GnuCash           *
+ *                                                                  *
+ * Copyright (C) 2022 Bob Fewell                                    *
+ *                                                                  *
+ * This program is free software; you can redistribute it and/or    *
+ * modify it under the terms of the GNU General Public License as   *
+ * published by the Free Software Foundation; either version 2 of   *
+ * the License, or (at your option) any later version.              *
+ *                                                                  *
+ * This program is distributed in the hope that it will be useful,  *
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of   *
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the    *
+ * GNU General Public License for more details.                     *
+ *                                                                  *
+ * You should have received a copy of the GNU General Public License*
+ * along with this program; if not, contact:                        *
+ *                                                                  *
+ * Free Software Foundation           Voice:  +1-617-542-5942       *
+ * 51 Franklin Street, Fifth Floor    Fax:    +1-617-542-2652       *
+ * Boston, MA  02110-1301,  USA       gnu@gnu.org                   *
+ *                                                                  *
+\********************************************************************/
+
+#include <config.h>
+#include <gtk/gtk.h>
+#include <glib/gi18n.h>
+
+#include "gnc-report-combo.h"
+#include "gnc-ui-util.h"
+#include "gnc-engine.h"
+#include "dialog-utils.h"
+
+/** The debugging module used by this file. */
+static QofLogModule log_module = GNC_MOD_GUI;
+
+static void gnc_report_combo_init       (GncReportCombo      *grc);
+static void gnc_report_combo_class_init (GncReportComboClass *klass);
+static void gnc_report_combo_dispose    (GObject *object);
+static void gnc_report_combo_finalize   (GObject *object);
+
+#define GNC_REPORT_COMBO_PATH "gnc-report-combo-path"
+
+enum
+{
+    RC_NAME,
+    RC_GUID,
+    RC_MISSING
+};
+
+/** The instance private data for a content plugin. */
+typedef struct _GncReportComboPrivate
+{
+    GtkWidget   *combo;
+    GtkWidget   *warning_image;
+
+    const gchar *rpt_guids;
+
+    gboolean     block_signal;
+    gboolean     popup_shown;
+
+    gchar       *active_report_guid;
+    gchar       *active_report_name;
+
+} GncReportComboPrivate;
+
+G_DEFINE_TYPE_WITH_PRIVATE(GncReportCombo, gnc_report_combo, GTK_TYPE_BOX)
+
+#define GET_PRIVATE(o)  \
+   ((GncReportComboPrivate*)g_type_instance_get_private ((GTypeInstance*)o, GNC_TYPE_REPORT_COMBO))
+
+enum
+{
+    SIGNAL_0,
+    CHANGED,
+    LAST_SIGNAL
+};
+
+static guint report_combo_signals [LAST_SIGNAL] = {0};
+
+enum
+{
+    PROP_0,
+    PROP_POPUP_SHOWN,
+    N_PROPERTIES
+};
+
+static GParamSpec *report_combo_properties [N_PROPERTIES] = {NULL,};
+
+static void
+gnc_report_combo_get_property (GObject    *object,
+                               guint       property_id,
+                               GValue     *value,
+                               GParamSpec *pspec)
+{
+    GncReportCombo        *grc = GNC_REPORT_COMBO(object);
+    GncReportComboPrivate *priv = GET_PRIVATE(grc);
+
+    switch (property_id)
+    {
+    case PROP_POPUP_SHOWN:
+        g_value_set_boolean (value, priv->popup_shown);
+        break;
+
+    default:
+        /* We don't have any other property... */
+        G_OBJECT_WARN_INVALID_PROPERTY_ID (object, property_id, pspec);
+        break;
+    }
+}
+
+/** Initialize the GncReportCombo class object.
+ *
+ *  @internal
+ *
+ *  @param klass A pointer to the newly created class object.
+ */
+static void
+gnc_report_combo_class_init (GncReportComboClass *klass)
+{
+    GObjectClass   *object_class = G_OBJECT_CLASS(klass);
+    GtkWidgetClass *widget_class = GTK_WIDGET_CLASS(klass);
+
+    object_class->get_property = gnc_report_combo_get_property;
+    object_class->dispose  = gnc_report_combo_dispose;
+    object_class->finalize = gnc_report_combo_finalize;
+
+    report_combo_signals [CHANGED] =
+        g_signal_new ("changed",
+                      G_OBJECT_CLASS_TYPE(object_class),
+                      G_SIGNAL_RUN_FIRST,
+                      G_STRUCT_OFFSET(GncReportComboClass, changed),
+                      NULL,
+                      NULL,
+                      g_cclosure_marshal_VOID__VOID,
+                      G_TYPE_NONE,
+                      0);
+
+    report_combo_properties [PROP_POPUP_SHOWN] =
+        g_param_spec_boolean ("popup-shown",
+                              "State of PopUp",
+                              "State of PopUp",
+                              FALSE /* default value */,
+                              G_PARAM_READABLE);
+
+    g_object_class_install_properties (object_class,
+                                       N_PROPERTIES,
+                                       report_combo_properties);
+}
+
+/** Initialize a GncReportCombo object.
+ *
+ *  @internal
+ *
+ *  @param grc A pointer to the newly created object.
+ */
+static void
+gnc_report_combo_init (GncReportCombo *grc)
+{
+    GncReportComboPrivate *priv;
+
+    g_return_if_fail (grc != NULL);
+    g_return_if_fail (GNC_IS_REPORT_COMBO(grc));
+
+    priv = GET_PRIVATE(grc);
+
+    // Set the name for this widget so it can be easily manipulated with css
+    gtk_widget_set_name (GTK_WIDGET(grc), "gnc-id-report-combo");
+
+    priv->block_signal = FALSE;
+    priv->active_report_guid = NULL;
+    priv->active_report_name = NULL;
+    priv->popup_shown = FALSE;
+}
+
+/** Dispopse the GncReportCombo object. This function is called from
+ *  the G_Object level to complete the destruction of the object.  It
+ *  should release any memory not previously released by the destroy
+ *  function (i.e. the private data structure), then chain up to the
+ *  parent's destroy function.
+ *
+ *  @param object The object being destroyed.
+ *
+ *  @internal
+ */
+static void
+gnc_report_combo_dispose (GObject *object)
+{
+    /* Do not free the private data structure itself. It is part of
+     * a larger memory block allocated by the type system. */
+
+    G_OBJECT_CLASS (gnc_report_combo_parent_class)->dispose (object);
+}
+
+/** Finalize the GncReportCombo object.  This function is called from
+ *  the G_Object level to complete the destruction of the object.  It
+ *  should release any memory not previously released by the destroy
+ *  function (i.e. the private data structure), then chain up to the
+ *  parent's finalize function.
+ *
+ *  @param object The object being finalized.
+ *
+ *  @internal
+ */
+static void
+gnc_report_combo_finalize (GObject *object)
+{
+    GncReportComboPrivate *priv;
+    GncReportCombo *grc;
+
+    g_return_if_fail (object != NULL);
+    g_return_if_fail (GNC_IS_REPORT_COMBO(object));
+
+    grc = GNC_REPORT_COMBO(object);
+    priv = GET_PRIVATE(grc);
+
+    g_free (priv->active_report_guid);
+    g_free (priv->active_report_name);
+
+    G_OBJECT_CLASS (gnc_report_combo_parent_class)->finalize (object);
+}
+
+/** This function sets the active combo entry based on the private
+ *  report guid and also checks to see if the report guid is in the
+ *  list of reports, if not a warning image and tooltip is shown.
+ *
+ *  @internal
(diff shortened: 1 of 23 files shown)


libgnucash/app-utils/gnc-optiondb.cpp
```diff
@@ -1194,13 +1204,20 @@ gnc_option_db_book_options(GncOptionDB* odb)
     gnc_register_string_option(odb, business_section, N_("Company ID"), "c5",
                                N_("The ID for your company (eg 'Tax-ID: 00-000000)."),
                                empty_string);
-
+    gnc_register_invoice_print_report_option(odb, business_section,
+                                 OPTION_NAME_DEFAULT_INVOICE_REPORT, "e1",
+                                 N_("The invoice report to be used for printing."),
+                                 empty_string);
```
jralls, line 1210: That last value is the default value. You want it to be the guid-string for the Printable Invoice Report, not empty. Ensuring that there's a correct default value means that you don't have to handle setting a default in code anywhere.

gnucash/gnome-utils/gnc-report-combo.c
```diff
@@ -0,0 +1,595 @@
...
+static void
+update_report_list (GncReportCombo *grc)
+{
+    GncReportComboPrivate *priv = GET_PRIVATE(grc);
+    GtkTreeModel *model = gtk_combo_box_get_model (GTK_COMBO_BOX(priv->combo));
+
+    SCM template_menu_name = scm_c_eval_string ("gnc:report-template-menu-name/report-guid");
+    SCM get_rpt_guids = scm_c_eval_string (priv->rpt_guids);
+    SCM reportlist;
+    SCM rpt_guids;
```
jralls, line 292: MVC separation failure. This is a brand-new control so let's do it right. Keep the model/controller code in gnome-business-utils.c as `GList *gnc_invoice_report_list()`; the list can be of `struct { char* report_guid; char* report_name}`.

gnucash/gnome-utils/gnc-report-combo.c
```diff
@@ -0,0 +1,595 @@
...
+                                RC_NAME, name,
+                                RC_GUID, guid_str,
+                                RC_MISSING, FALSE,
+                                -1);
+            g_free (name);
+            g_free (guid_str);
+
+            rpt_guids = SCM_CDR(rpt_guids);
+        }
+    }
```
jralls, line 327: Most of this goes in `gnc_invoice_report_list` and here you iterate over the list and stuff the members into the list store.

gnucash/gnome/business-options-gnome.cpp
```diff
@@ -209,6 +209,61 @@ create_option_widget<GncOptionUIType::TAX_TABLE>(GncOption& option,
     return widget;
 }
 
+class GncGtkInvReportUIItem : public GncOptionGtkUIItem
```
jralls, line 213: Why does this need a new UIItem instead of using a GncMultiselectOption?

jralls (changes requested):

gnucash/gnome-utils/gnc-report-combo.c
```diff
@@ -0,0 +1,595 @@
...
+    {
+        gchar *guid;
+        gtk_tree_model_get (model, &iter, RC_GUID, &guid, -1);
+
+        if (g_strcmp0 (priv->active_report_guid, guid) == 0)
+        {
+            gtk_combo_box_set_active_iter (GTK_COMBO_BOX(priv->combo), &iter);
+            g_free (guid);
+            found = TRUE;
+            break;
```
gjanssens, line 247: Technically, if you break out of the while loop because the current iter is the one you like, `valid_iter` is still TRUE, so you don't need the separate `found` boolean. You can just continue with the state of `valid_iter` further down.

And more even, if you found what you need, you can just return rather than breaking. The second part of this function will then only be run if there was no active selection yet and hence you can drop the if clause around it in that case and just return false.

gjanssens: You may also want to document what the returned boolean means. Otherwise it takes some code deciphering before the reader gets it.
With that I mean, the function name `select_active` suggests an imperative action, but the function has side effects (it may add an entry in the combo list!) and returns a value. Without carefully examining the code and the context in which it's called this is not too obvious. So better add a comment and perhaps improve the function name.

gnucash/gnome-utils/gnc-report-combo.c
```diff
@@ -0,0 +1,595 @@
...
+static void
+update_warning_tooltip (GncReportCombo *grc)
+{
+    GncReportComboPrivate *priv = GET_PRIVATE(grc);
+    gchar *tool_tip;
+
+    if (priv->active_report_name)
+        /* Translators: %s is the report name. */
+        tool_tip = g_strdup_printf (_("'%s' is missing"),
+                                      priv->active_report_name);
```
gjanssens, line 339: "Report x is missing" may be misleading. From reading the gnc_report_combo_set_active function any report guid, name combination can be passed to this widget. Some of these combinations may be as you suggest be saved reports that no longer exist in the book, but it could just as well be a combination that refers to a report that's not a printable invoice report. So perhaps that's a useful second check to give a more precise warning message to the user.

gnucash/gnome-utils/gnc-report-combo.c
```diff
@@ -0,0 +1,595 @@
...
+    update_report_list (grc);
+
+    if (!select_active (grc))
+        update_warning_tooltip (grc);
+
+    priv->block_signal = FALSE;
+}
+
+GtkWidget *
+gnc_report_combo_get_warning_image (GncReportCombo *grc)
```
gjanssens, line 537: Unless I missed something, I don't see this function used anywhere. Why does it exist and why is it exposed in the public interface of this widget ?

gjanssens (changes requested):
Some more feedback on this.

gnucash/gnome/dialog-invoice.c
```diff
@@ -846,9 +994,20 @@ gnc_invoice_window_printCB (GtkWindow* parent, gpointer data)
...
         gnc_plugin_page_report_reload (GNC_PLUGIN_PAGE_REPORT (iw->reportPage));
     else
-        iw->reportPage = gnc_invoice_window_print_invoice
-            (parent, iw_get_invoice (iw));
+    {
+        gchar *report_guid = use_default_report_template_or_change (parent);
+
+        if (!report_guid)
+        {
+            g_free (report_guid);
```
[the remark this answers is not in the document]
jralls: > If report_guid is NULL there's no point in g_free'in it. You can just return.

But if it's not null it does need to be freed, he got it backwards. But `g_free` is null-protected so it doesn't need the conditional.

libgnucash/app-utils/gnc-optiondb.cpp
```diff
@@ -853,6 +853,16 @@ gnc_register_taxtable_option(GncOptionDB* db, const char* section,
...
 }
 
+void
+gnc_register_invoice_print_report_option(GncOptionDB* db, const char* section,
+                                         const char* name, const char* key,
+                                         const char* doc_string, std::string value)
+{
+    GncOption option{section, name, key, doc_string,
+                     value, GncOptionUIType::INV_REPORT};
+    db->register_option(section, std::move(option));
```
jralls, line 863: I didn't notice this before. *It doesn't work like that.* `GncOption` is a variant, you need to construct one of the GncOptionValue classes to pass in to the GncOption constructor.

gnucash/gnome-utils/gnc-report-combo.c
Bob-IT: I was thinking that the scheme was part of as it was only to do with reports but have changed it.

gnucash/gnome/dialog-invoice.c
```diff
@@ -789,24 +790,171 @@ gnc_invoice_window_blankCB (GtkWidget *widget, gpointer data)
...
-    if (!reportname)
-        reportname = "5123a759ceb9483abf2182d01c140e8d"; // fallback if the option lookup failed
+
+    is_invoice_guid = scm_c_eval_string ("gnc:report-is-invoice-report?");
+    scm_guid = scm_from_utf8_string (report_guid);
+
+    if (scm_is_false (scm_call_1 (is_invoice_guid, scm_guid)))
+        use_report_guid = g_strdup (PRINTABLE_INVOICE_GUID); // fallback if the option lookup failed
+    else
+        use_report_guid = g_strdup (report_guid);
```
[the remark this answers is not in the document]
Bob-IT: Changed...

gnucash/gnome/dialog-invoice.c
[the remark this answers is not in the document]
Bob-IT: Yes, another case of forgetting that g_free is null-protected and totally backwards, changed

gnucash/gnome-utils/gnc-report-combo.c
Bob-IT: Changed...

gnucash/gnome-utils/gnc-report-combo.c
Bob-IT: I added this at the start and as you pointed out never used so deleted.

gnucash/gnome/dialog-invoice.c
```diff
@@ -789,24 +790,171 @@ gnc_invoice_window_blankCB (GtkWidget *widget, gpointer data)
...
+}
+
+static void
+combo_changed_cb (GtkComboBox *widget, gpointer user_data)
+{
+    g_source_remove_by_user_data (user_data);
+}
+
+static char*
+use_default_report_template_or_change (GtkWindow *parent)
```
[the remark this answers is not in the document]
Bob-IT: Yes it will always return a guid, either the default invoice report setting or the builtin default.

gnucash/gnome/business-gnome-utils.h
```diff
@@ -37,6 +37,45 @@
 
 #define GNC_PREF_AUTO_PAY "auto-pay"
 
+#define PRINTABLE_INVOICE_GUID "5123a759ceb9483abf2182d01c140e8d"
+#define TAX_INVOICE_GUID       "0769e242be474010b4acf264a5512e6e"
+#define EASY_INVOICE_GUID      "67112f318bef4fc496bdc27d106bbda4"
+#define FANCY_INVOICE_GUID     "3ce293441e894423a2425d7a22dd1ac6"
```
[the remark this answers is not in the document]
Bob-IT: I moved them thinking I was going to need them at the start but as you pointed out the only one used else where is `PRINTABLE_INVOICE_GUID` so have added them back to `business-gnome-utils.c`. These are realy only needed in the migration from prefs to book properties, after that is done the list is dynamic with only `PRINTABLE_INVOICE_GUID` being kept to be used as the builtin default.
Will need to write a test for its existence.

libgnucash/app-utils/gnc-optiondb.cpp
Bob-IT: I left it as `empty_string` as I did not want to change it to the guid value of `PRINTABLE_INVOICE_GUID`. Instead I change this in `business-options-gnome.cpp` where I can also use the define.

libgnucash/app-utils/gnc-optiondb.cpp
Bob-IT: Not sure about this, it seems to be working OK.

Bob-IT:
I have re-based and made most of the changes and seems to work OK.
The reason for adding the new options type was two fold...
The first was that I wanted to display a visual indication that the user set default invoice report template may be missing.
The other was that I sort of understood what was needed doing it this way.

I suppose the other way would be to just silently drop the saved default template. In that case then the existing combo option could be used but am still not sure how.

There are a couple of things that may need changing.
The migration of the preferences to book properties is triggered when `gnc_get_default_invoice_print_report` is used to print an invoice and only updates the book properties if the existing preference is not `PRINTABLE_INVOICE_GUID`. This was done to  try to keep the book properties free of default entries but this may need changing. I can not do the migration from the book options page as the values have already been loaded and can not see another way apart from calling the migration function from some where else and maybe setting / testing for the book properties so it does not keep being called.

The other thing was to do with `pref_transormations.xml`, I tagged onto the end for the section dealing with `release version="5000"` to depreciate the old preference path which I think works but may need further testing. I then added a new `release version="6000"` to make it obsolete but doing that causes the preference setting to always return 0 and I am not sure why.

libgnucash/app-utils/gnc-optiondb.cpp
jralls: My error, it does work like that for simple option types.

jralls:
> I presume the answer is: to be able to use the GncReportCombo widget with all of its specialization to display a subset of report templates based on a filtering guile function.

Multichoices [can do that already](https://github.com/Gnucash/gnucash/blob/master/libgnucash/engine/gnc-option-impl.hpp#L432), the filter function just has to output the filter of tuples. 

Bob's reasons are better, though I think that it would be better to use a Multichoice or a List option to back up the new control.

jralls:
Tested and it does work, so I've merged it to get it into 4.901 this weekend. A couple of observations:
* We might need to bump the max timeout a bit, 10 seconds is a little short at least until one gets used to it.
* When the selected default is missing ISTM the timeout box should say that the Printable Invoice will be printed. One easy way to do that would be to change the value displayed in the list box (it's not a combo, you can't type into it) and have the warning icon's tooltip say what the missing former default is.
* If the selected default is missing and the timeout is 0 the print button runs the Printable Invoice report without any indication to the user until they open File>Properties. We might pop a message box saying that the default has gone missing and we've run the Printable Invoice instead.
* If you run the Printable Invoice, use report options to customize it, and then save that config, the default in File>Properties is automatically changes to the newly-created saved config. No doubt some users will find that a nice touch and others will complain that it's surprising and shouldn't happen without permission.

christopherlam:
Does this line of work require any scheme changes?

Bob-IT:
I have worked out what is going on here..
> The other thing was to do with `pref_transormations.xml`, I tagged onto the end for the section dealing with `release version="5000"` to depreciate the old preference path which I think works but may need further testing. I then added a new `release version="6000"` to make it obsolete but doing that causes the preference setting to always return 0 and I am not sure why.

It all points to `transform_settings` in `gnc-settings.cpp`, my current `old_maj_min` value is 4900 and so both the 5000 and 6000 sections were being processed and the obsolete section resets the key to its default value.

@gjanssens
Is this correct, shouldn't values greater than `old_maj_min` not be processed ?
The other thing I was trying to work out is what would needs to be done in the long run and when, does the key need to be removed from `...business.scheme.xml.in` file after being marked as obsolete but when ?

Bob-IT:
> * We might need to bump the max timeout a bit, 10 seconds is a little short at least until one gets used to it.

Can change max value.
> 
> * When the selected default is missing ISTM the timeout box should say that the Printable Invoice will be printed. One easy way to do that would be to change the value displayed in the list box (it's not a combo, you can't type into it) and have the warning icon's tooltip say what the missing former default is.

I wonder if it is easier to just stop and let the user cancel or change the template to use.
> 
> * If the selected default is missing and the timeout is 0 the print button runs the Printable Invoice report without any indication to the user until they open File>Properties. We might pop a message box saying that the default has gone missing and we've run the Printable Invoice instead.

Again, one could possibly display the timeout dialog and wait for user intervention.
> 
> * If you run the Printable Invoice, use report options to customize it, and then save that config, the default in File>Properties is automatically changes to the newly-created saved config. No doubt some users will find that a nice touch and others will complain that it's surprising and shouldn't happen without permission.

I will need to test this, if I am reading this correctly it was not intended.

jralls:
Using the timeout box without the timeout is also reasonable for missing saved configs.

gjanssens:
> I have worked out what is going on here..
> 
> > The other thing was to do with `pref_transormations.xml`, I tagged onto the end for the section dealing with `release version="5000"` to depreciate the old preference path which I think works but may need further testing. I then added a new `release version="6000"` to make it obsolete but doing that causes the preference setting to always return 0 and I am not sure why.
> 
> It all points to `transform_settings` in `gnc-settings.cpp`, my current `old_maj_min` value is 4900 and so both the 5000 and 6000 sections were being processed and the obsolete section resets the key to its default value.
> 
> @gjanssens Is this correct, shouldn't values greater than `old_maj_min` not be processed ? The other thing I was trying to work out is what would needs to be done in the long run and when, does the key need to be removed from `...business.scheme.xml.in` file after being marked as obsolete but when ?

I think I never intended this to have transformations for future gnucash versions... But you can add a restriction in there that will not do transformations if the transformation version is higher than current gnucash. I think that's quite convenient, because it's very likely we'll forget to add the obsoletions if we have to wait a full major release cycle before we can add them.

PR #1335, opened 2022-05-14
Bug 771095, partially solved. Inserting a help button in the Since-Last-Run dialog.

CWehli:
Since-Last-Run dialog.

### DIFF
--- a/gnucash/gnome/dialog-sx-since-last-run.c
+++ b/gnucash/gnome/dialog-sx-since-last-run.c
@@ -1137,6 +1137,10 @@ dialog_response_cb (GtkDialog *dialog, gint response_id, GncSxSinceLastRunDialog
     GList* creation_errors = NULL;
     switch (response_id)
     {
+    case GTK_RESPONSE_HELP:
+        gnc_gnome_help (GTK_WINDOW(dialog), HF_HELP, HL_SX_SLR);
+        break;
+
     case GTK_RESPONSE_OK:
         // @@fixme validate current state(GError *errs);
         // - [ ] instance state constraints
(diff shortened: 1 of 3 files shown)


fellen:
Seems OK from my documenter view.
BTW, @CWehli is the german translation  pending?

CWehli:
> BTW, @CWehli is the german translation pending?

@fellen: I haven't started that yet.

gnucash/gnome/dialog-sx-since-last-run.c
```diff
@@ -1187,6 +1187,10 @@ dialog_response_cb (GtkDialog *dialog, gint response_id, GncSxSinceLastRunDialog
     g_list_free (app_dialog->created_txns);
     app_dialog->created_txns = NULL;
 
+    case GTK_RESPONSE_HELP:
+            gnc_gnome_help (GTK_WINDOW(dialog), HF_HELP, HL_SX_SLR);
+            return;
+
```
Bob-IT, line 1193: In this position, the help page is opened when you press the OK button as it using falling through to the help case.
So move these lines to be the first case, above the OK case and change the 'return' to be 'break'. Also make sure the alignment is correct, in this case remove 4 spaces from the two lines.
With these changes and a rebase, this should be OK.

Bob-IT (changes requested):

gnucash/gnome/dialog-sx-since-last-run.c
jralls: @Bob-IT For your technical-debt-reduction list, the body of `GTK_RESPONSE_OK` desperately needs to be moved to at least one and maybe two functions.

jralls:
Thanks!

PR #1342, opened 2022-05-27
close-book: option to close a limited account tree

christopherlam:
from bug 798534. option to select a subgroup of inc&exp accounts to close. 

image


derekatkins:
I'm curious -- what is the rationale for this change?

derekatkins:
Interesting -- just read the bug report.  Seems like a tiny corner-case to support, but I see no harm in supporting it.

jralls:
@christopherlam Why did you decide to use a simple combo with a single account root instead of a full account selector?

Also, this is a significant UI change that should go in master.

PR #1344, opened 2022-06-02
[dialog-fincalc.glade] Inserting a help button

CWehli:


Bob-IT (approved):
Looks OK to me so will push, thanks.

PR #1345, opened 2022-06-05
Show paths

christopherlam:
Show paths in About box. Paths are clickable. Any other paths to be included?

- [x] to add onto cli `--paths`
- [x] untested on Windows

image

### DIFF
--- a/gnucash/gnome-utils/gnc-gtk-utils.c
+++ b/gnucash/gnome-utils/gnc-gtk-utils.c
@@ -297,3 +297,38 @@ gnc_style_context_get_border_color (GtkStyleContext *context,
     *color = *c;
     gdk_rgba_free (c);
 }
+
+static gpointer
+find_widget_func (GtkWidget *widget, const gchar *id)
+{
+    const gchar *name = gtk_buildable_get_name (GTK_BUILDABLE(widget));
+    GtkWidget *ret = NULL;
+
+    if (g_strcmp0 (name, id) == 0)
+        return widget;
+
+    if (GTK_IS_CONTAINER(widget))
+    {
+        GList *container_list = gtk_container_get_children (GTK_CONTAINER(widget));
+        for (GList *n = container_list; !ret && n; n = n->next)
+            ret = find_widget_func (n->data, id);
+        g_list_free (container_list);
+    }
+
+    return ret;
+}
+
+/** Find the Widget defined by 'id' in the dialog
+ *
+ *  @param dialog The dialog to search for 'id'.
+ *
+ *  @param id The widget name to find in the dialog.
+ *
+ *  @returns The widget defined by id in the dialog or NULL.
+ */
+GtkWidget *
+gnc_get_dialog_widget_from_id (GtkDialog *dialog, const gchar *id)
+{
+    GtkWidget *content_area = gtk_dialog_get_content_area (dialog);
+    return find_widget_func (content_area, id);
+}
--- a/libgnucash/core-utils/gnc-filepath-utils.cpp
+++ b/libgnucash/core-utils/gnc-filepath-utils.cpp
@@ -69,6 +69,7 @@ extern "C" {
 #include <boost/filesystem.hpp>
 #include <boost/locale.hpp>
 #include <iostream>
+#include <numeric>
 
 /* Below cvt and bfs_locale should be used with boost::filesystem::path (bfs)
  * objects created alter in this source file. The rationale is as follows:
@@ -1306,5 +1307,27 @@ gnc_filepath_locate_doc_file (const gchar *name)
     return result;
 }
 
+GList *
+gnc_list_all_paths (void)
+{
+    if (gnc_userdata_home.empty())
+        gnc_filepath_init ();
+
+    std::vector<EnvPaths> paths
+        { { "GNC_USERDATA_DIR", gnc_userdata_home_str.c_str(), true},
+          { "GNC_USERCONFIG_DIR", gnc_userconfig_home_str.c_str(), true },
+          { "GNC_BIN", g_getenv ("GNC_BIN"), false },
+          { "GNC_LIB", g_getenv ("GNC_LIB"), false },
+          { "GNC_CONF", g_getenv ("GNC_CONF"), false },
+          { "GNC_DATA", g_getenv ("GNC_DATA"), false },
+        };
+    auto accum = [](const auto& a, const auto& b)
+    {
+        EnvPaths *ep = g_new0 (EnvPaths, 1);
+        *ep = b;
+        return g_list_prepend (a, ep);
+    };
+    return std::accumulate (paths.rbegin(), paths.rend(), (GList*) nullptr, accum);
+}
 
 /* =============================== END OF FILE ========================== */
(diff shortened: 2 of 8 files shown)


jralls:
> The links do not work on windows, first I thought it was down to file:// when it should be file:/// for windows, see gnc-uri-utils.c

It's [always three slashes](https://github.com/Gnucash/gnucash/blob/e14dfea3d168fbd4ae24a9150231a6156c349e8c/libgnucash/engine/gnc-uri-utils.c#L328)  as required by [RFC-8029](https://datatracker.ietf.org/doc/html/rfc8089#section-2), though our code allows for back-slashes..

libgnucash/core-utils/gnc-filepath-utils.cpp
```diff
@@ -1306,5 +1307,53 @@ gnc_filepath_locate_doc_file (const gchar *name)
...
+        auto& [ name, val, user_modifiable ] = *it;
+        ss << name << ": ";
+        if (use_markup)
+            ss << "<a href=\"file://" << val << "\">" << html_sanitize (val) << "</a>";
+        else
+            ss << val;
+        if (user_modifiable)
+            ss << " (user modifiable)";
+        if (std::next (it) != paths.end())
+            ss << '\n';
```
jralls, line 1354: Don't re-create URI generation, use `gnc_uri_create_uri ("file", NULL, 0, NULL, NULL, path)`.

libgnucash/core-utils/gnc-filepath-utils.h
```diff
@@ -172,4 +172,5 @@ gchar *gnc_filepath_locate_ui_file (const gchar *name);
  */
 gchar *gnc_filepath_locate_doc_file (const gchar *name);
 
+gchar *gnc_log_all_paths (gboolean use_markup);
```
jralls, line 175: This function should be in app-utils/gnc-help-utils. It also needs Doxygen documentation.

christopherlam:
and does a gnome link to `file:///C|/Users/user/filepath` actually open an explorer window?

christopherlam:
> No, it opens the file with the program that's registered for the extension. In the instant case all of the paths are for directories and there's no application registered for those so you get
> 
> ```
> Unable to show 'file:///C:\Users\John Ralls\AppData\Roaming\GnuCash': No application is registered as handling this file
> ```
> 
> That's [glib bug 1259](https://gitlab.gnome.org/GNOME/glib/-/issues/1259).

That's a bummer. Opening in Explorer is the whole point of this extra info. Looks like it needs a special `#ifdef WIN32` handler.

jralls:
Hmm, it looks like we're already catching GtkAboutDialog's [`activate-link` signal](https://docs.gtk.org/gtk3/signal.AboutDialog.activate-link.html) in `gnc_main_window_cmd_help_about` and calling `gnc_launch_doclink`. That function uses `gtk_open_uri_on_window` only for non-macOS Unix, providing native implementations on Win32 and macOS.

@Bob-IT what is the transaction linked-document code calling?

Bob-IT:
@jralls I think it is using the `gnc_launch_doclink` code but have not checked.
If I remember when I tried this, the call back only applies to the website label and not the version label which is where all this is being added.
As I said it maybe possible to add link button widgets to the dialog by deriving the containers from the dialog, that way they could also use an 'activate-link' callback.
I suppose one could change the website url to test but did not do that.

Bob-IT:
@jralls I see what you mean, I was reading the docs wrong, had a quick look at the source of GtkAboutDialog and see in `text_buffer_new` is where the `uri` data tag is set. From a quick glance it seems to only recognize `http://` and `https://` for setting `uri` other wise it sets the tag to `email` if one is found.

I am still thinking one could add individual link buttons to the content_area, I can have a look at that on Friday unless you see another approach.

jralls:
@Bob-IT It turns out that the dialog is the wrong control: The button event is getting handled by the version label so we'd need to connect to that to use `gnc_launch_doclink` on the paths.
```
--- a/libgnucash/core-utils/gnc-filepath-utils.cpp
+++ b/libgnucash/core-utils/gnc-filepath-utils.cpp
@@ -1345,7 +1345,9 @@ gnc_log_all_paths (gboolean use_markup)
         auto& [ name, val, user_modifiable ] = *it;
         ss << name << ": ";
         if (use_markup)
-            ss << "<a href=\"file://" << val << "\">" << html_sanitize (val) << "</a>";
+            ss << "<a href=\"file://" <<
+                g_uri_escape_string (val.c_str(), "/\\:", TRUE) <<
+                "\">" << html_sanitize (val) << "</a>";
         else
             ss << val;
         if (user_modifiable)
```
Fixes the space in the path problem on macOS. Doesn't do anything for Windows of course.

jralls:
> version label so we'd need to connect to that to use gnc_launch_doclink on the paths

And the label is unfortunately private so we can't get to it to connect to its signal. I looked at `follow_if_link` some more and fount that it's only connected to GtkTextViews, meaning the credits and license boxes. I don't think it would be useful to put the paths in the credits box though doing so with `gtk_about_dialog_add_credit_section` does work.

It's still eluding me how the website label is getting connected to the about dialog's `activate-link` signal.

jralls:
Adding a label--no need for buttons, just connect to its `activate-link` with `gnc_launch_doclink`--seems like the cleanest solution, better in fact than jamming the paths into the version string.

Or you could make another toggle-button that opens a GtkTextView with the paths.

Bob-IT:
This was more involved that originally thought.
Trying out [Gist](https://gist.github.com/Bob-IT/b2a371704c2a60154217517785f6eaaa) to share two patches of my changes.

The first patch does the same as Chris's changes and works on Windows and Linux. In this I am finding the vbox of the default about page and adding a GtkGrid with GtkLinkButtons populated from a GList. To get the same layout, I think hbox's could be used with the items being added to these if desired.
Applying the second patch also changes this by adding another stack and to do the switching a 'Paths' button. This was my original thought which was simple for Linux as it uses a header bar but on Windows it uses an action area which as you can see is more involved.

I will leave it to @christopherlam to make any additional changes required.

Here is an image on Windows...
about2

gnucash/gnome-utils/gnc-gtk-utils.c
```diff
@@ -297,3 +297,51 @@ gnc_style_context_get_border_color (GtkStyleContext *context,
     *color = *c;
     gdk_rgba_free (c);
 }
+
+struct about_data
```
christopherlam, line 301: maybe this shouldn't be named `about_data` because it applies to any container. This file's functions iterates all children even if the widget has been found. Shouldn't it return early if widget is found?

Bob-IT: I think I missed changing it when I moved the function to gnc-gtk-utils, so maybe change to `dialog_data`. As for finishing early if found, I did not think to do so as there weren't very many widgets and the rarity of use but if you think it worth doing, change it.

christopherlam:
I hadn't seen bob's second patch and cannot understand the deprecated action-area api.

The GtkAboutDialog also has a `system-information` page which could be used, but is available in Gtk4 onwards.

fellen:
Sorry that I didn't comment before, but from my POV this does not belong in this context and  is cluttering the About page. 

Perhaps it should be something like "View/edit predefined settings/pathes",
at least some kind of subpage "Pathes" like Credits

PR #1347, opened 2022-06-09
Implementing Co-Owner support

rzerres:
This branch is work in progress!!

# Version: 0.2

# Status
Given code does compile ... well the .c and .cpp part.

But i can't figure out, why the scm part can't be completed.
I need a reviewer how has more insight view on the scheme part.

# Abstract

In Germany, a co-owners' association (MEG) is free to decide to
self-manage the property in an owner administration. The tasks of
[property][property_mgmt] management can then be taken over by an
owner of the MEG appointed for this purpose.

The rules may be adopted to the needs in other countries as well.

[property_mgmt]: https://www.hausverwaltung-ratgeber.de/hausverwaltung.html).

## Implementation

Using the current version of 'gnucash', the property management can
create the account structure required for this and manage the
necessary bookings. The legal regulations that exist in Germany
stipulate that individual accounts must be created annually for the
owners, which in particular must show and proportionately allocate
apportionable amounts, non-apportionable amounts, the maintenance
reserve, as well as income, individual and heating costs.

## Property Management

It seems logical, to create a `business` entity `co-owner`.
This entity is created as a `child` of `owner` mutch like the implementation of `Employee` or `customer`.

You can find the actual code in my branch [Co-Owner](https://github.com/rzerres/gnucash/tree/wip-coowner)

## Reports

As a first goal, a new report will provide an owner-bill report.

The report will have the following structure:

* Apportionable amounts
* Individual costs
* Heating costs
* Non-Apportionalbe amounts
* Revenues
* Provision for maintenance

Sub-acounts inside this structure are listed in colums, showing the cumulated balances, the distribution key, the property-total, the apartment/Housing unit shares beside the calculated co-owner amount/balance.

A final section will print out the billing totals comparing the business-plan values with the actual realized values in the billing period (usual one year).
Again you will see the total amounts as well as the calculated Co-Owner amount based on the actual share.

The report will conclude with a Credit-Balance or with an Additional Payment.


jralls:
It's a shame that you've done so much work without first discussing it on the developer list. Unfortunately this is very much out of scope for GnuCash so we must decline it.

christopherlam:
The main issue is it could fix a very narrow use case. But GnuCash has a million moving parts which are not proven to be unbroken by this engine change.

PR #1350, opened 2022-06-14
Add action group to View menu for the tab position

nomis:
This makes it easier to change the tab position quickly without opening the preferences window.

With many tabs open there's more space to display all tabs vertically but when resizing GnuCash to half the display size they use up too much horizontal space which can be fixed by temporarily making the tabs horizontal.

----
PR to maint: #1369

### DIFF
--- a/gnucash/gnome-utils/gnc-main-window.cpp
+++ b/gnucash/gnome-utils/gnc-main-window.cpp
@@ -182,6 +182,7 @@ static void gnc_main_window_cmd_view_refresh (GtkAction *action, GncMainWindow *
 static void gnc_main_window_cmd_view_toolbar (GtkAction *action, GncMainWindow *window);
 static void gnc_main_window_cmd_view_summary (GtkAction *action, GncMainWindow *window);
 static void gnc_main_window_cmd_view_statusbar (GtkAction *action, GncMainWindow *window);
+static void gnc_main_window_cmd_view_tab_position (GtkAction *action, GtkRadioAction *current, GncMainWindow *window);
 static void gnc_main_window_cmd_actions_reset_warnings (GtkAction *action, GncMainWindow *window);
 static void gnc_main_window_cmd_actions_rename_page (GtkAction *action, GncMainWindow *window);
 static void gnc_main_window_cmd_window_new (GtkAction *action, GncMainWindow *window);
@@ -356,6 +357,10 @@ static GtkActionEntry gnc_menu_actions [] =
     /* View menu */
 
     {
+        "ViewTabPositionAction", nullptr, N_("Tab P_osition"), nullptr,
+        nullptr, nullptr
+    },
+    {
         "ViewSortByAction", nullptr, N_("_Sort By..."), nullptr,
         N_("Select sorting criteria for this page view"), nullptr
     },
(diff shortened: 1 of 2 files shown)


Bob-IT:
I had a look at this and the first thing to say is if you wanted it in the current version use the maint branch, master is used for the next major version which will be 5.0
I think it is a small change and so could go to maint but have created two patches on [Gist](https://gist.github.com/Bob-IT/22416137a5b8b4685e3635ccce9677c4#file-mychanges-for-tab-patch), the first are the changes I would make to your PR, the second on [Gist](https://gist.github.com/Bob-IT/22416137a5b8b4685e3635ccce9677c4#file-change-to-view-menu-patch) are to do with moving it to the 'View' menu.

I think it fits better under the 'view' menu that is why I added the second patch but I maybe wrong. Also with this change a reason could be made for removing the option under preferences possibly for the next major release.

jralls:
I'm inclined to agree with Simon that it belongs on the Windows menu because there we already have `Show Previous Tab`, `Show Next Tab`, and `Move Tab to New Window`.

nomis:
> have created two patches on [Gist](https://gist.github.com/Bob-IT/22416137a5b8b4685e3635ccce9677c4#file-mychanges-for-tab-patch), the first are the changes I would make to your PR

You've re-ordered the entries so that they are no longer consistent with the same list in the preferences window, probably so that you could do this other change which is unsafe because it makes an assumption that the GTK enum value used for positions is sequential, starts at 0 and exactly matches the list used in GnuCash:
```
+    if (position == 0)
+        action = first_radio_action;
+    else
+        action = gtk_action_group_get_action (priv->action_group, tab_pos_radio_entries[position].name);
```

The `position_set` check can't be removed. This isn't about there being no preferences; the problem is the preference code sets the old value to false and then the new value to true. These are two preference change events and they make the application slower (although I think *saving* the preference twice may be taking the most time).

I'll make the `first_radio_action`, use of `nullptr` (the code in `maint` is C) and whitespace changes soon.

Bob-IT:
> You've re-ordered the entries so that they are no longer consistent with the same list in the preferences window, probably so that you could do this other change which is unsafe because it makes an assumption that the GTK enum value used for positions is sequential, starts at 0 and exactly matches the list used in GnuCash:

Yes, I reordered the list to match the GtkPositionType which I think is very unlikely to change, the position on the radio buttons has not changed so personally see nothing wrong with that but I have been wrong before.

>The position_set check can't be removed. This isn't about there being no preferences; the problem is the preference code sets the old value to false and then the new value to true. These are two preference change events and they make the application slower (although I think saving the preference twice may be taking the most time).

Did you not see,
`    if (pref && !gnc_prefs_get_bool (GNC_PREFS_GROUP_GENERAL, pref))`
`        return;`

I thought that takes care of the double pref change.

Bob-IT:
> I'm inclined to agree with Simon that it belongs on the Windows menu because there we already have `Show Previous Tab`, `Show Next Tab`, and `Move Tab to New Window`.

I assume this is a MAC thing, I do not see that on my Linux VM and maybe if I had I would not of proposed the move.

nomis:
Ok, I understand what this is doing now - if the preference being changed is `false` it can be ignored because another one is about to be set to `true`:
> `if (pref && !gnc_prefs_get_bool (GNC_PREFS_GROUP_GENERAL, pref))`

I've pushed the updated version, which also changes the mnemonics to match those on the preferences window. They're isolated in their own menu so they could all use the first letter, but this makes translation easier.

christopherlam:
Apologies for irrelevant comment :-) I also use tabs and am *wishing* for the Window menu to be able to show all open tabs in current window (in addition to list of windows). The only interface for switching tabs is Ctrl-PgUp/PgDn and right-click on a tab. So, if there's any takers, someone will be very happy :+1:

Bob-IT:
> The only interface for switching tabs is Ctrl-PgUp/PgDn and right-click on a tab.

You can also use Ctrl+Alt+Space or menu key to bring up that tab menu and navigate from there.

jralls:
> You can also use Ctrl+Alt+Space or menu key to bring up that tab menu and navigate from there.

Or right-click in the tabs to do the same. BTW Ctrl-Alt-Space doesn't work on macOS.

ISTM this PR is superfluous.

nomis:
> ISTM this PR is superfluous.

But this PR is not to do what @christopherlam is asking for... it is to temporarily stop the tabs taking up so much space in the window.

nomis:
As requested I've moved this to the View menu

jralls:
Superseded by #1369.

PR #1359, opened 2022-06-24
Translation update

mocsa:
L10N: hu Translated Stock Split Assistant and other corrections.
3052 translated messages, 1371 fuzzy translations, 978 untranslated messages.


fellen:
Hello Molnár,

I have the following problem:
Szia Tomi updates the file using https://hosted.weblate.org/projects/gnucash/gnucash/hu/
and now I am getting conflicts. and do not know, which version I should commit.

Do you think, you can use weblate, too?
It has many nice features like adding comments, formal checks …

Regards
Frank

PR #1365, opened 2022-06-27
glib to stl

christopherlam:
~2~3 down... many more to go. There's some runtime advantages; the `unordered_map`s are populated during compilation in both qofchoice and qofclass. Most map keys are short strings; SSO makes using `std::string` *much* nicer than `char*`.

### DIFF
--- a/libgnucash/engine/qof-string-cache.cpp
+++ b/libgnucash/engine/qof-string-cache.cpp
@@ -25,7 +25,9 @@
  *   Author: Linas Vepstas (linas@linas.org)                        *
  *   Author: Phil Longstaff (phil.longstaff@yahoo.ca)               *
 \********************************************************************/
-#include <glib.h>
+
+#include <string>
+#include <unordered_map>
 
 extern "C"
 {
@@ -43,40 +45,16 @@ static QofLogModule log_module = QOF_MOD_UTIL;
 /* =================================================================== */
 /* The QOF string cache                                                */
 /*                                                                     */
-/* The cache is a GHashTable where a copy of the string is the key,    */
-/* and a ref count is the value                                        */
+/* The cache is an unordered_map where a copy of the string is the     */
+/* key, and a ref count is the value                                   */
 /* =================================================================== */
 
-static GHashTable* qof_string_cache = NULL;
-
-static GHashTable*
-qof_get_string_cache(void)
-{
-    if (!qof_string_cache)
-    {
-        qof_string_cache = g_hash_table_new_full(
-                               g_str_hash,               /* hash_func          */
-                               g_str_equal,              /* key_equal_func     */
-                               g_free,                   /* key_destroy_func   */
-                               g_free);                  /* value_destroy_func */
-    }
-    return qof_string_cache;
-}
+static std::unordered_map<std::string, guint> qof_string_cache;
 
 void
-qof_string_cache_init(void)
+qof_string_cache_clear (void)
 {
-    (void)qof_get_string_cache();
-}
-
-void
-qof_string_cache_destroy (void)
-{
-    if (qof_string_cache)
-    {
-        g_hash_table_destroy(qof_string_cache);
-    }
-    qof_string_cache = NULL;
+    qof_string_cache.clear();
 }
 
 /* If the key exists in the cache, check the refcount.  If 1, just
@@ -84,24 +62,18 @@ qof_string_cache_destroy (void)
 void
 qof_string_cache_remove(const char * key)
 {
-    if (key && key[0] != 0)
-    {
-        GHashTable* cache = qof_get_string_cache();
-        gpointer value;
-        gpointer cache_key;
-        if (g_hash_table_lookup_extended(cache, key, &cache_key, &value))
-        {
-            guint* refcount = (guint*)value;
-            if (*refcount == 1)
-            {
-                g_hash_table_remove(cache, key);
-            }
-            else
-            {
-                --(*refcount);
-            }
-        }
-    }
+    if (!key || !key[0])
+        return;
+
+    auto cache_iter = qof_string_cache.find (key);
+    if (cache_iter == qof_string_cache.end())
+        return;
+
+    auto& refcount = cache_iter->second;
+    if (refcount == 1)
+        qof_string_cache.erase (cache_iter);
+    else
+        refcount--;
 }
 
 /* If the key exists in the cache, increment the refcount.  Otherwise,
@@ -109,32 +81,16 @@ qof_string_cache_remove(const char * key)
 const char *
 qof_string_cache_insert(const char * key)
 {
-    if (key)
-    {
-        if (key[0] == 0)
-        {
-            return "";
-        }
+    if (!key)
+        return nullptr;
 
-        GHashTable* cache = qof_get_string_cache();
-        gpointer value;
-        gpointer cache_key;
-        if (g_hash_table_lookup_extended(cache, key, &cache_key, &value))
-        {
-            guint* refcount = (guint*)value;
-            ++(*refcount);
-            return static_cast <char *> (cache_key);
-        }
-        else
-        {
-            gpointer new_key = g_strdup(static_cast<const char*>(key));
-            guint* refcount = static_cast<unsigned int*>(g_malloc(sizeof(guint)));
-            *refcount = 1;
-            g_hash_table_insert(cache, new_key, refcount);
-            return static_cast <char *> (new_key);
-        }
-    }
-    return NULL;
+    if (!key[0])
+        return "";
+
+    auto [cache_iter, inserted] = qof_string_cache.insert ({key, 0});
+    (cache_iter->second)++;
+
+    return cache_iter->first.c_str();
 }
 
 const char *
--- a/libgnucash/engine/qofchoice.cpp
+++ b/libgnucash/engine/qofchoice.cpp
@@ -21,57 +21,40 @@
  *  Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
  */
 
+#include <algorithm>
+#include <numeric>
+#include <vector>
+#include <string>
+#include <unordered_map>
 #include <config.h>
-#include <glib.h>
 
 #include "qof.h"
 #include "qofchoice.h"
 
+using OptionVec = std::vector<char*>;
+using ParamMap = std::unordered_map<std::string, OptionVec>;
+using ChoiceMap = std::unordered_map<std::string, ParamMap>;
 static QofLogModule log_module = QOF_MOD_CHOICE;
-static GHashTable *qof_choice_table = NULL;
-
-/* To initialise, call qof_choice_add_class in
-qof_object_register for the choice object. */
-static gboolean qof_choice_is_initialized(void)
-{
-    if (!qof_choice_table)
-    {
-        qof_choice_table = g_hash_table_new(g_str_hash, g_str_equal);
-    }
-    if (!qof_choice_table)
-    {
-        return FALSE;
-    }
-    return TRUE;
-}
+static ChoiceMap qof_choice_table;
 
 gboolean qof_object_is_choice(QofIdTypeConst type)
 {
-    gpointer value = NULL;
+    g_return_val_if_fail(type != NULL, false);
 
-    if (!qof_choice_is_initialized())
+    auto choice_it = qof_choice_table.find (type);
+    if (choice_it != qof_choice_table.end())
     {
-        return FALSE;
-    }
-    g_return_val_if_fail(type != NULL, FALSE);
-    value = g_hash_table_lookup(qof_choice_table, type);
-    if ((GHashTable*)value)
-    {
-        return TRUE;
+        return true;
     }
     DEBUG (" QOF_TYPE_CHOICE setup failed for %s\n", type);
-    return FALSE;
+    return false;
 }
 
 gboolean
 qof_choice_create(char* type)
 {
-    GHashTable *param_table;
-
-    g_return_val_if_fail(type != NULL, FALSE);
-    g_return_val_if_fail(qof_choice_is_initialized() == TRUE, FALSE);
-    param_table = g_hash_table_new(g_str_hash, g_str_equal);
-    g_hash_table_insert(qof_choice_table, type, param_table);
+    g_return_val_if_fail(type != NULL, false);
+    qof_choice_table.insert_or_assign (type, ParamMap{});
     return TRUE;
 }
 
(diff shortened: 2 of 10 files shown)


libgnucash/engine/qof-string-cache.cpp
```diff
@@ -43,64 +45,42 @@ static QofLogModule log_module = QOF_MOD_UTIL;
...
-    return qof_string_cache;
-}
+static std::unordered_map<std::string, guint> qof_string_cache;
 
 void
 qof_string_cache_init(void)
 {
-    (void)qof_get_string_cache();
+    qof_string_cache.clear();
 }
```
gjanssens, line 58: Perhaps it's worth checking if this function is still necessary:

- As far as I can tell it's called only once exactly
- If I remember correctly an empty unordered map is the default state for a newly created one so clearing it doesn't do anything.

libgnucash/engine/qof-string-cache.cpp
```diff
@@ -43,64 +45,42 @@ static QofLogModule log_module = QOF_MOD_UTIL;
...
 void
 qof_string_cache_destroy (void)
 {
-    if (qof_string_cache)
-    {
-        g_hash_table_destroy(qof_string_cache);
-    }
-    qof_string_cache = NULL;
+    qof_string_cache.clear();
 }
```
gjanssens, line 64: Same here. A container cleans up after itself when it goes out of scope, so I don't think having an explicit destroy function that's called exactly once when the application quits is not very useful.

libgnucash/engine/qof-string-cache.cpp
christopherlam: Is there an (artificial) scenario where something could call init/destroy/init/destroy within 1 session? If so, both init and destroy should clear the map... :-/

gjanssens: Not that I know of. Still in such imaginary scenario it would also suffice to have only one function, with a more sensible name like for example `qof_string_cache_clear`. Your init and destroy function currently do exactly the same and are an artefact of how things had to be dealt with in plain C.

libgnucash/engine/qof-string-cache.cpp
```diff
@@ -62,48 +62,38 @@ qof_string_cache_clear (void)
...
-        else
-        {
-            auto& refcount = cache_iter->second;
-            refcount++;
-        }
+    auto cache_iter = qof_string_cache.find (key);
+    if (cache_iter == qof_string_cache.end())
+        cache_iter = qof_string_cache.insert_or_assign (key, 1).first;
+    else
+        (cache_iter->second)++;
```
gjanssens, line 95: How about:
```
auto [cache_iter, result] = qof_string_cache.insert(key, 0);
(cache_iter->second)++;
```

`insert` will return an iter to the pre-existing map element for key if it existed already or an iter to the newly inserted element if it didn't exist yet. By initializing a new value to 0, incrementing the value by 1 does the right thing regardless of the iter that's returned. For new elements the net result is the value is 1, for pre-existing elements the value is increased by one.

I think this will save iterating over the map twice (once for find and once for insert_or_assign).

Note the `auto [ a, b]` notation is a new feature in C++17 and allows to assign elements from a pair to individual variables.

libgnucash/engine/qof-string-cache.cpp
```diff
@@ -62,48 +62,38 @@ qof_string_cache_clear (void)
...
-            qof_string_cache.erase (cache_iter);
-        else
-            refcount--;
-    }
+    auto& refcount = cache_iter->second;
+    if (refcount == 1)
+        qof_string_cache.erase (cache_iter);
+    else
+        refcount--;
 }
```
gjanssens, line 77: I was looking for algorithms in the Algorithm library or member functions of an unordered map to reduce the number of `if` clauses, but in this case I didn't find much that's useful. The only possible change I'd consider to make the code slightly more in line with my suggestion for insert is to first decrease the value by one and then check if it has gone to 0 to delete it.

```
(cache_iter->second)--;
if (!cache_iter->second)
        qof_string_cache.erase (cache_iter);
```

libgnucash/engine/qof-string-cache.cpp
jralls: If you use C++17 structured binding then `cache_iter` isn't a pair any more and `cache_iter->second` doesn't exist. It's just `++cache_iter`.

jralls:
[Deja-vu](https://github.com/Gnucash/gnucash/pull/1280).

libgnucash/engine/qof-string-cache.cpp
gjanssens: > ```
> bool result;
> [cache_iter, result] = qof_string_cache.insert_or_assign({key, 1});
> ```
> 

It seems to me you're mixing up `insert` and `insert_or_assign` syntax. `insert` can use an initializer list to create a map element, `insert_or_assign` is a function with two parameters `key` and  `value`.
While not to the point it's confusing me here as I deliberately chose to use `insert` instead of `insert_or_assign` and it's not clear whether you understood why.

> is clearer than
> 
> ```
> cache_iter = qof_string_cache.insert_or_assign (key, 1).first;
> ```
> 
That I do agree to.

> Plus in either case you should test `result` to make sure that the insertion succeeded.

Why ? The way I [interpret the description in cppreference](https://en.cppreference.com/w/cpp/container/unordered_map/insert) this particular invocation of insert will always return an iterator that points to an element in the map with the key we attempted to insert. If insertion fails it will be the element that already existed with this key, otherwise it will be the element that got inserted. For our particular use case it doesn't matter which of these two it is - our next step is always to increment its counter.
Do you think my interpretation is wrong ?

christopherlam:
Ok this is modernisation and I feel should still go ahead... there are more efficient `unordered_map`s around which we can use -- https://tessil.github.io/2016/08/29/benchmark-hopscotch-map.html and https://www.reddit.com/r/cpp/comments/vbdr3z/new_boostunordered_containers_have_big/ but the longer GHashTables stick around, the more this code feels dated :-)

libgnucash/engine/qof-string-cache.cpp
jralls: > For our particular use case it doesn't matter which of these two it is

OK.

jralls:
> more this code feels dated :-)

Not a strong argument. The string cache gets used a *lot* so efficiency is important. 

Another point is that in a lot of cases we're using cached strings where we should be using something else like an enum or a GQuark. It would be better in those cases to fix the flaw.

christopherlam:
> > more this code feels dated :-)
> 
> Not a strong argument. The string cache gets used a _lot_ so efficiency is important.
> 
> Another point is that in a lot of cases we're using cached strings where we should be using something else like an enum or a GQuark. It would be better in those cases to fix the flaw.

GQuark uses a GHashTable internally, and there's no mechanism to remove an interned string, so, maybe leave things in qof-string-cache until maybe boost 1.80 is generally available.

jralls:
TL;DR: It makes more sense to get rid of the string cache than to convert it to unordered_map.

There's no mechanism to remove an enum either, and obviously using a GQuark--regardless of its internals--wouldn't further the goal of getting GLib out of libgnucash. The point of GQuarks is that they can be compared numerically instead of character-by-character. It's true that it's that it's a run-time mechanism for managing compile-time constants that is (like so much of the Gnome stack) a workaround for C's limitations. Modern C++ offers the ability through templates and constexpr to do pretty much the same thing with the expensive part, creating the hash, done at compile time.

Of the 116 `CACHE_INSERT` calls, 92 are string literals and one is a translation of a string literal ("Unnamed Budget" in gnc-budget.cpp). Two more are of QofIdType, also a compile-time constant. Of the remaining calls on run-time variables, 10 are in gnc-commodity.c, 6 in split-register-copy-ops.c, 3 in gncInvoice.c, and 2 in Split.c

gjanssens:
Interesting analysis. I agree the string cache is heavily overused. Would make for a nice cleanup to reduce its use to only those cases we either are unsure or pretty sure it may be useful.

gjanssens:
How do you see the combination of templates and constexpr being used here to deal with this in c++ ?

jralls:
Not a combination, an either-or. For C++17 and later constexpr is a lot easier to both read and write so it's preferred in those cases where it works; I think it would here.

The simplification is  to write a constexpr inline hash function to turn all of those string literals into uint64_t constants that can be easily passed around and compared. No allocs, no hash-table lookups, and by making it constexpr and declaring the result constexpr all of the hashing work happens at compile time. We could steal GLib's string hash for the purpose: It's simple, fast, and should produce sufficiently unique hashes for our purposes.

That does mean that all of the files that currently use `CACHE_INSERT` on a string literal have to be made C++.

For a lot of those cases an enum would be better because it enforces type consistency when compiled with C++; one could hypothetically have the same literal from different contexts that compare equal when the comparison should fail to compile. I doubt that we have that problem because it would have shown up as a weird bug, but it's also more idiomatic C++ (and C for that matter) to use enums for that kind of constant.

There's another wrinkle: ISTR that in some cases those string literals are also displayed, either in error messages or as msgids. That's no reason to allocate and cache them, though. If they're converted to enums then a simple corresponding lookup function will suffice. If there are enough separate cases I suppose that construct could be hoisted to a template meta-class.

christopherlam:
> I agree the string cache is heavily overused

At the very least I think `gnc-commodity.c` use of string-cache is wrong -- it caches all currency data eg names, numeric code, mnemonic. These are not reused anywhere else.

jralls:
That's true for most if not all of the string-cache uses. There are two potential benefits of a string cache: If the exact same string would be allocated several times, and if the cached char* can be used for a direct compare. Literals fail the first test, they're already in the DATA segment so there's no good reason to allocate them at all, ever. I don't think we actually do the second case in GnuCash but there might be an instance hiding in a corner that I haven't found.

One possible valid use case for a string cache would be transaction's description field, where it's common to have the same value (e.g. "Ralph's Pretty Good Grocery") month-after-month. But only careful and thorough profiling would support whether there's an actual benefit. We don't cache the transaction description field and there's no evidence that anyone has ever profiled using vs. not using the string cache. @nomis and I profiled the *implementation* a couple of years ago and got a nice pop in loading time, but we didn't consider the alternative of disabling it.

nomis:
> if the cached char* can be used for a direct compare. Literals fail the first test, they're already in the DATA segment so there's no good reason to allocate them at all, ever. I don't think we actually do the second case in GnuCash but there might be an instance hiding in a corner that I haven't found.
GnuCash does this in several places, particularly for sort ordering.

> there's no evidence that anyone has ever profiled using vs. not using the string cache. @nomis and I profiled the _implementation_ a couple of years ago and got a nice pop in loading time, but we didn't consider the alternative of disabling it.
I tried and it fails quickly because of the pointer comparisons. I have plenty of memory so it's not required.

jralls:
The rest of this is poking around the edges of QofQuery, which is targeted for replacement with SQL queries. I don't think there's any point to putting lipstick on that pig in the meantime.

jralls:
> GnuCash does this in several places, particularly for sort ordering.

That doesn't make sense, one can't order on char* addresses. You can use the addresses to compare two strings for equality provided that you've run `CACHE_INSERT` or `CACHE_REPLACE` on both of them. 

> I tried and it fails quickly because of the pointer comparisons.

What did you try? Redefining the macros to no-ops?

jralls:
>>GnuCash does this in several places, particularly for sort ordering.

>That doesn't make sense, one can't order on char* addresses. You can use the addresses to compare two strings for equality provided that you've run CACHE_INSERT or CACHE_REPLACE on both of them.

>>I tried and it fails quickly because of the pointer comparisons.

>What did you try? Redefining the macros to no-ops?

Nevermind, I tried that last dodge myself. The comparisons that blocked compiling  were of QofTypeId and that seemed pretty easily fixed by converting it to an enum. That seemed to be working OK until I got to [xaccSplitRegister](https://github.com/Gnucash/gnucash/blob/2d8bda80c3bda04d4148c5bdb615c1d153644967/libgnucash/engine/Split.c#L2259) where a variety of what should be different enum types are used for classifying QofQuery parameters. I'd rather replace QofQuery with SQL than try to redesign it, so I guess we're stuck with QofStringCache until we can get rid of QofQuery.

nomis:
> > GnuCash does this in several places, particularly for sort ordering.
> 
> That doesn't make sense, one can't order on char* addresses. You can use the addresses to compare two strings for equality provided that you've run `CACHE_INSERT` or `CACHE_REPLACE` on both of them.

Things like `xaccSplitEqual` use it. Except for the special case of `""` (which is probably quite popular) this isn't going to be faster.

> > I tried and it fails quickly because of the pointer comparisons.
> 
> What did you try? Redefining the macros to no-ops?

`strdup`/`free`

I think I remember the qof keys being inconvenient here too. The string cache could be limited to qof strings, as its name implies...

jralls:
OK. xaccSplitEqual is a boolean equality/equivalence check, not an ordering function. An ordering function has to either be less than or greater than (less than is what the STL requires) if it returns a boolean or it has to return -1/0/1 like `strcmp`. IIRC most GLib collection sort functions require the latter as a parameter.

jralls:
I was filling in some dead time this morning with Jason Turner vids on YouTube. [This one](https://www.youtube.com/watch?v=ABg4_EV5L3w&list=PLs3KjaCtOwSaqPapPV4pc1SRjypnwrXYV&index=16) got be thinking about all of the static strings in the string cache. That point near the end about the template function being the same for multiple instances of the string constant means that if the function returned a hardware-comparable int from some compile-time hash would make comparisons as fast as an enum. It wouldn't help for run-time strings like GUIDs or transaction descriptions, but it would be a nice improvement for places where we now string-cache string literals.

libgnucash/engine/qofchoice.cpp
```diff
@@ -21,107 +21,96 @@
...
 {
-    GHashTable *param_table;
-
-    g_return_val_if_fail(type != NULL, FALSE);
-    g_return_val_if_fail(qof_choice_is_initialized() == TRUE, FALSE);
-    param_table = g_hash_table_new(g_str_hash, g_str_equal);
-    g_hash_table_insert(qof_choice_table, type, param_table);
+    g_return_val_if_fail(type != NULL, false);
+    qof_choice_table.insert_or_assign (type, ParamMap{});
     return TRUE;
```
jralls, line 58: Be consistent: Either always return the macro or always return the C++ bool value. IMO the macro is better because it doesn't require the implicit cast.

libgnucash/engine/qofchoice.cpp
```diff
@@ -21,107 +21,96 @@
  *  Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
  */
 
+#include <algorithm>
+#include <numeric>
+#include <vector>
+#include <string>
+#include <unordered_map>
 #include <config.h>
-#include <glib.h>
```
jralls, line 25: This should be moved to qofchoice.h instead of removed altogether because several functions return `GList*`.

libgnucash/engine/qofclass.cpp
```diff
@@ -200,81 +153,38 @@ qof_class_get_parameter_type (QofIdTypeConst obj_name,
...
-
     if (!cb) return;
-    if (!classTable) return;
 
-    iter.fcn = cb;
-    iter.data = user_data;
-
-    g_hash_table_foreach (classTable, class_foreach_cb, &iter);
+    std::for_each (classTable.begin(), classTable.end(),
+                   [&cb, &user_data](const auto& it)
```
jralls, line 162: `cb` and `user_data` are pointers, so capture by value instead of by reference.

libgnucash/engine/qofclass.cpp
```diff
@@ -327,20 +237,16 @@ find_reference_param_cb(QofParam *param, gpointer user_data)
     {
         return;
     }
-    b->list = g_list_append(b->list, param);
+    auto list = static_cast<GList*>(user_data);
+    list = g_list_prepend (list, param);
```
jralls, line 241:
```
    auto list = static_cast<GList**>(user_data);
    *list = g_list_prepend (*list, param);
```

libgnucash/engine/qofclass.cpp
```diff
@@ -327,20 +237,16 @@ find_reference_param_cb(QofParam *param, gpointer user_data)
...
-    struct param_ref_list b;
-
-    ref_list = NULL;
-    b.list = NULL;
-    qof_class_param_foreach(type, find_reference_param_cb, &b);
-    ref_list = g_list_copy(b.list);
-    return ref_list;
+    GList *list = nullptr;
+    qof_class_param_foreach(type, find_reference_param_cb, &list);
+    return g_list_reverse (list);
```
jralls, line 249: This is silly, use `std::foreach` directly and replace `find_reference_param_cb` with a lambda. Three's no point in using C wrappers inside a C++ file.

libgnucash/engine/qofevent.cpp
```diff
@@ -43,26 +56,17 @@ static QofLogModule log_module = QOF_MOD_ENGINE;
...
     {
-        hi = static_cast<HandlerInfo*>(node->data);
-
-        if (hi->handler_id == handler_id)
+        if (hi.handler_id == handler_id)
         {
             handler_id++;
-            node = handlers;
-            continue;
+            goto restart;
```
jralls, line 68: Do you understand that this creates nested for loops?

libgnucash/engine/qofevent.cpp
```diff
@@ -43,26 +56,17 @@ static QofLogModule log_module = QOF_MOD_ENGINE;
...
+    auto handler_id = next_handler_id;
 
-    while (node)
+ restart:
+    for (const auto& hi : handlers)
     {
-        hi = static_cast<HandlerInfo*>(node->data);
-
-        if (hi->handler_id == handler_id)
+        if (hi.handler_id == handler_id)
```
jralls, line 65: This will never happen: `handler_id` is inited to `next_handler_id` that's incremented every time this function is called and never decremented, so there can never be an entry in the vector with that value.

jralls:
@christopherlam, you're wasting time on QOF. It's a crap design and changing bits and pieces of it to C++ doesn't help anything. It's too bad C++options and fixing Gtk3 on macOS ate so much time this dev cycle, I'd hoped to have QOF replaced with SQL by now, but that's going to be my first priority for the next one.

Next, it doesn't do any good to replace GLists or GHashTables with STL containers if you still have to hand out GLists or GHashTables in the class's API.

Finally, you can't not include glib.h unless you remove *all* of the glib functions and classes. That means no 'gint' variables, no `g_return_on_fail`, no GList* return values, and so on. Yes, the file might still compile without the explicit include because some other header pulls it in. That's poor practice, because that other header's glib dependency might get removed and then you have breakage in all of the files that depended on that other header to include glib.h.

christopherlam:
closed into https://github.com/christopherlam/gnucash/pull/8

PR #1367, opened 2022-06-27
[dialog-sx-since-last-run] Fix handling of last run dialog on file open

nomis:
When the "since last run" notification dialog is disabled on file open, it leaks a reference to GncSxInstanceModel and any creation errors won't be displayed. Avoid returning from the function too early.

### DIFF
--- a/gnucash/gnome/dialog-sx-since-last-run.c
+++ b/gnucash/gnome/dialog-sx-since-last-run.c
@@ -851,15 +851,16 @@ gnc_sx_sxsincelast_book_opened (void)
     if (summary.need_dialog)
     {
         gnc_ui_sx_since_last_run_dialog (gnc_ui_get_main_window (NULL), inst_model, auto_created_txns);
+        /* gnc_ui_sx_since_last_run_dialog now owns this list */
         auto_created_txns = NULL;
     }
     else
     {
-        if (summary.num_auto_create_no_notify_instances != 0)
-        {
-            if (!gnc_prefs_get_bool (GNC_PREFS_GROUP_STARTUP, GNC_PREF_SHOW_AT_FOPEN))
-                return;
+        g_list_free (auto_created_txns);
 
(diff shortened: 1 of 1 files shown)


jralls (changes requested):
Also unref `inst_model` right after `gnc_sx_instance_model_effect_change` and to free `auto_created_txns` as the first statement in the `else`.

`gnc_ui_sx_since_last_run_dialog` leaks the created_txns list if the user cancels, could you fix that too by moving the free and null out of the `GTK_RESPONSE_OK` block to after the `switch` statement?

nomis:
> Also unref `inst_model` right after `gnc_sx_instance_model_effect_change`

I can't do that because it's used by the call to `gnc_ui_sx_since_last_run_dialog()`.

> `gnc_ui_sx_since_last_run_dialog` leaks the `created_txns` list if the user cancels, could you fix that too by moving the free and null out of the `GTK_RESPONSE_OK` block to after the switch statement?

It's not freed in the `GTK_RESPONSE_OK` block of `dialog_response_cb`?

jralls:
> I can't do that because it's used by the call to gnc_ui_sx_since_last_run_dialog().

Ah, so it is. 

> It's not freed in the GTK_RESPONSE_OK block of dialog_response_cb?

That is where it's freed, and that's the problem: If the user clicks Cancel instead of OK that block doesn't execute and the list is leaked.

nomis:
> > It's not freed in the GTK_RESPONSE_OK block of dialog_response_cb?
> 
> That is where it's freed, and that's the problem: If the user clicks Cancel instead of OK that block doesn't execute and the list is leaked.

Ok... the indentation in that switch block is wrong. I've fixed that and moved the free to where `gnc_close_gui_component()` is called.

jralls (approved):

jralls:
Thanks!

PR #1369, opened 2022-07-06
Add action group to View menu for the tab position

nomis:
This makes it easier to change the tab position quickly without opening the preferences window.

With many tabs open there's more space to display all tabs vertically but when resizing GnuCash to half the display size they use up too much horizontal space which can be fixed by temporarily making the tabs horizontal.

----
PR to master: #1350

### DIFF
--- a/gnucash/gnome-utils/gnc-main-window.c
+++ b/gnucash/gnome-utils/gnc-main-window.c
@@ -166,6 +166,7 @@ static void gnc_main_window_cmd_view_refresh (GtkAction *action, GncMainWindow *
 static void gnc_main_window_cmd_view_toolbar (GtkAction *action, GncMainWindow *window);
 static void gnc_main_window_cmd_view_summary (GtkAction *action, GncMainWindow *window);
 static void gnc_main_window_cmd_view_statusbar (GtkAction *action, GncMainWindow *window);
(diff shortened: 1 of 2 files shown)


Bob-IT:
@nomis 
I was going to push this but had some errors like...
(gnucash:3684): Gtk-WARNING **: 11:17:25.834: ViewTabPositionTop: missing action ViewTabPositionTopAction

Can you move the `gtk_action_group_add_radio_actions` you added to the first line after the `if (merge_id)` and then drop all the changes for `gnc_main_window_window_menu`

Could you also add a note above the `g_signal_handlers_block_by_func`, something like...

    /* When GtkRadioAction buttons were setup with gtk_action_group_add_radio_actions,
     * the G_OBJECT for the callback is the first action so we need to use this
     * action to block/unblock for all radio buttons */

On your original PR I thought this was a mistake and had to prove it my self.

nomis:
I've fixed those issues now.

Does the View menu in `gnucash/ui/gnc-plugin-basic-commands-ui.xml` need to be changed? I can't tell which windows this file affects.

Bob-IT:
Thanks for the changes, regarding the basic-command file, I do not think so as I understand it is to do with supplying the more common commands across all plugin types and possibly over ride some.

PR #1371, opened 2022-07-09
refinement #1162 to allow edit multiple rows

christopherlam:
Today I wanted to mass-rename when importing, and could not.

This commit will allow renaming desc/notes/memo, but only enabled if rows' desc/notes/memo are identical.

The disabling works piecemeal i.e. if selection has identical description but varied memo and notes, then only the description will be allowed mass renaming.

@jeanlaroche from #1162

### DIFF
--- a/gnucash/import-export/import-main-matcher.c
+++ b/gnucash/import-export/import-main-matcher.c
@@ -96,8 +96,10 @@ enum downloaded_cols
     DOWNLOADED_COL_AMOUNT_DOUBLE, // used only for sorting
     DOWNLOADED_COL_DESCRIPTION,
     DOWNLOADED_COL_DESCRIPTION_ORIGINAL,
+    DOWNLOADED_COL_DESCRIPTION_STYLE,
     DOWNLOADED_COL_MEMO,
     DOWNLOADED_COL_MEMO_ORIGINAL,
+    DOWNLOADED_COL_MEMO_STYLE,
     DOWNLOADED_COL_NOTES_ORIGINAL,
     DOWNLOADED_COL_ACTION_ADD,
     DOWNLOADED_COL_ACTION_CLEAR,
@@ -932,62 +934,48 @@ input_new_fields (GtkWidget *parent, RowInfo *info, GtkTreeStore *store,
                   gboolean edit_desc, gboolean edit_notes, gboolean edit_memo,
                   char **new_desc, char **new_notes, char **new_memo)
 {
-    GtkWidget *desc_entry, *notes_entry, *memo_entry, *label, *grid;
-    GtkWidget *dialog =
-        gtk_dialog_new_with_buttons ("Edit imported transaction details",
-                                     GTK_WINDOW (parent),
-                                     GTK_DIALOG_MODAL | GTK_DIALOG_DESTROY_WITH_PARENT,
-                                     _("_OK"), GTK_RESPONSE_ACCEPT,
-                                     _("_Cancel"), GTK_RESPONSE_REJECT,
-                                     NULL);
-    GtkWidget *content_area = gtk_dialog_get_content_area (GTK_DIALOG (dialog));
-    gboolean retval;
+    GtkWidget  *desc_entry, *notes_entry, *memo_entry, *label;
+    GtkWidget  *dialog;
+    GtkBuilder *builder;
+    gboolean    retval = FALSE;
 
-    grid = gtk_grid_new ();
+    builder = gtk_builder_new ();
+    gnc_builder_add_from_file (builder, "dialog-import.glade", "transaction_edit_dialog");
 
-    label = gtk_label_new ("Description");
-    gtk_grid_attach (GTK_GRID (grid), label, 0, 0, 1, 1);
+    dialog = GTK_WIDGET(gtk_builder_get_object (builder, "transaction_edit_dialog"));
 
-    desc_entry = gtk_entry_new ();
-    gtk_widget_set_halign (desc_entry, GTK_ALIGN_START);
+    desc_entry = GTK_WIDGET(gtk_builder_get_object (builder, "desc_entry"));
+    memo_entry = GTK_WIDGET(gtk_builder_get_object (builder, "memo_entry"));
+    notes_entry = GTK_WIDGET(gtk_builder_get_object (builder, "notes_entry"));
+
     gtk_widget_set_sensitive (desc_entry, edit_desc);
     gtk_entry_set_text (GTK_ENTRY (desc_entry), xaccTransGetDescription (info->trans));
-    gtk_grid_attach (GTK_GRID (grid), desc_entry, 1, 0, 1, 1);
 
-    label = gtk_label_new ("Notes");
-    gtk_grid_attach (GTK_GRID (grid), label, 0, 1, 1, 1);
-
-    notes_entry = gtk_entry_new ();
-    gtk_widget_set_halign (notes_entry, GTK_ALIGN_START);
     gtk_widget_set_sensitive (notes_entry, edit_notes);
     gtk_entry_set_text (GTK_ENTRY (notes_entry), xaccTransGetNotes (info->trans));
-    gtk_grid_attach (GTK_GRID (grid), notes_entry, 1, 1, 1, 1);
 
-    label = gtk_label_new ("Memo");
-    gtk_grid_attach (GTK_GRID (grid), label, 0, 2, 1, 1);
-
-    memo_entry = gtk_entry_new ();
-    gtk_widget_set_halign (memo_entry, GTK_ALIGN_START);
     gtk_widget_set_sensitive (memo_entry, edit_memo);
     gtk_entry_set_text (GTK_ENTRY (memo_entry), xaccSplitGetMemo (info->split));
-    gtk_grid_attach (GTK_GRID (grid), memo_entry, 1, 2, 1, 1);
 
-    gtk_container_add_with_properties (GTK_CONTAINER (content_area), grid,
-                                       "position", 1, NULL);
+    gtk_window_set_transient_for (GTK_WINDOW (dialog), GTK_WINDOW (parent));
 
     // run the dialog
-    gtk_widget_show_all (grid);
+    gtk_widget_show_all (dialog);
 
-    retval = (gtk_dialog_run (GTK_DIALOG (dialog)) == GTK_RESPONSE_ACCEPT);
-
-    if (retval)
+    switch (gtk_dialog_run (GTK_DIALOG(dialog)))
     {
+    case GTK_RESPONSE_OK:
         *new_desc = g_strdup (gtk_entry_get_text (GTK_ENTRY (desc_entry)));
         *new_notes = g_strdup (gtk_entry_get_text (GTK_ENTRY (notes_entry)));
         *new_memo = g_strdup (gtk_entry_get_text (GTK_ENTRY (memo_entry)));
+        retval = TRUE;
+        break;
+    default:
+        break;
     }
 
     gtk_widget_destroy (dialog);
+    g_object_unref (G_OBJECT(builder));
     return retval;
 }
 
@@ -1027,8 +1015,11 @@ gnc_gen_trans_edit_fields (GtkMenuItem *menuitem, GNCImportMainMatcher *info)
             RowInfo *row = n->data;
             if (info->edit_desc)
             {
+                guint64 style = g_strcmp0 (new_desc, row->orig_desc) ?
+                    PANGO_STYLE_ITALIC : PANGO_STYLE_NORMAL;
                 gtk_tree_store_set (store, &row->iter,
                                     DOWNLOADED_COL_DESCRIPTION, new_desc,
+                                    DOWNLOADED_COL_DESCRIPTION_STYLE, style,
                                     -1);
                 xaccTransSetDescription (row->trans, new_desc);
             }
@@ -1038,8 +1029,11 @@ gnc_gen_trans_edit_fields (GtkMenuItem *menuitem, GNCImportMainMatcher *info)
 
             if (info->edit_memo)
             {
+                guint64 style = g_strcmp0 (new_memo, row->orig_memo) ?
+                    PANGO_STYLE_ITALIC : PANGO_STYLE_NORMAL;
                 gtk_tree_store_set (store, &row->iter,
                                     DOWNLOADED_COL_MEMO, new_memo,
+                                    DOWNLOADED_COL_MEMO_STYLE, style,
                                     -1);
                 xaccSplitSetMemo (row->split, new_memo);
             }
@@ -1060,7 +1054,7 @@ gnc_gen_trans_reset_edits_cb (GtkMenuItem *menuitem, GNCImportMainMatcher *info)
     GtkTreeModel *model;
     GtkTreeStore *store;
     GtkTreeSelection *selection;
-    GList *selected_rows;
+    GList *selected_rows, *row_info_list;
 
     g_return_if_fail (info != NULL);
     ENTER("gnc_gen_trans_reset_edits_cb");
@@ -1086,6 +1080,8 @@ gnc_gen_trans_reset_edits_cb (GtkMenuItem *menuitem, GNCImportMainMatcher *info)
         gtk_tree_store_set (store, &rowinfo->iter,
                             DOWNLOADED_COL_DESCRIPTION, rowinfo->orig_desc,
                             DOWNLOADED_COL_MEMO, rowinfo->orig_memo,
+                            DOWNLOADED_COL_DESCRIPTION_STYLE, PANGO_STYLE_NORMAL,
+                            DOWNLOADED_COL_MEMO_STYLE, PANGO_STYLE_NORMAL,
                             -1);
         rowinfo_free (rowinfo);
     };
@@ -1186,7 +1182,7 @@ gnc_gen_trans_view_popup_menu (GtkTreeView *treeview,
     GtkTreeModel *model;
     GtkTreeSelection *selection;
     GList *selected_rows;
-    const char *desc, *memo, *notes;
+    const char *desc = NULL, *memo = NULL, *notes = NULL;
     gboolean has_edits = FALSE;
 
     ENTER ("");
@@ -1346,6 +1342,12 @@ add_text_column (GtkTreeView *view, const gchar *title, int col_num, gboolean el
     else
         gtk_tree_view_column_set_sort_column_id (column, col_num);
 
+    if (col_num == DOWNLOADED_COL_DESCRIPTION)
+        gtk_tree_view_column_add_attribute (column, renderer, "style", DOWNLOADED_COL_DESCRIPTION_STYLE);
+
+    if (col_num == DOWNLOADED_COL_MEMO)
+        gtk_tree_view_column_add_attribute (column, renderer, "style", DOWNLOADED_COL_MEMO_STYLE);
+
     g_object_set (G_OBJECT(column),
                   "reorderable", TRUE,
                   "resizable", TRUE,
@@ -1389,8 +1391,9 @@ gnc_gen_trans_init_view (GNCImportMainMatcher *info,
     view = info->view;
     store = gtk_tree_store_new (NUM_DOWNLOADED_COLS, G_TYPE_STRING, G_TYPE_INT64,
                                 G_TYPE_STRING, G_TYPE_STRING, G_TYPE_DOUBLE,
-                                G_TYPE_STRING, G_TYPE_STRING, G_TYPE_STRING,
-                                G_TYPE_STRING, G_TYPE_STRING, G_TYPE_BOOLEAN,
+                                G_TYPE_STRING, G_TYPE_STRING, G_TYPE_UINT64, //description stuff
+                                G_TYPE_STRING, G_TYPE_STRING, G_TYPE_UINT64, //memo stuff
+                                G_TYPE_STRING, G_TYPE_BOOLEAN,
                                 G_TYPE_BOOLEAN, G_TYPE_BOOLEAN, G_TYPE_STRING,
                                 GDK_TYPE_PIXBUF, G_TYPE_POINTER, G_TYPE_STRING,
                                 G_TYPE_BOOLEAN);
@@ -1720,9 +1723,19 @@ update_child_row (GNCImportMatchInfo *sel_match, GtkTreeModel *model, GtkTreeIte
     ro_text = xaccPrintAmount (xaccSplitGetAmount (sel_match->split),
                                gnc_split_amount_print_info (sel_match->split, TRUE));
 
-    gtk_tree_store_set (store, &child, DOWNLOADED_COL_AMOUNT, ro_text, -1);
-    gtk_tree_store_set (store, &child, DOWNLOADED_COL_MEMO, memo, -1);
-    gtk_tree_store_set (store, &child, DOWNLOADED_COL_DESCRIPTION, desc, -1);
+    gtk_tree_store_set (store, &child,
+                        DOWNLOADED_COL_AMOUNT, ro_text,
+                        -1);
+
+    gtk_tree_store_set (store, &child,
+                        DOWNLOADED_COL_MEMO, memo,
+                        DOWNLOADED_COL_MEMO_STYLE, PANGO_STYLE_NORMAL,
+                        -1);
+
+    gtk_tree_store_set (store, &child,
+                        DOWNLOADED_COL_DESCRIPTION, desc,
+                        DOWNLOADED_COL_DESCRIPTION_STYLE, PANGO_STYLE_NORMAL,
+                        -1);
 
     gtk_tree_store_set (store, &child, DOWNLOADED_COL_ENABLE, FALSE, -1);
     g_free (text);
--- a/gnucash/import-export/import-main-matcher.c
+++ b/gnucash/import-export/import-main-matcher.c
@@ -894,6 +894,30 @@ typedef enum
     NOTES,
 } edit_field;
 
+typedef struct
+{
+    Split *split;
+    Transaction *trans;
+    GtkTreeIter iter;
+} RowInfo;
+
+static void rowinfo_free (RowInfo* info)
+{
+    g_free (info);
+}
+
+static RowInfo * row_get_info (gpointer row, GNCImportMainMatcher *info)
+{
+    GNCImportTransInfo *trans_info;
+    GtkTreeModel *model = gtk_tree_view_get_model (info->view);
+    RowInfo *retval = g_new (RowInfo, 1);
+    gtk_tree_model_get_iter (model, &retval->iter, row);
+    gtk_tree_model_get (model, &retval->iter, DOWNLOADED_COL_DATA, &trans_info, -1);
+    retval->trans = gnc_import_TransInfo_get_trans (trans_info);
+    retval->split = gnc_import_TransInfo_get_fsplit (trans_info);
+    return retval;
+}
+
 static void
 gnc_gen_trans_edit_fields (GtkMenuItem *menuitem, GNCImportMainMatcher *info,
                            edit_field field)
@@ -901,12 +925,9 @@ gnc_gen_trans_edit_fields (GtkMenuItem *menuitem, GNCImportMainMatcher *info,
     GtkTreeView *treeview;
     GtkTreeSelection *selection;
     GtkTreeModel *model;
-    GList *selected_rows;
-    GList *refs = NULL;
+    GList *selected_rows, *row_info_list;
     GtkTreeStore* store;
-    GNCImportTransInfo *trans_info;
     Transaction* trans;
-    GtkTreeIter iter;
 
     g_return_if_fail (info != NULL);
     ENTER("assign_transfer_account_to_selection_cb");
@@ -923,18 +944,8 @@ gnc_gen_trans_edit_fields (GtkMenuItem *menuitem, GNCImportMainMatcher *info,
         return;
     }
 
-    if (selected_rows->next)
-    {
-        LEAVE ("User selected multiple rows, not supported");
-        return;
-    }
-
-    g_return_if_fail (gtk_tree_model_get_iter (model, &iter,
-                                               selected_rows->data));
-
-    gtk_tree_model_get (model, &iter, DOWNLOADED_COL_DATA,
-                        &trans_info, -1);
-    trans = gnc_import_TransInfo_get_trans (trans_info);
+    row_info_list = gnc_g_list_map (selected_rows, (GncGMapFunc) row_get_info, info);
+    trans = ((RowInfo*)row_info_list->data)->trans;
 
     switch (field)
     {
@@ -945,24 +956,33 @@ gnc_gen_trans_edit_fields (GtkMenuItem *menuitem, GNCImportMainMatcher *info,
                                             _("Enter new Description"),
                                             xaccTransGetDescription (trans));
             if (!new_field) break;
-            xaccTransSetDescription (trans, new_field);
-            gtk_tree_store_set (store, &iter, DOWNLOADED_COL_DESCRIPTION,
-                                new_field, -1);
+            for (GList *n = row_info_list; n; n = g_list_next (n))
+            {
+                RowInfo *info = n->data;
+                xaccTransSetDescription (info->trans, new_field);
+                gtk_tree_store_set (store, &info->iter,
+                                    DOWNLOADED_COL_DESCRIPTION, new_field,
+                                    -1);
+            }
             g_free (new_field);
             break;
         }
         case MEMO:
         {
-            Split *first_split =
-                gnc_import_TransInfo_get_fsplit (trans_info);
+            Split *first_split = xaccTransGetSplit (trans, 0);
             char *new_field =
                 gnc_input_dialog_with_entry(info->main_widget, "",
                                             _("Enter new Memo"),
                                             xaccSplitGetMemo (first_split));
             if (!new_field) break;
-            xaccSplitSetMemo (first_split, new_field);
-            gtk_tree_store_set (store, &iter,
-                                DOWNLOADED_COL_MEMO, new_field, -1);
+            for (GList *n = row_info_list; n; n = g_list_next (n))
+            {
+                RowInfo *info = n->data;
+                xaccSplitSetMemo (info->split, new_field);
+                gtk_tree_store_set (store, &info->iter,
+                                    DOWNLOADED_COL_MEMO, new_field,
+                                    -1);
+            }
             g_free (new_field);
             break;
         }
@@ -973,11 +993,16 @@ gnc_gen_trans_edit_fields (GtkMenuItem *menuitem, GNCImportMainMatcher *info,
                                             _("Enter new Notes"),
                                             xaccTransGetNotes (trans));
             if (!new_field) break;
-            xaccTransSetNotes (trans, new_field);
+            for (GList *n = row_info_list; n; n = g_list_next (n))
+            {
+                RowInfo *info = n->data;
+                xaccTransSetNotes (info->trans, new_field);
+            }
             g_free (new_field);
             break;
         }
     }
+    g_list_free_full (row_info_list, (GDestroyNotify)rowinfo_free);
     g_list_free_full (selected_rows, (GDestroyNotify)gtk_tree_path_free);
     LEAVE("");
 }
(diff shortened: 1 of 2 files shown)


christopherlam:
Ok tested on real data and it's not yet selecting the right transactions: see attached file (it must be renamed to OFX): Import OFX, sort by Description column, highlight all transactions whose descriptions are DEPOSIT, rename to SALARY, and it will rename some invalid transactions. I suspect the gtk_tree_model is sorting the list prematurely thereby offering undesirable iters for modification?

[Data_export.txt](https://github.com/Gnucash/gnucash/files/9077516/Data_export.txt)

christopherlam:
I can anticipate a future request to store these renames to auto rename future imports... This would be a major project to provide a reasonable UI for it.

christopherlam:
and another feature: during import, the desc/notes/memo will be store internally; if they are edited during import, will allow them to be reset back as originally imported. The "Reset edits" menu is only enabled if any selection desc/notes/memo differs from the original.

image

Bob-IT:
> @Bob-IT any way to annotate the cells via markup (eg italics) to indicate a modified field?

What I think you would need to do is add a new uint field to the store, connect that to the style property of the cell renderer of the tree column you want to control, populate the model with the default value of 'Normal' and then change this value to 'Italics' on the row required.

Bob-IT:
@christopherlam 
While you are in this area, does it make sense to have three dialogs to edit Description/Memo/Notes ?
I would of thought one menu entry that brings up a single dialog with three Gtk entries would be more efficient.
Your multi selection change would still work by setting the sensitivity of the non matching entries to false.
I do not use this so ignore if existing option is best.

christopherlam:
@Bob-IT any way to annotate the cells via markup (eg italics) to indicate a modified field?

> What I think you would need to do is add a new uint field to the store, connect that to the style property of the cell renderer of the tree column you want to control, populate the model with the default value of 'Normal' and then change this value to 'Italics' on the row required.

In other words: too difficult!

> @christopherlam
> While you are in this area, does it make sense to have three dialogs to edit Description/Memo/Notes ?
> I would of thought one menu entry that brings up a single dialog with three Gtk entries would be more efficient.
> Your multi selection change would still work by setting the sensitivity of the non matching entries to false.
> I do not use this so ignore if existing option is best.

Seems to work... but the code (even if there is clear separation ui & data processing) starts to stink...

image

jralls:
> While you are in this area, does it make sense to have three dialogs to edit Description/Memo/Notes ?
I would of thought one menu entry that brings up a single dialog with three Gtk entries would be more efficient.
Your multi selection change would still work by setting the sensitivity of the non matching entries to false.
I do not use this so ignore if existing option is best.

Couldn't you decorate the string with [pango markup](https://docs.gtk.org/Pango/pango_markup.html) instead? Or is that not enabled in the text cell editor?

christopherlam:
> Couldn't you decorate the string with [pango markup](https://docs.gtk.org/Pango/pango_markup.html) instead? Or is that not enabled in the text cell editor?

No, "<i>%s<i>" doesn't work.

Bob-IT:
> In other words: too difficult!

@christopherlam I will pull this and have a look.

Bob-IT:
@christopherlam 
Here is a [gist](https://gist.github.com/Bob-IT/60e06d317cab58d2c918d83ae035a837) of my changes.
Check over this, especially when I change to italic for correctness.
You will see I have also changed the dialog to glade, much easier to tweek.

One question, if I change one row description from 'INTEREST' to 'NONE', and then include it with another row that has description of 'INTEREST', should that allow multi change as I have already changed it ?

Bob-IT:
> Couldn't you decorate the string with [pango markup](https://docs.gtk.org/Pango/pango_markup.html) instead? Or is that not enabled in the text cell editor?

I think you would have to do something similar and connect the GtkCellRendererText markup property to a model column instead of the text property. Did not try that but Chris can if he thinks it more appropriate.

christopherlam:
> @christopherlam Here is a [gist](https://gist.github.com/Bob-IT/60e06d317cab58d2c918d83ae035a837) of my changes. Check over this, especially when I change to italic for correctness. You will see I have also changed the dialog to glade, much easier to tweek.

Thank you, works well!

> One question, if I change one row description from 'INTEREST' to 'NONE', and then include it with another row that has description of 'INTEREST', should that allow multi change as I have already changed it ?

My view (IMHO) is if you change INTEREST to NONE, then it doesn't belong to the INTEREST logical group anymore, and selecting it together with other INTEREST transactions shouldn't allow multichange. I suspect this was an issue with your gist lines 237 and 250, and I think keeping my conditional seems to work as it should.

christopherlam:
A suppose a further development is to switch from a GtkEntry dropdown to a GtkComboBox which offers a list of candidate strings from the existing transactions. But not for maint.

christopherlam:
Ok I think the last commit has the right approach -- set cell to italic if it differs from original imported string.

christopherlam:
> A suppose a further development is to switch from a GtkEntry dropdown to a GtkComboBox which offers a list of candidate strings from the existing transactions. But not for maint.

I tried this change -- instead of GtkEntry, use GtkComboboxText to populate from existing account data. It works but is extremely unwieldly: the dropdown from GtkComboboxText is slow. So I think best drop it.

image

```patch
diff --git a/gnucash/gtkbuilder/dialog-import.glade b/gnucash/gtkbuilder/dialog-import.glade
index f7e3d001e..b9809129f 100644
--- a/gnucash/gtkbuilder/dialog-import.glade
+++ b/gnucash/gtkbuilder/dialog-import.glade
@@ -917,7 +917,7 @@
         </child>
         <child>
           <!-- n-columns=2 n-rows=3 -->
-          <object class="GtkGrid">
+          <object class="GtkGrid" id="transaction_edit_grid">
             <property name="row-spacing">3</property>
@@ -959,37 +959,13 @@
               </packing>
             </child>
             <child>
-              <object class="GtkEntry" id="desc_entry">
-                <property name="visible">True</property>
-                <property name="can-focus">True</property>
-                <property name="hexpand">True</property>
-              </object>
-              <packing>
-                <property name="left-attach">1</property>
-                <property name="top-attach">0</property>
-              </packing>
+              <placeholder/>
             </child>
             <child>
-              <object class="GtkEntry" id="notes_entry">
-                <property name="visible">True</property>
-                <property name="can-focus">True</property>
-                <property name="hexpand">True</property>
-              </object>
-              <packing>
-                <property name="left-attach">1</property>
-                <property name="top-attach">1</property>
-              </packing>
+              <placeholder/>
             </child>
             <child>
-              <object class="GtkEntry" id="memo_entry">
-                <property name="visible">True</property>
-                <property name="can-focus">True</property>
-                <property name="hexpand">True</property>
-              </object>
-              <packing>
-                <property name="left-attach">1</property>
-                <property name="top-attach">2</property>
-              </packing>
+              <placeholder/>
             </child>
           </object>
           <packing>
diff --git a/gnucash/import-export/import-main-matcher.c b/gnucash/import-export/import-main-matcher.c
index db00c9a68..f7534e2db 100644
--- a/gnucash/import-export/import-main-matcher.c
+++ b/gnucash/import-export/import-main-matcher.c
@@ -919,6 +919,11 @@ static RowInfo * row_get_info (gpointer row, GNCImportMainMatcher *info)
     return retval;
 }
 
+static void add_to_combobox_text (gpointer key, gpointer user_data)
+{
+    gtk_combo_box_text_append_text (GTK_COMBO_BOX_TEXT (user_data), (char*)key);
+}
+
 static gboolean
 input_new_fields (GtkWidget *parent, RowInfo *info, GtkTreeStore *store,
                   gboolean edit_desc, gboolean edit_notes, gboolean edit_memo,
@@ -928,24 +933,75 @@ input_new_fields (GtkWidget *parent, RowInfo *info, GtkTreeStore *store,
     GtkWidget  *dialog;
     GtkBuilder *builder;
     gboolean    retval = FALSE;
+    GHashTable *desc_hash, *notes_hash, *memo_hash;
+    GList      *keys;
 
     builder = gtk_builder_new ();
     gnc_builder_add_from_file (builder, "dialog-import.glade", "transaction_edit_dialog");
 
     dialog = GTK_WIDGET(gtk_builder_get_object (builder, "transaction_edit_dialog"));
+    grid = GTK_WIDGET(gtk_builder_get_object (builder, "transaction_edit_grid"));
 
-    desc_entry = GTK_WIDGET(gtk_builder_get_object (builder, "desc_entry"));
-    memo_entry = GTK_WIDGET(gtk_builder_get_object (builder, "memo_entry"));
-    notes_entry = GTK_WIDGET(gtk_builder_get_object (builder, "notes_entry"));
+    desc_hash = g_hash_table_new (g_str_hash, g_str_equal);
+    notes_hash = g_hash_table_new (g_str_hash, g_str_equal);
+    memo_hash = g_hash_table_new (g_str_hash, g_str_equal);
 
-    gtk_widget_set_sensitive (desc_entry, edit_desc);
-    gtk_entry_set_text (GTK_ENTRY (desc_entry), xaccTransGetDescription (info->trans));
+    for (GList *n = xaccAccountGetSplitList (xaccSplitGetAccount (info->split));
+         n; n = n->next)
+    {
+        const Split *s = n->data;
+        const Transaction *t = xaccSplitGetParent (s);
+        const gpointer one = GINT_TO_POINTER (1);
+        const gchar *key;
+
+        key = xaccTransGetDescription (t);
+        if (key && *key)
+            g_hash_table_insert (desc_hash, (gpointer)key, one);
+
+        key = xaccTransGetNotes (t);
+        if (key && *key)
+            g_hash_table_insert (notes_hash, (gpointer)key, one);
+
+        key = xaccSplitGetMemo (s);
+        if (key && *key)
+            g_hash_table_insert (memo_hash, (gpointer)key, one);
+    };
 
+    desc_entry = gtk_combo_box_text_new_with_entry ();
+    gtk_widget_set_sensitive (desc_entry, edit_desc);
+    g_hash_table_remove (desc_hash, xaccTransGetDescription (info->trans));
+    keys = g_hash_table_get_keys (desc_hash);
+    keys = g_list_sort (keys, (GCompareFunc)g_strcmp0);
+    g_list_foreach (keys, (GFunc)add_to_combobox_text, desc_entry);
+    g_list_free (keys);
+    g_hash_table_destroy (desc_hash);
+    gtk_combo_box_text_prepend_text (GTK_COMBO_BOX_TEXT (desc_entry), xaccTransGetDescription (info->trans));
+    gtk_combo_box_set_active (GTK_COMBO_BOX (desc_entry), 0);
+    gtk_grid_attach (GTK_GRID (grid), desc_entry, 1, 0, 1, 1);
+
+    notes_entry = gtk_combo_box_text_new_with_entry ();
     gtk_widget_set_sensitive (notes_entry, edit_notes);
-    gtk_entry_set_text (GTK_ENTRY (notes_entry), xaccTransGetNotes (info->trans));
-
+    g_hash_table_remove (notes_hash, xaccTransGetNotes (info->trans));
+    keys = g_hash_table_get_keys (notes_hash);
+    keys = g_list_sort (keys, (GCompareFunc)g_strcmp0);
+    g_list_foreach (keys, (GFunc)add_to_combobox_text, notes_entry);
+    g_list_free (keys);
+    g_hash_table_destroy (notes_hash);
+    gtk_combo_box_text_prepend_text (GTK_COMBO_BOX_TEXT(notes_entry), xaccTransGetNotes (info->trans));
+    gtk_combo_box_set_active (GTK_COMBO_BOX (notes_entry), 0);
+    gtk_grid_attach (GTK_GRID (grid), notes_entry, 1, 1, 1, 1);
+
+    memo_entry = gtk_combo_box_text_new_with_entry ();
     gtk_widget_set_sensitive (memo_entry, edit_memo);
-    gtk_entry_set_text (GTK_ENTRY (memo_entry), xaccSplitGetMemo (info->split));
+    g_hash_table_remove (memo_hash, xaccSplitGetMemo (info->split));
+    keys = g_hash_table_get_keys (memo_hash);
+    keys = g_list_sort (keys, (GCompareFunc)g_strcmp0);
+    g_list_foreach (keys, (GFunc)add_to_combobox_text, memo_entry);
+    g_list_free (keys);
+    g_hash_table_destroy (memo_hash);
+    gtk_combo_box_text_prepend_text (GTK_COMBO_BOX_TEXT(memo_entry), xaccSplitGetMemo (info->split));
+    gtk_combo_box_set_active (GTK_COMBO_BOX (memo_entry), 0);
+    gtk_grid_attach (GTK_GRID (grid), memo_entry, 1, 2, 1, 1);
 
     gtk_window_set_transient_for (GTK_WINDOW (dialog), GTK_WINDOW (parent));
 
@@ -955,9 +1011,9 @@ input_new_fields (GtkWidget *parent, RowInfo *info, GtkTreeStore *store,
     switch (gtk_dialog_run (GTK_DIALOG(dialog)))
     {
     case GTK_RESPONSE_OK:
-        *new_desc = g_strdup (gtk_entry_get_text (GTK_ENTRY (desc_entry)));
-        *new_notes = g_strdup (gtk_entry_get_text (GTK_ENTRY (notes_entry)));
-        *new_memo = g_strdup (gtk_entry_get_text (GTK_ENTRY (memo_entry)));
+        *new_desc = g_strdup (gtk_combo_box_text_get_active_text (GTK_COMBO_BOX_TEXT (desc_entry)));
+        *new_notes = g_strdup (gtk_combo_box_text_get_active_text (GTK_COMBO_BOX_TEXT (notes_entry)));
+        *new_memo = g_strdup (gtk_combo_box_text_get_active_text (GTK_COMBO_BOX_TEXT (memo_entry)));
         retval = TRUE;
         break;
     default:
```

Bob-IT:
> Ok I think the last commit has the right approach -- set cell to italic if it differs from original imported string.

Yes that works, did you want to do any thing with the old tooltip value?
So that just leaves the keyboard accelerator keys from what I have noticed.

Bob-IT:
> I tried this change -- instead of GtkEntry, use GtkComboboxText to populate from existing account data. It works but is extremely unwieldly: the dropdown from GtkComboboxText is slow. So I think best drop it.

In what way, slow to load or use ?
How many entries do they have ?

christopherlam:
> > Ok I think the last commit has the right approach -- set cell to italic if it differs from original imported string.
> 
> Yes that works, did you want to do any thing with the old tooltip value? So that just leaves the keyboard accelerator keys from what I have noticed.

IMHO old tooltips should show the ofx import data. This is debatable.

> > I tried this change -- instead of GtkEntry, use GtkComboboxText to populate from existing account data. It works but is extremely unwieldly: the dropdown from GtkComboboxText is slow. So I think best drop it.
> 
> In what way, slow to load or use ? How many entries do they have ?

Multiple scrolling pages of GtkComboboxText don't quite work. Using GtkEntry with completion seems to work very well now :-D

Bob-IT:
> IMHO old tooltips should show the ofx import data.

That is OK, at least it was thought about.

gnucash/import-export/import-main-matcher.c
```diff
@@ -885,26 +894,118 @@ gnc_gen_trans_assign_transfer_account_to_selection_cb (GtkMenuItem *menuitem,
...
+    gtk_entry_completion_set_model (completion, GTK_TREE_MODEL(list));
+    gtk_entry_completion_set_text_column (completion, 0);
+    gtk_entry_set_completion (GTK_ENTRY (entry), completion);
+}
+
+static gboolean
+input_new_fields (GtkWidget *parent, RowInfo *info, GtkTreeStore *store,
+                  gboolean edit_desc, gboolean edit_notes, gboolean edit_memo,
+                  GHashTable *desc_hash, GHashTable *notes_hash, GHashTable *memo_hash,
+                  char **new_desc, char **new_notes, char **new_memo)
```
jralls, line 957: Four of the parameters at the call site come from the GncMainMatcher* so maybe just take that as a parameter to cut down the size of the parameter list; wrapping the hash tables in a struct (see comment about creating that once) will help too.

gnucash/import-export/import-main-matcher.c
```diff
@@ -921,81 +1022,130 @@ gnc_gen_trans_edit_fields (GtkMenuItem *menuitem, GNCImportMainMatcher *info,
...
+            if (key && *key)
+                g_hash_table_insert (desc_hash, (gpointer)key, one);
+
+            key = xaccTransGetNotes (t);
+            if (key && *key)
+                g_hash_table_insert (notes_hash, (gpointer)key, one);
+
+            key = xaccSplitGetMemo (s);
+            if (key && *key)
+                g_hash_table_insert (memo_hash, (gpointer)key, one);
```
jralls, line 1056: Separate this out and do it as part of the matcher setup and add a hash entry if the user creates a new description/note/memo.

While you're at it create a struct with all three hash tables that you can pass around.

jralls (changes requested):
Looks generally OK.

gnucash/import-export/import-main-matcher.c
christopherlam: > Separate this out and do it as part of the matcher setup and add a hash entry if the user creates a new description/note/memo.
> 
> While you're at it create a struct with all three hash tables that you can pass around.

Agree about doing in the matcher setup but do you mean creating a signal to watch for new transactions?

jralls: Hmm, hadn't thought about that, I was thinking of using the dialog to create a new description/note/memo. Monitoring the QofEvents for `xaccTransCommitEdit` would be better, yes. I don't think you need to create anything, just catch and act on what's there.

christopherlam: > Hmm, hadn't thought about that, I was thinking of using the dialog to create a new description/note/memo. Monitoring the QofEvents for `xaccTransCommitEdit` would be better, yes. I don't think you need to create anything, just catch and act on what's there.

Don't think I'll add a signal... no idea how it works. The import matcher does not have ability to add a new transaction while it's running, therefore I'm not sure it's necessary to update the desc/notes/memo hash tables during the matcher lifetime. The code currently written will populate GtkListStore with the desc/notes/memo with *existing* data, plus the current field contents. So, if you've renamed a desc from "INTEREST" to "NONE", and re-edit it, the autocomplete will include NONE rather than INTEREST. I think this is a reasonable approach.

jralls: > if the user launches import matcher, clicks behind into the register, adds some more transactions, then the hash could be updated.

OK, probably a rare occurrence; it's more likely in the reconcile step. 

> So, if you've renamed a desc from "INTEREST" to "NONE", and re-edit it, the autocomplete will include NONE rather than INTEREST. I think this is a reasonable approach.

Erm, not quite. Suppose you have previous transactions with a description of "Interest" and you edit a batch of imported transactions' descriptions to "Interesting", a string that wasn't previously used. You now select a different batch of transactions and edit the descriptions. "Interesting" will not be in the list, and some user is going to file a bug saying that it should be.

christopherlam: Ok see very last commit -- editing a field causes the new text to be added into the hash. This means the hash will keep growing during a matcher session; editing "Interest" to "Interesting" then re-edit to "Interested" means both are now suggestions, until they are reset in the next matcher session.

christopherlam: > Ok see very last commit -- editing a field causes the new text to be added into the hash. This means the hash will keep growing during a matcher session; editing "Interest" to "Interesting" then re-edit to "Interested" means both are now suggestions, until they are reset in the next matcher session.

Oops doesn't quite work: The hash table keys are pointers to original string data, and an edited text points to garbage. Must be a rethink... Maybe a separate 'new_desc' hash.

gnucash/import-export/import-main-matcher.c
```diff
@@ -1100,10 +1105,16 @@ gnc_gen_trans_edit_fields (GtkMenuItem *menuitem, GNCImportMainMatcher *info)
                                     DOWNLOADED_COL_DESCRIPTION_STYLE, style,
                                     -1);
                 xaccTransSetDescription (row->trans, new_desc);
+                info->edits = g_list_prepend (info->edits, (gpointer)CACHE_INSERT (new_desc));
```
jralls, line 1108: I don't think the string cache adds any value here vs. just `strdup`ing the string.

jralls:
That last commit message should be mnemonics not accelerators. Accelerators are key combinations that invoke the action with out navigating the menu, mnemonics are a way to navigate the menu from the keyboard.

That aside looks fine to me. @Bob-IT any further comments?

Bob-IT:
@christopherlam 
These are my last thoughts in [gist](https://gist.github.com/Bob-IT/20fd53876a5635099aa7c1766007047b)
Basically they are, glade changes for keyboard mnemonics,
Do not use numbers for the model columns, create an enum so they can be named, you may like to change them to something other than 'key' to something more meaningful.
Lastly I created a custom completion function so any text entered can be found, not just starting text.

This last one I am not sure off as I do not have the data in my test file, give it a try and if you want to keep it you may like to add another model field that is stored case-folded and use that in the function instead of deriving it.

christopherlam:
> @christopherlam These are my last thoughts in [gist](https://gist.github.com/Bob-IT/20fd53876a5635099aa7c1766007047b) Basically they are, glade changes for keyboard mnemonics, Do not use numbers for the model columns, create an enum so they can be named, you may like to change them to something other than 'key' to something more meaningful. Lastly I created a custom completion function so any text entered can be found, not just starting text.
> 
> This last one I am not sure off as I do not have the data in my test file, give it a try and if you want to keep it you may like to add another model field that is stored case-folded and use that in the function instead of deriving it.

That's neat. If the `model_key` is normalized and casefolded, we might as well store it in the GtkListStore in `populate_list` which is called for every existing string *once* per dialog, rather than for every keystroke when editing fields.

Bob-IT:
> That's neat. If the `model_key` is normalized and casefolded, we might as well store it in the GtkListStore in `populate_list` which is called for every existing string _once_ per dialog, rather than for every keystroke when editing fields.

Yep, that is what I was trying to say at the end of the comment.

christopherlam:
> > That's neat. If the `model_key` is normalized and casefolded, we might as well store it in the GtkListStore in `populate_list` which is called for every existing string _once_ per dialog, rather than for every keystroke when editing fields.
> 
> Yep, that is what I was trying to say at the end of the comment.

With this last commit you're now just spoiling the users 😂

gnucash/import-export/import-main-matcher.c
```diff
@@ -1394,6 +1394,9 @@ gnc_gen_trans_view_popup_menu (GtkTreeView *treeview,
     gtk_menu_popup_at_pointer (GTK_MENU(menu), (GdkEvent*)event);
 
     g_list_free_full (selected_rows, (GDestroyNotify)gtk_tree_path_free);
+
+    g_object_take_ref (G_OBJECT(menu));
+    g_object_unref (G_OBJECT(menu));
```
christopherlam, line 1399: Any more `g_object_unref` fails. So, I presume this is the right approach to fix this leaking `menu` object.

christopherlam:
For maint, I cannot think of any further development ideas. Any further concerns before merging in?

There is an unresolved leak in import-main-matcher.c: the widget created with `menu = gtk_menu_new();` is never `g_object_unref`fed, and cannot be unreffed because it's floating; sinking it seems to be difficult.

Bob-IT:
@christopherlam 
Does this help, add after the `gtk_menu_new`
 `gtk_menu_attach_to_widget (GTK_MENU(menu), GTK_WIDGET(treeview), NULL);`
The menu should be destroyed when the treeview is destroyed.

Bob-IT:
Have not built win version for a while, will have a look tomorrow.

jralls:
You're a bit hamstrung on debugging because MSYS2 strips the symbols, but a check of the g_utf8_collate [code](https://gitlab.gnome.org/GNOME/glib/-/blob/main/glib/gunicollate.c#L111). This being Windows I'd expect it to take the wchar_t branch where it converts the strings to UTF16 and calls [wcscoll](https://en.cppreference.com/w/c/string/wide/wcscoll) on them, but I suppose it's possible that it doesn't in which case it calls `strcmp` directly. In either case strcmp will crash if a NULL gets through to it.

You don't call `g_utf8_collate`, you use [`g_utf8_collate_key`](https://gitlab.gnome.org/GNOME/glib/-/blob/main/glib/gunicollate.c#L382) that creates a hash for later sorting. The `g_utf8_collate` is probably coming from somewhere in the GtkTreeModel so you wouldn't be able to disable the Unicode handling. Unfortunately since MSYS2 strips the symbols you won't be able to get a better backtrace even from gdb. Can you test the same import into the same book on Linux? If it crashes there you can install the symbols for Gtk and GLib and find out what are the strings causing the crash and where they're coming from.

christopherlam:
Preliminary testing using the same datafile and OFX shows Linux doesn't crash. I'm not convinced an empty string is at fault; the hash table keys being fed to UTF8 are never `NULL` or `""`, see the numerous `(key && *key)` before insertion into hash. These are my findings so far. Will try again.

christopherlam:
(small feedback using my moderate sized datafile) storing desc_hash in the main struct may need to be changed; the GtkListStore sorting is a significant slowdown whenever opening the input dialog. I'd modify to the following:

- when creating the main matcher, load desc/notes/memo hash as before
- convert hash to GList, perform utf8 sorting; destroy hash
- when creating input dialog, copy GList onto GtkListStore, prepend the existing string to be modified.

Bob-IT:
> (small feedback using my moderate sized datafile) storing desc_hash in the main struct may need to be changed; the GtkListStore sorting is a significant slowdown whenever opening the input dialog.

Can you give some figures here, how many entries in the list stores, in `setup_entry` add this after the `populate_list` function call
`g_print("list store size is %d\n", gtk_tree_model_iter_n_children (GTK_TREE_MODEL(list), NULL));`

What is the slowdown?

christopherlam:
In my book, Desc has 1016 unique strings, Notes 31, Memo 471 strings.

Here's some logging in setup_entry; the 0.82+0.01+0.39 = 1.22s is very noticeable, and instaneous with #1383

```patch
@@ -1022,6 +1022,7 @@ setup_entry (GtkWidget *entry, gboolean sensitive, GHashTable *hash,
 {
     GtkEntryCompletion* completion;
     GtkListStore *list;
+    float starttime = (float)clock()/CLOCKS_PER_SEC;
 
     gtk_entry_set_text (GTK_ENTRY (entry), sensitive ? initial : _("Disabled"));
     gtk_widget_set_sensitive (entry, sensitive);
@@ -1029,12 +1030,16 @@ setup_entry (GtkWidget *entry, gboolean sensitive, GHashTable *hash,
         return;
 
     list = gtk_list_store_new (3, G_TYPE_STRING, G_TYPE_STRING, G_TYPE_STRING);
+    PWARN ("start: %f", (float)clock()/CLOCKS_PER_SEC - starttime);
     g_hash_table_foreach (hash, (GHFunc)populate_list, list);
+    PWARN ("after_hash: %f", (float)clock()/CLOCKS_PER_SEC - starttime);
     if (!g_hash_table_lookup (hash, (gpointer)initial))
         populate_list ((gpointer)initial, NULL, list);
+    PWARN ("after_adding_initial: %f", (float)clock()/CLOCKS_PER_SEC - starttime);
     gtk_tree_sortable_set_sort_column_id (GTK_TREE_SORTABLE (list),
                                           COMPLETION_LIST_COLLATED,
                                           GTK_SORT_ASCENDING);
+    PWARN ("after_sort: %f", (float)clock()/CLOCKS_PER_SEC - starttime);
 
     completion = gtk_entry_completion_new ();
     gtk_entry_completion_set_model (completion, GTK_TREE_MODEL(list));
```

```
* 17:24:07  WARN <gnc.import.main-matcher> [setup_entry()] start: 0.000498
* 17:24:07  WARN <gnc.import.main-matcher> [setup_entry()] after_hash: 0.030794
* 17:24:07  WARN <gnc.import.main-matcher> [setup_entry()] after_adding_initial: 0.031015
* 17:24:08  WARN <gnc.import.main-matcher> [setup_entry()] after_sort: 0.822418

* 17:24:08  WARN <gnc.import.main-matcher> [setup_entry()] start: 0.000368
* 17:24:08  WARN <gnc.import.main-matcher> [setup_entry()] after_hash: 0.001245
* 17:24:08  WARN <gnc.import.main-matcher> [setup_entry()] after_adding_initial: 0.001360
* 17:24:08  WARN <gnc.import.main-matcher> [setup_entry()] after_sort: 0.010481

* 17:24:08  WARN <gnc.import.main-matcher> [setup_entry()] start: 0.000347
* 17:24:08  WARN <gnc.import.main-matcher> [setup_entry()] after_hash: 0.012516
* 17:24:08  WARN <gnc.import.main-matcher> [setup_entry()] after_adding_initial: 0.012709
* 17:24:08  WARN <gnc.import.main-matcher> [setup_entry()] after_sort: 0.390121
```

Bob-IT:
Just to let you know, I have built on win and can crash it, will investigate further.

christopherlam:
> Just to let you know, I have built on win and can crash it, will investigate further.

Just a thought: if changing the sort key from `collated` to `original` in the `gtk_tree_sortable_set_sort_column_id` call fixes the crash, then it would suggest that the gtk sorting is calling g_utf8_collate rather than g_strcmp0 as I thought.

Bob-IT:
@christopherlam 
I think I have found the problem and pushed a change.
The first was the wrong type was used, I had used G_TYPE_UINT64 when it should really be G_TYPE_INT.
The second was the use of COMPLETION_LIST_COLLATED for the sort function, I have changed this to COMPLETION_LIST_ORIGINAL.

I have tested and all seems to work OK with updating fields and resetting them. Whether this has any baring on the lag you have seen I do not know but maybe you can get new figures as above. If there is still a lag then hopefully your other branch can be used, I create a CSV file that I imported into a new book which had some random description, notes and memos and it seemed OK but maybe not random enough.

christopherlam:
> I think I have found the problem and pushed a change. The first was the wrong type was used, I had used G_TYPE_UINT64 when it should really be G_TYPE_INT.

Was this change independently sufficient to fix the crash?

> I have tested and all seems to work OK with updating fields and resetting them. Whether this has any baring on the lag you have seen I do not know but maybe you can get new figures as above. If there is still a lag then hopefully your other branch can be used, I create a CSV file that I imported into a new book which had some random description, notes and memos and it seemed OK but maybe not random enough.

Further benchmarks sorting on the original strings, using my personal book on an underpowered machine -- ORIGINAL
```
* 23:11:47  WARN <gnc.import.main-matcher> [setup_entry()] --------
* 23:11:47  WARN <gnc.import.main-matcher> [setup_entry()] start: 0.000950
* 23:11:47  WARN <gnc.import.main-matcher> [setup_entry()] after hash: 0.011900
* 23:11:47  WARN <gnc.import.main-matcher> [setup_entry()] after adding initial: 0.012016
* 23:11:47  WARN <gnc.import.main-matcher> [setup_entry()] after sort: 0.110613

* 23:11:47  WARN <gnc.import.main-matcher> [setup_entry()] --------
* 23:11:47  WARN <gnc.import.main-matcher> [setup_entry()] start: 0.000401
* 23:11:47  WARN <gnc.import.main-matcher> [setup_entry()] after hash: 0.000830
* 23:11:47  WARN <gnc.import.main-matcher> [setup_entry()] after adding initial: 0.000910
* 23:11:47  WARN <gnc.import.main-matcher> [setup_entry()] after sort: 0.001896

* 23:11:47  WARN <gnc.import.main-matcher> [setup_entry()] --------
* 23:11:47  WARN <gnc.import.main-matcher> [setup_entry()] start: 0.000303
* 23:11:47  WARN <gnc.import.main-matcher> [setup_entry()] after hash: 0.005239
* 23:11:47  WARN <gnc.import.main-matcher> [setup_entry()] after adding initial: 0.005335
* 23:11:47  WARN <gnc.import.main-matcher> [setup_entry()] after sort: 0.053068
```

COLLATED:
```
* 23:13:46  WARN <gnc.import.main-matcher> [setup_entry()] --------
* 23:13:46  WARN <gnc.import.main-matcher> [setup_entry()] start: 0.001017
* 23:13:46  WARN <gnc.import.main-matcher> [setup_entry()] after hash: 0.025776
* 23:13:46  WARN <gnc.import.main-matcher> [setup_entry()] after adding initial: 0.025942
* 23:13:47  WARN <gnc.import.main-matcher> [setup_entry()] after sort: 0.827160

* 23:13:47  WARN <gnc.import.main-matcher> [setup_entry()] --------
* 23:13:47  WARN <gnc.import.main-matcher> [setup_entry()] start: 0.000402
* 23:13:47  WARN <gnc.import.main-matcher> [setup_entry()] after hash: 0.001137
* 23:13:47  WARN <gnc.import.main-matcher> [setup_entry()] after adding initial: 0.001213
* 23:13:47  WARN <gnc.import.main-matcher> [setup_entry()] after sort: 0.009834

* 23:13:47  WARN <gnc.import.main-matcher> [setup_entry()] --------
* 23:13:47  WARN <gnc.import.main-matcher> [setup_entry()] start: 0.000324
* 23:13:47  WARN <gnc.import.main-matcher> [setup_entry()] after hash: 0.011585
* 23:13:47  WARN <gnc.import.main-matcher> [setup_entry()] after adding initial: 0.011677
* 23:13:47  WARN <gnc.import.main-matcher> [setup_entry()] after sort: 0.330976
```

Conclusion we can keep the GtkListStore but don't need the collated key anymore.

Bob-IT:
> Was this change independently sufficient to fix the crash?

I started with a pull of maint and built it, ran it and could get it to crash especially when resetting.
Commented out the sort function and my completion function for elimination, built and still would crash.
This was what led me to changing the type, it was strange as the error I was getting implied the `gtk_tree_store_set` function did not have the terminating `-1` and possibly an invalid column, changed the type, built and it did not crash.
Uncommented the sort function, built and would crash but if I remember correctly only on multiple changes, a single change and reset would work.
Changed the sort function, built and all seemed to be OK, added back the completion function, built and still OK.

christopherlam:
> Uncommented the sort function, built and would crash but if I remember correctly only on multiple changes, a single change and reset would work. 

Thanks. `g_utf8_collated_key` creates completely incomprehensible strings from plain ascii. The intermittent crash suggests that Gtk sorting would compare the collated strings with `g_utf8_collate` (which may not handle these strings gracefully) instead of `g_strcmp0`.

jralls:
`g_utf8_collated_key` creates hashes that are guaranteed to sort in the correct order with `strcmp` but are not guaranteed to be valid utf8 so passing them to `g_utf8_collate` is likely to cause problems. If you're creating a hidden column to sort on and populating it with the results of `g_utf8_collated_key` then telling it to sort on `COMPLETION_LIST_ORIGINAL` is correct and should be substantially faster than using the actual strings and `COMPLETION_LIST_COLLATE`.

As for using `guint64 foo = strcmp(bar, baz)`, `strcmp` will return a negative number if `bar` is lexically less than `baz`. If you don't know why casting that to an unsigned is a bad idea it's time to learn about [two's complement](https://en.wikipedia.org/wiki/Two%27s_complement) binary number representation.

christopherlam:
The next ideas in this line of work are

1. the GtkEntryCompletion sorting - instead of alphabetical (with the UTF8 madness), we can sort via descending popularity. Moreover the completion CellRenderer could show, in addition to the string itself, the number of occurrences. IMHO it's nice to see "Walmart (258 occurrences)" "Walmart Refund (2 occurrences)" etc in the completion.
2. GtkEntryCompletion also used in the regular register (!!)

Bob-IT:
> As for using `guint64 foo = strcmp(bar, baz)`, `strcmp` will return a negative number if `bar` is lexically less than `baz`. 

This got me thinking, did I miss something as I had considered that to be wrong at the time, the actual line is...
`guint64 style = g_strcmp0 (new_desc, row->orig_desc) ? PANGO_STYLE_ITALIC : PANGO_STYLE_NORMAL;`
which I remember testing to check it was OK, values were always 0 or 2.
Just to reexamine the failure, I rebuilt maint from before my fix which failed as expected.
Did the same thing in commenting out sort function and completion function, still failed.
Split the `gtk_tree_store_set` into setting the text and setting the style, crashed and by looking at a log it was the style update that caused the crash.
Split the style update into two, and it is here I get two errors which implied the `gtk_tree_store_set` function did not have the terminating `-1` and possibly an invalid column.

So it turns out another fix would of been to do `(guint64)PANGO_STYLE_NORMAL` but changing to `G_TYPE_INT` I think was still the right fix.

jralls:
OK, so the correct type is actually PangoStyle, an enum. The base type of enums in C is int and enums can take negative values--although that one doesn't--so the comment about two's complement still applies. gint is better than unsigned anything but it would be better still to use the right type, PangoStyle. If and only if the compiler complains about the enum should you cast `style` to gint in the `gtk_tree_store_set` call.

christopherlam:
Tiny issue that I've encountered -- if the import includes both NEW and MATCHED transactions, we cannot include MATCHED transactions in multi-select. This is probably designed to avoid modifying existing matched transaction data. See https://github.com/christopherlam/gnucash/blob/b727b8740462e10c5b697db37f685f03070483f0/gnucash/import-export/import-main-matcher.c#L1264 

However we may want to edit the existing txn data from the import matcher.

jralls:
I think the present behavior is right. Editing an existing transaction is outside of the import matcher's responsibility. Besides, changing it may well require much deeper surgery because IIUC the match existing transactions and select transfer account/edit transaction text fields are separate.

PR #1375, opened 2022-07-17
start plugging leaks from qof_instance_get

christopherlam:
`qof_instance_get` comes from `g_object_get` which returns copies and must be freed via `g_free` or `g_object_unref`. see https://docs.gtk.org/gobject/method.Object.get.htmle-Object-Type.html#g-object-get

ditto `gtk_tree_model_get`.

This means the signatures for `gnc_import_get_split_online_id` and friends are wrong -- they're not `const` and must be `g_free`d after use. I guess fixing this means changing the signature throughout?


jralls:
Good catch. LGTM.

PR #1377, opened 2022-07-17
[Master] Remove depreciated GtkUIManager and friends from RecWin (Experiment)

Bob-IT:
I was wondering how the new menu actions fitted together with the possible follow up removing all the depreciated GtkUIManager and friends. I think the main application will need to be changed to GtkApplication but as an experiment the Reconcile Window could be used to try.
This commit is my first attempt at this, I may of missed using a new function or two but it all seems to work. The main difference is the lack of icons on the menus as below left, the space is for radio/toggle buttons...

recwin

From [here](https://docs.gtk.org/gtk3/class.ImageMenuItem.html) it is possible to create them but do we need them?
As an aside, I recently created a Fedora36 VM and did a build and the default was not to show them.

### DIFF
--- a/gnucash/gnome/window-reconcile.c
+++ b/gnucash/gnome/window-reconcile.c
@@ -1910,7 +1910,10 @@ recnWindowWithBalance (GtkWidget *parent, Account *account, gnc_numeric new_endi
         GtkAccelGroup *accel_group = gtk_accel_group_new ();
         const gchar *ui = "/org/gnucash/ui/gnc-reconcile-window.ui";
         GError *error = NULL;
-
+#ifdef MAC_INTEGRATION
+        GtkosxApplication *theApp = g_object_new (GTKOSX_TYPE_APPLICATION,
+                                                  NULL);
+#endif
         recnData->builder = gtk_builder_new ();
 
         gtk_builder_add_from_resource (recnData->builder, ui, &error);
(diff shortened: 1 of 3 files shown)


jralls:
Shouldn't this PR also remove gnucash/ui/gnc-reconcile-window-ui.xml?

The current way to put icons on menu items is described at the bottom of [menu design](https://developer.gnome.org/GMenu/documentation/tutorials/menus.html). That uses GMenu that we'll have to convert to before migrating to Gtk4 (and yes, IIUC that requires that we also use GApplication). In the meantime the [Gtk documentation](https://docs.gtk.org/gtk3/class.ImageMenuItem.html) says
> If you want to display an icon in a menu item, you should use GtkMenuItem and pack a GtkBox with a GtkImage and a GtkLabel instead. You should also consider using GtkBuilder and the XML GMenu description for creating menus, by following the [GMenu guide][https://developer.gnome.org/GMenu/]. You should consider using icons in menu items only sparingly, and for “objects” (or “nouns”) elements only, like bookmarks, files, and links; “actions” (or “verbs”) should not have icons.

That said I don't think that icons contribute much to the UX. Apple agrees, having removed icons from menus years ago. On MSWin a menu item is [either text or bitmap, not both](https://docs.microsoft.com/en-us/windows/win32/menurc/using-menus#using-menu-item-bitmaps).

jralls:
Overall I think this is a good step forward.

fellen:
> I don't think that icons contribute much to the UX

At least they are very helpful,if you are not very familiar with the selected language.
How will we find the exit while checking for RTL issues?

Bob-IT:
> Shouldn't this PR also remove gnucash/ui/gnc-reconcile-window-ui.xml?

@jralls 
Yes it should, just missed it along with adding 'ui' files to be translated.
Thanks for the macOS menu bar changes, that was a last minute comment out when I realised the mac CI would fail.
I see you linked to the same gtk documentation I had about menu items.
I am also inclined to get rid of all the menu icons, it makes it so much easier to create the menus.

Bob-IT:
> > I don't think that icons contribute much to the UX
> 
> At least they are very helpful,if you are not very familiar with the selected language. How will we find the exit while checking for RTL issues?

When I have done RTL testing, I resort to having a Fedora VM with an RTL language user running GnuCash and my development VM running Gnucash along side so I can navigate the menus easily.

Bob-IT:
Closing this as I have moved commits to PR #1456 so there is only one PR dealing with the removal of GtkUIManager.

PR #1382, opened 2022-07-21
[gnc-plugin-page-sx-list] enable popup for sx editor

christopherlam:
Enable right-click menu for SX editor.


jralls:
Use-case?

christopherlam:
Just to harmonise UI. A treeview begs to be R-clicked. And the toolbar seems far away.

jralls:
OK for master, it's time to stabilize maint, it's less than 9 months to releasing 5.0.

christopherlam:
into master.

PR #1388, opened 2022-07-25
import matcher

christopherlam:
further work --?

- sort by decreasing frequency
- show n(occurences) for dropdown

image

### DIFF
--- a/gnucash/import-export/import-main-matcher.c
+++ b/gnucash/import-export/import-main-matcher.c
@@ -127,8 +127,6 @@ enum downloaded_cols
 /*static QofLogModule log_module = GNC_MOD_IMPORT;*/
 static QofLogModule log_module = G_MOD_IMPORT_MATCHER;
 
-static const gpointer one = GINT_TO_POINTER (1);
-
 void on_matcher_ok_clicked (GtkButton *button, GNCImportMainMatcher *info);
 void on_matcher_cancel_clicked (GtkButton *button, gpointer user_data);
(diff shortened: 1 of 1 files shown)


jralls:
Thinking a bit more about this, I think the drop-down with completion candidates is a nice UX improvement over the keep typing until the match is what you like in the register's quickfill cell. I also think that quickfill should work the same across GnuCash. How do you feel about adding the dropdown and sorting to gnucash/register/register-gnome/quickfillcell-gnome.c?

BTW, we're now less than 6 months away from beginning beta testing for GnuCash 5, so it's time for new features to go to master and to focus on stabilizing maint.

christopherlam:
> Thinking a bit more about this, I think the drop-down with completion candidates is a nice UX improvement over the keep typing until the match is what you like in the register's quickfill cell. I also think that quickfill should work the same across GnuCash. How do you feel about adding the dropdown and sorting to gnucash/register/register-gnome/quickfillcell-gnome.c?

This would require a per-account hashtable of desc/notes/memo candidates... aka the old friend `qof-string-cache` :laughing:

> BTW, we're now less than 6 months away from beginning beta testing for GnuCash 5, so it's time for new features to go to master and to focus on stabilizing maint.

jralls:
> This would require a per-account hashtable of desc/notes/memo candidates... aka the old friend qof-string-cache 😆

Nope, not qof_string_cache. It works only on whole strings, has strings from all over the place, and doesn't know where they come from. For this you need a collection of just the entries from the one field for for the splits in the account in question--for typeahead in the register that would be the descriptions for that account's splits' parent transactions.

Regardless, it's no different from what you've done here.

PR #1391, opened 2022-07-27
Mac tests

fellen:
I got
```
mac-tests: .github#L1
The macOS-10.15 environment is deprecated, consider switching to macos-11(macos-latest), macos-12 instead. For more details see https://github.com/actions/virtual-environments/issues/5583
```
on my last real commit.

With this change the test succeeds again.

I don't know, if you have to do updates on the Mac bundle, too.


jralls:
I've made these fixes in a slightly different way.

PR #1396, opened 2022-07-30
[gnucash-style] free dimensions properly

christopherlam:
Previous code was trying to reference count dimensions upon destroy. However it was buggy: when refcount==1, both `style_dimensions_destroy` and `gnucash_sheet_style_destroy` were decrementing refcount, therefore the it never reached zero to trigger g_table_destroy and g_free.

The dimensions are held as GHashTable values, and were eventually g_freed; but the GTable was never destroyed.

### DIFF
--- a/gnucash/register/register-gnome/gnucash-sheet.c
+++ b/gnucash/register/register-gnome/gnucash-sheet.c
@@ -2521,9 +2521,8 @@ gnucash_sheet_new (Table *table)
     sheet->item_editor = gnc_item_edit_new (sheet);
 
     /* some register data */
-    sheet->dimensions_hash_table = g_hash_table_new_full (g_int_hash,
-                                   g_int_equal,
-                                   g_free, g_free);
+    sheet->dimensions_hash_table = g_hash_table_new_full (g_int_hash, g_int_equal,
+                                                          g_free, NULL);
 
     /* add tooltips to sheet */
     gtk_widget_set_has_tooltip (GTK_WIDGET(sheet), TRUE);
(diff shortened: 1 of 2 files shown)


jralls:
If your analysis is correct then ignoring the ref counting and not freeing the other reference is wrong. Correct is to implement `block_dimensions_ref()` and `block_dimensions_unref()` to manage the ref count; `block_dimensions_unref` frees the BlockDimensions object when the ref count goes to 0. Everything else calls `block_dimensions_ref()` when copying a BlockDimensions* and `block_dimensions_unref()` when it's done with its copy.

Making this more or less automatic is why object systems like GObject and C++'s std::shared_ptr exist.

christopherlam:
Even better will be to remove refcount altogether, I've only ever seen 2 dimensions being created per session.

christopherlam:
My issue with implementing refcounting is that the block dimensions don't have a record of their use as GHashTable values. Too difficult.

jralls:
> Even better will be to remove refcount altogether, I've only ever seen 2 dimensions being created per session.

If there's *any* shared ownership then it needs to use ref counting. That said, shared ownership is a code smell so if you can arrange things so that it's not needed that's indeed better.

jralls:
> My issue with implementing refcounting is that the block dimensions don't have a record of their use as GHashTable values. Too difficult.

If doing it right is too hard for you then add a FIXME comment or write a bug and leave the rest alone. Substituting one flavor of wrong for another isn't an improvement.

PR #1397, opened 2022-07-31
Update gnc-vcs-info

vuw-ecs-kevin:
As per email to gnucash-devel

Subject: util/gnc-vcs-info fails if building from a Git worktree (with possible solution)

### DIFF
--- a/util/gnc-vcs-info
+++ b/util/gnc-vcs-info
@@ -147,6 +147,47 @@ then
   fi
 fi
 
+# Maybe it's a git worktree?
+real_gitdir="${real_srcdir}"/.git
+if test -f "${real_gitdir}"
+then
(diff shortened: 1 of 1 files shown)


jralls:
Duplicating the whole block isn't necessary, you could just add the extra test to the first git check. But that seems unneccesary too because it's already `if test -d "${real_gitdir}" || test -e "$real_gitdir}"`. That `-e` makes both the `-d` and your addition of `-f` superfluous.

That second test [was added[(https://github.com/Gnucash/gnucash/commit/32a196122f33dbe342a29c57a5bbcefa5a8410ed) 5 years ago with the commit message "Properly detect git in case of linked worktree". Are you perhaps patching an old tarball?

PR #1399, opened 2022-08-01
use gtk_entry_get_text instead of gtk_editable_get_chars

christopherlam:
The former returns a `const char*` which does not need to be freed. I think this is self-explanatory.


jralls:
Good catch!

PR #1403, opened 2022-08-08
some cleanups in pricedb, and don't ref/unref from reports

christopherlam:
IMHO reports must access prices passively, therefore gnc-price-ref/unref possibly shouldn't be allowed.


libgnucash/engine/gnc-pricedb.c
```diff
@@ -698,60 +698,38 @@ compare_prices_by_date(gconstpointer a, gconstpointer b)
                          gnc_price_get_guid((GNCPrice *) b));
 }
 
-typedef struct
+static gint
+find_dupl_price (const GNCPrice *a, const GNCPrice *b)
```
jralls, line 702: `price_compare` or maybe `prcmp` if you want to pun `strcmp`. `find` implies that it's traversing a price list looking for a duplicate.

jralls (changes requested):
You need to think a little harder about the price reference counting in the portfolio reports. It's not about the reports mutating the prices, it's about the reports not crashing GnuCash with a use-after-free if the user deletes a price from the UI while the report is computing its results.

christopherlam:
> You need to think a little harder about the price reference counting in the portfolio reports. It's not about the reports mutating the prices, it's about the reports not crashing GnuCash with a use-after-free if the user deletes a price from the UI while the report is computing its results.

Ok I can see this risk, however it's pervasive across all reports and all objects - Account* Split* etc all liable to be nullified while a report is running. This risk is minimised by the whole app being single-threaded, and the guile exception handler will sternly inform the user that the data being modified during report generation leads to a crash. Wouldn't it be nicer to implement a data locking mechanism to start before launching a report, then release when complete? I think that only GNCPrice being protected this way is a sore point.

jralls:
> This risk is minimised by the whole app being single-threaded

The whole app isn't single threaded: Guile and webkit are both threaded. Plus anything that updates the progress bar runs the main loop every time it calls for a progress bar update so even functions on the main thread can be paused allowing other stuff, including queued user interactions, to happen.

> the guile exception handler will sternly inform the user that the data being modified during report generation leads to a crash

The Guile exception handler will never see it, it's not a guile exception. The program will get a `SIGSEGV` from the OS and quit. Game over.

> Wouldn't it be nicer to implement a data locking mechanism to start before launching a report, then release when complete? I think that only GNCPrice being protected this way is a sore point.

Yes, it's a sore point and it's a major failing in @derekatkins s conversion of the engine classes to GObject 18 years ago. GObject has a built-in ref-counting mechanism but Derek chose to not use it. @gjanssens  and I decided when we embarked on the C++ porting several years ago that we'd fix that as part of converting the engine to C++ to avoid having to rewrite everything twice. But that's not a good reason to gratuitously break the one case where memory management is being done right.

As for data-locking, you're confusing two completely separate principles, shared access and object lifetime management. The former is needed in a multithreaded environment to prevent race conditions; the latter is always needed when working with a non-garbage-collected language.

I'll stipulate that it could be done better, but there's no point in rewriting gnc-pricedb.c until we're ready to make it a C++ class.

PR #1408, opened 2022-08-15
Consider aqbanking's ultimateDebtor field when processing SEPA transactions.

hoehermann:
Some banks now use the payment processor's name in the remoteName field with the actual recipient being available in the new ultimateDebtor field. For details, see the mailing lists:

* https://lists.gnucash.org/pipermail/gnucash-de/2022-June/012426.html
* https://mailman.aqbanking.de/private/aqbanking-user/2022-August/thread.html

This patch works for me, but I cannot test it with other banks.

### DIFF
--- a/gnucash/import-export/aqb/gnc-ab-utils.c
+++ b/gnucash/import-export/aqb/gnc-ab-utils.c
@@ -414,7 +414,24 @@ gnc_ab_get_remote_name (const AB_TRANSACTION *ab_trans)
 
     g_return_val_if_fail (ab_trans, NULL);
 
+#if AQBANKING_VERSION_INT >= 60200
+    /* Prefer the ultimateDebtor, if available.
+     * Happens when processing payments done via VISA debit cards. For details,
+     * see https://lists.gnucash.org/pipermail/gnucash-de/2022-June/012426.html
+     * and https://mailman.aqbanking.de/private/aqbanking-user/2022-August/thread.html
+     * 
+     * This is only relevant when processing transactions received via SEPA.
+     * Not relevant when using gnc_ab_get_remote_name for user input validation.
+     */
+    ab_remote_name = AB_Transaction_GetUltimateDebtor (ab_trans);
+    if (!ab_remote_name)
+    {
+        // No ultimateDebtor given, fall back to the plain remoteName.
+        ab_remote_name = AB_Transaction_GetRemoteName (ab_trans);
+    }
+#else
     ab_remote_name = AB_Transaction_GetRemoteName (ab_trans);
+#endif
     if (ab_remote_name)
 #ifdef AQBANKING6
         gnc_other_name = g_strdup(ab_remote_name);


gnucash/import-export/aqb/gnc-ab-utils.c
```diff
@@ -414,7 +414,24 @@ gnc_ab_get_remote_name (const AB_TRANSACTION *ab_trans)
 
     g_return_val_if_fail (ab_trans, NULL);
 
+#if AQBANKING_VERSION_MAJOR >= 6 && AQBANKING_VERSION_MINOR >= 2
```
fellen, line 417: #if (AQBANKING_VERSION_INT >= 60200)

fellen:
This approach was discussed in the thread starting with
https://lists.gnucash.org/pipermail/gnucash-de/2022-June/012426.html
and continued in August starting with
https://lists.gnucash.org/pipermail/gnucash-de/2022-August/012495.html

There is also a discussion in
[Bug 798385 - Description to often only "Landesbank Hessen-Thuringen Girozentrale"](https://bugs.gnucash.org/show_bug.cgi?id=798385) and 
[Bug 797877 - Fields "ultimateDebtor" and "ultimateCreditor" are not used to create a description of a transaction](https://bugs.gnucash.org/show_bug.cgi?id=797877) of more  complex solutions.

But as currently nobody is volunteering to adjust the assistant, this PR is a first step.

jralls:
The update needs to check both `AB_Transaction_UlitimateCreditor` and `AB_Transaction_UlitimateDebtor`. I think it's very unlikely that both will be set so the result can go to the same local variable.

Based on the bug reports and those two mailing list discussions it appears to me that UltimateCreditor/Debtor is the most significant field and so should go at the front instead of the end. I'm not so sure about RemoteName; I'm not even convinced that it or TransactionText are even still emitted as those were part of the old Deutsche Kreditwirtschaft scheme used in HBCI. They're not even part of SWIFT, they're inserted as `:NS:` tags where NS stands for Not SWIFT.
However we might as well continue to check for them and just pass an empty string (`""`) to the format if they're not present:

```
gnc_ab_description_to_gnc:
char * ultimate_name = gnc_ab_ulitimate_creditor_debtor_to_gnc (ab_trans);
if (!ultimate_name)
    ultimate_name = "";
if (!description)
    description = "";
if (!other_name)
    other_name = "";
retval = g_strdup_printf("%s %s %s", ultimate_name, description, other_name);
```

hoehermann:
There has been a lengthy discussion about incorporating `ultimateCreditor` into this pull request. With the transactions happening in my civic life, the field is never set. This means, I would not be able to test the changes with real-world data. That is why I did not include `ultimateCreditor` in this pull request. Also I do not see a reasonable place where to put it. I simply added to the function `gnc_ab_get_remote_name`, which is used for describing the _destination_ of the transfer. However, `ultimateCreditor` denotes the _source_ of the transfer. As far as I can tell, the transaction importer does not even make this distinction.

Judging from the discussions and my personal experiences, as soon as `ultimateCreditor` is present, `remoteName` does not contain useful information. As a non-accountant, I find myself quickly overwhelmed with the information in the transaction descriptions. As a programmer, I want to reduce the clutter. That is why I decided against text concatenation. Instead, I rather prefer `ultimateCreditor` over `remoteName`. I hope to have this pull request accepted so users can try it and share their experiences. If all goes well, we can always add the `ultimateCreditor` in another pull request. :)

jralls:
> There has been a lengthy discussion about incorporating ultimateCreditor into this pull request.

Really? Where was that? 
If it was on the AQBanking list, that's closed to non-subscribers. It sure wasn't in the gnucash-de thread, there Martin told you to use both. As @fellen  pointed out we have a bug report specifically requesting that we include UltimateCreditor.

Have you ever gotten a refund to your card account? Wouldn't the bank in that case use the UltimateCreditor field to indicate who issued the refund?

As for RemoteName I think it very unlikely that an MT940 stream will include both it and Ultimate*, but considering how many banks do their own thing with SWIFT specs I can't rule it out completely and neither can you.

hoehermann:
> Really? Where was that [discussion]?

Private Mail between Frank and me.

> As fellen pointed out we have a bug report specifically requesting that we include UltimateCreditor.

Yes, that is https://bugs.gnucash.org/show_bug.cgi?id=797877. In the very same discussion, you wrote "Seems unlikely that tacking it to the end of the already lengthy concatenation (…) would be useful" – that is one reason why I went for the either-or-approach. Based on what I read in the mailing lists and bug reports, this is most likely to help many users without breaking anything for others.

> Have you ever gotten a refund to your card account?

No. But maybe someone out there does, knows how it is presented by their bank and come up with a solution which works for them. Hopefully, they share it with us.

> As for RemoteName I think it very unlikely that an MT940 stream will include both it and Ultimate*, but considering how many banks do their own thing with SWIFT specs I can't rule it out completely and neither can you.

And since we cannot know, I opt for simply giving it a go to see how this patch works for the users.

I do not know anything about MT940 or SWIFT specs or PAIN documentation. I am a simple man. I noticed the problem and came up with a solution that "works for me"™. The change is small and is unlikely to have serious drawbacks. If you want a more complete solution, we can use the one Michael posted in https://bugs.gnucash.org/show_bug.cgi?id=797877. If neither solution is deemed "good", we should probably find out what kind of solution we are looking for.

jralls:
>  In the very same discussion, you wrote "Seems unlikely that tacking it to the end of the already lengthy concatenation (…) would be useful" – that is one reason why I went for the either-or-approach.

It's still at the end. I think it needs to go first. Both bug reports complain that their banks put a bunch of useless boiler plate in the purpose strings.

> Private Mail between Frank and me.
> I do not know anything about MT940 or SWIFT specs or PAIN documentation. 

OK. I've been working on this with Martin--an expert in the subject matter, as he's been writing AQBanking for 20 years-- and I actually took the time to study the relevant specs and some other documentation that Martin pointed me at. Thanks for your efforts, but I'll write it my way.

PR #1409, opened 2022-08-16
Bug 797725 - Untranslatable string "For Period Covering ~a to ~a"

christopherlam:
use `(gnc:format)` strings.

### DIFF
--- a/bindings/guile/core-utils.scm
+++ b/bindings/guile/core-utils.scm
@@ -111,7 +111,7 @@
 (set-exception-printer! 'unbound-variable print-unbound-variable-error)
 
 ;; format.
-(define %regex (make-regexp "[$][{]([[:alnum:]]+)[}]"))
+(define %regex (make-regexp "[$][{]([[:alnum:]\\-]+)[}]"))
 (define (gnc:format str . bindings)
   (define hash (make-hash-table))
   (define (substitute m)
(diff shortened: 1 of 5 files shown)


fellen:
I wonder if we should create a report-msgs.scm or similar to collect such constructs.

jralls:
It sets the report title so it's more boilerplate than message. That aside it doesn't really merit a separate file just for itself, but if there's a bunch of common generation funcs or fragments that get copied from one report to the next, extracting them all to a common file would be a Good Thing™.

christopherlam:
The named arg syntax can be changed from `${var}` to `{var}` to match c++ fmt if desired.

jralls:
> The named arg syntax can be changed from ${var} to {var} to match c++ fmt if desired.

 I don't think that's terribly important and requiring `${` as an intro makes it much less likely that you'd ever need to provide an escape like {fmt}'s provision of `{{` and `}}` to insert curly brace literals into the format string.

jralls:
@christopherlam why did you decide to move this to master? It's not a major change.

christopherlam:
Can't remember much-- may be because of string freeze. I haven't changed
the strings generally because I hadn't received any feedback about it.

jralls:
There wasn't any string freeze on 18 August when you applied the `master` label nor on 6 Sept when you changed the branch. The latest version is fine and can go into maint immediately after the release.

christopherlam:
Forgot about this PR. So should we apply this for final release, or into master?

christopherlam:
merged in maint, 580efe091c

PR #1414, opened 2022-08-25
Stock Assistant: specialises stock split transactions.

christopherlam:
Stock splits and reverse stock splits are handled via a specialised stock_amount input. Instead of inputting new units, input new balance. Some error checks are included. This means the assistant is designed to generalise stock activity, yet has a special case. :man_shrugging: 

image

### DIFF
--- a/gnucash/gnome/assistant-stock-transaction.cpp
+++ b/gnucash/gnome/assistant-stock-transaction.cpp
@@ -113,6 +113,7 @@ FieldMask operator ^(FieldMask lhs, FieldMask rhs)
 struct TxnTypeInfo
 {
     FieldMask stock_amount;
+    bool input_new_balance;
     FieldMask stock_value;
     FieldMask cash_value;
     FieldMask fees_value;
@@ -131,6 +132,7 @@ static const TxnTypeVec starting_types
 {
     {
         FieldMask::ENABLED_DEBIT,          // stock_amt
+        false,                             // input_new_balance
         FieldMask::ENABLED_DEBIT,          // stock_val
         FieldMask::ENABLED_CREDIT,         // cash_amt
         FieldMask::ENABLED_DEBIT | FieldMask::ALLOW_ZERO,          // fees_amt
@@ -144,6 +146,7 @@ static const TxnTypeVec starting_types
     },
     {
         FieldMask::ENABLED_CREDIT,         // stock_amt
+        false,                             // input_new_balance
         FieldMask::ENABLED_CREDIT,         // stock_val
         FieldMask::ENABLED_DEBIT,          // cash_amt
         FieldMask::ENABLED_DEBIT | FieldMask::ALLOW_ZERO,          // fees_amt
@@ -161,6 +164,7 @@ static const TxnTypeVec long_types
 {
     {
         FieldMask::ENABLED_DEBIT,          // stock_amt
+        false,                             // input_new_balance
         FieldMask::ENABLED_DEBIT,          // stock_val
         FieldMask::ENABLED_CREDIT,         // cash_amt
         FieldMask::ENABLED_DEBIT | FieldMask::ALLOW_ZERO,          // fees_amt
@@ -174,6 +178,7 @@ static const TxnTypeVec long_types
     },
     {
         FieldMask::ENABLED_CREDIT,         // stock_amt
+        false,                             // input_new_balance
         FieldMask::ENABLED_CREDIT,         // stock_val
         FieldMask::ENABLED_DEBIT,          // cash_amt
         FieldMask::ENABLED_DEBIT | FieldMask::ALLOW_ZERO,          // fees_amt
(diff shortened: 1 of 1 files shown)


christopherlam:
I think the Stock Ratio calculator must be skipped: if the current stock balance is 100 units, and they announce a 1:3 reverse stock split with cash-in-lieu, then the new balance will be 33 units, with 1/3 units returned in cash. The stock ratio would be displayed as 33/100.

jralls:
Sounds good. What's the problem? That it's not 33/99?

christopherlam:
> Sounds good. What's the problem? That it's not 33/99?

That's right. The true ratio is 1/3 but this cannot be determined by the calculator purely by considering `old_bal = 100`, `new_bal = 33` - the broker fee and the cash-in-lieu are not available yet.

jralls:
True. But so what? That number is only user entertainment. Nothing in GnuCash needs the ratio. The whole ratio thing came up as a way to enter the transaction: You get a letter or see a news item that says XYZ did a 3:2 split yesterday. With that information and yesterday's closing price GnuCash can calculate the split amount, whether there's cash-in-lieu, and what it's likely to be. Do brokers charge a fee on stock splits in Australia? They don't here.

christopherlam:
@CDB-Man this works. The small oddity is inputting the `New Balance` requires the user *must* enter a negative balance if stock split occurs when shorting. However I think this is acceptable; `old_balance` and `new_balance` will thus both be negative. The ratio will then be positive, as expected.

jralls:
You could do a check box for short and flip the sign if it's true. OTOH I don't recall a lot of discussion about shorting on the mailing list over the years, so it's not really worth worrying about.

CDB-Man:
https://bugs.gnucash.org/show_bug.cgi?id=798004#c62
When using forward split in Stock Assistant: https://i.imgur.com/HV7IBdT.png
The output stock split transaction is broken: https://i.imgur.com/40GndaE.png

---

When using the reverse split in Stock Assistant:
https://i.imgur.com/WUD1ktd.png
https://i.imgur.com/BCrBl2O.png
The output is -150 instead of -50 in the transaction: https://i.imgur.com/bl8nqIl.png

---
Conclusion: it appears you have your signs reversed on ending balance.

gnucash/gnome/assistant-stock-transaction.cpp
```diff
@@ -704,7 +742,30 @@ to ensure proper recording."), new_date_str, last_split_date_str);
...
+        auto ratio = gnc_numeric_div (stock_amount, info->balance_at_date,
+                                      GNC_DENOM_AUTO, GNC_HOW_DENOM_REDUCE);
+        auto stock_pinfo = gnc_commodity_print_info
+            (xaccAccountGetCommodity (info->acct), true);
+        stock_amount = gnc_numeric_sub_fixed (stock_amount, info->balance_at_date);
+        line.units = xaccPrintAmount (stock_amount, stock_pinfo);
+        if (gnc_numeric_check (ratio))
+            add_error_str (errors, N_("Invalid stock new balance"));
+        else if (gnc_numeric_negative_p (ratio))
+            add_error_str (errors, N_("New and old balance must have same signs"));
```
fellen, line 762: Full stop or exclamation mark missing.

Also in the next messages!

gnucash/gnome/assistant-stock-transaction.cpp
```diff
@@ -704,7 +742,30 @@ to ensure proper recording."), new_date_str, last_split_date_str);
...
+        auto credit_side = (info->txn_type->stock_amount & FieldMask::ENABLED_CREDIT);
+        auto delta = gnc_numeric_sub_fixed (stock_amount, info->balance_at_date);
+        auto ratio = gnc_numeric_div (stock_amount, info->balance_at_date,
+                                      GNC_DENOM_AUTO, GNC_HOW_DENOM_REDUCE);
+        auto stock_pinfo = gnc_commodity_print_info
+            (xaccAccountGetCommodity (info->acct), true);
+        stock_amount = gnc_numeric_sub_fixed (stock_amount, info->balance_at_date);
+        line.units = xaccPrintAmount (stock_amount, stock_pinfo);
+        if (gnc_numeric_check (ratio))
+            add_error_str (errors, N_("Invalid stock new balance"));
```
fellen, line 760: I do not understand this message. Perhaps a verb is missing?

fellen (commented):
Some messages need improvements.

gnucash/gnome/assistant-stock-transaction.cpp
jralls: It means "The new stock balance that you entered is invalid because computing the split ratio from it caused an overflow, underflow, or divide by 0".

christopherlam: Nearly. It can't be div/0 because info->balance_at_date is never zero in stock splits. It means the balance is garbage e.g."abc", overflow/underflow, or after 1752f1d19, zero or blank "".

gnucash/gnome/assistant-stock-transaction.cpp
christopherlam: This one has been removed. It sounds silly.

PR #1421, opened 2022-09-03
gnc_g_list_stringjoin skips null

christopherlam:


christopherlam:
@jralls I wouldn't immediately merge this because I cannot test that it produces the correct output.

jralls:
> I wouldn't immediately merge this because I cannot test that it produces the correct output.

@christopherlam Did you get a bit cross-threaded between the PR and the proposed commit? The commit is a straightforward replay of one you've done a bunch of other places and shouldn't need testing. The PR title OTOH refers to a suggestion I made on IRC a couple of days ago that could make `gnc_ab_description_to_gnc` even more terse but isn't implemented here.

PR #1423, opened 2022-09-06
stock assistant - add units column, remove zero-fee line

christopherlam:
* add stock units column indicating gain/loss of number of units
* don't show Fee line if zero; a zero-fee split isn't generated.

### DIFF
--- a/gnucash/gnome/assistant-stock-transaction.cpp
+++ b/gnucash/gnome/assistant-stock-transaction.cpp
@@ -638,8 +638,6 @@ check_page (GtkListStore *list, gnc_numeric& debit, gnc_numeric& credit,
             FieldMask splitfield, Account *acct, GtkWidget *memo, GtkWidget *gae,
             gnc_commodity *comm, const char* page, StringVec& errors)
 {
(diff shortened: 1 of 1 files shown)


christopherlam:
It uses the existing string "Units" so I think it's ok to merge.

gnucash/gnome/assistant-stock-transaction.cpp
```diff
@@ -633,26 +634,55 @@ add_error_str (StringVec& errors, const char* str)
...
+
+
+static void
+check_page (SummaryLineInfo& line, gnc_numeric& debit, gnc_numeric& credit,
             FieldMask splitfield, Account *acct, GtkWidget *memo, GtkWidget *gae,
             gnc_commodity *comm, const char* page, StringVec& errors)
 {
-    if (splitfield == FieldMask::DISABLED)
-        return;
     const char* missing_str = N_("(missing)");
```
jralls, line 670: This needs a translator comment explaining what might be missing and maybe that we hope there aren't any languages with a different word for missing accounts and missing amounts.

jralls (changes requested):

PR #1427, opened 2022-09-07
Standardize the root account names of the business account template in French

bigon:
Hello,

For some reasons, the business account templates in French are using different names for the Assets and Equity root account than the other templates in French

In other language they are the same in all templates

This merge request fixes this


fellen:
Thanks!

PR #1429, opened 2022-09-08
[gnc-currency-edit] allow GtkEntry completion

christopherlam:
...thus typing 'dollar' would filter to all currencies using the dollar. It uses `gtk_bin_get_child` because `gtk_combo_box_get_child` is Gtk4 only.

image

### DIFF
--- a/gnucash/gnome-utils/gnc-currency-edit.c
+++ b/gnucash/gnome-utils/gnc-currency-edit.c
@@ -284,6 +284,12 @@ static void gnc_currency_edit_active_changed (GtkComboBox *gobject,
     }
 }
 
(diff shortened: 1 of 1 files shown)


jralls:
LGTM and works as advertised. @Bob-IT any comments?

Should we put it in master as a new feature for 5.0?

gjanssens:
I love it. It was on my todo list back when I was more active still.

I'd put it in master and focus maint purely on (important) bug fixing and possibly compatibility fixes with master (if there are any) for the last few releases.

Bob-IT:
Looks OK to me and should go to master.

PR #1431, opened 2022-09-11
Updated wording for assistant-stock-transaction.cpp

CDB-Man:
On the long side of the transaction list, dividend reinvestments with and without remainder can be merged.

On the short side of the transaction list, the 2 dividend reinvestment transactions are not relevant and should be deleted.  There is no such thing as a "compensatory dividend reinvestment".

Refer to https://bugs.gnucash.org/show_bug.cgi?id=798004#c49

### DIFF
--- a/gnucash/gnome/assistant-stock-transaction.cpp
+++ b/gnucash/gnome/assistant-stock-transaction.cpp
@@ -168,7 +168,7 @@ static const TxnTypeVec long_types
         FieldMask::DISABLED,               // dividend_amt
         FieldMask::DISABLED,               // capg_amt
         // Translators: this is a stock transaction describing
-        // purchase of stock.
+        // new purchase of stock.
         N_("Buy"),
         N_("Buying stock long.")
     },
@@ -180,8 +180,8 @@ static const TxnTypeVec long_types
         false,                  // fees_capitalize
         FieldMask::DISABLED,               // dividend_amt
         FieldMask::ENABLED_CREDIT | FieldMask::ALLOW_ZERO | FieldMask::ALLOW_NEGATIVE, // capgains_amt
-        // Translators: this is a stock transaction describing sale of
-        // stock, and recording capital gain/loss
+        // Translators: this is a stock transaction describing new
+        // sale of stock, and recording capital gain/loss
         N_("Sell"),
         N_("Selling stock long, and record capital gain/loss.")
     },
@@ -285,8 +285,8 @@ static const TxnTypeVec long_types
         FieldMask::DISABLED,               // dividend_amt
         FieldMask::ENABLED_CREDIT | FieldMask::ALLOW_ZERO | FieldMask::ALLOW_NEGATIVE, // capgains_amt
         // Translators: this is a stock transaction describing a
-        // reverse split. Some fractional stock is returned as cash.
-        N_("Reverse split with cash in lieu for fractional unit remainders"),
+        // reverse split. Some remainder stock units are returned as cash.
+        N_("Reverse split with cash in lieu for remainder units"),
         N_("Company redeems units, thereby increasing the stock price by a multiple, while keeping the total monetary value of the overall investment constant. Any remainder units are sold/redeemed for cash, with a capital gain/loss realized.")
     },
 };
@@ -420,9 +420,9 @@ static const TxnTypeVec short_types
         FieldMask::DISABLED,               // dividend_amt
         FieldMask::ENABLED_DEBIT | FieldMask::ALLOW_ZERO | FieldMask::ALLOW_NEGATIVE,          // capg_amt
         // Translators: this is a stock transaction describing a
-        // reverse split when shorting stock. Fractional remaining
-        // stock is retrieved as cash.
-        N_("Reverse split with cash in lieu for fractionals"),
+        // reverse split when shorting stock. Remainder stock
+        // units are retrieved as cash.
+        N_("Reverse split with cash in lieu remainder units"),
         N_("Company redeems units, thereby increasing the stock price by a multiple, while keeping the total monetary value of the overall investment constant. Any remainder units are closed and covered with a cash payment, with a capital gain/loss realized.")
     },
 };
--- a/gnucash/gnome/assistant-stock-transaction.cpp
+++ b/gnucash/gnome/assistant-stock-transaction.cpp
@@ -262,7 +262,7 @@ static const TxnTypeVec long_types
         // Translators: this is a stock transaction describing a stock
         // split
         N_("Stock split"),
-        N_("Company issues additional units, thereby reducing the stock price by a divisor, while keeping the $ value of shares constant.")
+        N_("Company issues additional units, thereby reducing the stock price by a divisor, while keeping the total monetary value the overall investment constant.")
     },
     {
         FieldMask::ENABLED_CREDIT,         // stock_amt
@@ -274,7 +274,7 @@ static const TxnTypeVec long_types
         FieldMask::DISABLED,               // capg_amt
         // Translators: this is a stock transaction describing a reverse split
         N_("Reverse split"),
-        N_("Company redeems units, thereby increasing the stock price by a multiple, while keeping the $ value of shares constant.")
+        N_("Company redeems units, thereby increasing the stock price by a multiple, while keeping the monetary value of the overall investment constant.")
     },
     {
         FieldMask::ENABLED_CREDIT,         // stock_amt
@@ -287,7 +287,7 @@ static const TxnTypeVec long_types
         // Translators: this is a stock transaction describing a
         // reverse split. Some fractional stock is returned as cash.
         N_("Reverse split with cash in lieu for fractional unit remainders"),
-        N_("Company redeems units, thereby increasing the stock price by a multiple, while keeping the $ value of shares constant. Any remainder units are sold/redeemed for cash, with a capital gain/loss realized.")
+        N_("Company redeems units, thereby increasing the stock price by a multiple, while keeping the monetary value of overall investment constant. Any remainder units are sold/redeemed for cash, with a capital gain/loss realized.")
     },
 };
 
@@ -370,7 +370,7 @@ static const TxnTypeVec short_types
         // Translators: this is a stock transaction describing return
         // of capital retrieved from holder when shorting stock
         N_("Compensatory return of capital"),
-        N_("Company returns capital, and the short stock holder must make a compensatory payment for the returned capital. This reduces the cost basis (less negative towards $0) without affecting # units.")
+        N_("Company returns capital, and the short stock holder must make a compensatory payment for the returned capital. This reduces the cost basis (less negative, towards 0.00 value) without affecting # units.")
     },
     {
         FieldMask::DISABLED,               // stock_amt
@@ -383,7 +383,7 @@ static const TxnTypeVec short_types
         // Translators: this is a stock transaction describing a
         // notional distribution when shorting stock
         N_("Compensatory notional distribution"),
-        N_("Company issues a notional distribution, and the short stock holder must make a compensatory payment for the notional distribution. This is recorded as a loss/negative dividend income amount, and increases the cost basis (more negative away from $0) without affecting # units.")
+        N_("Company issues a notional distribution, and the short stock holder must make a compensatory payment for the notional distribution. This is recorded as a loss/negative dividend income amount, and increases the cost basis (more negative, away from 0.00 value) without affecting # units.")
     },
     {
         FieldMask::ENABLED_CREDIT,         // stock_amt
@@ -396,7 +396,7 @@ static const TxnTypeVec short_types
         // Translators: this is a stock transaction describing a stock
         // split when shorting stock
         N_("Stock split"),
-        N_("Company issues additional units, thereby reducing the stock price by a divisor, while keeping the $ value of shares constant.")
+        N_("Company issues additional units, thereby reducing the stock price by a divisor, while keeping the monetary value of the overall investment constant.")
     },
     {
         FieldMask::ENABLED_DEBIT,          // stock_amt
@@ -409,7 +409,7 @@ static const TxnTypeVec short_types
         // Translators: this is a stock transaction describing a
         // reverse split when shorting stock.
         N_("Reverse split"),
-        N_("Company redeems units, thereby increasing the stock price by a multiple, while keeping the $ value of shares constant.")
+        N_("Company redeems units, thereby increasing the stock price by a multiple, while keeping the monetary value of the overall investment constant.")
     },
     {
         FieldMask::ENABLED_DEBIT,          // stock_amt
@@ -423,7 +423,7 @@ static const TxnTypeVec short_types
         // reverse split when shorting stock. Fractional remaining
         // stock is retrieved as cash.
         N_("Reverse split with cash in lieu for fractionals"),
-        N_("Company redeems units, thereby increasing the stock price by a multiple, while keeping the $ value of shares constant. Any remainder units are closed and covered with a cash payment, with a capital gain/loss realized.")
+        N_("Company redeems units, thereby increasing the stock price by a multiple, while keeping the monetary value of the overall investment constant. Any remainder units are closed and covered with a cash payment, with a capital gain/loss realized.")
     },
 };
 
--- a/gnucash/gnome/assistant-stock-transaction.cpp
+++ b/gnucash/gnome/assistant-stock-transaction.cpp
@@ -262,7 +262,7 @@ static const TxnTypeVec long_types
         // Translators: this is a stock transaction describing a stock
         // split
         N_("Stock split"),
-        N_("Company issues additional units, thereby reducing the stock price by a divisor, while keeping the total monetary value the overall investment constant.")
+        N_("Company issues additional units, thereby reducing the stock price by a divisor, while keeping the total monetary value of the overall investment constant.")
     },
     {
         FieldMask::ENABLED_CREDIT,         // stock_amt
@@ -274,7 +274,7 @@ static const TxnTypeVec long_types
         FieldMask::DISABLED,               // capg_amt
         // Translators: this is a stock transaction describing a reverse split
         N_("Reverse split"),
-        N_("Company redeems units, thereby increasing the stock price by a multiple, while keeping the monetary value of the overall investment constant.")
+        N_("Company redeems units, thereby increasing the stock price by a multiple, while keeping the total monetary value of the overall investment constant.")
     },
     {
         FieldMask::ENABLED_CREDIT,         // stock_amt
@@ -287,7 +287,7 @@ static const TxnTypeVec long_types
         // Translators: this is a stock transaction describing a
         // reverse split. Some fractional stock is returned as cash.
         N_("Reverse split with cash in lieu for fractional unit remainders"),
-        N_("Company redeems units, thereby increasing the stock price by a multiple, while keeping the monetary value of overall investment constant. Any remainder units are sold/redeemed for cash, with a capital gain/loss realized.")
+        N_("Company redeems units, thereby increasing the stock price by a multiple, while keeping the total monetary value of the overall investment constant. Any remainder units are sold/redeemed for cash, with a capital gain/loss realized.")
     },
 };
 
@@ -396,7 +396,7 @@ static const TxnTypeVec short_types
         // Translators: this is a stock transaction describing a stock
         // split when shorting stock
         N_("Stock split"),
-        N_("Company issues additional units, thereby reducing the stock price by a divisor, while keeping the monetary value of the overall investment constant.")
+        N_("Company issues additional units, thereby reducing the stock price by a divisor, while keeping the total monetary value of the overall investment constant.")
     },
     {
         FieldMask::ENABLED_DEBIT,          // stock_amt
@@ -409,7 +409,7 @@ static const TxnTypeVec short_types
         // Translators: this is a stock transaction describing a
         // reverse split when shorting stock.
         N_("Reverse split"),
-        N_("Company redeems units, thereby increasing the stock price by a multiple, while keeping the monetary value of the overall investment constant.")
+        N_("Company redeems units, thereby increasing the stock price by a multiple, while keeping the total monetary value of the overall investment constant.")
     },
     {
         FieldMask::ENABLED_DEBIT,          // stock_amt
@@ -423,7 +423,7 @@ static const TxnTypeVec short_types
         // reverse split when shorting stock. Fractional remaining
         // stock is retrieved as cash.
         N_("Reverse split with cash in lieu for fractionals"),
-        N_("Company redeems units, thereby increasing the stock price by a multiple, while keeping the monetary value of the overall investment constant. Any remainder units are closed and covered with a cash payment, with a capital gain/loss realized.")
+        N_("Company redeems units, thereby increasing the stock price by a multiple, while keeping the total monetary value of the overall investment constant. Any remainder units are closed and covered with a cash payment, with a capital gain/loss realized.")
     },
 };
 
--- a/gnucash/gnome/assistant-stock-transaction.cpp
+++ b/gnucash/gnome/assistant-stock-transaction.cpp
@@ -138,9 +138,9 @@ static const TxnTypeVec starting_types
         FieldMask::DISABLED,               // dividend_amt
         FieldMask::DISABLED,               // capg_amt
         // Translators: this is a stock transaction describing an
-        // Initial stock purchase
+        // Initial stock long purchase
         N_("Open buy"),
-        N_("Initial stock purchase")
+        N_("Initial stock long purchase.")
     },
     {
         FieldMask::ENABLED_CREDIT,         // stock_amt
@@ -151,9 +151,9 @@ static const TxnTypeVec starting_types
         FieldMask::DISABLED,               // dividend_amt
         FieldMask::DISABLED,               // capg_amt
         // Translators: this is a stock transaction describing an
-        // initial stock short-sale
+        // initial stock short sale
         N_("Open short"),
-        N_("Initial stock short-sale")
+        N_("Initial stock short sale.")
     }
 };
 
@@ -170,7 +170,7 @@ static const TxnTypeVec long_types
         // Translators: this is a stock transaction describing
         // purchase of stock.
         N_("Buy"),
-        N_("Buying stock.")
+        N_("Buying stock long.")
     },
     {
         FieldMask::ENABLED_CREDIT,         // stock_amt
@@ -183,7 +183,7 @@ static const TxnTypeVec long_types
         // Translators: this is a stock transaction describing sale of
         // stock, and recording capital gain/loss
         N_("Sell"),
-        N_("Selling stock, and record capital gain/loss")
+        N_("Selling stock long, and record capital gain/loss.")
     },
     {
         FieldMask::DISABLED,               // stock_amt
@@ -196,7 +196,7 @@ static const TxnTypeVec long_types
         // Translators: this is a stock transaction describing
         // dividends issued to holder
         N_("Dividend"),
-        N_("Company issues dividends to holder")
+        N_("Company issues cash dividends to holder.")
     },
     {
         FieldMask::ENABLED_DEBIT,          // stock_amt
@@ -207,10 +207,10 @@ static const TxnTypeVec long_types
         FieldMask::ENABLED_CREDIT,         // dividend_amt
         FieldMask::DISABLED,               // capg_amt
         // Translators: this is a stock transaction describing
-        // dividend issued to holder, and is reinvested. Some
+        // dividend issued to holder, and it may be reinvested. Some
         // dividends are distributed as cash.
-        N_("Dividend reinvestment (with remainder)"),
-        N_("Company issues dividend which is reinvested. Some dividends are paid to holder")
+        N_("Dividend reinvestment"),
+        N_("Company issues dividend, which may be reinvested. Non-reinvested dividends are paid out as cash.")
     },
     {
         FieldMask::ENABLED_DEBIT,          // stock_amt
@@ -222,7 +222,7 @@ static const TxnTypeVec long_types
         FieldMask::DISABLED,               // capg_amt
         // Translators: this is a stock transaction describing
         // dividend which is wholly reinvested.
-        N_("Dividend reinvestment (without remainder)"),
+        N_("Dividend reinvestment (without remainder) - N/A"),
         N_("Company issues dividend which is wholly reinvested.")
     },
     {
@@ -235,8 +235,8 @@ static const TxnTypeVec long_types
         FieldMask::DISABLED,               // capg_amt
         // Translators: this is a stock transaction describing return
         // of capital
-        N_("Return of Capital"),
-        N_("Stock returns capital to holder")
+        N_("Return of capital"),
+        N_("Company returns capital, reducing the cost basis without affecting # units.")
     },
     {
         FieldMask::DISABLED,               // stock_amt
@@ -249,7 +249,7 @@ static const TxnTypeVec long_types
         // Translators: this is a stock transaction describing a
         // notional distribution
         N_("Notional distribution"),
-        N_("Stock returns a notional distribution")
+        N_("Company issues a notional distribution, which is recorded as dividend income and increases the cost basis without affecting # units.")
     },
     {
         FieldMask::ENABLED_DEBIT,          // stock_amt
@@ -262,7 +262,7 @@ static const TxnTypeVec long_types
         // Translators: this is a stock transaction describing a stock
         // split
         N_("Stock split"),
-        N_("Stock price is fractionally reduced")
+        N_("Company issues additional units, thereby reducing the stock price by a divisor, while keeping the $ value of shares constant.")
     },
     {
         FieldMask::ENABLED_CREDIT,         // stock_amt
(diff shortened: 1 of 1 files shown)


gnucash/gnome/assistant-stock-transaction.cpp
```diff
@@ -262,7 +262,7 @@ static const TxnTypeVec long_types
         // Translators: this is a stock transaction describing a stock
         // split
         N_("Stock split"),
-        N_("Stock price is fractionally reduced")
+        N_("Company issues additional units, thereby reducing the stock price by a divisor, while keeping the $ value of shares constant.")
```
fellen, line 265: Can we replace `$` here and later?
Reason: Languages are often shared by several countries with different currency symbols like en in GB (Lira Symbol), de in several EU  countries (€) and CH (CHF), …

fellen (changes requested):
Nice improvement.

@christopherlam As long as the wrong MsgContexts are not  removed, we have set assistant-stock-transaction.* on xgettext's ignore list. So theis files are not affected by the string freeze.

gnucash/gnome/assistant-stock-transaction.cpp
jralls: Good catch. Let's use `currency` instead.

christopherlam: "while keeping the currency value of shares constant"? As a layperson I'd understand "while keeping the stock valuation constant"

jralls: Well, if we're going to word-smith it:
Valuation can be construed as price, so you don't want that, and the distinction between `of shares` and `of a share` is maybe a bit fine. Plus you started with `units` so you should use that instead of `share` or `shares` throughout the sentence. `Issues units` isn't always right either, sometimes they take units away.

How about `Issuer increases or decreases the units held, proportionally adjusting the price so that the total value doesn't change.`?

CDB-Man: > Company issues additional units, thereby reducing the stock price by a divisor, while keeping the $ value of shares constant.

Instead, we could use the word `monetary` in place of the $ sign.

---
> Plus you started with units so you should use that instead of share or shares throughout the sentence.

In terms of units vs shares, the different text is purposeful.  The number of units values, but the value of the overall total shares does not change.  They should not be the same word.  Instead, we could refer to it as the `investment`.

---
> Issues units isn't always right either, sometimes they take units away.

Reverse split is covered in the next transaction, to this is exclusively a forward split and therefore exclusively issuing units.

---
New version:
> Company issues additional units, thereby reducing the stock price by a divisor, while keeping the **total monetary** value the **overall investment** constant.

fellen (approved):

gnucash/gnome/assistant-stock-transaction.cpp
```diff
@@ -207,10 +207,10 @@ static const TxnTypeVec long_types
         FieldMask::ENABLED_CREDIT,         // dividend_amt
         FieldMask::DISABLED,               // capg_amt
         // Translators: this is a stock transaction describing
-        // dividend issued to holder, and is reinvested. Some
+        // dividend issued to holder, and it may be reinvested. Some
         // dividends are distributed as cash.
-        N_("Dividend reinvestment (with remainder)"),
-        N_("Company issues dividend which is reinvested. Some dividends are paid to holder")
+        N_("Dividend reinvestment"),
+        N_("Company issues dividend, which may be reinvested. Non-reinvested dividends are paid out as cash.")
```
christopherlam, line 213: @CDB-Man If this option will take care of both "dividend reinvestment" and "dividend reinvestment with remainer", then the `cash_amount` in the truth table will then need to allow ENABLE_DEBIT and ZERO. So, a few lines above the strings will need to be changed as follows:

```patch
@@ -201,7 +201,7 @@ static const TxnTypeVec long_types
     {
         FieldMask::ENABLED_DEBIT,          // stock_amt
         FieldMask::ENABLED_DEBIT,          // stock_val
-        FieldMask::ENABLED_DEBIT,          // cash_amt
+        FieldMask::ENABLED_DEBIT | FieldMask::ALLOW_ZERO,          // cash_amt
         FieldMask::ENABLED_DEBIT | FieldMask::ALLOW_ZERO,          // fees_amt
         true,                   // fees_capitalize
         FieldMask::ENABLED_CREDIT,         // dividend_amt
```

The other UX issue is that the updated transaction description `"Company issues dividend, which may be reinvested. Non-reinvested dividends are paid out as cash."` is very much clearer, however either

1. the "accounting"-type definition could/should either be extended to include the stock-assistant UI hints, e.g. `"Company issues dividend, which may be reinvested to purchase more units. Non-reinvested dividends are paid out as cash. The dividend amount is input in the Dividend tab. If there are no non-reinvested dividends, the cash amount in the assistant will be zero."`. As you can see it gets unwieldly very quickly. The advantage is there's only 1 long string per transaction type.

2. the "accounting"-type definition stays as is, and the assistant pages' descriptions may by individuali[s|z]ed with custom strings. In this dividend reinvestment, the cash description could be "Please input the non-reinvested cash remainder. If there are no non-reinvested dividends, please leave blank or input zero". This is cleaner, but generate a NxM number of strings which can annoy translators.

3. We can ignore these UX concerns altogether (for now) :)

christopherlam:
image

The next minor issue is that the last option "Reverse stock split with cash in lieu for fractional remainders" does now look oddly specific compared with others. It *would* be simpler to simply remove this option altogether -- we could augment the "Reverse split" description to nudge the user record a sale before the reverse stock split. This means creating two transactions instead of one; the reverse stock split description would be as follows:

`"Company redeems units, thereby increasing the stock price by a multiple, while keeping the total monetary value of the overall investment constant.  If there are any fractional units sold/redeemed for cash, the sale should be recorded first, followed by the reverse stock split"`

Thus #1414 can be merged in without the exact ratio issue. The disadvantage is that the assistant won't offer the combined reverse-stock-split+sale option which is a slight downgrade from the existing Stock Split Assistant.

gnucash/gnome/assistant-stock-transaction.cpp
CDB-Man: 1. The cash flag is changed as suggested.
2. See 6ceee1b for a compromise on the definition text.  I really don't want to create for ourselves a situation with NxM future strings to maintain.  Each of the individual steps are basically identical, so any variation should be front loaded.

CDB-Man:
> The next minor issue is that the last option "Reverse stock split with cash in lieu for fractional remainders" does now look oddly specific compared with others. It _would_ be simpler to simply remove this option altogether -- we could augment the "Reverse split" description to nudge the user record a sale before the reverse stock split. This means creating two transactions instead of one; the reverse stock split description would be as follows:
As I've stated multiple times on IRC and in https://bugs.gnucash.org/show_bug.cgi?id=798004#c49 -- I have always preferred the 2 transaction solution.  It is much cleaner, and shows the A-->B relationship, where remainders are sold/redeemed before a reverse split.

I would edit your string as follows... it very quickly becomes a giant wall of text.
`"Company redeems units, thereby increasing the stock price by a multiple, while keeping the total monetary value of the overall investment constant. If there are any remainder units sold/redeemed for cash during the reverse split, the sale/redemption remainder units of  should first be recorded in a separate transaction in stock assistant, followed by a second transaction in stock assistant for the reverse stock split."`

If the "rest of GNUcash" currently does things as a single transaction, and there is a desire to maintain that behaviour, then we'll just have to live with this oddity.  It also avoids the wall of text as a minor victory.

---
P.S. as a nomenclature issue issue, it is "more" correct to call them remainder units than it is fractional units.  This is because the redemption of remainders can technically be executed either before or after the reverse split.  If done pre-split, the units are not yet fractionals.  If done post-split, the units are fractionals.  By calling them remainders, we avoid the minor nomenclature issue altogether.

I have issued 9cd6645 to address this.

christopherlam:
> I would edit your string as follows... it very quickly becomes a giant wall of text. `"Company redeems units, thereby increasing the stock price by a multiple, while keeping the total monetary value of the overall investment constant. If there are any remainder units sold/redeemed for cash during the reverse split, the sale/redemption remainder units of should first be recorded in a separate transaction in stock assistant, followed by a second transaction in stock assistant for the reverse stock split."`

Ok I'll amend to add newlines and I think it'll be ok.

> If the "rest of GNUcash" currently does things as a single transaction, and there is a desire to maintain that behaviour, then we'll just have to live with this oddity. It also avoids the wall of text as a minor victory.

I think only the legacy portfolio and advanced portfolio reports are the only "rest of GnuCash" to worry about. The IFRS report handles them well. @jralls may offer final vote?

jralls:
The lot scrubber is the other piece of "the rest of GnuCash" and IIRC it creates separate transactions. I think the portfolio reports don't care about that part.

As for the long strings, those don't belong in the UI, they belong in the documentation. If you haven't started writing up this assistant for the T&CG and Manual then it's high time you start. Keep tooltips to no more than a couple of short sentences.

I don't see a need for separate transaction types for splits, reverse splits, and reverse splits with cash in lieu. They're all just splits and the existing split dialog has for 15 years done a reasonable job of handling them without confusing users much.

A Dividend reinvestment is just a buy. All my broker and mutual fund company statements for 40 years have reported them as two transactions, one DR cash (with CR income implied) and a second CR cash DR stock.  One can combine them as illustrated in the T&CG, but having the assistant do that might surprise some users. Either way I think from a UI perspective it's better to just have a dividend page with a reinvest checkbox and an entry for the number of shares that's sensitive only when the box is ticked.

jralls:
> Company redeems units

That's wrong. Redeem means to "gain or regain possession of (something) in exchange for payment." (Oxford Languages via Google) so that would describe what's known as a share or stock buyback, where instead of returning cash to the owners with a cash payment the issuer uses the money to buy shares in the market reducing the number of shares outstanding and increasing the portion of the company owned by the remaining shareholders. This is common in the US where it's treated favorably in the tax code, but it doesn't create an accounting event for shareholders.

A split is technically speaking the company withdrawing the existing shares and replacing them with new shares that represent a different ownership portion of the company. In a normal split a new share represents less of the company than an old share; in a reverse split the new share represents more of the company. Then there's a [stock dividend](https://www.investopedia.com/terms/s/stockdividend.asp) where the company releases stock existing shares that it holds (known as treasury stock) or has already authorized but not issued to shareholders. It's technically not a split because the total number of authorized shares (and in the case of treasury stock the total number of issued shares) doesn't change, but it has the same effect because there are now more shares available for trading that represent the ownership of the company.

CDB-Man:
How does the current split tool handle remainders?  Does it properly record a sale, with the capital gain/loss entered into both the stock account and the corresponding income account?  The default split window I believe does not provide an opportunity to input cash received, price, cost basis, or capital gain, so how would it realize it?  Splits and reverse splits could perhaps be merged from a UI perspective, but reverse with remainder is definitely a separate workflow.

Unless of course, we eliminate this option and mandate that sales of remainders are entered as a regular sale... But you've all been telling me that you "prefer it a single transaction to show the link".  If we want it a single transaction, it necessarily needs to be it's own workflow.  Whether that workflow is a top level dropdown, or a step 2 checkbox, is up to you.

Same comments for reinvested dividends vs pure cash.

CDB-Man:
> > Company redeems units
> 
> That's wrong. Redeem means to "gain or regain possession of (something) in exchange for payment." (Oxford Languages via Google) so that would describe what's known as a share or stock buyback, where instead of returning cash to the owners with a cash payment the issuer uses the money to buy shares in the market reducing the number of shares outstanding and increasing the portion of the company owned by the remaining shareholders. This is common in the US where it's treated favorably in the tax code, but it doesn't create an accounting event for shareholders.
> 
> A split is technically speaking the company withdrawing the existing shares and replacing them with new shares that represent a different ownership portion of the company. In a normal split a new share represents less of the company than an old share; in a reverse split the new share represents more of the company. Then there's a [stock dividend](https://www.investopedia.com/terms/s/stockdividend.asp) where the company releases stock existing shares that it holds (known as treasury stock) or has already authorized but not issued to shareholders. It's technically not a split because the total number of authorized shares (and in the case of treasury stock the total number of issued shares) doesn't change, but it has the same effect because there are now more shares available for trading that represent the ownership of the company.

Disagree.  A reverse split is a redemption and cancellation of shares. When there is no remainder, the consideration given is $0.  When there is a remainder, the consideration given is cash.  Destruction of remainder units is necessarily reducing your percentage investment in the company (albeit by the tiniest of amounts).  I don't know about the US tax treatment, but from an accounting perspective this is a sale for the investor, and a redemption for the company.

Splits are commonly executed via stock dividends.  It is much simpler from a depository and transfer agent legal documentation perspective to issue a stock dividend... Than it would be to redeem and reissue everyone's shares.

Both of these comments are from my direct experience auditing corporations as a CPA.

jralls:
> How does the current split tool handle remainders?

It doesn't, because it takes as input the number of shares added (put in a negative number for reverse splits) and creates a transaction with a zero-value split for that number of shares. Cash in lieu creates a pair of splits DR the selected asset account and CR the selected income account. It's up to the user to create the gain/loss split either by hand or with the lot scrubber.

> but reverse with remainder is definitely a separate workflow.

and 

>  If we want it a single transaction, it necessarily needs to be it's own workflow. 

In my view a split is a split, regardless of direction. Sometimes it has "remainder" shares that get sold for cash and sometimes it doesn't. When it does there's some extra work, but the event is still a split. It's one workflow with a branch: If there's no cash in lieu then done else account for the cash in lieu and the resulting gain/loss. Does your understanding of workflow preclude conditional branching?

Whether to use one transaction or several is a matter of taste. One way is that a transaction represents a single entry in one ledger and all of the corresponding entries in other ledgers. Another way is that a transaction represents an event and might involve multiple entries in some ledgers. Of course one can model e.g. a reinvested dividend as a single event or as  two. 

That said, the various reports--and this isn't necessarily restricted to investing as the business reports are quite particular--can't handle all the possible ways a user might want to model events, nor can an assistant reasonably have more than one way of doing things. That drives a design constraint: An assistant and the reports associated with what the assistant does have to work from the same design model. If the assistant creates a single many-split transaction then the reports have to be able to interpret that transaction; conversely if the reports expect a single transaction then the assistant had better make one.  We have the portfolio reports and a lot of users who have been accounting for their investments by hand to work with those reports. Those users will be very unhappy if we change the reports in a way that breaks their many years of work, so either the assistant has to generate transactions the way the reports want or @christopherlam needs to write a replacement portfolio report (and maybe that's the IFRS one) that works with the way the assistant does.

CDB-Man:
> Does your understanding of workflow preclude conditional branching?

Conditional branching would work, yes. I don't have any opinions against, if you think it's better from a UI perspective.

---
For the current existing reports, do they expect reverse split remainder as 1 or 2 transactions?

Agree with the principle that assistant and reports should be consistent with each other.

While I personally prefer 2 transactions, if there is backwards compatibility to be gained by going 1 transaction, that is a reasonable compromise.  If the existing is already a 2 transaction model, even better.

jralls:
> Disagree. A reverse split is a redemption and cancellation of shares.

No, just like a forward split it's an exchange of old shares for new ones. If there are old shares that would turn in to fractional new shares then *those* are redeemed for cash.

[This tutorial](https://www.principlesofaccounting.com/chapter-14/splits-and-dividends/) reflects my understanding of the difference between a split and a stock dividend: The former changes the number of issued shares without affecting the books while the latter includes a transfer from retained earnings to paid-in equity.

On the other hand [a recent Tesla filing](https://www.cnbc.com/2022/03/28/tesla-wants-to-split-its-stock-so-it-can-pay-a-stock-dividend-shares-gain.html) illustrates a bit of a terminology gap: In the filing the increase in the number of authorized shares is the split and the issuing of those shares to shareholders is a dividend. And [this FAQ from Eastman](https://s29.q4cdn.com/397220757/files/doc_downloads/faq/Stock-Split-FAQs.pdf), a company old enough to actually have stock certificates outstanding, illustrates another facet: Since they also issued their split shares to shareholders as a dividend they didn't have to recall and reissue the certificates. I suppose that's the benefit to the transfer agent that you referred to. 

There's no significant difference to the investor: Dividend or split the number of their shares increases or decreases; they get cash in exchange for a fractional amount but otherwise the book cash value of the investment doesn't change.

CDB-Man:
> > Disagree. A reverse split is a redemption and cancellation of shares.
> 
> No, just like a forward split it's an exchange of old shares for new ones. If there are old shares that would turn in to fractional new shares then _those_ are redeemed for cash.

This would likely be true in the era of physical certificates.  With digital certificates, it is easier for the transfer agent to redeem half the shares than to reissue all the shares.



> [This tutorial](https://www.principlesofaccounting.com/chapter-14/splits-and-dividends/) reflects my understanding of the difference between a split and a stock dividend: The former changes the number of issued shares without affecting the books while the latter includes a transfer from retained earnings to paid-in equity.

The latter would only impact retained earnings and paid-in equity if the company's shares have a par value.  $0.00 par value shares would have no required journal entry on the book.  It would be a "shares-only" stock dividend.

> On the other hand [a recent Tesla filing](https://www.cnbc.com/2022/03/28/tesla-wants-to-split-its-stock-so-it-can-pay-a-stock-dividend-shares-gain.html) illustrates a bit of a terminology gap: In the filing the increase in the number of authorized shares is the split and the issuing of those shares to shareholders is a dividend.

The increase in authorized shares would not be a split, as no actual shares were issued. It is only an increase in the allowable issuance cap.
1. The split was effected by issuing new shares and distributing them via stock dividend.
2. The authorized share cap is merely to give the company the ability to effect the former transaction 

> And [this FAQ from Eastman](https://s29.q4cdn.com/397220757/files/doc_downloads/faq/Stock-Split-FAQs.pdf), a company old enough to actually have stock certificates outstanding, illustrates another facet: Since they also issued their split shares to shareholders as a dividend they didn't have to recall and reissue the certificates. I suppose that's the benefit to the transfer agent that you referred to.

Indeed, direct registration of shares allows for fractional units if the company, their corporate secretary, and their transfer agent are willing to deal with it. In today's world with the depository (DTCC) and with brokered accounts (Schwab, etc), depositories do not want anything to do with fractional units.

> There's no significant difference to the investor: Dividend or split the number of their shares increases or decreases; they get cash in exchange for a fractional amount but otherwise the book cash value of the investment doesn't change.

Agreed.  Ultimately, from a functional perspective, I'm still of the opinion that separate transactions are the cleanest... Notwithstanding any backwards compatibility issues.

christopherlam:
WRT wording issues, I have no particular further comments. It would be nice to finalize this PR so that we can move on to technical issues whether to record capgains separately or not, and whether to merge split/rev-split and dividend/dividend+reinvesment.

In this assistant it's far easier to have separate split and rev-split types because the latter will remind the user to record the remainder unit sale.

Ditto dividend vs dividend+reinvestment vs dividend+reinvestment+remainder -- ultimately we could condense all into 1 Dividend txn_type.

I'm not too keen on a conditional workflow because it'll complicate the code endlessly -- if conditions are complex enough then we might as well have a separate dialog for each txn_type. I'll bail.

christopherlam:
Re: portfolio reports -- I did try to simplify them previously but didn't consider all edge cases. Maybe I can try again with a copy in Experimental submenu. Perhaps the more robust IFRS report can take its place instead because it won't crash when handling shorts. For better or for worse, users like the portfolio report because it gives an indication of stock "performance" in the "Gain" columns.

CDB-Man:
As a general principle, IFRS report either considers a transaction as disposition (sale) or non-disposition (everything else). Only long sale, cover short, and long + short reverse split with remainder generate a disposition and therefore capital gain/loss.  This is a major reason why it should be separate transactions in my opinion.

christopherlam:
> In this assistant it's far easier to have separate split and rev-split types because the latter will remind the user to record the remainder unit sale.
> 
> Ditto dividend vs dividend+reinvestment vs dividend+reinvestment+remainder -- ultimately we could condense all into 1 Dividend txn_type, and nudge the user to record the reinvestment separately.

I've contradicted myself. Let's try again:
- SS & RSS could be merged. But the description must warn user to record any remainder sale *prior* to entering the RSS.
- DIV & DIV+reinv & DIV+reinv+remainder could be merged into DIV. Nudge user to record any reinvestment *after* the dividend.
- ND and ROC have no variants and are left alone... I suppose if a user receives ND & ROC and wish to buy with the proceeds, it'll be easy enough to do afterwards.

jralls:
For portfolio reports I think it would be simpler and safer to add a gain column to the IFRS report than to try to modify the existing reports and risk breaking them for users who've tailored their books to the way they are now.

christopherlam:
@CDB-Man @jralls Any further thoughts about the strings?

Do we want to merge DIV & DIV+Reinv & DIV+Reinv+Remainder, and SS & RSS & RSS+cash-in-lieu in the Assistant? The IFRS will still interpret these combined transactions correctly.

CDB-Man:
From my perspective:
1. No further edits to strings at this time
2. Yes to merging the 3 dividend transactions (+ also delete the 2 reinvestment transactions on the short side).
3. SS and RSS could be merged.
4. RSS with remainder, my personal opinion is that it's sufficiently different to keep its own step. If not, we should make it clear that the remainder should first be sold (covered) with a long sale (short cover) before executing the combined SS/RSS step.

christopherlam:
> From my perspective:
> 
> 1. No further edits to strings at this time
> 2. Yes to merging the 3 dividend transactions (+ also delete the 2 reinvestment transactions on the short side).

Ok I'll add a short note to the "Dividend" description - "Any dividend reinvested must be recorded as a regular Buy".

> 3. SS and RSS could be merged.
> 4. RSS with remainder, my personal opinion is that it's sufficiently different to keep its own step. If not, we should make it clear that the remainder should first be sold (covered) with a long sale (short cover) before executing the combined SS/RSS step.

TBH the invalid price error will be fixed in a much easier way by nudging the user to record a sale first. "If the reverse stock split results in a cash in lieu for remainder units, please record the sale (or cover buy) using the Stock Split Assistant first, then record the reverse stock split."

-----

If SS & RSS are merged, and RSS+cash-in-lieu is removed, then #1414 will still be a nice addition and can handle them all.

-----

The IFRS report will still be much more capable and can interpret them all.

-----

Any further technical issues to be discussed in bug 798004, i.e. separate or built-in capgains splits.

CDB-Man:
I don't think capital gain splits were at issue? If yes, I must have missed it. Otherwise, I definitely think current behaviour of including splits is best practice, as that conforms with IFRS.

christopherlam:
From @jralls

> The lot scrubber is the other piece of "the rest of GnuCash" and IIRC it creates separate transactions. I think the portfolio reports don't care about that part.

IIUC the lot scrubber creates separate capgains transactions. I am not too sure how it links them with the approriate transactions.

jralls:
> IIUC the lot scrubber creates separate capgains transactions. I am not too sure how it links them with the approriate transactions.

It sets kvps on the splits named `gains-split` and `gains-source`.

christopherlam:
@CDB-Man Congrats on your first merged commits!

fellen (commented):
I made a test build of a pot file including this assistant and saw a few additional issues, which weblate also would report later similar to https://hosted.weblate.org/translate/gnucash/gnucash/en/?q=has:check.

Some error messages are missing a full stop.
They would also look nicer with a beginning article.

@christopherlam There a still several strange MsgContexts
(in the glade file?) like: 
```
#. Translators: this is a stock transaction describing
#. dividends issued to holder
msgid "Dividend"
msgstr ""

msgctxt "Stock Assistant: Action field"
msgid "Dividend"
msgstr ""

msgid "Dividend Account"
msgstr ""

msgid "Dividends"
msgstr ""
```
and
```
msgid "Fees"
msgstr ""

msgctxt "Stock Assistant: Action field"
msgid "Fees"
msgstr ""
```
In text use ` or ` instead of slash like in `received/spent`

`msgid "No errors found. Click Apply to create transaction."`
Sentence 1: verb missing;
sentence 2: Quote button labels.
```
#. Translators: this is a stock transaction describing a
#. notional distribution
msgid "Notional distribution"
```
As first meaning of `Notional` I get `fictitious`. Can we get a better comment?

```
#. Translators: this is a stock transaction describing return
#. of capital
msgid "Return of capital"
```
This comment has no entropy.

```
#. Translators: this is a stock transaction describing a reverse split
#. Translators: this is a stock transaction describing a
#. reverse split when shorting stock.
msgid "Reverse split"
```
The first comment would be sufficient.

Another example of a superflous and misleading comment on a general term:
```
#. Translators: this is a stock transaction describing new
#. sale of stock, and recording capital gain/loss
#: gnucash/gnome/assistant-stock-transaction.cpp:185
#: gnucash/gnome-utils/gnc-tree-model-split-reg.c:2872
#: gnucash/gnome-utils/gnc-tree-model-split-reg.c:2876
#: gnucash/gnome-utils/gnc-tree-model-split-reg.c:2887
#: gnucash/gnome-utils/gnc-tree-model-split-reg.c:2891
#: gnucash/gnome-utils/gnc-tree-model-split-reg.c:2908
#: gnucash/gnome-utils/gnc-tree-model-split-reg.c:2919
#: gnucash/gnome-utils/gnc-tree-model-split-reg.c:2924
#: gnucash/gnome-utils/gnc-tree-model-split-reg.c:2931
#: gnucash/gnome-utils/gnc-tree-model-split-reg.c:2951
#: gnucash/gnome-utils/gnc-tree-view-split-reg.c:2990
#: gnucash/register/ledger-core/split-register.c:2567
#: gnucash/register/ledger-core/split-register.c:2571
#: gnucash/register/ledger-core/split-register.c:2582
#: gnucash/register/ledger-core/split-register.c:2586
#: gnucash/register/ledger-core/split-register.c:2603
#: gnucash/register/ledger-core/split-register.c:2614
#: gnucash/register/ledger-core/split-register.c:2619
#: gnucash/register/ledger-core/split-register.c:2646
#: libgnucash/engine/Account.cpp:171 libgnucash/engine/Account.cpp:172
#: libgnucash/engine/Account.cpp:173
msgid "Sell"
```

@CDB-Man can you add an entry [pair] about short/long positions into [gnc-glossary.txt](https://github.com/Gnucash/gnucash/blob/maint/po/glossary/gnc-glossary.txt)
see https://wiki.gnucash.org/wiki/Translation#Terms_missing_or_inadequate_in_the_glossary_file

gnucash/gnome/assistant-stock-transaction.cpp
```diff
@@ -285,9 +285,9 @@ static const TxnTypeVec long_types
         FieldMask::DISABLED,               // dividend_amt
         FieldMask::ENABLED_CREDIT | FieldMask::ALLOW_ZERO | FieldMask::ALLOW_NEGATIVE, // capgains_amt
         // Translators: this is a stock transaction describing a
-        // reverse split. Some fractional stock is returned as cash.
-        N_("Reverse split with cash in lieu for fractionals"),
-        N_("Stock price is fractionally increased. Fractional remaining stock is returned as cash.")
+        // reverse split. Some remainder stock units are returned as cash.
+        N_("Reverse split with cash in lieu for remainder units"),
+        N_("Company redeems units, thereby increasing the stock price by a multiple, while keeping the total monetary value of the overall investment constant. Any remainder units are sold/redeemed for cash, with a capital gain/loss realized.")
```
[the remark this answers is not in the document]
christopherlam: I'm not sure whether it's possible to split the two strings. This is a part of `static const TxnTypeVec long_types` and the structs are `const char*`.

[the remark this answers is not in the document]
jralls: > An additional option could be to parametrize the 2. sentence

Not an option. It assumes that there is no language where the main sentence changes for covering a short or selling a long.  Even if it was possible to know that assumption is true the msgids "the sale" and "the cover buy" would have to have the main sentence as context and a translator comment on the main sentence explaining that it's used for those two separate cases and that translators must consider all three msgids in constructing their translations.

CDB-Man:
From @fellen 
> #. Translators: this is a stock transaction describing a
> #. notional distribution
> msgid "Notional distribution"
> ```
> 
> As first meaning of `Notional` I get `fictitious`. Can we get a better comment?

Well, that's not inaccurate, as a notional distribution does not actually distribute anything...
> A notional distribution does not distribute any cash or shares to the holder; instead, the holder's cost basis is increased, and this increase in cost basis is generally considered taxable income to the holder.  The notional distribution is effectively a mark-to-market gain.

I'm not sure how else to describe that (more simply), unless this explanation is what you're looking for,

---
> @CDB-Man can you add an entry [pair] about short/long positions into [gnc-glossary.txt](https://github.com/Gnucash/gnucash/blob/maint/po/glossary/gnc-glossary.txt) see https://wiki.gnucash.org/wiki/Translation#Terms_missing_or_inadequate_in_the_glossary_file

Did you have something like this in mind?
> "long stock position"	"The investor has purchased stock units to hold as an investment, looking to profit by selling the stock when the stock price appreciates."
> "short stock position"	"The investor has sold stocks they do not own by borrowing stocks from their broker, looking to profit by repurchasing the stock (and returning the borrowed units back to the broker) when the stock price depreciates."

CDB-Man:
@christopherlam 
> "If the reverse stock split results in a cash in lieu for remainder units, please record the sale (or cover buy) using the Stock **Split** Assistant first, then record the reverse stock split."

I think the word <split> is in error, since you're referring to the Stock Assistant?

CDB-Man:
> @christopherlam
> 
> > "If the reverse stock split results in a cash in lieu for remainder units, please record the sale (or cover buy) using the Stock **Split** Assistant first, then record the reverse stock split."
> 
> I think the word `Split` is included in error, since you're referring to the Stock Assistant?
> 
> **EDIT**: nevermind, I see that the Split Assistant may have been enhanced...? will add that to testing.

The Stock Split tool records cash in lieu as a 100% capital gain, which is  definitely incorrect.  In my opinion, Chris' Stock Assistant tool should replace the legacy Stock Split tool in its entirety.  Given this, I think the reference to Split assistant should be removed.

jralls:
> The Stock Split tool records cash in lieu as a 100% capital gain, which is definitely incorrect.

No, it records a split pair between a selected asset/cash/bank account (*not* stock or fund) and a  selected income account. You can tell it to use a capital gains income account, but it can't be a cap gain in the same way that one records a cap gain resulting from a sale. For forward splits I suppose the correct bin for cash-in-lieu payments would either be dividends (not sure offhand if they'd be "qualified" in the US nor where they'd show up on a 1099B) or a cash-in-lieu income account.

For a reverse split you'd have to create the transaction and then edit it in the register to show the notional sale of the excess "old" shares and the resulting cap gain by adding the two stock account splits, one for disposing of the shares and the other with no amount or price, and to adjust the income split to be the cap gain amount instead of the amount transferred tot he currency account. It's a little complicated; isn't that why you two were going on (somewhere else that I can't find right now) about doing the fractional sale as a separate action first?

CDB-Man:
Hmm, with that additional context, I'm not sure I see the purpose of having this feature.  But yes, this is why Chris and I were going on about having them as separate transactions.

jralls:
Sorry, which feature?

If it's the legacy Split Assistant I agree that @christopherlam's new Stock Transaction Assistant is a better replacement.

CDB-Man:
> Sorry, which feature?
> 
> If it's the legacy Split Assistant I agree that @christopherlam's new Stock Transaction Assistant is a better replacement.

The legacy split assistant as a whole, and the legacy tool's ability to add that cash in lieu transaction in particular.  Agreed of course that Chris' tool supersedes it entirely.

christopherlam:
@CDB-Man there's also some strings in https://github.com/Gnucash/gnucash/blob/maint/gnucash/gtkbuilder/assistant-stock-transaction.glade if you're keen to review.

christopherlam:
With regards to creating documentation, I think the simplest approach is that I can write a draft with screenshots on wiki.gnucash.org and @CDB-Man can proofread. Someone else must do the docbook conversion -- I 👎 it.

fellen:
> From @fellen
> 
> > #. Translators: this is a stock transaction describing a
> > #. notional distribution
> > msgid "Notional distribution"
> > ```
> > 
> > As first meaning of `Notional` I get `fictitious`. Can we get a better comment?
> > ```
> 
> Well, that's not inaccurate, as a notional distribution does not actually distribute anything...
> 
> > A notional distribution does not distribute any cash or shares to the holder; instead, the holder's cost basis is increased, and this increase in cost basis is generally considered taxable income to the holder.  The notional distribution is effectively a mark-to-market gain.

A ways better description! Can you replace the translator comment by it like
`/* Translators:  A notional distribution does not distribute any cash or shares to the holder; instead, the holder's cost basis is increased, and this increase in cost basis is generally considered taxable income to the holder.  The notional distribution is effectively a mark-to-market gain. */`

> I'm not sure how else to describe that (more simply), unless this explanation is what you're looking for,
> 
> > @CDB-Man can you add an entry [pair] about short/long positions into [gnc-glossary.txt](https://github.com/Gnucash/gnucash/blob/maint/po/glossary/gnc-glossary.txt) see https://wiki.gnucash.org/wiki/Translation#Terms_missing_or_inadequate_in_the_glossary_file
> 
> Did you have something like this in mind?
> 
> > "long stock position"	"The investor has purchased stock units to hold as an investment, looking to profit by selling the stock when the stock price appreciates."
> > "short stock position"	"The investor has sold stocks they do not own by borrowing stocks from their broker, looking to profit by repurchasing the stock (and returning the borrowed units back to the broker) when the stock price depreciates."

Right. And after enabbling MsgContext we want "stock position" as context? Then we can for now write it as "stock position: long" …

fellen:
> With regards to creating documentation, I think the simplest approach is that I can write a draft with screenshots on wiki.gnucash.org and @CDB-Man can proofread. Someone else must do the docbook conversion -- I -1 it.

No problem!

PR #1439, opened 2022-09-22
[dialog-custom-report] use GtkButtons instead of pictograms for Run/Edit/Delete actions on saved report configurations.

christopherlam:
This custom dialog amended to use regular GtkButton instead of cryptic icons on the rows. Still in draft...

image

### DIFF
--- a/gnucash/gnome/dialog-custom-report.c
+++ b/gnucash/gnome/dialog-custom-report.c
@@ -70,10 +70,6 @@ typedef struct _CustomReportDialog
     GncMainWindow *window;
     GtkTreeViewColumn *namecol;
     GtkCellRenderer *namerenderer;
-    GtkTreeViewColumn *runcol;
-    GtkTreeViewColumn *editcol;
-    GtkTreeViewColumn *delcol;
-    GtkTreeViewColumn *dummycol;
 
     /* data */
     SCM reportlist;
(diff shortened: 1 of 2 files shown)


jralls:
All UI changes to master at this point.

Eww. Buttons are so 1990s. It might be nice to have the icons hidden and revealed with a left swipe/drag, but that's Apple's paradigm not Gtk's. As for the icons being cryptic, that might be but they're what the Gnome Designers have ordained and are mostly the same as what we use in the toolbar. I think `text-editor` would be a better choice that `accessories-text-editor`, but that's a separate issue.

One might argue that a 'run' icon  is superfluous as the natural way to select and run an item is to double-click on it, which works.  On the other hand gtk3-demo has a run button in its titlebar; on the other other hand that's been there since gtk2 in the early 2000s so maybe my `eww` applies to it too.

Separately, title editing should be in place not in a separate dialog. That's the current behavior and it should stay that way.  The delete and backspace keys should trigger the delete action, which they don't at present; there should also be a way to start editing from the keyboard, maybe &lt;primary&gt;return. Plain return does run the report as expected.

christopherlam:
How about a compromise?
- Run via doubleclick or GtkButton
- Edit name via pictogram
- Delete via button

I find myself needing to hover to reveal tooltip to remember what they mean, every single time I need to act on a saved report.

jralls:
> How about a compromise?

No, consistency is important in a UI. Having different actions performed with different kinds of visual elements is ugly and confusing.

I agree that the gears icon for run (the reconcile window uses it for the Finish action; a check-mark would be better there but wouldn't work for the saved configs dialog) is a bit off since Apple, Google, and MS all use a gear for "preferences", but the Gnome design team obviously disagrees. We 

I suggested replacing `accessories-text-editor`'s steno notebook icon with `text-editor`'s pencil over a sqaure because the pencil by itself is used pretty widely for edit and it's pretty close.

The delete icon is repeated with a legend in the toolbar, never mind that an X in a circle isn't likely to mean 'run' or 'edit', so it eludes me why you'd need to resort to the tooltip for that one.

PR #1445, opened 2022-10-05
Use the WebKitGTK 4.1 API builds when available

bertogg:
Libsoup2 is being replaced by the newer libsoup3. Unfortunately both version are incompatible with each other so WebKitGTK provides two different builds of the packages: the 4.0 API build uses soup2 internally, and the 4.1 API build uses soup3.

Note that the version numbers are different so both can be installed in parallel during the transition but the actual APIs are the same.

Since gnucash only depends on libsoup through WebKitGTK switching to the new 4.1 API is trivial.


christopherlam:
70958279b2 has fixed it.

PR #1447, opened 2022-10-10
Register description cell

Bob-IT:
I think this may be of use and seem to recall it being requested.
This PR changes the register description cell from an entry cell to a combo cell which lists all the previous descriptions.
To make the search more akin to a modified completion, a function was added to combocell that sets the cell to use only the type ahead search. Below is an image of the description cell popped...
reg-combo

As is, there is a difference to a normal register and the journal. The journal with default settings, lists only the descriptions of the last 30 days where as the normal registers have all entries. This could be changed by adding the if clause at line 663 in split-register-load.c
`if (!info->first_pass)`
which will limit the descriptions to be the same as the register view. Not sure which is better ?

This changes the description cell only but could be done for notes and memo cells if warranted or maybe all could be controlled via preference settings on the 'Register Defaults' page.

I do not think the change has had an impact on loading times but I do not have a lot of transaction with different descriptions so maybe some one else could give it spin.

### DIFF
--- a/gnucash/register/ledger-core/split-register-layout.c
+++ b/gnucash/register/ledger-core/split-register-layout.c
@@ -675,7 +675,7 @@ gnc_split_register_layout_add_cells (SplitRegister* reg,
 
     gnc_register_add_cell (layout,
                            DESC_CELL,
-                           QUICKFILL_CELL_TYPE_NAME,
+                           COMBO_CELL_TYPE_NAME,
                            C_ ("sample", "Description of a transaction"),
                            CELL_ALIGN_LEFT,
                            TRUE,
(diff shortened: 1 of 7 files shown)


christopherlam:
> I think this may be of use and seem to recall it being requested.

Yes: https://github.com/Gnucash/gnucash/pull/1371#issuecomment-1193502925 👍

jralls:
Thanks!

christopherlam:
Hi @Bob-IT, small observation.

After this welcome change, clicking in the register description column of an existing split will lead to the drop-down arrow appearing. Clicking on drop-down will then open a 2-element list; the existing Description, and a blank one.

The user may click the Description cell to simply *select* an old transaction for other uses e.g. Duplicate, Expand splits, Jump, etc. It would be a good UX improvement to avoid showing the drop-down arrow unless the description was being actively edited, e.g. typing characters or backspace etc. WDYT?

christopherlam:
Addendum: or maybe it's absolutely fine to leave it as it is; after all clicking on Date & Account cells _also_ show down-arrows.

PR #1448, opened 2022-10-10
Bug769256 - Change New Account Dialog

Bob-IT:
Came across this bug regarding the account dialog layout and thought I would see what the changes would look like.

The main change is changing the account type field from a tree view to a combo and repositioning it. The parent account tree view has been set to expand to the available space. The bug also mentions changing the parent account field to a combo but I do not like that idea as I think it looks better as a tree.
Added some images below, the first one is when editing, the second is a new account and the third is the type popped...
Is it an improvement ?
Any further changes required before pushing ?

Screenshot_2022-10-10_13-27-07
account-edit1
account-edit2

### DIFF
--- a/gnucash/gnome-utils/dialog-account.c
+++ b/gnucash/gnome-utils/dialog-account.c
@@ -1390,6 +1390,16 @@ opening_equity_cb (GtkWidget *w, gpointer data)
     gtk_widget_set_sensitive (aw->transfer_account_scroll, !use_equity);
 }
 
+static void
+gnc_name_entry_realize (GtkWidget *widget, gpointer user_data)
+{
+    GtkWidget *commodity_box = user_data;
+    GtkAllocation alloc;
+
+    gtk_widget_get_allocation (widget, &alloc);
+    gtk_widget_set_size_request (commodity_box, -1, alloc.height);
+}
+
 /********************************************************************\
  * gnc_account_window_create                                        *
  *   creates a window to create a new account.                      *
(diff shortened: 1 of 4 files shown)


jralls:
Why does the commodity look like a label in the first screenshot?

Bob-IT:
> Why does the commodity look like a label in the first screenshot?

That is how it currently is, Down to once the account has transactions the commodity can not be changed, line 1452 of dialog-account.c. Probably needs some padding to increase the space taken when a label.

jralls:
Ah, I didn't notice that it was Edit instead of New. Yes, some padding to make it the same size would look better.

sunfish62:
As the bug reporter, I think you've done a great job. Thank you! 

As for the point about the spacing for the commodity, could it always be rendered as a button, but inactive for existing accounts?

christopherlam:
My only comment is that the colours give me a migraine :) -- how would the dialog be like if the parent-account selector were a `gnc_account_sel` instead? After all we only need one account, and this big box is taking up a lot of space.

jralls:
> As for the point about the spacing for the commodity, could it always be rendered as a button, but inactive for existing accounts?

I think that's what GUI Gurus recommend.

> My only comment is that the colours give me a migraine

Those are the colors assigned to the account. You don't have to set ones that give you a migraine, or indeed any at all.

> gnc_account_sel

The only catch is that the current behavior pre-selects the account that's selected in the Accounts tab when one clicks the New button, which ISTM is desirable behavior. That might make the search a bit more fiddly than it is already.

sunfish62:
"current behavior pre-selects the account that's selected in the Accounts tab when one clicks the New button, which ISTM is desirable behavior"

--VERY much so!

Bob-IT:
I have change the label space size to be the same as the name_entry, this time posted a non colourful image...

account-edit3

christopherlam:
Alternatively, disabling the widget where appropriate also works well.

image

Herewith snippet #L1447 onwards
```c

    box = GTK_WIDGET(gtk_builder_get_object (builder, "commodity_hbox"));
    aw->commodity_edit = gnc_general_select_new (GNC_GENERAL_SELECT_TYPE_SELECT,
                                                 gnc_commodity_edit_get_string,
                                                 gnc_commodity_edit_new_select,
                                                 &aw->commodity_mode);
    gtk_box_pack_start (GTK_BOX(box), aw->commodity_edit, TRUE, TRUE, 0);
    gtk_widget_show (aw->commodity_edit);

    // If the account has transactions, prevent changes and add explanatory tooltip
    if (xaccAccountGetSplitList (aw_get_account (aw)) != NULL)
    {
        gtk_widget_set_tooltip_text (aw->commodity_edit, tt);
        gtk_widget_set_sensitive (aw->commodity_edit, FALSE);
    }
```

I tried a GNCAccountSel which does work however I agree it's best to leave the `parent_tree` because the latter contains the root account.

jralls:
Thanks to both Bob and Chris. I made Chris's adjustment to disable the commodity selector instead of replacing it with a label for the better visual consistency.

PR #1449, opened 2022-10-10
Remove depreciated GtkUIManager and friends from the main window

Bob-IT:
This is my attempt to remove the depreciated GtkUIManager and friends from the main window.
The command call backs have been changed to use GActions along with being part of GSimpleActionGroups.
The main concern has been changing the menus. The existing menus were built up by adding and removing individual or groups of menu entries to the menu and this looked complicated to replicate so instead I have chosen to add hole sub menus.

The menus can be thought of having two half's. The action side which connects to the command call backs and was straight forward to change. For the menu, the menu model is defined in the ui files but when the menu bar is created from the model, there is no reference on the menu item to the action which makes it difficult to find. So to get round this in the ui file I set the label to that of the action and use it to find the menu item, set my own data object so it can subsequently found and change all the labels and add tool tips for the menu items when loaded.

The gnc-main-window.ui defines all of the top level menus along with all of File, Business, Tools, Extensions, Windows and Help.
When a plugin page is loaded and comes into focus, gnc_main_wiindow_add_sub_menus is called which then loads that plugin's sub menus which is generally Edit, View and Action along with the tool bar for that plugin.
Doing it this way does lead to duplication of items in different files but as they rarely change may be acceptable. This also reduces the ui files to 9 which are being added as a resources.

Translations are mostly in the same location in the source files with only the top level menu items in the gnc-main-window.ui

I may of over squashed my commits but I can add additional ones if required. Most of the changes were just being replicated across multiple files but I may of squashed to much or unrelated changes crept in. 

If this approach is acceptable, I will continue with some refactoring and moving of some parts to gnc-gtk-utils.c along with making sure there is nothing else to change / remove. You will notice `//FIXMEb` all over the source files which is my way of keeping track of what I have added and these will be removed before the final push. I also may change #1377 to follow the same approach.

If not acceptable, can it be made so or I can just scrap it.


gnucash/gnome-utils/gnc-plugin.c
```diff
@@ -231,155 +223,266 @@ gnc_plugin_get_name (GncPlugin *plugin)
...
+                                 gint n_updates, GtkWidget *statusbar)
 {
-    GtkAction    *action;
-    gint          i;
+    g_return_if_fail (menubar != NULL);
+    g_return_if_fail (toolbar != NULL);
+    g_return_if_fail (updates != NULL || n_updates == 0);
 
-    for (i = 0; action_names[i]; i++)
+    for (gint i = 0; n_updates == -1 ? updates[i].action_name != NULL : i < n_updates; i++)
```
gjanssens, line 276: Can you explain what's happening here ? The n_updates parameter is never explained, and it seems to have a double use case.

gjanssens: OIC, this is factored out from gnc_main_window.cpp and more or less used as is. It's a bit convoluted, but perhaps not worth to improve as part of this PR.

gnucash/gnome-utils/gnc-gtk-utils.c
```diff
@@ -332,3 +332,334 @@ gnc_get_dialog_widget_from_id (GtkDialog *dialog, const gchar *id)
...
+static gboolean
+tool_item_leave_event (GtkWidget *tool_item, GdkEvent *event,
+                       gpointer user_data)
+{
+    statusbar_pop (user_data);
+    return FALSE;
+}
+
+/** Setup the callback for manu bar items so the tooltip can be
+ *  displayed in the status bar.
```
gjanssens, line 611: Minor typo: manu bar -> menu bar

gnucash/gnome-utils/gnc-main-window.cpp
```diff
@@ -118,6 +119,7 @@ enum
 #define GNC_PREF_TAB_OPEN_ADJACENT    "tab-open-adjacent"
 
 #define GNC_MAIN_WINDOW_NAME "GncMainWindow"
+#define GNC_MAIN_WINDOW_MAX_NUMBER 10
```
gjanssens, line 122: From a cursory read through the rest of the code it looks like the list of windows to select from the Windows menu is limited to 10 and this define is to keep that same restriction in your added code.
Should we really keep it ? Or can we just rewrite the code to allow an arbitrarily long list of open windows to select ?

gnucash/ui/gnc-main-window.ui
```diff
@@ -0,0 +1,626 @@
...
+            <attribute name="label" translatable="no">LogReplayAction</attribute>
+            <attribute name="action">gnc-plugin-log-replay-actions.LogReplayAction</attribute>
+          </item>
+          <item>
+            <attribute name="label" translatable="no">OfxImportAction</attribute>
+            <attribute name="action">gnc-plugin-ofx-actions.OfxImportAction</attribute>
+            <attribute name="hidden-when">action-disabled</attribute>
+          </item>
+          <item>
+            <attribute name="label" translatable="no">Mt940ImportAction</attribute>
```
gjanssens, line 54: This breaks the modularity. GnuCash can be built without aqbanking support. If that's the case there shouldn't be traces from aqbanking in other parts, such as here in the UI files. The same may apply to ofx.

gnucash/gnome-utils/gnc-gtk-utils.c
```diff
@@ -332,3 +332,334 @@ gnc_get_dialog_widget_from_id (GtkDialog *dialog, const gchar *id)
...
+void
+gnc_menu_item_setup_tooltip_to_statusbar_callback (GtkWidget *menu_item,
+                                                   GtkWidget *statusbar)
+{
+    g_return_if_fail (menu_item != NULL);
+    g_return_if_fail (statusbar != NULL);
+
+    g_signal_connect (menu_item, "select",
+                      G_CALLBACK(menu_item_select_cb),
+                      statusbar);
```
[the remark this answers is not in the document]
Bob-IT: It's is emitted when the mouse hovers over a menu item.

gnucash/gnome-utils/gnc-main-window.cpp
Bob-IT: I did this as I am only using a single digit to identify the already added window menu entry action names like Window0Action to Window9Action. It looks like I will have to create and add the menu items and can use two digits so it would support 99 windows which I think should be adequate and that would do away with the define.

gnucash/gnome-utils/gnc-gtk-utils.c
Bob-IT: OK, will fix in next push.

gnucash/ui/gnc-main-window.ui
Bob-IT: I did it this way to avoid the complications of finding the location to add menu items and then add them. When built without aqbanking support they are all hidden. I will have a look at trying to do the addition in the three locations and do away with the already added menu items.

gnucash/gnome-utils/gnc-gtk-utils.c
```diff
@@ -332,3 +332,334 @@ gnc_get_dialog_widget_from_id (GtkDialog *dialog, const gchar *id)
...
+{
+    struct search_menu sm;
+
+    g_return_val_if_fail (GTK_IS_WIDGET(menu), NULL);
+
+    sm.search_action_name = NULL;
+    sm.menu_item = NULL;
+    sm.list = NULL;
+
+    gtk_container_foreach (GTK_CONTAINER(menu), search_menu_item_list, &sm);
```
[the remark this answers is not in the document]
Bob-IT: This use is part of `gnc_menu_get_items` which is returning a list of all entries so should be OK. I think your concern is with it's use in `gnc_find_menu_item` which calls `search_menu_item_lookup`. On line 415 I am testing for the search item being found and returning which seems to work as expected, am I missing some thing?

Bob-IT:
@gjanssens  Thanks for looking and suggestions.

gnucash/gnome-utils/gnc-main-window.cpp
jralls: > Or can we just rewrite the code to allow an arbitrarily long list of open windows to select ?

Let's not. That list is displayed in the Windows menu and AFAIK there's no scrollbar available in GTK menus. It looks to me like our nominal 800x600 minimum display size will handle only around 25 lines of menu and then only if the window is all the way at the top of the screen.

I also don't recall anyone ever complaining that they couldn't create enough windows.

> so it would support 99 windows which I think should be adequate and that would do away with the define.

No, it would just change the define to 100. You still need the number to test the index against.

That aside `static unsigned gnc_main_window_max_number{10};` is better than a macro unless you need the value to initialize a static array, which you don't. Since it's a C++ file `constexpr` is better than `static`, and you *can* use `constexpr` values to initialize static arrays.

gnucash/gnome-utils/gnc-gtk-utils.c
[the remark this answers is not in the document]
gjanssens: Right, I put my comment in the wrong location.

What you are missing is that the function is still *called* for each and every menu_item, even after sm is set. It's true it will return early in the function, but for long menus it remains a big number of extra function calls. Again, I don't know for sure if it really would impact performance, but it may.

gnucash/gnome-utils/gnc-plugin.c
gjanssens: I see. The bit I find convoluted is that it has an inline if/then/else clause as the for loop condition. While it works that's pretty hard to read. And it also suggests that in some cases n_updates can be -1 and that that would mean the list of updates is ended with a sentinel entry (an updates record with a NULL action_name). I think that as well is poor design - that you can have two different ways to declare the size of an updates list - either by passing a counter or by setting a NULL value for the last element in the list.
I haven' checked if gnucash has update arrays for both methods. If so, it would be best to rewrite this to use only one method (eitherr all update lists that can be passed to this function have a size counter or they end with a sentinel update value - meaning one that has a NULL entry).

I understand this is not your code originally, but as it's being moved around anyway, it may be worth updating it.

gjanssens:
You're welcome. I'm happy you took this up. It crossed my mind more than once.

gnucash/gnome-utils/gnc-main-window.cpp
Bob-IT: OK, I will leave this at 10 and change to use a `constexpr`. Interestingly even though there are only 10 entries [here](https://github.com/Gnucash/gnucash/blob/3c306eae658e3b4645eb51fc33a45205eaabf4dd/gnucash/gnome-utils/gnc-main-window.cpp#L476) you can still keep creating new windows, I stopped at 12. I suppose `gnc_main_window_cmd_window_move_page` should check the `active_windows` list length and stop at 10, should a dialog be presented advising max number of windows open?

jralls: Maybe a message box saying that the menu max has been exceeded and more windows won't be added to the Window menu.  Be sure to wrap it in `ifndef MAC_INTEGRATION`, macOS manages the Windows menu on its own.

gnucash/gnome-utils/gnc-gtk-utils.c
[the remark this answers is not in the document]
christopherlam: There's `find_widget_func` in the same file in which the comparison function `(g_strcmp0 (name, id) == 0)` can be generalised.

https://github.com/Gnucash/gnucash/blob/7c29bf7efced3f817b0f5f1a8d0babf7c74d5de7/gnucash/gnome-utils/gnc-gtk-utils.c#L302

gnucash/gnome-utils/gnc-plugin.c
Bob-IT: I have checked all occurrences and all were being passed the number of elements and none of the arrays had a sentinel so  changed this to always expect the number of elements, I did not notice this as I also used it for the GncDisplayItems.

gnucash/gnome-utils/gnc-gtk-utils.c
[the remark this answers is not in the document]
Bob-IT: Changed this to use a GList.

gnucash/ui/gnc-main-window.ui
Bob-IT: Changed this to create the Aqbanking and Ofx menu entries from the associated plugin so no need to hide the menu entries when not built.

gnucash/gnome-utils/gnc-main-window.cpp
```diff
@@ -255,12 +270,19 @@ typedef struct GncMainWindowPrivate
     GHashTable *merged_actions_table;
     /** Set when restoring plugin pages */
     gboolean restoring_pages;
+
+    const gchar   *previous_plugin_page_name; //FIXMEb added
+    const gchar   *previous_menu_quailifier; //FIXMEb added
```
gjanssens, line 275: There's a typo in the variable name. Better fix it before merging... (quailifier->qualifier).

gnucash/gnome-utils/gnc-gtk-utils.c
```diff
@@ -332,3 +332,320 @@ gnc_get_dialog_widget_from_id (GtkDialog *dialog, const gchar *id)
...
+
+
+static gpointer
+find_menu_item_func (GtkWidget *widget, const gchar *action_name)
+{
+    GtkWidget *ret = NULL;
+
+    if (GTK_IS_MENU_ITEM(widget))
+    {
+        const gchar *a_name = g_object_get_data (G_OBJECT(widget), "myaction-name");
```
gjanssens, line 413: Every GtkMenuItem also implements the GtkActionable interface. So I think it should be possible to simply do
```
const gchar *a_name = gtk_actionable_get_action_name (widget);
```

This may need a typecast for widget to GTK_ACTIONABLE.

However as to me it looks like the action name is readily available, I wonder if you really need to set the myaction-name property on each menu item separately.

gnucash/gnome-utils/gnc-main-window.cpp
```diff
@@ -269,237 +291,264 @@ GNC_DEFINE_TYPE_WITH_CODE(GncMainWindow, gnc_main_window, GTK_TYPE_WINDOW,
...
 typedef struct
 {
-    /** The merge identifier for this action group.  This number
-     *  is provided by the UI manager. */
-    guint merge_id;
     /** The action group itself.  This contains all actions added
      *  by a single menu or content plugin. */
-    GtkActionGroup *action_group;
+    GSimpleActionGroup *simple_action_group; //FIXMEb added
 } MergedActionEntry;
```
gjanssens, line 297: As the struct has only one member variable left, you might as well drop it completely and just just GSimpleActionGroup directly instead. Will improve readability of the code IMO.

gnucash/gnome-utils/gnc-main-window.cpp
```diff
@@ -3572,60 +3685,142 @@ gnc_main_window_unmerge_actions (GncMainWindow *window,
...
+    priv = GNC_MAIN_WINDOW_GET_PRIVATE(window);
+
+    if (priv->merged_actions_table == nullptr)
+        return nullptr;
+
+    iter.action = nullptr;
+    iter.action_name = name;
+    iter.group_name = group_name;
+
+    g_hash_table_foreach (priv->merged_actions_table, group_foreach_cb, &iter);
```
gjanssens, line 3757: I'm not convinced we need to keep track of merged actions in a merged_actions_table. So I'm systematically going through all uses and check if there are alternatives.
I *think* in this case what you really want to do is this:
```
auto action_group = gtk_widget_get_action_group (GTK_WIDGET(window), group_name);
if (action_group)
    action = g_action_map_lookup_action (G_ACTION_MAP(action_group), name);
```

gnucash/gnome-utils/gnc-main-window.cpp
```diff
@@ -3642,61 +3837,334 @@ gnc_main_window_get_action_group (GncMainWindow *window,
     if (entry == nullptr)
         return nullptr;
 
-    return entry->action_group;
+    return entry->simple_action_group;
```
gjanssens, line 3840: Same remark regarding merged_actions_table. I think this is what is needed here:
```
return gtk_widget_get_action_group (GTK_WIDGET(window), group_name);
```

All but the early sanity tests (g_return_val_if_fail) is no longer needed.

gnucash/gnome-utils/gnc-plugin.c
```diff
@@ -231,155 +223,266 @@ gnc_plugin_get_name (GncPlugin *plugin)
...
-        LEAVE("fail");
-        return 0;
-    }
+        GncDisplayItem *gdi;
+        GtkWidget *item = GTK_WIDGET(ptr->data);
+        const gchar *action_name = (const gchar*)g_object_get_data (G_OBJECT(item), "myaction-name");
 
-    merge_id = gtk_ui_manager_add_ui_from_file (ui_merge, pathname, &error);
-    DEBUG("merge_id is %d", merge_id);
+        PINFO("action_name is '%s'", action_name);
```
gjanssens, line 357: I'd put this after the test for a valid action_name below. I'm not sure the log formatter gracefully handles null pointers.

gnucash/gnome-utils/gnc-plugin.c
```diff
@@ -231,155 +223,266 @@ gnc_plugin_get_name (GncPlugin *plugin)
...
+        if (GTK_IS_ACTIONABLE(item))
+        {
+            const gchar *item_action_name = gtk_actionable_get_action_name (GTK_ACTIONABLE(item));
+            GncDisplayItem *gdi;
+            gchar *ptr;
+
+            PINFO("item_action_name is '%s'", item_action_name);
+
+            if (!item_action_name)
+                continue;
```
gjanssens, line 417: Same here: first test for a valid item_action_name, and then call PINFO.

gnucash/gnome-utils/gnc-plugin.c
```diff
@@ -231,155 +223,266 @@ gnc_plugin_get_name (GncPlugin *plugin)
...
+
+    ENTER("display_item_hash %p, toolbar %p", display_item_hash, toolbar);
+
+    for (gint i = 0; i < gtk_toolbar_get_n_items (GTK_TOOLBAR(toolbar)); i++)
     {
-        g_critical("Failed to load ui file.\n  Filename %s\n  Error %s",
-                   filename, error->message);
-        g_error_free(error);
-    }
+        GtkToolItem *item = gtk_toolbar_get_nth_item (GTK_TOOLBAR(toolbar), i);
```
gjanssens, line 406: As you iterate over all the child items in the toolbar it may be more efficient to use gtk_container_foreach rather than gtk_toolbar_get_nth_item.

gnucash/gnome-utils/gnc-gtk-utils.c
Bob-IT: That's what I thought but unfortunately when the menu is built from the model the GTK_ACTIONABLE is not populated. So to be able to find the menu items to hide them as required I needed another way to find them other that the label that is why I added this. Using inspector, all the menu items that are added by the model have the 'action-name' field blank unlike the ones I have built for the Aqbanking menu items that have the full action name.
I still maybe missing some thing...

gnucash/gnome-utils/gnc-main-window.cpp
Bob-IT: Yes I was wondering about that but was still unsure about whether I needed to track these merged actions at all. Will have a look.

gnucash/gnome-utils/gnc-plugin.c
Bob-IT: OK.

gnucash/gnome-utils/gnc-plugin.c
Bob-IT: OK.

gnucash/gnome-utils/gnc-plugin.c
Bob-IT: OK, will have a look.

gnucash/gnome-utils/gnc-main-window.cpp
Bob-IT: OK, will be in next push.

gnucash/gnome-utils/gnc-main-window.cpp
```diff
@@ -142,6 +146,8 @@ static QofLogModule log_module = GNC_MOD_GUI;
 static GObjectClass *parent_class = nullptr;
 /** An identifier that indicates a "main" window. */
 static GQuark window_type = 0;
+/** A hash table of the display items to update the menu and tool bars */
+static GHashTable *display_item_hash; //FIXMEb added
```
gjanssens, line 150: Similar to the merged_actions_table in gnc-plugin I doubt this extra hash is really needed. From what I gather this is all about widget related properties (labels, icon names,...). So instead of storing them in a hash, why not set these properties directly on the widgets as they are created ? If you can't set them to the exact widget property immediately, you can still add them with g_object_set_data_full as generic associated data to query back later. This will save us quite a bit of duplicate hash table queries imo.

The trick will be to figure our where exactly this is best done. I suggest to find where the widgets are created (probably when loading a ui file) and and that point set the display properties related to those widgets.

gjanssens (commented):
@Bob-IT It looks like my review will be iterative. My first scan was high level. This time I was triggered by how you changed the handling of ui elements in the plugins. I don't think that's really optimal at this point and that prompted me to grok how all of this works together a bit more.

One thing that still eludes me in general is this: we have a gnc-main-window and several plugin pages that add menu items. How does the code know where to insert the menu items from these plugin pages ? Do you know ?

I would assume the same logic should be applicable for the menu items inserted by aqbanking and ofx.

gnucash/gnome-utils/gnc-gtk-utils.c
gjanssens: That's surprising.
Perhaps this is because we are still loading old-style ui files instead of true glade files ?
If I load one of our ui files in glade only the toolbar appears, not the menubar. Note for example in gnc-main-window.ui the toolbar is defined as an `object`, and the actions are referred to via `action-name`. In the menubar definitions these are referred to via `action`. Perhaps that's the issue ?

gjanssens: If so, this means we may have to redo the ui files in glade or convert them to glade format in some way. Might as well change the extension to .glade then. Some may even be merged with glade files for the same object. For example there's both a gnc-plugin-page-budget.glade and a gnc-plugin-page-budget.ui. If we convert the latter to glade format, it could be added in the former.

gnucash/ui/gnc-main-window.ui
gjanssens: (This was supposed to be my first comment for today, but I got side-tracked trying to understand the code)

I guess this works, but I wonder why you chose to hard-code these menu entries instead of using a ui file and using gtk_builder to load it as you do with the other ui files ?

gnucash/gnome-utils/gnc-gtk-utils.c
jralls: More likely it's because we're still using GtkMenu instead of GMenu. That's not a simple change: It also requires GApplication and removing GtkOsxApplication and maybe modifying some of our other macOS customizations.

Bob-IT:
@gjanssens  The main thing I wanted to avoid is adding individual menu items based on trying to find the right menu location and insert the menu item. So I decided to try this...

As each plugin is added the actions get added to the window and the menu item parts, label, icon, action_name, tooltip get added to a GncDisplayItem and these are added to the common hash table for all windows.
The main menu bar is created when the main window is created from the gnc-main-window.ui file but only the top level file, tools, business and help are fully defined. Once the menu is created, it is scanned and it is here the labels and tool tips get added and by using the action name as the hash key I thought would be quick.

When a plugin page is loaded, the focus call back adds the appropriate edit, view and actions sub menus from there respective ui files. It is the ui file menu model that has holds the menu layout with the menu items label and action_name defined with the action_name. As each sub menu is created, it is scanned and the labels and tool tips get added.

I am also checking when changing from one page type to another to limit adding the sub menus so going from one register to another should not trigger the sub menus being recreated.
Hope this answers your question.

gnucash/gnome-utils/gnc-main-window.cpp
Bob-IT: I am not sure about this, see comment below. 
In a sense I am, when the menu parts are loaded from the builder object and the menu created, the menu is scanned and the menu items updated.

gjanssens:
@Bob-IT Thanks for the explanation. I more or less got the general idea. I'm mostly missing some details.

For example, the gnc-main-window.ui file defines an `Edit` submenu. Then in, say, gnc-plugin-page-register.ui there are the effective menu items for the edit menu. But where is it declared or decided that these menu items should be inserted in the edit menu that is declared in gnc-main-window.ui ? I know they should be there because I interpret the names and labels. But how does the code figure this out ?

Bob-IT:
Adding and hiding the sub menus is defined in GncAddSubMenu instances. If you do a search for `add_submenus` you will find each plugin page type has one.

gnucash/gnome-utils/gnc-gtk-utils.c
gjanssens: Perhaps. A simple test should give more details.
@Bob-IT can you experiment with setting property `action-name` instead (or in addition to) the `action` property ? If that fails to parse you may have to try and recreate one menubar in glade, such that it's described as an objects hierarchy (similar to how the toolbar is already described), and then check in the inspector whether the action-name property is set ?

gjanssens:
Ok, I found what I was looking for in `gnc_main_window_add_sub_menus`, which at some point calls `main_menu_add_submenus`. That's where the top-level menus are first shown or hidden and then later on the actual dropdown menus are added by iterating over the list of top level menus and searching a matching submenu for each top-level menu and specific for the current plugin page.

Is there really no more elegant way to declare where a menu should be added ? In the old GtkUiManager implementation the ui files were full of placholder elements in which menus and menu items could be added. I never bothered figuring out exactly how that worked, but can't we implement something similar based on GtkMenu ?

My general feeling is also we're doing a lot of tracking of parameters and lists and objects that are probably already tracked by Gtk or the GAction objects. I may be wrong but am willing to dig deeper (small bits at the time)... This is not all @Bob-IT 's fault though. I'm pretty sure the original code also did so. However as you're reworking this now, we may as well try to clean that legacy up.

gnucash/gnome-utils/gnc-main-window.cpp
gjanssens: Ok, then why the intermediary step via a hash ? The properties to set on the menu items are already in a static array.

Also are the same menu items created more than once per main window ? And how about setting the properties for these menu items, are these updates frequently or only once per main window ? For example are the items recreated or the properties set when changing plugin pages ? A hash table would be useful if setting these properties had to be done over and over again.

gnucash/gnome-utils/gnc-gtk-utils.c
```diff
@@ -332,3 +332,320 @@ gnc_get_dialog_widget_from_id (GtkDialog *dialog, const gchar *id)
...
+        ml->list = g_list_prepend (ml->list, widget);
+
+        if (!a_name)
+        {
+            GtkWidget *accel_label = gtk_bin_get_child (GTK_BIN(widget));
+
+            if (accel_label)
+            {
+                // use gtk_label_get_text to get text with no underlines
+                const gchar *al_name = gtk_label_get_label (GTK_LABEL(accel_label));
```
christopherlam, line 484: https://gist.github.com/christopherlam/565fc8d2fbd0511662ae6f7c14f7a30f

Bob-IT: @christopherlam When I made my original change I wanted to do as you have done but just could not get the correct syntax, I can see from your gist where I was going wrong, thank you.

gnucash/gnome-utils/gnc-gtk-utils.c
gjanssens: That's the workaround I was going to suggest if we couldn't make the glade ui files work :wink: 

More on this later.

gnucash/gnome-utils/gnc-plugin.c
Bob-IT: I am going to leave this as is for now, I think I did it this way as I would need to create a structure to pass the parameters needed in the container for each and as there were only small number of items did no think it worth doing but may come back to it once other changes are possibly made.

gnucash/gnome-utils/gnc-main-window.cpp
Bob-IT: I have removed all to do with merged_action_hash.

Bob-IT:
> Is there really no more elegant way to declare where a menu should be added ? In the old GtkUiManager implementation the ui files were full of placholder elements in which menus and menu items could be added. I never bothered figuring out exactly how that worked, but can't we implement something similar based on GtkMenu ?

This comes down to that I did not want to go finding locations in the menu to add items, then remove them when the focus changes to a different plugin. I do not think there is the concept of a place holder in the GtkMenu, The only way I could see is if dummy menu items were added and they were used to find locations to add items, they would ether have to be removed or made hidden when items added. As most of the changes were made in edit, action and view I thought it was simpler to just replace the hole sub menu. This does lead to some duplication in the ui files but thought it to be OK as a trade off on the dynamic replacement of menu items.

gnucash/gnome-utils/gnc-main-window.cpp
Bob-IT: I think the main reason for this intermediary step was that I am using `gnc_main_window_merge_actions` which is called when the plugin is added to the window and this is all part of creating the window. So when the plugin is added, the menu item properties are added to the hash table, I think there were around 150 and then subsequently used to update the menu. Doing it this way seemed like a good idea as I would not have to fetch the properties from different plugins when required.

I do not think there were any menu items that appeared in more than one location apart from being on the toolbar.

The `File`, `Business`, `Tools`, `Reports` and `Help` are mainly static with some items being shown/hidden when required.
The `Transaction` sub menu is static but the labels are changed when on transactions or splits and only shown for registers.
The `Schedule` sub menu is static but only shown for the schedule plugin.
The `Edit`, `View` and `Action` are the ones that change the most depending on plugin.

So the main part of the menu bar is created once and updated. The `Edit`, `View`, and `Action` are created and updated on plugin focus. It may be possible to keep these sub menus once created as part of the plugin but have not looked at that which may also apply to the right mouse popup menu that is also defined in the ui files, created and updated when required.

gjanssens:
> > Is there really no more elegant way to declare where a menu should be added ? In the old GtkUiManager implementation the ui files were full of placholder elements in which menus and menu items could be added. I never bothered figuring out exactly how that worked, but can't we implement something similar based on GtkMenu ?
> 
> This comes down to that I did not want to go finding locations in the menu to add items, then remove them when the focus changes to a different plugin. I do not think there is the concept of a place holder in the GtkMenu, The only way I could see is if dummy menu items were added and they were used to find locations to add items, they would ether have to be removed or made hidden when items added. As most of the changes were made in edit, action and view I thought it was simpler to just replace the hole sub menu. This does lead to some duplication in the ui files but thought it to be OK as a trade off on the dynamic replacement of menu items.

I have been studying this some more. And here are some extra findings.

1. With what I have seen so far I don't think it's a good idea to duplicate menu items as we do now. Our gui structure is rather hierarchical starting with a window, in which we add plugin-pages and each plugin page handles a specific plugin. In addition there are pageless plugins that also add functionality. This is done to allow for code reuse and easy extensibility. This hierarchy is lost in the ui files, which makes it harder to map them back to the proper subsection of the interface.
2. As to your remark about the placeholder concept missing in GtkMenu, I suppose you're right. However my guess is we could model this using permanently hidden menu items. There's an undocumentend (rather internal) GMenuModel attribute called [`hidden-when`](https://gitlab.gnome.org/GNOME/gtk/-/blob/main/gtk/gtkmenutrackeritem.c#L528) which could be set to [3 (HIDDEN_WHEN_ALWAYS)](https://gitlab.gnome.org/GNOME/gtk/-/blob/main/gtk/gtkmenutrackeritem.c#L103). Alternatively if we don't want to rely on internal implementation, we can set our own custom attribute (like for example `gnc-placeholder`) and hide the placeholders while loading the GMenu. To get the menu snippets provided by our plugins in the right locations we could add a custom attribute (like for example `gnc-insert-before`/`gnc-insert-after`) to these menu snippets. Our code could then generically parse this gnc-insert-... attributes and use those to find the location in the global menu to insert the menu snippen using the GMenuModel functions (`g_menu_model_insert`,...) 
3. Gtk's treatment of menus has clearly drastically changed and it's taking me some time to wrap my head around it. From what I gather, the primary object to interact with now is the GMenu rather than a GtkMenu. The latter can be generated from the former by Gtk when required. GMenu has features to add and remove menus at arbitrary locations. We build the complete menu model by composing GMenus that are provided by plugins (and plugin pages). If each of our plugin(-page)s maintains a list of GMenus it manages, that could be used to update the complete menu model upon opening/closing/switching tabs. This may imply we reconstruct the full GtkMenuBar on each such switch. I have no idea if that's a heavy process and whether that needs optimization. A potential optimization is that each plugin-page caches the complete menu bar that it wants to present and just switch menu bars when switching tabs.
4. My suggestion to go for GtkMenus to avoid GncDisplayItems is not a good one - GtkMenus are not the object we want to manage directly in our glade files, Gtk's new way to handle menus is via GMenu and GMenuModel objects.
5. Having said that though, we already *do* create GMenus or Toolbars for all actions that will ever be visible in our ui files. As far as I can see most of the data we currently track in GncDisplayItem can be perfectly added in the ui files instead. Both a GMenuItem and a ToolBarItem in a ui file have attributes for action name, label and icon. The short_label is just an replacement for labels in certain ToolBarItems. There's no need to track both for a single ToolItem, so it can replace label where ever it's used currently.
6. Accelerator in GncDisplayItem is unclear: [the gnome wiki page on GMenuModel](https://wiki.gnome.org/Projects/GLib/GApplication/GMenuModel) suggests the `accel` attribute for GMenu but [the GMenuModel reference page](https://docs.gtk.org/gio/class.MenuModel.html) doesn't mention it at all. Instead [this tutorial page](https://developer.gnome.org/GMenu/documentation/tutorials/actions.html#accelerators-keybindings-for-actions) suggests we can add accelerators directly to actions. Additionally [the Gtk internal implementation](https://gitlab.gnome.org/GNOME/gtk/-/blob/main/gtk/gtkmenutrackeritem.c#L726) still seems to support in GtkMenuTrackerItem.c (gtk) I found some support reading the "accel" attribute from a GMenuItem.  But perhaps that only works when using GtkApplication ? If we really can leverage them as accelerators directly in the GMenuModel, I would add them in the ui file. On the other hand if they are effectively used when associated with an action, I'd add them to actions as described in the link above.
7. And finally there's the tooltip. I don't see support for that in GMenuModel and on top of it we use it both for menuitems and toolitems. So that's probably the last remaining bit we have to set for ourselves once we generate GtkMenus and GtkToolbars.
8. Switching to a GtkApplicationWindow in GnuCash seems relatively straightforward. All that is needed is to change the base class of `GncMainWindow` from `GtkWindow` to `GtkApplicationWindow`. I have pushed my experiment to [the GtkGUI branch](https://github.com/gjanssens/gnucash/commits/GtkUI) of my github repo. If you like what you see, feel free to cherry-pick.
9. The same branch also contains a minor correction in one of the ui files to fix a lot of Gtk warnings about a missing toolbar.
10. I'm still puzzled why the action attributes as declared in our ui files are not properly converted to action-name properties in our GtkMenuItems. I'm getting more and more convinced this should be possible. Needs more study. In the interim we can choose to set them manually the moment we convert our GMenuModels into GtkMenu(Bar)s. But if you do, please add comments to indicate this is a workaround and perhaps even add DEBUG messages to note which action was set. I'm thinking of something like (pseudo-code):
```
Iterate-over-GMenuModel/GtkMenu;
if (CurrentGMenuItem-has-action-attribue)
  if (CurrentGMenuItem-action-attribute <> GtkMenu-action-name-property)
  {
    GtkMenu-action-name-property = CurrentGMenuItem-action-attribute;
    DEBUG("GtkMenuItem action-attribute differs from corresponding GMenuItem, set to %s", GtkMenu-action-name-property);
  }
  else
    DEBUG("GtkMenuItem action-attribute was preset to %s, do nothing", GtkMenu-action-name-property);
```
That could help in debugging this in the future.

gjanssens:
@Bob-IT That was a lot of text at once... If you make further changes, I would appreciate to do them in small chunks (where that's possible at least). I find this a very worthwhile conversion and like to keep track. Due to my limited time I have an easier keeping up if the blobs of change are smaller.

Bob-IT:
@gjanssens Yes that is a lot and will need time to think about it.

It looks like to me you want the plugins to update a bare minimum main window menu model, essentially all the actions from `gnc-main-window.cpp` and placeholders for the plugins to find and add there actions. Some of these will be permanent like the top of the reports menu and some temporary that are added / removed on page changes. As I stated this was not what I wanted to do but will have a think.
Regarding...
2. Already came across `hidden-when` and was using that for hiding the aqbanking menu options before I changed it. There was also a problem in finding items in GtkMenu marked as such but if all done in the model should be OK.
6. Have a look at #1377, you will see I have set all parameters in the ui file. Still needed to scan the GtKMenu after built to get the accelerator keys to work. 
8. Did not try that so maybe worth while.
10. If the action-name is not populated in the GtkMenu, I am struggling to see how it will work but maybe it just needs trying.

gjanssens:
> The only other thing you did not mention is translations, potentially you could have the same label / tool tip in multiple places.
> 
True. And that's a good motivation to limit duplication. The way we currently define GncDisplayItems was probably partly done with that in mind. The idea is that the same label can be used on both a menu and a tool bar button. However we already partly defeat that goal by defining separate short-labels for tool buttons. Also where a menu item can have an accelerator key (indicated with the '_' character, tool bar buttons don't use these. So technically these are almost always different strings. That's different for tooltips, which is why I'm cautious about those.

> Regarding...
> 2. Already came across `hidden-when` and was using that for hiding the aqbanking menu options before I changed it. There was also a problem in finding items in GtkMenu marked as such but if all done in the model should be OK. 

:+1: 

> 5. Is it feasible to just always use short labels on the toolbar and so forget about `short_label` setting.

The way I understand the current code we always set a short label if it is defined in GncDisplayItems. So in that case the normal label is not used for the ToolItem. And as far as I can see we define all toolbars in ui files. So the net result is that the toolitem label property is either set to GncDisplayItem's label or short-label property, but it never changes once set. So yes, I think this can be handled by simply setting the label correctly directly in the ui file.

> 6. Have a look at #1377, you will see I have set all parameters in the ui file. Still needed to scan the GtKMenu after built to get the accelerator keys to work, I actually started this with a fully defined ui file and then changed to what I did.

I had forgotten about that PR. I will have a closer look to learn how you did it there.

> 8. Did not try that so maybe worth while.

> 10. If the action-name is not populated in the GtkMenu, I am struggling to see how it will work but maybe it just needs trying.

The point is that normally GtkBuilder is supposed to fill in the action-name from the GMenu description when creating the menu. As that currently doesn't work, my suggestion is to do this ourselves temporarily (until we understand what's wrong). This is an alternative to your introduction of "myaction-name" data, which you suggested yourself. The best moment to do that IMO is whenever the menus are converted from GMenu to GtkMenu. 

The fact this is not happening is because we are doing something wrong, because all of the Gtk documentation I found on that subject so far suggest this should happen.

That may be because we're not using GtkApplication, though from what I have read so far, it looks like GtkApplication is primarily a convenience object that handles a lot of typical, early application tasks. Perhaps in those early tasks there is one we are currently not doing. It may however also be that we have a timing issue - maybe the actions don't exist yet when we generate the menus. As I said, needs more study (also for me).

As an aside, I'm currently estimating what's required to convert gnucash to be GtkApplication based. So part of this may eventually go away once I have figured that out.

Bob-IT:
@gjanssens 
You might like to give the GTK demo application a spin. If you look at the application.c file, it does exactly what i was doing but if you run the inspector you get this....
Under `DemoApplication` two GMenu's, one dealing with the 'app' menu and one dealing with 'Preferences and Help' which show action entries.
Under the `DemoApplicatioWindow` you have a GtkMenuBar and items and all do not have action-name populated.

Bob-IT:
Closing this as it is replaced by #1456

PR #1453, opened 2022-10-25
[gnc-features.cpp] convert to cpp

christopherlam:
A useful glib->stl conversion

- don't need to create/destroy GHashTable for each feature query
- plugs leak: `g_hash_table_unref (features_used)` was not always called properly
- to check 1 feature, don't need to traverse whole GHashTable

### DIFF
--- a/libgnucash/engine/gnc-features.cpp
+++ b/libgnucash/engine/gnc-features.cpp
@@ -19,28 +19,22 @@
  *                                                                  *
 \********************************************************************/
 
+#include <unordered_map>
+#include <string>
+#include <numeric>
+
 #include <config.h>
 
 #include <glib.h>
 #include <glib/gi18n.h>
-#include <errno.h>
-#include <fcntl.h>
-#include <string.h>
-#include <sys/stat.h>
-#include <sys/types.h>
+#include "qofbook.hpp"
 
-#include "qof.h"
-#include "gnc-features.h"
-#include "gnc-glib-utils.h"
-
-typedef struct
+extern "C"
 {
-    const gchar *key;
-    const gchar *desc;
-} gncFeature;
+#include "gnc-features.h"
+}
 
-static GHashTable *features_table = NULL;
-static gncFeature known_features[] =
+static const FeaturesTable features_table
 {
     { GNC_FEATURE_CREDIT_NOTES, "Customer and vendor credit notes (requires at least GnuCash 2.5.0)" },
     { GNC_FEATURE_NUM_FIELD_SOURCE, "User specifies source of 'num' field'; either transaction number or split action (requires at least GnuCash 2.5.0)" },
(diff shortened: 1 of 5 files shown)


jralls:
You should write some unit tests first.

QofBook's features is a KvpFrame that's implemented as a std::map and has a `get_slot(std::string path)` function. Using that directly would be a heck of a lot clearer, never mind more efficient, than this dance. Put the checking code in qofbook.cpp and call it from gnc-features.cpp to keep the implementation details hidden.

christopherlam (commented):
even faster test_feature

libgnucash/engine/gnc-features.cpp
```diff
@@ -54,42 +53,14 @@ static QofLogModule log_module = G_LOG_DOMAIN;
...
 gchar *gnc_features_test_unknown (QofBook *book)
 {
-    std::string str;
-    auto features_used = qof_book_get_features (book);
-
-    /* Iterate over the members of this frame for unknown features */
-    g_hash_table_foreach (features_used, &gnc_features_test_one, &str);
-    g_hash_table_unref (features_used);
-    return str.empty() ? nullptr : g_strdup (str.c_str());
+    return qof_book_test_features (book, (gpointer)&features_table);
```
jralls, line 63: Don't use `gpointer`, aka `void*`, in C++, it defeats the type system. Add this to qofbook.h:
```
#ifdef __cplusplus
using FeaturesTable = static_cast<std::unordered_map<std::string,std::string>*>(data);

char* qof_book_test_features(QofBook* book, const FeaturesTable& features_table);
bool qof_book_test_feature(QofBook* book, const char* feature);
#endif
```
You might consider having `qof_book_test_features` return a std::string and doing the `g_strdup` here.  `gnc_features_check_used` might need `return static_cast<gboolean>(qof_book_test_feature(book, feature);`  if there's a compiler warning.

christopherlam: Nice, I'd thought it would be compulsory to have a `gnc-book.hpp`.

christopherlam: Couldn't make it work. `gnc-book.hpp` it is.

libgnucash/engine/test/test-qofbook.c
```diff
@@ -768,6 +769,25 @@ test_book_get_collection( Fixture *fixture, gconstpointer pData )
...
+    g_assert_false (gnc_features_check_used (fixture->book, "Credit Notes"));
+
+    gnc_features_set_used (fixture->book, "Credit Notes");
+    g_assert_null (gnc_features_test_unknown (fixture->book));
+    g_assert_true (gnc_features_check_used (fixture->book, "Credit Notes"));
+
+    /* cannot set an unknown feature: it bails out. */
+    /* gnc_features_set_used (fixture->book, "Quantum Computing and Nanotech"); */
+    /* g_assert_nonnull (gnc_features_test_unknown (fixture->book)); */
+    g_assert_false (gnc_features_check_used (fixture->book, "Quantum Computing and Nanotech"));
```
christopherlam, line 788: I don't know how to feed a datafile with unknown features without using the gnc-features.h functions.

jralls: I'm not sure I understand the question. If "feed" means "load", you can't. That's the whole point of the features check, to prevent GnuCash from opening a file created with features it doesn't know about because it might do something that would break the part of the file that depends on that unknown feature.

If you mean something else, please expand on it.

christopherlam: That's exactly it. It's for the tests only. The `g_assert_nonnull (gnc_features_test_unknown (fixture->book))` cannot be tested in tests.

christopherlam:
I'm not sure you'll agree but I think qofbook.cpp to return a `vector<string>` of unknown features is tidier than returning the full error message.

gjanssens:
> I'm not sure you'll agree but I think qofbook.cpp to return a `vector<string>` of unknown features is tidier than returning the full error message.

That's fine, as long as the contents of the vector<string> are the feature descriptions of the missing features as found in the book. It can't be the feature keys only, because the error message is intended to be displayed for gnucash versions that don't know about these features. So the feature keys won't exist in that gnucash version and neither will the feature descriptions.

PR #1454, opened 2022-10-28
Expose all of the AQBanking Import functionality.

jralls:
Replace the File>Import menu items implemented by AQBanking with a dialog that permits selecting from any of the importers and profiles that AQBanking provides.

This improves things even for the import methods GnuCash currently supports (SWIFT MT940 and MT942 ~and the old German DTAUS~*Edit 2022-10-28 AQBanking no longer supports DTAUS*) because AQBanking provides profiles that customize the import format for particular banks; with the existing menu items those additional profiles aren't available.

This also adds support for a number of other file import formats, most significantly the CAMT format increasingly used instead of SWIFT MT940. That fixes [Bug 797675](https://bugs.gnucash.org/show_bug.cgi?id=797675).


CMakeLists.txt
```diff
@@ -48,7 +48,7 @@ include (TestBigEndian)
 # SQL, put -D WITH_SQL=OFF on the command line.
 
 option (WITH_SQL "Build this project with SQL (libdbi) support" ON)
-option (WITH_AQBANKING "Build this project with aqbanking (online banking) support" ON)
+option (WITH_AQBANKING "Build this project with aqbanking (online banking and SWIFT file imporrt) support" ON)
```
gjanssens, line 51: A small typo: imporrt -> import

gnucash/import-export/aqb/dialog-ab-select-imexporter.c
```diff
@@ -0,0 +1,257 @@
...
+    GtkWidget *select_profile;
+    GtkWidget *ok_button;
+
+    AB_BANKING* abi;
+};
+
+// Expose the selection handlers to GtkBuilder.
+G_MODULE_EXPORT gboolean imexporter_changed(GtkTreeSelection* sel,
+                                            gpointer data);
+G_MODULE_EXPORT gboolean profile_changed(GtkTreeSelection* sel, gpointer data);
```
gjanssens, line 53: It struck me you use G_MODULE_EXPORT here. However looking at our code I see it's used in several locations of the aqbanking code and also in the xml and dbi backends.

Other parts of the code that also work with GtkBuilder don't use this. I also remember a long time ago I added this to assistant-hierarchy.c and had to remove it because it caused issues on Windows.

I had to dig for it but eventually found this explanation in our Changelog.2011:
```
2011-04-23 18:58  cstim
          
        * [r20593] src/gnome-utils/dialog-account.c,
          src/gnome/assistant-hierarchy.c: Win32: Undefine G_MODULE_EXPORT
          in our internal libraries. Please do not use it. The rest of
          gnucash relies on *everything* being exported. This is done by
          mingw's linker, but only if no symbol at all was marked as being
          exported. As soon as at least one symbol is manually marked as
          export, nothing except those marked symbols are exported, which
          will immediately lead to "unresolved symbols" in the win32 build.
```

I have no idea if the same issue happens with MingW64. And also gnucash' modularity has changed a lot since then. Only very few bits are loaded on demand at runtime still: the xml/db backends, and three gnc_modules (aqbanking, ofx and python).

So my question here really is whether we really need it and/or really want it ? As most of the code lives happily without, we can certainly alter the code and/or build system to not require this. But perhaps there are advantages to a controlled exposure of functions. I have no idea at this point.

.github/workflows/ci-tests.yml
```diff
@@ -2,17 +2,17 @@ name: ci-tests
 on: [push, pull_request]
 permissions: {}
 jobs:
-  ci_tests_ubuntu-18:
-    runs-on: ubuntu-18.04
-    name: Ubuntu-18.04 CI Tests
+  ci_tests_ubuntu-22:
+    runs-on: ubuntu-22.04
+    name: Ubuntu-22.04 CI Tests
```
gjanssens, line 7: It's uncommon for gnucash to drop support for the previous LTS (20.04 in this case), but I understand why (lack of recent Aqb version).

And honestly I think the linux ecosystem has evolved over the years. The mindset is becoming more one of having a stable base system with current applications. The LTS' increasingly depend on snaps or flatpaks to provide current applications. So to me as application developer it makes less and less sense to keep old code alive for slow-moving distros. I don't think we have to go all in on bleeding edge either, but I'm quite fine with your current choice - aim for the most recent LTS (though even **that** distro version won't ever package what's currently our master branch - it will only go into the **next** LTS).

gnucash/import-export/aqb/dialog-ab-select-imexporter.c
```diff
@@ -0,0 +1,257 @@
...
+
+static gboolean
+clear_model_row (GtkTreeModel* model, GtkTreePath* path,
+                 GtkTreeIter* iter, gpointer data)
+{
+    char *name, *desc;
+    g_return_val_if_fail (model && iter, TRUE);
+    gtk_tree_model_get (model, iter, 0, &name, 1, &desc, -1);
```
Bob-IT, line 61: I know there are only two columns in model but I would create an enum so names can be used instead of numbers.

gnucash/import-export/aqb/dialog-ab-select-imexporter.c
```diff
@@ -0,0 +1,257 @@
...
+    imexd->parent = parent;
+    imexd->abi = abi;
+
+    g_signal_connect (parent, "destroy",
+                      G_CALLBACK (gtk_widget_destroyed), &imexd->parent);
+    builder = gtk_builder_new();
+    gnc_builder_add_from_file (builder, "dialog-ab.glade", "imexporter-list");
+    gnc_builder_add_from_file (builder, "dialog-ab.glade", "profile-list");
+    gnc_builder_add_from_file (builder, "dialog-ab.glade",
+                               "aqbanking-select-imexporter-dialog");
```
Bob-IT, line 112: Not sure it matters but all the other dialogs in glade file use '_' instead of '-' for name.

gnucash/import-export/aqb/dialog-ab-select-imexporter.c
```diff
@@ -0,0 +1,257 @@
...
+            gtk_widget_set_sensitive (imexd->ok_button, TRUE);
+            return FALSE;
+        }
+
+        if (!profiles->next)
+        {
+            GtkTreePath* path = gtk_tree_path_new_first();
+            GtkTreeSelection* profile_sel =
+                gtk_tree_view_get_selection (GTK_TREE_VIEW (imexd->select_profile));
+            gtk_tree_selection_select_path (profile_sel, path); //should call profile_changed
```
Bob-IT, line 199: I think you need to free the path with gtk_tree_path_free

Bob-IT (commented):
Looks OK with a couple of minor suggestions.
Originally I could not build this, so upgrade my gwenhwfar to 5.6.0 but still would not build despite having aqbanking 6.2.5, so upgraded to 6.4.0 and all was OK.

Is the minimum version correct at 6.1.0?

gnucash/import-export/aqb/dialog-ab-select-imexporter.c
christopherlam: Even better: I doubt that this whole `clear_model_row` is needed in the first place -- its purpose is to `g_free` the model data, however, `gtk_list_store_clear` will take care of it. AFAIU The `g_free` here is simply freeing the copies provided by `gtk_tree_model_get`.

gnucash/import-export/aqb/dialog-ab-select-imexporter.c
jralls: I copied that over from dialog-ab-trans.c because I was having trouble getting the gtkbuilder signal connecting to work. It didn't help and I switched to connecting the signal in code, but didn't remove the `G_MODULE_EXPORT` decl. Of course changing it to static also works because it's completely internal now. That should also be true of dialog-ab-trans.c as it also explicitly connects the signals in code instead of with gtkbuilder.

On the other hand, the uses in the backends date back to Andreas Köhler and Phil Longstaff in 2006 and 2008 respectively while the ones in dialog-ab-trans.c are from a patched that Christian merged a few months *after* writing that commit. All were compiling cleanly before the switch to MSYS2 and of course still do so I'm not entirely sure that Christian's commit comment is correct.

.github/workflows/ci-tests.yml
jralls: Well, for one thing this isn't about GnuCash supporting an older distro, it's that the older distro image in Github's workflows doesn't provide new enough AQBanking.  A user of that older distro can build a new AQBanking and then build GnuCash 5.

Next problem: We're going to [have to remove](https://github.blog/changelog/2022-08-09-github-actions-the-ubuntu-18-04-actions-runner-image-is-being-deprecated-and-will-be-removed-by-12-1-22/) the 18.04 runner in maint in a few weeks; fortunately that's at the end of our dev cycle for the current maint anyway. 20.04 will be similarly killed off in December 2024.

We do need to have the broader discussion about what our dependency requirements policy should be, but I think we should do so on gnucash-devel instead of on a random PR.

gnucash/import-export/aqb/dialog-ab-select-imexporter.c
jralls: You're both right.

gnucash/import-export/aqb/dialog-ab-select-imexporter.c
jralls: They're just strings to the compiler and GtkBuilder doesn't seem to care. Does Glade care? (I don't use it.)

gnucash/import-export/aqb/dialog-ab-select-imexporter.c
jralls: Right, thanks.

jralls:
> Looks OK with a couple of minor suggestions. Originally I could not build this, so upgrade my gwenhwfar to 5.6.0 but still would not build despite having aqbanking 6.2.5, so upgraded to 6.4.0 and all was OK.
> 
> Is the minimum version correct at 6.1.0?

No, sigh. I misread `#if AQBANKING_VERSION_INT >= 60400` as meaning newer than v6.0.4 and it's of course v6.4.0. 

6.1.0 isn't correct for another reason: The [Bug798385](https://github.com/Gnucash/gnucash/commit/1b384df62249d0f4601858a9839d583dc33f18e7) change depends on a feature in 6.2.0.

It doesn't really matter for the purposes of this PR. If I restore the version exclusion for `SWIFT_INTERNAL_TRANSFER" we can make the cutoff 6.2.0 which is sufficient for the current Debian stable. Debian testing, which is rumored to be released in January, is fully up to date with AQB 6.5.3 so we could just as reasonably make the requirement 6.4.0 and leave out the version checks.

CMakeLists.txt
jralls: Thanks. Since the file import is about a lot more than SWIFT now I changed that too.

gnucash/import-export/aqb/dialog-ab-select-imexporter.c
```diff
@@ -0,0 +1,245 @@
...
+    {
+        char* name = NULL;
+        GList* profiles = NULL;
+
+        gtk_tree_model_get (model, &iter, NAME_COL, &name, -1);
+        if (name && *name)
+        {
+             profiles = gnc_ab_imexporter_profile_list (imexd->abi, name);
+             g_free (name);
+        }
```
christopherlam, line 169: To prevent leakage whereby `name==""`
```suggestion
        if (name && *name)
             profiles = gnc_ab_imexporter_profile_list (imexd->abi, name);
        g_free (name);
```

.github/workflows/ci-tests.yml
gjanssens: Ok. If you start that discussion, I'll chime in.

gnucash/import-export/aqb/dialog-ab-select-imexporter.c
jralls: Thanks!

gnucash/import-export/aqb/dialog-ab-select-imexporter.c
jralls: I'll leave this in for now, it's easily changed if Glade gives you trouble.

.github/workflows/ci-tests.yml
jralls: > Ok. If you start that discussion, I'll chime in.

Did so.

gnucash/import-export/aqb/dialog-ab-select-imexporter.c
jralls: I've replaced `G_MODULE_EXPORT` with `static` in both files in aqb. I'll try the backends later, that's obviously out of scope for this PR.

jralls:
I changed the AQB and Gwen required versions to 6.4.0 and 5.8.0 respectively. That corresponds to Fedora 35 updates, so I think it's reasonable for something we'll release next year. If we get significant pushback then it's not too hard to roll back to 6.2 and 5.6.

I've squashed the fixup commits and I'll merge as soon as CI finishes. Thanks, everybody, for your review and comments!

gnucash/import-export/aqb/dialog-ab-select-imexporter.c
gjanssens: Good.

As to Christian's commit comment. I think it was factual. The fact that it wasn't an issue in the places where G_MODULE_EXPORT was used, was because other parts of gnucash effectively only used these exported symbols for those modules. So no issues there. But at the time of his commit I had accidentally introduced use of G_MODULE_EXPORT in parts of gnucash that were never written with that in mind. And suddenly other functions in the same source files as I had added the macros were no longer visible, but still called - causing linker errors on Windows. Those had to be removed again.

jralls: OK, I guess. Maybe that's changed, because having the symbols in AQB didn't affect the Windows linkage of libgncmod-aqbanking.dll. After all, the symbols with declspecs didn't have external usage at all, they were all callbacks connected to signals in the same files, but there were legitimately public symbols that linked correctly and weren't marked with declspecs.

In the case of the two backends, the marked symbols are the only ones used publicly, though there are unmarked aliases intended for initializing the backends if we were to statically link them.

PR #1456, opened 2022-10-30
Remove depreciated GtkUIManager and friends from the main window, TAKE2

Bob-IT:
To implement Geert's changes I decided to start again and use the old commits as a basis for these ones. The main thing is that I am trying to use GMenuModel and GMenu for the menus.

The command call backs have been changed to use GActions along with being part of GSimpleActionGroups.

Basically the menu can be thought of two half's, the action side which is connect to the command call backs and the menu.
The main window menu is defined in the `gnc-main-window.ui` which has hidden menu items deemed as place holders. Each of the plugin files like `gnc-plugin-basic-commands.c` has an array `gnc_plugin_load_ui_items` which is a list of placeholders to be updated. Each plugin has it's own `ui` file containing menu fragments and these fragments are loaded into the main window when these plugins are loaded and can be thought of as permanent menu entries.

When a plugin page is loaded, the focus widget call back is used to call `gnc_main_window_update_menu` which adds the page specific toolbar and also uses the placeholder array `gnc_plugin_load_ui_items` to add the page specific menu items, these are deemed temporary. When the page is switched, these temporary menu items are removed and new ones loaded according to the new page.

Most of the menu Translations have moved to the `ui` files but there are some left in the source where menu labels are changed.
Tool tips are defined as an extra attribute in the GMenuModel item and are retrieved for displaying in the status bar.

Part of the main window `ui` file is shown below along with a menu fragment.

         <item>
          <attribute name="label" translatable="yes">Tutorial and Concepts _Guide</attribute>
          <attribute name="action">mainwin.HelpTutorialAction</attribute>
          <attribute name="accel">&lt;Primary&gt;h</attribute>
          <attribute name="tooltip" translatable="yes">Open the GnuCash Tutorial</attribute>
        </item>
        <item>
          <attribute name="label" translatable="no">ToDPlaceholder</attribute>
          <attribute name="action">mainwin.HelpPlaceholder1</attribute>
          <attribute name="hidden-when">action-disabled</attribute>
        </item>

 ```
  <menu id="HelpPlaceholder1">
    <item>
      <attribute name="label" translatable="yes">_Tips Of The Day</attribute>
      <attribute name="action">gnc-plugin-basic-commands-actions.HelpTipsOfTheDayAction</attribute>
      <attribute name="tooltip" translatable="yes">View the Tips of the Day</attribute>
    </item>
  </menu>
```

Every thing seems to work but needs further testing. There are some things still to complete...
Convert the recent file list to be updated with GMenu calls.
Convert the Business menu updates to use GMenu calls.
Rework the Window menu list, this has been converted but I think needs some more work.
Also noticed some changes and refactoring that is needed.


jralls:
Amazingly the menus still work on macOS. I'd thought that gtkosx_application didn't work with GMenu, but I guess that there are enough remaining GtkMenu hooks for it to still make the connections.

jralls:
There's still some GtkUIManager code in gnome/window-reconcile.c.

Bob-IT:
> There's still some GtkUIManager code in gnome/window-reconcile.c.

That was covered in PR #1377, maybe I will move all of that to this one and have only on PR going forward.

Bob-IT:
I have moved the commits dealing with the reconcile window to this PR so there is only one PR dealing with the removal of GtkUIManager.

jralls:
@Bob-IT This is a substantial modernization. Thanks!

gjanssens:
I didn't have time to review before merge. Sorry about that.

I agree this is a great step forwards!

You can consider the amendments I pushed today as a partial review :)

The only other thing that struck me in my limited review is that gnc-main-window.ui contains menu definitions for menus that are only relevant for certain types of windows/tabs. For example the "Transaction" menu is only relevant for register-type tabs. Is that really necessary to define it top-level ? Can't it be added by the relevant plugin as a whole ? I vaguely remember seeing Gtk or GMenu related functions that would allow to insert a menu right after an existing one. So can't we just have the plugin add the Transaction menu after the View menu straight away, rather than requiring a placeholder-filled Transaction menu in the top level gnc-main-window.ui ? My concern here is locality of responsibility.

PR #1458, opened 2022-11-01
[gtest-qofevent.cpp] basic test for qofevent

christopherlam:
includes unusual macos tests discrepancy... when `TEST (qofevent, events)` completes I'd forgotten to unregister the last `id4` handler, which means `TEST (qofevent, compound_events)` would run with `id4` still registered. With ubuntu the `compound` tests increments data once, but on mac it increments data twice. The commented-out line 102 is crucial for macos tests. I think mac is right.


christopherlam:
combined into #1365

PR #1459, opened 2022-11-01
Bug 765920 - Running balance on report is not accurate when report is sorted in different ways

dawansv:
For https://bugs.gnucash.org/show_bug.cgi?id=765920

This bug is caused by the fact that the current running balance option pulls the running balance from the underlying account for each split. It has the advantage of also providing a starting prior balance from the account as well, but the disavantage that it only really makes sense if the transaction report is sorted by account and by ascending dates. Any other sort order and the running balance appears out of order.

So I added an option to either get a running balance as per the underlying accounts (the current option) or get a running balance calculated based on the report sort order which will then accomodate any sort variation (but will start at the first transaction, without a prior balance). I kept the original option as well because it does offer a prior balance as starting balance while my option cannot do that since it is disconnected from any account. Both make sense depending on what type of report is being made, so it's up to the user to decide.

I am new to scheme so ended up creating a global variable to keep track of the running balance as it changes going through the splits. I could not find a way to do it by passing the current running balance back into the loop; I am not clear how that part works yet (or if it's even possible). I am not entirely clear whether it's a problem or not to use a global variable here. If There is better way to do that part please point me in the right direction I am happy to make changes.

### DIFF
--- a/libgnucash/engine/test/test-qofbook.c
+++ b/libgnucash/engine/test/test-qofbook.c
@@ -773,6 +773,7 @@ test_book_get_collection( Fixture *fixture, gconstpointer pData )
 static void
 test_book_features (Fixture *fixture, gconstpointer pData)
 {
+    char* msg;
     g_test_message ("Testing book features");
 
     g_assert_null (gnc_features_test_unknown (fixture->book));
@@ -786,10 +787,18 @@ test_book_features (Fixture *fixture, gconstpointer pData)
     g_assert_null (gnc_features_test_unknown (fixture->book));
     g_assert_false (gnc_features_check_used (fixture->book, "Credit Notes"));
 
-    /* cannot set an unknown feature: it bails out. */
-    /* gnc_features_set_used (fixture->book, "Nanotech"); */
-    /* g_assert_nonnull (gnc_features_test_unknown (fixture->book)); */
-    g_assert_false (gnc_features_check_used (fixture->book, "Nanotech"));
+    /* cannot use gnc_features_set_used to set an unknown feature: it bails out.
+     * use qof_book_set_feature instead. */
+    qof_book_set_feature (fixture->book, "Nanotech", "With Quantum Computing");
+    g_assert_true (gnc_features_check_used (fixture->book, "Nanotech"));
+    msg = gnc_features_test_unknown (fixture->book);
+    g_assert_cmpstr (msg, ==, "This Dataset contains features not \
+supported by this version of GnuCash. You must use a newer version \
+of GnuCash in order to support the following features:\n* With Quantum Computing");
+    g_free (msg);
+
+    qof_book_unset_feature (fixture->book, "Nanotech");
+    g_assert_null (gnc_features_test_unknown (fixture->book));
 }
 
 static void


gnucash/report/trep-engine.scm
```diff
@@ -1032,7 +1032,21 @@ be excluded from periodic reporting.")
...
+      "n" (G_ "Display a running balance?")
+      running-balance
+      (list
+       (vector 'none   (G_ "None"))
+       (vector 'account (G_ "Retrieve from Underlying Accounts"))
+       (vector 'report (G_ "Calculate as per Report Sort Order")))
+      #f
+      (lambda (x)
+        (set! running-balance x)
+        (apply-selectable-by-name-display-options)))))
```
christopherlam, line 1049: A `gnc:make-multichoice-callback-option` is only needed if the option _must_ e.g. toggle other options' sensitivities. A `gnc:make-multichoice-option` is more than sufficient here. There's no need for the `running-balance` variable being kept and modified.

christopherlam (commented):
This is not a bad first attempt, however:
- it sums the splits' amounts on each call to `running-balance-report` - how should it sum the balances if the amounts are in different currencies? e.g. the accounts include your stock portfolio
- there are two nested levels of sorting and grouping; should the running balance be reset at each level? (or not?)

gnucash/report/trep-engine.scm
dawansv: Thanks for the pointer, that makes sense. I "borrowed" the code from the "Hide/Single column/Two column" above but you are right that option does more than mine. Will convert to the simpler gnc:make-multichoice-option

dawansv:
> it sums the splits' amounts on each call to running-balance-report - how should it sum the balances if the amounts are in different currencies? e.g. the accounts include your stock portfolio

Good point. I am going to do more testing in the context of multi-currency/stock

> there are two nested levels of sorting and grouping; should the running balance be reset at each level? (or not?)

Yes I considered this question as well because indeed we have the 2 nested levels+ the grand total, so actually 3 levels. Right now the running balance (the one I implemented) applies to the grand total only. I think that needs to be offered at a minimum, and its location right now in the UI on the display tab right above the "Totals" options is good because it only relates to that total.

On a side note that "Totals" option might be better named "Grand Total" because as far as I can see it only controls the Grand Total. The options for subtotals are on the Sort tab and independant to this one).
 
So then the next question would be whether there should be an option to also have a running balance on the other levels, similar to the total/subtotal option. The subtotal options are in the sort tab, so that would be I think were to put the other running balance option. In that case I suppose we might even offer to display several running balances, one for each level, matching each subtotal and grant total. 

I'll first focus on fixing the other stuff and finishing the implementation for grant total, then see what I can do for nested level. Possibly others might have some more feedback as to what would be the expectation here in regards to nested levels. 

There is an alternative approach as well. The "Running balance as per underlying account" could be kept as a completely separate option, and possibly the column heading would be better named "Account Balance" and the option explained as "Show account balance for each transaction". Then the "Running balance as per report" could be offered separately at each level and renamed "Running Total" to more deeply connect it with the total since it does in fact match the total at the last transaction. Now that I think about it, this might be a better approach. I am going to let it simmer a bit.

dawansv:
Just a quick update. Now that I have spent more time understanding the code, my solution here is definitely very rudimentary and can be ignored for now as it falls apart completely in the multi-currency context.

What I need to do is plug into the total and subtotal rountines and integrate this "running total" feature into that. The existing code is already collecting total and subtotal information in 3 total/subtotal collectors (one for each level) and dealing with the multicurrency issue. So in essence the running total is nothing more than calculating the total from these collectors at each split as it loops through the splits and storing it in the corresponding "runing total" columns. At the core this need to be implemented through a routine very similar to add-subtotal-row with the exception that it would be calculating and populating the runing total columns on each row instead of adding the total/subtotal row at the bottom of each section. 

We would still need to decide how to display the running total in the multicurrency context. I think repeating each commodity on each row (the way the total/subtotals row does it) is going to be super crowded. What might make more sense is to only show the running total for the current commodity on each row, with the total/subtotal line showing all of them (as it currently does). So using a simplified example it would show like this.

```
Date           Amount          Running Total
01/01/2021      EURO 100         EURO 100
02/01/2021      EURO 150         EURO 250
03/01/2021      USD  100         USD  100
04/01/2021      EURO  50         EURO 300
05/01/2021      USD   50         USD  150
--------------------------------------------
GRAND TOTAL                      EURO 300
                                 USD  150
```

Also in the 2 column format (debit/credit) the total woud be both the running total column and the credit or debit column (with running total signs inverted if credit). For the single column shown, I am not sure whether it's better to put it only under Running total (as shown), only under amount, or both. Seems like a small detail that can be decided later because there is a lot more work to do before getting anywhere close to a solution. But at least now I understand more specifically what needs to be done.

dawansv:
See next part with updated code here https://github.com/Gnucash/gnucash/pull/1460

PR #1460, opened 2022-11-05
Bug 765920 - Running balance on report is not accurate when report is sorted in different ways.

dawansv:
I think I got it done. I ended up using several gnc:make-commodity-collector which really works well for this purpose. It is fully multi-currency compliant. I ended up keeping the existing Running Balance as it was but renamed it Account Balance to avoid confusion (and also it's a better name because it is only a running balance when sorted by account and by register order; in any other sorts it doesn't work as a running balance). For the Running Totals, the main one (grand totals) are implemented under the Display option and the subtotals are implemented under the sorting option where the subtotals options are.

I got into a bit of a "fight" with github trying to get my changes to show up into a single commit so it's easier to review.

### DIFF
--- a/gnucash/report/trep-engine.scm
+++ b/gnucash/report/trep-engine.scm
@@ -79,6 +79,7 @@
 (define optname-prime-subtotal (N_ "Primary Subtotal"))
 (define optname-prime-sortorder (N_ "Primary Sort Order"))
 (define optname-prime-date-subtotal (N_ "Primary Subtotal for Date Key"))
+(define optname-running-prime (N_ "Running Primary Subtotal"))
 (define optname-full-account-name (N_ "Show Full Account Name"))
 (define optname-show-account-code (N_ "Show Account Code"))
 (define optname-show-account-description (N_ "Show Account Description"))
@@ -90,6 +91,7 @@
 (define optname-sec-subtotal (N_ "Secondary Subtotal"))
 (define optname-sec-sortorder  (N_ "Secondary Sort Order"))
 (define optname-sec-date-subtotal (N_ "Secondary Subtotal for Date Key"))
+(define optname-running-sec (N_ "Running Secondary Subtotal"))
 
 ;;General
 (define optname-startdate (N_ "Start Date"))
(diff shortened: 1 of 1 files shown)


dawansv:
Screenshot showing all 3 subtotal level with a dual common/original currency display
All running totals and subtotals

christopherlam:
Not bad at all! A couple comments to heat up the debate further:
1. How would the grouping & totals work when the primary sort key is "Other Account Name" - does it still display a reasonable transaction report?
2. We could argue that the Grand Total should show all currencies that they are tallying. (Ditto Secondary and Primary?)
3. There must be further changes in `options.scm` to ensure backward compatibility with saved reports which still expect to set `Running Balance` and `Totals` options.

jralls:
> There must be further changes in options.scm to ensure backward compatibility with saved reports which still expect to set Running Balance and Totals options.

Bad idea, you'll have to redo the work for master.

This is anyway too significant a change for the last 2 releases of 4.x.

dawansv:
> This is anyway too significant a change for the last 2 releases of 4.x.

Hi John. Yes I agree. 

As I got into fixing bug 765920 I realized that:

1. Bug 765920 is not really a bug. It's some users (including me) misunderstanding what the "running balance" option does. Now I understand that all it does is pull the balance of the underlying account for each transaction. That said it could be tweaked a bit (see below).
2. What I wanted (and confused with the running balance option) is a running total option which gives the running total for whatever transactions are returned by the report engine in whatever order they are returned. By definition that running total would build up to match the existing grand total shown at the bottom, or existing primary/secondary total for subtotal groups. 

With that in mind, this is what I propose to do.
1. For Bug 765920, I still think changing the option name and column heading from "Running Balance" to "Account Balance" would avoid confusion as to what this option does. I might even look at improving my code by changing the column heading in the report from "Account Balance" to "Running Balance" only when the report is sorted by the order of the underlying account (hence the balance shown is actually running). These is already code to add a balance forward in that case, so it should be easy enough. So I will still submit a PR just for that (it's not many changes) and then a discusion on that very specific point can take place there, and ultimately you can decide what to approve or reject. 
2. My code here also fixes another unreported unrelated bug with the Running Balance feature, relating to common currency conversion. The current code doesn't convert the retrieved balance to the currency chosen in the report. I will create a separate bug report for that issue and submit a sepated PR as well.
3. I will create an enhancement request for the new "running total" feature and then refer to that when I am ready to submit a PR for Master. Further discussions on this new feature can happen in that PR.

John, if this works for you then please close this PR for now and I'll see all of you later in the other PRs. If you have any suggestion about doing things differently let me know.

Christopher I will answer your follow-up question/comments in the other PRs.

To both of you, thank you for your feedback and patience (and your incredible contribution to gnuCash of course).

jralls:
@dawansv A big difference indeed between running balance and running total. I agree that "Account Balance" or perhaps "Account Balance at Transaction" makes more sense in cases where the report isn't a comprehensive list of transactions in standard order from a single account, though maybe it would make even more sense to display nothing at all in that case. It's a lot more likely to be confusing than useful. Further discussion should go on the bug report.

PR #1461, opened 2022-11-08
Translations update from Hosted Weblate

weblate:
Translations update from [Hosted Weblate](https://hosted.weblate.org) for [GnuCash/Program](https://hosted.weblate.org/projects/gnucash/gnucash/).

It also includes following components:

* [GnuCash/Glossary](https://hosted.weblate.org/projects/gnucash/glossary/)

Current translation status:

Weblate translation status

PR #1462, opened 2022-11-13
Update pl.po

Kaligula0:
correct critical errors, as for https://github.com/Gnucash/gnucash/pull/1461#issuecomment-1312606320


fellen:
Same content like PR #1463. We prefer weblate to avoid merge conflicts.
Thanks!

PR #1464, opened 2022-11-14
Add investment-lots.scm report

mcbridebt:
# Investment Lots report

## Summary
image

Provides detailed info about investment lots across multiple accounts. Includes purchases, sales, current balances, gains (realized and unrealized, short term and long term), return on investment (ROI), and various validation checks. 

Can help answer questions such as:
- Which lots would be the best candidates for tax loss/gain harvesting?
- Which lots held longer than one (or more) years have the largest unrealized gains, and would therefore be most advantageous (tax-wise) if donated to charity?
- What (taxable?) capital gains have been realized for a particular time period (i.e. this year, last year, etc)? Are they long term or short term gains?
- Are my lots (and related accounts and splits) well-formed?

## Report Options
The report is highly configurable and includes the following options.

### Accounts tab
Standard accounts selector, commonly used in other reports. 
Has additional options to filter out accounts with no shares and/or no lots.
image
The selection is filtered to only include accounts of type stock or mutual fund, similar to the **Investment Portfolio** and **Advanced Portfolio** reports.
Tip: Select only taxable accounts to scope the report to taxable capital gains.

### Chart tab
Controls the chart that summarizes capital gains for all lots in all report accounts.
image

**Show Chart**
Default: true

**Chart type**
`Stacked Bar Chart`
image
	
`Bar Chart`
image

**Chart location**
`Top`: Above the table
`Bottom`: Beneath the table

**Plot Width**
**Plot Height**
Controls chart dimensions

Note that the chart gain categories are controlled by the related **Columns** options

### Columns tab
Various options to control which columns to show in the report table.
image

The report is quite complex with many columns available. But if only a subset are desired, use these options to hide the others and make the report less busy.

Note that the gains related options at the bottom impact both the table and the chart.

**Group gains by age (short term and long term)**
Different governments may have different rules regarding how long shares must be held to qualify for different tax treatment. 
Set this option to `Gains Only` to group gains by age (in both table and chart.)
Set this option to `Gains and Sales` to additionally group the Sold columns by age:
image

**Long term gains age (years)**
Default: 1 year.
Shares held longer than this many years count as long term. This option is ignored when the **Group gains by age (short term and long term)** option is disabled.

### Display tab
Miscellaneous display options
image

**Show long account names**
Controls which account name to use in the report: long (fully-qualified) or short.

**Show mnemonic in amounts**
Controls whether the currency mnemonic will be use when reporting amounts. I.e. "123 RIO" versus just "123". This setting only impacts amounts not values, which are always displayed in the report's currency, with its mnemonic.

**Include closed lots**
Allows closed lots to be filtered out. This can be useful when only wanting to look at unrealized gains.

**Show blanks instead of zeros in table cells**
This can make the report look less cluttered.

zeros:
image

blanks:
image

**Show lot split rows**
When selected, rows for each of the bought and sold splits pertaining to the lot (within the report's date window) will be displayed (indented) after each lot row: 
image

### General tab
image

The **Start date** and **End date** define the lot's date window. Splits outside of this window are not included in the report. Except: Splits from before the start date are counted only insofar as needed to ensure correctness of **Sold/End Basis** and **End Amount** columns. But they are not included in the report's **Bought** and **Sold** columns.

### Validation tab
The report can show various validation warnings that can help identify potential lot problems. These warnings are optional and may be toggled in this tab.
image

**Include only accounts with warnings**
Disabled by default.
Set this option to see only accounts that have one of more of the selected warnings. Can be useful for quickly validating that lots are well-formed, especially in conjunction with selecting all accounts and all dates.

**Warn if a lot has more than one bought split**
As per GnuCash documentation, in the "Lots in Account" window, the "Scrub Account" button:
> Creates a lot for each buy transaction split that is not already linked to a lot.
> 	
> 	A new lot is created for each unlinked buy transaction split because gains on investments, in some jurisdictions, may be taxed at different rates depending on how long they were held before being sold. As a separate capital gains transaction is created for each lot, this makes it easier to determine tax on the capital gains based on the lot opened and closed dates.
> 	
> 	If you know that the capital gains from a sale are all to be taxed at the same rate, you can manually link multiple buys to a lot so scrubbing will only create 1 capital gains transaction.
https://www.gnucash.org/docs/v4/C/gnucash-help/tool-lots.html

Also:

> two buy transactions create different lots (because you can't add to a stock-investment lot after its been created, because the purchases occurred on different dates)
https://code.gnucash.org/docs/MAINT/lotsoverview.html

Actually, you can in GnuCash. The **Scrub** operation may only add one purchase per lot, but GnuCash does not stop a user from manually adding multiple purchase splits to the same lot. Even though this is arguably bad practice. 

Having multiple purchase dates makes the holding period ambiguous. But if this is desired, consider disabling both this warning and the **Columns** -> **Group gains by age (short term and long term)** option. If the column option is left enabled, then the report will use the date of the _last_ purchase split for determining the holding period, but continue to use the first purchase date for the lot **Opened** column. 

Why use the last date for the holding period? Because long term gains often receive favorable tax treatment, so it is probably worse to misidentify a gain as long-term rather than as short-term. But better still is to not have more than one purchase per lot, so the holding periods are unambiguously correct. 

Example:
`Warning: Above lot has 2 bought splits. Consider separating them into their own lots.`

**Warn if a lot's balance drops below zero**

> A lot is typically formed when an item is bought, and is closed when the item is sold. A lot need not be a single item, it can be a quantity of the same thing e.g. 500 gallons of paint (sold off a few gallons at a time)… A lot is 'closed' when the number of items in the lot has gone to zero.
https://wiki.gnucash.org/wiki/Concept_of_Lots

The **Scrub** operation won't create negative lot balances. But GnuCash does not prevent users from manually creating them. The report detects if at any time during the report date window the lot's balance became negative. 
Example:
`Warning: Above lot's balance is negative on 10/29/2022. Consider removing the responsible sale split from the lot and then scrubbing.`

**Warn if a lot has a blank title**
Disabled by default.
Lot titles are optional. Turning on this warning will flag lots with blank (empty or whitespace) titles.
Example:
`Warning: Above lot's title is blank.`

**Warn if the "Realized Gain/Loss" split(s) sum does not match the computed gains**
> The computation of (Realized) Gains/Losses is performed automatically by the lot 'scrub' routines
https://wiki.gnucash.org/wiki/Concept_of_Lots

The report recomputes the gain/loss amounts from each sale, compares them with the sum of the lot's generated "Realized Gain/Loss" splits and flags any found discrepancies. Example:
`Warning: Above lot's computed gain $(1,518.79) is not equal to the "Realized Gain/Loss" split(s) sum $(1,518.80). Difference: $0.01`

I have observed many instances, in my own GnuCash db, of these generated "Realized Gain/Loss" having incorrect values. It is unclear whether this corruption is caused by user error or code bug (or both). But when it happens, it can often be fixed by:
	1. Removing the sale split(s) from the lot
	2. Manually deleting the corrupt "Realized Gain/Loss"  transaction(s),
	3. Adding back the sales split(s) to the lot.

**Warn if a split is not assigned to a lot**
Realized and unrealized gains cannot be computed for splits that are not in a lot. This warning flags accounts containing such orphans.
Set this option to **Count** for the warning message to only identify the number of orphaned splits. Example: 
`Warning: 2 split(s) are not assigned to a lot. Do lots need to be scrubbed?`
Set this option to **List** to also list each one after the summary message.

**Warn if the account balance does not match the computed lots' end balance**
Example:
`Warning: End amount 236.5 RIO is not equal to actual account balance 234 RIO. Difference: 2.5 RIO. Do lots need to be scrubbed?`

This is often another indication that the account has splits that are not included in a lot.

------------------------------------------------------------------------------------------------

# Report genesis:

------------------------------------------------------------------------------------------------
From: Christopher Lam <redacted>
Sent: Saturday, September 3, 2022 9:54 PM
To: Brent McBride 
Subject: Re: [GNC] Lots report?
 
No such report exists. Please feel free to relearn scheme and submit as a PR on GitHub.

You may find the Lot Viewer report (exposed only when running GnuCash with --extra) a useful starting point.

Good luck!

On Sun, 4 Sept 2022, 12:27 pm Brent McBride wrote:
Hi GnuCash experts,

Does GnuCash have a report that shows the current unrealized gain/loss of open lots?

Such a report would help answer questions such as:

  *   Which lots held longer than one year have the largest unrealized gains, and would therefore be most advantageous (tax-wise) if donated to charity?
  *   Which lots would be the best candidates for tax loss/gain harvesting?

The Lots UI has almost everything needed to answer these questions: Date opened, balance, realized gains. But it is missing one key piece of info: Unrealized gains (what would the gain/loss be if the remaining balance was sold at the most recent price?).

Further, the Lots UI only shows a single commodity. The ideal would be a report that shows open lots in all accounts (perhaps configurable by date, accounts, gain/loss amount, etc). I thought maybe one of the existing reports would have this info, but haven't yet found one.

The "Investment Portfolio" report seems close. It shows accounts, but not lots within the accounts.

Any suggestions?

My background:
I recently moved ~20 years of data from Quicken to GnuCash, and am still learning the ropes.

I've written some scripts using the super-useful piecash and openpyxl tools to update/create various Excel docs from my GnuCash db. I suppose I could write a similar script to generate a Lots report in Excel.

Or if I'm in the mood for pain, I could relearn Scheme enough to create a custom GnuCash report (I briefly taught Scheme as a TA in college over 20 years ago—all I remember now is it uses lots of parentheses 🙂).

But before going either route, I want to verify with you experts that there is no preexisting report.

Thanks for your help!

-Brent
_______________________________________________
gnucash-user mailing list

### DIFF
--- a/gnucash/report/reports/standard/investment-lots.scm
+++ b/gnucash/report/reports/standard/investment-lots.scm
@@ -1,0 +1,2072 @@
+;; -*-scheme-*-
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;; investment-lots.scm
+;; by Brent McBride (mcbridebt@hotmail.com) Nov 2022
+;;
+;; This program is free software; you can redistribute it and/or
+;; modify it under the terms of the GNU General Public License as
+;; published by the Free Software Foundation; either version 2 of
+;; the License, or (at your option) any later version.
+;;
+;; This program is distributed in the hope that it will be useful,
+;; but WITHOUT ANY WARRANTY; without even the implied warranty of
+;; MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+;; GNU General Public License for more details.
+;;
+;; You should have received a copy of the GNU General Public License
+;; along with this program; if not, contact:
+;;
+;; Free Software Foundation           Voice:  +1-617-542-5942
+;; 51 Franklin Street, Fifth Floor    Fax:    +1-617-542-2652
+;; Boston, MA  02110-1301,  USA       gnu@gnu.org
+;;
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+
+
+(define-module (gnucash reports standard investment-lots))
+
+(use-modules (ice-9 format))
+(use-modules (ice-9 match))
+(use-modules (gnucash app-utils))
+(use-modules (gnucash core-utils))           ;for gnc-prefs-is-extra-enabled
+(use-modules (gnucash engine))
+(use-modules (gnucash html))
+(use-modules (gnucash report))
+(use-modules (gnucash utilities))
+(use-modules (srfi srfi-1))
+(use-modules (srfi srfi-11))            ;for let-values
+(use-modules (srfi srfi-13))            ;for string-trim
+
+(define pagename-chart (N_ "Chart"))
+(define pagename-columns (N_ "Columns"))
+(define pagename-validation (N_ "Validation"))
+
+;; Accounts
+(define optname-accounts (N_ "Accounts"))
+(define optname-zero-shares (N_ "Include accounts with no shares"))
+(define optname-include-lotless-accounts (N_ "Include accounts with no lots"))
+
+;; Chart
+(define optname-show-chart (N_ "Show Chart"))
+(define optname-chart-type (N_ "Chart type"))
+(define optname-chart-location (N_ "Chart location"))
+(define optname-plot-width (N_ "Plot width"))
+(define optname-plot-height (N_ "Plot height"))
+
+;; Columns
+(define optname-show-lot-guid-column (N_ "Show lot guid column"))
+(define optname-show-date-columns (N_ "Show date columns"))
+(define optname-show-bought-columns (N_ "Show bought columns"))
+(define optname-show-sold-columns (N_ "Show sold columns"))
+(define optname-show-end-columns (N_ "Show end columns"))
+(define optname-show-realized-gain-columns
+    (N_ "Show realized gain column(s)"))
+(define optname-show-unrealized-gain-columns
+    (N_ "Show unrealized gain column(s)"))
+(define optname-group-gains-by-age
+    (N_ "Group gains by age (short term and long term)"))
+(define optname-long-term-years (N_ "Long term gains age (years)"))
+
+;; Display
+(define optname-show-long-account-names (N_ "Show long account names"))
+(define optname-show-mnemonics (N_ "Show mnemonic in amounts"))
+(define optname-include-closed-lots (N_ "Include closed lots"))
+(define optname-show-blanks-for-zeros
+    (N_ "Show blanks instead of zeros in table cells"))
+(define optname-show-split-rows (N_ "Show lot split rows"))
+
+;; General
+(define reportname (N_ "Investment Lots"))
+(define optname-from-date (N_ "Start date"))
+(define optname-to-date (N_ "End date"))
+(define optname-report-currency (N_ "Report's currency"))
+(define optname-price-source (N_ "Price source"))
+
+;; Warnings
+(define optname-include-only-accounts-with-warnings
+    (N_ "Include only accounts with warnings"))
+(define optname-warn-if-multiple-bought-splits
+    (N_ "Warn if a lot has more than one bought split"))
+(define optname-warn-if-balance-negative
+    (N_ "Warn if a lot's balance drops below zero"))
+(define optname-warn-if-lot-title-blank
+    (N_ "Warn if a lot has a blank title"))
+(define optname-warn-if-gains-mismatch
+    (N_ "Warn if the 'Realized Gain/Loss' split(s) sum does not match the computed gains"))
+(define optname-warn-type-if-split-not-in-lot
+    (N_ "Warn if a split is not assigned to a lot"))
+(define optname-warn-if-balance-mismatch
+    (N_ "Warn if the account balance does not match the computed lots' end balance"))
+
+(define colname-lot-title (N_ "Lot Title"))
+(define colname-opened (N_ "Opened"))
+(define colname-closed (N_ "Closed"))
+(define colname-lot-guid (N_ "GUID"))
+(define colname-bought-amount (N_ "Bought Amount"))
+(define colname-bought-value (N_ "Bought Value (Basis)"))
+(define colname-bought-price (N_ "Bought Average Price"))
+(define colname-sold-splits (N_ "Sold Splits"))
+(define colname-sold-amount (N_ "Sold Amount"))
+(define colname-sold-basis (N_ "Sold Basis"))
+(define colname-sold-value (N_ "Sold Value"))
+(define colname-short-term-sold-amount (N_ "ST Sold Amount"))
+(define colname-short-term-sold-basis (N_ "ST Sold Basis"))
+(define colname-short-term-sold-value (N_ "ST Sold Value"))
+(define colname-long-term-sold-amount (N_ "LT Sold Amount"))
+(define colname-long-term-sold-basis (N_ "LT Sold Basis"))
+(define colname-long-term-sold-value (N_ "LT Sold Value"))
+(define colname-sold-price (N_ "Sold Average Price"))
+(define colname-end-amount (N_ "End Amount"))
+(define colname-end-basis (N_ "End Basis"))
+(define colname-end-value (N_ "End Value"))
+(define colname-realized-gain (N_ "Realized Gain"))
+(define colname-short-term-realized-gain (N_ "ST Realized Gain"))
+(define colname-long-term-realized-gain (N_ "LT Realized Gain"))
+(define colname-realized-roi (N_ "Realized ROI"))
+(define colname-unrealized-gain (N_ "Unrealized Gain"))
+(define colname-short-term-unrealized-gain (N_ "ST Unrealized Gain"))
+(define colname-long-term-unrealized-gain (N_ "LT Unrealized Gain"))
+(define colname-unrealized-roi (N_ "Unrealized ROI"))
+
+(define label-account-total (N_ "Account Lots Total"))
+(define label-grand-total (N_ "Grand Total"))
+
+;; This function will generate a set of options that GnuCash
+;; will use to display a dialog where the user can select
+;; values for the report's parameters.
+(define (options-generator)
+  (let* ((options (gnc:new-options)))
+
+    ;; This is just a helper function for making options.
+    ;; See libgnucash/app-utils/options.scm for details.
+    (define (add-option new-option)
+      (gnc:register-option options new-option))
+
+    ;; Accounts tab
+    (add-option
+     (gnc:make-account-list-option
+      gnc:pagename-accounts
+      optname-accounts
+      "a"
+      (N_ "Stock Accounts to report on.")
+      ;; default-getter
+      (lambda () (filter gnc:account-is-stock?
+                         (gnc-account-get-descendants-sorted
+                          (gnc-get-current-root-account))))
+      ;; value-validator
+      (lambda (accounts) (list  #t
+                                (filter gnc:account-is-stock? accounts)))
+      #t)) ;; multiple-selection
+
+    (add-option
+      (gnc:make-simple-boolean-option
+        gnc:pagename-accounts
+        optname-zero-shares
+        "b"
+        (N_ "Include accounts that have a zero share balances.")
+        #t))
+
+    (add-option
+      (gnc:make-simple-boolean-option
+        gnc:pagename-accounts
+        optname-include-lotless-accounts
+        "c"
+        (N_ "Include accounts with no lots")
+        #f))
+
+    ;; Chart tab
+    (add-option
+      (gnc:make-simple-boolean-option
+        pagename-chart
+        optname-show-chart
+        "a"
+        (N_ "Include a chart that shows lot gains, grouped by account and gain type")
+        #t))
+
+    (add-option
+      (gnc:make-multichoice-option  
+        pagename-chart
+        optname-chart-type
+        "b"
+        (N_ "What kind of chart to include")
+        'bar-stacked
+        (list (vector 'bar (N_ "Bar Chart"))
+              (vector 'bar-stacked (N_ "Stacked Bar Chart")))))
+
+    (add-option
+      (gnc:make-multichoice-option  
+        pagename-chart
+        optname-chart-location
+        "c"
+        (N_ "Where to place the chart")
+        'top
+        (list (vector 'top (N_ "Top"))
+              (vector 'bottom (N_ "Bottom")))))
+
+    (gnc:options-add-plot-size!
+      options
+      pagename-chart
+      optname-plot-width
+      optname-plot-height
+      "d"
+      (cons 'percent 100.0)
+      (cons 'percent 50.0))
+
+    ;; Columns tab
+    (add-option
+      (gnc:make-simple-boolean-option
+        pagename-columns
+        optname-show-lot-guid-column
+        "a"
+        (N_ "Show the lot GUID table column")
+        #f))
+
+    (add-option
+      (gnc:make-simple-boolean-option
+        pagename-columns
+        optname-show-date-columns
+        "b"
+        (N_ "Show the lot open and close table columns")
+        #t))
+
+    (add-option
+      (gnc:make-simple-boolean-option
+        pagename-columns
+        optname-show-bought-columns
+        "c"
+        (N_ "Show purchase-related table columns")
+        #t))
+
+    (add-option
+      (gnc:make-simple-boolean-option
+        pagename-columns
+        optname-show-sold-columns
+        "d"
+        (N_ "Show sale-related table columns")
+        #t))
+
+    (add-option
+      (gnc:make-simple-boolean-option
+        pagename-columns
+        optname-show-end-columns
+        "e"
+        (N_ "Show end date amount and value table columns")
+        #t))
+
+    (add-option
+      (gnc:make-simple-boolean-option
+        pagename-columns
+        optname-show-realized-gain-columns
+        "f"
+        (N_ "Show realized gain table column(s) for sold shares")
+        #t))
+
+    (add-option
+      (gnc:make-simple-boolean-option
+        pagename-columns
+        optname-show-unrealized-gain-columns
+        "g"
+        (N_ "Show unrealized gain table column(s) for unsold shares")
+        #t))
+
+    (add-option
+      (gnc:make-multichoice-option
+        pagename-columns
+        optname-group-gains-by-age
+        "h"
+        (N_ "Group gains (and sales?) by long-term (LT) and short-term (ST)")
+        'gains-only
+        (list (vector 'no (N_ "No"))
+              (vector 'gains-only (N_ "Gains Only"))
+              (vector 'gains-and-sales (N_ "Gains and Sales")))))
+
+    ;; Note: Different governments may have different rules regarding how long
+    ;; shares must be held to qualify for different tax treatment. So make
+    ;; configurable the boundary between short-term and long-term capital
+    ;; gains.
+    (add-option
+     (gnc:make-number-range-option
+      pagename-columns
+      optname-long-term-years
+      "i"
+      (N_ "Commodities held longer than this many years count as long-term (LT).")
+      1 ;; default-value. For USA federal taxes, shares held longer than 1
+        ;; year are long-term.
+      0 ;; lower-bound
+      10E9 ;; upper-bound
+      0 ;; num-decimals
+      1)) ;; step-size
+
+    ;; Display tab
+    (add-option
+      (gnc:make-simple-boolean-option
+        gnc:pagename-display
+        optname-show-long-account-names
+        "a"
+        (N_ "Show long (instead of short) account names")
+        #t))
+
+    (add-option
+      (gnc:make-simple-boolean-option
+        gnc:pagename-display
+        optname-show-mnemonics
+        "b"
+        (N_ "Show mnemonics with commodity amounts")
+        #t))
+
+    (add-option
+      (gnc:make-simple-boolean-option
+        gnc:pagename-display
+        optname-include-closed-lots
+        "c"
+        (N_ "Include closed lots in addition to open lots")
+        #t))
+
+    (add-option
+      (gnc:make-simple-boolean-option
+        gnc:pagename-display
+        optname-show-blanks-for-zeros
+        "d"
+        (N_ "Show blank text instead of zero values for inner table cells. Does not apply to footer rows.")
+        #t))
+
+    (add-option
+      (gnc:make-simple-boolean-option
+        gnc:pagename-display
+        optname-show-split-rows
+        "e"
+        (N_ "Add a row for each split belonging to a lot, under the lot row.")
+        #f))
+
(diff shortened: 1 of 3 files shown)


mcbridebt:
TODO: The report is not properly respecting the end date option. Fix it. Also, add a start date option. This should allow the report's output to be comparable to capital gains related tax forms (i.e. in USA, form 1099-B and form 1040 Schedule D) for any given year.

christopherlam:
Which guile version are you using? This file does not compile with guile-3.0

mcbridebt:
> Which guile version are you using? This file does not compile with guile-3.0

I don't know. How can I tell? I'm running GnuCash 4.12 on Win10. The report works for me in that environment.

mcbridebt:
I think the Guile issue is fixed now. 

This is all new to me, but I think the GnuCash 4.12 install on my Windows box is running Guile 2.2? Based on the number in the file path: "C:\Program Files (x86)\gnucash\share\guile\site\2.2".

Then I created a Linux VM (Ubuntu 22.04) and went through the instructions (https://wiki.gnucash.org/wiki/Building_On_Ubuntu) to build and install GnuCash from my topic branch. It appears to be using Guile 3.0: "/opt/gnucash/share/guild/site/3.0"

I ran my report in that environment and repro'ed the error, then pushed a fix. Does it work for you now?

christopherlam:
> I think the Guile issue is fixed now.

Mostly. There are compilation errors; the majority seem to be your use of the following form;

```scheme
 (case obj
  ('symbol ...))
```
which [should](https://www.gnu.org/software/guile/manual/html_node/OO-Closure.html#OO-Closure) be
```scheme
 (case obj
  ((symbol) ...))
```

> This is all new to me, but I think the GnuCash 4.12 install on my Windows box is running Guile 2.2? Based on the number in the file path: "C:\Program Files (x86)\gnucash\share\guile\site\2.2".

That's right windows still on 2.2.

> 
> Then I created a Linux VM (Ubuntu 22.04) and went through the instructions (https://wiki.gnucash.org/wiki/Building_On_Ubuntu) to build and install GnuCash from my topic branch. It appears to be using Guile 3.0: "/opt/gnucash/share/guild/site/3.0"
> 
> I ran my report in that environment and repro'ed the error, then pushed a fix. Does it work for you now?

Not quite; see above; there are still other `format` issues and you should add `(use-modules (ice-9 format))`.

Having said that, scheme has modernised a bit in the past 20 years; we'd want to avoid using `set!` as far as possible (!!), we don't _really_ need `gnc-numeric-add/sub/div/mul/zero?` because we can use `+ - * / zero?` directly. Otherwise this code is well written.

I'll paste a screenshot from a test investment book. I am not 100% sure how many assumptions it holds upon the underlying data, and it *must* be documented properly in the guide.

image

mcbridebt:
Thanks! How do I repro those compilation errors? I don't see any in the output when I run "make". Is there some special command I need to run to see those errors?

I do see some warnings about other scm files (like "options.scm:307:67: warning: possible unbound variable ...."), but I don't see a single reference to my new file investment-lots.scm. Which is odd. Shouldn't it be listed here, with all the other reports?

image

Hmm, I wonder if I need to add a reference to the new file in some build file somewhere...

mcbridebt:
Okay, I added it to CMakeLists.txt, and now I do see file compilations warnings. Fixing...

mcbridebt:
I've pushed more fixes, such that the build warnings no longer repro on my Linux box. 

> we'd want to avoid using `set!` as far as possible
I refactored away as many of these as I could figure out how.


> we don't _really_ need `gnc-numeric-add/sub/div/mul/zero?` because we can use `+ - * / zero?` directly
From looking at the code, I thought a purpose of these functions was to control where and how rounding happens. Can `+ - * / zero?` do the same?


> I am not 100% sure how many assumptions it holds upon the underlying data, and it _must_ be documented properly in the guide.

Yeah, the report could definitely use some documentation. I've made a first stab at it, and added it to the first comment of this PR. The section on the Validation warning options hopefully covers the "assumptions it holds upon the underlying data".

Side note: In that section I call out what appears to be a rounding error. It seems that GnuCash is not doing rounding according to my (possibly incorrect?) understanding how rounding is supposed to be done.  I'd appreciate it if you could take a look at let me know what you think. It seems either my report is wrong, or the GnuCash logic that computes gains is wrong.

As for "the guide", where is it and what is the process for updating it? I've seen (and quoted in the above writeup) various wikis and other places that have GnuCash documentation. Not sure if one of them is "the guide" you mentioned.  Is it in the source tree? looked in my local clone, but didn't find it. I see a /doc folder, but that doesn't appear to be it.

christopherlam:
> > we'd want to avoid using `set!` as far as possible
> 
> I refactored away as many of these as I could figure out how.

Thank you, it does help a lot with reading code. Plus guile has some further optimisations if a variable is never _set!_.

> From looking at the code, I thought a purpose of these functions was to control where and how rounding happens. Can `+ - * / zero?` do the same?

Correct, gnc-numeric-* offer rounding and most reports do not need exact rounding until the final display.

> Yeah, the report could definitely use some documentation. I've made a first stab at it, and added it to the first comment of this PR. The section on the Validation warning options hopefully covers the "assumptions it holds upon the underlying data".
> 
> Side note: In that section I call out what appears to be a rounding error. It seems that GnuCash is not doing rounding according to my (possibly incorrect?) understanding how rounding is supposed to be done. I'd appreciate it if you could take a look at let me know what you think. It seems either my report is wrong, or the GnuCash logic that computes gains is wrong.
> 
> As for "the guide", where is it and what is the process for updating it? I've seen (and quoted in the above writeup) various wikis and other places that have GnuCash documentation. Not sure if one of them is "the guide" you mentioned. Is it in the source tree? I looked in my local clone, but didn't find it. I see a /doc folder, but that doesn't appear to be it.

The official guide is https://github.com/Gnucash/gnucash-docs however I wouldn't recommend creating a PR until it is 100% certain that this report is fully thrashed out. Disclaimer: I maintain reports but do _not_ use gnucash at all for my investment tracking, and have no deep understanding of lots capgains rounding. It is very possible that you've found some capgains bugs within the lot editor. Maybe discussion needs to be in bugzilla including some sample datafiles.

My small concern is the report seems tailor made for US Capgains rules, with the legal STCG/LTCG distinction being 1yr; maybe it'll be nice to offer this arbitrary duration as an option.

This report is designed for determining capital gains via lot-tracking, isn't it? How would it work when the user uses LIFO/FIFO/Average-cost to determine capgains?

I'll defer any further comment regarding the math to someone else.

(FWIW you may have seen the "IFRS weighted-average cost basis report" which aims to track basis, and capital gains, using average-cost method).

mcbridebt:
> This report is designed for determining capital gains via lot-tracking, isn't it? How would it work when the user uses LIFO/FIFO/Average-cost to determine capgains?

I believe LIFO/FIFO/Average-cost/other is determined by how splits are assigned to lots. Once done, there's really only one way to compute gains (sale value minus basis). So, the report doesn't need to know or care whether the user set up their lots via LIFO or FIFO or something else.

- FIFO: Current behavior of the [Scrub operation](https://www.gnucash.org/docs/v4/C/gnucash-help/tool-lots.html):

> Scrub Account:
> - Creates a lot for each buy transaction split that is not already linked to a lot...
> - Links buy and sell transaction splits to lots using FIFO method.

- LIFO: A user who wants this can manually add each sales split to the newest open lot, instead of running Scrub to add them to the oldest.

- Average: Manually adding multiple purchase splits to the same lot will accomplish this. The report will 1) Optionally show a warning, and 2) Use the average cost of these purchases to compute the basis (total purchase value divided by total purchase amount. See the `set-bought-average-price` function).
The downside (and hence the optional warning) is that having multiple purchase splits in the same lot may make ambiguous the holding period, possibly breaking grouping by short and long term. The report will attempt this grouping anyways, using the date of the oldest bought split (see `earliest-bought-split-date`). 
Though I wonder if it would be better to use the newest purchase date instead? Since long term gains often have more favorable tax treatment, it's probably worse to wrongly put a sale into the long-term bucket than in the short term. Maybe I should change that. Hmm...

- Other: A user may use any other method to assign splits to lots. Such as picking the lot to sell that has the lowest unrealized gains (or biggest loss) to minimize taxes. I've done this before (and now with this report, I'll have an easier time identifying such lots!). When I sell shares, my brokers allow me to pick which lot to sell from (and I note, I've never seen a broker that puts more than a single purchase transaction in a lot!). Then in GnuCash I manually add the sale split to the chosen lot, rather than using Scrub to do FIFO. Thus when my broker sends me their year-end tax forms showing my realized capital gains, my GnuCash db should match (and I can use this new report to confirm).


> (FWIW you may have seen the "IFRS weighted-average cost basis report" which aims to track basis, and capital gains, using average-cost method).

Thanks, I'll take a look at it.


> Maybe discussion needs to be in bugzilla including some sample datafiles.

This brings up another question I had: Do we have a set of standard sample data files available somewhere? They would be useful for 1) Creating screenshots for the report documentation, and 2) Doing further testing of my report. Thus far, I've only tested against my own db. It is pretty complex (~20 years of data) with many investment lots. But I worry there may be bugs in my report that would only repro against other dbs that do things differently than mine.

mcbridebt:
TODO: 
I was pondering more the implications of lots that have multiple purchase splits, and I think my report may have a bug in how it computes gains when a purchase split happens after a sale split. Such a scenario seems like abusing the concept of lots. But GnuCash allows adding to a single lot multiple purchase and sale splits, in any order. So my report ought to handle such a situation similarly (nonsensical as it may be).

christopherlam:
> TODO: I was pondering more the implications of lots that have multiple purchase splits, and I think my report may have a bug in how it computes gains when a purchase split happens after a sale split. Such a scenario seems like abusing the concept of lots. But GnuCash allows adding to a single lot multiple purchase and sale splits, in any order. So my report ought to handle such a situation similarly (nonsensical as it may be).

I'm not convinced it is necessary to handle lots with multiple purchase splits. See the overview at https://code.gnucash.org/docs/MAINT/lotsoverview.html and written by the original lot coder.

I would recommend wider beta testing in gnucash-devel.

mcbridebt:
In troubleshooting the rounding discrepancy, I found where GnuCash computes lot capital gains: libgnucash/engine/cap-gains.c. Its approach differs from my report. It keeps a running tally of basis value, which it updates for both purchase and sale splits. Whereas mine computed basis by summing up the purchase splits, which works fine... unless a sales split precedes a purchase split. It also changes where rounding happens...hence all the report warnings about rounding errors. So I changed my report to follow the cap-gains.c approach. 

That got rid of all of the rounding warnings (right or wrong, at least the report now rounds consistently with the rest of GnuCash). And if a user does create an (arguably non-sensical) lot with multiple purchase splits, some of which don't precede all of the sales splits, the report will yield the same (possibly bogus) gains. Garbage in, garbage out. But at least the garbage out will be consistent! :-) But the optional warning will encourage users to fix their lots and avoid this situation.

Okay, so now I'm not aware of any more issues to be fixed in the report. 

> I would recommend wider beta testing in gnucash-devel.

Sounds good. How do I set that up? The report is currently parented under the **Experimental** menu, with  `gnc-prefs-is-extra-enabled` set to `#t`, similar to the **Lot Viewer** report. I figured that would allow bleeding edge users (beta testers?) to wring it out before it is deemed stable enough to release to all users. Or is there a different process I should follow?

gjanssens:
> In troubleshooting the rounding discrepancy, I found where GnuCash computes lot capital gains: libgnucash/engine/cap-gains.c. Its approach differs from my report. It keeps a running tally of basis value, which it updates for both purchase and sale splits. Whereas mine computed basis by summing up the purchase splits, which works fine... unless a sales split precedes a purchase split. It also changes where rounding happens...hence all the report warnings about rounding errors. So I changed my report to follow the cap-gains.c approach.
> 
Rather than your calculation "to be more like cap-gains.c" can you actually use the calculation method in cap-gains.c ?

In that case if someone in the future makes modifications to that calculation your report automatically benefits from those changes as well.

jralls:
>Rather than your calculation "to be more like cap-gains.c" can you actually use the calculation method in cap-gains.c ?

That's asking a lot: cap-gains.h isn't wrapped for Guile.

gjanssens:
> > Rather than your calculation "to be more like cap-gains.c" can you actually use the calculation method in cap-gains.c ?
> 
> That's asking a lot: cap-gains.h isn't wrapped for Guile.

I didn't check that. Agreed that would be a lot to ask.

mcbridebt:
> > Rather than your calculation "to be more like cap-gains.c" can you actually use the calculation method in cap-gains.c ?
> 
> That's asking a lot: cap-gains.h isn't wrapped for Guile.

Even if it were, xaccSplitComputeCapGains does not return gains, such that I could reuse it in my report. If I'm reading it right, it mutates the account by inserting the computed "Realized Gain/Loss" split and linking it to the sale split.

mcbridebt:
> Sounds good. How do I set that up? The report is currently parented under the **Experimental** menu, with `gnc-prefs-is-extra-enabled` set to `#t`, similar to the **Lot Viewer** report. I figured that would allow bleeding edge users (beta testers?) to wring it out before it is deemed stable enough to release to all users. Or is there a different process I should follow?

So, what are the next steps to get this PR competed? Is it good to go as-is (**Experimental** and `--extra`)?

christopherlam:
> > Sounds good. How do I set that up? The report is currently parented under the **Experimental** menu, with `gnc-prefs-is-extra-enabled` set to `#t`, similar to the **Lot Viewer** report. I figured that would allow bleeding edge users (beta testers?) to wring it out before it is deemed stable enough to release to all users. Or is there a different process I should follow?
> 
> So, what are the next steps to get this PR competed? Is it good to go as-is (**Experimental** and `--extra`)?

From my POV the commits *could* get squashed together into one. If @jralls has no further concerns I'd think it's ok to merge into `maint` rather than `master` as a hidden experimental report.

jralls:
Yes to squashing. 

AFAICT the only time the experimental/extra dodge has worked is with @christopherlam's stock entry assistant where he's used it to let @CDB-Man test nightlies. The other half of that collaboration is long running discussions on two bug reports and IRC. Without that specific recruitment and coordination there doesn't seem to be anyone routinely running GnuCash --extra looking for features to test so it doesn't happen.

 GnuCash 4.900 release is only a few weeks away and a new report in its proper location will be a lot more visible to the few users who undertake to test beta releases.

mcbridebt:
I've moved it to the Assets & Liabilities menu, and removed the `--extra` prerequisite. Also fixed a bug and added a new column option.

mcbridebt:
Can this PR be completed now? Anything else I need to do?

gnucash/report/reports/standard/investment-lots.scm
```diff
@@ -0,0 +1,2108 @@
...
+            (let ((date-cell
+                      (to-split-cell (qof-print-date trans-date) split)))
+              (add-data-row
+                splits-table
+                currency
+                #f ;; is-bold
+                #t ;; is-odd-row
+                (string-append
+                    ;; indentation
+                    "&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;"
```
christopherlam, line 1286: There's 1 remaining nbsp in html-acct-table.scm and is proving difficult to reomve. We want to avoid any further raw html in code. It should be replaced by a `bool` to indicate whether to use an additional column, or merge 2 columns.

gnucash/report/reports/standard/investment-lots.scm
```diff
@@ -0,0 +1,2108 @@
...
+              (list (format #f
+                      (G_ "Warning: Above lot's balance is negative on ~a. Consider removing the responsible sale split from the lot and then scrubbing.")
+                      (qof-print-date
+                        (split->date first-negative-split))))
+              '())
+
+            ;; Warn for blank (empty or only whitespace) lot titles.
+            (if (and 
+                  warn-if-lot-title-blank
+                  (string=? "" (string-trim (lot->title lot))))
```
christopherlam, line 1411: Use `string-null?`

gnucash/report/reports/standard/investment-lots.scm
```diff
@@ -0,0 +1,2108 @@
...
+                    (> (length lots) 0))))))
+
+      ;; Adds the current account to the chart.
+      (define (add-account-to-chart
+                short-term-realized-gain
+                long-term-realized-gain
+                short-term-unrealized-gain
+                long-term-unrealized-gain)
+        (if show-chart
+          (let ((values
```
christopherlam, line 1894: `values` is a guile keyword. Don't use it as a var name.

gnucash/report/reports/standard/investment-lots.scm
```diff
@@ -0,0 +1,2108 @@
...
+                  include-empty-accounts
+                  (not (gnc-numeric-zero-p (get-account-balance))))))
+             ;; Gets whether the account should be included in the report,
+             ;; based on the number of lots and the option to ignore accounts
+             ;; with no lots.
+             (check-lot-count (lambda()
+                (or include-lotless-accounts
+                  (and 
+                    (not (null? lots))
+                    (> (length lots) 0))))))
```
christopherlam, line 1885: These two conditionals linked with `and` are identical. Plus, `length` is O(N). It's best to use a simple `(pair? lots)`.

christopherlam:
A few comments so far. Is it useful to record the various (time) performance benchmarks? No other report does them. Plus, it adds quite a lot of clutter. Also curious why use `gnc-numeric-add/sub/mul/div` instead of guile's own `+ - * /` which are native... any rounding into (by default 2 decimal places) is done by the style sheet.

gnucash/report/reports/standard/investment-lots.scm
mcbridebt: I replaced the nbsp's with text-indent css style.

mcbridebt:
> Is it useful to record the various (time) performance benchmarks?

It came in handy when writing the report, but I removed it now.

> Also curious why use `gnc-numeric-add/sub/mul/div` instead of guile's own `+ - * /` which are native
I use them to control exactly where and how rounding happens, so that the report's calculations match those in the C code, less the report numbers have rounding discrepancies

christopherlam:
Ok looks good now. Remaining steps: 

- squash all commits together
- documentation in gnucash-docs (ideally) or wiki (a close second best)

Nice to have (can be done in another PR)
- unit tests with sample datafile (see test-ifrs-cost-basis.scm, test-portfolios.scm, )

christopherlam:
P.S. When this is merged we may perform other UX changes in the master branch.

Also @fellen may require string changes to fit i18n requirements; this means some options and report strings may change. This means if you have saved reports, the options would need to be reset. We'll aim to keep you informed.

mcbridebt:
> * squash all commits together

Is that something you do as part of the merge process, or do I need to do something on my end?

mcbridebt:
> * documentation in gnucash-docs (ideally) or wiki (a close second best)

I'm trying to figure out where to put the documentation. I see that there is a "Help Manual" and "Tutorial and Concepts Guide", but I am unclear on how to compare and contrast them. Are they redundant? What is the difference? Both docs have Reports sections, but both are rather sparse.

For the Assets & Liabilities reports, the Manual (Chapter 9.3.2) has fairly detailed writeups for Advanced Portfolio and Asset Barchar, but blank or one-sentence descriptions for other reports. The Guide (Chapter 10.3.2) has little more than a paragraph for each report.

The documentation I have in mind is based off of the first post in this PR. But that seems like overkill for both docs. I kind of wish each report had its own page, rather than just a small subsection of the standard reports page.

But at any rate, I don't think I should duplicate the same writeup in both docs, as that would be redundant. One option is I could put a detailed writeup in the Manual (since it already has the most detail) and a paragraph summary in the Guide. Or is there a better approach?

jralls:
Merged but a rebase changed the hash so github didn't pick it up. Thanks!

mcbridebt:
> Merged but a rebase changed the hash so github didn't pick it up. Thanks!

Thanks! I've raised a PR with the report documentation here: https://github.com/Gnucash/gnucash-docs/pull/294

PR #1474, opened 2022-11-25
Bug 403979 - Balance column shows only low order digits when too narrow

Bob-IT:
I came across this bug and thought it may be of use so had a go at implementing it, did not take much and I was curious to see what it would look like.
Basically if a PRICE_CELL_TYPE is too narrow to display the whole number, the digits are replaced with '*'s

column-width
I think I may of been caught out by this along time ago.
Worth while?

### DIFF
--- a/gnucash/register/register-gnome/gnucash-sheet-private.c
+++ b/gnucash/register/register-gnome/gnucash-sheet-private.c
@@ -514,6 +514,26 @@ draw_cell (GnucashSheet *sheet, SheetBlock *block,
     }
 #endif
 
+    if ((text != NULL) && (*text != '\0') && g_strcmp0 (PRICE_CELL_TYPE_NAME,
+         gnc_table_get_cell_type_name (table, virt_loc)) == 0)
+    {
+        int text_width;
+        int text_border_padding;
+
+        pango_layout_get_pixel_size (layout, &text_width, NULL);
+
+        text_border_padding = gnc_item_edit_get_margin (item_edit, left_right) +
+                              gnc_item_edit_get_padding_border (item_edit, left_right);
+
+        if (text_width + text_border_padding > width)
+        {
+            int pango_width = (width - text_border_padding) * PANGO_SCALE;
+
+            pango_layout_set_width (layout, pango_width); //pango units
+            pango_layout_set_ellipsize (layout, PANGO_ELLIPSIZE_END);
+        }
+    }
+
     /* If this is the currently open transaction and
        there is no text in this cell */
     if ((table->current_cursor_loc.vcell_loc.virt_row ==


gnucash/register/register-gnome/gnucash-sheet-private.c
```diff
@@ -514,6 +514,24 @@ draw_cell (GnucashSheet *sheet, SheetBlock *block,
     }
 #endif
 
+    if ((text != NULL) && (*text != '\0'))
+    {
+        if (g_strcmp0 (PRICE_CELL_TYPE_NAME, gnc_table_get_cell_type_name (table, virt_loc)) == 0)
+        {
```
jralls, line 520: Please combine the `if`s with `&&` and split the `g_strcmp0` args on the comma to reduce the indentation and line width.

jralls:
Sure, seems reasonable. I suppose jsled's suggestion to ellipsize instead is too hard?

Bob-IT:
> Sure, seems reasonable. I suppose jsled's suggestion to ellipsize instead is too hard?

Not really, just needed to remember to deal with pango units, can be done by replacing the inner if with this...

`            int pango_width = (width - text_border_padding) * PANGO_SCALE;`

`            pango_layout_set_width (layout, pango_width); //pango units`
`            pango_layout_set_ellipsize (layout, PANGO_ELLIPSIZE_MIDDLE);`

column-width

Do you prefer this, I do not mind, both convey what was intended ?

jralls:
Yeah, I think the ellipsized looks better. @christopherlam, @derekatkins, @fellen , @gjanssens what do you all think?

christopherlam:
This has never bothered me... a narrow cell frequently cuts a number mid-glyph so it's easy to detect errors. Maybe ellipsised one-sided would be better to preserve the decimal alignment?

```
    0.00
    1.23
   12.34
  123.45
 ...4.56
```

Bob-IT:
Here is the same register page with the ellipsis at the start...
column-width

gjanssens:
I like the idea to mark numbers that can't be displayed fully. I don't have a strong opinion on how to do that exactly.

Ellipsizing removes even more digits from the already not fully displayed number so there's even less information left to interpret the value. Perhaps that's why libreoffice just replaces such numbers with "#####" instead ?

The ellipsis is lighter on the eyes though so that's probably more elegant.

fellen:
I would ellipsize at the other end to have more significant digits.

jralls:
> I would ellipsize at the other end to have more significant digits.

That doesn't really matter because there's no way to tell how significant. But the drag handle is on the right side of the header cell so it's perhaps a little more intuitive that the ellipses are on the right.

fellen:
> it's perhaps a little more intuitive that the ellipses are on the right.

I am not sure, what pango understands, but to avoid RTL issues let's better talk about `at the end` .

christopherlam:
Bike shedding! Libreoffice Calc also does some smart stuff eg. 1,234,567,890 -> 1.03E+09 and maybe it'll be ok to write "1G", "1.2G", "1.23G", "1.234G"  etc. but this may be too hard.

jralls:
> Bike shedding! 

Indeed.

>  1,234,567,890 -> 1.23E+09

That's OK for floats, but as you know those aren't appropriate in financial programs.

> I am not sure, what pango understands, but to avoid RTL issues let's better talk about at the end .

Nope. RTL scripts like Arabic and Hebrew use LTR numbers, which is a good thing because GnuCash doesn't know how to write or parse numbers RTL. Other RTL scripts include number systems that are written RTL but they're like Roman numbers in that they use separate symbols instead of position for higher-order numbers. GnuCash doesn't support those.

christopherlam:
There's https://github.com/cfobel/si-prefix which could be ported to CPP

jralls:
Thanks!

PR #1477, opened 2022-12-04
Bug 798681 - Previously imported investment income transactions may not be filtered.

agwekixj:
Instantiate OFX investment income transactions with the investment account split before the income account.


jralls:
I've done a major refactor of `ofx_proc_transaction` and fixed this along with some other problems.

PR #1478, opened 2022-12-04
Bug 798680 - Not able to match a reverse transaction of a previously matched transaction

agwekixj:
Allow matching a reverse transaction of a previously matched transaction.

### DIFF
--- a/libgnucash/engine/Transaction.c
+++ b/libgnucash/engine/Transaction.c
@@ -2927,6 +2927,9 @@ xaccTransReverse (Transaction *orig)
 
     /* Make sure the reverse transaction is not read-only */
     xaccTransClearReadOnly(trans);
+	
+    /* Make sure the online id is unset in the reverse transaction */
+    qof_instance_set (QOF_INSTANCE(trans), "online-id", "", NULL);
 
     qof_instance_set_dirty(QOF_INSTANCE(trans));
     xaccTransCommitEdit(trans);


libgnucash/engine/Transaction.c
```diff
@@ -2927,6 +2927,9 @@ xaccTransReverse (Transaction *orig)
 
     /* Make sure the reverse transaction is not read-only */
     xaccTransClearReadOnly(trans);
+	
+    /* Make sure the online id is unset in the reverse transaction */
+    qof_instance_set (QOF_INSTANCE(trans), "online-id", "", NULL);
```
christopherlam, line 2932: It's neater to remove the `online-id` slot altogether:
```c
     qof_instance_set (QOF_INSTANCE(trans), "online-id", NULL, NULL);
```

agwekixj: I'm not certain using `NULL` actually removes the `online-id` _slot_.  I believe it just sets the value of key to NULL. Either value should suffice (see import-utilities.c: `gnc_import_trans_has_online_id()`).

jralls: It does, via [`qof_instance_set_kvp`](https://github.com/Gnucash/gnucash/blob/maint/libgnucash/engine/qofinstance.cpp#L1060)->[`KvpFrameImpl::set_path`](https://github.com/Gnucash/gnucash/blob/maint/libgnucash/engine/kvp-frame.cpp#L138)->[`KvpFrame::set_impl`](https://github.com/Gnucash/gnucash/blob/maint/libgnucash/engine/kvp-frame.cpp#L106).

The last unconditionally destroys the slot and recreates it only if the provided value is not `nullptr`.

jralls:
Regardless, it's the wrong place(s), we actually want to unset `"online-id"` in `xaccTransClone` and `xaccSplitCopyKvp`.

I fixed the bug in a different commit.

PR #1487, opened 2022-12-29
If a currency pair is not found, make a detour via EUR (maint branch)

zommuter:
Some pairs, e.g. XAU-CHF, are not directly available. Make a detour via EUR then. This could of course be improved to try other detour currencies etc...

### DIFF
--- a/libgnucash/quotes/gnc-fq-dump.in
+++ b/libgnucash/quotes/gnc-fq-dump.in
@@ -198,6 +198,16 @@ if ($exchange eq "currency") {
             $result = $inv_res;
         }
     }
+    # If quotes aren't directly available, try a detour via EUR
+    # TODO reverse direction as above, detour via USD etc...
+    unless (defined($result)) {
+        my $eur_res_from = $q->currency($from, "EUR");
+        my $eur_res_to = $q->currency($to, "EUR");
+        if (defined($eur_res_from) and defined($eur_res_to)) {
+            $result = $eur_res_from / $eur_res_to;
+        }
+    }
+    # end
     if (defined($result)) {
       printf "1 $from = $result $to\n";
     } else {
--- a/libgnucash/quotes/gnc-fq-helper.in
+++ b/libgnucash/quotes/gnc-fq-helper.in
@@ -358,6 +358,16 @@ while(<>) {
             $price = $inv_price;
         }
     }
+    # If quotes aren't directly available, try a detour via EUR
+    # TODO reverse direction as above, detour via USD etc...
+    unless (defined($price)) {
+        my $eur_price_from = $quoter->currency($from_currency, "EUR");
+        my $eur_price_to = $quoter->currency($to_currency, "EUR");
+        if (defined($eur_price_from) and defined($eur_price_to)) {
+            $price = $eur_price_from / $eur_price_to;
+        }
+    }
+    # end
 
     $quote_data{$from_currency, "success"} = defined($price);
     $quote_data{$from_currency, "symbol"} = $from_currency;


jralls:
Any logic like this belongs in Finance::Quote, not here.

zommuter:
@jralls By that logic 923b01e26952caf44f546b5c895dbb82d6ab9188 shouldn't be here either :wink: Though I basically agree, this is just where I fixed it for myself.

zommuter:
Side note, this is my "solution" for https://money.stackexchange.com/q/154391/5035 - alternatively an option to just use a different base currency without changing the locale might do the trick as well...

zommuter:
> Any logic like this belongs in Finance::Quote, not here.

VS

> F::Q and its various modules can only fetch data available from the various sources. Adding an indirect lookup would be the function of the program/script calling F::Q. -- https://github.com/finance-quote/finance-quote/issues/258#issuecomment-1367584314

bpschuck:
Additionally as posted on https://github.com/finance-quote/finance-quote/issues/258:

The correct solution would be to find a Swiss based source for precious metal quotes and create a new module specifically for that source. To create an indirect calculation for every conceivable currency pair not readily available from AlphaVantage or Yahoo would be trip down the old rabbit hole.

jralls:
@zommuter It turns out this won't work anyway: While GnuCash considers XAU a currency (only because ISO4217 does), [Alphavantage doesn't](https://www.alphavantage.co/physical_currency_list/). That's a csv download, not a webpage.

A bit of Googling found that bloomberg quotes XAU-CHF with the symbol XAUCHF:CUR 
```
gnc-fq-dump bloomberg XAUCHF:CUR
Finance::Quote fields Gnucash uses:
    symbol: XAUCHF:CUR           <=== required
      date: 12/29/2022           <=== recommended
  currency: CHF                  <=== required
      last: 1675.4900            <=\
       nav:                      <=== one of these
     price:                      <=/
  timezone:                      <=== optional
```

So you need to set up a non-currency commodity for that and set the quote source to bloomberg.

bpschuck:
@jralls  - AlphaVantage does have some precious metal quotes available through the `CURRENCY_EXCHANGE_RATE` API.

```
% curl "https://www.alphavantage.co/query?function=CURRENCY_EXCHANGE_RATE&from_currency=XAU&to_currency=USD&apikey=$ALPHAVANTAGE_API_KEY"
{
    "Realtime Currency Exchange Rate": {
        "1. From_Currency Code": "XAU",
        "2. From_Currency Name": null,
        "3. To_Currency Code": "USD",
        "4. To_Currency Name": "United States Dollar",
        "5. Exchange Rate": "1814.69950000",
        "6. Last Refreshed": "2022-12-29 21:43:01",
        "7. Time Zone": "UTC",
        "8. Bid Price": "1814.69950000",
        "9. Ask Price": "1814.69950000"
    }
}  
```

BTW - Nice find on using Bloomberg.

zommuter:
@jralls That's curious, thanks! Yeah I guess setting up a separate commodity is probably better. But now I'm at a loss, since your statement contradicts `gnc-fq-dump currency XAU EUR` actually working 🤔

... Ah @bpschuck already found the answer to that question while I was still typing, thanks 😅

But anyway, the way gnucash currently handles this could be improved one way or another...

bpschuck:
@jralls was likely not aware of AlphaVantage's "_undocumented_" feature. According to AV's API docs, the precious metals are not listed in the physical or crypto currencies that the `CURRENCY_EXCHANGE_RATE` API will convert.

jralls:
I looked at the Alphavantage currency list because it *wasn't* working for me:
```
$ gnc-fq-dump currency XAU USD
1 XAU = <unknown> USD
$ gnc-fq-dump currency XAU EUR
1 XAU = <unknown> EUR
$ gnc-fq-dump currency USD EUR
1 USD = 0.9378 EUR
```
That's because I hadn't yet upgraded to 1.54; it works correctly now.

bpschuck:
It worked in 1.52 and earlier (I was even using a self hacked version of GnuCash that treated crypto just like physical currency for a while). Then when AV changed the `CURRENCY_EXCHANGE_RATE` API to a premium, the exchange rate was not available in `FX_DAILY` which was what F::Q v1.53 was using. I noticed AV flipped back and so I reverted that merge in v1.54.

PR #1488, opened 2022-12-29
If a currency pair is not found, make a detour via EUR (master branch)

zommuter:
Some pairs, e.g. XAU-CHF, are not directly available. Make a detour via EUR then. This could of course be improved to try other detour currencies etc...

### DIFF
--- a/libgnucash/quotes/finance-quote-wrapper.in
+++ b/libgnucash/quotes/finance-quote-wrapper.in
@@ -169,6 +169,15 @@ sub parse_currencies {
                 $inverted = 1;
             }
         }
+        # If quotes aren't directly available, try a detour via EUR
+        # TODO reverse direction as above, detour via USD etc...
+        unless (defined($price)) {
+            my $eur_price_from = $quoter->currency($from_currency, "EUR");
+            my $eur_price_to = $quoter->currency($to_currency, "EUR");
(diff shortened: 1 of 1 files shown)


jralls:
You don't want to do any numeric computation in perl, it will [force the numbers to floats](https://perldoc.perl.org/perlnumber) with a severe loss of precision. [We've been down that road before](https://bugs.gnucash.org/show_bug.cgi?id=797233).

Fortunately GnuCash already knows how to compute single-level indirect prices so if you just add the XAU-EUR and EUR-CHF prices to the pricedb GnuCash can do the rest.

Even better would be to make a way to tell GnuCash that it wants to get XAU-EUR instead of XAU-CHF. There's a mechanism for that for non-currency commodities, you make the non-currency account a sub account of one denominated in the right currency. I guess the simplest (and still a little bit hacky) would be a "use parent account currency for transaction currency" boolean KVP slot controlled by a corresponding tick box on the account edit dialog.

jralls:
It's been 6 weeks with no response so I guess you're not able to do this yourself. Please open an [enhancement request](https://wiki.gnucash.org/wiki/Enhancement_Requests) in the [bug tracker](https://bugs.gnucash.org) and maybe someone will implement it someday.

PR #1493, opened 2023-01-08
[balsheet-pnl] Generate a separate exchange-fn for each date. Cache its result.

christopherlam:
Generate a separate exchange-fn for each date. Cache its result.

### DIFF
--- a/gnucash/report/reports/standard/balsheet-pnl.scm
+++ b/gnucash/report/reports/standard/balsheet-pnl.scm
@@ -47,13 +47,6 @@
   (split-balance-with-closing col-datum-get-split-balance-with-closing)
   (split-value-balance col-datum-get-split-value-balance))
 
-(define FOOTER-TEXT
-  (gnc:make-html-text
-   (G_ "WARNING: Foreign currency conversions, and unrealized gains
-calculations are not confirmed correct. This report may be modified
-without notice. Bug reports are very welcome at
-https://bugs.gnucash.org/")))
-
 ;; define all option's names and help text so that they are properly
 
 (define optname-startdate (N_ "Start Date"))
@@ -1285,9 +1278,6 @@ also show overall period profit & loss."))
       (gnc:html-document-add-object!
        doc multicol-table))
 
-    (gnc:html-document-add-object!
-     doc FOOTER-TEXT)
-
     (gnc:report-finished)
     ;; (gnc:html-document-set-style-text!
     ;;  doc " table, td{ border-width: 1px; border-style:solid; border-color: lightgray; border-collapse: collapse}")
@@ -1300,7 +1290,7 @@ also show overall period profit & loss."))
  'version 1
  'name balsheet-reportname
  'report-guid "065d5d5a77ba11e8b31e83ada73c5eea"
- 'menu-path (list gnc:menuname-experimental)
+ 'menu-path (list gnc:menuname-asset-liability)
  'options-generator (lambda () (multicol-report-options-generator 'balsheet))
  'renderer (lambda (rpt) (multicol-report-renderer rpt 'balsheet)))
 
@@ -1308,7 +1298,7 @@ also show overall period profit & loss."))
  'version 1
  'name pnl-reportname
  'report-guid "0e94fd0277ba11e8825d43e27232c9d4"
- 'menu-path (list gnc:menuname-experimental)
+ 'menu-path (list gnc:menuname-income-expense)
  'options-generator (lambda () (multicol-report-options-generator 'pnl))
  'renderer (lambda (rpt) (multicol-report-renderer rpt 'pnl)))
 
--- a/gnucash/report/reports/standard/balsheet-pnl.scm
+++ b/gnucash/report/reports/standard/balsheet-pnl.scm
@@ -769,11 +769,17 @@ also show overall period profit & loss."))
               (cons acc (map col-datum-get-split-balance-with-closing cols-data))))
            accounts-cols-data))
 
-         (exchange-fn (and common-currency
-                           (gnc:case-exchange-time-fn
-                            price-source common-currency
-                            (map xaccAccountGetCommodity accounts) enddate
-                            #f #f)))
+         ;; generate an exchange-fn for date, and cache its result.
+         (get-date-exchange-fn
+          (let ((h (make-hash-table)))
+            (lambda (date)
+              (or (hashv-ref h date)
+                  (let ((exchangefn (gnc:case-exchange-time-fn
+                                     price-source common-currency
+                                     (map xaccAccountGetCommodity accounts)
+                                     date #f #f)))
+                    (hashv-set! h date exchangefn)
+                    exchangefn)))))
 
          ;; from col-idx, find effective date to retrieve pricedb
          ;; entry or to limit transactions to calculate average-cost
@@ -799,9 +805,9 @@ also show overall period profit & loss."))
                        (gnc:gnc-monetary-commodity monetary)
                        common-currency))
                  (has-price? (gnc:gnc-monetary-commodity monetary))
-                 (exchange-fn
-                  monetary common-currency
-                  (col-idx->price-date col-idx)))))
+                 (let* ((col-date (col-idx->price-date col-idx))
+                        (exchange-fn (get-date-exchange-fn col-date)))
+                   (exchange-fn monetary common-currency col-date)))))
 
          ;; the following function generates an gnc:html-text object
          ;; to dump exchange rate for a particular column. From the


CDB-Man:
Inconsistency between `closest to report date` and `last up through report date` methods.  Legacy BS and multicol BS are consistent with each other, however the methods are not producing the expected result.

I have a USDCAD FX rate in pricedb for both `2019-12-31 = 1.2988` and `2020-01-01 = 1.2987`.  Using a report date of `2019-12-31`, for `last up through report date` the rate of `1.2988` is produced as expected, but for `closest to report date` the rate of `1.2987` was selected instead.

image

CDB-Man:
These methods have been tested
1. Most recent
2. Closest to report date
3. Last up through report date

The results are consistent between legacy vs multicol BS, and the results are also as expected.

For the other methods:
1. **Weighted average of all transactions in the past:** I do not understand the purpose of having this method, or the calculation behind it.  Also, given that my book has tens of thousands of transactions since 2015, this takes over 4 minutes to calculate on legacy BS (and I assume even longer on multicol).  I do not understand its inner workings, nor see the necessity of having this method available.
2. **Average cost of purchases weighted by volume**: same comment as above.

jralls:
@CDB-Man 
>Does the FX calculator have a built in preference / bias for F::Q quotes over user entered quotes? If yes, this should only be applied if there are 2 entries on the same date.

No. A couple of things have changed between 2020 and now: Before GC4.12 F::Q quotes were recorded at 16:00 local time, and manually entered quotes were recorded at 00:00 local. That caused the indicated day to change with the timezone. In 4.12 and later all quotes are recorded  at 10:59 UTC, a time that is date stable in almost all time zones. For all releases and git builds until two days ago the report date time was set at noon local, resulting in the "latest through" price to be the one from the day before in time zones east of UTC+2. I just changed that to midnight local to fix that.

There is a precedence for price sources, but it's applied when the price is created. The PriceDB stores only one price per day. The precedence is (low) price created by entering a transaction, (middle) Finance::Quote, (high) price created by the price editor. A new price is recorded if it is of equal or higher precedence than the one that's already there, otherwise it's silently discarded.

CDB-Man:
> There is a precedence for price sources, but it's applied when the price is created. The PriceDB stores only one price per day. The precedence is (low) price created by entering a transaction, (middle) Finance::Quote, (high) price created by the price editor. A new price is recorded if it is of equal or higher precedence than the one that's already there, otherwise it's silently discarded.

This makes a lot of sense to me.

> No. A couple of things have changed between 2020 and now: Before GC4.12 F::Q quotes were recorded at 16:00 local time, and manually entered quotes were recorded at 00:00 local. That caused the indicated day to change with the timezone. In 4.12 and later all quotes are recorded at 10:59 UTC, a time that is date stable in almost all time zones. For all releases and git builds until two days ago the report date time was set at noon local, resulting in the "latest through" price to be the one from the day before in time zones east of UTC+2. I just changed that to midnight local to fix that.

Hmm, I am following, but not sure if I follow all the way through.  In my case with the screenshot above, I have the following data in pricedb (my local time in UTC-5 EST):
1. 2019-12-31 00:00 EST 1.2988
2. 2020-01-01 16:00 EST 1.2987

In the previous report logic, `the report date time was set at noon local`, meaning my report is `2019-12-31 12:00 EST`. The "latest through" price is 1.2988. Should the "nearest price" not also be 1.2988?  `2019-12-31 00:00 EST` is 12 hours away from `2019-12-31 12:00 EST`, whereas `2020-01-01 16:00 EST` is 24+4 hours away from `2019-12-31 12:00 EST`.

For clarity, after applying `I just changed that to midnight local to fix that`, what will the displayed price be in my example here?

For context, I am running version `gnucash-4.13-2023-01-07-git-4.13-14-gf7324d16b4+.setup.exe`

christopherlam:
Thanks @CDB-Man I think this particular fix is correct then.

I can't really comment on weighted-average. This is >20years old.

With regards to price-source, there's https://bugs.gnucash.org/show_bug.cgi?id=798622 and f7324d16b4 very recently (i.e. 7-jan-23) merged in.

jralls:
@CDB-Man,
> Hmm, I am following, but not sure if I follow all the way through. In my case with the screenshot above, I have the following data in pricedb (my local time in UTC-5 EST):

>    2019-12-31 00:00 EST 1.2988
>    2020-01-01 16:00 EST 1.2987

> In the previous report logic, the report date time was set at noon local, meaning my report is 2019-12-31 12:00 EST. The "latest through" price is 1.2988. Should the "nearest price" not also be 1.2988? 2019-12-31 00:00 EST is 12 hours away from 2019-12-31 12:00 EST, whereas 2020-01-01 16:00 EST is 24+4 hours away from 2019-12-31 12:00 EST.

> For clarity, after applying I just changed that to midnight local to fix that, what will the displayed price be in my example here?

> For context, I am running version gnucash-4.13-2023-01-07-git-4.13-14-gf7324d16b4+.setup.exehttps://github.com

Which should have the ['midnight change'](/Gnucash/gnucash/commit/f7324d16b47ec180a0a79ebb5a50c169236f6074).
I shouldn't have said midnight, though, because what I changed it to is 23:59:59 local, 1 second before midnight the next day. I did that so that "up through" will include any price on the day regardless of time zone or what time it is. That has the downside that nearest still depends on timezone because whether 10:59 UTC on the date or on the next day is closer depends on what 23:59:59 is in UTC. That's not what's at play in the case you just posed because all of the times are local and 16:00 on the 1st is closer to 23:59:59 on the 31st than 00:00 on the 31st.

jralls:
@CDB-Man,
> Hmm, I am following, but not sure if I follow all the way through. In my case with the screenshot above, I have the following data in pricedb (my local time in UTC-5 EST):

>    2019-12-31 00:00 EST 1.2988
>    2020-01-01 16:00 EST 1.2987

> In the previous report logic, the report date time was set at noon local, meaning my report is 2019-12-31 12:00 EST. The "latest through" price is 1.2988. Should the "nearest price" not also be 1.2988? 2019-12-31 00:00 EST is 12 hours away from 2019-12-31 12:00 EST, whereas 2020-01-01 16:00 EST is 24+4 hours away from 2019-12-31 12:00 EST.

> For clarity, after applying I just changed that to midnight local to fix that, what will the displayed price be in my example here?

> For context, I am running version gnucash-4.13-2023-01-07-git-4.13-14-gf7324d16b4+.setup.exehttps://github.com

Which should have the ['midnight change'](/Gnucash/gnucash/commit/f7324d16b47ec180a0a79ebb5a50c169236f6074).
I shouldn't have said midnight, though, because what I changed it to is 23:59:59 local, 1 second before midnight the next day. I did that so that "up through" will include any price on the day regardless of time zone or what time it is. That has the downside that nearest still depends on timezone because whether 10:59 UTC on the date or on the next day is closer depends on what 23:59:59 is in UTC. That's not what's at play in the case you just posed because all of the times are local and 16:00 on the 1st is closer to 23:59:59 on the 31st than 00:00 on the 31st.

I guess that the "nearest to" case should go back to noon, I didn't really think that case through on Friday.

CDB-Man:
> I shouldn't have said midnight, though, because what I changed it to is 23:59:59 local, 1 second before midnight the next day.

Explaining it to myself so that I follow...

Referencing my datapoints:
1. 2019-12-31 00:00 EST 1.2988
2. 2020-01-01 16:00 EST 1.2987

Using a report time of 2019-12-31 **23:59** EST, then 1.2988 is 23:59 away, whereas 1.2987 is 16:01 away, hence 1.2987 is the nearest time.
By changing the report time to 2019-12-31 **12:00** EST, 1.2988 is now the latest time.

This all makes sense to me now.

====
@christopherlam re:
> I can't really comment on weighted-average. This is >20years old.

I assume you similarly also have no comment on the `Average cost of purchases weighted by volume` method as well?

jralls:
> Using a report time of 2019-12-31 23:59 EST, then 1.2988 is 23:59 away, whereas 1.2987 is 16:01 away, hence 1.2987 is the nearest time.
> By changing the report time to 2019-12-31 12:00 EST, 1.2988 is now the latest time.

Except that that last sentence should be "If the report time was put back to 12:00 EST, 1.2988 would be the nearest in time quote".

Expanding on that, the report time for "latest before" needs to stay at 23:59 to ensure that those old 16:00 quotes and new 10:59 UTC ones still get picked up on the report day regardless of TZ. That means different report times depending on the price source, which could be ugly.

Another way to get there might be to reinterpret the old quote times as 10:59 UTC regardless of what's actually recorded and to use the same 10:59 UTC for the report time as well.

CDB-Man:
> Another way to get there might be to reinterpret the old quote times as 10:59 UTC regardless of what's actually recorded and to use the same 10:59 UTC for the report time as well.

This seems like a reasonable approach.  The topic was recently discussed again here: https://code.gnucash.org/logs/2023/04/08.html#T19:55:33


```
19:55:33 <CDB-Man> chris: multicolumn income statement is showing the FX I have for 2020-12-30 instead of 2020-12-31. the legacy IS report shows the 31st rate correctly. jralls does this have something to do with the timezone change you mentioned a month ago?
19:57:13 <CDB-Man> weird that its only doing it for my 2020 year end values, not 2021. the 30th and 31st prices for 2020 and 2021 are both FQ downloaded rates
19:58:33 *** Kaligula has quit IRC
20:01:41 <chris> CDB-Man: I'd bet it's TZ yes
20:02:38 <chris> if you were to open the datafile in XML, look for a <price> near 2020-12-30 cf 2021-12-31 and report the TZ info
20:02:59 <chris> <price:time>
20:03:22 <CDB-Man> how can I unpack the .gnucash file?
20:23:34 <chris> you could copy to .zip !
20:32:48 <CDB-Man> like rename the extension to .zip?
20:36:37 <chris> yes
20:38:45 <chris> assuming you're using xml rather than sql
20:40:34 <CDB-Man> well a straight rename didnt work
20:40:59 <CDB-Man> i vaguely remember someone saying something about compressed save files
20:42:39 <chris> then try Edit > Preferences > General > compress [false]
20:42:58 <chris> ok it's likely .gz rather than .zip
20:43:12 <CDB-Man> no compression did not work, will try gz
20:43:30 <CDB-Man> .gz did not work
20:43:33 <CDB-Man> .tar.gz ?
20:43:49 <CDB-Man> nope
20:44:47 <chris> save as > xml
20:45:43 <CDB-Man> looks like uncompressed i can open in notepad++
20:48:09 <CDB-Man> <ts:date>2020-12-31 22:09:54 +0000</ts:date>
20:48:14 <CDB-Man> <ts:date>2020-12-30 23:46:32 +0000</ts:date>
20:52:30 <chris> yeah not sure what time the exchange converter is picking up
20:52:56 <CDB-Man> actually, i think i quoted the wrong stamp
20:53:27 <CDB-Man> nope, that's the right one
20:54:10 <CDB-Man> jralls had explained it at one point. if there's something i could change it to, or perhaps even just changing the Dec 30th price from an FQ price to a manual price
21:10:01 <chris> the dilemma is whether we'd consider the canonical datafile to be incorrect; and whether the report should adapt to it, or the datafile should be scrubbed
21:10:20 <chris> IIRC the F::Q was fixed fairly recently
21:10:27 <chris> to store more reliable datetimes
21:10:27 *** palerider has left #gnucash
21:12:54 <CDB-Man> [2023.04.08 21:10:20] <chris> IIRC the F::Q was fixed fairly recently <-- yes, jralls and i had talked about it in this channel.... back in January or so
```

PR #1501, opened 2023-01-13
Unique String Values different to Section Name

flywire:
* Check version
* Consider renaming `Hello, World!` section, perhaps to `Hello!`


jralls:
Why?

jralls:
Never mind, I saw the email. Dumb idea.

flywire:
@jralls I'd like to take this forward but it is failing the tests (and generating errors) despite only changing four lines in hello-world.scm after a Sync-fork. The only thing I can think of is a new PR uploading file as a single commit losing commit history. What do you recommend?

@christopherlam would you mind reviewing the latest [flywire:Hello:options-example.scm](https://github.com/flywire/gnucash/blob/Hello/gnucash/report/reports/example/options-example.scm) for consistency with current convention?

jralls:
@flywire, the reason that it's failing tests is because the tests check the option values that you changed. You have to change the tests to match the changes you made in the code.

BTW, you should be working on master, not maint, because the options system is rewritten with some API changes that would make this PR difficult to merge up.

gjanssens:
Following the discussion on [this gnucash-devel thread](https://lists.gnucash.org/pipermail/gnucash-devel/2023-January/046498.html) I'm changing the status of this PR from 'closed' to 'needs work'.

@flywire when ready please push your extra changes on this report to your github repo to update this PR.

gjanssens:
For some reason my browser didn't show your recent commits until right after I reopened the PR. I'll try reviewing soon.

gjanssens:
> > I'm changing the status of this PR from 'closed' to 'needs work'.
> 
> Errors for failed checks are likely too excessive for this repo and I don't know how to fix them.
> 
Well looking at the failed checks, it looks like other parts of the build system still depend on the hello-world.scm being named exactly that. In particular the file once was relocated and as a convenience to users we set a deprecation warning if they tried to access the old location in their own guile code. This warning has already been removed on the master branch, so it will go away after you have rebased on master.

> The report works but I'd be pleased if someone can check and fix these issues for consistency with current conventions:
> 
>     1. Suspect renderer is using old helper functions
> 
Possibly, I haven't searched for those and you won't be able to fix those unless you rebased on master. The newer api calls don't exist yet on maint.

>     2. Suspect older `gnc:html-markup` `format` functions being used
> 
I'm not sure those have been changed. But again, there's little point in checking for those while you're looking at maint.

>     3. Suspect `example/welcome-to-gnucash.scm` adds no value so should be deleted
> 
I agree, though make sure that happens in a separate commit (and still does after you have rebased your work on master)

>     4. Should be pushed to `Master` branch which is not in my fork
> 
It should be in your repository, but you'll have to learn to use git (or resort to a gui application such as gitkraken) in order to switch to a different branch.

Our wiki has [a introduction on using git](https://wiki.gnucash.org/wiki/An_Introduction_to_Git) which may get you started.

> 
> I can fix docs and wiki.

:+1: 


I'll note that switching to the master branch in itself is probably not too difficult to learn. Rebasing your current work on this branch is probably too advanced. I can help with that once you have figured out how to switch branches.

That aside, perhaps I'm stating the obvious but it may help to know that the maint branch in our sources is the branch from which stable gnucash versions are releases (the most recent one being 4.13), while the master branch is for future development, eventually leading up to the next stable series (soon to be 5.0). We generate a new Windows installer every night for each of these two source branches which you can download from https://code.gnucash.org/builds/win32/

For your testing you can download and install a recent master nightly build. Note this will overwrite your currently installed gnucash edition (but not your data). So if you need to switch regularly between a nightly version of gnucash and your regular stable version, you have to reinstall all the time. This may become cumbersome. If you have a separate Windows PC or VM you could save a lot of time installing the master nightly there to test your report modifications.

jralls:
> maint branch in our sources is the branch from which stable gnucash versions are releases (the most recent one being 4.13), while the master branch is for future development, eventually leading up to the next stable series (soon to be 5.0)

Not this quarter. The maint branch has only one terminal release left, 4.14, and the master branch is the beta testing branch from which the 4.90x series are released leading to the *imminent* release of [GnuCash 5.0 in March](https://wiki.gnucash.org/wiki/Release_Schedule#Goals_for_5.0).

@gjanssens since you've decided to humor @flywire's bikeshedding and reopen this (and probably rewrite it for them since they don't know how) I've changed the base to master to ensure that it uses the new options API. There's no point in merging something like this into maint for the final release.

flywire:
I will resubmit as https://github.com/flywire/gnucash/blob/hello2/gnucash/report/reports/example/hello-world.scm based on `master` when the original repo passes tests. The report has not been renamed but API updates are included.

PR #1507, opened 2023-01-22
Fix some bad code formatting

richardcohen:


jralls:
Thanks. Maint can benefit from this too so I merged it there and then merged that up to master.

PR #1508, opened 2023-01-22
Avoid touching the global errno

richardcohen:


jralls:
Thanks. I cherry-picked it onto maint and merged it up so that both branches get the benefit.

PR #1510, opened 2023-01-23
Use internal extern "C" { ... } for C++

richardcohen:
- removes warnings compiling swig engine ...
[ 10%] Generating swig-engine.cpp
.../libgnucash/engine/engine-helpers.h:31: Warning 313: Unrecognized extern type "C++". .../libgnucash/engine/gnc-date.h:83: Warning 313: Unrecognized extern type "C++". .../libgnucash/engine/qofquery.h:90: Warning 302: Identifier 'QofQuery' redefined (ignored), .../libgnucash/engine/gnc-option.hpp:55: Warning 302: previous definition of 'QofQuery'. .../libgnucash/engine/gnc-commodity.h:56: Warning 313: Unrecognized extern type "C++". .../libgnucash/engine/gncBusiness.h:40: Warning 313: Unrecognized extern type "C++". .../libgnucash/engine/gncEntry.h:37: Warning 313: Unrecognized extern type "C++".

I used G_BEGIN/END_DECLS unless it failed to compile


jralls (changes requested):
That's a lot of work, thanks for making the effort. We're trying to reduce our dependencies on GLib, particularly in C++ code, so I'd prefer not to use G_BEGIN/END_DECLS macros. One alternative would be to recreate them with GNC_BEGIN/END_DECLS, but that would require having a common header that gets included everywhere and we don't have one of those and it seems silly to introduce one just for a single macro pair. The other alternative is to replace the macros with their expansion, i.e. `G_BEGIN_DECLS` with `#ifdef __cplusplus…`. Please do the latter.

richardcohen:
The change was actually quicker to do than I expected :)

jralls:
Thanks again!

PR #1512, opened 2023-01-24
Bug 798732: Type of balance carry forward accounts for german skr49 account template fixed

rhabacker:
With this merge request [bug 798732](https://bugs.gnucash.org/show_bug.cgi?id=798732) will be fixed.


jralls:
@fellen Do you agree?

rhabacker:
I have added printouts of reports that are dependent on this change to [bug 798732](https://bugs.gnucash.org/show_bug.cgi?id=798732). Perhaps this will help in the assessment. 

If there is no clear pro or con, there is the possibility to add the changed template as an additional skr49 template. In this case, please inform me, so that I can update this pr.

jralls:
@fellen, silence is assent. ;-)
@rhabacker, Thanks!

fellen:
I have been busy in RL.

I am not sure about the "Carry forward" accounts, but they are common to all SKR templates and should behave similar.

Line 43-45 of the file shows the contriburors/maintainer. They should have been contacted.
Ideally this  should have been discussed on the german GnuCash mailing list. There are the maintainers of SKR49 and some other DATEV experts—perhaps with better knowledge of the  right type.

PR #1513, opened 2023-01-24
Cleanup

richardcohen:


jralls:
Thanks. Merged to master because I wasn't paying attention.

richardcohen:
Thanks. It would be good to have at least the my_strtok() change in maint, because clang 15.0.2 gives a warning:

gnucash/import-export/log-replay/gnc-log-replay.c:112:15: warning: a function definition without a prototype is deprecated in all versions of C and is not supported in C2x [-Wdeprecated-non-prototype]
static char * my_strtok (s, delim)
              ^
1 warning generated.

jralls:
You know that the next release from maint will be the last and will be concurrent with GnuCash 5.0 from master, right?

richardcohen:
Yes, I've seen the plan in the wiki. Thanks. It's probably not important.

PR #1516, opened 2023-01-25
Fix some unused variables

richardcohen:


christopherlam:
IMHO this needs to be scaled down a notch. For exaple: the `static QofLogModule log_module = GNC_MOD_GUI;` is required when debugging code via `PWARN` statements.

richardcohen:
I don't think it is a good idea to leave code around just in case a developer wants to use it.
After all, a developer adding `PWARN` can also add the `log_module = ...` line.
`G_GNUC_UNUSED` is used elsewhere in the code. I could add that?

I really want to end up with the code passing -Wunused - it is very helpful for finding memory leaks.

jralls:
@christopherlam didn't explain that very well. Most modules have some number of logging statements and many of them aren't obvious: `g_return_if_fail` logs an error and requires QofLogModule and G_LOG_DOMAIN to be set.

> G_GNUC_UNUSED is used elsewhere in the code. I could add that?

Better to spell it out: `__attribute__((unused))` or in C++ files `[[unused]]`.

You should do that for discarded return values as well; those are usually there because the compiler complained that the rv was ignored.

>  really want to end up with the code passing -Wunused - it is very helpful for finding memory leaks.

Not really. It misses nearly all and has too many false positives. There are much better tools.

richardcohen:
>`g_return_if_fail` logs an error and requires QofLogModule and G_LOG_DOMAIN to be set.
g_return_if_fail is a glib macro - it only requires LOG_DOMAIN
The log_module lines are definitely not used (successful compile)
 
> `__attribute__((unused))` or in C++ files `[[unused]]`.
`[[unused]]` is not a standard C++ attribute - I think it is almost always possible to avoid unused variables in C++.
[[maybe_unused]] is available, but it's better to remove the variable entirely, or actually use it.

> You should do that for discarded return values as well
Yes, but I don't think there are any Wunused-result warnings in gnucash at the moment.

> ... has too many false positives.
I haven't found any false positives. Can you give me an example?

See the draft pull request  #1519 that shows it building successfully with -Wunused

jralls:
> Yes, but I don't think there are any Wunused-result warnings in gnucash at the moment.

There would be if we merged this PR. I haven't yet looked at your new one.

> [[maybe_unused]] is available, but it's better to remove the variable entirely, or actually use it.

Right about `maybe_unused`. I'll accept that it's better to use the variable, but not that it's always better to remove it.

> I haven't found any false positives. Can you give me an example?

```
int foo;
```
If `foo` isn't used it will raise the warning. It's not a memory leak. Most of the changes in your PR are like this. `-Wunused-but-set-variable` would get you closer, but `int foo = 42;` will trip that and it isn't a leak either. One has to dig deeper in the rv case, especially if it uses C++ `auto`: `auto foo = bar();`. That might be a leak if `bar` returns a naked allocated pointer, but most of the time it's not. But even `Transaction *foo = bar();` is probably not a leak because nearly all engine objects are allocated once during the session and copies of that one ptr are freely passed around.

But the main reason `-Wunused` is useless for leak detection is the false negatives e.g.
```
char *foo = g_strdup_printf("Some message with %d and %s", bar, baz);
gtk_label_set_text(foo);
```
with no `g_free(foo)`. Another version of the same is `gtk_label_set_text(g_strdup_printf("Some message with %d and %s", bar, baz);`.

BTW, Github requires an empty line after a quoted one to end the quote. You can use the preview tab at the top left of the edit box to make sure your markdown does what you want before clicking the comment button.

richardcohen:
> > Yes, but I don't think there are any Wunused-result warnings in gnucash at the moment.
> 
> There would be if we merged this PR. I haven't yet looked at your new one.

For example?

> > I haven't found any false positives. Can you give me an example?
> 
> ```
> int foo;
> ```
> 
> If `foo` isn't used it will raise the warning. 

That is not a false positive -- it *is* unused, and should be removed.

> But the main reason `-Wunused` is useless for leak detection is the false negatives e.g.
> 
> ```
> char *foo = g_strdup_printf("Some message with %d and %s", bar, baz);
> gtk_label_set_text(foo);
> ```
> 
> with no `g_free(foo)`. Another version of the same is `gtk_label_set_text(g_strdup_printf("Some message with %d and %s", bar, baz));`.
> 
I never claimed that it would find all memory leaks. As you say, it will find the most basic.
It is better to clear out the existing errors, so that new ones are more obvious, and fail the build

jralls:
> That is not a false positive -- it is unused, and should be removed.

If you're looking for leaks it's a false positive: It's not a leak.

>  never claimed that it would find all memory leaks. As you say, it will find the most basic -- and it found 3.

I see only [one leak](https://github.com/Gnucash/gnucash/pull/1516/files#diff-16b37bbf0c88abc835b69568ddeb668cee7c41de4d752d0c99d0131448f2768e), and it's in a unit test where it doesn't matter much.

About #1519, it looks to me that you intend it to be only the CMakeLists.txt change and to depend on this. Is that right?

richardcohen:
> I see only [one leak](https://github.com/Gnucash/gnucash/pull/1516/files#diff-16b37bbf0c88abc835b69568ddeb668cee7c41de4d752d0c99d0131448f2768e), and it's in a unit test where it doesn't matter much. 

Yes, I separated out the interesting ones into #1517 
This patch is about unused variables, not memory leaks

> About #1519, it looks to me that you intend it to be only the CMakeLists.txt change and to depend on this. Is that right?

The other commits in #1519 remove other unused things, which are needed to get -Wunused to pass on gcc and clang.
#1519 could be separated into adding -Wunused-variable -Wunused-function, etc. to CMakeLists.txt, but I'm not sure it is worth it.
Yes, #1519 depends on this change.

gnucash/html/gnc-html-webkit2.c
```diff
@@ -52,12 +52,8 @@
 #include "gnc-prefs.h"
 #include "gnc-gui-query.h"
 #include "gnc-engine.h"
-#include "gnc-html.h"
 #include "gnc-html-webkit.h"
 #include "gnc-html-history.h"
-#include "print-session.h"
-#include "gnc-state.h"
-#include "print-session.h"
```
jralls, line 60: These aren't variables, nor does clang-tidy flag them as unused. We want to follow [include-what-you-use](https://google.github.io/styleguide/cppguide.html#Include_What_You_Use) so even if they are included via some other  include they still need to be here.

gnucash/gnome/dialog-fincalc.c
```diff
@@ -92,10 +92,6 @@ static unsigned int periods[] =
     365, /* daily (365) */
 };
 
-/* This static indicates the debugging module that this .o belongs to.  */
-static QofLogModule log_module = GNC_MOD_GUI;
```
jralls, line 96: Set `__attribute__((unused))` instead of deleting.

gnucash/gnome/window-autoclear.c
```diff
@@ -37,8 +37,6 @@
 
 #define WINDOW_AUTOCLEAR_CM_CLASS "window-autoclear"
 
-static QofLogModule log_module = GNC_MOD_GUI;
```
jralls, line 40: `__attribute__((unused))`

gnucash/gnucash-cli.cpp
```diff
@@ -44,9 +44,6 @@
 
 namespace bl = boost::locale;
 
-/* This static indicates the debugging module that this .o belongs to.  */
-static QofLogModule log_module = GNC_MOD_GUI;
```
jralls, line 48: `__attribute__((unused))`

gnucash/import-export/aqb/dialog-ab-select-imexporter.c
```diff
@@ -32,8 +32,6 @@
 #include "dialog-ab-select-imexporter.h"
 #include <dialog-utils.h>
 
-static QofLogModule log_module = G_LOG_DOMAIN;
```
jralls, line 35: `__attribute__((unused))`

libgnucash/backend/sql/gnc-budget-sql.cpp
```diff
@@ -51,8 +51,6 @@
 #define AMOUNTS_TABLE "budget_amounts"
 #define AMOUNTS_TABLE_VERSION 1
 
-static QofLogModule log_module = G_LOG_DOMAIN;
```
jralls, line 54: `__attribute__((unused))`

libgnucash/backend/sql/gnc-commodity-sql.cpp
```diff
@@ -44,8 +44,6 @@
 #include "splint-defs.h"
 #endif
 
-static QofLogModule log_module = G_LOG_DOMAIN;
```
jralls, line 47: `__attribute__((unused))`

libgnucash/backend/sql/gnc-order-sql.cpp
```diff
@@ -45,8 +45,6 @@
 
 #define _GNC_MOD_NAME   GNC_ID_ORDER
 
-static QofLogModule log_module = G_LOG_DOMAIN;
```
jralls, line 48: `__attribute__((unused))`

libgnucash/backend/xml/gnc-transaction-xml-v2.cpp
```diff
@@ -46,7 +46,6 @@
 
 #include "sixtp-dom-parsers.h"
 
-static const QofLogModule log_module = G_LOG_DOMAIN;
```
jralls, line 49: `__attribute__((unused))`

jralls (changes requested):
99% OK.

libgnucash/core-utils/binreloc.c
```diff
@@ -106,22 +106,23 @@ _br_find_exe (Gnc_GbrInitError *error)
...
-        return NULL;
+        /* Native Mac, but not Aqua */
+        uint32_t size2 = buf_size;
+        if (_NSGetExecutablePath (path2, &size2) != 0)
+        {
+            /* buffer not big enough or some other error */
+            if (error)
+                *error = GNC_GBR_INIT_ERROR_NOMEM;
+            return NULL;
+        }
```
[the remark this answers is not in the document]
richardcohen: I now realise that the block is not needed anyway -- those `ifdef`s are confusing.

BTW I notice that the C compile options include -Wdeclarations-after-statement, although it is not mentioned in the coding standard. Would it be OK to remove it?

libgnucash/backend/dbi/test/test-backend-dbi-basic.cpp
```diff
@@ -250,7 +250,6 @@ destroy_database (gchar* url)
     gchar* port = NULL;
     auto pgsql = "pgsql";
     dbi_conn conn = NULL;
-    auto errfmt = "Unable to delete tables in %s: %s";
```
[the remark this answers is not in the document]
richardcohen: The unused variable is "errnum" on line 299.
In order to use it, the format string needs to be changed.
It made sense to change the format string everywhere.
As a bonus, the compiler will check the format parameters

gnucash/html/gnc-html-webkit2.c
richardcohen: clang-tidy flags them all as unused here (15.0.2 on ubuntu 22.10)
`#include "gnc-state.h"` has to go because it is no longer being used (the call to gnc_state_get_current() is being removed)
The duplicate `#include "print-session.h"` is clearly wrong, but I can leave it if you want

libgnucash/core-utils/binreloc.c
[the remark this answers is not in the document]
jralls: > BTW I notice that the C compile options include -Wdeclarations-after-statement, although it is not mentioned in the coding standard. Would it be OK to remove it?

Yeah. I don't follow the C standard as closely as I do the C++ one so I didn't realize that C99 allows mixing decls and statements, I thought it was just a gcc extension. Since we require C11 (and should probably bump it to C17 so it matches the C++ standard requirement) we can certainly allow it. I'll go further and say we should encourage it in the coding standard because it makes it much easier to declare and initialize instead of assigning, and also allows clustering variables and their uses. @gjanssens Any comment on that?

gnucash/html/gnc-html-webkit2.c
jralls: I checked manually. gnc-state.h's only use is as you say also removed in this change. print-session.h isn't, but gnc-html.h declares several macros and functions used in this file so it should stay.

libgnucash/core-utils/binreloc.c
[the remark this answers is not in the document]
gjanssens: I'm all for allowing this. I agree it makes code much more readable.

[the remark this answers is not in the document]
gjanssens: I'm fine with switching to C17 as well by the way, though I have no idea what improvements have happened since C98 we could make use of.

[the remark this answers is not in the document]
jralls: I [dug into that](https://en.cppreference.com/w/c/language/history) a little bit yesterday. It turns out that there's only one significant change between C11 and C17, the behavior if you try to `realloc(foo, 0)`.  We don't use `realloc` so it doesn't matter to us, so there's no reason to change from C11.
 
Some C98 and C11 features that would let us make C code more C++-ish: Declaring variables at point of use, type `bool` instead of `TRUE` and `FALSE` macros to int, `inline` functions, and maybe some others that need a little more study to understand.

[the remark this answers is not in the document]
Bob-IT: OK with me, just need to remember to do it.

richardcohen:
Rebased onto master -- I'm not sure why it thinks there are still conflicts ??

gnucash/gnucash-cli.cpp
richardcohen: This is C++, so I think [[maybe_unused]] is better

richardcohen:
Seems to be some networking? problem in Azure -- Ubuntu can't download its dependencies.
Some of the comments have only just turned up.

gnucash/gnucash-cli.cpp
jralls: Yes it is.

jralls (approved):

jralls:
Thanks!

PR #1521, opened 2023-01-31
Use const references for for-loops to avoid unnecessary copies

richardcohen:
Found by clazy - clazy-range-loop-reference


jralls:
Thanks!

PR #1524, opened 2023-02-02
Unique values in option example

flywire:
See https://lists.gnucash.org/pipermail/gnucash-devel/2023-January/046498.html and https://github.com/Gnucash/gnucash/pull/1501#issuecomment-1386348045

I'd suggest two changes to the previous discussion:

```
'name (N_ "Sample Report")
'menu-name (N_ "Sample Options Report")
```

PR #1525, opened 2023-02-05
Unique values in option example

flywire:


jralls:
Please remove the last commit: Pull requests shouldn't have any merge commits. You can rebase on today's master to get that change.

The first commit would be better split into three: Renaming hello-world.scm to sample-report.scm, removing welcome-to-gnucash.scm, and your changes to sample-report.scm.

I think it would be better to remove the crash option from sample-report and the exception from test-stress-options so that sample-report is tested.

To force-push your repaired branch say `git push -f <remote> Hello5`, substituting the remote name for `<remote>`.

flywire:
> To force-push your repaired branch say `git push -f <remote> Hello5`, substituting the remote name for `<remote>`.

Forcing pushes back to the repo is failing possibly due to wrong <remote> reference. Can you give me that line explicitly referencing https://github.com/Gnucash/gnucash/pull/1524#issuecomment-1418263961 ?

jralls:
It failed when you tried to push Hello5 to hello4 because you left off the `-f` in `git push origin Hello5:hello4`. Maybe it will be easier to remember if you spell it out: `git push --force origin Hello5`.

A non-forced push expects the source branch (the one you're pushing) to include the commit that is HEAD on the destination (i.e. remote) branch and adds the commits on the source to that commit. A force push *replaces* the HEAD commit and its chain of parents on the destination branch with those from the source branch.

jralls:
Thanks!

PR #1527, opened 2023-02-07
minor changes

christopherlam:
A small collection of tidying up. Primarily: gtk_entry_get_text instead of gtk_editable_get_chars.

### DIFF
--- a/gnucash/gnome-utils/dialog-account.c
+++ b/gnucash/gnome-utils/dialog-account.c
@@ -2196,14 +2196,13 @@ static void
 gnc_account_renumber_update_examples (RenumberDialog *data)
 {
     gchar *str;
-    gchar *prefix;
     gint   interval;
     gint   digits;
     unsigned int num_digits = 1;
 
     g_return_if_fail (data->num_children > 0);
 
-    prefix = gtk_editable_get_chars (GTK_EDITABLE(data->prefix), 0, -1);
+    const gchar *prefix = gtk_entry_get_text (GTK_ENTRY(data->prefix));
     interval = gtk_spin_button_get_value_as_int (GTK_SPIN_BUTTON(data->interval));
     digits = gtk_spin_button_get_value_as_int (GTK_SPIN_BUTTON(data->digits));
 
@@ -2225,7 +2224,7 @@ gnc_account_renumber_update_examples (RenumberDialog *data)
     else
         num_digits = digits;
 
-    if (strlen (prefix))
+    if (prefix && *prefix)
         str = g_strdup_printf ("%s-%0*d", prefix, num_digits, interval);
     else
         str = g_strdup_printf ("%0*d", num_digits, interval);
@@ -2243,7 +2242,6 @@ gnc_account_renumber_update_examples (RenumberDialog *data)
     gtk_label_set_text (GTK_LABEL(data->example2), str);
 
     g_free (str);
-    g_free (prefix);
 }
 
 void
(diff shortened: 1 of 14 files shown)


gnucash/gnome-utils/dialog-account.c
```diff
@@ -2206,14 +2206,13 @@ static void
...
     gchar *str;
-    gchar *prefix;
     gint   interval;
     gint   digits;
     unsigned int num_digits = 1;
 
     g_return_if_fail (data->num_children > 0);
 
-    prefix = gtk_editable_get_chars (GTK_EDITABLE(data->prefix), 0, -1);
+    const gchar *prefix = gtk_entry_get_text (GTK_ENTRY(data->prefix));
```
jralls, line 2215: `gtk_entry_get_text` [can return `NULL`](https://gitlab.gnome.org/GNOME/gtk/-/blob/gtk-3-24/gtk/gtkentrybuffer.c#L525) and `strlen` will crash if fed a `NULL` glibc and clang's libc, so the test check at line 2237 and elsewhere needs to be `if (prefix && strlen(prefix))` or better `if (prefix && *prefix)` to not count the chars for no good reason.

gnucash/gnome/dialog-customer.c
```diff
@@ -399,28 +399,22 @@ void
...
 
-    id = gtk_entry_get_text (GTK_ENTRY (cw->id_entry));
-
-    fullname = g_strconcat (name, " (", id, ")", (char *)NULL);
-
-    if (cw->dialog_type == EDIT_CUSTOMER)
-        title = g_strconcat (_("Edit Customer"), " - ", fullname, (char *)NULL);
-    else
-        title = g_strconcat (_("New Customer"), " - ", fullname, (char *)NULL);
+    char *title = g_strdup_printf ("%s - %s (%s)", header, name,
```
jralls, line 413: `g_strdup_printf` is implemented with `g_vasprintf` which in turn calls either the system `vasprintf` if it's available or `vsprintf` otherwise. Being system functions they can't be trusted not to crash if fed a NULL argument, so `fullname` needs to hang around and get tested before being passed to `g_strdup_printf`.

gnucash/gnome/dialog-sx-from-trans.c
```diff
@@ -177,23 +177,22 @@ sxftd_get_end_info(SXFromTransInfo *sxfti)
 
     if (gtk_toggle_button_get_active( GTK_TOGGLE_BUTTON(sxfti->oc_but) ))
     {
-        gchar *text, *endptr;
         guint n_occs;
 
-        text = gtk_editable_get_chars(GTK_EDITABLE(sxfti->n_occurences), 0, -1);
+        const gchar *text = gtk_entry_get_text (GTK_ENTRY(sxfti->n_occurences));
         if (text == NULL || strlen(text) == 0)
```
jralls, line 183: `if (text && *text)`

gnucash/gnome/dialog-vendor.c
```diff
@@ -267,29 +267,22 @@ void
...
+    const char *name = gtk_entry_get_text (GTK_ENTRY(vw->company_entry));
+    if (!name || *name == '\0')
+        name = _("<No name>");
 
-    if (vw->dialog_type == EDIT_VENDOR)
-        title = g_strconcat (_("Edit Vendor"), " - ", fullname, (char *)NULL);
-    else
-        title = g_strconcat (_("New Vendor"), " - ", fullname, (char *)NULL);
+    char *title = g_strdup_printf ("%s - %s (%s)", header, name,
+                                   gtk_entry_get_text (GTK_ENTRY(vw->id_entry)));
```
jralls, line 282: `id` needs to stick around and be tested.

jralls (changes requested):

jralls (approved):

jralls:
@christopherlam LGTM, go ahead and merge.

PR #1528, opened 2023-02-08
Fix crash after upgrading v4 graph reports to v5

richardcohen:
Backtrace:
In ice-9/boot-9.scm:
  1752:10  5 (with-exception-handler _ _ #:unwind? _ #:unwind-for-type _)
...
In unknown file:
           0 (GncOption-set-value #<pointer 0x555da480fe78> (percent . 50.0))

ERROR: In procedure GncOption-set-value:
Wrong type (expecting exact integer): 50.0

### DIFF
--- a/bindings/guile/gnc-optiondb.i
+++ b/bindings/guile/gnc-optiondb.i
@@ -1369,7 +1369,11 @@ inline SCM return_scm_value(ValueType value)
                                   GncOptionRangeValue<int>>)
                     {
                         if (scm_is_pair(new_value))
-                            option.set_value(scm_to_int(scm_cdr(new_value)));
+                        {
+                            // in Gnucash 4, the value might have been saved as floating point
+                            auto double_value = scm_to_double(scm_cdr(new_value));
+                            option.set_value(static_cast<int>(double_value));
+                        }
                         else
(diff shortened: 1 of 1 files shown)


bindings/guile/gnc-optiondb.i
```diff
@@ -1369,7 +1369,11 @@ inline SCM return_scm_value(ValueType value)
                                   GncOptionRangeValue<int>>)
                     {
                         if (scm_is_pair(new_value))
-                            option.set_value(scm_to_int(scm_cdr(new_value)));
+                        {
+                            // in Gnucash 4, the value might have been saved as floating point
+                            auto double_value = scm_to_double(scm_cdr(new_value));
+                            option.set_value(static_cast<int>(double_value));
+                        }
```
jralls, line 1376: This will break reading plot size options set by GC5. The SCM input must be tested with `scm_is_integer` and processed accordingly:
```
if (scm_is_pair(new_value))
{
    auto value_num = scm_cdr(new_value);
    if (scm_is_int(value_num))
        option.set_value(scm_to_int(value_num));
    else
        option.set_value(static_cast<int>(scm_to_double(value_num)));
}
```

jralls:
There being no response, I've fixed this myself.

PR #1529, opened 2023-02-08
Restore the ability to set the plot size in pixels

richardcohen:
FWIW, I use this.

- Text never scales, and IMHO it looks better if the graph behaves the same
- Rescaling is sluggish on older machines
- The behaviour now matches the tooltip

It is unfortunate that the UI is not very discoverable

### DIFF
--- a/libgnucash/engine/gnc-optiondb.cpp
+++ b/libgnucash/engine/gnc-optiondb.cpp
@@ -817,9 +817,8 @@ gnc_register_number_plot_size_option(GncOptionDB* db,
                                      const char* key, const char* doc_string,
                                      int value)
 {
-    // Pixel values don't make much sense so always use percent.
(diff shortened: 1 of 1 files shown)


jralls:
Text usually doesn't change size when the window resizes. It may reflow or not depending on how the image is styled in the page (e.g `float: left` on the image will cause text to be placed to the right of it if there's enough room).

To get the UI to be more discoverable you could recreate the radio + entry from the Scheme code in gnucash/gnome-utils/gnc-option-gtk-ui.cpp,`GncGtkPlotSizeUIItem`. You could make it like the `BothDateEntry` except that it would use two `GtkSpinButton`s instead of an `AbsoluteDateEntry` and a `RelativeDateEntry`.

jralls:
There being no response, I've implemented this myself.

richardcohen:
Thanks for implementing this. It was on my list of things to do.

PR #1530, opened 2023-02-09
Upgrade chartjs to 4.2.0

richardcohen:


jralls:
@richardcohen Have you tested this on Windows? That's important because Windows is stuck at an antiquated version of WebKitGtk, that project having dropped Windows support in later versions. #759 depends on upgrading chartjs too and hasn't moved forward because of @christopherlam's worries and not having made time to resolve them.

richardcohen:
> @richardcohen Have you tested this on Windows? That's important because Windows is stuck at an antiquated version of WebKitGtk, that project having dropped Windows support in later versions. #759 depends on upgrading chartjs too and hasn't moved forward because of @christopherlam's worries and not having made time to resolve them.

No, only on Linux.

christopherlam:
Plus, thanks for the hard work of scrutinising the API changes. I wasn't
looking forward to it.

christopherlam (commented):
Would you be able to modify the tabs to spaces in scm files?

jralls:
Unfortunately @christopherlam's worries were justified. It breaks displaying charts on Windows.

christopherlam:
> Unfortunately @christopherlam's worries were justified. It breaks displaying charts on Windows.

And old libwebkit on windows doesn't have the handy developer console to attempt to fix this. Would chartjs3 be a safer option?

jralls:
> Would chartjs3 be a safer option?

Sadly no. I tried applying #759 and it breaks charts on Windows too.

So we need at least a new javascript solution. One possibility: Mingw64 supplies [v8](https://packages.msys2.org/package/mingw-w64-x86_64-v8?repo=mingw64). If we can figure out how to wire that into webkit1 we're home. Another approach would be to use the [webview](https://github.com/webview/webview) or [litehtml](https://github.com/litehtml/litehtml) in place of WebKitGtk; note that in addition to implementing the cairo renderer for litehtml one would also need to fill out [el_script.cpp](https://github.com/litehtml/litehtml/blob/master/src/el_script.cpp), wiring it into some JS engine.

gjanssens:
> > Would chartjs3 be a safer option?
> 
> Sadly no. I tried applying #759 and it breaks charts on Windows too.
> 
> So we need at least a new javascript solution. One possibility: Mingw64 supplies [v8](https://packages.msys2.org/package/mingw-w64-x86_64-v8?repo=mingw64). If we can figure out how to wire that into webkit1 we're home. Another approach would be to use the [webview](https://github.com/webview/webview) or [litehtml](https://github.com/litehtml/litehtml) in place of WebKitGtk; note that in addition to implementing the cairo renderer for litehtml one would also need to fill out [el_script.cpp](https://github.com/litehtml/litehtml/blob/master/src/el_script.cpp), wiring it into some JS engine.

I hadn't heard of webview before, looks interesting. The changes required for this swap of html renderer are quite important. I don't think that's a viable option for the upcoming 5.0 still. So getting on to a more recent chartjs may have to wait a while.

christopherlam:
> > So we need at least a new javascript solution. One possibility: Mingw64 supplies [v8](https://packages.msys2.org/package/mingw-w64-x86_64-v8?repo=mingw64). If we can figure out how to wire that into webkit1 we're home. Another approach would be to use the [webview](https://github.com/webview/webview) or [litehtml](https://github.com/litehtml/litehtml) in place of WebKitGtk; note that in addition to implementing the cairo renderer for litehtml one would also need to fill out [el_script.cpp](https://github.com/litehtml/litehtml/blob/master/src/el_script.cpp), wiring it into some JS engine.
> 
> I hadn't heard of webview before, looks interesting. The changes required for this swap of html renderer are quite important. I don't think that's a viable option for the upcoming 5.0 still. So getting on to a more recent chartjs may have to wait a while.

I think to swap in webview would be a radical and welcome change that shouldn't wait too long, and would benefit from a lot of beta testers... They could coexist hidden via a `--use-alternate-html` type switch to be used for the GnuCash 5.0 series only. This is how I'd structured the old and new aging and owner reports, and the swap was a simple guid dance.

richardcohen:
> Would you be able to modify the tabs to spaces in scm files?
It wasn't intentional. Thanks for checking

Bob-IT:
I had a look at `webview` by doing the following..

I built master with this PR and then exported an asset chart and balance report, also made a copy of the `chartjs` folder.
On my Linux VM built the `webview` basic.c example OK. Then modified it to display a `gtk` dialog with two buttons that launched each report from a file path which was OK. With the aid of a `js` function using the `webview` bind function was able to get the href links to print in the terminal and so could be fed back to GnuCash to open registers. I was able to have both charts displayed in separate windows and displayed quickly.

On my windows 10 build machine I tried to do the the same, already had MS Edge so did not need to install the runtime, with these observations but they may be down to my machine or probably the way I am building the `gtk` dialog.
I could only have one report window open, if I opened the chart and then pressed the dialog balance button it would not display until I had closed the chart.
When the report opens, the window is blank and smaller which then enlarges to display the report.

From my understanding, using `webview` creates native windows so they can not be embedded in the plugin report page.
Not sure what the reaction would be to that.

If this is the way to go, I suppose for testing, not users, `webview` could be added to borrowed and a dialog added to a menu.

jralls:
I hadn't looked at the webview source code before now. The window is the second argument to `win32_edge_engine::win32_edge_engine(bool, void*)` so if we can get the MS WebView2 that provides the backend to draw to a cairo surface we can send that to a GtkDrawing and put it in our `plugin-page-report`.  

It's a pretty simple implementation, the whole thing is in a single 2300 line header, I guess it's not too surprising that it can't handle multiple instances of itself. We might be better off taking the idea and the general outline and reimplmenting it ourselves.

christopherlam:
I may suggest to allow chartjs>=3 as a platform-specific code, because windows build will otherwise inhibit progress.

jralls:
@christopherlam OK, but consider that the vast majority of our users are on Windows so any progress will have a rather limited audience.

jralls:
This is superseded by #2110

PR #1531, opened 2023-02-11
Bug 798748 - Transaction Notes field's value does not appear in reversed transaction

agwekixj:
A work around is to copy and paste the value from the original transaction into the reverse transaction.  
Note: Other transaction field values are also not being propagated.  For example, the document link.
An appropriate fix appears to be to add the missing values in xaccTransCloneNoKvp()

### DIFF
--- a/libgnucash/engine/Transaction.c
+++ b/libgnucash/engine/Transaction.c
@@ -2432,6 +2432,8 @@ xaccTransGetDocLink (const Transaction *trans)
     {
         GValue v = G_VALUE_INIT;
         Transaction *t = (Transaction*) trans;
+	if (!qof_instance_has_kvp(QOF_INSTANCE (trans)))
+            return NULL;
         qof_instance_get_kvp (QOF_INSTANCE (trans), &v, 1, doclink_uri_str);
         t->doclink = G_VALUE_HOLDS_STRING (&v) ? g_value_dup_string (&v) : NULL;
         g_value_unset (&v);
@@ -2447,6 +2449,8 @@ xaccTransGetNotes (const Transaction *trans)
     {
         GValue v = G_VALUE_INIT;
         Transaction *t = (Transaction*) trans;
+	if (!qof_instance_has_kvp(QOF_INSTANCE (trans)))
+            return NULL;
         qof_instance_get_kvp (QOF_INSTANCE (trans), &v, 1, trans_notes_str);
         t->notes = G_VALUE_HOLDS_STRING (&v) ? g_value_dup_string (&v) : NULL;
         g_value_unset (&v);
@@ -2462,6 +2466,8 @@ xaccTransGetIsClosingTxn (const Transaction *trans)
     {
         Transaction* trans_nonconst = (Transaction*) trans;
         GValue v = G_VALUE_INIT;
+	if (!qof_instance_has_kvp(QOF_INSTANCE (trans)))
+            return FALSE;
         qof_instance_get_kvp (QOF_INSTANCE (trans), &v, 1, trans_is_closing_str);
         if (G_VALUE_HOLDS_INT64 (&v))
             trans_nonconst->isClosingTxn_cached = (g_value_get_int64 (&v) ? 1 : 0);
@@ -2604,6 +2610,8 @@ xaccTransGetReadOnly (Transaction *trans)
     if (trans->readonly_reason == is_unset)
     {
         GValue v = G_VALUE_INIT;
+	if (!qof_instance_has_kvp(QOF_INSTANCE (trans)))
+            return NULL;
         qof_instance_get_kvp (QOF_INSTANCE(trans), &v, 1, TRANS_READ_ONLY_REASON);
         trans->readonly_reason = G_VALUE_HOLDS_STRING (&v) ?
             g_value_dup_string (&v) : NULL;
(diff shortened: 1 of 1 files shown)


gjanssens:
While this PR is technically ok, I wonder if this is really the behaviour we want. I would expect the note on a reversed transaction to hold an explanation as to why the transaction was reversed, not the same note as on the original transaction. Same for the document link. To my mind it doesn't make sense that this is the same on a reversed transaction. If the reversal is because the transaction was credited, the document link would point to the credit document, not the debit note.

What is your use case that you want these fields to be copied automatically and unconditionally ?

agwekixj:
> While this PR is technically ok, I wonder if this is really the behaviour we want. I would expect the note on a reversed transaction to hold an explanation as to why the transaction was reversed, not the same note as on the original transaction. Same for the document link. To my mind it doesn't make sense that this is the same on a reversed transaction. If the reversal is because the transaction was credited, the document link would point to the credit document, not the debit note.
> 
> What is your use case that you want these fields to be copied automatically and unconditionally ?

Thanks for responding. 

While I agree it is likely the note and document link of a reversed transaction may be different than the original transaction, I argue they should initially be the same.  I have not been able to think of a use case where this behavior would not be what is expected.

Consider a return of a purchase in which the original transaction note contains the order number and the document link a reference to an invoice.  The reverse transaction should initially include these same details.  A user may subsequently modify the reversed transaction note, referenced document, the link, or any of the other fields in the reversed transaction as desired.

jralls:
> Consider a return of a purchase in which the original transaction note contains the order number and the document link a reference to an invoice.

That's not what reversing transactions are for. See [Add a reversing transaction](https://www.gnucash.org/docs/v4/C/gnucash-help/trans-delete.html) in the manual.

>  A user may subsequently modify the reversed transaction note, referenced document, the link, or any of the other fields in the reversed transaction as desired.

Or a user could, without any change to GnuCash, *duplicate* the transaction and move the amount from debit to credit, adjusting such other fields as appropriate to the user's purpose.

libgnucash/engine/Transaction.c
```diff
@@ -665,6 +665,10 @@ xaccTransCloneNoKvp (const Transaction *from)
 			    qof_instance_get_book(from));
 
     xaccTransBeginEdit(to);
+
+    xaccTransSetNotes(to, xaccTransGetNotes(from));
+    xaccTransSetDocLink(to, xaccTransGetDocLink(from));
+
```
jralls, line 671: Notes and DocLink are stored in KVP, so this breaks the documented purpose of the function.

agwekixj: Good point.  I've moved the change into xaccTransClone().

Thanks.

agwekixj:
> > Consider a return of a purchase in which the original transaction note contains the order number and the document link a reference to an invoice.
> 
> That's not what reversing transactions are for. See [Add a reversing transaction](https://www.gnucash.org/docs/v4/C/gnucash-help/trans-delete.html) in the manual.
> 

Your link points to the "Deleting a Transaction" page but, regardless, in my opinion, the transaction note and document link in the reverse transaction should initially be provided with the same values as the original transaction for the use case described in "Adding Reversing Transaction" for the same reasons.

> > A user may subsequently modify the reversed transaction note, referenced document, the link, or any of the other fields in the reversed transaction as desired.
> 
> Or a user could, without any change to GnuCash, _duplicate_ the transaction and move the amount from debit to credit, adjusting such other fields as appropriate to the user's purpose.

True but significantly more error prone than the previously mentioned work around of copying the note from the original transaction.

libgnucash/engine/Transaction.c
jralls: Where it shouldn't be necessary, `xaccTransClone` calls `qof_instance_copy_kvp` that should already copy the notes, doclink, and everything else stored in KVP.

`xaccTransReverse` calls `xaccTransClone`, meaning that the intended behavior is what you want and the discussion about use-case is irrelevant.

Testing shows that it does, that the doclink mark shows on the reversed transaction and Manage Document Links shows the same link as the original. *But* although the notes slot is duplicated in the reversed transaction it doesn't display in the register until one closes and reopens the book; simply closing and reopening the register isn't sufficient. It's pretty odd because the doclink code is identical both in Transaction.c and split-register-model.c.

christopherlam:
I think it's the `xaccTransCloneNoKvp` which must reset `trans->notes`

jralls:
Please sqaush the commits into one and force-push.

christopherlam:
Before this is merged in; prior to returning NULL, wouldn't it be worth
setting the various `char*` to NULL as well to speed up future getters?

agwekixj:
> 

Setting the transactions cached values when the transaction instance has an empty KVP frame is precisely what triggers the issue we're attempting to resolve.   The alternate approach of coercing their reinitialization in xaccTransClone() after the KVP frame is cloned was suggested earlier and found wanting.

christopherlam:
As discussed in the bug, the various `struct->char*` were meant to store a copy of the string while the GValue was unset, and resulted in less leaks. If `g_value_take_string` and `g_value_get_static_string` etc would be a better choice, I'm less familiar with the ownership transfer dances.

jralls:
As I just noted on the [bug](https://bugs.gnucash.org/show_bug.cgi?id=798748) I've removed the caches, so this PR is no longer valid. @agwekixj thanks for bashing through this, it was a lot of help in understanding the situation.

PR #1534, opened 2023-02-16
Gsettting Changes

Bob-IT:
This changes the `gnc_gsettings_version_upgrade` process so that release versions greater than the current GnuCash version in the `pref_transformations.xml` file are not processed. It also fixes a windows bug that was failing on the glib error...
`WARN <GLib-GIO> subscribe() failed: only 64 different paths may be watched.`

The way I have done this is that after each migrate/obsolete pass, the function `clear_gset_object`(will rename this to `clear_gsettings_object`), is called which finds the GSettings object in the schema_hash, removes it and then unref's the object to release the schema path.

The down side on this is that a two GSettings objects are created and removed for each migrate and one created and removed for each obsolete.

Not sure if this is an issue, another way I did think of is to have two static GSettings objects and compare the `schema-id` to see if it can be reused for the next change, if not unref it and create a new one, may also need a couple of duplicate functions to use them, not sure have not tried it.

### DIFF
--- a/gnucash/gschemas/pref_transformations.xml
+++ b/gnucash/gschemas/pref_transformations.xml
@@ -2152,16 +2152,20 @@
 <obsolete old-path="org.gnucash.window.pages.account-tree.summary"
           old-key="end-period"/>
 
+</release>
+
+
+<release version="5000">
+
 <deprecate old-path="org.gnucash.GnuCash.dialogs.business.invoice"
            old-key="invoice-printreport" />
 
 </release>
 
-<!--
 <release version="6000">
 
 <obsolete old-path="org.gnucash.GnuCash.dialogs.business.invoice"
           old-key="invoice-printreport"/>
 
 </release>
--->
+
--- a/libgnucash/app-utils/gnc-gsettings.cpp
+++ b/libgnucash/app-utils/gnc-gsettings.cpp
@@ -688,7 +688,7 @@ void gnc_gsettings_load_backend (void)
 
 static GVariant *
 gnc_gsettings_get_user_value (const gchar *schema,
-                         const gchar *key)
+                              const gchar *key)
 {
     GSettings *settings_ptr = gnc_gsettings_get_settings_ptr (schema);
     g_return_val_if_fail (G_IS_SETTINGS (settings_ptr), NULL);
--- a/libgnucash/app-utils/gnc-gsettings.cpp
+++ b/libgnucash/app-utils/gnc-gsettings.cpp
@@ -870,9 +870,12 @@ void gnc_gsettings_version_upgrade (void)
     auto ogG_maj_min = gnc_gsettings_get_user_value (GNC_PREFS_GROUP_GENERAL, GNC_PREF_VERSION);
     auto og_maj_min = gnc_gsettings_get_user_value (GSET_SCHEMA_OLD_PREFIX "." GNC_PREFS_GROUP_GENERAL, GNC_PREF_VERSION);
 
+    auto cur_maj_min = PROJECT_VERSION_MAJOR * 1000 + PROJECT_VERSION_MINOR;
+
     if (!ogG_maj_min && !og_maj_min) // new install
     {
-        LEAVE("");
+        gnc_gsettings_set_int (GNC_PREFS_GROUP_GENERAL, GNC_PREF_VERSION, cur_maj_min);
+        LEAVE ("Setting Previous compatibility level to current version: %i", cur_maj_min);
         return;
     }
 
@@ -886,8 +889,6 @@ void gnc_gsettings_version_upgrade (void)
     }
     if (og_maj_min)
         g_variant_unref (og_maj_min);
-
-    auto cur_maj_min = PROJECT_VERSION_MAJOR * 1000 + PROJECT_VERSION_MINOR;
 
     PINFO ("Previous setting compatibility level: %i, Current version: %i", old_maj_min, cur_maj_min);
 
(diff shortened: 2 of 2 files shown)


Bob-IT:
Geert's PR #1535 is better than my last commit so that can be ignored, will remove it later.

gnucash/gschemas/pref_transformations.xml
```diff
@@ -2152,16 +2152,20 @@
 <obsolete old-path="org.gnucash.window.pages.account-tree.summary"
           old-key="end-period"/>
 
+</release>
+
+
+<release version="5000">
+
```
gjanssens, line 2159: Why do you separate this out ?

libgnucash/app-utils/gnc-gsettings.cpp
```diff
@@ -823,8 +823,9 @@ void gnc_gsettings_version_upgrade (void)
 
     if (!ogG_maj_min && !og_maj_min) // new install
     {
-        LEAVE("");
-        return;
+        PINFO ("Setting Previous compatibility level to major version minus one: %i", PROJECT_VERSION_MAJOR * 1000 - 1);
+        gnc_gsettings_set_int (GNC_PREFS_GROUP_GENERAL, GNC_PREF_VERSION, PROJECT_VERSION_MAJOR * 1000 - 1);
+        ogG_maj_min = gnc_gsettings_get_user_value (GNC_PREFS_GROUP_GENERAL, GNC_PREF_VERSION);
```
gjanssens, line 872: On a new install there are no settings to migrate or obsolete. So the only thing to do is setting the current prefs-version. There's no need to go through all other migration steps.

libgnucash/app-utils/gnc-gsettings.cpp
```diff
@@ -672,6 +672,23 @@ gnc_gsettings_get_user_value (const gchar *schema,
...
+
+    // Now release the gsettings object
+    gset = static_cast<GSettings*> (g_hash_table_lookup (schema_hash, path->c_str()));
+    DEBUG ("Looking for schema %s to release gsettings, returned %p",  path->c_str(), gset);
+
+    if (gset)
+    {
+        DEBUG ("Clearing gsettings object %p", gset);
+        g_hash_table_remove (schema_hash, path->c_str());
+        g_clear_object (&gset);
```
gjanssens, line 688: g_clear_object doesn't add much here over g_object_unref.  gset will go out of scope right after this call, so there nowhere you'll ever test if the pointer gset itself is null. The object it points to is not affected by g_clear_object other than the usual refcounting machinery. If, when you call this function, there's only one reference left, the object will be freed from memory, otherwise its refcount will only be decreased. That's exactly the same as g_object_unref.

gjanssens: Other than that, I don't think we'll need this at all. IMO we can just stop adding most objects in our hash table.

gjanssens:
From first looks, the full solution will be a mix of your and my PR. I'm only focussing on the hash table bits. From your PR the parts that check for newer version numbers in the transformation tables is valid and needed. And a fixed version of setting the prefs-version when gnucash is called for the first time on a new install is also wanted.

Do we really want to add the debug dump function to the code ? It was very helpful for your local debugging, but I don't know if we have to carry it forward in production code.

gnucash/gschemas/pref_transformations.xml
Bob-IT: I just did this to make it clearer for me when testing, just to group the changes I made and maybe when that section is removed.

libgnucash/app-utils/gnc-gsettings.cpp
Bob-IT: I was thinking that if a change was was made for release 5.2, admittedly it may only be a depreciate which does not do much, and I did a new install of 5.4, I would want the previous changes to be processed.

libgnucash/app-utils/gnc-gsettings.cpp
Bob-IT: OK, a bit over the top and as pointed out you PR replaces this commit so can be dropped.

Bob-IT:
> Do we really want to add the debug dump function to the code ? It was very helpful for your local debugging, but I don't know if we have to carry it forward in production code.

I thought it may be useful but I can remove and keep locally to use when needed.

libgnucash/app-utils/gnc-gsettings.cpp
gjanssens: Unless it's me I think you may be missing the point. On a *new* install there are *no* settings in GSettings. So there is no history to transform to the current version. If it was an uninstall old version and install new version, that's not a new install in this case. In that case the prefs-version would still be set to 5.2 and hence anything that needs changing to get to 5.4 will be done.

For this to work, an empty prefs-version should be set on first-run to the current version. That would ensure that *future* settings transformations work properly.

gjanssens:
> > Do we really want to add the debug dump function to the code ? It was very helpful for your local debugging, but I don't know if we have to carry it forward in production code.
> 
> I thought it may be useful but I can remove and keep locally to use when needed.

Well, it was mostly a question. I don't have a strong opinion on this. If you prefer to keep it, that's fine with me as well.

libgnucash/app-utils/gnc-gsettings.cpp
Bob-IT: No It's me, over thinking it, forgot that there would be no preferences on a new install. When you start thinking one way it is sometimes hard to change track. Will set it to the current version.

Bob-IT:
Will tidy up and push later.

PR #1535, opened 2023-02-16
Bug 798749 - Fails to read gsettings on startup

gjanssens:
This alternative implementation gets rid of caching the GSettings objects except for keeping track of signal handlers.

Will slow down GSettings interactions, but in general they are already that fast, there is no noticeable side effect.

### DIFF
--- a/libgnucash/app-utils/gnc-gsettings.cpp
+++ b/libgnucash/app-utils/gnc-gsettings.cpp
@@ -89,32 +89,19 @@ static gboolean gnc_gsettings_is_valid_key(GSettings *settings, const gchar *key
 
 static GSettings * gnc_gsettings_get_settings_ptr (const gchar *schema_str)
 {
-    GSettings *gset = NULL;
-    gchar *full_name = gnc_gsettings_normalize_schema_name (schema_str);
-
     ENTER("");
-    if (!schema_hash)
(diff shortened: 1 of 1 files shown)


jralls:
Works on Windows.

Bob-IT:
@gjanssens 
I have pushed my changes but have noticed that there is a segfault on exit...

`Thread 1 "gnucash" received signal SIGSEGV, Segmentation fault.
0x00007ffff6f085dc in gnc_gsettings_remove_cb_by_func (schema=0x7ffff7a7a26f "history", key=0x0, func=0x7ffff7a32ca0 <gnc_plugin_history_list_changed>, user_data=0x555555c5c330) at /var/tmp/portage/app-office/gnucash-4.901/work/gnucash-4.901/libgnucash/app-utils/gnc-gsettings.cpp:274
274	        if (!G_IS_SETTINGS (settings_ptr))`

gjanssens:
@Bob-IT I got a similar crash yesterday, but not today. I am in the process of rewriting the signal handler tracking, so that may fix the segfault.

gjanssens:
@Bob-IT can you retest with the most recent master branch ?

Bob-IT:
@gjanssens  Looks OK, now to see what you have changed and get an idea of how it works now.

PR #1537, opened 2023-02-21
Warn for deprecated gtk and glib

richardcohen:


jralls:
Does a bit more than advertised up top, but they're all improvements. Thanks!

PR #1538, opened 2023-02-26
Make the scheme tests fail if they load an external gnucash file

richardcohen:


jralls:
What motivated this?

Under what circumstances would this happen and not cause the supposed test to fail in CI where the out-of-tree file couldn't exist?

richardcohen:
> Under what circumstances would this happen and not cause the supposed test to fail in CI where the out-of-tree file couldn't exist?

I have the Ubuntu distro version of gnucash installed locally in /usr and I noticed that after switching between branches, building etc., test-stress-options started failing with odd stack traces.

Long story short, scm files were being loaded from /usr/share/guile/site/3.0/gnucash/

The loading problem is solved by adding missing MAKE_LINKS, and an entry in the load-path. I have a patch to fix it, but it would be nice to include something like this to protect me when building locally -- obviously I would need to fix it for the Mac build.

common/cmake_modules/GncAddTest.cmake
```diff
@@ -109,7 +109,20 @@ endfunction()
 
 function(gnc_add_scheme_test _TARGET _SOURCE_FILE)
   add_test(${_TARGET} ${CMAKE_COMMAND} -E env
-    ${GUILE_EXECUTABLE} --debug -c "(load-from-path \"${_TARGET}\")(exit (run-test))"
+    ${GUILE_EXECUTABLE} --debug -c "
+    (format #t \"%load-path =~%\")
+    (for-each (lambda(x)(format #t \"~s~%\" x)) %load-path)
+    (format #t \"%load-compiled-path =~%\")
+    (for-each (lambda(x)(format #t \"~s~%\" x)) %load-compiled-path)
```
jralls, line 116: This will add a lot of noise to test output. How about making it conditional on loading a file from outside build-dir?

jralls:
Thanks.

christopherlam:
The 7cae61d97a commit (I think) is causing test-owner-report, test-new-owner-report, test-stress-options to fail, when I have gnucash installed in ubuntu. I'm not sure why suddenly only these tests would fail.

richardcohen:
> Loading guile/site file from outside build tree! "/usr/share/guile/site/3.0/gnucash/reports/standard/owner-report.scm"

`owner-report.scm` was renamed to `new-owner-report.scm`.
Perhaps you have an old dead symlink at `.../master-build/share/guile/site/3.0/gnucash/reports/standard/owner-report.scm`?

I am hopeful that blowing away `.../master-build/share/guile/site/3.0/gnucash/` and rebuilding will work.
Otherwise, try blowing away all of `.../master-build/`

christopherlam:
Ok this was it, thanks. I'd done a `ninja clean` which didn't rid symlinks, so, I did a fully rebuild and now there's no failure.

PR #1543, opened 2023-02-28
By default, filter out online_wiggle in test-gnc-quotes

richardcohen:
The test relies on a lot of external dependencies
- installed finance-quote-wrapper
- alpha vantage key
- network
- working remote server

Running ./bin/test-gnc-quotes from the command line will still include online_wiggle


jralls:
With the demise of yahoo_json this is needed more. Thanks.

PR #1544, opened 2023-02-28
I18 n

fellen:


gnucash/report/report-core.scm
```diff
@@ -148,6 +148,7 @@
 
 ;; Define those strings here to make changes easier and avoid typos.
 (define gnc:menuname-reports "Reports/StandardReports")
+;; Translators: In reality it is "_Activa & Passiva". The English use sometimes strange terms.
```
jralls, line 151: Activa & Passiva is pretty much everything including income and expense, so is that really the right thing to call the menu? Not that Assets & Liabilities is much better, so maybe we should rename it in 5.0. It has the balance sheets, asset and liability balance pie charts, asset, liability, and net worth balance-over-time charts, the portfolio reports, and the ledger and journal reports. Maybe Balances & Investments would be a better title?

gnucash/gnome/gnc-plugin-page-register.c
```diff
@@ -3405,22 +3405,22 @@ gnc_plugin_page_register_set_filter_tooltip (GncPluginPageRegister* page)
     {
         gchar* sdate = qof_print_date (priv->fd.start_time);
         t_list = g_list_prepend
-            (t_list, g_strdup_printf ("%s %s", _("Start Date:"), sdate));
+            (t_list, g_strdup_printf ("%s: %s", _("Start Date"), sdate));
```
jralls, line 3408: No, different languages use punctuation differently, and separating the msgid from the format is bad because it stops translators from reordering it. It should be `_("Start Date: %s")`.

jralls (changes requested):

gnucash/report/report-core.scm
fellen: Another approach: `Balance oriented` or based vs. `Income oriented` ,but then we should perhaps change the source strings.

gnucash/gnome/gnc-plugin-page-register.c
fellen: > No, different languages use punctuation differently, and separating the msgid from the format is bad because it stops translators from reordering it.

I tried to get less messages, but you are right, in fr there should be a nbsp between term and colon.

IIRC there are more hardcoded colons, mostly in reports. Perhaps a separate const string for the colon would be a general solution.?
> It should be `_("Start Date: %s")`.

gnucash/report/report-core.scm
jralls: Yes, I do mean to change the English name (or names) of the menu items. Rearranging the reports in each is also possible, as is adding one or more submenus.

The `Income & Expense` submenu contains various income and expense category charts, the cash flow report and barchart, Income Statement and Profit & Loss report (which I think are the same, will check later), and the Trial Balance report.

Our Cash Flow report is quite different from what [accountants call a cash flow statement](https://www.investopedia.com/terms/c/cashflowstatement.asp) though I suppose one might be able to get there with careful account selection. There's even an [old bug](https://bugs.gnucash.org/show_bug.cgi?id=522171) on the nomenclature. Regardless, whether it's an income/expense report depends on what accounts you select. Trial Balance is also only peripherally an income/expense report in that its primary use is to show whether or not you recorded all of your trading gains and losses--but it does that by examining balances of all accounts in the book.

Balance-oriented and Income-oriented are descriptive but seem to me a bit stilted. I like the idea but I want the wording to percolate for a bit to see if something more fluent comes to mind.

gnucash/gnome/gnc-plugin-page-register.c
jralls: > IIRC there are more hardcoded colons, mostly in reports. Perhaps a separate const string for the colon would be a general solution.?

[Not really](https://www.gnu.org/software/gettext/manual/gettext.html#Preparing-Strings), and that doesn't even mention that `printf` formatting is deficient because the translator can't reorder the inserted tokens, never mind that they're not descriptive and we often don't even provide comments explaining what gets put where.

jralls:
@fellen I hope this satisfies you for the stock transaction assistant strings. I removed all but one of the context translations and capitalized the page names so that they'd use the same msgid as the inserted actions. "Cash" in the context of a stock is a different meaning that e.g. Cash in Wallet: Nobody is handing over bank notes when buying stock.

PR #1546, opened 2023-03-03
Bug798759 saving register 'with sub-account' registers 

Bob-IT:
Register widths are saved in the .gcm file under sections based on the account guid. When you have for example an 'Assets' register open and the 'with sub-account' register page 'Assets+' open, they will both save there settings to the same section as there is only one guid.

To fix this, add a '+' to the end of the guid for the sub-account section as done for the title.

Also, I have modified @jralls commit earlier on to not use xaccAccountSetSortReversed as the 'sort order / sort by and filter by' are all stored in the register. This also fixes the general journal as there is no default account and if saved open segfaults Gnucash on restart.

Currently there is a function to remove those settings from used accounts and store them in the .gcm file, if I remember I added that at the beginning of 4.0 so maybe in 5.0 that can be removed but not sure how to handle the depreciation of the account functions.

The last commit fixes setting the cells insensitive when the register is in reverse order and also when the preference setting to move future transactions after the blank transaction. Could not think of another way of doing this.

### DIFF
--- a/gnucash/gnome/gnc-split-reg.c
+++ b/gnucash/gnome/gnc-split-reg.c
@@ -2152,11 +2152,14 @@ gnc_split_reg_set_sort_reversed(GNCSplitReg *gsr, gboolean rev, gboolean refresh
      *       In other words, qof_query_set_sort_increasing should
      *       always use the inverse of rev.
      */
+    SplitRegister *reg = gnc_ledger_display_get_split_register (gsr->ledger);
     Query *query = gnc_ledger_display_get_query( gsr->ledger );
+
+    gnc_split_register_set_reverse_sort (reg, rev);
+
     qof_query_set_sort_increasing (query, !rev, !rev, !rev);
     gsr->sort_rev = rev;
-    Account *acct = gnc_ledger_display_leader (gsr->ledger);
-    xaccAccountSetSortReversed(acct, rev);
+
     if (refresh)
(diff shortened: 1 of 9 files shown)


gnucash/gnome/gnc-split-reg.c
```diff
@@ -2152,11 +2152,14 @@ gnc_split_reg_set_sort_reversed(GNCSplitReg *gsr, gboolean rev, gboolean refresh
...
      */
+    SplitRegister *reg = gnc_ledger_display_get_split_register (gsr->ledger);
     Query *query = gnc_ledger_display_get_query( gsr->ledger );
+
+    gnc_split_register_set_reverse_sort (reg, rev);
+
     qof_query_set_sort_increasing (query, !rev, !rev, !rev);
     gsr->sort_rev = rev;
-    Account *acct = gnc_ledger_display_leader (gsr->ledger);
-    xaccAccountSetSortReversed(acct, rev);
```
jralls, line 2152: Should the calls to `xaccAccount[GS]etSortReversed` in `gsr_move_sort_and_filter_to_state_file` be similarly replaced? What about `xaccAccount[GS]etSortOrder` in that function?

That's the only use for either of those functions, and the corresponding QOF properties are similarly unused so ISTM they should be removed in favor of split register state.

That's out of scope for this PR and should probably go in master.

jralls:
Thanks!

PR #1551, opened 2023-03-04
[import-main-matcher] allow force editing non-identical desc/notes/memo

christopherlam:
Previously, only identical Desc/Notes/Memo were unlocked for editing. This commit allows editing selected transactions' textual data, even if they are not identical.

Mainly for banks like mine who add unique garbage in the Descriptions:

```
"SHOPRITE - Visa Purchase - Receipt 12345678In MELBOURNE Date 27 Feb 2016 Card 462263xxxxxx3981"
"SHOPRITE - Visa Purchase - Receipt 12345679In MELBOURNE Date 28 Feb 2016 Card 462263xxxxxx3981"
etc
```

### DIFF
--- a/gnucash/import-export/import-main-matcher.c
+++ b/gnucash/import-export/import-main-matcher.c
@@ -942,17 +942,43 @@ match_func (GtkEntryCompletion *completion, const char *entry_str,
     return ret;
 }
 
-static void
-setup_entry (GtkWidget *entry, bool sensitive, GHashTable *hash,
-             const char *initial)
+typedef struct
 {
+    GtkWidget *entry;
+    GObject *override;
+    bool *can_edit;
+    GHashTable *hash;
+    const char *initial;
+} EntryInfo;
+
+static void override_clicked (GtkWidget *widget, EntryInfo *entryinfo)
+{
+    gtk_widget_set_visible (GTK_WIDGET (entryinfo->override), false);
+    gtk_widget_set_sensitive (entryinfo->entry, true);
+    gtk_entry_set_text (GTK_ENTRY (entryinfo->entry), "");
+    gtk_widget_grab_focus (entryinfo->entry);
+    *entryinfo->can_edit = true;
+}
+
+static void
+setup_entry (EntryInfo *entryinfo)
+{
+    bool sensitive = *entryinfo->can_edit;
+    GtkWidget *entry = entryinfo->entry;
+    GtkWidget *override = GTK_WIDGET (entryinfo->override);
+    GHashTable *hash = entryinfo->hash;
+    const char *initial = entryinfo->initial;
+
     gtk_widget_set_sensitive (entry, sensitive);
+    gtk_widget_set_visible (override, !sensitive);
+
     if (sensitive && initial && *initial)
         gtk_entry_set_text (GTK_ENTRY (entry), initial);
     else if (!sensitive)
     {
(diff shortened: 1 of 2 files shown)


gnucash/import-export/import-main-matcher.c
```diff
@@ -925,17 +925,49 @@ match_func (GtkEntryCompletion *completion, const char *entry_str,
...
+        OverrideData *override_data = g_new (OverrideData, 1);
+        override_data->parent = info->main_widget;
+        override_data->override_edit = override_edit;
+        override_data->entry = entry;
+        override_data->sensitive = sensitive_ptr;
         gtk_entry_set_text (GTK_ENTRY (entry), _("Disabled"));
-        return;
+        g_signal_connect (override_edit, "clicked", G_CALLBACK (override_edit_clicked),
+                          override_data);
+        info->new_strings = g_list_prepend (info->new_strings, override_data);
```
christopherlam, line 970: `info->new_strings` is being reused to free OverrideData structs when import matcher completes. It could be renamed to `info->data_to_gfree`.

christopherlam:
Here's screenshot after 2nd commit. Desc&Memo were non-identical but can be unlocked for editing with `Edit` button.

image

christopherlam:
> @christopherlam Do not use stock buttons, they are depreciated and I am inclined to just use straight forward buttons as I do not think there are other places using images on buttons.
Ok changed

> Would you normally just type some text in or would you have copied the text from one of the entries ? Just wondering if there is a suitable way of showing the entries so the text can be copied.

The GtkEntry still has GtkEntryCompletion populated with the _existing_ transactions' data. I tried to populate it to select the most frequent strings within the current selected rows, but this proved too difficult, and would have a very marginal benefit.

christopherlam:
I've refactored this branch a bit -- instead of `setup_entry` having a large number of arguments, it gets a struct on the stack (instead of heap) containing relevant data. The struct is then reused in the `override_edit_clicked` callback. I think this is a bit unusual but works well.

jralls:
> I think this is a bit unusual but works well.

Well, it's the normal fix for too many arguments though 4 is a little low for "too many". It's also unusual to copy the struct members into locals.

Why are you passing `can_edit` as a `bool*`?

christopherlam:
> Well, it's the normal fix for too many arguments though 4 is a little low for "too many". It's also unusual to copy the struct members into locals.

Copying into locals ensures the majority of `setup_entry` is unchanged. Otherwise there'll be lots of `entryinfo->*` in it.

> Why are you passing `can_edit` as a `bool*`?

Because the original bools eg `info->can_edit_desc` will need to be flipped if the button is clicked to perform the change - see line 1155

jralls:
> Otherwise there'll be lots of entryinfo->* in it.

So?

> Because the original bools eg info->can_edit_desc will need to be flipped if the button is clicked to perform the change - see line 1155

Not in `setup_entry`: You immediately dereference `info->can_edit` at line 966 and never touch it again.

christopherlam:
> Not in `setup_entry`: You immediately dereference `info->can_edit` at line 966 and never touch it again.

Yes but line 981 it gets used as callback for the override button in `override_clicked` to flip the bool in 960.

jralls:
> Yes but line 981 it gets used as callback for the override button in override_clicked to flip the bool in 960.

OIC. That's really ugly, but I guess that's on whoever designed this mess.

PR #1553, opened 2023-03-05
Bug 798768 - Improve docs for gnucash.trace location

salomvary:
Used glib documentation as source of truth:
https://docs.gtk.org/glib/func.get_tmp_dir.html

Note that it is still not 100% accurate buecause on Windows the `TEMP` variable is used.

See also https://bugs.gnucash.org/show_bug.cgi?id=798768


jralls:
Don't change translated strings in po files, even when they're partly in English.

jralls:
Since you haven't made the requested changes I did it myself.

salomvary:
Thanks! Noted that I should be faster next time :)

PR #1554, opened 2023-03-06
Speedup Import matcher completion

christopherlam:
Speed up import matcher creating transactions by wrapping relevant accounts with BeginEdit and CommitEdit pair. EDIT: this branch should depend on #1586


gnucash/import-export/import-main-matcher.c
```diff
@@ -553,9 +565,18 @@ on_matcher_ok_clicked (GtkButton *button, GNCImportMainMatcher *info)
         if (first_tran)
         {
             Split* first_split = gnc_import_TransInfo_get_fsplit (trans_info);
-            xaccAccountSetAppendText (xaccSplitGetAccount(first_split), append_text);
+            Account* account = xaccSplitGetAccount (first_split);
+            xaccAccountSetAppendText (account, append_text);
             first_tran = false;
+            begin_edit (&accounts_modified, account);
```
gjanssens, line 571: Tangential to your change - this test for `first_tran` assumes an import is always into a single account. This is true for typical bank imports, but is it no longer true for generic csv imports. It should do this for all source accounts we're importing into.

As said, not directly related to your PR, I just noticed it while reviewing.

gnucash/import-export/import-main-matcher.c
```diff
@@ -553,9 +565,18 @@ on_matcher_ok_clicked (GtkButton *button, GNCImportMainMatcher *info)
...
         }
+
+        if (gnc_import_TransInfo_get_action (trans_info) == GNCImport_ADD)
```
gjanssens, line 574: Why do you only do this for added transactions and not for updated transactions ? Is there no change to the account while updating ? Amounts may be updated ?

gnucash/import-export/import-main-matcher.c
```diff
@@ -570,6 +591,9 @@ on_matcher_ok_clicked (GtkButton *button, GNCImportMainMatcher *info)
     }
     while (gtk_tree_model_iter_next (model, &iter));
 
+    DEBUG ("commit_edit");
+    g_list_free_full (accounts_modified, (GDestroyNotify)xaccAccountCommitEdit);
```
gjanssens, line 595: Nice trick :wink:

gnucash/import-export/import-main-matcher.c
christopherlam: Yes, I'm not too familiar with multi-account csv imports, hence I'd sought your expertise. I agree we shouldn't limit to first split. Would it be sufficient to analyze all import transactions->splits->account?

gnucash/import-export/import-main-matcher.c
christopherlam: > It looks technically ok. I presume you actually tested the performance improvement is some way ?

Anecdotally yes seems faster; the real pain point was importing 6months of CSV in Windows.

gnucash/import-export/import-main-matcher.c
```diff
@@ -553,9 +565,18 @@ on_matcher_ok_clicked (GtkButton *button, GNCImportMainMatcher *info)
...
         }
+
+        if (gnc_import_TransInfo_get_action (trans_info) == GNCImport_ADD)
+        {
+            Account *acc = gnc_import_TransInfo_get_destacc (trans_info);
+            begin_edit (&accounts_modified, acc);
+        }
+
         // Note: if there's only 1 split (unbalanced) one will be created with the unbalanced account,
         // and for that account the defer balance will not be set. So things will be slow.
```
jralls, line 581: Is this comment still true?

gnucash/import-export/import-main-matcher.c
```diff
@@ -553,9 +565,18 @@ on_matcher_ok_clicked (GtkButton *button, GNCImportMainMatcher *info)
         if (first_tran)
         {
             Split* first_split = gnc_import_TransInfo_get_fsplit (trans_info);
-            xaccAccountSetAppendText (xaccSplitGetAccount(first_split), append_text);
+            Account* account = xaccSplitGetAccount (first_split);
+            xaccAccountSetAppendText (account, append_text);
```
jralls: `xaccAccountSetAppendText` calls `xaccAccountBeginEdit` and `xaccAccountCommitEdit` so might as well be after the `begin_edit` call to stop those being processed until the list destruction.

jralls (changes requested):
Absent profiling data the case for this is a bit weak.

gnucash/import-export/import-main-matcher.c
jralls: [Profiling](https://wiki.gnome.org/Apps/Sysprof) data would be better than "anecdotally seems faster". Even wall-clock timing the same import with and without the change would be better than "seems faster".

christopherlam:
With BeginEdit/CommitEdit importing a 925k CSV with 3385 lines on my underpowered laptop.

```
* 18:01:33  WARN <gnc.engine> [gnc_option_db_destroy()] Direct Destroy called on GncOptionDB 93884251806080
* 18:03:45 ERROR <Gtk> gtk_assistant_set_page_has_padding: assertion 'child != NULL' failed
* 18:06:01 ERROR <Gtk> gtk_assistant_set_page_has_padding: assertion 'child != NULL' failed
* 19:58:36  WARN <gnc.import.main-matcher> [on_matcher_ok_clicked()] time =     26.63
```

Without BeginEdit/CommitEdit

```
* 20:00:20 ERROR <Gtk> gtk_assistant_set_page_has_padding: assertion 'child != NULL' failed
* 20:10:25  WARN <gnc.import.main-matcher> [on_matcher_ok_clicked()] time =     76.06
```

jralls:
> Without BeginEdit/CommitEdit

I trust that's with/without `begin_edit` and the list destructor calling `CommitEdit`!

christopherlam:
Here's patch for timing:

```patch
@@ -33,6 +33,7 @@
     @author Copyright (c) 2006 David Hampton <hampton@employees.org>
 */
 #include <config.h>
+#include <time.h>
 
 #include <gtk/gtk.h>
 #include <glib/gi18n.h>
@@ -520,6 +521,7 @@ static void begin_edit (GList **accounts_modified, Account *acc)
         return;
 
     DEBUG ("begin_edit acc %s", xaccAccountGetName (acc));
+    return;
     *accounts_modified = g_list_prepend (*accounts_modified, acc);
     xaccAccountBeginEdit (acc);
 }
@@ -545,6 +547,9 @@ on_matcher_ok_clicked (GtkButton *button, GNCImportMainMatcher *info)
     /* Don't run any queries and/or split sorts while processing the matcher
     results. */
     gnc_suspend_gui_refresh ();
+
+    clock_t start = clock();
+
     bool first_tran = true;
     bool append_text = gtk_toggle_button_get_active ((GtkToggleButton*) info->append_text);
     do
@@ -591,8 +596,12 @@ on_matcher_ok_clicked (GtkButton *button, GNCImportMainMatcher *info)
     }
     while (gtk_tree_model_iter_next (model, &iter));
 
+    clock_t end = clock();
+    double cpu_time_used = ((double) (end - start)) / CLOCKS_PER_SEC;
+
+    PWARN ("time = %9.2f", cpu_time_used);
     DEBUG ("commit_edit");
-    g_list_free_full (accounts_modified, (GDestroyNotify)xaccAccountCommitEdit);
+    /* g_list_free_full (accounts_modified, (GDestroyNotify)xaccAccountCommitEdit); */
 
     gnc_gen_trans_list_delete (info);
```

christopherlam:
merged in as 2b4f0b3b8c

PR #1555, opened 2023-03-06
Add blank transaction at the top - the last piece for 5.1

Bob-IT:
This adds the blank transaction at the top of the register when in reverse sort order which I think makes sense and there may be a bug for it.

As part of this, the toolbar icon and keyboard short cut did not seem appropriate so have changed them to 'go-jump' and 'CNTRL-B', this can be changed if there is a better combination. This would then free up 'CNTRL+Page_Down' for tab navigation which I think there is also a bug report for but have not looked if other changes are needed for this.

Unfortunately I did not realise that the tool tip was so specific so that needed changing and subsequently found the preference 'Future transactions after blank transaction' would also need changing, might need some better tool tip text for that one.

I have tested this and all seems to work OK but would appreciate if some one else could give it a try.

So what to do...
This introduces another copy of the same block of code for the blank transaction, I will try and refactor that and maybe create a structure to pass all the function parameters.

The string changes will need to wait till after the freeze but could the rest be applied?

### DIFF
--- a/gnucash/gtkbuilder/dialog-preferences.glade
+++ b/gnucash/gtkbuilder/dialog-preferences.glade
@@ -2762,13 +2762,13 @@ many months before the current month</property>
                 </child>
                 <child>
                   <object class="GtkCheckButton" id="pref/general.register/future-after-blank-transaction">
-                    <property name="label" translatable="yes">_Future transactions after blank transaction</property>
+                    <property name="label" translatable="yes">_Placement of future transactions</property>
                     <property name="visible">True</property>
                     <property name="can-focus">True</property>
                     <property name="receives-default">False</property>
                     <property name="has-tooltip">True</property>
-                    <property name="tooltip-markup">If checked, transactions with a date in the future will be displayed at the bottom of the register after the blank transaction. If clear, the blank transaction will be at the bottom of the register after all transactions.</property>
-                    <property name="tooltip-text" translatable="yes">If checked, transactions with a date in the future will be displayed at the bottom of the register after the blank transaction. If clear, the blank transaction will be at the bottom of the register after all transactions.</property>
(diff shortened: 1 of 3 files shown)


jralls:
> The string changes will need to wait till after the freeze but could the rest be applied?

Let's hold off the whole thing for 5.1. ISTM having the old tooltips with the blank transaction at or near the top would be confusing.

gjanssens:
> This adds the blank transaction at the top of the register when in reverse sort order which I think makes sense and there may be a bug for it.
> 

https://bugs.gnucash.org/show_bug.cgi?id=797639 ?

Bob-IT:
Would it be worth while changing the keyboard short cut and toolbar image now for 5.0 so they are changed mid series ?

Bob-IT:
I have changed the short cut and the toolbar icons. The rest will be changed for 5.1

jralls:
Thanks.

jralls:
@Bob-IT please fix the conflict and merge when you have a moment.

Bob-IT:
> @Bob-IT please fix the conflict and merge when you have a moment.

Will do, should be next week.

Bob-IT:
Conflict fixed, just looking at a rouge blue line in reverse sort.

jralls:
Is that maybe a *rogue* blue line? Do you have https://github.com/Gnucash/gnucash/commit/de9c0eb5408ed1f90fd36abc4f1d2d4a7bc4af0a in your history?

Bob-IT:
@jralls 
> Is that maybe a _rogue_ blue line? Do you have [de9c0eb](https://github.com/Gnucash/gnucash/commit/de9c0eb5408ed1f90fd36abc4f1d2d4a7bc4af0a) in your history?

Ah spelling, yes I do. I get the blue line after adding the blank transaction to the top of the register. The last commit is what I need to do to hide it when appropriate so it matches the normal order. I can not see a better way unless you can.

I have tried to test this in single, double, basic. auto, ledger, 'future after blank' and all seem to be OK but would appreciate some one else trying it.

I will squash the last commit to the first before pushing.

PR #1558, opened 2023-03-07
Bug798773 - Mac key combination 'command-H' launches help

Bob-IT:
Pressing 'command-H' should hide the application on a Mac but currently it also launches a browser with the 'tutorial and concepts guide'.

The first commit blocks this and the second commit fixes missing shortcut key menu entries.
@jralls Is this OK?


jralls:
@Bob-IT Well, it works, but I don't think it's quite right: All we really want to do is switch the accelerator for opening the tutorial from `<primary>h` to the mac-standard `<primary>?`.

Bob-IT:
Changed to finding the help menu item, disconnect the existing accelerator key and add a new one for '?'

gjanssens:
I would think just changing osx_accel_map in gnucash/ui/ would solve this without code changes ?

Bob-IT:
> I would think just changing osx_accel_map in gnucash/ui/ would solve this without code changes ?
> Or at least the conflicting accelerator. Perhaps the code for loading is still needed ?

May be, I did not know about it, I have never setup / changed any short cuts. I suppose there is an equivalent Linux file that I could test. The code to load and save is still there but maybe it is trying to load before the menu's are setup.

Bob-IT:
I had a look at my live version and see there is an `accelerator-map` file full of commented out `Action` lines. Obviously they are all wrong for the current version but pointed me in a direction.

I think this comes down to the way we are using `GMenuModel` to build the menus. I have been able to dump some of the actions out in the format of the `accelerator-map` file and paste them in as a start point. From here they can be loaded to the accelerator map which I can walk and update the `GMenuModel` with the new accelerators.

I am not sure this is the correct solution but seems to work. I need to tidy up and refactor before I post here.

One thing I do not understand is why the map is saved on exit, the keys can not be changed when the program is running right?

jralls:
> One thing I do not understand is why the map is saved on exit, the keys can not be changed when the program is running right?

In Gtk2 it was possible for users to change accelerators from the menus by some key-mouse combination that I've forgotten. I think that was removed in Gtk3 in favor of changing things through the Gnome control panel app--a facility that isn't available on macOS.

That said, saving the map on exit does allow the file in GNC_USER_DATA to have the current accelerators for the user to edit, they just have to run the current version of GnuCash first. Any obsolete items (at this point possibly the whole thing) from the osx_accel_map in git will be ignored, but I can copy my GNC_USER_DATA/accelerator_map in to create a new one if it would be helpful.

But even though it's not marked deprecated anywhere (nor is GtkMenu/GtkMenuItem) it's not in Gtk4: It appears to me that the new classes are [GAction](https://docs.gtk.org/gio/iface.Action.html), [GActionGroup](https://docs.gtk.org/gio/iface.ActionGroup.html) and [GMenuModel](https://docs.gtk.org/gio/class.MenuModel.html). Accelerators seem to be implemented by top-level functions in Gtk, e.g. [`accelerator_parse`](https://docs.gtk.org/gtk4/func.accelerator_parse.html) but I haven't studied this stuff enough to understand how it all works.

jralls:
Following up on my previous, an accelerator_map file from GnuCash 4.902 has only 4 entries:
```
(gtk_accel_path "<Actions>/GncPluginPageRegisterActions/DeleteTransactionAction" "<Primary><Control>d")
(gtk_accel_path "<Actions>/GncPluginPageRegisterActions/CutTransactionAction" "<Primary><Control>x")
(gtk_accel_path "<Actions>/GncPluginPageRegisterActions/PasteTransactionAction" "<Primary><Control>v")
(gtk_accel_path "<Actions>/GncPluginPageRegisterActions/CopyTransactionAction" "<Primary><Control>c")
```

Bob-IT:
This is what I have so far...

I have commented out saving of the gtk accelerator map as it will over write the accelerator-map file with only the ones loaded, will need to remove if appropiate.
I have used the last commit just to extract the actions and allow me to get I think a complete list to populate a default `accelerator-map` file for Linux and then looked at the `osx_accel_map` and made those changes to create an `accelerator-map-osx` file. I chose a different name just so that I did not over write it.

I have not done the cmake changes, may need a hand on that to handle the location of the default maps and then copy it to the correct location to make edits on. Maybe the copy part can be part of the gui init.

As I said, the last commit was used to create a list of unique actions which I then sorted to create the file and this commit will not be added. I suppose that could be achieved by looking at the ui files but maybe for another time. This shouldn't be needed much as small changes could just be made to the default files. There maybe some actions I need to remove after another read.

Bob-IT:
Hopefully this is now close to being done. I have added my two default files for cmake and removed a few entries.
Have changed the `gnc_gui_int` to look in the users data directory for the `accelerator-map` file, if not there it will copy the default one and open it. Saving the `accelerator-map` is not used.

Todo...
Test on Windows and Macs.
Some bits are commented out, need to remove.
Keep default file names or change the osx one back.

Bob-IT:
@jralls
I think I have made all the changes for this, tested on my LInux VM and also on Windows and I am able to change the key maps.
If OK I can push this.

jralls:
@Bob-IT unfortunately the guide still opens when one uses cmd-h to hide GnuCash. 

It also looks to me that most of the overrides in accelerator-map-osx aren't necessary anymore. `<meta>z` and `<primary>z` are equivalent on Macs. The whole munged accelerator file was a work around I implemented before Mitch Natterer invented the `<primary>` keyword to deal with macOS's use of the command key for common accelerators that use the control key on traditional Unixes and on Windows. Gdk maps both `<meta>` and `<primary>` to the command key on macOS.

Worse, if I try to set the accelerator in GNC_CONFIG_DIR/accelerator-map to anything, e.g.
```
 (gtk_accel_path "<Actions>/MainWindowActions/HelpAction" "<Primary>g")
```
the help menu doesn't even display and though it does suppress displaying of the guide with cmd-h cmd-g doesn't do anything.


On master doing that has no effect at all: The help menu displays and cmd-h both hides the window and displays the guide index.

Bob-IT:
@jralls 
I think the problem here is that you have an existing `accelerator-map` file so will not be replaced with the new default one.
The map entries should look like...
`(gtk_accel_path "<Actions>/mainwin/HelpTutorialAction" "<Meta>?")`, also `HelpAction` is the top level menu action which I have removed in the default map file.

I am not sure how one would communicate this or may be this should happen...
Test for `GNC_CONFIG_DIR/accelerator-map-old,` if not present copy/rename existing `accelerator-map` file to it.
Then copy in the new default `accelerator-map` file.

Another option would be to rename all `mainwin` file entries to `MainWindowActions`, 115 occurrences + the map files.

I suppose `recwin` would also change to `ReconcileWindowActions` to be consistent.

Also does this...

> It also looks to me that most of the overrides in accelerator-map-osx aren't necessary anymore. `<meta>z` and `<primary>z` are equivalent on Macs. The whole munged accelerator file was a work around I implemented before Mitch Natterer invented the `<primary>` keyword to deal with macOS's use of the command key for common accelerators that use the control key on traditional Unixes and on Windows. Gdk maps both `<meta>` and `<primary>` to the command key on macOS.

mean we only need one default `accelerator-map` file

jralls:
OK. I removed GNC_CONFIG_DIR/accelerator_map and ran GnuCash again. The newly installed accelerator map has 32 uncommented lines with the final one being 
```
(gtk_accel_path "<Actions>/mainwin/HelpTutorialAction" "<Meta>?")
```

It doesn't work, nor do `<Meta><Shift>?` nor `<Meta><Shift>/` but `<Meta><Shift>G` does, so <cmd>? must be like <cmd>q: macOS has it hardwired to something so it's not passed through to the application.

So let's make it `<Meta><Shift>G` and merge.

dbreiser:
> So let's make it `<Meta><Shift>G` and merge.

In 4.13, from my ~/Library/Application Support/Gnucash/accelerator-map, I'm showing:

(gtk_accel_path "<Actions>/MainWindowActions/HelpTutorialAction" "<Primary><Control>h")

Is there a reason not to use that?

dbreiser:
> So let's make it `<Meta><Shift>G` and merge.

In 4.13, my ~/Library/Application Support/Gnucash/accelerator-map shows:

(gtk_accel_path "\<Actions>/MainWindowActions/HelpTutorialAction" "\<Primary>\<Control>h")


Is there a reason not to use that for help?

Bob-IT:
Pushed and merged and changed the accelerator to "<Primary><Control>h"

PR #1560, opened 2023-03-07
Remove redundant uses of cmake -E env

richardcohen:
- In GncAddTest, set_tests_properties() is already setting the env
- In the other uses, there is no change to the environment


jralls:
Thanks!

PR #1563, opened 2023-03-08
Push -Wno-error=deprecated-declarations from C*_FLAGS down to the target

richardcohen:
- Apple Xcode deprecates SecKeychain* and mktemp
- ~~Leave mingw - it uses guile2~~
- mingw no longer needs it, according to the build logs at
  https://code.gnucash.org/builds/win32/build-logs/master

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -591,13 +591,11 @@ set(CMAKE_C_EXTENSIONS ON)
 
 if (UNIX)
   set( CMAKE_C_FLAGS "-Werror -Wall -Wmissing-prototypes -Wmissing-declarations ${CMAKE_C_FLAGS}")
-  set( CMAKE_C_FLAGS "-Wno-error=deprecated-declarations ${CMAKE_C_FLAGS}")
   set( CMAKE_CXX_FLAGS "-Werror -Wall -Wmissing-declarations ${CMAKE_CXX_FLAGS}")
-  set( CMAKE_CXX_FLAGS "-Wno-error=deprecated-declarations ${CMAKE_CXX_FLAGS}")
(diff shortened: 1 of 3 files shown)


jralls:
I'd prefer to guard `mktemp` with `#pragma GCC diagnostic ignored "-Wdeprecated-declarations" and to replace the very obsolete `SecKeychain` calls with  `SecItemAdd` and `SecItemCopyMatching`. The new functions have been available since Mac OS X 10.6 and the obsolete ones were deprecated in 10.10. We require 10.13 as a minimum because that's what supports C++17.

There are, BTW, only two Gdk deprecations, `gdk_screen_width` and `gdk_screen_height`, both in gnc-main-window.c.  We should fix those too and force any deprecation besides `mktemp` to error out, including GLib and Gtk/Gdk ones.

jralls:
Followup on GLib & Gtk: I turned up the versions to 2.72.0 and 3.24.30 respectively and found 
just one more, [`g_binding_get_target`](https://docs.gtk.org/gobject/method.Binding.get_target.html) in gnc-tree-view.c, deprecated in glib-2.68 and replaced in that version with `g_binding_dup_target` for better thread safety.

richardcohen:
https://github.com/Gnucash/gnucash/pull/1565 is for `g_binding_get_target`

I can do `gdk_screen_width` and `gdk_screen_height`.

Someone else will have to do the Mac SecKeychain changes.

jralls:
> Someone else will have to do the Mac SecKeychain changes.

I'll do those, but I won't be able to get to it until next week. If you want to move this along you can wrap them with the `diagnostic ignored` pragma in the meantime.

jralls:
Thanks!

PR #1564, opened 2023-03-08
Guard against some possible null pointer dereferences

richardcohen:
Found by gcc 12 when compiling with "-O2 -Wnull-dereference"

### DIFF
--- a/gnucash/gnome-search/search-core-type.c
+++ b/gnucash/gnome-search/search-core-type.c
@@ -230,7 +230,9 @@ editable_enters (GNCSearchCoreType *fe)
 void
 gnc_search_core_register_type (const char *type_name, GNCSearchCoreNew fcn)
 {
-    g_return_if_fail (type_name || *type_name || fcn);
+    g_return_if_fail (type_name);
+    g_return_if_fail (*type_name);
+    g_return_if_fail (fcn);
     g_return_if_fail (typeTable);
 
     g_hash_table_insert (typeTable, (char *) type_name, (gpointer) fcn);
--- a/gnucash/gnome/gnc-plugin-page-owner-tree.c
+++ b/gnucash/gnome/gnc-plugin-page-owner-tree.c
@@ -932,7 +932,11 @@ static int build_owner_report (GncOwner *owner, Account *acc)
         args = scm_cons (SCM_BOOL_F, args);
     }
 
-    arg = SWIG_NewPointerObj(owner, SWIG_TypeQuery("_p__gncOwner"), 0);
+    swig_type_info * qtype = SWIG_TypeQuery("_p__gncOwner");
+    g_return_val_if_fail (qtype, -1);
+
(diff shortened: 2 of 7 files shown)


gnucash/gnome-search/search-core-type.c
```diff
@@ -230,7 +230,9 @@ editable_enters (GNCSearchCoreType *fe)
 void
 gnc_search_core_register_type (const char *type_name, GNCSearchCoreNew fcn)
 {
-    g_return_if_fail (type_name || *type_name || fcn);
+    g_return_if_fail (type_name);
+    g_return_if_fail (*type_name);
+    g_return_if_fail (fcn);
```
jralls, line 235: Just change the `||` to `&&`.

gnucash/gnome/gnc-plugin-page-owner-tree.c
```diff
@@ -932,7 +932,11 @@ static int build_owner_report (GncOwner *owner, Account *acc)
         args = scm_cons (SCM_BOOL_F, args);
     }
 
-    arg = SWIG_NewPointerObj(owner, SWIG_TypeQuery("_p__gncOwner"), 0);
+    swig_type_info * qtype = SWIG_TypeQuery("_p__gncOwner");
+    g_return_val_if_fail (qtype, -1);
+
+    arg = SWIG_NewPointerObj(owner, qtype, 0);
+
```
jralls, line 939: `_p__gncOwner` is a `#define` in swig-engine.cpp (a generated file) so the only way that `SWIG_TypeQuery` could fail would be a link failure at compile time.

jralls (commented):
And a couple of impossible ones.

gnucash/gnome-search/search-core-type.c
richardcohen: Yes, that was my first instinct, but it's annoyingly inconsistent with the next line.

gnucash/gnome/gnc-plugin-page-owner-tree.c
jralls: > SWIG_TypeQuery() is looking for the type at runtime, not compile time.
You misunderstood: While SWIG_TypeQuery() is indeed a runtime check, it's checking a compile-time constant so as long as the correct swig library is linked in *it will always succeed* and if the correct swig library isn't linked in *the build will fail*. 

> I would prefer to keep it consistent with the code above, which does a similar type query for _p_Account.

If there's a check on `_p__Account` then remove it as unnecessary. That will still satisfy your preference for consistency.

gnucash/gnome-search/search-core-type.c
jralls: Not really, but maybe it's because you don't know that `g_return_if_fail` also logs an error. This one ensures that the function args precondition is met, the other one ensures that the static variable `typeTable` has been initialized. Two quite different preconditions that deserve separate error messages.

PR #1565, opened 2023-03-10
Replace deprecated g_binding_get_target, and increase glib version to 2.68

richardcohen:


jralls:
Unfortunately Debian Stable is still on 2.66 so I'm afraid we'll have to have a HAVE_GLIB_2_68 and conditionally use `g_binding_get_target` or `g_binding_dup_target`.

richardcohen:
In that case, this change can wait; it offers no particular advantages, and setting glib min !=glib max risks having other glib functions sneaking in.

Debian 12 with glib 2.74 will be released as stable in a few months.

christopherlam:
@richardcohen nice set of spring cleaning. We want to remove the WITH_SQL option and make it compulsory. This also means removing the non SQL dependent functions in code. Would you want to take up the task?

richardcohen:
> @richardcohen nice set of spring cleaning. We want to remove the WITH_SQL option and make it compulsory. This also means removing the non SQL dependent functions in code. Would you want to take up the task?

@christopherlam I am not entirely clear what the task is? Removing WITH_SQL is easy, so I suspect there is a lot of work lurking behind "removing the non SQL dependent function".

Is this the same as "Migrate from QofQuery to SQL Query"?

I am cautious about forcing a dependency on the unmaintained (if stable) libdbi. Perhaps it would be better to depend directly on sqllite. Perhaps create our own database facade?

richardcohen:
> That would make more sense, but there's a prerequisite: First we have to have a new XML backend that creates a SQL database to query.
> 

OK. I can take a look at that.

> ... We'll want to use stored procedures for common queries--like populating registers--and libdbi doesn't support that.

sqlite doesn't support stored procedures, either. I would imagine that prepared SQL would be fine.

> 9 years is a long time in software so that evaluation would need to be done again ...

odb still looks the most attractive -- it is maintained, and is in Debian.

jralls:
Thanks. Sorry this took so long to merge.

PR #1571, opened 2023-03-13
Fix crash with --log gnc.gui=debug

richardcohen:
gnc_main_window_update_one_menu_action() will try to print data.action_name


jralls:
Thanks!

PR #1576, opened 2023-03-14
File/Open should default to *.gnucash files only

christopherlam:
Why should File/Open show all `*.*` files initially? Limit to `*.gnucash` and exclude `*.gnucash.*.gnucash` files

### DIFF
--- a/gnucash/gnome-utils/dialog-file-access.c
+++ b/gnucash/gnome-utils/dialog-file-access.c
@@ -22,6 +22,7 @@
  * Boston, MA  02110-1301,  USA       gnu@gnu.org                   *
 \********************************************************************/
 
+#include <stdbool.h>
 #include <config.h>
 
 #include <gtk/gtk.h>
@@ -33,6 +34,7 @@
 #include "dialog-utils.h"
 #include "dialog-file-access.h"
 #include "gnc-file.h"
+#include "gnc-filepath-utils.h"
 #include "gnc-plugin-file-history.h"
 #include "gnc-session.h"
 
@@ -243,6 +245,15 @@ get_default_database( void )
     return default_db;
 }
 
+typedef bool (*CharToBool)(const char*);
+
+static bool datafile_filter (const GtkFileFilterInfo* filter_info,
+                             CharToBool filename_checker)
+{
+    return filter_info && filter_info->filename &&
+        filename_checker (filter_info->filename);
+}
+
 static void free_file_access_window (FileAccessWindow *faw)
 {
     g_free (faw->starting_dir);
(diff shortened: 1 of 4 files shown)


jralls:
> Why should File/Open show all *.* files initially?

Because not all gnucash files have a `.gnucash` suffix. Plus sometimes users need to load an older file.

Not that having the filters wouldn't be helpful most of the time, just that there needs to be tick-boxes to turn them off.

christopherlam:
there's always the `*.*` option.

gjanssens:
I think this is a good improvement, except the commit message is misleading: the code as you have written it will also list backup files, not only main data files. I think that's how it should be. If you really don't want that, it would be nice to have two filters, one for normal data files and one for backup files specifically - backup files are regular gnucash files in contents, and can be opened just like that.

gjanssens:
I see, I didn't read the code carefully enough.

The suggested filter list looks fine to me.

As an implementation suggestion, perhaps it's nicer to convert the file to cpp and use a regex to create the filters?

christopherlam:
Not so easy to create heuristics for such substring search, with or without regex.

gjanssens:
Backups are the easiest to define in regex:

```
".*[.](?:xac|gnucash)[.][0-9]{14}[.](?:xac|gnucash)$"
```

You can check for that one only if the backup filter is requested. If the normal data files filter is requested, you can first check for

```
".*[.](?:xac|gnucash)$"
```

And if that matches, check also if the backup filter does not match. This double test will give you plain data files only.

I don't think you have to worry about case sensitivity. GnuCash has never used capitals for extensions. If it is important, case sensitivity is usually an option you can pass to the regex processor.

gjanssens:
Also non-ascii searching is irrelevant because the bits we are searching for in the regex are all ascii.

gjanssens:
What may be a nice compliment to this PR: detect when a user opens a backup file, issue a dialog box informing the user of this and propose to resave to a new name. What you have implemented above is useful for the Open dialog box, but there are other ways a file can be opened - via the command line or from history.

christopherlam:
Ok using boost::regex -- the backup regex doesn't quite work yet.

jralls:
LGTM.

jralls:
Oh, wait, there is one thing: The translatable strings need translator notes explaining that ".gnucash" and ".xac" are file extensions and shouldn't be translated.

And of course this can't merge until after string freeze.

gjanssens:
I like it even better with std::regex :+1:

christopherlam:
> Oh, wait, there is one thing: The translatable strings need translator notes explaining that ".gnucash" and ".xac" are file extensions and shouldn't be translated.
> 
> And of course this can't merge until after string freeze.

If we want it to slip into the next release pronto we can remove the _ and reinstate it after string freeze 😭

jralls:
> If we want it to slip into the next release pronto we can remove the _ and reinstate it after string freeze 😭

Let's not. A week isn't that long to wait.

christopherlam:
Merged in 'stable'

PR #1577, opened 2023-03-14
Bug 798570 - Totals for the income, expenses and remaining to budget incorrect for one specific date

BrianRater:
https://bugs.gnucash.org/show_bug.cgi?id=798570

In a budget, if you increase the number of periods (ex. from 12 months to 26 fortnights), the code attempts to delete the "totals column" (the last column which shows totals for the entire budget), then add the extra columns and then add the totals column back.  When deleting the totals column in the "totals row" (the bottom rows which show totals for each period), the wrong column was being deleted.  The "totals column" would remain which displays the entire budget totals instead of just the given period, then columns added and a second "totals column" was added.  Calculating the column to delete was not taking into account the 2 hidden by default columns (code and description).

### DIFF
--- a/gnucash/gnome/gnc-budget-view.c
+++ b/gnucash/gnome/gnc-budget-view.c
@@ -1547,11 +1547,13 @@ The function will step through to only display the columns that are set
 void
 gnc_budget_view_refresh (GncBudgetView *budget_view)
 {
-	// Column identifiers
-	const gint GNC_BUDGET_VIEW_CODE_COL = 1;
-	const gint GNC_BUDGET_VIEW_DESC_COL = 2;
-	const gint GNC_BUDGET_VIEW_START_PERIODS_COL = 3;
-	// The Totals column will be after the periods columns.
+    // Column identifiers
+    enum {
+        code_column         = 1,
+        description_column  = 2,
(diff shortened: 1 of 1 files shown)


gnucash/gnome/gnc-budget-view.c
```diff
@@ -1547,6 +1547,12 @@ The function will step through to only display the columns that are set
 void
 gnc_budget_view_refresh (GncBudgetView *budget_view)
 {
+	// Column identifiers
+	const gint GNC_BUDGET_VIEW_CODE_COL = 1;
+	const gint GNC_BUDGET_VIEW_DESC_COL = 2;
+	const gint GNC_BUDGET_VIEW_START_PERIODS_COL = 3;
```
jralls, line 1553: Make this an enum. There's no need to use all caps (which should be reserved for preprocessor symbols) nor to elaborate  the names with `gnc_budget_view`  in function scope.

jralls (changes requested):

gnucash/gnome/gnc-budget-view.c
BrianRater: Done

jralls (approved):

PR #1578, opened 2023-03-15
Bug798558 - Irregular line spacing for securities accounts in CoA

Bob-IT:
Finally got back to looking at this, @jralls this is what I was thinking off.

In addition to fixing bug report, this PR changes the vertical alignment of the row contents to be at the top. This is only noticeable when say the notes field has more than one line. Also as the Account dialog notes field had word wrapping I have added it to the CoA notes cell.

Still need to do some more testing but seems to work.

### DIFF
--- a/libgnucash/app-utils/gnc-ui-util.c
+++ b/libgnucash/app-utils/gnc-ui-util.c
@@ -1486,6 +1486,9 @@ gnc_print_amount_with_bidi_ltr_isolate (gnc_numeric val, GNCPrintAmountInfo info
     static const char ltr_pop_isolate[] = { 0xe2, 0x81, 0xa9 };
     size_t offset = info.use_symbol ? 3 : 0;
 
+    if (!gnc_commodity_is_currency (info.commodity))
+        offset = 0;
+
     memset (buf, 0, BUFLEN);
     if (!xaccSPrintAmount (buf + offset, val, info))
     {
@@ -1493,7 +1496,7 @@ gnc_print_amount_with_bidi_ltr_isolate (gnc_numeric val, GNCPrintAmountInfo info
         return buf;
     };
 
-    if (!info.use_symbol)
+    if (offset == 0)
         return buf;
 
     memcpy (buf, ltr_isolate, 3);


gnucash/gnome-utils/gnc-tree-view-account.c
```diff
@@ -935,12 +949,13 @@ gnc_tree_view_account_new_with_root (Account *root, gboolean show_root)
...
-                                     GNC_TREE_MODEL_ACCOUNT_COL_COLOR_TOTAL_PERIOD,
-                                     GNC_TREE_VIEW_COLUMN_VISIBLE_ALWAYS,
-                                     sort_by_total_period_value);
+    column = gnc_tree_view_add_numeric_column(view, _("Total (Period)"), "total-period",
+                                              SAMPLE_ACCOUNT_VALUE,
+                                              GNC_TREE_MODEL_ACCOUNT_COL_TOTAL_PERIOD,
+                                              GNC_TREE_MODEL_ACCOUNT_COL_COLOR_TOTAL_PERIOD,
+                                              GNC_TREE_VIEW_COLUMN_VISIBLE_ALWAYS,
+                                              sort_by_total_period_value);
+    gnc_tree_view_set_column_cell_renderer_pango_wrap_word (column);
```
jralls, line 958: What does word-wrapping numeric columns do?

Bob-IT: It sets the cell render to wrap on word and also sets the wrap-width to that of the column width. Setting the wrap-width enables the wrapping, setting it to -1 does not fix the problem.

jralls: Yes of course. But what does it do for ***numeric*** columns that don't have words and shouldn't wrap?

Bob-IT: I under stand what you mean now, does not look good, will have a rethink over the weekend, maybe look at changing the original rtl fix.

Bob-IT:
@jralls This change fixes the bug and tested on my Fedora VM when logged in with rtl user.
I have dropped for now the vertical alignment as the expander icon stayed centered, might of missed something there or it may just not be possible. I am sure if the rest was aligned at the top some one would report it.
Also dropped the wrapping of the notes field, that may require a tree view preference to enable wrapping or not. Again I am sure someone will complain.

PR #1579, opened 2023-03-16
Eliminate translation ambiguity 

ljffytax:
Eliminate translation ambiguity of "For"


jralls:
Thanks. I cherry-picked and squashed the two real commits to get rid of the unwanted merge.

PR #1519, opened 2023-01-30
Build with -Wunused

richardcohen:
Builds with gcc 12.2.0 and clang 15.0.6 on ubuntu 22.10

PR #1582, opened 2023-03-18
Build fixes

mkbosmans:
Some small fixes to build without errors. These were mainly found when testing various nonstandard configurations, like RelWithDebInfo for profiling or enabling AddressSanitizer with GCC 11.


gnucash/gnome/dialog-report-column-view.cpp
```diff
@@ -531,6 +531,7 @@ gnc_column_view_edit_size_cb(GtkButton * button, gpointer user_data)
     if (r->contents_list.size() > static_cast<size_t>(r->contents_selected))
     {
         auto [id, wide, high] = r->contents_list[r->contents_selected];
+        (void) id; // unused
```
jralls, line 534: Then remove it or mark it with `[[maybe_unused]]`. Removal is probably better.

jralls (changes requested):
> During development it is desirable not to have to make the code
warning free for every build, as that slows down the testing cycle.

Nope. We set -Werror on those warnings because we don't want anything committed that emits those warnings.

> We do not control the contents of these files directly,
so we don't want to deal with compiler warnings as errors.

Actually we do, they're generated from code in `.i` files and those aren't checked until the resulting SWIG file is compiled.

mkbosmans:
Thanks for the review, will work on a new version of the PR.

> > During development it is desirable not to have to make the code
> > warning free for every build, as that slows down the testing cycle.
> 
> Nope. We set -Werror on those warnings because we don't want anything committed that emits those warnings.

My usual workflow would be to hack in debug builds and then when everything is ready do a final build and test run in release mode. So that would prevent any commit which adds warnings (which I agree would be pretty annoying).

When working on the code it is quite cumbersome to e.g. when commenting out a function call you also have to comment out the whole function if it otherwise becomes unused. I would assume the CI runners would catch any pull request that introduced a build error because the developer only ran without -Werror, right?

jralls:
> I would assume the CI runners would catch any pull request that introduced a build error because the developer only ran without -Werror, right?

We don't have a PR-only policy and core devs can make mistakes too.

jralls:
For your workflow notice that all of those lines end with `${CMAKE_FOO_FLAGS}`, appending the command-line value to the value so you can override the flags on the command line by passing e.g. `-DCMAKE_C_FLAGS="-Wno-error"` to cmake. Just make sure to do a build without that before committing!

richardcohen:
@jralls, -Werror causes problems for downstream users (e.g. old gcc - https://github.com/Gnucash/gnucash/pull/1593, or new python - https://github.com/Gnucash/gnucash/pull/1596).
I would suggest moving -Werror from CMakeLists into the CI and nightly build scripts.

jralls:
@richardcohen 

> @jralls, -Werror causes problems for downstream users (e.g. old gcc - https://github.com/Gnucash/gnucash/pull/1593, or new python - https://github.com/Gnucash/gnucash/pull/1596).
I would suggest moving -Werror from CMakeLists into the CI and nightly build scripts.

Amusing, given that these "problems for downstream users" arise from your #1519 and  #1563.

But [my objection](https://github.com/Gnucash/gnucash/pull/1582#issuecomment-1474967157) stands. I'd rather inconvenience one or two downstream users who want to build on old compilers than slow down core developers by requiring them to wait for CI to complete before they can push to code. Both goals could be satisfied with a change to the dist target that removes `-Werror` from CMakeLists.txt in distribution tarballs or something that depends on `CMAKE_BUILD_TYPE`. Ideally any build type that disables `-Werror` would also disable the test targets.

jralls:
@mkbosmans You refuse to accept that removing `-Werror` for your personal convenience isn't going to happen. Since the rest of this PR is cosmetic, I'm closing it.

PR #1587, opened 2023-03-19
Bug 798712 - Regional Decimal Point not recognised in budgets

BrianRater:
https://bugs.gnucash.org/show_bug.cgi?id=798712

In locales where a comma is used instead of a decimal in currency amounts, such as Portugal, the keypad period was not being properly translated to a comma in the budget view.  When someone would enter "5.23" via the keyboard intending for this to be "5,23" it would instead be interpreted as "523,00".

I copied code from gnc-amount-edit.c which also includes a patch to a known issue with Windows passing keypad events.  I am not running on Windows, so I can't test this part.  This code exists in 3 other places in gnucash, so it would probably be a good candidate for refactoring some day.  I tested this by changing my locale to portugese.  One issue I ran into is that I took an existing file in us dollars, turned on portugese localization, then added a budget.  The totals values would remain zero because it could not convert Euros to USD.  Likely an issue with my installation, but unrelated to this bug.   I created a new book using the Portugese settings and budgeting worked correctly, including totals.


jralls:
Thanks!

jralls:
Rats, I missed 
>[Merge branch 'Gnucash:maint' into 798712](https://github.com/Gnucash/gnucash/pull/1587/commits/36085837cc13fece0afc103ccc7e881e048146ee)

Please don't merge anything into your feature branches, only rebase them on the target branch. You don't usually need to do even that unless Github or your own testing tells you that there are conflicts that need to be resolved in order to merge or you need to squash a bunch of refinement commits.

BrianRater:
Thanks.  I understand.  I’ve just started using git at work a few months
ago and I’m still a novice.

PR #1593, opened 2023-03-28
Possible fix for -Wunused-but-set-variable in gcc 7.5.0

richardcohen:
See https://lists.gnucash.org/pipermail/gnucash-devel/2023-March/046617.html


jralls:
Possible for gcc 7.5, definitely for gcc-8 and 9 that I tested. Not a problem on gcc-10 and later.

Note that [README.dependencies](https://github.com/Gnucash/gnucash/blob/stable/README.dependencies) currently specifies at least gcc-8, but we may bump that to gcc-10 for 5.1.

PR #1594, opened 2023-03-28
Fix clang 15.0.7 build

richardcohen:
Will be in ubuntu 23.04


jralls:
Squashed and changed the commit message. Thanks.

PR #1597, opened 2023-03-30
Fix cmake 3.10 build

richardcohen:
I would have replied on the mailing list, but subscriptions seem to be broken at the moment.

For choosing a minimum version of cmake, see https://cliutils.gitlab.io/modern-cmake/chapters/intro/dodonot.html


jralls:
Wouldn't bumping the minimum Cmake version to 3.11 make more sense?

PR #1599, opened 2023-04-04
std::string cannot free() scm_to_utf8_string automatically

christopherlam:
all these `scm_to_utf8_string` must be freed manually.

### DIFF
--- a/bindings/guile/gnc-optiondb.i
+++ b/bindings/guile/gnc-optiondb.i
@@ -413,23 +413,25 @@ scm_to_value<gnc_commodity*>(SCM new_value)
     if (scm_is_list(new_value))
     {
         auto len{scm_to_uint(scm_length(new_value))};
-        std::string mnemonic{scm_to_utf8_string(scm_list_ref(new_value,
-                                                         scm_from_uint(0)))};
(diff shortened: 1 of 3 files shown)


jralls:
Good.

christopherlam:
The officious approach would have been to use `scm_dynwind_begin(0);` ... `scm_dynwind_free (str);` https://www.gnu.org/software/guile/manual/html_node/Dynamic-Wind.html 😮

jralls:
That would be a starting point. Ideally you'd write a class that does `scm_dynwind_begin` in the constructor and `scm_dynwind_free` in the destructor and implement an `operator()` to provide access to the char* to pass to `gnc_commodity_lookup`. That's the general C++98 way to construct a closure that does [RAII](https://en.wikipedia.org/wiki/Resource_acquisition_is_initialization). `const char* operator()` could be hoisted to `template <typename T> T operator()` with a bunch of `if constexpr (is_same_v<T, foo>)` to call the appropriate SCM conversion to return a pointer or reference to whatever you think the `SCM` should hold.

None of that's new C++, BTW. The original GnuCash code could have done it if the folks developing it back then had wanted to use C++.

PR #1600, opened 2023-04-09
Stable progress

christopherlam:
A few small steps forward. Any comments about the wording of error message?

### DIFF
--- a/gnucash/gnome/gnc-plugin-business.c
+++ b/gnucash/gnome/gnc-plugin-business.c
@@ -43,6 +43,7 @@
 #include "gnc-plugin-page-invoice.h"
 #include "gnc-plugin-page-owner-tree.h"
 #include "gncOwner.h"
+#include "gnc-ui.h"
 #include "gnc-ui-util.h"
 #include "gnc-date.h"
 #include "gnc-file.h"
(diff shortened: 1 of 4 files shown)


jralls:
There's a big difference between unsetting `defer_bal_computation`  and `xaccAccountCommitEdit`: The latter sorts the split list, computes the balance, *and writes out the result to the database*. Since `gnc_gen_trans_list_delete` is a cleanup function called in among other places `on_matcher_cancel_clicked` and `on_matcher_delete_event` that doesn't seem like the right thing to do.

gnucash/gnome/gnc-plugin-business.c
```diff
@@ -715,6 +716,13 @@ gnc_business_assign_payment (GtkWindow *parent,
     if (xaccTransCountSplits(trans) <= 1)
         return;
 
+    if (xaccTransGetFirstAPARAcctSplit (trans, FALSE) == NULL)
+    {
+        gnc_warning_dialog (parent, "%s",
+                            _("There must be an A/P or A/R transfer account."));
+        return;
+    }
+
```
gjanssens, line 725: I don't think this was ever my intention when reworking the Assign As Payment functionality (and this is even against prior advice -- avoid manually creating transactions involving the A/R or A/P accounts).

Why would you prevent such transactions from being considered for being transformed to a business payment ? From my point of view this is safe. If there's no A/R or A/P split, the user can select this in the dialog that will be shown. In that dialog it's the Post to account.

christopherlam: From my testing it seemed that if there's no APAR split then the assign as payment dialog appears but clicking OK then nothing was actioned in the book.

gjanssens: That's how it behaves now indeed and that's the bug IMO :)

I'll be looking into this before the 5.1 release. In my first experiment I see there's a currency conversion error showing up in the logs that shouldn't be there. That's my starting point.

christopherlam: Thanks, let's consider this branch a lost cause then 🙂

PR #1601, opened 2023-04-09
cmake: Try guile and guild executables from guile-N.N.pc first

ArsenArsen:
Hi there!

This patch adjusts the build system to first attempt to use the `guile` and `guild` executables specified in the pkg-config files for Guile, before trying to search the path.

This interacts better with a variety of possible program name transformations (for instance, on my system, Guild 3.0 is installed as guild-3.0).


jralls:
Thanks!

PR #1603, opened 2023-04-12
Check account limits at startup

christopherlam:
1. Accumulate startup warnings into 1 `gnc_info_dialog`
2. Check limits at startup

### DIFF
--- a/gnucash/gnome-utils/gnc-file.c
+++ b/gnucash/gnome-utils/gnc-file.c
@@ -694,6 +694,31 @@ gnc_file_query_save (GtkWindow *parent, gboolean can_cancel)
     return TRUE;
 }
 
+static void
+check_limit (Account *acct, GList **list)
+{
+    char *expl = gnc_ui_account_get_balance_limit_explanation (acct);
+    if (expl)
+        *list = g_list_prepend (*list, expl);
+}
+
+static char*
+collect_balance_limit_warnings (QofBook *book)
+{
+    g_return_val_if_fail (book, NULL);
+    GList *list = NULL;
+    gnc_account_foreach_descendant (gnc_book_get_root_account (book),
+                                    (AccountCb) check_limit, &list);
+    if (!list)
+        return NULL;
+
+    const char *header = N_("Account limit warnings:");
+    list = g_list_sort (list, (GCompareFunc)g_utf8_collate);
+    list = g_list_prepend (list, g_strdup (_(header)));
+    char *rv = gnc_g_list_stringjoin (list, "\n");
+    g_list_free_full (list, g_free);
+    return rv;
+}
 
 
 static char*
@@ -745,6 +770,10 @@ run_post_load_scrubs (GtkWindow *parent, QofBook *book)
     char *sep_warning = get_account_sep_warning (book);
     if (sep_warning)
         infos = g_list_prepend (infos, sep_warning);
+
+    char *limit_warnings = collect_balance_limit_warnings (book);
+    if (limit_warnings)
+        infos = g_list_prepend (infos, limit_warnings);
 
     qof_event_resume();
 


jralls:
Not a great design. I particularly object to `get_balance_limit_info (const Account *account, gboolean icon)` that returns a `const char*` if `icon` is true and an allocated `char*` if it's false. Meanwhile `higher_limit_explanation (const Account *account, GNCPrintAmountInfo pinfo)` and friends duplicate a lot of code, differing only in the condition and the string value.

christopherlam:
The most pressing question is whether there's a better interface to present these results other than as a `gnc_info_dialog` at startup...

jralls:
> The most pressing question is whether there's a better interface to present these results other than as a gnc_info_dialog at startup...

You mean besides adding the column to the accounts page?

I think that the dialog at startup is a bit intrusive and perhaps not all that useful. ISTM that if you're going to pop an info dialog the time to do it is when the user posts a transaction that causes the balance to go out of bounds. We haven't really documented or advertised this new feature much, but I suspect that when it starts getting some real usage users might ask for a warning to pop when they get close to the limit.

christopherlam:
> > The most pressing question is whether there's a better interface to present these results other than as a gnc_info_dialog at startup...
> 
> You mean besides adding the column to the accounts page?
> 
> I think that the dialog at startup is a bit intrusive and perhaps not all that useful. ISTM that if you're going to pop an info dialog the time to do it is when the user posts a transaction that causes the balance to go out of bounds. We haven't really documented or advertised this new feature much, but I suspect that when it starts getting some real usage users might ask for a warning to pop when they get close to the limit.

In my particular use case, I'd set limits to `Imbalance` and other suspense accounts which are waiting to be allocated to the correct account. A `gnc_info_dialog` would seem appropriate. Alternatively, a toolbar icon which only appears if there are such warnings.

jralls:
So you start up GnuCash and it has some accounts holding splits that really should belong to other accounts. and nothing else You have set up limits on those accounts that displays an icon on the accounts page when the balance falls outside those limits, and you naturally have that column displayed on your accounts page because what's the point of setting the limits if you don't also display the icons.

Now at startup you also get a display listing the accounts that have those icons displayed next to them. I guess that could be helpful if the accounts in question are scrolled of the display at the moment, but you'll either dismiss the dialog because you don't want to deal with it right now or you'll have to look at each account on the dialog and find it on the accounts page where it's made obvious by displaying the limits violation symbol. The list of accounts needing attention seems not that useful. A simple dialog that says that there are accounts out of limits would be a lot less work and make more sense.

Toolbars are for buttons that produce actions, not directly provide information. If it's just to provide information then a status bar message seems more appropriate. It could be decorated with an icon to help attract the user's attention. OTOH a toolbar button could be made sensitve when there are accounts that are out of limits and clicking it could take the user to the next such account in the accounts page.

gjanssens:
> Toolbars are for buttons that produce actions, not directly provide information. If it's just to provide information then a status bar message seems more appropriate. It could be decorated with an icon to help attract the user's attention. OTOH a toolbar button could be made sensitve when there are accounts that are out of limits and clicking it could take the user to the next such account in the accounts page.

One of those options is exactly what I would have suggested as well.

Or a third option, which is more work but may be very useful in other contexts as well:
Implement filtering on the accounts page. That is allow the user to set certain criteria to limit the list of displayed accounts.
For this particular use case you can then create a pre-defined filter that only shows accounts close to or over a balance limit and stick that after a toolbar button/menu item.

For other situations, the filtering code could allow the user to filter on arbitrary account related criteria (name, code, type, balance, projected balance,...)

christopherlam:
> > Toolbars are for buttons that produce actions, not directly provide information. If it's just to provide information then a status bar message seems more appropriate. It could be decorated with an icon to help attract the user's attention. OTOH a toolbar button could be made sensitve when there are accounts that are out of limits and clicking it could take the user to the next such account in the accounts page.
> 
> One of those options is exactly what I would have suggested as well.
> 
> Or a third option, which is more work but may be very useful in other contexts as well: Implement filtering on the accounts page. That is allow the user to set certain criteria to limit the list of displayed accounts. For this particular use case you can then create a pre-defined filter that only shows accounts close to or over a balance limit and stick that after a toolbar button/menu item.
> 
> For other situations, the filtering code could allow the user to filter on arbitrary account related criteria (name, code, type, balance, projected balance,...)

I tend to think that an account which exceeds limit is a warning requiring some action, rather than a filter to apply at leisure.

To summarise the following startup warnings each have their own intrusive dialog.
- sx are now due, and can be postponed
- business objects are now due, and can be postponed
- account names have separator, and needs fixing asap
- (one off) budgets have been scrubbed
- (being discussed) accounts are exceeding limits set in properties
- (another one?) future-dated transactions are now no longer future-dated (maybe any future-dated transaction should have its own reconcile-status e.g. "z" to be changed to "n" when the startup routine finds them.

It wouldn't be too difficult to centralize them into a report-page. As you can see I have several disjoint methods to remind myself of important financial tasks :) and want the process to be smoother.

christopherlam:
While this is being discussed, herewith screenshot of exposing explanation in account-tree

image

christopherlam:
The tooltip in main account tree code has been merged into stable.

The `gnc_info_dialog` mechanism to accumulate warnings and errors is also merged into stable. Further refinements are possible. see screenshot showing both budget-fix and accountsep-in-acctname messages at startup.

image

Bob-IT:
@christopherlam I will get to this at some point, just busy with other stuff right now.

luzpaz:
any progress here ?

PR #1605, opened 2023-04-13
Update subdir location for backend libraries

rzerres:
gnucash version 5 branch changed the location of module libraries
inside CMakeFile.txt
Let's adopt to make the linker happy again.

### DIFF
--- a/libgnucash/backend/xml/gnc-xml-backend.cpp
+++ b/libgnucash/backend/xml/gnc-xml-backend.cpp
@@ -297,7 +297,7 @@ GncXmlBackend::load(QofBook* book, QofBackendLoadType loadType)
 void
 GncXmlBackend::sync(QofBook* book)
 {
-        /* We make an important assumption here, that we might want to change
+    /* We make an important assumption here, that we might want to change
      * in the future: when the user says 'save', we really save the one,
      * the only, the current open book, and nothing else. In any case the plans
      * for multiple books have been removed in the meantime and there is just one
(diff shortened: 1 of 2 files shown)


jralls:
@rzerres 
> gnucash version 5 branch changed the location of module libraries inside CMakeFile.txt

Why do you think that (I'm looking for a specific commit here) and what linker is having difficulties?

rzerres:
> I'm looking for a specific commit here

I encountered error messages when compiling in my coowner branch, working in ArchLinux. They disappiered once i appied the PR.

jralls:
> I encountered error messages when compiling in my coowner branch, working in ArchLinux. They disappiered once i appied the PR.

That doesn't answer the question, which is: What specific commit makes you think that the build or install location of libgncmod-backend-dbi.so and libgncmond-backend-xml.so changed?

I ask it because in the relevant parts of libgnucash/backends/dbi/CMakeLists.txt and libgnucash/backends/xml/CMakeLists.txt the library destinations are unchanged since they were first written, and I'm pretty sure that goes back a long way in the autotools builds.


> compiling in my coowner branch

That sounds like a problem of your making in your fork. We use Arch Linux as one of the CI platforms so it's clearly not a problem for GnuCash proper.

rzerres:
> > I encountered error messages when compiling in my coowner branch, working in ArchLinux. They disappiered once i appied the PR.
> 
> That doesn't answer the question, which is: What specific commit makes you think that the build or install location of libgncmod-backend-dbi.so and libgncmond-backend-xml.so changed?
> 
> I ask it because in the relevant parts of libgnucash/backends/dbi/CMakeLists.txt and libgnucash/backends/xml/CMakeLists.txt the library destinations are unchanged since they were first written, and I'm pretty sure that goes back a long way in the autotools builds.
> 
> > compiling in my coowner branch
> 
> That sounds like a problem of your making in your fork. We use Arch Linux as one of the CI platforms so it's clearly not a problem for GnuCash proper.

All right. So please ignore the noise. Just closing the PR.

PR #1607, opened 2023-04-17
Chart Y axis scaling can cause some bar charts to become misleading

ZDBioHazard:
I guess I've been lucky, but all my quarters' cash flow has been positive. (Yay me)
Unfortunately, this causes the "Cash Flow Barchart" and other similar reports to start their Y axes at various non-zero auto-scaled values, which causes these reports to look misleading at a glance.

This PR adds a `gnc:html-chart-set-y-axis-begin-at-zero!` function that can be used in reports to toggle the `beginAtZero` chart property. I also updated the charts I have issues with to use it.

I'd be happy to update more reports, but these are the ones I'm familiar with. I'm also willing to add check boxes to the report options to make this behavior optional. Another option would be to simply enable this behavior for all bar charts, but this method seems less disruptive. Let me know what you need.

### DIFF
--- a/gnucash/report/html-chart.scm
+++ b/gnucash/report/html-chart.scm
@@ -64,6 +64,7 @@
 (export gnc:html-chart-set-stacking?!)
 (export gnc:html-chart-set-grid?!)
 (export gnc:html-chart-set-y-axis-label!)
+(export gnc:html-chart-set-y-axis-begin-at-zero!)
 (export gnc:html-chart-add-data-series!)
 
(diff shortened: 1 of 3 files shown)


christopherlam:
Is there a rationale that only these two reports are set to include zero?

ZDBioHazard:
Just that they are the only two reports I personally have issues with.

I'm not familiar enough with the other reports to know if they need this tweak, so instead I opted to ask here before messing with them.

jralls:
@ZDBioHazard Are you going to implement this on the other reports? I think adding a report option to disable it is worthwhile, some users will want the truncated version.

jralls:
Closing as abandoned.

PR #1608, opened 2023-04-19
Bug 798775 - Why is General Journal called "Register" in the tabs?

dawansv:
Set default report name to General Journal in options to overwrite default report name from underlying register report.

Maybe there is a reason why the report name is kept at "register" for the General Journal and if so please ignore this PR. But otherwise if it's just an oversight, this one line commit should fix this bug.


christopherlam:
Thanks!

PR #1609, opened 2023-04-21
Bug 798461 - balance sheet shows positions with zero balances despite report options

dawansv:
This occurs when accounts with different currencies or commodities are being rolled up to a higher level of subaccounts. Fixed by removing commodities with zero totals from commoditylist when "omit zero balance figures" option is selected.

Code's author is Christopher Lam.

### DIFF
--- a/gnucash/report/html-acct-table.scm
+++ b/gnucash/report/html-acct-table.scm
@@ -1023,6 +1023,8 @@
                                               (omit-bal . #f)))
                                   (bal-sym (assq-ref bal-syms bal-method))
                                   (comm-amt (get-val env bal-sym)))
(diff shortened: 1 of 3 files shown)


dawansv:
I have tested this and it seems to work well.

Here are some screenshots. This one show the balance sheet WITHOUT the "omit zero balance figures" option selected

image

This one shows the same balance sheet with the "omit zero balance figures" option selected. Notice how the zeros are also removed from the rolled up accounts in Assets > Other Assets > Cars and also Liabilities > Bank Loan > Bank ABC

image

christopherlam:
Thanks!

PR #1611, opened 2023-04-23
Edit local symbol for uzbek sum (UZS)

muhammadyusuf-kurbonov:
Uzbek sum has not a symbol yet, and unit name is used. It's international form is `sum`.
But it's actually called `so'm` and should be written after digit part.


muhammadyusuf-kurbonov:
GNUCash:
image

Expected:
image

jralls:
Thanks!

PR #1612, opened 2023-04-24
Bug 797879 - [Transaction Report] running balance column not consistent with amount column for SAME transaction date

dawansv:
Added QUERY-DEFAULT-SORT as second parameter to query sortkey to ensure transactions are sorted as per register within the same TRANS-DATE-POSTED

The problem only occurs with the query sort, not the custom sort. For the trep-engine, sorting by date is a regular query sort unless a primary subtotal per date key (e.g. per month or year) is chosen. In that second case it shifts to a custom sort (the sort is done in the report) which is not experiencing this problem. 

Hence it's only a query sort issue illustrated below. Correct register sort is on the left, incorrect date ascending sort on the right.

running-balance-ro-vs-da

Looking further it seems the problem is with the sortkey for date as set for the query. Currently it is set to 

SPLIT-TRANS TRANS-DATE-POSTED 

which means sorted by posted date then split order within each transaction (I assume). It seems that it's not enough. Evidently we need to tell the query engine to use the register sort within the same date otherwise it uses some way to sort within a same date. The register sort is call QUERY-DEFAULT-SORT so I added that.

SPLIT-TRANS QUERY-DEFAULT-SORT TRANS-DATE-POSTED

It seems to fix the problem as per my test. After this chage, date descending matches register sort, and date ascending matches the exact opposite of register sort, and the running balance is much happier.

### DIFF
--- a/gnucash/report/trep-engine.scm
+++ b/gnucash/report/trep-engine.scm
@@ -2178,7 +2178,7 @@ be excluded from periodic reporting.")
          query
          (keylist-get-info (sortkey-list BOOK-SPLIT-ACTION) primary-key 'sortkey)
          (keylist-get-info (sortkey-list BOOK-SPLIT-ACTION) secondary-key 'sortkey)
-         '())
+         (list QUERY-DEFAULT-SORT))
         (qof-query-set-sort-increasing
          query (eq? primary-order 'ascend) (eq? secondary-order 'ascend)
          #t))


christopherlam:
If this is right, then you'll be fixing a 21 year old bug. This is the earliest I can find on the sort keys: 0f0080c964888d59d51c001087d953a1bd3ca05e

dawansv:
> This is the earliest I can find on the sort keys: [0f0080c](https://github.com/Gnucash/gnucash/commit/0f0080c964888d59d51c001087d953a1bd3ca05e)

Interesting. That said it's not easy to spot without the running balance (which came later I think) and the specific data showing it. I suspect something I did with my sample data makes it more prominent in my test file. It has random data I have added overtime, so I don't even remember where and when it was entered but I have moved the transactions a lot more than would be usual to different dates and even accounts, hence why it's possibly so prominent in the test file.

I have attached my sample gc file with the random test data to the bug report for reference. The sample data in my screenshot comes from the mock checking account from Jan 2022 to June 2022.
https://bugs.gnucash.org/show_bug.cgi?id=797879#c7

Also I notice from a quick grep that the following scm files use the same SPLIT-TRANS TRANS-DATE-POSTED combination so we should probably investigate whether it needs to be addressed there as well depending on what it's used for in there. I'll try to have a look.

gnucash/report/report-utilities.scm
gnucash/report/reports/example/average-balance.scm
gnucash/report/reports/example/daily-reports.scm
gnucash/report/reports/standard/general-journal.scm
gnucash/report/reports/standard/new-aging.scm
gnucash/report/reports/standard/new-owner-report.scm

dawansv:
Converted this to draft for now. Doing some more investigation on this, trying to understand the query sort parameters.

In libgnucash/engine/qofquery.cpp we have the procedure that sets the sort. It can take up to 3 parameters, but each of these are lists themselves.

```c
void
qof_query_set_sort_order (QofQuery *q,
                          QofQueryParamList *params1, QofQueryParamList *params2, QofQueryParamList *params3)
{
    if (!q) return;
    if (q->primary_sort.param_list)
        g_slist_free (q->primary_sort.param_list);
    q->primary_sort.param_list = params1;
    q->primary_sort.options = 0;

    if (q->secondary_sort.param_list)
        g_slist_free (q->secondary_sort.param_list);
    q->secondary_sort.param_list = params2;
    q->secondary_sort.options = 0;

    if (q->tertiary_sort.param_list)
        g_slist_free (q->tertiary_sort.param_list);
    q->tertiary_sort.param_list = params3;
    q->tertiary_sort.options = 0;

    q->changed = 1;
}
```

Then I noticed that general-journal.scm also uses the "SPLIT-TRANS TRANS-DATE-POSTED" sort but it is NOT showing the same problem, the sort is correct. So I looked at the whole thing:

```scheme 
    (qof-query-set-sort-order query
                              (list SPLIT-TRANS TRANS-DATE-POSTED)
                              (list QUERY-DEFAULT-SORT)
                              '())
```

So they do have QUERY-DEFAULT-SORT, but as a second parameter. Although they use SPLIT-TRANS TRANS-DATE-POSTED together, so I am not sure what the difference is between passing them all as one parameter or 2. 

Also I still don't know whether SPLIT-TRANS TRANS-DATE-POSTED means sort by SPLIT-TRANS then TRANS-DATE-POSTED or the other way around. The problem is I have no clue what SPLIT-TRANS does,

Note that for trep-engine we need all of these to be in one parameter, as the second parameter is used by the next level in the sort options. 

I ran (SPLIT-TRANS TRANS-DATE-POSTED QUERY-DEFAULT-SORT) as an alternative but that does not fix the problem.
However (QUERY-DEFAULT-SORT SPLIT-TRANS TRANS-DATE-POSTED) does fix it as well as my original fix
(SPLIT-TRANS QUERY-DEFAULT-SORT TRANS-DATE-POSTED)

At the end all we need might just be QUERY-DEFAULT-SORT because it is already sorted by TRANS-DATE-POSTED. But still I need to understand what SPLIT-TRANS does, because I really don't like changing code I don't understand. 

I'll do more research but who is the query engine guru here (if there is one)?

dawansv:
Ok so not my brightest hour. First what I said earlier about "at the end all we need might just be QUERY-DEFAULT-SORT" is wrong, because for instance I might want to sort by "posted date" then within that "description" and QUERY-DEFAULT-SORT as primary would clearly break that because if it is set as primary then the secondary is always ignored since by then each primary sort is in its unique group. So my first fix proposal would break that.

Also on further look and testing, some of these sort parameters clearly work as a pair and SPLIT_TRANS is required on all TRANS-* parameters such as TRANS-DESCRIPTION, TRANS-DATE-POSTED, TRANS-DATE-ENTERED, TRANS-NUM. Removing SPLIT-TRANS completely breaks these sorts, all in the same way, so possibly it reverts to sorting by the underlying transaction id or something without it. Needless to say it's part of these sort keys that start with TRANS-* just work as a pair, they are not 2 different sort keys per se. So my original fix actually was just sorting on QUERY-DEFAULT-SORT which won't work as I said above.

So we are left with this challenge. In places where we really only have one sort level and can use 2 parameters of qof-query-set-sort-order, we can simply do what general-journal.scm does, call "SPLIT-TRANS TRANS-DATE-POSTED" as first sort level and QUERY-DEFAULT-SORT as second level. 

However in the trep-engine, the code is standardized to get ONE sort key for the primary sort and ONE sort key for the secondary sort level as chosen by the user. But luckily qof-query-set-sort-order has a third level we are not using, it's just set to an empty list in the current code. 

So the correct fix is to use the third (hidden to the user) parameter offered by qof_query_set_sort_order and ALWAYS set it to QUERY-DEFAULT-SORT. That would say the the very final level, if there is still something to sort, should be QUERY-DEFAULT-SORT which in fact is what would be expected without that parameter but evidently is not the case.

So this is the fix I propose, and I am much more confident about it.

Note: please ignore the last sentence in my commit comment "Also set the third level to follow the sort order of the second." That was another brain fart on my part wich I forgot to remove from the comment.

jralls:
Thanks!

PR #1614, opened 2023-04-25
Bug 798886 - [Transaction Report] Subtotal upper headings do not follow font style of lower headings

dawansv:
Added "total-label-cell" market to upper headings to match lower headings

This is a small bug that has really, well, bugged me. I usually fix it on my own copy but here is a PR to finally get it in.

**This is what it looks like BEFORE the fix**. Notice the font size of "Cash in Wallet" and "Checking Account" as well as Month subheadings.
image

**This is what it looks like after the fix**. Font sizes can now be set via the stylesheets as they are for lower subtotal headings. Eventually it would be nice to have separate cell and row styles for upper and lower, as well as for primary and secondary (row style is different but font isn't), but for now this is already an improvement.
image

### DIFF
--- a/gnucash/report/trep-engine.scm
+++ b/gnucash/report/trep-engine.scm
@@ -1445,12 +1445,12 @@ be excluded from periodic reporting.")
             (gnc:html-make-empty-cells left-indent)
             (if export?
                 (cons
-                 (gnc:make-html-table-cell data)
+                 (gnc:make-html-table-cell/markup "total-label-cell" data)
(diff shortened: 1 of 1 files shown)


christopherlam:
I think this is a welcome change. It did annoy me for a while, but at a lower threshold :) Any other opinions?

jralls:
Looks like a slam dunk to me. Merge it!

christopherlam:
It's an easy thanks then!

PR #1616, opened 2023-04-25
Restore register description quickfill

Bob-IT:
Here is a PR to restore the register quickfill to how it was.

I have been testing some changes that would of fixed most of the bugs but I do not think I will have enough time to fully test and complete the changes before the weekend so probably worth reverting my changes and clear half a dozen bugs.

What I have achieved with the test file from [Bug 798788](https://bugs.gnucash.org/show_bug.cgi?id=798788) which has 16804 transaction with 16804 unique descriptions that are 128 characters long.
Register load time, 0.22 seconds
Completion test from start of description, say pressing 9 resulted in a sorted list store with 265 descriptions in 0.12 seconds
Completion test any part of description, say pressing 9 resulted in a sorted list store with 10920 descriptions in 1.25 seconds

I achieved this by creating a new cell type and loading the descriptions in to a hash table and once the first character was typed create the sorted list store.
With the use of a secondary entry icon, I am able to choose to do the search from the start or any place and with a new preference a default could be set.

### DIFF
--- a/gnucash/gnome/gnc-plugin-page-register.c
+++ b/gnucash/gnome/gnc-plugin-page-register.c
@@ -4891,6 +4891,23 @@ gnc_plugin_page_register_cmd_schedule (GSimpleAction *simple,
     LEAVE (" ");
 }
 
+static gboolean
+strip_text (const gchar *text_in, gchar **text_out)
+{
+    if (text_in)
+    {
+        *text_out = g_strdup (text_in);
+        gnc_utf8_strip_invalid_and_controls (*text_out);
+        g_strstrip (*text_out);
+
+        if (g_strcmp0 (text_in, *text_out) != 0)
+            return TRUE;
+
+        g_free (*text_out);
+    }
+    return FALSE;
+}
+
 static void
 scrub_split (Split *split)
 {
@@ -4903,6 +4920,28 @@ scrub_split (Split *split)
     trans = xaccSplitGetParent (split);
     lot = xaccSplitGetLot (split);
     g_return_if_fail (trans);
+
+    const gchar *desc_text = xaccTransGetDescription (trans);
+    gchar *text_out = NULL;
+    if (strip_text (desc_text, &text_out))
+    {
+        xaccTransSetDescription (trans, text_out);
+        g_free (text_out);
+    }
+
+    const gchar *notes_text = xaccTransGetNotes (trans);
+    if (strip_text (notes_text, &text_out))
+    {
+        xaccTransSetNotes (trans, text_out);
+        g_free (text_out);
+    }
+
+    const gchar *memo_text = xaccSplitGetMemo (split);
+    if (strip_text (memo_text, &text_out))
+    {
+        xaccSplitSetMemo (split, text_out);
+        g_free (text_out);
+    }
 
     xaccTransScrubOrphans (trans);
     xaccTransScrubImbalance (trans, gnc_get_current_root_account(), NULL);
--- a/gnucash/register/ledger-core/split-register-load.c
+++ b/gnucash/register/ledger-core/split-register-load.c
@@ -878,8 +878,6 @@ gnc_split_register_load_desc_cells (SplitRegister* reg)
     cell = (ComboCell*)
            gnc_table_layout_get_cell (reg->table->layout, DESC_CELL);
 
-    gnc_combo_cell_use_type_ahead_only (cell);
-
     gnc_combo_cell_use_list_store_cache (cell, store);
 }
 /* ====================== END OF FILE ================================== */


gjanssens:
Unless you think it won't make 5.2 either, I would prefer to not revert. It's not optimal in some cases but is not horribly broken either.

As usual I also keep reservations with adding options in this case :wink:
At least I would wait until we have fully established we can't tweak the new behaviour to also satisfy proponents of the old way.

Most important objections I have seen so far are:
1. No way to  cancel quickfill. That one should be relatively easy to implement, no ? Can we listen for an escape character to abort any quickfill ?
2. Exact match may be lost in long list of matches. That should be a matter of customizing the sort algorithm.
3. Most recent match is more relevant than alphabetic match (which can be quite old). Also this is a matter of tweaking the sort algorithm.
4. Performance. It seems the current quickfill kicks in from the first character typed. Would it perform better if we only create a list starting from the third character typed ? I imagine the list of potential matches should be drastically reduced by then and I vaguely remember having seen this in use in other applications or dialogs, though I can't immediately find an example.

How much of this can be fixed in a short notice ?

jralls:
I don't think we want the feature disappearing and reappearing later, and I think most of the complaints could be addressed by changing the sort to MRU from alphabetic. A nice tweak would be to override that with an exact match if there is one being first and the MRU matches following. @Bob-IT can you do that by Friday?

The performance problems are another matter. We know one of them is due to newlines--in particular Windows newlines--mucking things up; the other is on Linux and is likely a different problem. We're obviously not going to get that fixed this week.

Bob-IT:
OK, maybe I was paying too much attention on the negative in the bugs.

As I understand it, all we want to achieve in the description cell is to get some string which is then used when one tabs to auto fill the transaction, I assume this is where it picks the last transaction with that string.

The first commit changes the search to start from the beginning of the text which is how it was, albeit now uses the combo quickfill code.

The second commit was what I was thinking for stripping control characters and also top and tailing spaces in the transaction check and repair for the description,notes and memo fields.

Bob-IT:
> As usual I also keep reservations with adding options in this case wink At least I would wait until we have fully established we can't tweak the new behaviour to also satisfy proponents of the old way.

That's OK, it was an option, for testing I am using this, works OK but sometimes I am having mouse clicks issues...
desc-test

> 
> Most important objections I have seen so far are:
> 
>     1. No way to  cancel quickfill. That one should be relatively easy to implement, no ? Can we listen for an escape character to abort any quickfill ?

Not sure what this is about, escape closes the list and as before adding a space stops the quickfill to allow changing 'fred' to 'Fred '
> 
>     2. Exact match may be lost in long list of matches. That should be a matter of customizing the sort algorithm.
> 

The new version would truncate the list as in image above.
>     3. Most recent match is more relevant than alphabetic match (which can be quite old). Also this is a matter of tweaking the sort algorithm.
> 

The list has nothing to do with how recent it was used, it was just sorted for convenience and that I was finding all entries with say a 't' somewhere in the description where some users would  like to start from the front.
>     4. Performance. It seems the current quickfill kicks in from the first character typed. Would it perform better if we only create a list starting from the third character typed ? I imagine the list of potential matches should be drastically reduced by then and I vaguely remember having seen this in use in other applications or dialogs, though I can't immediately find an example.

Originally I was going to use the `GtkEntryCompletion` but the list would not popup, it actually may of been popped behind the sheet but could not figure it out, I think I remember trying to do this before a while back and did not remember why I gave up. With the `GtkEntryCompletion` there was a property that controlled when to popup the list on the number of characters typed in.

The main performance issue with the current combo is two fold, by default the list store is set to sort and the overhead in adding to it when the register is loaded. To overcome the second, I just add all the descriptions to a hash table and only when the first character is typed do I start adding a reduced number of descriptions to the list store. With the sorting, I disable the sorting, disconnect the store from the treeview, populate the list store, enable the sorting but with collate, reconnect the treview.

Doing this I think performance is OK, when searching from the start, just takes a little longer when searching anywhere in the string. Currently I am always starting the searches from the hash with an empty list store but want to try to add a `GtkTreeModelFilter` and then just change the filter, only when the cell is empty would I start from the hash table.

jralls:
I just tried a build of this PR. It does no filtering, showing the whole (sorted alphabetically) list of all descriptions/memos/notes in that register with the typing so far making the first selection:

Screen Shot 2023-04-28 at 11 02 23 AM

That won't shut up the complainers because the old quickfill behavior supplied the most recent string that matched and this selects the first alphabetically.  It will also upset the folks who've embraced the new way of being able to filter on something from the middle of the field . (I'm in that camp.)

Bob-IT:
Looking more closely at this, I realise I was missing a bit of information, when the descriptions are added to the quickfill the order matters and by default they are retrieved in LIFO order. I thought it was just the descriptions that mattered with the auto complete code finding the last transaction. I suppose when the split list is reversed, it should be FIFO order.

I will change my new code to cater for this as an option, unfortunately I am busy this weekend but hopefully will have some thing to test in the next week or two.

dawansv:
> Not sure what this is about, escape closes the list and as before adding a space stops the quickfill to allow changing 'fred' to 'Fred '

Not quite what I see on my side at least. This is the behavior I see as of version 5.0 on Ubuntu.

Imagine I want to write "Bank of America" as my description. It's not currently in my quickfill but "Bank of America NY Branch" is in the list.

1. I type "Bank of America". It matches "Bank of America NY Branch" in the quickfill list. My description still says "Bank of America". Yet I can see that if I simpy press tab now then the "Bank of America NY Branch" will be used.
2. So I press Escape to close the list. That's promissing, because now the list is closed and my description just says "Bank of America".
3. However when I now press Tab, the quickfill still kicks in for "Bank of America NY Branch". That is not what I would expect. I wanted to enter "Bank of America". I would expect Escape Tab to let me do that.

Now compare this with the behavior in version 4.13:
1. I type "Bank of America". It matches "Bank of America NY Branch" in the description field as I type
2. As I reach the last a of America while typing, " NY Branch" is still seletced. I simply press Delete or Backscape to delete that selection, an intuitive key press since I see that " NY Branck" is selected and I want to get rid of it. This leaves me with just "Bank of America"
3. I press tab and no autofill takes place. This is what I want. My description now says "Bank of America" with no autofill for the account and amount.

As for adding a space after "Bank of America", right now in version 5.0 it simply reopens the list to "Bank of America NY Branch" which is actually what I would expect, so I don't have a problem with that but it still doesn't let me enter "Bank of America"

dawansv:
Just for reference, a mention that there is more user feedback about this on the bug report, which might help shape the tweaking of this feature: https://bugs.gnucash.org/show_bug.cgi?id=798829

Also would it be better to open a new PR at some point with a more "optimistic" outlook instead of this one? I understand the original intent here was not pessimistic per se, it was just a temporary pause, but if we want a place to discuss this further (assuming more user feedback is welcome) then should it be the bug report, or can we have a new PR for more flexibility? The interface of bugzila is fine for small bugs or long term discussion, but not always the easiest (compared to github) to share some specifics of enhancement requests as the implementation is evolving such as screenshots, etc.

Bob-IT:
I have created a new PR #1625 with a revised description completion control so closing this one.

PR #1617, opened 2023-04-26
Bug 798879 - RFE: [Transaction Report] add Running Total option

dawansv:
### Summary
Running total are calculated based on report transactions sort order as shown. Can be displayed on the main total as well as the primary and secondary subtotal levels. Compatible with multi-currency options and multi-line (splits) display.
### Context
The idea came when I was trying to use the running balance feature to create running totals and realized it wasn't doing what I wanted (although it's a usefull feature in itself). What I wanted was a running total based on the transactions in the report in  whatever order and filtering I had selected. 

I started to work on in gnucash version 4. My code was improved based on useful (and very patient) feedback I received on a couple of previous PR (code in the [first one](https://github.com/Gnucash/gnucash/pull/1459) was pretty bad, [second one](https://github.com/Gnucash/gnucash/pull/1460) better) and is now presented with additional improvements for version 5. It is a work in progress requiring more pairs of eyes both in terms of code but also presentation. Because it changes trep-engine, care must also be taken to not impact other reports and custom reports based on it, so additional testing will be done there as well. In the meantime I am opening this draft PR to get additional feedback.
### The Feature
The running totals work on all 3 levels: Grand Total, Primary Subtotal and Secondary Subtotal. They can be enabled or disabled separetely. In the screenshot below all 3 are enabled at once (with a small dataset to keep the screenshot manageable).
image

The options to enables them are right under each of their respective options (so in the Display section for the grand total and the Sorting section for the subtotals). 
image

image

The running totals are calculated as per the grouping and ordering presented in the report, not the underlying data. It is for the user to decide whether a particular grouping and ordering is relevant (such as combining transactions from several accounts in the same list ordered by date). But the running totals will follow whatever grouping and sorting option the user chooses. 

Here is an example of a running total calculated by combining transactions from both wallet and checking accounts mixed together with an unusual descending date sort order. This examples is meant to highlight that the relevance of the nature and order of transaction data presented is left to the user, the running total is not judgmental in that regard.
image

The feature is fully compliant with the multi currency display (including showing converted and original amounts, each with their own set of running total and subtotals columns). 
image

### Caveats

1. Although fully compliant with the multi currency display, in the multicurrency columns, only the running total of the current split’s currency is shown on each line, while all currency are shown in the final total at the bottom on the corresponding amount column (current behavior). Showing all currencies on each transaction line would not only require significant code change but also most likely clutter the report. Additionally, the total is not currently shown on the running total columns themselves as that would require significant additional changes to how subtotals are calculated and displayed. Minimizing the effect is the fact that running totals are only enables when their corresponding total or subtotal is enabled.
2. The running totals get their data from the splits, not the other calculated columns themselves. If one were to transform the data in a custom calculator-function for the amount columns, then of course the running total calculator-functions would need to be updated accordingly with their own custom calculator-functions. That is to be expected and fully supported by the trep-engine’s custom-calculated-cells feature. 
3. The running totals suffer from the same issue as the subtotals as related to the “Sign Reverses” option. Just as they are for the subtotals, signs are not reversed for the running totals since they have to match the totals on the last transaction of their respective level. The issues around the “Sign Reverses” option have been discussed other places and if anything is ever decided as relating to either that option or changes to the subtotals then the same would be done to the running totals. However there is nothing that can be done here differently until that other issue is resolved.

### Technical brief and code review
Although this introduces close to 200 lines of new code and a few lines of modified code, most additions are pretty benign boilerplate to create new options, new columns, etc. following existing code structures, and with plenty of comments to help with long-term maintainability.

There is only one major change (discussed in detail below) which calls for the introduction of a second argument in the calculated-cells calculator-functions (while preserving the support of calculator-functions with a single argument) with related code change in the add-split-row renderer.

A second smaller change was made to selecting which number columns are formatted with a link to the underlying transaction in the register is also discussed below (discussed further down).

A third small change is in removing columns that do not have subtotals enabled from calculated-cells when “Show subtotals only” is selected. This should have been done when that feature was developed, and the missing code caused empty columns to be displayed in that view (including the existing account balance column) an issue made more apparent with the new running total columns.

### Changes to add-split-row and the related calculator-functions
Add-split-row is the procedure that is called to add each row of data to the report. It received a split and runs it through the calculator-function for each cell. 

In the definition part of the report, the columns are divided in two groups:

1. the left-columns (date, description, notes, etc.), and 
2. the calculated-cells (also referred as right-columns) that contain the numeric data calculated from the split amount or retrieved based on it (single amount, credit, debit, running balance, etc).

One key difference between the left-columns and calculated-cells is that the function that are called to retrieve the content of the left-columns are provided with a second parameter called transaction-row? which differentiates the splits in a multiline situation (when "Display – Detail Level – One Split per Line” is selected). The flag is set to #t when the current split is the “main” split related to the account and #f when the current split is another split. This allows the left-columns to only show the date and description when the main split is shown, and leaves that information blank when not (since it is the same for the splits).

Unlike the left-columns, the calculator-functions for the calculated-cells do not have that argument. They only get the split with no way to tell if it’s a main split or “other” split. This is a problem for the running totals in a multiline environment since they should only be computed and shown on the main split, not the other splits. In fact the same problem exists on the Running Balance which for now is disabled in the multiline context. 

Hence my code changes implement this additional argument for the calculator-functions (calculator-function split transaction-row?) while preserving the ability to have calculator-functions with only one argument (calculator-function split) for legacy purpose. 

This results in one of the main code additions and changes in add-split-row

```
             ;; Returns the number of required arguments of a function
             (fn-arity (lambda (fn) (car (procedure-minimum-arity fn))))
             ;; Calls the calculator-function for the provided cell with either
             ;; two arguments (calculator-function split transaction-row?) or only
             ;; one argument  (calculator-function split) for legacy support.
             (calc-fn (lambda (cell split transaction-row?)
                              (if (= (fn-arity (vector-ref cell 1)) 1)
                                  ((vector-ref cell 1) split)
                                  ((vector-ref cell 1) split transaction-row?))))
```
The code uses a guile-specific procedure called procedure-minimum-arity. It is not widely documented because not standard RnRS scheme but it was introduced in Guile version 2.0.0 (it is mentioned in the changelog for that version) and is till supported as of the current version 3.0.9. It’s pretty straightforward as explained by (help procedure-minimum-arity)
`procedure-minimum-arity' is a procedure in the (guile) module.

- Scheme Procedure: procedure-minimum-arity proc
     Return the "minimum arity" of a procedure.

     If the procedure has only one arity, that arity is returned as a
     list of three values: the number of required arguments, the number
     of optional arguments, and a boolean indicating whether or not the
     procedure takes rest arguments.

     For a case-lambda procedure, the arity returned is the one with the
     lowest minimum number of arguments, and the highest maximum number
     of arguments.

     If it was not possible to determine the arity of the procedure,
     `#f' is returned.

See the specific commit for this procedure: https://git.savannah.gnu.org/cgit/guile.git/commit/?id=cb2ce548441824fe1284fc80a3a95394a9fc03d0

For our purpose we can ignore optional and rest argument, and are only looking at the number of required argument, assuming either 1 (for legacy purpose) or otherwise expecting 2 as the else. Hence we simply retrieve the car of the procedure-minimum-arity for the provided calculator-functions to decide to either call it with 1 or 2 arguments. This therefore requires no changes to existing calculator-functions in trep-engine or trep-engine-based report but allows for addition of new calculator-functions with the 2 arguments.

I feel the addition of the transaction-row? argument is a worthwhile addition to the calculated-cells to bring them in line with the left-columns even outside of this current feature. For instance it could also be used to fix the running balance display in the multiline context or used by custom reports for other reasons.

### Alternative to changes to add-row-split and calculator-functions’ number of arguments
If this change is considered to bold, the alternative would be to disable running totals in the multiline context for now, although that brings the smaller challenge of disabling “on the fly” the subtotal options in the ui (in the sorting section) when the “One Split per Line” is selected in the display section. I don’t think there is any way to do this right now while the dialog is active, or at least there is no hook I could find when changing tabs in the dialog (I could only get it done when it is opened the next time, or apply is clicked, or when a related option is clicked in the sorting window). That said the actual option can be disabled in the actual report output regardless of option graying in the dialog. So this is purely for the visual in the dialog and could possibly be left active with additional info in the option popup message stating it doesn’t work when “One Split per Line” is selected. 

### Changes to which number columns are formatted as links
The current implementation offers the “Enable Links” option in the display section. When enabled, it is applied to each cell that has a number in it. One limitation of this approach is that it is even applied to number cells whose calculated number is different from the transaction amount in the register (for instance a converted amount, or a running total). Also it can clutter the report visually since the link is essentially the same for each cell in a given row, pointing to the split in the register.

Ideally there would be an additional flag in the column definition to decide whether a link should or shouldn’t be present on that column, but that would require additional changes for such a small feature with potential impact on custom reports. The alternative I propose is to limit links (when enabled) to columns already marked as displaying a subtotal so as to declutter the report. To avoid an edge case of being left with no links on the report (which could happen if no columns with subtotals are enabled), the code reverts back to putting links on any number column in case no columns with subtotals are enabled.

### DIFF
--- a/gnucash/report/trep-engine.scm
+++ b/gnucash/report/trep-engine.scm
@@ -1636,14 +1636,19 @@ be excluded from periodic reporting.")
       ;; this part will check whether custom-calculated-cells were specified. this
       ;; describes a custom function which consumes an options list, and generates
       ;; an association list similar to default-calculated-cells as above.
-      (if custom-calculated-cells
-          (let ((cc (custom-calculated-cells options)))
-            (cond
-             ((not (pair? cc)) (gnc:error "welp" cc) default-calculated-cells)
-             ((vector? (car cc)) (upgrade-vector-to-assoclist cc))
-             ((any invalid-cell? cc) (gnc:error "welp" cc) default-calculated-cells)
-             (else cc)))
-          default-calculated-cells))
+      (let ((cc (if custom-calculated-cells
+                    (let ((ccc (custom-calculated-cells options)))
+                      (cond
+                        ((not (pair? ccc)) (gnc:error "welp" ccc)
+                          default-calculated-cells)
+                        ((vector? (car ccc)) (upgrade-vector-to-assoclist ccc))
+                        ((any invalid-cell? ccc) (gnc:error "welp" ccc)
+                          default-calculated-cells)
+                        (else ccc)))
+                    default-calculated-cells)))
+        ;; Only keep cells with subtotals when "Show subtotals only" is selected
+        ;; otherwise leave all calculated-cells as is.
+        (if (column-uses? 'subtotals-only) (filter cell-with-subtotals? cc) cc)))
 
     (define headings-left-columns
       (map (cut assq-ref <> 'heading) left-columns))
--- a/gnucash/report/trep-engine.scm
+++ b/gnucash/report/trep-engine.scm
@@ -74,6 +74,11 @@
 ;;Display
 (define optname-detail-level (N_ "Detail Level"))
 (define optname-grid (N_ "Subtotal Table"))
+;; Translators: a running total is a total that is continually adjusted on every line.
+;; To be consistent, also consider how the term "Running Balance" is translated.
+;; "Running Totals" is the plural form as it refers to the running total and running subtotals.
+;; To be consistent, also consider how the singular form "Running Total" is translated.
+(define optname-running-totals (N_ "Running Totals"))
 
 ;;Sorting
 (define pagename-sorting (N_ "Sorting"))
@@ -983,7 +988,16 @@ be excluded from periodic reporting.")
       gnc:pagename-display (N_ "Sign Reverses")
       "m1" (G_ "Reverse amount display for certain account types.")
       "global"
-      (keylist->vectorlist sign-reverse-list)))
+      (keylist->vectorlist sign-reverse-list))
+
+    (gnc-register-multichoice-option options
+      gnc:pagename-display optname-running-totals
+      "o2" (G_ "Display running totals as per report sort order?")
+      "none"
+      (list (vector 'none  (G_ "None"))
+            (vector 'all   (G_ "Grand Total and Subtotals"))
+            (vector 'grand (G_ "Grand Total Only"))
+            (vector 'sub   (G_ "Subtotals Only")))))
 
   ;; this hidden option will toggle whether the default
   ;; qof-query is run, or a different query which ensures
@@ -1058,6 +1072,17 @@ be excluded from periodic reporting.")
                      (or (primary-get-info 'renderer-fn)
                          (secondary-get-info 'renderer-fn))))
           (cons 'running-balance (opt-val gnc:pagename-display "Account Balance"))
+          (cons 'running-grand-total
+                (or (eq? (opt-val gnc:pagename-display optname-running-totals) 'grand)
+                    (eq? (opt-val gnc:pagename-display optname-running-totals) 'all)))
+          (cons 'running-prime
+                (and (primary-get-info 'renderer-fn)
+                     (or (eq? (opt-val gnc:pagename-display optname-running-totals) 'sub)
+                         (eq? (opt-val gnc:pagename-display optname-running-totals) 'all))))
+          (cons 'running-sec
+                (and (secondary-get-info 'renderer-fn)
+                     (or (eq? (opt-val gnc:pagename-display optname-running-totals) 'sub)
+                         (eq? (opt-val gnc:pagename-display optname-running-totals) 'all))))
           (cons 'account-full-name
                 (opt-val gnc:pagename-display (N_ "Use Full Account Name")))
           (cons 'memo (opt-val gnc:pagename-display (N_ "Memo")))
@@ -1142,6 +1167,18 @@ be excluded from periodic reporting.")
            (opt-val pagename-currency optname-currency)
            (gnc:accounts-get-commodities c_account_1 #f) enddate #f #f)
           gnc:exchange-by-pricedb-nearest))
+
+    ;; Returns #t if a calculated-cell definition has the subtotal flag
+    (define (cell-with-subtotals? cell)
+      (assq-ref cell 'subtotal?))
+
+    ;; Collectors for running total and subtotals
+    (define converted-running-total-collector (gnc:make-commodity-collector))
+    (define original-running-total-collector (gnc:make-commodity-collector))
+    (define converted-running-prime-collector (gnc:make-commodity-collector))
+    (define original-running-prime-collector (gnc:make-commodity-collector))
+    (define converted-running-sec-collector (gnc:make-commodity-collector))
+    (define original-running-sec-collector (gnc:make-commodity-collector))
 
     (define left-columns
       (let* ((add-if (lambda (pred? . items) (if pred? items '())))
@@ -1347,7 +1384,46 @@ be excluded from periodic reporting.")
                                           (gnc:monetary-neg (original-amount s tr?)))))
            (original-account-balance (lambda (s tr?)
                                        (gnc:make-gnc-monetary
-                                        (split-currency s) (xaccSplitGetBalance s)))))
+                                        (split-currency s) (xaccSplitGetBalance s))))
+           ;; Helper function to add the splits to the running total collectors.
+           ;; Third argument is the function that returns converted or original amount.
+           ;; Returns current monetary from specified running total collector.
+           (add-split-to-collector (lambda (s tr? coll amt-fn)
+                                        (coll 'add
+                                          (gnc:gnc-monetary-commodity (amt-fn s tr?))
+                                          (gnc:gnc-monetary-amount (amt-fn s tr?)))
+                                        (coll 'getmonetary
+                                          (gnc:gnc-monetary-commodity (amt-fn s tr?)) #f)))
+           (converted-running-total (lambda (s tr?)
+                                      (and tr?
+                                        (add-split-to-collector s tr?
+                                          converted-running-total-collector
+                                          converted-amount))))
+           (converted-running-prime (lambda (s tr?)
+                                      (and tr?
+                                        (add-split-to-collector s tr?
+                                          converted-running-prime-collector
+                                          converted-amount))))
+           (converted-running-sec (lambda (s tr?)
+                                      (and tr?
+                                        (add-split-to-collector s tr?
+                                          converted-running-sec-collector
+                                          converted-amount))))
+           (original-running-total (lambda (s tr?)
+                                      (and tr?
+                                        (add-split-to-collector s tr?
+                                          original-running-total-collector
+                                          original-amount))))
+           (original-running-prime (lambda (s tr?)
+                                      (and tr?
+                                        (add-split-to-collector s tr?
+                                          original-running-prime-collector
+                                          original-amount))))
+           (original-running-sec (lambda (s tr?)
+                                      (and tr?
+                                        (add-split-to-collector s tr?
+                                          original-running-sec-collector
+                                          original-amount)))))
         (append
          ;; each column will be a list of pairs whose car is a metadata header,
          ;; and whose cdr is the procedure, string or bool to obtain the metadata
@@ -1411,6 +1487,63 @@ be excluded from periodic reporting.")
                              (cons 'merge-dual-column? #f))))
              '())
 
+         (if (column-uses? 'running-sec)
+             (list (list (cons 'heading (header-commodity
+                           ;; Translators: this is the running total for the secondary subtotal.
+                           ;; For translation to be consistent, make sure it follows the same
+                           ;; pattern as for these other strings: “Running Totals”,
+                           ;; "Secondary Subtotal” and "Running Primary Subtotal"
+                           (G_ "Running Secondary Subtotal")))
+                         (cons 'calc-fn converted-running-sec)
+                         (cons 'reverse-column? #f)
+                         (cons 'subtotal? #f)
+                         (cons 'start-dual-column? #f)
+                         (cons 'friendly-heading-fn (const ""))
+                         (cons 'merge-dual-column? #f)))
+             '())
+
+         (if (column-uses? 'running-prime)
+             (list (list (cons 'heading (header-commodity
+                           (if (secondary-get-info 'renderer-fn)
+                               ;; Translators: this is the running total for the primary subtotal.
+                               ;; For translation to be consistent, make sure it follows the same
+                               ;; pattern as for these other strings: “Running Totals” and
+                               ;; “Primary Subtotal”
+                               (G_ "Running Primary Subtotal")
+                               ;; Translators: "Running Subtotal" is a shorter version of
+                               ;; "Running Primary Subtotal" used when a running primary subtotal
+                               ;; is displayed without a secondary subtotal.
+                               (G_ "Running Subtotal"))))
+                         (cons 'calc-fn converted-running-prime)
+                         (cons 'reverse-column? #f)
+                         (cons 'subtotal? #f)
+                         (cons 'start-dual-column? #f)
+                         (cons 'friendly-heading-fn (const ""))
+                         (cons 'merge-dual-column? #f)))
+             '())
+
+         (if (column-uses? 'running-grand-total)
+             (list (list (cons 'heading (header-commodity
+                           (if (or (primary-get-info 'renderer-fn)
+                                   (secondary-get-info 'renderer-fn))
+                               ;; Translators: this is the running total for the grand total.
+                               ;; For translation to be consistent, make sure it follows the same
+                               ;; pattern as for these other strings: “Running Totals” and
+                               ;; "Grand Total”
+                               (G_ "Running Grand Total")
+                               ;; Translators: "Running Total" is a shorter version of
+                               ;; "Running Grand Total" used when the running grand total is
+                               ;; displayed without subtotals. To be consistent, also consider
+                               ;; how the plural form "Running Totals" is translated.
+                               (G_ "Running Total"))))
+                         (cons 'calc-fn converted-running-total)
+                         (cons 'reverse-column? #f)
+                         (cons 'subtotal? #f)
+                         (cons 'start-dual-column? #f)
+                         (cons 'friendly-heading-fn (const ""))
+                         (cons 'merge-dual-column? #f)))
+             '())
+
          (if (and (column-uses? 'amount-original-currency)
                   (column-uses? 'amount-single))
              (list (list (cons 'heading (G_ "Amount"))
@@ -1457,6 +1590,46 @@ be excluded from periodic reporting.")
                              (cons 'start-dual-column? #f)
                              (cons 'friendly-heading-fn #f)
                              (cons 'merge-dual-column? #f))))
+             '())
+
+         (if (and (column-uses? 'amount-original-currency)
+                  (column-uses? 'running-sec))
+             (list (list (cons 'heading (G_ "Running Secondary Subtotal"))
+                         (cons 'calc-fn original-running-sec)
+                         (cons 'reverse-column? #f)
+                         (cons 'subtotal? #f)
+                         (cons 'start-dual-column? #f)
+                         (cons 'friendly-heading-fn (const ""))
+                         (cons 'merge-dual-column? #f)))
+             '())
+
+         (if (and (column-uses? 'amount-original-currency)
+                  (column-uses? 'running-prime))
+             (list (list (cons 'heading
+                           (if (secondary-get-info 'renderer-fn)
+                               (G_ "Running Primary Subtotal")
+                               (G_ "Running Subtotal")))
+                         (cons 'calc-fn original-running-prime)
+                         (cons 'reverse-column? #f)
+                         (cons 'subtotal? #f)
+                         (cons 'start-dual-column? #f)
+                         (cons 'friendly-heading-fn (const ""))
+                         (cons 'merge-dual-column? #f)))
+             '())
+
+         (if (and (column-uses? 'amount-original-currency)
+                  (column-uses? 'running-grand-total))
+             (list (list (cons 'heading
+                           (if (or (primary-get-info 'renderer-fn)
+                                   (secondary-get-info 'renderer-fn))
+                               (G_ "Running Grand Total")
+                               (G_ "Running Total")))
+                         (cons 'calc-fn original-running-total)
+                         (cons 'reverse-column? #f)
+                         (cons 'subtotal? #f)
+                         (cons 'start-dual-column? #f)
+                         (cons 'friendly-heading-fn (const ""))
+                         (cons 'merge-dual-column? #f)))
              '()))))
 
     (define calculated-cells
@@ -1770,7 +1943,13 @@ be excluded from periodic reporting.")
                      (and cell-content
                           (gnc:make-html-table-cell/markup
                            "number-cell"
-                           (if opt-use-links?
+                           ;; If link option is enabled, to avoid cluttering, we show links
+                           ;; only on number cells that are set to show a subtotal,
+                           ;; unless no columns are set to show a subtotal, in which case links
+                           ;; are shown on all number cells.
+                           (if (and opt-use-links? (or (cell-with-subtotals? cell)
+                                                       (not (any cell-with-subtotals?
+                                                                 cell-calculators))))
                                (gnc:html-split-anchor split cell-content)
                                cell-content)))))
                  cell-calculators))))
@@ -1902,7 +2081,11 @@ be excluded from periodic reporting.")
                 (when secondary-subtotal-comparator
                   (add-subheading (render-summary next 'secondary #t)
                                   def:secondary-subtotal-style next
-                                  'secondary))))
+                                  'secondary))
+                (converted-running-prime-collector 'reset #f #f)
+                (original-running-prime-collector 'reset #f #f)
+                (converted-running-sec-collector 'reset #f #f)
+                (original-running-sec-collector 'reset #f #f)))
 
              (else
               (when (and secondary-subtotal-comparator
@@ -1926,7 +2109,9 @@ be excluded from periodic reporting.")
                  secondary-subtotal-collectors)
                 (when next
                   (add-subheading (render-summary next 'secondary #t)
-                                  def:secondary-subtotal-style next 'secondary)))))
+                                  def:secondary-subtotal-style next 'secondary)
+                  (converted-running-sec-collector 'reset #f #f)
+                  (original-running-sec-collector 'reset #f #f)))))
 
             (loop rest (not odd-row?) (1+ work-done)))))
 
--- a/gnucash/report/trep-engine.scm
+++ b/gnucash/report/trep-engine.scm
@@ -74,6 +74,7 @@
 ;;Display
 (define optname-detail-level (N_ "Detail Level"))
 (define optname-grid (N_ "Subtotal Table"))
+(define optname-grand-total (N_ "Grand Total"))
 ;; Translators: a running total is a total that is continually adjusted on every line.
 ;; To be consistent, also consider how the term "Running Balance" is translated.
 ;; "Running Totals" is the plural form as it refers to the running total and running subtotals.
@@ -926,7 +927,7 @@ be excluded from periodic reporting.")
       ;; note the "Amount" multichoice option in between here
       (list optname-grid                        "m5" (G_ "Display a subtotal summary table.") #f)
       (list (N_ "Account Balance")              "n"  (G_ "Display the balance of the underlying account on each line?") #f)
-      (list (N_ "Totals")                       "o"  (G_ "Display the totals?") #t)))
+      (list optname-grand-total                 "o"  (G_ "Display a grand total section at the bottom?") #t)))
 
     (when BOOK-SPLIT-ACTION
       (gnc-register-simple-boolean-option options
@@ -1321,7 +1322,7 @@ be excluded from periodic reporting.")
 
         (if (or (column-uses? 'subtotals-only)
                 (and (null? left-cols-list)
-                     (or (opt-val gnc:pagename-display "Totals")
+                     (or (opt-val gnc:pagename-display optname-grand-total)
                          (primary-get-info 'renderer-fn)
                          (secondary-get-info 'renderer-fn))))
             `(((heading . "") (renderer-fn . ,(const #f))))
@@ -2005,7 +2006,7 @@ be excluded from periodic reporting.")
 
       (if (null? splits)
 
-          (when (opt-val gnc:pagename-display "Totals")
+          (when (opt-val gnc:pagename-display optname-grand-total)
             (gnc:html-table-append-row/markup!
              table def:grand-total-style
              (list
--- a/gnucash/report/reports/standard/income-gst-statement.scm
+++ b/gnucash/report/reports/standard/income-gst-statement.scm
@@ -226,6 +226,7 @@ with *EUGOODS* in the account description."))
   (GncOptionDBPtr-make-internal options gnc:pagename-display "Amount")
   (GncOptionDBPtr-make-internal options gnc:pagename-display "Sign Reverses")
   (GncOptionDBPtr-make-internal options gnc:pagename-display "Running Balance")
+  (GncOptionDBPtr-make-internal options gnc:pagename-display "Running Totals")
   ;; No multilines allowed
   (GncOptionDBPtr-make-internal options gnc:pagename-display "Detail Level")
   (GncOptionDBPtr-make-internal options pagename-sorting "Show Informal Debit/Credit Headers")
--- a/gnucash/report/reports/standard/test/test-transaction.scm
+++ b/gnucash/report/reports/standard/test/test-transaction.scm
@@ -487,7 +487,7 @@
          (set-option! options "Display" name #f))
        (list "Date" "Reconciled Date" "Num" "Description" "Memo" "Notes"
              "Account Name" "Other Account Name" "Shares" "Price" "Account Balance"
-             "Totals"))
+             "Grand Total"))
       (let ((sxml (options->sxml options "all columns off")))
         (test-assert "all display columns off, except amount and subtotals are enabled, there should be 2 columns"
           (= (length ((sxpath '(// (table 1) // (tr 1) // th)) sxml))
@@ -545,7 +545,7 @@
          (set-option! options "Display" name #t))
        (list "Date" "Reconciled Date" "Num" "Description" "Memo" "Notes"
              "Account Name" "Other Account Name" "Shares" "Price" "Account Balance"
-             "Totals" "Use Full Other Account Name" "Use Full Account Name"))
+             "Grand Total" "Use Full Other Account Name" "Use Full Account Name"))
       (set-option! options "Display" "Running Totals" 'grand)
       (let* ((sxml (options->sxml options "all columns on")))
         (test-equal "all display columns on, displays correct columns"
@@ -571,7 +571,7 @@
       (set-option! options "General" "Start Date" (cons 'absolute (gnc-dmy2time64 13 02 1971)))
       (set-option! options "General" "End Date" (cons 'absolute (gnc-dmy2time64 15 02 1971)))
       (set-option! options "Display" "Detail Level" 'multi-line)
-      (set-option! options "Display" "Totals" #f)
+      (set-option! options "Display" "Grand Total" #f)
       (set-option! options "Sorting" "Primary Subtotal" #f)
       (set-option! options "Sorting" "Primary Subtotal for Date Key" 'none)
       (set-option! options "Sorting" "Secondary Subtotal" #f)
@@ -735,7 +735,7 @@
       (set-option! options "General" "End Date" (cons 'absolute (gnc-dmy2time64 31 12 1970)))
       ;;(set-option! options "Accounts" "Accounts" (gnc-account-get-descendants (gnc-account-get-root bank)))
       (set-option! options "Accounts" "Accounts" (list bank))
-      (set-option! options "Display" "Totals" #f)
+      (set-option! options "Display" "Grand Total" #f)
       (set-option! options "Display" "Other Account Name" #t)
       (set-option! options "Filter" "Void Transactions" 'both)
       (set-option! options "Sorting" "Primary Subtotal" #f)
@@ -797,7 +797,7 @@
       (set-option! options "Sorting" "Primary Subtotal" #t)
       (set-option! options "Sorting" "Secondary Key" 'date)
       (set-option! options "Sorting" "Secondary Subtotal for Date Key" 'monthly)
-      (set-option! options "Display" "Totals" #t)
+      (set-option! options "Display" "Grand Total" #t)
       (set-option! options "Sorting" "Secondary Subtotal for Date Key" 'quarterly)
       (set-option! options "Sorting" "Show subtotals only (hide transactional data)" #t)
       (let* ((sxml (options->sxml options "sorting=account-name, date-quarterly, subtotals only")))
@@ -807,7 +807,7 @@
 
       (set! options (default-testing-options))
       (set-option! options "Accounts" "Accounts" (gnc-account-get-descendants (gnc-account-get-root bank)))
-      (set-option! options "Display" "Totals" #t)
+      (set-option! options "Display" "Grand Total" #t)
       (set-option! options "Display" "Amount" 'double)
       (set-option! options "Currency" "Show original currency amount" #t)
       (set-option! options "General" "Table for Exporting" #f)
@@ -827,7 +827,7 @@
           (get-row-col sxml 91 #f)))
 
       (set-option! options "Accounts" "Accounts" (list bank))
-      (set-option! options "Display" "Totals" #f)
+      (set-option! options "Display" "Grand Total" #f)
       (set-option! options "Sorting" "Show subtotals only (hide transactional data)" #t)
       (let* ((sxml (options->sxml options "sorting=date quarterly")))
         (test-equal "quarterly subtotals are correct"
@@ -868,7 +868,7 @@
 
       ;; test running subtotals
       (set! options (default-testing-options))
-      (set-option! options "Display" "Totals" #f)
+      (set-option! options "Display" "Grand Total" #f)
       (set-option! options "Display" "Running Totals" 'sub)
       (set-option! options "Sorting" "Primary Key" 'account-name)
       (set-option! options "Sorting" "Primary Subtotal" #t)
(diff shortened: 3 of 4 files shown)


gnucash/report/trep-engine.scm
```diff
@@ -1676,7 +1827,19 @@ be excluded from periodic reporting.")
...
+             (reversible-account? (acc-reverse? account))
+             ;; Returns the number of required arguments of a function
+             (fn-arity (lambda (fn) (car (procedure-minimum-arity fn))))
+             ;; Calls the calculator-function for the provided cell with either
+             ;; two arguments (calculator-function split transaction-row?) or only
+             ;; one argument  (calculator-function split) for legacy support.
+             (calc-fn (lambda (cell split transaction-row?)
+                              (if (= (fn-arity (vector-ref cell 1)) 1)
+                                  ((vector-ref cell 1) split)
+                                  ((vector-ref cell 1) split transaction-row?))))
```
christopherlam: I wonder if it may be simplified to:
```scheme
  (calc-fn (lambda (cell . rest) (apply (vector-ref cell 1) rest)))
```

EDIT: no.

dawansv: Thanks for the feedback! I will check that. I love it when you can turn several lines of my code into a single line. Scheme is pretty fascinating.

gnucash/report/trep-engine.scm
```diff
@@ -1676,7 +1827,19 @@ be excluded from periodic reporting.")
...
+             ;; Do any of the displayed columns have a subtotal flag?
+             (any-col-with-subtotals?
+               (< 0 (length(filter cell-with-subtotals? cell-calculators)))))
```
christopherlam: We don't like `length` - it is O(N) rather than O(1). This one I think can be optimized to `(any cell-with-subtotals? cell-calculators)` which is still O(N) but can terminate early.

christopherlam:
Overall the code is fine in itself, however, it is not desirable to add too many options in such a complicated report. What about a single "Show separate primary/secondary/total subtotals?" option somewhere in the Sorting tab?

gnucash/report/trep-engine.scm
christopherlam: In this case - I don't think it's necessary to define it here -- it's best to inline it into line 1867 so that the list traversal is only performed when necessary.

```scheme
                           (if (and opt-use-links? (or (cell-with-subtotals? cell)
                                                       (not (any cell-with-subtotals? cell-calculators))))
```

gnucash/report/trep-engine.scm
christopherlam: Didn't work. Here's a way to _upgrade_ the cell calculator without increasing arity. I'm not sure at all what's the clearest future-proof method.

```patch
index d5c2a4a71e..5b60abdc5a 100644
@@ -1387,36 +1387,36 @@ be excluded from periodic reporting.")
                                           (gnc:gnc-monetary-amount (amt-fn s)))
                                         (coll 'getmonetary
                                           (gnc:gnc-monetary-commodity (amt-fn s)) #f)))
-           (converted-running-total (lambda (s tr?)
+           (converted-running-total (lambda (s) (lambda (tr?)
                                       (and tr?
                                         (add-split-to-collector s
                                           converted-running-total-collector
-                                          converted-amount))))
-           (converted-running-prime (lambda (s tr?)
+                                          converted-amount)))))
+           (converted-running-prime (lambda (s) (lambda (tr?)
                                       (and tr?
                                         (add-split-to-collector s
                                           converted-running-prime-collector
-                                          converted-amount))))
-           (converted-running-sec (lambda (s tr?)
+                                          converted-amount)))))
+           (converted-running-sec (lambda (s) (lambda (tr?)
                                       (and tr?
                                         (add-split-to-collector s
                                           converted-running-sec-collector
-                                          converted-amount))))
-           (original-running-total (lambda (s tr?)
+                                          converted-amount)))))
+           (original-running-total (lambda (s) (lambda (tr?)
                                       (and tr?
                                         (add-split-to-collector s
                                           original-running-total-collector
-                                          original-amount))))
-           (original-running-prime (lambda (s tr?)
+                                          original-amount)))))
+           (original-running-prime (lambda (s) (lambda (tr?)
                                       (and tr?
                                         (add-split-to-collector s
                                           original-running-prime-collector
-                                          original-amount))))
-           (original-running-sec (lambda (s tr?)
+                                          original-amount)))))
+           (original-running-sec (lambda (s) (lambda (tr?)
                                       (and tr?
                                         (add-split-to-collector s
                                           original-running-sec-collector
-                                          original-amount)))))
+                                          original-amount))))))
         (append
          ;; each column will be a vector
          ;; (vector heading
@@ -1828,19 +1828,15 @@ be excluded from periodic reporting.")
     (define (add-split-row split cell-calculators row-style transaction-row?)
       (let* ((account (xaccSplitGetAccount split))
              (reversible-account? (acc-reverse? account))
-             ;; Returns the number of required arguments of a function
-             (fn-arity (lambda (fn) (car (procedure-minimum-arity fn))))
-             ;; Calls the calculator-function for the provided cell with either
-             ;; two arguments (calculator-function split transaction-row?) or only
-             ;; one argument  (calculator-function split) for legacy support.
-             (calc-fn (lambda (cell split transaction-row?)
-                              (if (= (fn-arity (vector-ref cell 1)) 1)
-                                  ((vector-ref cell 1) split)
-                                  ((vector-ref cell 1) split transaction-row?))))
-             ;; Do any of the displayed columns have a subtotal flag?
              (any-col-with-subtotals?
                (< 0 (length(filter cell-with-subtotals? cell-calculators)))))
 
+        (define (calc-fn cell split transaction-row?)
+          (match ((vector-ref cell 1) split)
+            ((? procedure? proc) (proc transaction-row?))
+            (mon mon)))
+
         (unless (column-uses? 'subtotals-only)
           (gnc:html-table-append-row/markup!
            table row-style
```

dawansv: Ha here is that match again! I like it but one issue I see is that it creates some unusual structure in the calculated-function definitions. I mean the double (lambda (s) (lambda (tr?)) instead of the more traditional (lambda (s tr?)). Since the calculated-functions are also part of custom reports, etc., it puts the onus on whoever is creating these custom reports to maintain this strange structure. 

Considering this complaint in the roadmap:
> Scheme is impenetrable to most programmers. Expecting users to be able to write reports in Scheme is completely unreasonable.

I don't know whether it's a good idea to make it look even more funky especially the custom definition part. Also in the long term, if eventually we deprecate support for one argument, then we are left with a bunch of existing custom-functions with this strange double lambda structure which we don't even need anymore.

My version (although not as clever) has the benefit of putting the code to support the 2 types of calculator-functions entirely in add-split-row, hidden from whoever is designing the custom reports. The calculator-functions themselves follow a more standard pattern. The only inconvenience of my version I can see is that it introduces the guile-specific function. I really can't judge how much of a problem it is using that function. I don't really know enough to make that sort of call of 'pure' scheme vs guile. I suppose it comes down to whether there is ever a chance we keep scheme without guile? I mean we are already using guile modules, right? For instance the match function, that already comes from a guile-specific module ice-9 , no?

So we probably need to see if others want to pitch in here... There is also my alternative implementation where we simply keep one argument and disable the running totals in the multiline context. I think it's a bit of a missed opportunity but I would be OK with it as well, not a big deal.

On a completely side note RE: the complaint in the roadmap, yes scheme is a bit unusual and possibly overkill for this, but to be honnest if we make the effort of putting more comments in these scm files including some explanation of scheme clever but unusual structures when we use them, I think it might help maintanability and invite more people to participate.

gnucash/report/trep-engine.scm
dawansv: Nice one! And funny too because I even wrote the any in my original code (as part of the function name). It's like I was 99% there but couldn't quite make it. Will update.

dawansv:
> Overall the code is fine in itself, however, it is not desirable to add too many options in such a complicated report. What about a single "Show separate primary/secondary/total subtotals?" option somewhere in the Sorting tab?

Yes the option dialog is getting crowded. That said keep in mind the total, subtotal options are already there. I am only adding 3 checkboxes, one each under their respective, already existing, totals or subtotal checkbox.

Now I can certainly look at reorganizing or combining some of these if that's OK (I was trying to reduce UI changes). For instance the Totals / runing total on the display could be a single drop down list instead of 2 checkboxes. The drop-down list would be called Totals and have 3 options: None, Totals, Running Total. Running totals automatically enables the totals line.

It could be similar for the subtotals except that then we have to deal with the "Primary/Secondary Subtotal to Date Key". So I would need to think more about how that would work.

gnucash/report/trep-engine.scm
christopherlam: My patch was simply a pure scheme-based idea.

Please hold off on it now and we'll think of a neat way to upgrade the trep-engine while maintaining backward compatibility, without needing to rely too much on guile's specific functions. Maybe the calculated-cells can be changed from a list to record with a `version` field and a `columns` list.

FWIW (ice-9 match) is macro-based and widely adopted by guile maintainers, and is pretty much standardised as SRFI-204 across many lisp implementations.

christopherlam:
Ok #1618 merged in this one just needs rebasing, refactoring and removal of arity code...

dawansv:
> Ok #1618 merged in this one just needs rebasing, refactoring and removal of arity code...

Great. I have rebased and updated my code. Tested and it all seems to work fine.

Also udated income-gst-statement.scm to disable runing total options in that context (that report uses all trep options by default, then removes the onces it doesn't one)

I still need to work on making some tests for the running totals

For your comment about the option dialog, I looked at simplifying the dialog but considering I am only adding 3 checkboxes (one under each of the 3 existing total/subtotal checkboxes) I am not sure I can simplify much. In particular the fact that the exsting subtotal options are already split between a checkbox and a drop-down list when the key is a date (to select the periods to subtotal: daily, weekly, monthly, etc) makes it difficult to combine the existing subtotal and new running subtotal options. So unless you have a more specific idea, then I don't see a way to improve this without redesigning the whole option panel which is not something I can embark on right now.

dawansv:
FYI still need to rebase with your more recent changes to left-columns. Will do that later as it's bed time for me.

dawansv:
I am trying to create a rather simple test from this html file (remove .log because I can't attached html here for some reason)
[test-trep-running-total.html.log](https://github.com/Gnucash/gnucash/files/11360643/test-trep-running-total.html.log)

The test simply grabs the last running total to make sure it equals the total of the test. It's -4 because there are 3 more rows for totals after that.

```
      ;; test running total
      (set! options (default-testing-options))
      (set-option! options "Display" "Running Total" #t)
      (let* ((sxml (options->sxml options "running total")))
        (test-equal "Last running total is $2,280.00"
          '("$2,280.00")
          (get-row-col sxml -4 -1)))
      )
```

Yet I keep getting this error, it returns #f instead of $2,880.00

> [fail] line:708, test: Last running total is $2,280.00
> transaction.scm/display options
>  -> expected: ("$2,280.00")
>  -> obtained: #f

Here is the kicker however, If I try row -5, the second to last line, I have no problem whatsover

```
      ;; test running total
      (set! options (default-testing-options))
      (set-option! options "Display" "Running Total" #t)
      (let* ((sxml (options->sxml options "running total")))
        (test-equal "Second-to-last running total is $2,302.00"
          '("$2,302.00")
          (get-row-col sxml -5 -1)))
      )
```

returns

> [pass] line:708, test: Second-to-last running total is $2,302.00

I am puzzled to say the least...

christopherlam:
The row/col analyses the xml tree exactly as written, rather than the rendered html table. This means a cell which spans 2 column will count as 1. To increase visibility of the xml try #f in row or col. ` (get-row-col sxml -5 #f)`.

Would it work better if the option "Table for export" is enabled?

dawansv:
### Summary

Here are a bunch of tests I ran with various results, but you can probably skip most of it and go to the last one at the bottom where I disable all totals, subtotals and runing total and yet I still get this odd behavior: every minus EVEN row fails and every minus ODD row works. 

This is confirmed by my earlier tests where it just happened that the row I wanted was a minus EVEN, but when I removed one subtotal line and it moved to a minus ODD then suddently it works. Same thing happened with the columns, -2 (even) fails but -1 and -3 (odd) works. 

Of note is that testing from TOP or LEFT (no minuses) works on all rows.

### This is with table for exporting option on (makes no difference)

Notice (after the fact for me) how it's a MINUS EVEN that fails (-4)

```
      ;; test running total
      (set! options (default-testing-options))
      (set-option! options "Display" "Running Total" #t)
      (set-option! options "General" "Table for Exporting" #t)
      (let* ((sxml (options->sxml options "running total")))
        (test-equal "Second-to-last running total is $2,302.00"
          '("(get-row-col sxml -5 #f)")
          (get-row-col sxml -5 #f))
        (test-equal "Last running total is $2,280.00"
          '("(get-row-col sxml -4 #f)")
          (get-row-col sxml -4 #f))
        (test-equal "Bank total row"
          '("(get-row-col sxml -3 #f)")
          (get-row-col sxml -3 #f)))
      )
```
> [fail] line:708, test: Second-to-last running total is $2,302.00
> transaction.scm/display options
>  -> expected: ("(get-row-col sxml -5 #f)")
>  -> obtained: ("12/09/23" "$109 income" "Root.Asset.Bank" "$109.00" "$2,302.00")
> [fail] line:711, test: Second-to-last running total is $2,280.00
> transaction.scm/display options
>  -> expected: ("(get-row-col sxml -4 #f)")
>  -> obtained: #f
> [fail] line:714, test: Bank total row
> transaction.scm/display options
>  -> expected: ("(get-row-col sxml -3 #f)")
>  -> obtained: ("Total For Bank" "$2,280.00")

### Testing from the top, it works

Exact same code but coming from the top (row 36 instead of -5, etc.) it works.

```
      ;; test from the top
      (set! options (default-testing-options))
      (set-option! options "Display" "Running Total" #t)
      (set-option! options "General" "Table for Exporting" #t)
      (let* ((sxml (options->sxml options "running total")))
        (test-equal "Second-to-last running total is $2,302.00"
          '("(get-row-col sxml 36 #f)")
          (get-row-col sxml 36 #f))
        (test-equal "Last running total is $2,280.00"
          '("(get-row-col sxml 37 #f)")
          (get-row-col sxml 37 #f))
        (test-equal "Bank total row"
          '("(get-row-col sxml 38 #f)")
          (get-row-col sxml 38 #f)))
```

> [fail] line:724, test: Second-to-last running total is $2,302.00
> transaction.scm/display options
>  -> expected: ("(get-row-col sxml 36 #f)")
>  -> obtained: ("12/09/23" "$109 income" "Root.Asset.Bank" "$109.00" "$2,302.00")
> [fail] line:727, test: Last running total is $2,280.00
> transaction.scm/display options
>  -> expected: ("(get-row-col sxml 37 #f)")
>  -> obtained: ("12/15/23" "$22 expense" "Root.Asset.Bank" "-$22.00" "$2,280.00")
> [fail] line:730, test: Bank total row
> transaction.scm/display options
>  -> expected: ("(get-row-col sxml 38 #f)")
>  -> obtained: ("Total For Bank" "$2,280.00")

### Retrieving columns, second-to-last fails

(Here again MINUS EVEN fails)

```
      ;; testing columns
      (set! options (default-testing-options))
      (set-option! options "Display" "Running Total" #t)
      (set-option! options "General" "Table for Exporting" #t)
      (let* ((sxml (options->sxml options "running total")))
        (test-equal "Third column from the end"
          '("(get-row-col sxml #f -3)")
          (get-row-col sxml #f -3))
        (test-equal "Second-to-last column"
          '("(get-row-col sxml #f -2)")
          (get-row-col sxml #f -2))
        (test-equal "Last column"
          '("(get-row-col sxml #f -1)")
          (get-row-col sxml #f -1)))
      )
```

> [fail] line:739, test: Third column from the end
> transaction.scm/display options
>  -> expected: ("(get-row-col sxml #f -3)")
>  -> obtained: ("Root.Asset.Bank" "Root.Asset.Bank" "Root.Asset.Bank" "Root.Asset.Bank" "Root.Asset.Bank" "Root.Asset.Bank" "Root.Asset.Bank" "Root.Asset.Bank" "Root.Asset.Bank" "Root.Asset.Bank" "Root.Asset.Bank" "Root.Asset.Bank" "Root.Asset.Bank" "Root.Asset.Bank" "Root.Asset.Bank" "Root.Asset.Bank" "Root.Asset.Bank" "Root.Asset.Bank" "Root.Asset.Bank" "Root.Asset.Bank" "Root.Asset.Bank" "Root.Asset.Bank" "Root.Asset.Bank" "Root.Asset.Bank" "Root.Asset.Bank" "Root.Asset.Bank" "Root.Asset.Bank" "Root.Asset.Bank" "Root.Asset.Bank" "Root.Asset.Bank" "Root.Asset.Bank" "Root.Asset.Bank" "Root.Asset.Bank" "Root.Asset.Bank" "Root.Asset.Bank" "Root.Asset.Bank")
> [fail] line:742, test: Second-to-last column
> transaction.scm/display options
>  -> expected: ("(get-row-col sxml #f -2)")
>  -> obtained: #f
> [fail] line:745, test: Last column
> transaction.scm/display options
>  -> expected: ("(get-row-col sxml #f -1)")
>  -> obtained: ("$103.00" "$212.00" "$190.00" "$293.00" "$402.00" "$380.00" "$483.00" "$592.00" "$570.00" "$673.00" "$782.00" "$760.00" "$863.00" "$972.00" "$950.00" "$1,053.00" "$1,162.00" "$1,140.00" "$1,243.00" "$1,352.00" "$1,330.00" "$1,433.00" "$1,542.00" "$1,520.00" "$1,623.00" "$1,732.00" "$1,710.00" "$1,813.00" "$1,922.00" "$1,900.00" "$2,003.00" "$2,112.00" "$2,090.00" "$2,193.00" "$2,302.00" "$2,280.00")

### Retrieving columns from the left works

> [fail] line:739, test: Third column from the end
> transaction.scm/display options
>  -> expected: ("(get-row-col sxml #f 5)")
>  -> obtained: ("Root.Asset.Bank" "Root.Asset.Bank" "Root.Asset.Bank" "Root.Asset.Bank" "Root.Asset.Bank" "Root.Asset.Bank" "Root.Asset.Bank" "Root.Asset.Bank" "Root.Asset.Bank" "Root.Asset.Bank" "Root.Asset.Bank" "Root.Asset.Bank" "Root.Asset.Bank" "Root.Asset.Bank" "Root.Asset.Bank" "Root.Asset.Bank" "Root.Asset.Bank" "Root.Asset.Bank" "Root.Asset.Bank" "Root.Asset.Bank" "Root.Asset.Bank" "Root.Asset.Bank" "Root.Asset.Bank" "Root.Asset.Bank" "Root.Asset.Bank" "Root.Asset.Bank" "Root.Asset.Bank" "Root.Asset.Bank" "Root.Asset.Bank" "Root.Asset.Bank" "Root.Asset.Bank" "Root.Asset.Bank" "Root.Asset.Bank" "Root.Asset.Bank" "Root.Asset.Bank" "Root.Asset.Bank")
> [fail] line:742, test: Second-to-last column
> transaction.scm/display options
>  -> expected: ("(get-row-col sxml #f 6)")
>  -> obtained: ("$103.00" "$109.00" "-$22.00" "$103.00" "$109.00" "-$22.00" "$103.00" "$109.00" "-$22.00" "$103.00" "$109.00" "-$22.00" "$103.00" "$109.00" "-$22.00" "$103.00" "$109.00" "-$22.00" "$103.00" "$109.00" "-$22.00" "$103.00" "$109.00" "-$22.00" "$103.00" "$109.00" "-$22.00" "$103.00" "$109.00" "-$22.00" "$103.00" "$109.00" "-$22.00" "$103.00" "$109.00" "-$22.00" "$2,280.00" "$2,280.00")
> [fail] line:745, test: Last column
> transaction.scm/display options
>  -> expected: ("(get-row-col sxml #f 7)")
>  -> obtained: ("$103.00" "$212.00" "$190.00" "$293.00" "$402.00" "$380.00" "$483.00" "$592.00" "$570.00" "$673.00" "$782.00" "$760.00" "$863.00" "$972.00" "$950.00" "$1,053.00" "$1,162.00" "$1,140.00" "$1,243.00" "$1,352.00" "$1,330.00" "$1,433.00" "$1,542.00" "$1,520.00" "$1,623.00" "$1,732.00" "$1,710.00" "$1,813.00" "$1,922.00" "$1,900.00" "$2,003.00" "$2,112.00" "$2,090.00" "$2,193.00" "$2,302.00" "$2,280.00")

### Without subtotals

If I turn off the subtotals (so I have only 2 total rows at the bottom instead of 3), then it is STILL row -4 that fails (which is now is the second-to-last in the transaction list). My last row (previous row -4) is now -3 and it comes back just fine.

```
      ;; test running total
      (set! options (default-testing-options))
      (set-option! options "Display" "Running Total" #t)
      (set-option! options "General" "Table for Exporting" #t)
      (set-option! options "Sorting" "Primary Subtotal" #f)
      (set-option! options "Sorting" "Primary Subtotal for Date Key" 'none)
      (set-option! options "Sorting" "Secondary Key" 'none)
      (set-option! options "Sorting" "Secondary Subtotal" #f)      
      (let* ((sxml (options->sxml options "running total")))
        (test-equal "Second-to-last running total is $2,302.00"
          '("(get-row-col sxml -4 #f)")
          (get-row-col sxml -4 #f))
        (test-equal "Last running total is $2,280.00"
          '("(get-row-col sxml -3 #f)")
          (get-row-col sxml -3 #f))
        (test-equal "Bank total row"
          '("(get-row-col sxml -1 #f)")
          (get-row-col sxml -1 #f)))
      )
```

> [fail] line:713, test: Second-to-last running total is $2,302.00
> transaction.scm/display options
>  -> expected: ("(get-row-col sxml -4 #f)")
>  -> obtained: #f
> [fail] line:716, test: Last running total is $2,280.00
> transaction.scm/display options
>  -> expected: ("(get-row-col sxml -3 #f)")
>  -> obtained: ("12/15/23" "$22 expense" "Root.Asset.Bank" "-$22.00" "$2,280.00")
> [fail] line:719, test: Bank total row
> transaction.scm/display options
>  -> expected: ("(get-row-col sxml -1 #f)")
>  -> obtained: ("Grand Total" "$2,280.00")

### Final test: With no subtotals or totals

OK so forger about totals and running totals. Here is a version with all of that disabled. So just a very basic table. 
[test-trep-running-total.html.log](https://github.com/Gnucash/gnucash/files/11361885/test-trep-running-total.html.log)

Yet here we see that testing the last 6 rows, every -EVEN fails while -ODD works

```
      ;; test minus odd and even rows
      (set! options (default-testing-options))
      (set-option! options "Display" "Totals" #f)
      (set-option! options "General" "Table for Exporting" #t)
      (set-option! options "Sorting" "Primary Subtotal" #f)
      (set-option! options "Sorting" "Primary Subtotal for Date Key" 'none)
      (set-option! options "Sorting" "Secondary Key" 'none)
      (set-option! options "Sorting" "Secondary Subtotal" #f)      
      (let* ((sxml (options->sxml options "running total")))
        (test-equal "Row -6"
          '("(get-row-col sxml -6 #f)")
          (get-row-col sxml -6 #f))      
        (test-equal "Row -5"
          '("(get-row-col sxml -5 #f)")
          (get-row-col sxml -5 #f))
        (test-equal "Row -4"
          '("(get-row-col sxml -4 #f)")
          (get-row-col sxml -4 #f))
        (test-equal "Row -3"
          '("(get-row-col sxml -3 #f)")
          (get-row-col sxml -3 #f))
        (test-equal "Row -2"
          '("(get-row-col sxml -2 #f)")
          (get-row-col sxml -2 #f))          
        (test-equal "Row -1"
          '("(get-row-col sxml -1 #f)")
          (get-row-col sxml -1 #f)))
      )
```

> [fail] line:713, test: Row -6
> transaction.scm/display options
>  -> expected: ("(get-row-col sxml -6 #f)")
>  -> obtained: #f
> [fail] line:716, test: Row -5
> transaction.scm/display options
>  -> expected: ("(get-row-col sxml -5 #f)")
>  -> obtained: ("11/09/23" "$109 income" "Root.Asset.Bank" "$109.00")
> [fail] line:719, test: Row -4
> transaction.scm/display options
>  -> expected: ("(get-row-col sxml -4 #f)")
>  -> obtained: #f
> [fail] line:722, test: Row -3
> transaction.scm/display options
>  -> expected: ("(get-row-col sxml -3 #f)")
>  -> obtained: ("12/03/23" "$103 income" "Root.Asset.Bank" "$103.00")
> [fail] line:725, test: Row -2
> transaction.scm/display options
>  -> expected: ("(get-row-col sxml -2 #f)")
>  -> obtained: #f
> [fail] line:728, test: Row -1
> transaction.scm/display options
>  -> expected: ("(get-row-col sxml -1 #f)")
>  -> obtained: ("12/15/23" "$22 expense" "Root.Asset.Bank" "-$22.00")

christopherlam:
Please feel free to add your WIP test commit. I may need to get back to this 2wks from now.

dawansv:
> Please feel free to add your WIP test commit. I may need to get back to this 2wks from now.

Will do and no rush. I can also most do the test without the minus row calls but will also keep investigating a bit. I don't find a lot of existing tests using a MINUS EVEN but I did find one, test-trial-balance.scm is calling "'sxml->table-row-col sxml 1 -2 #f" so I will have a look at that to see if I can find a clue there.

gnucash/report/trep-engine.scm
dawansv: This was resolved through PR #1618 . See next commit.

dawansv:
Rebased to 5.1. Tests included. Moving this out of draft.

As for the strange issue with the "minus even" rows and columns in the test, I am not using these in my test anymore. I will submit a separate PR to investigate this further since I can reproduce it on a much simpler example. 
Edit: See PR #1621

christopherlam:
Could you possibly perhaps merge into 3 separate commits to tidy this branch for merging? (1) the triad of running balances with tests (2) trep to filter `cell-with-subtotals?` with its own test, and (3) income-gst-statement ?

dawansv:
> Could you possibly perhaps merge into 3 separate commits to tidy this branch for merging? (1) the triad of running balances with tests (2) trep to filter `cell-with-subtotals?` with its own test, and (3) income-gst-statement ?

Yes good idea.

dawansv:
OK done with merge in 3 commits.

christopherlam:
Without delving too much into the technicalities of the old vs new running total, I wonder if the number of options can be minimised by _assuming_ that if the Display option specifies Running Total, then the visibility of Primary/Secondary/Total running totals depends on the *existence* of primary/secondary sort keys. IOW if secondary sort-key is None then disable its running total. Would this help?

gjanssens:
@christopherlam asked me to voice my opinion as well :wink: as I'm known as the 'option guy' around here, typically raising concerns about adding options. I'll admit I have no strong opinion on this one as I'm not using these reports and haven't touched its code.

If I understand correctly your feature enables running balances of three levels: globally, and at each grouping level. Your intro explains what they do, but not why they are useful in this report. So I guess my first question is which use cases do you target with each of these three running sums ?

dawansv:
> If I understand correctly your feature enables running balances of three levels: globally, and at each grouping level.

Your understanding of the feature is correct except possibly for the running balance vs running total part. Strictly speaking, a running balance (at least of the context of accounting) is related to an account balance, hence much more specific; this is the way the running balance works in the report. It simply takes the account balance for each transaction. Hence why it needed to be "fixed" in terms of presentation because it is only running when the transaction report is filtering, grouping and sorting exactly the same transaction as the account registers. This is what we did already in PR #1604. Hence the running balance feature is actually quite specific to a single use case: when the transaction report matches an account register (or several if using subtotals). Otherwise it's just showing the account balance at each transaction

Running totals have nothing to do with account registers nor how the underlying data is sorted outside of the report. It starts with a list of numbers in whatever order that list is presented, in this case whatever filtering, grouping, ordering, sorting, the transaction report is giving us. I assume you already know the transaction report can literally filter, group, sort, and order transaction in almost any imaginable way without any connection to the underlying registers. I could filter all transactions with the word "blue in them in a specific selection of accounts, sorted by description ascending. 

We shall define the resulting list of transactions as (1..n). Each time there is a list of items and a total (in any context), their can be a running total. Technically. taking the list (1..n), the total is the sum of (1..n) while the running total is the sum at every point down the list; (1), sum(1..2), sum(1..3), sum(1..4), etc until sum(1..n); Hence by definition the last running total of a series is always equal to the total of the series. There isn't much more to it than that. 

> So I guess my first question is which use cases do you target with each of these three running sums ?

First there are 3 **levels** of running totals because there are 3 **levels** of totals. (edit: It's worth adding that the subtotal levels typically show more than one total each, consecutively, since they are subtotals, and the same applies to running totals). I don't think users are likely to use all 3 at once, but since we give them a choice of any of these 3 **levels** of totals it seemed to make sense to give them a choice of any of the same 3 **levels** of running totals. 

As to what the use case of any of these running totals is, I could say it's the same as the use case for the totals themselves, but extended.

For instance; I have a list of expenses (of a certain type I selected using all of the report feature I have; it doesn't have to be expenses for a given account). We had agreed only to spend $500 for all of these things I have selected. 
What's the use case for the total? Well it tells me how much we actually spent on them. Imagine it's $750. 
What's the use case for the running total? It's the same but more. It tells me how much we sent on them (the last running total is = to the total) but also where in the list we went over the $500.
```

             Amount  Running Total
Expense 1    $100    $100
Expense 2    $150    $250
Expense 3    $300    $550
Expense 4    $200    $750
TOTAL        $750
```

This is just one use case. I could come up with more but to be honest running total is just a generic concept, just as total is, I am hoping the case is made with just this.

dawansv:
> Without delving too much into the technicalities of the old vs new running total

Indeed but again there never was an old running total. There was and still is a a running account balance (different use case in a restricted filtering and sorting context).

> The visibility of Primary/Secondary/Total running totals depends on the existence of primary/secondary sort keys

Worth noting that this is already the way in one direction: the running total for a given **level** (Primary/Secondary/Grand Total) is only available if the total for that **level** is enabled.

So I am not against that idea going the other way as well, meaning if you turn running totals on, you turn it on for all **levels** where totals are enabled. It's a bit more restrictive but if that's the compromise, I am fine with it. In this case yes, instead of one additional checkbox for each level we only have one checkbox which I assume would be the one on the display panel currently only targeting the grant total.

An alternative: Still only on the display and still a single option, but a drop down list with the following options:
```
Running Total...
* None
* Grand Total Only
* Grand Total and Subtotals
```

You still lose the granularity of one subtotal level vs the other but to be honest I am not sure this is that important. I can certainly show what that one would look like (we already know what the other one with just the check box would look like since it's the same as of now on the display tab)

dawansv:
Also just to add, since we are discussing option presentations. The current (not mine) option to just turn on the Grand Total on the Display tab says "Totals". Note the plural form. I don't understand why it does say that. It only turns on the "Grand Total" which by definition there is only one of for each report. That option has no impact on the totals for the other 2 levels that are enabled on the sorting display. Personally I feel it would be best to rename that option "Grand Total" which is what it is called in the report itself.

Edit: OK on second thought the plural could refer to more than one column being totaled, hence totals as in several columns on the Grant Total line. Still, in this context, Grand Total (singular) still makes more sense. Yes several columns might have a grand total, but there is still ONE Grand Total line (hence the singular). Totals still just makes it sound as if it would control the subtotals as well which it doesn't. Anyway just my brain going in overdrive now so possibly just forget about this comment.

jralls:
It should be consistent. If you make it Grand Total you don't have to change the subtotal labels on the Sorting page.

dawansv:
I have combined all running totals options into one option at the bottom of the display tab. Also renamed Totals to Grand Total.

image

christopherlam:
Nice. Now don't forget the options renamer mechanism somewhere in the options cpp code.

dawansv:
> Nice. Now don't forget the options renamer mechanism somewhere in the options cpp code.

The option renaming is the the last commit. Actually I have a question about this. Since the rename is not report specific, what would happen if 2 reports use 2 unrelated options with the same name; then one is renamed and the other is not; or worse both are renamed differently...

jralls:
Each report gets its own optiondb instance so name collisions can't happen.

dawansv:
> Each report gets its own optiondb instance so name collisions can't happen.

OK but isn't there only one global renaming list in gnc-optiondb.cpp? Look at my last commit https://github.com/Gnucash/gnucash/pull/1617/commits/6eaa2fe49f3f6399d1f7ee32d908f89bc5805b3e ; if there was also a completely different report with a completely different options called "Totals" that I wanted to rename to "Big Totals", how and where would I differentiate that with my "Totals" to "Grand Total" renaming as it is currently coded?

jralls:
Ah, sorry I misunderstood. No, there's only one alias possible for any particular string.

There are three non-trep reports that have "Totals" options: General Journal, General Ledger, and Register. The General Journal's and Register Report Totals options don't seem to do anything, the General Ledger's option produces a new line named "Grand Totals".  One might argue that you should apply your `Grand Totals` change to the General Ledger report and (in a separate commit) remove the Totals option from the General Journal and Register reports. 

That doesn't solve the more general problem; I suppose that the right way to fix that would be to make the alias list mutable and add instances that shouldn't apply to all reports from the report that needs it. I don't think we should add that complication until we really need it.

@christopherlam?

christopherlam:
Agree the current option renamer mechanism doesn't discriminate between different reports, and I don't particularly think this is currently an issue at this time. Hence I've typically preferred to use long report option names to be as descriptive as possible in any particular report.

gnucash/report/trep-engine.scm
```diff
@@ -983,7 +985,16 @@ be excluded from periodic reporting.")
...
+      (keylist->vectorlist sign-reverse-list))
+
+    (gnc-register-multichoice-option options
+      gnc:pagename-display optname-running-totals
+      "o2" (G_ "Display running totals as per report sort order?")
+      "none"
+      (list (vector 'none  (G_ "None"))
+            (vector 'all   (G_ "Grand Total and Subtotals"))
+            (vector 'grand (G_ "Grand Total Only"))
+            (vector 'sub   (G_ "Subtotals Only")))))
```
christopherlam, line 1001: I like the ability to choose the totals with a single multichoice. But is there a use case for "Subtotals only"?

gnucash/report/trep-engine.scm
```diff
@@ -1411,6 +1484,36 @@ be excluded from periodic reporting.")
...
+                         (cons 'calc-fn converted-running-prime)
+                         (cons 'reverse-column? #f)
+                         (cons 'subtotal? #f)
+                         (cons 'start-dual-column? #f)
+                         (cons 'friendly-heading-fn (const ""))
+                         (cons 'merge-dual-column? #f)))
+             '())
+
+         (if (column-uses? 'running-grand-total)
+             (list (list (cons 'heading (header-commodity (G_ "Running Total")))
```
christopherlam: To be consistent, I'd rename these to "Running Secondary Subtotal" "Running Primary Subtotal", "Running Grand Total".

christopherlam (changes requested):

gnucash/report/trep-engine.scm
dawansv: The short answer it is the same use case when the grant total is not useful yet I want to see subtotals. That's already an option, so this one is just extended to the running totals as well. The long answer is when the report shows a series of accounts, each presented with their own primary subtotal (as per the sorting options), but the grant total of all account is not relevant because the various accounts are not part of a group -- Maybe it's a mix of asset and expense accounts or whatever else I have to review and I don't need nor want to see the grand total nor running grant total because it is neither relevant nor useful.

gnucash/report/trep-engine.scm
dawansv: Yes I went back and forth on this and don't love it either way. 'Running Secondary Subtotal" "Running Primary Subtotal", "Running Grand Total" is nice because consistent. However it makes for long headings with unecessary info when not using all 3 . Most use case probably won't have secondary sutotals. Hence you are left with "Running Primary Subtotal", "Running Grand Total". I was just trying to be shorter. Or you just have "Running Grand Total" when these is no other total anyway so it seens redudant and looks a bit weird as a column heading (not really a problem on the last line as a row heading). 

What if I make the labels dynamic depending on what running totals are displayed?
- If only running grand total is showns with no subtotals, then we keep it as "Running Total" otherwise it will say "Running Grand Total" when subtotals are present as well (subtotals can be running or not, so it removes the confusion either way).
- For subtotals, if only primary subtotal exist without secondary, then we keep it as "Running Subtotal" otherwise "Running Primary Subtotal". 

So basically:
Just grand: "Running Total"
grand+primary: "Running Subtotal", "Running Grand Total"
all 3: as you show
primary only: "Running Subtotal"
primary and secondary: "Running Secondary Subtotal" "Running Primary Subtotal"

I don't think it adds a lot of code, just some logic on the grand and primary labels and all "column-uses" constants are already there for the needed logic. It would give us "best of both worlds". What do you think? I'll give it a try, it shouldn't be much to implement -- if it is I'll stop and consider your orignal ask.

jralls: Just be sure to include comments beginning with "Translators:" explaining what's going on so that the translators&mdash;who can't see the code without extra effort and might not be able to understand it anyway&mdash;have a chance of getting the right meanings in their own languages.

dawansv: See https://github.com/Gnucash/gnucash/pull/1617/commits/2236cd292fff825ac94707d0f999e76ff032e9d0
Much better like this I think. Will squash this commit with main one if we move forward with it

All 3:
image

Primary + Grand
image

Grand alone
image

dawansv: @jralls Are these regular comments or some other way to pass them to the translators? Also do the translator see each G_ string (pun intended I guess) separately or in context, together? Then they will see:
Running Total
Running Grand Total
Running Subtotal
Running Primary Subtotal
Running Secondary Subtotal

I suppose the comments would be for Running Total and Running Subtotal:
- Running Total is a shorter version of Running Grand Total used in the context where a Running Grand Total is displayed without subtotals.
- Running Subtotal is a shorter version of Running Primary Subtotal used when a Running Primary Subtotal is displayed without a Secondary Subtotal.

jralls: Just a regular comment beginning with the word "Translators:", see e.g. [this one](https://github.com/Gnucash/gnucash/blob/6111184c735f98da7b2f5dcf529f5660a16d6bb3/gnucash/report/trep-engine.scm#L1271) and on the [corresponding weblate page](https://hosted.weblate.org/translate/gnucash/gnucash/fr/?q=+source%3A%3DL+&sort_by=-priority%2Cposition&offset=2) you can see that they get just one string at a time. Along with the comment and context* when available there are links to the code in the String Information block, but very few translators are programmers so that seldom does them much good.

\* context in this case is a differentiator to create multiple instances of single words or short phrases that when used in different places might have different translations. In scheme the string is introduced with `(C_ "…")` instead of `(G_ "…")`.

christopherlam:
I have no further comments on this branch 👍

gnucash/report/trep-engine.scm
dawansv: See https://github.com/Gnucash/gnucash/pull/1617/commits/e928c693dfc7f7399d01a7672ae2037772eb25ae for translator comments. I will squash this as well but can somebody first check I put these where they should go (and whether they make sense)? Also I assume that if a term is used more than once, I only put the comment on the first instance.

christopherlam: The translator comments are fine.

dawansv:
I squashed these last 2 commits into the main one. I think this is good to go then.

christopherlam:
All good :)

gjanssens:
I didn't get back to this PR in time. I'll just end with a 'nice work' compliment :)

PR #1619, opened 2023-04-29
Put hostname (if available) and PID in XML backend lock file

jikamens:
Write the host name (if `gethostname` is available) and the PID of the GnuCash process into the XML backend lock file.

User story: I synchronize my GnuCash data across multiple computers using a cloud file synchronization service and run GnuCash on all of these computers. When I forget to close GnuCash on one computer and try to launch it on another, the lock file prevents me from doing that, but now I need to figure out which computer it's running on so I can close it there. Writing the host name and PID into the lock file make this easy to do, so I don't have to check each of my computers one by one to figure out where it's running.

### DIFF
--- a/libgnucash/backend/dbi/gnc-backend-dbi.hpp
+++ b/libgnucash/backend/dbi/gnc-backend-dbi.hpp
@@ -24,15 +24,8 @@
 #define GNC_BACKEND_DBI_HPP
 
 #include <dbi/dbi.h>
-#ifdef G_OS_WIN32
-#include <winsock2.h>
-#define GETPID() GetCurrentProcessId()
-#else
-#include <limits.h>
-#include <unistd.h>
-#define GETPID() getpid()
-#endif
 
+#include "../gnc-backend.hpp"
 #include <gnc-sql-backend.hpp>
 #include <gnc-sql-connection.hpp>
 
--- a/libgnucash/backend/gnc-backend.hpp
+++ b/libgnucash/backend/gnc-backend.hpp
@@ -1,0 +1,34 @@
+/********************************************************************
+ * gnc-backend.hpp: header stuff used by multiple backends          *
+ *                                                                  *
+ * This program is free software; you can redistribute it and/or    *
+ * modify it under the terms of the GNU General Public License as   *
+ * published by the Free Software Foundation; either version 2 of   *
+ * the License, or (at your option) any later version.              *
+ *                                                                  *
+ * This program is distributed in the hope that it will be useful,  *
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of   *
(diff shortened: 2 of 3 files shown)


jralls:
Adding a file at libgnucash/backend requires adding it to `backend_DIST_local` in libgnucash/backend/CMakeFiles.txt. But that's not really a good place for it, nor is leaving it as a macro. It really belongs in core-utils though there's no file there that really makes a good home for it. @gjanssens any thoughts on a good home? 

Everything should be wrapped in a function, something like `char* gnc_process_identifier()` that returns the string "hostname:pid".

Having the lock file contain the hostname and pid is only half the job. A complete solution would also add some way to get the hostname and pid into the error dialog so that the user (many of whom can't even find their data files after they've created them) doesn't need to find and open the lock file.

jikamens:
As a prelude to the comment below, let me say that as a relative stranger to the GnuCash code base, I endeavored in this PR to do as little as possible and be as minimally disruptive as possible because I don't feel comfortable making anything more than the simplest changes. With that in mind...

Regarding moving `GETPID()` into a new header file vs. turning it into a function, if you can offer me some guidance about about where to put the function I'm happy to do that, but having said that I'm not sure it's fair to burden this PR with the workload of converting something that was already implemented as a macro into a function. I understand the motivation of improving the code when it is touched, but as a stranger to the code-base I'm probably not the right person to be trying to do that, and it seems a little harsh to put that burden on me when I'm just trying to submit a relatively trivial change to add a small bit of useful functionality to the app.

Ditto for wrapping the functionality I added in a function. I don't see how that's necessary to implement the functionality I added, and it doesn't seem particularly Agile to put it into its own function until it's needed in more than one place.

Similarly, I agree that putting the info into the lock file is only half the battle, but getting the data into the error dialog would be a lot more work and a lot more disruptive to the code-base, and I don't feel like I'm knowledgeable enough about the code-base to do it. It would seem a bit churlish to reject the first half of the solution which I've submitted here, which is useful by itself, just because I'm not in a position to implement the whole solution.

I am also concerned about holding up this change until that part of it is implemented, because that would presumably require translation of the updated error message into all supported languages, which would significantly delay getting any part of the change into the app.

Having said all that, if you are able to offer me more guidance about how, within the constraints and conventions of the GnuCash code base, I would go about getting the info into the error dialog, I can give it a shot as time permits. I would ask that you consider accepting this PR in the meantime, though, so at least this first part of the functionality can be available to people.

jralls:
To be absolutely blunt: We're not interested in cheap hacks. We have far too many of them already. Either completely write a proper feature--and yes, that entails becoming familiar with the code base--or close this and maintain your hack as a private fork.

jikamens:
Wow. OK, then.

PR #1623, opened 2023-05-01
[trep-engine.scm] Proof of concept: Implement sort and subtotal by user-defined tags

dawansv:
This is a proof of concept related to the long discussed issue around behind able to use analytical dimensions. 
See https://bugs.gnucash.org/show_bug.cgi?id=113772

This proof of concept only focuses on producing a transaction report sorted and grouped by user-defined tags. 

A new tags sort key is made available in the primary and secondary key drop-down list, in the sorting options.

image

To see the transactions grouped by tags, one should select that key as well as enable the subtotal for the level selected.

More options are made available at the bottom of the sorting options.

image

### Tag Prefix

- By default, the sort engine will look for tags starting with #. However to accomodate the ability to tag multiple dimensions (meaning several tags per transaction), the user is able to change that prefix to any character (even the # could be changed). For instance you could use tags for a project dimension, for instance #P-XXX where XXX is your project reference, so #P-001 #P-002 etc. You could also use tags for different departments so #D for instance, #D-MKTING #D-FINANCE, etc. Then to run a report on the project dimension, set the tag to #P, for the department dimension set the tag to #D. 
- A checkbox allows to select the option to remove the prefix from the subtotal heading, so #D-FINANCE would simply show as FINANCE as the heading for instance (assuming #D- is entered as prefix)
- The tag search is first applied to the split memo, then if no match the transaction notes, then if no match the transaction description. Only the first matched tag (as per the prefix) is considered.

### No-Match Heading
This is the heading that will be used for the subtotal group listing transactions with no match. A checkbox allows to automatically add the prefix to it or not.

## Examples
Here is an example of a report sorted with the defaults on first level. Second level is sorted by account. I am using 2 tags in this sample file: #P-Anna and #P-Bob. Untagged transactions are listed at the bottom in the No Match # section.
image
image

Here is an example where I change the headings with these options.
image

Now my headings are a cleaner Anna, Bob and Nobody.
image
image

It is possible to remove the transactions with no matching tags by using the existing transaction filter (filter tab). Simply put the same tag prefix (# or whatever else you are using) in the Transaction Filter field.
image

Here only Anna and Bob are listed. No-match "Nobody" is no longer there.
image

Now I only want to see unmatched transactions.
image

These are just transactions that are not assigned to anybody
image

I could show Anna and Nobody, excluding Bob from the view (his transactions are excluded, not in nobody)
image

Here is just Anna and Nobody
image

Many more variations are possible, thanks to the wonderfull filter feature. 

### Credit

**The fact that this feature could be implemented with so little code is made possible by the work of all the gnucash developers that gave us the very versatile and clean structure of the transaction engine we have today. A special thanks to Christopher Lam for his contributions over the last few years in modulizing and cleaning the report as well as providing very kind and pointed advise to a new schemer.**

### DIFF
--- a/gnucash/report/trep-engine.scm
+++ b/gnucash/report/trep-engine.scm
@@ -59,6 +59,7 @@
 (use-modules (srfi srfi-9))
 (use-modules (srfi srfi-26))
 (use-modules (ice-9 match))
+(use-modules (ice-9 regex))
 
 (export gnc:trep-options-generator)
 (export gnc:trep-renderer)
@@ -98,6 +99,13 @@
 (define optname-sec-subtotal (N_ "Secondary Subtotal"))
 (define optname-sec-sortorder  (N_ "Secondary Sort Order"))
 (define optname-sec-date-subtotal (N_ "Secondary Subtotal for Date Key"))
+
+;;Tags
+(define pagename-tags (N_ "Tags"))
+(define optname-tag-prefix (N_ "Tag Prefix"))
+(define optname-remove-tp (N_ "Remove Tag Prefix from Headings"))
+(define optname-no-match-heading (N_ "No-Match Heading"))
+(define optname-append-tp (N_ "Append Tag to No-Match Heading"))
 
 ;;General
 (define optname-startdate (N_ "Start Date"))
@@ -159,6 +167,40 @@ in the Options panel."))
         (cons #\y (G_ "Reconciled"))
         (cons #\f (G_ "Frozen"))
         (cons #\v (G_ "Voided"))))
+
+;; Create a function which helps find a split's tag as per
+;; various user-defined options in pagename-tags section.
+;; Memorize both the sortvalue and rendering in hashtable
+;; for later retrieval because get-tag-from-split is slow.
+;; Returns either the sort value or rendered value as per sortvalue?
+(define (split-tag split parameters sortvalue?)
+  (let* ((tag-htable (assq-ref parameters 'tag/htable)))
+  (cond
+   ((hash-ref tag-htable split) =>
+    (lambda (found) (if sortvalue? (car found) (cadr found))))
+   (else
+    (let ((tag (get-tag-from-split split parameters)))
+      (hash-set! tag-htable split tag)
+          (if sortvalue? (car tag) (cadr tag)))))))
+
+;; Finds a split's tag as per user-defined options.
+;; Returns a list with both (tag-sortvalue tag-rendered)
+(define (get-tag-from-split split parameters)
+  (let* ((tag-prefix (assq-ref parameters 'tag/prefix))
+         (remove-tp? (assq-ref parameters 'tag/remove-tp))
+         (no-match-heading (assq-ref parameters 'tag/no-match-heading))
+         (append-tp? (assq-ref parameters 'tag/append-tp))
+         (regexp (assq-ref parameters 'tag/regexp))
+         (sm (or (regexp-exec regexp (xaccSplitGetMemo split))
+                 (regexp-exec regexp ((compose xaccTransGetNotes xaccSplitGetParent) split))
+                 (regexp-exec regexp ((compose xaccTransGetDescription xaccSplitGetParent) split))))
+         (tag-matched (and sm (match:substring sm))))
+    (cond ((not tag-matched) (list 'infinity-string
+                                    (string-append
+                                     (if (string-null? no-match-heading) "No Match" no-match-heading)
+                                     (if append-tp? (string-append " " tag-prefix) ""))))
+          (remove-tp? (list tag-matched (string-drop tag-matched (string-length tag-prefix))))
+          (else (list tag-matched tag-matched)))))
 
 (define (sortkey-list parameters)
   ;; Defines the different sorting keys, as an association-list
@@ -268,6 +310,12 @@ in the Options panel."))
               (cons 'split-sortvalue (compose xaccTransGetNotes xaccSplitGetParent))
               (cons 'text (G_ "Notes"))
               (cons 'renderer-fn (compose xaccTransGetNotes xaccSplitGetParent)))
+
+        (list 'tags
+              (cons 'sortkey #f)
+              (cons 'split-sortvalue (lambda (s) (split-tag s parameters #t)))
+              (cons 'text (G_ "Tags"))
+              (cons 'renderer-fn (lambda (s) (split-tag s parameters #f))))
 
         (list 'none
               (cons 'sortkey '())
@@ -1000,6 +1048,36 @@ be excluded from periodic reporting.")
             (vector 'all   (G_ "Grand Total and Subtotals"))
             (vector 'grand (G_ "Grand Total Only"))
             (vector 'sub   (G_ "Subtotals Only")))))
+
+  ;; Tags options
+
+  (gnc-register-string-option options
+    pagename-tags optname-tag-prefix
+    "a"
+    (G_ "Use this option along with the Tags primary or secondary key \
+and subtotals in the sorting options. \
+By default tags start with # but you can use any other prefix of any \
+length starting with any character. \
+The report engine will attempt to find a match in the split memo first, \
+then the transaction notes, and finally the transaction description. \
+The pattern is case sensitive, so it will match uppercase and lowercase \
+letters separately.")
+    "#")
+
+  (gnc-register-simple-boolean-option options
+    pagename-tags optname-remove-tp
+    "c" (G_ "Remove tag prefix from the subtotal headings displayed in report?") #f)
+
+  (gnc-register-string-option options
+    pagename-tags optname-no-match-heading
+    "e"
+    (G_ "The heading that should be displayed for the subtotal group that \
+contains the transactions with no matching tags. Default is 'No Match'.")
+    "No Match")
+
+  (gnc-register-simple-boolean-option options
+    pagename-tags optname-append-tp
+    "g" (G_ "Append tag prefix to no-match heading specified above?") #t)
 
   ;; this hidden option will toggle whether the default
   ;; qof-query is run, or a different query which ensures
@@ -2219,6 +2297,10 @@ be excluded from periodic reporting.")
                                (memq (opt-val gnc:pagename-display (N_ "Amount"))
                                      '(single double))))
          (infobox-display (opt-val gnc:pagename-general optname-infobox-display))
+         (sort-by-tags? (or (eq? primary-key 'tags) (eq? secondary-key 'tags)))
+         (tag-prefix-raw (opt-val pagename-tags optname-tag-prefix))
+         (tag-prefix (if (string-null? (string-trim tag-prefix-raw))
+                      "#" (string-trim tag-prefix-raw)))
          (query (qof-query-create-for-splits)))
 
     ;; define a preprocessed alist of report parameters.
@@ -2300,6 +2382,21 @@ be excluded from periodic reporting.")
                 (cons 'sort-account-description
                       (opt-val pagename-sorting (N_ "Show Account Description")))
                 (cons 'informal-headers (opt-val pagename-sorting optname-show-informal-headers))
+                ;; parameters based on tag options
+                (cons 'tag/prefix (and sort-by-tags? tag-prefix))
+                (cons 'tag/remove-tp (and sort-by-tags?
+                                    (opt-val pagename-sorting optname-remove-tp)))
+                (cons 'tag/no-match-heading (and sort-by-tags?
+                                          (opt-val pagename-sorting optname-no-match-heading)))
+                (cons 'tag/append-tp (and sort-by-tags? (opt-val pagename-sorting optname-append-tp)))
+                (cons 'tag/regexp (and sort-by-tags?
+                                (make-regexp
+                                  (string-append
+                                    (regexp-substitute/global
+                                      #f "[#-.]|[[-^]|[?|{}]" tag-prefix
+                                      'pre (lambda (m) (string-append "\\" (match:substring m))) 'post)
+                                    "[^ ]*"))))
+                (cons 'tag/htable (and sort-by-tags? (make-hash-table)))
                 ;; Parameters based on a mix of options
                 ;; This parameter is set to #t if an account balance can be displayed
                 ;; as a running balance with a balance forward at the top.
--- a/gnucash/report/trep-engine.scm
+++ b/gnucash/report/trep-engine.scm
@@ -99,9 +99,6 @@
 (define optname-sec-subtotal (N_ "Secondary Subtotal"))
 (define optname-sec-sortorder  (N_ "Secondary Sort Order"))
 (define optname-sec-date-subtotal (N_ "Secondary Subtotal for Date Key"))
-
-;;Tags
-(define pagename-tags (N_ "Tags"))
 (define optname-tag-prefix (N_ "Tag Prefix"))
 (define optname-remove-tp (N_ "Remove Tag Prefix from Headings"))
 (define optname-no-match-heading (N_ "No-Match Heading"))
@@ -169,7 +166,7 @@ in the Options panel."))
         (cons #\v (G_ "Voided"))))
 
 ;; Create a function which helps find a split's tag as per
-;; various user-defined options in pagename-tags section.
+;; various user-defined options in pagename-sorting section.
 ;; Memorize both the sortvalue and rendering in hashtable
 ;; for later retrieval because get-tag-from-split is slow.
 ;; Returns either the sort value or rendered value as per sortvalue?
@@ -744,7 +741,9 @@ be excluded from periodic reporting.")
              (sec-sortkey-enabled (not (eq? sec-sortkey 'none)))
              (sec-sortkey-subtotal-enabled
               (SUBTOTAL-ENABLED? sec-sortkey parameters))
-             (sec-date-sortingtype-enabled (memq sec-sortkey DATE-SORTING-TYPES)))
+             (sec-date-sortingtype-enabled (memq sec-sortkey DATE-SORTING-TYPES))
+             (tag-sortingtype-enabled
+              (or (eq? prime-sortkey 'tags) (eq? sec-sortkey 'tags))))
 
         (gnc-optiondb-set-option-selectable-by-name
          options pagename-sorting optname-prime-subtotal
@@ -802,7 +801,23 @@ be excluded from periodic reporting.")
 
         (gnc-optiondb-set-option-selectable-by-name
          options pagename-sorting optname-sec-date-subtotal
-         sec-date-sortingtype-enabled)))
+         sec-date-sortingtype-enabled)
+
+        (gnc-optiondb-set-option-selectable-by-name
+         options pagename-sorting optname-tag-prefix
+         tag-sortingtype-enabled)
+
+         (gnc-optiondb-set-option-selectable-by-name
+         options pagename-sorting optname-remove-tp
+         tag-sortingtype-enabled)
+
+         (gnc-optiondb-set-option-selectable-by-name
+         options pagename-sorting optname-no-match-heading
+         tag-sortingtype-enabled)
+
+         (gnc-optiondb-set-option-selectable-by-name
+         options pagename-sorting optname-append-tp
+         tag-sortingtype-enabled)))
 
     ;; primary sorting criterion
     (gnc-register-multichoice-callback-option options
@@ -905,7 +920,37 @@ be excluded from periodic reporting.")
       pagename-sorting optname-sec-sortorder
       "i" (G_ "Order of Secondary sorting.")
       "ascend"
-      ascending-choice-list))
+      ascending-choice-list)
+
+    ;; Tags options
+
+    (gnc-register-string-option options
+      pagename-sorting optname-tag-prefix
+      "k1"
+      (G_ "Use this option along with the Tags primary or secondary key \
+and subtotals in the sorting options. \
+By default tags start with # but you can use any other prefix of any \
+length starting with any character. \
+The report engine will attempt to find a match in the split memo first, \
+then the transaction notes, and finally the transaction description. \
+The pattern is case sensitive, so it will match uppercase and lowercase \
+letters separately.")
+      "#")
+
+    (gnc-register-simple-boolean-option options
+      pagename-sorting optname-remove-tp
+      "k2" (G_ "Remove tag prefix from the subtotal headings displayed in report?") #f)
+
+    (gnc-register-string-option options
+      pagename-sorting optname-no-match-heading
+      "k3"
+      (G_ "The heading that should be displayed for the subtotal group that \
+contains the transactions with no matching tags. Default is 'No Match'.")
+      "No Match")
+
+    (gnc-register-simple-boolean-option options
+      pagename-sorting optname-append-tp
+      "k4" (G_ "Append tag prefix to no-match heading specified above?") #t))
 
   ;; Display options
 
@@ -1048,36 +1093,6 @@ be excluded from periodic reporting.")
             (vector 'all   (G_ "Grand Total and Subtotals"))
             (vector 'grand (G_ "Grand Total Only"))
             (vector 'sub   (G_ "Subtotals Only")))))
-
-  ;; Tags options
-
-  (gnc-register-string-option options
-    pagename-tags optname-tag-prefix
-    "a"
-    (G_ "Use this option along with the Tags primary or secondary key \
-and subtotals in the sorting options. \
-By default tags start with # but you can use any other prefix of any \
-length starting with any character. \
-The report engine will attempt to find a match in the split memo first, \
-then the transaction notes, and finally the transaction description. \
-The pattern is case sensitive, so it will match uppercase and lowercase \
-letters separately.")
-    "#")
-
-  (gnc-register-simple-boolean-option options
-    pagename-tags optname-remove-tp
-    "c" (G_ "Remove tag prefix from the subtotal headings displayed in report?") #f)
-
-  (gnc-register-string-option options
-    pagename-tags optname-no-match-heading
-    "e"
-    (G_ "The heading that should be displayed for the subtotal group that \
-contains the transactions with no matching tags. Default is 'No Match'.")
-    "No Match")
-
-  (gnc-register-simple-boolean-option options
-    pagename-tags optname-append-tp
-    "g" (G_ "Append tag prefix to no-match heading specified above?") #t)
 
   ;; this hidden option will toggle whether the default
   ;; qof-query is run, or a different query which ensures
@@ -2298,7 +2313,7 @@ contains the transactions with no matching tags. Default is 'No Match'.")
                                      '(single double))))
          (infobox-display (opt-val gnc:pagename-general optname-infobox-display))
          (sort-by-tags? (or (eq? primary-key 'tags) (eq? secondary-key 'tags)))
-         (tag-prefix-raw (opt-val pagename-tags optname-tag-prefix))
+         (tag-prefix-raw (opt-val pagename-sorting optname-tag-prefix))
          (tag-prefix (if (string-null? (string-trim tag-prefix-raw))
                       "#" (string-trim tag-prefix-raw)))
          (query (qof-query-create-for-splits)))
--- a/gnucash/report/trep-engine.scm
+++ b/gnucash/report/trep-engine.scm
@@ -1032,6 +1032,14 @@ be excluded from periodic reporting.")
       (((? (cut assq <> cell)) . rest) (lp rest))
       ((fld . _) (gnc:error "field " fld " missing in cell " cell) #t))))
 
+(define (stringplus< a b)
+  (cond
+    ((eq? a 'infinity-string) #f)
+    ((eq? b 'infinity-string) #t)
+    (else (gnc:string-locale<? a b))))
+
+(define (stringplus> a b) (stringplus< b a))
+
 ;; ;;;;;;;;;;;;;;;;;;;;
 ;; Here comes the big function that builds the whole table.
 
@@ -2037,7 +2045,8 @@ be excluded from periodic reporting.")
   (cons (vector row col data) grid))
 (define (grid->html-table grid)
   (define (<? a b)
-    (cond ((string? (car a)) (gnc:string-locale<? (car a) (car b)))
+    (cond ((or (symbol? (car a)) (string? (car a)) (symbol? (car b)) (string? (car b)))
+           (stringplus< (car a) (car b)))
           ((number? (car a)) (< (car a) (car b)))
           (else (gnc:error "unknown sortvalue"))))
   (define list-of-rows (sort (delete 'row-total (grid-rows grid)) <?))
@@ -2388,8 +2397,8 @@ be excluded from periodic reporting.")
                       (lambda (s) #f))))
              (value-of-X (comparator-function split-X))
              (value-of-Y (comparator-function split-Y))
-             (op (if (string? value-of-X)
-                     (if ascend? gnc:string-locale<? gnc:string-locale>?)
+             (op (if (or (string? value-of-X) (symbol? value-of-X))
+                     (if ascend? stringplus< stringplus>)
                      (if ascend? < >))))
         (and value-of-X (op value-of-X value-of-Y))))
 


benbucksch:
> However to accomodate the ability to tag multiple dimensions (meaning several tags per transaction), the user is able to change that prefix to any character (even the # could be changed). For instance you could use tags for a project dimension, for instance #P-XXX where XXX is your project reference, so #P-001 #P-002 etc. You could also use tags for different departments so #D for instance, #D-MKTING #D-FINANCE, etc

That's a great idea. Low-key, simplistic, but allows the user to do what they need.

I do need multiple dimensions, so thanks a lot for this effort!

flywire:
My personal view is this workaround is the most significant improvement since V2. Obviously, there is no developer support for implementing tags properly as a relational table.

Suggested enhancements:

1. User customisable final group heading, default: `No prefix tag #[hashtag]`
1. Hide all tags in report

---

Example with Hashtag Prefix `#house` :

image

Sample data below without tags shown in above report which would properly require a multi-split file.

Cheque Account Data:

```
Date,Number,Description,Amount,Balance,Account,Entity
01/03/06,,Opening Balance,1000,1206.79,Equity:Opening Balance,Opening Balance
05/03/06,101,GG25j1546 Groceries    wtf 15:57 061124,-45.21,1206.79,Expense:Groceries,Old Bakery
05/03/06,,EFTPOS 5632185,-75,1206.79,Expense:Groceries,HomeTown Grocery
06/03/06,,Transfer to J&J Doe Savings Acc 5765-8397 589654259587,100,1206.79,Asset:Saving,J&J Doe
06/03/06,106,EFTPOS 98416,-100,1206.79,Expense:Hardware,ABC Hardware
14/03/06,,Direct Credit Salary from Employers R Us,670,1206.79,Income:Salary,Employers R Us
24/03/06,,Transfer to J&J Doe Savings Acc 5765-8397 589654259587,500,1206.79,Asset:Saving,J&J Doe
28/03/06,,Mmvoin515b  Internet Company bg??,-20,1206.79,Expense:Internet,FastFibre
28/03/06,102,Light Company Big City Branch   9g8k863,-78,1206.79,Expense:Electricity,Energy Plus
28/03/06,103,Phone Company Name    Autodebit 595642583,-45,1206.79,Expense:Phone,Telecom
28/03/06,104,April Rent        5 Short Road,-350,1206.79,Expense:Rent,HighTower
28/04/06,105,May Rent        5 Short Road,-350,1206.79,Expense:Rent,HighTower
```

flywire:
Tags in columns is also very useful. There could be many tags with the intention use report in a spreadsheet.

Note single column for amount, eg:

image

christopherlam:
It is a very very slippery slope to incomprehensible code if we start using a global variable. The transaction report is currently fairly hackable _because_ it doesn't use global vars, and I would recommend these must factored out before being considered for merging in.

gnucash/report/trep-engine.scm
```diff
@@ -154,6 +156,27 @@ in the Options panel."))
...
         (cons #\v (G_ "Voided"))))
 
+;; Define the hashtag-prefix globally so it can be used in the split-sortvalue and renderer
+;; procedures below; it will be set with the actual user-entered string in gnc:trep-renderer
+(define hashtag-prefix "#")
+
+;; Used by the hashtag sortkey to retrieve first matching hashtag from split
+(define (split-hashtag s)
+  (let* ((htp-trimmed (string-trim hashtag-prefix))
+         (htp (if (< 0 (string-length htp-trimmed)) htp-trimmed "#"))
```
christopherlam: instead of `(< 0 (string-length str))` prefer `(not (string-null? str))`

gnucash/report/trep-engine.scm
```diff
@@ -263,6 +286,12 @@ in the Options panel."))
               (cons 'text (G_ "Notes"))
               (cons 'renderer-fn (compose xaccTransGetNotes xaccSplitGetParent)))
 
+        (list 'hashtag
+              (cons 'sortkey #f)
+              (cons 'split-sortvalue (lambda (s) (split-hashtag s)))
```
christopherlam: a `(lambda (s) (proc s))` is simply reduced to `proc`

gnucash/report/trep-engine.scm
jralls: `(< 0 (string-length str))` means `#t`. `string-length` can't return a negative value.

dawansv:
> Suggested enhancements:
> 
> 1. User customisable final group heading, default: `No prefix tag #[hashtag]`
> 2. Hide all tags in report

These are good suggestions. Will try to add some of it in my next iteration.

dawansv:
> Tags in columns is also very useful.

Yes I have seem that type of report before as well in other systems. Good idea to keep as next step after we cover the transaction-type report first.

dawansv:
> It is a very very slippery slope to incomprehensible code if we start using a global variable.

Yes I totally agree with this, and I should have mentionned this is just a temporary hack so I can at least show proof of concept from a usability standpoint. I need to spend more time figuring out how I can make the option database available within the sortkey-list procedure. So the globial variable was just a shortcut for now.

gnucash/report/trep-engine.scm
dawansv: Thanks Christopher, I knew there had to be a better way. string-null? it is.

John, scheme is funny for sure. Everything is a function or procedure, even <. So < is a procedure that takes 2 arguments and then compares them as in arg1 < arg2, so 0 < (string-length str).  But yes when you read it in a non-scheme way it looks like it's backward...
Edit: possibly misread what John said, but still  (string-length str) doesn't need to go negative for (< 0 (string-length str)) to return #f. It returns #f when (string-length str) equals 0. Still, string-null? is a better option here anyway.

gnucash/report/trep-engine.scm
dawansv: Good point. I put a lambda because I actually need to pass a second argument to differentiate the sortvalue from the renderer. For sorting reason, I need to put a bunches of ZZZZZ in front of the sortvalue for the unclassified/no tag category. It's a bit dirty but I don't really want to have to rewrite the sort code just to push that one section to the bottom. But the renderer should not show it (it does right now). So in my next iteration I will put an additional #t flag for the sortvalue (lambda (s) (split-hashtag s #t)) vs the renderer (lambda (s) (split-hashtag s #f)) so I can skip the ZZZZ for the renderer.

gnucash/report/trep-engine.scm
jralls: Darn post-fix notation. `string-null?` is definitely clearer for that reason if no other.

BTW, operators are functions in all languages. C++ even names them (e.g. `operator<()`) and lets you write your own.

dawansv: > Darn post-fix notation

I think you mean prefix notation, no? I think postfix notation would be (0 (string-length str) <). Hardly better ;-)

dawansv:
Ok so I here is my version 2. I will update my top post with some of the added features such as the ability to customize the headings by removing the tag prefix and also customize the no match heading.
More importantly it is no longer using global variables. See this commit https://github.com/Gnucash/gnucash/pull/1623/commits/7373fc1fce9631fbd4ef9068070395a2bab94c2d for changes to pass options to the sortkey-list procedure. It wasn't too difficult.

gnucash/report/trep-engine.scm
jralls: Yeah, prefix. Post-fix is Forth and HP calculators. It sort of makes sense when you're operating on an inverted stack (where you push on the bottom instead of the top). Sort of. I can do it on an HP calculator, but I haven't had to program one for almost 50 years and when you're just calculating something the comparisons are in your head, not in the calculator. I hated Forth and never even got a program finished in it.

flywire:
Comments:
1. Tag seems a more reasonable name than hashtag given that it might not even have a hash
1. The options menu is at the top for all of my reports but this version forces it to the left hand side

image

dawansv:
> * Tag seems a more reasonable name than hashtag given that it might not even have a hash

Yes I had the same thought. In fact I was about to rename it to just Tags but then I thought it might be good to frame it as hashtah because it would be best avoided to use non-descript prefixes. So the # is really good practice I think. The fact it can be changed is then a bit hidden. Maybe I am overthinking it, oviously the renaming from Hashtah to Tag is easily done but I am going to sit on it a bit. Possibly get more feedback from others on this?

> * The options menu is at the top for all of my reports but this version forces it to the left hand side

Yes it did that all on its own... Evidently when more "pages" are added eventually it shifts to the left as it runs out of room, so one more page did it here.

benbucksch:
> > * Tag seems a more reasonable name than hashtag given that it might not even have a hash

> Possibly get more feedback from others on this?

+1 on just "tag"

christopherlam:
It will be a natural progression to implement multiple tags per transaction. How would it be best handled?

#project1 #renovations2022 etc

dawansv:
> It will be a natural progression to implement multiple tags per transaction. How would it be best handled?
> 
> #project1 #renovations2022 etc

The current implementation already accomodates that. It is indeed an important feature without which it would be quite limited. You can already have several tags on one transaction as long as they are in different tag groups, meaning tags with a different prefix (they are called dimensions in analytical accounting). See what I said in my intro above:

> However to accomodate the ability to tag multiple dimensions (meaning several tags per transaction), the user is able to change that prefix to any character (even the # could be changed). For instance you could use tags for a project dimension, for instance #P-XXX where XXX is your project reference, so #P-001 #P-002 etc. You could also use tags for different departments so #D for instance, #D-MKTING #D-FINANCE, etc. Then to run a report on the project dimension, set the hashtag to #P, for the department dimension set the hasgtag to #D.

You can have both the #P-001 tag and #D-MKTING tag on the same split. When running a project report you will use the #P- prefix and when running a Department report you will use the #D- prefix. The tagged split will show up in both report. 

Note that it is different than having 2 tags (analytical codes) in the same group (dimension) on the same split. That is not allowed, not even in analytical accounting, nor does it make sense to have the same transaction counted twice in the same dimension. In this case I explain that only the first tag is considered:

> The tag search is first applied to the split memo, then if no match the transaction notes, then if no match the transaction description. Only the first matched tag (as per the prefix) is considered.

However you can split your transaction between 2 tags on the same dimension since the tagging works on the split. So you could have a dining expenses split between #P-001 and #P-002. Just create 2 splits with whatever amounts need to be applied to each (both going to dining) and tag each with the appropriate tag.

In more advanced systems, you can even split transactions on several dimensions differently. You could have a 80/20 split between 2 codes on a Project dimension, and a 50/50 split on a Department dimension. These are pretty extreme cases. We can still do that with our split system by doing a 50/30/20 split and assigning each split to add up to 80/20 for the project dimension and 50/50 for the Department dimension. These are pretty edge cases even in larger accounting system, and even much more for this simple implementation in gnucash.

flywire:
> How would it be _best_ handled?

Clearly as a relational table with separate fields like splits.

Don't let the perfect be the enemy of the good. Many things worth doing are worth doing badly. This workaround is functional with tags in Description, Notes, or Memo.

Another nice thing about this solution is it is pretty easy to convert to a relational table, even on a temporary basis, or import from fields in other software.

dawansv:
Possibly what I need to do next to explain it better is create a sample gnucash file for a typical case, such as a small nonprofit that might have several grants over time (starting and ending at different times, overlapping, and getting replaced with new grants overtime) and run 2 or 3 projects (which can also start and end at different times, overlapping, and replaced with new projects overtime) with some grants funding several projects and some projects funded by several grants. Each project and each grant eventually needs its own report of income and expenses. All of this while keeping regular general-ledger-based organization-wide accounts as gnucash already provides. Possibly then it might demystifiy analytical accounting and make it OK for those that have not encountered it. The irony of not having this feature all these years is that gnucash would actually work quite well for small nonprofits that typically only has one volunteer accountant and little money to spend on accounting software, but today probably spends hundreds (thousands?) of dollars on quickbooks or such.

flywire:
The sample data I used in https://github.com/Gnucash/gnucash/pull/1623#issuecomment-1530788858 is from the guide ([eg](https://www.gnucash.org/docs/v5/C/gnucash-guide/cbook-together1.html)) and I envisage it could easily be extended to present a brief tag section. It might be worth looking at the next level down in the guide with the splits.

christopherlam:
> Don't let the perfect be the enemy of the good. Many things worth doing are worth doing badly. This workaround is functional with tags in Description, Notes, or Memo.

Disagree completely. Gnucash had 20+ years of imperfect solutions, many half baked, and many abandoned. And while battling ever changing underlying software infrastructure.

I agree this particular branch introduces a novel functionality but maintainers will need to bear the burden of bug fixing and trying to answer queries from users in perpetuity. Are there volunteers? 🙂

dawansv:
> I agree this particular branch introduces a novel functionality but maintainers will need to bear the burden of bug fixing and trying to answer queries from users in perpetuity. Are there volunteers?

That's a good point to bring up and also one reason why I labelled this "proof of concept" because I don't necessarily want to push this like a regular update. In part we certainly need to make sure there are enough users both interested in this feature but also (some) willing to do a little bit of work in terms of maintenance, documentation, etc. We can most likely find a way to go about it to ensure enough user (and volunteer) interest before making this a permanent feature.

The reports menu already has an "experimental' section. Right now there is only one report in it. Could that be the place to introduce a new report called "Analytical Transaction Report"  or "Transactions by Tags" or such? Its location in the experimental section would make it clear for that for the time being, unless we can get some user interest in feedback, documentation, etc, then it might not go anywhere?

In the background it could take the form of either a report based on trep-engine or a completely separate report. I believe we have several possible variations here (all of which are completely transparent to end-users):

- I think (but this needs confirmation) that if we merge this first commit (with no user-facing changes) https://github.com/Gnucash/gnucash/pull/1623/commits/7373fc1fce9631fbd4ef9068070395a2bab94c2d in trep-engine then the second commit that actually implements the feature can be put in a scm file that simply redefines the options and then calls trep-engine. I only add an option in that second commit, that's all. There are no changes to trep-renderer and related functions. The only tweak might be (use-modules (ice-9 regex)) which might need to be added to trep-engine (or can an additional module be added to the other scm file?)
- We can completely split this into a separate scm file (based on my modified trep-engine), only to be reunited with trep-engine much later if it gets out of its experimental status. That's not my first choice, because forcing ourselves to maintain a common trep-engine while adding new features in scm files calling trep-engine is not necessarily a bad excersise if we can make it work. In any case these 2 options are completely transparent to the end-user.
- I maintain this in a separate branch for a custom report. Users can then grab it as a custom report with instructions on how to install already on the wiki https://wiki.gnucash.org/wiki/Custom_Reports#Loading_Your_Report. But I think this misses an opportunity to figure out how to better push new (possibly experimental) reports (and even other features) to users to test their interest and possibly gain more volunteers.

As for user interest, only time will tell. But this feature in particular is # 1 on user voice with many many comments, a lot of opinions for sure, some misunderstanding (with the caveat that users that don't like it don't have to use it). https://gnucash.uservoice.com/forums/101223-feature-request

Also has a similar long list of comments in the bug report https://bugs.gnucash.org/show_bug.cgi?id=113772

That said I agree that if not enough users step up and provide comments, with some helping with documentation, etc, then possibly it never makes it out of experimental. I don't have a problem with that.

benbucksch:
> this particular branch introduces a novel functionality but maintainers will need to bear the burden of bug fixing

This is a feature that I needed since some 15 years, and I'm missing badly since then. I had [filed this as feature request almost 15 years ago](https://bugs.gnucash.org/show_bug.cgi?id=569200). But it was simply DUPped.
I was aware at that time that this would be a major change. But I felt it was necessary anyways.

To the urgency: If I have the choice between changing my bank, or changing GnuCash, I would probably change my bank. This is how important GnuCash is to me. And [this feature, as described in my bug report](https://bugs.gnucash.org/show_bug.cgi?id=569200), is for me by *far* the most important missing feature in GnuCash. And I'm surprised how small and confined the patch here is.

Here is why I need this urgently, as a user:
* I categorize expenses by their purpose. For example, Personal | Transport | Car | Purchase and Repairs.
* Now, I also want to see how much *each* car has cost me. So, I have to introduce a new level: Personal | Transport | Car | Silver Car | Purchase and Repairs , same with "Insurance", "Gas", "Taxes" etc. for that car. This allows me to see how much a specific car has cost, all combined. But now I cannot easily see anymore how much I paid for gas for all cars combined, because it's per-car now.
* Now, I make a vacation with the car, and spend a lot on gas on that vacation. But I also want to classify that gas both to the car, *and* to the vacation. I have specific accounts for travael expenses, like Hotel, Restaurants, etc., and I want to include the gas in the travel expenses. So, I need them in the Car | Gas _and_ in Travel | Transport | Gas. I cannot do that with the current system.
* Lastly, I would like to separate our specific suppliers. For example, Project A | Hosting | Servers, and I also have Project B | Hosting | Servers, but I get both servers from Hurrican Electric. I would like to see how much money I paid to Hurricane Electric, so that I can maybe compare better offers or see just how much I paid for hosting. I cannot do that in the current system.

I classify **either** by Project, or by expense type, or supplier, but I cannot classify as all of them, and I cannot make a sum per project, per expense type, **and** per supplier.

Back to the code:

With this patch, I am surprised how small the change is (esp. in the first patch with a global variable, but even now with the current patch without global variable). I thought this feature would require substantial changes in the user interface, in the database, and in all the reports.

The fact that this feature is confined in a single file and just in reports is amazing to me, and an incredibly low code impact compared to the utility of it. I would think that this is a no-brainer to take this patch.

Additionally, you have a very cooperative contributor (@dawansv) who made massive changes to the patch based on the review requests. He has done all you asked. I think it would only be fair to merge this. Both the contributor and the users would be happy about this.

christopherlam:
@dawansv Thank you for considering. And others for further feedback.

Ideas to consider: multiple tagging with dropdown of available tags in desc/notes/memo. Becomes unfriendly when users will inevitably want to mass-rename or mass-delete obsolete tags. Tags may want their own life cycle: start date, expiry date (we aren't interested in #marketing2010 tags anymore are we?), attached desc/notes. 

A much more forward-thinking approach is to introduce a new non-kvp based TAGS table, and will be reference-counted, and attached to individual transactions or splits. This TAGS table will need UI to maintain the list, and UI in search and register. Don't forget import/export, and adding dimensions in other reports. These won't need to be introduced all at once... but the underlying data structure must be robust.

dawansv:
> Here is why I need this urgently, as a user:

Thanks Ben, it's good to hear some more practical example. The car is a good one too. For years I only had one. For complex reasons, in the last 2 years I ended up with 3. Do I really want to restructure my chart of account for that. Absolutely not, it's not that important. Yet with analytical accounting features I can do it if needed without disturbing my chart of account. Cars come and go, my chart of account stays the same. 

Your example also capture the many dimensions you might want to 'follow', some more temporaty (a trip) than others (a car) yet criss-crossing at times. Again all of this done without messing up your main chart of account. It also doesn't stop you from using the chart of account for this purpose at times. 

For instance I own rental properties, and I actually keep them in my main chart of account each under a different main income and expense subaccount with their own expense and income sub-subaccounts for each type of income and expense. I could of course use a completely separate file for each, but that's my choice. I could combine them all and use analytical codes, but that wouldn't be my choice either. Why do I do it this was vs what I would do for the car with analytical tags? Well it's matter of permanency and importance. Each property is a separate sub-business, and the question "how mich electricity do I spend accross all properties" is not one I seek to answer. On the contrary each property is accounted for separately for tax purpose at least in my jurisdiction. But even in that setup it wouldn't stop me from using analytical codes to keep track of a particular thing accross my rentals if I needed.

Bottom line is there is more than one way to do things, users have choice, and each case can be a bit different.

dawansv:
> A much more forward-thinking approach is to introduce a new non-kvp based TAGS table, and will be reference-counted, and attached to individual transactions or splits. This TAGS table will need UI to maintain the list, and UI in search and register. Don't forget import/export, and adding dimensions in other reports. These won't need to be introduced all at once... but the underlying data structure must be robust.

Yes and yes on this, and that's how it works in full implementations in other systems. But the amount of work required is probaly litterally 100+ times what I put in so far. Hence why introducing this remarkably small code change (as I said in an experimental report if needed) might help us get more appetite and actual support for a bigger implementation later. 

> Ideas to consider: multiple tagging with dropdown of available tags in desc/notes/memo. Becomes unfriendly when users will inevitably want to mass-rename or mass-delete obsolete tags. Tags may want their own life cycle: start date, expiry date (we aren't interested in #marketing2010 tags anymore are we?), attached desc/notes.

We could fetch the free tags and present them in a list. However unless we have that data structure you describe we don't know what the prefixes are, there is no hierarchy to present, and having to guess this based on tag names is not a pleasant thought. I see 2 tags #PA01 and #PA02. I see #PB01 and #PB02. Is #P a prefix for these 4 codes in one dimension? Are #PA and PB prefixes for 2 separate dimensions? Tags exists in their own time period. A old #Poolparty tag from 2010 is not conflicting with a new set of #P tags introduced in 2020 because I will run the report on that period. Only the user knows all of that for now with free tags. Yes, in a full implementation you can have a hierarchical representation, so #P is a dimension and #PA and #PB subdimensions. You can mass rename, delete, etc. You can hide old tags. But only the user knows that in an open tags system (vs what you describe above). 

Pulling a list would also I think give the impression that we are offering more than we are. I think the current changes in the autofill in the register tell us we have to be careful, because pulling old entries from 10 years ago is unlikely to be helpful (one of the current issues in that case). 

For now it's just free tags, and yes there is no way to move, rename, etc. Yes you have to enter them by hand (that's the point of free tags). It's a big downside for sure, we can disclose that. Although your example is actually one where it doesn't matter. Old tags are irrelevant because they are no longer used. If for some reason you have a #marketing2010 tag, then leave it alone. I assume you are not using it today anymore. As long as you are not showing it in a list, then the tag is irrelevant if you forgot about it, unless you remember AND are choosing a time period that includes it. 

Ironically it is long lasting tags that can be more of a problem. If you want to rename a tag because you are reorganizing your dimensions or code. But even that is a bit of an edge case. In some cases you can reuse tags knowingly without a problem, because the period of also of interest. So for instance imagine I have a #car dimension with 2 tags, #car-mazda and #car-toyota. Now year later I by a second mazda, then I realize I should probably have used a more descriptive tag for the first one. Well in this case I can't rename it, but I could decide to simply keep it as is for the first car, start using a new tag in the new fiscal year if I only care about yearly reports, etc. That said long lasting tags are really not the main example of use of analytical codes as I have described before. In practice most tend to be relatively short lives, a few years max if even.

Bottom line is that I agree this is not by any stretch a feature-rich implementation. I agree with additional features, but then they require a full implementation as you describe. If we are only using free tags as a first taste of this, then we keep it consistent and simple, and I don't think offering mass delete and mass rename with free tags is a good idea. The idea of doing a mass replace in the actual memo, notes and descriptions that doesn't sound very appealing to me. But again this only matters for long lasting tags. Most tags arent, and they become obsolete on their own as they stop being used in the next period.

If this first implementation brings more interest and support from other developpers, then we can do more. I am to rusty to go back to C++ at this point, so for now my focus is mostly on scheme because they are so many low hanging fruits to fix in the report system (a whole other chapter being improving styles and stylesheets for instance, not technically difficult but very impactful). Hence it would require more developpers to join the effort because I certainly don't want to add more work for the existing core developpers, that is really not my intention.

I mean at the very minimum I will maintain it as a custom report for now if nothing else, for others to use if they want and get some traction. But I think putting it in in the actual built as experimental is just one step up that might get more support more easily. However we could also do it in 2 stages if it's really that worrisome. Custom report for a while, trying to push it out for more user feedback via various channels, then experimental in a later release of version 5 based on feedback we receive including whether users are OK with some of the limitations for now. Then much later and only then decide if there is appetite and support for a full implementation as you describe, which I think would require a stronger set of functional requirements before we start coding and might even be a version 6 goal in its full form.

benbucksch:
> litterally 100+ times [the work] ... this remarkably small code change (as I said in an experimental report if needed) might help us get more appetite and actual support for a bigger implementation later.

+1

Dear maintainers, please provide this feature to categorize a transaction using multiple dimensions in *some* way that is reasonably easy to use, and available out of the box in the standard release. If you reject this patch, then please provide it in another way. But the functionality is really needed. It's the one feature that's missing.

christopherlam:
Couple of advice I can offer:
1. If `split->hashtag` is slow, then it can be memoized. See `gnc:make-owner->split` for example code. It looks like you would need a hash table, and memoize both values of `sortvalue?` eg `(list hashtag-with-sortvalue? hashtag-without-sortvalue?)`
2. Don't use ZZZZZZ to make the heading sort last. It's ugly and insensitive to locale-sensitive sorting. You're looking for an object which will satisfy the `gnc:string-locale<?` similarly like `<` handles `+inf.0`. The only suggestion I can consider is: the split->hashtag returns `'infinity-string` and apply the following (untested) patch:

```patch
@@ -2160,6 +2160,14 @@ be excluded from periodic reporting.")
        (else
         (string-contains str transaction-matcher))))
 
+    (define (stringplus< a b)
+      (cond
+       ((eq? a 'infinity-string) #f)
+       ((eq? b 'infinity-string) #t)
+       (else (gnc:string-locale<? a b))))
+
+    (define (stringplus> a b) (stringplus< b a))
+
     (define (generic-less? split-X split-Y sortkey date-subtotal-key ascend?)
       ;; compare splits X and Y, whereby
       ;; sortkey and date-subtotal-key specify the options used
@@ -2179,8 +2187,8 @@ be excluded from periodic reporting.")
                       (lambda (s) #f))))
              (value-of-X (comparator-function split-X))
              (value-of-Y (comparator-function split-Y))
-             (op (if (string? value-of-X)
-                     (if ascend? gnc:string-locale<? gnc:string-locale>?)
+             (op (if (or (string? value-of-X) (symbol? value-of-X))
+                     (if ascend? stringplus< stringplus>)
                      (if ascend? < >))))
         (and value-of-X (op value-of-X value-of-Y))))
```

dawansv:
> * If `split->hashtag` is slow, then it can be memoized. See `gnc:make-owner->split` for example code. It looks like you would need a hash table, and memoize both values of `sortvalue?` eg `(list hashtag-with-sortvalue? hashtag-without-sortvalue?)`

Good point. I just ran a test with 4 years worth of expenses; that's about 5000 transactions. I just used the letter E as the tag prefix since it's a pretty common letter (the tag prefix will pick a match mid word, it only needs a blank at the end). Mind you this is on a 20-yo thinkpad running Xubuntu. 
Transaction report sorted and subtotaled by Account Name: 10 seconds
Transaction report sorted and subtotaled by Tag Prefix E: 27 seconds

So not insignificant 2.7x longer. I will try that solution you are refering to and see what it gets us.

> * Don't use ZZZZZZ to make the heading sort last. It's ugly and insensitive to locale-sensitive sorting. You're looking for an object which will satisfy the `gnc:string-locale<?` similarly like `<` handles `+inf.0`. The only suggestion I can consider is: the split->hashtag returns `'infinity-string` and apply the following (untested) patch:

Yes I am not surprised you are not impressed by this!  I was trying to limit my code changes but if you are opened to changing the sort code then I think your solution is as good as it gets without adding yet another function to handle exceptions to the sort position separately from the sortvalue for each split. Will test your code and keep it in a separate commit just for that part.

flywire:
A few comments:
1. The inability to effectively utilise the strengths and resources of the many volunteers more effectively, and recognise developers are not subject matter experts, is a project leadership issue few open source projects resolve
2. Extensive use case examples have been given where GnuCash isn't an option without multidimensional reporting
3. Don't confuse this workaround with dropdown lists etc which make no sense unless tags are implemented as fields in a relational table, multidimensional functionality far exceeding most financial software
4. The effort to migrate tags in descriptions/notes/memos to relational tables would be minimal if ever required
5. Seems the speed of reporting is not affected and is irrelevant for reporting without tags
6. This report is functional and part of an open source project so the community can use it as a hack now without it ever becoming part of the main distribution

dawansv:
Latest changes:
- added the 'infinity-string fix to generic-less? function in https://github.com/Gnucash/gnucash/pull/1623/commits/64a934e7d769a37d2abb2277b8dd9286f3047194
- Updated main commit with 'infinity-string and hashtable in split-hashtag. https://github.com/Gnucash/gnucash/pull/1623/commits/ab8cc0a4dd0265b42c63c9774517de603d72ea12 Speed is much improved compared to prior test, down to 16 seconds vs 27 seconds prior

gnucash/report/trep-engine.scm
```diff
@@ -154,7 +162,51 @@ in the Options panel."))
         (cons #\f (G_ "Frozen"))
         (cons #\v (G_ "Voided"))))
 
-(define (sortkey-list split-action?)
+;; Hashtable to memorize both the sortvalue and rendering of the
+;; split's matching hashtag 
+(define hashtag-htable (make-hash-table))
```
christopherlam: this `hashtag-htable` is a global var *within* trep-engine. It means repeated calls to this trep-engine will fail to refresh the hash table. Either reset the hash table on launching trep-renderer, or (better) create a split->hashtag generator similarly to `gnc:make-split->owner` and the generator will automatically clean up when gc runs.

gnucash/report/trep-engine.scm
```diff
@@ -154,7 +162,51 @@ in the Options panel."))
...
+    (gnc-optiondb-lookup-value (gnc:optiondb options) section name))
+  (let* ((hashtag-prefix (opt-val pagename-hashtag optname-hashtag-prefix))
+         (remove-htp? (opt-val pagename-hashtag optname-remove-htp))
+         (no-match-heading (opt-val pagename-hashtag optname-no-match-heading))
+         (append-htp? (opt-val pagename-hashtag optname-append-htp))
+         (htp (if (string-null? (string-trim hashtag-prefix))
+                   "#" (string-trim hashtag-prefix)))
+         (htp-sanitized (regexp-substitute/global #f "[#-.]|[[-^]|[?|{}]" htp
+                          'pre (lambda (m) (string-append "\\" (match:substring m))) 'post))
+         (regexp (make-regexp (string-append htp-sanitized "[^ ]*")))
```
christopherlam: IIUC `make-regexp` is slow... it should be generated once, and reused within a trep-renderer session

gnucash/report/trep-engine.scm
```diff
@@ -154,7 +162,51 @@ in the Options panel."))
...
+;; various user-defined options in pagename-hashtag section.
+;; Memorize both the sortvalue and rendering in hash table
+;; for later retrieval because get-hashtag-from-split is slow. 
+;; Returns either the sort value or rendered value as per sortvalue?
+(define (split-hashtag split options sortvalue?)
+    (or (let ((tag (hash-ref hashtag-htable (gncSplitGetGUID split))))
+          (and tag (if sortvalue? (car tag) (cadr tag))))
+        (let ((tag (get-hashtag-from-split split options)))
+          (hash-set! hashtag-htable (gncSplitGetGUID split) tag)
+          (if sortvalue? (car tag) (cadr tag)))))
```
christopherlam: clearer:
```scheme
 (cond
  ((hash-ref hashtag-htable (gncSplitGetGUID split)) =>
   (lambda (found) (if sortvalue? (car found) (cadr found))))
  (else ...))
```

christopherlam (commented):
A couple more feedback. IMV "hashtag" is not a good term to use. May I suggest "Description/Notes/Memo substring" which will unlock a substring string option. Moreover I still think 4 more options is too much. Ideally only 1 additional string option (enabled only if substring search is selected) requesting the format, but I don't know how to design it without requiring the user input the regex string.

Other items to consider:
- substring search in both primary and secondary sort key. what happens? what should happen?
- documentation and FAQ entry including all known limitations esp multitags in 1 txn.
- tests, tests, and more tests

christopherlam:
candidate [trep-chris.txt](https://github.com/Gnucash/gnucash/files/11398637/trep-chris.txt)

gnucash/report/trep-engine.scm
dawansv: I didn't realize anything survives pass a report run. I was (wrongly) under the impression everything gets cleaned up. So yes that won't work. I assumed (wrongly) your reference to `gnc:make-split->owner` was for the hashtable part, not the other stuff. In `gnc:make-split->owner` I do not understand how the hashtable created within `gnc:make-split->owner` remains initialized through what I assume must be repeated calls to `gnc:make-split->owner`. In fact in a first version I put mine in there as well and of course I even got worse performance. So I am clearly missing the part about how this "generator" is initialized somehow. Possibly the answer is in your candidate below, so if it is let me look in there.

gnucash/report/trep-engine.scm
dawansv: I am wondering if string-match would be faster. It's more limited because doesn't allow for case insensitive match but I am not using this any way. But possibly string-match is just a wrapper around make-regexp.
https://www.gnu.org/software/guile/manual/html_node/Regexp-Functions.html
As for generating make-regexp once we are back to my confusion as to where to do that if not in a global variable or passing it it all the way as an argument. This part of the overall structure of either scheme or the report engine is this eluding me.

dawansv:
> IMV "hashtag" is not a good term to use.

Yes not in love with it either. We also discussed just tag. If we go full analytical terminlogy they would be codes. The prefix would be their dimension. The Hash or whatever would be to differentiate them from anyother string. Hence the format #DimensionCode or @DimensionCode or whaterver the use wants to use as firt differentiating character. But # has become sort of standard in the free tagging world, hence hashtags.

> May I suggest "Description/Notes/Memo substring" which will unlock a substring string option. 

I hear what you are saying. In fact in my performance test I was just using the letter E to simulare lot's of tags. So it could be used on any substring. I am not sure what the use case is however, although that might be left to the enduser. What we are losing however is the introduction of a more specific terminlogy that could lead us to more support for a full implementation of analytical dimensions and codes.

> Moreover I still think 4 more options is too much. 

I don't disagree with the feeling in the context of the general transaction report whose option list is getting too crowded. Hence why I am more and more leaning towards a separate report.

>Ideally only 1 additional string option (enabled only if substring search is selected) requesting the format, but I don't know how to design it without requiring the user input the regex string.

I could see it work for just the substring part, but not the other options all passed in one string. You would essentially need to pass the options as some form of delimiter separated list. That would be really subpar user interface. So either we keep it as a simplified feature in the generic transaction report with only the prefix part (not the other options) and we do another specific analytical report (either custom or experimental report) with full options, or we just do the latter. I am not sure what the use-case is for this feature outside of the analitical tags. 

> * substring search in both primary and secondary sort key. what happens? what should happen?

Right now the same thing that happens as with the other selections (Account Name, etc.). The primary and secondary groups are simply the same nested groups, so it doesn't server any purpose. In theory we could offer a separate prefix for both, although the second prefix would need to be a sub of the first for it to make sense. That's another possibility in analytical accounting, you can have hierarchical dimensions if you provide that option. As is #DimensionSubdimensionCode; so primary can be on #Dimension and Secondary on #DimensionSubdimension. But that creates even more options; however it could be done and wouldn't be an issue if we are doing it in a separate custom or experimental report so as not to crowd the main transaction report.

> * documentation and FAQ entry including all known limitations esp multitags in 1 txn.
> * tests, tests, and more tests

Always a good reminder

dawansv:
> candidate [trep-chris.txt](https://github.com/Gnucash/gnucash/files/11398637/trep-chris.txt)

Thanks. Will have a look later today and come back...

gnucash/report/trep-engine.scm
christopherlam: > I didn't realize anything survives pass a report run. I was (wrongly) under the impression everything gets cleaned up. So yes that won't work. I assumed (wrongly) your reference to `gnc:make-split->owner` was for the hashtable part, not the other stuff. In `gnc:make-split->owner` I do not understand how the hashtable created within `gnc:make-split->owner` remains initialized through what I assume must be repeated calls to `gnc:make-split->owner`. In fact in a first version I put mine in there as well and of course I even got worse performance. So I am clearly missing the part about how this "generator" is initialized somehow. Possibly the answer is in your candidate below, so if it is let me look in there.

`gnc:make-split->owner` does not return the split's owner. It _creates_ a function (which is aptly named split->owner) which takes a split and returns its owner, and also maintains its own hash table. When the split->owner is out of scope, it is garbage collected as well as the hash table.

dawansv:
> candidate [trep-chris.txt](https://github.com/Gnucash/gnucash/files/11398637/trep-chris.txt)

Ok so lots of usefull things here that I have incorporated with some caveats

1. Code cleanup with useful tips such as using cond insteaf of if etc and reducing the use of some variables by nesting functions  (although at times I feel we lose in code readability and maintability, but that's not a hill I am willing to die for).
2. Good catch on the grid->htmltable function also using the string comparator with `gnc:string-locale<?` , although you forgot to changed that part to the new fancy `stringplus<`, so I did... Possibly a test to see if I am paying attention ;-) Note that I had to move the definitions for `stringplus<` and `stringplus>` up to the main level so they are in scope there as well.
3. Clearing the hashtag-htable at the end without which indeed you can only set the options once. That was not good!
4. The precalculation of the hash info in a record to improve speed, mostly as it relates to not rerunning make-regexp

The only caveat is on that last point. I have kept it in a separate commit because I am not certain it is worth it. My test (same as before) shows without it I am at 20 seconds (the 16 I had before was fake because I was reusing the hashtable), and with the hash-info I am at 19 seconds. I ran both test several times to be sure. Hence the impact of rerunning make-regexp seems low. I am not sure the difference justifies introducing a new specific hash-info argument that has to be passed along to all of the calls instead of the more generic options argument that could actually be used by other sortkeys as well at some other point. This especially since the trep-engine is a generic engine. So I think it is worth thinking whether that change, although technically superior, is worth it in terms of keeping the code more generic.

gnucash/report/trep-engine.scm
dawansv: OK I am going to spend more time looking at that `gnc:make-split->owner` if only to learn how it's doing this.

dawansv:
Cleanup up commits. Some recent changes that ended up in the third commit earlier belong to the first two.

christopherlam:
>     2. Good catch on the grid->htmltable function also using the string comparator with `gnc:string-locale<?` , although you forgot to changed that part to the new fancy `stringplus<`, so I did... Possibly a test to see if I am paying attention ;-) Note that I had to move the definitions for `stringplus<` and `stringplus>` up to the main level so they are in scope there as well.

It may be easier to augment `gnc:string-locale<?` itself, which will sort `'first-string` and `'last-string` preferentially to the strings themselves.

>     3. Clearing the hashtag-htable at the end without which indeed you can only set the options once. That was not good!

Ideally you won't use a global hash table.

> The only caveat is on that last point. I have kept it in a separate commit because I am not certain it is worth it. My test (same as before) shows without it I am at 20 seconds (the 16 I had before was fake because I was reusing the hashtable), and with the hash-info I am at 19 seconds. I ran both test several times to be sure. Hence the impact of rerunning make-regexp seems low. I am not sure the difference justifies introducing a new specific hash-info argument that has to be passed along to all of the calls instead of the more generic options argument that could actually be used by other sortkeys as well at some other point. This especially since the trep-engine is a generic engine. So I think it is worth thinking whether that change, although technically superior, is worth it in terms of keeping the code more generic.

Ok fair. Instead of passing `options` all over the place, it may be better to add a new entry in `used-columns`, which contains relevant options and the compiled regex too. The `used-columns` could also be renamed, and would carry the above hashtable.

dawansv:
> It may be easier to augment `gnc:string-locale<?` itself

Yes that sounds like a better plan. Will look into that.

> Ideally you won't use a global hash table.

Yes still a goal

> it may be better to add a new entry in used-columns

Sounds promissing, will have a look. I had noticed the other day how used-columns already has things that are not columns like 'subtotals-only so it has already oulived its original moniker, hence it would benefit from a cleanup regardless of our current needs.

dawansv:
Moved hashtable into hashtag-info so no more global variable https://github.com/Gnucash/gnucash/pull/1623/commits/9b6e403ad0f624b180758f244f1137338dfdf5c2
Also replaced hashtag-info with #f as parameter to sortkey-info when not needed.

Still wanting to combine hashtag-info into used-columns. used-columns (probably renamed to report-options or processed-options or such) will need to be defined in trep-renderer instead of make-split-table.  This might make sense anyway since trep-renderer is already loading a whole bunch of options, some of which make-split-table is also loading, so it migth make sense to combine them into one big process_report_options procedure. I'll give that a try on based on the current stable version first to see what it looks like.

dawansv:
@christopherlam see #1629 for refactor of used-columns into a more generic parameters alist that would allow us to carry the various parameters currently in hashtag-info

dawansv:
Here is the latest version refactored with the generic parameters alist (presented in PR #1629 )

I have also renamed hashtags to tags and moved the options back to the bottom of the sorting page

Edit: I have updated the screenshots in my main comment.
~~Edit2: Ooops, missed one more of these sortkey-list parameters argument. Will fix later tomorrow.~~ Done.

benbucksch:
@dawansv I am not part of the project and definitely not a reviewer, but I've looked at the code, and the latest version looks really clean to me. Even if I don't know Scheme, I can still read roughly what you change and why. The changes look reasonable and straight-forward to me, given the task here. Very clean code. Well done. Even though my opinion isn't worth anything, I still think this patch is worth the praise.

dawansv:
@benbucksch @flywire I have created a custom report version of this feature so that possibly more users could start using this and provide feedback. Before I advertise this on the enhancement report, if you have a minute, I would love if you could try installing it as per my instructions here https://github.com/dawansv/gnucash-transaction-tags . Thanks.

I will add a disclaimer on that same page to make sure that users understand that as a custom report there is no guarantee right now that this will be released as an official feature or even provided in the long term. But for now it's relatively easy for me to keep what I am doing here and the custom report version in sync, so we might as well do it. More feedback will make it a stronger feature in the long term.

flywire:
@dawansv I've been hacking the report straight into the Windows distribution and it has been working fine although I haven't tested the latest mods.

Ultimately, it will need documentation in the Guide and that really means extending the existing sample data with a credible scenario. I've looked through the Guide with that in mind but in my view, it has a few structural issues and the start is daunting for a new non-accounting user. The concept of a transaction isn't clear enough which flows through to no real explanation of description and note, and the comment pretty well applies to the memo too.

These comments sit behind my recent [post](https://lists.gnucash.org/pipermail/gnucash-user/2023-May/106929.html) and [PR](https://github.com/Gnucash/gnucash-docs/pull/319). In my view, this could be fixed without any changes to debits/credits by explaining the concept of a transaction before accounts at the start of the chapter. 

> Accounting is a record of transactions that record how money
moves from one place to another. Understandably, accounting calls those
places accounts and arranges them in standard types. Everyone is familiar
with withdrawing cash from a savings account. Those are two accounts,
savings, and cash. You could spend that cash on groceries or clothing, two
more accounts.

Again, the transaction needs more in the UI section, including identifying description and note, rather than the current emphasis on splits. I think it is time to bring @fellen and @gjanssens into the conversation.

dawansv:
> Again, the transaction needs more in the UI section, including identifying description and note, rather than the current emphasis on splits.

@flywire From the get go I need to say that because I have formal accounting training I have never really looked at the concept guide much. But I can agree that the last thing we need is people getting even more confused and using tags instead of general ledger accounts. So from the get go there might need to be a disclaimer about that, and about the fact that until a user has a good understanding of transactions, splits and account, probably using tags isn't a good idea. 

And yes the concept of transactions should probably be further up because they are a great way to introduce double entry since their main caracteristic is that there need to balance to zero.

The credit/debit inversion in every day language is unfortunate and we can thank the banking world for it. But now we are stuck with it. The idea of using informal labels as default is a worthwhile idea, considering the feature is already implemented. Maybe there should be PR about that as well so discussions stay in one place. 

As for the "double line" language and how it relates (or if fact doesn't relate) to splits is a good point as well. I will leave some comments in your PR about that because even the transaction report has a very odd way of dealing with the memo vs notes issues by inexplicably making it an either/or thing, so I'll leave something about that as well.

dawansv:
Rebased with current version of trep-engine. 
For now this lives as a custom report at https://github.com/dawansv/gnucash-transaction-tags
Putting this back into draft until I can figure out how and if we refactor trep-engine to be more flexible with additional parameters as in PR #1629

dawansv:
@christopherlam OK so with the parameters alist refactor this code change is now much shorter. 

Yet we sill have the infinity-string part https://github.com/Gnucash/gnucash/pull/1623/commits/3caddf841d24e7915fe248d8d8b6ad8ca1214f5e that can be implemented in a more generic way in core-utils.scm, as you said in an earlier comment:
>It may be easier to augment gnc:string-locale<? itself, which will sort 'first-string and 'last-string preferentially to the strings themselves.

Regardless of what we decide to do with this feature (keep as custom report or changing it to a generic "sort by substring" feature) I'd like to get that out of the way first, since I think it is a worthwhile feature in itself. So I'll submit a separate PR to augment gnc:string-locale<? in core-utils.scm if you are still on board with it.

christopherlam:
This branch is ok, however I still feel that it is not the greatest idea to use a transaction textual fields to denote individual tags. IMV Description/notes/memos should be static.

As a preliminary step, it would not be too difficult to reuse the doclink mechanism and UI to store/retrieve a string in a transaction KVP tree, labelled "transaction-tags". This would be a space or comma delimited string containing the desired metadata, freely modifiable via dialog (see dialog-doclink) and will be read and interpreted via a xaccTransGetTags which returns either a string or a list of strings. Later on, it will be less difficult to upgrade this to a full tagging system. If you'd use part of a description instead, then modifications of tags require changing the trans->description individually.

jralls:
> however I still feel that it is not the greatest idea to use a transaction textual fields to denote individual tags. IMV Description/notes/memos should be static.

Why do you think that adding tags to text fields would make them any less static? Nobody is going to edit a register-full of transactions to adjust report settings. The same would be true for a separate categories field, no matter how implemented.

The more pertinent question is why is this better than simply creating a deeper account tree? E.g. from the first example Expenses:Anna:Clothes, Expenses:Anna:Dining, Expenses:Nobody:Clothes, and Expenses:Nobody:Dining.

@flywire's matrix report is interesting, but the example suggests that it's out of scope: A construction company large enough to need to have separate columns for New Construction, Remodels, Overhead, and Unclassified is probably too big to be in GnuCash's target audience. If instead one uses @dawansv's categories so the table is
```
**Expense Type**              **Anna**         **Nobody**     **Total**
Clothes
Dinner
Total

One could implement that as easily with the four subaccounts as with category tags by providing a Display option something like "Turn like-named leaf accounts into columns".

dawansv:
@christopherlam 
> transaction KVP tree

Interesting. Didn't even know about this key-value pair system. Good thing it's documented here : https://github.com/Gnucash/gnucash/blob/stable/libgnucash/engine/kvp_doc.txt

And it looks like it's also available at the split level which is good since the main tagging level is actually the split (hence why memo is considered first). Tagging at the transaction level is more of a shortcut that says that the transaction tag is used by default for all splits that don't have a tag in that particular dimension. Hence ideally, we need a tagging mechanism at both levels, although we can start with the split level first since that's the main one. Anyway I'll look into it.

One parallel implementation that could happen faster would be to release this as a more generic "sort by substring" as you had mentioned earlier, with the ability to pass a regex that would give a prefix and main variable part of the string. The regex would actually allow you to put another end point instead of a space. Then it's not specifically for tagging but could be used for that just as any other substrings following a pattern. It would actually be a companion to the transaction filter since it can work in tandem as I show in my examples above (the filter can remove or only select some of the sorted tags/substrings using a similar regex). This doesn't need to conflict or replace this other idea of a more specific tagging feature using the KVP tree. I'll give that a try to time permitting. For now, this is also available as a custom report, so I don't think there is a need to rush to a decision anyway.

christopherlam:
> The more pertinent question is why is this better than simply creating a deeper account tree? E.g. from the first example Expenses:Anna:Clothes, Expenses:Anna:Dining, Expenses:Nobody:Clothes, and Expenses:Nobody:Dining.

I'll have an answer for this part only: when setting up a deeper account tree, if I'm a landlord with multiple properties, I'll have a dilemma whether to go for Expenses:House1:Utilities or Expenses:Utilities:House1. This is only 1 example which can be aided by a multidimensional tagging system. I am not sure I have answers otherwise.

christopherlam:
> And it looks like it's also available at the split level which is good since the main tagging level is actually the split (hence why memo is considered first). Tagging at the transaction level is more of a shortcut that says that the transaction tag is used by default for all splits that don't have a tag in that particular dimension. Hence ideally, we need a tagging mechanism at both levels, although we can start with the split level first since that's the main one. Anyway I'll look into it.

KVP is available at split, transaction, account book levels (and possibly more eg business objects). Hence, all sorts of rubbish and gems have been stored there. Budgets data, online_id for ofx, bayesian data, and etc all are stored in KVP. And it's not exactly desirable to add more.

flywire:
> why is this better than simply creating a deeper account tree?

Following on from @christopherlam example. It's all about independent dimensions. Extending the example, the same expense can be Utilities, group anywhere along Property:House:House1, Tenant, City, Project, Manager. ~Any~ All of these from a single transaction.

For example, how does water use compare between: property types (houses to units), tenants, cities, projects, or property managers? What about maintenance costs?

jralls:
> I'll have an answer for this part only: when setting up a deeper account tree, if I'm a landlord with multiple properties, I'll have a dilemma whether to go for Expenses:House1:Utilities or Expenses:Utilities:House1. This is only 1 example which can be aided by a multidimensional tagging system. I am not sure I have answers otherwise.

That question isn't answered with tags. You have either Expenses:Utilities and subdivide it by house or the other way around. Whether to use tags or subaccounts is orthogonal.

dawansv:
@jralls 
> The more pertinent question is why is this better than simply creating a deeper account tree?

This is going back to the root of the question for sure. There is actually a lot of information about this on the original bug report but it's lost in the sea of comments.

The very short version is that analytical accounting exists both in terms of accounting practice but also across many accounting packages and therefore (trust me) there must be a reason for it. I am saying this in jest a bit, but the point is that I am not inventing any of this, it is standard accounting practice in many places. That said I found this which I think explains why we are going in circle on this when talking to people outside of Europe.

> Some European countries use two types of accounting systems. (The United States and most other countries use only one.)
> 
> The general accounting system is a legal obligation. It must conform to certain accounting principles and must represent a fair picture of the financial situation of the company by producing a balance sheet and a profit and loss statement. Its foundation is the Chart of Accounts, made up of eight classes. Classes 1-5 are balance sheet accounts and classes 6 and 7 (expenses and revenues) are used for the profit and loss statement. All journals and accounts post to the General Ledger.
> 
> The second accounting system used in some countries is called Analytical Accounting. Its main purpose is to track expense and revenue accounts by categories in order to derive profit and loss by activity. Its foundation is a separate Chart of Accounts made up of a single class (class 9). Its journals and accounts post to a separate ledger (the Analytical Ledger.)

Source: https://docs.infor.com/ism/5.x/en-us/ism_onlinehelp/cei1464154169202.html

The slightly longer version is that analytical accounting provides the ability for many orthogonal classifications. To be honest it is used in the US as well, but maybe restricted to the nonprofit world. Even if you CAN address ONE extra classification via the chart of account, you will be stuck when a second type of classification is needed. For instance a non profit needs to classify all expenses:
1) in nature (electricity, wages, supplies) --> no problem, that's the standard chart of account
2) by project or program. OK fine let's split the chart of account as you show (Anna is a program, bob is a program, etc). OK but ugly because programs are time limited, so my chart of accounts keeps having more and more in them even though they are no longer active. Still OK it works.
3) by source of funds/grants. No you can't create grants under projects or projects under grant because some project are funder by several grants while some grants are funding several projects. Even if you could your chart of account is now a complete jumble as years go by and new projects and grants are started and old ones no longer active.

The non-profit example is just one which possibly is more common in the US. In other countries this sort of thing is very common in many other applications.

GnuCash does have a very narrow implementation of analytical accounting via the job feature. It's incomplete and specific while analytical accounting is generic, it can be used with any orthogonal categorization. In fact I would think it would appeal GnuCash developpers since it follows generic principles similar to how double accounting follows generic principles to set up any chart of accounts.

dawansv:
And here is somebody trying to explain the concept to American users: https://www.ktlsolutions.com/analytical-accounting-what-is-it-how-can-it-help-my-business/
It's unclear to me if they succeed.

jralls:
@dawansv 
> The slightly longer version is that analytical accounting provides the ability for many orthogonal classifications.

The simple response is: Of course. For large enterprises analytical accounting is enormously important. But GnuCash isn't suitable for large&mdash;or even small;mdash;enterprises and never will be. GnuCash is for tiny and simple enterprises; we often say individuals and sole proprietorships. Even having employees makes using GnuCash a dubious proposition (no payroll module), as does carrying more than a very few line items of inventory (no inventory module)  or performing any sort of manufacturing (no bill of materials or cost accounting modules). 

Adding those modules is a non-starter: Consider that it's been 10 years since we began working on separating libgnucash from the application, converting from GObject to C++, and changing from a program that can load its data from a SQL database and put it back to one that is built around SQL queries. We haven't gotten very far. 

Those seeking a F/LOSS accounting solution for larger enterprises should look to [Odoo](https://www.odoo.com).

dawansv:
@jralls I don't think anybody is asking for a full scale analytical accounting system. I was just trying to answer your question as to why creating a deeper account tree doesn't always work. I mean the tagging feature is just a low level basic implementation. I am keeping the books for a small nonprofit with a $20,000 budget and gnucash is fine for that. I have even used the workaround I explain in the bug report (using the include/exclude account feature of the transaction report) when I needed some grant accounting for another business. I don't personally have a problem using free tags for my use, and it's fine if this stays at the level of custom reports. 

I would be more interested to get feedback on the idea of simply releasing this as a more generic substring sort in the transaction report. Considering all the other filtering available in that report I would think it would be a useful addition. I assume the existing filtering feature was in part related to free tagging.  We are certainly not forcing anybody to use that feature nor a new sort option. Otherwise this will just live as a custom report.

jralls:
@dawansv, ISTM @christopherlam is arguing for turning this into a full feature using the already heavily-abused KVP misfeature, and @flywire persists in offering examples that are clearly applicable only to big businesses. ISTM even your funding/project matrix requirement is an indication that your non-profit, small as it's budget may be, has outgrown GnuCash. That's not just because it requires a CoA that's more complex than GnuCash can handle: It also implies that one or more of your grantors or funders will want to audit your accounting system, and GnuCash's complete lack of controls and traceability will ensure that the audit lasts only a few minutes and that you'll get no more funding from that source.

flywire:
> @flywire persists in offering examples that are clearly applicable only to big businesses.

This comment reflects a lack of understanding of how software is used.

I've offered three examples. Let's assume it doesn't relate to the [Guide example](https://github.com/Gnucash/gnucash/pull/1623#issuecomment-1530788858). The 2 dimensional [Quickbooks report](https://github.com/Gnucash/gnucash/pull/1623#issuecomment-1530802670) is applicable to any sized business and I can assure you the extension of the [rental water use example](https://github.com/Gnucash/gnucash/pull/1623#issuecomment-1583936430) is typical of the sort of comparisons I've done on rentals, hardly _big business_.

Still, it's trivial compared to the analysis of a zero-employee farm with cash accounting.

benbucksch:
@jralls :
> it requires a CoA that's more complex than GnuCash can handle: ... GnuCash's complete lack of controls and traceability will ensure that the audit lasts only a few minutes and that you'll get no more funding from that source.

Since over 20 years, I use GnuCash for all my personal finances and my business, and I need these multi-dimension classifications for both. I have survived audits. I can be analytical in my finances without having a legal requirement to do so, but simply for myself, because I want to analyse my finances, in a way that makes sense for me personally, to keep them better under control. For that, multiple dimensions as outlined in https://github.com/Gnucash/gnucash/pull/1623#issuecomment-1533253932 is a necessity and has been sorely lacking from GnuCash. I tried to use deep accounts trees, and they don't cut it, as explained in my comment.

Can we please all get along, and accept that for some of us, this is super helpful and in fact necessary, while not helpful for other users? And that this is fine, and that this is nonetheless a good and useful feature to add?

Open-Source is all about fixing an itch. That's how it matures and becomes more useful for more people.

jralls:
> Since over 20 years, I use GnuCash for all my personal finances and my business, and I need these multi-dimension classifications for both.

That implies that you've already got a way to do the multi-dimension analysis. Otherwise you would already have switched to a program that supports your more complex requirements.

> Accept…that this is fine, and that this is nonetheless a good and useful feature to add?

That depends on how it's framed as a feature. If we announce that GnuCash supports multidimensional analysis based on this hack (sorry @dawansv) we'll get no end of bug reports wanting extensions and complaining that it doesn't work exactly right. Worse, we'll be stuck with it forever because it exposes way too much implementation detail to the user.

On the other hand if we take out the language about multidimensional analysis and the special character introducing tags and do it as a report option to filter entries based on a string match on an arbitrary string or even a regular expression in a selected text field&mdash;any of description, notes, memo, num, or action&mdash;then users like you and @dawansv can write custom reports around it to support your analysis, other users can find other uses, and it doesn't constrain future development.

> Open-Source is all about fixing an itch. That's how it matures and becomes more useful for more people.

Open source enables you to modify the program to scratch your itch, and you can offer your mods back to the originating project. The project is in no way obligated to accept every offered change.

dawansv:
Thank you for all the useful feedback. I am closing this PR and moving the conversation to PR #1673 where I present a similar but generic feature implementing sort and subtotal by substring.

benbucksch:
@jralls I see that you have been one of the most active contributors to GnuCash in the last years. Thank you for your work on GnuCash. I deeply appreciate GnuCash. I like GnuCash very very much, and tell my friends about it. It has been very helpful in keeping my finances and spending rational and within the set limits.

There are some analysis that are not possible right now, and @dawansv 's work would make that possible, that's why I'm very happy about it.

> That implies that you've already got a way to do the multi-dimension analysis. Otherwise you would already have switched

I don't know how you arrive at that conclusion. I explicitly stated the opposite. I wrote:

"Since over 20 years, I use GnuCash for all my personal finances and my business"

That means
a) I currently use GnuCash ("since" means ongoing, "use" is in present tense)
b) I am probably one of your most loyal and happy users
Even though GnuCash meets most of my needs, it doesn't meet this one need, and it's an important need. The lack causes me problems.

I cannot run certain analysis on certain factors, which means I am lacking specific information that I need to make better informed financial decisions. I even gave specific examples in [my earlier comment](https://github.com/Gnucash/gnucash/pull/1623#issuecomment-1533253932), e.g. I book how much gas I spent with a particular car, and I book how much I spend for airplane and hotel, but I cannot find out how much a specific travel event (by car or flight) has cost me. I could theoretically book things on the purpose (e.g. the travel), but then I cannot know anymore how much I spent for hotels in total.

The words "sorely lacking" - meaning that it affects me strongly - in fact show that there is a real problem here, which is not addressed and where I did not find a solution yet. I don't know how you get in your logic from me stating 'there is an unfulfilled need' and "sorely lacking" to you stating "you've already got a way". Your response (the part which I quoted above) seems to me like a deliberate mis-representation of what I said. I have to say that I'm rather irritated by that.

> The project is in no way obligated to accept every offered change.

Of course, there is no legal obligation for you or any other GnuCash project member to do anything at all, given that you're not paid for it. I've been an Open Source dev since almost 25 years, and maintain software myself, so I do understand.
However, an Open Source project has a moral obligation to serve its end users, and has a self-interest in keeping developers/contributors happy.

This patch serves a real end-user need, and @dawansv took the time to implement it. You think the implementation is not good. OK. In such a case, it would be helpful to give _constructive and specific_ feedback how the goal could be accomplished in a way that you deem good.

You mentioned a "report option to filter entries based on a string match on an arbitrary string or even a regular expression in a selected text field—any of description, notes, memo, num, or action", but that's a little vague. Also, as a user, I wouldn't know how to use it.

@dawansv has done like 10 to 20 revisions already, including several major rewrites, based on previous feedback. He's met any reasonable request so far.

My wish is only that everybody could work together.

PR #1624, opened 2023-05-03
Use trep-engine lowest shown grouping for detail

flywire:
If the Transaction report option `Show subtotals only (hide transactional data)` is selected then the `Total For` prefix should not be shown.

Would be better to also use normal row style, see https://lists.gnucash.org/pipermail/gnucash-user/2023-January/105139.html

### DIFF
--- a/gnucash/report/trep-engine.scm
+++ b/gnucash/report/trep-engine.scm
@@ -138,6 +138,10 @@
 (define def:primary-subtotal-style "primary-subheading")
 (define def:secondary-subtotal-style "secondary-subheading")
 
+(define (option-ref opts section name)
+  (let ((opt (gnc:lookup-option opts section name)))
+    (and opt (gnc:option-value opt))))
+
 (define NO-MATCHING-TRANS-HEADER (G_ "No matching transactions found"))
 (define NO-MATCHING-TRANS-TEXT (G_ "No transactions were found that \
 match the time interval and account selection specified \
@@ -1062,6 +1066,17 @@ be excluded from periodic reporting.")
 
   (let* ((work-to-do (length splits))
          (table (gnc:make-html-table))
+         (subtotals-only? (option-ref options pagename-sorting optname-show-subtotals-only))
+         (sec-subtotal?   (option-ref options pagename-sorting optname-sec-subtotal))
+         (subtotal-label (lambda (str level)
+                           (let ((show-prefix? (not (and subtotals-only?
+                                                         (case level
+                                                           ((secondary) sec-subtotal?)
+                                                           ((primary) (not sec-subtotal?))
+                                                           (else #f))))))
+                             (if show-prefix? 
+                                 (string-append (G_ "Total For ") str) 
+                                 str))))
          (account-types-to-reverse
           (keylist-get-info sign-reverse-list
                             (report-uses? 'reversed-signs)
@@ -1613,7 +1628,9 @@ be excluded from periodic reporting.")
             (gnc:html-make-empty-cells left-indent)
             (if (report-uses? 'export-table)
                 (cons
-                 (gnc:make-html-table-cell/markup "total-label-cell" data)
+                 (gnc:make-html-table-cell/markup 
+                  (if (summary-style? level) "total-label-cell" "text-cell") 
+                  data)
                  (gnc:html-make-empty-cells
                   (+ right-indent width-left-columns -1)))
                 (list
@@ -1660,6 +1677,14 @@ be excluded from periodic reporting.")
       (and (pair? calculated-cells)
            (assq-ref (car calculated-cells) 'merge-dual-column?)))
 
+    (define (summary-style? level)
+      (let ((detail-level
+             (cond ((not (report-uses? 'subtotals-only)) 'transactions)
+                   ((report-uses? 'secondary-key/renderer-fn) 'secondary)
+                   ((report-uses? 'primary-key/renderer-fn) 'primary)
+                   (else 'grand))))
+        (not (eq? level detail-level))))
+
     (define (add-subtotal-row subtotal-string subtotal-collectors
                               subtotal-style level row col)
       (let* ((left-indent (case level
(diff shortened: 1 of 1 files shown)


dawansv:
Yes I have noticed this before and it would be nice to improve it. That said I think it might need a bit more than this. Ideally we should keep the "Total for" on the primary subtotal when both primary and secondary subtotals are present. In that case we should also bring back the header of the primary section which is currently not displayed when "Show subtotals only" is selected.

So for instance if you have primary on account (e.g cash and bank) and secondary on month, then you would have

```
Cash
  January 2023                $100
  February 2023               $200
  march 2023                  $100
Total for Cash                $400
Bank
  January 2023                $300
  February 2023               $400
  march 2023                  $200
Total for Bank                $900
----------------------------------
Grant Total                  $1300
```

flywire:
Good point which shouldn't be too hard to resolve.

Any thoughts on how to supress the highlighting? The code in the last line of linked post seemed obvious but didn't work:

```
    (define (total-string str)
      (if (opt-val pagename-sorting optname-show-subtotals-only)
        str
        (string-append (G_ "Total For ") str)))

    ; (set! def:primary-subtotal-style def:normal-row-style)
```

dawansv:
> Any thoughts on how to supress the highlighting? The code in the last line of linked post seemed obvious but didn't work

That's because the style is already passed earlier when calling add-subtotal-row, the function that builds the subtotal row (starts on line 1572).

If you do a search on add-subtotal-row you'll see where it's being called for the secondary (line 1874) and the primary (line 1887). These 2 are only called when the primary is active. This is where you have to add some logic to change `def:primary-subtotal-style` to `def:normal-row-style` when subtotals-only is selected. Note however that for the primary we should still keep `def:primary-subtotal-style` when the secondary is on (to achieve the example I gave earlier)

Further down at 1912 is also a call to add-subtotal-row when only the secondary is selected. There too you have to add some logic to change `def:primary-subtotal-style` to `def:normal-row-style` when subtotals-only is selected.

Now keep in mind all of this code is still in the context of the big function that build the whole table, make-split-table (line 1022 to 1945). In that context you can simply use `(column-uses? 'subtotals-only)` to check is subtotals-only are enabled.

Now the style above is only for the row style. For font you have to change the `"total-label-cell"` cell markup in add-subtotal-row. 

```
        (define (first-column string)
          (if export?
              (cons
               (gnc:make-html-table-cell/markup "total-label-cell" string)
               (gnc:html-make-empty-cells (+ right-indent width-left-columns -1)))
              (list
               (gnc:make-html-table-cell/size/markup
                1 (+ right-indent width-left-columns) "total-label-cell" string))))
```

I am thinking instead of redoing the full check here, it might be clean enough to just check subheading-style, the row style passed to the function. If `subheading-style` is equal to `def:normal-row-style` then we should create the cell without markup. That means `(gnc:make-html-table-cell string)` instead of `(gnc:make-html-table-cell/markup "total-label-cell" string)`
and `(gnc:make-html-table-cell/size 1 (+ right-indent width-left-columns) string)`  instead of `(gnc:make-html-table-cell/size/markup 1 (+ right-indent width-left-columns) "total-label-cell" string)`

Finally there is the whole piece about bringing back the primary header when a secondary subtotal is shown. That would need to be done in add-subheading function (line 1497). That part where it says `(unless (column-uses? 'subtotals-only)` needs to be change so that the code still runs on primary when both primary and secondary subtotals are on, even when 'subtotals-only are on.

If none of this makes sense, don't worry about it. I can work on this later on as well.

flywire:
> If none of this makes sense

This is terrific guidance. Let me start working through it, it's much easier for me to learn by working toward an objective.

dawansv:
> it's much easier for me to learn by working toward an objective.

Yes I am still new to scheme (and the gnucash report engine mechanism) and learning by doing is also what I like to do. I still have a lot to learn. Christopher Lam has been very patient with providing feedback as well (at time probably spending more time on that than doing the changes himself). Scheme is a bit esoteric, although I think a lot of fun. I also think there is room for developpers with various skill levels to pitch in, although they have to be willing to be patient and learn because being the core developper than enforce quality control is important but also a lot of work. On the other hand, without a more active community with at least some knowledge of the code base, it's hard to transition when people eventually have to move on...

jralls:
> [Merge branch 'Gnucash:stable' into Subtotal](https://github.com/Gnucash/gnucash/pull/1624/commits/35513b417f07843f85ca9542c37f4a1f5f393993)

Please don't do that, it makes loops in git history that screws up bisecting. *Rebase* the feature branch on the current target branch instead.

christopherlam:
I think this is a nice refinements to the subtotals... however it would be nice to be extremely explicit of the rules whether to insert "Total For ". What about the scenario if the `primary subtotal` is `none` and `secondary subtotal` is `date` or `account`?

flywire:
> What about the scenario if the primary subtotal is none and secondary subtotal is date or account?

Text seems fine in my test. Try it yourself.

BTW, concern re changing style too late:

>That's because the style is already passed earlier when calling add-subtotal-row, the function that builds the subtotal row

flywire:
Existing report:

image

image

Show subtotals only (hide transactions)

---

New report functionality with lowest shown grouping for detail:

image

flywire:
Consider this for `release 5.14`.

jralls:
> Consider this for release 5.14.

@christopherlam up to you.

christopherlam:
A PR needs illustration; IIUC this changes the trep subtotals from:
image
to:
image

IMHO this is a visual change that styles subtotals into normal text, which is slightly jarring because normal text would typically indicate transactional rather than aggregate sums. However there are other aggregate sums in normal text eg networth table, trep subtotal table. Ultimately I think it's a harmless change.

flywire:
> IMHO this is a visual change that styles subtotals into normal text

No, it actually changes the text (see [updated example](https://github.com/Gnucash/gnucash/pull/1624#issuecomment-3636584747)).

---

Here is a version selecting the non-default `Add indenting columns` with a more realistic secondary subtotal list:

image


Or without Primary subtotals:

image

christopherlam:
Ok merged in. At some point the Transaction Report can no longer be described as a transaction report ;-)

jralls:
> At some point the Transaction Report can no longer be described as a transaction report ;-)

[Not @flywire's doing](https://github.com/flywire/gnucash/commit/e6dcc0cc1eea910d1dbad29725825954bc14dd73)

PR #1626, opened 2023-05-05
WIP: Incorporate new Finance::Quote get_features() to automatically configure quote sources

vincentl:
# Background

This work in progress PR is a continuation of a discussion started on [GNG-dev](https://lists.gnucash.org/pipermail/gnucash-devel/2023-January/046534.html) in January.

This PR will:
- Use the new Finance::Quote `get_features()` to populate both the list of single quote sources and quote methods that use multiple sources
- Any security quote source discovered in a user's datafile that is not found in the `get_features()` list will be registered in the "unknown" source list in GnuCash
- Establish data structures in GnuCash to hold additional information returned by `get_features()` such are required API keys for certain sources

A future PR will implement GUI features that expose the additional information to users and enable setting of fields such as API keys.

# What is Implemented

- Add command line option to finance-quote-wrapper to call FQ::get_features() and return result on STDOUT as json
- C++ code that parses the features json blob from finance-quote-wrapper into map of map of list of strings

# Next Steps

- Implement `get` functions to retrieve single source, multi-source, and currency methods from the parsed json to populate menu lists in the GUI
- Add additional documentation to clarify changes

### DIFF
--- a/libgnucash/app-utils/gnc-quotes.cpp
+++ b/libgnucash/app-utils/gnc-quotes.cpp
@@ -232,11 +232,10 @@ m_version{}, m_currency_sources{}, m_single_sources{}, m_multiple_sources{}, m_a
             // info["quote_methods"]    = { "name" : ["module_name", ...] }
             // info["quote_modules"]    = { "name" : ["feature_name", ...] }
             //
-            // A quote_method is the name for one more quote sources, such as NYSE
-            // A quote_module is a single source with option "features"
+            // A quote_method is the name for one or more quote sources, such as nasdaq (multiple) or alphavantage (single)
+            // A quote_module is backs one or more methods and may have required features.
             // A "feature_name" is something like "API_KEY", a need value for the source
 
-            PWARN("TAG: %s", info["currency_modules"]["AlphaVantage"][0].c_str());
             m_version = std::move(version);
 
             for (auto& it : info["currency_modules"]) {
@@ -244,14 +243,14 @@ m_version{}, m_currency_sources{}, m_single_sources{}, m_multiple_sources{}, m_a
             }
             std::sort (m_currency_sources.begin(), m_currency_sources.end());
             
-            for (auto& it : info["quote_modules"]) {
-                m_single_sources.push_back(it.first);
+            for (auto& it : info["quote_methods"]) {
+                // Sort methods into single vs multiple sources
+                if (info["quote_methods"][it.first].size() == 1)
+                    m_single_sources.push_back(it.first);
+                else
+                    m_multiple_sources.push_back(it.first);
             }
             std::sort (m_single_sources.begin(), m_single_sources.end());
-
-            for (auto& it : info["quote_methods"]) {
-                m_multiple_sources.push_back(it.first);
-            }
             std::sort (m_multiple_sources.begin(), m_multiple_sources.end());
         }
         catch (std::exception &e)
@@ -555,7 +554,7 @@ GncQuotesImpl::comm_vec_to_json_string(const CommVec &comm_vec) const
                                return;
                        }
                        else
-                           comm_ns = gnc_quote_source_get_internal_name(gnc_commodity_get_quote_source(comm));
+                           comm_ns = gnc_quote_source_get_name(gnc_commodity_get_quote_source(comm));
 
                        pt.put (make_quote_path(comm_ns, comm_mnemonic), "");
                    }
--- a/libgnucash/app-utils/gnc-quotes.cpp
+++ b/libgnucash/app-utils/gnc-quotes.cpp
@@ -80,9 +80,6 @@ struct GncQuoteSourceError : public std::runtime_error
 CommVec
 gnc_quotes_get_quotable_commodities(const gnc_commodity_table * table);
 
-/**
- * GncQuotesSource provides an API for quote retrieval used by GncQuotesImpl.
- **/
 class GncQuoteSource
 {
 public:
@@ -94,11 +91,6 @@ public:
     virtual QuoteResult get_quotes(const std::string& json_str) const = 0;
 };
 
-/**
- * GncQuotesImpl defines the Quote API used by the core of GnuCash.  A quote source
- * such as Finance::Quote implements the GncQuotesSource API and this class translates
- * between that API and the public API this class defines for the rest of GnuCash.
- **/
 class GncQuotesImpl
 {
 public:
@@ -144,9 +136,6 @@ private:
     gnc_commodity *m_dflt_curr;
 };
 
-/**
- * GncFQQuotesSource implements the GncQuotesSources API for Finance::Quote. 
- **/
 class GncFQQuoteSource final : public GncQuoteSource
 {
     const bfs::path c_cmd;
--- a/libgnucash/app-utils/gnc-quotes.cpp
+++ b/libgnucash/app-utils/gnc-quotes.cpp
@@ -69,6 +69,8 @@ namespace bpt = boost::property_tree;
 namespace bio = boost::iostreams;
 
 using QuoteResult = std::tuple<int, StrVec, StrVec>;
+using ListMap = std::map<std::string, std::vector<std::string>>;
+using KeyMap = std::map<std::string, ListMap>;
 
 struct GncQuoteSourceError : public std::runtime_error
 {
@@ -151,35 +153,77 @@ GncFQQuoteSource::GncFQQuoteSource() :
 c_cmd{bp::search_path("perl")},
 m_version{}, m_sources{}, m_api_key{}
 {
-    char *bindir = gnc_path_get_bindir();
-    c_fq_wrapper = std::string(bindir) + "/finance-quote-wrapper";
-    g_free(bindir);
-    StrVec args{"-w", c_fq_wrapper, "-v"};
-    auto [rv, sources, errors] = run_cmd(args, empty_string);
-    if (rv)
     {
-        std::string err{bl::translate("Failed to initialize Finance::Quote: ")};
-        for (const auto& err_line : errors)
-            err += err_line.empty() ? "" : err_line + "\n";
-        throw(GncQuoteSourceError(err));
+        // Call Finance::Quote wrapper with info flag
+        char *bindir = gnc_path_get_bindir();
+        c_fq_wrapper = std::string(bindir) + "/finance-quote-wrapper";
+        g_free(bindir);
+        StrVec args{"-w", c_fq_wrapper, "-i"};
+        auto [rv, info, errors] = run_cmd(args, empty_string);
+
+        // Check for 0 exit code and exactly one line from wrapper
+        if (rv || info.size() != 1)
+        {
+            std::string err{bl::translate("Failed to initialize Finance::Quote: ")};
+            for (const auto& err_line : errors)
+                err += err_line.empty() ? "" : err_line + "\n";
+            throw(GncQuoteSourceError(err));
+        }
+        // errors should be empty too
+        if (!errors.empty())
+        {
+            std::string err{bl::translate("Finance::Quote check returned error ")};
+            for(const auto& err_line : errors)
+                err += err.empty() ? "" : err_line + "\n";
+            throw(GncQuoteSourceError(err));
+        }
+
+        // Parse JSON
+        try 
+        {
+            std::istringstream is(info.front());
+            bpt::ptree root;
+            bpt::read_json(is, root);
+
+            // Get version and remove it from the tree
+            auto version = root.get<std::string>("version");
+            root.erase("version");
+
+            // Remaining entries look like: key : {key : [value, ...], ...}
+            // Parse as KeyMap, which is string -> ListMap (string -> vec)
+            KeyMap info;
+            for (auto& [outer_k, outer_v]: root) {
+                auto& outer_entry = info[outer_k];
+                for (auto& [inner_k, inner_v]: outer_v) {
+                    auto& inner_entry = outer_entry[inner_k];
+                    for (auto& value : inner_v) {
+                        inner_entry.push_back(value.second.data());
+                    }
+                }
+            }
+
+            // info["currency_modules"] = { "name" : ["feature_name", ...] }
+            // info["quote_methods"]    = { "name" : ["module_name", ...] }
+            // info["quote_modules"]    = { "name" : ["feature_name", ...] }
+            //
+            // A quote_method is the name for one more quote sources, such as NYSE
+            // A quote_module is a single source with option "features"
+            // A "feature_name" is something like "API_KEY", a need value for the source
+
+            PWARN("TAG: %s", info["currency_modules"]["AlphaVantage"][0].c_str());
+            m_version = std::move(version);
+            for (auto& it : info["quote_methods"]) {
+                m_sources.push_back(it.first);
+            }
+            std::sort (m_sources.begin(), m_sources.end());
+        }
+        catch (std::exception &e)
+        {
+            std::string err{bl::translate("Finance::Quote failed in parsing quote sources ")};
+            err += e.what();
+            throw(GncQuoteException(err));
+        }
     }
-    if (!errors.empty())
-    {
-        std::string err{bl::translate("Finance::Quote check returned error ")};
-        for(const auto& err_line : errors)
-            err += err.empty() ? "" : err_line + "\n";
-        throw(GncQuoteSourceError(err));
-    }
-    auto version{sources.front()};
-    if (version.empty())
-    {
-        std::string err{bl::translate("No Finance::Quote Version")};
-        throw(GncQuoteSourceError(err));
-    }
-    m_version = std::move(version);
-    sources.erase(sources.begin());
-    m_sources = std::move(sources);
-    std::sort (m_sources.begin(), m_sources.end());
 
     auto av_key = gnc_prefs_get_string ("general.finance-quote", "alphavantage-api-key");
     if (!(av_key && *av_key))
--- a/libgnucash/app-utils/gnc-quotes.cpp
+++ b/libgnucash/app-utils/gnc-quotes.cpp
@@ -80,16 +80,25 @@ struct GncQuoteSourceError : public std::runtime_error
 CommVec
 gnc_quotes_get_quotable_commodities(const gnc_commodity_table * table);
 
+/**
+ * GncQuotesSource provides an API for quote retrieval used by GncQuotesImpl.
+ **/
 class GncQuoteSource
 {
 public:
     virtual ~GncQuoteSource() = default;
-    virtual const StrVec& get_sources() const noexcept = 0;
+    virtual const StrVec& get_currency_sources() const noexcept = 0;
+    virtual const StrVec& get_single_sources() const noexcept = 0;
+    virtual const StrVec& get_multiple_sources() const noexcept = 0;
     virtual const std::string & get_version() const noexcept = 0;
     virtual QuoteResult get_quotes(const std::string& json_str) const = 0;
 };
 
-
+/**
+ * GncQuotesImpl defines the Quote API used by the core of GnuCash.  A quote source
+ * such as Finance::Quote implements the GncQuotesSource API and this class translates
+ * between that API and the public API this class defines for the rest of GnuCash.
+ **/
 class GncQuotesImpl
 {
 public:
@@ -104,8 +113,14 @@ public:
     void report (const char* source, const StrVec& commodities, bool verbose);
 
     const std::string& version() noexcept { return m_quotesource->get_version(); }
-    const QuoteSources& sources() noexcept { return m_sources; }
-    GList* sources_as_glist ();
+ 
+    const QuoteSources& currency_sources() noexcept { return m_currency_sources; }
+    GList* currency_sources_as_glist ();
+    const QuoteSources& single_sources() noexcept { return m_single_sources; }
+    GList* single_sources_as_glist ();
+    const QuoteSources& multiple_sources() noexcept { return m_multiple_sources; }
+    GList* multiple_sources_as_glist ();
+ 
     bool had_failures() noexcept { return !m_failures.empty(); }
     const QFVec& failures() noexcept;
     std::string report_failures() noexcept;
@@ -119,24 +134,35 @@ private:
     GNCPrice* parse_one_quote(const bpt::ptree&, gnc_commodity*);
 
     std::unique_ptr<GncQuoteSource> m_quotesource;
-    QuoteSources m_sources;
+
+    QuoteSources m_currency_sources;
+    QuoteSources m_single_sources;
+    QuoteSources m_multiple_sources;
+    
     QFVec m_failures;
     QofBook *m_book;
     gnc_commodity *m_dflt_curr;
 };
 
+/**
+ * GncFQQuotesSource implements the GncQuotesSources API for Finance::Quote. 
+ **/
 class GncFQQuoteSource final : public GncQuoteSource
 {
     const bfs::path c_cmd;
     std::string c_fq_wrapper;
     std::string m_version;
-    StrVec m_sources;
+    StrVec m_currency_sources;
+    StrVec m_single_sources;
+    StrVec m_multiple_sources;
     std::string m_api_key;
 public:
     GncFQQuoteSource();
     ~GncFQQuoteSource() = default;
     const std::string& get_version() const noexcept override { return m_version; }
-    const StrVec& get_sources() const noexcept override { return m_sources; }
+    const StrVec& get_currency_sources() const noexcept override { return m_currency_sources; }
+    const StrVec& get_single_sources() const noexcept override { return m_single_sources; }
+    const StrVec& get_multiple_sources() const noexcept override { return m_multiple_sources; }
     QuoteResult get_quotes(const std::string&) const override;
 private:
     QuoteResult run_cmd (const StrVec& args, const std::string& json_string) const;
@@ -151,7 +177,7 @@ static const std::string empty_string{};
 
 GncFQQuoteSource::GncFQQuoteSource() :
 c_cmd{bp::search_path("perl")},
-m_version{}, m_sources{}, m_api_key{}
+m_version{}, m_currency_sources{}, m_single_sources{}, m_multiple_sources{}, m_api_key{}
 {
     {
         // Call Finance::Quote wrapper with info flag
@@ -212,10 +238,21 @@ m_version{}, m_sources{}, m_api_key{}
 
             PWARN("TAG: %s", info["currency_modules"]["AlphaVantage"][0].c_str());
             m_version = std::move(version);
-            for (auto& it : info["quote_methods"]) {
-                m_sources.push_back(it.first);
+
+            for (auto& it : info["currency_modules"]) {
+                m_currency_sources.push_back(it.first);
             }
-            std::sort (m_sources.begin(), m_sources.end());
+            std::sort (m_currency_sources.begin(), m_currency_sources.end());
+            
+            for (auto& it : info["quote_modules"]) {
+                m_single_sources.push_back(it.first);
+            }
+            std::sort (m_single_sources.begin(), m_single_sources.end());
+
+            for (auto& it : info["quote_methods"]) {
+                m_multiple_sources.push_back(it.first);
+            }
+            std::sort (m_multiple_sources.begin(), m_multiple_sources.end());
         }
         catch (std::exception &e)
         {
@@ -319,36 +356,59 @@ GncFQQuoteSource::run_cmd (const StrVec& args, const std::string& json_string) c
 
 /* GncQuotes implementation */
 GncQuotesImpl::GncQuotesImpl() : m_quotesource{new GncFQQuoteSource},
-                                 m_sources{}, m_failures{},
+                                 m_currency_sources{}, m_single_sources{}, m_multiple_sources{}, m_failures{},
                                  m_book{qof_session_get_book(gnc_get_current_session())},
                                  m_dflt_curr{gnc_default_currency()}
 {
-    m_sources = m_quotesource->get_sources();
+    m_currency_sources = m_quotesource->get_currency_sources();
+    m_single_sources = m_quotesource->get_single_sources();
+    m_multiple_sources = m_quotesource->get_multiple_sources();
 }
 
 GncQuotesImpl::GncQuotesImpl(QofBook* book) : m_quotesource{new GncFQQuoteSource},
-m_sources{}, m_book{book},
+m_currency_sources{}, m_single_sources{}, m_multiple_sources{}, m_book{book},
 m_dflt_curr{gnc_default_currency()}
 {
-    m_sources = m_quotesource->get_sources();
+    m_currency_sources = m_quotesource->get_currency_sources();
+    m_single_sources = m_quotesource->get_single_sources();
+    m_multiple_sources = m_quotesource->get_multiple_sources();
 }
 
 GncQuotesImpl::GncQuotesImpl(QofBook* book, std::unique_ptr<GncQuoteSource> quote_source) :
 m_quotesource{std::move(quote_source)},
-m_sources{}, m_book{book}, m_dflt_curr{gnc_default_currency()}
+m_currency_sources{}, m_single_sources{}, m_multiple_sources{}, m_book{book}, m_dflt_curr{gnc_default_currency()}
 {
-    m_sources = m_quotesource->get_sources();
+    m_currency_sources = m_quotesource->get_currency_sources();
+    m_single_sources = m_quotesource->get_single_sources();
+    m_multiple_sources = m_quotesource->get_multiple_sources();
 }
 
 GList*
-GncQuotesImpl::sources_as_glist()
+GncQuotesImpl::currency_sources_as_glist()
 {
     GList* slist = nullptr;
-    std::for_each (m_sources.rbegin(), m_sources.rend(),
+    std::for_each (m_currency_sources.rbegin(), m_currency_sources.rend(),
                     [&slist](const std::string& source) { slist  = g_list_prepend (slist, g_strdup(source.c_str())); });
     return slist;
 }
 
+GList*
+GncQuotesImpl::single_sources_as_glist()
+{
+    GList* slist = nullptr;
+    std::for_each (m_single_sources.rbegin(), m_single_sources.rend(),
+                    [&slist](const std::string& source) { slist  = g_list_prepend (slist, g_strdup(source.c_str())); });
+    return slist;
+}
+
+GList*
+GncQuotesImpl::multiple_sources_as_glist()
+{
+    GList* slist = nullptr;
+    std::for_each (m_multiple_sources.rbegin(), m_multiple_sources.rend(),
+                    [&slist](const std::string& source) { slist  = g_list_prepend (slist, g_strdup(source.c_str())); });
+    return slist;
+}
 
 void
 GncQuotesImpl::fetch (QofBook *book)
@@ -1120,14 +1180,34 @@ const std::string& GncQuotes::version() noexcept
     return m_impl->version ();
 }
 
-const QuoteSources& GncQuotes::sources() noexcept
+const QuoteSources& GncQuotes::currency_sources() noexcept
 {
-    return m_impl->sources ();
+    return m_impl->currency_sources ();
 }
 
-GList* GncQuotes::sources_as_glist ()
+GList* GncQuotes::currency_sources_as_glist ()
 {
-    return m_impl->sources_as_glist ();
+    return m_impl->currency_sources_as_glist ();
+}
+
+const QuoteSources& GncQuotes::single_sources() noexcept
+{
+    return m_impl->single_sources ();
+}
+
+GList* GncQuotes::single_sources_as_glist ()
+{
+    return m_impl->single_sources_as_glist ();
+}
+
+const QuoteSources& GncQuotes::multiple_sources() noexcept
+{
+    return m_impl->multiple_sources ();
+}
+
+GList* GncQuotes::multiple_sources_as_glist ()
+{
+    return m_impl->multiple_sources_as_glist ();
 }
 
 GncQuotes::~GncQuotes() = default;
--- a/libgnucash/app-utils/gnc-quotes.hpp
+++ b/libgnucash/app-utils/gnc-quotes.hpp
@@ -103,18 +103,49 @@ public:
      */
     const std::string& version() noexcept;
 
-    /** Get the available Finance::Quote sources as a std::vector
+    /** Get currency sources as a std::vector. Finance::Quote has a default
+     * currency module but may support multiple currency modules.
      *
-     * @return The quote sources configured in Finance::Quote
+     * @return The currency sources configured in Finance::Quote
      */
-    const QuoteSources& sources() noexcept;
+    const QuoteSources& currency_sources() noexcept;
 
-    /** Get the available Finance::Quote sources as a GList
+    /** Get currency sources as a GList
      *
-     * @return A double-linked list containing the names of the installed quote sources.
+     * @return A double-linked list containing the names of the installed currency sources.
      * @note the list and its contents are owned by the caller and should be freed with `g_list_free_full(list, g_free)`.
      */
-    GList* sources_as_glist () ;
+    GList* currency_sources_as_glist ();
+
+    /** Get quote sources that use a single upstream source as a std::vector. Finance::Quote
+     * has "modules" (which gather quotes from a single upstream source) and "methods" (which
+     * use a combination of modules to retrieve quotes, trying one and then
+     * falling back to other modules if the first fails.
+     *
+     * @return The single sources configured in Finance::Quote
+     */
+    const QuoteSources& single_sources() noexcept;
+    
+    /** Get single_sources as a GList.  See single_sources() for description of single sources.
+     *
+     * @return A double-linked list containing the names of the installed Fiannce::Quote modules.
+     * @note the list and its contents are owned by the caller and should be freed with `g_list_free_full(list, g_free)`.
+     */
+    GList* single_sources_as_glist ();
+
+    /** Get list of quotes souces that use multiple upstream sources.  See single_sources() for more details.
+     *
+     * @return A double-linked list containing the names of the available Fiannce::Quote methods.
+     * @note the list and its contents are owned by the caller and should be freed with `g_list_free_full(list, g_free)`.
+     */
+    const QuoteSources& multiple_sources() noexcept;
+    
+    /** Get multiple_sources as a GList.  See multiple_sources() for description of multiple sources.
+     *
+     * @return A double-linked list containing the names of the available Fiannce::Quote methods.
+     * @note the list and its contents are owned by the caller and should be freed with `g_list_free_full(list, g_free)`.
+     */
+    GList* multiple_sources_as_glist ();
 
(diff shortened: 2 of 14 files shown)


jralls:
@vincentl You still have this marked as WIP. Are you still working on it or is it ready for review?

vincentl:
> @vincentl You still have this marked as WIP. Are you still working on it or is it ready for review?

@jralls - This is not done and I'm thinking this PR should just cover getting the info from FQ into GnuCash at startup and no GUI changes.  So it will eliminate the defined list of sources in the GnuCash and get all the data structures in place for future GUI tools.

If this plan sounds good, I think I can get this ready for review in a few days.  I can then open a PR about how to simplify quote source management in the GUI.

jralls:
@vincentl Umm, not quite clear. It's OK to remove the defined list of sources and use only the "unknown" feature (with a new label) *provided it doesn't break existing selections*.

vincentl:
@jralls - I just updated the description to hopefully clarify the intent.  

The top priority for this PR is to make the change transparent to the users.  All working quote sources should continue to work.

jralls:
But you're still not ready for code review, right?

jralls:
@vincentl 

> Merge branch 'Gnucash:stable' into fq-features

Don't do that! Only *rebase* feature branches on the base-branch. Please reset the merge away, rebase on `stable`, and force-push.

vincentl:
> @vincentl
> 
> > Merge branch 'Gnucash:stable' into fq-features
> 
> Don't do that! Only _rebase_ feature branches on the base-branch. Please reset the merge away, rebase on `stable`, and force-push.

@jralls - Sorry!  I'm not used to working through a fork and I used the "Sync Fork" feature it helpfully merged stable into the feature branch.  I think I have correctly reset, rebased, and pushed.

vincentl:
@jralls - It appears that I haven't been making progress because I've tried a few attempts at replacing the hard coded sources lists in `gnc-commodity.c` and am having trouble finding a good solution.  The root of the issue is that `gnc-commodity.c` has 4 lists (currency_quote_source, single_quote_sources, multiple_quote_sources, and new_quote_sources) and the new FQ features is called from `gnc-quotes.cpp`, but the current API between `gnc-quotes.cpp` and `gnc-commodity.c` is a single `sources_as_glist()` function.

I still can't quite understand the division of roles between these two pieces of code.

I am also having trouble understanding the different between `GncFQQuoteSource` and `GncQuotesImpl`.  I can read the different functions these classes provide, but I'm still missing the design idea that resulted in having multiple classes in `gnc-quotes.cpp`.

Is there any documentation beyond code comments to help a new developer?

I want to keep the FQ features data structure in `GncFQQuoteSource` and then route the separate types (single, multi, currency) to `gnc-commodity` but then I need to change the `m_sources` in `gnc-quote`, which brings me back to not quite understanding the design of the class.

Any advice is appreciated and I'll keep experimenting.

jralls:
@vincentl 

There's no up-to-date documentation for developers other than the comments in the code. Many of those comments can be processed with Doxygen to make a nicer format, and we do that run nightly, see https://code.gnucash.org/docs/MASTER/. Unfortunately @derekatkins hasn't fixed the build script for the new branch name so it's a bit out of date, but that should have most of the price-quotes Doxygen comments.

`GncQuotesImpl` provides the API to the rest of GnuCash. `GncQuotesSource` provides an API for quote retrieval to `GncQuotesImpl`, and `GncFQQuotesSource` implements that API for Finance::Quote. We do that to comply with two fundamental (as in CS101-level) design principles, Encapsulation and the Single Responsibility Rule.

`gnc_commodity` encapsulates commodities. All commodities except the home currency have a price and there must be a chain of prices (and GnuCash chokes if that chain is more than two links, e.g. stock->trading currency->home currency). So gnc_commodity has API to set commodity prices. But because gnc_commodity is old code that was written by people who didn't pay much attention in CS101 or maybe never took it, there's also code that belongs somewhere else and some of it, like those hard-coded lists, needs to go away. Even if they were to stay they're for the Commodity Editor so they wouldn't belong in gnc-commodity.c; gnucash/gnome-utils/dialog-commodities.c (while you're at it you also need to look at gnucash/gnome/dialog-commodity.c and the respective GtkBuilder files in gnucash/gtkbuilder) would be a better place and is probably the right place for your replacement code. 

> The root of the issue is that gnc-commodity.c has 4 lists (currency_quote_source, single_quote_sources, multiple_quote_sources, and new_quote_sources) and the new FQ features is called from gnc-quotes.cpp, but the current API between gnc-quotes.cpp and gnc-commodity.c is a single sources_as_glist() function.

Yes, that is indeed the root of the issue and is what you signed up to fix. 

The current code gets a list of available sources from F::Q and compares it against the first 3 lists. Anything that's not in those lists gets added to the `new` list (that's displayed as "Unknown" in the commodity dialog); anything in one of the other lists that's not in the list from F::Q gets marked as disabled. F::Q has changed a lot since @fellen  last updated those lists so they mostly get disabled and most of what's actually available shows up in "unknown".

Don't feel at all constrained by the existing API. If you need to redesign it, do. Same for structs. That `sources_as_glist()` function returns a `GList` of sources. What `source` means in that context is whatever you need it to mean.

gjanssens:
> `GncQuotesImpl` provides the API to the rest of GnuCash. `GncQuotesSource` provides an API for quote retrieval to `GncQuotesImpl`, and `GncFQQuotesSource` implements that API for Finance::Quote. We do that to comply with two fundamental (as in CS101-level) design principles, Encapsulation and the Single Responsibility Rule.
> 
Perhaps I'll just add `GncQuotesImpl` is the internal implementation of the Quotes API of gnucash, the public API that should effectively be used is `GncQuotes`. Please do correct me if my understanding is incorrect.

> `gnc_commodity` encapsulates commodities. All commodities except the home currency have a price and there must be a chain of prices (and GnuCash chokes if that chain is more than two links, e.g. stock->trading currency->home currency). So gnc_commodity has API to set commodity prices. But because gnc_commodity is old code that was written by people who didn't pay much attention in CS101 or maybe never took it, there's also code that belongs somewhere else and some of it, like those hard-coded lists, needs to go away. Even if they were to stay they're for the Commodity Editor so they wouldn't belong in gnc-commodity.c; gnucash/gnome-utils/dialog-commodities.c (while you're at it you also need to look at gnucash/gnome/dialog-commodity.c and the respective GtkBuilder files in gnucash/gtkbuilder) would be a better place and is probably the right place for your replacement code.
> 

Each commodity object has a `quote_source` member variable, which is of type `gnc_quote_source`. As I understood last time I studied the code (which is when I wrote the initial implementation of GncQuotes) this is the same type as used in the lists's we're talking about. The commodity object itself doesn't do anything with this parameter though. In fact as far as I have been able to trace this bit of data is purely informational ?

Informational or not if we keep the commodity's member variable, we will have to keep its type definition somewhere inside the engine. From a high-level perspective this data bit is also strongly related to the GncQuotes code or probably even more so to GncFQQuotesSource as this information is directly derived from querying F::Q. So in that respect I'm tempted to check if we shouldn't merge the useful bits of `gnc_quote_source_...` and the additional feature data you want to query (like needed API keys) into `GncQuotesSource (as generic API) and GncFQQuotesSource (F::Q specific implementation) ? These are currently in app-utils, but would have to move to engine in that case to avoid circular dependencies.

> > The root of the issue is that gnc-commodity.c has 4 lists (currency_quote_source, single_quote_sources, multiple_quote_sources, and new_quote_sources) and the new FQ features is called from gnc-quotes.cpp, but the current API between gnc-quotes.cpp and gnc-commodity.c is a single sources_as_glist() function.
> 

When I did the initial GncQuotes implementation I deliberately didn't change gnc-commodity, to keep my changes somewhat focused. But I definitely had in mind to rework gnc-commodity as well and move the quote_source stuff to somewhere inside GncQuotes (I didn't have the extra classes GncQuotesSource and GncFQQuotesSource yet back then).

> The current code gets a list of available sources from F::Q and compares it against the first 3 lists. Anything that's not in those lists gets added to the `new` list (that's displayed as "Unknown" in the commodity dialog); anything in one of the other lists that's not in the list from F::Q gets marked as disabled. F::Q has changed a lot since @fellen last updated those lists so they mostly get disabled and most of what's actually available shows up in "unknown".

I'm definitely with you it would be much nicer if these lists are autopopulated from information you can query from F::Q itself. That would remove the need to manually maintain these lists.

jralls:
> The commodity object itself doesn't do anything with this parameter though. In fact as far as I have been able to trace this bit of data is purely informational ?

That's what drives retrieving all the quotes for a book--the `Get Quotes` button or `Gnucash::add_quotes`--via `GncQuotes::fetch(QofBook*)` . It gets a list of commodities with quote source set via  [`gnc_commodity_get_quotable_commodities`](https://github.com/Gnucash/gnucash/blob/stable/libgnucash/engine/gnc-commodity.c#L2183). It wends it's way through a few functions, ending up in GncQuotesImpl::comm_vec_to_json_string() which calls [`omm_ns = gnc_quote_source_get_internal_name(gnc_commodity_get_quote_source(comm));
`](https://github.com/Gnucash/gnucash/blob/stable/libgnucash/app-utils/gnc-quotes.cpp#L445).

> Informational or not if we keep the commodity's member variable, we will have to keep its type definition somewhere inside the engine. 

Umm, OK. But it's a struct that we use only two members of, and one of those is a boolean that's true if F::Q is installed and false otherwise. The other is `char* internal_name`. We might as well just use the string and then there's no type that needs to be dealt with.

We could also move it to a lookup table in app-utils, but that seems unnecessarily complex for no real benefit.

gjanssens:
> > The commodity object itself doesn't do anything with this parameter though. In fact as far as I have been able to trace this bit of data is purely informational ?
> 
> That's what drives retrieving all the quotes for a book--the `Get Quotes` button or `Gnucash::add_quotes`--via `GncQuotes::fetch(QofBook*)` . It gets a list of commodities with quote source set via [`gnc_commodity_get_quotable_commodities`](https://github.com/Gnucash/gnucash/blob/stable/libgnucash/engine/gnc-commodity.c#L2183). It wends it's way through a few functions, ending up in GncQuotesImpl::comm_vec_to_json_string() which calls [`omm_ns = gnc_quote_source_get_internal_name(gnc_commodity_get_quote_source(comm)); `](https://github.com/Gnucash/gnucash/blob/stable/libgnucash/app-utils/gnc-quotes.cpp#L445).
> 

That makes sense. I only superficially scanned the code and missed this path.

> > Informational or not if we keep the commodity's member variable, we will have to keep its type definition somewhere inside the engine.
> 
> Umm, OK. But it's a struct that we use only two members of, and one of those is a boolean that's true if F::Q is installed and false otherwise. The other is `char* internal_name`. We might as well just use the string and then there's no type that needs to be dealt with.

I had thought of that as well. I didn't suggest that as I was also considering @vincentl suggestion to store more quote source properties at some point, like whether and which API keys would be needed for each. In that case it may be useful to make a quote source object, or extend your quote source related classes to hold that data. I am only making high level suggestions at this point. I have not looked at the code in detail in a while.
> 
> We could also move it to a lookup table in app-utils, but that seems unnecessarily complex for no real benefit.

Right. Let's see what @vincentl will come up with. My feedback here is only to give a few starting leads based on my previous encounters with this area of the code. It may end up being something completley different :)

jralls:
> I was also considering @vincentl suggestion to store more quote source properties at some point, like whether and which API keys would be needed for each.

That info is very much Finance::Quote specific and therefor needs to be encapsulated in the quotes code, not in the engine. While I usually rant about the evils of canned strings in place of enums, in this case canned strings are the right solution because those strings are very mutable, their mutation is completely out-of-sync with our development cycles, and they're anyway what we have to pass to F::Q.

That also applies to API keys. We can use GSettings to store and retrieve them, but any API should be in the quotes code and the UI should be in the PriceDB window.

vincentl:
@jralls - I had some time over the last few days and was able to get a good start on expanding the notion of "sources" to currency_sources, single_sources, multiple_sources, and unknown_sources in `gnc-quotes.cpp` and `gnc-commodity.c`.  I've run into one issue that I'm hoping you can comment on.

Finance::Quote uses "AlphaVantage" as the default currency module and "alphavantage" as a quote source.  In places where a string is used to find a source, it looks like the type is known, so I wanted to change the look up by name function to `gnc_quote_source_lookup_by_name(const char * name, QuoteSourceType type)`.  This means there can be collisions in the currency and single/multiple source names without causing issues.

However, in `gnc-commodity-sql.cpp` there is the function `set_quote_source_name(gpointer pObject, gpointer pValue)` which is used (I think) as some kind of call-back in at line 82 in the same source file.

In the user data (XML and SQL), is there a record of the quote source kind (single vs multiple)?  Or, is the notion of "commodity" always a security and not a currency?

I pushed my latest changes if that will help in understanding my question.

jralls:
>However, in gnc-commodity-sql.cpp there is the function set_quote_source_name(gpointer pObject, gpointer pValue) which is used (I think) as some kind of call-back in at line 82 in the same source file.

>In the user data (XML and SQL), is there a record of the quote source kind (single vs multiple)? 

Nope, just the name. You can see the SQL schema at https://wiki.gnucash.org/wiki/SQL#Commodities and the XML one at https://github.com/Gnucash/gnucash/blob/e5f9e4ea6d04fb08ce275e5f7b789cdd02266a4c/libgnucash/doc/xml/gnucash-v2.rnc#L159.

> Or, is the notion of "commodity" always a security and not a currency?

No, currency is a subset of commodity so it can in theory have a non-empty price-source field, but in the current design that's hardcoded to 'currency', which is also a special F::Q source that takes two arguments instead of one.

vincentl:
I just pushed commit 5e8d7e that finishes an initial implementation of using the recently added Finance::Quote `get_features()`. In working through these changes, it became clear that I wasn't appreciating some details:
- Finance::Quote has "modules" and "methods"
   - A "module" is a perl module that fetches quotes from a source
   - A "method" is a routine that client code calls to fetch quotes
   - A module typically provides a method that only uses the defining module an optionally registers itself with methods (such as "usa" or "nasdaq" that attempt to fetch quotes by using multiple modules until one succeeds)
- GnuCash distinguishes F::Q methods that fetch from a single source from methods that use multiple modules
- GnuCash uses "currency" as the special name for a quote source that causes the F::Q currency conversion routine to be called
- Finance::Quote has multiple modules for fetching currency information, but there is no notion of multiple currency methods in GnuCash

Commit  5e8d7e
- maintains the distinction between single source and multiple source methods
- stores info about multiple currency methods but doesn't use it in any way
- does not have updated test code, which will need to be adjust for API changes before merge

Questions for @jralls 
- Do you know why there is the distinction between "single" and "multiple" sources?  I was confused for quite a while, thinking it was distinguishing "modules" from "methods", but I recorded my current understanding in the bullets above.
- What is the next step for moving this PR forward?

Note for @bpschuck - you indicating wanting to test GnuCash with the changed discussed in this PR.  I think the code is in a reasonable state for testing, but you would need to build it.

jralls:
@vincentl Dunno why they're separate, but the distinction seems to originate in https://github.com/Gnucash/gnucash/commit/eb868d71d555684c0cabd46f81ee3316f71d221f. Maybe back then F::Q required different code for the two types, or maybe David Hampton thought that it was useful for users to know which sources would try only one source vs. try more than one until it got an answer. If you don't think it makes sense now I have no strong feelings either way.

Next step is code review and test it out, I guess. I'll try to get to that in the next week.

Note on the `currency` source: That's an F::Q keyword for a default currency source. F::Q had only the one until pretty recently. It would be good to update GnuCash to use the newer API, but that's for another PR.

jralls:
@vincentl You need to change gtest-gnc-quotes.cpp to reflect your changes.

libgnucash/engine/gnc-commodity.c
```diff
@@ -381,95 +266,80 @@ gnc_quote_source_add_new (const char *source_name, gboolean supported)
...
-        break;
-
-    case SOURCE_SINGLE:
-        if (index < num_single_quote_sources)
-        {
-            LEAVE("found %s", single_quote_sources[index].user_name);
-            return &single_quote_sources[index];
-        }
-        break;
+        case SOURCE_CURRENCY: source_list = currency_quote_sources; break;
```
jralls: Separate lines for the `case` label and each statement, e.g.
```
case SOURCE_CURRENCY:
    source_list = currency_quote_sources;
    break;
```

libgnucash/engine/gnc-commodity.c
```diff
@@ -381,95 +266,80 @@ gnc_quote_source_add_new (const char *source_name, gboolean supported)
...
+gnc_quote_source *
+gnc_quote_source_lookup_by_name(const char * name)
+{
+    // "currency" is the internal GNC name that causes the F::Q wrapper to call the currency conversion routine
+    if (g_strcmp0(name, "currency") == 0)
+        return gnc_quote_source_lookup_by_name_and_type(name, SOURCE_CURRENCY);
+
+    // Otherwise, the source must be in one of the following since SOURCE_FQ_CURRENCY is not currently in use
+    gnc_quote_source *source = NULL;
+    if (NULL != (source = gnc_quote_source_lookup_by_name_and_type(name, SOURCE_SINGLE)))
```
jralls: Don't put the constant first in a test, and it's not necessary to test explicitly for NULL in this instance anyway.  You could do the whole block in a loop:
```
for (QuoteSourceType type = SOURCE_SINGLE; type < FQ_CURRENCY; type++)
      if (source = gnc_quote_source_lookup_by_name_and_type (name, type))
          return source;
```

libgnucash/app-utils/gnc-quotes.cpp
```diff
@@ -146,37 +174,89 @@ static const std::string empty_string{};
...
+            
+            for (auto& it : info["quote_methods"]) {
+                // Sort methods into single vs multiple sources
+                if (info["quote_methods"][it.first].size() == 1)
+                    m_single_sources.push_back(it.first);
+                else
+                    m_multiple_sources.push_back(it.first);
+            }
+            std::sort (m_single_sources.begin(), m_single_sources.end());
+            std::sort (m_multiple_sources.begin(), m_multiple_sources.end());
```
jralls, line 243: ISTM you've created a lot of complexity by maintaining these as separate containers. Why not a single container of `std::pair<gnc_quote_type, std::string>` and do the separation only once in `gnc_quote_source_add_new()`?

jralls (changes requested):
Somewhere along the line my [original request](https://github.com/finance-quote/finance-quote/discussions/255#discussioncomment-4578804) to have F::Q supply a human-friendly description to put in the source-selection list boxes got lost. Expecting users to know that "aex" is Amsterdam-Euronext and "asx" is the Australian Stock Exchange and to be able to distinguish them on the list is IMO asking a bit much.

The Unknown list box and radio button need to go away, there's no longer any value to them. There *is* value to collecting the unknown sources and the securities that use them and producing a message box containing the list and telling the user that they won't work.

bpschuck:
> Somewhere along the line my [original request](https://github.com/finance-quote/finance-quote/discussions/255#discussioncomment-4578804) to have F::Q supply a human-friendly description to put in the source-selection list boxes got lost. Expecting users to know that "aex" is Amsterdam-Euronext and "asx" is the Australian Stock Exchange and to be able to distinguish them on the list is IMO asking a bit much.

Well, it wasn't a hard request. You plainly stated "it would also be nice to have get_description()". ;) 
Users _could_ use the POD/man pages to see what each source gets data for. That would be SOP for a Perl programmer writing a tool that uses F::Q. But, I could definitely see where something like `get_description()` would be handy for program that somewhat insulates (for lack of a better word) users from the use of Perl.

And for telling users modules won't work, we have begun being more aggressive at identifying non-working modules and removing them if a fix is not anticipated.

jralls:
> Well, it wasn't a hard request. 

But it was the request on which this PR is predicated.  The complete paragraph was

>Along that line it would also be nice to have get_description() and is_multi() methods that could be used to populate a selection list. Lacking that GnuCash has to maintain a hard-coded [array of module names and descriptions](https://github.com/Gnucash/gnucash/blob/maint/libgnucash/engine/gnc-commodity.c#L177). We've done a remarkably poor job of keeping that in sync with the sources actually available.

> Users could use the POD/man pages to see what each source gets data for. That would be SOP for a Perl programmer writing a tool that uses F::Q.

The correspondence between sources and modules is limited. All of the multis are out as are some singles, e.g.
```
$ perldoc Finance::Quote::bet
No documentation found for "Finance::Quote::bet".
$ perldoc Finance::Quote::Bet
No documentation found for "Finance::Quote::Bet".
```
So not only are you suggesting that *users* know about and use perldoc to find out what exchange `bet` retrieves prices from, you're suggesting that they also know about `grep` and `@INC` to find what module `bet` belongs to so they can pass that to perldoc. Not happening for 99% of them.

> And for telling users modules won't work, we have begun being more aggressive at identifying non-working modules and removing them if a fix is not anticipated.

I understand why, but it's not very friendly. Sounds like we need a new type, `SOURCE_DISCONTINUED` and a way to populate that too.

bpschuck:
> > Well, it wasn't a hard request.

Well... to-may-toe, to-mah-toe...

> > Along that line it would also be nice to have get_description() and is_multi() methods that could be used to populate a selection list. Lacking that GnuCash has to maintain a hard-coded [array of module names and descriptions]

I still stand by my interpretation of "**would** be nice", along with "**could** be used". To me, that's a suggestion. Just say'n... 

Big difference from "Please add something like `get_description()` that GnuCash could use to populate the quote source selection list". <insert facetious grin here>

Too much of ambiguous and vague docs, change requests, etc. at the day job causing problems. Me, I usually like to over detail things. I like to say, explain something technical to 4 of my coworkers and it will get interpreted 8 different ways.

> > Users could use the POD/man pages to see what each source gets data for. That would be SOP for a Perl programmer writing a tool that uses F::Q.

> So not only are you suggesting that _users_ know about and use perldoc to find out what exchange `bet` retrieves prices from, you're suggesting that they also know about `grep` and `@INC` to find what module `bet` belongs to so they can pass that to perldoc. Not happening for 99% of them.

I said "But, I could definitely see where something like get_description() would be handy for program that somewhat insulates (for lack of a better word) users from the use of Perl."

Meaning, I said Perl programmers would usually do that when writing their scripts/programs using a module they are not familiar with. **Then** I noted that for programs such as GnuCash where Perl is somewhat "hidden" it could be a good idea. Never suggested or expected GnuCash users to know things like grep and what the `@INC` means. But... `man` or `perldoc Finance::Quote` will then list the different modules that one can do a `man` or `perldoc` on. Again, while it can be done, I do understand a good amount of GnuCash users would be lost. That was sort of implied in my one sentence quoted above, I perhaps should have taken my own advice and added detail but not expecting users of such a program to know about Perl PODs, man pages, and the like.  I'm not arguing, was only making a lighthearted rebuttal about the request/suggestion. 

In all seriousness, `get_description()` is a great idea as far as GnuCash and really won't hurt normal F::Q functionality otherwise. But the devil's advocate in me cannot think of any other library call, function, or library in general that a program makes a call to get a description of it. In a way it's like a call to `printf()` that returns a description of its syntax. ;) 

> > And for telling users modules won't work, we have begun being more aggressive at identifying non-working modules and removing them if a fix is not anticipated.

> I understand why, but it's not very friendly. Sounds like we need a new type, `SOURCE_DISCONTINUED` and a way to populate that too.

In my opinion it's just as easy to remove the module as it is to change what a method/function returns once it's reported as no longer working. Then when GnuCash gets a dynamic list of modules, methods, parameters required, and descriptions, that module will just not show up. And if the module is gone, one can safely assume that source has been discontinued. Now for a fixed list in GnuCash, yes an indicator of the module not functioning is handy. But I thought this change is all about a dynamic list. Or I'm missing something. 

In short, get_description() isn't a problem in my book. 

And now to watch me some SharkWeek and enjoy a scotch.

G'night sir and enjoy the weekend.

bpschuck:
All that said, perhaps have `description` another key returned by `get_features()`? Would that work for you John? An example of what it currently returns is in https://github.com/finance-quote/finance-quote/pull/260. Which was merged back in May.
Thanks.

jralls:
> All that said, perhaps have description another key returned by get_features()? Would that work for you John? An example of what it currently returns is in https://github.com/finance-quote/finance-quote/pull/260. Which was merged back in May.

I imagine it would, but the person to ask is @vincentl as he's the developer of both this and [F::Q#260](https://github.com/finance-quote/finance-quote/pull/260).

vincentl:
> @vincentl You need to change gtest-gnc-quotes.cpp to reflect your changes.

pushed ee5f8fde to fix this test (pending any additional updates need when other comments on this PR are addressed).

libgnucash/engine/gnc-commodity.c
vincentl: @jralls - do gnc developers use anything like clang-format (or similar)?  For now, fixed the case statements in commit e6467a3e.

vincentl:
> > All that said, perhaps have description another key returned by get_features()? Would that work for you John? An example of what it currently returns is in [finance-quote/finance-quote#260](https://github.com/finance-quote/finance-quote/pull/260). Which was merged back in May.
> 
> I imagine it would, but the person to ask is @vincentl as he's the developer of both this and [F::Q#260](https://github.com/finance-quote/finance-quote/pull/260).

This idea made me realize that the hash `get_features()` returns doesn't have enough structure to adequately support this idea.  But, now is a good time to fix that.  I need some time to think about the best change.

@bpschuck - Does opening a new PR for F::Q work for you to discuss proposed change?  Do we have time to make it into the next F::Q release?

libgnucash/engine/gnc-commodity.c
jralls: One of the items on my very long queue--one without much priority--is to set up a formatting pre-commit hook. I've set up the formatting rules for GnuCash in my emacs file. In the past we've used [Artistic Style](https://astyle.sourceforge.net/0) with the command `astyle -xd -k1 -m0 -M60 -xL -xC79 -OHpUcZns4 --brackets=break`.

bpschuck:
> @bpschuck - Does opening a new PR for F::Q work for you to discuss proposed change? Do we have time to make it into the next F::Q release?

@vincentl - Please do not open a PR until changes are ready (or close to ready) to be merged. Discussion can be on finance-quote-devel@lists.sourceforge.net or https://github.com/finance-quote/finance-quote/discussions. My least preferred would be an issue (Github discussions can be easily changed to issues).

Timing of F::Q releases usually depends on seriousness of issues that are fixed in the code but not yet pushed to CPAN. That said, I have been toying with idea of releasing 1.58 this coming weekend since it has a Yahoo_JSON fix. Depends if I decide to fix the Bloomberg module this week.

vincentl:
@jralls - Sorry this PR has stalled.  I opened a PR with Finance::Quote to add the description field (https://github.com/finance-quote/finance-quote/pull/334) but @bpschuck pointed out that short description fields weren't adding that much info for end users.  What were you thinking a description would look like?

jralls:
@vincentl Just a human-readable expansion of the source like the hard-coded list has now. I'll comment on the FQ PR.

libgnucash/app-utils/gnc-quotes.cpp
```diff
@@ -151,35 +153,77 @@ GncFQQuoteSource::GncFQQuoteSource() :
...
+                m_sources.push_back(it.first);
+            }
+            std::sort (m_sources.begin(), m_sources.end());
+        }
+        catch (std::exception &e)
+        {
+            std::string err{bl::translate("Finance::Quote failed in parsing quote sources ")};
+            err += e.what();
+            throw(GncQuoteException(err));
+        }
```
jralls, line 250: This isn't necessary and adds weight, let the caller handle the original GncQuoteExceptions.

libgnucash/app-utils/gnc-quotes.cpp
```diff
@@ -151,35 +153,77 @@ GncFQQuoteSource::GncFQQuoteSource() :
 c_cmd{bp::search_path("perl")},
 m_version{}, m_sources{}, m_api_key{}
 {
-    char *bindir = gnc_path_get_bindir();
-    c_fq_wrapper = std::string(bindir) + "/finance-quote-wrapper";
-    g_free(bindir);
-    StrVec args{"-w", c_fq_wrapper, "-v"};
-    auto [rv, sources, errors] = run_cmd(args, empty_string);
-    if (rv)
     {
```
jralls, line 171: This block is superfluous. Remove the curly braces and reindent.

jralls:
The 4th commit message says in part "Incomplete - need to clarify how quote sources are recorded in user data files - is both the name and type (currency, single, multiple) recorded in the data file"

Quote information is recorded like
```
  <cmdty:get_quotes/>
  <cmdty:quote_source>yahoo_json</cmdty:quote_source>
  <cmdty:quote_tz/>
```

Sticking questions in commit messages isn't the best way to get answers.

libgnucash/app-utils/gnc-quotes.cpp
```diff
@@ -80,16 +80,25 @@ struct GncQuoteSourceError : public std::runtime_error
 CommVec
 gnc_quotes_get_quotable_commodities(const gnc_commodity_table * table);
 
+/**
+ * GncQuotesSource provides an API for quote retrieval used by GncQuotesImpl.
+ **/
```
jralls: Please remove all of the doxygen comments. We put those in header files in GnuCash, and gnc-quotes.hpp already has the documentation for the public API. API declared in implementation (e.g. .cpp) files is implementation detail and should not be revealed in doxygen.

Even plain comments that just restate what's obvious from reading the code are superfluous noise and should be avoided, and that's the case for all of these comments.

libgnucash/app-utils/gnc-quotes.cpp
```diff
@@ -319,36 +356,59 @@ GncFQQuoteSource::run_cmd (const StrVec& args, const std::string& json_string) c
...
 GList*
-GncQuotesImpl::sources_as_glist()
+GncQuotesImpl::currency_sources_as_glist()
 {
     GList* slist = nullptr;
-    std::for_each (m_sources.rbegin(), m_sources.rend(),
+    std::for_each (m_currency_sources.rbegin(), m_currency_sources.rend(),
                     [&slist](const std::string& source) { slist  = g_list_prepend (slist, g_strdup(source.c_str())); });
     return slist;
 }
```
jralls, line 381: Make a helper function `quotesource_to_glist(const QuoteSources& source)` that does the work and call it from each of `currency_sources_as_glist`, `single_sources_as_glist`, and `multiple_sources_as_glist`.

libgnucash/app-utils/gnc-quotes.hpp
```diff
@@ -103,18 +103,49 @@ class GncQuotes
...
+     * @return A double-linked list containing the names of the installed Fiannce::Quote modules.
+     * @note the list and its contents are owned by the caller and should be freed with `g_list_free_full(list, g_free)`.
+     */
+    GList* single_sources_as_glist ();
+
+    /** Get list of quotes souces that use multiple upstream sources.  See single_sources() for more details.
+     *
+     * @return A double-linked list containing the names of the available Fiannce::Quote methods.
+     * @note the list and its contents are owned by the caller and should be freed with `g_list_free_full(list, g_free)`.
+     */
```
jralls: Pasted in the wrong comment, this one returns vector.

libgnucash/engine/gnc-commodity.c
```diff
@@ -301,73 +192,51 @@ gnc_quote_source_fq_version (void)
...
 
-
 /********************************************************************
  * gnc_quote_source_add_new
  *
- * Add a new price source. Called when unknown source names are found
- * either in the F::Q installation (a newly available source) or in
- * the user's data file (a source that has vanished but needs to be
- * tracked.)
+ * Add a new price source. Called from withing this source file to
```
jralls: s/withing/within/

libgnucash/engine/gnc-commodity.c
```diff
@@ -301,73 +192,51 @@ gnc_quote_source_fq_version (void)
...
+        case SOURCE_MULTI:
+            new_source->index = g_list_length(multiple_quote_sources);
+            multiple_quote_sources = g_list_append(multiple_quote_sources, new_source);
+            break;
+        case SOURCE_UNKNOWN:
+            new_source->index = g_list_length(unknown_quote_sources);
+            unknown_quote_sources = g_list_append(unknown_quote_sources, new_source);
+            break;
+    }
+
```
jralls: This isn't going to work: The source type isn't known by callers, and the design allows a source name to be used for each of the types. The names should be unique, so use hash tables and check `g_hash_table_contains` on all of them before inserting. GHashTable has an optimization for hash tables where the key and value are the same so it will be reasonably efficient. You can use `g_hash_table_get_keys` when you need a list, just be careful with the list's contents because it belongs to the hash table. You might consider having GncFQQuotesSource construct the hash tables instead of intermediate lists.

Better yet extract all of the quote source stuff, both from here and gnc-quotes.cpp into a new engine/gnc-quote-source.[h|c]pp to get rid of all of the passing lists around, plus better memory management and nicer loop code because C++.

libgnucash/engine/gnc-commodity.c
```diff
@@ -155,7 +155,21 @@ struct gnc_quote_source_s
...
+ * name as the indicator to call the F::Q->currency function.
+ *
+ * Here, currency_quote_sources is a 1-long list for the "currency" source.
+ * fq_currency_quote_sources stores the list of available currency sources assigned
+ * as reported by F::Q and can be used in the future to enable more user control of the 
+ * currency source.
+ **/
+
 static GList *currency_quote_sources = NULL;
+static GList *fq_currency_quote_sources = NULL;
```
jralls: Multiple currency sources should be its own commit.

jralls (changes requested):
Commit messages for anything non-trivial should be a >= 79 character summary followed by a blank line and a fuller description. Trivial ones can be just a summary if it will be less that 79 characters. The exception to the rule is bug fixes, but that doesn't apply here.

vincentl:
@jralls - I'm having second thoughts about the FQ changes based on your [comment](https://github.com/finance-quote/finance-quote/pull/334#issuecomment-1712962434)

I find the FQ module vs method design to be confusing even after working on the project for a while and trying to work on this integration into GnuCash is highlighting the confusion. I think of a module as the mechanism for pulling information from a source and a method as a way to use the module. Since multiple modules can link themselves to a single method, some methods use multiple sources. GnuCash cares about methods and tracks single vs multi-source methods separately.

But my current PR for FQ adds a description string to the **module**! I'm thinking what GnuCash needs is FQ to provide (in your words) "a human-friendly short name for each" **method** (my word).  Right?

Continuing with this line of thinking, when a module registers a method it should provide the human-friendly short name. Then single-source methods will have a single short name to display to the user in GnuCash and multi-source methods will have an ordered list of short names.

For example, AEX.pm might do

```perl
my $display = "Euronext Amsterdam";
sub methods { 
  return (dutch => {method => \&aex, display => $display},
              aex => {method => \&aex, display => $display})
}
```

Interested in your thoughts.

vincentl:
@jralls - how do you prefer the individual conversations for the code review be resolved?  Should I address each on a commit and reference the commit replying to each comment?

jralls:
>  @jralls - how do you prefer the individual conversations for the code review be resolved? Should I address each on a commit and reference the commit replying to each comment?

That depends. We want to end up with a series of commits to lead the reader through a stepwise (re)implementation of an object or process. 

For simple editorial changes you can just make a bunch of fixes and commit the lot; you can even go back to the original commit that made the particular change, fix it, and then force push the result.

Substantial architectural changes on the other hand deserve a commit on their own; if the result is that an earlier change in the PR is replaced then ultimately that commit should be removed. Note *ulitimately*: The review comments are implemented as notes on a commit so if you remove a commit the comments go with it. (Comments like this one made on the conversation page aren't affected by removing commits.) If we're having an ongoing discussion about some section of code it's better to keep the old commits until we're satisfied that we've gotten it right, then do a final rebase to squash out the intermediate commits.

jralls:
@vincentl 

>Continuing with this line of thinking, when a module registers a method it should provide the human-friendly short name. Then single-source methods will have a single short name to display to the user in GnuCash and multi-source methods will have an ordered list of short names.

I agree. GnuCash doesn't care about the modules, that's an F::Q implementation detail to us. Your extra level in Quote.pm's  %METHODS is as good a way of dealing with it as any, and offers an easy way to extend it.

@bpschuck What do you think?

bpschuck:
I prefer to utilize the KISS principle. While there is a difference in Module vs Method, for the most part a module's main (and usually only) method that returns the quote data is the same name. The methods allow GnuCash a way of checking different sources (utilizing different modules) that can return data for the given method name (nyse, amex, etc.). 

Starting down the rabbit hole of returning a description per method can start to get complex if trying retrieve a description for those methods that are defined in multiple modules.

While not perfect, I can see something where the description is module based for the methods that are main method in each module. Then for those methods that are defined in multiple modules, the description lists the different modules it can possibly draw data from.

In this scenario, `alphavantage` description may be "From AlphaVantage.co, requires API Key. NYSE, Nasdaq, and foreign exchange data.". But since `nyse` is defined in multiple modules, its description may be "Failover type method. Will try alphavantage, fool, yahooweb, googleweb, and yahoo_json."

This recent possible redesign or rethinking of the solution is exactly why I have been hesitant in merging [F::Q PR 335](https://github.com/finance-quote/finance-quote/pull/334). 

I hope you both had a good Thanksgiving.

vincentl:
@bpschuck - After all of this discussion, I think I've arrived at a simple design change to FQ - see [comment](https://github.com/finance-quote/finance-quote/pull/334#issuecomment-1826417507)

vincentl:
@bpschuck - Thanksgiving was nice - I hope you had a nice one too.

jralls:
> Starting down the rabbit hole of returning a description per method can start to get complex if trying retrieve a description for those methods that are defined in multiple modules.

Indeed. Maybe it makes more sense to jettison the multiple methods altogether. As you note it's not a great design and on top of that the implementation leaves much to be desired. For example, `usa` checks `alphavantage`, `fool`, and `IEXCloud`; `nyse` and `nasdaq` check those plus `googleweb` (and make no distinction about whether the symbol trades on the mentioned exchange). `europe` checks `asegr` and `tradegate`. If the multiple methods exist solely for GnuCash then maybe they should be in GnuCash's code instead of Finance::Quote's.

vincentl:
The history of modules vs methods predates my involvement by many years, so I'm not sure the lineage.  Where methods added and then GnuCash used it?  Or, the other way around?

bpschuck:
> If the multiple methods exist solely for GnuCash then maybe they should be in GnuCash's code instead of Finance::Quote's.

> The history of modules vs methods predates my involvement by many years, so I'm not sure the lineage. Where methods added and then GnuCash used it? Or, the other way around?

As far as I can determine, the failover ability has been part of F::Q for a long time. So it may not have been a feature solely for GnuCash.

jralls:
> As far as I can determine, the failover ability has been part of F::Q for a long time. So it may not have been a feature solely for GnuCash.

Umm, not quite. Finance::Quote came into existence in [April 2000](https://github.com/finance-quote/finance-quote/commit/15786ebf9621219dee050729f661feb68ec50fa1); it was forked from GnuCash's Quote.pm which was removed [a few months later](https://github.com/Gnucash/gnucash/commit/ecad85249174240675e59b0c093bc02be361b5fe). The [addition of the failover code](https://github.com/finance-quote/finance-quote/commit/31811310a068f5a36e1c4d79192fc3a2f36c3599) happened in between. It seems unlikely that anyone else was using F::Q at that point. [This thread](https://lists.gnucash.org/pipermail/gnucash-devel/2000-July/000176.html) on gnucash-devel has a bit of the history. 

But that's not really relevant to now. If GnuCash is the only program using failover, @bpschuck wants to lose it and @vincentl  is willing to write the replacement code then the functionality can move into GnuCash. If non-GnuCash users want those methods then it should stay.

bpschuck:
On 11/25/23 3:42 PM, John Ralls wrote:

> But that's not really relevant to now. If GnuCash is the only program
>  using failover, @bpschuck <https://github.com/bpschuck> wants to
> lose it and @vincentl <https://github.com/vincentl> is willing to
> write the replacement code then the functionality can move into
> GnuCash. If non-GnuCash users want those methods then it should
> stay.

I never said I want to lose, not sure what I said that may have been 
interpreted as such. But yes, a good point. Whatever changes are made 
need to be backwards compatible. I've run into a few non-GnuCash users 
while addressing issues and pull requests over the last 3 or 4 years. 
The failover methods stay.

vincentl:
My latest proposal at [FQ PR comment](https://github.com/finance-quote/finance-quote/pull/334#issuecomment-1826417507) would enable GnuCash to filter out methods that have multiple sources.  GnuCash can then create a GUI interface that allows GnuCash users to specify failover per security with a preferred order.

So if we move forward with the proposal on FQ, it will be backward compatible with the existing FQ API but expose the information GnuCash needs to do the filtering as well as obtain the concise "display names" per method.

libgnucash/engine/gnc-commodity.c
vincentl: fixed in ca574dad5f

libgnucash/engine/gnc-commodity.c
vincentl: resolved in 47f3a224fa

libgnucash/app-utils/gnc-quotes.cpp
vincentl: resolved in 66ecde69f6

libgnucash/app-utils/gnc-quotes.hpp
vincentl: resolved in 58b48ffbe8

libgnucash/engine/gnc-commodity.c
christopherlam, line 1: As discussed in #1866 this commit can be cherry picked. I haven't followed this thread closely. Would you be happy if this were merged into stable and #1866 rebased on it?

vincentl: Just to clarify - are you proposing just to cherry pick the change that removes `old_internal_name` or something more extensive that includes changes to `finance-quote-wrapper.in`?

I got stuck on this PR because it depends on both GnuCash and Finance Quote accepting changes and I could quite get to consensus.  So changes to `finance-quote-wrapper.in` and what it returns to GnuCash is tied to getting the associated PR on Finance Quote accepted.

But, if you just want the `old_internal_name` removal, I think you can safely grab it now.

christopherlam: > But, if you just want the `old_internal_name` removal, I think you can safely grab it now.

Correct -- it's incorporated into #1866. Would it be a risk that an old datafile could have a stock which refers to the capitalized old internal name though?

jralls: It would be a very old datafile, `old_internal_name` was introduced int [1.9.0](https://github.com/Gnucash/gnucash/commit/b9bf61723a8b0b1b2820d4ec7503759d2c128227).

gjanssens: ... and hence the file should first be opened with all intermediate major releases before we support opening it in 5.x

christopherlam: Ok I hadn't seen any scrub which converts old to new internal name anywhere though.

gjanssens: The 'scrub' happens automatically when saving the data file. GnuCash just switched to writing the new internal name from that point on. That would not be the case for sql based datafiles, but the sql backend didn't exist yet back then. So the sql backend has always used the new internal names.

vincentl:
@jralls - The PR on FQ has stalled and I'm not sure on a path forward, so I'm closing this PR.

PR #1629, opened 2023-05-07
[trep-engine] Refactor used-columns alist into more generic parameters alist

dawansv:
### Justification

This change is prompted by the need to be able to create and store various report parameters that can be generated or calculated once and reused several times throughout the report building functions, such as regexp object or a hashtable containing preprocessed items. The need was indentified in PR #1623 

### Current Situation

Procedure `make-split-table` creates a big alist called `used-columns`. It must originally have been created true to its name, so that each key named after a column would return #t or #f depending on whether a column is enabled or not in the report. 

However, overtime other pairs were added to cover other options that are not actually columns, such as whether links are displayed or whether subtotals-only is selected. Hence the alist has already outlived its original "columns" moniker in its current form.

### Proposed changes

The existing code already uses the moniker "param" in the function used to retrieve a specific value from the alist

```
    (define (column-uses? param)
      (assq-ref used-columns param))
```

Therefore I propose to rename the alist from "used-columns" to the more generic "parameters" for code readability and future extensibility.

The current (now called) "parameters" are already not a one-to-one match to corresponding options. For some of them there is already logic that combines several options, split drop-down list options, or uses global preferences to produce a result for a #t or #f value. 

In the future, the parameters alist can then be used to store other parameters that might be needed to produce the report within make-split-table and other related functions. The parameters alist is then passed as argument to these functions. 

Some of these other functions building the report are called from gnc:trep-renderer, so I propose to move the creation of the alist of "parameters" in to gnc:trep-renderer instead of make-split-table so it can be used in a wider scope. I am then simply adding one argument to make-split-table so it can received the parameters alist.

### Additional cleanups for improved code readability

- Reorganized the items in the alist so they are grouped by the option pagename they relate to.
- Renamed "column-uses? param" to the more generic "report-uses? param" throughout make-split-table

### DIFF
--- a/gnucash/report/trep-engine.scm
+++ b/gnucash/report/trep-engine.scm
@@ -160,7 +160,7 @@ in the Options panel."))
         (cons #\f (G_ "Frozen"))
         (cons #\v (G_ "Voided"))))
 
-(define (sortkey-list split-action?)
+(define (sortkey-list parameters)
   ;; Defines the different sorting keys, as an association-list
   ;; together with the subtotal functions. Each entry:
   ;;  'sortkey             - sort parameter sent via qof-query
@@ -238,7 +238,7 @@ in the Options panel."))
               (cons 'text (G_ "Description"))
               (cons 'renderer-fn (compose xaccTransGetDescription xaccSplitGetParent)))
 
-        (if split-action?
+        (if (assq-ref parameters 'split-action)
             (list 'number
                   (cons 'sortkey (list SPLIT-ACTION))
                   (cons 'split-sortvalue xaccSplitGetAction)
@@ -441,20 +441,20 @@ in the Options panel."))
       (keylist-get-info keylist (car item) 'text)))
    keylist))
 
-(define (SUBTOTAL-ENABLED? sortkey split-action?)
+(define (SUBTOTAL-ENABLED? sortkey parameters)
   ;; this returns whether sortkey *can* be subtotalled/grouped.
   ;; it checks whether a renderer-fn is defined.
-  (keylist-get-info (sortkey-list split-action?) sortkey 'renderer-fn))
+  (keylist-get-info (sortkey-list parameters) sortkey 'renderer-fn))
 
-(define (CUSTOM-SORTING? sortkey split-action?)
+(define (CUSTOM-SORTING? sortkey parameters)
   ;; sortkey -> bool
   ;;
   ;; this returns which sortkeys which *must* use the custom sorter.
   ;; it filters whereby a split-sortvalue is defined (i.e. the splits
   ;; can be compared according to their 'sortvalue) but the QofQuery
   ;; sortkey is not defined (i.e. their 'sortkey is #f).
-  (and (keylist-get-info (sortkey-list split-action?) sortkey 'split-sortvalue)
-       (not (keylist-get-info (sortkey-list split-action?) sortkey 'sortkey))))
+  (and (keylist-get-info (sortkey-list parameters) sortkey 'split-sortvalue)
+       (not (keylist-get-info (sortkey-list parameters) sortkey 'sortkey))))
 
 (define (lists->csv lst)
   ;; converts a list of lists into CSV
@@ -512,8 +512,9 @@ in the Options panel."))
 ;; Default Transaction Report
 ;;
 (define (gnc:trep-options-generator)
-  (define BOOK-SPLIT-ACTION
-    (qof-book-use-split-action-for-num-field (gnc-get-current-book)))
+  (define parameters
+    (list 
+      (cons 'split-action (qof-book-use-split-action-for-num-field (gnc-get-current-book)))))
 
   ;; (Feb 2018) Note to future hackers - this gnc:trep-options-generator
   ;; defines a long set of options to be assigned as an object in
@@ -678,7 +679,7 @@ be excluded from periodic reporting.")
   ;; Sorting options
 
   (let ((ascending-choice-list (keylist->vectorlist ascending-list))
-        (key-choice-list (keylist->vectorlist (sortkey-list BOOK-SPLIT-ACTION)))
+        (key-choice-list (keylist->vectorlist (sortkey-list parameters)))
         (date-subtotal-choice-list (keylist->vectorlist date-subtotal-list))
         (prime-sortkey 'account-name)
         (prime-sortkey-subtotal-true #t)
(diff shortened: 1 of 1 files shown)


jralls:
@christopherlam Any comments? Is this OK to merge?

christopherlam:
Haven't tested personally because I trust test-transaction and it lgtm

dawansv:
Actually let me put this back in draft for a minute. I had to redefine primary-get-info and secondary-get-info in gnc:trep-renderer while keeping it also in make-split-table. I think it's because BOOK-SPLIT-ACTION is also redefined at each level. I think I need to look whether I can move all of this one level up instead of redefining the same thing.

dawansv:
OK so here is where I am with this. Maybe @christopherlam can enlightened me

We have both gnc:trep-renderer and make-split-table both defined at the top level

Right now they already contain duplicate helper procedures opt-val and BOOK-SPLIT-ACTION

https://github.com/Gnucash/gnucash/blob/8e360bbe849afcf90a47c2966800ec40427bd2d1/gnucash/report/trep-engine.scm#L2056-L2059

In my refactor, they both also need to be able to use primary-get-info and secondary-get-info which right now in only in make-split-table https://github.com/Gnucash/gnucash/blob/8e360bbe849afcf90a47c2966800ec40427bd2d1/gnucash/report/trep-engine.scm#L1078-L1096

We could move make-split-table definition inside gnc:trep-renderer so it can inherit all of gnc:trep-renderer's definitions for the above procedures but make-split-table is such a big procedure I think it would make the code more difficult to maintain, so I am not even considering this option.

We could simply add all of these as 4 extra parameters to make-split-table. It's easy and requires very little changes. That's the easy solution but to me it looks a bit messy somehow. It's just an ever growing list of somewhat random arguments.
```
(define (make-split-table splits options parameters
                          custom-calculated-cells
                          BOOK-SPLIT-ACTION
                          opt-val
                          primary-get-info
                          secondary-get-info
                          begindate enddate c_account_1)
```

OR I could cleanup opt-val to make it a true helper function that only uses the arguments passed to it so it can be moved to the top. So instead of `(define (opt-val section name)` that relies on options to already be in scope, we have `(define (opt-val options section name)` with options passed on every call. I personally find this more readable even if it requires an extra options argument each time opt-val is called. It's not like we are paying by the number of characters used.

As for BOOK-SPLIT-ACTION it can either be passed as an argument to primary-get-info and secondary-get-info or these 2 procedures can simply go fetch that value. It's a bit more costly but these procedures are not called that many times, certainly not at every split in loop that I can see. I ran a huge report with the standard subtotal by account and see no difference in time.

That way Primary-get-info and secondary-get-info can also receive a new options parameter and be moved as helper functions to the top level as well out of the way of the main gnc:trep-renderer and make-split-table procedures with no more duplicates.

So in this new version I have done this second solution. The first one is very short (just some parameters to make-split-table) but this one although requiring more changes is cleaner I think. It's all search-replace changes mind you, I certainly wasn't about to do this manually.

So what do you think, first option (passing more arguments to make-split-table) or this one https://github.com/Gnucash/gnucash/pull/1629/commits/e7c6f507b6c2dc7fbbad1396993827f8e288c214 ?

christopherlam:
Is this refactoring truly necessary? I'm aware `used-columns` is no longer an accurate varname; it is simply a holdover from the previous transaction.scm https://github.com/Gnucash/gnucash/blob/abbd3abd32517b3944482b87a6b3aa0f10a02a43/src/report/standard-reports/transaction.scm#L1044

dawansv:
> Is this refactoring truly necessary? I'm aware `used-columns` is no longer an accurate varname; it is simply a holdover from the previous transaction.scm

The renaming of used-columns to parameters and column-uses to report-uses is merely a by-product. That said I think realigning varnames to their use as we change the code anyway goes a long way in improving readability and code maintenance especially if we want more people to submit patches and improvements. 

The underlying need is for trep-engine to offer a more generic way to store and pass parameters as was identified in PR #1623. Whether it would be useful for other reports in the future I cannot say. Although I don't see why trep-engine should only be used by the transaction report. I thought the idea if it being modular was so it could be used for other reports as well. Still possibly it's a new philosophy that adding new features to trep-engine doesn't need to be just for the transaction report, it can be for other reports even a framework for custom ones. For instance isn't the code handling csv export in trep-engine only used in custom reports? I certainly have not seen in in standard reports.

That said even with the refactor, adding new parameters to the list would still require updating trep-engine since we don't yet have a modular way to pass parameters. So that would be a change of philosophy, adding parameters to trep-engine that are not in the transaction report. Or it would require more work to make the parameter list modular. That said the bulk of the other changes in PR #1623 are in the option section which is already modular, so I think these could be added outside of trep-engine by redefining the options in the scm calling trep-engine.

Ultimately none of this is absolutely necessary. Nothing stops me or anybody from simply refactoring the existing trep-engine code in a custom report without trying to stay in sync with trep-engine. It's certainly easier that way as it doesn't require any coordination. The only downside is that we lose the opportunity to build a better common trep-engine. But I can also understand not wanting to add to it so that there isn't even more to maintain later.

christopherlam:
> Although I don't see why trep-engine should only be used by the transaction report. I thought the idea if it being modular was so it could be used for other reports as well.

It's being used in reconcile-report.scm and income-gst-statement.scm

> For instance isn't the code handling csv export in trep-engine only used in custom reports? I certainly have not seen in in standard reports.

It's used in income-gst-statement.scm -- File > Export > Export Report... will offer csv output.

At this time the `used-columns` mirrors the report options db creating a parallel association list whose keys are symbols and adds a few more, instead of the options db whose "keys" are the section & name pair. I do not particularly think it offers any major performance benefit, nor any major readibility improvement. It is currently a mess of options-db settings (e.g. the 'date setting), complex logic (eg 'other-account-name), and intermediate bools which will be combined later on (e.g. 'amount-original-currency).

christopherlam:
TBH if I were to refactor this used-columns I'd do the following:

1. rename all symbols into a combination of section+name eg. instead of `'date` use `'general/report-date`
2. collate all complex logic so that we don't need to push `primary-get-info` and `secondary-get-info`, nor `opt-val` or `BOOK-SPLIT-ACTION` into `make-split-table` i.e. I'd create an element whose key is `'primary-key/renderer-fn`, `'primary-key/split-sort-value` etc.

This may be too much work.

dawansv:
> 2\. collate all complex logic so that we don't need to push `primary-get-info` and `secondary-get-info`, nor `opt-val` or `BOOK-SPLIT-ACTION` into `make-split-table` i.e. I'd create an element whose key is `'primary-key/renderer-fn`, `'primary-key/split-sort-value` etc.

I am not against doing a bit more, in particular that second point. Putting BOOK-SPLIT-ACTION inside parameters would be nice. It's all over the place right now. It would require moving having a first definition of parameters in the option definition section because the whole logic with BOOK-SPLIT-ACTION and sortkey-list etc is used at that level, before gnc:trep-renderer  is even called. Putting BOOK-SPLIT-ACTION in parameters would simply make sortkey-list from the weirdly specific `define (sortkey-list split-action?)` into the more generic `(sortkey-list parameters)` which is what I also need in PR #1623 

This could even be the solution to making parameters modular, by passing it both to the option generator and gnc:trep-renderer from the report definition scm itself, so that it can be customized there then added onto by trep-engine. Then I think I might be able to implement PR #1623 as a separate report without any specific changes to trep-renderer other than the generic changes discussed here. That would be ideal and a further evolution of the modular aspect of trep-engine. 

I'll give that a try to see if what I am saying even makes sense and report back. None of this is time-critical or important, just a learning opportunity and I do appreciate you pushing back at times (as long as I am not taking to much of your time).

dawansv:
> 2\. collate all complex logic so that we don't need to push `primary-get-info` and `secondary-get-info`, nor `opt-val` or `BOOK-SPLIT-ACTION` into `make-split-table` i.e. I'd create an element whose key is `'primary-key/renderer-fn`, `'primary-key/split-sort-value` etc.

Done in this latest version.

> rename all symbols into a combination of section+name eg. instead of 'date use 'general/report-date

Didn't quite do that as there isn't actually a 1:1 correspondence between options and parameters so I think it's overkill + we end up with parameter names that are unnecessary long; but still they are presented as relating to options and some of the names have been improved. More could be done in renaming some of the other opaque variable names such as c_account_0, c_account_1, c_account_2, etc. in gnc:trep-renderer but to me that's less urgent.

christopherlam:
All good!

dawansv:
Thanks but see #1659 for a little additional fix

PR #1631, opened 2023-05-14
Clean up UI files

richardcohen:
and fix the AQBanking menu tooltip


richardcohen:
The test failure is not related to the patch

> The following tests FAILED:
> 	 29 - test-engine (SEGFAULT)
>

jralls:
Thanks!

PR #1634, opened 2023-05-16
Cleanup some gtk usages

richardcohen:


gnucash/gnome/gnc-split-reg.c
```diff
@@ -817,7 +822,6 @@ gnc_split_reg_ld_destroy( GNCLedgerDisplay *ledger )
     }
 
     gnc_ledger_display_set_user_data (ledger, NULL);
-    g_object_unref (gsr);
```
jralls: Maybe it would be better to `g_object_ref(gsr)` when setting the ledger's user data in `gnc_split_reg_init2` so that if the plugin_page_register gets destroyed first `gsr->reg` won't be a use-after-free.

richardcohen: Maybe. I'll have a think about it. I've removed that patch.

jralls:
Thanks!

PR #1637, opened 2023-05-19
Implement option for more aggressive match for Other Split

dawansv:
## Current situation
`xaccSplitGetOtherSplit` in `/libgnucash/engine/Split.c` is the function that currently retrieve the "other split" when provided a split. The other split account is used in register to show the transfer account as well as in various reports (such as the transaction report). Currently the match is conservative. If there is more than 2 splits, xaccSplitGetOtherSplit returns NULL and *Split Transaction* is typically displayed. In the register, the user can see the detail by opening the split view. In reports, not so much (although some have a split view it isn't quite as usable as the transaction view for some reports).

There are already 2 exceptions to the NULL return value when there are more than 2 splits:
1. Trading accounts are ignored, so that the transfer account can be shown in the transaction view even if additional trading account splits exists
2. Split that have been cut by the lot feature are regrouped, also to avoid the *split transaction* message

## Proposed changes
I'd like to propose an additional function called `xaccSplitGetOtherSplitExtended` that offers a more aggressive match to "other split" that what we currently have. This would not replace `xaccSplitGetOtherSplit` in all places but instead could be used in places where a more aggressive match is useful and doesn't have negative side effects, such as in reports. In the register, the current conservative match is probably prudent as it forces the user to open the split view instead of updating the split account in the transaction line without looking at the other splits as well.

In addition to the existing rules, the following rules are added for a more aggressive match:
 * Multiple splits on the same credit/debit side as main split are excluded.
 * Splits with zero value are excluded unless main split is zero as well.
 * A Split on the other side with same value (opposite sign) is matched unless there is more than one with same value.
 * If no obvious single split is found with these extra rules, it still returns NULL of course

## Example
Here is an example of this more aggressive match implemented in the transaction report. 

**First this view shows the example splits as entered in the register:**
image
image

**This is the transaction report with the more aggressive match:**
image
image

**For comparison, same report with the current match:**
image
image

### DIFF
--- a/gnucash/report/trep-engine.scm
+++ b/gnucash/report/trep-engine.scm
@@ -211,13 +211,13 @@ in the Options panel."))
               (cons 'sortkey (list SPLIT-CORR-ACCT-NAME))
               (cons 'split-sortvalue xaccSplitGetCorrAccountFullName)
               (cons 'text (G_ "Other Account Name"))
-              (cons 'renderer-fn (compose xaccSplitGetAccount xaccSplitGetOtherSplit)))
+              (cons 'renderer-fn (compose xaccSplitGetAccount xaccSplitGetOtherSplitExtended)))
 
         (list 'corresponding-acc-code
               (cons 'sortkey (list SPLIT-CORR-ACCT-CODE))
               (cons 'split-sortvalue xaccSplitGetCorrAccountCode)
               (cons 'text (G_ "Other Account Code"))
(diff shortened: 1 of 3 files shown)


dawansv:
This is in draft as it still needs testing, etc. I am just trying to validate the idea before doing more work.

dawansv:
For the moment I have implemented this change in [my custom report "Transaction Report Extended"](https://github.com/dawansv/gnucash-custom-reports)

PR #1638, opened 2023-05-19
Bug 798809 - Multicolumn report error when reopened after saving.

dawansv:
GncOption-set-value needed to be GncOption-get-value


jralls:
Brilliant, thanks. I must have stared at that line 100 times without realizing that it's backwards.

dawansv:
> Brilliant, thanks. I must have stared at that line 100 times without realizing that it's backwards.

Well it helped that I was looking at it from a compare between the old and new code and the correct GncOption-get-value popped up as part of the same refactoring in the next procedure. That was a strong hint something was amiss...
image

PR #1639, opened 2023-05-21
[budget.scm] budget report's actual amounts link to a trep

christopherlam:
showing the exact transactions comprising the *actual* period total.

it works well but not debugged yet.

image

### DIFF
--- a/gnucash/gnome/gnc-budget-view.c
+++ b/gnucash/gnome/gnc-budget-view.c
@@ -1219,9 +1219,6 @@ bgv_get_total_for_account (Account *account, GncBudget *budget, gnc_commodity *n
         }
     }
 
-    if (gnc_reverse_balance (account))
-        total = gnc_numeric_neg (total);
-
     return total;
 }
 
@@ -1317,7 +1314,6 @@ totals_col_source (GtkTreeViewColumn *col, GtkCellRenderer *cell,
     gint period_num;
     gnc_numeric value; // used to assist in adding and subtracting
     gchar amtbuff[100]; //FIXME: overkill, where's the #define?
-    gboolean neg;
     GNCPriceDB *pdb;
     gnc_commodity *total_currency, *currency;
     gnc_numeric total = gnc_numeric_zero ();
@@ -1342,8 +1338,6 @@ totals_col_source (GtkTreeViewColumn *col, GtkCellRenderer *cell,
         currency = gnc_account_get_currency_or_parent (account);
         acctype = xaccAccountTypeGetFundamental (xaccAccountGetType (account));
 
-        neg = gnc_reverse_balance (account);
-
         switch (row_type)
         {
             case TOTALS_TYPE_ASSET_LIAB_EQ:
@@ -1351,7 +1345,6 @@ totals_col_source (GtkTreeViewColumn *col, GtkCellRenderer *cell,
                     (acctype != ACCT_TYPE_LIABILITY) &&
                     (acctype != ACCT_TYPE_EQUITY))
                     continue;
-                neg = !neg;
                 break;
             case TOTALS_TYPE_EXPENSES:
                 if (acctype != ACCT_TYPE_EXPENSE)
@@ -1360,10 +1353,8 @@ totals_col_source (GtkTreeViewColumn *col, GtkCellRenderer *cell,
             case TOTALS_TYPE_INCOME:
                 if (acctype != ACCT_TYPE_INCOME)
                     continue;
-                neg = !neg;
                 break;
             case TOTALS_TYPE_REMAINDER:
-                neg = !neg;
                 break;
             default:
                 continue;       /* don't count if unexpected total row type is passed in... */
@@ -1383,11 +1374,13 @@ totals_col_source (GtkTreeViewColumn *col, GtkCellRenderer *cell,
                         gnc_budget_get_period_start_date (priv->budget, period_num));
         }
 
-        if (neg)
-            value = gnc_numeric_neg (value);
-
         total = gnc_numeric_add (total, value, GNC_DENOM_AUTO, GNC_HOW_DENOM_LCD);
     }
+
+    /* because we want income to be positive, expense and transfer to
+       be negative. i.e. income adds to the budget, expense and
+       transfer deducts from budget */
+    total = gnc_numeric_neg (total);
 
     xaccSPrintAmount (amtbuff, total,
                       gnc_commodity_print_info (total_currency,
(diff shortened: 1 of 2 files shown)


dawansv:
Looking good but I noticed a couple things.

When I get to a trep for a specific budget item, I can click on <back to come back to the main report. However, if I go in the options for the trep and change an option, I can´t go back anymore. Possibly this is by design although a bit off-putting.

The reason I went in the options in the first place is because my trep was showing several currencies, but I wanted it to match the total in the budget which is in common currency -- I don´t think there is a way to make a budget in multiple currencies. Hence I think by default the trep should be in common currency as well (Option - currency - common currency) so that the total match. If you want to be fancy, I suppose you could check for several currencies in the underlying data (but don´t know how difficult that would be), in which case you would select both common currency and show original currency amount.

Obviously none of these are deal breakers.

christopherlam:
> The reason I went in the options in the first place is because my trep was showing several currencies, but I wanted it to match the total in the budget which is in common currency -- I don´t think there is a way to make a budget in multiple currencies. Hence I think by default the trep should be in common currency as well (Option - currency - common currency) so that the total match.

Thanks for testing. And you've uncovered one of the many warts in budget -- the budget account values are denominated into each accounts currency, but summed in the budget editor into the "default currency" in global pref. However the budget values in budget report are shown in account currency.

The least bad choice is to link each budget value into the account's immediate non-commodity currency. e.g. assume your 'default currency' is GBP, and you have Asset[GBP]:USBroker[USD]:Apple[AAPL], a budget in the AAPL account will be denoted in USD.

PS of much greater interest is the thorough testing of budget editor totals (ie bottom 4 rows)

dawansv:
> the budget account values are denominated into each accounts currency, but summed in the budget editor into the "default currency" in global pref.

I totally missed that part but see it now when I increase the account display depth. I suppose then you can't really follow the currency of each level since the trep only allows for either default common currency or original as per each transaction, not whatever currency you want for the whole report. So probably as it is is the best we can do for now.

> PS of much greater interest is the thorough testing of budget editor totals (ie bottom 4 rows)

OK I will test that as well

dawansv:
OK I see that the income shows in positive and expense in negative which is the opposite of how it used to be (just using the regular debit/credit signs) and I assume part of the fix. Now the old "remaining to budget" was broken as well (it seemed to sum the income with the expenses both in negative); that is fixed somewhat but now it just shows zero no matter what. That doesn't seem right (although still preferable than before). 
image

Same budget in the current version shows this, certainly not better
image

christopherlam:
What is your account hierarchy and currencies? And default currency?

dawansv:
> What is your account hierarchy and currencies? And default currency?

It's a pretty standard test file with not much in it, random data. It has mostly USD accounts with a couple Canadian dollar ones. Here it is. The budget I was showing is called "Unnamed Budget" (yes original name) There isn't much in it. Just remove the .LOG from the name I couldn't put it here without adding an approved extension.
[GNC_test.gnucash.log](https://github.com/Gnucash/gnucash/files/11538694/GNC_test.gnucash.log)

christopherlam:
The last desirable change is to delineate income/expense/transfer against remaining to budget because the latter is the sum of the former. A horizontal line above the last row would be nice. But not sure how to draw it. @Bob-IT 🙂

Bob-IT:
> The last desirable change is to delineate income/expense/transfer against remaining to budget because the latter is the sum of the former. A horizontal line above the last row would be nice. But not sure how to draw it. @Bob-IT slightly_smiling_face

@christopherlam 
Have a look at this [gist](https://gist.github.com/Bob-IT/cfb7bdd76b704d824c13b786177f98da)
The commented out part adds a separator function to the tree view and so a line all the way across.
The uncommented part utilises the existing cell data function.

I have not checked the code so if you use please check for mistakes.

christopherlam:
Thanks Bob, how about a simple line separator? I was originally hoping for the excel-style top-gridline separator but it may not be possible.

```patch
modified   gnucash/gnome/gnc-budget-view.c
@@ -99,6 +99,7 @@ enum
     TOTALS_TYPE_INCOME,        /**< This total is Income type*/
     TOTALS_TYPE_EXPENSES,      /**< This total is Expenses type*/
     TOTALS_TYPE_ASSET_LIAB_EQ, /**< This total is Asset/Liab/Equity type*/
+    TOTALS_TYPE_SEPARATOR,     /**< Separator*/
     TOTALS_TYPE_REMAINDER      /**< This total is Remaining to Budget*/
 };
 /**< \brief ENUM for different budget totals types.
@@ -389,6 +390,14 @@ gbv_tree_view_model_row_changed_cb (GtkTreeModel *tree_model, GtkTreePath *path,
     g_idle_add ((GSourceFunc)gbv_totals_tree_view_redraw_idle, priv->totals_tree_view);
 }
 
+static gboolean
+separator_func (GtkTreeModel* model, GtkTreeIter* iter, gpointer user_data)
+{
+    gint type;
+    gtk_tree_model_get (model, iter, 1, &type, -1);
+    return (type == TOTALS_TYPE_SEPARATOR);
+}
+
 /****************************
  * GncPluginPage Functions  *
  ***************************/
@@ -506,11 +515,18 @@ gbv_create_widget (GncBudgetView *budget_view)
     gtk_list_store_set (totals_tree_model, &iter, 0, _("Transfer"),
                                                   1, TOTALS_TYPE_ASSET_LIAB_EQ, 2, " ", 3, " ", -1);
     gtk_list_store_append (totals_tree_model, &iter);
+    gtk_list_store_set (totals_tree_model, &iter, 0, " ",
+                                                  1, TOTALS_TYPE_SEPARATOR, 2, " ", 3, " ", -1);
+    gtk_list_store_append (totals_tree_model, &iter);
     gtk_list_store_set (totals_tree_model, &iter, 0, _("Remaining to Budget"),
                                                   1, TOTALS_TYPE_REMAINDER, 2, " ", 3, " ", -1);
 
     totals_tree_view = GTK_TREE_VIEW(gtk_tree_view_new ());
     priv->totals_tree_view = totals_tree_view;
+
+    gtk_tree_view_set_row_separator_func (GTK_TREE_VIEW(totals_tree_view),
+                                          separator_func, NULL, NULL);
+
     gtk_tree_selection_set_mode (gtk_tree_view_get_selection (totals_tree_view), GTK_SELECTION_NONE);
     gtk_tree_view_set_headers_visible (totals_tree_view, FALSE);
     gtk_tree_view_set_model (totals_tree_view, GTK_TREE_MODEL(totals_tree_model));
```

image

But ideally it would be:
image

PR #1640, opened 2023-05-21
Remove code that duplicates features of G_DEFINE_TYPE

richardcohen:
- declarations of *_init, *_class_init
- implementation of *_parent_class

- also, removed unnecessary NULL checks for destroy, dispose, finalize


jralls:
Thanks!

PR #1641, opened 2023-05-21
Use G_DEFINE_BOXED_* type macros

richardcohen:


jralls:
Looks like you're right about `ab-templates`. Let's remove it instead.

jralls:
Note, `gnc_value_list` removed in #1644.

Thanks!

PR #1642, opened 2023-05-21
BUGFIX: Actions > Online Actions > Show log Window does not open ...

richardcohen:
... when it is first clicked

- do not call gnc_restore_window_size() without a parent, to silence the existing error: 

> ERROR <gnc.gui> gnc_restore_window_size: assertion 'parent != NULL' failed


jralls:
The correct fix for the no-parent warning is to set a parent, not to disable restoring the window size.

richardcohen:
> The correct fix for the no-parent warning is to set a parent, not to disable restoring the window size.

That will need more substantial changes; I'll do it in a separate pull request.

jralls:
Thanks!

PR #1643, opened 2023-05-22
[trep-engine] Option to display full path of linked document

dawansv:
As per request on mailing list, this extends the existing "link" option (from a checkbox to a drop-down list) to also offer the option to display the full path of linked documents in addition to the short L anchor. Also renamed "link" to "linked document" for consistency with menu items.
Would like to know whether this is OK to add before writing tests and updating legacy option selection. Otherwise I can also make this as a custom report.

image

image


Bob-IT:
I am not sure about this but I have a couple of observations, can the column order be changed so the link path is at the end that would at least move the debit and credit values closer to the rest. Secondly the second document link above looks like a relative path and as such is not the 'Full document path', you would need to add the path head to it and am not sure it is available in the reports, it's a preference so maybe.

dawansv:
> can the column order be changed so the link path is at the end that would at least move the debit and credit values closer to the rest.

Thanks, Bob, for your comments. This is not a trivial fix however because right now the columns are split into 2 sections, the "left-columns" in which this is included and the "calculated cells" which are all the amount columns further to the right. Hence putting this all the way to the right would probably require adding a feature for "right-columns" similar to the "left-columns". But that's too big of a change for such a small use case.

dawansv:
> Secondly the second document link above looks like a relative path and as such is not the 'Full document path', you would need to add the path head to it and am not sure it is available in the reports, it's a preference so maybe.

Unfortunately there is no function that returns the full path. It's a bit more complicated than just adding the head because of the various "schemes" supported (file:, http:, etc). The document link dialog code has similar code that would be needed but it does open the link, not just return it. It would be easy enough to create a function to return the full path using similar code but I have no idea where it would live as relating to the report engine. The overall architecture still eludes me in that regard.

https://github.com/Gnucash/gnucash/blob/d214b2f05d014169307a02e523ba4cfb8875309e/gnucash/gnome/dialog-doclink.c#L86-L107

luzpaz:
Typo in PR title. Should be `display` not `dislay`

tctlrd:
This would be  great!
Side topic, documents link paths should to be accessible to the CSV importer and exporter IMHO:
https://bugs.gnucash.org/show_bug.cgi?id=797510

gjanssens:
> > Secondly the second document link above looks like a relative path and as such is not the 'Full document path', you would need to add the path head to it and am not sure it is available in the reports, it's a preference so maybe.
> 
> Unfortunately there is no function that returns the full path. It's a bit more complicated than just adding the head because of the various "schemes" supported (file:, http:, etc). The document link dialog code has similar code that would be needed but it does open the link, not just return it. It would be easy enough to create a function to return the full path using similar code but I have no idea where it would live as relating to the report engine. The overall architecture still eludes me in that regard.
> 
I am aware your attention shifted to other issues by now, though here's my 2 cents.

The general idea is that all code that's not GUI dependent and useful for more than one specific use case should live somewhere in libgnucash. This guideline has not been followed from the start so you will find plenty of examples of code duplication and non-gui logic in the gnucash subdirectory.

For this particular case, as we're talking about pathnames, the first suggestion that comes to mind is to factor out the path construction code from dialog-doclink.c into its own function somewhere in libgnucash/core-utils/gnc-filepath-utils.cpp or gnc-path.c in the same directory (sidenote, I think these could be merged into one single source of truth for file path management as yet another cleanup project). "filepath" in that sourcefile's name may then be too narrow as there can also be URI's in doc-links. So it may need a little bit more shuffling or renaming for best results. This may not work if doc-link's path construction code depends on gnc-uri-utils.c inside libgnucash/engine. I haven't had time to dig that deep, but wanted to give you a starting point.

dawansv:
Thanks @gjanssens for the pointers.

I turns out there was already a function in gnucash/gnome-utils/dialog-doclink-utils.c to retrieve the full path from a doclink uri. 

Trouble is it's not accessible where it needs to be so I ended up moving most non dialog related support functions from dialog-doclink-utils.c to gnc-uri-utils.c. The other files you mentioned (libgnucash/core-utils/gnc-filepath-utils.cpp or gnc-path.c) are for utils serving system files and I don't think it's a good idea to mix linked document stuff in there because it could lead to confusion, I think, since a lot of doclink functions use the doclink pathhead and you wouldn't want that mixed up with system files. 

[Whether these doclink functions should live in gnc-uri-utils.c or get their own gnc-doclink-utils.c I am not sure however.](https://github.com/Gnucash/gnucash/pull/1643/commits/31354df960f4e3915329e1fab1a9844bef832cb8)

jralls:
> moving most non dialog related support functions from dialog-doclink-utils.c to gnc-uri-utils.c. 

Yay! 

> Whether these doclink functions should live in gnc-uri-utils.c or get their own gnc-doclink-utils.c I am not sure however.

The fact that the function names start with `gnc_doclink_` strongly implies that there should be a gnc-doclink.h/c (or cpp). It also suggests a `GncDoclink` struct or class, and I think redesigning around one will generate a much cleaner API  compared to the rather random one you've exported from the dialog.

PR #1646, opened 2023-05-26
Bug 798796 - Account list incomplete in report options

nomis:
Allow stock/fund accounts that are descendants of Bank accounts to be selected for the Advanced Portfolio, Investment Lots and Investment Portfolio reports.


jralls:
Thanks!

PR #1647, opened 2023-05-27
Including brokerage fees in Money Out calculation (if not ignoring them)

fabrica64:
Correcting the Advanced Portfolio report to satisfy the equation "MoneyOut - MoneyIn + Value = TotalGain" when brokerage fees are not ignored.

Previously the report was not considering brokerage fees when selling stocks leading to an incorrect equation.

### DIFF
--- a/gnucash/report/reports/standard/income-gst-statement.scm
+++ b/gnucash/report/reports/standard/income-gst-statement.scm
@@ -226,6 +226,7 @@ with *EUGOODS* in the account description."))
   (GncOptionDBPtr-make-internal options gnc:pagename-display "Amount")
   (GncOptionDBPtr-make-internal options gnc:pagename-display "Sign Reverses")
   (GncOptionDBPtr-make-internal options gnc:pagename-display "Running Balance")
+  (GncOptionDBPtr-make-internal options gnc:pagename-display "Running Totals")
   ;; No multilines allowed
   (GncOptionDBPtr-make-internal options gnc:pagename-display "Detail Level")
   (GncOptionDBPtr-make-internal options pagename-sorting "Show Informal Debit/Credit Headers")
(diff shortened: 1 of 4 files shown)


jralls:
Looks reasonable to me. @mtalexander what do you think?

dawansv:
**EDIT: Never mind. I missed the fact that this change is for fees when selling. My example below is for fees when buying so please disregard for now. See my other comment below.**

I don't think Money out is meant to include internal expenses such as fees. It's the money out to you (the investor) in comparison to the money in.

Look at this example. I am buying 5 stocks at a price of $200 each. There is a $5 brokerage fee. I also enter a dividend later.
image

This is the report as of now (not with the current change):
image
It shows I have put $1000 (the 5 stocks x $200 each) + $5 (brokerage fee) in = $1005; I have not taken any money out, hence that has to be zero. The current stock value is $250, so it shows a gain of (5 * (250-200) - 5 [the fee]) = 245. The total return is 255 because I add the $10 dividend.

Now if I actually sell everything at $250 it changes nothing to the return. The only thing that changes is I see $1250 in money out (which is what I got when I sold the 5 stocks at $250 each) and my gain in now realized instead on unrealized.
image

To me it currently works exactly as expected. But let me do the same exercise with this change...

dawansv:
I guess I must have never noticed possibly because I don't typically pays fees when selling? But taking my previous example, I added a $5 fee when selling. This is the report as it currently shows.
image
The brokerage fee is correctly increased to $10 and reflected in gain and return but it's true that the money out is wrong. It should be $1245.

Here it is with the proposed changes:
image

That looks like the $1245.00 we want indeed. So I say that's a good change! Thanks.

mtalexander:
This seems like a reasonable change.

jralls:
@fabrica64 Did you mean to close this?

fabrica64:
> @fabrica64 Did you mean to close this?

No, I was trying to sync my fork and I did something wrong, sorry...

fabrica64:
> @fabrica64 Did you mean to close this?

I reopened it as PR #1652 
Sorry for it...

PR #1649, opened 2023-05-29
Bug 798695 - Deleting everything from the "Transfer" cell after suggestions pop-up restricts search to the first 30 accounts

nomis:
The combo box list is capable of displaying all accounts if expanded without starting searching, so changing the search string to "" shouldn't only return a small subset of accounts with no way to fix it.

Skip the search and return all accounts in the combo box as it normally does if the account list is opened without searching for something.

### DIFF
--- a/gnucash/register/register-gnome/combocell-gnome.c
+++ b/gnucash/register/register-gnome/combocell-gnome.c
@@ -633,6 +633,20 @@ gnc_combo_cell_type_ahead_search (const gchar* newval,
     gtk_list_store_clear (box->tmp_store);
     unblock_list_signals (cell);
 
+    if (strlen (newval) == 0) {
+        /* Deleting everything in the cell shouldn't provide a search result for
+         * "" because that will just be the first MAX_NUM_MATCHES accounts which
+         * isn't very useful.
+         *
+         * Skip the search show the popup again with all accounts. Clear the
+         * temp store or the cell will be pre-filled with the first account.
+         */
+        gnc_item_list_set_temp_store (box->item_list, NULL);
+        gnc_item_edit_show_popup (box->item_edit);
+        box->list_popped = TRUE;
+        goto cleanup;
+    }
+
     while (valid && num_found < MAX_NUM_MATCHES)
     {
         gchar* str_data = NULL;
@@ -659,6 +673,8 @@ gnc_combo_cell_type_ahead_search (const gchar* newval,
         gnc_item_edit_show_popup (box->item_edit);
         box->list_popped = TRUE;
     }
+
+cleanup:
     g_regex_unref (regex);
     return match_str;
 }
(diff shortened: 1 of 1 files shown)


nomis:
The outcome is the same as manually opening the combo box when entering the cell or when `auto_pop_combos` is enabled.

I'd probably prefer that the combo box be closed again when `!auto_pop_combos` but nothing else in the register interface behaves like that so I've left it open.

dawansv:
I have noticed this behavior as well and it has bugged me (only very mildly) so this is welcome. That said it prompt the following question: why do we limit the match to 30 accounts? I have had situations where I was looking for an account and only remember a part of it, but a lot of other accounts before it match as well; I have accounts for various properties so the property name matches a lof of accounts, more than 30; but it would be nice to see all accounts related to a property when putting the name. Yet right now it stops at 30. Since the full list is displayed and can be scrolled through when just using the arrow, why did we decide to restrict it to 30 when filtering? After all the filtering will always be smaller than the full list; I don't see why I shouldn't be able to scroll through it. If we remove the 30 limit then it would also fix this current bug I imagine.

nomis:
> If we remove the 30 limit then it would also fix this current bug I imagine.

That won't change the fact that we'd still be doing a search for `""`, which shouldn't happen.

dawansv:
> That won't change the fact that we'd still be doing a search for `""`, which shouldn't happen.

That's true. It would hide it from the user experience standpoint but that code definitely needs to be fixed regardless.

dawansv:
> why do we limit the match to 30 accounts?

Quoting myself... The 30 account limit was in the original implementation of this great feature, but I don't see it discussed in the PR. https://github.com/Gnucash/gnucash/pull/664

@jeanlaroche Do you remember why you put the 30 account limit in that feature (great feature by the way) and whether there was a strong case for it or it was just something that felt right? It can be a small annoyance in some cases. Since the filter list is by definition shorter than the full list which can be displayed when opening the list, what's the downside of displaying the whole matching list? The combo box already restricts the view on the screen obviously (one has to scroll down in the list) in the same way it does when opening the full list with the combo box arrow.

In my use case I notice the 30 limit because I have a gnucash file with several properties in it. Each property has a series of subaccounts in various places. For instance sub accounts in Capital Investments in the Asset section, sub accounts in Income section, sub account in Expenses section. That are all under the same heading (the property name) in their respective section (asset, income, expense) so they easily add to more than 30 account especially because for capital investments there is one sub account per year with even more subaccounts. Bottom line is when I type the property name, I see the first 30 accounts related but no more. It's just a great feature for that use, but the 30 limit really makes it less useful. That said I am trying to see whether I am missing the downside of lifting that limit.

jralls:
Thanks!

ISTR that the 30-line limit was to avoid having to scroll the list view, which was a problem in Gtk on Windows at the time.

PR #1651, opened 2023-05-30
[Scrub.c] refactor xaccAccountTreeScrubImbalance

christopherlam:
The scrubbing routines are transaction oriented. Instead of xaccAccountTreeScrubImbalance calling xaccAccountScrubImbalance for each descendant, refactor so that the transaction list is generated only once. The scrubbing is faster and progress bar is more accurate.

- [ ] ~Refactor other scrubbing routines~

### DIFF
--- a/libgnucash/engine/Scrub.c
+++ b/libgnucash/engine/Scrub.c
@@ -43,6 +43,7 @@
 #include <stdio.h>
 #include <string.h>
 #include <stdint.h>
+#include <stdbool.h>
 
 #include "Account.h"
 #include "AccountP.h"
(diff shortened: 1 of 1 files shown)


christopherlam:
An even better idea is to remove the xaccTransBeginEdit/xaccTransCommitEdit pair in `xaccTransScrubSplits` -- it creates a major slowdown for every transaction during scrubbing, and is not necessary at all. Any thoughts @jralls or @derekatkins ?

jralls:
@christopherlam If a split gets changed then the transaction's balancer needs to run and IIRC it's `xaccTransCommitEdit` that does that. So to use it parsimoniously you'd need to test for problems outside of Edit/Commit and wrap only the repairs.

PR #1653, opened 2023-06-01
Remove unused remnants of register2

richardcohen:


gnucash/gnome-utils/gnc-tree-view.h
```diff
@@ -287,40 +287,24 @@ gnc_tree_view_add_pix_column (GncTreeView *view,
...
+(??)gnc_tree_view_add_combo_column (GncTreeView *view,
+(??)                                const gchar *column_title,
+(??)                                const gchar *pref_name,
+(??)                                const gchar *sizing_text,
+(??)                                gint model_data_column,
+(??)                                gint model_visibility_column,
+(??)                                GtkTreeModel *combo_tree_model,
+(??)                                gint combo_model_text_column,
+(??)                                GtkTreeIterCompareFunc column_sort_fn);
+(??)
```
jralls: Looks like you kept the wrong decl here. I realize that you delete it again along with the definition of `gnc_tree_view_add_date_column` but having a build-breaker like this makes bisects difficult.

jralls (changes requested):

gnucash/gnome-utils/gnc-tree-view.h
richardcohen: Yep. Bad merge.

richardcohen:
I've updated the branch, but github doesn't seem to have noticed.

jralls:
Thanks!

PR #1654, opened 2023-06-01
Scrub speedup

christopherlam:
Instead of #1651 which does reduce the number of scrubs dramatically, this branch will augment `xaccSplitScrub` to take a `dry_run` option returning true/false if a scrub must take place.

Most splits wouldn't need scrubbing therefore the slow Begin/Commit can be skipped.

### DIFF
--- a/libgnucash/engine/Scrub.c
+++ b/libgnucash/engine/Scrub.c
@@ -332,6 +332,28 @@ xaccSplitScrub (Split *split)
 /* ================================================================ */
 
 void
+xaccTransScrubSplits (Transaction *trans)
+{
+    if (!trans) return;
+
+    gnc_commodity *currency = xaccTransGetCurrency (trans);
+    if (!currency)
+        PERR ("Transaction doesn't have a currency!");
+
+    xaccTransBeginEdit(trans);
+    /* The split scrub expects the transaction to have a currency! */
+
+    for (GList *n = xaccTransGetSplitList (trans); n; n = g_list_next (n))
+        xaccSplitScrub (n->data);
+
+    xaccTransCommitEdit(trans);
+}
+
+
+/* ================================================================ */
+
+
+void
 xaccAccountTreeScrubImbalance (Account *acc, QofPercentageFunc percentagefunc)
 {
     if (!acc) return;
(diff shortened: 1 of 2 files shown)


christopherlam:
Some benchmarks for scrub_all from main account tree:

* Neither: 280.591s
* #1651 only: 62.690s
* #1654 only: 2.447s
* Both PRs applied: 1.846s

christopherlam:
I found an issue with this branch: the refactor will use QofQuery to in `get_all_transactions` whereas the previous code used `xaccAccountGetSplitList`. The refactor unfortunately includes the blank transaction.

As a result, the blank transaction will be scrubbed, which results in unwanted Orphan accounts.

We can fix this by redefining `get_all_transactions` as follows, which is slightly ugly, but it is still faster than the workload from prior this branch because it skips duplicate transactions. Any better ideas?

```patch
modified   libgnucash/engine/Scrub.c
@@ -96,13 +96,13 @@ get_all_transactions (Account *account, bool descendants)
 {
     GList *accounts = descendants ? gnc_account_get_descendants (account) : NULL;
     accounts = g_list_prepend (accounts, account);
-    QofQuery *q = qof_query_create_for (GNC_ID_TRANS);
-    QofBook *book = qof_session_get_book (gnc_get_current_session ());
-    qof_query_set_book (q, book);
-    xaccQueryAddAccountMatch (q, accounts, QOF_GUID_MATCH_ANY, QOF_QUERY_AND);
-    GList *transactions = g_list_copy (qof_query_run (q));
-    qof_query_destroy (q);
-    return transactions;
+    GHashTable *ht = g_hash_table_new (g_direct_hash, g_direct_equal);
+    for (GList *n = accounts; n; n = g_list_next (n))
+        for (GList *m = xaccAccountGetSplitList (n->data); m; m = g_list_next (m))
+            g_hash_table_add (ht, xaccSplitGetParent (m->data));
+    GList *rv = g_hash_table_get_keys (ht);
+    g_hash_table_destroy (ht);
+    return rv;
 }
```

jralls:
Not just ugly, it'll be slow as heck on a large book with a lot of accounts. The blank transaction should have a non-zero edit level: You could test that and skip the scrub. That would also skip scrubbing any other transaction being edited and ISTM that's right.

PR #1655, opened 2023-06-03
Bug 798564 - GnuCash is slow when there are a lot of open tabs/registers

nomis:
Every time a change to a transaction is saved by navigating away from it, the UI freezes for almost a second.

Register refreshes should be deferred on tabs that are not visible until those tabs are focused again.

Use focus information to suppress the last part of the refresh_handler() if the ledger isn't currently visible. When it becomes visible again, run the deferred refresh step.

----

When loading the book on startup, GnuCash takes a long time. This is partly because gnc_restore_all_state() is loading data for every register page and I have 35+ registers open.

Don't automatically focus pages when they are recreated. This has to disregard a preference to get them to be in the right order so it is specifically only for recreates (and not a "background" or "without focus" way of creating pages).
    
Change the register page so that it doesn't assume it will be in focus on creation. Don't immediately load or refresh the ledger (this will happen the first time it is focused instead).
    
The previously active page is still focused automatically when loading is finished and this will load the ledger if it's a register.

### DIFF
--- a/gnucash/gnome-utils/gnc-main-window.cpp
+++ b/gnucash/gnome-utils/gnc-main-window.cpp
@@ -3035,7 +3035,8 @@ gnc_main_window_connect (GncMainWindow *window,
     priv = GNC_MAIN_WINDOW_GET_PRIVATE(window);
     notebook = GTK_NOTEBOOK (priv->notebook);
 
-    if (gnc_prefs_get_bool (GNC_PREFS_GROUP_GENERAL, GNC_PREF_TAB_OPEN_ADJACENT))
+    if (!priv->restoring_pages
+            && gnc_prefs_get_bool (GNC_PREFS_GROUP_GENERAL, GNC_PREF_TAB_OPEN_ADJACENT))
         current_position = g_list_index (priv->installed_pages, priv->current_page) + 1;
 
     priv->installed_pages = g_list_insert (priv->installed_pages, page, current_position);
@@ -3044,7 +3045,8 @@ gnc_main_window_connect (GncMainWindow *window,
                                    tab_hbox, menu_label, current_position);
     gtk_notebook_set_tab_reorderable (notebook, page->notebook_page, TRUE);
     gnc_plugin_page_inserted (page);
-    gtk_notebook_set_current_page (notebook, current_position);
+    if (!priv->restoring_pages)
+        gtk_notebook_set_current_page (notebook, current_position);
 
     if (GNC_PLUGIN_PAGE_GET_CLASS(page)->window_changed)
         (GNC_PLUGIN_PAGE_GET_CLASS(page)->window_changed)(page, GTK_WIDGET(window));
@@ -3187,7 +3189,8 @@ gnc_main_window_open_page (GncMainWindow *window,
 
     if (gnc_main_window_page_exists(page))
     {
-        gnc_main_window_display_page(page);
+        if (!gnc_main_window_is_restoring_pages (window))
+            gnc_main_window_display_page (page);
         return;
     }
 
--- a/gnucash/register/ledger-core/gnc-ledger-display.c
+++ b/gnucash/register/ledger-core/gnc-ledger-display.c
@@ -66,6 +66,8 @@ struct gnc_ledger_display
 
     gboolean loading;
     gboolean use_double_line_default;
+    gboolean visible; /* focus */
+    gboolean needs_refresh;
 
     GNCLedgerDisplayDestroy destroy;
     GNCLedgerDisplayGetParent get_parent;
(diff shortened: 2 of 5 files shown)


nomis:
I use the following commits for testing the timing of operations and the focus refresh behaviour:
* https://github.com/nomis/gnucash/commit/5f5016b150e91ea0ddc68ef0ba682d7b91aa6c87
* https://github.com/nomis/gnucash/commit/89cf392f5154cce252dd26e9dc5bef605e9adfc7

nomis:
I don't understand why, but the second commit ("Use focus information to suppress the last part of the refresh_handler() if the ledger isn't currently visible. When it becomes visible again, run the deferred refresh step.") has broken the tab fill behaviour of the "Transfer" column when entering a transaction.

If I enter a partial/complete account name **without the popup appearing** and then press tab, it becomes blank so it gets replaced with an Orphan account on commit. If the popup has appeared (even if it's then dismissed) or I press enter then the value stays there.

I also have a problem entering that column on existing transactions. It becomes empty when focused and the drop-down arrow is disabled.

nomis:
> I don't understand why, but the second commit has broken the tab fill behaviour of the "Transfer" column when entering a transaction.

This is because `gnc_ledger_display_internal` no longer calls `gnc_split_register_load`, preferring to wait until the first focus event, but that function does more than just load splits - it also does various first pass stuff like:
```
        /* load up account names into the transfer combobox menus */
        gnc_split_register_load_xfer_cells (reg, default_account);
        gnc_split_register_load_desc_cells (reg);
        gnc_split_register_load_doclink_cells (reg);
        gnc_split_register_load_recn_cells (reg);
        gnc_split_register_load_type_cells (reg);
```

I've fixed this by calling it to load an empty list of splits. I think this will work the same way as before but it's possible that the "go to blank on first pass" behaviour may be subtly different when there are no splits on the first pass.

gnucash/register/ledger-core/gnc-ledger-display.c
```diff
@@ -94,8 +94,8 @@ gnc_ledger_display_internal (Account* lead_account, Query* q,
                              gboolean is_template,
                              gboolean mismatched_commodities);
 
-static void gnc_ledger_display_refresh_internal (GNCLedgerDisplay* ld,
-                                                 GList* splits);
+static void
+gnc_ledger_display_refresh_internal (GNCLedgerDisplay* ld, gboolean first);
```
jralls: Eww. Functions that do different things based on a boolean parameter is poor design.
Instead move the code from `refresh_handler` to `gnc_ledger_display_refresh` and have `refresh_handler` call `gnc_ledger_display_refresh`.

gnucash/gnome-utils/gnc-main-window.cpp
```diff
@@ -3014,12 +3014,16 @@ gnc_main_window_engine_commit_error_callback( gpointer data,
  *
  *  @param menu_label The widget that should be added into the
  *  notebook popup menu for this page.  This should be a GtkLabel.
+ *
+ *  @param recreate The page is being recreated (while loading the book)
+ *  so always add the page the bottom and don't focus it after insertion.
```
jralls: "add the page to the bottom" ?

And note that "the end" is probably better than "the bottom" because the default behavior is for the tabs to be displayed horizontally across the top.

jralls (changes requested):
That last commit needs breaking up into several, one for each piece. `recreate` is even worse than `first` since it seems to travel quite a lot.

I think you're generally on the right track but need to do a more thorough refactoring instead of just bolting on a hack.

gnucash/register/ledger-core/gnc-ledger-display.c
nomis: This is now fixed.

gnucash/gnome-utils/gnc-main-window.cpp
nomis: This code is no longer present.

nomis:
I found `gnc_main_window_is_restoring_pages()` when I was looking for somewhere to put the `recreate` state which makes this a lot simpler.

gnucash/register/ledger-core/gnc-ledger-display.c
```diff
@@ -935,11 +909,38 @@ gnc_ledger_display_refresh (GNCLedgerDisplay* ld)
...
     // Exclude any template accounts for search register and gl
     if (!ld->reg->is_template && (ld->reg->type == SEARCH_LEDGER || ld->ld_type == LD_GL))
         exclude_template_accounts (ld->query, ld->excluded_template_acc_hash);
 
-    gnc_ledger_display_refresh_internal (ld, qof_query_run (ld->query));
+    /* It's not clear if we should re-run the query, or if we should
+     * just use qof_query_last_run().  It's possible that the dates
+     * changed, requiring a full new query.  Similar considerations
+     * needed for multi-user mode.
+     */
```
jralls: This comment goes with what is now line 880.

jralls (changes requested):
Only one commit should have the bug title, but maybe commit 4 can be squashed into commit 2?

gnucash/register/ledger-core/gnc-ledger-display.c
nomis: I've moved it now

jralls (approved):

jralls:
Thanks!

PR #1656, opened 2023-06-03
Bug 777472 - reconcile does not work if transaction selected

nomis:
If there's an existing transaction with pending edits when starting reconciliation, that transaction can appear in the list of transactions and be modified by the reconciliation but as soon as navigation to another transaction occurs the pending edit will be saved, overwriting the change to mark it as reconciled. The result is that reconciliation appears to have failed or not marked the transaction as reconciled.

This commonly happens when marking several transactions as cleared before performing reconciliation. If the last transaction is still being edited it will typically be saved after the reconciliation finishes and overwrite any change made.

<s>Potential modification or removal of transactions affecting reconciliation could be detected by looking for pending edits to transactions involving the account being reconciled but it's not trivial to determine if a pending edit in another register would add the transaction to the account being reconciled.

Check that there's no outstanding activity in any register pages before starting a reconciliation. It is still possible to start modifying a transaction after the reconciliation window is open but this will stop the most common issue with the process.</s>

Check that there's no outstanding activity in the current register page before starting a reconciliation.

It is still possible to start modifying a transaction after the reconciliation window is open but this will stop the most common issue with the process.

Starting a reconciliation from the account tree is left unprotected.

### DIFF
--- a/gnucash/gnome/gnc-plugin-page-account-tree.c
+++ b/gnucash/gnome/gnc-plugin-page-account-tree.c
@@ -1782,6 +1782,17 @@ gnc_plugin_page_account_tree_cmd_reconcile (GSimpleAction *simple,
     account = gnc_plugin_page_account_tree_get_current_account (page);
     g_return_if_fail (account != NULL);
 
+    /* To prevent mistakes involving saving an edited transaction after
+     * finishing a reconciliation (reverting the reconcile state), we could look
+     * at all open registers and determine if any of them have a transaction
+     * being edited that involves the account to be reconciled.
+     *
+     * However, the reconcile window isn't modal so it's still possible to start
+     * editing a transaction after opening it. Assume the user knows what
+     * they're doing if they start a reconciliation from the account tree and
+     * don't attempt to stop them.
+     */
+
     window = GNC_PLUGIN_PAGE (page)->window;
     recnData = recnWindow (window, account);
     gnc_ui_reconcile_window_raise (recnData);
--- a/gnucash/gnome/gnc-plugin-page-register.c
+++ b/gnucash/gnome/gnc-plugin-page-register.c
@@ -4327,6 +4327,17 @@ gnc_plugin_page_register_cmd_reconcile (GSimpleAction *simple,
 
     g_return_if_fail (GNC_IS_PLUGIN_PAGE_REGISTER (page));
 
+    /* To prevent mistakes involving saving an edited transaction after
+     * finishing a reconciliation (reverting the reconcile state), require
+     * pending activity on the current register to be finished.
+     *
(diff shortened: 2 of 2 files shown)


dawansv:
Ironically, I had this very thing happen to me for the first time just one week ago after using GC for over 20 years...

jralls:
I don't see the point of creating the new functions `gnc_main_window*_finish_pending_by_type` instead of simply adding the GType parameter to the existing `gnc_main_window*_finish_pending`. If you default `GType type = G_TYPE_OBJECT` then you won't even have to change the existing calls.

I also wonder what's the point of manually clearing transactions before running reconcile, but I suppose it's harmless.

Somebody will complain that they shouldn't have to finish up an edit of a transaction in order to reconcile an account that isn't involved in the transaction.

nomis:
> I don't see the point of creating the new functions `gnc_main_window*_finish_pending_by_type` instead of simply adding the GType parameter to the existing `gnc_main_window*_finish_pending`. If you default `GType type = G_TYPE_OBJECT` then you won't even have to change the existing calls.

I'll fix this.

> I also wonder what's the point of manually clearing transactions before running reconcile, but I suppose it's harmless.

For some accounts the register view is more convenient than the reconcile view to compare transactions with the statement because it's possible to edit the date/description, expand the transaction, change the amount on transactions in other currencies (which is only finalised when it appears on the statement) etc.

In those cases I could actually do with a "quick reconcile" mode where it just auto-approves it if the amount matches.

For other accounts I do bulk-select them in the reconcile view, because that happens to be easier.

> Somebody will complain that they shouldn't have to finish up an edit of a transaction in order to reconcile an account that isn't involved in the transaction.

Do you want me to add a preference?

jralls:
> Do you want me to add a preference?

And raise @gjanssens ire?

You can solve your specific problem by running `finish_pending` only on the current register. I don't think anyone would complain about that, it's an obvious check. I don't know for sure but I suppose that you decided it was too hard to examine the other splits in every transaction and check for a register open on that account.

nomis:
> You can solve your specific problem by running `finish_pending` only on the current register. I don't think anyone would complain about that, it's an obvious check. I don't know for sure but I suppose that you decided it was too hard to examine the other splits in every transaction and check for a register open on that account.

That's what I did initially, for reconciliation on the current register, but when I added a check for reconciliation on an account from the accounts tree I realised I'd have a lot more checking to do.

It shouldn't examine every transaction; it would be more efficient to check every open register for pending edits to a transaction involving a split on the reconciliation account.

Since I only need the simple case of the current register having outstanding activity to prevent reconciliation I'll do that and leave reconciliation from the accounts tree unprotected.

jralls:
Thanks!

PR #1658, opened 2023-06-04
[gnc-ui-balances.c] better wording for balance limits

christopherlam:
Bikeshedding requested please. @nomis @fellen @Bob-IT 
https://code.gnucash.org/logs/2023/06/04.html#T10:10:04

### DIFF
--- a/libgnucash/app-utils/gnc-ui-balances.c
+++ b/libgnucash/app-utils/gnc-ui-balances.c
@@ -366,9 +366,9 @@ get_balance_limit_info (const Account *account, gboolean icon)
     gboolean lower_limit_reached, higher_limit_reached;
     gboolean lower_is_zero = FALSE;
     gboolean higher_is_zero = FALSE;
-    const char *higher_template = N_("%s balance is %s, exceeds limit of %s.");
-    const char *lower_template = N_("%s balance is %s, subceeds limit of %s.");
-    const char *zero_template = N_("%s balance is %s, and should be zero.");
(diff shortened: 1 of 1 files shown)


dawansv:
> Bikeshedding

I love the irony of this PR removing the obscure word _subceed_ while I had to google _Bikeshedding_ as I wasn't clear what it meant in this context ;-)

Bob-IT:
I think I would just go with 'above set limit' and 'below set limit'

dawansv:
"above higher balance limit of..." sounds odd to me. Shouldn't it be "upper balance limit"? As in "upper limit" not "higher limit"?

Interesting how English uses lower for the opposite of both higher and upper. In French (and probably other languages) we have different words for both: "supérieur/inférieur" for upper/lower (what we would use in this case, talking about the limits) and "plus haut/plus bas" for higher/lower.

christopherlam:
Hence, bikeshedding. "higher balance limit" and "lower balance limit" are the actual option names in the account properties dialog. Any further votes? Or simply "above/below set limit"?

nomis:
The current version repeats "balance" twice, which I feel is unnecessary.

e.g. "above higher limit" and "below lower limit"

jralls:
What fun! I prefer "upper limit" to "higher limit"; it's much more common usage at least in American English. I agree with @nomis that "balance" is redundant. The template can be better still though: "%s balance of %s is above the upper limit %s" and "%s balance of %s should be 0".

PR #1659, opened 2023-06-04
[trep-engine] Define parameters in 2 steps to remove preparam

dawansv:
@christopherlam Ok so there was one little ugliness in PR #1629. To my defence it was still marked as draft ;-)

I was trying to get around sortkey-list needing split-action in the parameters alist before the alist is even created. I did this by using this ugly workaround using a preparam alist with split-actions in it:
https://github.com/Gnucash/gnucash/blob/c9c8d4891033cc464ca0d983f2f7a7ac0ab77072/gnucash/report/trep-engine.scm#L2189-L2190
https://github.com/Gnucash/gnucash/blob/c9c8d4891033cc464ca0d983f2f7a7ac0ab77072/gnucash/report/trep-engine.scm#L2197-L2202

Well needless to say while it works for now, as soon as one needs to add more parameters needed in sortkey-list it falls apart. It became apparent as I was refactoring PR #1623

The solution here is simple however and what I should have done in the first place. It is to build param alist in 2 steps, with the parameters requiring sortkey-list being added in the second step. No more ugly preparam here.

As for custom-sort?, it also needs sortkey-list but since it's used by gnc:trep-renderer only, its definition can simply be moved after the parameters alist is created.

### DIFF
--- a/gnucash/report/trep-engine.scm
+++ b/gnucash/report/trep-engine.scm
@@ -2187,42 +2187,21 @@ be excluded from periodic reporting.")
          (report-title (opt-val gnc:pagename-general gnc:optname-reportname))
          (detail-is-single? (eq? (opt-val gnc:pagename-display optname-detail-level) 'single))
          (split-action? (qof-book-use-split-action-for-num-field (gnc-get-current-book)))
-         (preparam (list (cons 'split-action split-action?)))
          (amount-setting (opt-val gnc:pagename-display (N_ "Amount")))
          (reversed-signs (opt-val gnc:pagename-display (N_ "Sign Reverses")))
          (primary-key (opt-val pagename-sorting optname-prime-sortkey))
          (primary-order (opt-val pagename-sorting optname-prime-sortorder))
          (primary-subtotal (opt-val pagename-sorting optname-prime-subtotal))
          (primary-date-subtotal (opt-val pagename-sorting optname-prime-date-subtotal))
-         (primary-get-info (lambda (info)
-                             (if (memq primary-key DATE-SORTING-TYPES)
-                                 (keylist-get-info date-subtotal-list primary-date-subtotal info)
-                                 (and (SUBTOTAL-ENABLED? primary-key preparam)
-                                      primary-subtotal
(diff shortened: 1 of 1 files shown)


gnucash/report/trep-engine.scm
```diff
@@ -2236,114 +2215,146 @@ be excluded from periodic reporting.")
...
+                             (eq? (opt-val gnc:pagename-display optname-running-totals) 'all))))
+              ;; additional parameters based on sorting options
+              (cons 'primary-key/renderer-fn primary-key/renderer-fn)
+              (cons 'primary-key/split-sortvalue (primary-get-info 'split-sortvalue))
+              (cons 'secondary-key/renderer-fn secondary-key/renderer-fn)
+              (cons 'secondary-key/split-sortvalue (secondary-get-info 'split-sortvalue))
+              (cons 'subtotals-only
+                    (and (opt-val pagename-sorting optname-show-subtotals-only)
+                          (or primary-key/renderer-fn secondary-key/renderer-fn)))
+            )))))
```
christopherlam: Instead of using `parameters` as both the shorter and expanded varname, I'd name the shorter `parameters-tail` and the expanded one `parameters` and use the following construct instead of `append`:

```scheme
(let* ((parameters-tail (list ...))
       (primary-get-info ...))
    (cons* (cons 'running-prime ...)
           (cons 'running-sec ...)
           parameters-tail))
```

It's still a two-step process but is tidier.

dawansv: Interesting. I originally looked at list* but then couldn't find it in guile. Didn't realize it's cons*. Now I see srfi-1 says

> This function is called list* in [Common Lisp](https://srfi.schemers.org/srfi-1/srfi-1.html#CommonLisp) and about half of the Schemes that provide it, and cons* in the other half.

dawansv: updated as per recommendation

christopherlam (approved):

christopherlam:
Is this one ready?

dawansv:
> Is this one ready?

Yes this is good to go. Thanks.

PR #1663, opened 2023-06-06
Y2038 fixes

richardcohen:


libgnucash/engine/gnc-date.h
```diff
@@ -80,9 +77,11 @@ extern "C"
...
 
 /**
- * Many systems, including Microsoft Windows and BSD-derived Unixes
- * like Darwin, are retaining the int-32 typedef for time_t. Since
- * this stops working in 2038, we define our own:
+ * Most systems, including Microsoft Windows, BSD-derived Unixes and
+ * Linux have moved to 64-bit time_t, even on 32-bit architectures.
+ * See https://en.wikipedia.org/wiki/Year_2038_problem
+ *
+ * For historical reasons, we define our own 64-bit time type.
```
jralls: Seems like this advocates for reverting to libc's time functions with a configuration test to ensure that `time()` returns an `int64_t`.

richardcohen: That seems like the kind of change that should only happen in a major release.

Forcing 64-bit time_t would probably cause problems for Debian, who don't seem to have decided how to do their transition yet -- https://groups.google.com/g/linux.debian.bugs.dist/c/BamYnv3caHs. Maybe they will just let their 32-bit desktop die.

Perhaps there could be conditional compilation with a fallback to gnc_time() if still 32-bit?

Is the Windows build 32-bit?

jralls: Yes, it would absolutely a GnuCash6 change. BTW, I was being a bit facetious about reverting to libc POSIX time functions. It doesn't do everything we want and it's not fully supported in MinGW--and not supported at all by MSVC.

But if Debian is still fumbling around with 32-bit time_t on some architectures then ISTM your proposed comment isn't true yet.

Yes, the Windows build is 32-bit.

richardcohen: Is that better?

jralls: Yes, thanks.

jralls:
Thanks!

PR #1665, opened 2023-06-08
Remove unused osx_accel_map

richardcohen:
Since a143f8b5edd3118aedff975d3a1b5b47867e2027 on 2023-03-18, accelerator-map-osx is the file that is used


jralls:
Thanks!

PR #1667, opened 2023-06-13
xaccTransGetTxnType must return even P if non-APAR is zero-value

christopherlam:
Amendment to #1201: A txn_type==P transaction which is later voided will be incorrectly changed to txn_type==L.

~Will create tests soon.~

Tests completed.

### DIFF
--- a/libgnucash/engine/Transaction.c
+++ b/libgnucash/engine/Transaction.c
@@ -2484,7 +2484,7 @@ xaccTransRetDateDue(const Transaction *trans)
 char
 xaccTransGetTxnType (Transaction *trans)
 {
-    gboolean has_nonAPAR_amount = FALSE;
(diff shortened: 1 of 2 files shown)


jralls:
> xaccTransGetTxnType must return even P if non-APAR is zero-value

> Otherwise a txn_type==P transaction which is later voided will be incorrectly changed to txn_type==L.

Should be "…must return TXN_TYPE_PAYMENT even if…"

But ISTM the logic is faulty. What transactions will have only APAR splits?

christopherlam:
> Should be "…must return TXN_TYPE_PAYMENT even if…"
> 
> But ISTM the logic is faulty. What transactions will have only APAR splits?

A transaction which marries up a credit note and an invoice.

PR #1672, opened 2023-06-15
Refactor: Remove some unnecessary uses of memcpy

richardcohen:
- Also, remove unnecessary "static" in gnucash-style.c

The second one in guid.cpp is UB:

libgnucash/engine/guid.cpp:137:5: warning: undefined behavior, source object type 'const gnc::GUID' is not TriviallyCopyable [bugprone-undefined-memory-manipulation]
        memcpy (&target, &source, sizeof (GncGUID));
        ^


jralls:
Thanks!

PR #1673, opened 2023-06-16
[trep-engine] Implement sort and subtotal by substring

dawansv:
As discussed in PR #1623, this creates a more generic feature in trep-engine that allows to select a substring match pattern as a sort and group key (in the sorting options).

Additional options are then presented further down in the Sorting Options dialog.

image

# Features

## Substring Match Pattern

In its simple form, the Match Pattern is a prefix that will match substrings with the same prefix up to the first space or end-of-string.

In its advanced form, the Match Pattern is a regular expression enabled when the 'use regular expression' checkbox is selected. 

The match is case sensitive, so it considers uppercase and lowercase letters separately, unless the 'case insensitive' checkbox is selected.

Only the first match is considered on each split, starting with the split memo, then the transaction notes, and finally the transaction description, similar to how the transaction filter works.

## Use regular expression for substring matching

Checking this box enables full POSIX regular expressions capabilities.

Examples: 
- `#work|#family` will match both #work and #family substrings.
- `#[^ ]*` will match any substring starting with # and ending with a space.

Use parenthesis to identify a submatch to retain for the sorting and grouping.

Examples: `#(work)|#(family)` and `#([^ ]*)` perform the same matches as the previous examples but remove the # from the final sorting and grouping.

Because POSIX doesn't provide word boundary capability, a space is automatically added on each end of the memos, notes and descriptions being searched so that spaces and other character classes can be used for word boundary matches.

Example: `[ ](work)[ ]|[ ](family)[ ]` will match the words work and family but not working and multifamily

## Substring Matching is case insensitive

If this option is selected, substring matching is not case sensitive. The resulting substrings will follow the same capitalization as the Custom Headings (either Title Case, lowercase, UPPERCASE).

## Custom Substring Headings

Use this option to customize the heading for the subtotal group containing the transactions with no matching substring, and optionally, the group containing the transactions with an empty matching substring (as can occur with regexp match).

If the second heading is required, it must be separated by a slash. Default is [No Match]/[Empty String].

# Testing and providing feedback

For testing without compiling or modifying source files, a custom report called Transaction Report Extended (transaction-extended.scm) includes this feature and is available here: https://github.com/dawansv/gnucash-custom-reports

### DIFF
--- a/gnucash/report/trep-engine.scm
+++ b/gnucash/report/trep-engine.scm
@@ -59,6 +59,8 @@
 (use-modules (srfi srfi-9))
 (use-modules (srfi srfi-26))
 (use-modules (ice-9 match))
+(use-modules (ice-9 regex))
+(use-modules (ice-9 i18n))
 
 (export gnc:trep-options-generator)
 (export gnc:trep-renderer)
@@ -98,6 +100,12 @@
 (define optname-sec-subtotal (N_ "Secondary Subtotal"))
 (define optname-sec-sortorder  (N_ "Secondary Sort Order"))
 (define optname-sec-date-subtotal (N_ "Secondary Subtotal for Date Key"))
+(define optname-substring-matcher (N_ "Substring Match Pattern"))
+(define optname-substring-custom-headings (N_ "Custom Substring Headings"))
+(define optname-substring-use-regexp
+  (N_ "Use regular expression for Substring Matching"))
+(define optname-substring-matcher-caseinsensitive
+  (N_ "Substring Matching is case insensitive"))
 
 ;;General
 (define optname-startdate (N_ "Start Date"))
@@ -159,6 +167,63 @@ in the Options panel."))
         (cons #\y (G_ "Reconciled"))
         (cons #\f (G_ "Frozen"))
         (cons #\v (G_ "Voided"))))
+
+;; Create a function which helps find a substring within a split
+;; or transaction as per various user-defined options in pagename-sorting section.
+;; Memorize both the sortvalue and rendering in hashtable for later retrieval
+;; because get-substring-from-split is slow.
+;; Returns either the sort value or rendered value as per sortvalue?
+(define (split-substring split parameters sortvalue?)
+  (let* ((substring-htable (assq-ref parameters 'substring/htable)))
+  (cond
+   ((hash-ref substring-htable split) =>
+    (lambda (found) (if sortvalue? (car found) (cadr found))))
+   (else
+    (let ((sstring (get-substring-from-split split parameters)))
+      (hash-set! substring-htable split sstring)
+          (if sortvalue? (car sstring) (cadr sstring)))))))
+
+;; Finds a substring withing a split as per user-defined options.
+;; Returns a list with both (substring-sortvalue substring-rendered)
+(define (get-substring-from-split split parameters)
+  (let* ((regexp (assq-ref parameters 'substring/matcher-regexp))
+         (no-match-heading (assq-ref parameters 'substring/no-match-heading))
+         (blank-heading (assq-ref parameters 'substring/blank-heading))
+         (change-case (assq-ref parameters 'substring/change-case))
+         (sm (or (regexp-exec regexp
+                  (string-append " " (xaccSplitGetMemo split) " "))
+                 (regexp-exec regexp
+                  (string-append " " ((compose xaccTransGetNotes xaccSplitGetParent) split) " "))
+                 (regexp-exec regexp
+                  (string-append " " ((compose xaccTransGetDescription xaccSplitGetParent) split) " "))))
+         (substring-count (and sm (match:count sm)))
+         (substring-matched
+           ;; If at least one match was returned
+           (if substring-count
+             ;; If there is at least one submatch
+             (if (> substring-count 1)
+               ;; then return a join of all submatches that return a string
+               (string-join
+                 (map (lambda (n) (if (string? (match:substring sm n)) (match:substring sm n) ""))
+                      (cdr (list-tabulate (match:count sm) values)))
+                 "")
+               ;; otherwise return the main match
+               (match:substring sm))
+             #f))
+         (substring-cased
+           (cond ((not substring-matched) #f)
+                 ((equal? (string-trim substring-matched) "") 'first-string)
+                 ((eq? change-case 'upcase) (string-locale-upcase substring-matched))
+                 ((eq? change-case 'downcase) (string-locale-downcase substring-matched))
+                 ((eq? change-case 'titlecase) (string-locale-titlecase substring-matched))
+                 (else substring-matched))))
+    (if substring-cased
+      (list substring-cased
+        (if (eq? substring-cased 'first-string)
+            (if (string-null? blank-heading) (G_ "[Empty String]") blank-heading)
+            substring-cased))
+      (list 'last-string
+            (if (string-null? no-match-heading) (G_ "[No Match]") no-match-heading)))))
 
 (define (sortkey-list parameters)
   ;; Defines the different sorting keys, as an association-list
@@ -268,6 +333,12 @@ in the Options panel."))
               (cons 'split-sortvalue (compose xaccTransGetNotes xaccSplitGetParent))
               (cons 'text (G_ "Notes"))
               (cons 'renderer-fn (compose xaccTransGetNotes xaccSplitGetParent)))
+
+        (list 'substring
+              (cons 'sortkey #f)
+              (cons 'split-sortvalue (lambda (s) (split-substring s parameters #t)))
+              (cons 'text (G_ "Substring"))
+              (cons 'renderer-fn (lambda (s) (split-substring s parameters #f))))
 
         (list 'none
               (cons 'sortkey '())
@@ -696,7 +767,9 @@ be excluded from periodic reporting.")
              (sec-sortkey-enabled (not (eq? sec-sortkey 'none)))
              (sec-sortkey-subtotal-enabled
               (SUBTOTAL-ENABLED? sec-sortkey parameters))
-             (sec-date-sortingtype-enabled (memq sec-sortkey DATE-SORTING-TYPES)))
+             (sec-date-sortingtype-enabled (memq sec-sortkey DATE-SORTING-TYPES))
+             (substring-sortingtype-enabled
+              (or (eq? prime-sortkey 'substring) (eq? sec-sortkey 'substring))))
 
         (gnc-optiondb-set-option-selectable-by-name
          options pagename-sorting optname-prime-subtotal
@@ -754,7 +827,23 @@ be excluded from periodic reporting.")
 
         (gnc-optiondb-set-option-selectable-by-name
          options pagename-sorting optname-sec-date-subtotal
-         sec-date-sortingtype-enabled)))
+         sec-date-sortingtype-enabled)
+
+        (gnc-optiondb-set-option-selectable-by-name
+         options pagename-sorting optname-substring-matcher
+         substring-sortingtype-enabled)
+
+        (gnc-optiondb-set-option-selectable-by-name
+         options pagename-sorting optname-substring-custom-headings
+         substring-sortingtype-enabled)
+
+        (gnc-optiondb-set-option-selectable-by-name
+         options pagename-sorting optname-substring-use-regexp
+         substring-sortingtype-enabled)
+
+        (gnc-optiondb-set-option-selectable-by-name
+         options pagename-sorting optname-substring-matcher-caseinsensitive
+         substring-sortingtype-enabled)))
 
     ;; primary sorting criterion
     (gnc-register-multichoice-callback-option options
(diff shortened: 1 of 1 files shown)


christopherlam:
> Because POSIX doesn't provide word boundary capability, a space is automatically added on each end of the memos, notes and descriptions being searched so that spaces and other character classes can be used for word boundary matches.

IMV this is overkill. I don't think anyone will be offended that "work" search will also match "working". To test whitespace means you'll also want to test punctuation and localised punctuation and whitespace.

dawansv:
> I don't think anyone will be offended that "work" search will also match "working"

Well I was just illustrating a word boundary search. It's meant to illustrate a string within a string. I can use "networker" as another example to make the point. The text in my post comes from the help text in the dialog, so I was trying to be as concise as possible to illustrate word boundary search. But yes "networker" would be better as an example.

 > To test whitespace means you'll also want to test punctuation and localised punctuation and whitespace.

Yes this is just as simplified example. I am assuming at this point people using this know regular expressions or will go look for more information. That said I can certainly add a link to this page in the help tip for instance https://en.wikibooks.org/wiki/Regular_Expressions/POSIX-Extended_Regular_Expressions

Hence that's why I say "spaces and other character classes can be used for word boundary matches". I imagine in practice something like `[^[:alnum:]](work|family)[^[:alnum:]]` is probably better, assuming [:alnum:] adapts to localised form although I have no idea how it deal with no latin alphabets. But that's a POSIX regexp problem, we are just offering the interface here. But point taken I could probably put that in the example instead of the over simplistic `[ ](work)[ ]|[ ](family)[ ]`

The real issue is that without offering some known characters to grab onto at the beginning and end of the full string, there is no way that I know to do reliable word boundary search in POSIX regexp. 

Even the more complex form `^(work|family)[^[:alnum:]]|[^[:alnum:]](work|family)[^[:alnum:]]|[^[:alnum:]](work|family)$` will not work because although it will match both terms at the beginning and end of the full string along with the middle, it will not match when work or family are the entire string and there is no way that I know to add ^(work|family)$ to this regexp. In any case, even if it did work, it's so long and unwieldy that even if it worked it's unusable. There is a reason why other regexp implementations have word boundary match capability.

The same problem already exists in the transaction filter and that is a simple fix there as well.

That said if you can show me a POSIX regexp that will match words with at least some reasonable level of boundaries including beginning and end of string then I stand corrected, but otherwise I don't see why this is overkill considering it literally has no negative side effects and almost no incidence on the code itself.

jralls:
> I am assuming at this point people using this know regular expressions or will go look for more information. That said I can certainly add a link to this page in the help tip for instance https://en.wikibooks.org/wiki/Regular_Expressions/POSIX-Extended_Regular_Expressions

It's not reasonable to expect users to know or learn regular expressions to use a GnuCash feature, but it's fine to allow regular expressions as an option.

Correctly handling non-Latin scripts is required. We have translations using Arabic, Chinese, Cyrillic, Devangari, Hebrew, and Japanese, and that implies that there are users of those scripts who care enough to contribute translations.

POSIX regex, even extended ones, aren't very good with Unicode though so PCRE would be better. We already have several dependencies that provide PCRE support but it means that you have to write the filtering code in C/C++ and SWIG it to make it available to Guile. Unfortunately I think the same is true of Guile string manipulation functions, so even doing a plain string search with UNICODE_IS_SPACE || UNICODE_IS_PUNCTUATION requires a C/C++ implementation.

dawansv:
> It's not reasonable to expect users to know or learn regular expressions to use a GnuCash feature, but it's fine to allow regular expressions as an option.

Yes it's the advanced form that assumes at least some knowledge of regexp, following the same presentation as for the current account and transaction filtering. The basic form doesn't require regular expression use.

> PCRE would be better

Yes and it's unfortunate Guile seems to only offer POSIX support out of the box, unless I am reading this wrong. But unfortunately, I don't have the skills nor time to reimplement that in C++. That said whatever regex support is currently provided in reports such as the account and transaction filters is POSIX only as well, so this is really just providing the same support. I imagine these features already break down with non latin scripts. Whether this means that we don't want to add more features using the same regexp support than what is already offered is not for me to say. 

For all I know there could be a way to check for non latin script use and disable all features using regexp support in that case. I imagine the regexp code might probably already bomb since it has its own error checking code. I wouldn't know and have no real way to check. But this is way beyond the scope of this added feature or at least what I am willing to be involved in.

Possibly this feature is also reaching the realm of another larger point.

I do understand the logic of not wanting to implement new report features if getting rid of Guile is a goal. I mean if there is ever an actual effort to redo these reports in something else besides guile it's going to be a serious undertaking (not that the rest of the code cleanup wasn't). I have no idea how likely this is to happen in the short-term. If this ends up taking years before it happens, and in the meantime we are limiting ourselves to what new features we can implement because Guile is on its way out at some point in the future then that might be quite limiting. As much as we can develop new register features, the reporting ultimately is a big part of what makes an accounting package great. 

At the end, as I have said, if this only lives as a custom report then it is what it is. I am probably going to add other features (unrelated to this) to my custom extended transaction report which I need or have limited use for others, such as PR #1637 and #1643. Some of these probably won't make it in the main report because they offer only a limited range of options, but are good enough for a custom report while upgrading them to full scope requires a lot more work with possibly limited use case.

We could also encourage more sharing of custom reports by publishing a repository of custom reports, or even just links on the official GC page, similar to how some products offers plugins, that make them a bit more accessible to users while making it clear that they are not maintained by the core team and could break at any point. I actually have no idea what other custom reports might be out there if any. I still think that there is value in putting these things out there for users to decide if they like these features or not. It's then easier for users to give feedback and developers to integrate features in the main repository in the future based on how popular a custom report has become. We certainly lack clarity at this point as to what report features are more useful to a larger group of users, and I think encouraging the sharing of custom reports could be one way to achieve that.

At the end however these conversations are starting to take more time than writing the code itself, so this probably has to reach its final step. If there is a go or at least tentative go, then I will write tests for it. Otherwise it's already available as a custom report which I will maintain for the time being. https://github.com/dawansv/gnucash-custom-reports

There truly is no pressure on my side. Either way this won't change my interest in both using or help fix what I can.

christopherlam:
What I meant was: is it truly necessary to restrict search to words rather than word fragments? IMV it's the users responsibility to find a good searchstring.

dawansv:
> What I meant was: is it truly necessary to restrict search to words rather than word fragments? IMV it's the users responsibility to find a good searchstring.

This isn't modifying the searchstring as entered by the user or restricting the search to word fragment. It's just offering an alternative for word boundary match which POSIX doesn't otherwise offer in a simple manner. I can get rid of it if it's such a point of contention, even though I don't understand why since it doesn't have any side effect I can think of.

There is an alternative without temporarily adding spaces at the start and end of the strings being searched. 
One can use the even longer `(^|[^[:alnum:]])(work|family)([^[:alnum:]]|$)`  Unfortunately, because it needs the parenthesis to include the | in the boundary submatches, one can no longer use the parenthesis only for the actual submatch to be retrieved. Also it doesn't work for more complex expression with several variations of | etc. But whatever, I understand this is an edge case. If interested you can see it discussed here as part of a longer discussion (near the middle) https://stackoverflow.com/questions/9792702/does-bash-support-word-boundary-regular-expressions

christopherlam:
If you need PCRE regex how about this for a start. Haven't checked for memory leaks. Even better would be to use std::regex however I haven't tried.

```patch
diff --git a/bindings/app-utils.i b/bindings/app-utils.i
index 96430c3034..94ec3b38e3 100644
--- a/bindings/app-utils.i
+++ b/bindings/app-utils.i
@@ -67,6 +67,13 @@ Account * gnc_get_current_root_account (void);
 
 #if defined(SWIGGUILE)
 
+/* not sure if needed: */
+/* %newobject gnc_regex_new; */
+
+GRegex * gnc_regex_new (const char *pattern);
+bool gnc_regex_match (GRegex *regex, const char *str);
+void gnc_regex_destroy (GRegex *regex);
+
 gnc_commodity * gnc_default_currency (void);
 gnc_commodity * gnc_default_report_currency (void);
 
index d2da97cc78..3444ec0972 100644
@@ -2422,6 +2422,12 @@ be excluded from periodic reporting.")
           (((? from-account?) . _) #t)
           ((_ . rest) (lp rest)))))
 
+    (let ((reg (gnc-regex-new "#[a-z]+")))
+      (pk 'reg reg)
+      (pk 'matches (gnc-regex-match reg "#abc"))
+      (pk 'doesn't-match (gnc-regex-match reg "abc"))
+      (gnc-regex-destroy reg))
+
     (cond
      ((or (null? c_account_1)
           (symbol? account-matcher-regexp)
diff --git a/libgnucash/app-utils/gnc-ui-util.c b/libgnucash/app-utils/gnc-ui-util.c
index 025b9b6e21..77bbec4655 100644
--- a/libgnucash/app-utils/gnc-ui-util.c
+++ b/libgnucash/app-utils/gnc-ui-util.c
@@ -2340,3 +2340,21 @@ gnc_filter_text_for_currency_commodity (const gnc_commodity *comm,
 
     return gnc_filter_text_for_currency_symbol (incoming_text, *symbol);
 }
+
+GRegex *
+gnc_regex_new (const char *pattern)
+{
+    g_return_val_if_fail (pattern, NULL);
+    return g_regex_new (pattern, G_REGEX_OPTIMIZE, 0, NULL);
+}
+
+bool
+gnc_regex_match (GRegex *regex, const char *str)
+{
+    g_return_val_if_fail (regex && str, false);
+    return g_regex_match (regex, str, 0, NULL);
+}
+
+void
+gnc_regex_destroy (GRegex *regex)
+{
+    g_return_if_fail (regex);
+    g_regex_unref (regex);
+}
diff --git a/libgnucash/app-utils/gnc-ui-util.h b/libgnucash/app-utils/gnc-ui-util.h
index 3ad7e6bdc9..bf40f79e4d 100644
--- a/libgnucash/app-utils/gnc-ui-util.h
+++ b/libgnucash/app-utils/gnc-ui-util.h
@@ -34,6 +34,7 @@
 
 #include <glib.h>
 #include <locale.h>
+#include <stdbool.h>
 
 #include "Account.h"
 #include "gncOwner.h"
@@ -456,6 +457,11 @@ gchar * gnc_filter_text_for_currency_commodity (const gnc_commodity *comm,
                                                 const gchar *incoming_text,
                                                 const gchar **symbol);
 
+
+GRegex* gnc_regex_new (const char *pattern);
+bool gnc_regex_match (GRegex *regex, const char *str);
+void gnc_regex_destroy (GRegex *regex);
+
 #ifdef __cplusplus
 }
 #endif
```

jralls:
> I do understand the logic of not wanting to implement new report features if getting rid of Guile is a goal. I mean if there is ever an actual effort to redo these reports in something else besides guile it's going to be a serious undertaking (not that the rest of the code cleanup wasn't). I have no idea how likely this is to happen in the short-term. If this ends up taking years before it happens, and in the meantime we are limiting ourselves to what new features we can implement because Guile is on its way out at some point in the future then that might be quite limiting. As much as we can develop new register features, the reporting ultimately is a big part of what makes an accounting package great.

@dawansv The short answer for re-doing the reports in something other than Guile is "not anytime soon". There's a massive amount of other work that needs to happen first, and as you observe the rewrite will itself be a pretty substantial job. So no, that's not an obstacle to adding features or new standard reports. Guile itself can be something of an obstacle, but that's a separate issue entirely.

My objective is to get a feature that normal people can use and that makes sense to them. Normal people don't understand regular expressions. Heck, there are lots of *programmers* who don't understand regular expressions. Statements like

>  I am assuming at this point people using this know regular expressions or will go look for more information.

and (from @christopherlam )

>  IMV it's the users responsibility to find a good searchstring.

strongly suggests that it's not for normal people.

> At the end however these conversations are starting to take more time than writing the code itself

That's not unusual for a new feature, and the earlier in the process the discussing starts the better. Working out what the feature will do, how it will be used, how it fits into GnuCash's objectives, and how to present it to users before beginning the design saves a lot of time and frustration later; writing a new feature and dumping a PR without any previous discussion means all of those things have to be hashed out with code already on the table, and that inevitably leads to frustration and disappointment.

One of the biggest sources of the enormous pile of technical debt in GnuCash is that earlier versions of the core team were too willing to accept drive-by donations of features that were at best half-baked. Budgets are a sterling example but there are lots of others.

> so this probably has to reach its final step.

That's unfortunate.

dawansv:
> > so this probably has to reach its final step.
> 
> That's unfortunate.

I didn't mean this out of frustration. I meant these last little points are not that critical so it's OK for core developers to eventually make an "executive' decision on this. I do actually enjoy these conversations but I trying to be mindful not to take too much of your and Christopher's time. It's not like there aren't 100 other things to fix. Even more so with reports where there is clearly somewhere a line between core report and custom report. Not every custom feature should be in core.

The regular expression thing is already an added piece and not really central to that feature anyway. Even more so with my word boundary thing, it's really not important. I was just doing it because why not. I also see @christopherlam is giving some good feedback about implementing PCRE. I'll certainly look at it as a learning opportunity, but I actually agree with you @jralls that the regex stuff is already a bit out there in terms of regular users, so I think as much as it's fun playing with it, maybe we leave it as it is currently using the Guile implementation. I'll remove the little hack on word boundaries (not really that important) and it will match what we have with the transaction filter in this new feature. Now regex might not work in some locale but until we hear back from users really experiencing the problem, I don't think we should remove or refactor the current regex support either. I don't see it as a priority either way.

The basic non regex feature will work without regex. Well at least I can make work. This latest code actually uses the regex internally when doing just the basic prefix match as well, I thought that was clever (it saves a few lines of code) but now I realize that if regex fails in other locale then my basic feature will also fail when it really doesn't need to. Hence I will update that part of the code to be done without regex as I had before.

dawansv:
> If you need PCRE regex how about this for a start. Haven't checked for memory leaks. Even better would be to use std::regex however I haven't tried.

Thanks @christopherlam I'll have a look. After writing my previous message, I got into thinking what happens in a right-to-left language context with even just basic prefix match? The prefix is now to the right. That in itself could be a justification to implement PCRE support if it offers more reliable support even for just basic prefix matching. I need to do more reading about that.

Clumi:
@dawansv Is there any chance of your revisiting this PR? I am very interested in your work both in #1623 and here and would love to see it merged.

dawansv:
@Clumi Unfortunately I don't have the time nor probably the skills to make the regex feature fully compliant with international languages requirement. But I am maintaining the reports (keeping them updated with trep-engine changes) as custom reports here: https://github.com/dawansv/gnucash-custom-reports

PR #1678, opened 2023-06-20
Bug 798952 - Unable to set day threshold or counters in properties

dawansv:
Make sure counter values are retrieved as int; allow counters back to zero; fix voucher default value.

@jralls please double check this. I think this fixes some of the issues we are still having with the counters not being retrieved in the properties and at least brings it back to 5.1 level.

However it doesn't fix the following that was already a problem in 5.1: when setting the counters to zero or when setting the counter format to empty string, it doesn't save. Counters 1 and above will save, but not 0. As for format, any string will save but not setting it empty. Hence clicking on default and saving won't work because default = zero counters with no format strings. As I said this was already like that in 5.1, so it's not related to these latest changes that were implemented for this bug.

### DIFF
--- a/libgnucash/engine/gnc-optiondb.cpp
+++ b/libgnucash/engine/gnc-optiondb.cpp
@@ -547,7 +547,7 @@ GncOptionDB::load_from_kvp(QofBook* book) noexcept
                             set_double();
                             break;
                         case KvpValue::Type::INT64:
-                            option.set_value(kvp->get<int64_t>());
+                            option.set_value(static_cast<int>(kvp->get<int64_t>()));
                             break;
                         case KvpValue::Type::STRING:
                             fill_option_from_string_kvp(option, kvp);
@@ -889,7 +889,7 @@ gnc_register_counter_option(GncOptionDB* db, const char* section,
                             const char* doc_string, int value)
 {
     GncOption option{GncOptionRangeValue<int>{section, name, key, doc_string,
-                value, 1, 999999999, 1}};
+                value, 0, 999999999, 1}};
     option.set_alternate(true);
     db->register_option(section, std::move(option));
 }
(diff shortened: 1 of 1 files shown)


libgnucash/engine/gnc-optiondb.cpp
```diff
@@ -547,7 +547,7 @@ GncOptionDB::load_from_kvp(QofBook* book) noexcept
                             set_double();
                             break;
                         case KvpValue::Type::INT64:
-                            option.set_value(kvp->get<int64_t>());
+                            option.set_value(static_cast<int>(kvp->get<int64_t>()));
```
dawansv, line 550: I believe this is what caused the counters not to be populated after the first save, when the doubles have been converted to int, then saved back as int64_t. What is less clear is since this is generic kvp code, what else is stored as int64_t? If other things are stored as int64_t and need to be loaded as such (not as int) then I suppose we need to do here the same thing that you did for the double above, meaning testing if we are dealing with counters and if so retrieve as int, otherwise retrieve as int64_t.

jralls: @dawansv Indeed, good catch.

This code fortunately applies only to file properties. The only reason this would break is if there's a GncOptionValue<int64_t> in there and I'm pretty sure there isn't.

dawansv: Oh good yes now I see it's only called from `gnc_book_options_dialog_cb` via `gnc_option_db_load`, and also from `assistant_insert_book_options_page` but that's the assistant populating the file properties so going to the same place. 
I just tested some of the other properties in that dialog; there aren't many that are numbers, and they all seem to work fine, so as you say I don't think we have any other int64_t options in there.

jralls:
Thanks!

PR #1680, opened 2023-06-20
Register description fixes / changes

Bob-IT:
This PR has a couple of fixes / changes to the description completion.

The first deals with updating the cursor position when deleting selected text.

The second changes the match text when editing to be from the start of the description to the cursor allowing similar entries to be shown, originally it was the whole of the description.

The third resets the horizontal scrolling position on subsequent popups.

The last one changes the horizontal scroll position to show the matching text. So if you have a matches in the list at the start and at the end of a long description as you move down with the arrow keys it will scroll to the left and right appropriately. This also helps when editing long description text as the popup list should be in the same area.

I am in two minds about this, I think it does what was asked but need other people to test. The other option could be to allow the description popup to use all the space to the right like the transfer field by doing the following...

Change the line 129 in split-register-load.c to...

    gnc_completion_cell_set_autosize (cell, FALSE);

### DIFF
--- a/gnucash/register/register-gnome/completioncell-gnome.c
+++ b/gnucash/register/register-gnome/completioncell-gnome.c
@@ -666,6 +666,10 @@ gnc_completion_cell_modify_verify (BasicCell* bcell,
          box->stop_searching = FALSE;
     }
 
+    // Are were deleting or inserting in the middle.
+    if (change == NULL || *cursor_position < bcell->value_chars)
+        *start_selection = *end_selection = *cursor_position;
+
     populate_list_store (cell, newval);
 
     if (g_strcmp0 (newval, "") == 0)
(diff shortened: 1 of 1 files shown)


jralls:
@Bob-IT Not quite right, at least on macOS. As long as there are entries where the description is visible on the screen it doesn't scroll horizontally. That part is probably right. But the end of the match for all the matching rows goes off the screen it zooms so that all but half a letter of the match is visible:
image
image

@nomis asked for the [cursor to be put in the middle in that case](https://bugs.gnucash.org/show_bug.cgi?id=798960#c14), not on the left edge. Can you do that?

Bob-IT:
This update may be better.

dawansv:
This last version seems fine to me, at least for what I am testing. There are so many variations it's not easy to cover them all.

jralls:
> This update may be better.

Much. Thanks!

PR #1681, opened 2023-06-22
Bug 798956 - On Flatpak, broken link to Chart.bundle.min.js in chart report html

dawansv:
Just sharing my work so far for some feedback because I am not sure whether this even makes sense.

If we want to embed Chart.bundle.min.js in the html for reports using it, we have two choices. We either do it in the regular report rendering into the tmp html produced by simply viewing the report OR we only do it when exporting. The first option is much easier since right now the option to export to html isn't actually rendering anything, it's reusing the existing tmp file. 

Yet for the sake of it I decided to give the second option a try. Right now just testing with one report category-barchart.scm. 

It's easy enoough to use the existing export feature (in that report used for CSV export) and add a second HTML export which calls the embedding js code instead of the regular renderer just calling include js. 

I have to do a little step dance in `gnc-plugin-page-report.cpp` in the procedure that shows the available options since the default HTML export is added in that procedure to whatever other custom export types the report is defining, and treated separately from regular exports; that means we need to bypass the default HTML export when an actual custom HTML export is offered otherwise we would end up with 2 HTML export options (one just reusing the existing html file, the other rerendering with the embedded js). **TODO**: now I realize my code doesn't take into account the case where the custom HTML rendering is the only option and therefore chosen by default without the dialog box even showing up, so that's still a TODO.

The alternative is we simply always call `gnc:html-js-embed` instead of `gnc:html-js-include` and simply create larger tmp files that will be the way we want them if export to html is used. I really don't know the downside of this except for bigger tmp files, but how much does that matter? Yet it feels ugly. Bit it cuts down on a lot of the gymnastics involved with removing the default HTML option.

Finally handling multi-column reports will create its own problems and I have not adressed that here. We could simply call the custom html export for each subreport offering it, but then we will end up with the same js embedded for each report in the multi-column report, creating an unnecessary large file. It might have to be the price to pay for now.

### DIFF
--- a/gnucash/gnome/gnc-plugin-page-report.cpp
+++ b/gnucash/gnome/gnc-plugin-page-report.cpp
@@ -1532,14 +1532,27 @@ gnc_get_export_type_choice (SCM export_types, GtkWindow *parent)
 
     if (!bad)
     {
+        const gchar * html_type = _("HTML");
+        gboolean custom_html_export = FALSE;
+
         choices = g_list_reverse (choices);
 
-        choices = g_list_prepend (choices, g_strdup (_("HTML")));
+        for (node = choices; node; node = node->next)
+            if (g_strcmp0 (static_cast<char*>(node->data), html_type) == 0)
+            {
+                custom_html_export = TRUE;
+                break;
+            }
+
+        if (!custom_html_export)
+            choices = g_list_prepend (choices, g_strdup (html_type));
 
         choice = gnc_choose_radio_option_dialog
             (GTK_WIDGET (parent), _("Choose export format"),
              _("Choose the export format for this report:"),
              nullptr, 0, choices);
+
+        if (custom_html_export && (choice >= 0)) ++choice;
     }
     else
         choice = -1;
--- a/gnucash/report/html-chart.scm
+++ b/gnucash/report/html-chart.scm
@@ -65,6 +65,8 @@
 (export gnc:html-chart-set-grid?!)
 (export gnc:html-chart-set-y-axis-label!)
 (export gnc:html-chart-add-data-series!)
+(export gnc:html-chart-embed-js?)
+(export gnc:html-chart-set-embed-js?!)
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;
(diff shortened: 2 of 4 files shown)


jralls:
My first reaction is that it makes more sense to always embed the js instead of re-rendering the report on export. It's way cleaner and avoids trying to find all of the corner cases (like custom reports). The file size isn't really a problem. Yes, it adds 226K to files that can be as small as 8K and that will pretty quickly accumulate to tens or hundreds of MB in users' `tmp` directories if they don't reboot often,  but nowadays that's still not much considering that a single 1/2 hour video file is over 1GB.

dawansv:
> My first reaction is that it makes more sense to always embed the js instead of re-rendering the report on export.

Playing the devil's advocate, would we then offer the option to toggle between embed and include as a global option in Preference - Reports, most likely with embed as the default?

I am asking because there is an alternative include that could be useful to some users even if the link is broken. When writing the include code, it is possible to instruct the browser to look for Chart.bundle.min.js into the current directory (where the html file is located) as an alternative if it is not loaded from the original path; all it takes is adding this script tag after the one linking to /[whatever path]/Chart.bundle.min.js"

`<script>window.Chart || document.write('<script src="Chart.bundle.min.js"><\/script>');</script>`

Essentially this checks whether window.Chart exists (it should if Chart.bundle.min.js was loaded) and if not it adds the script to try to load Chart.bundle.min.js from the current directory. What this allows is for advanced users to still only use the include but fix the broken link problem (if it occurs) by simply retrieving Chart.bundle.min.js from wherever and putting it in the same directory as their report html files. The nice thing if this is included is that you can 'fix' a broken link after the fact, when you notice your charts are not loading, by simply putting Chart.bundle.min.js in the current directory without having to change the path in each html file (helpful if you have a lot of files).

jralls:
I think that's over-thinking it, and it would be brittle because moving the exported report without moving or copying Chart.bundle.min.js will break it.

Here's an option: [this cdn](https://www.jsdelivr.com/package/npm/chart.js?path=dist). You could link that in the exported file. We don't want the ancient WebKit we bundle with Windows to go out in the cold, cruel internet so the temp reports should still link the installed one.

dawansv:
> You could link that in the exported file

Then I propose we simply add this cdn as alternate location to the existing include in the default html rendering (used for tmp and default html export). This will cover both using the local file first (which should happen 100% of the time when viewing the tmp file within GC itself) or if local is not found when viewing in an external browser then this cdn.

See #1685

dawansv:
Closing this PR as we have move to a different solution in PR #1696

PR #1683, opened 2023-06-23
[gnc-main-window] Fix null pointer dereference when opening existing pages

nomis:
When the page already exists in a window then we can't be restoring pages so stop checking. The "window" variable will be nullptr if we're trying to open the page "in any window".

This bug was introduced in the changes for bug 798564.


jralls:
Thanks!

PR #1684, opened 2023-06-24
Fixes based on clang static analysis

nomis:
I wanted to see if clang's static analysis could find the recent null pointer dereference bug... it couldn't but it found several other issues:

Bugs:
* [Fix return value of gnc_book_write_to_xml_file_v2()](https://github.com/Gnucash/gnucash/commit/4b83068c6b64d41e82cee1c48caa5a73e94fe8f9)
* [Check source parameter to GncQuotesImpl::report() is not null](https://github.com/Gnucash/gnucash/commit/fe526a6043459763dba2701ff74fdb7717e193e4)
* [Initialise toclear_value in gnc_autoclear_window_ok_cb()](https://github.com/Gnucash/gnucash/commit/b8591de88e5595c0f72a995a532ad86cac81bee9) (not sure how to reproduce this, the impact looks like an invalid value could get set to an uninitialised `gnc_numeric` (if it was valid) instead of being left unmodified when providing an error message)
* [Fix memory leak in gnc_main_window_update_menu_and_toolbar()](https://github.com/Gnucash/gnucash/commit/2b47fe48f474fb76e9c761b86a1a88c3bd773e66)
* [Fix null pointer dereference in gnc_split_register_save_to_copy_buffer()](https://github.com/Gnucash/gnucash/commit/89ea0ca2398b7df8a0de9ffb43a2095c2d95de12) (copied a modified split but couldn't reproduce this - changing the account name on a newly created account affects the `account` cell not the `transfer` cell?)

Bug in tests:
* [Avoid unnecessary memory allocation in dom_tree_to_gnc_numeric()](https://github.com/Gnucash/gnucash/commit/d7568b2dcaafda3b5db092d4d78a99d2246d294a) (fixes use after free by avoiding allocation)

Won't cause a problem today but passes an uninitialised value that may affect future code:
* [Fix use of uninitialised value in gnc_split_register_load()](https://github.com/Gnucash/gnucash/commit/e1448d796b15b7a80b32d0b2c7a58cb9d24ff41c)

Should never happen but inconsistent null checks/missing initialisation of values:
* [Add missing null pointer check in gnc_account_renumber_update_examples()](https://github.com/Gnucash/gnucash/commit/96dd88c79ede5cd84a1976ff3e2db91925097c9d)
* [Initialise all column widths in set_dimensions_pass_two()](https://github.com/Gnucash/gnucash/commit/5b860d2fa9dfb788a86e23b4666a5a18007f3e95)

Warning in tests:
* [Avoid the potential negative left shift in test_xaccAccountType_Compa…](https://github.com/Gnucash/gnucash/commit/d36557ebbecabfbf6d011458aea104dc78820c6f)

Unresolved:
* [Bug 798966](https://bugs.gnucash.org/show_bug.cgi?id=798966) - Uninitialised variable used in [dialog-account-picker.c:dialog_response_cb()](https://github.com/Gnucash/gnucash/blob/ba7b26066ee8313b4580604958ce284950ac06be/gnucash/import-export/qif-imp/dialog-account-picker.c#L363-L372)

### DIFF
--- a/gnucash/gnome-utils/gnc-main-window.cpp
+++ b/gnucash/gnome-utils/gnc-main-window.cpp
@@ -3751,7 +3751,6 @@ gnc_main_window_update_menu_and_toolbar (GncMainWindow *window,
     const gchar *menu_qualifier;
 
     GMenuModel *menu_model_part;
-    GncMenuModelSearch *gsm = g_new0 (GncMenuModelSearch, 1);
 #ifdef MAC_INTEGRATION
     auto theApp{static_cast<GtkosxApplication *>(g_object_new(GTKOSX_TYPE_APPLICATION, nullptr))};
 #endif
@@ -3792,6 +3791,7 @@ gnc_main_window_update_menu_and_toolbar (GncMainWindow *window,
     gnc_menubar_model_remove_items_with_attrib (priv->menubar_model,
                                                 GNC_MENU_ATTRIBUTE_TEMPORARY);
 
+    GncMenuModelSearch *gsm = g_new0 (GncMenuModelSearch, 1);
     for (gint i = 0; ui_updates[i]; i++)
     {
         gchar *menu_name;
--- a/libgnucash/backend/xml/gnc-bill-term-xml-v2.cpp
+++ b/libgnucash/backend/xml/gnc-bill-term-xml-v2.cpp
@@ -151,11 +151,7 @@ static gboolean
 set_numeric (xmlNodePtr node, GncBillTerm* term,
              void (*func) (GncBillTerm*, gnc_numeric))
 {
-    gnc_numeric* num = dom_tree_to_gnc_numeric (node);
-    g_return_val_if_fail (num, FALSE);
-
-    func (term, *num);
-    g_free (num);
+    func (term, dom_tree_to_gnc_numeric (node));
     return TRUE;
 }
 
--- a/gnucash/gnome-utils/dialog-account.c
+++ b/gnucash/gnome-utils/dialog-account.c
@@ -2232,7 +2232,7 @@ gnc_account_renumber_update_examples (RenumberDialog *data)
     gtk_label_set_text (GTK_LABEL(data->example1), str);
     g_free (str);
 
-    if (strlen (prefix))
+    if (prefix && *prefix)
         str = g_strdup_printf ("%s-%0*d", prefix, num_digits,
                                interval * data->num_children);
     else
--- a/gnucash/gnome/window-autoclear.c
+++ b/gnucash/gnome/window-autoclear.c
@@ -123,7 +123,7 @@ gnc_autoclear_window_ok_cb (GtkWidget *widget,
                             AutoClearWindow *data)
 {
     GList *toclear_list = NULL;
-    gnc_numeric toclear_value;
+    gnc_numeric toclear_value = gnc_numeric_error (GNC_ERROR_ARG);
     gchar *errmsg = NULL;
     GError* error = NULL;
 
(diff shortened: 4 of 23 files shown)


jralls:
@nomis What's so hard about initializing `placeholder` in `dialog_response_cb`?

libgnucash/backend/xml/gnc-bill-term-xml-v2.cpp
```diff
@@ -151,12 +151,14 @@ static gboolean
 set_numeric (xmlNodePtr node, GncBillTerm* term,
              void (*func) (GncBillTerm*, gnc_numeric))
 {
-    gnc_numeric* num = dom_tree_to_gnc_numeric (node);
-    g_return_val_if_fail (num, FALSE);
+    gnc_numeric num;
+    gboolean ret;
```
jralls: This is C++: *Always* declare and initialize variables in the same line, and unless there's a scoping reason do so at first use. If you need a variable to be set in a scope and used outside of it declare and init just before the beginning of the scope where it will be set.

Also C++ has a `bool` type. Don't use `gboolean` unless interfacing with GLib or Gtk.

Do the same in new C code. We've removed the C89 restriction on mixing decls and code, and C99 has a `bool` type (you have to `#include <bool.h>` to use it).

jralls: As for changing `dom_tree_to_gnc_numeric` from returning a `gnc_numeric*` to returning a `bool` with an in/out reference parameter, *eww*. Just return a `gnc_numeric`.  [Copy elision](https://en.wikipedia.org/wiki/Copy_elision) is mandatory in C++17 so the numeric will get created in place in the calling function's stack.

nomis:
> @nomis What's so hard about initializing `placeholder` in `dialog_response_cb`?

Should it be initialised to `true` or `false`?

It's not entirely clear which would be most appropriate because it has an impact on whether some signal should be stopped or not.

gnucash/gnome-utils/gnc-main-window.cpp
```diff
@@ -3792,6 +3792,7 @@ gnc_main_window_update_menu_and_toolbar (GncMainWindow *window,
     gnc_menubar_model_remove_items_with_attrib (priv->menubar_model,
                                                 GNC_MENU_ATTRIBUTE_TEMPORARY);
 
+    gsm = g_new0 (GncMenuModelSearch, 1);
```
jralls: Move the decl here.

jralls (changes requested):

libgnucash/backend/xml/gnc-bill-term-xml-v2.cpp
nomis: What do you want it to do when `dom_tree_to_text` in `dom_tree_to_gnc_numeric` returns NULL?

jralls:
> Should it be initialised to true or false?

`true`. The signal is blocked if the account is a placeholder, meaning that one can't post splits to it. One also can't post splits to no account, so the signal should be blocked in that case too.

libgnucash/backend/xml/gnc-bill-term-xml-v2.cpp
jralls: The general modern C++ way is `std::optional` but that's not necessary here. You could return `GNC_ERROR_ARG`, but consider that `dom_tree_to_gnc_numeric` returns `gnc_numeric_zero` if `dom_tree_to_text` returns an empty string instead of nullptr. nullptr and "" are equivalent in this context so you might as well return 0 in both cases.

nomis: Done

jralls:
Thanks!

PR #1685, opened 2023-06-24
Bug 798956 Broken link to Chart.bundle.min.js - alternate solution

dawansv:
Add alternate CDN location to local Chart.bundle.min.js

See #1681 for earlier discussion about this

Tested on linux built, but please wait after 5.2 to commit this because it needs to be tested on Windows (unlikely to break but better safe than sorry)


dawansv:
Closing this PR as we have moved to a different solution in PR #1696

PR #1688, opened 2023-06-25
Html report chart field

christopherlam:
#1681 and #1685 differing view -- insert chartjs in `<head>` section, depending on whether the report template definition requires it.

Re: exporting html - currently it calls the exporter in gnc-html; I'm not sure how the html exporter works. It may be an idea to simply call the document renderer again with a new parameter `#:export #t` --- @dawansv

### DIFF
--- a/gnucash/report/html-chart.scm
+++ b/gnucash/report/html-chart.scm
@@ -454,8 +454,6 @@ document.getElementById(chartid).onclick = function(evt) {
          ;; clashing on multi-column reports
          (id (symbol->string (gensym "chart"))))
 
-    (push (gnc:html-js-include "chartjs/Chart.bundle.min.js"))
-
     ;; the following hidden h3 is used to query style and copy onto chartjs
     (push "<h3 style='display:none'></h3>")
     (push (format #f "<div style='width:~a;height:~a;'>\n"
--- a/gnucash/report/html-document.scm
+++ b/gnucash/report/html-document.scm
@@ -30,6 +30,7 @@
 (use-modules (gnucash report html-style-sheet))
 (use-modules (gnucash report html-table))
 (use-modules (gnucash report html-text))
+(use-modules (gnucash report html-utilities))
 (use-modules (gnucash report report-utilities))
 (use-modules (gnucash utilities))
 (use-modules (ice-9 match))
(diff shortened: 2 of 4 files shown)


dawansv:
> depending on whether the report template definition requires it.

Interesting take and if I am not mistaken putting the chart flag in the report template definition would allow us to control the non-insertion in the subreports, although it would create the challenge of the main multi-report code needing to know whether it has a chart report as a subreport. Is there a way to retrieve that from the render function for a given report?

> Re: exporting html - currently it calls the exporter in gnc-html; I'm not sure how the html exporter works.

If you look at this report definition for instance:
https://github.com/Gnucash/gnucash/blob/683c7f614d73977ad82e5a33928ce521488529ad/gnucash/report/reports/standard/category-barchart.scm#L696-L714

You'll see it has 2 renderers. The main "show on the screen" one at the bottom just called 'renderer and another one for export called 'export-thunk which passes a export-type (instead of #f for the regular renderer).

However the default html exporter is not an actual exporter. It doesn't use the renderer defined by 'export-thunk. Instead it piggybacks on the "show on screen" renderer -- it simply saves the html stream coming from the temp file when it's being displayed and reuses that stream when "export to html" is called. So there in no second rendering occurring there by default, with the advantage that it's very fast. 

Here is where the saving the html stream occurs (saved there for later use when "export to html" is called).
https://github.com/Gnucash/gnucash/blob/683c7f614d73977ad82e5a33928ce521488529ad/gnucash/html/gnc-html-webkit2.c#L519-L525

So if we want to embed the js, then we certainly don't want to rerender every html export, only the ones for reports with charts. But it implies we are able to have this info deep into the rendering code. 

Now looking at that code, just above it I find this:
https://github.com/Gnucash/gnucash/blob/683c7f614d73977ad82e5a33928ce521488529ad/gnucash/html/gnc-html-webkit2.c#L507-L517

So it looks like at one point we were using the OBJECT html tag for chart and essentially the hmtl stream was modified here to be different for export. This is where we would have to do it if we don't want to rerender but simply modify the html code after rendering.

https://github.com/Gnucash/gnucash/blob/683c7f614d73977ad82e5a33928ce521488529ad/gnucash/html/gnc-html-webkit2.c#L383-L457

The alternative is to define an actual html export type for reports that need it which is what I do in PR #1681 but it has the disavantage of, well, having to define an actual html export type in all reports using charts

jralls:
>  So there in no second rendering occurring there by default, with the advantage that it's very fast.

It would be even faster if it just copied the file.

christopherlam:
IIUC this `handle_embedded_object` mechanism is from the wizards of yore -- before chartjs there was jqplot, and before that there was guppi and gtkhtml. I don't think it's elegant to modify the html. I would think it's safer to simply re-run the HTML renderer, with an additional `export?` type flag to trigger packaging the chartjs within the html. But it should be elegant and streamlined -- i.e. not need too many changes to existing reports, and potentially upgradeable to embed libraries other than chartjs.

dawansv:
@christopherlam See #1696 for my continuation of this

jralls:
@christopherlam Should this close in favor of #1696?

christopherlam:
The tiny dilemma is whether to retain the `chart?` bool or to use a list of symbols to describe the js libraries.

christopherlam:
Maybe best keep it simple

PR #1689, opened 2023-06-26
Bug 798967 - Cannot Save to Any Path After Upgrading to 5.2

dawansv:
Fix inverted if statement condition for save success flag


christopherlam:
Sorry wrong place for comment but the release notes didn't mention that scrubbing is now mucho fast, and still works... In case users complain that scrubbing no longer seems to be working. Chris from mobile.

dawansv:
Working for me now on Windows as well with the latest nightly build

PR #1693, opened 2023-06-27
Use G_DECLARE_FINAL_TYPE for some gnome-utils classes

richardcohen:
These are the easiest ones.

Also, remove the class GncRecurrenceComp -- which is only used in the (never compiled) test-gnc-recurrence


jralls:
Thanks!

PR #1695, opened 2023-06-27
Fix memory leaks logging with gnc_numeric_to_string() 

nomis:
* Fix assumption in gnc_num_dbg_to_string()
* Fix memory leaks logging with gnc_numeric_to_string()


christopherlam:
Easier to use `gnc_num_dbg_to_string`

PR #1696, opened 2023-06-28
Html report chart field continued...

dawansv:
@christopherlam Taking it from your work at #1688

It works well with the chart.js now either included or embedded (if export flag is set) only once in the html head, including with the multi-column report

Of note: in gnucash/gnome/gnc-plugin-page-report.cpp I am no longer using the old `gnc_html_export_to_file` from the gnc_html object. The rendering in report-core.scm is already cached so there no need to go treat the regular and export rendering differently (except for the flag). In the export flag is not set, report-core.scm will simply reuse the previous already cached rendering (that is not new).

Still TO DO (if we decide to go this route):
+ html export also needs to be implemented in gnucash-cli, simlar to how it's done in gnc-plugin-page-report.cpp

DONE
+ ~~`gnucash/report/reports/standard/dashboard.scm` uses a weird definition not calling gnc:define-report so I have no idea how to set the chart? flag for that report~~ This was never a problem since dashboard is just a custom multicolumn
+ ~~The 2 chart tests `gnucash/report/test/test-html-chart.scm` and `gnucash/report/reports/standard/test/test-cashflow-barchart.scm` fail. How do we modify the test engine to address the need for a chart? flag there as well?~~ This was caused by a small typo in cashflow-barchart.scm

### DIFF
--- a/gnucash/gnome/gnc-plugin-page-report.cpp
+++ b/gnucash/gnome/gnc-plugin-page-report.cpp
@@ -1724,7 +1724,6 @@ gnc_plugin_page_report_export_cb (GSimpleAction *simple,
     char * filepath;
     SCM export_types;
     SCM export_thunk;
-    gboolean result;
     SCM choice;
     GtkWindow *parent = GTK_WINDOW (gnc_plugin_page_get_window
                                     (GNC_PLUGIN_PAGE (report)));
@@ -1788,17 +1787,37 @@ gnc_plugin_page_report_export_cb (GSimpleAction *simple,
                                      "document object with export-string or "
                                      "export-error."));
         }
-        result = TRUE;
     }
     else
-        result = gnc_html_export_to_file (priv->html, filepath);
-
-    if (!result)
     {
-        const char *fmt = _("Could not open the file %s. "
-                            "The error is: %s");
-        gnc_error_dialog (parent, fmt, filepath ? filepath : "(null)",
-                          strerror (errno) ? strerror (errno) : "" );
+        SCM chart_report = scm_call_1 (scm_c_eval_string ("gnc:report-chart?"),
+                                       priv->cur_report);
+        SCM html_export = (scm_is_true (chart_report)) ? SCM_BOOL_T : SCM_BOOL_F;
+        SCM report_render_html = scm_c_eval_string ("gnc:report-render-html");
+        SCM apply_with_error_handling = scm_c_eval_string ("gnc:apply-with-error-handling");
+        SCM arg_list = scm_list_3 (priv->cur_report, SCM_BOOL_T, html_export);
+        SCM rendered = scm_call_2 (apply_with_error_handling, report_render_html, arg_list);
+        SCM document = SCM_CAR(rendered);
+        SCM err = SCM_CDR (rendered);
+
+        if (scm_is_false (document))
+        {
+            gchar *str = scm_to_utf8_string (err);
+            gnc_error_dialog (parent, "%s %s", _("Error rendering HTML:"), str);
+            g_free (str);
+        }
+        else
+        {
+            GError *err = nullptr;
+            gchar *exported = scm_to_utf8_string (document);
+            if (!g_file_set_contents (filepath, exported, -1, &err))
+                gnc_error_dialog (parent, "%s %s",
+                                  _("Error during export:"),
+                                  err->message);
+            g_free (exported);
+            if (err)
+                g_error_free (err);
+        }
     }
 
     g_free(filepath);
--- a/gnucash/report/html-chart.scm
+++ b/gnucash/report/html-chart.scm
@@ -454,8 +454,6 @@ document.getElementById(chartid).onclick = function(evt) {
          ;; clashing on multi-column reports
          (id (symbol->string (gensym "chart"))))
 
-    (push (gnc:html-js-include "chartjs/Chart.bundle.min.js"))
-
     ;; the following hidden h3 is used to query style and copy onto chartjs
     (push "<h3 style='display:none'></h3>")
     (push (format #f "<div style='width:~a;height:~a;'>\n"
--- a/gnucash/report/html-document.scm
+++ b/gnucash/report/html-document.scm
@@ -30,6 +30,7 @@
 (use-modules (gnucash report html-style-sheet))
 (use-modules (gnucash report html-table))
 (use-modules (gnucash report html-text))
+(use-modules (gnucash report html-utilities))
 (use-modules (gnucash report report-utilities))
 (use-modules (gnucash utilities))
 (use-modules (ice-9 match))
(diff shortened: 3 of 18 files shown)


gnucash/gnome/gnc-plugin-page-report.cpp
```diff
@@ -1793,17 +1792,37 @@ gnc_plugin_page_report_export_cb (GSimpleAction *simple,
...
+        SCM arg_list = scm_list_3 (priv->cur_report, SCM_BOOL_T, html_export);
+        SCM rendered = scm_call_2 (apply_with_error_handling, report_render_html, arg_list);
+        SCM document = SCM_CAR(rendered);
+        SCM err = SCM_CDR (rendered);
+
+        if (scm_is_false (document))
+        {
+            gnc_error_dialog (parent, "%s %s",
+                              _("Error rendering HTML:"),
+                              scm_to_utf8_string (err));
```
christopherlam: This scm_to_utf8_string must be g_freed after use.

gnucash/gnome/gnc-plugin-page-report.cpp
```diff
@@ -1793,17 +1792,37 @@ gnc_plugin_page_report_export_cb (GSimpleAction *simple,
...
     else
-        result = gnc_html_export_to_file (priv->html, filepath);
-
-    if (!result)
     {
-        const char *fmt = _("Could not open the file %s. "
-                            "The error is: %s");
-        gnc_error_dialog (parent, fmt, filepath ? filepath : "(null)",
-                          strerror (errno) ? strerror (errno) : "" );
+        SCM chart_report = scm_call_1 (scm_c_eval_string ("gnc:report-chart?"),
```
christopherlam, line 1793: IMHO it's not necessary to trigger re-rendering for charts only. What's the harm in always repeating the render? Maybe in the future we'd include other JS libraries eg "ai-library.js" etc.

gnucash/gnome/gnc-plugin-page-report.cpp
dawansv: Oopsy. Will fix.

gnucash/gnome/gnc-plugin-page-report.cpp
dawansv: > What's the harm in always repeating the render?

I imagine mostly the time it can take on some type of reports with large datasets? It's always a little bit concerning to take away a feature that exists because we don't know which users it's going to impact (if any). I can almost hear the complaint 'I have this mega-big report that already takes one minute to be generated and now when I try to export it's so slow!!!!".

For future maintainability, what if instead of testing the value of chart? here, I create a more generic helper function in report-core.scm, something like gnc:report-rerender-on-html-export? that can then be called here instead of gnc:report-chart? Right now the only test in that helper function would still be chart?, but then later if other features are added it will be possibly more obvious that this function has to be amended with whatever other tests are needed?

jralls: That's not easy to answer. I have a pretty large book--58MB uncompressed--and it can take almost a minute to create some reports. What I don't know is what exactly is taking the time, and since Guile's call stack is invisible to the profiler I can't find out.

christopherlam: > That's not easy to answer. I have a pretty large book--58MB uncompressed--and it can take almost a minute to create some reports. What I don't know is what exactly is taking the time, and since Guile's call stack is invisible to the profiler I can't find out.

Which reports? How about some good old fashioned `(display "checkpoint-A\n")`?

jralls: > Which reports? How about some good old fashioned (display "checkpoint-A\n")?

The Net Worth barchart used to pretty reliably take over a minute, but I just tested it in 5.3 and it took only a couple of seconds.

> How about some good old fashioned (display "checkpoint-A\n")?

IO is slow so inserting any IO statements will distort the timing.

christopherlam: > > Which reports? How about some good old fashioned (display "checkpoint-A\n")?
> 
> The Net Worth barchart used to pretty reliably take over a minute, but I just tested it in 5.3 and it took only a couple of seconds.

b9382d2cf22ae1549f144bd64b1fe9b7bfcaa3f5 is the likely culprit. Ditto #420.

gnucash/gnome/gnc-plugin-page-report.cpp
dawansv: Fixed

dawansv:
Fixed missing `g_free` and typo in `cashflow-barchart.scm` that caused test to fail.

While we still think about whether or not to always rerender, I'd like to work on gnucash-cli. However I'd like some feedback on the following.

Right now, gnucash-cli produces the html as per what we see in the interface. It also has an `export-type` option used mostly for csv or whatever else might be implemented within the reports. 

So we have 2 options in implementing the export flag there:

1. Either we simply always run the export flag on the regular report run
2. Or we add an html export-type that has to be specified for the export flag to be set, so that the regular run still produces the regular rendering shown in the interface while the `--export-type html` option produces an option identical to the Export - HTML option in the interface.

Personally, I think option 2 is better as it follows the same logic as the GUI.

gnucash/gnome/gnc-plugin-page-report.cpp
```diff
@@ -1788,17 +1787,37 @@ gnc_plugin_page_report_export_cb (GSimpleAction *simple,
...
+                                       priv->cur_report);
+        SCM html_export = (scm_is_true (chart_report)) ? SCM_BOOL_T : SCM_BOOL_F;
```
christopherlam, line 1795: lol, use `chart_report` directly...

christopherlam (commented):
I'm still thinking if it's clever to use a bool field member `chart?`, or a list of symbols `js-libraries` which may be #f or `'(chartjs)` which triggers the loading of relevant libraries. But this may be overengineering.

The main dilemma is whether to rerender the report on export (much easier) or use the cached .html file in /tmp. IMV a slow report is a bad report.

jralls:
> Personally, I think option 2 is better as it follows the same logic as the GUI.

But there's no GUI to display it so what good is it?

gnucash/gnome/gnc-plugin-page-report.cpp
dawansv: LOL indeed. Possibly all that SCM conversion gymnastics causing my brain to shut off...

jralls:
@dawansv Did this fall off of your stack?

dawansv:
@jralls Sorry just got busy with summer. Should be able to get back to it later this month...

jralls:
> @jralls Sorry just got busy with summer. Should be able to get back to it later this month...

@dawansv time flies when you're having fun. Just a week left until the 5.4 release!

luzpaz:
soft bump

jralls:
@dawansv I'm closing your PRs, you seem to have lost interest in GnuCash. If interest revives by all means rebase them on current code and reopen them.

PR #1699, opened 2023-07-01
Bug 798922 - No withdrawal/deposit fields in csv import dialog

dawansv:
**Set Amount column labels as per base account type and global preference for formal accounting labels.**
+ generic headings set to "Amount (sign-standard)" and "Amount (sign-reversed)"
+ for formal labels, headings set to "Account: +Debit -Credit" and "Amount: +Credit -Debit"
+ when base account is selected, headings are customized
+ e.g. for Bank account, headings are set to "Amount: +Deposit -Withdrawal" and "Amount: +Withdrawal -Deposit"
+ [See other base account labels in code](https://github.com/Gnucash/gnucash/pull/1699/commits/7e759d95422dd59f4eb1271f911bafc2a0995488#diff-cb38e23796dd773e0243e5f319ab581b3dadf7f6c85c9083adf7624dd6fe2b6a)

**Clarify instructions re:Amount columns on start page.**
image

**TODO**

+ There are still some messages with the old (negated) language that I need to change
+ As per @nomis  comment: When formal labels are selected, the main start page message should be simplified.
+ As per @nomis  comment: Whatever decision is made as to how account columns are presented, a similar logic would then apply to value and transfer amount columns

### DIFF
--- a/gnucash/gtkbuilder/assistant-csv-trans-import.glade
+++ b/gnucash/gtkbuilder/assistant-csv-trans-import.glade
@@ -43,12 +43,29 @@
         <property name="margin-end">12</property>
         <property name="label" translatable="yes">This assistant will help you import a delimited file containing a list of transactions. It supports both token separated files (such as comma separated or semi-colon separated) and fixed width data.
 
-For a successful import three columns have to be available in the import data:
-• a Date column
-• a Description column
-• a Deposit or Withdrawal column
+For a successful import at least three columns have to be available in the import data:
+• a Date column,
+• a Description column,
+• an Amount column.
 
 If there is no Account data available, a base account can be selected to which all data will be imported.
+
+After selecting a base account, the Amount column headings will guide you in selecting the correct label for you data column(s) depending on the number of amount columns in your data and signs used (positive or negative).
+
+For instance, when selecting a bank account as a base account, the column heading labelled "Amount: +Deposit -Withdrawal" could be used on a single column containing both positive deposit amounts and negative withdrawal amounts, or on a column with positive deposit amounts only. The same label could then be used on a second column with negative withdrawal amounts only. However, if that second column comes with POSITIVE withdrawal amounts, then the opposite column heading labelled "Amount: +Withdrawal -Deposit" would need be used for that column.
+
+When separate Account data is provided for each transaction, then the heading(s) to select for the Amount data depends on whether it follows this standard sign format or its opposite:
+• for bank and cash accounts, deposits are positive and withdrawals are negative;
+• for credit card accounts, charges are negative and payments/refunds are positive;
+• for mortgages and loans, debt proceeds are negative and repayments are positive;
+• for brokerage accounts, "shares bought" are positive and "shares sold" are negative;
+• for other account types, refer to the concept guide, chapter 2 (available in the help menu).
+
+If it follows that format, select the "Amount (sign-standard)" heading; otherwise select "Amount (sign-reversed)".
+
+Note that when "formal accounting labels" are selected in Preferences, the formal "Debit/Credit" labels are used in Amount column headings:
+• "Amount: +Debit -Credit" for the sign-standard label, accepting positive debit amounts and negative credit amounts,
+• "Amount: +Credit -Debit" for the sign-reversed label, accepting positive credit amounts and negative debit amounts.
 
 Apart from a choice of delimiter, there are several options to tweak the importer. For example a number of lines can be skipped at the start or the end of the data, as well as odd rows. Several date and number formats are supported. The file encoding can be defined.
 
--- a/gnucash/import-export/csv-imp/assistant-csv-trans-import.cpp
+++ b/gnucash/import-export/csv-imp/assistant-csv-trans-import.cpp
@@ -41,6 +41,7 @@
 #include "gnc-uri-utils.h"
 #include "gnc-ui-util.h"
 #include "dialog-utils.h"
+#include "gnc-prefs.h"
 
 #include "gnc-component-manager.h"
 
@@ -1530,9 +1531,23 @@ CsvImpTransAssist::preview_style_column (uint32_t col_num, GtkTreeModel* model)
 /* Helper to create a shared store for the header comboboxes in the preview treeview.
  * It holds the possible column types */
 static GtkTreeModel*
-make_column_header_model (bool multi_split)
+make_column_header_model (bool multi_split, const Account* acct=NULL)
 {
     auto combostore = gtk_list_store_new (2, G_TYPE_STRING, G_TYPE_INT);
+
+    /* Set labels for Amount columns as per global label preference and account type */
+    if (gnc_prefs_get_bool(GNC_PREFS_GROUP_GENERAL, GNC_PREF_ACCOUNTING_LABELS))
+    {
+        gnc_csv_col_type_strs[GncTransPropType::AMOUNT] = _("Amount: +Debit -Credit");
+        gnc_csv_col_type_strs[GncTransPropType::AMOUNT_NEG] = _("Amount: +Credit -Debit");
+    }
+    else
+    {
+        GNCAccountType acct_type = xaccAccountGetType (acct);
+        gnc_csv_col_type_strs[GncTransPropType::AMOUNT] = gnc_imp_amount_strs[acct_type];
+        gnc_csv_col_type_strs[GncTransPropType::AMOUNT_NEG] = gnc_imp_amtneg_strs[acct_type];
+    }
+
     for (auto col_type : gnc_csv_col_type_strs)
     {
         /* Only add column types that make sense in
@@ -1624,7 +1639,7 @@ void CsvImpTransAssist::preview_refresh_table ()
     }
 
     /* Reset column attributes as they are undefined after recreating the model */
-    auto combostore = make_column_header_model (tx_imp->multi_split());
+    auto combostore = make_column_header_model (tx_imp->multi_split(), tx_imp->base_account());
     for (uint32_t i = 0; i < ntcols; i++)
         preview_style_column (i, combostore);
 
--- a/gnucash/import-export/csv-imp/gnc-imp-props-tx.cpp
+++ b/gnucash/import-export/csv-imp/gnc-imp-props-tx.cpp
@@ -52,6 +52,46 @@ namespace bl = boost::locale;
 
 G_GNUC_UNUSED static QofLogModule log_module = GNC_MOD_IMPORT;
 
+/* This map contains a set of strings representing the AMOUNT column informal
+ * names for different account types. */
+std::map<GNCAccountType, const char*> gnc_imp_amount_strs = {
+    { ACCT_TYPE_NONE,       N_("Amount (sign-standard)") },
+    { ACCT_TYPE_BANK,       N_("Amount: +Deposit -Withdrawal") },
+    { ACCT_TYPE_CASH,       N_("Amount: +Receive -Spend") },
+    { ACCT_TYPE_CREDIT,     N_("Amount: +Payment -Charge") },
+    { ACCT_TYPE_ASSET,      N_("Amount: +Increase -Decrease") },
+    { ACCT_TYPE_LIABILITY,  N_("Amount: +Decrease -Increase") },
+    { ACCT_TYPE_STOCK,      N_("Amount: +Buy -Sell") },
+    { ACCT_TYPE_MUTUAL,     N_("Amount: +Buy -Sell") },
+    { ACCT_TYPE_CURRENCY,   N_("Amount: +Buy -Sell") },
+    { ACCT_TYPE_INCOME,     N_("Amount: +Charge -Income") },
+    { ACCT_TYPE_EXPENSE,    N_("Amount: +Expense -Rebate") },
+    { ACCT_TYPE_PAYABLE,    N_("Amount: +Payment -Bill") },
+    { ACCT_TYPE_RECEIVABLE, N_("Amount: +Invoice -Payment") },
+    { ACCT_TYPE_TRADING,    N_("Amount: +Decrease -Increase") },
+    { ACCT_TYPE_EQUITY,     N_("Amount: +Decrease -Increase") },
+};
+
+/* This map contains a set of strings representing the AMOUNT_NEG column informal
+ * names for different account types. */
+std::map<GNCAccountType, const char*> gnc_imp_amtneg_strs = {
+    { ACCT_TYPE_NONE,       N_("Amount (sign-reversed)") },
+    { ACCT_TYPE_BANK,       N_("Amount: +Withdrawal -Deposit") },
+    { ACCT_TYPE_CASH,       N_("Amount: +Spend -Receive") },
+    { ACCT_TYPE_CREDIT,     N_("Amount: +Charge -Payment") },
+    { ACCT_TYPE_ASSET,      N_("Amount: +Decrease -Increase") },
+    { ACCT_TYPE_LIABILITY,  N_("Amount: +Increase -Decrease") },
+    { ACCT_TYPE_STOCK,      N_("Amount: +Sell -Buy") },
+    { ACCT_TYPE_MUTUAL,     N_("Amount: +Sell -Buy") },
+    { ACCT_TYPE_CURRENCY,   N_("Amount: +Sell -Buy") },
+    { ACCT_TYPE_INCOME,     N_("Amount: +Income -Charge") },
+    { ACCT_TYPE_EXPENSE,    N_("Amount: +Rebate -Expense") },
+    { ACCT_TYPE_PAYABLE,    N_("Amount: +Bill -Payment") },
+    { ACCT_TYPE_RECEIVABLE, N_("Amount: +Payment -Invoice") },
+    { ACCT_TYPE_TRADING,    N_("Amount: +Increase -Decrease") },
+    { ACCT_TYPE_EQUITY,     N_("Amount: +Increase -Decrease") },
+};
+
 /* This map contains a set of strings representing the different column types. */
 std::map<GncTransPropType, const char*> gnc_csv_col_type_strs = {
         { GncTransPropType::NONE, N_("None") },
@@ -64,10 +104,10 @@ std::map<GncTransPropType, const char*> gnc_csv_col_type_strs = {
         { GncTransPropType::VOID_REASON, N_("Void Reason") },
         { GncTransPropType::ACTION, N_("Action") },
         { GncTransPropType::ACCOUNT, N_("Account") },
-        { GncTransPropType::AMOUNT, N_("Amount") },
-        { GncTransPropType::AMOUNT_NEG, N_("Amount (Negated)") },
+        { GncTransPropType::AMOUNT,  gnc_imp_amount_strs[ACCT_TYPE_NONE] },
+        { GncTransPropType::AMOUNT_NEG, gnc_imp_amtneg_strs[ACCT_TYPE_NONE] },
         { GncTransPropType::VALUE, N_("Value") },
-        { GncTransPropType::VALUE_NEG, N_("Value (Negated)") },
+        { GncTransPropType::VALUE_NEG, N_("Value (sign-reversed)") },
         { GncTransPropType::PRICE, N_("Price") },
         { GncTransPropType::MEMO, N_("Memo") },
         { GncTransPropType::REC_STATE, N_("Reconciled") },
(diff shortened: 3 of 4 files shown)


flywire:
> Moving the conversation to this PR for next steps.

The PR is moot while the import gives different results for users as per the [bug report](https://bugs.gnucash.org/show_bug.cgi?id=798964).

---

> e.g. for Bank account, headings are set to "Amount: +Deposit -Withdrawal" and "Amount: +Withdrawal -Deposit"

Why? The `+/-` prefix makes no sense. Existing account type Debit/Credit labels or alias (as set in the preferences) should be used:

```
* BALANCE ACCCOUNT TYPE       DEBIT    CREDIT    
  debit            NONE       Funds In Funds Out
  debit            BANK       Deposit  Withdrawal
  debit            CASH       Receive  Spend    
  credit           CREDIT     Payment  Charge    
* Debit            ASSET      Increase Decrease  
* Credit           LIABILITY  Decrease Increase  
  debit            STOCK      Buy      Sell      
  debit            MUTUAL     Buy      Sell      
  debit            CURRENCY   Buy      Sell      
* Credit           INCOME     Charge   Income    
* Debit            EXPENSE    Expense  Rebate    
* Credit           PAYABLE    Payment  Bill      
* Debit            RECEIVABLE Invoice  Payment  
  credit           TRADING    Decrease Increase  
* Credit           EQUITY     Decrease Increase  
```

I expect the account is the first data line in a multi-split file with other accounts being transfer accounts. Otherwise, I suppose account type `NONE` should apply.

nomis:
This is becoming very wordy; the text should change when formal accounting labels are enabled to avoid overloading the user with information that doesn't apply to them. I think it now needs headings because it's going to be hard to read.

I'd get rid of "Amount (sign-standard)" and "Amount (sign-reversed)" entirely and replace them with "Debit" and "Credit" when importing for multiple accounts. I think there's space on the page where columns are selected to have a help area with a combo box for account type and then context-specific information for each of them to explain how to select "Debit Amount" or "Credit Amount"?

Everything probably needs to be duplicated for "Value" and "Transfer Amount", or those need to use "Debit" and "Credit" in the names because even "sign-reversed" needs some explanation.

> Why? The `+/-` prefix makes no sense.

There needs to be a column name that works when the amount is only in one column that will have both positive and negative values. I think "Amount: +Deposit -Withdrawal" and "Amount: +Withdrawal -Deposit" make a lot of sense. The complication is that there are a lot of informal label options and three fields (Amount/Value/Transfer Amount) that they apply to.

There would be 66 options if all of them are visible at once. The total number of options would be 85. However, that may be easier to use than trying to explain what to select for different accounts.

dawansv:
@nomis 
> This is becoming very wordy; the text should change when formal accounting labels are enabled to avoid overloading the user with information that doesn't apply to them. I think it now needs headings because it's going to be hard to read.

No question there. Some ideas to improve it:
- As you say, easy to get rid of a lot of this when formal labels are selected. That certainly should be done so will add to TODO.
- For informal, we could skip the whole section about not selecting a base account, either sending people to a wiki page where it is further explained and essentially saying (in a nice way) "until you understand how this works it's probably best to first try importing one account at a time and let the column labels guide you". Not ideal but there is some truth in it and there is a limit to how much we can explain here.

> I'd get rid of "Amount (sign-standard)" and "Amount (sign-reversed)" entirely and replace them with "Debit" and "Credit" when importing for multiple accounts.

Definitely something that crossed my mind. As much as "Amount (negated)" is confusing, the reality is the alternative "Amount (sign-reversed)" is hardly less confusing without some explanation. Since it's confusing anyway, then going with Debit/Credit at least as the benefit of being standard even if it still requires an explanation. Note that IMO the column headings would still have to be "Amount: +Debit -Credit" and "Amount: +Credit -Debit", as they are with formal labels, to accommodate the single column option. Simply using Debit/Credit only makes sense in a "dual column with positive numbers only" settings, not single columns or negative numbers thrown in.

> Everything probably needs to be duplicated for "Value" and "Transfer Amount", or those need to use "Debit" and "Credit" in the names because even "sign-reversed" needs some explanation.

Yes I should have mentioned that whatever we decide to do for amounts has to be implemented there as well. But let's come with a solution fir the amount columns and then we can follow the same logic with the rest. But will add to TODO list as well to make this clear.

@flywire 
> Why? The +/- prefix makes no sense.

To add to what @nomis is saying. We **could** differentiate between single and dual column imports with a checkbox, and change the labels accordingly. But that's yet ANOTHER option. And it would not cover the case of people using dual columns YET still with negative amounts in one column. I have seen some banks providing CSV in that way, with deposit and withdrawsls in 2 columns, but the withdrawals are still negative numbers. So then that would be to be yet another option. So I am not saying what I am proposing is pretty, not am I attached to it in any wat, but the challenge is to come up with an alternative that covers all of these scenarios, not just the "dual column with positive debit/credit numbers only" scenario.

gjanssens:
I think this is a solution that strikes a nice balance.

While not actively involved in the discussion, I was pondering how to improve on this. My main worry was that if we wanted to support all the possible names we use for amount in the different account types, this would result in a very long list of column types to choose from. But by specializing the "Amount" column type only when a base account is selected, this issue is avoided entirely.

I'm also happy you don't change the number of possible types that are used in the presets. Internally the code continues to call this column "AMOUNT" or "AMOUNT_NEG", regardless of the displayed name.

So a :+1:  from me!

flywire:
> * "until you understand how this works it's probably best to first try importing one account at a time and let the column labels guide you"

This discussion offers very complicated ways of selecting two columns (ie Debit and Credit) where the worst outcome of not understanding the guidance is the need to rerun the process of selecting the columns in reverse order.

1. emphasise the need to try an import with one account
1. it would be unreasonable not to support the labels used for account types elsewhere in the software
1. add Amount as a Debit alias for a single column
1. again, `+/-` is confusing for no benefit

> I have seen some banks providing CSV in that way, with deposit and withdrawals in 2 columns, but the withdrawals are still negative numbers.

This is the most important outlier not covered by two columns and the ability to reverse them. It requires a special case, possibly just a note for the user to fix it before importing it.

> There would be 66 options if all of them are visible at once. The total number of options would be 85

This is completely wrong, there are still only a Debit and Credit with all other terms being an alias used in a lookup table displayed in column headings or help text according to the account type. It has no practical affect on programming.

gjanssens:
> > * "until you understand how this works it's probably best to first try importing one account at a time and let the column labels guide you"
> 
> This discussion offers very complicated ways of selecting two columns (ie Debit and Credit) where the worst outcome of not understanding the guidance is the need to rerun the process of selecting the columns in reverse order.

Do you consider this good or bad (I mean that worst case you have to rerun the process once to select columns in reverse order) ?
I assume that once you are satisfied with your selections you save the settings for your next import. So the burden to understand the exact order should happen only very occasionally.

> 
>     1. emphasise the need to try an import with one account
> 
>     2. it would be unreasonable not to support the labels used for account types elsewhere in the software

The main problem here is that when the preview page opens, there is insufficient data available to determine which labels would be preferred for this particular import. Is it a bank import ? Is it a credit card import ? Or something else ?

So you either have to list all possible labels in the column dropdown menus, or you present a generic label until you have more details to determine which ones are most appropriate. @nomis did the calculation - if you want to add all possible labels to the dropdown menu, you'd end up with 66 to 85 elements in that list only to select an amount or value column using the most suitable labels. Personally I think that would seriously degrade the usability of these dropdown menus. Keep in mind these options will be presented for each column header you have to select. So these 66-85 options will be listed as well when you try to select "Description" for your description column.

In addition, the importer preview window is designed to give you feedback on missing settings (like if you have a multi-currency transaction, but not price or value column selected). That message can now be limited to mention price and value. If your dropdown has all labels supported elsewhere in the system, you may have to repeat all these labels also in these messages. Again I think they would become unusably long.

The alternative approach as suggested by @dawansv is to start with generic names and as soon as there is sufficient detail to know which labels are most suitable, replace them with the more suitable ones. That keeps the dropdown menus usable and clean.

To me this approach makes more sense.

> 
>     3. add Amount as a Debit alias for a single column
> 

I'm not sure what problem you see here that needs fixing ?

>     4. again, `+/-` is confusing for no benefit
> 

Depends on who you ask. For me Debit/Credit remains confusing even after 15 years of gnucash usage.

But frankly I'm not too opinionated on what are the best terms. I have worked a lot on the csv importer to allow it to import a wide range of transaction types from very differently formatted csv files. It's called *generic* after all. And I think there are only very few edge cases that are currently not possible to import. The only ones I'm currently aware of are:
* a file in which the amounts are all in a single column and all positive, with a separate column that indicates whether the amount is a Debit or Credit
* date formats that are not short dates (like 15-mar-2023).

I do understand this comes at the expense of simplicity. There are effectively many knobs to turn. I'm open to user interface improvements, but I think we'll have to accept some level of complexity, simply due to the generic nature of the importer.

> 
> > I have seen some banks providing CSV in that way, with deposit and withdrawals in 2 columns, but the withdrawals are still negative numbers.
> 
> This is the most important outlier not covered by two columns and the ability to reverse them. It requires a special case, possibly just a note for the user to fix it before importing it.

It is covered: in this case you select the same amount header  `Amount (sign-standard)` twice: once for the column with the deposit values and once for the column with the withdrawal values. GnuCash will internally combine them into one amount.

Perhaps this bit is not obvious: for amounts and values, you can use the same label more than once on different columns. So the label `Amount (sign-standard)` can be used in multiple scenarios:
* if there's only one column containing both positive and negative amounts
* in case there's a deposit column with positive amounts and a withdrawal column with negative amounts. In this case the same label is set on both columns
* in case there are both a deposit and a withdrawal column with positive amounts. In this case the label above is used for the deposit column and the label `Amount (sign-reversed)` is used for the withdrawal amount, because you want these amount to be of opposite sign than they are in the csv data.

Again I don't mind what is used as descriptive label. It should be something that matches the three use cases above.

> 
> > There would be 66 options if all of them are visible at once. The total number of options would be 85
> 
> This is completely wrong, there are still only a Debit and Credit with all other terms being an alias used in a lookup table displayed in column headings or help text according to the account type. It has no practical affect on programming.

There are several levels here.

The first is what the user sees in the GUI: as I mentioned before I don't think adding 66-85 menu options just to offer all possible labels in all possible column menus will give a poor user experience IMO.

@dawansv compromise is much more manageable at the expense of only adding clarity once there's sufficient detail to do so.

Internally, you could work with only debit and credit in all cases if that helps code clarity.

gjanssens:
Final note, wouldn't the importer benefit a lot from having some good examples of csv files and how to import them in the documentation ? I imagine this could alleviate the initial confusion and complexity users are seeing now while using the importer for the first time.

dawansv:
@flywire Feedback is appreciated but I would ask if you could clarify because I have trouble understanding

> * emphasise the need to try an import with one account

I think we are in agreement here? I think in the intro text, it's a good idea to invite users to start with single account import first because it's easier for the interface to guide them via the column labels this way.

> * it would be unreasonable not to support the labels used for account types elsewhere in the software

Yes, that's the whole point of the change I am proposing. I am using the informal labels as used elsewhere in the software, unless of course formal labels are selected, in which case I am using the Debit/Credit labels. But I am only able to implement these labels when a base account is selected, for obvious reasons since the labels are based on the account type. If you have a better way to do this, then please propose it?

> * add Amount as a Debit alias for a single column

I don't really understand what you mean by "alias" in this context. That said in my proposal Amount is always presented as part of the label, including the formal Debit/Credit labels, but I assume you are talking about something else. Possibly you can provide examples so I can understand. 

> * again, `+/-` is confusing for no benefit

The importer allows both columns with mixed signs (+ and -) or columns with single signs (+ or -). As @gjanssens said, the only option not offered is to try importing debit/credit with the same sign in the same column, and for obvious reasons as there is no way to differentiate them. Now this offers a lot of flexibility but is also more complex; often these 2 go together. I we want to get rid of that complexity, then we lose functionality. Or we can try to improve the language and explain things better which is what I am trying to do. If you can think of a better way to improve language or functionality, then why not present it?

If you are referring specifically to not liking the labels saying (for instance for bank) "Amount: +Deposit -Withdrawal" and "Amount: +Withdrawal -Deposit" then what is your alternative, taking into account that the labels have to convey the use for both single column and dual column. "Amount: +Deposit -Withdrawal" means "use this label for columns that have either positive numbers for deposit amounts or negative numbers for withdrawal amount or both". The other label covers the other scenario for instance where withdrawals are positive numbers. 

This I attempt to explain in my intro text as such:

`For instance, when selecting a bank account as a base account, the column heading labelled "Amount: +Deposit -Withdrawal" could be used on a single column containing both positive deposit amounts and negative withdrawal amounts, or on a column with positive deposit amounts only. The same label could then be used on a second column with negative withdrawal amounts only. However, if that second column comes with POSITIVE withdrawal amounts, then the opposite column heading labelled "Amount: +Withdrawal -Deposit" would need be used for that column.`

Could this be clearer? Well of course, things can always be improved, but that's the best I can come up with. But why not provide specific feedback in the form of what you think is better language then?

As @gjanssens explains, this also answers the issue below. That case is covered, and that's why I mention it in my intro text and also why the labels have to be as they are with the double +/- description.

> > I have seen some banks providing CSV in that way, with deposit and withdrawals in 2 columns, but the withdrawals are still negative numbers.

> 
> This is the most important outlier not covered by two columns and the ability to reverse them. It requires a special case, possibly just a note for the user to fix it before importing it.

That said if you have a better alternative then by all means describe it.

dawansv:
> > There would be 66 options if all of them are visible at once. The total number of options would be 85
> 
> This is completely wrong, there are still only a Debit and Credit with all other terms being an alias used in a lookup table displayed in column headings or help text according to the account type. It has no practical affect on programming.

What I am proposing is as you describe: there are just 2 different column types with all other terms being an alias used in a lookup table displayed in column headings the account type. As you say the different informal labels have no practical effect on programming. We could also look at using help text to provide more information since the column labels are obviously very restrictive in terms of space. Finally for obvious reasons, this can only be done when a base account is selected.

Assuming you don't like the format of the labels themselves with the +/-, what do you propose as an alternative considering that the importer allows both columns with mixed signs (+ and -) or columns with single signs (+ or -). 

Using single labels as an alternative is simpler but unfortunately just doesn't convey the full meaning of these columns. How is a column that says (for instance for bank) "Deposit" or "Deposit Amount" going to indicate that this is also the label to use when you have negative withdrawal amount either mixed in the same column or as a separate column? How would a user know that a column with negative withdrawal amounts should use the "Deposit" label and not the "Withdrawal" label?

flywire:
The CSV importer/exporter is a really good feature of GnuCash setting it apart from a lot of commercial software.

> Final note, wouldn't the importer benefit a lot from having some good examples of csv files and how to import them in the documentation ?

Agree. Proposed help is much too complicated, better to move most of it to the documentation and refer to it.

> Do you consider this good or bad (I mean that worst case you have to rerun the process once to select columns in reverse order) ?

Good. Highlight up front if the demonstration data doesn't give the expected output just reverse the fields.

> The main problem here is that when the preview page opens, there is insufficient data available to determine which labels would be preferred for this particular import.

As explained by @dawansv, "only able to implement these labels when a base account is selected" (ie "present a generic label until you have more details to determine which ones are most appropriate" so "66 to 85 elements" never occurs).

> I'm not sure what problem you see here (`+/-`) that needs fixing ?

> If you are referring specifically to not liking the labels

It's confusing. Just use plain terms (eg Debit / Credit) unless they are being reversed in which case the sign before the term is meaningful to highlight usage is reversed and explain it is reversed in the docs..

> select the same amount header twice

Nice. I've never had this data so I wasn't aware of this.

> I think we are in agreement here?

Yes. In general our differences are only in finer detail about other points.

> I don't really understand what you mean by "alias" in this context.

An alias is another term for something but both terms are the same item. As I see it there are only two columns, formally Debit and Credit, with any other terms or use (including amount) being an alias of one of these (although selecting the same header twice is an exception).

flywire:
Linking other importer enhancement requests:
1. [Enable direct GST Splits when Importing File](https://bugs.gnucash.org/show_bug.cgi?id=798987)
1. [Request Support CSV Fieldname Header](https://bugs.gnucash.org/show_bug.cgi?id=797700)
1. should use Gnucash Preferences for Date format

gjanssens:
Final note, wouldn't the importer benefit a lot from having some good examples of csv files and how to import them in the documentation ? I imagine this could alleviate the initial confusion and complexity users are seeing now while using the importer for the first time.

> Linking other importer enhancement requests:
> 
>     1. [Enable direct GST Splits when Importing File](https://bugs.gnucash.org/show_bug.cgi?id=798987)
> 
>     2. [Request Support CSV Fieldname Header](https://bugs.gnucash.org/show_bug.cgi?id=797700)
> 
>     3. should use Gnucash Preferences for Date format

While related to the CSV importer, these are not related to this particular PR. Why do you add these links here ?

flywire:
> I imagine this could alleviate the initial confusion and complexity users are seeing now while using the importer for the first time.

I would have expected so until I saw opposite results from two users importing the same file using the same process.  Hence the suggestion to `Highlight up front if the demonstration data doesn't give the expected output just reverse the fields.`

> Why do you add these links here ?

Reviewing the importer in the context of these comments likely offers an opportunity for these related issues. Implementing the second and third points would likely take a single line each.

gjanssens:
> > I imagine this could alleviate the initial confusion and complexity users are seeing now while using the importer for the first time.
> 
> I would have expected so until I saw opposite results from two users importing the same file using the same process. Hence the suggestion to `Highlight up front if the demonstration data doesn't give the expected output just reverse the fields.`

Oh, I misunderstood the original intent of your earlier comment.

IMO the opposite results are a bug, that should be fixed. From that point of view it should not be mentioned in the clarifying statement. We don't write gui prose to explain known bugs. That's a waste of time better spent on actually fixing the bug.

If after deeper analysis this should turn out not to be a bug, but desired behaviour that can be deterministically explained, *then* is could be made part of the initial clarifications. Until that's sorted out this is not relevant to this PR.

> 
> > Why do you add these links here ?
> 
> Reviewing the importer in the context of these comments likely offers an opportunity for these related issues. Implementing the second and third points would likely take a single line each.

Please don't do that. A PR is intended to review code at hand. You should not hijack it to divert attention to issues that serve your own interest, even if they may be in the same area of code. That is very distracting in this context.

If you do have good reason to believe these are "single line each" fixes you could add your analysis supporting this idea to the respective bug reports  or submit separate PRs with these respective single lines. Note I don't consider "These likely take a single line each" a good analysis. That would require something more tangible.

I understand your desire to get these things fixed as they clearly are hampering your use of the gnucash import features. But please use each support tool as it's intended.

flywire:
> Oh, I misunderstood the original intent of your earlier comment.

No you didn't, the bug example was intended as an ultimate example where documentation and understanding could never be successful and user should rely on reversing the fields.

> add your analysis supporting this idea to the respective bug reports

Python prototype code was added to two of three examples but the third issue does not occur in Python so GnuCash setting was identified.

ysalmon:
Sorry to intervene on this long --- and difficult to follow -- conversation, but I need to point out that despite having been a GnuCash user for a decade, the change in vocabulary is making me currently unable to import transactions from a CSV file as I cannot decide what setting to choose.

This is due to the change in vocabulary not being reflected in the documentation, and to the fact that the French translation is rather odd (back translated to English, the options are "Raw amount", "Amount (negativity)", "Amount" and "Value (negativity)"), which in turn is probably because the French translator too did not get the meaning of the options.

My situation is the following :
 - there is a column for all amounts of money going out of the account, with a ``-`` sign in front of each amount ;
 - there is another column for all amounts of money going into the account, with a ``+`` sign in front of each amount.

Obviously, each line has a value in one or the other of those columns but not both.

So what setting should I use ?


And more generally, I think this approach (listing different typologies of bank statements rather that trying to find a name for each column) would help achieving the clarity you are trying to have : before the column assignation step, or in the documentation, show the user typical layouts (with either one or two columns), let him choose one and tell him what settings he needs to set. In the example, use example lines with descriptions that make it obvious in which way money is flowing (eg. "Cash deposited on this account" vs. "Money withdrawn at the ATM"). The name of the setting then becomes of no importance.

The typologies I can think of are :
 - the one I described above,
 - the same but with signs reverted,
 - with two columns but all amounts are written in absolute values,
 - only one column, with the convention that positive is an inflow,
 - only one column, with the convention that positive is an outflow,
 - the same as all above, but with ``+`` signs being omitted.

flywire:
> So what setting should I use ?

> Highlight up front `if the demonstration data doesn't give the expected output just reverse the fields.`

markokecman:
i am a new user of gnucash and this import functionality is very frustrating. i understand the idea is to pander to the "amateurs" who do not understand credit and debit concepts, but the chosen path does nobody any favours.

absent a "pro" import functionality (where a credit and debit account is specified along with the single amount number) i suggest the following workaround:

- instead of "amount" and "amount (negated)" we have "balance increase" and "balance decrease"

this way, in a multi-split import (i don't even want to think about one-line imports for amateurs) it would be more obvious what is asked. if i owe more money on a credit card, this is a balance increase. if i get money into a bank account, this is also a balance increase.

this means that you can have two balance increases (or decreases) in a single transaction, rather than one "+" and one "-".

PR #1700, opened 2023-07-04
Change g_assert() -> g_assert_true() in tests

richardcohen:
g_assert() can be compiled out, so should not be used for tests

Minimum glib has been increased to 2.56 since
https://bugs.gnucash.org/show_bug.cgi?id=792008


jralls:
The commit message, repeated above, is not correct. The fix for [Bug 792008](https://bugs.gnucash.org/show_bug.cgi?id=792008) is what caused `g_assert_true`'s removal, necessary because it wasn't introduced until glib-2.38 and we required only 2.26 at that time. glib >= 2.40 has been required since https://github.com/Gnucash/gnucash/commit/8acbc41c61f63149a0583e4fe49dc133eb081d5f.

I've fixed the commit message & force-pushed.

Thanks!

PR #1703, opened 2023-07-07
[gnc-plugin-business.c] free owners from business plugin

christopherlam:
@richardcohen is problably the expert

### DIFF
--- a/gnucash/gnome/gnc-plugin-business.c
+++ b/gnucash/gnome/gnc-plugin-business.c
@@ -241,6 +241,11 @@ gnc_plugin_business_finalize (GObject *object)
 {
     g_return_if_fail (GNC_IS_PLUGIN_BUSINESS (object));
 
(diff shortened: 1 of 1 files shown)


richardcohen:
Hi @christopherlam,
I would suggest using GNC_PLUGIN_BUSINESS() instead of the raw cast, not least to be consistent with the rest of the file.

I am not familiar with the business features, so I could be wrong, but ... those member variables are initialised to NULL values, and then never changed!!

jralls:
@richardcohen, That's not really necessary because the preceding line
```
 g_return_if_fail (GNC_IS_PLUGIN_BUSINESS (object));

```
accomplishes the same thing: The GObject cast macros run the GType check and return `NULL` if it fails and cast the argument if it passes. A more GObject-idiomatic way would be
```
      GncPluginBusiness *plugin = GNC_PLUGIN_BUSINESS (object);
      g_return_if_fail (plugin);
```
But I don't think it's worth the effort to change on its own.

PR #1705, opened 2023-07-10
gnc:string<? uses glib instead of guile string compare

christopherlam:
for [Bug 798991](https://bugs.gnucash.org/show_bug.cgi?id=798991) - Incorrect Account Name Order in Transaction Report

needs win32 build and test on localised environment...


jralls:
@christopherlam As I noted on the bug I can't replicate the problem so I can't tell if this makes a difference. OTOH it doesn't seem to do any harm. Maybe we should just merge it so that zWind can see if it changes anything on their system.

PR #1586, opened 2023-03-19
import main matcher to cpp and use stl

christopherlam:
Edit (April 2023) note this branch is currently damaged due to too many fixups. I'd consider it abandoned.

PR #1708, opened 2023-07-15
import-matcher convert to cpp classes

christopherlam:


jralls:
@christopherlam Are you planning to actually create any classes or is this just about replacing GLib containers with STL ones?

christopherlam:
Creating classes would be the goal, but much slower than the previous attempt #1586 which became buggy somewhere...

jralls:
Maybe write some more tests first so that you can make sure the refactors don't break anything. Also the matcher (like so much of GnuCash) is plagued by over-long and over-complex functions. Breaking those up and testing the pieces will make refactoring--both containers and classes--easier and safer.

christopherlam:
> Maybe write some more tests first so that you can make sure the refactors don't break anything. Also the matcher (like so much of GnuCash) is plagued by over-long and over-complex functions. Breaking those up and testing the pieces will make refactoring--both containers and classes--easier and safer.

Some data structures (eg import-backend.h stuff) could be tested, however many functions use gtk calls heavily, hence, are too labour intensive to test. It would be nice to simply perform each stage (glib->stl, then use cpp classes) stepwise and test thoroughly inbetween.

gnucash/import-export/ofx/gnc-ofx-import.cpp
```diff
@@ -1278,49 +1284,48 @@ reconcile_when_close_toggled_cb (GtkToggleButton *togglebutton, ofx_info* info)
...
-static gchar* make_date_amount_key (time64 date, gnc_numeric amount)
+static const std::string
+make_date_amount_key (time64 date, gnc_numeric amount)
 {
     // Create a string that combines date and amount, we'll use that for our hash
     gchar buf[64];
     gnc_numeric _amount = gnc_numeric_reduce(amount);
-    g_snprintf (buf, sizeof(buf), "%" PRId64 "%" PRId64 "%" PRId64, _amount.num , _amount.denom, date);
-    return g_strdup (buf);
+    g_snprintf (buf, sizeof(buf), "%s%" PRId64 "%" PRId64, GncDateTime(date).format("%y%m%d").c_str(), _amount.num , _amount.denom);
```
christopherlam, line 1296: this "%y%m%d" format ensure the resulting `buf` together with the num+denom is typically 15 bytes or less, ensuring SSO is done in both gcc and clang...

jralls:
> Some data structures (eg import-backend.h stuff) could be tested, however many functions use gtk calls heavily, hence, are too labour intensive to test. It would be nice to simply perform each stage (glib->stl, then use cpp classes) stepwise and test thoroughly inbetween.

MVC is the answer to that: Only the View code has Gtk calls, the model and view code is fully testable. Yes, it's a lot of work, but there's not much point to rewriting C into C++ without any redesign.

gnucash/import-export/ofx/gnc-ofx-import.cpp
```diff
@@ -1475,7 +1477,7 @@ void gnc_file_ofx_import (GtkWindow *parent)
 
         DEBUG("Opening selected file(s)");
         // Create the structure that holds the list of files to process and the statement info.
-        info = g_new(ofx_info,1);
+        info = new ofx_info;
```
christopherlam: I wonder if this `info = new ofx_info;` is safe? `info` is being passed to libofx.

gnucash/import-export/ofx/gnc-ofx-import.cpp
jralls: Typically in rich countries. 6 bytes for the date leaves 9; if denom is 100 that leaves only 6 for num, so any amount over 9999.99 overflows. That won't even buy a cup of coffee in inflated currencies like VND or COP.

If you want to save allocations create a stack `buf` in the calling function, wrap it in a std::string_view and pass the std::string_view& into `make_date_amount_key`.  That's safe to pass to `trans_hash.find` but not to store anywhere.

gnucash/import-export/ofx/gnc-ofx-import.cpp
jralls: If it's the info getting deleted at 1199 then it's safe.

Edit: Note that that's not necessarily true if you're replacing `g_new0` because that 0-initializes the memory and `new` doesn't. `new` will also call the class/struct default constructor and `g_new` won't. That may or may not matter. If there's no explicit default constructor the compiler will generate one that 0-initializes complex member variables (i.e. structs/classes) but not plain-old-data ones like ints or char*.

christopherlam: > If it's the info getting deleted at 1199 then it's safe.
> 
> Edit: Note that that's not necessarily true if you're replacing `g_new0` because that 0-initializes the memory and `new` doesn't. `new` will also call the class/struct default constructor and `g_new` won't. That may or may not matter. If there's no explicit default constructor the compiler will generate one that 0-initializes complex member variables (i.e. structs/classes) but not plain-old-data ones like ints or char*.

I understood `new` won't zero-initialize the struct; the query is because `info` will be sent via `ofx_set_statement_cb(libofx_context, ofx_proc_statement_cb, info)` in `gnc_file_ofx_import_process_file` to libofx and I wasn't sure if the upgraded `info` would be handled.

jralls: It's not upgraded. There's no difference in the returned pointer, it just comes from a different memory pool. 

> because info will be sent via ofx_set_statement_cb(libofx_context, ofx_proc_statement_cb, info)

Not exactly. `ofx_set_statement_cb` tells libofx that [`ofx_proc_statement_cb`](https://github.com/Gnucash/gnucash/blob/0a9e999cfd9f0f4c19cf8c1b79c9c8c1285af542/gnucash/import-export/ofx/gnc-ofx-import.cpp#L1016) is our statement-processing callback and that libofx should pass `info` as the second argument to `ofx_proc_statement_cb`.  [`struct _ofx_info`](https://github.com/Gnucash/gnucash/blob/0a9e999cfd9f0f4c19cf8c1b79c9c8c1285af542/gnucash/import-export/ofx/gnc-ofx-import.cpp#L72) contains GnuCash variables that libofx doesn't know anything about.

jralls:
Speaking of `ofx_info` a very early C++ change should be to remove the two `typedef`s at lines 69 and 72; the first can just go away as C++ understands `struct OfxTransactionData` as the declaration of type `OfxTransactionData`, the `typedef` is meaningless. For the same reason the second one should become `struct OfxInfo` with a global renaming of all uses to comply with our naming convention and to make it a proper C++ declaration.

gnucash/import-export/ofx/gnc-ofx-import.cpp
christopherlam: Ok. Last commit does upgrade info.

gnucash/import-export/ofx/gnc-ofx-import.cpp
```diff
@@ -1476,17 +1478,10 @@ void gnc_file_ofx_import (GtkWindow *parent)
...
-        info->num_trans_processed = 0;
-        info->statement = NULL;
-        info->last_investment_account = NULL;
-        info->last_import_account = NULL;
-        info->last_income_account = NULL;
-        info->parent = parent;
-        info->run_reconcile = FALSE;
-        info->file_list = selected_filenames;
-        info->response = 0;
+        auto info = new OfxInfo (parent, selected_filenames);
```
jralls, line 1483: If you make this a `std::unique_ptr` you won't have to worry about deleting it later, including worrying about there being a path that can escape the `delete`.

christopherlam: I'm not sure how to write this as a unique pointer and get automatic memory management. Additionally the info must be sent to ofx calls.

jralls: Replace `new OfxInfo` with `std::make_unique<OfxInfo>` to get a std::pointer. Use `info.get()` to get a plain `OfxInfo*` to hand to `ofx_set_foo_cb` to pass on to our libofx callbacks.

christopherlam: see latest commit for attempt. `.get()` is ok for unique_ptr->raw_ptr but we'd also need the return trip.

gnucash/import-export/ofx/gnc-ofx-import.cpp
```diff
@@ -1199,23 +1203,20 @@ gnc_ofx_process_next_file (GtkDialog *dialog, gpointer user_data)
     else
     {
         // Final cleanup.
-        delete info;
+        info.reset();
```
jralls: ~~Not necessary, info take care of itself when it goes out of scope.~~ Edit: Never mind, the problem is bigger than that.

gnucash/import-export/ofx/gnc-ofx-import.cpp
jralls: There is no return trip. The weird recursion has to be replaced with a loop in `gtk_file_osx_import`.

gnucash/import-export/ofx/gnc-ofx-import.cpp
christopherlam: > ~Not necessary, info take care of itself when it goes out of scope.~ Edit: Never mind, the problem is bigger than that.

Yups. Hence my preference to carefully commit small changes, and perform the bigger changes as a separate PR.

PR #1715, opened 2023-07-20
Fix test leaks from get_random_string()

richardcohen:


christopherlam:
Triple thanks #1713 #1714 #1715

PR #1719, opened 2023-07-21
Some fixes for (gnc-mktime)

richardcohen:


common/base-typemaps.i
```diff
@@ -119,7 +119,7 @@ typedef char gchar;
 %typemap(argout) struct tm * {
     struct tm* t = $1;
     SCM tm = $input;
-    if (t == NULL)
```
jralls, line 122: This egregious an error says the typemap should be simply removed.

bindings/guile/test/test-date-utilities.scm
```diff
@@ -193,3 +194,10 @@
   (test-equal "11/09/1997 = Q3"
     "Q3"
     (gnc:date-get-quarter-string (create-datevec '(1997 9 11 08 14 21)))))
+
+(define (test-mktime-updates-tm)
+  (let ((july-32 (create-datevec '(2023 7 32 0 0 1))))
+    (gnc-mktime july-32)
+    (test-equal "weekday is Tuesday" 2 (tm:wday july-32))
+    (test-equal "month wraps to August" 7 (tm:mon july-32))
+    (test-equal "day wraps to 1" 1 (tm:mday july-32))))
```
jralls: There is no point at all in testing a wrapper, and this is a wrapper of a wrapper since `gnc_mktime` just wraps `static_cast<time64>(GncDateTime(struct tm*))`. If you think this test is important and missing, add it to libgnucash/engine/test/gtest-gnc-datetime.cpp.

jralls (changes requested):

bindings/guile/test/test-date-utilities.scm
richardcohen: > There is no point at all in testing a wrapper

That is clearly not true, since a simple test exposes the obvious bug in the wrapper.
(gnc-mktime) is a public API, so should be held to a high standard of correctness.

> add it to libgnucash/engine/test/gtest-gnc-datetime.cpp.

The test is testing the (gnc-mktime) wrapper, and needs to be in scheme.

jralls: > since a simple test exposes the obvious bug in the wrapper.

It does not. It exposes a bug in the typemap code that happens to affect the gnc-mktime wrapper among others as you note in your commit message.  There are no unit tests for base-typemaps.i is tested. Obviously there should be. It misses the point of unit tests to test a unit by proxy somewhere unrelated to the actual code of interest.

jralls:
Thanks!

PR #1726, opened 2023-07-25
Further amendment to use new/delete to handle vector<string>

christopherlam:
I guess the `_main_matcher_info` should be inited via `new`/`delete` instead of `g_new0`/`g_free` because of the vector.


jralls:
I don't think this is a good idea without at least writing constructors and a destructor and getting rid of the coy typedef. It really should be the last step of a thorough rewrite into real C++ instead of "better C".

christopherlam:
It's too difficult right now to safely convert to c++ so I'd prefer to use GList instead of `vector<Transaction*>`

christopherlam:
I think it's much simpler to revert recent changes and move a few instructions in `gnc_gen_trans_list_add_trans_internal` which ~should~ does achieve the same result of fast imports.

PR #1728, opened 2023-07-27
Import Matcher: After clicking/toggling A/U+C/C checkbox, reselect the row

christopherlam:
thus it'll be quick to press `up`/`down` and press `space` again to repeat the same action to the next row.

### DIFF
--- a/gnucash/import-export/import-main-matcher.cpp
+++ b/gnucash/import-export/import-main-matcher.cpp
@@ -674,6 +674,15 @@ run_match_dialog (GNCImportMainMatcher *info,
 }
 
 static void
+set_treeview_selection_from_path (GtkTreeView* view, const char* path)
+{
+    auto selection = gtk_tree_view_get_selection (view);
+    auto tree_path = gtk_tree_path_new_from_string (path);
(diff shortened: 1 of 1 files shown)


gnucash/import-export/import-main-matcher.cpp
```diff
@@ -689,6 +689,12 @@ gen_trans_common_toggled_cb (GtkCellRendererToggle *cell_renderer, gchar *path,
     else
         gnc_import_TransInfo_set_action (transaction_info, action);
     refresh_model_row (gui, model, &tree_iter, transaction_info);
+
+    auto treeview = GTK_TREE_VIEW(gui->view);
+    auto selection = gtk_tree_view_get_selection (treeview);
+    auto tree_path = gtk_tree_path_new_from_string (path);
+    gtk_tree_selection_select_path (selection, tree_path);
+    gtk_tree_path_free (tree_path);
```
jralls: Let's make this a separate function `set_treeview_selection_from_path (GtkTreeView* view, const char* path)` and call it here for clarity.

PR #1731, opened 2023-08-01
Pricedb to stl

christopherlam:
first easy steps.

### DIFF
--- a/libgnucash/engine/gnc-pricedb.cpp
+++ b/libgnucash/engine/gnc-pricedb.cpp
@@ -690,14 +690,14 @@ compare_prices_by_date(gconstpointer a, gconstpointer b)
                          gnc_price_get_guid((GNCPrice *) b));
 }
 
-static bool
-price_is_duplicate (const GNCPrice *pPrice, const GNCPrice *cPrice)
+static int
+price_is_duplicate (const GNCPrice *p_price, const GNCPrice *c_price)
 {
     /* If the date, currency, commodity and price match, it's a duplicate */
-    return (gnc_numeric_equal (gnc_price_get_value (pPrice), gnc_price_get_value (cPrice)) &&
-            gnc_price_get_commodity (pPrice) == gnc_price_get_commodity (cPrice) &&
-            gnc_price_get_currency (pPrice) == gnc_price_get_currency (cPrice) &&
-            time64CanonicalDayTime (gnc_price_get_time64 (pPrice)) == time64CanonicalDayTime (gnc_price_get_time64 (cPrice)));
+    return time64CanonicalDayTime (gnc_price_get_time64 (p_price)) != time64CanonicalDayTime (gnc_price_get_time64 (c_price)) ||
+        gnc_numeric_compare (gnc_price_get_value (p_price), gnc_price_get_value (c_price)) ||
+        gnc_commodity_compare (gnc_price_get_commodity (p_price), gnc_price_get_commodity (c_price)) ||
+        gnc_commodity_compare (gnc_price_get_currency (p_price), gnc_price_get_currency (c_price));
 }
 
 gboolean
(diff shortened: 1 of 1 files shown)


jralls:
@christopherlam Re https://github.com/Gnucash/gnucash/commit/2d730e38d5aa4c1f54c658636c9b0f671dcc21e4 and https://github.com/Gnucash/gnucash/commit/6b741ddf320cbab9bc2953523e6d12cb02f66a8f

What's the point of having a PR--one marked Draft no less--when you're just going to push the commits directly to stable?

BTW, I think the struct->std::pair change isn't a great idea on its own: It adds a bunch of overhead, and reduces expressiveness because "first" and "second" are less meaningful than "to" and "from". I didn't comment here because you have it marked draft and I was waiting for you to do something that would actually use some feature of the pair.

christopherlam:
Ok we can revert the `std::pair`. The idea was to reduce num(LOC); lots of repetition in this file. Eventually (maybe) use `std::unordered_map`.

libgnucash/engine/gnc-pricedb.cpp
```diff
@@ -690,14 +690,14 @@ compare_prices_by_date(gconstpointer a, gconstpointer b)
                          gnc_price_get_guid((GNCPrice *) b));
 }
 
-static bool
-price_is_duplicate (const GNCPrice *pPrice, const GNCPrice *cPrice)
+static int
+price_compare (const GNCPrice *pPrice, const GNCPrice *cPrice)
```
jralls: Keep the old name. This isn't a real `GCompareFunc`, it can't return -1 because of the use of `operator||` so it can't be used for sorting.

You should also fix the variable names (p_price instead of pPrice) to comply with the coding standard, and since a price list should have only prices between identical commodity/currency pairs and should contain only one price per date I'd check date, value, commodity, then currency.

libgnucash/engine/gnc-pricedb.cpp
```diff
@@ -2338,6 +2324,11 @@ gnc_pricedb_lookup_nearest_in_time64(GNCPriceDB *db,
     return lookup_nearest_in_time(db, c, currency, t, FALSE);
 }
 
+// return 0 if price's time is less or equal to time
+static int price_less_or_equal (GNCPrice *p, time64 time)
```
jralls: `price_time64_less_or_equal`.

libgnucash/engine/gnc-pricedb.cpp
```diff
@@ -2193,30 +2178,31 @@ gnc_pricedb_lookup_day_t64(GNCPriceDB *db,
     return lookup_nearest_in_time(db, c, currency, t, TRUE);
 }
 
+// return 0 if price's time matches exactly
+static int price_equal (GNCPrice *p, time64 time)
+{
+    return !(gnc_price_get_time64 (p) == time);
+}
+
```
jralls: `price_same_time`.

jralls (changes requested):

libgnucash/engine/gnc-pricedb.cpp
```diff
@@ -2346,26 +2337,17 @@ gnc_pricedb_lookup_nearest_before_t64 (GNCPriceDB *db,
...
     GNCPrice *current_price = NULL;
-
     if (!db || !c || !currency) return NULL;
     ENTER ("db=%p commodity=%p currency=%p", db, c, currency);
-
     auto price_list = pricedb_get_prices_internal (db, c, currency, TRUE);
     if (!price_list) return NULL;
-
-    for (auto item = price_list; item; item = item->next)
+    auto p = g_list_find_custom (price_list, GUINT_TO_POINTER(t), (GCompareFunc)price_time64_less_or_equal);
```
jralls, line 2344: `GUINT_TO_POINTER` is an evil thing that subverts the type system. There are two problems in this case: First of all, `time64`s are signed and second they're int64_t. GUINT is an alias for uint32_t, and pointers on 32-bit builds (Windows!) are 32 bits. See the patch on [bug 799104](https://bugs.gnucash.org/show_bug.cgi?id=799104). The user doesn't explain themself well, but I guess they're doing forecasting and have dates past 2038 when time goes over 32 bits.

christopherlam: Oops.

PR #1736, opened 2023-08-06
Bug 798990 - Notes No Longer Autofills

Bob-IT:
With the change to delay loading the non focused registers, the notes and memo quick fills are empty. This is due to trying to populate the quick fill on the first load when there are no splits so change populating the quick fill lists to the subsequent load only.

This works on my Linux VM machine and is the only thing I thought of. Will test on my Windows build machine once it has finished a backup.

### DIFF
--- a/gnucash/register/ledger-core/split-register-load.c
+++ b/gnucash/register/ledger-core/split-register-load.c
@@ -354,6 +354,10 @@ update_info (SRInfo* info, SplitRegister* reg)
         gnc_split_register_get_current_trans_split (reg, NULL);
     info->cursor_hint_cursor_class =
         gnc_split_register_get_current_cursor_class (reg);
+
+    if (!info->first_pass && !info->quickfill_setup)
+        info->quickfill_setup = TRUE;
+
     info->hint_set_by_traverse = FALSE;
     info->traverse_to_new = FALSE;
     info->exact_traversal = FALSE;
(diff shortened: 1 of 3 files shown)


jralls:
@Bob-IT That would explain why memo doesn't get populated, but notes is in the transaction's KVP, not in splits. Is loading KVP also delayed?

Bob-IT:
@jralls `gnc_split_register_load` on the first load has an empty `slist` and on line 752 it tries to add the completions when looping over the list.
As the list is empty, it never gets to do the `add_quickfill_completions` in side the loop but `info->first_pass` is set to `false` near the end of the function.

On subsequent loads, `info->first_pass` is `false` and so `add_quickfill_completions` is not called.

`add_quickfill_completions` uses the current `split` from the list with the `trans` value obtained from the parent of the split. So the meno text is from the `split` by `xaccSplitGetMemo` and the note text from the parent of `split` by`xaccTransGetNotes`.

jralls:
Got it. Next question: Why load the quickfill completions only once but add to the description menu every time? Doesn't that mean that if the user adds a new note it won't be available for completion until they close and reopen the register?

Bob-IT:
With the existing quick fill cells, as far as I understand there is no easy way to remove or update entries. So if you amended say a description the old one would stay in the list until the register was reloaded like wise if you deleted a description that would stay. 

So with the new description cell I wanted to fix that and the easiest way was to refresh the hash table on every load. For notes and memo it is probably less annoying to have the old entries around.

The quick fill for the notes and memo text is updated as soon the cursor is moved by a callback in `gnc_basic_cell_set_value`, line 223  that calls `gnc_quickfill_insert`. In testing all you need to do is enter some text in these cells, Tab and it will be in the list.

jralls:
Thanks!

PR #1737, opened 2023-08-06
Bug799051 - Shortcut Ctrl + Tab not working in 5.3

Bob-IT:
In a register description cell pressing Ctrl+Tab would populate the description cell if the list had only one entry, i.e. all the register transaction descriptions were the same. This was removed by mistake so this commit adds it back.

I think this is OK, have tested it on my Linux VM and will test on my Windows build machine once a backup has finished.


jralls:
LGTM

Bob-IT:
Built and tested on my Windows PC and all OK, will push later.

PR #1742, opened 2023-08-10
gnome-utils removed from gnucash-cli

christopherlam:


gnucash/gnucash-core-app.hpp
```diff
@@ -70,6 +70,8 @@ class CoreApp
     char *sys_locale = nullptr;
 };
 
-void gnc_load_scm_config(void);
+typedef void (*MessageCb) (const char *msg);
```
jralls: That's the C way. Use [`std::function`](https://en.cppreference.com/w/cpp/utility/functional/function) instead and use lambdas instead of named callback functions.

gnucash/gnucash-commands.cpp
```diff
@@ -118,6 +118,12 @@ write_report_file (const char *html, const char* file)
     // ofs destructor will close the file
 }
 
+static void
+update_message_cli (const gchar *msg)
```
jralls: Leaving aside using lambdas, this name and its gui counterpart do a poor job of communicating intent. Since the parameter they're passed to is called `update_message_cb` they would be better named `cli_update_message_cb` and `gui_update_message_cb`.

jralls (changes requested):

gnucash/gnucash.cpp
```diff
@@ -155,7 +161,7 @@ scm_run_gnucash (void *data, [[maybe_unused]] int argc, [[maybe_unused]] char **
     /* Load the scm config files before starting up the gui. This ensures that
      * custom reports have been read into memory before the Reports
      * menu is created. */
-    Gnucash::gnc_load_scm_config();
+    Gnucash::gnc_load_scm_config (gui_update_message_cb);
```
jralls: I had in mind that you'd do
```
     Gnucash::gnc_load_scm_config(
                               [](const gchar *msg)
                               {
                                   gnc_update_splash_screen (msg, GNC_SPLASH_PERCENTAGE_UNKNOWN);
                                   PINFO ("%s", msg);
                                });
```
But using the named lambda is OK too, Just move the decls inside their respective function scopes. Right now they're global variables, not a Good Thing.

christopherlam: the cb is needed in several functions in `gnucash-commands.cpp`

jralls: It's also a one-liner that's mostly a no-op, so it wouldn't hurt anything to use the immediate lambda in all three places. It might even be better: By creating the lambda in each function it might capture enough context to print the right function in the log message. I'm not sure about that, but it might.

OTOH it doesn't matter if it's a lambda or a regular function, and `std::function` works with both. Either way declare it `static` so it's only file-global and not global-global.

christopherlam: Agree that inlining the single liner is better. A top-level lambda shows 

```
* 19:57:07  WARN <gnc.gui> [<lambda()] Loading system wide Guile extensions…
* 19:57:07  WARN <gnc.gui> [<lambda()] Loading user specific Guile extensions…
```

and the inlined lambda shows
```
* 19:57:41  WARN <gnc.gui> [scm_report_list()] Loading system wide Guile extensions…
* 19:57:41  WARN <gnc.gui> [scm_report_list()] Loading user specific Guile extensions…
```

gjanssens:
I love what you did here :+1:

PR #1744, opened 2023-08-11
Implement support for !Type:Prices QIF records.

jwhite66:


gnucash/import-export/qif-imp/qif-to-gnc.scm
```diff
@@ -449,6 +455,87 @@
...
+                  (if (equal? commodity #f)
+                      (qif-import:log progress-dialog
+                          "qif-import:qif-to-gnc"
+                          (G_ (format #f "Warning: cannot find commodity for symbol ~a." current-symbol))))))
+            (if (not (equal? commodity #f))
+                (if (not (duplicate-price? pricedb commodity default-currency price))
+                    (set! pruned-prices (append pruned-prices (list price)))
+                    (set! pruned-price-count (+ 1 pruned-price-count))))
+            (update-progress))
+         sorted-prices)
```
christopherlam:
```suggestion

        ;; Prune duplicate records before inserting
        (let lp ((sorted-prices sorted-prices)
                 (pruned '())
                 (count 0))
          (cond
           ((null? sorted-prices)
            (set! pruned-prices (reverse pruned))
            (set! pruned-price-count count))
           (else
            (update-progress)
            (unless (string-ci=? current-symbol (qif-price:symbol price))
              (set! current-symbol (qif-price:symbol price))
              (set! commodity (hash-ref symbol-hash current-symbol))
              (unless commodity
                (qif-import:log progress-dialog
                                "qif-import:qif-to-gnc"
                                (G_ (format #f "Warning: cannot find commodity for symbol ~a." current-symbol)))))
            (when commodity
              (if (duplicate-price? pricedb commodity default-currency price)
                  (lp (cdr sorted-prices)
                      pruned
                      (1+ count))
                  (lp (cdr sorted-prices)
                      (cons price pruned)
                      count))))))
```

christopherlam (commented):
just an o(n^2) fix.

gnucash/import-export/qif-imp/qif-to-gnc.scm
jwhite66: I'm confused.  I don't see an O(n^2) problem.  I have 70k records that are pruned in about 29 seconds, which feels reasonable to me.  The suggestion as provided does not work.  It does not do a (let ((price (cdr sorted-prices)) ..., and the first unless needs a call to (lp (cdr..., to keep the logic going.  And then the logic flow does not work when there are multiple prices with a bad symbol in a row (in fact, it incorrectly exits early), whereas my code pruned those just fine.  Can you explain the flaw you saw so that I can try to address it?  Is (append ...) expensive in some way I don't understand?

christopherlam: https://chat.openai.com/share/4f5758a1-1e00-4864-8e71-8f3903d7eedf ;)

christopherlam: If you wish to retain your code structure, then simply build pruned-list via `(set! pruned-prices (cons price pruned-prices))` then when the forloop completes, `(set! pruned-prices (reverse pruned-prices))`

jwhite66: Ah, yes, that is substantially faster.  I've made that change, and also caught an invalid data record in my large data set.

jralls: @jwhite66 do you understand *why* it's faster?

jwhite66: Well, I presume it's because the internal implementation of a scheme list is a singly linked list, and so append requires a full list traversal on each call.  For the record, I think the other point Christopher was making was that my code is not idiomatic scheme.  That is unsurprising, as it's my first substantive bit of scheme :-/.   I would prefer the code go in as I wrote it because I believe I understand it, and have some comfort that I have developed and tested those code paths as written, no matter how naive.   I did study the code shared and felt that I should embrace at least when and unless.  I never really enjoyed the recursion first approach of lisp like languages, so that may not come so readily to me.

jralls: Yes, that's right. 

I understand your preference but we need to be concerned about maintaining your contribution going forward, meaning that other people--maybe even your future self who's learned more scheme but forgotten this PR--can easily understand it and how it fits in with the rest of the importer.

gnucash/import-export/qif-imp/qif-to-gnc.scm
```diff
@@ -449,6 +455,88 @@
...
+                      (qif-import:log progress-dialog
+                          "qif-import:qif-to-gnc"
+                          (G_ (format #f "Warning: cannot find commodity for symbol ~a." current-symbol))))))
+            (if (not (equal? commodity #f))
+                (if (not (duplicate-price? pricedb commodity default-currency price))
+                    (set! pruned-prices (cons price pruned-prices))
+                    (set! pruned-price-count (+ 1 pruned-price-count))))
+            (update-progress))
+         sorted-prices)
+        (set! pruned-prices (reverse pruned-prices))
```
christopherlam, line 506: A few comments if you want: you're concatenating several lists into 1 long list via `append`, then the long list is sorted, and filtered. 

1. it's marginally faster to use `append-reverse` instead of `append` -- see https://github.com/cky/guile/blob/89ce9fb31b00f1f243fe6f2450db50372cc0b86d/libguile/srfi-1.c#L81. 

2. IMHO instead of `append-reverse` it's better to filter and accumulate at the same time, then sort at the end.

Would you want to squeeze a few milliseconds from this?

jwhite66: Thanks for the tip on append.

My instinct is that it's not worth the optimization.  The large file I'm using takes over 15 minutes to import, without the prices; the prices bit adds perhaps a minute to the overall timeline.  A few milliseconds won't make much difference to the user experience.   I also suspect that the vast majority of import operations are done against exactly one qif file.

I did initially write this code without the sorting, and it sorted + filtered as it went and was dissatisfied and rewrote it into this form. You get some efficiency from sorting, as you don't need to look up the commodity for each transaction that way.  It also makes for a natural way of emitting a warning when a symbol is not found only once.  The original code generated one warning for every price record with a missing symbol, which polluted the log.

gnucash/import-export/qif-imp/qif-to-gnc.scm
jwhite66: Sure, and as maintainers you have a duty to make sure the code base has the quality it needs, and I am willing to make changes until it passes muster.  From my naive and definitely inexperienced perspective, it seems to be Good Enough (TM).  The qif import code itself feels a bit on the naive side, so my code seemed to fit right in :-/.  Again, I'm willing to make changes, but I would need more specific direction on which parts were obviously wrong.

gnucash/import-export/qif-imp/qif-to-gnc.scm
christopherlam: It would be useful (for us) to understand your large file. e.g. number of accounts, transactions, prices including invalid prices, etc. A sample large datafile with bogus but realistic data _would_ be nice. IMO 15mins wait is 14min too long.

christopherlam: FWIW what would block this merge is that we can confirm it works well with a sample (small or large) datafile with transactions and prices. In bugzilla would be ok too.

jwhite66: Let me think on that.  It's 30 years of personal history and I have an obvious concern regarding security.   I've got some scripts that transform QIF files; I'll see if I can make myself comfortable that I can sanitize it.

I ran perf against it, and my initial results didn't give me any sense of aha!  The bulk of time (86%) is in the trans_cleanup_commit call stack which does not feel startling to me.

If I trust quiffen, it's got 54 accounts, 26 securities, and 115 categories.  There are 23K transactions, with the bulk of them in a single checking account.  The qif file itself is about 5.4M; about half of that is used for the ~70k prices records.  It's a typical qif file with those flaws, and Quicken had been allowed to 'archive' some data, which screws up the precision of the data in some frustrating ways.

Note that I have found a few other flaws with the import which I'm hoping to figure out and ideally submit patches for.  For example, a transfer of a security from one fund to another seems to fail to result in a shares only transaction; instead half is shares, half is cash.  I'd like to see if I can 'undo' the damage of the Quicken archive; I see a path to do it manually, and I'm unsure if it could be automated.  I'm also hoping to agitate for an auto prefixing of account names, but I need to do more homework first.

At the very least, I'll pull together some representative sample files that illustrate what I've seen.

jwhite66: @christopherlam : I checked with my wife, and she is uncomfortable with me releasing that data, no matter how sanitized I think it may be.  I'm willing to break out samples, and to run tests against it if that will help.  I did add a sample file that shows prices to this PR.

christopherlam: Thanks for adding this datafile, which does include an invalid price entry which is nicely handled:
```
Line 83: Could not parse price line: "ABC",," 1/18'38"
```

Is it intentional that it displays warnings about missing symbols DEF and GHI when importing into a blank datafile?
```
Warning: cannot find commodity for symbol DEF.
Warning: cannot find commodity for symbol GHI.
```

jwhite66: Yes, those are also intentional errors.

gnucash/import-export/qif-imp/qif-file.scm
```diff
@@ -1083,3 +1133,40 @@
...
+                 (string->number (cadr fracprice)))))
+              (set! price (format #f "~f" total))))
+
+          (qif-price:set-share-price! current-xtn price)
+
+          ;; A blank entry will not necessarily throw an exception, but is invalid
+          (if (or (string-null? symbol) (or (string-null? price) (string-null? date)))
+            (set! current-xtn #f)))))
+       (set! current-xtn #f))
+    current-xtn))
```
christopherlam: I'll rewrite it to avoid exceptions and use pattern matching instead. It also includes a home grown string to numeric parser which is too lenient. "1 3/4" is converted to 1.75 and "1 2" will be converted to "3" without erroring out. Does @jralls know of such a parser that can handle "1" "2 3/4" well and decline to parse "4 5" within the engine?

jralls: It's not just too lenient, it generates floating point when it should create either gnc_numeric or Scheme exact rational. 

Try `string-to-gnc-numeric` for parsing, that's what pretty much everything else uses. I don't know what it does with internal whitespace when the second number isn't a fraction.

christopherlam: https://github.com/Gnucash/gnucash/pull/1756 if gnc_numeric_to_string is upgraded.

christopherlam: Ok with #1756 merged in this function can be rewritten as
```scheme

(use-modules (ice-9 match))
(define (qif-file:parse-price-line line)
  (define (remove-punctuation s) (string-trim-both s char-set:punctuation))
  (and (string? line)
       (match (string-split line #\,)
         (((= remove-punctuation symbol) (= gnc-numeric-from-string value) (= remove-punctuation date))
          (cond
           ((string-null? symbol) #f)
           ((not value) #f)
           ((string-null? date) #f)
           (else
            (let ((xtn (make-qif-price)))
              (qif-price:set-symbol! xtn symbol)
              (qif-price:set-date! xtn date)
              (qif-price:set-share-price! xtn (number->string value))
              xtn))))
         (_ #f))))
```

jwhite66: Hmm.   That does not drop in and just work.; the import fails. The resulting strings from this function are, I presume, scheme rational numbers (i.e. I see "21/20" and so on).  But qif-file:parse-fields wants to reprocess all transactions, expecting string input.   And that seems to boil down to a call to  qif-parse:parse-number/format, which, afaict, does not handle this format - it has decimal, comma, and integer types.

christopherlam: How annoying. The qif-parser was written 20 years ago.

```scheme
modified   gnucash/import-export/qif-imp/qif-file.scm
@@ -832,7 +832,7 @@
       (start-sub 1)
       (check-and-parse-field
        qif-price:share-price qif-price:set-share-price! gnc-numeric-equal
-       qif-parse:check-number-format '(decimal comma)
+       qif-parse:check-number-format '(decimal comma rational)
        qif-parse:parse-number/format (qif-file:prices self)
        qif-parse:print-number
        'guess-on-ambiguity add-error 'share-price
@@ -883,7 +883,7 @@
       (start-sub 0.1)
       (check-and-parse-field
        qif-xtn:share-price qif-xtn:set-share-price! gnc-numeric-equal
-       qif-parse:check-number-format '(decimal comma)
+       qif-parse:check-number-format '(decimal comma rational)
        qif-parse:parse-number/format (qif-file:xtns self)
        qif-parse:print-number
        'guess-on-ambiguity add-error 'share-price
modified   gnucash/import-export/qif-imp/qif-parse.scm
@@ -404,6 +404,11 @@
   (and regexp-enabled?
        (make-regexp "^ *[$]?[+-]?[$]?[0-9]+[+-]?$|^ *[$]?[+-]?[$]?[0-9]?[0-9]?[0-9]?([\\.'][0-9][0-9][0-9])*(,[0-9]*)?[+-]? *$|^ *[$]?[+-]?[$]?[0-9]+,[0-9]*[+-]? *$")))
 
+;; eg 1/2 or 32/15 or 4
+(define rational-regexp
+  (and regexp-enabled?
+       (make-regexp "^-?[0-9]+(/[0-9]+|)$")))
+
 ;; eg 456 or 123
 (define integer-regexp
   (and regexp-enabled?
@@ -419,6 +424,7 @@
   (define numtypes-alist
     (list (cons 'decimal decimal-radix-regexp)
           (cons 'comma comma-radix-regexp)
+          (cons 'rational rational-regexp)
           (cons 'integer integer-regexp)))
   (filter (lambda (fmt) (regexp-exec (assq-ref numtypes-alist fmt) value-string))
           possible-formats))
@@ -439,7 +445,7 @@
                         ((comma) (gnc:string-replace-char
                                   (gnc:string-delete-chars filtered-string ".")
                                   #\, #\.))
-                        ((integer) filtered-string)))
+                        ((integer rational) filtered-string)))
          (num (or (string->number (string-append "#e" read-string)) 0)))
     (if has-minus? (- num) num)))
```

jwhite66: Good to know that I've learned everything I know about Guile from that code base.  If only this was 2001, I'd be so cool  :-).

jwhite66: I've in theory pushed your code into this PR, and, if I rebase to include the new code, it seems to work for me.

christopherlam:
LGTM...

gnucash/import-export/qif-imp/qif-to-gnc.scm
```diff
@@ -1259,3 +1346,98 @@
...
+;; qif-date-to-time64
+;;
+;; Convert our internal qif-date format into a time64
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+(define (qif-date-to-time64  qif-date)
+  (let ((tm (gnc-localtime (current-time))))
+    ;; Choose a 'neutral' time of day, instead of time of import
+    (set-tm:hour tm 11)
+    (set-tm:min tm 0)
+    (set-tm:sec tm 0)
```
jralls: We use 10:59:00Z for neutral time because 11:00:00Z is 00:00:00 the next day in New Zealand summer time.

jralls (changes requested):

jralls (approved):

jralls:
I took care of neutral time and merged. Thanks!

PR #1745, opened 2023-08-12
iter to stl

christopherlam:
~draft. too difficult~ to convert fully to `std::iterator` like methods.
EDIT: complete...

### DIFF
--- a/gnucash/gnome-utils/CMakeLists.txt
+++ b/gnucash/gnome-utils/CMakeLists.txt
@@ -113,6 +113,7 @@ set(gnome_utils_noinst_HEADERS
   gnc-autosave.h
   gnc-gobject-utils.h
   gnc-gtk-utils.h
+  gnc-tree-model-container.hpp
   gnc-option-gtk-ui.hpp
   search-param.h
 )
--- a/gnucash/gnome-utils/gnc-tree-model-container.hpp
+++ b/gnucash/gnome-utils/gnc-tree-model-container.hpp
@@ -1,0 +1,145 @@
+/********************************************************************\
+ * gnc-tree-model-container.hpp
+ *                                                                  *
+ * This program is free software; you can redistribute it and/or    *
+ * modify it under the terms of the GNU General Public License as   *
+ * published by the Free Software Foundation; either version 2 of   *
+ * the License, or (at your option) any later version.              *
+ *                                                                  *
+ * This program is distributed in the hope that it will be useful,  *
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of   *
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the    *
+ * GNU General Public License for more details.                     *
+ *                                                                  *
+ * You should have received a copy of the GNU General Public License*
+ * along with this program; if not, contact:                        *
+ *                                                                  *
+ * Free Software Foundation           Voice:  +1-617-542-5942       *
+ * 51 Franklin Street, Fifth Floor    Fax:    +1-617-542-2652       *
+ * Boston, MA  02110-1301,  USA       gnu@gnu.org                   *
+\********************************************************************/
+
+#ifndef GNC_TREE_MODEL_CONTAINER_HPP
+#define GNC_TREE_MODEL_CONTAINER_HPP
+
+#include <string>
+#include <optional>
+
+class GncTreeData
+{
+public:
+    GncTreeData (GtkTreeModel* model, GtkTreeIter iter) : m_model{model}, m_iter{iter} {};
+
+    template <typename T>
+    T get (int column)
+    {
+        gpointer rv;
+        gtk_tree_model_get(m_model, &m_iter, column, &rv, -1);
+        return static_cast<T>(rv);
+    }
+
+    GtkTreeIter& get_iter () { return m_iter; };
+
+    int get_int (int column)
+    {
+        int rv;
+        gtk_tree_model_get(m_model, &m_iter, column, &rv, -1);
+        return rv;
+    }
+
+    std::string get_string (int column)
+    {
+        auto str = get<char*>(column);
+        std::string rv{str};
+        g_free (str);
+        return rv;
+    }
+
+    bool operator==(const GncTreeData& other) const
+    {
+        return (m_model == other.m_model) &&
+            (m_iter.stamp == other.m_iter.stamp) &&
+            (m_iter.user_data == other.m_iter.user_data) &&
+            (m_iter.user_data2 == other.m_iter.user_data2) &&
+            (m_iter.user_data3 == other.m_iter.user_data3);
+    }
+
+private:
+    GtkTreeModel* m_model;
+    GtkTreeIter m_iter;
+};
+
+// Custom iterator class
+template <typename ModelType>
+class GncTreeIter
+{
+public:
+    GncTreeIter(GtkTreeModel* model, std::optional<GtkTreeIter> iter) : m_model(model), m_iter(iter) {}
+
+    GncTreeIter(GtkTreeModel* model) : m_model (model)
+    {
+        GtkTreeIter iter;
+        gtk_tree_model_get_iter_first(m_model, &iter);
+        m_iter = iter;
+    }
+
+    GncTreeIter& operator++()
+    {
+        if (!m_iter.has_value())
+            throw "no value, cannot increment";
+        if (!gtk_tree_model_iter_next(m_model, &m_iter.value()))
+            m_iter = std::nullopt;
+        return *this;
+    }
+
+    ModelType operator*() const
+    {
+        if (!m_iter.has_value())
+            throw "no value, cannot dereference";
+        return ModelType (m_model, *m_iter);
+    }
+
+    bool has_value() { return m_iter.has_value(); };
+
+    bool operator==(const GncTreeIter& other) const
+    {
+        if (!m_iter.has_value() && !other.m_iter.has_value())
+            return true;
+        if (!m_iter.has_value() || !other.m_iter.has_value())
+            return false;
+        return (ModelType (m_model, *m_iter) == ModelType (m_model, *other.m_iter));
+    }
+
+    bool operator!=(const GncTreeIter& other) const { return !(*this == other); }
+
+private:
+    GtkTreeModel* m_model;
+    std::optional<GtkTreeIter> m_iter;
+};
+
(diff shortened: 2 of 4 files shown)


gnucash/import-export/import-main-matcher.cpp
```diff
@@ -193,33 +195,83 @@ defer_bal_computation (GNCImportMainMatcher *info, Account* acc)
...
-    TreeViewIter (GtkTreeView *view)
-        : m_model{gtk_tree_view_get_model (view)}
-        , m_value{(bool)gtk_tree_model_get_iter_first (m_model, &m_iter)}
-    {};
-    gpointer get (gint col)
+    using iterator_category = std::forward_iterator_tag;
+    using difference_type = ptrdiff_t;
+    GtkTreeModelIterator(GtkTreeModel* model, std::optional<GtkTreeIter> iter)
+        : m_model(model), m_iter(iter) {}
+    GtkTreeModelIterator& operator++()
```
jralls: Please add a blank line between functions!

jralls:
Woot! A huge readability improvement!

gnucash/import-export/import-main-matcher.cpp
```diff
@@ -192,32 +195,95 @@ defer_bal_computation (GNCImportMainMatcher *info, Account* acc)
     }
 }
 
+
+// Custom iterator class
+class GtkTreeModelIterator
+{
+public:
+    using iterator_category = std::forward_iterator_tag;
+    using difference_type = ptrdiff_t;
```
christopherlam: these two lines are in the examples but removal doesn't harm.

gnucash/import-export/import-main-matcher.cpp
```diff
@@ -192,32 +195,95 @@ defer_bal_computation (GNCImportMainMatcher *info, Account* acc)
...
+    {
+        if (m_iter.has_value() && !gtk_tree_model_iter_next(m_model, &*m_iter))
+            m_iter = std::nullopt;
+        return *this;
+    }
+    template <typename T>
+    T get (int column) {
+        g_return_val_if_fail (m_iter.has_value(), T());
+        gpointer rv;
+        gtk_tree_model_get(m_model, &*m_iter, column, &rv, -1);
```
christopherlam: I'm using `std::nullopt` to denote the `end()` iterator. The original suggestions use `GtkTreeIter()` as the constructor for the `end()` iterator but I figure nullopt is safer, especially needing to test equality. This means we need the weird `&*m_iter` construct.

gnucash/import-export/import-main-matcher.cpp
```diff
@@ -192,32 +195,95 @@ defer_bal_computation (GNCImportMainMatcher *info, Account* acc)
...
+// Custom container class
+class GtkTreeModelContainer
+{
+public:
+    using iterator = GtkTreeModelIterator;
+    GtkTreeModelContainer(GtkTreeModel* model) : m_model(model), m_size{0}
+    {
+        for (auto it = begin(); it.has_value(); ++it)
+            m_size++;
+    }
```
christopherlam: this O(N) to calculate `size` is inevitable. Alternately we could forego size() and test empty via `return begin() == end()` in which both calls GtkTreeModelIterator constructor every time?

gnucash/import-export/import-main-matcher.cpp
```diff
@@ -192,32 +195,95 @@ defer_bal_computation (GNCImportMainMatcher *info, Account* acc)
...
-            if (info->transaction_processed_cb)
-            {
-                info->transaction_processed_cb (trans_info, false,
-                                                info->user_data);
-            }
-        }
-        while (gtk_tree_model_iter_next (model, &iter));
-    }
+    if (info->transaction_processed_cb)
+        for (auto iter : GtkTreeModelContainer(gtk_tree_view_get_model (info->view)))
```
christopherlam: instead of `for (auto iter : container)` I'd prefer `for (const auto& iter : container)` however the latter won't compile.

gnucash/import-export/import-main-matcher.cpp
```diff
@@ -192,32 +195,106 @@ defer_bal_computation (GNCImportMainMatcher *info, Account* acc)
...
+    GtkTreeModel* m_model;
+    std::optional<GtkTreeIter> m_iter;
+};
+
+// Custom container class
+template <typename ModelType>
+class GtkTreeModelContainer
+{
+public:
+    using iterator = GtkTreeModelIterator<ModelType>;
```
jralls: Capitalize type names.

gnucash/import-export/import-main-matcher.cpp
jralls: Constructing `end()` is trivial, so that would be much faster for `empty()`. This will anyway work to get the size only of a GtkListModel because a GtkTreeModel's iterators have children reflecting its tree structure.

gnucash/import-export/import-main-matcher.cpp
jralls: You could use `&m_iter.value()` instead.

gnucash/import-export/import-main-matcher.cpp
jralls: That's because the result of `GtkTreeModelContainer(gtk_tree_view_get_model (info->view))` is a temporary and you're not allowed to take references to temporaries. There's apparently a lifetime adjustment in C++23 that will allow it in for loops, see Temporary range expression in [this](https://en.cppreference.com/w/cpp/language/range-for). It will be a few years before we can use that, so a work around would be to create a local and iterate on that:
```
auto info_view{GtkTreeModelContainer(gtk_tree_view_get_model (info->view))};
for (const auto& iter : info_view)
  //…
```

christopherlam: Doesn't work...
```
/home/chris/sources/gnucash/stable/gnucash/import-export/import-main-matcher.cpp:309:27: error: cannot bind non-const lvalue reference of type 'ImportMatcherType&' to an rvalue of type 'ImportMatcherType'
  309 |         for (auto& iter : tree_model)
      |                           ^~~~~~~~~~
```

gnucash/import-export/import-main-matcher.cpp
christopherlam: > This will anyway work to get the size only of a GtkListModel because a GtkTreeModel's iterators have children reflecting its tree structure.

Not quite sure what this means. Anyway, amended.

gnucash/gnome-utils/gnc-gtk-utils.hpp
```diff
@@ -0,0 +1,130 @@
...
+
+    GtkTreeModelContainer(GtkTreeModel* model)
+        : m_model{model}
+        , m_begin_iter{TreeModelIterator(m_model, std::nullopt)}
+        , m_end_iter{TreeModelIterator(m_model, std::nullopt)}
+    {
+        GtkTreeIter iter;
+        gtk_tree_model_get_iter_first(m_model, &iter);
+        m_begin_iter = TreeModelIterator(m_model, iter);
+    }
```
christopherlam: If this last commit is desirable, this ctor will init m_begin_iter with a dummy value before finding the proper beginning iterator. If we remove the initializer line then compilation fails.

christopherlam:
I think the .hpp is now ready... I wouldn't merge the import-main-matcher.cpp until #1708

jralls:
We need a more descriptive name for the header. `gnc-gtk-treeview-container.hpp` would be the normal choice because it's based on the name of the class it declares but it seems a bit wordy.

gnucash/import-export/import-main-matcher.cpp
jralls: Suggesting that you could bind a `const lvalue reference` but IIUC you already tried that. Looks like the compiler is optimizing away tree_model and leaving you with the temporary anyway. That probably means that `iter` is the temporary's memory and taking a reference wouldn't improve performance.

gjanssens:
> We need a more descriptive name for the header. `gnc-gtk-treeview-container.hpp` would be the normal choice because it's based on the name of the class it declares but it seems a bit wordy.

`gnc-treeview-container.hpp` - whenever we inherit from gtk widgets we replace gtk with gnc. I think it will be ok to the the same here.

christopherlam:
Ok. `gnc-tree-model-container.hpp` it is then. (not tree-view).

jralls:
>  whenever we inherit from gtk widgets we replace gtk with gnc. I think it will be ok to the the same here.

It's composition not inheritance, but the `gtk` is reasonably implied. `view` and `model` are similarly irrelevant, so maybe just `gnc-tree-container.hpp` and the classes could be `GncTree(Container|Data|Iter)`.

christopherlam:
> > whenever we inherit from gtk widgets we replace gtk with gnc. I think it will be ok to the the same here.
> 
> It's composition not inheritance, but the `gtk` is reasonably implied. `view` and `model` are similarly irrelevant, so maybe just `gnc-tree-container.hpp` and the classes could be `GncTree(Container|Data|Iter)`.

It's tailored for GtkTreeModel though, with `gtk_tree_model_get_iter_first` in it.

jralls:
> It's tailored for GtkTreeModel though, with gtk_tree_model_get_iter_first in it.

Of course. What's your point?

christopherlam:
> > It's tailored for GtkTreeModel though, with gtk_tree_model_get_iter_first in it.
> 
> Of course. What's your point?

It'd be appropriate to name it GncTreeModel?

jralls:
> It'd be appropriate to name it GncTreeModel?

It would not, because it's not a stand-alone tree model. You can't create a GncTreeModel on its own, you have to create a GtkTreeModel and pass it in. It doesn't take ownership, and you need to keep the GtkTreeModel* handy so you can load it, mutate it, connect to its signals,  assign it to a GtkTreeView, and ultimately destroy it. You can't even safely keep it around because its `m_model` can be freed out from under it.

christopherlam:
Ok `GncTreeContainer` it is then.

christopherlam:
Re model it's probably better to pepper some `GTK_IS_TREE_MODEL` around then.

jralls:
> Re model it's probably better to pepper some GTK_IS_TREE_MODEL around then.

Should be needed only in the constructor to make sure that you really have one. Since you have to write a ctor you might as well ref the thing and write a dtor to unref it or make it a unique_ptr with a custom deleter that unrefs it. That will protect it from being freed while the container is alive.

PR #1749, opened 2023-08-15
combine 2 account.cpp fns into 1

christopherlam:
I figure that Account.cpp returning a `char*` is preferable to returning a `GList*`. These two functions are complementary and could be combined.

### DIFF
--- a/gnucash/gnome-utils/dialog-preferences.c
+++ b/gnucash/gnome-utils/dialog-preferences.c
@@ -123,22 +123,13 @@ GSList *add_ins = NULL;
 static gchar *gnc_account_separator_is_valid (const gchar *separator,
                                               gchar **normalized_separator)
 {
-    QofBook *book;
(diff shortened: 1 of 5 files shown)


jralls:
Bad idea because it turns a reasonably testable design into an untestable one: Testing the result string like that is fragile because it requires the string to be the same in two different files and because it will work only in the "C" locale. It's also not really what needs testing, which is the algo that assembles the list.

But the test string does make it apparent that using the separator to separate the account names that include the separator is a bad idea: `":foo:bar:waldo:pepper"` Which `:`s are in what names?  Newlines would be better.

Finally, it's a waste of your time: The function runs every session start, but seldom generates an error message and even more seldom is there more than one account. It might not even be necessary anymore: @Bob-IT did a lot of work a few years ago to replace identifying accounts by name with identifying them by GUID.

PR #1751, opened 2023-08-17
Iter to stl

christopherlam:
More iter changes.

- to fully be compatible with `std::find_if` and friends, the `GncTreeIter` seems to need iterator traits inherited from `std::iterator`. See the class definition
- `GncTreeContainer` adds an `append()` methods which adds a new `GtkTreeIter` and returns its `GncTreeIter`.
- `GncTreeIter` adds a `get_gtk_tree_iter` to return the `GtkTreeIter` to be useful when `find_if` returns the `GncTreeIter`. This creates another method to access the GtkTreeIter; one from the GncTreeIter and another from GncTreeData. The benefit is this allows:
```cpp
 auto iter = std::find_if (...);
 gtk_list_store_set (store, &iter.get_gtk_tree_iter(), ...)
```
instead of the uglier
```cpp
 auto iter = std::find_if (...);
 gtk_list_store_set (store, &(*iter).get_iter(), ...)
```
- Adds an example `std::find_if` use in `import-match-picker.cpp`.

### DIFF
--- a/gnucash/gnome-utils/gnc-tree-container.hpp
+++ b/gnucash/gnome-utils/gnc-tree-container.hpp
@@ -74,6 +74,12 @@ template <typename ModelType>
 class GncTreeIter
 {
 public:
+    using iterator_category = std::forward_iterator_tag;
+    using value_type = ModelType;
+    using difference_type = ModelType;
+    using pointer = ModelType*;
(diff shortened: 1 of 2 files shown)


christopherlam:
Closed because although this compiles, it segfaults when running.

PR #1752, opened 2023-08-18
Bug 798995 - Keystrokes ignored during ledger entry

Bob-IT:
Some themes add additional pixels to list rows and dialogs, in this case with 'TraditionalOK' 2 pixels are added to each entry in the list and 2 to the popup. This make the calculation for the pop up height wrong and so when compared to the allocation height it is different so tries to resolve by re-popping the list and this causes an endless loop.

To fix this, use the calculated height which works for Adwaita but if this is different to the popup allocation height use that.

I have tested this on the reported OS, Fedora38 with the reported 'TraditionalOK' theme and also on an Xfce VM with the 'Clearlooks-Phenix' theme. Also built and tested on Windows and all seems to be OK.

### DIFF
--- a/gnucash/register/register-gnome/combocell-gnome.c
+++ b/gnucash/register/register-gnome/combocell-gnome.c
@@ -920,29 +920,20 @@ gnc_combo_cell_gui_move (BasicCell* bcell)
 }
 
 static int
-popup_get_height (G_GNUC_UNUSED GtkWidget* widget,
+popup_get_height (GtkWidget* widget,
                   int space_available,
-                  int row_height,
+                  G_GNUC_UNUSED int row_height,
                   gpointer user_data)
 {
     PopBox* box = user_data;
     GtkScrolledWindow* scrollwin = GNC_ITEM_LIST(widget)->scrollwin;
-    GtkWidget *hsbar = gtk_scrolled_window_get_hscrollbar (scrollwin);
-    GtkStyleContext *context = gtk_widget_get_style_context (hsbar);
-    /* Note: gtk_scrolled_window_get_overlay_scrolling (scrollwin) always returns
-       TRUE so look for style class "overlay-indicator" on the scrollbar. */
-    gboolean overlay = gtk_style_context_has_class (context, "overlay-indicator");
-    int count, height;
+    int height;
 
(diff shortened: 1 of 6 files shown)


bparees:
can confirm, this appears to resolve the issue i was seeing when using `TraditionalOk` with `Mate` on `Fedora38`

thank you!

gnucash/register/register-gnome/completioncell-gnome.c
```diff
@@ -926,6 +927,10 @@ popup_get_height (G_GNUC_UNUSED GtkWidget* widget,
         height = height + minh;
     }
 
+    // if popup_allocation_height set use that
+    if (box->item_edit->popup_allocation_height != -1)
+        height = box->item_edit->popup_allocation_height;
+
```
jralls: Same as above, allocation first and count * cell height as a fallback.

gnucash/register/register-gnome/combocell-gnome.c
```diff
@@ -944,6 +944,10 @@ popup_get_height (G_GNUC_UNUSED GtkWidget* widget,
         height = height + minh;
     }
 
+    // if popup_allocation_height set use that
+    if (box->item_edit->popup_allocation_height != -1)
+        height = box->item_edit->popup_allocation_height;
+
```
jralls: ISTM this should go first with the `count * cell_height` as an `else` fallback. Is this partly to figure out how big a window to ask for or otherwise set `height` before the popup is realized and has an allocation?

Bob-IT:
I have added calculating the popup height to `gnucash-item-list` as the same calculation was being done in two places and added calling it from the `else` condition.

> Is this partly to figure out how big a window to ask for or otherwise set `height` before the popup is realized and has an allocation?

The calculation of height is correct for `Adwaita` but for other themes it may be correct or an approximation. The height is used to decide on the placement of the popup which is based on the greater space above or below the current position. Setting the height is needed when the list is large to restrict the popup to the register size.

I am not sure I can see a better way but will have a think.

ipilcher:
This also fixes the issue for me.  (Seen with the Breeze theme under KDE/Plasma/X11.)

jralls:
Thanks!

PR #1756, opened 2023-08-27
[gnc-engine-guile] exports gnc_string_to_numeric

christopherlam:
which can return #f on error, or a number on successful parse.

- [x] doesn't seem to handle "integer fraction" form eg `10 3/4`.

### DIFF
--- a/libgnucash/backend/xml/io-gncxml-v1.cpp
+++ b/libgnucash/backend/xml/io-gncxml-v1.cpp
@@ -472,6 +472,13 @@ simple_kvp_value_parser_new (sixtp_end_handler end_handler)
 
  */
 
+static gboolean
+string_to_gnc_numeric(const gchar* str, gnc_numeric *n)
+{
+    *n = gnc_numeric_from_string (str);
+    return (!gnc_numeric_check (*n));
+}
+
 /* ------------------------------------------------------------ */
 /* generic type copnversion for kvp types */
 #define KVP_CVT_VALUE(TYPE)                 \
--- a/libgnucash/engine/gnc-numeric.cpp
+++ b/libgnucash/engine/gnc-numeric.cpp
@@ -208,21 +208,24 @@ numeric_from_scientific_match(smatch &m)
 }
 
 GncNumeric::GncNumeric(const std::string &str, bool autoround) {
-    static const std::string numer_frag("(-?[0-9]*)");
-    static const std::string denom_frag("([0-9]+)");
+    static const std::string maybe_sign ("(-?)");
+    static const std::string opt_signed_int("(-?[0-9]*)");
+    static const std::string unsigned_int("([0-9]+)");
     static const std::string hex_frag("(0[xX][A-Fa-f0-9]+)");
     static const std::string slash("[ \\t]*/[ \\t]*");
+    static const std::string whitespace("[ \\t]+");
     /* The llvm standard C++ library refused to recognize the - in the
-     * numer_frag pattern with the default ECMAScript syntax so we use the
+     * opt_signed_int pattern with the default ECMAScript syntax so we use the
      * awk syntax.
      */
-    static const regex numeral(numer_frag);
+    static const regex numeral(opt_signed_int);
     static const regex hex(hex_frag);
-    static const regex numeral_rational(numer_frag + slash + denom_frag);
+    static const regex numeral_rational(opt_signed_int + slash + unsigned_int);
+    static const regex integer_and_fraction(maybe_sign + unsigned_int + whitespace + unsigned_int + slash + unsigned_int);
     static const regex hex_rational(hex_frag + slash + hex_frag);
-    static const regex hex_over_num(hex_frag + slash + denom_frag);
-    static const regex num_over_hex(numer_frag + slash + hex_frag);
-    static const regex decimal(numer_frag + "[.,]" + denom_frag);
+    static const regex hex_over_num(hex_frag + slash + unsigned_int);
+    static const regex num_over_hex(opt_signed_int + slash + hex_frag);
+    static const regex decimal(opt_signed_int + "[.,]" + unsigned_int);
     static const regex scientific("(?:(-?[0-9]+[.,]?)|(-?[0-9]*)[.,]([0-9]+))[Ee](-?[0-9]+)");
     static const regex has_hex_prefix(".*0[xX]$");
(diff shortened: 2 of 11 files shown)


bindings/engine.i
```diff
@@ -121,6 +121,8 @@ functions currently used in guile, but not all the functions that are
 wrapped.  So, we should contract the interface to wrap only the used
 functions. */
 
+gnc_numeric gnc_string_to_number (const char * in_str);
```
jralls: Not requred. Including gnc-engine-guile.hpp imports all of its declarations.

bindings/guile/gnc-engine-guile.cpp
```diff
@@ -1909,3 +1909,13 @@ gnc_parse_time_to_time64 (const gchar *s, const gchar *format)
...
 
+
+gnc_numeric
+gnc_string_to_number (const char * in_str)
+{
+    gnc_numeric rv;
+    if (!string_to_gnc_numeric (in_str, &rv))
+        rv = gnc_numeric_create (1, 0);
+    return rv;
+}
```
jralls: Adding a `%typemap(argout) gnc_numeric` to common/base-typemaps.i would allow `string_to_gnc_numeric` to be wrapped correctly and make this unnecessary.

libgnucash/engine/gnc-numeric.cpp
```diff
@@ -208,17 +208,20 @@ numeric_from_scientific_match(smatch &m)
 }
 
 GncNumeric::GncNumeric(const std::string &str, bool autoround) {
+    static const std::string integer("(-?[0-9]+)");
     static const std::string numer_frag("(-?[0-9]*)");
     static const std::string denom_frag("([0-9]+)");
```
jralls: `numer_frag` and `denom_frag` need new names because you're using `denom_frag` as a numerator.
How about `signed_int`, `opt_signed_int` and `unsigned_int`?

libgnucash/engine/gnc-numeric.cpp
```diff
@@ -256,6 +259,21 @@ GncNumeric::GncNumeric(const std::string &str, bool autoround) {
...
     }
+    if (regex_search(str, m, integer_and_fraction))
+    {
+        auto integer{stoll(m[1].str())};
+        if (integer == 0)
+            throw std::invalid_argument ("GncNumeric with integer and fraction, integer cannot be zero");
+        auto frac_num{stoll(m[2].str())};
+        auto frac_denom{stoll(m[3].str())};
+        if (frac_num >= frac_denom)
+            throw std::invalid_argument ("GncNumeric with integer and fraction, fraction should be less than 1");
```
jralls: Why does that matter?

libgnucash/engine/gnc-numeric.cpp
```diff
@@ -256,6 +259,21 @@ GncNumeric::GncNumeric(const std::string &str, bool autoround) {
         m_den = n.denom();
         return;
     }
+    if (regex_search(str, m, integer_and_fraction))
+    {
+        auto integer{stoll(m[1].str())};
+        if (integer == 0)
+            throw std::invalid_argument ("GncNumeric with integer and fraction, integer cannot be zero");
```
jralls: Why not?

jralls (changes requested):

bindings/guile/gnc-engine-guile.cpp
jralls: Alternatively `string_to_gnc_numeric` is used in only 6 places. You could rewrite it to `gnc_numeric gnc_numeric_from_string(const char*)` returning `GNC_ERROR_ARG` on failure and change those 6 uses.

libgnucash/engine/gnc-numeric.cpp
christopherlam: because it seems ridiculous to allow such weirdness like `0 2/3` or `1 8/5`?

jralls: That's being arbitrary. They're perfectly legitimate numbers and while a human wouldn't write them a lazily-written program might.

christopherlam: Perhaps... Then the parser will need to handle such cases like `-0 5/3` being -5/3. The regexes (regices?) will need more upgrades.

libgnucash/backend/xml/io-gncxml-v1.cpp
```diff
@@ -472,6 +472,13 @@ simple_kvp_value_parser_new (sixtp_end_handler end_handler)
 
  */
 
+static gboolean
+string_to_gnc_numeric(const gchar* str, gnc_numeric *n)
+{
+    *n = gnc_numeric_from_string (str);
+    return (!gnc_numeric_check (*n));
+}
+
```
jralls, line 481: Huh? It's not even used in io-gncxml-v1.

libgnucash/engine/gnc-numeric.cpp
```diff
@@ -1321,19 +1321,18 @@ gnc_num_dbg_to_string(gnc_numeric n)
     return p;
 }
 
-gboolean
-string_to_gnc_numeric(const gchar* str, gnc_numeric *n)
+gnc_numeric
+gnc_numeric_from_string (const gchar* str)
 {
```
jralls, line 1319: Maybe start with 
```
if (!str)
    return gnc_numeric_error(GNC_ERROR_ARG);
```
then in gnc-exp-parser you can say
```
 str_value = g_key_file_get_string(key_file, GEP_GROUP_NAME, *key, NULL);
 value = gnc_numeric_from_string (str_value);
 if (!gnc_numeric_check (value))
      gnc_exp_parser_set_value (*key, gnc_numeric_reduce (value));
```

libgnucash/backend/xml/io-gncxml-v1.cpp
christopherlam: It's used by the macro immediately afterwards

jralls: Which is easily enough dealt with, but I'll accept that as out of scope for this PR.

christopherlam:
Ready. It would be good for a fresh pair of eyes on all these pointer args.

jralls (approved):

jralls:
LGTM.

PR #1757, opened 2023-08-29
Remove reference to unused assistant subfield.

jwhite66:
This did not cause trouble because the only use of this was in a call to gnc_warning_dialog, which protects against a NULL pointer.


christopherlam:
Thanks!

PR #1758, opened 2023-09-01
Revise QIF Import to enable the back button.

jwhite66:
For bug 728875.

The current logic has each 'prepare' function jump us around in the assistant flow in order to skip pages that aren't needed.  But that logic only allows forward movement, effectively breaking the back button.

This patch introduces a custom 'next page' function, which GtkAssistant supports, along with a shift in the paradigm.  Instead of the prepare function jumping around, there is now a deliberate 'should I skip?' function that is used to avoid displaying a page.

This does not make the flow perfect; unexpected exceptions still end the process (see bug 698804) and going back with loaded files feels a bit clunky.

### DIFF
--- a/gnucash/import-export/qif-imp/assistant-qif-import.c
+++ b/gnucash/import-export/qif-imp/assistant-qif-import.c
@@ -217,6 +217,9 @@ typedef struct _qifnotebookpage QIFCommNotebookPage;
 static void gnc_ui_qif_import_assistant_destroy (GtkWidget *object, gpointer user_data);
 static void gnc_ui_qif_import_assistant_close_handler (gpointer user_data);
 
+static gboolean gnc_ui_qif_import_assistant_skip_page (GtkAssistant *assistant, GtkWidget *page, QIFImportWindow *wind);
+static int gnc_ui_qif_import_assistant_page_forward (int current_page, gpointer data);
+
 void gnc_ui_qif_import_cancel_cb (GtkAssistant *gtkassistant, gpointer user_data);
 void gnc_ui_qif_import_prepare_cb (GtkAssistant *assistant, GtkWidget *page, gpointer user_data);
 void gnc_ui_qif_import_finish_cb (GtkAssistant *gtkassistant, gpointer user_data);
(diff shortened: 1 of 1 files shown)


gnucash/import-export/qif-imp/assistant-qif-import.c
```diff
@@ -217,6 +217,9 @@ typedef struct _qifnotebookpage QIFCommNotebookPage;
 static void gnc_ui_qif_import_assistant_destroy (GtkWidget *object, gpointer user_data);
 static void gnc_ui_qif_import_assistant_close_handler (gpointer user_data);
 
+static gboolean gnc_ui_skip_page (GtkAssistant *assistant, GtkWidget *page, QIFImportWindow *wind);
+static int gnc_ui_forward (int current_page, gpointer data);
```
jralls: Those names imply that the functions should be global scope and implemented in gnucash/gnome-utils/gnc-ui.c. We prefer file-static functions to not have a `gnc_foo` prefix, so `assistant_skip_page` and `assistant_page_forward` would be best current practice. But given that all of the other functions in this file are prefixed `gnc_ui_qif_import` it's better to be consistent with that, so please rename these to `gnc_ui_qif_import_assistant_skip_page` and `gnc_ui_qif_import_assistant_page_forward`.

gnucash/import-export/qif-imp/assistant-qif-import.c
```diff
@@ -231,10 +234,11 @@ void gnc_ui_qif_import_load_progress_prepare (GtkAssistant *assistant, gpointer
 void gnc_ui_qif_import_load_progress_pause_cb (GtkButton *button, gpointer user_data);
 void gnc_ui_qif_import_load_progress_start_cb (GtkButton * button, gpointer user_data);
 
-void gnc_ui_qif_import_date_format_prepare (GtkAssistant *assistant, gpointer user_data);
+static gboolean gnc_ui_skip_qif_import_date_format (GtkAssistant *assistant, QIFImportWindow *wind);
```
jralls: Appropos the previous comment `gnc_ui_qif_import` is a prefix, so this should be named `gnc_ui_qif_import_skip_date_format_page`. Please fix the similarly named functions likewise.

gnucash/import-export/qif-imp/assistant-qif-import.c
```diff
@@ -246,19 +250,25 @@ static void update_file_page (QIFImportWindow * wind);
 
 void gnc_ui_qif_import_account_match_prepare (GtkAssistant *assistant, gpointer user_data);
 void gnc_ui_qif_import_account_doc_prepare (GtkAssistant *assistant, gpointer user_data);
+static gboolean gnc_ui_skip_qif_import_account_doc (QIFImportWindow *wind);
 void gnc_ui_qif_import_account_rematch_cb (GtkButton *button, gpointer user_data);
 
-void gnc_ui_qif_import_catagory_match_prepare (GtkAssistant *assistant, gpointer user_data);
-void gnc_ui_qif_import_catagory_doc_prepare (GtkAssistant *assistant, gpointer user_data);
+void gnc_ui_qif_import_category_match_prepare (GtkAssistant *assistant, gpointer user_data);
```
jralls, line 256: Please put spelling fixes in a separate commit.

jralls (changes requested):
Looks OK to me aside from the function naming. I don't like the `pagename` comparison but I don't see a better way given the limitations of GtkAssistant.

jralls (approved):

jralls:
I failed to notice until the last minute that this is a bug fix. We use the number and summary from those as the commit summary to make them easier to note in the release notes. I modified the second commit accordingly.

Thanks!

PR #1759, opened 2023-09-01
More stock assistant

jralls:
Fixes the bugs and issues, almost finishes the reorganization, adds Doxygen documentation.

### DIFF
--- a/gnucash/gnome/assistant-stock-transaction.cpp
+++ b/gnucash/gnome/assistant-stock-transaction.cpp
@@ -618,8 +618,11 @@ StockTransactionEntry::set_fieldmask(FieldMask mask)
 const char *
 StockTransactionEntry::print_account() const
 {
+    auto acct_required = m_enabled &&
+        !(m_allow_zero && (gnc_numeric_zero_p(m_value) ||
+                           gnc_numeric_check(m_value)));
     return m_account ? xaccAccountGetName(m_account) :
-        m_enabled && !gnc_numeric_zero_p(m_value) ? _("missing") : "";
+        acct_required ? _("missing") : "";
 }
 
 void
--- a/gnucash/gnome/assistant-stock-transaction.cpp
+++ b/gnucash/gnome/assistant-stock-transaction.cpp
@@ -404,7 +404,7 @@ static const TxnTypeVec short_types
         N_("Company issues additional units, thereby reducing the stock price by a divisor, while keeping the total monetary value of the overall investment constant.")
     },
     {
-        FieldMask::DISABLED | FieldMask::ENABLED_DEBIT | FieldMask::INPUT_NEW_BALANCE,          // stock_amt
+        FieldMask::DISABLED | FieldMask::AMOUNT_DEBIT | FieldMask::INPUT_NEW_BALANCE,          // stock_amt
         FieldMask::ENABLED_CREDIT | FieldMask::ALLOW_ZERO,          // cash_amt
         FieldMask::ENABLED_DEBIT | FieldMask::ALLOW_ZERO | FieldMask::CAPITALIZE_DEFAULT,          // fees_amt
         FieldMask::DISABLED,               // dividend_amt
@@ -543,6 +543,7 @@ struct StockTransactionEntry
     virtual const char* print_value(GNCPrintAmountInfo info);
     virtual const char* print_amount(gnc_numeric amt);
     virtual gnc_numeric calculate_price(bool) { return gnc_numeric_error(GNC_ERROR_ARG); }
+    virtual bool has_amount() { return false; }
 };
 
 using StockTransactionEntryPtr = std::unique_ptr<StockTransactionEntry>;
@@ -599,14 +600,14 @@ StockTransactionEntry::set_value(gnc_numeric amount, const char* page, Logger& l
         return;
     }
 
-    m_value = m_debit_side ? amount : gnc_numeric_neg (amount);
+    m_value = amount;
 }
 
 const char *
 StockTransactionEntry::print_value(GNCPrintAmountInfo pinfo)
 {
-    if (gnc_numeric_check(m_value) ||
-        (gnc_numeric_zero_p(m_value) && !m_allow_zero))
+    if ((gnc_numeric_check(m_value) || gnc_numeric_zero_p(m_value))
+        && !m_allow_zero)
         return _("missing");
     return xaccPrintAmount(m_value, pinfo);
 }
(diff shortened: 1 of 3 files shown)


christopherlam:
not running well yet...

jralls:
@christopherlam 
>not running well yet...

Can you be more specific, starting with what OS?

christopherlam:
Linux as usual

From stable: `git pull --no-rebase https://github.com/jralls/gnucash.git more-stock-assistant` and recompile, no issues. Then:

```
* 12:06:01 ERROR <gnc.gui> void StockAssistantController::prepare(GtkAssistant*, GtkWidget*): assertion 'm_model->txn_type_valid()' failed
```

jralls:
@christopherlam Got it, thanks. Should be fixed now.

christopherlam:
Much better. I played a bit, there _should_ be an error if eg. a dividend type transaction, and the dividend account is missing. Lots of the following:

```
* 19:40:57 ERROR <gnc.engine> gnc_commodity* gnc_account_get_currency_or_parent(const Account*): assertion 'account' failed
* 19:41:55 ERROR <gnc.engine> gnc_commodity* gnc_account_get_currency_or_parent(const Account*): assertion 'account' failed
* 19:42:39 ERROR <gnc.engine> gnc_commodity* gnc_account_get_currency_or_parent(const Account*): assertion 'account' failed
```

jralls:
OK, I added an error for there not being an account when there should be, to display the stock account on the fees line when capitalizing fees, and an if to not try to print the amount if there's no account to get the currency from.

The macOS test is failing because of a Sourceforge problem. We can ignore that for now.

christopherlam:
ok the GNCAccountSels don't seem to work now.

christopherlam:
> ok the GNCAccountSels don't seem to work now.

Ok further clarification: I used 797796's datafile. I tried to recreate the dividend transaction. I couldn't, because 797796.gnucash's SPY's parent is in CAD and the assistant will set the txn_currency to be CAD. Conclusion: given the assitant cannot create a txn whose currency isn't the stock's parent currency, I think this is acceptable.

Here's a dividend issue: creating a simple dividend should add a zero-amt, zero-value split in the stock acct, and it doesn't.

christopherlam:
Ok now this branch would want `gnc_account_sel_set_default_new_commodity (GNC_ACCOUNT_SEL(m_selector), currency);` somewhere

jralls:
@christopherlam Added the new account default currency, the tracking split for dividends, the multicurrency note to the intro, and fixed showing a missing account for fees when the amount was GNC_ERROR_ARG.

christopherlam:
It seems to work well so far.

A small comment: this being a stock assistant, it seems right to assume that *all* txn_types _will_ insert a stock-acct split. Therefore in my design I didn't specify a separate `MARKER_SPLIT` flag.

Moreover, `FieldMask::DISABLED` was intended to fully denote fields which should skip the page i.e. in my original design, the dividend page should skip both stock_amt and stock_val. In this branch, the "Stock Split" type will define stock_amt as `FieldMask::DISABLED | FieldMask::AMOUNT_DEBIT | FieldMask::INPUT_NEW_BALANCE` which is identical to `FieldMask::AMOUNT_DEBIT | FieldMask::INPUT_NEW_BALANCE`, i.e. it seems odd.

gnucash/gnome/assistant-stock-transaction.cpp
```diff
@@ -292,7 +261,10 @@ reinvested must be subsequently recorded as a regular stock purchase.")
         // Translators: this is a stock transaction describing a stock
         // split
         N_("Stock split"),
-        N_("Company issues additional units, thereby reducing the stock price by a divisor, while keeping the total monetary value of the overall investment constant.")
+        N_("Company issues additional units, thereby reducing the stock price by a divisor "
+           ", while keeping the total monetary value of the overall investment constant. "
+           "\n\nIf the reverse split results in a cash in lieu for remainder units, please "
```
christopherlam:
```suggestion
           "\n\nIf the split results in a cash in lieu for remainder units, please "
```

gnucash/gnome/assistant-stock-transaction.cpp
```diff
@@ -401,58 +387,221 @@ static const TxnTypeVec short_types
         // Translators: this is a stock transaction describing a stock
         // split when shorting stock
         N_("Stock split"),
-        N_("Company issues additional units, thereby reducing the stock price by a divisor, while keeping the total monetary value of the overall investment constant.")
+        N_("Company issues additional units, thereby reducing the stock price by a divisor, "
+           "while keeping the total monetary value of the overall investment constant. "
+           "\n\nIf the reverse split results in a cash in lieu for remainder units, please "
```
christopherlam:
```suggestion
           "\n\nIf the split results in a cash in lieu for remainder units, please "
```

christopherlam (approved):
from my testing today everything seems ok. comments about marker_split and FieldMask::DISABLED remains. minor explanation fix.

jralls:
> A small comment: this being a stock assistant, it seems right to assume that all txn_types will insert a stock-acct split. Therefore in my design I didn't specify a separate MARKER_SPLIT flag.

Yes, and that was bad because it created a hidden special case. The MARKER_SPLIT flag makes it clear at the top level that a Dividend transaction will create a marker split, that a buy will generate an AMOUNT_DEBIT and a VALUE_DEBIT while a stock split will generate only an AMOUNT_DEBIT.

> Moreover, FieldMask::DISABLED was intended to fully denote fields which should skip the page i.e. in my original design, the dividend page should skip both stock_amt and stock_val. In this branch, the "Stock Split" type will define stock_amt as FieldMask::DISABLED | FieldMask::AMOUNT_DEBIT | FieldMask::INPUT_NEW_BALANCE which is identical to FieldMask::AMOUNT_DEBIT | FieldMask::INPUT_NEW_BALANCE, i.e. it seems odd.

Correct, FieldMask::DISABLED can't be meaningfully combined with anything else because it sets no bit. That makes it useless to the program, but it's still useful as a placeholder in the FieldMask declarations. I agree that it's not helpful on the stock entry because there's always something else. I'll remove it tomorrow.

christopherlam:
Ok one more modification: `gnc_account_sel_set_new_account_modal (GNC_ACCOUNT_SEL(m_selector), true);` so that the "New account" button, having created the new account, will select it in the GNCAccountSel.

gnucash/gnome/assistant-stock-transaction.cpp
jralls: Good catch, thanks.

jralls:
> Ok one more modification: 

Done.

PR #1762, opened 2023-09-07
GNCAccountSel gets new account gets default commodity

christopherlam:
preparation for, and illustration of use to fix https://github.com/Gnucash/gnucash/pull/1759#issuecomment-1708431612

### DIFF
--- a/gnucash/gnome-utils/dialog-account.c
+++ b/gnucash/gnome-utils/dialog-account.c
@@ -2143,6 +2143,14 @@ gnc_ui_edit_account_window (GtkWindow *parent, Account *account)
     gtk_window_present (GTK_WINDOW(aw->dialog));
 }
 
+void
+gnc_ui_new_account_with_types_and_commodity (GtkWindow *parent, QofBook *book, GList *valid_types,
+                                             gnc_commodity *default_commodity)
+{
+    gnc_ui_new_account_window_internal (parent, book, NULL, NULL,
+                                        valid_types, default_commodity, FALSE);
+}
+
 /*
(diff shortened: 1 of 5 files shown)


jralls:
I guess this is based on [this IRC exchange](https://code.gnucash.org/logs/2023/09/06.html#T10:00:57). It doesn't limit anything, though, it just suggests. The user can override the suggestion in the account dialog. It does nothing at all about selecting an existing account in a different currency from the stock's imputed currency.

ISTM that instead the verification step should check that all of the accounts use the same currency and log an error if not.

christopherlam:
> the verification step

do you mean the stock assistant's final page? no it's impossible to select an account whose currency doesn't match the stock's parent currency - see `gnc_account_sel_set_acct_filters(GNC_ACCOUNT_SEL(m_selector), acct_list, curr_list);` with `curr_list` has only 1 element, the txn currency.

imho offering a "New account" whose default currency == stock's parent currency would be a sensible default. if the user wishes to create a different currency account, they're free; the new account would still not be selectable in the GNCAccountSel.

jralls:
Ah,  I see that now. Thanks.

Yes, the default seems sensible indeed, though it doesn't help the user who's confused about why they can't find the account they want, just as you were. Maybe the StockTransactionController should check the supplied account's currency against the root currency and add a warning to the intro page if they're different.

christopherlam:
If this is ok then I'll merge this branch (minus the stock assistant commit) and you can amend your branch appropriately. A warning on the intro page may be offered too.

PR #1773, opened 2023-09-16
Test qof plug leaks

christopherlam:
keeps valgrind happy but:
1. test-qofbook.c -- shouldn't `qof_book_destroy` take care of destroying root?
2. test-qofsession.cpp -- works but of doubtful elegance.

### DIFF
--- a/libgnucash/engine/test/test-qofbook.c
+++ b/libgnucash/engine/test/test-qofbook.c
@@ -64,6 +64,9 @@ setup( Fixture *fixture, gconstpointer pData )
 static void
 teardown( Fixture *fixture, gconstpointer pData )
 {
+    Account *root = gnc_book_get_root_account (fixture->book);
+    xaccAccountBeginEdit (root);
+    xaccAccountDestroy (root);
     qof_book_destroy( fixture->book );
 }
 
--- a/libgnucash/engine/test/test-qofsession.cpp
+++ b/libgnucash/engine/test/test-qofsession.cpp
@@ -36,8 +36,20 @@ static bool sync_called {false};
 static bool load_error {true};
 static bool data_loaded {false};
 
+struct DestroyAccount
+{
+    void operator()(Account *acct)
+    {
+        xaccAccountBeginEdit (acct);
+        xaccAccountDestroy (acct);
+    }
+};
+
+using AccountPtr = std::unique_ptr<Account, DestroyAccount>;
+
 class QofSessionMockBackend : public QofBackend
 {
+    AccountPtr m_root;
 public:
     QofSessionMockBackend() = default;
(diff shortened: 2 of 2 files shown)


christopherlam:
>     1. test-qofbook.c -- shouldn't `qof_book_destroy` take care of destroying root?

this issue is pervasive across many tests-- although qof_book_destroy is called in tests, accounts/splits/commodities etc are somehow surviving. e.g. `valgrind --leak-check=full build/bin/test-qofsession` is the simplest test which illustrates this.

jralls:
> test-qofbook.c -- shouldn't qof_book_destroy take care of destroying root?

Probably, but making it do so will touch a lot of code and not just in tests.

> test-qofsession.cpp -- works but of doubtful elegance.

Because of needing the custom deleter? Avoiding that requires Account to be a C++ class with a destructor. That's a big job too.

christopherlam:
and this last one works; remove `+` and the lambda can't be used....

jralls:
> and this last one works; remove + and the lambda can't be used....

That's an ugly hack that doesn't really have a point. Saying `static bool is_destroying(gpointer key, EventInfo* change, gpointer user_data)` is lighter weight (it doesn't need to create a function object under the hood), clearer, and doesn't risk somebody in the future breaking it by removing the "obviously unnecessary" `+`.

christopherlam:
thought so...

jralls:
That last force-push isn't right, you lost two of the fixes and left one of them spread across two commits.

christopherlam:
ok the lambda->function pointer one got pushed as c074b82224.

jralls:
> ok the lambda->function pointer one got pushed as https://github.com/Gnucash/gnucash/commit/c074b8222426012841a3326b956d58428d09d54e.

That's this PR's remaining commit.  The lambda->function commit is https://github.com/Gnucash/gnucash/commit/41167e3270c1380c3d84c50ce7675359cc70ba9d, and the other one is https://github.com/Gnucash/gnucash/commit/09474bb251c00de68aea81f8ab7010da2fff2b21. So what's keeping you from pushing this one?

christopherlam:
Hmm just pondering if it's better to clean up accounts manually on teardown, or whether the test code may clean up somewhere else centrally.

jralls:
How much work are you up for? 

For a really good test program, there should be a fixture class with all of the setup and teardown code. QofSessionMockBackend should be in its own file in libgnucash/engine/mocks and use GMock macros (, and all of the classes except QofSession (the class being tested) should also be mocks from that directory.

If that's too much you could settle for just moving the setup and teardown into a fixture class. See the Test Fixtures section in [the Googletest primer](http://google.github.io/googletest/primer.html).

PR #1775, opened 2023-09-18
[assistant-stock-transaction] check account irrespective of numeric field

christopherlam:
checking account should occur even if numeric field is invalid.

further issues:
1. run an assistant session, with valid amounts/accounts. the final page should show 'ready to create'. press back and erase some numeric fields (i.e. the GtkEntry is now blank). complete the assistant again. the numeric field is still retaining the old number.
2. (just reverted as 095d32f73b) a fees page amount being empty shouldn't lead to an exception. it should undergo validation in the final page, erroring out if appropriate. In my original design, fees are always optional. empty = no fee deducted, positive number = fee deducted, negative number = invalid; non-number = invalid (? unsure).

### DIFF
--- a/gnucash/gnome/assistant-stock-transaction.cpp
+++ b/gnucash/gnome/assistant-stock-transaction.cpp
@@ -656,6 +656,8 @@ StockTransactionEntry::validate_amount(Logger& logger) const
         g_free (buf);
     };
 
+    if (!gnc_numeric_zero_p(m_value) && !m_account)
+        add_error(N_("The %s amount has no associated account."), m_action);
 
     if (gnc_numeric_check (m_value))
     {
@@ -669,9 +671,6 @@ StockTransactionEntry::validate_amount(Logger& logger) const
 
     if (!m_allow_zero && !gnc_numeric_positive_p (m_value))
         add_error (N_("Amount for %s must be positive."), m_action);
-
-    if (!gnc_numeric_zero_p(m_value) && !m_account)
-        add_error(N_("The %s amount has no associated account."), m_action);
 }
 
 const char *
@@ -1004,6 +1003,9 @@ StockTransactionFeesEntry::validate_amount(Logger& logger) const
     };
 
 
+    if (!gnc_numeric_zero_p(m_value) && !m_account && !m_capitalize)
+        add_error(N_("The %s amount has no associated account."), m_action);
+
     if (gnc_numeric_check (m_value))
     {
         if (!m_allow_zero)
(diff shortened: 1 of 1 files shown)


jralls:
> checking account should occur even if numeric field is invalid.

Why?

And if so then it should be in a separate function `check_account()`.

> run an assistant session, with valid amounts/accounts. the final page should show 'ready to create'. press back and erase some numeric fields (i.e. the GtkEntry is now blank). complete the assistant again. the numeric field is still retaining the old number.

OK.

>(just reverted as https://github.com/Gnucash/gnucash/commit/095d32f73bede5c9e8114a8d60d890a0c5a3bed2) a fees page amount being empty shouldn't lead to an exception. it should undergo validation in the final page, erroring out if appropriate. In my original design, fees are always optional. empty = no fee deducted, positive number = fee deducted, negative number = invalid; non-number = invalid (? unsure).

Where did you get an exception? I haven't seen that. Likewise, what does "non number" mean? Some error other than `GNC_ERROR_ARG`?  (`NaN` is a concept from IEEE floating point, the result of a division by zero. There's no division going on here and `gnc_numeric` uses a 0 denominator to encode error values.)

jralls:
Fixed 1 & 2 after finding the answer about the exception [on IRC](https://code.gnucash.org/logs/2023/09/17.html#T22:06:51).

The optimizer in macOS's clang causes `g_hash_table_find` to return a pointer even if `is_destroy` returns false. That in turn causes premature destruction of the assistant (while it's still displayed!) when a New Account dialog closes when trying to create a new account, promptly crashing on a use-after-free. I reverted 41167e and added a `g_list_free` to fix the leak.

>> checking account should occur even if numeric field is invalid.

>Why?

Make that "rubbish". That's only true for a tracer split in the stock account, and the stock account is guaranteed to be valid. Meanwhile enforcing that will cause an ignored fees page--one with neither account nor value entered--to fail validation.

christopherlam:
> The optimizer in macOS's clang causes g_hash_table_find to return a pointer even if is_destroy returns false.

Very very naughty

christopherlam:
> > checking account should occur even if numeric field is invalid.
> 
> Why?

The idea is that if the Fees page Account is empty, and Amount is non-number eg "tada", the Summary should show both "fees account is missing" and "fees amount is invalid" or something. In the stable branch, the account validation was skipped because amount was non-numeric.

jralls:
And to distinguish between a non-number string and an empty string, both of which will return `GNC_ERROR_ARG` when converted to `gnc_numeric` you have to separately record that there's a non-empty string in the entry so that you can raise the "missing" and "invalid" messages. What's the point of that vs. simply assuming that there's no fee entry? If the user intended a fee the cash value will reflect it and the transaction won't balance. The Fee amount will be blank. They'll go back and fix the bad input.

PR #1784, opened 2023-09-25
[utest-gnc-pricedb] fix test failure due to callback type mismatch

nevivurn:
I found the test `test_gnc_pricedb_foreach_price` can fail randomly, I believe this is because the callback to `gnc_pricedb_foreach_price` is supposed to return a `gboolean`.

Currently, depending on the compiler, this can randomly exit the iterator early here
https://github.com/Gnucash/gnucash/blob/5.4/libgnucash/engine/gnc-pricedb.cpp#L2633-L2635

### DIFF
--- a/libgnucash/engine/test/utest-gnc-pricedb.c
+++ b/libgnucash/engine/test/utest-gnc-pricedb.c
@@ -1499,14 +1499,17 @@ gboolean
 gnc_pricedb_foreach_price(GNCPriceDB *db,// C: 2 in 2  Local: 6:0:0
 */
 
-static void prepend_price_time64 (GNCPrice *p, GList **lst)
+static gboolean prepend_price_time64 (GNCPrice *p, GList **lst)
 {
     *lst = g_list_prepend (*lst, GUINT_TO_POINTER(gnc_price_get_time64 (p)));
+    return TRUE;
 }
 
-static void inc_counter (GNCPrice *p, guint *count)
+static gboolean inc_counter (GNCPrice *p, guint *count)
 {
+
     (*count)++;
+    return TRUE;
 }
 
 static void


christopherlam:
Interesting... haven't seen it. I'm not sure this is the right solution... `gnc_pricedb_foreach_price` shouldn't test the output from `unstable_price_traversal`, and `g_hash_table_foreach` shouldn't ever return before scanning the whole GHashTable. What compiler and flags do you use?

christopherlam:
> Interesting... haven't seen it. I'm not sure this is the right solution... `gnc_pricedb_foreach_price` shouldn't test the output from `unstable_price_traversal`, and `g_hash_table_foreach` shouldn't ever return before scanning the whole GHashTable. What compiler and flags do you use?

Oops scratch my comment.

christopherlam:
Having reviewed my commits fc21c55a0b and 6b741ddf32 I'm fairly confident I haven't effected this change. It may be the e841ceaaa1 cpp conversion which optimises more aggressively.

I think these `foreach` functions are misnamed... they behave like `foreach_until`. I'd vote to rename them as `foreach_until` instead of amending the tests as you suggest.

Any other option wouldn't be right because [/home/chris/sources/gnucash/stable/libgnucash/backend/xml/gnc-pricedb-xml-v2.cpp:489](https://github.com/Gnucash/gnucash/blob/15193c5ce04e583782b42287fdfb8132577c7048/libgnucash/backend/xml/gnc-pricedb-xml-v2.cpp#L489) relies on the `foreach_until` behaviour. @jralls ?

I'd also add `foreach` functions which wraps the callback, and should never return early.

christopherlam:
Thanks!

PR #1785, opened 2023-09-25
rename gnc_pricedb_foreach_price to gnc_pricedb_foreach_price_until

christopherlam:
and create a `gnc_pricedb_foreach_price` without ability to return early should fix #1784

### DIFF
--- a/libgnucash/backend/sql/gnc-price-sql.cpp
+++ b/libgnucash/backend/sql/gnc-price-sql.cpp
@@ -222,7 +222,7 @@ GncSqlPriceBackend::write (GncSqlBackend* sql_be)
     write_objects_t data{sql_be, true, this};
 
     auto priceDB = gnc_pricedb_get_db (sql_be->book());
-    return gnc_pricedb_foreach_price (priceDB, write_price, &data, TRUE);
+    return gnc_pricedb_foreach_price_while (priceDB, write_price, &data, TRUE);
 }
 
 /* ========================== END OF FILE ===================== */
--- a/libgnucash/backend/sql/gnc-price-sql.cpp
+++ b/libgnucash/backend/sql/gnc-price-sql.cpp
@@ -204,15 +204,15 @@ write_price (GNCPrice* p, gpointer data)
 {
     auto s = reinterpret_cast<write_objects_t*>(data);
 
-    g_return_val_if_fail (p != NULL, FALSE);
-    g_return_val_if_fail (data != NULL, FALSE);
+    g_return_val_if_fail (p != NULL, true);
+    g_return_val_if_fail (data != NULL, true);
 
     if (s->is_ok && gnc_price_get_source (p) != PRICE_SOURCE_TEMP)
     {
         s->commit (QOF_INSTANCE(p));
     }
 
-    return s->is_ok;
+    return !s->is_ok;
 }
 
 bool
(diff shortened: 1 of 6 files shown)


christopherlam:
Addendum; since the iteration continues as long as the function returns `true`, maybe this should be renamed to the more accurate `foreach_while`

jralls:
> and create a gnc_pricedb_foreach_price without ability to return early should fix https://github.com/Gnucash/gnucash/pull/1784

Sort of, because you also need unit tests for the new function whose callbacks will look just like the ones from #1784; that test will also need callbacks that terminate early to test that control path in the function.

christopherlam:
Ok. prefer `foreach_while` or `foreach_until_false`?

jralls:
`foreach_while`.

Although `gnc_account_foreach_until` is suggestive, it runs until its callback returns something non-null and that becomes `gnc_account_foreach_until`'s rv. That's very different semantics.

`gnc_pricedb_foreach_price_until_false` is windy and ugly.

christopherlam:
@jralls this is a bit more complicated than I thought, because the `foreach_while` will do `g_hash_table_foreach` within `g_hash_table_foreach`, and the innermost loop will be `g_list_find_custom`. I think it _would_ be simpler to just create a `foreach_until` whereby the callback output being `false` or `nullptr` would keep searching.

Thus the pricelist would be traversed via `g_hash_table_find`, `g_hash_table_find`, `g_list_find_custom`.

christopherlam:
(in case it's not obvious) with the prior _hash_table_foreach > g_hash_table_foreach > g_list_find_custom, the GHashtables were not returning early, causing the tests to be brittle. This latest iteration now has stable tests.

jralls:
Hmm, no, returning a pointer isn't the right thing to do here.

I think we should back up a bit and think about what those functions are really about. ISTM the account ones are really find functions, with the callback acting as a predicate and the function returning the first `Account*` that matches the predicate. The price function really is a foreach: The intention in the backends is to write all of the prices to the database/DOM. A failure is an error that's supposed to stop the save (I found [an error](https://github.com/Gnucash/gnucash/blob/15193c5ce04e583782b42287fdfb8132577c7048/libgnucash/backend/sql/gnc-sql-backend.cpp#L502) where the SQL backend fails to check the return value).  The right name is really `gnc_pricedbd_foreach_price_unless_fail`, though that's rather cumbersome, and the original semantic of returning `true` for success is correct.

christopherlam:
How about the the view that creating `foreach_until` is _much_ easier because the it mirrors g_hash_table_find? I'd rather push the above and add a `foreach_unless_fail` which wraps and negates the callback output.

jralls:
What has `g_hash_table_find` got to do with anything?

I'd rather accept #1784, fix `GncSqlBackend::Sync` and leave the rest alone until it's refactor into C++ time. It's not like anything else is broken.

christopherlam:
If the callback is `foreach_until_false` which should stop iterating through the hashtables, and the xml/sql pricedb writers mean to stop iteration when the write fails, the g_hash_table_foreach won't actually stop iterating yet.

The small hiccup was the tests. The tests `g_assert_cmpint (count, ==, 10);` would fail because count became 14 or 15 depending on the hashing.

jralls:
> If the callback is foreach_until_false which should stop iterating through the hashtables, and the xml/sql pricedb writers mean to stop iteration when the write fails, the g_hash_table_foreach won't actually stop iterating yet.

OK, another reason to leave it alone. The backends use [`stable_price_traversal`](https://github.com/Gnucash/gnucash/blob/d8d8614064d6a5ef921e13049d651fe71827cf32/libgnucash/engine/gnc-pricedb.cpp#L2736). It returns immediately if the predicate returns `false` on a price.

The other use case, the one that the tests are attempting to test, just counts the number of prices.  Surely it would be more efficient, never mind more straightforward, to call `g_hash_table_get_values` on the outer hash table, then iterate on that summing the results of `g_hash_table_size` on each inner hash table.

christopherlam:
Ok

PR #1789, opened 2023-09-30
Improve Price Import Assistant clarity

flywire:


jralls:
The first sentence is better, but it's sort of negated by the second. How about
```
The importer needs at least Date and Price columns when importing prices for a single commodity and currency pair set in the Commodity From and Currency To drop-downs. To import prices multiple commodities the importer requires Date, Amount, From Namespace, From Symbol and Currency To columns.
```

flywire:
Fair point, it needs further work below too. Let's see what comes up in the [list discussion](https://lists.gnucash.org/pipermail/gnucash-user/2023-September/108998.html) which runs into [next month](https://lists.gnucash.org/pipermail/gnucash-user/2023-October/109005.html).

PR #1791, opened 2023-10-02
Icu list formatter

christopherlam:
locale-sensitive list_of_strings to string... :smile: (inserted into gnc-date.cpp for convenient access to icu). now looking for a locale-sensitive quoter better than "string". Maybe we need to localise `_("\"%s\"")` @fellen ?

en_AU
image

fr_FR
image

ja_JP
image

zh_TW
image

hi_IN
image

### DIFF
--- a/gnucash/gnome/gnc-plugin-page-register.c
+++ b/gnucash/gnome/gnc-plugin-page-register.c
@@ -3230,7 +3230,7 @@ gnc_plugin_page_register_set_filter_tooltip (GncPluginPageRegister* page)
 
         if (show)
         {
-            char *str = gnc_g_list_stringjoin (show, ", ");
+            char *str = gnc_list_formatter (show);
             t_list = g_list_prepend
                 (t_list, g_strdup_printf ("%s %s", _("Show:"), str));
             g_free (str);
@@ -3238,7 +3238,7 @@ gnc_plugin_page_register_set_filter_tooltip (GncPluginPageRegister* page)
 
         if (hide)
         {
-            char *str = gnc_g_list_stringjoin (hide, ", ");
+            char *str = gnc_list_formatter (hide);
             t_list = g_list_prepend
                 (t_list, g_strdup_printf ("%s %s", _("Hide:"), str));
             g_free (str);
--- a/gnucash/import-export/import-main-matcher.cpp
+++ b/gnucash/import-export/import-main-matcher.cpp
@@ -1876,7 +1876,7 @@ get_peer_acct_names (Split *split)
         g_free (name);
     }
     names = g_list_sort (names, (GCompareFunc)g_utf8_collate);
-    gchar *retval = gnc_g_list_stringjoin (names, ", ");
+    auto retval = gnc_list_formatter (names);
     g_list_free_full (names, g_free);
     g_list_free (accounts_seen);
     return retval;
--- a/libgnucash/engine/gnc-date.cpp
+++ b/libgnucash/engine/gnc-date.cpp
@@ -46,6 +46,7 @@
 
 #include <cinttypes>
 #include <unicode/calendar.h>
+#include <unicode/listformatter.h>
 
 #include "gnc-date.h"
 #include "gnc-date-p.h"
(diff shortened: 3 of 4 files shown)


fellen:
> now looking for a locale-sensitive quoter better than "string". Maybe we need to localise `_("\"%s\"")` @fellen ?

We have already https://github.com/Gnucash/gnucash/blob/9a241ea9d6bbcfb9b381ad786bdb94798661bc13/po/glossary/de.po#L37-L39

Another approach has html:
`<q></q> ` is interpreted language dependent.
Store this as quote_test.html:
```
<!DOCTYPE html>
<html dir="auto" lang="en" prefix="og: https://ogp.me/ns#">
  <head>
    <meta http-equiv="content-type" content="text/html; charset=UTF-8" />
    <title>Quotation Test | GnuCash</title>
  </head>
  <body>
    <p lang="de">Ein Satz mit <q><q><q>einem Zitat</q> in einem Zitat</q> in einem Zitat</q>.</p>
    <p lang="en">A sentence with <q><q><q>a quote</q> within a quote</q> within a quote</q>.</p>
    <p lang="es">Una oración con <q><q><q>una cita</q> dentro de una cita</q> dentro de una cita</q>.</p>
    <p lang="fr">Une phrase avec <q><q><q>une citation</q> dans une citation</q> dans une citation</q>.</p>
    <p lang="uk">Речення з <q><q><q>цитатою</q> в цитаті</q> в цитаті</q>.</p>
    <p lang="zh">一个句子，其中<q><q><q>引用</q>引用内</q>引用内</q>。</p></q>.</p>
  </body>
</html>
```
and view it in your broswer. The 3. level is not always very well, but the first two are fine IMHO.

jralls:
Does `icu::listFormatter` not localize the quotes if you include them in the list-member strings?  If not it's probably too hard to get it right any other way.

christopherlam:
@jralls no
@fellen these quotes aren't actually being used anywhere in code... the cformat string would be more useful.

jralls:
Then what does this do for us?

Plus perhaps localizing the quotes is too hard, because not only do we have to insert the right quote symbols we also have to deal with whether or not the quotes around list members are appropriate in the locale.

christopherlam:
> Then what does this do for us?

You mean localising `"\"%s\"`? It means better quoting to have `«Biens:Banque» «Depenses:Impots» et «Revenue:Boulot»` showing in the import marcher.

jralls:
No, I mean this PR. I understand perfectly well what localizing `\"%s\"` would do. I also know something it won't do by itself, which is put the quotes in the right places in a list.If `icu::listformatter` doesn't apply localization to the list elements then there's no advantage to it over `gnc_g_list_stringjoin`; in fact it's a pessimization because you have to allocate a new `std::vector<icu::UnicodeString>` (BTW you can't stuff a `icu::UnicodeString` with UTF8, it's UTF16. You have to at least use `emplace_back` so that the UnicodeString ctor gets called and you have make sure that ` U_CHARSET_IS_UTF8=1` so that the right converter is called by the ctor.)

Since the use-case here is to wrap each non-localizable (or maybe already-localized) member of a list in quotes you need a special function `gnc_g_list_stringjoin_quoted` that does `g_list_prepend(g_strdup_printf(C_("list member", "\"%s\""), node->data)` and sends the result of that to `gnc_g_list_stringjoin`.

christopherlam:
The purpose of this PR is to localise the commas and the `and` words.

fellen:
> The purpose of this PR is to localise the commas and the `and` words.

Then that should be `, and`, which in de would become `und` without comma.
[Edit:]
Perhaps you want also a or-list: `, or`?

jralls:
> Then that should be , and

That [comma](https://en.wikipedia.org/wiki/Serial_comma) is optional and some people hate on it. The documentation  [icu list formatter](https://unicode-org.github.io/icu-docs/apidoc/dev/icu4c/classicu_1_1ListFormatter.html) isn't consistent about whether `listFormatter` uses it: Some examples have it, others don't.
> Perhaps you want also a or-list: , or?

Depends on the [type](https://unicode-org.github.io/icu-docs/apidoc/dev/icu4c/ulistformatter_8h.html#af311a9fd3c2f3124bb448cf2ac8d1f1b) field passed to the [full factory function](https://unicode-org.github.io/icu-docs/apidoc/dev/icu4c/classicu_1_1ListFormatter.html#aa41c9fc38d6d24b6b831a8c41340e25d

jralls:
Ducking the quoting issue completely. OK, merge at will.

christopherlam:
> Ducking the quoting issue completely. OK, merge at will.

We can definitely localise quoting simply via _ at the cost of annoying the translators! Or we can wait for ICU to gain this capability.

jralls:
> We can definitely localise quoting simply via _ at the cost of annoying the translators! Or we can wait for ICU to gain this capability.

Or we can ignore the problem and leave the quote off of list elements. Quoting list elements is seldom good usage in English anyway.

PR #1793, opened 2023-10-04
Use the default version of Python 3

nomis:
Python scripts that run with the default version of Python 3 by executing with `/usr/bin/python3` that try to `import gnucash` can't find it if it has been built for a different version.

Instead of using other installed versions of Python 3 that happen to be present, default to using the default "unversioned" version.

It doesn't look like CMake are going to fix the default behaviour, so every project has to do this:
https://gitlab.kitware.com/cmake/cmake/-/issues/24878
https://gitlab.kitware.com/cmake/cmake/-/issues/24126
https://gitlab.kitware.com/cmake/cmake/-/merge_requests/8287

This is only supported on CMake 3.20 or newer, so users of older versions will still get the broken behaviour.

The default behaviour for `Python3_FIND_STRATEGY` also changes depending on which version of CMake is used, and needs to be set to `LOCATION` for the change to `Python3_FIND_UNVERSIONED_NAMES` to have any effect:
https://cmake.org/cmake/help/latest/policy/CMP0094.html

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,6 +2,11 @@
 
 cmake_minimum_required (VERSION 3.14.5)
 
+# CMake 3.15+ Python3_FIND_STRATEGY=LOCATION
+if (POLICY CMP0094)
+  cmake_policy(SET CMP0094 NEW)
+endif()
+
 project (gnucash
     VERSION 5.4
 )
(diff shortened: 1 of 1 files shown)


jralls:
ISTM it would be better to set CMP0094 to NEW instead of explicitly setting Python3_FIND_STRATEGY. 

This change will need a [wiki](https://wiki.gnucash.org/wiki/Python_Bindings) update.

nomis:
> ISTM it would be better to set CMP0094 to NEW instead of explicitly setting Python3_FIND_STRATEGY.

I've changed this and updated the wiki to note the change in behaviour present in 5.4

jralls:
Thanks!

PR #1801, opened 2023-10-21
Bug 799108 - "Since Last Run" crashes if there are any errors creating a scheduled transaction

nomis:
Pass a `creation_errors` list to `gnc_sx_instance_model_effect_change()` so that errors creating SX can be stored, and then display them.

### DIFF
--- a/gnucash/gnome/dialog-sx-since-last-run.c
+++ b/gnucash/gnome/dialog-sx-since-last-run.c
@@ -759,8 +759,8 @@ gnc_sx_slr_tree_model_adapter_new (GncSxInstanceModel *instances)
     return rtn;
 }
 
-static void
-creation_error_dialog (GList **creation_errors)
+void
+gnc_ui_sx_creation_error_dialog (GList **creation_errors)
 {
     GtkWidget *dialog = NULL;
     gchar *message = NULL;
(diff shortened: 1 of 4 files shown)


jralls:
@nomis you said in IRC that there's an obvious `nullptr` deref. Can you add a commit to fix that too?

nomis:
> @nomis you said in IRC that there's an obvious `nullptr` deref. Can you add a commit to fix that too?

This is that fix. The function uses `creation_errors` and it was `NULL`:
```
-    gnc_sx_instance_model_effect_change(sx_instances, TRUE, &auto_created_txns, NULL);
+    gnc_sx_instance_model_effect_change(sx_instances, TRUE, &auto_created_txns,
+                                        &creation_errors);
```

jralls:
No., I mean explicitly fix whatever crashed because gnc_sx_instance_model_effect_change was called with a NULL error list as well as not passing the null.

nomis:
Ok, but now it would just fail silently

nomis:
There's still a potential memory leak here... because now `instance_errors` isn't freed

nomis:
That is now fixed

jralls:
Thanks!

PR #1802, opened 2023-10-22
Address, Undefined Behavior, and Leak sanitizers

jralls:
This creates a new build type `Asan` that configures the compiler address, UB, and (gcc only because Apple clang doesn't have one) leak sanitizers. It adds a job that builds it on ubuntu-latest and runs all of the tests.

The first iteration fixes all of the errors that Apple Clang 15 found.


jralls:
@christopherlam Some more test leaks for you to work on!😉

christopherlam:
[here you go ](https://github.com/christopherlam/gnucash/tree/asan) - not sure if sql genuinely requires g_strdup'd strings

jralls:
@christopherlam 
> [here you go ](https://github.com/christopherlam/gnucash/tree/asan)- not sure if sql genuinely requires g_strdup'd strings

404

christopherlam:
https://github.com/christopherlam/gnucash/commit/73c06b86dc2faed589b7e793625dc31d024a9622

christopherlam:
> URI is https://gnucash.github.io/

this uri not activated?

jralls:
The gnucash one won't activate until the PR is merged because branches. It's working [in my repo](https://jralls.github.io/gnucash/Coverage-HTML/).

christopherlam:
> The gnucash one won't activate until the PR is merged because branches. It's working [in my repo](https://jralls.github.io/gnucash/Coverage-HTML/).

Cool..... fwiw guile can also generate lcov data! I'll need to revisit this with guile-3.0

https://github.com/Gnucash/gnucash/blob/f5ee0c07d8c9eb036422319ca0622f079c67d19d/gnucash/report/reports/standard/test/test-transaction.scm#L31

https://www.gnu.org/software/guile/manual/html_node/Code-Coverage.html

jralls:
> fwiw guile can also generate lcov data!

I'm already working on it by wrapping https://github.com/Gnucash/gnucash/blob/f5ee0c07d8c9eb036422319ca0622f079c67d19d/common/cmake_modules/GncAddTest.cmake#L110.

So far it doesn't work. I first tried to wrap just the `(exit (run-test))` in `with-coverage-test` but it made only 4 result files and while all of the tests passed for several it was because they skipped all their test cases for want of dependencies. I'd rather have a more boiler-plate solution than wiring it into every test program.

jralls:
I misunderstood the message about missing dependencies. That turns out to have been from `call-with-coverage`. Every scheme test has messages like "Skipping unknown source file: html-test.scm" and consequently only 4, test-cash-flow, test-c-interface, test-create-accounts, and test-test-extras, write info files and those contain only counts for Guile sources, not for ours.

So I tried your way by changing the `#f` to `#t` in test-transaction.scm. It crashed because `(current-filename)` returned `#f`. Maybe it doesn't work with compiled scheme? I guess you were trying to get `call-with-coverage` to find the source files, but the load path including everything guile should need is passed in via `$GUILE_LOAD_PATH` so is that really likely to make a difference?

christopherlam:
From memory I think `current-filename` is now a macro so it needs to have its parentheses removed. Am not at terminal, will check later.

christopherlam:
I'm afraid I can't make it work anymore. IIRC `add-to-load-path` was crucial to get the source location for the various guile modules. I removed the dirname section and it runs (slowly) but can't find `/tmp/lcov.info` anymore.

christopherlam:
> ~I'm afraid I can't make it work anymore. IIRC `add-to-load-path` was crucial to get the source location for the various guile modules. I removed the dirname section and it runs (slowly) but can't find `/tmp/lcov.info` anymore.~

Found it -- The `with-code-coverage` part wasn't generating coverage data because `run-test-proper` was running the `gnc:test-runner`, and the latter was calling `exit`[*] with the number of failed tests. The following patch will re-enable lcov data generation, but the modules (eg html-chart trep-engine) etc are not available in load-path, therefore they won't be included in lcov output.

[*] the `exit` is/was necessary to allow gnc:test-runner to signal test failures to ninja check. Therefore removing it will make lcov work, but will prevent ninja from detecting a test failure. 

If we remove the `gnc:test-runner` completely then both lcov and ninja will work well, however the test output is not as nice, and it'll stop at the first test failure instead of (my preference) continuing tests.

So, ultimately the lcov guile problem was due to my decisions...

```patch
diff --git a/bindings/guile/test/srfi64-extras.scm b/bindings/guile/test/srfi64-extras.scm
index e572e988a8..b41baee432 100644
--- a/bindings/guile/test/srfi64-extras.scm
+++ b/bindings/guile/test/srfi64-extras.scm
@@ -45,5 +45,5 @@
       (lambda (runner)
         (format #t "Source:~a\npass = ~a, fail = ~a\n"
                 (test-result-ref runner 'source-file) num-passed num-failed)
-        (exit (zero? num-failed))))
+        (zero? num-failed)))
     runner))
diff --git a/gnucash/report/reports/standard/test/test-transaction.scm b/gnucash/report/reports/standard/test/test-transaction.scm
index 100692ea6a..9a9012ceed 100644
--- a/gnucash/report/reports/standard/test/test-transaction.scm
+++ b/gnucash/report/reports/standard/test/test-transaction.scm
@@ -46,17 +46,12 @@
 (setlocale LC_ALL "C")
 
 (define (run-test)
-  (if #f
+  (if #t
       (coverage-test)
       (run-test-proper)))
 
 (define (coverage-test)
-  (let* ((currfile (dirname (current-filename)))
-         (path (string-take currfile (string-rindex currfile #\/))))
-    (add-to-load-path path))
-  (call-with-values
-      (lambda()
-        (with-code-coverage run-test-proper))
+  (call-with-values (lambda () (with-code-coverage run-test-proper))
     (lambda (data result)
       (let ((port (open-output-file "/tmp/lcov.info")))
         (coverage-data->lcov data port)
```

jralls:
> -        (exit (zero? num-failed))))
> +        (zero? num-failed)))


Did the trick for me, along with adding a bunch of share/guile/site/xx/gnucash directories to `GUILE_LOAD_PATH` before starting the tests. At least it did on Debian Unstable, we'll see what happens here and on Debian Stable.

The Guile coverage tests do take 15 minutes in my VM, so maybe they should get a separate option and CI job that doesn't run on every push & PR.

jralls:
[Works here.](https://jralls.github.io/gnucash/Coverage-HTML).

PR #1812, opened 2023-11-05
recurrencePeriodTypeToString and recurrenceWeekendAdjustToString g_strdup only when necessary

christopherlam:
for sql use only. xml use assumes they return a const char*.

I can't set up libsql on my machine and cannot verify this strdup is necessary. From https://github.com/Gnucash/gnucash/pull/1802#issuecomment-1791602981

### DIFF
--- a/libgnucash/backend/sql/gnc-recurrence-sql.cpp
+++ b/libgnucash/backend/sql/gnc-recurrence-sql.cpp
@@ -379,14 +379,13 @@ upgrade_recurrence_table_1_2 (GncSqlBackend* sql_be)
 
     /* Step 2: insert a default value in the newly created column */
     {
-        gchar* weekend_adj_str = recurrenceWeekendAdjustToString (WEEKEND_ADJ_NONE);
+        const gchar* weekend_adj_str = recurrenceWeekendAdjustToString (WEEKEND_ADJ_NONE);
         std::stringstream sql;
         sql << "UPDATE " << TABLE_NAME << " SET " <<
             weekend_adjust_col_table[0]->name() << "='" <<
             weekend_adj_str << "'";
         auto stmt = sql_be->create_statement_from_sql(sql.str());
         sql_be->execute_nonselect_statement(stmt);
-        g_free (weekend_adj_str);
     }
 
     /* Step 3: rewrite the table, requiring the weekend_adj column to be non-null */
--- a/libgnucash/engine/Recurrence.c
+++ b/libgnucash/engine/Recurrence.c
@@ -518,10 +518,10 @@ recurrenceListToString(const GList *r)
     return g_string_free(str, FALSE);
 }
 
-gchar *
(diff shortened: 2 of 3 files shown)


jralls:
There's no such thing as "libsql". There's libmysql (or more commonly libmariadb), libpq (postgresql), and libsqlite3. All three are used by various Linux services, not to mention that they're all generally pulled in by the distro gnucash package  so it's very unlikely that you don't have all of them. The SQL and DBI backend tests use SQLite3 by default; you test the other two by setting URLs to servers with the necessary credentials via [environment variables](https://github.com/Gnucash/gnucash/blob/5937b5d542503311542e454f8f6ca954945e73be/libgnucash/backend/dbi/test/test-backend-dbi-basic.cpp#L669). That's not generally necessary, and certainly isn't for these leaks: Using SQLite3 will find them.

libgnucash/backend/sql/gnc-recurrence-sql.cpp
```diff
@@ -164,8 +164,9 @@ get_recurrence_period_type (gpointer pObject)
     g_return_val_if_fail (pObject != NULL, NULL);
     g_return_val_if_fail (pInfo->pRecurrence != NULL, NULL);
 
-    return (gpointer)recurrencePeriodTypeToString (
-               recurrenceGetPeriodType (pInfo->pRecurrence));
+    return (gpointer)g_strdup (recurrencePeriodTypeToString
+                               (recurrenceGetPeriodType
+                                (pInfo->pRecurrence)));
```
christopherlam: I mean this change preserves the `strdup`ed string. Is it truly necessary? I'd test by saving a book with SX into SQLite3 but I can't seem to enable sqlite3.

jralls: I don't think it is. It's eventually stuffed into a std::string without a subsequent free so making it const char* should be right.

christopherlam: > I don't think it is. It's eventually stuffed into a std::string without a subsequent free

Ok where's that?

jralls: [here](https://github.com/Gnucash/gnucash/blob/4b6a9b85605e3cd9adecf2c60a839fd97c45226a/libgnucash/backend/sql/gnc-sql-column-table-entry.cpp#L153) via [here](https://github.com/Gnucash/gnucash/blob/4b6a9b85605e3cd9adecf2c60a839fd97c45226a/libgnucash/backend/sql/gnc-sql-column-table-entry.hpp#L349), `get_recurrence_period_type` etc. having been passed as the `getter`.

jralls:
You can merge after you update the commit summary to reflect the new reality.

christopherlam:
Oops missed out on your last comment :(

PR #1816, opened 2023-11-06
Fix minor potential memory leak

adamwight:
Returning the literal string will cause the caller to free unallocated memory.


jralls:
Thanks!

PR #1818, opened 2023-11-07
More leaks from asan

christopherlam:


libgnucash/backend/sql/gnc-sql-column-table-entry.cpp
```diff
@@ -152,6 +152,7 @@ GncSqlColumnTableEntryImpl<CT_STRING>::add_to_query(QofIdTypeConst obj_name,
         stream << s;
         vec.emplace_back (std::make_pair (std::string{m_col_name},
                                           quote_string(stream.str())));
+        g_free (s);
```
adamwight: If I understand the code right, there could be a problem here.  `get_row_value_from_object` can produce its result through several paths with mixed pointer ownership.  The `g_object_get` path returns a string owned by the caller, but the `qof_class_get_parameter_getter` path returns from `g_hash_table_lookup` and I think that's a direct copy of the pointer location into the hash table.  This string would need to be duplicated in order to have consistent memory ownership.

christopherlam: Yeah I can't understand this code; g_object_get would occasionally return a newly malloc'd object which should be freed.

bindings/guile/gnc-engine-guile.cpp
```diff
@@ -378,6 +380,8 @@ gnc_query_path2scm (const GSList *path)
     return scm_reverse (path_scm);
 }
 
+std::unordered_set<std::string> params;
```
jralls: Needs to be `static` to keep it out of the global namespace.  But isn't this just a local reinvention of the string cache?

libgnucash/backend/xml/sixtp-dom-generators.cpp
```diff
@@ -246,18 +246,26 @@ add_kvp_value_node (xmlNodePtr node, const gchar* tag, KvpValue* val)
...
+    }
     case KvpValue::Type::NUMERIC:
-        add_text_to_node (val_node, "numeric",
-                          gnc_numeric_to_string (val->get<gnc_numeric> ()));
+    {
+        char *num_str = gnc_numeric_to_string (val->get<gnc_numeric> ());
+        add_text_to_node (val_node, "numeric", num_str);
+        g_free (num_str);
         break;
+    }
```
jralls: `std::ostringstream` would be more idiomatic:
```
std::ostringstream ss;
switch {val->get_type())
{
     case KvpValue::Type::INT64:
          ss << val->get<int64_t>();
          add_text_to_node(val_node, "integer", ss.string().c_str());
          break;
     case KvpValue::Type::DOUBLE:
          ss << val->get<double>();
          add_text_to_node(val_node, "double", ss.string().c_str());
          break;
     case KvpValue::Type::NUMERIC:
          ss << GncNumeric{val->get<gnc_numeric>());
          add_text_to_node(val_node, "numeric", ss.string().c_str());
          break;
```

libgnucash/backend/sql/gnc-sql-column-table-entry.cpp
jralls: So we need specializations of [`get_row_value_from_object<char*>(QofIdTypeConst obj_name,
                                                  const void* pObject, Z)`](https://github.com/Gnucash/gnucash/blob/906d4113c15267175b5ef40c38f9e1b9197263b5/libgnucash/backend/sql/gnc-sql-column-table-entry.hpp#L335) where `Z` is `std::true_type` and `std::false_type` that `g_strdup`s the rv from the Qof Getter. I'll do that when I have time this weekend.

bindings/guile/gnc-engine-guile.cpp
christopherlam: Yes it's a string cache however I don't see where the strings refcount would be decremented.

jralls: In [`gnc_query_path_free`](https://github.com/Gnucash/gnucash/blob/8f7eecd3e94df935549a7e8cecd155ba99ad122a/bindings/guile/gnc-engine-guile.cpp#L406). Which, BTW, will crash spectacularly with this change.

There's another use, [`%typemap(in) QofQueryParamList*`](https://github.com/Gnucash/gnucash/blob/8f7eecd3e94df935549a7e8cecd155ba99ad122a/bindings/engine.i#L224) that leaks the list and its contents. It needs a corresponding [freearg typemap](https://www.swig.org/Doc1.3/Typemaps.html#Typemaps_nn33) to clean up.

christopherlam: Uh no because in general (non-bindings) code the `const char*` are static eg https://github.com/Gnucash/gnucash/blob/8f7eecd3e94df935549a7e8cecd155ba99ad122a/gnucash/import-export/csv-exp/assistant-csv-export.c#L630 and https://github.com/Gnucash/gnucash/blob/8f7eecd3e94df935549a7e8cecd155ba99ad122a/gnucash/gnome/gnc-split-reg.c#L1956 onwards

libgnucash/backend/xml/sixtp-dom-generators.cpp
christopherlam: Not quite: `checked_char_cast (val)` will sanitize for invalid chars therefore the arg cannot be `const`.

bindings/guile/gnc-engine-guile.cpp
christopherlam: Ok I didn't see that `gnc_query_path_free` is bindings only.

libgnucash/backend/xml/sixtp-dom-generators.cpp
jralls: OK, I guess. The alternative of having `std::string` overloads for `add_text_to_node` and `checked_char_cast` is too much to ask at this point.

bindings/guile/gnc-engine-guile.cpp
christopherlam: Rewritten to use the built-in string cache.

christopherlam: and swigged.

christopherlam: ...and swigging seems lotta trouble. reasoning: there are some static QofQueryParamList being used as described in https://github.com/Gnucash/gnucash/pull/1818#discussion_r1390017645 and some may be freed incorrectly.

jralls: For example?

christopherlam: Not sure... and I don't know how to run reports in builddir under gdb.

jralls: You can run the report tests easily enough: Start with `ctest -V -R <test name>`. The top of the output is the command line and environment variables. Start with `gdb guile`, set up the environment (you might create a .gdb for that part, it gets tedious), set any breakpoints you want, and then `run -c <the Scheme fragment to start the test>`.

Tor run an actual report you might try `gdb bin/gnucash-cli` then set up the environment and `run -R run --name <report name> <test datafile>`.

christopherlam:
subsumed into #1831

PR #1820, opened 2023-11-07
boost->std::optional

adamwight:
A few usages weren't converted in libgnucash/app-utils because these belong to dependencies.


adamwight:
boost::program_options doesn't like the mix of optionals, abandoning for now.

PR #1821, opened 2023-11-07
Port some safe bits to c++

adamwight:
Just some minor patches which update the code inside of modules.  Changes on this branch are intended to not ripple over to dependencies outside of each file.

### DIFF
--- a/gnucash/gnucash.cpp
+++ b/gnucash/gnucash.cpp
@@ -58,7 +58,9 @@
 #include <search-core-type.h>
 #include <top-level.h>
 
+#include <boost/filesystem.hpp>
 #include <boost/locale.hpp>
+#include <boost/none.hpp>
 #include <boost/optional.hpp>
 #ifdef __MINGW32__
 #include <boost/nowide/args.hpp>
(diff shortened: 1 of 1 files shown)


gnucash/gnucash.cpp
```diff
@@ -195,9 +195,9 @@ scm_run_gnucash (void *data, [[maybe_unused]] int argc, [[maybe_unused]] char **
 
     gnc_hook_run(HOOK_STARTUP, NULL);
 
-    boost::optional<boost::filesystem::path> fn;
     if (!user_file_spec->nofile)
     {
+        boost::optional<boost::filesystem::path> fn;
         if (auto fn = get_file_to_load (user_file_spec->file_to_load))
```
gjanssens, line 200: You redefine fn inside the `if` clause in the next line. I think you can drop this declaration.

adamwight: Thanks, pushing a correction now!

gnucash/gnucash.cpp
```diff
@@ -114,14 +116,14 @@ load_gnucash_modules()
...
 {
-    if (file_to_load && *file_to_load != '\0')
-        return g_strdup(file_to_load);
-    else
-        /* Note history will always return a valid (possibly empty) string */
-        return gnc_history_get_last();
+    if (!file_to_load.empty())
+        return file_to_load;
+    if (auto last = gnc_history_get_last())
+        return boost::filesystem::path(last);
```
jralls, line 125: Leaks `last`.

jralls:
The boost::filename commit adds a great deal of overhead for no discernible purpose. The `user data_migration_msg` commit would add substantial overhead to no purpose except that `userdata_migration_msg` got moved to gnucash-core-app.cpp and this little fragment was left behind.  `userdata_migration_msg` is never changed from `nullptr` in gnucash.cpp, so the fragment should simply be removed.

PR #1824, opened 2023-11-15
Bug 799090 - Right Clicking scheduled transaction

Bob-IT:
The first commit fixes the bug by changing the right click action to...
If you right click on an unselected row, it does an unselect of all rows and then selects the one under the mouse to apply the popup action to. If you right click on a selected row, all selected rows are used in the popup action.

The second commit changes the delete warning message to list all the schedules selected for deletion.

The last commit is some thing I noticed when looking at the bug, the tool tip for the 'Edit' schedule was not being displayed in the status bar. This was down to a conflict with the main 'Edit' menu entry and can be easily fixed by renaming the scheduled transaction menu entries to 'Edit Schedule'/New Schedule' and 'Delete Schedule' which also brings them in line with the account menu entries.

I think this is good to go unless changes are required so will push end of week.

### DIFF
--- a/gnucash/gnome/gnc-plugin-page-sx-list.c
+++ b/gnucash/gnome/gnc-plugin-page-sx-list.c
@@ -390,19 +390,19 @@ treeview_popup (GtkTreeView *treeview, GdkEvent *event, GncPluginPageSxList *pag
 
     menu = gtk_menu_new();
 
-    menuitem = gtk_menu_item_new_with_mnemonic (_("_New"));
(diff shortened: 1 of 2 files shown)


jralls (approved):
LGTM.

Is the tooltip-update code that works off of the menu entry title ours or from Gtk? ISTM this is showing that it's a bit fragile.

Bob-IT:
Ours, it was only way I could think of doing it and I did not think there were such duplicates when I did a scan for such. One day I may have another look to see if there is a better way.

PR #1828, opened 2023-11-17
Since Last Run changes

Bob-IT:
Some possible bug fixes

The first commit, Bug799241 allows the description and status tree view columns to be sorted. With the description column it is sorted by the date of the first instance in the list and then the description of the schedule. From testing it works OK but can be tweaked if required.
The second commit just allows for the sorting to be saved.

The third commit, Bug607000 removes all the schedules that have any empty status from the view, there is no pending operation.

The fourth commit, Bug798760 allows the OK button to complete the value entry field instead of loosing any changes. A second part to this was for a reminder that required a value, if a value was entered it automatically changes the status to 'To Create'.

The last commit, Bug798025 I was originally going to close but had this idea which I am not sure about. It is all about when the cell renderer text entry is focused / active, the underlying tree view can be scrolled which can lead to confusion as all the entry values are marked 'Need Value'. This commit blocks the scrolling of the tree view while an entry is being made.

### DIFF
--- a/gnucash/gnome/dialog-sx-since-last-run.c
+++ b/gnucash/gnome/dialog-sx-since-last-run.c
@@ -72,6 +72,8 @@ struct _GncSxSinceLastRunDialog
     GtkTreeView *instance_view;
     GtkToggleButton *review_created_txns_toggle;
     GList *created_txns;
+
+    GtkCellEditable *temp_ce; // used when editing values
 };
 
 /* ------------------------------------------------------------ */
@@ -922,6 +924,8 @@ variable_value_changed_cb (GtkCellRendererText *cell,
 
     DEBUG ("variable to [%s] at path [%s]", value, path);
 
+    dialog->temp_ce = NULL;
+
     if (!gtk_tree_model_get_iter (GTK_TREE_MODEL(dialog->editing_model), &tree_iter, model_path))
     {
         gtk_tree_path_free (model_path);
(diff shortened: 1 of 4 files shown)


gnucash/gnome/dialog-sx-since-last-run.c
```diff
@@ -435,6 +435,10 @@ gsslrtma_populate_tree_store (GncSxSlrTreeModelAdapter *model)
             }
         }
 
+        // if there are no instances for the instance skip adding
+        if (g_list_length (instances->instance_list) == 0)
```
jralls, line 441: An empty glist is indicated by GList* being NULL, so `if (instances->instance_list)` does the same with less cost when there are lots of instances.

Also shouldn't this go before populating `last_occur_date_buf`?

jralls (changes requested):
Please use the bug summary for the commit summary; with a two-parter you can append "part1" and "part2" or similar if you like, but put the explanation of what the commit fixes in the body of the commit message. If you don't think that the bug's summary line is good you can edit it and then use that for your commit summary.

It appears to me that the first commit enables sorting by name, date, and status. If that's correct please say so in the commit message.

The second commit saves the SLR sort order in preferences so that it would apply to all books. Shouldn't it be saved in the metadata file so that it will be per-book?

gnucash/gnome/dialog-sx-since-last-run.c
Bob-IT: I think I had a print statement for `last_occur_date_buf` and forgot to move this test after I had removed it.
Changed.

Bob-IT:
> The second commit saves the SLR sort order in preferences so that it would apply to all books. Shouldn't it be saved in the metadata file so that it will be per-book?

I did take the easy option as there was already a per-book setting for the 'Review created transactions'. But thinking about I do not see why you would want to set it differently per book, if your preference is one way, it would be for all.

jralls (approved):

jralls:
> I did take the easy option as there was already a per-book setting for the 'Review created transactions'. 
> But thinking about I do not see why you would want to set it differently per book, if your preference is one way, it would 
> be for all.

You mean per-user, i.e. in Preferences, for Review created transactions and that's along with several others including when to run the dialog and Transaction Editor defaults for auto-create, create-in-advance, and remind-in-advance.

One might have different SLR sorting needs in different books depending on the purpose of each book, just like one might have different sorting needs for accounts or registers. I don't feel strongly about that though. If enough users whine about it we can change it later.

Merge when ready.

PR #1829, opened 2023-11-19
Mark 2 regular expressions as raw strings

jamesjer:
When running gnucash with python 3.12, I see these warnings on startup:
```
/usr/share/gnucash/python/pycons/console.py:167: SyntaxWarning: invalid escape sequence '\['
  self.color_pat = re.compile('\x01?\x1b\[(.*?)m\x02?')
/usr/share/gnucash/python/pycons/shell.py:50: SyntaxWarning: invalid escape sequence '\s'
  self.complete_sep = re.compile('[\s\{\}\[\]\(\)]')
```
Mark both strings as raw strings so the escapes are passed through to the regular expression engine.


jralls:
Thanks!

PR #1831, opened 2023-11-27
[dialog-report-column-view.cpp] new/delete instead of g_new/malloc

christopherlam:
because the struct has cpp (eg vector) objects which would otherwise leak.

### DIFF
--- a/bindings/guile/gnc-engine-guile.cpp
+++ b/bindings/guile/gnc-engine-guile.cpp
@@ -386,17 +386,16 @@ gnc_query_scm2path (SCM path_scm)
     if (!scm_is_list (path_scm))
         return nullptr;
 
-    while (!scm_is_null (path_scm))
+    for (; !scm_is_null (path_scm); path_scm = scm_cdr (path_scm))
     {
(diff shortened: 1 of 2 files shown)


jralls:
Where's the leak? Are `gncp_column_view_edit`s getting destroyed without `gncp_column_view_edit_destroy` being called?

christopherlam:
No, gncp_column_view_edit_destroy is being called appropriately but doesn't nullify the string vector

jralls:
That seems like a compiler bug if calling `free` doesn't run the default destructor. Did you try calling `available_list.clear()` in `gncp_column_view_edit_destroy` and find that it still leaked something?

christopherlam:
I thought `free()` would never call the cpp destructors.

jralls:
Oy, [you're](https://godbolt.org/z/WK7e5rYnf) [right](https://en.cppreference.com/w/cpp/language/delete). That's disappointing because the compiler notices other cases where a variable becomes invalid and arranges for its destruction.

jralls (approved):

PR #1833, opened 2023-12-05
Bug 799152: normalize_reldate_tm() does not handle dates with months greater than 11 or less than -11 correctly.

agwekixj:
Update normalize_reldate_tm() to handle dates with months greater than 11 or less than -11.


jralls:
Not the PR title, that stays on Github. The commit summary needs to match the bug title. And bug titles have the bug number separated by ` - ` not `: `.

jralls:
BTW, code looks a lot cleaner than mine.

jralls:
I fixed the commit myself and force-pushed it back here so the merge would show up.

Thanks for the fix.

PR #1835, opened 2023-12-08
Bug 799156: normalize_struct_tm() does not normalize seconds, minutes, and hours correctly.

agwekixj:
Seconds and minutes should range from 0 to 59 and hours from 0 to 23.  As coded, normalize_struct_tm() will allow seconds and minutes to be 60 and hours to be 24.

### DIFF
--- a/libgnucash/engine/gnc-date.cpp
+++ b/libgnucash/engine/gnc-date.cpp
@@ -126,54 +126,73 @@ gnc_localtime_r (const time64 *secs, struct tm* time)
 }
 
 static void
-normalize_time_component (int *inner, int *outer, int divisor, int base)
+normalize_time_component (int *inner, int *outer, int divisor)
 {
-    if (*inner < base)
+    if (*inner < 0)
     {
         do
         {
             --(*outer);
             *inner += divisor;
-        } while (*inner < base) ;
-        return;
+        } while (*inner < 0) ;
     }
-    while (*inner >= divisor + base)
+    else while (*inner >= divisor)
     {
         ++(*outer);
         *inner -= divisor;
     }
 }
-  
+
 static int
 days_in_month (struct tm* time)
 {
-    normalize_time_component (&(time->tm_mon), &(time->tm_year), 12, 0);
+    normalize_time_component (&(time->tm_mon), &(time->tm_year), 12);
     return gnc_date_get_last_mday (time->tm_mon, time->tm_year + 1900);
 }
-  
+
 static void
 normalize_struct_tm (struct tm* time)
 {
     int days;
     
-    normalize_time_component (&(time->tm_sec), &(time->tm_min), 60, 0);
-    normalize_time_component (&(time->tm_min), &(time->tm_hour), 60, 0);
-    normalize_time_component (&(time->tm_hour), &(time->tm_mday), 24, 0);
-  
-    if (time->tm_mday < 1)     
+    normalize_time_component (&(time->tm_sec), &(time->tm_min), 60);
+    normalize_time_component (&(time->tm_min), &(time->tm_hour), 60);
+    normalize_time_component (&(time->tm_hour), &(time->tm_mday), 24);
+
+    if (time->tm_mday < 1)
     {
         do
         {
             time->tm_mon--;
             time->tm_mday += days_in_month (time);
         } while (time->tm_mday < 1) ;
-        return;
     }
-
-    while (time->tm_mday > (days = days_in_month (time)))
+    else while (time->tm_mday > (days = days_in_month (time)))
     {
         time->tm_mon++;
         time->tm_mday -= days;
+    }
+     
+    /* Gregorian_date throws if it gets an out-of-range year
+     * so bring the time into gregorian_date's range.
+     */
+    if (time->tm_year + 1900 < 1400)
+    {
+        time->tm_sec = 0;
+        time->tm_min = 0;
+        time->tm_hour = 0;
+        time->tm_mday = 1;
+        time->tm_mon = 0;
+        time->tm_year = 1400 - 1900;
+    }
+    else if (time->tm_year + 1900 > 9999)
+    {
+        time->tm_sec = 59;
+        time->tm_min = 59;
+        time->tm_hour = 23;
+        time->tm_mday = 31;
+        time->tm_mon = 11;
+        time->tm_year = 9999 - 1900;
     }
 }
 
--- a/libgnucash/engine/gnc-date.cpp
+++ b/libgnucash/engine/gnc-date.cpp
@@ -126,61 +126,55 @@ gnc_localtime_r (const time64 *secs, struct tm* time)
 }
 
 static void
-normalize_time_component (int *inner, int *outer, unsigned int divisor,
-                          int base)
+normalize_time_component (int *inner, int *outer, int divisor, int base)
 {
-     while (*inner < base)
-     {
-          --(*outer);
-          *inner += divisor;
-     }
-     while (*inner > static_cast<gint>(divisor))
-     {
-          ++(*outer);
-          *inner -= divisor;
-     }
+    if (*inner < base)
+    {
+        do
+        {
+            --(*outer);
+            *inner += divisor;
+        } while (*inner < base) ;
+        return;
+    }
+    while (*inner >= divisor + base)
+    {
+        ++(*outer);
+        *inner -= divisor;
+    }
 }
-
-static void
-normalize_month(int *month, int *year)
+  
+static int
+days_in_month (struct tm* time)
 {
-    ++(*month);
-    normalize_time_component(month, year, 12, 1);
-    --(*month);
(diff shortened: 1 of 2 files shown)


libgnucash/engine/gnc-date.cpp
```diff
@@ -126,61 +126,55 @@ gnc_localtime_r (const time64 *secs, struct tm* time)
...
 normalize_struct_tm (struct tm* time)
 {
-     gint year = time->tm_year + 1900;
-     gint last_day;
-
-     /* Gregorian_date throws if it gets an out-of-range year
-      * so clamp year into gregorian_date's range.
-      */
-     if (year < 1400) year += 1400;
-     if (year > 9999) year %= 10000;
```
jralls, line 162: This needs to stay for the reason the comment explains.

libgnucash/engine/gnc-date.cpp
```diff
@@ -126,61 +126,55 @@ gnc_localtime_r (const time64 *secs, struct tm* time)
...
-normalize_month(int *month, int *year)
+  
+static int
+days_in_month (struct tm* time)
 {
-    ++(*month);
-    normalize_time_component(month, year, 12, 1);
-    --(*month);
+    normalize_time_component (&(time->tm_mon), &(time->tm_year), 12, 0);
+    return gnc_date_get_last_mday (time->tm_mon, time->tm_year + 1900);
```
jralls: With this change `base` is superfluous. You might as well remove it.

libgnucash/engine/gnc-date.cpp
```diff
@@ -126,61 +126,55 @@ gnc_localtime_r (const time64 *secs, struct tm* time)
...
+    if (*inner < base)
+    {
+        do
+        {
+            --(*outer);
+            *inner += divisor;
+        } while (*inner < base) ;
+        return;
+    }
+    while (*inner >= divisor + base)
```
jralls: So the actual change that fixes the bug is here: `>=` instead of `>`, and `divisor + base` to account for the (unnecessary) `base = 1` from `normalize_month`. That really should be isolated as the bug-fix commit and the rest as a separate "Clean up struct tm normalization code" commit.

libgnucash/engine/gnc-date.cpp
```diff
@@ -126,61 +126,55 @@ gnc_localtime_r (const time64 *secs, struct tm* time)
...
-     }
+    if (*inner < base)
+    {
+        do
+        {
+            --(*outer);
+            *inner += divisor;
+        } while (*inner < base) ;
+        return;
+    }
```
jralls: I'm less keen on the `if`-`do`-`while`-`return` construct here. It looks clumsier and [doesn't save a compare, just moves it](https://godbolt.org/z/d463q4ad3).

libgnucash/engine/gnc-date.cpp
agwekixj: I think its bogus but, if you want it, I suggest using something more like:

  ```
if (year < 1400) year = 1400;
  else if (year > 9999) year = 9999;
```

jralls: Neither corrects the likely mistakes of a 2-digit or 5-digit year, and it seems unlikely that anyone would be making book keeping entries in either century. Arguably the first should add something more than 1400.

libgnucash/engine/gnc-date.cpp
agwekixj: The `return` does save a compare.

libgnucash/engine/gnc-date.cpp
agwekixj: Correct.  That is the actual bug and the rest of the changes are "cleanup" that improves the readability.

libgnucash/engine/gnc-date.cpp
agwekixj: I agree.  This is currently the only call of `normalize_time_component()` with a base that is not 0 but I didn't want to assume the intent.

libgnucash/engine/gnc-date.cpp
agwekixj: I agree.  The original code is just wrong, so I removed it.  Limiting the year range here, may mask a problem.  If we do, we should do so after we've normalized - not before.  Should we assert?  Show there be similar year range checks in `gnc-option-date.cpp:normalize_reldate_tm()`?

libgnucash/engine/gnc-date.cpp
jralls: It doesn't. It just trades it for the `if (*inner  < base)`.

agwekixj: It does.  The `return` saves the `*inner >= divisor + base` .

libgnucash/engine/gnc-date.cpp
jralls: > Limiting the year range here, may mask a problem.

What part of `Gregorian_date throws if it gets an out-of-range year` eludes your understanding?

> we should do so after we've normalized - not before

That would make sense only if there was a chance that the precondition fails because of an edge case. Nobody's using GnuCash to handle accounting events in either the 14th or 100th centuries. 

> Should we assert?

Asserts only work in debug builds. Dates come from user input in release builds.

>  Show there be similar year range checks in gnc-option-date.cpp:normalize_reldate_tm()?

No, because it's not part of the constructor for GncDateTime so it's not passing its result to boost::date_time.

agwekixj: > > Limiting the year range here, may mask a problem.
> 
> What part of `Gregorian_date throws if it gets an out-of-range year` eludes your understanding?

I'm fine with `gregorian_date` throwing the error when the date is out-of-range.  I would prefer the `out-of-range year` be flagged on entry.

Since normalize_struct_tm() doesn't know why the year it receives is out-of-range, adjusting the year to make it be within range may make it more difficult to find the issue that caused the year to be out-of-range in the first place.

> 
> > we should do so after we've normalized - not before
> 
> That would make sense only if there was a chance that the precondition fails because of an edge case. 

There is a chance.

> Nobody's using GnuCash to handle accounting events in either the 14th or 100th centuries.

I agree.  So, why should normalize_struct_tm() need to check?

> 
> > Should we assert?
> 
> Asserts only work in debug builds. Dates come from user input in release builds.

Should we throw or return an error?

> 
> > Show there be similar year range checks in gnc-option-date.cpp:normalize_reldate_tm()?
> 
> No, because it's not part of the constructor for GncDateTime so it's not passing its result to boost::date_time.

So, we'll allow some dates to be out-of-range but not others.  An idiot could ask to see transactions  pre-14th or post-100th, but couldn't post them.

jralls: >  Should we throw or return an error?

Neither is an option. The functions that use this `normalize_struct_tm` are `extern "C"` so they can't propagate C++ exceptions. They're 64-bit replacements for the POSIX functions `localtime` and `mktime` which provide no error mechanism. 

> So, we'll allow some dates to be out-of-range but not others. An idiot could ask to see transactions pre-14th or post-100th, but couldn't post them.

The sole objective here is not crashing. The alternatives are to make sure we pass a valid year or calling `GncDateTime(const struct tm)` inside a `try` block. The second case needs a `catch` block that does something that will return valid output for the respective function, so that will involve adjusting the date and calling GncDateTime() again. Fixing the date up front is simpler.

>  I would prefer the out-of-range year be flagged on entry.

Me too, but I didn't have time to protect the 133 calls to `gnc_mktime`, `gnc_localtime`, and `gnc_localtime_r` (some of which are in Scheme code). I still don't have that time. Maybe even less now. Regardless, even if someone did that it wouldn't be good enough because someone could come along later and add a new unprotected call or, not fully understanding the ramifications, remove the protections... just like you're trying to do.

It was a bit lazy of me to bury the guard in `normalize_struct_tm` since it's not strictly normalization. It would be clearer to put it in a separate function, maybe `ensure_valid_year(int&)`, and call that after `normalize_struct_tm`.

agwekixj: I've reimplemented a guard in `normalize_struct_tm` that will set an date with a year earlier than 1400 to January 1, 1400 00:00:00 and any date with a year greater than  9999 to December 31, 9999 23:59:59.  

On an aside, there are 3 additional fields which are not being _normalized_: `tm_wday`, `tm_yday`, `tm_isdst`

jralls: `mtkime` and `timegm` (sorry, I scrambled it with `localtime`; that emits a struct tm, it doesn't consume one) are specified to ignore those fields, as is boost::date_time's posix_time constructor.

jralls:
> I've reimplemented a guard in normalize_struct_tm that will set an date with a year earlier than 1400 to January 1, 1400 00:00:00 and any date with a year greater than 9999 to December 31, 9999 23:59:59.


And I'm done arguing with you. I'll fix the bug myself.

PR #1836, opened 2023-12-09
Trep subtotal links

christopherlam:
There's more to add in the trep.

Subtotal Table can be modified from a boring 2D grid of numbers into a super interactive table with links to the relevant subtotal in the main report. eg the "-$50.60" links to the "Total For 08/10/2023 to 14/10/2023 		$50.60" row. It would be nicer to link to the header instead of the footer but is much more difficult.

image


jralls:
@christopherlam What report and options make that table?

christopherlam:
Transaction report, display/subtotal table set to true.

Sorting/primary = account
Sorting/secondary = date monthly

PR #1842, opened 2023-12-18
Update AUTHORS

Stephan-P:
Added "Stephan Paternotte  Dutch translation" as suggested by jralls


jralls:
@fellen any objection?

PR #1844, opened 2023-12-22
Add spacing for long Invoice ID's

user8446:
This just simply adds a space between "Reference:" and long Invoice ID's printed on the invoice. If it is long enough it appears on the invoice as:

`Reference:MyInvoiceID`

Without the space:

Screenshot 2023-12-22 165516

---

With the space:

Screenshot 2023-12-22 165350


christopherlam:
This breaks translations for strings. It would be much better to improve the css.

user8446 (commented):
Thanks, I didn't think about that. Here is the spacing via CSS

christopherlam:
Thanks, this works!

PR #1849, opened 2024-01-11
### Experiment ### Use browser instaed of Webkit

Bob-IT:
This is just an experiment to see if it was possible to communicate from a Web Browser to Gnucash application.

To achieve this I have created a very basic WebSocket server as part of Gnucash, it is only dealing with messages of 125 text characters or less. With some added javascript to `html-document.scm` to create a web socket connection and intercept the on click requests which are sent to Gnucash as a web socket message. As the connection is started form the web browser page, I needed a consistent identifier for the page so at the moment I am using the temporary file name. This is loaded into a hash table along with the connection pointer and is used to send a message to a specific web page.

The report pages still exist in Gnucash and at the moment all they have is a GtkLabel but could have the report name, options button, reload button and some kind of connection status.

For the reload, a 'RELOAD' message is sent from Gnucash to the web page over the connection with some javascript to do a `location.reload()`. For this to work I had to change how the creation of the temporary file is created so that the same file name is used for the life of the report. There is an issue here as it also creates another web page but that can be fixed in the reload function.

I have also added a 'PRINT' message that opens up the print dialog on the web browser, how useful that is is debated-able.

As you can see, there are print statements for status and errors which would need changing and some messages on the web page for diagnosis.

All of this was just to see if it was possible with no thought to code so there may be leaks which obviously would need checking / fixing and probably a lot more code could be removed..

I built this on my Gentoo VM with a Firefox browser and it did work. I also have tested it on my Windows10 machine with Edge which did work with the exception of the print dialog and bring Gnucash back into focus.

The biggest issue may be changing focus from Gnucash to browser to Gnucash, whether that is fixable for all platforms I do not know.

I will leave this for now for people to ponder if this approach is worth pursuing. I am going to look at `litehtml` and as a start get it to work in a report page. It would still leave the issue of charts not working but I did see that the charts could be created as an image and maybe that could be used.

`Oh` one last thing, the tests that fail are all to do with reports and maybe as a result of me changing `htm-document.scm` and have not looked at but would need to be sorted if this was pursued.

### DIFF
--- a/gnucash/gnome/gnc-plugin-page-report.cpp
+++ b/gnucash/gnome/gnc-plugin-page-report.cpp
@@ -127,7 +127,6 @@ typedef struct GncPluginPageReportPrivate
     /// the gnc_html abstraction this PluginPage contains
 //        gnc_html *html;
     GncHtml *html;
-    gboolean webkit2;
 
     /// the container the above HTML widget is in.
     GtkContainer *container;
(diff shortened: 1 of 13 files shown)


jralls:
This is at least similar to @christopherlam's #1006.

jralls:
Superseded by #2275

PR #1856, opened 2024-01-24
Fix build with GCC 14.

limburgher:


jralls:
What's the error?

limburgher:
/builddir/build/BUILD/gnucash-5.5/redhat-linux-build/bindings/python/gnucash_core.c: error: -Wno-implicit-int detected - is this intentional ? [-Werror]
/builddir/build/BUILD/gnucash-5.5/redhat-linux-build/bindings/python/gnucash_core.c: error: -Wno-implicit-function-declaration detected - is this intentional ? [-Werror]

I'm not sure if ,y fix is optimal but it worked for me. I can test others if you'd like.

jralls:
> I'm not sure if ,y fix is optimal but it worked for me. I can test others if you'd like.

But as you can see from the test results it breaks everything else, so it's obviously not right. 

The errors seem to be related to [some modernization](https://fedoraproject.org/wiki/Changes/PortingToModernC) [efforts at Fedora](https://fedoraproject.org/wiki/Toolchain/PortingToModernC). The author of those changes contacted [other](https://lists.opensuse.org/archives/list/factory@lists.opensuse.org/thread/5OL76NH5AX75WOTZ43O3ZF2JOS3ABBXL/) [distros](https://lists.debian.org/debian-devel/2023/04/msg00067.html) last fall about his proposed changes, but they're not reflected in the [draft gcc 14 release notes](https://gcc.gnu.org/gcc-14/changes.html). Furthermore it seems removing `-Wno-implicit` allows GnuCash to build on Fedora's hacked gcc14 while making it fail *everywhere else* suggests that the hack itself is broken. That's a Fedora bug. @wenottingham FYI.

Another complaint: According to [repology](https://repology.org/project/gcc/versions) while several other distros provide gcc-14 packages, all except Rawhide make it clear that it's not released and provide something else. [Rawhide](https://packages.fedoraproject.org/pkgs/gcc/gcc/fedora-rawhide.html) falsely make it appear that it's a released compiler (they even assign a bogus second release number!) there is also no indication that they've added non-standard errors.

Mind, I agree with the underlying proposal: It's about 30 years past time to get rid of these implicit declarations that were deprecated in the original C-89 standard, but he's doing it wrong. The right place to make it illegal is in the ISO standard committee for the next C standard. Until that happens it's legitimate for compilers to issue a warning (as I believe they all do already) but it is not an error: What compilers may issue errors for is controlled by the standard. It's also a bit of an over-reach for a distro to try to set coding standards for projects that they distribute.

All of which said I support the effort: Ideally we'd have a clean build with the compiler set to its strictest warning level. To achieve that we have to remove the warning exclusion *and fix the resulting errors*. You won't be able to find them with Fedora Rawhide's hack so you'll have to use something else.

limburgher:
Understood. Rawhide isn't meant to be used by end users, and is where we prepare for new releases of compilers, languages, etc. By the time Fedora 40 is released, gcc 14 will be as well. I can't speak to the strategy behind the modernization effort; all I can do is attempt to keep gnucash building.

At present I can't build gnucash either on gcc14 or gcc13 on Fedora without my patch. As it's not suitable for other situations I'll close this PR. If I can make corrections that are applicable generally, I'll submit those.

PR #1858, opened 2024-01-26
Stock acct metadata

christopherlam:
stock-acct can have metadata to save and retrieve the associated proceeds/dividend/capgains/fees accounts.


jralls:
Seems harmless aside from adding more KVP.

PR #1861, opened 2024-01-28
Gnc numeric operators

christopherlam:
@jralls the `GncNumeric` <=> `int64_t` comparison operators don't compile.

test-gnc-numeric.cpp:443:23: note: candidate 2: 'operator>=(double, int)' (built-in)
```
  443 |     EXPECT_TRUE (b >= 0);
      |                       ^
```
test-gnc-numeric.cpp:444:23: error: ISO C++ says that these are ambiguous, even though the worst conversion for the first is better than the worst conversion for the second: [-Werror]

### DIFF
--- a/libgnucash/engine/test/gtest-gnc-numeric.cpp
+++ b/libgnucash/engine/test/gtest-gnc-numeric.cpp
@@ -427,6 +427,28 @@ TEST(gncnumeric_functions, test_cmp)
     EXPECT_EQ(1, a.cmp(INT64_C(12500)));
 }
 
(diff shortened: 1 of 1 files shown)


jralls:
@christopherlam That's because `0` isn't an `int64_t` it's an `int`, i.e. an `int32_t`. The error is telling you that it can be implicitly cast either to a `bool` or an `int64_t` so the compiler can't tell which one you want.

The solution is actually part of your patch, in the first line of context: Change it to `INT64_C(0)` and it should work.

christopherlam:
Ok, wasn't obvious (to me).

PR #1864, opened 2024-02-05
More engine cpp

christopherlam:
Not sure why the tests are failing. From my testing it seems `g_return_val_if_fail` behaves differently in C vs C++. In other words, in C if the express isn't true, it'd log some `g_log` fatal warning, wheras in C++ it's not logging. Amirite?

### DIFF
--- a/libgnucash/engine/CMakeLists.txt
+++ b/libgnucash/engine/CMakeLists.txt
@@ -164,7 +164,7 @@ set (engine_SOURCES
   gnc-features.cpp
   gnc-hooks.c
   gnc-int128.cpp
-  gnc-lot.c
+  gnc-lot.cpp
   gnc-numeric.cpp
   gnc-option-date.cpp
   gnc-option.cpp
--- a/libgnucash/engine/CMakeLists.txt
+++ b/libgnucash/engine/CMakeLists.txt
@@ -147,7 +147,7 @@ set (engine_SOURCES
   Scrub3.c
   ScrubBusiness.c
   ScrubBudget.c
-  Split.c
+  Split.cpp
   TransLog.c
   Transaction.c
   cap-gains.c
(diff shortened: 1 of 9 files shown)


jralls:
No, it's logging: The error messages are printed out just below the test failure message.  Something you did is causing `g_log` to write out the [clang](https://github.com/Gnucash/gnucash/blob/b7e68144154c3677293148a40d66bea36d492a96/libgnucash/engine/test/utest-Account.cpp#L861) version of the function signature instead of just its name so the error strings don't match the tested-for message.

christopherlam:
> Something you did is causing `g_log` to write out the [clang](https://github.com/Gnucash/gnucash/blob/b7e68144154c3677293148a40d66bea36d492a96/libgnucash/engine/test/utest-Account.cpp#L861) version of the function signature instead of just its name so the error strings don't match the tested-for message.

Good catch... although I can only blame the gremlins on this one. No clue.

christopherlam:
Ok nearly there... not sure how to redefine the conditions appropriate for the `#define USE_CLANG_FUNC_SIG 1` line.

jralls:
What do you need to change?

christopherlam:
See latest commit for attempt - setting `USE_CLANG_FUNC_SIG` on mac.

jralls:
`MAC_INTEGRATION` has to be set in the directory's CMakeLists.txt and it isn't in libgnucash/engine/test.

We can lose the Clang version check and just use `#if defined (__clang__)`: We require C++17 and clang v5 was the first to support it. The problem of course is that Clang is now at version 18, so `__clang_major__ == 5` fails. 

The original problem addressed by the hack was that *only* clang versions 5 and <=3.5 wrote that longer function id for C but everything including gcc does so for C++. Now the difference is the space between the identifier and the pointer `*`.

christopherlam:
Ok done. The only CPP change is the use of std::swap instead of the SWAP macro.

christopherlam:
> Ok done. The only CPP change is the use of std::swap instead of the SWAP macro.

Confirmed it works well in test-engine:

```

**** SWAP 0x2288480 and (nil)

**** SWAPPED (nil) and 0x2288480
```

PR #1866, opened 2024-02-12
More engine cpp

christopherlam:
Partial work, but safe.

~`new_quote_sources` is still a GList which leaks the list and contents (and the strdup'ed struct contents).~
- [x] `new_quote_sources` is a vector which frees itself.
- [x] `gnc_quote_source_set_fq_installed` modified to accept a `GncQuotes` directly instead of converting to GList.

### DIFF
--- a/libgnucash/engine/gnc-commodity.cpp
+++ b/libgnucash/engine/gnc-commodity.cpp
@@ -151,11 +151,27 @@ static std::string fq_version;
 
 struct gnc_quote_source_s
 {
-    gboolean supported;
-    QuoteSourceType type;
-    gint index;
-    const char *user_name;		/* User friendly name incl. region code*/
-    const char *internal_name;	/* Name used internally and by finance::quote. */
+private:
+    gboolean m_supported;
+    QuoteSourceType m_type;
+    gint m_index;
+    std::string m_user_name;		/* User friendly name incl. region code*/
+    std::string m_internal_name;	/* Name used internally and by finance::quote. */
+public:
+    bool get_supported () const { return m_supported; }
+    void set_supported (bool supported) { m_supported = supported; }
+    QuoteSourceType get_type () const { return m_type; }
+    size_t get_index () const { return m_index; }
+    void set_index (size_t index) { m_index = index; }
+    const char* get_user_name () const { return m_user_name.c_str(); }
+    const char* get_internal_name () const { return m_internal_name.c_str(); }
+    gnc_quote_source_s (gboolean supported, QuoteSourceType type, gint index,
+                        const char* username, const char* int_name)
+        : m_supported{supported}
+        , m_type{type}
+        , m_index{index}
+        , m_user_name{username}
+        , m_internal_name{int_name} { };
 };
 
 /* To update the following lists scan
@@ -267,7 +283,7 @@ static const int num_single_quote_sources =
     sizeof(single_quote_sources) / sizeof(gnc_quote_source);
 static const int num_multiple_quote_sources =
     sizeof(multiple_quote_sources) / sizeof(gnc_quote_source);
-static GList *new_quote_sources = NULL;
+static std::vector<gnc_quote_source> new_quote_sources;
 
 
 /********************************************************************
@@ -311,7 +327,7 @@ gint gnc_quote_source_num_entries(QuoteSourceType type)
     if (type == SOURCE_MULTI)
         return num_multiple_quote_sources;
 
-    return g_list_length(new_quote_sources);
+    return new_quote_sources.size();
 }
 
 /********************************************************************
@@ -326,12 +342,12 @@ gnc_quote_source_init_tables (void)
 
     for (i = 0; i < num_single_quote_sources; i++)
     {
-        single_quote_sources[i].index = i;
+        single_quote_sources[i].set_index(i);
     }
 
     for (i = 0; i < num_multiple_quote_sources; i++)
     {
-        multiple_quote_sources[i].index = i;
+        multiple_quote_sources[i].set_index(i);
     }
 }
 
@@ -347,24 +363,14 @@ gnc_quote_source_init_tables (void)
 gnc_quote_source *
 gnc_quote_source_add_new (const char *source_name, gboolean supported)
 {
-    gnc_quote_source *new_source;
-
     DEBUG("Creating new source %s", (source_name == NULL ? "(null)" : source_name));
-    new_source = g_new0 (gnc_quote_source, 1);
-    new_source->supported = supported;
-    new_source->type = SOURCE_UNKNOWN;
-    new_source->index = g_list_length(new_quote_sources);
-
     /* This name can be changed if/when support for this price source is
      * integrated into gnucash. */
-    new_source->user_name = g_strdup(source_name);
-
     /* This name is permanent and must be kept the same if/when support
      * for this price source is integrated into gnucash (i.e. for a
      * nice user name). */
-    new_source->internal_name = g_strdup(source_name);
-    new_quote_sources = g_list_append(new_quote_sources, new_source);
-    return new_source;
+    return &new_quote_sources.emplace_back (supported, SOURCE_UNKNOWN, (gint)new_quote_sources.size(),
+                                            source_name, source_name);
 }
 
 /********************************************************************
@@ -375,21 +381,18 @@ gnc_quote_source_add_new (const char *source_name, gboolean supported)
 gnc_quote_source *
 gnc_quote_source_lookup_by_ti (QuoteSourceType type, gint index)
 {
-    gnc_quote_source *source;
-    GList *node;
-
     ENTER("type/index is %d/%d", type, index);
     switch (type)
     {
     case SOURCE_CURRENCY:
-        LEAVE("found %s", currency_quote_source.user_name);
+        LEAVE("found %s", currency_quote_source.get_user_name());
         return &currency_quote_source;
         break;
 
     case SOURCE_SINGLE:
         if (index < num_single_quote_sources)
         {
-            LEAVE("found %s", single_quote_sources[index].user_name);
+            LEAVE("found %s", single_quote_sources[index].get_user_name());
             return &single_quote_sources[index];
         }
         break;
@@ -397,19 +400,18 @@ gnc_quote_source_lookup_by_ti (QuoteSourceType type, gint index)
     case SOURCE_MULTI:
         if (index < num_multiple_quote_sources)
         {
-            LEAVE("found %s", multiple_quote_sources[index].user_name);
+            LEAVE("found %s", multiple_quote_sources[index].get_user_name());
             return &multiple_quote_sources[index];
         }
         break;
 
     case SOURCE_UNKNOWN:
     default:
-        node = g_list_nth(new_quote_sources, index);
-        if (node)
+        if ((size_t)index < new_quote_sources.size())
         {
-            source = static_cast<gnc_quote_source*>(node->data);
-            LEAVE("found %s", source->user_name);
-            return source;
+            auto& source = new_quote_sources.at(index);
+            LEAVE("found %s", source.get_user_name());
+            return &source;
         }
         break;
     }
@@ -421,8 +423,6 @@ gnc_quote_source_lookup_by_ti (QuoteSourceType type, gint index)
 gnc_quote_source *
 gnc_quote_source_lookup_by_internal(const char * name)
 {
-    gnc_quote_source *source;
-    GList *node;
     gint i;
 
     if ((name == NULL) || (g_strcmp0(name, "") == 0))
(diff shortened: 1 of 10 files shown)


christopherlam:
Interestingly, although new_quote_sources is currently a GList* of structs containing strdup'd chars, running valgrind doesn't currently report any leaks.

jralls:
Why would it? The glist is correctly freed after use.

christopherlam:
> Why would it? The glist is correctly freed after use.

huh where?

libgnucash/engine/gnc-commodity.cpp
```diff
@@ -354,20 +354,20 @@ gnc_quote_source_add_new (const char *source_name, gboolean supported)
     gnc_quote_source *new_source;
 
     DEBUG("Creating new source %s", (source_name == NULL ? "(null)" : source_name));
-    new_source = g_new0 (gnc_quote_source, 1);
+    new_source = new gnc_quote_source;
```
gjanssens: This change from `g_new0` to `new` caught my attention. You are in the process of converting a C-struct into a C++ one. There are plenty of places in our C code that are still accessing this as a C-struct  (like dialog-commodity.c, gnc-tree-model-commodity.c, ...). I don't know if this is a safe practice. I mean I don't know if the memory layout is guaranteed to be the same between a C and a C++ struct. So just a heads up from my side.

gjanssens: And that aside, as a minor nitpick - in the commit history this would better fit with your next commit (where you also add a constructor to the struct) than in this commit (which is about replacing `const char *` with `std::string` in the definition.

libgnucash/engine/gnc-commodity.cpp
```diff
@@ -167,100 +167,100 @@ struct gnc_quote_source_s
  * Apply changes here also to the FQ appendix of help.
  */
 static gnc_quote_source currency_quote_source =
-{ TRUE, 0, 0, "Currency", "CURRENCY", "currency" };
+{ TRUE, SOURCE_SINGLE, 0, "Currency", "CURRENCY", "currency" };
```
gjanssens: This one should immediately be set to `SOURCE_CURRENCY` rather than setting it in the init routine further down.

libgnucash/engine/gnc-commodity.cpp
```diff
@@ -326,13 +326,11 @@ gnc_quote_source_init_tables (void)
 
     for (i = 0; i < num_single_quote_sources; i++)
     {
-        single_quote_sources[i].type = SOURCE_SINGLE;
         single_quote_sources[i].index = i;
```
gjanssens: A further potential refactoring would be to get rid of this `index`. I went through the code and as far as I can tell it's not used in any meaningful way within gnc-commodity.cpp and only exposed publicly via `gnc_quote_source_get_index`.

That function is only consumed once in `gnc_ui_commodity_update_quote_info` in dialog-commodity.c.

The usage there is a poor man's way to keep track of which quote source to activate in a listbox. The more Gtk idiomatic way would be to keep a pointer to each source quote directly in the gtk_liststore and search for that one when you need to figure out which line to activate. In theory searching for a line in the list store each time would be slower than accessing it by index. I believe using an index to be a premature optimization: the list of quote sources is small and there is only a single potential search while loading the dialog.

gjanssens: More design consideration: if refactoring dialog-commodity.c is outside of your current scope, you could at the interim also rewrite `gnc_quote_source_get_index`, to calculate the index on the fly. It would have a similar impact on performance, yet eliminate this init routine.

Note that there is a slight difference between these two approaches. The current implementation and the second proposed rewrite build on the assumption that order of quote sources in the internal tables in gnc-commodity.c[pp] and as presented in dialog-commodity.c is equal and can't change. The first proposed alternative doesn't have that assumption. It instead only assumes the quote sources loaded when creating the dialog continue to exist until the dialog is closed again. That would leave room to reorder the quote sources in the user-visible list if we would want to. That difference really only matters if the dialog would be refactored at some point in the future though.

libgnucash/engine/gnc-commodity.cpp
christopherlam: > I mean I don't know if the memory layout is guaranteed to be the same between a C and a C++ struct. So just a heads up from my side.

fixed by making struct members private, and using getters/setters.

libgnucash/engine/gnc-commodity.cpp
```diff
@@ -488,15 +464,38 @@ gnc_quote_source_get_type (const gnc_quote_source *source)
...
-    return source->get_index();
+    switch (source->get_type())
+    {
+    case SOURCE_CURRENCY:
+        return 0;
+    case SOURCE_SINGLE:
+        for (auto i = 0; i < num_single_quote_sources; ++i)
+            if (&single_quote_sources[i] == source)
+                return i;
+        break;
```
gjanssens: I think this can work. If you like yet another step further in the C++ conversion, you may want to look into converting the static arrays of quote sources into a std::vector instead. That would open the doors for some  C++ algorithm fun. With a vector you could replace the generic for-loops with something more expressive like:
```
    auto quote_it = std::find (single_quote_sources.begin(), single_quote_sources.end(), source);
    auto i = std::distance (single_quote_sources.begin(), quote_it);
    if (i == single_quote_sources.size())
        i = 0; // no match found
    return i;
```
Perhaps the std::find will have to be std::find_if with a custom search lambda, but I think it may work as suggested. With a little more refactoring you can reduce the switch to only select the vector to work on as the std::vectors would all be the same type.

```
    auto sources_vec = single_quote_sources(); // can be whatever, just to initialize the vector
    switch (source->get_type())
    {
    case SOURCE_SINGLE:
        sources_vec = single_quote_sources;
        break;
    case SOURCE_MULTI:
        ....
    }
  
    auto quote_it = std::find (sources_vec.begin(), sources_vec.end(), source);
    auto i = std::distance (sources_vec.begin(), quote_it);
    if (i == sources_vec.size())
        i = 0; // no match found
    return i;
```
  
And if it were me, I'd even make the currency_source a std::vector instead of just a struct. That would allow you to treat it just the same as the other vectors everywhere.

With vectors there will be plenty of other generic for loops you can replace with more expressive algorithms in gnc-commodity.cpp by the way. Just tell me when to stop :wink:

christopherlam: Ok if you want. This requires testing (via test-commodities) and UI testing (still works!).

jralls: Changing the quote sources to std::vector will also help #1626. A good next step (not for this PR) will move the F::Q specific code out of gnc-commodity to app-utils/gnc-quotes.cpp or maybe a new app-utils/gnc-finance-quote.cpp.

libgnucash/engine/gnc-commodity.cpp
```diff
@@ -278,12 +282,16 @@ static gnc_quote_source multiple_quote_sources[] =
...
 };
 
-static const int num_single_quote_sources =
-    sizeof(single_quote_sources) / sizeof(gnc_quote_source);
-static const int num_multiple_quote_sources =
-    sizeof(multiple_quote_sources) / sizeof(gnc_quote_source);
-static std::vector<gnc_quote_source> new_quote_sources;
+static QuoteSourceVec new_quote_sources;
 
+// cannot use map or unordered_map because order must be preserved
```
gjanssens, line 280: What other part of the code relies on the order to be preserved ?

libgnucash/engine/gnc-commodity.cpp
```diff
@@ -470,30 +434,14 @@ gnc_quote_source_get_index (const gnc_quote_source *source)
...
-        for (size_t i = 0; i < new_quote_sources.size(); ++i)
-            if (&new_quote_sources[i] == source)
-                return i;
-        break;
-    }
-    default:
-        break;
-    }
+    auto& sources = get_quote_source_from_type (source->get_type());
+    auto is_source = [&source](const auto& source_iter)
```
gjanssens: Calling the parameter to this lambda function `source_iter` is slightly misleading. It's a genuine `gnc_quote_source`, not an iter to one. It set me on the wrong foot for a while...

gjanssens (commented):
This is c++-ifying nicely :)

libgnucash/engine/gnc-commodity.cpp
christopherlam: I didn't want to the `gnc_quote_source_lookup_by_internal` search function to modify its behaviour-- it currently searches currency, single, multi, and new quote sources in this order.

gjanssens: That's what I guessed and fair enough.

Just fyi you could also have achieved that with an (unordered) map by looping over the source type enum and doing a lookup in the map of the corresponding QuoteSourcesVec associated with each loop iterator. The enum is in the proper order as well. But I don't think it matters so your implementation is fine.

I also briefly thought to reduce the 4 QuoteSourcesVec's to a single one and keeping a separate set of iterators into that vector to indicate start and end of each type. However that's probably something that's best considered in the light of #1626 .

libgnucash/engine/gnc-commodity.cpp
gjanssens: I had forgotten about that PR...

While glancing over it I noticed a commit that does away with the `old_internal_name` variable. That's a nice cleanup standing on its own. Perhaps we could cherry-pick that commit to ~~maint~~ stable and rebase this PR on top of it ? That honours the authors work and avoids having to redo it after this PR (I anticipate only a few easy to resolve merge conflicts in this rebase, the other parts of #1626 touching gnc-commodity.cpp will need more significant rework).

gjanssens: I couldn't resist and tried this cherry-pick and merge myself. Turns out it affects almost every commit in this PR... Good news is all merge conflicts are trivial. If you want I can force-push this work to this PR, so you don't have to redo the merges.

christopherlam: > I couldn't resist and tried this cherry-pick and merge myself. Turns out it affects almost every commit in this PR... Good news is all merge conflicts are trivial. If you want I can force-push this work to this PR, so you don't have to redo the merges.

Go ahead :) I'm not entirely sure why old_internal_name exists anyway.

gjanssens: Ok, done. I have also added two commits of my own. One with just minor NULL/nullptr cleanups and the other with two more c++ algorithm conversions.

gjanssens:
Did one more force push to your PR, to fixup some unwanted additions to po/POTFILES.in and rebase on current stable.

gjanssens:
Force-pushed one last time - my nullptr cleanups broke the gnc-quotes test...

christopherlam:
... and cleaned up.

I *think* most &references are correct. I cannot see any unnecessary vector copying.

gjanssens:
> ... and cleaned up.
> 
> I _think_ most &references are correct. I cannot see any unnecessary vector copying.

Thanks, I figured that needed some more attention... Out of curiosity, what do you use to detect vector copying ?

christopherlam:
> > ... and cleaned up.
> > I _think_ most &references are correct. I cannot see any unnecessary vector copying.
> 
> Thanks, I figured that needed some more attention... Out of curiosity, what do you use to detect vector copying ?

I don't know any tool... Here's my approach: add `const` and `&` all over the place and remove if the compiler complains!

libgnucash/engine/gnc-commodity.cpp
```diff
@@ -318,10 +318,13 @@ gnc_quote_source_fq_version (void)
 static QuoteSourceVec&
 get_quote_source_from_type (QuoteSourceType type)
 {
-    for (const auto& [s_type, sources] : quote_sources_map)
-        if (type == s_type)
-            return sources;
+    auto quote_sources_it = std::find_if (quote_sources_map.begin(), quote_sources_map.end(),
```
gjanssens, line 317: I also notice you replaced my use of `unknown_as_default` construct. What was wrong with it ? Too exotic or is it because you wanted to keep the log warning in case an invalid quote source was used ?

christopherlam: The latter. Also this branch includes slight change of behaviour when the quote source type is invalid: some functions return `new_quote_sources` info; others return 0. I'm pondering what's the best approach to mimic previous behaviour.

gjanssens: Ok, that's reasonable. I didn't compare to how it worked originally. I have just followed your changes. So feel free to tweak to mimic previous behaviour.

christopherlam: I rechecked: all functions with an invalid QuoteSourceType will pick `new_quote_sources` instead.

libgnucash/engine/gnc-commodity.cpp
christopherlam: > While glancing over it I noticed a commit that does away with the `old_internal_name` variable. That's a nice cleanup standing on its own. Perhaps we could cherry-pick that commit to stable and rebase this PR on top of it ?

@vincentl this is now done. Your `old_internal_name` cleanup merged in as f5b7bf3401, and this branch is rebased on stable. I've added a tiny modification to modify test to illustrate the change.

christopherlam:
Last commit is attempt to consolidate into 1 vector but is proving difficult so I'll probably remove it later.

libgnucash/engine/gnc-commodity.cpp
```diff
@@ -351,24 +347,13 @@ gnc_quote_source_init_tables (void)
...
      * integrated into gnucash. */
-    new_source->user_name = g_strdup(source_name);
-
     /* This name is permanent and must be kept the same if/when support
      * for this price source is integrated into gnucash (i.e. for a
      * nice user name). */
-    new_source->internal_name = g_strdup(source_name);
-    new_quote_sources = g_list_append(new_quote_sources, new_source);
-    return new_source;
+    return &new_quote_sources.emplace_back (supported, SOURCE_UNKNOWN, source_name, source_name);
```
nomis, line 356: You're returning a reference to an object that can be moved when a vector reallocation occurs.

* https://en.cppreference.com/w/cpp/container/vector/push_back
* https://bugs.gnucash.org/show_bug.cgi?id=799305

christopherlam: I'm struggling a bit to find an elegant solution for this. I don't think it's nice to do vector.reserve(100000) 😭

jralls: 100000 would be silly. 100 would be sufficient, F::Q only has 54, but it's still an ugly and brittle way to do it.

christopherlam: @nomis does the bug disappear when you redefine QuoteSourceVec as std::list<quote_source_info>?

christopherlam: @nomis #1936

nomis: > does the bug disappear when you redefine QuoteSourceVec as std::list<quote_source_info>?

It would do, because std::list doesn't invalidate references when the list changes.

PR #1867, opened 2024-02-14
Add missing file headers in import source files.

BrianRater:
I set the copyright according to who first checked in each file.

### DIFF
--- a/gnucash/import-export/csv-imp/gnc-tokenizer-csv.cpp
+++ b/gnucash/import-export/csv-imp/gnc-tokenizer-csv.cpp
@@ -1,3 +1,27 @@
+/********************************************************************\
+ * gnc-tokenizer-csv.cpp - takes a csv file and converts it into a  *
+ *                         two-dimensional vector of strings (table)*
+ *                                                                  *
+ *  Copyright (C) 2015 Geert Janssens <geert@kobaltwit.be>          *
+ *                                                                  *
+ * This program is free software; you can redistribute it and/or    *
+ * modify it under the terms of the GNU General Public License as   *
+ * published by the Free Software Foundation; either version 2 of   *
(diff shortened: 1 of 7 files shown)


jralls:
You seem to have copied from the header files and neglected to change the file extension on the second line.

@gjanssens These are yours, any comments?

gjanssens:
Thanks for fixing the copyright omissions. I agree with @jralls the file names should be adjusted.

Also I wouldn't duplicate the doxygen comment from each header file to the corresponding source file. For reference the doxygen comment is the comment block that starts with `/** @file`.  This block is parsed by our document generator and it's an unecessary burden to maintain the same block in two different files.

BrianRater:
I think I've updated this correctly.  I'm a novice at git and new to github.  This PR seems to have updated automatically when I pushed my changed to my fork, which is good to know.

jralls:
@BrianRater Good, but we want them as a single commit, so you need to squash the two commits into one and force-push the result. Are you using git in the command line or with a GUI tool?

BrianRater:
Command line.

jralls:
@BrianRater 
Ok, here are the steps:
`git rebase -i stable` will open a buffer in your selected terminal window editor with the lines
```
pick 05525e84bf Adding missing file headers in import source files.
pick 309051e179 cleanup
```
Depending on the state of your local `stable` branch there may be more, you can ignore those. Change the word 'pick' to 'fixup' and save the file. It should tell you that it successfully rebased, and `git log --oneline stable..` should show only the one commit. Once you've satisfied yourself that everything is correct you can `git push -f <fork> import_file_headers` where <fork> is whatever you've named your fork's remote.

BrianRater:
I think I may have it now?

jralls:
👍

Thanks!

PR #1873, opened 2024-02-22
[gnc-plugin-page-register.c] add Desc substring filter

christopherlam:
~Very buggy:~ Desc substring filter in register

~Desc filter needs to be string-sanitized or encoded before saving into state file~

For discussion

image


christopherlam:
I think this is ready for beta testing. It works well, saves and retrieves in .gcm; and the desc_filter is stored as a GString because a std::string was too difficult with the struct in struct.

jralls:
How is this a UX improvement over Find?

christopherlam:
> How is this a UX improvement over Find?

I occasionally temporarily want to work on all transactions with the same Description, and the single-split account register is tidier than the multi-split Find window.

PR #1874, opened 2024-02-24
Gnc plugin page register to cpp

christopherlam:
and will rebase #1873 on top of this.

### DIFF
--- a/gnucash/gnome/gnc-plugin-page-register.cpp
+++ b/gnucash/gnome/gnc-plugin-page-register.cpp
@@ -244,7 +244,6 @@ static void gnc_plugin_page_register_event_handler (QofInstance* entity,
                                                     GncEventData* ed);
 
 static GncInvoice* invoice_from_split (Split* split);
-static GList* invoices_from_transaction (Transaction* trans);
 
 /************************************************************/
 /*                          Actions                         */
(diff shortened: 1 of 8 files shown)


jralls:
That's a rather expansive PR title. "Plugin page register to C++" would be better. "CPP" is ambiguous because it can also mean the C pre-processor, the program that handles `#define`, `#if` and such.

gnucash/gnome/gnc-plugin-page-register.cpp
```diff
@@ -3898,29 +3898,31 @@ gnc_plugin_page_register_cmd_reverse_transaction (GSimpleAction *simple,
     {
         const char *rev = _("A reversing entry has already been created for this transaction.");
         const char *jump = _("Jump to the transaction?");
-        if (gnc_verify_dialog (GTK_WINDOW (window), TRUE, "%s\n\n%s", rev, jump))
-            goto jump_to_trans;
```
jralls, line 3877: A better way to accomplish this would be to extract into new functions the two blocks, say `maybe_jump_to_existing` and `maybe_create_reverse_transaction`, then
```
if (auto new_trans{xaccTransactionGetReversedBy(trans)
{
    if (!maybe_jump_to_existing(/*…*/))
    {
        LEAVE ("reverse already exists");
        return;
    }
}
else
{
     if (!maybe_create_reverse_transaction(/*…*/))
     {
          LEAVE ("reverse cancelled");
          return;
      }
}
```

christopherlam:
uhm. maybe i'll split this into cpp conversion and cppisms in another branch.

PR #1880, opened 2024-03-09
[txn-columns.scm] report which lists transactions and breaks into accounts

christopherlam:
A more generalised version of income-gst-statement.scm

This is ideal for analysing e.g. salaried income payslips with PAYG taxes and insurance deductions.

image

and with complex transactions involving >1 currencies; note horizontal sums values to zero, vertical sums amount to the account balance at report end-date
image

### DIFF
--- a/gnucash/report/reports/CMakeLists.txt
+++ b/gnucash/report/reports/CMakeLists.txt
@@ -45,6 +45,7 @@ set (reports_standard_SCHEME
     standard/reconcile-report.scm
     standard/transaction.scm
     standard/trial-balance.scm
+    standard/txn-columns.scm
     standard/view-column.scm
     standard/taxinvoice.scm
     standard/receipt.scm
--- a/gnucash/report/reports/standard/txn-columns.scm
+++ b/gnucash/report/reports/standard/txn-columns.scm
@@ -1,0 +1,262 @@
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;;
+;; This program is free software; you can redistribute it and/or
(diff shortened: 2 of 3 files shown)


christopherlam:
Herewith transactions from a book with 3 USD and 3 GBP accounts. The accounts of type INCOME/BANK/EXPENSE accounts. There's a regular income->bank->expense account for each, and a USD->GBP and GBP->USD transfer.

Note the blue clickable amounts are designed so that horizontally they add up to zero, and are denominated in the transaction currency, and click to the transaction within the account.

The black amounts are designed to show the currency amount denoted in the account column currency whereby the account currency differs from transaction currency.

This may not be intuitive, but ensures no information is lost.

USD Bank:
image

GBP Bank:
image

christopherlam:
Or, Mk.2: the blue amounts are denoted in the account currency. If the currency differs from transaction currency, the transaction currency value is shown first in `[brackets]`. Thus, the account numbers will line up with the totals.

USD BANK
image

GBP BANK
image

jralls:
They're both hard to read, Mk2 is harder because one must pay closer attention to what each number's currency is. Most clear would have separate split-account-commodity and transaction currency columns each with a header telling the viewer what the column contains.

The concept only works well for a limited number, maybe <8 or <10, of accounts. If there are more the matrix becomes sparse making it more difficult to analyze.

christopherlam:
The intent of this report will be to analyse a subset of transactions, mainly multi split ones, so they shouldn't have multiple currencies nor have too many columns. I think an option to limit the number of columns would be indicated.

I'd also prefer Mk.2 because the vertical amounts summed are used to calculate the total, whereas the horizonal sum is always known to balance to zero, therefore should take a secondary role.

PR #1883, opened 2024-03-11
New txn warnings

christopherlam:
adds mechanism to show easily verifiable warnings into register. in this example, the fuel and groceries accounts were reconciled 11/03/24 and we tried to set posting date 08/03/24.

image

### DIFF
--- a/gnucash/import-export/import-main-matcher.cpp
+++ b/gnucash/import-export/import-main-matcher.cpp
@@ -40,7 +40,9 @@
 
 #include <memory>
 #include <algorithm>
+#include <numeric>
 #include <vector>
+#include <string>
 
 #include "import-main-matcher.h"
 
@@ -59,6 +61,7 @@
 #include "guid.h"
 #include "gnc-session.h"
 #include "Query.h"
+#include <optional>
 
 #define GNC_PREFS_GROUP "dialogs.import.generic.transaction-list"
 #define IMPORT_MAIN_MATCHER_CM_CLASS "transaction-matcher-dialog"
@@ -116,6 +119,7 @@ enum downloaded_cols
     DOWNLOADED_COL_ACTION_UPDATE,
     DOWNLOADED_COL_ACTION_INFO,
     DOWNLOADED_COL_ACTION_PIXBUF,
+    DOWNLOADED_COL_ACTION_PIXBUF_TOOLTIP,
     DOWNLOADED_COL_DATA,
     DOWNLOADED_COL_COLOR,
     DOWNLOADED_COL_ENABLE,
@@ -1258,6 +1262,11 @@ gnc_gen_trans_row_activated_cb (GtkTreeView *treeview,
                                            &first, is_selection, path,
                                            &assigned_account, info);
 
+    auto model = gtk_tree_view_get_model (info->view);
+    GtkTreeIter iter;
+    gtk_tree_model_get_iter (model, &iter, path);
+    refresh_model_row (info, model, &iter, get_trans_info (model, &iter));
+
     gtk_tree_selection_select_path (gtk_tree_view_get_selection (treeview), path);
 
     gchar *namestr = gnc_account_get_full_name (assigned_account);
@@ -1559,7 +1568,7 @@ gnc_gen_trans_init_view (GNCImportMainMatcher *info,
                                 G_TYPE_STRING, G_TYPE_STRING, G_TYPE_INT, //memo stuff
                                 G_TYPE_STRING, G_TYPE_BOOLEAN,
                                 G_TYPE_BOOLEAN, G_TYPE_BOOLEAN, G_TYPE_STRING,
-                                GDK_TYPE_PIXBUF, G_TYPE_POINTER, G_TYPE_STRING,
+                                GDK_TYPE_PIXBUF, G_TYPE_STRING, G_TYPE_POINTER, G_TYPE_STRING,
                                 G_TYPE_BOOLEAN);
     gtk_tree_view_set_model (view, GTK_TREE_MODEL(store));
     g_object_unref (store);
@@ -1909,6 +1918,16 @@ get_peer_acct_names (Split *split)
     return retval;
 }
 
+static inline std::optional<time64>
+acct_recndate_warning (time64 trans_date, const Account* acc)
+{
+    time64 recn_date;
+    if (xaccAccountGetReconcileLastDate (acc, &recn_date) && trans_date < recn_date)
+        return recn_date;
+    else
+        return {};
+}
+
 static void
 refresh_model_row (GNCImportMainMatcher *gui,
                    GtkTreeModel *model,
@@ -1976,6 +1995,7 @@ refresh_model_row (GNCImportMainMatcher *gui,
     ro_text = text = NULL;
     const gchar *color = NULL;
     bool show_pixbuf = true;
+    std::vector<std::string> warnings;
     switch (gnc_import_TransInfo_get_action (info))
     {
     case GNCImport_ADD:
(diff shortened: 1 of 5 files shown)


jralls:
I'm pretty sure that `gnc_split_register_save` fires on edits of existing txns as well as creation of new ones, never mind that `gnc_transaction_get_warnings` will operate on a transaction any time you call it.

This strikes me as something many users will find more annoying than helpful.

christopherlam:
> I'm pretty sure that `gnc_split_register_save` fires on edits of existing txns as well as creation of new ones, never mind that `gnc_transaction_get_warnings` will operate on a transaction any time you call it.

This shows how little I know of the register. The primary purpose of this is to prevent import errors. I was caught by an errant $20 matched to an incorrect account. The import matcher would verify the new transactions prior to committing. In any case this is a Proof of Concept.

gnucash/import-export/import-main-matcher.cpp
```diff
@@ -1909,6 +1909,20 @@ get_peer_acct_names (Split *split)
...
+acct_recndate_warning (time64 trans_date, const Account* acc)
+{
+    time64 recn_date;
+    if (xaccAccountGetReconcileLastDate (acc, &recn_date))
+        return trans_date > recn_date;
+
+    return false;
+}
+
+const char* recn_date_warning = N_("The import date is earlier than reconcile date. \
```
jralls: `static const char*` and inside `refresh_model_row` so that it isn't global and lives in the data segment instead of having to be created on the stack at every invocation.

gnucash/import-export/import-main-matcher.cpp
```diff
@@ -1909,6 +1909,20 @@ get_peer_acct_names (Split *split)
     return retval;
 }
 
+static bool
```
jralls: `static inline bool`. The compiler will probably inline it anyway, but it's good to put the hint in for humans.

jralls (changes requested):
Please rebase away the reverted commits.

christopherlam:
`xaccAccountGetReconcileLastDate` faster and tested 100% including an unusual behaviour confirmed in test.

christopherlam:
Last two commits to have a warning icon with tooltip, instead of adding into the additional comments. Identifying the GtkPixBuf column was not straightforward; its `gtk_tree_view_column_get_sort_column_id` returns `-1`...

Screenshot from 2024-04-04 06-28-19

gnucash/import-export/import-backend.cpp
```diff
@@ -405,6 +405,33 @@ GdkPixbuf* gen_probability_pixbuf(gint score_original, GNCImportSettings *settin
     return retval;
 }
 
+GdkPixbuf* gen_warning_pixbuf ()
```
jralls: Gtk provides standard icons, please use them instead.

gnucash/import-export/import-main-matcher.cpp
```diff
@@ -2028,6 +2043,22 @@ refresh_model_row (GNCImportMainMatcher *gui,
                                      imbalance, acct_full_name);
 
                 }
+                if (auto recn_date = acct_recndate_warning (date, dest_acc))
+                {
+                    static const char* recn_date_warning = N_("The import date %s is \
+earlier than destination account %s whose last reconciled date is %s. Further \
+reconciliation of the destination account may be difficult.");
```
jralls, line 2055: I think you want to say `"The import includes at least one transaction whose posted date %s is before the last reconciliation date %s of account %s for which it has a split. This may make reconciliation of that account difficult."

But that's utter nonsense. GnuCash's reconciliation mechanism handles that situation very nicely. It's a pretty common occurence and raising a warning about it will only be annoying.

christopherlam: I'd offer an alternative view... it's common to have multiple accounts, eg checking, savings, spouse bank accounts tracked.

If I have reconciled my checking account with the bank statement dated 31-dec-2023 successfully, it means all bank's transactions have been matched with my book account. It would be _unusual_ that I'd want to add another split in my checking account _before_ 31-dec-2023. In my view, any import which automatically or manually matches to the checking account before 31-dec-2023 is a mismatch, and it would be nice to detect this.

jralls: In my usage it's quite common for a payment I posted not to clear the bank for a few days, sometimes even weeks and in that case the transaction won't get reconciled until a later statement. It would be pretty annoying for GnuCash to pop a warning dialog every time that happens. It certainly doesn't make reconciling more difficult, the older transactions keep showing up in the reconcile window until they're reconciled, and you don't reconcile them until they show up on the bank statement. 

Of course that doesn't work if you only enter transactions by importing them from the bank, but if you do that you're keeping the bank's books not your own and you have no way of noticing when a transaction doesn't clear. There's not really any reconciliation going on either: You're just downloading transactions from the bank and marking them reconciled. Since both the downloaded transactions and the statement come from the same database if not the same computer it would be pretty strange if they didn't agree.

christopherlam: This warning isn't about handling unreconciled splits in an account already reconciled. It's about popping up a soft warning that we're adding a new split in a reconciled account via import. The warning will only pop up as a yellow triangle instead of the red/yellow/green bars, which is not intrusive.

christopherlam: In my scenario I will exercise much care in _manually_ entering data in my main checking account: it will receive all paychecks, it is regularly used to pay various operating expenses. It is also often reconciled with the bank statement.

I also hold other accounts, eg savings accounts. I'd rely on the bank imports.

This work implements a soft warning only, ensuring the main checking account doesn't receive new transactions before the reconciliation date.

jralls: > The warning will only pop up as a yellow triangle instead of the red/yellow/green bars, which is not intrusive.

 Not exactly, at least according to your screenshot. You add a new column that displays an icon if there's a warning on the transaction. (Oddly the example is warning about an expense account. Equity accounts normally don't represent an external account to reconcile with.) It's not as intrusive as the dialog box in your first pass but it's still intrusive.

> This work implements a soft warning only, ensuring the main checking account doesn't receive new transactions before the reconciliation date.

Ensuring nothing. It adds a flag with a misleading tooltip attached when an import adds a transaction to the import account when the *matched* account has a reconciled date after the transaction's post date.

How might that happen? Either the matcher selected the wrong "other" account or the transaction exists but the matcher failed to match it or the transaction exists  in the "other" account but with a different other account than the one you're importing now. Relying on the "other" account's last reconciled date to catch any of those is dicey: It depends on the reconcile periods of the various accounts being the same and on the order that the accounts are reconciled. It's more reliable to review carefully the import in the matcher window before clicking OK. In your case any added transaction with the checking account as a match is a problem regardless of reconciliation date. If you've already done savings account A's import and it turns up as a matched account in an add in savings account B's import that's a problem too.

christopherlam: > > The warning will only pop up as a yellow triangle instead of the red/yellow/green bars, which is not intrusive.
> 
> Not exactly, at least according to your screenshot. You add a new column that displays an icon if there's a warning on the transaction. (Oddly the example is warning about an expense account. Equity accounts normally don't represent an external account to reconcile with.) It's not as intrusive as the dialog box in your first pass but it's still intrusive.

This was a contrived example. There's no additional column. It reuses the existing automatch pixbuf column.

A better example with standard icon:

image

> > This work implements a soft warning only, ensuring the main checking account doesn't receive new transactions before the reconciliation date.
> 
> Ensuring nothing. It adds a flag with a misleading tooltip attached when an import adds a transaction to the import account when the _matched_ account has a reconciled date after the transaction's post date.

As above it won't prevent the user from adding the transaction, but will flag that the account is _probably_ an incorrect auto-match or manual-match.

> How might that happen? Either the matcher selected the wrong "other" account or the transaction exists but the matcher failed to match it or the transaction exists in the "other" account but with a different other account than the one you're importing now. Relying on the "other" account's last reconciled date to catch any of those is dicey: It depends on the reconcile periods of the various accounts being the same and on the order that the accounts are reconciled.

> It's more reliable to review carefully the import in the matcher window before clicking OK. 

Exactly^ it would be nice to flag automatically these imports whereby an algorithm can warn of obvious warnings that the user _may_ miss. See similar in assistant-stock-transaction.cpp -- it flags, but doesn't prevent adding a new transaction dated _before_ the stock account _latest_ split, which could affect capgains calculations.

PR #1882, opened 2024-03-10
[gnc-numeric.cpp] shortcut parsing num/denom as gnc_numeric

christopherlam:
The gnc_numeric is written as "num/denom" with no whitespace, and denom > 0. This function takes advantage of https://en.cppreference.com/w/cpp/utility/from_chars to parse it.

Benchmarks from my console cli session. From 7.5s to 5.5s i.e. 25% speedup!

```shell
chris@pop-os:~/sources/gnucash [env]$ # WITH fast_char
chris@pop-os:~/sources/gnucash [env]$ time stable-install/bin/gnucash-cli ~/GNC/OZ-xml.gnucash -R show --name "Transaction Report"
This is a development version. It may or may not work.
Report bugs and other problems to gnucash-devel@gnucash.org
You can also lookup and file bug reports at https://bugs.gnucash.org
To find the last stable version, please refer to https://www.gnucash.org/
hello from config-user.scm

* name: Transaction Report
  guid: 2fe3b9833af044abb929a88d5a59620f

real	0m5.711s
user	0m5.629s
sys	0m0.217s
chris@pop-os:~/sources/gnucash [env]$ time stable-install/bin/gnucash-cli ~/GNC/OZ-xml.gnucash -R show --name "Transaction Report"
This is a development version. It may or may not work.
Report bugs and other problems to gnucash-devel@gnucash.org
You can also lookup and file bug reports at https://bugs.gnucash.org
To find the last stable version, please refer to https://www.gnucash.org/
hello from config-user.scm

* name: Transaction Report
  guid: 2fe3b9833af044abb929a88d5a59620f

real	0m5.540s
user	0m5.485s
sys	0m0.203s
chris@pop-os:~/sources/gnucash [env]$ time stable-install/bin/gnucash-cli ~/GNC/OZ-xml.gnucash -R show --name "Transaction Report"
This is a development version. It may or may not work.
Report bugs and other problems to gnucash-devel@gnucash.org
You can also lookup and file bug reports at https://bugs.gnucash.org
To find the last stable version, please refer to https://www.gnucash.org/
hello from config-user.scm

* name: Transaction Report
  guid: 2fe3b9833af044abb929a88d5a59620f

real	0m5.658s
user	0m5.570s
sys	0m0.234s
chris@pop-os:~/sources/gnucash [env]$ # without fast_char
[...]

PR #1885, opened 2024-03-12
[gnc-datetime.cpp] faster datetime parsing for "YYYY-MM-DD HH:MM:SS +0000" format

christopherlam:
use howard hinnant's d/m/y -> time64 converter

benchmarks -- (see #1882 for older benchmarks)

```bash
chris@pop-os:~/sources/gnucash [env]$ time stable-install/bin/gnucash-cli ~/GNC/OZ-xml.gnucash -R show --name "Transaction Report"
This is a development version. It may or may not work.
Report bugs and other problems to gnucash-devel@gnucash.org
You can also lookup and file bug reports at https://bugs.gnucash.org
To find the last stable version, please refer to https://www.gnucash.org/
hello from config-user.scm

* name: Transaction Report
  guid: 2fe3b9833af044abb929a88d5a59620f

real	0m3.426s
user	0m3.403s
sys	0m0.149s
chris@pop-os:~/sources/gnucash [env]$ time stable-install/bin/gnucash-cli ~/GNC/OZ-xml.gnucash -R show --name "Transaction Report"
This is a development version. It may or may not work.
Report bugs and other problems to gnucash-devel@gnucash.org
You can also lookup and file bug reports at https://bugs.gnucash.org
To find the last stable version, please refer to https://www.gnucash.org/
hello from config-user.scm

* name: Transaction Report
  guid: 2fe3b9833af044abb929a88d5a59620f

real	0m3.498s
user	0m3.399s
sys	0m0.216s
chris@pop-os:~/sources/gnucash [env]$ time stable-install/bin/gnucash-cli ~/GNC/OZ-xml.gnucash -R show --name "Transaction Report"
This is a development version. It may or may not work.
Report bugs and other problems to gnucash-devel@gnucash.org
You can also lookup and file bug reports at https://bugs.gnucash.org
To find the last stable version, please refer to https://www.gnucash.org/
hello from config-user.scm

* name: Transaction Report
  guid: 2fe3b9833af044abb929a88d5a59620f

real	0m3.509s
user	0m3.430s
sys	0m0.202s
chris@pop-os:~/sources/gnucash [env]$ time stable-install/bin/gnucash-cli ~/GNC/OZ-xml.gnucash -R show --name "Transaction Report"
This is a development version. It may or may not work.
Report bugs and other problems to gnucash-devel@gnucash.org
You can also lookup and file bug reports at https://bugs.gnucash.org
To find the last stable version, please refer to https://www.gnucash.org/
hello from config-user.scm

* name: Transaction Report
  guid: 2fe3b9833af044abb929a88d5a59620f

real	0m3.653s
user	0m3.569s
sys	0m0.215s
```

### DIFF
--- a/libgnucash/engine/gnc-datetime.cpp
+++ b/libgnucash/engine/gnc-datetime.cpp
@@ -37,6 +37,8 @@
 #include <sstream>
 #include <string>
 #include <vector>
+#include <optional>
+#include <charconv>
 #ifdef __MINGW32__
 #include <codecvt>
 #endif
@@ -282,7 +284,8 @@ public:
     GncDateTimeImpl(const time64 time) : m_time(LDT_from_unix_local(time)) {}
     GncDateTimeImpl(const struct tm tm) : m_time(LDT_from_struct_tm(tm)) {}
     GncDateTimeImpl(const GncDateImpl& date, DayPart part = DayPart::neutral);
-    GncDateTimeImpl(std::string str);
+    GncDateTimeImpl(const std::string& str) : GncDateTimeImpl (str.c_str()) {};
+    GncDateTimeImpl(const char* str);
     GncDateTimeImpl(PTime&& pt) : m_time(pt, tzp->get(pt.date().year())) {}
     GncDateTimeImpl(LDT&& ldt) : m_time(ldt) {}
 
@@ -339,6 +342,50 @@ GncDateTimeImpl::GncDateTimeImpl(const GncDateImpl& date, DayPart part) :
 /* Member function definitions for GncDateTimeImpl.
  */
 
+static bool
+parse_chars_into_num (const char* ptr, const char *end_ptr, int32_t& rv) noexcept
+{
+    auto result = std::from_chars (ptr, end_ptr, rv);
+    return (result.ec == std::errc() && result.ptr == end_ptr);
+}
+
+static std::optional<PTime>
+fast_iso8601_utc_parse (const char* str)
+{
+    int32_t year, month, mday, hour, min, sec;
+
+    // parse the first 4 bytes into year
+    if (!str || !parse_chars_into_num (str, str + 4, year))
+        return {};
+
+    // parse iso-8601 utc format "YYYY-MM-DD HH:MM:SS +0000"
+    if (str[4] == '-' &&
+        parse_chars_into_num (str +  5, str +  7, month) && str[ 7] == '-' &&
+        parse_chars_into_num (str +  8, str + 10, mday)  && str[10] == ' ' &&
+        parse_chars_into_num (str + 11, str + 13, hour)  && str[13] == ':' &&
+        parse_chars_into_num (str + 14, str + 16, min)   && str[16] == ':' &&
+        parse_chars_into_num (str + 17, str + 19, sec)   && str[19] == ' ' &&
+        !strcmp (str + 20, "+0000"))
+    {
+        return PTime (boost::gregorian::date (year, month, mday),
+                      boost::posix_time::time_duration (hour, min, sec));
+    }
+
+    // parse compressed iso-8601 format "YYYYMMDDHHMMSS"
+    if (parse_chars_into_num (str +  4, str +  6, month) &&
+        parse_chars_into_num (str +  6, str +  8, mday)  &&
+        parse_chars_into_num (str +  8, str + 10, hour)  &&
+        parse_chars_into_num (str + 10, str + 12, min)   &&
+        parse_chars_into_num (str + 12, str + 14, sec)   &&
+        str[14] == '\0')
+    {
+        return PTime (boost::gregorian::date (year, month, mday),
+                      boost::posix_time::time_duration (hour, min, sec));
+    }
+
+    return {};
+}
+
 static TZ_Ptr
 tz_from_string(std::string str)
 {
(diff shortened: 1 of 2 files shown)


christopherlam:
Hmm premature, closing down. It seems the 3.3s benchmark was via my cached `faster_iso8601` byte parsing.

christopherlam:
with latest commit: less drastic changes, and benchmark from 5.4s to 4.8s = 11% speedup

jralls:
The PIMPL and impl need the same signature so [`GncDateTime(std::string)](https://github.com/Gnucash/gnucash/blob/27c16517e968479d98552486db7cf4bfa69a6615/libgnucash/engine/gnc-datetime.hpp#L108) and [its def](https://github.com/Gnucash/gnucash/blob/27c16517e968479d98552486db7cf4bfa69a6615/libgnucash/engine/gnc-datetime.cpp#L636) need to change to `std::string_view` too. That will save another string construction and copy; I've always hoped without actually testing that an ISO date-time fits in std::string's short-string buffer, but I suspect that even the no-delimiter version doesn't, so there's allocations going on too.

But this change doesn't use `std::string_view` correctly. The point of `std::string_view` is to provide `std::string` semantics without `std::string`'s memory management. It doesn't do anything for you if you use only `data()`.

Plus there's a risk with passing `std::string_view.data()` to a function expecting a `char*` string: `std::string_view::data()` isn't guaranteed by the standard to be null terminated. It might be if the source is a null-terminated `char*`, but the implementor is allowed to drop the `'\0'`. If the source is a `std::string` it's very unlikely that `std::string_view::data()` is null-terminated because `std::string::data()` isn't.

So a clearer and safer way to do this optimization is to change `GncDateTime(std::string)` to `GncDateTime(const std::string&)` and add `GncDateTime(const char*)`. The latter does the work, so `GncDateTimeImpl(std::string)` becomes `GncDateTimeImpl(const char*)` and `GncDateTimeImpl(const std::string&)` is just
```
     GncDateTimeImpl(const std::string& str) : GncDateTimeImpl(str.c_str());
```

That shouldn't require any changes in client code because the compiler should prefer the explicit `const char*` ctor when presented with a `char*` over doing the implicit construction of a `std::string` from it.

christopherlam:
Ok if we avoid boost's `time_from_string` which demands an std::string and collect the y/m/d h:m:s manually, herewith the latest:


```bash
chris@pop-os:~/sources/gnucash [env]$ time stable-install/bin/gnucash-cli ~/GNC/OZ-xml.gnucash -R show --name "Transaction Report"
This is a development version. It may or may not work.
Report bugs and other problems to gnucash-devel@gnucash.org
You can also lookup and file bug reports at https://bugs.gnucash.org
To find the last stable version, please refer to https://www.gnucash.org/
hello from config-user.scm

* name: Transaction Report
  guid: 2fe3b9833af044abb929a88d5a59620f

real	0m3.814s
user	0m3.740s
sys	0m0.204s
chris@pop-os:~/sources/gnucash [env]$ time stable-install/bin/gnucash-cli ~/GNC/OZ-xml.gnucash -R show --name "Transaction Report"
This is a development version. It may or may not work.
Report bugs and other problems to gnucash-devel@gnucash.org
You can also lookup and file bug reports at https://bugs.gnucash.org
To find the last stable version, please refer to https://www.gnucash.org/
hello from config-user.scm

* name: Transaction Report
  guid: 2fe3b9833af044abb929a88d5a59620f

real	0m3.841s
user	0m3.795s
sys	0m0.179s
chris@pop-os:~/sources/gnucash [env]$ time stable-install/bin/gnucash-cli ~/GNC/OZ-xml.gnucash -R show --name "Transaction Report"
This is a development version. It may or may not work.
Report bugs and other problems to gnucash-devel@gnucash.org
You can also lookup and file bug reports at https://bugs.gnucash.org
To find the last stable version, please refer to https://www.gnucash.org/
hello from config-user.scm

* name: Transaction Report
  guid: 2fe3b9833af044abb929a88d5a59620f

real	0m3.886s
user	0m3.772s
sys	0m0.249s
```

christopherlam:
and avoiding `tzptr->base_utc_offset().seconds()` means benchmarks improve to under 3.8s

christopherlam:
Any idea about asan failure?

libgnucash/engine/gnc-datetime.cpp
```diff
@@ -339,6 +342,34 @@ GncDateTimeImpl::GncDateTimeImpl(const GncDateImpl& date, DayPart part) :
...
+    int32_t year, month, mday, hour, min, sec;
+
+    if (str &&
+        parse_chars_into_num (str,      str +  4, '-', year) &&
+        parse_chars_into_num (str +  5, str +  7, '-', month) &&
+        parse_chars_into_num (str +  8, str + 10, ' ', mday) &&
+        parse_chars_into_num (str + 11, str + 13, ':', hour) &&
+        parse_chars_into_num (str + 14, str + 16, ':', min) &&
+        parse_chars_into_num (str + 17, str + 19, ' ', sec) &&
+        !strcmp (str + 20, "+0000"))
```
jralls: There are two ISO-ish formats, one with and one without delimiters. I don't remember offhand which SQL backend uses which format, but we should optimize both.

libgnucash/engine/gnc-datetime.cpp
```diff
@@ -353,17 +384,23 @@ tz_from_string(std::string str)
...
     m_time(unix_epoch, utc_zone)
 {
-    if (str.empty()) return;
+    if (!str || !str[0]) return;
     TZ_Ptr tzptr;
     try
     {
+        if (auto res = fast_iso8601_utc_parse (str))
+        {
+            static const TZ_Ptr utc_tz = tz_from_string("+0000");
```
jralls: Use [`utz_zone`](https://github.com/Gnucash/gnucash/blob/57e3b6f57661964f036192e76e44c707a50de645/libgnucash/engine/gnc-datetime.cpp#L67) instead of creating a new one.

libgnucash/engine/gnc-datetime.cpp
```diff
@@ -633,7 +670,9 @@ GncDateTime::GncDateTime(const time64 time) :
     m_impl(new GncDateTimeImpl(time)) {}
 GncDateTime::GncDateTime(const struct tm tm) :
     m_impl(new GncDateTimeImpl(tm)) {}
-GncDateTime::GncDateTime(std::string str) :
+GncDateTime::GncDateTime(const std::string& str) :
+    m_impl(new GncDateTimeImpl(str.c_str())) {}
```
jralls: Let GncDateTimeImpl(const std::string&) take care of the conversion.

jralls (changes requested):

jralls (approved):

christopherlam:
This one offers 30% speedup in my book.

PR #1886, opened 2024-03-13
Correct misleading description of SX action

Kaligula0:
Correct misleading description about creating Scheduled Transaction. Using this action on a user-created transaction creates a new Scheduled Transaction, but using this action on a transaction that was created by Schedule just opens its Scheduled Transaction.

### DIFF
--- a/gnucash/ui/gnc-plugin-page-register.ui
+++ b/gnucash/ui/gnc-plugin-page-register.ui
@@ -14,14 +14,14 @@
     <item>
       <attribute name="label" translatable="yes">Assign as payment…</attribute>
       <attribute name="action">gnc-plugin-business-actions.RegisterAssignPayment</attribute>
-      <attribute name="tooltip" translatable="yes">Assign the selected transaction as payment</attribute>
+      <attribute name="tooltip" translatable="yes">Assign the selected transaction as payment.</attribute>
       <attribute name="temp" translatable="no">yes</attribute>
       <attribute name="hidden-when">action-disabled</attribute>
     </item>
     <item>
       <attribute name="label" translatable="yes">Edit payment…</attribute>
       <attribute name="action">gnc-plugin-business-actions.RegisterEditPayment</attribute>
-      <attribute name="tooltip" translatable="yes">Edit the payment this transaction is a part of</attribute>
+      <attribute name="tooltip" translatable="yes">Edit the payment this transaction is a part of.</attribute>
       <attribute name="temp" translatable="no">yes</attribute>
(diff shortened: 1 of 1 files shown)


gnucash/ui/gnc-plugin-page-register.ui
```diff
@@ -294,7 +294,7 @@
     <item>
       <attribute name="label" translatable="yes">Sche_dule…</attribute>
       <attribute name="action">GncPluginPageRegisterActions.ScheduleTransactionAction</attribute>
-      <attribute name="tooltip" translatable="yes">Create a Scheduled Transaction with the current transaction as a template</attribute>
+      <attribute name="tooltip" translatable="yes">Create a Scheduled Transaction with the current transaction as a template or edit the Scheduled Transaction that current transaction was created by</attribute>
```
fellen: Please finish the sentence with a full stop `.`!

Kaligula0: Are you sure? I know the grammar, but there are no full stops at the end of strings/sentences in this file (except for the line 99).

fellen: It would be nice, if you fixed all then.
In the times of DOS almost nobody applied them, but today [weblate [old message]](https://hosted.weblate.org/translate/gnucash/gnucash/en/?checksum=fd06b08929b6ca69&q=Create+a+Scheduled+Transaction+with&sort_by=-priority%2Cposition) is sending us warnings.

jralls:
Thanks!

gnucash/ui/gnc-plugin-page-register.ui
```diff
@@ -846,7 +846,7 @@
         <property name="can-focus">False</property>
         <property name="label" translatable="yes">Sche_dule…</property>
         <property name="action-name">GncPluginPageRegisterActions.ScheduleTransactionAction</property>
-        <property name="tooltip-text" translatable="yes">Create a Scheduled Transaction with the current transaction as a template</property>
+        <property name="tooltip-text" translatable="yes">Create a Scheduled Transaction with the current transaction as a template or edit the Scheduled Transaction that current transaction was created by</property>
```
fellen, line 849: Full stop still missing.
I will fix it despite string freeze, because it will be the same string as we already have.

Kaligula0: Sorry! My Regex didn't fix all of them, because I didn't notice there are &lt;attributes&gt; as well as &lt;properties&gt;. Now I found even more full stops, and fixed them all, I think → #1895.

PR #1889, opened 2024-03-17
More engine cpp

christopherlam:
add more c->cpp conversion.... this is now adequate to make AccountP.h a cpp header with access to std::vector.


libgnucash/engine/Scrub.cpp
```diff
@@ -721,7 +721,7 @@ gnc_transaction_balance_trading (Transaction *trans, Account *root)
     for (imbalance_commod = imbal_list; imbalance_commod;
          imbalance_commod = imbalance_commod->next)
     {
-        gnc_monetary *imbal_mon = imbalance_commod->data;
+        gnc_monetary *imbal_mon = (gnc_monetary*)imbalance_commod->data;
```
jralls: Does neither `static_cast` nor `reinterpret_cast` compile?

libgnucash/engine/cap-gains.cpp
```diff
@@ -126,34 +126,34 @@ latest_pred (time64 earl, time64 tran)
 static gpointer
 finder_helper (GNCLot *lot,  gpointer user_data)
 {
-    struct find_lot_s *els = user_data;
+    struct find_lot_s *els = (struct find_lot_s *)user_data;
```
jralls: Remove `struct` from both sides. In C++ `struct` defines a type just like `class`, so you can refer to it as just `find_lot_s`. Better still, also rename it to `FindLot` to comply with our naming policy.

Also prefer `static_cast<FindLot*>` to C-style `(FindLot*)`. If `static_cast` won't compile use `reinterpret_cast`.

jralls (changes requested):
Good start. Don't put too much effort into any of the Query stuff, that's getting replaced with SQL queries as soon as I can make time to write the new code.

christopherlam:
The point here is to let `splits` become a `vector`. I thought that a significant slowdown when loading datafile is the `g_list_find` in `gnc_account_insert_split`, which would be faster as a vector. However I think a GHashTable would be even faster: I've done some more benchmarks adding an GArray `splits_array` or ghashtable `splits_hash` to cache splits already added. Benchmarks improve from 3.66s to 3.22s (GArray) and 3.08s (GHashTable as follows).

```patch
modified   libgnucash/engine/Account.cpp
@@ -336,6 +336,7 @@ gnc_account_init(Account* acc)
     priv->include_sub_account_balances = TriState::Unset;
 
     priv->splits = NULL;
+    priv->splits_hash = g_hash_table_new (nullptr, nullptr);
     priv->sort_dirty = FALSE;
 }
 
@@ -1516,6 +1517,7 @@ xaccAccountCommitEdit (Account *acc)
         {
             g_list_free(priv->splits);
             priv->splits = NULL;
+            g_hash_table_destroy (priv->splits_hash);
         }
 
         /* It turns out there's a case where this assertion does not hold:
@@ -1930,15 +1932,13 @@ gboolean
 gnc_account_insert_split (Account *acc, Split *s)
 {
     AccountPrivate *priv;
-    GList *node;
 
     g_return_val_if_fail(GNC_IS_ACCOUNT(acc), FALSE);
     g_return_val_if_fail(GNC_IS_SPLIT(s), FALSE);
 
     priv = GET_PRIVATE(acc);
-    node = g_list_find(priv->splits, s);
-    if (node)
-        return FALSE;
+    if (!g_hash_table_add (priv->splits_hash, s))
+        return false;
 
     if (qof_instance_get_editlevel(acc) == 0)
     {
@@ -1966,17 +1966,16 @@ gboolean
 gnc_account_remove_split (Account *acc, Split *s)
 {
     AccountPrivate *priv;
-    GList *node;
 
     g_return_val_if_fail(GNC_IS_ACCOUNT(acc), FALSE);
     g_return_val_if_fail(GNC_IS_SPLIT(s), FALSE);
 
     priv = GET_PRIVATE(acc);
-    node = g_list_find(priv->splits, s);
-    if (NULL == node)
-        return FALSE;
 
-    priv->splits = g_list_delete_link(priv->splits, node);
+    if (!g_hash_table_remove (priv->splits_hash, s))
+        return false;
+
+    priv->splits = g_list_remove (priv->splits, s);
     //FIXME: find better event type
     qof_event_gen(&acc->inst, QOF_EVENT_MODIFY, NULL);
     // And send the account-based event, too
modified   libgnucash/engine/AccountP.h
@@ -130,6 +130,7 @@ typedef struct AccountPrivate
     gboolean balance_dirty;     /* balances in splits incorrect */
 
     GList *splits;              /* list of split pointers */
+    GHashTable *splits_hash;
     gboolean sort_dirty;        /* sort order of splits is bad */
 
     LotList   *lots;		/* list of lot pointers */
```

jralls:
> The point here is to let splits become a vector.

That might not be the right goal. Account split lists get used a bunch of different ways and  choice of container class will optimize some uses and pessimize others.

What's optimal in this case depends a lot on what you think is most important to optimize. A pre-allocated std::vector or a (mostly equivalent) C dynamic array like GArray will yield the best load and destroy performance, the best indexed retrieval performance, and the best traversal (for e.g. calculating balances) performance. Sorting a linked list like a GList or std::list will perform better than a vector/array, and random retrieval of a split via a hashed parameter will be most performant with a hash table. The performance difference between list and array for nserting a split depends on where in the list it's inserted: The expense of traversing the list to the insertion point is offset by the cost of allocating and moving the array items after the insertion, so it's much faster to insert near the beginning of a list or the end of an array.

Traversal of hashtables is really slow because of the btree arrangement. Sorting of hashtables by what the hash indexes is unnecessary but sorting on anything else is impossible: They're ordered solely by their btrees. To get a sorted list you have to extract the hashtable values to a list or array, sort it, and then iterate on the sorted list or array.

christopherlam:
Ok let's leave `splits` alone then. How about using the `splits_hash` as above? The GList search at insertion is currently a bad O(N**2) slowdown. 

EDIT: memory management becomes more difficult. The leaks in test-engine increases from 140Kb to 170Kb

jralls:
Are you proposing to keep a hashtable on top of the split list just for the duplication check? Account can't have only a split hashtable, it needs to be able to sort the splits to compute the balances and hashtables are unsortable.

The fundamental problem is in [the xml backend](https://github.com/Gnucash/gnucash/blob/2c25d0bbc056e52801fe2b9fcd2ecbdcc3578f37/libgnucash/backend/xml/gnc-transaction-xml-v2.cpp#L292) where the splits are loaded into the account one at a time with `xaccAccountInsertSplit`.   Reading XML is a serial process and the splits are inside the transactions so the splits get loaded trannsaction-wise. The solution would be to create a `std::unordered_map<GList<Split*>>` where the unordered map is hashed on the account's GUID. `spl_account_handler` would insert each `Split*` (with a straight `g_list_prepend`) into the account's GList, then after processing the whole file just hand each account's list to the account for sorting and de-duplication. Any duplicates are a serious problem with the book and should be flagged to the user.

Another approach is to convert the XML into an in-memory SQL database. That takes care of de-duplication because the GUID column is [the primary key](https://github.com/Gnucash/gnucash/blob/2c25d0bbc056e52801fe2b9fcd2ecbdcc3578f37/libgnucash/backend/sql/gnc-transaction-sql.cpp#L105) and primary keys must be unique. What I'm not sure about offhand is whether the SQL code also checks whether the split is already inserted in the split list. It doesn't need to so that would be an optimization opportunity.

christopherlam:
> Are you proposing to keep a hashtable on top of the split list just for the duplication check? Account can't have only a split hashtable, it needs to be able to sort the splits to compute the balances and hashtables are unsortable.

I meant the `splits_hash` would be an addition to AccountP.h simply to do the duplicate check.

> The fundamental problem is in [the xml backend](https://github.com/Gnucash/gnucash/blob/2c25d0bbc056e52801fe2b9fcd2ecbdcc3578f37/libgnucash/backend/xml/gnc-transaction-xml-v2.cpp#L292) where the splits are loaded into the account one at a time with `xaccAccountInsertSplit`. Reading XML is a serial process and the splits are inside the transactions so the splits get loaded trannsaction-wise. The solution would be to create a `std::unordered_map<GList<Split*>>` where the unordered map is hashed on the account's GUID. `spl_account_handler` would insert each `Split*` (with a straight `g_list_prepend`) into the account's GList, then after processing the whole file just hand each account's list to the account for sorting and de-duplication. Any duplicates are a serious problem with the book and should be flagged to the user.
> 
> Another approach is to convert the XML into an in-memory SQL database. That takes care of de-duplication because the GUID column is [the primary key](https://github.com/Gnucash/gnucash/blob/2c25d0bbc056e52801fe2b9fcd2ecbdcc3578f37/libgnucash/backend/sql/gnc-transaction-sql.cpp#L105) and primary keys must be unique. What I'm not sure about offhand is whether the SQL code also checks whether the split is already inserted in the split list. It doesn't need to so that would be an optimization opportunity.

I don't know either. If the duplicate check is needed only during load, then the hash table could be moved into the gnc-transaction-xml-v2.cpp#L292 instead. I'll leave this issue for now. As I mentioned above, leaks increase.

PR #1892, opened 2024-03-20
Sixtp string converters use from_char

christopherlam:
This speeds up `string_to_int{32|64}` by 40% and introduces `string_to_guint`, but IRL speedup is less impressive because xml stores most numbers in gnc_numeric "num/denom" format.

This also IIUC fixes some maxint bugs in previous code, ~see commit with test illustrating bug, and the test case fixed in later commits.~

EDIT: ~looks like mac clang cannot use `from_chars` for float~ the `string_to_float` will test cpp features and use `strtod` if required.

Benchmarks:
```
Benchmarks running tests 4E6 times:

using sscanf

$ bin/test-string-converters
 elapsed=7.33942s
Executed 92000006 tests. All tests passed.
$ bin/test-string-converters
 elapsed=7.3811s
Executed 92000006 tests. All tests passed.
$ bin/test-string-converters
 elapsed=7.3455s
Executed 92000006 tests. All tests passed.

with std::from_chars

$ bin/test-string-converters
 elapsed=4.47369s
Executed 92000006 tests. All tests passed.
$ bin/test-string-converters
 elapsed=4.46908s
Executed 92000006 tests. All tests passed.
$ bin/test-string-converters
 elapsed=4.47067s
Executed 92000006 tests. All tests passed.
$ bin/test-string-converters
 elapsed=4.48706s
Executed 92000006 tests. All tests passed.
```

### DIFF
--- a/libgnucash/backend/xml/sixtp-utils.cpp
+++ b/libgnucash/backend/xml/sixtp-utils.cpp
@@ -42,6 +42,9 @@
 #include "sixtp.h"
 #include "sixtp-utils.h"
 
+#include <charconv>
+#include <cctype>
+
 static QofLogModule log_module = GNC_MOD_IO;
 
 gboolean
@@ -138,6 +141,27 @@ concatenate_child_result_chars (GSList* data_from_children)
  */
 
 
+template <typename T>
+static bool parse_chars_into_num (const char* str, T* num_ptr)
+{
+    if (!str || !num_ptr)
+        return false;
+
+    while (std::isspace (*str))
+        ++str;
+
+    const char* end_ptr = str + std::strlen (str);
+
+    auto res = std::from_chars (str, end_ptr, *num_ptr);
+    if (res.ec != std::errc{})
+        return false;
+
+    while (std::isspace (*res.ptr))
+        ++res.ptr;
+
+    return (res.ptr == end_ptr);
+}
+
 /*********/
 /* double
  */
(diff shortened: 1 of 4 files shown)


libgnucash/backend/xml/sixtp-utils.cpp
```diff
@@ -138,58 +141,61 @@ concatenate_child_result_chars (GSList* data_from_children)
...
  */
 
 gboolean
 string_to_double (const char* str, double* result)
 {
-    char* endptr = 0x0;
-
-    g_return_val_if_fail (str, FALSE);
-    g_return_val_if_fail (result, FALSE);
+#if __cpp_lib_to_chars >= 201611L
```
gjanssens, line 172: It may be worth adding a comment here to explain this is due to clang not having std::from_char yet for doubles ? That may help to remember it's (hopefully) a temporary workaround.

christopherlam: Done...

christopherlam:
https://www.reddit.com/r/cpp/comments/dgj89g/cppcon_2019_stephan_t_lavavej_floatingpoint/ this from_chars rabbit hole is deep, with lots of recent developments...

jralls:
Indeed it is. STL gave a lecture at one of the C++ conferences in the before time about the complexities. It's on YouTube, should be pretty easy to find.

jralls:
BTW, the first bit of the STL lecture is about floating point. As long as we stay away from those we avoid most of the potions, bits of cake, and mushroom stuff, leaving only the cheshire cat's teleportation to improve efficiency. ;-)

christopherlam:
More cleanup, remove unused `string_to_int32` and harmonise func calls with templates.

PR #1900, opened 2024-04-01
Allow building with c++20 and greater

limburgher:


limburgher:
This commit: https://github.com/Gnucash/gnucash/commit/b217553cf50b2a86f1069736a200dd1d4f5c6551

Introduces a bug where all returned install paths are in /usr/lib, when some should be in /usr/lib64.  This corrects that but has only been tested on Fedora.

jralls:
The python install location is a bit more complicated. The `sysconfig.get_path` dance is needed to install the GnuCash bindings to `CMAKE_INSTALL_PREFIX` and *only* the `user` scheme allows that by setting `userbase`. The `posix_path` scheme returns a hard-coded value that's independent of passed-in variables. Most everybody who isn't a distro packager building GnuCash needs the `user` scheme, the exception being those who want to install to `/usr/local`, but they want the default scheme.

Interestingly while Debian Bookworm's python3.11 returns `/usr/local/lib/python3.11/dist-packages` from `get_path('platlib')` and `/usr/lib/python3.11/site-packages` from `get_path('platlib', 'posix_prefix')` providing an easy option for packagers, Fedora 39's python3.12 returns `/usr/local/lib64/python3.12/site-packages` for both. I suspect that you want the RPM to install the python bindings to `/usr/lib64/python3.12/site-packages` so neither version of `get_path` is going to work for you unless it's fixed in later versions of Fedora.

I don't think that there's a clean generic way to get both behaviors. The only way at all that I can think of is to introduce a cache variable "DISTRO_BUILD" that uses the posix_prefix scheme if true and the user scheme if not, but even that won't work on Fedora 39.

limburgher:
Maybe the setuptools.distutils.util.sysconfig would work?

jralls:
Maybe, but not for long. That's [provided as a porting tool](https://setuptools.pypa.io/en/latest/deprecated/distutils-legacy.html) for migrating from distutils proper.

CMakeLists.txt
```diff
@@ -510,7 +510,7 @@ if (WITH_PYTHON)
 
   # Determine where to install the python libraries.
   execute_process(
-    COMMAND ${Python3_EXECUTABLE} -c "from sysconfig import get_path; print(get_path('platlib', 'posix_user', {'userbase': '${CMAKE_INSTALL_PREFIX}'}))"
+    COMMAND ${Python3_EXECUTABLE} -c "from sysconfig import get_path; print(get_path('platlib', 'posix_prefix', {'userbase': '${CMAKE_INSTALL_PREFIX}'}))"
```
yselkowitz:
```suggestion
    COMMAND ${Python3_EXECUTABLE} -c "from sysconfig import get_path; print(get_path('platlib', 'posix_prefix', {'platbase': '${CMAKE_INSTALL_PREFIX}'}))"
```

jralls:
That works. Thanks!

PR #1901, opened 2024-04-03
Bug799258 gnc:make-date-exchange-time-fn as a better gnc:make-exchange-time-fn

christopherlam:
creates a better `gnc:make-exchange-time-fn` which doesn't search transactions dated after the date selected.

heavily memory intensive.


christopherlam:
some c++ conversions can definitely take place for `commodity-utilities.scm` functions.

christopherlam:
nopes no good

PR #1895, opened 2024-03-24
More missing full stops

Kaligula0:
Correction of missing full stops in `gnucash/ui/gnc-plugin-page-register.ui`, as @fellen requested in #1886.

PR #1903, opened 2024-04-03
Improve translatable strings.

CWehli:
These strings are labeled as "Source needs review" in Weblate.

### DIFF
--- a/gnucash/gnome/assistant-stock-transaction.cpp
+++ b/gnucash/gnome/assistant-stock-transaction.cpp
@@ -268,9 +268,9 @@ static const TxnTypeVec long_types
         // Translators: this is a stock transaction describing a stock
         // split
         N_("Stock split"),
-        N_("Company issues additional units, thereby reducing the stock price by a divisor "
-           ", while keeping the total monetary value of the overall investment constant. "
-           "\n\nIf the split results in a cash in lieu for remainder units, please "
+        N_("Company issues additional units, thereby reducing the stock price by a divisor, "
+           "while keeping the total monetary value of the overall investment constant.\n\nIf "
+           "the split results in a cash in lieu for remainder units, please "
            "record the sale using the Stock Transaction Assistant first, then record the split.")
     },
     {
@@ -310,8 +310,8 @@ static const TxnTypeVec short_types
         // Translators: this is a stock transaction describing cover
         // buying stock, and recording capital gain/loss
         N_("Buy to cover short"),
-        N_("Buy back stock to cover short position, and record capital gain/loss. "
-           "\n\nIf you are unable to calculate capital gains you can enter a placeholder "
+        N_("Buy back stock to cover short position, and record capital gain/loss.\n\nIf "
+           "you are unable to calculate capital gains you can enter a placeholder "
            "amount and correct it in the transaction later.")
     },
     {
@@ -352,7 +352,7 @@ static const TxnTypeVec short_types
         N_("Company returns capital, and the short stock holder must make a compensatory "
            "payment for the returned capital. This reduces the cost basis (less negative, "
            "towards 0.00 value) without affecting # units. A distribution previously recorded "
-           "as a compensatory dividend is reclassified to compensatory return of capital,"
+           "as a compensatory dividend is reclassified to compensatory return of capital, "
            "often due to end-of-year tax information.")
     },
     {
@@ -395,8 +395,8 @@ static const TxnTypeVec short_types
         // split when shorting stock
         N_("Stock split"),
         N_("Company issues additional units, thereby reducing the stock price by a divisor, "
-           "while keeping the total monetary value of the overall investment constant. "
-           "\n\nIf the split results in a cash in lieu for remainder units, please "
+           "while keeping the total monetary value of the overall investment constant.\n\nIf "
+           "the split results in a cash in lieu for remainder units, please "
            "record the cover buy using the Stock Transaction Assistant first, then record the split.")
     },
     {
--- a/gnucash/gtkbuilder/dialog-preferences.glade
+++ b/gnucash/gtkbuilder/dialog-preferences.glade
@@ -326,7 +326,7 @@ Press 'Close' to return to the preference window
                     <property name="visible">True</property>
                     <property name="can-focus">False</property>
                     <property name="halign">start</property>
-                    <property name="label" translatable="yes">&lt;b&gt;Summarybar Content&lt;/b&gt;</property>
+                    <property name="label" translatable="yes">&lt;b&gt;Summary Bar Content&lt;/b&gt;</property>
                     <property name="use-markup">True</property>
                   </object>
                   <packing>
--- a/gnucash/gtkbuilder/dialog-preferences.glade
+++ b/gnucash/gtkbuilder/dialog-preferences.glade
@@ -182,7 +182,10 @@
               <object class="GtkLabel" id="sep_label">
                 <property name="visible">True</property>
                 <property name="can-focus">False</property>
-                <property name="label" translatable="yes">&lt;b&gt;Separator Character&lt;/b&gt;</property>
+                <property name="label" translatable="yes">Separator Character</property>
+                  <style>
+                    <class name="gnc-class-strong"/>
+                  </style>
                 <property name="use-markup">True</property>
               </object>
               <packing>
@@ -326,7 +329,10 @@ Press 'Close' to return to the preference window
                     <property name="visible">True</property>
                     <property name="can-focus">False</property>
                     <property name="halign">start</property>
-                    <property name="label" translatable="yes">&lt;b&gt;Summary Bar Content&lt;/b&gt;</property>
+                    <property name="label" translatable="yes">Summary Bar Content</property>
+                    <style>
+                      <class name="gnc-class-strong"/>
+                    </style>
                     <property name="use-markup">True</property>
                   </object>
                   <packing>
@@ -376,9 +382,9 @@ Press 'Close' to return to the preference window
                     <property name="can-focus">False</property>
                     <property name="halign">start</property>
                     <property name="label" translatable="yes">Start Date</property>
-                    <attributes>
-                      <attribute name="weight" value="bold"/>
-                    </attributes>
+                    <style>
+                      <class name="gnc-class-strong"/>
+                    </style>
                   </object>
                   <packing>
                     <property name="left-attach">0</property>
@@ -391,9 +397,9 @@ Press 'Close' to return to the preference window
                     <property name="can-focus">False</property>
                     <property name="halign">start</property>
                     <property name="label" translatable="yes">End Date</property>
-                    <attributes>
-                      <attribute name="weight" value="bold"/>
-                    </attributes>
+                    <style>
+                      <class name="gnc-class-strong"/>
+                    </style>
                   </object>
                   <packing>
                     <property name="left-attach">0</property>
@@ -596,7 +602,10 @@ Press 'Close' to return to the preference window
                     <property name="visible">True</property>
                     <property name="can-focus">False</property>
                     <property name="halign">start</property>
-                    <property name="label" translatable="yes">&lt;b&gt;Separator Character&lt;/b&gt;</property>
+                    <property name="label" translatable="yes">Separator Character</property>
+                    <style>
+                      <class name="gnc-class-strong"/>
+                    </style>
                     <property name="use-markup">True</property>
                   </object>
                   <packing>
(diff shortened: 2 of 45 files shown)


gnucash/gnome/assistant-stock-transaction.cpp
```diff
@@ -309,8 +309,8 @@ static const TxnTypeVec short_types
         // Translators: this is a stock transaction describing cover
         // buying stock, and recording capital gain/loss
         N_("Buy to cover short"),
-        N_("Buy back stock to cover short position, and record capital gain/loss. "
-           "\n\nIf you are unable to calculate capital gains you can enter a placeholder "
+        N_("Buy back stock to cover short position, and record capital gain/loss.\n\nIf "
```
fellen, line 313: Better split the message and keep the layout "\n\n" outside of the translation, e.g.:
`printf ("%s\n\n%s", _(para1), _(para2))`

gnucash/gnome/assistant-stock-transaction.cpp
```diff
@@ -394,8 +394,8 @@ static const TxnTypeVec short_types
         // split when shorting stock
         N_("Stock split"),
         N_("Company issues additional units, thereby reducing the stock price by a divisor, "
-           "while keeping the total monetary value of the overall investment constant. "
-           "\n\nIf the split results in a cash in lieu for remainder units, please "
+           "while keeping the total monetary value of the overall investment constant.\n\nIf "
```
fellen, line 398: Same as above

gnucash/gtkbuilder/dialog-preferences.glade
```diff
@@ -326,7 +326,7 @@ Press 'Close' to return to the preference window
                     <property name="visible">True</property>
                     <property name="can-focus">False</property>
                     <property name="halign">start</property>
-                    <property name="label" translatable="yes">&lt;b&gt;Summarybar Content&lt;/b&gt;</property>
+                    <property name="label" translatable="yes">&lt;b&gt;Summary Bar Content&lt;/b&gt;</property>
```
fellen: See https://wiki.gnucash.org/wiki/GUI_Guidelines#Markups_in_Text

gnucash/import-export/aqb/dialog-ab.glade
```diff
@@ -1727,7 +1727,7 @@
                     <property name="model">imexporter-list</property>
                     <property name="visible">True</property>
                     <property name="can-focus">True</property>
-                    <property name="tooltip-text" translatable="yes">This list contains the file formats that AQBanking knows how to import. The profile list contains specializations and is populated according to which format you select in this list.</property>
+                    <property name="tooltip-text" translatable="yes">This list contains the file formats that AqBanking knows how to import. The profile list contains specializations and is populated according to which format you select in this list.</property>
```
fellen, line 1733: Ideally we would in a separate "preparing patch for msgmerge" replace all occurrences of "AQBanking" in all .po  files using sed. Before grep also the sources.

gnucash/import-export/bi-import/dialog-bi-import-gui.c
```diff
@@ -191,7 +191,7 @@ gnc_plugin_bi_import_getFilename(GtkWindow *parent)
     gtk_file_filter_set_name (filter, "text files (*.txt)");
     gtk_file_filter_add_pattern (filter, "*.txt");
     filters = g_list_append( filters, filter );
-    filename = gnc_file_dialog(parent, _("Import Bills or Invoices from csv"), filters, NULL, GNC_FILE_DIALOG_IMPORT);
+    filename = gnc_file_dialog(parent, _("Import Bills or Invoices from CSV"), filters, NULL, GNC_FILE_DIALOG_IMPORT);
```
fellen, line 194: Perhaps also a term to grep for.

gnucash/report/reports/standard/taxinvoice.scm
```diff
@@ -317,9 +317,9 @@
        (set-opt options headingpage optname-tax-rate (G_ "GST Rate"))
        (set-opt options headingpage optname-tax-amount (G_ "GST Amount"))
        (set-opt options headingpage2 optname-amount-due (G_ "Amount Due (inc GST)"))
-       (set-opt options headingpage2 optname-invoice-number-text (G_ "Invoice #: "))
-       (set-opt options headingpage2 optname-ref-text (G_ "Reference: "))
-       (set-opt options headingpage2 optname-jobname-text (G_ "Engagement: "))
+       (set-opt options headingpage2 optname-invoice-number-text (G_ "Invoice #:"))
```
fellen, line 320: Do we need to insert the space somewhere else?

libgnucash/engine/gnc-optiondb.cpp
```diff
@@ -1335,11 +1335,11 @@ gnc_option_db_book_options(GncOptionDB* odb)
                                      N_("Length of time to change the used invoice report. A value of 0 means disabled."),
                                      0.0, 0.0, 20.0, 1.0);
     gnc_register_taxtable_option(odb, business_section,
-                                 N_("Default Customer TaxTable"), "f1",
+                                 N_("Default Customer Tax Table"), "f1",
```
fellen, line 1339: Also a grep candidate.

fellen (commented):
Good approach, but there is still room for improvements. ;-)

gnucash/report/reports/standard/taxinvoice.scm
CWehli: no, i've checked the report and it looks ok for me.

gnucash/gtkbuilder/dialog-preferences.glade
CWehli: What is wrong with this change? The formatting described in number 3 of the linked wiki page is exactly what you see here.

CWehli:
Are there any errors in the content of the changes? If not, please accept the PR, as I can't work through the tasks you suggested.
I am not familiar enough with `grep` to be able to change all files automatically. And the suggestions with `printf()` would be applied in many places.

fellen:
To be done after PR #1895

gnucash/gtkbuilder/dialog-preferences.glade
fellen: Eben, das ist die unerwünschteste Form. Besser:
```
<property name="label" translatable="yes">Summary Bar Content</property>
<style>
  <class name="gnc-class-strong"/>
</style>

```
[Edit:] Tippfehler und Link gefixed im Wiki.

CWehli: ok, that's not even 100 hits, I'll bring that as a separate commit.

fellen:
Very good, but I see still markups in glade files:
```
grep -inr '&lt;.&gt;' gnucash/gtkbuilder/
gnucash/gtkbuilder/assistant-csv-trans-import.glade:1074:If one of those &lt;i&gt;checks fails&lt;/i&gt; you'll be automatically redirected to the preview page to try and correct.
gnucash/gtkbuilder/assistant-csv-trans-import.glade:1078:If this is your &lt;i&gt;initial import into a new file&lt;/i&gt;, you will first see a dialog for setting book options, since these can affect how imported data are converted to GnuCash transactions. If this is an existing file, the dialog will not be shown.
gnucash/gtkbuilder/assistant-csv-trans-import.glade:1080:If this is the &lt;i&gt;first time importing&lt;/i&gt;, you will find that all lines may need to be associated. On subsequent imports, the importer will try to associate the transactions based on previous imports.
gnucash/gtkbuilder/dialog-new-user.glade:101:                <property name="label" translatable="yes">If you press the &lt;i&gt;Yes&lt;/i&gt; button, the &lt;i&gt;Welcome to GnuCash&lt;/i&gt; dialog will be displayed again next time you start GnuCash. If you press the &lt;i&gt;No&lt;/i&gt; button, it will not be displayed again.</property>
gnucash/gtkbuilder/dialog-new-user.glade:230:                    <property name="label" translatable="yes">There are some predefined actions available that most new users prefer to get started with GnuCash. Select one of these actions from below and click the &lt;i&gt;OK&lt;/i&gt; button or press the &lt;i&gt;Cancel&lt;/i&gt; button if you don't want to perform any of them.</property>
gnucash/gtkbuilder/assistant-hierarchy.glade:542:&lt;b&gt;Note:&lt;/b&gt; all accounts except Equity and placeholder accounts may have an opening balance.</property>
```
If we can replace the last remaining `&lt;b&gt;` in
gnucash/gtkbuilder/assistant-hierarchy.glade:542:&lt;b&gt;Note:&lt;/b&gt; all accounts except Equity […]
we can patch the po files with sed.
If you are unsure, perhaps @Bob-IT can help you to split this label?

CWehli:
@Bob-IT , @fellen 
I wonder if ‘Note:’ really needs to be written in bold. Other labels are not formatted either.

Bob-IT:
> @Bob-IT , @fellen I wonder if ‘Note:’ really needs to be written in bold. Other labels are not formatted either.

I think this one is easy to fix by splitting the text into separate labels like...
[0001-assistant-hierarchy.txt](https://github.com/Gnucash/gnucash/files/15346061/0001-assistant-hierarchy.txt)


As for the other two files, if you want to remove the markup, I do not think splitting the text is worth while or appropriate. The only thing I can suggest is for the dialog-user, remove all the italics and just put `"` around the `Yes` and `No` text. For csv-trans-import just remove the italics.

fellen:
Seems OK for me, but now someone (I fear I) has to create the preparing patch for the po files.

jralls:
@fellen
>but now someone has to create the preparing patch for the po files.

I started on that but quickly realized that there are wording changes as well as formatting changes. I can robotically remove the markup and the associated fuzzies but while I can manage to fix the translation of
```diff
-"the starting balance.\n"
-"\n"
-"<b>Note:</b> all accounts except Equity and placeholder accounts may have an "
-"opening balance."
+"the starting balance."
```
in some European languages there's no way I could do so in the languages using non-latin scripts. What do you do in those cases?

fellen:
> @fellen
> 
> > but now someone has to create the preparing patch for the po files.
> 
> I started on that but quickly realized that there are wording changes as well as formatting changes. I can robotically remove the markup and the associated fuzzies

If you patch the po files before the msgmerge, no new fuzzy flags should be created. That is also the logical error in @cstim's [remove-suffix.sh](https://github.com/Gnucash/gnucash/blob/stable/po/remove-suffix.sh). It would also remove before existing fuzzy flags.
> but while I can manage to fix the translation of
> 
> ```diff
> -"the starting balance.\n"
> -"\n"
> -"<b>Note:</b> all accounts except Equity and placeholder accounts may have an "
> -"opening balance."
> +"the starting balance."
> ```
This label was split in 2 or 3. IMHO it is OK to mark them as fuzzy.
> in some European languages there's no way I could do so in the languages using non-latin scripts. What do you do in those cases?

I see 2 tasks:

1. Remove markups: Here it is possible, that we end with duplicate MsgIds, e.g.`<b>Note:</b>` maps to `Note:` . That can be checked by a "normal" `msgformat -c …` and then remove the duplicate.
2. Append full stops: Ideally in ja and both zh writing it should use  `。` instead of `.`.

fellen:
@jralls other patches  can map

- `AQBanking` and similar forms -> `AqBanking` in msgid and msgstr
- `csv` -> `CSV`  in msgid and msgstr, but not location
- `: "` -> `:"`
English inaccuracies, msgid only:
- `Summarybar` -> `Summary Bar`
- `TaxTable` -> `Tax Table`

jralls:
Largely manual then. Well, that isn't happening by Sunday so I guess these changes get to sit out another cycle.

As for 
>Append full stops: Ideally in ja and both zh writing it should use  。 instead of ..

Uh-huh. We also have Arabic, Bengali, Cyrillic, Devanagari, Hangul,  Hebrew, Kannada, Manipuri, Marathi, Tamil, Telugu, Urdu, and Vietnamese. I guess one can look at existing msgids with full stops and see if there's a corresponding character at the end of the line. Maybe by December.

jralls:
So it's two days to December and I haven't had time to do any of the mods. This is Too Hard to automate so I'm pushing both PRs as is and they'll get fuzzies. Translators will have to do the cleanup.

> If you patch the po files before the msgmerge, no new fuzzy flags should be created. That is also the logical error in @cstim's [remove-suffix.sh](https://github.com/Gnucash/gnucash/blob/stable/po/remove-suffix.sh). It would also remove before existing fuzzy flags.

If you want to avoid the fuzzy flags you have to patch the msgid too, but doing that automatically runs the risk of changing a string in cases where it wasn't changed in source. `remove-suffix.sh` uses the fuzzy to only match the msgids that `msgmerge` actually changed and where the suffix to be removed was present at the end of the msgstr, avoiding most of the non-latin script problems. But that's a very simple case and it would be dangerous to try to apply it in the adding-suffix situation because there's no programmatic way to tell if the new suffix belongs there or what code point to use.

So thanks, @CWehli, for your work here and sorry it took so long to merge.

PR #1904, opened 2024-04-03
Use correct literal suffix for `size_type`.

madpilot78:
`std::vector<T,Allocator>::size` returns `size_type`. `%ld` works fine on 64 bit arches, but fails on 32 bit.

Ref.: https://en.cppreference.com/w/cpp/container/vector/size
      https://en.cppreference.com/w/cpp/types/size_t


jralls:
Thanks!

PR #1906, opened 2024-04-06
Fix invalid-$DISPLAY error being non-text

nabijaczleweli:
Given
```
nabijaczleweli@tarta:~$ DISPLAY=komputerek-x40.:0 gnucash
Run 'g --help' to see a full list of available command line options.
Error: could not initialise graphical user interface and option add-price-quotes was not set.
Perhaps you need to set the $DISPLAY environment variable?nabijaczleweli@tarta:~$
we can obviously see the error output isn't a text file.
```

Half-mechanically updated with
```sh
sed -i '/eed to set the $DISPLAY environmen/s/?/?\\n/' $(git grep -l 'eed to set the $DISPLAY environmen')
```
and then
```sh
vi $(git grep -l 'eed to set the $DISPLAY environmen')
```

### DIFF
--- a/gnucash/gnucash.cpp
+++ b/gnucash/gnucash.cpp
@@ -296,7 +296,8 @@ main(int argc, char ** argv)
         << "\n"
         // Translators: Do not translate $DISPLAY! It is an environment variable for X11
         << _("Error: could not initialize graphical user interface and option add-price-quotes was not set.\n"
-        "Perhaps you need to set the $DISPLAY environment variable?");
+        "Perhaps you need to set the $DISPLAY environment variable?")
+        << "\n";
         return 1;
     }
 


gnucash/gnucash.cpp
```diff
@@ -296,7 +296,7 @@ main(int argc, char ** argv)
         << "\n"
         // Translators: Do not translate $DISPLAY! It is an environment variable for X11
         << _("Error: could not initialize graphical user interface and option add-price-quotes was not set.\n"
-        "Perhaps you need to set the $DISPLAY environment variable?");
+        "Perhaps you need to set the $DISPLAY environment variable?\n");
```
zopsicle: why not just `?") << "\n"`

nabijaczleweli: weird workaround instead of correct solution, breaks some existing translations

zopsicle: ok

gjanssens: What's weird about @chloekek 's suggestion ? That's perfectly normal c++ and it avoids having to add coded newline characters in translations. Note how line 296 does exactly the same thing.

As for breaking existing translations, so does your proposal :wink: albeit the exact opposite set of translations. You even added fixes for them in your PR.

I'd be tempted to go with @chloekek 's suggestion.

nabijaczleweli:
applied

fellen:
Thanks!

PR #1907, opened 2024-04-06
Fix first-day-of-week detexion in dense calendar (Upcoming Transactions)

nabijaczleweli:
GnuCash depends on GTK3, and not GTK2. Thus:
```
$ gettext -d gtk20 calendar:week_start:0; echo
calendar:week_start:0
$ LC_LANG=pl_PL.UTF-8 gettext -d gtk20 calendar:week_start:0; echo
calendar:week_start:0

$ gettext -d gtk30 calendar:week_start:0; echo
calendar:week_start:1
$ LC_LANG=pl_PL.UTF-8 gettext -d gtk30 calendar:week_start:0; echo
calendar:week_start:1
```
because the "gtk20" catalog, provided by GTK2, doesn't exist.

Before:
image

After:
image

This corresponds to https://bugs.gnucash.org/show_bug.cgi?id=797623 and  https://bugs.gnucash.org/show_bug.cgi?id=752395

### DIFF
--- a/gnucash/gnome-utils/gnc-dense-cal.c
+++ b/gnucash/gnome-utils/gnc-dense-cal.c
@@ -496,30 +496,7 @@ gnc_dense_cal_init(GncDenseCal *dcal)
     dcal->initialized = TRUE;
 
 
-    dcal->week_starts_monday = 0;
-    {
-        gchar **parts;
-        const char *week_start_str;
-
-        /* Use this renaming macro to avoid extraction of the message
-           string into the gnucash.pot file when calling xgettext. */
-#define dgettext_noextract dgettext
-        /* Translators: This string must not show up in gnucash.pot as
-           it is looked up in the "gtk20" translation domain
-           instead. */
-        week_start_str = dgettext_noextract("gtk20", "calendar:week_start:0");
-#undef dgettext_noextract
-
-        parts = g_strsplit(week_start_str, ":", 3);
-        if (parts[0] != NULL
-                && parts[1] != NULL
-                && parts[2] != NULL)
-        {
-            if (strcmp("1", parts[2]) == 0)
-                dcal->week_starts_monday = 1;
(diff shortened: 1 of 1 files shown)


christopherlam:
Would be better to use https://github.com/Gnucash/gnucash/blob/cab6173cd56db08c755a1176c3de2f0fe472e214/libgnucash/engine/gnc-date.cpp#L194 which returns 1..7 for Sun..Sat

nabijaczleweli:
Indeed; applied.

Bob-IT:
I agree that this was an over site from gtk2 to 3 but curious on how have you tested this, what OS's.

nabijaczleweli:
As evident from the top-level post, I just ran it and looked at the calendar. bookworm.

nabijaczleweli:
3-week bump

christopherlam:
It's a simple fix which looks inadequate... Do you think it's feasible to support start of week on any of the weekdays?

Bob-IT:
> It's a simple fix which looks inadequate... Do you think it's feasible to support start of week on any of the weekdays?

@christopherlam  I did look at this a while back that did allow any start day but required some more testing and refinement.
I will dig it out and create a PR for comment next week.

jralls:
It should be a localization setting. ICU has [getFirstDayOfWeek](https://github.com/unicode-org/icu/blob/69cb085e12764b84269c31be5a054e7c510be443/icu4c/source/i18n/unicode/calendar.h#L984). I found hints that `nl_langinfo` has added it in the last couple of years but it doesn't show up in the Linux docs.

nabijaczleweli:
ftr gnucash already uses getFirstDayOfWeek() for this

Bob-IT:
I have added #PR1935 which was what I was working on last year.

jralls:
Closing in favor of #1935.

PR #1909, opened 2024-04-07
Start of Gtk3 to Gtk4

Bob-IT:
I have had a look at moving from Gtk3 to 4 and thought I would push what I have done.

Starting with 5.6 built without any Aqbanking and OFX support to try and make life easier I did some of the pre-migration changes as in the first 7 commits. This was mainly around not accessing the event structure directly but by function calls.

After that I pushed Webkit out of the way and changed to `gtk version 4.8.3` and started compiling, on error I tried to replace the offending function and do a global find/replace or comment it out for later investigation. In total this was 69 commits, up to `Change gtk_init_check` commit. The next commit fixed any errors in the trace file when I tried to start a built version.

There are a lot of changes and some widgets have gone like `GtkContainer`, `GtkToolBar` and `GtkStatusBar` is on it's way out. The main layout widgets are Box and Grid with special functions for adding a child to windows and frames.

Another big change is that there is no `gtk_dialog_run` function so one would need to use a `response` callback for possibly 140 dialogs but GtkDialog is on it's way also and being replaced by a GtkWindow.

The migration document is [here](https://docs.gtk.org/gtk4/migrating-3to4.html) which shows what is likely to be involved.

I definitely have not got every thing I changed correct but it has enabled me to get a sort of working framework to test the changes and make further ones. There is still plenty to change, not least the `ui` and `glade` files. As can be seen below, there is no toolbar as that is depreciated and I have removed that section from a couple of `ui` files and commented out any reference to it.
Not sure about replacing it or maybe not, need to think about it more.
The other thing I am not sure about is the menu, `ALT+` the top first letter works but not on the actual menu or the `Ctrl+` entries. I think the latter may be fixed by the use of a `GtkShortcutController`.

The other thing worth mentioning is that I am not using `GtkApplication`, see `Change gtk_main` commit.

The last 6 commits are to do with trying to get a handle on what is needed to have a working chart of accounts page as seen below, the height of the combo is not playing ball so that may need to be changed to a `GtkDropDown`...

`One thing I would like to clarify is what version of Gtk4 are we aiming at, as stated above I am building against 4.8.3, 4.10.5 would be my next choice with more depreciated widgets and was released August last year, the latest version is 4.14.2`

gtk4-image1


jralls:
I guess it's time for a future branch to base this on.

That aside I think you might be going at this the hard way. Step 1 is to turn on deprecation warnings, still building with Gtk3, and get all of those fixed. We could conceivably merge that into stable.

I think Step 2 has to be to convert to GApplication.

Step 3 is to rewrite as many uses of removed API listed in the migration document as possible while remaining with Gtk3. For example we have to [replace GtkToolBar with a GtkBox styled as a Toolbar](https://docs.gtk.org/gtk4/migrating-3to4.html#gtktoolbar-has-been-removed).

On the subject of menu accelerators, they've [removed the old `primary` hack](https://gitlab.gnome.org/GNOME/gtk/-/issues/3421).

Is this so much work that we should consider putting it into changing GUI toolkits instead?

Gtk4 version shouldn't matter that much at this stage: Anything they deprecate after Gtk-4.0.0 won't be removed until Gtk5 so we can keep using it while we migrate, but we can do ourselves a favor going forward by trying to clean out the deprecations as part of major releases, e.g. if GC6's benchmark distro is Ubuntu 24.04LTS then whatever is in that (I suppose it will be Gtk-4.14.0) then we should fix all deprecations to that Gtk version.

Bob-IT:
> Step 1 is to turn on deprecation warnings, still building with Gtk3, and get all of those fixed. We could conceivably merge that into stable.

I did do this first and if I did it correctly there are none in GnuCash. The only one it did find was in WebKit where they are still using GtkAction.
 
> I think Step 2 has to be to convert to GApplication.

Would need to think about that...

> Step 3 is to rewrite as many uses of removed API listed in the migration document as possible while remaining with Gtk3. For example we have to [replace GtkToolBar with a GtkBox styled as a Toolbar](https://docs.gtk.org/gtk4/migrating-3to4.html#gtktoolbar-has-been-removed).

I saw that and will have a think about what that would involve. Maybe I misunderstood some of the prep work or just thought I would see what building with gtk4 would throw up and then fix it.

> Is this so much work that we should consider putting it into changing GUI toolkits instead?

I do not know, I would like to try/fix a few more changes.

jralls:
> I saw that and will have a think about what that would involve. Maybe I misunderstood some of the prep work or just thought I would see what building with gtk4 would throw up and then fix it.

No harm (other than time expended) in experimenting. Looks like we'll be doing a lot of that in this migration. My thought on getting as much as possible done in Gtk3 is to reduce the variables at play as we rewrite the code.

> I do not know, I would like to try/fix a few more changes.

OK.

christopherlam:
Whoops sorry didn't see these new conflicts dialog-commodity.c[pp] and gnc-currency-edit.c[pp]

Bob-IT:
> Whoops sorry didn't see these new conflicts dialog-commodity.c[pp] and gnc-currency-edit.c[pp]

No worries, I will fix when I next push, just carry on committing as normal.

fellen:
> built without any Aqbanking

That would require a GTK4 module  in 
https://github.com/aqbanking/gwenhywfar/tree/master/gui

jralls:
>  That would require a GTK4 module in https://github.com/aqbanking/gwenhywfar/tree/master/gui

Which we'll probably have to do and submit as a PR.

gjanssens:
> > I think Step 2 has to be to convert to GApplication.
> 
> Would need to think about that...
> 
This was what I was studying when I ran out of time for gnucash. I had my thoughts on how to go forward with this, but hadn't tried anything in code just yet.

I still would like to work this out and may find some time in the coming weeks. Can you leave the GApplication conversion part aside for me ?

Bob-IT:
> Can you leave the GApplication conversion part aside for me ?

Will do, I am just looking at conversion of dialogs and some custom widgets.

jralls:
@gjanssens One thing to consider for migrating to GApplication is timing of loading the book in relation to starting the event loop. As you know a long-running complaint from Mac users is that they can't double-click a data file to get GnuCash to open it because macOS uses an event to tell the app what to load and GnuCash loads the last-used file before it can handle events.

gjanssens:
> @gjanssens One thing to consider for migrating to GApplication is timing of loading the book in relation to starting the event loop. As you know a long-running complaint from Mac users is that they can't double-click a data file to get GnuCash to open it because macOS uses an event to tell the app what to load and GnuCash loads the last-used file before it can handle events.

Indeed. I don't know the insides yet, but from what I remember from my reading so far is that in the GApplication world the request to open a file comes in in the form of an event. I don't know how the MacOS backend implements this, but I was hoping on macos such an event would be generated as well when the user clicks on a GnuCash data book.

So my idea here was to process the pending event queue at the appropriate moment early on in the application startup and once empty check whether or no a file has been opened. If not, open the last opened file as found in our history list.

Would that work ?

jralls:
Should do. Here's how gtk-mac-integration (the Gtk2/3 library that preceded GtkApplication) [handles it](https://gitlab.gnome.org/GNOME/gtk-mac-integration/-/blame/master/src/GtkApplicationDelegate.c?ref_type=heads#L49). I'm not as familiar with the GApplication implementation but I expect that it would be pretty similar.

There is, BTW, a [UX regression](https://gitlab.gnome.org/GNOME/gtk/-/issues/3421) for macOS with Gtk4.

Bob-IT:
I have been trying some changes on and off as seen here, the first image shows a right mouse menu and the next shows the column menu. The text size can be controlled via CSS but on first looks could not see how to change the height/spacing but that will come to me...

gtk4-image2
gtk4-image3

You will also notice I have added a toolbar, seen below. This image shows the difference in named icon image sizes, most of them are `large` but the last is `normal`, the actual size I think is theme based but you can set the size in pixels. This was fairly easy to do with some modification to the existing `ui` files, each button holds a vertical box packed with an image and label and sizes can be controlled by CSS. There is nothing fancy here so the Gnucash window may increase if there many buttons.

gtk4-image5

The other area I have been looking at is the tools menu, I have done a first attempt at converting those dialogs to windows, still modal but possibly returning values via callbacks.

There are still many changes required moving to gtk4 but am getting some messages in the trace file which appear to of been fixed in the 4.10 series.

My next attempt at changes will be possibly changing a GtkCombo to a GtkDropDown or the dense calendar/register as they use a lot of cairo. If that is successful, I think I will up my gtk4 version to 4.10.5 and see what that shows up.

jralls:
@Bob-IT Nice work. Do you mind if I push a fix to the CI?

Bob-IT:
> @Bob-IT Nice work. Do you mind if I push a fix to the CI?

For stable, yes just do it, for this PR I am building with just `-Werror -DGDK_DISABLE_DEPRECATED -DGTK_DISABLE_DEPRECATED`. I commented out several calls to functions and did not want the extra hassle of fixing variables/function defined but not used. I will of cause change that to the original parameters and fix any issues.

jralls:
> For stable, yes just do it,

I meant for this PR: https://github.com/jralls/gnucash/commit/e4bdad80e354cbd5883bf7adf5910d4daf3dc50c

It turns out that the current Ubuntu-latest image had gtk-4.6 and of course there's other work to get it to build on macOS so I turned off everything except the Arch Linux docker test.

jralls:
> On the subject of menu accelerators, they've [removed the old primary hack](https://gitlab.gnome.org/GNOME/gtk/-/issues/3421).

Progress on this front, of a sort: https://gitlab.gnome.org/GNOME/gtk/-/merge_requests/7260 in which [Emmanuele Bassi explains](https://gitlab.gnome.org/GNOME/gtk/-/merge_requests/7260#note_2117155) that for the most part it's the application's responsibility to provide the common keyboard shortcuts. Gtk4 only provides shortcuts for its own widgets. [I derive](https://gitlab.gnome.org/GNOME/gtk/-/merge_requests/7260#note_2117205) from that that apps can pretty easily create alternate versions and Arjen Molenaar redid the MR to provide the alternate versions in Gtk4. 

I don't think Emmanuele or Matthias Clasen would go for a macro but nothing keeps us from doing one:
```
#ifdef __APPLE__ 
#define GNC_PRIMARY_MASK GDK_META_MASK
#else 
#define GNC_PRIMARY_MASK GDK_CONTROL_MASK 
#endif
```

Bob-IT:
This is just a rebase of the existing commits onto 5.11 with some added commits that I had locally but had not pushed before I ran out of time.

Below is a picture of the summary bar after a first attempt at changing it to a GtkDropdown. As you can see it fixes previous issue and also the alignment issue when expanded.
gtk4-image6

The picture below is an attempt to change the dense calendar which seems to work but with an occasional lockup issue with the popup when in motion, the month selection is also a GtkDropdown.
gtk4-image7

The other thing of note is a first attempt at changing a GtkTreeView to a GtkColumnview for the doc link dialog.

The last thing I was looking at was trying to get the register to display so will try and do that time permitting.

jralls:
Fixed a conflict from #2076.

The test harnesses need to be updated to pull in Gtk4 instead of 3.

copystring:
I have been continuing the GTK4 preparation work by replacing removed APIs while retaining GTK3-compatible paths.

I understand the preparation patches intended for `stable` should continue to build with GTK3. The open question is only the eventual GTK4 future branch: once the GApplication work and preparatory changes are in place, should that branch remain dual-toolkit, or should it become GTK4-only with GTK3 remaining on the legacy maintenance line?

Knowing that boundary would help keep the preparation work structured correctly and avoid carrying compatibility branches into the future code unnecessarily.

Bob-IT:
> I have been continuing the GTK4 preparation work by replacing removed APIs while retaining GTK3-compatible paths.
> 
> I understand the preparation patches intended for `stable` should continue to build with GTK3. The open question is only the eventual GTK4 future branch: once the GApplication work and preparatory changes are in place, should that branch remain dual-toolkit, or should it become GTK4-only with GTK3 remaining on the legacy maintenance line?
> 
> Knowing that boundary would help keep the preparation work structured correctly and avoid carrying compatibility branches into the future code unnecessarily.

@copystring 
This was a test to see what could be done and possible changes that were required. It also gave some exposure of the new GTK4 way of doing things. I think I may of done some further changes locally, really should rebase this to reflect current changes

There are some changes that could be applied to `stable` as a start as some GTK4 functions exist in GTK3 namely some event/controller access functions and possibly some other functions have been 'back ported'/exist now with the current required GTK3 version 3.22.30. Like wise some GTK3 functions exist in GTK4 but are marked as depreciated which will take a bit longer to convert.

Going forward, I think a branch of only GTK4 is the way to go setting a lower required version and progressively updating it. I started at 4.8.3 upped it to 4.10.5 and may be locally it is higher. Not sure what the current version of GTK4 will be on supported platforms when this is live.

jralls:
@Bob-IT is there a backwards-compatibility issue that would justify making changes on stable? If not I don't see any reason to do so: We have a future branch now, which we didn't when you made the PR.

We really need to plan out what we need to change to prepare. Implementing GApplication is surely a big one. We can reduce the amount of conversion work by going through the Gtk3 code and looking for business logic that belongs in libgnucash: One instance that I know about is the transfer dialog but I'm sure there are others.

PR #1910, opened 2024-04-09
gnc_commodity_table_get_namespaces returns a vector

christopherlam:
`gnc_commodity_table_get_namespaces` currently returns a `GList*` which isn't always released. There's more leaks similar to 4f5ee5dc4e. Instead of hunting down leaks, forward-proof it to return a `std::vector`. ~Should it return a `std::vector<std::string>`?~

~Ditto `gnc_commodity_table_get_commodities` isn't always being released properly.~ postponed to later branch

### DIFF
--- a/gnucash/gnome-utils/dialog-commodity.cpp
+++ b/gnucash/gnome-utils/dialog-commodity.cpp
@@ -429,15 +429,6 @@ gnc_ui_select_commodity_namespace_changed_cb (GtkComboBox *cbwe,
 /********************************************************************
  * gnc_ui_update_commodity_picker
  ********************************************************************/
-static int
-collate(gconstpointer a, gconstpointer b)
-{
-    if (!a)
-        return -1;
-    if (!b)
-        return 1;
-    return g_utf8_collate (static_cast<const char*>(a), static_cast<const char*>(b));
-}
 
 
 void
@@ -445,15 +436,13 @@ gnc_ui_update_commodity_picker (GtkWidget *cbwe,
                                 const gchar * name_space,
                                 const gchar * init_string)
 {
-    GList      * iterator = nullptr;
-    GList      * commodity_items = nullptr;
+    std::vector<std::string> commodity_items;
     GtkComboBox *combo_box;
     GtkEntry *entry;
     GtkTreeModel *model;
     GtkTreeIter iter;
     gnc_commodity_table *table;
     gint current = 0, match = 0;
-    gchar *name;
 
     g_return_if_fail(GTK_IS_COMBO_BOX(cbwe));
     g_return_if_fail(name_space);
@@ -470,25 +459,23 @@ gnc_ui_update_commodity_picker (GtkWidget *cbwe,
     gtk_combo_box_set_active(combo_box, -1);
 
     table = gnc_commodity_table_get_table (gnc_get_current_book ());
+
     for (auto comm : gnc_commodity_table_get_commodities(table, name_space))
-    {
-        commodity_items = g_list_prepend (commodity_items, (gpointer) gnc_commodity_get_printname(comm));
-    }
+        commodity_items.push_back (gnc_commodity_get_printname(comm));
 
-    commodity_items = g_list_sort(commodity_items, collate);
-    for (iterator = commodity_items; iterator; iterator = iterator->next)
+    std::sort (commodity_items.end(), commodity_items.begin());
+
+    for (auto name : commodity_items)
     {
-        name = (char *)iterator->data;
         gtk_list_store_append(GTK_LIST_STORE(model), &iter);
-        gtk_list_store_set (GTK_LIST_STORE(model), &iter, 0, name, -1);
+        gtk_list_store_set (GTK_LIST_STORE(model), &iter, 0, name.c_str(), -1);
 
-        if (init_string && g_utf8_collate(name, init_string) == 0)
+        if (init_string && name == init_string)
             match = current;
         current++;
     }
 
     gtk_combo_box_set_active(combo_box, match);
-    g_list_free(commodity_items);
 }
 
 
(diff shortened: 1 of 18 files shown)


gnucash/gnome-utils/dialog-commodity.cpp
```diff
@@ -470,25 +459,23 @@ gnc_ui_update_commodity_picker (GtkWidget *cbwe,
...
     {
-        name = (char *)iterator->data;
         gtk_list_store_append(GTK_LIST_STORE(model), &iter);
         gtk_list_store_set (GTK_LIST_STORE(model), &iter, 0, name, -1);
 
-        if (init_string && g_utf8_collate(name, init_string) == 0)
+        if (init_string && name == init_string)
             match = current;
         current++;
     }
```
jralls: Might as well use 
```
if (init_string)
{
    const std::string findme{init_string};
    auto iter = std::find(commodity_items.begin(), commodity_items.end(), init_string);
    if (iter != commodity_items.end())
        match = iter - commodity_items.begin();
}
```
And clear out the resulting cruft.

libgnucash/engine/gnc-commodity.cpp
```diff
@@ -1853,7 +1853,7 @@ gnc_commodity_table_remove(gnc_commodity_table * table,
     if (!nsp) return;
 
     nsp->cm_list = g_list_remove(nsp->cm_list, comm);
-    g_hash_table_remove (nsp->cm_table, priv->mnemonic);
+    nsp->cm_table.erase (priv->mnemonic);
     /* XXX minor mem leak, should remove the key as well */
```
jralls, line 1882: Remove the comment. There's no function to delete the key if `erase` deletes only the value, and the unordered_map probably won't deallocate memory anyway, it will keep it around for reuse.

libgnucash/engine/gnc-commodity.cpp
```diff
@@ -2184,8 +2158,9 @@ gnc_commodity_table_delete_namespace(gnc_commodity_table * table,
     g_list_free(ns->cm_list);
     ns->cm_list = nullptr;
 
-    g_hash_table_foreach_remove(ns->cm_table, ns_helper, nullptr);
-    g_hash_table_destroy(ns->cm_table);
+    std::for_each (ns->cm_table.begin(), ns->cm_table.end(),
+                   [](auto it){ gnc_commodity_destroy (it.second); });
+    StrCommodityMap().swap(ns->cm_table);
```
jralls: Sigh. Not your fault but this function should call only 
```
{
    gnc_commodity_namespace *ns = gnc_commodity_table_find_namespace(table, name_space);
    g_object_unref(ns);
}
```
The cleanup stuff--removing ns from `table->ns_table` and `table->ns_list` and clearing the StrCommodityMap should be in `gnc_commodity_namespace_dispose_real()`. Call `ns->cm_table.~StrCommodityMap()` in `gnc_commodity_namespace_finalize_real()`. ~If you use placement new for constructing the StrCommodityMap then you can call `delete ns->cm_table` instead, which is all-around cleaner.~ *Edit* No, I remember now that you still have to use `~StrCommodityMap()` with placement new.

There's no point to having GObject memory management if we don't bleeping use it.

Better still of course would be to rewrite `gnc_commodity_namespace` as a proper `GncCommodityNamespace` C++ class, but that's probably more work than you want to do at this point.

libgnucash/engine/gnc-commodity.cpp
```diff
@@ -1592,7 +1593,7 @@ gnc_commodity_table *
 gnc_commodity_table_new(void)
 {
     gnc_commodity_table * retval = g_new0(gnc_commodity_table, 1);
-    retval->ns_table = g_hash_table_new(&g_str_hash, &g_str_equal);
+    StrCommodityNSMap().swap (retval->ns_table);
```
jralls: As for StrCommodityMap, better to use placement new.

libgnucash/engine/gnc-commodity.cpp
```diff
@@ -2114,7 +2097,7 @@ gnc_commodity_table_add_namespace(gnc_commodity_table * table,
     if (!ns)
     {
         ns = static_cast<gnc_commodity_namespace*>(g_object_new(GNC_TYPE_COMMODITY_NAMESPACE, nullptr));
-        ns->cm_table = g_hash_table_new(g_str_hash, g_str_equal);
+        StrCommodityMap().swap(ns->cm_table);
```
jralls: Risky. I haven't looked at the standard but [c++ reference](Swapping with a default-constructed temporary doesn't say anything about swapping to 0-filled memory running the constructor on that memory.
Better to use placement new to be sure that the default constructor is called.

libgnucash/engine/gnc-commodity.cpp
```diff
@@ -2242,8 +2242,7 @@ gnc_commodity_table_destroy(gnc_commodity_table * t)
 
     g_list_free(t->ns_list);
     t->ns_list = nullptr;
-    g_hash_table_destroy(t->ns_table);
-    t->ns_table = nullptr;
+    StrCommodityNSMap().swap(t->ns_table);
```
jralls: Again, placement new allows operator delete here.

jralls (changes requested):
Maybe we should split base-typemaps.i into base-typemaps-guile.i and base-typemaps-python.i.

libgnucash/engine/gnc-commodity.cpp
christopherlam: > As for StrCommodityMap, better to use placement new.

or `auto retval = new gnc_commodity_table;` which will properly initialize the map.

jralls: It will, but you'll need an explicit `~gnc_commodity_table()` to clean up the GList. Note that mixing `new` with `malloc` like that is considered risky. The compiler will probably do what you want but it doesn't have to.

christopherlam:
This branch getting rather crowded. I wonder if it would be _safer_ to push well-tested commits one by one. The aim is still to use c++ classes.

christopherlam:
last commit is fake... don't know how to fix python bindings.

PR #1913, opened 2024-04-10
[gnc-ui-utils.cpp] use icu::RBNF to convert numbers to long strings

christopherlam:
result:

Number 123456789.010000 in words: one hundred twenty-three million four hundred fifty-six thousand seven hundred eighty-nine point zero one

Number 12.000000 in words: 十二

Number 123456789.010000 in words: 一億二千三百四十五万六千七百八十九・〇一

Number 2345.23 in words: deux mille trois cent quarante-cinq virgule deux trois

Number 2345.23 in words: 二千三百四十五・二三

### DIFF
--- a/bindings/app-utils.i
+++ b/bindings/app-utils.i
@@ -81,7 +81,6 @@ GNCPrintAmountInfo gnc_price_print_info (const gnc_commodity *curr,
 GNCPrintAmountInfo gnc_share_print_info_places (int decplaces);
 const char * xaccPrintAmount (gnc_numeric val, GNCPrintAmountInfo info);
 
-gchar *number_to_words(gdouble val, gint64 denom);
 const gchar *printable_value (gdouble val, gint denom);
 
 gboolean gnc_reverse_balance (const Account *account);
--- a/libgnucash/app-utils/gnc-ui-util.cpp
+++ b/libgnucash/app-utils/gnc-ui-util.cpp
@@ -1500,18 +1500,41 @@ gnc_wrap_text_with_bidi_ltr_isolate (const char* text)
 /********************************************************************\
  ********************************************************************/
 
+using StrStrVec = std::vector<std::pair<std::string, std::string>>;
+
+static std::string
+string_replace_substring (const std::string& input, const StrStrVec& replacements)
+{
+    std::string result = input;
+    for (const auto& [key, val] : replacements)
+        for (auto pos = result.find(key); pos != std::string::npos; pos = result.find(key))
+            result.replace(pos, key.length(), val);
+    return result;
+}
+
 static std::string
 number_to_words(double val, int64_t denom)
 {
     double int_part;
     const int frac_part = std::round(std::modf (std::fabs(val), &int_part) * denom);
-    const std::vector<std::string> tail =
-        { " ", _("and"), " ", std::to_string (frac_part), "/", std::to_string (denom) };
     std::ostringstream ss;
-
     ss.imbue(gnc_get_boost_locale());
-    ss << boost::locale::as::spellout << int_part;
-    return std::accumulate (tail.begin(), tail.end(), ss.str());
+    auto num_to_words = [&ss](auto num)
+    {
+        ss.str("");
+        ss << boost::locale::as::spellout << num;
+        return ss.str();
+    };
+    StrStrVec replacements = {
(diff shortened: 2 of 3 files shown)


jralls:
Is there a use-case for this other than printing checks? Does anybody besides Americans write out the check's value in words as well as write the numerals? (Back in 1972 when my family went to England for 4 months my mom got a bank account there and wrote checks. I think they were pretty much like American ones, but that was more than 50 years ago.)

When writing a check in the USA we write (using your shorter example) "Two thousand three hundred forty-five and 23/100". "Dollars" is pre-printed at the end of the line. 
I've never encountered anyone writing out "point", though it would make sense if the string was destined for an audio function that was going to speak it.

christopherlam:
rewritten in the _much_ nicer boost::locale

christopherlam:
https://www.co-operativebank.co.uk/help-and-support/payments/cheques/

They advocate writing "only" irrespective of the amount...

jralls:
And also writing out "pounds", "and" and "pence", with the number of pence written out as words, rather different from the custom in the USA. How do you write the amount in Australia?
@fellen How do you do it in Germany?

Incidentally some in the US recommend when handwriting checks to draw a line from the fraction to the end if there's space, for example https://www.thebalancemoney.com/how-to-write-a-check-4019395, though only one of the half-dozen or so "how to write a check" sites I looked at did that, but all of them inserted "and" before the fraction.

It seems we should create a US format string and make it translatable with instructions on the placeholders in a translator comment.

christopherlam:
@jralls see word substitution for check printing

jralls:
@christopherlam That's not localized, and what's the point of concocting a private formatting function? Use [`boost::locale::format`](https://www.boost.org/doc/libs/1_84_0/libs/locale/doc/html/localized_text_formatting.html).

There's a much better formatting and I/O library, [`{fmt}`](https://github.com/fmtlib/fmt) that's being piecewise added to the C++ standard as `std::format` beginning with C++20, but both inclusion and implementation in libc++ and libstdc++ have been slow. We should aim towards replacing boost::locale::format with it in the near term and all use of std::iostreams in the longer term.

christopherlam:
I didn't use boost format because it requires that all format variables are used. And I felt it would be better to have more flexibility, allowing eg integer words, integer numbers, fraction words, fraction numbers as options.

jralls:
OK, but that doesn't do any good unless you mark the string for translation and document the construction thoroughly for translators with a translator comment. Not only can most translators not understand code, they can't even *see* it in Weblate without significant effort.

christopherlam:
The format string would be best IMHO be a global preference rather than a translatable string. I figure most users would prefer being able to fine tune rather than rely on random translators. I felt this branch would be a temporary until we plug into the preference mechanism.

jralls:
> The format string would be best IMHO be a global preference rather than a translatable string. I figure most users would prefer being able to fine tune rather than rely on random translators. I felt this branch would be a temporary until we plug into the preference mechanism.

OK, that might be true at least for some users. Temporary tends to stay for a long time around here so at least comment it thoroughly including a FIXME or TODO about creating a preference.

It does occur to me that @cstim commented about the untranslatability 17 years ago and there are [no bugs](https://bugs.gnucash.org/buglist.cgi?bug_status=NEW&bug_status=ASSIGNED&bug_status=NEEDINFO&bug_status=REOPENED&component=Check%20Printing&list_id=21335&query_format=advanced) about it so maybe only English speakers (maybe even only Americans) use the check printing and there's no actual point to localizing it. Maybe you should add one as an enhancement request to create another reminder to add the preference.

cstim:
Yes, I would claim this isn't used in any other country anyway. Hence you can quite safely ignore the translatability here.

jralls:
> Yes, I would claim this isn't used in any other country anyway. Hence you can quite safely ignore the translatability here.

Hmm, it kind of follows that localizing the number-strings doesn't make sense either and it should be just left alone.

jralls:
@christopherlam Do you still want to pursue this?

christopherlam:
> @christopherlam Do you still want to pursue this?

This branch aims to use ICU instead of hand rolling the number->string routine... It could be limited to English words and two decimal places only if you like

jralls:
The problem is that we have a particular use for that conversion and the ICU wording doesn't fit it.

PR #1914, opened 2024-04-12
Bug 799281 - Deleting a transaction may trigger a crash

agwekixj:
Update qof_collection_lookup_entity() to prevent returning instances marked to be destroyed.


jralls:
Thanks!

PR #1915, opened 2024-04-13
account->splits is a vector

christopherlam:
and `xaccAccountGetSplitList` still returns a `GList*`.

compared to #1887 the best benchmarks are now 3.491s

### DIFF
--- a/gnucash/gnome-utils/dialog-transfer.cpp
+++ b/gnucash/gnome-utils/dialog-transfer.cpp
@@ -44,6 +44,7 @@
 #include "gnc-ui.h"
 #include "Transaction.h"
 #include "Account.h"
+#include "Account.hpp"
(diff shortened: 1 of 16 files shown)


christopherlam:
too soon

jralls:
> too soon

Yeah, it's harder than it might seem. Too much exposed implementation and too much spaghetti in general.

Do your benchmarking tests include any sorting?

christopherlam:
> > too soon
> 
> Yeah, it's harder than it might seem. Too much exposed implementation and too much spaghetti in general.
> 
> Do your benchmarking tests include any sorting?

Huh the main benchmark is loading my primary data file. The sorting takes place automatically periodically. I think I'm nearly done, just need to expose `const vector`, the `vector.size()` and the `GList*` which must be freed.

PR #1917, opened 2024-04-14
Account splits vector

christopherlam:
#1915 for review. ~`test-sx` is still failing, not sure why.~

best benchmarks now 3.477s

### DIFF
--- a/gnucash/gnome-utils/dialog-transfer.cpp
+++ b/gnucash/gnome-utils/dialog-transfer.cpp
@@ -44,6 +44,7 @@
 #include "gnc-ui.h"
 #include "Transaction.h"
 #include "Account.h"
+#include "Account.hpp"
 #include "engine-helpers.h"
 #include "QuickFill.h"
 #include <gnc-commodity.h>
@@ -472,10 +473,8 @@ gnc_xfer_dialog_reload_quickfill( XferDialog *xferData )
     gnc_quickfill_destroy( xferData->qf );
     xferData->qf = gnc_quickfill_new();
 
-    auto splitlist = xaccAccountGetSplitList( account );
-    for ( GList *node = splitlist; node; node = node->next )
+    for (auto split : xaccAccountGetSplits (account))
     {
-        auto split = static_cast<Split *> (node->data);
         auto trans = xaccSplitGetParent (split);
         gnc_quickfill_insert( xferData->qf,
                               xaccTransGetDescription (trans), QUICKFILL_LIFO);
--- a/gnucash/gnome/gnc-plugin-page-account-tree.cpp
+++ b/gnucash/gnome/gnc-plugin-page-account-tree.cpp
@@ -39,6 +39,7 @@
 #include "gnc-plugin-page-account-tree.h"
 #include "gnc-plugin-page-register.h"
 
+#include "Account.hpp"
 #include "Scrub.h"
 #include "Scrub3.h"
 #include "ScrubBusiness.h"
@@ -1145,22 +1146,19 @@ static gpointer
 delete_account_helper (Account * account, gpointer data)
 {
     auto helper_res = static_cast<delete_helper_t*>(data);
-    GList *splits;
+    auto splits{xaccAccountGetSplits (account)};
 
-    splits = xaccAccountGetSplitList (account);
-    if (splits)
+    if (!splits.empty())
     {
         helper_res->has_splits = TRUE;
-        while (splits)
+        for (auto s : splits)
         {
-            auto s = GNC_SPLIT(splits->data);
             Transaction *txn = xaccSplitGetParent (s);
             if (xaccTransGetReadOnly (txn))
             {
                 helper_res->has_ro_splits = TRUE;
                 break;
             }
-            splits = splits->next;
         }
     }
 
@@ -1383,7 +1381,6 @@ account_delete_dialog (Account *account, GtkWindow *parent, Adopters* adopt)
     gchar *title = NULL;
     GtkBuilder *builder = gtk_builder_new();
     gchar *acct_name = gnc_account_get_full_name(account);
-    GList* splits = xaccAccountGetSplitList(account);
     GList* filter = g_list_prepend(NULL, (gpointer)xaccAccountGetType(account));
 
     if (!acct_name)
@@ -1416,7 +1413,7 @@ account_delete_dialog (Account *account, GtkWindow *parent, Adopters* adopt)
                   account, FALSE);
 
     // Does the selected account have splits
-    if (splits)
+    if (!xaccAccountGetSplits(account).empty())
     {
         delete_helper_t delete_res2 = { FALSE, FALSE };
 
@@ -1537,8 +1534,7 @@ gnc_plugin_page_account_tree_cmd_delete_account (GSimpleAction *simple,
     }
 
     // If no transaction or children just delete it.
-    if (!(xaccAccountGetSplitList (account) != NULL ||
-          gnc_account_n_children (account)))
+    if (xaccAccountGetSplits (account).empty() && gnc_account_n_children (account) == 0)
     {
         do_delete_account (account, NULL, NULL, NULL);
         return;
@@ -1581,7 +1577,6 @@ confirm_delete_account (GSimpleAction *simple, GncPluginPageAccountTree *page,
                         delete_helper_t delete_res)
 {
     Account *account = gnc_plugin_page_account_tree_get_current_account (page);
-    GList* splits = xaccAccountGetSplitList(account);
     GtkWidget* window = gnc_plugin_page_get_window(GNC_PLUGIN_PAGE(page));
     gint response;
 
@@ -1595,7 +1590,7 @@ confirm_delete_account (GSimpleAction *simple, GncPluginPageAccountTree *page,
                                 acct_name);
     g_free(acct_name);
 
-    if (splits)
+    if (!xaccAccountGetSplits (account).empty())
     {
         if (ta)
         {
--- a/libgnucash/engine/Account.cpp
+++ b/libgnucash/engine/Account.cpp
@@ -34,6 +34,7 @@
 #include <string.h>
 
 #include "AccountP.hpp"
+#include "Account.hpp"
 #include "Split.h"
 #include "Transaction.h"
 #include "TransactionP.h"
@@ -331,7 +332,7 @@ gnc_account_init(Account* acc)
     priv->lower_balance_limit = {};
     priv->include_sub_account_balances = {};
 
-    priv->splits = nullptr;
+    priv->splits.clear();
     priv->sort_dirty = FALSE;
 }
 
@@ -1136,6 +1137,29 @@ xaccInitAccount (Account * acc, QofBook *book)
 /********************************************************************\
 \********************************************************************/
 
+Split*
+gnc_account_find_split (const Account *acc, std::function<bool(const Split*)> predicate,
+                        bool reverse)
+{
+    if (!GNC_IS_ACCOUNT (acc))
+        return nullptr;
+
+    auto splits{GET_PRIVATE(acc)->splits};
+    if (reverse)
+    {
+        auto latest = std::find_if(splits.rbegin(), splits.rend(), predicate);
+        return (latest == splits.rend()) ? nullptr : *latest;
+    }
+    else
+    {
+        auto earliest = std::find_if(splits.begin(), splits.end(), predicate);
+        return (earliest == splits.end()) ? nullptr : *earliest;
+    }
+}
+
+/********************************************************************\
+\********************************************************************/
+
 QofBook *
 gnc_account_get_book(const Account *account)
 {
@@ -1368,22 +1392,18 @@ xaccFreeAccount (Account *acc)
     /* NB there shouldn't be any splits by now ... they should
      * have been all been freed by CommitEdit().  We can remove this
      * check once we know the warning isn't occurring any more. */
-    if (priv->splits)
+    if (!priv->splits.empty())
     {
-        GList *slist;
         PERR (" instead of calling xaccFreeAccount(), please call\n"
               " xaccAccountBeginEdit(); xaccAccountDestroy();\n");
 
         qof_instance_reset_editlevel(acc);
 
-        slist = g_list_copy(priv->splits);
-        for (lp = slist; lp; lp = lp->next)
+        for (auto s : priv->splits)
         {
-            Split *s = (Split *) lp->data;
             g_assert(xaccSplitGetAccount(s) == acc);
             xaccSplitDestroy (s);
         }
-        g_list_free(slist);
 /* Nothing here (or in xaccAccountCommitEdit) nullptrs priv->splits, so this asserts every time.
         g_assert(priv->splits == nullptr);
 */
@@ -1416,6 +1436,7 @@ xaccFreeAccount (Account *acc)
     priv->type = ACCT_TYPE_NONE;
     gnc_commodity_decrement_usage_count(priv->commodity);
     priv->commodity = nullptr;
+    priv->splits.clear();
 
     priv->balance_dirty = FALSE;
     priv->sort_dirty = FALSE;
@@ -1483,7 +1504,6 @@ xaccAccountCommitEdit (Account *acc)
     priv = GET_PRIVATE(acc);
     if (qof_instance_get_destroying(acc))
     {
-        GList *lp, *slist;
         QofCollection *col;
 
         qof_instance_increase_editlevel(acc);
@@ -1500,18 +1520,12 @@ xaccAccountCommitEdit (Account *acc)
            themselves will be destroyed by the transaction code */
         if (!qof_book_shutting_down(book))
         {
-            slist = g_list_copy(priv->splits);
-            for (lp = slist; lp; lp = lp->next)
-            {
-                Split *s = static_cast<Split *>(lp->data);
+            for (auto s : priv->splits)
                 xaccSplitDestroy (s);
-            }
-            g_list_free(slist);
         }
         else
         {
-            g_list_free(priv->splits);
-            priv->splits = nullptr;
+            priv->splits.clear();
         }
 
         /* It turns out there's a case where this assertion does not hold:
@@ -1528,7 +1542,7 @@ xaccAccountCommitEdit (Account *acc)
             qof_collection_foreach(col, destroy_pending_splits_for_account, acc);
 
             /* the lots should be empty by now */
-            for (lp = priv->lots; lp; lp = lp->next)
+            for (auto lp = priv->lots; lp; lp = lp->next)
             {
                 GNCLot *lot = static_cast<GNCLot*>(lp->data);
                 gnc_lot_destroy (lot);
@@ -1822,37 +1836,23 @@ xaccAccountEqual(const Account *aa, const Account *ab, gboolean check_guids)
     /* no parent; always compare downwards. */
 
     {
-        GList *la = priv_aa->splits;
-        GList *lb = priv_ab->splits;
-
-        if ((la && !lb) || (!la && lb))
+        if (priv_aa->splits.size() != priv_ab->splits.size())
         {
-            PWARN ("only one has splits");
-            return FALSE;
+            PWARN ("number of splits differs");
+            return false;
         }
 
-        if (la && lb)
+        for (auto it_a = priv_aa->splits.begin(), it_b = priv_ab->splits.begin();
+             it_a != priv_aa->splits.end() && it_b != priv_ab->splits.end();
+             ++it_a, ++it_b)
         {
-            /* presume that the splits are in the same order */
-            while (la && lb)
+            Split *sa = *it_a;
+            Split *sb = *it_b;
+
+            if (!xaccSplitEqual(sa, sb, check_guids, TRUE, FALSE))
             {
-                Split *sa = (Split *) la->data;
-                Split *sb = (Split *) lb->data;
-
-                if (!xaccSplitEqual(sa, sb, check_guids, TRUE, FALSE))
-                {
-                    PWARN ("splits differ");
-                    return(FALSE);
-                }
-
-                la = la->next;
-                lb = lb->next;
-            }
-
-            if ((la != nullptr) || (lb != nullptr))
-            {
-                PWARN ("number of splits differs");
-                return(FALSE);
+                PWARN ("splits differ");
+                return false;
             }
         }
     }
@@ -1922,30 +1922,29 @@ gboolean gnc_account_get_defer_bal_computation (Account *acc)
 /********************************************************************\
 \********************************************************************/
 
+static bool split_cmp_less (const Split* a, const Split* b)
+{
+    return xaccSplitOrder (a, b) < 0;
+}
+
 gboolean
 gnc_account_insert_split (Account *acc, Split *s)
 {
     AccountPrivate *priv;
-    GList *node;
 
     g_return_val_if_fail(GNC_IS_ACCOUNT(acc), FALSE);
     g_return_val_if_fail(GNC_IS_SPLIT(s), FALSE);
 
     priv = GET_PRIVATE(acc);
-    node = g_list_find(priv->splits, s);
-    if (node)
+    if (std::find (priv->splits.begin(), priv->splits.end(), s) != priv->splits.end())
         return FALSE;
 
+    priv->splits.push_back (s);
+
     if (qof_instance_get_editlevel(acc) == 0)
-    {
-        priv->splits = g_list_insert_sorted(priv->splits, s,
-                                            (GCompareFunc)xaccSplitOrder);
-    }
+        std::sort (priv->splits.begin(), priv->splits.end(), split_cmp_less);
     else
-    {
-        priv->splits = g_list_prepend(priv->splits, s);
-        priv->sort_dirty = TRUE;
-    }
+        priv->sort_dirty = true;
 
     //FIXME: find better event
     qof_event_gen (&acc->inst, QOF_EVENT_MODIFY, nullptr);
@@ -1962,17 +1961,17 @@ gboolean
 gnc_account_remove_split (Account *acc, Split *s)
 {
     AccountPrivate *priv;
-    GList *node;
 
     g_return_val_if_fail(GNC_IS_ACCOUNT(acc), FALSE);
     g_return_val_if_fail(GNC_IS_SPLIT(s), FALSE);
 
     priv = GET_PRIVATE(acc);
-    node = g_list_find(priv->splits, s);
-    if (nullptr == node)
+
+    auto it = std::remove (priv->splits.begin(), priv->splits.end(), s);
+    if (it == priv->splits.end())
         return FALSE;
 
-    priv->splits = g_list_delete_link(priv->splits, node);
+    priv->splits.erase (it, priv->splits.end());
     //FIXME: find better event type
     qof_event_gen(&acc->inst, QOF_EVENT_MODIFY, nullptr);
     // And send the account-based event, too
@@ -1993,7 +1992,7 @@ xaccAccountSortSplits (Account *acc, gboolean force)
     priv = GET_PRIVATE(acc);
     if (!priv->sort_dirty || (!force && qof_instance_get_editlevel(acc) > 0))
         return;
-    priv->splits = g_list_sort(priv->splits, (GCompareFunc)xaccSplitOrder);
+    std::sort (priv->splits.begin(), priv->splits.end(), split_cmp_less);
     priv->sort_dirty = FALSE;
     priv->balance_dirty = TRUE;
 }
@@ -2166,7 +2165,7 @@ xaccAccountInsertLot (Account *acc, GNCLot *lot)
 /********************************************************************\
 \********************************************************************/
 static void
-xaccPreSplitMove (Split *split, gpointer dummy)
+xaccPreSplitMove (Split *split)
 {
     xaccTransBeginEdit (xaccSplitGetParent (split));
 }
@@ -2193,7 +2192,7 @@ xaccAccountMoveAllSplits (Account *accfrom, Account *accto)
 
     /* optimizations */
     from_priv = GET_PRIVATE(accfrom);
-    if (!from_priv->splits || accfrom == accto)
+    if (from_priv->splits.empty() || accfrom == accto)
         return;
 
     /* check for book mix-up */
@@ -2203,7 +2202,7 @@ xaccAccountMoveAllSplits (Account *accfrom, Account *accto)
     xaccAccountBeginEdit(accfrom);
     xaccAccountBeginEdit(accto);
     /* Begin editing both accounts and all transactions in accfrom. */
-    g_list_foreach(from_priv->splits, (GFunc)xaccPreSplitMove, nullptr);
+    std::for_each (from_priv->splits.begin(), from_priv->splits.end(), xaccPreSplitMove);
 
     /* Concatenate accfrom's lists of splits and lots to accto's lists. */
     //to_priv->splits = g_list_concat(to_priv->splits, from_priv->splits);
@@ -2220,10 +2219,11 @@ xaccAccountMoveAllSplits (Account *accfrom, Account *accto)
      * Convert each split's amount to accto's commodity.
      * Commit to editing each transaction.
      */
-    g_list_foreach(from_priv->splits, (GFunc)xaccPostSplitMove, (gpointer)accto);
+    std::for_each (from_priv->splits.begin(), from_priv->splits.end(),
+                   [accto](Split *s){ xaccPostSplitMove (s, accto); });
 
     /* Finally empty accfrom. */
-    g_assert(from_priv->splits == nullptr);
+    g_assert(from_priv->splits.empty());
     g_assert(from_priv->lots == nullptr);
     xaccAccountCommitEdit(accfrom);
     xaccAccountCommitEdit(accto);
@@ -2268,7 +2268,6 @@ xaccAccountRecomputeBalance (Account * acc)
     gnc_numeric  noclosing_balance;
     gnc_numeric  cleared_balance;
     gnc_numeric  reconciled_balance;
-    GList *lp;
 
     if (nullptr == acc) return;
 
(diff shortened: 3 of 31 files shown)


jralls:
It's not even getting to tests, the build is failing:
```
/home/runner/work/gnucash/gnucash/build/gnucash-5.6/libgnucash/engine/Account.cpp:37:10: fatal error: Account.hpp: No such file or directory
   37 | #include "Account.hpp"
      |          ^~~~~~~~~~~~~
compilation terminated.
```

libgnucash/engine/Account.cpp
```diff
@@ -333,7 +334,7 @@ gnc_account_init(Account* acc)
     priv->lower_balance_limit = {};
     priv->include_sub_account_balances = {};
 
-    priv->splits = nullptr;
+    priv->splits = {};
```
jralls: `priv->splits.clear()` is better than assigning to an empty vector, but no initialization is needed, the `std::vector` constructor has already done that.

libgnucash/engine/Account.cpp
```diff
@@ -1370,22 +1371,19 @@ xaccFreeAccount (Account *acc)
     /* NB there shouldn't be any splits by now ... they should
      * have been all been freed by CommitEdit().  We can remove this
      * check once we know the warning isn't occurring any more. */
-    if (priv->splits)
+    if (priv->splits.size())
```
jralls: `if (!priv->splits.empty()`) communicates the intent better.

libgnucash/engine/Account.cpp
```diff
@@ -1370,22 +1371,19 @@ xaccFreeAccount (Account *acc)
...
-        GList *slist;
         PERR (" instead of calling xaccFreeAccount(), please call\n"
               " xaccAccountBeginEdit(); xaccAccountDestroy();\n");
 
         qof_instance_reset_editlevel(acc);
 
-        slist = g_list_copy(priv->splits);
-        for (lp = slist; lp; lp = lp->next)
+        std::vector<Split*> slist = priv->splits;
+        for (const auto& s : slist)
```
jralls: I don't think there's any benefit to copying the vector. Since the vector contains Split* getting references just slows things down by adding a useless indirection. Do `for (auto s : priv->splits)` instead.

libgnucash/engine/Account.cpp
```diff
@@ -1502,18 +1499,13 @@ xaccAccountCommitEdit (Account *acc)
            themselves will be destroyed by the transaction code */
         if (!qof_book_shutting_down(book))
         {
-            slist = g_list_copy(priv->splits);
-            for (lp = slist; lp; lp = lp->next)
-            {
-                Split *s = static_cast<Split *>(lp->data);
+            std::vector<Split*> slist = priv->splits;
+            for (const auto& s : slist)
```
jralls: Same thing here: `for (auto s : priv->splits)`.

libgnucash/engine/Account.cpp
```diff
@@ -1502,18 +1499,13 @@ xaccAccountCommitEdit (Account *acc)
...
                 xaccSplitDestroy (s);
-            }
-            g_list_free(slist);
         }
         else
         {
-            g_list_free(priv->splits);
-            priv->splits = nullptr;
+            priv->splits = {};
```
jralls: `priv->splits.clear()`.

libgnucash/engine/Account.cpp
```diff
@@ -1824,38 +1816,25 @@ xaccAccountEqual(const Account *aa, const Account *ab, gboolean check_guids)
...
 
-            if ((la != nullptr) || (lb != nullptr))
-            {
-                PWARN ("number of splits differs");
-                return(FALSE);
-            }
+        if (it_a != priv_aa->splits.end() || it_b != priv_ab->splits.end())
+        {
+            PWARN ("number of splits differs");
+            return(FALSE);
```
jralls: Change this to 
```
if (priv_aa->splits.size() != priv_ab->splits.size())
{
    PWARN ("number of splits differs");
   return FALSE;
}
```
and put it before the loop as a short-circuit. We did it this way with the GList implementation to save a list traversal, but that's an antipattern for a std::vector.

libgnucash/engine/Account.cpp
```diff
@@ -1964,17 +1942,17 @@ gboolean
...
     priv = GET_PRIVATE(acc);
-    node = g_list_find(priv->splits, s);
-    if (nullptr == node)
+
+    auto it = std::find (priv->splits.begin(), priv->splits.end(), s);
+    if (it == priv->splits.end())
         return FALSE;
 
-    priv->splits = g_list_delete_link(priv->splits, node);
+    priv->splits.erase (it);
```
jralls: Use [`std::remove(priv->splits.begin(), priv->splits.end(), s)`](https://en.cppreference.com/w/cpp/algorithm/remove) instead of rolling your own loop. See https://en.cppreference.com/w/cpp/container/vector/erase#Notes.

libgnucash/engine/Account.cpp
```diff
@@ -2168,17 +2146,17 @@ xaccAccountInsertLot (Account *acc, GNCLot *lot)
...
 }
 
 static void
-xaccPostSplitMove (Split *split, Account *accto)
+xaccPostSplitMove (Split *split)
 {
     Transaction *trans;
 
-    xaccSplitSetAccount(split, accto);
+    xaccSplitSetAccount(split, nullptr);
```
jralls: 1, that's a [no-op](https://github.com/Gnucash/gnucash/blob/c1ee59c39bf5287f30dc2ef489b5f4b71db66029/libgnucash/engine/Split.cpp#L978) and 2, the whole point of `xaccMoveSplits` is to change the splits' account to accto. You've broken it.

libgnucash/engine/Account.cpp
```diff
@@ -2287,9 +2264,8 @@ xaccAccountRecomputeBalance (Account * acc)
 
     PINFO ("acct=%s starting baln=%" G_GINT64_FORMAT "/%" G_GINT64_FORMAT,
            priv->accountName, balance.num, balance.denom);
-    for (lp = priv->splits; lp; lp = lp->next)
+    for (const auto& split : priv->splits)
```
jralls: No point in getting a reference, it's already a pointer.

libgnucash/engine/Account.cpp
```diff
@@ -3593,11 +3566,11 @@ GetBalanceAsOfDate (Account *acc, time64 date, gboolean ignclosing)
...
     xaccAccountRecomputeBalance (acc); /* just in case, normally a noop */
 
-    for (GList *lp = GET_PRIVATE(acc)->splits; lp; lp = lp->next)
+    for (const auto& split : GET_PRIVATE(acc)->splits)
     {
-        if (xaccTransGetDate (xaccSplitGetParent ((Split *)lp->data)) >= date)
+        if (xaccTransGetDate (xaccSplitGetParent (split)) >= date)
             break;
-        latest = (Split *)lp->data;
+        latest = split;
```
jralls:
```
auto priv{GET_PRIVATE(acc)};
auto latest{std::find_if_not(priv->splits.rbegin(), priv->splits.rend(), [date](auto s)->bool { 
        return xaccTransGetDate(xaccSplitGetParent(s)) >= date;
    })};
if (latest == priv->splits.rend())
   return gnc_numeric_zero();
```

libgnucash/engine/Account.cpp
```diff
@@ -3628,9 +3601,8 @@ xaccAccountGetReconciledBalanceAsOfDate (Account *acc, time64 date)
 
     g_return_val_if_fail(GNC_IS_ACCOUNT(acc), gnc_numeric_zero());
 
-    for (GList *node = GET_PRIVATE(acc)->splits; node; node = node->next)
+    for (const auto& split : GET_PRIVATE(acc)->splits)
     {
-        Split *split = (Split*) node->data;
         if ((xaccSplitGetReconcile (split) == YREC) &&
             (xaccSplitGetDateReconciled (split) <= date))
             balance = gnc_numeric_add_fixed (balance, xaccSplitGetAmount (split));
```
jralls:
```
auto priv{GET_PRIVATE(acc)};
return std::accumulate(priv->splits.begin(), priv->splits.end(), gnc_numeric_zero(),
                                      [date](auto balance, auto split)->gnc_numeric {
                                             if ((xaccSplitGetReconcile(split)  == YREC) &&
                                                   xaccSplitGetDateReconciled(split) <= date))
                                                  balance = gnc_numeric_add( balance, xaccSplitGetAmount (split));
                                              return balance;
                                });
```

libgnucash/engine/Account.cpp
```diff
@@ -5408,9 +5388,9 @@ finder_help_function(const Account *acc, const char *description,
...
      * returned!?  */
     priv = GET_PRIVATE(acc);
-    for (slp = g_list_last(priv->splits); slp; slp = slp->prev)
+    for (auto it = priv->splits.rbegin(); it != priv->splits.rend(); it++)
     {
-        Split *lsplit = static_cast<Split*>(slp->data);
+        auto lsplit = *it;
         Transaction *ltrans = xaccSplitGetParent(lsplit);
 
         if (g_strcmp0 (description, xaccTransGetDescription (ltrans)) == 0)
```
jralls: Lose `finder_help_function`, implement `xaccAccountFindSplitByDesc` with `std::find_if(priv->splits.rbegin(), priv->splits.rend(), /* lambda to do the comparison */)` and `xaccAccountFindTransByDesc` with `xaccSplitGetParent(xaccAccountFindSplitByDesc())`.

libgnucash/engine/Account.cpp
```diff
@@ -3552,9 +3525,9 @@ xaccAccountGetProjectedMinimumBalance (const Account *acc)
 
     priv = GET_PRIVATE(acc);
     today = gnc_time64_get_today_end();
-    for (node = g_list_last(priv->splits); node; node = node->prev)
+    for (auto it = priv->splits.rbegin(); it != priv->splits.rend(); it++)
     {
-        Split *split = static_cast<Split*>(node->data);
+        Split *split = *it;
```
jralls:
```
xaccAccountGetProjectedMinimumBalance (const Account *acc)
{
    bool seen_a_transaction{false};
    auto today{gnc_time64_get_today_end
    auto priv{GET_PRIVATE(acc);
    auto last_real_trans{std::find_if(priv->spllits.rbegin(), priv->splits.rend(), 
                                      [date](auto split)->bool {
                                          return xaccTransGetDate(xaccSplitGetParent(split) <= today);
                                      });
    return std::accumulate(priv->splits.rbegin(), last_real_trans, gnc_numeric_zero(),
                         [today, seen_a_transaction&](auto lowest, auto split)->gnc_numeric
                         {
                             if (seen_a_transaction)
                                 return gnc_numeric_compare(xaccSplitGetBalance(split), lowest) < 0) ?   
                                     xaccSplitGetBalance(split) : lowest;

                              seen_a_transaction = true;
                              return xaccSplitGetBalance(split);
                           });
}
```

libgnucash/engine/Account.cpp
```diff
@@ -5596,7 +5570,7 @@ static void do_one_split (Split *s, gpointer data)
 static void do_one_account (Account *account, gpointer data)
 {
     AccountPrivate *priv = GET_PRIVATE(account);
-    g_list_foreach(priv->splits, (GFunc)do_one_split, nullptr);
+    std::for_each (priv->splits.begin(), priv->splits.end(), do_one_split);
```
jralls: Move `do_one_split` into a lambda.

libgnucash/engine/Account.cpp
```diff
@@ -5617,24 +5591,21 @@ xaccAccountStagedTransactionTraversal (const Account *acc,
...
     priv = GET_PRIVATE(acc);
-    for (split_p = priv->splits; split_p; split_p = next)
+    for (auto it = priv->splits.begin(); it != priv->splits.end();)
     {
         /* Get the next element in the split list now, just in case some
          * naughty thunk destroys the one we're using. This reduces, but
          * does not eliminate, the possibility of undefined results if
          * a thunk removes splits from this account. */
-        next = g_list_next(split_p);
+        auto next = std::next (it);
```
jralls: That doesn't work for vectors. If a thunk destroys a split or moves it to a different account (either of which will remove it from `splits` behind our back) then `next` and all further iterators including `splits->end()` are invalid and continuing the loop would be undefined behavior. I think in practice what will happen is that we'd wind up skipping over the next split (because it's now pointed at by the current iterator and next points at the next one; vector iterators are just indexes) and then run off the end of the vector because end() points one split too far.

One safety measure would be to copy the vector and iterate on the copy, the other would be to set a flag on the Account and assert in `gnc_account_remove_split` that it's not set. The first is more in the spirit of the current design, though it conceivably could make for weird results. The second will crash. That might be an over-reaction depending on what's going on, but the backtrace will make it easy to figure out what happened.

gnucash/gnome-utils/dialog-transfer.cpp
```diff
@@ -472,10 +473,8 @@ gnc_xfer_dialog_reload_quickfill( XferDialog *xferData )
     gnc_quickfill_destroy( xferData->qf );
     xferData->qf = gnc_quickfill_new();
 
-    auto splitlist = xaccAccountGetSplitList( account );
-    for ( GList *node = splitlist; node; node = node->next )
+    for (const auto& split : xaccAccountGetSplits (account))
```
jralls: No references.

gnucash/gnome/gnc-plugin-page-account-tree.cpp
```diff
@@ -1595,7 +1592,7 @@ confirm_delete_account (GSimpleAction *simple, GncPluginPageAccountTree *page,
                                 acct_name);
     g_free(acct_name);
 
-    if (splits)
+    if (!splits.empty())
```
jralls: It looks like this is the only use of `splits`, so lose the decl at line 1581 and say 
`if (!xaccAccountGetSplits(account).empty())` 
here.

gnucash/gnome/window-reconcile.c
```diff
@@ -1612,15 +1612,15 @@ recn_set_watches_one_account (gpointer data, gpointer user_data)
 {
     Account *account = (Account *)data;
     RecnWindow *recnData = (RecnWindow *)user_data;
-    GList *node;
+    GList *splits = xaccAccountGetSplitList (account);
```
jralls: Put the new decl next to where it's used.

jralls (changes requested):
Good first (or maybe second) draft. 

I didn't mark all of the places where you used `for (const auto& s :  priv->splits)` but you should fix them all.

christopherlam:
Ok will fix these cppisms; the design of this branch was to introduce the vector and perform minimal changes rather than perform larger refactoring which could be error prone. Still haven't found out why test sx fails. In the vector version it segfaults because split is null during account staged reversals during sx->proper transaction creation.

jralls:
> Still haven't found out why test sx fails. In the vector version it segfaults because split is null during account staged reversals during sx->proper transaction creation.

Interesting that it does build all the way on Arch but fails for different reasons on macOS and Ubuntu. I take it that it builds completely and test-sx fails rather than segfaulting on your system.

christopherlam:
In

> > Still haven't found out why test sx fails. In the vector version it segfaults because split is null during account staged reversals during sx->proper transaction creation.
> 
> Interesting that it does build all the way on Arch but fails for different reasons on macOS and Ubuntu. I take it that it builds completely and test-sx fails rather than segfaulting on your system.

In mine debian based system it seems to run well, and all tests pass except test-sx segfaults in account.CPP one of the stage traversal functions.

jralls:
Tested in my bookworm VM. The segfault in test-sx is because the iterator manipulation in `xaccAccountStagedTransactionTraversal` doesn't work:
```

Thread 1 "test-sx" received signal SIGSEGV, Segmentation fault.
0x00007ffff78fc71f in xaccAccountStagedTransactionTraversal (
    acc=0x555555634b80 [Account], stage=42, 
    thunk=0x7ffff7fbbf5a <create_each_transaction_helper>, 
    cb_data=0x7fffffffda10)
    at /home/john/gnucash-git/libgnucash/engine/Account.cpp:5609
5609	        trans = s->parent;
(gdb) p s
$1 = 0x69746361782d6465
(gdb) p *s
Cannot access memory at address 0x69746361782d6465
(gdb) p priv->splits.size()
$2 = 2
(gdb) p priv->splits[0]
$3 = 0x5555556401c0 [Split]
(gdb) p priv->splits[1]
$4 = 0x555555640090 [Split]
(gdb) p it
$5 = 0x69746361782d6465
(gdb) p *it
$6 = (split_s *&) @0x55555563d4f8: 0x69746361782d6465
(gdb) p it->parent
Could not find operator->.
(gdb) p priv->splits.end()
$7 = 0x555555640090 [Split]
(gdb) p priv->splits.begin()
$8 = 0x5555556401c0 [Split]
(gdb) p next
$9 = 0x6e6f
```
This fixes it:
```

--- a/libgnucash/engine/Account.cpp
+++ b/libgnucash/engine/Account.cpp
@@ -5590,34 +5590,21 @@ xaccAccountStagedTransactionTraversal (const Account *acc,
                                        TransactionCallback thunk,
                                        void *cb_data)
 {
-    AccountPrivate *priv;
-    Transaction *trans;
-    int retval;
-
     if (!acc) return 0;
 
-    priv = GET_PRIVATE(acc);
-    for (auto it = priv->splits.begin(); it != priv->splits.end();)
+    auto splits = GET_PRIVATE(acc)->splits;
+    for (auto s : splits)
     {
-        /* Get the next element in the split list now, just in case some
-         * naughty thunk destroys the one we're using. This reduces, but
-         * does not eliminate, the possibility of undefined results if
-         * a thunk removes splits from this account. */
-        auto next = std::next (it);
-
-        auto s = *it;
-        trans = s->parent;
+        auto trans = s->parent;
         if (trans && (trans->marker < stage))
         {
             trans->marker = stage;
             if (thunk)
             {
-                retval = thunk(trans, cb_data);
+                auto retval = thunk(trans, cb_data);
                 if (retval) return retval;
             }
         }
-
-        it = next;
     }
 
     return 0;
```

libgnucash/engine/Account.cpp
christopherlam: > `priv->splits.clear()` is better than assigning to an empty vector, but no initialization is needed, the `std::vector` constructor has already done that.

But the `AccountPrivate` object is created via `g_object_new` or `g_malloc` rather than `new` so the constructors won't get called?

libgnucash/engine/Account.cpp
christopherlam: can't make it work. we need to return early if no removal took place.

libgnucash/engine/Account.cpp
christopherlam: > 1, that's a [no-op](https://github.com/Gnucash/gnucash/blob/c1ee59c39bf5287f30dc2ef489b5f4b71db66029/libgnucash/engine/Split.cpp#L978) and 2, the whole point of `xaccMoveSplits` is to change the splits' account to accto. You've broken it.

it's because `xaccPostSplitMove` was being called with `accto==nullptr` in the only caller below.

libgnucash/engine/Account.cpp
christopherlam: see previous comment regarding `GetBalanceAsOfDate`

libgnucash/engine/Account.cpp
christopherlam: i found a better way, combined with your next comment regarding `xaccAccountGetReconciledBalanceAsOfDate`...

libgnucash/engine/Account.cpp
christopherlam: see previous comment, refactoring...

libgnucash/engine/Account.cpp
christopherlam: ditto refactored to be IMHO much more functional and tidy

libgnucash/engine/Account.cpp
christopherlam: > it's because `xaccPostSplitMove` was being called with `accto==nullptr` in the only caller below.

uh... fixednow.

christopherlam:
here's a performance regression - on book close, `gnc_account_remove_split` empty the accounts. however the deletion is called by `gnc_transaction_book_end` which destroys the transactions as per the arbitrary hash order. therefore the vector.erase will keep shifting elements to the left for each erase. "it would be nice" if the `gnc_transaction_book_end` were to destroy the transactions in reverse chrono order.

christopherlam:
I think that's enough refactoring and cleaning up the main branch.

libgnucash/engine/Account.cpp
jralls: It all gets initialized to 0s, and that's what an empty vector is:
```
   frame #0: 0x00007ffff7ae6fd5 libgnc-engine.so`::gnc_account_init(acc=0x000055555561e1f0) at Account.cpp:303:20
   300 	    AccountPrivate *priv;
   301 	
   302 	    priv = GET_PRIVATE(acc);
-> 303 	    priv->parent   = nullptr;
   304 	    priv->children = nullptr;
   305 	
   306 	    priv->accountName = qof_string_cache_insert("");
(lldb) p sizeof(AccountPrivate)
(unsigned long) $1 = 360
(lldb) x/36wx priv
0x55555561e030: 0x00000000 0x00000000 0x00000000 0x00000000
0x55555561e040: 0x00000000 0x00000000 0x00000000 0x00000000
0x55555561e050: 0x00000000 0x00000000 0x00000000 0x00000000
0x55555561e060: 0x00000000 0x00000000 0x00000000 0x00000000
0x55555561e070: 0x00000000 0x00000000 0x00000000 0x00000000
0x55555561e080: 0x00000000 0x00000000 0x00000000 0x00000000
0x55555561e090: 0x00000000 0x00000000 0x00000000 0x00000000
0x55555561e0a0: 0x00000000 0x00000000 0x00000000 0x00000000
0x55555561e0b0: 0x00000000 0x00000000 0x00000000 0x00000000
...
   frame #0: 0x00007ffff7ae719a libgnc-engine.so`::gnc_account_init(acc=0x000055555561e1f0) at Account.cpp:333:40
   330 	
   331 	    priv->higher_balance_limit = {};
   332 	    priv->lower_balance_limit = {};
-> 333 	    priv->include_sub_account_balances = {};
   334 	
   335 	    priv->splits.clear();
   336 	    priv->sort_dirty = FALSE;
(lldb) p priv->splits
(std::vector<split_s *, std::allocator<> >) $5 = size=1 {
  std::_Vector_base<split_s *, std::allocator<> > = {
    _M_impl = {
      std::_Vector_base<split_s *, std::allocator<> >::_Vector_impl_data = {
        _M_start = nullptr
        _M_finish = nullptr
        _M_end_of_storage = nullptr
      }
    }
  }
}
...
Process 837193 stopped
* thread #1, name = 'test-engine', stop reason = step over
    frame #0: 0x00007ffff7ae71c7 libgnc-engine.so`::gnc_account_init(acc=0x000055555561e1f0) at Account.cpp:336:22
   333 	    priv->include_sub_account_balances = {};
   334 	
   335 	    priv->splits.clear();
-> 336 	    priv->sort_dirty = FALSE;
   337 	}
   338 	
   339 	static void
(lldb) p priv->splits
(std::vector<split_s *, std::allocator<> >) $6 = size=1 {
  std::_Vector_base<split_s *, std::allocator<> > = {
    _M_impl = {
      std::_Vector_base<split_s *, std::allocator<> >::_Vector_impl_data = {
        _M_start = nullptr
        _M_finish = nullptr
        _M_end_of_storage = nullptr
      }
    }
  }
}
```
So all of that setting stuff to `{}` is redundant, as is `priv->splits.clear()`.

jralls:
> "it would be nice" if the gnc_transaction_book_end were to destroy the transactions in reverse chrono order.

It would be cleaner and more efficient to have `xaccAccountCommitEdit` free the splits and let `xaccFreeTransaction` just free its split list.

jralls (approved):

libgnucash/engine/Account.cpp
```diff
@@ -5538,14 +5461,11 @@ gnc_account_merge_children (Account *parent)
 /* Transaction Traversal functions                                  */
 
 
-void
-xaccSplitsBeginStagedTransactionTraversals (GList *splits)
+static void
+xaccSplitsBeginStagedTransactionTraversals (SplitsVec splits)
```
christopherlam, line 5465: oops should have been a `SplitsVec&`

christopherlam:
not quite sure why the vector still leaks on a small file. I've added a trace (++count on insert, --count on remove) and count is zero on exit.

```
==643006== 2,336 bytes in 18 blocks are definitely lost in loss record 18,892 of 19,272
==643006==    at 0x4844ED1: operator new(unsigned long) (in /gnu/store/kr9ic1mdw3c79dzwygicv2g3y7f41gsc-valgrind-3.22.0/libexec/valgrind/vgpreload_memcheck-amd64-linux.so)
==643006==    by 0x6E75BC7: __gnu_cxx::new_allocator<split_s*>::allocate(unsigned long, void const*) (new_allocator.h:127)
==643006==    by 0x6E73F25: std::allocator_traits<std::allocator<split_s*> >::allocate(std::allocator<split_s*>&, unsigned long) (alloc_traits.h:464)
==643006==    by 0x6E715F5: std::_Vector_base<split_s*, std::allocator<split_s*> >::_M_allocate(unsigned long) (stl_vector.h:346)
==643006==    by 0x6E6E283: void std::vector<split_s*, std::allocator<split_s*> >::_M_realloc_insert<split_s* const&>(__gnu_cxx::__normal_iterator<split_s**, std::vector<split_s*, std::allocator<split_s*> > >, split_s* const&) (vector.tcc:440)
==643006==    by 0x6E6BAD5: std::vector<split_s*, std::allocator<split_s*> >::push_back(split_s* const&) (stl_vector.h:1198)
==643006==    by 0x6E53DE5: gnc_account_insert_split (Account.cpp:1944)
==643006==    by 0x6E8D311: xaccSplitCommitEdit (Split.cpp:1031)
==643006==    by 0x6E95631: trans_cleanup_commit(QofInstance_s*) (Transaction.cpp:1583)
==643006==    by 0x704FFFB: qof_commit_edit_part2 (qofinstance.cpp:1039)
==643006==    by 0x6E959E8: xaccTransCommitEdit (Transaction.cpp:1666)
==643006==    by 0x62D31DA: dom_tree_to_transaction(_xmlNode*, _QofBook*) (gnc-transaction-xml-v2.cpp:613)
==643006== 
```

jralls:
What inserts and deletes are you counting?  Most of the time `priv->splits` is simply cleared in `xaccAccountCommitEdit` when the account is destroyed during book shutdown.

christopherlam:
here's my valgrind script:

```bash
cd stable-build
ninja -j 3
ninja install > /dev/null
cd ../stable-install/bin
G_SLICE=always-malloc G_DEBUG=gc-friendly valgrind --tool=memcheck --keep-debuginfo=yes --track-origins=yes --leak-check=full ./gnucash --logto stdout --extra 2> valgrind.txt
```

and the diff

```diff
diff --git a/libgnucash/engine/Account.cpp b/libgnucash/engine/Account.cpp
index f25f7c4feb..8460d4fe2f 100644
--- a/libgnucash/engine/Account.cpp
+++ b/libgnucash/engine/Account.cpp
@@ -1927,6 +1927,8 @@ static bool split_cmp_less (const Split* a, const Split* b)
     return xaccSplitOrder (a, b) < 0;
 }
 
+static int count = 0;
+
 gboolean
 gnc_account_insert_split (Account *acc, Split *s)
 {
@@ -1946,6 +1948,7 @@ gnc_account_insert_split (Account *acc, Split *s)
     else
         priv->sort_dirty = true;
 
+    PWARN ("count=%d", ++count);
     //FIXME: find better event
     qof_event_gen (&acc->inst, QOF_EVENT_MODIFY, nullptr);
     /* Also send an event based on the account */
@@ -1971,6 +1974,7 @@ gnc_account_remove_split (Account *acc, Split *s)
     if (it == priv->splits.end())
         return FALSE;
 
+    PWARN ("count=%d", --count);
     priv->splits.erase (it, priv->splits.end());
     //FIXME: find better event type
     qof_event_gen(&acc->inst, QOF_EVENT_MODIFY, nullptr);
```

libgnucash/engine/Account.cpp
```diff
@@ -3578,27 +3559,16 @@ xaccAccountGetProjectedMinimumBalance (const Account *acc)
...
-    {
-        if (xaccTransGetDate (xaccSplitGetParent ((Split *)lp->data)) >= date)
-            break;
-        latest = (Split *)lp->data;
-    }
+    auto is_before_date = [date](auto s) -> bool
+    { return xaccTransGetDate(xaccSplitGetParent(s)) < date; };
 
-    return latest ? split_to_numeric (latest) : gnc_numeric_zero();
+    auto latest_split{gnc_account_find_split (acc, is_before_date, true)};
```
christopherlam, line 3570: This would be an ideal use for `std::upper_bound`

christopherlam: ... or not. IRL most GetBalanceAsOfDate's dates would be towards the later end of the splits vector...

christopherlam:
> "it would be nice" if the `gnc_transaction_book_end` were to destroy the transactions in reverse chrono order.

this does it well. benchmark drop to 3.420s

```patch
diff --git a/libgnucash/engine/Account.cpp b/libgnucash/engine/Account.cpp
index f5ca47cb15..65b600613f 100644
--- a/libgnucash/engine/Account.cpp
+++ b/libgnucash/engine/Account.cpp
@@ -1971,6 +1971,9 @@ gnc_account_remove_split (Account *acc, Split *s)
     if (it == priv->splits.end())
         return FALSE;
 
+    auto date{qof_print_date (xaccTransGetDate (xaccSplitGetParent (s)))};
+    PWARN ("deleting %s = %s", xaccAccountGetName (acc), date);
+    g_free (date);
     priv->splits.erase (it, priv->splits.end());
     //FIXME: find better event type
     qof_event_gen(&acc->inst, QOF_EVENT_MODIFY, nullptr);
diff --git a/libgnucash/engine/Transaction.cpp b/libgnucash/engine/Transaction.cpp
index bc17b5d18e..9713b1b448 100644
--- a/libgnucash/engine/Transaction.cpp
+++ b/libgnucash/engine/Transaction.cpp
@@ -3070,6 +3070,12 @@ destroy_tx_on_book_close(QofInstance *ent, gpointer data)
     xaccTransDestroy(tx);
 }
 
+static int
+trans_cmp_less (const Transaction* ta, const Transaction* tb)
+{
+    return xaccTransOrder (ta, tb) < 0;
+}
+
 /** Handles book end - frees all transactions from the book
  *
  * @param book Book being closed
@@ -3080,7 +3086,8 @@ gnc_transaction_book_end(QofBook* book)
     QofCollection *col;
 
     col = qof_book_get_collection(book, GNC_ID_TRANS);
-    qof_collection_foreach(col, destroy_tx_on_book_close, nullptr);
+    qof_collection_foreach_sorted (col, destroy_tx_on_book_close, nullptr,
+                                   (GCompareFunc)trans_cmp_less);
 }
 
 #ifdef _MSC_VER
diff --git a/libgnucash/engine/qofid.cpp b/libgnucash/engine/qofid.cpp
index cd23c46365..9d613fa2f2 100644
--- a/libgnucash/engine/qofid.cpp
+++ b/libgnucash/engine/qofid.cpp
@@ -318,8 +318,8 @@ foreach_cb (gpointer item, gpointer arg)
 }
 
 void
-qof_collection_foreach (const QofCollection *col, QofInstanceForeachCB cb_func,
-                        gpointer user_data)
+qof_collection_foreach_sorted (const QofCollection *col, QofInstanceForeachCB cb_func,
+                               gpointer user_data, GCompareFunc sort_cb)
 {
     struct _qofid_iterate iter;
     GList *entries;
@@ -333,9 +333,19 @@ qof_collection_foreach (const QofCollection *col, QofInstanceForeachCB cb_func,
     PINFO("Hash Table size of %s before is %d", col->e_type, g_hash_table_size(col->hash_of_entities));
 
     entries = g_hash_table_get_values (col->hash_of_entities);
+    if (sort_cb)
+        entries = g_list_sort (entries, sort_cb);
     g_list_foreach (entries, foreach_cb, &iter);
     g_list_free (entries);
 
     PINFO("Hash Table size of %s after is %d", col->e_type, g_hash_table_size(col->hash_of_entities));
 }
+
+void
+qof_collection_foreach (const QofCollection *col, QofInstanceForeachCB cb_func,
+                        gpointer user_data)
+{
+    qof_collection_foreach_sorted (col, cb_func, user_data, nullptr);
+}
+
 /* =============================================================== */
diff --git a/libgnucash/engine/qofid.h b/libgnucash/engine/qofid.h
index a7c9bc0bcf..cf099a3a25 100644
--- a/libgnucash/engine/qofid.h
+++ b/libgnucash/engine/qofid.h
@@ -145,6 +145,9 @@ QofInstance * qof_collection_lookup_entity (const QofCollection *, const GncGUID
 /** Callback type for qof_collection_foreach */
 typedef void (*QofInstanceForeachCB) (QofInstance *, gpointer user_data);
 
+void qof_collection_foreach_sorted (const QofCollection*, QofInstanceForeachCB,
+                                    gpointer, GCompareFunc);
+
 /** Call the callback for each entity in the collection. */
 void qof_collection_foreach (const QofCollection *, QofInstanceForeachCB,
                              gpointer user_data);
```

jralls:
In re leaks, it looks to me like a libstdc++ problem. `xaccFreeAccount` would be logging an error if `priv->splits` wasn't empty, so there must be a failure of std::vector to free its storage memory. 2336 bytes is 292 pointers. Does the book in question have that many splits in 18 accounts?

christopherlam:
> In re leaks, it looks to me like a libstdc++ problem. `xaccFreeAccount` would be logging an error if `priv->splits` wasn't empty, so there must be a failure of std::vector to free its storage memory. 2336 bytes is 292 pointers. Does the book in question have that many splits in 18 accounts?

This book has 233 splits.

christopherlam:
> > In re leaks, it looks to me like a libstdc++ problem. `xaccFreeAccount` would be logging an error if `priv->splits` wasn't empty, so there must be a failure of std::vector to free its storage memory. 2336 bytes is 292 pointers. Does the book in question have that many splits in 18 accounts?
> 
> This book has 233 splits.

Ok not a bug, vector needs https://en.cppreference.com/w/cpp/container/vector/shrink_to_fit to release memory, because the `AccountPrivate` doesn't have RAII.

jralls:
> because the AccountPrivate doesn't have RAII.

That's not about [RAII](https://en.cppreference.com/w/cpp/language/raii) because `splits` is a direct member of `AccountPrivate`.  If left to its own devices the compiler creates  a [default constructor and destructor](https://godbolt.org/z/fvzKj3P4c) for `AccountPrivate` the latter of which in turn calls the vector's destructor. As you've observed elsewhere using `malloc` and `free` [short circuit that](https://godbolt.org/z/T5cv4zejM), but we can work around it by [expressly calling the default destructor](https://godbolt.org/z/8sPrvah4v); if `Account` was correctly implemented we do that in a function named `gnc_account_finalize`, but it isn't so best to put that at the end of `xaccFreeAccount`.

PR #1920, opened 2024-04-17
commodity-utilities.scm -> c++

christopherlam:
step 1 of many for speed up currency conversions for bug 799258: make `gnc_get_match_commodity_splits` a cpp function. why does report eg asset chart fail?

```
* 00:18:52  WARN <gnc.gui> [gnc_run_report_with_error_handling()] Error in report: In ice-9/boot-9.scm:
  1752:10 11 (with-exception-handler _ _ #:unwind? _ # _)
In unknown file:
          10 (apply-smob/0 #<thunk 7a3fdf514300>)
In c-interface.scm:
     34:5  9 (gnc:call-with-error-handling _ _)
In ice-9/boot-9.scm:
  1747:15  8 (with-exception-handler #<procedure 7a3fcb9258a0 at ic…> …)
  1752:10  7 (with-exception-handler _ _ #:unwind? _ # _)
In c-interface.scm:
    38:40  6 (_)
In report-core.scm:
   748:25  5 (gnc:report-render-html #<<report> type: "e9cf815f79db…> …)
In standard/category-barchart.scm:
   282:26  4 (category-barchart-renderer #<<report> type: "e9cf815f…> …)
In commodity-utilities.scm:
   824:31  3 (gnc:case-exchange-time-fn _ _ _ _ _ _)
   192:29  2 (gnc:get-commoditylist-totalavg-prices (#<swig-poin…> …) …)
In unknown file:
           1 (gnc-get-match-commodity-splits (#<swig-pointer Acc…> …) …)
In ice-9/boot-9.scm:
  1685:16  0 (raise-exception _ #:continuable? _)

In procedure gnc-get-match-commodity-splits: Wrong type argument in position 1: (#<swig-pointer Account * 17a7080> #<swig-pointer Account * 17a6e00> #<swig-pointer Account * 17a9b90> #<swig-pointer Account * 17a9d80> #<swig-pointer Account * 17a73d0> #<swig-pointer Account * 17abc90> #<swig-pointer Account * 17ac340> #<swig-pointer Account * 17a75c0> #<swig-pointer Account * 17aa240> #<swig-pointer Account * 17a77b0> #<swig-pointer Account * 17a7c40> #<swig-pointer Account * 17a8890> #<swig-pointer Account * 17a7eb0> #<swig-pointer Account * 17a8cf0> #<swig-pointer Account * 17ac7b0> #<swig-pointer Account * 17acde0> #<swig-pointer Account * 17a8f60> #<swig-pointer Account * 17a8280> #<swig-pointer Account * 17acfd0> #<swig-pointer Account * 17ad1c0> #<swig-pointer Account * 17ad3b0> #<swig-pointer Account * 17ad790> #<swig-pointer Account * 17ad5a0> #<swig-pointer Account * 17ad980> #<swig-pointer Account * 17adb70> #<swig-pointer Account * 17add60> #<swig-pointer Account * 17adf50>)
```

### DIFF
--- a/gnucash/report/commodity-utilities.scm
+++ b/gnucash/report/commodity-utilities.scm
@@ -61,18 +61,10 @@
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;; Functions to get splits with interesting data from accounts.
 
-;; helper function. queries book for all splits in accounts before
-;; end-date (end-date can be #f)
-(define (get-all-splits accounts end-date)
-  (let ((query (qof-query-create-for-splits)))
-    (qof-query-set-book query (gnc-get-current-book))
-    (xaccQueryAddClearedMatch
-     query (logand CLEARED-ALL (lognot CLEARED-VOIDED)) QOF-QUERY-AND)
-    (xaccQueryAddAccountMatch query accounts QOF-GUID-MATCH-ANY QOF-QUERY-AND)
-    (xaccQueryAddDateMatchTT query #f 0 (and end-date #t) (or end-date 0) QOF-QUERY-AND)
-    (let ((splits (qof-query-run query)))
-      (qof-query-destroy query)
-      splits)))
+(define (get-all-commodity-splits currency-accounts end-date commodity sort?)
+  (gnc-get-match-commodity-splits currency-accounts (and end-date #t)
+                                  (or end-date 0) (or commodity '()) sort?))
+
 
 ;; Returns a list of all splits in the 'currency-accounts' up to
 ;; 'end-date' which have two different commodities involved, one of
@@ -80,15 +72,7 @@
 ;; 'commodity' != #f ).
 (define (gnc:get-match-commodity-splits
          currency-accounts end-date commodity)
-  (filter
-   (lambda (s)
-     (let ((txn-comm (xaccTransGetCurrency (xaccSplitGetParent s)))
-           (acc-comm (xaccAccountGetCommodity (xaccSplitGetAccount s))))
-       (and (not (gnc-commodity-equiv txn-comm acc-comm))
-            (or (not commodity)
-                (gnc-commodity-equiv commodity txn-comm)
-                (gnc-commodity-equiv commodity acc-comm)))))
-   (get-all-splits currency-accounts end-date)))
+  (get-all-commodity-splits currency-accounts end-date commodity #f))
 
 ;; Returns a sorted list of all splits in the 'currency-accounts' up
 ;; to 'end-date' which have the 'commodity' and one other commodity
@@ -96,12 +80,7 @@
 (define (gnc:get-match-commodity-splits-sorted currency-accounts
                                                end-date
                                                commodity)
-  (sort (gnc:get-match-commodity-splits currency-accounts
-                                        end-date commodity)
-        (lambda (a b)
-          (<
-           (xaccTransGetDate (xaccSplitGetParent a))
-           (xaccTransGetDate (xaccSplitGetParent b))))))
+  (get-all-commodity-splits currency-accounts end-date commodity #t))
 
 
 ;; Returns a list of all splits in the currency-accounts up to
--- a/gnucash/report/commodity-utilities.scm
+++ b/gnucash/report/commodity-utilities.scm
@@ -124,10 +124,12 @@
   (gnc:get-commodity-totalavg-prices-internal
    currency-accounts end-date price-commodity report-currency
    (gnc:get-match-commodity-splits-sorted
-    currency-accounts end-date price-commodity)))
+    currency-accounts end-date price-commodity)
+   #f))
 
 (define (gnc:get-commodity-totalavg-prices-internal
-         currency-accounts end-date price-commodity report-currency commodity-splits)
+         currency-accounts end-date price-commodity report-currency commodity-splits
+         hide-warnings?)
   (let loop ((tot-foreign 0)
              (tot-domestic 0)
              (commodity-splits commodity-splits)
@@ -173,16 +175,17 @@
                          (cons (list txn-date (/ new-domestic new-foreign)) result)))))
 
            (else
-            (warn "gnc:get-commodity-totalavg-prices: "
-                  "Sorry, currency exchange not yet implemented:"
-                  (gnc:monetary->string
-                   (gnc:make-gnc-monetary txn-comm value-amt))
-                  " (buying "
-                  (gnc:monetary->string
-                   (gnc:make-gnc-monetary price-commodity share-amt))
-                  ") =? "
-                  (gnc:monetary->string
-                   (gnc:make-gnc-monetary report-currency 0)))
+            (unless hide-warnings?
+              (warn "gnc:get-commodity-totalavg-prices: "
+                    "Sorry, currency exchange not yet implemented:"
+                    (gnc:monetary->string
+                     (gnc:make-gnc-monetary txn-comm value-amt))
+                    " (buying "
+                    (gnc:monetary->string
+                     (gnc:make-gnc-monetary price-commodity share-amt))
+                    ") =? "
+                    (gnc:monetary->string
+                     (gnc:make-gnc-monetary report-currency 0))))
             (loop tot-foreign
                   tot-domestic
                   (cdr commodity-splits)
@@ -194,9 +197,9 @@
 ;; extended to a commodity-list. Returns an alist. Each pair consists
 ;; of the foreign-currency and the appropriate list from
 ;; gnc:get-commodity-totalavg-prices, see there.
-(define (gnc:get-commoditylist-totalavg-prices
-         commodity-list report-currency end-date
-         start-percent delta-percent)
+(define* (gnc:get-commoditylist-totalavg-prices
+          commodity-list report-currency end-date
+          start-percent delta-percent #:key hide-warnings?)
   (let* ((currency-accounts
           (gnc-account-get-descendants-sorted (gnc-get-current-root-account)))
          (interesting-splits (gnc:get-match-commodity-splits-sorted currency-accounts end-date #f))
@@ -213,7 +216,8 @@
        (cons c
              (gnc:get-commodity-totalavg-prices-internal
               currency-accounts end-date c report-currency
-              (filter split-has-commodity? interesting-splits))))
+              (filter split-has-commodity? interesting-splits)
+              hide-warnings?)))
      commodity-list
      (iota work-to-do))))
 
@@ -306,7 +310,7 @@
 ;; (described in gnc:get-exchange-totals) and returns a report-list, This
 ;; resulting alist can immediately be plugged into gnc:make-exchange-alist.
 
-(define (gnc:resolve-unknown-comm sumlist report-commodity)
+(define* (gnc:resolve-unknown-comm sumlist report-commodity #:key hide-warnings?)
   ;; reportlist contains all known transactions with the
   ;; report-commodity, and now the transactions with unknown
   ;; currencies should be added to that list (with an appropriate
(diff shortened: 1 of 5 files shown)


christopherlam:
Ok function converted to cpp is much more memory-friendly than guile, because in scheme the `get-all-splits` would retrieve _all_ 10'000s of splits, filtered into the 100s of foreign conversion splits; converted to c++ which does filtering within the lambda. This does seem to work.

christopherlam (commented):
don't sort unless requested

christopherlam:
The new function is in Account.cpp because I can't fix `report.i` to swig `gnc_get_match_commodity_splits` properly. Anyway:

1. `gnc_get_match_commodity_splits` will scan all book splits and return only splits involving foreign exchange. Previously the whole book splitlist would be generated, scanned, filtered, gc'd.
2. `gnc:case-exchange-time-fn` which creates an `exchange-fn` can now accept a hash-table which will store the date-specific exchange-data. The hash-table should be scoped within a report render only.
3. The relevant multi-date reports amended to use the upgraded `gnc:case-exchange-time-fn`

Both 1 and 2 should mean multiyear reports would be much faster and would fix 799258 to boot...

gnucash/report/commodity-utilities.scm
```diff
@@ -819,24 +820,26 @@
 ;;    foreign  - foreign commodity/currency
 ;;    domestic - a gnc-monetary pair
 ;;    date     - time64 price
-(define (gnc:case-exchange-time-fn
-         source-option report-currency commodity-list to-date-tp
-         start-percent delta-percent)
+(define* (gnc:case-exchange-time-fn
+          source-option report-currency commodity-list to-date-tp
+          start-percent delta-percent #:key date-hash)
+  (define-syntax-rule (memoize date expensive-fn)
```
jralls, line 833: What is `memoize` and where is it documented? I can't find it in the Guile docs.

jralls:
> I can't fix report.i to swig gnc_get_match_commodity_splits properly

Probably because you didn't change `swig-report` to write C++. `swig-engine` is already C++.

gnucash/report/commodity-utilities.scm
christopherlam: Its the definition right there. It is a macro.

christopherlam:
> > I can't fix report.i to swig gnc_get_match_commodity_splits properly
> 
> Probably because you didn't change `swig-report` to write C++. `swig-engine` is already C++.

If you could move the function to gnc-report.cpp it would be a tremendous speedup and memory+gc improvement already.

gnucash/report/commodity-utilities.scm
christopherlam: You could copy the whole `gnc:case-exchange-time-fn` definition, and paste into guile repl after `,expand`:

```scheme
cheme@(guile-user)> ,expand (define* (gnc:case-exchange-time-fn
          source-option report-currency commodity-list to-date-tp
          start-percent delta-percent #:key date-hash)
  (define-syntax-rule (memoize date expensive-fn)
    (if date-hash
        (or (hash-ref date-hash date) (hashv-set! date-hash date expensive-fn))
        expensive-fn))
  (case source-option
    ;; Make this the same as gnc:case-exchange-fn
    ((average-cost) (lambda (foreign domestic date)
                      ((memoize date (gnc:make-exchange-function
                                      (gnc:make-exchange-cost-alist
                                       report-currency date)))
                       foreign domestic)))
    ((weighted-average) (lambda (foreign domestic date)
                          (gnc:exchange-by-pricealist-nearest
                           (memoize date (gnc:get-commoditylist-totalavg-prices
                                          commodity-list report-currency date
                                          start-percent delta-percent))
                           foreign domestic date)))
    ((pricedb-before) gnc:exchange-by-pricedb-nearest-before)
    ((pricedb-latest) (lambda (foreign domestic date)
                        (gnc:exchange-by-pricedb-latest foreign domestic)))
    ((pricedb-nearest) gnc:exchange-by-pricedb-nearest)
    (else
     (begin
       (gnc:warn "gnc:case-exchange-time-fn: bad price-source value: "
                 source-option ". Using pricedb-nearest.")
       ;; FIX-ME another hack to prevent report crashing when an
       ;; unimplemented source-option comes through
       gnc:exchange-by-pricedb-nearest))))
$1 = (define* (gnc:case-exchange-time-fn
          source-option
          report-currency
          commodity-list
          to-date-tp
          start-percent
          delta-percent
          #:key
          (date-hash #f #:date-hash))
  (let ((key source-option))
    (cond (((@@ (guile) memv) key '(average-cost))
           (lambda (foreign domestic date)
             ((if date-hash
                (or (hash-ref date-hash date)
                    (hashv-set!
                      date-hash
                      date
                      (gnc:make-exchange-function
                        (gnc:make-exchange-cost-alist
                          report-currency
                          date))))
                (gnc:make-exchange-function
                  (gnc:make-exchange-cost-alist
                    report-currency
                    date)))
              foreign
              domestic)))
          (((@@ (guile) memv) key '(weighted-average))
           (lambda (foreign domestic date)
             (gnc:exchange-by-pricealist-nearest
               (if date-hash
                 (or (hash-ref date-hash date)
                     (hashv-set!
                       date-hash
                       date
                       (gnc:get-commoditylist-totalavg-prices
                         commodity-list
                         report-currency
                         date
                         start-percent
                         delta-percent)))
                 (gnc:get-commoditylist-totalavg-prices
                   commodity-list
                   report-currency
                   date
                   start-percent
                   delta-percent))
               foreign
               domestic
               date)))
          (((@@ (guile) memv) key '(pricedb-before))
           gnc:exchange-by-pricedb-nearest-before)
          (((@@ (guile) memv) key '(pricedb-latest))
           (lambda (foreign domestic date)
             (gnc:exchange-by-pricedb-latest foreign domestic)))
          (((@@ (guile) memv) key '(pricedb-nearest))
           gnc:exchange-by-pricedb-nearest)
          (else
           (gnc:warn
             "gnc:case-exchange-time-fn: bad price-source value: "
             source-option
             ". Using pricedb-nearest.")
           gnc:exchange-by-pricedb-nearest))))
```

jralls: And what would that do for me?

christopherlam: It illustrates why I inserted the macro. To avoid calling expensive-fn unless it's necessary.

christopherlam:
> If you could move the function to gnc-report.cpp it would be a tremendous speedup and memory+gc improvement already.

@jralls Do you think you could do this for me? i.e. move `gnc_get_match_commodity_splits` to `gnc-report.cpp`. The first 2 commits in this branch are good to go.

jralls:
> @jralls Do you think you could do this for me? i.e. move gnc_get_match_commodity_splits to gnc-report.cpp. The first 2 commits in this branch are good to go.

I have worked on it a bit yesterday and today and I'm leaning towards it really belonging in sw_engine rather than sw_reports because it exposes a bit too much of Account's internals and consequently needs to wrap Account*, AccountList*, and SplitList* that are already wrapped by sw_engine. Why do you think it would be better or faster in sw_report than in sw_engine?

christopherlam:
> > @jralls Do you think you could do this for me? i.e. move gnc_get_match_commodity_splits to gnc-report.cpp. The first 2 commits in this branch are good to go.
> 
> I have worked on it a bit yesterday and today and I'm leaning towards it really belonging in sw_engine rather than sw_reports because it exposes a bit too much of Account's internals and consequently needs to wrap Account*, AccountList*, and SplitList* that are already wrapped by sw_engine. Why do you think it would be better or faster in sw_report than in sw_engine?

Umm because engine is getting rather polluted, and this is a "analyze and don't modify splits" type function

christopherlam:
> > > @jralls Do you think you could do this for me? i.e. move gnc_get_match_commodity_splits to gnc-report.cpp. The first 2 commits in this branch are good to go.
> > 
> > 
> > I have worked on it a bit yesterday and today and I'm leaning towards it really belonging in sw_engine rather than sw_reports because it exposes a bit too much of Account's internals and consequently needs to wrap Account*, AccountList*, and SplitList* that are already wrapped by sw_engine. Why do you think it would be better or faster in sw_report than in sw_engine?
> 
> Umm because engine is getting rather polluted, and this is a "analyze and don't modify splits" type function

instead of swig-report it could go to swig-app-utils... but even then, i failed to swig it.

jralls:
> Umm because engine is getting rather polluted, and this is a "analyze and don't modify splits" type function

I don't know what you mean by "polluted", but modification or lack thereof isn't significant. The function itself needs access to Engine internals, and swigging it requires guile&lt;-&gt;C/C++ conversions that live in bindings, in particular converting a Scheme list of SWIG_p_account_s into a GList of Account and a GList of Split into a Scheme list of SWIG_p_split_s. Combine that with a hefty serving of Spaghetti GnuCashese and you wind up having to duplicate a huge part of sw_engine in whatever other swig module you try to put it in on top of making the spaghetti problem worse by leaking engine implementation details.

gnucash/report/test/test-commodity-utils.scm
```diff
@@ -655,15 +655,15 @@
                           (gnc-dmy2time64-neutral 20 02 2016)
                           #f #f)))
         (test-equal "gnc:case-exchange-time-fn weighted-average 20/02/2012"
-          307/5
+          0
```
jralls, line 658: The commit message needs to explain why the answers changed, otherwise it looks like you've done something wrong and patched the test to hide it.

gnucash/report/commodity-utilities.scm
```diff
@@ -173,16 +173,16 @@
...
+            ;; (warn "gnc:get-commodity-totalavg-prices: "
+            ;;       "Sorry, currency exchange not yet implemented:"
+            ;;       (gnc:monetary->string
+            ;;        (gnc:make-gnc-monetary txn-comm value-amt))
+            ;;       " (buying "
+            ;;       (gnc:monetary->string
+            ;;        (gnc:make-gnc-monetary price-commodity share-amt))
+            ;;       ") =? "
+            ;;       (gnc:monetary->string
+            ;;        (gnc:make-gnc-monetary report-currency 0)))
```
jralls: Why are you removing these warnings? What has it got to do with caching the splits?

jralls (changes requested):
Otherwise looks mostly OK.

gnucash/report/commodity-utilities.scm
christopherlam: These warnings will slow down the scanning of Forex splits. If the user chooses a multi date report eg net worth line chart spanning 10 years at daily resolution, the warnings will be shown 3650 times give or take.

gnucash/report/test/test-commodity-utils.scm
christopherlam: I'll amend when rewriting the commit message.

gnucash/report/commodity-utilities.scm
jralls: So set a flag and print the error once. Arguably it should abort the report because it will be wrong, but it might be friendlier to print a message on the report.

Regardless it should be its own commit.

christopherlam: ok it'll be tricky because it'll be the _last_ report-date which captures the most splits which matters the most.

christopherlam: My thoughts about showing the warnings once per report: we'd need to augment the case-exchange-date-fn to accept a "show-warnings?" optional argument which scans the splits and dumps any warnings. This would need to be called by reports once per report render.

christopherlam:
so:

1. keep it in swig-engine but function lives in engine.i ?
2. add `gnc_account_foreach_split` which takes a std::function?
3. augment `gnc:case-exchange-time-fn` which accepts a `#:show-warnings?` optarg to show these warnings?

gnucash/report/commodity-utilities.scm
christopherlam: solved.

gnucash/report/test/test-commodity-utils.scm
christopherlam: test fix commit message amended.

jralls (approved):
Much better. Thanks for separating the rework for easier review, don't forget to squash out the original versions before merging.

christopherlam:
> Much better. Thanks for separating the rework for easier review, don't forget to squash out the original versions before merging.

Squashed up, also added `SplitVec` and `AccountVec` in base-typemaps.i which could shave off a few more nanoseconds and could be used going forward...

PR #1926, opened 2024-04-23
[engine.i] even faster gnc_get_match_commodity_splits

christopherlam:
`gnc_get_match_commodity_splits` needs to scan the account splitlists to find suitable splits upto `end_date`. We can stop scanning each account at `end_date` because the splitlists are sorted by date. This saves significant time because majority of splits can by bypassed in the early report dates.

My primary book net worth linechart benchmarks -- fastest speeds from 1.05s to 0.54s, now practically unnoticeable

### DIFF
--- a/bindings/engine.i
+++ b/bindings/engine.i
@@ -129,19 +129,27 @@ static const GncGUID * gncBudgetGetGUID(GncBudget *x)
 SplitsVec gnc_get_match_commodity_splits (AccountVec accounts, bool use_end_date,
                                           time64 end_date, gnc_commodity *comm, bool sort)
 {
-    auto match = [use_end_date, end_date, comm](const Split* s) -> bool
+    SplitsVec rv;
+
+    auto maybe_accumulate = [&rv, comm](auto s)
     {
-        if (xaccSplitGetReconcile (s) == VREC) return false;
-        auto trans{xaccSplitGetParent (s)};
-        if (use_end_date && xaccTransGetDate(trans) > end_date) return false;
-        auto txn_comm{xaccTransGetCurrency (trans)};
+        auto txn_comm{xaccTransGetCurrency (xaccSplitGetParent (s))};
         auto acc_comm{xaccAccountGetCommodity (xaccSplitGetAccount (s))};
-        return (txn_comm != acc_comm) && (!comm || comm == txn_comm || comm == acc_comm);
+        if ((xaccSplitGetReconcile (s) != VREC) &&
+            (txn_comm != acc_comm) &&
+            (!comm || comm == txn_comm || comm == acc_comm))
+            rv.push_back (s);
     };
-    std::vector<Split*> rv;
-    auto maybe_accumulate_split = [&rv, match](auto s){ if (match(s)) rv.push_back (s); };
-    for (const auto acc : accounts)
-        gnc_account_foreach_split (acc, maybe_accumulate_split, true);
+
+    std::function<void(Account*)> scan_account;
+    if (use_end_date)
+        scan_account = [end_date, maybe_accumulate](auto acc)
+            { gnc_account_foreach_split_until_date (acc, end_date, maybe_accumulate); };
+    else
+        scan_account = [maybe_accumulate](auto acc)
+            { gnc_account_foreach_split (acc, maybe_accumulate, false); };
+
+    std::for_each (accounts.begin(), accounts.end(), scan_account);
     if (sort)
         std::sort (rv.begin(), rv.end(), [](auto a, auto b){ return xaccSplitOrder (a, b) < 0; });
     return rv;
--- a/libgnucash/engine/Account.cpp
+++ b/libgnucash/engine/Account.cpp
@@ -1151,6 +1151,22 @@ gnc_account_foreach_split (const Account *acc, std::function<void(Split*)> func,
         std::for_each(splits.begin(), splits.end(), func);
 }
 
+void
+gnc_account_foreach_split_until_date (const Account *acc, time64 end_date,
+                                      std::function<void(Split*)> f)
+{
+    if (!GNC_IS_ACCOUNT (acc))
+        return;
+
+    auto after_date = [](time64 end_date, auto s) -> bool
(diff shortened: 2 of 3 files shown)


bindings/engine.i
```diff
@@ -129,19 +129,20 @@ static const GncGUID * gncBudgetGetGUID(GncBudget *x)
...
-        return (txn_comm != acc_comm) && (!comm || comm == txn_comm || comm == acc_comm);
+        if ((txn_comm != acc_comm) && (!comm || comm == txn_comm || comm == acc_comm))
+            rv.push_back (const_cast<Split*>(s));
+        return false;
     };
-    std::vector<Split*> rv;
-    auto maybe_accumulate_split = [&rv, match](auto s){ if (match(s)) rv.push_back (s); };
-    for (const auto acc : accounts)
-        gnc_account_foreach_split (acc, maybe_accumulate_split, true);
+    auto scan_account = [&](auto acc){ gnc_account_find_split (acc, maybe_accumulate, false); };
```
jralls: Yuk. Passing a function with side effects to `gnc_account_find_split`'s predicate is ugly and obscure. 

To correctly indicate what you want we need a `gnc_account_foreach_split_until` function. Unfortunately I think you need C++20's ranges library to implement that efficiently with standard algorithms. A proper use of std::find_if and std::accumulate or std::for_each would loop over the desired part of the vector twice, once to find the last desired split and the second to do the accumulation, so you'd have to write a loop with an escape, something like 
```
template <typename IterType, typename FuncType, typename PredType> void
foreach_until(IterType begin, IterType end, FuncType func, PredType pred)
/* Insert appropriate enable_ifs to constrain the types */
{
    for (auto local_iter = begin; local_iter != end && pred(local_iter); ++local_iter)
       func(local_iter);
}
```

christopherlam: In this case it'd be `std::upper_bound` to find the date, then `std::for_each`. Further Account.h methods would be needed.

jralls: Except that std::upper_bound and friends work with forward iterators only and you want to be able to use reverse iterators.

christopherlam: IMO this limitation is acceptable...

libgnucash/engine/Account.cpp
```diff
@@ -1151,6 +1151,19 @@ gnc_account_foreach_split (const Account *acc, std::function<void(Split*)> func,
         std::for_each(splits.begin(), splits.end(), func);
 }
 
+void
+gnc_account_foreach_until_date (const Account *acc, time64 end_date, std::function<void(Split*)> f)
+{
+    auto splits{GET_PRIVATE(acc)->splits};
+
+    auto is_before = [](time64 end_date, auto s) -> bool
+    { return (xaccTransGetDate(xaccSplitGetParent (s)) <= end_date); };
```
jralls: That makes it "is_before_or_on" unless `end_date` is after 10:59:59Z on the day.

bindings/engine.i
```diff
@@ -130,18 +130,25 @@ SplitsVec gnc_get_match_commodity_splits (AccountVec accounts, bool use_end_date
...
         auto acc_comm{xaccAccountGetCommodity (xaccSplitGetAccount (s))};
-        if ((txn_comm != acc_comm) && (!comm || comm == txn_comm || comm == acc_comm))
-            rv.push_back (const_cast<Split*>(s));
-        return false;
+        if ((xaccSplitGetReconcile (s) != VREC) &&
+            (txn_comm != acc_comm) &&
+            (!comm || comm == txn_comm || comm == acc_comm))
+            rv.push_back (s);
     };
-    auto scan_account = [&](auto acc){ gnc_account_find_split (acc, maybe_accumulate, false); };
```
jralls: Keep the decl down here next to the first use.

christopherlam:
no idea why it won't compile on github... all passes well on my pop_os end.

jralls:
> no idea why it won't compile on github... all passes well on my pop_os end.

It's just macOS (so far anyway). That's the problem [I posted about on IRC](https://code.gnucash.org/logs/2024/04/24.html#T00:40:34).

jralls:
What's going wrong with `std::lower_bound`? It's not really the distro, of course, it's some version of libstdc++ that has a problem. Does it work the way you expect in some other version?

christopherlam:
> What's going wrong with `std::lower_bound`? It's not really the distro, of course, it's some version of libstdc++ that has a problem. Does it work the way you expect in some other version?

Beats me... Try reverting the fix-up commit and push: The Ubuntu 22.04 ci shows lots of test errors proving that the lower_bound approach is giving incorrect results!

christopherlam:
https://github.com/Gnucash/gnucash/actions/runs/8865090288/job/24340972868

jralls:
> he Ubuntu 22.04 ci shows lots of test errors proving that the lower_bound approach is giving incorrect results!

You got the comparison backwards in `is_after_date`. [`std::lower_bound`](https://en.cppreference.com/w/cpp/algorithm/lower_bound) returns the first iterator where the passed-in function returns false so the comparison needs to be `txn_date <= end_date`. Alternatively change to `std::upper_bound` and reverse the parameters in `is_after_date`. I think the latter better expresses the intent because you're evaluating `[begin(), is_after_date_iter)`; lower_bound would make sense for `[is_after_date_iter, end())`.

christopherlam:
> > he Ubuntu 22.04 ci shows lots of test errors proving that the lower_bound approach is giving incorrect results!
> 
> You got the comparison backwards in `is_after_date`. [`std::lower_bound`](https://en.cppreference.com/w/cpp/algorithm/lower_bound) returns the first iterator where the passed-in function returns false so the comparison needs to be `txn_date <= end_date`. Alternatively change to `std::upper_bound` and reverse the parameters in `is_after_date`. I think the latter better expresses the intent because you're evaluating `[begin(), is_after_date_iter)`; lower_bound would make sense for `[is_after_date_iter, end())`.

Ok you're right :-)

christopherlam:
The next question is why did arch ci pass the tests then...

PR #1932, opened 2024-04-30
[dialog-price-edit-db.cpp] Copy the latest price when creating a new entry from a commodity selection

boomer41:
When a price entry is selected, it's contents get copied 1:1 to the newly created price to facilitate faster entry based on existing values (e.g. the type).
When selecting a commodity instead, the user is presented with an empty price entry where only the commodity is set.

IMO the user ends up with the same values (except the price of course) as the latest entry.
So we check whether we have a (latest) entry and use that as the template,
so that e.g. the type gets copied, too.
This spares the user some clicks.

### DIFF
--- a/gnucash/gnome/dialog-price-edit-db.cpp
+++ b/gnucash/gnome/dialog-price-edit-db.cpp
@@ -532,9 +532,23 @@ gnc_prices_dialog_add_clicked (GtkWidget *widget, gpointer data)
     {
         if (!gnc_list_length_cmp (comm_list, 1)) // make sure it is only one parent
         {
-            price = gnc_price_create (pdb_dialog->book);
-            auto comm = static_cast<gnc_commodity *> (comm_list->data);
-            gnc_price_set_commodity (price, comm);
+            auto comm = GNC_COMMODITY (comm_list->data);
+            auto latest_price = gnc_pricedb_lookup_latest_any_currency (pdb_dialog->price_db, comm);
+
+            if (latest_price)
+            {
+                price = GNC_PRICE (latest_price->data);
+                gnc_price_ref (price);
+
+                gnc_price_list_destroy (latest_price);
+            }
+
+            if (!price)
+            {
+                price = gnc_price_create (pdb_dialog->book);
+                gnc_price_set_commodity (price, comm);
+            }
+
             unref_price = TRUE;
         }
         g_list_free (comm_list);


gnucash/gnome/dialog-price-edit-db.cpp
```diff
@@ -532,9 +532,23 @@ gnc_prices_dialog_add_clicked (GtkWidget *widget, gpointer data)
...
         {
-            price = gnc_price_create (pdb_dialog->book);
             auto comm = static_cast<gnc_commodity *> (comm_list->data);
-            gnc_price_set_commodity (price, comm);
+            auto latest_price = gnc_pricedb_lookup_latest_any_currency (pdb_dialog->price_db, comm);
+
+            if (latest_price)
+            {
+                price = static_cast<GNCPrice *> (latest_price->data);
+                gnc_price_ref (price);
```
jralls: This will edit the existing price instead of creating a new one; you want `gnc_price_clone` instead. It's null-checked so it's safe to say
```
price = gnc_price_clone(GNC_PRICE(latest_price->data), pdb_dialog->book);
```
It's created with a ref count of 1 so there's no need to ref it. Notice the cast macro: GncPrices are GObjects so you should always use the cast macro that checks its GType instead of a static_cast or C-cast.

jralls:
> IMO the user ends up with the same values (except the price of course) as the latest entry.

There's nothing here that empties the price. Was it your intention to? Might it make sense to change the date to today instead of defaulting to the last price's date?

gnucash/gnome/dialog-price-edit-db.cpp
boomer41: See gnc_price_edit_dialog on L636, which will be called by the code you mentioned some lines below:
The price is copied there, only a reference is given. Or am I overseeing something?

```c++
    case GNC_PRICE_NEW:
        if (price)
        {
            price = gnc_price_clone(price, pedit_dialog->book);

            gnc_price_set_source (price, PRICE_SOURCE_EDIT_DLG);
            gnc_price_set_time64 (price, gnc_time (NULL));
            gnc_price_set_value (price, gnc_numeric_zero ());
        }
```

boomer41:
> > IMO the user ends up with the same values (except the price of course) as the latest entry.
> 
> There's nothing here that empties the price. Was it your intention to? Might it make sense to change the date to today instead of defaulting to the last price's date?

See my reply in the discussion thread. It will be zeroed.

gnucash/gnome/dialog-price-edit-db.cpp
boomer41: I changed the casts to the GLib-Style-Casts as requested.

jralls: > See gnc_price_edit_dialog on L636

OK. The memory management is a bit tortured but that's not your doing.

jralls:
Thanks!

boomer41:
> Thanks!

You may have misspelled my name in the merge commit message, but all good :D
Thx!

PR #1936, opened 2024-05-09
Bug 799305 - Crash when there is more than one unknown quote source for commodities

christopherlam:
instead of vector which requires reallocation, use a c++ linked list.

### DIFF
--- a/libgnucash/engine/test/test-commodities.cpp
+++ b/libgnucash/engine/test/test-commodities.cpp
@@ -23,6 +23,7 @@
  *  02110-1301, USA.
  */
 #include <glib.h>
+#include <string>
 
 #include <config.h>
 #include "gnc-commodity.h"
@@ -53,6 +54,17 @@ test_quote_sources ()
 
     // internal name:
     do_test (gnc_quote_source_lookup_by_internal("treasure") == nullptr, "lookup_by_internal: treasure doesn't exist");
+
+    auto first = gnc_quote_source_lookup_by_ti (SOURCE_UNKNOWN, 0);
+    g_assert (first != nullptr);
+
+    auto name = gnc_quote_source_get_user_name (first);
+    do_test (!g_strcmp0 (name, "test-source"), "get_user_name: name is as expected");
+
+    for (auto i = 0; i < 20; ++i)
+        gnc_quote_source_add_new (std::string(std::to_string(i)).c_str(), false);
+
+    do_test (gnc_quote_source_get_user_name (first) == name, "get_user_name hasn't moved");
 }
 
 static void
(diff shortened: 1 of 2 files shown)


libgnucash/engine/test/test-commodities.cpp
```diff
@@ -53,6 +53,21 @@ test_quote_sources ()
 
     // internal name:
     do_test (gnc_quote_source_lookup_by_internal("treasure") == nullptr, "lookup_by_internal: treasure doesn't exist");
+
+    auto first =gnc_quote_source_lookup_by_ti (SOURCE_UNKNOWN, 0);
```
jralls: whitespace error.

libgnucash/engine/test/test-commodities.cpp
```diff
@@ -53,6 +53,21 @@ test_quote_sources ()
...
+    g_assert (first != nullptr);
+    auto name = gnc_quote_source_get_user_name (first);
+    do_test (!g_strcmp0 (name, "test-source"), "get_user_name: name is as expected");
+    printf ("name=%s\n", name);
+
+    for (auto i = 0; i < 10000; ++i)
```
jralls: You don't need anywhere near that many iterations. Somewhere in the 10-20 range should be sufficient; it will depend a bit on the c++ library implementation.

jralls (changes requested):
The first commit duplicates #1900, but something's fishy about it: CMakeLists.txt [specifies C++ 17](https://github.com/Gnucash/gnucash/blob/f3b203f4dbd4e9bb02ebecf8b9beaf19b57cbd76/CMakeLists.txt#L599), not C++ 20. Is that not making it into the flags passed to the compiler or has a recent release of GCC decided that C++20 compiance is required regardless of what's passed as `-std=`?

christopherlam:
> The first commit duplicates #1900, but something's fishy about it: CMakeLists.txt [specifies C++ 17](https://github.com/Gnucash/gnucash/blob/f3b203f4dbd4e9bb02ebecf8b9beaf19b57cbd76/CMakeLists.txt#L599), not C++ 20. Is that not making it into the flags passed to the compiler or has a recent release of GCC decided that C++20 compiance is required regardless of what's passed as `-std=`?

Dunno latest ci is failing on arch only

jralls:
It's a GCC bug. The [documentation](https://gcc.gnu.org/onlinedocs/gcc-14.1.0/gcc/C_002b_002b-Dialect-Options.html) says "-Wtemplate-id-cdtor is enabled by default with -std=c++20; it is also enabled by -Wc++20-compat", the args passed to the compiler say `-std=c++17`.

I don't have a gcc bugzilla account and you have to request one by email so I'm not going to file one. I've merged #1900 so the problem goes away.

PR #1939, opened 2024-05-15
Bug 799298 - Shortcut Ctrl-G does not work in the General Journal 

Bob-IT:
I think this fixes the bug which is caused by the blank splits of other open registers being included in the General Journal split list and so have to be excluded.

While looking at this I noticed an inconsistency in the jump to blank action. In the General Journal, the highlighted cell is the split `action` field whereas I think it should be the `date` field. I vaguely remember a bug for it that I found, bug798822 but strangely it referenced an already fixed bug.

### DIFF
--- a/gnucash/gnome/gnc-plugin-page-register.cpp
+++ b/gnucash/gnome/gnc-plugin-page-register.cpp
@@ -4674,6 +4674,15 @@ gnc_plugin_page_register_cmd_blank_transaction (GSimpleAction *simple,
     LEAVE (" ");
 }
 
+static bool
+find_after_date (Split *split, time64* find_date)
+{
+    auto trans = xaccSplitGetParent (split);
+    return !(xaccSplitGetAccount (split) != nullptr && 
+             xaccTransGetDate (trans) >= *find_date &&
+             xaccTransCountSplits (trans) != 1);
+}
+
 static void
 gnc_plugin_page_register_cmd_goto_date (GSimpleAction *simple,
                                         GVariant      *paramter,
@@ -4689,6 +4698,7 @@ gnc_plugin_page_register_cmd_goto_date (GSimpleAction *simple,
     g_return_if_fail (GNC_IS_PLUGIN_PAGE_REGISTER (page));
 
     auto date = input_date (window, _("Go to Date"), _("Go to Date"));
+
     if (!date)
     {
         LEAVE ("goto_date cancelled");
@@ -4700,14 +4710,14 @@ gnc_plugin_page_register_cmd_goto_date (GSimpleAction *simple,
     splits = g_list_copy (qof_query_run (query));
     splits = g_list_sort (splits, (GCompareFunc)xaccSplitOrder);
 
-    for (GList *lp = splits; lp; lp = lp->next)
-    {
-        if (xaccTransGetDate (xaccSplitGetParent (GNC_SPLIT(lp->data))) >= date.value())
-        {
-            gnc_split_reg_jump_to_split (gsr, GNC_SPLIT(lp->data));
-            break;
-        }
-    }
+    // if gl register, there could be blank splits from other open registers
+    // included in split list so check for and ignore them
+    auto it = g_list_find_custom (splits, &date.value(), (GCompareFunc)find_after_date);
+
+    if (it)
+        gnc_split_reg_jump_to_split (gsr, GNC_SPLIT(it->data));
+    else
+        gnc_split_reg_jump_to_blank (gsr);
 
     g_list_free (splits);
     LEAVE (" ");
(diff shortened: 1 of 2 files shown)


gnucash/gnome/gnc-plugin-page-register.cpp
```diff
@@ -4702,12 +4704,23 @@ gnc_plugin_page_register_cmd_goto_date (GSimpleAction *simple,
...
+            if (xaccTransGetDate (trans) >= date.value())
+            {
+                found = true;
+                gnc_split_reg_jump_to_split (gsr, split);
+                break;
+            }
         }
     }
+    if (!found)
+       gnc_split_reg_jump_to_blank (gsr);
```
christopherlam: haven't tested:

```patch
diff --git a/gnucash/gnome/gnc-plugin-page-register.cpp b/gnucash/gnome/gnc-plugin-page-register.cpp
index 92732a4be3..b5f137c7c6 100644
--- a/gnucash/gnome/gnc-plugin-page-register.cpp
+++ b/gnucash/gnome/gnc-plugin-page-register.cpp
@@ -4674,6 +4674,14 @@ gnc_plugin_page_register_cmd_blank_transaction (GSimpleAction *simple,
     LEAVE (" ");
 }
 
+static bool
+find_after_date (Split *split, time64* find_date)
+{
+    auto trans = xaccSplitGetParent (split);
+    return !(xaccSplitGetAccount (split) != nullptr && xaccTransGetDate (trans) >= *find_date &&
+             xaccTransCountSplits (trans) != 1);
+}
+
 static void
 gnc_plugin_page_register_cmd_goto_date (GSimpleAction *simple,
                                         GVariant      *paramter,
@@ -4700,14 +4708,12 @@ gnc_plugin_page_register_cmd_goto_date (GSimpleAction *simple,
     splits = g_list_copy (qof_query_run (query));
     splits = g_list_sort (splits, (GCompareFunc)xaccSplitOrder);
 
-    for (GList *lp = splits; lp; lp = lp->next)
-    {
-        if (xaccTransGetDate (xaccSplitGetParent (GNC_SPLIT(lp->data))) >= date.value())
-        {
-            gnc_split_reg_jump_to_split (gsr, GNC_SPLIT(lp->data));
-            break;
-        }
-    }
+    auto it = g_list_find_custom (splits, &date.value(), (GCompareFunc)find_after_date);
+
+    if (it)
+        gnc_split_reg_jump_to_split (gsr, GNC_SPLIT(it->data));
+    else
+        gnc_split_reg_jump_to_blank (gsr);
 
     g_list_free (splits);
     LEAVE (" ");
```

christopherlam: Hmmmm then we'd worry about 32 bit gpointer shenanigans

christopherlam: ok patch amended, should be fine on 32-bit

jralls: @christopherlam TBC the issue with GINT_TO/FROM_POINTER is that it's a 32-bit operation because int is a 32-bit type, at least on the architectures we're interested in. You can't use it with time64 even where pointers are 64-bit because you can't be sure that the compiler won't cast the time64 to int, discarding the top half, before casting the result to a pointer.

gnucash/gnome/gnc-plugin-page-register.cpp
```diff
@@ -4684,11 +4684,13 @@ gnc_plugin_page_register_cmd_goto_date (GSimpleAction *simple,
     Query* query;
     GList *splits;
     GtkWidget *window = gnc_plugin_page_get_window (GNC_PLUGIN_PAGE (page));
+    gboolean found = false;
```
jralls: Don't mix types. Either (preferred) `bool found = false` or `gboolean found = FALSE`.

gnucash/gnome/gnc-plugin-page-register.cpp
christopherlam: > @christopherlam TBC the issue with GINT_TO/FROM_POINTER is that it's a 32-bit operation because int is a 32-bit type, at least on the architectures we're interested in. You can't use it with time64 even where pointers are 64-bit because you can't be sure that the compiler won't cast the time64 to int, discarding the top half, before casting the result to a pointer.

Is the amended patch any better? No GINT_ macros

jralls: Yes, much. I just think it's important for you to understand the underlying problem with casting between ints and pointers. The same problem applies to casting between numeric types of different sizes like doubles to ints: Bits get lost.

gnucash/gnome/gnc-plugin-page-register.cpp
Bob-IT: Missed that, thanks.

gnucash/gnome/gnc-plugin-page-register.cpp
Bob-IT: I have updated the PR with Chris's suggested change.

jralls:
@Bob-IT Looks fine, merge when ready.

PR #1940, opened 2024-05-18
Statusbar during ofx download

christopherlam:
While processing ofx download, it will delete previously imported[*] transactions. Due to the slow transaction deletion performance, this delay may be significant e.g. ~4 seconds while processing an ofx with only 63 transactions.

Add a statusbar message to show the UI isn't frozen. 

Yak shaving requested: any better wording for the statusbar message?

I thought about utilising the progress bar, however, due to the deletion of transactions, it causes the UI to update too often[**].

image

[*] `gnc_gen_trans_list_add_trans` will check `(gnc_import_exists_online_id (trans, gui->acct_id_hash))` which triggers the transaction deletion.
[**] try this patch to test:
```patch
modified   gnucash/import-export/ofx/gnc-ofx-import.cpp
@@ -1304,6 +1304,7 @@ runMatcher (ofx_info* info, char * selected_filename, gboolean go_to_next_file)
      * create duplicate entries.
      */
     info->num_trans_processed = 0;
+    auto len = g_list_length (info->trans_list);
 
     gnc_window_show_progress (_("Processing and removing already imported transactions..."), 100);
 
@@ -1314,6 +1315,7 @@ runMatcher (ofx_info* info, char * selected_filename, gboolean go_to_next_file)
     // appear.
     for(GList* node = info->trans_list; node; node=node->next)
     {
+        gnc_window_show_progress ("Processing transactions", info->num_trans_processed * 100 / len);
         auto trans = static_cast<Transaction*>(node->data);
         Split* split = xaccTransGetSplit (trans, 0);
         Account* account = xaccSplitGetAccount (split);
```

### DIFF
--- a/gnucash/import-export/ofx/gnc-ofx-import.cpp
+++ b/gnucash/import-export/ofx/gnc-ofx-import.cpp
@@ -49,6 +49,7 @@
 #include "gnc-glib-utils.h"
 #include "gnc-prefs.h"
 #include "gnc-ui.h"
+#include "gnc-window.h"
 #include "dialog-account.h"
 #include "dialog-utils.h"
 #include "window-reconcile.h"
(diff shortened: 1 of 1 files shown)


jralls:
The function you put that in de-duplicates the imported transaction list and adds the transactions to the match list. Removing already-imported transactions is done by the matcher with that match list, so I'd go with "Removing duplicate transactions".

PR #1947, opened 2024-06-01
[SX-ttinfo.cpp] tidier, using c++ stl and algo

christopherlam:


libgnucash/engine/SX-ttinfo.cpp
```diff
@@ -22,337 +22,235 @@
...
+    void operator= (const char* str) { m_str = str ? std::make_optional<std::string>(str) : std::nullopt; };
+    void reset () { m_str = std::nullopt; };
+private:
+    std::optional<std::string> m_str;
+};
 
-    GList *splits; /* list of template splits, owned by us */
+static void delete_splitinfo (TTSplitInfo* tt_info);
+
+struct TTInfo_s
```
jralls: Lose the `_s` and the `typedef TTInfo_s TTInfo;`, ditto for `TTSplitInfo`.

libgnucash/engine/SX-ttinfo.cpp
```diff
@@ -22,337 +22,235 @@
...
+{
+    OptionalString description;
+    OptionalString num;
+    OptionalString notes;
+    gnc_commodity *common_currency = nullptr;
+    TTSplitInfoVec splits;
```
jralls: Instance member variable names [should begin with `m_`](https://wiki.gnucash.org/wiki/CodingStandard#C.2B.2B_Format).

libgnucash/engine/SX-ttinfo.cpp
```diff
@@ -22,337 +22,235 @@
...
-    g_list_free(info->splits);
-
-    g_free(info);
-
-    return;
+    delete info;
 }
 
 void
 gnc_ttinfo_set_description(TTInfo *tti, const char *description)
```
jralls: Make all of the getters and setters class functions or better yet get rid of the setters and use a single constructor.

libgnucash/engine/SX-ttinfo.cpp
```diff
@@ -22,337 +22,235 @@
...
-
-    if (ttsi->credit_formula)
-    {
-        g_free(ttsi->credit_formula);
-        ttsi->credit_formula = NULL;
-    }
-    return;
+    auto str{gnc_numeric_to_string (debit)};
+    gnc_ttsplitinfo_set_debit_formula (ttsi, str);
+    g_free (str);
```
jralls:
```
TTSplitInfo::set_debit_formula_numeric(gnc_numeric(debit))
{
   m_debit_formula = GncNumeric(debit).to_string();
}
```

libgnucash/engine/SX-ttinfo.hpp
```diff
@@ -48,13 +52,13 @@ void gnc_ttinfo_set_currency(TTInfo *tti, gnc_commodity *common_currency);
...
 /* no deep copy occurs - if you want a deep copy make one yourself ! */
-void gnc_ttinfo_set_template_splits(TTInfo *tti, GList *splits);
+void gnc_ttinfo_set_template_splits(TTInfo *tti, TTSplitInfoVec splits);
 
 const char    * gnc_ttinfo_get_description(TTInfo *tti);
 const char    * gnc_ttinfo_get_num(TTInfo *tti);
 const char *gnc_ttinfo_get_notes (TTInfo *tti);
 gnc_commodity * gnc_ttinfo_get_currency(TTInfo *tti);
-GList         * gnc_ttinfo_get_template_splits(TTInfo *tti);
+const TTSplitInfoVec gnc_ttinfo_get_template_splits(TTInfo *tti);
```
jralls: `const TTSplitInfoVec& ...`

libgnucash/engine/SX-ttinfo.hpp
```diff
@@ -48,13 +52,13 @@ void gnc_ttinfo_set_currency(TTInfo *tti, gnc_commodity *common_currency);
 
 
 /* no deep copy occurs - if you want a deep copy make one yourself ! */
-void gnc_ttinfo_set_template_splits(TTInfo *tti, GList *splits);
+void gnc_ttinfo_set_template_splits(TTInfo *tti, TTSplitInfoVec splits);
```
jralls: `TTSplitInfoVec&& splits`

libgnucash/engine/SX-ttinfo.hpp
```diff
@@ -31,9 +31,13 @@
 #include "Account.h"
 #include "gnc-commodity.h"
 
+#include <vector>
+
 typedef struct TTInfo_s TTInfo;
 typedef struct TTSplitInfo_s TTSplitInfo;
 
+using TTSplitInfoVec = std::vector<TTSplitInfo*>;
```
jralls: `std::vector<TTSplitInfo>`. That way the std::vector will take care of managing the memory for you and TTInfo won't need a custom destructor. You can `reserve(2)` in the TTInfo constructor to avoid an expensive reallocation for the vast majority of cases; `reserve(3)` will waste a little memory and cover the investment cases with an empty tracker split. You could test for trading accounts and bump it up 2 more.

libgnucash/engine/SX-ttinfo.cpp
```diff
@@ -22,337 +22,235 @@
...
 }
 
-GList *
+const TTSplitInfoVec
 gnc_ttinfo_get_template_splits(TTInfo *tti)
 {
-    g_return_val_if_fail(tti, NULL);
+    g_return_val_if_fail(tti, TTSplitInfoVec{});
     return tti->splits;
 }
```
jralls: Eww. Make the RV `std::optional<TTSplitInfoVec&>` or better yet replace this with `TTInfo::find_split_with_acct(Account*)` and `TTInfo::add_splits_to_txn(Transaction* trans)`.

libgnucash/engine/SX-ttinfo.cpp
```diff
@@ -22,337 +22,235 @@
...
 gnc_ttinfo_get_currency(TTInfo *tti)
 {
     g_return_val_if_fail(tti, NULL);
-
     return tti->common_currency;
 }
 
 
-void gnc_ttinfo_set_template_splits(TTInfo *tti, GList *splits)
+void gnc_ttinfo_set_template_splits(TTInfo *tti, TTSplitInfoVec splits)
```
jralls: `TTSplitInfoVec&& splits` to move the `TTSplitInfoVec` created in `sxftd_add_template_trans` or better still replace this with `TTInfo::add_template_split(Split*)`.

jralls (changes requested):
The performance benefit here is marginal at best, it's a rare SX transaction that has more than a half-dozen splits, so if you're going to bother with a C++ conversion you might as well make it really C++.

christopherlam:
> The performance benefit here is marginal at best, it's a rare SX transaction that has more than a half-dozen splits, so if you're going to bother with a C++ conversion you might as well make it really C++.

This is mainly a simple file to practise c++ structs... and I'm not sure it's safe to use auto memory management since there's still too many pointers being manipulated: look at the mess in assistant-loan.cpp!

jralls:
> This is mainly a simple file to practise c++ structs... and I'm not sure it's safe to use auto memory management since there's still too many pointers being manipulated: look at the mess in assistant-loan.cpp!

The most important thing to realize about C++ structs is that they're C++ classes where the default exposure is `public` instead of `private`.  That means that all of the rules about compiler-generated constructors apply equally to both and that you should think of them as bundles of data *and the code that operates on that data*.  The less that outside code interacts with the class's internals the less you have to worry about changing those internals breaking things somewhere else.

> and I'm not sure it's safe to use auto memory management since there's still too many pointers being manipulated: look at the mess in assistant-loan.cpp!

Exactly. So part of converting from a C struct to a C++ class is moving all of that pointer manipulation inside the class and providing functions for users of the class to to whatever higher-level thing they need to do.

On the other hand having pointers to the contents of a vector is only a problem if you keep the pointers. Using a pointer in stack context is safe regardless of how the pointed-to object's memory is managed. Keeping pointers is just as much a problem in C as it is in C++ and it's a major source of bugs in GnuCash. Trying to manually manage the object's lifetime is no protection, it's far too error prone.

There is a safe way to keep pointers: shared ownership. It's now considered an obsolete model in C++, but it's also still widely used. Garbage-collected languages are based on it. To use it everything has to be separately allocated on the heap/free store and gets an embedded reference count. When a pointer is copied the reference count is increased and when  one goes out of scope it's decremented. In C++ `std::shared_ptr` and garbage-collected languages that's handled automatically. In C you have to do it manually: That's what `g_object_ref` and `g_object_unref` are about. In garbage-collected languages there's a separate GC thread that periodically looks at all of the pointers and deallocates any that have 0 ref counts; C++ shared pointers and GObjects include code in the dereference function that deallocates when the ref count goes below 1. Note that C-style manually-managed shared ownership is just as error prone as manually managed object lifetime.

libgnucash/engine/SX-ttinfo.hpp
christopherlam: > `std::vector<TTSplitInfo>`. That way the std::vector will take care of managing the memory for you and TTInfo won't need a custom destructor. You can `reserve(2)` in the TTInfo constructor to avoid an expensive reallocation for the vast majority of cases; `reserve(3)` will waste a little memory and cover the investment cases with an empty tracker split. You could test for trading accounts and bump it up 2 more.

How about this branch is reviewed for correctness and the auto memory management postponed to a later branch?

jralls: [Correctness](https://en.wikipedia.org/wiki/Correctness_(computer_science)) is not somewhere you want to go. 

> This is mainly a simple file to practise c++ structs... and I'm not sure it's safe to use auto memory management

If you're going to practice then you should practice doing it the [right way](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rr-newdelete). Automatic memory management exists because manual memory management is inherently unsafe.

christopherlam:
Moving the whole struct into hpp seems necessary so that CPP files know the struct methods. Isit desirable?

jralls:
It's a tradeoff. The alternatives are to wrap the class functions with free functions or to use a [PIMPL](https://en.cppreference.com/w/cpp/language/pimpl) class that exposes only the class functions. 

The argument in favor of exposing the class declaration is that it usually makes for simpler code that the compiler is better able to optimize. If it keeps class instances on the stack it's also inherently more performant because no heap allocations.
The argument against is that the more there is in the header the more likely you'll have to change something in the header and when you do everything that includes that header has to recompile. 

This case is pretty simple, limited in scope, and doesn't appear in any bindings so I lean toward putting the class declaration in the header.

libgnucash/engine/SchedXaction.hpp
```diff
@@ -0,0 +1,36 @@
...
+#ifndef XACC_SCHEDXACTION_HPP
+#define XACC_SCHEDXACTION_HPP
+
+#include "qof.h"
+#include "gnc-engine.h"
+
+#include <vector>
+#include "SX-ttinfo.hpp"
+
+/** \brief Set the schedxaction's template transaction.  tt_vec is a
```
gjanssens, line 30: Since you are using doxygen's `\brief` tag here, you may as well use `\param` to explain the tt_vec parameter.

PR #1949, opened 2024-06-02
Bug 799324 - Avoid invalid free in gvalue_from_kvp_value()

living180:
As of ddc3f28899c71579abd9b2bd822baa6b95fc41d6, `gvalue_from_kvp_value()` takes a `GValue` pointer from the caller, which in some cases points to memory on the stack.  If that is the case and the code also hits the `default` case in the `switch` statement, the unconditional `g_slice_free()` call will attempt to free stack memory, causing the program to abort.

Fix by requiring the caller to always pass in a valid `GValue` pointer, making the caller responsible for freeing it if necessary.  This also means that it is no longer necessary for `gvalue_from_kvp_value()` to return a value, so make it a `void` function.

### DIFF
--- a/libgnucash/engine/kvp-frame.cpp
+++ b/libgnucash/engine/kvp-frame.cpp
@@ -232,14 +232,11 @@ int compare(const KvpFrameImpl & one, const KvpFrameImpl & two) noexcept
     return 0;
 }
 
-GValue*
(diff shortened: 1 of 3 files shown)


jralls:
Thanks!

PR #1951, opened 2024-06-03
[account.cpp] account->children is a vector

christopherlam:
1 step towards better memory management...

### DIFF
--- a/bindings/engine-common.i
+++ b/bindings/engine-common.i
@@ -50,14 +50,10 @@ AccountList * gnc_account_get_descendants (const Account *account);
 %newobject gnc_account_get_descendants_sorted;
 AccountList * gnc_account_get_descendants_sorted (const Account *account);
 
-%newobject gnc_accounts_and_all_descendants;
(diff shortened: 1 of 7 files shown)


christopherlam (commented):
some more c++isms

christopherlam:
This branch causes segfault when deleting an account from ui...
https://lists.gnucash.org/pipermail/gnucash-user/2024-July/112255.html

jralls:
OK. What do you want/intend to do about it?

christopherlam:
Can't find anything particularly wrong so I'd be inclined to revert

jralls:
It's [Bug 799347](https://bugs.gnucash.org/show_bug.cgi?id=799347). I can reproduce it in the debugger but haven't quite figured out what it's trying to dereference that's crashing, nor what exactly is different about what we're handing to the tree model.

jralls:
@christopherlam I think https://github.com/Gnucash/gnucash/commit/60ccca017f1be3af2caebb8a9d0c82a197f1da4f fixes it. Comments should explain why.

PR #1953, opened 2024-06-04
Allow building from a github code zip file (non distro tarball)

gjanssens:
Until now our build system would fail to build the sources downloaded from github as a plain zip file because it had no way of extracting the version data from that zip file.

This PR relaxes this a bit: you can now use such a zip file as source to build gnucash, but it has a few limitations:
1. the version reported by the built gnucash executable is rather vague: it's whatever was set as project version in CMakeLists.txt. So for any zipfile that was downloaded from a non-tagged build, this will only report the gnucash version the zipfile branched off of, but not the actual commit from that branch that was downloaded.
2. this way of building will fail to generate a distribution tarball. I think of that rather as a feature than a bug. Distribution tarballs should only be made from git checkout (and ideally only from tagged commits).

To inform the builder (or a dev helping someone building from such a zip file) a few indicators have been added:
1. when running cmake on an unversioned source tree a warning will be printed
2. the build id of the executable will be amended with "-no-vcs-details"

Some bikeshedding opportunities :wink: : what should the warning message tell the user and what suffix would you like to see in the build-id to indicate this build is not originating from a git checkout or dist-tarball ?

As a sidenote: our ./util/gnc-vcs-info script contains tests for svn, bzr, git and svk. Over time however our build system has slowly been assuming our source tree is in git exclusively. So any objections I clean out that script to only test for git/no git ?


jralls:
What exactly would we be warning the user about? That their build won't be traceable to an exact commit? Will anyone care?

In the bikeshedding spirit how about "-unknown-commit" instead of "-no-vcs-info" for the version suffix?

Yes, please clean out the old VCS cruft. You left that in so as not to break anything during the conversion.

gjanssens:
> What exactly would we be warning the user about? That their build won't be traceable to an exact commit? Will anyone care?

I have a vague recollection of a discussion we had in the past about this. I think the concern was mostly on our side when trying to offer support someone that chose a random github zipfile download. So I guess my warning was meant to be in that sense: tell the user this way of building makes it more difficult to get support. But in retrospect I see this of much less of a concern today than I did back then. There are only few that will attempt to build gnucash and most of them follow the guidelines in the wiki. So I propose to remove the warning.

> 
> In the bikeshedding spirit how about "-unknown-commit" instead of "-no-vcs-info" for the version suffix?

I like your suggestion better than mine. It's more neutral and more expressive. I'll amend the commit.
> 
> Yes, please clean out the old VCS cruft. You left that in so as not to break anything during the conversion.

I don't remember I did. It seems to have been in a previous life :smile: . I'll clean it out, the conversion happened a long time ago.

PR #1955, opened 2024-06-09
qofclass uses stl and ~qofchoice.cpp

christopherlam:
although the qof_choice param lists were being populated, they were never queried.

### DIFF
--- a/libgnucash/engine/qofclass-p.h
+++ b/libgnucash/engine/qofclass-p.h
@@ -32,9 +32,6 @@
 
 #include "qofclass.h"
 
-void qof_class_init(void);
-void qof_class_shutdown (void);
-
 QofSortFunc qof_class_get_default_sort (QofIdTypeConst obj_name);
 
 /* @} */
--- a/libgnucash/engine/qofclass.cpp
+++ b/libgnucash/engine/qofclass.cpp
@@ -22,326 +22,87 @@
 \********************************************************************/
 
 #include <config.h>
-#include <glib.h>
+#include <string>
+#include <algorithm>
+#include <unordered_set>
+#include <unordered_map>
 
 #include "qof.h"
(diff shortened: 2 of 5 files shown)


jralls:
Don't waste time converting QOF to C++, we want to replace it with SQL.

christopherlam:
I'd though it'd be easier to remove layers piecemeal?

jralls:
It depends on the layers. To implement [Roadmap:Database and QOF](https://wiki.gnucash.org/wiki/Roadmap#Database_and_QOF):
1. Write xml<->sql backend with in-memory SQLite db. 
2. Replace all QofQuery calls with SQL queries.
3. Remove QofQuery and the parts of QOF that don't directly support QofSession, QofInstance, and QofCollection. AFAICT QofClass exists only to provide access to engine objects for QofQuery.

I'd like to get that done for GnuCash 6. The big payoff for users is
4. Write a report where the SQL query terms are part of the report options. No more need to write custom reports in Scheme ([Roadmap:Reports](https://wiki.gnucash.org/wiki/Roadmap#Reports)). It would be really sweet to have a drag-and-drop UI for constructing queries like Microsoft Access and formatting the results, but that's a major project all on its own. 

That's the easy part. The more difficult next step is to replace accessing QofCollections with SQL queries. Once that's done we can stop loading the whole DB into memory and lose the rest of QOF. At that point GnuCash is a real SQL app. We can implement record-level locking in the MySql and Postgresql backends and allow simultaneous user access.

christopherlam:
Ok let's remove qofchoice.cpp at the very least.

jralls:
> Ok let's remove qofchoice.cpp at the very least.

Yes please. Cruft removal is encouraged, I'm just trying to keep you from rewriting future cruft into C++.

PR #1956, opened 2024-06-11
add CAGR columns to Investment Lots report

mcbridebt:
Adds Compound Annual Growth Rate (CAGR) columns (realized and unrealized) to the Investment Lots report.

Currently, the report shows investment gain/loss percentages via the Return on Investment (ROI) columns (both realized and unrealized). This can be useful in some situations, but ROI has its limitations. It does not recognize the time value of money. But CAGR does. It is useful for the report to support both ROI and CAGR.

For example, suppose there are two investment lots, [A] and [B]. Both were purchased for $100 and sold for $110. At first glance, it might seem that they performed equally. The ROI for each is 10% (gain divided by basis). But suppose [A] was held for only one year before being sold while [B] was held for 10 years. Clearly [A] grew at a much faster rate than [B], which ROI does not reflect. The CAGR is 10% for [A] but only 0.96% for [B], which is much more illustrative of how much [A] has outperformed [B].

See https://www.investopedia.com/terms/c/cagr.asp for CAGR description and formula. 
Bug: https://bugs.gnucash.org/show_bug.cgi?id=799335

New options:
image

New table columns:
image

TODO: Update report documentation for the new columns

### DIFF
--- a/gnucash/report/reports/standard/investment-lots.scm
+++ b/gnucash/report/reports/standard/investment-lots.scm
@@ -60,12 +60,14 @@
 (define optname-show-sold-columns (N_ "Show sold columns"))
 (define optname-show-end-columns (N_ "Show end columns"))
 (define optname-show-realized-gain-columns
-    (N_ "Show realized gain column(s)"))
+    (N_ "Show realized gain columns"))
 (define optname-show-unrealized-gain-columns
-    (N_ "Show unrealized gain column(s)"))
+    (N_ "Show unrealized gain columns"))
 (define optname-group-gains-by-age
     (N_ "Group gains by age (short term and long term)"))
 (define optname-long-term-years (N_ "Long term gains age (years)"))
+(define optname-show-roi-columns (N_ "Show ROI columns"))
+(define optname-show-cagr-columns (N_ "Show CAGR columns"))
 
 ;; Display
 (define optname-show-long-account-names (N_ "Show long account names"))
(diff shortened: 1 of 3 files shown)


christopherlam:
Hi I'm the resident scheme maintainer.

I don't have any opinion regarding the calculations... I'll assume that they are correct; to learn enough scheme to create a custom report sets a non-trivial bar of entry ;-)

However: can the report handle the full gamut of stock-type activity? ie. buy/sell, dividend payment, stock split, reverse stock split, return of capital, notional distribution? and the short-position counterparts? If not, it may need to be explicit about safely bailing out when encountering these scenarios. Example: see advanced-portfolio.scm which cannot handle short positions, and will simply bail out if the stock balance is negative.

I'll have an opinion regarding programming style: you'll notice most existing reports will _minimise_ the use of `set!` in the report. `set!` can redefine any variable, whether local or global; and makes the report arguably more difficult to follow. Ideally they would be rewritten; the scheme idiom is to use named let loops to achieve similar results.

Moreover there should be tests. See example at test/test-ifrs-cost-basis.scm to create stock transactions and tests the ifrs report output. Note the ifrs cost basis report doesn't use lots. See test-new-owner-report.scm which has some lot-creation routines.

jralls:
@mcbridebt You haven't responded to @christopherlam's  question. It has been two years. Is this a drive-by?

mcbridebt:
I've updated the PR to:
1. Minimize "set!" usage, and
2.  Add tests, which involved refactoring report code to be more testable

Regarding the activity types: split classification consults xaccSplitGetType and xaccSplitGetAction in addition to amount sign, with an explicit non-realizing-share-reduction path. That fixes reverse stock splits, which previously produced a phantom realized loss. Buy/sell, forward splits, and reinvested dividends remain correct on the existing purchase/sale paths. Cash dividends are not lot events. Return of capital, notional distributions, and short positions are flagged by the existing validation warnings (warn-if-gains-mismatch, warn-if-balance-negative). I believe the report behaves correctly across all the cases you listed.

mcbridebt:
@christopherlam  please let me know if the latest changes resolve your concerns. Thanks!

jralls:
> @christopherlam please let me know if the latest changes resolve your concerns. Thanks!

Silence is assent! Thanks, @mcbridebt.

PR #1961, opened 2024-06-17
Bug799047 - AutoComplete Only Considers Visible Transaction

Bob-IT:
In the past, all transactions were used on the first load of the register and this is when the autocomplete would be populated, on
subsequent loads it would be a filtered list. With the change to delay loading the register, the first load split list is empty and the autocomplete is now being populated by the filtered list.

To fix this, make a copy of the original query and compare this query to the query used for refresh. When different, pass a second full list of splits to the register to populate the autocomplete lists.

### DIFF
--- a/gnucash/register/ledger-core/gnc-ledger-display.c
+++ b/gnucash/register/ledger-core/gnc-ledger-display.c
@@ -59,6 +59,7 @@ struct gnc_ledger_display
     GncGUID leader;
 
     Query* query;
+    Query* pre_filter_query;
 
     GNCLedgerDisplayType ld_type;
 
@@ -655,6 +656,9 @@ close_handler (gpointer user_data)
     qof_query_destroy (ld->query);
     ld->query = NULL;
 
+    qof_query_destroy (ld->pre_filter_query);
+    ld->pre_filter_query = NULL;
+
     g_free (ld);
 }
 
@@ -834,6 +838,8 @@ gnc_ledger_display_internal (Account* lead_account, Query* q,
     else
         gnc_ledger_display_make_query (ld, limit, reg_type);
 
+    ld->pre_filter_query = qof_query_copy (ld->query);
+
     ld->component_id = gnc_register_gui_component (klass,
                                                    refresh_handler,
                                                    close_handler, ld);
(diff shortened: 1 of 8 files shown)


gnucash/register/ledger-core/gnc-ledger-display.c
```diff
@@ -466,9 +465,9 @@ gnc_ledger_display_gl (void)
     // Exclude any template accounts for search register and gl
      exclude_template_accounts (query, exclude_template_accounts_hash);
 
-    gnc_tm_get_today_start (&tm);
-    tm.tm_mon--; /* Default the register to the last month's worth of transactions. */
-    start = gnc_mktime (&tm);
+    start = gnc_time64_get_today_start ();
+    start = start - 2592000; // the default is to show last 30 days
```
jralls: Let's not have the magic number:
```C
static const time64 secs_per_thirty_days = 2592000;
start = gnc_time64_get_today_start () - secs_per_thirty_days;
```

gnucash/register/ledger-core/split-register-load.c
```diff
@@ -256,8 +256,9 @@ _find_split_with_parent_txn (gconstpointer a, gconstpointer b)
     return xaccSplitGetParent (split) == txn ? 0 : 1;
 }
 
-static void add_quickfill_completions (TableLayout* layout, Transaction* trans,
-                                       Split* split, gboolean has_last_num)
+static void
+add_completions (TableLayout* layout, Transaction* trans,
+                 Split* split, gboolean has_last_num)
```
jralls: Why the name change? It's adding completions to quick fill cells.

jralls (changes requested):

gnucash/register/ledger-core/gnc-ledger-display.c
Bob-IT: Changed, thank you.

gnucash/register/ledger-core/split-register-load.c
Bob-IT: I think I changed the name when I was going to use it for all completions but decided to do it as this PR but forgot to change it back, done.

jralls:
Thanks.

PR #1962, opened 2024-06-17
Bug 799334 - GnuCash re-opens to incorrect page

Bob-IT:
If there are transient pages open on close, like search result pages, they are not restored which results in the wrong page being selected on restore.

To fix this, remember the pages that are not restored so the position of pages in the saved order can be adjusted.

### DIFF
--- a/gnucash/gnome-utils/gnc-main-window.cpp
+++ b/gnucash/gnome-utils/gnc-main-window.cpp
@@ -435,13 +435,16 @@ gnc_main_window_foreach_page (GncMainWindowPageFunc fn, gpointer user_data)
  *  installed.
  *
  *  @param data A data structure containing state about the
- *  window/page restoration process. */
-static void
+ *  window/page restoration process.
+ * 
+ *  @return true if page was added else false.
+ *  */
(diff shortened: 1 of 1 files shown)


jralls:
@Bob-IT wouldn't it make more sense to stop the temporary tabs from being added to the .gcm?

Bob-IT:
> @Bob-IT wouldn't it make more sense to stop the temporary tabs from being added to the .gcm?

@jralls  I did wonder about this and I decided to do it on the restore for the following reasons...

- On save, you would need to loop over all the open pages to decide which to exclude, adjust total number of pages, exclude the pages from the page order and then loop over the pages again to add the page sections with amended page references.
- Doing on restore seemed simpler.
- I have often pondered whether one could actually save the search on an open search page.
- Lastly, doing it on restore takes care of a variation to search pages, having two computers with separate .gcm files sharing the same data file and with the same open account saved in the open pages list. If that account is deleted on one computer and then the other one used the current page should be corrected.

jralls:
OK, that's reasonable. Thanks.

PR #1965, opened 2024-06-19
Better c++

christopherlam:
to be selectively merged after 5.7.

- adds an overloaded `gnc_account_foreach_descendant` for c++ use.
- use c++ containers instead of creating new GLists

### DIFF
--- a/libgnucash/engine/Account.cpp
+++ b/libgnucash/engine/Account.cpp
@@ -2482,72 +2482,117 @@ xaccAccountSetDescription (Account *acc, const char *str)
 }
 
 static void
-set_kvp_string_path (Account *acc, std::vector<std::string> const & path,
-                     const char *value)
+set_kvp_gnc_numeric_path (Account *acc, const std::vector<std::string>& path,
+                          std::optional<gnc_numeric> value)
 {
-    g_return_if_fail(GNC_IS_ACCOUNT(acc));
-
     xaccAccountBeginEdit(acc);
-    if (value && *value)
-    {
-        GValue v = G_VALUE_INIT;
-        g_value_init (&v, G_TYPE_STRING);
-        g_value_set_static_string (&v, value);
-        qof_instance_set_path_kvp (QOF_INSTANCE (acc), &v, path);
-        g_value_unset (&v);
-    }
-    else
-    {
-         qof_instance_set_path_kvp (QOF_INSTANCE (acc), nullptr, path);
-    }
-    mark_account (acc);
+    qof_instance_set_path_kvp<gnc_numeric> (QOF_INSTANCE(acc), value, path);
     xaccAccountCommitEdit(acc);
 }
 
+static std::optional<gnc_numeric>
+get_kvp_gnc_numeric_path (const Account *acc, const Path& path)
+{
+    return qof_instance_get_path_kvp<gnc_numeric> (QOF_INSTANCE(acc), path);
+}
+
 static void
-set_kvp_string_tag (Account *acc, const char *tag, const char *value)
+set_kvp_string_path (Account *acc, std::vector<std::string> const & path,
+                     const char *value)
 {
-    set_kvp_string_path (acc, {tag}, value);
+    std::optional<const char*> val;
+    if (value && *value)
+        val = g_strdup(value);
+
+    xaccAccountBeginEdit(acc);
+    qof_instance_set_path_kvp<const char*> (QOF_INSTANCE(acc), val, path);
+    xaccAccountCommitEdit(acc);
 }
 
 static const char*
-get_kvp_string_path (const Account *acc, std::vector<std::string> const & path,
-                     GValue *v)
+get_kvp_string_path (const Account *acc, const Path& path)
 {
-    *v = G_VALUE_INIT;
-    if (acc == nullptr) return nullptr; // how to check path is valid??
-    qof_instance_get_path_kvp (QOF_INSTANCE (acc), v, path);
-    return G_VALUE_HOLDS_STRING (v) ? g_value_get_string (v) : nullptr;
+    auto rv{qof_instance_get_path_kvp<const char*> (QOF_INSTANCE(acc), path)};
+    return rv ? *rv : nullptr;
 }
 
-static const char*
-get_kvp_string_tag (const Account *acc, const char *tag, GValue *v)
+static void
+set_kvp_account_path (Account* acc, const Path& path, const Account* kvp_account)
 {
-    return get_kvp_string_path (acc, {tag}, v);
+    std::optional<GncGUID*> val;
+    if (kvp_account)
+        val = guid_copy(xaccAccountGetGUID (kvp_account));
+
+    xaccAccountBeginEdit(acc);
+    qof_instance_set_path_kvp<GncGUID*> (QOF_INSTANCE(acc), val, path);
+    xaccAccountCommitEdit(acc);
+}
+
+static Account*
+get_kvp_account_path (const Account *acc, const Path& path)
+{
+    auto val{qof_instance_get_path_kvp<GncGUID*> (QOF_INSTANCE(acc), path)};
+    return val ? xaccAccountLookup (*val, gnc_account_get_book (acc)) : nullptr;
+}
+
+static void
+set_kvp_boolean_path (Account *acc, const Path& path, gboolean option)
+{
+    set_kvp_string_path (acc, path, option ? "true" : nullptr);
+}
+
+static gboolean
+get_kvp_boolean_path (const Account *acc, const Path& path)
+{
+    auto slot{QOF_INSTANCE(acc)->kvp_data->get_slot(path)};
+    if (!slot) return false;
+    switch (slot->get_type())
+    {
+    case KvpValueImpl::Type::INT64:
+        return slot->get<int64_t>() != 0;
+    case KvpValueImpl::Type::STRING:
+        return g_strcmp0 (slot->get<const char*>(), "true") == 0;
+    default:
+        return false;
+    }
+}
+
+static void
+set_kvp_int64_path (Account *acc, const Path& path, std::optional<gint64> value)
+{
+    xaccAccountBeginEdit(acc);
+    qof_instance_set_path_kvp<int64_t> (QOF_INSTANCE(acc), value, path);
+    xaccAccountCommitEdit(acc);
+}
+
+static const std::optional<gint64>
+get_kvp_int64_path (const Account *acc, const Path& path)
+{
+    return qof_instance_get_path_kvp<int64_t> (QOF_INSTANCE(acc), path);
 }
 
 void
 xaccAccountSetColor (Account *acc, const char *str)
 {
-    set_kvp_string_tag (acc, "color", str);
+    set_kvp_string_path (acc, {"color"}, str);
 }
 
 void
 xaccAccountSetFilter (Account *acc, const char *str)
 {
-    set_kvp_string_tag (acc, "filter", str);
+    set_kvp_string_path (acc, {"filter"}, str);
 }
 
 void
 xaccAccountSetSortOrder (Account *acc, const char *str)
 {
-    set_kvp_string_tag (acc, "sort-order", str);
+    set_kvp_string_path (acc, {"sort-order"}, str);
 }
 
 void
 xaccAccountSetSortReversed (Account *acc, gboolean sortreversed)
 {
-    set_kvp_string_tag (acc, "sort-reversed", sortreversed ? "true" : nullptr);
+    set_kvp_boolean_path (acc, {"sort-reversed"}, sortreversed);
 }
 
 static void
@@ -2571,7 +2616,7 @@ qofAccountSetParent (Account *acc, QofInstance *parent)
 void
 xaccAccountSetNotes (Account *acc, const char *str)
 {
-    set_kvp_string_tag (acc, "notes", str);
+    set_kvp_string_path (acc, {"notes"}, str);
 }
 
 
@@ -2581,25 +2626,7 @@ xaccAccountSetAssociatedAccount (Account *acc, const char *tag, const Account* a
     g_return_if_fail (GNC_IS_ACCOUNT(acc));
     g_return_if_fail (tag && *tag);
 
-    std::vector<std::string> path = { "associated-account", tag };
-    xaccAccountBeginEdit(acc);
-
-    PINFO ("setting %s assoc %s account = %s", xaccAccountGetName (acc), tag,
-           assoc_acct ? xaccAccountGetName (assoc_acct) : nullptr);
-
-    if (GNC_IS_ACCOUNT(assoc_acct))
-    {
-        GValue v = G_VALUE_INIT;
-        g_value_init (&v, GNC_TYPE_GUID);
-        g_value_set_static_boxed (&v, xaccAccountGetGUID (assoc_acct));
-        qof_instance_set_path_kvp (QOF_INSTANCE (acc), &v, path);
-        g_value_unset (&v);
-    }
-    else
-        qof_instance_set_path_kvp (QOF_INSTANCE (acc), nullptr, path);
-
-    mark_account (acc);
-    xaccAccountCommitEdit(acc);
+    set_kvp_account_path (acc, {"associated-account", tag}, assoc_acct);
 }
 
 void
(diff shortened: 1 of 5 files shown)


jralls:
> to be selectively merged after 5.7.

That would be better as multiple PRs. It's ok to make a chain of them, just say in the comments who depends on who.

libgnucash/engine/Account.cpp
```diff
@@ -2753,11 +2753,9 @@ DxaccAccountSetCurrency (Account * acc, gnc_commodity * currency)
 \********************************************************************/
 
 static void
-account_foreach_descendant (const Account *acc, AccountCb thunk,
-                            void* user_data, bool sort)
+account_foreach_descendant (const Account *acc, std::function<void(Account*)> f, bool sort)
```
jralls: Let's have longer function name than `f`. Static analyzers and linters whine about One or two letter function names.

Do *not* make a function do different jobs based on a `bool`. Write separate functions.
I'm not keen on sorting the actual children vector either. Doing that while the user has a tree view open will make a mess. Since the one use case constructs a GList you could do that then sort the list at each level before returning.

gnucash/register/ledger-core/split-register-control.cpp
```diff
@@ -668,23 +669,14 @@ static Split *
...
-
-        split = gnc_find_split_in_trans_by_memo (trans, memo, unit_price);
-
-        if (split)
-            rv = split;
-    }
+    const auto& splits = xaccAccountGetSplits (account);
+    for (auto it = splits.rbegin(); it != splits.rend(); it++)
+        if (auto split = gnc_find_split_in_trans_by_memo (xaccSplitGetParent (*it), memo, unit_price))
+            return split;
```
jralls: Surely you can do better than a raw loop here.

gnucash/gnome/window-reconcile.cpp
```diff
@@ -2258,66 +2258,45 @@ recn_key_press_cb(GtkWidget *widget, GdkEventKey *event, gpointer data)
...
 
-    g_list_free (list);
-    return rv;
-}
-
-typedef void (*AccountProc) (Account *a);
-static void traverse_fn (Account *acct, AccountProc fn)
-{
-    fn (acct);
+    return nullptr;
```
jralls: Can't you do this with nested `find_if` instead of raw loops?

libgnucash/engine/Account.cpp
```diff
@@ -3396,34 +3390,12 @@ xaccAccountGetCommodity (const Account *acc)
 
 gnc_commodity * gnc_account_get_currency_or_parent(const Account* account)
 {
-    gnc_commodity * commodity;
-    g_return_val_if_fail (account, nullptr);
+    g_return_val_if_fail (GNC_IS_ACCOUNT (account), nullptr);
+
+    for (auto acc = account; acc; acc = gnc_account_get_parent (acc))
+        if (auto comm = xaccAccountGetCommodity (acc); gnc_commodity_is_currency (comm))
+            return comm;
```
jralls, line 3372: std::find_if

jralls (changes requested):

libgnucash/engine/Account.cpp
christopherlam: find_if wants c++ iterators though. this is acc climbing the parent tree.

gnucash/gnome/window-reconcile.cpp
christopherlam: not yet because `xaccTransGetSplitList` isn't yet a vector

gnucash/register/ledger-core/split-register-control.cpp
christopherlam: find_if will search correctly; but it returns the iterator rather than the result of the predicate. hence the above construct to avoid calling `gnc_find_split_in_trans_by_memo` again

libgnucash/engine/Account.cpp
jralls: Oh, right.

gnucash/register/ledger-core/split-register-control.cpp
jralls: Mrmmph. Transaction split lists tend to be very short so it's not a big deal to loop over one twice, but this is sufficiently simple that I guess there's not much clarity gained with an algo.

libgnucash/engine/Account.cpp
```diff
@@ -4826,196 +4860,86 @@ xaccAccountSetLastNum (Account *acc, const char *num)
...
+    if (priv_limit.has_value())
     {
-        *balance = GET_PRIVATE(acc)->higher_balance_limit.value();
-
-        if (gnc_numeric_check (*balance) == 0)
-            return true;
-        else
-            return false;
+        auto limit = *priv_limit;
+        if (limit) *balance = *limit;
```
jralls: That relies on an implicit cast from gnc_numeric to GncNumeric. Better to say `*balance = gnc_numeric_create(limit->num, limit->denom);`

libgnucash/engine/Account.cpp
```diff
@@ -4826,196 +4860,86 @@ xaccAccountSetLastNum (Account *acc, const char *num)
...
+        return limit.has_value();
     }
-    else
+    else if (auto priv_limit = get_kvp_gnc_numeric_path (acc, { KEY_BALANCE_LIMIT, key }))
```
jralls: Please keep KVP implementation details in QofInstance.

libgnucash/engine/AccountP.hpp
```diff
@@ -113,8 +113,10 @@ typedef struct AccountPrivate
     gnc_numeric cleared_balance;
     gnc_numeric reconciled_balance;
 
-    std::optional<gnc_numeric> higher_balance_limit;
-    std::optional<gnc_numeric> lower_balance_limit;
+    // outer optional to denote if balance_limit is cached
```
jralls: Too clever by half. Change your design so that the limits are loaded from KVP at object initialization and save the value to KVP in xaccAccountCommitEdit.

gnucash/import-export/import-main-matcher.cpp
```diff
@@ -1043,22 +1043,19 @@ input_new_fields (GNCImportMainMatcher *info, RowInfo& rowinfo,
...
        first available entry and give it focus. */
-    for (const auto& entryinfo : entries)
-        if (entryinfo.can_edit)
-        {
-            gtk_widget_grab_focus (entryinfo.entry);
-            break;
-        }
+    if (auto it = std::find_if (entries.begin(), entries.end(),
+                                [](auto info){ return info.can_edit; });
+        it != entries.end())
```
jralls: That syntax is a little odd. Let's do 
```
    auto it = std::find_if (entries.begin(), entries.end(),
                                [](auto info){ return info.can_edit; });
    if (it != entries.end())
```
instead.

jralls (changes requested):

libgnucash/engine/AccountP.hpp
christopherlam: > Too clever by half. Change your design so that the limits are loaded from KVP at object initialization and save the value to KVP in xaccAccountCommitEdit.

Not too sure about that. `gnc_account_init` shouldn't assume a book is available to read kvp from. I'm thinking a 'balance_limits_are_cached' bool may be used. Or, forego caching them altogether.

jralls: What? You can't create any other kind of QofInstance without a book, not that that has anything to do with this.

You have the wrong mental model of KVP. KVP is a storage extension to provide class member variables that don't require a matching XML element name and therefore require adding handler code to the backend. Ideally KVP wouldn't exist and we'd maintain and version datafiles, but we don't. Next best would have the KVP code in the backends and map directly to instance member variables so the engine wouldn't have to worry about them.

jralls:
That last commit about centralizing GValue got me to realize that since you're writing C++ here and accessing the KvpFrames directly, there's no need for GValues at all, use `KvpFrame::set_path(std::vector<std::string> path, KvpValue* value)` and `KvpValue* KvpFrame::get_slot(std::vector<std::string> path)`. To get a particular type for the KvpValue* use `value->get<T>()` substituting the type you want for T, as in `value->get<time64>();`. To create a KvpValue you can just construct it from the source variable `KvpValue val(the_time);`.

Edit: Better yet move them to qofinstance. The whole point of the GValue dodge was to provide a generic interface similar to g_object_[sg]et for C code. It doesn't make sense to have a bunch of `set_kvp_foo_path` functions wrapping the abstraction. You could go full C++ and do
```cpp
template <typename T> T
qof_instance_get_path_value(QofInstance* inst, Path path)
{
     auto kvp_value{inst->get_slot(path)};
     return kvp_value->get<T>();
}

template <typename T> void
qof_instance_set_path_value(QofInstance* inst, Path path, T value) 
{
     KvpValue kvp_value(value);
     inst->kvp_data->set_path(path, kvp_value);
}
```
Then to handle the edit level and marking-dirty in Account.cpp you'd have
```cpp
template <typename T>
set_slot_from_path(Account* acc, Path path, T value)
 {
      xaccAccountBeginEdit(acc);
      qof_instance_set_path_value(QOF_INSTANCE(acc), path, value);
      mark_account(acc);
      xaccAccountCommitEdit(acc);
}
```
When calling `qof_instance_get_path_value` you'll specify the type that you want, e.g.
`qof_instance_get_path_value<gnc_numeric>(acc, path);`.

christopherlam:
> That last commit about centralizing GValue got me to realize that since you're writing C++ here and accessing the KvpFrames directly, there's no need for GValues at all, use `KvpFrame::set_path(std::vector<std::string> path, KvpValue* value)` and `KvpValue* KvpFrame::get_slot(std::vector<std::string> path)`. To get a particular type for the KvpValue* use `value->get<T>()` substituting the type you want for T, as in `value->get<time64>();`. To create a KvpValue you can just construct it from the source variable `KvpValue val(the_time);`.
> 
> Edit: Better yet move them to qofinstance. The whole point of the GValue dodge was to provide a generic interface similar to g_object_[sg]et for C code. It doesn't make sense to have a bunch of `set_kvp_foo_path` functions wrapping the abstraction. You could go full C++ and do
> 
> ```cpp
> template <typename T> T
> qof_instance_get_path_value(QofInstance* inst, Path path)
> {
>      auto kvp_value{inst->get_slot(path)};
>      return kvp_value->get<T>();
> }
> ```

Not quite because `kvp_value` may be a nullptr. Also the `get_kvp_boolean_path` is a polyglot.

Edit: alternatively we could return `std::optional<T>`

christopherlam:
The last issue is the `optional<optional<numeric>>` which I admit is confusing but I'm not sure where to initialise the balance limits except on the first getter call.

libgnucash/engine/Account.cpp
```diff
@@ -4856,211 +4678,93 @@ xaccAccountGetLastNum (const Account *acc)
...
```
jralls: Didn't notice this before: You're masking the paramerter, you ~should change the name of this `priv_limit`~ drop `auto`: You want to set the parameter.

libgnucash/engine/AccountP.hpp
jralls: > The last issue is the optional<optional> which I admit is confusing but I'm not sure where to initialise the balance limits except on the first getter call.

Init them in gnc_account_init. Slots go with the QofInstance, and that's part of struct Account. It has nothing to do with the book.

Or don't cache them at all. Retrieving a KVP isn't that expensive.

jralls:
If you're going to bury the template declaration in a cpp file, which speeds up compilation especially when rebuilding and avoids ODR problems if the template gets used with the same args in more than one compilation unit, then you have to force instantiation in the cpp file for all of the types you want to use it with. That's just some simple declarations, see [gnc-options.cpp](https://github.com/Gnucash/gnucash/blob/fef0124e45e0ec7ee053d858714acd2a0c317f83/libgnucash/engine/gnc-option.cpp#L495) for examples.

libgnucash/engine/Account.cpp
```diff
@@ -2477,107 +2477,147 @@ xaccAccountSetDescription (Account *acc, const char *str)
...
+
+template <typename T> std::optional<T>
+qof_instance_get_path_kvp (QofInstance* inst, StrVec path)
+{
+     auto kvp_value{inst->kvp_data->get_slot(path)};
+     return kvp_value ? std::make_optional<T>(kvp_value->get<T>()) : std::nullopt;
+}
+
+template <typename T> void
+qof_instance_set_path_kvp (QofInstance* inst, StrVec path, std::optional<T> value)
```
jralls: This will be confusing because the parameter order is different from `qof_instance_set_path_kvp(QofInstance*, GValue*, Path)`. Also it would be clearer to use the same `Path` alias for `std::vector<std::string>>` as the Kvp code.

libgnucash/engine/Account.cpp
```diff
@@ -2477,107 +2477,147 @@ xaccAccountSetDescription (Account *acc, const char *str)
...
+}
+
 static void
 set_kvp_string_path (Account *acc, std::vector<std::string> const & path,
                      const char *value)
 {
     g_return_if_fail(GNC_IS_ACCOUNT(acc));
 
-    xaccAccountBeginEdit(acc);
+    std::optional<char*> val;
```
jralls: The KvpValue type is [`const char*`](https://github.com/Gnucash/gnucash/blob/fef0124e45e0ec7ee053d858714acd2a0c317f83/libgnucash/engine/kvp-value.cpp#L86)

libgnucash/engine/Account.cpp
```diff
@@ -2477,107 +2477,147 @@ xaccAccountSetDescription (Account *acc, const char *str)
     xaccAccountCommitEdit(acc);
 }
 
+using StrVec = std::vector<std::string>;
+
+template <typename T> std::optional<T>
+qof_instance_get_path_kvp (QofInstance* inst, StrVec path)
+{
+     auto kvp_value{inst->kvp_data->get_slot(path)};
```
jralls: Both get and set need a null-check on `inst`. Doing so makes the `get_kvp_T_path` functions redundant.

jralls (changes requested):
> note these are overloaded functions returning type T instead of
getting/setting GValues

That's unclear. They're template overrides with std::optional<T> replacing the GValue containing T.

This is pretty close, just a few tweaks left.

libgnucash/engine/Account.cpp
christopherlam: I didnt const the char* because of https://discourse.gnome.org/t/copy-a-gchar-explication/3105

libgnucash/engine/qofinstance-p.h
```diff
@@ -166,13 +166,11 @@ void qof_instance_get_path_kvp (QofInstance *, GValue *, std::vector<std::string
 
 void qof_instance_set_path_kvp (QofInstance *, GValue const *, std::vector<std::string> const &);
 
-using StrVec = std::vector<std::string>;
-
 template <typename T> std::optional<T>
-qof_instance_get_path_kvp (QofInstance*, const StrVec&);
+qof_instance_get_path_kvp (QofInstance*, const std::vector<std::string>&);
```
jralls: `qof_instance_get_path_kvp (QofInstance*, const Path&)`

`using Path = std::vector<std::string>>;` is in kvp-frame.hpp, which is included in qofinstance-p.h.

jralls (changes requested):
I still want the QofInstance* NULL check in qof_instance_get_path_kvp and get_kvp_gnc_foo_path removed.

You made a typo in your last commit that causes tests to fail.

christopherlam:
if the last two are ok I'll tidy up and merge in.

libgnucash/engine/qofinstance.cpp
```diff
@@ -1063,6 +1063,30 @@ qof_instance_set_kvp (QofInstance * inst, GValue const * value, unsigned count,
     delete inst->kvp_data->set_path (path, kvp_value_from_gvalue (value));
 }
 
+template <typename T> std::optional<T>
+qof_instance_get_path_kvp (QofInstance* inst, const Path& path)
+{
```
jralls, line 1068: Add here:
```
      g_return_val_if_fail (QOF_INSTANCE (inst), std:::nullopt);
```

libgnucash/engine/qofinstance.cpp
```diff
@@ -1063,6 +1063,30 @@ qof_instance_set_kvp (QofInstance * inst, GValue const * value, unsigned count,
...
+    auto kvp_value{inst->kvp_data->get_slot(path)};
+    return kvp_value ? std::make_optional<T>(kvp_value->get<T>()) : std::nullopt;
+}
+
+template <typename T> void
+qof_instance_set_path_kvp (QofInstance* inst, std::optional<T> value, const Path& path)
+{
```
jralls, line 1076: Add here:
```
     g_return_val_if_fail (QOF_INSTANCE (inst), std::nullopt);
```

libgnucash/engine/Account.cpp
```diff
@@ -2481,73 +2477,122 @@ xaccAccountSetDescription (Account *acc, const char *str)
...
+get_kvp_int64_path (const Account *acc, const Path& path)
+{
+    return qof_instance_get_path_kvp<int64_t> (QOF_INSTANCE(acc), path);
 }
 
 void
 xaccAccountSetColor (Account *acc, const char *str)
 {
-    set_kvp_string_tag (acc, "color", str);
+    g_return_if_fail(GNC_IS_ACCOUNT(acc));
```
jralls: No need for this (or any of the other's you've added) once `qof_instance_set_path_kvp` and `qof_instance_get_path_kvp` do the checks. Mind the additional checks are harmless and do report the error a little closer to its source so arguably it makes for a better trace file.

jralls (changes requested):
Not quite, you went in the opposite direction I had in mind. The null/wrong type check needs to be at the lowest level to protect against some future use that doesn't do it. There's no harm in doing it also nearer the source to make a more usefult tracefile.

jralls (approved):

jralls:
OK, merge when ready.

PR #1971, opened 2024-06-25
Require gettext 0.22 for make pot

fellen:


CMakeLists.txt
```diff
@@ -262,15 +262,19 @@ find_library (REGEX_LIBRARY NAMES regex)
...
-    message (WARNING "Gettext version 0.20 or more recent is required to translate the 'developer_name' tag in gnucash.appdata.xml. All but that tag will be translated in the generated file.")
+if (${GETTEXT_VERSION_STRING} VERSION_LESS 0.22)
+    message (WARNING "Gettext version 0.22 or more recent is required to detect c++-formats.")
     if(BUILD_GNUCASH_POT)
         # Only emit warning if potfile generation was enabled otherwise
-        message (WARNING "Gettext version 0.20 or more recent is required to extract all translatable strings. Potfile generation will be disabled.")
+        message (WARNING "Gettext version 0.22 or more recent is required to detect c++-formats. Potfile generation will be disabled.")
     endif()
     set (BUILD_GNUCASH_POT OFF)
 endif()
```
fellen: In this if branch it would break the Ubuntu build.

jralls:
I'm not convinced this is necessary. There's no effect on string extraction or translation. The only difference between 0.21 and 0.22 is that the latter adds a comment `#, c++-format` when the msgstr uses `{#}` style variable reference. Weblate displays that under **Flags** about halfway down the string information sidebar. It seems unlikely that many translators will even notice and of those who do few will understand what it's trying to tell them.

jralls:
The gettext documentation [for the feature](https://www.gnu.org/software/gettext/manual/html_node/c_002b_002b_002dformat.html) says that it's about [C++20 std::format](https://en.cppreference.com/w/cpp/utility/format), which is derived from [{fmt}](https://github.com/fmtlib/fmt). While I very much want to replace all of our iostreams usage with that, it's not going into stable this round.

PR #1975, opened 2024-06-30
Expand widget to full width in GncOption

christopherlam:
The `GncAccountSel` should be expanded.

Before:
image

After:
image

### DIFF
--- a/gnucash/gnome-utils/gnc-option-gtk-ui.hpp
+++ b/gnucash/gnome-utils/gnc-option-gtk-ui.hpp
@@ -150,7 +150,7 @@ wrap_widget (const GncOption& option, GtkWidget* widget, GtkGrid* page_box, int 
 {
     auto enclosing{gtk_box_new (GTK_ORIENTATION_HORIZONTAL, 5)};
     gtk_box_set_homogeneous (GTK_BOX (enclosing), FALSE);
-    gtk_box_pack_start(GTK_BOX(enclosing), widget, FALSE, FALSE, 0);
+    gtk_box_pack_start(GTK_BOX(enclosing), widget, false, true, 0);
     set_name_label(option, page_box, row, false);
     set_tool_tip(option, enclosing);
     gtk_widget_show_all(enclosing);


jralls:
It surprises me that your proposed change makes any difference: The [documentation](https://docs.gtk.org/gtk3/method.Box.pack_start.html) says "This parameter has no effect if expand is set to FALSE." 

Anyway, I don't think this is the right place to make the change, nor do I think we want the disclose arrow and option button pushed all the way to the right. Something about the account selection listbox is messed up: It should be aligned left, not centered, and should be wider without expanding all the way to the right edge of the dialog box. As a related issue why are the GtkTextEntries expanding when nothing else is? @Bob-IT any ideas?

Bob-IT:
I will have a look.

Bob-IT:
The reason for the GtkTextEntries being so long is that the GtkGrid halign is set to `GTK_ALIGN_FILL` and so is the size of the window. As the text entries do not have set widths, they also expand to this width.
This could be fixed by adding a line to set it to `GTK_ALIGN_START` in dialog-options.cpp, line 220,
`    gtk_widget_set_halign (GTK_WIDGET(options_box), GTK_ALIGN_START);`

 and so will be the size of the biggest containing widget which would be the 'Start Date'.

To fix the account selection width, you need to set the child properties of the GtkBox it is contained in so to make it specific in gnc-option-gtk-ui.cpp, line 1108 add...

```
    gtk_container_child_set (GTK_CONTAINER(gtk_widget_get_parent(widget)),
                             widget, "fill", true, "expand", true,
                             nullptr);
```
As seen below...
options

If you want more text for the Account field, one could add a function like gnc_account_select_set_width_chars to gnc-account-sel.c to set the entry width. This would also mean the text entries will appear the same length.

Bob-IT:
It turns out I have already added the last bit as a `property`, just could not see it.

christopherlam:
:-O Nice one... However instead of the GncAccountSel expanding to the widest widget, I think it'd be nice if the GncAccountSel could default to be able to show the longest gnc_account_full_name width, but flexibly narrowing to the window size. Is this feasible?

christopherlam:
> :-O Nice one... However instead of the GncAccountSel expanding to the widest widget, I think it'd be nice if the GncAccountSel could default to be able to show the longest gnc_account_full_name width, but flexibly narrowing to the window size. Is this feasible?

I'll elaborate further: the account tree may be arbitrarily deep, e.g. Expenses:Taxes:Federal:Income:2019; if the whole account fullname is visible (up to a limit) then it'll be a nice visual indicator of the selected account.

Bob-IT:
I am a bit undecided at the moment, if one changes the grid to align start then the max grid width is the largest contained widget. On the balance sheet there is a tree view, so the max width size would be limited to the width of the four buttons at the bottom so if you have really long account names you will not see the ends as there is no horizontal scrolling.
It maybe possible to add, I have not looked.

The other possible solution is to leave the grid as is and set character width limits to the text entries and the account select. In the following image the entries are both set to 50 characters.

I am assuming your motivation is to see the account name at the end of the full name to this end one could move the cursor to the text end on change so for 'Assets:Current Assets2:Unknown OFX current account "Bank account 678901"' you get...

accoumt-sel

christopherlam:
@Bob-IT I'll let you be the judge for aesthetics... A wider GncAccountSel is miles better than current code, and I'm not too worried about displaying whole fullname. IMHO showing the tail end of fullname by default is jarring.

jralls:
Maybe we could ellipsize it? Though somebody's sure to complain that the ellipsis zaps the one account that they use to differentiate two 12-deep account trees. :-/

jralls:
There being no further discussion, I've committed @Bob-IT's [changes](https://github.com/Gnucash/gnucash/pull/1975#issuecomment-2200124041). @Bob-IT if you're motivated to do more, have at it.

PR #1979, opened 2024-07-04
Overload account foreach descendant

christopherlam:
base of #1965 -- introduces overload for `gnc_account_foreach_descendant`

### DIFF
--- a/libgnucash/engine/Account.cpp
+++ b/libgnucash/engine/Account.cpp
@@ -2752,23 +2752,35 @@ DxaccAccountSetCurrency (Account * acc, gnc_commodity * currency)
 /********************************************************************\
 \********************************************************************/
 
-static void
-account_foreach_descendant (const Account *acc, AccountCb thunk,
-                            void* user_data, bool sort)
+void
+gnc_account_foreach_descendant (const Account *acc, std::function<void(Account*)> account_cb)
 {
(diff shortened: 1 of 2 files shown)


jralls:
See comments on the corresponding commits in #1965.

libgnucash/engine/Account.cpp
```diff
@@ -2752,23 +2752,34 @@ DxaccAccountSetCurrency (Account * acc, gnc_commodity * currency)
 /********************************************************************\
 \********************************************************************/
 
+void
+gnc_account_foreach_descendant (const Account *acc, std::function<void(Account*)> account_cb)
+{
+    g_return_if_fail (GNC_IS_ACCOUNT(acc));
+
+    // some callers eg xaccAccountTreeScrubLots will modify the children
+    auto children = GET_PRIVATE(acc)->children;
```
christopherlam, line 2763: Sadly if a reference is created instead of a vector copy, test-lots will fail.

jralls: That shouldn't surprise you.

jralls:
The code looks generally OK. Do you intend to turn the old `gnc_account_foreach_descendant` into a C wrapper for the new one?

christopherlam:
> The code looks generally OK. Do you intend to turn the old `gnc_account_foreach_descendant` into a C wrapper for the new one?

yes already present in this branch

PR #1989, opened 2024-07-20
Optiondb set changed cb

christopherlam:
Adds a generic `gnc-option-set-changed-callback` instead of adding a specialised `gnc-register-complex-account-sel-limited-option`.

Unfortunately `scm_from_value` isn't defined in `gnome-utils/gnc-option-gtk-ui.cpp`.


jralls:
> Unfortunately scm_from_value isn't defined in gnome-utils/gnc-option-gtk-ui.cpp.

With good reason. SCM belongs bindings and nowhere else.

As for this PR: No. Absolutely not. Ever.

The whole point of the options rewrite was to get rid of garbage like this. Scheme may call C. C/C++ may not call Scheme.

PR #1990, opened 2024-07-20
Budget Report Features: Rollover (Envelope) Budgeting and Unselected Account Exclusion

rmehyde:
Hi Gnucash team! I'm a longtime user and have opened a PR with two features I added to my custom Budget Report, in case they're useful to others in the community. The features are:

1. **Budget Rollover:** the option to roll over budget surplus or deficit to the following period
2. **Unselected Account Exclusion:** the option to exclude balances of deselected accounts from their ancestor totals.
    - e.g. if your "Groceries" account is not selected to display in the budget report the value of transactions in that account will not be included in "Expenses"

I implemented these because they're valuable to me, and I thought others in the community might find them useful as well. If maintainers agree that either or both make sense to have in the main program, I'll add unit tests and implement any suggested changes. If they're best left to a Custom Report, I'll keep using them myself that way!

Thanks,
Reese

### Budget Rollover

The implementation of budget rollover is similar to the Accumulated Amounts feature: when the option is selected, previous periods are added up along with the current period. The difference as that with Budget Rollver, the Budget totals from previous periods are added to the current period Budget but Actual totals from previous periods are subtracted from the current period budget, rather than being added to the current period Actuals. Only one of the Use Accumulated Amounts and Rollover Budget Difference options can be selected.

I found quite a few mailing list messages requesting this feature, so I believe this one in particular may be valuable. A sampling:
- [How can I do something like the "Envelope System" (2009)](https://lists.gnucash.org/pipermail/gnucash-user/2009-January/027884.html)
- [Envelope budgeting with credit cards (2010)](https://lists.gnucash.org/pipermail/gnucash-user/2010-February/033469.html)
- [How to Budget Investment Accounts? (2014)](https://lists.gnucash.org/pipermail/gnucash-user/2014-August/055807.html)
- [Budgets - how does the community want to use them (2015)](https://lists.gnucash.org/pipermail/gnucash-user/2015-December/063202.html)

### Unselected Account Exclusion

In general, Unselected Acount Exclusion works by subtracting the balance of unselected accounts from selected parent/ancestor accounts. The nuance is that sometimes balances need to be added as well. For example, if a grandparent account is selected, parent account is unselected, and child account is selected, the parent account balance will be subtracted from the grandparent balance and child account added to the grandparent balance. This is explained in greater detail in [this code comment](https://github.com/rmehyde/gnucash/blob/5a820356155afc53bc55829ddf7d51183116d039/gnucash/report/reports/standard/budget.scm#L427-L464).

An ideal solution would probably just be a C++ function to get an account balance with these offsets. But unfortunately my C(++) skills are almost nonexistent, so learning Scheme and implementing it this way worked best for me.

I didn't find any mailing list messages for this one so there may not be much other interest but I was also less sure what to search for here.

### DIFF
--- a/gnucash/report/reports/standard/budget.scm
+++ b/gnucash/report/reports/standard/budget.scm
@@ -40,6 +40,17 @@
 
 (define reportname (N_ "Budget Report"))
 
+
+;; TEMP: debug logging
+(define (debug-log message)
+  (let* ((out (open-file
+                (string-append "/home/reese/.local/share/gnucash/report-logs/custom-budget-report-" current-time-string ".log")
+                "a")))
+    (display message out)
+    (newline out)
+    (close-output-port out)))
+
+
 ;; define all option's names so that they are properly defined
 ;; in *one* place.
 
@@ -60,6 +71,8 @@
 (define opthelp-show-difference (N_ "Display the difference as budget - actual."))
 (define optname-accumulate (N_ "Use accumulated amounts"))
 (define opthelp-accumulate (N_ "Values are accumulated across periods."))
+(define optname-rollover (N_ "Roll over difference"))
+(define opthelp-rollover (N_ "Budget period surplus or deficit is rolled over to next period."))
 (define optname-show-totalcol (N_ "Show Column with Totals"))
 (define opthelp-show-totalcol (N_ "Display a column with the row totals."))
 (define optname-show-zb-accounts (N_ "Include accounts with zero total balances and budget values"))
@@ -123,9 +136,17 @@
       "a" (N_ "Budget to use.")
       (gnc-budget-get-default (gnc-get-current-book)))
 
-    (gnc-register-simple-boolean-option options
+    (gnc-register-complex-boolean-option options
       gnc:pagename-general optname-accumulate
-      "b" opthelp-accumulate #f)
+      "b1" opthelp-accumulate #f
+      (lambda (new-val)
+        (set-option-enabled options gnc:pagename-general optname-rollover (eqv? new-val #f))))
+
+    (gnc-register-complex-boolean-option options
+      gnc:pagename-general optname-rollover
+      "b2" opthelp-rollover #f
+      (lambda (new-val)
+        (set-option-enabled options gnc:pagename-general optname-accumulate (eqv? new-val #f))))
 
     (gnc-register-complex-boolean-option options
       gnc:pagename-general optname-use-budget-period-range
@@ -277,6 +298,7 @@
          (show-note? (get-val params 'show-note))
          (footnotes (get-val params 'footnotes))
          (accumulate? (get-val params 'use-envelope))
+         (rollover? (get-val params 'rollover))
          (show-totalcol? (get-val params 'show-totalcol))
          (use-ranges? (get-val params 'use-ranges))
          (num-rows (gnc:html-acct-table-num-rows acct-table))
@@ -340,7 +362,7 @@
              (reverse-balance? (gnc-reverse-balance acct))
              (maybe-negate (lambda (amt) (if reverse-balance? (- amt) amt)))
              (allperiods (filter number? (gnc:list-flatten column-list)))
-             (total-periods (if (and accumulate? (not (null? allperiods)))
+             (total-periods (if (and (or accumulate? rollover?) (not (null? allperiods)))
                                 (iota (1+ (apply max allperiods)))
                                 allperiods))
              (income-acct? (eqv? (xaccAccountGetType acct) ACCT-TYPE-INCOME)))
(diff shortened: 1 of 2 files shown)


rmehyde:
@christopherlam it looks like you've worked a bit on the budget report, would you be a good person to weigh in here?

christopherlam:
I haven't verified the actual code, but the existing "Use accumulated amounts" option aims to accumulate budget + actual values across periods, which means any budget surplus or deficit is accounted for in the subsequent periods. Would your change be different from the existing option?

The other concern about excluding accounts is a nice idea, however, it will be a fundamental difference in the derivation of parent account totals. How do other budget reports handle subaccounts? I haven't checked them all. How does the budget editor and the budget editor 'Estimate' tool handle subaccounts?

rmehyde:
@christopherlam I apologize for the very long delay, my GitHub notifications don't seem to be working as expected but I'll look to fix that. Thank you for the great questions, and sorry for the long reply.

### Rollover (Envelope) Budget vs Accumulated Amounts

This is a great question about the difference between Accumulated Amounts and Budget Rollover. It's in the Budget and Actual values for each period: Accumulated Amounts shows a total across all budget periods, while Budget Rollover shows only each period's value, with adjustments to the Budget value for the previous periods' differences.

For example, if you budget $1000 per month for Groceries, with Accumulated Amounts in the 6th budget period your Actual value would be your total spend across the six periods, and your Budget value will always be $6000. With Rollover, your Actual value would be only the amount during the 6th period, and your Budget value in the 6th period will be $1000 plus or minus the total amount you overshot or undershot the budget during the previous five periods.

More concretely, suppose we have the following Budget and Actual for six periods:

|            |  Jan  |  Feb  |  Mar  |  Apr  |  May  |  Jun  |
| ---------- | ----- | ----- | ----- | ----- | ----- | ----- |
| Budget     | $1000 | $1000 | $1000 | $1000 | $1000 | $1000 |
| Actual     |  $980 | $1200 | $1000 |  $900 | $1050 |  $890 |
| Difference |  +$20 | -$200 |    $0 | +$100 |  -$50 | +$110 |

With Accumulated Amounts enabled, we would get the following:

|            |  Jan  |  Feb  |  Mar  |  Apr  |  May  |  Jun  |
| ---------- | ----- | ----- | ----- | ----- | ----- | ----- |
| Budget     | $1000 | $2000 | $3000 | $4000 | $5000 | $6000 |
| Actual     |  $980 | $2180 | $3180 | $4080 | $5130 | $6020 |
| Difference |  +$20 | -$180 | -$180 |  -$80 | -$130 |  -$20 |

With Rollover Budget enabled, the "Difference" column will be the same as with Accumulated Amounts, but we'll see the Budget and Actual values for each period instead of running totals. The $1000 Budget value for each period will be adjusted to include the difference from the previous period:

|            |  Jan  |  Feb  |  Mar  |  Apr  |  May  |  Jun  |
| ---------- | ----- | ----- | ----- | ----- | ----- | ----- |
| Budget     | $1000 | $1020 |  $820 |  $820 |  $920 |  $870 |
| Actual     |  $980 | $1200 | $1000 |  $900 | $1050 |  $890 |
| Difference |  +$20 | -$180 | -$180 |  -$80 | -$130 |  -$20 |

### Excluding Subaccounts

Great questions, thank you! I had looked a bit at the budget editor but not considered the Estimate tool or other budget reports. Responding inline:

> How does the budget editor... handle subaccounts?

If a parent account has its own value specified directly in the budget, that value is used as the budget value. If it has no value specified, but its children do have values specified, the sum of its children is used for the budget value. In this case where the parent account is just the total of children, the text is light grey instead of black to indicate as much.

> How does... the budget editor 'Estimate' tool handle subaccounts?

This appears to use the total including all subaccounts for estimation. This makes sense since there is no account selection in the budget editor.

> How do other budget reports handle subaccounts?

This was a very interesting question, the answer is a bit all over the place. My main discovery is the "Parent account balances" option which offers similar functionality but only works as expected in one of the three available reports. I think in the long term, it would make sense to consolidate all possibilities into this option and fix it across all reports.

**Budget Balance Sheet:** Already includes this feature. Either by setting "Parent account balances" to "Calculate Subtotal" (sum of only selected children), or "Parent account subtotals" to "Show subtotals" (shows at bottom on new row instead of as parent account balance). However I found the "Account balance" option of "Parent account balances" did not work: it displayed 0 even when the parent account had a balance.
**Budget Chart:** Not applicable as it only shows one level at a time. Suppose you choose "Level of Subbacounts" = 3. Then if you select Account A at Level 2 but no children accounts, the total of Account A will be shown on a chart. However if you select any children account(s) under Account A (at Level 3) then only the individual children accounts will be shown but there will be no chart for Account A.
**Budget Flow:** Never calculates subtotals, only uses direct account values.
**Budget Income Statement:** Bug. Always computes subtotal for selected and unselected children accounts. Except when "Report for a range of budget periods" is selected, then parent account value is always 0. However, it has the same "Parent account balances" and "Parent account subtotals" options as Budget Balance Sheet, the options just don't function correctly.
**Budget Report Profit & Loss:** Bug. Similarly has "Parent account balances" and "Parent account subtotals" options. However "Account balance" actually computes children subtotal, and "Calculate Subtotal" results in sum of (all children) + (selected children), e.g. 2X sum of children balances if all children are selected

jralls (commented):
It all looks reasonable to me, but surely these changes warrant corresponding changes in test-budget-report.scm.

rmehyde:
> surely these changes warrant corresponding changes in test-budget-report.scm.

Yep absolutely! Just didn't want to write them upfront if these were best left to my custom reports:
> If maintainers agree that either or both make sense to have in the main program, I'll add unit tests and implement any suggested changes.

I'll add tests and then mark this PR as ready.

rmehyde:
Apologies for the very long delay but I've updated tests here! That helped catch a bug in budget note rendering in both Accumulated Amounts and Rollover.

This is now ready for review if you have a chance @jralls :pray:

jralls:
> That helped catch a bug in budget note rendering in both Accumulated Amounts and Rollover.

That's why you write the tests first, then the code. ;-)

rmehyde:
> That's why you write the tests first, then the code. ;-)

Maybe next time I'll be comfortable enough in Scheme for TDD :)

Just let me know if there's anything else you need from me for review @jralls

rmehyde:
Friendly bump here @christopherlam and @jralls, are y'all good to merge this or any concerns or blockers? Thanks!

PR #1995, opened 2024-07-26
Optiondb changes

christopherlam:
because the account_list options were getting some spurious null accounts.


jralls:
Looks good.

christopherlam:
let's revert this for 799533

jralls:
@christopherlam do you remember what motivated it in the first place? I thought that it was some discussion of a problem in gnucash-user but I couldn't find it.

christopherlam:
3e5d6a5 was written to suppress some warnings, I recall there was a prior recent commit of yours (and I cannot find anymore) which was rather chatty especially when the account from a guid didn't exist anymore. However I cannot recall the exact action which was dumping warnings.

PR #1997, opened 2024-07-27
Support "Show long account names" option in Advanced Portfolio report

Blacksmoke16:
My portfolio consists of roughly the same securities in multiple different accounts. As such, this results in there being duplicate account names in the advanced portfolio report which makes it pretty confusing to know which account each line actually represents. This PR adds an opt-in feature to let it use the long account names like you can in the investment lots report.

This is my first contribution, so happy to make any required changes! Will open as a draft to start.


Blacksmoke16:
Bump, any feedback on this?

PR #1999, opened 2024-08-01
Python example: new year balance

quazgar:
# What's this? #

Following [this bug about an outdated Python example](https://bugs.gnucash.org/show_bug.cgi?id=799327), I started working on `new_book_with_opening_balances.py` and created an updated script with the same purpose: to create a new file with an identical account structure and balances, but without all the transactions of the year past.

As a result, the new script does:

- Create a new GnuCash file.
- Duplicate the origin's accounts.
- Create one or more opening transactions to initialize the accounts with the correct value.
  - Multiple transactions and custom target accounts are new.
- Duplicate business entities such as vendors, customers, employees.
  - This is also new.

Also there are a few new Python utility functions to make the development a bit easier.

# Notes #

- The new script was developed on [Gitlab.com](https://gitlab.com/wiese28/gnucash-happy-new-year), that is also where future development will probably happen.  I added it to the GnuCash repo with `git subtree`, so for users the different source repository should be completely transparent.
- There are tests which can be run with pytest.  I do not know enough about Github's workflows to trigger these tests.  But if someone wants to have a look at the pipeline scripts for GitLab, they can be found in `.gitlab-ci.yml`: it's really just `pytest tests`, once the environment is set up.
- The license is GPLv3, so it should be compatible with the rest of GnuCash, which is GPLv2-or-later.


jralls:
Wow. That's a lot more than you indicated on the bug report. Thanks for offering it but I don't think it belongs as part of GnuCash itself. I've added a link in the Wiki's [Python Bindings](https://wiki.gnucash.org/wiki/Python_Bindings#Example_Usages) page instead.

fellen:
@jralls To fix the [bug](https://bugs.gnucash.org/show_bug.cgi?id=799327) we should IMHO replace the [broken example](https://github.com/Gnucash/gnucash/blob/stable/bindings/python/example_scripts/new_book_with_opening_balances.py) by this PR or at least a link to it.

jralls:
@fellen It doesn't fix the bug, so no. @quazgar, who also filed the bug, went off in a different direction. The issue with the bug seems to be that calling `Session.save()` in Python throws an exception in the SQL backend. That's completely orthogonal to writing a new (and apparently SQL-only) new-file-with-balances application.

PR #2009, opened 2024-08-26
Gnc commodity cpp

christopherlam:
tested via add/remove custom commodity in ui

### DIFF
--- a/libgnucash/engine/gnc-commodity.cpp
+++ b/libgnucash/engine/gnc-commodity.cpp
@@ -109,6 +109,7 @@ static void commodity_free(gnc_commodity * cm);
 static void gnc_commodity_set_default_symbol(gnc_commodity *, const char *);
 
 using StringCommodityMap = std::unordered_map<std::string, gnc_commodity*>;
+using StringCommNSMap = std::unordered_map<std::string, gnc_commodity_namespace*>;
 
 struct gnc_commodity_namespace_s
 {
@@ -126,8 +127,7 @@ struct _GncCommodityNamespaceClass
 
 struct gnc_commodity_table_s
 {
-    GHashTable * ns_table;
-    GList      * ns_list;
+    StringCommNSMap ns_table;
 };
 
 static const std::unordered_map<std::string,std::string> gnc_new_iso_codes =
@@ -1606,8 +1606,7 @@ gnc_commodity_table *
 gnc_commodity_table_new(void)
 {
     gnc_commodity_table * retval = g_new0(gnc_commodity_table, 1);
-    retval->ns_table = g_hash_table_new(&g_str_hash, &g_str_equal);
-    retval->ns_list = nullptr;
+    new (&retval->ns_table) StringCommNSMap ();
     return retval;
 }
 
@@ -1649,12 +1648,12 @@ gnc_commodity_obtain_twin (const gnc_commodity *from, QofBook *book)
  ********************************************************************/
 
 static void
-count_coms(gpointer key, gpointer value, gpointer user_data)
+count_coms (const std::string& key, gpointer value, gpointer user_data)
 {
     auto& tbl = ((gnc_commodity_namespace*)value)->cm_table;
     guint *count = (guint*)user_data;
 
-    if (g_strcmp0((char*)key, GNC_COMMODITY_NS_CURRENCY) == 0)
+    if (key == GNC_COMMODITY_NS_CURRENCY)
     {
         /* don't count default commodities */
         return;
@@ -1670,9 +1669,9 @@ gnc_commodity_table_get_size(const gnc_commodity_table* tbl)
 {
     guint count = 0;
     g_return_val_if_fail(tbl, 0);
-    g_return_val_if_fail(tbl->ns_table, 0);
 
-    g_hash_table_foreach(tbl->ns_table, count_coms, (gpointer)&count);
+    std::for_each (tbl->ns_table.begin(), tbl->ns_table.end(),
+                   [&count](auto& str_ns){ count_coms (str_ns.first, str_ns.second, &count); });
 
     return count;
 }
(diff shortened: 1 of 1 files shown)


jralls:
Code mostly looks OK except for the last commit, see below. 

The first commit is dependent on the following two so it will cause crashes in bisects. The simplest fix would be to reorder it to after them but it would be more correct to combine the `g_list_free` changes with the respective conversion commit.
The two "use stl" commit messages are incorrect as they replace one STL algo (`for_each`) with a different one (`all_of`). Please fix them to say that.

The last commit is wrong. `ns->name` (not `ns->mnemonic` as claimed in the commit message) *is* `CACHE_INSERT`ed and does need to be `CACHE_REMOVE`d.
The ASAN fail (below) is a use-after-free:
```
==11961==ERROR: AddressSanitizer: heap-use-after-free on address 0x6020000528b0 at pc 0x7fb7844925b3 bp 0x7fffa61efb80 sp 0x7fffa61ef328
READ of size 1 at 0x6020000528b0 thread T0
    #0 0x7fb7844925b2 in __interceptor_strcmp ../../../../src/libsanitizer/sanitizer_common/sanitizer_common_interceptors.inc:466
    #1 0x7fb778da68a8 in gnc_commodity_table_map_namespace /home/runner/work/gnucash/gnucash/libgnucash/engine/gnc-commodity.cpp:1588
    #2 0x7fb778dabc76 in gnc_commodity_table_find_namespace /home/runner/work/gnucash/gnucash/libgnucash/engine/gnc-commodity.cpp:2123
    #3 0x7fb778dac1bc in gnc_commodity_table_delete_namespace /home/runner/work/gnucash/gnucash/libgnucash/engine/gnc-commodity.cpp:2152
    #4 0x7fb778dacfab in gnc_commodity_table_destroy /home/runner/work/gnucash/gnucash/libgnucash/engine/gnc-commodity.cpp:2202

0x6020000528b0 is located 0 bytes inside of 5-byte region [0x6020000528b0,0x6020000528b5)
freed by thread T0 here:
    #0 0x7fb7844b4537 in __interceptor_free ../../../../src/libsanitizer/asan/asan_malloc_linux.cpp:127
    #1 0x7fb77b4ed3b8  (/lib/x86_64-linux-gnu/libglib-2.0.so.0+0x383b8)
    #2 0x7fb77b4f5956 in g_hash_table_remove (/lib/x86_64-linux-gnu/libglib-2.0.so.0+0x40956)
    #3 0x7fb77932baa0 in qof_string_cache_remove /home/runner/work/gnucash/gnucash/libgnucash/engine/qof-string-cache.cpp:94
    #4 0x7fb778dac7e9 in gnc_commodity_table_delete_namespace /home/runner/work/gnucash/gnucash/libgnucash/engine/gnc-commodity.cpp:2163
    #5 0x7fb778dacfab in gnc_commodity_table_destroy /home/runner/work/gnucash/gnucash/libgnucash/engine/gnc-commodity.cpp:2202
```
Line 2202 is
```
for (auto ns : t->ns_vec)
        gnc_commodity_table_delete_namespace(t, ns->name);
```
The problem is that `for (auto ns: t->ns_vec)` is syntactic sugar for `for(auto it = ns_vec.begin(); it != ns_vec.end(); ++it) auto ns = *it;`. `gnc_commodity_table_delete_namespace` calls `table->ns_vec.erase (std::remove (table->ns_vec.begin(), table->ns_vec.end(), ns));`, invalidating `it` and `ns_vec.end()`. That causes two problems: `ns_vec` doesn't get completely cleared out and `for (auto ns : t->ns_vec)` will run off the end of the shrunken `ns_vec` causing use-after-free errors. As a band-aid you could instead just reverse that outer loop:
```
for (auto ns = t->ns_vec.rbegin(); ns != ns_vec.rend(); ++ns)
  gnc_commodity_table_delete_namespace(t, ns->name);
```
But consider the stupidity of the algorithm: We have an iterator in the vector. We get the string name from that iterator and search the vector again with the string to find the iterator that we already have, then convert the iterator to its object and search the vector yet again to remove the iterator. Then we destroy the object's members (ignoring its ref count and not NULLing any of them) and finally call the object's destructor (`g_object_unref(ns)`) which decrements the ref count and frees the memory assuming nothing has reffed it anywhere.

If the object cleanup was in `gnc_commodity_namespace_dispose_real` where it belongs then `gnc_commodity_table_delete_namespace` could just find the object from the name and `g_object_unref` it, and the table destructor could do
```
   while(!ns_vec.empty())
       g_object_unref(ns_vec.back());
```
Unfortunately `gnc_commodity_namespace` is widely used in C code so a C++ class would need C wrapper functions, but it doesn't use any of the features of QofInstance; it perhaps should use GObject ref counting, but considering how rare it is to do so in any GnuCash code it probably doesn't.

christopherlam:
> The first commit is dependent on the following two so it will cause crashes in bisects

The first commit is one that makes the get_commodities and get_namespaces return GLists that _must_ be freed by the caller, so I was actually thinking best to merge it separately first, so that this branch is purely c++ refactoring.

jralls:
> The first commit is one that makes the get_commodities and get_namespaces return GLists that must be freed by the caller

Ah, missed the `g_list_copy`s. OK.

christopherlam:
I see the silly object cleanup code paths... and hopefully with the lesser LOC here it would be easier to simplify.

libgnucash/engine/gnc-commodity.cpp
```diff
@@ -1830,11 +1819,9 @@ gnc_commodity_table_insert(gnc_commodity_table * table,
     nsp = gnc_commodity_table_add_namespace(table, ns_name, book);
 
     PINFO ("insert %p %s into nsp=%p %s", priv->mnemonic, priv->mnemonic,
-           nsp->cm_table, nsp->name);
-    g_hash_table_insert(nsp->cm_table,
-                        (gpointer)CACHE_INSERT(priv->mnemonic),
-                        (gpointer)comm);
-    nsp->cm_list = g_list_append(nsp->cm_list, comm);
+           &nsp->cm_table, nsp->name);
+    nsp->cm_table[priv->mnemonic] = comm;
```
christopherlam, line 1830: this removal of CACHE_INSERT should cause refcount error but in practice it doesn't...

jralls: No, `cache_remove()` just [returns if the key isn't in the cache](https://github.com/Gnucash/gnucash/blob/a99491d03425a3565bf2a2de6d3dbb8574a04815/libgnucash/engine/qof-string-cache.cpp#L89).

The string cache use here is pretty questionable for the GHashTable and utterly useless for the std::unordered_map: The main advantage to using the string cache is to be able to directly compare the pointers, no string comparison or string hash required. . The GHashTables were declared with g_string_hash so g_hashtable was looking through the cache pointer and hashing the string, discarding the string cache's benefit. The std::unordered_map goes a step further and creates a std::string, completely isolating it from the string cache's value.

jralls:
@christopherlam did you click the wrong button by mistake?

christopherlam:
I wasn't sure if there's still interest in glist->stl conversion...

PR #2010, opened 2024-09-01
Csv import: GncDateFormat ->date_locale

christopherlam:
offshoot #1710 -- here's attempt.

all GncDateFormat changed to use locales. The locale combobox is currently 799 items long!

but it doesn't seem to parse all dates. see example with debugging code. looks like ICU type issue which cannot understand "Sep" yet can understand "Apr".

image

### DIFF
--- a/libgnucash/engine/gnc-datetime.cpp
+++ b/libgnucash/engine/gnc-datetime.cpp
@@ -552,10 +552,11 @@ locale_to_formatter_and_calendar (const std::string locale_str)
     if (!tuple)
     {
         auto locale = icu::Locale::createCanonical (locale_str.c_str());
-        std::shared_ptr<icu::DateFormat> formatter(icu::DateFormat::createDateInstance(icu::DateFormat::kDefault, locale));
+        std::shared_ptr<icu::DateFormat> formatter(icu::DateFormat::createDateInstance(icu::DateFormat::kShort, locale));
         if (formatter == nullptr)
             throw std::invalid_argument ("Cannot parse string");
 
+        formatter->setLenient (false);
         UErrorCode status = U_ZERO_ERROR;
         std::shared_ptr<icu::Calendar> calendar(icu::Calendar::createInstance(locale, status));
         if (U_FAILURE(status))
@@ -573,7 +574,7 @@ GncDateImpl::GncDateImpl(const std::string str, const std::string locale_str) :
     /* Temporarily initialized to today, will be used and adjusted in the code below */
     m_greg(boost::gregorian::day_clock::local_day())
 {
-    std::cout << locale_str << '|' << str << ": ";
+    // std::cout << locale_str << '|' << str << ": ";
 
     auto [formatter, calendar] = locale_to_formatter_and_calendar (locale_str);
     icu::UnicodeString input = icu::UnicodeString::fromUTF8(str);
@@ -582,7 +583,7 @@ GncDateImpl::GncDateImpl(const std::string str, const std::string locale_str) :
     UDate date = formatter->parse(input, parsePos);
     if (parsePos.getErrorIndex() != -1)
     {
-        std::cout << "cannot parse " << std::endl;
+        // std::cout << "cannot parse " << std::endl;
         throw std::invalid_argument ("Cannot parse string");
     }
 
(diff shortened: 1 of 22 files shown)


jralls:
It didn't like Jun either. Does changing the locale change anything?

christopherlam:
> It didn't like Jun either. Does changing the locale change anything?

not really! "Jun" "Jul" are also off. "September" is accepted.

christopherlam:
oops

with en_US.utf8 "Sep":fail "Jun":fail "Jul":fail -- I guess it expects m/d/y
with en_AU.utf8 "Sep":fail "Jun":fail "Jul":fail
with en_GB.utf8 "Sep":fail "Jun":pass "Jul":pass
with en_CA.utf8 similar to en_GB

jralls:
It's partly explained here: [en_US](https://icu4c-demos.unicode.org/icu-bin/locexp/en/_/help.html#Month): Jun, Jul, Sep; [en_AU](https://icu4c-demos.unicode.org/icu-bin/locexp/en/_/help.html#Month): June, July, Sept; [en_GB](https://icu4c-demos.unicode.org/icu-bin/locexp/en/_/help.html#Month): Jun, Jul, Sept; [en_CA](https://icu4c-demos.unicode.org/icu-bin/locexp/en/_/help.html#Month): Jun, Jul, Sep.

That says it should work for US and CA. A decent parser should be able to deal with dd MMM YYYY regardless of the locale's normal output format.

libgnucash/engine/gnc-datetime.cpp
```diff
@@ -541,54 +541,62 @@ GncDateTimeImpl::timestamp()
...
+        std::shared_ptr<icu::DateFormat> formatter(icu::DateFormat::createDateInstance(icu::DateFormat::kDefault, locale));
+        if (formatter == nullptr)
+            throw std::invalid_argument ("Cannot parse string");
+
+        UErrorCode status = U_ZERO_ERROR;
+        std::shared_ptr<icu::Calendar> calendar(icu::Calendar::createInstance(locale, status));
+        if (U_FAILURE(status))
+            throw std::invalid_argument ("Cannot parse string");
+
+        tuple = std::make_tuple<DateFormatPtr, CalendarPtr>(std::move(formatter), std::move(calendar));
```
jralls: You forgot to add the tuple to the cache.

christopherlam: I didn't: see the magic & in `auto&` ☺️
I did test via printfs

christopherlam: The std::move is slightly beyond my current understanding but it's a requirement for compilation to succeed. Chatgpt did suggest it.

flywire:
> It's partly explained here: [en_AU](https://icu4c-demos.unicode.org/icu-bin/locexp/en/_/help.html#Month): June, July, Sept

Seems it is based on poor information. I've never seen that month format in Australia, normally numeric or 3 character alpha.

libgnucash/engine/gnc-datetime.cpp
jralls: > see the magic & in auto&

Very clever. I did miss that indeed.

> The std::move is slightly beyond my current understanding 

```
[std::tuple](http://en.cppreference.com/w/cpp/utility/tuple)<unwrap_decay_t<Types>...> make_tuple(Types&&... args)
```
In template context `Types&&` means [forwarding reference](https://en.cppreference.com/w/cpp/language/reference#Forwarding_references). It needs either a reference or an lvalue. A named variable is an rvalue so it needs to be cast into an lvalue. `std::move` does that. `std::move` is apparently confusing. Maybe the committee should have called it `lvalue_cast` instead.

christopherlam:
Ok let's kill this stinky hack in favour of #2011

gjanssens:
Ok.

For posterity, there was a bug in the last attempt, where you try two parsers. If the short parser failed, the medium parser would always fail as well because the `parsePos` was not reset. Due to this the medium parser always had only a partial string left to parse. By resetting `parsePos` before attempting the medium parser, I got better results.

Still all these parse attempts are costly. In the ideal case - where only a few parsers are valid - this is ok. However if I test on a string that's valid in many parsers the time to run all the tests add up quickly.

For example with 500 date strings like '06/09/2021', over 600 parsers return a valid date, and the time it takes varies from 2,5 seconds to 3,5 seconds on my system. This can likely be optimized some more by eliminating failing short parsers from the cache. However I don't know if this "intelligent" guessing is really worth it in the end.

PR #2012, opened 2024-09-02
Price scatter date xaxis

christopherlam:
image

### DIFF
--- a/gnucash/report/reports/standard/price-scatter.scm
+++ b/gnucash/report/reports/standard/price-scatter.scm
@@ -55,9 +55,6 @@
      options gnc:pagename-general
      optname-from-date optname-to-date "a")
 
-    (gnc:options-add-interval-choice! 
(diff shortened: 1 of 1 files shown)


christopherlam:
Any particular opinion on this? @jralls I think the uneven date axis is a feature rather than a bug...

jralls (commented):
I don't think the chart is particularly useful, there are better ways to get stock price charts. It's anyway misnamed, it's a plain line chart, not a scatterplot. That's a good thing, a price scatterplot wouldn't be very useful either. 

I'm all for getting rid of the different intervals, they don't even change the displayed data, just the numbers along the X axis, and those numbers are themselves pretty useless.
However using the dates of the prices to set the vertical lines makes for a strange-looking plot and it's going to get congested for plots with either a long period or showing frequent quotes. I think it would be better to break the period up into some consistent number of intervals and put date labels on each interval.

jralls:
> @christopherlam requested a review from @jralls yesterday

Why? Nothing has changed.

PR #2011, opened 2024-09-01
csv import date: Add dateformat "Locale" to pick current locale

christopherlam:
another approach to #2010 -- add "Locale" which uses current icu parser with current locale. still has same fault as #2010.

PR #2015, opened 2024-09-08
[gnc-datetime] add boost UK/US/ISO date parsers

christopherlam:
From https://github.com/Gnucash/gnucash/pull/1710#issuecomment-2336692212

UK: "30 Sep 2024" and "30 September 2024" are accepted. "30 Sept 2024" is rejected.
US: "May 2, 2024" is accepted.
ISO delimited: "2024 Dec 20" and "2024 December 20" are accepted.

Note boost's parser requires 4-digit years.

### DIFF
--- a/libgnucash/engine/gnc-datetime.cpp
+++ b/libgnucash/engine/gnc-datetime.cpp
@@ -77,13 +77,16 @@ static constexpr auto ticks_per_second = INT64_C(1000000);
 static constexpr auto ticks_per_second = INT64_C(1000000000);
 #endif
 
-/* Vector of date formats understood by gnucash and corresponding regex
- * to parse each from an external source
+/* Vector of date formats understood by gnucash and corresponding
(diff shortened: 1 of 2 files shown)


gjanssens:
> From [#1710 (comment)](https://github.com/Gnucash/gnucash/pull/1710#issuecomment-2336692212)
> 
> UK: "30 Sep 2024" and "30 September 2024" are accepted. "30 Sept 2024" is rejected. US: "May 2, 2024" is accepted. ISO delimited: "2024 Dec 20" and "2024 December 20" are accepted.
> 
> Note boost's parser requires 4-digit years.

As @jralls already remarked on #1710 , the issue remains that this solution is very limited. It solves your case, but will not add anything for non-English speaking regions. If we merge this I expect users to request similar options for other regions, which we can't offer via boost.

christopherlam:
Closed in favour of #2011

PR #2019, opened 2024-09-15
Transaction Cut and Paste problems

Bob-IT:
A couple of bug fixes...
The first commit I think is correct but am unsure on second one, it works but may be I am missing something.

Below is an image illustrating problem...
cutpaste

The last transaction was copied and pasted to the blank transaction and as you can see the transfer account is that of the current register. The transaction is still correct but could be confusing and hence the bug report.

To fix this, when the register loads an existing check is made to ensure the transaction and splits being edited are in the split list we are loading and as part of that I am recording the anchor split of the pending transaction. If the pending transaction is the blank transaction I use that to set the leading virtual cell in the register.

There is one situation when this would fail, if the copied transaction has two splits for the same account as it would always pick the same anchor split. I do not think this is a common occurrence so for most users not a problem. I did think of another way to fix this which involved passing the copied current cursor split to `gnc_txn_to_float_txn` and making sure it was the first in the floating split list but decided to do it register load.

### DIFF
--- a/gnucash/register/ledger-core/split-register.c
+++ b/gnucash/register/ledger-core/split-register.c
@@ -73,6 +73,7 @@ typedef struct
     };
     CursorClass cursor_class;
     GncGUID leader_guid;
+    gint anchor_split_index;
 } ft_fs_store;
 
 static ft_fs_store copied_item = { 0, { NULL } };
@@ -101,6 +102,7 @@ clear_copied_item()
     copied_item.ft = NULL;
     copied_item.cursor_class = CURSOR_CLASS_NONE;
     copied_item.leader_guid = *guid_null();
+    copied_item.anchor_split_index = 0;
 }
 
 static void
@@ -828,6 +830,7 @@ gnc_split_register_copy_current_internal (SplitRegister* reg,
             }
 
             copied_item.leader_guid = info->default_account;
+            copied_item.anchor_split_index = xaccTransGetSplitIndex (trans, split);
         }
     }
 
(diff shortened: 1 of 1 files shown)


jralls:
Everything else aside these changes illustrate the poor quality of the splt register code.

`clear_copied_item` needs to be moved because it also clears `copied_class` and `copied_leader_guid`. That's easily improved by making them part of the struct so they can be referred to in code as `copied_item->class` and `copied_item->leader_guid` so that the need to call `clear_copied_item` before using them is evident.

The fact that you had to make the same change 3 times in the *****576-line***** `gnc_split_register_load` shows that it's a function desperately in need of refactoring as does the fact that its use to paste a single transaction when it's documented to load the split register from a list. 

>  I did think of another way to fix this which involved passing the copied current cursor split to gnc_txn_to_float_txn and making sure it was the first in the floating split list but decided to do it register load.

Why? the `gnc_txn_to_float_txn` seems cleaner in principal.

jralls:
Tested it with a transaction with two splits to the register account, works fine. They're actually pretty common for stock sale transactions: One split for the actual sale (the "other" split being the account where the proceeds go) and a second capital gain/loss split (the other being the cap gains income account).

Bob-IT:
OK, I have added a commit to add the `copied_class` and `copied_leader_guid` to `copied_item`. I always start from a position that someone more experienced has done it that way for a reason.

Also fixed the second bug a different way by adding the copied anchor split index to `copied_item` and using that index to reference the blank split when copying onto the blank transaction.

jralls:
Thanks, Bob. It looks good.

PR #2028, opened 2024-09-30
Delete commodity show accounts

christopherlam:
If deleting commodity is impossible because there are accounts using the commodity, show them.


christopherlam:
any feedback on i18n strings

jralls:
Looks fine.

PR #2030, opened 2024-10-03
Close tabs when middle-clicked

rkok:
A feature I've been wanting for a while, that I'm used to having with browser tabs. ;-)

I've had to split up tab_hbox somewhat, to make it possible to catch middle clicks. Despite some attempts to make it work, I could not find a way to handle middle clicks on GtkNotebook tabs natively. Documentation on the matter seems non-existent.

### DIFF
--- a/gnucash/gnome-utils/gnc-main-window.cpp
+++ b/gnucash/gnome-utils/gnc-main-window.cpp
@@ -1607,6 +1607,33 @@ gnc_main_window_update_all_titles (void)
                    nullptr);
 }
 
+/*  Callback function invoked when the user clicks on a GtkNotebook tab.
+ *
+ *  This function is needed to make it possible to close a tab
+ *  when it's clicked using the middle mouse button;
+ *  there does not seem to be a way to do this with GtkNotebook natively.
+ *
+ *  @param widget The event box in the tab, which was clicked.
+ *
+ *  @param event The event parameter describing where on the screen
+ *  the mouse was pointing when clicked, type of click, modifiers,
+ *  etc.
+ *
+ *  @param page This is the GncPluginPage corresponding to the tab.
+ *
+ *  @return Returns TRUE if this was a middle-click, meaning Gnucash
+ *  handled the click.
+ */
+static gboolean
+gnc_tab_clicked_cb(GtkWidget *widget, GdkEventButton *event, GncPluginPage *page) {
+    if (event->type == GDK_BUTTON_PRESS && event->button == 2)
+    {
+        gnc_main_window_close_page(page);
+        return TRUE;
+    }
+    return FALSE;
+}
+
 static void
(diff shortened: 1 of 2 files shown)


jralls:
I'm inclined to reject this as not being cross-platform: Neither macOS nor Windows support a middle button. @gjanssens what do you think?

rkok:
> I'm inclined to reject this as not being cross-platform: Neither macOS nor Windows support a middle button.

Hi @jralls, could you elaborate a bit on what additional work would need to be done to support those platforms? I'm happy to give it a try - I want to contribute.

In case I wasn't clear enough, this PR is to catch middle _mouse_ button clicks; when pressing the middle mouse button (usually the scroll wheel) on a tab heading, the window closes, just like with Firefox/Chrome/etc. browser tabs.

It uses [native features of Gdk](https://docs.gtk.org/gdk3/struct.EventButton.html) to identify which mouse button was pressed, which as far as I could see requires no OS-specific tweaking. Whatever compatibility layer is used to run this on Windows/macOS should let it gracefully degrade by simply not catching the click, I would think. (Though I did not test it on platforms other than Linux)

middleclick

Bob-IT:
I was not sure about this so fired up chrome on my MS Windows 10 box, opened up a couple of tabs and used the middle wheel button to close one, also using that button on the Gnucash toolbar pops the menu.

Regarding the code, there is no need for changes to the .h file, just make your call back function static and add a similar entry to the start of the `gnc-main-window.cpp` file.

Also I am not sure about using an event box, I wonder if it would be better to have...

> tab_container (GtkBox)
> > tab_content (GtkButton)
> > > button_container (GtkBox)
> > > > image (GtkImage, optional)
> > > > label (GtkLabel)
> > > > entry (GtkEntry, hidden)

> close_button (GtkButton, if not immutable)

I would have to try it and see if it would work and whether the tabs would be coloured correctly. Another reason for going this way is that Gtk4 has dropped GtkEventBox should we stick with Gtk.


Either way, if changing the layout you may need to change the functions that gain access to these widgets, some may be ...
`gnc_main_window_update_tab_*`
`main_window_find_tab_items`
`main_window_find_tab_widget`

Also I think there may be an enhancement bug for this, will need to find it, found it 493595.

rkok:
Thanks for your input @Bob-IT!

Indeed some other places needed touching up due to the changed box structure. That's now done, and that callback function is now also static.

I've tested by:
- changing the name and color of an account
- checking the tab tooltip
- changing settings under Preferences > Window

All of that works well now.

I've tried your suggested alternative to GtkEventBox as well as some other types of containers, but GtkEventBox really seems to be the only one that'll catch middle-clicks.

Let me know if you have any further feedback. Hope this is getting ready for prime-time :-)

jralls:
Thanks!

PR #2031, opened 2024-10-11
Bug799430 Split auto complete bug

Bob-IT:
I think the first commit fixes the issue that if you try to auto complete a split using the memo text from a split that has no values it fails to fill in the transfer account. This has stemmed from a change in the return value of [xaccSplitSetSharePrice](https://github.com/Gnucash/gnucash/commit/87b0a41e9b0b4db6556523ce5e3009fa03dd7734#diff-154ee983af0cf96a53b09614667e634d03e6a3b4321c6913046d04bbbfc8494cR1196) in version 4.10

Just looking for confirmation of change before I push this.

The second commit is just realigning the source file after the change to cpp.

### DIFF
--- a/gnucash/register/ledger-core/split-register-control.cpp
+++ b/gnucash/register/ledger-core/split-register-control.cpp
@@ -696,7 +696,7 @@ gnc_find_split_in_trans_by_memo (Transaction *trans, const char *memo,
         if (unit_price)
         {
             gnc_numeric price = xaccSplitGetSharePrice (split);
-            if (!gnc_numeric_equal (price, gnc_numeric_create (1, 1)))
+            if (!gnc_numeric_zero_p (price))
                 continue;
         }
 
--- a/gnucash/register/ledger-core/split-register-control.cpp
+++ b/gnucash/register/ledger-core/split-register-control.cpp
@@ -1,5 +1,5 @@
 /********************************************************************\
- * split-register-control.c -- split register control object        *
+ * split-register-control.cpp -- split register control object      *
  * Copyright (C) 2017 Aaron Laws                                    *
  *                                                                  *
  * This program is free software; you can redistribute it and/or    *
@@ -50,20 +50,20 @@
 static QofLogModule log_module = GNC_MOD_LEDGER;
 
 static inline bool
-check_imbalance_fraction (const SplitRegister* reg,
-                          const gnc_monetary* imbal_mon,
-                          const Transaction* trans)
+check_imbalance_fraction (const SplitRegister *reg,
+                          const gnc_monetary *imbal_mon,
+                          const Transaction *trans)
 {
-    auto commodity_fraction{gnc_commodity_get_fraction (imbal_mon->commodity)};
+    auto commodity_fraction {gnc_commodity_get_fraction (imbal_mon->commodity)};
     auto denom_diff = imbal_mon->value.denom > commodity_fraction;
     if (!denom_diff)
     {
         const auto imbal_comm = imbal_mon->commodity;
-        for (auto node = xaccTransGetSplitList(trans); node;
-             node = g_list_next(node))
+        for (auto node = xaccTransGetSplitList (trans); node;
+             node = g_list_next (node))
         {
-            auto acc = xaccSplitGetAccount(GNC_SPLIT (node->data));
-            if (xaccAccountGetCommodity(acc) == imbal_comm &&
+            auto acc = xaccSplitGetAccount (GNC_SPLIT(node->data));
+            if (xaccAccountGetCommodity (acc) == imbal_comm &&
                 imbal_mon->value.denom > xaccAccountGetCommoditySCU (acc))
             {
                 denom_diff = true;
@@ -73,8 +73,11 @@ check_imbalance_fraction (const SplitRegister* reg,
     }
 
     if (denom_diff)
-        gnc_error_dialog(gnc_ui_get_main_window(GTK_WIDGET(reg)), "%s", _("This transaction cannot be balanced: The imbalance is a fraction smaller than the commodity allows."));
-
+    {
+        gnc_error_dialog (gnc_ui_get_main_window (GTK_WIDGET(reg)),
+                          "%s",
+                          _("This transaction cannot be balanced: The imbalance is a fraction smaller than the commodity allows."));
+    }
     return denom_diff;
 }
 
@@ -93,7 +96,6 @@ gnc_split_register_balance_trans (SplitRegister *reg, Transaction *trans)
     Split *other_split;
     gboolean two_accounts;
     gboolean multi_currency;
-
 
     if (xaccTransIsBalanced (trans))
         return FALSE;
@@ -114,8 +116,8 @@ gnc_split_register_balance_trans (SplitRegister *reg, Transaction *trans)
         {
             imbal_mon = static_cast<gnc_monetary*>(imbal_list->data);
             if (!imbal_list->next &&
-                    gnc_commodity_equiv(gnc_monetary_commodity(*imbal_mon),
-                                        xaccTransGetCurrency(trans)))
+                gnc_commodity_equiv (gnc_monetary_commodity (*imbal_mon),
+                                     xaccTransGetCurrency (trans)))
                 multi_currency = FALSE;
             else
                 multi_currency = TRUE;
@@ -126,7 +128,7 @@ gnc_split_register_balance_trans (SplitRegister *reg, Transaction *trans)
 
         /* We're done with the imbalance list, the real work will be done
            by xaccTransScrubImbalance which will get it again. */
-        gnc_monetary_list_free(imbal_list);
+        gnc_monetary_list_free (imbal_list);
     }
     else
         multi_currency = FALSE;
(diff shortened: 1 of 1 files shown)


gnucash/register/ledger-core/split-register-control.cpp
```diff
@@ -86,17 +89,16 @@ gnc_split_register_balance_trans (SplitRegister *reg, Transaction *trans)
...
     const char *message = _("The current transaction is not balanced.");
     Split *split;
     Split *other_split;
     gboolean two_accounts;
     gboolean multi_currency;
 
-
     if (xaccTransIsBalanced (trans))
-        return FALSE;
+        return false;
```
jralls: `FALSE` is correct here because the function return type is gboolean, an alias for int. `FALSE` is a macro `#define`d as `0`. `false` and `true` are the possible values of the C++ `bool` and C `_Bool` types. While it compiles either way with an implicit cast it's more correct to use the value that matches the type.

gnucash/register/ledger-core/split-register-control.cpp
```diff
@@ -109,46 +111,48 @@ gnc_split_register_balance_trans (SplitRegister *reg, Transaction *trans)
...
     }
     else
-        multi_currency = FALSE;
+        multi_currency = false;
 
     split = xaccTransGetSplit (trans, 0);
     other_split = xaccSplitGetOtherSplit (split);
 
-    if (other_split == NULL)
+    if (other_split == nullptr)
```
jralls: @This is better without the comparison, i.e. `if (!other_split)`.

gnucash/register/ledger-core/split-register-control.cpp
```diff
@@ -424,10 +426,9 @@ gnc_split_register_move_cursor (VirtualLocation *p_new_virt_loc,
     gboolean saved;
     SRInfo *info;
 
-    ENTER("reg=%p, p_new_virt_loc=%p (%d, %d)",
-          reg, p_new_virt_loc,
-          new_virt_loc.vcell_loc.virt_row,
-          new_virt_loc.vcell_loc.virt_col);
+    ENTER("reg=%p, p_new_virt_loc=%p (%d, %d)", reg, p_new_virt_loc,
+                                                new_virt_loc.vcell_loc.virt_row,
+                                                new_virt_loc.vcell_loc.virt_col);
```
jralls: The old version is indented correctly: Aligned with the opening parenthesis.

gnucash/register/ledger-core/split-register-control.cpp
```diff
@@ -616,9 +619,8 @@ gnc_split_register_move_cursor (VirtualLocation *p_new_virt_loc,
 
     *p_new_virt_loc = new_virt_loc;
 
-    PINFO ("after move %d %d\n",
-           new_virt_loc.vcell_loc.virt_row,
-           new_virt_loc.vcell_loc.virt_col);
+    PINFO("after move %d %d", new_virt_loc.vcell_loc.virt_row,
+                              new_virt_loc.vcell_loc.virt_col);
```
jralls: As with the `ENTER` call above the old indentation is correct.

gnucash/register/ledger-core/split-register-control.cpp
```diff
@@ -908,79 +910,77 @@ gnc_split_register_auto_completion (SplitRegister *reg,
...
             gnc_split_register_save_data_destroy (sd);
         }
 
         gnc_resume_gui_refresh ();
 
         /* now move to the non-empty amount column unless config setting says not */
-        if ( !gnc_prefs_get_bool(GNC_PREFS_GROUP_GENERAL_REGISTER,
-                                 GNC_PREF_TAB_TRANS_MEMORISED) )
+        if (!gnc_prefs_get_bool (GNC_PREFS_GROUP_GENERAL_REGISTER,
+                                  GNC_PREF_TAB_TRANS_MEMORISED))
```
jralls: Adds an extra space to the second line.

gnucash/register/ledger-core/split-register-control.cpp
```diff
@@ -1141,11 +1141,11 @@ gnc_split_register_check_stock_action (SplitRegister *reg,
         return;
     shares = gnc_price_cell_get_value ((PriceCell *) cell);
 
-    if ((buy  && !gnc_numeric_positive_p (shares)) ||
-            (sell &&  gnc_numeric_positive_p (shares)))
+    if ((buy && !gnc_numeric_positive_p (shares)) ||
+            (sell && gnc_numeric_positive_p (shares)))
```
jralls: `(sell` should line up with `(buy`.

gnucash/register/ledger-core/split-register-control.cpp
```diff
@@ -1176,12 +1176,12 @@ gnc_split_register_check_stock_shares (SplitRegister *reg,
         return;
     name = ((ComboCell *)cell)->cell.value;
 
-    if (!g_strcmp0(name, "") ||
-            !g_strcmp0(name, buy ? ACTION_SELL_STR : ACTION_BUY_STR))
+    if (!g_strcmp0 (name, "") ||
+            !g_strcmp0 (name, buy ? ACTION_SELL_STR : ACTION_BUY_STR))
```
jralls: The `!g_strcmp0`s should line up.

gnucash/register/ledger-core/split-register-control.cpp
```diff
@@ -846,59 +848,59 @@ gnc_split_register_auto_completion (SplitRegister *reg,
...
         if (!gnc_cell_name_equal (cell_name, DESC_CELL))
-            return FALSE;
+            return false;
 
         /* nothing but the date, num, and description should be changed */
         /* FIXME, this should be refactored. */
         if (gnc_table_layout_get_cell_changed (reg->table->layout,
-                                               XFRM_CELL, TRUE)  ||
+                                               XFRM_CELL, true) ||
                 gnc_table_layout_get_cell_changed (reg->table->layout,
```
jralls: All the function calls should line up with the first one.

jralls (commented):
The first commit makes sense and is fine. The spacing one needs some QA. What tool did you use for indentation and spaces-before-braces?

gnucash/register/ledger-core/split-register-control.cpp
Bob-IT: Changed.

gnucash/register/ledger-core/split-register-control.cpp
Bob-IT: Changed.

gnucash/register/ledger-core/split-register-control.cpp
Bob-IT: My error, changed.

gnucash/register/ledger-core/split-register-control.cpp
Bob-IT: Changed back.

gnucash/register/ledger-core/split-register-control.cpp
Bob-IT: Changed back.

Bob-IT:
> The first commit makes sense and is fine. The spacing one needs some QA. What tool did you use for indentation and spaces-before-braces?

No tool used, I just find it easier to read the source files if there is a consistent use of spaces and just changed it visually.

gnucash/register/ledger-core/split-register-control.cpp
Bob-IT: I was not aware of this, I use Geany to edit the source files and it highlights `nullptr`, `true`, and `false` for `cpp` files so just thought I would do that also. 
Maybe I should just change them back and see if I can change Geany in some way.

gnucash/register/ledger-core/split-register-control.cpp
Bob-IT: I did not want to make this sort of change due to possible errors as the other changes were just find and replace.

gnucash/register/ledger-core/split-register-control.cpp
jralls: An IDE can't do that right without being able to parse the code. There's a protocol called [LSP](https://en.wikipedia.org/wiki/Language_Server_Protocol) that connects editors/IDEs to compilers and there's a [bare-bones Geany plugin](https://github.com/techee/geany-lsp), but it looks like it leaves a lot of configuration work to the user.

jralls:
> No tool used, I just find it easier to read the source files if there is a consistent use of spaces and just changed it visually.

Using a tool is easier. We've used [astyle](http://astyle.sourceforge.net/) in the past, usually with args `--indent=spaces=4 --brackets=break --suffix=none` though the last one I did was `-xd -k1 -m0 -M60 -xL -xC79 -OHpUcZns4 --brackets=break`. That was on the SQL backend before I started the C++ rewrite. Phil Longstaff had a rather unique formatting style and it needed a lot of adjustments.

Since you did it by eye, a question: Why do you like a space between a function name and its opening parenthesis but not between a preprocessor macro and its opening parenthesis?

jralls:
Go ahead and merge when you're ready.

gnucash/register/ledger-core/split-register-control.cpp
Bob-IT: Thank you, will have a look at that when I have time.

Bob-IT:
Will install `astyle` and do some testing. 

>Since you did it by eye, a question: Why do you like a space between a function name and its opening parenthesis but not between a preprocessor macro and its opening parenthesis?

If I remember correctly, in the early days when I was submitting patches on bug reports it was suggested to do it that way.

PR #2036, opened 2024-10-19
Bug 799443 - Edit find account jump then open, open the last highlighted account not the jumped to account

christopherlam:
but `gnc_tree_view_account_set_selected_account` is called from multiple places. @Bob-IT ?

### DIFF
--- a/gnucash/gnome-utils/gnc-tree-view-account.c
+++ b/gnucash/gnome-utils/gnc-tree-view-account.c
@@ -1485,7 +1485,7 @@ gnc_tree_view_account_set_selected_account (GncTreeViewAccount *view,
     }
     gtk_tree_path_free(parent_path);
 
-    gtk_tree_selection_select_path (selection, s_path);
+    gtk_tree_view_set_cursor (GTK_TREE_VIEW (view), s_path, 0, FALSE);
 
     /* give gtk+ a chance to resize the tree view first by handling pending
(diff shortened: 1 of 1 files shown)


Bob-IT:
I think there are several possibilities for this...

It does seem reasonable that if you select an account then the cursor should follow but you would have to check where used.

Another option would be to add a parameter to `gnc_tree_view_account_set_selected_account` that selectively did the set cursor but that would involve updating all occurrences accordingly.

So possibly create `gnc_tree_view_account_set_selected_account_with_cusor` which calls the modified and renamed static version of `gnc_tree_view_account_set_selected_account` as above, this would then be called from the exposed version of `gnc_tree_view_account_set_selected_account`. Then `gnc_tree_view_account_set_selected_account_with_cusor` would replace the call in `gnc_plugin_page_account_tree_open`

jralls:
> It does seem reasonable that if you select an account then the cursor should follow but you would have to check where used.

All of the other uses are account selectors in dialogs where keyboard editing (which is what the cursor is for) isn't an issue. Having the cursor follow the selection will be harmless.

Separately it's not necessary to call both `gtk_tree_view_set_cursor` and `gtk_tree_selection_select_path`, GtkTreeSelection handles the `cursor_changed` signal emitted by `gtk_tree_view_set_cursor` and moves the selection to the new cursor.

PR #2042, opened 2024-10-29
Bug 799451 - Cannot translate different meanings of string  "_Balance" to Chinese.

taoyouh:
The previous fix (https://github.com/Gnucash/gnucash/commit/2e52086451549fc58355ac656e702f952bc352f4) for this ticket added context for some uses of string "Balance" but that's not part of the original ticket and doesn't have ambiguity (it's always used as a noun). The real problem "_Balance" (the underscore indicates it is a label of an action) however is not fully covered. This ticket adds context all uses of "_Balance" and removed the unnecesary contexts for "Balance".


jralls:
Sorry for the misunderstanding and thanks for fixing it.

PR #2043, opened 2024-11-01
Add missing `#include <cstdint>`

SoapGentoo:
* GCC 15 has reduced transitive dependencies between headers.

Bug: https://bugs.gentoo.org/939856

### DIFF
--- a/bindings/guile/gnc-kvp-guile.cpp
+++ b/bindings/guile/gnc-kvp-guile.cpp
@@ -2,6 +2,7 @@
 #include <kvp-frame.hpp>
 #include <libguile.h>
 #include <numeric>
+#include <cstdint>
 
 #include <config.h>
 
--- a/gnucash/gnome-utils/gnc-option-gtk-ui.cpp
+++ b/gnucash/gnome-utils/gnc-option-gtk-ui.cpp
@@ -25,6 +25,7 @@
 #include "gnc-option-gtk-ui.hpp"
 #include <config.h>  // for scanf format string
 #include <memory>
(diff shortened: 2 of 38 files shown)


bindings/guile/gnc-kvp-guile.cpp
```diff
@@ -2,6 +2,7 @@
 #include <kvp-frame.hpp>
 #include <libguile.h>
 #include <numeric>
+#include <stdint.h>
```
jralls: stdint.h is the C header. Please use cstdint in C++ files.

jralls (changes requested):

bindings/guile/gnc-kvp-guile.cpp
SoapGentoo:
```
#include <cstdint> -> std::intX_t
#include <stdint.h> -> intX_t
```
one problem of gnucash is that it relies on `<cstdint>` providing unqualified `intX_t` which is not guaranteed by the standard. `<stdint.h>` guarantees unqualified types.

jralls: You're being pedantic. While it's true that the standard doesn't require exporting the standard integer types to the global namespace, all three of the remaining C++ standard libraries (I.e. those that are continuing to support the C++ standard as it evolves) do so.

SoapGentoo: updated

jralls (approved):

jralls:
Thanks!

PR #2045, opened 2024-11-21
fix build with boost 1.87

landryb:
replace the now dropped `recursive_directory_iterator` usage by https://en.cppreference.com/w/cpp/filesystem/copy which is present in c++17. tested to be backwards compatible at build and runtime with boost 1.84.


jralls:
What exactly is broken? `recursive_directory_iterator` is [still in the latest git](https://github.com/boostorg/filesystem/blob/0848f5347b69bd9f8f0459dc0cb88e7a52714448/include/boost/filesystem/directory.hpp#L842) without any indication of being deprecated.

While I'm in favor of switching to std::filesystem, it must be a complete replacement of boost::filesystem, not a mixed bag.

Obviously it's got to pass CI. Considering that none of the CI jobs even built you must have turned off -Werror when you tested. Don't do that.

landryb:
> What exactly is broken? `recursive_directory_iterator` is [still in the latest git](https://github.com/boostorg/filesystem/blob/0848f5347b69bd9f8f0459dc0cb88e7a52714448/include/boost/filesystem/directory.hpp#L842) without any indication of being deprecated.

it fails to build in our ports bulk build testing the update to 1.87b1, cf https://marc.info/?l=openbsd-ports&m=173219077014769&w=2

i've fixed the first error by replacing io_service by io_context, then stumbled upon the `recursive_directory_iterator` being missing. I dont have the error at hand, but gnucash 5.9 failed to build. The gnucash code didnt change since then afaict..
> 
> While I'm in favor of switching to std::filesystem, it must be a complete replacement of boost::filesystem, not a mixed bag.

my goal was to have gnucash building, i'm not a c++ hacker, and don't intend to convert the complete gnucash codebase. The patch is here as a start if other packagers for other OSes want to fix the potential build failure they see if they try with boost 1.87b1.
 
> Obviously it's got to pass CI. Considering that none of the CI jobs even built you must have turned off -Werror when you tested. Don't do that.

i didnt build from git so didn't switch `-Werror` off, but from the 5.9 tarball, then forward ported the patch to a throwaway git clone i did to upstream the patch. I can drop the unused var just to please the CI if you like :)

landryb:
i can't reproduce the error i was seeing, but it was a linking error that lead me to think that `recursive_directory_iterator` was mssing.

Now that i think of it, it's probably because i had a mixture of boost 1.84 & 1.87b1 installed at the same time, and since i've cleaned up the system i cant reproduce anymore. So technically, to fix the build with boost 1.87 only the `io_service`>`io_context` change is relevant.

Whether you want to take the rest of the patch to convert `copy_recursive` to `std::filesystem` is your choice :) i can amend the PR to only do the build fix if you prefer.

jralls:
> Whether you want to take the rest of the patch to convert copy_recursive to std::filesystem is your choice :) i can amend the PR to only do the build fix if you prefer.

Yes, please. 

The boost/asio/io_context.h change requires at least Boot 1.66. We already have a soft-ish requirement on Boost 1.67, with an exception for >1.60 with a patch removing the use of the old `auto_ptr` in boost::locale.  You can either remove the [exception code](https://github.com/Gnucash/gnucash/blob/764157ddeb6e7eb90e4c9fdae00ec51dcfda35ea/CMakeLists.txt#L584) and make 1.67 required, in which case [the note](https://github.com/Gnucash/gnucash/blob/764157ddeb6e7eb90e4c9fdae00ec51dcfda35ea/README.dependencies#L69) in README.dependencies also needs editing, or you can create a `HAVE_BOOST_1_67` configure variable and ifdef io_service.h/io_context.h.

landryb:
fwiw there has to be something weird with directory iterators in boost 1.87, because when working on gnuradio i face those build errors:
```
/usr/obj/ports/gnuradio-3.8.2.0/gnuradio-3.8.2.0/gnuradio-runtime/lib/prefs.cc:66:31: error: expected ';' after expression
        fs::directory_iterator diritr(dir);
                              ^
                              ;
/usr/obj/ports/gnuradio-3.8.2.0/gnuradio-3.8.2.0/gnuradio-runtime/lib/prefs.cc:66:13: error: no member named 'directory_iterator' in namespace 'boost::filesystem'
```
im definitely not comfy with boost but this feels strange.

jralls:
> im definitely not comfy with boost but this feels strange.

It looks to me like a basic C++ issue: `boost::filesystem::directory_iterator` is declared in filesystem/directory.hpp, and prefs.cc doesn't include it, nor does any other boost::filesystem header. Lacking that inclusion the compiler doesn't recognize `fs::directory_iterator` as a type so it expects a semicolon after it.  I'm not sufficiently interested in GnuRadio to figure out for you what changed when.

landryb:
> in which case [the note](https://github.com/Gnucash/gnucash/blob/764157ddeb6e7eb90e4c9fdae00ec51dcfda35ea/README.dependencies#L69) in README.dependencies also needs editing

that note already specifies 1.67 is required..

jralls:
>> in which case [the note](https://github.com/Gnucash/gnucash/blob/764157ddeb6e7eb90e4c9fdae00ec51dcfda35ea/README.dependencies#L69) in README.dependencies also needs editing

> that note already specifies 1.67 is required..

Right, but the bit at the end about some distributions having patched earlier versions needs to go because you're removing the exception code that recognizes the patches.

jralls:
Thanks!

PR #2046, opened 2024-11-29
SX editor - use integer-type schema parameters rather than string and…

gjanssens:
… double

GnuCash expects these values to be integers anyway, so this avoids needless type conversions

### DIFF
--- a/gnucash/gnome/gnc-plugin-page-sx-list.cpp
+++ b/gnucash/gnome/gnc-plugin-page-sx-list.cpp
@@ -508,7 +508,7 @@ gnc_plugin_page_sx_list_create_widget (GncPluginPage *plugin_page)
 
     /* Set the paned position from the preferences, default 160 */
     gtk_paned_set_position (GTK_PANED(priv->widget),
-                            gnc_prefs_get_float (GNC_PREFS_GROUP_SXED,
+                            gnc_prefs_get_int (GNC_PREFS_GROUP_SXED,
                                                  GNC_PREF_DIVIDER_POS));
 
     {
@@ -585,9 +585,8 @@ gnc_plugin_page_sx_list_create_widget (GncPluginPage *plugin_page)
         g_object_ref_sink (priv->gdcal);
 
         /* Set number of months from preference, default 12 */
-        gchar *num_months = gnc_prefs_get_string (GNC_PREFS_GROUP_SXED, GNC_PREF_NUM_OF_MONTHS);
-        gnc_dense_cal_set_num_months (priv->gdcal, atoi (num_months));
-        g_free (num_months);
+        auto num_months = gnc_prefs_get_int (GNC_PREFS_GROUP_SXED, GNC_PREF_NUM_OF_MONTHS);
+        gnc_dense_cal_set_num_months (priv->gdcal, num_months);
 
         gtk_container_add (GTK_CONTAINER(swin), GTK_WIDGET(priv->gdcal));
     }
@@ -766,23 +765,19 @@ gnc_plugin_page_sx_list_cmd_save_layout (GSimpleAction *simple,
 {
     auto plugin_page = GNC_PLUGIN_PAGE_SX_LIST(user_data);
     GncPluginPageSxListPrivate *priv;
-    gchar *num_of_months;
     gint paned_position;
 
     g_return_if_fail (GNC_IS_PLUGIN_PAGE_SX_LIST(plugin_page));
 
     priv = GNC_PLUGIN_PAGE_SX_LIST_GET_PRIVATE(plugin_page);
 
-    num_of_months = g_strdup_printf ("%d", gnc_dense_cal_get_num_months (priv->gdcal));
     paned_position = gtk_paned_get_position (GTK_PANED(priv->widget));
 
-    gnc_prefs_set_float (GNC_PREFS_GROUP_SXED, GNC_PREF_DIVIDER_POS,
+    gnc_prefs_set_int (GNC_PREFS_GROUP_SXED, GNC_PREF_DIVIDER_POS,
                          paned_position);
 
-    gnc_prefs_set_string (GNC_PREFS_GROUP_SXED, GNC_PREF_NUM_OF_MONTHS,
-                          num_of_months);
-
-    g_free (num_of_months);
+    gnc_prefs_set_int (GNC_PREFS_GROUP_SXED, GNC_PREF_NUM_OF_MONTHS,
+                          gnc_dense_cal_get_num_months (priv->gdcal));
 }
 
 static void
--- a/gnucash/gnome/gnc-plugin-page-sx-list.cpp
+++ b/gnucash/gnome/gnc-plugin-page-sx-list.cpp
@@ -586,6 +586,11 @@ gnc_plugin_page_sx_list_create_widget (GncPluginPage *plugin_page)
 
         /* Set number of months from preference, default 12 */
         auto num_months = gnc_prefs_get_int (GNC_PREFS_GROUP_SXED, GNC_PREF_NUM_OF_MONTHS);
+        if (num_months == 0)
+        {
+            PWARN ("Got invalid value '0' for number of months to display. This suggests a gsettings configuration issue. Continuing with a default value of 12 instead.");
+            num_months = 12;
+        }
         gnc_dense_cal_set_num_months (priv->gdcal, num_months);
 
         gtk_container_add (GTK_CONTAINER(swin), GTK_WIDGET(priv->gdcal));
--- a/gnucash/gschemas/org.gnucash.GnuCash.dialogs.sxs.gschema.xml.in
+++ b/gnucash/gschemas/org.gnucash.GnuCash.dialogs.sxs.gschema.xml.in
@@ -54,12 +54,12 @@
       <summary>How many days in advance to notify the user.</summary>
       <description>How many days in advance to notify the user.</description>
     </key>
-    <key name="number-of-months" type="s">
-      <default>'12'</default>
+    <key name="number-of-months" type="i">
+      <default>12</default>
       <summary>The number of months to be shown in editor.</summary>
       <description>The number of months to be shown in editor.</description>
     </key>
-    <key name="divider-position" type="d">
+    <key name="divider-position" type="i">
       <default>160</default>
       <summary>The horizontal position of the editor divider.</summary>
       <description>The horizontal position of the editor divider.</description>


gnucash/gnome/gnc-plugin-page-sx-list.cpp
```diff
@@ -585,9 +585,13 @@ gnc_plugin_page_sx_list_create_widget (GncPluginPage *plugin_page)
...
-        gchar *num_months = gnc_prefs_get_string (GNC_PREFS_GROUP_SXED, GNC_PREF_NUM_OF_MONTHS);
-        gnc_dense_cal_set_num_months (priv->gdcal, atoi (num_months));
-        g_free (num_months);
+        auto num_months = gnc_prefs_get_int (GNC_PREFS_GROUP_SXED, GNC_PREF_NUM_OF_MONTHS);
+        if (num_months == 0)
+        {
+            PWARN ("Got invalid value '0' for number of months to display. This suggests a gsettings configuration issue. Continuing with a default value of 12 instead.");
+            num_months = 12;
+        }
+        gnc_dense_cal_set_num_months (priv->gdcal, num_months);
```
jralls, line 594: This fixes [Bug 799470](https://bugs.gnucash.org/show_bug.cgi?id=799470) so please use a bugfix commit message.

But `number-of-months` doesn't show up in dconf in Linux nor org.gnucash.GnuCash.plist on mac, there's no setting for it in Preferences>Scheduled transactions, and changing it in an SX Editor tab doesn't persist if I close the tab and open a new one so I'm a bit mystified about why it works and why it failed for the reporter of that bug.

jralls: From[Bug 799470 Comment 15] https://bugs.gnucash.org/show_bug.cgi?id=799470#c15):
> The code to store the number of months is only called when you select View->Save layout as default.

That explains it: After Save layout `number-of-months` is in org.gnucash.GnuCash.plist.

gjanssens: > This fixes [Bug 799470](https://bugs.gnucash.org/show_bug.cgi?id=799470) so please use a bugfix commit message.
> 

I'm not sure yet this fixes the bug. It does indeed prevent the crash, but as I explain on the bug the crash is the result of not properly being able to access the schema in the first place. We haven't figured out yet why that happens for the bug reporter. If we do need to add code to fix that part as well, I'll use the bugfix commit message for it.

jralls: OK, no harm in waiting.

Bob-IT:
Sorry for the inconvenience this has caused, it took a while to remember why I used those settings, When I first looked at implementing this I added a couple of default preference settings using a spin button for the panel setting and a combo for the month, these two widgets were giving a double and text as values and that is why I chose them. I then thought of adding the menu action 'Save Layout as default' which was simpler and decided to drop the preference changes as they were not realy needed but unfortunately did bot revisit the choices for the settings.

These values are only saved by the menu action and used when the schedule editor page is created. When Gnucash is closed with that page open, the values are saved as part of the `gcm` page settings.

I do not think I have ever tested when the key did not exist, did not think it was possible.

By changing the `gschemas`, does that reset the two values to default entries ?
I do not think it is a problem for these entries, unfortunately I have not tried it as I broke my PC.

gjanssens:
> Sorry for the inconvenience this has caused, it took a while to remember why I used those settings, When I first looked at implementing this I added a couple of default preference settings using a spin button for the panel setting and a combo for the month, these two widgets were giving a double and text as values and that is why I chose them. I then thought of adding the menu action 'Save Layout as default' which was simpler and decided to drop the preference changes as they were not realy needed but unfortunately did bot revisit the choices for the settings.

Ah, that makes sense.

> I do not think I have ever tested when the key did not exist, did not think it was possible.

As far as I know that can only happen if the wrong schema is loaded at runtime. Schemas are compiled and installed by our build scripts, but read by gsettings based on the paths found in XDG_DATA_DIRS. If where gnucash installs them is different from where gsettings is looking for them you may get mismatches and hence preference read/write issues. gsettings is quite blunt at this and simply aborts the running program. We have written some code around it to first test for a key and only if it's defined return it. Note that if a key is defined querying it will always return a value. Either whatever was set or the default value. I'm still puzzled how the reporter of bug 799470 gets into the situation where no value is returned from gsettings.

> 
> By changing the `gschemas`, does that reset the two values to default entries ? I do not think it is a problem for these entries, unfortunately I have not tried it as I broke my PC.

Thanks for asking, I did test before submitting this PR. By changing the type of the gsettings setting, the eventually already stored value is discarded and the default value of the new type is returned. So this seems to be safe to do (at least on Fedora 39).

jralls:
> By changing the type of the gsettings setting, the eventually already stored value is discarded and the default value of the new type is returned. So this seems to be safe to do (at least on Fedora 39).

Safe but annoying. Please add a notice about that in the commit message for the release notes.

jralls:
@gjanssens release is only a few days away. Are you ready to wrap this up?

gjanssens:
Sure.

How to though ? I'm fairly sure the root cause of the crash is not that we missed a sanity check on the gsettings query, but that for some reason the wrong gsettings schema is picked up by the user. We can easily work around this in this case by the extra check, but it can happen again each time the schema from a system-installed gnucash lags behind a user built gnucash - and likely when installed in /usr/local rather than an arbitrary local installation directory.

As there's no response on the bug any more, I'm afraid we won't be able to get to the bottom of it via that user.

To not delay this further I propose to make it two commits: one with the change to use integers and a second one to add the validation which prevents the crash. That way it's easier to convey the purpose of each commit in the commit messages.

jralls:
> To not delay this further I propose to make it two commits: one with the change to use integers and a second one to add the validation which prevents the crash. That way it's easier to convey the purpose of each commit in the commit messages.

If that means the commits currently in the PR, good enough for this release, but we need to null-protect all calls to any g_settings_get method that returns a pointer unless we can ensure that they never will.

To that end: Since every build comes with its own set of schemas how about creating a preference with a key equal to the current build id and checking for its existence at startup? If it doesn't match we log an error (for command line use) and if we have a gui put up an error box explaining that there's a mismatch and that Bad Things will happen if it isn't fixed?

gjanssens:
> > To not delay this further I propose to make it two commits: one with the change to use integers and a second one to add the validation which prevents the crash. That way it's easier to convey the purpose of each commit in the commit messages.
> 
> If that means the commits currently in the PR, good enough for this release, but we need to null-protect all calls to any g_settings_get method that returns a pointer unless we can ensure that they never will.
> 

Way back when I initially did the port from gconf to gsettings, the Gtk developers (naively?) considered a missing schema a programming error, which is why gsettings bluntly aborts when it can't find the proper schema. It seems to me they missed the complexity of reality in which multiple installs step on each other's toes.

> To that end: Since every build comes with its own set of schemas how about creating a preference with a key equal to the current build id and checking for its existence at startup? If it doesn't match we log an error (for command line use) and if we have a gui put up an error box explaining that there's a mismatch and that Bad Things will happen if it isn't fixed?

I was hoping to detect whether the issue in bug 779470 was due to a loophole in our build scripts or due to in improper install or just something we simply can't avoid. I think only in the latter case we really do need to guard against schema mismatches. But on the other hand it doesn't hurt to have a canary like the one you propose. It can serve to alert us devs when a user found yet another way to run gnucash with the wrong schema.

gjanssens:
Hmm, long-winded train of thoughts and slightly beside de point. We already guard against gtk's aborts by first checking whether a gsettings key exists before looking up its value. The issue here is that we return some sane defaults in case a key doesn't exist. The default we chose for a string is a null pointer. Yes, the calling code should check that indeed before dereferencing.

That aside, I still like your suggestion of making a unique key in the schema that changes with each gnucash version and then testing if the right one exists early at boot.

jralls:
I think it's something we can't avoid, at least on Linux where we have a much less controlled execution environment than on Flatpak, macOS, and Windows. Even with perfect installation scripts it's too easy for the distribution's installed schema to come before the running GnuCash's schemas. That's unfortunately also true of shared libraries and guile modules.

> The issue here is that we return some sane defaults in case a key doesn't exist. The default we chose for a string is a null pointer. Yes, the calling code should check that indeed before dereferencing.

Are you going to work on that? GVariant* returns are safe because both uses test it with `g_variant_is_of_type` and that returns `FALSE` if passed a nullptr.

Alternatively you could change the default string value to `g_strdup ("")` on the premise that leaks are better than crashes.

gjanssens:
For what it's worth, I quickly set up a Ubuntu system in an attempt to reproduce the issue found in the bug.

I installed system gnucash which is 5.5 as on the mint system
I then built gnucash 5.9 and installed it in /usr/local
When I run that gnucash, it perfectly finds the proper schema files in /usr/local/share/glib2.0
So I strongly suspect something is wrong in the build environment of the original poster but we'll probably won't get more info.

> I think it's something we can't avoid, at least on Linux where we have a much less controlled execution environment than on Flatpak, macOS, and Windows. Even with perfect installation scripts it's too easy for the distribution's installed schema to come before the running GnuCash's schemas. That's unfortunately also true of shared libraries and guile modules.

I actually think we can avoid it in much the same way as we do it on all other platforms - enforce that the directory leading to the correct set of schemas is always the first path in `XDG_DATA_DIRS`. When I first did the conversion I was very conservative and [avoided messing with `XDG_DATA_DIRS` if the install location was either /usr/local or /usr](https://github.com/Gnucash/gnucash/blob/stable/gnucash/CMakeLists.txt#L212). I think it's safe to change this and always prepend `{GNC_HOME}/share` to `XDG_DATA_DIRS`. That would at least ensure that our own schema is always found first while running gnucash.

That should reduce the schema related issues to installation errors. The only one I can think of that would not show a build error is if the user disables compiling the schemas. We have a build-time switch for that, intended for packagers. But if that switch is disabled during a normal install on linux, the schema compilation will not happen and hence the schemas won't be found. That's a user error however.

> 
> > The issue here is that we return some sane defaults in case a key doesn't exist. The default we chose for a string is a null pointer. Yes, the calling code should check that indeed before dereferencing.
> 
> Are you going to work on that? GVariant* returns are safe because both uses test it with `g_variant_is_of_type` and that returns `FALSE` if passed a nullptr.

Ok. I think though that all our gsettings queries in themselves are safe these days. The all return *a* value (nullptr being one possible option). It's the calling code that has to check whether the return value is within the bounds it expects. Out of bounds can happen in several situations: querying a non-existing key, could return an unexpected default value. A settings manipulation outside of gnucash (via say dconf-editor or the defaults system on MacOS) could also set a key to an unsupported value.

I don't intend to work on this, no. TBH I think the issue is rare enough that there are more pressing matters to attend to should I have spare time :(

If you agree with my proposal above about unconditionally prefixing XDG_DATA_DIRS, I can still push that small change. I don't have enough bandwith to implement a canary or do a deep dive on validity checking for all our gsettings queries.

gjanssens:
Oh small sidenote: while testing on Ubuntu I noticed our uninstall code doesn't do a very good job. Several file were left behind, among which all our schema files and several tools in the bin directory. I'll write a bug report later so this titbit is at least recorded.

jralls:
> If you agree with my proposal above about unconditionally prefixing XDG_DATA_DIRS,

Yes, that seems like a good idea.

> It's the calling code that has to check whether the return value is within the bounds it expects.

Of course. And of course it's a bug to dereference a pointer without null-checking it when nullptr is a possible result from the function used to initialize said pointer.

> I don't intend to work on this, no.

OK, I'll do it today. There are only 21 calls to `gnc_prefs_get_string`, it won't take long.

jralls:
> OK, I'll do it today. There are only 21 calls to gnc_prefs_get_string, it won't take long.

It turned out that they were all checked already, either immediately after the call or by passing the rv to a function that null checked the parameter.

PR #2050, opened 2024-12-10
Bug 799420 - gnucash 5.8 crashes on windows 10 while importing qif

jralls:
xaccTransCommitEdit will delete the transaction if there aren't at least 2 splits and that will cause a UAF in xaccTransRecordPrice so test the splits before creating the gnc-transaction.


christopherlam:
scm is ok
Would it be safer to handle UAF in xaccTransRecordPrice?

jralls:
> Would it be safer to handle UAF in xaccTransRecordPrice?

 Too late. Once the F has happened any access is UA. ;-)

PR #2052, opened 2024-12-18
Bug 799487 - Unable to save gnucash DB file as XML file

jralls:
Not quite true, it just takes a really long time for a large database.

The underlying problem is gnc_file_do_save_as reloads the data to make sure that the save-as saves everything. On the SQL backend that triggers a scrub. The scrub itseld doesn't take long, but every transaction commit was logged in the transaction log and did a refresh of the registers. So:
* Suspend logging while doing the scrub.
* Suspend UI refreshes while reloading the data.
* Make the register page event handler obey the refresh being suspended.

@Bob-IT I'm a bit concerned about that last bullet: Is there a reason that `gnc_plugin_page_register_event_handler` doesn't respect the suspend? Or is it that none of the handlers check whether the GUI is suspended because suspension is supposed to be handled somewhere else?

### DIFF
--- a/gnucash/gnome-utils/gnc-file.c
+++ b/gnucash/gnome-utils/gnc-file.c
@@ -1593,8 +1593,12 @@ gnc_file_do_save_as (GtkWindow *parent, const char* filename)
     }
 
     /* Make sure all of the data from the old file is loaded */
+    qof_event_suspend ();
+    gnc_suspend_gui_refresh ();
     qof_session_ensure_all_data_loaded(session);
-
+    gnc_resume_gui_refresh ();
+    qof_event_resume ();
+    
     /* -- this session code is NOT identical in FileOpen and FileSaveAs -- */
 
     save_in_progress++;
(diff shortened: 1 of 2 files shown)


Bob-IT:
I would need to check this but I think it is how the handler was registered, there seems to be two ways either with `qof_event_register_handler` or `gnc_register_gui_component`. I think with the second method, the gui suspend level is checked but with the first function it is a different suspend level.

jralls:
> I think with the second method, the gui suspend level is checked but with the first function it is a different suspend level.

Right, `qof_event_suspend` and `qof_event_resume`. I bracketed the data reload with those instead of checking for gui refreshes being suspended in `gnc_plugin_page_register_event_handler`. It's more appropriate too because that latter function doesn't actually do a UI refresh in the normal sense: It does a ton of stuff that looks to me like it should be done once in `gnc_plugin_page_register_init`. Doing it every time a QOF Event fires seems insane.

PR #2053, opened 2024-12-24
Improve zh_CN translation for accuracy, naturalness and formality.

xuxinhang:
Those updates are introduced during [my translating gnucash-docs screenshots](/Gnucash/gnucash-docs/pull/347).

Many items are updated, involving punctuation usage, accounting terms and confusing expressions.

I modified the po file directly, so no WebTranslate. I have built and checked 5.10 version, and the new po file seems to break nothing.


jralls:
Thanks!

PR #2055, opened 2025-01-05
Fix incorrect deallocation type

peter15914:
Objects created with ```dom_tree_to_guid``` must be destroyed with ```guid_free```.


jralls:
Thanks!

PR #2056, opened 2025-01-07
corrected 2 grammatical mistakes in fr.po

ferdinandhubbard981:
Fixed two grammatical errors in the french translation (fr.po). 

Updated `~a jusqu'à ~a` to `du ~a au ~a`, as the latter is the correct grammatical form for date ranges in french. This is also how it is translated in other reports.


jralls:
Please use [Weblate](https://hosted.weblate.org/projects/gnucash/gnucash/fr/) for translation changes.

PR #2057, opened 2025-01-10
Bug 799514 - Pie chart hovering shows wrong information in pop-up

agwekixj:
Set intersect to true for pie charts.


jralls:
@christopherlam Any comment? The bug is a side effect of your #2047.

christopherlam:
Thata's a reasonable approach. Alternatively `account-piecharts.scm` sets the option manually.

jralls:
Thanks, merged after rebase.

PR #2058, opened 2025-01-19
Use std mismatch

christopherlam:


libgnucash/engine/Account.cpp
```diff
@@ -3259,35 +3262,18 @@ gnc_account_get_full_name(const Account *account)
...
+    std::for_each (path.rbegin(), path.rend(),
+                   [&gs](auto a)
+                   {
+                       if (gs->len)
+                           gs = g_string_append (gs, account_separator);
+                       gs = g_string_append (gs, xaccAccountGetName (a));
+                   });
 
-    return fullname;
+    return g_string_free (gs, false);
```
jralls: `FALSE`, not `false`: `g_string_free` takes a gboolean not a bool.

libgnucash/engine/Account.cpp
```diff
@@ -3259,35 +3262,18 @@ gnc_account_get_full_name(const Account *account)
...
-     * becomes the terminating nullptr pointer for the array of strings. */
-    names = (const gchar **)g_malloc(level * sizeof(gchar *));
-    names[--level] = nullptr;
-    for (a = account; level > 0; a = priv->parent)
-    {
-        priv = GET_PRIVATE(a);
-        names[--level] = priv->accountName;
-    }
+    auto path{gnc_account_get_path (account)};
+    auto gs = g_string_new (nullptr);
```
jralls: This is severely pessimized over what you're replacing because of the repeated allocations. Call [`g_string_sized_new`](https://docs.gtk.org/glib/ctor.String.sized_new.html) to do the allocation in one go. You could get the size with another loop or just allocate a big buffer, maybe 512 or 1024 bytes.

libgnucash/engine/Account.cpp
```diff
@@ -3242,15 +3242,18 @@ xaccAccountGetName (const Account *acc)
     return GET_PRIVATE(acc)->accountName;
 }
 
+std::vector<const Account*>
+gnc_account_get_path (const Account *account)
```
jralls: `path` might be a little confusing in this context, how about `ancestry` or `all_parents`?

jralls (changes requested):
"std mismatch"?

libgnucash/engine/Account.cpp
```diff
@@ -3268,35 +3262,22 @@ gnc_account_get_full_name(const Account *account)
...
-    /* Build the full name */
-    fullname = g_strjoinv(account_separator, (gchar **)names);
-    g_free(names);
-
-    return fullname;
+    auto path{gnc_account_get_all_parents (account)};
+    auto seps_size{path.empty() ? 0 : strlen (account_separator) * (path.size() - 1)};
+    auto alloc_size{std::accumulate (path.begin(), path.end(), seps_size,
+                                     [](auto sum, auto acc)
+                                     { return sum + strlen (xaccAccountGetName (acc)); })};
```
jralls, line 3269: You need more than one `seps_size` in the result:
```
    auto alloc_size{std::accumulate (path.begin(), path.end(), 0,
                                     [seps_size](auto sum, auto acc)
                                     { return sum + strlen (xaccAccountGetName (acc)) + seps_size; })};
```
The trailing `seps_size` conveniently makes at least enough space for the trailing `'\0'` and a few extra bytes does no harm.

christopherlam: My original was correct; checked via printing `strlen()` and `alloc_size`. If we're not using GString reallocation then lets' fill the buffer manually...

jralls: Sorry, I misinterpreted `seps_size`.

jralls (approved):

PR #2060, opened 2025-01-24
Bug 797824 - Inconsistent behavior in Loan Review

Genie-Liu:
fix the shift issue in loan review. More discussion is https://bugs.gnucash.org/show_bug.cgi?id=797824

### DIFF
--- a/gnucash/gnome/assistant-loan.cpp
+++ b/gnucash/gnome/assistant-loan.cpp
@@ -1977,10 +1977,11 @@ loan_rev_get_loan_range( LoanAssistantData *ldd, GDate *start, GDate *end )
 
     endDateMath = g_new0( struct tm, 1 );
     g_date_to_struct_tm (ldd->ld.startDate, endDateMath);
-    monthsTotal = ( (ldd->ld.numPer - 1)
+    monthsTotal = ( ldd->ld.numPer
                     * ( ldd->ld.perSize == GNC_MONTHS ? 1 : 12 ) );
     endDateMath->tm_mon += monthsTotal;
     gnc_gdate_set_time64 (end, gnc_mktime (endDateMath));
+    g_date_subtract_days( end, 1 );
     g_free (endDateMath);
 }
 
@@ -2108,13 +2109,13 @@ loan_rev_recalc_schedule( LoanAssistantData *ldd )
             char *eloc;
             rowNumData =
                 (gnc_numeric*)g_hash_table_lookup( repayment_schedule,
-                                                   &curDate );
+                                                   &nextDate );
             if ( rowNumData == NULL)
             {
                 int j;
                 GDate *dateKeyCopy = g_date_new();
 
-                *dateKeyCopy = curDate;
+                *dateKeyCopy = nextDate;
                 rowNumData = g_new0( gnc_numeric, ldd->ld.revNumPmts );
                 g_assert( rowNumData != NULL );
                 for ( j = 0; j < ldd->ld.revNumPmts; j++ )
@@ -2198,13 +2199,13 @@ loan_rev_recalc_schedule( LoanAssistantData *ldd )
                     | GNC_HOW_RND_ROUND_HALF_UP;
                 gnc_numeric val;
                 rowNumData = (gnc_numeric*)g_hash_table_lookup( repayment_schedule,
-                             &curDate );
+                             &nextDate );
                 if ( rowNumData == NULL )
                 {
                     int j;
                     GDate *dateKeyCopy = g_date_new();
 
-                    *dateKeyCopy = curDate;
+                    *dateKeyCopy = nextDate;
                     rowNumData = g_new0( gnc_numeric, ldd->ld.revNumPmts );
                     g_assert( rowNumData != NULL );
                     for ( j = 0; j < ldd->ld.revNumPmts; j++ )
@@ -3084,4 +3085,3 @@ gnc_ui_sx_loan_assistant_create (void)
 
     gnc_window_adjust_for_screen (GTK_WINDOW(ldd->window));
 }
-


jralls:
Hmm, still not right. 6 Months of loan but only 5 in the  summary. Whole loan has the same problem.
Screenshot 2025-01-24 at 15 16 28

Screenshot 2025-01-24 at 15 15 57

It's still an improvement: Before your change it only went for 4 months.

Genie-Liu:
Hmm, interesting. Thank you for the feedback, It seems that after the change, the first payments is correct now. But I have a question. In my current version, the number of loan payments is accurate, yet you mentioned there are only four? Could there be a difference in settings? Here are my parameters:

Screenshot 2025-01-25 at 15 20 57
Screenshot 2025-01-25 at 15 21 14
Screenshot 2025-01-25 at 15 21 21

jralls:
Interesting. The difference is the start date: You're starting the loan and the payments on Feb. 1, I'm leaving the loan start date as today, the default, and changing the payment start to Feb 1. Your screenshot also shows the bad first payment meaning it's without your change. Here's what it looks like with the change:
image

So it's correct as long as the loan start and payment start are the same.

Genie-Liu:
I see. I think my fix has resolved the issue with the first incorrect payment.

The missing payment in your previous figure is due to the start date configuration. The current mechanism calculates the end date based on the start date and the number of periods. In your earlier setting, the start date was set to the default, let’s assume it was 01/25/2025, with a 6-month duration. The end date would then be calculated as 06/25/2025. However, if we start the payment from 02/01/2025 and aim for 6 payments, the last payment would fall on 07/01/2025, which is after the calculated end date and it will be filtered during calculating the review.

A propose solution is to calculate the payments review using the remain months. This also aligns with the final created schedules.

Genie-Liu:
If we aim to create a loan schedule starting in the middle of the loan term, we will still have issues. For example, suppose the loan begins on 12/01/2024 with a 6-month period, and we intend to start payments on 02/01/2025. 

The current result is incorrect cause the first payment is wrong and the last payment is missing:
image

If we apply this change the fixed version should produce the payments starting at 02/01/2025 and end at 05/01/2025. The payments' date is correct. But the interest is still not the right one.

jralls:
>The missing payment in your previous figure is due to the start date configuration. The current mechanism calculates the end date based on the start date and the number of periods.

Fair enough, but a partial payment on the first day of the loan is called a down payment and technically isn't part of the loan. For house mortgages the first payment is usually near the beginning of the following month; auto loans may defer payments for up to a year.

> If we aim to create a loan schedule starting in the middle of the loan term, we will still have issues. For example, suppose the loan begins on 12/01/2024 with a 6-month period, and we intend to start payments on 02/01/2025.

>If we apply this change the fixed version should produce the payments starting at 02/01/2025 and end at 05/01/2025. The payments' date is correct. But the interest is still not the right one.

A 6-month loan beginning on 12/1 is due on 5/31. Strictly speaking if we start monthly payments on 2/1 then we would expect to have 5 payments of roughly $210 with the last one on 5/31.  But we have a problem in the Months Remaining box: If we leave it to itself it changes the number to 4 and assumes that we've already made two payments; if we push it to 6 it decides that we have until 7/1 to pay. Fixing that's out of scope for the bug and is deeper in the logic of the assistant. I'm satisfied that your current change fixes the date issues sufficiently to satisfy the bug. I'll do a code review tomorrow (it's already past my bedtime), but assuming it all looks good I'm willing to merge this as-is and I'll leave it up to you whether to work on getting the interest calculations right on a separate PR.

Genie-Liu:
> Fixing that's out of scope for the bug and is deeper in the logic of the assistant. I'm satisfied that your current change fixes the date issues sufficiently to satisfy the bug.

Yeah, I agree. The generation logic need more discussion. I will revert the repayments generation code that based on remain month. 

> I'll leave it up to you whether to work on getting the interest calculations right on a separate PR.

The interest fix is also included in this pr. It depends on the remain month. Maybe you can help take a look.

jralls:
You locked in on monthly payments but the payments option has several options. For example setting payments to semi-monthly produces:
Screenshot 2025-01-26 at 08 57 43

Screenshot 2025-01-26 at 08 57 56
Screenshot 2025-01-26 at 08 58 23

**Edit:**
~~Without your changes the result is more correct though it of course ignores the start of payments date:~~
Without your changes the result has the right number of payments but doesn't adjust the payment amount so it overpays by 2x:

Screenshot 2025-01-26 at 09 04 12

jralls:
Notice that the formula (labeled Amount) is unaffected by the payments schedule. That explains a lot of the problem. I wondered if it had been noticed and found [a lot of bugs](https://bugs.gnucash.org/buglist.cgi?bug_status=NEW&bug_status=ASSIGNED&bug_status=NEEDINFO&bug_status=REOPENED&component=Scheduled%20Transactions&list_id=24303&longdesc=mortgage&longdesc_type=allwordssubstr&order=Importance&query_format=advanced) documenting problems with it.

Genie-Liu:
It appears there are several issues that still require to fix. I believe there's at least four issues need to be addressed.

1. The initial payment date is incorrect.. This is what I plan to address. 
2. If the start date does not align with the repayment date, the final payment is omitted. This issue arises when the start date is left as default, . The root cause is how we calculate the end date. (If you set the length to be 1 year, you will observe that no payments are generated in the review at all)
3. The calculations for interest and principal are inaccurate when the payment date does not align with the loan's start date. This issue arises due to the review calculation beginning with an index of 1.
4. The current system does not support non-monthly repayment frequencies. 

This PR will focus on addressing issues 1 and 2, the monthly repayment schedule as it is the most commonly used. 
Issues 3 and 4 will be tackled in a separate PR.

Genie-Liu:
@jralls Please help test if the first and second issue are resolved after the new change.

The expected behavior should be: 
1. Setting loan start date to default like 01/27/2025 with length 6 months, and repayment start date at 02/01/2025. There should be 6 repayments in the review. 
2. Changing length to 1 year will also fix the no repayments issue.

jralls:
@Genie-Liu the date part seems to work as expected now. Please squash out the extra commits.

A note about item 3: The interest is wrong *unless* the the loan start date is exactly one month before the start of payments. The GnuCash [pmt function](https://github.com/Gnucash/gnucash/blob/55465e88ed33ba3ecb9a08cc4c6decb52182372e/bindings/guile/fin.scm#L48) works exactly like the spreadsheet function of the same name, and the last argument is a boolean `Type` where `false`/`0` means to calculste the interest at the *end* of the payment period while a `true` (anything not `0`) means to calculate it at the *beginning* of the period. As the Amount formula shows--and you can verify this easily with a spreadsheet--we're using `0` for `Type`.

Genie-Liu:
Commits squashed. Thanks for the tips.

jralls:
Thanks!

PR #2062, opened 2025-02-24
Add 'Customer's Invoices' to customer list popup

altera2015:
Adds a new menu entry called: "Customer's Invoices" to the popup menu in the Customer List.

### DIFF
--- a/gnucash/ui/gnc-plugin-page-owner-tree.ui
+++ b/gnucash/ui/gnc-plugin-page-owner-tree.ui
@@ -42,7 +42,7 @@
       <attribute name="temp" translatable="no">yes</attribute>
     </item>
   </menu>
-  
+
   <menu id="EditPlaceholder2-e">
     <item>
       <attribute name="label" translatable="yes">E_dit Employee</attribute>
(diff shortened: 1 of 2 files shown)


christopherlam:
I think this is a nice addition.

Suggestions: 
- instead of "Customer's Invoices" maybe "Search Invoices..."
- also add to Vendor/Employee tabs
- menuitem immediately after "New Invoice..."

gnucash/ui/gnc-plugin-page-owner-tree.ui
```diff
@@ -170,6 +170,11 @@
         <attribute name="action">GncPluginPageOwnerTreeActions.OTProcessPaymentAction</attribute>
         <attribute name="tooltip" translatable="yes">Process Payment</attribute>
       </item>
+      <item>
+        <attribute name="label" translatable="yes">Customer's Invoices</attribute>
+        <attribute name="action">GncPluginPageOwnerTreeActions.OTSearchInvoicesAction</attribute>
+        <attribute name="tooltip" translatable="yes">Customer's Invoices</attribute>
```
jralls, line 176: The menu item label and tooltip are respectively misleading and redundant. The former should be "Find Customer Invoice…" (the ellipsis indicates that the menu item opens a dialog that requires further input to do anything useful) and the tooltip should be something like "Open the Find Invoice Dialog with the invoices already filtered for this customer."

jralls (changes requested):
Please provide the same feature for vendor bills and employee vouchers.

I agree with Chris that the new entry belongs either before or after the New XXX entry.

christopherlam:
this branch merged in, with amendments as discussed. thanks!

PR #2063, opened 2025-02-26
Bug 798946 - start/end of current/last quarter have off-by-one error

agwekixj:
Correct determination of current quarter.

PR #2064, opened 2025-02-27
Use the fiscal start day of the month for relative quarter reports

agwekixj:
Correct determination of current quarter.

### DIFF
--- a/libgnucash/engine/gnc-option-date.cpp
+++ b/libgnucash/engine/gnc-option-date.cpp
@@ -552,8 +552,8 @@ gnc_relative_date_to_time64(RelativeDatePeriod period)
             auto month = now.tm_mon - (12 * year_delta);
             auto year = now.tm_year + year_delta;
             now.tm_mday = days_in_month(month, year);
-	        if (offset == RelativeDateOffset::QUARTER && acct_per.tm_mday > 1 && acct_per.tm_mday <= now.tm_mday)
-	            now.tm_mday = acct_per.tm_mday - 1;
+            if (offset == RelativeDateOffset::QUARTER && acct_per.tm_mday > 1 && acct_per.tm_mday <= now.tm_mday)
+                now.tm_mday = acct_per.tm_mday - 1;
             gnc_tm_set_day_end(&now);
         }
     }
--- a/libgnucash/engine/gnc-option-date.cpp
+++ b/libgnucash/engine/gnc-option-date.cpp
@@ -536,9 +536,10 @@ gnc_relative_date_to_time64(RelativeDatePeriod period)
 
     switch (checked_reldate(period).m_type)
     {
-        case RelativeDateType::LAST:
+        case RelativeDateType::ABSOLUTE:
+	case RelativeDateType::LAST:
         case RelativeDateType::NEXT:
-            // Do nothing for LAST and NEXT.
+            // Do nothing
             break;
         case RelativeDateType::START:
             now.tm_mday = offset == RelativeDateOffset::QUARTER && acct_per.tm_mday > 1 ? acct_per.tm_mday : 1;
--- a/libgnucash/engine/gnc-option-date.cpp
+++ b/libgnucash/engine/gnc-option-date.cpp
@@ -463,26 +463,6 @@ normalize_reldate_tm(struct tm& now)
     }
 }
 
-static void
-reldate_set_day_and_time(struct tm& now, RelativeDateType type)
-{
-    if (type == RelativeDateType::START)
-    {
-        gnc_tm_set_day_start(&now);
-        now.tm_mday = 1;
-    }
-    else if (type == RelativeDateType::END)
-    {
-        /* Ensure that the month is between 0 and 11*/
-        auto year_delta = (now.tm_mon / 12) + (now.tm_mon < 0 ? -1 : 0);
-        auto month = now.tm_mon - (12 * year_delta);
-        auto year = now.tm_year + year_delta;
-        now.tm_mday = days_in_month(month, year);
-        gnc_tm_set_day_end(&now);
-    }
-    // Do nothing for LAST and NEXT.
-};
-
 time64
 gnc_relative_date_to_time64(RelativeDatePeriod period)
 {
@@ -502,10 +482,12 @@ gnc_relative_date_to_time64(RelativeDatePeriod period)
                             GNC_PREF_START_CHOICE_ABS))
         acct_per = static_cast<tm>(GncDateTime(gnc_accounting_period_fiscal_start()));
 
-    switch(reldate_offset(period))
+    RelativeDateOffset offset = reldate_offset(period);
+
+    switch(offset)
     {
         case RelativeDateOffset::NONE:
-// Report on today so nothing to do
+            // Report on today so nothing to do
             break;
         case RelativeDateOffset::YEAR:
             if (reldate_is_prev(period))
@@ -517,7 +499,7 @@ gnc_relative_date_to_time64(RelativeDatePeriod period)
             else if (gnc_relative_date_is_ending(period))
                 now.tm_mon = 11;
             break;
-       case RelativeDateOffset::SIX:
+        case RelativeDateOffset::SIX:
             if (reldate_is_prev(period))
                 now.tm_mon -= 6;
             else if (reldate_is_next(period))
@@ -537,9 +519,9 @@ gnc_relative_date_to_time64(RelativeDatePeriod period)
             else if (reldate_is_next(period))
                 now.tm_mon += 3;
             if (gnc_relative_date_is_ending(period))
-                now.tm_mon += 2;
+                now.tm_mon += offset == RelativeDateOffset::QUARTER && acct_per.tm_mday > 1 ? 3 : 2;
             break;
-       case RelativeDateOffset::MONTH:
+        case RelativeDateOffset::MONTH:
             if (reldate_is_prev(period))
                 --now.tm_mon;
             else if (reldate_is_next(period))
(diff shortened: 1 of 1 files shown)


jralls:
@agwekixj I guess you're trying to accommodate non-month-aligned accounting periods like @christopherlam's UK tax year example. That would also need adjusting the start and end days of the quarter. I think it should be a separate commit.

agwekixj:
@jralls I agree which is why I created this as a separate patch.   The descriptions are currently the same but that's editable, right?

jralls:
> @jralls I agree which is why I created this as a separate patch. The descriptions are currently the same but that's editable, right?

You can edit the PR description here but you need to change the commit description locally and then force-push it.

jralls:
@agwekixj I've merged #2063 [as you suggested](https://bugs.gnucash.org/show_bug.cgi?id=798946#c42) so now you need to rebase on it to resolve the conflicts. When you do please change the first commit's name since it's no longer directly about the bug. You can also squash out the later commits.

libgnucash/engine/gnc-option-date.cpp
```diff
@@ -551,7 +533,31 @@ gnc_relative_date_to_time64(RelativeDatePeriod period)
...
+            auto year_delta = (now.tm_mon / 12) + (now.tm_mon < 0 ? -1 : 0);
+            auto month = now.tm_mon - (12 * year_delta);
+            auto year = now.tm_year + year_delta;
+            now.tm_mday = days_in_month(month, year);
+            if (offset == RelativeDateOffset::QUARTER && acct_per.tm_mday > 1 && acct_per.tm_mday <= now.tm_mday)
+                now.tm_mday = acct_per.tm_mday - 1;
+            gnc_tm_set_day_end(&now);
+        }
+    }
+
```
jralls: I would rather this stay a separate function. `gnc_relative_date_to_time64` is already 70 lines; this pushes it to just short of 100.

libgnucash/engine/gnc-option-date.cpp
```diff
@@ -537,9 +519,9 @@ gnc_relative_date_to_time64(RelativeDatePeriod period)
             else if (reldate_is_next(period))
                 now.tm_mon += 3;
             if (gnc_relative_date_is_ending(period))
-                now.tm_mon += 2;
+                now.tm_mon += (offset == RelativeDateOffset::QUARTER && acct_per.tm_mday > 1 ? 3 : 2);
```
jralls: I don't care much for the conditional referring to the previous case here. How about instead having 
```
    if (gnc_relative_date_is_ending(period))
        --delta;
    now.tm_mon -= delta;
}
    [[fallthrough]]
```
in the `QUARTER` case?

libgnucash/engine/gnc-option-date.cpp
```diff
@@ -551,7 +533,31 @@ gnc_relative_date_to_time64(RelativeDatePeriod period)
...
+
+    switch (checked_reldate(period).m_type)
+    {
+        case RelativeDateType::ABSOLUTE:
+	case RelativeDateType::LAST:
+        case RelativeDateType::NEXT:
+            // Do nothing
+            break;
+        case RelativeDateType::START:
+            now.tm_mday = offset == RelativeDateOffset::QUARTER && acct_per.tm_mday > 1 ? acct_per.tm_mday : 1;
```
jralls: `acct_per.tm_mday > 1` is unnecessary.

libgnucash/engine/gnc-option-date.cpp
```diff
@@ -551,7 +533,31 @@ gnc_relative_date_to_time64(RelativeDatePeriod period)
...
+            break;
+        case RelativeDateType::END:
+        {
+            /* Ensure that the month is between 0 and 11*/
+            auto year_delta = (now.tm_mon / 12) + (now.tm_mon < 0 ? -1 : 0);
+            auto month = now.tm_mon - (12 * year_delta);
+            auto year = now.tm_year + year_delta;
+            now.tm_mday = days_in_month(month, year);
+            if (offset == RelativeDateOffset::QUARTER && acct_per.tm_mday > 1 && acct_per.tm_mday <= now.tm_mday)
+                now.tm_mday = acct_per.tm_mday - 1;
```
jralls: Doesn't cover quarters ending in February when the  accounting period begins on the 30th or 31st of a longer month.

jralls (changes requested):

libgnucash/engine/gnc-option-date.cpp
agwekixj: It should now.

libgnucash/engine/gnc-option-date.cpp
agwekixj: When gnc_prefs_get_bool (GNC_PREFS_GROUP_ACCT_SUMMARY, GNC_PREF_START_CHOICE_ABS) returns false, acct_per.tm_mday remains 0.  Without the test, the start of the quarter would be normalized to the last day of the previous quarter.

libgnucash/engine/gnc-option-date.cpp
agwekixj: Done.

libgnucash/engine/gnc-option-date.cpp
agwekixj: I've reestablished the separate function passing two additional arguments.

libgnucash/engine/gnc-option-date.cpp
jralls: Hmm. Perhaps `acct_per` should always be initialized with `gnc_accounting_period_fiscal_start`: It should return something sane (generally the 1 January in either the current or previous calendar year) even when the accounting period preference isn't set to an absolute date.

jralls (commented):
Other than the open initializing of `acct_per` it looks good.

libgnucash/engine/gnc-option-date.cpp
agwekixj: I don't think that's necessary.   The selection of _Relative_ or _Absolute_ in the _Start Date_ preference should be sufficient.  I suspect the default date value may be 31 December 1969.

agwekixj:
Thanks for taking the time.

libgnucash/engine/gnc-option-date.cpp
jralls: >  I suspect the default date value may be 31 December 1969.

The current null-initialized value is 1900-01-01 00:00:00 UTC; what that ends up being in local time depends on one's timezone.

The relative selections are start of this or the previous (calendar) month, quarter, or year, with the default being the current year. `gnc_accounting_period_fiscal_start` returns whichever of those is selected.

It may not be necessary, but ISTM it's clearly *better* to always initialize it to `gnc_accounting_period_fiscal_start`. The null-initialized value is wrong and having it imposes an otherwise unnecessary test at the start of the function and another one here.

agwekixj: > > I suspect the default date value may be 31 December 1969.
> 
> The current null-initialized value is 1900-01-01 00:00:00 UTC; what that ends up being in local time depends on one's timezone.

The date that appears by default on the _Accounting Period_ Preference page is 31 December 1969.  This is the date value I suspected would be returned by `gnc_accounting_period_fiscal_start()` unless the user had specified another value. 

> 
> The relative selections are start of this or the previous (calendar) month, quarter, or year, with the default being the current year. `gnc_accounting_period_fiscal_start` returns whichever of those is selected.

Having now reviewed `gnc_accounting_period_fiscal_start()`, I can see _gnc_accounting_period_fiscal_start()_ will return a _Relative_ start date when _Absolute_ is not selected.  Note: The relative selection also includes _Today_

>
> It may not be necessary, but ISTM it's clearly _better_ to always initialize it to `gnc_accounting_period_fiscal_start`. The null-initialized value is wrong and having it imposes an otherwise unnecessary test at the start of the function and another one here.

As _Today_ would not reliably be the first day of the month, the code would still need to test for _Absolute_.

I'll update the patch again and run some more tests.

jralls: > As Today would not reliably be the first day of the month, the code would still need to test for Absolute.

Right, but it won't be 0. If the user has today for their accounting period start, then why would that not be the day that quarters should align on?

Edit: I suppose we could instead set `acct_per.tm_mday = 1` if ABSOLUTE isn't set. It would be nicer to just initialize `struct tm acct_per{0, 0, 0, 1, 0, 100, 0, 0, 0, 0}` ( 1 Jan. 2000) but IIRC that would need an ifdef because MinGW's `struct tm` doesn't have the timezone fields.

agwekixj: > Right, but it won't be 0. If the user has today for their accounting period start, then why would that not be the day that quarters should align on?

In stable, the quarter's start month is adjusted when _Absolute_ is selected in the _Start Date_  on the _Accounting Period_ preferences.  The change I proposed was to also adjust the quarter's start day.  The change you appear to be proposing is to also adjust the quarter's start day and month when a _Today_, _Start of this month_, or _Start of previous month_ are selected.

My remaining concern is simply the scope of the impact.   I'm not an expert on how others have set up their preferences and their expectations.  

I'll update the patch without the _Absolute_ check.

jralls: The `tm_mday` in all of those cases is `1`, which is always the start of period now. Where's the change? What am I missing?

agwekixj: > The `tm_mday` in all of those cases is `1`, which is always the start of period now. Where's the change? What am I missing?

When _Start of this month_ and _Start of previous month_ are selected the `tm_mday` will be `1`.  However, prior to this change, _Today_, _Start of this month_, or _Start of previous month_, would not impact the quarter's start month.  For example, _Start of this month_  is now 1 March 2025.  The current quarter in _stable_ would start 1 January 2025 but, after we introduce this change, the current quarter will start 1 March 2025.

Essentially, when _Today_, _Start of this month_, or _Start of previous month_ are used, the start of the quarters in the reports will become relative to the current date.  It's not bad.  It's just more than I thought we wanted to accomplish.

agwekixj:
I decided to start a fresh pull request: https://github.com/Gnucash/gnucash/pull/2065.

PR #2065, opened 2025-03-03
Use the fiscal start day of the month for relative quarter reports

agwekixj:
Only the quarter's start month in reports is adjusted when Absolute is selected in the Start Date on the Accounting Period preferences. This branch adjusts both the quarter's start day and month when Absolute is selected or when Relative and Today, Start of this month, or Start of previous month are selected.

### DIFF
--- a/libgnucash/engine/gnc-option-date.cpp
+++ b/libgnucash/engine/gnc-option-date.cpp
@@ -516,7 +516,7 @@ reldate_set_day_and_time(
 };
 
 time64
-gnc_relative_date_to_time64(RelativeDatePeriod period)
+gnc_relative_date_to_time64(RelativeDatePeriod period, time64 now_t)
 {
     if (period == RelativeDatePeriod::TODAY)
         return static_cast<time64>(GncDateTime());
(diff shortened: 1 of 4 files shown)


agwekixj:
The test errors are expected as we're changing the prior relative quarter start behavior.

jralls:
> The test errors are expected as we're changing the prior relative quarter start behavior.

In only one place, so that the reports' understanding of the relative quarter differs from the rest of the program's. This is particularly amusing because it answers my "what am I missing" from #2064: In the absence of preferences, which is the case with unit tests, the default accounting period is `TODAY`.

That aside it appears from grepping around that the only other place that would be affected by this policy change is in the price database, where leaving the last quote of each quarter is an option for deleting old quotes. That's not exactly high-risk.

However while I don't advocate as extensive a testing range as [Graham Menhemnitt](https://bugs.gnucash.org/show_bug.cgi?id=798946#c31) it does deserve more testing, not less.

agwekixj:
I have updated the expected test results for the new relative quarter start behavior.

agwekixj:
The end of a quarter should always be one day before the start of the next quarter.

jralls:
@agwekixj Don't merge stable into a PR, always rebase the PR on stable.

I did that and force-pushed.

I've also replaced your today-relative quarters hack with a fake prefs backend that sets the FY so that the original tests work correctly and leveraged it to write new tests for the fy-relative quarter values.

Thanks for working on this.

libgnucash/engine/gnc-option-date.cpp
```diff
@@ -525,10 +525,9 @@ gnc_relative_date_to_time64(RelativeDatePeriod period)
     if (period == RelativeDatePeriod::END_ACCOUNTING_PERIOD)
         return gnc_accounting_period_fiscal_end();
 
-    GncDateTime now_t;
     if (period == RelativeDatePeriod::TODAY)
```
agwekixj, line 528: This condition is already handled.

jralls: I don't understand the comment. The change you've highlighted is part of making `now_t` a parameter of the function.

agwekixj: Examine the condition handled on the first two lines of the function (521 - 522) and then on lines (528 - 529).

jralls: Got it, thanks.

PR #2070, opened 2025-03-28
[Transaction.hpp] restore trans->is_closing_cached struct member

christopherlam:
because `trans->is_closing` is heavily accessed during account balance calculations, therefore it's reasonable to cache the kvp value.

was removed in 8a7c53925870501c8f103c8f09961ecba3341c84
```
time ./gnucash-cli datafile.gnucash -R run --name "Account Summary"
```

benchmarks improve from 5.5s to 5.0s

### DIFF
--- a/libgnucash/engine/Transaction.cpp
+++ b/libgnucash/engine/Transaction.cpp
@@ -273,6 +273,7 @@ gnc_transaction_init(Transaction* trans)
     trans->marker = 0;
     trans->orig = nullptr;
     trans->txn_type = TXN_TYPE_UNCACHED;
+    trans->is_closing_cached = -1;
     LEAVE (" ");
 }
 
@@ -2129,6 +2130,7 @@ xaccTransSetIsClosingTxn (Transaction *trans, gboolean is_closing)
 {
     if (!trans) return;
     xaccTransBeginEdit(trans);
+    trans->is_closing_cached = is_closing ? 1 : 0;
 
     if (is_closing)
     {
@@ -2337,16 +2339,15 @@ xaccTransGetIsClosingTxn (const Transaction *trans)
 {
     if (!trans) return FALSE;
 
-    GValue v = G_VALUE_INIT;
-    gboolean rv;
-    qof_instance_get_kvp (QOF_INSTANCE (trans), &v, 1, trans_is_closing_str);
-    if (G_VALUE_HOLDS_INT64 (&v))
-        rv = (g_value_get_int64 (&v) ? 1 : 0);
-    else
-        rv = 0;
-    g_value_unset (&v);
-
-    return rv;
+    if (trans->is_closing_cached == -1)
+    {
+        auto t = const_cast<Transaction*>(trans);
+        GValue v = G_VALUE_INIT;
+        qof_instance_get_kvp (QOF_INSTANCE (trans), &v, 1, trans_is_closing_str);
+        t->is_closing_cached = G_VALUE_HOLDS_INT64 (&v) && g_value_get_int64 (&v) ? 1 : 0;
+        g_value_unset (&v);
+    }
+    return trans->is_closing_cached != 0;
 }
 
 /********************************************************************\
--- a/libgnucash/engine/TransactionP.hpp
+++ b/libgnucash/engine/TransactionP.hpp
@@ -116,6 +116,7 @@ struct transaction_s
      */
     char txn_type;
 
+    signed char is_closing_cached;
 };
 
 struct _TransactionClass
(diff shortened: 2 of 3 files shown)


jralls:
Where's the profiling data showing that a) there's a problem b) that caching improves it, and c) that caching is better than the
 kvp direct-access [you wrote](https://github.com/Gnucash/gnucash/commit/a26a6c4a6f9effe3dc6a42d067c9bf54b1c2c945)?

christopherlam:
> Where's the profiling data showing that a) there's a problem b) that caching improves it, and c) that caching is better than the kvp direct-access [you wrote](https://github.com/Gnucash/gnucash/commit/a26a6c4a6f9effe3dc6a42d067c9bf54b1c2c945)?

See first comment which is the only profiling I've done. 

https://github.com/Gnucash/gnucash/pull/2070#issue-2955391167

jralls:
Either I need a bigger file or you need a faster computer. ;-) With a 54MB (uncompressed) book, average of the last 4 of 5 runs each (because the first run was several hundred ms longer in both cases):
||stable: | PR2070:|
|---------|---------|--------|
|real: |1.954|1.905|
|user:| 1.684|1.631|
|system:| 0.385|0.388|

Not exactly compelling. BTW, profiling with Xcode instruments shows 75% of the time is loading the db. The report takes only 388 ms.

christopherlam:
I've done more profiling, and the modes are still 5.5s without caching, and 5.1s with caching.
Maybe let's park this. I'm still trying to optimize loading which is heavily dragged by the xaccTransBegin|CommitEdit burden.

jralls:
If you're always getting the same times it means that you're always running the same code on the same data file. That I got different results is probably down to my having a different data file and might have to do that it doesn't have any closing transactions because I never close the book in that file. Mind, even your 9% improvement isn't exactly compelling.

Running `gnucash-cli` wrapped in `time` isn't profiling. Profiling means using a profiler, and to improve user experience that has to be [timed profiling](https://www.brendangregg.com/perf.html#TimedProfiling). The Linux way is explained in excruciating detail in that link. The easy way is to get a Mac and use its free (part of Xcode) and very easy to use Instruments application. It's really easy, too: A single app that does all the work and doesn't require a special build that can mess with the compiler and linker optimizations and so yield misleading results.

What you get when you use a profiler is an extended stack trace that shows you how long the program spent in each function of each execution branch. That enables you to quickly identify problem functions--or recognize that there aren't any problem functions and that what you're asking the computer to do just takes that long. That might lead you to reconsider the design to see if you can get the same result with less work; see my comment on loading the file being the dominant time-user. IIRC it's because xaccAccountCommitEdit is sorting its splits after every transaction insertion. Sorting calls xaccTransOrder and one of the things it orders on is isClosingTrans. Some observations: 
1: When loading the splits are probably already in order. 
2. You changed the Account split container from a GList to a std::vector last year. Sorting linked lists is faster than sorting vectors. That's largely ameliorated by 1, but 
3. Even when the splits are already in order a sort is doing a pairwise comparison of every split in the list/vector. The split and transaction comparison functions are pretty expensive so it's slow.

An obvious way to speed that up is calling xaccAccountBeginEdit on every account before starting to load the transactions, but while our XML files are laid out with the transactions last--and have to be because the accounts have to all be created before the transactions can be--there's no way for a SAX parser to know that. The SQL backend knows what in order it's lloading objects so it can do those optimizations.

That aside, caching KVP  merely works around GValue's inefficiency. You've already written an actual fix that gets rid of GValue. Use it. It's a maintainability improvement that doesn't need profiling for support.

christopherlam:
From my understanding the xaccAccountBegin/CommitEdit is already happening... maybe in https://github.com/Gnucash/gnucash/blob/5bedd7fad1a7a01cb89b498d057af3656d789587/libgnucash/backend/xml/gnc-account-xml-v2.cpp#L492 and the split sorting isn't a hotspot. Whenever I've profiled loading the slowdowns have always been in xaccTransBegin/CommitEdit.

I'll see about porting the GValue->kvp to transaction.cpp functions, but I feel it's more of a code prettify than CPU optimization exercise. IIUC GValue is a simple struct with a datatype and union data see https://github.com/GNOME/glib/blob/f1a498e178709dcd5c26303a5b54ae5007733fc5/gobject/gvalue.h#L113

jralls:
> From my understanding the xaccAccountBegin/CommitEdit is already happening.

OK, that's good.

>  IIUC GValue is a simple struct with a datatype and union data

Right, a [discriminated or tagged union](https://en.wikipedia.org/wiki/Tagged_union). It still has to be initialized, have a value copied in, have its type tested, and the value copied out. `std::variant` is another example of the same with the advantage that C++ can resolve the type testing at compile time and maybe optimize out the extra copying. But using a templated retrieval function instead is faster still.

The other time waster here is resolving the KVP path. We should be able to move that to compile time with some `constexpr` pixie dust.

PR #2073, opened 2025-03-30
L10N:fi: Restore lost translations

kohtala:
Commit 63714e5e93b2091b35bbd0d28f69bce03a397d3e removed large number of translations and replaced many with invalid fuzzy translations.

This merges back translations from the parent commit tree. It also updates some translations.


kohtala:
@jralls , could you take a look at this? The translations were lost on your commit.

jralls:
@kohtala Sorry for the delay, I was traveling for 2 weeks. 

It's odd that msgmerge deleted all of those msgstrs, but all of the fuzzies I looked at seemed legit and resulted in changed translations.

PR #2074, opened 2025-04-10
Fixed Transaction.GetInvoiceFromTxn to convert to the right type of Python object

featherfeet:
GetInvoiceFromTxn method for Transaction objects in the Python bindings binds to a C function that returns a GncInvoice*, but it was attempting to return a Transaction, which causes errors. Fixed it to return a Python Invoice object.

### DIFF
--- a/bindings/python/gnucash_core.py
+++ b/bindings/python/gnucash_core.py
@@ -822,9 +822,9 @@ class Transaction(GnuCashCoreClass):
         return self.GetSplitList().pop(n)
 
     def GetInvoiceFromTxn(self):
-        from gnucash.gnucash_business import Transaction
+        from gnucash.gnucash_business import Invoice
         return self.do_lookup_create_oo_instance(
-            gncInvoiceGetInvoiceFromTxn, Transaction )
+            gncInvoiceGetInvoiceFromTxn, Invoice )
 
     def __eq__(self, other):
         return self.Equal(other, True, False, False, False)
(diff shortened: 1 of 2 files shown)


bindings/python/gnucash_core.py
```diff
@@ -822,9 +822,9 @@ def GetNthSplit(self, n):
         return self.GetSplitList().pop(n)
 
     def GetInvoiceFromTxn(self):
-        from gnucash.gnucash_business import Transaction
+        from gnucash.gnucash_business import Invoice
         return self.do_lookup_create_oo_instance(
-            gncInvoiceGetInvoiceFromTxn, Transaction )
+            gncInvoiceGetInvoiceFromTxn, Invoice )
```
jralls, line 827: I think this needs to be
```
def GetInvoiceFromTxn(self):
        from gnucash.gnucash_business import Invoice
        return self.do_lookup_create_oo_instance(
         gncInvoiceGetInvoiceFromTxn, Invoice, self )
```
The same applies to `GetInvoiceFromLot` a few lines up.

jralls (changes requested):
This also needs a unit test.

bindings/python/gnucash_core.py
featherfeet: This causes "gncInvoiceGetInvoiceFromTxn() takes 1 positional argument but 2 were given" error. My original code seems to work (I guess self gets passed implicitly). I have added a unit test in the next push.

featherfeet:
@jralls , sorry for the delay. I believe the last "self" argument is not needed. I have added a Python unit test that shows the fixed binding works.

jralls:
Thanks!

PR #2075, opened 2025-04-14
Create tax report for Canada using GIFI codes

tienne-B:
This PR implements the list of Canadian GIFI codes from [RC 4088](https://www.canada.ca/en/revenue-agency/services/forms-publications/publications/rc4088/general-index-financial-information-gifi.html) as a locale-dependent tax feature + report, mostly copying from the US equivalent.


libgnucash/tax/ca/txf.scm
```diff
@@ -0,0 +1,809 @@
...
+(export txf-help-categories)
+(export txf-income-categories)
+(export txf-liab-eq-categories)
+(export txf-tax-entity-types)
+
+(define txf-tax-entity-types
+  (list
+   (cons 'F1040 #("Individual, Joint, etc." "Files US Form 1040 Tax Return"))
+   (cons 'F1065 #("Partnership" "Files US Form 1065 Tax Return"))
+   (cons 'F1120 #("Corporation" "Files US Form 1120 Tax Return"))
```
jralls: Surely these don't apply for Canada!

jralls (commented):
This appears to be mostly a mapping of Canadian category codes to US TXF equivalents, and only for a particular flavor of Canadian business tax return. Is that right?

libgnucash/tax/ca/txf.scm
tienne-B: lol; removed.

I only know of three types of Canadian business returns (T2125 on a personal return like for sole props, T2, and T2SHORT), which uses these same codes.

I'm having some difficulty building GNUCash (which is why I kept this as a draft). Is there a Docker container that'd be standardized for builds? My Mac doesn't like the dependencies.

jralls: But what about Canadian personal returns? 

No docker containers that I know about, but there's [this tarball](https://sourceforge.net/projects/gnucash/files/Dependencies/gnucash-5.6-mac-dependencies.tar.xz/download) that we use to set things up for CI. See [macos.yaml](https://github.com/Gnucash/gnucash/blob/stable/.github/workflows/mac-tests.yaml) for how to use it. 

Otherwise you might be able to hack [gnucash-on-osx](https://github.com/gnucash/gnucash-on-osx) to build with Homebrew instead of [gtk-osx](https://gitlab.gnome.org/GNOME/gtk-osx) but you'd have to manually assemble the dependencies. The alternative is [MacPorts](https://wiki.gnucash.org/wiki/MacOS/MacPortsDetail).

tienne-B: > But what about Canadian personal returns?

I'm much less concerned about non-sole-prop personal returns (T1 without T2125), as I don't know how often the categorization would come up. I think people just auto-fill their T1s with slips provided to them, and there are not a lot of options for deductions (and even then, unsure how much people would use the field numbers)

jralls: The general point of TXF is feeding [tax software](https://www.canada.ca/en/revenue-agency/services/e-services/digital-services-individuals/netfile-overview/certified-software-netfile-program.html) instead of hand-keying it. I don't understand why that would be something that only business users would want to do.

Yes, most people with only wage income don't need tax software, but they generally don't need GnuCash either.

tienne-B: Other than T2125s, people generally don't hand-key the form fields; they get auto-populated through the slips and [Auto-fill my return](https://www.canada.ca/en/revenue-agency/services/e-services/about-auto-fill-return.html). The additional fields are very few.

jralls: OK.

jralls (changes requested):
Your cavalier treatment of copyright notices is unacceptable. Since most of this change is copying the files from `us` to `ca` and making insubstantial modifications the original copyright statements should remain on all of them.

Moving the bulk of the US taxtxf.scm to the `locale-specific` directory doesn't make sense. It's still US-specific. Leave it in `us` with a different name, maybe `taxtxf-shared.scm` and adjust the `use-module`s accordingly.

Providing an empty `txf-help-strings` isn't very helpful.

You never answered my original question, so I'll rephrase it: TXF reports are for importing into US Tax programs and are keyed to US Tax forms. The format changes every year to match the changes in those forms. Canada has completely different forms. How is this useful to Canadians?

PR #2076, opened 2025-04-15
Bug 799590 - Pressing enter after changing date on initial reconcile …

agwekixj:
…information window does not change ending balance.

If the date has changed, call the date changed callback.


jralls:
Thanks!

PR #2077, opened 2025-04-18
Stock assistant better memo defaults

christopherlam:


jralls:
"proceeds" doesn't make sense for buy or stock split transactions and is a little odd for dividends and returns of capital.

christopherlam:
@CDB-Man

CDB-Man:
Principles:
1. Have the words "gross" and "net" appear in the text
2. Have uniform text as much as possible

We could substitute the word "proceeds" with "value", which would work in all cases. "Proceeds" for dividends and ROC don't bother me at least, but I can understand why (non-accountants) would find it a bit weird.

I still also think it makes sense for the user to somehow be able to edit the defaults.

jralls:
> I still also think it makes sense for the user to somehow be able to edit the defaults.

You mean a setting/preference somewhere to have their own defaults?

CDB-Man:
https://bugs.gnucash.org/show_bug.cgi?id=798004#c94

> Suggestion: make the default texts in these fields editable / definable under Edit > Preferences menu.

PR #2078, opened 2025-04-20
[engine.i] convert gnc_account_accumulate_at_dates to c++

christopherlam:
to speed up traversal of account->splits through std::vector instead of scm_list.

ample testing in test-report-utilities.scm and test-balsheet-pnl.scm

### DIFF
--- a/bindings/engine.i
+++ b/bindings/engine.i
@@ -66,6 +66,22 @@ using AccountVec = std::vector<Account*>;
 SplitsVec gnc_get_match_commodity_splits (AccountVec accounts, bool use_end_date,
                                           time64 end_date, gnc_commodity *comm, bool sort);
 
+/* scans account splits, in posted date order, calling split_fn(split)
+   at each split and accumulate the result at each date in the SCM
+   list dates into a new SCM list. Parameters are:
+
+   acc            the account
+   dates          the SCM list of posted dates, assumed to be in chronological order
+   init           the result to be pushed into the result for dates prior to first split date
+   split_fn       the split->elt procedure whose result of (split_fn split) will be pushed
+                  into the returned SCM list */
+SCM gnc_account_accumulate_to_dates (const Account *acc, SCM dates,
+                                     SCM split_fn, SCM init);
+
+/* as above, but the split_to_date function to use a different date order. */
+SCM gnc_account_accumulate_to_dates (const Account *acc, SCM dates,
+                                     SCM split_fn, SCM init, SCM split_to_date);
+
 AccountVec gnc_accounts_and_all_descendants (AccountVec accounts);
 
 extern "C"
@@ -159,6 +175,51 @@ SplitsVec gnc_get_match_commodity_splits (AccountVec accounts, bool use_end_date
         std::sort (rv.begin(), rv.end(), [](auto a, auto b){ return xaccSplitOrder (a, b) < 0; });
     return rv;
 }
+
+static SCM
+accumulate_splits_by_dates (const SplitsVec& splits, SCM dates, SCM result,
+                            std::function<bool(Split*,SCM)> no_later_than_date,
+                            std::function<SCM(Split*)> get_result)
+{
+    SCM rv = SCM_EOL;
+    auto splits_it = splits.begin();
+    for (; !scm_is_null(dates); dates = scm_cdr (dates))
+    {
+        while (splits_it != splits.end() && no_later_than_date (*splits_it, scm_car (dates)))
+            result = get_result (*splits_it++);
+
+        rv = scm_cons (result, rv);
+    }
+    return scm_reverse_x (rv, SCM_EOL);
+}
+
+SCM
+gnc_account_accumulate_to_dates (const Account *acc, SCM dates,
+                                 SCM split_fn, SCM init)
+{
+    const auto& splits = xaccAccountGetSplits(acc);
+    auto get_result = [&](Split* s) -> SCM { return scm_call_1(split_fn, gnc_split_to_scm(s)); };
+    auto no_later_than_date = [&](Split* s, SCM date) -> bool
+        { return xaccTransGetDate (xaccSplitGetParent (s)) <= scm_to_int64 (date); };
+
+    return accumulate_splits_by_dates (splits, dates, init, no_later_than_date, get_result);
+}
+
+SCM
+gnc_account_accumulate_to_dates (const Account *acc, SCM dates,
+                                 SCM split_fn, SCM init, SCM split_to_date)
+{
+    auto splits = xaccAccountGetSplits(acc);
+    auto less_scm = [](SCM a, SCM b) -> bool { return scm_is_true(scm_less_p(a, b)); };
+    auto get_date = [&](Split* s) -> SCM { return scm_call_1(split_to_date, gnc_split_to_scm(s)); };
+    auto get_result = [&](Split* s) -> SCM { return scm_call_1(split_fn, gnc_split_to_scm(s)); };
+    auto no_later_than_date = [&](auto s, SCM date) -> bool { return !less_scm(date, get_date(s)); };
+    std::sort(splits.begin(), splits.end(), [&](auto a, auto b) -> bool
+              { return less_scm(get_date(a), get_date(b)); });
+
+    return accumulate_splits_by_dates (splits, dates, init, no_later_than_date, get_result);
+}
+
 
 using AccountSet = std::unordered_set<Account*>;
 static void maybe_add_descendants (Account* acc, AccountSet* accset)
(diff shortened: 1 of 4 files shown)


christopherlam:
ready...

bindings/engine.i
```diff
@@ -160,6 +166,50 @@ SplitsVec gnc_get_match_commodity_splits (AccountVec accounts, bool use_end_date
     return rv;
 }
 
+static SCM
+accumulate_splits_by_dates (const SplitsVec& splits, SCM dates, SCM last_result,
+                            std::function<bool(Split*,SCM)> no_later_than_date,
+                            std::function<SCM(Split*)> get_result)
+{
+    SCM rv = SCM_EOL;
+    for (auto splits_it = splits.begin(); !scm_is_null(dates); dates = scm_cdr (dates))
```
jralls: That's really obtuse. One expects a for loop to iterate on the variable in the first clause. 
```
auto splits_it{splits.begin()};
for (; !scm_is_null(dates); dates = scm_cdr (dates))
{
```
is more idiomatic.

jralls (commented):
Maybe a little too clever. Some comments on the engine.1 functions explaining the SCM parameters would be a good addition.

christopherlam:
even better I'm keen to remove the `split_to_date` parameter... I never ended up using it.

bindings/engine.i
```diff
@@ -66,6 +66,22 @@ using AccountVec = std::vector<Account*>;
...
 
+/* scans account splits, in posted date order, accumulating the result
+   of split_to_elt(split) at each date in the SCM list dates into a
+   new SCM list. Parameters are:
+
+   acc - the account
+   dates - the SCM list of posted dates
+   last_result - the result to be pushed into the result for dates prior to first split date
+   split_to_elt - the split->elt transformer to be pushed into the result */
+SCM gnc_account_accumulate_at_dates (const Account *acc, SCM dates, SCM last_result,
```
jralls: Please explain what an 'elt' is.

jralls: It occurs to me that there's nothing in *this* code restricting what function gets passed in `split_to_elt`, so it should be renamed `transform_fn` or `transformer` or perhaps `unary_op` to borrow the equivalent parameter name from `std::transform`. That leaves the mysterious `elt` term in report-utilities.scm where there's presumably enough context to figure out what an `elt` is.

Along those lines, this is a transform, not an accumulate: std::accumulate returns a single value and this returns a range of transformed values.

christopherlam: > It occurs to me that there's nothing in _this_ code restricting what function gets passed in `split_to_elt`, so it should be renamed `transform_fn` or `transformer` or perhaps `unary_op` to borrow the equivalent parameter name from `std::transform`. That leaves the mysterious `elt` term in report-utilities.scm where there's presumably enough context to figure out what an `elt` is.
> 
> Along those lines, this is a transform, not an accumulate: std::accumulate returns a single value and this returns a range of transformed values.

It's both a transform (returns a list of values) and an accumulator (calls fn on each split up to date). I think `split->elt` is appropriate to distinguish from `split->date` and `elt` refers to a scheme list element.

jralls: > It's both a transform (returns a list of values) and an accumulator (calls fn on each split up to date).

No, both accumulate (aka [fold](https://en.wikipedia.org/wiki/Fold_(higher-order_function)) and transform (aka [map](https://en.wikipedia.org/wiki/Map_(higher-order_function) apply an operation to each member of a collection. The difference is that transform returns a new collection and accumulate returns *a single value*.

>  I think split->elt is appropriate to distinguish from split->date and elt refers to a scheme list element

`split->date` is kind of appropriate because in C++ it means "get `split`'s `date` member. The fact that in Scheme it implies "convert the split into a date", which isn't possilbe, is irrelevant here because this is C++ code. The function hard-codes that parameter to returning a date so naming the parameter "function to get a date from the split" is reasonable.

But the other operation is arbitrary. The *current scheme-side usage* happens to need a list element but *this* function will work for any operation that takes only a split as an argument. Naming the operation parameter too specifically makes it harder to recognize that.

jralls:
> even better I'm keen to remove the split_to_date parameter... I never ended up using it.

Plus it won't work with anything besides transaction date because it depends on the the split vector being ordered on the desired date field and while that's not necessarily true for the other dates.

christopherlam:
> > even better I'm keen to remove the split_to_date parameter... I never ended up using it.
> 
> Plus it won't work with anything besides transaction date because it depends on the the split vector being ordered on the desired date field and while that's not necessarily true for the other dates.

It does work with split->reconciled_date see test-report-utilities.scm -- splitlist is sorted via std::sort

jralls:
> splitlist is sorted via std::sort

Ah, missed that. That's bad because it's a side effect on the vector that will affect other uses of the vector unless they re-sort it. It's also a potential concurrency problem if updating the progress bar runs an event that causes an account to sort its splits. Maybe better to copy the split vector and sort that.

Your rename (though impressively ugly) helped my understand what you were trying to tell me with "it's both a transform and an accumulate": It's a transform on the date list where the function is an accumulate on the split vector truncated by the current date list member. I'll have to think a bit about how to name that.

christopherlam:
> > splitlist is sorted via std::sort
> 
> Ah, missed that. That's bad because it's a side effect on the vector that will affect other uses of the vector unless they re-sort it. It's also a potential concurrency problem if updating the progress bar runs an event that causes an account to sort its splits. Maybe better to copy the split vector and sort that.

It's being copied already!

jralls:
> It's being copied already!

How's that? You get a `SplitVec& splits` and call `std::sort(splits.begin(), splits.end(), ...)`. I see no copying here.

bindings/engine.i
```diff
@@ -160,6 +177,51 @@ SplitsVec gnc_get_match_commodity_splits (AccountVec accounts, bool use_end_date
...
+        rv = scm_cons (last_result, rv);
+    }
+    return scm_reverse_x (rv, SCM_EOL);
+}
+
+SCM
+gnc_account_foreach_split_transform_dates (const Account *acc, SCM dates, SCM last_result,
+                                 SCM split_fn)
+{
+    const auto& splits = xaccAccountGetSplits(acc);
```
christopherlam, line 200: no copy needed here

bindings/engine.i
```diff
@@ -160,6 +177,51 @@ SplitsVec gnc_get_match_commodity_splits (AccountVec accounts, bool use_end_date
...
+        { return xaccTransGetDate (xaccSplitGetParent (s)) <= scm_to_int64 (date); };
+
+    return accumulate_splits_by_dates (splits, dates, last_result, no_later_than_date, get_result);
+}
+
+SCM
+gnc_account_foreach_split_transform_dates (const Account *acc, SCM dates, SCM last_result,
+                                           SCM split_fn, SCM split_to_date)
+{
+    auto splits = xaccAccountGetSplits(acc);
```
christopherlam, line 212: copy needed here to sort immediately before calling helper fn

jralls: Right. The easiest way is to change this line to
```
std::vector<Split*> splits{xaccAccountGetSplits(acc)};
```

christopherlam: from my understanding `auto splits = xaccAccountGetSplits(acc);` (cf `auto& splits = xaccAccountGetSplits(acc);`) with the `&` will already create a vector copy but I've changed to explicitly specify type.

jralls: No, `auto` will make splits whatever the initializer returns, [in this case a `const SplitVec&`](https://github.com/Gnucash/gnucash/blob/e81d2846192a368096d00e645d5cad1fffc88ced/libgnucash/engine/Account.hpp#L43).

christopherlam: Ok if it's a const then the std::sort shouldn't compile?

jralls: Hmm, yes, it should fail. So I went [to play](https://godbolt.org/z/hEzr7rhzd), changing  You're right, `auto` by itself makes a copy no matter what the qualifiers on the function return value. Since it's a copy `const`ness is discarded. `auto&` will preserve `const`ness if that's what the function returns. Trying to `auto&` a non-reference return value barfs because you can't take a reference to an rvalue.

Meaning I've missed a subtlety on returning references for 13+ years: It's not sufficient to return a reference from a function if you want to avoid a copy, you also have to explicitly say that you want a reference at the receiving end. There's probably a lot of places where I've gotten that wrong.

jralls (approved):

christopherlam:
how about rename to `gnc_account_accumulate_splits_to_dates`

christopherlam:
I've benchmarked and this C++ version (balsheet-multicol = 0.30s) is actually slower than the guile one (balsheet-multicol = 0.20s)... this is because of the overhead of `gnc_split_to_scm` which calls `SWIG_TypeQuery ("_p_Split")` for each split. I've added another commit which caches these `swig_type_info*` and improves performance.

jralls:
> I've added another commit which caches these swig_type_info* and improves performance.

I noticed that, but your comment led me to go look at why it's so slow. It's because the types are stored as two circular linked lists of strings, one for mangled type names and the other for non-mangled ones. Every call to TypeQuery loops through both of them. It screams for a hash table.

PR #2082, opened 2025-05-05
Guard against duplicate definitions of __EXTENSIONS__

Bill-Sommerfeld:
For whatever reason, g++ on illumos predefines `__EXTENSIONS__`.

gnucash defines it in four places; it's guarded with an `#ifdef` in `config.h`, but three other files define it unconditionally in a way that clashes with the predefinition, resulting in compile errors like:

```
/z/ws/gnucash-upstream/libgnucash/engine/gnc-date.cpp:28:9: error: "__EXTENSIONS__" redefined [-Werror]
   28 | #define __EXTENSIONS__
      |         ^~~~~~~~~~~~~~
<built-in>: note: this is the location of the previous definition
```
The fix is to change the other three files to match `config.h`.


jralls:
Meh. `__EXTENSIONS__` seems to be a SunOS/Solaris thing. Let's just get rid of them.

jralls:
I just saw your wiki account application explaining that openindiana is an open source solaris. The reason for `__EXTENSIONS__` is explained in [standards(5)](https://docs.oracle.com/cd/E26502_01/html/E29043/standards-5.html) under "feature test macros".

Bill-Sommerfeld:
I understand why `__EXTENSIONS__` exists in the system headers (looks like it was also used on hp-ux).  What I'm not sure of yet is why g++ (but not gcc) is always setting it.

Dropping the #define `__EXTENSIONS__` from these three files and from config.h makes sense if it can land in `CFLAGS` instead.  I'll revise this pull request accordingly..

jralls:
It will also need to come out of CMakeLists.txt.

Test thoroughly, it's not going to have anything to do with modern Unix that I can test.

jralls:
>  I'll revise this pull request accordingly..

Ping?

Bill-Sommerfeld:
Will be getting back to this shortly (had it mostly ready when something else captured my attention..)

jralls:
Thanks!

PR #2083, opened 2025-05-05
Add testbuild target which builds tests without running them

Bill-Sommerfeld:
The openindiana packaging framework includes cmake integration that lets you sensibly compare a test run with prior test results checked into the packaging repo.  It works better if you can just let the system run `ctest` separately so the output isn't cluttered with build messages.

The only way I've found to force a build of all tests is through the `check` target, which also runs the tests once they're all built.  

This change adds a new build target, `testbuild`, which builds tests without running them.
(I'm new to using cmake and just picked a plausible name; I'm open to changing it if anyone's got a better idea).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -693,7 +693,8 @@ endif()
 gnc_gtest_configure()
 
 # There are targets that need to build before tests will run
-add_custom_target(check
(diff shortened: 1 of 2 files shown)


jralls:
Great idea, thanks!

PR #2085, opened 2025-05-11
Bug 799602 - Unit Price on Invoice will NOT display as decimals

jralls:
Add a new form gnc:price->string to complement gnc:monetary->string that uses the gnc-price-print-info and so observes the price-force-decimal preference. Add an eguile form fmtprice that wraps it and replace fmtmoney with fmtprice when printing the unit price in the tas invoice report.

### DIFF
--- a/gnucash/report/reports/support/taxinvoice.eguile.scm
+++ b/gnucash/report/reports/support/taxinvoice.eguile.scm
@@ -331,7 +331,7 @@
         <td align="right"><?scm:d (fmtnumeric qty) ?></td>
       <?scm )) ?>
       <?scm (if (or units? qty? discount?) (begin ?>
-        <td align="right"><?scm:d (fmtmoney currency each) ?></td>
+        <td align="right"><?scm:d (nbsp (gnc:default-price-renderer currency each)) ?></td>
       <?scm )) ?>
       <?scm (if discount? (begin ?>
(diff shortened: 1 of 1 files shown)


gnucash/report/reports/support/taxinvoice.eguile.scm
```diff
@@ -331,7 +331,7 @@
         <td align="right"><?scm:d (fmtnumeric qty) ?></td>
       <?scm )) ?>
       <?scm (if (or units? qty? discount?) (begin ?>
-        <td align="right"><?scm:d (fmtmoney currency each) ?></td>
+        <td align="right"><?scm:d (fmtprice currency each) ?></td>
```
christopherlam: There's already `gnc:default-price-renderer`:

```suggestion
        <td align="right"><?scm:d (gnc:default-price-renderer currency each) ?></td>
```

jralls: OK. Using that suggests that I should also replace `fmtmoney` with `gnc:default-html-monetary-renderer` and maybe do the same for `gnc:monetary->string`.

jralls: > `fmtmoney`…

Nah. `gnc:default-html-monetary-renderer` seems excessively convoluted.

christopherlam:
Already merged in

PR #2086, opened 2025-05-12
Avoid non-POSIX "echo -n" behavior in util/gnc-vcs-info

Bill-Sommerfeld:
POSIX declined to standardize -n and recommends against using it. Instead, for -r, do a single echo with both the commithash and the modified flag.

Tested on illumos, MacOS, and linux.


jralls:
Thanks!

PR #2087, opened 2025-05-12
Let cmake probe for the existence of HAVE_STRUCT_TM_GMTOFF

Bill-Sommerfeld:
Use cmake facilities to probe for the existence of the tm_gmtoff field of struct tm, and set HAVE_STRUCT_TM_GMTOFF based on the result.

Tested on illumos and linux.


jralls:
Thanks!

PR #2090, opened 2025-05-24
Export qof_book_mark_session_dirty to the Python bindings.

rotdrop:
Use case: implement a command line "save-as" script. This works by using the gnucash_core_c bindings like follows but for the problem that qof_book_mark_session_dirty() is not exported, and without "dirty" the function `qof_session_save(outputSession)` does nothing.

```
...
inputSession = gnucash_core_c.qof_session_new(gnucash_core_c.qof_book_new())
gnucash_core_c.qof_session_begin(inputSession, inputURI, SessionOpenMode.SESSION_READ_ONLY)
gnucash_core_c.qof_session_load(inputSession, None)

outputSession = gnucash_core_c.qof_session_new(gnucash_core_c.qof_book_new())
gnucash_core_c.qof_session_begin(outputSession, outputURI, SessionOpenMode.SESSION_NEW_OVERWRITE)

gnucash_core_c.qof_session_swap_data(inputSession, outputSession)
gnucash_core_c.qof_book_mark_session_dirty(gnucash_core_c.qof_session_get_book(outputSession))

gnucash_core_c.qof_session_save(outputSession, None)
gnucash_core_c.qof_session_end(outputSession)
```

Complete script is here https://github.com/rotdrop/gnucash-python/blob/main/src/save-as.py

PR #2091, opened 2025-05-24
Expose the percentage-function to the Python-bindings by a typemap

rotdrop:
The idea is to be able to use the percentage function with a python function callback. Maybe a little bit hacky as SWIG was not designed for these kind of things (i.e. callbacks from the target language back to Python are difficult to achieve with SWIG).

Usage:
```
...
percentageCallback =  gnucash_core_c.qof_percentage_func(lambda a, b: print(a, b) if a else print(b))
...
gnucash_core_c.qof_session_load(inputSession, percentageFunc)
...
```
Complete "save-as" script is here https://github.com/rotdrop/gnucash-python/blob/main/src/save-as.py

### DIFF
--- a/bindings/python/gnucash_core.i
+++ b/bindings/python/gnucash_core.i
@@ -44,6 +44,29 @@
 %feature("autodoc", "1");
 %module(package="gnucash") gnucash_core_c
 
+%typemap(in) QofPercentageFunc {
+  if ($input == Py_None) {
+    $1 = (QofPercentageFunc)0;
+  } else {
+    $1 = (QofPercentageFunc)PyLong_AsVoidPtr($input);
+  }
+}
+
+%pythoncode
+%{
+import ctypes
+
+qof_percentage_func_type = ctypes.CFUNCTYPE(None, ctypes.c_char_p, ctypes.c_double)
+
+def qof_percentage_func(py_callback):
+    if py_callback == None:
+        f_ptr = None
+    else:
+        f = qof_percentage_func_type(py_callback)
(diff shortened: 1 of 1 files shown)


bindings/python/gnucash_core.i
```diff
@@ -44,6 +44,29 @@
...
+  } else {
+    $1 = (QofPercentageFunc)PyLong_AsVoidPtr($input);
+  }
+}
+
+%pythoncode
+%{
+import ctypes
+
+qof_percentage_func_type = ctypes.CFUNCTYPE(None, ctypes.c_char_p, ctypes.c_double)
```
jralls: IIUC this is a type so according to our [style guide](https://wiki.gnucash.org/wiki/CodingStandard) it should be camel-case.

jralls (changes requested):
Very clever.

Needs a test to go with it.

bindings/python/gnucash_core.i
rotdrop: Yes, this is a type. Thanks for pointing me to the style guide.

rotdrop:
> Very clever.
> 
> Needs a test to go with it.

The credits go actually to

https://gist.github.com/sterin/32b5c7c42a177dacf455

Test would also need to be written for my other PR #2090, I'll try to have a look at it ...

bindings/python/gnucash_core.i
rotdrop: Should be cured by e7e3b55358a663913e7989bd7569b469f92b648d

jralls:
@rotdrop Still waiting for tests...

PR #2092, opened 2025-05-24
wip: swigify gnc_options_copy_values

christopherlam:
I think this function is a significant slowdown in the multicolumn profit&loss -- because each cell (income or expense in period) generates a link to a transaction report with the particular exact splits, and each link creates a whole new optiondb copied from the transaction report. However I'm not 100% sure how to complete this.

### DIFF
--- a/bindings/guile/gnc-optiondb.i
+++ b/bindings/guile/gnc-optiondb.i
@@ -2139,6 +2139,26 @@ gnc_register_multichoice_callback_option(GncOptionDBPtr& db,
                     });
             });
     }
+
+
+    void
+    gnc_options_copy_values (GncOptionDBPtr& src, GncOptionDBPtr& dest)
+    {
+        if (!dest) return;
+        auto make_copy = [&](auto src_option)
+        {
+            const auto& section = GncOption_get_section (src_option);
+            const auto& name = GncOption_get_name (src_option);
+            auto dest_option = gnc_lookup_option (dest, section, name);
+            if (dest_option)
+                GncOption_set_value (dest_option, GncOption_get_value (src_option));
+        };
+        src->foreach_section([&](const GncOptionSectionPtr& section)
+                             {
+                                 section->foreach_option([&](auto& option)
+                                                         { make_copy (&option); });
+                             });
+    }
 %}
 
 #endif //SWIGGUILE
--- a/bindings/guile/gnc-optiondb.i
+++ b/bindings/guile/gnc-optiondb.i
@@ -2140,24 +2140,33 @@ gnc_register_multichoice_callback_option(GncOptionDBPtr& db,
             });
     }
 
-
-    void
+    static void
     gnc_options_copy_values (GncOptionDBPtr& src, GncOptionDBPtr& dest)
     {
         if (!dest) return;
-        auto make_copy = [&](auto src_option)
+        auto make_copy = [&](const auto& src_option)
         {
-            const auto& section = GncOption_get_section (src_option);
-            const auto& name = GncOption_get_name (src_option);
-            auto dest_option = gnc_lookup_option (dest, section, name);
-            if (dest_option)
-                GncOption_set_value (dest_option, GncOption_get_value (src_option));
+            const auto& section = src_option.get_section();
+            const auto& name = src_option.get_name().c_str();
+            auto dest_option = dest->find_option(section, name);
+            /* std::cout << section << '/' << name << ": " << dest_option << '\n'; */
+            if (!dest_option) return;
+            switch (src_option.get_ui_type())
+            {
+            case GncOptionUIType::BOOLEAN:
+                dest_option->set_value (src_option.template get_value<bool>());
+                break;
+            case GncOptionUIType::TEXT:
+                dest_option->set_value (src_option.template get_value<std::string>());
+                break;
+            case GncOptionUIType::ACCOUNT_LIST:
+                dest_option->set_value (src_option.template get_value<std::string>());
+                break;
+            default:
+                break;
+            };
         };
-        src->foreach_section([&](const GncOptionSectionPtr& section)
-                             {
-                                 section->foreach_option([&](auto& option)
-                                                         { make_copy (&option); });
-                             });
+        src->foreach_section([&](auto& src_section){ src_section->foreach_option (make_copy); });
     }
 %}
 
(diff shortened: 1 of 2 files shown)


christopherlam:
@jralls any clues? Is it possible to use a higher level than bool/int/string etc?
```cpp
    dest_option->set_value (src_option.template get_value<GncOptionValue>())
```

jralls:
@christopherlam To answer the question directly, yes: Expose the GncOptionImpl copy ctor to GncOption and use that.

But is there an opportunity for the transaction reports' options to mutate independently? If not they can just take a reference to the optiondb instead.

Are the transaction reports used for anything if the user doesn't follow the link? If not they shouldn't be created unless they're actually called for.

christopherlam:
> Are the transaction reports used for anything if the user doesn't follow the link? If not they shouldn't be created unless they're actually called for.

Correct the SCM transaction reports options are being created for each cell. Without them the report renders in 0.5s instead of several seconds. I don't know how to delay creating the options db.

jralls:
>  I don't know how to delay creating the options db.

Have `gnc:make-report-anchor` make a closure having the data needed to generate the report and return a URL that fires a callback to actually create the report using the closure contents. A lambda with captures (which in C++ is syntactic sugar for a function object) is an example of a closure. GObject signal handlers is another. I don't know offhand how to make one in Scheme.

christopherlam:
> > I don't know how to delay creating the options db.
> 
> Have `gnc:make-report-anchor` make a closure having the data needed to generate the report and return a URL that fires a callback to actually create the report using the closure contents. A lambda with captures (which in C++ is syntactic sugar for a function object) is an example of a closure. GObject signal handlers is another. I don't know offhand how to make one in Scheme.

Ok here's how I've structured a "delayed report" scenario. It wastes memory because the hash tables to store the environment doesn't clear. It also doesn't yet work.

```diff
diff --git a/gnucash/gnome/gnc-plugin-page-report.cpp b/gnucash/gnome/gnc-plugin-page-report.cpp
index a700a233c2..f63db3738a 100644
--- a/gnucash/gnome/gnc-plugin-page-report.cpp
+++ b/gnucash/gnome/gnc-plugin-page-report.cpp
@@ -677,6 +677,19 @@ gnc_plugin_page_report_load_cb (GncHtml * html, URLType type,
         report_id = atoi(location + 3);
         DEBUG( "parsed id=%d", report_id );
     }
+    if (!g_strcmp0 (type, URL_TYPE_REPORT)
+        && location
+        && !strncmp("closure-id=", location, 11))
+    {
+        auto closure_id = atoi (location + 11);
+        const SCM trigger_cb = scm_c_eval_string ("gnc:trigger-delayed-report");
+        // printf ("trigger_cb = %p. closure_id = %d\n", trigger_cb, closure_id);
+        // scm_write (trigger_cb, scm_current_output_port ());
+        // printf ("\n");
+        auto rv = scm_call_1 (trigger_cb, scm_from_uint (closure_id));
+        report_id = scm_to_uint (rv);
+        DEBUG( "parsed id=%d", report_id );
+    }
     else if (!g_strcmp0( type, URL_TYPE_OPTIONS)
              && location
              && (strlen(location) > 10)
diff --git a/gnucash/report/html-utilities.scm b/gnucash/report/html-utilities.scm
index 904184a4bd..cf93d65bf7 100644
--- a/gnucash/report/html-utilities.scm
+++ b/gnucash/report/html-utilities.scm
@@ -180,6 +180,34 @@
             (gnc:report-anchor-text id)))
         (warn "gnc:make-report-anchor: No such report: " reportname))))
 
+(define closures (make-hash-table))
+(define size 0)
+
+(use-modules (srfi srfi-69))
+
+(define-public (gnc:make-delayed-report-anchor reportname src-report optionlist)
+  (set! size (1+ size))
+  (hash-set! closures size (list reportname src-report optionlist))
+  (gnc-build-url URL-TYPE-REPORT (format #f "closure-id=~a" size) ""))
+
+(define-public (gnc:trigger-delayed-report id)
+  ;; (pk 'gnc:trigger-delayed-report id)
+  (match (hash-ref closures id)
+    ((reportname src-report optionlist)
+     ;; (pk 'gnc:trigger-delayed-report reportname src-report optionlist)
+     (let* ((src-options (gnc:report-options src-report))
+            (options (gnc:make-report-options reportname))
+            (db (gnc:optiondb options)))
+       ;; (pk '(reportname src-report optionlist) (list reportname src-report optionlist))
+       ;; (pk 'src-options src-options)
+       ;; (pk 'options:before options)
+       ;; (pk 'db db)
+       (gnc:options-copy-values src-options options)
+       ;; (pk 'options:after options)
+       (for-each (lambda (l) (gnc-set-option db (car l) (cadr l) (caddr l))) optionlist)
+       ;; (pk 'options:after-setting options)
+       (pk 'rv (gnc:make-report reportname options))))
+    (_ (gnc:error "invalid id " id))))
 
 ;; returns the account name as html-text and anchor to the register.
 (define (gnc:html-account-anchor acct)
diff --git a/gnucash/report/reports/standard/balsheet-pnl.scm b/gnucash/report/reports/standard/balsheet-pnl.scm
index da80ad300e..1c581988db 100644
--- a/gnucash/report/reports/standard/balsheet-pnl.scm
+++ b/gnucash/report/reports/standard/balsheet-pnl.scm
@@ -1122,7 +1122,7 @@ also show overall period profit & loss."))
                       (show-orig? (and common-currency #t))
                       (curr (or common-currency book-main-currency))
                       (accts (if (pair? account) account (list account))))
-                  (gnc:make-report-anchor
+                  (gnc:make-delayed-report-anchor
                    trep-uuid report-obj
                    (list
                     (list "General" "Start Date" (cons 'absolute (car datepair)))
```

jralls:
That's not a [closure](https://en.wikipedia.org/wiki/Closure_(computer_programming)). Apparently in [Scheme](https://docs.scheme.org/schintro/schintro_toc.html#SEC152) `(let ((foo)) (lambda() …))` is the same as `[&foo]() {…};`.

So you want a `gnc:make-subreport-generator` that takes the report Guid, optiondbref, and option list and returns a procedure/closure with those values embedded. You park it wherever you're parking the created reports and `gnc_main_window_open_report_url` or `gnc_plugin_page_report_new` retrieves the procedure and runs it to generate the report.

PR #2094, opened 2025-05-27
Bug 799363 - When selecting "load" two layer ...

Oscar65:
The Back and Forward buttons now work correctly.
Added pressedBACK and pressedFWD to _gnc_html_history because gnc_html_history_append needs to know if a button has been pressed.

I had to remove the Back and Forward options from the right-click menu because they interfere with the buttons' functionality.

Clicking and selecting Load clears the existing Forward order. I can't change it because it's JavaScript inside the HTML.

### DIFF
--- a/gnucash/html/gnc-html-webkit2.c
+++ b/gnucash/html/gnc-html-webkit2.c
@@ -120,6 +120,11 @@ static void impl_webkit_print (GncHtml* self,const gchar* jobname);
 static void impl_webkit_cancel( GncHtml* self );
 static void impl_webkit_set_parent( GncHtml* self, GtkWindow* parent );
 static void impl_webkit_default_zoom_changed(gpointer prefs, gchar *pref, gpointer user_data);
+static gboolean webkit_context_menu_cb(WebKitWebView         *web_view,
+                                       WebKitContextMenu     *context_menu,
+                                       GdkEvent              *event,
+                                       WebKitHitTestResult   *hit_test_result,
+                                       gpointer               user_data);
 
 static GtkWidget*
 gnc_html_webkit_webview_new (void)
@@ -206,6 +211,10 @@ gnc_html_webkit_init( GncHtmlWebkit* self )
      g_signal_connect (priv->web_view, "resource-load-started",
                        G_CALLBACK (webkit_resource_load_started_cb),
                        self);
+     g_signal_connect(priv->web_view, "context-menu",
+                       G_CALLBACK(webkit_context_menu_cb),
+                       self);
+
      gnc_prefs_register_cb (GNC_PREFS_GROUP_GENERAL_REPORT,
                             GNC_PREF_RPT_DFLT_ZOOM,
                             impl_webkit_default_zoom_changed,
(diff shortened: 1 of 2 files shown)


gnucash/html/gnc-html-webkit2.c
```diff
@@ -727,6 +736,56 @@ webkit_resource_load_started_cb (WebKitWebView *web_view,
...
+          {
+                webkit_context_menu_remove(context_menu, item);
+                // Because GList has changed we cannot remove another item.
+                break;
+          } else if (action == WEBKIT_CONTEXT_MENU_ACTION_GO_FORWARD) {
+                webkit_context_menu_remove(context_menu, item);
+                // Because GList has changed we cannot remove another item.
+                break;
+          }
+     }
```
jralls, line 784: Don't duplicate code. You *could* extract this to a separate function that takes the WebKitContextMenuAction as an argument and call it twice.

But you don't need to do either. You can keep on working with the list, you just need to advance the node before you delete it:
```
for (GList* node = menu_items; node;)
{
    WebKitContextMenuItem *item = WEBKIT_CONTEXT_MENU_ITEM (node->data);
    WebKitContextMenuAction action = webkit_context_menu_item_get_stock_action (item);
    node = g_list_next (node);
    if (action == WEBKIT_CONTEXT_MENU_ACTION_GO_BACK ||
        action == WEBKIT_CONTEXT_MENU_ACTION_GO_FORWARD)
        webkit_context_menu_remove(context_menu, item);
}
```

gnucash/html/gnc-html-webkit2.c
```diff
@@ -727,6 +736,56 @@ webkit_resource_load_started_cb (WebKitWebView *web_view,
...
+
+     // We need to get items again.
+     menu_items = webkit_context_menu_get_items(context_menu);
+     for (GList *l = menu_items; l != NULL; l = g_list_next(l)) {
```
jralls, line 771: Avoid single-letter names except for integer indexes. `node` is a typical name for list members.
Always put curly braces on their own line. [GnuCash's style](https://wiki.gnucash.org/wiki/CodingStandard) specifies that they should be flush with the introducing line.

gnucash/html/gnc-html-webkit2.c
```diff
@@ -727,6 +736,56 @@ webkit_resource_load_started_cb (WebKitWebView *web_view,
                        data);
 }
 
+/*
+ *
+ * We need to delete the Back and Forward items from the right-click menu because
+ * otherwise the Back and Forward buttons won't work properly.
```
jralls, line 742: That makes the context menu different from those of most browsers. Better to replace the menu items with new ones that use the same actions as the toolbar buttons.

jralls (changes requested):
Everything you do in gnc-html-webkit2.c you must also do in gnc-html-webkit1.c.

Oscar65:
I managed to create "Back" and "Forward" items in the context menu, but I don't know how to call gnc-plugin-page-report.c::gnc_plugin_page_report_back_cb from a function in gnc-html-webkit2.c with report as a parameter.

Is possible?

jralls:
> Is possible?

Yes, via the navigation policy in `perform_navigation_policy` in gnc-html-webkit2.c and `webkit_navigation_requested_cb` in gnc-html-webkit1.c.

Oscar65:
@jralls Could you please give me an example?, because I don't find any examples on code nor documentation about `perform_navigation_policy`

jralls:
@Oscar65 

> because I don't find any examples on code nor documentation about perform_navigation_policy

[WebkitGtk navigation policy documentation](https://webkitgtk.org/reference/webkit2gtk/2.25.4/WebKitNavigationPolicyDecision.html) and [`perform_navigation_polcy`](https://github.com/Gnucash/gnucash/blob/stable/gnucash/html/gnc-html-webkit2.c#L599) *is* the example as well as the obvious place for you to hook in your handler.

Oscar65:
> Everything you do in gnc-html-webkit2.c you must also do in gnc-html-webkit1.c.

@jralls I did in gnc-html-webkit2.c, how can I build with webkit1 to develop and test it?

jralls:
You'll need to set up a [Windows build environment](https://wiki.gnucash.org/wiki/Building_on_Windows). @Bob-IT provided some more advice in [this thread on gnucash-devel](https://lists.gnucash.org/pipermail/gnucash-devel/2025-March/047024.html).

PR #2095, opened 2025-05-28
Bug 799538 - Clipboard pastes to wrong field in...

Oscar65:
Fixed Cut, Copy and Paste when focus is not in GnucashSheet.

### DIFF
--- a/gnucash/gnome/gnc-plugin-page-register.cpp
+++ b/gnucash/gnome/gnc-plugin-page-register.cpp
@@ -3654,6 +3654,17 @@ gnc_plugin_page_register_cmd_cut (GSimpleAction *simple,
 
     ENTER ("(action %p, page %p)", simple, page);
     priv = GNC_PLUGIN_PAGE_REGISTER_GET_PRIVATE (page);
+
+    GtkWidget *widget = gtk_window_get_focus(GTK_WINDOW (priv->gsr->window));
+    const char *name = gtk_widget_get_name(widget);
+    if (strcmp(name, "GnucashSheet") != 0)
+    {
+        gtk_editable_cut_clipboard( GTK_EDITABLE(widget));
+        LEAVE("Not cut from GnucashSheet");
+
+        return;
+    }
+
     gnucash_register_cut_clipboard (priv->gsr->reg);
     LEAVE ("");
 }
@@ -3671,6 +3682,17 @@ gnc_plugin_page_register_cmd_copy (GSimpleAction *simple,
 
     ENTER ("(action %p, page %p)", simple, page);
     priv = GNC_PLUGIN_PAGE_REGISTER_GET_PRIVATE (page);
+
+    GtkWidget *widget = gtk_window_get_focus(GTK_WINDOW (priv->gsr->window));
+    const char *name = gtk_widget_get_name(widget);
+    if (strcmp(name, "GnucashSheet") != 0)
+    {
+        gtk_editable_copy_clipboard( GTK_EDITABLE(widget));
+        LEAVE("Not copied from GnucashSheet");
+
+        return;
+    }
+
     gnucash_register_copy_clipboard (priv->gsr->reg);
     LEAVE ("");
 }
@@ -3688,6 +3710,17 @@ gnc_plugin_page_register_cmd_paste (GSimpleAction *simple,
 
     ENTER ("(action %p, page %p)", simple, page);
     priv = GNC_PLUGIN_PAGE_REGISTER_GET_PRIVATE (page);
+
+    GtkWidget *widget = gtk_window_get_focus(GTK_WINDOW (priv->gsr->window));
+    const char *name = gtk_widget_get_name(widget);
+    if (strcmp(name, "GnucashSheet") != 0)
+    {
+        gtk_editable_paste_clipboard( GTK_EDITABLE(widget));
+        LEAVE("Not pasted to GnucashSheet");
+
+        return;
+    }
+
     gnucash_register_paste_clipboard (priv->gsr->reg);
     LEAVE ("");
 }
--- a/gnucash/gnome/dialog-sx-editor.c
+++ b/gnucash/gnome/dialog-sx-editor.c
@@ -1128,6 +1128,30 @@ sxed_delete_event (GtkWidget *widget, GdkEvent *event, gpointer ud)
     return FALSE;
 }
 
+static gboolean
+focus_idle_callback(gpointer user_data)
+{
+    GNCLedgerDisplay  *ledger_display = (GNCLedgerDisplay  *)user_data;
+
+    if (ledger_display)
+        gnc_ledger_display_refresh(ledger_display);
+
+    return FALSE;
+}
+
+static void
+on_notebook_switch_page(GtkNotebook *notebook, GtkWidget *page,
+                        guint page_num, gpointer user_data)
+{
+    GtkWidget *current_page = gtk_notebook_get_nth_page(notebook, page_num);
+    if (current_page && page_num == 2)
+    {
+        GncSxEditorDialog *sxed = (GncSxEditorDialog *)user_data;
+
+        // Wait until Gtk is idle to refresh the display.
+        g_idle_add (focus_idle_callback, sxed->ledger);
+    }
+}
 
 /*************************************
  * Create the Schedule Editor Dialog *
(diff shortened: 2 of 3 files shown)


gnucash/gnome/gnc-plugin-page-register.cpp
```diff
@@ -3654,6 +3654,15 @@ gnc_plugin_page_register_cmd_cut (GSimpleAction *simple,
 
     ENTER ("(action %p, page %p)", simple, page);
     priv = GNC_PLUGIN_PAGE_REGISTER_GET_PRIVATE (page);
+
+    GtkWidget *widget = gtk_window_get_focus(GTK_WINDOW (priv->gsr->window));
+    const char *name = gtk_widget_get_name(widget);
+    if (strcmp(name, "GnucashSheet") != 0) {
+      gtk_editable_cut_clipboard( GTK_EDITABLE(widget));
+
+      return;
```
Bob-IT: ENTER at the start of the function is followed by LEAVE when you leave the function so you need to add some thing like LEAVE("Not cut from GnucashSheet"); above the return. Also the opening brace is on the wrong line and the indent should be 4 spaces.
This applies to all three functions.

Bob-IT:
I had a look at this bug and got side tracked so have tested your fix. It is an improvement but there was one thing I found when testing as follows...
Highlight some text on scheduled editor sheet and ctrl+c
Navigate to title with cursor at end and ctrl+v
Now select all the title and ctrl+c
Try pasting it back to same cell on sheet, it ends up in the title.

This is when I ran out of time so did not identify the reason.

Oscar65:
I can repeat the bug, but I think it's a different one. When you click on the same field as above in the scheduled editor sheet after going to title, the title doesn't lose focus. You need to press TAB or double-click the scheduled editor sheet field for the title to lose focus.

jralls:
> When you click on the same field as above in the scheduled editor sheet after going to title, the title doesn't lose focus.

Right. This is directly observable: The I-beam cursor remains in the title field. Interestingly clicking in a *different* field in the register does move the focus.

Oscar65:
After investigating, I found that the scheduled editor sheet loses focus, but not the box indicating that it is being edited.

If it doesn't have focus and you click on the box editing, it doesn't gain focus, but if you click on the same field, but in the small space outside the box editing, it does gain focus.

The problem I'm having is that I don't know how to get the sheet variable to gain focus every time the "Template Transaction" tab is clicked or run gnucash_sheet_deactivate_cursor_cell(reg->sheet);.

If I run gtk_widget_grab_focus(GTK_WIDGET(sxed->embed_window)); or gtk_widget_grab_focus(GTK_WIDGET(sxed->nameEntry)); every time the "Template Transaction" tab is clicked, it does nothing. The "Duplicate" button always has focus. But to test if it works, I run gtk_widget_set_margin_top (GTK_WIDGET(sxed->embed_window), 500); and it does.

If you press the TAB key when entering the "Template Transaction" tab
when the focus is on the "Duplicate" button, then the field with the box editing gains focus.

Any ideas on how to fix this bug?

Bob-IT:
I dug through my saved working on patches and came across one that was looking into the gnucash-item-edit.c line 892

    // Make sure the Entry can not have focus and no frame
    gtk_widget_set_can_focus (GTK_WIDGET(item_edit->editor), FALSE);

Changing this to TRUE seems to sort of fix this but I can not remember why it was set that way.
May be a pointer.

jralls:
@Bob-IT That line is from 1ed4f3d4d7. In the commit message you wrote 
> Most of the call backs and procedures rely on the sheet being the one in focus so prevent the entry being in focus. In doing this there is no text cursor so add a draw function to draw one after the entry drawn. Also remove the frame from the entry.

So is the paste supposed to go to the sheet and the sheet figures out what frame to put it into?

Bob-IT:
I think the focus can be fixed by adding this to the end of the existing `button_press_cb` in gnucash-item-edit.c at line 860 before the return FALSE;

    if (!gtk_widget_has_focus (GTK_WIDGET(sheet)))
        gtk_widget_grab_focus (GTK_WIDGET(sheet));

When the title entry is focused at the top of the dialog, the sheet focus out event is triggered which then triggers a focus out event for the item_edit. Possibly as that can not have focus, it upsets the events. When you click back in the same cell, there is no sheet in focus event so this change makes sure there is.

jralls:
LGTM. @Bob-IT ?

Bob-IT:
To start with you do not need the `gnucash-sheet.c` changes, with them the transfer cell popup does not work, probably all popups.
The source file changes in `dialog-sx-editor.c` need a bit off realignment.

I am not sure why there is a need for the `IdleType` structure, why not just do...


```

static gboolean
focus_idle_callback (gpointer user_data)
{
    GNCLedgerDisplay *ledger_display = (GNCLedgerDisplay *)user_data;

    if (ledger_display)
        gnc_ledger_display_refresh (ledger_display);

    return FALSE;
}

static void
on_notebook_switch_page (GtkNotebook *notebook, GtkWidget *page,
                         guint page_num, gpointer user_data)
{
    GtkWidget *current_page = gtk_notebook_get_nth_page (notebook, page_num);
    if (current_page && page_num == 2)
    {
        GncSxEditorDialog *sxed = (GncSxEditorDialog *)user_data;

        // Wait until Gtk is idle to refresh the display.
        g_idle_add (focus_idle_callback, sxed->ledger);
    }
}

```

Oscar65:
@Bob-IT Do you mean Action / Transfer...? On my PC it works. Could you be more specific about why does not work?

Sorry, I forgot to delete IdleType. I used before because I needed to pass two parameters.

Bob-IT:
@Oscar65, With those two lines left in, if I mouse click on the transfer popup button, the button just disappears. The same happens in a normal register, mouse click on the date popup button and it just disappears.

Remove the two lines and the popup button works.

Bob-IT:
@Oscar65
I am not really sure about your addition of changing the focus setting of the item-edit widget. I am always reluctant in changing stuff in the register as I cannot test every thing. I think my other change would have less of an impact.

Also please try to follow the source style, compare your 'IdleType'  changes to my suggested changes.

I did also try this on Windows and was OK.

gnucash/register/register-gnome/gnucash-sheet.c
```diff
@@ -1273,6 +1273,7 @@ gnucash_sheet_focus_out_event (GtkWidget *widget, GdkEventFocus *event)
         (widget, event);
 
     gnc_item_edit_focus_out (GNC_ITEM_EDIT(sheet->item_editor));
+
```
Bob-IT: @Oscar65 Can you remove this add blank line

gnucash/gnome/gnc-plugin-page-register.cpp
```diff
@@ -3688,6 +3710,17 @@ gnc_plugin_page_register_cmd_paste (GSimpleAction *simple,
 
     ENTER ("(action %p, page %p)", simple, page);
     priv = GNC_PLUGIN_PAGE_REGISTER_GET_PRIVATE (page);
+
+    GtkWidget *widget = gtk_window_get_focus(GTK_WINDOW (priv->gsr->window));
+    const char *name = gtk_widget_get_name(widget);
+    if (strcmp(name, "GnucashSheet") != 0)
+    {
+        gtk_editable_paste_clipboard( GTK_EDITABLE(widget));
+        LEAVE("Not paste from GnucashSheet");
```
Bob-IT: @Oscar65 Maybe change this to LEAVE("Not pasted to GnucashSheet");

gnucash/gnome/gnc-plugin-page-register.cpp
```diff
@@ -3671,6 +3682,17 @@ gnc_plugin_page_register_cmd_copy (GSimpleAction *simple,
 
     ENTER ("(action %p, page %p)", simple, page);
     priv = GNC_PLUGIN_PAGE_REGISTER_GET_PRIVATE (page);
+
+    GtkWidget *widget = gtk_window_get_focus(GTK_WINDOW (priv->gsr->window));
+    const char *name = gtk_widget_get_name(widget);
+    if (strcmp(name, "GnucashSheet") != 0)
+    {
+        gtk_editable_copy_clipboard( GTK_EDITABLE(widget));
+        LEAVE("Not copy from GnucashSheet");
```
Bob-IT: @Oscar65 Maybe change this to LEAVE("Not copied from GnucashSheet");

Bob-IT:
@Oscar65
After the suggested changes can you squash all the commits to one commit and make sure the first commit title line is the same as that of the bug report. Once done I think this will be ready to push.

jralls:
@Bob-IT @Oscar65 has made changes. Are they sufficient to merge this?

PR #2096, opened 2025-05-29
Modified quote sources to match current F::Q

bpschuck:
Modified libgnucash/engine/gnc-commodity.cpp to match current Finance::Quote.

In addition to removing a few modules from `single_quote_sources`, a bunch of sources were removed from `multiple_quote_sources` which refers to what Finance::Quote calls "failover methods". At the current time only 7 methods exist in F::Q where alternate methods defined in other modules will be attempted until data is successfully returned.

### DIFF
--- a/libgnucash/engine/gnc-commodity.cpp
+++ b/libgnucash/engine/gnc-commodity.cpp
@@ -201,15 +201,12 @@ static QuoteSourceList single_quote_sources =
     { false, SOURCE_SINGLE, NC_("FQ Source", "comdirect, DE"), "comdirect" },
     { false, SOURCE_SINGLE, NC_("FQ Source", "Consors Bank, DE"), "consorsbank" },
     { false, SOURCE_SINGLE, NC_("FQ Source", "Deka Investments, DE"), "deka" },
-    { false, SOURCE_SINGLE, NC_("FQ Source", "DWS, DE"), "dwsfunds" },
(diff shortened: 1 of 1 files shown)


libgnucash/engine/gnc-commodity.cpp
```diff
@@ -239,22 +235,14 @@ static QuoteSourceList single_quote_sources =
     { false, SOURCE_SINGLE, NC_("FQ Source", "YH Finance (FinanceAPI)"), "financeapi" },
 };
 
+// Finance::Quote calls these failover sources
```
jralls: How about "provides" instead of "calls"?

bpschuck: I meant "calls" as an "also known as" or a definition, not a programmatic call.
Perhaps change it to "Finance::Quote defines these as failover methods"?

jralls: Sure, anything that's less ambiguous.

bpschuck: Change has been made.

jralls:
Thanks!

bpschuck:
> Thanks!

You're welcome.

Just to confirm, the https://wiki.gnucash.org/wiki/Online_Quotes#Finance::Quote_Sources tables are no longer part of the documentation source, correct?

Bruce S.

jralls:
Correct. https://github.com/Gnucash/gnucash-docs/pull/346 was merged last December.

PR #2097, opened 2025-05-31
Bug 799521 - Segmentation fault on Autocomplete of Description with ß 

jralls:
Use [ICU String Search](https://unicode-org.github.io/icu/userguide/collation/string-search.html) to fix both the crash reported in [bug 799521](https://bugs.gnucash.org/show_bug.cgi?id=799521) and the markup-misalignment, both resulting from trying to apply an index and length from a normalized and case-folded string to the original string. When normalizing and case folding change the number of code points in the string the results are unsatisfactory.

This initial version of `gnc_utf8_has_substring` is hard-coded to the broadest matches to support this immediate use case. ISTM it would make sense to enable stricter matching for other uses.

### DIFF
--- a/libgnucash/core-utils/gnc-unicode.h
+++ b/libgnucash/core-utils/gnc-unicode.h
@@ -1,0 +1,144 @@
+/********************************************************************
+ * gnc-icu-locale.h -- Localization with ICU.                        *
+ *                                                                  *
+ * Copyright (C) 2025 John Ralls <jralls@ceridwen.us                *
+ *                                                                  *
+ * This program is free software; you can redistribute it and/or    *
+ * modify it under the terms of the GNU General Public License as   *
+ * published by the Free Software Foundation; either version 2 of   *
+ * the License, or (at your option) any later version.              *
+ *                                                                  *
+ * This program is distributed in the hope that it will be useful,  *
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of   *
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the    *
+ * GNU General Public License for more details.                     *
+ *                                                                  *
+ * You should have received a copy of the GNU General Public License*
+ * along with this program; if not, contact:                        *
+ *                                                                  *
+ * Free Software Foundation           Voice:  +1-617-542-5942       *
+ * 51 Franklin Street, Fifth Floor    Fax:    +1-617-542-2652       *
+ * Boston, MA  02110-1301,  USA       gnu@gnu.org                   *
+ ********************************************************************/
+
+#pragma once
+#include <stdbool.h>
+
+/** @addtogroup Localization These functions perform string comparison
+    and collation according to the Unicode Common Locale Data
+    Repository rules. The CLDR specifies five levels of
+    comparison.
+
+    - The primary or base level considers all variant codepoints
+    representing a character to be equivalent regardless of case or
+    decorations like accents and vowel or tone marks.
+    - The secondary level differentiates between letters with
+    decorations but still ignores case.
+    - The tertiary level differentiates based on case, decorations,
+    and variants, for example A and Ⓐ.
+    - The Quaternary level differentiates words with punctuation, for
+    example "ab" and "a-b".
+    - Identical differentiates all codepoints with no implicit
+      normalization so a character constructed with combining marks
+      will compare different from the same character represented as a
+      single codepoint.
+*/
+#ifdef __cplusplus
+extern "C"
+{
+#endif
+
+/** Find the first Unicode-equivalent UTF-8-encoded substring in a
+ * UTF-8-encoded string comparing characters at the CLDR primary
+ * level, setting the starting position and length of the matching
+ * part of the string.
+ * @param needle The substring to search for
+ * @param haystack the string to search in
+ * @param output the position of needle in haystack
+ * @param output the length of the match
+ * @return true if needle is found in haystack
+ */
+bool gnc_unicode_has_substring_base_chars(const char* needle, const
+                                          char* haystack, int* position,
+                                          int* length);
+
+/** Find the first Unicode-equivalent UTF-8-encoded substring in a
+ * UTF-8-encoded string comparing characters at the CLDR secondary
+ * level, setting the starting position and length of the matching
+ * part of the string.
+ * @param needle The substring to search for
+ * @param haystack the string to search in
+ * @param output the position of needle in haystack
+ * @param output the length of the match
+ * @return true if needle is found in haystack
+ */
+bool gnc_unicode_has_substring_accented_chars(const char* needle, const
+                                              char* haystack, int* position,
+                                              int* length);
+
+/** Find the first Unicode-equivalent UTF-8-encoded substring in a
+ * UTF-8-encoded string comparing characters at the CLDR tertiary
+ * level, setting the starting position and length of the matching
+ * part of the string.
+ * @param needle The substring to search for
+ * @param haystack the string to search in
+ * @param output the position of needle in haystack
+ * @param output the length of the match
+ * @return true if needle is found in haystack
+ */
+bool gnc_unicode_has_substring_accented_case_sensitive(const char* needle, const
+                                                       char* haystack, int* position,
+                                                       int* length);
+/** Find the first Unicode-equivalent UTF-8-encoded substring in a
+ * UTF-8-encoded string comparing characters at the CLDR identical
+ * level, setting the starting position and length of the matching
+ * part of the string.
+ * @param needle The substring to search for
+ * @param haystack the string to search in
+ * @param output the position of needle in haystack
+ * @param output the length of the match
+ * @return true if needle is found in haystack
+ */
+bool gnc_unicode_has_substring_identical(const char* needle, const
+                                         char* haystack, int* position,
+                                         int* length);
+/** Compare two UTF-8 encoded strings for equivalence at the CLDR
+ * primary level in the current locale. Errors are logged to
+ * gnc.locale.
+ * @param one a string
+ * @param two another string
+ * @return 0 if one and two are equivalent, -1 if one is less that
+ * two, 1 if one is greater than two, -99 on error.
+ */
+int gnc_unicode_compare_base_chars(const char* one, const char* two);
+
+/** Compare two UTF-8 encoded strings for equivalence at the CLDR
+ * secondary level in the current locale. Errors are logged to
+ * gnc.locale.
+ * @param one a string
+ * @param two another string
+ * @return 0 if one and two are equivalent, -1 if one is less that
+ * two, 1 if one is greater than two, -99 on error.
+ */
+int gnc_unicode_compare_accented_chars(const char* one, const char* two);
+
+/** Compare two UTF-8 encoded strings for equivalence at the CLDR
+ * tertiary level in the current locale. Errors are logged to
+ * gnc.locale.
+ * @param one a string
+ * @param two another string
+ * @return 0 if one and two are equivalent, -1 if one is less that
+ * two, 1 if one is greater than two, -99 on error.
+ */
+int gnc_unicode_compare_accented_case_sensitive(const char* one, const char* two);
+/** Compare two UTF-8 encoded strings for equivalence at the CLDR
+ * identical level in the current locale.
+ * @param one a string
+ * @param two another string
+ * @return 0 if one and two are equivalent, -1 if one is less that
+ * two, 1 if one is greater than two, -99 on error.
+ */
+int gnc_unicode_compare_identical(const char* one, const char* two);
+#ifdef __cplusplus
+}
+#endif
--- a/libgnucash/core-utils/test/gtest-gnc-unicode.cpp
+++ b/libgnucash/core-utils/test/gtest-gnc-unicode.cpp
@@ -1,0 +1,135 @@
+/********************************************************************\
+ * test-icu-locale.cpp -- Unit tests for GncQuotes            *
+ *                                                                  *
+ * Copyright 2025 John Ralls <jralls@ceridwen.us>                   *
+ *                                                                  *
+ * This program is free software; you can redistribute it and/or    *
+ * modify it under the terms of the GNU General Public License as   *
+ * published by the Free Software Foundation; either version 2 of   *
+ * the License, or (at your option) any later version.              *
+ *                                                                  *
+ * This program is distributed in the hope that it will be useful,  *
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of   *
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the    *
+ * GNU General Public License for more details.                     *
+ *                                                                  *
+ * You should have received a copy of the GNU General Public License*
+ * along with this program; if not, contact:                        *
+ *                                                                  *
+ * Free Software Foundation           Voice:  +1-617-542-5942       *
+ * 51 Franklin Street, Fifth Floor    Fax:    +1-617-542-2652       *
+ * Boston, MA  02110-1301,  USA       gnu@gnu.org                   *
+ *                                                                  *
+\********************************************************************/
+
+#include <gnc-unicode.h>
+#include <stdbool.h>
+#include <gtest/gtest.h>
+
+TEST(GncUnicode, test_ss_base_chars)
+{
+    int pos = 0, len = 0;
+
+    auto result = gnc_unicode_has_substring_base_chars("besi", "Necklace for Bessie",
+                                         &pos, &len);
+    EXPECT_FALSE(result);
+    EXPECT_EQ(0, pos);
+    EXPECT_EQ(0, len);
+
+    result = gnc_unicode_has_substring_base_chars("bessi", "Necklace for Bessie",
+                                         &pos, &len);
+    EXPECT_TRUE(result);
+    EXPECT_EQ(13, pos);
+    EXPECT_EQ(5, len);
+    }
+TEST(GncUnicode, test_ss_accented)
+{
+    int pos = 0, len = 0;
+    auto result = gnc_unicode_has_substring_accented_chars("bessi", "Necklace for Bessie",
+                                         &pos, &len);
+    EXPECT_TRUE(result);
+    EXPECT_EQ(13, pos);
+    EXPECT_EQ(5, len);
+}
+
+TEST(GncUnicode, test_ss_accented_case_sensitive)
+{
+    int pos = 0, len = 0;
+    auto result = gnc_unicode_has_substring_accented_case_sensitive("bessi", "Necklace for Bessie",
+                                         &pos, &len);
+    EXPECT_FALSE(result);
+    EXPECT_EQ(0, pos);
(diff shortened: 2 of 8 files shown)


Bob-IT:
The change has fixed the highlight issue and also the out of bounds error but some thing seems to of changed in the order of the list, seems to be backwards, before change...

original-list

after...
new-list

jralls:
@Bob-IT good catch. I misunderstood the exact match comparison and did it wrong. Should be fixed now.

jralls:
Now a more general question. @christopherlam and @gjanssens please consider weighing in here:
I see this as a widely usable improvement over `g_utf8` comparisons and roll-our-own string searches, but we won't always want [level 1](https://unicode-org.github.io/icu/userguide/collation/concepts.html#comparison-levels) and case-in∂sensitive dependent comparisons. ICU makes that easy to do. We can expose it either by adding  `accent-sensitive` and `case-sensitive` parameters to `gnc_utf8_has_substring` and `gnc_utf8_compare` or with separate functions, e.g. `gnc_utf8_compare_base`, `gnc_utf8_compare_base_nocase` (the current `gnc_utf8_compare` implementation), `gnc_utf8_compare_accented`, and `gnc_utf8_compare_accented_nocase`.  My inclination is the latter because I think it's clearer than parameters, even if the parameters are enums.

Also, any thoughts on the function and file names?

christopherlam:
My preference would be `gnc_icu_compare_base|accented_case[in]sensitive` to remind that it uses the almighty heavy duty ICU instead of the generic utf8 functions. Also `nocase` is ambiguous; better be explicit.

Also see gnc-date.cpp#get_icu_resources -- I think if we need to ICU compare lots of strings, we should make the collator `static`.

Also `locale` must be g_freed.

jralls:
> My preference would be gnc_icu_compare_base|accented_case[in]sensitive to remind that it uses the almighty heavy duty ICU instead of the generic utf8 functions. Also nocase is ambiguous; better be explicit.

While ICU is substantially more capable than GLib's `g_utf8_` functions "heavy duty" implies they're more expensive. I don't think there's much evidence of that: I profiled the autocomplete list creation on a largish account (several thousand transactions yielding 673 entries in the hash table) and there wasn't a significant difference.

`gnc_icu_compare_base` doesn't express what's being compared while `utf8` in `gnc_utf8_compare` is clearly a string comparison, plus putting `icu` in the public API exposes internals. How about `gnc_unicode_compare` etc. and the files renamed to `gnc-unicode.x`? 

> I think if we need to ICU compare lots of strings, we should make the collator static.

I profiled `gnc_utf8_has_substring` and creating the collator was only 7% of the search time. 90% of the time was split evenly between constructing the StringSearch and `search.first`.

> Also locale must be g_freed.

Thanks.

jralls:
It turns out that even though it documents the possibility of case-sensitive primary comparisons it's not actually implemented, so there's just `gnc_unicode_has_substring_base`.

> My preference would be `gnc_icu_compare_base|accented_case[in]sensitive `

That's pretty long winded and even more so with `gnc_unicode`:
`gnc_unicode_compare_base`
`gnc_unicode_compare_accented_case_sensitive`
`gnc_unicode_compare_accented_case_insensitive`
`gnc_unicode_has_substring_base`
`gnc_unicode_has_substring_accented_case_sensitive`
`gnc_unicode_has_substring_accented_case_insensitive`

That's getting a bit long at 50 characters, but I've made the changes so you can see them and added some more tests to make sure all of the variants work as expected.

gjanssens:
I think those are very useful functions that will save us headaches when gnucash is used beyond the ascii world.

Reading the comments in the header file I find ICU has 5 strictness levels, where we only expose 3. The Quaternary level may be to subtle to expose. Have you considered the identical level though ? And found no proper use case for it ?

Other than that I can only do some bike shedding on the function names :wink: 

* "base" was misleading me until I read the code. It thought it to mean a base function which the others build upon, so the variant which takes extra parameters to specify the strictness of the match. We have several patterns like that in gnucash, albeit those usually use "internal" to indicate so. Perhaps "liberal" reflects the intent better ?
* I understand the choice of explicit naming in C code. But perhaps a C++ variant where a single function (with defaults set for the strictness parameters) could be nice too ? The "liberal" function call could then be done without setting the parameters, hence falling back to the defaults. "Accented/Case Insensitive" might be accomplished by setting the accented parameter and "Accented/Case sensitive" would require two parameters to be set ? That's just how I like C++ code to be written, but perhaps that's a practice that's frowned upon these days ?
* I agree the names become fairly long. I don't know for the accented part, but as far as I'm concerned summarizing case insensitive as "ci" and case sensitive as "cs" is still expressive enough.

libgnucash/core-utils/gnc-unicode.h
```diff
@@ -0,0 +1,77 @@
...
+ ********************************************************************/
+
+#pragma once
+#include <stdbool.h>
+
+/** @addtogroup Localization These functions perform string comparison
+    and collation according to the Unicode Common Locale Data
+    Repository rules. The CLDR specifies five levels of
+    comparison. The primary or base level considers all variant
+    codepoints representing a character to be equivalent regardless of
```
gjanssens: More minor bike shedding: this description calls for a (un)numbered list for readability.

libgnucash/core-utils/gnc-unicode.h
```diff
@@ -0,0 +1,77 @@
...
+    case or decorations like accents and vowel or tone marks. The
+    secondary level differentiates between letters with decorations
+    but still ignores case. The tertiary level differentiates based on
+    case, decorations, and variants--for example A and Ⓐ, The
+    Quaternary level differentiates words with punctuation, for
+    example "ab" and "a-b". Identical differentiates all codepoints.
+
+    In the following functions "base" means primary level comparison,
+    "accented case" means secondary level comparison, and
```
gjanssens: I'dd add this bit in the comment blocks of the actual functions and refer to the icu levels in this group comment block from there.

One typically sees the function's comment first, so that link direction makes more sense.

libgnucash/core-utils/test/gtest-gnc-unicode.cpp
```diff
@@ -0,0 +1,135 @@
...
+{
+    int pos = 0, len = 0;
+    auto result = gnc_unicode_has_substring_base("be\xc3\x9fi", "Necklace for Be\xc3\x9fie",
+                                         &pos, &len);
+    EXPECT_TRUE(result);
+    EXPECT_EQ(13, pos);
+    EXPECT_EQ(4, len);
+
+    pos = 0, len = 0;
+    result = gnc_unicode_has_substring_accented_case_insensitive("be\xc3\x9fi", "Necklace for Be\xc3\x9fie",
```
gjanssens: Out of curiosity, is it bad practice to put the actual unicode character directly in a string constant in C++ ? Don't we have [string literal qualifiers](http://www.en.cppreference.com/w/cpp/language/string_literal.html) now that help express the string is supposed to be interpreted as unicode ?

For example isn't `u8"Be쎟ie"` equivalent to `"Be\xc3\x9fi"` ?

jralls: We do, but emacs's mule IM is a PITA and I didn't bother to copy-and-paste the character. On macOS I can just type <option>s. I'm also distrustful that Windows editors won't either convert the explicit character to UTF-16 or worse convert it to some code-page codepoint. Using the escape is bulletproof or maybe Windows-proof.

BTW, `"Be\xc3\x9fi"` is `u8"Beßie"`. The character in the middle of  u8"Be쎟ie" is `\xec\x8e\x9f`, [U+C39F](https://codepoints.net/U+C39F?lang=en). It looks like you read the UTF-8 `\xc3\x9f` as UTF-16 `\xc39f`.

jralls:
> We have several patterns like that in gnucash, albeit those usually use "internal" to indicate so.

As we do with `unicode_has_string_internal` and `unicode_compare_internal`.

> "base" was misleading me until I read the code. It thought it to mean a base function which the others build upon, so the variant which takes extra parameters to specify the strictness of the match. Perhaps "liberal" reflects the intent better ?

I understand the ambiguity but "liberal" has unfortunate overtones in US English these days. `unaccented` would be the natural opposite to `accented` but the two-letters-in-ten difference will be hard to notice in a long function name; that's also true of `case_insensitive`. How about "plain" instead of "base"? Case sensitivity is enabled by tertiary strength and that also turns on differentiating by character variants like `C` and `©` so maybe we need a different name-pair for that too.

jralls:
> I understand the choice of explicit naming in C code. But perhaps a C++ variant where a single function (with defaults set for the strictness parameters) could be nice too ? The "liberal" function call could then be done without setting the parameters, hence falling back to the defaults. "Accented/Case Insensitive" might be accomplished by setting the accented parameter and "Accented/Case sensitive" would require two parameters to be set ? That's just how I like C++ code to be written, but perhaps that's a practice that's frowned upon these days ?

What's frowned on as being error prone is a bunch of `bool` parameters like the old `void qof_session_begin (QofSession *session, const char * new_uri, gboolean ignore_lock, gboolean create, gboolean force);` that I fixed in https://github.com/Gnucash/gnucash/commit/6a1cb5eecd794d65ab0d15161b1628750b29acb6. [comment 1](https://github.com/Gnucash/gnucash/pull/2097#issuecomment-2931891910) offered enum parameters as an alternative to function names and @christopherlam expressed a preference for the latter.

Note that since I've discovered that ICU's implementation ignores the `UCOL_CASE_LEVEL` attribute there's only the one enum, `ComparisonStrength`.  It's a C++ `enum class` to avoid polluting the global namespace, but there wouldn't be any real harm in making it a C enum. I think a good argument can be made that `gnc_unicode_compare(foo, bar, COMPARE_STRENGTH_SECONDARY)` and `gnc_unicode_compare(foo, bar, COMPARE_STRENGTH_TERTIARY)` are easier to tell apart than are `gnc_unicode_compare_accented_case_insensitve` and`gnc_unicode_compare_accented_case_sensitve` .

jralls:
> Have you considered the identical level though ? And found no proper use case for it ?

`Identical` turns off normalization so `U+0060U+0041` and `U+00E0` (à) compare as different. That can be accomplished more quickly with `strncmp` and `strstr` so I don't see a point to exposing it.

libgnucash/core-utils/test/gtest-gnc-unicode.cpp
gjanssens: Yeah, I didn't really know how to read these characters... I **imagined** it to mean ß, but didn't know how to check that :face_with_diagonal_mouth: 

Other than that good enough.

jralls: Search the web for "utf-8 c3 9f" or use any of the myriad utf-8 calculator websites. The [details](https://en.wikipedia.org/wiki/UTF-8#Description) in case you're interested.

gjanssens: You're correct about my misinterpretation as UTF-16. My confusion comes from the Character map tool on my linux system that shows these UTF-16 code points for each symbol. I didn't realize that. Looking better I find it also displays UTF-8 further down on the info panel, but I can't search on it...

gjanssens:
> > I understand the choice of explicit naming in C code. But perhaps a C++ variant where a single function (with defaults set for the strictness parameters) could be nice too ? The "liberal" function call could then be done without setting the parameters, hence falling back to the defaults. "Accented/Case Insensitive" might be accomplished by setting the accented parameter and "Accented/Case sensitive" would require two parameters to be set ? That's just how I like C++ code to be written, but perhaps that's a practice that's frowned upon these days ?
> 
> What's frowned on as being error prone is a bunch of `bool` parameters like the old `void qof_session_begin (QofSession *session, const char * new_uri, gboolean ignore_lock, gboolean create, gboolean force);` that I fixed in [6a1cb5e](https://github.com/Gnucash/gnucash/commit/6a1cb5eecd794d65ab0d15161b1628750b29acb6). [comment 1](https://github.com/Gnucash/gnucash/pull/2097#issuecomment-2931891910) offered enum parameters as an alternative to function names and @christopherlam expressed a preference for the latter.
> 

Got it.

> Note that since I've discovered that ICU's implementation ignores the `UCOL_CASE_LEVEL` attribute there's only the one enum, `ComparisonStrength`. It's a C++ `enum class` to avoid polluting the global namespace, but there wouldn't be any real harm in making it a C enum. I think a good argument can be made that `gnc_unicode_compare(foo, bar, COMPARE_STRENGTH_SECONDARY)` and `gnc_unicode_compare(foo, bar, COMPARE_STRENGTH_TERTIARY)` are easier to tell apart than are `gnc_unicode_compare_accented_case_insensitve` and`gnc_unicode_compare_accented_case_sensitve` .

I agree here. As long as we clearly explain somewhere what the different levels mean exactly. You already do in the new header file. If we go for a plain C enum, that enum's declaration would be the place to document it.

gjanssens:
> > Have you considered the identical level though ? And found no proper use case for it ?
> 
> `Identical` turns off normalization so `U+0060U+0041` and `U+00E0` (à) compare as different. That can be accomplished more quickly with `strncmp` and `strstr` so I don't see a point to exposing it.

Maybe. In C++ circles I hear a lot of about expressing intent. For me an `IDENTICAL` enum would express the intent more clearly than `strncmp` or `strstr` and helps maintaining a single, consistent interface to use for string comparions. I don't follow conversations in the C world, so I don't know trends there. On the other hand, the main argument in favour of the latter two would be performance. So if the ICU based code performs significantly worse, that would be a good reason to use the other two.

We could of course optimize `gnc_unicode_compare(foo, bar, COMPARE_STRENGTH_IDENTICAL)`, to just call `strncmp` or `strstr` internally to keep a unified interface and still have the performance benefits. I wonder even if ICU internally might not already optimize this path itself.

No strong opinion here on my side. I'm just adding some wider context to consider for these new functions.

jralls:
> We could of course optimize gnc_unicode_compare(foo, bar, COMPARE_STRENGTH_IDENTICAL), to just call strncmp or strstr internally to keep a unified interface and still have the performance benefits. I wonder even if ICU internally might not already optimize this path itself.

I like the idea of optimizing IDENTICAL in `gnc_unicode_compare` and `gnc_unicode_has_substring`. There might be an optimization in the former case as `icu::Collator::compare` has a `char*` overload but `icu::StringSearch()` doesn't so we always incur the cost of an `icu::UnicodeString` conversion.

jralls:
>>I think a good argument can be made that gnc_unicode_compare(foo, bar, COMPARE_STRENGTH_SECONDARY) and gnc_unicode_compare(foo, bar, COMPARE_STRENGTH_TERTIARY) are easier to tell apart than are gnc_unicode_compare_accented_case_insensitve andgnc_unicode_compare_accented_case_sensitve .

> I agree here. As long as we clearly explain somewhere what the different levels mean exactly. 

But we could also name the functions that way, i.e. `gnc_unicode_compare_primary` etc. That would be friendlier to bindings than using the enum.

Apropos that I just looked at the Guile [string comparison doc](https://www.gnu.org/software/guile/manual/html_node/String-Comparison.html). They take the same simplistic character-by-character approach that `g_utf8_foo` does, requiring manual normalization, though they do offer built-in case-insensitive comparisons. It doesn't appear that there's any way to get CLDR Primary comparisons. [Python](https://docs.python.org/3/library/stdtypes.html#textseq) doesn't explicitly define a comparison function but the presence of both [`casefold`](https://docs.python.org/3/library/stdtypes.html#str.casefold) and [normalize](https://docs.python.org/3/library/unicodedata.html#unicodedata.normalize) indicates a similar situation. We likely want to bind overloads of these functions that skip the charset conversions since both Guile and Python internally use UTF-16.

gjanssens:
> > > I think a good argument can be made that gnc_unicode_compare(foo, bar, COMPARE_STRENGTH_SECONDARY) and gnc_unicode_compare(foo, bar, COMPARE_STRENGTH_TERTIARY) are easier to tell apart than are gnc_unicode_compare_accented_case_insensitve andgnc_unicode_compare_accented_case_sensitve .
> 
> > I agree here. As long as we clearly explain somewhere what the different levels mean exactly.
> 
> But we could also name the functions that way, i.e. `gnc_unicode_compare_primary` etc. That would be friendlier to bindings than using the enum.
> 

Fair point.

> Apropos that I just looked at the Guile [string comparison doc](https://www.gnu.org/software/guile/manual/html_node/String-Comparison.html). They take the same simplistic character-by-character approach that `g_utf8_foo` does, requiring manual normalization, though they do offer built-in case-insensitive comparisons. It doesn't appear that there's any way to get CLDR Primary comparisons. [Python](https://docs.python.org/3/library/stdtypes.html#textseq) doesn't explicitly define a comparison function but the presence of both [`casefold`](https://docs.python.org/3/library/stdtypes.html#str.casefold) and [normalize](https://docs.python.org/3/library/unicodedata.html#unicodedata.normalize) indicates a similar situation. We likely want to bind overloads of these functions that skip the charset conversions since both Guile and Python internally use UTF-16.

This goes beyond the level I have looked into this, but I trust you understand it better than I do.

gjanssens:
With the name change for the `basic` function the commit message of your second commit now refers to a non-exisiting function name...

jralls:
I'm not really keen on primary/secondary/tertiary in function names or a public enum. How about 
* `gnc_unicode_compare_base_chars`
* `gnc_unicode_compare_accented_chars`
* `gnc_unicode_compare_accented_cap_sensitive`
* `gnc_unicode_compare_identical`

gjanssens:
I understand. primary/secondary/tertiary are not very descriptive.

What does `cap` stand for in the third option ? I would have expected `case` but perhaps you deliberately chose it to express a wide scope ? Other than that I'm fine with the proposed function names.

jralls:
> What does cap stand for in the third option ?

Capitals. You're right, case would be better.

@christopherlam what do you think?

christopherlam:
Agree case better than cap[tain].

Im probably influenced by the scheme name string-compare-ci which seems intuitive, but naive; so, I have no further comments in this.

jralls:
OK, I've made the API
* `gnc_unicode_foo_base_chars`
* `gnc_unicode_foo_accented_chars`
* `gnc_unicode_foo_accented_cap_sensitive`
* `gnc_unicode_foo_identical`

Anything else?

gjanssens:
LTGM

christopherlam:
Forgot to ask which string-comparison variant should be used in reports, which currently use `gnc:string-locale<?` which currently use `g-utf8-collate`?

jralls:
Depends on what level comparison you want. `(define (gnc:string-locale<? a b) (< (g-utf8-collate a b) 0))` so it's going to be identical strength. That will work OK as long as there aren't normalization variants in the strings, but if you want to switch to `gnc_unicode_compare` I'd go with `gnc_unicode_compare_accented_case_sensitive`.  Note that I haven't yet added gnc-unicode.h to core-utils.i so you can't do that yet.

> That can be accomplished more quickly with strncmp and strstr so I don't see a point to exposing it.

I realized from looking at the docs for `g_utf8_collate` that that's only half right: `strncmp` won't do because it does a strictly numeric ordering on the bytes. That's only alphabetic for the ASCII characters. I'll have to fix `gnc_unicode_compare_identical` to use `unicode_compare_internal`.

PR #2098, opened 2025-06-02
Delay report generation

christopherlam:
better than #2092 

This will change behaviour of all html anchors from report to another report.

Previously the resulting report would be generated `(gnc:make-report)` and the options also generated, which is slow. This change will simply store the required data to generate the report into a hash table within the `report` SCM object. Thus a 5-year monthly multicolumn income statment renders in 1.3s instead of 4-5s.

~Unfortunately the report page isn't updated yet. With this change try:~

~1. load "Income Statemnt (Multicolumn)" report.~
~2. click on any amount cell, which _should_ switch to the transaction report. However the report tab is not yet updated.~
~3. quit and relaunch gnucash; the tab now shows the transaction report.~

~Something is missing between 2 and 3 to update the tab.~

Note this change uses srfi-69 which has a O(1) `hash-table-size`

### DIFF
--- a/gnucash/gnome/gnc-plugin-page-report.cpp
+++ b/gnucash/gnome/gnc-plugin-page-report.cpp
@@ -674,7 +674,12 @@ gnc_plugin_page_report_load_cb (GncHtml * html, URLType type,
             && (strlen(location) > 3)
             && !strncmp("id=", location, 3))
     {
-        report_id = atoi(location + 3);
+        report_id = gnc_report_id_string_to_report_id (location + 3);
+        if (report_id < 0)
+        {
+            LEAVE ("id_string error %s", location);
+            return;
+        }
         DEBUG( "parsed id=%d", report_id );
     }
     else if (!g_strcmp0( type, URL_TYPE_OPTIONS)
--- a/gnucash/report/gnc-report.cpp
+++ b/gnucash/report/gnc-report.cpp
@@ -257,12 +257,29 @@ gnc_report_name( SCM report )
     return gnc_scm_call_1_to_string(get_name, report);
 }
 
+gint
+gnc_report_id_string_to_report_id (const char *id_string)
+{
+    g_return_val_if_fail (id_string, -1);
+
+    char *end_ptr;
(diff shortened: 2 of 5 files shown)


Bob-IT:
I had a look at this and I see you have changed the `id` value to incorporate a `|` so I wonder if you have to change `gnc_html_parse_url`. I think it all leads to `impl_webkit_show_url` for Linux in gnc-html-webkit2.c, look from line 874.

For `new_location` I am getting `id=0|1`

christopherlam:
> I had a look at this and I see you have changed the `id` value to incorporate a `|` so I wonder if you have to change `gnc_html_parse_url`. I think it all leads to `impl_webkit_show_url` for Linux in gnc-html-webkit2.c, look from line 874.
> 
> For `new_location` I am getting `id=0|1`

Thanks.... I'm completely lost in the url handlers, and not sure how to trigger a tab refresh; or whether a tab refresh is the right approach. You're right `id=0|1` means gnc_report 0 and the 2nd report anchor.

EDIT: also, I'm not 100% happy that the `id=` handler in my commit needs an `if (strchr (start, '|'))` but it seems necessary.

Bob-IT:
OK, I think what is happening is that the  "Income Statement (Multicolumn)" is being reloaded from here...

gnc-report.cpp, line 261, id_string is `id=0|1` and creates report_id as `0` and hence creates the report again.

I suppose you would need to keep the report_id as id_string and this is then passed to `gnc_run_report_with_error_handling` which would need modifying to include some thing like your change to `gnc_plugin_page_report_load_cb`.

Does that make sense?

christopherlam:
Thanks @Bob-IT! Now this seems to work well! For maximum speedup check out Income Statement (Multicolumn), short period (eg 1month) long date range (eg 5yrs); this creates a spreadsheet of 60 columns, multiply by num(accounts), equal a lotta trep reports which would be formerly prepared. With this change, the trep parameters are simply stored until launched via clicking. The number of reports stored in gnc-report.cpp hash-table should also drastically reduce.

jralls (approved):
Looks OK.

PR #2099, opened 2025-06-04
[gnc-optiondb] allow any report option to have a "changed" callback

christopherlam:
Thus options other than `boolean` and `multichoice` can have callbacks. e.g. we can attach a callback to the `account-sel` option to set other options.

Adding another option type callback is as easy as:

```diff
modified   gnucash/gnome-utils/gnc-option-gtk-ui.cpp
@@ -215,6 +215,11 @@ public:
         auto widget{GTK_ENTRY(get_widget())};
         option.set_value(std::string{gtk_entry_get_text(widget)});
     }
+    SCM get_widget_scm_value(const GncOption& option) const override
+    {
+        auto widget{GTK_ENTRY(get_widget())};
+        return scm_from_utf8_string (gtk_entry_get_text (widget));
+    }
 };
 
 template<> void
modified   gnucash/report/trep-engine.scm
@@ -595,6 +595,11 @@ Expenses:Car and Expenses:Flights. Use a period (.) to match a single character
 '20../.' will match 'Travel 2017/1 London'. ")
     #f)
 
+  (gnc-option-set-changed-callback
+   options pagename-filter optname-account-matcher
+   (lambda (txt) (gnc:pk "new account matcher: " txt)))
+
   (gnc-register-simple-boolean-option options
     pagename-filter optname-account-matcher-exclude "a7"
     (G_ "If this option is selected, accounts matching filter are excluded.")
```

TODO: after updating other options, need to find a way to refresh ui.

### DIFF
--- a/bindings/guile/gnc-engine-guile.cpp
+++ b/bindings/guile/gnc-engine-guile.cpp
@@ -1788,6 +1788,13 @@ gnc_split_to_scm (const Split *split)
     return gnc_generic_to_scm (split, stype);
 }
 
+SCM
+gnc_account_to_scm (const Account *account)
+{
+    static auto stype = get_swig_type ("_p_Account");
+    return gnc_generic_to_scm (account, stype);
+}
+
 GncAccountValue * gnc_scm_to_account_value_ptr (SCM valuearg)
 {
     GncAccountValue *res;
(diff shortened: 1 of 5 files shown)


jralls:
> Thus options other than boolean and multichoice can have callbacks. e.g. we can attach a callback to the account-sel option to set other options.

That's a terrible idea. It makes it impossible to reason about the code. I consider it a very serious bug that I couldn't remove the callbacks from the two types that have it without breaking the reports that abuse it.

PR #2101, opened 2025-06-11
[reports] better gnc_report_id_string_to_report_id with from_chars

christopherlam:
- no implicit cast
- no locale-sensitive numeric decoding
- no nibble or hex decoding

and the user can have measly MAXINT reports instead of MAXUINT reports in memory...

### DIFF
--- a/gnucash/report/gnc-report.cpp
+++ b/gnucash/report/gnc-report.cpp
@@ -42,6 +42,7 @@
 #include <gnc-guile-utils.h>
 #include <gnc-engine.h>
 #include "gnc-report.h"
+#include <charconv>
 
 extern "C" SCM scm_init_sw_report_module(void);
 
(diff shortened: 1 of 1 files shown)


gnucash/report/gnc-report.cpp
```diff
@@ -262,19 +263,20 @@ gnc_report_id_string_to_report_id (const char *id_string)
 {
     g_return_val_if_fail (id_string, -1);
 
-    char *end_ptr;
-    guint rpt_id = std::strtoul (id_string, &end_ptr, 10);
-    if (end_ptr == id_string) return -1;
-    if (*end_ptr == '\0') return rpt_id;
-    if (*end_ptr != '|') return -1;
+    const char *end_ptr = id_string + std::strlen (id_string);
+    gint rpt_id;
```
gjanssens, line 267: You don't seem to expect negative report id's. So why wouldn't you then make `rpt_id` a `guint` (or `uint32_t` for that matter) ?
I may read the documentation on `std::from_chars` wrongly (as I did with `std::strtoul`), but I **think** it has overloads for all common integer types. It seems to me `uint23_t` is a valid integer type, no ?

christopherlam: Ok the motivation is to minimise the need for casting and tests. If we make `rpt_id` a `uint` and the function returns an `int`, then we need to range check `uint` before casting to, and returning `int`. I'd rather make the range check takes place within `from_chars` which will then error out appropriately.

christopherlam: > It seems to me `uint23_t` is a valid integer type, no ?

And yes `from_chars` handles any and every integer or float type, except the inexistent `uint23_t` ;-) see numerous int/uint 16/32/64 tests in 249ec9f43a0780b743e2fb2f0be2955399daad11

gjanssens: > Ok the motivation is to minimise the need for casting and tests. If we make `rpt_id` a `uint` and the function returns an `int`, then we need to range check `uint` before casting to, and returning `int`. I'd rather make the range check takes place within `from_chars` which will then error out appropriately.

You do the range check anyway as you verify the parsed value is greater than or equal to 0.
The conclusion is that the function return value is also wrong. But yes, perhaps that's too much of a change to handle here...

jralls: > You do the range check anyway as you verify the parsed value is greater than or equal to 0.

 He means the other end. INT32_MAX = UINT32_MAX / 2  - 1.

jralls: > Ok the motivation is to minimise the need for casting and tests. If we make rpt_id a uint and the function returns an int, then we need to range check uint before casting to, and returning int. I'd rather make the range check takes place within from_chars which will then error out appropriately.

But the reason for the rv to be signed is to signal a conversion error. Returning `std::optional<unsigned int>` would better express that.

gjanssens: I ignored the purpose of the negative return value indeed.

On a related note I recently read about [std::expected](https://en.cppreference.com/w/cpp/utility/expected.html) as the perfect fit for this use case. Unfortunately it's C++23 only, so unfortunately we can't use that one yet...

christopherlam:
Merged as is; because returning `optional<unsigned int>` would require changing the function signature to be c++.

jralls:
> Merged as is; because returning optional<unsigned int> would require changing the function signature to be c++.

So? It's only used in gnucash/gnome/gnc-plugin-page-report.cpp and you introduced it only two weeks ago.

PR #2102, opened 2025-06-14
[reports] use ICU for Unicode string comparison

christopherlam:
following #2097 use icu for unicode aware string comparison for use in reports

### DIFF
--- a/bindings/core-utils.i
+++ b/bindings/core-utils.i
@@ -29,6 +29,7 @@
 #include <gnc-locale-utils.h>
 #include <glib.h>
 #include <glib/gi18n.h>
+#include <gnc-unicode.h>
 #include <gnc-version.h>
 #include <libintl.h>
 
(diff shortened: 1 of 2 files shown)


bindings/core-utils.i
```diff
@@ -104,7 +105,7 @@ const char * gnc_locale_default_iso_currency_code (void);
 %newobject gnc_locale_name;
 gchar *gnc_locale_name (void);
 
-int g_utf8_collate (const char *a, const char *b);
+int gnc_unicode_compare_accented_case_sensitive (const char *a, const char *b);
```
jralls: `%include <gnc-unicode.h>` at line 53 to pull in all of the declarations instead of declaring just the one function here. Do remove `g_utf8_collate`.

jralls (dismissed):

jralls:
You need both `#include` *and* `%include`: `%include` tells SWIG to parse the file but it's the `#include` that gets  added to the swig file (swig-core-utils.c in this case).

jralls:
Ok, merge when ready.

PR #2104, opened 2025-06-18
[import-main-matcher.cpp] Add tooltips to transaction action columns

rkok:
A small usability improvement for novice users: this PR adds descriptive tooltips to the A, U+C, and C action columns in the generic transaction import matcher.

gc-tooltip


jralls:
Thanks, that's a good idea.

Note, though, that we only use context translations for very short strings, both to help translators know what they're for and in case they might be reused with different meanings that might have different words in some languages. The strings you've added are long enough that that's not a concern so I've removed the context in the merge commit. Normally I'd ask you to change the PR but today is string-freeze day for 5.12 and I want to get this in.

PR #2105, opened 2025-06-21
Modified gnc-commodity.cpp to match F::Q

bpschuck:
Modules Bloomberg.pm, Tiaacref.pm, Troweprice.pm removed. They will not be included in the soon to released F::Q v1.66.


jralls:
Thanks!

PR #2107, opened 2025-06-25
engine kvp access: remove need for GValue

christopherlam:
~however last commit won't pass ...~ the solution appears as soon as PR is created
// https://en.wikipedia.org/wiki/Rubber_duck_debugging

### DIFF
--- a/libgnucash/engine/Transaction.cpp
+++ b/libgnucash/engine/Transaction.cpp
@@ -1945,16 +1945,13 @@ xaccTransSetDatePostedSecsNormalized (Transaction *trans, time64 time)
 void
 xaccTransSetDatePostedGDate (Transaction *trans, GDate date)
 {
-    GValue v = G_VALUE_INIT;
     if (!trans) return;
 
     /* We additionally save this date into a kvp frame to ensure in
      * the future a date which was set as *date* (without time) can
      * clearly be distinguished from the time64. */
-    g_value_init (&v, G_TYPE_DATE);
-    g_value_set_static_boxed (&v, &date);
-    qof_instance_set_kvp (QOF_INSTANCE(trans), &v, 1, TRANS_DATE_POSTED);
-    g_value_unset (&v);
+    qof_instance_set_path_kvp<GDate> (QOF_INSTANCE(trans), date, {TRANS_DATE_POSTED});
+    qof_instance_set_dirty (QOF_INSTANCE(trans));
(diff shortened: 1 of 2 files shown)


jralls:
Looks reasonable. Are you ready to merge?

christopherlam:
> Looks reasonable. Are you ready to merge?

not yet. More tests are pending. I'm keen to harmonize the setters a bit more -- there some variability in how they handle "" vs `nullptr` for `const char*` values eg `xaccTransSetReadOnly`, `xaccTransSetNotes`, `xaccTransSetDocLink` with `nullptr` const char*; maybe the users need to do the null-check instead of the setter functions?

christopherlam:
@jralls what's your opinion on harmonizing the `xaccObjectSetProperty(obj,str)` whereby `str==nullptr`? Some will handle it as "remove property", but others seem to handle `str=nullptr` as NOP.

jralls:
Hmm, that nullptr->remove is a KVP paradigm. Properties that are implemented as instance variables can't be deleted. That sort of forces an API difference unless you define a default value for a KVP-backed variable and instantiate the KVP only when the variable isn't the default.

Mind, I'd like to get rid of GObject properties too and replace them with normal setters and getters along with proper full-on constructors instead of the GObject construct a null object and populate the member variables one at a time paradigm.

libgnucash/engine/Transaction.cpp
```diff
@@ -2416,18 +2332,9 @@ xaccTransRetDateEntered (const Transaction *trans)
...
-    if (G_VALUE_HOLDS_BOXED (&v))
-    {
-        ret = ((Time64*)g_value_get_boxed (&v))->t;
-        g_value_unset (&v);
-    }
-    if (!ret)
-        return xaccTransRetDatePosted (trans);
-    return ret;
+    auto res = qof_instance_get_path_kvp<Time64> (QOF_INSTANCE (trans), {TRANS_DATE_DUE_KVP});
+    return (res && res->t) ? res->t : xaccTransRetDatePosted (trans);
```
christopherlam: this line currently reproduces behaviour described in 7f82098c11

jralls: The bug is `&& res->t`: If `res` is false, meaning that the slot isn't set, then returning `xaccTransRetDatePosted` is correct, but in the extremely unlikely event that it's actually set to 0, i.e. 1970-01-01T00:00:00Z, then that's indeed what it should return.

PR #2108, opened 2025-06-28
Test load xml datafile

christopherlam:
includes a seed xml datafile to test stability of XML serialization/deserialization routines

The .gnucash file may be modified later on by UI so that further tests can be written

### DIFF
--- a/libgnucash/backend/xml/test/CMakeLists.txt
+++ b/libgnucash/backend/xml/test/CMakeLists.txt
@@ -38,6 +38,7 @@ set_local_dist(test_backend_xml_DIST_local
   test-load-backend.cpp
   test-load-example-account.cpp
   gtest-load-save-files.cpp
+  gtest-xml-contents.cpp
   test-load-xml2.cpp
   test-real-data.sh
   test-save-in-lang.cpp
@@ -69,6 +70,10 @@ target_compile_options(test-load-example-account PRIVATE -DU_SHOW_CPLUSPLUS_API=
 add_xml_gtest(test-load-save-files gtest-load-save-files.cpp
   GNC_TEST_FILES=${CMAKE_CURRENT_SOURCE_DIR}/test-files/load-save
 )
+add_xml_gtest(test-xml-contents gtest-xml-contents.cpp
+  GNC_TEST_FILES=${CMAKE_CURRENT_SOURCE_DIR}/test-files/xml-contents
+)
+
 add_xml_test(test-string-converters "test-string-converters.cpp")
 add_xml_test(test-xml-account "test-xml-account.cpp;test-file-stuff.cpp")
 add_xml_test(test-xml-commodity "test-xml-commodity.cpp;test-file-stuff.cpp")
--- a/libgnucash/backend/xml/test/gtest-xml-contents.cpp
+++ b/libgnucash/backend/xml/test/gtest-xml-contents.cpp
@@ -1,0 +1,204 @@
+/********************************************************************\
+ * This program is free software; you can redistribute it and/or    *
+ * modify it under the terms of the GNU General Public License as   *
+ * published by the Free Software Foundation; either version 2 of   *
+ * the License, or (at your option) any later version.              *
+ *                                                                  *
+ * This program is distributed in the hope that it will be useful,  *
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of   *
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the    *
+ * GNU General Public License for more details.                     *
+ *                                                                  *
+ * You should have received a copy of the GNU General Public License*
+ * along with this program; if not, contact:                        *
+ *                                                                  *
+ * Free Software Foundation           Voice:  +1-617-542-5942       *
+ * 51 Franklin Street, Fifth Floor    Fax:    +1-617-542-2652       *
+ * Boston, MA  02110-1301,  USA       gnu@gnu.org                   *
+ *                                                                  *
+\********************************************************************/
+#include <glib.h>
+
+#include <config.h>
+#include <sys/types.h>
+#include <sys/stat.h>
+
+#include <cashobjects.h>
+#include <TransLog.h>
+#include <gnc-engine.h>
+#include <gnc-lot.h>
+#include <gnc-prefs.h>
+#include <Account.hpp>
+#include <gnc-datetime.hpp>
+
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wcpp"
+#include <gtest/gtest.h>
+#pragma GCC diagnostic pop
+#include <unittest-support.h>
+
+#include "../gnc-backend-xml.h"
+#include "../io-gncxml-v2.h"
+
+#define GNC_LIB_NAME "gncmod-backend-xml"
+#define GNC_LIB_REL_PATH "xml"
+
+class LoadFile : public testing::Test
+{
+public:
+    static void SetUpTestSuite ()
+    {
+        g_setenv ("GNC_UNINSTALLED", "1", TRUE);
+        qof_init ();
+        cashobjects_register ();
+        ASSERT_TRUE(qof_load_backend_library (GNC_LIB_REL_PATH, GNC_LIB_NAME)) << "loading gnc-backend-xml GModule failed";
+        xaccLogDisable ();
+    }
+
+    static void TearDownTestSuite () { qof_close (); }
+};
+
+static QofBook*
+session_load (QofSession* session, const char* filename)
+{
+    if (!session || !filename) return nullptr;
+
+    qof_session_begin (session, filename, SESSION_READ_ONLY);
+    if (qof_session_get_error(session) != 0)
+    {
+        std::cerr << "Session begin failed: " << qof_session_get_error_message(session);
+        return nullptr;
+    }
+
+    qof_session_load(session, nullptr);
+    if (qof_session_get_error(session) != 0)
+    {
+        std::cerr << "Session begin failed: " << qof_session_get_error_message(session);
+        return nullptr;
+    }
+
+    return qof_session_get_book(session);
+}
+
+TEST_F(LoadFile, LoadAndVerifyKVP)
+{
+    auto location{g_getenv ("GNC_TEST_FILES")};
+    if (!location)
+        location = "test-files/load-save";
+    auto filename{"xml-datafile.gnucash"};
+    std::shared_ptr<gchar> to_open{g_build_filename (location, filename, (gchar*)nullptr), g_free};
+    std::shared_ptr<QofSession> session{qof_session_new(qof_book_new()), qof_session_destroy};
+    auto book{session_load (session.get(), to_open.get())};
+    ASSERT_NE(book, nullptr);
+
+    auto bank_acct = gnc_account_lookup_by_name (gnc_book_get_root_account (book), "Bank");
+    ASSERT_TRUE (bank_acct != nullptr);
+
+    // THE FOLLOWING TESTS BANK ACCOUNT
+    const auto& bank_splitlist = xaccAccountGetSplits (bank_acct);
+    ASSERT_EQ (bank_splitlist.size(), static_cast<uint>(4));
+
+    // first split is from a regular transaction
+    auto bank_reg_split{bank_splitlist[0]};
+    auto bank_reg_txn{xaccSplitGetParent(bank_reg_split)};
+    EXPECT_STREQ (xaccTransGetDescription (bank_reg_txn), "income");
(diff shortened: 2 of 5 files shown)


jralls:
ISTM this would be better as an extension of gtest-load-save-files.cpp with the elements added to test/test-files/load-save/sample.gnucash.

The directories test-files/xml1 and xml2 contain files in the old (pre-GnuCash 2.0) and current (GnuCash 2.0 and later) schemas respectively. An xml3 directory won't be appropriate unless we adopt a new xml schema.

christopherlam:
Seems the original tests compression/decompression sizes. Not sure why adding 1 additional file means the whole thing fails. @nomis 

With this PR complete with all setters&getters from #2107 added, I'll have much greater confidence that #2107 will still read/write the same kvp data.

nomis:
> Seems the original tests compression/decompression sizes. Not sure why adding 1 additional file means the whole thing fails. @nomis

There was a [bug](https://bugs.gnucash.org/show_bug.cgi?id=798967) where saving with compression always fails, which is why I added the test.

It includes checks that saving the file with and without compression results in the same data:
```
    /* Verify that we can write a compressed version of the original file that
     * has the original content when uncompressed.
     */
    auto new_compressed_file = filename + "-test-compressed~";
    /* Verify that we can read a compressed file and write an uncompressed file
     * that has the original content.
     */
    auto new_uncompressed_file = filename + "-test-uncompressed~";
```

The first step in checking this is that the file size matches, before it checks that the content matches.

It's unavoidable that when GnuCash is changed to write different output, the test will fail and the [README](https://github.com/Gnucash/gnucash/blob/stable/libgnucash/backend/xml/test/test-files/load-save/README) in the test directory explains how to fix it. It always requires manual developer intervention to verify that the change in uncompressed XML is intended.

libgnucash/backend/xml/test/gtest-load-save-files.cpp
```diff
@@ -319,3 +320,68 @@ std::vector<std::string> ListTestCases ()
...
+TEST_F(LoadSaveFiles, LoadAndVerifyKVP)
+{
+    auto location{g_getenv ("GNC_TEST_FILES")};
+    if (!location)
+        location = "test-files/load-save";
+
+    std::shared_ptr<QofSession> session{qof_session_new(qof_book_new()), qof_session_destroy};
+    ASSERT_NE(session.get(), nullptr);
+
+    auto filename{"xml-datafile.gnucash"};
```
nomis: This needs a better name reflecting what its purpose is because all of the files are "XML data files" so this is too generic.

libgnucash/backend/xml/test/gtest-load-save-files.cpp
```diff
@@ -319,3 +320,68 @@ std::vector<std::string> ListTestCases ()
     EXPECT_FALSE(files.empty()) << "no files found in " << location;
     return files;
 }
+
+TEST_F(LoadSaveFiles, LoadAndVerifyKVP)
```
nomis: I think you can remove this test completely. There's already a test that checks that the file can be loaded and then saved to get the same content. You don't need to check anything individually because if GnuCash understood the input correctly then the output will be the same (e.g. all `bug-746937-sx-split-order.gnucash` is doing is making sure that splits are saved in the same order each time).

Something unusual happens when this test is added, parts of all the output  files get duplicated:
```
$ diff -U4 xml-datafile.gnucash <(zcat xml-datafile.gnucash-test-compressed~)
--- xml-datafile.gnucash	2025-06-29 09:29:03.964840649 +0100
+++ /dev/fd/63	2025-06-29 09:41:05.344780112 +0100
@@ -27,8 +27,22 @@
      xmlns:order="http://www.gnucash.org/XML/order"
      xmlns:owner="http://www.gnucash.org/XML/owner"
      xmlns:taxtable="http://www.gnucash.org/XML/taxtable"
      xmlns:tte="http://www.gnucash.org/XML/tte"
+     xmlns:vendor="http://www.gnucash.org/XML/vendor"
+     xmlns:addr="http://www.gnucash.org/XML/addr"
+     xmlns:billterm="http://www.gnucash.org/XML/billterm"
+     xmlns:bt-days="http://www.gnucash.org/XML/bt-days"
+     xmlns:bt-prox="http://www.gnucash.org/XML/bt-prox"
+     xmlns:cust="http://www.gnucash.org/XML/cust"
+     xmlns:employee="http://www.gnucash.org/XML/employee"
+     xmlns:entry="http://www.gnucash.org/XML/entry"
+     xmlns:invoice="http://www.gnucash.org/XML/invoice"
+     xmlns:job="http://www.gnucash.org/XML/job"
+     xmlns:order="http://www.gnucash.org/XML/order"
+     xmlns:owner="http://www.gnucash.org/XML/owner"
+     xmlns:taxtable="http://www.gnucash.org/XML/taxtable"
+     xmlns:tte="http://www.gnucash.org/XML/tte"
      xmlns:vendor="http://www.gnucash.org/XML/vendor">
 <gnc:count-data cd:type="book">1</gnc:count-data>
 <gnc:book version="2.0.0">
 <book:id type="guid">f70b5dc6711143939c0dd7ae25e2e773</book:id>
@@ -69,8 +83,11 @@
 <gnc:count-data cd:type="transaction">3</gnc:count-data>
 <gnc:count-data cd:type="gnc:GncCustomer">1</gnc:count-data>
 <gnc:count-data cd:type="gnc:GncEntry">1</gnc:count-data>
 <gnc:count-data cd:type="gnc:GncInvoice">1</gnc:count-data>
+<gnc:count-data cd:type="gnc:GncCustomer">1</gnc:count-data>
+<gnc:count-data cd:type="gnc:GncEntry">1</gnc:count-data>
+<gnc:count-data cd:type="gnc:GncInvoice">1</gnc:count-data>
 <gnc:commodity version="2.0.0">
   <cmdty:space>CURRENCY</cmdty:space>
   <cmdty:id>AUD</cmdty:id>
   <cmdty:get_quotes/>
@@ -316,8 +333,96 @@
 <gnc:GncCustomer version="2.0.0">
   <cust:guid type="guid">2e8b6716d9cb4d70b8b7bea13315f939</cust:guid>
   <cust:name>customer-name</cust:name>
   <cust:id>000001</cust:id>
+  <cust:addr version="2.0.0">
+    <addr:name>billing-name</addr:name>
+    <addr:addr1>addr1</addr:addr1>
+    <addr:addr2>addr2</addr:addr2>
+    <addr:addr3>addr3</addr:addr3>
+    <addr:addr4>addr4</addr:addr4>
+    <addr:phone>cust-phone</addr:phone>
+    <addr:fax>cust-fax</addr:fax>
+    <addr:email>cust-email</addr:email>
+  </cust:addr>
+  <cust:shipaddr version="2.0.0"/>
+  <cust:notes>cust-notes</cust:notes>
+  <cust:taxincluded>USEGLOBAL</cust:taxincluded>
+  <cust:active>1</cust:active>
+  <cust:discount>0/1</cust:discount>
+  <cust:credit>0/1</cust:credit>
+  <cust:currency>
+    <cmdty:space>CURRENCY</cmdty:space>
+    <cmdty:id>AUD</cmdty:id>
+  </cust:currency>
+  <cust:use-tt>0</cust:use-tt>
+  <cust:slots>
+    <slot>
+      <slot:key>last-posted-to-acct</slot:key>
+      <slot:value type="guid">0656224544884e8e8196f23af8e67a1e</slot:value>
+    </slot>
+    <slot>
+      <slot:key>payment</slot:key>
+      <slot:value type="frame">
+        <slot>
+          <slot:key>last_acct</slot:key>
+          <slot:value type="guid">0d0c14d1661b46819c18319c21dc2666</slot:value>
+        </slot>
+      </slot:value>
+    </slot>
+  </cust:slots>
+</gnc:GncCustomer>
+<gnc:GncEntry version="2.0.0">
+  <entry:guid type="guid">81bb7b390a6240b4bab220f1bed4ebbb</entry:guid>
+  <entry:date>
+    <ts:date>2025-02-01 04:00:00 +0000</ts:date>
+  </entry:date>
+  <entry:entered>
+    <ts:date>2025-06-28 15:03:18 +0000</ts:date>
+  </entry:entered>
+  <entry:description>entry-desc</entry:description>
+  <entry:qty>1/1</entry:qty>
+  <entry:i-acct type="guid">5fb86abd10e94abba7463655767a5e63</entry:i-acct>
+  <entry:i-price>2/1</entry:i-price>
+  <entry:i-discount>3/1</entry:i-discount>
+  <entry:invoice type="guid">2d4e6dbfba6949abb6e4e1b0fe0ad54a</entry:invoice>
+  <entry:i-disc-type>PERCENT</entry:i-disc-type>
+  <entry:i-disc-how>PRETAX</entry:i-disc-how>
+  <entry:i-taxable>0</entry:i-taxable>
+  <entry:i-taxincluded>0</entry:i-taxincluded>
+</gnc:GncEntry>
+<gnc:GncInvoice version="2.0.0">
+  <invoice:guid type="guid">2d4e6dbfba6949abb6e4e1b0fe0ad54a</invoice:guid>
+  <invoice:id>000001</invoice:id>
+  <invoice:owner version="2.0.0">
+    <owner:type>gncCustomer</owner:type>
+    <owner:id type="guid">2e8b6716d9cb4d70b8b7bea13315f939</owner:id>
+  </invoice:owner>
+  <invoice:opened>
+    <ts:date>2025-02-01 10:59:00 +0000</ts:date>
+  </invoice:opened>
+  <invoice:posted>
+    <ts:date>2025-02-01 10:59:00 +0000</ts:date>
+  </invoice:posted>
+  <invoice:active>1</invoice:active>
+  <invoice:posttxn type="guid">372e0950eebc4bc0982768f6ad920c99</invoice:posttxn>
+  <invoice:postlot type="guid">73ab1ddf8d4943799c6705e1813ae582</invoice:postlot>
+  <invoice:postacc type="guid">0656224544884e8e8196f23af8e67a1e</invoice:postacc>
+  <invoice:currency>
+    <cmdty:space>CURRENCY</cmdty:space>
+    <cmdty:id>AUD</cmdty:id>
+  </invoice:currency>
+  <invoice:slots>
+    <slot>
+      <slot:key>credit-note</slot:key>
+      <slot:value type="integer">0</slot:value>
+    </slot>
+  </invoice:slots>
+</gnc:GncInvoice>
+<gnc:GncCustomer version="2.0.0">
+  <cust:guid type="guid">2e8b6716d9cb4d70b8b7bea13315f939</cust:guid>
+  <cust:name>customer-name</cust:name>
+  <cust:id>000001</cust:id>
   <cust:addr version="2.0.0">
     <addr:name>billing-name</addr:name>
     <addr:addr1>addr1</addr:addr1>
     <addr:addr2>addr2</addr:addr2>
```

You're causing `SetUpTestSuite()` to be called twice which breaks the output because `cashobjects_register()` is called twice. If you remove this test then it passes.

nomis (changes requested):

libgnucash/backend/xml/test/gtest-load-save-files.cpp
christopherlam: > You don't need to check anything individually because if GnuCash understood the input correctly then the output will be the same (e.g. all `bug-746937-sx-split-order.gnucash` is doing is making sure that splits are saved in the same order each time).

Thanks for looking into it. My concern leading to this bug was subtle; an earlier iteration of #2107 incorrectly handled `xaccTransSetDateDue` for example and would store/retrieve an incorrect kvp format; getter/setter tests would pass, they would only fail if a crafted datafile contained the `TRANS_DATE_DUE_KVP` kvp that was incorrectly read. Hence I'm still keen to add individual getter tests.

> You're causing `SetUpTestSuite()` to be called twice which breaks the output because `cashobjects_register()` is called twice. If you remove this test then it passes.

Hmm then `cashobjects_register` need to return early if called twice?

nomis: If you're testing round-trip of load/save, put it in this test suite without a second test case.

If you're testing something invalid in the XML you'll need a separate test suite. Presumably that one doesn't write the same content that it reads?

christopherlam: > If you're testing round-trip of load/save, put it in this test suite without a second test case.

Yes this new test is intended to test roundtrip load&save, and ensure that the xml serialization is identical. e.g. some bool are stored as string `true`, others are stored as integer `1`. We wouldn't want any engine refactoring to accidentally switch kvp bool from string to integer, which would mean old datafiles would suddenly find bools being `false`.

nomis: Then it's not a _new test_ but a _new test xml file_.

Put your getter tests in a separate test suite with a copy of the file. This one isn't designed to keep files in older formats forever.

christopherlam: Thanks, I've remove from your test suite completely.

PR #2110, opened 2025-07-07
Chartjs 4

christopherlam:
Thanks to @richardcohen for the heavy lifting in #1530 

~By default it picks chartjs-2 with WEBKIT1 or chartjs-4 otherwise. Can be overridden via `GNC_MODERN_JS=1` env variable.~ Thanks to webkit upgrade we can now cleanly upgrade to chartjs4

~Introduces `gnc_has_modern_javascript()` returning TRUE/FALSE.~

~Future:~
- ~any future reports which require chartjs-4 can be enabled only if chartjs4 is available.~
- ~maybe, refactor common code in html-chartjs-2 and html-chartjs-4 to reduce duplication~


gnucash/report/gnc-report.cpp
```diff
@@ -450,3 +450,16 @@ gnc_get_optiondb_from_dispatcher(SCM dispatcher)
...
+    static const char* env = g_getenv ("GNC_MODERN_JS");
+    if (env)
+        return *env == '1';
+
+#ifdef WEBKIT1
+    return FALSE;
+#else
+    return TRUE;
+#endif
+}
```
jralls: 1. Testing whether sufficiently new Javascript is available belongs in the root CMakeLists.txt.
2. WebKitGtk 2.14 has WebKit2, it just doesn't build on Win32. It'll still have the same obsolete Javascript. The best way to test is to test for some Javascript feature(s) that Chartjs 4 needs and set the variable based on whether it works. Failing that you need to at least figure out what is the earliest version of WebKitGtk that works and test for that. 
3. Your CMake test will set the preprocessor macro `GNC_MODERN_JS`. Since you're selecting what to call in Scheme I'd think that `gnc_has_modern_javascript` could go in report.i and return `SCM_TRUE` or `SCM_FALSE`, avoiding the typemap overhead.

jralls (changes requested):
Mentioning #1530 since it's the source of html-chartjs-4.scm

You still need to add chartjs4 to borrowed.

gnucash/report/gnc-report.cpp
christopherlam: I don't think it's possible to test javascript feature from C or guile. `html-chart` will build the js string into the html to be rendered in webkit2. So it's possible only to test WEBKIT1|2 and pick chartjs2|4 appropriately. I don't have the manpower to test Webkit versions either I'm afraid. So it'll be either `#ifdef WEBKIT` or `#ifdef G_OS_WIN32`.

jralls: Not only possible, not even hard. Make a [JavaScriptCore::Context](https://webkitgtk.org/reference/jsc-glib/2.43.4/class.Context.html) ([Non-Gtk-specific docs](https://docs.webkit.org/Deep%20Dive/JSC/JavaScriptCore.html)) and pass it the javascript function test. [`JavaScriptCore::Context::check_syntax`](https://webkitgtk.org/reference/jsc-glib/2.43.4/method.Context.check_syntax.html) might work, but I think you're more likely to need to use [`JavaScriptCore::Context::evaluate`](https://webkitgtk.org/reference/jsc-glib/2.43.4/method.Context.evaluate.html) and then check [`JavaScriptCore::Context::get_exception`](https://webkitgtk.org/reference/jsc-glib/2.43.4/method.Context.get_exception.html) to find out if it worked.

ECMAScript is the official standard name for JavaScript, and Guile can run [ECMAscript directly](https://www.gnu.org/software/guile/manual/html_node/ECMAScript.html), but of course that's not going to do us any good for chartjs.

> I don't have the manpower to test Webkit versions either I'm afraid. So it'll be either #ifdef WEBKIT or #ifdef G_OS_WIN32.

Wrong answer in two ways. First, this is a mostly pointless change anyway because it doesn't improve GnuCash for the majority of users, so there's no urgency and no reason to take on the additional technical debt. Second, we're [already testing](https://github.com/Gnucash/gnucash/blob/2b0c90304e96b7a8ca3f18f970ae9c23ac98f380/CMakeLists.txt#L231) webkit versions and `#ifdef WEBKIT` isn't a thing because GnuCash won't even build if there's no WebKit.

jralls: #2275 takes care of the obsolete javascript problem. 

Separate issue: Since we seem to need only two files it would perhaps make more sense to fetch just those from the chartjs repo with Cmake's [`FetchContent`](hhttps://cmake.org/cmake/help/latest/module/FetchContent.html) or [`file(DOWNLOAD)`](https://cmake.org/cmake/help/latest/command/file.html) instead of copying the entire repository snapshot into our `borrowed` directory. Apparently the chartjs folks have removed the build products from their repository so they [recommend](https://www.chartjs.org/docs/latest/getting-started/installation.html) getting built files from [CDNJS](https://cdnjs.com/libraries/Chart.js).

christopherlam: > #2275 takes care of the obsolete javascript problem.
> 
> Separate issue: Since we seem to need only two files it would perhaps make more sense to fetch just those from the chartjs repo with Cmake's `FetchContent` or [`file(DOWNLOAD)`](https://cmake.org/cmake/help/latest/command/file.html) instead of copying the entire repository snapshot into our `borrowed` directory. Apparently the chartjs folks have removed the build products from their repository so they [recommend](https://www.chartjs.org/docs/latest/getting-started/installation.html) getting built files from [CDNJS](https://cdnjs.com/libraries/Chart.js).

Does this mean that building from source will require installed packages, _and_ access to the wider Internet to relieve the latest and greatest chartjs?

jralls: > Does this mean that building from source will require installed packages, and access to the wider Internet to relieve the latest and greatest chartjs?

Depends. If you mean [building Chartjs from source](https://www.chartjs.org/docs/latest/developers/contributing.html#building-and-testing), yes, that will require installing the [pnpm package manager](https://www.chartjs.org/docs/latest/developers/contributing.html#building-and-testing).
We can avoid that effort by retrieving the built chart.js and chart.umd.js from npm (via the CDNJS links from [chartjs installation](https://www.chartjs.org/docs/latest/getting-started/installation.html#cdnjs). There are three options: Keeping them in our borrowed directory, retrieving at build time via Cmake, or linking directly in html `script` tags. 

Note that npm has a history of supply chain vulnerabilities, the latest of which was [discovered just last Tuesday](https://www.microsoft.com/en-us/security/blog/2026/08/04/chaindrop-supply-chain-compromise-anatomy-self-propagating-worm/). Building chart.js from source doesn't help, one must install a [bunch of packages](https://github.com/chartjs/Chart.js/blob/master/pnpm-lock.yaml) from npm just to build it.

Maybe the safest way to stay up to date on chartjs is to periodically pull the two files from CDNJS and sit on them for a couple of weeks while monitoring for malware reports. If none come up we can merge them into borrowed/chartjs.

christopherlam: +1 I think getting the compiled js is the best approach here to prevent surprises.

christopherlam:
my `future` branch isn't compilable locally (my glib too old!)...

jralls:
> my future branch isn't compilable locally (my glib too old!)...

Probably more than that. Future needs one of the latest distros, i.e. Ubuntu 26.04 or Fedora 46. You don't have to upgrade your machine, you can set up a VM or a Docker.

gjanssens:
> > my future branch isn't compilable locally (my glib too old!)...
> 
> Probably more than that. Future needs one of the latest distros, i.e. Ubuntu 26.04 or Fedora 46. You don't have to upgrade your machine, you can set up a VM or a Docker.

Or perhaps a [toolbox](https://containertoolbx.org/) container. I find them easier to work with than docker for development. Docker containers are typically more geared towards running services. Toolbox containers (while based on podman technology) offer isolated workstations at your fingertips, with ability to run gui applications and access to the useful bits of your host system (most usefully your HOME directory). It is tested and available on arch, fedora, ubuntu en rhel, and can create containers for each of those platforms as well. Which means it's very easy to have a container running ubuntu 26.04 on a Fedora 44 host, or a Fedora 46 container on an Ubuntu 24.04 LTS for example. Quite nice for buiding an running gnucash on multiple distributions and versions on one workstation.

(Note I am not affiliated with toolbox, I just enjoy using this tool)

jralls:
@christopherlam a rebase should fix the windows tests.

christopherlam:
upgrade chartjs-4 including html-chart changes from #2304

christopherlam:
I've tested this branch by applying it into `stable` and all seems OK.

PR #2112, opened 2025-07-15
D3 sankey

christopherlam:
For entertainment. Has enormous tech debt! :open_mouth: 

- uses an old D3.js v3 which is still ES5 compatible, increases chances of WIN32 compatibility
- uses an old D3-sankey which is compatible with D3v3
- `html-sankey` offers a thin wrapper similar to `html-chart`
- `cashflow-sankey` uses `html-sankey`

Screenshot 2025-07-15 at 23-23-35 Cash Flow Sankey Chart


jralls:
```
 Syntax error:
standard/cashflow-sankey.scm:132:4: definition in expression context, where definitions are not allowed, in form (define all-flows (hash-map->list cons totals-hash))
```

That aside, the lack of color in the screenshot makes it impossible to interpret the chart. Is that an artifact or a limitation from the antiquity of the dependencies?

christopherlam:
> That aside, the lack of color in the screenshot makes it impossible to interpret the chart. Is that an artifact or a limitation from the antiquity of the dependencies?

Dunno... what colour is missing? It's likely customisable via the js string.

jralls:
Most sankey charts I see have colors for the source categories to make it easier to see how much of which source end up at each sink. I thought the #759 chart was like that but when I looked at it just now I see that it only has gradients in place of the blue bars in this one. Prettier but not more usable.

jralls:
@christopherlam can we close this in favor of #2281 or do you prefer a strictly chartjs approach in light of #2275?

christopherlam:
> @christopherlam can we close this in favor of #2281 or do you prefer a strictly chartjs approach in light of #2275?

I'm leaning towards reusing industry standard tooling rather than a pure scheme Sankey layout engine... Easier to let others lead development, and they'll have much better documentation.

jralls:
> I'm leaning towards reusing industry standard tooling rather than a pure scheme Sankey layout engine... Easier to let others lead development, and they'll have much better documentation.

But can you get the colors to work with the chartjs-based Sankey chart? 

As for industry standard, does that mean you're in favor of (eventually) rewriting the reports module in Javascript or (better) TypeScript?

christopherlam:
> > I'm leaning towards reusing industry standard tooling rather than a pure scheme Sankey layout engine... Easier to let others lead development, and they'll have much better documentation.
> 
> But can you get the colors to work with the chartjs-based Sankey chart?

I haven't touched the Sankey charts for a couple years but I imagine they should be customisable.

> 
> As for industry standard, does that mean you're in favor of (eventually) rewriting the reports module in Javascript or (better) TypeScript?

Not much time I'm afraid. If there is a definite standardised set of accountancy reports in JavaScript, then presumably it should be possible to create an engine->glue->js mechanism, which could run in parallel to the current scheme ones.

PR #2113, opened 2025-07-17
Better UX - instead of Yes/No or OK/Cancel, offer Action/Cancel choices

christopherlam:
I'd postulate that most `gnc_verify_dialog` could use `gnc_action_dialog` to have better UX. Example: "Do you want to revert?" changed from Yes/No to Cancel/Revert.

Downisde: users with muscle memory may object.

### DIFF
--- a/gnucash/gnome-utils/dialog-tax-table.c
+++ b/gnucash/gnome-utils/dialog-tax-table.c
@@ -733,7 +733,7 @@ tax_table_delete_table_cb (GtkButton *button, TaxTableWindow *ttw)
         return;
     }
 
-    if (gnc_verify_dialog (GTK_WINDOW(ttw->dialog), FALSE,
+    if (gnc_action_dialog (GTK_WINDOW(ttw->dialog), _("_Delete"), FALSE,
                            _("Are you sure you want to delete \"%s\"?"),
                            gncTaxTableGetName (ttw->current_table)))
     {
@@ -780,7 +780,7 @@ tax_table_delete_entry_cb (GtkButton *button, TaxTableWindow *ttw)
         return;
     }
 
-    if (gnc_verify_dialog (GTK_WINDOW(ttw->dialog), FALSE, "%s",
+    if (gnc_action_dialog (GTK_WINDOW(ttw->dialog), _("_Delete"), FALSE, "%s",
                            _("Are you sure you want to delete this entry?")))
     {
         /* Ok, let's remove it */
--- a/gnucash/gnome-utils/gnc-file.c
+++ b/gnucash/gnome-utils/gnc-file.c
@@ -315,7 +315,7 @@ show_session_error (GtkWindow *parent,
     case ERR_BACKEND_NO_SUCH_DB:
         fmt = _("The database %s doesn't seem to exist. "
                 "Do you want to create it?");
-        if (gnc_verify_dialog (parent, TRUE, fmt, displayname))
+        if (gnc_action_dialog (parent, _("_Create"), TRUE, fmt, displayname))
         {
             uh_oh = FALSE;
         }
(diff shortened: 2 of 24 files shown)


jralls:
You can avoid the muscle memory problem by keeping the affirmative on the left even though having it on the right is customary for Cancel/OK dialogs.

The short response strings need translation context: Imperative verbs take a different form in many languages. Alternatively you could make the strings a little longer so that the imperative intent is clear, e.g. "Revert the changes" or "Delete the configuration".

christopherlam:
> You can avoid the muscle memory problem by keeping the affirmative on the left even though having it on the right is customary for Cancel/OK dialogs.

Seems windows has affirmative on left but Mac has affirmative on right. 

https://www.nngroup.com/articles/ok-cancel-or-cancel-ok/

> The short response strings need translation context: Imperative verbs take a different form in many languages. Alternatively you could make the strings a little longer so that the imperative intent is clear, e.g. "Revert the changes" or "Delete the configuration".

Yeah do we make multiple 3 word short statements, or lots of single word verb strings with context?

jralls:
> Seems windows has affirmative on left but Mac has affirmative on right.

That's immaterial to muscle memory for GnuCash users. `gnc_verify_dialog` has affirmative on the left so long time users' muscle memory is going to click left. 

> Yeah do we make multiple 3 word short statements, or lots of single word verb strings with context?

I think big wordy buttons are weird so I'd prefer single-word-with-context.

jralls (commented):
Is your goal is to replace all instances of `gnc_verify_dialog`. `grep -rl gnc_verify_dialog gnucash | wc -l` returns 29 and this covers 16, so halfway there. If that's not your goal is there some criterion for where you prefer `gnc_action_dialog`?

BTW, `gnc_verify_dialog` is wrapped in gnome-utils.i but AFAICT isn't used.

christopherlam:
> Is your goal is to replace all instances of `gnc_verify_dialog`. `grep -rl gnc_verify_dialog gnucash | wc -l` returns 29 and this covers 16, so halfway there. If that's not your goal is there some criterion for where you prefer `gnc_action_dialog`?

The goal is to complete half, check the desired wording, and continue ;)

There are some dialogs that I don't know how to change... `gnc_verify_dialog("Are you sure you wish to cancel the slow task?")` with Yes/No _would_ be changed to `gnc_action_dialog(...)` with Cancel/Continue, whereby _Cancel_ means cancel the cancellation and _Continue_ means proceed with cancellation would be very confusing...

christopherlam:
`gnc_action_dialog` works well, what remains is whether there's any wordsmithing needed.

jralls:
> There are some dialogs that I don't know how to change... gnc_verify_dialog("Are you sure you wish to cancel the slow task?") with Yes/No would be changed to gnc_action_dialog(...) with Cancel/Continue, whereby Cancel means cancel the cancellation and Continue means proceed with cancellation would be very confusing...

A case where "Yes" and "No" are clearer.

Are there any cases where "Yes" and "No" are unclear? I'm not aware of anyone ever complaining about this on any uses of `gnc_verify_dialog` so maybe it's a solution in search of a problem.

christopherlam:
> Are there any cases where "Yes" and "No" are unclear? I'm not aware of anyone ever complaining about this on any uses of `gnc_verify_dialog` so maybe it's a solution in search of a problem.

Agreed it's not a necessary change, however, it would seem to be the best practice going forward. This PR hesitantly suggests whether there's any need for modifying current `gnc_verify_dialog` (Yes/No) or `gnc_ok_cancel_dialog` (OK/Cancel).

jralls:
Why do you think it's a best practice to avoid "Yes" and "No" responses? Does anybody's HIG say to do that?

christopherlam:
https://developer.apple.com/design/human-interface-guidelines/alerts

https://learn.microsoft.com/en-us/windows/apps/design/controls/dialogs-and-flyouts/dialogs

Both encourage verbs

PR #2114, opened 2025-07-27
Bug 799650 - Reconcile Ending Balance Ignored When Statement Date Set

agwekixj:
Connected an activate callback to the Ending Balance entry widget to handle the value when the Enter key is used.


KevinHHobbs:
GnuCash built from this branch does not exhibit the bug on Fedora 42.

jralls:
Thanks!

PR #2116, opened 2025-08-04
GdPDU Export

jhs-s:
German tax authorities expect an standardized export of the electronic accounting in a Format called GoBD or GdPDU. This is essentially an CSV Format which can contain multiple files and an index.xml file describing those files. The format is rather flexible but it still expects that a booking is always on Record.

This branch adds an Option for exporting all bookings in a certain CSV-Format and takes care that "multiple bookings" are exported as several single bookings. The standard CSV export cannot do that and the extended CSV export is not able to but a booking into a single line.

This involved a refactoring of the CSV-Export, creating a new class that takes care of the different options.

There might be open points as I could only test with my accounting data which might not be fully representative. It also cannot cope with multiple bookings that do not have a single origin account. In this case it will abort. However, normal invoice bookings with VAT shouldn't have any issues.

There is a (german) description here: https://wiki.gnucash.org/wiki/De/howto


jralls:
> [Merge branch 'master' into gdpdu-export](https://github.com/Gnucash/gnucash/pull/2116/commits/819a5e24e3f58c0d6a1f9fa264d5da1d5bcb3609)

Don't do that, it messes up history. Always rebase your feature branch on the main branch (in this case stable).

> [contrib: Add necessary files for full gdpdu-export](https://github.com/Gnucash/gnucash/pull/2116/commits/13bf6186567bf0ae722d3676372d36223c3e32aa)

I think these files would fit better in `data`.

gnucash/import-export/csv-exp/csv-export-helpers.cpp
```diff
@@ -38,56 +39,61 @@
...
+#define EOLSTR "\r\n"
 #endif
 
 #define QUOTE '"'
 
-bool
-gnc_csv_add_line (std::ostream& ss, const StringVec& str_vec,
-                  bool use_quotes, const char* sep)
+bool gnc_csv_add_line(std::ostream &ss, const StringVec &str_vec,
+                      bool use_quotes, const char *sep)
```
jralls, line 50: Please follow the [Coding Standard](https://wiki.gnucash.org/wiki/CodingStandard): In a definition the function name should start on a new line to make it easy to grep for, e.g. `grep -rl ^gnc_csv_add_line gnucash/import-export`. Declarations should have the return type, storage class, etc. on the same line as the function name.

gnucash/import-export/csv-exp/csv-export-helpers.cpp
```diff
@@ -38,56 +39,61 @@
...
+                ss << QUOTE;
+
+            if (ss.fail())
+                return false;
+        }
+        ss << EOLSTR;
+    }
+    catch (const std::exception &e)
+    {
+        std::cerr << e.what() << '\n';
```
jralls, line 86: GnuCash is primarily a GUI application that runs without a console so output to std::cout and std::cerr are lost; on Windows they're lost even with a console. We have a logging library that writes to a file instead with handy macros `PERR`, `PWARN`, `PINFO`, and `PDEBUG`. In this case `PWARN` seems appropriate: `PWARN("Exception %s", e.what(). The macro will add a timestamp and the name of the function.

gnucash/import-export/csv-exp/csv-transaction-export-line.cpp
```diff
@@ -0,0 +1,376 @@
+/*******************************************************************\
+ * csv-transactions-export-line.cpp -- Convert Transaction to line    *
+ *                                                                  *
+ * Copyright (C) 2025 Johannes Triegel                                *
+ *                                                                  *
```
jralls, line 5: Please fix up the whitespace so the ending `*`s line up.

gnucash/import-export/csv-exp/csv-transaction-export-line.cpp
```diff
@@ -0,0 +1,376 @@
...
+                                                   std::ofstream &ss) : m_split(split),
+                                                                        m_transaction(transaction),
+                                                                        m_separator(separator),
+                                                                        m_use_quotes(use_quotes),
+                                                                        m_simple(simple),
+                                                                        m_ss(ss),
+                                                                        m_is_debit_split(false),
+                                                                        m_is_credit_split(false),
+                                                                        m_base_split(NULL),
+                                                                        m_base_split_account(NULL)
```
jralls: Please start the initializers on a new line and put several on a line to cut down on the empty real estate.

gnucash/import-export/csv-exp/csv-transaction-export-line.cpp
```diff
@@ -0,0 +1,376 @@
...
+                debit++;
+            }
+            else if (gnc_numeric_negative_p(amount))
+            {
+                credit++;
+            }
+        }
+        if (debit > 1 && credit > 1)
+        {
+            std::cerr << "Multi split booking that needs to be booked on an intermediate account, no implemented!" << std::endl;
```
jralls: Use `PINFO` instead of `std::cerr`.

Also a typo: "not implemented" instead of "no implemented".

gnucash/import-export/csv-exp/csv-transactions-export.cpp
```diff
@@ -261,99 +58,96 @@ using TransSet = std::unordered_set<Transaction*>;
...
-                     std::ofstream& ss, TransSet& trans_set)
+export_query_splits(CsvExportInfo *info, bool is_trading_acct,
+                    std::ofstream &ss, TransSet &trans_set)
 {
-    g_return_if_fail (info);
+    g_return_if_fail(info);
 
     /* Run the query */
-    for (GList *splits = qof_query_run (info->query); !info->failed && splits;
+    for (GList *splits = qof_query_run(info->query); /*!info->failed && */ splits;
```
jralls, line 67: Don't comment out code, remove it.

This should be a separate commit with the commit message explaining why you don't want to abort the loop on a `print_csv` failure.

gnucash/import-export/csv-exp/csv-transactions-export.cpp
```diff
@@ -261,99 +58,96 @@ using TransSet = std::unordered_set<Transaction*>;
...
-        auto trans{xaccSplitGetParent (split)};
+        auto split{static_cast<Split *>(splits->data)};
+        auto trans{xaccSplitGetParent(split)};
 
         // Look for trans already exported in trans_set
-        if (!trans_set.emplace (trans).second)
+        if (!trans_set.emplace(trans).second)
+        {
             continue;
+        }
```
jralls, line 77: Don't add braces around single-line conditional or loop bodies..

gnucash/import-export/csv-exp/csv-transactions-export.cpp
```diff
@@ -261,99 +58,96 @@ using TransSet = std::unordered_set<Transaction*>;
...
         }
 
-        // Write complex Transaction Line.
-        auto line = make_complex_trans_line (trans, split);
-        info->failed = !gnc_csv_add_line (ss, line, info->use_quotes,
-                                          info->separator_str);
+        CsvTransactionExportLine line(split, trans, info->separator_str, info->use_quotes, info->simple_layout, ss);
+        info->failed = !line.print_csv();
 
+#if 0
```
jralls: `#if 0` is the same as comments: Just remove the code.

jralls (changes requested):
Out of time for now, and I got through only one commit. More later.

gnucash/import-export/csv-exp/csv-transaction-export-line.hpp
```diff
@@ -0,0 +1,90 @@
...
+    CsvTransactionExportLine(Split *split,
+                             Transaction *transaction,
+                             const char *separator,
+                             bool use_quotes,
+                             bool simple,
+                             std::ofstream &ss);
+
+    bool print_csv();
+
+protected:
```
jralls, line 51: `protected` is to allow access to derived classes. `CsvTransactionExportLine` isn't an abstract class--it doesn't even have any virtual members--so there should be no `protected` members, only `private` and `public`. These are all getters so there's no good reason for them to be `private`.

gnucash/import-export/csv-exp/csv-transaction-export-line.hpp
```diff
@@ -0,0 +1,90 @@
...
+    StringVec make_simple_trans_line(Split *split);
+
+private:
+    Split* m_split;
+    Transaction *m_transaction;
+
+    const char *m_separator;
+    bool m_use_quotes;
+    bool m_simple;
+    std::ofstream &m_ss;
```
jralls: These should be passed as parameters to the relevant function, not held as object state.

gnucash/import-export/csv-exp/csv-transaction-export-line.cpp
jralls: This is C++. `NULL` is spelled `nullptr`.

gnucash/import-export/csv-exp/csv-transaction-export-line.cpp
```diff
@@ -0,0 +1,376 @@
...
+        {
+            m_is_debit_split = true;
+        }
+        else
+        {
+            m_is_credit_split = true;
+        }
+
+        // Find the base_split
+        Split *m_base_split = NULL;
```
jralls, line 97: You already did that in the initializers.

gnucash/import-export/csv-exp/csv-transactions-export.cpp
```diff
@@ -42,217 +42,14 @@
...
-        get_value (split, t_void, true),
-        get_value (split, t_void, false),
-        get_reconcile (split),
-        get_reconcile_date (split),
-        get_price (split, t_void)
-    };
-}
-
-using TransSet = std::unordered_set<Transaction*>;
+using TransSet = std::unordered_set<Transaction *>;
```
jralls, line 52: That's innapropriate whitespace. `Transaction*` is a C++ type. It has been C++ convention basically forever that type names are kept together, though because C++'s backward compatibility with C allows a line like
  `Transaction* foo, bar;` 
would make `foo` a `Transaction*` and `bar` a `Transaction` it's necessary to either write 
`Transaction *foo, *bar;` with the pointer indicator stuck to the variables for consistency or
```
Transaction *foo;
Transaction *bar;
```
Stroustroup (Bjarne Stroustoup, The C++ Programming Language Fourth Edition, Addison Wesley, 2013, p 155) writes of the first way "Such declarations with multiple names and nontrivial declarators make a program harder to read and should be avoided." While GnuCash has a C heritage and most of the C++ files are largely renamed C code containing C conventions that's no reason to use C coding conventions in new C++ code.

gnucash/import-export/csv-exp/csv-transactions-export.cpp
```diff
@@ -261,99 +58,72 @@ using TransSet = std::unordered_set<Transaction*>;
...
-                (xaccAccountGetType (tsplit_acc) == ACCT_TYPE_TRADING))
-                continue;
-
-            // Write complex Split Line.
-            auto line = make_complex_trans_line (trans, t_split);
-            info->failed = !gnc_csv_add_line (ss, line, info->use_quotes,
-                                              info->separator_str);
-        }
+        CsvTransactionExportLine line(split, trans, info->separator_str, info->use_quotes, info->simple_layout, info->gdpdu_layout, is_trading_acct, ss);
+        info->failed = !line.print_csv();
```
jralls, line 94: That's it? All of this to create a class, call a single instance function on it, and discard it? What a waste.

gnucash/import-export/csv-exp/assistant-csv-export.h
```diff
@@ -74,13 +74,21 @@ typedef struct
 
     Query          *query;
 
+    GtkBuilder     *builder;
```
jralls, line 77: Why?

gnucash/import-export/csv-exp/csv-transaction-export-line.cpp
```diff
@@ -272,46 +293,92 @@ CsvTransactionExportLine::get_price(Split *split, bool t_void)
...
+                                  m_separator);
+            if (!ok)
+                break;
         }
-        else
+        return ok;
+    }
+    else
+    {
+        /* Loop through the list of splits for the Transaction */
```
jralls, line 346: Pointless comment. What else would `for (auto node = xaccTransGetSplitList…)` do?

gnucash/import-export/csv-exp/csv-transactions-export.cpp
```diff
@@ -261,99 +58,96 @@ using TransSet = std::unordered_set<Transaction*>;
...
             if (!is_trading_acct &&
-                (xaccAccountGetType (tsplit_acc) == ACCT_TYPE_TRADING))
+                (xaccAccountGetType(tsplit_acc) == ACCT_TYPE_TRADING))
                 continue;
 
             // Write complex Split Line.
-            auto line = make_complex_trans_line (trans, t_split);
-            info->failed = !gnc_csv_add_line (ss, line, info->use_quotes,
-                                              info->separator_str);
+            auto line = make_complex_trans_line(trans, t_split);
```
jralls: I guess you fix this later, but you deleted `make_complex_trans_line`  so this commit won't compile. That will break bisecting.

gnucash/import-export/csv-exp/csv-transaction-export-line.cpp
```diff
@@ -153,6 +157,23 @@ CsvTransactionExportLine::get_account_name(Split *split, bool full)
     return full ? account_get_fullname_str(account) : xaccAccountGetName(account);
 }
 
+// Account ode
```
jralls, line 162: Huh? An ode is a kind of poem. What's that got to do with anything?

gnucash/import-export/csv-exp/csv-transaction-export-line.hpp
```diff
@@ -68,16 +72,20 @@ class CsvTransactionExportLine
...
 
 private:
-    Split* m_split;
+    Split *m_split;
     Transaction *m_transaction;
 
     const char *m_separator;
     bool m_use_quotes;
     bool m_simple;
+    bool m_gdpdu;
```
jralls: Another case where an enum would be better than an accumulation of bools plus a hidden default when both are false.

gnucash/import-export/csv-exp/csv-transactions-export.cpp
```diff
@@ -168,6 +144,20 @@ void csv_transactions_export(CsvExportInfo *info)
             _("Rate/Price"),
         };
     }
+    /* Header string */
+    if (info->gdpdu_layout)
+    {
+        /* Translators: The following symbols will build the header
+           line of exported CSV files for german GdPDU-Export: */
```
jralls, line 151: That doesn't work: Each translatable string is presented separately on Weblate; translator comments are presented only when the very next line is a translatable string and this one would be applied to only "Date". Also all of these strings are already in use without context. On the one hand that's good because you don't have to worry about the translations. Just check de.po to make sure that the msgstrs for each one are correct for this case.

gnucash/import-export/csv-exp/csv-transaction-export-line.cpp
```diff
@@ -360,28 +372,46 @@ StringVec
...
 StringVec
 CsvTransactionExportLine::make_gdpdu_trans_line(Split *split)
 {
     auto t_void{xaccTransGetVoidStatus(m_transaction)};
+    auto amount = xaccSplitGetAmount(split);
+    auto account = xaccSplitGetAccount(split);
+    auto type = xaccAccountGetType(account);
+    bool pos = gnc_numeric_positive_p(amount);
+
+    /* It is possible that other types of accounts need to be taken into account here */
```
jralls, line 398: This looks like a FIXME: GnuCash's account types aren't tested hierarchically so even though `ACCT_TYPE_INCOME` and `ACCT_TYPE_EXPENSE` are Equity they won't set `exchange` true.

A comment explaining why the presence of an Equity account justifies switching the order of the account numbers in a single transaction would be useful.

gnucash/import-export/csv-exp/csv-transaction-export-line.cpp
```diff
@@ -379,26 +384,6 @@ CsvTransactionExportLine::make_gdpdu_trans_line(Split *split)
...
-        get_description(m_transaction),
-        pos ? get_account_name(split, true) : xaccAccountGetName(m_base_split_account),
-        get_reconcile(split),
-        get_amount(split, t_void, true),
-        get_amount(split, t_void, false),
-        get_value(split, t_void, true),
-        get_value(split, t_void, false),
-        get_rate(split, t_void)};
-}
-
```
jralls: This should be removed from the second commit instead of adding it there and throwing it away later.

gnucash/import-export/csv-exp/csv-transaction-export-line.cpp
```diff
@@ -249,9 +249,11 @@ CsvTransactionExportLine::get_commodity(Transaction *trans)
 
 // Amount with Symbol or not
 std::string
-CsvTransactionExportLine::get_amount(Split *split, bool t_void, bool symbol)
+CsvTransactionExportLine::get_amount(Split *split, bool t_void, bool symbol, bool positive)
```
jralls, line 252: Multiple boolean function args is poor design, it's too easy to screw up the order.

jralls (changes requested):
I don't see how `CsvTransactionExportLine` contributes anything except unnecessary overhead, and a major refactoring like this demands that you start by adding tests that prove the behavior of the existing code, then do the refactor with the tests still passing.

A PR shouldn't have "fix" commits in it: All of that should be cleaned up before submission.

But even if this were well designed and coded--and it isn't--I'd be hesitant to merge it because it's of use in only in Germany and we have no German core developers to maintain it.

jhs-s:
Thanks for all your comments and sorry for not replying - I was assuming that github would somehow send me Mails on comments.

I will try to resolve the issues...

jhs-s:
> > [Merge branch 'master' into gdpdu-export](https://github.com/Gnucash/gnucash/pull/2116/commits/819a5e24e3f58c0d6a1f9fa264d5da1d5bcb3609)
> 
> Don't do that, it messes up history. Always rebase your feature branch on the main branch (in this case stable).

Is there a clever way to do that without creating a new pull request. I guess I cannot change the existing history of the branch.

> > [contrib: Add necessary files for full gdpdu-export](https://github.com/Gnucash/gnucash/pull/2116/commits/13bf6186567bf0ae722d3676372d36223c3e32aa)
> 
> I think these files would fit better in `data`.

ok, no problem.

jralls:
> Is there a clever way to do that without creating a new pull request. I guess I cannot change the existing history of the branch.

You can change the branch history as much as you want back to the branch point, then force-push it (`git push -f github gdpdu-export` where `github` stands in for whatever you call this remote). That will update the PR, no new PR required.

PR #2120, opened 2025-08-29
Bug 799654 - Zombie entries in the recently accessed file list.

agwekixj:
An update to an existing menu entry should remove the menu entry.

### DIFF
--- a/gnucash/gnome-utils/gnc-plugin-file-history.c
+++ b/gnucash/gnome-utils/gnc-plugin-file-history.c
@@ -407,8 +407,6 @@ gnc_history_update_action (GncMainWindow *window,
     GncMenuModelSearch *gsm = g_new0 (GncMenuModelSearch, 1);
     gchar *action_name;
     gint limit;
-    gboolean add_item = FALSE;
-    gint pos;
 
     ENTER("window %p, index %d, filename %s", window, index,
           filename ? filename : "(null)");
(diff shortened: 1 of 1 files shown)


jralls:
I know that you did this backwards, creating the PR before [the bug](https://bugs.gnucash.org/show_bug.cgi?id=799654), but having created the bug you need to change the commit message so that fixing the bug shows up in the release notes.

Otherwise it looks fine and resolves the (somewhat obscure) problem. Too bad it turns out not to be what @christopherlam complained about in IRC.

agwekixj:
I wasn't aware of @christopherlam complaint at the time.  This was just something I tripped across while testing Chang Wang's average cost issue...  

Which reminds me, I haven't seen a bug posted on the issue.  S/he said something about not being able to create a BZ account.  Perhaps this was due to same issue that got me banned.

christopherlam:
> I wasn't aware of @christopherlam complaint at the time. This was just something I tripped across while testing Chang Wang's average cost issue...

My issue seems to be my `guix --pure` setup which is weird and I don't have gnucash glib schemas.

agwekixj:
> > I wasn't aware of @christopherlam complaint at the time. This was just something I tripped across while testing Chang Wang's average cost issue...
> 
> My issue seems to be my `guix --pure` setup which is weird and I don't have gnucash glib schemas.

I found your IRC post on 8/21 in the log.  The name used in the MRU list is the name in the file path value for the file# key.  The accessibility of a file in the MRU list is only determined later.

jralls:
> I don't have gnucash glib schemas.

I'm pretty sure GnuCash would refuse to start in that case. Even if it did not having the schemas would crash GnuCash as soon as it tried to read a preference. There may well be something off about one or more of your schemas, but glib is surely finding them.

PR #2121, opened 2025-09-02
Bug 799652 - Invoice price decimal places

agwekixj:
The "+ 2 places" should only be applied when force decimal is enabled.


christopherlam:
(just a commit style comment) I'd vote that these bugfix + test amendments should  be in the same commit to ensure future bisects are less painful?

agwekixj:
I don't have a "good" unit test environment but I could/should have collected test changes in a single commit.

jralls:
Thanks!

PR #2122, opened 2025-09-02
Bug 799656 - Invoice and bills still display prices as fractional amounts with "Force Prices to display as decimals"

agwekixj:
gncEntryLedgerModel.c:get_pric_entry() is changed to coerce the forced-decimal amount in to the price's display cell.

gncEntryLedger.c:gnc_entry_ledger_compute_value() is changed to pull the price from the ledger (instead of the cell).


jralls:
Thanks!

PR #2124, opened 2025-09-05
Bug 343711 - Splits to Imbalance-USD do not disappear when zero-valued.

agwekixj:
To mitigate potential impact, the Imbalance split is only removed when the split transitions to zero value during the commit of the transaction and the split's memo and action have no value.


jralls:
Thanks!

PR #2125, opened 2025-09-05
Get CI tests to pass on MSYS2/Win32 builds.

jralls:
Just as it says in the title.


gjanssens:
I have browsed through the changes, but haven't run the actual tests on Windows. Looks like a big improvement!

How do you run these tests on Windows ? Last I remember the build was fully managed by jhbuild, so I'm not too sure how one can go about to run the tests.

jralls:
> How do you run these tests on Windows ?

Assuming that GnuCash is already built, from a MINGW32 shell:
```
cd /c/gcdev64/src/gnucash-on-windows
TARGET=gnucash-stable jhbuild -f jhbuildrc shell
cd $PREFIX/../build/gnucash-git
ninja check
```

That's not going to work for actual CI testing because it would take too long to build all of the dependencies, so I'm working on a `pacman`-based `gnucash-on-windows` that will have prebuilt packages for all of the dependencies hosted in a repo on our Sourceforge project. That will also make it easier for people wanting to build/develop GnuCash on Windows.

PR #2127, opened 2025-09-08
add podman scripts

christopherlam:
This PR is mainly to discuss and work on creating a container for building. Not intending to merge in.

The aim is that the dev will need `podman` and nothing else.

See `pod.sh` which creates podman containers, caching the gcc compilation and storing build artifacts in separate volumes.

Unfortunately it barfs thus:

```
FAILED: libgnucash/quotes/finance-quote-wrapper.1 /app/build/libgnucash/quotes/finance-quote-wrapper.1 
cd /app/build/libgnucash/quotes && /usr/sbin/perl POD2MAN_EXECUTABLE-NOTFOUND /app/build/bin/finance-quote-wrapper /app/build/libgnucash/quotes/finance-quote-wrapper.1
Can't open perl script "POD2MAN_EXECUTABLE-NOTFOUND": No such file or directory
```


jralls:
> podman

I think you mean `pacman`, the Arch Linux (and Mingw) package management tool.

I'm actually working on this for Mingw, including PKGBUILD stuff for all of the packages that they don't support. That's obviously going to [gnucash-on-windows](https://github.com/gnucash/gnucash-on-windows). If you want something similar for arch the code would go in a new gnucash-on-arch repo.

Arch [already packages GnuCash](https://archlinux.org/packages/extra/x86_64/gnucash/), though they're 1 rev behind. Have you contacted their packager?

christopherlam:
I do mean podman the Docker alternative, setting up an arch container with Pacman packages... The workflow can be: install Linux, install podman, download sources, run script to build and run.

jralls:
So arch linux was just an example. So gnucash-on-docker instead of gnucash-on-arch.

christopherlam:
> So arch linux was just an example. So gnucash-on-docker instead of gnucash-on-arch.

Chatgpt did initially suggest podman pulling the Fedora image, but resulted in a different error. I picked arch because it seems to compile on ci

Containerfile
```diff
@@ -0,0 +1,11 @@
+FROM docker.io/library/archlinux:latest
+
+ENV DEBIAN_FRONTEND=noninteractive
+
+RUN pacman -Syu --quiet --noconfirm glibc gcc cmake make boost python3 pkg-config gettext gtk3 guile git ninja gtest gmock sqlite3 webkit2gtk swig gwenhywfar aqbanking intltool libxslt libofx postgresql-libs libmariadbclient libdbi libdbi-drivers wayland-protocols ccache valgrind perl && pacman -Scc --noconfirm
+
+ENV CCACHE_DIR=/ccache
+ENV PATH="/usr/lib/ccache/bin:${PATH}"
```
gjanssens: What's in /usr/lib/ccache/bin that is needs to be added to the PATH environment variable ? On Fedora 41 that directory doesn't exist.

pod.sh
```diff
@@ -0,0 +1,47 @@
...
+  -v gnucash-build:/app/build:Z \
+  -v gnucash-ccache:/ccache:Z \
+  -e DISPLAY=$DISPLAY \
+  -v /tmp/.X11-unix:/tmp/.X11-unix:Z \
+  $IMAGE_NAME bash -c "
+    mkdir -p /app/build
+    cd /app/build
+    if [ '$ACTION' = 'rebuild' ]; then
+        echo 'Cleaning build directory...'
+        rm -rf ./*
```
gjanssens, line 42: My experience with building gnucash taught me it's better to do `rm -fr ./* ./??*` to also remove hidden files for a really clean build.

gjanssens (commented):
Preliminary note: I have not yet run your scripts, but I have entertained a similar idea in the past.

I have been working with podman containers for some time now, so  a few suggestions:

1. The container is created to run bash by default (as per Dockerfile:11). However pod.sh replaces this command with a large inline script (starting at line 28). It would make more sense to move this inline script into a file of its own and include it in the image. A typical name would be `entrypoint.sh`. And then change the Dockerfile to have an `ENTRYPOINT` referring to this script. You can always override this default start script by passing a custom entrypoint to the podman run invocation should that be needed for some reason or another. Instead of a the explicit `ENTRYPOINT` in the Dockerfile, one could also achieve this by pointing to the entrypoint.sh script in the `CMD` line. Both have their advantages and drawbacks.
2. I don't see any step to download and install the gnucash sources ? Is the user supposed to add this manually somewhere ?
3. Your script is explicitly creating podman volumes. While that works, it doesn't give you much flexibility over where you want the persistent data to be stored on your filesystem. I have stopped using these "internal" podman volumes and instead I simply pass ordinary directories on my file system as volumes like `-v ${WORKDIR}/app:/app:Z`. These directories have to exist before running podman though.

As for your goal 'Requiring only podman', the wider goal no doubt is to make it as easy as possible for newcomers to experiment with the gnucash code base or to self-build.

There are a few hurdles to take in general:
* getting the source code
* getting the required dependencies
* setting up a directory structure in which you don't shoot yourself in the foot

This PR focuses on the second of these.

How do you imagine the first one ? Getting the source code for this container based build ?

After some reflection I see two possible directions to go:
1. Make your scripts part of the gnucash source tree. So each time the user gets the source in some form (git clone, download and extract source tar balls), they have both the sources and the necessary scripts to create the image and run the container. For this scenario it would be useful of the user can tell the pod script which directory contains the sources and which directory to use for building.
2. We could add a github action that creates an image for each commit and store it in the github container repo (no idea how that works yet). In that case you could include the source directory directly into the images you build and work from there. Note that the includes sources would be immutable, which means any change made to them would be lost. That could be fixed by altering the entrypoint code in yet another way: there would still be a persistent source volume. If it's empty when the container starts, the source tree included in the image would be copied over to the persistent source volume. This makes it easy to make changes (which persist), reset the sources if you're not happy with your changes (just empty the persistent source volume), or even choose to work with another source tree altogether (like a git clone that allows you to switch branches as you see fit).

Finally as for your build error: you're still missing a build dependency: pod2man
I don't know which arch package provides this. On Fedora 41, it comes with package perl-podlators-6.0.2-2.fc41.noarch

gjanssens:
> > So arch linux was just an example. So gnucash-on-docker instead of gnucash-on-arch.
> 
> Chatgpt did initially suggest podman pulling the Fedora image, but resulted in a different error. I picked arch because it seems to compile on ci

As containers a completely isolated, it shouldn't matter what distro you take as starting point. It mostly shows chatgpt is not very good a coding, or docker image building in this case... It will only give you something that statistically looks like what you ask for, but it doesn't understand any of what it spits out.

Out of curiosity, what was the error you got when starting from the fedora image ?

gjanssens:
Also, from what I know docker and podman were never designed to run graphical applications. You probably can, but it likely needs extra tweaks in what you share between host system and container. That's a whole challenge on its own. It's solved by flatpak and snap, but both are very opinionated wrt their build mechanisms which make them poorer options for development. An alternative could be [toolbx](https://containertoolbx.org/). It builds on podman, but handles all the details you need to properly integrate with the host system. I don't know if there is an equivalent of a Dockerfile to easily set up a toolbx container ready for gnucash development though.

gjanssens:
> I don't know if there is an equivalent of a Dockerfile to easily set up a toolbx container ready for gnucash development though.

[There is](https://containertoolbx.org/doc/#from-a-containerfile).

christopherlam:
Thanks @gjanssens the missage package was indeed `perl-podlators` and it now compiles and runs \o/ however the absent `dconf` means the gui is raw. Some response to comments inline:

> Preliminary note: I have not yet run your scripts, but I have entertained a similar idea in the past.
> 
> I have been working with podman containers for some time now, so a few suggestions:
> 
>   The container is created to run bash by default (as per Dockerfile:11). However pod.sh replaces this command with a large inline script (starting at line 28). It would make more sense to move this inline script into a file of its own and include it in the image. A typical name would be `entrypoint.sh`. And then change the Dockerfile to have an `ENTRYPOINT` referring to this script. You can always override this default start script by passing a custom entrypoint to the podman run invocation should that be needed for some reason or another. Instead of a the explicit `ENTRYPOINT` in the Dockerfile, one could also achieve this by pointing to the entrypoint.sh script in the `CMD` line. Both have their advantages and drawbacks.

agree, however I felt it would be opportune to try have a single `pod.sh` as much as possible

>   I don't see any step to download and install the gnucash sources ? Is the user supposed to add this manually somewhere ?

My setup has:

* ./pod.sh
* ./stable/ <-- sources

I didn't specify a url to download sources. Maybe the whole repo could be retrieved somewhere and the toplevel contains `pod.sh`? Alternatively `pod.sh` could `git clone` or `curl` from somewhere (eg default github or code) into `./stable`?

> As for your goal 'Requiring only podman', the wider goal no doubt is to make it as easy as possible for newcomers to experiment with the gnucash code base or to self-build.

^ agree this is the aim of this project

> After some reflection I see two possible directions to go:
> 
>   We could add a github action that creates an image for each commit [snip]

don't fancy relying on github at all! preferable the dev retrieves the sources separately.

> Finally as for your build error: you're still missing a build dependency: pod2man I don't know which arch package provides this. On Fedora 41, it comes with package perl-podlators-6.0.2-2.fc41.noarch

Containerfile
christopherlam: This was to hold the `ccache` data but it was causing more problems than expected (i.e. couldn't compile tip-of-the-day) so I've dropped it.

christopherlam:
Btw I forgot to annotate, this PR now works.

pod.sh
```diff
@@ -0,0 +1,82 @@
...
+podman build -t "$IMAGE_NAME" --build-arg "FEDORA_VERSION=$FEDORA_VERSION" -f - <<'EOF'
+ARG FEDORA_VERSION
+FROM fedora:${FEDORA_VERSION}
+RUN dnf -y install \
+      gcc gcc-c++ make automake autoconf libtool cmake ninja-build git \
+      guile30{,-devel} pkg-config gtk3{,-devel} webkit2gtk4.0{,-devel} \
+      libdbi{,-devel} libdbi-dbd-sqlite boost-devel gtest-devel gmock-devel \
+      libxml2-devel libxslt-devel swig gettext{,-devel} glib2-devel \
+      python3{,-devel} valgrind perl-podlators \
+    && dnf clean all
```
jralls: The Debian and Fedora package managers know what the dependencies are so you can just say `dnf builddep gnucash` instead of listing them. If we change dependencies for `future` you'd need to manually add the new deps.

pod.sh
```diff
@@ -0,0 +1,82 @@
...
+    set -euo pipefail
+    mkdir -p /app/build
+    cd /app/build
+    if [ "$ACTION" = rebuild ]; then
+        echo "Cleaning build directory..."
+        rm -rf ./*
+    fi
+    echo "Configuring CMake with install prefix..."
+    cmake -GNinja .. -DCMAKE_BUILD_TYPE=Debug -DWITH_OFX=OFF \
+          -DWITH_AQBANKING=OFF -DCMAKE_INSTALL_PREFIX=/usr/local
```
jralls: An official script should build everything, so instead of turning off OFX and AqB it should turn on Python (`-WITH_PYTHON=ON`).

pod.sh
```diff
@@ -0,0 +1,82 @@
...
+    echo "Building with Ninja..."
+    ninja
```
jralls, line 64: `ninja check` too.

christopherlam: `ninja check` to run with `./pod.sh check`, to avoid slowdown when simply calling `./pod.sh run` or `./pod.sh shell`

PR #2128, opened 2025-09-12
Bug 799661 - Exchange Rate entries displayed in reports in an unstabl…

agwekixj:
…e order

Modify gnc:html-make-rates-table to sort the list returned by gnc:accounts-get-commodities using gnc-commodity-get-nice-symbol


jralls:
@christopherlam any comments?

christopherlam:
It's fine... there are other commodities-list being reported eg in balsheet-pnl, so, if this matters then it would be better with a `gnc:accounts-get-commodities-sorted` function.

christopherlam:
Also the accounts would have currencies and commodities, and it would be better to stable-sort twice; first by `nice-symbol` then by `namespace`? Stable sort ensures multi-level sorting.

```scheme
(define (gnc:accounts-get-commodities-sorted accounts exclude-commodity)
  (stable-sort!
   (stable-sort!
    (gnc:accounts-get-commodities accounts exclude-commodity)
    (lambda (a b)
      (gnc:string-locale<?
       (gnc-commodity-get-nice-symbol a) (gnc-commodity-get-nice-symbol b))))
   (lambda (a b)
     (gnc:string-locale<?
      (gnc-commodity-get-namespace a) (gnc-commodity-get-namespace b)))))
```

agwekixj:
I've managed to screw up this branch when the changes to test-balsheet-pnl.scm collided.  The simplest solution to me is to abort this PR and create a new one.

agwekixj:
The new PR is: https://github.com/Gnucash/gnucash/pull/2129

PR #2129, opened 2025-09-13
Bug 799661 - Exchange Rate entries displayed in reports in an unstable order

agwekixj:
Added gnc:accounts-get-commodities-sorted


christopherlam:
Lgtm

PR #2130, opened 2025-09-17
Bug 799662 - Importing accounts from UTF8 csv file does not work correctly

agwekixj:
Modifies the imports of Accounts from CSV, Bills & Invoices, and Customer & Vendors, to only call g_locale_to_utf8() if the file is determined to be not UTF-8 compliant.

Removes the potential XML-prohibited codepoints.


jralls:
Thanks!

PR #2131, opened 2025-09-19
[new-owner-report.scm] modify highlight: visible on hover

christopherlam:
Slight modification to highlight behaviour. As per original implementation by @gjanssens. js added only if link-option is `'detailed`. I think improves visibility.

* hover to highlight matching transactions.
* click to freeze/unfreeze.


jralls:
That's going to be confusing without documentation somewhere.

PR #2132, opened 2025-09-21
Bug 799344 - Add "Postpone All" button to Scheduled Transactions dialog

crossan007:
Adds a "Postpone All" button to the "Since Last Run" Dialog.

Useful for cases where a user has a large number of "To Create" scheduled transactions and and wants to avoid accidentally dismissing scheduled transactions.

Especially helpful if it's been a while since opening a database and the user wants to manually validate that "auto pay" transactions have properly cleared.

image

Relates to:
* https://bugs.gnucash.org/show_bug.cgi?id=799344
* https://bugs.gnucash.org/show_bug.cgi?id=792121
* https://bugs.gnucash.org/show_bug.cgi?id=797005

### DIFF
--- a/gnucash/gnome/dialog-sx-since-last-run.c
+++ b/gnucash/gnome/dialog-sx-since-last-run.c
@@ -72,6 +72,7 @@ struct _GncSxSinceLastRunDialog
     GncSxSlrTreeModelAdapter *editing_model;
     GtkTreeView *instance_view;
     GtkToggleButton *review_created_txns_toggle;
+    GtkButton *postpone_all_button;
     GList *created_txns;
 
     GtkCellEditable *temp_ce; // used when editing values
@@ -139,6 +140,7 @@ static void _show_created_transactions (GncSxSinceLastRunDialog *app_dialog, GLi
 static void close_handler (gpointer user_data);
 static void dialog_destroy_cb (GtkWidget *object, GncSxSinceLastRunDialog *app_dialog);
 static void dialog_response_cb (GtkDialog *dialog, gint response_id, GncSxSinceLastRunDialog *app_dialog);
+static void postpone_all_button_clicked_cb (GtkButton *button, GncSxSinceLastRunDialog *app_dialog);
 
 #define debug_path(fn, text, path) {\
     gchar *path_string = gtk_tree_path_to_string (path);\
@@ -1340,6 +1342,7 @@ gnc_ui_sx_since_last_run_dialog (GtkWindow *parent, GncSxInstanceModel *sx_insta
 
     dialog->editing_model = gnc_sx_slr_tree_model_adapter_new (sx_instances);
     dialog->review_created_txns_toggle = GTK_TOGGLE_BUTTON(gtk_builder_get_object (builder, "review_txn_toggle"));
+    dialog->postpone_all_button = GTK_BUTTON(gtk_builder_get_object (builder, "postpone_all_button"));
 
     gtk_toggle_button_set_active (GTK_TOGGLE_BUTTON(dialog->review_created_txns_toggle),
                                   gnc_prefs_get_bool (GNC_PREFS_GROUP_STARTUP, GNC_PREF_SET_REVIEW));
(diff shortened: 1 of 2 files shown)


crossan007:
I think there's either a bug in my code, or I misunderstood how `SX_INSTANCE_STATE_POSTPONED` is supposed to work.

This seems to work at first, but as soon as I close and re-open GNUCash, all of my "Postponed" transactions are gone.

jralls:
No, you need to mark the db as dirty so that the changes get saved. Notice that the Save button isn't lighted up. For example, click postpone all then click OK to get rid of the SLR dialog and create a transaction. Creating the transaction will light up the Save button and then when you restart the transactions will all have status Postponed.

This does it:
```
--- a/gnucash/gnome/dialog-sx-since-last-run.c
+++ b/gnucash/gnome/dialog-sx-since-last-run.c
@@ -1540,6 +1540,7 @@ postpone_all_button_clicked_cb (GtkButton *button, GncSxSinceLastRunDialog *app_
     GncSxInstanceModel *instance_model;
     GList *sx_instances_list;
     GList *sx_iter;
+    gboolean changed = FALSE;

     g_return_if_fail (app_dialog != NULL);
     g_return_if_fail (app_dialog->editing_model != NULL);
@@ -1561,10 +1562,16 @@ postpone_all_button_clicked_cb (GtkButton *button, GncSxSinceLastRunDialog *app_
             // Only change instances that are not already created
             if (instance->state != SX_INSTANCE_STATE_CREATED)
             {
+                changed = TRUE;
                 gnc_sx_instance_model_change_instance_state (instance_model, instance, SX_INSTANCE_STATE_POSTPONED);
             }
         }
     }
+    if (changed)
+    {
+        QofBook *book = qof_session_get_book(gnc_get_current_session());
+        qof_book_mark_session_dirty(book);
+    }
 }

 static void
```

jralls:
Don't worry about the test failure, it's already fixed. You can rebase on stable to make it go away.

crossan007:
@jralls I don't quite understand why that's necessary here for bulk-changes to the SX_INSTANCE_STATE of each sx but isn't necessary for individual changes, but I just tried it and it seems to work, so thank you.

crossan007:
I take that back - this patch did not cause my "Postponed" scheduled transactions to persist after closing / re-opening my MySQL (real) database

jralls:
@crossan007 it turns out not to require it, it's just really slow to light up the save button. With a DBI backend it appears to be ignored: You don't have to close GnuCash to observe this, just look at the Last Occur column in the Scheduled Transactions list.

That's also the case if you postpone the transactions one at a time. The problem is that there's no routine to save the deferred list corresponding to the one in the [XML backend](https://github.com/Gnucash/gnucash/blob/01f0d252ec4008243a3efe3d67962015ce5a6119/libgnucash/backend/xml/gnc-schedxaction-xml-v2.cpp#L169). Somebody filed a bug about this in the last year, but Bugzilla is turned off so we can't search for it.

crossan007:
I've been (ab)using the "Remind" status to "save forater validation" my "To Create" transactions


Maybe I'll just adjust this MR to use that instead of postpone.


> @crossan007 it turns out not to require it, it's just really slow to light up the save button. With a DBI backend it appears to be ignored: You don't have to close GnuCash to observe this, just look at the Last Occur column in the Scheduled Transactions list.
> 
> That's also the case if you postpone the transactions one at a time. The problem is that there's no routine to save the deferred list corresponding to the one in the [XML backend](https://github.com/Gnucash/gnucash/blob/01f0d252ec4008243a3efe3d67962015ce5a6119/libgnucash/backend/xml/gnc-schedxaction-xml-v2.cpp#L169). Somebody filed a bug about this in the last year, but Bugzilla is turned off so we can't search for it.

jralls:
@crossan007 The postponed SXes not getting saved in SQL is [799536](https://bugs.gnucash.org/show_bug.cgi?id=799536). [799664](https://bugs.gnucash.org/show_bug.cgi?id=799664) points out that the SLR window doesn't do a good job with showing postponed transactions even on the XML backend.

jralls:
@crossan007 Are you still working on this?

crossan007:
@jralls Yes* 

* I'm running this branch locally for my day-do-day finances, but haven't had the time to get it working exactly the way I want it, nor adjust the code to be in compliance with the project's standards.

jralls:
@crossan007 OK. I've changed it to draft. Change it back when you're ready for us to take another look.

PR #2136, opened 2025-09-26
[gnucash-header.c] don't draw header->surface until it's realized

christopherlam:
@jralls maybe consider this after next release. I don't really understand unrealize vs finalize.

### DIFF
--- a/gnucash/register/register-gnome/gnucash-header.c
+++ b/gnucash/register/register-gnome/gnucash-header.c
@@ -52,6 +52,11 @@ G_DEFINE_TYPE (GncHeader, gnc_header, GTK_TYPE_LAYOUT)
 static void
 gnc_header_draw_offscreen (GncHeader *header)
 {
+    g_return_if_fail (GTK_IS_WIDGET(header));
+
+    if (!gtk_widget_get_realized (GTK_WIDGET(header)))
+        return;
+
     SheetBlockStyle *style = header->style;
     GncItemEdit *item_edit = GNC_ITEM_EDIT(header->sheet->item_editor);
     Table *table = header->sheet->table;


jralls:
@christopherlam a widget is `realized` when it's displayed and `unrealized` when it's undisplayed; in the case of a register that's equivalent to creation and destruction; `gtk_widget_unrealize` is called from `gtk_widget_dispose`. `unrealize` is also called when the widget is resized, followed by an immediate `realize` with the new size.

Since `unrealize` is called from `dispose` and `dispose` should run before `finalize` (see https://docs.gtk.org/gobject/concepts.html#object-memory-management) this *should* be unnecessary. Did you encounter a leaked cairo surface that prompted this?

christopherlam:
> Since `unrealize` is called from `dispose` and `dispose` should run before `finalize` (see https://docs.gtk.org/gobject/concepts.html#object-memory-management) this _should_ be unnecessary. Did you encounter a leaked cairo surface that prompted this?

Yes and large blobs of 200k too

jralls:
OK. Can you apply your patch and set a breakpoint on the `cairo_surface_destroy` to how `finalize` is getting called wiithout `dispose` getting called first? It really would be better to fix the problem than to put a band-aid on it.

christopherlam:
I think I can reproduce: open a session with at least 2 registers open. Activate only the first register. Quit. Relaunch under valgrind. Quit. The hidden registers are never drawn. Only the first register is `unrealize`. All registers are `finalize`d. See my debugging patch, and the result.

```patch
modified   gnucash/register/register-gnome/gnucash-header.c
@@ -49,6 +49,21 @@ enum
 
 G_DEFINE_TYPE (GncHeader, gnc_header, GTK_TYPE_LAYOUT)
 
+static guint
+get_index (GncHeader* header)
+{
+    static GPtrArray *hdrs = NULL;
+    if (!hdrs)
+        hdrs = g_ptr_array_new ();
+    guint idx;
+    if (!g_ptr_array_find (hdrs, header, &idx))
+    {
+        g_ptr_array_add (hdrs, header);
+        idx = hdrs->len - 1;
+    }
+    return idx;
+}
+
 static void
 gnc_header_draw_offscreen (GncHeader *header)
 {
@@ -76,6 +91,8 @@ gnc_header_draw_offscreen (GncHeader *header)
     color_type = gnc_table_get_color (table, virt_loc, NULL);
     gnucash_get_style_classes (header->sheet, stylectxt, color_type, FALSE);
 
+    printf ("header %p[%d] generate surface = %p\n",
+            header, get_index (header), header->surface);
     if (header->surface)
         cairo_surface_destroy (header->surface);
     header->surface = cairo_image_surface_create (CAIRO_FORMAT_ARGB32,
@@ -249,6 +266,8 @@ static void
 gnc_header_unrealize (GtkWidget *widget)
 {
     GncHeader *header = GNC_HEADER(widget);
+    printf ("header %p[%d] unrealize surface = %p\n",
+            header, get_index (header), header->surface);
     if (header->surface)
         cairo_surface_destroy (header->surface);
     header->surface = NULL;
@@ -272,6 +291,8 @@ gnc_header_finalize (GObject *object)
     GncHeader *header;
 
     header = GNC_HEADER(object);
+    printf ("header %p[%d] finalize surface = %p\n",
+            header, get_index (header), header->surface);
 
     g_free (header->cursor_name);
     header->cursor_name = NULL;
```

christopherlam:
Output; 3 registers; only first one drawn. Only the first register[0] is unrealized. All are finalized.

```
header 0x3266df20[0] generate surface = (nil)
header 0x3266df20[0] generate surface = 0x331b4350
header 0x332c8cd0[1] generate surface = (nil)
header 0x332c8cd0[1] generate surface = 0x332dc950
header 0x33360e90[2] generate surface = (nil)
header 0x33360e90[2] generate surface = 0x33369160
header 0x3266df20[0] generate surface = 0x331b4350
header 0x3266df20[0] generate surface = 0x331b4350
header 0x332c8cd0[1] generate surface = 0x332dc950
header 0x332c8cd0[1] generate surface = 0x332dc950
header 0x33360e90[2] generate surface = 0x33369160
header 0x33360e90[2] generate surface = 0x33369160
header 0x3266df20[0] generate surface = 0x331b4350
header 0x332c8cd0[1] generate surface = 0x332dc950
header 0x33360e90[2] generate surface = 0x33369160
header 0x33360e90[2] finalize surface = 0x33369160
header 0x332c8cd0[1] finalize surface = 0x332dc950
header 0x3266df20[0] unrealize surface = 0x3368e940
header 0x3266df20[0] finalize surface = (nil)
```

jralls:
OK. Maybe we should check `gtk_widget_is_realized` at the start of `gnc_header_draw_offscreen`. There's no point in drawing an unrealized widget.

jralls:
Not creating the surface in an unrealized widget should make the destroying the surface in finalize unnecessary, and that's the case in my (simple) testing. On the one hand I suppose it could stay as insurance; on the other it prevents leak detection from flagging a problem if there's another path where a surface is created unnecessarily or manages to evade `gtk_widget_dispose`.

jralls:
Looks good, go ahead and merge when ready.

PR #2137, opened 2025-10-06
Fix saving account page changes

agwekixj:
Partially reverts a fix for bug 799249 that prevented preserving modifications to the account tree.

For further discussion: https://lists.gnucash.org/pipermail/gnucash-user/2025-October/117752.html


jralls:
Thanks. I didn't notice that the unref was redundant to gtk_widget_destroy a few lines later.

PR #2138, opened 2025-10-08
Reset gui component session

agwekixj:
When saving a new file, reset the managed gui components to the new session prior to clearing the old session.

This is the follow-on patch noted in: https://lists.gnucash.org/pipermail/gnucash-user/2025-October/117787.html


jralls:
Thanks!

PR #2139, opened 2025-10-12
Fix stale account balances and grand total during transaction import

dl-g:
The bug:

After importing transactions from an existing account, the grand total net assets is updated to N, and the ending balance of the imported account is X.

Quit and restart gnucash, the net assets changes to N+X.

on_matcher_ok_clicked invokes gnc_resume_gui_refresh before the account is fully commited. xaccAccountRecomputeBalance skips an account if it has pending edits.

The fix:
Invoke gnc_resume_gui_refresh after all accounts have fully commited.

Tests:

make check

reran the above test, the net assets balance is updated to N+X without restarting gnc.


jralls:
Thanks!

PR #2140, opened 2025-10-14
Some char occasional leaks

christopherlam:
any commits of interest?


libgnucash/engine/TransLog.cpp
```diff
@@ -141,13 +141,12 @@ gboolean
...
     gchar *base;
-    gint result;
 
-    if (!name || !trans_log_name)
+    if (!name || trans_log_name.empty())
         return FALSE;
 
     base = g_path_get_basename(name);
-    result = (strcmp(base, trans_log_name) == 0);
+    gboolean result = trans_log_name.compare(base) == 0;
```
jralls: This is c++: Boolean operators return type `bool`, not `int`.

libgnucash/engine/TransLog.cpp
```diff
@@ -266,37 +265,27 @@ xaccTransWriteLog (Transaction *trans, char flag)
...
+                         << gnc_numeric_num(val) << '/' << gnc_numeric_denom(val) << '\t'
+                         << drecn << '\n';
     }
 
-    fprintf (trans_log, "===== END\n");
-
-    /* get data out to the disk */
-    fflush (trans_log);
+    trans_log_stream << "===== END\n";
+    trans_log_stream.flush();
```
jralls: You can do that more efficiently with `trans_log_stream << "===== END" << std::endl;

jralls (changes requested):
A couple of nits.

Both qoflog and the main matcher code could be made a lot more efficient (in a separate PR and if you're so inclined) by rethinking how `qof_logger_format` and `color` are managed: There are only two formats, so they could be static const char*, no allocations required and (I think) four colors that could be created and destroyed with the main matcher window instead of for every row.

PR #2143, opened 2025-10-25
Removed China Merchants Bank from quote sources.

bpschuck:
The Finance::Quote module is no longer working and will be removed in the next release.

I will update the Quote Sources Wiki page within the next couple days.


jralls:
Thanks!

PR #2144, opened 2025-10-30
Bug 799682 - Unable to "Save As" (or open) a non-xml data file

agwekixj:
Fixes a generic bug in gnc_module_init_backend_dbi() when HAVE_LIBDBI_R is defined.  Specifically, when the initial dbi_initialize_r() returns no drivers, the dbi_instance variable is still set.


jralls:
Thanks!

PR #2147, opened 2025-11-08
Autoclear. now usable.

christopherlam:
much more efficient with cacheable memory.

- ~instead of hashtable, use sorted vector, merge sort as required~ 
- ~don't store list of splits, but store previous index and backtrack to create result splits~
- don't store reachable sums and constituent splits. use recursion, taking candidate splits and narrower possible splits, tracking the target amount. store solution if found, and abort if more than 1 solution found.

result: `test-autoclear` completes in under ~20ms~40ms instead of 170ms. clearing _k_ splits with worst-case scenario amounts 2^k up to $1,048,576 will ~create a sums vector with 2097152 elements~ recurse through the `subset_sum` function >2million times, but with only O(k) stack and heap memory.

@cristiklein


gnucash/gnome-utils/gnc-autoclear.cpp
```diff
@@ -0,0 +1,198 @@
+/********************************************************************
+ * gnc-autoclear.c -- Knapsack algorithm functions                  *
+ *                                                                  *
+ * Copyright 2020 Cristian Klein <cristian@kleinlabs.eu>            *
+ * Modified  2021 Christopher Lam to clear same-amount splits       *
+ * Modified  2025 ChatGPT to merge without building a shifted vec   *
```
jralls: The FSF had a [panel last month](https://lwn.net/Articles/1040888/) on use of LLMs in Free software; I recommend the short article, it's an interesting read. My takeaway is that until the courts get some sort of consensus on how LLMs affect intellectual property claims and rights we have to be a bit conservative. You have to write the code yourself so that you can claim copyright without reservation. Consult ChatGPT the same way you would Stack Overflow or programming texts.

Also, don't say "Modified" here. The point of this block is to assert copyright, so say "Copyright Christopher Lam 2021-2025". Or just 2021: The date doesn't matter anymore as long as there is one, copyright lasts until 70 years after your death.

gnucash/gnome-utils/gnc-autoclear.cpp
```diff
@@ -0,0 +1,198 @@
...
+#include "Split.h"
+#include "Transaction.h"
+#include "gnc-autoclear.h"
+
+#include <vector>
+#include <numeric>
+#include <algorithm>
+#include <cstdint>
+
+#define log_module "autoclear"
```
jralls: This should be
```
static QofLogModule log_module = "gnc.autoclear"
```
`gnc` is important, it allows adjusting the logging level for the whole application.

On the topic of logging, this module should use more, at least `ENTER()` and `LEAVE()` macros on the major functions.

gnucash/gnome-utils/gnc-autoclear.cpp
```diff
@@ -0,0 +1,198 @@
...
+
+struct Entry
+{
+    int64_t sum;
+    int64_t pred_index;   // index of last split
+    bool unique;          // true if sum is still unique
+    bool operator<(const Entry& o) const { return sum < o.sum; }
+};
+
+/**
```
jralls: `subset_sum_unique` isn't part of the public API so this should be a regular comment, not a Doxygen one.

That aside, you've thoroughly screwed up the Doxygen commands. Use the [command reference](https://www.doxygen.nl/manual/commands.html#cmdresult) when writing Doxygen comments.

gnucash/gnome-utils/gnc-autoclear.cpp
```diff
@@ -0,0 +1,198 @@
...
+ * subset_sum_unique
+ * @splits: list of split amounts
+ * @target: target sum
+ * @result: output indices of splits that sum to target if unique
+ * Returns nullptr on success, otherwise returns error message.
+ *
+ * This implementation merges the existing sums vector with a virtual
```
jralls: It's not existing after this commit. History goes in git commit messages, comments should explain the current code.

gnucash/gnome-utils/gnc-autoclear.cpp
```diff
@@ -0,0 +1,198 @@
...
+{
+    if (gnc_numeric_zero_p (target))
+        return "Account is already at Auto-Clear Balance.";
+
+    // start with base case: sum 0 reachable with no pred
+    std::vector<Entry> sums = {{0, -1, true}};
+
+    for (int64_t i = 0; i < (int64_t)splits.size(); ++i)
+    {
+        int64_t amt = xaccSplitGetAmount(splits[i]).num;
```
jralls: Using only `num` assumes that all of the splits have the same `denom`.  That's not a valid assumption. You need to get the account's scu and run `gnc_numeric_convert(amount, scu, GNC_HOW_DENOM_EXACT | GNC_HOW_ROUND_NEVER)` on target and on each split.

gnucash/gnome-utils/gnc-autoclear.cpp
```diff
@@ -0,0 +1,198 @@
...
+    bool unique;          // true if sum is still unique
+    bool operator<(const Entry& o) const { return sum < o.sum; }
+};
+
+/**
+ * subset_sum_unique
+ * @splits: list of split amounts
+ * @target: target sum
+ * @result: output indices of splits that sum to target if unique
+ * Returns nullptr on success, otherwise returns error message.
```
jralls: Returning `nullptr` on success flips convention. You've already got a [GError](https://docs.gtk.org/glib/struct.Error.html)** in the outer function. Pass it in and use it, returning `result`: [RVO](https://en.cppreference.com/w/cpp/language/copy_elision.html) will construct it in the outer stack frame for you.

gnucash/gnome-utils/gnc-autoclear.cpp
```diff
@@ -0,0 +1,198 @@
...
+ * "shifted" view (sums + amt) by keeping two independent indices and
+ * computing the shifted value on the fly, avoiding allocation of a
+ * separate shifted vector.
+ */
+
+static const char*
+subset_sum_unique (const SplitVec& splits, gnc_numeric target, SplitVec& result)
+{
+    if (gnc_numeric_zero_p (target))
+        return "Account is already at Auto-Clear Balance.";
```
jralls: What if the account has two unreconciled splits, a purchase and a refund?

gnucash/gnome-utils/gnc-autoclear.cpp
```diff
@@ -0,0 +1,198 @@
...
+        std::vector<Entry> merged;
+        merged.reserve(sums.size() * 2);
+
+        size_t a = 0, b = 0;
+        // Note: 'b' indexes into the same 'sums' vector but we treat it as the
+        // source for shifted sums (sums[b].sum + amt).
+        while (a < sums.size() && b < sums.size())
+        {
+            int64_t sum_a = sums[a].sum;
+            int64_t sum_b_shifted = sums[b].sum + amt;
```
jralls: Use `auto` instead of `int64_t`.

gnucash/gnome-utils/gnc-autoclear.cpp
```diff
@@ -0,0 +1,198 @@
...
+ * subset_sum_unique
+ * @splits: list of split amounts
+ * @target: target sum
+ * @result: output indices of splits that sum to target if unique
+ * Returns nullptr on success, otherwise returns error message.
+ *
+ * This implementation merges the existing sums vector with a virtual
+ * "shifted" view (sums + amt) by keeping two independent indices and
+ * computing the shifted value on the fly, avoiding allocation of a
+ * separate shifted vector.
```
jralls: That's a commit message, not a description of the function. This function is sufficiently opaque that an explanation of what sum and sum_shifted are about would be very useful.

jralls (changes requested):
`gnc-autoclear` should be in libgnucash/app-utils or libgnucash/engine. Only GUI code should be in gnucash/gnome-utils.

cristiklein:
Hi @christopherlam ! Really cool that you brought this feature to the next step. This is exactly what I intended to achieve back in 2010.

Sadly, I no longer use GnuCash, so I cannot test.

But it does look really exciting!

gnucash/gnome-utils/gnc-autoclear.cpp
christopherlam: > What if the account has two unreconciled splits, a purchase and a refund?

Then the algorithm is designed to always fail because there's more than 1 solution: including both signed transactions, and excluding both.

christopherlam:
Ok I think this is as optimized as possible. Duplicate amounts are handled via elimination -- handle all or nothing. e.g amounts 3 3 10 means only 6 and 10 are considered, with targets 6 10 16.

Complexity analysis with k = unique amounts

Aspect | Worst-case | Average / typical
-- | -- | --
Time | O(2^k)  | O(k) to O(k⋅logk)
Stack space | O(k) | O(k)
Heap space | O(k) | O(k)
Max runtime (20 splits) | ≤ 1 s (due to runtime monitor) | ~10–100 ms

</div></div>

gjanssens:
> Ok I think this is as optimized as possible. Duplicate amounts are handled via elimination -- handle all or nothing. e.g amounts 3 3 10 means only 6 and 10 are considered, with targets 6 10 16.

I haven't gone through the details so perhaps I'm about to ask a silly question. With amounts 3 3 10, why wouldn't you consider 3 and 13, but only 6 and 10 ?

christopherlam:
> > Ok I think this is as optimized as possible. Duplicate amounts are handled via elimination -- handle all or nothing. e.g amounts 3 3 10 means only 6 and 10 are considered, with targets 6 10 16.
> 
> I haven't gone through the details so perhaps I'm about to ask a silly question. With amounts 3 3 10, why wouldn't you consider 3 and 13, but only 6 and 10 ?

not silly... if there are 2 splits with $3 and one with $10, auto-clear aims to find the unique set of splits which reaches the target. and if the target is $3 or $13 then the algorithm recognises that it cannot choose from either $3 split.

christopherlam:
example log illustrates algorithm nicely. assume 3 splits amounts 2,3,5. targets 2, 3, 7, 8, 10 are reachable; target 5 is ambiguous.

first run with target 7 finds a unique solution `2|5`:
```
 [enter subset_sum()]  this=2 path=<empty> target=7
     [enter subset_sum()]  this=3 path=2 target=5
         [enter subset_sum()]  this=5 path=2|3 target=2
             [enter subset_sum()]  this=0 path=2|3|5 target=-3
             [leave subset_sum()] end of splits
             [enter subset_sum()]  this=0 path=2|3 target=2
             [leave subset_sum()] end of splits
         [leave subset_sum()] 
         [enter subset_sum()]  this=5 path=2 target=5
             [enter subset_sum()]  this=0 path=2|5 target=0
             [leave subset_sum()] solution found 2|5
             [enter subset_sum()]  this=0 path=2 target=5
             [leave subset_sum()] end of splits
         [leave subset_sum()] 
     [leave subset_sum()] 
     [enter subset_sum()]  this=3 path=<empty> target=7
         [enter subset_sum()]  this=5 path=3 target=4
             [enter subset_sum()]  this=0 path=3|5 target=-1
             [leave subset_sum()] end of splits
             [enter subset_sum()]  this=0 path=3 target=4
             [leave subset_sum()] end of splits
         [leave subset_sum()] 
         [enter subset_sum()]  this=5 path=<empty> target=7
             [enter subset_sum()]  this=0 path=5 target=2
             [leave subset_sum()] end of splits
             [enter subset_sum()]  this=0 path=<empty> target=7
             [leave subset_sum()] end of splits
         [leave subset_sum()] 
     [leave subset_sum()] 
 [leave subset_sum()] 
```

second run with target 5 will abort:
```
[enter subset_sum()]  this=2 path=<empty> target=5
    [enter subset_sum()]  this=3 path=2 target=3
        [enter subset_sum()]  this=5 path=2|3 target=0
        [leave subset_sum()] solution found 2|3
        [enter subset_sum()]  this=5 path=2 target=3
            [enter subset_sum()]  this=0 path=2|5 target=-2
            [leave subset_sum()] end of splits
            [enter subset_sum()]  this=0 path=2 target=3
            [leave subset_sum()] end of splits
        [leave subset_sum()] 
    [leave subset_sum()] 
    [enter subset_sum()]  this=3 path=<empty> target=5
        [enter subset_sum()]  this=5 path=3 target=2
            [enter subset_sum()]  this=0 path=3|5 target=-3
            [leave subset_sum()] end of splits
            [enter subset_sum()]  this=0 path=3 target=2
            [leave subset_sum()] end of splits
        [leave subset_sum()] 
        [enter subset_sum()]  this=5 path=<empty> target=5
            [enter subset_sum()]  this=0 path=5 target=0
            [leave subset_sum()] abort: found another solution 5
```

btw @jralls the ENTER and LEAVE macros won't indent/dedent properly when throwing exceptions.

libgnucash/app-utils/gnc-autoclear.cpp
```diff
@@ -0,0 +1,219 @@
...
+        const auto elapsed = std::chrono::duration_cast<std::chrono::duration<double>>(std::chrono::steady_clock::now() - m_start).count();
+        return elapsed > m_seconds;
+    }
+};
+
+struct SplitInfo
+{
+    int64_t m_amount;
+    Split* m_split;
+    friend bool operator<(const SplitInfo& a, const SplitInfo& b)
```
jralls: `friend` isn't needed here: It's a `struct` with no `private:` declaration so there's nothing that needs special access. 

This is presumably for use in STL algorithms. It doesn't match the normal comparison signature because of the hidden `this` argument. It should be a non-class (aka free) function.

jralls:
> btw @jralls the ENTER and LEAVE macros won't indent/dedent properly when throwing exceptions.

You mean here?
```
 if (!solution.empty())
        {
            LEAVE ("abort: found another solution %s", path_to_str (path));
            throw std::runtime_error ("The selected amount cannot be cleared.");
        }
```
Complain to your code editor. And BTW, that block is indented in the gnu/gnome style with the braces in 2: Our style has the braces flush with the code still indented 4.

jralls (changes requested):
Otherwise looks good.

christopherlam:
> ```
> 
>             [leave subset_sum()] end of splits
>             [enter subset_sum()]  this=0 path=3 target=2
>             [leave subset_sum()] end of splits
>         [leave subset_sum()] 
>         [enter subset_sum()]  this=5 path=<empty> target=5
>             [enter subset_sum()]  this=0 path=5 target=0
>             [leave subset_sum()] abort: found another solution 5
> ```
> btw @jralls the ENTER and LEAVE macros won't indent/dedent properly when throwing exceptions.

no: see the `abort: found another solution` log entry; it's indented a few levels due to the recursion; it aborts via `throw`, and won't dedent back to original level. it means the subsequent log entries (not shown here) will be indented a few levels.

another note: this PR offers autoclear during every reconciliation. timeouts at 1s. is this in agreement?

libgnucash/app-utils/gnc-autoclear.cpp
```diff
@@ -0,0 +1,219 @@
...
+
+    // 1st path: use current_num
+    path.insert(path.end(), iter, next);
+    subset_sum (next, end, path, solution, target - current_num * dupe_count,
+                monitor);
+    path.erase(path.end() - dupe_count, path.end());
+
+    // 2nd path: skip current_num
+    subset_sum (next, end, path, solution, target, monitor);
+    LEAVE ("");
```
christopherlam: Fwiw With this exit macro, the preceding recursive call is no longer tail call optimisable...

christopherlam:
after further testing, reverted to testing all 2^k combinations instead of deduping which was buggy. still processes 20 splits comfortably within 1s. also using exceptions is slow; now tests run in 15ms (cf 20-30ms with throw/catch).

gnucash/gnome/window-reconcile.cpp
```diff
@@ -1719,6 +1720,36 @@ close_handler (gpointer user_data)
...
+
+    if (error)
+    {
+        PWARN ("cannot autoclear: %s", error->message);
+        g_error_free (error);
+        return;
+    }
+
+    const char* autoclear = N_("Found %d splits up to %s which \
+will clear to %s. Do you want to auto-clear?");
```
jralls: Instead of a continuation you can split the string constant,
```
    const char* autoclear = N_("Found %d splits up to %s which"
                               " will clear to %s. Do you want to auto-clear?");
```

libgnucash/app-utils/gnc-autoclear.cpp
jralls: Having ENTER and LEAVE in a recursive function is also extremely noisy. I'd change the ENTER and error exit logs to DEBUG and drop the routine LEAVE.

OTOH you've got two recursion calls. Doesn't that break the optimization anyway?

jralls:
> also using exceptions is slow

They do add a lot of overhead and unwinding the stack between the throw and catch can be expensive. On the other hand an error return in a double recursion like `subset_sum` will only terminate the one leg so if you need to terminate the whole thing (e.g. the timeout) then you have a choice of throwing or testing for an error before the second recursion call. Which is faster/better/more efficient is a matter of some debate among the C++ punditry, but you do neither so you have the potential of testing the timeout many times as the recursive stack unwinds. 

When there isn't a need to abort the recursion you can wring out a little more performance by declaring the recursion function `noexcept` that tells the compiler not to generate the stack-unwinding and `catch` remembering overhead for that function. Just make sure that nothing the function calls can throw.

christopherlam:
With #2151 merged in, I find that autoclear 20 splits is fast with runtime O(2<sup>n</sup>) memory O(n). Apparently there's another algorithm which will handle up to 40 splits with runtime O(2<sup>n/2</sup>), memory O(2<sup>n/2</sup>). however I can't understand it, and I'll just park it here.

```diff
diff --git a/libgnucash/app-utils/gnc-autoclear.cpp b/libgnucash/app-utils/gnc-autoclear.cpp
index 11fd4d58a4..028e33c654 100644
--- a/libgnucash/app-utils/gnc-autoclear.cpp
+++ b/libgnucash/app-utils/gnc-autoclear.cpp
@@ -83,6 +83,8 @@ struct Solution
     SplitVec splits;
 };
 
+// subset sum
+
 static const char*
 path_to_str(const SplitInfoVec& path)
 {
@@ -162,6 +164,130 @@ subset_sum (SplitInfoVec::const_iterator iter,
     subset_sum (next, end, path, solution, target, monitor);
 }
 
+// meet-in-the-middle
+
+
+struct SubsetEntry
+{
+    int64_t sum;
+    uint64_t mask;   // up to 64 splits — fine
+};
+
+static void
+meet_in_the_middle (const SplitInfoVec& splits, Solution& solution,
+                    int64_t target, RuntimeMonitor& monitor)
+{
+    const size_t N = splits.size();
+    if (N == 0) return;
+
+    const size_t mid = N / 2;
+
+    // Left: [0, mid)
+    // Right: [mid, N)
+    const size_t NL = mid;
+    const size_t NR = N - mid;
+
+    const size_t left_count = 1ULL << NL;
+    const size_t right_count = 1ULL << NR;
+
+    std::vector<SubsetEntry> left;
+    std::vector<SubsetEntry> right;
+    left.reserve(left_count);
+    right.reserve(right_count);
+
+    // ----- generate left subsets -----
+    for (uint64_t mask = 0; mask < left_count; ++mask)
+    {
+        if (monitor.should_abort()) {
+            solution.abort = true;
+            return;
+        }
+
+        int64_t sum = 0;
+        for (size_t i = 0; i < NL; ++i)
+            if (mask & (1ULL << i))
+                sum += splits[i].m_amount;
+
+        left.push_back({sum, mask});
+    }
+
+    // ----- generate right subsets -----
+    for (uint64_t mask = 0; mask < right_count; ++mask)
+    {
+        if (monitor.should_abort()) {
+            solution.abort = true;
+            return;
+        }
+
+        int64_t sum = 0;
+        for (size_t i = 0; i < NR; ++i)
+            if (mask & (1ULL << i))
+                sum += splits[mid + i].m_amount;
+
+        right.push_back({sum, mask});
+    }
+
+    // Sort right by sum for binary search
+    std::sort(right.begin(), right.end(),
+              [](const SubsetEntry& a, const SubsetEntry& b)
+              { return a.sum < b.sum; });
+
+    // Helper lambda: find all right subsets with a given sum
+    auto range_equal_sum = [&](int64_t need)
+    {
+        return std::equal_range (right.begin(), right.end(), SubsetEntry{need, 0},
+                                   [](const SubsetEntry& a, const SubsetEntry& b)
+                                   { return a.sum < b.sum; });
+    };
+
+    bool found_once = false;
+
+    // ----- search -----
+    for (const auto& L : left)
+    {
+        if (monitor.should_abort()) {
+            solution.abort = true;
+            return;
+        }
+
+        int64_t need = target - L.sum;
+
+        auto pr = range_equal_sum(need);
+        if (pr.first == pr.second) continue; // nothing found
+
+        // If > 1 right subset matches → ambiguity
+        size_t count = std::distance(pr.first, pr.second);
+        if (count >= 2 || found_once)
+        {
+            solution.abort = true;
+            return;
+        }
+
+        // count == 1 and not found_before → record solution
+        const auto& R = *pr.first;
+        found_once = true;
+
+        // Construct final list of Split*
+        uint64_t Lmask = L.mask;
+        uint64_t Rmask = R.mask;
+
+        solution.splits.clear();
+        solution.splits.reserve(N);
+
+        for (size_t i = 0; i < NL; ++i)
+            if (Lmask & (1ULL << i))
+                solution.splits.push_back(splits[i].m_split);
+
+        for (size_t i = 0; i < NR; ++i)
+            if (Rmask & (1ULL << i))
+                solution.splits.push_back(splits[mid + i].m_split);
+    }
+
+    if (!found_once)
+        solution.splits.clear();
+}
+
+
 GList *
 gnc_account_get_autoclear_splits (Account *account, gnc_numeric toclear_value,
                                   time64 end_date, GError **error,
@@ -220,12 +346,17 @@ gnc_account_get_autoclear_splits (Account *account, gnc_numeric toclear_value,
 
     RuntimeMonitor monitor{max_seconds};
     Solution solution;
-    SplitInfoVec path;
-    path.reserve (splits.size());
 
-    subset_sum (splits.begin(), splits.end(), path, solution, target, monitor);
+    if (splits.size() < 20)
+    {
+        SplitInfoVec path;
+        path.reserve (splits.size());
+        subset_sum (splits.begin(), splits.end(), path, solution, target, monitor);
+    }
+    else
+        meet_in_the_middle (splits, solution, target, monitor);
 
-    DEBUG ("finished subset_sum in %f seconds", monitor.get_elapsed());
+    DEBUG ("finished autoclear algorithm in %f seconds", monitor.get_elapsed());
 
     if (solution.splits.empty() || solution.abort)
     {
```

jralls:
It's dividing the splits into two parts and computing vectors of sums for each of the two parts, then iterating over the first (left) vector and seeing if there's one and only one sum in the second (right) vector that equals the target. The right vector is sorted by sum to allow binary search for a match and quick determination if the match is unique. If at the end it finds only one match then it succeeds. 

If there are 24 splits then with a single list the vector of sums has 2<sup>24</sup> elements but with a pair each has only 2<sup>12</sup>, a 2<sup>11</sup> reduction in the memory use and the number of iterations. That's a pretty good speedup but I don't think it would scale much beyond 32 total splits: 2<sup>16</sup> entries in each vector and 16 bytes per entry is 4GiB per vector, about 1M iterations to stuff each one, then sorting a vector with 64K entries, and finally 64K iterations through the left vector. You could get some more scale by parallelizing computing the sums and searching the left vector's values. That's not standard C++ until 26 but there's a [gcc extension](https://gcc.gnu.org/onlinedocs/gcc/Vector-Extensions.html) you could use now should you be so inclined.

jralls:
There's [feedback](https://lists.gnucash.org/pipermail/gnucash-user/2025-December/118641.html) on the user list. It's not positive.

jralls:
@christopherlam I propose to revert https://github.com/Gnucash/gnucash/pull/2147/changes/43561deda4a29f9fceea92f5e261614157b85008 before Sunday's string freeze so that this isn't shoved at users who don't want it.

PR #2148, opened 2025-11-08
Prevent user's environment from messing with GNC_HOME.

jralls:
https://lists.gnucash.org/pipermail/gnucash-user/2025-November/118103.html reported that GnuCash crashes when he runs it with GNC_HOME=<somewhere random>. We use GNC_HOME to set XDG_DATA_DIR in the environment file to accommodate relocating GnuCash on macOS and Windows and GnuCash crashes when the schemas aren't found.

@gjanssens can you think of any reason not to do this? ISTM letting the user set any of these is asking for trouble.

### DIFF
--- a/libgnucash/core-utils/gnc-environment.c
+++ b/libgnucash/core-utils/gnc-environment.c
@@ -22,6 +22,9 @@
  */
 
 #include <glib.h>
+#include <stdbool.h>
+#include <stdio.h>
+
 #include <string.h>
 #include "gnc-environment.h"
 #include "gnc-path.h"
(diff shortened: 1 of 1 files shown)


gjanssens:
No, I agree preventing these variables from being overwritten is a good thing. When I coded it years ago I tried to make it as flexible as possible to allow maximum user freedom. I assumed that if a user set GNC_HOME they would have a good reason for it. But I think I went way too far there. It only makes sense that these parameters are set to what matches the proper installation paths as gnucash derives them at startup.

Note that users can still mess up if they explicitly add a new definition for GNC_HOME in the environment file somewhere. If you want to close that door as well, we should add a filter in the environment parsing code that filters out any attempt to set one of the gnucash defined variables you touch in the PR.

PR #2149, opened 2025-11-09
Make a link in the build directory for options.scm

richardcohen:
Should fix bug 799159


jralls:
Thanks!

> Should fix bug 799159

Well, it fixes the first bit, the scheme tests are more resilient to having an installed GnuCash. It doesn't do anything about test-userdata-dir-invalid-home.

richardcohen:
> It doesn't do anything about test-userdata-dir-invalid-home.

I think that is a red herring - I am running Ubuntu 25.10, with distro-installed  gnucash 5.13, and all the tests pass.

jralls:
Maybe a red herring or maybe a separate Gentoo issue. Apparently they've had it disabled for some time.

PR #2153, opened 2025-11-24
About copy

Bob-IT:
This was what I was think for coping the About configuration text in one go. I have added some text and copy icon that can be clicked on to copy the setting to the clipboard. The first image below shows the about dialog and the second how it was pasted to a text file.

Screenshot_2025-11-24_14-46-02
about-copy

The second commit adds the operating system to the dialog as seen below...
Screenshot_2025-11-24_15-00-00

### DIFF
--- a/gnucash/gnome-utils/gnc-main-window.cpp
+++ b/gnucash/gnome-utils/gnc-main-window.cpp
@@ -5266,22 +5266,162 @@ url_signal_cb (GtkAboutDialog *dialog, gchar *uri, gpointer data)
     return TRUE;
 }
 
-static gboolean
-link_button_cb (GtkLinkButton *button, gpointer user_data)
+#define DEFAULT_MARGIN 5
+
+static void
+set_text_cursor (GdkWindow *win, GdkCursorType type)
 {
-   const gchar *uri = gtk_link_button_get_uri (button);
-   gchar *escaped_uri = g_uri_escape_string (uri, ":/.\\", true);
-   gnc_launch_doclink (GTK_WINDOW(user_data), escaped_uri);
-   g_free (escaped_uri);
-   return TRUE;
+    if (!win && !type)
+        return;
+
+    GdkCursor *current = gdk_window_get_cursor (win);
+    if (type == gdk_cursor_get_cursor_type (current))
+        return;
+
+    GdkCursor *cur = gdk_cursor_new_for_display (gdk_window_get_display (win), type);
+    gdk_window_set_cursor (win, cur);
+}
+
+static gboolean
+textview_motion_notify_cb (GtkWidget *textview,
+                           GdkEventMotion *event,
+                           gpointer user_data)
+{
+    if (event->state & GDK_BUTTON1_MASK)
+        return false;
+
+    GtkTextIter iter;
+    gboolean valid = gtk_text_view_get_iter_at_location (GTK_TEXT_VIEW(textview),
+                                                         &iter, event->x, event->y);
+
+    if (valid && (event->y > DEFAULT_MARGIN))
+    {
+        GSList *tt_list = gtk_text_iter_get_tags (&iter);
+
+        if (tt_list)
+        {
+            GtkTextTag *tt = (GtkTextTag*)g_slist_nth_data (tt_list, 0);
+
+            if (g_object_get_data (G_OBJECT(tt), "link"))
+                set_text_cursor (event->window, GDK_HAND1);
+            else
+                set_text_cursor (event->window, GDK_XTERM);
+
+            g_slist_free (tt_list);
+        }
+        else
+            set_text_cursor (event->window, GDK_XTERM);
+    }
+    else
+        set_text_cursor (event->window, GDK_XTERM);
+
+    return true;
+}
+
+static gboolean
+textview_url_activate (GtkTextTag *tag,
+                       GObject *object,
+                       GdkEvent *event,
+                       GtkTextIter *iter,
+                       gpointer user_data)
+{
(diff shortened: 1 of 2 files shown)


jralls:
I guess including the OS is harmless, but it seems unnecessary to me. Surely the user knows what OS they're running.

gnucash/gnome-utils/gnc-main-window.cpp
```diff
@@ -5276,6 +5276,65 @@ link_button_cb (GtkLinkButton *button, gpointer user_data)
    return TRUE;
 }
 
+static gboolean
+image_button_callback (GtkWidget *event_box,
+                       GdkEventButton *event,
+                       gpointer user_data)
+{
+    GtkClipboard *clipboard = gtk_clipboard_get_default (gdk_display_get_default());
+    GList *entries = nullptr;
```
jralls: This is a C++ file, use a `std::string` to collect the entries and pass `entries.c_str()` to `gtk_clipboard_set_text`.

jralls (changes requested):
I think it would be better to replace the labels and buttons with a [GtkTextView](https://docs.gtk.org/gtk3/class.TextView.html) using markup to make the clickable links. That would remove the need for duplication and let the user select what they want to copy instead of having to copy the whole thing and delete the bits they don't want after pasting.

christopherlam:
> I guess including the OS is harmless, but it seems unnecessary to me. Surely the user knows what OS they're running.

I haven't a major opinion on having the OS... perhaps my main motivation for exposing the env paths was to assist in unusual packages. The user may run on flatpak or docker on ubuntu, and if the envs are deep within the packages, exposing the paths would be useful.

In my real-life job, I've had the pleasure of using weird remote desktop apps whereby I'd log into a win10 desktop somewhere, use it proficiently, launch a different app, use it natively, and find that printing was broken. The cause was, of course, the different app was presented locally with minimal delays, but was running from another windows server somewhere, and didn't access the local (i.e. "remote") printers.

How would the OS be shown if run within WSL2 on Win11?

jralls:
> How would the OS be shown if run within WSL2 on Win11?

When run in an Arch Linux WSL2 VM it says Arch Linux. You'd expect that, the OS inside a VM doesn't know it's not on a dedicated computer.

I wasn't able to get a flatpak built with this PR incorporated but based on the code I'd expect it to return the same as a normal build. We'd still be able to tell it's a flatpak from the GNC_BIN path being /app/bin.

bhardwajs:
Since the OS name is coming from build time and inside WSL `G_OS_WIN32` is not defined, it would be the Linux disto. Would you like me to try building the branch on my WSL instance?

Bob-IT:
> I guess including the OS is harmless, but it seems unnecessary to me. Surely the user knows what OS they're running.

Yes they should, it was just for convenience, you see a lot of requests asking for version and machine OS. This was just a thought about coping the top section and pasting in email, bug or what ever.

Bob-IT:
With a GtkTextview this is what I have come up with, I have left the grid in for comparison. Using link buttons in the grid you had tooltips for the links, is that required for the GtkTextview...

Screenshot_2025-11-26_14-31-32

jralls:
I can't think of any good use for the tooltips. Do you think that it would look better if you made the background blend in with the rest of the dialog?

Bob-IT:
Changes required for the About dialog to appear as above with a textview widget. The last commit changes the background to match rest of the dialog which can be over ridden by the user. This was the only way I could get a good result, tried using the same method as with the tab colors, getting the background color of the dialog and creating some CSS to apply which did work but it then upset the selection colors.

gnucash/gnome-utils/gnc-gnome-utils.c
```diff
@@ -171,6 +171,53 @@ gnc_add_css_file (void)
 
 #ifdef MAC_INTEGRATION
 
+char *gnc_get_os_version (void)
+{
+    NSString *operatingSystemVersionString = [[NSProcessInfo processInfo] operatingSystemVersionString];
+
+    return g_strdup_printf ("Mac %s",[operatingSystemVersionString UTF8String]);
```
jralls: s/Mac/macOS/

gnucash/gnome-utils/gnc-gnome-utils.c
```diff
@@ -171,6 +171,53 @@ gnc_add_css_file (void)
...
+    dwMinorVersion = (DWORD)(HIBYTE(LOWORD(dwVersion)));
+
+    // Get the build number.
+    if (dwVersion < 0x80000000)
+        dwBuild = (DWORD)(HIWORD(dwVersion));
+
+    if (dwBuild >= 22000 && dwMajorVersion == 10)
+        dwMajorVersion = 11;
+
+    return  g_strdup_printf ("Windows Version %ld.%ld (%ld)",
```
jralls: s/Windows/Microsoft Windows/

jralls:
Aside from the OS name changes I like the look though I'm still not convinced that the OS info adds anything useful.

gnucash/gnome-utils/gnc-gnome-utils.c
Bob-IT: dropped

gnucash/gnome-utils/gnc-gnome-utils.c
Bob-IT: dropped

jralls (approved):

Bob-IT:
I forgot to account for using Adwaita-dark theme. Tested on Linux and Windows and looks OK.
Will push this tomorrow.

PR #2157, opened 2025-12-05
Faster guid and other xml speedups

christopherlam:
this does shave off ~5-10%~ 20% xml loading time.

### DIFF
--- a/libgnucash/backend/xml/gnc-budget-xml-v2.cpp
+++ b/libgnucash/backend/xml/gnc-budget-xml-v2.cpp
@@ -85,17 +85,6 @@ gnc_budget_dom_tree_create (GncBudget* bgt)
 }
 
 /***********************************************************************/
-static inline gboolean
-set_string (xmlNodePtr node, GncBudget* bgt,
-            void (*func) (GncBudget* bgt, const gchar* txt))
-{
-    gchar* txt = dom_tree_to_text (node);
-    g_return_val_if_fail (txt, FALSE);
-
-    func (bgt, txt);
-    g_free (txt);
-    return TRUE;
-}
 
 static gboolean
 budget_id_handler (xmlNodePtr node, gpointer bgt)
@@ -109,13 +98,13 @@ budget_id_handler (xmlNodePtr node, gpointer bgt)
 static gboolean
 budget_name_handler (xmlNodePtr node, gpointer bgt)
 {
-    return set_string (node, GNC_BUDGET (bgt), gnc_budget_set_name);
+    return apply_xmlnode_text (gnc_budget_set_name, GNC_BUDGET (bgt), node);
 }
 
 static gboolean
 budget_description_handler (xmlNodePtr node, gpointer bgt)
 {
-    return set_string (node, GNC_BUDGET (bgt), gnc_budget_set_description);
+    return apply_xmlnode_text (gnc_budget_set_description, GNC_BUDGET (bgt), node);
 }
 
 static gboolean
--- a/libgnucash/backend/xml/gnc-entry-xml-v2.cpp
+++ b/libgnucash/backend/xml/gnc-entry-xml-v2.cpp
@@ -225,18 +225,6 @@ struct entry_pdata
 };
 
 static inline gboolean
-set_string (xmlNodePtr node, GncEntry* entry,
-            void (*func) (GncEntry* entry, const char* txt))
-{
-    char* txt = dom_tree_to_text (node);
-    g_return_val_if_fail (txt, FALSE);
-
-    func (entry, txt);
-    g_free (txt);
-    return TRUE;
-}
-
-static inline gboolean
 set_time64 (xmlNodePtr node, GncEntry* entry,
               void (*func) (GncEntry* entry, time64 ts))
 {
@@ -349,7 +337,7 @@ entry_description_handler (xmlNodePtr node, gpointer entry_pdata)
 {
     struct entry_pdata* pdata = static_cast<decltype (pdata)> (entry_pdata);
 
-    return set_string (node, pdata->entry, gncEntrySetDescription);
+    return apply_xmlnode_text (gncEntrySetDescription, pdata->entry, node);
 }
 
 static gboolean
@@ -357,7 +345,7 @@ entry_action_handler (xmlNodePtr node, gpointer entry_pdata)
 {
     struct entry_pdata* pdata = static_cast<decltype (pdata)> (entry_pdata);
 
-    return set_string (node, pdata->entry, gncEntrySetAction);
+    return apply_xmlnode_text (gncEntrySetAction, pdata->entry, node);
 }
 
 static gboolean
@@ -365,7 +353,7 @@ entry_notes_handler (xmlNodePtr node, gpointer entry_pdata)
 {
     struct entry_pdata* pdata = static_cast<decltype (pdata)> (entry_pdata);
 
-    return set_string (node, pdata->entry, gncEntrySetNotes);
+    return apply_xmlnode_text (gncEntrySetNotes, pdata->entry, node);
 }
 
 static gboolean
@@ -405,40 +393,30 @@ static gboolean
 entry_idisctype_handler (xmlNodePtr node, gpointer entry_pdata)
 {
     struct entry_pdata* pdata = static_cast<decltype (pdata)> (entry_pdata);
-    GncAmountType type;
-    char* str;
-    gboolean ret;
-
-    str = dom_tree_to_text (node);
-    g_return_val_if_fail (str, FALSE);
-
-    ret = gncAmountStringToType (str, &type);
-    g_free (str);
-
-    if (ret)
-        gncEntrySetInvDiscountType (pdata->entry, type);
-
-    return ret;
+    auto entry = pdata->entry;
+    auto set_discount_type = [entry](auto str)
+    {
+        GncAmountType type;
+        if (!gncAmountStringToType (str, &type)) return false;
+        gncEntrySetInvDiscountType (entry, type);
+        return true;
+    };
+    return apply_xmlnode_text (set_discount_type, node, FALSE);
 }
 
 static gboolean
 entry_idischow_handler (xmlNodePtr node, gpointer entry_pdata)
 {
     struct entry_pdata* pdata = static_cast<decltype (pdata)> (entry_pdata);
-    GncDiscountHow how;
-    char* str;
-    gboolean ret;
-
-    str = dom_tree_to_text (node);
-    g_return_val_if_fail (str, FALSE);
-
-    ret = gncEntryDiscountStringToHow (str, &how);
-    g_free (str);
-
-    if (ret)
-        gncEntrySetInvDiscountHow (pdata->entry, how);
-
-    return ret;
+    auto entry = pdata->entry;
+    auto set_discount_how = [entry](auto str)
+    {
+        GncDiscountHow how;
+        if (!gncEntryDiscountStringToHow (str, &how)) return false;
+        gncEntrySetInvDiscountHow (entry, how);
+        return true;
+    };
+    return apply_xmlnode_text (set_discount_how, node, FALSE);
 }
 
 static gboolean
@@ -526,20 +504,15 @@ static gboolean
 entry_billpayment_handler (xmlNodePtr node, gpointer entry_pdata)
 {
     struct entry_pdata* pdata = static_cast<decltype (pdata)> (entry_pdata);
-    GncEntryPaymentType type;
-    char* str;
-    gboolean ret;
-
-    str = dom_tree_to_text (node);
-    g_return_val_if_fail (str, FALSE);
-
-    ret = gncEntryPaymentStringToType (str, &type);
-    g_free (str);
-
-    if (ret)
-        gncEntrySetBillPayment (pdata->entry, type);
-
-    return ret;
+    auto entry = pdata->entry;
+    auto set_billpayment = [entry](auto str)
+    {
+        GncEntryPaymentType type;
+        if (!gncEntryPaymentStringToType (str, &type)) return false;
+        gncEntrySetBillPayment (entry, type);
+        return true;
+    };
+    return apply_xmlnode_text (set_billpayment, node, FALSE);
 }
 
 /* The rest of the stuff */
--- a/libgnucash/backend/xml/gnc-transaction-xml-v2.cpp
+++ b/libgnucash/backend/xml/gnc-transaction-xml-v2.cpp
@@ -200,26 +200,6 @@ struct split_pdata
 };
 
 static inline gboolean
-set_spl_string (xmlNodePtr node, Split* spl,
-                void (*func) (Split* spl, const char* txt))
-{
-    if (auto txt = dom_node_to_text (node))
-    {
-        func (spl, txt);
-        return TRUE;
-    }
-
-    gchar* tmp = dom_tree_to_text (node);
-    g_return_val_if_fail (tmp, FALSE);
-
-    func (spl, tmp);
-
-    g_free (tmp);
-
-    return TRUE;
-}
-
-static inline gboolean
 set_spl_gnc_num (xmlNodePtr node, Split* spl,
                  void (*func) (Split* spl, gnc_numeric gn))
 {
@@ -243,35 +223,25 @@ static gboolean
 spl_memo_handler (xmlNodePtr node, gpointer data)
 {
     struct split_pdata* pdata = static_cast<decltype (pdata)> (data);
-    return set_spl_string (node, pdata->split, xaccSplitSetMemo);
+    return apply_xmlnode_text (xaccSplitSetMemo, pdata->split, node);
 }
 
 static gboolean
 spl_action_handler (xmlNodePtr node, gpointer data)
 {
     struct split_pdata* pdata = static_cast<decltype (pdata)> (data);
-    return set_spl_string (node, pdata->split, xaccSplitSetAction);
+    return apply_xmlnode_text (xaccSplitSetAction, pdata->split, node);
 }
 
 static gboolean
 spl_reconciled_state_handler (xmlNodePtr node, gpointer data)
 {
     struct split_pdata* pdata = static_cast<decltype (pdata)> (data);
-
-    if (auto txt = dom_node_to_text (node))
+    auto set_reconciled = [](Split* s, const char *txt)
     {
-        xaccSplitSetReconcile (pdata->split, txt[0]);
-        return TRUE;
-    }
-
-    gchar* tmp = dom_tree_to_text (node);
-    g_return_val_if_fail (tmp, FALSE);
-
-    xaccSplitSetReconcile (pdata->split, tmp[0]);
-
-    g_free (tmp);
-
-    return TRUE;
+        xaccSplitSetReconcile(s, txt[0]);
+    };
+    return apply_xmlnode_text (set_reconciled, pdata->split, node);
 }
 
 static gboolean
@@ -408,29 +378,6 @@ struct trans_pdata
     QofBook* book;
 };
 
-static inline gboolean
-set_tran_string (xmlNodePtr node, Transaction* trn,
-                 void (*func) (Transaction* trn, const char* txt))
-{
-    if (auto txt = dom_node_to_text (node))
-    {
-        func (trn, txt);
-        return TRUE;
-    }
-
-    gchar* tmp;
-
-    tmp = dom_tree_to_text (node);
-
-    g_return_val_if_fail (tmp, FALSE);
-
-    func (trn, tmp);
-
-    g_free (tmp);
-
-    return TRUE;
-}
-
 static gboolean
 set_tran_time64 (xmlNodePtr node, Transaction * trn,
         void (*func) (Transaction *, time64))
@@ -474,7 +421,7 @@ trn_num_handler (xmlNodePtr node, gpointer trans_pdata)
     struct trans_pdata* pdata = static_cast<decltype (pdata)> (trans_pdata);
     Transaction* trn = pdata->trans;
 
-    return set_tran_string (node, trn, xaccTransSetNum);
+    return apply_xmlnode_text (xaccTransSetNum, trn, node);
 }
 
 static gboolean
@@ -501,7 +448,7 @@ trn_description_handler (xmlNodePtr node, gpointer trans_pdata)
     struct trans_pdata* pdata = static_cast<decltype (pdata)> (trans_pdata);
     Transaction* trn = pdata->trans;
 
-    return set_tran_string (node, trn, xaccTransSetDescription);
+    return apply_xmlnode_text (xaccTransSetDescription, trn, node);
 }
 
 static gboolean
--- a/libgnucash/backend/xml/sixtp-dom-parsers.cpp
+++ b/libgnucash/backend/xml/sixtp-dom-parsers.cpp
@@ -47,74 +47,46 @@ dom_node_to_text (xmlNodePtr node) noexcept
 std::optional<GncGUID>
 dom_tree_to_guid (xmlNodePtr node)
 {
-    if (!node->properties)
-    {
+    auto type = xmlGetProp (node, BAD_CAST "type");
+    if (!type)
         return {};
-    }
 
-    if (strcmp ((char*) node->properties->name, "type") != 0)
-    {
-        PERR ("Unknown attribute for id tag: %s",
-              node->properties->name ?
-              (char*) node->properties->name : "(null)");
+    bool ok = !g_strcmp0 ((char*)type, "guid") || !g_strcmp0 ((char*)type, "new");
+
+    xmlFree (type);
+
+    if (!ok)
         return {};
-    }
 
+    auto extract_guid = [](auto str) -> std::optional<GncGUID>
     {
-        char* type;
+        if (GncGUID guid; string_to_guid (str, &guid))
+            return guid;
 
-        type = (char*)xmlNodeGetContent (node->properties->xmlAttrPropertyValue);
+        return {};
+    };
 
-        /* handle new and guid the same for the moment */
-        if ((g_strcmp0 ("guid", type) == 0) || (g_strcmp0 ("new", type) == 0))
-        {
-            GncGUID gid;
-            char* guid_str;
-
-            guid_str = (char*)xmlNodeGetContent (node->xmlChildrenNode);
-            string_to_guid (guid_str, &gid);
-            xmlFree (guid_str);
-            xmlFree (type);
-            return gid;
-        }
-        else
-        {
-            PERR ("Unknown type %s for attribute type for tag %s",
-                  type ? type : "(null)",
-                  node->properties->name ?
-                  (char*) node->properties->name : "(null)");
-            xmlFree (type);
-            return {};
-        }
-    }
+    return apply_xmlnode_text<std::optional<GncGUID>>(extract_guid, node);
 }
 
 static KvpValue*
 dom_tree_to_integer_kvp_value (xmlNodePtr node)
 {
-    gchar* text;
-    gint64 daint;
-    KvpValue* ret = NULL;
-
-    text = dom_tree_to_text (node);
-
-    if (string_to_gint64 (text, &daint))
+    auto node_to_int_kvp = [](auto txt) -> KvpValue*
     {
-        ret = new KvpValue {daint};
-    }
-    g_free (text);
+        if (gint64 daint; string_to_gint64 (txt, &daint))
+            return new KvpValue{daint};
 
-    return ret;
+        return nullptr;
+    };
+    return apply_xmlnode_text<KvpValue*> (node_to_int_kvp, node, nullptr);
 }
 
 template <typename T>
 static bool
 dom_tree_to_num (xmlNodePtr node, std::function<bool(const char*, T*)>string_to_num, T* num_ptr)
 {
-    auto text = dom_tree_to_text (node);
-    auto ret = string_to_num (text, num_ptr);
-    g_free (text);
-    return ret;
+    return apply_xmlnode_text<T>([&](auto txt){ return string_to_num (txt, num_ptr);}, node, false);
 }
 
 gboolean
@@ -138,46 +110,36 @@ dom_tree_to_guint (xmlNodePtr node, guint* i)
 gboolean
 dom_tree_to_boolean (xmlNodePtr node, gboolean* b)
 {
-    gchar* text;
-
-    text = dom_tree_to_text (node);
-    if (g_ascii_strncasecmp (text, "true", 4) == 0)
+    auto set_bool = [b](auto text) -> gboolean
     {
-        *b = TRUE;
-        g_free (text);
-        return TRUE;
-    }
-    else if (g_ascii_strncasecmp (text, "false", 5) == 0)
-    {
-        *b = FALSE;
-        g_free (text);
-        return TRUE;
-    }
-    else
-    {
-        *b = FALSE;
-        g_free (text);
-        return FALSE;
-    }
+        if (g_ascii_strncasecmp (text, "true", 4) == 0)
+        {
+            *b = TRUE;
+            return TRUE;
+        }
+        else if (g_ascii_strncasecmp (text, "false", 5) == 0)
+        {
+            *b = FALSE;
+            return TRUE;
+        }
+        else
+        {
+            *b = FALSE;
+            return FALSE;
+        }
+    };
+    return apply_xmlnode_text<gboolean> (set_bool, node);
 }
 
 static KvpValue*
 dom_tree_to_double_kvp_value (xmlNodePtr node)
 {
-    gchar* text;
-    double dadoub;
-    KvpValue* ret = NULL;
-
-    text = dom_tree_to_text (node);
-
-    if (string_to_double (text, &dadoub))
+    auto node_to_double_kvp = [](auto txt) -> KvpValue*
     {
-        ret = new KvpValue {dadoub};
-    }
-
-    g_free (text);
-
-    return ret;
+        if (double dadoub; string_to_double (txt, &dadoub)) return new KvpValue{dadoub};
+        return nullptr;
+    };
+    return apply_xmlnode_text<KvpValue*> (node_to_double_kvp, node, nullptr);
 }
 
 static KvpValue*
@@ -189,30 +151,18 @@ dom_tree_to_numeric_kvp_value (xmlNodePtr node)
 static KvpValue*
 dom_tree_to_string_kvp_value (xmlNodePtr node)
 {
-    const gchar* datext;
-    KvpValue* ret = NULL;
-
-    datext = dom_tree_to_text (node);
-    if (datext)
+    auto node_to_string_kvp = [](auto txt) -> KvpValue*
     {
-        ret = new KvpValue {datext};
-    }
-
-    return ret;
+        return new KvpValue {g_strdup (txt)};
+    };
+    return apply_xmlnode_text<KvpValue*> (node_to_string_kvp, node, nullptr);
 }
 
 static KvpValue*
 dom_tree_to_guid_kvp_value (xmlNodePtr node)
 {
-    KvpValue* ret = NULL;
-
     auto daguid = dom_tree_to_guid (node);
-    if (daguid)
-    {
-        ret = new KvpValue {guid_copy (&*daguid)};
-    }
-
-    return ret;
+    return daguid ? new KvpValue {guid_copy (&*daguid)} : nullptr;
 }
 
 static KvpValue*
@@ -225,19 +175,8 @@ dom_tree_to_time64_kvp_value (xmlNodePtr node)
 static KvpValue*
 dom_tree_to_gdate_kvp_value (xmlNodePtr node)
 {
-    GDate* date;
-    KvpValue* ret = NULL;
-
-    date = dom_tree_to_gdate (node);
-
-    if (date)
-    {
-        ret = new KvpValue {*date};
-    }
-
-    g_free (date);
-
-    return ret;
+    auto date = dom_tree_to_gdate (node);
+    return date ? new KvpValue {*date} : nullptr;
 }
 
 gboolean
@@ -314,17 +253,8 @@ dom_tree_to_list_kvp_value (xmlNodePtr node)
 static KvpValue*
 dom_tree_to_frame_kvp_value (xmlNodePtr node)
 {
-    KvpFrame* frame;
-    KvpValue* ret = NULL;
-
-    frame = dom_tree_to_kvp_frame (node);
-
-    if (frame)
-    {
-        ret = new KvpValue {frame};
-    }
-
-    return ret;
+    KvpFrame* frame = dom_tree_to_kvp_frame (node);
+    return frame ? new KvpValue {frame} : nullptr;
 }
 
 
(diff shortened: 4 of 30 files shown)


libgnucash/app-utils/gnc-autoclear.cpp
```diff
@@ -79,7 +79,7 @@ using SplitVec = std::vector<Split*>;
 
 struct Solution
 {
-    bool abort = false;
+    std::optional<std::string> abort;
```
jralls: Make `abort` a `const char *` and the two strings `static const char *` members. Saves the string alloc and converting char* -> string and back.

libgnucash/backend/xml/sixtp-dom-parsers.cpp
```diff
@@ -705,77 +707,62 @@ dom_tree_to_commodity_ref_no_engine (xmlNodePtr node, QofBook* book)
...
-
-    table = gnc_commodity_table_get_table (book);
-
-    g_return_val_if_fail (table != NULL, NULL);
-
-    ret =  gnc_commodity_table_lookup (table,
-                                       gnc_commodity_get_namespace (daref),
-                                       gnc_commodity_get_mnemonic (daref));
-
-    gnc_commodity_destroy (daref);
```
christopherlam: This last commit is a good one. Every transaction has a currency node, which would be creating a new gnc_commodity and immediately destroying it. Refactor will avoid this wasteful trip.

libgnucash/backend/xml/sixtp-dom-parsers.cpp
```diff
@@ -92,6 +98,10 @@ dom_tree_to_guid (xmlNodePtr node)
 static KvpValue*
 dom_tree_to_integer_kvp_value (xmlNodePtr node)
 {
+    if (auto txt = dom_node_to_text (node))
+        if (gint64 daint; string_to_gint64 (txt, &daint))
+            return new KvpValue{daint};
+
```
jralls: This is sloppy. At least put the new code after the decls and use `text` instead of creating a new `txt` and `daint`. Better yet would be 
```
 {
    gchar* text{dom_node_to_text (node)};
    gint64 daint;
    gboolean valid_int{FALSE};

    if (text)
        valid_int = string_to_gint64 (text, &daint);
    else
    {
        text = dom_tree_to_text (node);
        valid_int = string_to_gint64 (text, &daint);
        g_free (text);
    }
    if (valid_int)
        return new KvpValue {daint};
    return nullptr
}
```

jralls: Note you can make it even cleaner by changing `string_to_gint64` and friends to take a `const std::string_view` instead of a `char*` and `dom_tree_to_text` to return a `std::string`. Then you can write
```
{
     auto text{dom_node_to_text(node)};
     if (!text)
         text = dom_tree_to_text (node);
     gint64 daint;
     if (string_to_gint64 (text, &daint))
         return new KvpValue {daint};
     return nullptr;
}
```

libgnucash/engine/guid.cpp
```diff
@@ -266,15 +256,9 @@ guid_hash_to_guint (gconstpointer ptr)
...
-
-    guint hash {0};
-    std::for_each (temp.begin (), temp.end (), [&hash] (unsigned char a) {
-        hash <<=4;
-        hash |= a;
-    });
-    return hash;
+    const GncGUID* g = static_cast<const GncGUID*>(ptr);
+    const std::string_view data{reinterpret_cast<const char*>(g->reserved), GUID_DATA_SIZE};
+    return std::hash<std::string_view>{}(data);
```
jralls: `std::hash` produces a `size_t` that fits in an `gint` only on 32-bit systems. You need to change the return type here and in all uses.

jralls: Also it would be better to make a gnc:GUID member function and wrap it for use in gnc-sx-instance-model.c and call it directly in import-pending-matches.cpp.

jralls (changes requested):
Nice optimizations.

libgnucash/backend/xml/sixtp-dom-parsers.cpp
christopherlam: > Note you can make it even cleaner by changing `string_to_gint64` and friends to take a `const std::string_view` instead of a `char*` and `dom_tree_to_text` to return a `std::string`. Then you can write
> 
> ```
> {
>      auto text{dom_node_to_text(node)};
>      if (!text)
>          text = dom_tree_to_text (node);
>      gint64 daint;
>      if (string_to_gint64 (text, &daint))
>          return new KvpValue {daint};
>      return nullptr;
> }
> ```

hmm then `text` will be a `const char*` or `std::string`?

jralls: > hmm then text will be a const char* or std::string?

Oh, right. It would have to be 
```
     std::string tree_text{};
     std::string_view text{dom_node_to_text(node)};
     if (text.empty())
     {
          tree_text = dom_tree_to_text(node);
          text = tree_text;
      }
      ...
```

christopherlam: > > hmm then text will be a const char* or std::string?
> 
> Oh, right. It would have to be
> 
> ```
>      std::string tree_text{};
>      std::string_view text{dom_node_to_text(node)};
>      if (text.empty())
>      {
>           tree_text = dom_tree_to_text(node);
>           text = tree_text;
>       }
>       ...
> ```

Not quite because the dom_node_to_text(node) needs to be freed

jralls: > Not quite because the dom_node_to_text(node) needs to be freed

No it doesn't, it's `reinterpret_cast<const char*>(node->children->content)`. Isn't that why you have separate a separate block with no free for it?

christopherlam: > > Not quite because the dom_node_to_text(node) needs to be freed
> 
> No it doesn't, it's `reinterpret_cast<const char*>(node->children->content)`. Isn't that why you have separate a separate block with no free for it?

oops my bad... I mean `dom_tree_to_text` needs to be `g_free`d.

libgnucash/engine/guid.cpp
christopherlam: > `std::hash` produces a `size_t` that fits in an `gint` only on 32-bit systems. You need to change the return type here and in all uses.

as follows?

```cpp

guint
guid_hash_to_guint (gconstpointer ptr)
{
    if (!ptr)
    {
        PERR ("received nullptr guid pointer.");
        return 0;
    }
    const GncGUID* g = static_cast<const GncGUID*>(ptr);
    const std::string_view data{reinterpret_cast<const char*>(g->reserved), GUID_DATA_SIZE};
    uint64_t h = std::hash<std::string_view>{}(data);
    return static_cast<guint>(h ^ (h >> 32));
}
```

libgnucash/backend/xml/sixtp-dom-parsers.cpp
jralls: It does now, that's why I said

> Note you can make it even cleaner by changing string_to_gint64 and friends to take a const std::string_view instead of a char* and dom_tree_to_text to return a std::string.

libgnucash/engine/guid.cpp
jralls: `h ^ (h >> 32)` is a weird thing to do:
```
>>> hash=6919333181322027406
>>> hex(hash)
'0x600668de4345e18e'
>>> hex(hash >> 32)
'0x600668de'
>>> hex(hash ^ (hash >> 32))
'0x600668de23438950'
```
`6919333181322027406` comes from `[std::hash](https://en.cppreference.com/w/cpp/utility/hash.html)<[std::string_view](https://en.cppreference.com/w/cpp/string/basic_string_view.html)>{}("A"sv)`

You might as well just chop off the top half.

I looked again at the original hash func,
```
guint hash {0};
    std::for_each (temp.begin (), temp.end (), [&hash] (unsigned char a) {
        hash <<=4;
        hash |= a;
    });
```
It also shifts the top half of the GUID to oblivion. 

A GUID is already a random number so simply returning `*(std::reinterpret_cast<guint*>(&g->reserved[12]))` should be a good enough hash for our purposes.

jralls:
Heh. The failing test is the one that #2158 fixes. Arch must have picked up guile 3.0.11.

libgnucash/backend/xml/sixtp-dom-parsers.cpp
christopherlam: > It does now, that's why I said
> 
> > Note you can make it even cleaner by changing string_to_gint64 and friends to take a const std::string_view instead of a char*

this is easily done

>> and dom_tree_to_text to return a std::string.

and this one less so because it can return `nullptr` for invalid conditions, and is heavily used by `backend/xml`. should it return `optional<string>`?

jralls: That's the current preferred way, and you do have some practice with it. ;-)

christopherlam: > That's the current preferred way, and you do have some practice with it. ;-)

Ok this is now present in last commit ... It's technically another layer of copying from xmlChar to char* to `optional<string>` to the engine data though. We can build the string directly by walking the XML tree though.

christopherlam: > > That's the current preferred way, and you do have some practice with it. ;-)
> 
> Ok this is now present in last commit ... It's technically another layer of copying from xmlChar to char* to `optional<string>` to the engine data though. We can build the string directly by walking the XML tree though.

I've reverted the `dom_tree_to_text` change; it does not offer any speedup. The various engine setters (eg entry + kvp) using `dom_node_to_text` seem more effective; it reduces the unwanted temporary char* from 21000 to 350.

jralls: It's not meant to be a speedup, it's to allow cleaner code and safer memory management. As long as it isn't a significant slowdown (and it shouldn't be, see below) it's still better.

> It's technically another layer of copying from xmlChar to char* to optional<string>

It's not. `xmlNodeListGetString` allocates a string and `temp` points at it. The old code allocates a copy with `strdup` and the new code allocates a copy in the `std::string`. No change. It's [theoretically possible](https://en.cppreference.com/w/cpp/string/basic_string/basic_string) to make a std::string class that uses libxml2's allocator and could use the xmlChar* directly from `xmlNodeListGetString`, eliminating the second allocation but I don't think we want to go there.

Note that `return rv` move-constructs the std::optional from rv into the caller's stack frame, see the example in [cpp_reference](https://en.cppreference.com/w/cpp/utility/optional/optional.html).

libgnucash/backend/xml/gnc-transaction-xml-v2.cpp
```diff
@@ -193,6 +193,18 @@ gnc_transaction_dom_tree_create (Transaction* trn)
...
 /***********************************************************************/
 
+template<typename Obj, typename F>
+static inline gboolean
+set_string_from_node(xmlNodePtr node, Obj* obj, F&& func)
+{
+    if (const char* txt = dom_node_to_text(node)) func(obj, txt);
+    else if (gchar* txt = dom_tree_to_text(node)) { func (obj, txt); g_free (txt); }
+    else return FALSE;
+    return TRUE;
```
jralls: A little more white space please.

libgnucash/backend/xml/gnc-budget-xml-v2.cpp
```diff
@@ -89,11 +89,10 @@ static inline gboolean
 set_string (xmlNodePtr node, GncBudget* bgt,
             void (*func) (GncBudget* bgt, const gchar* txt))
 {
-    gchar* txt = dom_tree_to_text (node);
+    auto txt = dom_tree_to_text (node);
     g_return_val_if_fail (txt, FALSE);
 
-    func (bgt, txt);
-    g_free (txt);
+    func (bgt, txt->c_str());
```
jralls: Looks like this and bill-term above are expecting a std::string* from `dom_tree_to_text`.

jralls: Oh, never mind, I missed that this is the commit that adds that.

libgnucash/backend/xml/gnc-entry-xml-v2.cpp
```diff
@@ -401,44 +399,41 @@ entry_idiscount_handler (xmlNodePtr node, gpointer entry_pdata)
...
-    ret = gncAmountStringToType (str, &type);
-    g_free (str);
-
-    if (ret)
-        gncEntrySetInvDiscountType (pdata->entry, type);
-
-    return ret;
+    auto entry = pdata->entry;
+    return parse_and_set (node, [entry](auto str)
+    {
```
jralls: This takes close reading to notice that the braces are the start of the lambda. Maybe declare the lambda first, then call:
```
auto entry = pdata->entry;
auto set_from_entry = 
    [entry](auto str)
    { 
          //...
    };
return parse_and_set (node, set_from_entry);
```

jralls (changes requested):
Please follow the coding guidelines!
```
if (foo)
    bar();
``` 
not `if (foo) bar();` and especially not `if (foo) { bar(); baz(); }`

libgnucash/backend/xml/sixtp-dom-parsers.cpp
```diff
@@ -60,6 +60,11 @@ dom_tree_to_guid (xmlNodePtr node)
         return {};
     }
 
+    if (auto type_str = dom_node_to_text ((xmlNodePtr)node->properties); !strcmp (type_str, "guid"))
```
jralls: Better use `g_strcmp0` here, `type_str` could be NULL and `strcmp` isn't NULL-safe.

christopherlam:
Any thoughts about naming the two templates functions in the header?

jralls:
How about `apply_string` and `apply_xml_node_text`? I'm not keen on starting either name with `gnc` because the function isn't exported outside of the xml backend and `xml` doesn't really belong up front: Neither does anything to the DOM tree and the first one doesn't even have involve xml.

christopherlam:
Ok how about apply_node_text and apply_node_text_to_object

jralls:
> the first one doesn't even have involve xml

Oops, that's wrong, it's the one getting the text from the xmlNode. So `apply_string` is wrong. `apply_node_text` isn't specific enough, it should be `apply_xml_node_text` (or `apply_xmlnode_text` since the first arg is an xmlNode.) The function argument really shouldn't be part of the template pack, only `std::function<T(const char*)>` will work.

The only differences for the second function are the return type (arbitrary in the first, gboolean in the second) and that the function argument is `std::function<gboolean(Obj*, const char*)>`. Using the different function argument also moves you from template type deduction to normal function argument type resolution so you can use the same name for both. I think this would be a good place to do that.

christopherlam:
I think this is nearly complete... The goal was to replicate all `dom_tree_to_text` with new templated functions. Hopefully more eyes == less bugs. My cli to load xml load and run report improves from 5.0s to 4.0s.

christopherlam:
On reflection the helper functions should have their arguments swapped around; should be the following to be more "intuitive"/"natural"/etc. Will try to modify.

`apply_xmlnode_text(f, node, [default])` --> returns `f(node_text);` or `default`
`apply_xmlnode_text(f, obj, node)` --> calls `f(obj, node_text);` returning `true` if ok

similar to scheme `(apply f (list arg1 arg2...))` is equivalent to `(f arg1 arg2...)`

jralls:
> On reflection the helper functions should have their arguments swapped around; should be the following to be more "intuitive"/"natural"/etc.

Makes sense.

jralls:
But you're still passing `func` as an unconstrained template parameter. That works but it will produce hard to understand errors if somebody tries to pass in something that doesn't match the signature of how it's used.

christopherlam:
added thanks to chatgpt...

because std::function has overhead
and function pointer cannot take lambda, and cannot inline
and F&& with static_assert is zero-cost

jralls:
I think chatgpt gave you the wrong answer, or maybe you asked the wrong question.  It's far better for the compiler to say that it can't find a matching `apply_xmlmode_text` at the line with the wrong parameters than to raise an assert fail inside `apply_xmlnode_text` with the source of the problem buried in a note message somewhere after the error.

It also doesn't make sense to pretend the function is generic in a way that it isn't: F is exactly one type, `std::function` in each version of `apply_xmlmode_text`. Tell the compiler that up front. The simplest way to do that is to specify the type in the function arguments without templating. There's a more complicated way using `std::enable_if` on the template paramter, but the extra work and clunkiness doesn't get you anything. (C++20 adds "concepts light" that provides a prettier way to constrain the template type deduction but it's still extra work that doesn't make sense when there's only one type that satisfies the template.
```
template<typename T>
inline T
apply_xmlnode_text (std::function<T, xmlNodePtr, T> f, xmlNodePtr node, T default_val = T{})

template<typename Obj>
inline bool
apply_xmlnode_text (std::function<void, Obj*, const char*> f, Obj* obj, xmlNodePtr node)
```

BTW there's a stray `;` at the end of the first definition.

christopherlam:
it should be `std::function<T(const char*)>` but std::function has runtime overhead, and I can't currently make it compile.

jralls:
```
Template<typename T, typename F, typename std::enable_if<std::is_invocable_r_v<void, F&, Obj*, const char*>, int> == 0>
inline T
apply_xmlnode_text (F&& f, xmlNodePtr node, T default_val = T{})
```
Is the ugly way. It saves the virtual lookup from std::function and won't break inlining.

What compile error are you getting from std::function?

christopherlam:
> ```
> Template<typename T, typename F, typename std::enable_if<std::is_invocable_r_v<void, F&, Obj*, const char*>, int> == 0>
> inline T
> apply_xmlnode_text (F&& f, xmlNodePtr node, T default_val = T{})
> ```
> 
> Is the ugly way. It saves the virtual lookup from std::function and won't break inlining.
> 
> What compile error are you getting from std::function?

```
/home/chris/sources/gnucash/stable/libgnucash/backend/xml/gnc-address-xml-v2.cpp: In function ‘gboolean address_name_handler(xmlNodePtr, gpointer)’:
/home/chris/sources/gnucash/stable/libgnucash/backend/xml/gnc-address-xml-v2.cpp:105:31: error: no matching function for call to ‘apply_xmlnode_text(void (&)(GncAddress*, const char*), GncAddress*&, xmlNode*&)’
  105 |     return apply_xmlnode_text (gncAddressSetName, pdata->address, node);
      |            ~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from /home/chris/sources/gnucash/stable/libgnucash/backend/xml/gnc-address-xml-v2.cpp:36:
/home/chris/sources/gnucash/stable/libgnucash/backend/xml/sixtp-dom-parsers.h:79:1: note: candidate: ‘template<class T> T apply_xmlnode_text(std::function<T(const char*)>, xmlNodePtr, T)’
   79 | apply_xmlnode_text (std::function<T(const char*)> f, xmlNodePtr node, T default_val = T{})
      | ^~~~~~~~~~~~~~~~~~
/home/chris/sources/gnucash/stable/libgnucash/backend/xml/sixtp-dom-parsers.h:79:1: note:   template argument deduction/substitution failed:
/home/chris/sources/gnucash/stable/libgnucash/backend/xml/gnc-address-xml-v2.cpp:105:31: note:   mismatched types ‘std::function<T(const char*)>’ and ‘void (*)(GncAddress*, const char*)’ {aka ‘void (*)(_gncAddress*, const char*)’}
  105 |     return apply_xmlnode_text (gncAddressSetName, pdata->address, node);
      |            ~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/chris/sources/gnucash/stable/libgnucash/backend/xml/sixtp-dom-parsers.h:95:1: note: candidate: ‘template<class Obj> bool apply_xmlnode_text(std::function<void(Obj*, const char*)>, Obj*, xmlNodePtr)’
   95 | apply_xmlnode_text (std::function<void(Obj*, const char*)> f, Obj* obj, xmlNodePtr node)
      | ^~~~~~~~~~~~~~~~~~
/home/chris/sources/gnucash/stable/libgnucash/backend/xml/sixtp-dom-parsers.h:95:1: note:   template argument deduction/substitution failed:
/home/chris/sources/gnucash/stable/libgnucash/backend/xml/gnc-address-xml-v2.cpp:105:31: note:   mismatched types ‘std::function<void(Obj*, const char*)>’ and ‘void (*)(GncAddress*, const char*)’ {aka ‘void (*)(_gncAddress*, const char*)’}
  105 |     return apply_xmlnode_text (gncAddressSetName, pdata->address, node);
      |            ~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/chris/sources/gnucash/stable/libgnucash/backend/xml/gnc-address-xml-v2.cpp: In function ‘gboolean address_addr1_handler(xmlNodePtr, gpointer)’:
/home/chris/sources/gnucash/stable/libgnucash/backend/xml/gnc-address-xml-v2.cpp:113:31: error: no matching function for call to ‘apply_xmlnode_text(void (&)(GncAddress*, const char*), GncAddress*&, xmlNode*&)’
  113 |     return apply_xmlnode_text (gncAddressSetAddr1, pdata->address, node);
      |            ~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/chris/sources/gnucash/stable/libgnucash/backend/xml/sixtp-dom-parsers.h:79:1: note: candidate: ‘template<class T> T apply_xmlnode_text(std::function<T(const char*)>, xmlNodePtr, T)’
   79 | apply_xmlnode_text (std::function<T(const char*)> f, xmlNodePtr node, T default_val = T{})
      | ^~~~~~~~~~~~~~~~~~
```

jralls:
Apparently it's too hard for compilers (clang has the same problem) to both convert a C-style function ptr to std::function and perform template type deduction on the template arguments to std::function in one call. You can work around it by e.g.
```apply_xmlnode_text(static_cast<std::function<void(GncAddress*, const char*)>(gncAddressSetName), pdata->address, node);``` but that's just as ugly and in more places than enable_if on the templates.

I had a couple of errors in the template decls, so here are the tested-correct ones:
```
template <typename T, typename F,
          std::enable_if_t<std::is_invocable_r_v<void, F, const char*>, int> = 0>
inline T
apply_xmlnode_text (F&& f, xmlNodePtr node, T default_val = T{})

template <typename Obj, typename F,
    std::enable_if_t<std::is_invocable_r_v<void, F, Obj*, const char*>, int> = 0>
inline bool
apply_xmlnode_text (F&& f, Obj* obj, xmlNodePtr node)
```

To illustrate the difference in the error messages I mutated the lambda in the second `apply_xmlnode_text` to `    auto set_str = [&f](Obj* obj, auto txt)`. With enable_if the error is
```
/Users/john/Development/Gnucash-git-11.0-arm64/src/gnucash-git/libgnucash/backend/xml/sixtp-dom-parsers.h:110:12: error: no matching function for call to 'apply_xmlnode_text'
  110 |     return apply_xmlnode_text<bool> (set_str, node, false);
      |            ^~~~~~~~~~~~~~~~~~~~~~~~
/Users/john/Development/Gnucash-git-11.0-arm64/src/gnucash-git/libgnucash/backend/xml/gnc-account-xml-v2.cpp:179:12: note: in instantiation of function template specialization 'apply_xmlnode_text<account_s, void (&)(account_s *, const char *), 0>' requested here
  179 |     return apply_xmlnode_text (xaccAccountSetName, pdata->account, node);
      |            ^
/Users/john/Development/Gnucash-git-11.0-arm64/src/gnucash-git/libgnucash/backend/xml/sixtp-dom-parsers.h:80:1: note: candidate template ignored: requirement 'std::is_invocable_r_v<void, (lambda at /Users/john/Development/Gnucash-git-11.0-arm64/src/gnucash-git/libgnucash/backend/xml/sixtp-dom-parsers.h:105:20) &, const char *>' was not satisfied [with T = bool, F = (lambda at /Users/john/Development/Gnucash-git-11.0-arm64/src/gnucash-git/libgnucash/backend/xml/sixtp-dom-parsers.h:105:20) &]
   80 | apply_xmlnode_text (F&& f, xmlNodePtr node, T default_val = T{})
      | ^
/Users/john/Development/Gnucash-git-11.0-arm64/src/gnucash-git/libgnucash/backend/xml/sixtp-dom-parsers.h:100:1: note: candidate template ignored: requirement 'std::is_invocable_r_v<void, (lambda at /Users/john/Development/Gnucash-git-11.0-arm64/src/gnucash-git/libgnucash/backend/xml/sixtp-dom-parsers.h:105:20) &, bool *, const char *>' was not satisfied [with Obj = bool, F = (lambda at /Users/john/Development/Gnucash-git-11.0-arm64/src/gnucash-git/libgnucash/backend/xml/sixtp-dom-parsers.h:105:20) &]
  100 | apply_xmlnode_text (F&& f, Obj* obj, xmlNodePtr node)
      | ^
1 error generated.
```
and with static_assert it's
```
/Users/john/Development/Gnucash-git-11.0-arm64/src/gnucash-git/libgnucash/backend/xml/sixtp-dom-parsers.h:81:20: error: static assertion failed due to requirement 'std::is_invocable_r_v<bool, (lambda at /Users/john/Development/Gnucash-git-11.0-arm64/src/gnucash-git/libgnucash/backend/xml/sixtp-dom-parsers.h:104:20) &, const char *>': apply_xmlnode_text expects a callable F(const char*) -> T
   81 |     static_assert (std::is_invocable_r_v<T, F&, const char*>,
      |                    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/Users/john/Development/Gnucash-git-11.0-arm64/src/gnucash-git/libgnucash/backend/xml/sixtp-dom-parsers.h:109:12: note: in instantiation of function template specialization 'apply_xmlnode_text<bool, (lambda at /Users/john/Development/Gnucash-git-11.0-arm64/src/gnucash-git/libgnucash/backend/xml/sixtp-dom-parsers.h:104:20) &>' requested here
  109 |     return apply_xmlnode_text<bool> (set_str, node, false);
      |            ^
/Users/john/Development/Gnucash-git-11.0-arm64/src/gnucash-git/libgnucash/backend/xml/gnc-account-xml-v2.cpp:179:12: note: in instantiation of function template specialization 'apply_xmlnode_text<account_s, void (&)(account_s *, const char *), 0>' requested here
  179 |     return apply_xmlnode_text (xaccAccountSetName, pdata->account, node);
      |            ^
In file included from /Users/john/Development/Gnucash-git-11.0-arm64/src/gnucash-git/libgnucash/backend/xml/gnc-account-xml-v2.cpp:38:
/Users/john/Development/Gnucash-git-11.0-arm64/src/gnucash-git/libgnucash/backend/xml/sixtp-dom-parsers.h:88:16: error: no matching function for call to object of type '(lambda at /Users/john/Development/Gnucash-git-11.0-arm64/src/gnucash-git/libgnucash/backend/xml/sixtp-dom-parsers.h:104:20)'
   88 |         return f(txt);
      |                ^
/Users/john/Development/Gnucash-git-11.0-arm64/src/gnucash-git/libgnucash/backend/xml/sixtp-dom-parsers.h:104:20: note: candidate function template not viable: requires 2 arguments, but 1 was provided
  104 |     auto set_str = [&f](Obj* obj, auto txt)
      |                    ^    ~~~~~~~~~~~~~~~~~~
/Users/john/Development/Gnucash-git-11.0-arm64/src/gnucash-git/libgnucash/backend/xml/sixtp-dom-parsers.h:91:16: error: no matching function for call to object of type '(lambda at /Users/john/Development/Gnucash-git-11.0-arm64/src/gnucash-git/libgnucash/backend/xml/sixtp-dom-parsers.h:104:20)'
   91 |         return f(txt->c_str());
      |                ^
/Users/john/Development/Gnucash-git-11.0-arm64/src/gnucash-git/libgnucash/backend/xml/sixtp-dom-parsers.h:104:20: note: candidate function template not viable: requires 2 arguments, but 1 was provided
  104 |     auto set_str = [&f](Obj* obj, auto txt)
      |                    ^    ~~~~~~~~~~~~~~~~~~
3 errors generated.
```

**Comment by christopherlam** [comment]
thanks this did it...

**Review by jralls — APPROVED** [review]
OK, merge when ready.

**Comment by jralls** [comment]
@christopherlam You don't have to rebase before merging unless there are conflicts that you don't want to resolve in the merge commit.

**Comment by christopherlam** [comment]
ugh. I think we need to revert `apply_xmlnode_text` and `[backend/xml] dom_tree_to_text returns std::optional<std::string>`. there's leaks. can't figure it out.
[Valgrind memory leak analysis.html](https://github.com/user-attachments/files/24157336/Valgrind.memory.leak.analysis.html)

**Comment by jralls** [comment]
Most of what ChatGPT is complaining about is either nonsensical (this isn't multi-threaded code so `apply_xmlnode_text` can't return early. There's no path where it can call what ChatGPT is calling a handler more than once.)
It complains
```
f (auto txt = dom_node_to_text(node))
    ...
else
    txt = dom_tree_to_text(node);

This means:
   *  borrowed vs owned strings
   *  different lifetimes
   *  different failure behaviour
```
That ignores that in the second case we're passing txt->c_str(), i.e. a borrowed string, to `f()` and that in all cases `f()` immediately copies the `char*` and then the std::string if there is one goes out of scope so there's no lifetime issue.

Different failure behavior is a valid complaint and might be the cause of some of the leaks.
For example the lambda in `owner_type_handler` returns `true` even if the type doesn't match one of the choices; It doesn't know that that's defensive and that it's not actually possible. Nevertheless it would be reasonable to have the lambda throw a `std::runtime_error` if it falls through. Looks like a similar problem in `dom_tree_to_guid` and perhaps others. 

I dunno how you constructed valgrind_memory_leak_analysis.html but the valgrind output is rendered illegible for lack of being wrapped in `<pre>...</pre>`: HTML converts all whitespace including newlines to a single space so the output is rendered as a single line. I suggest you go through the changes and restore all of the error conditions to the lambdas and see if that fixes the leaks. The exceptions will make the resulting object code a little bigger and maybe a little slower. If it turns out to be a lot slower then we can consider another error handling mechanism: Maybe the lambdas could capture a bool& success and set it to true if they succeed. Also any logging in the original should also be in the lambdas.

**Comment by christopherlam** [comment]
Maybe best revert for release, I'll do tomorrow.

**Comment by jralls** [comment]
How about just reverting the replacements where the lambda doesn't do what the original code did?

**Comment by christopherlam** [comment]
> How about just reverting the replacements where the lambda doesn't do what the original code did?

Uh because I'm not confident I can cleanly identify the offending functions, and thought it'd be cleaner to go back to the drawing board

**Comment by jralls** [comment]
I've taken a [stab at a partial revert](https://github.com/jralls/gnucash/tree/revert-xmlnode) but it still needs some cleanup to build. I'll finish that up tomorrow morning and you can try it in Valgrind.

Was yesterday's Valgrind run the test suite or did you open GnuCash and try some stuff?

**Comment by christopherlam** [comment]
> I've taken a [stab at a partial revert](https://github.com/jralls/gnucash/tree/revert-xmlnode) but it still needs some cleanup to build. I'll finish that up tomorrow morning and you can try it in Valgrind.
> 
> Was yesterday's Valgrind run the test suite or did you open GnuCash and try some stuff?

It's from a regular datafile.

**Comment by christopherlam** [comment]
ok as of current `stable` theres some valgrind logs

```

==465037== 768 (464 direct, 304 indirect) bytes in 2 blocks are definitely lost in loss record 21,595 of 22,755
==465037==    at 0x5C5DCCD: g_type_create_instance (in /gnu/store/r99aqxzrfkhcijp2ni72nzcahynb9kkv-glib-2.83.3/lib/libgobject-2.0.so.0.8303.0)
==465037==    by 0x5C4222F: ??? (in /gnu/store/r99aqxzrfkhcijp2ni72nzcahynb9kkv-glib-2.83.3/lib/libgobject-2.0.so.0.8303.0)
==465037==    by 0x5C434DD: g_object_new_with_properties (in /gnu/store/r99aqxzrfkhcijp2ni72nzcahynb9kkv-glib-2.83.3/lib/libgobject-2.0.so.0.8303.0)
==465037==    by 0x5C44310: g_object_new (in /gnu/store/r99aqxzrfkhcijp2ni72nzcahynb9kkv-glib-2.83.3/lib/libgobject-2.0.so.0.8303.0)
==465037==    by 0x62BC4C2: gncInvoiceCreate (gncInvoice.c:319)
==465037==    by 0x5DD65D4: entry_invoice_handler(_xmlNode*, void*) (gnc-entry-xml-v2.cpp:554)
==465037==    by 0x5DEB266: gnc_xml_set_data (sixtp-dom-parsers.cpp:694)
==465037==    by 0x5DEB266: dom_tree_generic_parse(_xmlNode*, dom_tree_handler*, void*) (sixtp-dom-parsers.cpp:725)
==465037==    by 0x5DD625C: dom_tree_to_entry (gnc-entry-xml-v2.cpp:684)
==465037==    by 0x5DD625C: gnc_entry_end_handler(void*, _GSList*, _GSList*, void*, void*, void**, char const*) (gnc-entry-xml-v2.cpp:732)
==465037==    by 0x5DEFA67: sixtp_sax_end_handler(void*, unsigned char const*) (sixtp.cpp:540)
==465037==    by 0x5F37BC7: ??? (in /gnu/store/71v7zvadm07b7zh1z22d9242xln0q843-libxml2-2.14.6/lib/libxml2.so.16.0.6)
==465037==    by 0x5F384B6: ??? (in /gnu/store/71v7zvadm07b7zh1z22d9242xln0q843-libxml2-2.14.6/lib/libxml2.so.16.0.6)
==465037==    by 0x5F3E137: ??? (in /gnu/store/71v7zvadm07b7zh1z22d9242xln0q843-libxml2-2.14.6/lib/libxml2.so.16.0.6)
==465037==


==465037== 288 (224 direct, 64 indirect) bytes in 4 blocks are definitely lost in loss record 20,876 of 22,755
==465037==    at 0x48457B2: malloc (in /gnu/store/3i5px8576v3zw9836cz5ji1zk1ffbyki-valgrind-3.26.0/libexec/valgrind/vgpreload_memcheck-amd64-linux.so)
==465037==    by 0x693C27D: g_malloc (in /gnu/store/r99aqxzrfkhcijp2ni72nzcahynb9kkv-glib-2.83.3/lib/libglib-2.0.so.0.8303.0)
==465037==    by 0x6912819: ??? (in /gnu/store/r99aqxzrfkhcijp2ni72nzcahynb9kkv-glib-2.83.3/lib/libglib-2.0.so.0.8303.0)
==465037==    by 0x6912E99: ??? (in /gnu/store/r99aqxzrfkhcijp2ni72nzcahynb9kkv-glib-2.83.3/lib/libglib-2.0.so.0.8303.0)
==465037==    by 0x5C40994: ??? (in /gnu/store/r99aqxzrfkhcijp2ni72nzcahynb9kkv-glib-2.83.3/lib/libgobject-2.0.so.0.8303.0)
==465037==    by 0x5C41017: g_object_unref (in /gnu/store/r99aqxzrfkhcijp2ni72nzcahynb9kkv-glib-2.83.3/lib/libgobject-2.0.so.0.8303.0)
==465037==    by 0x62D7C90: qof_commit_edit_part2 (qofinstance.cpp:1035)
==465037==    by 0x5DDD11F: taxtable_guid_handler(_xmlNode*, void*) (gnc-tax-table-xml-v2.cpp:261)
==465037==    by 0x5DEB266: gnc_xml_set_data (sixtp-dom-parsers.cpp:694)
==465037==    by 0x5DEB266: dom_tree_generic_parse(_xmlNode*, dom_tree_handler*, void*) (sixtp-dom-parsers.cpp:725)
==465037==    by 0x5DDCB23: dom_tree_to_taxtable (gnc-tax-table-xml-v2.cpp:378)
==465037==    by 0x5DDCB23: gnc_taxtable_end_handler(void*, _GSList*, _GSList*, void*, void*, void**, char const*) (gnc-tax-table-xml-v2.cpp:418)
==465037==    by 0x5DEFA67: sixtp_sax_end_handler(void*, unsigned char const*) (sixtp.cpp:540)
==465037==    by 0x5F37BC7: ??? (in /gnu/store/71v7zvadm07b7zh1z22d9242xln0q843-libxml2-2.14.6/lib/libxml2.so.16.0.6)
==465037== 


==465037== 48 bytes in 1 blocks are definitely lost in loss record 13,283 of 22,755
==465037==    at 0x4845F18: operator new(unsigned long) (in /gnu/store/3i5px8576v3zw9836cz5ji1zk1ffbyki-valgrind-3.26.0/libexec/valgrind/vgpreload_memcheck-amd64-linux.so)
==465037==    by 0x62D5C34: qof_instance_init(QofInstance_s*) (qofinstance.cpp:262)
==465037==    by 0x5C5DBD2: g_type_create_instance (in /gnu/store/r99aqxzrfkhcijp2ni72nzcahynb9kkv-glib-2.83.3/lib/libgobject-2.0.so.0.8303.0)
==465037==    by 0x5C4222F: ??? (in /gnu/store/r99aqxzrfkhcijp2ni72nzcahynb9kkv-glib-2.83.3/lib/libgobject-2.0.so.0.8303.0)
==465037==    by 0x5C434DD: g_object_new_with_properties (in /gnu/store/r99aqxzrfkhcijp2ni72nzcahynb9kkv-glib-2.83.3/lib/libgobject-2.0.so.0.8303.0)
==465037==    by 0x5C44310: g_object_new (in /gnu/store/r99aqxzrfkhcijp2ni72nzcahynb9kkv-glib-2.83.3/lib/libgobject-2.0.so.0.8303.0)
==465037==    by 0x62C4EE2: gncTaxTableCreate (gncTaxTable.c:411)
==465037==    by 0x5DDCFBA: set_parent_child(_xmlNode*, taxtable_pdata*, void (*)(_gncTaxTable*, _gncTaxTable*)) (gnc-tax-table-xml-v2.cpp:239)
==465037==    by 0x5DEB266: gnc_xml_set_data (sixtp-dom-parsers.cpp:694)
==465037==    by 0x5DEB266: dom_tree_generic_parse(_xmlNode*, dom_tree_handler*, void*) (sixtp-dom-parsers.cpp:725)
==465037==    by 0x5DDCB23: dom_tree_to_taxtable (gnc-tax-table-xml-v2.cpp:378)
==465037==    by 0x5DDCB23: gnc_taxtable_end_handler(void*, _GSList*, _GSList*, void*, void*, void**, char const*) (gnc-tax-table-xml-v2.cpp:418)
==465037==    by 0x5DEFA67: sixtp_sax_end_handler(void*, unsigned char const*) (sixtp.cpp:540)
==465037==    by 0x5F37BC7: ??? (in /gnu/store/71v7zvadm07b7zh1z22d9242xln0q843-libxml2-2.14.6/lib/libxml2.so.16.0.6)
==465037== 
```

and I missed out on `g_date_free` in `dom_tree_to_gdate_kvp_value`:

```diff
modified   libgnucash/backend/xml/sixtp-dom-parsers.cpp
@@ -176,7 +176,10 @@ static KvpValue*
 dom_tree_to_gdate_kvp_value (xmlNodePtr node)
 {
     auto date = dom_tree_to_gdate (node);
-    return date ? new KvpValue {*date} : nullptr;
+    if (!date) return nullptr;
+    auto rv = new KvpValue {*date};
+    g_date_free (date);
+    return rv;
 }
```

jralls:
The [partial revert\(https://github.com/jralls/gnucash/tree/revert-xmlnode) is cleaned up so you can try valgrind on it. 

Are those 3 the only leaks in stable?

jralls:
Are you sure that those leaks are new? None of them look to me like they have anything to do with this refactor.

jralls:
I think this is the cause of #3:
```
static void
gnc_taxtable_finalize(GObject* ttp)
{
    G_OBJECT_CLASS(gnc_taxtable_parent_class)->dispose(ttp);
}
```
That should be *finalize*, not *dispose*.

PR #2158, opened 2025-12-07
[Bug 799702](https://bugs.gnucash.org/show_bug.cgi?id=799702)

jralls:
Guile 3.0.11 includes a [complete rewrite of SRFI-64](https://codeberg.org/guile/guile/commit/ad90f45a8c4fd00add44c214863850a425f787a0) that is a bit more strict in a few places and that broke the build and a few tests.

A possibly separate issue is that if a test raises an exception the runner no longer sets the expected and actual values on the results. I've extended `gnc:test-runner` to report the exception kind if there is one. It also no longer accepts an empty expected kind nor `'error` to mean any exception. Use `#t` for that, and `#f` if it's not actually supposed to throw.

Finally, and I'm not completely sure about this, it seems that the old implementation would pass `(test-equal "foo" #f (bar baz))` if `(bar baz)` raises an exception; we had a few of the form `(cdr #f)` that throw 'wrong-type-arg. The last commit changes those tests to `test-error`.

### DIFF
--- a/bindings/guile/test/srfi64-extras.scm
+++ b/bindings/guile/test/srfi64-extras.scm
@@ -34,11 +34,12 @@
         (case (test-result-kind runner)
           ((pass xpass) (set! num-passed (1+ num-passed)))
           ((fail xfail)
-           (if (test-result-ref runner 'expected-value)
+           (if (not (equal? 'no (test-result-ref runner 'expected-value 'no)))
                (format #t "~a\n -> expected: ~s\n -> obtained: ~s\n"
                        (string-join (test-runner-group-path runner) "/")
                        (test-result-ref runner 'expected-value)
-                       (test-result-ref runner 'actual-value)))
+                       (test-result-ref runner 'actual-value))
+               (format #t "~%~a~%" (test-result-alist runner)))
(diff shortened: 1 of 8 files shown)


christopherlam:
I don't know how to install guile 3.0.11 but if they pass then they must be ok?

jralls:
> I don't know how to install guile 3.0.11 but if they pass then they must be ok?

Not that hard to build on Unix, just the usual configure/make/make install. But they pass now because I changed the tests to make them pass. Are those changes OK, especially the `test-equal #f` to `test-error 'wrong-type-arg` changes?

christopherlam:
> > I don't know how to install guile 3.0.11 but if they pass then they must be ok?
> 
> Not that hard to build on Unix, just the usual configure/make/make install. But they pass now because I changed the tests to make them pass. Are those changes OK, especially the `test-equal #f` to `test-error 'wrong-type-arg` changes?

From distant memory, and testing with old srfi-64 just now, yes `test-equal` would return #f for a scheme exception, and changing to `test-error` seems right.

jralls:
fast-forward merged.

PR #2159, opened 2025-12-10
Bug 799701 - SX Editor error When Frequency=Once

Bob-IT:
When the frequency is once, in the SX edit frequency tab the scheduled date is not marked. This is due to the `first_date` variable not being set and subsequent test for it being valid causing the calendar to be cleared.

This can be seen in the image below...

bug799701

Note, I will change bug title to reflect commit before pushing.

### DIFF
--- a/gnucash/gnome/dialog-sx-editor.c
+++ b/gnucash/gnome/dialog-sx-editor.c
@@ -1662,9 +1662,21 @@ gnc_sxed_update_cal (GncSxEditorDialog *sxed)
 
     if (!g_date_valid (&first_date))
     {
+        /* Note: There are no recurrences for PERIOD_NONE and on initial setting
+         * of PERIOD_WEEKLY (no days set), so still need to 'do nothing' */
+        gboolean do_nothing = TRUE;
+        if (recurrences)
+        {
+            Recurrence *r = g_list_nth_data (recurrences, 0);
+            if (r && r->ptype == PERIOD_ONCE)
+                do_nothing = FALSE;
+        }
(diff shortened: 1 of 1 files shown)


gnucash/gnome/dialog-sx-editor.c
```diff
@@ -1662,9 +1662,20 @@ gnc_sxed_update_cal (GncSxEditorDialog *sxed)
...
+        {
+            GList *r_iter = g_list_first (recurrences);
+            Recurrence *r = r_iter->data;
+            if (r->ptype == PERIOD_ONCE)
+                do_nothing = FALSE;
+        }
         /* Nothing to do. */
-        gnc_dense_cal_store_clear (sxed->dense_cal_model);
-        goto cleanup;
+        if (do_nothing)
```
jralls, line 1675: Suggestion:
```
if (recurrences)
{
    Recurrence *r = g_list_nth_data(recurrences, 0);
    if (r && r->ptype != PERIOD_ONCE)
    {
        gnc_dense_cal_store_clear (sxed->dense_cal_model);
        goto cleanup;
     }
 }
```

Bob-IT: I had forgotten about nth_data so have changed to use that.

The reason for using `gboolean do_nothing` is that `recurrences` is not set for a couple of runs, for `PERIOD_NONE` and `PERIOD_WEEKLY` when days have not been set. I added a note.

jralls: OK.

PR #2160, opened 2025-12-13
[gnc-main-window.cpp] addendum to ed03e61dab cancel on click+drag

christopherlam:
@Bob-IT making the env paths selectable is ok however I think we should cancel activating url if a selection is active.


Bob-IT:
Seems reasonable to me.

PR #2165, opened 2025-12-31
Bug 799706 - Rename Namespace

Bob-IT:
Add a rename namespace button to the Security editor that is enabled by the selection of a namespace row in the tree view.

### DIFF
--- a/gnucash/gnome/dialog-commodities.cpp
+++ b/gnucash/gnome/dialog-commodities.cpp
@@ -28,6 +28,7 @@
 #include <glib/gi18n.h>
 
 #include "dialog-commodity.h"
+#include "gnc-commodity.hpp"
 #include "dialog-utils.h"
 #include "gnc-commodity.h"
 #include "gnc-component-manager.h"
@@ -62,6 +63,7 @@ typedef struct
     GtkWidget * edit_button;
     GtkWidget * remove_button;
     gboolean    show_currencies;
+    GtkWidget * rename_namespace_button;
 
     gboolean is_new;
 } CommoditiesDialog;
@@ -74,6 +76,9 @@ void gnc_commodities_dialog_add_clicked (GtkWidget *widget, gpointer data);
 void gnc_commodities_dialog_edit_clicked (GtkWidget *widget, gpointer data);
 void gnc_commodities_dialog_remove_clicked (GtkWidget *widget, gpointer data);
 void gnc_commodities_dialog_close_clicked (GtkWidget *widget, gpointer data);
+
+void gnc_commodities_dialog_rename_namespace_clicked (GtkWidget *widget, gpointer data);
+
 void gnc_commodities_show_currencies_toggled (GtkToggleButton *toggle, CommoditiesDialog *cd);
 }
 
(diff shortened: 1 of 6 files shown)


libgnucash/engine/gnc-commodity.cpp
```diff
@@ -2134,6 +2134,43 @@ gnc_commodity_table_add_namespace(gnc_commodity_table * table,
...
+                                     const char * namespace_name,
+                                     const char * new_namespace_name)
+{
+    if (!table || !namespace_name || !new_namespace_name)
+        return false;
+
+    if (g_strcmp0(namespace_name, new_namespace_name) == 0)
+        return false;
+
+    gnc_commodity_namespace *ns = gnc_commodity_table_find_namespace(table, namespace_name);
```
jralls: This is a C++ file, use `auto`.

libgnucash/engine/gnc-commodity.cpp
```diff
@@ -2134,6 +2134,43 @@ gnc_commodity_table_add_namespace(gnc_commodity_table * table,
...
+
+    if ((g_strcmp0 (new_namespace_name, GNC_COMMODITY_NS_ISO) == 0) ||
+        (g_strcmp0 (new_namespace_name, GNC_COMMODITY_NS_ISO_GUI) == 0) ||
+        (g_strcmp0 (new_namespace_name, _(GNC_COMMODITY_NS_ISO_GUI)) == 0))
+        return false;
```
jralls: Add this to the same-namespace check above and add the existing namespace check from the next line. Then you can write the if statement as
```
if (auto ns = gnc_commodity_table_find_namespace(table, namespace_name); ns)
```

gnucash/gnome/dialog-commodities.cpp
```diff
@@ -270,6 +274,68 @@ gnc_commodities_dialog_close_clicked (GtkWidget *widget, gpointer data)
...
+void
+gnc_commodities_dialog_rename_namespace_clicked (GtkWidget *widget, gpointer data)
+{
+    auto cd = static_cast<CommoditiesDialog*>(data);
+    gnc_commodity_namespace *ns = gnc_tree_view_commodity_get_selected_namespace (cd->commodity_tree);
+
+    if (!ns)
+        return;
+
+    const char *ns_name = gnc_commodity_namespace_get_name (ns);
```
jralls: `auto`

gnucash/gnome/dialog-commodities.cpp
```diff
@@ -270,6 +274,68 @@ gnc_commodities_dialog_close_clicked (GtkWidget *widget, gpointer data)
...
+    // Set our parent
+    gtk_window_set_transient_for (GTK_WINDOW(dialog),
+                                  GTK_WINDOW(gtk_widget_get_toplevel(widget)));
+
+    gtk_builder_connect_signals_full (builder, gnc_builder_connect_full_func, nullptr);
+    g_object_unref (G_OBJECT(builder));
+
+    gtk_dialog_set_default_response (GTK_DIALOG(dialog), GTK_RESPONSE_CANCEL);
+
+    gboolean retval = false;
```
jralls: The function doesn't return a value so you shouldn't have a variable called `retval`. `label_reset` indicates its meaning better.  Since it's for local use it should be a bool instead of a gboolean.

libgnucash/engine/gnc-commodity.cpp
```diff
@@ -2134,6 +2134,43 @@ gnc_commodity_table_add_namespace(gnc_commodity_table * table,
     return ns;
 }
 
+gboolean
+gnc_commodity_table_rename_namespace(const gnc_commodity_table * table,
+                                     const char * namespace_name,
+                                     const char * new_namespace_name)
+{
+    if (!table || !namespace_name || !new_namespace_name)
+        return false;
```
jralls: Either change the return type to bool (since this is called from a C++ file) or the return value to `FALSE`.

gnucash/gnome/dialog-commodities.cpp
```diff
@@ -270,6 +274,68 @@ gnc_commodities_dialog_close_clicked (GtkWidget *widget, gpointer data)
...
+                                                                       new_ns_name);
+            if (rename_ok)
+            {
+                qof_book_mark_session_dirty (gnc_get_current_book());
+                retval = true;
+            }
+            else
+                gtk_label_set_text (GTK_LABEL(label),
+                                    _("Rename failed, possibly new name exists"));
+        }
```
jralls, line 330: add `else gtk_label_set_text (label, _("No new name"));`

gnucash/gnome/dialog-commodities.cpp
```diff
@@ -270,6 +274,68 @@ gnc_commodities_dialog_close_clicked (GtkWidget *widget, gpointer data)
...
+    gtk_window_set_transient_for (GTK_WINDOW(dialog),
+                                  GTK_WINDOW(gtk_widget_get_toplevel(widget)));
+
+    gtk_builder_connect_signals_full (builder, gnc_builder_connect_full_func, nullptr);
+    g_object_unref (G_OBJECT(builder));
+
+    gtk_dialog_set_default_response (GTK_DIALOG(dialog), GTK_RESPONSE_CANCEL);
+
+    gboolean retval = false;
+    while (gtk_dialog_run (GTK_DIALOG(dialog)) == GTK_RESPONSE_OK)
```
jralls: `while (gtk_dialog_run (GTK_DIALOG(dialog)) == GTK_RESPONSE_OK && !label_reset)` instead of the conditional `break` at the end.

jralls (changes requested):
Generally good, some details need polish.

libgnucash/engine/gnc-commodity.cpp
Bob-IT: ok

gnucash/gnome/dialog-commodities.cpp
Bob-IT: OK

gnucash/gnome/dialog-commodities.cpp
Bob-IT: Good idea, the message was an after thought.

gnucash/gnome/dialog-commodities.cpp
Bob-IT: Needed to change the order to get this to work.

libgnucash/engine/gnc-commodity.cpp
Bob-IT: I think I have changed this but may need some more work.

libgnucash/engine/gnc-commodity.cpp
Bob-IT: I am not sure I have done what you were thinking but have made some changes.

gnucash/gnome/dialog-commodities.cpp
```diff
@@ -270,6 +275,68 @@ gnc_commodities_dialog_close_clicked (GtkWidget *widget, gpointer data)
...
+
+        if (new_ns_name && *new_ns_name)
+        {
+            rename_ok = gnc_commodity_table_rename_namespace (commodity_table,
+                                                              ns_name,
+                                                              new_ns_name);
+            if (rename_ok)
+            {
+                qof_book_mark_session_dirty (gnc_get_current_book());
+                rename_ok = true;
```
Bob-IT: Just noticed this, this line is not required so will remove.

libgnucash/engine/gnc-commodity.cpp
jralls: What I had in mind is to do all of the `new_namespace` checks at once:
```
if (!table || !namespace_name || !new_namespace_name ||
    (g_strcmp0 (namespace_name, new_namespace_name) == 0) ||
    (g_strcmp0 (new_namespace_name, GNC_COMMODITY_NS_ISO_GUI) == 0) ||
    (g_strcmp0 (new_namespace_name, _(GNC_COMMODITY_NS_ISO_GUI)) == 0) ||
    gnc_commodity_table_find_namespace(table, namespace_name))
    return false;

if (auto ns = gnc_commodity_table_find_namespace(table, namespace_name); ns)
    {
```
Compilers usually want to see double parens when there's an assignment in an if statement, so either double parens
```
if ((gnc_commodity_table_find_namespace(table, namespace_name)))
```
or the assignment and condition separated by `;` (added in C11 and C++11) is needed to avoid a warning.

It's maybe a little clearer to have all of the `return false` conditions together so
```
auto ns = gnc_commodity_table_find_namespace(table, namespace_name);
if (!ns)
   return false;

ns->name = CACHE_INSERT(static_cast<const char*>(new_namespace_name));
...
```
might be better.

Bob-IT: OK.

jralls (approved):

PR #2167, opened 2026-01-12
Bug 799705 - Remove Old Prices by Namespace

Bob-IT:
Adds the namespace to the list view and the next commit add a namespace combo so that you can select a specific namespace. the combo also has an entry so that you can type the namespace if known.

I will probably squash these commits when ready.


jralls (approved):
LGTM.

PR #2168, opened 2026-01-12
Account tab -- show/sort earliest date

christopherlam:
Adds `earliest date` which usually reflects the account opening date.

Will be useful when we have multiple loan or asset accounts.

<sub><sup>very useful for me</sup></sub>


jralls:
Harmless, but I don't understand the use case.

christopherlam:
Because I've managed books for parents who are super conservative, and maintain a bunch of CDs which renew onto new CDs for a number of years. So there's a lot of matured $0 CDs and a few active ones.

Also I have a large loan which is being split over time into many small loans, as a tax strategy.

For both above it'll be useful to keep it in strict opening-date order without needing to rely on the account->name or account->code.

jralls:
Hmm. I can see that being useful for reports where you can limit it to just the CD and bond accounts. FWIW I mark such accounts hidden and placeholder so I don't see them in the account tree or pickers either.

Sorting the whole account tree by earliest date seems of limited use. It might make sense for some investments but I'd think it would make life harder for everything else. For CDs and bonds don't you care more about maturity than purchase anyway? 

BTW `gnc_account_get_earliest_date` belongs in libgnucash/engine/Account.cpp. It has nothing to do with the UI.

jralls (approved):
OK

PR #2169, opened 2026-01-15
Implement full test coverage of libgnucash/engine/qofid

stefan-koch-micro:
NOTE:  This does not have full coverage because the "if (!target || !ent)" body in the
collection_compare_cp function cannot be reached. There is too  much safety in the rest of
the system to test this error.

NOTE: I made this a separate test from the test-engine so that I was able to check that it
by itself could test the full coverage of the qofid.cpp file. If it was part of the larger
test, I could have missed some parts that were covered incidentally elsewhere.

Also: 
1. Remove unused qof_collection_from_glist function
2. Explicitly set the is_dirty attribute in qof_collection_new 
3. Fix undefined behaviour in collection_compare_cb function

### DIFF
--- a/libgnucash/engine/test/test-qofid.cpp
+++ b/libgnucash/engine/test/test-qofid.cpp
@@ -134,6 +134,10 @@ TEST(QofIDTest, collection_compare)
     qof_instance_set_guid(QOF_INSTANCE(book4), guid_null());
     EXPECT_EQ(-1, qof_collection_compare(target, merge));
     qof_instance_set_guid(QOF_INSTANCE(book4), old_guid);
+
+    qof_collection_destroy(target);
+    qof_collection_destroy(merge);
+    qof_collection_destroy(incompat);
 }
 
 static void cb_visit(GncJob* job, std::vector<const char*>* results_ptr)
@@ -182,6 +186,8 @@ TEST(QofIDTest, collection_foreach)
     EXPECT_STREQ(results[1], "ggg");
     EXPECT_STREQ(results[2], "qqq");
     EXPECT_STREQ(results[3], "zzz");
+
+    qof_collection_destroy(col);
 }
 
 TEST(QofIDTest, collection_print_dirty)
@@ -202,4 +208,5 @@ TEST(QofIDTest, collection_print_dirty)
     output = testing::internal::GetCapturedStdout();
     EXPECT_EQ(output.find("Book collection is dirty."), size_t(0));
     qof_collection_mark_clean(col);
+    qof_collection_destroy(col);
 }
--- a/libgnucash/engine/test/test-qofid.cpp
+++ b/libgnucash/engine/test/test-qofid.cpp
@@ -1,0 +1,205 @@
+/********************************************************************
+ * test-qofid.c: GLib g_test test suite for qofobject.c.	    *
+ * Copyright 2025 Stefan Koch <stefan.koch.micro@gmail.com>	    *
+ *                                                                  *
+ * This program is free software; you can redistribute it and/or    *
+ * modify it under the terms of the GNU General Public License as   *
+ * published by the Free Software Foundation; either version 2 of   *
+ * the License, or (at your option) any later version.              *
+ *                                                                  *
+ * This program is distributed in the hope that it will be useful,  *
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of   *
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the    *
+ * GNU General Public License for more details.                     *
+ *                                                                  *
+ * You should have received a copy of the GNU General Public License*
+ * along with this program; if not, contact:                        *
+ *                                                                  *
+ * Free Software Foundation           Voice:  +1-617-542-5942       *
+ * 51 Franklin Street, Fifth Floor    Fax:    +1-617-542-2652       *
+ * Boston, MA  02110-1301,  USA       gnu@gnu.org                   *
+********************************************************************/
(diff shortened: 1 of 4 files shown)


libgnucash/engine/test/test-qofid.cpp
```diff
@@ -0,0 +1,205 @@
...
+********************************************************************/
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wcpp"
+#include <gtest/gtest.h>
+#pragma GCC diagnostic pop
+#include <glib.h>
+#include <algorithm>
+#include <vector>
+
+#include "../qofid.h"
```
jralls: No need for the `..` path: `gnc_engine_INCLUDES` contains `MODULEPATH` and that's set to `${CMAKE_SOURCE_DIRECTORY}/libgnucash/engine`. You can say `#include <qofid.h>`, using angle-brackets so as not to waste time searching `test/` before walking the include path.

libgnucash/engine/test/test-qofid.cpp
```diff
@@ -0,0 +1,205 @@
...
+{
+    return g_strcmp0(gncJobGetID(job1), gncJobGetID(job2));
+}
+
+static bool is_in(const char* str, const std::vector<const char*> str_vect)
+{
+    for(auto id : str_vect)
+        if(!g_strcmp0(id, str))
+            return true;
+    return false;
```
jralls: Prefer STL algorithms over raw loops, e.g. (untested, might need a tweak)
`std::find_if(str_vect.begin(), str_vect.end(), [str](id)->bool{ !g_strcmp0(id, str); }) != `str_vect.end()`.

libgnucash/engine/test/test-qofid.cpp
```diff
@@ -0,0 +1,205 @@
...
+{
+    results_ptr->push_back(gncJobGetID(job));
+}
+
+static gint cb_compare(GncJob* job1, GncJob* job2)
+{
+    return g_strcmp0(gncJobGetID(job1), gncJobGetID(job2));
+}
+
+static bool is_in(const char* str, const std::vector<const char*> str_vect)
```
jralls: `str_vect` should be a reference, no point in copying it.

libgnucash/engine/test/test-qofid.cpp
```diff
@@ -0,0 +1,205 @@
...
+        if(!g_strcmp0(id, str))
+            return true;
+    return false;
+}
+
+TEST(QofIDTest, collection_foreach)
+{
+    auto col = qof_collection_new(GNC_ID_JOB);
+    auto book = qof_book_new();
+    auto job_ids = std::vector<const char*>{"zzz", "ggg", "qqq", "aaa"};
```
jralls: That's a weird way to write the `job_ids` decl. The idiomatic way is 
```
std::vector<const char*> job_ids = {"zzz", "ggg", "qqq", "aaa"};
```
You can leave out the `=` since C++17.

jralls (changes requested):
Pretty good. My comments are mostly about style.

An overall comment about the coverage, though: Don't make a fetish of 100% coverage. Not everything needs to be tested: For example `qof_collection_mark_dirty` and `qof_collection_mark_clean` set and unset a bool and `qof_collection_is_dirty` returns the value of that bool. Testing one-liners like that isn't productive.

libgnucash/engine/test/test-qofid.cpp
stefan-koch-micro: Done.  (Copied from another test).

libgnucash/engine/test/test-qofid.cpp
stefan-koch-micro: Not exactly like that, but done.

libgnucash/engine/test/test-qofid.cpp
stefan-koch-micro: Duh.  Thank you.

libgnucash/engine/test/test-qofid.cpp
stefan-koch-micro: Where I'm from the convention is to do it like I had done.  That way it makes it really easy to pick out the variable names in a column of auto declarations.  But if gnucash  convention is different, then that is OK with me.  (Took me a while to like the auto convention.)  Changed.

stefan-koch-micro:
Thank you.  I saw the comments on github before I saw this.

I have fixed them all (and commented there).
I have also fixed the ninja distcheck where I forgot to include the new
test driver file in the source distribution.

I have changed the std vector declaration.

instead of
auto job_ids = std::vector<const char*>{"zzz", "ggg", "qqq", "aaa"};
I have:
std::vector<const char*> job_ids{"zzz", "ggg", "qqq", "aaa"};

Where I'm from the convention is to do it like I had done. That way it
makes it really easy to pick out the variable names in a column of auto
declarations. That style gorws on you.

stefan-koch-micro:
So, I’m still confused on process.  I have made the changes you requested,
and fixed the ci issues with distcheck.  I think I have even asked for a
re-review.  But, the pr is still marked as changes requested.

What should I do next to move this forward?

jralls:
> What should I do next to move this forward?

Be patient. We're all volunteers here who work on GnuCash when time is available.

stefan-koch-micro:
Sorry. Thanks.  I was more concerned that it seemed to be in a state that
was my next action.

jralls (approved):

jralls:
Thanks!

stefan-koch-micro:
No, thank you for all your help.

PR #2170, opened 2026-01-15
[trep-engine] add per-column totals in subtotal table

christopherlam:
adds column totals in subtotal table


christopherlam:
this branch needs a small upgrade to `gnc_commodity_compare` so that it's stable across runs, because `row->commodities` needs to output a stable list of commodities.

PR #2171, opened 2026-01-18
Bug 799722 - Inconsistent wording: Bill and Invoice

Bob-IT:
I do not know much about the Business side but this is an attempt to fix the inconsistent 'Jump to invoice' menu label.

The first commit is just some entries that I do not think are required, will adjust commit text before pushing.

The second commit is a fix based on static structures which does work but does not seem to be very flexible.

The final commit was a change to using a GSList of menu update structures which again works and I believe is more fixable catering for more than one unrelated menu update.

The last thing I thought of was possibly just changing the `Jump to invoice` to be less specific like `Jump to business document`.

I will adjust commits if this progresses.

### DIFF
--- a/gnucash/gnome/gnc-plugin-page-register.cpp
+++ b/gnucash/gnome/gnc-plugin-page-register.cpp
@@ -276,7 +276,7 @@ static GncInvoice* invoice_from_split (Split* split);
 #define LINK_TRANSACTION_OPEN_LABEL      N_("_Open Linked Document")
 /* Translators: This is a menu item that will open the bill, invoice, or voucher
    that is posted to the current transaction if there is one. */
-#define JUMP_LINKED_INVOICE_LABEL        N_("Jump to Invoice")
+#define JUMP_LINKED_INVOICE_LABEL        N_("Jump to Business item")
 #define CUT_SPLIT_LABEL                  N_("Cu_t Split")
 #define COPY_SPLIT_LABEL                 N_("_Copy Split")
 #define PASTE_SPLIT_LABEL                N_("_Paste Split")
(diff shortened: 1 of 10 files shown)


gnucash/gnome/gnc-plugin-page-register.cpp
```diff
@@ -1030,14 +1076,25 @@ gnc_plugin_page_register_ui_update (gpointer various,
...
+    bool is_credit_note = false;
+    GncInvoiceType invoice_type = GNC_INVOICE_UNDEFINED;
     if (trans)
     {
         auto invoices = invoices_from_transaction (trans);
+        if (invoices.size() == 1)
+        {
+            is_credit_note = gncInvoiceGetIsCreditNote (invoices[0]);
+            invoice_type = gncInvoiceGetType (invoices[0]);
+        }
```
christopherlam: you may have noticed the `invoices` is retrieved; a payment may cover multiple invoices or bills. if it's too difficult to modify the texts, it may be simpler to rename to "Jump to invoice, bill or expense"?

Bob-IT: I did see that,
Question will they all be of one kind, all invoices or all bills?

If not then I think is a non starter, easier to change the menu label to be more general and possibly the tool-tip.
I like, as you have used documents for the dialog that pops up with multiples..

Label,  'Jump to Business Documents' 
Tip, 'Jump to invoice, bill, expense or credit note.'

jralls: Document doesn't work for me, it implies something separate from GnuCash's data like a linked document. How about Item or Record?

Bob-IT: OK, I went looking in the en-GB PO file for the word 'document' and there are about 20 places where it is used in relation to Business like...
```
dialog-invoice.c:3902
                       ngettext("The following vendor document is due:",
                               "The following %d vendor documents are due:",

dialog-invoice.c:3914
                      ngettext("The following customer document is due:",
                               "The following %d customer documents are due:",

dialog-payment.c:293
        conflict_msg = _("No documents were selected to assign this payment to. This may create an unattached payment.");

dialog-payment.glade:335
                    <property name="label" translatable="yes">Documents</property>

gncOwner.c:1012
    gchar *memo_prefix = _("Offset between documents: ");

```
`Item` would also work, so you would have...
Label, '`Jump to Business Item`'
Tip, `'Jump to invoice, bill, expense or credit note.`'

Question, should all the other references be changed for consistency?

jralls: I think `customer item` or `business item` works pretty well in all those cases except the first. There's only one kind of vendor item so that one might as well say `bill`.

Yes, consistency is good. String freeze is just over a week away so we need to get this cleaned up and merged soon if it's going to make 5.15.

Bob-IT: I have dropped all the changes above  to now changing the wording to use 'Business Item'.
I looked again through the en_GB Po file to identify the changes in the following commit which hopefully takes care of the business related items.

I am not sure of the '`business-prefs.glade`' changes and the associated `gscheme.xml.in`, possibly could of been left as was but easy to drop before pushing.

jralls (approved):
The changes in org.gnucash.GnuCash.dialogs.business.gschema.xml.in make the schema documentation consistent with everything else, so that's good. There's no change to keys or values so there's no backwards compatibility problem. Merge when ready.

PR #2172, opened 2026-01-19
Cleanup and threadlocal

christopherlam:
using `thread_local` will, at least, remove the need for comments...


jralls:
Unfortunately just declaring variables `thread_local` makes them thread safer but not thread safe: It means that separate threads can call the function and use the returned variable without stepping on another thread doing the same. But if you try to pass the reference or pointer to a function in another thread bad things will happen.

It would be much better to convert the functions using static buffers to instead take a `buffer&` parameter and let the caller deal with the lifetime.

PR #2173, opened 2026-01-24
Bug - When importing csv-transactions, currency_format is not always properly applied.

royhansen99:
### The bug:
When importing csv-transactions, currency_format is not always properly applied.
If you start out by setting `currency_format` and `date_format etc, and only choose account as the last action before proceeding, the result of your import will not use the currency_format you chose.

The issue also occurs if you:
- First set account, then currency format.
- Set a random account, then set the correct account, then set currency format.

As seen in screenshots below, gnucash does not properly use the chosen currency format with "comma" as the decimal separator and "dot" as the thousand separator:

1-import-preview

2-import-match

Here is the sample csv-file I used:
[sample.txt](https://github.com/user-attachments/files/24836641/sample.txt)

### Fix:
Reparse/tokenize all import lines when changing date/currency formats. In addition, make sure this also happens when setting base_account. After this change it works.


jralls:
Thanks!

PR #2175, opened 2026-01-26
Improve usability of csv-transaction-import by adding informal text for date-format selection.

royhansen99:
**The date-format selection box displays all variations with a "-" separator, however this separator also matches "/" and ".".
By adding a short informal text to explain this behavior, it is much clearer to the user.**

_In 2015 I started using Gnucash for accounting in my small Norwegian consulting business.
I've regularly been using the csv-import to import transactions from my bank, which export csv-files with date formatted using dots "." as separators.
It was never clear to me that the dates displayed in the selection box would also match dots ".", so I have been manipulating my csv's to use "-" every year since 2015.
I was going to contribute a PR to add support for dots, when I realized gnucash already supports it._

Hopefully with this change it becomes immediately clear to everyone that the separator also matches dots and slashes.

A screenshot of the updated version:
Screenshot from 2026-01-26 12-15-43

### DIFF
--- a/gnucash/gtkbuilder/assistant-csv-trans-import.glade
+++ b/gnucash/gtkbuilder/assistant-csv-trans-import.glade
@@ -591,6 +591,8 @@ To know which lines belong to the same transaction, the importer will compare th
                             <property name="can-focus">False</property>
                             <property name="halign">end</property>
                             <property name="label" translatable="yes">Date Format</property>
+                            <property name="tooltip-text" translatable="yes">Choose the order of day, month and year, to match the formatting in the imported file.
+Choices are displayed with a Hyphen (-) separator, but will also match Dot (.) and Slash (/) as separator.</property>
                           </object>
                           <packing>
                             <property name="left-attach">0</property>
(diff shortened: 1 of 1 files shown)


jralls:
This should be a tooltip on the listbutton, not text added to the body of the assistant page. Also the option is about what fields are present in what order so that the importer can disambiguate a string like "12/11/10" so it might be better for the tooltip to say that instead of focusing on the separators.

Separately, if you must use glade please load and save the file once and commit those changes, then add the tooltip as a separate commit so that the noise from Glade gratuitously reordering the elements is separate from the material change.

royhansen99:
> This should be a tooltip on the listbutton, not text added to the body of the assistant page. Also the option is about what fields are present in what order so that the importer can disambiguate a string like "12/11/10" so it might be better for the tooltip to say that instead of focusing on the separators.
> 
> Separately, if you must use glade please load and save the file once and commit those changes, then add the tooltip as a separate commit so that the noise from Glade gratuitously reordering the elements is separate from the material change.

Thanks for the review!

I altered the text a little bit and added two tooltips instead of the inlined text, one for the label and one for the selection-box.

jralls:
Perfect, thanks!

PR #2176, opened 2026-01-28
Fix 798873

stefan-koch-micro:


jralls:
The GLIST stuff should be a separate commit from the bug fix, but as I said on [gnucash-devel](https://lists.gnucash.org/pipermail/gnucash-devel/2026-January/047214.html) you discovered how the GLIST form gets created and that means we shouldn't delete it until we have good tests for capgains.c that exercise qof_instance_kvp_add/has/merge/remove_guid to make sure we don't break anything.

jralls:
Thanks!

PR #2178, opened 2026-02-07
Some fixes for schedule transactions

Bob-IT:
The first commit was a patch I started a while back which I found and is an enhancement for identifying the a scheduled transaction which does not balance, displays the difference.

While working on that I noticed that only the first transaction was ever tested so fixed that.

Also the duplicate/paste of scheduled transaction only copied the description/notes/memo fields so fixed that.

The last commit only enables the 'Paste Transaction/Split' menu entry only when there is something to paste.

### DIFF
--- a/gnucash/gnome/dialog-sx-editor.c
+++ b/gnucash/gnome/dialog-sx-editor.c
@@ -101,9 +101,9 @@ typedef enum _EndTypeEnum
 
 struct _GncSxEditorDialog
 {
-    GtkWidget *dialog;
-    GtkBuilder *builder;
-    GtkNotebook *notebook;
+    GtkWidget    *dialog;
+    GtkBuilder   *builder;
+    GtkNotebook  *notebook;
     SchedXaction *sx;
     /* If this is a new scheduled transaction or not. */
     int newsxP;
@@ -111,34 +111,33 @@ struct _GncSxEditorDialog
     /* The various widgets in the dialog */
     GNCLedgerDisplay *ledger;
 
-    GncFrequency *gncfreq;
+    GncFrequency     *gncfreq;
     GncDenseCalStore *dense_cal_model;
-    GncDenseCal *example_cal;
+    GncDenseCal      *example_cal;
 
-    GtkEntry *nameEntry;
-
-    GtkLabel *lastOccurLabel;
+    GtkEntry        *nameEntry;
+    GtkLabel        *lastOccurLabel;
 
     GtkToggleButton *enabledOpt;
     GtkToggleButton *autocreateOpt;
     GtkToggleButton *notifyOpt;
     GtkToggleButton *advanceOpt;
-    GtkSpinButton *advanceSpin;
+    GtkSpinButton   *advanceSpin;
     GtkToggleButton *remindOpt;
-    GtkSpinButton *remindSpin;
+    GtkSpinButton   *remindSpin;
 
     GtkToggleButton *optEndDate;
     GtkToggleButton *optEndNone;
     GtkToggleButton *optEndCount;
-    EndType end_type;
-    GtkEntry *endCountSpin;
-    GtkEntry *endRemainSpin;
-    GNCDateEdit *endDateEntry;
+    EndType          end_type;
+    GtkEntry        *endCountSpin;
+    GtkEntry        *endRemainSpin;
+    GNCDateEdit     *endDateEntry;
 
-    char *sxGUIDstr;
+    char            *sxGUIDstr;
 
     GncEmbeddedWindow *embed_window;
-    GncPluginPage *plugin_page;
+    GncPluginPage     *plugin_page;
 };
 
 /** Prototypes **********************************************************/
@@ -436,8 +435,8 @@ gnc_sxed_check_changed (GncSxEditorDialog *sxed)
  ****************************************************************************/
 typedef struct _txnCreditDebitSums
 {
-    gnc_numeric creditSum;
-    gnc_numeric debitSum;
+    gnc_numeric    creditSum;
+    gnc_numeric    debitSum;
 } txnCreditDebitSums;
 
 static txnCreditDebitSums *
--- a/gnucash/gnome/dialog-sx-editor.c
+++ b/gnucash/gnome/dialog-sx-editor.c
@@ -792,7 +792,7 @@ check_transaction_splits (Transaction *txn, gpointer data)
             return FALSE;
         }
     }
-    return TRUE;
+    return FALSE; // return FALSE to continue to next transaction
 }
 
 /*******************************************************************************
--- a/gnucash/gnome/dialog-sx-editor.c
+++ b/gnucash/gnome/dialog-sx-editor.c
@@ -437,6 +437,9 @@ typedef struct _txnCreditDebitSums
 {
     gnc_numeric    creditSum;
     gnc_numeric    debitSum;
+    gnc_commodity *base_cmdty;
+    GtkWindow     *window;
+    gboolean       multi_commodity;
 } txnCreditDebitSums;
 
 static txnCreditDebitSums *
@@ -444,6 +447,9 @@ tcds_new (void)
 {
     txnCreditDebitSums *tcds = g_new0 (txnCreditDebitSums, 1);
     tcds->creditSum = tcds->debitSum = gnc_numeric_zero ();
+    tcds->base_cmdty = NULL;
+    tcds->window = NULL;
+    tcds->multi_commodity = FALSE;
     return tcds;
 }
 
@@ -455,6 +461,8 @@ set_sums_to_zero (gpointer key,
     txnCreditDebitSums *tcds = (txnCreditDebitSums*)val;
     tcds->creditSum = gnc_numeric_zero ();
     tcds->debitSum  = gnc_numeric_zero ();
+    tcds->base_cmdty = NULL;
+    tcds->multi_commodity = FALSE;
 }
 
 inline static gnc_numeric
@@ -467,6 +475,7 @@ static void
 check_credit_debit_balance (gpointer key, gpointer val, gpointer ud)
 {
     txnCreditDebitSums *tcds = (txnCreditDebitSums*)val;
+    Transaction *txn = GNC_TRANSACTION(key);
     gboolean *unbalanced = (gboolean*)ud;
     gnc_numeric diff = tcds_difference (tcds);
     const char *result = gnc_numeric_zero_p (diff) ? "true" : "false";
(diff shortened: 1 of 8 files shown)


gnucash/gnome/dialog-sx-editor.c
```diff
@@ -467,6 +475,7 @@ static void
 check_credit_debit_balance (gpointer key, gpointer val, gpointer ud)
 {
     txnCreditDebitSums *tcds = (txnCreditDebitSums*)val;
+    Transaction *txn = (Transaction*)key;
```
jralls: Use `GNC_TRANSACTION(key)` on GObject-derived types to do the run-time verification.

gnucash/gnome/dialog-sx-editor.c
```diff
@@ -871,7 +890,7 @@ gnc_sxed_check_consistent (GncSxEditorDialog *sxed)
     {
         const char *msg =
             _("The Scheduled Transaction Editor cannot automatically "
-              "balance this transaction. Should it still be entered?");
+              "balance this schedule. Should it still be entered?");
```
jralls: Schedules don't balance. You're aiming for "cannot automatically balance all of the transactions in this Scheduled Transaction" but I guess you think that's a bit too wordy.

gnucash/register/ledger-core/split-register.c
```diff
@@ -105,6 +105,14 @@ clear_copied_item()
     copied_item.anchor_split_index = 0;
 }
 
+gboolean
+gnc_split_register_has_copied_item (void)
+{
+    if (!copied_item.ft && !copied_item.fs)
+        return FALSE;
+    return TRUE;
+}
```
jralls, line 112: `return copied_item.ft || copied_item.fs;` does the same and is simpler.

gnucash/gnome/dialog-sx-editor.c
```diff
@@ -436,15 +435,21 @@ gnc_sxed_check_changed (GncSxEditorDialog *sxed)
  ****************************************************************************/
 typedef struct _txnCreditDebitSums
 {
-    gnc_numeric creditSum;
-    gnc_numeric debitSum;
+    gnc_numeric    creditSum;
+    gnc_numeric    debitSum;
+    gnc_commodity *base_cmdty;
+    gpointer       window;
```
jralls: Why a `void*` instead of a `GtkWindow*`? It's never used generically (like as user data in a callback).

gnucash/register/ledger-core/split-register-copy-ops.c
```diff
@@ -138,11 +140,98 @@ void gnc_float_split_set_value (FloatingSplit *fs, const gnc_numeric value)
...
+
+    guid_free (ftsd->m_account_guid);
+    g_free (ftsd->m_credit_numeric);
+    g_free (ftsd->m_debit_numeric);
+
+    g_free (ftsd);
+}
+
+static
+FloatingTemplateSxData *gnc_split_to_float_template_sx_data (Split *split)
```
jralls: Return type on the first line so the function name starts the second, i.e.
```
static FloatingTemplateSxData*
gnc_split_to_float_template_sx_data (Split *split)
```

gnucash/register/ledger-core/split-register-copy-ops.c
```diff
@@ -168,21 +262,74 @@ FloatingSplit *gnc_split_to_float_split (Split *split)
    status and date are not copied. The split's guid is,
    of course, unchanged.
 */
-void gnc_float_split_to_split (const FloatingSplit *fs, Split *split)
+void gnc_float_split_to_split (const FloatingSplit *fs, Split *split, Account *template_account)
```
jralls: ISTM you're trying to cram too much into a single function. The tip off is the complex set of top-level if statements. Each of those branches deserves a separate function, with `gnc_float_split_to_split` limited to dispatch.

gnucash/register/ledger-core/split-register-copy-ops.c
```diff
@@ -368,6 +519,72 @@ FloatingTxn *gnc_txn_to_float_txn (Transaction *txn, gboolean use_cut_semantics)
...
+        FloatingSplit *fs = iter->data;
+        if (!fs)
+            continue;
+        if (fs->m_template_sx_data)
+            ftsd_exists = TRUE;
+    }
+    return ftsd_exists;
+}
+
+void gnc_float_txn_to_template_txn (const FloatingTxn *ft, Transaction *txn, Account *template_account, gboolean do_commit)
```
jralls: Return type on a separate line.

gnucash/register/ledger-core/split-register-copy-ops.h
```diff
@@ -113,10 +127,17 @@ void gnc_float_txn_set_splits (FloatingTxn *ft, SplitList *splits);
 
 void gnc_float_txn_append_float_split (FloatingTxn *ft, FloatingSplit *fs);
 
-FloatingTxn *gnc_txn_to_float_txn (Transaction *txn, gboolean use_cut_semantics);
+gboolean gnc_float_txn_has_template (const FloatingTxn *ft);
+
+FloatingTxn *gnc_txn_to_float_txn (Transaction *txn, gboolean use_cut_semantics, gboolean is_template);
+
+void gnc_float_txn_to_template_txn (const FloatingTxn *ft, Transaction *txn,
+                                    Account *template_account, gboolean do_commit);
```
jralls, line 135: I'm not keen on this API with its nullable `template_account` and boolean arg to `gnc_txn_to_float_txn`, but I guess it's unavoidable without refactoring `split_register_duplicate_current`, a major undertaking.

gnucash/register/ledger-core/split-register-copy-ops.c
```diff
@@ -138,11 +140,98 @@ void gnc_float_split_set_value (FloatingSplit *fs, const gnc_numeric value)
...
+    const gchar *cf = "", *df = "";
+    gnc_numeric cn = gnc_numeric_zero(), dn = gnc_numeric_zero();
+
+    if (ftsd->m_account_guid)
+        qof_instance_set (QOF_INSTANCE (split), "sx-account", ftsd->m_account_guid, NULL);
+
+    if (ftsd->m_credit_formula)
+        cf = ftsd->m_credit_formula;
+
+    if (!gnc_numeric_zero_p (*ftsd->m_credit_numeric))
```
jralls: m_credit_numeric can be NULL, so this must be
```
if (ftsd->m_credit_numeric && !gnc_numeric_zero_p (*fstd->m_credit_numeric))
```

gnucash/register/ledger-core/split-register-copy-ops.c
```diff
@@ -168,21 +262,74 @@ FloatingSplit *gnc_split_to_float_split (Split *split)
...
+        if (!gnc_numeric_negative_p (fs->m_amount))
+            gnc_split_set_sx_settings (split, "", zero, string, abs_diff);
+
+        split_account = template_account;
+    }
+    else if (!template_account && fs->m_template_sx_data) // From Sched to Reg
+    {
+        FloatingTemplateSxData *ftsd = fs->m_template_sx_data;
+
+        if (!gnc_numeric_zero_p (*ftsd->m_credit_numeric))
```
jralls: `if (ftsd->m_credit_numeric && !gnc_numeric_zero_p (*ftsd->m_credit_numeric))`

gnucash/register/ledger-core/split-register-copy-ops.c
```diff
@@ -168,21 +262,74 @@ FloatingSplit *gnc_split_to_float_split (Split *split)
...
+            xaccSplitSetAmount (split, gnc_numeric_neg(*ftsd->m_credit_numeric));
+            xaccSplitSetValue (split, gnc_numeric_neg(*ftsd->m_credit_numeric));
+        }
+
+        if (!gnc_numeric_zero_p (*ftsd->m_debit_numeric))
+        {
+            xaccSplitSetAmount (split, *ftsd->m_debit_numeric);
+            xaccSplitSetValue (split, *ftsd->m_debit_numeric);
+        }
+
```
jralls: If there's a variable in the formula then the numeric isn't set because it can't be resolved until the SLR gets the variable value. In those cases pasting a template txn into a real one will have 0-value splits and that will confuse less cluefull users. Should we pop a dialog warning about this?

jralls (changes requested):

gnucash/gnome/dialog-sx-editor.c
Bob-IT: Changed, not sure why I did it that way, maybe just copied line above.

gnucash/gnome/dialog-sx-editor.c
Bob-IT: Not really, changed.

gnucash/gnome/dialog-sx-editor.c
Bob-IT: I have no idea why I did not use GtkWindow, maybe I was going in a different direction when I started way back and did not notice, changed.

gnucash/register/ledger-core/split-register.c
Bob-IT: Sometimes I just do not see it, changed.

gnucash/register/ledger-core/split-register-copy-ops.c
Bob-IT: Typo, changed.

gnucash/register/ledger-core/split-register-copy-ops.c
Bob-IT: Changed, I would normally do that. I think I didn't do it as there were a lot of functions in the file that did not so just made it the same.
I will note the file and realign the source file some other time.

gnucash/register/ledger-core/split-register-copy-ops.c
Bob-IT: Changed.

gnucash/register/ledger-core/split-register-copy-ops.c
Bob-IT: Changed.

gnucash/register/ledger-core/split-register-copy-ops.c
Bob-IT: Added a couple of functions to simplify.

gnucash/register/ledger-core/split-register-copy-ops.c
Bob-IT: I am not sure, I thought it may of made sense they would be empty with a formula/variable specified.
That would apply to transactions and splits?

jralls: It will make sense to the more advanced users who understand how SX variables work. The other 90% will be confused. On the other hand the other 90% probably won't have SXes with variables other than those created by the mortgage assistant, so maybe it doesn't matter. Let's go with it as-is and if anyone notices we can add the message box.

gnucash/register/ledger-core/split-register-copy-ops.c
```diff
@@ -159,30 +248,101 @@ FloatingSplit *gnc_split_to_float_split (Split *split)
...
-    xaccSplitSetValue (split, fs->m_value);
-    if (fs->m_account)
+
+    if (template_account && fs->m_template_sx_data) // From Sched to Sched
+    {
+        split_account = template_account;
+        gnc_float_template_sx_data_to_split (fs, split);
+    }
+
+    if (template_account && !fs->m_template_sx_data) // From Reg to Sched
```
jralls: Doesn't this need to be `else if` so that the final `else` doesn't overwrite the split amount and value set in `gnc_float_template_sx_data_to_split`?

Bob-IT: As I understand it, those values are not used in the scheduled transactions instead the values come from the kvp data but to make it clear I have added.

jralls: Right. Thanks, it does make it clearer.

jralls (approved):

PR #2179, opened 2026-02-09
Fix segfault in gnc_gsettings_get_settings_obj when GSettings schemas are not installed

Noerr:
## Summary

Fixes segfault in `gnc_gsettings_get_settings_obj()` when GSettings schemas are missing (e.g., library-only builds with `-DWITH_GNUCASH=OFF`).

## Problem

When `g_settings_schema_source_get_default()` returns NULL (no compiled schemas), the code was passing NULL directly to `g_settings_schema_source_lookup()`, causing a segmentation fault. The existing `G_IS_SETTINGS()` check came too late—after three potential NULL dereferences.

**Crash sequence:**
```
import gnucash (Python)
→ gnc_prefs_init()
→ gnc_gsettings_version_upgrade()
→ gnc_gsettings_get_settings_obj("general")
→ SEGFAULT at line 108
```

## Solution

Added NULL guards before dereferencing GLib function return values:
- Check `schema_source` before passing to `g_settings_schema_source_lookup()`
- Check `schema` before passing to `g_settings_new_full()`
- Added same guard in `gnc_settings_dump_schema_paths()` for consistency

All callers already handle NULL returns via `G_IS_SETTINGS()` checks.

## Testing Status

⚠️ **NEEDS TESTING** - Requires specific build configuration:
- Build with `-DWITH_GNUCASH=OFF` (library-only, due to bug, currently installs no schemas)
- Test Python import: `python3 -c "import gnucash"`
- Should complete without segfault (will log warnings but not crash)

Without this fix, the import immediately segfaults. With the fix, it should gracefully degrade.

## Related

- Bug report: https://bugs.gnucash.org/show_bug.cgi?id=799740
- Companion bug: https://bugs.gnucash.org/show_bug.cgi?id=799739

### DIFF
--- a/libgnucash/app-utils/gnc-gsettings.cpp
+++ b/libgnucash/app-utils/gnc-gsettings.cpp
@@ -105,20 +105,10 @@ static GSettings * gnc_gsettings_get_settings_obj (const gchar *schema_str)
     auto full_name_str = normalize_schema_name (schema_str);
     auto full_name = full_name_str.c_str();
     auto schema_source {g_settings_schema_source_get_default()};
-    if (!schema_source)
-    {
-        PWARN ("No GSettings schema source available; cannot access schema %s", full_name);
-        LEAVE("");
-        return nullptr;
-    }
+    g_return_val_if_fail(schema_source, nullptr);
 
     auto schema {g_settings_schema_source_lookup(schema_source, full_name, true)};
-    if (!schema)
-    {
-        PWARN ("GSettings schema %s not found", full_name);
-        LEAVE("");
-        return nullptr;
-    }
+    g_return_val_if_fail(schema, nullptr);
 
     auto gset = g_settings_new_full (schema, nullptr, nullptr);
     DEBUG ("Created gsettings object %p for schema %s", gset, full_name);
@@ -559,11 +549,7 @@ gnc_settings_dump_schema_paths (void)
     gchar **non_relocatable;
 
     auto schema_source {g_settings_schema_source_get_default()};
-    if (!schema_source)
-    {
-        PWARN ("No GSettings schema source available; cannot dump schema paths");
-        return;
-    }
+    g_return_if_fail(schema_source);
 
     g_settings_schema_source_list_schemas (schema_source, true,
                                            &non_relocatable, nullptr);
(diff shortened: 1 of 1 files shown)


libgnucash/app-utils/gnc-gsettings.cpp
```diff
@@ -105,7 +105,21 @@ static GSettings * gnc_gsettings_get_settings_obj (const gchar *schema_str)
...
     auto full_name = full_name_str.c_str();
     auto schema_source {g_settings_schema_source_get_default()};
+    if (!schema_source)
+    {
+        PWARN ("No GSettings schema source available; cannot access schema %s", full_name);
+        LEAVE("");
+        return nullptr;
+    }
+
     auto schema {g_settings_schema_source_lookup(schema_source, full_name, true)};
```
jralls: `g_return_val_if_fail(schema_source, nullptr);` is sufficient in all 3 cases. It logs an error so it's evident that the function is aborted, no need to call `LEAVE` too.

Noerr:
Thanks for the suggestion @jralls . Much cleaner.
I've had a chance to test and this works as expected when the schema templates are not present.  No SEGFAULT, instead lots of error messages like,

```
ERROR <gnc.app-utils> T gnc_gsettings_get(const char*, const char*,
      T (*)(GSettings*, const char*), T) [with T = int; GSettings = GSettings]: assertion
     'G_IS_SETTINGS (gs_obj)' failed
```

I built and tested with Docker.  Key here is that -DWITH_GNUCASH=OFF  prevents the schemas from installing ([bug 799739](https://bugs.gnucash.org/show_bug.cgi?id=799739) ) so that this PR fix is exercised.
[Dockerfile.test-segfault-fix.txt](https://github.com/user-attachments/files/25269801/Dockerfile.test-segfault-fix.txt)

jralls:
Thanks!

libgnucash/app-utils/gnc-gsettings.cpp
christopherlam: One cosmetic issue is `LEAVE` does call `qof_log_dedent()` which means if the `ENTER` isn't balanced by `LEAVE` the log indenting will keep increasing.

PR #2181, opened 2026-02-17
Bug 799734 - mapped OFX income account not listed

Bob-IT:
Change the Import Map Editor to look for the KVP slot "ofx/associated-income-account" so it can be listed. The account
reference stored is a GUID so add a function to retrieve that so the existing gnc_account_delete_map_entry function can be used to remove it.

Will probably squash these two commits before pushing.

### DIFF
--- a/gnucash/gnome/dialog-imap-editor.c
+++ b/gnucash/gnome/dialog-imap-editor.c
@@ -704,6 +704,97 @@ get_account_info_nbayes (ImapDialog *imap_dialog, GList *accts)
 }
 
 static void
+add_online_entry_to_store (ImapDialog *imap_dialog, GtkTreeIter *toplevel,
+                           const gchar *based_on, GncImapInfo *imapInfo)
+{
+    // Add top level entry and pass iter to add_to_store
+    gtk_tree_store_append (GTK_TREE_STORE(imap_dialog->model), toplevel, NULL);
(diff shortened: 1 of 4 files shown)


gnucash/gnome/dialog-imap-editor.c
```diff
@@ -703,6 +703,97 @@ get_account_info_nbayes (ImapDialog *imap_dialog, GList *accts)
...
+{
+    // Add top level entry and pass iter to add_to_store
+    gtk_tree_store_append (GTK_TREE_STORE(imap_dialog->model), toplevel, NULL);
+    add_to_store (imap_dialog, toplevel, based_on, imapInfo);
+
+    imapInfo->map_account = NULL;
+}
+
+static void
+check_on_line (ImapDialog *imap_dialog, GtkTreeIter *toplevel, Account *acc, GncImapInfo *imapInfo)
```
jralls: `check_online`, please. `check_on_line` implies something quite different.

jralls (changes requested):
1 nit, otherwise fine.

gnucash/gnome/dialog-imap-editor.c
Bob-IT: I think that was a typo, changed.

jralls:
@Bob-IT we're getting close to 5.15 release. Can you get this and your other two PRs merged soon?

jralls (approved):

PR #2182, opened 2026-02-20
Store log and backup files in logs subdirectory

jshrek:
When opening a .gnucash file, GnuCash creates:
- a .log file
- timestamped backup files (e.g. .YYYYMMDD.gnucash)

This change creates a "logs" subdirectory beside the main data file and writes the .log and timestamped backup files into that directory (creating it if needed).

The .LCK lock file remains in the same directory as the main .gnucash file (unchanged).

### DIFF
--- a/libgnucash/backend/xml/gnc-xml-backend.cpp
+++ b/libgnucash/backend/xml/gnc-xml-backend.cpp
@@ -145,7 +145,21 @@ GncXmlBackend::session_begin(QofSession* session, const char* new_uri,
     /* ---------------------------------------------------- */
     /* We should now have a fully resolved path name.
      * Let's start logging */
-    xaccLogSetBaseName (m_fullpath.c_str());
+    /* Build logs directory inside main file directory */
+    auto logs_dir = g_build_filename(m_dirname.c_str(), "logs", NULL);
+    g_mkdir_with_parents(logs_dir, 0700);
+    
+    /* Build new base path for logs */
+    auto filename = g_path_get_basename(m_fullpath.c_str());
+    auto new_base = g_build_filename(logs_dir, filename, NULL);
+    
+    /* Set log base name */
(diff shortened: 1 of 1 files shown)


jshrek:
This patch meant to address these:

https://gnucash.uservoice.com/forums/101223-feature-request/suggestions/1470507-configuration-option-for-backup-location

https://bugs.gnucash.org/show_bug.cgi?id=619119

jralls (changes requested):
This behavior change must be optional.

This would fix [Bug 619119](https://bugs.gnucash.org/show_bug.cgi?id=619119), but the two bugs marked duplicates of that one request that the user be able to specify the log directory. Although not articulated, some might want to specify the backup directory separately. Regardless, the commit message needs to conform to the [bug-fix format](https://wiki.gnucash.org/wiki/Git#Link_Bugzilla_Entries)

Although the DBI/SQL backend doesn't make backup files it does write transaction logs so the log basename change must apply there as well (it's in libgnucash/backend/dbi/gnc-backend-dbi.cpp).

These are C++ files, so please use the C++ string and [boost::filesystem](https://www.boost.org/doc/libs/latest/libs/filesystem/doc/reference.html) functions instead of GLib ones. Note that there are already some standard file-path facilities in `libgnucash/core-utils/gnc-filepath-utils`. It would be reasonable to add the log and backup directory functions there.

Please don't litter your code with comments that state the obvious.

jralls:
@jshrek are you going to address my comments?

jshrek:
@jralls Thanks for comments, I did see them. It took me quite a while to get as far as I did, so it may take me another "quite a while" before I can fix this properly.

jralls:
@jshrek OK. I've marked it as draft so that I can stop looking at it every time I'm reviewing PRs. Just click the Ready for review button when you've gotten it fixed.

PR #2185, opened 2026-02-26
Correct matching against description and memo

jpschewe:
When the description or the memo is of length zero or one, doing a fuzzy match causes the the check to be against zero characters and this always comes back true. Because of this, skip the match when the string is too short.

Also enhance the debugging to make it easier to track why some transactions aren't considered possible matches.

This is a fix for https://bugs.gnucash.org/show_bug.cgi?id=799745

### DIFF
--- a/gnucash/import-export/import-backend.cpp
+++ b/gnucash/import-export/import-backend.cpp
@@ -667,7 +667,9 @@ void split_find_match (GNCImportTransInfo * trans_info,
 
     /* Check number heuristics */
     auto new_trans_str = gnc_get_num_action(new_trans, new_trans_fsplit);
-    if (new_trans_str && *new_trans_str)
+    auto split_str = gnc_get_num_action (xaccSplitGetParent (split), split);
+    DEBUG("number download: '%s' to match: '%s'", new_trans_str, split_str);
+    if (new_trans_str && *new_trans_str && split_str && *split_str)
     {
         char *endptr;
         auto conversion_ok = true;
@@ -679,7 +681,6 @@ void split_find_match (GNCImportTransInfo * trans_info,
                         numbers on string and string empty */
         conversion_ok = !(errno || endptr == new_trans_str);
 
-        auto split_str = gnc_get_num_action (xaccSplitGetParent (split), split);
         errno = 0;
         auto split_number = strtol(split_str, &endptr, 10);
         conversion_ok =  !(errno || endptr == split_str);
@@ -701,9 +702,11 @@ void split_find_match (GNCImportTransInfo * trans_info,
 
     /* Memo heuristics */
     auto memo = xaccSplitGetMemo(new_trans_fsplit);
-    if (memo && *memo)
+    auto match_memo = xaccSplitGetMemo(split);
+    if (memo && *memo && match_memo && *match_memo)
     {
-        if (safe_strcasecmp(memo, xaccSplitGetMemo(split)) == 0)
+        DEBUG("memo download: '%s' to match: '%s'", memo, match_memo);
+        if (safe_strcasecmp(memo, match_memo) == 0)
         {
             /* An exact match of memo gives a +2 */
             prob = prob + 2;
@@ -723,10 +726,11 @@ void split_find_match (GNCImportTransInfo * trans_info,
 
     /* Description heuristics */
     auto descr = xaccTransGetDescription(new_trans);
-    if (descr && *descr)
+    auto match_descr = xaccTransGetDescription(xaccSplitGetParent(split));
+    if (descr && *descr && match_descr && *match_descr)
     {
-        if (safe_strcasecmp(descr,
-                xaccTransGetDescription(xaccSplitGetParent(split))) == 0)
+        DEBUG("description: download: '%s' to match: '%s'", descr, match_descr);
+        if (safe_strcasecmp(descr, match_descr) == 0)
         {
             /*An exact match of Description gives a +2 */
             prob = prob + 2;
--- a/gnucash/import-export/import-backend.cpp
+++ b/gnucash/import-export/import-backend.cpp
@@ -593,9 +593,9 @@ void split_find_match (GNCImportTransInfo * trans_info,
     /* Amount heuristics */
     auto downloaded_split_amount =
         gnc_numeric_to_double (xaccSplitGetAmount(new_trans_fsplit));
-    /*DEBUG(" downloaded_split_amount=%f", downloaded_split_amount);*/
+    DEBUG(" downloaded_split_amount=%f", downloaded_split_amount);
     auto match_split_amount = gnc_numeric_to_double(xaccSplitGetAmount(split));
-    /*DEBUG(" match_split_amount=%f", match_split_amount);*/
+    DEBUG(" match_split_amount=%f", match_split_amount);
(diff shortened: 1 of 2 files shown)


gnucash/import-export/import-backend.cpp
```diff
@@ -696,53 +698,57 @@ void split_find_match (GNCImportTransInfo * trans_info,
...
         {
             /* An exact match of memo gives a +2 */
             prob = prob + 2;
-            /* DEBUG("heuristics:  probability + 2 (memo)"); */
+            DEBUG("heuristics:  probability + 2 (memo)");
         }
-        else if ((strncasecmp(memo, xaccSplitGetMemo(split),
-                    strlen(xaccSplitGetMemo(split)) / 2) == 0))
+        else if (strlen(match_memo) > 1 && (strncasecmp(memo, match_memo,
+                    strlen(match_memo) / 2) == 0))
```
jralls: `strlen` and `strncasecmp` need to be null-protected. I'd also go for a bit more than a more restrictive match if one of the strings is substantially longer than the other, maybe test 1/2 the length of whichever is longer.

jralls (changes requested):
Please separate uncommenting the DEBUG statements and fixing the bug into separate commits.

And since you created a companion bug (no need to do that), please title the bug fix commit with the bug number and title$mdash;"[Bug 799745] - Import matcher doesn't handle zero length memo and description properly" so that [the release note script](https://github.com/Gnucash/gnucash/blob/stable/util/git-release-notes.pl) categorizes it correctly.

gnucash/import-export/import-backend.cpp
jpschewe: I was trying for the minimal set of changes to fix the issue. The change to the length of the match would be a larger change in behavior. I can make the larger change if you want though.

As far as the companion bug, I was just following the developer guidelines.

jpschewe: Changes are in place.

gnucash/import-export/import-main-matcher.cpp
```diff
@@ -2268,7 +2268,7 @@ gnc_gen_trans_list_add_trans_internal (GNCImportMainMatcher *gui, Transaction *t
     {
         /* If it does, abort the process for this transaction, since
            it is already in the system. */
-        DEBUG("%s", "Transaction with same online ID exists, destroying current transaction");
+        DEBUG("Transaction with same online ID exists, destroying current transaction: %s", qof_print_date(xaccTransGetDate(trans)));
```
christopherlam: This is a useful change, but will leak `qof_print_date`. Options: g_free later

```cpp
char* date = qof_print_date(...);
DEBUG("...%s",date);
g_free(date);
```

Alternatively just show the pointer ```DEBUG("...%p", trans);```.

Or even better, also debug-log the online_id from within `gnc_import_exists_online_id`:

```patch
modified   gnucash/import-export/import-backend.cpp
@@ -1052,6 +1052,10 @@ gboolean gnc_import_exists_online_id (Transaction *trans, GHashTable* acct_id_ha
     }
 
     auto online_id_exists = g_hash_table_contains (online_id_hash, source_online_id);
+
+    if (online_id_exists)
+        DEBUG ("transaction with online id %s already exists", source_online_id);
+
     g_free (source_online_id);
     return online_id_exists;
 }
```

jpschewe: THank you, I had forgotten about freeing the pointers. 2 questions:
1) any thoughts on using shared pointers to avoid issues with free?
2) is it all right to have the fix as a separate commit or do you want me to rebase the changes into the previous commits?

jralls: std::shared_ptr carries a fair amount of overhead for instance-counting so you only want to use it if you're actually sharing ownership of the pointer. std::unique_ptr has less overhead and is intended for automatic freeing of pointers where ownership isn't shared, so that's what you'd use here. *But* both of them use operator `delete` by default so if you use them on pointers that aren't allocated with operator `new` you must provide a custom deleter; that can be a simple lambda that calls `free`. `g_free` is just a null-protected wrapper around `free`. 
I don't think that using `std::unique_ptr` here would get you anything. You can't be sure when using it inline like
```
DEBUG("Transaction with same online ID exists, destroying current transaction: %s", (std::unique_ptr<char*, (void(*)(char*)>(qof_print_date(xaccTransGetDate(trans)[](char* p){ g_free(p); })));
```
that the temporary will survive long enough for the `printf` buried in the `DEBUG` macro to use it before it's freed so you would still need to create a local variable. Plus you're creating a lot of extra work for yourself and the compiler. The problem for most of us is noticing that we need to free the rv from a function like `qof_print_date` and therefore can't use it as a temporary.

You should fix the first commit. You can force-push your branch and the PR will update automatically.

gnucash/import-export/import-backend.cpp
jralls: > As far as the companion bug, I was just following the developer guidelines.

Where did you find that?

jpschewe: I was reading https://wiki.gnucash.org/wiki/Git#Pull_Requests and saw the requirement for bug reports under patches and assumed it was for all changes. I've had other projects refuse PRs for not opening a bug when I see such language.

gnucash/import-export/import-main-matcher.cpp
jpschewe: Thank you for the explanation about shared and unique pointers. I don't see why there would be any issue with the compiler executing the free before the printf gets the pointer. I expect that would be a bug in the compiler.

That all said, I'm using g_free in my updated commits.

gnucash/import-export/import-backend.cpp
jpschewe: > `strlen` and `strncasecmp` need to be null-protected. I'd also go for a bit more than a more restrictive match if one of the strings is substantially longer than the other, maybe test 1/2 the length of whichever is longer.

@jralls  I've updated the PR do use this algorithm. Let me know what you think.

gnucash/import-export/import-main-matcher.cpp
jpschewe: > This is a useful change, but will leak `qof_print_date`. Options: g_free later
> 
@christopherlam  I've updated the PR with the freeing of the memory. I've also added the debug statement that you suggested in `gnc_import_exists_online_id`.

jralls: > I don't see why there would be any issue with the compiler executing the free before the printf gets the pointer. I expect that would be a bug in the compiler.

I got distracted by the deleter lambda and forgot to `get()` the `char*` from the `unique_ptr` in my example. Here's the corrected one:
```
DEBUG("Transaction with same online ID exists, destroying current transaction: %s", (std::unique_ptr<char*, (void(*)(char*)>(qof_print_date(xaccTransGetDate(trans)[](char* p){ g_free(p); }))))->get());
```
[It's complicated](https://en.cppreference.com/w/cpp/language/lifetime.html) and I think implementation dependent because the actual temporary is the `char*` from the `get`, though that might qualify as creating a reference that extends the lifetime of the `unique_ptr`. As it happens both gcc and clang call the deleter after the printf, see [compiler explorer](https://godbolt.org/z/Gc3x61d1j).

jralls:
Thanks!

PR #2188, opened 2026-03-15
Fix completion cell gui leak

agwekixj:
The `gui_realize` test in `gnc_completion_cell_gui_destroy() `is corrected.  The `item_store` allocation is moved to `gnc_completion_cell_gui_realize()` and the `item_store` is released in `gnc_completion_cell_gui_destroy()`.  The patch also modifies `item_store_clear()` and `populate_list_store()` for clarity.


jralls:
Thanks!

PR #2189, opened 2026-03-16
Avoid ub in fast_iso8601_utc_parse

christopherlam:
if `const char*` argument was 3 bytes or less.
also some comment and typo fixes.

### DIFF
--- a/libgnucash/engine/gnc-datetime.cpp
+++ b/libgnucash/engine/gnc-datetime.cpp
@@ -381,13 +381,12 @@ static std::optional<PTime>
 fast_iso8601_utc_parse (const char* str)
 {
     int32_t year, month, mday, hour, min, sec;
-
-    // parse the first 4 bytes into year
-    if (!str || !parse_chars_into_num (str, str + 4, year))
-        return {};
+    const size_t len = str ? strnlen (str, 26) : 0;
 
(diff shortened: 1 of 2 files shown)


libgnucash/engine/gnc-datetime.cpp
```diff
@@ -381,13 +381,14 @@ static std::optional<PTime>
 fast_iso8601_utc_parse (const char* str)
 {
     int32_t year, month, mday, hour, min, sec;
+    const size_t len = str ? strnlen (str, 26) : 0;
 
     // parse the first 4 bytes into year
-    if (!str || !parse_chars_into_num (str, str + 4, year))
+    if (len < 14 || !parse_chars_into_num (str, str + 4, year))
```
jralls: Magic numbers, so 
```
constexpr int compact_iso_string_len = 14;
constexpr int expanded_iso_string_len = 25;

if (len < compact_iso_string_len ...)
```
But maybe better `if (len != compact_iso_string_len || len != expanded_iso_string_len || !parse_chars_into_num (str, str + 4, year))`. Then you wouldn't need any more length checks

jralls (changes requested):

jralls (approved):

PR #2190, opened 2026-03-20
Fix translation of Post invoice and Duplicate invoice

unhammer:
"Oppføring" means "Item" (which could be a synonym of "Post", but not in this context).

"Duplisert" is "DuplicateD", we want the imperative.


jralls:
Please use https://hosted.weblate.org/projects/gnucash/gnucash/ for translations.

PR #2193, opened 2026-03-20
Fix plugin page report gui leak

agwekixj:
Added `g_object_unref()` of `priv->container` to `gnc_plugin_page_report_destroy()`.


jralls:
Thanks!

PR #2195, opened 2026-03-23
fix plugin page register gui leak

agwekixj:
Added `g_object_ref()` of `gsr` to `gnc_plugin_page_register_create_widget()` and `g_object_unref()` of `priv->widget` to `gnc_plugin_page_register_destroy_widget()`.  Also, removed the `gtk_widget_destroy()` of `priv->gsr `as `priv->gsr` will be handled by the `g_object_unref ()` of `priv->widget`.


jralls:
That works, thanks.

PR #2196, opened 2026-03-23
Converted the gnc-html target to C++

russ-gorby:
Converted the gnc-html target to C++
	  - no longer minimal conversion
	  - has some C++ constructs: nullptr, "using" directives, some STL containers
tested the Build and unittests on Windows 11 and Ubuntu (Jammy Jellyfish)
unittests configured with -DWITH_PYTHON=ON
	  - passes on Ubuntu
	  - only the python-bindings test fails on Windows (multiple pre-existing issues)

BTW, This is my first GNC PR

### DIFF
--- a/gnucash/html/gnc-html-factory.hpp
+++ b/gnucash/html/gnc-html-factory.hpp
@@ -1,5 +1,5 @@
 /********************************************************************
- * gnc-html-webkit.h -- display html with gnc special tags          *
+ * gnc-html-factory.hpp -- display html with gnc special tags         *
  * Copyright (C) 2009 Phil Longstaff <plongstaff@rogers.com>        *
  *                                                                  *
  * This program is free software; you can redistribute it and/or    *
@@ -20,45 +20,11 @@
  * Boston, MA  02110-1301,  USA       gnu@gnu.org                   *
 \********************************************************************/
 
-#ifndef GNC_HTML_WEBKIT_H
-#define GNC_HTML_WEBKIT_H
+#ifndef GNC_HTML_FACTORY_H
+#define GNC_HTML_FACTORY_H
 
-#include <glib-object.h>
 #include "gnc-html.h"
 
-G_BEGIN_DECLS
-
-#define GNC_TYPE_HTML_WEBKIT       (gnc_html_webkit_get_type())
-#define GNC_HTML_WEBKIT(o)         (G_TYPE_CHECK_INSTANCE_CAST ((o), GNC_TYPE_HTML_WEBKIT, GncHtmlWebkit))
-#define GNC_HTML_WEBKIT_CLASS(k)   (G_TYPE_CHECK_CLASS_CAST((k), GNC_TYPE_HTML_WEBKIT, GncHtmlWebkitClass))
-#define GNC_IS_HTML_WEBKIT(o)      (G_TYPE_CHECK_INSTANCE_TYPE((o), GNC_TYPE_HTML_WEBKIT))
-#define GNC_IS_HTML_WEBKIT_CLASS(k)   (G_TYPE_CHECK_CLASS_TYPE((k), GNC_TYPE_HTML_WEBKIT))
-#define GNC_HTML_WEBKIT_GET_CLASS(o)  (G_TYPE_INSTANCE_GET_CLASS((o), GNC_TYPE_HTML_WEBKIT, GncHtmlWebkitClass))
-
-typedef struct _GncHtmlWebkit GncHtmlWebkit;
-typedef struct _GncHtmlWebkitClass GncHtmlWebkitClass;
-typedef struct _GncHtmlWebkitPrivate GncHtmlWebkitPrivate;
-
-/** Key for saving the PDF-export directory in the print settings */
-#define GNC_GTK_PRINT_SETTINGS_EXPORT_DIR "gnc-pdf-export-directory"
-
-struct _GncHtmlWebkit
-{
-    GncHtml parent_instance;
-
-    /*< private >*/
-    GncHtmlWebkitPrivate* priv;
-};
-
-struct _GncHtmlWebkitClass
-{
-    GncHtmlClass parent_class;
-};
-
-GType gnc_html_webkit_get_type( void );
-
-GncHtml* gnc_html_webkit_new( void );
-
-G_END_DECLS
+extern "C" GncHtml* gnc_html_factory_create_html( void ) NOEXCEPT;
 
 #endif
--- a/gnucash/html/gnc-html-history.cpp
+++ b/gnucash/html/gnc-html-history.cpp
@@ -23,7 +23,7 @@
 #include <config.h>
 
 #include <gtk/gtk.h>
-#include <string.h>
+#include <string>
 
 #include "gnc-html-history.h"
 
@@ -43,12 +43,12 @@ struct _gnc_html_history
  ********************************************************************/
 
 gnc_html_history *
-gnc_html_history_new(void)
+gnc_html_history_new(void) noexcept
 {
     gnc_html_history * hist = g_new0(gnc_html_history, 1);
-    hist->nodes         = NULL;
-    hist->current_node  = NULL;
-    hist->last_node     = NULL;
+    hist->nodes         = nullptr;
+    hist->current_node  = nullptr;
+    hist->last_node     = nullptr;
     return hist;
 }
 
@@ -59,11 +59,9 @@ gnc_html_history_new(void)
  ********************************************************************/
 
 void
-gnc_html_history_destroy(gnc_html_history * hist)
+gnc_html_history_destroy(gnc_html_history * hist) noexcept
 {
-    GList * n;
-
-    for (n = hist->nodes; n ; n = n->next)
+    for (GList *n = hist->nodes; n ; n = n->next)
     {
         if (hist->destroy_cb)
         {
@@ -74,9 +72,9 @@ gnc_html_history_destroy(gnc_html_history * hist)
     }
     g_list_free(hist->nodes);
 
-    hist->nodes         = NULL;
-    hist->current_node  = NULL;
-    hist->last_node     = NULL;
+    hist->nodes         = nullptr;
+    hist->current_node  = nullptr;
+    hist->last_node     = nullptr;
     g_free(hist);
 }
 
@@ -87,7 +85,7 @@ gnc_html_history_destroy(gnc_html_history * hist)
 void
 gnc_html_history_set_node_destroy_cb(gnc_html_history * hist,
                                      gnc_html_history_destroy_cb cb,
-                                     gpointer cb_data)
+                                     gpointer cb_data) noexcept
 {
     hist->destroy_cb = cb;
     hist->destroy_cb_data = cb_data;
@@ -121,14 +119,11 @@ g_strcmp(char * a, char * b)
  ********************************************************************/
 void
 gnc_html_history_append(gnc_html_history * hist,
-                        gnc_html_history_node * node)
+                        gnc_html_history_node * node) noexcept
 {
-    GList * n;
-    gnc_html_history_node * hn;
-
     if (hist->current_node)
     {
-        hn = hist->current_node->data;
+        auto hn = static_cast<gnc_html_history_node *>(hist->current_node->data);
         if ((hn->type == node->type) &&
                 !g_strcmp(hn->location, node->location) &&
                 !g_strcmp(hn->label, node->label))
@@ -142,7 +137,7 @@ gnc_html_history_append(gnc_html_history * hist,
         }
 
         /* blow away the history after this point, if there is one */
-        for (n = hist->current_node->next; n; n = n->next)
+        for (GList *n = hist->current_node->next; n; n = n->next)
         {
             if (hist->destroy_cb)
             {
@@ -152,14 +147,14 @@ gnc_html_history_append(gnc_html_history * hist,
             gnc_html_history_node_destroy((gnc_html_history_node *)n->data);
         }
         g_list_free(hist->current_node->next);
-        hist->current_node->next = NULL;
+        hist->current_node->next = nullptr;
         hist->last_node = hist->current_node;
     }
 
-    n = g_list_alloc();
+    GList *n = g_list_alloc();
     n->data = (gpointer) node;
-    n->next = NULL;
-    n->prev = NULL;
+    n->next = nullptr;
+    n->prev = nullptr;
 
     if (hist->nodes && hist->last_node)
     {
@@ -187,11 +182,11 @@ gnc_html_history_append(gnc_html_history * hist,
  ********************************************************************/
 
 gnc_html_history_node *
-gnc_html_history_get_current(gnc_html_history * hist)
+gnc_html_history_get_current(gnc_html_history * hist) noexcept
 {
-    if (!hist || !(hist->current_node)) return NULL;
+    if (!hist || !(hist->current_node)) return nullptr;
 
-    return hist->current_node->data;
+    return static_cast<gnc_html_history_node *>(hist->current_node->data);
 }
 
 
@@ -200,11 +195,11 @@ gnc_html_history_get_current(gnc_html_history * hist)
  ********************************************************************/
 
 gnc_html_history_node *
-gnc_html_history_forward(gnc_html_history * hist)
+gnc_html_history_forward(gnc_html_history * hist) noexcept
 {
     if (!hist || !(hist->current_node))
     {
-        return NULL;
+        return nullptr;
     }
 
     if (hist->current_node->next)
@@ -212,7 +207,7 @@ gnc_html_history_forward(gnc_html_history * hist)
         hist->current_node = hist->current_node->next;
     }
 
-    return hist->current_node->data;
+    return static_cast<gnc_html_history_node *>(hist->current_node->data);
 }
 
 
@@ -221,12 +216,12 @@ gnc_html_history_forward(gnc_html_history * hist)
  ********************************************************************/
 
 gnc_html_history_node *
-gnc_html_history_back(gnc_html_history * hist)
+gnc_html_history_back(gnc_html_history * hist) noexcept
 {
 
     if (!hist || !(hist->current_node))
     {
-        return NULL;
+        return nullptr;
     }
 
     if (hist->current_node->prev)
@@ -234,7 +229,7 @@ gnc_html_history_back(gnc_html_history * hist)
         hist->current_node = hist->current_node->prev;
     }
 
-    return hist->current_node->data;
+    return static_cast<gnc_html_history_node *>(hist->current_node->data);
 }
 
 
@@ -244,7 +239,7 @@ gnc_html_history_back(gnc_html_history * hist)
  ********************************************************************/
 
 int
-gnc_html_history_back_p(gnc_html_history * hist)
+gnc_html_history_back_p(gnc_html_history * hist) noexcept
 {
     if (hist && hist->current_node && hist->current_node->prev)
     {
@@ -263,7 +258,7 @@ gnc_html_history_back_p(gnc_html_history * hist)
  ********************************************************************/
 
 int
-gnc_html_history_forward_p(gnc_html_history * hist)
+gnc_html_history_forward_p(gnc_html_history * hist) noexcept
 {
     if (hist && hist->current_node && hist->current_node->next)
     {
@@ -282,7 +277,7 @@ gnc_html_history_forward_p(gnc_html_history * hist)
 
 gnc_html_history_node *
 gnc_html_history_node_new(URLType type, const gchar * location,
-                          const gchar * label)
+                          const gchar * label) noexcept
 {
     gnc_html_history_node * rv = g_new0(gnc_html_history_node, 1);
 
@@ -298,7 +293,7 @@ gnc_html_history_node_new(URLType type, const gchar * location,
  ********************************************************************/
 
 void
-gnc_html_history_node_destroy(gnc_html_history_node * node)
+gnc_html_history_node_destroy(gnc_html_history_node * node) noexcept
 {
 
     /* free the url resources and cached text */
@@ -306,8 +301,8 @@ gnc_html_history_node_destroy(gnc_html_history_node * node)
     g_free(node->location);
     g_free(node->label);
 
-    node->location = NULL;
-    node->label    = NULL;
+    node->location = nullptr;
+    node->label    = nullptr;
 
     g_free(node);
 }
--- a/gnucash/html/gnc-html-history.h
+++ b/gnucash/html/gnc-html-history.h
@@ -28,8 +28,12 @@ typedef struct _gnc_html_history gnc_html_history;
 
 #include "gnc-html.h"
 
+typedef void (* gnc_html_history_destroy_cb)(gnc_html_history_node * n,
+        gpointer user_data);
+
 #ifdef __cplusplus
-extern "C" {
+extern "C"
+{
 #endif
 
 struct _gnc_html_history_node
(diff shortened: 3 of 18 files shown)


gnucash/html/gnc-html-history.cpp
```diff
@@ -119,16 +122,17 @@ g_strcmp(char * a, char * b)
 /********************************************************************
  * gnc_html_history_append
  ********************************************************************/
+extern "C"
```
jralls: No need for `extern "C"` on function definitions, it's only needed on the declarations.

gnucash/html/gnc-html-history.cpp
```diff
@@ -29,9 +29,9 @@
 
 struct _gnc_html_history
 {
-    GList * nodes;
-    GList * current_node;
-    GList * last_node;
+    GList *nodes;
+    GList *current_node;
+    GList *last_node;
```
jralls: Whitespace changes should be a separate commit.

jralls (changes requested):
Why do you think cmake needs to be told to use C++ linkage?
What's the benefit of extracting the compler flags to macros that are called only once?
The third commit message says that it includes adding C++ compile flags, but it doesn't.

Do NOT merge stable into your feature branch. Always rebase your feature branch on stable.

russ-gorby:
Why do you think cmake needs to be told to use C++ linkage?
[russ] eventually the C++ linker will be needed, and one of the common first steps in
converting large SW projects to C++ is to convert to using the C++ linker throughout
as it understands name mangling and can handle linking both C and C++ modules. 
I understand that is not strictly required at this time but it made sense to me to do that as a first step.
Do you have an issue with this or just requesting clarity?

What's the benefit of extracting the compler flags to macros that are called only once?
[russ] mostly to keep the top level CMakeFile.txt from becoming too cluttered and to keep the
functions/macros that control the compile flags co-located. There is a new macro to modify the
compile time flags on MINGW (-Wall -Werror …) that will be called from the cmakefile for the targets that are converted,
so that will be called from a number of cmakefiles as targets are converted.
It just made sense to me to keep all the compile time flags logic co-located. 
If you disagree, I can split them up again but the modify macro needs to stay to avoid duplication skew.


The third commit message says that it includes adding C++ compile flags, but it doesn't.
[russ] Good Catch. Thank you.
The cmakefile is supposed to have a call to mingw_promote_compile_flags()
I missed this when I segregatged patches. 


Do NOT merge stable into your feature branch. Always rebase your feature branch on stable.
[russ] WHY? Most github environments I’ve worked in prefer, and some even mandated merge over rebase. I can’t recall the details but I remember being in some situations where rebase was more clunky and error prone esp. when needing to include code from one or more other PRs in review. Anyway, I’m happy to abide but curious.

jralls:
> eventually the C++ linker will be needed,

Sorry, I guess I didn't state the question well. Why do you think that cmake can't figure out on its own to link those files as C++? Consider carefully: They're already C++, making heavy use of the C++ standard library and some use of Boost. If cmake tried to link them without `-lstdc++`/`libc++` it would fail.

> one of the common first steps in converting large SW projects to C++ 

Have you even looked through the code? We're *way* beyond first steps.

> mostly to keep the top level CMakeFile.txt from becoming too cluttered and to keep the
functions/macros that control the compile flags co-located…It just made sense to me to keep all the compile time flags logic co-located.

It's already in single block in the top-level CMakeLists.txt. Moving it to macros does nothing but move the clutter around.

> There is a new macro to modify the compile time flags on MINGW (-Wall -Werror …) that will be called from the cmakefile for the targets that are converted, so that will be called from a number of cmakefiles as targets are converted.

 Those flags not being set on Mingw's  CMAKE_CXX_FLAGS globally is an oversight that should be fixed in the compiler flags block in the top level CMakeLists.txt. 
Also using a macro like that to set target compiler options is wrong. That should be done with [target_compile_options](https://cmake.org/cmake/help/latest/command/target_compile_options.html).

>>  Do NOT merge stable into your feature branch. Always rebase your feature branch on stable.
>[russ] WHY? 

Because it makes loops in history that screw up bisects. 

> Most github environments I’ve worked in prefer, and some even mandated merge over rebase.

Examples?

> being in some situations where rebase was more clunky and error prone esp. when needing to include code from one or more other PRs in review

Easy to fix compared to the mess resulting from that series of PRs having a bunch of merge commits from different points in stable. Chains of PRs are kind of rare, but the way to test them in aggregate is to create a local branch off of main (which we call stable) and merge in the PRs in order.

gnucash/html/gnc-html-history.cpp
russ-gorby: Thats not been my experience. I tried that and it did not work. I haven't drilled down to exactly why.

gnucash/html/gnc-html-history.cpp
russ-gorby: OK

gnucash/html/gnc-html-history.cpp
russ-gorby: Why Definitions Need It Too
The extern "C" specifier controls linkage (name mangling and calling convention) for both declarations and definitions—it must match exactly for the linker to resolve symbols correctly. A declaration like extern "C" void func(); tells C++ callers to expect an unmangled C name, but if the definition lacks extern "C" (e.g., plain void func() {}), C++ mangles it (e.g., _Z4funcv), causing "undefined reference" linker errors

russ-gorby: did you mean "noexcept"?

jralls: > The extern "C" specifier controls linkage (name mangling and calling convention) for both declarations and definitions—it must match exactly for the linker to resolve symbols correctly. A declaration like extern "C" void func(); tells C++ callers to expect an unmangled C name, but if the definition lacks extern "C" (e.g., plain void func() {}), C++ mangles it (e.g., _Z4funcv), causing "undefined reference" linker errors

[Nope](https://en.cppreference.com/w/cpp/language/language_linkage.html). See the first section which says that it is applied only to declarations. See also the [C++ Wiki](https://isocpp.org/wiki/faq/mixing-c-and-cpp#call-cpp) where it's applied only to the declarations.

Note also that in libgnucash alone it's applied in 109 header files and only 9 cpp files; inspecting those 9 files will show that it's applied to declarations, mostly because the test file is called from the GLib test harness.

> did you mean "noexcept"?

No,  of course not. We're using C++17 so `noexcept` [is part of the function type](https://en.cppreference.com/w/cpp/language/noexcept_spec.html) and that must match for both the declaration and definition.
***Edit***
> Thats not been my experience. I tried that and it did not work. I haven't drilled down to exactly why.

Might it have been that you tried without both?

russ-gorby: Most of what I've read recommends adding them to both but if that is the standard for the project I'll fix those.
Thanks.

CMakeLists.txt
```diff
@@ -645,6 +645,10 @@ endif()
 if (MINGW)
   set( CMAKE_C_FLAGS "-Werror -Wall -Wmissing-prototypes -Wmissing-declarations ${CMAKE_C_FLAGS}")
   set( CMAKE_CXX_FLAGS "-DWINVER=0x0500 -D_EMULATE_GLIBC=0 ${CMAKE_CXX_FLAGS}") # Workaround for bug in gtest on mingw, see https://github.com/google/googletest/issues/893 and https://github.com/google/googletest/issues/920
+  function(mingw_promote_target_compile_flags tgt)
+    target_compile_options(${tgt} PRIVATE
+      $<$<COMPILE_LANGUAGE:CXX>:-Werror -Wall -Wmissing-declarations>)
+  endfunction(mingw_promote_target_compile_flags)
```
jralls: I'm going to fix the oversight in stable next week. It raises a bunch of errors and I've got to do a release this weekend so I don't think I'll have time to fix everything.

gnucash/html/gnc-html-webkit1.cpp
```diff
@@ -478,9 +483,8 @@ load_to_stream( GncHtmlWebkit* self, URLType type,
 
     if ( gnc_html_stream_handlers != NULL )
     {
-        GncHTMLStreamCB stream_handler;
-
-        stream_handler = g_hash_table_lookup( gnc_html_stream_handlers, type );
+        const gpointer p = g_hash_table_lookup( gnc_html_stream_handlers, type );
+        GncHTMLStreamCB stream_handler = reinterpret_cast<GncHTMLStreamCB>(p);
```
jralls, line 475: ~~Did you get a compile error with `static_cast<GncHTMLStreamCB>(p)`? `reinterpret_cast` isn't usually needed for casting to/from `void*`.~~
Nevermind, I tried in Win32 and found that GCC objects.

jralls (approved):
This is fine except for the MINGW C++ compiler flags.  Do you intend to make the code actually C++ with things like `nullptr`, replacing `typedef` with `using` and removing it from `typedef struct`, using STL containers, etc.? If so do you want to do it in another PR or add to this one?

CMakeLists.txt
russ-gorby: I did test the build and unittests on my 2 platforms I have access to (MINGW32/Wiindows11 and Ubnuntu(JammyJellyfish-WSL).
I saw no errors.
When you get a chance can you help me understand what I'm missing? Thank you John.

russ-gorby:
> This is fine except for the MINGW C++ compiler flags. Do you intend to make the code actually C++ with things like `nullptr`, replacing `typedef` with `using` and removing it from `typedef struct`, using STL containers, etc.? If so do you want to do it in another PR or add to this one?

My intent was to go wide before deep, and do minimal conversions to C++ of other targets first, then cycle back for more in-depth C++ work.

My thought was that nullptr, using, const, etc. are  mostly self contained in-line changes. They should probably be done at the same time as STL container use to avoid refactoring.
If we are talking STL containers just for internal use to replace C-constructs that could all be done together.
I'd ask for guidance from you on what you think is best for the project in this regard.
Once the interface functions become/use objects for their implementation logic, we will need unittest support which is the big win but this step could possibly require some redesign as well.

Let me know what you think.
And if we go width-first is there an target order preference?
I've got several other targets I've done a good chunk of work on but they were randomly chosen.

Thank you John.

CMakeLists.txt
jralls: That just means that there aren't any C++ errors in gnucash/html. The ones I encountered in my initial test are in Windows-specific code in libgnucash, which is why I'm implementing and fixing it instead of asking you to: It's out of scope for this PR. The end result will be that this macro and its invocation in gnucash/html/CMakeLists.txt won't be needed.

jralls:
> My intent was to go wide before deep, and do minimal conversions to C++ of other targets first, then cycle back for more in-depth C++ work.

I'd prefer you go a little deeper before proceeding to another directory. The already-done C++ work is at two levels: Keyword replacement with some work towards STL containers (most of libgnucash) or complete rewrites (SQL/DBI backend, Quotes, Options, a few engine classes, CSV importer, and the executables). It's easier to keep track of progress if conversion completes at least keyword replacement. For this html code we need to figure out a better way to support Windows than a 10 year old WebKit build before going any further because it doesn't make sense to do too much C++ conversion work if we're going to replace the whole thing. FWIW @Bob-IT has made a couple of skeleton alternatives in #1849 and #1868, neither of which is ideal.

Bob-IT:
One of my early tests was with [webview](https://github.com/webview/webview), it looks different from what I remember but maybe worth looking at again. If I can find my old test source files will give it a another try.

CMakeLists.txt
russ-gorby: OK thank you. I'll wait to see your changes. appreciate all the help on this.

jralls:
> One of my early tests was with [webview](https://github.com/webview/webview), it looks different from what I remember but maybe worth looking at again. If I can find my old test source files will give it a another try.

Unfortunately I don't think that's going to work for us: Its cross-platform model is native rather than Gtk-based, so while macOS wraps WebKit it's not WebKitGtk, and Windows (indirectly) wraps the built-in (and Chromium-based) WebView2 via .Net and a third-party library.

This is a bit off-topic for this PR and I don't want to distract @russ-gorby so let's have any further discussion on gnucash-devel.

russ-gorby:
> > My intent was to go wide before deep, and do minimal conversions to C++ of other targets first, then cycle back for more in-depth C++ work.
> 
> I'd prefer you go a little deeper before proceeding to another directory. The already-done C++ work is at two levels: Keyword replacement with some work towards STL containers (most of libgnucash) or complete rewrites (SQL/DBI backend, Quotes, Options, a few engine classes, CSV importer, and the executables). It's easier to keep track of progress if conversion completes at least keyword replacement. For this html code we need to figure out a better way to support Windows than a 10 year old WebKit build before going any further because it doesn't make sense to do too much C++ conversion work if we're going to replace the whole thing. FWIW @Bob-IT has made a couple of skeleton alternatives in #1849 and #1868, neither of which is ideal.

ok, I see. I can do that for this PR (go a little deeper). 
thank you.

CMakeLists.txt
jralls: @russ-gorby I've made the changes and pushed them.

russ-gorby: Just got them. Thanks!

russ-gorby:
> > > My intent was to go wide before deep, and do minimal conversions to C++ of other targets first, then cycle back for more in-depth C++ work.
> > 
> > 
> > I'd prefer you go a little deeper before proceeding to another directory. The already-done C++ work is at two levels: Keyword replacement with some work towards STL containers (most of libgnucash) or complete rewrites (SQL/DBI backend, Quotes, Options, a few engine classes, CSV importer, and the executables). It's easier to keep track of progress if conversion completes at least keyword replacement. For this html code we need to figure out a better way to support Windows than a 10 year old WebKit build before going any further because it doesn't make sense to do too much C++ conversion work if we're going to replace the whole thing. FWIW @Bob-IT has made a couple of skeleton alternatives in #1849 and #1868, neither of which is ideal.
> 
> ok, I see. I can do that for this PR (go a little deeper). thank you.

more to review now with the deeper cut.

gnucash/html/gnc-html-factory.h
```diff
@@ -26,12 +26,9 @@
 #include "gnc-html.h"
 
 #ifdef __cplusplus
-#define NOEXCEPT noexcept
-extern "C"
-{
-#else
-#define NOEXCEPT
+extern "C" {
```
jralls: Please revise the other commit instead.

gnucash/html/gnc-html-history.cpp
```diff
@@ -123,12 +121,9 @@ void
 gnc_html_history_append(gnc_html_history * hist,
                         gnc_html_history_node * node) noexcept
 {
-    GList * n;
-    gnc_html_history_node * hn;
-
     if (hist->current_node)
     {
-        hn = static_cast<gnc_html_history_node *>(hist->current_node->data);
+        gnc_html_history_node *hn = static_cast<gnc_html_history_node *>(hist->current_node->data);
```
jralls: You can use `auto` instead of repeating the cast type.

gnucash/html/gnc-html-history.h
```diff
@@ -23,18 +23,35 @@
 #ifndef GNC_HTML_HISTORY_H
 #define GNC_HTML_HISTORY_H
 
+#ifdef __cplusplus
+
+struct _gnc_html_history_node;
+struct _gnc_html_history;
+using gnc_html_history_node = struct _gnc_html_history_node;
+using gnc_html_history = struct _gnc_html_history;
```
jralls: `struct _gnc_html_history` here is C, not C++. `struct` in C++ declares a type and once declared you just use the type name.

gnucash/html/gnc-html-webkit1.cpp
```diff
@@ -121,47 +121,41 @@ static void impl_webkit_default_zoom_changed(gpointer prefs, gchar *pref, gpoint
...
-    PangoFontDescription *font_desc;
-    gdouble zoom = 1.0;
+    PangoFontDescription *font_desc = nullptr;
 
-    new_priv = static_cast<GncHtmlWebkitPrivate*>(
+    GncHtmlWebkitPrivate *new_priv = static_cast<GncHtmlWebkitPrivate*>(
         g_realloc( GNC_HTML(self)->priv, sizeof(GncHtmlWebkitPrivate) )
     );
-    priv = self->priv = new_priv;
+    GncHtmlWebkitPrivate *priv = self->priv = new_priv;
```
jralls: You can use `auto` for most if not all of the variable decls you've moved.

gnucash/html/gnc-html-webkit1.cpp
```diff
@@ -270,9 +264,9 @@ gnc_html_webkit_finalize( GObject* obj )
 {
     GncHtmlWebkit* self = GNC_HTML_WEBKIT(obj);
 
-//	if( self->priv != NULL ) {
+//	if( self->priv != nullptr ) {
```
jralls: This is a good opportunity to remove commented-out code.

gnucash/html/gnc-html-webkit1.cpp
```diff
@@ -472,60 +467,59 @@ static gboolean
...
         GncHTMLStreamCB stream_handler = reinterpret_cast<GncHTMLStreamCB>(p);
         if ( stream_handler )
         {
             GncHtml *weak_html = GNC_HTML(self);
-            gboolean ok;
 
             g_object_add_weak_pointer(G_OBJECT(self), (gpointer *)(&weak_html));
 
-            ok = stream_handler(location, &fdata, &fdata_len);
+            gboolean ok = stream_handler(location, &fdata, &fdata_len);
```
jralls: Use `bool` in C++ code instead of `gboolean`.

gnucash/html/gnc-html-webkit1.cpp
```diff
@@ -472,60 +467,59 @@ static gboolean
...
-            if (g_strstr_len(fdata, -1, "<object classid=") != NULL) {
-              gchar *new_fdata;
-              new_fdata = handle_embedded_object(self, fdata);
-              g_free(fdata);
-              fdata = new_fdata;
-            }
+                // Until webkitgtk supports download requests, look for "<object
+                // classid=" indicating the beginning of an embedded graph.  If
+                // found, handle it
+                if (g_strstr_len(fdata, -1, "<object classid=") != nullptr) {
```
jralls: Since you're fixing whitespace, put the opening curly brace on its own line to comply with our [coding standard](https://wiki.gnucash.org/wiki/CodingStandard).

gnucash/html/gnc-html-webkit1.h
```diff
@@ -35,10 +35,23 @@ G_BEGIN_DECLS
 #define GNC_IS_HTML_WEBKIT_CLASS(k)   (G_TYPE_CHECK_CLASS_TYPE((k), GNC_TYPE_HTML_WEBKIT))
 #define GNC_HTML_WEBKIT_GET_CLASS(o)  (G_TYPE_INSTANCE_GET_CLASS((o), GNC_TYPE_HTML_WEBKIT, GncHtmlWebkitClass))
 
+#ifdef __cpluusplus
```
jralls: gnc-html-webkit1.h is `included` only by gnc-html-webkit1.cpp; same for gnc-html-webkit2.h and gnc-html-webkit2.cpp. They should therefore be renamed `hpp` and don't need the `#ifdef __cplusplus` (which, BTW, you misspelled).

This `typedef` is anyway misplaced. Gnome does this as a data-hiding exercise by defining `struct _Foo` in the implementation file and typedeffing the name without the leading underscore for public use. But these files define the structs after typedeffing them, so it's the equivalent of just writing
```
typedef struct
{
//stuff
} Foo;
```
and of course in C++ we write that as
```
struct Foo
{
//stuff
};
```

gnucash/html/gnc-html.cpp
```diff
@@ -764,33 +749,30 @@ gnc_html_encode_string(const char * str) noexcept
...
 
 
 char *
 gnc_html_decode_string(const char * str) noexcept
 {
     static const gchar * safe = "$-._!*(),"; /* RFC 1738 */
+
+    if (!str) return nullptr;
+
     GString * decoded  = g_string_new ("");
```
jralls: Replace the `GString` with a `std::string`.

jralls (changes requested):

gnucash/html/gnc-html-webkit1.cpp
russ-gorby: ok, didn't want to step on anyone's toes, but sure.

gnucash/html/gnc-html-webkit1.cpp
russ-gorby: OK, what about all the other glib types? are they also candidates for C++ type replacements?

gnucash/html/gnc-html-history.cpp
russ-gorby: ah, yes. thank you.

gnucash/html/gnc-html-factory.h
russ-gorby: you mean make the { on separate line like before? I mean to, but missed this one.
The NOEXCEPT defines are not needed since they come in from gnc-html.h

gnucash/html/gnc-html-history.h
russ-gorby: I thought that struct in C++ was essentially a class with some special properties so this is just a forward reference to a class. I was trying to not redesign the header files for this target that rely heavily on fwd declarations of structs. I'd rather leave that for another PR and keep the same layout for this pass.

gnucash/html/gnc-html-webkit1.cpp
russ-gorby: my only concern is that over use of auto can make the code more opaque to read. E.g. the compiler knows what type some random function returns but many code readers may not. If its obvious from the context what type is in use then sure. Is that acceptable?

gnucash/html/gnc-html-webkit1.cpp
russ-gorby: so far I've head gboolean and GString should be wholesale replaced.

gnucash/html/gnc-html.cpp
russ-gorby: OK

gnucash/html/gnc-html-webkit1.h
russ-gorby: OK let me look at this

gnucash/html/gnc-html-factory.h
jralls: No, I mean remove the NOEXCEPT defines from the first commit instead of adding them there and removing them in the second commit.

gnucash/html/gnc-html-history.h
jralls: In C++ `struct` and `class` are the same except that in a `struct` the default access is `public` while in a `class` the default access is `private`.  They both declare type names so the keyword is needed only in declarations and definitions. When using you need only the name. In C the type name of a struct includes the `struct` keyword unless you `typedef` it. 

So having declared`struct _gnc_html_history_node` you don't need to use `struct` in the `using` expression, you can just say `using gnc_html_history_node = _gnc_html_history_node;`. Of course if we didn't have to let C functions know about it we could just declare it as `gnc_html_history_node` (or to better comply with our coding standard `GncHTMLHistoryNode`) and be done with it. Since we're defining the struct just below separating the typedef from the definition is dumb. It would be better to just say 
```
typedef struct
{
//stuff
} gnc_html_history_node;
```
The separation of the typedef and definition with the leading underscore (discouraged in C++ where that's supposed to be reserved to the compiler and standard library)  is a Gnome data-hiding convention where the typedef goes in the header and the definition in the implementation file. 

For headers that are used in both languages I think using both `typedef` and `using` separated by a preprocessor macro just adds clutter without contributing anything.

gnucash/html/gnc-html-webkit1.cpp
jralls: Mostly. The problem is that as you get more familiar with the code the population of what's obvious increases, so a lot more of them are obvious to me than will be to you.

But in the instant case, `new_priv` is statically cast one line up so it should be pretty obvious to anybody what type `priv` is. It should be pretty obvious that `gtk_widget_get_style_context` returns a `GtkStyleContext*` too. It might be less obvious that `pango_font_description_get_family` returns a `const char*`, in particular the `const` part meaning that unlike most strings coming from Gnome functions it doesn't need to be freed when we're done with it so that's a good reason to leave the type in.

gnucash/html/gnc-html-webkit1.cpp
jralls: Only if local. If you change parameters or return values then they have to be changed at the points of use too, which may not be feasible. And yes, in general prefer standard types to the Gnome typedefs, and avoid all flavors of `long` because it means `int32_t` on WIN64 and `int64_t` on all other 64-bit platforms.

gnucash/html/gnc-html.cpp
russ-gorby: I don't think I can get rid of GString w/o changing the callers since these functions return a GString and the callers will call g_free() on them - can I?

gnucash/html/gnc-html-webkit1.h
russ-gorby: right. these can go to .hpp and eliminate all the typedef stull

gnucash/html/gnc-html-webkit1.cpp
russ-gorby: ok

gnucash/html/gnc-html.cpp
jralls: No, the function returns a `char*`. That's a C string, not a `GString`. You can use a `std::string` internally and return `strdup(my_string.c_str())` for the `char*` that must be `g_free`'d.

gnucash/html/gnc-html-history.h
russ-gorby: For headers that are used in both languages, just use the typedef for both then?

gnucash/html/gnc-html-factory.h
russ-gorby: oh I see. perhaps I should just squash the commits together.

gnucash/html/gnc-html-history.h
jralls: I think so, and where both the decl and def are in the same file combine them into a single def.

russ-gorby: after seeing the difference it is much cleaner. thank you.

gnucash/html/gnc-html.cpp
russ-gorby: Oh, ok. so you can g_free() just a raw string allocation, it doesn't expect some kind of GString anatomy.
I'll fix this.

jralls: The current last line of the function is [`g_string_free`](https://docs.gtk.org/glib/method.String.free.html). It steals the `char*` (meaning that it copies it to a variable and nulls the `GString`'s member variable so that the `GString`'s `dispose` can't free it) from the `GString` and frees the latter, keeping the GString* local to the function.

Spend some time getting familiar with GLib, GObject, GIO, and Gtk so you understand what you're converting from. GLib is primarily an abstraction library to smooth over the differences between Unix and Windows (and originally the differences between compilers, though that's much less of a problem in the 2020s than it was in the 1990s). GObject is a not entirely successful attempt to support object-oriented programming in C. GIO is a services library; most of its pieces started off in Gtk and were abstracted and moved down the food chain to focus Gtk on the GUI components of an application. The [documentation](https://docs.gtk.org/) is generally pretty good.

gnucash/html/gnc-html-factory.h
russ-gorby: done

gnucash/html/gnc-html.cpp
russ-gorby: thank you

gnucash/html/gnc-html-factory.hpp
```diff
@@ -25,14 +25,6 @@
...
-#ifdef __cplusplus
-extern "C" {
-#endif
-
-GncHtml* gnc_html_factory_create_html( void );
-
-#ifdef __cplusplus
-}
-#endif
+extern "C" GncHtml* gnc_html_factory_create_html( void ) NOEXCEPT;
```
jralls, line 28: If it's only called from C++ it doesn't need C linkage.

PR #2198, opened 2026-04-03
Bug 799759 - Users can't Enable entries via Checkboxes on Scheduled T…

agwekixj:
…ransactions Page

Implement the toggle of the enabled box in on the Scheduled Transactions list.

### DIFF
--- a/gnucash/gnome-utils/gnc-sx-list-tree-model-adapter.c
+++ b/gnucash/gnome-utils/gnc-sx-list-tree-model-adapter.c
@@ -342,13 +342,12 @@ gsltma_proxy_sort_column_changed (GtkTreeSortable *sortable, gpointer user_data)
     g_signal_emit_by_name (user_data, "sort-column-changed");
 }
 
-static gint
-_name_comparator (GtkTreeModel *model, GtkTreeIter *a, GtkTreeIter *b, gpointer user_data)
+static gint 
+_sort_iterators_to_instances (gpointer *user_data, GtkTreeIter *a, GtkTreeIter *b, 
+        gint (*instances_comparator)(GncSxInstances *a, GncSxInstances *b))
 {
-    gint rtn;
     GncSxListTreeModelAdapter *adapter = GNC_SX_LIST_TREE_MODEL_ADAPTER(user_data);
     GncSxInstances *a_inst, *b_inst;
-    gchar *a_caseless, *b_caseless;
 
     a_inst = gsltma_get_sx_instances_from_orig_iter (adapter, a);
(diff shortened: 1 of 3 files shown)


jralls:
LGTM. @Bob-IT ?

Bob-IT:
Have tested this and it does work as expected, looks like it was based on a similar call back in gnc-tree-view-account.

One question, how likely once released into the wild do you think a request will come in to be able to change multiple rows at once.

Could possibly be achieved by adding a couple of entries to the right mouse popup menu.

jralls:
> Could possibly be achieved by adding a couple of entries to the right mouse popup menu.

Or making the spacebar toggle the checkbox instead of opening the editor on the first entry in the selection. That would mirror the behavior in the Reconcile Window and wouldn't require adding a menu item. IMO &lt;return&gt; should open the selected row's editor, and it should be disabled as ambiguous when there's a multiple selection.

Bob-IT:
> Or making the spacebar toggle the checkbox instead of opening the editor on the first entry in the selection. That would mirror the behavior in the Reconcile Window and wouldn't require adding a menu item. IMO <return> should open the selected row's editor, and it should be disabled as ambiguous when there's a multiple selection.

Seems reasonable to me, I had forgotten there was an example in the Reconcile Window.

jralls:
>  able to change multiple rows at once.

@Bob-IT It doesn't appear Sherlock has much interest in adding that. Would it be an easy thing for you to do?

Bob-IT:
> > able to change multiple rows at once.
> 
> @Bob-IT It doesn't appear Sherlock has much interest in adding that. Would it be an easy thing for you to do?

I will have a look next week.

agwekixj:
> > > able to change multiple rows at once.
> > 
> > 
> > @Bob-IT It doesn't appear Sherlock has much interest in adding that. Would it be an easy thing for you to do?
> 
> I will have a look next week.

Honestly, we’ve used this patch since 4.12 and haven’t had the need.  

Any how, I’ve completed an implementation which mimics Reconcile Window spacebar behavior. 

I am just working on a couple GTK quirks I noticed in the original patch while testing.  I should have an update shortly.

agwekixj:
It may also be good idea to provide a secondary column sort using the Name column for all the other columns.

agwekixj:
Latest push includes using the Name column to provide the secondary column sort.

jralls:
@agwekixj Are you done?

agwekixj:
> @agwekixj Are you done?

Yes.  Unless you see something else?

jralls:
Nope, I was about to merge it and you pushed that last commit while I was giving it a last look-through. I don't want to merge something you're still working on.

PR #2199, opened 2026-04-06
Bug799751 - autocomplete crash and some

Bob-IT:
The first two commits are really just some cosmetic changes regarding space and naming of GncPluginPage... 

The third commit fixes the bug, it was not crashing for me but I think it fixes the issue regarding the General Journal autocomplete. What I think was happening is that the default query for only 30 days was being used to create the register view when selecting 'Show All' and the order was changed from standard order, the query was being updated but the register view was not being refreshed which I think this was a side effect from the change to just load the register when it is visible.
The fix for this is not do any refreshing when the register is created or recreated and leave it to the register focus call back to request a refresh when it is visible.
As a follow up to this the next commit stops the register refreshing when the application is quitting.

The next three commits are something I have wanted to do and that is split out the filter and sort dialogs to separate files and hopefully making it more manageable.

The last five commits is my attempt to convert those files to c++ with no doubt numerous mistakes so can be easily discarded if required.

One last thing to note, while testing this, the settings for 'today', 'earliest' and 'latest' are not saved correctly which reminded me of a bug Chris raised some time ago. I am not really sure the point of saving 'Earliest' and 'Latest' so was considering changing the dialog to use the existing date widget and a period select widget similar to the preference page.
This would obviously be a separate PR.

### DIFF
--- a/gnucash/gnome/gnc-plugin-page-register-filter.cpp
+++ b/gnucash/gnome/gnc-plugin-page-register-filter.cpp
@@ -107,6 +107,18 @@ static struct status_action status_actions[] =
     { "filter_status_unreconciled", CLEARED_NO, nullptr },
     { nullptr, 0, nullptr },
 };
+#ifdef skip
+static inline bool
+gboolean_to_bool (gboolean value)
+{
+    return value ? true : false;
+}
+#endif
+static inline gboolean
+bool_to_gboolean (bool value)
+{
+    return value ? TRUE : FALSE;
+}
 
 static const gchar*
 get_filter_default_num_of_days (GNCLedgerDisplayType ledger_type)
@@ -484,7 +496,7 @@ gnc_ppr_filter_update_register (GncPluginPage* plugin_page)
     int filter_changed = 0;
 
     /* Set the filter for the split register and status of save filter button */
-    fd->save_filter = FALSE;
+    fd->save_filter = false;
 
     gchar* filter_str = gnc_ppr_filter_get_filter (gsr, ledger_type);
     gchar** filter = g_strsplit (filter_str, ",", -1);
@@ -529,7 +541,7 @@ gnc_ppr_filter_update_register (GncPluginPage* plugin_page)
     }
 
     if (filter_changed != 0)
-        fd->save_filter = TRUE;
+        fd->save_filter = true;
 
     fd->original_save_filter = fd->save_filter;
     g_strfreev (filter);
@@ -548,7 +560,7 @@ gnc_ppr_filter_update_register (GncPluginPage* plugin_page)
         {
             fd->days = 0;
             fd->cleared_match = (cleared_match_t)g_ascii_strtoll (DEFAULT_FILTER, nullptr, 16);
-            fd->save_filter = FALSE;
+            fd->save_filter = false;
         }
 
         fd->original_days = fd->days;
@@ -941,9 +953,9 @@ gnc_ppr_filter_save_cb (GtkToggleButton* button,
 
     /* Compute the new save filter status */
     if (gtk_toggle_button_get_active (button))
-        fd->save_filter = TRUE;
+        fd->save_filter = true;
     else
-        fd->save_filter = FALSE;
+        fd->save_filter = false;
 
     LEAVE(" ");
 }
@@ -1042,10 +1054,9 @@ gnc_ppr_filter_response_cb (GtkDialog* dialog,
 
 void
 gnc_ppr_filter_by (GncPluginPage *plugin_page, Query *query,
-                   FilterData *fd, gboolean show_save_button)
+                   FilterData *fd, bool show_save_button)
 {
     time64 start_time, end_time, time_val;
-    gboolean sensitive, value;
 
     /* Create the dialog */
     auto builder = gtk_builder_new();
@@ -1069,9 +1080,9 @@ gnc_ppr_filter_by (GncPluginPage *plugin_page, Query *query,
     {
         auto toggle = GTK_WIDGET(gtk_builder_get_object (builder,
                                                      status_actions[i].action_name));
-        value = fd->cleared_match & status_actions[i].value;
+        bool value = fd->cleared_match & status_actions[i].value;
         status_actions[i].widget = toggle;
-        gtk_toggle_button_set_active (GTK_TOGGLE_BUTTON(toggle), value);
+        gtk_toggle_button_set_active (GTK_TOGGLE_BUTTON(toggle), bool_to_gboolean (value));
     }
     fd->original_cleared_match = fd->cleared_match;
 
@@ -1080,7 +1091,7 @@ gnc_ppr_filter_by (GncPluginPage *plugin_page, Query *query,
         gtk_toggle_button_set_active (GTK_TOGGLE_BUTTON(button), TRUE);
 
     // hide the save button if appropriate
-    gtk_widget_set_visible (GTK_WIDGET(button), show_save_button);
+    gtk_widget_set_visible (GTK_WIDGET(button), bool_to_gboolean (show_save_button));
 
     /* Set up number of days */
     fd->num_days = GTK_WIDGET(gtk_builder_get_object (builder,
@@ -1125,13 +1136,14 @@ gnc_ppr_filter_by (GncPluginPage *plugin_page, Query *query,
     fd->end_date_choose = GTK_WIDGET(gtk_builder_get_object (builder, "end_date_choose"));
     fd->end_date_today = GTK_WIDGET(gtk_builder_get_object (builder, "end_date_today"));
 
+    bool sensitive;
     {
         /* Start date info */
         if (start_time == 0)
         {
             button = GTK_WIDGET(gtk_builder_get_object(builder, "start_date_earliest"));
             time_val = xaccQueryGetEarliestDateFound (query);
-            sensitive = FALSE;
+            sensitive = false;
         }
         else
         {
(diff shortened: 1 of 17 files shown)


gnucash/gnome/gnc-split-reg.c
```diff
@@ -1939,7 +1939,8 @@ gnc_split_reg_double_line_cb (GtkWidget *w, gpointer data)
 }
 
 static void
-gnc_split_reg_sort_force( GNCSplitReg *gsr, SortType sort_code, gboolean force )
+gnc_split_reg_sort_force( GNCSplitReg *gsr, SortType sort_code,
+                          gboolean force, gboolean refresh )
```
jralls: Functions with more than one boolean argument are error prone and hard to interpret at the call site. You can improve that with enums:
```
typedef enum { no_force, force } ForceSort;
typedef enum { no_refresh, refresh } Refresh;
/*...*/
static void gnc_split_reg_sort_maybe_force_maybe_refresh(GNCSpllitReg*, SortType, ForceSort, Refresh);
```
The function name there is a little unwieldy. `gnc_split_reg_sort` is used in the file as an abbreviation of `gnc_split_reg_sort_force(foo, bar, no_force, refresh)`. IMO it would be cleaner to remove that and rename `gnc_split_reg_sort_force` to `gnc_split_reg_sort` instead of to `gnc_split_reg_sort_maybe_force_maybe_refresh`. That way the force and refresh values are explicit at every callback definition.

gnucash/gnome/gnc-split-reg.c
```diff
@@ -2557,15 +2565,18 @@ gnc_split_reg_get_sort_type( GNCSplitReg *gsr )
...
 
 void
-gnc_split_reg_set_sort_type( GNCSplitReg *gsr, SortType t )
+gnc_split_reg_set_sort_type( GNCSplitReg *gsr, SortType t, gboolean refresh )
 {
-    gnc_split_reg_sort( gsr, t );
+    if (refresh)
+        gnc_split_reg_sort( gsr, t );
+    else
+        gnc_split_reg_sort_without_refresh( gsr, t);
```
jralls: `gnc_split_reg_sort_force(gsr, t, FALSE, refresh)` (or better with the mods suggested at the gnc_split_reg_sort_force definition) is cleaner.

gnucash/gnome/gnc-split-reg.h
```diff
@@ -207,7 +207,7 @@ GtkWidget *gsr_create_summary_bar( GNCSplitReg *gsr );
  * query to be redone even if SortType has not changed.
  **/
 SortType gnc_split_reg_get_sort_type( GNCSplitReg *gsr );
-void gnc_split_reg_set_sort_type( GNCSplitReg *gsr, SortType t );
+void gnc_split_reg_set_sort_type( GNCSplitReg *gsr, SortType t, gboolean refresh);
 void gnc_split_reg_set_sort_type_force( GNCSplitReg *gsr, SortType t, gboolean force);
```
jralls: Not part of this PR but I don't like the naming: `set_sort_type` indicates that you're setting `SortType` for future sort actions, not performing a sort. It also implies that `refresh` is part of the type that you're setting so `get_sort_type` should return a struct instead of a single value.

gnucash/gnome/gnc-plugin-page-register-filter.cpp
```diff
@@ -49,15 +49,16 @@
...
+#include "gnc-plugin-page-register-filter.hpp"
 
 #define DEFAULT_FILTER_NUM_DAYS_GL  "30"
-#define CLEARED_VALUE "cleared_value"
 #define DEFAULT_FILTER "0x001f"
 
 /* This static indicates the debugging module that this .o belongs to.  */
 static QofLogModule log_module = GNC_MOD_GUI;
 
+extern "C"
```
jralls, line 90: Needs a comment explaining that the callbacks are passed to Gtk functions and so require C linkage.

gnucash/gnome/gnc-plugin-page-register-filter.cpp
```diff
@@ -1078,13 +1073,13 @@ gnc_ppr_filter_by (GncPluginPage *plugin_page, Query *query,
                                   gnc_window_get_gtk_window (GNC_WINDOW(GNC_PLUGIN_PAGE(plugin_page)->window)));
 
     /* Translators: The %s is the name of the plugin page */
-    title = g_strdup_printf (_ ("Filter %s by…"),
-                             gnc_plugin_page_get_page_name (GNC_PLUGIN_PAGE(plugin_page)));
+    gchar* title = g_strdup_printf (_ ("Filter %s by…"),
```
jralls: You can use `auto` instead of an explicit type when the compiler (and anyone reading the code) can deduce the type from the return type of the function.

gnucash/gnome/gnc-plugin-page-register-filter.cpp
```diff
@@ -672,7 +670,7 @@ gnc_ppr_filter_status_clear_all_cb (GtkButton* button,
     FilterData *fd = gnc_plugin_page_register_get_filter_data (GNC_PLUGIN_PAGE(page));
 
     /* Turn off all the check menu items */
-    for (i = 0; status_actions[i].action_name; i++)
+    for (int i = 0; status_actions[i].action_name; i++)
```
jralls: If you change `status_actions` from a C array into a C++ `std::array` you can get rid of the sentinel and have 
```
     for (auto action : status_actions)
     {
            g_signal_handlers_block_by_func(action.widget, 
```

jralls (changes requested):
The design of the filter module can be greatly improved by having a struct or tuple of the filter components in their proper types (e.g. an int for the number of days) and leave the string stuff for explicit IO functions that compose and write or read and parse.

gnucash/gnome/gnc-split-reg.c
Bob-IT: I have change this to as suggested.

gnucash/gnome/gnc-plugin-page-register-filter.cpp
Bob-IT: Added comment.

gnucash/gnome/gnc-plugin-page-register-filter.cpp
Bob-IT: added some more uses of auto

gnucash/gnome/gnc-plugin-page-register-filter.cpp
Bob-IT: Changed to use std:array but found that if I used this format for the setup in gnc_ppr_filter_by the widget value would be null outside of the setup loop so used the original format for this one.

jralls: Sorry, you've lost me. What widget value is null? You mean [here](https://github.com/Bob-IT/gnucash/blob/10c8a9cd05e59bbaa7b8606d9f304c8e4707163c/gnucash/gnome/gnc-plugin-page-register-filter.cpp#L1156), right?

Bob-IT: Sorry for the lack clarity, yes that location. I was going to replace that loop with this minus the print statement...
```
    for (auto action : status_actions)
    {
        auto toggle = GTK_WIDGET(gtk_builder_get_object (builder, action.action_name.c_str()));
        bool value = fd->cleared_match & action.value;
        action.widget = toggle;
g_print("action.name, '%s', action.widget %p\n", action.action_name.c_str(), action.widget);
        gtk_toggle_button_set_active (GTK_TOGGLE_BUTTON(toggle), bool_to_gboolean (value));
    }
```
action.name, 'filter_status_reconciled', action.widget 0x55b382d113d0
action.name, 'filter_status_cleared', action.widget 0x55b382d12ed0
action.name, 'filter_status_voided', action.widget 0x55b382d14ae0
action.name, 'filter_status_frozen', action.widget 0x55b382d166f0
action.name, 'filter_status_unreconciled', action.widget 0x55b382d0f380

but if I try and retrieve the widget with this in the other locations...
```
    for (auto action : status_actions)
    {
g_print("  action.name, '%s', action.widget %p\n", action.action_name.c_str(), action.widget);
    }
```

I get this...

  action.name, 'filter_status_reconciled', action.widget (nil)
  action.name, 'filter_status_cleared', action.widget (nil)
  action.name, 'filter_status_voided', action.widget (nil)
  action.name, 'filter_status_frozen', action.widget (nil)
  action.name, 'filter_status_unreconciled', action.widget (nil)

jralls: Ah. Try `auto& action : status_actions` so you get a reference to the `status_action` instead of a copy. Everywhere else you'll want `const auto& action` to indicate that you're not changing the action's contents.

Bob-IT: Thanks, that fixed this.

Bob-IT:
@jralls I mucked up an interactive rebase of this branch by using the wrong commit after I had rebased this branch from an updated stable branch.
Is it easy to correct or I will just scrap this and start a new branch?

jralls:
@Bob-IT did you try `git rebase stable bug799751` and then a force push, or is that what messed it up in the first place?

Bob-IT:
> @Bob-IT did you try `git rebase stable bug799751` and then a force push, or is that what messed it up in the first place?

Thanks, that gave me an idea of doing another interactive rebase, drop those commits and then do a normal rebase from stable which seems to of done the trick, whether I needed to do the dropping I am not sure but it looks OK now. 

I think I have done all the changes you have suggested unless I have missed something.

jralls (approved):
Looks good. Merge when ready.

Bob-IT:
Merged.

PR #2201, opened 2026-04-09
Add port field to database connection dialog

jlaustill:
## Summary

- Adds a **Port** entry field to the PostgreSQL/MySQL database connection dialog
- When left blank, behavior is unchanged (backend defaults to 5432 for PostgreSQL, 3306 for MySQL)
- The backend already fully supported custom ports via `gnc_uri_create_uri()` and `dbi_conn_set_option_numeric()` — this change wires up the UI to pass the port through instead of hardcoding `0`

## Motivation

Managed database services (Digital Ocean, AWS RDS, Google Cloud SQL, etc.) commonly use non-standard ports. There was no way to specify a port in the connection dialog, making it impossible to connect to these services without workarounds.

## Changes

- **`gnucash/gtkbuilder/dialog-file-access.glade`** — Added Port label and entry field (row 4) with placeholder text "Default: 5432"
- **`gnucash/gnome-utils/dialog-file-access.c`** — Added `tf_port` to `FileAccessWindow` struct, reads port value in `geturl()`, passes it to `gnc_uri_create_uri()`

## Test plan

- [ ] Open File → Save As → select PostgreSQL — verify Port field appears below Password
- [ ] Leave Port blank, connect to localhost:5432 — verify default behavior unchanged
- [ ] Enter a custom port (e.g., 25060), connect — verify connection uses the specified port
- [ ] Verify MySQL connection dialog also shows the Port field

### DIFF
--- a/gnucash/gnome-utils/dialog-file-access.c
+++ b/gnucash/gnome-utils/dialog-file-access.c
@@ -64,12 +64,15 @@ typedef struct FileAccessWindow
     GtkEntry            *tf_database;
     GtkEntry            *tf_username;
     GtkEntry            *tf_password;
+    GtkEntry            *tf_port;
 } FileAccessWindow;
 
 void gnc_ui_file_access_file_activated_cb( GtkFileChooser *chooser,
         FileAccessWindow *faw );
 void gnc_ui_file_access_response_cb( GtkDialog *, gint, GtkDialog * );
 static void cb_uri_type_changed_cb( GtkComboBoxText* cb );
+static void port_insert_text_cb( GtkEditable *editable, const gchar *text,
+                                 gint length, gint *position, gpointer data );
 
 static gchar*
 geturl( FileAccessWindow* faw )
@@ -82,6 +85,8 @@ geturl( FileAccessWindow* faw )
     gchar* type = NULL;
     /* Not const as return value of gtk_file_chooser_get_filename must be freed */
     gchar* path = NULL;
+    gint32 port = 0;
+    const gchar* port_text = NULL;
 
     type = gtk_combo_box_text_get_active_text (faw->cb_uri_type);
     if (gnc_uri_is_file_scheme (type))
@@ -101,7 +106,12 @@ geturl( FileAccessWindow* faw )
         password = gtk_entry_get_text( faw->tf_password );
     }
 
-    url = gnc_uri_create_uri (type, host, 0, username, password, path);
+    g_assert (faw->tf_port != NULL);
+    port_text = gtk_entry_get_text (faw->tf_port);
+    if (port_text && *port_text)
+        port = atoi (port_text) & 0xffff;
+
+    url = gnc_uri_create_uri (type, host, port, username, password, path);
 
     g_free (type);
     g_free (path);
@@ -210,10 +220,26 @@ set_widget_sensitivity_for_uri_type( FileAccessWindow* faw, const gchar* uri_typ
     else if ( strcmp( uri_type, "mysql" ) == 0 || strcmp( uri_type, "postgres" ) == 0 )
     {
         set_widget_sensitivity( faw, /* is_file_based_uri */ FALSE );
+        gtk_entry_set_placeholder_text( faw->tf_port,
+            strcmp( uri_type, "mysql" ) == 0 ? _("Default: 3306") : _("Default: 5432") );
     }
     else
     {
         g_assert( FALSE );
+    }
+}
+
+static void
+port_insert_text_cb( GtkEditable *editable, const gchar *text, gint length,
+                     gint *position, gpointer data )
+{
+    for ( gint i = 0; i < length; i++ )
+    {
+        if ( !g_ascii_isdigit( text[i] ) )
+        {
+            g_signal_stop_emission_by_name( G_OBJECT(editable), "insert-text" );
+            return;
+        }
     }
 }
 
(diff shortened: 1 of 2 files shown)


jralls:
This isn't really necessary, you can set the port in the usual way by specifying e.g. `localhost:1234` in the host field just like you would in a web browser.

jlaustill:
> This isn't really necessary, you can set the port in the usual way by specifying e.g. `localhost:1234` in the host field just like you would in a web browser.

That host:port workaround is not documented anywhere in the UI, and has been a recurring source of confusion for users for over 
  17 years:                                                                                                                       
                                                                                                                                  
  - GNOME Bugzilla[ #592357](https://bugzilla.gnome.org/show_bug.cgi?id=592357) (2009) — The original report that port couldn't be specified at all. The "fix" was to make the         
  host:port colon syntax work rather than add a dedicated field — a workaround that has required explaining to users ever since.
  - [gnucash-user mailing list, Nov 2018](https://lists.gnucash.org/pipermail/gnucash-user/2018-November/080891.html) — A developer explicitly acknowledges "there is no separate field for it" when a user is  
  confused about port configuration.                                                                                              
  - [gnucash-user mailing list, Aug 2022](https://lists.gnucash.org/pipermail/gnucash-user/2022-August/102511.html) — Yet another user had no idea they could/should type host:port together, requiring manual
   instruction.                                                                                                                   
                             
  Every other database client (pgAdmin, MySQL Workbench, DBeaver, etc.) provides a dedicated port field. The host:port convention 
  belongs in browser address bars, not database connection dialogs. It also behaves inconsistently across platforms — it worked
  for me on Windows but failed on Linux.                                                                                          
                             
  A dedicated port field is the standard, and the work is already done. Happy to address any concerns about the implementation.

gnucash/gtkbuilder/dialog-file-access.glade
```diff
@@ -278,6 +278,34 @@
...
+                    <property name="top-attach">4</property>
+                  </packing>
+                </child>
+                <child>
+                  <object class="GtkEntry" id="tf_port">
+                    <property name="visible">True</property>
+                    <property name="can-focus">True</property>
+                    <property name="invisible-char">●</property>
+                    <property name="activates-default">True</property>
+                    <property name="placeholder-text" translatable="yes">Default: 5432</property>
```
jralls: That displays the wrong default for MySQL.

gnucash/gtkbuilder/dialog-file-access.glade
```diff
@@ -278,6 +278,34 @@
...
+                    <property name="primary-icon-activatable">False</property>
+                    <property name="secondary-icon-activatable">False</property>
+                  </object>
+                  <packing>
+                    <property name="left-attach">1</property>
+                    <property name="top-attach">4</property>
```
jralls: Port belongs in position 1, right after host. It really belongs on the same line, it's part of the host specification, but that would make the line too long.

gnucash/gnome-utils/dialog-file-access.c
```diff
@@ -308,6 +317,7 @@ gnc_ui_file_access (GtkWindow *parent, int type)
     gtk_entry_set_text( faw->tf_database, default_db );
     faw->tf_username = GTK_ENTRY(gtk_builder_get_object (builder, "tf_username" ));
     faw->tf_password = GTK_ENTRY(gtk_builder_get_object (builder, "tf_password" ));
+    faw->tf_port = GTK_ENTRY(gtk_builder_get_object (builder, "tf_port" ));
```
jralls, line 337: Add an insert-text event handler to limit port input to numbers. [This stackoverflow answer](https://stackoverflow.com/a/55893352) has an example implementation.

jralls (changes requested):

gnucash/gnome-utils/dialog-file-access.c
```diff
@@ -101,7 +108,22 @@ geturl( FileAccessWindow* faw )
         password = gtk_entry_get_text( faw->tf_password );
     }
 
-    url = gnc_uri_create_uri (type, host, 0, username, password, path);
+    g_assert (faw->tf_port != NULL);
+    port_text = gtk_entry_get_text (faw->tf_port);
+    if (port_text && *port_text)
+    {
+        parsed_port = g_ascii_strtol (port_text, &port_endptr, 10);
```
jralls: There is no `g_ascii_strtol`. This is the cause of the CI failure.

gnucash/gnome-utils/dialog-file-access.c
```diff
@@ -101,7 +108,22 @@ geturl( FileAccessWindow* faw )
...
+        if (port_endptr != port_text && *port_endptr == '\0' && parsed_port > 0 && parsed_port <= 65535)
+            port = (gint32) parsed_port;
```
jralls: This seems unnecessary considering that the text entry is constrained to accepting only numerals. You can just do
```
port = atoi(port_text) & 0xffff; // or % 65536 
```

jralls (changes requested):

gnucash/gnome-utils/dialog-file-access.c
jlaustill: Fixed. Replaced `g_ascii_strtol` (which doesn't exist in GLib) with `atoi` as suggested in comment #3127424486.

gnucash/gnome-utils/dialog-file-access.c
jlaustill: Fixed. Simplified to `port = atoi(port_text) & 0xffff;` and removed the now-unnecessary `port_endptr` and `parsed_port` variables.

jralls (approved):
Looks good, now squash it back to a single commit for merging.

jralls:
Thanks!

PR #2203, opened 2026-04-13
Python binding: example and test to create new book

quazgar:
Added a few more examples and one test to Python, regarding new book creation.

Especially noted that session.book.get_root_account() is necessary to mark the book as dirty, otherwise the tables will not be created.

### DIFF
--- a/bindings/python/example_scripts/simple_book.py
+++ b/bindings/python/example_scripts/simple_book.py
@@ -9,14 +9,12 @@ from gnucash import Session, SessionOpenMode
 
 # We need to tell GnuCash the data format to create the new file as (xml://)
 uri = "xml:///tmp/simple_book.gnucash"
-# Alternatively, use "sqlite3://..." instead to save as sqlite file.
 
 print("uri:", uri)
 with Session(uri, SessionOpenMode.SESSION_NEW_STORE) as ses:
     book = ses.get_book()
 
-    # Call some methods that produce output to show that Book works and to mark it as "something
-    # needs to be saved".
+    # Create root account to mark the book as "needs saving".
     book.get_root_account().SetDescription("hello, book")
     print("Book is saved:", not book.session_not_saved())
 
--- a/bindings/python/example_scripts/simple_book.py
+++ b/bindings/python/example_scripts/simple_book.py
@@ -9,17 +9,19 @@ from gnucash import Session, SessionOpenMode
 
 # We need to tell GnuCash the data format to create the new file as (xml://)
 uri = "xml:///tmp/simple_book.gnucash"
+# Alternatively, use "sqlite3://..." instead to save as sqlite file.
 
 print("uri:", uri)
 with Session(uri, SessionOpenMode.SESSION_NEW_STORE) as ses:
     book = ses.get_book()
 
-    #Call some methods that produce output to show that Book works
+    # Call some methods that produce output to show that Book works and to mark it as "something
+    # needs to be saved".
     book.get_root_account().SetDescription("hello, book")
     print("Book is saved:", not book.session_not_saved())
 
-    #As long as there's no exceptions, book is automatically saved
-    #when session ends.
+    # As long as there's no exception, book is automatically saved
+    # when session ends.
     print("saving...")
 
 print("Book is saved:", not book.session_not_saved())
--- a/bindings/python/example_scripts/simple_sqlite_create.py
+++ b/bindings/python/example_scripts/simple_sqlite_create.py
@@ -3,12 +3,13 @@
 #   @brief Example Script simple sqlite create 
 #   @ingroup python_bindings_examples
 
+import os
 from gnucash import Session, Account, SessionOpenMode
 from os.path import abspath
 from gnucash.gnucash_core_c import ACCT_TYPE_ASSET
 
 s = Session('sqlite3://%s' % abspath('test.blob'), SessionOpenMode.SESSION_NEW_STORE)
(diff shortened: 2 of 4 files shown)


quazgar:
Especially the `session.book.get_root_account()` is a new requirement in GnuCash 5, it was not necessary in GnuCash 4, as I could verify in Docker images of Debian 12 vs. Debian 13.

jralls:
> Especially noted that session.book.get_root_account() is necessary to mark the book as dirty, otherwise the tables will not be created.

No, it's one way (and probably the minimal way) to get the book to contain something, and if the book contains nothing then there's nothing to save. That was indeed changed in GnuCash 5 to stop GnuCash asking if you wanted to save after starting it with no file and using File>New, File>Open, or Quit.

quazgar:
Ah, I understand! Unfortunately, this would lead the file in a locked state then, so it could not be opened again without forcibly breaking the lock. So for working with Python, this is the only way I could find out.

bindings/python/example_scripts/simple_book.py
```diff
@@ -9,17 +9,19 @@
 
 # We need to tell GnuCash the data format to create the new file as (xml://)
 uri = "xml:///tmp/simple_book.gnucash"
+# Alternatively, use "sqlite3://..." instead to save as sqlite file.
```
jralls, line 12: Not a useful comment. You added the sqlite3 test in simple_sqlite_create.py.

bindings/python/example_scripts/simple_book.py
```diff
@@ -9,17 +9,19 @@
...
 
 print("uri:", uri)
 with Session(uri, SessionOpenMode.SESSION_NEW_STORE) as ses:
     book = ses.get_book()
 
-    #Call some methods that produce output to show that Book works
+    # Call some methods that produce output to show that Book works and to mark it as "something
+    # needs to be saved".
```
jralls, line 19: Better wording: "Create a root account so that the book has something to save."

bindings/python/example_scripts/simple_sqlite_create.py
```diff
@@ -3,12 +3,13 @@
...
 #   @ingroup python_bindings_examples
 
+import os
 from gnucash import Session, Account, SessionOpenMode
 from os.path import abspath
 from gnucash.gnucash_core_c import ACCT_TYPE_ASSET
 
 s = Session('sqlite3://%s' % abspath('test.blob'), SessionOpenMode.SESSION_NEW_STORE)
-# this seems to make a difference in more complex cases
+# This seems to make a difference in more complex cases
```
jralls, line 12: If you're going to make it a sentence put a period at the end.

bindings/python/example_scripts/simple_sqlite_create.py
```diff
@@ -23,3 +24,21 @@
...
+
+########################################
+# Similar, but create empty file first.
+
+targetpath = os.path.abspath("./test.gnucash")
+targetfile = f"sqlite3://{targetpath}"
+
+# Remove if it existst.
+if os.path.exists(targetpath):
+    os.remove(targetpath)
```
jralls, line 36: Better to use `TemporaryDirectory()` so that it doesn't.
That aside, comments that explain blindingly obvious code are unwelcome noise.

jralls (changes requested):

jralls:
> Unfortunately, this would lead the file in a locked state then, so it could not be opened again without forcibly breaking the lock. So for working with Python, this is the only way I could find out.

IIUC that sounds like a bug in the SQL backend: Creating the session creates a database (a SQLite3 file in this case) with a lock table and the lock set. When the session ends with no other content it bails out without clearing the lock.

quazgar:
> IIUC that sounds like a bug in the SQL backend

Yes, that's exactly the result I am seeing. Would a test, for example in test_book.py, be appreciated?

bindings/python/example_scripts/simple_sqlite_create.py
quazgar: Thanks for your comments, I reworked my PR, I hope the result should match your suggestions.

bindings/python/example_scripts/simple_book.py
quazgar: New wording: 
> Create root account to mark the book as "needs saving".

bindings/python/example_scripts/simple_sqlite_create.py
quazgar: New wording:

> The save() seems to make a difference in more complex cases.

quazgar:
> > IIUC that sounds like a bug in the SQL backend
> 
> Yes, that's exactly the result I am seeing. Would a test, for example in test_book.py, be appreciated?

Anyway, here's the test: [e2bad35](https://github.com/Gnucash/gnucash/pull/2203/commits/e2bad35d9b8a04d293d870e24aa9bb377106dd2b)

jralls:
> Anyway, here's the test:

Thanks, I'll give it a spin.

jralls:
The locking problem turned out to be that `Session.__exit__` didn't destroy the session, it just ended it. On the C++ side it's the QofSession destructor that calls the DBI backend's unlock. Of course it also was leaking the QofSession and QofBook objects. I've pushed a commit that fixes that.

So you can remove  `get_root_account()` with its comment from `test_minimal_new_sqlite` after rebasing on stable. While rebasing, squash back down to a single commit leaving off the `TestBookSqlite` commit.

PR #2205, opened 2026-04-16
Bug 799314 - Add Run command to Schedule Transactions

agwekixj:
This patch adds a “Run” command to the “Scheduled Transactions” tab to allow users to select the scheduled transactions to be included in a “Since Last Run…” window.   If there are no instances of a selected transaction triggered by today’s date, the next instance is triggered.

### DIFF
--- a/gnucash/gnome/gnc-plugin-page-sx-list.cpp
+++ b/gnucash/gnome/gnc-plugin-page-sx-list.cpp
@@ -53,6 +53,7 @@
 #include "Transaction.h"
 #include "dialog-sx-editor.h"
 #include "dialog-utils.h"
+#include "dialog-sx-since-last-run.h"
 #include "gnc-commodity.h"
 #include "gnc-component-manager.h"
 #include "gnc-date.h"
@@ -124,6 +125,7 @@ static void gppsl_row_activated_cb (GtkTreeView *tree_view, GtkTreePath *path,
 static void gnc_plugin_page_sx_list_cmd_new (GSimpleAction *simple, GVariant *paramter, gpointer user_data);
 static void gnc_plugin_page_sx_list_cmd_edit (GSimpleAction *simple, GVariant *paramter, gpointer user_data);
 static void gnc_plugin_page_sx_list_cmd_delete (GSimpleAction *simple, GVariant *paramter, gpointer user_data);
+static void gnc_plugin_page_sx_list_cmd_run (GSimpleAction *simple, GVariant *paramter, gpointer user_data);
 static void gnc_plugin_page_sx_list_cmd_refresh (GSimpleAction *simple, GVariant *paramter, gpointer user_data);
 static void gnc_plugin_page_sx_list_cmd_save_layout (GSimpleAction *simple, GVariant *paramter, gpointer user_data);
 static void gnc_plugin_page_sx_list_cmd_edit_tax_options (GSimpleAction *simple, GVariant *paramter, gpointer user_data);
@@ -135,6 +137,7 @@ static GActionEntry gnc_plugin_page_sx_list_actions [] =
     { "SxListNewAction", gnc_plugin_page_sx_list_cmd_new, nullptr, nullptr, nullptr },
     { "SxListEditAction", gnc_plugin_page_sx_list_cmd_edit, nullptr, nullptr, nullptr },
     { "SxListDeleteAction", gnc_plugin_page_sx_list_cmd_delete, nullptr, nullptr, nullptr },
+    { "SxListRunAction", gnc_plugin_page_sx_list_cmd_run, nullptr, nullptr, nullptr },
     { "ViewRefreshAction", gnc_plugin_page_sx_list_cmd_refresh, nullptr, nullptr, nullptr },
     { "ViewSaveLayoutAction", gnc_plugin_page_sx_list_cmd_save_layout, nullptr, nullptr, nullptr },
     { "EditTaxOptionsAction", gnc_plugin_page_sx_list_cmd_edit_tax_options, nullptr, nullptr, nullptr },
@@ -315,17 +318,19 @@ static void
 gppsl_selection_changed_cb (GtkTreeSelection *selection, gpointer user_data)
 {
     GncPluginPage *page;
-    GAction *edit_action, *delete_action;
+    GAction *edit_action, *delete_action, *run_action;
     gboolean selection_state = true;
 
     page = GNC_PLUGIN_PAGE(user_data);
     edit_action = gnc_plugin_page_get_action (page, "SxListEditAction");
     delete_action = gnc_plugin_page_get_action (page, "SxListDeleteAction");
+    run_action = gnc_plugin_page_get_action (page, "SxListRunAction");
     selection_state = gtk_tree_selection_count_selected_rows (selection) == 0
                       ? false
                       : true;
     g_simple_action_set_enabled (G_SIMPLE_ACTION(edit_action), selection_state);
     g_simple_action_set_enabled (G_SIMPLE_ACTION(delete_action), selection_state);
+    g_simple_action_set_enabled (G_SIMPLE_ACTION(run_action), selection_state);
 }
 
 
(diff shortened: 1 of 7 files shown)


gnucash/gnome/gnc-plugin-page-sx-list.cpp
```diff
@@ -951,5 +964,80 @@ gnc_plugin_page_sx_list_cmd_delete (GSimpleAction *simple,
...
+gnc_plugin_page_sx_list_cmd_run (GSimpleAction *simple,
+                                    GVariant      *parameter,
+                                    gpointer       user_data)
+{
+    auto plugin_page = GNC_PLUGIN_PAGE_SX_LIST(user_data);
+    GncPluginPageSxListPrivate *priv = GNC_PLUGIN_PAGE_SX_LIST_GET_PRIVATE(plugin_page);
+    GtkTreeSelection *selection = gtk_tree_view_get_selection (priv->tree_view);
+    GtkTreeModel *model;
+    GList *selected_paths = gtk_tree_selection_get_selected_rows (selection, &model);
+
```
jralls, line 977: All of these except `model` can be `auto`, though arguably `selected_paths` should be type-specified for clarity since it isn't entirely obvious what `gtk_tree_selection_get_selected_rows` returns.

gnucash/gnome/gnc-plugin-page-sx-list.cpp
```diff
@@ -951,5 +964,80 @@ gnc_plugin_page_sx_list_cmd_delete (GSimpleAction *simple,
...
+    if (!gnc_list_length_cmp (selected_paths, 0))
+    {
+        g_warning ("no selection for run.");
+        return;
```
jralls: Empty `GList*` are `NULL` so the fastest way to test for being empty is `if (!selected_paths)`.
Please use the QofLog macros, e.g. `PWARN`, instead of using the GLib message functions directly.

libgnucash/app-utils/gnc-sx-instance-model.c
```diff
@@ -555,6 +555,41 @@ _gnc_sx_gen_instances(gpointer *data, gpointer user_data)
     return instances;
 }
 
+static GncSxInstances*
+_gnc_sx_gen_select_instances(gpointer *data, gpointer user_data)
```
jralls: Per the [C Standard](https://www.open-std.org/jtc1/sc22/wg14/www/docs/N2912.pdf) p71: "All identifiers that begin with an underscore are reserved for use as identifiers with file scope in both the ordinary and tag name spaces." and "If the program declares or defines an identifier in a context in which it is reserved (other than as allowed by ??), the behavior is undefined."

`static` function names are at file scope. I'm aware that there are other functions in this file that violate this rule.

A more legal way to distinguish `static` function names from global ones is to leave off the `gnc` prefix.

jralls (changes requested):
If the sx's first recurrence is after today the SLR will contain a single instance of that future recurrence; otherwise the last recurrence is the one previous to today. Is that intentional?

gnucash/gnome/gnc-plugin-page-sx-list.cpp
agwekixj: True.  However, mimicking the existing code style improves readability.  See the other static gnc_plugin_page_sx_list_cmd_... () code.

gnucash/gnome/gnc-plugin-page-sx-list.cpp
agwekixj: Again,  mimicking the existing code.

libgnucash/app-utils/gnc-sx-instance-model.c
agwekixj: And again, mimicking.

agwekixj:
> If the sx's first recurrence is after today the SLR will contain a single instance of that future recurrence; otherwise the last recurrence is the one previous to today. Is that intentional?

Yes.  It is intentional.  It is what I meant by:  "If there are no instances of a selected transaction triggered by today’s date, the next instance is triggered. "  It is the enhancement being requested.

Thank you for taking the time to review the code.  Regarding your requested changes, are you expecting these modifications to be limited to the new code or are we attempting to "clean" the files?

jralls:
> are you expecting these modifications to be limited to the new code or are we attempting to "clean" the files

Fixing the whole thing would be nice but I'll be satisfied with just fixing the new code.

gnucash/gnome/gnc-plugin-page-sx-list.cpp
agwekixj: Done.

gnucash/gnome/gnc-plugin-page-sx-list.cpp
agwekixj: Done.

libgnucash/app-utils/gnc-sx-instance-model.c
agwekixj: Done.

agwekixj:
> > are you expecting these modifications to be limited to the new code or are we attempting to "clean" the files
> 
> Fixing the whole thing would be nice but I'll be satisfied with just fixing the new code.

I've applied your comments to the new code.  Let me know if they're satisfactory and I'll squash and force-push.

gnucash/ui/gnc-plugin-page-sx-list.ui
```diff
@@ -201,6 +207,21 @@
...
     </child>
+    <child>
+      <object class="GtkToolButton" id="but6">
+        <property name="visible">True</property>
+        <property name="can-focus">False</property>
+        <property name="label" translatable="yes">_Run</property>
+        <property name="action-name">GncPluginPageSxListActions.SxListRunAction</property>
+        <property name="tooltip-text" translatable="yes">Run the selected scheduled transaction</property>
+        <property name="use-underline">True</property>
+        <property name="icon-name">document-send</property>
```
Bob-IT: I am not sure about the icon, maybe "system-run"

gnucash/gnome/gnc-plugin-page-sx-list.cpp
```diff
@@ -951,5 +964,80 @@ gnc_plugin_page_sx_list_cmd_delete (GSimpleAction *simple,
...
+        g_list_free (auto_created_txns);
+    }
+
+    g_object_unref (G_OBJECT(sx_instances));
+
+    if (creation_errors)
+    {
+        gnc_ui_sx_creation_error_dialog (&creation_errors);
+    }
+
```
Bob-IT: This looks like a copy from dialog-sx-since-last-run.c
Can you not create a new function there with this part and call it from both locations?

Bob-IT:
> "If there are no instances of a selected transaction triggered by today’s date, the next instance is triggered.

Could this be used in the tool tip in some form?

gnucash/ui/gnc-plugin-page-sx-list.ui
agwekixj: I went back and forth on the two icons.  Are you okay with the location?

gnucash/gnome/gnc-plugin-page-sx-list.cpp
agwekixj: I had actually used code from gnc-plugin-basic-commands.c.

I've reimplemented the gnc_ui_sx_since_last_run_dialog() abstraction to accomodate the code common to gnc_sx_sxsincelast_book_opened(), gnc_main_window_cmd_actions_since_last_run(), and gnc_plugin_page_sx_list_cmd_run().

Good catch.

agwekixj:
> > "If there are no instances of a selected transaction triggered by today’s date, the next instance is triggered.
> 
> Could this be used in the tool tip in some form?

Would this suffice?

_Run the selected scheduled transaction.  If no transactions are scheduled to be entered, the next occurrence is used._

agwekixj:
I intend to simplify the modified gnc_ui_sx_since_last_run_dialog() abstraction.  Please hold off on the merge or any review in this area of code change until the next push.  Thanks.

jralls (approved):
Looks OK to me.

Bob-IT (approved):
Changes are OK with me, I think just a squash required.

Just an after thought, why did you allow it to work on non enabled schedules?

agwekixj:
> Changes are OK with me, I think just a squash required.
> 
> Just an after thought, why did you allow it to work on non enabled schedules?

If a users includes a non enabled schedules in their selection and issues any other option, we allow it to work.  I didn't see a reason not to be consistent.  I think of enabled like hidden on the Accounts.  I even considered hiding the non enabled schedules when the Enabled column wasn't viewed.

Again, thanks for taking the time.

agwekixj:
Ready for merge.

jralls:
Thanks!

PR #2206, opened 2026-04-23
[gnucash.css] the register entry selection needs to be set

christopherlam:
because on dark mode it'll be light on light green with poor contrast. set it to dark forest green on light green.

before:
Screenshot_2026-04-23_21-52-49

after:
image

alternative: dark charcoal #2c2c2c
image

### DIFF
--- a/gnucash/gnucash.css
+++ b/gnucash/gnucash.css
@@ -18,6 +18,11 @@ gnc-id-cursor entry {
   padding: 0px 2px 0px 2px;
 }
 
+/* Selection text in entry */
+gnc-id-cursor entry selection {
+  color: #1b4332;
+}
+
 gnc-id-cursor button {
   margin: 1px 1px 1px 1px;
 }


Bob-IT:
I will have a look, I would think I would of noticed that.
What theme are you using ?

christopherlam:
> I will have a look, I would think I would of noticed that. What theme are you using ?

I use pop os cosmic

Bob-IT:
@christopherlam
As there are many themes we can not support all so the only two I test against are `Adwaita` and `Adwaita-dark`. The following screenshot is from `Adwaita-dark` and as you can see it is OK...

css

I did some digging and found there was an ebuild for the cosmic desktop and downloaded/installed the theme which was is called `adw-gtk3` and `adw-gtk3-dark` and I did get the first images you sent above. Looking at both theme gtk.css files line 194, if you change the alpha value from 0.25 to 0.8 you will get a similar image to mine or add your CSS locally.

So I do not think it appropriate to make the change you are suggesting.

gnucash/gnucash.css
```diff
@@ -18,6 +18,11 @@ gnc-id-cursor entry {
   padding: 0px 2px 0px 2px;
 }
 
+/* Selection text in entry */
+gnc-id-cursor entry selection {
+  color: #1b4332;
+}
+
```
Bob-IT, line 25: See comment.

christopherlam:
> @christopherlam As there are many themes we can not support all so the only two I test against are `Adwaita` and `Adwaita-dark`. The following screenshot is from `Adwaita-dark` and as you can see it is OK...
>
> I did some digging and found there was an ebuild for the cosmic desktop and downloaded/installed the theme which was is called `adw-gtk3` and `adw-gtk3-dark` and I did get the first images you sent above. Looking at both theme gtk.css files line 194, if you change the alpha value from 0.25 to 0.8 you will get a similar image to mine or add your CSS locally.
> 
> So I do not think it appropriate to make the change you are suggesting.

Ok I am not at all familiar with cosmic internals except that it's now the default for pop_OS. Did you mean modify the `~/.config/gtk-3.0/gtk.css` or `~/.config/gtk-4.0/gtk.css` files? I don't know the exact line to modify.

[gtk.css](https://github.com/user-attachments/files/27095082/gtk.css)

Bob-IT:
> Ok I am not at all familiar with cosmic internals except that it's now the default for pop_OS. Did you mean modify the `~/.config/gtk-3.0/gtk.css` or `~/.config/gtk-4.0/gtk.css` files? I don't know the exact line to modify.

No, for the dark theme, on my Gentoo box I was looking at, `/usr/share/themes/adw-gtk3-dark/gtk.css` and line 194 points to the selected background color for various widgets. It uses the `accent_bg_color` with an alpha value of 0.25, I changed that to 0.8. Further up the file you will see the color defined as blue_3 and further up the actual color. The CSS you provided above just defines all/some of these color values which it says is generated by cosmic so I assume there is a desktop application that allows you to override the default ones. By using the Gtk Inspector and looking at the CSS nodes for that entry widget you can see the location in the CSS file that `background-color` is specified.

As I said, whether you want to change that or just use a local CSS change is up to you.

jralls:
Why on earth would you set alpha to anything other than 1 on a text color? 

Anyway, I agree with @Bob-IT. This is a theme being stupid and overriding theme choices in GnuCash just breaks things for all of the themes that aren't broken.

PR #2207, opened 2026-04-23
Convert gnc-register-gnome to C++

russ-gorby:


gnucash/register/register-core/table-allgui.h
```diff
@@ -192,6 +192,29 @@ typedef enum {
...
+{
+#endif
+
+/*
+ * These 4 functions are not defined in the source file for table-allgui
+ * from register-core but rather in the source file for table-gnome from register-gnome
+ */
+void  gnc_table_save_state (Table *table, const gchar *state_section); // RRG: table-gnome.cpp
+
+void        gnc_table_init_gui (Table *table); // RRG: table-gnome.cpp
```
jralls, line 206: Excess space between the return type and function name.

gnucash/register/register-gnome/combocell-gnome.cpp
```diff
@@ -108,7 +108,7 @@ gnc_combo_cell_autopop_init (gpointer unused)
 
     id = gnc_prefs_register_cb (GNC_PREFS_GROUP_GENERAL_REGISTER,
                                 GNC_PREF_AUTO_RAISE_LISTS,
-                                gnc_combo_cell_set_autopop,
+                                reinterpret_cast<gpointer>(gnc_combo_cell_set_autopop),
```
jralls, line 112: Is `static_cast<gpointer>` raising an error? It's [usually OK](https://stackoverflow.com/a/310489) to `static_cast` to/from `void*`.

gnucash/register/register-gnome/combocell-gnome.cpp
```diff
@@ -526,18 +516,21 @@ gnc_combo_cell_type_ahead_search (const gchar* newval,
...
-    PopBox* box = cell->cell.gui_private;
+    auto box = static_cast<PopBox *>(cell->cell.gui_private);
     int num_found = 0;
     gchar* match_str = NULL;
     const char* sep = gnc_get_account_separator_string ();
     char* escaped_sep = g_regex_escape_string (sep, -1);
     char* escaped_newval = g_regex_escape_string (newval, -1);
     gchar* newval_rep = g_strdup_printf (".*%s.*", escaped_sep);
-    GRegex* regex0 = g_regex_new (escaped_sep, 0, 0, NULL);
+    const GRegexMatchFlags flag_0 = static_cast<GRegexMatchFlags>(0);
```
jralls, line 528: You want `constexpr` here, not `const`.

gnucash/register/register-gnome/completioncell-gnome.cpp
```diff
@@ -631,7 +632,8 @@ select_first_entry_in_list (PopBox* box)
 static void
 populate_list_store (CompletionCell* cell, gchar* str)
 {
-    PopBox* box = cell->cell.gui_private;
+    auto box = static_cast<PopBox *>(cell->cell.gui_private);
+    char dont_text[] = DONT_TEXT;
```
jralls, line 635: Why `char[]` instead of `const char*` (should be `static` or better `constexpr` either way so it doesn't get regenerated every time the function is called) or better yet changing `DONT_TEXT` from a macro to `static const char*` at file scope?

gnucash/register/register-gnome/gnucash-color.cpp
```diff
@@ -73,19 +74,17 @@ color_equal (gconstpointer v, gconstpointer w)
...
 
-    color = g_hash_table_lookup (color_hash_table, &key);
+    auto color = static_cast<GdkRGBA *>(g_hash_table_lookup (color_hash_table, &key));
 
     if (color)
         return color;
 
     color = g_new0(GdkRGBA, 1);
-    newkey = g_new0(guint32, 1);
+    std::uint32_t *newkey = g_new0(guint32, 1);
```
jralls, line 87: The stdint types don't need to be prefixed with `std::`, `cstdint` includes `stdint.h` that puts them in the global namespace and aliases them into `namespace std` for the standard library's use.

gnucash/register/register-gnome/gnucash-sheet.cpp
```diff
@@ -84,8 +84,11 @@ static void gnucash_sheet_activate_cursor_cell (GnucashSheet *sheet,
 static void gnucash_sheet_stop_editing (GnucashSheet *sheet);
 static gboolean gnucash_sheet_check_direct_update_cell (GnucashSheet *sheet,
                                                         const VirtualLocation virt_loc);
+extern "C" {
+// not accessed externally, should this be static?
```
jralls, line 88: Yes. Callbacks should nearly always be static, and generally shouldn't have the `gnc` or in this case `gnucash` prefix to make it more obvious that they're private.

gnucash/register/register-gnome/combocell-gnome.cpp
```diff
@@ -459,6 +456,13 @@ gnc_combo_cell_add_account_menu_item (ComboCell* cell, char* menustr)
         return;
 
     auto box = static_cast<PopBox *>(cell->cell.gui_private);
+    auto strdelimt = [](std::string& s, const std::string_view& delims, char repl) {
+        for (char& ch : s) {
+            if (delims.find(ch) != std::string_view::npos) {
+                ch = repl;
+            }
+        }
+    };
```
jralls, line 465: Put the lambda in the scope where it's used.

gnucash/register/register-gnome/datecell-gnome.cpp
```diff
@@ -131,9 +130,8 @@ check_readonly_threshold (const gchar *datestr, GDate *d, gboolean warn)
 static void
 gnc_parse_date (struct tm *parsed, const char * datestr, gboolean warn)
 {
-    int day, month, year;
-    gboolean use_autoreadonly = qof_book_uses_autoreadonly(gnc_get_current_book());
-    GDate *test_date;
+    int day = 0, month = 0, year = 0;
+    bool use_autoreadonly = qof_book_uses_autoreadonly(gnc_get_current_book());
```
jralls, line 134: `auto` or better yet, lose the variable and change line 163 to
```
if (qof_book_uses_autoreadonly(gnc_get_current_book()))
```

gnucash/register/register-gnome/datecell-gnome.cpp
```diff
@@ -150,7 +148,7 @@ gnc_parse_date (struct tm *parsed, const char * datestr, gboolean warn)
         year = tm_today.tm_year + 1900;
     }
 
-    test_date = g_date_new_dmy (day, static_cast<GDateMonth>(month), year);
+    GDate *test_date = g_date_new_dmy (day, static_cast<GDateMonth>(month), year);
```
jralls, line 151: `auto`

gnucash/register/register-gnome/datecell-gnome.cpp
```diff
@@ -217,19 +212,21 @@ gnc_date_cell_init (DateCell *cell)
     cell->cell.direct_update = gnc_date_cell_direct_update;
     cell->cell.set_value = gnc_date_cell_set_value_internal;
 
-    box = g_new0 (PopBox, 1);
+    PopBox *box = g_new0 (PopBox, 1);
```
jralls, line 215: `auto box = static_cast<PopBox*>(g_new0 (PopBox, 1));` to be consistent with your dozens of the same invocation in earlier commits.

gnucash/register/register-gnome/formulacell-gnome.cpp
```diff
@@ -56,15 +56,13 @@ gnc_formula_cell_direct_update( BasicCell *bcell,
...
     auto cell = reinterpret_cast<FormulaCell *>(bcell);
     auto event = static_cast<GdkEventKey *>(gui_data);
-    struct lconv *lc;
-    gboolean is_return;
 
     if (event->type != GDK_KEY_PRESS)
         return FALSE;
 
-    lc = gnc_localeconv ();
+    struct lconv *lc = gnc_localeconv ();
```
jralls, line 63: `auto`

gnucash/register/register-gnome/formulacell-gnome.cpp
```diff
@@ -112,9 +110,7 @@ gnc_formula_cell_direct_update( BasicCell *bcell,
 BasicCell *
 gnc_formula_cell_gnome_new (void)
 {
-    BasicCell *cell;
-
-    cell = gnc_formula_cell_new ();
+    BasicCell *cell = gnc_formula_cell_new ();
```
jralls, line 113: `auto`

gnucash/register/register-gnome/gnucash-cursor.cpp
```diff
@@ -48,49 +48,42 @@ G_DEFINE_TYPE (GnucashCursor, gnucash_cursor, G_TYPE_OBJECT)
...
     vcell_loc.virt_row = cursor->row;
     vcell_loc.virt_col = cursor->col;
 
-    block = gnucash_sheet_get_block (sheet, vcell_loc);
+    SheetBlock *block = gnucash_sheet_get_block (sheet, vcell_loc);
     if (!block)
         return;
 
-    vcell = gnc_table_get_virtual_cell (sheet->table, vcell_loc);
+    VirtualCell *vcell = gnc_table_get_virtual_cell (sheet->table, vcell_loc);
```
jralls, line 65: `auto`

gnucash/register/register-gnome/gnucash-cursor.cpp
```diff
@@ -48,49 +48,42 @@ G_DEFINE_TYPE (GnucashCursor, gnucash_cursor, G_TYPE_OBJECT)
...
-    SheetBlock *block;
-    gint col;
+    VirtualCellLocation vcell_loc{};
+    int col = 0;
 
     vcell_loc.virt_row = cursor->row;
     vcell_loc.virt_col = cursor->col;
 
-    block = gnucash_sheet_get_block (sheet, vcell_loc);
+    SheetBlock *block = gnucash_sheet_get_block (sheet, vcell_loc);
```
jralls, line 61: `auto`

gnucash/register/register-gnome/gnucash-header.cpp
```diff
@@ -106,36 +102,31 @@ gnc_header_draw_offscreen (GncHeader *header)
...
-            PangoLayout *layout;
+            int width = 0;
             PangoRectangle logical_rect;
             GdkRectangle rect;
-            int x_offset;
 
             virt_loc.phys_col_offset = j;
 
-            cd = gnucash_style_get_cell_dimensions (style, i, j);
+            CellDimensions *cd = gnucash_style_get_cell_dimensions (style, i, j);
```
jralls, line 129: `auto`

gnucash/register/register-gnome/gnucash-header.cpp
```diff
@@ -144,7 +135,7 @@ gnc_header_draw_offscreen (GncHeader *header)
             else
                 width = cd->pixel_width;
 
-            cell = gnc_cellblock_get_cell (cb, i, j);
+            BasicCell *cell = gnc_cellblock_get_cell (cb, i, j);
```
jralls, line 138: `auto`

gnucash/register/register-gnome/gnucash-header.cpp
```diff
@@ -157,11 +148,11 @@ gnc_header_draw_offscreen (GncHeader *header)
 
             virt_loc.vcell_loc =
                 table->current_cursor_loc.vcell_loc;
-            text = gnc_table_get_label (table, virt_loc);
+            const char *text = gnc_table_get_label (table, virt_loc);
```
jralls, line 151: `auto`

gnucash/register/register-gnome/gnucash-item-edit.cpp
```diff
@@ -551,30 +534,29 @@ draw_text_cursor_cb (GtkWidget *widget, cairo_t *cr, gpointer user_data)
...
 
     // Get the layout x offset
     gtk_entry_get_layout_offsets (GTK_ENTRY(widget), &x_offset, nullptr);
 
     // Get the foreground color
+    GdkRGBA color;
     gdk_rgba_parse (&color, "black");
     gtk_style_context_get_color (stylectxt, flags, &color);
-    fg_color = &color;
+    GdkRGBA *fg_color = &color;
```
jralls, line 550: `auto`

gnucash/register/register-gnome/gnucash-item-list.cpp
```diff
@@ -121,11 +118,12 @@ _gnc_item_find_selection (GtkTreeModel* model, GtkTreePath* path,
...
 {
     auto to_find = static_cast<FindSelectionData *>(data);
-    gchar* iterStr;
-    gboolean found;
 
+    char* iterStr;
     gtk_tree_model_get (model, iter, 0, &iterStr, -1);
-    found = g_strcmp0 (to_find->string_to_find, iterStr) == 0;
+
+    bool found = g_strcmp0 (to_find->string_to_find, iterStr) == 0;
```
jralls, line 125: `auto`

gnucash/register/register-gnome/gnucash-item-list.cpp
```diff
@@ -138,21 +136,18 @@ _gnc_item_find_selection (GtkTreeModel* model, GtkTreePath* path,
...
+    FindSelectionData *to_find_data = static_cast<FindSelectionData *>(g_new0 (FindSelectionData, 1));
     to_find_data->item_list = item_list;
     to_find_data->string_to_find = string;
 
     gtk_tree_model_foreach (GTK_TREE_MODEL (item_list->list_store),
                             _gnc_item_find_selection,
                             to_find_data);
 
-    result = (to_find_data->found_path != nullptr);
+    gboolean result = (to_find_data->found_path != nullptr);
```
jralls, line 150: `auto`, and that's going to be a `bool` (result of `operator !=`) not a `gboolean`.

gnucash/register/register-gnome/gnucash-register.cpp
```diff
@@ -364,9 +339,7 @@ gnucash_register_update_hadjustment (GtkAdjustment *adj,
 static void
 gnucash_register_class_init (GnucashRegisterClass *klass)
 {
-    GObjectClass *gobject_class;
-
-    gobject_class = G_OBJECT_CLASS (klass);
+    GObjectClass *gobject_class = G_OBJECT_CLASS (klass);
```
jralls, line 342: `auto`

gnucash/register/register-gnome/gnucash-sheet-private.cpp
```diff
@@ -57,15 +58,14 @@ find_block_by_pixel (GnucashSheet *sheet,
                      gint x, gint y,
                      VirtualCellLocation *vcell_loc)
 {
-    SheetBlock *block;
     VirtualCellLocation vc_loc = { 1, 0 };
 
     g_return_val_if_fail (y >= 0, nullptr);
     g_return_val_if_fail (x >= 0, nullptr);
 
+    SheetBlock *block = gnucash_sheet_get_block (sheet, vc_loc);
```
jralls, line 66: `auto`

gnucash/register/register-gnome/gnucash-sheet-private.cpp
```diff
@@ -539,7 +528,10 @@ draw_cell (GnucashSheet *sheet, SheetBlock *block,
     {
         text = gnc_table_get_label (table, virt_loc);
         if ((text == nullptr) || (*text == '\0'))
-            goto exit;
+        {
+            do_exit();
+            return;
+        }
```
jralls, line 534: Nice!

gnucash/register/register-gnome/gnucash-sheet.cpp
```diff
@@ -177,23 +177,20 @@ gnucash_sheet_virt_cell_out_of_bounds (GnucashSheet *sheet,
...
+        SheetBlockStyle *style = gnucash_sheet_get_style (sheet, virt_loc.vcell_loc);
 
         valid = (virt_loc.phys_row_offset >= 0 &&
                  virt_loc.phys_row_offset < style->nrows &&
                  virt_loc.phys_col_offset >= 0 &&
                  virt_loc.phys_col_offset < style->ncols);
     }
 
-    return valid;
+    return valid ? TRUE : FALSE;
```
jralls, line 193: Not necessary, it's syntactic sugar all the way down. `true` is the same as `TRUE` is the same as `1`. If you really want to be pedantic just `return static_cast<gboolean>(valid);`.

gnucash/register/register-gnome/gnucash-sheet.cpp
```diff
@@ -1552,16 +1471,12 @@ gnucash_sheet_button_press_event (GtkWidget *widget, GdkEventButton *event)
     if (do_popup)
         gtk_menu_popup_at_pointer (GTK_MENU(sheet->popup), reinterpret_cast<GdkEvent *>(event));
 
-    return button_1 || do_popup;
+    return (button_1 || do_popup) ? TRUE : FALSE;
```
jralls, line 1474: Not necessary, the compiler can be trusted to do the right thing.

gnucash/register/register-gnome/gnucash-sheet.cpp
```diff
@@ -1572,12 +1487,12 @@ gnucash_sheet_refresh_from_prefs (GnucashSheet *sheet)
     sheet->use_vertical_lines = gnc_prefs_get_bool (GNC_PREFS_GROUP_GENERAL_REGISTER,
                                                     GNC_PREF_DRAW_VERT_LINES);
 
-    item_edit = GNC_ITEM_EDIT(sheet->item_editor);
+    GncItemEdit *item_edit = GNC_ITEM_EDIT(sheet->item_editor);
```
jralls, line 1490: `auto`

gnucash/register/register-gnome/gnucash-style.cpp
```diff
@@ -357,19 +330,19 @@ set_dimensions_pass_two (GnucashSheet *sheet, int default_width)
         }
     }
 
-    cursors = gnc_table_layout_get_cursors (sheet->table->layout);
+    GList *cursors = gnc_table_layout_get_cursors (sheet->table->layout);
```
jralls, line 333: `auto`

gnucash/register/register-gnome/gnucash-style.cpp
```diff
@@ -357,19 +330,19 @@ set_dimensions_pass_two (GnucashSheet *sheet, int default_width)
...
 
     /* adjust widths to be consistent */
-    for (node = cursors; node; node = node->next)
+    for (GList *node = cursors; node; node = node->next)
```
jralls, line 336: `auto`

gnucash/register/register-gnome/pricecell-gnome.cpp
```diff
@@ -130,9 +127,7 @@ gnc_price_cell_direct_update (BasicCell *bcell,
 BasicCell *
 gnc_price_cell_gnome_new (void)
 {
-    BasicCell *cell;
-
-    cell = gnc_price_cell_new ();
+    BasicCell *cell = gnc_price_cell_new ();
```
jralls, line 130: `auto`

gnucash/register/register-gnome/pricecell-gnome.cpp
```diff
@@ -146,15 +141,12 @@ gnc_basic_cell_insert_decimal(BasicCell *bcell,
                               int *start_selection,
                               int *end_selection)
 {
-    GString *newval_gs;
-    gint start, end;
-
     /* allocate space for newval_ptr : oldval + one letter ( the
        decimal_point ) */
-    newval_gs = g_string_new("");
+    GString *newval_gs = g_string_new("");
```
jralls, line 146: `auto`, but it would be better to replace it with `std::string`.

jralls (changes requested):
I got tired of sticking in `auto` comments: Nearly all of the local variable declarations should use it instead of an explicit type; the obvious exceptions being those where the variable is passed by reference to another function.

`bool` and `gboolean` and their values are interchangeable as far as the compiler is concerned. There's no need to cast from one to the other and no need for `return bool ? TRUE : FALSE;`.

jralls:
@russ-gorby Are you going to finish this up in time for 5.16?

russ-gorby:
I guess I'd like to push back a little on the auto-everywhere comment. I understand the AAA logic but when its capturing the result of a function call I think the explicit typing is more clear for those of us not as familiar with all the exact return types of all the calls. It means that I have to now lookup the callee to understand what type I'm working with.

gnucash/register/register-gnome/combocell-gnome.cpp
russ-gorby: hmm, let me check.

gnucash/register/register-core/table-allgui.h
russ-gorby: ooh, that's not good

gnucash/register/register-gnome/combocell-gnome.cpp
russ-gorby: K

gnucash/register/register-gnome/completioncell-gnome.cpp
russ-gorby: oh yeah, that's wasteful.

gnucash/register/register-gnome/gnucash-sheet.cpp
russ-gorby: I kind of disagree with this, and this pattern is ubiquitous throughout the code.
It makes the code less readable in my opinion.

jralls: It's ubiquitous because C doesn't have `auto`. How is it less readable when the cast specifies the return type?

jralls:
> I guess I'd like to push back a little on the auto-everywhere comment. I understand the AAA logic but when its capturing the result of a function call I think the explicit typing is more clear for those of us not as familiar with all the exact return types of all the calls. It means that I have to now lookup the callee to understand what type I'm working with.

Sometimes, maybe, and I'm OK with leaving the explicit decl when the type isn't obvious from the function name or a cast on the rhs of `=`. Only one case here, `gnc_table_layout_get_cursors (sheet->table->layout);`, comes close.

PR #2209, opened 2026-04-25
Bug 799762 - Poor handling of cases where hidden/placeholder accounts…

agwekixj:
… are used in the account register

If the cell value has not changed, gnc_combo_cell_leave() shouldn't reject it.


jralls:
Thanks!

gjanssens:
Quite a fast response, thanks @agwekixj !

PR #2210, opened 2026-04-25
Bug421610 add Logical dates to register View->Filter

Bob-IT:
This is my followup PR to change the register 'Filter By' dialog to include logical dates.
Chris raised Bug 798487 which was about not saving the 'Today', 'Earliest and 'Latest' setting of the filter properly and while looking at that found this enhancement bug which I thought would be better.

I have changed the 'Filter By' dialog to use GncPeriodSelect widgets as in the Preference dialog as shown below...

filter

`Note: the dates on the GncPeriodSelect lines are just for testing and will be disabled on final push.`

I have changed the radio buttons in the 'Select Range' to exclusive check buttons as I did not want to give the impression 'Earliest/Latest' were settings that could be saved.

The values saved for these period select widgets is some text combined with the enum of the accounting period value and these values are stored in the same positions as the date would of been in the saved filter text.

If a .gcm file is opened which has these values in it from an earlier Gnucash version, it would be treated as a date and fail to pass resulting in a value of 0 which is equivalent to no date filter and so the register will show all entries. I do not think it is worth guarding against this with a feature change.

The third commit I am unsure about and possibly could be dropped. I based setting up the GncPeriodSelect widgets from the preference dialog which is created from dialog-preferences.c line 1339. On lines 1410 to 1417 there is a check for the book value 'Fiscal year End' and supposedly used on lines 1424 and 1432 when setting up the GncPeriodSelect widgets but 'date_is_valid' always FALSE.

I can not find any place that would set this value, I created the KVP value in my test file manually just to test my code worked and that is how I found out that when used it would segfault and hence the following commit to fix that.
So maybe it was left over from some other change, I do not know, I can drop my commit and maybe remove the code from dialog-preferences.c but still push the last two commits.

### DIFF
--- a/gnucash/gnome/gnc-plugin-page-register-filter.cpp
+++ b/gnucash/gnome/gnc-plugin-page-register-filter.cpp
@@ -397,7 +397,7 @@ gpp_update_match_filter_text (cleared_match_t match, const guint mask,
  *
  *  @param page A pointer to the GncPluginPageRegister that is
  *  associated with this filter dialog.
- * 
+ *
  *  @param fd A pointer to the filter data used for filter state.
  */
 void
@@ -621,7 +621,7 @@ gnc_ppr_filter_update_date_query (GncPluginPage* plugin_page)
     LEAVE(" ");
 }
 
-/** This function is used to clear the current filter so that a 
+/** This function is used to clear the current filter so that a
  *  specific split can be shown in the register.
  *
  *  @param page A pointer to the GncPluginPageRegister that is
@@ -1113,7 +1113,7 @@ gnc_ppr_filter_response_cb (GtkDialog* dialog,
 /** This function is called to create the filter dialog.
  *
  *  @param rfd A pointer to the filter dialog structure.
- * 
+ *
  *  @param fd The filter data structure for remembering state.
  *
  *  @param query A pointer to the current register query.
@@ -1287,9 +1287,9 @@ gnc_ppr_filter_dialog_create (RegisterFilterDialog* rfd, FilterData *fd, Query *
  *
  *  @param plugin_page  A pointer to the GncPluginPageRegister that is
  *  associated with this filter dialog.
- * 
+ *
  *  @param query A pointer to the current register query.
- * 
+ *
  *  @param fd A pointer to the filter data structure for remembering state.
  *
  *  @param show_save_button Set to True to show save button.
--- a/gnucash/gnome/gnc-plugin-page-register-filter.cpp
+++ b/gnucash/gnome/gnc-plugin-page-register-filter.cpp
@@ -160,7 +160,7 @@ get_filter_default_num_of_days (GNCLedgerDisplayType ledger_type)
 
 /* This function converts a time64 value date to a string */
 static std::string
-gnc_ppr_filter_time2dmy (time64 raw_time)
+ppr_filter_time2dmy (time64 raw_time)
 {
     struct tm* timeinfo;
     char date_string[11];
@@ -175,7 +175,7 @@ gnc_ppr_filter_time2dmy (time64 raw_time)
 
 /* This function converts a string date to a time64 value */
 static time64
-gnc_ppr_filter_dmy2time (std::string date_string)
+ppr_filter_dmy2time (std::string date_string)
 {
     struct tm when;
 
@@ -206,8 +206,8 @@ split_filter_by_delimiter (std::string str, char delimiter)
 }
 
 static void
-gnc_ppr_check_for_empty_group (GKeyFile *state_file,
-                               const gchar *state_section)
+ppr_check_for_empty_group (GKeyFile *state_file,
+                           const gchar *state_section)
 {
     gsize num_keys;
     gchar **keys = g_key_file_get_keys (state_file, state_section, &num_keys, nullptr);
@@ -219,7 +219,7 @@ gnc_ppr_check_for_empty_group (GKeyFile *state_file,
 }
 
 static std::string
-gnc_ppr_filter_load_filter (GNCSplitReg *gsr, GNCLedgerDisplayType ledger_type)
+ppr_filter_load_filter (GNCSplitReg *gsr, GNCLedgerDisplayType ledger_type)
 {
     // get the filter from the .gcm file
     GKeyFile* state_file = gnc_state_get_current();
@@ -255,7 +255,7 @@ set_filterdata_to_defaults (FilterData *fd)
 }
 
 static void
-gnc_ppr_filter_load_filter_parts (GNCSplitReg *gsr, GNCLedgerDisplayType ledger_type, FilterData *fd)
+ppr_filter_load_filter_parts (GNCSplitReg *gsr, GNCLedgerDisplayType ledger_type, FilterData *fd)
 {
     set_filterdata_to_defaults (fd);
     fd->dialog = nullptr;
@@ -263,7 +263,7 @@ gnc_ppr_filter_load_filter_parts (GNCSplitReg *gsr, GNCLedgerDisplayType ledger_
     if (!gsr)
         return;
 
-    std::string filter_str = gnc_ppr_filter_load_filter (gsr, ledger_type);
+    std::string filter_str = ppr_filter_load_filter (gsr, ledger_type);
 
     PINFO("Loaded Filter String is %s", filter_str.c_str());
 
@@ -282,7 +282,7 @@ gnc_ppr_filter_load_filter_parts (GNCSplitReg *gsr, GNCLedgerDisplayType ledger_
     {
         PINFO("Loaded Filter Start Date is %s", split_filter[1].c_str());
 
-        fd->start_time = gnc_ppr_filter_dmy2time (split_filter[1]);
+        fd->start_time = ppr_filter_dmy2time (split_filter[1]);
         fd->start_time = gnc_time64_get_day_start (fd->start_time);
         fd->save_filter = true;
     }
@@ -291,7 +291,7 @@ gnc_ppr_filter_load_filter_parts (GNCSplitReg *gsr, GNCLedgerDisplayType ledger_
     {
         PINFO("Loaded Filter End Date is %s", split_filter[2].c_str());
 
-        fd->end_time = gnc_ppr_filter_dmy2time (split_filter[2]);
+        fd->end_time = ppr_filter_dmy2time (split_filter[2]);
         fd->end_time = gnc_time64_get_day_end (fd->end_time);
         fd->save_filter = true;
     }
@@ -310,7 +310,7 @@ gnc_ppr_filter_load_filter_parts (GNCSplitReg *gsr, GNCLedgerDisplayType ledger_
 }
 
 static void
-gnc_ppr_filter_save_filter (GNCSplitReg *gsr, std::string filter)
+ppr_filter_save_filter (GNCSplitReg *gsr, std::string filter)
 
 {
     GNCLedgerDisplayType ledger_type = gnc_ledger_display_type (gsr->ledger);
@@ -327,7 +327,7 @@ gnc_ppr_filter_save_filter (GNCSplitReg *gsr, std::string filter)
         if (g_key_file_has_key (state_file, state_section, KEY_PAGE_FILTER, nullptr))
             g_key_file_remove_key (state_file, state_section, KEY_PAGE_FILTER, nullptr);
 
-        gnc_ppr_check_for_empty_group (state_file, state_section);
+        ppr_check_for_empty_group (state_file, state_section);
     }
     else
     {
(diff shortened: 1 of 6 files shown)


jralls:
> So maybe it was left over from some other change, I do not know, I can drop my commit and maybe remove the code from dialog-preferences.c but still push the last two commits.

It's got a bit of history to it. The KVP key `fy_end` is my mistake from https://github.com/Gnucash/gnucash/commit/7faed6540162ccd1fe34ad9ea425164c3a61cdd6, before that there were two keys, `fyear_end/month` and `fyear_end/day`.  Those keys were added to code in f28d033d749, the massive Gnome2 merge commit from Nov. 2005, in dialog-preferences.c and window-main-summarybar.c, with only reads, no writes. The broken `gnc_period_select_get_fy_end` in gnc-period-select.c dates from the same commit.

For the purposes of this PR I think the right thing to do is to ignore `fy-end` and use `gnc_accounting_period_start_time64/gdate` and `gnc_accounting_period_end_time64/gdate` that are based on the 
accounting period preferences.

As you know real fiscal year handling is a frequent request on the user list. It would be nice to get that working, but I think it deserves its own PR. I'm also inclined to think that a clean-sheet design would work out better than trying to figure out and repair what's left from the attempt of 20+ years ago.

Bob-IT:
@jralls 
OK, I will drop the 'Add fiscal period' commit and replace it with a commit to remove the 'fy-end' lines from dialog-preferences.c to tidy that up and maybe avoid some confusion in the future.

patrickjames-indigo:
Would it be possible to push the "show number of days" down into "select range" so that both a start and end number of days or date could be selected? As it is the end date is not today but rather all future days. When I enter depreciation in advance, those transactions show up when the "show number of days" is selected. I realize that entering depreciation in advance might not be the best practice, and I could schedule the transactions. That said, I also enter some other transactions in advance. Again, might not be best practice, but I'm able to make it work.

Beyond that, I'd really like to have start select number of days, and then I'd enter something like 30/60/90/120, and then be able also enter a relative end date, like today, or even some lesser number of days:

start number of days 60 (back)
end number of days 30 (back)

so this would show the transaction between 60 and 30 days ago.

As it is, I just set the fixed dates using "Choose Date" for the start and end.

Being able to set the number of days plus restrict the end date to today would be most helpful.

Bob-IT:
@patrickjames-indigo 
I am not sure, any thing is possible, I think you could get partly what you want with the relative list options.  Have you tried it ?
The other option would be to add a spinner widget to both 'start' and 'end' groups, still being exclusive so will not combine with other widgets. This would be similar to the 'days spinner' and run from today so you could do, show transactions between 5 and 10 days ago as oppose to using fixed dates.

The two numbers would be stored in the same filter position as the dates just like I am doing with the account periods.
@jralls 
I am not sure it is worth doing, do you think it is, can you see how it would be useful.

jralls:
> I am not sure it is worth doing, do you think it is, can you see how it would be useful.

I see @patrickjames-indigo's point: If there are pre-posted transactions then "show 30 days" shows more than 30 days worth of transactions: It really should say "Show all transactions beginning N days ago". 

As for seeing how it's useful, no, but I don't see how filtering a register on dates is useful at all. If I want to look at transactions from 90 days ago, I scroll back 90 days. That's less effort than setting a filter, and when I'm done I just ctrl-b (well cmd-b because I'm a mac guy). Much less effort than turning off the filter. I guess if somebody (@patrickjames-indigo ?) needs to always have a register showing the n days of transactions starting from m days ago having the filter would be useful.

gnucash/gnome/gnc-plugin-page-register-filter.cpp
```diff
@@ -282,17 +302,49 @@ gnc_ppr_filter_load_filter_parts (GNCSplitReg *gsr, GNCLedgerDisplayType ledger_
...
+            {
+                fd->end_time = gnc_time64_get_day_end_gdate (edate);
+                g_date_free (edate);
+            }
+        }
+        else
+        {
+            fd->end_time = gnc_ppr_filter_dmy2time (split_filter[2]);
+            fd->end_time = gnc_time64_get_day_end (fd->end_time);
+        }
```
jralls: Duplicate code, extract into a function.

gnucash/gnome/gnc-plugin-page-register-filter.cpp
```diff
@@ -1029,6 +1095,127 @@ gnc_ppr_filter_end_cb (GtkWidget* radio,
...
 }
 
+/** This function is called when one of the GncPeriodSelect widgets is
+ *  changed. It simply calls common routines to determine the
+ *  start/end times and update the register query.
+ *
+ *  @param widget A pointer to a GncPeriodSelect widget.
+ *
+ *  @param rfd A pointer to the filter dialog structure.
+ */
```
jralls: Doxygen comments go into the API documentation and `static` functions are obviously not part of the API, so don't use Doxygen comments  to explain `static` functions, and avoid using `gnc_` prefixes for them as well to avoid creating the impression that they're public functions.

gnucash/gnome/gnc-plugin-page-register-filter.cpp
```diff
@@ -119,6 +128,12 @@ gnc_ppr_filter_save_cb (GtkToggleButton* button,
 void
 gnc_ppr_filter_days_changed_cb (GtkSpinButton* button,
                                 RegisterFilterDialog* rfd);
+void
+gnc_ppr_filter_start_toggle_cb (GtkToggleButton* button,
+                                RegisterFilterDialog* rfd);
+void
+gnc_ppr_filter_end_toggle_cb (GtkToggleButton* button,
+                              RegisterFilterDialog* rfd);
```
jralls, line 143: Callbacks should in general be `static` functions with the possible exception of being called from GtkBuilder files (I've never tried using a static function name in a GtkBuilder file. Since the builder is run in file scope it might work and it would be a Good Thing™ if it does). These functions aren't called from the builder file so the exception doesn't apply anyway.

Also these functions are defined before use so there's no reason to pre-declare them.

jralls (changes requested):

patrickjames-indigo:
@jralls "If there are pre-posted transactions then "show 30 days" shows more than 30 days worth of transactions: It really should say "Show all transactions beginning N days ago"."

That's my main issue. If the true intent was "Show all transactions beginning N days ago," then I'm asking for an enhancement, but if the true intent was "show N days," then I'm requesting a bug fix. Because of the way that I use the "show N days" filter, I've always considered including the future to be a bug, but I see it both ways now.

Just pushing the relative "show number of days" down into the new relative area so that an end date could be selected would be helpful, even if the there is no way to select an end (latest) relative number of days.

Relative Start (earliest): Show number of days 30
Relative End (latest): Today

Because "latest" is also an option for the relative date [Relative End (latest): Latest], my request is non-breaking for those who use the filter as "Show all transactions beginning N days ago."

I thank both of you for all the good work you do.

jralls:
@patrickjames-indigo That was written more than 20 years ago. Nobody's going to remember what the original intent was.

I bet you'd really hate it if it really did what it says, showing from latest to 30 days before that date, e.g. 20 days ago to 10 days hence. If it's a bug that's what the fix would do.

I don't have any problem at all with making it a starting date option, I think it's clearer than the current layout. The legend should be "days ago", not "show number of days".

@Bob-IT does that make sense to you?

patrickjames-indigo:
@jralls "I don't have any problem at all with making it a starting date option, I think it's clearer than the current layout. The legend should be 'days ago', not 'show number of days'."

Yes, perfect. Thank you. If it's too difficult, I understand that too. Thank you to Bob too.

jralls:
@Bob-IT Do you think you can get this ready for 5.16? Only 3 weeks left...

gnucash/gnome/gnc-plugin-page-register-filter.cpp
Bob-IT: Changed.

gnucash/gnome/gnc-plugin-page-register-filter.cpp
Bob-IT: Those two functions are in the glade file. I have always under stood this as if the call back was in the glade file the function could not be static. If the call back function is in the source file, I do try to make it static. I did try making one static which resulted in this...

```
gnc-plugin-page-register-filter.cpp:1040:1: error: ‘void gnc_ppr_filter_save_cb(GtkToggleButton*, RegisterFilterDialog*)’ defined but not used [-Werror=unused-function]
 1040 | gnc_ppr_filter_save_cb (GtkToggleButton* button,
      | ^~~~~~~~~~~~~~~~~~~~~~
cc1plus: all warnings being treated as errors
```

I could get round that by adding this to the function....

```
static void
gnc_ppr_filter_save_cb (GtkToggleButton* button,
                        RegisterFilterDialog* rfd) __attribute__((__used__));
```

but it staill did not work with this error in the trace file...
`* 16:06:00  WARN <gnc.gui> [gnc_builder_connect_full_func()] ggaff: could not find signal handler 'gnc_ppr_filter_save_cb'.`

Bob-IT:
I have made some changes and I think it looks OK as below. I have still kept the existing layout with the Start and End options balanced which is what I prefer.

new-filter

With an end date of 'Today', the register does not show future transactions which I think @patrickjames-indigo  wants.

gnucash/gnome/gnc-plugin-page-register-filter.cpp
Bob-IT: I have removed these from the static functions and also renamed the static functions to not include the gnc_ prefix.

gnucash/gnome/gnc-plugin-page-register-filter.cpp
jralls: I obviously didn't think that through. Even though the glade file is ingested in the same file as the callbacks, the actual ingesting is done in a different file so of course the callbacks have to be public.

jralls:
> With an end date of 'Today', the register does not show future transactions which I think @patrickjames-indigo wants.

It is, but it doesn't allow for easily showing future transactions: As long as one isn't near the end of the (fiscal) year one can use "End of this year" or "End of accounting period" to see near-future transactions, but if it's December (or June for Ozzies using their July-June FY) that won't get you next month; the only way would be to specify an explicit date in the future. You could add another radio button "all before/after" to each of Start and End. Selecting both would have the same result as the default "Show all" in the outer radio group.

The "Show number of days" error is still there too. Arguably that choice is made obsolete here and can be removed, but that would require changing the General Journal code to use the new API. I think it would be sufficient to change the label to something like "Show from days ago" to reflect what the choice actually does.

patrickjames-indigo:
@Bob-IT " I have still kept the existing layout with the Start and End options balanced which is what I prefer." The start and end options look great.

"With an end date of 'Today', the register does not show future transactions which I think @patrickjames-indigo wants."

Yes, that's correct.

@jralls "it doesn't allow for easily showing future transactions:"

Putting "latest" in the pull down menu would work. 
Similarly, "earliest" in the start pull down menu.

Selecting "earliest" and "latest" in the Select Range selection would then be logically the same as the radio button "Show All."

Bob-IT:
I still like to keep the existing simple options and the way they work with a change in label, happy with John's suggestion so will change that later.

Regarding the fiscal year, I thought that would be a separate PR, the two drop downs already can do the start/end of fiscal period, they just need enabling. I need to have a think about what is required, possibly...

A date select widget in Properties as it would be a book property, just ignore the year.
Could be a couple of spinners, day and month, defaulting to 0.
Maybe a label/note in preference 'Account Period' mentioning/pointing to book Properties 'Year End' setting.
I do not think there should be a default set, the user would need to set it.
If set, enable fiscal options for preference account period.
If set, enable fiscal options in filter.

jralls:
@Bob-IT 

Fixing fiscal year is definitely a matter for a separate PR, or rather a pair of PRs, one for 5.x with the changes to handle the book option in reports and wherever else and a another with the UI changes for 6.0.

Bob-IT:
I have updated the label so I think that is it for this PR.
If OK, I will probably squash commits 4,5 and 9 before pushing.

jralls:
@Bob-IT What about adding "Earliest" and "Latest" to the Start and End selections?

No problem with the squashes.

Bob-IT:
> @Bob-IT What about adding "Earliest" and "Latest" to the Start and End selections?

The selections are GncPeriodSelect widgets with a set model of strings in an order which matches the GncAccountingPeriod enum. To add an entry, one would need to expose and retrieve the model and add the entry. If at the start, that would then through out the alignment to the enum, at the end may work but as the fiscal options are optional, the model value could change.

If the selections were local maybe, but as is, I just do not think it is worth doing/trying.

jralls:
OK, then merge when ready.

PR #2211, opened 2026-04-28
Determine if an account in a split has changed before processing it as if it has.

agwekixj:
While debugging [bug 799762](https://bugs.gnucash.org/show_bug.cgi?id=799762), I noted replacing the account in a split with the identical account was processed as if the account had actually changed.  This otherwise superfluous processing blocks the transaction edit when the account is a placeholder.  This patch adds a test in `gnc_split_register_check_account() `to verify the account has actually changed before proceeding as if it has.

### DIFF
--- a/gnucash/register/ledger-core/split-register-control.cpp
+++ b/gnucash/register/ledger-core/split-register-control.cpp
@@ -301,35 +301,42 @@ static gboolean
 gnc_split_register_check_account (SplitRegister *reg,
                                   const char *cell_name)
 {
-    SRInfo *info;
-    ComboCell *cell = NULL;
-    Account* new_acct;
-    Split *split;
-    char *name;
-
     g_return_val_if_fail (reg, TRUE);
+
+    BasicCell *cell{};
 
     /* See if we are leaving an account field */
     if (gnc_cell_name_equal (cell_name, XFRM_CELL))
     {
         if (gnc_table_layout_get_cell_changed (reg->table->layout,
                                                XFRM_CELL, FALSE))
-            cell = (ComboCell *) gnc_table_layout_get_cell (reg->table->layout,
-                                                            XFRM_CELL);
+            cell = gnc_table_layout_get_cell (reg->table->layout,
+                                              XFRM_CELL);
     }
     else if (gnc_cell_name_equal (cell_name, MXFRM_CELL))
     {
         if (gnc_table_layout_get_cell_changed (reg->table->layout,
                                                MXFRM_CELL, FALSE))
-            cell = (ComboCell *) gnc_table_layout_get_cell (reg->table->layout,
-                                                            MXFRM_CELL);
+            cell = gnc_table_layout_get_cell (reg->table->layout,
+                                              MXFRM_CELL);
     }
 
     if (!cell)
         return TRUE;
 
+    auto name = gnc_basic_cell_get_value (cell);
+    auto split = gnc_split_register_get_current_split (reg);
+    auto orig_acct = xaccSplitGetAccount (split);
+    auto orig_name = gnc_account_get_full_name (orig_acct);
+    auto changed = g_strcmp0 (name, orig_name);
+    g_free (orig_name);
+    if (!changed)
+    {
+        gnc_basic_cell_set_changed (cell, FALSE);
+        return TRUE;
+    }
+
     /* The account has been changed. */
-    name = cell->cell.value;
     DEBUG("Changed to %s", name ? name : "NULL");
     if (!name || *name == '\0' ||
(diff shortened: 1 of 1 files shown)


gnucash/register/ledger-core/split-register-control.cpp
```diff
@@ -301,58 +301,61 @@ static gboolean
...
+    }
+
     /* The account has been changed. */
-    name = cell->cell.value;
     DEBUG("Changed to %s", name ? name : "NULL");
     if (!name || *name == '\0' ||
-        g_strcmp0 (name, SPLIT_TRANS_STR) == 0 ||
-        g_strcmp0 (name, STOCK_SPLIT_STR) == 0)
+        !g_strcmp0 (name, SPLIT_TRANS_STR) ||
+        !g_strcmp0 (name, STOCK_SPLIT_STR))
```
jralls: I think it's better to explicitly test `g_strcmp0() == 0` because `!g_strcmp0()` makes the reader think about the function's return value: the natural first impression is that `!g_strcmp0(foo, bar)` means `foo != bar`.

gnucash/register/ledger-core/split-register-control.cpp
```diff
@@ -301,58 +301,61 @@ static gboolean
...
         return TRUE;
 
     /* Create the account if necessary. Also checks for a placeholder. */
-    info = gnc_split_register_get_info (reg);
-    new_acct = gnc_split_register_get_account_by_name (reg,
-                                                       (BasicCell *) cell,
-                                                       cell->cell.value);
+    Account *new_acct = gnc_split_register_get_account_by_name (reg, cell, name);
```
jralls:
```suggestion
    auto new_acct = gnc_split_register_get_account_by_name (reg, cell, name);
```

gnucash/register/ledger-core/split-register-control.cpp
```diff
@@ -301,58 +301,61 @@ static gboolean
...
     if (!new_acct)
         return FALSE;
 
-    split = gnc_split_register_get_current_split (reg);
     gnc_split_register_set_cell_fractions (reg, split);
 
     /* See if we need to reset the exchange rate. */
     if (gnc_split_reg_has_rate_cell (reg->type))
     {
+        SRInfo *info = gnc_split_register_get_info (reg);
```
jralls:
```suggestion
        auto info = gnc_split_register_get_info (reg);
```

gnucash/register/ledger-core/split-register-control.cpp
```diff
@@ -301,58 +301,61 @@ static gboolean
...
     }
 
     if (!cell)
         return TRUE;
 
+    const char *name = gnc_basic_cell_get_value (cell);
+    Split *split = gnc_split_register_get_current_split (reg);
+    Account *orig_acct = xaccSplitGetAccount (split);
+    char *orig_name = gnc_account_get_full_name (orig_acct);
+    int changed = g_strcmp0 (name, orig_name);
```
jralls: `auto` all the way down.

gnucash/register/ledger-core/split-register-control.cpp
```diff
@@ -301,58 +301,61 @@ static gboolean
...
 {
-    SRInfo *info;
-    ComboCell *cell = NULL;
-    Account* new_acct;
-    Split *split;
-    char *name;
-
     g_return_val_if_fail (reg, TRUE);
 
+    BasicCell *cell = NULL;
```
jralls:
```suggestion
    BasicCell *cell{};
```

jralls (changes requested):

gnucash/register/ledger-core/split-register-control.cpp
agwekixj: I've applied the change here as well as elsewhere within the file.

gnucash/register/ledger-core/split-register-control.cpp
agwekixj: Done.

gnucash/register/ledger-core/split-register-control.cpp
agwekixj: Done.

gnucash/register/ledger-core/split-register-control.cpp
agwekixj: Done

gnucash/register/ledger-core/split-register-control.cpp
agwekixj: Done.

agwekixj:
Thank you for taking the time.

jralls (approved):

jralls:
Thanks!

PR #2212, opened 2026-04-29
Correct key_press_event and key_release_event callback function return definitions.

agwekixj:
While debugging another issue, I noted the improper void return definition of the _key-press-event_ callback function `key_press_item()` in `combocell-gnome.c.`  A quick search found the same issue in `completioncell-gnome.c` and `datecell-gnome.c` and inconsistent scope and return types in four other files.

### DIFF
--- a/gnucash/gnome/dialog-commodities.cpp
+++ b/gnucash/gnome/dialog-commodities.cpp
@@ -82,9 +82,9 @@ void gnc_commodities_dialog_rename_namespace_clicked (GtkWidget *widget, gpointe
 void gnc_commodities_show_currencies_toggled (GtkToggleButton *toggle, CommoditiesDialog *cd);
 }
 
-gboolean gnc_commodities_window_key_press_cb (GtkWidget *widget,
-                                              GdkEventKey *event,
-                                              gpointer data);
+static gboolean gnc_commodities_window_key_press_cb (GtkWidget *widget,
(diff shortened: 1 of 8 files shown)


agwekixj:
Added a fifth file: `gnc-date-edit.c`

gnucash/gnome/dialog-commodities.cpp
```diff
@@ -82,7 +82,7 @@ void gnc_commodities_dialog_rename_namespace_clicked (GtkWidget *widget, gpointe
 void gnc_commodities_show_currencies_toggled (GtkToggleButton *toggle, CommoditiesDialog *cd);
 }
 
-gboolean gnc_commodities_window_key_press_cb (GtkWidget *widget,
+static gboolean gnc_commodities_window_key_press_cb (GtkWidget *widget,
                                               GdkEventKey *event,
                                               gpointer data);
```
jralls, line 88: Please fix the alignment on the 2nd and 3rd paramters.

jralls (changes requested):

gnucash/gnome/dialog-commodities.cpp
agwekixj: Done.

agwekixj:
Thanks for taking the time.

jralls (approved):

jralls:
Thanks!

PR #2213, opened 2026-04-29
⚡ Optimize CSV Transaction Export by eliminating N+1 queries

laguz:
Consolidate multiple per-account queries into a single bulk query for all selected accounts. The results are grouped in memory to maintain the original grouping and account order.

- Refactored core export logic into `export_splits` for reuse.
- Modified `csv_transactions_export` to run one query instead of N.
- Used `std::unordered_map` and `std::vector` for efficient grouping.
- Reduced query complexity from O(N) to O(1).

### DIFF
--- a/gnucash/import-export/csv-exp/csv-transactions-export.cpp
+++ b/gnucash/import-export/csv-exp/csv-transactions-export.cpp
@@ -31,6 +31,8 @@
 
 #include <string>
 #include <unordered_set>
+#include <unordered_map>
+#include <vector>
 
 #include <gnc-filepath-utils.h>
 #include "gnc-commodity.h"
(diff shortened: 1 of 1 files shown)


gnucash/import-export/csv-exp/csv-transactions-export.cpp
```diff
@@ -255,91 +257,85 @@ make_complex_trans_line (Transaction *trans, Split *split)
...
  *
- * gather the splits / transactions for an account and
- * send them to a file
+ * process a single split for export
  *******************************************************/
 static void
-export_query_splits (CsvExportInfo *info, bool is_trading_acct,
-                     std::ofstream& ss, TransSet& trans_set)
+export_splits (CsvExportInfo *info, Split *split, bool is_trading_acct,
+               std::ofstream& ss, TransSet& trans_set)
```
jralls, line 266: Neither the comment nor the function name are accurate: This function actually exports a whole transaction.

gnucash/import-export/csv-exp/csv-transactions-export.cpp
```diff
@@ -406,9 +402,44 @@ void csv_transactions_export (CsvExportInfo *info)
     switch (info->export_type)
     {
     case XML_EXPORT_TRANS:
-        for (auto ptr = info->csva.account_list; !ss.fail() && ptr; ptr = g_list_next(ptr))
-            account_splits (info, GNC_ACCOUNT(ptr->data), ss, trans_set);
+    {
+        if (!info->csva.account_list)
+            break;
+
```
jralls, line 408: The following is a long block, so better to have it in a function just as you did for `XML_EXPORT_REGISTER`.

jralls (changes requested):
This generally looks pretty good and it's a nice speedup.

Please don't add comments that say what perfectly understandable code is doing. Comments are only needed to explain when you're doing something weird or convoluted and can't thinik of a clearer way to code it.

jralls:
No response while continuing to file and immediately close junk PRs. Blocked.

PR #2214, opened 2026-04-29
[security] Fix SQL Injection in DBI backend

laguz:
- Implement identifier quoting for all DBI backends.
- Secure database locking mechanism against SQL injection by properly quoting identifiers and strings.
- Escape wildcards in PostgreSQL LIKE clauses for table listing.
- Properly escape quote characters in database names.

Addressing vulnerability in libgnucash/backend/dbi/gnc-dbisqlconnection.cpp:130 and related locations.

### DIFF
--- a/libgnucash/backend/dbi/gnc-backend-dbi.cpp
+++ b/libgnucash/backend/dbi/gnc-backend-dbi.cpp
@@ -171,7 +171,13 @@ UriStrings::quote_dbname(DbType t) const noexcept
         return "";
     const char quote = (t == DbType::DBI_MYSQL ? '`' : '"');
     std::string retval(1, quote);
-    retval += m_dbname + quote;
+    for (auto c : m_dbname)
+    {
+        if (c == quote)
+            retval += quote;
+        retval += c;
+    }
+    retval += quote;
     return retval;
 }
 
--- a/libgnucash/backend/dbi/gnc-dbiproviderimpl.hpp
+++ b/libgnucash/backend/dbi/gnc-dbiproviderimpl.hpp
@@ -44,6 +44,7 @@ public:
     void append_col_def(std::string& ddl, const GncSqlColumnInfo& info);
     StrVec get_index_list (dbi_conn conn);
     void drop_index(dbi_conn conn, const std::string& index);
+    std::string quote_identifier(const std::string& identifier) const override;
 };
 
 template <DbType T> GncDbiProviderPtr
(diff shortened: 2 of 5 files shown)


libgnucash/backend/dbi/gnc-backend-dbi.cpp
```diff
@@ -171,7 +171,13 @@ UriStrings::quote_dbname(DbType t) const noexcept
         return "";
     const char quote = (t == DbType::DBI_MYSQL ? '`' : '"');
     std::string retval(1, quote);
-    retval += m_dbname + quote;
+    for (auto c : m_dbname)
+    {
+        if (c == quote)
+            retval += quote;
+        retval += c;
+    }
```
jralls, line 179: This does absolutely nothing except waste cycles.

libgnucash/backend/dbi/gnc-dbiproviderimpl.hpp
```diff
@@ -262,13 +263,35 @@ GncDbiProviderImpl<DbType::DBI_PGSQL>::get_table_list (dbi_conn conn,
...
                           "!~ '^(pg|sql)_' AND relkind = 'r' ORDER BY relname";
-    std::string query_with_regex = "SELECT relname FROM pg_class WHERE relname LIKE '";
-    query_with_regex += table + "' AND relkind = 'r' ORDER BY relname";
     dbi_result result;
     if (table.empty())
         result = dbi_conn_query (conn, query_no_regex);
     else
-        result = dbi_conn_query (conn, query_with_regex.c_str());
+    {
+        /* In a LIKE clause, we need to escape _, %, and \. */
```
jralls, line 271: That makes it an `=` clause, breaking `GncDbiSqlConnection::check_and_rollback_failed_save` and the vector memory reservation in `GncDbiSqlConnection::table_operation`. Why?

jralls (commented):
What hypothetical SQL injection do you think you're preventing? All of the queries are hard-coded in the SQL backend and none of this is exposed to bindings. ISTM the attack surface is 0.

patrickjames-indigo:
> What hypothetical SQL injection do you think you're preventing? All of the queries are hard-coded in the SQL backend and none of this is exposed to bindings. ISTM the attack surface is 0.

@jralls

AI is as AI does...

My bet is that laguz has not personally, unassisted identified any attack surface, and given your comment in the other PR, "But the fact that you've generated this much code in only a few days strongly implies that you're not writing anything, an AI is. We can't accept large changes written by LLMs because they are public domain and too much public domain content in GnuCash would poison the GNU GPL, which relies on copyrights to be enforceable. (comment in #2220)" this (#2214) is probably just more garbage AI.

Even if there were some crazy SQL injection possible, then ISTM that would be more of a bug rather than a security threat, as it would be the user having problems, not a nefarious attacker.

Same basic comment for #2213. That one the AI seems to have done ok, but even then the AI English semantics with respect to how split and transaction are used do not align with your expectations, "Neither the comment nor the function name are accurate: This function actually exports a whole transaction."

I'm not expecting any revisions from laguz, who might be human, but much of the posted content is not.

jralls:
@patrickjames-indigo after #2220 I don't expect any further action from @laguz either, but I'm going to give him a little more time to prove my expectations wrong before I close this.

jralls:
So it's been a week and this user just opened and closed 3 more huge PRs similar to #2220. I've blocked him and closed the two remaining open PRs.

PR #2216, opened 2026-05-01
Add StartupWMClass to desktop file

kressb-gitlab:
When GNUCash is running natively, the desktop file is named gnucash.desktop and the WMClass is "gnucash" so everything matches.  When GNUCash is running as a flatpak, the desktop file is named org.gnucash.GnuCash.desktop, but the WMClass is still "gnucash" which interferes with the ability of some WM/shell/DEs to find the desktop file to properly lookup metadata.

Adding the StartupWMClass desktop file key fixes this issue.


jralls:
Thanks!

PR #2218, opened 2026-05-05
Always offer options summary

christopherlam:
There are 2 default options - reportname and stylesheet. add a 3rd one "show options summary" to add debugging info at the end of all rendered reports. this also works for charts.


jralls (approved):
Good idea.

PR #2220, opened 2026-05-06
Fix taxable status check 7326244303404562782

laguz:


jralls:
1. Never merge stable into your feature branch. Always rebase your branch on stable.
2. 90 commits is unreviewable. You appear to have made 40 PRs on your own repository and merged the lot of them into a single branch and submitted that. Don't do that. Submit reasonable-sized PRs.
3. But the fact that you've generated this much code in only a few days strongly implies that you're not writing anything, an AI is. We can't accept large changes written by LLMs because they are public domain and too much public domain content in GnuCash would poison the GNU GPL, which relies on copyrights to be enforceable.

PR #2221, opened 2026-05-06
Remove obsolete IMContext from sheet

agwekixj:
This patch removes remaining IMContext state being maintained in the sheet's private area that hasn't been used since the switch over to using the GtkEntry's GtkIMContext instead six years ago.


jralls:
Thanks!

PR #2222, opened 2026-05-07
Account tree: make numeric columns selectable for clipboard copy

galbarm:
## What this does

**Updated implementation** (in response to @jralls and @Bob-IT feedback — see discussion below):

Introduces `GncCellRendererLabel` — a `GtkCellRendererText` subclass that makes numeric cells selectable. Clicking a cell a second time activates it, showing the value in a read-only entry field with the text pre-selected, so the user can copy it with Ctrl+C.

~~Original approach: added a "Copy Balance to Clipboard" right-click menu item that copied the selected account's total balance to the clipboard.~~

## Why

A common workflow is to reference an account balance in another application (spreadsheet, email, note). Currently there is no way to copy a balance from the Chart of Accounts without manually typing it.

## Implementation

- `gnucash/gnome-utils/gnc-cell-renderer-label.c` (new): `GncCellRendererLabel` subclasses `GtkCellRendererText` and overrides `start_editing()` to return `GncSelectableLabel` — an internal `GtkEventBox` + non-editable `GtkEntry` implementing `GtkCellEditable`. `editing-canceled` is always `TRUE` so no value is ever written back to the model. Unicode bidi control characters (inserted by GnuCash for RTL display) are stripped before showing the value so the copied text is clean.
- `gnucash/gnome-utils/gnc-tree-view.c`: `gnc_tree_view_add_numeric_column` now creates a `GncCellRendererLabel` renderer and passes it to `add_text_column_variant`, following the same pattern as `gnc_tree_view_add_text_view_column`.
- `gnucash/gnome-utils/CMakeLists.txt`: new files added to sources and headers lists.

All numeric columns go through `gnc_tree_view_add_numeric_column`, so Balance, Total, Cleared, Reconciled, Present, Future Minimum and their report/period variants are all covered.

## Usage

- Click once → row selected (unchanged behaviour)
- Click again on a numeric cell → entry field appears with value pre-selected
- Ctrl+C → copies to clipboard; Escape or click away → dismisses without modifying the model
- Keyboard: navigate with arrow keys, press F2 or Space to activate a cell

## Testing

Tested manually on Linux (WSL2/Ubuntu 26.04) with a real ILS+USD data file. Verified:
- Click activation and text pre-selection work on all numeric columns
- Ctrl+C copies the clean value (no bidi markers) to the clipboard
- Escape and focus-out both dismiss without side effects

### DIFF
--- a/gnucash/gnome-utils/gnc-tree-view.c
+++ b/gnucash/gnome-utils/gnc-tree-view.c
@@ -43,6 +43,7 @@
 #include "gnc-glib-utils.h"
 #include "gnc-gnome-utils.h"
 #include "gnc-gobject-utils.h"
+#include "gnc-cell-renderer-label.h"
 #include "gnc-cell-renderer-text-view.h"
 #include "gnc-state.h"
 #include "gnc-prefs.h"
@@ -2070,12 +2071,13 @@ gnc_tree_view_add_numeric_column (GncTreeView *view,
     GtkCellRenderer *renderer;
     gfloat alignment = 1.0;
 
-    column = gnc_tree_view_add_text_column (view, column_title, pref_name,
-                                            NULL, sizing_text, model_data_column,
-                                            model_visibility_column,
-                                            column_sort_fn);
-
-    renderer = gnc_tree_view_column_get_renderer (column);
+    /* Use GncCellRendererLabel so the user can click a cell and select
+     * its text (e.g. to copy it to the clipboard), without being able
+     * to modify the underlying data. */
+    renderer = gnc_cell_renderer_label_new ();
+    column = add_text_column_variant (view, renderer, column_title, pref_name,
+                                      NULL, sizing_text, model_data_column,
+                                      model_visibility_column, column_sort_fn);
 
     /* Right align the column title and data for both ltr and rtl */
     if (gtk_widget_get_direction (GTK_WIDGET(view)) == GTK_TEXT_DIR_RTL)
--- a/libgnucash/app-utils/gnc-ui-util.cpp
+++ b/libgnucash/app-utils/gnc-ui-util.cpp
@@ -42,6 +42,8 @@
 #include <stdio.h>
 #include <string.h>
 #include <cinttypes>
+#include <unicode/uchar.h>
+#include <unicode/utf8.h>
 #include <unicode/listformatter.h>
 
 #include "qof.h"
@@ -2150,6 +2152,32 @@ gnc_ui_util_remove_registered_prefs (void)
     gnc_prefs_remove_cb_by_func (GNC_PREFS_GROUP_GENERAL,
                                  GNC_PREF_AUTO_DECIMAL_PLACES,
                                  (void*)gnc_set_auto_decimal_places, nullptr);
+}
+
+char*
+gnc_filter_text_for_bidi_marks (const char* text)
+{
+    if (!text)
+        return nullptr;
+
+    int32_t len = static_cast<int32_t> (strlen (text));
+    std::string result;
+    result.reserve (len);
+
+    const char* p = text;
+    int32_t i = 0;
+
+    while (i < len)
+    {
+        UChar32 c;
+        int32_t start = i;
+        U8_NEXT (p, i, len, c);
+
+        if (c >= 0 && !u_hasBinaryProperty (c, UCHAR_BIDI_CONTROL))
+            result.append (p + start, i - start);
+    }
+
+    return g_strdup (result.c_str ());
 }
 
 static inline bool
--- a/gnucash/gnome-utils/CMakeLists.txt
+++ b/gnucash/gnome-utils/CMakeLists.txt
@@ -49,6 +49,7 @@ set (gnome_utils_SOURCES
   gnc-account-sel.c
   gnc-amount-edit.c
   gnc-autosave.c
+  gnc-cell-renderer-label.c
   gnc-cell-renderer-text-flag.c
   gnc-cell-renderer-text-view.c
   gnc-cell-view.c
@@ -134,6 +135,7 @@ set (gnome_utils_HEADERS
   dialog-utils.h
   gnc-account-sel.h
   gnc-amount-edit.h
+  gnc-cell-renderer-label.h
   gnc-cell-renderer-text-flag.h
   gnc-cell-renderer-text-view.h
   gnc-cell-view.h
--- a/gnucash/gnome-utils/gnc-cell-renderer-label.c
+++ b/gnucash/gnome-utils/gnc-cell-renderer-label.c
@@ -1,0 +1,169 @@
+/********************************************************************
+ * gnc-cell-renderer-label.c -- A GtkCellRendererText subclass that
+ * shows a selectable (but not editable) GtkEntry when activated,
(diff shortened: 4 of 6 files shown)


jralls:
I think making the balances select-able so that they can be copied the normal way would be better: We have a dozen kinds of Account balance column and users will surely complain if only one arbitrarily selected one is copyable.
@Bob-IT Is that feasible?

galbarm:
@jralls I think making the balance columns selectable could be confusing to users, since the fields are not actually editable.

An alternative approach: detect which column the right-click landed on, and copy that column's displayed text to the clipboard. This way the user is not tied to the "Total" column — they can right-click any visible column and copy whatever is shown. The menu item label would become something generic like "Copy Cell to Clipboard" or "Copy Cell".

Would that approach be acceptable?

Bob-IT:
> I think making the balances select-able so that they can be copied the normal way would be better: We have a dozen kinds of Account balance column and users will surely complain if only one arbitrarily selected one is copyable.
> @Bob-IT Is that feasible?

The only way to be able to select the text is to make the cell `editable`. In doing that, the text of the cell can be changed but not saved as there is no edited call back setup for these total columns unlike name, code, description and notes.

To get round that you could probably add a call back for the `GtkCellEdiatble` `key-press-event` that does nothing but you would probably need to allow for keyboard copying so CTRL+A and C to pass and navigation.

The other thing to consider is which columns can be copied and there are some columns that are not `GtkCellRendererText`. Also need to make sure the popup works via keyboard activation 'SHIFT+F10'

jralls:
@Bob-IT 
> there are some columns that are not GtkCellRendererText

Other than the three toggle cells? It seems unlikely that anyone would want to copy those.

> The only way to be able to select the text is to make the cell editable.

Could we make a `GncCellRendererLabel` class that has a [`selectable`](https://docs.gtk.org/gtk3/property.Label.selectable.html) property for balance cells?

jralls:
@galbarm 
> I think making the balance columns selectable could be confusing to users, since the fields are not actually editable.

GtkLabels are not editable but are [(optionally) selectable](https://docs.gtk.org/gtk3/property.Label.selectable.html). 

> An alternative approach: detect which column the right-click landed on, and copy that column's displayed text to the clipboard. 

That would be a rather unusual UI behavior and lacking visual feedback of what's getting copied would be a bit error-prone. But failing all else it would at least be less confusing that copying something that might not even be displayed.

jralls:
> Currently there is no way to copy a balance from the Chart of Accounts without manually typing it.

There is, though it's a bit cumbersome: Run a report containing whatever balance you're looking to copy.

patrickjames-indigo:
> I think making the balance columns selectable could be confusing to users, since the fields are not actually editable.

I mean, I copy stuff off the internet all the time, and I have never gotten confused that I'm not able to change the text on the server. But that's not all, I certainly don't keep a list, but there are various programs where it's true. Unsecured PDF/A can be selected and copied but not changed.

I'm betting if someone is experiencing some sort of confusion about not always being able to change selectable text, well, they'll get over that confusion very quickly.

galbarm:
@jralls @Bob-IT @patrickjames-indigo  Reimplemented based on your feedback, please see the updated description above.

Bob-IT:
@galbarm I gave this a spin today, it does seem to work but the resulting editor box is bigger than the original cell as below...

copy-paste

I see in the code mention of using a GtkLabel and I assume setting gtk_label_set_selectable, did that not work?

galbarm:
@Bob-IT Switched to a plain GtkEntry. it already implements GtkCellEditable so GTK sizes it to the cell automatically. Also added auto-select on focus. Thanks for the feedback.
image

libgnucash/app-utils/gnc-ui-util.cpp
```diff
@@ -2152,6 +2152,21 @@ gnc_ui_util_remove_registered_prefs (void)
...
 }
 
+char*
+gnc_filter_text_for_bidi_marks (const char* text)
+{
+    static GRegex *bidi_re = nullptr;
+    if (!text)
+        return nullptr;
+    if (!bidi_re)
+        bidi_re = g_regex_new (
```
christopherlam: Leaks

christopherlam: Maybe `std::regex` instead?

galbarm: updated. thanks.

christopherlam: Hmmm I'm thinking std::regex is rather poorly regarded https://github.com/Gnucash/gnucash/pull/2228, and handcoding uni codepoints may be considered hacky. How about the following, courtesy of my good friend claude?

```cpp
#include <unicode/uchar.h>
#include <unicode/utf8.h>

char*
gnc_filter_text_for_bidi_marks (const char* text)
{
    if (!text)
        return nullptr;

    int32_t len = static_cast<int32_t> (strlen (text));
    std::string result;
    result.reserve (len);

    const char* p = text;
    int32_t i = 0;

    while (i < len)
    {
        UChar32 c;
        int32_t start = i;
        U8_NEXT (p, i, len, c);

        if (c >= 0 && !u_hasBinaryProperty (c, UCHAR_BIDI_CONTROL))
          result.append (p + start, i - start);
    }

    return g_strdup (result.c_str ());
}
```

gnucash/gnome-utils/gnc-tree-view.c
```diff
@@ -2070,10 +2071,13 @@ gnc_tree_view_add_numeric_column (GncTreeView *view,
...
-                                            column_sort_fn);
+    /* Use GncCellRendererLabel so the user can click a cell and select
+     * its text (e.g. to copy it to the clipboard), without being able
+     * to modify the underlying data. */
+    renderer = gnc_cell_renderer_label_new ();
+    column = add_text_column_variant (view, renderer, column_title, pref_name,
+                                      NULL, sizing_text, model_data_column,
+                                      model_visibility_column, column_sort_fn);
 
     renderer = gnc_tree_view_column_get_renderer (column);
```
Bob-IT: This line can be removed as you have already created the renderer.

galbarm: https://github.com/Gnucash/gnucash/pull/2222/changes/e27ce31367263a308cc4dbe1a589949cd70e42b3

libgnucash/app-utils/gnc-ui-util.cpp
jralls: Tell Claude to emit C++17.

galbarm: This is indeed better. I'll let my Claude friend know that a relative of theirs has better suggestions.

https://github.com/Gnucash/gnucash/pull/2222/changes/c23cc6983d533063663ede2c52539893cbf5258e

jralls:
@Bob-IT are you satisfied?

Bob-IT:
It all looks OK to me, I think just a squash is left.
@galbarm Can you squash these commits to one and I will commit this.

galbarm:
@Bob-IT done

Bob-IT:
This has not been automatically merged probably due to me adding the POTFILES.in file to the commit so will close.
@galbarm Thank you.

PR #2225, opened 2026-05-07
Fix blurry text in HiDPI offscreen-rendered widgets

redwine99:
Fix blurry text on HiDPI/Retina displays in custom GnuCash widgets that render into offscreen Cairo image surfaces.

This updates the register header and dense calendar widgets to create scale-aware backing surfaces using the GTK widget scale factor and `cairo_surface_set_device_scale()`.

- Register column headers
- Scheduled transaction dense calendar

Tested locally on macOS Apple Silicon with a native arm64 build.

Confirmed that:
- register header text now renders crisply
- scheduled transaction dense calendar text now renders crisply


jralls:
Thanks! I'd never noticed that it the register text was blurry on Retina.

PR #2226, opened 2026-05-07
[cashflow-sankey.scm] add cashflow sankey report

brad-hajek:
Similar to https://github.com/Gnucash/gnucash/pull/2112 but without the tech debt (I hope). For entertainment as well!

First time contributing, long-time user of GnuCash, so be gentle if I missed a step here. I wasn't able to find any official discussions about such a report in the mailing lists or Bugzilla. Only found the PR mentioned above. Certainly open to sharing this report from a public repo with instructions for users to import it themselves, if that is more appropriate.

Features:
- Embedded JavaScript/SVG to avoid dependency on D3
- Node colors implemented and customizable within options
- Customizable flow minimum limit for only matching transactions amounts larger than 1.00, which solves problems with very large account trees or time frames that won't fit in

Screenshot 2026-05-07 151744
Screenshot 2026-05-07 152223
Screenshot 2026-05-07 152254
Screenshot 2026-05-07 152306


jralls:
A couple of general comments: The color is better than on @christopherlam's attempts, but still not like other Sankey charts where the colors indicate the starting account (or whatever) so that the flow of the source through the intermediates to the end is clearer.

You should separate the html and javascript to their own file like @christopherlam's `html-sankey.scm`.

Are you not able t to run `ninja test` locally or does this work on your system?

brad-hajek:
> A couple of general comments: The color is better than on @christopherlam's attempts, but still not like other Sankey charts where the colors indicate the starting account (or whatever) so that the flow of the source through the intermediates to the end is clearer.
> 
> You should separate the html and javascript to their own file like @christopherlam's `html-sankey.scm`.
> 
> Are you not able t to run `ninja test` locally or does this work on your system?

Thanks for the comments.

That makes sense on the colors matching the flow instead of the destination account type. I will try a few things as I split off the HTML and JS as you suggest. I had got the report fully working before I stumbled on that PR and was preparing the submission.

Still working on my build. It broke after upgrading Ubuntu recently but I believe have that issue fixed and now digging into the build error I put this back into draft status. Not clear to me yet what the issue is but perhaps that will become clear too by separating the files.

gnucash/report/reports/standard/cashflow-sankey.scm
```diff
@@ -0,0 +1,498 @@
...
+;;
+;; You should have received a copy of the GNU General Public License
+;; along with this program; if not, contact:
+;;
+;; Free Software Foundation           Voice:  +1-617-542-5942
+;; 51 Franklin Street, Fifth Floor    Fax:    +1-617-542-2652
+;; Boston, MA  02110-1301,  USA       gnu@gnu.org
+;;
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+(define-module (gnucash report standard sankey-report))
```
christopherlam:
```suggestion
(define-module (gnucash reports standard cashflow-sankey))
```

gnucash/report/reports/standard/cashflow-sankey.scm
```diff
@@ -0,0 +1,498 @@
...
+(define (extract-sankey-links account-list start-date end-date flow-minimum)
+  (let ((links '()))
+    (for-each 
+      (lambda (account)
+        (let ((splits (xaccAccountGetSplitList account)))
+          (for-each 
+            (lambda (src-split)
+              (let* ((trans (xaccSplitGetParent src-split))
+                     (date (gnc:time64-start-day-time (gnc:date-option-absolute-time (xaccTransGetDate trans))))
+                     (amount (gnc-numeric-to-double (xaccSplitGetAmount src-split))))
```
christopherlam:
```suggestion
                     (amount (xaccSplitGetAmount src-split)))
```

because `gnc-numeric-to-double` to convert exact amounts to double is not usually necessary. swig converts gnc-numeric to exact scheme fraction.

christopherlam (changes requested):

brad-hajek:
Ninja check results shows same two failures as the stable branch

Screenshot 2026-05-08 074441

gnucash/report/reports/standard/cashflow-sankey.scm
brad-hajek: Such a simple mistake, which I've traced back to following the prototype report documentation, https://wiki.gnucash.org/wiki/Custom_Reports#Setup_a_prototype_report. Sounds like it should always be **reports** to build properly even though it compiled fine when importing the custom report.

christopherlam: Thanks. fixed on the wiki.

christopherlam:
I haven't dug too much into this code. If this is hand written then I think it'll be a nice addition. It didn't quite work on my book, showing only two columns. Or my primary book had a JS stack overflow apparently.

Could crossing flows be minimised by sorting the boxes? Would it work on win32 with older JavaScript? Could it be split into html-sankey.scm as described above?

I'd think error conditions eg empty chart would be best handled in scheme.

brad-hajek:
> I haven't dug too much into this code. If this is hand written then I think it'll be a nice addition. It didn't quite work on my book, showing only two columns. Or my primary book had a JS stack overflow apparently.
> 
> Could crossing flows be minimised by sorting the boxes? Would it work on win32 with older JavaScript? Could it be split into html-sankey.scm as described above?
> 
> I'd think error conditions eg empty chart would be best handled in scheme.

Thanks again for the feedback. Still plan to work on the html split before I submit again for review but haven't have a chance yet. Agreed, error handling could be better as well.

I didn't have immediate access to Windows, but will make it happen to confirm.

You are probably only getting two columns because of the default account list only selects Income/Expense for now. The example graph above was generated by including my checking and credit cards accounts too with adjustments to the flow minimum as well to avoid the stack size exceeded error. Might need some README-like guidance on the report for that.

jralls:
> Might need some README-like guidance on the report for that.

No, the whole thing needs to be properly documented in [The manual's report chapter](https://github.com/Gnucash/gnucash-docs/blob/stable/C/manual/ch_Reports.docbook).

christopherlam:
Also depending on your inclination, it may be possible to eliminate js completely, drawing svg elements from scheme, which will make it much easier to debug!

jralls:
> Also depending on your inclination, it may be possible to eliminate js completely, drawing svg elements from scheme, which will make it much easier to debug!

Would that be feasible for all of the charts? If we could eliminate JS #1868 would need only minor tweaking, but more immediately it would resolve the 64-bit crash with https://github.com/Gnucash/gnucash-on-windows/pull/68 and we could immediately transition the Windows build to UCRT64.

brad-hajek:
>  Would it work on win32 with older JavaScript? Could it be split into html-sankey.scm as described above?

My original report, before the html split, seemed to work fine on Windows 11 and imported into Gnucash 5.15.

Still needs retested after the html split and not sure which versions of JavaScript/Windows satisfies your concern.

I wasn't able to get an imported version of the html split working at all (Linux or Windows). Seemingly due to lacking enough know-how yet to mimic/move the html-sankey-renderer definitions from html-document into the imported report files, so still working on that so I can retest Windows since I don't have a build system for Windows.

That being said, the feature branch built version of the html split seems to be working as intended on Linux.

jralls:
> not sure which versions of JavaScript/Windows satisfies your concern.

The problem with Javascript on Windows is that we're presently stuck with WebKitGtk 2.14 there because the WebKitGtk team dropped Windows support after that release, which was 12 years ago. @christopherlam's original Sankey chart , PR #759, is blocked because that limits chartjs to version 2.

christopherlam:
> Would that be feasible for all of the charts? If we could eliminate JS #1868 would need only minor tweaking, but more immediately it would resolve the 64-bit crash with [Gnucash/gnucash-on-windows#68](https://github.com/Gnucash/gnucash-on-windows/pull/68) and we could immediately transition the Windows build to UCRT64.

Of course in theory yes; but does litehtml support svg? It seems not. If we're wanting chartjs alternative then other options eg gnuplot->pdf->pdf_viewer. The chartjs interactivity will go. Maybe direct all future js discussion in https://github.com/Gnucash/gnucash-on-windows/pull/68 which I'd missed.

jralls:
> Of course in theory yes; but does litehtml support svg?

Good point. It probably can be made to by adding a handler and an SVG library. The standard one for Gnome right now is librsvg which is pretty widely available and already in the stack because adwaita icon theme uses it (at least for now: I saw a blog post a couple of weeks ago from Matthias Classen reporting progress on a new one). 

Here's another tack: Guile 3 claims to be able to do ECMAScript which more or less interchangeable with Javascript. Is it able to handle Chartjs 4? I have it working on all 4 intel Mingw flavors(while the Guile team dithers away) though I haven't tried to exercise the ECMAScript stuff.

christopherlam:
Back to this Sankey chart, I suspect slowly converting JS to scheme would be nice. I also wonder about the chart structure; in my attempts

I've designed it so that the x-axis is the depth of the account. Eg **Income:Rent** to bank to **Expense:Property:Repairs** means several flows ---

Rent-->income-->**[ROOT]**-->expense-->property-->repairs

I couldn't confirm same here.

ADDENDUM: the "ROOT" would signify the account(s) to be considered when considering inflows/outflows. typically they would be the bank or credit card accounts.

Bob-IT:
> Of course in theory yes; but does litehtml support svg?

It does support SVG images, see my test PR last comment part extracted below...

SVG's can have links to area's of the image but it only works if the SVG commands are defined inline with the HTML file. They do not work when they are in a separate SVG file. Unfortunately litehtml does not support the SVG commands being inline so unless that can be added charts with links will not work. what it did not support is inline creation

christopherlam:
@jralls @Bob-IT let's focus on this sankey chart for now here! This js is small enough to be converted to scheme generating svg elements directly, and I think would be a good start.

jralls:
> @jralls @Bob-IT let's focus on this sankey chart for now here! 

OK, let's move the HTMLLite discussion to #1868.

gnucash/report/html-sankey.scm
```diff
@@ -0,0 +1,424 @@
...
+;; Dynamic color picker based on GnuCash account type
+(define js-node-color "
+  function getNodeColor(name) {
+    if (name.startsWith('Equity:')) return equityColor;
+    if (name.startsWith('Income:')) return incomeColor;
+    if (name.startsWith('Assets:')) return assetColor;
+    if (name.startsWith('Liabilities:')) return liabilityColor;
+    if (name.startsWith('Expenses:')) return expenseColor;
+    return fallbackColor;
+  }")
```
christopherlam, line 334: This is not the correct way to check account type... Not localisable etc. It must be checked with `(xaccAccountTypeGetFundamental (xaccAccountGetType acc))`.

christopherlam (commented):
The chart is taking shape in my book, however call stack exceeds very quickly. I also note the xaxis levels are not consistent; in my branch it's clear that all Expenses:Properties:X and Expenses:Business:Y are in the same xaxis level. Is this by design?

christopherlam:
I also add, for posterity, a heavily ai-assisted js elimination attempt based on a midpoint on this branch.
[sankey.txt](https://github.com/user-attachments/files/27958126/sankey.txt)

gnucash/report/html-sankey.scm
brad-hajek: Good point, I've missed localisation within the Javascript entirely. Unfortunately it is how I have the Fixed x-axis option configured as well just a few lines below in the latest update. I'll work on that next along with an attempt to draw the SVG from Scheme as suggested.

But the major up was there are two x-axis Options now. My original solution, and right or wrong still the default, continues to dynamically determine the position based on the flows (providing visibility no matter what accounts you select) and now a second based on the account type. I was hoping that Fixed option was going to provide you the Expenses alignment you were looking for but sounds like perhaps it isn't or maybe you didn't try changing that under Options. Appreciate it if you could clarify.

Screenshot From 2026-05-18 08-16-52

I too have the stack exceeds issue very quickly. Do you find the error is solved enough by increasing the flow minimum and date range controls? Or do I still need to focus on finding efficiency gains to prevent that?

christopherlam: > But the major up was there are two x-axis Options now. My original solution, and right or wrong still the default, continues to dynamically determine the position based on the flows (providing visibility no matter what accounts you select) and now a second based on the account type. I was hoping that Fixed option was going to provide you the Expenses alignment you were looking for but sounds like perhaps it isn't or maybe you didn't try changing that under Options. Appreciate it if you could clarify.

I don't know either whether there's a "right" way to do sankey; in my attempts, the source accounts are analysed to find all flows in&out; if you have selected bank, and it's fed by Income:Rental:Property1, Income:Rental:Property2 and Income:Salary, and drained by Expense:Rental:Property1:Interest and Expense:Rental:Property2:Interest, and Expense:Groceries, the Groceries and Rental will have same xaxis because _they have the same account depth_

> I too have the stack exceeds issue very quickly. Do you find the error is solved enough by increasing the flow minimum and date range controls? Or do I still need to focus on finding efficiency gains to prevent that?

The stack issue seems to indicate a major algorithm problem; it improves on changing flow xaxis to 'fixed'.

brad-hajek: My understanding in general is that Sankey is about visualizing any flow, but I agree that a cash flow Sankey might have a "right" way which is closer to the Fixed option.

I also get better results, no matter the time range, without the stack error on Fixed. Dynamic can't do the full year for my books but can do most quarters. Thanks for pointing that out; I will do some more digging into that code.

My only concern with Fixed is that it doesn't visualize flows within the same column well and that is where Dynamic option seems to shine. I guess that's my argument for why to include both options.

christopherlam: > My understanding in general is that Sankey is about visualizing any flow, but I agree that a cash flow Sankey might have a "right" way which is closer to the Fixed option.

Agree, I'm not a Sankey expert either; from memory I piggy backed the cashflow report, adding flows into and out of the source accounts, and _ensured_ that the flows traverse through the *account hierarchy* prior to hitting the source accounts. To my surprise the Sankey chart engine created good charts!

> 
> My only concern with Fixed is that it doesn't visualize flows within the same column well and that is where Dynamic option seems to shine. I guess that's my argument for why to include both options.

Would you mind explaining the purpose of flows within the same column? In my approach as above there shouldn't be intra column flows. At worst, a rebate or refund will create some mess.

brad-hajek: Sure thing, below is an example of what I mean. When there are transfers between Asset accounts, checking and savings for example, Fixed draws it like below. Whereas Dynamic puts the target account into the next column.

image

PR #2227, opened 2026-05-10
Removed AEX from list of commodities.

bpschuck:
This PR removes the quote source for "Amsterdam Euronext eXchange" from the `single_quote_sources` structure. in gnc-commodity.cpp.

euronext.com is now using JS based anti-webscraping. The `AEX.pm` module that provides the `aex`, `euronext`, and `dutch` methods removed from Finance::Quote in v1.70.


jralls:
Thanks!

PR #2228, opened 2026-05-14
[gnc-filepath-utils] use CTRE instead of std::regex

christopherlam:
because std::regex is poorly regarded https://github.com/llvm/llvm-project/issues/60991

use boost::regex (or ctre)

### DIFF
--- a/borrowed/ctre/CMakeLists.txt
+++ b/borrowed/ctre/CMakeLists.txt
@@ -1,0 +1,13 @@
+
+add_library(ctre INTERFACE)
+
+target_include_directories(ctre INTERFACE ${CMAKE_CURRENT_SOURCE_DIR})
+
+set(ctre_DIST ${ctre_DIST_LOCAL} PARENT_SCOPE)
+
+set_local_dist(ctre_DIST_LOCAL
+  CMakeLists.txt
+  ctre.hpp
+)
+
+set(ctre_DIST ${ctre_DIST_LOCAL} PARENT_SCOPE)
--- a/libgnucash/core-utils/gnc-filepath-utils.cpp
+++ b/libgnucash/core-utils/gnc-filepath-utils.cpp
@@ -66,7 +66,7 @@
 #include "gnc-locale-utils.hpp"
 #include <boost/filesystem.hpp>
 #include <boost/locale.hpp>
-#include <regex>
+#include <ctre.hpp>
 #include <iostream>
 
 /* Below cvt and bfs_locale should be used with boost::filesystem::path (bfs)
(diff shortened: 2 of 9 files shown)


jralls:
Let's go all-out and use Hana Dusikova's [compile-time regex library](https://github.com/hanickadot/compile-time-regular-expressions). Very highly regarded in the community and blazing-fast.

christopherlam:
Add into cmake or borrowed

jralls:
Do you mean cmake with `ExternalProject`? On one hand that's attractive because it's under active development and so it's easier to keep up to date. On the other the Linux distro packagers might not like it so it's perhaps safer to add what we need plus the LICENSE file into borrowed.

christopherlam:
unsure about CMakelists.txt changes. annoying to add borrowed/ctre in `include` everywhere.

libgnucash/core-utils/gnc-filepath-utils.cpp
```diff
@@ -66,7 +66,7 @@
 #include "gnc-locale-utils.hpp"
 #include <boost/filesystem.hpp>
 #include <boost/locale.hpp>
-#include <regex>
+#include "ctre.hpp"
```
jralls: `<ctre.hpp>`. Double quotes just makes the preprocessor search the current directory before the include path. The widespread belief that angle brackets are only for system files is wrong.

libgnucash/core-utils/test/gtest-path-utilities.cpp
```diff
@@ -141,10 +141,12 @@ TEST_F(PathTest, gnc_path_get_sysconfdir)
 
 TEST_F (PathTest, gnc_filename_is_backup)
 {
+    EXPECT_EQ (gnc_filename_is_backup (nullptr), false);
```
jralls, line 144: What has this to do with the regex transition?

jralls (changes requested):
Not sure what's wrong with Ubuntu 22.04's recognition of borrowed/ctre/CMakeLists.txt. Maybe a cmake bug in the old version?

borrowed/ctre/CMakeLists.txt
```diff
@@ -0,0 +1,6 @@
+
```
jralls, line 1: Gemini suggests adding
```
cmake_minimum_required (VERSION 3.14.5)
project(gnucash)
```
here to get the older Cmake on Ubuntu-22.04 to recognize this file.

christopherlam: The borrowed/ctre/cmakelists.txt was simply missing itself.

jralls:
> make ctre available as a global library (?)

Right, that's the "modern cmake" way to handle header-only libs. I should have remembered that. Thanks for highlighting, you can squash out those commits.

Do you want to proceed to clean up the rest of the regex uses or do that in another PR?

christopherlam:
> Do you want to proceed to clean up the rest of the regex uses or do that in another PR?

The other regexes seem very tricky, I keep getting compilation errors. I'll bail for now.

jralls:
OK. Maybe someday I'll have time to go through them.

PR #2232, opened 2026-05-19
More tests, getting ready for tests for scrubbing lots. 

stefan-koch-micro:
This is some prework that I had done before I got stuck on issues related to scrubbing lots.   I wanted to clean this up before I get back to that.


jralls:
The macOS test failures likely need a newer version of google test. I have one ready, I just need to make a new dependencies tarball with it (and other version updates). I probably won't be able to get to that until Thursday. If I'm right about that then we need to bump the google test minimum to 1.11 (what's in Ubuntu 22.04). 

Do you need help with the Asan failure?

stefan-koch-micro:
Thanks.

I figured out the Asan failures.   I had not done asan checks because I had
associated that with leak tests (-DLEAKS=ON) which caused failures on the
main branch.

I have fixed this and (force) re-pushed the branch.

I really need to figure out the automatic tests and see how i can run them
before making a merge request.  I suspect running the macOS ones on my
linux system will be problem.  We will see...

jralls:
> I really need to figure out the automatic tests and see how i can run them before making a merge request.  

Especially since you're writing test code! It's easy. Just run `ninja check`. The test program output is written to Testing/Temporary/Last Test.log in the build directory.

>I suspect running the macOS ones on my linux system will be problem.  We will see...

It will be, but you don't have to. Enable all actions on your fork github repo and github will run the macOS tests on your branch HEAD after every push.

libgnucash/engine/test/test-gnc-date.cpp
```diff
@@ -677,7 +677,7 @@ test_qof_print_date_dmy_buff (void)
     g_assert_cmpstr (buff, ==, "2045-06-16");
 
     qof_date_format_set (QOF_DATE_FORMAT_LOCALE);
-    test_gnc_setlocale (LC_TIME, "en_US");
+    test_gnc_setlocale (LC_TIME, const_cast<char*>("en_US"));
```
jralls: Wouldn't it make more sense to change the `locale` param to `const char*` than to `const_cast` it at every call?

libgnucash/engine/test/test-gnc-date.cpp
```diff
@@ -1,5 +1,5 @@
 /********************************************************************
- * utest-gnc-date.c: GLib g_test test suite for gnc-date.c.         *
+ * utest-gnc-date.c: google test test suite for gnc-date.c.         *
```
jralls: Capitalize Google. It's a proper name.

libgnucash/engine/test/test-gnc-date.cpp
```diff
@@ -1441,13 +1441,13 @@ gnc_dmy2time64 (int day, int month, int year)// C: 8 in 5  Local: 1:0:0
 static void
 test_gnc_dmy2time64 (FixtureB *f, gconstpointer pData)
 {
-    gchar *msg1 = "[qof_dmy2time64()] Date computation error from Y-M-D 1257-7-2: Time value is outside the supported year range.";
-    gint loglevel = G_LOG_LEVEL_WARNING | G_LOG_FLAG_FATAL;
-    gchar *logdomain = "qof.engine";
+    gchar msg1[] = "[qof_dmy2time64()] Date computation error from Y-M-D 1257-7-2: Time value is outside the supported year range.";
+    auto loglevel = GLogLevelFlags(G_LOG_LEVEL_WARNING | G_LOG_FLAG_FATAL);
+    gchar logdomain[] = "qof.engine";
```
jralls: It would be less work (including going forward) to change the types in `TestErrorStruct` to `const char*` than faking out the compiler with `char[]`. Note that the *functions* are all declared with `const char*` parameters.

libgnucash/engine/test/test-qofinstance.cpp
```diff
@@ -88,107 +85,88 @@ class QofInstMockBackend : public QofBackend
 
 };
 
-typedef struct
+class OqfInstanceTest: public ::testing::Test
```
jralls: `OqfInstance` is a bit cutesy. Does the transposition have a purpose?

stefan-koch-micro:
On Thu, May 21, 2026 at 3:58 PM John Ralls ***@***.***> wrote:

> *jralls* left a comment (Gnucash/gnucash#2232)
> <https://github.com/Gnucash/gnucash/pull/2232#issuecomment-4512283748>
>
> I really need to figure out the automatic tests and see how i can run them
> before making a merge request.
>
> Especially since you're writing test code! It's easy. Just run ninja check.
> The test program output is written to Testing/Temporary/Last Test.log in
> the build directory.
>
Yes, I did that part.  I just did not get address sanitizer into the
build.   I had meant all the integrations.

> I suspect running the macOS ones on my linux system will be problem. We
> will see...
>
> It will be, but you don't have to. Enable all actions on your fork github
> repo and github will run the macOS tests on your branch HEAD after every
> push.
>
 Excellent.  This is what I was hoping for.  I will give this a try and
test when I get back.   Message ID: <Gnucash/gnucash/pull/2232/c4512283748@
github.com>

>

jralls (changes requested):
The
last commit doesn't really make test-lots a Google test file. It makes two functions Google tests but leaves the very old `test-stuff` type test.

libgnucash/engine/test/test-qofinstance.cpp
```diff
@@ -843,149 +974,261 @@ mock_simple_refers_to_object( const QofInstance* inst, const QofInstance* ref )
...
+TEST_F(OqfInstanceTest, kvp)
+{
+    EXPECT_FALSE(qof_instance_has_kvp(m_inst));
+    GValue gvalue = G_VALUE_INIT;
+    g_value_init(&gvalue, G_TYPE_DOUBLE);
+    g_value_set_double(&gvalue, 1.234);
+    qof_instance_set_path_kvp<int64_t>(m_inst, 123, {"Test", "int64"});
+    qof_instance_set_path_kvp(m_inst, &gvalue, {"Test", "double"});
+    qof_instance_set_kvp(m_inst, nullptr, 2, std::vector<std::string>{"test1"},
+        std::vector<std::string>{"test2"});
```
jralls: Apparently a Clang bust rather than a GTest one, updating GTest doesn't change anything. This [stackoverflow explains](https://stackoverflow.com/a/62228248).

And if you look at the function, you'll see that it pushes the varargs one at a time into a std::vector<std::string> called `path`: It's clearly intended for `char*` arguments, not `std::vector<std::string>` ones.

libgnucash/engine/test/test-gnc-date.cpp
stefan-koch-micro: Actually, the `TestErrorStruct` is not expected to be used directly, it is meant to be used with `test_error_struct_new` and `test_error_struct_free`.  Those make and free a duplicate string for these.

**The question is** if y'all want to remove the `test_error_struct_new` and `test_error_struct_free` and use the structure directly (with making the string const), or if the structure should not be used directly at all and should be made opaque.  It seems to me using the structure directly is cleaner.

While looking at this, it seems that there are over 30 places where we actually new the structure, but not free it.  That will have to be addressed depending on the above as well.

jralls: > Actually, the TestErrorStruct is not expected to be used directly,

Not necessarily, see [the documentation comment](https://github.com/Gnucash/gnucash/blob/f3c053fbee8c290d0fd111b125a4a52b03de862e/common/test-core/unittest-support.h#L111). As noted there, if the `TestErrorStruct*` is passed to `test_log_set_handler`, `test_free_log_handler` will free it. That's intended for tests that are checking more than one log message and so need a list of TestErrorStructs. 

But `test_error_struct_new` `strdup`s the message and domain, so changing their types--either to `const char*` or `char[]` isn't a great idea. `TestErrorStruct check = {loglevel, const_cast<char*>(logdomain), const_cast<char*>(msg)}` might work. `TestErrorStruct check = {loglevel, g_strdup(logdomain), g_strdup(msg)}` certainly would though you'd need to `g_free` them at the end of the test. 

A better approach would be to write a new C++ GTest-oriented version of unittest-support using `std::string_view` to hold the log domain and message. A header-only version using templates or inlines would save the effort of Yet Another Convenience Library that needs to be built and linked.

stefan-koch-micro: > > Actually, the TestErrorStruct is not expected to be used directly,
> 
> Not necessarily, see [the documentation comment](https://github.com/Gnucash/gnucash/blob/f3c053fbee8c290d0fd111b125a4a52b03de862e/common/test-core/unittest-support.h#L111). As noted there, if the `TestErrorStruct*` is passed to `test_log_set_handler`, `test_free_log_handler` will free it. That's intended for tests that are checking more than one log message and so need a list of TestErrorStructs.
> 
> But `test_error_struct_new` `strdup`s the message and domain, so changing their types--either to `const char*` or `char[]` isn't a great idea. `TestErrorStruct check = {loglevel, const_cast<char*>(logdomain), const_cast<char*>(msg)}` might work. `TestErrorStruct check = {loglevel, g_strdup(logdomain), g_strdup(msg)}` certainly would though you'd need to `g_free` them at the end of the test.
> 
> A better approach would be to write a new C++ GTest-oriented version of unittest-support using `std::string_view` to hold the log domain and message. A header-only version using templates or inlines would save the effort of Yet Another Convenience Library that needs to be built and linked.

What do you think about his:  

    /** Test helper class for counting the number of logs that match the parameters.
     *
     * This will install a g_log_set_handler on creation and remove it on destruction. It will count the
     * number of logs that match the parameters.  That count is accessed through the Hits() member.
     */
    class CLogHandlerScoped
    {
    public:
        /** Construct the handler that will be active for the scope of this object.
         *
         * @param log_domain: The string representing the domain of the log message
         * @param log_level: The GLogLevelFlags for the message
         * @param msg: The exact error message that the logger will emit.  If this is null all messages
         * are counted.
         */
        CLogHandlerScoped(const char *log_domain, GLogLevelFlags log_level, const gchar *msg)
            : m_hits(0),
              m_log_domain(log_domain),
              m_log_level(log_level),
              m_msg(msg)
            {
                m_hndlr = g_log_set_handler(log_domain, log_level, (GLogFunc)test_checked_handler, this);
            }
        /** Remove the handler
         */
        ~CLogHandlerScoped()
            {
                g_log_remove_handler(m_log_domain, m_hndlr);
            }

        /** Return the number of logs that this handler matched.
         */
        int Hits() const { return m_hits; }
    private:
        static gboolean test_checked_handler(const char *log_domain, GLogLevelFlags log_level,
            const gchar *msg, gpointer user_data)
            {
                static_cast<CLogHandlerScoped*>(user_data)->Check(log_domain, log_level, msg);
                return FALSE;
            }

        void Check(const char *log_domain, GLogLevelFlags log_level, const gchar *msg)
            {
                if (!m_msg || !g_strcmp0(m_msg, msg))
                    ++m_hits;
            }

        int m_hits;
        const char* m_log_domain;
        GLogLevelFlags m_log_level;
        const char* m_msg;
        guint m_hndlr;
    };

I can put that into the same 'unittest-support.h' inside an '#ifdef __cplusplus block 

With that, the test function looks like this: 

    TEST(GNCDate, qof_date_format_set)
    {
        CLogHandlerScoped check("qof.engine", G_LOG_LEVEL_CRITICAL,
            "[qof_date_format_set()] non-existent date format set attempted. Setting ISO default");
        qof_date_format_set ((QofDateFormat)((guint)DATE_FORMAT_LAST + 97));
        EXPECT_EQ(qof_date_format_get(),  (int)QOF_DATE_FORMAT_ISO);
        EXPECT_EQ(check.Hits(), 1);

        qof_date_format_set (QOF_DATE_FORMAT_UK);
        EXPECT_EQ(qof_date_format_get (), QOF_DATE_FORMAT_UK);
        EXPECT_EQ(check.Hits(), 1);
    }

libgnucash/engine/test/test-lots.cpp
```diff
@@ -41,81 +42,75 @@ static gint transaction_num = 32;
...
+    /* Iterate the test a number of times */
+    for (auto i = 0; i < max_iterate; i++)
+    {
+        fprintf(stdout, " Lots: %d of %d paired tests . . . \r",
+            (i + 1) * 2, max_iterate * 2);
+        fflush(stdout);
 
+        /* --------------------------------------------------------- */
+        /* In the first test, we will merely try to see if we can run
+         * without crashing.  We don't check to see if data is good. */
```
jralls: This comment should be outside the loop.

jralls (changes requested):
@stefan-koch-micro Don't ask for a review until you've fixed the comments from my first review.

PR #2233, opened 2026-05-21
remove heavy libraries from #include

christopherlam:
culprits: regex, iostream, sstream, algorithm
should reduce compilation times

### DIFF
--- a/gnucash/gnome-utils/dialog-options.cpp
+++ b/gnucash/gnome-utils/dialog-options.cpp
@@ -43,8 +43,6 @@
 #include "gnc-ui.h" // for DF_MANUAL
 
 #include <any>
-#include <iostream>
-#include <sstream>
 
 #include "gnc-option-gtk-ui.hpp"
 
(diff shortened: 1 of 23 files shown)


jralls:
Reducing the compilation time to 1 minute by not compiling the whole program isn't a viable improvement. ;-)

MacOS uses clang and it has a different implementation of the C++ standard library. If a file uses a symbol but doesn't include the header that declares the symbol the file may still compile if some other header that the file includes does include the symbol's header. That might work on gcc but not clang, or clang but not gcc, because of the different standard libraries. Multiple includes of a single header don't change the compile time--after all, the declaration has to be provided somehow--and guard macros (`#ifndef FOO_H; #define FOO_H;/* rest of header */;#endif`) or `#pragma once` ensure that only one instance of the header code is part of the compilation unit (the source file plus all of the included code).

jralls:
BTW, [ccache](https://ccache.dev/) is a more reliable way to improve compile times than messing with headers.

PR #2234, opened 2026-05-25
Bug 799767 - POST /accounts route in rest-api example is broken

peterskim:
addAccount was rewritten to accept (name, currency, account_type_id, parent_account_guid, description, code), validate each field, look up the parent by GUID, and return gnucash_simple.accountToDict(account). The route handler was updated to pull the form fields and call it with the new signature.

README was updated to reflect these changes.


jralls:
Thanks!

PR #2235, opened 2026-05-26
ensure boost::regex is compiled at startup

christopherlam:
instead of per-call

### DIFF
--- a/gnucash/import-export/csv-imp/gnc-imp-props-price.cpp
+++ b/gnucash/import-export/csv-imp/gnc-imp-props-price.cpp
@@ -41,6 +41,7 @@
 #include <boost/regex/icu.hpp>
 #include <gnc-locale-utils.hpp>
 #include "gnc-imp-props-price.hpp"
+#include <ctre.hpp>
 
 namespace bl = boost::locale;
 
@@ -65,7 +66,8 @@ std::map<GncPricePropType, const char*> gnc_price_col_type_strs = {
 GncNumeric parse_amount_price (const std::string &str, int currency_format)
 {
     /* If a cell is empty or just spaces return invalid amount */
-    if(!boost::regex_search(str, boost::regex("[0-9]")))
+    static constexpr ctll::fixed_string digit_re{"[0-9]"};
+    if(!ctre::search<digit_re>(str))
         throw std::invalid_argument (_("Value doesn't appear to contain a valid number."));
 
     static const auto expr = boost::make_u32regex("[[:Sc:]]");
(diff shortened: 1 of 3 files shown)


gnucash/import-export/csv-imp/gnc-imp-props-price.cpp
```diff
@@ -65,7 +65,8 @@ std::map<GncPricePropType, const char*> gnc_price_col_type_strs = {
 GncNumeric parse_amount_price (const std::string &str, int currency_format)
 {
     /* If a cell is empty or just spaces return invalid amount */
-    if(!boost::regex_search(str, boost::regex("[0-9]")))
+    static const boost::regex digit{"[0-9]"};
```
gjanssens: Could it be constexpr instead of static const to generate the boost::regex at compile time ?

libgnucash/engine/gnc-datetime.cpp
```diff
@@ -722,12 +722,11 @@ GncDateImpl::GncDateImpl(const std::string str, const std::string fmt) :
         catch (...) {}          // with any string->date exception, try regex
     }
 
-    if (iter->m_re.empty())
+    if (!iter->m_re.has_value())
         throw std::invalid_argument ("No regex pattern available");
```
gjanssens, line 726: It may be faster to test an optional vs an empty boost::regex, but your commit message is not explaining very well why you introduce an optional here. You may want to split this in two commits or improve the wording of the commit message to also cover this.

christopherlam: didn't know about `empty()` thanks

gnucash/import-export/csv-imp/gnc-imp-props-price.cpp
christopherlam: `constexpr` doesn't seem to be possible for `boost::basic_regex`

jralls: In order for `constexpr func()` to compile, `func` has to have been declared `constexpr`. Almost nothing in boost::regex is declared `constexpr`.

gnucash/import-export/csv-imp/gnc-imp-props-price.cpp
```diff
@@ -65,10 +65,11 @@ std::map<GncPricePropType, const char*> gnc_price_col_type_strs = {
...
```
jralls: Let's make it clear everywhere that it's a regular expression by naming it (e.g. for this case) `digit_re` and so on.

gnucash/import-export/csv-imp/gnc-imp-props-price.cpp
jralls: To get compile-time regexp evaluation use CTRE as in #2228.

PR #2236, opened 2026-05-30
[qofid.cpp] use boost::unordered_flat_map

christopherlam:
to store guid->qof entities

note this is appropriate if the map values are QofInstance* pointers, because unordered_flat_map does not guarantee stable iterators.

my `time ./cli.sh` which loads my primary datafile and runs a report improves from around 4.1s to 3.4s

(thanks claude)

### DIFF
--- a/libgnucash/engine/qofid.cpp
+++ b/libgnucash/engine/qofid.cpp
@@ -27,9 +27,13 @@
 #include <config.h>
 #include <string.h>
 
+#include <guid.hpp>
 #include "qof.h"
 #include "qofid-p.h"
 #include "qofinstance-p.h"
+
+#include <boost/unordered/unordered_flat_map.hpp>
+using GuidEntityMap = boost::unordered_flat_map<GncGUID, QofInstance*, std::hash<GncGUID>>;
 
 static QofLogModule log_module = QOF_MOD_ENGINE;
 
@@ -38,17 +42,15 @@ struct QofCollection_s
     QofIdType    e_type;
     gboolean     is_dirty;
 
-    GHashTable * hash_of_entities;
+    GuidEntityMap guid_entity_map;
     gpointer     data;       /* place where object class can hang arbitrary data */
 
     QofCollection_s (QofIdType type) : e_type{static_cast<QofIdType>(CACHE_INSERT(type))}
                                      , is_dirty{FALSE}
-                                     , hash_of_entities{guid_hash_table_new()}
                                      , data{NULL} {}
     ~QofCollection_s ()
     {
         CACHE_REMOVE (e_type);
-        g_hash_table_destroy (hash_of_entities);
     }
 };
 
@@ -87,48 +89,30 @@ qof_collection_remove_entity (QofInstance *ent)
     col = qof_instance_get_collection(ent);
     if (!col) return;
     guid = qof_instance_get_guid(ent);
-    g_hash_table_remove (col->hash_of_entities, guid);
+    col->guid_entity_map.erase(*guid);
     qof_instance_set_collection(ent, NULL);
 }
 
 void
 qof_collection_insert_entity (QofCollection *col, QofInstance *ent)
 {
-    const GncGUID *guid;
-
     if (!col || !ent) return;
-    guid = qof_instance_get_guid(ent);
+    const GncGUID *guid = qof_instance_get_guid(ent);
     if (guid_equal(guid, guid_null())) return;
     g_return_if_fail (col->e_type == ent->e_type);
     qof_collection_remove_entity (ent);
-    g_hash_table_insert (col->hash_of_entities, (gpointer)guid, ent);
+    col->guid_entity_map[*guid] = ent;
     qof_instance_set_collection(ent, col);
 }
 
 gboolean
 qof_collection_add_entity (QofCollection *coll, QofInstance *ent)
 {
-    QofInstance *e;
-    const GncGUID *guid;
-
-    e = NULL;
-    if (!coll || !ent)
-    {
-        return FALSE;
-    }
-    guid = qof_instance_get_guid(ent);
-    if (guid_equal(guid, guid_null()))
-    {
-        return FALSE;
-    }
+    if (!coll || !ent) return FALSE;
+    const GncGUID *guid = qof_instance_get_guid(ent);
+    if (guid_equal(guid, guid_null())) return FALSE;
     g_return_val_if_fail (coll->e_type == ent->e_type, FALSE);
-    e = qof_collection_lookup_entity(coll, guid);
-    if ( e != NULL )
-    {
-        return FALSE;
-    }
-    g_hash_table_insert (coll->hash_of_entities, (gpointer)guid, ent);
-    return TRUE;
+    return coll->guid_entity_map.try_emplace (*guid, ent).second;
 }
 
 
@@ -208,22 +192,18 @@ qof_collection_compare (QofCollection *target, QofCollection *merge)
 QofInstance *
 qof_collection_lookup_entity (const QofCollection *col, const GncGUID * guid)
 {
-    QofInstance *ent;
     g_return_val_if_fail (col, NULL);
     if (guid == NULL) return NULL;
-    ent = static_cast<QofInstance*>(g_hash_table_lookup (col->hash_of_entities,
-							 guid));
-    if (ent != NULL && qof_instance_get_destroying(ent)) return NULL;	
-    return ent;
+    auto it = col->guid_entity_map.find(*guid);
+    if (it == col->guid_entity_map.end() || qof_instance_get_destroying(it->second))
+        return nullptr;
+    return it->second;
 }
 
 guint
 qof_collection_count (const QofCollection *col)
 {
-    guint c;
-
-    c = g_hash_table_size(col->hash_of_entities);
-    return c;
+    return col->guid_entity_map.size();
 }
 
 /* =============================================================== */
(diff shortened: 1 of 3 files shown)


jralls:
Requires Boost 1.81, which in turn means minimum Ubuntu of 24.04. That makes it time to create the `future` branch and plan for GnuCash 6.0 next spring.

jralls:
Wrong answer. Upgrading dependency minimum versions requires a new major version of GnuCash.

christopherlam:
#2237 was to create a `future` branch. I'll leave this for you.

jralls:
> https://github.com/Gnucash/gnucash/pull/2237 was to create a future branch. I'll leave this for you.

Sorry, that wasn't at all clear from the PR. There's a lot more to it than just updating the ubuntu version in CI. I'll have a PR ready in a couple of days.

libgnucash/engine/qofid.cpp
jralls, line 1: This seems unrelated to the boost::unordered_flat_map change. If you're going to improve memory management then I think you should go all-in and replace the raw `QofCollection*` with `std::unique_ptr<QofCollection>` held in QofBook and QofQueryPredData and used in test-qofid and test-qofinstance.

As part of the same effort you can make QofCollection a real class with member functions and no Glib artifacts.

jralls:
> I'll have a PR ready in a couple of days.

A bit of a delay while I get WebKitGtk 2.53.3 to build on macOS so that the non-Win32 build is consistent with Linux.

libgnucash/engine/qofid.cpp
christopherlam: I think this is slightly beyond my comfort zone for now... and eventually `QofCollection` should be accessible directly without `unique_ptr` ?

jralls: `unique_ptr` is strictly a memory-management tool. Accessors like `qof_book_get_collection` would use `unique_ptr::get()` to retrieve the underlying pointer, but only because they're called from C. Eventually, when the function isn't called from C anymore and has C++ binding they can return a `const unique_ptr&`.

christopherlam: > `unique_ptr` is strictly a memory-management tool. Accessors like `qof_book_get_collection` would use `unique_ptr::get()` to retrieve the underlying pointer, but only because they're called from C. Eventually, when the function isn't called from C anymore and has C++ binding they can return a `const unique_ptr&`.

The idea of avoiding `unique_ptr` is that eventually the `book->hash_of_collections->hash_of_entities` may eventually become flat_maps therefore avoiding need to chase pointers...

jralls:
> A bit of a delay while I get WebKitGtk 2.53.3 to build on macOS so that the non-Win32 build is consistent with Linux.

That failed, so I created #2242 with WebKitGtk 2.32.0 as the minimum version.

libgnucash/engine/qofid.cpp
jralls: That's fine when we get there, but in the meantime unique_ptr is the best way to prevent leaks when we have to allocate the memory ourselves.

christopherlam:
How about this, conditional compilation... 🤣

christopherlam:
also I'd like to merge in the small c++ ctor/dtor struct changes with `new` and `delete`; it'll be easier to reason about with other `map` changes using `unique_ptr` later on. otherwise I find the extra step when switching GHashTable to map difficult. If the conditional compilation was acceptable I think this PR could be merged as is.

libgnucash/engine/qofid.cpp
```diff
@@ -145,27 +136,11 @@ qof_collection_insert_entity (QofCollection *col, QofInstance *ent)
...
-    if (!coll || !ent)
-    {
-        return FALSE;
-    }
-    guid = qof_instance_get_guid(ent);
-    if (guid_equal(guid, guid_null()))
-    {
-        return FALSE;
-    }
+    if (!coll || !ent) return false;
```
jralls: Return type is `gboolean` so this should be `FALSE`. The distinction is for humans, compiler doesn't care; also to be consistent with line 141.

jralls:
> also I'd like to merge in the small c++ ctor/dtor struct changes with new and delete; it'll be easier to reason about with other map changes using unique_ptr later on. 

OK, with a couple of extra items: 
* Remove all of the collection code from qof-querycore.*. AFAICT it's not used.
* Fix the leaked `QofCollection*`s in gncInvoice. There are two, where `gncInvoiceGetEntries` (which returns the result of `qof_collection_new`) is used to initialize a for loop variable.

> If the conditional compilation was acceptable I think this PR could be merged as is.

I think it makes more sense to do this in `future` where the boost minimum is 1.90 and you have the option of using `std::flat_map` instead (assuming the implementation isn't screwed up).

libgnucash/engine/qofid.cpp
```diff
@@ -27,44 +27,73 @@
...
+
+using GuidEntityMap = boost::unordered_flat_map<GncGUID, QofInstance*>;
+#else
+
+template<> struct std::hash<GncGUID>
+{
+    std::size_t operator()(const GncGUID& g) const noexcept
+    {
+        std::size_t h;
+        std::memcpy(&h, g.reserved, sizeof(h) < sizeof(g.reserved) ? sizeof(h) : sizeof(g.reserved));
```
jralls: `size_t` isn't likely to grow beyond 64 bits (20 yottabytes of address space) in your or GnuCash's lifetime. But we've already got `guid_hash_to_guint(void*)` so better to use that or its guts `memcpy (&rv, &g->reserved[12], sizeof (uint32_t))` for compatibility.

christopherlam:
>     * Remove all of the collection code from qof-querycore.*. AFAICT it's not used.
>     * Fix the leaked `QofCollection*`s in gncInvoice. There are two, where `gncInvoiceGetEntries` (which returns the result of `qof_collection_new`) is used to initialize a for loop variable.

No it's `qofInvoiceGetEntries` and both above are removed with #2243

> I think it makes more sense to do this in `future` where the boost minimum is 1.90 and you have the option of using `std::flat_map` instead (assuming the implementation isn't screwed up).

see improved benchmarks with STL or Boost

from 4.0s to std::unordered_map 3.7s or boost::unordered_flat_map 3.3s

christopherlam:
hmm adding map.reserve() seems to make the benchmarks more consistent...

jralls:
I've created the `future` branch so you can switch this PR.

> hmm adding map.reserve() seems to make the benchmarks more consistent...

It should also make them a bit faster as there's only one allocation for the map.

jralls:
@christopherlam You've got too many commits here. Did you rebase your use-boost-flat-map onto future or do something else?

christopherlam:
> @christopherlam You've got too many commits here. Did you rebase your use-boost-flat-map onto future or do something else?

tried, yes... not sure which future to use... upstream?

jralls:
> tried, yes... not sure which future to use... upstream

Shouldn't matter, they should be the same. OTOH code is down right now so you're stuck with github. Anyway, good that you got it and recognized that future needs the g_module_build_path fix merged in...and I can't do that until code comes back.

jralls:
Arg. I pushed future from the wrong computer and I can't fix it with code down.

christopherlam:
I note this is already in `jralls/future` -- I think the `guid_entity_map.reserve()` call is slightly ugly. Other ideas welcome...

jralls:
> I note this is already in jralls/future 

It is, but I screwed up the merge commit so it fails to compile. I'll have another try soon.

> I think the guid_entity_map.reserve() call is slightly ugly. Other ideas welcome...

The call or the name?

christopherlam:
> > I note this is already in jralls/future
> 
> It is, but I screwed up the merge commit so it fails to compile. I'll have another try soon.
> 
> > I think the guid_entity_map.reserve() call is slightly ugly. Other ideas welcome...
> 
> The call or the name?

I mean the QofCollection ctor decides to reserve space for splits.

The other idea is in Split.cpp

```diff
@@ -1072,7 +1072,7 @@ xaccSplitLookup (const GncGUID *guid, QofBook *book)
 {
     QofCollection *col;
     if (!guid || !book) return nullptr;
-    col = qof_book_get_collection (book, GNC_ID_SPLIT);
+    col = qof_book_get_collection (book, GNC_ID_SPLIT, 10000);
     return (Split *) qof_collection_lookup_entity (col, guid);
 }
```

jralls:
The only place it makes sense to reserve memory is when loading the backend. In the SQL backend you can easily get a count of the number of splits you need from the database. It's possible in XML too, but very slow: You have to traverse the whole DOM and count the split elements.

jralls:
> It is, but I screwed up the merge commit so it fails to compile. I'll have another try soon.

Now fixed and in the public `future`. That makes for some conflicts that will need to be cleaned up here and in #2244.

christopherlam:
The problem with benchmarks is it's difficult to do right. The speedup above 4s->3.3s is true, but complicated by having a slow `Balance Sheet` report. 

FWIW the "Balance Sheet (Multicolumn)" is much faster -- "Balance Sheet" needs 1.0s, multi column completes in 0.15s.

I've added some fancy benchmarking in `qof_session_load` and running a report, and this branch does _not_ seem to speed up `qof_session_load` which remains 1.6s, but it speeds up running a Balance Sheet report from 1.6s to 1.0s.

So it seems somewhere in the engine<->SCM interface is hitting the `hash_of_entities` map hard.

SOLVED: the speedup is balance sheet's `qof_query_run` which improves from 1.6s to 1.0s

jralls:
Benchmarking is indeed hard to get right. That's why I keep pushing you towards using time-slice based profiling instead.

PR #2237, opened 2026-05-31
[ci-tests.yml] require minimum ubuntu 24.04

christopherlam:
like this as a new branch in code?
there's no libwebkitgtk-4.0-dev apparently


jralls:
No. The minimum versions of several packages--including boost--are anchored to Ubuntu 22.04 and CI is set to use that distribution specifically to catch PRs that require newer versions so that they can be shunted to the next major release.

PR #2239, opened 2026-06-07
Qofbook pimpl

christopherlam:
move `_QofBook` to cpp -- the first step in modernising struct hash tables


libgnucash/engine/qofbook.cpp
```diff
@@ -74,6 +74,91 @@ enum
     PROP_OPT_FY_END,                        /* KVP */
 };
 
+/* Book structure */
+struct _QofBook
```
jralls: Symbols beginning with `_` and a capital letter [is reserved to the compiler or standard library in C++](https://devblogs.microsoft.com/oldnewthing/20230109-00/?p=107685). ~~Gnome's `typedef` wheeze (where the struct is named with a leading underscore in a private header and then typedeffed to the underscore-less public name in the public one) isn't really appropriate in C++ because the `public:` and `private:` labels in the class/struct declaration enforce access rules while letting the compiler compute the object size for memory allocation. That also means that the class/struct declaration needs to be in the header where the compiler can see it when building other files that use the struct/class.~~

~~Making the member variables `private` and providing `public` accessors is fine, but they should be member functions instead of free ones, and those member functions wrapped in free functions if *and only if* C-binding access is needed. Both gnc-main-window and QofSession are already compiled with C++ so no free-function wrappers are needed for these three accessors.~~

**Edit:** Sorry, I went off on a tangent without considering that QofBook is a GObject class. The safest thing to do is to leave the struct decl in qofbook.h where it's wrapped in `extern "C"`. The accessors are OK.

christopherlam: But if it's C then we can't change ghashtable to an appropriate flavour of boost::map?

In this branch I'm not sure what to do about the gmock errors

jralls: Ah, *that's* what you're up to. As long as the hash table implementation doesn't leak to C it doesn't matter what it is. qofbook.h doesn't have any functions that return the hashtable so it should be OK. 

The GMock errors are from the mock class being derived from QofBook. That's done more or less the same way that it's done in GObject: A QofMockBook starts off with a QofBook and any additional members are tacked on after. To create one the compiler needs to see the complete declaration of a QofBook, and you've hidden that away in qofbook.cpp which isn't (and shouldn't be) included in gmock-qofbook.h. The usual workaround is to create qofbook-p.h, put the QofBook decl in there, and include it in qmock-qofbook.h. If the only thing that needs to know the size of a QofBook is QofMockBook then it will be used only by C++ code and you're safe. To make that clear to the world you can reasonably name it qofbook-p.hpp (and while you're at it also rename gmock-qofbook.h and gmock-Split.h to be hpp since they also work only with C++).

libgnucash/engine/test/test-qofbook.cpp
```diff
@@ -838,19 +838,20 @@ test_book_foreach_collection( Fixture *fixture, gconstpointer pData )
...
+    "void qof_book_foreach_collection(const QofBook *, QofCollectionForeachCB, gpointer)";
 #else
-#define _func "void qof_book_foreach_collection(const QofBook*, QofCollectionForeachCB, gpointer)"
-//#define _func "qof_book_foreach_collection"
+    "void qof_book_foreach_collection(const QofBook*, QofCollectionForeachCB, gpointer)";
 #endif
-    gchar *msg1 = _func ": assertion 'book' failed";
-    gchar *msg2 = _func ": assertion 'cb' failed";
+    std::string msg1 = std::string (func_sig) + ": assertion 'book' failed";
+    std::string msg2 = std::string (func_sig) + ": assertion 'cb' failed";
```
jralls: `func_sig` is big enough to fail SSO so create it as a `static const std::string` instead of a `constexpr const char*`, avoiding one of the temporary allocations... or leave it as a macro so that `msg1` and `msg2` are created at compile time as stack frame constants, no allocations at all.

libgnucash/engine/test/test-qofbook.cpp
```diff
@@ -838,19 +838,20 @@ test_book_foreach_collection( Fixture *fixture, gconstpointer pData )
...
 #undef _func
-    gchar *log_domain = "qof";
-    guint loglevel = G_LOG_LEVEL_CRITICAL | G_LOG_FLAG_FATAL, hdlr;
-    TestErrorStruct check1 = { loglevel, log_domain, msg1 };
-    TestErrorStruct check2 = { loglevel, log_domain, msg2 };
+    const gchar *log_domain = "qof";
+    auto loglevel = static_cast<GLogLevelFlags>(G_LOG_LEVEL_CRITICAL | G_LOG_FLAG_FATAL);
+    guint hdlr;
+    TestErrorStruct check1 = { loglevel, (char*) log_domain, (char*) msg1.c_str() };
+    TestErrorStruct check2 = { loglevel, (char*) log_domain, (char*) msg2.c_str() };
```
jralls: Use `const_cast<char*>` instead of a C-style cast.

libgnucash/engine/qofbook-p.hpp
```diff
@@ -41,10 +41,93 @@
 #include "qofid.h"
 #include "qofid-p.h"
 #include "qofinstance-p.h"
-#ifdef __cplusplus
-extern "C"
+
+
+/* Book structure */
```
jralls: We don't really need the comment...

jralls:
Looks good. Merge when ready.

libgnucash/engine/test/test-qofbook.cpp
christopherlam: Couldn't get it to compile as macro.

jralls:
```
diff --git a/libgnucash/engine/test/test-qofbook.cpp b/libgnucash/engine/test/test-qofbook.cpp
index e1d7c2b5fb..834b2dd519 100644
--- a/libgnucash/engine/test/test-qofbook.cpp
+++ b/libgnucash/engine/test/test-qofbook.cpp
@@ -838,20 +838,19 @@ test_book_foreach_collection( Fixture *fixture, gconstpointer pData )
     QofIdType my_type = "my_type", my_type2 = "my_type2";
     guint param = (guint) g_test_rand_int();
     /* GLib assertion messages which aren't filtered to make clang's output like gcc's */
-    constexpr const char* func_sig =
 #if defined(__clang__)
-    "void qof_book_foreach_collection(const QofBook *, QofCollectionForeachCB, gpointer)";
+#define _func "void qof_book_foreach_collection(const QofBook *, QofCollectionForeachCB, gpointer)"
 #else
-    "void qof_book_foreach_collection(const QofBook*, QofCollectionForeachCB, gpointer)";
+#define _func "void qof_book_foreach_collection(const QofBook*, QofCollectionForeachCB, gpointer)"
 #endif
-    std::string msg1 = std::string (func_sig) + ": assertion 'book' failed";
-    std::string msg2 = std::string (func_sig) + ": assertion 'cb' failed";
+    const char* msg1 = _func ": assertion 'book' failed";
+    const char* msg2 = _func ": assertion 'cb' failed";
 #undef _func
     const gchar *log_domain = "qof";
     auto loglevel = static_cast<GLogLevelFlags>(G_LOG_LEVEL_CRITICAL | G_LOG_FLAG_FATAL);
     guint hdlr;
-    TestErrorStruct check1 = { loglevel, (char*) log_domain, (char*) msg1.c_str() };
-    TestErrorStruct check2 = { loglevel, (char*) log_domain, (char*) msg2.c_str() };
+    TestErrorStruct check1 = { loglevel, const_cast<char*>(log_domain), const_cast<char*>(msg1) };
+    TestErrorStruct check2 = { loglevel, const_cast<char*>(log_domain), const_cast<char*>(msg2) };

     /* need this as long as we have fatal warnings enabled */
     g_test_log_set_fatal_handler ( ( GTestLogFatalFunc )handle_faults, NULL );
@@ -870,7 +869,7 @@ test_book_foreach_collection( Fixture *fixture, gconstpointer pData )
     qof_book_foreach_collection( fixture->book, NULL, (gpointer)(&param) );
     g_assert_true( !col_struct.col1_called );
     g_assert_true( !col_struct.col2_called );
-    g_assert_cmpstr( test_struct.msg, == , msg2.c_str());
+    g_assert_cmpstr( test_struct.msg, == , msg2);
     g_free( test_struct.msg );
     g_log_remove_handler (log_domain, hdlr);
     test_clear_error_list ();
```
Builds and tests for me. If you're getting `expression result unused` make sure you deleted the `;` from the end of the `_func` definition.

